PR #11671, opened 2018-03-26
extlib: blacklist 1.7.4 on 32-bit archs (broken)

ygrek:
ref https://github.com/ygrek/ocaml-extlib/issues/46


kit-ty-kate:
Thanks ! Is it also the case for previous versions or just this one ?

ygrek:
only this one

PR #11673, opened 2018-03-27
r2pipe is now deprecated

binoopang:
r2pipe is now deprecated. we recommend to use `radare2` package instead of `r2pipe`.


hcarty:
Thanks!

PR #11676, opened 2018-03-27
`conf-snappy`

gasche:
I got caught by the lack of a fedora depext for snappy in leveldb, so here is a conf-package for snappy.

### DIFF
--- a/packages/conf-snappy/conf-snappy.1/descr
+++ b/packages/conf-snappy/conf-snappy.1/descr
@@ -1,0 +1,2 @@
+Virtual package relying on snappy
+This package can only install if the snappy library is installed on the system.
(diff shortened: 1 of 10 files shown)


gasche:
It looks like Ubuntu/Debian don't actually provide a `pkg-config` package, so I'll have to write a compilation test instead.

gasche:
I included a short compile test in the package, waiting to see what CI results give.

I don't plan for now to fix the lint errors on the snappy.* and leveldb.* packages, which are independent of the current change.

gasche:
After some CI golfing fun, it looks like I fixed the problems in the package metadata: Travis and the V2.0 datakit are happy. I'm moving on to merge.

PR #11698, opened 2018-03-28
conduit-lwt-unix has depopt on lwt_ssl, not ssl

dra27:
While setting up cohttp with @FooB4r, we discovered that if you install `ssl` and then `lwt_ssl`, `conduit-lwt-unix` is not recompiled.

cc @avsm, @rgrinberg, @samoht


rgrinberg:
Yeah, that makes sense. Could you make this PR upstream as well?

dra27:
Good point - will do!

hannesm:
thanks!

PR #11664, opened 2018-03-26
Package haxe 3.4.7

andyli:
Initial packaging of Haxe.
It is built as a native binary so most dependencies are build only.
The use of `git:` in `url` is because there are git submodules which are not included in the Github archive.

PR #11700, opened 2018-03-28
Fix camlp4 for INSTALL_LOCAL test

dra27:
If camlp4 is required by the packages under test, then camlp4 needs to built, since the camlp4.*+system expect to find a working system camlp4, rather than building it themselves. There's a new universal of camlp4 in the pipeline which would allow this commit to be reverted in future, as long as the package constraints don't prevent its installation.

Should unblock #11664.


mseri:
If there is no objection, I think we should merge this and see how it goes

kit-ty-kate:
Yes, let's do that. Thanks !

PR #11707, opened 2018-03-30
KaSim 4.0

pirbo:


packages/KaSim/KaSim.4.0.0/opam
```diff
@@ -0,0 +1,39 @@
+opam-version: "1.2"
+name: "KaSim"
+version: "4.0"
```
hcarty: Version numbers should not be included in opam files - they are inferred from the parent directory name.

hannesm:
thanks, merged!

PR #11709, opened 2018-03-30
upper bound for calculon

c-cube:
see https://github.com/ocaml/opam-repository/pull/11701#issuecomment-377490120 (cc @kit-ty-kate)


kit-ty-kate:
Thanks ! :)

PR #11711, opened 2018-03-31
Camomile version 1.0.0

yoriyuki:
This is an update of Camomile.
	new file:   camomile.1.0.0/descr
	new file:   camomile.1.0.0/opam
	new file:   camomile.1.0.0/url


yoriyuki:
> Looks like your tests fail in some cases (see Travis)

Yes, they are all known issues.

> Also it seems that there are no versions of liquidsoap compatible with this release of Camomile. Could you add the appropriated constraints ?

Looking at https://github.com/savonet/liquidsoap/blob/master/configure.ac, I'm not sure why.

kit-ty-kate:
> Yes, they are all known issues.

Could you disable the tests then ?

yoriyuki:
> Could you disable the tests then ?

Did it but as you see still have build errors.  
We did some incompatible changes so some packages may not be able to compile with v.1.0.0
@rgrinberg Do you have any suggestion ?

kit-ty-kate:
@toots if you have the time I would rather prefer to have a new release of liquidsoap rather than having the latest patched. Users might want to have your latest fixes & camomile < 1.0.0

If you don't have time to make a release I think this is a case where you could have a new ```1.3.3-1``` package instead (where ```1.3.3-1``` would be ```1.3.3``` + patch and ```1.3.3``` would just require ```camomile < 1.0.0```)

toots:
@kit-ty-kate Will do the `1.3.3-1` then, and also add support for `OCaml 4.06`.

toots:
Ok, PR pushed. Didn't incorporate the changes to support `4.06`, it's too much.

yoriyuki:
Looks good.  All failures are caused by gasoline, patoline and Alpine linux.  (I don't checked arm64 because my browser complains its certificate)

kit-ty-kate:
Yes looks fine. Thanks a lot !

@yoriyuki can't you add an exception for the certificate ?

PR #11714, opened 2018-04-01
+bigstringaf.0.1.0

seliopou:


seliopou:
It looks like installation of the `m4`package is failing on Alpine.

hcarty:
Agreed, CI failures are the usual unrelated cases.

Thanks!

PR #11718, opened 2018-04-03
leveldb.1.0.3: Remove duplicated dev-repo

kit-ty-kate:
This was introduced in #11676 and caused opam2's opam1-to-opam2 format converter to fail on updates.

cc @gasche


gasche:
Hm, apologies... I thought that I had caught this mistake (but I guess it happened in two different opam files). Someday hopefully camelus will run on each force-push, to catch these errors more easily.

kit-ty-kate:
No worries. Regarding Camelus it think it already run on each force-push but I think it doesn't check this kind of error or maybe it's just opam2 related and I believe it doesn't use opam2.

PR #11719, opened 2018-04-03
Add genlet.a to the genlet.install file.

yallop:
Add a file that was missing from the installation instructions in the `genlet` package.


hannesm:
thanks!

PR #11712, opened 2018-03-31
graphql 0.6.0, graphql-lwt 0.6.0 and ppx_graphql 0.2.0

andreas:

PR #11720, opened 2018-04-03
angstrom < 0.6.0 are not safe-string safe

hannesm:
/cc @seliopou (but newer releases play fine with 4.06.0)

/cc @andreas discovered while browsing through CI issues in #11712


hannesm:
imho good to merge, there are remaining revdeps issues (when angstrom got split into multiple packages, these need some constraints AFAIU)... I won't have time to fix this in here...

hcarty:
@hannesm Agreed that this seems good to merge

PR #11716, opened 2018-04-02
Constrain packages for Lwt 4.0.0

aantron:
See https://github.com/ocaml/opam-repository/pull/11708#issuecomment-377951313.

cc @hannesm 

I'll rebase the "main" Lwt 4.0.0 PR over this, once this is merged.

PR #11721, opened 2018-04-04
More constraints for Lwt 4.0.0

aantron:
See #11708, #11716. cc @hannesm


aantron:
```
[ERROR] async < 111.25.00 & async >= 113.00.00 & lwt (< 2.5.0 | >= 4.0.0) &
        js_of_ocaml >= 3.0 is not a valid conjunction
```

Does this mean I can't write conflicts using `|`, and must use two conflicts entries to get the same effect?

kit-ty-kate:
Seems like this syntax works only with opam2 but yes I would say. Sorry :/
@AltGr can you confirm ?

aantron:
I converted the constraints not to use `|`. The build failures are not related to the new constraints, so I would say this is ready for merge.

kit-ty-kate:
Yep, thank you ! :)

PR #11722, opened 2018-04-04
ounit.2.0.8

gildor478:


packages/ounit/ounit.2.0.8/opam
```diff
@@ -0,0 +1,16 @@
...
+authors: [ "Maas-Maarten Zeeman" "Sylvain Le Gall" ]
+maintainer: "https://github.com/ocaml/opam-repository/issues"
+homepage: "http://ounit.forge.ocamlcore.org"
+bug-reports: "https://forge.ocamlcore.org/tracker/?func=browse&group_id=162&atid=730"
+dev-repo: "https://github.com/gildor478/ounit.git"
+doc: ["http://ounit.forge.ocamlcore.org/api-ounit/index.html"]
+build: [make "build"]
+remove: [["ocamlfind" "remove" "oUnit"]]
+depends: [
+  "ocamlfind"
```
kit-ty-kate: does ounit use the ```Findlib``` library or just the ocamlfind binary ? In the later case you might want to add a ```{build}``` attribute there.

gildor478: I use ocamlfind at build time and to "remove" the package. So I am really not sure that "ocamlfind" is only for "build"ing.

gildor478: Done/

PR #11723, opened 2018-04-04
oasis.0.4.11

gildor478:


packages/oasis/oasis.0.4.11/opam
```diff
@@ -0,0 +1,32 @@
...
+]
+install: ["ocaml" "setup.ml" "-install"]
+remove: [
+  ["ocaml" "%{etc}%/oasis/setup.ml" "-C" "%{etc}%/oasis" "-uninstall"]
+]
+build-doc: [ "ocaml" "setup.ml" "-doc" ]
+depends: [
+  "base-unix"
+  "ocamlbuild"
+  "ocamlfind" {>= "1.3.1"}
```
kit-ty-kate: does ounit use the ```Findlib``` library or just the ocamlfind binary ? In the later case you might want to add a ```{build}``` attribute there.

gildor478: Same as for ounit, the remove operation will use ocamlfind to remove libraries when "removing" the package.

kit-ty-kate: ok we should really write that somewhere in bold letters or something:
```
Build depends: you may add the key build in front of the version constraints, e.g. "package" {build & >= "3.2"}, to indicate that there is no run-time dependency to this package: it is required but won't trigger rebuilds of your package when changed.
```
(source: https://opam.ocaml.org/doc/Packaging.html)

In summary the ```{build}``` attribute just tells the solver to not rebuild its reverse dependencies when it gets updated but it stays a dependency that cannot be removed without removing all of its reverse dependencies.

gildor478: So I have fixed the code. BUT oasis is kind of strange because that is a build tool. So it actually uses ocamlfind/ocamlbuild at runtime.

I think I should keep ocamlfind/ocamlbuild.

What do you think?

kit-ty-kate: Again, no. ```ocamlbuild``` is fine because the ocamlbuild plugin contains call to the ocamlbuild library so it needs to be rebuilt on updates et so on but I can't see any uses of the ```Findlib``` library so really ```ocamlfind``` should be marked as with a  ```{build}``` attribute since, again, this will still be a regular "runtime" dependency but oasis doesn't need to be rebuilt every time ```ocamlfind``` gets updated.

PR #11726, opened 2018-04-04
oasis2debian.0.1.4

gildor478:


gasche:
The `build-test` test is attempted by the Travis CI and fails (see [this log example](https://travis-ci.org/ocaml/opam-repository/jobs/362391735)). It looks like 4 out of the 5 tests fail, with error logs such as:

```
File "/home/travis/.opam/4.04.2/build/oasis2debian.0.1.4/_build/oUnit-oasis2debian-travis-job-9558937e-99f8-4593-9ad9-547abb31d9fa#01.log", line 263, characters 1-1:
Error: oasis2debian:1:ocamlmod-0.0.3.tar.gz (in the log).

File "test/test.ml", line 118, characters 1-1:
Error: oasis2debian:1:ocamlmod-0.0.3.tar.gz (in the code).
```

Do you have an idea of why these tests fail on the CI machines?

gildor478:
@gasche These tests are trying to build debian packages. The Travis CI probably doesn't have the right tool to build the Debian packages (i.e. dpkg-buildpackage exec). Should I disable the tests?

gasche:
I think the idea of the `build-test` field in the opam package is to enable some kind of large-scale automated testing (for example: checking that a change in the compiler does not break user code, where "break" checks not only the build but also that the package still work somewhat correctly), so tests that only work in very restricted environment should not be listed there indeed.

(Note that it is possible to have `{test}` qualifiers on opam dependencies, so tests depending on additional OPAM packages compared to normal build can be made to work fine. But I suppose the debian tools are not exposed to the opam universe, even as `conf-*` packages.)

gildor478:
@gasche I have deactivate it for now and will look at that later ocaml/oasis2debian#4

I think the PR is good to go as is.

gasche:
I think so as well, I'll just wait for the results of the automated CI.

gasche:
The CI fails on 4.01 and older releases because of a build failure in `debian-formats` (which advertises `>= 3.12.1` on opam):

```
File "src/DFWatch.ml", line 28, characters 15-26:
Error: Unbound value String.trim
```

The `String` module of the standard library includes `trim` since 4.00, but the failure may be caused by the use of `extlib-compat 1.6.1`, whose [String module](https://github.com/ygrek/ocaml-extlib/blob/1.6.1/extlib/extString.mli) does not re-export `trim` -- they were only made available in 1.7.0.

kit-ty-kate:
Tests looks good. However looking at the opam file I find it a bit doubtful to have ~every single dependencies as build dependencies. Looking at the git repository it seems that most of them are used to build the executable. I don't think you'd want oasis2debian to be out-of-sync with the oasis library for example.

gildor478:
Well, I am fine with going out of sync this way.

gasche:
@kit-ty-kate feel free to merge once you have decided whether you like the `opam` file or not :-)

gildor478:
Friendly ping. Can we merge this PR?

kit-ty-kate:
Sorry I've been busy on other things and I missed the first ping by @gasche.
Sorry but don't really like that. Having out-of-sync dependencies is really a mess to debug and I feel is just not "right".
Do you expect this tool to behave differently with different versions of oasis or are you confident enough to be sure this won't have different behaviours ?

Anyway I just noticed the previous versions are also defined with build depends so I'll merge it and we can discuss later on that topic. Thanks and sorry for the delay.

PR #11727, opened 2018-04-05
New package: inferno.

fpottier:


fpottier:
Hello,
All tests seem to fail with the error message `Couldn't extract the version number from inferno-20180405`. I am not sure what's wrong?

hcarty:
The directory name should be in the form `name.version` rather than `name-version`

Could you add a license field to the opam file?

PR #11729, opened 2018-04-05
Mark ocamlfind as incomptible with OCaml >= 4.07

kit-ty-kate:
Fixed upstream in https://gitlab.camlcity.org/gerd/lib-findlib/merge_requests/10

### DIFF
--- a/packages/ocamlfind/ocamlfind.1.3.2/opam
+++ b/packages/ocamlfind/ocamlfind.1.3.2/opam
@@ -20,5 +20,5 @@ depends: [
   "conf-m4" {build}
   "num" {= "0"}
 ]
-available: ocaml-version > "3.12.2"
+available: [ocaml-version > "3.12.2" & ocaml-version < "4.07"]
 install: [make "install"]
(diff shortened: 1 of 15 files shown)


kit-ty-kate:
cc @gerdstolpmann

packages/ocamlfind/ocamlfind.1.3.2/opam
```diff
@@ -20,5 +24,5 @@ depends: [
   "conf-m4" {build}
   "num" {= "0"}
 ]
-available: ocaml-version > "3.12.2"
+available: [ocaml-version > "3.12.2" & ocaml-version < "4.07"]
```
mseri: The opam file of 1.3.3 says
``` 
available: [(ocaml-version >= "3.08") & (ocaml-version <= "4.01.0")] # ocamlfind uses Arg.align of 3.08
```
I think 1.3.2 should be at least as restricted as 1.3.3

kit-ty-kate: You're probably right. Done

mseri:
What is going on with the CI? I think that this could be merged otherwise

PR #11731, opened 2018-04-05
New release of Earley / Earley-OCaml + small version fix in earlier version

rlepigre:


packages/earley-ocaml/earley-ocaml.1.0.1/opam
```diff
@@ -1,5 +1,5 @@
 opam-version: "1.2"
-available   : [ ocaml-version >= "4.03.0" ]
```
kit-ty-kate: Why ```<=``` ? Did you mean ```<``` instead ?

rlepigre: No, I exactly meant `<=`. This package requires per-version stuff (including minor versions), and at the time of the previous release there was no `4.06.1` (which fails to build). For the more recent version I added an upper-bound right away, to make sure that this will not happen again.

kit-ty-kate:
Ok fair enough. Thanks !

rlepigre:
Thank you!

PR #11739, opened 2018-04-09
Package Lascar 0.5

jserot:
Second try (after canceling the first PR).

### DIFF
--- a/packages/lascar/lascar.0.5/descr
+++ b/packages/lascar/lascar.0.5/descr
@@ -1,0 +1,1 @@
+A library for manipulating Labeled Transition Systems in OCaml
--- a/packages/lascar/lascar.0.5/opam
+++ b/packages/lascar/lascar.0.5/opam
@@ -24,6 +24,6 @@ remove: [
 ]
 depends: [
   "camlp4"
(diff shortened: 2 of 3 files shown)


jserot:
Build now seems ok but the so-called "distros" step fail.
Had a look at the logs : the pb seems to be on tools / path i don't have control on.
For ex, the "V1.2 Common Distro" step fails with "m4: not found".. 
What to do now ?

kit-ty-kate:
I added the ```{build}``` attribute to ocamlfind because you seem to use only ocamlfind and not Findlib which means you don't need to rebuild your package every time ocamlfind gets updated or changed.
If you can change that in the local opam file in your repository it would be perfect (otherwise you are probably going to forget this change next time you make a release).

Thanks !

jserot:
Thanks !
I'll do the change in the local opam file (altough i'm not sure to fully understand what it means).
I also noticed that the rules "make lib" and "make install-lib" have been renamed "make" and make "install" resp. Is there some specific reason for this ? The fact is that "make" and "make install" may build (resp. install) also non-caml parts, so that are may not be required for OPAM.
Thanks again

PR #11741, opened 2018-04-09
More constraints for Lwt 4.0.0

aantron:
See #11708.

Issue #10531, opened 2017-10-25
package revisionism can break things


toolslive:
Sometimes, a package is patched after a release, and then it's released again, with **the same** version.

This practice makes it hard to track dependency changes, but it also (probably) changes the checksums of the modules in the package. As a consequence, this can break the usage of this module in a dynamically loadable plugin context. It has happened quite a few times for us, and I think it's just as easy to just release a new version whenever the code changed.

As a concrete example, 
the [recent patching of oUnit.2.0.0](https://github.com/ocaml/opam-repository/blob/master/packages/ounit/ounit.2.0.0/files/safe-string.patch#L55) causes a

```
   Dynlink.Error "interface mismatch on OUnitConf"
```

hannesm:
I'd be very much in favour if you add custom patches to a released package, bump its version (a lengthy discussion is https://github.com/ocaml/opam-repository/pull/10350#issuecomment-332588180).  there's no harm in having `foo-1.0-1` etc. around, and makes it obvious which package is actually installed.  if the source checksum of a package changes while that package is released to opam-repository, there shoud as well be a version bump (at least a `-n` suffix, and the old version may need to be removed).

if the maintainers of opam-repository agree with this ("no patches / modification of released packages"), it should be communicated in CONTRIBUTING.md.  It would make this opam repository more immutable (easier to debug issues).  the downside is that upon a new version opam recompiles all reverse dependencies - but imho this is required (as @toolslive outlined above).

hcarty:
I ran into something similar today - I tried to `opam update -u` which triggered a rebuild of effectively everything due to a change to ocamlfind 1.7.3 (the version I already have/had installed).  The result left my switch in a broken state, with Lwt unable to build for example.

The fix was to install a fresh switch, but in this case the error message was not very helpful, saying that Lwt's configuration script could not build a basic toplevel.

pmetzger:
MacPorts has a notion of a revision number for all packages. Packages start (naturally) at revision number 0. If you make any substantive change to a package (altering dependencies, patches, anything else), you bump the rev number. When the package gets upgraded to the next maintainer release, you put the rev number back to 0 again. For simplicity, the revision field is assumed to be 0 if it isn't listed in the Portfile, so in the usual case (in which nothing gets changed between maintainer releases), you don't have to edit anything at all.

toolslive:
https://github.com/ocaml/opam-repository/commit/b610e6dd913b027a4deb6a6faf2973edfee671ba
[...]

PR #11742, opened 2018-04-09
bson: not available with safe-string

avsm:
noticed in revdeps for #11741


kit-ty-kate:
cc @MassD 

One other solution to this would be to prefix the build phase with ```env OCAMLPARAM=safe-string=0,_``` because people are complaining about this here: https://github.com/ocaml/opam-repository/issues/10531

hannesm:
thanks, merging.

PR #11744, opened 2018-04-09
modify 4.07 compilers to point to the newly created 4.07 branch

avsm:
This fixes these packages as trunk is now 4.08
cc @damiendoligez


damiendoligez:
Note to OPAM maintainers: don't bother making a switch for 4.07.0+beta1, I'm already making a beta2. I'll submit a PR for the switches when I'm done.
@ocaml/opam-repository

PR #11751, opened 2018-04-10
Fix package compilers

rjbou:
This PR contains some fix for ocaml compilers: 
  - Configure option
     * ocaml-variants.4.06.1+trunk+safe-string
     * ocaml-variants.4.07.0+trunk+unsafe-string
  - Move install command on install field (fail due to sandboxing)
    * ocaml-variants.3.09.1+metaocaml
    * ocaml-variants.4.00.1+BER
    * ocaml-variants.4.00.1+mirage-xen
    * ocaml-variants.4.00.1+mirage-xen
    * ocaml-variants.4.01.0+BER
    * ocaml-variants.4.02.1+BER
    * ocaml-variants.4.02.1+modular-implicits-ber
    * ocaml-variants.4.04.0+BER
    * ocaml-variants.4.04.0+afl


avsm:
Thanks for this extremely useful fix! The change is only going into the 2.0.0 branch -- would it be possible to have the fixes into the `master` branch (for opam 1.2) so that the translation bot can then apply them to the 2.0 series?  That will fix all the opam clients :)

packages/ocaml-variants/ocaml-variants.4.06.1+trunk+safe-string/opam
```diff
@@ -11,14 +11,14 @@ conflict-class: "ocaml-core-compiler"
 flags: compiler
 setenv: CAML_LD_LIBRARY_PATH = "%{lib}%/stublibs"
 build: [
-  ["./configure" "-prefix" prefix "-with-debug-runtime" "-safe-string"]
```
kit-ty-kate: This switch is uninstallable because the ```-safe-string``` option doesn't exists. Could you rename it to make it consistent with the stable variants (i.e. ```4.06.1+trunk+force-safe-string```). Also ```4.06.1+trunk+default-unsafe-string``` is missing but I don't know if it change something at this point.

packages/ocaml-variants/ocaml-variants.4.07.0+trunk+unsafe-string/opam
```diff
@@ -12,14 +12,15 @@ conflict-class: "ocaml-core-compiler"
 flags: compiler
 setenv: CAML_LD_LIBRARY_PATH = "%{lib}%/stublibs"
 build: [
-  ["./configure" "-prefix" prefix "-with-debug-runtime" "-unsafe-string"]
```
kit-ty-kate: Same comment as the above + I think having ```4.07.0+trunk+default-unsafe-string``` would be good here though.

rjbou:
> The change is only going into the 2.0.0 branch -- would it be possible to have the fixes into the master branch (for opam 1.2) so that the translation bot can then apply them to the 2.0 series?

@avsm Commit d2c22083c4d8fc60c23a0c05d0abcb269637f72b can not be ported to `master` branch. In 1.2 opam file, there is no differentiation between `build` and `install` commands. 

@kit-ty-kate @avsm I'll port changes of commit c509cd9b5c8bf8172b1a8471ebf9ec7285ca666e on master, and in the same time change the name of the compilers on both branches. Like that, there is no inconsistency/conflicts and it is possible to refuse camelus PR.

kit-ty-kate:
Seems good now. Thanks a lot !

PR #11753, opened 2018-04-10
Fix old ocaml compat

yoriyuki:
fix a bug https://github.com/yoriyuki/Camomile/issues/68 in 1.0.0 which prevents build by OCaml <= 4.04.0


kit-ty-kate:
Thanks ! So the problem appeared on OSX only right ?

yoriyuki:
I'm not sure which problem you mention.  If you mention the problem which 1.0.1 fixes, this is an `ar` issue on Mac but potentially others.  If you mention build errors in CI of opam, CentOS7 also fails.  opam fails even before going to build Camomile on CentOS7, looks failing resolve dependency but not sure.

PR #11755, opened 2018-04-10
Add compilers 4.07.0+beta2 and 4.08.0+trunk

damiendoligez:


kit-ty-kate:
Could you make the same changes than #11757 (i.e. rename & fix the options) ? Otherwise ```4.07.0+beta2+unsafe-string``` will fail to compile.

PR #11757, opened 2018-04-10
Compiler string option

rjbou:
This PR is related to PR #11751. It fixes safe string option for some 1.2 compilers, and change the name accordingly.


kit-ty-kate:
Seems good. Thanks ! :)

PR #11761, opened 2018-04-10
add download2.camlcity.org mirrors

fdopen:
download.camlcity.org is down far too often ...

### DIFF
--- a/packages/ocamlfind/ocamlfind.1.5.6/url
+++ b/packages/ocamlfind/ocamlfind.1.5.6/url
@@ -1,2 +1,3 @@
 archive: "http://download.camlcity.org/download/findlib-1.5.6.tar.gz"
 checksum: "91585dd5459cb69bfd9a0689bf222403"
+mirrors: [ "http://download2.camlcity.org/download/findlib-1.5.6.tar.gz" ]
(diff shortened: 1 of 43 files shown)


packages/ocamlfind/ocamlfind.1.5.6/url
```diff
@@ -1,2 +1,3 @@
 archive: "http://download.camlcity.org/download/findlib-1.5.6.tar.gz"
 checksum: "91585dd5459cb69bfd9a0689bf222403"
+mirrors: [ "http://download2.camlcity.org/download/findlib-1.5.7.tar.gz" ]
```
hcarty: Mismatch here - `1.5.7` in the mirror, should be probably be `1.5.6`.

fdopen: Indeed. Fixed by a forced push.

hcarty: Thanks!

hcarty:
I think this is OK to merge as it only affects opam's fallback if the main download site is down.  I'd like another reviewer to approve (and/or merge) though, out of a general sense of paranoia.

hannesm:
thanks, lgtm

PR #11754, opened 2018-04-10
ctypes: lwt < 4.0.0 for tests

fdopen:
The test instructions refer to `lwt.preemptive` that is gone now ...
CC @yallop

PR #11766, opened 2018-04-11
mariadb: ocamlbuild is a build dependency

fdopen:
spotted here: https://github.com/ocaml/opam-repository/pull/11754
reported upstream: https://github.com/andrenth/ocaml-mariadb/pull/21


kit-ty-kate:
Thanks !

PR #11780, opened 2018-04-12
Actually opam-lib.1.3.1 seems to work for 4.06

samoht:
This partially reverts #10988 @minishrink do you remember why this constraint was needed?


hannesm:
merging, travis is happy with their 4.06.0 builds!  thanks!

PR #11782, opened 2018-04-12
Package dune-release.0.1.0

samoht:
### `dune-release.0.1.0`

Release dune packages in opam

`dune-release` is a tool to streamline the release of Dune packages in
[opam](https://opam.ocaml.org). It supports only projects be built
with [Dune](https://github.com/ocaml/dune) and released on
[GitHub](https://github.com).

---
* Homepage: https://github.com/samoht/dune-release
* Source repo: https://github.com/samoht/dune-release.git
* Bug tracker: https://github.com/samoht/dune-release/issues

---

---
## 0.1.0 (2018-04-12)

Initial release.

Import some code from [topkg](http://erratique.ch/software/topkg).

- Use of `Astring`, `Logs`, `Fpath` and`Bos` instead of custom
  re-implementations;
- Remove the IPC layer which is used between `topkg` and `topkg-care`;
- Bundle everything as a single binary;
- Assume that the package is built using [dune](https://github.com/ocaml/dune);
- Do not read/neeed a `pkg/pkg.ml` file.

### DIFF
--- a/packages/dune-release/dune-release.0.1.0/descr
+++ b/packages/dune-release/dune-release.0.1.0/descr
@@ -1,0 +1,6 @@
+Release dune packages in opam
+
+`dune-release` is a tool to streamline the release of Dune packages in
+[opam](https://opam.ocaml.org). It supports only projects be built
+with [Dune](https://github.com/ocaml/dune) and released on
+[GitHub](https://github.com).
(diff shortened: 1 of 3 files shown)


samoht:
Of course, this was generated by doing `dune-release tag && dune-release bistro` :-)

samoht:
/cc @ocaml/dune-dev 

The dependency cone of this thing is way too big to be included in `dune` as it is today, but I am interested to make it more self-contained.

Note: an other option is to call the tool `opam-release` and install it as a plugin (so it will be installed just once for all the switches and then having a lot of depencies is not so important). Not sure what @AltGr things about that.

dbuenzli:
> an other option is to call the tool opam-release

`opam` is supposed to be build-system agnostic.

hannesm:
thanks, merging.

diml:
@samoht, thanks, this looks nice!

Making it fully self contained seems like it would be a lot of work, I'm not sure it's worth doing. Having it as a separate package doesn't seem too bad to me.

PR #11784, opened 2018-04-13
camltc.0.9.5 (Lwt.4.0.0)

toolslive:


packages/camltc/camltc.0.9.5/opam
```diff
@@ -0,0 +1,32 @@
...
+]
+
+remove: [
+  ["ocamlfind" "remove" "camltc"]
+]
+
+depends: [
+  "ocamlfind" {build}
+  "ounit"
+  "lwt" {>= "4.0.0"}
```
aantron: @toolslive, I glanced at the update of camltc. It looks to me like the constraint should be `>= 3.2.0`: the only new change that requires a lower bound is using `Lwt_preemptive` without `lwt.preemptive`, and that is possible since 3.2.0.

toolslive:
Is there a way to inspect what happened to @camelus ?

packages/camltc/camltc.0.9.5/url
```diff
@@ -0,0 +1,2 @@
+git: "https://github.com/toolslive/camltc.git#64dd82ab1586b81b2939839bc2d3cba9ecee9a2a"
```
hannesm: in the opam-repository, we only want releases, not versions pointing to a git commit.. could you please do a release and point to a tarball and checksum.  thanks!

toolslive: The reason the git commit is used here is that the build uses a git submodule for tokyo cabinet.
So the build would need to be changed to fetch a tarbal iso a git submodule.
What's the reason you don't want versions pointing to a git commit ?
Btw, we've had this discussion before:
https://github.com/ocaml/opam-repository/pull/7661

kit-ty-kate:
Fine. I still don't like having to fetch git submodules here but since it was done before I guess it doesn't change anything. Thanks.

Is there no other way to do otherwise ? Like having a custom archive somewhere or depend on a compiled package of tokyocabinet ?

PR #11788, opened 2018-04-13
Constrain js_of_ocaml for Lwt 4.0.0

aantron:
cc @hannesm @hhugo

Constraints are based on looking in the git history of js_of_ocaml. The dependency on `lwt.log` first appeared in 2.3, and seems not to be guarded by its own condition.

EDIT: see https://github.com/ocaml/opam-repository/pull/11708#issuecomment-381119399.

PR #11793, opened 2018-04-14
Use Lwt 2.4.3 GitHub archive

aantron:
`http://ocsigen.org/download/lwt-2.4.3.tar.gz` seems to no longer be available (@hannesm in https://github.com/ocaml/opam-repository/pull/11788#issuecomment-381311848).

Please don't merge this right away. We should probably try to get `ocsigen.org` to restore the archive first. cc @Drup, @balat.


mseri:
I think we could merge this if there is no followup

aantron:
I have no objection, thanks for the reminder :)

PR #11797, opened 2018-04-15
Package freetds.0.5.2

Chris00:
### `freetds.0.5.2`

Binding to the FreeTDS library

FreeTDS is a set of libraries for Unix and Linux that allows your
programs to natively talk to Microsoft SQL Server and Sybase
databases.

---
* Homepage: https://github.com/kennknowles/ocaml-freetds
* Source repo: https://github.com/kennknowles/ocaml-freetds.git
* Bug tracker: https://github.com/kennknowles/ocaml-freetds/issues

---

---
0.5.2 2018-04-15
----------------

- Better error messages (commits #94134f7 and #bdc2c99)
- Port to Jbuilder/Dune (Brendan Long) and topkg
- Add basic unit tests (Brendan Long)
- Fix all build warnings
- Improve Travis CI scripts

### DIFF
--- a/packages/freetds/freetds.0.5.2/descr
+++ b/packages/freetds/freetds.0.5.2/descr
@@ -1,0 +1,5 @@
+Binding to the FreeTDS library
+
+FreeTDS is a set of libraries for Unix and Linux that allows your
+programs to natively talk to Microsoft SQL Server and Sybase
+databases.
--- a/packages/freetds/freetds.0.5.2/opam
+++ b/packages/freetds/freetds.0.5.2/opam
@@ -1,0 +1,35 @@
+opam-version: "1.2"
+authors: [ "Christophe Troestler <Christophe.Troestler@umons.ac.be>"
+           "Kenn Knowles <kenn.knowles@gmail.com>" ]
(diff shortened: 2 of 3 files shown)


Chris00:
There is a failure about `m4: not found`, another about OPAM repository conversion, and then the problem that, on CentOS, `No package freetds-devel available.` and it is not clear it is available in the default repo.

brendanlong:
Hmm I'm not sure why we even need `m4` anymore. We could add it as a depext if we needed to but I thought Dune replaced that?

`freetds-devel` doesn't seem to be in the case CentOS repo but is in EPEL: https://mirrors.sonic.net/epel/7/x86_64/Packages/f/

Chris00:
> Why do you need ocamlfind?

We don't — but it is installed nonetheless in the container.

brendanlong:
Ohh I think `ocamlfind` is a transitive dependency of `ounit`: https://opam.ocaml.org/packages/ounit/ounit.2.0.8/

I'm still really confused about this problem though, since I would expect a common package like ounit to work fine :\

mseri:
I think this should be merged. All the failures are due to known infrastructure issues

PR #11799, opened 2018-04-16
 bddapron.*: handle safe strings of OCaml ≥ 4.06

nberth:


packages/bddapron/bddapron.2.3.2/opam
```diff
@@ -0,0 +1,25 @@
...
+install: [
+  [make "install"]
+]
+remove: [
+  ["ocamlfind" "remove" "bddapron"]
+  ["rm" "-r" "-f" "%{bddapron:doc}%"]
+]
+tags: [ "flags:light-uninstall" ]
+depends: [
+  "ocamlfind"
```
yomimono: It seems likely that the `build` tag should be here as well.

kit-ty-kate: I checked manually and there is no use of the `Findlib` library. Edited.

kit-ty-kate:
Looks good. Thanks !

PR #11803, opened 2018-04-17
Package opam-publish.2.0.0~beta

AltGr:
### `opam-publish.2.0.0~beta`

---
* Homepage: https://github.com/ocaml/opam-publish
* Source repo: git+https://github.com/ocaml/opam-publish.git
* Bug tracker: https://github.com/ocaml/opam-publish/issues

---

### DIFF
--- a/packages/opam-publish/opam-publish.2.0.0~beta/opam
+++ b/packages/opam-publish/opam-publish.2.0.0~beta/opam
@@ -14,14 +14,15 @@ depends: [
   "opam-core" {build & >= "2.0.0~beta5"}
   "opam-format" {build & >= "2.0.0~beta5"}
   "opam-state" {build & >= "2.0.0~beta5"}
-  "ocamlfind" {build}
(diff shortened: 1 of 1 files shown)


rgrinberg:
Curious why jbuilder isn't listed as a dependency here. And for that matter why is ocamlfind a dependency?

AltGr:
The Alpine failure is unrelated, but a note about it:
it seems their provided `patch` command fails to deal with the huge patch generated by the repository update (`opam update` uses `git` to generate a patch then applies it to its repository, possibly after validation of signatures using `conex`).

Also, the fact that the patch is so big is in part because here:
```
Step 6/7 : RUN git remote add local git://github.com/ocaml/opam-repository &&   git fetch local pull/11803/head &&   git merge e7a366ca752220dc932060435e126a7d88b91ab3 &&   opam update -uy
```
the merge is done with the master rather than the 2.0.0 branch.

PR #11804, opened 2018-04-17
conf-openblas: update to latest homebrew package for OS X

mor1:


packages/conf-openblas/conf-openblas.0.2.0/files/centos_install.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+yes | yum update
+yes | yum install git
+yes | yum install gcc gcc-c++ gcc-gfortran make kernel-devel
+
+git clone https://github.com/xianyi/OpenBLAS.git
```
mseri: Could we shallow-clone a specific version of OpenBLAS?

packages/conf-openblas/conf-openblas.0.2.0/files/centos_install.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+yes | yum update
```
mseri: You can do `yum update -y` and `yum install -y git gcc gcc-c++ gcc-gfortran make kernel-devel`

packages/conf-openblas/conf-openblas.0.2.0/files/centos_install.sh
```diff
@@ -0,0 +1,11 @@
...
+yes | yum install git
+yes | yum install gcc gcc-c++ gcc-gfortran make kernel-devel
+
+git clone https://github.com/xianyi/OpenBLAS.git
+cd OpenBLAS
+make && make install
```
mseri: Does it require sudo for the `make install` step?

packages/conf-openblas/conf-openblas.0.2.0/files/centos_install.sh
mor1: Sure

packages/conf-openblas/conf-openblas.0.2.0/files/centos_install.sh
mor1: ok

packages/conf-openblas/conf-openblas.0.2.0/files/centos_install.sh
mor1: TBH no idea, this is left as-is from the 0.1 version.
/cc @ryanrhymes who wrote this originally...

packages/conf-openblas/conf-openblas.0.2.0/files/centos_install.sh
ryanrhymes: For CentOS, we don't need to build from source, takes way to long time. We can just install from binary. Check this Dockerfile https://github.com/owlbarn/owl/blob/6de0d5aafa85a179bbfcc8234a6cb77f50b2a617/docker/Dockerfile.centos#L15

ryanrhymes: So we can probably remove this sh file, but you may also need to install epel-release

packages/conf-openblas/conf-openblas.0.2.0/files/centos_install.sh
ryanrhymes: See my previous comment. I think this file is obsolete now, you can find all the dependency in [CentOS dockerfile](https://github.com/owlbarn/owl/blob/master/docker/Dockerfile.centos)

Similarly, all the dependency on various distributions can be found in [this folder](https://github.com/owlbarn/owl/tree/master/docker)

mseri: Nice work with those `Docker` files!! Thanks for pointing them out

packages/conf-openblas/conf-openblas.0.2.0/files/centos_install.sh
mseri: I think a depext on epel-release is not a problem, I would imagine lots of packages needing it anyway

packages/conf-openblas/conf-openblas.0.2.0/files/centos_install.sh
mseri: I think also the yum steps will need sudo

mseri:
As much as I would not like to leave the centos users alone, I think that it would be better if this package was doing nothing in centos and was instead producing a warning saying something on the line of: "OpenBLAS on centos needs to be installed manually, you can follow the procedure in [url/to/Dockerfile]" or even explicitly mentioning the necessary steps. Ping @kit-ty-kate or @avsm for an additional opinion

ryanrhymes:
why do we have to leave centos user behind? I think we only need to add openblas-devel this in opam for centos.

mor1:
it seems to have a failure when doing `opam install` with v1.2, but I don't understand the logging well enough to understand what is wrong, or how to replicate and fix... :/

kit-ty-kate:
The error on opam 1.2 is indeed really weird :/

@AltGr might be able to help ? opam pin seems to fail for some reason:
```
$ opam pin add -k version -yn conf-openblas 0.2.0
conf-openblas is now version-pinned to 0.2.0

conf-openblas is now unpinned from version 0.2.0
```

mseri:
I will test the `CentOS7` locally tomorrow as the CI is currently broken for CentOS

kit-ty-kate:
@mseri the CI doesn't seem to be broken for opam2 (I think it is for opam1 for an unknown reason):
```
#=== ERROR while compiling conf-openblas.0.2.0 ================================#
# context              2.0.0~rc | linux/x86_64 | ocaml-base-compiler.4.05.0 | pinned
# path                 ~/.opam/4.05.0/.opam-switch/build/conf-openblas.0.2.0
# command              /usr/bin/sh -exc cc $CFLAGS test.c -lopenblas
# exit-code            1
# env-file             ~/.opam/log/conf-openblas-17-9b8556.env
# output-file          ~/.opam/log/conf-openblas-17-9b8556.out
### output ###
# + cc test.c -lopenblas
# test.c:1:19: fatal error: cblas.h: No such file or directory
#  #include <cblas.h>
#                    ^
# compilation terminated.
```

mseri:
The CentOs CI for opam 1 is broken because the remote solver is not working, they do not provide a real aspcud and use some kind of remote service instead

kit-ty-kate:
@mseri yes, but the thing is that it worked before for months and suddenly stopped working even if no change was made to the docker image (last update: 5 months ago)

mor1:
Is there anything I can do to push this along? We're making a release of the `owl` numeric library and tools at the moment as well, which has a dependency on this updated package. Is there a way I can recreate the CI environment locally so that I can try to recreate and understand the error? Thanks!

kit-ty-kate:
> The joy of depending on an external service?

Not really. I mean here my guess is the image was updated on the docker hub but not locally on the server. I don't have access to this server nor the dockerfiles so I don't know.

I'm going to merge this since Travis (which has opam1) is doing fine. Sorry for the delay. Thanks !
If you can fix the opensuse build in an other PR when you have time it would be perfect.

mseri:
Sorry I have arrived late. For CentOS and OpenSuse you can grab aspcud rpms from https://github.com/xapi-project/xapi-travis-scripts/tree/master/aspcud 

This is what I was going to use for testing them

PR #11808, opened 2018-04-17
opam-fu: Update opam-lib dependency

bluddy:
This is a bottleneck on building the opam.org site, at least on 4.06.0 and above.


yomimono:
Thanks for this, @bluddy !

PR #11813, opened 2018-04-17
add mpp.0.3.2

pw374:


packages/mpp/mpp.0.3.2/findlib
```diff
@@ -0,0 +1 @@
+mpp
```
mseri: This file has no longer a meaning, you can safely remove it

pw374: OK, done. :)

packages/mpp/mpp.0.3.2/opam
```diff
@@ -0,0 +1,26 @@
+opam-version: "1.2"
+maintainer: "Philippe Wang <philippe.wang@gmail.com>"
+authors: [ "Philippe Wang <philippe.wang@gmail.com>" ]
+license: "ISC"
+homepage: "https://github.com/ocaml/MPP-language-blender"
+dev-repo: "https://github.com/ocaml/MPP-language-blender.git"
+bug-reports: "https://github.com/ocaml/MPP-language-blender/issues"
+build: [
```
mseri: Something is wrong here. The mpp tarball pointed by URL contains:
```
x mpp-0.3.2/
x mpp-0.3.2/mpp_libload.ml
x mpp-0.3.2/mpp_charset.ml
x mpp-0.3.2/mpp_dynload.ml
x mpp-0.3.2/Makefile
x mpp-0.3.2/META
x mpp-0.3.2/mpp_init.ml
x mpp-0.3.2/mpp_load.ml
x mpp-0.3.2/mpp_main.ml
x mpp-0.3.2/mpp_stringmap.ml
x mpp-0.3.2/mpp_core.ml
x mpp-0.3.2/mpp_charstream.ml
x mpp-0.3.2/mpp_out.ml
x mpp-0.3.2/mpp_actions.ml
```

That contains a makefile for building the code, does not require oasis at all in fact is making the Travis build fail with `missing _oasis`. The build, test and install steps here then are doing the wrong thing.

The `Makefile` is doing `ocamlfind install mpp META mpp.native` for the install, so uninstall is likely being similar.

This opam file seems more related to the code in https://github.com/ocaml/MPP-language-blender. I think the best way to deal with this is to tag the 0.3.2 version on the git repository and use the tarball associated to the tag instead.

pw374:
Let's see what happens now...

mseri:
It already looks much better. I think you need to drop support for ocaml < 4.02.0 as the String.init function used in moo was introduced in 4.02.0 (https://caml.inria.fr/pub/docs/manual-ocaml/libref/String.html)

It should be enough to add `available: ocaml-version >= "4.02”` to the opam file

pw374:
> It should be enough to add available: ocaml-version >= "4.02" to the opam file

Done :)

mseri:
I think it's fine now, but camlcity is not responding and making the CI fail. I'll re-trigger the builds once camlcity is back online

PR #11814, opened 2018-04-18
add mpp.0.3.2

pw374:
It doesn't appear obvious to me which one is better, this one, or https://github.com/ocaml/opam-repository/pull/11813


pw374:
OK, this one doesn't work at all...

PR #11817, opened 2018-04-18
Use a slightly less unstable tarball URL

### LINKED ISSUE #11557: scrypt package release url is not static
see https://github.com/ocaml/opam/issues/3251 which I erroneously opened on the `opam` issue tracker

Leonidas-from-XIV:
Closes #11557.

References the URL mentioned in https://github.com/ocaml/opam/issues/3251

### DIFF
--- a/packages/scrypt/scrypt.0.2.1/url
+++ b/packages/scrypt/scrypt.0.2.1/url
@@ -1,2 +1,2 @@
-archive: "https://github.com/pacemkr/ocaml-scrypt/tarball/v0.2.1"
+archive: "https://github.com/constfun/ocaml-scrypt/archive/v0.2.1.tar.gz"
 checksum: "24704b12638932a14400f74a354c0617"


Leonidas-from-XIV:
I have opted into adding a spurious dependency on `ctypes-foreign`, because the `META` file requires it (but nothing else as far as I can see). I can follow up with a patch that removes this dependency if desired.

yomimono:
[it looks like the checksum is not correct](https://travis-ci.org/ocaml/opam-repository/jobs/368048898#L889):

```
=-=- Gathering sources =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
[ERROR] Bad checksum for
        /home/travis/.opam/packages.dev/scrypt.0.2.1/v0.2.1.tar.gz:
          - 24704b12638932a14400f74a354c0617 [expected result]
          - 4e0b272f5438be475dc5ff5e851d9837 [actual result]
        This may be fixed by running `opam update`.
          - scrypt.0.2.1
        (This may be fixed by running 'opam update')
'opam install -t -v scrypt.0.2.1' failed
```

Leonidas-from-XIV:
Thanks @yomimono for looking into the failure. I've addressed it in a commit, here's my reasoning why the checksum probably changed, copied from the commit for reference:

> The old URL made GitHub generate a tarball with a folder containing <author>-<project>-<hash> and since the repository was moved from pacemkr to constfun account the current checksum is incorrect already.
> The new URL generates a tarball in the form <project>-<tagname> which is less likely to break and should contain the same content as the previous 0.2.1 release.

yomimono:
I think the MacOSX error in Travis can be fixed by replacing the `depext`s with a dependency on the opam package `conf-openssl`, which has `depext` entries for more systems than Debian and Ubuntu ([see here](https://github.com/ocaml/opam-repository/blob/master/packages/conf-openssl/conf-openssl.1/opam) .

Leonidas-from-XIV:
Thanks @yomimono, that's a good idea. I've always wondered what the advantage of these packages is, but this explains it very well 🙂

Is there a trick on how to read the build results? I find them very hard to understand and there are some failures where I can't do much about it (e.g. internal solver failures).

kit-ty-kate:
Don't worry about the failures on Alpine & CentOS, it's a long-standing issue that is going be fixed soon, hopefully…
Otherwise the build is still failing on Mac. Can you do something about it or do you think it's ok as is ?
```
#=== ERROR while installing scrypt.0.2.1 ======================================#
# opam-version 1.2.2
# os           darwin
# command      make
# path         /Users/travis/.opam/system/build/scrypt.0.2.1
# compiler     system (4.06.1)
# exit-code    2
# env-file     /Users/travis/.opam/system/build/scrypt.0.2.1/scrypt-10557-a00279.env
# stdout-file  /Users/travis/.opam/system/build/scrypt.0.2.1/scrypt-10557-a00279.out
# stderr-file  /Users/travis/.opam/system/build/scrypt.0.2.1/scrypt-10557-a00279.err
### stdout ###
# [...]
# checking sys/param.h presence... yes
# checking for sys/param.h... yes
# checking for posix_memalign... yes
# checking for special C compiler options needed for large files... no
# checking for _FILE_OFFSET_BITS value needed for large files... no
# configure: creating ./config.status
# config.status: creating Makefile
# config.status: creating config.h
# config.status: executing depfiles commands
# gcc -g -O2 -fPIC -lcrypto -DHAVE_CONFIG_H -Iscrypt-1.1.6 -Iscrypt-1.1.6/lib/crypto -Iscrypt-1.1.6/lib/scryptenc -Iscrypt-1.1.6/lib/util   -c -o scrypt-1.1.6/lib/crypto/crypto_aesctr.o scrypt-1.1.6/lib/crypto/crypto_aesctr.c
### stderr ###
# clang: warning: -lcrypto: 'linker' input unused
# scrypt-1.1.6/lib/crypto/crypto_aesctr.c:34:10: fatal error: 'openssl/aes.h' file not found
# #include <openssl/aes.h>
#          ^
# 1 error generated.
# make[1]: *** [scrypt-1.1.6/lib/crypto/crypto_aesctr.o] Error 1
# make: *** [libscrypt/libscrypt.a] Error 2
```

Leonidas-from-XIV:
@kit-ty-kate The problem is that Apple seems to have removed OpenSSL headers in the recent macOS, so you'd need to make sure to install openssl via brew and then do some trickery like [here](http://masnun.com/2015/12/01/fixing-fatal-error-opensslaes-h-file-not-found-on-os-x.html). This requires some more patching in the build process which in my opinion is better left for the upstream project. But that project is pretty much dormant.
 
Since I personally don't use the library and just wanted to fix up the most pressing issues reported on the bug tracker here I am also ok with adding a constraint that the package is just not installable on macOS.

yomimono:
>  I am also ok with adding a constraint that the package is just not installable on macOS.

That seems like a reasonable approach to me.

kit-ty-kate:
@Leonidas-from-XIV for opam1 the correct answer is `darwin` and for opam2 it's `macos`. See similar issue: https://github.com/ocaml/opam/issues/3293

PR #11820, opened 2018-04-18
Add an upper bound to deprecated ppx packages

xclerc:
Rationale: these package are now thin compatibility wrappers
around ppxlib, but we do not want ppxlib to be tied by retro
compatibility w.r.t. these packages.

@trefis


yomimono:
Is `<= 0.2.0` meant here?  I would have expected `< 0.2.0` for an upper bound.

yomimono:
Since `0.2.0` exists and is known to be compatible it makes sense, and I have no problem with it.  Just wanted to make sure that what was written was what was intended.  Thanks for the constraints!

xclerc:
Thanks!

PR #11823, opened 2018-04-18
conf-openblas: fix for opensuse

mor1:


packages/conf-openblas/conf-openblas.0.2.0/opam
```diff
@@ -6,17 +6,14 @@ bug-reports: "https://github.com/ocaml/opam-repository/issues"
...
+  ["sh" "-exc" "cc $CFLAGS test.c -lopenblas"] {os != "fedora" & os != "centos" & os != "opensuse" & os != "darwin"}
 ]
 depexts: [
-  [["fedora"] ["openblas-devel"]]
-  [["centos"] ["epel-release" "openblas-devel"]]
   [["alpine"] ["libc-dev" "openblas-dev"]]
-  [["ubuntu"] ["libopenblas-dev" "liblapacke-dev"]]
-  [["debian"] ["libopenblas-dev" "liblapacke-dev"]]
+  [["centos"] ["epel-release" "openblas-devel"]]
+  [["debian" "ubuntu"] ["libopenblas-dev" "liblapacke-dev"]]
```
kit-ty-kate: this is not the expected syntax here. The first element expects a list of flags that are going to be there on the user's machine. In this case the machine cannot be both `debian` and `ubuntu`.
These should be separated as they were before.

packages/conf-openblas/conf-openblas.0.2.0/opam
```diff
@@ -6,17 +6,14 @@ bug-reports: "https://github.com/ocaml/opam-repository/issues"
 dev-repo: "https://github.com/ocaml/opam-repository.git"
 license: "BSD"
 build: [
-  ["sh" "-exc" "cc $CFLAGS test.c -lopenblas"] {os = "alpine"}
-  ["sh" "-exc" "cc $CFLAGS test.c -lopenblas"] {os = "debian"}
```
kit-ty-kate: `debian` has been removed. This shouldn't be the case I suppose

mor1: This case should still be covered by the following line (`{os != "fedora" ...}`) I think?

packages/conf-openblas/conf-openblas.0.2.0/opam
mor1: Ah apologies, misinterpreted the format. (Couldn't find docs specifying it :(

kit-ty-kate: here is the documentation: https://opam.ocaml.org/doc/Manual.html

packages/conf-openblas/conf-openblas.0.2.0/opam
kit-ty-kate: Yes, you are right. Sorry.

kit-ty-kate:
Looks ok now.

kit-ty-kate:
Thanks.

mor1:
Ah thanks! Could've sworn I looked at that before, must've missed the
relevant bit...

kit-ty-kate:
@mor1 no worries. It happens. The CIs are here to check that.

PR #11830, opened 2018-04-19
Package oseq.0.1

c-cube:
### `oseq.0.1`

Simple list of suspensions, as a composable lazy iterator that behaves like a value.

Will be compatible with the new standard library's `Seq` module.

---
* Homepage: https://github.com/c-cube/oseq/
* Source repo: https://github.com/c-cube/oseq.git
* Bug tracker: https://github.com/c-cube/oseq/issues

---

### DIFF
--- a/packages/oseq/oseq.0.1/descr
+++ b/packages/oseq/oseq.0.1/descr
@@ -1,0 +1,3 @@
+Simple list of suspensions, as a composable lazy iterator that behaves like a value.
+
+Will be compatible with the new standard library's `Seq` module.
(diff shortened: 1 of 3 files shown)


hcarty:
Is this meant to depend on seq?

c-cube:
Yes, this depends on `seq` so it can reuse the same type :slightly_smiling_face:

hcarty:
Sorry - the question wasn't meant to be obtuse! It looks like the seq dependency is missing from the opam file.

c-cube:
Oh, sorry. I think opam-publish doesn't properly re-update the opam file by itself, I should have been more careful.

hcarty:
Thanks!

PR #11831, opened 2018-04-19
Release of stdcompat.2

thierry-martinez:
Support for OCaml 4.07.0


thierry-martinez:
Thanks! The previous versions do work with OCaml 4.07.0. They just don't include the new definitions introduced in 4.07.0.

PR #11834, opened 2018-04-19
Package cfstream.1.3.0

agarwal:
### `cfstream.1.3.0`

Stream operations in the style of Core's API.

---
* Homepage: https://github.com/biocaml/cfstream
* Source repo: https://github.com/biocaml/cfstream.git
* Bug tracker: https://github.com/biocaml/cfstream/issues

---
### opam-lint failures

---

### DIFF
--- a/packages/cfstream/cfstream.1.3.0/opam
+++ b/packages/cfstream/cfstream.1.3.0/opam
@@ -1,0 +1,24 @@
+opam-version: "1.2"
+homepage: "https://github.com/biocaml/cfstream"
+dev-repo: "https://github.com/biocaml/cfstream.git"
+bug-reports: "https://github.com/biocaml/cfstream/issues"
+license: "LGPL + linking exception"
+maintainer: "Ashish Agarwal <agarwal1975@gmail.com>"
+authors: [
+  "Philippe Veber <philippe.veber@gmail.com>"
+  "Ashish Agarwal <agarwal1975@gmail.com>"
+  "Drup <drupyog@zoho.com>"
+]
+
+build: [
+  ["jbuilder" "build" "-p" name "-j" jobs]
+]
+
+depends: [
+  "jbuilder" {build}
+  "core_kernel" {>= "v0.11.0"}
+]
+
+available: [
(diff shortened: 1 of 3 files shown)


packages/cfstream/cfstream.1.3.0/opam
```diff
@@ -0,0 +1,25 @@
...
+]
+
+build: [
+  ["jbuilder" "build" "-p" name "-j" jobs]
+]
+
+depends: [
+  "jbuilder" {build}
+  "core_kernel" {>= "v0.11.0"}
+  "ounit" {test}
```
kit-ty-kate: This dependency is unnecessary as there is no `build-test` phase.

packages/cfstream/cfstream.1.3.0/opam
```diff
@@ -0,0 +1,25 @@
...
+]
+
+available: [
+  ocaml-version >= "4.06.0"
```
kit-ty-kate: Is ther a reason you need this constraint ? I tried using 4.05.0 and it compiled fine.

packages/cfstream/cfstream.1.3.0/opam
agarwal: I deleted this.

packages/cfstream/cfstream.1.3.0/opam
agarwal: Thanks for noticing this. In fact, the package compiles fine with ocaml 4.04.1. I fixed this constraint.

kit-ty-kate:
Looks good, thanks ! :)

For info: `biocaml` seems to be incompatible with `>= core_kernel.v0.11.0`. Full log here: https://ci.ocamllabs.io/log/saved/docker-run-72604ab69cbdd34f7bf3f135082d5d38/954f849a9773bb5ea5db9e975e18d9d96d2c459a
This can be fixed in a later PR when you or anybody else have time.

agarwal:
@kit-ty-kate Thanks for the careful reviews! Yes, we are thinking about a new biocaml release too. Will investigate when we do that.

PR #11844, opened 2018-04-21
camlpdf.2.2.1: patch to add '-unsafe-string' flag when switch >= 4.06

ttamttam:
See https://github.com/johnwhitington/camlpdf/issues/22


pmetzger:
It might be better to fix camlpdf to not need mutable strings?

kit-ty-kate:
I agree with @pmetzger. In the meantime there is an error:
```

=-=- Processing actions -=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
[ERROR] Patch file
        "/home/opam/.opam/4.06.0/build/camlpdf.2.2.1/files/unsafe_string.patch"
        not found.

#=== ERROR while installing camlpdf.2.2.1 =====================================#
These patches didn't apply at /home/opam/.opam/4.06.0/build/camlpdf.2.2.1:
  - files/unsafe_string.patch
```

kit-ty-kate:
Ok, fixed it.

ttamttam:
Thanks Kate.

About the camlpdf fix: it is what I suggested upstream. I understood his (short-term?) suggestion to use the compilation flag after I had a look at the library: `camlpdf` is mainly a string processing library, and it defined a `bytes` type that is used a lot, and predates Pervasives's one.

kit-ty-kate:
Ok, fair enough. Thanks !

PR #11845, opened 2018-04-21
Package qtest.2.9

c-cube:
### `qtest.2.9`

qtest : Inline (Unit) Tests for OCaml.

qtest extracts inline unit tests written using a special
syntax in comments. Those tests are then run using the oUnit framework
and the qcheck library.  The possibilities range from trivial tests --
extremely simple to use -- to sophisticated random generation of test cases.

---
* Homepage: https://github.com/vincent-hugot/qtest
* Source repo: git@github.com:vincent-hugot/qtest.git
* Bug tracker: https://github.com/vincent-hugot/qtest/issues

---

### DIFF
--- a/packages/qtest/qtest.2.9/descr
+++ b/packages/qtest/qtest.2.9/descr
@@ -1,0 +1,6 @@
+qtest : Inline (Unit) Tests for OCaml.
+
+qtest extracts inline unit tests written using a special
+syntax in comments. Those tests are then run using the oUnit framework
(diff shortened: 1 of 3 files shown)


kit-ty-kate:
ping @camelus (maybe that would restart it ?)

kit-ty-kate:
Ok doesn't seem so. Whatever, seems good. Thanks, merging.
Do you still need `base-bytes` as a dependency ? (can be fixed in an other PR)

PR #11847, opened 2018-04-23
Package obelisk.0.3.1

Lelio-Brun:
### `obelisk.0.3.1`

Obelisk is a simple tool which produces pretty-printed output from a Menhir parser file (.mly).

---
* Homepage: https://github.com/Lelio-Brun/Obelisk
* Source repo: https://github.com/Lelio-Brun/Obelisk.git
* Bug tracker: https://github.com/Lelio-Brun/Obelisk/issues

---
### opam-lint failures

---

### DIFF
--- a/packages/obelisk/obelisk.0.3.1/descr
+++ b/packages/obelisk/obelisk.0.3.1/descr
@@ -1,0 +1,1 @@
+Obelisk is a simple tool which produces pretty-printed output from a Menhir parser file (.mly).
--- a/packages/obelisk/obelisk.0.3.1/opam
+++ b/packages/obelisk/obelisk.0.3.1/opam
@@ -1,0 +1,17 @@
+opam-version: "1.2"
+maintainer: "Lélio Brun <lelio.brun@inria.fr>"
+authors: "Lélio Brun <lelio.brun@inria.fr>"
+homepage: "https://github.com/Lelio-Brun/Obelisk"
+bug-reports: "https://github.com/Lelio-Brun/Obelisk/issues"
+license: "MIT"
+dev-repo: "https://github.com/Lelio-Brun/Obelisk.git"
+build: [make]
+install: [make "install" "BINDIR=%{bin}%"]
+build-test: [make "tests"]
+remove: [make "uninstall" "BINDIR=%{bin}%"]
+depends: [
+  "ocamlfind" {build}
+  "ocamlbuild" {build}
+  "menhir" {build}
+]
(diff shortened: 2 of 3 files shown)


Lelio-Brun:
My tests involve pdflatex, that is what Travis errors on. It was not the case before, should I remove mys tests ? 
For V1.2 Common Distros Alpine, m4 seems to be missing. For others, I am not sure of the errors.

kit-ty-kate:
For the tests there are two solutions:
* Either create a new `conf-pdflatex` package that would pull `pdflatex` from the system package manager
* Or remove the tests

I would rather prefer the first solution but it's up to you. Otherwise everything is good. The other failures are unrelated.

Lelio-Brun:
Thanks @kit-ty-kate, in [.travis-ci.sh](https://github.com/ocaml/opam-repository/blob/master/.travis-ci.sh) I see the following lines: 
```sh
    echo "====== External dependency handling ======"
    opam install depext
    depext=$(opam depext -ls $pkg --no-sources)
    opam depext $pkg
```
Would it be a third solution that I add external dependence in my opam file following https://opam.ocaml.org/doc/Manual.html#opamfield-depexts ? 

Also, how does it work for correcting this release? I use `opam-publish`, so https://opam.ocaml.org/doc/Packaging.html#With-opam-publish_1 mentions that I can update my pull-request but will the linked archive be updated also? Or should I close this PR and continue with another release?

kit-ty-kate:
Yes, this is what I was implying in the first solution (`conf-*` package are meta packages to centralize external dependencies).
In this case having external dependencies directly in your package is not really good because it is only needed during tests and there is no way to tell to install these depexts only for tests so it's going to install everytime for those who are using the `opam depext` command to install your package.
Also, having meta packages avoids having the same depexts repeated over and over for each packages that need it.

I don't use opam-publish so I don't know but if you know how to use git you can push new commit on top and this will update the PR.

Lelio-Brun:
All right I see. As I need some particular `texlive-*` packages, I don't think it's a good idea to build such a specialized `conf-*` package, and a `conf-texlive-full` package is probably overkill.
I might guard my tests with conditionals testing for the presence of LaTeX tools as LaTeX tests are relevant only if the end user is willing to use LaTeX. 

My question was more about OPAM and the setup with Github, because if I update my PR, I have to provide an updated `url` file linking to an updated archive, which in my case is automatically built by Github through the release process and I think cannot be updated with the same version name. So I think I have my answer: I will have to close this PR, issue a new release and open another PR.

PR #11855, opened 2018-04-24
+ott.0.28

hannesm:
- Bugfix: add missing case of Lem auxparam type name hack when it's hom (Brian Campbell)
- copy ocaml_light example sources from svn (rsem/ott/old_pre_github/examples/caml) and partially de-bitrot (Peter Sewell and palmskog for Coq)
- add pointer to VSCode plugin (JoeyEremondi)
- clean opam install of Emacs ott-mode (hannesm, Blaisorblade)


kit-ty-kate:
Thanks !

PR #11538, opened 2018-03-08
Add JsOfOCairo.1.1.0

jacquev6:
New version

PR #11857, opened 2018-04-25
Add JsOfOCairo.1.1.1 (and General.0.5.0)

jacquev6:
For reference this follows #11538.

### DIFF
--- a/.travis-ci.sh
+++ b/.travis-ci.sh
@@ -110,8 +110,9 @@ function build_one {
     echo
     echo "====== External dependency handling ======"
     opam install depext
-    depext=$(opam depext -ls $pkg --no-sources)
-    opam depext $pkg
+    # No --build-tests option on opam depext, workaround from https://github.com/ocaml/opam-depext/issues/10#issuecomment-93882764
+    depext=$(OPAMBUILDTEST=1 opam depext -ls $pkg --no-sources)
+    OPAMBUILDTEST=1 opam depext $pkg
     echo
     echo "====== Installing dependencies ======"
     opam install --deps-only $pkg
--- a/packages/General/General.0.5.0/descr
+++ b/packages/General/General.0.5.0/descr
@@ -1,0 +1,1 @@
+Rich functionality for built-in and basic OCaml types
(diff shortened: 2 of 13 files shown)


jacquev6:
The build [failed on Travis](https://travis-ci.org/ocaml/opam-repository/jobs/370964154#L1863) because, for JsOfOCairo's *tests*, I'm using a [Node.js package (node-canvas)](https://github.com/Automattic/node-canvas/#compiling) that requires the following APT packages:
    - libcairo2-dev
    - libjpeg8-dev
    - libpango1.0-dev
    - libgif-dev

I'll try and investigate how to make sure these are installed.

jacquev6:
I've created conf- packages for the missing system dependencies. I've based conf-pango on the existing conf-cairo, and conf-libjpeg and conf-libgif on the existing conf-libpng.

Please be careful when reviewing these packages: it's the first time I create such package, and I don't have a deep understanding of how they should be. Please let me know if I can do anything else.

hcarty:
This looks ok to me in general.  A lot of the CI complaints could be cleaned up with brew (macos) and alpine depext definitions.  I'm not sure that should prevent this from being merged though.

jacquev6:
So! There was more than just some missing homebrew depexts: the script used on Travis was not installing *test* external dependencies. I fixed this in a2d8644. @samoht, I see you had to do the same in ocaml-ci-scripts, so maybe you'll want to review this PR?

I got it working on Travis on my fork: https://travis-ci.org/jacquev6/opam-repository/jobs/371430628#L1045 so I'm quite confident it should work here as well.

jacquev6:
The [V1.2 Build test](https://ci.ocamllabs.io/ocaml/opam-repository/pr/11857?test=1%20V1.2%20Build) is still failing on my three new conf- packages but the V2.0 Build test is passing. The [logs](https://ci.ocamllabs.io/log/saved/docker-build-fcb649281057ab07a8bae6717c7796ac/d09a978ce1e8773a1f104dabe192308fd255233a) give very little information about the cause. Please let me know if I can do anything to help.

kit-ty-kate:
The error is really weird indeed. I think opam1 doesn't support pinning meta package or something like that. @AltGr could you confirm ?

In the meantime I think it's ok as is. Thanks a lot !

PR #11835, opened 2018-04-20
added packages/get_line/get_line.4.0.0

UnixJunkie:
the package went from obuild to jbuilder

PR #11862, opened 2018-04-25
Fix camlp4 system compiler for test dependencies

dra27:
#11700 only looked at the dependencies of the package being installed, but the package is then actually installed with tests so it can break if only test dependencies require camlp4 as in https://github.com/ocaml/opam-repository/pull/11835#issuecomment-383900346 (in this case, camlp4 is drawn in because of the buggy behaviour of `-t` in opam 1.2.2, but the same situation could still arise in future with opam 2.0.0)

This commit does fix #11835 - see [this build log](https://travis-ci.org/dra27/opam-repository/builds/371238156)


kit-ty-kate:
Thanks a lot !

PR #11868, opened 2018-04-26
added packages/obuild/obuild.0.1.10

UnixJunkie:
bugfix by @jeromemaloberti concerning new warning keyword in META files


UnixJunkie:
I think it is OK to merge

PR #11870, opened 2018-04-26
elpi 1.0

gares:
This PR packages ELPI, an OCaml library providing an extension language to a host OCaml application.

### DIFF
--- a/packages/elpi/elpi.1.0.2/descr
+++ b/packages/elpi/elpi.1.0.2/descr
@@ -1,0 +1,49 @@
+ELPI - Embeddable λProlog Interpreter
+
+ELPI implements a variant of λProlog enriched with Constraint Handling Rules,
+a programming language well suited to manipulate syntax trees with binders.
+
+ELPI is designed to be embedded into larger applications written in OCaml as
+an extension language. It comes with an API to drive the interpreter and 
+with an FFI for defining built-in predicates and data types, as well as
+quotations and similar goodies that are handy to adapt the language to the host
+application.
+
+This package provides both a command line interpreter (elpi) and a library to
+be linked in other applications (eg by passing -package elpi to ocamlfind).
+
+The ELPI programming language has the following features:
+
+- Native support for variable binding and substitution, via an Higher Order
+  Abstract Syntax (HOAS) embedding of the object language. The programmer needs
+  not to care about De Bruijn indexes.
+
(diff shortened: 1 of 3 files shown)


gares:
The `build-test` stanza requires `lua5.1` to be installed in order to run.

Which is the recommended way express that? There is already a `lua-config` but it pulls in only the header files, not the command line interpreter (and it pulls in lua5.2, while elpi uses 5.1)

gares:
The package builds fine here on OSX but I had to `brew install gnu-time`

kit-ty-kate:
Seems like you also depend on lua. Could you not invoke such command in your Makefile please ? (if I understand correctly what you said). There are meta package that are just here for that.
`conf-time` and `conf-lua` are what you want, though you might want to update them with the homebrew packages.

gares:
I've added `conf-time {test}` but I don't see it being installed.
Am I missing something?

kit-ty-kate:
the correct syntax was `{test}`. Also you missed `conf-lua` I guess ?
Also these won't work on OSX until you add the correct depext to `conf-lua` and `conf-time`.

gares:
Sorry I did not understand I had to patch conf-time. But looking at its description
```
    new conf-time package for the time command
    
    This package does *not* enforce availability of "GNU time", it may be
    BSD time (BSD, OSX) or Busybox's (Alpine). The only common
    functionality is "time <command>".
```
Do you want me to add a new conf-gnu-time package?

gares:
> Also you missed conf-lua I guess ?

I double checked and (despite an error being printed) `lua5.1` is not necessary (it is only to generate a performance graph, not to run the tests).

On the contrary (a decent version of) `time` is needed.

kit-ty-kate:
mmh I see. So if you are willing to do that you can remove the `build-test` field and no tests will be build.
If you prefer to keep it but really need a GNU time, maybe you can create a new `conf-gnu-time` but I don't know if it's worth it.

gares:
I've disabled tests, but maybe I should have asked first what is the policy for merging a PR here.
If it needs all green lights, then I'm fine with what I did.
If it does not require all CI to get green, I prefer to have the tests be run on most of the cases and have some platform fail for silly reasons.

gares:
Can you help me understand the other failures? I quite don't get what's wrong from the logs

kit-ty-kate:
Don't worry with that. Internal long-standing issues. Thanks !!

PR #11871, opened 2018-04-26
fix bun dependency on spawn by adding v prefix

yomimono:
It's `v0.10.1`, not `0.10.1`, in `packages/spawn`.  Refer to it properly to avoid installing `v0.9.0` and getting link errors on pthread.


mseri:
Looks like the failures are the usual suspects

kit-ty-kate:
Thanks a lot !

PR #11874, opened 2018-04-27
added packages/lbvs_consent/lbvs_consent.1.1.3

UnixJunkie:
fix so that it compiles with the current version of the cpm library

### DIFF
--- a/packages/lbvs_consent/lbvs_consent.1.1.3/descr
+++ b/packages/lbvs_consent/lbvs_consent.1.1.3/descr
@@ -1,0 +1,8 @@
+Chemoinformatics software for Ligand-Based Virtual Screening (LBVS) using consensus queries.
+
+Similarity search for molecules using several known active molecules
+(for a protein of interest) instead of a single query search.
+Details can be found in the paper:
+"Consensus queries in ligand-based virtual screening experiments".
+F. Berenger, O. Vu and J., Meiler.
+Journal of Cheminformatics, November 2017.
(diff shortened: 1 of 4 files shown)


kit-ty-kate:
The only failure is on OSX. Could you explain the reason behind having:
```
["sh" "-c" "brew tap rdkit/rdkit && brew install rdkit"] {os = "darwin"}
```
in the build phase ? It's a bit unrelated to this but I find it a bit dubious :/

UnixJunkie:
The command you are mentioning comes from the conf-rdkit package.
This brew recipe is very often broken and users have trouble installing it.
However, last time I checked, it was the official way to install rdkit on mac os x.
You can complain here https://github.com/rdkit/homebrew-rdkit/issues if it repeatedly does not work.

UnixJunkie:
I think you are asking for changes into conf-rdkit's opam file.
I don't think it is a good idea: if the brew formula for rdkit was working, then conf-rdkit would install automatically on a mac.
Of course, if in the depexts field there was a way to specify a brew tap, then it could go into the depexts
but I don't think there is such a mechanism currently.

kit-ty-kate:
I think the problem is for users who want to use another version of rdkit than the one available through this method. They wouldn't use `opam depext` and thus wouldn't expect to have the build phase install something on itself (+ it doesn't seem to work on Travis anyway).

UnixJunkie:
cf. https://github.com/ocaml/opam-repository/pull/11931

kit-ty-kate:
Looks good, thanks !

PR #11801, opened 2018-04-16
+flowtype.0.70.0

avikchaudhuri:

PR #11875, opened 2018-04-27
+flowtype.0.71.0

fishythefish:


kit-ty-kate:
Looks good but I find a bit dubious to have a dependency to `ocamlbuild` but not `ocamlfind` here :/ Is it really not needed anyway. I could find calls to `ocamlfind` in the source code.

kit-ty-kate:
It's failing on Alpine with the following error:
```
# In file included from hh_shared.c:98:0:
# /usr/include/sys/errno.h:1:2: warning: #warning redirecting incorrect #include <sys/errno.h> to <errno.h> [-Wcpp]
#  #warning redirecting incorrect #include <sys/errno.h> to <errno.h>
#   ^~~~~~~
# hh_shared.c:182:26: fatal error: asm/unistd.h: No such file or directory
#    #include <asm/unistd.h>
#                           ^
# compilation terminated.
# make: *** [Makefile:283: _build/hack/heap/hh_shared.o] Error 2
```
Could you do something about it ? Otherwise if there is no other failures I suppose it would be \~mostly\~ ok as is.

kit-ty-kate:
Closing this. Same reason as in #11801

Issue #11852, opened 2018-04-23
Build dependency on `ppx_sexp_conv` should be runtime dependency?


paurkedal:
The `ipaddr` package installs a `META` file with a dependency on `ppx_sexp_conv.runtime-lib`, but `ppx_sexp_conv` is only declared a build dependency in opam.  This can cause dependent packages to fail building when the `ipaddr` itself is not rebuilt, due since opam does not know that `ppx_sexp_conv` must be rebuilt first.  An example is show below.

Note that `ipaddr` has a runtime dependency on `sexplib`, so the `ppx_sexp_conv.runtime-lib` may have been an unexpected dependency due to the `requires(-ppx_driver) = "ppx_sexp_conv.runtime-lib"` line in the `META` of `ppx_sexp_conv`.
```
#=== ERROR while compiling mirage-stack-lwt.1.2.0 =============================#
# context      2.0.0~rc | linux/x86_64 | base-bigarray.base base-threads.base base-unix.base ocaml-base-compiler.4.06.1 | https://opam.ocaml.org/2.0#6927894a
# path         ~/.opam/4.06.1/.opam-switch/build/mirage-stack-lwt.1.2.0
# command      ~/.opam/4.06.1/bin/jbuilder build -p mirage-stack-lwt -j 6
# exit-code    1
# env-file     ~/.opam/log/mirage-stack-lwt-23800-575881.env
# output-file  ~/.opam/log/mirage-stack-lwt-23800-575881.out
### output ###
# File "/home/urkedal/.opam/4.06.1/lib/ipaddr/META", line 1, characters 0-0:
# Error: Library "ppx_sexp_conv.runtime-lib" not found.
# -> required by library "ipaddr" in /home/urkedal/.opam/4.06.1/lib/ipaddr
# Hint: try: jbuilder external-lib-deps --missing -p mirage-stack-lwt @install


#=== ERROR while compiling mirage-runtime.3.0.7 ===============================#
# context      2.0.0~rc | linux/x86_64 | base-bigarray.base base-threads.base base-unix.base ocaml-base-compiler.4.06.1 | https://opam.ocaml.org/2.0#6927894a
# path         ~/.opam/4.06.1/.opam-switch/build/mirage-runtime.3.0.7
# command      ~/.opam/4.06.1/bin/jbuilder build -p mirage-runtime -j 6
# exit-code    1
# env-file     ~/.opam/log/mirage-runtime-23800-406bbc.env
[...]

PR #11880, opened 2018-04-28
otr, tls, x509: upper bound for sexplib

hannesm:
see https://github.com/ocaml/opam-repository/issues/11852

### DIFF
--- a/packages/otr/otr.0.3.1/opam
+++ b/packages/otr/otr.0.3.1/opam
@@ -24,7 +24,7 @@ depends: [
   "ppx_cstruct" {build}
   "sexplib"
   "ppx_deriving" {build}
-  "ppx_sexp_conv" {build}
+  "ppx_sexp_conv" {build & < "v0.11.0"}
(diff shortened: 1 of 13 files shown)


packages/otr/otr.0.3.1/opam
```diff
@@ -22,7 +22,7 @@ depends: [
   "ppx_tools" {build}
   "cstruct" {>= "1.9.0"}
   "ppx_cstruct" {build}
-  "sexplib"
+  "sexplib" {< "v0.11.0"}
```
kit-ty-kate: Isn't it more of an issue with `ppx_sexp_conv` instead of `sexplib` directly ?

hannesm:
@kit-ty-kate yes, thanks.  I force-pushed a commit which changed this (from sexplib to ppx_sexp_conv)

kit-ty-kate:
There are some failures in the revdeps but they are unrelated. Thanks a lot !! :)

PR #11881, opened 2018-04-29
logs-syslog 0.1.0 is not compatible with lwt 4.0.0, 0.1.0 and 0.1.1 r…

hannesm:
…equire newish x509 and tls

as observed in #11880


kit-ty-kate:
Thanks once again !! :)

PR #11891, opened 2018-05-01
Package sanddb.0.1

StrykerKKD:
### `sanddb.0.1`

A simple immutable database for the masses

SandDB is a simple immutable database, which is also:
 - Simple: It only does one thing, which is persisting data in a file.
 - Easy to use: SandDB's API is extremely small, so you only need to know few functions to use it.
 - Type safe: Every common dangerous operation (like parsing) is covered by the Result type, so you will know where to expect errors.
 - Immutable: Database is based on the immutable stack idea, where you can only push onto the stack.
 - Crud capable: Even though the database is immutable you still can update and delete records, by shadowing them.
 - Version keeping: Every update and delete operation will produce a new version of the affected record, without modifying the original, so you will have all versions of your data.
 - Concurrent: SandDB is based on lwt, so every database operation is asynchronous.
 - Supports multiple serializers: SandDB supports both json and biniou serialization format thanks to the atdgen library.

---
* Homepage: https://github.com/StrykerKKD/SandDB
* Source repo: http://github.com/StrykerKKD/SandDB.git
* Bug tracker: https://github.com/StrykerKKD/SandDB/issues

---

### DIFF
--- a/packages/sanddb/sanddb.0.1/descr
+++ b/packages/sanddb/sanddb.0.1/descr
@@ -1,0 +1,11 @@
+A simple immutable database for the masses
+
+SandDB is a simple immutable database, which is also:
+ - Simple: It only does one thing, which is persisting data in a file.
+ - Easy to use: SandDB's API is extremely small, so you only need to know few functions to use it.
+ - Type safe: Every common dangerous operation (like parsing) is covered by the Result type, so you will know where to expect errors.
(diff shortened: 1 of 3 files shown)


mseri:
Looks good, thanks! Feel free to announce the release on the discuss.ocaml.org forum. We have a community category and an announcement tag for this purpose

PR #11897, opened 2018-05-03
ocamlmod 0.0.3 0.0.4 0.0.7: mark as unavailable with ocaml 4.06.0+

hannesm:
0.0.8 includes a custom patch for `-safe-string`, 0.0.9 is working out of the box with `-safe-string`.

discovered via https://github.com/mirage/mirage-skeleton/issues/256#issuecomment-386223775


kit-ty-kate:
Thanks !

PR #11898, opened 2018-05-03
avoid ppx_sexp_conv v0.11.0 since it comes with a runtime depedency on base

hannesm:
see https://github.com/ocaml/opam-repository/issues/11852 and the latest mirage call for more reasoning behind this upper bound

### DIFF
--- a/packages/charrua-core/charrua-core.0.3/opam
+++ b/packages/charrua-core/charrua-core.0.3/opam
@@ -15,7 +15,7 @@ depends: [
   "ocamlfind" {build}
   "ppx_deriving" {build}
   "ppx_cstruct" {build}
-  "ppx_sexp_conv" {build}
+  "ppx_sexp_conv" {build & < "v0.11.0"}
   "ppx_type_conv" {build}
   "cstruct" {>="1.9.0" & <"3.0.0"}
   "cstruct-unix"
--- a/packages/charrua-core/charrua-core.0.4/opam
+++ b/packages/charrua-core/charrua-core.0.4/opam
@@ -14,7 +14,7 @@ depends: [
   "ocamlfind" {build}
   "ppx_deriving" {build}
   "ppx_cstruct" {build}
-  "ppx_sexp_conv" {build}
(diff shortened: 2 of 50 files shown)


avsm:
Hm, this lgtm from a mirage perspective, but it does make two packages uninstallable, probably due to bad transitive dependencies somewhere.

diml:
BTW, ppx_sexp_conv v0.11.1 won't depend on Base, so you should add a conflict with v0.11.0 only.

avsm:
thanks for the quick turnaround @diml! I think this constraint works as it's a <0.11.0; what's the eta on the 0.11.1 release?  We could test both together

diml:
Actually, I can do it now. However, I don't see how the <v0.11.0 constraint works as it will prevent ppx_sexp_conv.v0.11.1 to be installed together with these packages, which is going to be a pain for users.

diml:
Just did the release: https://github.com/ocaml/opam-repository/pull/11901

diml:
The runtime dependency can be queried from the META file of ppx_sexp_conv. For instance the following command gives the runtime dependency of ppx_sexp_conv:

```
$ grep 'requires(-ppx_driver)' `ocamlfind query ppx_sexp_conv`/META| cut -d\" -f2
ppx_sexp_conv.runtime-lib
```

It should work with both ppx_sexp_conv < v0.11.0 and >= v0.11.0. This could be used to generate the META file.

BTW dune handles all this already, so this is only needed for packages that don't use dune. If there are still technical limitations that prevent these packages from switching to dune, we can look into what can be done.

samoht:
@hannesm could you rebase this if this is still needed? Thanks

samoht:
ping @hannesm 

As `ppx_sexp_conv` added a runtime dependency, I think this is still needed.

hannesm:
sorry, short on time to fix the merge conflicts..  I'd guess the touched packages will still need to declare a run-time dependency for ppx_sexp_conv.runtime_lib, and may not work with the current opam-repository.

PR #11901, opened 2018-05-03
ppx_sexp_conv v0.11.1

diml:
Remove the dependency from `ppx_sexp_conv.runtime-lib` to base

### DIFF
--- a/packages/msgpck/msgpck.1.0/opam
+++ b/packages/msgpck/msgpck.1.0/opam
@@ -12,7 +12,7 @@ depends: [
   "ocamlbuild" {build}
   "topkg" {build}
   "ppx_deriving" {build}
-  "ppx_sexp_conv" {build}
+  "ppx_sexp_conv" {build & < "v0.11.0"}
   "sexplib"
   "ocplib-endian"
 ]
--- a/packages/ocaml-topexpect/ocaml-topexpect.0.1/opam
+++ b/packages/ocaml-topexpect/ocaml-topexpect.0.1/opam
@@ -8,7 +8,7 @@ dev-repo: "https://github.com/let-def/topexpect.git"
 build: [make]
 depends: [
   "ocamlfind" {build}
-  "ppx_sexp_conv"
+  "ppx_sexp_conv" {< "v0.11.0"}
   "ppx_deriving"
 ]
 available: [ocaml-version >= "4.04.0"]
--- a/packages/ocaml-topexpect/ocaml-topexpect.0.1/opam
+++ b/packages/ocaml-topexpect/ocaml-topexpect.0.1/opam
@@ -11,4 +11,4 @@ depends: [
   "ppx_sexp_conv" {< "v0.11.0"}
   "ppx_deriving"
 ]
-available: [ocaml-version >= "4.04.0"]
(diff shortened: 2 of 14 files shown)


hannesm:
thanks.  are the revdeps failures expected?  I suggest we merge this, and afterwards a PR similar to #11898 which (a) conflicts with `ppx_sexp_conv {= "v0.11.0"}` and (b) turns the build dependency `ppx_sexp_conv {build}` into a runtime dependency `ppx_sexp_conv`.  if I follow https://github.com/mirleft/ocaml-nocrypto/pull/144 along, there is as well the requirement to expose this runtime dependency via META!?  since when does `ppx_sexp_conv` provide a `runtime-lib` (i.e. do we need as well a lower bound of `ppx_sexp_conv` in the respective packages?

diml:
Looking at the errors, some of them should already occur with ppx_sexp_conv v0.11.0, I'll add some constraints. Some are just missing a dependency on sexplib.

Basically, the runtime dependency of `ppx_sexp_conv` used to be `sexplib`, but there was a change between v0.10 and v0.11 to add a `ppx_sexp_conv.runtime-lib` library and make it the runtime dependency of `ppx_sexp_conv`. This runtime library simply re-exports some functions under a `Ppx_sexp_conv_lib` toplevel module and the generated code refers to it. This was part of a change that affected all our ppx rewriters and was motivated by:
- keeping the generated code as small as possible and move more code to runtime libraries, since it's much easier to review and maintain
- hygiene: we had some name clash problems and by systematically having a runtime library we have a better guarantee on what the identifiers produced by ppx rewriters refer too

diml:
For `shared-block-ring` and `vchan-unix` the issue seems to be unrelated to this PR so I didn't touch them.

@avsm and @let-def, after this PR, `ocaml-topexpect` and `yaml` needs to be updated to add an explicit dependency on sexplib. For `ocaml-toplexpect`, the `sexplib` dependency was already in the jbuild file so I just added it to the opam file, there should be no need for another release, but for yaml, I think we'll need a new release.

diml:
Not sure what's going on with the CI here, I constrained things more so there should be less failures... The new errors look unrelated to this PR

diml:
Ok, it looks better now. I tested manually and ocaml-topexpect 0.1, 0.2 and 0.3 don't build with OCaml >= 4.05, so I added more constraints. The last two failures - qcow 0.6 and 0.7 - are unrelated.

I think this PR is good to merge.

diml:
Actually I added the missing dependency on `io-page-unix` for qcow 0.6 and 0.7 and the result is a bit better now. The last failures are really unrelated to this PR

trefis:
Ping, it would be nice to unsplit the world ASAP.

(x-ref https://github.com/mirage/ocaml-uri/pull/120 )

trefis:
Forget about the "unsplit the world part", apparently my opam has just gone crazy on its own.
Still, it would be nice to get this released.

hannesm:
sorry for the late reply.  This is good to go.  to unsplit the universe we'll still need to carefully adapt packages to the new ppx_sexp_conv world (thanks @diml for explaining the design considerations).

PR #11906, opened 2018-05-03
Add some incorrect jbuilder minimum bounds

samoht:
/cc @diml


samoht:
These packages are using the `package` field which was not available before jbuilder 1.0+beta9.

samoht:
I am going to split that PR into to to avoid the explosion of revdeps.

PR #11907, opened 2018-05-03
Fix depext for old versions of secp256

samoht:


samoht:
The missing depext on ubuntu cannot be fixed easily, so merging

PR #11908, opened 2018-05-03
Fix missing depext for old versions of capnp-rpc

samoht:
/cc @talex5


samoht:
The missing depexts on alpine cannot be fixed easily, so merging this one.

PR #11915, opened 2018-05-03
Package fstar.0.9.6.0~alpha1

tahina-pro:
Upcoming new release of the F* programming language with dependent types and effects.
In preparation for several courses and tutorials, we would like to make an opam package already available to our students for the pre-release, 0.9.6.0~alpha1.

0.9.6.0~alpha1 brings at least the following improvements in:
* two-phase type checking (towards disentangling type checking from checking of verification conditions), more robust wrt. inference of implicit arguments
* extraction to OCaml preserves the structure of `let`-bindings
* extraction now erases all pure unit functions, lemmas and ghost functions
* etc.

See https://github.com/FStarLang/FStar/blob/v0.9.6.0-alpha1/CHANGES.md for more details.

(Also, this new version of the opam package works with `opam pin --dev-repo`, while the one for 0.9.5.0 no longer does.)


catalin-hritcu:
Would very much appreciate if someone could merge this quickly so that we can send instructions to our students for a summer school next week. @samoht @kit-ty-kate helped us with such things in the past :)

kit-ty-kate:
Looks about right. CIs are all green. However there is just one thing: do you use the `Findlib` library from the `ocamlfind` package ? I couldn't find any mentions of it in the sources.
If not, it would be great to have `ocamlfind` flag as a build dependency (as is `ocamlbuild`)

In the meantime I can merge this. So if you can do this change for all versions if necessary in another PR it would be wonderful.

Thanks a lot !

catalin-hritcu:
@kit-ty-kate Thanks for the quick merge! Do you have an idea how long it usually takes until a normal `opam update && opam install` will pick up the new version?

I hope that @tahina-pro can look at the Findlib issues soon, but he's in Redmond and it's still night there.

kit-ty-kate:
hm, usually one or two hours, I don't remember.

PR #11916, opened 2018-05-04
ocaml-migrate-parsetree.1.0.8 does not work with 4.07.0

damiendoligez:
Specifically, when you install ocaml-migrate-parsetree.1.0.8 on ocaml.4.07.0+beta2, then ppx_tools_versioned.5.1 produces preprocessors that output the wrong magic number, which makes lwt.3.1 (and a number of other packages) fail to compile.


kit-ty-kate:
Thanks a lot !

PR #11917, opened 2018-05-04
Package bamboo.0.0.03

pirapira:
### `bamboo.0.0.03`

A compiler targeting Ethereum Virtual Machine

Bamboo compiles a simple language to Ethereum Virtual Machine. The
language is designed to keep programmers away from common mistakes. It
features: state transition as recursion with potentially changing
arguments, mandatory reentrance continuation when calling out, no
loops, no assignments except to mappings and partial compliance with
common Ethereum ABI.

---
* Homepage: https://github.com/pirapira/bamboo
* Source repo: https://github.com/pirapira/bamboo.git
* Bug tracker: https://github.com/pirapira/bamboo/issues

---

### DIFF
--- a/packages/bamboo/bamboo.0.0.03/descr
+++ b/packages/bamboo/bamboo.0.0.03/descr
@@ -1,0 +1,8 @@
+A compiler targeting Ethereum Virtual Machine
+
+Bamboo compiles a simple language to Ethereum Virtual Machine. The
+language is designed to keep programmers away from common mistakes. It
+features: state transition as recursion with potentially changing
+arguments, mandatory reentrance continuation when calling out, no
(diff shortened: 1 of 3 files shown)


pirapira:
The build failure happens in a dependency `cryptokit.1.12` and the issue seems to be https://github.com/ocaml/ocamlbuild/blob/4c7dae3e5b4e369e4c961e3b012823cd4f45c31e/src/ocaml_specific.ml#L555
>  ocamlc -c -o dir/foo.o dir/foo.c
       does not work as expected.

https://caml.inria.fr/mantis/view.php?id=6475

I'll try specifying an `ocamlbuild` version with this workaround.

PR #11919, opened 2018-05-04
add bound on ocaml version

damiendoligez:
`ocamlnet` embeds an old version of cppo that doesn't work on 4.07 because the parsing of `# line` directives has changed. Patch sent upstream: https://gitlab.camlcity.org/gerd/lib-ocamlnet3/merge_requests/4/diffs

cc @gerdstolpmann


gerdstolpmann:
Thanks, I'll fix during the weekend

kit-ty-kate:
Thanks !

PR #11920, opened 2018-05-04
extlib: add bound on ocaml version

damiendoligez:
`extlib` does not build with OCaml 4.07 because the interface of `Hashtbl` has changed. Search for `@since 4.07` in `hashtbl.mli` to see what needs to be added.
[update: `extlib-compat` fails for the same reason]

cc @ygrek


kit-ty-kate:
`extlib.1.7.2` tests fails with 4.06 (safe-string) but let's merge it anyway.

Thanks a lot !

PR #11922, opened 2018-05-05
base-bytes.legacy should not be an upgrade to base-bytes.base

dra27:
Versionographically speaking, "legacy" is newer than "base" which means that OPAM considers `base-bytes.legacy` as an upgrade to `base-bytes.base` for OCaml 4.02+ and only doesn't do it because of the version constraint.

It results in a slightly poor UI (reported in https://github.com/ocaml/opam/issues/3331):

```
Everything as up-to-date as possible.
The following newer versions couldn't be installed:
  - base-bytes.legacy is not available because your system doesn't comply with ocaml-version < "4.02.0".
```

Given that nothing should depend specifically on the legacy version, can I suggest that it's renamed to `base-bytes.backport`. This means that OCaml 4.02+ now doesn't get a somewhat illogical message about being too new and OCaml 4.01 and earlier get a much more meaningful:

```
dra@bionic:~$ opam upgrade --verbose
Everything as up-to-date as possible.
The following newer versions couldn't be installed:
  - base-bytes.base is not available because your system doesn't comply with ocaml-version >= "4.02.0".
```


mseri:
It seems that no other package requirest the legacy version explicitly, so this shoul not break anything. I think this is a good idea but I'd like to wait for more feedback before merging

mseri:
Ping @kit-ty-kate if you also agree I think we could merge

hannesm:
I agree with @mseri and approve this by merging :)

PR #11925, opened 2018-05-06
extlib: one more 32-bit arch

ygrek:
follow up to #11671


kit-ty-kate:
Is this specific to `extlib.1.7.4` ? Otherwise I think it would be wise to do it for every versions if you have some spare time. If you don't have time to do this I (or someone-else) will do it sometime this week.

kit-ty-kate:
Ok good. Thanks a lot ! :)

PR #11928, opened 2018-05-06
Fix ocaml and ocaml-system packages to take .exe into account

dra27:
This is a manual application of the change to the upgrade process in https://github.com/ocaml/opam/pull/3337 intended to unblock https://github.com/ocaml/opam/issues/3336.

I don't know if this affects Camelus's rewriting as well - I think not, as the ocaml packages are maintained by hand but could opam-repository maintainers wait for @AltGr to review before merging this one.

### DIFF
--- a/packages/ocaml-system/ocaml-system.3.07/files/gen_ocaml_config.ml.in
+++ b/packages/ocaml-system/ocaml-system.3.07/files/gen_ocaml_config.ml.in
@@ -1,5 +1,13 @@
 let () =
-  let ocamlc = Sys.executable_name ^ "c" in
+  let exe = ".exe" in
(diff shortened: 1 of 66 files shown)


AltGr:
If this is in line with opam's rewrite code, it's OK to merge. We just should recompile and restart @camelus around the same time, so that it doesn't detect conflicts on changes to the wrapper compiler packages (which shouldn't happen often, and in the worst case, will make it file a PR that we'll have to ignore, so nothing critical)

In other words: good to merge.

samoht:
@AltGr feel free to merge whenever you want. This is currently killing our CI as as basically recompile all the existing packages ...

PR #11930, opened 2018-05-07
Package fstar.0.9.6.0~alpha1 (tweaks)

tahina-pro:
- Flag ocamlfind as build dependency
- Make `opam remove fstar` actually remove `fstar.exe` (which is the actual name of the F* binary, not `fstar`)


kit-ty-kate:
Looks good ! :) Can you consider doing this for previous versions as well if you have some spare time ?

kit-ty-kate:
In the meantime, let's merge this.

PR #11931, opened 2018-05-07
maybe better opam file

UnixJunkie:


kit-ty-kate:
Thanks for that !!
I'm not totally sure it would work but what about adding `& os = "darwin"` to the post-message ?

UnixJunkie:
I think I tried but it did not parse (I use opam-2.0).

kit-ty-kate:
Let's try that. If it doesn't work I'll revert it.

kit-ty-kate:
Looks like it's working. Are you ok with the new message ?

UnixJunkie:
I updated the post-message, thanks.

kit-ty-kate:
Oops, sorry. Thanks a lot !!

PR #11940, opened 2018-05-08
Regenerate 0.1

Drup:
Regenerate is a tool to generate test-cases for regular expression engines.

Regenerate takes a regular expression and generates strings that match it.
It handles most posix extended regular expressions along with
complement (~a) and intersection (a&b).
Since it handles complement, it can also generate strings that
*don't* match a given regular expression.

- repo: https://github.com/Drup/regenerate
- website: https://drup.github.io/regenerate/
- release: https://github.com/Drup/regenerate/releases/tag/0.1

----

This marked an attempt at releasing with `dune-release`. It was, honestly, mostly unsuccesful (the good point is that I found many issues). I hope the packaging is still correct! :)

### DIFF
--- a/packages/regenerate/regenerate.0.1/descr
+++ b/packages/regenerate/regenerate.0.1/descr
@@ -1,0 +1,7 @@
+Regenerate is a tool to generate test-cases for regular expression engines.
+
+Regenerate takes a regular expression and generates strings that match it.
+It handles most posix extended regular expressions along with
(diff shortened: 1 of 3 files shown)


Drup:
[Huhu, well, yes, regenerate does require quite a bit of memory](https://travis-ci.org/ocaml/opam-repository/jobs/376493761#L1273) :D 

That's the only failure afaik. So I guess it's good to merge. :)

PR #11941, opened 2018-05-09
added conf-dbm

UnixJunkie:
inspired by the configure file from dbm

### DIFF
--- a/packages/conf-dbm/conf-dbm.1.0.0/descr
+++ b/packages/conf-dbm/conf-dbm.1.0.0/descr
@@ -1,0 +1,3 @@
+Virtual package relying on gdbm
+
+This package can only install if the gdbm library is installed on the system.
--- a/packages/conf-dbm/conf-dbm.1.0.0/files/test.c
+++ b/packages/conf-dbm/conf-dbm.1.0.0/files/test.c
@@ -1,6 +1,7 @@
-#include <ndbm.h>
+#include <gdbm.h>
 
 int main() {
-  (void) dbm_open ("foo", 0, 0);
+  GDBM_FILE db = gdbm_open("conf-dbm-testdb", 512, GDBM_WRCREAT, 0600, 0);
+  gdbm_close (db);
   return 0;
 }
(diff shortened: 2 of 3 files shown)


kit-ty-kate:
Well now that we have https://github.com/ocaml/dbm/pull/5 merged I would avoid requesting for the compat version of gdbm. The difference is really messy and quite impossible to deal with correctly it seems.

UnixJunkie:
What is the C header file you want to rely on then?
You might know better than me what this conf- package should be, please feel free to edit it.

kit-ty-kate:
`libgdbm-dev` should be sufficient here as on debian <= 9 it is the equivalent of debian >= 10's `libgdbm-compat-dev`. `dbm.1.2` now support both however previous versions does not support debian >= 10's `libgdbm-dev` but I think that's ok

UnixJunkie:
ok, thanks for the edits

kit-ty-kate:
Ok, the failure here comes from the fact that `ndbm.h` is not installed in the debian package.
Instead there is `dbm.h` (not to be used here I think), `gdbm-ndbm.h` and `gdbm.h`
See: https://packages.debian.org/stretch/amd64/libgdbm-dev/filelist

UnixJunkie:
I knew it:
````
$ locate ndbm.h
/usr/include/gdbm-ndbm.h
/usr/include/ndbm.h
$ dpkg -S /usr/include/ndbm.h
libgdbm-compat-dev: /usr/include/ndbm.h
````
(that's on a debian buster/sid)

UnixJunkie:
So, how do we go from here? I add back the libgdbm-compat-dev that was in there for debian and Ubuntu?

kit-ty-kate:
Well it depends. Do you want to test for \*a\* dbm library or specifically ndbm ?

UnixJunkie:
I changed the test file to use gdbm, finally. I guess that's what most people might use.
Should we run the produced exe, by the way?
I think that most opam conf-* packages are happy if they can compile and link against the desired library.

UnixJunkie:
All CI tests are OK, what a rare event!
Sorry for the rather piggy build section in the opam file.
If someone has a proposal to do this in a nicer way, please feel free to contribute it.

kit-ty-kate:
Awesome ! Thanks a lot ! Sorry for the delay.

PR #11945, opened 2018-05-09
Add version 1.3 of the OByteLib library.

bvaugon:
This new version of OByteLib simply fix ocamlfind compatibility and expose some useful functions.


kit-ty-kate:
Seems like you only use ocamlbuild as a build tool and not as a library so it seems fair to flag it as a build dependency.
If for some reason I missed something please correct me by removing my commit.

kit-ty-kate:
Looks good now. Thanks!

PR #11949, opened 2018-05-10
deriving.0.8.1: require num with ocamlfind package

ygrek:
As build system expects to find num ocamlfind package, and it is not provided by base-num.
This effectively limits to ocaml-version >= 4.06.0


kit-ty-kate:
That seems surprising to me as ocamlfind installs a META file for the `num` package. Do you have a build log of some kind where I can see the error message ?

ygrek:
Sorry, looks like I misdiagnosied the issue. Cannot reproduce now. Thanks for noticing.

PR #11950, opened 2018-05-10
astring: depends on base-bytes

ygrek:
Witness
```
$ ocamlfind query astring  -r -format %p
bytes
astring
```

Omission of this dependency causes build failures.


samoht:
Thanks

PR #11951, opened 2018-05-10
Tidy 2.0.0 conf-* packages

dra27:
The opam repo upgrade always adds a dependency on the `ocaml` package when upgrading (@AltGr - is it worth a special case for the re-writer to assume that packages beginning `conf-` don't depend on `ocaml`. There are only two exceptions...)

The first commit of this branch removes that dependency from all the conf packages except `conf-libev` (which uses an OCaml script for discovery) and `conf-pic-switch` (which is actually a probe for OCaml configuration itself).

The next commit adds a lot of `{build}` predicates - largely academic. More importantly, a lot of missing `conf-pkg-config` dependencies are added in the following commit (this is quite important on platforms like Cygwin, where pkg-config isn't by default installed, even with the C compiler).

There are then a few miscellaneous bits of tidying - quick ping to @avsm just to be aware of the conf-ncurses changes and the two editor packages.


avsm:
This lgtm; do we need to modify the rewriter scripts to also do this for new conf packages?

dra27:
Yes - I was also going to work on at least getting Travis to recognise that it's a 2.0.0 branch PR and use opam 2 instead of opam 1 to run the matrix of tests.

AltGr:
Apparently the lint check for the `extra-files:` field is broken in Camelus : the errors above can be ignored.

AltGr:
Camelus fixed!

PR #11952, opened 2018-05-10
conduit: add missing constraints

samoht:


samoht:
Errors are due to too many concurrent downloads on opam 1.2, so merging.

PR #11954, opened 2018-05-10
lwt_react: better minimal constraint for jbuilder

samoht:
lwt_react.1.1.1 uses `copy_files` which is not available in old versions of
jbuilder.

/cc @aantron

### DIFF
--- a/packages/lwt_react/lwt_react.1.1.1/opam
+++ b/packages/lwt_react/lwt_react.1.1.1/opam
@@ -18,7 +18,7 @@ build: [ [ "jbuilder" "build" "-p" name "-j" jobs ] ]
 build-test: [ [ "jbuilder" "runtest" "-p" name ] ]
 
 depends: [
-  "jbuilder" {build & >= "1.0+beta10"}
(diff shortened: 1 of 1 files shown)


mseri:
Looks good thanks!

aantron:
Thanks @samoht, I've updated it in Lwt as well.

`lwt_react` actually *doesn't* use `copy_files`. It is used in `lwt` (for `lwt.unix`), which happens to be in the same repo. I guess Jbuilder hits the `copy_files` while figuring out *not* to build `lwt.unix`.

aantron:
@rgrinberg, @diml: this is a relatively minor issue, but the problem is that package `lwt` requires Jbuilder 1.0+beta14, because it uses `copy_files`. That makes `lwt_react` and `lwt_ppx` (#11999) also require 1.0+beta14, because they are in the same repo, even their build doesn't actually use `copy_files`.

It's probably not worth fixing, but I want to ping you in case it is somehow easy to fix, or is more important than I think.

diml:
So far Jbuilder has been pretty good at being backward compatible, so bumping the minimum required version shouldn't be an issue in general.

PR #11955, opened 2018-05-10
Old versions of ocaml-git do not compile on 4.06

samoht:


samoht:
Yup, just pushed a fix

samoht:
Hum, interesting error on 4.04.2 while compiling nocrypto:

```
 File "_none_", line 1:
# Error: broken invariant in parsetree: Let with no bindings.
# Command exited with code 2.
```

samoht:
nocrypto error fixed in #11964

samoht:
The remaining errors are travis checks failing to find a solution to install git 1.4.10 (which is ok I think -- not sure why this is considered an error). So I think this is good to merge.

mseri:
I agree. Thanks!

PR #11956, opened 2018-05-10
old websockets needs an old version of ocplib-endian

samoht:
/cc @vbmithr

### DIFF
--- a/packages/websocket/websocket.2.2/opam
+++ b/packages/websocket/websocket.2.2/opam
@@ -17,6 +17,7 @@ depends: [
   "ppx_deriving" {>= "2.0"}
   "nocrypto" {>= "0.5.0"}
   "conduit" {>= "0.8.3"}
+  "cstruct" {< "3.2.0"}
   "ocamlbuild" {build}
 ]
 depopts: [


mseri:
TheCI error now is because the old cohttp wasn’t safe-string compliant. I guess we cannot do much about that. Same for stog

mseri:
Stog 0.14.0 fails in a weird way:
```
# ### checking required tools and libraries ###
# checking for xmlm... Package xmlm found with version v1.3.0, but wanted >= 1.1
# 
# You way have a look at config_check.log for details.
```

hannesm:
@mseri I address this in #11960 -- xmlm upstream (git master, unreleased) uses since just after `v1.3.0` `%%VERSION_NUM%%` which will translate to `1.3.x` / `.1.4.x` (without leading `v`) in the future.

hannesm:
though travis complains about build-test in cohttp 0.20.2 / 0.21.1 with 4.06.0

samoht:
@hannesm I am looking at these now

samoht:
The remaining issues should be fixed by https://github.com/ocaml/opam-repository/pull/11962 and https://github.com/ocaml/opam-repository/pull/11963

mseri:
There may or may not be further failures in the stog revdeps. What we see currently is an infrastructure error: 
```
docker: Error: No such image: sha256:40a245857eea654981be4e612cc80dde9e0979615f394386af9b0859be6eb8b2.
```

I have tried rerunning all or part of the build multiple times without success, I keep getting the same error. I think it's good to merge

PR #11957, opened 2018-05-10
yurt.0.1 does not work on 4.06

samoht:
/cc @zshipko


samoht:
The conduit issue on 4.04 should be fixed by https://github.com/ocaml/opam-repository/pull/11958

hannesm:
thanks!

PR #11960, opened 2018-05-10
fix stog 0.13.0 and stog 0.14.0

hannesm:
- stog 0.13 is not safe-string compatible
- stog 0.13 and 0.14 require xmlm not 1.2.0 or 1.3.0 (which both use a `v` prefix in ocamlfind)


mseri:
Thanks!!

PR #11961, opened 2018-05-10
base64: add missing lower bounds for alcotest

samoht:
Alcotest.check only appeared in alcotest.4.0.0


samoht:
Some revdeps are broken but I don't intend to fix them this time. Although a bit sad to these broken packages :-(

PR #11962, opened 2018-05-10
ppx_type_conv.v0.9.0: use an old version of ppx_core

samoht:
Otherwise duplicated entries are generated. /cc @diml 

Not too sure about why this bug happens (which is similar to https://github.com/mirage/ocaml-cohttp/issues/573 but with ppx_deriving.4.2.1)

To reproduce it:

```
$ opam install --switch=4.06.0 cohttp.0.20.2 ppx_deriving.4.2.1 ppx_type_conv.v0.9.0 ppx_core.v0.9.3
```


diml:
That seems fine. Basically ppx_type_conv.v0.9.1 + ppx_core.v0.9.3 contain changes to work well in combination with ppx_deriving 4.2. Other combination of the v0.9 versions are not expected to work so well.

PR #11964, opened 2018-05-10
nocrypto: use a non-buggy version of ppx_sexp_conv

samoht:
/cc @hannesm and @diml 

Without this, there is a risk to hit something similar to https://github.com/mirage/ocaml-uri/issues/92

@diml is there any reason to keep that buggy package in the repository? Would probably be safer to just remove it instead of adding bounds to every packages.


diml:
I don't know what is the policy for removing packages in opam, but if that's possible, then yes I'm all for removing the buggy version.

samoht:
Just making the package unavailable should be enough (with `available: [false]`)

PR #11968, opened 2018-05-11
Build opam 2 for 2.0.0 branch builds

dra27:
This tweaks the Travis build to use opam 2 for PRs or builds of the 2.0.0 branch.

The change would become fractionally simpler with https://github.com/ocaml/opam-depext/pull/96 and it will obviously want bumping to RC2 when it comes out. It does mean that opam version passed in the Travis matrix is completely ignored.

### DIFF
--- a/.travis-ci.sh
+++ b/.travis-ci.sh
@@ -32,6 +32,21 @@ else
   fi
 fi
 
+if [ "$TRAVIS_BRANCH" = "2.0.0" ]; then
+  echo Building opam 2.0.0
+  cd $TRAVIS_BUILD_DIR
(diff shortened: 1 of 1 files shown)


dra27:
For now I did it on purpose to leave code in place for if it needed to jump to a newer opam build quickly ... I guess once RC2 is out it could be commented out or converted to a binary

kit-ty-kate:
Ok fair enough. Hum otherwise there seems to be some failures that look related in Travis.

kit-ty-kate:
ping @dra27

XVilka:
Seems not needed anymore?

PR #11970, opened 2018-05-11
old cohttp tests do not work on 4.06

samoht:


samoht:
The issue with tyxml.2.2.0 (which makes podge fails) has been reported upstream: https://github.com/ocsigen/tyxml/issues/203

PR #11972, opened 2018-05-12
Add sqlexpr, ppx_sqlexpr and pa_sqlexpr 0.9.0.

mfp:
Hand-tested on 4.02.3, 4.05.0, 4.06.0.
Not installable on 4.07 because ppx_core isn't either yet.


mfp:
The pa_sqlexpr failure for 4.06 should be addressed by the last commit (the installation of its estring fails on 4.06).

As for the 4.03 one, I don't know why opam is trying to install sqlexpr 0.7.1 (which fails because it tries to use the now-removed lwt.syntax package) instead of the latest version 0.9.0.

PR #11980, opened 2018-05-14
Package radis.0.1

dinosaure:
### `radis.0.1`

Radix tree implementation

Radis is a little library to provide an implementation of a Radix tree
(specialized with a scalar key - like a string, bigarray or array). This project
is a part of ocaml-git to have a fast access to an immutable store. In this way,
remove operation should be the slowest operation and this data-structure is
focused on lookup operation when the key is specifically a hash (see
[digestif](https://github.com/mirage/digestif)).

The idea behind this library is to provide a fast access to an immutable store
(like git). So, in my mind, if we put a new object in this store, it can not be
deleted. It's why remove operation should be slow - of course, in a git store,
it's possible to delete an useless object (see `git gc`), however this
computation does not appear in this way.

---
* Homepage: https://github.com/dinosaure/radis
* Source repo: https://github.com/dinosaure/radis.git
* Bug tracker: https://github.com/dinosaure/radis/issues

---

---
v0.1 2018-05-07 Paris (France)
----------------------------------------

- First release of `radis`

### DIFF
--- a/packages/radis/radis.0.1/descr
+++ b/packages/radis/radis.0.1/descr
@@ -1,0 +1,14 @@
+Radix tree implementation
+
+Radis is a little library to provide an implementation of a Radix tree
+(specialized with a scalar key - like a string, bigarray or array). This project
+is a part of ocaml-git to have a fast access to an immutable store. In this way,
+remove operation should be the slowest operation and this data-structure is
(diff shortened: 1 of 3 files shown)


kit-ty-kate:
All green. Thanks !

PR #11988, opened 2018-05-14
ocaml-cgi v0.9

rixed:
With agreement from JC Filliatre (@backtracking) I pointed the URL to my
github fork of ocaml-cgi, which fixes a few issues and add a couple of
features. In particular, the code is now compatible with latest version
of OCaml compiler.
Also, I tried to improve the opam file.


kit-ty-kate:
Looks ok. Sorry for the time it took. Thanks a lot!

PR #11990, opened 2018-05-15
Add 2 new Verimag packages: lustre-v6 and lutin. Update lutils and rdbg  along the way.

jahierwan:
Add 2 new Verimag packages: lustre-v6 and lutin. Update lutils and
rdbg to their current version along the way.

Also add a post-install message to say where to find the very last
version of those 4 Verimag packages.


kit-ty-kate:
Mmh looks good so far but camelus is giving an error message that I don't understand. cc @AltGr

jahierwan:
Well I've managed to install it several times
cf for instance here : https://ci.ocamllabs.io/log/saved/docker-build-547eea3f67e186f9bd29f76e6627be9f/605a4fd2543c1b1acfa9ebc374f726490883362c

It is a dependency issue. I've fixed this problem in the past.

jahierwan:
i need to depend on mlgmpidl, and not mlgmp

jahierwan:
So, according to travis jobs,  my problem is at opam::41 : 
   depexts: [ "libgmp-dev" "libmpfr-dev" "m4"]
On alpine arch, the package "libgmp-dev" does not exist

@kit-ty-kate  your commit on mlgmp is unrelated to this PR AFAIUnderstand, since I normaly do not depend on it [...] Ah, but my opam file says differently (it is generated by oasis2opam, i missed that). I should rather remove that dependency on mlgmp I suppose.

Drup:
Please do not remove old packages. Just add the new versions.

jahierwan:
I still have one red flag in https://travis-ci.org/ocaml/opam-repository/jobs/380600137
The only difference with https://travis-ci.org/ocaml/opam-repository/jobs/380600138 
(that works fine) is the INSTALL_LOCAL env variable that is set to 1, which force the use of the system switch.

The thing is that I cannot reproduce the problem. I've even tried anaked ubuntu docker image where i've installed ocaml and opam from scratch (forcing the use of system switch) but it worked fine also.

Since it is a "num package since ocaml 4.06" issue, maybe I would need a 4.06 system switch, which I don't know where to find a docker image of.

If anyone has an idea...

jahierwan:
Yes, I know. The thing is that
- I've added this dep in the opam file
- i can not reproduce the pb locally (via docker) on  arch running ocaml 4.06 or higher using my repo

When looking at the #19073.2 log file, i can read line 6012 "Lutin.2.54 installed successfully". 
But for some reason, the travis script tries again and the second times it fails. 

which script generates this log ? It would help to understand in which context it fails

kit-ty-kate:
Sorry, that line is missleading :/ In travis we first need to run opam install with `--dry-run` to get the list of dependencies and then really call `opam install`. So `lutin.2.54` was never installed.

The script itself is quite tedious to read so I would not advice this. Both failures appear on 4.06.0 with the system switch. You might want to try with that.

jahierwan:
ok, i'm now able to reproduce the problem on a docker machine, after installing ocaml 4.06 from source. 

googling, i realized that the z3 package has the same issue (Big_int not found). And indeed, 'opam install z3' fails which exactly the same message. So how z3 can be green on ocaml 4.06 there http://51.145.134.72/ ?

jahierwan:
Ok, thanks. I think I've figured it out (cf https://gricad-gitlab.univ-grenoble-alpes.fr/verimag/synchrone/lutin/commit/847a16d7b3412e1c8d923e3255766d4b67528101)

dra27:
It looks as though your build system relied on the num library being installed to OCaml's lib directory. When installed on OCaml 4.06.0, the num package installs itself to OCaml's lib directory by default to maximise compatibility (for some old packages which still don't use ocamlfind).

However, opam cannot install packages in a system switch to OCaml's lib directory (because for a system switch, that directory is outside opam's control). So opam's num package does a "proper" ocamlfind install to a package directory. That means that the .cmi files for num are no longer in the standard library directory, so your build system needs to know to add the appropriate `-I` where before it was getting away with it (as I think you've discovered).

This is relatively easy to test locally as you can always compile OCaml 4.06.1 by hand and install it to /usr/local temporarily (that usually won't interfere with your actual system OCaml compiler). In opam 2, you can then do `PATH` mangling to maintain multiple system switches, or just let opam 1.2.2/2.0.0 automatically recompile the system switch when it notices the change in compiler.

Drup:
Build system bugs notwithstanding, you are still removing old packages.

dra27:
I agree - also note that `findlib` files aren’t needed any more.

jahierwan:
@Drup which old packages ? the very old ones that I promised to put back (once I've fixed all travis issues) or the intermediate ones (Lutin 2.53,2.54, 2.55)? 

If is the latter case, what is the rationale? In my last commit, I've forgotten to remove 2.54, and now 
travis fails to build version 2.54 even tough I've committed version 2.56!

jahierwan:
I've put back old packages. Is there something more you want me to do before merging ?

samoht:
Everything is green, thanks for your efforts!

PR #11991, opened 2018-05-15
Fixed hash of qtest.2.4

dinosaure:
I'm not sure about the process release of `qtest` and I think it's because GitHub regenerate assets (and hash) automatically. In this case, we need to wait a new _fixed_ asset of `qtest.2.4` before to merge this PR.


dinosaure:
cc @c-cube

samoht:
That seems to address the issue, so I am merging this :-)

PR #11992, opened 2018-05-15
ppx_type_conv.v0.9.1 needs a recent version of deriving

samoht:
Fixes https://github.com/mirage/irmin/issues/494

To reproduce:

```
$ opam install ppx_deriving.4.1.5 ppx_type_conv.v0.9.1 -y && opam reinstall nocrypto -y -v
...
File "src/hash.ml", line 115, characters 12-16:
Error: Cannot locate deriver sexp
...
```

whereas:

```
$ opam install ppx_deriving.4.2 ppx_type_conv.v0.9.1 -y && opam reinstall nocrypto -y -v
Done.
```

### DIFF
--- a/packages/ppx_type_conv/ppx_type_conv.v0.9.1/opam
+++ b/packages/ppx_type_conv/ppx_type_conv.v0.9.1/opam
@@ -16,4 +16,7 @@ depends: [
   "ocaml-migrate-parsetree" {>= "0.4"}
   "ppx_derivers"
 ]
+conflicts: [
+  "ppx_deriving" {< "4.2"}
+]
 available: [ocaml-version > "4.03.0"]


samoht:
I am not sure why that dependency didn't existed before, so waiting for @diml to confirm that's the right fix.

diml:
The way ppx_type_conv interacts with ppx_deriving changed in v0.9.1. ppx_type_conv >= v0.9.1 doesn't use ppx_deriving, however it is indeed not compatible with ppx_deriving < 4.2. So this should be a conflict with ppx_deriving < 4.2 instead of a dependency constraint.

samoht:
That makes sense, thanks for the explanation! I have force-pushed to add a conflict instead of a dependency.

samoht:
Revdeps should be fixed by https://github.com/ocaml/opam-repository/pull/11997

samoht:
All green, so merging.

PR #12001, opened 2018-05-16
+ qfs 0.7

ygrek:


ygrek:
my mistake, updated url file with correct checksum for existing tarball

samoht:
Thanks it's failing now because of depext, but I guess that's what is expected so merging.

ygrek:
yes, indeed, thanks

PR #12003, opened 2018-05-16
aws-s3.3.0.0

andersfugmann:
* Add parameter to specify scheme(http|https)
* Fix IAM handling for lwt version.
* Always use http when accessing s3.
* Support copying s3 -> s3
* Support multipart upload
* Support permanent/temporary redirect error code
* Return an error indicating the redirect to successive calls
* Switch to use Digistif 
* Remove support for gzip (Users are encurraged to set the encoding and use ezgzip instead)
* Compatibility with base.v0.11.0

### DIFF
--- a/packages/aws-s3-async/aws-s3-async.3.0.0/descr
+++ b/packages/aws-s3-async/aws-s3-async.3.0.0/descr
@@ -1,0 +1,10 @@
+Ocaml library for accessing Amazon S3 - Async version
+
+This library provides access to Amazon Simple Storage Solution (S3).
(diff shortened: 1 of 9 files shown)


hcarty:
@andersfugmann The source URL is giving a 404.

andersfugmann:
Thanks. Works much better now.

hcarty:
Thanks for the quick fix!  Core 0.11.x [doesn't build on Alpine](https://github.com/janestreet/core/issues/109) so agreed those failures are unrelated.

PR #12005, opened 2018-05-16
Adding first version of partition maps to opam.

rleonid:


rleonid:
Ugh, sorry this isn't ready.

PR #12011, opened 2018-05-17
flowtype.0.27.0: restrict to os != linux

avsm:
In bulk builds, this particular (historical) version of Flow
causes a timeout in builds as it seems to be in an infinite loop
in a Makefile.  The problem is fixed in later versions, so it is
just this one version affected, and is only happening on Linux
toolchains.

This PR just disables Linux support, which will allow the old
version to be built by disabling that constraint, but also unblocks
CI resources in the short term.

cc @gabelevi


avsm:
Since this is a historical version and blocking builds, I'll merge this for now

PR #12013, opened 2018-05-17
Bulk-remove build tag from ppx dependencies

copy:
See the discussion in https://github.com/ocaml/opam-repository/issues/11852

This PR removes the `build` tag from ppx dependencies in all packages in the opam universe.

It removes the `build` tag from these dependencies:

- "bisect_ppx"
- "js_of_ocaml-ppx"
- "lwt_ppx"
- "ppx_bitstring"
- "ppx_blob"
- "ppx_cstruct"
- "ppx_deriving"
- "ppx_deriving_yojson"
- "ppx_distr_guards"
- "ppx_fields_conv"
- "ppx_import"
- "ppx_jane"
- "ppx_monadic"
- "ppx_optcomp"
- "ppx_sexp_conv"
- "ppx_tools"
- "ppx_tools_versioned"
- "ppx_typerep_conv"

Which was found using:

```sh
grep -r 'ppx.*{.*build'  | grep -v ppxfind|grep -v ocamlbuild |grep -v ppx_core|grep -v ppx_driver |grep -v ppx_metaquot |grep -v ppx_type_conv |grep -v base-no-ppx |grep -v '"bisect"' |grep -o '"[^"]*"' |sort |uniq
```

`ppx_core`, `ppx_driver`, `ppx_metaquot`, `ppx_type_conv` and `ppx_type_conv`
were excluded because they are deprecated and unlikely to see further updates.

Removal of the build tag was performed using two sed scripts, which I suggest
to be reviewed:

```sed
s/\("bisect_ppx"\|"js_of_ocaml-ppx"\|"lwt_ppx"\|"ppx_bitstring"\|"ppx_blob"\|"ppx_cstruct"\|"ppx_deriving"\|"ppx_deriving_yojson"\|"ppx_distr_guards"\|"ppx_fields_conv"\|"ppx_import"\|"ppx_jane"\|"ppx_monadic"\|"ppx_optcomp"\|"ppx_sexp_conv"\|"ppx_tools"\|"ppx_tools_versioned"\|"ppx_typerep_conv"\)\s*{\s*build\s*}/\1/
s/\("bisect_ppx"\|"js_of_ocaml-ppx"\|"lwt_ppx"\|"ppx_bitstring"\|"ppx_blob"\|"ppx_cstruct"\|"ppx_deriving"\|"ppx_deriving_yojson"\|"ppx_distr_guards"\|"ppx_fields_conv"\|"ppx_import"\|"ppx_jane"\|"ppx_monadic"\|"ppx_optcomp"\|"ppx_sexp_conv"\|"ppx_tools"\|"ppx_tools_versioned"\|"ppx_typerep_conv"\)\(\s*{\s*\)build &/\1\2/
```

After executing these scripts the above `grep` command returns nothing.

### DIFF
--- a/packages/JsOfOCairo/JsOfOCairo.1.0.1/opam
+++ b/packages/JsOfOCairo/JsOfOCairo.1.0.1/opam
@@ -11,7 +11,7 @@ build-test: ["jbuilder" "runtest" "-p" name]
 depends: [
   "jbuilder" {build}
   "js_of_ocaml-compiler" {build & >= "3.0" & < "4.0"}
-  "js_of_ocaml-ppx" {build & >= "3.0" & < "4.0"}
+  "js_of_ocaml-ppx" { >= "3.0" & < "4.0"}
   "js_of_ocaml" {>= "3.0" & < "4.0"}
   "cairo2"
 ]
--- a/packages/JsOfOCairo/JsOfOCairo.1.1.1/opam
+++ b/packages/JsOfOCairo/JsOfOCairo.1.1.1/opam
@@ -11,7 +11,7 @@ build-test: ["jbuilder" "runtest" "-p" name]
 depends: [
   "jbuilder" {build & >= "1.0+beta17"}
   "js_of_ocaml-compiler" {build & >= "3.0" & < "4.0"}
-  "js_of_ocaml-ppx" {build & >= "3.0" & < "4.0"}
+  "js_of_ocaml-ppx" { >= "3.0" & < "4.0"}
   "js_of_ocaml" {>= "3.0" & < "4.0"}
   "General" {test & >= "0.5.0"}
   "cairo2" {test & >= "0.5"}
--- a/packages/amf/amf.0.1.0/opam
+++ b/packages/amf/amf.0.1.0/opam
@@ -27,13 +27,13 @@ remove: ["ocamlfind" "remove" "amf"]
 depends: [
   ("oasis" {build} | "oasis-mirage" {build})
 
-  "ppx_deriving" {build}
-  "ppx_sexp_conv" {build}
+  "ppx_deriving"
+  "ppx_sexp_conv"
   "base-threads"
   ("core" {>= "v0.9.1"})
   "stdint"
   "sexplib"
-  "bisect_ppx" {build}
+  "bisect_ppx"
   "bisect_ppx-ocamlbuild" {build}
   "ounit" {build}
 ]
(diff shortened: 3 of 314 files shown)


hannesm:
I'm not sure about the reasoning in the referenced issue.  Applying the same argument to build systems (version x may lead to invalid code, version (x+1) fixes this behaviour), should all the `build` dependencies for build systems be dropped in opam-repositories? //cc @avsm @AltGr @samoht

I can see the value, being able to derive from `opam list` output (plus system packages) a reproducible binary.  I'm worried about too many recompilations just because `ocamlbuild` or `dune` got released.

copy:
> I'm not sure about the reasoning in the referenced issue. Applying the same argument to build systems (version x may lead to invalid code, version (x+1) fixes this behaviour), should all the build dependencies for build systems be dropped in opam-repositories? //cc @avsm @AltGr @samoht

I tend to agree, but this PR only changes ppx dependencies, which need to trigger rebuilds after upgrades as they usually contain runtime libraries.

mseri:
@copy I see your point, but I am not sure about this. For example `bisect_ppx` is generally used for coverage builds and not opam builds, I think its presence in those opam files has been an oversight and should not trigger a rebuild if the library is updated. Similarly I think ppx_cstruct should not trigger a rebuild unless cstruct is updated, and probably cstruct should always be listed as a dependency if ppx_cstruct is used.

copy:
@mseri Indeed, I also included ppxs that don't have runtime libraries, as per [this comment](https://github.com/ocaml/opam-repository/issues/11852#issuecomment-387732193):

> BTW, if I understand the meaning of {build} correctly, then ppx rewriters should never be flagged as build only dependencies, even when they don't have runtime dependencies. The reason is that if the code they generate changes, then all their reverse dependencies must be rebuilt as well.

However, I'd glady revert to the less controversial change of only changing ppxs that have runtime libraries.

mseri:
:+1: Let’s get more opinions before reverting anything

rgrinberg:
I agree with this change. That said, this does not solve the issue in the long term as people will still continue submitting packages with `{build}` dependencies. How about a discuss thread to warn people to stop using this feature for ppx.

avsm:
I also agree with this change; we should err on the side of being conservative with recompiles and doing more if there's a risk of missing a material change.

Instead of or as well as a discuss thread, I think we should get into the habit of adding checks to the Camelus bot to automate this as part of the lint step.

copy:
Is there anything missing from this PR to be merged? Any interjections?

I had a quick look over the failures and they mostly come from failing old packages (that were touched by this PR), network errors and already broken packages.

samoht:
Merging as this seems to be a safe thing to do.

PR #12015, opened 2018-05-17
Packages fstar.0.9.6.0, kremlin.0.9.6.0

tahina-pro:
# New F* version: 0.9.6.0
F* is a functional programming language with dependent types, refinement types, SMT-based verification, tactics and extraction to OCaml.

Latest changes:
- Meta-F*: A metaprogramming and tactic framework, as described in this [report](https://arxiv.org/abs/1803.06547). Code samples are in examples/tactics, examples/native_tactics and the `FStar.Tactics` and `FStar.Reflection` libraries. Many people contributed a lot to this work, especially Guido Martinez.

- Improved type inference with two-phase typechecking: We now build verification conditions for a program after a first phase of type inference. This improves inference of implicit arguments and reduces our trust in the type inference. Thanks to Aseem Rastogi!

- Caching typechecked modules: F* emits ".checked" files, an on-disk representation of a typechecked module that can be read back later. This significantly reduces the time to load a module's dependences.

- Many other changes. For more info, see https://github.com/FStarLang/FStar/releases/tag/v0.9.6.0

# New package: KreMLin 0.9.6.0
KreMLin is a companion tool for F*, by @msprotz , to extract F* programs to C. To be eligible, such F* programs must fit in Low*, a low-level subset of F* supporting first-order effectful functions with non-recursive data types, while their specifications and proofs can still enjoy all F* features such as dependent types. F* comes with a library of buffers to model the C memory, thus enabling high-level verification for low-level code.

For more information:
- on this KreMLin release: https://github.com/FStarLang/kremlin/releases/tag/v0.9.6.0
- on using KreMLin: https://fstarlang.github.io/lowstar/html/
- on the theory of Low* : https://arxiv.org/abs/1703.00053

### DIFF
--- a/packages/fstar/fstar.0.9.6.0/descr
+++ b/packages/fstar/fstar.0.9.6.0/descr
@@ -1,0 +1,1 @@
+An ML-like language with a type system for program verification.
--- a/packages/fstar/fstar.0.9.6.0/opam
+++ b/packages/fstar/fstar.0.9.6.0/opam
@@ -18,6 +18,7 @@ depends: [
   "ppx_deriving"
   "ppx_deriving_yojson"
   "process"
+  "ocaml-migrate-parsetree"
 ]
 depexts: [
   [ ["osx" "homebrew"] ["coreutils"] ]
--- a/packages/fstar/fstar.0.9.6.0/opam
+++ b/packages/fstar/fstar.0.9.6.0/opam
@@ -1,0 +1,46 @@
+opam-version: "1.2"
+version: "0.9.6.0"
+maintainer: "taramana@microsoft.com"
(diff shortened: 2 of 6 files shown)


tahina-pro:
Note to the official opam repository managers: from this version on, ocamlfind is now more than just a build dependency. Indeed, in such usage scenarios as native tactics, F* needs to dynamically load some of its libraries that it needs to find through ocamlfind, such as fstar-tactics-lib, which is installed as part of this package.

avsm:
Thanks for the package, and the note about the ocamlfind dependency! You may want to announce this on <https://discuss.ocaml.org>, where we have a Community category and an `announce` tag for this purpose.

Issue #10337, opened 2017-09-25
Fedora 26 uses pkgconf-pkg-config as an alternative to pkgconfig


mauny:
(OPAM version 1.2.2)

Fedora 26 uses `pkgconf-pkg-config` as an alternative to `pkgconfig`:
  - `yum` correctly installs `pkgconf-pkg-config` when requested to install `pkgconfig`
  - but the test used by OPAM after installation (`rpm -q pkgconfig`) fails.

mauny:
More info about this:
  - the change is mentioned [here](https://fedoraproject.org/wiki/Fedora_26_talking_points#pkgconf_as_system_pkg-config_implementation)
  - according to file depext.ml (line 214), Fedora seems to be the only distribution that double-checks (`yum install -y`, then `rpm -q`) that the package installation has been successful. If installation fails, `yum install` has a non-zero status AFAIK, which should be sufficient.

Unless I am missing something, a simple fix could be to suppress the call to `rpm -q`: `yum install -y pkgconfig` correctly installs, when necessary, `pkgconf-pkg-config`, and returns 0 status in both cases.

dbuenzli:
@mauny I'm not what you are reporting here (erroring context is missing). Is it a problem with the `conf-pkg-config` package (i.e. a problem with `opam-repository`) or a problem with opam's `depext` functionality. If it's the later I suggest you report this on https://github.com/ocaml/opam-depext tracker.

mauny:
You are right: the problem is with opam's `depext` functionality. I'll report it where you suggest. Thanks and sorry for the noise.

dbuenzli:
No problem.

PR #12018, opened 2018-05-18
conf-pkg-config: Install pkg-config symlink on OpenBSD

mato:
I'm submitting this as a version 1.1 of `conf-pkg-config` since that seems like the right thing to do for this "meta" package.

From the commit:

> The /usr/bin/pkg-config provided by OpenBSD does not support ${pcfiledir}, which breaks .pc files used by MirageOS. Unlike FreeBSD, installing devel/pkgconf does not install a binary called "pkg-config", so we add a symlink to it in the OPAM "bin" directory at package installation time.
> 
> We also inform the user, since this will change the pkg-config binary on their PATH.

Tested on OpenBSD current with OPAM 2 (including `opam lint`), all seems to work as expected.

/cc @hannesm @adamsteen Solo5/solo5#226

### DIFF
--- a/packages/conf-pkg-config/conf-pkg-config.1.1/opam
+++ b/packages/conf-pkg-config/conf-pkg-config.1.1/opam
@@ -8,6 +8,15 @@ license: "GPL"
 build: [
   ["pkg-config" "--help"]
 ]
+install: [
+  ["ln" "-s" "/usr/local/bin/pkgconf" "%{bin}%/pkg-config"] {os = "openbsd"}
+]
+remove: [
+  ["rm" "-f" "%{bin}%/pkg-config"] {os = "openbsd"}
+]
+post-messages: [
+  "conf-pkg-config: A symlink to /usr/local/bin/pkgconf has been installed in the OPAM bin directory (%{bin}%) on your PATH as 'pkg-config'. This is necessary for correct operation." {os = "openbsd"}
+]
 depexts: [
   [["debian"] ["pkg-config"]]
   [["ubuntu"] ["pkg-config"]]
@@ -23,4 +32,3 @@ depexts: [
   [["osx" "homebrew"] ["pkg-config"]]
   [["freebsd"] ["pkgconf"]]
 ]
-available: [ os != "openbsd" ]
--- a/packages/conf-pkg-config/conf-pkg-config.1.1/descr
+++ b/packages/conf-pkg-config/conf-pkg-config.1.1/descr
@@ -1,0 +1,3 @@
+Virtual package relying on pkg-config installation.
+This package can only install if the pkg-config package is installed
+on the system.


packages/conf-pkg-config/conf-pkg-config.1.1/opam
```diff
@@ -0,0 +1,34 @@
...
+authors: ["Francois Berenger"]
+homepage: "http://www.freedesktop.org/wiki/Software/pkg-config/"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+dev-repo: "git://github.com/ocaml/opam-repository"
+license: "GPL"
+build: [
+  ["pkg-config" "--help"]
+]
+install: [
+  ["ln" "-s" "/usr/local/bin/pkgconf" "%{bin}%/pkg-config" {os = "openbsd"}]
```
hannesm: shouldn't the `{ os = "openbsd" }` be outside of the `[ ... ]` build instructions?

mato: The manual seems to suggest that either way is fine: 

> `install: [ [ <string> { <filter> } ... ] { <filter> } ... ]`: the list of commands that will be run in order to install the package. [...]

and `opam lint` does not complain, so I guess it should be fine?

hannesm: according to a `git grep`, both are used in opam-repository, so this is fine with me.

mato:
CI seems upset on Fedora 27, but that is unrelated to this change.

packages/conf-pkg-config/conf-pkg-config.1.1/opam
kit-ty-kate: I agree with @hannesm that should be outside. Otherwise only `"%{bin}%/pkg-config"` will be present on openbsd but the rest will be present on any plateform.

The CI doesn't fail because it seems that `ln -s foo` is a valid command.

hannesm: @kit-ty-kate raises a good point here.

hannesm:
@kit-ty-kate IMHO bumping the version number is crucial here, for the sake of avoiding to debug potential issues: consider that there are N people with OpenBSD who have conf-pkg-config.1.0 installed -- their systems won't reinstall/upgrade to 1.1.  but in case they want to install the upcoming solo5-kernel-ukvm, this will fail.  most `conf-pkg-config` are `{build}` dependencies, and won't be recompiled anyways.

mato:
@kit-ty-kate Good point about the different interpretation of the filter when it's inside the `[ ... ]`. Fixed.

packages/conf-pkg-config/conf-pkg-config.1.1/opam
```diff
@@ -0,0 +1,34 @@
+opam-version: "1.2"
+maintainer: "unixjunkie@sdf.org"
+authors: ["Francois Berenger"]
+homepage: "http://www.freedesktop.org/wiki/Software/pkg-config/"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+dev-repo: "git://github.com/ocaml/opam-repository"
+license: "GPL"
+build: [
+  ["pkg-config" "--help"]
```
kit-ty-kate: on openbsd is there a `pkg-config` binary installed or is it named `pkgconf` ?

packages/conf-pkg-config/conf-pkg-config.1.1/opam
```diff
@@ -0,0 +1,34 @@
...
+]
+install: [
+  ["ln" "-s" "/usr/local/bin/pkgconf" "%{bin}%/pkg-config"] {os = "openbsd"}
+]
+remove: [
+  ["rm" "-f" "%{bin}%/pkg-config"] {os = "openbsd"}
+]
+post-messages: [
+  "conf-pkg-config: A symlink to /usr/local/bin/pkg-config has been installed in the OPAM bin directory (%{bin}%) on your PATH. This is necessary for correct operation." {os = "openbsd"}
```
kit-ty-kate: this message does not match the install directive

packages/conf-pkg-config/conf-pkg-config.1.1/opam
mato: The issue is that OpenBSD has a `/usr/bin/pkg-config` as part of the base system, however that version does not support `${pcfiledir}`. The version installed as the `devel/pkgconf` package _does_ support this, but gets installed only as `pkgconf`. Hence the necessity for the symlink to get OPAM packages to use the correct version.

packages/conf-pkg-config/conf-pkg-config.1.1/opam
mato: > this message does not match the install directive

Fixed, and also reworded the commit message to be clearer.

mato:
@kit-ty-kate Regarding the version number change, I agree with @hannesm; not incrementing the version number would mean that existing installations will not get the fix.

avsm:
The older versions also need to be marked as `!openbsd` I think

kit-ty-kate:
> @kit-ty-kate Regarding the version number change, I agree with @hannesm; not incrementing the version number would mean that existing installations will not get the fix.

Why ? Any changed/upgraded packages will be rebuilt as expected.

mato:
@kit-ty-kate:

> Why ? Any changed/upgraded packages will be rebuilt as expected.

Are you saying that if a switch has an _existing_ installation of `conf-pkg-config.1.0`, then merely updating the `opam` file in the repository will cause `opam update && opam upgrade` to _reinstall_ the `conf-pkg-config` package? Are you sure about this? I realise it will get the "new" metadata from the repository, but that is not enough since in this case a reinstall needs to happen for the fix to be applied.

Further, not changing the version number makes it impossible to diagnose if the fixed version has been installed or not, since `opam list`, `opam show`, etc. would be identical for fixed/non-fixed versions.


> Otherwise why adding a new version for this ? Why not just change the current one ? That would avoid some redundancies. :/

Can you elaborate on that? What redundancies?

@avsm:

> The older versions also need to be marked as !openbsd I think

I can do that if you prefer, but need agreement on the version number change.

avsm:
I think opam uses a subset of the `opam` file metadata to determine if a package needs recompilation if any of those fields change -- not just the version number.

The function that extracts that subset of fields is `opamFile.effective_part` https://github.com/ocaml/opam/blob/master/src/format/opamFile.ml#L2738

cc @samoht @altgr for confirmation

mato:
> So yes, bumping the version number is crucial if you want to trigger a general recompilation for user installs.

Just to confirm the way forward, so we keep the version number bump to `1.1`, and I'll amend to commit to mark `1.0` as unavailable on openbsd?

@samoht @avsm please confirm.

samoht:
@mato that sounds like a good plan

mato:
Updated to add `available: [ os != "openbsd" ]` to conf-pkg-config.1.0.

Note that the Fedora 27 CI failure is due to https://github.com/ocaml/opam-repository/issues/10337, I'm not going to address that here.

PR #12019, opened 2018-05-19
Revert "Use Lwt 2.4.3 GitHub archive"

aantron:
Reverts ocaml/opam-repository#11793.

<p>
The Lwt 2.4.3 archive on ocsigen.org is now back online, so I propose we switch back to using it. We were using the Lwt 2.4.3 git tag since #11793. The ocsigen.org archive is a known-working release, while the git tag might have problems due to not generating the OASIS-generated build system correctly, for example.  

<p>
See https://github.com/ocsigen/lwt/issues/589#issuecomment-389113679. I'm also going to download all the ocsigen.org archives, and attach them to the GitHub releases, so that we have a backup.

<p>
cc @hannesm @mseri


hannesm:
thanks, I adjusted the revdeps failure (osc < 0.1.4) in https://github.com/ocaml/opam-repository/pull/12020

PR #12020, opened 2018-05-19
osc < 0.1.4 is not compatible with safe-string

hannesm:
discovered by revdeps build in https://github.com/ocaml/opam-repository/pull/12019


hannesm:
//cc @johnelse (NB: the most recent osc release in opam-repository is compatible with safe-string / 4.06)

johnelse:
This is fine by me, thanks for the cc.

kit-ty-kate:
Looks good. Thanks !

PR #12021, opened 2018-05-19
+webmachine.0.6.0

seliopou:


samoht:
Thanks! Would you be able to add constraints to fix the revdeps failures? This is needed for datakit-ci and irmin-http.

seliopou:
Done.

seliopou:
Looks like it's failing now because `git-unix` is missing constraints on `cohttp` or some related module; and `irmin-unix` is failing because of a safe-string issue? The last one I'm not too sure about.

samoht:
Thanks! I will fix the remaining issues in an other PR.

PR #12025, opened 2018-05-21
ocamldap 2.4.0

kit-ty-kate:


mseri:
I know this kind of renames are annoying, we had to deal with a lot of them in xapi, but I think they are a good thing. Do you think it would be better to release as `ocamldap.dummy` or `ocamldap.legacy` and `ldap.2.4.0` to emphasize the change? Or to `ocamlldap.3.0.0` and `ldap.3.0.0` to signal the breaking change?

kit-ty-kate:
I don't think `dummy` or `legacy` are good names since they seem to imply that the package does nothing in the former case or "keep the old behaviour" (see below) in the second.
The thing here is that there are no breaking changes so I don't feel comfortable bumping the version number more than expected. Everything should work as before and have the same findlib packages.

mseri:
Sounds reasonable to me

mseri:
Fedora is failing with `package pkgconfig is not installed` but I think this is due to the pkgconfig name being inferred as `pkgconf-pkg-config-1.3.12-2.fc27.x86_64`. It is related to some of the dependencies and I am not sure what is the right fix, so I would not consider it blocking the merge unless you believe you can also fix that

kit-ty-kate:
Yes, this has been fixed by https://github.com/ocaml/opam-depext/pull/94 but not yet available for the CI because of the docker cache :/ Since you are happy with the current PR I'll merge it. Thanks for your review !

PR #12036, opened 2018-05-22
Package lbfgs.0.9

Chris00:
### `lbfgs.0.9`

Bound-constrainted optimization in many variables

This library provide an algorithm to minimize and maximize functions
of many variables on bounded or unbounded domains.  This is a binding
to L-BFGS-B, a library for Large-scale Bound-constrained Optimization.

---
* Homepage: https://github.com/Chris00/L-BFGS-ocaml
* Source repo: https://github.com/Chris00/L-BFGS-ocaml.git
* Bug tracker: https://github.com/Chris00/L-BFGS-ocaml/issues

---

---
0.9 2018-05-22
--------------

- Compile with Dune/Jbuilder
- OPAM depext for OSX

### DIFF
--- a/packages/lbfgs/lbfgs.0.9/descr
+++ b/packages/lbfgs/lbfgs.0.9/descr
@@ -1,0 +1,6 @@
+Bound-constrainted optimization in many variables
+
+This library provide an algorithm to minimize and maximize functions
+of many variables on bounded or unbounded domains.  This is a binding
+to L-BFGS-B, a library for Large-scale Bound-constrained Optimization.
+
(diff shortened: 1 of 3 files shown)


Chris00:
The revdep failure is due to `cppo_ocamlbuild` not found  — unrelated to this PR then.

Chris00:
@kit-ty-kate I saw that and just pushed an update.

Chris00:
@kit-ty-kate A `conf-fortran` would be nice but IMHO it should go beyond ensuring the presence of a fortran compiler but also provide means of calling it with the right options (don't know where to open an issue for that though).

Chris00:
If `gfortran` is not present, the current package also checks for `g95` and `g77`.  It also passes `-fPIC` but all FORTRAN compilers may not accept that.  For MacOSX, it also calls `gfortran --print-file-name libgfortran.dylib` to pass the right location for `-lgfortran` to the script building the library.  I haven't explored Win$ but options are likely to be different.  That's why I think the package should come with some helpers for configuration scripts.

Chris00:
Do not know what the problem with Debian is.  It works well on my (Debian) machine.

PR #12037, opened 2018-05-22
Package fftw3.0.8.1

Chris00:
### `fftw3.0.8.1`

Binding to the famous Fast Fourier Transform library FFTW

---
* Homepage: https://github.com/Chris00/fftw-ocaml
* Source repo: https://github.com/Chris00/fftw-ocaml.git
* Bug tracker: https://github.com/Chris00/fftw-ocaml/issues

---
### opam-lint failures

---

---
0.8.1 2018-05-22
----------------

- Fix documentation (issue #11).
- Add bigarray dependency.

### DIFF
--- a/packages/fftw3/fftw3.0.8.1/opam
+++ b/packages/fftw3/fftw3.0.8.1/opam
@@ -1,0 +1,38 @@
+opam-version: "1.2"
+authors: ["Christophe Troestler <Christophe.Troestler@umons.ac.be>"]
+maintainer: "Christophe.Troestler@umons.ac.be"
+license: "LGPL-2.1 with OCaml linking exception"
+homepage: "https://github.com/Chris00/fftw-ocaml"
+dev-repo: "https://github.com/Chris00/fftw-ocaml.git"
+bug-reports: "https://github.com/Chris00/fftw-ocaml/issues"
+doc: "https://Chris00.github.io/fftw-ocaml/doc"
+tags: ["FFT"]
+build: [
+  [ "jbuilder" "subst" ] {pinned}
+  [ "jbuilder" "build" "-p" name "-j" jobs ] { os != "darwin" }
+  [ "env" "FFTW3_CFLAGS=-I /usr/local/include -L /usr/local/lib -I /usr/local/include -L /usr/local/lib"
+    "jbuilder" "build" "-p" name "-j" jobs ] { os = "darwin" }
+]
+build-test: [[ "jbuilder" "runtest" "-p" name "-j" jobs ]]
+depends: [
+  "jbuilder" {build}
+  "cppo"     {build}
+  "configurator" {build}
+  "base"     {build}
+  "stdio"    {build}
+  "lacaml" {test}
+]
+available: [ ocaml-version >= "4.03.0" ]
+depexts: [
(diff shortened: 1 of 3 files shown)


kit-ty-kate:
All green apart from the OSX build.
Could you have a look at the failure on archimedes ?
```
#=== ERROR while installing archimedes.0.4.18 =================================#
# opam-version 1.2.2
# os           darwin
# command      ocaml setup.ml -configure --prefix /Users/travis/.opam/system --enable-graphics
# path         /Users/travis/.opam/system/build/archimedes.0.4.18
# compiler     system (4.06.1)
# exit-code    1
# env-file     /Users/travis/.opam/system/build/archimedes.0.4.18/archimedes-17816-548b09.env
# stdout-file  /Users/travis/.opam/system/build/archimedes.0.4.18/archimedes-17816-548b09.out
# stderr-file  /Users/travis/.opam/system/build/archimedes.0.4.18/archimedes-17816-548b09.err
### stderr ###
# File "setup.ml", line 7667, characters 15-32:
# Warning 3: deprecated: String.capitalize
# Use String.capitalize_ascii instead.
# ocamlfind: Package `graphics' not found
# W: Field 'pkg_graphics' is not set: Command ''/Users/travis/.opam/system/bin/ocamlfind' query -format %d graphics > '/var/folders/my/m6ynh3bn6tq06h7xr3js0z7r0000gn/T/oasis-88c180.txt'' terminated with error code 2
# E: Cannot find findlib package graphics
# E: Failure("1 configuration error")
```

packages/fftw3/fftw3.0.8.1/opam
```diff
@@ -0,0 +1,39 @@
...
+  "jbuilder" {build}
+  "cppo"     {build}
+  "configurator" {build}
+  "base"     {build}
+  "stdio"    {build}
+  "archimedes" {test}
+  "lacaml" {test}
+]
+available: [ ocaml-version >= "4.03.0" ]
+depexts: [
```
kit-ty-kate: Also, it would be wise to move the depexts to an external meta `conf-fftw` package I think. This would avoid redundancies, but it can be done later.

Chris00: What redundancies?  IMHO, any package that wants to use FFTW should use this one (or contribute to it if it is not enough) and not depend again on the C code.  Or did I misunderstand your point?

Chris00:
>  Could you have a look at the failure on archimedes ?

I guess that it is because the Graphics module is not available — and Archimedes depends on it.
(As a side note, I do not intend to develop the current version of Archimedes any further.  I am in the process of splitting its functionality in several small packages and then only have a new version of Archimedes raise from its ashes...  I just lack time...)

packages/fftw3/fftw3.0.8.1/opam
kit-ty-kate: I'm just saying that gaving a single meta package that lists the external dependency for every systems is slightly better than having several versions of this package repeating this list (it might also need some adjustments and it's much better to have one meta package to handle everything than several that might be forgotten.

kit-ty-kate:
Other than the point I just mention (that can be done later if needed) everything is green. Thanks !

PR #12046, opened 2018-05-24
functory: new release 0.6

backtracking:
fixed compilation with OCaml 4.06
fixed OCaml version for package functory 0.5 as well


backtracking:
It seems that the remaining errors (on packages cubicle and mezzo) are not coming from this patch.
So could it be merged as is?

samoht:
The mezzo error should be fixed by https://github.com/ocaml/opam-repository/pull/12056

@mebsout are you aware that http://cubicle.lri.fr/cubicle-1.0.2.tar.gz disappeared and so cubicle.1.0.2 cannot be installed anymore?

PR #12057, opened 2018-05-25
lbfgs: not available on 4.06 (safe-string) and fix the version of oas…

samoht:
…is to use

/cc @Chris00


Chris00:
Thanks.  lbfgs 0.9 works with the latest version of OCaml — see https://github.com/ocaml/opam-repository/pull/12036

PR #12059, opened 2018-05-25
old versions of owl do not work with safe-string

samoht:


samoht:
The eigen issue should be fixed by https://github.com/ocaml/opam-repository/pull/12061

kit-ty-kate:
Thanks !

PR #12062, opened 2018-05-26
Nocrypto patches for compatibility with sexplib/ppx_sexp_conv > v0.11.0

copy:
This commit submits @gasche's fixes from https://github.com/mirleft/ocaml-nocrypto/pull/144, as nocrypto's maintainers are not responsive. This fixes https://github.com/mirleft/ocaml-nocrypto/issues/143.

cc @hannesm

### DIFF
--- a/packages/nocrypto/nocrypto.0.5.4-1/descr
+++ b/packages/nocrypto/nocrypto.0.5.4-1/descr
@@ -1,0 +1,9 @@
+Simpler crypto
+
+
+nocrypto is a small cryptographic library that puts emphasis on the applicative
+style and ease of use. It includes basic ciphers (AES, 3DES, RC4), hashes (MD5,
+SHA1, SHA2), public-key primitives (RSA, DSA, DH) and a strong RNG (Fortuna).
+
+RSA timing attacks are countered by blinding. AES timing attacks are avoided by
+delegating to AES-NI.
--- a/packages/nocrypto/nocrypto.0.5.4-1/files/0002-add-missing-runtime-dependencies-in-_tags.patch
+++ b/packages/nocrypto/nocrypto.0.5.4-1/files/0002-add-missing-runtime-dependencies-in-_tags.patch
@@ -1,0 +1,47 @@
+From cad7cfe15ae8eca95c4e284f3a679c35842659f2 Mon Sep 17 00:00:00 2001
+From: Gabriel Scherer <gabriel.scherer@gmail.com>
+Date: Mon, 26 Mar 2018 16:09:16 +0200
+Subject: [PATCH 2/6] add missing runtime dependencies in _tags
+
+Binaries in <bench/*>, <tests/*> depend on ppx_sexp_conv's runtime
+library within ppx_sexp_conv.
+
+The packed modules <src/nocrypto.cm{x,o}> also depend on the package
+ppx_sexp_conv: its presence at pack-creation time influences the
+generated .cmi interface, see
+
+  https://github.com/ocaml/opam-repository/pull/11628#issuecomment-375697444
+
+Note: the package ppx_sexp_conv.runtime-lib would suffice, but it is
+only available as such under recent ppx_sexp_conv versions, so its
+explicit use would make the build description (needlessly)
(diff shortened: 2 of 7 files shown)


hannesm:
AFAIU, the patched META file in here will include the entire `ppx_sexp_conv` library (which includes compiler-libs?).  I tried to incorporate all the open PRs in a branch https://github.com/hannesm/ocaml-nocrypto/tree/safely which I'm testing atm, where I ensured nocrypto for my use cases (esp. MirageOS unikernels) to work (and newer sexplib).

copy:
@hannesm You're right, this shouldn't be merged as-is (I thought I had checked this [earlier](https://github.com/mirleft/ocaml-nocrypto/pull/144#issuecomment-381364182), but I can now see that I was mistaken, as `-I …/compiler-libs` and others are included when compiling a library with nocrypto after my changes). I will use @diml's method instead, which should select `ppx_sexp_conv` on < v0.11 and `ppx_sexp_conv.runtime-lib` on >= v0.11.

copy:
Rebased with @diml's patches. I checkd and this version doesn't include base or compiler-libs.

hannesm:
another remark (I didn't have time to look in detail or test this PR): could you please increase the patch level (i.e. a 0.5.4-1) of nocrypto (see https://github.com/ocaml/opam-repository/issues/10531), otherwise it's hard to tell which nocrypto variant (the one with your patches, or the one without) is installed and makes debugging tricky.

copy:
Good point, moved the patches to a new release: nocrypto.0.5.4-1.

mseri:
I think we should merge this release. Does anybody have any objection?

hcarty:
No objection here.

hannesm:
I didn't have time to test this PR completely (but AFAICT my main concern - adding unnecessary runtime dependencies - has been addressed).  did anyone test this PR with both >0.11 and <0.11 ppx_sexp_conv/sexplib releases!?

concerns: why is it patch 0002, 0004, 0005, 0006 (why not 0001 and 0003)?  also, 0004 adds `ppx_sexp_conv` to `META`, which is then adjusted by 0005 -- can we remove the chunk from 0004?

mseri:
I am using it already with ppx_sexp_lib >0.11, did not test with older ones. I agree that the numbering is a bit weird, but maybe it is better to keep them as they are give that is practically the import of the PR of a different author (and, hopefully, we will get a new nocrypto soon)

copy:
>concerns: why is it patch 0002, 0004, 0005, 0006 (why not 0001 and 0003)?

The reason for this is that 0003 reverts 0001, see [gasche's branch](https://github.com/mirleft/ocaml-nocrypto/compare/v0.5.4...gasche:v0.5.4-ocamlbuild-pack).

> also, 0004 adds ppx_sexp_conv to META, which is then adjusted by 0005 -- can we remove the chunk from 0004?

We could do that. We could also collect all changes into a single patch, but I think it's better to leave the patches directly derived from the commits (they were created using `git format-patch`).

gasche:
I think I reverted this commit because I pushed a fix in ocamlbuild, but then the fix will only be available in the next ocamlbuild release (which I've been slow in putting out...).

I'm afraid that I forgot why I pushed the revert on my fork. I think that it would be better if you did *not* include the revert commit: you would work with older ocamlbuild releases.

(I just tested to confirm that having the workaround in myocamlbuild.ml *and* a fixed version of ocamlbuild would keep working.)

copy:
> I think I reverted this commit because I pushed a fix in ocamlbuild, but then the fix will only be available in the next ocamlbuild release (which I've been slow in putting out...).
> 
> I'm afraid that I forgot why I pushed the revert on my fork. I think that it would be better if you did not include the revert commit: you would work with older ocamlbuild releases.

I  already included the reverted commit, it's the last patch (I initially kept the reverting commit, and cherry-picked the reverted commit in the end).

copy:
If there are no more objections, can we move on with merging this?

mseri:
I think we should

hcarty:
Thanks!

PR #12069, opened 2018-05-27
notty_async.0.1

yminsky:
Trying again, this time against opam 1.2, and not using the opam-publish script!

### DIFF
--- a/packages/notty_async/notty_async.0.1/opam
+++ b/packages/notty_async/notty_async.0.1/opam
@@ -1,12 +1,10 @@
 opam-version: "1.2"
-maintainer:   "yminsky@gmail.com"
+maintainer:   "Yaron M. Minsky <yminsky@gmail.com>"
 authors:      "Yaron M. Minsky"
 license:      "MIT"
 homepage:     "https://github.com/yminsky/notty_async"
 bug-reports:  "https://github.com/yminsky/notty_async/issues"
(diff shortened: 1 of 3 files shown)


packages/notty_async/notty_async.0.1/opam
```diff
@@ -0,0 +1,20 @@
...
+  ["jbuilder" "subst"] {pinned}
+  ["jbuilder" "build" "-p" name "-j" jobs]
+]
+
+depends: [
+  "jbuilder" {build & >="1.0+beta20"}
+  "notty"    {>= "0.2.1"}
+  "async"    {>= "v0.11.0"}
+  "base"     {>= "v0.11.0"}
+  "ocaml"    {>= "4.06.0"}
```
dra27: `ocaml` is a 2.0.0-only package, so this wants to be deleted...

packages/notty_async/notty_async.0.1/opam
```diff
@@ -0,0 +1,20 @@
...
+]
```
dra27: ... and below here add:
```
available: ocaml-version >= "4.06.0"
```

yminsky:
Better, now?

kit-ty-kate:
The Alpine failure is caused by https://github.com/janestreet/core/issues/109. Otherwise all seem good. Thanks!

PR #12074, opened 2018-05-28
Restrict atd->menhir dependency to build time

mjambon:
Addresses https://github.com/mjambon/atd/issues/113


kit-ty-kate:
Looks good. Thanks !

PR #12075, opened 2018-05-29
added packages/bst/bst.2.0.1

UnixJunkie:


UnixJunkie:
let's directly go to 3.0.0.
Cf. here: https://github.com/ocaml/opam-repository/pull/12076

PR #12080, opened 2018-05-30
added packages/minicli/minicli.4.0.0

UnixJunkie:
some type aliases are declared and used in the interface to make it more readable


kit-ty-kate:
Looks good. Thanks !

PR #12087, opened 2018-05-31
Package user-setup.0.7

rjbou:
### `user-setup.0.7`

Helper for the configuration of editors for the use of OCaml tools

This tool knows about several editors, and several OCaml editing tools existing
as opam packages. It automates the configuration of these editors, providing
base templates when appropriate, and suitably installing the editing tools in
the editor's configuration.

For example, it will configure your emacs or Vim to indent OCaml files using
[ocp-indent](http://www.typerex.org/ocp-indent.html) if you have that installed.

Opam-user-setup is designed to be suitable both to beginners not wanting to be
bothered with configuration files, and to people who manage them carefully.

It's customisable and reversible.

---
* Homepage: https://github.com/OCamlPro/opam-user-setup
* Source repo: https://github.com/OCamlPro/opam-user-setup.git
* Bug tracker: https://github.com/OCamlPro/opam-user-setup/issues

---

### DIFF
--- a/packages/user-setup/user-setup.0.7/descr
+++ b/packages/user-setup/user-setup.0.7/descr
@@ -1,0 +1,14 @@
+Helper for the configuration of editors for the use of OCaml tools
+
+This tool knows about several editors, and several OCaml editing tools existing
+as opam packages. It automates the configuration of these editors, providing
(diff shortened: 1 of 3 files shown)


kit-ty-kate:
All green. Thanks!

PR #12091, opened 2018-05-31
easy_xlsx 1.0

brendanlong:
Initial release

Also includes `open_packaging` and `spreadsheetml` for lower-level
interfaces.

### DIFF
--- a/packages/open_packaging/open_packaging.1.0/opam
+++ b/packages/open_packaging/open_packaging.1.0/opam
@@ -1,0 +1,25 @@
+opam-version: "1.2"
+authors: ["Brendan Long <self@brendanlong.com>"]
+maintainer: "self@brendanlong.com"
+homepage: "https://github.com/brendanlong/ocaml-ooxml"
+dev-repo: "https://github.com/brendanlong/ocaml-ooxml.git"
+bug-reports: "https://github.com/brendanlong/ocaml-ooxml/issues"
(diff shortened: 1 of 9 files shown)


brendanlong:
Hopefully this one will be easier now that I know what I'm doing..

packages/open_packaging/open_packaging.1.0/opam
```diff
@@ -0,0 +1,25 @@
...
+homepage: "https://github.com/brendanlong/ocaml-ooxml"
+dev-repo: "https://github.com/brendanlong/ocaml-ooxml.git"
+bug-reports: "https://github.com/brendanlong/ocaml-ooxml/issues"
+doc: "https://brendanlong.github.io/ocaml-ooxml/doc"
+
+build: ["jbuilder" "build" "-p" name "-j" jobs]
+build-test: ["jbuilder" "runtest" "-p" name "-j" jobs]
+
+depends: [
+  "base" {= "v0.10.0"}
```
kit-ty-kate: it is generally better to have something like `>= "v0.10.0" & < "v0.11.0"` instead because there might be patches (from upstream or from opam) in the future.
For now it's ok I guess and it can be fixed in a future PR.

I'll merge it as soon as the MacOS build finishes.

kit-ty-kate:
All green. Thanks!

PR #12095, opened 2018-06-01
calendar 2.04 was released 2014-10-29

Leonidas-from-XIV:


Leonidas-from-XIV:
I've added a `dev-repo` but commented it out (just like every other SVN package seems to do), let's see if it'll make the linter happy.

PR #12098, opened 2018-06-01
delimcc.2018.03.16

gasche:
This somewhat-newly released version of the Delimcc library fixes a compilation failure under GCC 7.x, which affects previous delimcc versions under all OCaml versions.

### DIFF
--- a/packages/delimcc/delimcc.2018.03.16/descr
+++ b/packages/delimcc/delimcc.2018.03.16/descr
@@ -1,0 +1,1 @@
+Oleg's delimited continuations library for byte-code and native OCaml
--- a/packages/delimcc/delimcc.2018.03.16/opam
+++ b/packages/delimcc/delimcc.2018.03.16/opam
@@ -10,4 +10,4 @@ build: [
 remove: [["ocamlfind" "remove" "delimcc"]]
 depends: ["ocamlfind" {build}]
 install: [make "findlib-install"]
-available: [ ocaml-version >= "4.04.0" ]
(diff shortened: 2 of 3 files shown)


gasche:
Note: previous release tarballs of delimcc were sourced from <https://github.com/zinid/delimcc/>, which is a modified version of Oleg's upstream delimcc archive, meant for opam deployment. The author of this for-opam distribution doesn't have much time to maintain it anymore.

The reason why a for-opam fork was necessary is somewhat trite: for a long time Oleg only maintained a single archive at a non-stable URL <http://okmij.org/ftp/continuations/caml-shift.tar.gz>, which was regularly updated. I got in touch with Oleg and he agreed to also upload timestamped archives on his webpage, so that the present PR uses the archive <http://okmij.org/ftp/continuations/caml-shift-20180316.tar.gz>.

(Of course I also tried to nudge Oleg into making a git repository for delimcc publicly available, but that didn't succeed. I could maintain a copy as a git repository on the side, but I found it easier to deal with just one upstream rather than two, so I am fine with the current situation instead.)

gasche:
Travis CI reports failures on MacOS, which I suspect are due to a combination of the MacOS test setting being a bit arcane (it seems to use a `system` switch where the `ocamlc` and `ocaml` are not installed in the same directory?) and the Makefile for delimcc being very arcane as well. I could patch them to better support MacOS, or just mark this release incompatible with MacOS, and wait until I get the motivation to update Oleg's Makefile code to something that pleases everyone and works there as well. I am leaning towards the second option.

kit-ty-kate:
All green. Thanks a lot !

PR #12099, opened 2018-06-01
lambda-term 1.13

diml:
Fixes compatibility with lwt>=4.0.0 and camomile>=1.0.0


mseri:
Now it is hardcaml.1.2.0 that is failing for unrelated reasons. If you know the fix by hearth it would be good to add the necessary bounds as well, otherwise imho we could merge this

diml:
Looking at the log, I would say this is a safe-string issue. I don't know the proper fix though

PR #12100, opened 2018-06-01
Ppxlib 0.2.1

diml:
There is a bug in ppxlib 0.1.0 and 0.2.0 that make them incompatible with the upcoming 1.0.0 release of Dune. This PR marks these two packages as incompatible with `dune` and release a 0.2.1 version that fixes the issue.

0.3.0 is not affected but it is not co-installable with ppx_driver v0.11.0, so we need this intermediate version.

### DIFF
--- a/packages/ppx_ast/ppx_ast.v0.11.0/opam
+++ b/packages/ppx_ast/ppx_ast.v0.11.0/opam
@@ -10,6 +10,6 @@ build: [
 ]
 depends: [
   "jbuilder" {build & >= "1.0+beta18.1"}
-  "ppxlib"   {>= "0.1.0" & <= "0.2.0"}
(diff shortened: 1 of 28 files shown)


diml:
I also updated the various `"ppxlib" <= "0.2.0"` constraints to `"ppxlib" < "0.3.0"`

diml:
The [diff between 0.2.0 and 0.2.1](https://github.com/ocaml-ppx/ppxlib/commit/b69b1a4ea01aef47c9297d156c8df506461ab82e) is extremely simple and doesn't touch the API, so I expect that all the failures already existed with 0.2.0. As a result I simply marked the various ppx packages as incompatible with ppxlib. There were a few other failures but they seemed unrelated.

PR #12103, opened 2018-06-02
irmin 1.0 is not compatible with -safe-string

samoht:
See http://51.145.134.72/4.06.1/bad/irmin.1.0.0


kit-ty-kate:
Tests fail but it's still an improvement from before and tests can be fixed later.
```
# ocamlfind: Package `irmin-watcher' not found
```

Thanks!

PR #12104, opened 2018-06-02
lazy-trie.1.0 is not compatible with -safe-string

samoht:
See http://51.145.134.72/4.06.1/bad/lazy-trie.1.0.0


kit-ty-kate:
Thanks!

PR #12106, opened 2018-06-02
Mark benchmark 1.3.1 available only for OCaml < 4.06.0

Chris00:
There are newer versions of benchmark that support more recent
compilers.


kit-ty-kate:
This looks good. Thanks !

PR #12123, opened 2018-06-06
ocaml-migrate-parsetree v1.0.11

diml:
### Changelog:

- Fix handling of `--impl/--intf`. Before the driver would crash if
  the file extension was neither `.ml` nor `.mli`


diml:
Most errors are download failres or unrelated. However it seems that wcs-lib is related, I can reproduce the failure by pinning ocaml-migrate-parsetree to the development version. I'm looking into the issue.

diml:
Actually I still get the error on a fresh switch without pinning anything. I can install wcs-lib in my default switch but not in a fresh one, so there must be another issue.

diml:
In summary, I think this PR is good to merge :)

PR #12124, opened 2018-06-06
Package mariadb.1.1.0

andrenth:
### `mariadb.1.1.0`

OCaml bindings for MariaDB

OCaml-MariaDB provides Ctypes-based bindings for MariaDB, including its
nonblocking API.

---
* Homepage: https://github.com/andrenth/ocaml-mariadb
* Source repo: https://github.com/andrenth/ocaml-mariadb.git
* Bug tracker: https://github.com/andrenth/ocaml-mariadb/issues

---

### DIFF
--- a/packages/mariadb/mariadb.1.1.0/descr
+++ b/packages/mariadb/mariadb.1.1.0/descr
@@ -1,0 +1,4 @@
+OCaml bindings for MariaDB
+
+OCaml-MariaDB provides Ctypes-based bindings for MariaDB, including its
+nonblocking API.
--- a/packages/mariadb/mariadb.1.1.0/opam
+++ b/packages/mariadb/mariadb.1.1.0/opam
@@ -1,0 +1,23 @@
+opam-version: "1.2"
+maintainer: "Andre Nathan <andrenth@gmail.com>"
+authors: "Andre Nathan <andrenth@gmail.com>"
+homepage: "https://github.com/andrenth/ocaml-mariadb"
+bug-reports: "https://github.com/andrenth/ocaml-mariadb/issues"
+license: "MIT"
+dev-repo: "https://github.com/andrenth/ocaml-mariadb.git"
+build: [
+  ["./configure" "--prefix=%{prefix}%"]
+  [make]
+]
+install: [make "install"]
+remove: [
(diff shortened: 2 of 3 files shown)


gasche:
Continous integration fails (see [log example](https://travis-ci.org/ocaml/opam-repository/jobs/388876775) because it apparently doesn't know about some C-side MariaDB constants.

Maybe you add a dependency to the `conf-mariadb` package, which encapsulates system dependencies on the MariaDB client library?

andrenth:
This isn't done because in Ubuntu 14.04 (used in the CI environment) those dependencies are not available.

gasche:
I'm a bit confused, but then I am not familiar with MariaDB package so apologies if I make you repeat things.

- Not all CI machines use Ubuntu, for example [this log failure](https://ci.ocamllabs.io/log/saved/docker-build-f573ddbb889cd2acaf7399e85a0eb28c/ff7f0b1ad3a3de0ca7e7e3bd1f5b7dba7647b564) apparently comes from a Debian Stretch machine
- I am not sure whether you mean that the conf-db package uses the wrong name for Ubunutu packages (in which case a solution would be to fix them), whether you mean that no Ubunutu package exists that expresses the right dependency, or whether you mean that the Ubuntu packages that would be needed are not available on the CI environment (which is weird, given that the CI uses `depext` to know which system packages to install, so using the right `conf-*` package should make things work better, not worse)

In any case, the CI is not the only thing to consider for justifying proper `depext` choices (and in particular `conf-*` dependencies); they are also useful in giving a clearer error to users than just an obscure build failure (as is observed in your Travis results).

hcarty:
@gasche Unfortunately the opam 1.x version of depext doesn't support enough nuance to have a properly working conf package due to package name changes across OS versions and file name changes across mariadb versions. From what I've seen, proper depext support wasn't feasible for the mysql package either.

gasche:
I see. @hcarty, can I nominate you as a more informed reviewer for this PR then? :-)

hcarty:
Unfortunately, given the CI + depext constraints we currently have, I don't think we can do much beyond check the validity of the content in the PR and trust @andrenth to not break their own code!

Thanks for the PR!

PR #12126, opened 2018-06-06
Update archive URL after repo migration

xguerin:


kit-ty-kate:
Looks ok. Thanks a lot!

PR #12130, opened 2018-06-07
Package ocaml-r.0.1.0

pveber:
### `ocaml-r.0.1.0`

Objective Caml bindings for the R interpreter

OCaml-R is a library that can be used to construct R values in memory,
convert them to OCaml values, and build clean wrappers to R
functions. It provide a simple means to develop bindings to any R
package.

---
* Homepage: https://github.com/pveber/ocaml-r
* Source repo: https://github.com/pveber/ocaml-r.git
* Bug tracker: https://github.com/pveber/ocaml-r/issues

---

### DIFF
--- a/packages/ocaml-r/ocaml-r.0.1.0/opam
+++ b/packages/ocaml-r/ocaml-r.0.1.0/opam
@@ -1,0 +1,16 @@
+opam-version: "1.2"
+maintainer: "philippe.veber@gmail.com"
+authors: ["Guillaume Yzyquel" "Maxence Guesdon" "Philippe Veber"]
+homepage: "https://github.com/pveber/ocaml-r"
+bug-reports: "https://github.com/pveber/ocaml-r/issues"
+license: "GPL"
+dev-repo: "https://github.com/pveber/ocaml-r.git"
+build: ["jbuilder" "build" "-p" name "-j" jobs]
+depends: [
+  "base"           { build }
+  "configurator"   { build }
+  "conf-r"         { build }
+  "conf-r-mathlib" { build }
+  "jbuilder"       { build }
+  "stdio"          { build }
+]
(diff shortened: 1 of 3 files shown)


packages/ocaml-r/ocaml-r.0.1.0/opam
```diff
@@ -0,0 +1,21 @@
...
+homepage: "https://github.com/pveber/ocaml-r"
+bug-reports: "https://github.com/pveber/ocaml-r/issues"
+license: "GPL"
+dev-repo: "https://github.com/pveber/ocaml-r.git"
+build: ["jbuilder" "build" "-p" name "-j" jobs]
+depends: [
+  "base" {build}
+  "configurator" {build}
+  "jbuilder" {build}
+  "ocamlfind"
```
kit-ty-kate: Do you still require ocamlfind ? I don't see any reference to `ocamlfind` or `Findlib` in your repository.

packages/ocaml-r/ocaml-r.0.1.0/opam
```diff
@@ -0,0 +1,21 @@
...
+  "stdio" {build}
+]
+depexts: [
```
kit-ty-kate: Is it possible to use the `conf-r` package instead ? Do you rely on the mathlib version ? (I don't know how it works)

packages/ocaml-r/ocaml-r.0.1.0/opam
pveber: Indeed I don't!

packages/ocaml-r/ocaml-r.0.1.0/opam
pveber: Actually I think it'd be safer to put both.

mseri:
Thanks. Currently the only failure seems to be due to some problem with R-mathlib on alpine, I am going to merge this. I will open a new PR if I understand what’s wrong on alpine

pveber:
Thanks a lot! It's not clear to me either what's wrong with Alpine. Maybe the package `R-mathlib` doesn't configure `pkg-config` adequately?

PR #12133, opened 2018-06-07
Add async_graphics 0.6.0

lpw25:
Adds async_graphics version 0.6.0. This updates async_graphics to work with recent versions of async. This PR also fixes the constraints on the earlier versions of async_graphics to stop them being used with some versions of async that they won't work with.

### DIFF
--- a/packages/async_graphics/async_graphics.0.6.0/opam
+++ b/packages/async_graphics/async_graphics.0.6.0/opam
@@ -13,7 +13,7 @@ remove: [["ocamlfind" "remove" "async_graphics"]]
 depends: [
   "ocamlfind"
   "graphics"
-  "async" {>= "109.09.00" & < "113.00.00"}
+  "async" {>= "113.00.00"}
   "ocamlbuild" {build}
 ]
 install: "./install.sh"
--- a/packages/async_graphics/async_graphics.0.6.0/opam
+++ b/packages/async_graphics/async_graphics.0.6.0/opam
@@ -13,7 +13,7 @@ remove: [["ocamlfind" "remove" "async_graphics"]]
 depends: [
   "ocamlfind" {build}
   "graphics"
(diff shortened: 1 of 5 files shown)


yomimono:
A few non-spurious-looking failures in Travis:

[OCAML_VERSION=4.02: an attempted build against async 113.33.03, which fails with `Unbound module Deferred`](https://travis-ci.org/ocaml/opam-repository/jobs/389245925#L2142)
[OCAML_VERSION=4.00: an attempted build against async 109.15.00, which fails in `lib/unix_syscalls.ml`](https://travis-ci.org/ocaml/opam-repository/jobs/389245927#L1023)

The Alpine failure is in building `core`, which seems like a bigger problem.

hcarty:
@yomimono I think the `core` failure is related to https://github.com/janestreet/core/issues/109

packages/async_graphics/async_graphics.0.6.0/opam
```diff
@@ -0,0 +1,19 @@
...
+bug-reports: "https://github.com/lpw25/async_graphics/issues"
+dev-repo: "https://github.com/lpw25/async_graphics.git"
+license: "LGPL-2.0 with OCaml linking exception"
+tags: [
+  "async"
+  "graphics"
+]
+remove: [["ocamlfind" "remove" "async_graphics"]]
+depends: [
+  "ocamlfind"
```
kit-ty-kate: Could you add the `{build}` attribute to `ocamlfind` ? It seems that your package doesn't use the `Findlib` library.

lpw25: Done

lpw25:
Ok, I've pushed a tigher async dependency, which should fix the frist Travis failure. The second seems unrelated to this package.

PR #12134, opened 2018-06-07
atdgen: fix deps

damiendoligez:
Newer versions of `atdgen` need `atd` >= 1.2.0.


kit-ty-kate:
Looks good. Thanks a lot !

PR #12135, opened 2018-06-07
sequence: add bound on ocaml version

damiendoligez:
`sequence` doesn't compile on 4.07.0 because we added the stdlib module `Seq`, with new functions in `Set` and `Map`.

A new release was done last week.

cc @c-cube


yomimono:
Thanks, @damiendoligez !

PR #12136, opened 2018-06-07
js_of_ocaml-compiler: add bound on ocaml version

damiendoligez:
Doesn't compile with 4.07 because of the change to `Env.summary`.


yomimono:
Thanks!

PR #12137, opened 2018-06-07
ppx_import: add bound on ocaml version

damiendoligez:
`ppx_import` won't compile on 4.07 because `Ident.t` is now an abstract type.
Fix sent upstream: ocaml-ppx/ppx_import#23.

cc @whitequark 

I also had to fix the download URLs because the project moved from whitequark to ocaml-ppx.


yomimono:
Revdep failure is a failure to pull the source for `systemverilog` and everything else looks fine.  Thanks, @damiendoligez !

PR #12138, opened 2018-06-07
add depexts for fedora in several conf-* packages

maroneze:


maroneze:
I just wanted to add Fedora constraints to some `conf-` packages to ensure `depext` finds them, but is it worth updating the files to remove some errors/warnings? Some of them seem *very* old.

mseri:
Given that we are modifying the files, a cleanup to make camelus happy as well is welcome :)

maroneze:
For the `authors` field, I followed the pattern "The <lib> authors" I found in some conf-* packages, otherwise I have no idea what to put in them. For fields `dev-repo` and `bug-reports` I just followed what seems to be the established convention. 
Finally, `conf-brotli` used `install` instead of `build` as target, so I changed it to `build`, which is what everyone else does.

mseri:
Thank you very much. I think what you did for the authors field is a reasonable compromise. I'll wait for the CI and if the fedora images are fine I'll merge.

maroneze:
The Travis build failed on Mac ("The sources of the following couldn't be obtained, aborting: depext 1.0.5", not sure what I can do about it) and one of the builds errored with an apparent loss of connection, and I don't know how to retry it.

mseri:
I have restarted them but I wouldn’t really worry about them, they are not testing fedora

kit-ty-kate:
I fixed some issues with some of the packages so that they can be tested on fedora

kit-ty-kate:
Oh well it won't test fedora… mmh ok let's merge this anyway. Thanks a lot!

PR #12144, opened 2018-06-08
ppxx: add bound on OCaml version

damiendoligez:
`ppxx` won't compile on 4.07.0 because `Ident.t` has become abstract.

reported upstream with a patch: https://bitbucket.org/camlspotter/ppxx/issues/3/fix-for-407

cc @camlspotter


kit-ty-kate:
Thanks a lot!

PR #12147, opened 2018-06-08
ocp-index: add bound on jbuilder version

damiendoligez:
Older versions of jbuilder don't like ocp-index's `jbuild` file when lambda-term is not installed.

cc @AltGr @Drup


Drup:
I'm slightly surprised, as I think jbuilder beta19 was *not* out when the last version of ocp-index was released, but I'll trust you and CI!

kit-ty-kate:
CI is fine. I guess I'll merge it, feel free to ask for a finer constraint later if you feel the need.

PR #12152, opened 2018-06-09
hidapi: fix release 1.0 (critical bugfix)

vbmithr:


mseri:
I think it would be better to release a new version instead of overwriting the old one. Users would not know what they have installed in this way

vbmithr:
Would you leave a package that will segfault your program if some function is used in opam?

mseri:
How would I know that I have installed the correct one? I'd rather modify 1.0.0 to be uninstallable and release 1.0.1

This is a nontrivial discussion to be hones, there is an open issue about it https://github.com/ocaml/opam-repository/issues/10531 and a previous discussion on the forum: https://discuss.ocaml.org/t/introducing-a-yanked-mechanism-in-opam/1707

vbmithr:
Actually I think I deleted the tag 1.0 corresponding to the old release. I understand the point of not overwriting releases, I have read the discussions you showed me, but it would really help if you could make an exception for this package, I just released the broken version one week ago, and not likely used by anybody.

Or give me clear instructions in what you want to merge this (knowing that I might not be able to reproduce the old tag/commit).

mseri:
I am merging this, please create a new package `hidapi.1.0-1` with the exact same files as this one, so that people that had the old library can update automatically

PR #12162, opened 2018-06-10
old versions of irmin need an old version of re

samoht:
Fix http://51.145.134.72/4.05.0/bad/irmin-unix.0.9.9


samoht:
The `ppx_cstruct` issues should be fixed by #12167 
`mirage-tcpip`'s test should be fixed by #12169

samoht:
The remaining issues in Travis are due to https://github.com/ocaml/ocaml-ci-scripts/issues/229

PR #12164, opened 2018-06-10
dns: fix installation of optional sub-packages with `-t`

samoht:
Fix `ocamlfind: Package `dns.mirage' not found` errors


samoht:
The `ppx_cstruct` error should be fixed by #12167

PR #12168, opened 2018-06-10
+angstrom.0.10.0

seliopou:


mseri:
`emile` is failing also with angstrom `0.9.0`

kit-ty-kate:
The failure on Alpine is due to https://github.com/janestreet/core/issues/109

Looks good otherwise. Thanks!

PR #12171, opened 2018-06-11
really tell R to execute the check.r script

UnixJunkie:


UnixJunkie:
I think it is OK to merge this one
@pveber

mseri:
Would it be possible to fix also the revdeps issues or they are unrelated?

UnixJunkie:
Hmmm. Those packages are difficult to install fully automatically. Because they assume that the user has some R packages installed in user space (or can install them). So, there are many dependencies that are needed and that are difficult to track (R might need to compile Fortran or C++ programs and god knows what else).
Users of those packages are encouraged to proceed as follows:
```
post-messages:
  """
Please interact with R to install needed things in user-space:
R
install.packages('Marix', repos='http://cran.r-project.org')
install.packages('xgboost', repos='http://cran.r-project.org')"""
    {failure}
```
This is an example from the orxgboost package in opam.

UnixJunkie:
I am sorry that my packages taping into the R ecosystem are a little hackish.
But, they do the job and provide some ML functionalities to OCaml users.
Though, they don't play that nicely with opam.

mseri:
It’s ok, I was just trying to understand. I am going to merge this then, thanks!

PR #12177, opened 2018-06-11
amqp-client.2.0.2

andersfugmann:
* Do not fetch remote amqp spec to fix builder with opam 2.0 with sandboxing enabled.


andersfugmann:
The errors on alpine is from compiling core, and is unrelated to these packages.

hcarty:
Thanks!

PR #12175, opened 2018-06-11
digestif.0.5: needs newer ocamlbuild

damiendoligez:
With older `ocamlbuild` I get a mysterious linker error.

cc @Eyyub @dinosaure

PR #12181, opened 2018-06-12
rawlink: fix dependency

damiendoligez:
rawlink needs a recent version of ocamlbuild.

This is probably the same problem as #12175 

cc @haesbaert


kit-ty-kate:
Thanks a lot!

PR #12182, opened 2018-06-12
syslog: add bound on OCaml version (was: add -unsafe-string flag)

damiendoligez:
syslog wasn't updated for the switch to safe-string and I can't find a maintainer for it, so here is a minimal backward-compatible patch.


pmetzger:
> syslog wasn't updated for the switch to safe-string and I can't find a maintainer for it, so here is a minimal backward-compatible patch.

Wider question: as there are a bunch of such packages, would adopting some of them in a new GitHub project be a good idea?

mseri:
The safe/unsafe string patches are dangerous. When used with old compilers they can make packages fail to build because of inconsistent interface assumptions. I’d rather mark it as incomparible with 4.06

damiendoligez:
>I’d rather mark it as incomparible with 4.06.

That was my first idea too, but there is no maintainer, so that means the death of this package. I could also produce a source patch and add that to the opam package, but maybe killing the package is the best option.

mseri:
If it create inconsistent interfaces issues it would be its death as well. I'd imagine and hope that as soon as somebody needs the ocaml-4.06 version you would get a PR

mseri:
Is the package available on github? From the opam file it's unclear if there is any documentation or a website, and the url points to a debian repository

pmetzger:
> That was my first idea too, but there is no maintainer, so that means the death of this package.

Adopting it and changing the Strings to Bytes in appropriate places would surely be straightforward, would it not? Were there a "life support" repository somewhere, I'd happily help with a group effort of keeping useful but abandoned packages compiling with the latest OCaml.

mseri:
Yes I was asking about the support repo because I wanted to send the PR :)

damiendoligez:
This idea should be discussed in a wider forum than this PR, but I don't know where exactly. @avsm any suggestion?

avsm:
I'm tracking this in https://github.com/ocaml/obi/issues/19 -- I'll add this to the (very shortly released) opam-ci tool and start a discuss thread.

damiendoligez:
I've updated the PR to just mark the package incompatible with 4.06+

kit-ty-kate:
Looks good. Thanks!

PR #12186, opened 2018-06-13
camltc.0.9.6: ocaml 4.06.1

toolslive:


UnixJunkie:
All tests are green, this looks OK to merge.
camltc looks like the only simple key-value store solution in the ocaml ecosystem nowadays so it has some importance.
I miss the berkeleydb days; not sure if tokyocabinet is higher performance than bdb.

mseri:
Is Tokyo Cabinet actually maintained?

toolslive:
Tokyo Cabinet is **not** maintained. This library actually uses the last official release + some bugfixing patches. It's still used in the non-entreprise editions of Arakoon. The combination of the bindings and tokyo cabinet are really stable and has been running in production environments for years. 

However, we have seen tokyo cabinets go corrupt on power failures occasionally (we have seen the same with leveldb and rocksdb as well, so nothing new there, and it has more to do with SSD/NVMe design than database design).

mseri:
Good to know! Thanks for the informations. If you have time I think it would be worth mentioning this in ocamlverse: https://ocamlverse.github.io/

toolslive:
mention it there, or add it do the `descr` file of this opam lib ?

UnixJunkie:
I am considering adding things to camltc (the hashtbl backend).

toolslive:
I would not do it, or at least, first measure the performance difference with a c-based benchmark (iirc tokyo cabinet provides some tooling for this).  The simple reason is that a 2009 SSD is not a 2018 SSD/NVMe.

UnixJunkie:
some computers still have spinning disks (usually the very large ones)

toolslive:
The tokyo cabinet B+ tree uses the hash database for node placement. Just measure so you can see what you could gain, before you write the bindings, and don't blame me if your disappointed.

mseri:
@toolslive please add it to the `descr` file, I'll send a PR to add the libraries to ocamlverse

PR #12180, opened 2018-06-12
ppx_ast: add bound on OCaml version

damiendoligez:
ppx_ast.v0.10.0 fails to compile since 4.06.1, apparently because ocaml/ocaml@f6d53cc38f87c67fbf49109f5fb79a0334bab17a changed the type `Warnings.t`.

cc @janestreet

PR #12192, opened 2018-06-14
ppx_ast: add bound on OCaml version (second attempt)

damiendoligez:
Follow-up to #12180 .

@samoht the error message is:
```
- File "src/location_helper.ml", line 1, characters 63-99:
- Error: The constructor Ocaml_common.Warnings.Deprecated
-        expects 3 argument(s), but is applied here to 1 argument(s)
-       ocamlc src/.ppx_ast.objs/ppx_ast__Location_helper.{cmi,cmo,cmt} (exit 2)
```


kit-ty-kate:
Thanks!

PR #12195, opened 2018-06-14
omod 0.0.1 install: deal with toplevel eco-system mess.

dbuenzli:


dbuenzli:
So you can't use `ocaml:lib` in `opam v1.2.0` ?

kit-ty-kate:
All green. Thanks!

PR #12196, opened 2018-06-14
Add protocol-9p{,-unix}.0.12.1

djs55:
Also adjust protocol-9p-tool.0.12.0 to work with both.


kit-ty-kate:
It seems that `Named_pipe_lwt` was accessible through `protocol-9p-unix/protocol_9p_unix.cma` in the previous version and used directly by `datakit.0.1{1,2}.0` without linking against `named-pipe.lwt`.

Could you add a constraint to those two versions of `datakit` ?

kit-ty-kate:
All green now. Thanks a lot! (sorry I was not clear enough on the constraints I was asking on datakit)

djs55:
Thanks for the constraints and merge!

PR #12202, opened 2018-06-15
Solo5: Add solo5-kernel-* for v0.3.0

mato:
Solo5 v0.3.0 was released today, so starting from the bottom up for the
relevant MirageOS packages.


kit-ty-kate:
Thanks a lot!!

PR #12203, opened 2018-06-15
orocksdb.0.4.0

domsj:


mseri:
The failures are due to missing liborocksdb in certain old systems. Nothing to worry about

PR #12204, opened 2018-06-15
old portia versions that are unmaintained won't work with newer compiler

rixed:
But to be honest, I see no reason to keep these old unmaintained versions into the archive.
Shall I delete them instead?

### DIFF
--- a/packages/portia/portia.0.1/opam
+++ b/packages/portia/portia.0.1/opam
@@ -18,4 +18,4 @@ depexts: [
 post-messages: [
   "This package requires asciidoc to build the doc." {failure}
 ]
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
Currently some maintainers would like to keep packages around as much as possible. If you really want to "delete" them you'd have to add `available: false` to those packages.

For now this is good enough. Thanks a lot!

rixed:
When I mean to delete I mean to delete the directory altogether. I don't think that adding `available:false` would improve the archive over a more accurate `available` info.
This is more a question of policy I guess. i guess one day some old versions of everything will have to be deleted, but I might still be living a bit in the XXth century :)
Thank you!

PR #9328, opened 2017-05-26
Remove findlib files

whitequark:
See ocaml/opam#2948.

PR #12205, opened 2018-06-16
repository cleanups: remove 'findlib' files, rename desc and desr to descr

hannesm:
this PR does multiple changes
- removes leftover `findlib` files (see #9328 for earlier discussion)
- moves `desc` and `desr` files to `descr`
- removes `.chamo.sourceview.buffers` which seems to be accidentally commited


hcarty:
This looks good to me. The Travis failure is unrelated.

pmetzger:
Just a side note: a request name that is just "cleanups" is a bad idea. It means people don't know whether to pay attention to the particular request or not because they don't know what it is about.

hannesm:
@pmetzger good point, renamed.  sorry.

kit-ty-kate:
Looks good. Thanks a lot!!

PR #12206, opened 2018-06-16
ocaml-freestanding 0.3.0

mato:
Solo5 v0.3.0 is now merged in #12202, so continuing up the stack with a v0.3.0 release of ocaml-freestanding.


hannesm:
thanks! all green!

PR #12209, opened 2018-06-16
mirage-solo5 v0.3.0

mato:
Continuing up the stack from #12202, #12206.

(this time with the correct release artifact run through topkg)


mato:
After discussion on IRC with @hannesm, revdeps here show some missing upper bounds on `mirage-solo5 {< "0.3.0"}` on historic Solo5 packages, so added those in 7a0e0ae.

hannesm:
thanks, all green! :)

PR #12214, opened 2018-06-18
Heptagon version 1.05.00

gdelaval:
- Option -simple-scheduler active for scheduling post-ctrln code generation
 - Handle implication operator in Sigali backend
 - Compatibility with reatk >= 0.14


kit-ty-kate:
Looks good. Thanks!

PR #12217, opened 2018-06-18
Heptagon version 1.05.00

gdelaval:


kit-ty-kate:
What is the reason behind this hot fix ? Why not make a new release ?

gdelaval:
Sorry, I realized just a bit later that I took the wrong source archive (without the features announced for this new version).

kit-ty-kate:
Ok fair enough. Thanks!

PR #12218, opened 2018-06-19
added centos, rhel and alpine depexts to lz4

UnixJunkie:


UnixJunkie:
if I was not in a rush, I also think creating a conf-lz4 package would be the way to go.
But anyway, we can create the conf package out of this one, if we have added all systems we want.

kit-ty-kate:
Ok I'll merge this for now. Thanks!

PR #12221, opened 2018-06-19
Add a conf-liblz4 package and clean the lz4 package descriptions

kit-ty-kate:
cc @whitequark @UnixJunkie

### DIFF
--- a/packages/conf-liblz4/conf-liblz4.1/descr
+++ b/packages/conf-liblz4/conf-liblz4.1/descr
@@ -1,0 +1,1 @@
+Virtual package relying on liblz4 system installation.
--- a/packages/conf-liblz4/conf-liblz4.1/opam
+++ b/packages/conf-liblz4/conf-liblz4.1/opam
@@ -1,0 +1,18 @@
+opam-version: "1.2"
+maintainer: "https://github.com/ocaml/opam-repository/issues"
+homepage: "http://lz4.org"
+license: "GPLv2 & BSD-2-clause"
+build: ["pkg-config" "liblz4"]
+depends: [
+  "conf-pkg-config" {build}
+]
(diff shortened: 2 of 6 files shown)


kit-ty-kate:
Arf `liblz4-dev` on Ubuntu 14.04 doesn't have any pkg-config file.. :((

UnixJunkie:
You can probably get away by compiling a C program that includes lz4.h and that links against lz4.
If the program compiles, it means lz4-dev is present and usable on the system.

mseri:
Is it wort the hassle? Trusty end of suppport is in less than one year

UnixJunkie:
Is pkg-config really that portable?
Is it available on windows and BSDs?

mseri:
It is definitely available in OpenBSD and FreeBSD, I believe also in NetBSD. No idea about windows.

mseri:
At least since xenial, it contains the metadata for `pkg-config`:
```
$ apt-file list liblz4-dev
liblz4-dev: /usr/include/lz4.h
liblz4-dev: /usr/include/lz4frame.h
liblz4-dev: /usr/include/lz4hc.h
liblz4-dev: /usr/lib/x86_64-linux-gnu/liblz4.a
liblz4-dev: /usr/lib/x86_64-linux-gnu/liblz4.so
liblz4-dev: /usr/lib/x86_64-linux-gnu/pkgconfig/liblz4.pc
liblz4-dev: /usr/share/doc/liblz4-dev/changelog.Debian.gz
liblz4-dev: /usr/share/doc/liblz4-dev/copyright
```

I think we should merge this PR, especially given that in 6 months Trusty will no longer be supported.

hcarty:
Makes sense to me.  Thanks all!

PR #12222, opened 2018-06-19
add switches for 4.07.0+rc1

gasche:


kit-ty-kate:
Seems good. Could you also add the `force-safe-string` variant in another PR ? If you don't have time don't worry I can do it at some point today.

kit-ty-kate:
Thanks!

PR #12224, opened 2018-06-20
Update hiredis, resp-server. Add hiredis-value.

zshipko:


zshipko:
Sorry, it looks like I need to fix the tests for the resp-server release.

zshipko:
Okay, should be fixed now. It seems like the remaining errors are due to issues with 4.07 not being availabe for Alpine Linux yet.

kit-ty-kate:
The error on Alpine is unrelated don't worry.
Could you add `"-p" name` to you test commands ? It would avoid problems in the future if you, for instance, add tests for `hiredis-value`.

zshipko:
Yeah, no problem. I've add `"-p" name` for both `hiredis` and `resp-server`. Let me know if there's anything else that needs attention.

kit-ty-kate:
It's all good. Thanks a lot!

PR #12226, opened 2018-06-20
+mirage{-runtime,-types,-types-lwt}.3.1.0

hannesm:


hannesm:
the datakit-CI issue (can't find `alpine 3.7 ocaml`) seems to be unrelated -- I pressed the `rebuild` button several times, but the result is still the same.

hannesm:
good to merge imho

kit-ty-kate:
Looks good indeed. Thanks!!

PR #12236, opened 2018-06-21
Add the force-safe-string variant for 4.07.0

kit-ty-kate:
cc @gasche


gasche:
Thanks! I thought that I got all the `+beta2` switches (and only those), maybe I forgot it or there was not beta2+force-safe-string switch, but this is worth having in any case.

kit-ty-kate:
don't worry there wasn't. I added it in this PR.

PR #12238, opened 2018-06-21
fixed link error

dbousque:
Lymp 0.2.4 on opam is currently broken, due to a link error, a fix was made for this release, and so the checksum changed.


kit-ty-kate:
fair enough. Thanks!

PR #12241, opened 2018-06-23
caqti-driver-mariadb needs mariadb < 1.1.0.

paurkedal:
I need this version constraint for the Caqti MariaDB driver, since I managed to break revdep of my own package with andrenth/ocaml-mariadb#22.  (According to `opam search --depends-on mariadb` this is the only package which needs the constraint.)


kit-ty-kate:
Looks ok. Thanks a lot!

PR #12244, opened 2018-06-24
omake: patch bugfix from upstream

madroach:
fix: configuration of C compiler for the case that "ocamlc -config" prints several words for "c_compiler"


madroach:
Added a commit to add ndbm support to conf-dbm.
Does this need a separate pull-request?

PR #12245, opened 2018-06-25
add why3 1.0.0

silene:


mseri:
The debian failure is probably due to an out of date apt repository. I don’t think it can be solved by us, ping @kit-ty-kat

kit-ty-kate:
i don't know where the failures come from but it's definitely unrelated. Thanks.

mseri:
@kit-ty-kate usually you get that kind of error if you `apt-get install` something with an out of date local repository, and the fix is just to run `apt-get update` before the install

mseri:
mmh weird. Sorry, I had missed that in the logs from the ipad

PR #12250, opened 2018-06-26
Lwt 4.1.0: concurrency library, now under MIT license

aantron:
The main purpose of this release is that it is the first one under the MIT license.

From the changelog:

> Additions
>
> - Change license to MIT (ocsigen/lwt#560).
> - [`Lwt_fmt`](https://github.com/ocsigen/lwt/blob/33c37b550304b8d18b5dea4818fb1bfbc9a86aaf/src/unix/lwt_fmt.mli), an Lwt-aware version of [`Format`](http://caml.inria.fr/pub/docs/manual-ocaml/libref/Format.html) (ocsigen/lwt#548, Gabriel Radanne).
> - [`Lwt_io.establish_server_with_client_socket`](https://github.com/ocsigen/lwt/blob/33c37b550304b8d18b5dea4818fb1bfbc9a86aaf/src/unix/lwt_io.mli#L533) (ocsigen/lwt#586).
>
> Bugs fixed
>
> - [`Lwt_stream.iter_n`](https://github.com/ocsigen/lwt/blob/33c37b550304b8d18b5dea4818fb1bfbc9a86aaf/src/core/lwt_stream.mli#L314): rename `?max_threads` argument to `?max_concurrency` (ocsigen/lwt#581, Hezekiah Carty).
> - PPX: reject `match` expressions that have only `exception` cases (ocsigen/lwt#597, Raphaël Proust).
>
> Miscellaneous
>
> - Improvements to [`Lwt_pool`](https://github.com/ocsigen/lwt/blob/33c37b550304b8d18b5dea4818fb1bfbc9a86aaf/src/core/lwt_pool.mli) docs (ocsigen/lwt#575, Bobby Priambodo).
> - Restore all PPX tests (ocsigen/lwt#590, Jess Smith).

### DIFF
--- a/packages/amqp-client/amqp-client.1.0.1/opam
+++ b/packages/amqp-client/amqp-client.1.0.1/opam
@@ -24,4 +24,4 @@ conflicts: [
   "lwt" {< "2.4.6"}
   "async_unix" {>="v0.9"}
 ]
-available: [ ocaml-version >= "4.03.0" ]
+available: [ ocaml-version >= "4.03.0" & ocaml-version < "4.06.0" ]
--- a/packages/amqp-client/amqp-client.1.0.2/opam
+++ b/packages/amqp-client/amqp-client.1.0.2/opam
@@ -24,4 +24,4 @@ conflicts: [
   "lwt" {< "2.4.6"}
   "async" {>="v0.9"}
 ]
-available: [ ocaml-version >= "4.03.0" ]
+available: [ ocaml-version >= "4.03.0" & ocaml-version < "4.06.0" ]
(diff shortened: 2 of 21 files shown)


aantron:
I suggest merging this. There were no failures due to Lwt 4.1.0 in the original CI build, and the build is not running since I added the constraints, because some of the constrained packages have broken builds themselves.

We could split this up into two PRs to get a rebuild of Lwt against the new constraints, but I'd rather not do that. I think we will only find a few more packages that are not `-safe-string`-compatible, which can be done later by another package or a future release of Lwt.

avsm:
Agreed. Thanks for doing all the revdeps work!  By tomorrow, `opam ci status` should show any revdep or safe-string fallout with the results of this merged PR, which should help with subsequent PRs.

aantron:
Thanks!

PR #12253, opened 2018-06-26
Nocrypto: Fix patches for *BSD

copy:
Patches were created using `git format-patch --no-renames`. See https://github.com/ocaml/opam-repository/pull/12252.


copy:
It looks like the release tarball differs from the 0.5.4 branch in that `%%VERSION_NUM%%` was replaced by `0.5.4`. So the patches (which were created from the 0.5.4 branch) didn't apply to the release tarball. I have edited the patch files manually for now.

mseri:
Seems to have worked. Thanks

hannesm:
did anyone test this on either FreeBSD or OpenBSD?  At least on my FreeBSD-11 system, patch 0003 results in an empty `META` file which is then installed.

PR #12254, opened 2018-06-27
add virtual package for seq on OCaml >= 4.07

c-cube:


c-cube:
Well CI seems broken :disappointed:

packages/seq/seq.base/opam
```diff
@@ -0,0 +1,8 @@
+opam-version: "1.2"
+maintainer: " "
+authors: " "
+homepage: " "
+depends: ["ocamlfind" {build}]
```
hcarty: Out of curiosity - why does this require `ocamlfind`?

c-cube: Because that's the simplest way I know to install the `META` file at the proper place. I'm not on a crusade to kill ocamlfind.

hcarty: I was only asking because I don't see it used - does `opam-installer` use `ocamlfind` internally to install `META` from your `seq.install` file?

hcarty:
Given that this package is rather trivial and passes all the CI that is functional, I think it's ok to merge.

hcarty:
Thanks!

PR #12255, opened 2018-06-27
More lame depext.

toots:


toots:
Centos fails. The package isn't available in the main distribution but through additional sources. If possible, though, it would be great to include it in the depext nevertheless, that would help centos users a lot when trying to nagivate which package they need to install for each ocaml bindings.

toots:
Also, hopefully centos adds it at some point in the future `lame-devel` is already included in other RPM-based distros such as Fedora and etc.

toots:
Also, not sure why debian unstable fails. The package is clearly in the distro: https://packages.debian.org/sid/libmp3lame-dev

hcarty:
I think this is worth merging - it's beneficial for several OSs already.  Could the Debian Unstable failure be due to some transient packaging version conflict?

toots:
Thanks!

toots:
Not sure about debian unstable, we'll see if that gets better.

PR #12257, opened 2018-06-28
New package version why 2.41

claudemarche:


claudemarche:
Any reason why not accepting this pull request?

hcarty:
@claudemarche Can you look at the Travis failures?  It looks like why3 is failing to build there.

Beyond that I think this is ok to merge.

claudemarche:
Indeed, all the errors I see come from other packages than the why package, so I don't see how to do better.

hcarty:
Thanks!

PR #12259, opened 2018-06-29
Package Coq 8.8.1

RalfJung:


hcarty:
@RalfJung Is the sibylfs-lem failure related to coq?  It looks like it may need a version constraint.

RalfJung:
I have no idea what sibylfs-lem is... The coq dependencies and constraints did not change at all since 8.8.0.

The [sibylfs-lem](https://github.com/ocaml/opam-repository/blob/master/packages/sibylfs-lem/sibylfs-lem.0.4.0/opam) opam file just says it conflicts with any Coq >= 8.5. I have no idea why its build fails. The error is
```
# File "ulib/batText.ml", line 237, characters 20-21:
# Error: This expression has type bytes but an expression was expected of type
#          UTF8.t = string
```
so it looks to me like sibylfs-lem is just broken?

avsm:
It's broken with ocaml 4.06 due to the switch to safe-string -- it can be ignored for the purpose of this PR

hcarty:
I think this is good to merge then.  Thank you @avsm and @RalfJung!

RalfJung:
Yeah I was about to report that I reproduced the sibylfs-lem failure in a 4.06 switch without Coq being involved at all.

PR #12261, opened 2018-06-30
Plotkicad v0.3.0

jnavila:
adding also kicadsch v0.3.0


hcarty:
@jnavila Any idea what is causing the Travis failure on OCaml 4.04?

jnavila:
@hcarty I have strictly no idea why it fails. It seems that the version of lwt which is pulled for ocaml 4.04 has some issues with the ppx_lwt extension. Earlier versions are disabled by a limit on ocaml version, but it used to work with 4.04. It's difficult to know when we are not future-proof, specially with the evolutions of dependencies. I guess it's just a matter of taking a picture of the versions of dependencies at the time the new version of the package was introduced and passed the tests.

hcarty:
Thanks!

PR #12262, opened 2018-06-30
Lens version 1.2.1 (using jbuilder)

pdonadeo:


hcarty:
@pdonadeo Could you address the opam-lint error?  Then this should be good to merge.

pdonadeo:
@hcarty of course yes. I saw the error but I had no time to correct immediately.

hcarty:
Thanks!

PR #12263, opened 2018-07-01
More depext for fdk-aac.

toots:
Same as #12255, depext for centos isn't in the main distribution but would be very useful to have mentioned for end-user.


hcarty:
Thanks!

PR #12278, opened 2018-07-06
+checkseum

dinosaure:


packages/checkseum/checkseum.0.0.1/opam
```diff
@@ -0,0 +1,21 @@
...
+dev-repo:     "https://github.com/dinosaure/checkseum.git"
+doc:          "https://dinosaure.github.io/checkseum/"
+license:      "MIT"
+
+build: [ "jbuilder" "build" "-p" name "-j" jobs ]
+
+depends: [
+  "ocamlbuild"     {build}
+  "ocamlfind"      {build}
+  "topkg"          {build}
```
kit-ty-kate: mmmh, all the dependencies above doesn't seem to match how the package is built

kit-ty-kate:
All good. Thanks a lot!

PR #12279, opened 2018-07-06
Root 'jbuilder subst' calls

samoht:
This ensure that jbuilder doesn't try to escape the build directory.
See https://github.com/ocaml/dune/issues/954 for some context


samoht:
Note: this is just an incomplete change (just a quick sed), I plan to submit fixes to more packages later.

samoht:
Ok I think I've replaced all occurrences of `jbuilder subst` now...

kit-ty-kate:
Seems ok. A lot of maintainers need to be notified but that can be done later (poke @hannesm)

Thanks a lot!

PR #12280, opened 2018-07-06
Release frama-c.20180502

maroneze:
Minor (bugfix) release for Frama-C Chlorine.

(note: the .tar.gz is maybe not yet available in our website, so the build may fail, in which case @dbuhle will retry once the file is uploaded)

### DIFF
--- a/packages/frama-c/frama-c.20180502/descr
+++ b/packages/frama-c/frama-c.20180502/descr
@@ -1,0 +1,15 @@
+Platform dedicated to the analysis of source code written in C.
+
+Frama-C gathers several analysis techniques in a single collaborative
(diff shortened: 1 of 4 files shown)


kit-ty-kate:
Ok, could you please notify us when the tarball is ready?

maroneze:
Hello, the tarball is ready now and I updated the hash to it. There are just a few Debian Unstable builds that failed, I believe they are unrelated to this PR.

kit-ty-kate:
Looks good. Thanks a lot! You could've avoided the `run_autoconf_if_needed.sh` script by using the `{pinned}` attribute but for now I guess that's good enough. It can be removed later in a following PR if you want to.

PR #12281, opened 2018-07-06
Update md5 of xtmpl 0.17.0 archive

zoggy:
META file in previous archive contained a bad version number.


kit-ty-kate:
Could you have a look at the revdeps? It seems that they are mostly your packages and that they are broken in 4.06 / Lwt >= 4.0 (or 3.2 I don't remember)

Otherwise seems ok. Thanks!

PR #12284, opened 2018-07-06
mirrors the FrontC package

ivg:
The upstream package is no longer available, so as a temporary measure I propose to mirror it (thanks @pmetzger for the archive). We have contacted the upstream author, and if he won't reply or if he will refuse to fix, then we will put the package on a more reliable media (e.g., github).


pmetzger:
We seem to have some failures, but this doesn't make the situation _worse_ than it was...

mseri:
I agree. They seem to be due to mising safe string support but this could be modified in a separate PR if there is the need

PR #12286, opened 2018-07-06
nocrypto.0.5.4-1 still does not work with BSD patch

hannesm:
At least on my FreeBSD-11 system, patch `0003` results in an empty `META` file which is then installed - leading to an unusable nocrypto.  Would be nice if next time someone "unbreaks" this on Free/OpenBSD, they'd _test_ their changes.

see #12062 #12252 #12253
//cc @copy @hcarty @mseri


copy:
Sorry! I should have pinged someone for testing on the BSDs before getting another non-applying patch merged.

Considering how different BSD and GNU `apply` are, it's probably better if this was fixed by using `git apply` (https://github.com/ocaml/opam/issues/2160).

PR #12287, opened 2018-07-06
Fix craml files

samoht:


hcarty:
Thanks for the cleanup - I should have noticed that during the PR review.

PR #12291, opened 2018-07-07
domain-name.0.1.1

hannesm:
## Domain-name - [RFC 1035](https://tools.ietf.org/html/rfc1035) Internet domain names

A domain name is a sequence of labels separated by dots, such as `foo.example`. Each label may contain any bytes. The length of each label may not exceed 63 charactes.  The total length of a domain name is limited to 253 (byte representation is 255), but other protocols (such as SMTP) may apply even smaller limits.  A domain name label is case preserving, comparison is done in a case insensitive manner.

## Changes in 0.1.1 
* `to_string` and `to_strings` now have an optional labeled `trailing` argument of type bool
* support for FQDN with trailing dot: `of_string "example.com."` now returns `Ok`, and is equal to `of_string "example.com"`
* fix and add tests for `drop_labels` and `drop_labels_exn`, where the semantics of the labeled `back` argument was inversed.

### DIFF
--- a/packages/domain-name/domain-name.0.1.1/descr
+++ b/packages/domain-name/domain-name.0.1.1/descr
@@ -1,0 +1,12 @@
+[RFC 1035](https://tools.ietf.org/html/rfc1035) Internet domain names
+
+
+A domain name is a sequence of labels separated by dots, such as `foo.example`.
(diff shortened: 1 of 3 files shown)


kit-ty-kate:
Looks good. Thanks!

PR #12293, opened 2018-07-07
Add ffmpeg.0.1.2

kit-ty-kate:
Has been released two years ago but was missing in opam and the previous version (0.1.1) is unusable.

cc @smimram @toots


kit-ty-kate:
There doesn't seem to be any package in EPEL for CentOS but somehow in another repository. I think this is fine as is.

PR #12295, opened 2018-07-08
lzo 0.0.2

ygrek:


ygrek:
OSX build fails because of missing depext

kit-ty-kate:
Could you add the missing depexts?
Here is a table to search for your package: https://github.com/ocaml/opam-depext/blob/master/README.md

For instance it'll be `lzo-dev` under Alpine, `lzo-devel` under CentOS and `lzo` for Homebrew/OSX

ygrek:
> Could you add the missing depexts?

ftr I consider this requirement an abuse of my time

PR #12296, opened 2018-07-08
geoip 0.0.3

ygrek:


ygrek:
OSX build fails because of missing depext

kit-ty-kate:
That should fix the OSX build. Could you put those two changes upstream?

kit-ty-kate:
Also, same answer as in #12295

PR #12300, opened 2018-07-08
cbor 0.2

ygrek:


ygrek:
CI failure caused by : 
```
W: GPG error: http://ppa.launchpad.net/avsm/ocaml42+opam12/ubuntu trusty Release: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 5B2D0C5561707B09
[...]
WARNING: The following packages cannot be authenticated!
  ocaml-base-nox ocaml-compiler-libs ocaml-interp ocaml-nox camlp4
  camlp4-extra ocaml-native-compilers opam-docs opam
E: There were unauthenticated packages and -y was used without --allow-unauthenticated
```

kit-ty-kate:
All green now. Thanks!

PR #12301, opened 2018-07-08
gperftools 0.3

ygrek:


ygrek:
CI fails on OSX because of missing depext

kit-ty-kate:
Same comment as in #12295

ygrek:
there is no gperftools for alpine

kit-ty-kate:
Thanks!

PR #12306, opened 2018-07-08
[new release] ppx_tools_versioned.5.2

let-def:
Port to jbuilder by @rgrinberg.

### DIFF
--- a/packages/ppx_tools_versioned/ppx_tools_versioned.5.2/opam
+++ b/packages/ppx_tools_versioned/ppx_tools_versioned.5.2/opam
@@ -1,0 +1,21 @@
+opam-version: "1.2"
+maintainer: "frederic.bour@lakaban.net"
+authors: [
+  "Frédéric Bour <frederic.bour@lakaban.net>"
+  "Alain Frisch <alain.frisch@lexifi.com>"
+]
+license: "MIT"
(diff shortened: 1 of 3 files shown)


packages/ppx_tools_versioned/ppx_tools_versioned.5.2/opam
```diff
@@ -0,0 +1,19 @@
...
+homepage: "https://github.com/let-def/ppx_tools_versioned"
+bug-reports: "https://github.com/let-def/ppx_tools_versioned/issues"
+license: "MIT"
+tags: "syntax"
+dev-repo: "git://github.com/let-def/ppx_tools_versioned.git"
+build: [make "all"]
+install: [make "install"]
+remove: ["ocamlfind" "remove" "ppx_tools_versioned"]
+depends: [
+  "ocamlfind" {>= "1.5.0"}
```
kit-ty-kate: Since it has been switched to `jbuilder` I suppose those lines should change now? (starting from the `build` instruction)

let-def: Thanks, my bad!

kit-ty-kate:
Thanks a lot!

Issue #12302, opened 2018-07-08
ocamlfind-1.8.0 download link down


vassilmladenov:
It appears that `camlcity.org` is down right now, and the package's two sources are 
- Main: http://download.camlcity.org/download/findlib-1.8.0.tar.gz
- Mirror: http://download2.camlcity.org/download/findlib-1.8.0.tar.gz

I'm on the 2.0.0 branch but the urls are the same in master. Might be worth mirroring on a different server if the author and/or maintainer are OK with it, as it is a dependency of a great many packages in the tree.

PR #12311, opened 2018-07-08
Add fedora mirror for ocamlfind

vassilmladenov:
Per #12302, this PR allows a fallback to install ocamlfind 1.8.0. The Fedora project mirrors this package at https://src.fedoraproject.org/lookaside/pkgs/ocaml-findlib/, so if this is acceptable then I can go back and replace the mirrors for the previous versions as well.

### DIFF
--- a/packages/ocamlfind/ocamlfind.1.8.0/opam
+++ b/packages/ocamlfind/ocamlfind.1.8.0/opam
@@ -60,5 +60,5 @@ extra-files: [
 url {
   src: "http://download.camlcity.org/download/findlib-1.8.0.tar.gz"
   checksum: "md5=a710c559667672077a93d34eb6a42e5b"
-  mirrors: "http://download2.camlcity.org/download/findlib-1.8.0.tar.gz"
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Can't we keep the previous mirror as well ? I think `mirrors` can take a list but I'm not sure

vassilmladenov:
I searched the master branch, but I can't find an example of multiple mirrors for a package. Each is a singleton list, and I think the list was just dropped in the 2.0.0 branch. Otherwise, I'm not sure why the list brackets are present in master but not on 2.0.0. Can you elaborate on how the 2.0.0 branch is synchronized?

kit-ty-kate:
Oops sorry I forgot to answer you here. Hmm, well, you can have [this](https://github.com/ocaml/opam-repository/blob/master/packages/bap-beagle/bap-beagle.1.4.0/url) in the master branch and the 2.0.0 branch is synchronized with the master branch using the translator from 1.2 to 2.0 format included in opam2. This is done automatically for each commits on master.
For now there is nowhere in the opam-repository where there are a list of mirrors larger than one so I guess the translator just get ride of the unnecessary list.

PR #12313, opened 2018-07-09
+decompress.0.8

dinosaure:


dinosaure:
ping @rlepigre, I constrained `imagelib` to use only `decompress.0.7`, It's easy to move on the new interface 👍 .

dinosaure:
@rlepigre 

An update to the new interface should close to: https://github.com/mirage/ocaml-git/pull/303
However, you need to choose an implementation of `checkseum` (C or OCaml) and put it at the beginning of your dependencies,

kit-ty-kate:
camelus failure is unrelated. Thanks a lot! (could you use the more conventional method for `build` and `build-test` next time if that's ok?

PR #12316, opened 2018-07-09
+mirage-flow.1.5.0 with safe-string support

avsm:
Signed-off-by: Anil Madhavapeddy <anil@recoil.org>


kit-ty-kate:
Could you add `"-p" name` to the jbuilder subst commands? See https://github.com/ocaml/opam-repository/pull/12279

avsm:
fixed!

kit-ty-kate:
Looks good. Thanks!

PR #12326, opened 2018-07-11
Fix conf-mpi on macOS and become maintainer

tbrk:
Agreement with @gasche by private mail.


gasche:
I re-ran the OSX test, which suffered a transient download failure. Alpine fails because it is a very dependency-restricted environment that cannot set up most of the depexts, I think -- the other distributions in the CI install them automatically.

PR #12331, opened 2018-07-11
Release stdcompat.5

thierry-martinez:
- Interfaces are auto-generated.

- stdcompat is now free from any required dependency. There are still
  optional dependencies with respect to the packages result, seq and
  uchar: stdcompat takes care of providing types compatible with
  these packages if they are installed.

- Preprocessing is performed by "./configure" script (generated by
  autoconf). cppo and the C preprocessor are no longer used.

- Makefile is generated by automake. ocamlfind is no longer required
  for installation.

- Split implementation into one module for each standard library module
  (suggested by Yotam Barnoy:
   https://github.com/thierry-martinez/stdcompat/issues/4)

- All modules are now exported as sub-modules of Stdlib module
  (as in OCaml 4.07) -- Bigarray is not exported to allow programs not
  to be compiled with this module, this may change in the future.
  (suggested by Yotam Barnoy:
   https://github.com/thierry-martinez/stdcompat/issues/4)

- Compatibility with uchar package

### DIFF
--- a/packages/stdcompat/stdcompat.5/descr
+++ b/packages/stdcompat/stdcompat.5/descr
@@ -1,0 +1,1 @@
+Compatibility module for OCaml standard library allowing programs to use some recent additions to the OCaml standard library while preserving the ability to be compiled on former versions of OCaml.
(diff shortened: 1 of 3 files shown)


mseri:
Thanks

PR #12336, opened 2018-07-11
fix typo in url for lablgtk.8.06.5

garrigue:
Since this is a 4.07 only package, the checks didn't run on the previous PR.
This fixes the url. The md5 is correct too.


kit-ty-kate:
Thanks!

PR #12337, opened 2018-07-12
added packages/parmap/parmap.1.0-rc10

UnixJunkie:
bugfix in Parmap.array_parmapi
(i was very often wrong)

### DIFF
--- a/packages/parmap/parmap.1.0-rc10/opam
+++ b/packages/parmap/parmap.1.0-rc10/opam
@@ -1,0 +1,29 @@
+opam-version: "1.2"
+maintainer: "Roberto Di Cosmo <roberto@dicosmo.org>"
+authors: "Roberto Di Cosmo <roberto@dicosmo.org>"
+homepage: "https://github.com/rdicosmo/parmap"
+dev-repo: "https://github.com/rdicosmo/parmap.git"
+bug-reports: "https://github.com/rdicosmo/parmap/issues"
+build: [
+  ["aclocal" "-I" "m4"]
+  ["autoconf"]
+  ["autoheader"]
+  ["./configure"] 
+  [make "DESTDIR=%{prefix}%" "OCAMLLIBDIR=lib" ]
+]
+install: [
+  [make "install" "DESTDIR=%{prefix}%" "OCAMLLIBDIR=lib"]
+]
+remove: [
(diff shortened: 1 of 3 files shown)


packages/parmap/parmap.1.0-rc10/opam
```diff
@@ -0,0 +1,29 @@
...
+]
+remove: [
+  ["aclocal" "-I" "m4"]
+  ["autoconf"]
+  ["autoheader"]
+  ["./configure"] 
+  [make "uninstall" "DESTDIR=%{prefix}%" "OCAMLLIBDIR=lib"]
+]
+depends: [
+  "ocamlfind"
```
kit-ty-kate: `Findlib` doesn't seem to be used here. I'd suggest to add a `{build}` attribute. (same for `conf-autoconf` & `conf-aclocal`)

Also is there any ways to avoid using autoconf/aclocal? It looks like the archive linked in the github release not is different from the one you use here and already have autoconf/... ran. See https://github.com/rdicosmo/parmap/releases/tag/1.0-rc10

kit-ty-kate: Mmh nope my bad, it's not different in fact but the autoconf & cie. doesn't seem necessary since `./configure` is already present anyway.

UnixJunkie:
I added the {build} thing to all deps.
But, I'm not sure it is wise to remove the configure generation step.
@rdicosmo should know.

UnixJunkie:
FYI ocamlfind is used by 'make uninstall' (which issues an 'ocamlfind remove parmap').
I hope it qualifies as a {build} dependency, but I'm not sure.

kit-ty-kate:
I did a diff between the two and I could see no difference after `make`:
```
Only in parmap: aclocal.m4
Only in parmap: autom4te.cache
Files parmap/_build/_digests and parmap-1.0-rc10/_build/_digests differ
Files parmap/config.log and parmap-1.0-rc10/config.log differ
Files parmap/config.status and parmap-1.0-rc10/config.status differ
Files parmap/configure and parmap-1.0-rc10/configure differ
Only in parmap: .git
```
I think it's safe to remove

kit-ty-kate:
Thanks a lot!

PR #12330, opened 2018-07-11
Camlp4 for OCaml 4.07

diml:

PR #12338, opened 2018-07-12
camlp4.4.07+1 remove applied patch

andyli:
The patch was applied in https://github.com/ocaml/camlp4/commit/f0ea53725465260556832398096cef8d3f20b49d.

It fixes the error:
```
#=== ERROR while installing camlp4.4.07+1 =====================================#
These patches didn't apply at /home/andy/.opam/4.07.0/build/camlp4.4.07+1:
  - f0ea53725465260556832398096cef8d3f20b49d.patch
```

The #12330 CI was green because ocaml 4.07 was not tested...


kit-ty-kate:
Thanks!

PR #12340, opened 2018-07-12
+httpaf.0.4.1

seliopou:


packages/httpaf-async/httpaf-async.0.4.1/opam
```diff
@@ -0,0 +1,23 @@
+opam-version: "1.2"
+name: "httpaf-async"
+maintainer: "Spiros Eliopoulos <spiros@inhabitedtype.com>"
+authors: [ "Spiros Eliopoulos <spiros@inhabitedtype.com>" ]
+license: "BSD-3-clause"
+homepage: "https://github.com/inhabitedtype/httpaf"
+bug-reports: "https://github.com/inhabitedtype/httpaf/issues"
+dev-repo: "https://github.com/inhabitedtype/httpaf.git"
+build: [
+  ["jbuilder" "subst"] {pinned}
```
kit-ty-kate: Could you either add `"-p" name` to this command or use `dune` directly (replacing every mentions or `jbuilder` by `dune` in this file should work I guess). See https://github.com/ocaml/opam-repository/pull/12279 to know more about the original problem.

kit-ty-kate:
Could you take the change back to the source repository?

Thanks a lot!

PR #12341, opened 2018-07-12
Sundials v3.1.1p0 release

tbrk:
New release of Sundials. Update to install package.

### DIFF
--- a/packages/sundialsml/sundialsml.3.1.1p0/descr
+++ b/packages/sundialsml/sundialsml.3.1.1p0/descr
@@ -1,0 +1,31 @@
+Interface to the Sundials suite of numerical solvers
+
+Sundials is a collection of six numerical solvers: CVODE, CVODES, IDA, IDAS,
+ARKODE, and KINSOL. This interface provides access to all features of the
+underlying library except the Hypre and PETSC nvectors.
+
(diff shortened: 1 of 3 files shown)


tbrk:
The Ubuntu LTS and Alpine builds fail because the CI does not seem to be able to install the `depexts`.
I have to adapt the Sundials/ML configure script for CentOS.
Debian testing and unstable fail because the `depexts` are different than those for stable; is it possible to give different depexts for different versions of debian in the `opam` files (I read some of the issue reports for opam but without really understanding what I should do).

tbrk:
OK. *CentOS* is fixed.

Without some kind of versioning for `depexts`, it's not possible to simultaneously support *Debian* *stable*, *testing*, and *unstable* because the required package names change.

I don't think I can fix the *Alpine* and *Ubuntu LTS* CI dependency problems.

kit-ty-kate:
Yeah, that's a bit annoying but oh well. Hopefully it'll get better once opam2 will be released.
Could you add the `{build}` attribute to `ocamlfind` though? I'll merge it afterwards.

tbrk:
*{build}* added. Thanks for spotting that.

kit-ty-kate:
Thanks!

PR #12343, opened 2018-07-13
Release ppx_bigarray v3.0.0

akabe:
Release note: https://github.com/akabe/ppx_bigarray/releases/tag/v3.0.0

### DIFF
--- a/packages/ppx_bigarray/ppx_bigarray.3.0.0/descr
+++ b/packages/ppx_bigarray/ppx_bigarray.3.0.0/descr
@@ -1,0 +1,1 @@
+A PPX extension for big array literals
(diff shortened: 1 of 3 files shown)


akabe:
Hmm. I don't know why this is broken.

kit-ty-kate:
It's probably due to the fact that Camelus doesn't merge automatically the master branch with yours and directly test with your branch instead. Yours seems to be not up-to-date (i.e. does not contain `dune`) so it fails because your package relies to `dune` which does not exist on your version.

That said, everything else seems fine so merging now. Thanks a lot!

akabe:
I understood. Thank you!

PR #12346, opened 2018-07-13
Fix dependency of js_of_ocaml-lwt

AltGr:
it doesn't compile with js_of_ocaml 3.1.0.
Also removed redundant conflict


kit-ty-kate:
Thanks!!

PR #12347, opened 2018-07-13
Update some Jane Street libraries to support OCaml 4.07

xclerc:


hhugo:
Fix https://github.com/janestreet/core_kernel/issues/84

PR #12350, opened 2018-07-14
Fix a race condition between jbuilder, dune and ocaml-migrate-parsetree

diml:
Fix the issue described [here](https://discuss.ocaml.org/t/is-it-currently-possible-to-use-ppx-deriving-in-a-jbuilder-project/426/11). In summary, when ocaml-migrate-parsetree is built and installed by dune, it will copy one more information to the installation directory compared to when it is built and installed by jbuilder. This is not an issue for existing jbuilder projects but it is for newer pure dune projects. A few users have been bitten by this issue.

To avoid this issue, this PR makes `dune` an explicit dependency of ocaml-migrate-parsetree. This should ensure that omp is always built and installed with dune. Given that all versions of omp are fully backward compatible and that some older packages might depend on omp and have a upper bound on jbuilder, I only made this change for the latest version of omp and marked older versions as conflicting with dune.

I did the same for ppxlib which has the same issue but made both 0.2.1 and 0.3.0 depend on dune since 0.3.0 has breaking changes compared to 0.2.1.

### DIFF
--- a/packages/ocaml-migrate-parsetree/ocaml-migrate-parsetree.0.6/opam
+++ b/packages/ocaml-migrate-parsetree/ocaml-migrate-parsetree.0.6/opam
@@ -26,3 +26,4 @@ depends: [
   "jbuilder" {build & >= "1.0+beta2"}
 ]
 available: [ocaml-version >= "4.02.0" & ocaml-version < "4.06.0"]
+conflicts: [ "dune" ]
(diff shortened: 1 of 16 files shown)


kit-ty-kate:
Looks fine. Since most failures seem related to a server reboot and this PR only changes a few part I'm going to merge this right away.

Thanks a lot!

PR #12360, opened 2018-07-16
Fix jbuilder's post-messages field

diml:
@AltGr is it better like this?


kit-ty-kate:
Looks good, thanks!

AltGr:
Yup, thanks! :)

PR #12361, opened 2018-07-16
Package uwt.0.3.2

fdopen:
### `uwt.0.3.2`

libuv bindings

uwt provides OCaml bindings for libuv.
The main loop of libuv is integrated into lwt,
the light-weight cooperative threads library.

---
* Homepage: https://github.com/fdopen/uwt
* Source repo: https://github.com/fdopen/uwt.git
* Bug tracker: https://github.com/fdopen/uwt/issues

---

### DIFF
--- a/packages/uwt/uwt.0.3.2/descr
+++ b/packages/uwt/uwt.0.3.2/descr
@@ -1,0 +1,5 @@
+libuv bindings
+
+uwt provides OCaml bindings for libuv.
+The main loop of libuv is integrated into lwt,
+the light-weight cooperative threads library.
--- a/packages/uwt/uwt.0.3.2/opam
+++ b/packages/uwt/uwt.0.3.2/opam
@@ -1,0 +1,37 @@
+opam-version: "1.2"
+maintainer: "andreashauptmann@t-online.de"
+authors: [ "andreashauptmann@t-online.de" ]
+license: "MIT"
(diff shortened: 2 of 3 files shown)


kit-ty-kate:
There is only one (unrelated) failure (cc @dra27)
```
  - uwt -> ppx_import >= 1.0
  - ppx_import>=1.0 is not available because your system doesn't comply with opam-version >= "1.2.2" & ocaml-version >= "4.02.0" & ocaml-version < "4.07.0".
```
I'm not sure why there is this error on 4.07 only though

Anyway, merging. Thanks!

dra27:
What's the policy on what the opam file itself should have when a dependency isn't available - this package is effectively not available for OCaml 4.07.0, although presumably the second ppx_import supports 4.07.0 it would then work, so being forced to restrict this package seems artificial?

fdopen:
The regulare dependencies are available, but the test dependencies are missing for OCaml 4.07. The travis script can't deal with it.
It checks for normal availability - without test dependencies:
https://github.com/ocaml/opam-repository/blob/e7ce7dbcd7e37ce4961029733501a36fed9a4a4d/.travis-ci.sh#L98
But then tries to build and run the tests as well:
https://github.com/ocaml/opam-repository/blob/e7ce7dbcd7e37ce4961029733501a36fed9a4a4d/.travis-ci.sh#L121

PR #12362, opened 2018-07-17
lablqml.0.6: switching to dune

Kakadu:


packages/lablqml/lablqml.0.6/opam
```diff
@@ -0,0 +1,28 @@
...
+
+build: [
+  ["./configure"]
+  [make]
+]
+install: [make "install"]
+remove: [make "uninstall"]
+flags: [ light-uninstall ]
+depends: [
+  "ocamlfind"
```
kit-ty-kate: Do you use the Findlib library ? I couldn't see any reference of this in the source code. If so could you flag this dependency with the `{build}` attribute?

kit-ty-kate:
Also there is a failure on Ubuntu 16.04 which seems related:
```
# startup_stubs.cpp: In function 'value caml_QGuiApplication_exec(value)':
# startup_stubs.cpp:198:19: error: 'nullptr' was not declared in this scope
#    Q_ASSERT(app != nullptr);
```
Something tells me you need to explicitly require for cpp11 or something

Kakadu:
@kit-ty-kate I think that it's fine now

kit-ty-kate:
Thanks!

PR #12364, opened 2018-07-18
+faraday.0.6.1

seliopou:


kit-ty-kate:
Could you add the `"-p" name` to the `jbuilder subst` commands? Same change & reason as in #12340

kit-ty-kate:
Thanks!

PR #12366, opened 2018-07-18
Use https for the compiler archive sources.

andyli:
I've verified the updated links still works.

There are 2 remaining links which don't support https:
 * https://github.com/ocaml/opam-repository/blob/21aad325af7a100424609557443620f3164f6433/compilers/4.06.1/4.06.1%2Btermux/4.06.1%2Btermux.comp#L17
 * https://github.com/ocaml/opam-repository/blob/21aad325af7a100424609557443620f3164f6433/compilers/4.01.0/4.01.0%2BBER.comp#L5


mseri:
It seems to work for me locally

kit-ty-kate:
There is no way of testing this with the current CIs but it looks good. Thanks a lot!

PR #12374, opened 2018-07-20
add gnomecanvas depext for Archlinux

vprevosto:
Everything is in the title: this one-liner allows depext to install libgnomecanvas on Arch as needed.


kit-ty-kate:
Looks good. Thanks!

PR #12375, opened 2018-07-20
Package ocaml-varint.1.0

XnuKernPoll:
### `ocaml-varint.1.0`

A simple VarInt implementation.

---
* Homepage: https://github.com/XnuKernPoll/ocaml-varint
* Source repo: http://github.com/XnuKernPoll/ocaml-varint.git
* Bug tracker: https://github.com/XnuKernPoll/ocaml-varint/issues

---
### opam-lint failures

---

### DIFF
--- a/packages/ocaml-varint/ocaml-varint.1.0/descr
+++ b/packages/ocaml-varint/ocaml-varint.1.0/descr
@@ -1,0 +1,1 @@
+A simple VarInt implementation.
--- a/packages/ocaml-varint/ocaml-varint.1.0/opam
+++ b/packages/ocaml-varint/ocaml-varint.1.0/opam
@@ -1,0 +1,19 @@
+opam-version: "1.2"
+maintainer: "flatmapds@gmail.com"
+authors: "Sam Riyad"
+homepage: "https://github.com/XnuKernPoll/ocaml-varint"
+          
+license: "MIT"
+tags: ["mirage" "encoding"]
+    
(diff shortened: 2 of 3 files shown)


packages/ocaml-varint/ocaml-varint.1.0/descr
```diff
@@ -0,0 +1 @@
+A simple VarInt implementation.
```
pmetzger: I note there is no \n at the end of this line.

pmetzger:
Just an aside: when I see interesting package names go by, I often try to look at the package and see if it is interesting to me. The entire documentation (including in your README and your mli files) for this package is

> A simple VarInt implementation.

which makes it hard for potential users to figure out that they might want to use your code.

hannesm:
This may be not well documented (it is in https://github.com/ocaml/opam-repository/wiki/PR-checks), in opam-repository we try to avoid `ocaml-` prefixes and `-ocaml` suffixes for package names.  Would you be so kind to rename your package to `varint` (and I agree with @pmetzger that varint should be a bit better documented).  Thanks for your contribution!

XnuKernPoll:
Alright, I'll fix the package descr, change the name, and add a tutorial in the read me, and add some comments in the mli files.

XnuKernPoll:
I added the changes you guys gave me, fixed the api, and added Length Prefixing
https://github.com/ocaml/opam-repository/pull/12377

Out of curiosity do append, concat, or sub copy, because I know that bigarray can use layout transformation to avoid copies.

PR #12376, opened 2018-07-20
[compilers] add 4.06.1+no-naked-pointers+flambda

jvillard:
This new switch turns on both `-no-naked-pointers` and `-flambda` to get more performance out of the compiler.

There's currently no opam switch to get a compiler with the `-no-naked-pointers` option. Compiling with `-no-naked-pointers` gives good results on infer for example, with as much as 4% CPU win on some workloads.


kit-ty-kate:
Thanks!

PR #12377, opened 2018-07-20
Package varint.1.0

XnuKernPoll:
### `varint.1.0`

A simple varint implementation modeled after the one found in Go's standard library.

Originally I wrote it because I wanted to implement a protocol in ocaml that used it, but did not want to use piqi, or protobuf. 

What varint encoding does is that you can input an int32 or int64, and for smaller numbers, it will take up less space, protobuf uses this technique for field length prefixing,  as a result it is more space efficient than using an 32 bit or 64 bit int, but on the other hand it does take more CPU time.

---
* Homepage: https://github.com/XnuKernPoll/ocaml-varint
* Source repo: https://github.com:XnuKernPoll/ocaml-varint.git
* Bug tracker: https://github.com/XnuKernPoll/ocaml-varint/issues

---

### DIFF
--- a/packages/varint/varint.1.0/opam
+++ b/packages/varint/varint.1.0/opam
@@ -1,0 +1,21 @@
+opam-version: "1.2"
+maintainer: "flatmapds@gmail.com"
+authors: ["Sam Riyad"]
+tags: ["mirage" "encoding"]
+license: "MIT"
+         
+homepage: "https://github.com/XnuKernPoll/ocaml-varint"
+dev-repo: "https://github.com:XnuKernPoll/ocaml-varint.git"
+bug-reports: "https://github.com/XnuKernPoll/ocaml-varint/issues"
+             
+          
+build: [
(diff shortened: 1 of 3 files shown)


packages/varint/varint.1.0/opam
```diff
@@ -0,0 +1,21 @@
+opam-version: "1.2"
+maintainer: "flatmapds@gmail.com"
+authors: ["Sam Riyad"]
+tags: ["mirage" "encoding"]
+license: "MIT"
+         
+homepage: "https://github.com/XnuKernPoll/ocaml-varint"
+dev-repo: "https://github.com:XnuKernPoll/ocaml-varint.git"
```
kit-ty-kate: This syntax doesn't seem to be correct
```
fatal: unable to access 'https://github.com:XnuKernPoll/ocaml-varint.git/': Port number ended with 'X'
```

PR #12365, opened 2018-07-18
Cpdf unsafe string fix

UnixJunkie:

PR #12379, opened 2018-07-23
added packages/camltc/camltc.0.9.7.0

UnixJunkie:
this one tries to get rid of the git submodule so that the build of the package is more reproducible

### DIFF
--- a/packages/camltc/camltc.0.9.7.0/descr
+++ b/packages/camltc/camltc.0.9.7.0/descr
@@ -1,0 +1,1 @@
+OCaml bindings for tokyo cabinet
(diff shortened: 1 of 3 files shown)


UnixJunkie:
probably it is for tests only, I made the change

kit-ty-kate:
It apparently is not :/

Also, please can you make the requested change https://github.com/ocaml/opam-repository/pull/12365#issuecomment-407030724 ?! Not that I'm getting a bit pissed but I am... 🙃

UnixJunkie:
I removed the test flag for ounit.

kit-ty-kate:
Thanks!

PR #12367, opened 2018-07-18
qocamlbrowser 0.2.10

Kakadu:

PR #12384, opened 2018-07-23
qocamlbrowser.0.2.10: update opam bound

mseri:
This fixes a too strict bound merged in https://github.com/ocaml/opam-repository/pull/12367. Thanks @kit-ty-kate for spotting it.


mseri:
Oh I had forgotten about that.

You are right, it seems that the fix is backward compatible, but I think you made a reasonable guess there.

kit-ty-kate:
Looks good. Thanks!

PR #12389, opened 2018-07-24
[new release] digestif (v0.6.1)

dinosaure:
CHANGES:

- *breaking change* API: Digestif implements a true linking trick. End-user need
  to explicitely link with `digestif.{c,ocaml}` and it needs to be the first of
  your dependencies.
- move to `jbuilder`/`dune`


kit-ty-kate:
Looks good. Thanks!

PR #12390, opened 2018-07-24
opam 2.0.0~rc4 libs and devel packages

AltGr:
I was unsure wether to name this 2.0.0 or 2.0.0~rc4, since they are expected to be equal, and renaming the package would cause unnecessarily recompilations. But since the upstream archives are called 'rc4' at this point, it seemed more consistent.

### DIFF
--- a/packages/opam-client/opam-client.2.0.0~rc4/descr
+++ b/packages/opam-client/opam-client.2.0.0~rc4/descr
@@ -1,0 +1,3 @@
+opam 2.0 development libraries
+
+Actions on the opam root, switches, installations, and front-end.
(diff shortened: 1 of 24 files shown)


kit-ty-kate:
I don't know why but for some reason camelus is detecting that most packages are broken.
As for the version naming mmh I would probably be more in favour of `2.0.0` if both are really equal

rjbou:
Re-taggued on `2.0.0~rc4` and camelus is happy!

AltGr:
Camelus was failing because the packages had mutual constraints on version `"2.0.0"`, but were actually on version `"2.0.0~rc4"`. I fixed it but Camelus took some time to catch up.

Thankfully, we will soon be able to express version constraints on the same version with opam 2!
I'll open another PR with 2.0.0 and let the maintainers decide then.

AltGr:
We're still getting concurrent dl errors (429) from Github in the CI.

PR #12394, opened 2018-07-25
opam 2.0.0 libs and devel packages

AltGr:
This is exactly the same as #12390, except that the packages are named with version `2.0.0` instead of `2.0.0~rc4`. The two are the same anyway.

Note that the download links point to the `2.0.0-rc4` archives, since these will only be renamed when we do the repository format shift. But the version reported by `--version` **is** `2.0.0`.


rjbou:
Merging this one instead of #12390. There is still some errors that are not specific to this PR.

PR #12401, opened 2018-07-26
Add elina.1.1

skcho:


mseri:
It seems to be incompatible with 4.07.0 because of ppx_import:

```
  - jupyter -> ppx_deriving_yojson >= 3.0 -> ppx_import >= 1.1
```

Should we add a bound, or try to fix `ppx_import`?

PR #12402, opened 2018-07-26
Release ocaml-jupyter and jupyter-archimedes v2.3.2

akabe:
Release note: https://github.com/akabe/ocaml-jupyter/releases/tag/v2.3.2


mseri:
It seems to be incompatible with 4.07.0 because of ppx_import:

```
  - jupyter -> ppx_deriving_yojson >= 3.0 -> ppx_import >= 1.1
```

I'll see if it is reasonable to fix `ppx_import`. I think the fact that the installation is prevented with a clear message on 4.07.0 is good enough to allow a merge

PR #12403, opened 2018-07-26
Fix paths for NetBSD in conf-gmp and zarith

maroneze:
On NetBSD (8.0 at least), by default (using pkgsrc/pkgin), files are installed in `/usr/pkg`.

I am not an expert NetBSD user, but fixing these paths allowed me to compile `conf-gmp` and `zarith`. However, the string `/usr/pkg` is entirely absent from `opam-repository`, which makes me wonder if previous NetBSD users have just not installed these packages, or if there is another way to configure NetBSD to avoid having to do so.

If there are more experienced NetBSD users in the opam community, I'd like them to review my commit, but I do not know how to find them.

### DIFF
--- a/packages/conf-gmp/conf-gmp.1/opam
+++ b/packages/conf-gmp/conf-gmp.1/opam
@@ -6,8 +6,9 @@ bug-reports: "https://github.com/ocaml/opam-repository/issues"
 dev-repo: "https://github.com/ocaml/opam-repository.git"
 license: "GPL"
 build: [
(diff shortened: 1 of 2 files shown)


maroneze:
Well, according to a comment in https://github.com/ocaml/opam/issues/3482, it seems my patch is incorrect, so I am closing it to avoid inconsistencies between different BSDs (I'd put it under WIP as in Gitlab, but apparently Github does not have such a feature). Will reopen if a solution to the hardcoded-paths-conundrum is found.

PR #12409, opened 2018-07-27
[DO NOT MERGE] Trigger Lwt revdeps build

aantron:
See https://github.com/ocsigen/lwt/issues/603#issuecomment-407918060. We are making a change to the type signature of `Lwt.async`, and we want to see which dependees will be affected.

Please don't merge this. We just want to trigger the revdeps CI.

### DIFF
--- a/packages/lwt/lwt.5.0.0/descr
+++ b/packages/lwt/lwt.5.0.0/descr
@@ -1,0 +1,10 @@
+Promises, concurrency, and parallelized I/O
+
+A promise is a value that may become determined in the future.
+
+Lwt provides typed, composable promises. Promises that are resolved by I/O are
+resolved by Lwt in parallel.
+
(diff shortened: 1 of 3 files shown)


aantron:
@hannesm What is the command or script the CI uses to list the deps? I get a much wider list on my machine, running

```
opam list --unavailable --depends-on=lwt --depopts --recursive --short
```

Or, which repo can I find the CI code in?

kit-ty-kate:
The datakit-ci source code is here: https://github.com/avsm/mirage-ci/
As for the reverse dependencies we don't do recursive dependencies, otherwise it would just take too much resources and would be unbearable. All the steps are written on the web interface: Here is the command we use to get the revdeps: `opam list -s --color=never --depends-on lwt.5.0.0 --installable --all-versions --depopts`

aantron:
Thanks, I suppose that is good enough :)

PR #12411, opened 2018-07-30
cpdf.2.2.1: unsafe string fix

UnixJunkie:


pmetzger:
Just FYI, calling a PR "unsafe string fix" means people have no idea what package is associated with the request without clicking through. Many people monitor the pull request and issues queues looking for things they care about. It is always best if a pull request or issue references the specific package it is associated with.

samoht:
Seems that there is an issue on 4.07:

```
[cpdf: OCAMLPARAM=safe-string=0,_ make] Command started
+ OCAMLPARAM=safe-string=0,_ "make" (CWD=/home/travis/.opam/system/build/cpdf.2.2.1)
[ERROR] The compilation of cpdf failed at "OCAMLPARAM=safe-string=0,_ make".
```

samoht:
The PR is a bit strange; you should probably embed the `safe-string` detection in your build script (or even better into your build system) instead of relying on opam to use the right command. And I am not sure that the build command is doing what you want: setting environment variable in subshells is always a bit tricky.

UnixJunkie:
I am OK to close this PR (which has taken way too much of my time for what it's worth) and wait for the software maintainer to fix his software in the right way.

PR #12415, opened 2018-07-31
ptmap 2.0.4

andyli:
OCaml 4.07 compatibility. See https://github.com/backtracking/ptmap/pull/13.


kit-ty-kate:
Could you also constrain the previous versions to OCaml < 4.07 and flag the `ocamlfind` dependency as a `{build}` dependency?

kit-ty-kate:
Thanks a lot!

PR #12418, opened 2018-07-31
Add Genspio 0.0.1

smondet:


smondet:
The Travis failure is not in the package, it is for OCaml 4.00 but we have `available: [ ocaml-version >= "4.03.0" ]` anyway.

kit-ty-kate:
Looks good. Thanks!

PR #12422, opened 2018-08-01
Fix lutin typo

rjbou:
Changed the path of package lutin. As they weren't the same (capital), it was not ported on the 2.0 branch.
Related to #11990 and /cc @jahierwan.


mseri:
Thanks

PR #12423, opened 2018-08-01
remove bencode-rpc

c-cube:
this has always been a toy project, in retrospect, and there are many more serious alternatives these days.


mseri:
Does not break dependencies, I would be happy to merge. Will https://opam.ocaml.org/packages/bencode_rpc be automatically removed if we merge or we need to do something for that?

c-cube:
I have no idea, sorry.

kit-ty-kate:
Removing packages is not something opam-repository maintainers agree on.
For now the consensus seems to be to switch the packages to "remove" to `available: false` instead of removing them from the repository.

c-cube:
Why not? I'm pretty sure no one depends on it, and it just pollutes the list of results for `rpc`…

kit-ty-kate:
I don't know, I personally don't care but other people from @ocaml/opam-repository might have arguments against it.

samoht:
If that's a toy project that no one is depending on, I am fine to remove it :-)

hcarty:
One downside that will hopefully go away as opam2 spreads is that an opam1 user's switch can end up in a broken state if a definition for an installed package disappears from the repository. opam2 does not have the same problem as far as I understand.

PR #12425, opened 2018-08-01
mirage.3.1.1

hannesm:
- for the unix target, add `-tags thread`, as done for the mac osx target (suggested by @cfcs)
- bump minimum mirage-solo5* and solo5-kernel* to 0.3.0 (by @hannesm, as suggested by @mato)
- use the exposed signature in functoria for `Key` modules (by @Drup)
- add `?group` param to all generic devices (by @samoht)

### DIFF
--- a/packages/mirage/mirage.3.1.1/opam
+++ b/packages/mirage/mirage.3.1.1/opam
@@ -1,0 +1,33 @@
+opam-version: "1.2"
+maintainer:   ["anil@recoil.org" "thomas@gazagnaire.org"]
+authors:      ["Thomas Gazagnaire" "Anil Madhavapeddy" "Gabriel Radanne"
+               "Mindy Preston" "Thomas Leonard" "Nicolas Ojeda Bar"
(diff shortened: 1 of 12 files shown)


packages/mirage/mirage.3.1.1/opam
```diff
@@ -0,0 +1,33 @@
...
+  "astring"
+  "logs"
+  "mirage-runtime"     {>= "3.0.0"}
+]
+conflicts: [
+  "nocrypto" {< "0.4.0"}
+  "cstruct"  {< "1.0.1"}
+  "io-page"  {< "1.4.0"}
+  "crunch"   {< "1.2.2"}
+  "jbuilder" {= "1.0+beta18"}
```
kit-ty-kate: Is there no other compatible versions?

kit-ty-kate: nevermind

kit-ty-kate:
Thanks!

PR #12441, opened 2018-08-06
Add rpclib

samoht:
/cc @jonludlam


mseri:
@jonludlam @gaborigloi we should announce this (now that also the lwt, ashnc and ppx parts are released) in th discourse forum. Esoecially given the improvements i the documentation (readme) added by gabor

gaborigloi:
@mseri Should we wait for https://github.com/mirage/ocaml-rpc/pull/117 to be merged and the interface to be stabilized first? That way we'd avoid breaking changes after the announcement.
I think we should merge that PR and test it, and possibly remove the exn stuff that we don't need, and then we can have another release.

mseri:
@gaborigloi ok, makes sense to me

PR #12443, opened 2018-08-06
More rpclib packages

samoht:
/cc @jonludlam


mseri:
Thanks for the release! 

I was waiting to finish breaking the interfaces to do the release (a big breaking change is still pending in review)... but I think it was a bad idea, we should have relased at least the jbuildered version.

PR #12444, opened 2018-08-06
Add an extra-source to compile mirage-xen-minios with sandboxing

samoht:
/cc @avsm @talex5

PR #12451, opened 2018-08-08
Use opam2 in travis scripts

samoht:


samoht:
TravisCI is green and the script has been tested successfully with #12444 so merging.

PR #12452, opened 2018-08-08
Package resource-pooling.0.2

jrochel:
### `resource-pooling.0.2`

library for pooling resources like connections, threads, or similar

This package is derived from the module Lwt_pool from the lwt package,
which implements resource pooling. With Resource_pool this package
provides a modified version with additional features. Also there is a
module called Server_pool that manages resource clusters, specifically
a cluster of servers each with its own connection pool.

---
* Homepage: https://github.com/ocsigen/resource-pooling
* Source repo: https://github.com/ocsigen/resource-pooling.git
* Bug tracker: https://github.com/ocsigen/resource-pooling/issues

---


kit-ty-kate:
Same remarks than before but it's good to merge. Thanks!

PR #12458, opened 2018-08-09
ocaml-freestanding 0.3.1

mato:


mato:
The CI failures are due to https://github.com/Solo5/solo5/pull/267; this will be fixed by a Solo5 point release which I'll make on Monday. So this can either wait till then, or go ahead and merge anyway.

packages/ocaml-freestanding/ocaml-freestanding.0.3.1/opam
```diff
@@ -0,0 +1,25 @@
...
+bug-reports: "https://github.com/mirage/ocaml-freestanding/issues/"
+license: "MIT"
+tags: "org:mirage"
+dev-repo: "https://github.com/mirage/ocaml-freestanding.git"
+build: [make]
+install: [make "install" "PREFIX=%{prefix}%"]
+remove: [make "uninstall" "PREFIX=%{prefix}%"]
+depends: [
+  "conf-pkg-config"
+  "ocamlfind"
```
kit-ty-kate: Are you using the `Findlib` library? If not you might want to flag this as a `{build}` dependency.
Same with the one above I think.

mato: @kit-ty-kate I only use the `ocamlfind` command at build time, so yes, this can be a build dependency. In fact, same for `ocaml-src`. `conf-pkg-config` has to stay as-is since `pkg-config` (which it installs via depext) is required at "run" time to build unikernels. Will push a fix.

mato:
Solo5 fixes for the build failures are in #12459, I will push the `{build}` change here once that is merged.

kit-ty-kate:
Done

kit-ty-kate:
All green. Thanks!!

PR #12462, opened 2018-08-12
Package exenum.0.86

lebotlan:
### `exenum.0.86`

Build efficient enumerations for datatypes. Inspired by Feat for Haskell.

The exenum library offers constructors to build enumerations for
datatypes, that is, functions from (arbitrarily large) integers to
values. Such enumerations are typically used for unit testing. The
library is efficient: the n-th element of an enumeration is returned
without having computed the (n-1) previous elements. Complexity is in
log(n), except for some pathological datatypes. See the homepage for
details: https://github.com/lebotlan/ocaml-exenum
Inspired by Feat: Functional Enumeration of Algebraic Types, by
Duregard, Jansson, Wang, Chalmers University.

As an example, consider the following datatype:
type term = Var of string | App of term * term | Lambda of string * term

Using exenum, one may easily generate zillions of different
lambda-terms. In our specific example, term number 2000000000000
happens to be
((((x v) (fun u -> y)) ((fun u -> y) (fun y -> y))) (((x
v) (fun u -> v)) (fun u -> y)))

---
* Homepage: https://github.com/lebotlan/ocaml-exenum
* Source repo: git://github.com/lebotlan/ocaml-exenum.git
* Bug tracker: https://github.com/lebotlan/ocaml-exenum/issues

---


packages/exenum/exenum.0.86/opam
```diff
@@ -0,0 +1,14 @@
...
+authors: "Didier Le Botlan <github.lebotlan@dfgh.net>"
+homepage: "https://github.com/lebotlan/ocaml-exenum"
+bug-reports: "https://github.com/lebotlan/ocaml-exenum/issues"
+license: "MIT License"
+dev-repo: "git://github.com/lebotlan/ocaml-exenum.git"
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+]
+depends: ["dune" "zarith"]
```
kit-ty-kate: Could you flag `dune` as a `{build}` dependency?

lebotlan: Sure! (It would be worth adding it to the dune manual.)
Besides, the opam manual lacks an example. Is the following correct?
depends: ["dune" {build} "zarith"]

kit-ty-kate: > Is the following correct? depends: ["dune" {build} "zarith"]

Yes, though a more common/readable way would be to indent each dependencies as follow:
```
depends: [
  "dune" {build}
  "zarith"
]
```

lebotlan: Thanks for your advice. I have opam-published the package again.

PR #12470, opened 2018-08-14
orocksdb works fine on darwin as-is

cmr:


kit-ty-kate:
cc @domsj

kit-ty-kate:
Could you at least add the right external package to install on osx to the `conf-rocksdb` package?
I believe this one might be enough: https://formulae.brew.sh/formula/rocksdb

kit-ty-kate:
> Could you at least add the right external package to install on osx to the conf-rocksdb package?

otherwise it can't be tested by the CI

cmr:
unfortunately out of the box the homebrew package isn't sufficient. there's no pkg-config and the artifacts get installed outside of `/usr/local/lib`, in `/usr/local/Cellar/rocksdb/{version}/lib`. the orocksdb build system listens to ROCKS_LIBDIR, though. is there some way to export that for CI?

kit-ty-kate:
> is there some way to export that for CI?

mmh no there is not except maybe have special rules in the build phase.

But I guess it's good enough as-is. Let's merge that and this can be corrected later if needed/possible.
Thanks!

PR #12472, opened 2018-08-16
Add gapi-ocaml 0.3.7, google-drive-ocamlfuse 0.6.26

astrada:


hcarty:
Any idea why the Travis tests are failing on 4.05.0 and older?

astrada:
The tests fails only for the old version that I kept because it's the last version that works with very old compilers (3.12). The problem is that the tests work only if pa_monad_custom is installed and this dependency is marked as deptopt. But the Travis env doesn't install it, so the tests don't compile.

kit-ty-kate:
> The problem is that the tests work only if pa_monad_custom is installed and this dependency is marked as deptopt.

Then could you add `"pa_monad_custom" {test}` to the dependencies? (`depends` field)

packages/gapi-ocaml/gapi-ocaml.0.3.4/opam
```diff
@@ -28,9 +28,7 @@ depends: [
   "ocurl"
   "xmlm"
   "yojson"
-]
-depopts: [
-  "ounit" {build & test}
-  "pa_monad_custom" {build}
+  "ounit" {test}
+  "pa_monad_custom" {test}
```
kit-ty-kate: is it really only needed for the tests then?

astrada: Yes, more or less. There are other targets that use that dependency but you cannot build them with opam anyway.

kit-ty-kate:
Any reasons you would remove every previous versions? Could you avoid this?

astrada:
> Any reasons you would remove every previous versions? Could you avoid this?

It's not a library, it's an executable, the latest versions have bugfixes/new functionalities. The old versions require adding new constraints to compile, and it's a lot of work for something no one uses anymore.

astrada:
Is there something I still need to do to get this PR merged?

astrada:
I rebased the patch

kit-ty-kate:
@ocaml/opam-repository can we have a consensus on whether we accept to remove packages and in which cases?

astrada:
I can restore the deleted packages, but the revdeps will be broken. If you don't mind, I will rebase this patch.

samoht:
If nobody is keen to fix old constrains, I am fine to remove the old packages. Having broken constraints in the repo is not good for our users and opam is able to deal with deleted packages anyway.

rgrinberg:
If deleting packages doesn't affect the workflow of users using https://github.com/AltGr/opam-lock then I think it's ok to delete.

astrada:
I agree not to remove library packages, because there can be other projects depending on specific versions, but for an application I don't think that's a problem. Is there a way to deprecate old versions? (Something like NPM [deprecate](https://docs.npmjs.com/cli/deprecate)?)

kit-ty-kate:
Sorry for the time it took to deal with that. Let's take some decision and merge it.
Thanks and sorry again.

astrada:
No problem! :smile:

PR #12481, opened 2018-08-18
Fix ppx_deriving dependency of nocrypto

smondet:
The library seems to depend on `ppx_deriving.runtime` at runtime.

### DIFF
--- a/packages/nocrypto/nocrypto.0.5.4-1/opam
+++ b/packages/nocrypto/nocrypto.0.5.4-1/opam
@@ -21,7 +21,7 @@ depends: [
   "topkg" {build}
   "cpuid" {build}
   "ocb-stubblr" {build}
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
cc @pqwy @hannesm

kit-ty-kate:
Thanks!

kit-ty-kate:
(it might also have been worth doing on every versions, not just the latest but this PR is good enough as is)

smondet:
@kit-ty-kate for some reason, the `{build}` tagging was only in this version, it was added here:
https://github.com/ocaml/opam-repository/commit/5a3b33bd67594bd4b86f7303c2d6f66cbc47866d
(I actually don't understand why it was added as a new `...-1` package-version)

PR #12484, opened 2018-08-19
faad.0.4.0

toots:


toots:
CentOS distribution doesn't have `faad2-devel` but it's available in secondary repositories:https://pkgs.org/download/faad

kit-ty-kate:
Seems like this library requires either OCaml >= 4.02 or to depend on the `base-bytes` package, but it's probably the former considering the error message:
```
- File "faad.mli", line 46, characters 16-23:
- Error: Unbound module Bytes
```

kit-ty-kate:
mmh that fix was indeed unlikely to work. Could you add a lower-bound instead or fix your release?

toots:
@kit-ty-kate I don't get it. Isn't `base-bytes` supposed to provide the `Bytes` module?

kit-ty-kate:
It only provides a `Bytes` module in a `bytes` package. If you didn't use the `bytes` package in the first place, the compiler doesn't see it.

toots:
Ya makes sense, I don't know what confused me.. Just uploaded a fixed release.

PR #12490, opened 2018-08-19
Fixed old AIFAD cfg version constraints

mmottl:


kit-ty-kate:
Fair enough. Thanks!

Since you are the maintainer of the cfg package, fwi it seems that it is not compatible with OCaml >= 4.06:
```
#=== ERROR while compiling cfg.2.0.4 ==========================================#
# context              2.0.0 | linux/x86_64 | ocaml-base-compiler.4.06.1 | git+file:///home/opam/opam-repository
# path                 ~/.opam/4.06/.opam-switch/build/cfg.2.0.4
# command              ~/.opam/4.06/.opam-switch/build/cfg.2.0.4/./configure --prefix /home/opam/.opam/4.06
# exit-code            2
# env-file             ~/.opam/log/cfg-19-df0ac0.env
# output-file          ~/.opam/log/cfg-19-df0ac0.out
### output ###
# File "./setup.ml", line 247, characters 8-26:
# Warning 3: deprecated: String.set
# Use Bytes.set instead.
# File "./setup.ml", line 247, characters 8-11:
#          bytes
```

mmottl:
Fixing compiler version constraints in a separate pull request...

PR #12492, opened 2018-08-20
Release ppx_regexp.0.4.0 and ppx_tyre.0.4.0.

paurkedal:
This PR updates `ppx_regexp` and adds `ppx_tyre` (syntax support for tyre).  Both packages are based on the same repo and the plan is to keep versions is sync.


kit-ty-kate:
Thanks!

PR #12499, opened 2018-08-21
ordma.0.0.5

andersfugmann:
New release on behalf of @toolslive.


andersfugmann:
Alpine is not supported, as it does package librdmacm - hence the fail on alpine.

kit-ty-kate:
The failure on Alpine is indeed legit. However the failures on Travis are not:
```
  - ordma -> lwt_test
  - No package named lwt_test found.
```

kit-ty-kate:
thanks a lot!

PR #12506, opened 2018-08-23
Add a version of ocaml-manual for OCaml 4.07.0

pmetzger:
The changes required seemed very small and this was absent.

### DIFF
--- a/packages/ocaml-manual/ocaml-manual.4.07.0/descr
+++ b/packages/ocaml-manual/ocaml-manual.4.07.0/descr
@@ -1,0 +1,1 @@
+The OCaml system manual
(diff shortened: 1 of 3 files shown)


pmetzger:
Can someone explain this to me? The error message is a bit obscure:

> warning 41: Some packages are mentionned in package scripts of features, but there is no dependency or depopt toward them: "ocaml-manual"

kit-ty-kate:
oh it's probably because `ocaml-manual` is referencing itself in `ocaml-manual:doc` (where the better version should be `doc`) but I think that's minor since previous versions all use this without any problems.

Merging, can be fixed later. Thanks!

PR #12511, opened 2018-08-24
+x509.0.6.2

hannesm:
the changes only include support for `ppx_sexp_conv > v0.11.0`, which is required for 4.07.0 compatibility


pmetzger:
An FYI: usenix15.nqsb.io doesn't seem to be reachable.

hannesm:
@pmetzger thanks for noticing, i know and am working on getting it back up again soon...

pmetzger:
Ah, that's fine then. I'd worried that it was pointing at the wrong documentation site.

hcarty:
Thank you @hannesm!

hannesm:
@pmetzger the server (well, virtual machine) is back now :)

PR #12512, opened 2018-08-24
+tls 0.9.2

hannesm:
NEWS in 0.9.2 (from changes):
- 4.07 support by adding an optional runtime dependency of `ppx_sexp_conv` (required since `v0.11.0`)
- ALPN (RFC [7301](https://tools.ietf.org/html/rfc7301)) support by @bobbypriambodo


hcarty:
Thanks!

Issue #12513, opened 2018-08-24
ocp-index 1.1.6 not installing under 4.07 on MacOS


pmetzger:
I get the following errors (ultimately fatal) while building:

```
#=== ERROR while installing ocp-index.1.1.6 ===================================#
# opam-version 1.2.2
# os           darwin
# command      jbuilder build -p ocp-index
# path         /Users/perry/.opam/4.07.0/build/ocp-index.1.1.6
# compiler     4.07.0
# exit-code    1
# env-file     /Users/perry/.opam/4.07.0/build/ocp-index.1.1.6/ocp-index-6281-d2c37b.env
# stdout-file  /Users/perry/.opam/4.07.0/build/ocp-index.1.1.6/ocp-index-6281-d2c37b.out
# stderr-file  /Users/perry/.opam/4.07.0/build/ocp-index.1.1.6/ocp-index-6281-d2c37b.err
### stderr ###
# [...]
# Use Re.Posix
#     ocamlopt libs/.indexLib.objs/indexScope.{cmx,o}
# File "_none_", line 1:
# Warning 58: no cmx file was found in path for module Nstream, and its interface was not compiled with -opaque
# File "_none_", line 1:
# Warning 58: no cmx file was found in path for module Pos, and its interface was not compiled with -opaque
#     ocamlopt libs/.indexLib.objs/indexBuild.{cmx,o} (exit 2)
# (cd _build/default && /Users/perry/.opam/4.07.0/bin/ocamlopt.opt -w -40 -g -I libs/.indexLib.objs -I /Users/perry/.opam/4.07.0/lib/ocaml/compiler-libs -I /Users/perry/.opam/4.07.0/lib/ocp-indent/lexer -I /Users/perry/.opam/4.07.0/lib/ocp-indent/utils -intf-suffix .ml -no-alias-deps -o libs/.indexLib.objs/indexBuild.cmx -c -impl libs/indexBuild.pp.ml)
# File "libs/indexBuild.ml", line 413, characters 30-40:
# Error: Unbound record field Ident.name
```

hcarty:
This has been reported on the ocp-index issue tracker - https://github.com/OCamlPro/ocp-index/issues/117

@pmetzger Could you submit a PR against opam-repository restricting ocp-index to OCaml < 4.07.0?  This unfortunately was not caught during the OCaml 4.07.0 release.

pmetzger:
[...]

PR #12514, opened 2018-08-24
ocp-index: constrain to ocaml versions below 4.07

pmetzger:
As requested by @hcarty in #12513


hcarty:
This is still imperfect wrt 1.1.4, but it at least avoids the build failure for OCaml 4.07.0.

Thank you @pmetzger!

hcarty:
The 1.1.4 issue would require more digging - I'm not sure it's worth holding up this PR over that.

PR #12520, opened 2018-08-27
camltc.0.9.8

andersfugmann:
Switch to dune

### DIFF
--- a/packages/camltc/camltc.0.9.8/files/osx.patch
+++ b/packages/camltc/camltc.0.9.8/files/osx.patch
@@ -1,13 +1,31 @@
 diff --git a/src/dune b/src/dune
-index be85e03..e1c7ca9 100644
+index be85e03..05d61a4 100644
 --- a/src/dune
 +++ b/src/dune
-@@ -23,7 +23,7 @@
+@@ -9,6 +9,7 @@
+  (flags -w -27)
+  (c_names otc_wrapper)
+  (c_flags -O2 -I 3rd-party/tokyocabinet)
++ (modes native)
+ )
+ 
+ (library
+@@ -16,15 +17,15 @@
+  (public_name camltc.tc)
+  (self_build_stubs_archive (tc))
+  (modules)
++ (modes native)
+ )
+ 
+ (rule
+- (targets libtc_stubs.a dlltc_stubs.so)
++ (targets libtc_stubs.a)
   (action
    (bash "
     (../../../src/3rd-party/tokyocabinet/configure --disable-bzip --disable-zlib
 -   make libtokyocabinet.so libtokyocabinet.a -j $(nproc)) > /dev/null 2>&1
-+   make libtokyocabinet.so libtokyocabinet.a) > /dev/null 2>&1
-    cp libtokyocabinet.so dlltc_stubs.so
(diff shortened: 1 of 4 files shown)


andersfugmann:
continuous-integration/travis-ci/pr failed on network error for ocaml 3.12 (which is unsupported anyways)

packages/camltc/camltc.0.9.8/files/osx.patch
```diff
@@ -0,0 +1,31 @@
+diff --git a/src/dune b/src/dune
+index be85e03..05d61a4 100644
+--- a/src/dune
++++ b/src/dune
+@@ -9,6 +9,7 @@
+  (flags -w -27)
+  (c_names otc_wrapper)
+  (c_flags -O2 -I 3rd-party/tokyocabinet)
++ (modes native)
```
mseri: I am surprised that this is needed, I thought `native` was the default value

andersfugmann: I think native is the default indeed, but as I dont know how to build a shared lib under osx (I dont have access to mac), I believe this will disable it. Anyways it will not compile without it.

packages/camltc/camltc.0.9.8/opam
```diff
@@ -0,0 +1,26 @@
...
+authors: ["Jan Doms" "Joost Damad" "Romain Slootmaekers" "Nicolas Trangez"]
+homepage: "http://github.com/toolslive/camltc"
+bug-reports: "http://github.com/toolslive/camltc/issues"
+dev-repo: "https://github.com/toolslive/camltc.git"
+license: "Apache-2.0"
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+]
+build-test: [
+  ["dune" "build" "@runtest" "-j" jobs]
```
mseri: I think the idiomatic way is `["dune" "runtest" "-p" name "-j" jobs]`, they are nearly equivalent but specifying the name will restrict the compiled units and tests to the relevant ones. i suggest the idiomatic form only because using that uncovered some transitive dependencies issues in packages that I was maintaining in the past

mseri:
I have restarted that test but it keeps failing, there may be some issues with the network or ubuntu trusty mirrors. Overall it looks good. I suggested a change, please have a look and let me know what you prefer to do.

packages/camltc/camltc.0.9.8/opam
andersfugmann: I've updated the rule for build-test to follow your suggestion. 
thanks for the comments.

packages/camltc/camltc.0.9.8/files/osx.patch
andersfugmann: just to elaborate a bit: Under linux, If I remove the rule to generate the 'dlltc_stubs.so', the build will fail. Adding `(modes native)` to the library camltc.tc make is complete, and I assume the same goes for the osx build. 
I would agree that the `(modes native)` is slightly redundant for the camltc library, as I assume its inferred from camltc.tc lib, but is causes no harm.

andersfugmann:
Travis seems to have network problems again :-/

mseri:
Thanks for the patience. Looks good now

PR #12530, opened 2018-08-28
New release: slacko 0.14.1

Leonidas-from-XIV:


mseri:
Looks incompatible with ocaml 4.07.0 because of a transitive dependency:
```
  - slacko -> ppx_deriving_yojson >= 3.0 -> ppx_import >= 1.1
```
I think it is fixed on ppx_import master and only requires a release.

Leonidas-from-XIV:
@mseri This is a bit weird since it seems to build on Travis on 4.07 and it doesn't seem to depend on ppx_import: https://travis-ci.org/Leonidas-from-XIV/slacko/jobs/421709181

mseri:
That's strange. From what I can see in the repo, `ppx_deriving_yojson` 3.1 does depend on `ppx_import` https://github.com/ocaml-ppx/ppx_deriving_yojson/blob/v3.1/opam while the current master does not https://github.com/ocaml-ppx/ppx_deriving_yojson/blob/master/ppx_deriving_yojson.opam

mseri:
Oh, I see. It does depend on it only for tests and the CI also runs the tests (and for an opam 1 bug, it installs the test dependencies also of all the dependencies). I think we can merge

mseri:
One last thing. `slack-backup` seems to require an upper bound: 
```
# ocamlfind: [WARNING] Package `threads': Linking problems may arise because of the missing -thread or -vmthread switch
# File "lib/slack_backup_lib.mli", line 44, characters 6-26:
# Error: Unbound type constructor Slacko.authed_result
# Command exited with code 2.
```

Leonidas-from-XIV:
Thanks for looking into it. I've added the constraint.

mseri:
Thanks. It looks like the bound should be `< 0.13.0`. I made the change directly in the PR, if it succeeds I'll merge. I am surprised we did not catch that before

mseri:
Thanks. The remaining test failures are the we have already discussed. Merging

PR #12532, opened 2018-08-29
Add conflict between pkcs11 and asn1-comb >= 0.2.0

NathanReb:
Old version of `pkcs11` (> `0.9.0`) conflicts with `asn1-combinators` `0.2.0` and upwards.

Those `pkcs11` versions depend directly on `asn1-combinators` although they don't declare it (it used to work because they depend on `key-parsers` which in turns depends on it).

The best solution is probably actually to add this dependency to the opam files but the conflict seemed like an easier solution. I'm happy to change that if you feel like it's required.


NathanReb:
Any idea what's wrong with good old camelus here?

PR #12534, opened 2018-08-29
vorbis.0.7.1

toots:


toots:
The revdep checks failed on outdated packages. I can remove them but it doesn't seem like this is the usual way to proceed. Let me know!

mseri:
Looks good. Please add `available: [ocaml-version >= "4.03"]` to the opam files

mseri:
It would also be great if you can add the necessary upper bounds to the outdated packages

toots:
I'll do what I can. I think `schroedinger` should be removed, the upstream library has been sunset now.

toots:
Ok, I've added the `[ocaml-version >= "4.03"]`. I've tried to add upper-bound and such for the outdated modules but it's pushing them further into the CI, with reveps and distros testing. I really think these old packages should be left as they are or simply removed. There's no need to keep maintaining them, they're mostly and only deps for liquidsoap.

mseri:
I think it is fine as it is, I’ll see what we can do for the rest. Thanks for the help.

toots:
Thanks!

PR #12535, opened 2018-08-30
Elpi 1.0.5

gares:


mseri:
In the two CI failures, for some reasons, `opam` has chosen to install `ppx_tools_versioned` version `5.0beta1` instead of `5.1`. Maybe we should just say `"ppx_tools_versioned" {= "5.1"}`

mseri:
I've just noticed your commit appearing! Awesome, thanks

gares:
Thanks for the quick diagnosis! Maybe I should have separated the two changes :-/

ejgallego:
Umm, the strict dependency looks very fishy, wouldn't that make elpi basically uninstallable once some package starts to require `ppx_tools_versioned > 5.1` ?

mseri:
I don't think it's a problem: in that case one can install `elpi` 1.0.5 that does work with the new `ppx_tools_versioned`

gares:
elpi 1.0.5 works with both versions of ppx and is 100% compatible with 1.0.4, so I believe that users should have no problem upgrading.

I fixed the metadata of 1.0.4 just to "signal" the problem (unsatisfiable depends is much better than a compilation error), eg maybe someone (else) had a `opam install elpi.1.0.4` in a script that started to fail two days ago. I don't know the largest range of ppx_tools_versioned that works, esp the lower bound, and I've no time to try, so I'm ok with the strict constraint.

PR #12538, opened 2018-08-30
Upgrade OByteLib to version 1.4.

bvaugon:


hcarty:
Could you add license information to the opam metadata?  Then it should be good to merge.

mseri:
Thanks. Looks good now

PR #12542, opened 2018-08-31
[new release] eqaf (v0.1)

dinosaure:
Constant time equal function on `string`

- Project page: <a href="https://github.com/dinosaure/eqaf">https://github.com/dinosaure/eqaf</a>
- Documentation: <a href="https://dinosaure.github.io/eqaf/">https://dinosaure.github.io/eqaf/</a>

##### CHANGES:

* First release

### DIFF
--- a/packages/eqaf/eqaf.0.1/descr
+++ b/packages/eqaf/eqaf.0.1/descr
@@ -1,0 +1,14 @@
+Constant time equal function on `string`
+
+From some crypto libraries like
+[digestif](https://github.com/mirage/digestif.git) or
+[Callipyge](https://github.com/oklm-wsh/Callipyge.git), it needed to have a
+constant time equal function to avoid timing attacks. To avoid replication of
+code and ensure maintainability of this kind of function, we decide to provide a
+little package which implements `equal` function on `string`.
+
+This package provides 2 implementations (a C implementation which comes from
+OpenBSD's code and an OCaml implementation which is simply a translation of the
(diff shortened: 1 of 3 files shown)


dinosaure:
Did you know a way to constraint tests to be available only on `ocaml > "4.03.0"` ?

mseri:
I think (but I've never tried) that you can add `{ ocaml-version > "4.03.0" }` as a filter to the `build-test` section

dinosaure:
Ok, so `ppx_expect` does not compile on `4.03.0` and we continue to have a bug on Mac OSX about expected results - the situation could be better if we provide only the C implementation of the equal function, @hannesm and @cfcs, do you have any opinion about that?

NOTE: I can not test the code, I did not have any access on a Mac.

cfcs:
@dinosaure I think that seems sensible, it's not much code, and if the whole reason for having this is that the OCaml implementation is not in fact constant-time, I don't see a good way around this.

dinosaure:
CI fails because benchmark/test runs on `4.03.0` but I explicitly put a constraint like what @mseri said on `build-test`. Any other idea?

mseri:
the problem is that it tries to install the test dependencies even if the tests are not really running. I'll keep your last commit because it shows that the tests are expected to fail for 4.03.0. I'll merge and if we find a better solution will send a separate PR. Thanks

dinosaure:
Thanks a lot 👍 !

PR #11947, opened 2018-05-09
+flowtype.0.72.0

panagosg7:

PR #11302, opened 2018-01-24
+flowtype.0.64.0

avikchaudhuri:

PR #12545, opened 2018-08-31
+flowtype.0.80.0

gabelevi:


fdopen:
There are still no install instructions. And probably there is still no `*.install` file shipped or created. See https://github.com/ocaml/opam-repository/pull/11947#issuecomment-388826752 and https://github.com/ocaml/opam-repository/pull/11302#issuecomment-361242351

PR #12546, opened 2018-08-31
+flow_parser.0.80.0

gabelevi:
We haven't sync'd flow_parser to opam in ages. Let's see if this works...


gabelevi:
Oh right...let's see if pinning `ppx_tools_versioned` to 5.2 will fix that...

mseri:
Looks like it worked, although ppx_tools_versioned 5.2 is kind of broken. I am merging, we can change it later if sedlex if fixed

dexterleng:
Could a newer version be published?

PR #12548, opened 2018-09-01
Fix ezsqlite opam file

zshipko:
I accidentally included `str` in the opam file instead of the `dune` file. Sorry about that!


mseri:
I wonder why the CI did not fail, and only camelus complained in an uninformarive way

mseri:
Camleus keeps being unhappy, I still don’t understand why

zshipko:
Yeah, that’s weird! I’m happy to fix whatever the issue is but can’t find anything on my end either.

PR #12553, opened 2018-09-03
[frama-c] blacklist incompatible why3 version

maroneze:
why3 1.0.0 was released shortly before Frama-C-20180502, but both are incompatible. Compilation works, but some specific parts of Frama-C are not compatible with this version of why3, so it is blacklisted in the latest versions of Frama-C (to avoid opam from trying to downgrade Frama-C if this version of why3 is already installed).

### DIFF
--- a/packages/frama-c/frama-c.20171101/opam
+++ b/packages/frama-c/frama-c.20171101/opam
@@ -88,10 +88,6 @@ build-doc: [
    [make "-C" "doc" "install"]
 ]
 
-build-test: [
-  [make "-j%{jobs}%" "PTESTS_OPTS=-error-code" "tests"]
-]
-
 depends: [
   "ocamlgraph" { >= "1.8.8" & < "1.9~" }
   "ocamlfind"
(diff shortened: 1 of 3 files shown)


kit-ty-kate:
```
[ERROR] why3-base (< 0.86 | >= 1.0.0) & coq < 8.4.6 & coq >= 8.7 & lablgtk <
        2.18.2 & frama-c-e-acsl & frama-c-base is not a valid conjunction
```
I believe the correct way should probably be:
```
"why3-base" {< "0.86"}
"why3-base" {>= "1.0.0" }
```

maroneze:
Thanks, I tried using the syntax mentioned in https://github.com/ocaml/opam/issues/2958 this time.

maroneze:
I also disabled the tests in the 20171101 version, as had been already done in the 2018* versions. 

It seems that the remaining errors in the Travis CI are related to `alt-ergo` and not to `frama-c` directly.

kit-ty-kate:
Looks ok. Thanks!

PR #12555, opened 2018-09-03
Added 4.07.0+32bit switch.

toots:
This switch is particularly useful for [ocaml-cross/opam-cross-windows](https://github.com/ocaml-cross/opam-cross-windows/)


kit-ty-kate:
Thanks!

PR #12556, opened 2018-09-03
Fix sedlex and ppx_tools_versioned interaction

samoht:


mseri:
Travis Times out in ocaml 4.03 to 4.05 but I don’t see errors. I think it is because it is only trying to build and test too many packages

PR #12561, opened 2018-09-04
[new release] callipyge (v0.2)

dinosaure:
CHANGES:

* New version of Callipyge (**break API**) (thx to @cfcs, @hannesm and @emillon)
* Fix tests
* Benchmarks


dinosaure:
`ppx_expect` has definitely a problem on `ocaml.4.03.0` (already noticed when I released `eqaf`).

PR #12562, opened 2018-09-04
mirage-{protocols,stack} reverse dependencies adjust upper bounds (prepare for API changes)

hannesm:
(as explained in https://github.com/mirage/mirage-protocols/pull/11) -- I hope I got most of them in here :)

PR #12564, opened 2018-09-04
restrict mirage-{protocols,stack}-lwt.X to mirage-{protocols,stack}.X

hannesm:
see #12562 for more information


kit-ty-kate:
Thanks!

PR #12566, opened 2018-09-04
Release ppx_tyre 0.4.1.

paurkedal:
We could use a bugfix release of ppx\_tyre 0.4.1, due to a issue which limits usability of the initial release.


kit-ty-kate:
Thanks!

PR #12573, opened 2018-09-07
Package submit.0.9

Niols:
### `submit.0.9`

---

---

### DIFF
--- a/packages/submit/submit.0.9/opam
+++ b/packages/submit/submit.0.9/opam
@@ -1,0 +1,34 @@
+opam-version: "2.0"
+
+maintainer: "Nicolas Jeannerod <nicolas.jeannerod@irif.fr>"
+authors: [
+  "Yann Régis-Gianas <yann.regis-gianas@irif.fr>"
+  "Ralf Treinen <ralf.treinen@irif.fr>"
+  "Nicolas Jeannerod <nicolas.jeannerod@irif.fr>"
+]
+license: "GPL3"
+
+homepage: "https://github.com/colis-anr/morbig"
+bug-reports: "https://github.com/colis-anr/morbig/issues"
+dev-repo: "git://github.com/colis-anr/morbig.git"
+
+depends: [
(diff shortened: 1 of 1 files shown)


Niols:
Hello there,

The only test that does not pass is the one on OS X. This is not surprising as we don't pass our continuous integration on OS X either. I'm not so sure about the error we get though:

          - submit.0.9:
            Failure("openssl error")

In any case, we are planning to add OS X's support for the next 1.0 version; we will see then. In the meantime, should I add in the OPAM file that OS X is not supported? And remove that flag for the next release?

mseri:
I think it's good to mark the unavailability of osx support in the opam file. Please add the relevant available stanza in the opam file.

Usually we have the content of `url` in a separate file named `url` and a `descr` file with the description of the package, see e.g. https://github.com/ocaml/opam-repository/pull/12576/files
Can you please also do that change?

Niols:
Alright. Just to confirm, I should add a `available: [ os != "osx" ]` field to my OPAM file?

I know about the separate url and descr files. I generated them, but then I used `opam publish` that merged that in one file. I was a bit surprised but I thought it had just changed. I'll fix that.

Thank you for your help!

Niols:
OK, it's not "osx" but "macos". Also, `opam publish` changed a lot since last time I used it. I found the `--split` option that does what you want. However, here is its description:

    --split
       Split the URL and description of packages into separate files
       (`url` and `descr`). This can be useful on legacy repositories, but
       is deprecated.

Should I really use-it?

mseri:
Yes please, until we switch to opam2, let's stick with the standard structure

Niols:
Thank you for your time and comments! I also realise I used `opam publish submit` which created a package named… "submit". I file a PR that, I hope, is much better, here: #12577.

mseri:
I thought it was the actual name of the package 😆

Niols:
Haha, yup. Sorry again about that :-o

PR #12577, opened 2018-09-07
Package morbig.0.9

Niols:
### `morbig.0.9`
A trustworthy parser for POSIX shell
them, and constructs a concrete syntax tree for each of them.  The

---

---

### DIFF
--- a/packages/morbig/morbig.0.9/opam
+++ b/packages/morbig/morbig.0.9/opam
@@ -1,0 +1,27 @@
+opam-version: "1.2"
+
+maintainer: "Yann Régis-Gianas <yann.regis-gianas@irif.fr>"
+authors: [
+  "Yann Régis-Gianas <yann.regis-gianas@irif.fr>"
+  "Ralf Treinen <ralf.treinen@irif.fr>"
+  "Nicolas Jeannerod <nicolas.jeannerod@irif.fr>"
+]
+license: "GPL3"
+
+homepage: "https://github.com/colis-anr/morbig"
+bug-reports: "https://github.com/colis-anr/morbig/issues"
+dev-repo: "git://github.com/colis-anr/morbig.git"
+
+available: [os != "macos" & ocaml-version >= "4.03.0"]
+depends: [
+  "ocamlbuild"           {build}
+  "menhir"               {build}
+  "yojson"
+  "ppx_deriving_yojson"
+  "visitors"
+]
+
+build: [make]
+install: [make "install"]
+remove: [make "uninstall"]
+run-test: [make "check"]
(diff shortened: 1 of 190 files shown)


mseri:
Sorry again, the package format is opam 2.0.0, but the PR should be still targeting opam 1.2 (there is an automatic process that will convert the merged files for the opam 2.0.0 branch)

Niols:
Oh, you don't have to be sorry, I feel like I am wasting your time here. I should never have gone for opam-publish 2.0 ;-)

Is it better now?

packages/morbig/morbig.0.9/opam
```diff
@@ -0,0 +1,28 @@
...
+]
+license: "GPL3"
+
+homepage: "https://github.com/colis-anr/morbig"
+bug-reports: "https://github.com/colis-anr/morbig/issues"
+dev-repo: "git://github.com/colis-anr/morbig.git"
+
+available: [os != "macos"]
+depends: [
+  "ocaml"                {build & >= "4.03.0"}
```
mseri: This does not work in opam 1.2, you need to modify availabe as follows:
```
available: [os != "macos" & ocaml-version >= "4.03.0"]
```
or something like that. I can't remember if opam 1.2 wants "osx", "darwin" or "macos", sorry.

Niols:
How come my PR still passes the tests for 1.2 by the way?

Anyway, I've looked in the repo, it seems to be "macos" and I've made the right changes.

What do you think of the url file? Is it OK to have the two checksums or is it also 2.0 only?

mseri:
I thought the multiple checksums where fine, but after looking at the documentation I think they are a thing of opam2 as well. Thanks for spotting it.

The CI is already using opam2-based images, so I think it may be a bit weak with respect to that. Probably camelus will complain.

I really can't wait to move to opam2 to be honest :)

Niols:
Uh. Turns out camelus complains about opam-1.2 features :-o
Also, when I take a look at the files in the repository, they seem to have been updated to use the 2.0 scheme. Am I really supposed to send a PR using the 1.2 scheme?

Niols:
Maybe we don't have to wait for OPAM 2.0 anymore. Maybe it's already here, it the middle of us. Maybe we have been blind for years! :-o

*(can't wait for it either)*

mseri:
The problem is that your PR is not against `master` but against the `2.0.0` branch. You need to point it to the other branch

Niols:
Indeed, thank you for the remark, I'll change all that.

Niols:
I tried to use GitHub's "change branch" mechanism, but it appeared it confused the CI that tries to build the whole world. I'll guess I'll start again from scratch (we'll manage, eventually!)

Niols:
Here it is: #12579. Sorry for all the trouble.

PR #12581, opened 2018-09-07
ocamlbuild: constraint ocaml-version to 4.07

rjbou:
As packaged ocamlbuilds don't compile with 4.08.0, this PR set constraints on ocaml-versions.

/cc @gasche

### DIFF
--- a/packages/ocamlbuild/ocamlbuild.0.11.0/opam
+++ b/packages/ocamlbuild/ocamlbuild.0.11.0/opam
@@ -25,7 +25,7 @@ build: [
   [make "check-if-preinstalled" "all" "opam-install"]
 ]
 
-available: [ocaml-version >= "4.03"]
+available: [ocaml-version >= "4.03" & ocaml-version < "4.08.0"]
 depends: [ ]
 conflicts: [
   "base-ocamlbuild"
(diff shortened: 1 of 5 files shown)


gasche:
Fine with me! Out of curiosity, why are you testing 4.08 builds of the opam ecosystem? Is it for opam2 development?

rjbou:
No, i just found it by coincidence :)
While doing some unrelated tests, I created an empty switch then installed a package, so the whole needed packages where installed. As opam installs the last version of a package, [even ocaml](https://github.com/ocaml/opam/issues/3510), it installed ocaml 4.08 and ocamlbuild, and then, error!
At least, with this PR, it won't block users with an error unrelated to what they aim to do.

kit-ty-kate:
I would rather have `& ocaml-version < "4.08.0"` than the current `& ocaml-version <= "4.07.0"`

gasche:
Thanks @kit-ty-kate! That's a excellent point (sorry for not thinking of it myself), because we have plans for a 4.07.1.

rjbou:
@kit-ty-kate PR updated!

kit-ty-kate:
oops sorry I should've merged this PR earlier. Thanks a lot!

PR #12585, opened 2018-09-08
+conex{-nocrypto}.0.10.1

hannesm:
the only change in contrast to 0.10.0 is that a non-empty LICENSE.md file was added (2 clause BSD license, as advertised in the opam files) //cc @avsm


kit-ty-kate:
Thanks!

PR #12586, opened 2018-09-08
ocamlbuild.0.13.1

gasche:
OCamlbuild 0.13.0 contains new features (`ppopt(..)` and `ppxopt(...)` flags,
ocamlbuild options) and bugfixes (to pack production, for 4.08+dev support).

- ocaml/ocamlbuild#45, ocaml/ocamlbuild#190: add ppopt(arg) and ppxopt(package,arg) when -use-ocamlfind
  (Gabriel Scherer, review by whitequark,
   request by Gabriel Scherer, Gabriel Radanne and Pavel Argentov)

- ocaml/ocamlbuild#268, ocaml/ocamlbuild#269: add flag support for some ocamlmklib options:
  custom, debug, failsafe, linkall, ccopt(..), cclib(..), rpath(..), ldopt(..)
  (Gabriel Scherer, report by Hannes Mehnert, review by whitequark)

* ocaml/ocamlbuild#272: add the "link" tag to "pack" actions
  Instead of a separate category, "pack" is now another form of linking
  like "program", "library", "toplevel" and "output_obj". This fixes
  the issue that package(...) tags where not passed at pack-production
  time, spotted by Jérémie Dimino. More generally, this extends
  the meaning of all "link" flags to "pack", which seems to be the correct
  behavior for all the rules we inspected.
  (Gabriel Scherer, original issue diagnosis by Jérémie Dimino)

- ocaml/ocamlbuild#278: typo fixes in the manual
  (Xinzhe Yang)

- ocaml/ocamlbuild#282: fix compilation with trunk OCaml (4.08+dev)
  (Xavier Clerc and Nandor Licker)


mseri:
All the failures seems to be independent of this. I've seen not updated ast for 4.06, missing support for safe-string and network probles. I think this is safe to merge. I'll leave it open if somebody wants to double check first

gasche:
I agree, and/but I started working on a small PR to version-bound some of the unsafe-string revdeps appropriately: 

  https://github.com/ocaml/opam-repository/compare/master...gasche:ocamlbuild-revdep-fixes

soon in a PR near you!

fdopen:
> All the failures seems to be independent of this. I've seen not updated ast for 4.06, missing support for safe-string and network probles. 

Some failures are new and don't seem to be related to safe string, e.g:
https://ci.ocamllabs.io/log/saved/docker-run-af7d9f4b635e5949f2984b87fb1467ef/c4298e0875728ebbc7d36ed691591db38d7c1095

gasche:
@fdopen thanks! I'll have a look. Out of curiosity, how did you spot this one? Are there more when it comes from?

gasche:
So, this is indeed a regression due to a change of behavior in ocamlbuild, more precisely the addition of the "link" tag to "pack" actions. This package is using `true: linkpkg, linkall`, and as a result `ocamlfind` tries to link .cma archives when producing a `-pack` archive.

I think that the package is wrong here (I don't see a reason to use `true: linkpkg`), it should probably get fixed, but it could also be nice to have protection against this in ocamlbuild upstream -- by disabling `linkpkg` in `pack` actions.

I'm happy to do another round and replace this with a 0.13.1 release, but it would be nicer if we could have a reasonable sense of whether other regressions are waiting in the CI results, lost in the noise of `-unsafe-string` packages.

gasche:
I just rebased the present PR with a 0.13.1 release that fixes the webidl build (and wasm, which was also affected) without requiring upstream package changes. I will try to look at the revdeps more carefully this time to see if more packages are affected.

fdopen:
This sounds still like the previous error:
https://ci.ocamllabs.io/log/saved/docker-run-6715970e15d0e6eebad9a62dc7a8d074/7be5f93bfa6829ee669fc13d0d327d76411b576b
> Error: /home/opam/.opam/4.07/lib/ocaml/bigarray.cma is not a bytecode object file

gasche:
This is weird, I thought I had checked that `wasm` builds on ocamlbuild master...

Anyway, I won't have time to work on this before early October, so the release will have to wait. Sorry for the delay.

avsm:
I'll close this until @gasche gets a chance to revisit in October.

PR #12588, opened 2018-09-09
Ocamlbuild revdep fixes

gasche:
metadata update for the first N=6 packages in ocamlbuild reverse dependencies that require an ocaml version bound, mostly because they need `-unsafe-string`.

### DIFF
--- a/packages/argon2/argon2.0.1/opam
+++ b/packages/argon2/argon2.0.1/opam
@@ -20,4 +20,4 @@ depends: [
   "result"
   "ppx_deriving" {>= "2.0"}
 ]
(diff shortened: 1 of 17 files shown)


gasche:
Re. camelus: I won't fix all opam files (they were missing fields before, it's just fine), but I will look at the argon2 issues.

hannesm:
@gasche argon2 issue is your usage of `&&`, which should be `&` afaict

gasche:
Indeed! I just fixed and rebased, thanks @hannesm.

gasche:
I checked the Debian failures in the CI, they come from GTK/cairo packaging issues that are beyond my control. Merging.

PR #12590, opened 2018-09-09
Add gstreamer 0.3.0

toots:


toots:
The debian build errors seem to be issues with the CI system to me. The old `mm` dependency is outdated. I'd suggest to leave it or remove the package. Thks for your time reviewing this!

mseri:
That's fine, we can bound mm as incompatible with ocaml 4.06+ in separate PR

PR #12591, opened 2018-09-09
Add ffmpeg.0.2.0

toots:


kit-ty-kate:
The opam file needs to be changed accordingly to the opam file in the source repository: https://github.com/savonet/ocaml-ffmpeg/blob/master/ffmpeg.opam

kit-ty-kate:
ok done.

toots:
Thanks!

kit-ty-kate:
looks good now. Thanks!

PR #12592, opened 2018-09-10
Add hvsock.1.0.3

djs55:

### DIFF
--- a/packages/hvsock/hvsock.1.0.3/opam
+++ b/packages/hvsock/hvsock.1.0.3/opam
@@ -1,0 +1,36 @@
+opam-version: "1.2"
+maintainer: "dave@recoil.org"
+authors: [ "David Scott" "Rolf Neugebauer" "Anil Madhavapeddy" "Simon Ferquel"]
+license: "ISC"
+homepage: "https://github.com/mirage/ocaml-hvsock"
+dev-repo: "https://github.com/mirage/ocaml-hvsock.git"
+bug-reports: "https://github.com/mirage/ocaml-hvsock/issues"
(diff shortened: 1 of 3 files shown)


packages/hvsock/hvsock.1.0.3/opam
```diff
@@ -0,0 +1,36 @@
...
+maintainer: "dave@recoil.org"
+authors: [ "David Scott" "Rolf Neugebauer" "Anil Madhavapeddy" "Simon Ferquel"]
+license: "ISC"
+homepage: "https://github.com/mirage/ocaml-hvsock"
+dev-repo: "https://github.com/mirage/ocaml-hvsock.git"
+bug-reports: "https://github.com/mirage/ocaml-hvsock/issues"
+doc: "https://mirage.github.io/ocaml-hvsock"
+
+build: [
+  [ "jbuilder" "subst"] {pinned}
```
kit-ty-kate: if you are using jbuilder you need to pass `"-p" name` to the subst command. See https://github.com/ocaml/opam-repository/pull/12279

packages/hvsock/hvsock.1.0.3/opam
```diff
@@ -0,0 +1,36 @@
...
+  [ "jbuilder" "build" "-p" name "-j" jobs ]
+]
+
+build-test:[
+  [ "jbuilder" "subst"] {pinned}
```
kit-ty-kate: Same here, although I don't think you need to call subst for the tests

djs55:
@kit-ty-kate thanks for the info! I've also found some bugs in this particular release so I'll close this PR and make a new one after making another point release :)

PR #12596, opened 2018-09-10
stdcompat-6

thierry-martinez:
- Support for OCaml 4.08.0

- Support VPATH build, i.e. when configure is executed in a build directory
  distinct from the source directory

- Lexing.new_line

- Bigarray: only available from 4.02.0. From 4.02.0 and before 4.07.0,
  --no-alias-deps is used to allow the alias to appear in Stdcompat without
  requiring the library to be linked. Bigarray library should be linked if it is
  effectively used.

- Fix implementations with --disable-magic

- Fix license conflict: The project was intended to be under BSD license
  since the beginning, and a LICENSE file was provided accordingly.
  However, automake generated a COPYING file with the GPLv3 by default. The
  COPYING file now contains the BSD license, and the LICENSE file is removed.
  (reported by Török Edwin,
   https://github.com/thierry-martinez/stdcompat/issues/5)

- Fix auto-generated interfaces for Hashtbl.MakeSeeded

- Exceptions are reexported again. They were missing in auto-generated
  interfaces.

- min_binding_opt and max_binding_opt are not redefined if OCaml >=4.05.

- Array.of_seq is redefined in OCaml 4.07.0 to circumvent a bug in the
  implementation of the standard library. See:
    - https://caml.inria.fr/mantis/view.php?id=7820
    - https://github.com/ocaml/ocaml/pull/1897

### DIFF
--- a/packages/stdcompat/stdcompat.6/descr
+++ b/packages/stdcompat/stdcompat.6/descr
@@ -1,0 +1,1 @@
+Compatibility module for OCaml standard library allowing programs to use some recent additions to the OCaml standard library while preserving the ability to be compiled on former versions of OCaml.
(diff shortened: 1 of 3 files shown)


kit-ty-kate:
Thanks!

PR #12598, opened 2018-09-11
Added liquidsoap.1.3.4

toots:


toots:
The package isn't broken anymore, that was due to the latest `duppy` not being merged yet.

PR #12602, opened 2018-09-11
Add (safe-string compatible) ledit 2.04 and make ledit 2.03 explicitly not safe string

pirbo:


packages/ledit/ledit.2.04/opam
```diff
@@ -0,0 +1,16 @@
+opam-version: "1.2"
+authors: "Daniel de Rauglaudre"
+homepage:"http://pauillac.inria.fr/~ddr/ledit/"
+maintainer: "Pierre Boutillier <pierre.boutillier@laposte.net>"
+
+build: make
+depends: ["camlp5"]
+install: [
```
kit-ty-kate: There seem to be a problem here. What does `make install` do? There should be a `remove` rule somewhere if there is an `install` rule (though you have an `install` rule AND a `.install` file which seems weird.

pirbo:
You're right. I just copied/pasted the build system of ledit.2.03 for ledit.2.04 but it was ugly. I cleaned things u a little now...

kit-ty-kate:
Looks ok. Thanks!

PR #12606, opened 2018-09-12
cppo 1.6.5

mjambon:
Compared to the package for cppo 1.6.4, we removed `"base-bytes"` from the list of dependencies in the `opam` file (explained at https://github.com/mjambon/cppo/pull/57).


kit-ty-kate:
Revdeps failures look mostly unrelated. Thanks!

mjambon:
Thank you, @kit-ty-kate !

PR #12610, opened 2018-09-13
Add piqilib-0.6.14 and piqi-0.7.6

alavrik:


kit-ty-kate:
Could you add the `{build}` attribute to `ocamlfind` as it seems to be only needed for building and installation and not as a runtime dependency.

mseri:
Re-running ocaml 4.04 test, the ocaml 4.06 failure does not depend on this PR:
```
#=== ERROR while compiling camlp4.4.06+system =================================#
# context     2.0.0 | linux/x86_64 | ocaml-system.4.06.1 | file:///home/travis/build/ocaml/opam-repository
# path        ~/.opam/ocaml-system/.opam-switch/build/camlp4.4.06+system
# command     ~/.opam/opam-init/hooks/sandbox.sh build sh ./make.sh
# exit-code   1
# env-file    ~/.opam/log/camlp4-12940-d18c27.env
# output-file ~/.opam/log/camlp4-12940-d18c27.out
### output ###
# preexisting installation rather than compile one from scratch).
# [...]
# If you cannot install a system-wide camlp4, then you can still use OCaml
# 4.02 by switching to a local installation via `opam switch 4.02.0`.
# 
# Here are some installation instructions for camlp4 if you obtained OCaml
# via the OPAM binary packages:
# 
# http://software.opensuse.org/download.html?project=home%3Aocaml&package=ocaml
# 
# * Debian/Ubuntu: sudo apt-get install camlp4-extra
# * RHEL/CentOS/Fedora: sudo yum install ocaml-camlp4
# 
```

@avsm is this failure a known one? I don't think I had seen it on the opam 1.2 branch

alavrik:
@mseri  @kit-ty-kate should we just go ahead and merge it? It looks like there's nothing wrong with the packages themselves. There are some problems with CI checks, but I am not sure what I can do at this point. I don't think I can even rerun them checks myself to see if the problems went away.

mseri:
Yes, sorry it took so long

PR #12612, opened 2018-09-13
Solo5: Add constraints for 0.4.0 release

mato:
As part of the 0.4.0 release, the Solo5 packages (solo5-kernel-*) will
be renamed. Add the appropriate constraints to existing published
packages that depend/make use of the old package names.

See Solo5/solo5#274 for details.


hannesm:
thanks, all CI are green, merging!

PR #12614, opened 2018-09-13
[new release] lpd (1.2.2)

Chris00:
CHANGES:

- Use Dune to compile.
- Fix safe-string issues so it works with OCaml ≥ 4.06


kit-ty-kate:
Thanks!

PR #12615, opened 2018-09-13
fix packages that have `setup.ml` from an old oasis

damiendoligez:
The packages `forkwork.0.3.1, forkwork.0.3.2, genspir, indexmap, libvhd, lwt-binio.0.2.1, nlopt-ocaml.0.4, nlopt-ocaml.0.5, nlopt-ocaml.0.5.1, pci-db, qrencode, rtime, should, sqlite3EZ, udunits.0.1.1, udunits.0.2.0, zbar` all have this in common:

- they use oasis
- they have a `setup.ml` file generated by an old version of oasis, which was broken by the switch to safe strings in 4.06.0
- they are otherwise not broken by safe strings

So we can fix them by changing the build instructions to regenerate `setup.ml` before building as before. This fix doesn't need any patch file or upstream change. The only drawback is that it introduces a build-time dependency on `oasis` (not a big deal IMO).

### DIFF
--- a/packages/forkwork/forkwork.0.3.1/opam
+++ b/packages/forkwork/forkwork.0.3.1/opam
@@ -1,6 +1,7 @@
 opam-version: "1.2"
 maintainer: "mlin@mlin.net"
 build: [
+  ["oasis" "setup"]
   ["ocaml" "setup.ml" "-configure" "--prefix" prefix]
   ["ocaml" "setup.ml" "-build"]
 ]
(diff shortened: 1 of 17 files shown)


damiendoligez:
Sorry, I forgot to lint before sending the PR. I'll add a commit to make camelus happy.

dbuenzli:
LGM. Note that `rtime` is [formally](http://erratique.ch/software/rtime) deprecated but I never found a way to express that in opam (see https://github.com/ocaml/opam/issues/2046). As far as I'm concerned it could be removed from the repo but I think it's against policies.

damiendoligez:
AFAICT, all the CI failures are caused by missing external dependencies.

kit-ty-kate:
Looks good. Thanks!

PR #12617, opened 2018-09-13
pcap-format require ppx_cstruct > 0

hannesm:
as discovered in https://travis-ci.org/mirage/mirage-protocols/jobs/428313682 -- only affects master and most recent release (due to switch to jbuilder).


hannesm:
PR for ocaml-pcap repo at https://github.com/mirage/ocaml-pcap/pull/30

PR #12623, opened 2018-09-14
Constrain all packages non-safe-string-ready to OCaml < 4.06

kit-ty-kate:


gasche:
Thanks a lot! I didn't remark this before, and it will be very useful for ocamlbuild revdeps checking.

PR #12633, opened 2018-09-15
Fix liquidsoap 1.3.4 depopts

toots:
This PR fixes two issues with `gstreamer` missing and `bjack` conflict version. Sorry for the oversight in the previous PR.


kit-ty-kate:
Thanks!

PR #12646, opened 2018-09-18
delete bleeding edge 4.08 compilers as they might be selected by default

avsm:
temporary solution for https://github.com/ocaml/opam/issues/3510

these bleeding edge compilers will move into a new remote so they
are still accessible to advanced users, but this PR just removes
them for now as it is opam 2.0.0 release day :)


mseri:
Let's merge then. Will you need to remove the other `+fature` variants for older compilers as well before the merge?

mseri:
Also, do you need to remove https://github.com/ocaml/opam-repository/tree/2.0.0/packages/ocaml/ocaml.4.08.0 (and `ocaml-compiler.4.08.0*` and `ocaml-compiler-libs-4.08.0*`) now that there is no more an `ocaml-variant` for `ocaml` `4.08.0`?

mseri:
Done. I'll double check with camelus and if it's ok merge

avsm:
thanks @mseri !

PR #12648, opened 2018-09-18
opam 2.0.0

rjbou:
Url changes


AltGr:
Actually, since these are the same files, let's not trigger rebuilds by changing the checksums.

PR #12649, opened 2018-09-18
opam admin upgrade run

kit-ty-kate:
cc @rjbou @AltGr 

Most of the change is removing parenthesis around ors but some change are about changing arch name  and some other change for `opam-publish.2.0.0~beta`. Could you both check that everything is ok?


AltGr:
Ok with me, but I'd rather keep the parentheses part out. I reviewed it, and just keeping the lines containing the word `arch` would do (except for [this line](https://github.com/ocaml/opam-repository/pull/12649/files#diff-daeedde09dddb3df0fabb5e95ca7ee63R14), which isn't very important anyway)

samoht:
why do the parenthesis got removed?

PR #12650, opened 2018-09-19
added conf-git.1.0

UnixJunkie:


UnixJunkie:
Tests say it works everywhere (what a happy surprise, for once).
I used the conf-which package as a template, mostly.

packages/conf-git/conf-git.1.0/opam
```diff
@@ -0,0 +1,22 @@
+opam-version: "2.0"
+maintainer: "unixjunkie@sdf.org"
+homepage: "https://git-scm.com"
+authors: "Linus Torvalds"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+dev-repo: "git+https://github.com/ocaml/opam-repository.git"
+license: "GPL-2+"
+build: [["which" "git"]]
```
mseri: Does this mean that it depends on `conf-which`?

kit-ty-kate: yes.

kit-ty-kate: fixed.

UnixJunkie:
Yes, it depends on conf-which.
Nice catch and thanks for the fix.

PR #12660, opened 2018-09-20
Add ocamlfind dependency to ppxfind

diml:
`ppxfind` uses `findlib.dynload` so it should depend on `ocamlfind` explicitly.


diml:
I added a conflict with dune 1.2.{0,1}, there is a bug in these versions that prevents ppxfind to build. It's fixed in the master of Dune.

diml:
The travis failure looks unrelated. The rest is all green.

mseri:
The OS X failure affects all packages. Merging. Thanks!

PR #12661, opened 2018-09-20
jingoo: upgrade to version 1.2.19

tategakibunko:


pmetzger:
Just an FYI, but it's rather difficult for someone reading this commit message to know _what_ was upgraded to 1.2.19. It would be better if the commit messages said:

```
jingoo: upgrade to version 1.2.19
```

tategakibunko:
Thanks for your advice!

mseri:
The `url` file is still using the opam 1.2 syntax, please update that as well

tategakibunko:
Thanks!

PR #12665, opened 2018-09-21
added packages/conf-g++/conf-g++.1.0

UnixJunkie:


UnixJunkie:
Looks like it is working.
Someone might want to edit the PR and fill in the nixos line, if they know what it should be.

mseri:
Thanks. I'll leave it open a few days in case somebody knows what to put there

kit-ty-kate:
On nixos it seems to be `gcc` (https://nixos.org/nixos/packages.html#gcc)

However I'm not a fan of enforcing `g++` as the default `c++` compiler. Why not just call the package `conf-g++` ?

ygrek:
this package looks confusing indeed. E.g. on debian there is already alternatives system, and there is `c++` symlink available that should be used instead of hardcoding g++ (unless g++ specifically is required for build of course)

UnixJunkie:
I fixed all suggested things.

packages/conf-g++/conf-g++.1.0/opam
```diff
@@ -0,0 +1,22 @@
+opam-version: "2.0"
+maintainer: "unixjunkie@sdf.org"
+authors: "Francois Berenger"
+homepage: "https://github.com/ocaml/opam-repository"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+dev-repo: "git+https://github.com/ocaml/opam-repository.git"
```
dra27: There's no need for `dev-repo` for conf packages.

UnixJunkie: but opam lint gives me a warning if I remove it

kit-ty-kate:
Thanks a lot!

packages/conf-g++/conf-g++.1.0/opam
dra27: Really? What version of opam are you running. `opam lint` does not give a warning for no `dev-repo` if there's no `url`.

dra27: I’m not an opam-repository maintainer so, no, I couldn’t edit your PR...

dra27: Incidentally, the change to `opam lint` was made for the release candidate 8 months ago - you should upgrade!

PR #12670, opened 2018-09-21
add zenon.0.8.4

damiendoligez:


packages/zenon/zenon.0.8.4/opam
```diff
@@ -0,0 +1,31 @@
+opam-version: "2.0"
+homepage: "http://zenon-prover.org/"
+dev-repo: "https://github.com/zenon-prover/zenon.git"
```
mseri: Please use `git+https://`

mseri: I'll change it

PR #12671, opened 2018-09-21
add compiler packages for ocaml 4.07.1+rc1

damiendoligez:


mseri:
Wouldn’t this have the same problem of ocaml 4.08.0, with opam trying to upgrade the compiler to the rc1 for all users?

avsm:
Manually tested.

PR #12673, opened 2018-09-21
nocrypto.0.5.4-1 works on bsd with opam2 (which uses gpatch instead of bsd patch)

hannesm:
now that opam2 is default, which uses `gpatch` on *BSD (see https://github.com/ocaml/opam/pull/3444), the patches for nocrypto 0.5.4-1 apply nicely on these OS, /cc @copy
#12062 #12252 #12253 #12286


hannesm:
CI failure is unrelated, and there's no CI system for FreeBSD/OpenBSD yet, merging

PR #12675, opened 2018-09-22
Adding ppx_relit 0.1 and relit_helper 0.1

charlesetc:
Hi — I'm hoping to expose these two packages, both stored in the same .tar at the same github url. Do you think this will work?
Thanks very much!
Charles

### DIFF
--- a/packages/ppx_relit/ppx_relit.0.1/opam
+++ b/packages/ppx_relit/ppx_relit.0.1/opam
@@ -26,5 +26,5 @@ depends: [
 synopsis: "An implementation of Typed Literal Macros for Reason"
 url {
   src: "https://github.com/cyrus-/relit/archive/0.1.tar.gz"
-  checksum: "md5=7a684c44cc57cacc8d0a3d741a7907ad"
+  checksum: "md5=d5332e3cf643f75bfe84f778d619b167"
 }
--- a/packages/ppx_relit/ppx_relit.0.1/opam
+++ b/packages/ppx_relit/ppx_relit.0.1/opam
@@ -21,7 +21,7 @@ depends: [
   "base64"
   "ppxlib"
   "base-unix"
-  "ocamlbuild"
+  "ocamlbuild" {test}
 ]
 synopsis: "An implementation of Typed Literal Macros for Reason"
 url {
--- a/packages/ppx_relit/ppx_relit.0.1/opam
+++ b/packages/ppx_relit/ppx_relit.0.1/opam
@@ -22,10 +22,11 @@ depends: [
   "ppxlib"
   "base-unix"
   "ocamlbuild" {test}
+  "ocaml"      {= "4.06.1"}
   "relit_helper"
 ]
 synopsis: "An implementation of Typed Literal Macros for Reason"
 url {
   src: "https://github.com/cyrus-/relit/archive/0.1.tar.gz"
-  checksum: "md5=d5332e3cf643f75bfe84f778d619b167"
+  checksum: "md5=bb4d8ed18bdf1a5471ac7a7d6ee8b4dd"
 }
--- a/packages/ppx_relit/ppx_relit.0.1/opam
+++ b/packages/ppx_relit/ppx_relit.0.1/opam
@@ -1,4 +1,4 @@
-opam-version: "1.2"
+opam-version: "2.0"
 name: "ppx_relit"
 version: "0.1"
 maintainer: "Charles Chamberlain <charlespipin@gmail.com>"
@@ -9,7 +9,7 @@ authors: [
 homepage: "https://github.com/cyrus-/relit"
 bug-reports: "https://github.com/cyrus-/relit/issues"
 license: "MIT"
-dev-repo: "https://github.com/cyrus-/relit"
+dev-repo: "git://github.com/cyrus-/relit.git"
 build: [
   ["dune" "build" "-p" "ppx_relit"]
 ]
--- a/packages/ppx_relit/ppx_relit.0.1/opam
+++ b/packages/ppx_relit/ppx_relit.0.1/opam
@@ -22,6 +22,7 @@ depends: [
   "ppxlib"
   "base-unix"
   "ocamlbuild" {test}
+  "relit_helper"
 ]
 synopsis: "An implementation of Typed Literal Macros for Reason"
 url {
--- a/packages/ppx_relit/ppx_relit.0.1/opam
+++ b/packages/ppx_relit/ppx_relit.0.1/opam
@@ -1,0 +1,23 @@
+opam-version: "1.2"
+name: "ppx_relit"
+version: "0.1"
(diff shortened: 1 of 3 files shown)


packages/ppx_relit/ppx_relit.0.1/opam
```diff
@@ -0,0 +1,23 @@
...
+authors: [
+  "Charles Chamberlain <charlespipin@gmail.com"
+  "Cyrus Omar <cyrus.omar@gmail.com"
+]
+homepage: "https://github.com/cyrus-/relit"
+bug-reports: "https://github.com/cyrus-/relit/issues"
+license: "MIT"
+dev-repo: "https://github.com/cyrus-/relit"
+build: [
+  ["jbuilder" "build"]
```
mseri: Yous should add `”-p”  name` To make sure it is building only the package itself. Also the install and remove will be automatically generated by jbuilder, so you should delete them below

mseri: This will allow you to also drop the dependency on ocamlfind that seems to be used only to remove the package

packages/relit_helper/relit_helper.0.1/opam
```diff
@@ -0,0 +1,22 @@
...
+authors: [
+  "Charles Chamberlain <charlespipin@gmail.com>"
+  "Cyrus Omar <cyrus.omar@gmail.com>"
+]
+homepage: "https://github.com/cyrus-/relit"
+bug-reports: "https://github.com/cyrus-/relit/issues"
+license: "MIT"
+dev-repo: "https://github.com/cyrus-/relit"
+build: [
+  ["jbuilder" "build"]
```
mseri: See the comments on the other opam file

packages/ppx_relit/ppx_relit.0.1/opam
```diff
@@ -0,0 +1,23 @@
...
+]
+install: ["jbuilder" "install" "ppx_relit"]
+remove: ["ocamlfind" "remove" "project"]
+depends: "ocamlfind"
```
mseri: This depends at least on jbuilder, but looking at the jbuild file there are a lot of dependencies missing: 
https://github.com/cyrus-/relit/blob/master/ppx_relit/jbuild

packages/relit_helper/relit_helper.0.1/opam
```diff
@@ -0,0 +1,22 @@
...
+]
+install: ["jbuilder" "install" "relit_helper"]
+remove: ["ocamlfind" "remove" "project"]
+depends: "ocamlfind"
```
mseri: Also this package needs to fix the dependencies

mseri:
Thanks. Can you do the cleanup suggested in the comments? To fix the dependencies you could move to dune instead, the latest version is able to generate the correct list for you

charlesetc:
Thanks very much! I'll address your comments and move to the latest dune

charlesetc:
I think I've added the appropriate dependencies: thanks again! I've also added a relit-reason package which is a modified version of reason that supports the syntax relit uses. Have I done this properly?

mseri:
Thanks. Looks good! Let’s wait for the CI to finish, but it should be fine now

charlesetc:
:disappointed: looks like the last check failed. Do you know why or how I could debug that?

charlesetc:
Looks like it's failing since bumping to opam-version 2.0, it doesn't have m4 available? Is there a way I can require that as a dependency?

avsm:
The CI is currently very broken, but a fix has been deployed and is rebuilding :)

packages/ppx_relit/ppx_relit.0.1/opam
```diff
@@ -0,0 +1,30 @@
...
+]
+depends: [
+  "dune"
+  "ocamlfind"
+  "ocaml-migrate-parsetree"
+  "relit-reason"
+  "base64"
+  "ppxlib"
+  "base-unix"
+  "ocamlbuild"
```
avsm: does this still depend on ocamlbuild? Looks like it's been ported to Dune

charlesetc: Ah good point, we use ocamlbuild for some tests but it shouldn't be needed to build/use it.

mseri: If you use it in tests you can still mark it as a test dependency by adding {test} next to it

charlesetc:
Do you know if this will be fixed by tomorrow? Totally understand if not; it'd be good to get a sense of how we should plan things.

mseri:
This is a new package, the ci is not fundamental, let’s ignore it for the moment. There is still something wrong, the travis builds are failing with

```
#=== ERROR while compiling relit-reason.0.0.1 =================================#
# context     2.0.0 | linux/x86_64 | ocaml-base-compiler.4.05.0 | file:///home/travis/build/ocaml/opam-repository
# path        ~/.opam/4.05.0/.opam-switch/build/relit-reason.0.0.1
# command     ~/.opam/opam-init/hooks/sandbox.sh build dune build -p relit-reason -j 1
# exit-code   1
# env-file    ~/.opam/log/relit-reason-11093-075801.env
# output-file ~/.opam/log/relit-reason-11093-075801.out
### output ###
# Error: I don't know about package relit-reason (passed through --only-packages/--release)

```

charlesetc:
Ah good catch, thanks! I'm hoping I've fixed it but am not sure: where did you see that error?

charlesetc:
Okay well it looks like the packages are now being found and unpackaged and it's a different type error for each ocaml version. I'll do some thinking and see if this can be fixed. Thanks for your help!

mseri:
I had a brief look at the errors, ocaml 4.05 and 4.06 seem to be similar. I think fixing those and ocaml 4.07 may be easier. On the other hand I think you need 4.02 if you want to be compatible with bucklescript, right? Maybe in this case using multiple dune workspaces targeting all the compilers, and cppo to specialize some fixes could help, but it is going to be quite an amout of work.

Given that this is the first release, if you need it in time for ICFP, maybe you can fix it for one specific compiler, and ask for that in the depends bounds. Then you will have more time to fix it for all compilers, but you will have the package available and tested for the compiler version you have chosen.

packages/ppx_relit/ppx_relit.0.1/opam
```diff
@@ -22,10 +22,11 @@ depends: [
   "ppxlib"
   "base-unix"
   "ocamlbuild" {test}
+  "ocaml"      {= "4.06.1"}
```
mseri: I think you can relax kt to >= 4.06.0 & < 4.07.0 but I am not picky about that

charlesetc:
It seems reason doesn't support 4.07 yet and 4.02 is so long ago that 4.06 seems like a good compromise for now. Thanks again for your help! :tada:

mseri:
Thank you!

Issue #12550, opened 2018-09-02
ocaml-http does not build with ocaml-base-compiler.4.07.0


sacerdot:
# File "http_misc.ml", line 126, characters 29-32:
# Error: This expression has type string but an expression was expected of type
#          bytes
# make: *** [Makefile:74: http_misc.cmo] Error 2

<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
┌─ The following actions failed
│ λ build ocaml-http 0.1.5
└─ 
╶─ No changes have been performed

gares:
That must be because `-safe-string` is on by default on that compiler.
Maybe it can be switched off in the opam file, if the makefile reads ENV variables...

mseri:
Switching off safe-string locally leads often to incompatible interface error though. It would be better to update the pckge

avsm:
Thanks for the report! I've added a constraint in #12679 to stop trying to build it on an unsupported version of ocaml; the package seems unmaintained so you may want something like cohttp or httpaf instead.  We can't fix this directly in opam-repository unfortunately, so closing this for now.

PR #12679, opened 2018-09-23
ocaml-http: does not compile with safe-string on by default

avsm:
see #12550


mseri:
The ci failure seem unrelated. Thanks for the bounds

PR #12682, opened 2018-09-23
new release: mirage{-runtime,-types,-types-lwt}.3.2.0

hannesm:
### 3.2.0 (2018-09-23)

* adapt to solo5 0.4.0 changes  (by @mato)
Upgrading from Mirage 3.1.x or earlier

Due to conflicting packages, opam will not upgrade mirage to version 3.2.0 or newer if a version of mirage-solo5 older than 0.4.0 is installed in the switch. To perform the upgrade you must run `opam upgrade mirage` explicitly.

Changes required to rebuild and run ukvm unikernels

As of Solo5 0.4.0, the ukvm target has been renamed to hvt. If you are working out of an existing, dirty, source tree, you should initially run:

```
mirage configure -t hvt
mirage clean
mirage configure -t hvt
```

and then proceed as normal. If you are working with a clean source tree, then simply configuring with the new hvt target is sufficient:

`mirage configure -t hvt`

Note that the build products have changed:

The unikernel binary is now named `<unikernel>.hvt`,
the `ukvm-bin` binary is now named `solo5-hvt`.

* adapt to mirage-protocols, mirage-stack, tcpip changes (by @hannesm)

This is a breaking change: mirage 3.2.0 requires mirage-protocols 1.4.0, mirage-stack 1.3.0, and tcpip 3.5.0 to work (charru-client-mirage 0.10 and mirage-qubes-ipv4 0.6 are adapted to the changes).  An older mirage won't be able to use these new libraries correctly.  Conflicts were introduced in the opam-repository.

In more detail,  direct and socket stack initialisation changed, which is automatically generated by the mirage tool for each unikernel (as part of `main.ml`).  A record was built up, which is no longer needed.

Several unneeded type aliases were removed:
  `netif` from Mirage_protocols.ETHIF
  `ethif` and `prefix` from Mirage_protocols.IP
  `ip` from Mirage_protocols.{UDP,TCP}
  `netif` and `'netif config` from Mirage_stack.V4
  `'netif stackv4_config` and `socket_stack_config` in Mirage_stack

* squash unnecessary warning from `mirage build` (by @mato)

### DIFF
--- a/packages/mirage-runtime/mirage-runtime.3.2.0/opam
+++ b/packages/mirage-runtime/mirage-runtime.3.2.0/opam
@@ -1,0 +1,52 @@
+opam-version: "2.0"
+maintainer: ["anil@recoil.org" "thomas@gazagnaire.org"]
+authors: [
+  "Thomas Gazagnaire"
+  "Anil Madhavapeddy"
+  "Gabriel Radanne"
+  "Mindy Preston"
+  "Thomas Leonard"
+  "Nicolas Ojeda Bar"
+  "Dave Scott"
+  "David Kaloper"
+  "Hannes Mehnert"
+  "Richard Mortier"
+]
+tags: ["org:mirage" "org:xapi-project"]
+homepage: "https://github.com/mirage/mirage"
+doc: "https://mirage.github.io/mirage/"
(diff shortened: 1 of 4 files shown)


mseri:
Looks good thanks

PR #12700, opened 2018-09-25
Detect graphics.cmi in system compiler

dra27:
Re-run of #11946.

cc @damiendoligez, @gasche - this affects the release scripts for ocaml.


gasche:
I'm not completely sure I follow, how does that change the release process?

dra27:
@gasche - if the creation of the new `ocaml-system` package for the released version involves copying the previous one, then you're fine ... but if it's done from a template, then that template needs updating (see the changes in anyone of the gen scripts in this PR)

dra27:
I'm not totally clear why some build jobs succeeded on Travis and some didn't - but equally the script change is the same on all versions and Travis can't test system compiler packages anyway.

kit-ty-kate:
Thanks!

PR #12702, opened 2018-09-26
open 0.2.2

smolkaj:


kit-ty-kate:
The `descr` and `url` files should not be there and be included in the `opam` file with opam2.

smolkaj:
> The `descr` and `url` files should not be there and be included in the `opam` file with opam2.

Should be fixed now.

kit-ty-kate:
Thank you!

PR #12724, opened 2018-09-30
bigstringaf: depopt on mirage-xen, as the conditional compilation in lib/xen/jbuild is done

hannesm:
//cc @seliopou see https://github.com/reynir/qubes-mirage-ssh-agent/issues/7 for the reported issue.  upstream is a slightly better fix https://github.com/inhabitedtype/bigstringaf/pull/17

this fix works since mirage-xen depends on mirage-xen-posix, which is ultimately used when calling out to `pkg-config`


kit-ty-kate:
Thanks!

PR #12728, opened 2018-10-01
[opam2 cleanup] merge descr and url files into opam files

mseri:
Some descr and url files have been merged in the
days after the merge to opam2. This PR is to
make it right and include them in the opam file
instead.

A forthcoming PR will add a synopsis or description
to the packages that are currently missing one
when it is possible to get it.


mseri:
Any failure coming with this should have already been there. I don't think this is the right PR to address them. Note that due to the base packages part of the CI is completely useless anyways.

kit-ty-kate:
Could you at least fix the opam-version errors? That's good otherwise

mseri:
Done! I also checked if there are additional version `1` or `1.*` around but it does not seem the case

kit-ty-kate:
Thanks!

PR #12730, opened 2018-10-01
Fix `with-test` filter on depends field

rjbou:
Correct filter is [`with-test`](https://opam.ocaml.org/doc/Manual.html#pkgvar-with-test), not `test`.

### DIFF
--- a/packages/jingoo/jingoo.1.2.19/opam
+++ b/packages/jingoo/jingoo.1.2.19/opam
@@ -17,7 +17,7 @@ depends: [
   "pcre"
   "uutf"
   "uucp"
-  "ounit" {test}
+  "ounit" {with-test}
 ]
 
 description: """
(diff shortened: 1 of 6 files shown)


rjbou:
removed `nocrypto.0.5.4-1` modification because at build opam tries to patch the unmodified opam file that is in the archive using the modified patch file . ping @pqwy

rjbou:
ping @Drup for [`tree_layout` error](https://travis-ci.org/ocaml/opam-repository/jobs/436113027#L2607).

Drup:
Strange. It should go away if you remove the line for `--enable-benchmark` or add a test dependency towards `benchmark`. I would tend to favor the first solution.

Drup:
Ah, that means it should be `tyxml < 4.0` ! It seems test deps are bit undertested. :(

kit-ty-kate:
Thanks a lot!

PR #12732, opened 2018-10-01
Add ppx_deriving_cmdliner.0.4.1

smondet:


smondet:
Travis error:

```

#=== ERROR while compiling ppx_deriving_cmdliner.0.4.1 ========================#
# context     2.0.0 | linux/x86_64 | ocaml-base-compiler.4.07.0 | file:///home/travis/build/ocaml/opam-repository
# path        ~/.opam/4.07.0/.opam-switch/build/ppx_deriving_cmdliner.0.4.1
# command     ~/.opam/opam-init/hooks/sandbox.sh build jbuilder runtest -p ppx_deriving_cmdliner -j 1
# exit-code   1
# env-file    ~/.opam/log/ppx_deriving_cmdliner-17882-6cc03f.env
# output-file ~/.opam/log/ppx_deriving_cmdliner-17882-6cc03f.out
### output ###
# File "test/jbuild", line 5, characters 14-22:
# 5 |   (libraries (alcotest result))
#                   ^^^^^^^^
# Error: Library "alcotest" not found.
# Hint: try: dune external-lib-deps --missing -p ppx_deriving_cmdliner @runtest
```

Why isn't the script grabbing the `{test}` dependencies before running the tests?

mseri:
In `opam2` you need to use `with-test` in place of `test`. And you run the test with `run-test` instead of `build-test`

smondet:
Now the only failure happens while running the tests on 4.03 (I don't really understand the error), the package still builds fine.

packages/ppx_deriving_cmdliner/ppx_deriving_cmdliner.0.4.1/opam
```diff
@@ -0,0 +1,45 @@
...
+  "ocamlfind"    {build}
+  "ppxfind"      {build}
+  "cppo"         {build}
+  "alcotest"     {with-test}
+  "ppx_import"   {with-test & >= "1.1"}
+]
+synopsis: "Cmdliner.Term.t generator"
+description: """
+ppx_deriving_cmdliner is a ppx_deriving plugin that generates
+a Cmdliner Term.t for a record type."""
```
pmetzger: Just an FYI, but the description ends up getting folded into the opam web pages, and if I were reading through them, I'm not sure this would be enough for me to figure out why I want to use the package.

smondet: The package-page also links to the webpage that has properly formatted examples and all.
This description also shows up in terminals (`opam show ...`).

PRs welcome in any case.

pmetzger: I don't think most people would paste the URL and look at the page if they don't already know that they need something. There are a truly huge number of packages in OPAM, and one doesn't have time to do that all day long. I think it's pretty important that a package explain what it's for, especially so people who are skimming descriptions can find the right one.

And, as I don't know what this package is for, I can't submit a PR because to do that better, as I'd first have to know what to say.

kit-ty-kate:
Let's merge it and worry about the failing test later. Could you look into it anyway?

kit-ty-kate:
Thanks!

PR #12734, opened 2018-10-02
Fix automatic conversion dev-repo field

Leonidas-from-XIV:
I don't know what happened but it looks wrong.


Leonidas-from-XIV:
Also broken in `ordma` so I pushed a fix to it as well.

kit-ty-kate:
Indeed. Thanks a lot!

PR #12735, opened 2018-10-02
mlcuddidl.3.0.4

nberth:
Bugfix w.r.t memory management


nberth:
@avsm What's with all those `unmet availability conditions: ocaml-version >= "4.02.0"` on every build? Examining CI traces for other packages, it seems to happen a lot (but does not cause build failures…).

PR #12736, opened 2018-10-02
Support ndbm, bugfix for omake

madroach:
These are two small changes I carried for some time in my local repository.


kit-ty-kate:
mmh can you explain your patch for omake? Also I wonder why it works without stating the patch in the `extra-files` field

The ndbm support seems fine indeed tho

mseri:
I think the osx failure is related to an issue with the version of openssl shipped with it and the supported hashes. We have seen it in many different packages. Can you try replacing the url checksum with md5 and see if it resolves the issue?

madroach:
I added the _optional_ ``extra-files:`` field.

As to why the patch is needed:
```
*** omake error:
   File lib/build/OCaml.om: line 184, characters 4-31
   Array value where string expected:
      Use the $(string ...) function if you really want to do this
      The array has length 3:
         [0] = <data "cc"> : String
         [1] = <data "-O2"> : String
         [2] = <data "-pipe"> : String
```

madroach:
I added another commit adding a ``with-doc`` build-dep to sequence. Or is this supposed to be filtered ``{build with-doc}`` ?
Please feel free to cherry-pick any of the commits I added to this branch.

UnixJunkie:
@madroach just FYI, you might be interested by the dokeysto package in opam

kit-ty-kate:
Sorry I neglected this PR. Merging

kit-ty-kate:
Thanks!

MdeLv:
Hi, 
can you have a look at [ conf-dbm triggers false "upstream changes" #16308 ](https://github.com/ocaml/opam-repository/issues/16308#issue-608433728)?
The issue seems related to this merge: [https://github.com/ocaml/opam-repository/commit/9ecbe4e1cb23338275ab0b394ce278c63ba4d1c7](https://github.com/ocaml/opam-repository/commit/9ecbe4e1cb23338275ab0b394ce278c63ba4d1c7).

Thanks.

PR #12743, opened 2018-10-04
add OCaml release 4.07.1

damiendoligez:


packages/ocaml-system/ocaml-system.4.07.1/opam
```diff
@@ -0,0 +1,14 @@
...
+  "ocaml" {post}
+  "base-unix" {post}
+  "base-threads" {post}
+  "base-bigarray" {post}
+]
+conflict-class: "ocaml-core-compiler"
+available: sys-ocaml-version = "4.07.1"
+flags: compiler
+build: ["ocaml" "gen_ocaml_config.ml"]
+substs: "gen_ocaml_config.ml"
```
kit-ty-kate: the previous versions seem to have:
```
extra-files: [
  "gen_ocaml_config.ml.in" "md5=44ea2d0f7abd8e1b57f16ae1b28fcd2c"
]
```
I'm not sure it will work as is.

pmetzger: It would also be nice to fix the warnings...

kit-ty-kate: It would also be nice to avoid having a passive-aggressive attitude...

pmetzger: It is difficult to judge people's attitude or emotion in electronic media. The absence of facial expressions or voice tone makes knowing the counterparty's intent difficult. I generally try (when I remember!) to presume good faith on the part of others, and that makes my life better, and my interactions freer of friction. I can't claim that I'm perfect at it, of course.

kit-ty-kate:
should be fine now I guess. Thanks!

kit-ty-kate:
mmh on second thought the older one seem to not have the `extra-files` statement.. Is there something special here? @AltGr @rjbou

kit-ty-kate:
Anyway, merging

PR #12751, opened 2018-10-05
[coq-serapi] Fix constraint on old coq-serapi.

ejgallego:
Unfortunately this older release is not compatible with OCaml >= 4.07
due to a problem with `Stdlib` packing.


kit-ty-kate:
Thanks!

PR #12752, opened 2018-10-05
Jbuilder dune rename

tuncer:
As discussed with @kit-ty-kate.


tuncer:
The motivation is that dune provides `dune` plus `jbuilder`, and therefore depending on the package named `jbuilder` is not the right thing anymore. This doesn't mean projects will be automatically migrated to `dune`. Rather, it's a fix for the post-install message. So, projects will still use jbuilder until they're adapted to the dune changes. https://github.com/ocaml/opam-repository/blob/d8d70fb664f9fac3f64807704d6817bbfb2de06f/packages/jbuilder/jbuilder.transition/opam#L14-L15

tuncer:
5e0e5d9?

kit-ty-kate:
mmh I'm not sure to follow? Why remove jbuilder constraints when they are relevant? Only the one that you switched to dune are not relevant anymore

tuncer:
That makes more sense. One moment.

tuncer:
Done.

packages/camltc/camltc.0.9.8/opam
```diff
@@ -11,7 +11,7 @@ build: [
 ]
 depends: [
   "ocaml" {> "4.02.3"}
-  "dune" {build & >= "1.1.0"}
+  "dune" {build}
```
kit-ty-kate: This change is wrong :/

packages/cfg/cfg.2.2.0/opam
```diff
@@ -14,7 +14,7 @@ build: [
 
 depends: [
   "ocaml" {>= "4.04"}
-  "dune" {build & >= "1.1.0"}
+  "dune" {build}
```
kit-ty-kate: Same comment

packages/cohttp-lwt-jsoo/cohttp-lwt-jsoo.1.1.1/opam
```diff
@@ -15,7 +15,7 @@ homepage: "https://github.com/mirage/ocaml-cohttp"
 bug-reports: "https://github.com/mirage/ocaml-cohttp/issues"
 depends: [
   "ocaml" {>= "4.03.0"}
-  "dune" {build & >= "1.1.0"}
+  "dune" {build}
```
kit-ty-kate: same

kit-ty-kate:
I'm not going to go through every dune constraints removed because it would take too much time and it would be redundant but they should be let as is.

kit-ty-kate:
Anybody can comment on this? @rgrinberg @diml ?

kit-ty-kate:
Okay, good. I reviewed everything, fixed some mistakes and I'm just gonna wait for Camelus to tell us it doesn't break anything and move forward with the merge.

kit-ty-kate:
`these releases are not installable anymore (78)` mmh :/

tuncer:
What do we do?

kit-ty-kate:
> What do we do?

Mmh I don't know. This makes a lot of packages (even if they are mostly old versions) not installable anymore, and it's annoying. We can try to see where it comes from and maybe try to fix it but I'm not sure it's feasible.

avsm:
This PR seems to be heading in the right direction. Why not simply leave the `jbuilder` dependency on the packages that are uninstallable (since they require an older beta of jbuilder) and keep the rest?  Once you get to a set of diffs that have no uninstallable packages it should be good to merge!

avsm:
I'm closing this one as it's causing some epic CI churn.  Please do reopen when it's appropriate to merge it.   I am pro merging this with just the installable packages included.

PR #12768, opened 2018-10-09
remove version field

LucaFranceschini:
It is wrong and it is not used anymore


kit-ty-kate:
Indeed. Thanks for fixing that!

PR #12774, opened 2018-10-10
[new release] dune (1.4.0)

rgrinberg:
CHANGES:

- Do not fail if the output of `ocamlc -config` doesn't include
  `standard_runtime` (ocaml/dune#1326, @diml)

- Let `Configurator.V1.C_define.import` handle negative integers
  (ocaml/dune#1334, @Chris00)

- Re-execute actions when a target is modified by the user inside
  `_build` (ocaml/dune#1343, fix ocaml/dune#1342, @diml)

- Pass `--set-switch` to opam (ocaml/dune#1341, fix ocaml/dune#1337, @diml)

- Fix bad interaction between multi-directory libraries the `menhir`
  stanza (ocaml/dune#1373, fix ocaml/dune#1372, @diml)

- Integration with automatic formatters (ocaml/dune#1252, fix ocaml/dune#1201, @emillon)

- Better error message when using `(self_build_stubs_archive ...)` and
  `(c_names ...)` or `(cxx_names ...)` simultaneously.
  (ocaml/dune#1375, fix ocaml/dune#1306, @nojb)

- Improve name detection for packages when the prefix isn't an actual package
  (ocaml/dune#1361, fix ocaml/dune#1360, @rgrinberg)

- Support for new menhir rules (ocaml/dune#863, fix ocaml/dune#305, @fpottier, @rgrinberg)

- Do not remove flags when compiling compatibility modules for wrapped mode
  (ocaml/dune#1382, fix ocaml/dune#1364, @rgrinberg)

- Fix reason support when using `staged_pps` (ocaml/dune#1384, @charlesetc)

- Add support for `enabled_if` in `rule`, `menhir`, `ocamllex`,
  `ocamlyacc` (ocaml/dune#1387, @diml)

- Exit gracefully when a signal is received (ocaml/dune#1366, @diml)

- Load all defined libraries recursively into utop (ocaml/dune#1384, fix ocaml/dune#1344,
  @rgrinberg)

- Allow to use libraries `bytes`, `result` and `uchar` without `findlib`
  installed (ocaml/dune#1391, @nojb)

- Take argument to self_build_stubs_archive into account. (ocaml/dune#1395, @nojb)

- Fix bad interaction between `env` customization and vendored
  projects: when a vendored project didn't have its own `env` stanza,
  the `env` stanza from the enclosing project was in effect (ocaml/dune#1408,
  @diml)

- Fix stop early bug when scanning for watermarks (ocaml/dune#1423, @diml)

### DIFF
--- a/packages/dune/dune.1.4.0/opam
+++ b/packages/dune/dune.1.4.0/opam
@@ -1,0 +1,44 @@
+opam-version: "2.0"
+maintainer: "opensource@janestreet.com"
+authors: ["Jane Street Group, LLC <opensource@janestreet.com>"]
+homepage: "https://github.com/ocaml/dune"
+bug-reports: "https://github.com/ocaml/dune/issues"
+dev-repo: "git+https://github.com/ocaml/dune.git"
+license: "MIT"
+depends: [
+  "ocaml" {>= "4.02"}
+]
+build: [
+  # opam 2 sets OPAM_SWITCH_PREFIX, so we don't need a hardcoded path
+  ["ocaml" "configure.ml" "--libdir" lib] {opam-version < "2"}
+  ["ocaml" "bootstrap.ml"]
+  ["./boot.exe" "--release" "--subst"] {pinned}
+  ["./boot.exe" "--release" "-j" jobs]
+]
+conflicts: [
+  "jbuilder" {!= "transition"}
+]
+
+synopsis: "Fast, portable and opinionated build system"
+description: """
+dune is a build system that was designed to simplify the release of
+Jane Street packages. It reads metadata from "dune" files following a
+very simple s-expression syntax.
+
+dune is fast, it has very low-overhead and support parallel builds on
+all platforms. It has no system dependencies, all you need to build
+dune and packages using dune is OCaml. You don't need or make or bash
+as long as the packages themselves don't use bash explicitly.
+
+dune supports multi-package development by simply dropping multiple
+repositories into the same directory.
+
+It also supports multi-context builds, such as building against
+several opam roots/switches simultaneously. This helps maintaining
+packages across several versions of OCaml and gives cross-compilation
+for free.
+"""
+url {
+  src: "https://github.com/ocaml/dune/releases/download/1.4.0/dune-1.4.0.tbz"
+  checksum: "md5=dc862e5d821ff4d8bef16a78bd472431"
+}


kit-ty-kate:
poke @mseri for the `eigen`/`owl` related problems: https://ci.ocamllabs.io/log/saved/docker-run-f22b1fa17d182dc0684d8dabe5c30828/11cc8cb6dc8255040ebf775a45da98d9a3bb40ef

Otherwise there seem to be a related problem that wasn't there before:
```
#=== ERROR while compiling jupyter-archimedes.2.3.2 ===========================#
# context              2.0.0 | linux/x86_64 | ocaml-base-compiler.4.07.0 | git+file:///home/opam/opam-repository
# path                 ~/.opam/4.07/.opam-switch/build/jupyter-archimedes.2.3.2
# command              ~/.opam/4.07/bin/dune build -p jupyter-archimedes -j 127
# exit-code            1
# env-file             ~/.opam/log/jupyter-archimedes-18-58c0d3.env
# output-file          ~/.opam/log/jupyter-archimedes-18-58c0d3.out
### output ###
# File "jupyter-archimedes/src/dune", line 10, characters 14-30:
# 10 |               archimedes.cairo))
#                    ^^^^^^^^^^^^^^^^
# Error: Library "archimedes.cairo" in /home/opam/.opam/4.07/lib/archimedes is hidden (unsatisfied 'exist_if').
# Hint: try: dune external-lib-deps --missing -p jupyter-archimedes @@default
```

mseri:
Thanks @kit-ty-kate the issue with `eigen` is due to a too old `gcc`, unfortunately the install does not fail properly and makes the owl package unusable. We need to find a better way to deal with the external c++ libraries there, but unfortunately there seems to be no easy or right solution

mseri:
The archimedes failure seems to be there also with `dune 1.3.1`: https://ci.ocamllabs.io/log/saved/docker-build-92a97f1479c3e3d14a715749ec2aa264/dd2bcc2c1ef76359e2430188c28507b78a217251

rgrinberg:
If I had to guess, it's probably a race condition. Is it possible to restart the build?

pmetzger:
> If I had to guess, it's probably a race condition. Is it possible to restart the build?

Yes, but you can do it on your own just by force pushing the branch even though it hasn't actually changed. That will trigger a new build.

mseri:
We have been imprecise. It is jupyter-Archimedes that fails because Archimedes.cairo appears to be an hidden library

Chris00:
On 11 October 2018 at 16:36 GMT, Marcello Seri wrote:
>
> We have been imprecise. It is jupyter-Archimedes that fails because
> Archimedes.cairo appears to be an hidden library

I guess, this is because Archimedes predates the trend to have several packages for a given library so that dependencies are more explicit.  jupyter-Archimedes should depend on cairo — but that may not completely fix the fact that Archimedes will compile archimedes.cairo when cairo is present, that will depend on the order of installation... ☹

The next version of Archimedes will be released as separate packages — but do not hold your breath, I intend to rewrite it from the ground up and split the functionality into several libraries.

kit-ty-kate:
> Yes, but you can do it on your own just by force pushing the branch even though it hasn't actually changed. That will trigger a new build.

as it's just from datakit ci, everyone can just relaunch the build by clicking "rebuild". It would avoid unnecessary rebuilds

mseri:
Anyways it doesn’t really matter now as @Chris00 clarified how jupyter-archimedes will build correctly only if cairo happens to be compiled before archimedes. So this failure is not dune’s fault

PR #12775, opened 2018-10-10
Fix cairo2 upper bound for archimedes

Chris00:


mseri:
Thanks for the fast fix

PR #12776, opened 2018-10-10
[new release] cairo2, cairo2-pango and cairo2-gtk (0.6)

Chris00:
Interface to Cairo

- Project page: <a href="https://github.com/Chris00/ocaml-cairo">https://github.com/Chris00/ocaml-cairo</a>
- Documentation: <a href="https://Chris00.github.io/ocaml-cairo/doc">https://Chris00.github.io/ocaml-cairo/doc</a>

##### CHANGES:

- New `Ft` module to support FreeType fonts.  This is enabled if the
  package `conf-freetype` is installed.  On the C side, the exported
  header file `cairo_ocaml.h` defines the macro `OCAML_CAIRO_HAS_FT`
  when the Cairo bindings were compiled with TrueType support.
- New package `Cairo2-pango` providing the module `Cairo_pango`.
- Remove labels that were not bringing a clear benefit.  With Dune
  default behavior, users will feel compelled to write labels which
  was cluttering the code with the previous interface.  With Merlin,
  it is now possible to have the documentation of a function under the
  cursor displayed with a simple keystroke which should alleviate
  having slightly less documentation in the types.
- Improve the documentation.
- Use Dune (not the former Jbuilder) to compile.

### DIFF
--- a/packages/cairo2-gtk/cairo2-gtk.0.6/opam
+++ b/packages/cairo2-gtk/cairo2-gtk.0.6/opam
@@ -1,0 +1,33 @@
+opam-version: "2.0"
+maintainer: "Christophe Troestler <Christophe.Troestler@umons.ac.be>"
+authors: [ "Christophe Troestler <Christophe.Troestler@umons.ac.be>"
+           "Pierre Hauweele <pierre@hauweele.net>" ]
+license: "LGPL-3.0 with OCaml linking exception"
+homepage: "https://github.com/Chris00/ocaml-cairo"
+dev-repo: "git+https://github.com/Chris00/ocaml-cairo.git"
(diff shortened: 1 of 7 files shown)


Chris00:
> new broken packages (2): cairo2-gtk.0.6 cairo2-pango.0.6

That is no very explicit on the reason why...

mseri:
I don't know why camelus complains. Often (but not always) it is right but the reasons can be obscure. I have added some missing bounds, then if the CI is green I think this is good to merge. If we find that there is an issue as described by camelus, we can fix it in a subsequent PR

mseri:
The CI failures with jupyter-archimedes seem related to dune >= 1.3.1 and are independent of this

Issue #12729, opened 2018-10-01
CI should fail if a package does not present a description or a synopsis


mseri:
See the initial discussion with @pmetzger  here: https://github.com/ocaml/opam-repository/pull/12727

mseri:
It should also complain if `test` is used instead of `with-test`: 
https://github.com/ocaml/opam-repository/pull/12732

pmetzger:
I obviously agree this should be fixed. :)

rjbou:
The `test` issue is more complicated because of [variables handling](https://opam.ocaml.org/doc/Manual.html#Package-variables). Some variables are opam defined  but user can also defines its own. Plus, as variables can be undefined, and filter expressed according, we can't raise an error if `{test}` is present (while packager meant `{with-test}`), or if the `test` variable is null.

Note that the opam package upgrader rewrite `test` and `doc` variables into `with-test` and `with-doc`.

c-cube:
A warning (in the linter) would still do the trick, right? if someone wants to define `test` nonetheless they would be able to, but for most users using `{test}` is probably a mistake. Also, it's the first time I hear of a package upgrader, which is why I was confused by the same problem.

rjbou:
The two PR merged on master.

pmetzger:
I'm not sure that closing this was a good idea, merely because it hadn't had much activity of late. The problem remains interesting to fix.

PR #12780, opened 2018-10-11
[cleanup]: Add descr and or synopsis to packages that were missing it

mseri:
See https://github.com/ocaml/opam-repository/issues/12729#issuecomment-425925584

The values for synopis and/or description have been copied from previous (or newer) version of the relative packages when possible, otherwise come from the description of their github repository.


mseri:
Ping @pmetzger

Issue #12640, opened 2018-09-17
cppo_ocamlbuild missing depends on ocamlfind


gares:
```shell
gares@gargamel:/tmp$ opam install -y cppo_ocamlbuildThe following actions will be performed:
  - install cppo_ocamlbuild 1.6.0

=-=- Gathering sources =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
[cppo_ocamlbuild] Archive in cache

=-=- Processing actions -=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
[ERROR] The compilation of cppo_ocamlbuild failed at "jbuilder build -p
        cppo_ocamlbuild -j 2".

#=== ERROR while installing cppo_ocamlbuild.1.6.0 =============================#
# opam-version 1.2.2
# os           linux
# command      jbuilder build -p cppo_ocamlbuild -j 2
# path         /tmp/xxx/4.02.3/build/cppo_ocamlbuild.1.6.0
# compiler     4.02.3
# exit-code    1
# env-file     /tmp/xxx/4.02.3/build/cppo_ocamlbuild.1.6.0/cppo_ocamlbuild-18926-1872ec.env
# stdout-file  /tmp/xxx/4.02.3/build/cppo_ocamlbuild.1.6.0/cppo_ocamlbuild-18926-1872ec.out
# stderr-file  /tmp/xxx/4.02.3/build/cppo_ocamlbuild.1.6.0/cppo_ocamlbuild-18926-1872ec.err
### stderr ###
# [...]
# 5 | (ocamlyacc (cppo_parser))
#     ^^^^^^^^^^^^^^^^^^^^^^^^^
# Warning: File cppo_parser.mli is both generated by a rule and present in the source tree.
# As a result, the rule is currently ignored, however this will become an error in the future.
# To keep the current behavior and get rid of this warning, add a field (fallback) to the rule.
# File "ocamlbuild_plugin/jbuild", line 8, characters 14-24:
# 8 |   (libraries (ocamlbuild))))
#                   ^^^^^^^^^^
# Error: Library "ocamlbuild" not found.
# Hint: try: dune external-lib-deps --missing -p cppo_ocamlbuild @install



[...]

PR #12782, opened 2018-10-11
cppo_ocamlbuild: add missing ocamlfind dependency

mseri:
Should fix https://github.com/ocaml/opam-repository/issues/12640

I thought that using dune could have been enough but we have often seen a similar failure in the CI even when `jbuilder.transition` is used, so I went with adding the dependency without too many cases and combinations of packages.


kit-ty-kate:
Looks good. Thanks

PR #12788, opened 2018-10-12
osx-cf: add a missing conflict with lwt

samoht:
/cc @dsheets 

See the error in https://travis-ci.org/ocaml/opam-repository/jobs/440262234


mseri:
LGTM. The osx failure can be ignored and seems due to lack of support for safe-strings for the old `ctypes` required by `osx-cf.0.1.0`

PR #12789, opened 2018-10-12
[WIP] Add a conf-docker package

Chris00:


Chris00:
@avsm @kit-ty-kate  Having a `conf-docker` package that tries to add a new repository is not a good solution IMHO because it is likely to fail on end-user systems (will only work for CI).  How about to add Docker repository to Travis?

avsm:
I think I need a bit more context :-) Why do we need a `conf-docker` package at all for the `docker-api` package? Doesn't it work perfectly fine accessing a remote docker socket on another host?

Chris00:
It does.  By default, it connects to a Unix socket.  The package compiles fine without Docker installed.  It is just the tests — that the CI runs — that depend on Docker being available.

avsm:
I see, that makes sense. I think we should simply skip those tests in the opam published version, and reserve them for the docker-api repo itself.  The issue is that granting `docker` access from within CI basically gives root level access to the host without having a bunch of docker-in-docker security policies in place...

Chris00:
> skip those tests in the opam published version,

They are executed but do not fail whenever Docker is not present.

PR #12792, opened 2018-10-14
Package sturgeon.0.4.2

let-def:
### `sturgeon.0.4.2`
A toolkit for communicating with Emacs
Sturgeon implements various tools for manipulating Emacs from OCaml:
- `Sturgeon_sexp` implements the Emacs dialect of S-expressions
- `Sturgeon_session` implements an "session protocol" to make RPC to Emacs from OCaml and vice versa
- `Sturgeon_stui` is a session implementing an [Inuit](https://github.com/let-def/inuit) backend: one can now runs text user-interface on an Emacs buffer
- `Sturgeon_recipes_*` offers different "rendez-vous" points to connect to Emacs

Sturgeon is distributed under the ISC license.

---
* Homepage: https://github.com/let-def/sturgeon
* Source repo: git+https://github.com/let-def/sturgeon.git
* Bug tracker: https://github.com/let-def/sturgeon/issues

---


mseri:
This PR is moving `sturgeon.0.4.1` to `sturgeon.0.4.2` effectively deleting the first. Please keep the old version as well, then it's good to merge

let-def:
The old version does not install anyway, it was a mistake to publish it. The new version is exactly the same with a correct install script

mseri:
I think this goes against the policy of not removing packages, even if broken. I'd prefer to stick with that and in case mark the package as uninstallable but I'd like to hear also @kit-ty-kate opinion on that

kit-ty-kate:
If it was not installable it's ok to remove it as it shouldn't break anything. Merging. Thanks!

PR #12807, opened 2018-10-15
[new release] digestif (0.7)

dinosaure:
CHANGES:

- Fixed HMAC on BLAKE2{S,B} (@emillon)
- Fixed `convenient_of_hex` (@dinosaure, @hannesm, @cfcs)
- Add `of_raw_string`/`to_raw_string` (@samoht)
- Test `digestif` on solo5 and xen backends (@amoht)
- *breaking change*, commont type `t` is an abstract type
- Fixed META file (@dinosaure, @g2p)
- New dependency `eqaf` (@dinosaure, @cfcs, @hannesm) (constant-time equal function)
- Remove `Obj.magic` in common implementation (@dinosaure, @samoht)
- Add conveniences functions in common implementation (@hcarty)
- Add option-returning functions in common implementation (@harcty)
- Verify length of string on `of_raw_string` function (@hcarty)
- Release runtime lock (@andersfugmann, @dinosaure, @cfcs)
- Bounds check (@cfcs, @dinosaure)
- Fixed linking problem (@andersfugmann, @g2p, @dinosaure)
- Update OPAM file (@dinosaure)

### DIFF
--- a/packages/digestif/digestif.0.7/opam
+++ b/packages/digestif/digestif.0.7/opam
@@ -9,11 +9,13 @@ bug-reports:  "https://github.com/mirage/digestif/issues"
 dev-repo:     "git+https://github.com/mirage/digestif.git"
 doc:          "https://mirage.github.io/digestif/"
 license:      "MIT"
-synopsis:     "Hashes implementations (SHA*, RIPEMD160, BLAKE2* and MD5)"
+synopsis:     "Hashes implementations (SHA*, RIPEMD160, BLAKE2* and MD5) and HMAC of them"
 description: """
 Digestif is a toolbox to provide hashes implementations in C and OCaml.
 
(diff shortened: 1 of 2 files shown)


dinosaure:
@emillon `noise` does not compile because we decided to abstract the type `t` (you should use `to_raw_string`/`of_raw_string` now instead `(hash :> string)`.

cfcs:
The opam description doesn't seem to mention anything about `HMAC`? Maybe that would make sense to mention here?

I also think it would be nice to have a more extensive description of the "linking trick" thing, ie mention that `checkseum.ocaml` or `checkseum.c` needs to be mentioned before `digestif` in your `_tags`. (?and `META`?)

emillon:
@dinosaure you can add a version constraint to noise, I'll prepare an updated release.

kit-ty-kate:
@dinosaure Could you replicate the fix to upstream?

PR #12823, opened 2018-10-18
New ACGtk releases

Nyavlys:
This pull request is about an improved version of ACGtk and dropping some dependencies (in particular to camlp2 via BOLT) by using other librairies.


kit-ty-kate:
Hi, please upgrade your package to opam2. opam1.2 is not supported anymore and your PR won't be merged.

kit-ty-kate:
With opam2 it is better to have url and descr be integrated into the opam file. See https://opam.ocaml.org/doc/Upgrade_guide.html#New-features_1

PR #12830, opened 2018-10-19
+conduit.1.3.0 and associated packages

avsm:
* Mark `ppx_sexp_conv` as not just a build dependency.
* Switch build to dune from jbuilder.
* Depend on more precise versions of `mirage-types` instead
  of just the generic package.
* Fix ocamldoc headings to work with latest odoc (level 2 not 1).


mseri:
Thanks for the yurt fix. Everything else looks good

PR #12831, opened 2018-10-19
yurt.0.4: upper bound on lwt due to use of lwt_log

avsm:


avsm:
fixes revdep failure in #12830

mseri:
The osx issue is the usual one. Thanks for the fix

PR #12842, opened 2018-10-21
cohttp-async: add missing uri dependency and bounds

mseri:
This should prevent further issues when different versions of uri, cohttp, and cohttp-async are installed or pinned


mseri:
The failure for osx is known and due to the opennssl bug in opam2. The bug in websockets is due to a build concurrency issue, ssl is not yet present when conduit is built. Imho this is good to merge

kit-ty-kate:
Thanks!

PR #12843, opened 2018-10-21
reason.3.3.7, rtop.3.3.7

anmonteiro:


packages/rtop/rtop.3.3.7/opam
```diff
@@ -0,0 +1,23 @@
+opam-version: "2.0"
+maintainer: "Jordan Walke <jordojw@gmail.com>"
+authors: [ "Jordan Walke <jordojw@gmail.com>" ]
+license: "MIT"
+homepage: "https://github.com/facebook/reason"
+doc: "http://reasonml.github.io/"
+bug-reports: "https://github.com/facebook/reason/issues"
+dev-repo: "git://github.com/facebook/reason.git"
+tags: [ "syntax" ]
+build: [
```
mseri: Please  add description and/or synopsis to this opam file

mseri: Can you also add it to rtop.3.3.5 that was missing it as well?

PR #12845, opened 2018-10-22
[new release] gsl (1.23.0)

mmottl:
CHANGES:

* Switched to dune and dune-release


pmetzger:
Just FYI, there's no description in your opam file. That means people won't know what this package is for and thus might fail to know they would want to use it.

mmottl:
It seems that's an issue with `dune-release`.  It complains about missing `description` and `synopsis` entries during certain parts of the release process even though a `descr` file is provided containing both.  This seems to be more than just a `dune-release` linting issue, it apparently also doesn't copy the `descr` file when submitting a new package.  See https://github.com/samoht/dune-release/issues/68 for the discussion.

I'll make a new submission once this issue has been fixed.

PR #12861, opened 2018-10-25
rename all frama-c versions

maroneze:
Creating PR in WIP state to see what the CI builds report.


maroneze:
I tried to add constraints when the builds failed, but there is one case which is intriguing: `why.2.34` on OCaml 4.03.0. Frama-C's build is failing due to some relocation error, but I cannot reproduce it. Given the age of the package (2014), the fact that `why` itself is deprecated in favor of `why3`, and the availability of other versions to be installed, I'd consider this is as a very minor issue. In the worst case, I might try adding constraints to render Frama-C incompatible with this version of `why3`, but I'm not sure it's worth the effort.

I was also unable to reproduce the error when compiling `why3.0.88.3` in OCaml 4.06.1 (`Unbound module Big_int`). I tried installing the exact same versions of the packages indicated in the log (https://api.travis-ci.org/v3/job/446147216/log.txt) in a fresh opam switch, but it succeeded. Is there a more reliable way to reproduce Travis CI's errors to try and debug them?

maroneze:
Also, in other build failures, I'm not sure about what to do. For instance, in https://travis-ci.org/ocaml/opam-repository/jobs/446147221 (OCaml 4.01, package why.0.81), the build fails with a patch which cannot be applied. I can reproduce the error, but simply trying to install the package by itself already fails. It's as if the package never worked. I tried a different OCaml version and it fails with the same error. What should be done in this case? Just ignore it, or try to blacklist all affected OCaml compiler versions (even if in the end it means no version will remain)?

maroneze:
> Shoukd we merge this after #12877?

I will rebase this PR after #12877 is merged, renaming the updated package in the process. I will also try to fix the issue with Frama-C 10.0 in Debian with OCaml 4.03.0.

mseri:
I merged  #12877

maroneze:
Just to confirm, I realized after testing that `frama-c.10.0` does not compile on OCaml 4.03.0 on Debian (I am able to compile it on Fedora), so I tried adding a constraint:

    depends: [
       "ocaml" {>= "3.12" & != "4.02.0" & < "4.04"}
    +  "ocaml" {!= "4.03" | os-distribution != "debian"}

Some constraints were already present, so I added a new line, to signify "either OCaml is not 4.03.0, or we are not in a Debian", expecting it to work as a conjunction with the previous line; however, apparently the CI is still trying to install Frama-C 10.0 on a 4.03.0 Debian.

Is the previous constraint incorrect? Wouldn't it be equivalent to:

    depends: [
      "ocaml" {>= "3.12" & != "4.02.0" & < "4.04" &
               (!= "4.03" | os-distribution != "debian")}

hannesm:
@maroneze I'd try a separate conflict section:
```
confiict: [
  "ocaml" {= "4.03.0" & os-distribution = "debian"}
]
```

maroneze:
Thanks @hannesm, seems to work better.

I expect some Travis CI builds to fail, since old why3 versions seem to have some problems compiling, but I don't think it's related to this PR, and short of debugging old why3 versions, there's not much I can do about it, I'm afraid.

mseri:
I am fine with the merge. @hannesm ?

PR #12863, opened 2018-10-25
Add conf flag for conf-* packages

rjbou:
Add only `conf` flags for all, except:
- `conf-netsnmp.1.0.0` - `conf-numa.0.1.0` - `conf-secp256k1.1.0.0`:  remove url
- `conf-pkg-config` - `conf-rdkit`: remove light_uninstalll flag

For `conf-pkg-config`, it's not a real [conf](http://opam.ocaml.org/doc/Manual.html#opamflag-conf) package as it adds a link on openbsd.

As there is an error on conf flaggued packages check (cf https://github.com/ocaml/opam/issues/3625 & https://github.com/ocaml/opam/pull/3631), linting will certainly fail. Note that even after the PR merged on opam, `conf-pkg-config` will still fail linting.

### DIFF
--- a/packages/conf-aclocal/conf-aclocal.1.0.0/opam
+++ b/packages/conf-aclocal/conf-aclocal.1.0.0/opam
@@ -19,3 +19,4 @@ depends: ["conf-which" {build}]
 synopsis: "Virtual package relying on aclocal"
 description:
   "This package can only install if the aclocal program is installed on the system."
+flags: conf
--- a/packages/conf-autoconf/conf-autoconf.0.1/opam
+++ b/packages/conf-autoconf/conf-autoconf.0.1/opam
@@ -28,3 +28,4 @@ synopsis: "Virtual package relying on autoconf installation"
 description: """
 This package can only install if the autoconf command
 is available on the system."""
+flags: conf
--- a/packages/conf-bap-llvm/conf-bap-llvm.1/opam
+++ b/packages/conf-bap-llvm/conf-bap-llvm.1/opam
@@ -19,3 +19,4 @@ depexts: [
 conflicts: ["conf-llvm"]
 synopsis: "Virtual package relying on llvm library installation for BAP"
 extra-files: ["configure" "md5=47b8674130ba68afeeabebb719960c0f"]
+flags: conf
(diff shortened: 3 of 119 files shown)


mseri:
@rjbou I sm not sure this is reviewable aside looking at the opam files. Do you think it is good to merge?

rjbou:
@mseri Regarding opam file, it is good to merge, but 
* linting of conf flagged package [wrongly fails with opam 2.0.0/2.0.1](https://github.com/ocaml/opam/pull/3631). The patch is merged in master and will be available on 2.0.2.
* `conf-pkg-config` is not conf package [spec](http://opam.ocaml.org/doc/Manual.html#opamflag-conf) conform (ping @UnixJunkie) , so adding the flag makes it error linting
```
$ opam lint packages/conf-pkg-config/conf-pkg-config.1.1/opam 
packages/conf-pkg-config/conf-pkg-config.1.1/opam: Errors.
    error 46: Package is flagged "conf" but has source, install or remove instructions
```

UnixJunkie:
@rjbou you can have a look at that
https://github.com/ocaml/opam-repository/pull/12892

mseri:
There is also this new conf package that will need to be added `conf-glfw3.1`

rjbou:
@mseri updated PR & `conf-glfw3.1`.

We don't have yet a release date for 2.0.2.

I see two side-effect to merging : 
- it can show a wrong error to user, while no opam version corrects it (except `dev`). But A good side 
- it will help (new) maintainers to add the conf flag and linting will be more accurate regarding install, remove fields.

UnixJunkie:
@mato is the person who modified conf-pkg-config for openbsd

mseri:
Can we backport the fix for this error and release `opam` 2.0.1-1, so at least new users would not see the issue? Or make opam 2.0.2 another small fix release instead of waiting for long?

mato:
See #12018 for an explanation of the OpenBSD pkg-config change. What is the 
issue here? If it's just that conf-* SHOULD NOT install anything, I think 
this is a valid exception to the rule, esp. since reverting that behaviour 
will break existing setups.

UnixJunkie:
Maybe, a proper fix (to be confirmed by opam experts) is to create a new package.
Let's call it pkg-config-openbsd.
conf-pkg-config should depend on pkg-config-openbsd, only if the current os is openbsd.
The pkg-config-openbsd package would be the one that creates the symbolic link upon install, and remove it upon uninstall.

kit-ty-kate:
`conf-netsnmp.1.0.0` archive cannot be removed. It contains a file that is used in the `build` phase.

kit-ty-kate:
Looks good, thanks!

PR #12864, opened 2018-10-25
Change the dependency on menhir into a build dependency.

fpottier:
This allows me not to lose atd (and its descendants)
whenever I remove menhir from my system.


kit-ty-kate:
I'm not sure why it fails on Travis but your change shouldn't be the cause of this. Merging. Thanks!

PR #12875, opened 2018-10-25
Added conf-glfw package.

SylvainBoilard:
This PR adds a conf-glfw package that other packages can depend on to insure the presence of a system libglfw installation.


SylvainBoilard:
@kit-ty-kate Thank you for trying to fix the CentOS installation. Apparently the package name is in fact `glfw-devel` but it is found in group epel which does not seem to be available in the current CI configuration. I’m not sure what I should do now ?

mseri:
I think it’s better to avoid the Ubuntu version filter to see already the failure at depext level. I can do the change myself and merge tomorrow. Thanks

PR #12884, opened 2018-10-27
[jingoo] release version 1.2.21

tategakibunko:


packages/jingoo/jingoo.1.2.21/descr
```diff
@@ -0,0 +1 @@
+Template engine almost compatible with Jinja2(python template engine)
```
kit-ty-kate: Could you use the `synopsis` field instead of a descr file?

packages/jingoo/jingoo.1.2.21/opam
```diff
@@ -0,0 +1,25 @@
...
+]
+install: [make "install"]
+remove: ["ocamlfind" "remove" "jingoo"]
+depends: [
+  "ocaml" {>= "4.02.3"}
+  "ocamlfind" {build}
+  "re"
+  "uutf" {>= "0.9.4"}
+  "uucp"
+  "ounit" {test & >= "2.0.0"}
```
kit-ty-kate: The new syntax for the `test` attribute is `with-test`

kit-ty-kate:
@rjbou do you have any hints on why Camelus says it's a broken package by any chance? It seems to install fine on the CIs.

packages/jingoo/jingoo.1.2.21/descr
tategakibunko: Thank you for your advice!

mseri:
Pinning the v1.2.21 branch with the latest opam the install is successful but I get some warnings:
```
[WARNING] Failed checks on jingoo package definition from source at git+file:///home/mseri/code/jingoo#v1.2.21:
    error 57: Synopsis and description must not be both empty
  warning 58: Found `test` variable, predefined one is `with-test`
```
Can that be related to the camelus failure?

I think we can merge this. If any failure comes out of it, we may find out what to do

kit-ty-kate:
Yeah, let's merge this for now. Thanks a lot!

tategakibunko:
Thanks for many helps!

PR #12887, opened 2018-10-27
Kicadsch version 0.4.0

jnavila:


kit-ty-kate:
Thx!

jnavila:
BTW, do you know when this version will be visible in opam?

kit-ty-kate:
mmh, usually the script runs every few hours but it seems broken right now. I'll ping the people responsible for the script.

kit-ty-kate:
It seems fixed now

PR #12890, opened 2018-10-28
Correctly submit plotkicadsch

jnavila:
It's better to stick to dune-release generated files.

It seems the files that were generated by a copied Makefile are not suitable for opam.

Reviewers: please confirm that *.opam files are not allowed in the repo.


mseri:
@jnavila you are right. Thanks for the fix

kit-ty-kate:
Indeed. Thanks!

PR #12891, opened 2018-10-29
Package tldr.0.1

RosalesJ:
### `tldr.0.1`
An ocaml tldr client

---
* Homepage: https://github.com/RosalesJ/tldr-ocaml/
* Source repo: git+https://github.com/RosalesJ/tldr-ocaml.git
* Bug tracker: https://github.com/RosalesJ/tldr-ocaml/issues

---

### DIFF
--- a/packages/tldr/tldr.0.1/opam
+++ b/packages/tldr/tldr.0.1/opam
@@ -1,0 +1,32 @@
+opam-version: "2.0"
+synopsis: "An ocaml tldr client"
+maintainer: "coby@case.edu"
+authors: "Jacob Rosales Chase <coby@case.edu>"
+license: "MIT"
+homepage: "https://github.com/RosalesJ/tldr-ocaml/"
+bug-reports: "https://github.com/RosalesJ/tldr-ocaml/issues"
+depends: [
+  "dune" {build}
+  "core" {>= "v0.10.0"}
+  "ppx_jane"
+  "bos"
+  "cohttp-lwt-unix"
+  "lwt_ssl"
+  "ANSITerminal"
+  "angstrom"
+]
+conflicts: [
+  "ssl" {= "0.5.6"}
+]
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+]
+dev-repo: "git+https://github.com/RosalesJ/tldr-ocaml.git"
+url {
+  src: "https://github.com/RosalesJ/tldr-ocaml/archive/v0.1.tar.gz"
+  checksum: [
+    "md5=26f9547395b11ddda7ce09c58495b03c"
+    "sha512=cc5e754b2249b3df5e20ecb742219bea09ea81ada74094059e53a02e45f5744ccb19476541a02e589c66f3806ad54831831f7583af54bb069f909afd08aa4185"
+  ]
+}


mseri:
Please add a synpsis or a description to the opam file

packages/tldr/tldr.0.1/opam
```diff
@@ -0,0 +1,26 @@
...
+  "dune" {build}
+  "core"
+  "lwt"
+  "ppx_jane"
+  "terminal_size"
+  "bos"
+  "cohttp"
+  "cohttp-lwt-unix"
+  "lwt_ssl"
+  "ssl" {= "0.5.5"}
```
kit-ty-kate: Is there no way of having a more relaxed constraint?

RosalesJ: Indeed ```lwt_ssl``` requires ```ssl```, but I require ```"ssl" {= "0.5.5"}``` because of an unfixed issue in [ssl 0.5.6](https://github.com/savonet/ocaml-ssl/issues/40).

I can, however, get rid of ```cohttp``` and ```lwt```

mseri: You could say 

```
conflicts: [ "ssl" {= "0.5.6"} ]
```

so that you don't need to change the opam file to pick up the fix in a later version instead of having the ssl bound on the deps

mseri: Is the patch in ssl.0.5.7?

RosalesJ:
Adding ```conflicts: [ "ssl" {= "0.5.6"} ]``` failed the macOS travis test.
My travis macOS tests pass though:
 [Build Status](https://travis-ci.org/RosalesJ/tldr-ocaml)

mseri:
The failure is a known issue with some old versions of osx and opam.2.0.0, not related to your changes

mseri:
Thanks for the changes

mseri:
I made one last lint. If the tests are fine I'll merge tomorrow morning

mseri:
Wait, this opam file contains no url field!

mseri:
You need to `git tag v0.1` or `dune-release tag` it, push the tag online, and then add something like
```
url {
  src:
    "https://github.com/RosalesJ/tldr-ocaml/releases/download/v0.1/tldr-v0.1.tbz"
  checksum: "md5=ed09e8305e54fefb4343a5b67bd94f57"
}
```
(don't look at that specific checksum, it is on a private branch with the `"-p" name` fix on the opam file)

RosalesJ:
Ah alright, I was using
```
opam publish
```
Where I should have done
```
opam publish https://github.com/RosalesJ/tldr-ocaml/archive/v0.1.tar.gz
```
so that it would append the url and checksums for me. Looks like it builds fine now except for ocaml 4.03 which gives a compile error. I've seen this in my own tests, but it doesn't happen every time. Should I restrict ocaml >= 4.04?

kit-ty-kate:
Looks like it needs more of a lower-bound constraint on `core` rather than a constraint on `ocaml`

RosalesJ:
I did some tests to restrict core to ```>= v0.10.0```, took some trial and error but all tests are passing. Any advice on figuring out the least constraint without trial and error?

Thanks for all your help.

mseri:
I don’t have enough experience with core changes between versions. I think he current version of he opam file is good and we can always tewak it more in a second stage. Thanks for the patience

PR #12892, opened 2018-10-29
proposal tweak for pkg-config

UnixJunkie:


hannesm:
it is not clear to me what issue this change aims to address.  if anyone can explain the issue to-be-fixed, and please test this PR on a OpenBSD system (you can install it in a virtual machine, get it from http://www.openbsd.org/faq/faq4.html#Download) to ensure this is still working, please report here. thanks.

packages/conf-pkg-config/conf-pkg-config.1.1/opam
```diff
@@ -5,13 +5,8 @@ homepage: "http://www.freedesktop.org/wiki/Software/pkg-config/"
 bug-reports: "https://github.com/ocaml/opam-repository/issues"
 license: "GPL"
 build: [
-  ["pkg-config" "--help"]
-]
-install: [
   ["ln" "-s" "/usr/local/bin/pkgconf" "%{bin}%/pkg-config"] {os = "openbsd"}
```
kit-ty-kate: this definitely shouldn't work when ran during the build phase (with sandboxing on)

rjbou:
@hannesm cf. https://github.com/ocaml/opam-repository/pull/12863#issue-225742222

rjbou:
@UnixJunkie conf package should [just check](http://opam.ocaml.org/doc/Manual.html#opamflag-conf) and not install things (neither remove).
On solution is to add a [message](http://opam.ocaml.org/doc/Manual.html#opamfield-messages) for openbsd user, asking them to add this link and relaunching the conf package install. Or create another package that _installs_ this link on openbsd (and other packages would depend on this new package with an openbsd filter).

UnixJunkie:
OK, my bad.

UnixJunkie:
@rjbou adding a message would be bad. The current setup is fully automatic; it should stay that way. I am not an openbsd user, but as a standard unix user, I don't like when things that were working in the past stop working.

PR #12898, opened 2018-10-30
added git+https to bap-emacs urls

gitoleg:
This PR adds the `https` protocol to the url field of `bap-emacs-dot` and `bap-emacs-mode` packages. 

Thanks  @jishoon  for reporting


ivg:
Note, users under a proxy or in other restrictive environments like firewalls may have difficulties accessing the git port, thus it is necessary to have the `+https` part.

PR #12899, opened 2018-10-30
Allow installation of coqide with coq.8.7.1+2

maroneze:


maroneze:
The build fails are:
- On Travis, macOS fails (seems systematic nowadays, for any build);
- On Debian Unstable, it is unable to download lablgtk (https://forge.ocamlcore.org/frs/download.php/1726/lablgtk-2.18.6.tar.gz); manually trying to download the file works for me. I tried rebuilding a few times, but I always get the same error.

mseri:
Thanks

PR #12903, opened 2018-11-01
3 packages from cyrus-/relit at 0.2.0

charlesetc:
This pull-request concerns:
-`ppx_relit.0.2.0`: An implementation of Typed Literal Macros for Reason
-`relit_helper.0.2.0`: A helper library for those wishing to write TLMs using Relit

---
* Homepage: https://github.com/cyrus-/relit
* Source repo: git://github.com/cyrus-/relit.git
* Bug tracker: https://github.com/cyrus-/relit/issues

---

### DIFF
--- a/packages/relit_helper/relit_helper.0.2.0/opam
+++ b/packages/relit_helper/relit_helper.0.2.0/opam
@@ -12,10 +12,10 @@ build: [
   ["dune" "build" "-p" "relit_helper"]
 ]
 depends: [
-  "dune"
+  "dune" {build}
   "extlib"
-  "ppxlib" {= "0.3.1"}
-  "ocaml-migrate-parsetree" {= "1.0.11"}
+  "ppxlib" {>= "0.3.1"}
+  "ocaml-migrate-parsetree"
   "base64"
   "relit-reason"
   "ocaml" {>= "4.07.0" & < "4.08.0"}
--- a/packages/relit_helper/relit_helper.0.2.0/opam
+++ b/packages/relit_helper/relit_helper.0.2.0/opam
@@ -14,7 +14,7 @@ build: [
 depends: [
   "dune"
   "extlib"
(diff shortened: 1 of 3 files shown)


packages/relit_helper/relit_helper.0.2.0/opam
```diff
@@ -0,0 +1,30 @@
...
+license: "MIT"
+dev-repo: "git://github.com/cyrus-/relit.git"
+build: [
+  ["dune" "build" "-p" "relit_helper"]
+]
+depends: [
+  "dune"
+  "extlib"
+  "ppxlib" {= "0.3.1"}
+  "ocaml-migrate-parsetree" {= "1.0.11"}
```
mseri: Do you really need so strict version bounds?

kit-ty-kate:
As @mseri noticed I wonder why there are so many strict version bounds. It would be better to avoid those. Also the `dune` dependencies don't have the `{build}` attribute. Please add it.
And the `ocamlfind` dependency is legit for `ppx_relit` but not for `relit-reason` as far as I could see. Also the dependency is only a dependency on the executable so the `{build}` attribute can also be added.

charlesetc:
Thanks for the help! I've relaxed the restrictions a bit: we specifically do need ppxlib greater than or equal to 0.3.1, but probably not pinned so I've fixed that. Also added `{build}` to the dune dependencies. OCamlfind is a runtime dependency of relit so I think that should stay without the `{build}`? Anyways hope the tests pass and thanks again!

charlesetc:
Also removed the pin of ocaml-migrate-parsetree!

charlesetc:
How does it look now?

kit-ty-kate:
Good. Thanks!

charlesetc:
Thanks very much! :balloon:

PR #12904, opened 2018-11-01
Pkg config openbsd

UnixJunkie:


UnixJunkie:
@mato if you can test this on an actual openbsd, that would be nice.

UnixJunkie:
related to https://github.com/rdkit/rdkit/issues/1037

UnixJunkie:
related to https://github.com/ocaml/opam-repository/pull/12863

mato:
@UnixJunkie I don't actually have an OpenBSD system readily available. My work on that change (in #12018) was to help our OpenBSD users (of MirageOS) get a working out-of-the-box experience. Perhaps @adamsteen can test this? Also /cc @avsm @hannesm.

Having said that, the change looks reasonable, but I'm not too happy (as always) about retroactive changes to the behaviour of a package (in this case `conf-pkg-config`) _without_ a version number change _and_ I still don't understand the underlying issue; where is this policy of "conf-* packages MUST not install anything" documented and when was it put in place?

UnixJunkie:
I'm OK to bump the version number of the package for this change.

UnixJunkie:
I bumped the version number and added a new version, leaving intact the previous version of the package.

mato:
Ah, found the "spec", buried in comments on #12863: http://opam.ocaml.org/doc/Manual.html#opamflag-conf.

So, this needs testing on an OpenBSD machine, mainly from a clean install (i.e. empty switch) to verify that the transitive dependency to a depext works with `mirage` as intended.

packages/pkg-config-openbsd/pkg-config-openbsd.1.0/opam
```diff
@@ -0,0 +1,26 @@
+opam-version: "2.0"
+maintainer: "unixjunkie@sdf.org"
+authors: ["Francois Berenger"]
+homepage: "http://www.freedesktop.org/wiki/Software/pkg-config/"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: "GPL"
+build: [
+  ["pkgconf" "--help"] {os = "openbsd"}
```
kit-ty-kate: I wonder if instead of having `os = "openbsd"` everywhere we could just add `available: [os = "openbsd"]` once?

packages/conf-pkg-config/conf-pkg-config.1.2/opam
```diff
@@ -0,0 +1,31 @@
...
+  ["pkgconfig"] {os-distribution = "alpine"}
+  ["devel/pkgconf"] {os = "freebsd"}
+  ["pkg-config"] {os = "macos" & os-distribution = "homebrew"}
+  ["pkgconf"] {os = "freebsd"}
+]
+synopsis: "Virtual package relying on pkg-config installation"
+description: """
+This package can only install if the pkg-config package is installed
+on the system."""
+flags: light-uninstall
```
kit-ty-kate: I think the wanted flag here would be `conf`? cc @rjbou

packages/pkg-config-openbsd/pkg-config-openbsd.1.0/opam
UnixJunkie: That would be one small improvement, if that works.

packages/conf-pkg-config/conf-pkg-config.1.2/opam
UnixJunkie: I changed it.

packages/pkg-config-openbsd/pkg-config-openbsd.1.0/opam
UnixJunkie: I factorized the OS test thing.

packages/conf-pkg-config/conf-pkg-config.1.2/opam
rjbou: @kit-ty-kate yes! it's juste the wrong linting that complains about `light_uninstall` flag

adamsteen:
Hi All
I have been laid up for a while, unable to do much, but i should have a little bit of free time. just want to confirm how i can test this?

Should i just pin "UnixJunkie/pkg-config-openbsd" branch and do a test build of mirage-skeleton?

Cheers
Adam

rjbou:
@adamsteen To test it, the simplest, cleaner way is to add @UnixJunkie opam-repo in an empty switch:
```
$ opam switch create pkg_config --empty #no compiler needed to test just the conf package
$ opam repo add unixjunkie git+https://github.com/UnixJunkie/opam-repository#pkg-config-openbsd
$ opam show conf-pkg-config #here if you have pkg-config-openbsd on the depends, it confirms that you have the good repo
$ opam install conf-pkg-config
```
After, to revert those changes, you just have to remove the repo and the switch
```
$ opam repo remove unixjunkie
$ opam switch <your normal switch>
$ opam switch remove pkg_config
```

mato:
@rjbou @UnixJunkie I still don't entirely follow the reasoning behind this change. If the newly introduced `pkg-config-openbsd` package also has a `conf` flag (added in 38bf7f294), does that not _also_ make it subject to the "must not install anything" requirement? If it does not, then why the need for the extra package at all?

rjbou:
https://github.com/ocaml/opam-repository/commit/38bf7f2943e4790a3bcaf19d5a1211f5e866cbaa is a `conf-pkg-config` change, `pkg-config-openbsd` have just the [`light-uninstall` flag](https://github.com/ocaml/opam-repository/pull/12904/files#diff-acbf7a833eaa57b9983fe9ca0937d868R27).

packages/pkg-config-openbsd/pkg-config-openbsd.1.0/opam
```diff
@@ -0,0 +1,27 @@
...
+post-messages: [
+  "pkg-config-openbsd: A symlink to /usr/local/bin/pkgconf has been installed in the OPAM bin directory (%{bin}%) on your PATH as 'pkg-config'." {os = "openbsd"}
+]
+depexts: [
+  ["devel/pkgconf"]
+]
+available: [os = "openbsd"]
+synopsis: "OpenBSD fix for conf-pkg-config"
+description: """
+This package can only install if the pkgconf package is installed
```
rjbou: Erroneous (copied) description ?

UnixJunkie: Can't you edit the PR if you spot a typo?

UnixJunkie: What is the correction that you are suggesting?

rjbou: It was just about the description that was not the right one for this package. Updated

madroach:
> @adamsteen To test it, the simplest, cleaner way is to add @UnixJunkie opam-repo in an empty switch:
> 
> ```
> $ opam switch create pkg_config --empty #no compiler needed to test just the conf package
> $ opam repo add unixjunkie git+https://github.com/UnixJunkie/opam-repository#pkg-config-openbsd
> $ opam show conf-pkg-config #here if you have pkg-config-openbsd on the depends, it confirms that you have the good repo
> $ opam install conf-pkg-config
> ```
> 
> After, to revert those changes, you just have to remove the repo and the switch
> 
> ```
> $ opam repo remove unixjunkie
> $ opam switch <your normal switch>
> $ opam switch remove pkg_config
> ```

Tested on OpenBSD and working. The only think I noticed is that the ``post-messages`` is output even when ``pkgconf`` was not installad and the symlinking failed.

UnixJunkie:
maybe you can constrain with a {success} flag the post-message?

avsm:
Hm, I don't like this approach since it requires yet another manual step for the user.  It seems that it really ought to be a base package on openbsd. *not* a conf- package that has a side-effect. If it's a base package that is passed as part of the `opam init` or `opam switch` then it will be installed reliably on all switches as well.

dra27:
This whole fix (also in #12018) feels incorrect - even with a message, why should using `opam` require an OpenBSD user to forgo their system `pkg-config`? i.e. why should opam-repository ever be creating the symlink?

Isn't this solvable by the Mirage packages themselves using a mix of `build-env` (to set `PKG_CONFIG`, if it needs passing through to other build systems) or package variables to provide the name of the `pkg-config` binary to OCaml packages themselves?

UnixJunkie:
I'm closing this issue since it is too old and not agreed upon.

PR #12905, opened 2018-11-01
Put upper bound on cairo2 dependency

jacquev6:
An non-backward-compatible change was introduced in cairo2 0.6
https://github.com/Chris00/ocaml-cairo/commit/9aa9ce403fd16c56245c695eb0108aebdedec150#diff-d9fad5803a4c2c22f5c1be3854b69e50

This should fix https://github.com/jacquev6/DrawGrammar/issues/7

PR #12906, opened 2018-11-01
Put upper bound on cairo2 dependency (opam 1.2)

jacquev6:
This is a backport of #12905 to the 1.2 branch


kit-ty-kate:
mmh `cairo2.0.6` doesn't exist in the `1.2` branch and backports to this branch should be kept for "critical fixes". I'm considering closing this PR, would that be ok for you?

jacquev6:
That's OK assuming the same policy applies to cairo2 forever, i.e. version 0.6 will never appear on master.

Thanks!

kit-ty-kate:
> i.e. version 0.6 will never appear on master.

it is already on `master` but not on `1.2`

jacquev6:
Absolutely, my bad. I typed too fast.

Things are good just as they are.

PR #12913, opened 2018-11-01
Add package mlmpfr.4.0.0

thvnx:
This commit introduces mlmpfr 4.0.0, some OCaml bindings for MPFR 4.0.0.
This version is an update of the opam package mlmpfr.3.1.6 which was release for MPFR 3.1.6.

The new features of mlmpfr 4.0.0 are:
- root function replaced by rootn_ui
- support failthful rounding mode support
- support for mpfr_flags_* functions
- support mpfr_beta
- support mpfr_fmodquo
- support mpfr_fmma and mpfr_fmms
- support mpfr_log_ui and mpfr_gamma_inc
- support mpfr_rint_roundeven and mpfr_roundeven

See https://github.com/thvnx/mlmpfr/tree/release_400.

### DIFF
--- a/packages/mlmpfr/mlmpfr.4.0.0/files/test.c
+++ b/packages/mlmpfr/mlmpfr.4.0.0/files/test.c
@@ -1,0 +1,19 @@
+#include <string.h>
+#include <stdio.h>
+#include <mpfr.h>
+
+int main(int argc, char **argv)
+{
+  const char *version = mpfr_get_version();
+  char subversion[5];
+  memcpy(subversion, version, 5);
+  subversion[5] = '\0';
+
+  if(strcmp(subversion, "4.0.0") == 0)
+    return 0;
+
+  if(strcmp(subversion, "4.0.1") == 0)
+    return 0;
+
+  return 1;
+}
--- a/packages/mlmpfr/mlmpfr.4.0.0/opam
+++ b/packages/mlmpfr/mlmpfr.4.0.0/opam
@@ -8,7 +8,7 @@ bug-reports:  "https://github.com/thvnx/mlmpfr/issues"
 license:      "LGPL-3.0"
 dev-repo: "git+https://github.com/thvnx/mlmpfr.git"
 build: [
-  ["cc" "test.c" "-lmpfr"]
+  ["cc" "test.c" "-lmpfr" "&&" "./a.out"]
   ["oasis" "setup"]
   ["./configure" "--prefix=%{prefix}%"]
   [make]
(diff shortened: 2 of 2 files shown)


thvnx:
I expect validation failures due to the fact that MPFR 4.0.0 is required by mlmpfr and probably not installed on validation machines.

thvnx:
I've added travis CI on my [repo](https://github.com/thvnx/mlmpfr) if you want to check that this package is actually working well.

packages/mlmpfr/mlmpfr.4.0.0/opam
```diff
@@ -0,0 +1,31 @@
...
+  ["oasis" "setup"]
+  ["./configure" "--prefix=%{prefix}%"]
+  [make]
+]
+install: [make "install"]
+remove:  ["ocamlfind" "remove" "mlmpfr"]
+depends: [
+  "ocaml" {>= "4.04"}
+  "ocamlfind"
+  "oasis"
```
mseri: Are these two just build dependencies?

mseri: You need to specify `extra-files` and then opam will fetch them for you, then you have to make sure in your project that the extra files are actually used. If you grep for it in the opam-repository there are some examples, but I can't remember the names specifically.

thvnx: OK, thanks! I'll take a look at it.

packages/mlmpfr/mlmpfr.4.0.0/files/test.c
```diff
@@ -0,0 +1,19 @@
+#include <string.h>
+#include <stdio.h>
+#include <mpfr.h>
+
+int main(int argc, char **argv)
+{
+  const char *version = mpfr_get_version();
```
mseri: If you also want to account for the missing library, you could add
```
#ifndef MPFR_FLAGS_ALL
   return 2
#else
   const char ...
   ...
#endif
```

mseri:
Other packages do `"sh" "-exc"` as you were doing. I don't understand why it didn't fail

packages/mlmpfr/mlmpfr.4.0.0/opam
```diff
@@ -8,7 +8,7 @@ bug-reports:  "https://github.com/thvnx/mlmpfr/issues"
 license:      "LGPL-3.0"
 dev-repo: "git+https://github.com/thvnx/mlmpfr.git"
 build: [
-  ["cc" "test.c" "-lmpfr"]
+  ["cc" "test.c" "-lmpfr" "&&" "./a.out"]
```
mseri: You can add them to two separate lines:
```
["cc" "test.c" "-lmpfr" "-o" "test_installed"]
["./test_installed"]

thvnx: Thanks!

thvnx:
I think its fixed now?

mseri:
I think it is fine now. I have tested it locally and it is working

thvnx:
Thanks again for your help!

PR #12914, opened 2018-11-01
[new release] conf-glfw3 (2)

SylvainBoilard:
Added an external dependency for mesa-libGL-devel on CentOS which provides the OpenGL header files used in the GLFW headers.

I will also raise the issue on the CentOS bug tracker since this looks like a dependency problem in their GLFW package.


mseri:
Thanks

PR #12910, opened 2018-11-01
[new release] dune (1.5.0)

rgrinberg:
CHANGES:

- Filter out empty paths from `OCAMLPATH` and `PATH` (ocaml/dune#1436, @rgrinberg)

- Do not add the `lib.cma.js` target in lib's directory. Put this target in a
  sub directory instead. (ocaml/dune#1435, fix ocaml/dune#1302, @rgrinberg)

- Install generated OCaml files with a `.ml` rather than a `.ml-gen` extension
  (ocaml/dune#1425, fix ocaml/dune#1414, @rgrinberg)

- Allow to use the `bigarray` library in >= 4.07 without ocamlfind and without
  installing the corresponding `otherlib`. (ocaml/dune#1455, @nojb)

- Add `@all` alias to build all targets defined in a directory (ocaml/dune#1409, fix
  ocaml/dune#1220, @rgrinberg)

- Add `@check` alias to build all targets required for type checking and tooling
  support. (ocaml/dune#1447, fix ocaml/dune#1220, @rgrinberg)

- Produce the odoc index page with the content wrapper to make it consistent
  with odoc's theming (ocaml/dune#1469, @rizo)

- Unblock signals in processes started by dune (ocaml/dune#1461, fixes ocaml/dune#1451,
  @diml)

- Respect `OCAMLFIND_TOOLCHAIN` and add a `toolchain` option to contexts in the
  workspace file. (ocaml/dune#1449, fix ocaml/dune#1413, @rgrinberg)

- Fix error message when using `copy_files` stanza to copy files from
  a non sub directory with lang set to dune < 1.3 (ocaml/dune#1486, fixes ocaml/dune#1485,
  @NathanReb)

- Install man pages in the correct subdirectory (ocaml/dune#1483, fixes ocaml/dune#1441, @emillon)

- Fix version syntax check for `test` stanza's `action` field. Only
  emits a warning for retro-compatibility (ocaml/dune#1474, fixes ocaml/dune#1471,
  @NathanReb)

- Fix interpretation of paths in `env` stanzas (ocaml/dune#1509, fixes ocaml/dune#1508, @diml)

- Add `context_name` expansion variable (ocaml/dune#1507, @rgrinberg)

- Use shorter paths for generated on-demand ppx drivers. This is to
  help Windows builds where paths are limited in length (ocaml/dune#1511, fixes
  ocaml/dune#1497, @diml)

- Fix interpretation of environment variables under `setenv`. Also forbid
[...]

PR #12918, opened 2018-11-02
cohttp-lwt{,-jsoo}: add missing sexplib dependency

mseri:
Discovered by https://github.com/ocaml/opam-repository/pull/12910

### DIFF
--- a/packages/cohttp-lwt-jsoo/cohttp-lwt-jsoo.1.1.1/opam
+++ b/packages/cohttp-lwt-jsoo/cohttp-lwt-jsoo.1.1.1/opam
@@ -16,7 +16,7 @@ bug-reports: "https://github.com/mirage/ocaml-cohttp/issues"
 depends: [
   "ocaml" {>= "4.03.0"}
   "dune" {build & >= "1.1.0"}
-  "cohttp" {>= "1.0.0"}
+  "cohttp" {>= "1.0.0" & < "1.2.0"}
   "cohttp-lwt" {>= "1.0.0"}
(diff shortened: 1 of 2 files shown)


mseri:
The failure for `calculon-web.0.2` has nothing to do with this PR. Travis osx failure is the usual one

kit-ty-kate:
Would there be a simpler fix to just constrain `ppx_sexp_conv` to `< "v0.11.0"` in `cohttp-lwt.1.1.1` as `v0.11.0` now depends over `sexplib0` and not `sexplib` anymore?

mseri:
Yes but this other fix is forward and backward compatible

kit-ty-kate:
well if it doesn't work in the current state it's not really forward compatibility

mseri:
Why not? With this fix, the package works independently of the version of ppx_sexp_conv. Also I believe this fix is more correct, as both packages are explicitly calling Sexplib functions but not declaring it as a delendency

kit-ty-kate:
Well, in the current state the package doesn't compile so it's not really forward compatibility.
Plus I'm not very fond of having more patches added to the repository. Even more if it's an old version we're fixing.

mseri:
We are disagreeing on terminology. Anyways it's not such a big deal. I have tested it locally and the issue is that sexplib was linked transitively via uri that was coming transitively from cohttp. From the test I have done seems that the right fix is to add a bound to cohttp.

mseri:
Looks like it wiorked

kit-ty-kate:
Thanks!

PR #12920, opened 2018-11-03
add +spacetime switches for 4.07.*

cmr:
is there some automation or templates that generate these so they don't get forgotten?


kit-ty-kate:
Thanks!

mseri:
After an `opam update` this morning `opam` has decided that in all my `4.07` switches it needed to compile and install `ocaml-variants.4.07.1+spacetime`. I fear this is not expected, I would like to understand if this happened by chance or we have a bug

mseri:
opam says: `∗ install ocaml-variants 4.07.1+spacetime [required by ocaml]`

PR #12922, opened 2018-11-03
[new release] bwrap (0.1)

Chris00:
Use Bubblewrap to sandbox executables

- Project page: <a href="https://github.com/Chris00/ocaml-bwrap">https://github.com/Chris00/ocaml-bwrap</a>
- Documentation: <a href="https://Chris00.github.io/ocaml-bwrap/doc">https://Chris00.github.io/ocaml-bwrap/doc</a>

##### CHANGES:

- Initial release.

### DIFF
--- a/packages/bwrap/bwrap.0.1/opam
+++ b/packages/bwrap/bwrap.0.1/opam
@@ -1,0 +1,43 @@
+opam-version: "2.0"
+maintainer: "Christophe Troestler <Christophe.Troestler@umons.ac.be>"
+authors: ["Christophe Troestler <Christophe.Troestler@umons.ac.be>"]
+tags: ["sandbox" "bubblewrap"]
+license: "ISC"
+homepage: "https://github.com/Chris00/ocaml-bwrap"
+dev-repo: "git+https://github.com/Chris00/ocaml-bwrap.git"
+bug-reports: "https://github.com/Chris00/ocaml-bwrap/issues"
+doc: "https://Chris00.github.io/ocaml-bwrap/doc"
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "build" "@doc"] {with-doc}
+]
+run-test: [
+  ["dune" "runtest" "-p" name "-j" jobs]
+]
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
I fixed the issue with the tests. They won't be run as `with-test` is not set inside the `run-test` field.

I'm going to wait for the CI to finish and merge it afterwards

Chris00:
> They won't be run as with-test is not set inside the run-test field.

Thanks.  But shouldn't `["dune" "runtest" "-p" name "-j" jobs]` stay inside `run-test`?

Chris00:
Note that the documentation should be corrected [as it says](https://opam.ocaml.org/doc/Manual.html#Package-variables):

> Additionally, the following are limited to some package fields (depends:, depopts:, build:, install:, remove:, run-test:):
>
>    with-test: only true if tests have been enabled for this specific package
>    with-doc: similarly for documentation

Chris00:
I have left the tests within `run-test` and marked the package Linux-only.

Chris00:
Seems the tests are not run... but it seems to me that it is the right place to put the test instructions.

Chris00:
Travis CI: note that, when a package is not marked available on a platform, the build should quit immediately.

Chris00:
Correction: the tests are run on Travis — but not on ci/datakit.

Chris00:
The tests fail on Travis because the bubblewrap package is at version 0.1.7 while the current one is 0.3.1.

Chris00:
Well, option `--die-with-parent` is neither supported by old bwrap versions... :-(

mseri:
The tests are failing in Travis, but I think this has to be expected:
```
- No permissions to creating new namespace, likely because the kernel does not allow non-privileged user namespaces. On e.g. debian this can be enabled with 'sysctl  kernel.unprivileged_userns_clone=1'.
```

For what concerns me this could be merged. @Chris00 are you happy with it?

Chris00:
> The tests are failing in Travis, but I think this has to be expected:

Yes, there is nothing I can do about this.  For me it is good to merge.

mseri:
Thanks

PR #12927, opened 2018-11-04
Add a relit as an alias for ppx_relit

charlesetc:
Hi! I'm not sure what the status quo is for package aliases. Worst case scenario, it's totally fine just to have `ppx_relit` and `relit_helper` be separate, but it'd be nice to be able to run `opam install relit` and get the right packages. Does this sound reasonable? Thanks!

### DIFF
--- a/packages/relit/relit.0.2.0/opam
+++ b/packages/relit/relit.0.2.0/opam
@@ -8,17 +8,9 @@ homepage: "https://github.com/cyrus-/relit"
 bug-reports: "https://github.com/cyrus-/relit/issues"
 license: "MIT"
 dev-repo: "git://github.com/cyrus-/relit.git"
-build: []
 depends: [
   "ppx_relit" {= "0.2.0"}
   "relit_helper" {= "0.2.0"}
   "ocaml" {>= "4.07.0" & < "4.08.0"}
 ]
 synopsis: "An implementation of Typed Literal Macros for Reason"
-url {
-  src: "https://github.com/cyrus-/relit/archive/0.2.0.tar.gz"
-  checksum: [
-    "md5=c2350b29ed2b2005846e0a60f1790de8"
-    "sha512=64553171caec9a0348d2327880722b5f860ce9c42b156e14058b4ad5c37b57583f2ca552f2f6b63e714b42d8105358a32274b07cd5c854e7345d31334b05e09b"
-  ]
-}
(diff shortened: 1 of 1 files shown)


packages/relit/relit.0.2.0/opam
```diff
@@ -0,0 +1,24 @@
...
+maintainer: "Charles Chamberlain <charlespipin@gmail.com>"
+authors: [
+  "Charles Chamberlain <charlespipin@gmail.com>"
+  "Cyrus Omar <cyrus.omar@gmail.com>"
+]
+homepage: "https://github.com/cyrus-/relit"
+bug-reports: "https://github.com/cyrus-/relit/issues"
+license: "MIT"
+dev-repo: "git://github.com/cyrus-/relit.git"
+build: []
```
mseri: You can remove the empty build stanza directly

packages/relit/relit.0.2.0/opam
```diff
@@ -0,0 +1,24 @@
...
+depends: [
+  "ppx_relit" {= "0.2.0"}
+  "relit_helper" {= "0.2.0"}
+  "ocaml" {>= "4.07.0" & < "4.08.0"}
+]
+synopsis: "An implementation of Typed Literal Macros for Reason"
+url {
```
mseri: You don't need url, this "meta" package is not building anything

mseri:
I don't know if there is a specific policy or opam flag for this kind of meta packages. @rjbou or @kit-ty-kate do you know?

rjbou:
@mseri, no opam flag dedicated to, there is [_transition_](https://opam.ocaml.org/doc/Tricks.html#How-to-rename-a-package) packages for rename, but for aliases, afaik, no process defined

mseri:
Ok, thanks. Then I have nothing against merging this. Do you agree?

kit-ty-kate:
I wouldn't be against personally. However if you could explicitly say this is a meta/alias package in the synopsis it would be perfect

hannesm:
I'm not very convinced of `aliases` in opam-repository. While I understand that it easier for a user to type `opam install relit`, when they want to use it, they will need to learn about a new package name `ppx_relit`, and when they want to use the (helper library?), the need to depend on `relit_helper`. Wouldn't it be more natural to rename `relit_helper` to `relit`? In `relit` you can (build-?)depend on `ppx_relit`, so users will get both if they `opam install relit`.

Additionally, from an opam side, the universe grows (for imho no obvious reason), and the namespace gets more populated.

charlesetc:
Yeah I think that's better, thanks!

The relit ppx doesn't depend on relit_helper but everyone who's using the relit ppx will also probably use the relit_helper in one way or another. Next version bump I'll rename it to relit instead, thanks again!

PR #12928, opened 2018-11-05
Package resource-pooling.0.5.1

jrochel:
### `resource-pooling.0.5.1`
Library for pooling resources like connections, threads, or similar
This package is derived from the module Lwt_pool from the lwt package, which implements resource pooling. With Resource_pool this package provides a modified version with additional features. Also there is a module called Server_pool that manages resource clusters, specifically a cluster of servers each with its own connection pool.

---
* Homepage: https://github.com/ocsigen/resource-pooling
* Source repo: git+https://github.com/ocsigen/resource-pooling.git
* Bug tracker: https://github.com/ocsigen/resource-pooling/issues

---
:camel: Pull-request generated by opam-publish v2.0.0

PR #12824, opened 2018-10-18
Package prepare.0.5.1

jrochel:
### `prepare.0.5.1`
Library for pooling resources like connections, threads, or similar
This package is derived from the module Lwt_pool from the lwt package, which implements resource pooling. With Resource_pool this package provides a modified version with additional features. Also there is a module called Server_pool that manages resource clusters, specifically a cluster of servers each with its own connection pool.

---
* Homepage: https://github.com/ocsigen/resource-pooling
* Source repo: git+https://github.com/ocsigen/resource-pooling.git
* Bug tracker: https://github.com/ocsigen/resource-pooling/issues

---
:camel: Pull-request generated by opam-publish v2.0.0

PR #12929, opened 2018-11-05
Revert "Package prepare.0.5.1 and prepare.0.5"

mseri:
Reverts ocaml/opam-repository#12824

The package name is incorrect. The correct PR is the resource-pooling one: https://github.com/ocaml/opam-repository/pull/12928


mseri:
Should we simply remove the package as in this PR, or we need to replace it with something unavailable and pointing to the correct one?

kit-ty-kate:
Removing is good. Thanks!

PR #12939, opened 2018-11-06
hardens the installation of bap on Fedora

gitoleg:
This PR solves problems with the installation of `bap` on Fedora


kit-ty-kate:
Thanks!

PR #12944, opened 2018-11-07
Mirage/Solo5: Sync OPAM metadata with master, upper bounds

mato:
- solo5-bindings-*: Sync metadata with that on master. Adds conflicts
for upcoming solo5-bindings-genode, drop ocaml dependency here (handled
by ocaml-freestanding).
- mirage-{block,net}-solo5: Add upper bounds for mirage-solo5.


hannesm:
thanks!

PR #12950, opened 2018-11-08
Solo5: Release 0.4.1, various Mirage/Solo5 updates

mato:
/cc mirage/mirage#939

### DIFF
--- a/packages/mirage-block-solo5/mirage-block-solo5.0.4.0/opam
+++ b/packages/mirage-block-solo5/mirage-block-solo5.0.4.0/opam
@@ -1,0 +1,30 @@
+opam-version: "2.0"
+maintainer:   "martin@lucina.net"
+homepage:     "https://github.com/mirage/mirage-block-solo5"
+dev-repo:     "git+https://github.com/mirage/mirage-block-solo5.git"
+bug-reports:  "https://github.com/mirage/mirage-block-solo5/issues"
+authors:      ["Dan Williams" "Martin Lucina"]
+tags: [
+  "org:mirage"
+]
+build: [
+  ["ocaml" "pkg/pkg.ml" "build" "--pinned" "%{pinned}%"]
+]
+depends: [
+  "ocaml" {>= "4.04.2"}
+  "ocamlfind"  {build}
+  "ocamlbuild" {build}
+  "topkg"      {build}
+  "lwt" {>= "2.4.3"}
+  "cstruct" {>= "1.0.1"}
+  "mirage-block-lwt" {>= "1.0.0"}
+  "mirage-solo5" {>= "0.5.0" & < "0.6.0"}
+  "fmt"
+]
+synopsis: "Solo5 implementation of MirageOS block interface"
+description:
+  "This library implements the MirageOS block interface for Solo5 targets."
+url {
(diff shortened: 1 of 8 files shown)


mato:
Reopened and force-pushed 41f2e360a227649ce4ca0132cc05c67a5f1ab952, this will hopefully work now.

mato:
It looks like `alpine` is failing due to either the `linux-headers` depext on `solo5-bindings-hvt` being wrong or not getting installed. I can't tell from the log which.

mato:
... it seems to be the latter, running from an `alpine:latest` docker image `apk add build-base linux-headers` is sufficient to get `make hvt` to complete successfully.

mato:
@avsm AFAICT the Alpine CI is not running a `depext` step at all. Unclear where to report issues with the Datakit-based CIs, can you look into it please?

kit-ty-kate:
> @avsm AFAICT the Alpine CI is not running a `depext` step at all. Unclear where to report issues with the Datakit-based CIs, can you look into it please?

Mmh no the only failure on Alpine is for `mirage-net-solo5.0.4.0` and this package doesn't have `linux-headers` as a depexts, so of course it won't be installed.

hannesm:
@kit-ty-kate AFAIU: mirage-net-solo5 depends on mirage-solo5 (which provides the C symbols) -- and this one depends on a disjunction of solo5-bindings-* -- where the one selected here (solo5-bindings-hvt) has a depext on linux-headers. the compilation of solo5-bindings-hvt fails (due to missing linux/if.h). not sure what the root of the issue is (and mirage-solo5-block is very similar in terms of dependencies, but this one compiles fine)

kit-ty-kate:
Oh... I see where is the problem. It must be somewhere inside opam-depext.. Not sure how to fix it though. I'll look at it after lunch.

rjbou:
Internally,  `opam depext` calls `opam list --resolve` to get all their external dependencies and then calls `opam install`. The thing is that opam list and install don't call the solver the same way, with the same universe (install take into account current switch) , it results on two different choices between them.

hannesm:
thanks for looking into this @rjbou. how should we move forward?
I tried to click on `rebuild` on the CI, which now leads to a different error message which I don't fully understand `pull access denied for sha256, repository does not exist or may require 'docker login'`

since the issue only happens sometimes on alpine (or will it happen always on alpine? why are other distributions not affected?), I'd suggest to merge this PR as is, and open an issue on the opam issue tracker (or opam-depext) regarding @rjbou comment "opam list and install don't call the solver the same way, with the same universe (install take into account current switch) , it results on two different choices between them" -- which IIUC should be fixed to avoid potential problems in the future (with the universe getting bigger, my intuition is that there will be more divergence).

WDYT? especially, I can't see a way to apply any fix on this PR for the issue at hand.

kit-ty-kate:
Should be fine to merge I guess. We can test this later anyway. The reason this only fails on Alpine is that on all other distro the required headers are already installed by default.
The reason for the mysterious second error message is that docker seems to remove the base images after some time (I have no idea why) and the base images are not accessible anymore. One way to fix that is to rebuild up to the base image of the branch you want to test (the rebuild button one step above most of the time)

Anyway, thanks!

PR #12958, opened 2018-11-10
[new release] earley (2.0.0)

rlepigre:
Earley parser combinator library

- Project page: <a href="https://github.com/rlepigre/ocaml-earley">https://github.com/rlepigre/ocaml-earley</a>
- Documentation: <a href="https://rlepigre.github.io/ocaml-earley/">https://rlepigre.github.io/ocaml-earley/</a>

##### CHANGES:

New major release, introducing incompatibilities due to wrapping of library
modules. Here are the most important changes:
 - Switch to `dune` (build system).
 - New (ocamlfind) packing: `earley.core`, `earley.str`, `earley.ocaml`.
 - Modules are wrapped in packages (`Earley` is now `Earley_core.Earley`).
 - `EarleyStr` is now `Earley_str`.


packages/earley/earley.2.0.0/opam
```diff
@@ -0,0 +1,32 @@
...
+homepage: "https://github.com/rlepigre/ocaml-earley"
+dev-repo: "git+https://github.com/rlepigre/ocaml-earley.git"
+authors: [
+  "Christophe Raffalli <christophe@raffalli.eu>"
+  "Rodolphe Lepigre <rodolphe.lepigre@inria.fr>" ]
+license: "CeCILL-B_V1"
+doc: "https://rlepigre.github.io/ocaml-earley/"
+
+depends: [
+  "ocaml" { >= "4.03.0" & <= "4.07.1" }
```
kit-ty-kate: Why is it constrained for 4.08? If there is a good reason for this could you at least have the upper-bound to `< "4.08"` instead? It would probably make more sense if there is a new security release 4.07.2 or something

rlepigre: I know this is a bit weird, but I don't really have a choice here. We need to generate version-specific code, and we also rely on `compiler_libs` code that is marked unstable (changes may occur between minor versions, this append before.).

PR #12960, opened 2018-11-10
duppy.0.7.4

toots:


toots:
Question: What is the proper way to push this to legacy `1.2` opam users? This release has an important bugfix..

kit-ty-kate:
mmh not sure why `camlp4` fails here. cc @dra27 ?

toots:
Ok. I was planning on removing the `camlp4` part, I'll do another release without it soon.

dra27:
Something looks wrong with the builds for the specific compiler versions (especially given that macOS is passing). The local test would be expected to fail - that needs updating to build a "system" camlp4.

toots:
Closing, see #12977

PR #12962, opened 2018-11-10
[new release] ocal (0.2.2)

mor1:
An improved Unix `cal` utility

- Project page: <a href="https://github.com/mor1/ocal">https://github.com/mor1/ocal</a>
- Documentation: <a href="https://mor1.github.io/ocal/">https://mor1.github.io/ocal/</a>

##### CHANGES:

* Dunification

### DIFF
--- a/packages/ocal/ocal.0.2.2/opam
+++ b/packages/ocal/ocal.0.2.2/opam
@@ -21,7 +21,6 @@ depends: [
   "dune"      {build}
   "notty"     {build & >="0.2.0"}
   "ocaml"     {>="4.02.3"}
-  "ocamlfind" {build}
 ]
 
 synopsis: "An improved Unix `cal` utility"
--- a/packages/ocal/ocal.0.2.2/opam
+++ b/packages/ocal/ocal.0.2.2/opam
@@ -1,0 +1,37 @@
+opam-version: "2.0"
+maintainer: "Richard Mortier <mort@cantab.net>"
+authors: [ "Richard Mortier" ]
+license: "ISC"
+
+homepage: "https://github.com/mor1/ocal"
+dev-repo: "git+https://github.com/mor1/ocal.git"
+bug-reports: "https://github.com/mor1/ocal/issues"
+doc: "https://mor1.github.io/ocal/"
+
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+]
+
+depends: [
+  "astring"   {build}
+  "calendar"  {build}
+  "cmdliner"  {build}
+  "dune"      {build}
+  "notty"     {build & >="0.2.0"}
+  "ocaml"     {build & >="4.02.3"}
+  "ocamlfind" {build}
+]
+
+synopsis: "An improved Unix `cal` utility"
+
+description: """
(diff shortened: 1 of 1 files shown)


packages/ocal/ocal.0.2.2/opam
```diff
@@ -0,0 +1,40 @@
...
+bug-reports: "https://github.com/mor1/ocal/issues"
+doc: "https://mor1.github.io/ocal/"
+
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+]
+
+run-test: [
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
```
kit-ty-kate: This doesn't work. It should be either tagged with `{with-test}` inside `build` or without inside `run-test`.

kit-ty-kate:
Fixed. Could you return the fix upstream?

packages/ocal/ocal.0.2.2/opam
```diff
@@ -0,0 +1,37 @@
...
+]
+
+depends: [
+  "astring"   {build}
+  "calendar"  {build}
+  "cmdliner"  {build}
+  "dune"      {build}
+  "notty"     {build & >="0.2.0"}
+  "ocaml"     {build & >="4.02.3"}
```
kit-ty-kate: mmh, something tells me `ocaml` shouldn't be a build dependency. What if `dune` decides to compile the executable to bytecode if the platform doesn't support native code?

mor1: This is as recommended by `dune-release lint`:
```
: mort@greyjay:ocal#master= $; make lint
dune build @lint
dune-release lint
[FAIL] lint opam file ocal.opam:
       opam lint -s messages:
       /Users/mort/src/ocal/ocal.opam: Errors.
                    error 32: Field 'ocaml-version:' and variable 'ocaml-version' are deprecated, use a dependency towards the 'ocaml' package instead for availability, and the 'ocaml:version' package variable for scripts
[FAIL] lint /Users/mort/src/ocal failure: 1 errors.
make: *** [lint] Error 1
```
(Or at least, that's how I interpreted the error message! Am I supposed to do something else here?)

packages/ocal/ocal.0.2.2/opam
mor1: Ah ok -- I think this was me misinterpreting the error from `dune-release lint` indicating what to do to remove the `build-test` field. Will fix!

packages/ocal/ocal.0.2.2/opam
kit-ty-kate: oh sorry, I should have been more clear. I meant to insist on the `build` dependency part, not the `dependency` part of the sentence ^^

OCaml should of course be a dependency but not flagged as a `{build}` dependency probably. Anyone has a thought on that btw? @rjbou @rgrinberg?

rgrinberg: Setting `ocaml` as `{build}` seems wrong. At the very least, the compiler contains pervasives and it's highly unlikely that Pervasives are only used for building a package.

kit-ty-kate:
Sorry I just noticed the thing about the unneeded ocamlfind dependency. Could you return the fix upstream?

I'll merge it tomorrow

kit-ty-kate:
Thanks!

PR #12974, opened 2018-11-13
Update existing package descriptions for nunchaku and frenetic.

fpottier:
nunchaku: update existing packages to require menhir <= 20181026.
frenetic: update existing packages to require menhir <= 20181026.
          also note that menhir is only a build dependency.


kit-ty-kate:
Thanks!

PR #12976, opened 2018-11-13
ffmpeg.0.2.1

toots:


toots:
`ffmpeg-devel` isn't an official package on centos. However, it is the unofficial package for it and, as for the previous releases, it seems like a good thing to have on depext for the user to know what to install.

toots:
As for travis, it looks like the distribution used is a Ubuntu from 4 years ago so one of the depext package is missing there.

kit-ty-kate:
looks fine. Thanks!

PR #12977, opened 2018-11-13
duppy.0.8.0

toots:
This PR updates `ocaml-duppy` with a release without `camlp4` which seems to have been buggy. It also carries over an important bugfix from `0.7.4` which would be nice to see pushed soon. Accordingly, `liquidsoap.0.3.4` dependency on `duppy` has been bumped to make sure that the fixed version is used.


kit-ty-kate:
Thanks a lot!

mseri:
@toots I just noticed that the examples are still using the camlp4 syntax. Just wanted to let you know, I will not be able to send a PR before Friday or Monday

PR #12978, opened 2018-11-13
afl-persistent.1.3

stedolan:
New release of afl-persistent.

The only change is to replace a `/bin/bash` with `/bin/sh`, making the package installable on bash-less systems.

cc @hannesm


kit-ty-kate:
This package doesn't seem to use `ocamlfind`. I've just fix that in the PR. Could you return the fix upstream?

kit-ty-kate:
Thanks!

stedolan:
> This package doesn't seem to use ocamlfind. I've just fixed that in the PR. Could you return the fix upstream?

Done, thanks.

PR #12980, opened 2018-11-14
duppy.0.8.0 for opam 1.2.2

toots:
Please consider an update of this package in the `1.2` branch. It fixes an important bug in `liquidsoap` and it would be nice to also push it to users of the old version.


kit-ty-kate:
Looks fine. Thanks!

PR #12998, opened 2018-11-17
Package lmdb.0.1

Drup:
### `lmdb.0.1`
Bindings for LMDB, a fast in-file database with ACID transactions

---
* Homepage: https://github.com/Drup/ocaml-lmdb
* Source repo: git+https://github.com/Drup/ocaml-lmdb.git
* Bug tracker: https://github.com/Drup/ocaml-lmdb/issues

---


Drup:
The library doesn't currently has any versioning for the underlying C library, and is likely to fail for distributions that use old versions of lmdb (debian, ubuntu, ...).

Drup:
So, Centos doesn't have a package for lmdb, and for some reason, alpine doesn't install the `.h`.

Drup:
As far as I'm concerned, this is as green as it will ever be.

hannesm:
thanks

PR #13000, opened 2018-11-18
move mirage-xen to correct subfolder

hannesm:
oops, I wish @camelus would complain if you put package `x` into `y/x.a.b.c` (here, x being mirage-xen, and y mirage-unix)


hannesm:
merging

PR #13017, opened 2018-11-20
Prevents 'opam install' from proposing javalib 3.0 in the future

anscoil:
- javalib 3.0 will not be compatible with this version of sawja
- a new version of sawja will be released just after javalib 3.0

### DIFF
--- a/packages/sawja/sawja.1.5.4/opam
+++ b/packages/sawja/sawja.1.5.4/opam
@@ -15,7 +15,7 @@ homepage: "http://sawja.inria.fr"
 depends: [
   "ocaml" {>= "4.02"}
   "ocamlfind" {build}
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Thanks!

anscoil:
Apparently there are 2 bugs:

- perl is not installed in some environments where CI takes place
- opam 2.0 is sandboxed and prevents us from installing ptrees with **ocamlfind install** during the installation of _javalib_ itself

For the first issue, I don't know what I can do, except not using perl in the Makefile in the future.

For the second issue, it is indeed a bad practice and this is corrected in the forthcoming version of _javalib_.

I've just seen that the request has been accepted anyway, so thank you.

PR #13019, opened 2018-11-20
Fix error 39

xclerc:


xclerc:
(I meant `opam lint` error.)

kit-ty-kate:
Thanks!

PR #13020, opened 2018-11-20
Fix error 31

xclerc:


xclerc:
(I meant `opam lint` error.)

kit-ty-kate:
mmh I need a bit more context for this one. Could you explain what was the error 31 or what the PR does?

xclerc:
Sorry, `opam lint` complains when the same package appears
in both `depends` and `depopts`. To be conservative, I removed
the occurrences in `depopts`.

kit-ty-kate:
oh, fair enough. Thanks a lot!

xclerc:
I beg to differ; in my opinion this does not scale.
I would recommend using `diffstat` on the commit
diff to see the list of impacted packages.

pmetzger:
> I beg to differ; in my opinion this does not scale.

How does it fail to scale? FYI, this is the rule in MacPorts which has a couple of orders of magnitude more packages managed in it, and it scales quite well.

xclerc:
You end up with huge commit messages, while as I said
the information is available through other means.

That being said, I am not overly attached to commit
messages, feel fee to rewrite them.

pmetzger:
> You end up with huge commit messages

Not really, no. You just end up with commit messages that give the reader enough information to know whether or not they wish to bother looking at the commit, which is an important goal of documentation, and commit messages are a form of documentation. However, I'm not willing to argue over it longer.

kit-ty-kate:
@pmetzger please stop arguing. Put it to the forum if you really want to continue..

avsm:
Indeed; historically the opam-repository commit messages have had no strong policy on commit hygiene, so the history is littered with all sorts of intermediate commits.  This in turn means that opam-repo is not a good target for tools such as git-bisect.

I would welcome an effort to fix this by designing a git-bisect-friendly workflow for opam-repository PRs, which in turn would lead to better commit messages.  This should not be done by picking on an individual PR, but by writing down rules that apply consistently across all commits. That's best done on discuss.ocaml.org instead of the issue tracker.

pmetzger:
@kit-ty-kate I already said I was ceasing to discuss this. It's unfriendly to tell someone to stop discussing something when they've already said they're not discussing it further.

avsm:
@pmetzger Your last message was an attempt to get the last word in with a counterargument, rather than simply indicating you were ceasing the discussion -- so please don't have a go at our hard working repository maintainers. I agree with @kit-ty-kate's message to indicate that you should move this elsewhere.  Apologies to @xclerc for the noise on your contribution via this PR.

pmetzger:
It's exceptionally impolite to impute psychological motives to other people, since one has no privileged access to the mental states of other people; one cannot know what others are thinking, and one has no way to know what their context might be like. Imputing motives to others also has the result of making a discussion personal, resulting in acceleration of arguments and bad feeling instead of cooling things off and reducing animosity. That said, I'm _truly_ dropping this. I recommend others do the same, before anything worse is said.

avsm:
Your original note to this PR was generic advice that did not consider the history of this particular repository.  You then didn't bother to comment on either of the repository maintainers replies with anything constructive. I'm disappointed with your interactions here, @pmetzger.

pmetzger:
The feeling is mutual.

PR #13027, opened 2018-11-21
opam-ed.0.1: relax opam bound, it is compatible with 4.02.3 (at least)

mseri:


kit-ty-kate:
Thanks!

PR #13030, opened 2018-11-21
Remove md5 sums for memprof variants of ocaml compiler, since these forks can be updated.

jhjourdan:


avsm:
Thanks; please do feel free to put the checksums back when you stop maintaining the older versions.

PR #13031, opened 2018-11-21
Add gapi-ocaml 0.3.8

astrada:


kit-ty-kate:
the `light-uninstall` flag was useless. Could you return the fix upstream?

kit-ty-kate:
Thanks!

astrada:
@kit-ty-kate sorry to bother you, but why I'm still not seeing version 0.3.8 when I do `opam update && opam upgrade`? I'm asking because in version 0.3.8 I fixed a severe bug that broke my application with http/2, so I would like to make it available as soon a possible. Thanks
screenshot from 2018-11-25 10-51-39

mseri:
@astrada Should be fixed now

astrada:
Thanks!

PR #13033, opened 2018-11-22
Package range.0.2

aldrikfr:
### `range.0.2`
Fold on integer range
Enable folding on integer range with split capacities
for making distributed computing easier with libraries
like functory

---
* Homepage: https://github.com/aldrikfr/range
* Bug tracker: https://github.com/aldrikfr/range/issues

---


aldrikfr:
Hi !
I'm experimenting spam packaging, and I'm a bit lost to understand the root cause of the error. Help would be appreciated if someone have spare time.

kit-ty-kate:
there are 2/3 issues in your package. First of all you don't sepecify the url of the archive (see other packages in the repo for more examples), then you are missing the `{build}` attribute to the `dune` dependency (as it is only used during build time), and also (but it's minor) you forgot to put `"-j" jobs` to your dune build command to use the full power of parallelisation

PR #13042, opened 2018-11-23
(do not merge) conf-pkg-config: add version 1.2 with setenv

emillon:
Hi,

I believe that when `conf-pkg-config` is installed, `pkg-config` should see libraries installed by opam itself in `%{lib}%/pkgconfig`. I described the use case in ocaml/opam#3676 and the linked dune issues.

This adds a warning because it is discouraged for non-compiler packages, but it looks like this is what we want here. But I'm a bit wary since this is the only such case in the repository, so I'm curious about the opinion of the repository maintainers (cc @UnixJunkie).

As for the patch itself, this makes sure that the `PKG_CONFIG_PATH` environment variable is properly extended in the opam environment, and in particular that `pkg-config` can automatically find the libraries installed by opam itself. In addition, the post-message changelog from 1.1 is removed.

What do you think about this?

Thanks!

### DIFF
--- a/packages/conf-pkg-config/conf-pkg-config.1.2/opam
+++ b/packages/conf-pkg-config/conf-pkg-config.1.2/opam
@@ -1,0 +1,36 @@
+opam-version: "2.0"
+maintainer: "unixjunkie@sdf.org"
+authors: ["Francois Berenger"]
+homepage: "http://www.freedesktop.org/wiki/Software/pkg-config/"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: "GPL"
+build: [
+  ["pkg-config" "--help"]
+]
+install: [
+  ["ln" "-s" "/usr/local/bin/pkgconf" "%{bin}%/pkg-config"] {os = "openbsd"}
+]
+remove: [
+  ["rm" "-f" "%{bin}%/pkg-config"] {os = "openbsd"}
+]
+setenv: PKG_CONFIG_PATH += "%{lib}%/pkgconfig:%{share}%/pkgconfig"
+depexts: [
+  ["pkg-config"] {os-distribution = "debian"}
+  ["pkg-config"] {os-distribution = "ubuntu"}
+  ["pkg-config"] {os-distribution = "archlinux"}
+  ["pkgconfig"] {os-distribution = "fedora"}
+  ["pkgconfig"] {os-distribution = "centos"}
+  ["pkgconfig"] {os-distribution = "mageia"}
+  ["pkgconfig"] {os-distribution = "rhel"}
+  ["pkgconfig"] {os-distribution = "oraclelinux"}
+  ["pkgconfig"] {os-distribution = "alpine"}
+  ["devel/pkgconf"] {os = "freebsd"}
+  ["devel/pkgconf"] {os = "openbsd"}
+  ["pkg-config"] {os = "macos" & os-distribution = "homebrew"}
+  ["pkgconf"] {os = "freebsd"}
+]
+synopsis: "Virtual package relying on pkg-config installation"
+description: """
+This package can only install if the pkg-config package is installed
+on the system."""
+flags: conf


hannesm:
This looks very sensible to me. I'm not sure about the `PKG_CONFIG_PATH` directory in opam -- should it be `share/pkgconfig` or `lib/pkgconfig`? (the former contains `mirage-solo5.pc`, the latter `gmp-freestanding.pc`, `ocaml-freestanding.pc`, and `solo5-bindings-hvt.pc`) -- @mato do you remember the reasoning why there is both `share` and `lib`?

emillon:
By default it scans at least `/usr/local/lib/pkgconfig` and `/usr/local/share/pkgconfig`, so we can use the same pattern and add both.

The usual distinction is whether the files are arch-dependent or not, but I'm not sure what this means in the context of mirage.

hannesm:
The current state is to have both `lib` and `share` - but I'm in favor of using only one of them (and cleanup packages that use the other one) unless there's good reason to support both.

emillon:
The concept of arch independence in the context of opam is a bit tricky, since I don't think it's commonly used in multi-arch systems. I am not sure whether there are cross-compilation concerns, but this might be related as well.

If we want to remove the distinction between arch-dependent and arch-independent files, the correct solution is to install everything in `lib`.

In the case of `mirage-solo5`, the `.pc` file (in `share/`) refers to a binary file in to `lib/`. I don't think that it's correct, since that file cannot be used on a different architecture, so this probably needs to be moved.

But moving these files can cause problems, since there is probably some code that manually sets `PKG_CONFIG_PATH` to just `share`, and which would break. As another data point, `ocb-stubblr` is conservative about this and [considers both directories](https://github.com/pqwy/ocb-stubblr/blob/34dcbede6b51327172a0a3d83ebba02843aca249/src/ocb_stubblr.mli#L185).

So, I suggest the following plan:
- add both directories, for consistency with what is done for system wide directories
- encourage installing `pkg-config` configuration files by default in `lib/`
- review the existing packages that install in `share/` and move the arch-dependent `.pc` files to `lib/`, with a changelog entry for dependent code that calls `pkg-config` by hand.

Sounds good?

hannesm:
I digged up earlier discussion:
- https://github.com/ocaml/opam/issues/2153 (opam now supports `lib_root` for `.install` files (is this exposed by any build system))
- original "why mirage-solo5 installs pc file below `share/pkgconfig`" https://github.com/mirage/mirage-solo5/pull/15
- https://github.com/mirage/mirage/pull/669 (mirage to support both `share` and `lib`) - which turned out to be brittle: some linux distribution's (rather old) pkg-config required a specific order of the directories in `PKG_CONFIG_PATH`.

Since opam2 should allow to install below `lib/pkgconfig` easily (since https://github.com/ocaml/opam/commit/905d6258c803b6f306a34d78c7f776357795c0eb), we should investigate which packages are installing atm to `share` and modify them to install below `lib`.

emillon:
Cool, thanks for the pointers! My understanding is that it was not possible to install directly in `lib_root` at the time, but `share_root` was available first.

Now that it's easier to use `lib_root`, we should encourage that while keeping compatibility. We can try to eliminate all uses of `share/pkgconfig`, but I don't think it's that much of a problem if this directory is still scanned.

hannesm:
I went through the list in https://github.com/mirage/mirage/pull/669 and opened a PR for mirage-solo5 (https://github.com/mirage/mirage-solo5/pull/40) - the only package I could find that installs into `share`

mato:
@emillon:

> As for the patch itself, this makes sure that the PKG_CONFIG_PATH environment variable is properly extended in the opam environment, and in particular that pkg-config can automatically find the libraries installed by opam itself. 

Currently, OPAM packages that need pkg-config must manually set `PKG_CONFIG_PATH` during their build/install process. While I agree that this is annoying from the package maintainer's point of view and presumably makes some Dune integration hard(er), be aware that this change does more than the existing setup:

- In the existing setup (manually set in build/install scripts): `PKG_CONFIG_PATH` is modified only _during the build and install process_, and otherwise left alone.
- In the proposed setup (`setenv`): `PKG_CONFIG_PATH` will essentially be modified for _the entirety of any OPAM user's shell session_, assuming they install `conf-pkg-config`.

Are you **sure** that this will not have any negative side effects on system (non-OPAM) packages that use pkg-config? Specifically:

1. Do all pkg-config versions (BSD, GNU/Linux and OSX) behave the same and only use `PKG_CONFIG_PATH` to _extend_ the default search path? Or are there versions out there that _replace_ the search path?
2. What should/will happen if an OPAM package installs a `.pc` file with the same name as a system package? In the proposed setup this may cause such an OPAM package to override the system definition -- this is probably not what the user wants? In the current manual scheme this behaviour would be confined only to other OPAM packages.

@hannesm:

> some linux distribution's (rather old) pkg-config required a specific order of the directories in PKG_CONFIG_PATH

I remember now that this was an issue, but can't find this specific piece of information in the PRs you linked, can you point me to it?

mato:
@emillion:

Another thing to check is the behaviour of existing OPAM packages, both with respect to any manipulation of `PKG_CONFIG_PATH` and to point (2) in my previous comment. Doing a quick recursive grep of opam-repository for `PKG_CONFIG_PATH` reveals more than just Mirage packages that touch this variable. Granted, most of the uses seem to be to just set it to the OPAM prefix, but I've not looked in detail.

hannesm:
> I remember now that this was an issue, but can't find this specific piece of information in the PRs you linked, can you point me to it?

From https://github.com/mirage/mirage/blob/master/lib/mirage.ml#L659-L663
```
(* the order here matters (at least for ancient 0.26, distributed with ubuntu 14.04 versions): use share before lib! *)
```
and https://github.com/mirage/mirage/pull/676 contains further discussion.

hannesm:
I agree with @mato that we should find a suitable solution for MirageOS 4.0 (now that opam2 is deployed). Modifying `PKG_CONFIG_PATH` globally is not a good solution for MirageOS opam packages.

emillon:
@mato thanks for the input!

At first I tried to add this to `dune.configurator`, but it would add an indirect dependency on opam; then I tried add this in opam itself; and finally was pointed to this `setenv` feature, which sounds like it work. For example, packages can install man pages, and opam properly extends `MANPATH` so that the environment works. At some point, it even extended `PERL5LIB`, so I think it's pretty similar.

I can't be sure that this patch won't break anything of course, but it seems reasonable since it's already how other variables are extended, and since `pkg-config` will always fall back to the system directories. It means that it will pick up `libpng.pc` if an opam package installs one, but the same way any opam package that installs a `ls` binary can mess up your system.

That being said, I agree that our use of `pkg-config` to pass some strings around in mirage is not ideal and that it's a great occasion to use opam2 package variables. I'll run some experiments and check if that can replace our use of `pkg-config` in mirage.

I especially look forward to this, since that removes all the lib vs share choices as well as whether or not to work around old toolchain bugs :)

mato:
> I can't be sure that this patch won't break anything of course, but it seems reasonable since it's already how other variables are extended, and since pkg-config will always fall back to the system directories. It means that it will pick up libpng.pc if an opam package installs one, but the same way any opam package that installs a ls binary can mess up your system.

Well, I'm not an opam-repository maintainer, but in general I take a dim view of things that manipulate the user's "global" view of the system in potentially unexpected ways. I'm not familiar with Dune and why the integration you're working on can't use the current set-manually-when-needed scheme, but if this is your only option then at a minimum I would use a `post-message` to inform the user that their effective `PKG_CONFIG_PATH` has been modified.

That said, I think more repository maintainers should comment or vote on this change.

emillon:
>  I'm not familiar with Dune and why the integration you're working on can't use the current set-manually-when-needed scheme

Shelling out to opam in dune (or using `OPAM_SWITCH_PREFIX`, which is set by opam already) is still an option. We thought that it would make sense if opam handled this instead, but we didn't think of these scoping and compatibility problems :)

UnixJunkie:
@emillon you can become the maintainer of this package.
You know pkg-config more than I do.
I don't even remember why I had to create conf-pkg-config.

emillon:
It looks like it's a bit intrusive to do that at the opam level. We're trying to find a solution in mirage itself, see mirage/mirage#969. In the meantime I'm closing this, thanks! (cc @thelortex - feel free to reopen later if in the end it's required but that seems unlikely)

PR #13046, opened 2018-11-24
Add the conf-tensorflow package.

LaurentMazare:
This PR adds a conf-tensorflow package.
This linux only packages triggers the install of the libtensorflow library. As this library is not available for most linux distributions the binaries produced by Google gets downloaded and copied to the opam lib directory.
This will be used for a new version of the tensorflow package (which has not been updated since building requires the presence of the libtensorflow shared library).

### DIFF
--- a/packages/conf-tensorflow/conf-tensorflow.0.1/opam
+++ b/packages/conf-tensorflow/conf-tensorflow.0.1/opam
@@ -1,0 +1,37 @@
+opam-version: "2.0"
+homepage: "https://github.com/LaurentMazare/tensorflow-ocaml"
+maintainer: "lmazare@gmail.com"
+bug-reports: "https://github.com/LaurentMazare/tensorflow-ocaml/issues"
+authors: [
+  "Laurent Mazare"
+]
+install: [
+  [
+    "sh"
+    "-c"
+    """
+test -d %{lib}%/tensorflow/libtensorflow.so ||
+  ( tar xzf libtensorflow.tar.gz &&
+    mkdir -p %{lib}%/tensorflow &&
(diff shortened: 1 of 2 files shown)


LaurentMazare:
@mseri is there anything I can do to help get this PR in ?

mseri:
There has been a long discussion of conf packages not installing anything and having to require additional packages to install their dependencies, see the discussion at https://github.com/ocaml/opam-repository/pull/12018 and the linked PRs. So I am not sure what we should be doing at the moment, and that is why I added the help-wanted flag. Ping @kit-ty-kate and @avsm

LaurentMazare:
@mseri would it be better for this package to just check whether the library is already installed and if that's not the case fail the installation and print the install command for the user to run ?
And if this is the way to go, would we have a way to bypass CI for this (or to run the install command) ?

avsm:
I think this package is fine to have for local opam installations, but it should indeed just lose the `conf-` prefix and focus on installing a local version of tensorflow that is compatible with the opam OCaml bindings.

There is some precedent for this with the mirage C library packages that do not actually install any OCaml :-) But these packages all support some OCaml binding that is also present in opam-repository, hence justifying their existence here.

Would it also be possible to get macOS support for this @LaurentMazare?  And the binary version is x86 only if I'm not mistaken, so the constraint should take that into account too.

LaurentMazare:
Thanks @avsm I just pushed a change to rename the package dropping the conf prefix and also add some specific macos bits, I don't have access to a mac to test them though.
One slightly sad thing is that the package will now download both the linux and darwin versions as I haven't found a way to have a filter on extra-source. Let me know if you know a way to do this.

mseri:
Thanks for the help. I think this is fine to merge now

PR #13053, opened 2018-11-26
Replace jbuilder dep with dune

tuncer:


tuncer:
v2 of #12752. @avsm, as per your suggestion in #12752: let's try to pinpoint the opam files that should be left as is, based on CI results, and then revert those files. I wouldn't mind help with making sense of CI output.

tuncer:
v2 of #12752. @avsm, as per your suggestion in #12752: let's try to pinpoint opam files that should be left as is, based on CI results, and then revert those. I wouldn't mind help with making sense of CI output and extracting the list of files to revert.

tuncer:
v2 of #12752. @avsm, as per your suggestion in #12752, I'll try to revert files that led to _'these releases are not installable anymore (78)'_. I wouldn't mind help figuring out the root causes (relevant opam files), if you have a script to extract the data from camelus results.

avsm:
Thanks! I think the easiest way to get the results might be to cut and paste from a browser and do `cat foo | xargs -n 1 echo` or similar :-)

tuncer:
I can see travis-ci logs like https://api.travis-ci.org/v3/job/459748898/log.txt but no OCaml Labs CI anymore. Also, you have to load the raw log, or you'll only see a trimmed travis log. So, if you load that file and search for "build switch:", you'll land at a location where actual builds start. Before that there's setup and a very long section for `git diff`. Still, I failed to find the error for, say, `datakit.0.11.0` or `jupyter.2.2.0`.

tuncer:
Reverted changes in base dir of each failed package (e.g. `packages/re2/*`).

avsm:
I've resolved conflicts against the master branch

tuncer:
Thanks, @avsm. I'll check the fresh logs and see if I can find the other culprits. 534e742 seems insufficient.

kit-ty-kate:
> @tuncer we could also just break this up into smaller PRs to feed it through gradually, if you think that's easier for you

I agree. I'm going to close this one. Feel free to request to open it again if you don't agree

PR #13056, opened 2018-11-27
Javalib 2.3.6

anscoil:
* solves installation problem due to sandboxing with opam 2.0


anscoil:
As usual the old versions of `sawja` fail to install `javalib` because of sandboxing.
It should be the last time I need to modify these old packages.
`javalib 2.3.6` and `sawja 1.5.5` are updates of `javalib 2.3.5` and `sawja 1.5.4` with no API or functionality changes, in order to be compatible with `opam 2.0`.

mseri:
Tested locally, seems to work fine. Thanks

PR #13063, opened 2018-11-29
[new release] ahrocksdb (0.2.0)

Engil:
A binding to RocksDB

- Project page: <a href="https://github.com/ahrefs/ocaml-ahrocksdb">https://github.com/ahrefs/ocaml-ahrocksdb</a>
- Documentation: <a href="https://ahrefs.github.io/ocaml-ahrocksdb/">https://ahrefs.github.io/ocaml-ahrocksdb/</a>

##### CHANGES:

Initial release.

### DIFF
--- a/packages/ahrocksdb/ahrocksdb.0.2.0/opam
+++ b/packages/ahrocksdb/ahrocksdb.0.2.0/opam
@@ -1,0 +1,46 @@
+opam-version: "2.0"
+synopsis: "A binding to RocksDB"
+description: """
+This is a binding to Facebook's RocksDB.
+
+Early prototype of this library based on [orocksdb](https://github.com/domsj/orocksdb), we decided to rewrite our own binding to make use of Ctypes's stubs generators instead of the dynamic mode used in orocksdb.
+
+It is currently based and was tested against RocksDB 5.14fb, and should work with newer versions of this library.
+
+## API changes and contributions
+
+While we do not plan big changes in what is already implemented, we do not guarantee the stability of these APIs.
+
+Some APIs could definitely use improvements (moving the current configuration system to a builder-like pattern),
+and some breakage may or may-not happen.
+
+Pull requests to improve any parts of the library are however welcome, whether they are related to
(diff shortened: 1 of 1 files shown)


packages/ahrocksdb/ahrocksdb.0.2.0/opam
```diff
@@ -0,0 +1,47 @@
...
+  "bos" {with-test}
+  "cryptokit" {with-test}
+  "alcotest" {with-test}
+]
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest"] {with-test}
+]
+depexts: ["librocksdb-dev"] {os-distribution = "debian"}
```
mseri: It would be nice to extract a `conf-rocksdb` package for the depexts in other distributions as well in the future

mseri:
It's a pity that we cannot really test this as the necessary depexts are missing from the distribution used by travis

Engil:
I can try to adjust it perhaps ? Albeit it seems it's using Trusty, there's no simple way to get the latest version there.
As of Debian stable we cannot test because the required RocksDB release (>= 5.14.fb) is not available until unstable, the recommended would be a custom install from sources (we check the "default" path used by RocksDB makefile)

packages/ahrocksdb/ahrocksdb.0.2.0/opam
kit-ty-kate: there is already such package. This line is redundant.

mseri:
Don't worry, it's fine. I'll check if we get the other CI back anytime soon and if not I'll merge.

mseri:
@Engil can we depend on `conf-rocksdb` and remove the depext line though?

kit-ty-kate:
cc @avsm for ci.ocamllabs.io

Engil:
> @Engil can we depend on `conf-rocksdb` and remove the depext line though?
Right, we depend already on conf-rocksdb. leaving depexts was an oversight, pushed a new version.

mseri:
Thanks!

PR #13064, opened 2018-11-29
Package opam-bundle.0.4

AltGr:
### `opam-bundle.0.4`
A tool that creates stand-alone source bundles from opam packages
opam-bundle is a command-line tool that, given a selection of packages,
generates a .tar.gz (and optionally a self-extracting) archive containing their
sources, and everything needed to bootstrap and compile them:
- the sources of their dependencies
- the sources of the chosen version of OCaml
- the sources of opam
- a set of scripts to bootstrap, check and install external dependencies,
  compile all the above, install the packages within a sandbox, and optionally
  put wrapper scripts within your PATH

This is expected to be done as normal user, with constrained calls to `sudo`
when needed for depexts and wrappers installation.

---
* Homepage: https://github.com/AltGr/opam-bundle
* Source repo: git+https://github.com/AltGr/opam-bundle
* Bug tracker: https://github.com/AltGr/opam-bundle/issues

---

### DIFF
--- a/packages/opam-bundle/opam-bundle.0.4/opam
+++ b/packages/opam-bundle/opam-bundle.0.4/opam
@@ -5,14 +5,14 @@ license: "GPL-3"
 tags: "org:ocamlpro"
 homepage: "https://github.com/AltGr/opam-bundle"
 bug-reports: "https://github.com/AltGr/opam-bundle/issues"
-dev-repo: "git+https://github.com/AltGr/opam-bundle.git"
 depends: [
-  "ocaml"
-  "ocamlfind"
-  "cmdliner" {>= "1.0.0"}
-  "opam-client" {>= "2.0.0"}
+  "ocaml" {build}
+  "ocamlfind" {build}
+  "cmdliner" {build & >= "1.0.0"}
+  "opam-client" {build & >= "2.0.0"}
 ]
 build: make
+dev-repo: "git+https://github.com/AltGr/opam-bundle"
 flags: plugin
 synopsis: "A tool that creates stand-alone source bundles from opam packages"
 description: """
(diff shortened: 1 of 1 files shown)


packages/opam-bundle/opam-bundle.0.4/opam
```diff
@@ -0,0 +1,38 @@
+opam-version: "2.0"
+maintainer: "Louis Gesbert <louis.gesbert@ocamlpro.com>"
+authors: "Louis Gesbert <louis.gesbert@ocamlpro.com>"
+license: "GPL-3"
+tags: "org:ocamlpro"
+homepage: "https://github.com/AltGr/opam-bundle"
+bug-reports: "https://github.com/AltGr/opam-bundle/issues"
+depends: [
+  "ocaml"
+  "ocamlfind"
```
mseri: Is ocamlfind just a biuld dependency (as ocaml)?

AltGr: Yes, technically the package just installs  a binary that doesn't manage OCaml sources or artifacts, so everything should be a build dependency.

AltGr: (but then there is always the old dilemma of "should it still get recompiled if e.g. there is a fix in `opam-client` for a bug that affects it"... I guess in that case we could file a new release to force recompilation)

mseri: I think that we should keep `cmdliner` and `opam-client` out, as you actively use the libraries (and indeed want the package to be rebuilt in case of bugfixes), while ocaml and ocamlfind are only invoked externally during the build

mseri: I've made the change. I'll let the tests re-run before merging

PR #13076, opened 2018-12-03
CharInfo_width 1.0.0

kandu:
* remove the width_utext function
* add the String functor to handle all kinds of unicode strings

dune runtest will fail with OCaml 4.02 and 4.03 due to ppx issue.


kit-ty-kate:
Thanks!

PR #13079, opened 2018-12-04
[new release] dune (1.6.1)

rgrinberg:
CHANGES:

- Fix hash collision for on-demand ppx rewriters once and for all
  (ocaml/dune#1602, fixes ocaml/dune#1524, @diml)

- Add `dune external-lib-deps --sexp --unstable-by-dir` so that the output can
  be easily processed by a machine (#1599, @diml)


kit-ty-kate:
Thanks!

PR #13080, opened 2018-12-04
Frama-C 18.0 (Argon) for opam 1.2

dbuhle:


maroneze:
The build errors seem to be related to depext (on macOS) and due to external issues in the case of Linux distributions. The package itself seems fine, I was able to install and run it locally on a `ocaml/opam` docker machine.

Is there something that we can do to make the builds pass?

kit-ty-kate:
Is it a critical release? Only critical fixes are supposed to be backported to opam 1.2

dbuhle:
Oh, that is unfortunate. No, this is not a critical fix.

However, many of our users install Frama-C through the opam provided by their linux distribution, which is still opam 1.2 on Ubuntu and Debian. Even if this release is not critical, we would like to distribute Frama-C 18 to our users without requiring them to switch to opam 2.0.   
Some users also reported issues with bubblewrap (e.g. on Docker).

Overall, we expected this to be the final opam 1.2 release of Frama-C, to help with the transition, since the last release happened before the default repository switch.

kit-ty-kate:
Ok fair enough. For the record, with docker bubblewrap can be disabled (e.g. https://github.com/kit-ty-kate/opam-check-all/blob/9e27d9e74e6cb97ef5118e3210728b6509f5d61b/server/backend/check.ml#L88-L90)

Thanks!

PR #13082, opened 2018-12-04
charInfo_width 1.1.0

kandu:
* loosen the restriction of parameter module of the `CharInfo_width.String` functor (backward-compatible)

### DIFF
--- a/packages/charInfo_width/charInfo_width.1.1.0/opam
+++ b/packages/charInfo_width/charInfo_width.1.1.0/opam
@@ -7,7 +7,7 @@ license: "MIT"
 dev-repo: "hg://https://bitbucket.org/zandoye/charinfo_width"
 build: [
   ["dune" "build" "-p" name "-j" jobs]
-  ["dune" "runtest" "-p" name "-j" jobs] {with-test & (switch > "4.03.0") & (switch < "999.0~")}
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test & (ocaml:version >= "4.04.0")}
 ]
 depends: [
   "ocaml" {>= "4.02.3"}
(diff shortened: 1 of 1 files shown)


packages/charInfo_width/charInfo_width.1.1.0/opam
```diff
@@ -0,0 +1,27 @@
+opam-version: "2.0"
+maintainer: "zandoye@gmail.com"
+authors: [ "ZAN DoYe" ]
+homepage: "https://bitbucket.org/zandoye/charinfo_width/"
+bug-reports: "https://bitbucket.org/zandoye/charinfo_width/issues"
+license: "MIT"
+dev-repo: "hg://https://bitbucket.org/zandoye/charinfo_width"
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test & (switch > "4.03.0") & (switch < "999.0~")}
```
kit-ty-kate: What is the goal of `switch < "999.0~"` exactly?
Otherwise, for some reason the tests don't seem to be run anyway, even on 4.04.2 for example. The `switch` variable doesn't seem to be what you want (https://opam.ocaml.org/doc/Manual.html#opamvar-switch)

cc @rjbou would know maybe?

rjbou: `switch` variable is just the switch name, it can be whatever user want...
@kandu, do you want to set a restriction on the ocaml version of the current switch? If so, filter `with-test & (ocaml:version > "4.03.3")` is maybe what  you need (cf. [packages variables](https://opam.ocaml.org/doc/Manual.html#Package-variables))

kandu: Ah, thank you two! That's it!

mseri:
Thanks

PR #13083, opened 2018-12-04
Remove mesh 0.7.4, mindstorm 0.5.3 & 0.5.4, etc.

Chris00:
According to http://check.ocamllabs.io, this version does not compile
on OCaml ≥ 4.02.3 and newer versions exist that work on all OCaml
version ≥ 4.02.3.


kit-ty-kate:
`mindstorm` failures seem to be from a missing dependency to camlp4. `oasis2opam` failures I bet come from a missing constraint on oasis. `odepack` failure seems to come from a missing external dependency to fortran.

Removing those packages are not advisable.

Chris00:
On  6 December 2018 at 15:28 GMT, Kate wrote:
>
> Ok so it seems I could fix everything except mesh.0.7.4 which I think was broken from the start (I tried with every compiler versions and versions of ocamlbuild possible)
>
> Are you ok with my fixes? (especially concerning conf-gfortran) If so I'll go ahead and merge it

Thanks for your work Kate.  I'll look at your fixes ASAP.

mseri:
Ping @Chris00, if you are happy with it we could merge

packages/conf-gfortran/conf-gfortran.0/opam
```diff
@@ -0,0 +1,20 @@
+opam-version: "2.0"
+synopsis: "Virtual package relying on a gfortran system installation"
+maintainer: "https://github.com/ocaml/opam-repository/issues"
+authors: "The GCC team"
+license: "GPL 3"
+homepage: "https://gcc.gnu.org/fortran/"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+flags: conf
+build: ["gfortran" "--version"]
```
Chris00: Are you sure all external dependencies install `gfortran`?  Odepack `discover` script also looks for `g95` and `g77`.

Chris00:
Besides the above remark, I am fine with the changes.

packages/conf-gfortran/conf-gfortran.0/opam
kit-ty-kate: I looked at the packages for freebsd and openbsd and they seem wrong (it installs something unrelated). I'll fix this but the rest seems to install `gfortran` "at least"

kit-ty-kate:
Thanks!

Chris00:
@kit-ty-kate Thanks for your work!

PR #13084, opened 2018-12-04
add conf-cargo

c-cube:


c-cube:
I don't even know where this error comes from, I imitated existing conf packages.

mseri:
The flag error is expected, will be fixed in the next version of opam afaik

c-cube:
I'm worried that centos is too old to have a cargo package in its main repositories…

mseri:
I am fairly confident it is :D

PR #13086, opened 2018-12-05
Packages: bindings for opencc0 & opencc1

kandu:
The new releases of OpenCC 1.x are API incompatible with the previous 0.x releases.
This PR contains two bindings for OpenCC0 and OpenCC1, separately.
After this PR is merged, I'll turn the `opencc` package in opam to a virtual transition package.

### DIFF
--- a/packages/conf-opencc0/conf-opencc0.1/files/test.c
+++ b/packages/conf-opencc0/conf-opencc0.1/files/test.c
@@ -1,0 +1,3 @@
+int main(void) {
+  return 0;
+}
--- a/packages/conf-opencc0/conf-opencc0.1/opam
+++ b/packages/conf-opencc0/conf-opencc0.1/opam
@@ -1,0 +1,30 @@
+opam-version: "2.0"
+maintainer: "zandoye@gmail.com"
+authors: [ "ZAN DoYe" ]
+homepage: "https://bitbucket.org/zandoye/ocaml-opencc0/"
+bug-reports: "https://bitbucket.org/zandoye/ocaml-opencc0/issues"
+license: "MIT"
+tags: [
+  "opencc"
+  "Chinese conversion"
+]
+
+build: [
+  ["sh" "-exec" "cc $CFLAGS test.c -l:libopencc.so.1"]
+]
+
+depexts: [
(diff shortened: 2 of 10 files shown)


mseri:
Thanks. Would it make sense to create a `conf-opencc` package?

kandu:
It makes sense to create a conf-opencc1 for opencc1, though it's not necessary for now, IMHO.

As for opencc0, almost all up-to-date Linux distributions dropped it in their repository.
opencc0 has a significant benefit that opencc1 lacks: selecting of conversion modes: `fast` `segment` `candidates`. In contrast, opencc1 has only one `fast` mode, which is not quite qualified for serious converting process, so we build user-managed opencc0.
In this situation, to create a conf-opencc0 makes no sense.

I think a post-message after installation is well enough. any suggestion?

kandu:
well, at least openSUSE still provides libopencc0 and a conf-xxx package has its uses: to make sure the relying environment is setup correctly. I've made these conf-opencc(0|1) packages. If it's highly recommended under the opam policy, I'll commit these conf-xxx packages. suggestion?

kandu:
opam lint doesn't pass if the flags is set to `conf`.
The error msg is `error 52: Package is needlessly flagged "light-uninstall"`

How to correct it?

mseri:
Thanks for looking into it. Please add them to this PR, don't worry about the failure, it will be fixed in the next release of opam. The `conf` flag is the correct one

kandu:
Can we delete the old opencc package directly or keep it as a transition package for a while?

mseri:
I'd keep it as a transition package, if you want add a ponstinstall message referring to the new packages, to make things clearer for the users

kandu:
ok, tested from 4.02.3 to 4.06.1 by hand, both installation and practical converting(mainly for user managed libopencc0) in debian.
And I extracted and checked libopencc1(v0) libopencc2(v1) packages from fedora, opensuse, gentoo, freebsd, the suffix of .so files are the same(so.1 for libopencc0, so.2 for libopencc1), so the two package will work properly on these OS with no problem, at least, I hope so  ;)

mseri:
Thank you very much!

PR #13088, opened 2018-12-05
Fix patch files of old versions of coq and coqide

eponier:
The coqide files seem to have been ported to opam 2 without updating accordingly the .patch files. Indeed, opam 2 uses `patch -p1` instead of `patch -p0`, thus some paths in the .patch files needed to be prefixed.

### DIFF
--- a/packages/coqide/coqide.8.4pl2/opam
+++ b/packages/coqide/coqide.8.4pl2/opam
@@ -34,10 +34,10 @@ synopsis: "IDE of the coq formal proof management system"
 extra-files: [
   [
     "MAKEFILE_remove_useless_for_coqide.patch"
-    "md5=4726acb2340f2d808344f8c2a7c3b685"
+    "md5=0ef819341127690e51ba1ee0f73df43e"
   ]
-  ["CONFIGURE_allow_make4.patch" "md5=de289bd72e77531351acf47e3575ddc9"]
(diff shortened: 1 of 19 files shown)


packages/coqide/coqide.8.4pl2/opam
```diff
@@ -34,7 +34,7 @@ synopsis: "IDE of the coq formal proof management system"
 extra-files: [
   [
     "MAKEFILE_remove_useless_for_coqide.patch"
-    "md5=4726acb2340f2d808344f8c2a7c3b685"
+    "md5=0ef819341127690e51ba1ee0f73df43e"
   ]
   ["CONFIGURE_allow_make4.patch" "md5=de289bd72e77531351acf47e3575ddc9"]
```
mseri: Should this (and the next one) also have an updated md5?

eponier: Indeed, my sed commands were not powerful enough. According to Travis, some versions of coq (not coqide) also have the same problems. Will have a look.

eponier: I have just pushed a version with correct md5 for patch files. I have also added fixes for some versions of coq.

eponier:
Apparently there are some more issues with old versions of coq. This PR does not solve all of them, but still improves the situation overall.

mseri:
This seems to be fine. Thanks for the patience

PR #13093, opened 2018-12-05
Destroy fix.20181130, as it is broken.

fpottier:


kit-ty-kate:
Indeed it doesn't install anything. Thanks

PR #13096, opened 2018-12-06
Lablgtk3.0.beta1

garrigue:
An attempt at packaging the freshly released lablgtk3.
Note that I get syntax errors on the depexts, whereas I have just copied
the files from lablgtk and conf-gtksourceview,
Also, I did only test on macOS.


packages/lablgtk3/lablgtk3.0.beta1/opam
```diff
@@ -0,0 +1,50 @@
...
+  ["./configure" "--prefix" prefix]
+  [make "world"]
+]
+install: [
+  [make "install"]
+]
+remove: [["ocamlfind" "remove" "lablgtk3"]]
+depends: [
+  "ocaml" {>= "4.05"}
+  "ocamlfind" {>= "1.2.1"}
```
mseri: Are you also linking against `ocamlfind` or it is only a build dependency?

packages/conf-gtksourceview3/conf-gtksourceview3.0/opam
```diff
@@ -0,0 +1,24 @@
...
+  ["gtksourceview3"] {os = "freebsd"}
+  ["gtksourceview3"] {os = "openbsd"}
+  ["gtksourceview3-devel"] {os-distribution = "opensuse"}
+  ["libgtksourceview-3.0-dev"] {os-distribution = "ubuntu"}
+  ["gtksourceview3" "libxml2"] {os = "macos" & os-distribution = "homebrew"}
+]
+synopsis: "Virtual package relying on a GtkSourceView-3 system installation"
+description:
+  "This package can only install if libgtksourceview-3.0-dev is installed on the system."
+flags: conf
```
mseri: Don't worry about linting failures with this flag, they are known and will be fixed in the next opam release. This file looks correct

packages/lablgtk3/lablgtk3.0.beta1/opam
```diff
@@ -0,0 +1,50 @@
...
+]
+depopts: [
+  "conf-gtksourceview3"
+]
+depexts: [
```
mseri: Why we didn't put this on a `conf-gtk3`?

garrigue: Because I was basing myself on the lablgtk2 opal file, written by somebody else.
I suppose it would indeed make more sense to have a separate conf-gtk3 package.
Should I do that?
By the way, the depexts are quite arbitrary.

kit-ty-kate: Having a separate conf- file is always better as it allows to have to improve/change stuff only once and not in every opam files (also it might also be handy for other packages in the future)

garrigue:
OK, I created a conf-gtk3 package.
Should be ready for merging now. Note that the version of Gtk3 in ubuntu-16.04 is too old, and the feature seems to be required in CoqIDE.

mseri:
Looks good, thanks. Can I tag `ocamlfind` as `{build}` dependency? I could not see any direct linking to it in the code, it seems just used to for installation and removal

garrigue:
Yes, findlib just needed during installation (not even during build).

mseri:
@kit-ty-kate this is good for merge imho. Do we need to do something about #13098 ?

kit-ty-kate:
Yeah good enough. #13098 can be closed

PR #13098, opened 2018-12-06
Add conf-gtksourceview3

AltGr:
The depexts probably need fixing, I only adjusted them for the Debian family. CI should be able to help with that.


kit-ty-kate:
Mmh this PR overlaps with #13096

AltGr:
ah, indeed, @garrigue beat me to it ^^

PR #13102, opened 2018-12-07
Package parany.5.0.0

UnixJunkie:
### `parany.5.0.0`
Parallelize any computation
Generalized map reduce for parallel computers (not distributed computing).
Can process in parallel an infinite stream of elements.

Can process a very large file in parallel on a multicore computer;
provided there is a way to cut your file into independent blocks
(the 'demux' function).
The processing function is called 'work'.
The function gathering the results is called 'mux'.
The number of processors running your computation in parallel is called
'nprocs'.
The chunk size (number of items) processed by one call to the 'work' function
is called 'csize'.

Read the corresponding ocamldoc before using.

---
* Homepage: https://github.com/UnixJunkie/parany
* Source repo: git+https://github.com/UnixJunkie/parany.git
* Bug tracker: https://github.com/UnixJunkie/parany/issues

---

### DIFF
--- a/packages/parany/parany.5.0.0/opam
+++ b/packages/parany/parany.5.0.0/opam
@@ -1,0 +1,41 @@
+opam-version: "2.0"
+maintainer: "unixjunkie@sdf.org"
+authors: "Francois Berenger"
+license: "LGPL"
+homepage: "https://github.com/UnixJunkie/parany"
+bug-reports: "https://github.com/UnixJunkie/parany/issues"
+dev-repo: "git+https://github.com/UnixJunkie/parany.git"
+depends: [
+  "ocaml" {>= "4.03.0"}
+  "dune" {build}
+  "base-unix"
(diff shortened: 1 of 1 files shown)


UnixJunkie:
cool this opam-publish thingy

UnixJunkie:
Thanks go to to @mseri for dunification of the library.

UnixJunkie:
all CI tests are green

kit-ty-kate:
Sounds good enough. `"-j" jobs` might be worth adding to the `dune` command next time though. Thanks!

UnixJunkie:
@kit-ty-kate Do you want that I make a feature request to opam-publish about that?

kit-ty-kate:
opam-publish? Why would it be a feature in opam-publish? From what I understand opam-pushlish is not responsible of the opam file. It would be something to change upstream in your source code instead.

PR #13016, opened 2018-11-20
Add upper bounds to reverse dependencies of Jane Street packages.

xclerc:
(Some PRs were merged wille I was working on this.)

PR #13010, opened 2018-11-19
Add upper bounds to reverse dependencies of Jane Street packages

xclerc:

PR #13077, opened 2018-12-04
Package obeam.0.1.1

amutake:
### `obeam.0.1.1`
A utility library for parsing BEAM format

---
* Homepage: https://github.com/yutopp/obeam
* Source repo: git+https://github.com/yutopp/obeam.git
* Bug tracker: https://github.com/yutopp/obeam/issues

---
:camel: Pull-request generated by opam-publish v2.0.0

PR #13107, opened 2018-12-07
Add upper bounds to reverse dependencies of Jane Street packages.

xclerc:
Follow-up to https://github.com/ocaml/opam-repository/pull/13010 and https://github.com/ocaml/opam-repository/pull/13016.


xclerc:
For the record, the OpenSUSE error has already been discussed
[here](https://github.com/ocaml/opam-repository/pull/13077), and it is expected to be fixed on the [opam-depext side](https://github.com/ocaml/opam-depext/issues/109).

kit-ty-kate:
Thanks!

PR #13112, opened 2018-12-09
+angstrom.0.11.0

seliopou:


mseri:
The revdeps failures are related to issues in old versions of `angstrom-async`and `angstrom-unix` that were transitively depending on `result` from `angstrom` (see https://github.com/inhabitedtype/angstrom/commit/ce7f2baf951c37f797f39e848725ad258db26071 for the change that broke them). Can you please add an upper bound on angstrom to those packges?

mseri:
I am going to fix the revdeps in a separate PR with some other packages

seliopou:
Thank you!

PR #13117, opened 2018-12-09
Add a libtorch package.

LaurentMazare:
This is similar to #13046 but for libtorch. This package just downloads and extracts pre-built binaries for libtorch 1.0.0. This will be used as a dependency by the torch package based on [ocaml-torch](https://github.com/LaurentMazare/ocaml-torch).

One current limitation is that both the linux and macos binaries are downloaded as I'm not sure how to apply a filter to extra-source.


mseri:
Thanks

PR #13121, opened 2018-12-10
conf-r trial fix

UnixJunkie:
Don't merge this.
We might have to upgrade the bugfix version number of this package if the fix is correct.


kit-ty-kate:
It's fixed after #13126. Thanks!

kit-ty-kate:
oops I didn't see the `don't merge this` in the original message. Would you like me to revert it?

kit-ty-kate:
Reverted. What was the original goal of this PR?

UnixJunkie:
I will bump the version number of the incriminated package and include your fix.
Thanks!

UnixJunkie:
original goal: apparently, the R command was not installed by the package.
While this package is supposed to install the R interpreter (command: R).

kit-ty-kate:
@UnixJunkie it was, it's just that the opensuse image changed an broke the way the system was detected so no packages were actually installed. `conf-r` works as it is now.

PR #13122, opened 2018-12-10
stdcompat.8

thierry-martinez:
- Fix: module Fun was not exported

- Fix: value Stdlib.protect was not exported

- Fix: equality between aliases of Stdlib.result and of Seq.t

- Fix: Stdlib.result redefined before 4.03 (instead of 4.06)


thierry-martinez:
The revdep failure on electrod looks like a bug unrelated to stdcompat: as far as I understand, the interface of `Containers.List.Assoc.update` from the containers package has changed and does no longer take a ~f labeled argument.

mseri:
Thanks

mseri:
I have added the bound on electrod in a separate PR

PR #13126, opened 2018-12-10
fix opensuse metadata for packages

avsm:
OpenSUSE leap self-identifies as "opensuse-leap" now, so this
patch uses `os-family` instead so that any suse distribution
will return the depext. This should maintain compatability with
opensuse 42.3 while working with opensuse-leap 15.0 as well.

### DIFF
--- a/packages/alsa/alsa.0.2.3/opam
+++ b/packages/alsa/alsa.0.2.3/opam
@@ -14,7 +14,7 @@ depends: ["ocaml" "ocamlfind" "conf-pkg-config" {build}]
 depexts: [
   ["alsa-lib-devel"] {os-distribution = "centos"}
   ["alsa-lib-devel"] {os-distribution = "fedora"}
-  ["alsa-lib-devel"] {os-distribution = "opensuse"}
+  ["alsa-lib-devel"] {os-family = "suse"}
   ["libasound2-dev"] {os-distribution = "debian"}
   ["libasound2-dev"] {os-distribution = "ubuntu"}
 ]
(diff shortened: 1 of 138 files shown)


mseri:
I have merged a few packages before seeing this PR, can you double check if there are more instances of opensuse that need translation?

kit-ty-kate:
I'm not sure about this change. Can't we fix opam instead?

kit-ty-kate:
I looked at how `os-distribution` and `os-family` are detected and I at least think we need to make sure opensuse **always** had `suse` as first item of `ID_LIKE` in `/etc/os-release`. Otherwise it won't work for other versions

kit-ty-kate:
Found in their documentation:
```
ID_LIKE= A space separated list of IDs for related operating systems with common behaviour to ID=. eg "opensuse suse". This is so scripts don't need to micro-manage every possible option of ID=. The entry of "suse" represents all openSUSE, SUSE, SUSE Linux Enterprise distributions and derivatives. "opensuse" repesents only openSUSE distributions and derivatives. This field should be safe for parsing in scripts.
```
It doesn't say that `suse` will be the first one but from the versions I could test from hub.docker.com, they all have it except 13.1 (and previous I guess) which have their own non-standard file (so unsupported and will fall back to `linux`)

So I think this change is correct

kit-ty-kate:
@mseri I checked manually and none of the PR merged recently had depexts. We're good to go.

Thanks!

UnixJunkie:
Shouldn't you bump the minor version number of all those packages?

PR #13128, opened 2018-12-11
Fix some broken revdeps adding some upper bounds

mseri:
This should fix multiple revdeps that were currently broken by package updates


mseri:
The two revdeps failure are due to some dnf issues that are independent

kit-ty-kate:
Thanks!

PR #13132, opened 2018-12-12
added packages/conf-r/conf-r.1.0.1

UnixJunkie:
fix for openSUSE: install R-base instead of R-core

### DIFF
--- a/packages/conf-r/conf-r.1.0.1/files/check.r
+++ b/packages/conf-r/conf-r.1.0.1/files/check.r
@@ -1,0 +1,1 @@
+quit(status=0)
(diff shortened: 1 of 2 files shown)


UnixJunkie:
@kit-ty-kate will know if this needs to be merged.

kit-ty-kate:
@UnixJunkie as said in https://github.com/ocaml/opam-repository/pull/13121 I think it was fine before (the R compiler is already installed by `R-core`). So unless you find some problem on opensuse I would say it was fine.
Also I'm not at all convinced this needs a new package (especially if it lets the old one around). IMO for conf- packages there should be only one package in the general case or several if different version constraints are needed. I guess you can bump a revision version or something like that but it seems to me that the previous one should be removed/moved instead.

PR #13135, opened 2018-12-12
Add support for Nix openssl package on macOS

anmonteiro:
Nix can also be used as a package manager in macOS

### DIFF
--- a/packages/conf-openssl/conf-openssl.1/files/osx-build.sh
+++ b/packages/conf-openssl/conf-openssl.1/files/osx-build.sh
@@ -2,7 +2,7 @@
 
 if [ -d "$HOME/.nix-profile/lib/pkgconfig/" ]; then
   # Nix on macOS
-  PKG_CONFIG_PATH=$HOME/.nix-profile/lib/pkgconfig openssl
+  PKG_CONFIG_PATH=$HOME/.nix-profile/lib/pkgconfig pkg-config openssl
 elif [ -d "/usr/local/opt/openssl/" ]; then
   # Homebrew
   PKG_CONFIG_PATH=/usr/local/opt/openssl/lib/pkgconfig pkg-config openssl
--- a/packages/conf-openssl/conf-openssl.1/files/osx-build.sh
+++ b/packages/conf-openssl/conf-openssl.1/files/osx-build.sh
@@ -1,6 +1,9 @@
 #!/bin/sh
 
-if [ -d "/usr/local/opt/openssl/" ]; then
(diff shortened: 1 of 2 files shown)


mseri:
This requires also updating the hashes in the opam file

anmonteiro:
Ah wasn't aware. Done, thanks

packages/conf-openssl/conf-openssl.1/files/osx-build.sh
```diff
@@ -1,6 +1,9 @@
 #!/bin/sh
 
-if [ -d "/usr/local/opt/openssl/" ]; then
+if [ -d "$HOME/.nix-profile/lib/pkgconfig/" ]; then
+  # Nix on macOS
+  PKG_CONFIG_PATH=$HOME/.nix-profile/lib/pkgconfig
```
mseri: Sorry, I just noticed this: I think you need to add ` pkg-config openssl` to this line

anmonteiro: oh wow my bad, nice catch. Will do right now

anmonteiro: Done and updated the hash too

packages/conf-openssl/conf-openssl.1/files/osx-build.sh
```diff
@@ -1,6 +1,9 @@
 #!/bin/sh
 
-if [ -d "/usr/local/opt/openssl/" ]; then
+if [ -d "$HOME/.nix-profile/lib/pkgconfig/" ]; then
+  # Nix on macOS
+  PKG_CONFIG_PATH=$HOME/.nix-profile/lib/pkgconfig openssl
```
mseri: Still not it, I think the correct line should be
```suggestion
  PKG_CONFIG_PATH=$HOME/.nix-profile/lib/pkgconfig pkg-config openssl
```

mseri:
Thanks for the patience

pelzlpj:
👎 

This is actually a regression on my nix config. osx-build.sh seems to assume that `openssl` is installed via "nix-env -i". This breaks development environment isolation; the recommended path is to pull in libraries like `openssl` via a custom `nix-shell` invocation.

mseri:
@pelzlpj would you mind opening a PR to revert this change and add a postinstall message in case of failure that explains what to do if the user uses nix?

anmonteiro:
@pelzlpj Apologies, that is indeed the case. I'm quite new to nix and clearly didn't think there would be other workflows.

PR #13136, opened 2018-12-12
Module Explicits uses 4.02.1 as base

Leonidas-from-XIV:
Currently it fails with

```
### output ###
# OCaml version mismatch: 4.02.1, expected 4.02.0
```

As far as I see this is due to modular implicits using 4.02.1 as a base so I'm basically renaming the variant.


mseri:
The CI build is green but in reality it fails with
```
No solution for ocaml-variants.4.02.1+modular-implicits
```

I don't really know how this kind of packages are tested/reviewed

Leonidas-from-XIV:
@mseri Yes, I also noticed that but the previous version was not installable either (though the checksum didn't change, so I assume there is some opam2 magic that is testing that the compiler versions match). opam-repository checks are very confusing to me. Any input greatly appreciated.

PR #13146, opened 2018-12-14
ppx_deriving_yojson.3.2

gasche:


gasche:
@kit-ty-kate thanks for the fix!

The only remaining failure in the CI is on MacOS ([logs](https://travis-ci.org/ocaml/opam-repository/jobs/467996128)), and it seems to be unrelated to our project: it looks like the installation of `opam` failed. Is there a known configuration issue for the OSX CI?

anmonteiro:
It also looks like revdeps is failing and some packages will need upper bounds on `ppx_deriving_yojson`.

3.2 introduces a new (generated) function, and people using the attribute in interface files are getting those errors.

gasche:
Hm... Maybe we should try to fix https://github.com/ocaml-ppx/ppx_deriving_yojson/issues/76 to improve backward-compatibility, and scratch the current release... @anmonteiro, what do you think?

anmonteiro:
I'm OK with improving backwards compatibility

gasche:
I'll close this for now, then.

PR #13152, opened 2018-12-16
lambdasoup.0.6.3: incorrect checksum

avsm:
cc @aantron to confirm that the upstream checksum did indeed change
validly

Signed-off-by: Anil Madhavapeddy <anil@recoil.org>


avsm:
I can't find an archive of the old checksum to look for differences, unfortunately...

aantron:
The checksum looks correct w.r.t. the archive on GitHub (of course). I downloaded that archive, looked at the build system, and diffed the code against what I have locally for the tag. Everything looks correct, so I think we can merge this.

PR #13163, opened 2018-12-17
ppx_deriving_yojson incompatible with dune < 1.2

Leonidas-from-XIV:
ppx_deriving_yojson 3.3 includes a dune-project file which specifies the 1.2 language format, so the opam file has to depend on a newer dune.

A pull request to resolve this issue exists upstream: https://github.com/ocaml-ppx/ppx_deriving_yojson/pull/88 where compatibility with older dune is restored, but for opam it is easier to patch the opam file than to patch the dune-project.


kit-ty-kate:
Thanks!

PR #13165, opened 2018-12-17
switch objsize to backup copy since upstream is gone

ygrek:


ygrek:
what do you mean about old tags?

ygrek:
ocamlforge has only very old versions, for one they will not work with ocaml < 4.03.0

kit-ty-kate:
> what do you mean about old tags?

Was your old releases not handled by git tags?

> ocamlforge has only very old versions, for one they will not work with ocaml < 4.03.0

Yes but opam has this thing for keeping old versions around as long as it's not broken

ygrek:
This is not my project, I just setup a mirror from the checkout I have. Author is @gdsfh and it was hosted in bitbucket mercurial previously. Now the bitbucket account is no more and download links are broken. I only need the latest version so I recovered it from my local checkout and opam cache (note that md5sum checksum didn't change). I don't have release tarbal of 0.17 anywhere.

kit-ty-kate:
Is there no way of getting the archive for 0.17 through the opam archive cache? cc @rjbou ?

kit-ty-kate:
@ygrek hi, could you upload the old archive (https://opam.ocaml.org/cache/md5/fa/fa3e3f30fa9da04c7634738ec9726720) to bitbucket, put objsize.0.17 back with the new archive url and rebase your changes on top of master?

pmetzger:
I'll just note that there's a place for project that lost their upstream in ocaml-community, provided they continue to have users.

ygrek:
@kit-ty-kate done
@pmetzger I adopted it

kit-ty-kate:
Thanks a lot!

PR #12982, opened 2018-11-14
Fix archive access for users using recent openssl

kit-ty-kate:
With recent openssl (at least >= 1.1.1, maybe before as well) and curl the result of getting those archives is:
```
curl: (35) error:1425F102:SSL routines:ssl_choose_client_version:unsupported protocol
```
I contacted @gildor478 and the solution is to use either http or forge-static.ocamlcore.org. In this PR I choose the later.

PR #13167, opened 2018-12-17
Use the correct url to archives in ocamlforge

kit-ty-kate:
SSL is not correctly configured on the former (see https://github.com/ocaml/opam-repository/pull/12982)


kit-ty-kate:
cc @gildor478

gildor478:
Because they have been released after the snapshot for forge-static (forge-static is a snapshot).

PR #13170, opened 2018-12-18
Added cloudi.1.7.5

okeuday:


kit-ty-kate:
Could you return the fix upstream? `num.0` already pulls `base-num` so the disjunction complicates things unnecessarily. I'll merge it once the CI is all green

okeuday:
@kit-ty-kate I think the disjunction removal had a timeout in Travis, causing one of the builds there to fail, but if you hit the restart button at https://travis-ci.org/ocaml/opam-repository/jobs/469461300 , it should succeed.  That should be all that is required to make the CI green.

kit-ty-kate:
Thanks!

PR #13192, opened 2018-12-22
Add mpris.0.1.2

johnelse:


johnelse:
Actually this needs some rework, so closing for now.

PR #13193, opened 2018-12-23
ipaddr.2.9.0 uses 'lang dune 1.6' in dune-project, add this version constraint in opam

hannesm:
see https://github.com/mirage/ocaml-ipaddr/issues/78


kit-ty-kate:
Thanks!

PR #13199, opened 2018-12-25
Added liquidsoap 1.3.5 and gd.1.0a5

toots:


toots:
Looking good here from what I can see. The `gd` build on `CentOS` will always fail b/c the distribution does not provide the `depext` package. This isn't the first time. For `RPM`-based distributions CI testing, I'd recommend switch to e.g. `Fedora`.

kit-ty-kate:
Have you considered creating a new conf- package for the external dependencies of gd?

toots:
@kit-ty-kate I'm not familiar with the ins and outs of having a separate `conf-` package for external dependencies. What are the pros and cons?

kit-ty-kate:
Thanks!

PR #13212, opened 2018-12-28
Add spotify-web-api.0.2.1

johnelse:


johnelse:
Done, and I've added some version constraints since those versions of `spotify-cli` are quite old now. I don't *think* the travis failures are related, any ideas?

johnelse:
Thanks!

PR #13226, opened 2019-01-05
restrict mirage-net-xen,netchannel to ipaddr < 3.0.0

hannesm:
//cc @avsm


hannesm:
NB: CI fails due to broken mirage-net-xen 1.6.X releases, I won't fix them (or if in a separate PR) -- this PR is good to be merged IMHO

kit-ty-kate:
thanks

PR #13232, opened 2019-01-07
[new release] conduit-lwt-unix, conduit, conduit-async, mirage-conduit

avsm:
CHANGES:

* Use Ipaddr 3.0.0+ interfaces (mirage/ocaml-conduit#284 by @avsm).
* Update opam metadata files to the opam 2.0 format (mirage/ocaml-conduit#284 by @avsm)
* Hook in an introduction ocamldoc page to the `conduit` odoc (mirage/ocaml-conduit#284 by @avsm)

### DIFF
--- a/packages/conduit-async/conduit-async.1.2.0/opam
+++ b/packages/conduit-async/conduit-async.1.2.0/opam
@@ -18,6 +18,7 @@ depends: [
   "ppx_sexp_conv" {< "v0.12"}
   "sexplib" {< "v0.12"}
   "conduit"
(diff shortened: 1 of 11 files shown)


hannesm:
to me it looks like `mirage-conduit` and `conduit-async` should both explicitly depend on `ipaddr {>= "3.0.0"}` since they use Ipaddr in their code

avsm:
It works due to the dependent conduit having the right constraint, so i'll just merge this for now and fix up the metadata in conduit directly.

PR #13237, opened 2019-01-09
conduit-async: direct dependency on ipaddr

hannesm:
as discussed https://github.com/ocaml/opam-repository/pull/13232#issuecomment-452397042


kit-ty-kate:
Thanks!

PR #13242, opened 2019-01-10
[ocaml-base-compiler]: fix the install section for old compilers

oandrieu:
move some install-related actions from the `build` section to the `install` section, otherwise they fail with sandboxing.

(tested with 3.09.3)


kit-ty-kate:
Thanks!

PR #13247, opened 2019-01-10
mirage-protocols: upper bound for ipaddr 3

hannesm:


hannesm:
macos ci times out after 50 minutes, apart from that looks good to merge :)

avsm:
This is leading to some uninstallable releases, but that'll be fixed with a new package

PR #13256, opened 2019-01-14
Add libtorch 1.0.1

LaurentMazare:
The difference with 1.0.0 is that mklml gets downloaded and installed for macos. This seems necessary in order to have the `torch` package working on macos.


avsm:
Thanks, we appreciate the extra effort you've gone to for making macOS work as well!

PR #13264, opened 2019-01-14
openai-gym version 0.01

mandel:
openai-gym is an OCaml binding for [openai-gym](https://github.com/openai/gym) open-source toolkit. 

Could you accept this new package?

Thanks


kit-ty-kate:
The lint errors pointed by camelus should be fixed now. Could you return the fix upstream?

mandel:
Thank you @kit-ty-kate for the fix!

kit-ty-kate:
Thanks

PR #13265, opened 2019-01-15
Torch 0.3.

LaurentMazare:
This is a tentative release for macos support now that the libtorch 1.0.1 opam package has been released.


LaurentMazare:
It seems that libtorch requires CLOCK_MONOTIC to be defined which only has been introduced in macOS Sierra 10.12.

kit-ty-kate:
Ok, I guess the CI can't test the package. I'll trust you it works and move on with merging. Thanks!

PR #13279, opened 2019-01-17
{gmp,zarith}-freestanding updates for Solo5 0.5.0+

mato:
In preparation for updates to {gmp,zarith,ocaml}-freestanding:

1. Constrain existing releases of {gmp,zarith}-freestanding to ocaml-freestanding {<= "0.5.0"}.

2. Constrain existing releases of zarith-freestanding to gmp-freestanding {<= "6.1.2-1"}.

----

The SSP changes in Solo5/solo5#294 break gmp-freestanding and zarith-freestanding due to issues with the various cross-compiling hacks used to build these packages.

Publish updated versions, whose availability will be "gated" on an upcoming ocaml-freestanding 0.5.0 release, which itself will depend on Solo5 0.5.0+ bindings packages.

For details, see mirage/ocaml-freestanding#47.

----

@hannesm:

gmp-freestanding 6.1.2-2 should technically work with ocaml-freestanding / Solo5 < 0.5.0, but I don't want to "support" those combinations, so over and above what we discussed and you tested in mirage/ocaml-freestanding#47 I've added an explicit dependency on ocaml-freestanding {>= "0.5.0"} to the new gmp-freestanding.

/cc @emillon FYI/review, just so you're aware of the general horrible-ness of the cross-compiling hacks currently involved here.

### DIFF
--- a/packages/gmp-freestanding/gmp-freestanding.6.0.0-1/opam
+++ b/packages/gmp-freestanding/gmp-freestanding.6.0.0-1/opam
@@ -14,7 +14,7 @@ remove: [
 ]
 depends: [
   "ocaml"
-  "ocaml-freestanding"
+  "ocaml-freestanding" { < "0.5.0" }
   "conf-m4" {build}
 ]
 available: arch = "x86_64" | arch = "x86_64"
--- a/packages/gmp-freestanding/gmp-freestanding.6.0.0/opam
+++ b/packages/gmp-freestanding/gmp-freestanding.6.0.0/opam
@@ -14,7 +14,7 @@ remove: [
 ]
 depends: [
   "ocaml"
-  "ocaml-freestanding"
+  "ocaml-freestanding" { < "0.5.0" }
   "conf-m4" {build}
 ]
 available: arch = "x86_64" | arch = "x86_64"
--- a/packages/gmp-freestanding/gmp-freestanding.6.1.2-1/opam
+++ b/packages/gmp-freestanding/gmp-freestanding.6.1.2-1/opam
@@ -14,7 +14,7 @@ remove: [
 ]
 depends: [
   "ocaml"
-  "ocaml-freestanding" {>= "0.2.3"}
+  "ocaml-freestanding" {>= "0.2.3" & < "0.5.0"}
 ]
 synopsis: "The GNU Multiple Precision Arithmetic Library"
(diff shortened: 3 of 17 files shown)


hannesm:
thanks! IMHO we should try to upstream the configure patch of zarith-freestanding (not sure whether someone already did this?) to zarith (https://github.com/ocaml/zarith)

mato:
@hannesm I considered that, but if you look at what I need to do in the patch (including the comments at the top), I couldn't figure out an immediately sensible way of making those changes suitable for upstream.

mato:
@hannesm Hmm, so I thought I could push the new gmp-freestanding, zarith-freestanding without the corresponding ocaml-freestanding 0.5.0, but it seems not. Am going to try and drop the `depends: ocaml-freestanding { >= 0.5.0 }` on those and re-push, that seems like the simplest option.

hannesm:
uhm, why? I agree with your "which combinations of gmp/zarith and ocaml-freestanding are supported" -- and would rather appreciate to have this merged (yes, it adds some packages that are not installable, but there'll be a ocaml-freestanding.0.5.0 soon, and it doesn't hurt to have the newer gmp&zarith-freestanding around in opam-repository IMHO).

and yes, i looked deeper in the zarith configure patch, looks like various things that are not really upstreamable (apart from maybe the nodynlink configure option)

mato:
@hannesm Well, that's up to you, but with non-installable packages I have no way of passing CI which is not ideal, as I'd like at least some confirmation that the new packages don't introduce any unexpected problems.

How about this: I could relax the ocaml-freestanding dependency to `{ >= 0.4.1 }` (i.e. after the migration to OPAM 2 and with the new target names introduced in 0.4.x). That way we get at least some indication of CI passing, but don't bother with "historic" versions.

WDYT?

hannesm:
sure, as you like. we can also once ocaml-freestanding 0.5.0 is out, move the lower bounds to 0.5.0.

mato:
AFAICT, the Travis failures are all due to the job getting terminated due to the build log being too long. The Fedora failures are due to "cannot obtain transaction lock" when installing depexts, so presumably transient. So, if camelus goes green I think we can merge this.

hannesm:
thanks, I restarted the fedora builders and now they're happy. the travis failure is "log too long", and camelus is happy with new installable packages! merging...

PR #13097, opened 2018-12-06
Fix compilation issues discovered by CI

Kakadu:
* bad checksum
* missing dependency

It would be great if compilation errors would be sent on my email upon discovery.

PR #13287, opened 2019-01-18
Export extra libs in labltk META

pirbo:
Labltk installs some extra libraries without advertising them in the META file. This patch makes the advertisement.

I' sorry and sad to work at the opam level. I tried to modify that upstream 4 years ago: https://forge.ocamlcore.org/tracker/index.php?func=detail&aid=1444&group_id=343&atid=1353 without any reaction.

It is fully backward compatible (it simply add opportinuties). It would have make my life so easier 4 years ago when I moved the build system of KaSim to ocamlbuild and it simplifies so much the migration to dune today that I'm considering this non optimal solution.

### DIFF
--- a/packages/conf-tcl/conf-tcl.1/opam
+++ b/packages/conf-tcl/conf-tcl.1/opam
@@ -9,6 +9,8 @@ depexts: [
   ["tcl-dev"] {os-distribution = "debian"}
   ["tcl-dev"] {os-distribution = "ubuntu"}
   ["tcl"] {os-distribution = "nixos"}
+  ["tcl-dev"] {os-family = "alpine"}
+  ["tcl-devel"] {os-family = "rhel"}
 ]
 synopsis: "Virtual package relying on tcl"
 description:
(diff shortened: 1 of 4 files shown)


avsm:
Many thanks for this patch! I notice that @garrigue has got a GitHub mirror of the project up at https://github.com/garrigue/labltk 

Since this is a long-standing project that was formerly in the compiler distribution, should we move it to the GitHub ocaml/ organisation and migrate away from the deprecated OCamlForge, Jacques?  I can assist with the repo migration if so, and we could integrate this patch in a new release at the same time.

garrigue:
I would be perfectly happy with your taking over labltk. Note however that there is something I would like seen done : separating OCamlbrowser into a separate package. This would avoid the need to update labltk for each OCaml version.

garrigue:
By the way, I have recently done some important fixes to OCamlbrowser, so it would be a good time to make a release (but I have little time to do it)

mseri:
Is this PR related to https://github.com/ocaml/opam-repository/pull/13097 ?

garrigue:
I don't think so. labltk and lablqt are unrelated projects.

mseri:
The ivy failures in the CI are unrelated (and due to safe-string incompatibility), but the installation is still failing on Alpine with 
```
# Error: Error on dynamically loaded library: ../support/dlllabltk.so: Error relocating ../support/dlllabltk.so: Tcl_Merge: symbol not found
```
This is probably due to the specific Alpine versions but I would like to understand from you if it is an expected failure or there is more needed to do

pirbo:
The problem with alpine is a problem of alpine: the package `tk-dev` does not depend upon the package `tk`...

kit-ty-kate:
Thanks!

PR #13289, opened 2019-01-18
Fix menhir dependency

Gbury:
Dolmen has recently switched to dune, which seems to use menhir options introduced very recently, thus the version bound on menhir actually need to be updated.


kit-ty-kate:
Thanks

PR #13299, opened 2019-01-21
[new release] odepack (0.7)

Chris00:
Binding to ODEPACK

- Project page: <a href="https://github.com/Chris00/ocaml-odepack">https://github.com/Chris00/ocaml-odepack</a>
- Documentation: <a href="https://chris00.github.io/ocaml-odepack/doc/">https://chris00.github.io/ocaml-odepack/doc/</a>

##### CHANGES:

- Set the `-custom` flag for bytecode because `libgfortan` is not
  compiled with `-fPIC`.
- Fix package name in configure script.
- Use `conf-gfortran`.

### DIFF
--- a/packages/odepack/odepack.0.7/opam
+++ b/packages/odepack/odepack.0.7/opam
@@ -11,8 +11,6 @@ build: [
   ["dune" "subst"] {pinned}
   ["dune" "build" "-p" name "-j" jobs]
   ["dune" "build" "@doc"] {with-doc}
-]
-run-test: [
   ["dune" "runtest" "-p" name "-j" jobs] {with-test}
 ]
 depends: [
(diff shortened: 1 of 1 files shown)


packages/odepack/odepack.0.7/opam
```diff
@@ -0,0 +1,33 @@
...
+bug-reports: "https://github.com/Chris00/ocaml-odepack/issues"
+doc: "https://chris00.github.io/ocaml-odepack/doc/"
+tags: [ "ODE" "scientific-computing"  ]
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "build" "@doc"] {with-doc}
+]
+run-test: [
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
```
kit-ty-kate: this doesn't work. It's either a command in `run-test` without `{with-test}` or a command in `build` with `{with-test}`. The combination of the two will just do nothing

kit-ty-kate: it's fixed now. Could you return the fix upstream?

Chris00: [Done](https://github.com/Chris00/ocaml-odepack/commit/32e92b57093333db4acb2d8b6a8b8eac123df63c).  Thanks.

PR #13300, opened 2019-01-21
Package tablecloth-native.0.0.2

pbiggar:
### `tablecloth-native.0.0.2`
Native OCaml library implementing Tablecloth, a cross-platform standard library for OCaml, Bucklescript and ReasonML
Longer description

---
* Homepage: https://github.com/darklang/tablecloth
* Source repo: git://github.com/darklang/tablecloth
* Bug tracker: https://github.com/darklang/tablecloth/issues

---

### DIFF
--- a/packages/tablecloth-native/tablecloth-native.0.0.2/opam
+++ b/packages/tablecloth-native/tablecloth-native.0.0.2/opam
@@ -1,0 +1,23 @@
+opam-version: "2.0"
+synopsis:
+  "Native OCaml library implementing Tablecloth, a cross-platform standard library for OCaml, Bucklescript and ReasonML"
+description: "Longer description"
+maintainer: "Paul Biggar <paul@darklang.com>"
+authors: "Paul Biggar <paul@darklang.com>"
+license: "MIT with some exceptions"
+homepage: "https://github.com/darklang/tablecloth"
+bug-reports: "https://github.com/darklang/tablecloth/issues"
+depends: [
+  "ocaml"
+  "dune" {build}
+  "base" {>= "v0.10.0"}
+]
+build: ["dune" "build" "-p" name "-j" jobs]
+dev-repo: "git://github.com/darklang/tablecloth"
+url {
+  src: "https://github.com/darklang/tablecloth/archive/0.0.2.tar.gz"
+  checksum: [
+    "md5=5671b9cd3aef00a767d2f7811065e68d"
+    "sha512=3330a63e0d828e3c891f4cd66922310747e89f666ecab7fd0c5b7564d4c32bce192cdb08293dace0a1255cd9da4e793e0c1f142cce7d25e260bd0b0649e88348"
+  ]
+}


pbiggar:
Are there logs for camelus? How do I debug this?

kit-ty-kate:
Unfortunately there is no detailed log for Camelus but the problems are very obvious here: the package can't be installed because `base >= v10.0` doesn't exist. You probably mean `base >= v0.10.0` instead.

packages/tablecloth-native/tablecloth-native.0.0.2/opam
```diff
@@ -0,0 +1,25 @@
...
+  "Native OCaml library implementing Tablecloth, a cross-platform standard library for OCaml, Bucklescript and ReasonML"
+description: "Longer description"
+maintainer: "Paul Biggar <paul@darklang.com>"
+authors: "Paul Biggar <paul@darklang.com>"
+license: "MIT with some exceptions"
+homepage: "https://github.com/darklang/tablecloth"
+bug-reports: "https://github.com/darklang/tablecloth/issues"
+depends: [
+  "ocaml"
+  "ocamlfind"
```
kit-ty-kate: This dependency doesn't seem to be used anywhere in your source code. Could you remove it?

packages/tablecloth-native/tablecloth-native.0.0.2/opam
```diff
@@ -0,0 +1,25 @@
...
+  "dune"
```
kit-ty-kate:
```suggestion
  "dune" {build}
```
This dependency should be tagged with the `{build}` attribute

packages/tablecloth-native/tablecloth-native.0.0.2/opam
```diff
@@ -0,0 +1,25 @@
...
+  "base" {>= "v10.0"}
+  "str"
+]
+build: ["dune" "build" "-p" name]
```
kit-ty-kate:
```suggestion
build: ["dune" "build" "-p" name "-j" jobs]
```
The idiomatic call to dune usually have also `"-j" jobs` so than people can parallelize (or not) as they wish

packages/tablecloth-native/tablecloth-native.0.0.2/opam
```diff
@@ -0,0 +1,25 @@
...
+maintainer: "Paul Biggar <paul@darklang.com>"
+authors: "Paul Biggar <paul@darklang.com>"
+license: "MIT with some exceptions"
+homepage: "https://github.com/darklang/tablecloth"
+bug-reports: "https://github.com/darklang/tablecloth/issues"
+depends: [
+  "ocaml"
+  "ocamlfind"
+  "dune"
+  "base" {>= "v10.0"}
```
kit-ty-kate:
```suggestion
  "base" {>= "v0.10.0"}
```

packages/tablecloth-native/tablecloth-native.0.0.2/opam
pbiggar: 👍

packages/tablecloth-native/tablecloth-native.0.0.2/opam
pbiggar: 👍

packages/tablecloth-native/tablecloth-native.0.0.2/opam
pbiggar: 👍

packages/tablecloth-native/tablecloth-native.0.0.2/opam
pbiggar: 👍

pbiggar:
Thanks for the review and for catching that base version problem! Fixed them all and resubmitted.

packages/tablecloth-native/tablecloth-native.0.0.2/opam
```diff
@@ -0,0 +1,24 @@
...
+maintainer: "Paul Biggar <paul@darklang.com>"
+authors: "Paul Biggar <paul@darklang.com>"
+license: "MIT with some exceptions"
+homepage: "https://github.com/darklang/tablecloth"
+bug-reports: "https://github.com/darklang/tablecloth/issues"
+depends: [
+  "ocaml"
+  "dune" {build}
+  "base" {>= "v0.10.0"}
+  "str"
```
kit-ty-kate:
```suggestion
```
There is still a problem. This package doesn't exists. My guess is that you wanted to tell that you are using the `Str` module. However this is a builtin module installed by the compiler.

pbiggar: I added this in response to a build failure on this CI. I can't find the log but the error was "No implementations provided for the following modules: Str", and stackoverflow suggested this: https://stackoverflow.com/questions/38547815/no-implementations-provided-for-the-following-modules-str.

I'll remove it and see what happens and we can debug from there.

Thanks again :)

kit-ty-kate: str is vendored as a separate library even if it's installed by the compiler. You need to add it explicitly in your dune file for it to be linked

pbiggar: Gotcha, thanks! That seems to have fixed it :)

PR #13301, opened 2019-01-22
Configuration package for zstd

xclerc:
We have pushed a library depending on zstd [here](https://github.com/janestreet/zstandard), but
I was told that it would be better to have a configuration
package rather than embed the external dependencies
in the [opam file](https://github.com/janestreet/zstandard/blob/master/zstandard.opam).

(If this package is deemed correct / useful, I will then open
another pull request to make [other packages](https://github.com/ocaml/opam-repository/tree/master/packages/zstd) depending
on `zstd` use this configuration package.)

### DIFF
--- a/packages/conf-zstd/conf-zstd.0.1/opam
+++ b/packages/conf-zstd/conf-zstd.0.1/opam
@@ -7,9 +7,9 @@ license: "BSD"
 depexts: [
   ["libzstd-dev"] {os-distribution = "debian"}
   ["libzstd-dev"] {os-distribution = "ubuntu"}
-  ["libzstd-devel"] {os-distribution = "centos"}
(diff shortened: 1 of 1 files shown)


xclerc:
Looks like the package name was wrong for Fedora/CentOS;
just pushed a tentative fix.

kit-ty-kate:
Is there no ways of testing if the library is indeed installed, using pkg-config for instance?

kit-ty-kate:
The problem on macos seems related to the homebrew package. Merging. Thanks!

xclerc:
Great; thanks!

I have just opened https://github.com/ocaml/opam-repository/pull/13303 to patch the dependencies
of the packages mentioned in my first message.

PR #13302, opened 2019-01-22
Add ocaml variants for branch 4.08

damiendoligez:
also add ocaml versions 4.08.0 and 4.09.0, as they are needed to install the corresponding variants.

Can someone confirm that this is the right thing to do? I'm surprised that ocaml.4.08.0 didn't exist already.


kit-ty-kate:
@avsm wasn't there a problem with unreleased ocaml versions like 4.08?

fdopen:
Yes, they were removed with intent:
https://github.com/ocaml/opam-repository/pull/12646
https://github.com/ocaml/opam/issues/3510

damiendoligez:
Ah yes now I remember. So I removed them from this PR.

kit-ty-kate:
the thing is that now the packages are not installable since they require ocaml = 4.0{8,9}.0. I'm not sure what to do to be honest

avsm:
Until the solver selection is fixed, I don't think we can add 4.08 beta1 packages in the default remote.  @rjbou @altgr is there an issue open on the opam tracker about avoiding beta versions of the compiler in default selection?

We could add these into a separate remote for users instead; perhaps ocaml/opam-beta-repository or similar that they add as part of trying the 4.08 branches out.

kit-ty-kate:
should we close this PR?

avsm:
@damiendoligez since it'll take a while to deploy a fix to opam, I've pushed all the beta compilers into a new remote at https://github.com/ocaml/ocaml-beta-repository. That repository has CI for amd64 and arm64 for the compilers in there, so it'll be build tested. The README has instructions for how users can activate the remote, so it should be good to push the 4.08 beta releases to.

If this is a good enough solution, we can close this PR and redirect users there.  opam-repo, caml-devel groups both have write access to the ocaml-beta-repository git repo as well.

rjbou:
@avsm ocaml/opam#3510 is open, and as said @AltGr in https://github.com/ocaml/opam/issues/3510#issuecomment-414221365 there is no satisfying solution using the solver ftm.
_Best_ approach is what you set up: move unreleased or even `ocaml-variant` compiler packages to another repo, like that users have choice on _usual_ compilers, and compiler experimenter users who want to test / use variant compilers, just have to add a repo to access all compilers.

avsm:
I don't quite understand why we need a totally general solution though. This is in the specific case of switch creation, so isn't there some metadata tag we can or some heuristic to detect and filter out the dev versions before they even hit the solver?  For instance, if the compiler is a `git://` url that's a reasonable hint to not select it by default, or we could just add `beta` tags to the package.

damiendoligez:
I'll close this PR because the interim solution suits me, but I'd like to see a more permanent solution: IMO it is really important to let everyone try the beta versions with a very low barrier to entry (i.e. one-line installation instructions).

AltGr:
A one-liner is possible:
```
opam switch create ocaml-variants.4.08.0+trunk --repositories=default,beta=git+https://github.com/ocaml/ocaml-beta-repository.git
```

I would also be in favor of still keeping the `ocaml.4.08.0`, `ocaml.4.09.0` in the default repository: they won't do harm if there is no implementation package available (they'll just be uninstallable), and will help when we add them, of for e.g. peaple wanting to try pinnings, etc.

PR #13311, opened 2019-01-23
flac.0.1.4

toots:


packages/flac/flac.0.1.4/opam
```diff
@@ -0,0 +1,46 @@
...
+  ] {os = "macos"}
+  [make]
+]
+install: [
+  [make "install"]
+]
+remove: ["ocamlfind" "remove" "flac"]
+depends: [
+  "ocaml" {>= "4.02"}
+  "ocamlfind"
```
kit-ty-kate:
```suggestion
  "ocamlfind" {build}
```
ocamlfind should probably be flagged with the `{build}` attribute as it doesn't seem to be used anywhere at runtime

kit-ty-kate:
Could you return the change to the source repository? I'll merge this as soon as the CI is done

kit-ty-kate:
Thanks!

PR #13314, opened 2019-01-23
Package batsat.0.3.1

c-cube:
### `batsat.0.3.1`
OCaml bindings for batsat, a SAT solver in rust

---
* Homepage: https://github.com/c-cube/batsat-ocaml/
* Source repo: git+https://github.com/c-cube/batsat-ocaml.git
* Bug tracker: https://github.com/c-cube/batsat-ocaml/issues

---

### DIFF
--- a/packages/batsat/batsat.0.3.1/opam
+++ b/packages/batsat/batsat.0.3.1/opam
@@ -1,0 +1,26 @@
+opam-version: "2.0"
+synopsis: "OCaml bindings for batsat, a SAT solver in rust"
+maintainer: "simon.cruanes.2007@m4x.org"
+build: [
+  ["dune" "build" "-p" name]
+  ["dune" "build" "@doc" "-p" name] {with-doc}
+  ["dune" "runtest" "-p" name] {with-test}
+]
+# TODO: add dependency on cargo?
+depends: [
+  "dune" {build & >= "1.3.0" }
+  "odoc" {with-doc}
+  "conf-rust" {build}
+]
+tags: [ "minisat" "solver" "SAT" ]
+homepage: "https://github.com/c-cube/batsat-ocaml/"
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Getting a precise version of an external package is a bit tricky.

I'm not sure how to test this package. Your own Travis runs seem to fail too. Could you try to make them pass?

c-cube:
I think maybe the `rust-conf` package could have versions that match _capabilities_ (not the exact rust version). Something like `rust-conf.with-2018` (that will successfully install iff the rust compiler is recent enough)?

Anyway, let me close this for now :/

kit-ty-kate:
> I think maybe the `rust-conf` package could have versions that match _capabilities_ (not the exact rust version). Something like `rust-conf.with-2018` (that will successfully install iff the rust compiler is recent enough)?

One of the few conf- package I know that have something like that is `conf-gtk3.18` which just checks that the current version installed is at least `3.18` but it can't make the user install the correct version because that would mean dealing with distribution versions and it would be extremely painful. I guess something like that can be done for rust however it doesn't really help the CI problem

c-cube:
For the CI system I don't know what to do except wait for it to use a distro with recent enough packages. Even for my own CI, which is pretty annoying.

kit-ty-kate:
> Even for my own CI, which is pretty annoying.

Fixed by https://github.com/c-cube/batsat-ocaml/pull/1

c-cube:
I'll try to make a package `conf-rust.2018` first, I think. Thanks for your help!

kit-ty-kate:
I'm not sure whether `2018` is a good version name to be honest. I'd rather keep it the same as rust version number to avoid getting stuck with `conf-rust.2018.1.32` or something of the sort if some packages need further constraints.

PR #13315, opened 2019-01-24
nlopt-ocaml: remove ocaml-version upper bound

UnixJunkie:
also added depext for brew and missing opam fields


kit-ty-kate:
The build seems to have been canceled :/ I don't know where it comes from but I've restarted the osx job at least. The other builds are a bit irrelevant. I'll merge it as soon as it's finished

kit-ty-kate:
The osx job failed because of the dependency of nlopt on mac is gcc and it tries to compile gcc which takes over the time limit allocated by Travis :/

I think it's still fine to merge though. Thanks!

UnixJunkie:
ok, thanks for the feedback.
Why don't they put gcc in a bottle (the brew stuffs which install fast)?!

PR #13316, opened 2019-01-24
Package morsmall.0.1

Niols:
### `morsmall.0.1`
A concise AST for POSIX shell
A concise AST for POSIX shell

---
* Homepage: https://github.com/colis-anr/morsmall
* Source repo: git+ssh://git@github.com/colis-anr/morsmall.git
* Bug tracker: https://github.com/colis-anr/morsmall/issues

---

### DIFF
--- a/packages/morsmall/morsmall.0.1/opam
+++ b/packages/morsmall/morsmall.0.1/opam
@@ -1,0 +1,32 @@
+opam-version: "2.0"
+
+synopsis: "A concise AST for POSIX shell"
+description: """
+A concise AST for POSIX shell
+"""
+
+license: "GPL3"
+
+maintainer: "Nicolas Jeannerod <nicolas.jeannerod@irif.fr>"
+authors: [ "Nicolas Jeannerod <nicolas.jeannerod@irif.fr>" ]
+
+homepage: "https://github.com/colis-anr/morsmall"
+bug-reports: "https://github.com/colis-anr/morsmall/issues"
+dev-repo: "git+ssh://git@github.com/colis-anr/morsmall.git"
+
+depends: [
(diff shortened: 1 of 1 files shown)


packages/morsmall/morsmall.0.1/opam
```diff
@@ -0,0 +1,32 @@
...
+authors: [ "Nicolas Jeannerod <nicolas.jeannerod@irif.fr>" ]
+
+homepage: "https://github.com/colis-anr/morsmall"
+bug-reports: "https://github.com/colis-anr/morsmall/issues"
+dev-repo: "git+ssh://git@github.com/colis-anr/morsmall.git"
+
+depends: [
+  "dune"    {build}
+  "morbig"
+  "ocaml"   {build & >= "4.04"}
```
kit-ty-kate: I don't think that would necessary cause problems but I think you'd expect to have `ocaml` as a normal dependency here and not a build dependency

Niols: Oh? Why?

kit-ty-kate: Because you're making a library which intrinsically relies on OCaml to be used. If a user of your library is using the bytecode files, their resulting binary will call `ocamlrun` at runtime. In this case it's not **really** necessary but I think it's nice to have it at least a hint when reading the opam file.

kit-ty-kate:
Thanks!

packages/morsmall/morsmall.0.1/opam
Niols: OK, it makes sense. Thank you for the explanation :-)

PR #13317, opened 2019-01-24
Add some sdl depexts on osx/homebrew.

toots:


kit-ty-kate:
Something is very wrong with the osx machine but it's fine and we can merge this anyway. Thanks!

PR #13328, opened 2019-01-28
added virtual package for ocamlbrowser

UnixJunkie:


kit-ty-kate:
I'm not too sure about having a virtual package like this. At least the versioning seems a bit off as it feels like it's a normal version number but it can pull any versions of labltk

UnixJunkie:
Well, it is a virtual package for the moment.
I guess that if people want a special version of labltk, they know how to do this.

garrigue:
This should not be a virtual package, rather labltk should be split two packages, since ocamlbrowser shouldn't depend on the labltk version, and labltk shouldn't depend on the OCaml version. If anybody is interested to help, tell me.

kit-ty-kate:
I agree with @garrigue. Closing

PR #13329, opened 2019-01-28
Package headache.1.04

pbaudin:
New version for `headache` package including patches for UTF-8 encoding.

As said at Ocaml forge https://forge.ocamlcore.org/projects/headache, the project as migrated to https://github.com/Frama-C/headache.

### DIFF
--- a/packages/headache/headache.1.04/opam
+++ b/packages/headache/headache.1.04/opam
@@ -23,25 +23,23 @@ authors: [
 maintainer: "Patrick Baudin"
 homepage: "https://github.com/Frama-C/headache/"
 bug-reports: "https://github.com/Frama-C/headache/issues"
-dev-repo: "https://github.com/Frama-C/headache.git"
+dev-repo: "git+https://github.com/Frama-C/headache.git"
 
 depends: [
   "camomile"
-  "ocamlfind" {build}
-  "ocamlbuild" {build}
+  "dune" {build}
 ]
 
 build: [
-  [make]
+  [ "dune" "build" "-p" "headache" ]
 ]
 
-install: [
+install:  [
   [make "INSTALLDIR=%{prefix}%/bin" "install"]
   [make "DOC_INSTALLDIR=%{prefix}%/doc/headache" "install-doc"] {with-doc}
 ]
 
-extra-files: ["headache.install" "md5=698fde7c6498aa82223d679fae952d69"]
 url {
   src: "https://github.com/Frama-C/headache/archive/v1.04.tar.gz"
-  checksum: "md5=35d75338793d1f17de6de504909221ae"
+  checksum: "md5=d551319dcb2616d660f2cb04021602aa"
 }
--- a/packages/headache/headache.1.04/opam
+++ b/packages/headache/headache.1.04/opam
@@ -1,0 +1,45 @@
+opam-version: "2.0"
+name: "headache"
+version: "1.04"
+
+license: "GNU Library General Public License"
+
+synopsis: "Automatic generation of files headers"
+description: """
+Lightweight tool for managing headers in source code files. It can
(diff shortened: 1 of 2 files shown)


packages/headache/headache.1.04/opam
```diff
@@ -0,0 +1,47 @@
+opam-version: "2.0"
+name: "headache"
+version: "1.04"
```
mseri: name and version are redundant, the are inferred from the path (and overridden). I can remove them before merging, unless you need to make further changes to the opam file

packages/headache/headache.1.04/opam
```diff
@@ -0,0 +1,47 @@
...
+  "camomile"
+  "ocamlfind" {build}
+  "ocamlbuild" {build}
+]
+
+build: [
+  [make]
+]
+
+install: [
```
kit-ty-kate: Does the `install` clause do anything different than installing what's already listed in the `headache.install` file that you furnished?

pbaudin: We just modify the build state in order to use `dune` that generates such a file. This file is no more needed into the `opam-repository`.

packages/headache/headache.1.04/opam
pbaudin: Thanks for the proposal. I will leave you remove such lines once the automatic checks will be ok.
Could you also have a look to the `dev-repo` field ?

packages/headache/headache.1.04/opam
kit-ty-kate: ok perfect, then with dune this clause shouldn't be necessary. Could you remove it?

packages/headache/headache.1.04/opam
```diff
@@ -0,0 +1,45 @@
...
+bug-reports: "https://github.com/Frama-C/headache/issues"
+dev-repo: "git+https://github.com/Frama-C/headache.git"
+
+depends: [
+  "camomile"
+  "dune" {build}
+]
+
+build: [
+  [ "dune" "build" "-p" "headache" ]
```
kit-ty-kate: The more standard way of calling dune build is `"dune" "build" "-p" name "-j" jobs`. `"headache"` is replaced by `name` which is more generic and avoid to do mistakes if it gets renamed or copy-paste and `"-j" jobs` is useful for people wanting to let dune have the full power of their CPU cores or on the contrary limit to be able to do other task while it's installing

packages/headache/headache.1.04/opam
pbaudin: The file `headache.install` generated by `dune` is similar to the proposed one. So,
-  `make "INSTALLDIR=%{prefix}%/bin" "install` command installs only the binary listed into `headache.install`
- `make "DOC_INSTALLDIR=%{prefix}%/doc/headache" "install-doc"`  command (executed on the request of `with-doc`) installs the documentation files listed into `headache.install`.

packages/headache/headache.1.04/opam
pbaudin: Of course, let's go for the standard way...

pbaudin:
So, as far I can undertand, I need to:
- remove `version:`, `name:` and `install:` fields
- modify the `dune` command of the `build:` field

That will be done into my next commit.

packages/headache/headache.1.04/opam
kit-ty-kate: Those two commands install things that are already installed through `headache.install` (generated by `dune build`) for what I can see. I'm pretty sure those two commands are unnecessary

packages/headache/headache.1.04/opam
pbaudin: Fields `name:` and `version:` have been removed (see commit  4c202b3)

PR #13332, opened 2019-01-28
Package mssql.1.0

brendanlong:
### `mssql.1.0`
Mssql wraps FreeTDS in a nicer and safer interface, with support for parameterized queries, thread-based async IO, and a thread pool.

---
* Homepage: https://github.com/arenadotio/ocaml-mssql
* Source repo: git+https://github.com/arenadotio/ocaml-mssql.git
* Bug tracker: https://github.com/arenadotio/ocaml-mssql/issues

---

### DIFF
--- a/packages/mssql/mssql.1.0/opam
+++ b/packages/mssql/mssql.1.0/opam
@@ -1,0 +1,35 @@
+opam-version: "2.0"
+authors: ["Arena Developers <silver-snakes@arena.io>"]
+maintainer: "silver-snakes@arena.io"
+homepage: "https://github.com/arenadotio/ocaml-mssql"
+dev-repo: "git+https://github.com/arenadotio/ocaml-mssql.git"
+bug-reports: "https://github.com/arenadotio/ocaml-mssql/issues"
+doc: "https://arenadotio.github.io/ocaml-mssql/doc"
+synopsis: "Async SQL Server client using FreeTDS"
+description: "Mssql wraps FreeTDS in a nicer and safer interface, with support for parameterized queries, thread-based async IO, and a thread pool."
+
+build: [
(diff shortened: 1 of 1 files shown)


packages/mssql/mssql.1.0/opam
```diff
@@ -0,0 +1,35 @@
...
+maintainer: "silver-snakes@arena.io"
+homepage: "https://github.com/arenadotio/ocaml-mssql"
+dev-repo: "git+https://github.com/arenadotio/ocaml-mssql.git"
+bug-reports: "https://github.com/arenadotio/ocaml-mssql/issues"
+doc: "https://arenadotio.github.io/ocaml-mssql/doc"
+synopsis: "Async SQL Server client using FreeTDS"
+description: "Mssql wraps FreeTDS in a nicer and safer interface, with support for parameterized queries, thread-based async IO, and a thread pool."
+
+build: [
+  ["jbuilder" "build" "-p" name "-j" jobs]
```
kit-ty-kate: you seem to be using `dune` instead of jbuilder according to your source code. Is that correct? If so I think it would be slightly better to use the `dune`commands instead of `jbuilder`

brendanlong: Running `dune` turns more warnings into errors so it won't actually build that way. I'll look into making that change in a future release but it's a non-trivial change.

brendanlong: Actually maybe it's only a few warnings to fix..

brendanlong:
See #13334

(I'm not sure if there's a better way to update these when making changes? I just ran `opam publish` again)

kit-ty-kate:
> (I'm not sure if there's a better way to update these when making changes? I just ran `opam publish` again)

opam-publish should be able to force-push into this PR for what I understand

PR #13333, opened 2019-01-28
plain -no-flat-float-array ocaml variants for 4.{06,07}.{0,1}

gasche:
The -no-flat-float-array switches currently available are only with
+flambda. To benchmark the performance impact this option, it is
better if non-flambda switches with the option are also available.

(-no-flat-float-array was made available in 4.06.0, so this commit
covers all possible versions)


kit-ty-kate:
Thanks

PR #13337, opened 2019-01-29
added packages/parany/parany.5.0.1

UnixJunkie:
performance bug correction, seen on long running
computation that can make the jobs queue full


mseri:
Thanks

PR #13338, opened 2019-01-29
Hts shrink 210

UnixJunkie:
minor release


packages/hts_shrink/hts_shrink.2.1.0/opam
```diff
@@ -0,0 +1,39 @@
+opam-version: "2.0"
+name: "hts_shrink"
```
mseri: name and version are overridden by the values inferred from the path. I'll delete this line before merging.

kit-ty-kate:
Thanks!

PR #13342, opened 2019-01-30
[new release] yojson (1.6.0)

NathanReb:
Yojson is an optimized parsing and printing library for the JSON format

- Project page: <a href="https://github.com/ocaml-community/yojson">https://github.com/ocaml-community/yojson</a>
- Documentation: <a href="https://ocaml-community.github.io/yojson/">https://ocaml-community.github.io/yojson/</a>

##### CHANGES:

*2019-01-30*

### Deprecate

- `json` types are deprecated in favor of their new `t` aliases, ahead of their removal in the next
  major release (ocaml-community/yojson#73, @Leonidas-from-XIV)

### Add

- Add a type `t` and monomorphic `equal`, `pp` and `show` (ocaml-community/yojson#73, @Leonidas-from-XIV)

### DIFF
--- a/packages/yojson/yojson.1.6.0/opam
+++ b/packages/yojson/yojson.1.6.0/opam
@@ -1,0 +1,38 @@
+opam-version: "2.0"
+maintainer: "martin@mjambon.com"
+authors: ["Martin Jambon"]
+homepage: "https://github.com/ocaml-community/yojson"
+bug-reports: "https://github.com/ocaml-community/yojson/issues"
+dev-repo: "git+https://github.com/ocaml-community/yojson.git"
+doc: "https://ocaml-community.github.io/yojson/"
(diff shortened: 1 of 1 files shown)


NathanReb:
This triggers warning indeed but I don't think they're fatal and the revdeps builds seem to fail for another reason but I haven't been through all the failures, I'll try to see of some of those are due to the warning.

If some indeed are, the best thing to do would be to make it so the build command doesn't fail on warning 3 when that's possible.

NathanReb:
Doesn't seem so, I think this is good to go!

NathanReb:
@mseri could you please take a quick look at this when you have time? I don't think there's anything blocking!

kit-ty-kate:
Thanks!

PR #13344, opened 2019-01-30

xclerc:
Follow-up to https://github.com/ocaml/opam-repository/pull/13010, https://github.com/ocaml/opam-repository/pull/13016, and https://github.com/ocaml/opam-repository/pull/13107.

### DIFF
--- a/packages/biocaml/biocaml.0.9.0/opam
+++ b/packages/biocaml/biocaml.0.9.0/opam
@@ -29,20 +29,20 @@ depends: [
   "ocaml"
   "base64" {< "3.0.0"}
   "dune" {build}
-  "core_kernel" {>= "v0.11.0"}
-  "sexplib"
+  "core_kernel" {>= "v0.11.0" & < "v0.12"}
+  "sexplib" {< "v0.12"}
   "camlzip" {>= "1.05"}
   "xmlm"
   "cfstream"
-  "ppx_compare"
-  "ppx_sexp_conv"
+  "ppx_compare" {< "v0.12"}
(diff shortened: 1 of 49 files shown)


xclerc:
It looks like most problems are related to `ahrocksdb`,
and are apparently expected to be fixed by https://github.com/ocaml/opam-repository/pull/13347.

packages/biocaml/biocaml.0.9.0/opam
```diff
@@ -29,20 +29,20 @@ depends: [
...
+  "ppx_compare" {< "v0.12"}
+  "ppx_sexp_conv" {< "v0.12"}
   "re"
   "rresult"
   "uri"
 ]
 depopts: [
-  "async"
-  "core"
+  "async" {< "v0.12"}
```
kit-ty-kate: this shouldn't be define like this if I'm right. It has a different semantic than `conflicts: ["async" {>= "v0.12"}]`

The current one becomes a dependency only if `async < v0.12` is installed but doesn't say anything for `async >= v0.12`

xclerc: Indeed, this is an embarrassing mistake; thanks for noticing!

kit-ty-kate: Mmh, the fix you just push is not correct. It's still more or less the same thing.
The constraint needs to be in the `conflicts` block only

xclerc: I was confused, the "fix" was definitely wrong.

avsm:
thanks!

PR #13351, opened 2019-01-31
add conf-rust.2018

c-cube:
see https://github.com/ocaml/opam-repository/pull/13314 for some context. This should test whether a version of rust that supports `--edition 2018` is present.


packages/conf-rust/conf-rust.2018/opam
```diff
@@ -0,0 +1,27 @@
...
+  ["cargo"] {os-distribution = "fedora"}
+  ["cargo"] {os-family = "suse"}
+  ["cargo"] {os-distribution = "debian"}
+  ["cargo"] {os-distribution = "ubuntu"}
+  ["cargo"] {os-distribution = "nixos"}
+  ["cargo"] {os-distribution = "alpine"}
+  ["rust"] {os-distribution = "archlinux"}
+  ["rust"] {os = "macos" & os-distribution = "homebrew"}
+]
+synopsis: "Virtual package relying on cargo (rust build system)"
```
mseri: Don’t you need to add extra-files: `[“test.rs” “md5=...”]`

c-cube:
Well, CI now fails because its rust version is too old… which is exactly the point ^^

mseri:
Given the way the compiler will evolve I think this is the most reasonable and robust approach. Releases have backward compatible support for editions and if a feature is really necessary one releases directly the next edition as an opam package

c-cube:
It might be better to have a separate package (`conf-rust-2018`) so that different libraries can require different versions of rust (say, `conf-rust-2022`) without these versions colliding.

mseri:
Maybe this could be a good idea, but rust.2022 does support also projects using rust.2018 by design, so one will only need the lower bound anyways

mseri:
Adding conf-rust-2018 may be better if we want to reserve the possibility of introducing more specific release versioning in the future, not just to diversify editions. Would that be better for you as well  @kit-ty-kate ?

kit-ty-kate:
> Adding conf-rust-2018 may be better if we want to reserve the possibility of introducing more specific release versioning in the future, not just to diversify editions. Would that be better for you as well @kit-ty-kate ?

I've been thinking about this and I think conf-rust-2018 would be the best solution. It would avoid some of the problem, for example, `conf-llvm` has (picking the latest version of llvm even if it's not available on the host) and we have to resolve manually using `opam pin`.. (I should fix that I think)

mseri:
@kit-ty-kate I agree with your comment. I have tried to ask for comments also on discord, so it may be worth waiting a day before changing this to see if other people come up with different ideas or further changes.

mseri:
Given that there was no additional feedback, @c-cube can you rename the package into `conf-rust-2018.1`. I don’t have access to a computer to di it myself at the moment

kit-ty-kate:
it should be fine. Thanks!

mseri:
Thanks!

PR #13360, opened 2019-02-01
[new release] uri (2.1.0)

avsm:
An RFC3986 URI/URL parsing library

- Project page: <a href="https://github.com/mirage/ocaml-uri">https://github.com/mirage/ocaml-uri</a>
- Documentation: <a href="https://mirage.github.io/ocaml-uri/">https://mirage.github.io/ocaml-uri/</a>

##### CHANGES:

* Expose a `compare` function in `Uri_sexp` so that it will work
  with `deriving compare,sexp`.
* Upgrade the opam metadata to the 2.0 format.
* Update Travis to test OCaml 4.03->4.07.
* Minimum OCaml version is now 4.04.0+ due to sexplib0 dependency.

### DIFF
--- a/packages/uri/uri.2.1.0/opam
+++ b/packages/uri/uri.2.1.0/opam
@@ -29,5 +29,5 @@ build: [
 url {
   src:
     "https://github.com/mirage/ocaml-uri/releases/download/v2.1.0/uri-v2.1.0.tbz"
-  checksum: "md5=36b973b283cce2f3b31b121eccc5c2d1"
+  checksum: "md5=a5d9ed86f9f26d7936523afecd623d61"
 }


avsm:
a mistake in the CHANGES file obliterated the old release by mistake :-) doing a quick rerelease here

pirbo:
Sorry, I don't know where else to write that:
Having changed like that in place the tgz of uri.2.1.0 is an absolute disaster for the tezos blockchain.
Basically, no one can compile tezos anymore!

In tezos, we want to live in a controlled environment so we use a hash of the opam-repository we know works for us to compile our software. Now on, our snapshot contains the old hash of `uri-v2.1.0.tbz` so opam refuses to compile it (as we expect it to)!

kit-ty-kate:
relying on a specific hash of opam-repository directly (which heavily relies on github archives) seems a very bad idea. Github archives break quite frequently and we have to update the hash for some packages from time to time.

I would advice to use a local archive cache using `opam admin cache` for example.

avsm:
@pirbo the problem is that the old archive got overwritten (due to a duplicated entry in CHANGES), so I don't have a copy of the old archive. We could replace it; it's frustrating that github doesn't keep track of binary updates.

How bad a blocker for Tezos is this? Can you upgrade to a newer opam-repo hash or do you want me to hunt down and replace the old archive?

pirbo:
I guested you didn't replaced a release in place "on purpose".
Yes, we'll indeed consider `opam admin cache`, thanks.

I have the old archive in my opam cache. But, updating the tezos-opam-repository to put a new url is as painful as updating it to put new hashes so...

avsm:
> But, updating the tezos-opam-repository to put a new url is as painful as updating it to put new hashes so...

Why this is so difficult? Is that something we can help with?

pirbo:
It's not technically difficult (while I'm struggling now with ci issue but that's our internal nomadic problems...) It is communication wise expensive. Some participants in the network freaks out each time the "released" branch is updated: "Why have you changed? Is it a security issue? What has changed? Should I update? ...."

PR #13357, opened 2019-02-01
[new release] hex (1.3.0)

avsm:
Library providing hexadecimal converters

- Project page: <a href="https://github.com/mirage/ocaml-hex">https://github.com/mirage/ocaml-hex</a>
- Documentation: <a href="https://mirage.github.io/ocaml-hex/">https://mirage.github.io/ocaml-hex/</a>

##### CHANGES:

* Add `to/of_bytes/bigstring` functions. (mirage/ocaml-hex#27 @vbmithr)
* Install toplevel printers automatically in modern utop (@avsm)
* Port from jbuilder to dune (@avsm)
* Use `dune-release` instead of topkg (@avsm)
* Update opam metadata to 2.0 format (@avsm)
* Improve ocamldoc (@avsm)

PR #13362, opened 2019-02-01
tighten up github constraints

avsm:
fixes failures in #13357


avsm:
https://github.com/ocaml/opam-repository/pull/13377 should do the trick instead

PR #13388, opened 2019-02-04
[new release] cohttp-lwt-jsoo, cohttp, cohttp-lwt, cohttp-lwt-unix, cohttp-top, cohttp-async and cohttp-mirage (2.0.0)

avsm:
CoHTTP implementation for the Js_of_ocaml JavaScript compiler

- Project page: <a href="https://github.com/mirage/ocaml-cohttp">https://github.com/mirage/ocaml-cohttp</a>
- Documentation: <a href="https://mirage.github.io/ocaml-cohttp/">https://mirage.github.io/ocaml-cohttp/</a>

##### CHANGES:

Compatibility breaking interface changes:

Async: Expert response action no longer writes empty HTTP body (mirage/ocaml-cohttp#647 by @andreas)

In cohttp.0.99, a number of subpackages were turned into explicit
opam packages to simplify dependency management.
To aid migration, some compatability shims were left in place so that
the old findlib names would continue to work. They have now been removed
as of this release.  If you were still using them, then please rename
them as follows in your dune or ocamlbuild files:
- `cohttp.lwt-core` -> `cohttp-lwt`
- `cohttp.lwt` -> `cohttp-lwt-unix`
- `cohttp.js` -> `cohttp-lwt-jsoo`
- `cohttp.async` -> `cohttp-async`
- `cohttp.top` -> `cohttp-top`

Other changes and bugfixes:
* Lwt, Mirage: Add log warnings for uncaught exceptions (mirage/ocaml-cohttp#592 by @ansiwen)
* Log invalid client input and do not catch out of memory exceptions (mirage/ocaml-cohttp#652 @hannesm)
* Port opam files to opam2 and add local synopsis and descriptions.
* Lwt: Add Expert response action for servers (mirage/ocaml-cohttp#647 by @andreas)
* Use the namespaced `js_of_ocaml` interfaces from 3.3.0 onwards (mirage/ocaml-cohttp#654 @avsm)
* Use Base64 3.1.0 interfaces (mirage/ocaml-cohttp#655 @avsm)
* Clean up redundant conflicts in the `opam` files (@avsm)

PR #13390, opened 2019-02-04
fix aws-s3 upper bound on base64

avsm:
cc @hcarty, found in #13388 revdeps


avsm:
This fixing was masking another one with yojson deprecation, so ill fix that separately

PR #13391, opened 2019-02-04
tighten conduit constraints and fix websocket-async

avsm:
found in revdeps for #13388


packages/conduit-async/conduit-async.1.2.0/opam
```diff
@@ -17,7 +17,7 @@ depends: [
   "core" {< "v0.12"}
   "ppx_sexp_conv" {< "v0.12"}
   "sexplib" {< "v0.12"}
-  "conduit"
+  "conduit" {="1.2.0"}
```
mseri: There is no conduit `1.2.0`. I think this is `1.1.0`

PR #13392, opened 2019-02-05
add upper bounds for calculon-web and tighten cohttp constraints

avsm:
part of revdeps for #13388


kit-ty-kate:
`calculon-web.0.2` fails the same way already. Thanks!

PR #13393, opened 2019-02-05
add missing base64 dep on datakit-ci with upper bound

avsm:
found in revdeps for #13388


avsm:
solver failures in early compilers, merging

PR #13394, opened 2019-02-05
nsq breaks with Lwt_log being separate, so upper bound on lwt

avsm:
revdeps for #13388


avsm:
macos timeout, rest green

PR #13399, opened 2019-02-05
aws-s3.0.9.0: has warnings as errors on, so restrict yojson

avsm:
it now issues deprecation warnings for the json type
cc @hcarty


kit-ty-kate:
Thanks

PR #13416, opened 2019-02-07
added packages/ancient/ancient.0.9.1

UnixJunkie:
This is just the continuation of the ancient package.
However, since ancient is no more maintained, I prefer to rename the opam package and declare
it incompatible with ancient.
So that, people are not automatically switching to this version without knowing it.


UnixJunkie:
all tests are green

avsm:
Just before this is merged, has someone contacted the maintainer Richard Jones?  Since his `ancient` package is already in opam, he may prefer that name to be updated instead of having two versions in opam.

UnixJunkie:
I e-mailed him.
Also, I mostly renamed the package because I don't want people to blame me if newer versions of ancient break software that was using the old version.
People have to switch willingly, and at their own risk, to this version.

UnixJunkie:
do not merge, the mmalloc library was not installed

UnixJunkie:
I'll have to retag the package after the Makefile is fixed

avsm:
I would really prefer if this had a cleaner line of communication between the original and this fork.  Did the author reply to your email?  And won't opam version constraints on the compiler version prevent older software from picking this up anyway?  What has materially changed in the new version that might break packages -- I'm not sure that any existing opam packages actually depend on Ancient at all currently...

UnixJunkie:
I did not get any reply to my two emails to the original author.
Currently, I just corrected compilation warnings and updated the code so that it compiles with recent versions of ocaml.
I.e. there should be no breaking changes.

UnixJunkie:
I updated this PR.
There is my name as the maintainer of the opam package, but I am not sure I am very competent to maintain this package in fact.
I think that the functionality to deep-copy a data structure to a then read-only, non-GCed, shared memory should be provided by the stdlib's GC.

UnixJunkie:
all tests are green

mseri:
Ping @avsm before doing anything

packages/ancient/ancient.0.9.1/opam
```diff
@@ -0,0 +1,39 @@
...
+maintainer: "unixjunkie@sdf.org"
+build: [
+  ["sh" "-c" "cd mmalloc && ./configure"]
+  [make "depend"]
+  [make]
+]
+remove: ["ocamlfind" "remove" "ancient"]
+depends: [
+  "ocaml" {>= "4.03.0"}
+  "ocamlfind"
```
kit-ty-kate:
```suggestion
  "ocamlfind" {build}
```
do you need ocamlfind as a runtime dependency? I couldn't find any use in the source code

kit-ty-kate:
Thanks a lot!

PR #13420, opened 2019-02-07
3 packages from OCamlPro/ocplib-json-typed at 0.7

chambart:
This pull-request concerns:
-`ocplib-json-typed.0.7`: Type-aware JSON and JSON schema utilities
-`ocplib-json-typed-browser.0.7`: Json_repr interface over JavaScript's objects
-`ocplib-json-typed-bson.0.7`: A Json_repr compatible implementation of the JSON compatible subset of BSON

---
* Homepage: https://github.com/ocamlpro/ocplib-json-typed
* Source repo: git+https://github.com/ocamlpro/ocplib-json-typed.git
* Bug tracker: https://github.com/ocamlpro/ocplib-json-typed/issues

---

### DIFF
--- a/packages/ocplib-json-typed-browser/ocplib-json-typed-browser.0.7/opam
+++ b/packages/ocplib-json-typed-browser/ocplib-json-typed-browser.0.7/opam
@@ -1,0 +1,24 @@
+opam-version: "2.0"
+synopsis: "Json_repr interface over JavaScript's objects"
+maintainer: "Benjamin Canou <benjamin@ocamlpro.com>"
+authors: "Benjamin Canou <benjamin@ocamlpro.com>"
+homepage: "https://github.com/ocamlpro/ocplib-json-typed"
+bug-reports: "https://github.com/ocamlpro/ocplib-json-typed/issues"
+license: "LGPLv3 w/ linking exception"
+dev-repo: "git+https://github.com/ocamlpro/ocplib-json-typed.git"
+
+build: [ "dune" "build" "-j" jobs "-p" name "@install" ]
+run-test: [ "dune" "runtest" "-p" name "-j" jobs ]
+depends: [
+  "ocaml" {>= "4.03.0"}
+  "dune" {build & >= "1.0.1"}
+  "ocplib-json-typed" {= "0.7" }
+  "js_of_ocaml" {>= "3.3.0"}
+]
+url {
+  src: "https://github.com/OCamlPro/ocplib-json-typed/archive/v0.7.tar.gz"
+  checksum: [
+    "md5=afa5aed32fcd7ff3ffae7c0adfb5de58"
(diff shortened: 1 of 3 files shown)


packages/ocplib-json-typed-browser/ocplib-json-typed-browser.0.7/opam
```diff
@@ -0,0 +1,24 @@
+opam-version: "2.0"
+synopsis: "Json_repr interface over JavaScript's objects"
+maintainer: "Benjamin Canou <benjamin@ocamlpro.com>"
+authors: "Benjamin Canou <benjamin@ocamlpro.com>"
+homepage: "https://github.com/ocamlpro/ocplib-json-typed"
+bug-reports: "https://github.com/ocamlpro/ocplib-json-typed/issues"
+license: "LGPLv3 w/ linking exception"
+dev-repo: "git+https://github.com/ocamlpro/ocplib-json-typed.git"
+
+build: [ "dune" "build" "-j" jobs "-p" name "@install" ]
```
kit-ty-kate: the `@install` is unnecessary but for now it's fine as is

packages/ocplib-json-typed-bson/ocplib-json-typed-bson.0.7/opam
```diff
@@ -0,0 +1,24 @@
...
+synopsis: "A Json_repr compatible implementation of the JSON compatible subset of BSON"
+maintainer: "Benjamin Canou <benjamin@ocamlpro.com>"
+authors: "Benjamin Canou <benjamin@ocamlpro.com>"
+homepage: "https://github.com/ocamlpro/ocplib-json-typed"
+bug-reports: "https://github.com/ocamlpro/ocplib-json-typed/issues"
+license: "LGPLv3 w/ linking exception"
+dev-repo: "git+https://github.com/ocamlpro/ocplib-json-typed.git"
+
+build: [ "dune" "build" "-j" jobs "-p" name "@install" ]
+run-test: [ "dune" "runtest" "-p" name "-j" jobs ]
```
avsm: this should be in the `build` stanza but annotated with `{with-test}` now

kit-ty-kate: mmh, both does the same thing as far as I am aware

avsm: then why do we have both... :-)

kit-ty-kate: I think the proper way is using `run-test` which runs the commands during the `test phase`. If you annotate with `{with-test}` it's done during the `build phase`. I think the annotation is primarily here to enable tests flags or something like that during configuration or compilation.

I'm not sure what's the difference between the `build phase` and the `test phase` (maybe there can be some differences from the bubblewrap point of view) but they are morally the same in the end, just one usage is more a distorted use from its original use I suppose.

cc @AltGr @rjbou

chambart:
Indeed it doesn't

kit-ty-kate:
Fair enough. Thanks

PR #13424, opened 2019-02-08
ocsigen-start.1.1.0: correct wrong md5 checksum

jrochel:
I'm not sure whether what I'm doing is considered bad practise, but
it seems that version does not compile anymore. As far as I can see,
there is no-one willing to maintain a pretty outdated version of that
package, nor anyone depending on it.


jrochel:
see: https://ci.ocamllabs.io/log/saved/docker-run-f6fb11d4a370d9eb0574085f69717bdc/041f07bbba3b08f040463a4439eac37b0f106707

Leonidas-from-XIV:
Wouldn't it be enough to just fix the checksum of the tarball?

kit-ty-kate:
Yes, it works! Thanks!

PR #13427, opened 2019-02-08
graphql, graphql_parser, graphql-lwt, graphql-async, graphql-cohttp 0.9.0

andreas:
# 0.9.0 2019-02-08
Skip and include directives (andreas/ocaml-graphql-server#117)
Expose more data to resolvers (andreas/ocaml-graphql-server#127)
Add the errors key first in the response JSON (andreas/ocaml-graphql-server#131)
Rewrite parser to Menhir and replace sexp with fmt (andreas/ocaml-graphql-server#132)
Support for websockets as transport (andreas/ocaml-graphql-server#133)

### DIFF
--- a/packages/graphql-async/graphql-async.0.9.0/opam
+++ b/packages/graphql-async/graphql-async.0.9.0/opam
@@ -14,7 +14,7 @@ build: [
 depends: [
   "ocaml" {>= "4.03.0"}
   "dune" {build}
-  "graphql" {= "0.8.0"}
+  "graphql" {>= "0.9.0"}
   "async_kernel" {>= "v0.9.0" & < "v0.12"}
   "alcotest" {with-test}
   "async_unix" {with-test & >= "v0.9.0" & < "v0.12"}
@@ -26,6 +26,6 @@ description: """
 `graphql-async` adds support for Async to `graphql`, so you can use Async in your GraphQL schema resolver functions."""
 
 url {
-archive: "https://github.com/andreas/ocaml-graphql-server/releases/download/0.8.0/graphql-0.8.0.tbz"
-checksum: "582564e3de2095ce7512cee28b406920"
+archive: "https://github.com/andreas/ocaml-graphql-server/releases/download/0.9.0/graphql-0.9.0.tbz"
+checksum: "2385ce2f537413e9466d13acb11ae7e4"
 }
--- a/packages/graphql-cohttp/graphql-cohttp.0.9.0/opam
+++ b/packages/graphql-cohttp/graphql-cohttp.0.9.0/opam
@@ -14,18 +14,21 @@ build: [
 depends: [
   "ocaml" {>= "4.03.0"}
   "dune" {build}
-  "graphql" {= "0.8.0"}
-  "async_kernel" {>= "v0.9.0" & < "v0.12"}
-  "alcotest" {with-test}
-  "async_unix" {with-test & >= "v0.9.0" & < "v0.12"}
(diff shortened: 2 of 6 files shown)


packages/graphql-async/graphql-async.0.9.0/opam
```diff
@@ -0,0 +1,31 @@
...
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+]
+
+depends: [
+  "ocaml" {>= "4.03.0"}
+  "dune" {build}
+  "graphql" {>= "0.9.0"}
+  "async_kernel" {>= "v0.9.0"}
+  "alcotest" {with-test}
+  "async_unix" {with-test & >= "v0.9.0"}
```
andreas: It's unclear to me whether I should keep the upper bound constraint on `async_kernel` and `async_unix` which was added to `graphql-async.0.8.0` in [this PR](https://github.com/ocaml/opam-repository/pull/13344).

mseri: Ping @xclerc. I would probably keep it and relax it after the release of the new janestreet libraries

xclerc: Indeed, that is the plan.

andreas: Ok, I’ll add the constraint later today.

andreas: I've added the constraint now.

andreas:
Travis due to a timeout error, but I don't have permissions to restart the build ([link](https://travis-ci.org/ocaml/opam-repository/jobs/492243856)):

```
The job exceeded the maximum time limit for jobs, and has been terminated.
```

packages/graphql-cohttp/graphql-cohttp.0.9.0/opam
```diff
@@ -0,0 +1,34 @@
...
+dev-repo: "git+https://github.com/andreas/ocaml-graphql-server.git"
+
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+]
+
+depends: [
+  "ocaml" {>= "4.03.0"}
+  "jbuilder" {build}
```
hannesm: shouldn't this be `dune {buld}`?

andreas: Indeed it should -- thanks!

packages/graphql_parser/graphql_parser.0.9.0/opam
```diff
@@ -0,0 +1,28 @@
...
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+]
+
+depends: [
+  "ocaml" {>= "4.03.0"}
+  "dune" {build}
+  "menhir" {build}
+  "alcotest" {with-test & >= "0.8.1"}
+  "result"
```
hannesm: the dependency on `result` should not be needed, since you require `ocaml >=4.03.0` anyways (and result is a backwards compatibility package for ocaml <4.03.0). NB: this may require changes to code / build system, so no need to modify it for this release, but you may want to drop it in future releases.

hannesm:
thanks, merging. before the last force-push the CI worked, apart from running in a timeout on travis, and the last force-push only changed `jbuilder` -> `dune`.

PR #13434, opened 2019-02-10
New clangml release: 4.0.0beta1

thierry-martinez:
First public release: complete rewriting of the library.
Bindings now call directly the libclang API by a C/OCaml interface.
Compatible with all versions of clang from 3.8.0 to 8.0.0(rc1).
**Edit:** Compatible with all versions of clang from 3.4 to 8.0.0(rc1). (Added compatibility for old versions to be able to compile on CentOS)

### DIFF
--- a/packages/clangml/clangml.4.0.0beta1/opam
+++ b/packages/clangml/clangml.4.0.0beta1/opam
@@ -17,8 +17,8 @@ depends: [
   "ppx_deriving"
   "visitors"
   "dune" {>= "1.2"}
-  "ocaml" {>= "4.02.0" & < "4.08.0"}]
+  "ocaml" {>= "4.04.0" & < "4.08.0"}]
 url {
   src: "https://gitlab.inria.fr/tmartine/clangml/-/archive/v4.0.0beta1/clangml-v4.0.0beta1.tar.gz"
-  checksum: "md5=47630f9fc99676f85e5b940cbea27b61"
+  checksum: "md5=d6e62fadc5b0eac266cbb35c3ac898ed"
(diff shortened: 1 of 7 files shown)


thierry-martinez:
centos failure is due to the very old version of llvm shipped with the distribution: I don't plan to support llvm 3.4.2, at least for this beta release.

I don't understand travis failure: could we have a look to config.log to know why configure does not manage to compile C executable?

thierry-martinez:
Travis CI spots a failure on the installation of `deriving` (`ocamlfind: Package 'camlp4.quotations.o' not found`), which does no seem to be due to `clangml`.

As usual, the opam-lint error 52 looks like the lint bug reported here [ocaml/opam#3625](https://github.com/ocaml/opam/issues/3625) since `conf: flag` is present.

packages/clangml/clangml.4.0.0beta1/opam
```diff
@@ -0,0 +1,25 @@
...
+  ["./configure" "--prefix=%{prefix}%"]
+  ["dune" "build" "-p" name "-j" jobs]]
+depends: [
+  "stdcompat"
+  "conf-libclang"
+  "conf-ncurses"
+  "conf-zlib"
+  "ppx_deriving"
+  "visitors"
+  "dune" {>= "1.2"}
```
kit-ty-kate:
```suggestion
  "dune" {build & >= "1.2"}
```
this should probably be flagged as a build dependency

kit-ty-kate:
Looks ok. Thanks!

PR #13438, opened 2019-02-11
win-error.1.0: remove bad `version` field from metadata

djs55:
See [mirage/ocaml-win-error#6]

### DIFF
--- a/packages/win-error/win-error.1.0/opam
+++ b/packages/win-error/win-error.1.0/opam
@@ -1,6 +1,5 @@
 opam-version: "2.0"
 name: "win-error"
-version: "0.2"
 maintainer: "dave@recoil.org"
 authors: "David Scott"
 license: "ISC"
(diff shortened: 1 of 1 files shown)


packages/win-error/win-error.1.0/opam
```diff
@@ -1,6 +1,5 @@
 opam-version: "2.0"
 name: "win-error"
```
avsm:
```suggestion
```

avsm:
did you release it via `dune-release` originally? seems like a bug

djs55:
Hm, I can't remember if I used `dune-release` for this one specifically. I've been trying to use it more consistently but I might not have had it on the laptop I was using.

avsm:
@kit-ty-kate it might be worth putting in an `opam admin check` into check.ocamllabs.io -- I think we still need https://github.com/ocaml/opam/issues/3342 to make the output easier to parse

kit-ty-kate:
> @kit-ty-kate it might be worth putting in an `opam admin check` into check.ocamllabs.io -- I think we still need [ocaml/opam#3342](https://github.com/ocaml/opam/issues/3342) to make the output easier to parse

is `opam admin check` morally the same thing as `opam update --strict`? (at least for this kind of error)

kit-ty-kate:
Thanks!

PR #13440, opened 2019-02-11
why3 1.2.0

silene:


silene:
The only failure is the macOS job, due a timeout, but the job nonetheless succeeded in building all three packages (it reached the timeout while building the base package the third time). So, everything looks good.

packages/why3-coq/why3-coq.1.2.0/opam
```diff
@@ -0,0 +1,65 @@
...
+tags: [
+  "deductive"
+  "program verification"
+  "formal specification"
+  "automated theorem prover"
+  "interactive theorem prover"
+]
+
+build: [
+  ["./autogen.sh"] {dev} # when pinning, there might be no configure file
```
kit-ty-kate:
```suggestion
  ["./autogen.sh"] {pinned}
```
`pinned` should be used here instead

packages/why3-ide/why3-ide.1.2.0/opam
```diff
@@ -0,0 +1,68 @@
...
+tags: [
+  "deductive"
+  "program verification"
+  "formal specification"
+  "automated theorem prover"
+  "interactive theorem prover"
+]
+
+build: [
+  ["./autogen.sh"] {dev} # when pinning, there might be no configure file
```
kit-ty-kate:
```suggestion
  ["./autogen.sh"] {pinned} # when pinning, there might be no configure file
```

packages/why3/why3.1.2.0/opam
```diff
@@ -0,0 +1,79 @@
...
+tags: [
+  "deductive"
+  "program verification"
+  "formal specification"
+  "automated theorem prover"
+  "interactive theorem prover"
+]
+
+build: [
+  ["./autogen.sh"] {dev} # when pinning, there might be no configure file
```
kit-ty-kate:
```suggestion
  ["./autogen.sh"] {pinned}
```

kit-ty-kate:
ah sorry I tried using the "review" feature but it doesn't seem to work as I wanted. Anyway, `pinned` should be used and you can leave the comment if you want

silene:
The `autogen.sh` command has to be run only when using the git repository as a source, not when pinning a released version. So, if I understand opam's documentation correctly, it seems like `{dev}` is the proper trigger ("true if this is a development package, i.e. it was not built from a release archive"). Is that right? (In which case, I should make the comment clearer.)

kit-ty-kate:
mmh that's true. Now I'm wondering if `{pinned}` is the right flag for all the other packages which for example have it for `dune subst`. @AltGr @rjbou ?

In the meantime it's fine I suppose. Thanks and sorry for the delay

AltGr:
There isn't much difference, but the scope is slightly different. I believe `dev` is indeed better here:
- `dev` can apply to non-pinned repository packages when they are bound to a VC address (which shouldn't happen in `opam-repository`, but is allowed for custom ones)
- `pinned` can apply to packages that are just pinned to a specific version, without diverting from their repository definition or source archive. This might be more of a problem here.

This could probably be added to the FAQ ?

PR #13443, opened 2019-02-11
Package disml.0.2.4

Mishio595:
### `disml.0.2.4`
An OCaml library for interfacing with the Discord API
Dis.ml is a library that provides a high-level interface to the Discord API.
Key features include:
* Automatic sharding
* Deserialization of Discord objects to record types with related helper methods
* Automatic rate-limiting

For examples, see `/bin` in the git repo.

---
* Homepage: https://gitlab.com/Mishio595/disml
* Source repo: git+https://gitlab.com/Mishio595/disml
* Bug tracker: https://gitlab.com/Mishio595/disml/issues

---

### DIFF
--- a/packages/disml/disml.0.2.4/opam
+++ b/packages/disml/disml.0.2.4/opam
@@ -1,0 +1,44 @@
+opam-version: "2.0"
+maintainer: "Adelyn Breedlove <tenseiken595@gmail.com>"
+authors: "Adelyn Breedlove <tenseiken595@gmail.com>"
+license: "MIT"
+homepage: "https://gitlab.com/Mishio595/disml"
+doc: "https://mishio595.gitlab.io/disml/"
+dev-repo: "git+https://gitlab.com/Mishio595/disml"
+bug-reports: "https://gitlab.com/Mishio595/disml/issues"
+tags: ["discord"]
+synopsis: "An OCaml library for interfacing with the Discord API"
+description: """
+Dis.ml is a library that provides a high-level interface to the Discord API.
+Key features include:
+* Automatic sharding
+* Deserialization of Discord objects to record types with related helper methods
+* Automatic rate-limiting
+
+For examples, see `/bin` in the git repo.
+"""
+depends: [
+  "ocaml" {>= "4.07.0"}
+  "dune" {build & >= "1.3.0"}
+  "async_ssl" {>= "v0.11.0"}
+  "core" {>= "v0.11.3"}
+  "cohttp-async" {>= "1.2.0"}
+  "ppx_sexp_conv" {>= "v0.11.2"}
+  "ppx_deriving_yojson" {>= "3.1"}
+  "websocket-async" {>= "2.12"}
+  "decompress"
+  "odoc" {with-doc & >= "1.3.0"}
+  "yojson" {>= "1.6.0"}
+]
+build: [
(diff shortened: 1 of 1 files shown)


packages/disml/disml.0.2.4/opam
```diff
@@ -0,0 +1,44 @@
...
+Dis.ml is a library that provides a high-level interface to the Discord API.
+Key features include:
+* Automatic sharding
+* Deserialization of Discord objects to record types with related helper methods
+* Automatic rate-limiting
+
+For examples, see `/bin` in the git repo.
+"""
+depends: [
+  "ocaml" {>= "4.07.0"}
```
kit-ty-kate: is there a reason why you rely on such a recent version of OCaml?

packages/disml/disml.0.2.4/opam
```diff
@@ -0,0 +1,44 @@
...
+  "ppx_deriving_yojson" {>= "3.1"}
+  "websocket-async" {>= "2.12"}
+  "decompress"
+  "odoc" {with-doc & >= "1.3.0"}
+  "yojson" {>= "1.6.0"}
+]
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "build" "@doc" "-j" jobs] {with-doc}
```
kit-ty-kate: this line probably also needs `"-p" name` like the previous one

packages/disml/disml.0.2.4/opam
```diff
@@ -0,0 +1,44 @@
...
+depends: [
+  "ocaml" {>= "4.07.0"}
+  "dune" {build & >= "1.3.0"}
+  "async_ssl" {>= "v0.11.0"}
+  "core" {>= "v0.11.3"}
+  "cohttp-async" {>= "1.2.0"}
+  "ppx_sexp_conv" {>= "v0.11.2"}
+  "ppx_deriving_yojson" {>= "3.1"}
+  "websocket-async" {>= "2.12"}
+  "decompress"
```
kit-ty-kate: is there no lower-bound constraints for decompress. The API changed a lot between versions and I'm worried it wouldn't compile with earlier versions. Have you tried earlier versions?

kit-ty-kate:
something is wrong with the dependency constraints but I'm not sure why. Pinning the package directly from the repository seems to work locally but the CI says otherwise..

packages/disml/disml.0.2.4/opam
Mishio595: I haven't tested with older compiler versions, testing how far back this is compatible currently.

packages/disml/disml.0.2.4/opam
Mishio595: Wasn't aware that was needed, still fairly new to opam and dune. I'll add that in, thanks.

Mishio595:
I think I see the problem. Due to the changes in `yojson.1.6.0` I had updated to using the `t` types instead of `json`, but the latest release of `ppx_deriving_yojson` does not yet have this merged. I had been locally pinned to the dev repo for the ppx.

I'll make some changes to require yojson < 1.6.0 so the ppx is happy until we get a new release from them.

Mishio595:
Since this would require a new release on my end to revert, should I go ahead and close this PR and open a new one for the subsequent release?

PR #13449, opened 2019-02-13
Prevent yojson 1.4.x from building with dune > 1.7

Leonidas-from-XIV:
The deprecations in dune 1.7 cause failures to build in yojson 1.4 (though not older versions, 1.3.0 works fine). It also works with newer versions, like 1.5 or 1.6.

Here's a Dockerfile + output which reproduces the issue: https://gist.github.com/Leonidas-from-XIV/ad89a093ded4835f96707ec382bee96f


mseri:
Looks good, thanks

PR #13475, opened 2019-02-16
Add ocaml-integers 0.3.0

yallop:
See the [Changelog](https://github.com/ocamllabs/ocaml-integers/blob/0.3.0/CHANGES.md) for details.

### DIFF
--- a/packages/ctypes/ctypes.0.12.0/opam
+++ b/packages/ctypes/ctypes.0.12.0/opam
@@ -23,7 +23,7 @@ remove: [
 depends: [
   "ocaml" {>= "4.01.0"}
   "base-bytes"
-  "integers"
+  "integers" { < "0.3.0" }
   "ocamlfind" {build}
   "conf-pkg-config" {build}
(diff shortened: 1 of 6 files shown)


kit-ty-kate:
Thank you

kit-ty-kate:
apparently there was a lot more package that were using `integers` without being explicit in their dependencies... A lot of packages broke.. oh well.

kit-ty-kate:
I don't understand why the CI didn't get the failure on `lz4` for example.
Locally I get `Error: Error on dynamically loaded library: dllintegers.so: dllintegers.so: cannot open shared object file: No such file or directory` when I try to compile it with `integers.0.3.0`.

I get similar results with `yaml.1.0.0` and `nocrypto.0.5.4-1`. Any idea?

kit-ty-kate:
It seems that `dllintegers.so` is not installed anymore in this version.
I'm still wondering how it passed the CI. I'm reverting this PR now for us to find the cause of this issue

ping @yallop

kit-ty-kate:
ok nevermind something messed up with a local repository

PR #13481, opened 2019-02-18
added packages/bst/bst.6.1.0

UnixJunkie:
there is a new feature: you can partition the bisector tree
using a query point and a distance/tolerance to that point.
This is closely related to the already existing 'neighbors' query.


UnixJunkie:
All CI tests are green, amazing!

mseri:
Thanks!

PR #13495, opened 2019-02-19
+mirage-platform.3.3.0

avsm:
supports forthcoming dune support as part of mirage/mirage#969:

* Add support for OCaml 4.08 (mirage/mirage-platform#206 by @anmonteiro)
* `libxenasmrun.a` is symlinked in the ocaml directory in order to be able to use ocamlopt's `-runtime-variant` option.
* Expose flags through files as well as pkg-config (mirage/mirage-platform#205 by @TheLortex).  Instead of pkg-config, one can use the following files to get compilation flags:
  * mirage-xen-ocaml/libs
  * mirage-xen-ocaml/cflags
  * mirage-xen-posix/minios-cflags
  * mirage-xen-posix/minios-libs
  * mirage-xen-posix/posix-cflags
  * mirage-xen-posix/posix-libs
  With dune this allows us to write `%{lib:mirage-xen-posix:posix-libs}` to get the flags instead of having a script invoking `pkg-config mirage-xen-posix --libs`.

### DIFF
--- a/packages/mirage-xen-ocaml/mirage-xen-ocaml.3.3.0/opam
+++ b/packages/mirage-xen-ocaml/mirage-xen-ocaml.3.3.0/opam
@@ -1,0 +1,29 @@
+opam-version: "2.0"
+maintainer: "anil@recoil.org"
+authors: "The MirageOS team"
+tags: "org:mirage"
+homepage: "https://github.com/mirage/mirage-platform"
+bug-reports: "https://github.com/mirage/mirage-platform/issues/"
+depends: [
+  "ocaml" {>= "4.04.2" & < "4.09.0"}
+  "mirage-xen-posix" {>= "2.6.0"}
+  "conf-pkg-config"
+  "ocamlfind" {build}
+  "ocaml-src"
+]
+substs: [
+  "xen-ocaml/flags/cflags.tmp"
+  "xen-ocaml/flags/libs.tmp"
+  ]
+available: os = "linux"
+build: [make "xen-ocaml-build"]
+install: [make "xen-ocaml-install" "PREFIX=%{prefix}%"]
+remove: [make "xen-ocaml-uninstall" "PREFIX=%{prefix}%"]
+dev-repo: "git+https://github.com/mirage/mirage-platform.git"
+synopsis: "MirageOS headers for the OCaml runtime"
+description:
+  "The package contains the OCaml runtime patches and build system."
+url {
+  src: "https://github.com/mirage/mirage-platform/archive/v3.3.0.tar.gz"
+  checksum: "md5=25b42ec49cdd2c4503eb994d0279ef35"
+}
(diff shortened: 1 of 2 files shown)


mseri:
The failure with the system switch is expected?

avsm:
Hm, no - will respin release

avsm:
Hm, after some discussion with @TheLortex, I think we need to restrict this package to non-system installations. That won't disrupt existing users of mirage-platform, but will unblock the dune release of Mirage.  @samoht @hannesm are you ok with this?  The reason is that we need to symlink into the ocaml libdir in order to support `-output-complete-obj` for the mirage runtime variants, and you cant do that on a system switch.

hannesm:
@avsm is there at some point in the future an idea how to support system switches? or will the `-output-complete-obj` approach only ever work with non-system switches?

is a system switch detectable from within an opam file - and is it possible to print a message "you're using a system switch that is currently not supported"? (or inform the user "in order to use a system switch, you'll have to manually `ln -s YYY /usr/local/lib/ocaml/runtime/ZZZ`?)

I'm slightly concerned to introduce behaviour that only works on non-system opam switches, and would be in favour to figure out how system switches can be supported.

TheLortex:
It could be supported by patching OCaml so that the -runtime-variant option is more flexible: if we're able to provide a path to `libasmrun.a` then everything could work fine even on system switches.

hannesm:
IIUC, runtime-variant is located in the default search path (`Asmlink.runtime_lib` uses `Load_path.find`, which uses the path initialized by `Compmisc.init_path`, using `Clflags.std_include_dir`), which is `Config.standard_library`, and can be set by the environment variable `OCAMLLIB`. Maybe not doing a symlink, but instead setting this environment variable when compiling/linking is sufficient (sorry, haven't tested this, only tried to read the code in the OCaml compiler).

TheLortex:
It looks like `OCAMLLIB` allows only one single path, so that means we have to import the whole `/usr/lib/ocaml/` content inside a directory that opam can write into (like `.../ocaml-system/lib/ocaml`).

hannesm:
@TheLortex true. a plan!?:
- (a) figuring out in the opam file whether we're a system switch or not (and print a message which asks the user do the symlink manually) - once this is in place, we can merge this PR imho
- (b) ask OCaml distribution whether we can have a more mature `-runtime-variant` option (likely in 4.09 then!?)

hannesm:
I suggest a conflict of `mirage-platform=3.3.0` with `ocaml-system` would be feasible to move forward. WDYT?

TheLortex:
Wouldn't an `ocaml-system` conflict cause trouble with CI ?
I'll make a patch to print a warning message if it's not possible to do the symlink.

hannesm:
why should it cause trouble with CI? it will lead to not being installable if a system switch is used, which is the currently proposed support level (warning messages have the downside that people will ignore them). there is a variety of CI systems around, some use system switches, but most use opam switches, so they should be fine :)

TheLortex:
Okay that's fine with me !

avsm:
ive pushed the ocaml-system conflict

kit-ty-kate:
Thanks

hannesm:
@avsm fine, the conflict should be synchronised at upstream

hannesm:
we discussed this a bit further in marrakesh, and it turns out we can use `-I` for including paths, and don't need any symlink. imho we should revise this 3.3.0 release to deal with that (and then remove the system compiler conflict, and remove the symlink). see also https://github.com/ocaml/ocaml/issues/7926 https://github.com/ocaml/ocaml/pull/2309

PR #13499, opened 2019-02-20
Fix OCaml constraints on old versions of alcotest

samoht:
Detected via http://check.ocamllabs.io/4.03.0/bad/alcotest.0.4.6 and http://check.ocamllabs.io/4.03.0/bad/alcotest.0.4.7


mseri:
Thanks

PR #13502, opened 2019-02-21
Fix pkcs11 packages dependency over dune

NathanReb:
Dune was registered as a regular dependency instead of a build dependency.


mseri:
Thanks

PR #13514, opened 2019-02-24
base-unsafe-string

gasche:
We can use a dependency on the base-unsafe-string package to mark
packages that fail to build when -safe-string is enabled. This is
easier to do, and clearer in the metadata, than remember which version
of OCaml made -safe-string the default (4.06).

Note: this package is currently *not* available under >=4.06 switches
that explicitly enable -unsafe-string (they should in theory support
this flag), as I'm not sure how to do this and whether it is worth the
effort.


mseri:
Oh, nice idea

kit-ty-kate:
I like the idea!! Thanks a lot!

PR #13524, opened 2019-02-24
Migrate all forge(-static).ocamlcore.org links.

gildor478:
The forge will stop serving files as per [this article]. This change
migrates the links from the forge to https://download.ocamlcore.org.

[this article]: https://forge.ocamlcore.org/forum/forum.php?forum_id=966

It also removes the packages that are not available anymore. They should
not be served by the forge, since matching projects are not public
anymore:
- ocaml-rdf is now hosted there https://www.good-eris.net/ocaml-rdf
- sexplib is now hosted there https://github.com/janestreet/sexplib


kit-ty-kate:
If `rdf.0.{2,3,4,5}` and `sexplib.7.0.5` archives are still present in the cache or something, we can retrieve them and host them somewhere else, thus avoiding to remove them.

cc @zoggy @xclerc

gildor478:
I have done my work and I have already checked ;-)

- sexplib 7.0.5 is not available anymore, since it was replaced by the version from janestreet and is now marked as private on forge.o.o. If @mmottl wants me to continue to publish on download.ocamlcore.org, I am fine with that (will literally takes me 5 minutes to do so)

- ocaml-rdf 0.2 - 0.5 can be downloaded from https://www.good-eris.net/ocaml-rdf/download.html BUT it doesn't have the same hash -- that the reason I was reluctant to add it to this change. @zoggy to decide if he wants us to use the one from this other website.

My take on both cases is that they are both 4+ years old and it is probably not worth maintaining them -- but that's IMHO.

avsm:
Thanks very much for this Sylvain! In the interests of a quick merge of the majority of the changes  (since it is destroying opam ci with the number of packages), would it be possible to have a version of this without the rdf changes and sexplib removal?  That can be merged immediately.

For rdf, I think it's safe to use the ones from the other distsite -- it looks like they are compressed by framagit and so a different checksum.  This way genet.0.3 will also remain installable

It would be excellent if sexplib 7.0.5 could be preserved as well, but again, as a separate PR would be great so we can merge the bulk of the changes.

mmottl:
@gildor478 Thanks for your work, I don't have any strong opinions on what to do with old Forge packages that were maintained by me or Jane Street.  Sexplib is now fully maintained by Jane Street anyway.  I don't think many (if any) people would notice if they got removed, since there are much newer versions hosted on GitHub.

gildor478:
@avsm Done for adding back rdf and sexplib. See #13535 for RDF move to github.com -- it looks like they have the same MD5 after all. 

Given the answer of @mmottl, I'll be in favor of removing sexplib 7.0.5. Let me know if you agree and I'll send another PR.

mmottl:
@gildor478 Removing sexplib 7.0.5 is fine with me!

mseri:
I’ll merge his one in the meanwhile. Thanks

avsm:
@mmotl it's a pity to remove it from a historical perspective, since sexplib.7.0.5 was such a trooper of a version with ocaml 3.12.1 for so long -- which still installs fine through opam :-)

avsm:
Many thanks for taking the time to proactively fixing these links in opam-repo, @gildor478!

mmottl:
@avsm Thanks, I guess all things eventually come to an end.  sexplib.7.0.5 was still based on `camlp4` and certainly benefited from the transition to `ppx`.

gildor478:
FYI, tracking number of downloads:
https://forge-static.ocamlcore.org/metrics-using-old-download.html

PR #13497, opened 2019-02-20
constrain existing packages to mirage-kv{-lwt} <2.0.0 for an upcoming…

hannesm:
… breaking release (https://github.com/mirage/mirage-kv/pull/14)

PR #13525, opened 2019-02-24
[new release] mirage-kv and mirage-kv-lwt (2.0.0)

hannesm:
CHANGES:

* Major revision of the `RO` signature:
 - values are of type `string`
 - keys are segments instead of a string
 - `read` is now named `get`, and does no longer take an offset and length
 - the new function `list` is provided
 - the new functions `last_modified` and `digest` are provided
* A module `Key` is provided with convenience functions to build keys
* An `RW` signature is provided, extending `RO` with
 - a function `set` to replace a value
 - a function `remove` to remove a key
 - `batch` to batch operations

upper bounds for existing packages were added in #13497


mseri:
Thanks

PR #13528, opened 2019-02-25
[new release] lablgtk3-sourceview3, lablgtk3-gtkspell3 and lablgtk3 (3.0.beta5)

garrigue:
OCaml interface to GTK+ gtksourceview library

- Project page: <a href="https://github.com/garrigue/lablgtk">https://github.com/garrigue/lablgtk</a>
- Documentation: <a href="https://garrigue.github.io/lablgtk/lablgtk-sourceview">https://garrigue.github.io/lablgtk/</a>

##### CHANGES:

2019.02.12 [Emilio]
  * Update OPAM files so they work better with dune-release


ejgallego:
@garrigue did thinks work fine this time?

garrigue:
This seems to work (waiting for the CI), but I wonder why it adds a .tbz file in addition to the .tar.gz file GitHub automatically generates. Is there a way to customize its contents?

garrigue:
Also, the documentation link in the PR is broken: it point to lablgtk3-sourceview3, not lablgtk3 itself.
By the way, the automatically generated documentation is weaker than the ocamldoc one we were generating before. It lacks links to the Gtk site. However, I'm not the one who did that configuration.

ejgallego:
> This seems to work (waiting for the CI), but I wonder why it adds a .tbz file in addition to the .tar.gz file GitHub automatically generates. Is there a way to customize its contents?

I'd guess the reason is that it also supports the notion of releasing over a regular website, it thus builds it's own tarball and uploads.

> Also, the documentation link in the PR is broken: it point to lablgtk3-sourceview3, not lablgtk3 itself.

Should be fixed, let's open an issue so we don't forget.

> By the way, the automatically generated documentation is weaker than the ocamldoc one we were generating before. It lacks links to the Gtk site. However, I'm not the one who did that configuration.

Good point, I see the `gtkdoc` utility; indeed this can also be fixed IMO, so let's open an issue.

garrigue:
Somehow, the CI got stuck, so trying again. I'm adding proper depexts for gtkspell3 this time.

garrigue:
Is the CI working at all? I would like to see the result on All Distros. The previous run didn't even go that far, probably caused by internal failure (unrelated to the packages).

mseri:
The CI is currently kept busy behind a PR that updated 250something opam package urls :(

ejgallego:
why3-ide has been indeed been broken since beta4 I think [wrong dependencies on gtksourceview on the opam / ocamlfind level]

ejgallego:
Regarding gtkspell we should likely relax the dependency to `>= 3.0.3`

garrigue:
Are you sure this is the problem? The error message on centos looks more like what we had with dune 1.7.0.

Or can we say that we're done? gtkspell3 is not an essential functionality, it already works on most systems, and all the rest works.

ejgallego:
> Are you sure this is the problem? The error message on centos looks more like what we had with dune 1.7.0.

It looks like it:
```
Package gtkspell3-devel.x86_64 0:3.0.3-4.el7 will be installed
...
configure error: gtkspell3-3.0 >= 3.0.4 not found.
```
so as you wish, but the version check is too strong.

ejgallego:
> The error message on centos looks more like what we had with dune 1.7.0.

The message is the same because the Dune bug made it think that libraries were not installed [when they really were]

garrigue:
Ok, I see.
IIRC, there is a pkg-config flag that gives a more explicit error message (see the conf packages).
The trouble is that to change the version retirement I need to make a new release, since it's in dune rather than opam. 
And I won't have time until tonight at the earliest.

ejgallego:
> The trouble is that to change the version retirement I need to make a new release, since it's in dune rather than opam.
> And I won't have time until tonight at the earliest.

You can delete the tag and redo indeed [don't forget to delete the github release too, the UI is not the besŧ]

Or this can be merged and the bound corrected in `master` for beta6, so as people prefer, IMHO it is indeed OK to merge this.

garrigue:
Ok, then let's say it's ready, and address the centos problem in the next version.
For Alpine, gtkspell3 is not available in 3.9, but it is in edge.

mseri:
Can I merge?

ejgallego:
I think so @mseri

mseri:
Thanks

ejgallego:
Thanks to you!

PR #13530, opened 2019-02-25
conf-openblas: fix switches

mseri:
They were using incorrectly os instead of `os-{distribution,family}`, and this was causing the installation to fail in some systems (See e.g. https://travis-ci.org/owlbarn/owl/jobs/497892028)


mseri:
The old opam file is broken in most systems, I suggest to replace it instead of producing a new version (it souldn't affect the working cases, but will unlock some failing ones)

mseri:
I thought we had fixed camelus for linting conf packages...

mseri:
I am going to merge this and free the CI a bit

PR #13533, opened 2019-02-25
dns and tls have depopts on mirage-kv-lwt and need mirage-kv-lwt conflicts

hannesm:


hannesm:
CI failures are unrelated -- io-page-unix missing somewhere, opam-publish trying to link both extlib and base64 >=3.0.0 (leading to multiple definitions of camlBase64__data_begin). these can be fixed in subsequent PRs.

PR #13536, opened 2019-02-26
Dummy test release: omp with snapshotted Location.error

Armael:
This is just to use the CI and test that rev-deps do not get broken by this branch...
Will close if CI passes.


avsm:
this is now done i believe...

Armael:
yep, thanks.

PR #13537, opened 2019-02-26
Sqlite3 metadata fixes

gasche:
- fix a CI build failure with 3.0.0 due to a missing `pkg-config` dependency
- consistently use `conf-sqlite3` (already used in latter versions) to improve depext quality on older versions

### DIFF
--- a/packages/sqlite3/sqlite3.3.0.0/opam
+++ b/packages/sqlite3/sqlite3.3.0.0/opam
@@ -23,6 +23,7 @@ depends: [
   "ocaml" {>= "3.12"}
   "ocamlfind" {build & >= "1.3.1"}
   "ocamlbuild" {build}
+  "conf-pkg-config" {build}
 ]
 depexts: [
   ["libsqlite3-dev"] {os-distribution = "debian"}
--- a/packages/sqlite3/sqlite3.3.0.0/opam
+++ b/packages/sqlite3/sqlite3.3.0.0/opam
@@ -24,13 +24,7 @@ depends: [
   "ocamlfind" {build & >= "1.3.1"}
   "ocamlbuild" {build}
   "conf-pkg-config" {build}
-]
-depexts: [
-  ["libsqlite3-dev"] {os-distribution = "debian"}
-  ["database/sqlite3"] {os = "freebsd"}
-  ["database/sqlite3"] {os = "openbsd"}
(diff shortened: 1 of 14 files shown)


gasche:
The CI failure appears to be due to a failure in sqlite3's `test` target under 4.0.0. I don't know what causes the error (see [logs](https://travis-ci.org/ocaml/opam-repository/jobs/498619631)), but this is unrelated to the present PR. (cc @mmottl)

mmottl:
@gasche The initial "tests" were not intended to be run as part of a test suite.  Before version 4.0.2, which addressed this problem, one of them was intended to fail.  I'm not sure what the best way forward would be.  Maybe we could just delete the `with-test` declarations from `3.0.0`, `4.0.0`, and `4.0.1`?

gasche:
That sounds fine to me, will do. Thanks!

mmottl:
Thanks for your work!

gasche:
There remains a [travis failure](https://travis-ci.org/ocaml/opam-repository/jobs/498780083) under OSX when installing 4.0.2.

```
+ /Users/travis/.opam/ocaml-system/bin/ocamlfind ocamlopt -g -I lib -linkpkg -I lib -I test lib/sqlite3.cmxa test/test_agg.cmx -o test/test_agg.native
Undefined symbols for architecture x86_64:
  "_sqlite3_enable_load_extension", referenced from:
      _caml_sqlite3_enable_load_extension in libsqlite3_stubs.a(sqlite3_stubs.o)
     (maybe you meant: _caml_sqlite3_enable_load_extension)
ld: symbol(s) not found for architecture x86_64
File "caml_startup", line 1:
Error: Error during linking
Command exited with code 2.
```

I'll still merge the PR in the current state, given that it seems to be a net improvement in any case. If you happen to know what's going on under OSX and 4.0.2, I'm happy to, for example, mark that release uninstallable under that system.

14 reverse dependencies are failing, but I'm confident that those are due to metadata issues on their own, to be investigated and fixed separately.

mmottl:
On OSX SQLite3 loadable extensions are not usually available as also noted in the footnote here: https://docs.python.org/3/library/sqlite3.html#f1

Version 4.0.2 did not account for this problem yet.  I guess it would be ok to disable this version on OSX.

PR #13540, opened 2019-02-26
conf-maxminddb

Leonidas-from-XIV:
Added libmaxminddb as `conf` package because we're working on a `ctypes`-based binding to it so it made sense to us to not just make `depexts` but rather a true `conf-libmaxminddb` package.

### DIFF
--- a/packages/conf-libmaxminddb/conf-libmaxminddb.1/opam
+++ b/packages/conf-libmaxminddb/conf-libmaxminddb.1/opam
@@ -11,11 +11,12 @@ license: "Apache-2.0"
 build: [["pkg-config" "libmaxminddb"]]
 depends: ["conf-pkg-config" {build}]
 depexts: [
-  ["libmaxminddb-dev"] {os-distribution = "debian"}
-  ["libmaxminddb-dev"] {os-distribution = "ubuntu"}
+  ["libmaxminddb-dev"] {os-distribution = "debian" | os-distribution = "ubuntu"}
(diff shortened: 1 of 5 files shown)


Leonidas-from-XIV:
This could very well fix the [build failures](https://discuss.ocaml.org/t/help-wanted-crowd-sourcing-package-build-fixes-in-opam-repository/3406/2) of the `maxminddb` package reported by @gasche.

Leonidas-from-XIV:
Truth be told I don't understand the error about `light-uninstall` since the package is not flagged as such and running `opam lint` locally just gives me a `Passed`.

mseri:
The lint error is due to a bug in the older  version of opam jsed by camelus, you can ignore it

Leonidas-from-XIV:
So CI said it failed on Ubuntu 16.04 (which should have the package according to their own package directory) and on CentOS.

So I added CentOS package names, along with Fedora and Mageia and SuSE.

Leonidas-from-XIV:
Upon further inspection this pull request requires libmaxminddb 1.1.2 because only that added `pkg-config` support. This is newer than 1.0.4 which is shipped by Ubuntu 16.04. Similarly for CentOS, the `libmaxminddb-devel` was added in CentOS 6, but even that one is 1.1.1 and only CentOS 7 had a version with `pkg-config` support.

*sigh*, めんどくさい

mseri:
That is out of your control but I think this is an improvement anyways. I'll merge, thanks

PR #13549, opened 2019-02-28
Release crunch 3.0.0

hannesm:
CHANGES
* Implement the mirage-kv 2.0.0 interface, using mirage-kv-mem (@linse @hannesm)
* Removed no longer used size function (@linse @hannesm)


hannesm:
upstream patch for datakit-ci at https://github.com/moby/datakit/pull/641

PR #13552, opened 2019-02-28
introduce conflicts between base64 >=3.0.0 and extlib, both define the base64 module

hannesm:


hannesm:
(upstreamed at https://github.com/mirage/ocaml-base64/pull/29)

mseri:
Thanks

jvillard:
This breaks infer which requires base64 *and* extlib 3.1.0 (since https://github.com/facebook/infer/pull/1056). Extlib is a transitive dependency of javalib.2.3.6 which infer uses, and infer itself uses base64.3.1.0. Would it be possible to revert this change?

avsm:
hm yes, this should be reverted. in the case of module namespace clashes, the conflict should be at the point of linking, not the point of compiling the modules involved.

dinosaure:
About the last version of `infer`, they didn't use `extlib` anymore and a patch was send to update OPAM file.

hannesm:
@dinosaure maybe @avsm and/or @jvillard can explain this in more detail. from @jvillard comment:
> Extlib is a transitive dependency of javalib.2.3.6 which infer uses, and infer itself uses base64.3.1.0

dinosaure:
Ah, I see ... too many `base64` impl.

PR #13553, opened 2019-02-28
release tls 0.10.1

hannesm:
CHANGES:
* tls-mirage: fix compilation
* tls: fix extensions length (used to include the 2 byte extension length field) if padding is inserted (introduced on May 5, 2014)
* tls-mirage: adapt to mirage-kv 2.0.0 API (@samoht)


hannesm:
revdeps opam-publish addressed in #13552

PR #13557, opened 2019-03-01
Revert "introduce conflicts between base64 >=3.0.0 and extlib, both d…

hannesm:
…efine the base64 module"

This reverts commit 634a9973befd7c8be87082d862f5471da267b4a8.

as requested by @avsm and @jvillard in https://github.com/ocaml/opam-repository/pull/13552


hannesm:
see upstream PR https://github.com/mirage/ocaml-base64/pull/30 reverting the same change. sorry about that (the change originated from a CI error during the tls release where opam-publish failed to link -- looks like some other conflict is needed)

jvillard:
Many thanks for the quick fix!

PR #13558, opened 2019-03-01
Fix base upper bound (pcre)

xclerc:
I made a small mistake when adding the upper bounds.
It is probably harmless, but certainly annoying; this PR
restore compatibility of the latest pcre with v0.12.0 of
the Jane Street packages.

### DIFF
--- a/packages/pcre/pcre.7.4.1/opam
+++ b/packages/pcre/pcre.7.4.1/opam
@@ -16,7 +16,7 @@ depends: [
   "ocaml" {>= "4.04"}
   "dune" {build & >= "1.4.0"}
   "conf-libpcre" {build}
-  "base" {build & < "v0.12"}
+  "base" {build & < "v0.13"}
   "base-bytes"
 ]
 


xclerc:
Just to clarify: it is "probably harmless" because it affects
7.4.1 (but not 7.4.0), and is "certainly annoying" because it
will trigger a useless recompilation of the pcre package.

mmottl:
@xclerc What are these bounds needed for?  I've tested PCRE, and there doesn't seem to be any problem compiling it with newer versions of `base`.  Can I remove these constraints again?

mseri:
Afaik the bounds are added in bulk before new janestreet releases to avoid breaking the world as done in the past. If the library works with v0.13 we can now bump the bound to v0.14 in a separate PR. Would you mind sending one @mmottl ?

xclerc:
Anil described the process in [1]; the last message
of the thread indicates that the unnecessary bounds
are removed through `opam-check-all`.

[1] https://discuss.ocaml.org/t/experimenting-with-a-new-opam-repository-release-strategy-for-large-libraries/2918/23

mmottl:
@mseri I've just submitted https://github.com/ocaml/opam-repository/pull/15354.  The constraints have been completely removed, because I have no good reason to believe that future versions of `base` and `stdio` will break my packages.  I hope that automated processes will be able to take care of adding necessary constraints in the future.

PR #13566, opened 2019-03-02
remove upper bound on base and stdio from configurator

jberdine:
Builds locally with base v0.12, so seems ok to relax the upper bound.

I'm not sure if sending this PR contradicts the desired workflow for these upper bounds. If so, please just close this.


mseri:
Seems to work fine, all the failures in the revdeps are unrelated

PR #13568, opened 2019-03-03
+session.0.4.1

seliopou:


avsm:
Same comment for these; the `session` constraints here could be good to keep together in the subpackages. Can merge this like if you prefer to keep them looser (but we may find build failures in the future if the solver finds a combination that doesn't work together)

mseri:
I am going to merge, this can be changed in a second stage

PR #13574, opened 2019-03-03
alcotest: old versions do not work on OCaml 4.08

samoht:
Because Pervasives is used directly and warning-as-errors is enabled.


samoht:
See http://check.ocamllabs.io/4.08.0+beta1/bad/alcotest.0.7.2

mseri:
Thanks

PR #13575, opened 2019-03-03
datakit-bridge-github: use a fix version of datakit-github

samoht:
See http://check.ocamllabs.io/4.03.0/bad/datakit-bridge-github.0.11.0


kit-ty-kate:
The failures are due to solver timeout. We can always check again that it worked with opam-check-all. Thanks!

PR #13580, opened 2019-03-04
Added the initial release of the mmdb package

martinslota:
This package probably won't build on older Ubuntu because the corresponding `libmaxminddb` does not support `pkg-config`.

### DIFF
--- a/packages/mmdb/mmdb.0.1.0/opam
+++ b/packages/mmdb/mmdb.0.1.0/opam
@@ -8,15 +8,15 @@ maintainer: "Team Platypus <platypus@issuu.com>"
 build: [["dune" "build" "-p" name "-j" jobs]]
 run-test: [["dune" "runtest" "-p" name "-j" jobs]]
 depends: [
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
I'm wondering why you have all those strict upper-bound constraints. It's of course very good to have them if you know the next version is gonna break your package but over-using them usually leads to users having to downgrade a lot of their environment and it's not very good

Leonidas-from-XIV:
We use that mostly in our software because newer versions keep on breaking builds and it is very annoying to have the environment continuously shift under your feet because something in ppx-land changed and something else now fails with warnings but we can remove the upper bounds…

kit-ty-kate:
Well, the revdeps tests we do on every opam-repository PRs are here to tell us if this package needs new constraints or not

Leonidas-from-XIV:
Constraints are gone now.

kit-ty-kate:
Thanks a lot!

PR #13584, opened 2019-03-05
[new release] owl-zoo, owl-top, owl-base, owl and owl-plplot (0.5.0)

mseri:
OCaml Scientific and Engineering Computing - Zoo

- Project page: <a href="https://github.com/owlbarn/owl">https://github.com/owlbarn/owl</a>
- Documentation: <a href="https://owlbarn.github.io/">https://owlbarn.github.io/</a>

##### CHANGES:

* Improve building and installation.
* Fix bugs and improve performance.
* Add more functions to Algodiff.
* Split plot module out as sub library.
* Split Tfgraph module out as sub library.

PR #13585, opened 2019-03-05
owl: add upper bounds

mseri:
This is to address the changes in interface and build related to the failures in https://github.com/ocaml/opam-repository/pull/13584


mseri:
I don't really expect anything to build here

PR #13596, opened 2019-03-06
Package morbig.0.10.3

Niols:
### `morbig.0.10.3`
A trustworthy parser for POSIX shell
them, and constructs a concrete syntax tree for each of them. The

---

---

### DIFF
--- a/packages/morbig/morbig.0.10.3/opam
+++ b/packages/morbig/morbig.0.10.3/opam
@@ -1,0 +1,43 @@
+opam-version: "2.0"
+
+synopsis: "A trustworthy parser for POSIX shell"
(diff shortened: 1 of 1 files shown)


Niols:
Hum. The test that fails is because the dependencies of Morsmall (which I am packaging too) are a bit optimistic and don't specify on which version of Morbig they work. I have a new package for Morsmall that follows Morbig's changes, should I file a PR for the two of them at the same time?

Also, more generally, when facing tools whose API can change (Morbig is not considered stable yet), how should we specify dependencies in those tools? Should we put the exact version that works and then update the package whenever a new version that does work is out?

PR #13597, opened 2019-03-06
Packages Morbig 0.10.3 and Morsmall 0.2.0

Niols:
### Morbig 0.10.3

them, and constructs a concrete syntax tree for each of them. The

### Morsmall 0.2.0

A concise AST for POSIX shell.

### DIFF
--- a/packages/morbig/morbig.0.10.3/opam
+++ b/packages/morbig/morbig.0.10.3/opam
@@ -24,7 +24,7 @@ dev-repo: "git://github.com/colis-anr/morbig.git"
 depends: [
   "dune"                 {build & >= "1.4.0"}
   "menhir"               {>= "20170509"}
-  "ocaml"                {build & >= "4.04"}
+  "ocaml"                {>= "4.04"}
   "odoc"                 {with-doc}
   "ppx_deriving_yojson"
   "visitors"             {>= "20180513"}
--- a/packages/morbig/morbig.0.10.3/opam
+++ b/packages/morbig/morbig.0.10.3/opam
@@ -27,7 +27,7 @@ depends: [
   "ocaml"                {build & >= "4.04"}
   "odoc"                 {with-doc}
   "ppx_deriving_yojson"
-  "visitors"             {build & >= "20180513"}
+  "visitors"             {>= "20180513"}
   "yojson"
 ]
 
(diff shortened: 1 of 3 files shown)


packages/morsmall/morsmall.0.2.0/opam
```diff
@@ -0,0 +1,33 @@
...
+
+homepage: "https://github.com/colis-anr/morsmall"
+bug-reports: "https://github.com/colis-anr/morsmall/issues"
+dev-repo: "git+ssh://git@github.com/colis-anr/morsmall.git"
+
+depends: [
+  "dune"          {build}
+  "morbig"        {>= "0.10.0"}
+  "ocaml"         {>= "4.04"}
+  "ppx_deriving"  {build}
```
mseri: This could have (now or in the future) a runtime, it should not be tagged `buil` if it is used in morsmall

Niols: Is that always so? PPXes without runtimes shouldn't be tagged `build` in case they end up having one? Shouldn't we just change that in case it happens?

kit-ty-kate: there seems to be runtime dependencies for both `visitors` and `ppx_deriving` already:
* https://gitlab.inria.fr/fpottier/visitors/blob/master/src/META
* https://github.com/ocaml-ppx/ppx_deriving/blob/master/pkg/META.in

Niols: OK, fair enough, I just pushed this.

packages/morbig/morbig.0.10.3/opam
```diff
@@ -0,0 +1,43 @@
...
+license: "GPL3"
+
+homepage: "https://github.com/colis-anr/morbig"
+bug-reports: "https://github.com/colis-anr/morbig/issues"
+dev-repo: "git://github.com/colis-anr/morbig.git"
+
+depends: [
+  "dune"                 {build & >= "1.4.0"}
+  "menhir"               {>= "20170509"}
+  "ocaml"                {build & >= "4.04"}
```
kit-ty-kate:
```suggestion
  "ocaml"                {>= "4.04"}
```
ocaml shouldn't be a build dependency imo

mseri:
Thanks!

PR #13602, opened 2019-03-07
Better zstd dependencies

xclerc:
The `zstandard` package fails to build if the version of
Zstandard installed on the system is too old (I have
seen this *e.g.* on Debian Stretch).

This pull request adds a revised version of the
`conf-zstd` package that ensures that at least 1.3.8
is present, and make `zstandard` depend on it.

However, I am not certain it is the best (or even a
correct) way to achieve that; feedback more than
welcome.

### DIFF
--- a/packages/zstandard/zstandard.v0.12.0/opam
+++ b/packages/zstandard/zstandard.v0.12.0/opam
@@ -13,7 +13,7 @@ depends: [
   "ocaml"       {>= "4.07.0"}
   "core_kernel" {>= "v0.12" & < "v0.13"}
   "ppx_jane"    {>= "v0.12" & < "v0.13"}
-  "conf-zstd"   {>= "1.3.8"}
+  "conf-zstd"   {= "1.3.8"}
   "ctypes"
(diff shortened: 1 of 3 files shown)


xclerc:
(Note about the required version: at this point I know
that 1.3.8 is OK and that 1.3.5 is KO.)

packages/zstandard/zstandard.v0.12.0/opam
```diff
@@ -13,7 +13,7 @@ depends: [
   "ocaml"       {>= "4.07.0"}
   "core_kernel" {>= "v0.12" & < "v0.13"}
   "ppx_jane"    {>= "v0.12" & < "v0.13"}
-  "conf-zstd"
+  "conf-zstd-1.3.8"
```
mseri: This should be `conf-zstd {=“1.3.8”}`

xclerc: Indeed; that is an embarrassing mistake...
Thanks!

mseri:
Thanks

PR #13606, opened 2019-03-07
The great Iter migration.

Drup:
Following the addition of `Seq` to the standard library, @c-cube and I decided to rename `Sequence` to `Iter`. Unfortunately, due to github tarballs being obnoxious, this breaks all the checksums

So this commit fixes them all, and rewrite the path on the way. I also upgraded the hashing algorithm. :)

For the curious, and due to https://github.com/ocaml/opam/issues/3767 , I had to use the following script:
```zsh
#!/usr/bin/zsh -f
for i in *(/); do
    url="$(opam show -f "url.src:" "./${i}/opam")"
    url="${(Q)url}"
    tarball="/tmp/bla/${i}.tar.gz"
    wget -O "${tarball}" "${url}";
    checksum="$(sha256sum ${tarball} | cut -d " " -f 1 -)"
    sub="s/checksum:.*/checksum: \"sha256=${checksum}\"/" 
    sed -i -e $sub ./$i/opam
    sed -i -e "s/\/sequence/\/iter/g" ./$i/opam
done
```

### DIFF
--- a/packages/sequence/sequence.0.1/opam
+++ b/packages/sequence/sequence.0.1/opam
@@ -1,7 +1,7 @@
 opam-version: "2.0"
 maintainer: "simon.cruanes@inria.fr"
 authors: "Simon Cruanes"
-homepage: "https://github.com/c-cube/sequence/"
+homepage: "https://github.com/c-cube/iter/"
 build: make
 remove: [["ocamlfind" "remove" "sequence"]]
 depends: [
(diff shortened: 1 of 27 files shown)


Drup:
(We will submit the new iter package soon)

kit-ty-kate:
The checksums for sequence.0.1, 0.2 and 1.1 don't match

Drup:
(For the curiouser: everything was broken because shell script SUCKS. I should have used shcaml)

c-cube:
The rev-dep failures are caused by wrong bounds on old versions of containers, which I plan to fix in another PR (it's related to @gasche 's large effort to fix old packages).

kit-ty-kate:
Thanks!

PR #13609, opened 2019-03-07
Upgrade obytelib to version 1.5, ocamlclean to version 2.2 and ocapic to version 3.4.

bvaugon:


bvaugon:
Build failed on MacOS because of the **graphics** library which seems to do not compile on MacOS. Since the problem comes from graphics and not from the commited packages, and only on MacOS, I think this pull request may be merged.

packages/ocapic/ocapic.3.4/opam
```diff
@@ -0,0 +1,28 @@
...
+]
+remove: [
+  ["./configure" "-prefix" prefix]
+  [make "uninstall"]
+]
+depends: [
+  "ocaml" {>= "4.07.0"}
+  "camlp4"
+  "graphics"
+  "ocamlbuild"
```
kit-ty-kate:
```suggestion
  "ocamlbuild" {build}
```
do you need ocamlbuild at runtime?

packages/ocamlclean/ocamlclean.2.2/opam
```diff
@@ -0,0 +1,25 @@
...
+  ["./configure" "-prefix" prefix]
+  [make "all"]
+]
+remove: [
+  ["./configure" "-prefix" prefix]
+  [make "uninstall"]
+]
+depends: [
+  "ocaml" {>= "4.07.0"}
+  "ocamlbuild"
```
kit-ty-kate: same comment here

kit-ty-kate:
otherwise apart from my two comments, it looks indeed fine to merge

packages/ocamlclean/ocamlclean.2.2/opam
bvaugon: Yes, indeed. Thanks!

mseri:
Thanks!

PR #13621, opened 2019-03-08
owl-plplot: add missing dep

mseri:
It gets it transitively but it is always good to keep things fixed.
I have also sent the fix upstream.


mseri:
Failures are the expected ones

PR #13623, opened 2019-03-08
Fix cstruct-*.3.6.0 dependencies

pascutto:


samoht:
The error is not related, so merging.

PR #13631, opened 2019-03-08
Package glfw-ocaml (3.2.1).

SylvainBoilard:
CHANGES:

* Expose the details of the image type; keep it private.


mseri:
The failures are due to incompatible library versions, so not much can be done on that front

SylvainBoilard:
Indeed. I will try to fix the MacOS build though.

Thank you for merging.

PR #13620, opened 2019-03-08
[new release] dune (1.8.1)

rgrinberg:
CHANGES:

- Correctly write `dune-package` when version is empty string (ocaml/dune#1919, fix ocaml/dune#1918,
  @rgrinberg)

PR #12810, opened 2018-10-15
3 packages from ryanslade/influxdb-ocaml at 0.2.0

ryanslade:
This pull-request concerns:
-`influxdb.0.2.0`: InfluxDB client library
-`influxdb-async.0.2.0`: InfluxDB client library using async for concurrency
-`influxdb-lwt.0.2.0`: InfluxDB client library using lwt for concurrency

---
* Homepage: https://github.com/ryanslade/influxdb-ocaml
* Source repo: git://github.com/ryanslade/influxdb-ocaml.git
* Bug tracker: https://github.com/ryanslade/influxdb-ocaml/issues

---
:camel: Pull-request generated by opam-publish v2.0.0

PR #13632, opened 2019-03-09
fix influxdb constraints

avsm:
cc @ryanslade and #12810 -- the older packages may be selected
by the opam solver, so these constraints ensure the 'gang' of packages
are selected together

fixes build failure observed in #13620


mseri:
Looks good, thanks

PR #13652, opened 2019-03-11
[WIP] pre-release of ppx_deriving.4.3

gasche:
This is a pre-release commit to test the ppx_deriving revdeps in the
opam-repository CI. Don't merge yet!

### DIFF
--- a/packages/ppx_deriving/ppx_deriving.4.3/opam
+++ b/packages/ppx_deriving/ppx_deriving.4.3/opam
@@ -1,0 +1,36 @@
+opam-version: "2.0"
+maintainer: "whitequark <whitequark@whitequark.org>"
+authors: [ "whitequark <whitequark@whitequark.org>" ]
+license: "MIT"
+homepage: "https://github.com/whitequark/ppx_deriving"
+doc: "https://whitequark.github.io/ppx_deriving"
+bug-reports: "https://github.com/whitequark/ppx_deriving/issues"
+dev-repo: "git+https://github.com/whitequark/ppx_deriving.git"
+tags: [ "syntax" ]
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+  ["dune" "build" "@doc" "-p" name "-j" jobs] {with-doc}
+]
+depends: [
+  "dune"     {build >= "1.0.0"}
+  "cppo"     {build}
+  "ppxfind"  {build}
+  "ocaml-migrate-parsetree"
+  "ppx_derivers"
+  "ppx_tools"  {>= "4.02.3"}
+  "result"
+  "ounit"      {with-test}
+  "ocaml" {>= "4.02"}
+]
+synopsis: "Type-driven code generation for OCaml >=4.02"
+description: """
+ppx_deriving provides common infrastructure for generating
+code based on type definitions, and a set of useful plugins
+for common tasks.
+"""
+url {
+  src: "https://github.com/gasche/ppx_deriving/archive/release-4.3.tar.gz"
+  checksum: "sha256=e555c56bcdc6d58997a3df5b785de0f1602c75b36a3df683c166c01d13efd9b1"
+}
--- a/packages/ppx_deriving/ppx_deriving.4.3/opam
+++ b/packages/ppx_deriving/ppx_deriving.4.3/opam
@@ -2,7 +2,7 @@ opam-version: "2.0"
 maintainer: "whitequark <whitequark@whitequark.org>"
 authors: [ "whitequark <whitequark@whitequark.org>" ]
 license: "MIT"
-homepage: "https://github.com/whitequark/ppx_deriving"
+homepage: "https://github.com/ocaml-ppx/ppx_deriving"
 doc: "https://whitequark.github.io/ppx_deriving"
 bug-reports: "https://github.com/whitequark/ppx_deriving/issues"
 dev-repo: "git+https://github.com/whitequark/ppx_deriving.git"
--- a/packages/ppx_deriving/ppx_deriving.4.3/opam
+++ b/packages/ppx_deriving/ppx_deriving.4.3/opam
@@ -5,7 +5,7 @@ license: "MIT"
 homepage: "https://github.com/ocaml-ppx/ppx_deriving"
 doc: "https://whitequark.github.io/ppx_deriving"
 bug-reports: "https://github.com/whitequark/ppx_deriving/issues"
-dev-repo: "git+https://github.com/whitequark/ppx_deriving.git"
+dev-repo: "git+https://github.com/ocaml-ppx/ppx_deriving.git"
 tags: [ "syntax" ]
 build: [
   ["dune" "subst"] {pinned}


mseri:
Not bad! I only gave a broef look, but the only two failures, that probably just require an upper bound on ppx_deriving, are for “old” `ppx_deriving_{yojson,cmcliner}` failing with
```
# Error: Unbound module Ppx_deriving_runtime
```

avsm:
Wow, nice job with the port!

In the future, you can now create a "draft PR" which will prevent it being merged by mistake.

gasche:
Thanks @avsm, but most of the credit goes to @rgrinberg and @diml who did the actual work.

@mseri thanks a lot for the revdeps analysis, this is very helpful (I'm busy with other things as well). I'll update this PR updated upper-bounds and a final .opam, hopefully ready for merging, as soon as possible.

packages/ppx_deriving/ppx_deriving.4.3/opam
```diff
@@ -0,0 +1,36 @@
+opam-version: "2.0"
+maintainer: "whitequark <whitequark@whitequark.org>"
+authors: [ "whitequark <whitequark@whitequark.org>" ]
+license: "MIT"
+homepage: "https://github.com/whitequark/ppx_deriving"
```
avsm:
```suggestion
homepage: "https://github.com/ocaml-ppx/ppx_deriving"
```

packages/ppx_deriving/ppx_deriving.4.3/opam
```diff
@@ -0,0 +1,36 @@
...
+doc: "https://whitequark.github.io/ppx_deriving"
+bug-reports: "https://github.com/whitequark/ppx_deriving/issues"
+dev-repo: "git+https://github.com/whitequark/ppx_deriving.git"
```
avsm:
```suggestion
dev-repo: "git+https://github.com/ocaml-ppx/ppx_deriving.git"
```

...etc

packages/ppx_deriving/ppx_deriving.4.3/opam
```diff
@@ -0,0 +1,36 @@
...
+tags: [ "syntax" ]
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+  ["dune" "build" "@doc" "-p" name "-j" jobs] {with-doc}
+]
+depends: [
+  "dune"     {build >= "1.0.0"}
```
avsm: Does this really work with dune 1.0, @rgrinberg @diml? istr lots of pps fixes post 1.0 that might or might not affect this package

rgrinberg: I'm not sure. It depends if you want to be pessimistic or optimistic. Off the top of my head, I cannot think of any reason why dune 1.0 wouldn't work with ppx_deriving, but as you've said, there's definitely been some bugs along the way.

diml: Personally, I never take any chances and always put the version of Dune I'm currently using as a lower bound. Since Dune is really good at backward compatibility, upping the lower-bound is a no-brainer.

rgrinberg: In this case, the version that I used to develop the port was 1.6.3.

gasche:
I haven't looked in-depth yet but the [revdeps log](https://ci.ocamllabs.io/ocaml/opam-repository/pr/13652?test=2%20Revdeps) are giving me pause. Yes, the newest version of ppx_deriving_yojson works, but older version are broken, and some ppx-deriving plugins appear to be completely broken by the change (ppx_crowbar, ppx_hardcaml, ppx_deriving_morphism, visitors...).

mseri:
Are those? Visitor, like many others, fails with unknown cppo_ocamlbuild lobrary (probably just need upper bound on cppo). The others are failing due to that unbound ppx_deriving_runtime, that indeed may have to do with the port to dune (but maybe just requires upper bounds and linking to the runtime when updating the lobraries, in which case I would suggest to bump the major version)

rgrinberg:
The issue with `ppx_deriving_hardcaml`, `ppx_deriving_morphism`, `visitors` can safely be ignored as well. It's simply missing ocamlbuild or ocamlbuild_cppo as a dependency. This is all a consequence of opam files being rarely accurate.

> The others are failing due to that unbound ppx_deriving_runtime, that indeed may have to do with the port to dune

Indeed, I think that in the dune port, we dropped the dependency on the runtime in some places where it wasn't necessary. I suppose some users were relying on it being present transitively.

So I don't quite yet see a cause for concern. I am going to use this opportunity to remind @avsm to put up his WIP .opam generation branch as these problems demonstrate how important it is :)

rgrinberg:
don't merge this yet however, i'm still investigating the other failures

gasche:
My plan is to fix the opam descriptions of these dependencies before merging anyway, so that we can ensure that we have found all the cause for package failures. (Not all packages are in scope, but let's say the `ppx_*` stuff.) I could just update this PR with repo-fixes for the dependencies, or proceed with independent PRs, not sure yet.

gasche:
(And: thanks again to you all for the help triaging these failures, this is very helpful.)

rgrinberg:
> Indeed, I think that in the dune port, we dropped the dependency on the runtime in some places where it wasn't necessary. I suppose some users were relying on it being present transitively.

Actually, this is caused by a breaking change in the dune port. In the previous ppx_deriving, `ppx_deriving` was just an alias for `ppx_deriving.runtime` that introduced the preprocessor. But since all the cmi's were in one directory, this allowed one to use `ppx_deriving.api` as well.

The new dune port is strict about this stuff so `ppx_deriving` is basically an empty library. In fact, I'm not sure why dune is allowing us to use it.

I recommend that we do the following:

* Turn the current `ppx_deriving.api` into `ppx_deriving`.

* Change `ppx_deriving.api` into a deprecated alias of `ppx_deriving`.

The reasoning is as follows:

It's much better if the main module name matches the name of the library. Since the main module name for the api is `ppx_deriving`, I think we should prefer that as the library name for the API.

If this is acceptable to @gasche , I can prepare a PR.

gasche:
I'm not as familiar with `ppx_deriving` as I should be, so I'm a bit slow here. Let me make sure I understood properly:

- *users* of ppx_deriving plugins want to link to the runtime, so in the past depending on `ppx_deriving`, which was an alias for `ppx_deriving.runtime`, did exactly that. On the other hand, using of ppx_deriving plugins do not need to do this, if the plugins correctly records its own dependency on `ppx_deriving.runtime`, depending on just the plugin should be enough. If users of ppx_deriving are broken by a change in the `ppx_deriving` package, we can fix it by fixing the plugin's metadata.

- *developers* of ppx_deriving plugins want to export a dependency of their plugins on the runtime, but at build-time they need to link to `ppx_deriving.api`. Before the dune port this magically worked when depending on `ppx_deriving` because the cmi files were at the same place.

gasche:
If the above analysis is correct, I think we could/should do one of the two following things (please let me know what you think!):

- *not* have a ppx_deriving library at all, to force people to be extra-explicit about what they mean, the runtime or the api. Given that (if my understanding is correct) plugin users should never need to mention either anyway, only plugin developers need to think about it, I think this would not be user-unfriendly.
- if we keep having ppx_deriving as this library of unclear scope, have it depend on both the runtime and the api, so that everyone (plugin users and developers) get served the meal they expect.

I would go for option (1), given that the failures introduced by this change can be fixed at the opam-repository level, and so we can make it totally unobservable to plugin users¹. We need to tell the plugin developers to be careful in future release, but this is a small set of people so it's all right.

¹: I'm assuming here that there are no cases where users forgot to properly document their usage of specific plugins in their build script (`show`, `ord`), and it worked magically because the plugins where at a location exposed by `ppx_deriving`. If that happens, then we need to fix user build scripts in addition to their opam metadata, and this is a breaking change. Do we know for sure that this does not happen?

rgrinberg:
Your assessment is spot on. Just a couple of points to add:

* Using `ppx_deriving` was never a full solution as other plugins would have their own runtimes that users would need to remember to add as dependencies.

* Users who are using dune will not experience any breakage due to missing runtime dependencies, as those are managed automatically by dune when the plugin authors specify them.

Given that dune makes the runtimes pretty much transparent to users, I don't think the api vs. runtime distinction is necessary. Users should never use `ppx_deriving` directly, and rely on particular plugins. While plugin authors should just use `ppx_deriving` and specify runtime deps accurately.

Also, we are still introducing this ugly naming inconsistency. It's quite annoying that the library `ppx_deriving.api` contains the `Ppx_deriving` module. It means that it will be very hard to move this library to `(wrapped true)` mode in the future.

diml:
Note that the `ppx_deriving` library is currently required for using ppx_deriving with ocamlfind. It is the library that declares the ppx_deriving dynlinker, so all ppx_deriving plugin need to depend on it if they want to be compatible with `ocamlfind ocamlc -package ppx_deriving_blah`.

gasche:
There are two things I don't completely understand in your comment:

> Given that dune makes the runtimes pretty much transparent to users

If I understand correctly, you assume that plugin authors have specified that they depend on the plugin runtime, and you mean that dune will see post-processed files depending on the `Ppx_deriving_runtime` module and do the right thing from that dependency indication. (Or does it work magically even if the plugin authors does not do this?)

> While plugin authors should just use ppx_deriving and specify runtime deps accurately.

What do you mean by "and specify runtime deps accurately"?

gasche:
@rgrinberg, @diml, so let me recapitulate the emerging proposal:

- have `ppx_deriving` be the package that introduces the `Ppx_deriving` module in scope (currently ppx_deriving.api) *and* "declares the ppx_deriving dynlinker". (Would there be a way to split those two things and would that be useful?) Plugin authors depend on that in their build scripts (for the API) and export it as a dependency to their users (for the ocamlfind support).
- have `ppx_deriving.runtime` be a package that does not depend on `ppx_deriving`, and just indicates a dependency on the `Ppx_deriving_runtime` module. Plugin authors export this as a dependency to their users (for the runtime support). We could consider turning this subpackage into a `ppx-deriving-runtime` package, but let's maybe not do all sorts of compatibility-impacting changes at once.

rgrinberg:
> Note that the ppx_deriving library is currently required for using ppx_deriving with ocamlfind. It is the library that declares the ppx_deriving dynlinker, so all ppx_deriving plugin need to depend on it if they want to be compatible with ocamlfind ocamlc -package ppx_deriving_blah.

I see. Okay, then I think we should just keep `ppx_deriving.api`. While it's uglier, it will keep things simple for us.

@gasche if a plugin author uses dune, then they may specify the `ppx_runtime_libraries` field. For example, given a hypothetical `ppx_deriving_show`:

```
(library
 (name ppx_deriving_show)
 (libraries ppx_deriving.api)
 (ppx_runtime_libraries ppx_deriving_show.runtime))
```

Now, whenever users of this plugin will add `ppx_deriving_show` as a preprocessor, then `ppx_deriving_show.runtime` will be introduced as a library.

The end result is that end users never have to worry about runtime dependencies, as dune introduces them automatically whenever a ppx is used.

rgrinberg:
> have ppx_deriving be the package that introduces the Ppx_deriving module in scope (currently ppx_deriving.api) and "declares the ppx_deriving dynlinker". (Would there be a way to split those two things and would that be useful?) Plugin authors depend on that in their build scripts (for the API) and export it as a dependency to their users (for the ocamlfind support).

I think this might require more hackery than we're willing to stomach. Let's see if @diml thinks this is reasonable. Personally, it seems to be too complicated to be worth the effort so I think your original proposal works better.

> have ppx_deriving.runtime be a package that does not depend on ppx_deriving, and just indicates a dependency on the Ppx_deriving_runtime module. Plugin authors export this as a dependency to their users (for the runtime support). We could consider turning this subpackage into a ppx-deriving-runtime package, but let's maybe not do all sorts of compatibility-impacting changes at once.

This is actually not that useful as the runtime is fairly useless as a standalone package. The only situation where users might want to depend on the runtime alone is when they checked in generated code. Not a very common scenario.

gasche:
So:
- `ppx_deriving` declares stuff to ocamlfind
- `ppx_deriving.api` is used in package developers' build scripts (to get the `Ppx_deriving` module)
- `ppx_deriving.runtime` is indicated as a runtime dependency (in opam and in dune) by package developers

If I understand correctly, this requires no change to the current ppx_deriving sources (the `ppx_deriving` library is empty as pointed out by Rudi, but its META file has the right ocamlfind magic as pointed out by Jérémie), and if we fix plugin and user's opam metadata the fixes should be backward-compatible with the pre-dune ppx_deriving versions. (So that sounds quite nice.)

diml:
Agreed, fixing plugins seems like the nicest solution to me as well.

rgrinberg:
>  and if we fix plugin and user's opam metadata the fixes should be backward-compatible with the pre-dune ppx_deriving versions. (So that sounds quite nice.)

Not quite. Plugin authors who have specified `ppx_deriving` instead of `ppx_deriving.api` will need their build systems fixed. But I'm in favor of keeping the status quo nonetheless.

rgrinberg:
It just occurred to me that we could also add `ppx_deriving.api` as a dependency of `ppx_deriving`. This would fix all existing plugins but at the cost of making the plugin api available at runtime (which is certainly not desirable).

gasche:
Okay, so let's try that, and see what the revdep testing tells us. Thanks to both for your time!

diml:
Wait, if you add ppx_deriving.api as a dependency of ppx_deriving, then that means that anybody doing this:

```
$ ocamlfind ocamlc -o foo -linkpkg -linkall -package ppx_deriving_blah foo.ml
```

will end up linking the `Ppx_deriving` module inside `foo`.

diml:
TBH, every time I changed something at this level, it gave me a headache and I ended up somehow breaking something. I strongly suggest to try fixing the plugins instead, it is the most straightforward solution. I can help with the task.

BTW, independently of the port to dune, incorrectly using ppx_deriving instead of ppx_deriving.api has several negative consequences:

- it means that the code of the plugin itself will be preprocessed for no reason
- it means that the code of the plugin will see the modules of `Ppx_deriving_runtime`

So fixing these dependencies is a good thing.

gasche:
@rgrinberg and @diml, I wrote a complete (up-to-date) explanation of the issue in a new comment in https://github.com/ocaml-ppx/ppx_deriving/pull/170#issuecomment-472796569. Let me know (here or there) if something is missing/wrong and I'll update it. I think everyone agrees with the idea of fixing the incorrect dependencies.

gasche:
(And: thanks for your time and effort dealing with this issue.)

packages/ppx_deriving/ppx_deriving.4.3/opam
```diff
@@ -0,0 +1,36 @@
+opam-version: "2.0"
+maintainer: "whitequark <whitequark@whitequark.org>"
+authors: [ "whitequark <whitequark@whitequark.org>" ]
+license: "MIT"
+homepage: "https://github.com/ocaml-ppx/ppx_deriving"
+doc: "https://whitequark.github.io/ppx_deriving"
```
XVilka: Do you need to change the maintainer and bugs address too? Since it is under ocaml-ppx now.

avsm:
I'm just closing this in order to reduce the load on the CI due to the large number of revdeps. Please do feel free to reopen when you push new changes.

gasche:
Yesterday and today I went through all the failing revdeps to understand what the issue was, and I believe that I should now have fixed all of them. My notes are attached below; the failures in `ppx_deriving_*` plugins where all related to a missing `ppx_deriving.runtime` dependency (the conversation we had above was very helpful for me to navigate there), some related to a crufty `myocamlbuild.ml` script with hardcoded module paths (the problem that those crufty myocamlbuild.ml was solving was to let developers test their plugins without installing it, that is, before the ocamlfind packages were available; Dune build systems make this much easier, but it's also possible (and reasonably easy) to fix the myocamlbuild.ml to be more robust, so I sent those fixes upstream when that made sense).

Once we the corresponding opam-repository changes have been merged (I've merged a number of them already, and am waiting for CI reports on some others), I will create a new temporary issue for the ppx_deriving release, to check that the revdeps are passing as intended, and hopefully turn that into the final release.

----

revdep analysis logs

https://ci.ocamllabs.io/ocaml/opam-repository/pr/13652?test=2%20Revdeps

- clangml: due to visitors failure

- electrod: due to visitors failure

- visitors: missing `cppo_ocamlbuild` dependency (used to be a transitive dependency through ppx_deriving),
   upstream fix https://gitlab.inria.fr/fpottier/visitors/issues/5
   opam-repository fix https://github.com/ocaml/opam-repository/pull/14003

- elpi: old versions require a ppx_tools_versioned upper bound
   opam-repository fix https://github.com/ocaml/opam-repository/pull/14004

   https://ci.ocamllabs.io/log/saved/docker-run-708c4d7b798737602087ba7605559965/e28720ec572b2aa125585ed35443c5af22df30f8

   https://ci.ocamllabs.io/log/saved/docker-run-af0437f1efd0867e391f079336a9ae66/26a7e05c256aab17bfdd4337c4a44fe96e6ff7ce

- js_of_ocaml-ppx:

   building with dune-built ppx_deriving needs the following commit,
   present in 3.1.0 but not 3.0.2:

   ```
   commit ec8bac63e285139c61ec8c94b8db6350b10d4f22
   Author: Hugo Heuzard <hugo.heuzard@gmail.com>
   Date:   Sat Feb 10 04:09:37 2018 +0000

       Ppx: fix deps for ppx_deriving_json
   ```

   opam-repository fix https://github.com/ocaml/opam-repository/pull/14005

   https://ci.ocamllabs.io/log/saved/docker-run-374103b757f4bbdc465c6769636b14c5/4387038ee6a19e9a344f15849996bae4c5457858

- links:

   links.0.7.3 fails to build with a type error (a weak type variable
   issue) due to an incompatibility with ocaml 4.07. This was fixed upstream in

   https://github.com/links-lang/links/commit/e4ae8a44d72251f362846f388ef92091315929ad

   which is only included in links.0.8

   opam-repository fix https://github.com/ocaml/opam-repository/pull/14006

   https://ci.ocamllabs.io/log/saved/docker-run-0bcc57fbc508a2dbc928a0da9774370d/5bb974aaddfb7beb2a0f5793d75cee313067a22c

- morsmall: visitors failure

- ppx_deriving_cmdliner

   Before 0.4.1, this package forgot a dependency on the ppx_deriving
   runtime (ocamlfind packages 'ppx_deriving' or
   'ppx_deriving.runtime') when building its own runtime
   ppx_deriving_cmdliner_runtime. This worked correctly as the object
   files of ppx_deriving.runtime and ppx_deriving.api were installed in
   the same directory, hiding the error.

   0.4.1 uses dune as its build system and everything works fine there.

   For the record, applying the following patch to 0.4.0 fixes the
   build (I wanted to make sure that plugin authors using ocamlbuild
   could easily fix their build):

   ```diff
   diff --git a/_tags b/_tags
   index 3cbcb9e..0d9f14e 100644
   --- a/_tags
   +++ b/_tags
   @@ -2,4 +2,5 @@ true: warn(@5@8@10@11@12@14@23@24@26@29@40), bin_annot, safe_string, cppo_V_OCAM
    
    "src": include
    <src/*.{ml,mli,byte,native}>: package(ppx_tools.metaquot), package(ppx_deriving.api), package(result)
   +<src/ppx_deriving_cmdliner_runtime.*>: package(ppx_deriving)
    <src_test/*.{ml,byte,native}>: debug, package(result), package(oUnit), package(cmdliner), use_cmdliner
   ```

   Build failure logs:
   + https://ci.ocamllabs.io/log/saved/docker-run-3bda0bcda9a1dc89c776e08086048b6a/5b4c2eb661b15ee9d24ac806a33399e9c8ac05c7
   + https://ci.ocamllabs.io/log/saved/docker-run-0059c5c9a3e1fdba6dbe648ce4951669/64f0542938967251be5854a462142c5160de94f7
   + https://ci.ocamllabs.io/log/saved/docker-run-ec68749a9a8433ff8f2c36e0bb848a6e/02b25b7070c3ce8a4ec81aa5e3895008940bdce1
   + https://ci.ocamllabs.io/log/saved/docker-run-64a10aa432775f710725c4729ca6dd7a/47b8e4b7ddf84ed2ad206e8d1247fda50d71bcbe

- ppx_deriving_crowbar

  As for ppx_deriving_cmdliner (see above), there was a missing dependency on ppx_deriving.runtime.

  opam-repository fix https://github.com/ocaml/opam-repository/pull/14008

  upstream fix https://github.com/yomimono/ppx_deriving_crowbar/pull/4

  Build failure logs:
  + https://ci.ocamllabs.io/log/saved/docker-run-1ccccffee439ee8c3df71576664bfbac/d2595746f647d351d1dcbd827285cc96172fe3d8
  + https://ci.ocamllabs.io/log/saved/docker-run-495da4ce6bbd65b0697b9d7f3b7e042a/40d462d6e14e27ecb156bb6568f37722daee978f

- ppx_deriving_hardcaml

  incorrect myocamlbuild.ml setup, incompatible with ppx_deriving.4.3

  upstream fix https://github.com/xguerin/ppx_deriving_hardcaml/pull/12

  opam-repository fix https://github.com/ocaml/opam-repository/pull/14013

  Build failure logs:
  + https://ci.ocamllabs.io/log/saved/docker-run-61a831500ef1cd867764eac765458682/7797de6206516b241664d27d9da95011aae98cc7
  + https://ci.ocamllabs.io/log/saved/docker-run-617164bbb06b71cd90d20021126172f5/3a78ff6eeb9f717aa2ca774ed38e01a227c5475a
  + https://ci.ocamllabs.io/log/saved/docker-run-5ba4343bac8c0744f8d39cdd71bdc7e8/db58ee8b88ff888d2f7a688462086d16bfb82f24

- ppx_deriving_morphism

  incorrect myocamlbuild.ml setup, incompatible with ppx_deriving.4.3

  upstream fix https://github.com/choeger/ppx_deriving_morphism/pull/2

  opam-repository fix https://github.com/ocaml/opam-repository/pull/14014

  Build failure logs:
  + https://ci.ocamllabs.io/log/saved/docker-run-6614b63bdcd813869ec8f8c8641347f9/e75b4edd3a2701f41eea773a99abcdb288a63aaf

- ppx_deriving_yojson

  incorrect myocamlbuild.ml setup, incompatible with ppx_deriving.4.3

  opam-repository fix https://github.com/ocaml/opam-repository/pull/14015

  Build failure logs:
  + https://ci.ocamllabs.io/log/saved/docker-run-b2b45318da2088bbbe4ce1951871851e/6e24acd84c36f0e128eb766c742e38506d0fdca1

PR #13655, opened 2019-03-12
nocrypto: use the modern cstruct findlib packages

avsm:
This removes an upper bound which is splitting the opam package universe around nocrypto.

Interim patch until upstream nocrypto has a release (this issue is fixed in trunk), as it's blocking a number of activities in opam-repository. cc @hannesm @copy @pqwy


avsm:
No new failures in revdeps, and approved remotely by email by @hannesm so merging

PR #13660, opened 2019-03-12
Fix all oraclelinux depexts

kit-ty-kate:
The name turned out to be `ol` not `oraclelinux`


avsm:
This used to work in opam 1.2 -- what's the source of all this disagreement in the distro name?

PR #13664, opened 2019-03-12
remove mirage-console-xen-cli

avsm:
This is in the awkward position of not building due to xen revisions
and constantly breaks revdeps. There are no known users, but please
contact @avsm or @djs55 if you do happen to be one.


mseri:
Should we also get rid of `xenctrl`? It builds almost nowhere at the moment...

avsm:
could do -- but it rarely shows up on revdeps so it might be ok to leave it for now too

PR #13674, opened 2019-03-13
Older version of sexplib are not coinstallable with sexplib0.

xclerc:
The `sexplib0` package appeared as a split from `sexplib`
in `v0.11.0`. Previous versions of `sexplib` cannot coexist
with `sexplib0` because both contain a module named
`Sexplib0`.


mseri:
Mixing old sexplib and sexplib0 allows to replicate the issue. This patch seems to solve it, at least on the local test I've made.

mseri:
I think there is not much point in waiting for Travis in this case. I'll merge

xclerc:
@mseri Thanks!

Issue #13636, opened 2019-03-09
conf-pkg-config.1.1 no longer installs on Windows


talex5:
I don't have a Windows system, but my AppVeyor builds have started failing when installing conf-pkg-config.

Example of failing output (from https://ci.appveyor.com/project/talex5-ci/0install-yp593):

```
Augmented Transaction List:
   0 install                       mingw64-x86_64-pkg-config             1.6.0-1
   1 install                             mingw64-x86_64-zlib            1.2.11-1
   2 install                          mingw64-x86_64-nghttp2            1.31.0-1
   3 install                          mingw64-x86_64-openssl            1.0.2o-1
   4 install                          mingw64-x86_64-libssh2             1.7.0-2
   5 install                             mingw64-x86_64-curl            7.59.0-1
Downloaded C:\cygwin64\packages/http%3a%2f%2fcygwin.mirror.constant.com%2f/x86_64/release/pkgconf/mingw64-x86_64-pkg-config/mingw64-x86_64-pkg-config-1.6.0-1.tar.xz
[...]
The following actions will be performed:
  - install conf-which      1
  - install conf-pkg-config 1.1
[...]
Processing  1/36: [conf-pkg-config: pkg-config]
+ C:\cygwin64\home\appveyor\.opam\ocaml-variants.4.07.1+mingw64c\bin\pkg-config.exe "--help" (CWD=C:/cygwin64/home/appveyor/.opam/ocaml-variants.4.07.1+mingw64c/.opam-switch/build/conf-pkg-config.1.1)
+ C:\cygwin64\home\appveyor\.opam\ocaml-variants.4.07.1+mingw64c\bin\pkg-config.exe "--help" (CWD=C:/cygwin64/home/appveyor/.opam/ocaml-variants.4.07.1+mingw64c/.opam-switch/build/conf-pkg-config.1.1)
- pkg-config ist not installed. Install it with 'opam depext conf-pkg-config'
- pkg-config ist not installed. Install it with 'opam depext conf-pkg-config'
[ERROR] The compilation of conf-pkg-config failed at "C:\\cygwin64\\home\\appveyor\\.opam\\ocaml-variants.4.07.1+mingw64c\\bin\\pkg-config.exe --help".
```

Example from an earlier successful run (https://ci.appveyor.com/project/talex5-ci/0install-yp593/builds/22306992):

```
[00:06:45] Augmented Transaction List:
[...]

PR #13675, opened 2019-03-13
Add conf-libX11 package

dra27:
Extracted from #12674, just to shed some weight from the debugging...

### DIFF
--- a/packages/conf-libX11/conf-libX11.1/files/pkg-osx.sh
+++ b/packages/conf-libX11/conf-libX11.1/files/pkg-osx.sh
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+if [ -n "$PKG_CONFIG_PATH" ] ; then
+  PKG_CONFIG_PATH=$PKG_CONFIG_PATH:
+fi
+
+PKG_CONFIG_PATH=$PKG_CONFIG_PATH/usr/local/lib/pkgconfig:/opt/X11/lib/pkgconfig pkg-config x11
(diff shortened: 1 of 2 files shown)


dra27:
The error in Travis is a timeout, so that one failing job could be restarted. Equally, I do appear to have solved what was going on in #12674, and that PR includes in this commit.

mseri:
Thanks!

mseri:
I'll merge when travis finishes

packages/conf-libX11/conf-libX11.1/files/pkg-osx.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+if [ -n "$PKG_CONFIG_PATH" ] ; then
+  PKG_CONFIG_PATH=$PKG_CONFIG_PATH:
+fi
+
+PKG_CONFIG_PATH=$PKG_CONFIG_PATH/usr/local/lib/pkgconfig:/opt/X11/lib/pkgconfig pkg-config x11
```
mseri: Will this work on windows, or it will suffer from the same issue seen here: https://github.com/ocaml/opam-repository/issues/13636 ?

mseri: I guess it will, given that depends on conf-pkg-config

dra27: This script only runs on macOS!

mseri: Oops, indeed :)

PR #13685, opened 2019-03-17
Patch ocamlfind to elimitate the check for awk 

kit-ty-kate:
While testing the new opam-depext on Archlinux I encountered a problem while building `ocamlfind`: `awk` wasn't installed.

This PR adds a `conf-awk` package and adds the dependencies to it in all ocamlfind package as the dependency on awk has been there for at least the last 12 years (https://gitlab.camlcity.org/gerd/lib-findlib/commit/e400c8e6da02cd18343a321325e3d8f3314f19dd)

One thing I'm not sure about is how to name `conf-awk`. Is it fine like this or should we name it `conf-gawk`. I'm not sure of the differences between all the implementations.

### DIFF
--- a/packages/ocamlfind/ocamlfind.1.3.1/files/no-awk-check.patch
+++ b/packages/ocamlfind/ocamlfind.1.3.1/files/no-awk-check.patch
@@ -1,0 +1,19 @@
+commit 40142bc941e6e308686e86be6fc2da92f346a22f
+Author: Kate <kit.ty.kate@disroot.org>
+Date:   Tue Mar 19 16:29:06 2019 +0000
+
+    Remove awk from the set of checked unix tools as it's not used anywhere
+
+diff --git a/configure b/configure
+index d9b587c..20e8dca 100755
+--- a/configure
++++ b/configure
+@@ -184,7 +184,7 @@ echo "Configuring core..."
+ 
+ # Some standard Unix tools must be available:
+ 
+-for tool in sed awk ocaml ocamlc uname rm make cat m4 dirname basename; do
++for tool in sed ocaml ocamlc uname rm make cat m4 dirname basename; do
+     if in_path $tool; then true; else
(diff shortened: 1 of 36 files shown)


kit-ty-kate:
it seems that `gawk` is not installed by default (at least on debian) and `mawk` is used instead..
I'm not sure what to do tbh. I'm not really for adding another mandatory external dependency. Maybe we can just say "install gawk if on archlinux only"?

hannesm:
NB: on my FreeBSD system I have BSD awk:
```
STANDARDS
     The awk utility is compliant with the IEEE Std 1003.1-2008 (“POSIX.1”)
     specification, except awk does not support {n,m} pattern matching.

     The flags [-dV] and [-safe], as well as the commands fflush, compl, and,
     or, xor, lshift, rshift, are extensions to that specification.

HISTORY
     An awk utility appeared in Version 7 AT&T UNIX.
```

which works fine with (my universe of) opam packages. there's no need for an external `gawk`, but it would be great to figure out whether ocamlfind (and other packages) require more than POSIX-specified awk.

more concretely: does `mawk` cause any trouble? is it posix-compliant? should we demand "some posix-compliant awk named `awk`"?

kit-ty-kate:
any awk implementation should be fine for most opam packages. The problem arose only on Archlinux so far where no awk implementation is installed, and maybe on other distributions without any awk. But with the set of distributions we had on the CI, it worked at least

kit-ty-kate:
I just found out that `awk` is not used anyway in `ocamlfind` and wasn't used for something like 13-14 years. It's just checked for nothing. I might just do a patch instead, applied to all ocamlfind versions.

kit-ty-kate:
Merge request in `ocamlfind` here: https://gitlab.camlcity.org/gerd/lib-findlib/merge_requests/23

mseri:
Thanks!

PR #13688, opened 2019-03-18
containers is not compatible with OCaml 4.08

kit-ty-kate:

cc @c-cube

### DIFF
--- a/packages/containers/containers.0.12/opam
+++ b/packages/containers/containers.0.12/opam
@@ -28,7 +28,7 @@ remove: [
     ["ocamlfind" "remove" "containers"]
 ]
 depends: [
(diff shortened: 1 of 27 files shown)


c-cube:
yes, I'll make a release when 4.08 is out. I blame it all on the tiny use of compiler-libs to generate unlabelled files :(

kit-ty-kate:
the main error I've seen in the logs is the conflict between `Stdlib.List.t` and `CCList.t`

c-cube:
ah yes, that too. Anyway https://github.com/c-cube/ocaml-containers/tree/wip-4.08 should build and I'll merge it when 4.08 is released.

PR #13689, opened 2019-03-18
ocaml-migrate-parsetree is not compatible with OCaml 4.08

kit-ty-kate:

cc @let-def


let-def:
wip https://github.com/ocaml-ppx/ocaml-migrate-parsetree/issues/50

PR #13690, opened 2019-03-18
lwt is not compatible with OCaml 4.08

kit-ty-kate:

cc @aantron


aantron:
Thanks @kit-ty-kate. We are tracking the actual compatibility in https://github.com/ocsigen/lwt/issues/658. It should be doable now that `mmap` is merged, just have to actually make the change in Lwt.

PR #13693, opened 2019-03-18
Add Cygwin depext for pkg-config and fftw3

Chris00:


mseri:
We can’t really test this here. Let’s merge. Thanks!

PR #13700, opened 2019-03-19
Package easy_logging.0.1

sapristi:
### `easy_logging.0.1`
Module to easily log messages
Logging infrastructure inspired by the Python logging module.
The aim of this module is to provide a quick and easy to use logging
infrastructure.

It has the following features :
   * one line logger creation
   * log messages are either [string] or [string lazy_t]
   * log level adaptable at runtime from anywhere in the program
   * handlers associated to each logger will format and treat the message independantly.

---
* Homepage: https://sapristi.github.io/easy_logging/easy_logging/
* Source repo: git+https://github.com/sapristi/easy_logging.git
* Bug tracker: https://github.com/sapristi/easy_logging/issues

---

### DIFF
--- a/packages/easy_logging/easy_logging.0.1/opam
+++ b/packages/easy_logging/easy_logging.0.1/opam
@@ -1,0 +1,37 @@
+opam-version: "2.0"
+synopsis: "Module to easily log messages"
+
+maintainer: "mathiasmillet@gmail.com"
+authors: "Mathias Millet"
+license: "GPL"
+homepage: "https://sapristi.github.io/easy_logging/easy_logging/"
+bug-reports: "https://github.com/sapristi/easy_logging/issues"
+dev-repo: "git+https://github.com/sapristi/easy_logging.git"
+depends: [
+  "ocaml" {>= "4.04.0"}
+  "dune" {build & >= "1.0"}
+  "ppx_deriving"{>= "4.0" & < "5.0"}
+]
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name] {with-test}
+]
+
+description: """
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Your package doesn't have a `url` section and thus doesn't know where to get the archive from. Something went probably wrong when you used `dune-release` I suppose.

packages/easy_logging/easy_logging.0.1/opam
```diff
@@ -0,0 +1,23 @@
...
+   * log level adaptable at runtime from anywhere in the program
+   * handlers associated to each logger will format and treat the message independantly
+"""
+maintainer: "mathiasmillet@gmail.com"
+authors: "Mathias Millet"
+license: "GPL"
+homepage: "https://sapristi.github.io/easy_logging/easy_logging/"
+bug-reports: "https://github.com/sapristi/easy_logging/issues"
+dev-repo: "git+https://github.com/sapristi/easy_logging.git"
+depends: [ "ocaml" "dune" "ppx_deriving"]
```
kit-ty-kate: dune needs to be tagged as a `{build}` dependency. Moreover, one advice I could give is that it would be more readable if the dependencies were indented.

kit-ty-kate: I also see that your `dune` file in the source repository uses `batteries` which is not listed in the dependencies

sapristi: Thanks, I'll fix that.
I remove batteries from the code but forgot about dune files.

sapristi:
Please do not merge now since I still have some modifications to do.

mseri:
Looks good, thanks!

Feel free to announce it on discuss.ocaml.org, there is a `community` category and an `announce` tag for this purpose.

PR #13703, opened 2019-03-19
+ qfs 0.9

ygrek:
CI build is expected to fail due to absent libqfs

### DIFF
--- a/packages/qfs/qfs.0.9/opam
+++ b/packages/qfs/qfs.0.9/opam
@@ -7,8 +7,7 @@ dev-repo: "git://github.com/ahrefs/ocaml-qfs.git"
 bug-reports: "https://github.com/ahrefs/ocaml-qfs/issues"
 tags: [ "org:ahrefs" "clib:stdc" "clib:qfs"  ]
 build: [
-  ["ocaml" "setup.ml" "-configure" "--prefix" prefix]
-  ["ocaml" "setup.ml" "-configure" "--prefix" prefix "--enable-tests"] {with-test}
+  ["ocaml" "setup.ml" "-configure" "--prefix" prefix "--enable-tests" {with-test}]
   ["ocaml" "setup.ml" "-build"]
   ["ocaml" "setup.ml" "-test"] {with-test}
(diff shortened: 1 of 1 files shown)


packages/qfs/qfs.0.9/opam
```diff
@@ -0,0 +1,47 @@
...
+license: "LGPL-2.1 with OCaml linking exception"
+homepage: "https://github.com/ahrefs/ocaml-qfs"
+dev-repo: "git://github.com/ahrefs/ocaml-qfs.git"
+bug-reports: "https://github.com/ahrefs/ocaml-qfs/issues"
+tags: [ "org:ahrefs" "clib:stdc" "clib:qfs"  ]
+build: [
+  ["ocaml" "setup.ml" "-configure" "--prefix" prefix]
+  ["ocaml" "setup.ml" "-build"]
+  ["ocaml" "setup.ml" "-configure" "--enable-tests"] {with-test}
+  ["ocaml" "setup.ml" "-build"] {with-test}
```
kit-ty-kate: the build and configure steps are going to be done twice. I don't think this is what you wanted to do. I would suggest to place the `{with-test}` attribute to the argument for the configure and to omit the second build command line as it is redundant with the first one which is done unconditionally

packages/qfs/qfs.0.9/opam
```diff
@@ -0,0 +1,47 @@
...
+  ["ocaml" "setup.ml" "-test"] {with-test}
+  ["ocaml" "setup.ml" "-doc"] {with-doc}
+]
+install: ["ocaml" "setup.ml" "-install"]
+remove: [
```
kit-ty-kate: the remove field can be omitted with opam2

ygrek:
thanks, fixed

kit-ty-kate:
> I would suggest to place the {with-test} attribute to the argument for the configure and to omit the second build command line as it is redundant with the first one which is done unconditionally

sorry, what I meant about this comment is my latest commit

kit-ty-kate:
It should be ok now. Thanks!

ygrek:
I didn't know one could do that, nice!

PR #13704, opened 2019-03-19
[new release] mindstorm and mindstorm-lwt (0.8)

Chris00:
CHANGES:

- In addition to the package `mindstorm`, a version of the interface
  using LWT promises has been developed in the package `mindstorm-lwt`.
- LWT: transactions with the LEGO© brick are atomic.
- Use [dune][] and `dune.configurator`.
- Improve the documentation

[dune]: https://github.com/ocaml/dune

### DIFF
--- a/packages/mindstorm-lwt/mindstorm-lwt.0.8/opam
+++ b/packages/mindstorm-lwt/mindstorm-lwt.0.8/opam
@@ -1,0 +1,39 @@
+opam-version: "2.0"
+maintainer: "Christophe Troestler <Christophe.Troestler@umons.ac.be>"
+authors: [ "Christophe Troestler"
+           "Julie De Pril"
+           "Marc Ducobu"
+           "Dany Maslowski" ]
+license: "LGPL-2.1 with OCaml linking exception"
+homepage: "https://github.com/Chris00/ocaml-mindstorm"
+dev-repo: "git+https://github.com/Chris00/ocaml-mindstorm.git"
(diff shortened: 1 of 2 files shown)


Chris00:
`bluetooth/bluetooth.h` is missing on opensuse.

Chris00:
I do not understand why given that `bluez-devel` is [supposed to provide `bluetooth/bluetooth.h`](http://www.rpmfind.net/linux/rpm2html/search.php?query=%2Fusr%2Finclude%2Fbluetooth%2Fbluetooth.h&submit=Search+...&system=&arch=)

mseri:
It's not the first time I see this kind of errors on opensuse. It would be useful to have a look at pkg-config output to see if we need additional flags in that case

mseri:
I would not consider this a reason not to merge unless the fix is immediate

kit-ty-kate:
That should fix the issue. That said, it would be a lot better with a conf-libbluez or something like that to avoid redundancy.

mseri:
I keep forgetting about that difference for suse... we should probably write down clearly these subtleties somewhere visible (maybe even a secion on the opam manual)

Chris00:
>  it would be a lot better with a conf-libbluez or something like that 

Sure.  Will try to set up one now.

Chris00:
Once [conf-bluetooth](https://github.com/ocaml/opam-repository/pull/13708) is merged, I'll update this PR.

mseri:
Thanks

PR #13710, opened 2019-03-20
Update conf-cairo for macos with homebrew

Chris00:


Chris00:
Alternative suggestions from Mac users welcome.

Chris00:
[The problem](https://travis-ci.org/ocaml/opam-repository/jobs/508885854#L1324) this PR is trying to solve.

Chris00:
Could the original problem be with the sandboxing?

Chris00:
[cairo seems to be installed](https://travis-ci.org/ocaml/opam-repository/jobs/509194532#L1152) — even more than once.  This may be the problem.

mseri:
I incurred in this sometimes last year, on my mac I was able to install by passing some paths to homebrew via shell variables before invoking opam: I believe it was due to some conficts with pango though.

I cannot recover what I did from the history any longer and I don't know if they are still relevant in the latest homebrew given that it can see the pc files. Note that these
```
- /usr/local//Cellar/cairo/1.16.0/lib/pkgconfig/cairo.pc
- /usr/local//lib/pkgconfig/cairo.pc
```
are actually the same file (one is the symlink of the other).

Chris00:
@mseri Thanks for your input.  If you eventually remember how to make it work, please chime in!

Chris00:
I tried
```
  ["sh" "-exc"
   "export PKG_CONFIG_PATH=/usr/local/lib/pkgconfig/; pkg-config --libs cairo"]
```
and it seems to work.

Chris00:
The current [solution](https://github.com/ocaml/opam-repository/pull/13710/files#diff-a9b25a3ce4019caedaa48708ecc7cb3dR8) should pass the tests.  It remains to then be seen whether `dune.configurator` also needs a tweak to detect _cairo_.

Chris00:
For some reason, without the [`--libs`](https://travis-ci.org/ocaml/opam-repository/jobs/509543046#L1127), it [does not work](https://travis-ci.org/Chris00/opam-repository/jobs/509495170#L1145)...

Chris00:
@mseri Does that solution seem acceptable to you?

mseri:
Funny that it doesn’t work without `—libs`. Thanks for the effort, my only complain is that it does not work if you have nix or a non-linked installation, but we can try to fix it in that case if and once the problem arises

mseri:
Thanks for the help

Chris00:
> it does not work if you have a non-linked installation

Not sure what you mean.  Doesn't `brew` automatically make the symlink?

mseri:
I was thinking about my setup with how few but Cairo not linked due to some conflicts with pango. But it is not a big deal, just a corner case, and I no longer have that setup so it may have just been a case

PR #13712, opened 2019-03-20
Package easy_logging.0.1

sapristi:
### `easy_logging.0.1`
Module to easily log messages
Logging infrastructure inspired by the Python logging module.
The aim of this module is to provide a quick and easy to use logging
infrastructure.

It has the following features :
   * one line logger creation
   * log messages are either [string] or [string lazy_t]
   * log level adaptable at runtime from anywhere in the program
   * handlers associated to each logger will format and treat the message independantly

---
* Homepage: https://sapristi.github.io/easy_logging/easy_logging/
* Source repo: git+https://github.com/sapristi/easy_logging.git
* Bug tracker: https://github.com/sapristi/easy_logging/issues

---

### DIFF
--- a/packages/prepare/prepare.0.1/opam
+++ b/packages/prepare/prepare.0.1/opam
@@ -1,0 +1,34 @@
+opam-version: "2.0"
+synopsis: "Module to easily log messages"
+description: """
+     Logging infrastructure inspired by the Python logging module.
+The aim of this module is to provide a quick and easy to use logging
+infrastructure.
+
+It has the following features :
+   * one line logger creation
+   * log messages are either [string] or [string lazy_t]
+   * log level adaptable at runtime from anywhere in the program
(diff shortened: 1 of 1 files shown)


mseri:
It inferred the wrong name (`prepare` instead of `easy_logging`)

sapristi:
As mseri saw,  I was lost in opam-publish varying documentations, and misused opam-publish command.
The package should be named easy_logging.

sapristi:
Ok so I changed the PR messages, but I have no idea if that is enough. Should I wait for actions in this PR or should I make a new one ?

mseri:
You can force push on this PR adding the necessary fixes, but do as you find more convenient

sapristi:
Ok so it seems putting in the right name pushes to the other PR, si I guess this one should be closed. Thank you mseri !

PR #13713, opened 2019-03-21
+ devkit 1.0

cyberhuman:


cyberhuman:
There is some issue with libevent on OpenSuse :-(
Also some camlp4 issue in travis.

ygrek:
camlp4 problem same as observed before - for some reason different camlp4 version is installed with apt and takes over the one from opam
 https://travis-ci.org/ocaml/opam-repository/jobs/509218730#L288

mseri:
I have sent a PR to fix opensuse, it was missing the depext for libevent. I am more worried about the TravisCI failure

mseri:
In the case of ocaml 4.03 and 4.04, the main failure seems to be due to devkit.0.4 dependencies, maybe some upper bound or some dependency is incorrect there.

The 4.06 and 4.07 system compiler case is clear, there is a conflict between opam and the system.camlp4.

In the case of ocaml 4.05 and 4.06 (and partly also in 4.03 and 4.04), it is extunix.0.1.6 the culprit, having issues finding camlp4:
```
#=== ERROR while compiling extunix.0.1.6 ======================================#
# context     2.0.0 | linux/x86_64 | ocaml-base-compiler.4.07.1 | file:///home/travis/build/ocaml/opam-repository
# path        ~/.opam/ocaml-base-compiler.4.07.1/.opam-switch/build/extunix.0.1.6
# command     ~/.opam/opam-init/hooks/sandbox.sh build ocaml setup.ml -build
# exit-code   1
# env-file    ~/.opam/log/extunix-12903-bc35d5.env
# output-file ~/.opam/log/extunix-12903-bc35d5.out
### output ###
# [...]
# File "src/pa_have.ml", line 21, characters 6-38:
# While finding quotation "ctyp" in a position of "patt":
#   There is no quotation expander available.
#   Camlp4: Uncaught exception: Not_found
# 
# 
# File "src/pa_have.ml", line 1:
# Error: Error while running external preprocessor
# Command line: camlp4 -nolib '-I' '/home/travis/.opam/ocaml-base-compiler.4.07.1/lib/camlp4' '-I' '/home/travis/.opam/ocaml-base-compiler.4.07.1/lib/camlp4' 'pa_o.cmo' 'pa_op.cmo' 'pr_dump.cmo' 'pa_extend.cmo'  'src/pa_have.ml' > /tmp/ocamlppaba21f
# 
# Command exited with code 2.
# E: Failure("Command ''/home/travis/.opam/ocaml-base-compiler.4.07.1/bin/ocamlbuild' src/libextunix_stubs.a src/dllextunix_stubs.so src/extunix.cma src/extunix.cmxa src/extunix.a src/extunix.cmxs -tag debug' terminated with error code 10")
```

ygrek:
eh, I don't think it is extunix fault, in all cases it is conflict between system and opam camlp4, e.g. https://travis-ci.org/ocaml/opam-repository/jobs/509218732#L287

Arguably extunix could try harder to use the opam one, but I think we agree that compiling in presence of two different compilers in path is "best effort" at best.

cyberhuman:
I think it should not be built for 4.03 and 4.04 at all, because
```The following dependencies couldn't be met:
  - devkit → ocaml >= 4.05.0
```
Also it fails to build lwt, so it's not extunix fault for sure.

ygrek:
> In the case of ocaml 4.03 and 4.04, the main failure seems to be due to devkit.0.4 dependencies

I don't see it. The failure is the same camlp4 problem, witness pa_lwt build failure in same log.

mseri:
> I don't see it. The failure is the same camlp4 problem, witness pa_lwt build failure in same log.

Yes you are right. I disagree a bit on extunix: I would understand failing in the case of system+opam camlp4, but it should not fail trying to use a camlp4 that has nothing to do with the ocaml version in use

mseri:
I am going to merge because this is not a new problem and I don't think we can do anything as part of this PR. 

Maybe camlp4 could be split out as a separate devkit subpackage, to ease up installation (especially given that camlp4 is sort of deprecated in favour of ppx at this point).

cyberhuman:
Thank you!

ygrek:
> I would understand failing in the case of system+opam camlp4, but it should not fail trying to use a camlp4 that has nothing to do with the ocaml version in use

I don't get it. It is what happens, system camp4 of different version is present in path and opam camlp4 too, and apparently system camlp4 is first in PATH, so there is not much extunix can do (it could call camlp4 with full path, but I claim that it should not too)

> Maybe camlp4 could be split out as a separate devkit subpackage, to ease up installation (especially given that camlp4 is sort of deprecated in favour of ppx at this point).

it is likely we will get rid of it at some point, but it is not obvious how to do it for extunix.

PR #13722, opened 2019-03-22
[DRAFT] ocaml-migrate-parsetree for OCaml 4.08.0

xclerc:
This version is expected to be used with the current beta
(*i.e.* ocaml-variants.4.08.0+beta2), another version will
be needed due to https://github.com/ocaml/ocaml/pull/8535.

### DIFF
--- a/packages/ocaml-migrate-parsetree/ocaml-migrate-parsetree.1.3.0/opam
+++ b/packages/ocaml-migrate-parsetree/ocaml-migrate-parsetree.1.3.0/opam
@@ -1,0 +1,33 @@
+opam-version: "2.0"
+maintainer: "frederic.bour@lakaban.net"
+authors: [
+  "Frédéric Bour <frederic.bour@lakaban.net>"
+  "Jérémie Dimino <jeremie@dimino.org>"
+]
+license: "LGPL-2.1"
+homepage: "https://github.com/ocaml-ppx/ocaml-migrate-parsetree"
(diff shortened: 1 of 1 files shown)


avsm:
awesome! i am ok with merging this once ci passes to unblock bulk testing. Thanks very much for the hard work on OMP porting @Armael @let-def @xclerc and anyone else I've missed :-)

mseri:
If this is only for `ocaml-variants.4.08.0+beta2` should we be stricter in the depends?

xclerc:
@mseri  Sure; is `"ocaml" {= "4.08.0+beta2"}` a valid dependency
specification?
Also, how can I access the camelus log to see what it is complaining
about?

mseri:
I have no idea about camelus, but I think it complaints because there is no compiler that satisifies >= 4.08 and < 4.09.

xclerc:
Indeed, that makes sense!

mseri:
I think the specification you suggest should be right but I'd like to ping @avsm to double check

avsm:
I'm not sure -- it might have to be ocaml-variants.4.08.0+beta2. @AltGr or @dra27 would probably know for sure :)

In the meanwhile, I'll merge this since it unblocks vast tracts of 4.08 packages for @kit-ty-kate to run opam-check-all over

damiendoligez:
A bit of testing reveals that you need to put the constraint on `ocaml-variants`. I've opened #13870 to that effect.

PR #13723, opened 2019-03-22
Add missing bound on the cmdliner->topkg dependency

damiendoligez:
This is needed because these versions of `cmdliner` use `Pkg.Flatten`.


yomimono:
Failures look spurious and this is clearly an improvement. Thanks, @damiendoligez !

PR #13733, opened 2019-03-23
Travis - always use latest release of opam 2.0

dra27:
After https://github.com/ocaml/ocaml-ci-scripts/pull/274/ was merged, the matrix should have been updated to specify `2.0`, rather than specifically `2.0.0`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,23 +8,23 @@ matrix:
   include:
   - os: osx
     osx_image: xcode10.1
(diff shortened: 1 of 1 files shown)


mseri:
Thanks!!!

dra27:
No problem - see also https://github.com/ocaml/ocaml-ci-scripts/pull/287 which would add a warning for if this happens in others' CI

PR #13734, opened 2019-03-23
+ 0install.2.14

talex5:


talex5:
- Tests failed under Travis/macOS because the CI sandboxing is preventing (local) network tests there.
- SUSE seems to have failed to install its own curl package (`Problem: libcurl-devel-7.59.0-lp150.1.1.x86_64 requires libcurl4 = 7.59.0, but this requirement cannot be provided`).

mseri:
Looks good, thanks

PR #13736, opened 2019-03-23
sexplib: older sexplib should conflict with sexplib0

avsm:
otherwise sexplib0.v0.12.0 is getting selected with
sexplib.113.33.00+4.03 on ocaml 4.04 switches, with the
results not going well...

cc @xclerc


xclerc:
Thanks!

PR #13739, opened 2019-03-24
New package iso639.0.0.4

paurkedal:
This is a new package.  It provides a simple API to data files from the ISO-639 standard for language codes.  It's primary purpose is to supply a canonical representation of a language and conversions to and from 2- and 3-letter language codes.  The data files are packaged in the release tarballs.

API: https://paurkedal.github.io/ocaml-iso639/iso639/Iso639/index.html

### DIFF
--- a/packages/iso639/iso639.0.0.4/opam
+++ b/packages/iso639/iso639.0.0.4/opam
@@ -1,0 +1,29 @@
+opam-version: "2.0"
+name: "iso639"
+maintainer: "Petter A. Urkedal <paurkedal@gmail.com>"
+authors: "Petter A. Urkedal <paurkedal@gmail.com>"
+license: "LGPL-3 with OCaml linking exception"
+homepage: "https://github.com/paurkedal/ocaml-iso639/"
+doc: "http://paurkedal.github.io/ocaml-iso639/index.html"
+bug-reports: "https://github.com/paurkedal/ocaml-iso639/issues"
+depends: [
+  "ocaml" {>= "4.04.2"}
+  "dune" {build}
+]
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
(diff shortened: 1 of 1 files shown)


paurkedal:
Seems this needs to work.

mseri:
Ok, then I will not merge for now. Let me know once it's ready.

packages/iso639/iso639.0.0.4/opam
```diff
@@ -0,0 +1,30 @@
...
+maintainer: "Petter A. Urkedal <paurkedal@gmail.com>"
+authors: "Petter A. Urkedal <paurkedal@gmail.com>"
+license: "LGPL-3 with OCaml linking exception"
+homepage: "https://github.com/paurkedal/ocaml-iso639/"
+doc: "http://paurkedal.github.io/ocaml-iso639/index.html"
+bug-reports: "https://github.com/paurkedal/ocaml-iso639/issues"
+depends: [
+  "ocaml" {>= "4.04.2"}
+  "dune" {build}
+  "ocamlfind" {build}
```
kit-ty-kate:
```suggestion
```
you don't seem to use ocamlfind anywhere in this package

paurkedal: I'll remove it. Thanks!

paurkedal:
I suggest merging now, even though the former issue is not understood:

There was an assert failure in one of the travis runs, then one with `INSTALL_LOCAL=1`.  This being an isolated failure already seemed a bit odd, since the test is deterministic, single-threaded, and the  package is self-contained OCaml code.  I could not reproduce it in a local docker container, though I could only approximate the environment as I didn't see which docker image the travis build used, and resorted to a generic Ubuntu 16.04.  After pushing the opam update, the error went away.  I think the best I can do at the moment is to add more informative output to the test, in case the issue reoccurs in a future build.

PR #13745, opened 2019-03-25
Cstruct works with Jane Street 0.12 libs

XVilka:
Apparently the restriction is not required, wasn't updated after the 0.12 libraries release.
See https://github.com/mirage/ocaml-cstruct/pull/243


XVilka:
Note that failing Travis is due to its network connectivity issues.

mseri:
Thanks!

PR #13746, opened 2019-03-25
Add two variations for BuckleScript patched compiler

bobzhang:
It would help track our patches would not break native stuff, thanks!

### DIFF
--- a/packages/ocaml-variants/ocaml-variants.4.02.3+BS/opam
+++ b/packages/ocaml-variants/ocaml-variants.4.02.3+BS/opam
@@ -1,0 +1,32 @@
+opam-version: "2.0"
+synopsis: "4.02.3+BuckleScript patches"
+maintainer: "platform@lists.ocaml.org"
+depends: [
+  "ocaml" {= "4.02.3" & post}
+  "base-unix" {post}
+  "base-bigarray" {post}
+  "base-threads" {post}
+  "base-ocamlbuild" {post}
+]
+conflict-class: "ocaml-core-compiler"
+flags: compiler
+setenv: CAML_LD_LIBRARY_PATH = "%{lib}%/stublibs"
+build: [
+  ["./configure" "-prefix" prefix "-with-debug-runtime"]
+    {os != "openbsd" & os != "freebsd" & os != "macos"}
+  [
+    "./configure"
+    "-prefix"
+    prefix
+    "-with-debug-runtime"
+    "-cc"
+    "cc"
+    "-aspp"
+    "cc -c"
+  ] {os = "openbsd" | os = "freebsd" | os = "macos"}
+  [make "world.opt"]
+]
+install: [make "install"]
+url {
+  src: "https://github.com/bucklescript/ocaml/archive/4.02.3+BS.zip"
+}
--- a/packages/ocaml-variants/ocaml-variants.4.02.3+BS/opam
+++ b/packages/ocaml-variants/ocaml-variants.4.02.3+BS/opam
@@ -28,5 +28,5 @@ build: [
 ]
 install: [make "install"]
 url {
-  src: "https://github.com/bucklescript/ocaml/archive/4.02.3+BS.zip"
+  src: "https://github.com/bucklescript/ocaml/archive/4.02.3+BS.tar.gz"
 }
--- a/packages/ocaml-variants/ocaml-variants.4.06.1+BS/opam
+++ b/packages/ocaml-variants/ocaml-variants.4.06.1+BS/opam
@@ -28,5 +28,5 @@ build: [
 ]
 install: [make "install"]
 url {
-  src: "https://github.com/bucklescript/ocaml/archive/4.06.1+BS.zip"
+  src: "https://github.com/bucklescript/ocaml/archive/4.06.1+BS.tar.gz"
 }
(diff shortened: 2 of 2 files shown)


packages/ocaml-variants/ocaml-variants.4.02.3+BS/opam
```diff
@@ -0,0 +1,32 @@
...
+    "-cc"
+    "cc"
+    "-aspp"
+    "cc -c"
+  ] {os = "openbsd" | os = "freebsd" | os = "macos"}
+  [make "world.opt"]
+]
+install: [make "install"]
+url {
+  src: "https://github.com/bucklescript/ocaml/archive/4.02.3+BS.zip"
```
kit-ty-kate: is this release going to change? If it doesn't you need to add a checksum to the archive. Also maybe tar.gz would be a better choice of archive.

bobzhang:
changed to .tar.gz.
Note that 4.02.3+BS is supposed to not change that much, but it may have changes occasionally

avsm:
It does need to have an archive and a checksum for inclusion here.  If you're changing the archive, we recommend bumping a minor version number (e.g. `4.02.3+BS1`) to mark the change and submit a new version.

Altering upstream source archives once published makes it difficult to cache sources, and to distinguish between malicious changes (due to the server being compromised) and author-intended changes, hence this request.

bobzhang:
The 4.02.3+BS or 4.06.1+BS is only used for BuckleScript devs (not for general user land consumer).
I used to use compiler-conf but it is broken after opam2 upgrade.
If we add a checksum for each change, that would defeat the purpose of easy testing for our patches. The testing cycle would be much longer

hcarty:
If these are not intended for general users, would it be better to have them in a separate repository?

bobzhang:
Indeed, I am asking for a favor since compiler-conf is broken for opam2 and my previous opam variant (4.02.3+buckle-master) was merged(it works).
If it would takes lots of effort for each change, that would defeat the purpose of this PR and it may cause some inconvenience for contributors to BuckleScript.

bobzhang:
I tried to add my own repo, but failed:
```
opam-repository>opam repo add buckle https://github.com/bobzhang/opam-repository
[ERROR] Could not update repository "buckle": OpamDownload.Download_fail(_, "Download command failed: \"/usr/local/bin/wget
        --content-disposition -t 3 -O
        /private/var/folders/hj/ymj_3np92svgkr06tb511qq80000gn/T/opam-2660-5a61c5/index.tar.gz.part
        https://github.com/bobzhang/opam-repository/index.tar.gz -U opam/2.0.3\" exited with code 8 \"2019-03-29 10:56:53 ERROR
        404: Not Found.\"")
```

I added a local version:
```
opam-repository>opam repo add buckle .
[buckle] Initialised
[NOTE] Repository buckle has been added to the selections of switch 4.06.2+local-git-4.06 only.
       Run `opam repository add buckle --all-switches|--set-default' to use it in all existing switches, or in newly created
       switches, respectively.
```
```
opam-repository>opam repository add buckle --all-switches
```
```
opam-repository>opam switch list-available | grep BS
ocaml-variants      4.01.0+armv6-freebsd                   OCaml 4.01.0 with patches for native code compilation on armv6 FreeBSD10-RELEASE (ie raspberrypi)
```
Could not find it, so my last resort is to downgrade opam...

mseri:
@bobzhang To point to a Github repo you need to use git:// or git+https://

The repo needs to be added to the active switch (or all). There is a flag, I will let you know which as soon as I am at the laptop

avsm:
Sorry, I hadn't noticed that these are compiler variants, and not packages. I think there's precedent for including these with git:// urls then, although it does make archive caching a bit more of a pain from our end.  If you submit the PR but with the url being a `git+https://<...>` to the source repo, then it can be merged.  That ensures that opam will correctly pick up any commits to the remote tree and rebuild packages appropriately.

avsm:
@bobzhang in general, when you submit these requests that a little non-standard, it would be extremely helpful if you included some more context.  I had to read quite far down this thread to finally figure out that you were using compiler-conf before, and then most of the conversation happened over a weekend...

bobzhang:
@mseri I used git+https://, the opam repo add works and did `opam repository add buckle --all-switches`, but `opam switch list-available` does not show the `+BS` ocaml variant.
If I can manage to do this in my own git repo, I can set up a local repo so that other people can do the switch too.

@avsm 
> If you submit the PR but with the url being a git+https://<...> to the source repo, then it can be merged. 

What would be the branch then, (sorry for my poor knowledge about the git protocol), would `git+https://github.com/bucklescript/ocaml/archive/4.02.3+BS.tar.gz`
`git+https://github.com/bucklescript/ocaml/archive/4.06.1+BS.tar.gz`
work? thanks

mseri:
It does not show even after running `opam update`?

For the git branch it should work by adding #branchname to the repo url

bobzhang:
I tried to learn something from the recent PR 
https://github.com/ocaml/opam-repository/pull/14172/files
It looks similar to what I had here, can we revisit this PR?
Note if there is an easy way to set up my local opam repo, I am happy to do that, but it seems not that easy, see my previous work above

mseri:
This is just my opinion, but if you add a dependency to the ocaml-beta virtual package as done there, I wouldn’t see a problem with it

kit-ty-kate:
@bobzhang in which way is it similar? beta{1,2,3} and rc1 are fixed packages and their archive is attributed to a git tag, not a git branch as it is in your case.
There was a missing checksum for those packages but it will be fixed by https://github.com/ocaml/opam-repository/pull/14176.

XVilka:
Should be this closed then? Or updated instead?

bobzhang:
This is not relevant any more. But it still makes sense for opam to provide a way for people to use their own customized compiler directly

PR #13750, opened 2019-03-26
Disable camlp4 detection on non-system switches

dra27:
ocamlfind will install camlp4 `META` files if it detects `camlp4` in `PATH`, but includes no detection for whether the `camlp4` binary corresponds to the compiler.

This is a problem for OCaml 4.02.0+, where `camlp4` ceased to be part of the distribution as it means that a system-installed `camlp4` gets registered, but won't work.

The change here is twofold:

1. Require ocamlfind 1.5 or greater on OCaml 4.02.0+ (this is when the `-no-camlp4` configure-time option was introduced)
2. For non-system switches, use `-no-camlp4` to prevent ocamlfind's detection of `camlp4`.

### DIFF
--- a/packages/ocamlfind/ocamlfind.1.4.0/opam
+++ b/packages/ocamlfind/ocamlfind.1.4.0/opam
@@ -22,7 +22,7 @@ build: [
   [make "opt"]
 ]
 depends: [
-  "ocaml" {>= "3.08" & < "4.07"}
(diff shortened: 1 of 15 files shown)


mseri:
The failures in the CI seem to bre fine ones. I am going to release a bit the CI and merge

PR #13751, opened 2019-03-26
Relax constraints on 'ocaml-system' versions

AltGr:
For use with custom compilers. See https://github.com/AltGr/ocaml-system for details

### DIFF
--- a/packages/ocaml/ocaml.3.08.0/opam
+++ b/packages/ocaml/ocaml.3.08.0/opam
@@ -8,7 +8,7 @@ depends: [
   "ocaml-config"
   "ocaml-base-compiler" {= "3.08.0"} |
   "ocaml-variants" {>= "3.08.0" & < "3.08.1~"} |
-  "ocaml-system" {= "3.08.0"}
+  "ocaml-system" {>= "3.08.0" & < "3.08.1~"}
 ]
 setenv: [
   [CAML_LD_LIBRARY_PATH = "%{_:stubsdir}%"]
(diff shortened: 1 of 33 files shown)


rjbou:
Need to be propagated to ocaml-beta-repository?

AltGr:
Ah, the newer `ocaml-system` versions were migrated there too ?
I don't see why it would be a problem for these to stay in the main repo (together with the newer versions of the `ocaml` package); only the `ocaml-variants` package is problematic.

rjbou:
ping @avsm @gasche

gasche:
I'm not sure I understand the full context (but @AltGr tool looks like a potential successor for my [opam-compiler-conf](https://github.com/gasche/opam-compiler-conf/) script which has sadly not survived the opam2 migration... nice!), but let me give this a try.

The problem that `ocaml-beta-repository` is designed to solve is that, before it existed, unsuspecting users starting from an empty switch would have opam install not-yet-released versions of the compiler to minimize version lag. The main question is thus whether this user-experience failure can also occur with the `ocaml-system` package. I would guess "no", which goes in @AltGr's direction of suggesting to keep bleeding-edge `ocaml-system` versions in the main repository for convenience.

That said, I'm not convinced that this is the best choice, because it makes the administration of the repositories slightly harder, by making things less homogeneous: now we have to remembre which things should go into the separate repo and which shouldn't, whereas before it was just "all unreleased versions go to -beta". I like the simplicity of the previous policy.

Personally I would say that it is a good compromise to keep dev-version ocaml-system packages in the -beta repository. It adds an extra step for expert users that want to use Louis' `ocaml-system` trick, but those are expert users anyway, in the process of testing a development version. It makes life simpler for the maintainers, and people writing PRs to those repositories when a release or release-branch event happens, etc., and those people don't want to think about ocaml-system at this time.

avsm:
> I don't see why it would be a problem for these to stay in the main repo

iirc, this is because we would then have an uninstallable package in opam-repository.  If that's not the case, I have no problem with it being in the main repo. (I haven't had a chance to look at this PR yet)

avsm:
This is destroying CI with thousands of revdeps tests but seems safe to merge, so doing so now

PR #13748, opened 2019-03-25
[new release] cstruct-async, ppx_cstruct, cstruct, cstruct-unix, cstruct-sexp and cstruct-lwt (4.0.0)

avsm:
Access C-like structures directly from OCaml

- Project page: <a href="https://github.com/mirage/ocaml-cstruct">https://github.com/mirage/ocaml-cstruct</a>
- Documentation: <a href="https://mirage.github.io/ocaml-cstruct/">https://mirage.github.io/ocaml-cstruct/</a>

##### CHANGES:

- Sexplib is now an optional library for the base `Cstruct` module.
  A new `Cstruct_sexp` module has been introduced with the serialiser
  functions, contained within the `cstruct-sexp` opam package.

  To convert old code, simply use `Cstruct_sexp.t` instead of
  `Cstruct.t` in a record type for which you are using `[@@deriving sexp]`.
  This is a type alias to `Cstruct.t` but also has the right
  sexp-conversion functions in scope.  There is an example of this
  in the `ppx_test/with-sexp` directory in the source repo.

  When you have converted and released your library, add an
  opam constraint of `cstruct {>="4.0.0"}` to your own opam
  packages to ensure that they pick up this version of the library.
  (fixes mirage/ocaml-cstruct#222, @avsm)

- JavaScript stubs are now installed using the jsoo mechanism
  rather than a manual specification (mirage/ocaml-cstruct#241 @jonludlam)

- Use computed versions in opam files to ensure that dependent
  opam packages such as cstruct-async get the same base version
  of cstruct to avoid mismatches. (@avsm)

- Add a ppx test suite to checks that all error paths in ppx
  handling are reachable and have meaningful location info (mirage/ocaml-cstruct#238 @emillon)

PR #13753, opened 2019-03-26
add upper bounds on cstruct.4.0.0 due to sexplib removal

avsm:
see #13748


mseri:
I went through all of the.  It cannot make things worse, if anything by merging we can reduce the ci load a bit

avsm:
thanks!

PR #13755, opened 2019-03-26
nocrypto: support cstruct 4.0.0

avsm:
add an explicit sexplib dependency rather than assuming it is transitive. This should work with all cstruct releases including 4.0.0

cc @hannesm
fixes build failure in #13748


avsm:
This is all green and a pretty simple patch, so I'll just merge it to unblock cstruct.4

mseri:
Thanks

PR #13757, opened 2019-03-26
stop old jupyter-kernel from being selected

avsm:
these all fail with some lwt ppx error, so stop them from trying
to build (especially in revdeps) until debugged

cc @c-cube


c-cube:
Seems good, i guess.

PR #13758, opened 2019-03-27
Add {conf-,}llvm.8.0.0

kit-ty-kate:
Will fail as usual.
Tested locally using the `ocaml/opam2:debian-unstable` docker image.
Might work on macos, I don't remember


mseri:
Indeed, green on mac

PR #13760, opened 2019-03-27
constrain qcow to cstruct<4

avsm:
due to making sexp optional in cstruct, see #13748

### DIFF
--- a/packages/qcow/qcow.0.10.0/opam
+++ b/packages/qcow/qcow.0.10.0/opam
@@ -12,7 +12,7 @@ tags: [
 depends: [
   "ocaml" {>= "4.03.0"}
   "astring"
-  "cstruct"
+  "cstruct" {< "4.0.0"}
   "result"
   "prometheus"
   "mirage-types-lwt" {>= "3.0.0"}
--- a/packages/qcow/qcow.0.10.2/opam
+++ b/packages/qcow/qcow.0.10.2/opam
@@ -18,7 +18,7 @@ build: [
 depends: [
   "ocaml" {>= "4.03.0"}
   "astring"
-  "cstruct"
+  "cstruct" {< "4.0.0"}
   "result"
   "prometheus"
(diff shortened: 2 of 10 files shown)


mseri:
Travis fails due to the sandbox preventing to write the test output in /tmp

mseri:
Solver failure for qcow 0.81, 0.9.4 and 0.9.5 on ocaml 4.04 reliably in the CI. Should I wait before merging?

kit-ty-kate:
No it's fine, the thing to look for, for this kind of PR, is at least that the linter/Camelus passes.

Thanks!

avsm:
@kit-ty-kate and @altgr are looking at the solver problem more generally via opam-check-all, so i'd say this is ok to merge too !  I've created https://github.com/mirage/ocaml-qcow/issues/110 to track the /tmp sandbox failure

AltGr:
Could reproduce the hard-to-solve issue (the problem took 39s on my laptop with `builtin-mccs+glpk`); indeed, the z3 backend will probably help for these kind of issues, but it's always good to have more problematic cases to test on. (the opam-check-all tool — which is the same as "marracheck" I take it? — is not directly related, though, it's just another user of the z3 backend)
The z3 backend could resolve the same problem in 7.7s (which is already a lot) on the same machine. This falls to 3.0s if we simplify the criteria a little (getting a solution with less changes, but some slightly older versions; that change has an opposite effect on mccs)


> Travis fails due to the sandbox preventing to write the test output in /tmp

I don't understand, the sandbox should always allow write access to /tmp ?

avsm:
unsure why its failing, but the /tmp log is here https://api.travis-ci.org/v3/job/511899585/log.txt

kit-ty-kate:
`the opam-check-all tool — which is the same as "marracheck" I take it?`

Not at all, `opam-health-check` has more or less the same goal as "marrakesh" but the two are two distinct software, use distinct techniques and the former is just over a year old

PR #13769, opened 2019-03-28
Z3 package: build and install the static library

AltGr:
by default, only the dynamic lib is installed (and that requires an
update to `LD_LIBRARY_PATH` for running linked binaries)


kit-ty-kate:
The error on 4.04 existed before. Thanks!

PR #13782, opened 2019-03-30
cstruct: constrain older version to ocaml<4.08.0

avsm:
due to Array.map_file change


kit-ty-kate:
MacOS build exceeded time limit and `conf-time` seems to be missing for OpenSUSE. Good enough. Thanks a lot!

PR #13784, opened 2019-03-30
mirage-unix: ensure older versions use older lwt

avsm:
to get camlp4


kit-ty-kate:
The failures on Travis are due to the builds exceed time limit. Thanks a lot!

PR #13796, opened 2019-04-01
Package syslog.2.0.0

sagotch:
### `syslog.2.0.0`
syslog(3) routines for ocaml

---
* Homepage: http://github.com/geneanet/ocaml-syslog
* Source repo: git://github.com/geneanet/ocaml-syslog
* Bug tracker: https://github.com/geneanet/ocaml-syslog/issues

---

### DIFF
--- a/packages/syslog/syslog.2.0.0/opam
+++ b/packages/syslog/syslog.2.0.0/opam
@@ -1,0 +1,28 @@
+opam-version: "2.0"
+
+synopsis: "syslog(3) routines for ocaml"
+
+authors: [
+  "Shawn Wagner"
+  "Eric Stokes"
+]
+
+maintainer: [ "Julien Sagot <julien.sagot@geneanet.org>" ]
+
+homepage: "http://github.com/geneanet/ocaml-syslog"
+dev-repo: "git://github.com/geneanet/ocaml-syslog"
+bug-reports: "https://github.com/geneanet/ocaml-syslog/issues"
+
+build: [ "dune" "build" ]
+
+depends: [
+  "dune" { build }
+  "ocaml" {>= "4.03.0"}
+]
+url {
+  src: "https://github.com/geneanet/ocaml-syslog/archive/v2.0.0.tar.gz"
+  checksum: [
+    "md5=5bf1565ee78aaf8cdc5505b2b386ab19"
+    "sha512=e69f34da1b873fc29db7e154e0ea799427aef38022f3e44246221245916f057cc68085f6e8eae0998c397af3578ecb8a824ac37e308a799fc521f3f170ae0d24"
+  ]
+}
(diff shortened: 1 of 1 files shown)


avsm:
Thanks for this! Just one thing I noticed is that it is licensed under the LGPL but without the linking exception (which is important for static linking that is default in OCaml). Is that intentional?

sagotch:
I could ask to original developper, but he did not answered my previous mail, so I have no idea. My contribution is mainly to update an original package so it can be used with recent OCaml, but most of the lines are from someone else.

packages/syslog/syslog.2.0.0/opam
```diff
@@ -0,0 +1,31 @@
...
+dev-repo: "git://github.com/geneanet/ocaml-syslog"
+bug-reports: "https://github.com/geneanet/ocaml-syslog/issues"
+
+build: [ make "build" ]
+install: [ make "install" ]
+remove: [ [ "ocamlfind" "remove" "syslog" ] ]
+flags: light-uninstall
+
+depends: [
+  "dune"
```
kit-ty-kate:
```suggestion
  "dune" {build}
```
This is most likely a build dependency

packages/syslog/syslog.2.0.0/opam
```diff
@@ -0,0 +1,31 @@
...
+  "Eric Stokes"
+]
+
+maintainer: [ "Julien Sagot <julien.sagot@geneanet.org>" ]
+
+homepage: "http://github.com/geneanet/ocaml-syslog"
+dev-repo: "git://github.com/geneanet/ocaml-syslog"
+bug-reports: "https://github.com/geneanet/ocaml-syslog/issues"
+
+build: [ make "build" ]
```
kit-ty-kate: Could you use the `dune` commands instead of the build/install/remove fields you have here? Look for other packages using dune to know which command to use (only the `build` field is necessary, the others shouldn't be there)

packages/syslog/syslog.2.0.0/opam
```diff
@@ -0,0 +1,31 @@
...
+install: [ make "install" ]
+remove: [ [ "ocamlfind" "remove" "syslog" ] ]
+flags: light-uninstall
```
kit-ty-kate: This flag shouldn't be there with dune, see previous comment.

packages/syslog/syslog.2.0.0/opam
sagotch: No `install` or `remove` field needed?

kit-ty-kate: no

kit-ty-kate: dune build already creates an `.install` file that opam uses to install and uninstall all the necessary files

sagotch: Should that `syslog.install` file be versioned in the repo so that opam do not need to build the library for uninstall? Or would it build it anyway?

mseri: You don't need to version it, it is kept by opam on install and used for the uninstall

sagotch:
I do not think that original authors will answer me about the license. If the rest is okay, you could merge this one and I'll make a new release if they agree to change the license.

mseri:
I made the license explicit, just to be on the safe side. @sagotch can you upstream (or try to) the fixes to the opam file? Please submit a new release if it is possible to change the license

mseri:
Thanks

PR #13797, opened 2019-04-01
 mlgmpidl.1.2.9

nberth:


packages/mlgmpidl/mlgmpidl.1.2.9/opam
```diff
@@ -0,0 +1,36 @@
...
+homepage: "https://www.inrialpes.fr/pop-art/people/bjeannet/mlxxxidl-forge/mlgmpidl/"
+license: "LGPL-2.1 with static linking exception"
+build: [
+  ["sh" "./configure" "--absolute-dylibs" {os = "macos"}]
+  [make]
+]
+install: [
+  [make "install"]
+]
+remove: [
```
mseri: The remove section is no longer needed in opam2 when you are just removing what you have installed. This does not prevent the merge though

nberth: Interesting… just out of curiosity, how is "what you have installed" recorded?

mseri: Opam2 tracks the files that are installed

PR #13799, opened 2019-04-01
Specify depopts for h2-lwt-unix

anmonteiro:
I forgot to include this in the original release and users were seeing build failures with e.g. esy which needs to track this information.


hcarty:
The macos build failure on travis looks unrelated.  Could someone with access restart that build?

PR #13803, opened 2019-04-01
graphql_parser.0.11.0: fix missing `result` dependency.

dbuenzli:
It seems graphql_parser.0.11.0 is getting its `result` dependency indirectly through `fmt`. But the latest release of `fmt` drops the `result` dependency which results in a [build failure](https://ci.ocamllabs.io/log/saved/docker-run-f5501ba2891862ebca2f3088e59459a6/23bf11a4c8a0535463c58a64ab76cfa9e404f8b7).

This PR adds a dependency on `result` (note that e.g. `0.9.0` has the dependency on `result).

/cc @andreas

### DIFF
--- a/packages/graphql_parser/graphql_parser.0.11.0/opam
+++ b/packages/graphql_parser/graphql_parser.0.11.0/opam
@@ -16,6 +16,7 @@ depends: [
   "dune" {build}
   "menhir" {build}
   "alcotest" {with-test & >= "0.8.1"}
+  "result"
   "fmt"
   "re" {>= "1.5.0"}
 ]


andreas:
Sorry about that, and thanks for the fix. I'll make sure it's fixed with next release.

phated:
This is now causing my build to fail. I'm not sure why.  I'm using ocaml 4.06 so it shouldn't be an issue, right?

dbuenzli:
@phated without failure details it's going to be hard to say anything.

dbuenzli:
Also it seems this was merged after the `fmt` PR. I should have mentioned it should have been merged before. Are you sure you were not witnessing the state with `fmt` in and that one not in ?

phated:
It was using `fmt` at 0.8.6 which I believe is the latest.  The error is just "Error: Unbound module Result" but I'm not sure if the error log is helpful since it's from esy.

dbuenzli:
> It was using `fmt` at 0.8.6 which I believe is the latest. The error is just "Error: Unbound module Result" but I'm not sure if the error log is helpful since it's from esy.

Precisely this PR fixes this error.

phated:
But I was on graphql_parser 0.11.0 unless this overwrote an 0.11.0 version instead of incrementing the version?

dbuenzli:
> But I was on graphql_parser 0.11.0 unless this overwrote an 0.11.0 version instead of incrementing the version?

Yes it id. It seems we still don't have a good answer to this see https://github.com/ocaml/opam-repository/issues/10531

samoht:
I don't understand how this PR is fixing the issue ; `result` should be added to the build files, not opam files, right?

dbuenzli:
Humpf. Right, to both. I'll PR a constraint.

dbuenzli:
See See https://github.com/ocaml/opam-repository/pull/13808

PR #13804, opened 2019-04-02
Lwt 4.2.1

aantron:
Just a small bugfix release.

> Bug fixed
>
> - Detect libev correctly when building under esy (ocsigen/lwt#679, Antonio Nuno Monteiro).

### DIFF
--- a/packages/lwt/lwt.4.2.1/opam
+++ b/packages/lwt/lwt.4.2.1/opam
@@ -1,0 +1,58 @@
+opam-version: "2.0"
+
+synopsis: "Promises and event-driven I/O"
+
+license: "MIT"
+homepage: "https://github.com/ocsigen/lwt"
+doc: "https://ocsigen.org/lwt/manual/"
+bug-reports: "https://github.com/ocsigen/lwt/issues"
+
+authors: [
+  "Jérôme Vouillon"
+  "Jérémie Dimino"
+]
+maintainer: [
+  "Anton Bachin <antonbachin@yahoo.com>"
+  "Mauricio Fernandez <mfp@acm.org>"
+  "Simon Cruanes <simon.cruanes.2007@m4x.org>"
+]
+dev-repo: "git+https://github.com/ocsigen/lwt.git"
+
+depends: [
+  "cppo" {build & >= "1.1.0"}
+  "dune" {build}
+  "mmap"  # mmap is needed as long as Lwt supports OCaml < 4.06.0.
+  "ocaml" {>= "4.02.0"}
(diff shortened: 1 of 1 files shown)


packages/lwt/lwt.4.2.1/opam
```diff
@@ -0,0 +1,58 @@
...
+depends: [
+  "cppo" {build & >= "1.1.0"}
+  "dune" {build}
+  "mmap"  # mmap is needed as long as Lwt supports OCaml < 4.06.0.
+  "ocaml" {>= "4.02.0"}
+  "result" # result is needed as long as Lwt supports OCaml 4.02.
+  "seq" # seq is needed as long as Lwt supports OCaml < 4.07.0.
+
+  "bisect_ppx" {dev & >= "1.3.0"}
+  "ocamlfind" {dev & >= "1.7.3-1"}
```
kit-ty-kate: This is not blocker at all but I'm curious why there is still a dependency on ocamlfind

packages/lwt/lwt.4.2.1/opam
```diff
@@ -0,0 +1,58 @@
...
+  "Mauricio Fernandez <mfp@acm.org>"
+  "Simon Cruanes <simon.cruanes.2007@m4x.org>"
+]
+dev-repo: "git+https://github.com/ocsigen/lwt.git"
+
+depends: [
+  "cppo" {build & >= "1.1.0"}
+  "dune" {build}
+  "mmap"  # mmap is needed as long as Lwt supports OCaml < 4.06.0.
+  "ocaml" {>= "4.02.0"}
```
kit-ty-kate: Is this release also compatible with OCaml 4.08?

anmonteiro: It is since 4.2.0 (https://github.com/ocsigen/lwt/releases/tag/4.2.0)

kit-ty-kate: nevermind I didn't see the 4.2 release

kit-ty-kate:
Revdeps failures don't look too bad, some might be related but as it's only a small release I'm merging this now. Thanks!

packages/lwt/lwt.4.2.1/opam
aantron: The PPX expect tests [use](https://github.com/ocsigen/lwt/blob/6213e180686f7776781e9a50305792710df8d915/test/ppx_expect/main.ml#L54) it as an easy way of invoking the compiler.

aantron:
Thank you!

PR #13806, opened 2019-04-02
[new releases] x509 (0.6.3) tls (0.10.2)

hannesm:
X509 CHANGES:

* provide X509.Encoding.distinguished_name_of_cs -- similar to mirleft/ocaml-x509#87 which provided distinguished_name_to_cs
* provide X509.Encoding.{public_key_of_cstruct,public_key_to_cstruct}, as requested by @dinosaure
* support of cstruct 4.0.0, which split up the sexp de&encoders
* removes result dependency (now requires >= 4.04.2)
* upgrades opam file to version 2.0
* build system is now dune

TLS CHANGES:

* support for cstruct 4.0.0+
* remove support for < 4.04.2 (same as x509 in master)
* remove result (part of 4.03.0)
* enhance mirage/example2 to work on more platforms than unix

### DIFF
--- a/packages/x509/x509.0.6.3/opam
+++ b/packages/x509/x509.0.6.3/opam
@@ -1,0 +1,51 @@
+opam-version: "2.0"
+name: "x509"
+maintainer: [
+  "Hannes Mehnert <hannes@mehnert.org>" "David Kaloper <david@numm.org>"
+]
+authors: [
+  "David Kaloper <david@numm.org>" "Hannes Mehnert <hannes@mehnert.org>"
+]
+license: "BSD2"
+tags: "org:mirage"
+homepage: "https://github.com/mirleft/ocaml-x509"
+doc: "https://mirleft.github.io/ocaml-x509/doc"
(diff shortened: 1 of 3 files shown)


packages/x509/x509.0.6.3/opam
```diff
@@ -0,0 +1,51 @@
...
+]
+license: "BSD2"
+tags: "org:mirage"
+homepage: "https://github.com/mirleft/ocaml-x509"
+doc: "https://mirleft.github.io/ocaml-x509/doc"
+bug-reports: "https://github.com/mirleft/ocaml-x509/issues"
+depends: [
+  "ocaml" {>= "4.04.2"}
+  "dune" {build & >= "1.2"}
+  "ppx_sexp_conv"
```
avsm:
```suggestion
  "ppx_sexp_conv" {>="v0.11.0"}
```

This constraint ensures that older non-dune ppx_sexp_convs wont be picked up (which can happen on older compilers)

kit-ty-kate:
Thank you!

PR #13808, opened 2019-04-02
graphql_parser: really fix missing `result` dependency by constrainting fmt

dbuenzli:
The fix in a4d2989b5c7c41c3f is wrong.

See https://github.com/ocaml/opam-repository/pull/13803


samoht:
Sounds like the right fix, merging.

PR #13814, opened 2019-04-03
Update the pprint package description so as to use Makefile.

fpottier:
Update the pprint package description so as to use Makefile, not GNUmakefile. This should fix a build failure on systems that do not have bash.


mseri:
Thanks

PR #13815, opened 2019-04-03
ppx_tools_versioned not compatible with OMP 1.3.0

aantron:
See ~~https://github.com/ocaml-ppx/ppx_tools_versioned/pull/15~~ https://github.com/ocaml-ppx/ppx_tools_versioned/pull/12.


mseri:
I was scared by some CI failures, but looking at the CI it comes out it did use the old ppx_tools_versioned in all tests. So all should be fine

PR #13816, opened 2019-04-03
conf-openssl: (macos) try `pkg-config openssl` first

mattjbray:
Try just executing `pkg-config openssl` before looking explicitly for Nix -> Homebrew -> MacPorts.

This handles the case where the user has set `PKG_CONFIG_PATH` themselves.

In my case I'm actually using Nix, but rather than having `openssl` in my nix user profile I want to use `nix-shell` with the following `shell.nix` in my project directory:

```nix
{ pkgs ? import <nixpkgs> {} }:
  pkgs.mkShell {
    buildInputs = [
        pkgs.pkg-config
        pkgs.openssl
      ];
}
```

Then Nix sets `PKG_CONFIG_PATH` for me:

```shell
$ nix-shell

[nix-shell:~/code/project]$ echo $PKG_CONFIG_PATH
/nix/store/1zp3il13n33albk772nr3jrv8achkx09-openssl-1.0.2r-dev/lib/pkgconfig

```

### DIFF
--- a/packages/conf-openssl/conf-openssl.1/files/osx-build.sh
+++ b/packages/conf-openssl/conf-openssl.1/files/osx-build.sh
@@ -1,5 +1,15 @@
 #!/bin/sh
 
+# Try just executing pkg-config openssl before looking
+# explicitly for Nix -> Homebrew -> MacPorts.
+# This handles the case where the user has set
+# PKG_CONFIG_PATH themselves.
+res=$(pkg-config openssl)
+if [ $? -eq 0 ]; then
+    echo $res
+    exit 0
+fi
+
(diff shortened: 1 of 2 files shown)


mseri:
Please update the comment as suggested by @anmonteiro. Probably using the first two line of the PR description would be enough

```
# Try just executing pkg-config openssl before looking 
# explicitly for Nix -> Homebrew -> MacPorts.
# This handles the case where the user has set
# PKG_CONFIG_PATH themselves.
```

After that, I am happy to merge.

mseri:
Thanks!

anmonteiro:
@mattjbray just a quick note to tell you this works fine locally after `opam update`. Thanks again!

mattjbray:
Great, thank you!

PR #13821, opened 2019-04-04
Package crontab.0.1

yurug:
### `crontab.0.1`
Interacting with cron from OCaml
With this library, an ocaml program can interact with cron:
parse/print crontab and update installed crontabs.

---
* Homepage: https://github.com/yurug/ocaml-crontab
* Source repo: git://github.com/yurug/ocaml-crontab.git
* Bug tracker: https://github.com/yurug/ocaml-crontab/issues

---

### DIFF
--- a/packages/crontab/crontab.0.1/opam
+++ b/packages/crontab/crontab.0.1/opam
@@ -1,0 +1,32 @@
+opam-version: "2.0"
+
+synopsis: "Interacting with cron from OCaml"
+description: """
+With this library, an ocaml program can interact with cron:
+parse/print crontab and update installed crontabs.
+"""
+
+maintainer: "Yann Régis-Gianas <yrg@irif.fr>"
+authors: [
+  "Yann Régis-Gianas <yann.regis-gianas@irif.fr>"
+]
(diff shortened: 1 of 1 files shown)


packages/crontab/crontab.0.1/opam
```diff
@@ -0,0 +1,33 @@
...
+
+homepage: "https://github.com/yurug/ocaml-crontab"
+bug-reports: "https://github.com/yurug/ocaml-crontab/issues"
+dev-repo: "git://github.com/yurug/ocaml-crontab.git"
+
+depends: [
+  "dune"                 {build & >= "1.4.0"}
+  "ocaml"
+  "odoc"                 {with-doc}
+  "ocamlfind"
```
mseri: You don't seem to be using ocamlfind in the library and tests code, only in the example when running as a script.

yurug: You are right. Let's minimize the dependencies.

yurug: Thanks

mseri:
Thanks

PR #13827, opened 2019-04-05
[new release] codept 0.10.3

Octachron:
Codept 0.10.3 - Alternative ocaml dependency analyzer
-------
This is essentially a 4.08 compatibility release with a compiler upper bound on 4.09 since available data points show that codept generally breaks on new compiler releases.  

### Support

  * OCaml 4.08

### Features

* Flatter json and sexp output
* Less noisy ambiguous module warning

### Bug fixes

  * Missing dependencies while matching first-class modules

### DIFF
--- a/packages/codept/codept.0.10.3/opam
+++ b/packages/codept/codept.0.10.3/opam
@@ -1,0 +1,33 @@
+opam-version: "2.0"
+name: "codept"
+version: "0.10.3"
+author: "octachron <octa@polychoron.fr>"
+maintainer: "octachron <octa@polychoron.fr>"
+homepage: "https://github.com/Octachron/codept"
+bug-reports: "https://github.com/Octachron/codept/issues"
+license: "gpl-3"
+dev-repo: "git+https://github.com/Octachron/codept.git"
+build: [
+  ["dune" "build" "-p" name]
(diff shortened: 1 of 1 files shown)


packages/codept/codept.0.10.3/opam
```diff
@@ -0,0 +1,34 @@
...
+author: "octachron <octa@polychoron.fr>"
+maintainer: "octachron <octa@polychoron.fr>"
+homepage: "https://github.com/Octachron/codept"
+bug-reports: "https://github.com/Octachron/codept/issues"
+license: "gpl-3"
+dev-repo: "git+https://github.com/Octachron/codept.git"
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+  ["dune" "build" "@private-docs"] {with-doc}
```
mseri: Are you sure you want to bundle the private docs to the documentation?

Octachron: Right, it is probably better to remove it for now.

Octachron: I have removed this line (and the odoc dependency).

mseri:
Thanks

PR #13829, opened 2019-04-05
[new release] re (1.9.0)

rgrinberg:
CHANGES:

* Fix regression in `Re.exec_partial` (ocaml/ocaml-re#164)
* Mov gen related functions to `Re.Gen` and deprecate the old names (ocaml/ocaml-re#167)
* Introduce `Re.View` that exposes the internal representation (ocaml/ocaml-re#163)

### DIFF
--- a/packages/re/re.1.9.0/opam
+++ b/packages/re/re.1.9.0/opam
@@ -1,0 +1,42 @@
+opam-version: "2.0"
+
+maintainer: "rudi.grinberg@gmail.com"
+authors: [
+  "Jerome Vouillon"
+  "Thomas Gazagnaire"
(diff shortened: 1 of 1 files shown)


mseri:
Is the tyre failure a new one, or some recent version introduced breaking changes and we need to add an upper bound?
```
# File "src/tyre.ml", line 333, characters 19-29:
# Error: Unbound value Re.all_seq
# Hint: Did you mean all_gen?
```

rgrinberg:
that was not intentional. i pushed a fix.

mseri:
Thanks. I did not find suspicious errors in this CI run, merging

PR #13847, opened 2019-04-08
Patches to opam-publish released versions

dra27:
While checking something else, I found that the latest versions of `opam-publish` were missing dependencies (not relevantly, obviously).

cc @rjbou, @AltGr - did the fields get lost because of the opam file in the main repo? This file used to have `flags: plugin` as well?


rjbou:
Updated pr ocaml/opam-publish/pull/79 accordingly.

kit-ty-kate:
Thanks!

PR #13850, opened 2019-04-08
tyxml: use either a recent version of OCaml or Re

samoht:
/cc @Drup


Drup:
Thanks! I confirm that this is the right constraint.

kit-ty-kate:
Thanks

PR #13863, opened 2019-04-09
[new release] psq (0.2.0)

pqwy:
Functional Priority Search Queues

- Project page: <a href="https://github.com/pqwy/psq">https://github.com/pqwy/psq</a>
- Documentation: <a href="https://pqwy.github.io/psq/doc">https://pqwy.github.io/psq/doc</a>

##### CHANGES:

Semantics cleanup.

- flipped args to `adjust` **breaking**
- `of_list` now always chooses the rightmost binding
- `update`, `(++)`, `add_seq`, `to_priority_list`
- somewhat faster

### DIFF
--- a/packages/psq/psq.0.2.0/opam
+++ b/packages/psq/psq.0.2.0/opam
@@ -9,7 +9,7 @@ bug-reports: "https://github.com/pqwy/psq/issues"
 synopsis: "Functional Priority Search Queues"
 build: [ [ "dune" "subst" ] {pinned}
          [ "dune" "build" "-p" name "-j" jobs ]
-         [ "dune" "runtest" "-p" name ] {with-test} ]
+         [ "dune" "runtest" "-p" name ] {with-test & ocaml:version >= "4.07.0"} ]
 depends: [
   "ocaml" {>="4.03.0"}
   "dune" {build & >= "1.7"}
--- a/packages/psq/psq.0.2.0/opam
+++ b/packages/psq/psq.0.2.0/opam
@@ -1,0 +1,28 @@
+opam-version: "2.0"
+maintainer: "David Kaloper Meršinjak <dk505@cam.ac.uk>"
+authors: ["David Kaloper Meršinjak <dk505@cam.ac.uk>"]
+homepage: "https://github.com/pqwy/psq"
+doc: "https://pqwy.github.io/psq/doc"
+license: "ISC"
+dev-repo: "git+https://github.com/pqwy/psq.git"
+bug-reports: "https://github.com/pqwy/psq/issues"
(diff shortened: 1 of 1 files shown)


avsm:
Thanks! @anmonteiro is h2 affected by the breaking change to `adjust` above? Just checking it doesnt need an upper bound.

pqwy:
It's not. It vendors psq and never uses the one incompatible function.

anmonteiro:
H2 never uses `adjust`. I'm not sure what David means by "it vendors psq". H2 uses Psq from opam

avsm:
Thanks for checking, both. The older `lru` releases need a `psq {<"0.2.0"}` constraint as well.

pqwy:
Sorry, my bad -- noticed this [`esy.lock`](https://github.com/anmonteiro/ocaml-h2/blob/3f25694b0c6b5436b78f6a7ff3081347ba8e5a7b/esy.lock/index.json) while auditing its API usage of psq.

anmonteiro:
Ah! that's just a lockfile that esy uses for deterministic builds. I can see how it might be confusing at first.

avsm:
The tests fail in <4.07 due to:
```
# File "test/test.ml", line 121, characters 8-19:
# Error: Unbound value List.of_seq`
```

Not sure exactly what the constraint runes are for an ocaml constraint for just with-test... @altgr?

pqwy:
Tests are >= 4.07 only. Psq's [own .travis file](https://github.com/pqwy/psq/blob/master/.travis.yml) only enables them at that version, but this CI does not that that into account.

IOW spurious failure. See [here](https://travis-ci.org/pqwy/psq/builds/517949074?utm_source=github_status&utm_medium=notification).

avsm:
The problem is, if you do an `opam install -t psq` on a version earlier than 4.07 (or indeed, `opam install -t h2`), it'll fail due to being unable to compile the tests.

pqwy:
> or indeed, opam `install -t h2`

`opam install --help`:

       -t, --with-test, --build-test
           Build and run the package unit-tests. This only affects packages
           listed on the command-line. [...]
 
Perhaps not supporting `opam install -t psq` with older compilers is not that bad.

avsm:
aha! That's an opam2 improvement indeed -- opam1 used to apply it to the transitive dependencies.  Let's wait until tomorrow for the other maintainers to chirp in -- if there's no easy constraint formula then we can merge this.

pqwy:
I just noticed that @mseri added the correct incantation to opam:

```
[ "dune" "runtest" "-p" name ] {with-test & ocaml:version >= "4.07.0"}
```

Thank you!

PR #13871, opened 2019-04-10
biocaml.0.{5,6,7}.0 is incompatible with re >= 1.8.0 which uses seq (incompatible interface)

kit-ty-kate:

cc @pveber I'm not sure why recent versions are not affected but those 3 at least are (maybe more ancient ones too but I don't know)


pveber:
I'm late to the party, thanks a lot @kit-ty-kate!

PR #13876, opened 2019-04-10
Fix liquidsoap.1.3.7 md5

toots:
Hi,

Thanks for the quick merge earlier. Unfortunately, there was a small issue with the build tools in the released tarball. I have generated a new one and updated the `md5` signature here.


kit-ty-kate:
Thanks

PR #13872, opened 2019-04-10
z3.4.8.1 doesn't support parallel compilation

kit-ty-kate:
Detected by [check.ocamllabs.io](http://check.ocamllabs.io)

@c-cube are the other versions also affected?

PR #13879, opened 2019-04-10
z3: Fix parallel compilation

kit-ty-kate:
Backport of https://github.com/Z3Prover/z3/pull/2227

Better alternative to #13872


kit-ty-kate:
cc @c-cube

PR #13885, opened 2019-04-10
sequence.1.1 is not compatible with OCaml >= 4.08

kit-ty-kate:

cc @c-cube only very old versions of sequence are compatible with OCaml >= 4.08


c-cube:
there's a version of Iter that should support this, once 4.08 is released.

PR #13911, opened 2019-04-11
mdx is not compatible with OCaml >= 4.08

kit-ty-kate:

cc @samoht


samoht:
Thanks!

PR #13922, opened 2019-04-11
Fix otetris dependencies and remove its install rule

kit-ty-kate:


kit-ty-kate:
cc @vzaliva

PR #13926, opened 2019-04-11
protocol-9p: Add missing dependency and constraints

kit-ty-kate:

cc @djs55 (maybe the fix in `2.0.0` can be returned upstream)


djs55:
Thanks!

PR #13928, opened 2019-04-11
Fix various packages relying on Lwt >= 3.1

kit-ty-kate:


kit-ty-kate:
cc @pqwy for the change in `cpuid`

PR #13938, opened 2019-04-11
Add gapi-ocaml 0.3.10

astrada:


packages/gapi-ocaml/gapi-ocaml.0.3.10/opam
```diff
@@ -0,0 +1,38 @@
...
+  "ocaml" {>= "4.02.3"}
+  "cryptokit"
+  ("extlib" | "extlib-compat")
+  "dune" {build}
+  "ocamlnet" {>= "4.1.4"}
+  "ocurl"
+  "ounit" {with-test}
+  "ppx_monadic" {with-test}
+  "xmlm"
+  "yojson"
```
mseri: Are you sure that this does not need a lower bound?

mseri:
Looks good, just a question concerning yojson, as I remember that there were breaking changes lately

astrada:
> Looks good, just a question concerning yojson, as I remember that there were breaking changes lately

AFAIK, version 2.0.0 which will break compatibility is not yet out. I will add a constraint when it gets released (if it breaks my package).

mseri:
I was thinking more about the `Yojson.json` -> `Yojson.t` in one of the last versions. The package definitely works with the latest version, that's why I thought it may have needed a lower bound

astrada:
I'm using `Yojson.Safe.json` and I think that they added an alias to keep compatibility.

mseri:
Thanks

PR #13942, opened 2019-04-12
 H2-lwt-unix depends on ocamlfind for the build

anmonteiro:


kit-ty-kate:
fair enough. Reading the source code, it seems the reason you use ocamlfind can be replaced with dune's `(optional)` alone. At least I'm guessing

kit-ty-kate:
Thanks

anmonteiro:
I'm interested in knowing how to make it work with `optional`, actually. If I well understand, it would have to be in a sublibrary which wouldn't share the same toplevel module anymore.

kit-ty-kate:
mmh, I looked a bit more and yes it seems that they would need to be in separate sublibs. However that would be a better design anyway I think.

PR #13949, opened 2019-04-13
datakit-ci: depend on cstruct<4 due to sexp fix

avsm:
found in revdeps for #13947


avsm:
Solver angst ... merging

PR #13953, opened 2019-04-14
Fix bap packages: fix their dependencies to use the same version as the package version

kit-ty-kate:

cc @gitoleg could you check if there is no mistake from me here?
cc @ivg is there a reason why `bap-saluki` has a rather weird version number? (`bap-1.5` for instance)

### DIFF
--- a/packages/bap-abi/bap-abi.1.0.0/opam
+++ b/packages/bap-abi/bap-abi.1.0.0/opam
@@ -16,7 +16,11 @@ remove: [["ocamlfind" "remove" "bap-plugin-abi"]
          ["ocamlfind" "remove" "bap-abi"]
          ["bapbundle" "remove" "abi.plugin"]
          ]
-depends: ["ocaml" "bap-std" "cmdliner"]
+depends: [
+  "ocaml"
+  "bap-std" {= "1.0.0"}
+  "cmdliner"
+]
 synopsis: "A pass that applies ABI specific information"
 url {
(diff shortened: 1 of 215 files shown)


ivg:
I think it's accidental, it coincides with the tag version. It should be 1.5.0 of course. Thanks for the PR, it looks reasonable, but let us run it on our CI, just to double check.

kit-ty-kate:
i think that should fix most of the breakage, though there is still the missing `bap-ida-python.1.4.0` package that messes things a bit. Does anyone know how to fix it properly?

gitoleg:
@kit-ty-kate 
I'm currently working on it, so I'll let you know when everything will be ready

ivg:
Yep, it looks like that `bap-ida-python.1.4.0`  wasn't ever released. Probably, it would be easier to release it retrospectively (basically to rerelease bap-ida-python.1.3.0 as bap-ida-python.1.4.o), just to make automated versioning easier and to remove this nasty whole in versioning.

kit-ty-kate:
I suddenly remembered I could have used `{= version}` instead of `{= "<the same version as the package I'm modifying>"}`. It would have avoided some problems and be less of a hassle. Anyway, I'll remember for next time..

gitoleg:
@kit-ty-kate 
hey, there is a PR in your `opam-repository` that is waiting for you)

kit-ty-kate:
Ah sorry, apparently when you're forking a repository, by default github doesn't make you "watch" it so I didn't get notified

kit-ty-kate:
Thanks a lot for your help

PR #13955, opened 2019-04-14
doc-ock is not compatible with OCaml >= 4.08

kit-ty-kate:

cc @lpw25 there is no version of `doc-ock` compatible with OCaml 4.08


avsm:
The new odoc doesn't depend on doc-ock so this is safe to merge. When odoc 1.5.0 when released it will have OCaml 4.08.0 support (see https://github.com/ocaml/odoc/issues/350 for the tracking issue and related links)

PR #13956, opened 2019-04-14
fkie-cad-cwe-checker.0.1: Add missing dependencies and constraints

kit-ty-kate:

cc @tbarabosch you're using at least those packages explicitly in your code but they weren't explicitly listed in the dependencies. It worked only because they were dependencies of `bap-std`. Your package wasn't compatible with `core_kernel >= v0.11.0` so I added the required constraint.

### DIFF
--- a/packages/fkie-cad-cwe-checker/fkie-cad-cwe-checker.0.1/opam
+++ b/packages/fkie-cad-cwe-checker/fkie-cad-cwe-checker.0.1/opam
@@ -21,6 +21,9 @@ depends: [
   "ocaml"
   "bap-std" {>= "1.5.0"}
   "yojson" {= "1.4.1"}
+  "base-unix"
+  "ppx_jane" {< "v0.11"}
+  "core_kernel" {< "v0.11"}
 ]
 synopsis:
   "A plugin for bap that detects common bug classes, e.g. use of dangerous functions and simple integer overflows"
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Mmh there is still a failure:
```
#=== ERROR while installing fkie-cad-cwe-checker.0.1 ==========================#
# context              2.0.4 | linux/x86_64 | ocaml-base-compiler.4.03.0 | pinned(https://github.com/fkie-cad/cwe_checker/archive/v0.1.tar.gz)
# path                 ~/.opam/4.03/.opam-switch/build/fkie-cad-cwe-checker.0.1
# command              /bin/cp src/config.json /home/opam/.opam/4.03/share/bap/
# exit-code            1
# env-file             ~/.opam/log/fkie-cad-cwe-checker-20-7e54d9.env
# output-file          ~/.opam/log/fkie-cad-cwe-checker-20-7e54d9.out
### output ###
# /bin/cp: cannot create regular file '/home/opam/.opam/4.03/share/bap/': Not a directory
```
cc @gitoleg @ivg

tbarabosch:
Hey, 
we are transitioning to BAP 1.6 and by that we will use Core_kernel >= 0.11.0. The code in our branch should work already with newer versions of Core_kernel. We are going to tag it as cwe_checker 0.2 in the next week or two, depending on some open PRs.

kit-ty-kate:
@tbarabosch thanks! Do you have any idea about the other failure above though?

kit-ty-kate:
Let's see if that works

ivg:
The `$prefix/share/bap` is not created by bap-std, and is not actually used by bap. It is probably a bad idea to install the configuration file into the shared folder of the other package, even if you are depending on it. First of all, it may conflict with the other package own files (it is quite possible that we will decide to put our own config.json into our folder).

Also, probably share is not the most conventional place for configuration files. There is  a `$prefix/etc/$package`  for opam-wide configuration, aka site configuration.

tbarabosch:
Hey there,
for now this solution is not the best but it works in the current constellation. So I think this fixes the issue at hand. However, as we are going to tag cwe_checker 0.2 soon, we are going to adjust the place of the default configuration.

kit-ty-kate:
ok let's merge this then. Thanks!

PR #13958, opened 2019-04-14
liquidsoap is not compatible with OCaml >= 4.08

kit-ty-kate:

@toots there is no version of `liquidsoap` compatible with OCaml >= 4.08. Also it would be sensible to avoid the use of `warning-as-errors` for future releases.


toots:
Gotcha, will look a it, thanks!

PR #13964, opened 2019-04-15
Add ocaml.4.08.0 and ocaml.4.09.0 virtual packages

gretay-js:
Add missing ocaml packages for ocaml 4.08.0 and 4.09.0 to allow pinning a local version of ocaml compiler using ocaml-variants.opam, as discussed in https://github.com/ocaml/ocaml/pull/8604#issuecomment-482799149 and https://github.com/ocaml/ocaml/pull/2207#issuecomment-449417070

### DIFF
--- a/packages/ocaml-system/ocaml-system.4.08.0/files/gen_ocaml_config.ml.in
+++ b/packages/ocaml-system/ocaml-system.4.08.0/files/gen_ocaml_config.ml.in
@@ -1,0 +1,43 @@
+let () =
+  let exe = ".exe" in
+  let ocamlc =
+    let (base, suffix) =
+      let s = Sys.executable_name in
+      if Filename.check_suffix s exe then
+        (Filename.chop_suffix s exe, exe)
+      else
+        (s, "") in
+    base ^ "c" ^ suffix in
+  if Sys.ocaml_version <> "%{_:version}%" then
+    (Printf.eprintf
+       "ERROR: The compiler found at %%s has version %%s,\n\
+        and this package requires %{_:version}%.\n\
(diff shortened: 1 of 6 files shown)


avsm:
This does frustratingly lead to uninstallable packages in opam-repository, which will need to be special cased in `opam admin lint`.  With apologies if this has been covered in the ocaml/ocaml PR, but is there any reason why we can't just put these `ocaml.opam` files upstream into the relevant branches so they are picked up alongside the variants package?

gretay-js:
We get this error because we do not have the corresponding version of ocaml compiler package, right? The same problem as with ocaml-variants.opam. If so, I do not think there is a way around keeping ocaml.opam in ocaml/ocaml repository alongside ocaml-variants.opam until the official release and update of ocaml-base-compiler package. @dra27

dra27:
It seems better to have the packages in one place rather than in several (even taking `pin-depends` into account)

I had forgotten about the installability check, but it occurs to me that there is a simple solution - if `ocaml-system.4.08.0` and `ocaml-system.4.09.0` are added, then the packages are all “installable”.

That reduces the problem to being happy that ocaml-system packages exist prior to release, but I’m not sure I see an issue with this, beyond the slightly strange display of “latest version”?

avsm:
@dra27:
> I had forgotten about the installability check, but it occurs to me that there is a simple solution - if ocaml-system.4.08.0 and ocaml-system.4.09.0 are added, then the packages are all “installable”.

Yes, that seems fine -- if a system version of the development compiler _is_ installed, then that would work as intended (e.g. through a PPA or other development build).

@gretay-js thanks again for the PR.  I've pushed the ocaml-system packages to your branch, so can merge this as soon as that hopefully passes.

PR #13966, opened 2019-04-15
Jsoo 3.4.0

hhugo:
- adds support for OCaml 4.08


kit-ty-kate:
```
[ERROR] undefined filter variable in dependencies of js_of_ocaml-compiler.3.4.0: test
```

kit-ty-kate:
Fixed. Could you return the fix upstream?

hhugo:
@kit-ty-kate, thanks, the opam file is already fixed upstream.

hhugo:
The remaining error seems to be unrelated.

kit-ty-kate:
thanks!

PR #13974, opened 2019-04-16
Add constraints before jsoo 3.4.0

hhugo:


hhugo:
This is just adding some constraints. Errors seem to be unrelated.

avsm:
lgtm - just be careful to call the new jsoo `3.4.0` and not `3.4` as the versions are used interchangeably in the constraints above.

kit-ty-kate:
Thanks!

PR #13977, opened 2019-04-16
ppx_cstruct: remove unused test-dep to ppx_driver

emillon:
See https://github.com/mirage/ocaml-cstruct/pull/256

Thanks!


kit-ty-kate:
Thanks!

PR #13978, opened 2019-04-16
Patch nocrypto.0.5.{0,1,2,3} to fix build failures with recent versions of GCC

kit-ty-kate:

cc @hannesm


hannesm:
you should //cc @pqwy here, who is the author and maintainer of nocrypto.

PR #13984, opened 2019-04-17
added packages/conf-python-3-6

UnixJunkie:


kit-ty-kate:
there is no `python3.6` package in debian. Do you specifically need 3.6 or just 3?

UnixJunkie:
Ok, I pushed some more changes.
However, note that some operating systems/packagers don't have a python3 package.
In which case I ask for python3.6.
Maybe this conf package could be renamed to conf-python-3, since it does not enforce python3.6 in all environments.

UnixJunkie:
I just fixed the python script also.

kit-ty-kate:
> Maybe this conf package could be renamed to conf-python-3, since it does not enforce python3.6 in all environments.

I think it would be best

UnixJunkie:
I renamed it.

UnixJunkie:
Still not OK?

kit-ty-kate:
The name of the package doesn't match the directory it is in:
`conf-python-3-1.0.0` vs. `conf-python-3.1.0.0`. The minor and micro versions might be a bit overkill for conf packages too but it's not big deal.

UnixJunkie:
I renamed the directory.

kit-ty-kate:
Thanks!

UnixJunkie:
Thanks for the fixes!

PR #13985, opened 2019-04-17
Add octavius.1.2.1

lpw25:
Bugfix release so that it works with 4.08

### DIFF
--- a/packages/octavius/octavius.1.2.1/opam
+++ b/packages/octavius/octavius.1.2.1/opam
@@ -12,7 +12,6 @@ tags: ["doc" "ocamldoc" "org:ocaml-doc"]
 
 depends: [
   "ocaml" {>= "4.03.0"}
-  "ocamlfind" {build}
   "jbuilder" {build & >= "1.0+beta7"}
 ]
 build: [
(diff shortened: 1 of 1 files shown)


packages/octavius/octavius.1.2.1/opam
```diff
@@ -0,0 +1,27 @@
...
+homepage: "https://github.com/ocaml-doc/octavius"
+doc: "http://ocaml-doc.github.io/octavius/"
+license: "ISC"
+dev-repo: "git+http://github.com/ocaml-doc/octavius.git"
+bug-reports: "https://github.com/ocaml-doc/octavius/issues"
+tags: ["doc" "ocamldoc" "org:ocaml-doc"]
+
+depends: [
+  "ocaml" {>= "4.03.0"}
+  "ocamlfind" {build}
```
kit-ty-kate:
```suggestion
```
`octavius` doesn't seem to use ocamlfind at all in its source code

lpw25: It uses ocamlbuild with `-use-ocamlfind`

lpw25: Wait, no it doesn't do that anymore, I forgot that it got changed to jbuilder

mseri:
Thanks

PR #13989, opened 2019-04-18
ezsqlite.0.4: remove dependency on nonexistent `str` package

damiendoligez:
`str` is still built into the OCaml distribution.


mseri:
Thanks

PR #13990, opened 2019-04-18
gpr.1.3.0: fix dependency on `threads`, which does not exist

damiendoligez:
All other versions of `gpr` depend on `base-threads` instead.


mseri:
Thanks

PR #13992, opened 2019-04-18
Fix omd-2.0.0~alpha not being a directory

anmonteiro:
cc @rjbou @andreypopp


avsm:
Thanks!

avsm:
this triggered a latent bug in opam and had to be reverted; filed as https://github.com/ocaml/opam/issues/3830

sorry about that -- not the fault of the PR submitter in any way!

rjbou:
This kind of fix will be able to merge once we switch to another patch implementation.

PR #13993, opened 2019-04-18
logs-async.1.0: incompatible with logs.6.3.0

dbuenzli:
It is getting it's result package dependency via Logs and the latter no longer depends on it.


dbuenzli:
Datakit doesn't seem to be happy.

avsm:
restarted job

avsm:
could you rebase this against opam-repo master? I think its because of 3044264138073fb14576d59b8278c4f8f9b43e51 which is now reverted

mseri:
Thanks

vbmithr:
Ok will fix.

PR #13996, opened 2019-04-19
try to repair the install of rdkit on osx with brew

UnixJunkie:


UnixJunkie:
Not that I would _love_ to install a properly tagged and released version of rdkit on the Mac.
But, the bew recipe for it is constantly broken and unmaintained.
This brew recipe should be taken care of by competent people.
So, the only way I currenlty know that consistently work is to force installing the HEAD...

mseri:
I would rather leave the burden and decision to the user.
For example one could leave a post-install failure message for osx that recommend to install a working rdkit version before retrying and suggests, if the use uses brew, to try the commands you bave used there

mseri:
For example I don’t like that a package messes around with my pip environment

kit-ty-kate:
Why create a new version again? Why not just fix the current one?

UnixJunkie:
I'll fix this PR on Monday.
@kit-ty-kate because, if we bump the version number, then people who upgrade automatically install the fixed package.

kit-ty-kate:
@UnixJunkie it's also what opam does when a package description changes. There is only downsides to duplicates versions for conf packages. It just ends-up leaving broken versions everywhere

UnixJunkie:
I updated the PR.
It is just an update of the post-message for Mac users.

UnixJunkie:
this is pretty safe to merge

kit-ty-kate:
Yes, indeed, sorry. Thanks!

PR #14000, opened 2019-04-19
Opam publish ppx protocol conv.5.0.0

andersfugmann:


andersfugmann:
The only failing test is the osx build which, as usual,  times out.

packages/ppx_protocol_conv_yaml/ppx_protocol_conv_yaml.5.0.0/opam
```diff
@@ -0,0 +1,33 @@
...
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+]
+depends: [
+  "ocaml" {>= "4.04.0"}
+  "ppx_protocol_conv" {= "5.0.0"}
+  "yaml"
+  "yaml" {with-test & >= "2.0.0"}
```
kit-ty-kate: yaml is present twice. Which one is true?

andersfugmann: Nice catch, Thanks.

kit-ty-kate:
Thanks!

PR #14001, opened 2019-04-19
[new release] cstruct-async, ppx_cstruct, cstruct, cstruct-unix, cstruct-sexp and cstruct-lwt (5.0.0)

avsm:
Access C-like structures directly from OCaml

- Project page: <a href="https://github.com/mirage/ocaml-cstruct">https://github.com/mirage/ocaml-cstruct</a>
- Documentation: <a href="https://mirage.github.io/ocaml-cstruct/">https://mirage.github.io/ocaml-cstruct/</a>

##### CHANGES:

**Security**: This release tightens bounds checks to ensure
that data outside a given view (but still inside the underlying
buffer) cannot be accessed.

- `sub` does more checks (mirage/ocaml-cstruct#244 mirage/ocaml-cstruct#245 @hannesm @talex5 review by @dinosaure)
- `add_len` and `set_len` are now deprecated and will be removed
  in a future release. (mirage/ocaml-cstruct#251 @hannesm)
- do not add user-provided data for bounds checks 
  (mirage/ocaml-cstruct#253 @hannesm, report and review by @talex5)
- improve CI to add fuzzing (mirage/ocaml-cstruct#255 mirage/ocaml-cstruct#252 @avsm @yomimono @talex5)

**Remove Unix dependency**: cstruct now uses the new `bigarray-compat`
library instead of Bigarray directly, to avoid a dependency on Unix
when using OCaml compilers less than 4.06.0.  This will break downstream
libraries that do not have a direct dependency on `Bigarray`.  Simply
fix it in your library by adding a `bigarray` dependency in your dune
file. (mirage/ocaml-cstruct#247 @TheLortex)

**Capability module**: To improve the safety of future code with stronger type
checking, this release introduces a new `Cstruct_cap` module which makes the
underlying Cstruct an abstract type instead of a record. In return for this
extra abstraction, the module can enforce read-only, write only, and read/write
buffers by tracking them as phantom type variables.  Although this library
shares an implementation internally with classic `Cstruct`, it is a significant
revision and so we will be gradually migrating to it.  Feedback on it is
welcome! (mirage/ocaml-cstruct#237 @dinosaure and many excited reviewers)

**Ppx compare functions**: A new `compare_X` function is generated for
`cenum` declarations. This respects custom ids supplied in the cenum
declaration and so is more robust than polymorphic compare (mirage/ocaml-cstruct#248 @emillon)

The CI has also been switched over to both Azure Pipelines and Drone in
addition to Travis, and as a result the tests all run on Windows, macOS,
various Linux distributions, on x86 and arm64 machines, and runs AFL
fuzz tests on the Drone cloud (mirage/ocaml-cstruct#255 @avsm).


avsm:
not for merge just yet -- need to check the revdeps

avsm:
not too much damage; https://github.com/ocaml/opam-repository/pull/14010

mrt-format needs a patch to explicitly add a `bigarray` depend; cc @mor1

hannesm:
@avsm looks to me that asn1-combinators needs a bigarray(-compat?) dependency as well (revdep conex-nocrypto is failing) -- which also looks like you don't see failures in libraries, but only when linking executables, so maybe a revdeps-with-testrun is worth it!?

TheLortex:
@hannesm indeed failures caused by the switch to bigarray-compat will be runtime failures.
For asn1-combinators it's fairly easy to fix: https://github.com/dune-universe/ocaml-asn1-combinators/commit/03825a368d400309ec3237a5b28eb54651a35054

avsm:
oh dear, time to patch mirage-ci. I guess we might as well run all builds with `opam install -t` ...

avsm:
opam-ci now add tests for reverse dependencies @hannesm.  There is a slight blocker in #14036 being merged and retested in order to install test dependencies, but the current results should spot some link errors i hope

mor1:
`mrt-format` updated, PR at https://github.com/ocaml/opam-repository/pull/14041

mseri:
@avsm do you want to wait for the ci to be back before doing anything here?

avsm:
Yeah, ill try to fix the CI today -  very odd lack of notifications

PR #14006, opened 2019-04-22
links: fix dependencies on the ocaml version

gasche:
Versions of links older than 0.7.2 requires a <4.06 upper boud, and 0.7.3 requires a <4.07 upper bound.

CI failure log for 0.7.3: https://ci.ocamllabs.io/log/saved/docker-run-0bcc57fbc508a2dbc928a0da9774370d/5bb974aaddfb7beb2a0f5793d75cee313067a22c

I tested manually that links.0.7.1 does not build with ocaml.4.07.1 (so adding the upper bound seems right).

### DIFF
--- a/packages/links/links.0.6.1/opam
+++ b/packages/links/links.0.6.1/opam
@@ -41,7 +41,7 @@ remove: [
   ["rm" "-rf" "%{links:doc}%/README"]
 ]
 depends: [
-  "ocaml" {>= "4.04.0"}
+  "ocaml" {>= "4.04.0" & < "4.06"}
   "ocamlfind" {build}
   "deriving" {build}
   "lwt"
--- a/packages/links/links.0.7.1/opam
+++ b/packages/links/links.0.7.1/opam
@@ -52,7 +52,7 @@ depends: [
   "linenoise"
   "ANSITerminal"
   "lwt"
-  "cohttp"
+  "cohttp" {< "0.99.0"}
   "websocket-lwt"
 ]
 depopts: ["mysql" "postgresql" "sqlite3"]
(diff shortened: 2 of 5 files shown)


gasche:
It looks like older versions of links (<=0.7.1) are now failing to build in the CI due to a missing Cohttp upper bound:

```
# File "proc.mli", line 53, characters 27-44:
# Error: Unbound module Cohttp_lwt_body
```

I'm quite confident this is not related to the present PR, which could be merged nonetheless.

This build failure was fixed in the upstream repository in https://github.com/links-lang/links/commit/9d0fa87f9d0635a3f013d7664fd3b707b9c430ea (first in 0.7.3), which doesn't say which version of cohttp introduced the incompatibility -- changing `Cohttp_lwt_body` into `Cohttp_lwt.Body`.

gasche:
I would guess that the Cotthp change that requires this fix is https://github.com/mirage/ocaml-cohttp/commit/99ce2af466870287b8bca064afca2c1cbabfe2ab#diff-ecf7627dafd277c80e22fac0168d8fce , first released in 1.0.0. I'll update this PR with a cohttp<1.0.0 upper bound for older links releases.

gasche:
PR updated; I found out that this change had already been applied for links<=0.6.1 in 2ae95716868086277b09f0d53e7a9ad8a3cefa8b. It looks like the maintainers did not change their upstream dependencies after repository-side change, so they kept pushing new versions with incorrect dependencies.

gasche:
The only remaining failure in the CI comes from a dependency-solver timeout on an old version of links. This PR should be good to go.

PR #14011, opened 2019-04-23
Package ocolor.1.0

marc-chevalier:
### `ocolor.1.0`

---

---

### DIFF
--- a/packages/ocolor/ocolor.1.0/opam
+++ b/packages/ocolor/ocolor.1.0/opam
@@ -1,0 +1,35 @@
+opam-version: "2.0"
+synopsis: "Print with style in your terminal using Format's semantic tags"
+maintainer: "Marc Chevalier <ocolor@marc-chevalier.com>"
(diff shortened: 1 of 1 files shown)


packages/ocolor/ocolor.1.0/opam
```diff
@@ -0,0 +1,35 @@
+opam-version: "2.0"
+synopsis: "Print with style in your terminal using Format's semantic tags"
+maintainer: "Marc Chevalier <ocolor@marc-chevalier.com>"
+authors: "Marc Chevalier <ocolor@marc-chevalier.com>"
+depends: [
+  "ocaml" {>= "4.05.0"}
+  "dune" {>= "1.6.3"}
```
kit-ty-kate:
```suggestion
  "dune" {build & >= "1.6.3"}
```
This should probably be tagged as a build dependency.

kit-ty-kate:
Thanks

marc-chevalier:
Thanks to you. I get lost in opam documentation and missed the build dependency flag.

PR #14012, opened 2019-04-23
Fix various hardcaml issues

xclerc:
The script used to add upper bounds to the reverse dependencies
of Jane Street packages added a number of incorrect bounds.

The "hardcaml" package exists in two sets of versions, "1.x" and
"v0.x", and the script should have added upper bounds only for
the latter (versions "1.x" are not Jane Street packages). This results
in nonsensical constraints such as `{>= "1.1.0" & < "v0.13.0"}`.

This pull request fixes such constraints, and also adds a conflict
between the "hardcaml-waveterm" and "hardcaml_waveterm" packages.


kit-ty-kate:
Thanks!

PR #14016, opened 2019-04-23
Update ocaml-migrate-parsetree for OCaml 4.08+beta3

xclerc:
The AST has changed between beta2 and beta3 (precisely to
allow ocaml-migrate-parsetree to support OCaml 4.02), so the
package needs to be updated.


packages/ocaml-migrate-parsetree/ocaml-migrate-parsetree.1.3.1/opam
```diff
@@ -0,0 +1,34 @@
...
+doc: "https://ocaml-ppx.github.io/ocaml-migrate-parsetree/"
+tags: [ "syntax" "org:ocamllabs" ]
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+]
+depends: [
+  "result"
+  "ppx_derivers"
+  "dune" {build & >= "1.6.0"}
+  "ocaml" {>= "4.02.3" & < "4.09.0"}
```
kit-ty-kate: why are those bound here if the only compiler it can be installed with is `4.08.0+beta3`?

kit-ty-kate: issue fixed.

xclerc: I think this version is pretty close to the "final" one,
so I thought we would just drop the line (the `conflicts`
after your change) once 4.08.0 is out.

kit-ty-kate: I don't think it costs us anything to allow it already. Also it allows people using other variants of `ocaml.4.08.0+beta3` (with `flambda` or whatever) to use it out of the box

kit-ty-kate:
I just realized, why are those versions on your personal repo only and not on the official one? (at least as far as I can see)
Has the official repo changed? If so it should be changed in the appropriate fields in the opam file

xclerc:
My understanding is that this code was mainly for testing
the upcoming release, and at this point it has not been
reviewed by the maintainers of the project. I can try and
create a branch on `ocaml-ppx/ocaml-migrate-parsetree`
if you think it is better.

kit-ty-kate:
For now it's ok, but I'd be very much in favour of keeping `1.3.0` as a placeholder package for when your branch is going to be merged in `ocaml-ppx/ocaml-migrate-parsetree`'s master and released. Otherwise it just creates an unused package just here for testing that is probably not even going to be released since it's broken (https://github.com/ocaml/opam-repository/pull/14017 shows that `zarith-ppx.0.1` and `ppx_bigarray.3.0.0` were broken with `1.3.0` at least)

xclerc:
You are right, sorry for that.

kit-ty-kate:
I removed my commit and added a commit changing the name of the package to match with #14017. Is that ok?

cc @avsm

PR #14019, opened 2019-04-23
Add conf-clang and fix dependencies of some bap packages

kit-ty-kate:

cc @gitoleg @ivg


gitoleg:
cc @kit-ty-kate  looks reasonable!

Issue #14023, opened 2019-04-23
Checksums for charrua-core.0.{1,3,4} have changed


kit-ty-kate:
Detected with [check.ocamllabs.io](http://check.ocamllabs.io)

@hannesm it seems the checksums have changed for those releases. Could you look into that?

PR #14025, opened 2019-04-23
charrua: update checksums

hannesm:
@kit-ty-kate it looks like the repository rename `charrua-core` -> `charrua` lead to regeneration of github-provided tarballs, which were used in those earlier releases 

I downloaded the tarballs and checked with `diff -ur` against the respective git tag.

as requested in #14023


kit-ty-kate:
Thanks a lot!

Issue #3000, opened 2014-11-03
zarith and conf-gmp need to be smarter on macos x


avsm:
gmp.h is in `/usr/local/include` but there's no pkg-config.  So to compile `zarith` on MacOS X I need to:

```
LDFLAGS=-L/usr/local/lib CFLAGS=-I/usr/local/include opam install zarith -v
```

I suspect we need to pkg-config up gmp to make this opam-friendly

amirmc:
I just hit this issue again on a fresh switch (and came back here for the appropriate incantation).  How do we fix this?

aluuu:
Ran into this issue during following Mirage [tutorial screencast](https://mirage.io/blog/getting-started-screencasts). 
Defining `LDFLAGS` and `CFLAGS` worked for me.

MagnusS:
It's not possible to override CFLAGS and LDFLAGS on OS X anymore, as they are now set to `/usr/local/*` in [opam](https://github.com/ocaml/opam-repository/blob/master/packages/zarith/zarith.1.4/opam). If homebrew/gmp.h is installed in a different location than `/usr/local` installation fails.

jbulow:
Macports installs in /opt/local/include. Does this mean that opam is completely broken on OSX by design?

francesco-zappa-nardelli:
Dear all, I have encountered the same problem of jbulow: 

> Macports installs gmp.h in /opt/local/include. Does this mean that opam is broken on OSX and it will not be fixed?

This is a showstopper for me.  Is there any workaround (using brew instead of macport is not a workaround)?  Can it / will it be fixed?  Thanks.  

-francesco

jbulow:
Still broken in opam2-beta

psilord:
I have run into the same problem as jbulow as well, any suggestions on a decent workaround to tell opam that /opt/local and /usr/local can both exist at the same time? Thank you!

dbuenzli:
So basically on `macOS` by checking two locations it seems that we catch `99%` of the install base of `gmp`. 

So in #9728 I simply move to the solution of checking both locations. If I got my shell quoting right I also reinstated the ability @MagnusS mentioned was lost to explicitely specify LDFLAGS and and CFLAGS.

This may not be the best solution but it should work reasonably well for the time being.

[...]

PR #14027, opened 2019-04-24
malfunction v0.3

stedolan:
As well as adding malfunction v0.3, this PR also removes the base-flambda package: it was only ever used by malfunction, and malfunction no longer has a hard dependency on flambda. (The relevant version constraints were inlined directly in the opam files of old versions of malfunction, so nothing changes there).


gasche:
I'm happy to learn about a new release of malfunction, do you plan to someday describe the changes in the language somewhere? Here is what I can gather from the github commit log:

- support for OCaml 4.04 to 4.07 (dropped support for 4.03)
- support for non-flambda setups
- support for Lazy
- some floating-point support

Have you considered writing, like, release notes, and telling people about it?

stedolan:
The build failure on travis is due to zarith producing a spurious warning on OS X (see https://github.com/ocaml/dune/issues/883, https://github.com/ocaml/opam-repository/issues/3000). It's only breaking here because the malfunction tests check exact command output, and some of them include this warning.

Since this isn't a bug in malfunction, and the malfunction command works fine even with the bug (although you get a spurious warning), I don't think this should block merging. Any maintainers agree/disagree?

@gasche 
> Have you considered writing, like, release notes,

There's a CHANGES.md file in the repo (which contains nothing more than what you've gleaned from git logs), and `docs/spec.md` is kept in sync with language changes.

> and telling people about it?

This is a dangerous move, and runs a serious risk of acquiring users :)

gasche:
Thanks for the CI analysis. Merging.

stedolan:
Thanks!

PR #14033, opened 2019-04-24
2 packages from issuu/ppx_mysql at 1.0

darioteixeira:
This pull-request concerns:
-`ppx_mysql.1.0`: Syntax extension for facilitating usage of MySQL bindings
-`ppx_mysql_identity.1.0`: Convenience package for using ppx_mysql with Mysql and the identity monad for IO

---
* Homepage: https://github.com/issuu/ppx_mysql
* Source repo: git+https://github.com/issuu/ppx_mysql.git
* Bug tracker: https://github.com/issuu/ppx_mysql/issues

---

### DIFF
--- a/packages/ppx_mysql/ppx_mysql.1.0/opam
+++ b/packages/ppx_mysql/ppx_mysql.1.0/opam
@@ -1,0 +1,32 @@
+opam-version: "2.0"
+maintainer: "Dario Teixeira <dte@issuu.com>"
+synopsis: "Syntax extension for facilitating usage of MySQL bindings"
+description: """
+This syntax extension aims to reduce the pain and boilerplate associated with
+using MySQL bindings from OCaml.  It is similar in spirit to PG'OCaml, but
+without the compile-time communication with the DB engine for type inference.
+"""
+homepage: "https://github.com/issuu/ppx_mysql"
+dev-repo: "git+https://github.com/issuu/ppx_mysql.git"
+bug-reports: "https://github.com/issuu/ppx_mysql/issues"
+doc: "https://issuu.github.io/ppx_mysql/"
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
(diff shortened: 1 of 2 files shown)


darioteixeira:
This PR needs the good judgment of human eyes. It's failing for older versions of the compiler because it depends on a version of OCamlformat that requires OCaml >= 4.06. I can edit the package's constraint to make it require also OCaml >= 4.06 -- and thus make Travis happy -- but that feels wrong.

darioteixeira:
So in the meantime I did add a constraint on OCaml >= 4.06. 
The only build that still fails is for OSX, but it seems like a false alarm.

mseri:
Do you remember what was the failure with older compilers?

darioteixeira:
> Do you remember what was the failure with older compilers?

The `ppx_mysql` package depends on `OCamlformat >= 0.9` (only `with-test` though), which is not available for older compilers.

mseri:
Oh ok. But then we could simply disable tests for ocaml < 4.06

mseri:
You just need to add `& ocaml:version < “4.06.0”` to the build-test filter in the build stanza

darioteixeira:
Thanks @mseri. Though don't you mean `>=` instead of `<`? If I understood your suggestion correctly, I ought to replace line `["dune" "runtest" "-p" name "-j" jobs] {with-test}` with `["dune" "runtest" "-p" name "-j" jobs] {with-test & ocaml:version >= "4.06.0"}`, right? (And restore the old bound on the compiler, of course)

darioteixeira:
Though mind you, I have no problems in keeping the `>= 4.06.0` constraint, even if it's only required for testing. It's not like I actually tested the project with older versions...

mseri:
Yes, you are right! If you don't mind keeping 4.06 I am also happy to merge as is. Let me know what you prefer

darioteixeira:
I say it gets merged as is. Thanks in advance! :-)

PR #14035, opened 2019-04-24
coqide: use 'version' to automatically depend on the matching coq version

RalfJung:
This also means we can just copy-paste that part in the future.


kit-ty-kate:
Thanks!

PR #14043, opened 2019-04-25
Add version: fields to lwt.4.2.x opam files

aantron:
See https://github.com/ocsigen/lwt/issues/690.

### DIFF
--- a/packages/lwt/lwt.4.2.0/opam
+++ b/packages/lwt/lwt.4.2.0/opam
@@ -1,6 +1,7 @@
 opam-version: "2.0"
 
 synopsis: "Promises and event-driven I/O"
+version: "4.2.0"
 
 license: "MIT"
 homepage: "https://github.com/ocsigen/lwt"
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
wait, how does anything is those opam files affect builds? AFAIK it is not copied in the build directory and if anything is to be changed it's the opam file inside the archives. Maybe to be patched instead? Did you check it actually fixes the issue? I mean I don't know every bit of opam but I'd be very surprised if that did anything tbh. But maybe I'm wrong.

mseri:
I may be he one that did object about their redundancy. I don’t think this fixes the issues, as also @kit-ty-kate pointed out. I think the problem is that the opam file in the source tarball has incorrect metadata (and that creeps up to the META file because dune parses the local opam files afaik): https://github.com/ocsigen/lwt/blob/045d64bbeb52b204e8f19f65eaf2cc6cd43c90af/lwt.opam#L5

mseri:
Here is where it happens: https://github.com/ocaml/dune/blob/817f6ed220c3e519a41b1d9a871d2db0da0907d6/src/install_rules.ml#L71

mseri:
I think to solve this we should add patches to replace the version fields (or, even better, remove them) from the opam files in the sources.

kit-ty-kate:
Closing. Feel free to reopen if necessary

PR #14045, opened 2019-04-25
Package orewa.0.1.0

Leonidas-from-XIV:
After tinkering with my new Redis bindings I thought it might be time to do an initial release to let people play with it, since it supports a pretty useful subset of Redis commands already. More to follow soon but I think it is nice to get the word out it exists.


thomasblanc:
I'm doing some optimization tests on Camelus, which automatically modified this post in an erroneous way. The installability check should indicate that there is one new installable package.

Sorry for the inconvenience. Won't happen again.

kit-ty-kate:
Thanks!

PR #14049, opened 2019-04-26
Corrected checksums for frenetic.1.0 and ox.1.0.0

arjunguha:
I accidentally deleted this tarballs from my homepage and had to rebuild them.


mseri:
Checksum is correct now, failures probably were already there and can be fixed by a separate PR

PR #14055, opened 2019-04-29
add upper bound in preparation for the next major release of zed, lambda-term and utop

kandu:
Hi, We are working on improving unicode support in zed, lambda-term and utop and are going to release major versions of these projects which will introduce backward incompatible API changes. see https://github.com/ocaml-community/lambda-term/issues/2#issuecomment-487348878

Reverse dependencies without an upper bound will fail the compilation when these new major versions release.

### DIFF
--- a/packages/ocp-index/ocp-index.1.1.1/opam
+++ b/packages/ocp-index/ocp-index.1.1.1/opam
@@ -23,7 +23,7 @@ depends: [
   "re"
   "cmdliner"
 ]
-depopts: "lambda-term"
+depopts: "lambda-term" {< "2.0"}
 conflicts: "lambda-term" {< "1.7"}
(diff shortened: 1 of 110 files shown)


packages/ocp-index/ocp-index.1.1.1/opam
```diff
@@ -23,7 +23,7 @@ depends: [
   "re"
   "cmdliner"
 ]
-depopts: "lambda-term"
+depopts: "lambda-term" {< "2.0"}
```
kit-ty-kate: You put constraints in `depopts`, this is not what you want. You want to use the `conflicts` field instead.

kandu: Thanks, Kate.

kit-ty-kate:
Thanks!

PR #14057, opened 2019-04-29
Fix packages requiring pkg-config

kit-ty-kate:

cc @toots


toots:
Thanks!

PR #14060, opened 2019-04-30
Move Ke v0.3 to Ke 0.3

dinosaure:
Fast fix about `ke` and a misnaming version. In my back, I did this PR mirage/ke#6 on my side.


kit-ty-kate:
sorry for not noticing the problem in the first place... Thanks.

PR #14067, opened 2019-05-01
Upgrade the torch packages.

LaurentMazare:
This PR updates the torch and libtorch packages to be compatible with PyTorch v1.1.0.


LaurentMazare:
Looks like CI is failing on the mac build, I'll try to get access to a mac to debug this.

LaurentMazare:
The mac build should now be fixed.
The 2 remaining errors are caused by me tweaking the old version of the torch packages (torch-0.1 to torch-0.3) so that they don't use libtorch-1.1.0 as they are not compatible with it. Installing torch-0.1 seems to fail because of the -fPIC flag missing, I don't think this is much of an issue as this flag is present on the latest torch package (torch-0.4) which works fine.

mseri:
Thanks!

PR #14068, opened 2019-05-01
+ qfs 0.10

ygrek:
build on CI is expected to fail due to absence of qfs package (same as all previous PRs)


mseri:
Thanks

aantron:
Is there any way to exclude such packages from revdeps builds? Alternatively, we could install qfs in the build containers, but going down this path in the long term seems like it would be a maintenance headache.

mseri:
@kit-ty-kate had a plan for this using some special opam flags. I don’t know what is the status of it and if itt only affects the check.ocamllabs.io service

PR #14065, opened 2019-05-01
Restore ocaml-variants beta versions

dra27:
Following a proposal on caml-devel, this PR:

 - Adds the `compiler` flag to the `ocaml` packages (@rjbou - I believe this should have been there all along; it was being added to the new `ocaml` packages?)
 - Adds the unavailable package `ocaml-beta.disabled`
 - Moves all other packages from ocaml-beta-repository back to opam-repository, but with a dependency for the ocaml-variants packages on the `ocaml-beta` package (see https://github.com/ocaml/ocaml-beta-repository/pull/8)

Users installing betas should notice no changes, since you still need to add the ocaml-beta-repository in order to install beta releases. However, this change in layout means that ocaml-beta-repository never needs amending again, which reduces the risk of package duplication, reduced CI testing, selecting wrong branches, etc.

PR #14070, opened 2019-05-02
Pass lint for ocaml-beta.disabled

dra27:
There was a lint error in `ocaml-beta.disabled` in #14065 (I was using an old opam version on the VM I prepared the PR on 😊).

Here is a lint-passing alternative, relying on the behaviour of undefined variables rather than undefined operating system names!


kit-ty-kate:
Thank you

PR #14078, opened 2019-05-02
[new release] dune (1.9.2)

rgrinberg:
Fast, portable and opinionated build system

##### CHANGES:

- Put back library variants in development mode. We discovered a
  serious unexpected issue and we might need to adjust the design of
  this feature before we are ready to commit to a final version. Users
  will need to write `(using library_variants 0.1)` in their
  `dune-project` file if they want to use it before the design is
  finalized. (ocaml/dune#2116, @diml)

- Forbid to attach a variant to a library that implements a virtual
  library outside the current project (ocaml/dune#2104, @rgrinberg)

- Fix a bug where `dune install` would install man pages to incorrect
  paths when compared to `opam-installer`. For example dune now
  installs `(foo.1 as man1/foo.1)` correctly and previously that was
  installed to `man1/man1/foo.1`. (ocaml/dune#2105, @aalekseyev)

- Do not fail when a findlib directory doesn't exist (ocaml/dune#2101, fix ocaml/dune#2099, @diml)

- [coq] Rename `(coqlib ...)` to `(coq.theory ...)`, support for
  `coqlib` will be dropped in the 1.0 version of the Coq language
  (ocaml/dune#2055, @ejgallego)

- Fix crash when calculating library dependency closure (ocaml/dune#2090, fixes ocaml/dune#2085,
  @rgrinberg)

### DIFF
--- a/packages/dune/dune.1.9.2/opam
+++ b/packages/dune/dune.1.9.2/opam
@@ -1,0 +1,50 @@
+opam-version: "2.0"
+maintainer: "opensource@janestreet.com"
+authors: ["Jane Street Group, LLC <opensource@janestreet.com>"]
+homepage: "https://github.com/ocaml/dune"
+bug-reports: "https://github.com/ocaml/dune/issues"
+dev-repo: "git+https://github.com/ocaml/dune.git"
+license: "MIT"
+depends: [
+  "ocaml" {>= "4.02"}
+  "base-unix"
+  "base-threads"
+]
+build: [
+  # opam 2 sets OPAM_SWITCH_PREFIX, so we don't need a hardcoded path
+  ["ocaml" "configure.ml" "--libdir" lib] {opam-version < "2"}
+  ["ocaml" "bootstrap.ml"]
+  ["./boot.exe" "--release" "--subst"] {pinned}
+  ["./boot.exe" "--release" "-j" jobs]
+]
+conflicts: [
+  "jbuilder" {!= "transition"}
+  "odoc" {< "1.3.0"}
+]
+
+synopsis: "Fast, portable and opinionated build system"
+description: """
+dune is a build system that was designed to simplify the release of
+Jane Street packages. It reads metadata from "dune" files following a
+very simple s-expression syntax.
+
+dune is fast, it has very low-overhead and support parallel builds on
+all platforms. It has no system dependencies, all you need to build
+dune and packages using dune is OCaml. You don't need or make or bash
+as long as the packages themselves don't use bash explicitly.
+
+dune supports multi-package development by simply dropping multiple
+repositories into the same directory.
+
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
@avsm I (especially) cannot check this PR properly without mirage-ci. Would it be possible to revert to version of cohttp previously installed on the server or something like that, to have it working again?

hannesm:
thanks @rgrinberg - is the [issue we had with mirage-xen](https://github.com/mirage/mirage/issues/969#issuecomment-482226514) in respect to https://github.com/ocaml/dune/pull/2041/ fixed in 1.9.2?

rgrinberg:
@hannesm I don't think so. I will try and prepare something for 1.9.3

kit-ty-kate:
Can someone add the appropriate constraint if that's the case?

mseri:
Should we close this and move on directly with 1.9.3?

rgrinberg:
What is the reasoning for skipping a release?

mseri:
There is a known issue, and the fixed release is already ready

mseri:
We could even release 1.9.3 as 1.9.2, I have nothing against that but I don’t know if you are happy about retagging it.

rgrinberg:
> There is a known issue, and the fixed release is already ready

The known issue also existed in 1.9.1

I'm not happy about re-tagging it because it introduces problems for other users (those who package dune for other package managers) and I don't see the need.

To be honest, I'm not sure if there's anything relevant to discuss here with the maintainers of opam-repository. The release cycle of dune is already agreed upon within the dune team, the goal is to make sure that a release does introduce problems for the rest of the packages in the opam-repository. I think we've already established that.

mseri:
Fine with me, I was asking for opinions, there are packages that prefer to jump the release in case of known issues

avsm:
I would like to test revdeps on this before merging -- I'm still trying to diagnose where the CI is failing (its sadly not as simple as reverting it) but am making progress on finding where the notification is being gobbled.

kit-ty-kate:
I don't think this one should be merge since `1.9.3` has been merged already, or at least without the proper constraints on `mirage-xen`

rgrinberg:
If this PR isn't merged then it will create a discrepancy between opam and other package managers that do not have a problem with packaging dune 1.9.2. I'm not sure it's such a bad thing, but it is a bit weird that opam only contains a subset of the dune versions.

avsm:
There's no objection to merging 1.9.2; just that the mirage-xen conflict also needed to be extended to include this new version or else breakage would ensue for users.  We should indeed have all released versions of dune in opam-repository.

I've done the mirage-xen fix in https://github.com/ocaml/opam-repository/pull/14125, and then this PR can be merged after that goes in

PR #14080, opened 2019-05-03
[new release] tcpip (3.7.5)

hannesm:
CHANGES:

* drop IPv4 packets which destination address is not us or broadcast (mirage/mirage-tcpip#407 by @hannesm)

### DIFF
--- a/packages/tcpip/tcpip.3.7.5/opam
+++ b/packages/tcpip/tcpip.3.7.5/opam
@@ -1,0 +1,73 @@
+opam-version: "2.0"
+maintainer:   "anil@recoil.org"
+homepage:     "https://github.com/mirage/mirage-tcpip"
+dev-repo:     "git+https://github.com/mirage/mirage-tcpip.git"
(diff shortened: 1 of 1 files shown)


hannesm:
the CI failure on macOS was present in earlier tcpip releases as well, I'd be happy to debug this, but don't have access to a computer (or VM) with that respective environment

mseri:
@hannesm I can check it on mac. I am going to merge, and if I see a bug I'll get in touch with you (maybe with a fix)

mseri:
@hannesm I tested the package: the tests are passing outside the sandbox and failing inside. I think we could disable them on osx or just keep in mind that some of them will fail

hannesm:
@mseri thanks! since I don't know the underlying reason of the failure, do you think there is there a way to fix the tests to succeed within the sandbox on macOS? (I also remember there were some sandbox adjustments in recent opam releases, which opam version do you test?)

mseri:
The latest. As far as I remember the adjustments were just to have a writable temp folder, they did not touch networking. I will try to have a look as soon as I have a moment

PR #14084, opened 2019-05-04
Fix mariadb and conf-mariadb

kit-ty-kate:

cc @andrenth


kit-ty-kate:
Also it seems that there is no version compatible with OCaml 4.08 (it would be sensible to avoid having `warning-as-error` enabled for future releases), but as there are only two versions that work I cannot check the final error message for the other packages so I'm just gonna wait that they fail again in the next run of `opam-health-check`

mseri:
@kit-ty-kate can I merge or you want to run some additional tests/check?

kit-ty-kate:
The CI cannot give anything. Let's merge it and see if it improve things in opam-health-check for the next run

PR #14087, opened 2019-05-05
Various fixes for OCaml 4.08

kit-ty-kate:

cc @IndiscriminateCoding for `clarity`
cc @okeuday for `cloudi`
cc @yutopp for `loga`
cc @jberdine for `ocamlformat_support`
cc @marc-chevalier for `ocolor`
cc @jrochel for `ocsigen-i18n`
cc @akabe for `ppx_bigarray`
cc @whitequark for `ppx_include`
cc @xclerc for `ppx_view`
cc @mandel for `rmlbuild`

### DIFF
--- a/packages/aperf/aperf.0.1.1/opam
+++ b/packages/aperf/aperf.0.1.1/opam
@@ -8,7 +8,7 @@ build: [
   ["ocaml" "pkg/pkg.ml" "build" "--pinned" pinned ]
 ]
 depends: [
-  "ocaml" {>= "4.03"}
+  "ocaml" {>= "4.03" & < "4.08.0"}
   "ocamlfind" {build}
   "ocamlbuild" {build}
   "topkg" {build}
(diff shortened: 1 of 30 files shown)


okeuday:
@kit-ty-kate it looks like the `cloudi` error shows a problem with `ocaml 4.08.0+beta3`:

    File "cloudi.ml", line 165, characters 12-27:
    165 |     let _ = Queue.pop value in
                      ^^^^^^^^^^^^^^^
    Error (warning 5): this function application is partial,

(at https://github.com/CloudI/cloudi_api_ocaml/blob/103ff869d9caab4f381125389cca341a1bb37aec/cloudi.ml#L165)

kit-ty-kate:
cc @Armael for the warning in cloudi

@okeuday in any case, it would be sensible to avoid "warnings as errors" for future releases.

Armael:
@okeuday The warning/error in cloudi seems to be a false positive indeed. It is the consequence of PR https://github.com/ocaml/ocaml/pull/1974, which was added to catch mistakes of the form `let _ = Printf.printf "hello %s" in ...`, or `let _ = f x y z in ...` and afterwards `f` is refactored to take one more argument (which would silently turn the call to `f` into dead code).

In your case, the code looks correct since you are popping (and dropping) a function, but since the warning is typing-based, it looks like a partial application, and therefore the warning gets triggered.

To fix the issue, you can add a type annotation: `let _ : type_of_a_callback = Queue.pop value in ... ` or give a name to the result, starting with an underscore: `let _callback = Queue.pop value in ...`.

okeuday:
@Armael thanks for the information

Issue #42, opened 2012-09-14
Unstable repository


vbmithr:
Maybe we should have an unstable version of opam-repository in addition to this one: inclusion of packages from unstable to this one would be done after careful testing on different platforms (At least Linux and OSX let’s say).

PR #14089, opened 2019-05-05
[new release] h2-mirage, h2-lwt, h2 and h2-lwt-unix (0.3.0)

anmonteiro:
##### CHANGES:

- h2-mirage: Provide `Server` and `Client` functors that take a
  `Mirage_flow_lwt.S` module as an argument
  ([#37](https://github.com/anmonteiro/ocaml-h2/pull/37))
- h2: Fix bug in the client implementation that didn't report connection
  preface errors as soon as they happened
  ([#38](https://github.com/anmonteiro/ocaml-h2/pull/38))
- h2: optimize the stream scheduler: previously when the writer yielded between
  writes, a wake up function was registered with all the (active) streams,
  which required a linear traversal of all the streams. The optimization is to
  allow every stream to wake up a global writer to which they hold a reference
  ([#40](https://github.com/anmonteiro/ocaml-h2/pull/40))
- h2: improve handling of received frames against closed streams
  ([#40](https://github.com/anmonteiro/ocaml-h2/pull/40))
- h2: in the client implementation, call the stream level error handler when
  receiving an `RST_STREAM` frame
  ([#42](https://github.com/anmonteiro/ocaml-h2/pull/42))
- h2-lwt-unix: fail earlier when setting up a SSL/TLS server without the
  depopts being available
  ([#46](https://github.com/anmonteiro/ocaml-h2/pull/46))
- h2-lwt-unix: improve the default ALPN negotiation mechanism in the SSL
  binding ([#46](https://github.com/anmonteiro/ocaml-h2/pull/46))


anmonteiro:
The only failure here seems to be a solver timeout with 4.05.

Additionally, is the removal of OPAM CI intentional? Where can I see the result of revdeps?

kit-ty-kate:
It's not intentional unfortunately. There is a chance github is blocking the IP for the server for some reason or some technical problem is occurring. I'm not the one in charge of the server so I don't know much more.

As for this specific PR, there is no revdeps for `h2` and all its subpackages so it's ok. Merging. Thanks!

anmonteiro:
Thanks!

> As for this specific PR, there is no revdeps for h2 and all its subpackages so it's ok. Merging. Thanks!

The reason I asked is because I have had to add upper bounds to some earlier versions. If it was happy in this case, good!

kit-ty-kate:
oh, I didn't think of that you're right. Mmh as a general rule for bundle of packages like this one, I think it is best to have upper-bounds already when you're releasing them. It's what janestreet packages do and it works pretty well tbh.

e.g. for this current release `"h2" {>= "0.3.0" & < "0.4.0"}`

If you're really don't want to package one of the package for any reason you can always un-tighten the constraints later. Could you do a PR that do that for earlier releases and this one if you have some time and if that correspond to your release schedule?

anmonteiro:
> Could you do a PR that do that for earlier releases and this one if you have some time and if that correspond to your release schedule?

For sure, I'll do it now. I think earlier releases already have it but I'll double check.

anmonteiro:
@kit-ty-kate done in https://github.com/ocaml/opam-repository/pull/14097

PR #14097, opened 2019-05-05
Upper bounds for h2 packages

anmonteiro:
As requested in https://github.com/ocaml/opam-repository/pull/14089#issuecomment-489453253


kit-ty-kate:
Thanks!

PR #14100, opened 2019-05-06
irmin 0.12.0 needs an old version of crunch

samoht:
Fixes http://check.ocamllabs.io/4.05.0/partial/datakit.0.8.0


mseri:
The installation in the 4.02 to 4.05 environments fails due to solver timeout. This seems to happen often lately, I am merging

samoht:
The difference between CI containers and check.ocaml.org is the solver being used. I'm wondering when the "experimental" solver used by check.ocaml.org will go upstream as it seems to be working much more reliably :-) /cc @AltGr @kit-ty-kate

PR #14102, opened 2019-05-06
datakit-github-bridge: requires github 4+ for latest

avsm:


mseri:
Travis is fine. I am relying on that

PR #14108, opened 2019-05-07
[new release] encore (0.3)

dinosaure:
Library to generate encoder/decoder which ensure isomorphism

- Project page: <a href="https://github.com/mirage/encore">https://github.com/mirage/encore</a>
- Documentation: <a href="https://mirage.github.io/encore/">https://mirage.github.io/encore/</a>

##### CHANGES:

- Fix OPAM file about how to ask to `dune` to build `encore` (@kit-ty-kate)
- Use `ke` as common implementation of ring-buffer
- Support of 4.07.0 in Travis CI
- Provide a C stubs to know if a bigarray overlaps an other
 * Compilation fixed with MirageOS (@hannesm)
- Tests about printer, parsers and combinators
- Add `from` function to let user to allocate internal buffer
- Use rev_iter instead fold and rev on internal queue
- Delete tag argument on bijection objects to be more usable
- Delete dependency of ocplib-endian and use bigstringaf instead
- Use angstrom.0.10.0 (@andreas)

### DIFF
--- a/packages/encore/encore.0.3/opam
+++ b/packages/encore/encore.0.3/opam
@@ -1,0 +1,38 @@
+opam-version: "2.0"
+name:         "encore"
+maintainer:   "Romain Calascibetta <romain.calascibetta@gmail.com>"
+authors:      "Romain Calascibetta <romain.calascibetta@gmail.com>"
+homepage:     "https://github.com/mirage/encore"
+bug-reports:  "https://github.com/mirage/encore/issues"
+dev-repo:     "git+https://github.com/mirage/encore.git"
+doc:          "https://mirage.github.io/encore/"
+license:      "MIT"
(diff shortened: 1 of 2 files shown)


dinosaure:
Reround, sorry for this duplicate but we decide internally to let this release. I hope OPAM solver will not timeout.

hannesm:
should its version really be `v0.3-1-g624af25`? to me it looks like you forgot a "dune-release tag" at the right moment...

dinosaure:
I think all is ready to merge where Travis CI fails only because it can not reach some endpoints to properly make an OCaml env. on `4.00`.

dinosaure:
`REVDEP` should be fine too where I added the constraint on `git` which is the only package which uses `encore`.

mseri:
Thanks

PR #14114, opened 2019-05-07
owl.0.2.9: safe-string constraint

pveber:
The [current check](http://check.ocamllabs.io/4.07.1/bad/owl.0.2.9) shows this version is broken due to:
```
# ocamlfind: Package `ocaml-compiler-libs' not found
```
but event after installing this library, the build fails due to:
```
#          bytes
```
So I added a constraint on the compiler, < 4.06.0 to avoid safe-string by default.


mseri:
Thanks

PR #14115, opened 2019-05-07
ppx_tools_versioned for OCaml 4.08

kit-ty-kate:
This PR fixes the constraints for ppx_tools_versioned (which doesn't support OCaml 4.08), and also add a new test version for this package from the latest stable version patched to support OCaml 4.08.

The patch to support 4.08 is taken from https://github.com/xclerc/ppx_tools_versioned/commit/25e23b3620849df4c60d99da44225ea74fdacb0c
The section for OCaml 4.07 has been removed because there is currently no stable versions of `ppx_tools_versioned` with support for the 4.07 AST (see https://github.com/ocaml-ppx/ppx_tools_versioned/issues/13)

This PR follows more or less the same package scheme than the test packages for `ocaml-migrate-parsetree` added in these PRs:
* https://github.com/ocaml/opam-repository/pull/13722
* https://github.com/ocaml/opam-repository/pull/14016

cc @xclerc @let-def

### DIFF
--- a/packages/ppx_tools_versioned/ppx_tools_versioned.5.0.1/opam
+++ b/packages/ppx_tools_versioned/ppx_tools_versioned.5.0.1/opam
@@ -12,7 +12,7 @@ build: [make "all"]
 install: [make "install"]
 remove: ["ocamlfind" "remove" "ppx_tools_versioned"]
 depends: [
-  "ocaml" {>= "4.02.0"}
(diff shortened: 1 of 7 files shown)


mseri:
@kit-ty-kate can you update the pR with just the version bounds? I have merged https://github.com/ocaml/opam-repository/pull/14141 that includes the patches, there is no longer need to backport them

PR #14118, opened 2019-05-08
ocaml compiler distribution: use parallel build for OCaml 4.07 and above

gasche:
In the old days, the OCaml compiler distribution did not support
parallel build. This support has been contributed over time, and while
the dependencies are still not 100% reliable (in particular, parallel
rebuild from a partial build often fails with cmi inconsistencies),
builds from a clean state are now very robust: they are routinely used
by developers and tested daily on our CI machines, and we haven't seen
a failure in a long while.

Enabling parallel build improves build times for everyone, at the cost
of a very low risk of breaking the build. Some context:

- On my machine, a recent sequential build took 5m15s, and a parallel
  build from the same sources took 2m53s: we can save 2m20s for every
  new switch creation for users with similar hardware. When creating
  a switch with a small number of packages, the compiler build time
  may dominate the total installation time.

- New switches are created much more often now that opam2 makes local
  switches easy. It is typical for users to create a switch for each
  project to isolate their dependencies, ending up with dozens of
  switches. The install time quickly builds up.

If a user was to find a build failure due to parallel build,
a workaround is to pass `-j 1` to opam to force sequential switch
creation:

    opam switch create <switch-name> <compiler-version> -j 1

setting the environment variable OPAMJOBS=1 would also work.

This commit was semi-automatically generated using the following script:

```
for f in ocaml-base-compiler/*4.07.*/opam; do sed -i 's/make "world/make "-j%{jobs}%" "world/g' $f; done
for f in ocaml-variants/*4.0[789].*/opam; do sed -i 's/make "world/make "-j%{jobs}%" "world/g' $f; done
for f in ocaml-variants/*4.10.*/opam; do sed -i 's/make "world/make "-j%{jobs}%" "world/g' $f; done
```

### DIFF
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.4.07.0/opam
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.4.07.0/opam
@@ -23,8 +23,8 @@ build: [
     "-aspp"
     "cc -c"
   ] {os = "openbsd" | os = "freebsd" | os = "macos"}
-  [make "world"]
-  [make "world.opt"]
+  [make "-j%{jobs}%" "world"]
+  [make "-j%{jobs}%" "world.opt"]
 ]
 install: [make "install"]
 url {
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.4.07.1/opam
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.4.07.1/opam
@@ -23,8 +23,8 @@ build: [
     "-aspp"
     "cc -c"
   ] {os = "openbsd" | os = "freebsd" | os = "macos"}
-  [make "world"]
-  [make "world.opt"]
+  [make "-j%{jobs}%" "world"]
+  [make "-j%{jobs}%" "world.opt"]
 ]
 install: [make "install"]
 url {
--- a/packages/ocaml-variants/ocaml-variants.4.07.0+32bit/opam
+++ b/packages/ocaml-variants/ocaml-variants.4.07.0+32bit/opam
@@ -42,8 +42,8 @@ build: [
     "-host"
     "i386-apple-darwin13.2.0"
   ] {os = "macos"}
-  [make "world"]
-  [make "world.opt"]
+  [make "-j%{jobs}%" "world"]
+  [make "-j%{jobs}%" "world.opt"]
 ]
 install: [make "install"]
 url {
--- a/packages/ocaml-variants/ocaml-variants.4.07.0+afl/opam
+++ b/packages/ocaml-variants/ocaml-variants.4.07.0+afl/opam
@@ -23,8 +23,8 @@ build: [
     "-aspp"
     "cc -c"
   ] {os = "openbsd" | os = "freebsd" | os = "macos"}
-  [make "world"]
-  [make "world.opt"]
+  [make "-j%{jobs}%" "world"]
+  [make "-j%{jobs}%" "world.opt"]
 ]
 install: [make "install"]
 url {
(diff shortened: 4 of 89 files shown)


gasche:
See also: previous discussion in https://github.com/ocaml/opam/issues/3455.

dra27:
Cygwin needs excluding from this - parallel builds have been broken for a while, and will only be fixed in 4.09

avsm:
@gasche I'm temporarily closing this due to the CI load it's imposing, but please do feel free to reopen when the Cygwin fix is applied

gasche:
The version I proposed is

```
build: [
  ["./configure" "-prefix" prefix "-with-debug-runtime"]
    {os != "openbsd" & os != "freebsd" & os != "macos"}
  [
    "./configure"
    "-prefix"
    prefix
    "-with-debug-runtime"
    "-cc"
    "cc"
    "-aspp"
    "cc -c"
  ] {os = "openbsd" | os = "freebsd" | os = "macos"}
  [make "-j%{jobs}%" "world"]
  [make "-j%{jobs}%" "world.opt"]
]
```

do we agree that the following should work?

```
build: [
  ["./configure" "-prefix" prefix "-with-debug-runtime"]
    {os != "openbsd" & os != "freebsd" & os != "macos"}
  [
    "./configure"
    "-prefix"
    prefix
    "-with-debug-runtime"
    "-cc"
    "cc"
    "-aspp"
    "cc -c"
  ] {os = "openbsd" | os = "freebsd" | os = "macos"}
  [make "world"] {os = "cygwin"}
  [make "world.opt"] {os = "cygwin"}
  [make "-j%{jobs}%" "world"] {os != "cygwin"}
  [make "-j%{jobs}%" "world.opt"] {os != "cygwin"}
]
```

One thing I have been wondering is why we build `world` firs then `world.opt`; the [INSTALL](https://github.com/ocaml/ocaml/blob/4.09/INSTALL.adoc) file recommends running `world.opt` directly. There is no performance difference in doing one or the other, but would we agree to simplify the build script to run world.opt directly?

(I tried to trace the addition of the "world" step in the opam-repository history, but it seems that it was there from the start and not added for a particular reason. The only justification I could find is that, for systems that don't support native compilation, running "make world" first will at least work the message will be slightly more specific; but those systems should instead be correctly configured to not run `.opt` targets at all.)

With this simplification we would have the following:

```
build: [
  ["./configure" "-prefix" prefix "-with-debug-runtime"]
    {os != "openbsd" & os != "freebsd" & os != "macos"}
  [
    "./configure"
    "-prefix"
    prefix
    "-with-debug-runtime"
    "-cc"
    "cc"
    "-aspp"
    "cc -c"
  ] {os = "openbsd" | os = "freebsd" | os = "macos"}
  [make "world.opt"] {os = "cygwin"}
  [make "-j%{jobs}%" "world.opt"] {os != "cygwin"}
]
```

pmetzger:
> would we agree to simplify the build script to run world.opt directly

That seems pretty sane to me.

gasche:
I updated my branch to carry the following diff:

```diff
--  [make "world"]
--  [make "world.opt"]
++  [make "world.opt"] {os = "cygwin"}
++  [make "-j%{jobs%}" "world.opt"] {os != "cygwin"}
```

Could someone confirm that `os = "cygwin"` is the right condition? (@kit-ty-kate?). Some packages use `os-distribution = "cygwinports"`, which tests something else, but conf-libX11 uses `os-distribution = "cygwin"`.

gasche:
@dra27 I see that you have re-appeared, so this is a gentle ping (and hi!) on whether you think the new state is good and would be worth launching the CI.

kit-ty-kate:
According to https://github.com/ocaml/opam-depext/blob/master/depext.ml#L150 and https://github.com/ocaml/opam/blob/master/src/state/opamSysPoll.ml#L62, the correct answer appear to be `os = "cygwin"`

kit-ty-kate:
Could you reopen a/the PR? Github won't let me do it on my own for some reason

gasche:
I also cannot reopen this PR, so I will open a new one. Thanks for feedback on the condition, and the motivating messages :-)

PR #14123, opened 2019-05-10
[new release] eqaf (0.3)

dinosaure:
Constant-time equal function on string

- Project page: <a href="https://github.com/mirage/eqaf">https://github.com/mirage/eqaf</a>
- Documentation: <a href="https://mirage.github.io/eqaf/">https://mirage.github.io/eqaf/</a>

##### CHANGES:

- Provide `Eqaf_bigstring`
- Provide `Eqaf_cstruct`
- New check tool and delete any dependencies on `eqaf` package (@dinosaure, @hannesm, @cfcs)

### DIFF
--- a/packages/eqaf/eqaf.0.3/opam
+++ b/packages/eqaf/eqaf.0.3/opam
@@ -1,0 +1,31 @@
+opam-version: "2.0"
+name:         "eqaf"
+maintainer:   [ "Romain Calascibetta <romain.calascibetta@gmail.com>" ]
+authors:      [ "Romain Calascibetta <romain.calascibetta@gmail.com>" ]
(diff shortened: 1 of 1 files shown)


dinosaure:
Ok, so errors occur because linear regression on time needed to compute `equal` function is too noisy in CI.

cfcs:
@dinosaure would it help to increase the amount of samples, or do you need to turn off that regression test?

dinosaure:
@cfcs I believe we need to turn off tests in some contexts

avsm:
We'll work on improving the opam ci to support isolation for tests like this, but this is fine to merge now

PR #14128, opened 2019-05-11
angstrom.0.6.0: tests fail with safe-string

avsm:
spotted in revdeps for #14001

### DIFF
--- a/packages/angstrom/angstrom.0.6.0/opam
+++ b/packages/angstrom/angstrom.0.6.0/opam
@@ -18,6 +18,7 @@ depends: [
   "ocplib-endian" {>= "0.6"}
   "result"
 ]
+conflicts: [ "ocaml" {>="4.06.0" & with-test } ]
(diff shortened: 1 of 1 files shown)


avsm:
Looks like a few revdeps need lower bounds on whenever Angstrom introduced the `parse_string` function

packages/angstrom/angstrom.0.6.0/opam
```diff
@@ -8,11 +8,10 @@ dev-repo: "git+https://github.com/inhabitedtype/angstrom.git"
 build: [
   ["jbuilder" "subst" "-p" name] {pinned}
   ["jbuilder" "build" "-p" name "-j" jobs]
-  ["jbuilder" "runtest" "-p" name] {with-test}
+  ["jbuilder" "runtest" "-p" name] {with-test & ocaml:version < "4.06.0"}
```
avsm: much better, thanks :)

PR #14129, opened 2019-05-11
tcpip: versions before 3.5.1 required mirage-clock<2 (type unit change)

avsm:
spotted in revdeps for #14001

### DIFF
--- a/packages/mirage-clock-unix/mirage-clock-unix.1.2.0/opam
+++ b/packages/mirage-clock-unix/mirage-clock-unix.1.2.0/opam
@@ -14,8 +14,8 @@ depends: [
   "ocamlfind" {build}
   "ocamlbuild" {build}
   "topkg" {build & >= "0.8.0"}
-  "mirage-clock" {>= "1.2.0"}
-  "mirage-clock-lwt" {>= "1.2.0"}
+  "mirage-clock" {>= "1.2.0" & <"1.3.0"} 
+  "mirage-clock-lwt" {>= "1.2.0" & <"1.3.0"}
   "lwt"
 ]
 build: [
--- a/packages/mirage-clock-unix/mirage-clock-unix.1.3.0/opam
+++ b/packages/mirage-clock-unix/mirage-clock-unix.1.3.0/opam
@@ -12,8 +12,8 @@ tags: ["org:mirage"]
 depends: [
   "ocaml"
   "jbuilder" {build & >= "1.0+beta9"}
-  "mirage-clock" {>= "1.2.0"}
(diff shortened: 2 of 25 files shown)


avsm:
@hannesm just running this one past you if you have a moment -- was it intentional to not constrain the earlier mirage-types after the clock.2.0.0->unit change?  This PR is conservative since it'll keep older code working (e.g. tests in tcpip)

hannesm:
@avsm I don't fully understand this: the `type t = unit` constrained was introduced (in `mirage-clock` 2.0.0), which should be fully backwards compatible (i.e. earlier users passed `PCLOCK.now_d_ps pclock`, now they can either `PCLOCK.now_d_ps ()` or `PCLOCK.now_d_ps pclock`, since `pclock = unit`). What is the error message in <tcpip.3.5.1? certainly code using `PCLOCK.now_d_ps ()` requires `mirage-clock >= 2.0.0`.

avsm:
Hm it looks like older tcpip did use the representation for the tests

```
- File "test/test_tcp_window.ml", line 20, characters 38-43:
- Error: Signature mismatch:
-        ...
-        Type declarations do not match:
-          type t = Clock.t = { time : int64; }
-        is not included in
-          type t = unit
-        File "src/mirage_clock.ml", line 54, characters 31-44:
-          Expected declaration
-        File "test/test_tcp_window.ml", line 6, characters 2-26:
-          Actual declaration
-       ocamlc test/.test.eobjs/byte/test_ipv6.{cmi,cmo,cmt}
```

So i think that ec321bf43b7afd71b1380d8beb0f4bfcd057ef02 is necessary in this PR, and 8e842e8b0345c763d1830adc9db649efd023bc67 adds consistency (there's no need to select a -unix clock package that's different from the base clock version, as we released them together). But 
b454d35 may be unnecessarily conservative

(these failures are showing up in datakit-ci only now since i activated "run tests on revdeps")

hannesm:
ah, spot on - https://github.com/mirage/mirage-tcpip/commit/c5ea1b48b3eb8b4d15befc57a6103e1d696d5a69 which according to github's ui has been in tcpip 3.6.0+ (so everything < 3.6.0 should be constrained to mirage-clock with-test & < 2.0.0)

avsm:
ok thanks for checking that. i'm  primarily wondering whether to merge b454d35866c7210d881b0c8636fd59211529dc84 now -- it seems weird to pull in a newer time interface to an older published mirage-types.  i'd be inclined to add those upper bounds in to avoid surprisingly older applications like the tcpip tests

avsm:
This seems mergeable now

mseri:
Thanks

PR #14131, opened 2019-05-13
added packages/cpm/cpm.6.0.0

UnixJunkie:
added ROC AUC for arrays; previously there was only for lists of score labels;
better opam file also.


UnixJunkie:
CI tests are all green

mseri:
Thanks

PR #14139, opened 2019-05-14
as my peers, I do have an e-mail address

UnixJunkie:


UnixJunkie:
;^)

UnixJunkie:
@rixed maybe there should be you name in there too, with an e-mail, since you are also one of the active maintainers

mseri:
@rixed feel free to send a PR adding your name if that is the case

PR #14142, opened 2019-05-14
Package ocaml-monadic.0.4.1

zepalmer:
### `ocaml-monadic.0.4.1`
A PPX extension to provide an OCaml-friendly monadic syntax

---
* Homepage: http://github.com/zepalmer/ocaml-monadic
* Source repo: git+https://github.com/zepalmer/ocaml-monadic.git
* Bug tracker: https://github.com/zepalmer/ocaml-monadic/issues

---

### DIFF
--- a/packages/ocaml-monadic/ocaml-monadic.0.4.1/opam
+++ b/packages/ocaml-monadic/ocaml-monadic.0.4.1/opam
@@ -1,0 +1,24 @@
+opam-version: "2.0"
+synopsis: "A PPX extension to provide an OCaml-friendly monadic syntax"
+maintainer: "JHU PL Lab <pl.cs@jhu.edu>"
+authors: "JHU PL Lab <pl.cs@jhu.edu>"
+license: "BSD-3-clause"
+homepage: "http://github.com/zepalmer/ocaml-monadic"
+bug-reports: "https://github.com/zepalmer/ocaml-monadic/issues"
+depends: [
+  "dune" {build & >= "1.0.0"}
+  "ocaml-migrate-parsetree"
+  "ppx_tools_versioned"
+  "ocaml" {>= "4.04.0"}
+]
+build: ["dune" "build" "-p" name "-j" jobs]
+run-test: ["dune" "runtest" "-p" name "-j" jobs]
+dev-repo: "git+https://github.com/zepalmer/ocaml-monadic.git"
+url {
+  src:
+    "https://github.com/zepalmer/ocaml-monadic/archive/dd42d4ae6e618f38dac93223bd58aa04386f9a02.zip"
+  checksum: [
+    "md5=408b32bc4b5225eb228b5494318bbdc9"
+    "sha512=35d16c6d7671fecbb698e9ef6a71bac25afc99293f90a070a4a044f4c939f95a9ebcf4c24ff64b4ebfa7c7aa9f3d5bad2b27c0617b23aa18f41912df9607cfec"
+  ]
+}


zepalmer:
Hello!  I'm trying to upgrade my packages to the OPAM 2 metadata format so I can publish updates.  The Travis CI builds for applicable versions of OCaml all fail with something akin to the following message:

```
+ /home/travis/.opam/opam-init/hooks/sandbox.sh "build" "dune" "build" "-p" "ocaml-monadic" "-j" "1" (CWD=/home/travis/.opam/ocaml-base-compiler.4.04.2/.opam-switch/build/ocaml-monadic.0.4.1)
- Error: I don't know about package ocaml-monadic (passed through --only-packages/--release)
[ERROR] The compilation of ocaml-monadic failed at
        "/home/travis/.opam/opam-init/hooks/sandbox.sh build dune build -p
        ocaml-monadic -j 1".
```

I attempted to recreate this scenario on my own machine by running the following from my local Git repository for the `ocaml-monadic` project:

```
$ opam switch create 4.04.2
$ opam install .
```

I am unable to reproduce the Travis CI error; everything installs just fine on my machine.  Can anyone advise?

mseri:
@zepalmer the url field is missing

zepalmer:
Thanks for calling that out.  I'm a little perplexed, as other projects (e.g. ppx_deriving) don't seem to have a URL in their .opam files and yet have published with v2.0 package description files.  In my case, i have been following the instructions at https://opam.ocaml.org/doc/Packaging.html , which indicate that I should run

```
git tag -a 0.4.1
git push 0.4.1
```

and then "run `opam publish` ... and follow the steps".  Doing so produces

```
The following will be published:
  - ocaml-monadic version 0.4.1 with opam file at /home/zpalmer/work/research/development/ocaml-monadic/ocaml-monadic.opam

[WARNING] These will be virtual packages: ocaml-monadic.0.4.1

Continue ? You will be shown the patch before submitting. [Y/n]
```

Is this warning about the "virtual package" the culprit here?  If so, I can use the instructions for non-GitHub-hosted projects (even though my project is hosted on GitHub) and see if that fares any better.

Thanks with your help on this!

mseri:
I am not familiar with the error. The url should usually be added automatically. Can you raise an issue on the opam-publish page or ask on the discuss forum or the discord channel?

Also, on a different note, among the guidelines for new packages is that new packages do not contain ocaml in their names Can you please also rename the package?

mseri:
I think your issue may be due to the fact that the new tag has not been pushed to the repo, but it is just a guess

zepalmer:
Thanks for the feedback!

I deleted the tag once I saw that the CI build failed.  (Using OPAM 1.2, I never used to tag until after Travis CI succeeded since something would often be a bit off with my config.)  I'd pushed the tag when the original PR was submitted.

Regarding the prefix: when I originally submitted the 0.1.0 version of this package some years ago, I made the case that the "`ocaml`" prefix wasn't just because it's an OCaml library.  The package provides an OCaml-friendly syntax for monadic code, as opposed to previously-existing PPX rewriters which take after Haskell's `do` notation.  That seemed to be sufficient reason to give the package its name at the time and there are some dependencies on it in other projects now.  Is that okay?

I'll look into the error and try giving an explicit URL later.  Thanks!  :)

mseri:
@zepalmer sorry, I double checked only after writing the comment. Given that there are already a number of releases of it for few years, and the guideline was introduced later, I don’t think you should change the name of the package at this point.

zepalmer:
I re-issued the pull request with a different command:

```
$ opam publish https://github.com/zepalmer/ocaml-monadic/archive/dd42d4ae6e618f38dac93223bd58aa04386f9a02.zip
```

I am no longer getting the "virtual package" warning, which I suspect comes from not having a URL.  (I'm guessing that, like other package management systems, a virtual package is expected to have dependencies but no build or install files of its own.)  It seems reasonable to conclude, then, that the original

```
opam publish
```

apparently failed to identify the URL of my repository from the GitHub clone in which I was operating.  I'm not sure why that would be the case, but I guess I'm not especially worried about it.  Once everything is tidy, I'll try to make sure I report to the `opam-publish` page.  Thanks again for the help!

p.s. The Xcode build on the Travis CI shows success in the terminal but has stalled.  Please do let me know if I need to do anything about that.

mseri:
You get the ugly url instead of the nice tagged url, but this time it worked. Do you want to tag and update the pr with the tagged tarball, or you would like to merge as is?

zepalmer:
I'm happy to merge as-is.  The beautification of the URL would be a bonus, but I'm not fond of having to push a tag before I know the CI builds will succeed.  So long as my odd little workflow doesn't cause anyone else any grief, I'm happy to go with it.  :)

Thanks again!

mseri:
If you have not yet done it, please consider announcing this on the discuss.ocaml.org forum. We have a community category and an announcement tag for this purpose.

zepalmer:
Thanks for the advice; I have taken it!  :)

PR #14150, opened 2019-05-16
unix-fcntl: check if upper bound is a hard constraint

UnixJunkie:


UnixJunkie:
yes it is, and the fix is not that trivial.

PR #14159, opened 2019-05-20
[new release] syndic (1.6.1)

Chris00:
CHANGES:

* `Syndic.Date.of_rfc822` accepts dates such as “May 15th, 2019”.

### DIFF
--- a/packages/syndic/syndic.1.6.1/opam
+++ b/packages/syndic/syndic.1.6.1/opam
@@ -1,0 +1,43 @@
+opam-version: "2.0"
+name:         "syndic"
+maintainer:   "Romain Calascibetta <romain.calascibetta@gmail.com>"
+authors:      [ "Romain Calascibetta"
+                "Christophe Troestler" ]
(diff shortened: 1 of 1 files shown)


Chris00:
The failure for 4.03 is with the tests — and on the line `#use "topfind" ;;`.

mseri:
I can add a filter to disable the tests on 4.03 if you want, otherwise I’ll remove the name variable and merge

Chris00:
> disable the tests on 4.03

OK, let's do that.  Thanks.
CC @dinosaure in case he has another solution (the failure must have happened with the previous version as it is unrelated to the changes).

dinosaure:
Related to: https://github.com/ocaml/ocaml/issues/7555 so we can just drop `4.03.0`.

mseri:
@Chris00 I’ll let you decide, either way is fine for me

Chris00:
I think one can only run the tests for OCaml >= 4.04

PR #14161, opened 2019-05-20
Add coq and coqide 8.9.1.

silene:


silene:
The only failure is on MacOS, due to the timeout (50' is too short to build Coq several times). It seems like the build would have failed anyway for `coqide`, since its dependency `conf-gtksourceview` errored out. So, everything looks fine to me.

mseri:
Thanks!

PR #14170, opened 2019-05-22
Add pkg idd

etw33:


packages/idd/idd.0.1/opam
```diff
@@ -0,0 +1,21 @@
+opam-version: "2.0"
+name: "idd"
+depends: [
+  "ocaml" {>= "4.04.0"}
+  "dune" {build}
+  "base"
+  "ppx_jane"
+  "ppx_inline_test"
```
mseri: Do you need this always or is it just a test dependency?

packages/idd/idd.0.1/opam
```diff
@@ -0,0 +1,21 @@
+opam-version: "2.0"
+name: "idd"
+depends: [
+  "ocaml" {>= "4.04.0"}
+  "dune" {build}
+  "base"
```
mseri: I think you need a lower bound on v1.12 o v1.13. There have been various breaking changes in the releases of base

etw33: Why v1.12 or v1.13 if v1.0 hasn't been released yet?

mseri: Sorry, ot was a typo, I meant 0.

packages/idd/idd.0.1/opam
```diff
@@ -0,0 +1,20 @@
+opam-version: "2.0"
+name: "idd"
+depends: [
+  "ocaml" {>= "4.04.0"}
+  "dune" {build}
+  "base" {>= "0.12.0"}
+  "ppx_jane"
+  "ppx_deriving"
+]
+build: ["dune" "build" "-p" name]
```
kit-ty-kate:
```suggestion
build: ["dune" "build" "-p" name "-j" jobs]
```
you might want to allow users to control the number of jobs they use for the compilation of this package

packages/idd/idd.0.1/opam
```diff
@@ -0,0 +1,20 @@
+opam-version: "2.0"
+name: "idd"
+depends: [
+  "ocaml" {>= "4.04.0"}
+  "dune" {build}
+  "base" {>= "0.12.0"}
```
kit-ty-kate:
```suggestion
  "base" {>= "v0.12.0"}
```
the `v` prefix is non-optional for recent janestreet packages

kit-ty-kate:
Thanks!

PR #14173, opened 2019-05-22
z3.4.8.4: fix build for static linking by default

AltGr:
The two patches have been ported to Z3 master and are backported
here. This makes the installed z3 lib link statically by default,
avoiding the problem with a runtime dependency to a .so file within
the switch (needing an update of LD_LIBRARY_PATH for the binary to
run)

### DIFF
--- a/packages/z3/z3.4.8.4/files/0001-OCaml-static-fix-linking-options.patch
+++ b/packages/z3/z3.4.8.4/files/0001-OCaml-static-fix-linking-options.patch
@@ -1,0 +1,39 @@
+From c1ad450f7cde79149b73f25cc632d15830b75829 Mon Sep 17 00:00:00 2001
+From: Louis Gesbert <louis.gesbert@ocamlpro.com>
+Date: Tue, 21 May 2019 16:09:18 +0200
+Subject: [PATCH 1/2] OCaml / static: fix linking options
+
+---
+ scripts/mk_util.py | 6 ++++--
+ 1 file changed, 4 insertions(+), 2 deletions(-)
+
+diff --git a/scripts/mk_util.py b/scripts/mk_util.py
+index 9076b582f..f076817d2 100644
+--- a/scripts/mk_util.py
++++ b/scripts/mk_util.py
+@@ -2253,10 +2253,12 @@ class MLComponent(Component):
+ 
+             OCAMLMKLIB = 'ocamlmklib'
+ 
+-            LIBZ3 = '-L. -lz3'
++            LIBZ3 = '-lz3'
+             if is_cygwin() and not(is_cygwin_mingw()):
+                 LIBZ3 = 'libz3.dll'
+ 
++            LIBZ3 = LIBZ3 + ' ' + ' '.join(map(lambda x: '-cclib ' + x, LDFLAGS.split()))
++
+             if DEBUG_MODE and not(is_cygwin()):
+                 # Some ocamlmklib's don't like -g; observed on cygwin, but may be others as well.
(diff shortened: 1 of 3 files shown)


AltGr:
This will allow z3 support in opam to be merged \o/

AltGr:
Re-run Travis once https://github.com/ocaml/opam-repository/pull/14173 is merged.

mseri:
Thanks

giltho:
This merge request completely broke z3 on our end :/
We were using z3 with opam and doing some tricks to link the .so / .dylib at runtime.
This merge request broke our build while the version didn't change which is kind of annoying, we have no way of correcting that on our end.
Although it would be amazing to have a completely static build of z3 ! (so thank you for the initiative).

The error we get is a very very long error ending with
```      recfun::util::util(ast_manager&) in libz3-static.a(recfun_decl_plugin.o)
      recfun::decl::plugin::u() const in libz3-static.a(recfun_decl_plugin.o)
      ...
  "___gxx_personality_v0", referenced from:
      _Z3_mk_model in libz3-static.a(api_model.o)
      _Z3_model_inc_ref in libz3-static.a(api_model.o)
      _Z3_model_dec_ref in libz3-static.a(api_model.o)
      _Z3_model_get_const_interp in libz3-static.a(api_model.o)
      _Z3_model_has_interp in libz3-static.a(api_model.o)
      _Z3_model_get_func_interp in libz3-static.a(api_model.o)
      _Z3_model_get_num_consts in libz3-static.a(api_model.o)
      ...
ld: symbol(s) not found for architecture x86_64
File "caml_startup", line 1:```

I'd be happy to make a PR to correct that :)

avsm:
Sorry about that -- we should definitely bump the version number as this package is now a backport. As for the error, it looks like you are missing a `-lstdc++` somewhere...

giltho:
your reactivity is impressive, thanks for your answer !
I'm not really an expert in C++ compilation, do you have an idea of where I should add the missing `lstdc++` ? :)

giltho:
Ah annoying...
It doesn't fail when installing through `opam` actually, but it does when installing through `esy`. The issue might be on the `esy` side for that one

avsm:
Whatever build system you're using, you need to pass `-lstdc++` to the underlying linker. For dune, that would be `(c_library_flags -lstdc++)`. But I know nothing about your setup than that little fragment above, so I'm inferring this from the missing `__gxx_personality` symbol

giltho:
I'm using dune, but it's not my project that is not building, it's only z3 that fails to build. My project cannot start its build because z3 dies before.
I guess esy's environment variable are being annoying for some reason...

giltho:
Ok, removed previous comment which was false !
Indeed you are right, I'm using dune and adding `(c_library_flags -lstdc++)` in the dune file that specifies `z3` as a dependency of the lib worked !
I though it was the build of z3 itself that was dying but it was actually the build of our project.
Isn't it weird that I have to add some c library flags to our project's dune files, while I'm using z3 as an ocaml library, and not compiling any c++ myself ?

In any case, thank you so much for your answers, it helped a lot !!!

avsm:
It's to do with how c++ generates its shared libraries. You could also compile z3 with `c++` instead of `cc` and it may have some linking magic there. But generally it's hard to make the use of c++ invisible to the linker,

c-cube:
This actually breaks my code too, I'm just realizing it. I think it also prevents Z3 to be used in a toplevel (and not one made with mktop, no thanks)?

The `-L.` was there for a good reason…

PR #14175, opened 2019-05-24
[new release] eqaf (0.4)

dinosaure:
CHANGES:

- Distribution integrate an attack example
- Fuzzer to test `equal` function
- Unroll internal loop over 16 bits integers instead 32 bits
- Put x86 ASM output in implementation (and audit)
- Do second check even if first on fails (bad r²)
- Avoid indirection to `Pervasives` functions


kit-ty-kate:
Thanks!

PR #14179, opened 2019-05-25
arp{-mirage} >= 1.0.0 has dune as build dependency

hannesm:
as pointed out by @TheLortex in https://github.com/mirage/arp/pull/15


mseri:
The CI error should be due to osx sandbox

PR #14184, opened 2019-05-27
[new release] functoria and functoria-runtime (2.2.4)

hannesm:
CHANGES:

* fix app_info - executing "opam list --installed" (mirage/functoria#170, by @hannesm)


hannesm:
test failure is fine -- looks like some paths in myocamlbuild needs to be fixed for the tests, I tried and failed. the upside is that functoria-master uses a different approach (using dune instead of ocamlbuild) already

kit-ty-kate:
ok fair enough. Thanks!

PR #14186, opened 2019-05-27
ocaml-compiler-libs-v0.12.0

xclerc:
Switch to dune (using `dune upgrade`).

### DIFF
--- a/packages/ocaml-compiler-libs/ocaml-compiler-libs.v0.12.0/opam
+++ b/packages/ocaml-compiler-libs/ocaml-compiler-libs.v0.12.0/opam
@@ -6,7 +6,7 @@ bug-reports: "https://github.com/janestreet/ocaml-compiler-libs/issues"
 dev-repo: "git+https://github.com/janestreet/ocaml-compiler-libs.git"
 license: "Apache-2.0"
 build: [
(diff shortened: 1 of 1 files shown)


packages/ocaml-compiler-libs/ocaml-compiler-libs.v0.12.0/opam
```diff
@@ -0,0 +1,23 @@
+opam-version: "2.0"
+maintainer: "opensource@janestreet.com"
+authors: ["Jane Street Group, LLC <opensource@janestreet.com>"]
+homepage: "https://github.com/janestreet/ocaml-compiler-libs"
+bug-reports: "https://github.com/janestreet/ocaml-compiler-libs/issues"
+dev-repo: "git+https://github.com/janestreet/ocaml-compiler-libs.git"
+license: "Apache-2.0"
+build: [
+  ["jbuilder" "build" "-p" name "-j" jobs]
```
avsm:
```suggestion
  ["dune" "build" "-p" name "-j" jobs]
```

xclerc: Indeed... sorry for that!

PR #14180, opened 2019-05-26
[new release] cppo_ocamlbuild and cppo (1.6.6)

avsm:
Plugin to use cppo with ocamlbuild

- Project page: <a href="http://mjambon.com/cppo.html">http://mjambon.com/cppo.html</a>
- Documentation: <a href="https://ocaml-community.github.io/cppo/">https://ocaml-community.github.io/cppo/</a>

##### CHANGES:

- [pkg] port build system to dune from jbuilder.
- [pkg] upgrade opam metadata to 2.0 format.
- [pkg] remove topkg and use dune-release.
- [compat] Use `String.capitalize_ascii` to remove warning.

PR #14187, opened 2019-05-27
older decoders releases do not build with new dune

avsm:
due to ppx changes, but newer decoders releases work fine

found in revdeps for #14180


mseri:
Thanks

PR #14194, opened 2019-05-27
Ppxlib 0.8.0

xclerc:
Ppxlib used to embed its own (patched) copy of the parser,
but now uses the parser from the compiler.

note: this change was not part of 0.7.0 (also submitted today)
because we expect 0.7.0 to never break user code while 0.8.0
can possibly (although in corner cases) break code.


kit-ty-kate:
Thanks!

PR #14196, opened 2019-05-27
Various fixes for cstruct 5

avsm:
see #14001 for the revdeps failures


avsm:
Failure was an actual test failure, not packaging, rest looks fine

PR #14199, opened 2019-05-28
added packages/bst/bst.7.0.0

UnixJunkie:
there is a simplify function now

you can use it to down-sample a point set, in case the tree
was constructed with a bucket size > 1


kit-ty-kate:
Thanks!

PR #14200, opened 2019-05-28
ppxlib.0.5.0: not compatible with compiler 4.08

pveber:
Compilation under 4.08 compiler triggers many errors, among others this one:
```
File "ast/syntaxerr.ml", line 37, characters 10-93:
37 | ..........Location.errorf ~loc:opening_loc
38 |             "This '%s' might be unmatched" opening
Error: This expression has type Location/1.report
       but an expression was expected of type
         Location/2.msg = (Format.formatter -> unit) Location/1.loc
```
There used to be a type `Location.error` which has apparently been refined in two types `msg` and `report`. So I guess it's safe to say that this version of ppxlib is not compatible with 4.08.


kit-ty-kate:
Thanks!

PR #14201, opened 2019-05-28
ppxlib.0.6.0: not compatible with compiler 4.08

pveber:
Same problem than described in #14200


xclerc:
Thanks; `ppxlib` indeed supports OCaml 4.08 in 0.7.0+.

kit-ty-kate:
Thanks!

PR #14205, opened 2019-05-28
fix some bounds on faraday packages

damiendoligez:
* `faraday-lwt-unix.0.4.0` doesn't compile with recent versions of `lwt`
* `faraday-lwt.0.5.1` needs at least `faraday.0.5.0` (and not 0.3.0).

cc @seliopou


kit-ty-kate:
Thanks!

PR #14206, opened 2019-05-28
labltk: add bound on OCaml version

damiendoligez:
`labltk` doesn't compile on 4.08 because of the changes to configure.

reported upstream at https://forge.ocamlcore.org/tracker/index.php?func=detail&aid=1768&group_id=343&atid=1351

cc @garrigue @shindere


garrigue:
I'll try to make a new release shortly. ocamlbrowser is probably broken too.

kit-ty-kate:
Thanks!

PR #14208, opened 2019-05-28
Add an upper bounds to ppxlib reverse dependencies

xclerc:
The next release of ppxlib (0.9.0) will change the selected
AST from 4.07 to 4.08. It will hence break *a lot* of code.

This pull request adds an upper bounds to all reverse
dependencies, so that the next version can be released
safely.


kit-ty-kate:
Thanks!

PR #14209, opened 2019-05-29
added packages/conf-lz4/conf-lz4.1.0.0

UnixJunkie:
also makes lz4_chans rely on it

### DIFF
--- a/packages/conf-lz4/conf-lz4.1.0.0/opam
+++ b/packages/conf-lz4/conf-lz4.1.0.0/opam
@@ -1,0 +1,17 @@
+opam-version: "2.0"
+homepage: "http://lz4.org"
+authors: ["Yann Collet"]
+license: "GPLv2 & BSD-2-clause"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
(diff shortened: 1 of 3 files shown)


packages/conf-lz4/conf-lz4.1.0.0/opam
```diff
@@ -0,0 +1,17 @@
+opam-version: "2.0"
+homepage: "http://lz4.org"
+authors: ["Yann Collet"]
+license: "GPLv2 & BSD-2-clause"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+maintainer: "https://github.com/ocaml/opam-repository/issues"
+depexts: [
```
kit-ty-kate: I would suggest to add a way of checking if the command is there. For example:
```suggestion
build: ["lz4" "-V"]
depexts: [
```

kit-ty-kate:
could you also use it for previous versions of lz4_chans?

UnixJunkie:
ok, updated.

kit-ty-kate:
Thanks a lot!!

PR #14218, opened 2019-05-31
LablTk 8.06.6 and OCamlBrowser 4.08.0

garrigue:
Packages for LablTk 8.06.6 and OCamlBrowser 4.08.0.
They are compatible with ocaml 4.08, and are built from the same archive.

### DIFF
--- a/packages/labltk/labltk.8.06.6/opam
+++ b/packages/labltk/labltk.8.06.6/opam
@@ -29,5 +29,5 @@ description: "ocamlbrowser is now a separate package.\n\
 flags: light-uninstall
 url {
   src: "https://github.com/garrigue/labltk/archive/8.06.6.tar.gz"
-  checksum: "md5=b52555cd25141e9b78868a9493d409f2"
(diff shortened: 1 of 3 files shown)


packages/labltk/labltk.8.06.6/opam
```diff
@@ -0,0 +1,33 @@
...
+  ["./configure" "-use-findlib" "-installbindir" bin]
+  [make "library" "opt"]
+]
+install: [
+  [make "install"]
+]
+remove: [["ocamlfind" "remove" "labltk"]]
+depends: [
+  "ocaml" {>= "4.07"}
+  "ocamlfind"
```
kit-ty-kate: both of these packages don't seem to use ocamlfind as a library or use it at runtime. Could you add the `{build}` attribute to them?

garrigue: Sure. Done.

kit-ty-kate:
Thanks!

PR #14219, opened 2019-05-31
[new release] merlin (3.3.0)

trefis:

##### CHANGES:

Fri May 31 11:09:08 BST 2019

  + backend
    - interpret `-pp` flag
    - backtrack warnings in all versions, not just 4.06
    - silence C compiler warnings (by David Allsopp and Bernhard Schommer)
    - remove sturgeon support
    - allow to select sections to log
    - better error message on ocaml version mismatch
    - locate:
      + handle functors and functor applications
      + do not use the location coming from the environment
    - tweaked caching policy
    - fix environment when a file disappears
    - fix -short-paths handling of classes and class types (by Leo White)
    - don't select deprecated paths in -short-paths (by Leo White)
    - return type info in outline query (by Andrey Popp)
    - properly handle new lines in the lexer
    - better tracking of errors reported by the parser and by preprocessors
    - add support for OCaml 4.08
    - tweaked the recovery strategy in presence of syntax errors
    - timing information in replies now includes wall clock time.
    - dump command can new dump the parsetree post preprocessing

  + editors modes
    - emacs
      + fix merlin-xref.el install (by Emilio Jesus Gallego Arias)
      + keep labels matching the prefix the user has typed rather than
        dropping them (by Mitchell Plamann)
      + remove unused `merlin--overlay` function (by Wilfred Hughes)
      + show the number of errors in the modline (by Wilfred Hughes)
      + call a logger on the client side if one is defined
      + allow user to disable completion inside comments and strings
      + show errors and types even when buffer is narrowed (by Wilfred Hughes)
      + make sure PATH is updated when merlin-command is 'opam

    - vim
      + better FindBinary
      + make the log buffer a scratch buffer (by Tom Johnson)
      + execute buffer switching silently (by Fabian)
      + restore view after updating merlin type buffer (by Fabian)

  + testsuite
    - Switched to mdx with cram syntax.

Special thanks to Rudi Grinberg for helping us in reviewing and merging
pull-requests.

### DIFF
--- a/packages/merlin/merlin.3.3.0/opam
+++ b/packages/merlin/merlin.3.3.0/opam
@@ -1,0 +1,71 @@
+opam-version: "2.0"
+name:         "merlin"
+maintainer:   "defree@gmail.com"
+authors:      "The Merlin team"
+homepage:     "https://github.com/ocaml/merlin"
+bug-reports:  "https://github.com/ocaml/merlin/issues"
+dev-repo: "git+https://github.com/ocaml/merlin.git"
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+]
+depends: [
+  "ocaml" {>= "4.02.1" & < "4.09"}
+  "dune" {build & >= "1.8.0"}
+  "ocamlfind" {>= "1.5.2"}
+  "yojson"
+  "mdx" {with-test & >= "1.3.0"}
+]
+synopsis:
+  "Editor helper, provides completion, typing and source browsing in Vim and Emacs"
+description:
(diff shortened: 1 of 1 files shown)


trefis:
The revdeps failure are just jupyter, which appears to be failing because of omp (cf. [log](https://ci.ocaml.org/log/saved/docker-run-940c6f92f7b88f981e1ecc0c948668dc/f7413be28ff0aeaa1f727e979b38322bd0b654e2)).

The other failures are due to merlin's tests failing to run because they depend on `jq` which is not available. Should I just disable the tests?

kit-ty-kate:
> The other failures are due to merlin's tests failing to run because they depend on `jq` which is not available. Should I just disable the tests?

I think that'd be appropriate

trefis:
All fixed.
The only tests that still fail are the unrelated revdeps failures.

kit-ty-kate:
Thanks!

PR #14225, opened 2019-06-01
Package qcstm.0.1

jmid:
### `qcstm.0.1`
A simple state-machine framework for OCaml based on QCheck

---
* Homepage: https://github.com/jmid/qcstm
* Source repo: git+https://github.com/jmid/qcstm.git
* Bug tracker: https://github.com/jmid/qcstm/issues

---

### DIFF
--- a/packages/qcstm/qcstm.0.1/opam
+++ b/packages/qcstm/qcstm.0.1/opam
@@ -12,6 +12,10 @@ bug-reports: "https://github.com/jmid/qcstm/issues"
 authors: [ "Jan Midtgaard" ]
 maintainer: [ "Jan Midtgaard <mail@janmidtgaard.dk>" ]
 dev-repo: "git+https://github.com/jmid/qcstm.git"
+url {
+  src: "https://github.com/jmid/qcstm/archive/0.1.tar.gz"
+  checksum: "md5=e802a61964397ba73277d0da583a5269"
+}
 depends: [
   "ocaml" {>= "4.04.0"}
   "qcheck" {>= "0.6"}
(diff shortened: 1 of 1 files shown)


mseri:
The url field, pointing at the sources to build, is missing

jmid:
> The url field, pointing at the sources to build, is missing

Thanks, that certainly helped!
A meta-comment: I was simply trying to mimic the example given at the top of https://opam.ocaml.org/doc/Packaging.html 
I think adding a `url`-field to this first opam-file example could be beneficial for a lot newcomers like me (I've noticed I'm not the only one making this mistake). I've thus created an issue to this effect.
Thanks again 🙂

kit-ty-kate:
ocamlfind and ocamlbuild don't seem to be used at runtime so I tagged them as build dependencies.
Could you return this fix upstream?

Seems good otherwise.

jmid:
> ocamlfind and ocamlbuild don't seem to be used at runtime so I tagged them as build dependencies.

OK to the build dependencies. If users want to play with this, e.g., in utop, they'll need to load the module. utop uses ocamlfind underneath as far as I understand, but then users will probably already have ocamlfind installed anyway.

> Could you return this fix upstream?

Sure, just to clarify as I'm not much of a git user: Do you want me to merge this in my opam-repository branch, make the change in the package's git repo qcstm.opam file (so it is addressed for future releases), or the former + retag as 0.1 release and recalculate md5-sum?

kit-ty-kate:
> make the change in the package's git repo qcstm.opam file (so it is addressed for future releases)

just that if that's ok. Your opam-repository branch is already updated and there is no need for retagging as the opam file inside the archive doesn't matter.

kit-ty-kate:
Thanks!

jmid:
Thanks for merging - yes, I'll make the change in repo's qcstm.opam file shortly 🙂

PR #14230, opened 2019-06-02
Package owee.0.3

let-def:
### `owee.0.3`
OCaml library to work with DWARF format
Owee is an experimental library to work with DWARF format.
It can parse ELF binaries and interpret DWARF debugline programs.

It can also be used to find locations of functions from the current process.

---
* Homepage: https://github.com/let-def/owee
* Source repo: git+https://github.com/let-def/owee.git
* Bug tracker: https://github.com/let-def/owee

---


let-def:
This release has been ported to dune and is compatible with OCaml 4.08 thanks to @lpw25.

kit-ty-kate:
Thanks!

PR #14235, opened 2019-06-03
Package semver2.1.0.2

etaque:
### `semver2.1.0.2`
Semantic version handling for OCaml

---
* Homepage: https://github.com/dividat/ocaml-semver
* Bug tracker: https://github.com/dividat/ocaml-semver/issues

---

### DIFF
--- a/packages/semver2/semver2.1.0.2/opam
+++ b/packages/semver2/semver2.1.0.2/opam
@@ -9,7 +9,7 @@ depends: [
   "ocaml" {>= "4.04.0"}
   "dune" {build & >= "1.2.0"}
   "angstrom" {>= "0.11.2"}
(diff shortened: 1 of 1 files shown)


mseri:
The failure in tests is because the library name for the `ounit` package is `oUnit`

etaque:
> The failure in tests is because the library name for the `ounit` package is `oUnit`

Oh... and it isn't failing on my macos machine because its filesystem isn't case sensitive I guess?

mseri:
Oh, that's probable: we had a similar issue not long ago (and also some name clashes due to case insensitivity), I think there is a plan to solve this at dune level at a certain point, but for now we don't have a good way of detecting it early :(

PR #14237, opened 2019-06-03
ptset: add bound on OCaml version

damiendoligez:
`ptset` fails to compile on all ocaml versions since 4.05.0 because some functions were added to the module type `Set.S`.

cc @backtracking


kit-ty-kate:
Thanks!

PR #14245, opened 2019-06-04
Add capnp-rpc-mirage upper-bounds on base64 and tcpip

talex5:
These constraints will be needed once https://github.com/ocaml/opam-repository/pull/14240 is merged.

/cc @avsm

### DIFF
--- a/packages/capnp-rpc-mirage/capnp-rpc-mirage.0.3.1/opam
+++ b/packages/capnp-rpc-mirage/capnp-rpc-mirage.0.3.1/opam
@@ -17,11 +17,12 @@ depends: [
   "fmt"
   "logs"
   "ipaddr" {<"3.0.0"}
+  "base64" {<"3.0.0"}
   "mirage-dns"
   "mirage-stack-lwt"
   "alcotest-lwt" {with-test}
   "io-page-unix" {with-test}
-  "tcpip" {with-test}
+  "tcpip" {with-test & <"3.5.0"}
   "mirage-vnetif" {with-test}
   "jbuilder" {build & >= "1.0+beta10"}
 ]
--- a/packages/capnp-rpc-mirage/capnp-rpc-mirage.0.3.1/opam
+++ b/packages/capnp-rpc-mirage/capnp-rpc-mirage.0.3.1/opam
@@ -11,7 +11,7 @@ build: [
 ]
 depends: [
   "ocaml" {>= "4.03.0"}
-  "capnp" {>= "3.1.0"}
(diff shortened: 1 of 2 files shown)


avsm:
Thanks! shouldn't this be merged first as they appear to be independent of #14240?

talex5:
Yes, this PR should be merged first (sorry; my original comment was unclear).

talex5:
Hmm. `OCaml 4.03, OCaml 4.04, OCaml 4.05, OCaml 4.06 failed` with:

```
$ opam depext -uivyj 2 capnp-rpc-mirage.0.3.1
        Try to specify a simpler request, use a different solver, or increase the allowed time by setting OPAMSOLVERTIMEOUT to a bigger value (currently, it is set to 60.0 seconds).
```

I'm going to try fixing this as part of the original PR, as that version has fewer dependencies.

avsm:
That's a bit of a known issue and should be fixed by the z3 backend...

talex5:
In that case it might be best to merge #14240 first, then rebuild this against the new master branch. Hopefully that will succeed, as the new capnp has fewer dependencies to confuse the solver.

avsm:
done, this can be rebased now

talex5:
I had to add a constraint `"capnp" {>= "3.3.0"}` to get past the solver timeout, even though the packages actually work fine with older versions. `Alpine, CentOS failed`, as expected, due to not having capnp packages.

talex5:
Travis still timed-out, but only *after* successfully picking the versions!

```
The following actions will be simulated:
  ∗ install conf-gmp                1          [required by zarith]
  ∗ install mirage-no-xen           1          [required by nocrypto]
  ∗ install conf-m4                 1          [required by ocamlfind]
...
Installing capnp-rpc-lwt.0.3.1.
∗ installed capnp-rpc-lwt.0.3.1
Installing capnp-rpc-mirage.0.3.
∗ installed capnp-rpc-mirage.0.3
Done.
...
        Try to specify a simpler request, use a different solver, or increase the allowed time by setting OPAMSOLVERTIMEOUT to a bigger value (currently, it is set to 60.0 seconds).
Command failed: opam list --readonly --with-test --external  --resolve=capnp-rpc-mirage.0.3 returned 60
```

talex5:
I think this PR should still be merged, because it's an improvement over the current (broken) state of the repository and the timeouts only happen (I think) when trying to install the test dependencies too, which normal users won't be doing, and only happen sometimes even then.

kit-ty-kate:
Thanks a lot!

PR #14246, opened 2019-06-05
remove version field

LucaFranceschini:
`version` field is outdated (causing warnings in `opam update`) and, if I correctly understand opam, not used anymore anyway.


kit-ty-kate:
Fixed by https://github.com/ocaml/opam-repository/pull/14253. Thanks!

PR #14247, opened 2019-06-05
elpi 1.3

gares:


gares:
Failure are in the test target that uses `/usr/bin/time` which is not available.
What is the preferred way to fix this? I guess I can override which `time` utility is used by passing arguments to the makefile.

gares:
All good now

kit-ty-kate:
It's of course fine without tests but I'm guessing it would've worked if the package depended on `"conf-time" {with-test}`

Thank you!

PR #14249, opened 2019-06-05
mlgmpidl.1.2.10

nberth:


kit-ty-kate:
Thanks! (note that the remove section is not necessary anymore since opam 2.0)

PR #14251, opened 2019-06-05
add upper bounds on mirage-xen packages that use OS.Xen

avsm:
preparing for https://github.com/mirage/mirage-xen/pull/16 in mirage-xen.4.0.0

cc @TheLortex


kit-ty-kate:
Thanks!

PR #14253, opened 2019-06-06
fix version field for core.v0.12.2

damiendoligez:
Fix a copy-paste mistake.

cc @janestreet

### DIFF
--- a/packages/core/core.v0.12.2/opam
+++ b/packages/core/core.v0.12.2/opam
@@ -1,5 +1,5 @@
 opam-version: "2.0"
-version: "v0.12.1"
+version: "v0.12.2"
 maintainer: "opensource@janestreet.com"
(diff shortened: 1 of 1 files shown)


mseri:
Ideally we should really remove that field

packages/core/core.v0.12.2/opam
```diff
@@ -1,5 +1,5 @@
 opam-version: "2.0"
-version: "v0.12.1"
+version: "v0.12.2"
```
mseri:
```suggestion
```

kit-ty-kate: (sorry I wasn't able to commit your suggestion for some unknown reason) but thanks!!

kit-ty-kate:
Thanks!

PR #14257, opened 2019-06-08
ocaml compiler distribution: use parallel build for OCaml 4.07 and above

gasche:
TL;DR: This is a follow-up on #14118 that enables parallel builds on 4.07 and more recent, except for Cygwin.

In the old days, the OCaml compiler distribution did not support
parallel build. This support has been contributed over time, and while
the dependencies are still not 100% reliable (in particular, parallel
rebuild from a partial build often fails with cmi inconsistencies),
builds from a clean state are now very robust: they are routinely used
by developers and tested daily on our CI machines, and we haven't seen
a failure in a long while.

Enabling parallel build improves build times for everyone, at the cost
of a very low risk of breaking the build. Some context:

- On my machine, a recent sequential build took 5m15s, and a parallel
  build from the same sources took 2m53s: we can save 2m20s for every
  new switch creation for users with similar hardware. When creating
  a switch with a small number of packages, the compiler build time
  may dominate the total installation time.

- New switches are created much more often now that opam2 makes local
  switches easy. It is typical for users to create a switch for each
  project to isolate their dependencies, ending up with dozens of
  switches. The install time quickly builds up.

If a user was to find a build failure due to parallel build,
a workaround is to pass `-j 1` to opam to force sequential switch
creation:

    opam switch create <switch-name> <compiler-version> -j 1

setting the environment variable OPAMJOBS=1 would also work.

This commit was semi-automatically generated
using the following script run from the `packages/` directory:

```
for f in ocaml-base-compiler/*4.07.*/opam ocaml-variants/*4.0[789].*/opam ocaml-variants/*4.10.*/opam; do
    echo $f
    sed 's/  \[make "world.opt"\]/  [make "-j%{jobs}%" "world.opt"] {os != "cygwin"}/' -i $f
    sed 's/  \[make "world"\]/  [make "world.opt"] {os = "cygwin"}/' -i $f
done
```

### DIFF
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.4.07.0/opam
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.4.07.0/opam
@@ -23,11 +23,20 @@ build: [
     "-aspp"
     "cc -c"
   ] {os = "openbsd" | os = "freebsd" | os = "macos"}
-  [make "world"]
-  [make "world.opt"]
+  [make "-j%{jobs}%" {os != "cygwin"} "world"]
+  [make "-j%{jobs}%" {os != "cygwin"} "world.opt"]
 ]
 install: [make "install"]
 url {
   src: "https://github.com/ocaml/ocaml/archive/4.07.0.tar.gz"
   checksum: "md5=1f78bb35a2f15d5ec737ee6a8dc6890d"
 }
+post-messages: [
+  "A failure in the middle of the build may be caused by build parallelism
+   (enabled by default).
+   Please file a bug report at https://github.com/ocaml/ocaml/issues"
+  {failure & jobs > 1 & os != "cygwin"}
+  "You can try installing again including --jobs=1
+   to force a sequential build instead."
+  {failure & jobs > 1 & os != "cygwin" & opam-version >= "2.0.5"}
+]
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.4.07.1/opam
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.4.07.1/opam
@@ -23,11 +23,20 @@ build: [
     "-aspp"
     "cc -c"
   ] {os = "openbsd" | os = "freebsd" | os = "macos"}
-  [make "world"]
-  [make "world.opt"]
+  [make "-j%{jobs}%" {os != "cygwin"} "world"]
+  [make "-j%{jobs}%" {os != "cygwin"} "world.opt"]
 ]
 install: [make "install"]
 url {
   src: "https://github.com/ocaml/ocaml/archive/4.07.1.tar.gz"
   checksum: "md5=352fe8d46cb238a26aa10c38bad6ecb6"
 }
+post-messages: [
+  "A failure in the middle of the build may be caused by build parallelism
+   (enabled by default).
+   Please file a bug report at https://github.com/ocaml/ocaml/issues"
+  {failure & jobs > 1 & os != "cygwin"}
+  "You can try installing again including --jobs=1
+   to force a sequential build instead."
+  {failure & jobs > 1 & os != "cygwin" & opam-version >= "2.0.5"}
+]
--- a/packages/ocaml-variants/ocaml-variants.4.07.0+32bit/opam
+++ b/packages/ocaml-variants/ocaml-variants.4.07.0+32bit/opam
@@ -42,11 +42,20 @@ build: [
     "-host"
     "i386-apple-darwin13.2.0"
   ] {os = "macos"}
-  [make "world"]
-  [make "world.opt"]
+  [make "-j%{jobs}%" {os != "cygwin"} "world"]
+  [make "-j%{jobs}%" {os != "cygwin"} "world.opt"]
 ]
 install: [make "install"]
 url {
   src: "https://github.com/ocaml/ocaml/archive/4.07.0.tar.gz"
   checksum: "md5=1f78bb35a2f15d5ec737ee6a8dc6890d"
 }
+post-messages: [
+  "A failure in the middle of the build may be caused by build parallelism
+   (enabled by default).
+   Please file a bug report at https://github.com/ocaml/ocaml/issues"
+  {failure & jobs > 1 & os != "cygwin"}
+  "You can try installing again including --jobs=1
+   to force a sequential build instead."
+  {failure & jobs > 1 & os != "cygwin" & opam-version >= "2.0.5"}
+]
--- a/packages/ocaml-variants/ocaml-variants.4.07.0+afl/opam
+++ b/packages/ocaml-variants/ocaml-variants.4.07.0+afl/opam
@@ -23,11 +23,20 @@ build: [
     "-aspp"
     "cc -c"
   ] {os = "openbsd" | os = "freebsd" | os = "macos"}
-  [make "world"]
-  [make "world.opt"]
+  [make "-j%{jobs}%" {os != "cygwin"} "world"]
+  [make "-j%{jobs}%" {os != "cygwin"} "world.opt"]
 ]
 install: [make "install"]
 url {
   src: "https://github.com/ocaml/ocaml/archive/4.07.0.tar.gz"
   checksum: "md5=1f78bb35a2f15d5ec737ee6a8dc6890d"
 }
+post-messages: [
+  "A failure in the middle of the build may be caused by build parallelism
+   (enabled by default).
+   Please file a bug report at https://github.com/ocaml/ocaml/issues"
+  {failure & jobs > 1 & os != "cygwin"}
+  "You can try installing again including --jobs=1
+   to force a sequential build instead."
+  {failure & jobs > 1 & os != "cygwin" & opam-version >= "2.0.5"}
+]
--- a/packages/ocaml-variants/ocaml-variants.4.07.0+beta2+afl/opam
+++ b/packages/ocaml-variants/ocaml-variants.4.07.0+beta2+afl/opam
@@ -23,11 +23,20 @@ build: [
     "-aspp"
     "cc -c"
   ] {os = "openbsd" | os = "freebsd" | os = "macos"}
-  [make "world"]
-  [make "world.opt"]
+  [make "-j%{jobs}%" {os != "cygwin"} "world"]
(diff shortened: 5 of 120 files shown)


pmetzger:
As I've said before, I'm _very_ much in favor.

packages/ocaml-base-compiler/ocaml-base-compiler.4.07.0/opam
```diff
@@ -23,8 +23,8 @@ build: [
     "-aspp"
     "cc -c"
   ] {os = "openbsd" | os = "freebsd" | os = "macos"}
-  [make "world"]
-  [make "world.opt"]
+  [make "world.opt"] {os = "cygwin"}
+  [make "-j%{jobs}%" "world.opt"] {os != "cygwin"}
```
dra27: Sorry for being late to this - you can use filters on the terms in a command as well, so the line above could be deleted and this changed to:
```suggestion
  [make "-j%{jobs}%" {os != "cygwin"} "world.opt"] 
```

I'm not particularly sure which more clearly conveys that parallelism is disabled on Cygwin!

dra27:
Is it definitely a good idea to enable this for the `+trunk` variants? On the one hand, trunk is broken for parallel build at various times, on the other hand only advanced users should tend to be trying it, on the third hand (!!), users do occasionally use master/trunk versions of things even when it's not necessarily the sensible thing to be doing...

packages/ocaml-base-compiler/ocaml-base-compiler.4.07.0/opam
kit-ty-kate: I think your proposal is clearer and more concise.

gasche:
No strong opinion on `+trunk` variants, but I note that the person who introduced pin-friendly `ocaml-variants.opam` files in the compiler distribution (I think Leo) did use the parallel option (independently of this PR), so it's not worse to have it in the opam switch. We can always backtrack on `+trunk` if we get reports of failures. I could also add a compiler CI test to ensure that `make depend` is a no-op?

Thanks for the local-condition suggestion, I'll amend the PR.

dra27:
Pinning is quite explicit - I agree that it’s fine there. I was thinking of possible users create their switches based on the trunk branch (if they exist - but I see this in other projects, and marvel somewhat...)

gasche:
I rebased this PR and used the `"-j%{jobs}%" {os != "cygwin"}` style.

I kept the option for `+trunk` trees. Any failure there is fixable with OPAMJOBS=1, and I find it more natural to be consistent with in-tree pinning instruction. This makes it slightly harder to use `+trunk` switches, and makes maintainers slightly more motivated to keep trunk parallel-build-friendly, and I think both are reasonable trade-offs.

pmetzger:
How do we decide if we have consensus to merge? (I think it should be merged, FWIW.)

mseri:
+1 for merge

pmetzger:
And clearly @gasche wants to merge this or he wouldn't have proposed it, so that's three people. Any other votes?

pmetzger:
@kit-ty-kate ?

gasche:
Thanks for the show of support! But note that this is something that has to be done carefully (this is why we dragged our feet for a long time). In particular, I would like to be sure to hear another round of feedback from @dra27 (playing the role of the opponent here, which improved the PR quite a bit) before you consider merging.

dra27:
Thinking more about the `+trunk` packages, the way upgrading is (or rather isn’t) done at the moment means that a package using the trunk variant probably tends towards stability.

One last thought might be to put a build error message in the opam file to recommend retrying with `OPAMJOBS=1` on error?

gasche:
Let's agree on a message formulation before I update the PR. Would the following work:

```
post-messages: [
  "A failure in the middle of the build may be due to building in parallel (the default).
   You can try installing again with the OPAMJOBS=1 variable set to enforce a sequential build instead."
  {failures & os != "cygwin"}
]
```

Do we agree with the idea of adding this message to *all* parallel-build-enabled compiler descriptions, not just the +trunk ones?

pmetzger:
That message seems fine. (I would have said "may be caused by a parallel build" and "to force a sequential build" (not "enforce") but it's understandable either way.

gasche:
I changed "enforce" into "force", thanks. "Build parallelism" sounds more complex but the idea is that it's really the fact of being parallel that causes a failure, not "a parallel build".

pmetzger:
As I said, either phrasing for either of them would be understood. Personally I'm pretty happy; on the question of adding this on all parallel builds, that seems fine, too.

dra27:
Note that the predicate variable is `failure` not `failures`. What do you think of this for the actual trunk branch (recalling that after we release, ocaml-variants.4.09.0+trunk actually points to branch 4.09, not trunk):

```
https://github.com/ocaml/opam-repository/pull/14257

**Comment by dra27** [comment]
Note that it's `failure` not `failures` and I'd filter the message on `jobs` as well so it is not displayed for `OPAMJOBS=1` (but see note further on). I'd suggest that we want two slightly different messages. For the actual trunk builds (that means 4.10 variants at the moment - not 4.07, 4.08 and 4.09 which don't actually point to the trunk branch on ocaml/ocaml) perhaps this:

```
post-messages: [
  "A failure in the middle of the build may be caused by build parallelism
   (enabled by default). You can try installing again including --jobs=1
   to force a sequential build instead.
   See https://github.com/ocaml/opam-repository/pull/14257 for more info."
  {jobs > 1 & failure & os != "cygwin"}
]
```

and for released branches, the message should be more severe for reporting a bug:

```
post-messages: [
  "A failure in the middle of the build may be caused by build parallelism
   (enabled by default). You can try installing again including --jobs=1
   to force a sequential build instead.
   Please file a bug report at https://github.com/ocaml/ocaml/issues"
  {jobs > 1 & failure & os != "cygwin"}
]

```

There is a problem, however, which is that the `jobs` variable does not at present mean the value passed to `opam install`, it only refers to the configured value at `opam init` time (you can see this with `OPAMJOBS=1 opam config var jobs`). This is a problem, since it means that the workaround doesn't work.

**Comment by dra27** [comment]
Just an update on this. The filters on the `-j%{jobs}%` want to become `{opam-version >= "2.0.5" & os != "cygwin"}`. The idea is that parallel build is only attempted on a version of opam with either ocaml/opam#3916 or ocaml/opam#3881 where specifying `--jobs=1` will work to disable the parallelism.

This can obviously (and should, given the nuisance that rebasing it is) be merged prior to opam 2.0.5 being released, although that's due in the next few weeks.

**Comment by gasche** [comment]
Actually I never "rebase" this PR in the traditional sense, I update the script that is included at the end of the commit message and then I run it from a fresh branch off the current `master`. I don't mind waiting for 2.0.5, and now that you caught me red-handed with an untested error-handling path I would be in favor of waiting to test the fix.

**Comment by avsm** [comment]
Hang on, I'm really not ok with this message formulation that the build 'might work'.  opam isn't a package manager where builds work some of the time -- they should either be work every time for a given build environment and dependency graph, or fixed until they do work every time.

In the case of OCaml itself, we compile it thousands of times a week in CI, and so we will find parallel build problems fairly quickly, and thus can either fix it (by reverting this parallel build option) or patching the offending rule.

Exposing this workaround to users in the message for such a core package seems like extremely poor form to me.  Do we or do we not believe that parallel builds work? If we're not sure, then we should leave a multicore machine running in a build loop of the compiler with different -j options for a few days and check it works reliably.  Has anyone done that yet? If not, and noone else can do it, I'll do it when back -- but I'm travelling this week.

**Comment by dra27** [comment]
@avsm - the most severe message is for *trunk* (i.e. developer builds) where such a guarantee can never be made (which was my original suggestion not to have it at all for trunk builds). This message is also only displayed if it goes wrong, being released in the belief that it should work all the time. Your bar for this is too high, given OCaml’s present build system - the overwhelming number of CI builds of OCaml are not run in parallel and, even on CI systems where they can be, they’re likely to be run with low parallelism.

**Comment by dra27** [comment]
(I’m wondering if you misread it - the message is not formulated that it _might work_ but that it _may have failed because_?)

**Comment by gasche** [comment]
opam 2.0.5 [was just released](https://opam.ocaml.org/blog/opam-2-0-5/) with the fix that makes OPAMJOBS=1 work as expected in this circumstance. I would like to converge on the messages to update the PR once again with something that people agree on.

@avsm: with the wording proposed by @dra27 above, the message shown to non-trunk-build users is as follows:

```
post-messages: [
  "A failure in the middle of the build may be caused by build parallelism
   (enabled by default). You can try installing again including --jobs=1
   to force a sequential build instead.
   Please file a bug report at https://github.com/ocaml/ocaml/issues"
  {jobs > 1 & failure & os != "cygwin"}
]
```

Does that work with you? Would you like to suggest an alternative formulation, or do you strongly think that it is better not to have a message at all?

dra27:
(@gasche - I realise looking again that the message should also have the `opam-version >= "2.0.5"`, so it will be `{failure & jobs > 1 & opam-version >= "2.0.5" & os != "cygwin"}`)

avsm:
@gasche @dra27: I discussed this with @dra27 and am now convinced that this is the right approach :-) Thanks for the multiple iterations and explanations! Happy for this to be merged once rebased

gasche:
I rebased the PR with the new version. I used the following variant on the proposed message, where only the workaround suggestion is conditioned on the opam version:

    post-messages: [
      "A failure in the middle of the build may be caused by build parallelism
       (enabled by default).
       Please file a bug report at https://github.com/ocaml/ocaml/issues"
      {failure & jobs > 1 & os != "cygwin"}
      "You can try installing again including --jobs=1
       to force a sequential build instead."
      {failure & jobs > 1 & os != "cygwin" & opam-version >= "2.0.5"}
    ]

kit-ty-kate:
Do everyone agree on this PR?

pmetzger:
I think there is general consensus.

kit-ty-kate:
Thanks a lot!

pmetzger:
This being committed makes me happy!

dra27:
I think because this PR was so long-lived, `ocaml-base-compiler.4.08.0` doesn't have the change - @gasche, is it easy for you to run your script over it and any other 4.08.0 packages which got missed?

gasche:
Done, thanks!

PR #14261, opened 2019-06-08
New packages: conjury and orsetto.

jhwoodyatt:
- Conjury (version 2.0) is https://bitbucket.org/jhw/conjury
- Orsetto (version 1.0) is https://bitbucket.org/jhw/orsetto

### DIFF
--- a/packages/conjury/conjury.2.0/opam
+++ b/packages/conjury/conjury.2.0/opam
@@ -1,0 +1,23 @@
+opam-version: "2.0"
+name: "conjury" version: "2.0"
+synopsis: "Conjury library for OMake"
+maintainer: "james woodyatt <jhw@conjury.org>"
+authors: "james woodyatt <jhw@conjury.org>"
+homepage: "https://bitbucket.org/jhw/conjury/"
+bug-reports: "https://bitbucket.org/jhw/conjury/issues"
+dev-repo: "hg+https://bitbucket.org/jhw/conjury"
+tags: [ "org:conjury.org" ]
+depends: [
+    "omake" { >= "0.10.3" }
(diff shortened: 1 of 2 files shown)


jhwoodyatt:
Always happens. Noticed a last minute goblin and pushed a quick update to fix it.

packages/conjury/conjury.2.0/opam
```diff
@@ -0,0 +1,23 @@
...
+name: "conjury" version: "2.0"
+synopsis: "Conjury library for OMake"
+maintainer: "james woodyatt <jhw@conjury.org>"
+authors: "james woodyatt <jhw@conjury.org>"
+homepage: "https://bitbucket.org/jhw/conjury/"
+bug-reports: "https://bitbucket.org/jhw/conjury/issues"
+dev-repo: "hg+https://bitbucket.org/jhw/conjury"
+tags: [ "org:conjury.org" ]
+depends: [
+    "ocaml" { >= "4.04" }
```
kit-ty-kate: Does this package really requires OCaml (outside of tests)

jhwoodyatt: I think so. It includes OMake rules for invoking the OCaml compiler and related tools. And those rules don't work for OCaml prior to 4.04 (because the `-bytecode` option). It has a transitive dependency on `ocaml` through its dependency on the `omake` package, which requires 4.03 or more recent.

The `omake` package really should have its dependency on `ocaml` downgraded to a build-only dependency. But that's another thing.

jhwoodyatt: Tell you what, though... I'll downgrade its dependency to `build` and `test`.

mseri:
Looks good. Thanks!

Issue #14185, opened 2019-05-27
remove all `{build}` directives from dune packages in opam-repo


avsm:
See https://github.com/ocaml/opam/issues/3850 for background

The `dune-package` files installed by dune lead to runtime failures if it is subsequently downgraded and the installed packages are not recompiled.  Therefore we need to replace `dune {build}` with just `dune` in opam-repository.

The change to opam-repository is quite big though:
```
$ grep -r '{build' packages/|grep dune|wc -l
    1135
```

@AltGr @kit-ty-kate  @rjbou is this something that could be automated via an enhancement to `opam admin add-constraints` perhaps?

mseri:
Oh god, but that would mean that at each dune update one has to rebuild the whole world :(

kit-ty-kate:
I'm not sure to understand what this issue has to do with the `build` attribute. A lot of dune packages forget they require a specific version of dune (1.8 or something) and they don't have this constraint in the opam file.

avsm:
Can you show a concrete example of a package that you are referring to @kit-ty-kate ?

kit-ty-kate:
oh... ok, I thought the content of the `dune-package` installed had the version number specified in the local `dune-package` but it doesn't seem to be the case. It feels to me more like a dune bug than something to do in opam (though it's already too late I suppose :/)

avsm:
The dune-package installed does have the version number specified -- it's just that opam isn't enforcing the version constraint due to `{build}`.  This is my point about why `{build}` is so dangerous -- as software evolves, older uses of `{build}` can just become wrong silently.

kit-ty-kate:
> The dune-package installed does have the version number specified

I've tried earlier with `containers.2.5` and it's not what I got.
In the source code, the content of `dune-package` gives:
```
(lang dune 1.0)
```
but once installed the content of the `dune-package` is:
```
(lang dune 1.9)
[...]
```
in my case

avsm:
Right, because the installed dune-package uses metadata from version 1.9 of dune.   Dune uses dune-package to store structured metadata about the installation and adds to it on every release, so it's backwards but not forwards compatible.

AltGr:
>  it's just that opam isn't enforcing the version constraint due to {build}.

[...]

PR #14266, opened 2019-06-10
Mark dune as a non-build dependency

kit-ty-kate:
One possible fix for https://github.com/ocaml/opam-repository/issues/14185

Accepting the merge this is going to be painful for a lot of users. I'm not really looking forward to this but I feel like we have only a few choices:
* Either merge this PR and accept we'll annoy and loose users every time dune gets updated.
* Change the behaviour of `{build}` in opam itself to make it only rebuild on downgrades
* Fix the problem in `dune` and disable the affected versions (@diml would you know which versions are affected?)

I'm basically rewording @AltGr's comment here (see https://github.com/ocaml/opam/issues/3850#issuecomment-496248618). See there and in the issue above for further context and discussions.

cc @avsm 

*Note: I'm going to cancel manually all the CI run for this PR so that it doesn't take all the CI time available*


kit-ty-kate:
I'm not too sure what's wrong with the CI. Maybe the patch is too big?!
```
<><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>
[default] synchronised from git+file:///home/opam/opam-repository
[ERROR] Could not update repository "default": "/usr/bin/patch -p1 -i /home/opam/.opam/log/processed-patch-18-05296d" exited with code 1
```

kit-ty-kate:
@ocaml/opam-repository we should now be vigilant to not merge PRs with `dune` flagged as build dependencies

PR #14267, opened 2019-06-11
Mitigate dune downgrades

kit-ty-kate:
Mitigate https://github.com/ocaml/opam-repository/issues/14185 by avoiding downgrades as much as possible. This should work for all opam packages in opam-repository but there might still be some issues for external packages or remotes.

This PR disables those packages from the repository and cannot be installed anymore:
* decoders.0.1.0
* decoders.0.1.1
* lwt.3.1.0
* lwt.3.2.0
* lwt.3.2.1
* lwt.3.3.0
* ocp-index.1.1.6
* ocp-index.1.1.7
* yojson.1.4.0
* yojson.1.4.1

None of these packages are their latest versions but this might also make some more packages uninstallable. This will be detected by Camelus.

cc @avsm @diml @AltGr

### DIFF
--- a/packages/decoders/decoders.0.1.0/opam
+++ b/packages/decoders/decoders.0.1.0/opam
@@ -171,6 +171,7 @@ depends: [
 ]
 build: ["jbuilder" "build" "-p" name "-j" jobs]
 conflicts: [ "dune" {>="1.7.0"} ]
+available: false
 dev-repo: "git+ssh://git@github.com/mattjbray/ocaml-decoders.git"
 url {
   src:
--- a/packages/decoders/decoders.0.1.1/opam
+++ b/packages/decoders/decoders.0.1.1/opam
@@ -260,6 +260,7 @@ depends: [
   "containers"
 ]
 conflicts: [ "dune" {>="1.7.0"} ]
+available: false
 build: ["jbuilder" "build" "-p" name "-j" jobs]
 dev-repo: "git+ssh://git@github.com/mattjbray/ocaml-decoders.git"
 url {
--- a/packages/lwt/lwt.3.1.0/opam
+++ b/packages/lwt/lwt.3.1.0/opam
@@ -66,6 +66,7 @@ conflicts: [
   "ocaml-variants" {= "4.02.1+BER"}
   "dune" {>= "1.7.0"}
 ]
+available: false
 url {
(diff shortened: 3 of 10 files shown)


mseri:
I understand your reasons, but I am worried by this solution. For example `lwt.3` is not that old, and it was the release before the big breakage: there are many packages out (some of which not here but using opam) that work with it but not with `lwt 4`.

diml:
To give a bit more context, this PR follows a discussion we had this week with @kit-ty-kate, @AltGr and @avsm. I raised some concerns about #14266 as I was worried about the impact for opam and dune users.

We then had a discussion between dune developers yesterday to see if we could avoid this problem entirely by making dune forward compatible. It turns out that while backward compatibility is good and already supported by dune, forward compatibility introduces other issues for both dune and users of dune. Ideally, you shouldn't have to rebuild everything when you upgrade dune, but when you downgrade dune it's cleaner and safer to rebuild everything. We can't express that fact in opam files at the moment, so I suppose the safest approach is @kit-ty-kate's original PR to remove the `{build}` flag on dune dependencies.

mseri:
Thanks for the context. I had only partially followed the original discussion, and indeed raised some concerns also for the removal of the build flags.

But as mucg as I agree that the removal of the `{build}` flag is disruptive, and that we should probably make opam deal with rebuilding on downgrades, I would favour that instead of disabling important packages.

avsm:
I’m not in favour of this PR. My view is that we should get opam repository correct and coinstallable and not worry about a few rebuilds at this stage.

Once we have a version of dune that is forwards compatible we can hopefully apply a safer optimisation in opam. But for now, correctness should come first to avoid breakage on upgrade.

diml:
@avsm dune will not become forward compatible. That's what we discussed yesterday: making dune forward compatible would solve the downgrade problem but would create other more complicated ones.

In the end, it's better to focus our efforts on the shared artifact cache, it will make everything a lot faster and we will make the `{build}` flag irrelevant for projects using dune.

avsm:
That’s completely fine by me.  Indeed, binary caching will eliminate a lot of rebuild time, so that’s a good way forward.  I’m really keen to see any build unsoundness disappear from opam before we focus on any speed ups at the opam metadata level. The solver is complex enough :-)

avsm:
@kit-ty-kate is preparing another PR that will remove `{build}`, so closing this one

PR #14270, opened 2019-06-11
[new release] mdx (1.4.0)

avsm:
Executable code blocks inside markdown files

- Project page: <a href="https://github.com/realworldocaml/mdx">https://github.com/realworldocaml/mdx</a>
- Documentation: <a href="https://realworldocaml.github.io/mdx/">https://realworldocaml.github.io/mdx/</a>

##### CHANGES:

- Add `--force-output` option to force generation of diff file (realworldocaml/mdx#118 @clecat)
- Support OCaml 4.08.0 (realworldocaml/mdx#121 @xclerc)
- README and documentation fixes (realworldocaml/mdx#122 realworldocaml/mdx#118 @andreypopp @clecat @samoht)
- Use latest ocaml-migrate-parsetree interfaces (@avsm)

PR #14271, opened 2019-06-11
msat.0.8: missing test depend on sequence

avsm:
fixes revdep in #14270

### DIFF
--- a/packages/msat/msat.0.8/opam
+++ b/packages/msat/msat.0.8/opam
@@ -13,6 +13,8 @@ depends: [
   "iter" {>= "1.2"}
   "containers" {with-test}
   "sequence" {with-test}
+  "menhir" {with-test}
+  "conf-time" {with-test}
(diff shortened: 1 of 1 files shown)


mseri:
It currently fails with
```
# File "tests/dune", line 28, characters 0-144:
# 28 | (alias
# 29 |   (name runtest)
# 30 |   (deps   ./../src/main/main.exe ./run (source_tree .))
# 31 |   (locks test)
# 32 |   (action (run /usr/bin/time -f "%e" ./run sat)))
# Error: File unavailable: /usr/bin/time
# Error: Program menhir not found in the tree or in PATH (context: default)
# Hint: try: opam install menhir
#     test_api alias tests/runtest
```

Let's see if my attempted fix is enough

kit-ty-kate:
Seems like tests are incompatible with MacOS's version of `time`. But it's good enough for now. Thanks to you two

PR #14273, opened 2019-06-11
asn1-combinators: constrain to cstruct<5

avsm:
see https://github.com/mirleft/ocaml-asn1-combinators/issues/28


mseri:
The failure does not depend on this.
Thanks

PR #14279, opened 2019-06-13
* packages/pfff/pfff.0.37.1/opam: new release

aryx:
The librairies in this package are needed for another project I want to submit soon to OPAM (efuns).

### DIFF
--- a/packages/pfff/pfff.0.37.1/opam
+++ b/packages/pfff/pfff.0.37.1/opam
@@ -22,6 +22,7 @@ depends: [
   "ocamlfind"
   "camlp4"
   "num"
+  "ocamlgraph"
 ]
 bug-reports: "https://github.com/aryx/pfff/issues"
 dev-repo: "git+https://github.com/aryx/pfff.git"
--- a/packages/pfff/pfff.0.37.1/opam
+++ b/packages/pfff/pfff.0.37.1/opam
@@ -13,13 +13,8 @@ build: [
 install: [
   [make "install-libs"]
 ]
-remove: [
-  ["./configure" "--novisual" "--nocmt" "--nobytecode"]
-  [make "uninstall-libs"]
-]
 depends: [
   "ocaml" {>= "4.02.3"}
-  "ocamlfind"
   "camlp4"
   "num"
   "ocamlgraph"
@@ -41,8 +36,6 @@ even TeX. There is also very good support for OCaml and noweb
 itself.
 
 For more information see https://github.com/facebook/pfff/wiki/Main"""
-extra-files: [
-]
 url {
   src: "https://github.com/aryx/pfff/archive/0.37.1.tar.gz"
   checksum: "md5=637a4f055de86bf02ad372b45e5567a9"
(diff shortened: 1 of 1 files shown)


aryx:
Is there a way to see the output of the failing command running for the Alpine distro. I can click on the logs but then the log just tells me there is an error in the configure command without any output I can look at.

packages/pfff/pfff.0.37.1/opam
```diff
@@ -0,0 +1,49 @@
...
+build: [
+  ["./configure" "--novisual" "--nocmt" "--nobytecode"]
+  [make "depend"]
+  [make]
+  [make "opt"]
+]
+install: [
+  [make "install-libs"]
+]
+remove: [
```
kit-ty-kate: This part is unnecessary. Opam 2 already takes care of removing files that have been installed. Could you remove it?

packages/pfff/pfff.0.37.1/opam
```diff
@@ -0,0 +1,49 @@
...
+support for Erlang, Python, C#, Lisp, Scheme, Haskell, OPA, Sql, and
+even TeX. There is also very good support for OCaml and noweb
+(literate programming) so that pfff can be used on the code of pfff
+itself.
+
+For more information see https://github.com/facebook/pfff/wiki/Main"""
+extra-files: [
+]
+url {
+  src: "https://github.com/aryx/pfff/archive/0.37.1.tar.gz"
```
kit-ty-kate: Could you clarify the status of this? I see facebook stopped working on this a few years ago. Is it the new official repository or just your fork?

packages/pfff/pfff.0.37.1/opam
```diff
@@ -0,0 +1,49 @@
...
+install: [
+  [make "install-libs"]
+]
+remove: [
+  ["./configure" "--novisual" "--nocmt" "--nobytecode"]
+  [make "uninstall-libs"]
+]
+depends: [
+  "ocaml" {>= "4.02.3"}
+  "ocamlfind"
```
kit-ty-kate:
```suggestion
```
Where is this used? I don't even see use of ocamlfind or Findlib anywhere in the source code

packages/pfff/pfff.0.37.1/opam
aryx: I am the main author of pfff and I left facebook in 2014. It is not maintained anymore by facebook
so I've forked it. I did not migrate yet the wiki to my fork, hence the mix of URLs.

packages/pfff/pfff.0.37.1/opam
aryx: There is some occurences of ocamlfind in some Makefile, but for opam related stuff it's not used indeed. I've removed the dependency. Thx.

packages/pfff/pfff.0.37.1/opam
aryx: sure

aryx:
So how can I see the actual error with "Alipine failed" above. Clicling on Details and then Logs just show that configure returns an error without any output.

mseri:
I don’t think there is a way from the web interface to get mor information than what is you have already seen

aryx:
Can someone still accept my patch? I mean only Alpine users seem to be affected.

kit-ty-kate:
Thanks!

PR #14287, opened 2019-06-14
Introduce a .gitattributes file

dra27:
The purpose of this PR is to ensure that all text files in the repository are LF-normalised. This is done by introducing a `.gitattributes` file which enables Git's text file detection and always *stores* text files with Unix line endings, regardless of local settings. The file contains global instructions to ensure that shell scripts are checked out with LF-endings, even on Windows machines.

Although opam 2 will automatically adapt patches which are checked out with the "wrong" line endings, it is, in my experience, better to treat patch and diff files as binary (i.e. no normalization) which is done here.

Some packages have scripts which don't end with `.sh` - these have been assigned specific `.gitattributes` files in their *package* directory (i.e. not with the `opam` file). This is so that the package can be renamed (potentially) without having to update `.gitattributes` and also so that new versions of a package automatically gain the attribute.

In two packages, I renamed patch files to have the `.patch` extension so that no specific attribute was required.

The last commit corrects 3 files which have been clearly been written on a Windows machine which didn't have `core.autocrlf` set - it changes the line ending stored in the repository from Windows to Unix. The commit can be verified as having no actual change by using `git diff --ignore-space-at-eol`.


kit-ty-kate:
Looks good. Thanks a lot!

PR #14288, opened 2019-06-14
Prune various invalid files

dra27:
- `findlib` files aren't used any more
- `desr` and `desc` were presumably typos for `descr` which also isn't used any more
- `packages/ojs-base/ojs-base.0.1.0/.chamo.sourceview.buffers` should never have been committed

Does/should Camelus check for these?


kit-ty-kate:
> Does/should Camelus check for these?

I don't think it does.

Thanks a lot!

PR #14290, opened 2019-06-14
FrontC 3.4.2 - adds trailing comma to enums

ivg:
This is a minor update, that adds the abovementioned feature.


kit-ty-kate:
The remove field is unnecessary with opam 2.0. Also, you don't seem to use `ocamlfind` outside of the said `remove` field so I removed it.

ivg:
Thanks guys!

PR #14294, opened 2019-06-15
Temporarily have only one .gitattributes file

dra27:
Having `.gitattributes` in project directories is causing pain to `esy` - cross reference to bug fix to follow (after which this PR can be reverted).


dra27:
Cross-referencing https://github.com/esy/esy/issues/936

PR #14295, opened 2019-06-15
Revert #14294 and restore package-specific .gitattributes

dra27:
**DO NOT MERGE YET!**

This PR can merged when https://github.com/esy/esy/issues/936 is resolved.

Also note https://github.com/ocaml/opam/pull/3873, though this change wouldn't affect `esy` since it doesn't use `index.tar.gz`

cc @jordwalke, @andreypopp

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -10,12 +10,6 @@ pkgconfigure text eol=lf
 # NB Package-specific attributes should go in packages/NAME/.gitattributes,
 # not here, in order to ensure that they survive packages being renamed.
 
-# They are temporarily here, working around a bug in esy
-packages/core/*/files/corebuild text eol=lf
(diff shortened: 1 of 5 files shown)


dra27:
This issue has been fixed in esy 0.5.8, so this PR just needs a thumbs up from the esy team that they're confident that enough users have migrated.

mseri:
@dra27 this is now conflicting with gitattributes

@jordwalke, @andreypopp can we merge this?

dra27:
The conflict was just from #14348 - rebased to avoid it.

kit-ty-kate:
I'm not sure this PR should stay open indefinitely as it is just a superficial fix and things seems fine as they are. Please reopen if needed.

PR #14297, opened 2019-06-16
MD5 update for the macos pytorch binary package

LaurentMazare:
The PyTorch binary package for macos has recently been updated and its md5 checksum changed, this updates the opam package to match the new checksum.


mseri:
Thanks!

PR #14301, opened 2019-06-17
Package: lambda-term.2.0.1

kandu:
A package for the minor release: https://github.com/ocaml-community/lambda-term/releases/tag/2.0.1


mseri:
The dependency change to zed 2 seems to have broken the revdeps for ocaml-9p (ping @djs55), but I think this must have been the case already for lambda-term 2.0.

kandu:
Ah, I've already restricted the upper bounder for protocol-9p-tool in PR #14055. It seems somebody loosened it by accident.

This PR will fix this issue for protocol-9p-tool: https://github.com/mirage/ocaml-9p/pull/141

mseri:
I'll merge this, and update the bounds separately

PR #14302, opened 2019-06-17
core.v0.12.0 is not compatible with OCaml 4.08

xclerc:
As shown by [this CI log](http://check.ocamllabs.io/log/1560752820-1bb7c1833f0451f7fc390731698f137850b3125a/4.08.0+rc1/bad/core.v0.12.0).

notes: v0.12.1 was already marked as incompatible, and v.0.12.2 is compatible.


kit-ty-kate:
Thanks!

PR #14324, opened 2019-06-20
Various fixes

kit-ty-kate:
Detected using [check.ocamllabs.io](http://check.ocamllabs.io/)

cc @aryx for `pfff.0.37.1`. Could you return the fix in the upstream opam file if you have one?
cc @leamingrad `ppx_enum.0.0.2` is not compatible with OCaml 4.08

### DIFF
--- a/packages/bistro/bistro.0.0.0/opam
+++ b/packages/bistro/bistro.0.0.0/opam
@@ -19,7 +19,7 @@ depends: [
   "ocamlfind" {build}
   "core" {< "v0.13"}
   "dbm"
-  "lwt"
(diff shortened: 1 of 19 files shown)


aryx:
I don't have an upstream opam package. The one in the opam repository is
the source of truth.

leamingrad:
Thanks for the heads up - I'll try and fix then release a new version (rather than patching up the existing one)

mseri:
opam install -t ocplib-json-typed.0.4 fails on 4.08 and maybe other compilers, similarly digest.0.2 tests are not compatible with 4.08 due to their use of Pervasives

PR #14325, opened 2019-06-20
elpi 1.4.0

gares:


mseri:
Most tests are failing on osx, is that expected?

kit-ty-kate:
Could you return my fix to your opam file upstream?
* `dune install` is not necessary when using opam as dune already create a `.install` file used by opam.
* You don't seem to use dune substitutions
* You don't seem to use ocamlfind either

One other point I'll let you address is your tests: this make use of `dune build` directly and will fail most likely in the future (as `dune build` without `-p` or the release profile transforms warnings into errors). Could you put your tests into dune so that you just have to run `dune runtest -p name -j jobs`?

gares:
Thanks for the fixes.

On my ci pax does fine, so I really don't know what is wrong here.

Wrt putting the tests under dune, I don't get exactly what you ask, I'm not very fluent in dune.

mseri:
Might be due to OSX sandbox

PR #14329, opened 2019-06-21
added packages/cpm/cpm.8.0.0

UnixJunkie:


UnixJunkie:
The TopKeeper module is more generic now.
It can hold any data ('a), instead of just string previously.
The data must still be associated to a float score.

mseri:
Thanks

PR #14330, opened 2019-06-21
camlzip version 1.08

xavierleroy:
Version 1.08 of Camlzip was just released.

### DIFF
--- a/packages/camlzip/camlzip.1.08/opam
+++ b/packages/camlzip/camlzip.1.08/opam
@@ -1,0 +1,34 @@
+opam-version: "2.0"
+maintainer: "Etienne Millon <etienne@cryptosense.com>"
+authors: ["Xavier Leroy"]
+homepage: "https://github.com/xavierleroy/camlzip"
+bug-reports: "https://github.com/xavierleroy/camlzip/issues"
+license: "LGPL-2.1+ with OCaml linking exception"
+build: [
+  [make "all"]
+  [make "allopt"]
+]
+remove: [
+  ["ocamlfind" "remove" "zip"]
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
I improved the opam file by:
* Removing the `remove` field which is unnecessary with opam 2
* Using `conf-zlib` instead of custom depexts
* Tagging `ocamlfind` as a build dependency since you seem to use it only during compilation

xavierleroy:
Thanks a lot for the review and improvements.  However I'm not sure what you mean by "Tagging ocamlfind as a build dependency since you seem to use it only during compilation" ?  Users of this package are supposed to use ocamlfind... (In order to link with the library.)

kit-ty-kate:
@xavierleroy Tagging `ocamlfind` as a build dependency means camlzip won't get rebuilt if ocamlfind is updated. It stays an hard dependency, it's just about rebuilding.
But other than that, `ocamlfind` is not used by `dune` for example, so users won't necessary need it to use `camlzip`.

kit-ty-kate:
Sounds like `alt-ergo < 2.3.0` and `alt-ergo-free` are incompatible with this version of camlzip. Could you add the appropriate constraints to those packages in this PR if you have some time? Otherwise I'll do it later.

xavierleroy:
I'm not comfortable with editing a bunch of packages to add version constraints: I'm sure there are tools to help with this but I don't know them.  It would be very appreciated if someone more familiar with these tools could edit the alt-ergo packages.

mseri:
Thanks and sorry for the delay. I can deal with alt-ergo later

PR #14331, opened 2019-06-21
Package for "num" version 1.2

xavierleroy:
The legacy library for arbitrary-precision arithmetic.  Now with better findlib support, thanks to @dra27.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -17,6 +17,7 @@ packages/freetennis/*/files/freetennis text eol=lf
 packages/ocamlfind/*/files/ocaml-stub text eol=lf
 
 # Treat patches as binary for safety
-*.patch binary
-*.patch.in binary
-*.diff binary
+# NOTE: Temporary commenting the next 3 lines until opam 2.1 is released (see: https://github.com/ocaml/opam/pull/3879)
+#*.patch binary
+#*.patch.in binary
+#*.diff binary
--- a/packages/num/num.1.2/files/installation-warning.patch
+++ b/packages/num/num.1.2/files/installation-warning.patch
@@ -1,0 +1,59 @@
+From db8d748b2cad0adc2698e9fcf28727083a711bae Mon Sep 17 00:00:00 2001
+From: David Allsopp <david.allsopp@metastack.com>
+Date: Wed, 24 Jan 2018 16:01:56 +0000
+Subject: [PATCH] Warn about installations broken by previous faulty package
+
+---
+ Makefile | 33 +++++++++++++++++++++++++++++++++
+ 1 file changed, 33 insertions(+)
+
+diff --git a/Makefile b/Makefile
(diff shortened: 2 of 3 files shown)


xavierleroy:
CI complains:
```
Bad hash for /home/opam/.opam/repo/default/packages/num/num.1.2/files/installation-warning.patch
```
However, the file and the hash were taken verbatim from num.1.1, and the hash looks correct to me.

kit-ty-kate:
I took a look at it and there is indeed no difference. @dra27 could you take a look at it?

kit-ty-kate:
oh, it works on Travis.. Something to do with `.gitattributes` maybe

dra27:
It shouldn’t be .gitattributes, but that’s not to say that it isn’t! However, I wasn’t expecting the new release to need patches / let me have a proper look (but probably tomorrow)

xavierleroy:
The package for 1.1 had two patches, one that is now part of the 1.2 upstream distribution, the other that I kept in the package for 1.2.  It's the MD5 of the latter patch that is causing trouble, perhaps because of a LF/CRLF Windows issue.  I did look around for a `.gitignore` file in the 1.1 package but didn't see any.  Thanks for looking into this.

yakobowski:
@kit-ty-kate can you comment on why you removed the `remove` part of the Opam file?

dra27:
@yakobowski - opam 2 records the files installed by package commands and automatically removes them, so there’s no need for the uninstall steps (I can’t remember if I originally wrote that patch before opam 2 had that, or if I was just forgetting...)

yakobowski:
Neat! Thanks for the explanation.

kit-ty-kate:
I just remembered that I have now access to the server and I can take a deeper look at why it's failing.
And the reason is: .... the file is empty :/
```
-rw-r--r-- 1 opam opam 0 Jun 21 12:52 /home/opam/.opam/repo/default/packages/num/num.1.2/files/installation-warning.patch
```
I don't understand how this is the case but it is.. stay tune for more info~

kit-ty-kate:
With some minor local twitching in the CI and some patches this is now fixed. I'll merge it as soon as possible

kit-ty-kate:
Thanks a lot!

liyishuai:
OPAM [complains](//travis-ci.com/coq-community/reduction-effects/jobs/210175100#L667) upon compiling `num.1.2` that:
> #=== ERROR while compiling num.1.2 ============================================#
Bad hash for /home/travis/.opam/repo/default/packages/num/num.1.2/files/installation-warning.patch

Should I refrain from upgrading `num` for now?

kit-ty-kate:
@liyishuai I assume you have set the target for the opam-repository to be the git repository itself and not http://opam.ocaml.org ? In which case just do `opam repository remove default && opam repository add default http://opam.ocaml.org`

PR #14332, opened 2019-06-21
Package: CharInfo_width 1.1.1

kandu:
* update the combining marks set. remove range (0x200B, 0x200F), some general punctuation marks from it.

### DIFF
--- a/packages/charInfo_width/charInfo_width.1.1.1/opam
+++ b/packages/charInfo_width/charInfo_width.1.1.1/opam
@@ -1,0 +1,27 @@
+opam-version: "2.0"
+maintainer: "zandoye@gmail.com"
+authors: [ "ZAN DoYe" ]
+homepage: "https://bitbucket.org/zandoye/charinfo_width/"
+bug-reports: "https://bitbucket.org/zandoye/charinfo_width/issues"
+license: "MIT"
+dev-repo: "hg+https://bitbucket.org/zandoye/charinfo_width"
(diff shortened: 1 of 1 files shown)


packages/charInfo_width/charInfo_width.1.1.1/opam
```diff
@@ -0,0 +1,27 @@
...
+dev-repo: "hg+https://bitbucket.org/zandoye/charinfo_width"
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test & (ocaml:version >= "4.04.0")}
+]
+depends: [
+  "ocaml" {>= "4.02.3"}
+  "result"
+  "camomile" {>= "1.0.0" & < "2.0~"}
+  "dune" {build}
```
kit-ty-kate:
```suggestion
  "dune" {build & >= "1.4"}
```
Your `dune-project` file says this package requires at least dune 1.4

kit-ty-kate:
Thanks!

kandu:
Sorry. But this PR need to be reverted.

At first, I get the unicode info from https://jrgraphix.net/r/Unicode/2000-206F which confounds [200b, 200f] with [201b, 201f].
After looking into https://unicode.org/charts/PDF/U2000.pdf, I realized that I made a mistake submitting this PR.

Sorry, Kate.

PR #14335, opened 2019-06-21
Package: zed.2.0.2

kandu:
* Zed\_utf8: fix an offset-stepping bug in function `unsafe_extract_prev`

### DIFF
--- a/packages/zed/zed.2.0.2/opam
+++ b/packages/zed/zed.2.0.2/opam
@@ -7,7 +7,7 @@ dev-repo: "git://github.com/ocaml-community/zed.git"
 license: "BSD3"
 depends: [
   "ocaml" {>= "4.02.3"}
-  "dune" {build & >= "1.0.0"}
+  "dune" {build & >= "1.1.0"}
   "base-bytes"
   "camomile" {>= "1.0.1"}
   "react"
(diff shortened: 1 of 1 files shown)


packages/zed/zed.2.0.2/opam
```diff
@@ -0,0 +1,33 @@
+opam-version: "2.0"
+maintainer: "opam-devel@lists.ocaml.org"
+authors: ["Jérémie Dimino"]
+homepage: "https://github.com/ocaml-community/zed"
+bug-reports: "https://github.com/ocaml-community/zed/issues"
+dev-repo: "git://github.com/ocaml-community/zed.git"
+license: "BSD3"
+depends: [
+  "ocaml" {>= "4.02.3"}
+  "dune" {build & >= "1.0.0"}
```
kit-ty-kate:
```suggestion
  "dune" {build & >= "1.1.0"}
```
Your `dune-package` file indicates that this package requires dune >= 1.1

packages/zed/zed.2.0.2/opam
```diff
@@ -0,0 +1,33 @@
...
+  "base-bytes"
+  "camomile" {>= "1.0.1"}
+  "react"
+  "charInfo_width" {>= "1.1.0" & < "2.0~"}
+]
+build: [
+  ["dune" "subst"] {pinned}
```
kit-ty-kate:
```suggestion
```
You don't seem to use dune substitutions anywhere. Is there a reason why you are using `dune subst` here?

kandu: no, this is a good time to clean up some old useless code.
Thank you.

kit-ty-kate:
Thanks!

PR #14346, opened 2019-06-24
Package merlin-extend.0.4

let-def:
### `merlin-extend.0.4`
A protocol to provide custom frontend to Merlin
This protocol allows to replace the OCaml frontend of Merlin.
It extends what used to be done with the `-pp' flag to handle a few more cases.

---
* Homepage: https://github.com/let-def/merlin-extend
* Source repo: git+https://github.com/let-def/merlin-extend.git
* Bug tracker: https://github.com/let-def/merlin-extend

---


kit-ty-kate:
Thanks!

PR #14349, opened 2019-06-24
odoc 1.4.1

aantron:
See the changelog [here](https://github.com/ocaml/odoc/releases/tag/1.4.1).


avsm:

Issue #14344, opened 2019-06-24
Deprecate ulex in favor of ulex-campl5


XVilka:
Because original ulex is abandoned, will never be updated it seems, and will not build with OCaml 4.08, because it depends on campl4, which is [not supporting 4.08](https://github.com/ocaml/camlp4/issues/151) either. Moreover, seems that [ulex-campl5](https://github.com/ocaml/opam-repository/tree/master/packages/ulex-camlp5) is a drop-in replacement for ulex, so shouldn't cause any issues.

See ulex [original](https://github.com/whitequark/ulex) repository, and ulex-campl5 [fork](https://github.com/sacerdot/ulex/tree/ulex-camlp5).

pmetzger:
What still depends on ulex?

XVilka:
@pmetzger: piqi and piqilib, I am working on making BAP work on OCaml 4.08, and those are dependencies.

XVilka:
See the whole list at [opam site](https://opam.ocaml.org/packages/ulex/):
```
    beluga
    ccss<1.6
    cduce
    cow<2.0.0
    frenetic>=3.2.0 & <5.0.0
    fstar>=0.9.5.0
    higlo
    kremlin
    lambdoc
    mezzo
    opa-base
    piqi<0.7.0
    piqilib
    pxp
    rdf>=0.5 & <0.10.0
    sociaml-vcard
    wdialog
```

XVilka:
Focusing on porting to sedlex instead.

PR #14351, opened 2019-06-25
Rename ulex-camlp5 into ulex

XVilka:
Since the ulex-camlp5 is compatible with its campl4 clone, and camlp4 is basically dead, it makes sense to make a swap once and for all. See https://github.com/ocaml/opam-repository/issues/14344 for more information.

### DIFF
--- a/packages/ulex/ulex.1.2-1/opam
+++ b/packages/ulex/ulex.1.2-1/opam
@@ -1,19 +1,19 @@
 opam-version: "2.0"
 maintainer: "claudio.sacerdoticoen@unibo.it"
 authors: ["Alain.Frisch@inria.fr"]
-homepage: "https://github.com/whitequark/ulex"
-dev-repo: "git+https://github.com/whitequark/ulex.git"
-bug-reports: "https://github.com/whitequark/ulex/issues"
+homepage: "https://github.com/sacerdot/ulex"
+dev-repo: "git+https://github.com/sacerdot/ulex.git"
+bug-reports: "https://github.com/sacerdot/ulex/issues"
 synopsis: "A lexer generator for Unicode (backported to camlp5)"
 build: [
   [make]
   [make "all.opt"]
 ]
 install: [make "install"]
-remove: [["ocamlfind" "remove" "ulex-camlp5"]]
+remove: [["ocamlfind" "remove" "ulex"]]
 flags: light-uninstall
 depends: [
-  "ocaml" {>="4.02.0"} 
+  "ocaml" {>="4.02.0"}
   "ocamlfind" {build}
   "camlp5"
   "ocamlbuild" {build}


XVilka:
Seems everything is green, thus such renaming doesn't break anything and good to merge.

XVilka:
Travis failures are because of its network problems (as usual):
```
<><> Gathering sources ><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
[camlp5.7.06.10-g84ce6cc4] downloaded from https://github.com/camlp5/camlp5/archive/84ce6cc4d7738b3c43dfeadd5e9717bb9b50c786.tar.gz
[WARNING] ocamlfind.1.8.0: download of
          http://download.camlcity.org/download/findlib-1.8.0.tar.gz failed
          (http://download.camlcity.org/download/findlib-1.8.0.tar.gz (Curl
          failed: "/usr/bin/curl --write-out %{http_code}\\n --retry 3
          --retry-delay 2 --user-agent opam/2.0.3 -L -o
          /home/travis/.opam/ocaml-base-compiler.4.03.0/.opam-switch/sources/ocamlfind.1.8.0/findlib-1.8.0.tar.gz.part
          http://download.camlcity.org/download/findlib-1.8.0.tar.gz" exited
          with code 7)), trying mirror
[ocamlfind.1.8.0] downloaded from http://download2.camlcity.org/download/findlib-1.8.0.tar.gz
          - ocamlbuild.0.14.0: Curl failed
The command "bash -e .travis-ci.sh" exited with 40.
```
and
```
gpg: keyring `/tmp/tmpip_dda26/secring.gpg' created
gpg: keyring `/tmp/tmpip_dda26/pubring.gpg' created
gpg: requesting key 7BB9C367 from hkp server keyserver.ubuntu.com
Error: retrieving gpg key timed out.
gpg: /tmp/tmpip_dda26/trustdb.gpg: trustdb created
gpg: key 7BB9C367: public key "Launchpad PPA for Ansible, Inc." imported
gpg: Total number processed: 1
gpg:               imported: 1  (RSA: 1)
OK
The command "bash -e .travis-ci-install.sh" failed and exited with 1 during .
```
You might want to restart these jobs just to be sure.

avsm:
cc @alainfrisch -- are you ok with this switch?

XVilka:
Updated PR by keeping the old version of ulex-1.1 as is, lets update compatible ulex-1.2 only.

alainfrisch:
> cc @alainfrisch -- are you ok with this switch?

IIUC, the camlp4 ulex package was maintained by @whitequark , based on his clone.  So this is more a question for him.

whitequark:
I'm rather confused by this package description. It lists `whitequark/ulex` in the `bug-reports` field, but `sacerdot/ulex` in the `src` field. It should be one or the other. /cc @sacerdot

> So this is more a question for him.

Her*

XVilka:
@whitequark yes, I was confused by that description too, but I decided not to touch anything, just move the package, since it was "just working".

XVilka:
Now even camelus is happy. Should this be merged then? The changes required for the opam description/etc can be done separately.

whitequark:
Well, I'd like to not be responsible for bugs for a package whose source code I have no control over.

XVilka:
@whitequark I didn't change it - it was already like that in the opam repository before, just under ulex-camlp5 name.

whitequark:
That doesn't make it any less of a problem.

XVilka:
Let me change the corresponding fields then.

XVilka:
So? What was decided for this one?

XVilka:
Note that camlp5 was updated to support OCaml 4.08 version: https://github.com/camlp5/camlp5/commit/9b9eb124c13f404a0f41bd67f26c1555d97883f9

kit-ty-kate:
I think this change makes sense. However the revdeps only got tested with OCaml 4.08 and we have no view on whether it breaks packages with previous versions of the compiler and it appears that none of the ulex revdeps are themselves compatible with OCaml >= 4.08.

I'll do a separate run to make sure that works first. Ulex was used in conjunction with camlp4 so I'm not too sure a camlp5 extension can be used with camlp4 (I mean I have no idea tbh)
I'll merge this once that succeeds.

kit-ty-kate:
Oh however, I think it makes more sense if this packages has its own package and not just change `ulex.1.2`. Could you move it to `ulex.1.2-1` or `ulex.1.2+camlp5` or something like that, and let `ulex.1.2` as it was?

XVilka:
@kit-ty-kate it had its own package. But I moved this because Camlp5 already provided  OCaml 4.08 version support, while Camlp4 does not, so renaming will help all packages depending on ulex be compatible with OCaml 4.08 automatically, without any changes (if there are no incompatibilites in their own code or other dependencies).

kit-ty-kate:
It's not what I meant. Your PR changes `ulex.1.2`, what I'm saying is just that it would be better to not modify this version and to create a new **version** of `ulex`. It would still be the `ulex` package but it would be more recent version of it.

XVilka:
@kit-ty-kate done. Renamed to ulex.1.2-1

kit-ty-kate:
As expected all the revdeps fail (apart from fstar.0.9.5.0 which apparently doesn't use it if camlp4 is not present or something)
ulex-revdeps
Could you constrain every packages which use `ulex` (simply git grepping for ulex and adding `{<= "1.2"}` should be sufficient)

XVilka:
Hmm, then I am not sure it is needed at all. Idea was to replace the need
of campl4, but if it is used together with ulex so often, it makes sense to
abandon the rename, and just wait for Camlp4 to be upgraded, or port the
packages to sedlex and ditch campl4 in them.

kit-ty-kate:
Also for such a change I would expect the change in version number to be more meaningful than adding a `-1` suffix (sorry for suggesting this). `2.0` seems better in my opinion.

XVilka:
Ok, I am closing this one, since focusing on porting to sedlex instead.

PR #14354, opened 2019-06-25
+mirage-xen-{ocaml/posix}.3.3.1

avsm:
disable 4.08.0 on mirage-xen-ocaml.3.3.0 which fails to compile
on the release version of 4.08.0

* Fix CFLAGS generation with repeated spaces on Linux (mirage/mirage-platform#210 @TheLortex)
* Add support for 4.08 final release (continued from mirage/mirage-platform#206 by @avsm)


avsm:
Bah, Fedora 30 doesn't have find/xargs?!

avsm:
I'll update the fedora base image to deal with this failure

PR #14360, opened 2019-06-26
elpi 1.4.1

gares:


gares:
Linux failure is spurious (network error).
The OSX failure is like last week I guess #14325

packages/elpi/elpi.1.4.1/opam
```diff
@@ -0,0 +1,82 @@
+opam-version: "2.0"
+name: "elpi"
```
mseri:
```suggestion
```

This field is redundant. Opam will override it with the name inferred from the path

gares:
Thanks, I'll fix and push

packages/elpi/elpi.1.4.1/opam
gares: done

kit-ty-kate:
Thanks!

PR #14362, opened 2019-06-26
+ elasticsearch-cli 1.0

cyberhuman:


packages/elasticsearch-cli/elasticsearch-cli.1.0/opam
```diff
@@ -0,0 +1,28 @@
+opam-version: "2.0"
+version: "1.0"
```
mseri:
```suggestion
```

This field is redundant, opam will infer it from the path and override it. I am not allowed to commit the change myself for some reason

cyberhuman: Ok, removed it.
It was added because `ocamlfind query -format %v elasticsearch-cli` would return `[unspecified]` without it. But I can't make it work anymore and am not willing to investigate further at this time.

mseri:
Thanks!

PR #14369, opened 2019-06-26
mirage-console: older versions do not work with new mirage-xen

avsm:
due to xen-gnt addition into mirage-xen.3.3.0+

revdeps for #14366


avsm:
solver timeouts, merging

PR #14382, opened 2019-06-27
netchannel / mirage-block-xen: require mirage-xen 4.0.0

hannesm:
/cc @avsm

(see failing CI at https://travis-ci.org/ocaml/opam-repository/jobs/550981012 and https://travis-ci.org/ocaml/opam-repository/jobs/550996224)


avsm:
So strange; I'm sure I checked this in the opam file with the pin-depends! Thanks!

avsm:
Ah it's because I was pinning to `.dev`

PR #14390, opened 2019-06-28
Package bnfgen.2.0

dmbaturin:
### `bnfgen.2.0`
Random text generator that takes context-free grammars from BNF files
BNFGen generates random texts based on user-defined context-free grammars
specified in a BNF-like syntax.
You can specify "weight" for rules with alternation to change the probabilities,
for example, in `<foo> ::= 10 <foo> "foo" | "foo";` the recursive option will be
ten times more likely to be taken.

---
* Homepage: https://github.com/dmbaturin/bnfgen
* Source repo: git+https://github.com/dmbaturin/bnfgen
* Bug tracker: https://github.com/dmbaturin/bnfgen/issues

---

### DIFF
--- a/packages/bnfgen/bnfgen.2.0/opam
+++ b/packages/bnfgen/bnfgen.2.0/opam
@@ -19,12 +19,12 @@ build: [
 ]
 depends: [
   "menhir" {build}
-  "dune" {build & >= "1.4.0"}
+  "dune" {>= "1.4.0"}
 ]
 url {
   src: "https://github.com/dmbaturin/bnfgen/archive/2.0.tar.gz"
   checksum: [
-    "md5=89ae8d4e9db009b8b54ebcc9df18715d"
-    "sha512=98ee1c9cf2506b6aa3b7ea24c8fb7f3b7166c81be677d06895ba2e76ae34e259315ec99308457e121e82f6ab727325afb4ddd03b297c420e7c41583510e35c48"
+    "md5=efb520633eac4197509ddc9e020c53a9"
+    "sha512=840be6c7bb98590457aa9fd64593eb5995a5ae9d77cb4a883cb2f5ff65008de5dd29f3b2d86a746d5edafb7c449b0223a7c82d6ec1a3209c98c9430609e117a7"
   ]
 }
--- a/packages/bnfgen/bnfgen.2.0/opam
+++ b/packages/bnfgen/bnfgen.2.0/opam
@@ -1,0 +1,31 @@
+opam-version: "2.0"
+synopsis: "Random text generator that takes context-free grammars from BNF files"
+description: """
+BNFGen generates random texts based on user-defined context-free grammars
+specified in a BNF-like syntax.
+You can specify "weight" for rules with alternation to change the probabilities,
+for example, in `<foo> ::= 10 <foo> "foo" | "foo";` the recursive option will be
+ten times more likely to be taken.
+"""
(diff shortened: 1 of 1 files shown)


dmbaturin:
This is a new release and a rename of https://opam.ocaml.org/packages/cfgen/

I've mostly been using it for my own needs, but then I found that there are two more projects with that name, for different purposes (one not related to formal languages at all, the other extracts grammars from a corpus rather than takes an explicit definition), so I tried to make the name more descriptive.

I couldn't find if there's a procedure for renaming opam packages of whether it ever happened before. If any steps are needed to properly rename the package, please let me know.

avsm:
Thanks! It depends what you prefer -- if there are no other packages depending on it, then you can do a repository rename. Alternatively just put a prominent message in the metadata of the old package that it haas been deprecated and to direct users to the new name.

dmbaturin:
@avsm That's the reason I decided to go through with the rename really—nothing _can_ depend on it because it only provides executables, not modules (for better or worse, maybe something to consider for a future update, along with ABNF support).

I'll make a separate pull request to add a deprecation warning to the original cfgen then. jbuilder seems to provide a good example of how it's done.

packages/bnfgen/bnfgen.2.0/opam
```diff
@@ -0,0 +1,31 @@
...
+license: "MIT"
+homepage: "https://github.com/dmbaturin/bnfgen"
+bug-reports: "https://github.com/dmbaturin/bnfgen/issues"
+dev-repo: "git+https://github.com/dmbaturin/bnfgen"
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name]
+]
+depends: [
+  "ocamlfind" {build}
```
mseri: Is ocamlfind really still used in the build?

dmbaturin: Good point. I've updated the PR.

mseri:
Have you modified the tarball?
```
bnfgen.2.0:
            https://github.com/dmbaturin/bnfgen/archive/2.0.tar.gz (Bad checksum, expected md5=89ae8d4e9db009b8b54ebcc9df18715d)
```

dmbaturin:
In this case, I have to admit, yes. I've found a stupid mistake and thought it was safe to sneakily overwrite the tag until the PR is merged and get away with it. Sorry for making this PR so much trouble.

packages/bnfgen/bnfgen.2.0/opam
```diff
@@ -0,0 +1,30 @@
...
+homepage: "https://github.com/dmbaturin/bnfgen"
+bug-reports: "https://github.com/dmbaturin/bnfgen/issues"
+dev-repo: "git+https://github.com/dmbaturin/bnfgen"
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name]
+]
+depends: [
+  "menhir" {build}
+  "dune" {build & >= "1.4.0"}
```
mseri:
```suggestion
  "dune" {>= "1.4.0"}
```

mseri:
If you update the opam file and use the correct checksum for the tarball, this could be merged

packages/bnfgen/bnfgen.2.0/opam
dmbaturin: Fixed, please check.

mseri:
Thanks. The CI is a bit busy right now. I'll wait for Travis to do its job before merging.

PR #14401, opened 2019-06-28
frama-c: mark incompatible with 4.08

maroneze:
We just identified an issue with unmarshalling in OCaml 4.08, which prevents Frama-C from working properly in some cases.


mseri:
Thanks

PR #14403, opened 2019-06-28
Deprecate cfgen, since it was renamed to bnfgen

dmbaturin:
Related to #14390 
I've added prominent deprecation warnings to the description and post-messaged

Since github now leaves proper redirects after renaming repos, and the 1.0 tag is still there, cfgen still seems to install fine.

### DIFF
--- a/packages/cfgen/cfgen.1.0/opam
+++ b/packages/cfgen/cfgen.1.0/opam
@@ -2,31 +2,15 @@ opam-version: "2.0"
 maintainer: "Daniil Baturin <daniil@baturin.org>"
 authors: "Daniil Baturin <daniil@baturin.org>"
 homepage: "https://github.com/dmbaturin/cfgen"
-bug-reports: "https://github.com/dmbaturin/cfgen/issues"
+bug-reports: "https://github.com/dmbaturin/bnfgen/issues"
 license: "MIT"
-dev-repo: "git+https://github.com/dmbaturin/cfgen.git"
-build: [
-  ["./configure" "--prefix=%{prefix}%"]
-  [make]
-]
-install: [make "install"]
-remove: ["ocamlfind" "remove" "cfgen"]
-depends: [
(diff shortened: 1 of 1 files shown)


dmbaturin:
The build failed with:
```
          - cfgen.1.0:
            https://github.com/dmbaturin/cfgen/archive/1.0.zip (Bad checksum, expected md5=8036d9283ae8c83a710872a9455f4435)
```

I wonder why it works on my machine. Where is the hash sum stored?

mseri:
The hash used by opam is here: https://github.com/ocaml/opam-repository/pull/14403/files#diff-42adeb10eff6feab17b4bb9a52e6d72fR31

If I download the zip file and check the hash I get 996c7456bef1a758b05c8592d7e99bbd (that differs from the one in the opam file)

dmbaturin:
I see. Yeah, repository redirects, but inside, the dir name is different, so the hash is obviously different as well.
Would it be better to make it a transition package with bnfgen in dependencies, like jbuilder maintainers did when they renamed it to dune?

mseri:
I think it would be fine to update the checksum (maybe also of previous packages if present) or we can fish the old tarballs from the opam cache and upload them in the github releases sources sections (but I can never remember where the cache is, I will need to ask)

gasche:
If the content of the archives is exactly the same, then I think that changing the checksum would be fine.

dmbaturin:
@gasche The contents are exactly the same up to the name inside the ZIP archive (github changed it for the old tag), but that's likely problematic enough.
I've pushed a change to make it a transition package instead, in the same way as jbuilder did it.

Frankly, since it's a package nobody cares about (whole 7 installations in the last month), I don't think anyone would notice if it was just removed.

mseri:
Waiting for https://github.com/ocaml/opam-repository/pull/14390 to be merged

PR #14410, opened 2019-07-01
added packages/parany/parany.6.0.0

UnixJunkie:
There are two new functions to make the library safer for users:
Parany.set_copy_on_work() and
Parany.set_copy_on_mux()


mseri:
Thanks

PR #14411, opened 2019-07-01
Release ocaml-jupyter kernel v2.5.0

akabe:
Release note: https://github.com/akabe/ocaml-jupyter/releases/tag/v2.5.0


mseri:
I have tested it on two different osx versions, and it seems to work fine. I don't want to re-run the test as the install segfault may be useful for opam maintainers.

Issue #14318, opened 2019-06-19
-static switch: undefined reference to `secure_getenv'


copy:
Compiling a program on a `4.08.0+musl+static+flambda` with `-ccopt -static -cc musl-gcc` fails:

```
(cd _build/default && ~/.opam/4.08.0+musl+static+flambda/bin/ocamlopt.opt -w @a-4-29-40-41-42-44-45-48-58-59-60-40 -strict-sequence -strict-formats -short-paths -keep-locs -ccopt -static -cc musl-gcc -g -o ping2.exe .ping2.eobjs/native/ping2.cmx)
/usr/bin/ld: ~/.opam/4.08.0+musl+static+flambda/lib/ocaml/libasmrun.a(unix_n.o): in function `caml_secure_getenv':
~/.opam/4.08.0+musl+static+flambda/.opam-switch/build/ocaml-variants.4.08.0+musl+static+flambda/runtime/unix.c:414: undefined reference to `secure_getenv'
collect2: error: ld returned 1 exit status
File "caml_startup", line 1:
Error: Error during linking
```

musl doesn't have `secure_getenv`, and I believe this case is handled by OCaml, so something must be going wrong while this switch is being configured.

etanol:
Try the following:

1. Enable the switch you are using of the *musl+static* variant: `eval $(opam env)`
1. Get the source of the `ocaml-variants` package and pin it: `opam source --pin ocaml-variants.4.08.0+musl+static+flambda`
    - Don't build it yet
1. Go to the created directory: `cd ocaml-variants.4.08.0+musl+static+flambda`
1. Edit the `ocaml-variants.opam` file
    - Look for the `build` section and see if the call to `configure` looks like this:
      ```
      ["./configure" "--prefix=%{prefix}%"]
      ```
    - Replace it with the following:
      ```
      [
        "./configure"
        "--prefix=%{prefix}%"
        "CC=musl-gcc -Os"
        "ASPP=musl-gcc -c"
        "--enable-flambda"
        "LIBS=-static"
        "--disable-graph-lib"
        "--disable-shared"
      ]
      ```
1. Build it: `opam install .`
    - Don't worry about the version *downgrade*
1. Compile your program again
    - Note that you should **not** need `-cc musl-gcc`

copy:
@etanol That fixed the problem, cheers!

copy:
[...]

PR #14414, opened 2019-07-01
Enable musl variant under Linux.

etanol:
The `musl+static+flambda` variants for 4.07.1 and 4.08.0 were disabled when running under linux.  Even when the OPAM description explicitly mentions [Arch Linux][a] and [Debian][d].

This changes the os filters to use the [musl][m] static build configuration options.

[a]: https://www.archlinux.org/
[d]: https://www.debian.org/
[m]: http://www.musl-libc.org/

### DIFF
--- a/packages/ocaml-variants/ocaml-variants.4.07.1+musl+static+flambda/opam
+++ b/packages/ocaml-variants/ocaml-variants.4.07.1+musl+static+flambda/opam
@@ -15,7 +15,7 @@ flags: compiler
 setenv: CAML_LD_LIBRARY_PATH = "%{lib}%/stublibs"
 build: [
   ["./configure" "-prefix" prefix "-with-debug-runtime"]
-    {os != "openbsd" & os != "freebsd" & os != "macos"}
+    {os != "openbsd" & os != "freebsd" & os != "macos" & os != "linux"}
   [
     "./configure"
     "-prefix"
--- a/packages/ocaml-variants/ocaml-variants.4.08.0+musl+static+flambda/opam
+++ b/packages/ocaml-variants/ocaml-variants.4.08.0+musl+static+flambda/opam
@@ -19,7 +19,7 @@ flags: compiler
 setenv: CAML_LD_LIBRARY_PATH = "%{lib}%/stublibs"
 build: [
   ["./configure" "--prefix=%{prefix}%"]
-    {os != "openbsd" & os != "freebsd" & os != "macos"}
+    {os != "openbsd" & os != "freebsd" & os != "macos" & os != "linux"}
   [
     "./configure"
     "--prefix=%{prefix}%"
@@ -29,7 +29,7 @@ build: [
     "LIBS=-static"
     "--disable-graph-lib"
     "--disable-shared"
-  ] {os = "openbsd" | os = "freebsd" | os = "macos"}
+  ] {os = "openbsd" | os = "freebsd" | os = "macos" | os = "linux"}
   [make "world"]
   [make "world.opt"]
 ]


mseri:
I don't understand why there are two configure steps that may run twice, and why running configure once or twice actually makes a difference. Probably somebody else needs to review this

etanol:
> I don't understand why there are two configure steps that may run twice, and why running configure once or twice actually makes a difference. Probably somebody else needs to review this

Those don't run twice.  Notice the [Filter between braces][f] after each command.  One is the complement of the other, so only one `configure` command will run.

The issue here is that, currently, building under Linux does **not** compile OCaml using musl, from the `musl+static+flambda` flavor when it actually should.  This is because the filters are incomplete.

[f]: https://opam.ocaml.org/doc/Manual.html#Filters

copy:
Could this fix https://github.com/ocaml/opam-repository/issues/14318?

etanol:
> Could this fix #14318?

I believe so, because it seems that the OCaml internal libs are linked against glibc instead of musl when building on Linux.

avsm:
Could you give me an example of where this is failing? I don't understand this PR at all -- you see to be skipping the configure script when running on Linux?

For me,

```
$ opam switch create musl ocaml-variants.4.07.1+musl+static+flambda 
$ ocamlc -config|grep native
native_c_compiler: musl-gcc -Os -O2 -fno-strict-aliasing -fwrapv -D_FILE_OFFSET_BITS=64 -D_REENTRANT
native_c_libraries: -static  -lm 
native_pack_linker: ld -r -o 
```

which seems correct...?

etanol:
> Could you give me an example of where this is failing? I don't understand this PR at all -- you see to be skipping the configure script when running on Linux?

The issue happens on 4.08.0+musl+static+flambda only, for which the `opam` file differs slightly.

- [This is the code for `4.07.1+musl+static+flambda`][1]. In this case, under Linux, `./configure` is run twice.  But it doesn't affect the result, just the build time.
- [This is the code for `4.08.0+musl+static+flambda`][2]. Here you can see that the two `./configure` commands are filtered with complementary predicates.  Therefore, under Linux, only the **first** `./configure` is run.

Steps to reproduce:

```
opam switch create buggy 4.08.0+musl+static+flambda
opam exec --switch=buggy -- ocamlc -config | grep native
```

What you get under Linux is:

```
native_c_compiler: gcc -O2 -fno-strict-aliasing -fwrapv -D_FILE_OFFSET_BITS=64 -D_REENTRANT
native_c_libraries: -lm -ldl 
native_pack_linker: ld -r -o 
```

My apologies for not being more clear from the beginning.

[1]: https://github.com/ocaml/opam-repository/blob/master/packages/ocaml-variants/ocaml-variants.4.07.1%2Bmusl%2Bstatic%2Bflambda/opam#L17-L34
[2]: https://github.com/ocaml/opam-repository/blob/master/packages/ocaml-variants/ocaml-variants.4.08.0%2Bmusl%2Bstatic%2Bflambda/opam#L21-L32

avsm:
Thanks for the repro case! Why is the PR modifying the 4.07.1 opam file as well, if the bug only appears on 4.08.0?

etanol:
The idea was to avoid calling `./configure` two times.  But I'm happy to update the PR to leave it as it is now.

avsm:
Ah I see. The logic seems a little convoluted that was there before your PR.  It doesn't make much sense to have a non-Linux option at all for a musl-specific switch. We should probably make this available only on `os=linux`. But that can be done in a separate PR -- thanks for this fix!

etanol:
Thanks!

PR #14417, opened 2019-07-01
Tentative fix for hdf5.0.1.5

xclerc:
[This log](http://check.ocamllabs.io/log/1561966781-8775a4b07059f078d092484a05a235c3c9ced60e/4.08.0/bad/hdf5.0.1.5) exhibits the issue.

I am not positive this is the best fix, so feedback is more than welcome.


xclerc:
@vbrankov

mseri:
I think that is the correct fix. It would have been better if the package was called `hdf5`, and the library as well instead of having a name and a different library name (according to the source the library seems to be `hdf5-ocaml.caml`). Please consider making the names uniforms in a future version

mseri:
Thanks

PR #14423, opened 2019-07-02
Update package override.0.2.0

thierry-martinez:
- compatibility with OCaml 4.08.0

- remove dependency to ppx_tools

- add examples/typedtree_collect_texp_apply

- bootstrapped equivalence checking for Parsetree.core_type: matching
  between types for applying rewriting rules is now complete, and is
  implemented by overriding Parsetree and deriving eq. The former
  incomplete equivalence checking is used to bootstrap.

- support aliases even if the target module is defined in the same
  module (we do not take declaration order into account yet, so there
  can be wrong shadowings, and even loops, even if we suppress the
  trivial ones).

- renamed types are substituded globally in mutually recursive type
  definitions.

### DIFF
--- a/packages/override/override.0.2.0/opam
+++ b/packages/override/override.0.2.0/opam
@@ -1,0 +1,31 @@
+opam-version: "2.0"
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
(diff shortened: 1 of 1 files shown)


thierry-martinez:
Thank you, @samoht !

samoht:
Thank you for writing the package :-)

PR #14426, opened 2019-07-04
added packages/pardi/pardi.1.0.1/opam

UnixJunkie:
added two options -ie and -oe for programs that enforce input
and output file extensions.
Better help message too.


mseri:
Thanks

PR #14427, opened 2019-07-04
[bug-fix release] override 0.2.1

thierry-martinez:
Changes with respect to 0.2.0:
- attributes applied on `[%%types]` are applied once by declaration
  group. (Previously, attributes were applied to each types in a group,
  leading ppxlib's deriving to make n * n expansions for a group of
  n types -- one for each type/attribute pair).

### DIFF
--- a/packages/override/override.0.2.1/opam
+++ b/packages/override/override.0.2.1/opam
@@ -1,0 +1,31 @@
+opam-version: "2.0"
+build: [
+  ["dune" "subst"] {pinned}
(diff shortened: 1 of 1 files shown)


thierry-martinez:
The failure on Travis CI with OCaml 4.02 looks unrelated to the package itself. I think everything else is ok (despite camelus is still in progress for days!).

kit-ty-kate:
Thanks!

thierry-martinez:
Thank you very much, @kit-ty-kate !

PR #14431, opened 2019-07-04
obus: update to new github repository and add new release

Freyr666:
* a new release of `obus` library (OCaml native D-Bus interface).
* `obus` moved to `ocaml-community` repo
* mpris.0.1.1: add missing lwt_camlp4 dependency


Freyr666:
It seems that mpris.0.1.1 fails not due to obus, but due to the lack of lwt.syntax. Should I add the constraints?

mseri:
@Freyr666 if you could do that, it would be great and very appreciated.

mseri:
Thanks!

PR #14440, opened 2019-07-05
Use os-family = "debian"

toots:
We've had one case of a user whose `os-distribution` was `raspbian` but `os-family` is `debian`. If both `debian` and `ubuntu` depexts are the same then there's no reason to not use `os-family = "debian"` instead, which is what this PR does.

### DIFF
--- a/packages/0install/0install.2.10/opam
+++ b/packages/0install/0install.2.10/opam
@@ -22,8 +22,7 @@ depends: [
 ]
 depopts: [ "obus" "lablgtk" ]
 depexts: [
-  ["unzip"] {os-distribution = "ubuntu"}
-  ["unzip"] {os-distribution = "debian"}
+  ["unzip"] {os-family = "debian"}
 ]
 synopsis: "The antidote to app-stores"
 description: """
--- a/packages/0install/0install.2.11/opam
+++ b/packages/0install/0install.2.11/opam
@@ -22,8 +22,7 @@ depends: [
 ]
 depopts: [ "obus" "lablgtk" ]
 depexts: [
-  ["unzip"] {os-distribution = "ubuntu"}
-  ["unzip"] {os-distribution = "debian"}
+  ["unzip"] {os-family = "debian"}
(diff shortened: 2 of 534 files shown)


toots:
Wrote a parser/printer for it in case that can be helpful for someone else:
```ocaml
open OpamParser
open OpamParserTypes

exception Found of string

let get_distro name values =
  try
    List.iter (function
      | Option (_, List (_,[String (_,pkg)]), [
          Relop (_, `Eq, Ident (_, "os-distribution"),
                         String (_, distro))
        ]) when distro = name -> raise (Found pkg)
      | _ -> ()) values;
    raise Not_found
  with Found pkg -> pkg

let filter_distro values =
  let values =
    List.filter (function
      | Option (_, _, [
          Relop (_, `Eq, Ident (_, "os-distribution"),
                         String (_, "ubuntu"))
        ]) -> false
      | _ -> true) values
  in
  List.map (function
    | Option (pos1, values, [
       Relop (pos2, `Eq, Ident (pos3, "os-distribution"),
                         String (pos4, "debian"))
      ]) -> Option (pos1, values, [
       Relop (pos2, `Eq, Ident (pos3, "os-family"),
                         String (pos4, "debian"))
       ])
    | x -> x) values

let filter_depext values =
  try
    let debian = get_distro "debian" values in
    let ubuntu = get_distro "ubuntu" values in
    if debian = ubuntu then
      filter_distro values
    else
      values
  with Not_found -> values

let check_depext_value = function
  | List (pos, values) -> List (pos, filter_depext values)
  | x -> x

let check_content = List.map (function
  | Variable (pos, "depexts", v)  -> Variable (pos, "depexts", check_depext_value v)
  | x -> x)

let () =
  let orig = Sys.argv.(1) in
  let f = OpamParser.file orig in
  let f = {f with
    file_contents=check_content f.file_contents
  } in
  print_string
    (OpamPrinter.Preserved.opamfile ~format_from:orig f)
```

PR #14442, opened 2019-07-05
Lo is not available on alpine.

smimram:
An user informed us that lo-dev is not available on alpine. See https://github.com/savonet/ocaml-lo/issues/1.


mseri:
Thanks

PR #14452, opened 2019-07-08
Package omonad.omonad-0.4.0

yallop:
### `omonad.omonad-0.4.0`
Monad programming using ppx preprocessor

---
* Homepage: https://github.com/yallop/omonad
* Source repo: git+https://github.com/yallop/omonad.git
* Bug tracker: https://github.com/yallop/omonad/issues

---


kit-ty-kate:
Seems like opam-publish based the version for this package upon the tag name which is not correct. I fixed that.

I also removed some unnecessary fields (since opam 2.0) and I'm pretty sure oasis and ocamlfind don't are not dependencies that need omonad to be recompiled so it's better to tag them as build dependencies.

Could you return those fixes to the opam file upstream?

yallop:
Thanks!  Done: https://github.com/yallop/omonad/commit/a84d059cf070bef47e307cbcb5ec8758c26c5578

kit-ty-kate:
Thanks!

PR #14460, opened 2019-07-09
Package ocolor.1.2

marc-chevalier:
### `ocolor.1.2`

---

---

### DIFF
--- a/packages/ocolor/ocolor.1.2/opam
+++ b/packages/ocolor/ocolor.1.2/opam
@@ -8,7 +8,6 @@ depends: [
   "cppo" {build & >= "1.6.5"}
 ]
 build: [
-  ["dune" "subst"] {pinned}
   ["dune" "build" "-p" name "-j" jobs]
 ]
 homepage: "https://github.com/marc-chevalier/ocolor"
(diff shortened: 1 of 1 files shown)


packages/ocolor/ocolor.1.2/opam
```diff
@@ -0,0 +1,36 @@
...
+synopsis: "Print with style in your terminal using Format's semantic tags"
+maintainer: "Marc Chevalier <ocolor@marc-chevalier.com>"
+authors: "Marc Chevalier <ocolor@marc-chevalier.com>"
+depends: [
+  "ocaml" {>= "4.02.3"}
+  "dune" {build & >= "1.6.3"}
+  "cppo" {build & >= "1.6.5"}
+]
+build: [
+  ["dune" "subst"] {pinned}
```
kit-ty-kate:
```suggestion
```
You don't seem to use dune substitutions

marc-chevalier: No, indeed. A stupid copy-paste of some documentation. Thanks!

gasche:
Thanks! Merging.

PR #14465, opened 2019-07-10
ppx_deriving_yojson: 3.5 release with support for OCaml 4.08

gasche:
3.5
---

  * use tail-recursive functions to (de)serialize long lists
    (ocaml-ppx/ppx_deriving_yojson#97)
    Alex Knauth
  * Support for OCaml 4.08
    (ocaml-ppx/ppx_deriving_yojson#99)
    Antonio Nuno Monteiro

### DIFF
--- a/packages/jupyter/jupyter.0.0.0/opam
+++ b/packages/jupyter/jupyter.0.0.0/opam
@@ -26,7 +26,7 @@ remove: [
   [ "rm" "-rf" "%{share}%/ocaml-jupyter" ]
 ]
 depends: [
-  "ocaml" {>= "4.02.0"}
+  "ocaml" {>= "4.02.0" & < "4.08.0"}
   "base-threads"
   "base-unix"
   "uuidm"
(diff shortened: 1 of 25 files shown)


gasche:
The remaining revdeps faliure were all `jupyter`-related. Some of them I could reproduce and fix, some others I was unable to reproduce locally, so I plan to just leave them as they are.

(In case anyone is curious, the error message for the builds that I cannot reproduce look as follows. It looks like a jbuilder+ppx_deriving+ocaml-migrate-parsetree integration issue, but reusing the exact same version of all those packages gives me a working build, so I cannot find the problem.

```
+ /home/opam/.opam/4.02/bin/jbuilder "build" "-p" "jupyter" "-j" "2" (CWD=/home/opam/.opam/4.02/.opam-switch/build/jupyter.2.2.0)
- Error: I couldn't find 'ocaml-migrate-parsetree.driver-main'.
- I need this library in order to use ppx rewriters.
- See the manual for details.
- Hint: opam install ocaml-migrate-parsetree
```
)

PR #14468, opened 2019-07-10
Package prepare.3.1

anscoil:
### `prepare.3.1`
Javalib is a library written in OCaml with the aim to provide a high level representation of Java .class files
Thus it stands for a good starting point for people who want to
develop static analyses for Java byte-code programs, benefiting from
the strength of OCaml language.

---
* Homepage: https://javalib-team.github.io/javalib/
* Source repo: git+https://github.com/javalib-team/javalib.git
* Bug tracker: https://github.com/javalib-team/javalib/issues

---

### DIFF
--- a/packages/prepare/prepare.3.1/opam
+++ b/packages/prepare/prepare.3.1/opam
@@ -1,0 +1,36 @@
+opam-version: "2.0"
+maintainer: "Nicolas Barré <nicolas.barre@irisa.fr>"
+authors: "Javalib Development team"
+homepage: "https://javalib-team.github.io/javalib/"
+bug-reports: "https://github.com/javalib-team/javalib/issues"
+license: "LGPL-2.1 with OCaml linking exception"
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
I've removed the remove field which is unecessary since OPAM 2.0. Could you return the fix in the opam file upstream?

kit-ty-kate:
Everything is good otherwise. Thanks!

mseri:
@kit-ty-kate @anscoil is the name “prepare” correct? I would have expected “javalib”

anscoil:
Yes you're right ! I did a mistake with opam-publish, sorry...
Should I do it properly and republish ?
Thanks

mseri:
No, need, I'll move the file and send a PR. Thanks

anscoil:
> I've removed the remove field which is unecessary since OPAM 2.0. Could you return the fix in the opam file upstream?

Thanks, I did it so it will not appear anymore in the next release.

PR #14471, opened 2019-07-10
Markup.ml 0.8.1: standards-compliant HTML5 parser

aantron:
From the [changelog](https://github.com/aantron/markup.ml/releases/tag/0.8.1):

> Additions
>
> - Ability to optionally pass custom escaping functions to `Markup.write_html` (aantron/markup.ml#42, Julien Sagot).
>
> Bugs fixed
>
> - `Markup.html5` and `Markup.xhtml` filtered their inputs incorrectly (aantron/markup.ml#41).
> - XML parser placed unnecessary restrictions on attribute names (aantron/markup.ml#43, @jeffa5).
> - HTML parser dropped characters in URL query parameter names (aantron/markup.ml#44, Julien Sagot).

### DIFF
--- a/packages/markup/markup.0.8.1/opam
+++ b/packages/markup/markup.0.8.1/opam
@@ -1,0 +1,54 @@
+opam-version: "2.0"
+
+synopsis: "Error-recovering functional HTML5 and XML parsers and writers"
+
+version: "0.8.1"
+license: "BSD"
+homepage: "https://github.com/aantron/markup.ml"
+doc: "http://aantron.github.io/markup.ml"
+bug-reports: "https://github.com/aantron/markup.ml/issues"
+
+authors: "Anton Bachin <antonbachin@yahoo.com>"
+maintainer: "Anton Bachin <antonbachin@yahoo.com>"
+dev-repo: "git+https://github.com/aantron/markup.ml.git"
+
+depends: [
+  "dune" {build}
+  "uchar"
+  "uutf" {>= "1.0.0"}
+
+  "bisect_ppx" {dev & >= "1.3.0"}
+  "ounit" {dev}
+]
+# Markup.ml implicitly requires OCaml 4.02.3, as this is a contraint of Dune.
+# Without that, Markup.ml implicitly requires OCaml 4.01.0, as this is a
+# constraint of Uutf. Without *that*, Markup.ml works on OCaml 3.11 or earlier.
+
+build: [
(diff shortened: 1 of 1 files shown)


packages/markup/markup.0.8.1/opam
```diff
@@ -0,0 +1,54 @@
...
+maintainer: "Anton Bachin <antonbachin@yahoo.com>"
+dev-repo: "git+https://github.com/aantron/markup.ml.git"
+
+depends: [
+  "dune" {build}
+  "uchar"
+  "uutf" {>= "1.0.0"}
+
+  "bisect_ppx" {dev & >= "1.3.0"}
+  "ounit" {dev}
```
kit-ty-kate:
```suggestion
  "ounit" {with-test}
```
Are you sure it's `dev` and not `with-test` that you wanted here?

aantron: I chose `dev` because on a development pin, `ounit` should be installed. It won't be installed if `with-test` is used.

kit-ty-kate: I'm surprised. How do you manage to detect if it's a development pin or a release? How do you use `ounit` for the compilation?

aantron: I don't know exactly how opam detects this, but compare the dependency cone of `opam install odoc` and `opam pin add odoc --dev-repo`. The extra dependencies are marked with `{dev}` and their transitive dependencies.

I think the issue is that there isn't a good way (AFAIK) to distinguish between a pin for development and a git pin for another reason. So far, I've preferred to overestimate the dependencies for non-development git pins, rather than underestimate dependencies for development.

aantron:
Ping!

kit-ty-kate:
cc @rjbou would you know why Camelus is giving an error but the rest of the CI succeed. I'm guessing it's about the size of the hash (which changed in the last force-push)

packages/markup/markup.0.8.1/opam
kit-ty-kate: It's not really what I meant but it's ok. It's not really important anyway.

aantron:
There were some stray characters in the hash in the original commit.

avsm:
Looks fine to merge, as the bot error isn't a blocker for the package becoming available.

aantron:
Thanks, @kit-ty-kate and @avsm!

PR #14476, opened 2019-07-11
New version of visitors.

fpottier:
This version should be compatible with 4.08.


kit-ty-kate:
The remove command is unnecessary since OPAM 2.0 and the opam linter complained about the dot at the end of the synopsis. Could you return the fix to the opam file upstream?

kit-ty-kate:
Thanks!

fpottier:
> Could you return the fix to the opam file upstream?

Done, thanks!

PR #14477, opened 2019-07-11
Add info for nixos (opam2nix)

bobot:
Mainly add nixos precise packages, for use with @timbertson's opam2nix.


packages/zarith/zarith.1.7/opam
```diff
@@ -33,7 +33,9 @@ depends: [
   "ocamlfind"
   "conf-gmp"
   "conf-perl" {build}
+  "conf-gcc" {build}
 ]
+patches: [ "absolute_CC.patch" ]
```
avsm: I think we need to bump the version of zarith here rather than adding this patch unconditionally. Perhaps `zarith.1.7a`.  Otherwise if it causes issues for other users, they cannot revert to the old version easily.

timbertson:
Thanks @bobot ! The os-distribution opam2nix declares is `nixpkgs` though, not `nixos`: https://github.com/timbertson/opam2nix/blob/c4c66eda70e40998180eb086aa6dbcc4d92dd83b/src/opam_metadata.ml#L107

(I debated this, but nixpkgs is the name of the package set, so it seemed the most appropriate)

packages/zarith/zarith.1.7/opam
kit-ty-kate: `zarith.1.7-1` would be more standard I think

packages/conf-gcc/conf-gcc.1.0/opam
```diff
@@ -0,0 +1,22 @@
+opam-version: "2.0"
+maintainer: "unixjunkie@sdf.org"
+authors: ["Francois Berenger" "Francois Bobot"]
+homepage: "https://github.com/ocaml/opam-repository"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: "GPL-2+"
+build: [["which" "gcc"]]
```
kit-ty-kate: I think it would be simpler to just call `gcc --version` instead of relying on `which`

packages/coq/coq.8.8.2/opam
```diff
@@ -21,7 +21,7 @@ build: [
     "-docdir" doc
     "-libdir" "%{lib}%/coq"
     "-datadir" "%{share}%/coq"
-    "-camlp5dir" "%{lib}%/camlp5"
+    "-camlp5dir" "%{camlp5:lib}%"
```
kit-ty-kate: why is the reason behind this change? Also why change it for almost all versions but the latests

packages/zarith/zarith.1.7/opam
bobot: ok

packages/coq/coq.8.8.2/opam
bobot: Nix put each packages in its own directory. The second form is more precise so works in any cases.

packages/conf-gcc/conf-gcc.1.0/opam
bobot: Done

packages/opus/opus.0.1.2/opam
```diff
@@ -23,6 +23,7 @@ depexts: [
   ["opus-devel"] {os-distribution = "fedora"}
   ["opus-devel"] {os-family = "suse"}
   ["libopus"] {os-distribution = "nixos"}
```
kit-ty-kate: Could you remove the `nixos` line if it's not used instead? (it's definitly not used by opam-depext)

packages/zarith/zarith.1.7a/opam
```diff
@@ -0,0 +1,50 @@
+opam-version: "2.0"
+name: "zarith"
+version: "1.7"
```
kit-ty-kate:
```suggestion
```
This line makes the package not work. Also could you use the `-1` suffix convention instead of `1.7a`

kit-ty-kate:
> It should be clarified in opam which uses in some way both names: https://github.com/ocaml/opam/blob/master/src/state/opamFormatUpgrade.ml#L24

This is only the format upgrade module, some parts are not actually used by opam-depext. However, if it were to be used at some point this needs to match the `ID` field in `/etc/os-release` (if your distribution has it)

packages/coq/coq.8.8.2/opam
kit-ty-kate: I see. Thanks

packages/opus/opus.0.1.2/opam
bobot: It is true that depext doesn't detect this os. But there is code in opam (for upgrade) which take into account `nixos` (and prefers it compared to nix). Does people use `nixos` themselves in depext?

I can remove the nixos line, I have no opinion on the previous question.

Do we have a list of the nomenclature for `os-family`, `os`, `os-distribution`?

timbertson:
> It should be clarified in opam which uses in some way both names: https://github.com/ocaml/opam/blob/master/src/state/opamFormatUpgrade.ml#L24

Uh, that code is just wrong I'm afraid. `opam2nix` has _always_ used `nixpkgs`, never `nixos`. I don't know why this code exists, someone must have gotten confused.

packages/opus/opus.0.1.2/opam
kit-ty-kate: I clarify this in https://github.com/ocaml/opam-repository/pull/14477#issuecomment-510849631
You can see how opam detects distribution/os/os-family here: https://github.com/ocaml/opam/blob/c35eb977d19457d3f13b31c1aa60de4fb210865d/src/state/opamSysPoll.ml

What is the `ID` field in `/etc/os-release` in NixOS?

timbertson:
I'd like to find out where (if anywhere) `nixos` is actually used. Duplicating deps for both `nixos` and `nixpkgs` is just going to get confusing.

@AltGr , It looks like the first introduction of `nixos` was in https://github.com/ocaml/opam/commit/e236be540c0a747481d13ed13c62da5d83f71f04. I don't suppose you remember?

From the comments in the [related issue](https://github.com/ocaml/opam/issues/2919) "NixOs" was casually mentioned, I'm wondering if that accidental inaccuracy simply made its way into the code unchecked? Or do you know of a tool which actually uses "nixos"? (opam2nix doesn't, it only uses "nixpkgs")

timbertson:
I use fedora with nixpkgs, so opam's builtin depext is always going to do the wrong thing there. Is it ever going to support `nixos`? I guess doing `nix-env -ia <attr>` would be the closest analogy.

Maybe using "os" for depexts is just wrong in general, the information you really want is "which package sets are available". For most distros one implies the other, but not when third-party package managers come into it. I assume OSX/Windows would have the same kind of thing, since there are a number of package managers which might be present.

timbertson:
Sorry for stalling for so long on this. I have in the past added `nixpkgs` depexts in this repo, so I was under the impression there was now a mix of both `nixpkgs` and `nixos`, and I can't make `opam2nix` support both. But I've just re-checked, and it looks like the upgrade code has migrated all those to `nixos`, which makes them not work.

So since the current code doesn't have any effect since nobody's actually using `nixpkgs`, I'm going to just change opam2nix to look at `nixos` so that it works :man_shrugging:. @bobot , want to remove all the `nixpkgs` lines?

While we're at it, any thoughts on `os-family`? It's not currently set, maybe that's best for clarity because there aren't any other members of the family.

kit-ty-kate:
The CI was cancelled several times on this PR to avoid freezing the whole CI as it modifies a lot of packages.

Is this PR ready to be reviewed? Is there a consensus?

timbertson:
Agreed, as the source of contention for the distribution name (sorry 'bout that) I'm happy with this PR now :+1:

packages/conf-gcc/conf-gcc.1.0/opam
```diff
@@ -0,0 +1,21 @@
...
+license: "GPL-2+"
+build: [["gcc" "--version"]]
+depexts: [
+  ["gcc"] {os-distribution = "centos"}
+  ["gcc"] {os-distribution = "fedora"}
+  ["gcc"] {os-distribution = "opensuse"}
+  ["gcc"] {os-distribution = "debian"}
+  ["gcc"] {os-distribution = "ubuntu"}
+  ["gcc"] {os-distribution = "nixos"}
+  ["gcc"] {os-distribution = "archlinux"}
```
avsm:
```suggestion
  ["gcc"] {os-distribution = "arch"}
```

PR #14480, opened 2019-07-11
+bigstringaf.0.5.1

seliopou:


packages/bigstringaf/bigstringaf.0.5.1/opam
```diff
@@ -0,0 +1,42 @@
+opam-version: "2.0"
+maintainer: "Spiros Eliopoulos <spiros@inhabitedtype.com>"
+authors: [ "Spiros Eliopoulos <spiros@inhabitedtype.com>" ]
+license: "BSD-3-clause"
+homepage: "https://github.com/inhabitedtype/bigstringaf"
+bug-reports: "https://github.com/inhabitedtype/bigstringaf/issues"
+dev-repo: "git+https://github.com/inhabitedtype/bigstringaf.git"
+build: [
+  ["dune" "subst"] {pinned}
```
kit-ty-kate:
```suggestion
```
You don't seem to use any dune substitutions

packages/bigstringaf/bigstringaf.0.5.1/opam
```diff
@@ -0,0 +1,42 @@
...
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name] {with-test}
+]
+depends: [
+  "dune" {build}
```
kit-ty-kate:
```suggestion
  "dune" {build & >= "1.6"}
```
According to your `dune-package` file, this package requires dune >= 1.6

packages/bigstringaf/bigstringaf.0.5.1/opam
seliopou: Is there any harm in leaving these around?

packages/bigstringaf/bigstringaf.0.5.1/opam
seliopou: Fixed in inhabitedtype/bigstringaf#38. I don't know if dune does validation, but the build files certainly don't use any new features.

packages/bigstringaf/bigstringaf.0.5.1/opam
kit-ty-kate: no but it's just some more noise in the package description that could be avoided

packages/bigstringaf/bigstringaf.0.5.1/opam
kit-ty-kate: ok thanks! This PR will need to be updated though

PR #14483, opened 2019-07-12
Release clangml.4.0.0

thierry-martinez:
- C++ support: all the code examples from C++14 and C++17 norms are covered
  by the AST.

- AST pretty-printers, comparators and lifters are now defined in sub-libraries
  (clangml.show, clangml.ord, clangml.lift). Visitors are no longer derived by
  default (`override` PPX library can be used to derive visitors from the AST
  if needed.)

- A very limited C/C++ pretty-printer is provided in the sub-library
  clangml.printer. It should be extended in next releases.

- Compatible with OCaml 4.08

- Bug fixes

- Flag dune as a build dependency
  (suggested by @kit-ty-kate in
   https://github.com/ocaml/opam-repository/pull/13434)

### DIFF
--- a/packages/clangml/clangml.4.0.0/opam
+++ b/packages/clangml/clangml.4.0.0/opam
@@ -14,7 +14,7 @@ depends: [
   "conf-libclang"
   "conf-ncurses"
   "conf-zlib"
-  "ocamlfind" {>= "1.8.0"}
+  "ocamlfind" {build & >= "1.8.0"}
   "dune" {>= "1.10.0"}
   "ppxlib" {>= "0.6.0"}
   "stdcompat" {>= "10"}
--- a/packages/clangml/clangml.4.0.0/opam
+++ b/packages/clangml/clangml.4.0.0/opam
@@ -21,7 +21,7 @@ depends: [
   "ppx_show" {>= "0.1.0"}
   "ppx_compare" {>= "v0.12.0"}
   "override" {>= "0.1.0"}
-  "ocaml" {>= "4.04.0"}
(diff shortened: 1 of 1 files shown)


packages/clangml/clangml.4.0.0/opam
```diff
@@ -0,0 +1,32 @@
...
+dev-repo: "git+https://gitlab.inria.fr/memcad/clangml"
+synopsis: "OCaml bindings for Clang API"
+description: """
+clangml provides bindings to call the Clang API from OCaml.
+"""
+depends: [
+  "conf-libclang"
+  "conf-ncurses"
+  "conf-zlib"
+  "ocamlfind" {>= "1.8.0"}
```
kit-ty-kate:
```suggestion
  "ocamlfind" {build & >= "1.8.0"}
```
I took a look at your code and you're using ocamlfind only in `configure.ac` however I don't really see the point of using ocamlfind if you're using dune as build system. Maybe it can be removed from the dependencies in future versions?

thierry-martinez: Thanks for suggesting moving `ocamlfind` as build dependency. I use `ocamlfind` to check whether packages are available or not at `configure` time: even if it is not mandatory now (I could check nothing at `configure` time and fail at build time), such `configure` time checks could be useful in future version (for instance, to handle better optional dependencies). I agree that `dune` should replace most of the need of `configure`, but there are still things that I don't know how to ask `dune` to do (such as finding llvm/clang and gathering required C flags).

kit-ty-kate:
Thanks!

PR #14494, opened 2019-07-12
Fix/Add omd.2.0.0~alpha

kit-ty-kate:
This package was present in the source tree but not under the right name and was not available.

cc @pw374

PR #14496, opened 2019-07-12
[DO NOT MERGE YET] Fix/Add omd.2.0.0~alpha

kit-ty-kate:
Reopening #14494

This PR might cause opam update to fail.


pw374:
@kit-ty-kate I don't really know what's going on (not an opam expert). All I can say is I think omd 2 alpha being or not being in opam is not a big issue because afaik omd 2 is not release-ready yet.
Perhaps @nobj @clecat can provide better insight.

kit-ty-kate:
Ok this has been done and reverted already: https://github.com/ocaml/opam-repository/pull/13992
I think we can just remove this package since you were the one to add the package in the first place and are saying it's not release-ready yet.

PR #14501, opened 2019-07-13
bigstringaf.0.5.2: Untag dune a build dependency

kit-ty-kate:
https://github.com/ocaml/opam-repository/pull/14497 was merged in between https://github.com/ocaml/opam-repository/pull/14266 and was missing the fix

cc @seliopou could you change that upstream?

### DIFF
--- a/packages/bigstringaf/bigstringaf.0.5.2/opam
+++ b/packages/bigstringaf/bigstringaf.0.5.2/opam
@@ -11,7 +11,7 @@ build: [
   ["dune" "runtest" "-p" name] {with-test}
 ]
 depends: [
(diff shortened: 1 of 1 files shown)


seliopou:
Why is this change necessary? Dune is just a build dependency, after all.

kit-ty-kate:
The definition of a build dependency according to opam is that is does not need to rebuild the package if the dependency changed. Dune does not fit this definition anymore as it installs metadata files dependent of the current dune version. If one downgrade the dune version they can't read this metadata file anymore and will fail any builds requiring this package. (see pull requests linked above for more info)

PR #14509, opened 2019-07-15
Package herdtools7.7.54

maranget:
### `herdtools7.7.54`
The herdtools suite for simulating and studying weak memory models

---
* Homepage: http://diy.inria.fr/
* Source repo: git+https://github.com/herd/herdtools7.git
* Bug tracker: http://github.com/herd/herdtools7/issues/

---

### DIFF
--- a/packages/herdtools7/herdtools7.7.54/opam
+++ b/packages/herdtools7/herdtools7.7.54/opam
@@ -15,7 +15,7 @@ install: ["./dune-install.sh" "%{prefix}%"]
 # @todo Add "build-doc" field
 # @todo Add "build-test" field
 depends: [
-  "ocaml" {>= "4.02.0"}
+  "ocaml" {>= "4.05.0"}
   "dune"
 ]
 url {
--- a/packages/herdtools7/herdtools7.7.54/opam
+++ b/packages/herdtools7/herdtools7.7.54/opam
@@ -25,3 +25,4 @@ url {
     "sha512=63ebe937b8a651ec9918a8c4a70a41cba611fa8528a20b246732e0ec1bdb6e852e5936b8860f566d4973a9945cd44d1df6eb87c4e5c7704073565c07c8f75f20"
   ]
 }
+patches: ["chatouille.patch"]
--- a/packages/herdtools7/herdtools7.7.54/opam
+++ b/packages/herdtools7/herdtools7.7.54/opam
@@ -14,7 +14,6 @@ build: ["./dune-build.sh" "%{prefix}%"]
 install: ["./dune-install.sh" "%{prefix}%"]
 # @todo Add "build-doc" field
 # @todo Add "build-test" field
-remove: ["./dune-uninstall.sh" "%{prefix}%"]
 depends: [
   "ocaml" {>= "4.02.0"}
   "dune"
--- a/packages/herdtools7/herdtools7.7.54/opam
+++ b/packages/herdtools7/herdtools7.7.54/opam
@@ -17,7 +17,7 @@ install: ["./dune-install.sh" "%{prefix}%"]
 remove: ["./dune-uninstall.sh" "%{prefix}%"]
 depends: [
   "ocaml" {>= "4.02.0"}
-  "dune" {build}
+  "dune"
 ]
 url {
   src: "https://github.com/herd/herdtools7/archive/7.54.tar.gz"
--- a/packages/herdtools7/herdtools7.7.54/opam
+++ b/packages/herdtools7/herdtools7.7.54/opam
@@ -10,8 +10,8 @@ homepage: "http://diy.inria.fr/"
 bug-reports: "http://github.com/herd/herdtools7/issues/"
 doc: "http://diy.inria.fr/doc/index.html"
 dev-repo: "git+https://github.com/herd/herdtools7.git"
-build: ["./dune-build.sh" "%{prefix}%"]
-install: ["./dune-install.sh" "%{prefix}%"]
+build: ["sh -x ./dune-build.sh" "%{prefix}%"]
+install: ["sh -x ./dune-install.sh" "%{prefix}%"]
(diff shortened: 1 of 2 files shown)


packages/herdtools7/herdtools7.7.54/opam
```diff
@@ -0,0 +1,28 @@
...
+doc: "http://diy.inria.fr/doc/index.html"
+dev-repo: "git+https://github.com/herd/herdtools7.git"
+build: ["./dune-build.sh" "%{prefix}%"]
+install: ["./dune-install.sh" "%{prefix}%"]
+# @todo Add "build-doc" field
+# @todo Add "build-test" field
+remove: ["./dune-uninstall.sh" "%{prefix}%"]
+depends: [
+  "ocaml" {>= "4.02.0"}
+  "dune" {build}
```
kit-ty-kate:
```suggestion
  "dune"
```
Dune should not be flagged as a build dependency anymore (see #14266)

packages/herdtools7/herdtools7.7.54/opam
```diff
@@ -0,0 +1,28 @@
...
+authors: [
+  "Luc Maranget <Luc.Maranget@inria.fr>"
+  "Jade Alglave <j.alglave@ucl.ac.uk>"
+  "Vincent Jacques <vincent@russian-dolls-sunflowers.com>"
+]
+homepage: "http://diy.inria.fr/"
+bug-reports: "http://github.com/herd/herdtools7/issues/"
+doc: "http://diy.inria.fr/doc/index.html"
+dev-repo: "git+https://github.com/herd/herdtools7.git"
+build: ["./dune-build.sh" "%{prefix}%"]
```
kit-ty-kate: I looked at the code and I'm not too sure what all this shell boilerplate is about

packages/herdtools7/herdtools7.7.54/opam
```diff
@@ -0,0 +1,28 @@
...
+install: ["./dune-install.sh" "%{prefix}%"]
+# @todo Add "build-doc" field
+# @todo Add "build-test" field
+remove: ["./dune-uninstall.sh" "%{prefix}%"]
```
kit-ty-kate:
```suggestion
```
This in unnecessary since OPAM 2.0

packages/herdtools7/herdtools7.7.54/opam
kit-ty-kate: Also, I've tried your package locally and it doesn't work. It just installs empty directories

kit-ty-kate: I've tried to make it work using dune + a simple configure script to send a PR back to you but I'm not sure to understand what you wanted installed in the first place. What executables do you want to install? Is `herd` a library or an executable?

maranget: I am working on it. I do not understand what happens either.
To answer your question, the intend is to install several exectables (herd7,litmus7 etc.) in PREFIX/bin, as well as files in PREFIX/share.

kit-ty-kate: I'm pretty sure you can do all that with dune without any shell scripts

maranget: It may be the case that one can do all that with dune, but I do not know how to proceed...

maranget:
Hi, the package now has passed all checks. Would you consider merging in ?

--Luc

kit-ty-kate:
Ok fine. I tried to have everything using `dune` but I stumbled upon https://github.com/ocaml/dune/issues/1868 so I give up for now but once this is fixed. As I understand you want a way to also install it locally without opam so this requires the dune issue to be fixed to get rid of every shell scripts. I might have a partial solution in the meantime but for now this is more or less ok.

Thanks.

maranget:
Hi Kate,

Installing the contents of various repositories <d>/libdir into PREFIX/share/herdtools7/<d> is not the only difficulty I ran across. I could not find a simple way to instruct dune to install various executables herd.exe, litmus.exe etc. as PREFIX/bin/herd7, PREFIX/bin/litmus7, etc. Hence the shell scripts

kit-ty-kate:
> Hi Kate,
> 
> Installing the contents of various repositories /libdir into PREFIX/share/herdtools7/ is not the only difficulty I ran across. I could not find a simple way to instruct dune to install various executables herd.exe, litmus.exe etc. as PREFIX/bin/herd7, PREFIX/bin/litmus7, etc. Hence the shell scripts

That is easily fixed with `(public_name)`. For instance:
```
(executable
  (name litmus)
  (public_name litmus7))
```

kit-ty-kate:
Moreover, without `(public_name)` nothing gets installed nor built automatically by dune.

maranget:
> > Hi Kate,
> > 
> > Installing the contents of various repositories /libdir into PREFIX/share/herdtools7/ is not the only difficulty I ran across. I could not find a simple way to instruct dune to install various executables herd.exe, litmus.exe etc. as PREFIX/bin/herd7, PREFIX/bin/litmus7, etc. Hence the shell scripts
> 
> That is easily fixed with `(public_name)`. For instance:
> ```
> (executable
>   (name litmus)
>   (public_name litmus7))
> ```
> 
Thanks, however there are tens for executable to install in directory tools
and I am using (executables ...). I guess I could replace this by
tens of (executable ...) stanzas.

kit-ty-kate:
If you use `(names)` you can use its counterpart `(public_names)`

kit-ty-kate:
It just follows the same order. The whole thing then would be:
```
(executables
  (names litmus herd)
  (public_names litmus7 herd7))
```

maranget:
> It just follows the same order. The whole thing then would be:
> ```
> (executables
>   (names litmus herd)
>   (public_names litmus7 herd7))
> ```
Thanks, I'll try this for the next release.

PR #14508, opened 2019-07-15
Package ssl.0.5.9

smimram:
### `ssl.0.5.9`
Bindings for OpenSSL

---
* Homepage: https://github.com/savonet/ocaml-ssl
* Source repo: git+https://github.com/savonet/ocaml-ssl.git
* Bug tracker: https://github.com/savonet/ocaml-ssl/issues

---
:camel: Pull-request generated by opam-publish v2.0.0

PR #14511, opened 2019-07-15
ssl.0.5.9: Unflag dune as a build dependency

kit-ty-kate:
Fixing https://github.com/ocaml/opam-repository/pull/14508

cc @smimram Dune should not be flagged as a build dependency anymore (see #14266)
cc @avsm don't forget to check for build flags on dune


smimram:
Thanks I did not know about that. It's also fixed in my opam for next releases.

Issue #287, opened 2013-01-11
Would be nice to have a package for ocamlbrowser


UnixJunkie:
This thing comes quite handy when inspecting source code
from the standard library.

samoht:
Please see with upstream.

PR #14516, opened 2019-07-15
+conduit.1.5.0

avsm:
* lwt-unix: Do not close file descriptors more than once, which led to a lot of
  log spam due to EBADF (#294 @hcarty @avsm)
* lwt-unix: Always close channels after handling an event (#283 @hcarty)
* Allow TCP to be established from existing file descriptors
  (for example, an inherited systemd socket) (#144 @SGrondin #282 @timbertson)
* async: add `Conduit_async.V3` which provides convenience functions for
  resolving URIs to addresses (#287 @vbmithr)
* `Lwt_ssl`: Enable certification validation (#291 @vouillon)
* `Async_ssl`: fix exception raised when other side disconnects
  due to sharing underlying fd (#295 @bogdan2412)

### DIFF
--- a/packages/conduit-async/conduit-async.1.5.0/opam
+++ b/packages/conduit-async/conduit-async.1.5.0/opam
@@ -1,0 +1,37 @@
+opam-version: "2.0"
+maintainer: "anil@recoil.org"
+authors: [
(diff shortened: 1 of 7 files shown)


samoht:
```

PR #14519, opened 2019-07-15
Add ocaml-variants.4.08.1+trunk

kit-ty-kate:
This PR adds the `ocaml-variants.4.08.1+trunk+*` packages, as well as the `ocaml.4.08.1` virtual package.
This allows to test for the next maintenance release of OCaml 4.08 which amongst other things allows `nocrypto` to be built with OCaml 4.08 (as it doesn't compile with `ocaml.4.08.0`)

cc @damiendoligez

### DIFF
--- a/packages/ocaml-variants/ocaml-variants.4.08.1+trunk+afl/opam
+++ b/packages/ocaml-variants/ocaml-variants.4.08.1+trunk+afl/opam
@@ -1,0 +1,39 @@
+opam-version: "2.0"
+synopsis: "latest 4.08 development, with afl-fuzz instrumentation"
+maintainer: "platform@lists.ocaml.org"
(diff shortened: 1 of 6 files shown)


kit-ty-kate:
@damiendoligez could you confirm there is going to be a maintenance release of OCaml 4.08 at some point in the future? I would rather not mess up and add a virtual ocaml package that is not going to be released.

damiendoligez:
I didn't want to do a 4.08.1 release because we have already launched the release cycle for 4.09, but the `nocrypto` problem looks big enough to justify a bugfix release. I'll try to do the release this week.

avsm:
much appreciated; the nocrypto issue is blocking most http-related packages in opam on 4.08.0

PR #14522, opened 2019-07-16
bitstring: make ocaml migrate parsetree a runtime dependency

jorisgio:
OMP has a runtime lib, and upgrading omp breaks ppx_bitstring as it is not rebuilt to match interface change.


kit-ty-kate:
I'm sorry but I can't see the official `3.1.1` release. Furthermore you still need to modify the old versions for this be correct. This is a fix to a package, not a release.

PR #14524, opened 2019-07-16
Flac conflicts with old Liquidsoap versions.

smimram:
An user of Liquidsoap was recently hit by this incompatibility of latest version of flac with old Liquidsoap versions: https://github.com/savonet/liquidsoap/issues/881.


kit-ty-kate:
Is this also incompatible for the previous versions of `flac` available on opam? If yes, could you also add a constraint for them?

smimram:
No, the versions of ocaml-flac prior 0.1.5 are fine.

kit-ty-kate:
Perfect. Thank you!

PR #14526, opened 2019-07-16
Genprint 0.1

progman1:
library for printing of values a la toplevel but within compilation units.


progman1:
I put in the url - that was the cause of the problem before.
when is an archive release tarball made? there was one there but I did not invoke it manually
but then again it didn't update automatically after I pushed the opam file either...

progman1:
the url is in with an updated checksum
what's up now?!

packages/genprint/genprint.0.1/opam
```diff
@@ -0,0 +1,29 @@
...
+
+depends: [
+  "ocaml" {build}
+  "dune"
+  "ppxlib"
+]
+
+url {
+  src: "https://github.com/progman1/genprintlib/archive/0.1.tar.gz"
+  checksum: "4c7518439c1c4a4d6679190661e84bd0b6b21391"
```
mseri: This is not a sha256 or and md5. It should be `"md5=176d76e7b541d535f1aee7e194971d0b"` and/or `"sha256=6318065feb4ac2d5000d30e1fced95c9f2987185f7d889ad8c710d03b4765a47"`. How did you get that value?

packages/genprint/genprint.0.1/opam
```diff
@@ -0,0 +1,29 @@
...
+bug-reports: "https://github.com/progman1/genprintlib/issues"
+dev-repo: "git+https://github.com/progman1/genprintlib.git"
+
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+]
+
+depends: [
+  "ocaml" {build}
```
mseri:
```suggestion
  "ocaml"
```

packages/genprint/genprint.0.1/opam
```diff
@@ -0,0 +1,29 @@
+opam-version: "2.0"
+version: "0.1"
```
mseri:
```suggestion
version: "0.1"
```
This is inferred by opam from the folder name

packages/genprint/genprint.0.1/opam
mseri:
```suggestion
  checksum: "md5=176d76e7b541d535f1aee7e194971d0b"
```

progman1:
I got the checksum after generating a new release tarball - what I assumed was the checksum was actually the commit hash.
surely I don't have to download the tarball and gen a checksum manually?
the use of 'name: genprint' is redundant I am told. 
it was surprising to me that the name is inferred and then used in  'dune -p name' - this seems 
a crossover from symbolic labelling to the module namespace (equals confusion).
if there has to be inference of the symbolic name (ignoring the explicit 'name: ...'  form) how best to make the separation ?

mseri:
I'll try to replu to all your questions.

That's correct. If you publish using `opam-publish` or `dune-release`, they will generate, download and checksum the tarball for you, otherwise you need to do it by hand

Both `name` and `version` are redundant, and are inferred by opam from the path in the opam repository. They can be used as variables in the opam file and will be evaluated to the inferred values. I don't understand what separation you mean. You could use a different name for the library, and then explicitly write `"-p" "thenameyouwant"` but it is not considered good practice. Dune can deal with multiple opam files in the same repository without issue (see e.g. https://github.com/mirage/ocaml-cohttp)

progman1:
this error message:
error 3: File format error in 'url.checksum' at line 28, column 53: while expecting checksum: (Invalid_argument OpamHash.of_string)

did I have access to it?
I only saw a generic error...

mseri:
I think you may do if you run `opam lint` on the generated opam file

progman1:
still not going through with a pass from the linter

mseri:
Looks like the wrong tarball. Looking at your repo, it seems that the correct release is tagged "v0.1" and not "0.1"

mseri:
Now the error is probably related to ocaml 4.08:

```
#          Outcometree.out_name
#     ocamlopt src/.genprint.objs/native/genprint__Genprintval.{cmx,o} (exit 2)
# (cd _build/default && /home/opam/.opam/4.08/bin/ocamlopt.opt -w -40 -g -I src/.genprint.objs/byte -I src/.genprint.objs/native -I /home/opam/.opam/4.08/lib/ocaml/compiler-libs -intf-suffix .ml -no-alias-deps -open Genprint__ -o src/.genprint.objs/native/genprint__Genprintval.cmx -c -impl src/genprintval.ml)
# File "src/genprintval.ml", line 105, characters 44-47:
# 105 |             list := Oval_constr (Oide_ident "_", []) :: !list
#                                                   ^^^
#          Outcometree.out_name
```

progman1:
opam switch list-available doesn't even have an official 4.08 release!
yet there are 'official' releases of earlier versions and I know 4.08 has been announced on discuss.
assuming that the ocaml team have there own repo, is the practice to add that repo to one's opam?

progman1:
opam switch list-available doesn't even have an official 4.08 release!
yet there are 'official' releases of earlier versions and I know 4.08 has been announced on discuss.
quite confusing...
is the practice to add the ocaml repo directly to one's opam?

mseri:
You probably need to update your opam cache (run `opam update`), as `4.08.0` as been released already since some time.

mseri:
Usually tweaks on the version bounds or the structure of the repository are needed only on the first submission. The common practice is to use the opam docker images pinning the package and testing the build on travis automatically.

Of course for the release, you need to create the tarball and get its checksum, but usually that is the only thing, and maybe add some version bounds coming out of the release tests, but these don't require to regenerate the tarball as the opam file in the repository is the one that will be used by opam, so one can just update the master version of the opam file and move on

progman1:
did you push a change to my repo?
I am not familiar enough with this stuff to handle that - before I saw it I noticed I had problems pushing
my change and forced it through. 
and now the CI is not kicking off automagically. I will put 4.06 &  < 4.08 in manually and resubmit but how to get CI going again without doing another PR?
thanks for your help. this byzantine system is difficult but amazing!

progman1:
it looks like I was wrong - the CI progress below has come back from the dead!

packages/genprint/genprint.0.1/opam
```diff
@@ -0,0 +1,28 @@
+opam-version: "2.0"
+synopsis: "PPX syntax extension and library package for printing values of any type"
+description: """
+A PPX syntax extension and library package enabling printing of values of any type using OCaml's internal printing facilities a la toplevel evaluation.
+Useful for debugging as a quick alternative to ocamldebug/ppx_deriving/#install_printer.
+"""
+
+maintainer: "progman1"
```
kit-ty-kate: Could you provide an email address?

progman1:
and invite being spammed? 
I thought everything that needed to be communicated could be done via 'issues', no?

kit-ty-kate:
> and invite being spammed?
> I thought everything that needed to be communicated could be done via 'issues', no?

There might be matters that need a more direct way to discuss. Issues tend to be let to perish.
If you're only worried about being spam, there are ways to counteract spam (`name <at> blah.com` is one for example, but there are several other ways)

kit-ty-kate:
> There might be matters that need a more direct way to discuss. Issues tend to be let to perish.

Also some people might not have any github accounts for a reason or another.

progman1:
is it ok to access the source of a dependency as a step in building ? ie.

cp ~/.opam/$OPAMSWITCH/.opam-switch/sources/ocaml-base-compiler.$OPAMSWITCH/....

mseri:
No, the sandbox will prevent you to do that. Why do you need that?

progman1:
something amiss here with dune on 1.10 ....

# Error: 'allow_approximate_merlin' is only available since version 1.9 of the dune language

<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
| - build genprint 0.1
+- 
+- The following changes have been performed
| - install base                    v0.12.2
| - install dune                    1.10.0

progman1:
for 4.06: Error: Unbound module Stdlib

not much info about Stdlib-shims on the homepage but I put it in the deps and it hasn't worked...

packages/genprint/genprint.0.1/opam
```diff
@@ -0,0 +1,35 @@
...
+authors: [ "progman1" ]
+homepage: "https://github.com/progman1/genprintlib"
+bug-reports: "https://github.com/progman1/genprintlib/issues"
+dev-repo: "git+https://github.com/progman1/genprintlib.git"
+
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build"  "@epoc1"] {ocaml:version  <"4.08.0"}
+  ["dune" "build"  "@epoc2"] {ocaml:version >="4.08.0"}
+  ["dune" "build" "-p" "genprint" "-j" jobs]
```
kit-ty-kate:
```suggestion
  ["dune" "build" "-p" name "-j" jobs]
```

packages/genprint/genprint.0.1/opam
```diff
@@ -0,0 +1,35 @@
...
+]
+
+depends: [
+#  "ocaml" {>= "4.06.0" & < "4.08.0" }
```
kit-ty-kate:
```suggestion
```

packages/genprint/genprint.0.1/opam
```diff
@@ -0,0 +1,35 @@
...
+maintainer: "progman1"
+authors: [ "progman1" ]
+homepage: "https://github.com/progman1/genprintlib"
+bug-reports: "https://github.com/progman1/genprintlib/issues"
+dev-repo: "git+https://github.com/progman1/genprintlib.git"
+
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build"  "@epoc1"] {ocaml:version  <"4.08.0"}
+  ["dune" "build"  "@epoc2"] {ocaml:version >="4.08.0"}
```
kit-ty-kate: those two call to dune are only advised if you're developing, not for releases.
I looked at the code and it would be much more simple to just have a rule that generate the `genprintval.ml` from a generator program (in OCaml and thus would have access to the version number)

progman1:
> it would be much more simple to just have a rule that generate the genprintval.ml from a generator program 

can you point me to a simple example? - I'm unfamiliar with all this - it's been a struggle to get this far.

progman1:
I mean it's been a struggle to get this far and the prospect of grubbing around for inspiration in the obscureness of the dune documentation is very unappealing.

progman1:
what does the final stage of camelus actually do?

progman1:
happy to replace the dune calls but why is distinction made between dev and release exactly?

kit-ty-kate:
Should be fixed by this PR https://github.com/progman1/genprintlib/pull/1

kit-ty-kate:
I'm sorry why did you not simply merge my PR?! It would have been much simpler than just fixing mistakes over and over

progman1:
I wanted a bit of involvement .. to burn it all into my head!
thanks for the code snippet - it was enormously helpful and saved me from any more thrashing (though I just did a bit more with the CI I believe!)

progman1:
how can it still be pending? have I been rate-limited?

progman1:
what do I have to do?!

progman1:
it would be helpful to know what I left out in order for it to auto-merge (the status message implied the 
possibility).

also, why did it take two days (late friday to sunday afternoon) to pass the unspecified camelus stage?
thanks.

kit-ty-kate:
You can't merge this yourself, only opam-repository maintainers can. Camelus was being very slow, it is being investigated.

PR #14532, opened 2019-07-17
add 4.08.1+default-unsafe-string variant

avsm:


kit-ty-kate:
sorry I thought this was removed from the compiler

PR #14536, opened 2019-07-17
Fix hdf5.0.1.5

vbrankov:
hdf5.0.1.5 is was still broken: I tried installing it and although the install doesn't raise any errors, I installs nothing.  I'm trying a new attempt at fixing it by renaming the library to hdf5 so that the package name and the library name are the same.

See:
https://github.com/ocaml/opam-repository/pull/14417
https://github.com/ocaml/opam-repository/pull/14356


mseri:
Thanks. Let's hope this will do it

Issue #12989, opened 2018-11-16
conf-blas fails to install under Manjaro Linux 18


quipa:
When installing conf-blas under Manjaro Linux 18.0.0 x86_64 I get the error below, it seems to be related to [this test](https://github.com/ocaml/opam-repository/blob/master/packages/conf-openblas/conf-openblas.0.2.0/files/test.c). I am not familiar with cblas or lapacke, so no idea what's going on..!
```bash
$ opam install owl
The following actions will be performed:
  - install conf-openblas 0.2.0 [required by owl]
  - install owl           0.4.0
===== 2 to install =====
Do you want to continue? [Y/n] y

<><> Gathering sources ><><><><><><><><><><><><><><><><><><><><><><><><><><><>
[owl.0.4.0] found in cache

<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><>
[ERROR] The compilation of conf-openblas failed at
        "/home/possum/.opam/opam-init/hooks/sandbox.sh build sh -exc cc
        $CFLAGS test.c -lopenblas".

#=== ERROR while compiling conf-openblas.0.2.0 ==============================#
# context     2.0.1 | linux/x86_64 | base-bigarray.base base-threads.base base-unix.base ocaml-system.4.07.0 | https://opam.ocaml.org#0c9ec43d
# path        ~/.opam/system/.opam-switch/build/conf-openblas.0.2.0
# command     ~/.opam/opam-init/hooks/sandbox.sh build sh -exc cc $CFLAGS test.c -lopenblas
# exit-code   1
# env-file    ~/.opam/log/conf-openblas-17857-05296d.env
# output-file ~/.opam/log/conf-openblas-17857-05296d.out
### output ###
# + cc test.c -lopenblas
# /usr/bin/ld: /tmp/ccFJ2mRN.o: in function `main':
# test.c:(.text+0x63): undefined reference to `cblas_dnrm2'
# collect2: error: ld returned 1 exit status



<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+- The following actions failed
| - build conf-openblas 0.2.0
+- 
- No changes have been performed```

sconeyard:
Same on a fresh install. Anyone got a hint?

mseri:
It probably uses the wrong flags. What do `pkg-config --libs openblas` and `pkg-config --cflags openblas` return?

sconeyard:
libs returns > -lopenblas 
[...]

PR #14541, opened 2019-07-17
conf-openblas: add Arch Linux support

mreppen:
Fixes https://github.com/ocaml/opam-repository/issues/12989 for `os-distribution = "arch"`

This is the cause: https://bugs.archlinux.org/task/59046


mseri:
Thanks

PR #14543, opened 2019-07-18
move 4.08.1 default-unsafe-string to the right +trunk prefix

avsm:


avsm:
doing a quick merge to unblock container builds

PR #14545, opened 2019-07-18
[new release] ppxlib (0.8.1)

NathanReb:
Base library and tools for ppx rewriters

- Project page: <a href="https://github.com/ocaml-ppx/ppxlib">https://github.com/ocaml-ppx/ppxlib</a>
- Documentation: <a href="https://ocaml-ppx.github.io/ppxlib/">https://ocaml-ppx.github.io/ppxlib/</a>

### Fixed

- Report errors according to the value of `OCAML_ERROR_STYLE` and
  `OCAML_COLOR` in the standalone driver (ocaml-ppx/ppxlib#83, @NathanReb)

### DIFF
--- a/packages/ppxlib/ppxlib.0.8.1/opam
+++ b/packages/ppxlib/ppxlib.0.8.1/opam
@@ -11,7 +11,7 @@ build: [
   ["dune" "build" "-p" name "-j" jobs]
 ]
 run-test: [
-  ["dune" "runtest" "-p" name "-j" jobs] { ocaml:version >= "4.06" }
+  ["dune" "runtest" "-p" name "-j" jobs] { ocaml:version >= "4.06" & ocaml:version < "4.08" }
 ]
 depends: [
   "ocaml"                   {>= "4.04.1"}
--- a/packages/ppxlib/ppxlib.0.8.1/opam
+++ b/packages/ppxlib/ppxlib.0.8.1/opam
@@ -1,0 +1,45 @@
+opam-version: "2.0"
+maintainer: "opensource@janestreet.com"
+authors: ["Jane Street Group, LLC <opensource@janestreet.com>"]
+homepage: "https://github.com/ocaml-ppx/ppxlib"
+bug-reports: "https://github.com/ocaml-ppx/ppxlib/issues"
+dev-repo: "git+https://github.com/ocaml-ppx/ppxlib.git"
+doc: "https://ocaml-ppx.github.io/ppxlib/"
+license: "MIT"
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+]
+run-test: [
+  ["dune" "runtest" "-p" name "-j" jobs] { ocaml:version >= "4.06" }
+]
+depends: [
+  "ocaml"                   {>= "4.04.1"}
+  "base"                    {>= "v0.11.0"}
+  "dune"
+  "ocaml-compiler-libs"     {>= "v0.11.0"}
+  "ocaml-migrate-parsetree" {>= "1.3.1"}
+  "ppx_derivers"            {>= "1.0"}
+  "stdio"                   {>= "v0.11.0"}
+  "ocamlfind"               {with-test}
+]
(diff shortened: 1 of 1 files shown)


NathanReb:
The revdep failure with `cconv-ppx` seems to be unrelated to this release as it appears the package tests are broken on `4.08` with older versions of `ppxlib` as well.

packages/ppxlib/ppxlib.0.8.1/opam
```diff
@@ -0,0 +1,45 @@
...
+bug-reports: "https://github.com/ocaml-ppx/ppxlib/issues"
+dev-repo: "git+https://github.com/ocaml-ppx/ppxlib.git"
+doc: "https://ocaml-ppx.github.io/ppxlib/"
+license: "MIT"
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+]
+run-test: [
+  ["dune" "runtest" "-p" name "-j" jobs] { ocaml >= "4.06" }
```
kit-ty-kate:
```suggestion
  ["dune" "runtest" "-p" name "-j" jobs] { ocaml:version >= "4.06" }
```

kit-ty-kate: `ocaml >= "4.06"` is always false otherwise

packages/ppxlib/ppxlib.0.8.1/opam
```diff
@@ -0,0 +1,45 @@
...
+]
+depends: [
+  "ocaml"                   {>= "4.04.1"}
+  "base"                    {>= "v0.11.0"}
+  "dune"                    {build}
```
kit-ty-kate:
```suggestion
  "dune"
```
Dune should not be flagged as a build dependency anymore (see #14266)

kit-ty-kate:
> The revdep failure with `cconv-ppx` seems to be unrelated to this release as it appears the package tests are broken on `4.08` with older versions of `ppxlib` as well.

It is indeed unrelated. I've seen this error before

NathanReb:
Looks like the test don't work on 4.08, I'll disable them!

XVilka:
There are a lot of errors on 4.06 as well.

NathanReb:
I'm a bit puzzled about the 4.06 failures. I tried it locally on a fresh `4.06.1` switch and the tests passed.

NathanReb:
I'm guessing it's due to the `INSTALL_LOCAL`, I'll take a look but I'm wondering if it's worth holding up the release for that.

NathanReb:
Gentle ping @kit-ty-kate, is this an absolute no go or do you think that's okay?

Is there by any chance a way to disable test for a system compiler switch? Couldn't find one in the doc.

kit-ty-kate:
I think that should ok but do you know why it's failing on the system switch? I'm running Camelus manually and it's good to go otherwise

hannesm:
FWIW AFAICT ppxlib 0.8.0 compiled and tested ok on system switch (but I don't know anything about ppxlib -- it may as well be that there was a less extensive test suite back then) https://travis-ci.org/ocaml/opam-repository/builds/537812691?utm_source=github_status&utm_medium=notification

kit-ty-kate:
Thanks!

NathanReb:
The 4.06 system compiler build for `0.8.0` was successful because the tests weren't run since the ocaml version constraint was incorrect.

By fixing it in this PR we re-enabled them and therefore found out they weren't working with a system compiler switch.

We have our own `ppx_expect` in ppxlib which we use for testing since. I suspect there's an issue in the way it's set up. The issue seems to be tied to `ocamlfind`/`findlib` somehow as the `#use "topfind";;` statement isn't working as intended and the subsequent `#require` trigger errors in the toplevel.

I'm not yet super familiar with these parts but I'm going to have to look into `findlib`'s internal for another project soon and I hope to kill two birds with one stone (I love birds so I wish to emphasize that no actual bird will be harmed in the process) and figure this out.

hannesm:
@NathanReb thanks for your investigation! :) please excuse my noise above then

PR #14549, opened 2019-07-18
QOcamlBrowser 0.2.11 for OCaml >= 4.08

Kakadu:
it was lagging behind.


kit-ty-kate:
The `remove` field is unnecessary since opam 2.0

kit-ty-kate:
Could you return the fix to the upstream opam file?

Kakadu:
> Could you return the fix to the upstream opam file?

Is it so important (changing opam file in project's repo)?

kit-ty-kate:
> Is it so important (changing opam file in project's repo)?

yes, otherwise you'll have to do the same fixes next time (also you don't have to change the tag/archive or anything, it's just for the next release)

mseri:
Thanks

PR #14556, opened 2019-07-19
Package mccs.1.1+10

AltGr:
### `mccs.1.1+10`
MCCS (which stands for Multi Criteria CUDF Solver) is a CUDF problem solver
developed at UNS during the European MANCOOSI project

---
* Homepage: http://www.i3s.unice.fr/~cpjm/misc/
* Source repo: git+https://github.com/AltGr/ocaml-mccs.git
* Bug tracker: https://github.com/AltGr/ocaml-mccs/issues

---

### DIFF
--- a/packages/mccs/mccs.1.1+10/opam
+++ b/packages/mccs/mccs.1.1+10/opam
@@ -1,0 +1,29 @@
+opam-version: "2.0"
+maintainer: "Louis Gesbert <louis.gesbert@ocamlpro.com>"
+authors: [
+  "Claude Michel <claude.michel@unice.fr>"
+  "Louis Gesbert <louis.gesbert@ocamlpro.com>"
+]
+homepage: "http://www.i3s.unice.fr/~cpjm/misc/"
+bug-reports: "https://github.com/AltGr/ocaml-mccs/issues"
+license: "LGPL-2.1 with OCaml linking exception, BSD-3-clause, GPL-3.0"
+dev-repo: "git+https://github.com/AltGr/ocaml-mccs.git"
+build: [
+  ["dune" "build" "-p" name]
+  ["sh" "-c" "dune build @settests --auto-promote || true"] {with-test}
+  ["dune" "runtest"] {with-test}
+]
+depends: [
+  "ocaml"
+  "dune" {>= "1.0"}
+  "cudf" {>= "0.7"}
+]
+synopsis: "MCCS (which stands for Multi Criteria CUDF Solver) is a CUDF problem solver
+developed at UNS during the European MANCOOSI project"
+url {
+  src: "https://github.com/AltGr/ocaml-mccs/archive/1.1+10.tar.gz"
+  checksum: [
+    "md5=21fa1652179d47baebd2e3f3cfdbf1d7"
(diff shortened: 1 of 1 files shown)


packages/mccs/mccs.1.1+10/opam
```diff
@@ -0,0 +1,29 @@
...
+  "Louis Gesbert <louis.gesbert@ocamlpro.com>"
+]
+homepage: "http://www.i3s.unice.fr/~cpjm/misc/"
+bug-reports: "https://github.com/AltGr/ocaml-mccs/issues"
+license: "LGPL-2.1 with OCaml linking exception, BSD-3-clause, GPL-3.0"
+dev-repo: "git+https://github.com/AltGr/ocaml-mccs.git"
+build: [
+  ["dune" "build" "-p" name]
+  ["sh" "-c" "dune build @settests --auto-promote || true"] {with-test}
+  ["dune" "runtest"] {with-test}
```
kit-ty-kate:
```suggestion
  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
```
Wouldn't it be better to use at least `-p` here?

packages/mccs/mccs.1.1+10/opam
```diff
@@ -0,0 +1,29 @@
...
+authors: [
+  "Claude Michel <claude.michel@unice.fr>"
+  "Louis Gesbert <louis.gesbert@ocamlpro.com>"
+]
+homepage: "http://www.i3s.unice.fr/~cpjm/misc/"
+bug-reports: "https://github.com/AltGr/ocaml-mccs/issues"
+license: "LGPL-2.1 with OCaml linking exception, BSD-3-clause, GPL-3.0"
+dev-repo: "git+https://github.com/AltGr/ocaml-mccs.git"
+build: [
+  ["dune" "build" "-p" name]
```
kit-ty-kate:
```suggestion
  ["dune" "build" "-p" name "-j" jobs]
```

packages/mccs/mccs.1.1+10/opam
```diff
@@ -0,0 +1,29 @@
...
+  ["sh" "-c" "dune build @settests --auto-promote || true"] {with-test}
```
kit-ty-kate: Shouldn't we need a `--profile` here?

AltGr: Forwarding to @dra27 :grin:

packages/mccs/mccs.1.1+10/opam
dra27: Agreed!

packages/mccs/mccs.1.1+10/opam
dra27: It shouldn't make a difference (at some point I should switch the tests to use the newer promotion system, but that's a battle for another day) as all this does is run an OCaml script, but it would do harm to have `--profile=release`, I agree

packages/mccs/mccs.1.1+10/opam
dra27: `--profile=release` is probably a good idea, yes. Not `-p`, though - or does that automatically select sub-packages too (there's a PR in the pipeline for ocaml-mccs to use virtual packages for the linking methods of the solver backend)

kit-ty-kate:
Thanks!

PR #14560, opened 2019-07-19
Add ctypes 0.15.0

yallop:
Closes ocamllabs/ocaml-ctypes#605

### DIFF
--- a/packages/ctypes/ctypes.0.15.0/opam
+++ b/packages/ctypes/ctypes.0.15.0/opam
@@ -1,0 +1,62 @@
+opam-version: "2.0"
+maintainer: "yallop@gmail.com"
+homepage: "https://github.com/ocamllabs/ocaml-ctypes"
+doc: "http://ocamllabs.github.io/ocaml-ctypes"
+dev-repo: "git+http://github.com/ocamllabs/ocaml-ctypes.git"
+bug-reports: "http://github.com/ocamllabs/ocaml-ctypes/issues"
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
There is an issue in `pkcs11 < 0.18.0`:
```
# ocamlfind: Package `ctypes.foreign.unthreaded' not found - required by `ctypes.foreign'
```
Do you know where the problem is?

kit-ty-kate:
(also check.ocamllabs.io says there weren't failing before)

kit-ty-kate:
`zbar.0.9` also has the same issue

kit-ty-kate:
Otherwise the rest of the dependencies seem to be ok

yallop:
I think the problem is that `pkcs11` [has the following optional dependencies](https://github.com/ocaml/opam-repository/blob/aa6c2114cecc1d52cdc6d01d59eb81e21573d9e2/packages/pkcs11/pkcs11.0.17.1/opam#L40-L41):

```
"ctypes"
"ctypes-foreign"
```

but it actually requires exactly both or neither of `ctypes` and `ctypes-foreign` to be installed.  If `ctypes` is installed but `ctypes-foreign` is not then `pkcs11` will try and fail to build the optional driver component.

yallop:
The issue with `zbar` appears to be similar.

kit-ty-kate:
Fair enough then. Thanks!

PR #14562, opened 2019-07-20
[new release] cohttp-lwt-jsoo, cohttp, cohttp-lwt, cohttp-lwt-unix, cohttp-top, cohttp-async and cohttp-mirage (2.2.0)

avsm:
CoHTTP implementation for the Js_of_ocaml JavaScript compiler

- Project page: <a href="https://github.com/mirage/ocaml-cohttp">https://github.com/mirage/ocaml-cohttp</a>
- Documentation: <a href="https://mirage.github.io/ocaml-cohttp/">https://mirage.github.io/ocaml-cohttp/</a>

##### CHANGES:

- Previously if the client closed the connection while cohttp was
  handling a request, the server would crash (by default, unless the
  user overrode that using `on_exn` or changing Lwt's async exception
  handler). Now, cohttp will just log this at `info` level and
  continue. Exceptions produced by user code are logged as errors,
  while other exceptions generated by cohttp call back to the conduit
  exception handler, as before. (mirage/ocaml-cohttp#669 @talex5)

### DIFF
--- a/packages/cohttp-async/cohttp-async.2.2.0/opam
+++ b/packages/cohttp-async/cohttp-async.2.2.0/opam
@@ -24,13 +24,13 @@ doc: "https://mirage.github.io/ocaml-cohttp/"
 bug-reports: "https://github.com/mirage/ocaml-cohttp/issues"
 depends: [
   "ocaml" {>= "4.04.1"}
-  "dune" {build & >= "1.1.0"}
+  "dune" {>= "1.1.0"}
   "async_kernel" {>= "v0.11.0"}
   "async_unix" {>= "v0.11.0"}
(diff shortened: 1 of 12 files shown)


kit-ty-kate:
Older versions of the Cohttp packages need updated constraints

kit-ty-kate:
In my opinion it's a bit of a sloppy fix but it's a fix nonetheless. Something tells me it would be slightly better if every cohttp packages had a fixed constraint `"cohttp" {= version}` but not all packages have a matching version, also I was not sure if people were actually using different version of cohttp with subpackages of totally different versions, it allows more combinations but it's also more tedious to work with

kit-ty-kate:
Should be alright now. Thanks!

PR #14568, opened 2019-07-22
Fix dune version for sedlex.2.1

toots:
This is the same fix as ocaml-community/sedlex#82. The `dune` files for this package now use `promote-until-clean` which was only introduced starting with dune 1.8 and seems to not be handled correctly with earlier versions.

### DIFF
--- a/packages/sedlex/sedlex.2.1/opam
+++ b/packages/sedlex/sedlex.2.1/opam
@@ -23,7 +23,7 @@ build: [
 ]
 depends: [
   "ocaml" {>= "4.02.3"}
-  "dune" {>= "1.0"}
+  "dune" {>= "1.8"}
   "ppx_tools_versioned"
   "ocaml-migrate-parsetree"
   "gen"


packages/sedlex/sedlex.2.1/opam
```diff
@@ -23,12 +23,15 @@ build: [
...
 depends: [
   "ocaml" {>= "4.02.3"}
-  "dune" {>= "1.0"}
+  "dune" {>= "1.8"}
   "ppx_tools_versioned"
   "ocaml-migrate-parsetree"
   "gen"
   "uchar"
 ]
+conflicts: [
```
kit-ty-kate: why does this introduce two unrelated conflicts?

kit-ty-kate: I would rather see those be modified in their own packages rather than this one if it's just about dune

toots: Ok!

toots:
@kit-ty-kate Changes done. The `wcs-lib` packages are still failing to build, not sure if that's related to `sedlex`. Let me know how you would like to proceed.

kit-ty-kate:
I'm not sure what has changed but this PR is breaking `wcs-lib`. Results on [check.ocamllabs.io](http://check.ocamllabs.io) say both versions worked before with `sedlex.2.1`

kit-ty-kate:
I removed the two latest commit as they don't seem to fix the problem and the problem itself in wcs-lib seems unrelated but I cannot find the cause. Thanks a lot!

toots:
Thanks!

PR #14575, opened 2019-07-23
Update eqaf.0.5 with optional dependencies

dinosaure:
Add depopts to `eqaf.0.5` to be able to provide `eqaf.cstruct` if `cstruct` will be/is available.

### DIFF
--- a/packages/eqaf/eqaf.0.5/opam
+++ b/packages/eqaf/eqaf.0.5/opam
@@ -21,7 +21,7 @@ depends: [
   "alcotest"       {with-test}
   "crowbar"        {with-test}
 ]
-depotps: [
+depopts: [
   "cstruct" {>= "4.0.0"}
   "bigarray-compat"
 ]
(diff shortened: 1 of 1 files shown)


dinosaure:
Seems weird, tests was not launched even if I put:

```
runtest: [ "dune" "runtest" "-p" name ]
```

I'm wrong?

kit-ty-kate:
> Seems weird, tests was not launched even if I put:
> 
> ```
> runtest: [ "dune" "runtest" "-p" name ]
> ```
> 
> I'm wrong?

the name of the field is `run-test` not `runtest`

kit-ty-kate:
Same for `depopts`

kit-ty-kate:
There is a failure on OCaml 4.03 for some reason:
```
# > Merge results.
# > Start linear regression.
# eqaf: 9559.614005 ns/run.
# string.equal: 94.026041 ns/run.
# Fail with B¹ = -32.902366, B² = -87.233013.
# Command [31] exited with code 1:
# $ (cd _build/default/check && ./check.exe)
```

dinosaure:
Yes, the only way to fix it is to relaunch the test. Currently, it records some noises from virtualization and stamps about time does say the truth ... So tests are fragile :/

dinosaure:
As a counter example, 4.03.0 works with Travis ...

kit-ty-kate:
thanks

PR #14577, opened 2019-07-24
added packages/molenc/molenc.2.0.0

UnixJunkie:


kit-ty-kate:
There are various problems in system rdkit packages but the package seems good enough (according to the Travis checks)

kit-ty-kate:
Thanks!

UnixJunkie:
Its strange: I updated conf-rdkit recently so I guess it was working on most systems at that time.
But thanks for merging.

kit-ty-kate:
debian stable has been upgraded, this is maybe why

PR #14580, opened 2019-07-24
opam parallel build: include ocaml-base-compiler.4.08.0

gasche:
See https://github.com/ocaml/opam-repository/pull/14257#issuecomment-514561088


kit-ty-kate:
Thanks!

PR #14591, opened 2019-07-26
[new release] certify (0.3.0)

yomimono:
CLI utilities for simple X509 certificate manipulation

- Project page: <a href="https://github.com/yomimono/ocaml-certify">https://github.com/yomimono/ocaml-certify</a>
- Documentation: <a href="https://yomimono.github.io/ocaml-certify/doc">https://yomimono.github.io/ocaml-certify/doc</a>

##### CHANGES:

* breaking change: `sign`, `selfsign`, and `csr` are now subcommands of a `certify` binary.  To update your workflow, simply prepend `certify` to any command you previously ran. (@reynir)
* improvements: allow --key for key filename in `sign`; write to csr.pem by default in `csr`. (@yomimono)
* maintenance: use newer x509 (v.0.7.0); make opam 2.0-compliant; use dune instead of topkg/ocamlfind; fix tests for some platforms. (@hannesm, @yomimono)

PR #14600, opened 2019-07-27
Add conf-openssl-bin

kit-ty-kate:
In #14591 the CI fails because the `certify` package requires `openssl` but the `conf-openssl` package is not the right package, it does not pull the `openssl` binary but `libssl` instead. The reason it works on some systems is that either openssl is already installed by default, or there is only one package that installs both `libssl` and `openssl`.

* To remedy this, this PR creates a new `conf-openssl-bin` package and specify in the description of `conf-openssl` that it installs only the library. This is the least invasive change, but in my opinion, not the best.
* The other option would be to just rename `conf-openssl` to `conf-libssl` and rename the `conf-openssl-bin` created by this PR to `conf-openssl`. This is a bit more invasive because it requires changing the packages using it, but the `conf-openssl` package should have been correctly named `conf-libssl` from the beginning in my opinion.

What do people think about the two different proposals?
(pinging maintainers of packages using `conf-openssl`)
cc @yomimono @xclerc @savonet @constfun


kit-ty-kate:
Does anyone have any opinion on the matter?

xclerc:
I don't, but @tiash might.

hannesm:
I like clean names, i.e. renaming `conf-openssl` to `conf-libssl`; and introducing this new package as `conf-openssl`. since the current `conf-openssl` is not used much, I wouldn't be afraid of the invasive change.

hcarty:
Renaming a package would break anything using `opam lock` with a lockfile pointing at `conf-openssl`.  Those would live outside of opam-repository so they wouldn't be visible from the local CI.

kit-ty-kate:
@hcarty what if we create a new version for `conf-openssl`? (`conf-openssl.2`) and then comment the first one by saying not to use it and that it is only there for opam lock users?

hcarty:
@kit-ty-kate That sounds like a very good solution to me - thank you!

kit-ty-kate:
I made the necessary changes to the packages using `conf-openssl`. Thanks everyone for your inputs (and sorry to have let this drag on)

PR #14603, opened 2019-07-29
Add ocamlfind.1.8.1

kit-ty-kate:
OCamlfind 1.8.1 has just been released. This release adds support for OCaml >= 4.09


kit-ty-kate:
The MacOS build keeps going through internal failures even after 2 rebuild. Seems ok otherwise.

PR #14604, opened 2019-07-29
Various fixes (ter)

kit-ty-kate:
Pinging maintainers whose latest versions were broken to fix their opam file upstream:
* @whitequark @yallop for lz4
* @tokenrove for tsdl-mixer


tokenrove:
I've updated the upstream repo for tsdl-mixer.  Anything else I should do immediately?

kit-ty-kate:
> I've updated the upstream repo for tsdl-mixer. Anything else I should do immediately?

No it's all good, thanks a lot.

PR #14618, opened 2019-07-31
js_of_ocaml.3.4.0 is not compatible with OCaml 4.09

xclerc:
(actually tested with 4.09.0+beta1)

@hhugo


kit-ty-kate:
Thanks!

TypesLogicsCats:
I cloned the latest version of js_of_ocaml and was able to install it just fine on the 4.09 branch. The only thing that didn't build was something in `toplevel/examples/lwt_toplevel` because the graphics package is no longer distributed with OCaml. Could the version constraint please be lifted?

PR #14620, opened 2019-08-01
Ignore Vim swap files

dra27:
9e0f6ee27cf5ee95e4128294c9901d603de8a634 added the pattern for the emacs temp files - finally, after however many years, here's the Vim pattern!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *~
+*.sw?
 archives/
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Thanks!

dra27:
One of my more profound contributions this week 🙂

PR #14624, opened 2019-08-01
ppxlib.0.9.0

xclerc:
This release bumps the AST version from 4.07 to 4.08,
and it thus likely to break reverse dependencies.

For this reason, constraints have been added to said
reverse dependencies (https://github.com/ocaml/opam-repository/pull/14208 and https://github.com/ocaml/opam-repository/pull/14623).

This goes without saying, but this pull request should
only be merged after https://github.com/ocaml/opam-repository/pull/14623 is.

### DIFF
--- a/packages/ppxlib/ppxlib.0.9.0/opam
+++ b/packages/ppxlib/ppxlib.0.9.0/opam
@@ -1,0 +1,45 @@
+opam-version: "2.0"
+maintainer: "opensource@janestreet.com"
+authors: ["Jane Street Group, LLC <opensource@janestreet.com>"]
+homepage: "https://github.com/ocaml-ppx/ppxlib"
+bug-reports: "https://github.com/ocaml-ppx/ppxlib/issues"
+dev-repo: "git+https://github.com/ocaml-ppx/ppxlib.git"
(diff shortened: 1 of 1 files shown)


xclerc:
(The reported "revdeps" failures are fixed by https://github.com/ocaml/opam-repository/pull/14623.)

kit-ty-kate:
There is a whole heap of errors popping up with a local install of OCaml 4.06: https://travis-ci.org/ocaml/opam-repository/jobs/566384537

Could you have a look. Otherwise it looks good but I want to be sure

xclerc:
I will have a look.

xclerc:
I have not been able to reproduce the problem locally
(after carefully installing the very same set of packages).

If I read the Travis log correctly, ppxlib builds (line 3581)
but then the tests fail. The first error:

    `Unknown directive `require'`

is usually caused by an ocamlfind mis-installation, but it
looks like version 1.8.1 has been installed. In other words,
I do not know what is going on.

kit-ty-kate:
Ok, let's merge it and figure if there is really something wrong later. Thanks!

PR #14626, opened 2019-08-01
Speed up re-configuring OCaml sources

dra27:
The `graphics` package has to re-`configure` the OCaml source tree when the package is added to a compiler which was built without it.

The `configure` script takes a bit longer in OCaml 4.08.0 but, since it is built with GNU autoconf, there is a solution which is the `-C` parameter to `configure` (with thanks to @stedolan for originally bringing to my attention just how much faster it makes things!). This creates a file `config.cache` which is used in subsequent invocations to prevent probing the system again.

This alters `ocaml-base-compiler.4.08.0` to have the `-C` switch which has no effect on the build other than to cause `config.cache` to be created when `configure` is run and then this file is installed to `share/ocaml/config.cache` in the switch.

It makes a considerable difference installing graphics.4.08.0, knocking 20 seconds off the package's build time!

This would obviously also be wanted in ocaml-base-compiler.4.08.1 when it comes out. I don't propose doing this for `ocaml-variants`. Obviously the `graphics` package is entirely standalone from OCaml 4.09.0 onwards, but I'd propose keeping this tweak in ocaml-base-compiler.4.09.0 as well, since at some point there are other packages I'd like to be able to build from the source tree, so I anticipate other packages which will want to re-`configure` the tree.

### DIFF
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.4.08.0/files/ocaml-base-compiler.install
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.4.08.0/files/ocaml-base-compiler.install
@@ -1,0 +1,1 @@
+share_root: ["config.cache" {"ocaml/config.cache"}]
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.4.08.0/opam
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.4.08.0/opam
@@ -18,6 +18,7 @@ build: [
   [
     "./configure"
     "--prefix=%{prefix}%"
+    "-C"
     "CC=cc" {os = "openbsd" | os = "freebsd" | os = "macos"}
     "ASPP=cc -c" {os = "openbsd" | os = "freebsd" | os = "macos"}
   ]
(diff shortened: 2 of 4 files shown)


dra27:
(Travis failure is transient)

avsm:
Thanks! Has this been submitted upstream to ocaml/ocaml so that the install file will be in the 4.09 release?

avsm:
Also, does this also need to be applied to the various variant compilers, or is the base compiler package sufficient?

dra27:
My understanding from https://github.com/ocaml/opam-repository/pull/12700#issuecomment-425105053 is that the change will be picked up automatically when `ocaml-base-compiler.4.08.0` is used as a template for 4.09.0 (I will shortly do a PR for the missing 4.08.1 change, as it was released between this PR being opened and merged!)

dra27:
I decided for simplicity to leave the variant packages alone, mainly because I worried that if you're on a `trunk` variant, the cache could become obscurely invalid.

mseri:
Sorry for the delay

dra27:
No problem!

Octachron:
Writing down what I said to @dra27 offline, the change should indeed be picked automatically when creating the 4.09.0 ocaml base compiler package.

PR #14627, opened 2019-08-01
Add graphics.4.08.0 and graphics.4.08.1

dra27:
This started as adding `graphics.4.08.0`, anyway! It also addresses a few subtleties:

- The `graphics` package can fail when installing to a system compiler where the OCaml `Makefile` still used `cp` rather than `install` (I saw this with OCaml 4.05.0 on Ubuntu) - the fix is a pair of `mkdir -p`s in the script
- `opam reinstall graphics` was broken - because opam 2.0 builds the new package before uninstalling the old one, it meant that more care was needed for detecting whether the library should be built or was just a no-op. The new packages now write a `state` file (there's a dirty hack to read the `.changes` file otherwise, for the transition - that works with sandboxing now, the idea being that if it stopped working in the future, the old package description would already have gone).
- Although it never results in more than a warning, if `ocamlfind` and `graphics` are simultaneously rebuilt, the results were not entirely predictable. I've changed `graphics` to be a `depopt` of `ocamlfind` which is certainly philosophically correct.

graphics.4.08.0 takes advantage of #14626, but also works without it.

The files are generated, so it's only necessary to review the 3 files from one directory. The scripts used are in https://github.com/dra27/opam-repository/tree/graphics-gen-script/packages/graphics/scripts.


mseri:
@avsm I trust @dra27 but I don't feel entitled to rewiew this PR. Would you mind having a look?

dra27:
Thanks @mseri 🙂

The PR needs updating to add `graphics.4.08.1`, but I can't push updates to it without knowing that someone can cancel datakit CI, since the updates to `ocamlfind` find result in too many tasks. I don't think that the two changes can be separated, unfortunately.

kit-ty-kate:
Thanks a lot!

PR #14632, opened 2019-08-02
[new release] facteur (0.1.0)

dinosaure:
CHANGES:

- First release

### DIFF
--- a/packages/facteur/facteur.0.1.0/opam
+++ b/packages/facteur/facteur.0.1.0/opam
@@ -15,8 +15,8 @@ run-test: [ "dune" "runtest" "-p" name "-j" jobs ]
 
 depends: [
   "ocaml" {>= "4.03.0"}
-  "mrmime"
-  "colombe"
+  "mrmime" {>= "0.1.0"}
+  "colombe" {>= "0.1.0"}
   "sendmail-lwt"
   "logs"
   "fmt"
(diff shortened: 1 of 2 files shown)


packages/facteur/facteur.0.1.0/opam
```diff
@@ -0,0 +1,52 @@
...
+depexts: [
+  ["file-dev"] {os-distribution = "alpine"}
+  ["file-devel"] {os-distribution = "centos"}
+  ["file-devel"] {os-distribution = "fedora"}
+  ["file-devel"] {os-family = "suse"}
+  ["libmagic-dev"] {os-family = "debian"}
+  ["libmagic"] {os = "macos" & os-distribution = "homebrew"}
+]
+
+conflicts: [
```
kit-ty-kate: Why are those conflicts when they already are hard dependencies. Could you just add the constraint `{>= "0.1.0"}` directly to the dependencies above?

packages/facteur/facteur.0.1.0/opam
```diff
@@ -0,0 +1,52 @@
...
+  "astring"
+  "fpath"
+  "cmdliner"
+  "domain-name" {>= "0.3.0"}
+  "x509" {>= "0.7.0"}
+  "emile" {>= "0.5"}
+  "dune" {>= "1.3"}
+]
+
+depexts: [
```
kit-ty-kate: Would it be possible to have those depexts be defined in their own package? I'm not too sure if the depexts are well defined but if it's possible it would be better off in a meta package in my opinion.

dinosaure: Currently, this depext comes from the `magic` package which uses `libmagic-dev` too. About a meta-package, I don't know well how to do that.

dinosaure:
Should be good now, I think 👍 !

kit-ty-kate:
Thanks a lot!

PR #14635, opened 2019-08-04
labrys.0.1: Fix compilation

kit-ty-kate:
This PR introduces a patch to fix the compilation of `labrys.0.1` (patch from upstream). This fix requires `dune >= 1.10`


kit-ty-kate:
Failing because LLVM 8.0 is not available on debian-stable. The PR has been tested locally, it should be fine.

PR #14639, opened 2019-08-05
Update ocp-pp.1.99.17-beta/opam

Et7f3:
Same as https://github.com/ocaml/opam-repository/pull/14617
Avoid big bold text.


kit-ty-kate:
Thanks!

PR #14642, opened 2019-08-05
Release of OCaml 4.08.1

Octachron:
This is the official release of OCaml 4.08.1 .
This is exactly the third release candidate, with a new and shiny tag, version number, and a few more variants available to match the 4.08.0 variants.

### DIFF
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.4.08.1/opam
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.4.08.1/opam
@@ -1,0 +1,42 @@
+opam-version: "2.0"
+synopsis: "Official release 4.08.1"
+maintainer: "platform@lists.ocaml.org"
+authors: "Xavier Leroy and many contributors"
+homepage: "https://ocaml.org"
+bug-reports: "https://github.com/ocaml/ocaml/issues"
+dev-repo: "git://github.com/ocaml/ocaml"
+depends: [
+  "ocaml" {= "4.08.1" & post}
+  "base-unix" {post}
+  "base-bigarray" {post}
+  "base-threads" {post}
+]
+conflict-class: "ocaml-core-compiler"
+flags: compiler
+setenv: CAML_LD_LIBRARY_PATH = "%{lib}%/stublibs"
+build: [
+  ["./configure" "--prefix=%{prefix}%"]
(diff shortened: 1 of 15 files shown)


kit-ty-kate:
cc @dra27 could it be improved by #14626 ?

Octachron:
I hesitated to integrate the configure cache option, but I thought it would better to let #14626 take care of the option uniformly across versions.

dra27:
Indeed - I wasn’t expected #14626 to be merged before 4.08.1 (if at all), and I’ll update it once this one’s merged.

Octachron:
Are the failures for the musl, 32bit and bytecode-only variants expected?

dra27:
The bytecode failure is ocaml/ocaml#8855 - looking at the PR when 4.08.0 was released to opam-repository, I think this bug slipped into 4.08.0 because of CI trouble at the time.

dra27:
The musl and 32-bit failures look like packaging changes resulting from the 4.08.0 autoconf switch.

Octachron:
The bytecode failure does not seems that directly related to ocaml/ocaml#8855 ? The CI fails on a call to `ocamlopt` when building package. However, in the bytecode-only case,  it is normal that `ocamlopt` was not installed.

Concerning the musl and 32bit variant, it seems that I forget to add a depext on the musl and multilib gcc required by those variants. I will fix this (depexts are supported for compiler variants, aren't they?).

kit-ty-kate:
wait, there shouldn't be any CI failures (or tests for that matter). How is ocaml-variants co-installable with ocaml-base-compiler?

dra27:
@Octachron - you're of course right, too many bytecode PRs at the moment! Looking at the history, it's a bug which has always been there in `opam-depext`'s build system. The test in https://github.com/ocaml/opam-depext/blob/14a9aa6093bfb510b83f2ba07205b899b7e89989/src_ext/Makefile#L22 is incorrect (that's been pulled out of an old opam `src_ext/Makefile` where `OCAMLOPT` was set by a `configure` test)

kit-ty-kate:
oh it's not coinstallable, it's just ocaml-version which had set the latest version to 4.08.1 already. Anyway, I guess it should be ok to merge but I'm still surprised we can still upgrade from one version of `ocaml-variants` to another. It doesn't matter for now, thanks a lot!

PR #14645, opened 2019-08-06
Relax constraint to latest ocaml patch release.

bschommer:
I was just a tad too fast. The patch release should also be fine.

### DIFF
--- a/packages/camlp5/camlp5.7.08/opam
+++ b/packages/camlp5/camlp5.7.08/opam
@@ -13,7 +13,7 @@ doc: "https://camlp5.github.io/doc/html"
 install: [make "install"]
 synopsis: "Preprocessor-pretty-printer of OCaml"
 depends: [
-  "ocaml" {>= "4.02" & <= "4.08.0"}
+  "ocaml" {>= "4.02" & <= "4.08.1"}
 ]
 extra-files: ["camlp5.install" "md5=b16bcaf28bfb364236b2181fb50a6564"]
 url {


mseri:
Are all the `coq`/`coqide` failures expected?

bschommer:
> Are all the `coq`/`coqide` failures expected?

Not really, maybe they are a result of the dynlinking patches in OCaml 4.08.1. Maybe one of the coq maintainers knows more?

bschommer:
It seems that at least another person has the issue, see coq/coq#10404

mseri:
PIng @ejgallego

ejgallego:
What is the make vendor / version?

packages/camlp5/camlp5.7.08/opam
```diff
@@ -13,7 +13,7 @@ doc: "https://camlp5.github.io/doc/html"
 install: [make "install"]
 synopsis: "Preprocessor-pretty-printer of OCaml"
 depends: [
-  "ocaml" {>= "4.02" & <= "4.08.0"}
+  "ocaml" {>= "4.02" & <= "4.08.1"}
```
kit-ty-kate:
```suggestion
  "ocaml" {>= "4.02" & < "4.09.0"}
```
Why not just use strict comparaisons?

ejgallego: In particular camlp5 does not support 4.08.2 would that version be released, each OCaml version does require special support upstream, see https://github.com/camlp5/camlp5/tree/master/ocaml_stuff

ejgallego:
As pointed out by @skyskimmer in the Coq BT, this is due to change in semantics for dynlink:
```
Error while loading "/home/opam/.opam/4.08/lib/camlp5/pr_dump.cmo": The module `Pr_dump' is already loaded (either by the main program or a previously-dynlinked library).
```
The invocation Coq uses fails. This could be fixed in Coq, but would require a new point release.

Coq 8.10 and `master` don't use camlp5 so this issue went unnoticed.

ejgallego:
> This could be fixed in Coq, but would require a new point release.

After a bit of research I am not sure this is Coq's fault, it seems to me that `camlp5` may indeed be broken in itself.

bschommer:
Since it seems that the issue will be fixed in the next release of camlp5 I will close this PR.

PR #14647, opened 2019-08-06
Set an upper bound on merlin-extend for Reason < 3.5.0

anmonteiro:
`merlin-extend` changed the dune library name in 0.4 from `merlin_extend` to `merlin-extend`, and Reason has only adapted to this change since 3.5.0


anmonteiro:
This was quite the rabbit hole but I'm happy with the results now. The Travis failures just look like timeouts so I think it's safe to go ahead and merge this if the maintainers are happy with it.

mseri:
Thanks

PR #14659, opened 2019-08-09
[new release] x509 (0.7.1)

hannesm:
CHANGES:

* revert General_name.t (DNS and IP components) to string/Cstruct.t list
  - NameConstraints uses DNSname with strings with leading dots (.example.com),
    which are not valid RFC1034 domain names
  - NameConstraints uses IP with IP/netmask, i.e. using 8 octets for an IPv4
    address
  - X509.Certificate.hostnames still return a Domain_name.Set.t
  - reported by @reynir
* drop ipaddr dependency


hannesm:
see discussion in #14591 in respect to the failed CI job. this is imho good to be merged.

avsm:

PR #14662, opened 2019-08-09
BER MetaOCaml N101 (4.01.0+BER)

yallop:
Fixes so that BER MetaOCaml N101 works again:
* Since the patch on http://pim.happyleptic.org is no longer available, retrieve the code from the repository at https://github.com/metaocaml/ber-metaocaml/ instead
* Add 4.01.0+BER to the OCaml version constraints in base-metaocaml-ocamlfind
* Fix the checksum for the clang patch


avsm:
Thanks!

PR #14664, opened 2019-08-10
Don't disable dynlink in -static switches

### LINKED ISSUE #14663: ppx_deriving cannot be installed in the -static switches
For example in a `ocaml-variants.4.08.1+musl+static+flambda` switch:

```
    ocamlopt src/.ppx_deriving_main.eobjs/native/findlib_initl.{cmx,o}
File "_none_", line 1:
Warning 58: no cmx file was found in path for module Dynlink, and its interface was not compiled with -opaque
    ocamlopt src/.ppx_deriving_main.eobjs/native/ppx_deriving_main.{cmx,o}
File "_none_", line 1:
Warning 58: no cmx file was found in path for module Dynlink, and its interface was not compiled with -opaque
    ocamlopt src/ppx_deriving_main.exe (exit 2)
(cd _build/default && ~/.opam/test/bin/ocamlopt.opt -w -9 -g -o src/ppx_deriving_main.exe -linkall -linkall -I ~/.opam/test/lib/findlib -I ~/.opam/test/lib/ocaml-migrate-parsetree -I ~/.opam/test/lib/ocaml/compiler-libs -I ~/.opam/test/lib/ppx_derivers -I ~/.opam/test/lib/ppx_tools -I ~/.opam/test/lib/result -I src/api ~/.opam/test/lib/ocaml/compiler-libs/ocamlcommon.cmxa ~/.opam/test/lib/ppx_tools/ppx_tools.cmxa ~/.opam/test/lib/result/result.cmxa ~/.opam/test/lib/ppx_derivers/ppx_derivers.cmxa ~/.opam/test/lib/ocaml-migrate-parsetree/migrate_parsetree.cmxa src/api/ppx_deriving_api.cmxa ~/.opam/test/lib/findlib/findlib.cmxa ~/.opam/test/lib/findlib/findlib_dynload.cmxa src/.ppx_deriving_main.eobjs/native/findlib_initl.cmx src/.ppx_deriving_main.eobjs/native/ppx_deriving_main.cmx)
File "_none_", line 1:
Error: No implementations provided for the following modules:
         Dynlink referenced from ~/.opam/test/lib/findlib/findlib_dynload.cmxa(Fl_dynload),
           src/.ppx_deriving_main.eobjs/native/findlib_initl.cmx,
           src/.ppx_deriving_main.eobjs/native/ppx_deriving_main.cmx
```

Might be the same as https://github.com/backtracking/ocamlgraph/issues/80.

copy:
This removes the `-no-shared-libs` flag from the configuration of static switches.  Note that this switch doesn't produce static binaries by default, only when `-ccopt -static` is passed to `ocamlopt`.

This fixes #14663 and https://github.com/backtracking/ocamlgraph/issues/80 (verified by running `opam install ppx_deriving ocamlgraph`). It's also related to #3608.

One downside of this change is that it replace a compile-time error (`Dynlink` isn't built when `-no-shared-libs` is configured) with a run-time error (`Failure "Dynamic loading not supported"`). This only affects executables compiled with `-ccopt -static`.

cc @etanol

### DIFF
--- a/packages/ocaml-variants/ocaml-variants.4.01.0+musl+static/opam
+++ b/packages/ocaml-variants/ocaml-variants.4.01.0+musl+static/opam
@@ -28,7 +28,6 @@ build: [
     "-no-tk"
     "-no-curses"
     "-no-graph"
-    "-no-shared-libs"
   ]
   [make "world.opt"]
 ]
--- a/packages/ocaml-variants/ocaml-variants.4.02.1+musl+static/opam
+++ b/packages/ocaml-variants/ocaml-variants.4.02.1+musl+static/opam
@@ -27,7 +27,6 @@ build: [
     "-static"
     "-no-curses"
     "-no-graph"
-    "-no-shared-libs"
   ]
   [make "world.opt"]
 ]
--- a/packages/ocaml-variants/ocaml-variants.4.02.3+musl+static/opam
+++ b/packages/ocaml-variants/ocaml-variants.4.02.3+musl+static/opam
@@ -27,7 +27,6 @@ build: [
     "-static"
     "-no-curses"
     "-no-graph"
-    "-no-shared-libs"
   ]
   [make "world.opt"]
 ]
--- a/packages/ocaml-variants/ocaml-variants.4.05.0+musl+flambda/opam
+++ b/packages/ocaml-variants/ocaml-variants.4.05.0+musl+flambda/opam
@@ -1,0 +1,37 @@
+opam-version: "2.0"
+synopsis:
+  "Official 4.05.0 release compiled with musl-clang and with flambda activated"
+description:
+  "Requires musl-clang to be installed (package musl on Arch Linux or musl-tools on Debian)"
+maintainer: "platform@lists.ocaml.org"
+depends: [
+  "ocaml" {= "4.05.0" & post}
+  "base-unix" {post}
+  "base-bigarray" {post}
+  "base-threads" {post}
+]
+conflict-class: "ocaml-core-compiler"
+flags: compiler
+setenv: CAML_LD_LIBRARY_PATH = "%{lib}%/stublibs"
+build: [
+  [
+    "./configure"
+    "-prefix"
+    prefix
+    "-with-debug-runtime"
+    "-flambda"
+    "-cc"
+    "musl-clang -Os"
+    "-aspp"
+    "musl-clang -c"
+    "-no-curses"
+    "-no-graph"
+  ]
+  [make "world"]
+  [make "world.opt"]
+]
+install: [make "install"]
+url {
+  src: "https://github.com/ocaml/ocaml/archive/4.05.0.tar.gz"
+  checksum: "md5=7e0079162134336a24b9028349c756bb"
+}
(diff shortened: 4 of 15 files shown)


etanol:
Thanks for this, and apologies for not taking a look sooner.

> One downside of this change is that it replace a compile-time error (`Dynlink` isn't built when `-no-shared-libs` is configured) with a run-time error (`Failure "Dynamic loading not supported"`). This only affects executables compiled with `-ccopt -static`.

Also, thanks for the explanation.  I had been confused a bit about musl's dynamic loading support.

I didn't notice that musl has **two** `dlopen` implementations:

1. [The dummy one](http://git.musl-libc.org/cgit/musl/tree/src/ldso/dlopen.c), used in statically linked musl
1. [The one from the dynamic linker/loader](http://git.musl-libc.org/cgit/musl/tree/ldso/dynlink.c#n1956), for dynamically linked executables compiled with `musl-gcc`

In summary, these changes are perfectly fine.  But they raise a different question.  Since these variants now allow for **both** dynamic and static linking against musl libc, shouldn't we remove the `+static+` tag from the name?  It looks confusing.

osener:
Is this switch intended for distributions that ship with GCC by default? I've been successfully using `<version>+flambda` to build statically linked binaries on Alpine Linux without using this switch. I thought I had observed smaller binaries with `+musl+static+flambda`, but now I can not reproduce it.

Also, this switch does not build on Alpine Linux as it is, since Alpine does not have a `musl-gcc` binary, and instead comes with `gcc` and `x86_64-alpine-linux-musl-gcc`.

kit-ty-kate:
For what it's worth, the original ocaml-variants package included both with +static and without +static: https://github.com/ocaml/opam-repository/pull/2025

etanol:
@kit-ty-kate good point.  Plus the `musl` variant could be identical to the standard one when compiled under Apline Linux.

copy:
@osener That's correct, these switches are intended for static compilation on platforms that use glibc by default, but allow you to install a musl toolchain separately (Debian, Arch Linux, …). On Alpine, you can continue using a regular switch. The size difference might have been due to the `-Os` flag that this switch adds by default, although I believe it's overwritten by most build systems anyway.

> Since these variants now allow for both dynamic and static linking against musl libc, shouldn't we remove the +static+ tag from the name? It looks confusing.

> For what it's worth, the original ocaml-variants package included both with +static and without +static

I'm a bit wary about removing switches that people might be using. I'll leave the old switches with the fix and add new +musl switches. The last remaining difference (from the original MR) between the +musl+static and the +musl switches is then that in the former, the OCaml compiler itself is compiled statically.

copy:
Something else I found: Starting in 4.09, you will be able to create a musl switch simply by setting `CC` at switch creation time: `CC=musl-gcc opam switch create test 4.09.0+beta1`.

Previously this didn't work unless you had X11 header files in your musl toolchain to compile graphslib. [graphs was moved out of the compiler recently](https://github.com/ocaml/ocaml/pull/2318), fixing this issue.

copy:
This should be good to merge from my point of view. Any objections/questions?

etanol:
No objection.  I think the `static` tag is a bit of a lie currently, because the generated OCaml compiler binaries are still dynamically linked.

mseri:
I think this cleans up confusion, I am going to merge

PR #14668, opened 2019-08-11
allow compilation of coq 8.9 with ocaml 4.08

garrigue:
Coq 8.9.1 doesn't compiler with ocaml 4.08 due to an incompatibility in dynamic loading when calling camlp5.
This patch fixes that.

### DIFF
--- a/packages/coq/coq.8.9.1/files/ocaml408_compat.patch
+++ b/packages/coq/coq.8.9.1/files/ocaml408_compat.patch
@@ -1,0 +1,13 @@
+diff --git a/Makefile.build b/Makefile.build
+index ed29e21c82..02a92cd404 100644
+--- a/Makefile.build
++++ b/Makefile.build
+@@ -262,7 +262,7 @@ endef
+ CAMLP5DEPS:=grammar/grammar.cma
+ CAMLP5USE=pa_extend.cmo q_MLast.cmo pa_macro.cmo -D$(CAMLVERSION)
+ 
+-PR_O := $(if $(READABLE_ML4),pr_o.cmo,pr_dump.cmo)
++PR_O := $(if $(READABLE_ML4),pr_o.cmo,)
+ 
+ # Main packages linked by Coq.
+ SYSMOD:=-package num,str,unix,dynlink,threads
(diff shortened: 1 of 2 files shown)


garrigue:
Unfortunately, due to the bump of ocaml to 4.08.1 and awaiting a compatible camlp5 version, there seems to be no way to test this patch with 4.08.0 in the CI.
Or is there?
At least, for me it works locally.

garrigue:
Not sure how to fix opam-lint errors.
I don't understand the mismatch error, and I didn't touch the build instructions.

avsm:
The CI only tests the very latest patch release of each major release due to the combinatorial explosion otherwise.  I'll manually test this.

garrigue:
I see. For this kind of situation, it would be nice to be able to test with the latest version of ocaml that satisfies the constraints. Am I correct that opam does that for everything except ocaml itself?

avsm:
That's right -- we do test every compiler from 4.02 onwards, but only the latest patch release of each. So your PR was tested on 4.02.3, 4.03.0, 4.04.2, 4.05.0, 4.06.0, 4.07.1 as well as 4.08.1 (see the "Compilers" section of https://ci.ocaml.org/ocaml/opam-repository/pr/14668?test=2%20Revdeps)

Thanks for the PR! Merging.

avsm:
Ah sorry, just one query before merging it.  Do you really want this patch applied on previous versions of OCaml as well? That changes the Coq package and may be surprising for users who uses `pr_dump` find on <4.08.  I can modify your patch to only apply the 4.08 patch on OCaml 4.08.0 until a fixed camlp5 is available.

garrigue:
Indeed, when I wrote the patch it was not decided whether we would fix camlp5 or not. Now it is sufficient to have it applied in the interim. (But I wouldn't know how to do that)

garrigue:
Note however that the use of `pr_dump.cmo` is actually a long standing bug of coq, which was just silently ignored before ocaml 4.08.0. So applying the patch to older versions of ocaml is not dangerous (as the successful CI demonstrates).

mseri:
@garrigue I tried to add a filter to limit the patch presence. Let's see if it works

mseri:
It did not work. I'll merge as it was

PR #14675, opened 2019-08-13
checkseum / digestif: with variants: conflict with mirage-runtime < 4.0.0

hannesm:
require ocaml-freestanding at least 0.4.3 (the first one to install cflags et al).

//cc @dinosaure


hannesm:
PRed upstream https://github.com/mirage/checkseum/pull/33 and https://github.com/mirage/digestif/pull/89

dinosaure:
Agreed, thx :+1:

PR #14678, opened 2019-08-14
otf v0.1 submission.

pictyeye:


packages/otf/otf.0.1/opam
```diff
@@ -0,0 +1,16 @@
+opam-version: "2.0"
+name: "otf"
+version: "0.1"
+synopsis: "otf is a simple Output Test Framework"
+maintainer: "Olivier Levillain <olivier.levillain@telecom-sudparis.eu>"
+authors: "Olivier Levillain <olivier.levillain@telecom-sudparis.eu>"
+license: "BSD 3-Clause License"
+homepage: "https://gitlab.com/pictyeye/otf"
+bug-reports: "https://gitlab.com/pictyeye/otf/issues"
+depends: ["ocaml" "dune"]
```
kit-ty-kate:
```suggestion
depends: [
  "ocaml"
  "dune" {>= "1.6"}
]
```
According to your `dune-project` file, your package requires dune >= 1.6

packages/otf/otf.0.1/opam
```diff
@@ -0,0 +1,16 @@
...
+build: ["dune" "build" "-p" name]
```
kit-ty-kate:
```suggestion
build: ["dune" "build" "-p" name "-j" jobs]
```
It would be more user friendly to allow users to set the number of jobs needed to build this package.

packages/otf/otf.0.1/opam
kit-ty-kate: Sounds like you also need a constraint on the OCaml compiler:
```
#=== ERROR while compiling otf.0.1 ============================================#
# context              2.0.5 | linux/x86_64 | ocaml-base-compiler.4.04.2 | pinned(https://gitlab.com/pictyeye/otf/-/archive/0.1/otf-0.1.tar.bz2)
# path                 ~/.opam/4.04/.opam-switch/build/otf.0.1
# command              ~/.opam/4.04/bin/dune build -p otf
# exit-code            1
# env-file             ~/.opam/log/otf-15-9496e5.env
# output-file          ~/.opam/log/otf-15-9496e5.out
### output ###
#       ocamlc .otf.eobjs/byte/otf.{cmi,cmo,cmt} (exit 2)
# (cd _build/default && /home/opam/.opam/4.04/bin/ocamlc.opt -w -40 -g -bin-annot -I .otf.eobjs/byte -no-alias-deps -o .otf.eobjs/byte/otf.cmo -c -impl otf.ml)
# File "otf.ml", line 104, characters 8-23:
# Error: Unbound value Config.find_opt
```
I've edited the suggestion

pictyeye: Thank you for your insight. I added the corresponding constraints.

However, there was a dependency on a recent version of dune (strictly > 1.6.2 which is in Debian Buster) so I changed the metadata a little. I will refresh my PR soon.

packages/otf/otf.0.1/opam
pictyeye: No problem. However, I copy/pasted this from https://opam.ocaml.org/doc/Packaging.html (the "Creating a package definition file" section). Maybe you should consider updating this page?

pictyeye:
Even though I added the constraints in otf.opam, the Compilers check still rejects the package. What should I do?

kit-ty-kate:
Thanks a lot!

PR #14679, opened 2019-08-14
ocveralls will not be compatible with Bisect_ppx 1.5.0

aantron:
[ocveralls](https://github.com/sagotch/ocveralls) is deprecated. Its functionality was incorporated directly into Bisect_ppx in aantron/bisect_ppx#176.

Bisect_ppx will change in 1.5.0 in a way that will break existing ocveralls releases, so this PR adds the needed constraints.


mseri:
Thanks

PR #14681, opened 2019-08-15
[new release] dns, dns-certify, dns-mirage, dns-resolver, dns-client, dns-server, dns-tsig and dns-cli (4.0.0)

hannesm:
CHANGES:

* Switch to uDNS implementation, developed from scratch since 2017, primarily
  focusing on a recursive caching resolver. The server part supports dynamic
  updates (RFC 2135), transaction authentication with HMAC (RFC 2845), zone
  transfer (RFC 5936), incremental zone transfer (RFC 1995), change
  notifications (RFC 1996) amongst others.
* The core library uses a GADT for resource record sets, where the key (resource
  record type) specifies the value type.
* The API does not leak exceptions, but uses the result type where appropriate.
* TCP transport is well supported and used widely (client uses it by default)
* Naming: client is a DNS client, resolver is the recursive resolver library
* The DNS library is split into the following opam packages and sublibraries:
  - `dns` - the core library
  - `dns-tsig` - transaction signatures
  - `dns-zone` - zone file parser (mostly taken from the 1.x series)
  - `dns-cli` - command line utilities (odig, onotify, ..)
  - `dns-client` - pure client implementation
    - `.unix` - DNS client using the Unix module for communication
    - `.lwt` - DNS client using Lwt_unix for communication
    - `.mirage` - DNS client using MirageOS for communication
  - `dns-certify` - helpers for let's encrypt provisioning
    - `.mirage` - certificate provisioning with MirageOS
  - `dns-mirage` - generic MirageOS communication layer
  - `dns-server` - pure server implementation
    - `.mirage` - MirageOS primary and secondary server
  - `dns-resolver` - pure recursive resolver implementation
    - `.mirage` - MirageOS recursive resolver
* Only OCaml 4.07.0 and above are supported
* Multicast DNS has been dropped for now
* A client using async from JS has not been implemented yet

### DIFF
--- a/packages/dns-certify/dns-certify.4.0.0/opam
+++ b/packages/dns-certify/dns-certify.4.0.0/opam
@@ -8,7 +8,7 @@ bug-reports: "https://github.com/mirage/ocaml-dns/issues"
 license: "BSD2"
 
 depends: [
-  "dune" {build & >= "1.2.0"}
+  "dune" {>= "1.2.0"}
   "ocaml" {>= "4.07.0"}
   "dns" {= version}
   "dns-tsig" {= version}
--- a/packages/dns-certify/dns-certify.4.0.0/opam
+++ b/packages/dns-certify/dns-certify.4.0.0/opam
@@ -1,0 +1,45 @@
+opam-version: "2.0"
+maintainer: "team AT robur dot io"
+authors: ["Hannes Mehnert <hannes@mehnert.org>"]
+homepage: "https://github.com/mirage/ocaml-dns"
+doc: "https://mirage.github.io/ocaml-dns/"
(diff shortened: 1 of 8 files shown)


hannesm:
I don't quite understand the travis 4.08.0 failure tbh, first it finds a solution what to install and then it complains about constraints. I suspect some dependency is not yet 4.08-ready!? (but turns out, I use this release on a 4.08 switch, hmmm)

kit-ty-kate:
The CI failure should be fixed by https://github.com/ocaml/ocaml-ci-scripts/pull/306. All good otherwise, thanks!

hannesm:
thanks @kit-ty-kate! :)

PR #14689, opened 2019-08-17
[new release] conduit, conduit-lwt, conduit-async, conduit-mirage and conduit-lwt-unix (2.0.0)

avsm:
A network connection establishment library

- Project page: <a href="https://github.com/mirage/ocaml-conduit">https://github.com/mirage/ocaml-conduit</a>
- Documentation: <a href="https://mirage.github.io/ocaml-conduit/">https://mirage.github.io/ocaml-conduit/</a>

##### CHANGES:

* lwt-unix: obtain client IP correctly when using TLS connections (mirage/ocaml-conduit#277 @victorgomes)
* lwt-unix: replace the dune/ocaml file with a `(select)` build form.
  This avoids invoking `ocamlfind` from the build, and fits in with the
  rest of dune builds much more naturally (@avsm).
* lwt-unix: force callers to give a custom callback `on_exn` in case of exceptions
  to avoid random crashes (mirage/ocaml-conduit#261 @kit-ty-kate)
* mirage: use `dns-client>=4.0.0` which is the `udns` implementation (mirage/ocaml-conduit#290 @hannesm)
* mirage: rename `mirage-conduit` to `conduit-mirage` to fit the naming structure
  of this library suite more.  All new users of Mirage should use `conduit-mirage`,
  and migrating should involve simply swapping the name in your `dune` and `opam`
  files (mirage/ocaml-conduit#302 @hannesm @avsm)
* async: expose `verify_mode` correctly in `Conduit_async` (mirage/ocaml-conduit#298 @brendanlong)

PR #14690, opened 2019-08-18
min bound on calculon-web sequence

avsm:
due to Sequence.flat_map and find

found in revdeps for #14689


avsm:
Solver timeouts, but it does compile now, so merging

PR #14703, opened 2019-08-20
simplify constraints on the mmap package

UnixJunkie:
the dune version constraint should be enough


kit-ty-kate:
Mmh, I'm not too sure it would be even valid. Are you sure this package doesn't have a minimal ocaml version required?

UnixJunkie:
I meant, it relies on dune, and dune has an ocaml version constraint already (>= 4.02).
But this is not so important, you can close if you want.

kit-ty-kate:
It's fine, I can merge this.

PR #14708, opened 2019-08-20
Camlp5 logger syntax extension 0.4.2

Kakadu:
It will not be very popular but it is a dependency for stuff that I actually want to pulish in opam.

### DIFF
--- a/packages/logger-p5/logger-p5.0.4.2/opam
+++ b/packages/logger-p5/logger-p5.0.4.2/opam
@@ -17,7 +17,7 @@ bug-reports: "https://github.com/dboulytchev/logger/issues"
 
 depends: [
   "ocaml" { >= "4.07.0" }
-  "ocamlfind" 
+  "ocamlfind" {build}
   "camlp5"
 ]
 build: [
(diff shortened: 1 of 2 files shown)


Kakadu:
Mac is failing but it is temporary inconvenience. 

Ready to review.

packages/logger/logger.0.4.2/opam
```diff
@@ -0,0 +1,36 @@
+opam-version: "2.0"
+
+name: "logger-p5"
```
mseri: The `name` and `version` fields are redundant, `opam` will replace them with what infers from the path.
Moreover, is the name `logger` (as in the path) or `logger-p5` (as here)?

Kakadu: I made a transition to `logger-p5`. Thanks

packages/logger-p5/logger-p5.0.4.2/opam
```diff
@@ -0,0 +1,34 @@
...
+"""
+
+maintainer: "kakadu.hafanana@gmail.com"
+authors: "https://github.com/dboulytchev"
+homepage: "https://github.com/dboulytchev/logger"
+bug-reports: "https://github.com/dboulytchev/logger/issues"
+
+depends: [
+  "ocaml" { >= "4.07.0" }
+  "ocamlfind"
```
kit-ty-kate:
```suggestion
  "ocamlfind" {build}
```

kit-ty-kate:
Sounds ok. Thanks

PR #14714, opened 2019-08-21
Fix lablqml.0.6.2

kit-ty-kate:
cc @Kakadu @mseri 

This PR fixes several flaws with `lablqml.0.6.2`:
* The command prefixed by `PATH` shouldn't be needed outside of the configure part. I had a look at it and if discover (https://github.com/Kakadu/lablqml/blob/master/lib/config/discover.ml) fails with this PATH, the other commands will do to.
* `dune build -p` was missing for the case with `PATH`
* `build & with-test` is redundant I think (or does nothing, I'm not quite sure)

There are several things I'm not happy with. The package doesn't use dune the correct way:
* lablqml and ppx_qt should be either two separate packages or ppx_qt should have been `lablqml.ppx` or something. Currently this packages calls `dune install` which shouldn't be the case. opam files should be in the root directory, install files shouldn't be in the repository, there shouldn't be any `dune-project` in subdirectories. If it was done that way, dune install could have been avoided and opam would have just taken the install files generated in the root directory.
* In my opinion the PATH differentiation should be done by `discover.ml` and not by opam. It's a huge pain to read and a duplication which can leads to bugs (and there was since on one side make was called and on the other `dune build -p` was called.
* Are you sure ppxlib <= 0.9.0 is required? What if they made a micro release?

### DIFF
--- a/packages/lablqml/lablqml.0.6.2/opam
+++ b/packages/lablqml/lablqml.0.6.2/opam
@@ -9,29 +9,17 @@ tags: [ "gui" "ui" "qt" ]
 build: [
   ["sh" "-exc" "PATH=/usr/lib64/qt5/bin:/usr/lib/qt5/bin:$PATH ./configure"]
       {os-distribution = "alpine" | os-distribution = "centos" | os-distribution = "fedora" }
-  ["sh" "-exc" "PATH=/usr/lib64/qt5/bin:/usr/lib/qt5/bin:$PATH make"]
-      {os-distribution = "alpine" | os-distribution = "centos" | os-distribution = "fedora" }
-  ["sh" "-exc" "PATH=/usr/lib64/qt5/bin:/usr/lib/qt5/bin:$PATH make demos"]
-      { (os-distribution = "alpine"  | os-distribution = "centos" | os-distribution = "fedora") & with-test }
-
   ["./configure"]
       { os-distribution != "alpine" & os-distribution != "centos" & os-distribution != "fedora" }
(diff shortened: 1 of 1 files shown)


mseri:
@Kakadu can you please upstream the fixes and the suggestions from @kit-ty-kate for a future release?

packages/lablqml/lablqml.0.6.2/opam
```diff
@@ -9,29 +9,17 @@ tags: [ "gui" "ui" "qt" ]
...
-      {os-distribution = "alpine" | os-distribution = "centos" | os-distribution = "fedora" }
-  ["sh" "-exc" "PATH=/usr/lib64/qt5/bin:/usr/lib/qt5/bin:$PATH make demos"]
-      { (os-distribution = "alpine"  | os-distribution = "centos" | os-distribution = "fedora") & with-test }
-
   ["./configure"]
       { os-distribution != "alpine" & os-distribution != "centos" & os-distribution != "fedora" }
-  ["dune" "build" "-p" "lablqml,ppx_qt"]
-      { os-distribution != "alpine" & os-distribution != "centos" & os-distribution != "fedora" }
+  ["dune" "build" "-p" "lablqml,ppx_qt" "-j" jobs]
   [make "demos"]
```
mseri: Have you removed `{with-test}` on purpose?

kit-ty-kate: oops indeed, my bad

PR #14723, opened 2019-08-22
Package zarith.1.9

antoinemine:
### `zarith.1.9`
Implements arithmetic and logical operations over arbitrary-precision integers
The Zarith library implements arithmetic and logical operations over
arbitrary-precision integers. It uses GMP to efficiently implement
arithmetic over big integers. Small integers are represented as Caml
unboxed integers, for speed and space economy.

---
* Homepage: https://github.com/ocaml/Zarith
* Source repo: git+https://github.com/ocaml/Zarith.git
* Bug tracker: https://github.com/ocaml/Zarith/issues

---

### DIFF
--- a/packages/zarith/zarith.1.9/opam
+++ b/packages/zarith/zarith.1.9/opam
@@ -25,7 +25,6 @@ build: [
 install: [
   [make "install"]
 ]
-remove:  ["ocamlfind" "remove" "zarith"]
 depends: [
   "ocaml"
   "ocamlfind"
@@ -39,7 +38,6 @@ The Zarith library implements arithmetic and logical operations over
 arbitrary-precision integers. It uses GMP to efficiently implement
 arithmetic over big integers. Small integers are represented as Caml
 unboxed integers, for speed and space economy."""
-flags: light-uninstall
 url {
   src: "https://github.com/ocaml/Zarith/archive/release-1.9.tar.gz"
   checksum: [
--- a/packages/zarith/zarith.1.9/opam
+++ b/packages/zarith/zarith.1.9/opam
@@ -1,0 +1,49 @@
+opam-version: "2.0"
+maintainer: "Xavier Leroy <xavier.leroy@inria.fr>"
+authors: [
+  "Antoine Miné"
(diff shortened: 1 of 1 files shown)


Leonidas-from-XIV:
It looks like the `configure` script, line 258 is invalid and fails.

antoinemine:
I see... It looks like we added some bash-specific code in ./configure, but Debian/Ubuntu interprets #!/bin/sh in a strict way (it works on my Gentoo, so I did not catch it). I could replace it with a sed code or force #!/bin/bash, not sure which one is the most portable. The bash-specific code is used only for Cygwin btw.

Leonidas-from-XIV:
IMHO it would make the most sense to make a patch release with of zarith 1.9.1 or so to set that to `#!/bin/bash`. I don't think it is a great idea to do some sed trickery in OPAM since that only benefits OPAM users, but people installing the tarball on Debian or Ubuntu (e.g. packagers) will run into the exact same issue.

Debian and Ubuntu use `dash` as `/bin/sh` so they fail on bashisms, but the CI runs on Debian and also triggers that line so it doesn't seem to exclusively affect Cygwin.

hhugo:
will fix https://github.com/ocaml/Zarith/issues/54

packages/zarith/zarith.1.9/opam
```diff
@@ -0,0 +1,49 @@
...
+  ] {os = "macos"}
+  [make]
+]
+install: [
+  [make "install"]
+]
+remove:  ["ocamlfind" "remove" "zarith"]
+depends: [
+  "ocaml"
+  "ocamlfind"
```
kit-ty-kate:
```suggestion
  "ocamlfind" {build}
```

kit-ty-kate:
The `remove` field is unnecessary since opam 2.0. Could you return the fix in your opam file upstream?

antoinemine:
Thanks for the feedback. I've made a 1.9.1 bug-fix release and opam-publish-ed it as #14753

kit-ty-kate:
This one can be closed then.

Issue #14650, opened 2019-08-08
Set the last supported OCaml version for these camlp4-dependent packages to 4.08


XVilka:
I exported the published packages list that still rely on the camlp4 (I removed all `pa_*` packages as obsolete, and some of the packages that dropped camlp4 already, but not yet released, like coq or cocinelle):
* [abt](http://opam.ocaml.org/packages/abt)
* [archimedes](http://opam.ocaml.org/packages/archimedes/archimedes.0.4.19/)>=0.4.15
* [batsh](http://opam.ocaml.org/packages/batsh/batsh.0.0.6/)build & >=0.0.6
* [bisect](http://opam.ocaml.org/packages/bisect)
* [bolt](http://opam.ocaml.org/packages/bolt)
* [bookaml](http://opam.ocaml.org/packages/bookaml)
* [CamlGI](http://opam.ocaml.org/packages/CamlGI)
* [config-file](http://opam.ocaml.org/packages/config-file)
* [custom_printf](http://opam.ocaml.org/packages/custom_printf)
* [deriving](http://opam.ocaml.org/packages/deriving)
* [deriving-ocsigen](http://opam.ocaml.org/packages/deriving-ocsigen)
* [devkit](http://opam.ocaml.org/packages/devkit)build
* [dyntype](http://opam.ocaml.org/packages/dyntype)
* [enumerate](http://opam.ocaml.org/packages/enumerate)
* [erm_xmpp](http://opam.ocaml.org/packages/erm_xmpp)
* [estring](http://opam.ocaml.org/packages/estring)
* [extprot](http://opam.ocaml.org/packages/extprot/extprot.1.6.0/)build & >=1.3.0
* [extunix](http://opam.ocaml.org/packages/extunix)build
* [faillib](http://opam.ocaml.org/packages/faillib)
* [format](http://opam.ocaml.org/packages/format)
* [froc](http://opam.ocaml.org/packages/froc)
* [gettext](http://opam.ocaml.org/packages/gettext)
* [glsurf](http://opam.ocaml.org/packages/glsurf)build
* [hardcaml-yosys](http://opam.ocaml.org/packages/hardcaml-yosys)
* [haxe](http://opam.ocaml.org/packages/haxe)build
* [heptagon](http://opam.ocaml.org/packages/heptagon)build
* [herelib](http://opam.ocaml.org/packages/herelib)
* [integration1d](http://opam.ocaml.org/packages/integration1d)build
[...]

PR #14737, opened 2019-08-25
Label ocveralls and bisect as deprecated

aantron:
...in favor of package `bisect_ppx`.

- `ocveralls` is [archived](https://github.com/sagotch/ocveralls). Its functionality was integrated into Bisect_ppx in https://github.com/aantron/bisect_ppx/pull/176.
- `bisect` is the original source of `bisect_ppx`, but it is [now deprecated](https://github.com/gasche/bisect#deprecation) in favor of the latter. See also https://github.com/gasche/bisect/issues/2#issuecomment-506602009.

I did `opam search coverage`, and it would have helped to settle on `bisect_ppx` faster if these packages had been marked deprecated in the synopsis.

cc @gasche, @sagotch

### DIFF
--- a/packages/bisect/bisect.1.3.1/opam
+++ b/packages/bisect/bisect.1.3.1/opam
@@ -18,8 +18,10 @@ depends: [
   "camlp4"
   "ocamlbuild" {build}
 ]
-synopsis: "Code coverage tool for the OCaml language"
+synopsis: "Code coverage tool for the OCaml language (deprecated)"
 description: """
+Bisect is deprecated in favor of package bisect_ppx.
+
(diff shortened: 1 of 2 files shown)


XVilka:
It's an interesting idea, may be can be done for more deprecated packages, e.g. from https://github.com/ocaml/opam-repository/issues/14650 list (not all, only some of them though).

aantron:
I'll do it for lwt_camlp4 from that list, since I'm in charge of it. Also lwt_log.

aantron:
https://github.com/ocaml/opam-repository/pull/14744

Looks like I already did this for lwt_camlp4 and lwt_log, but I found another package of mine, namespaces, which needed to be marked.

PR #14739, opened 2019-08-25
Package fileutils.0.6.0

gildor478:
### `fileutils.0.6.0`
API to manipulate files (POSIX like) and filenames

---
* Homepage: http://ocaml-fileutils.forge.ocamlcore.org/
* Source repo: git+https://github.com/gildor478/ocaml-fileutils.git
* Bug tracker: https://github.com/gildor478/ocaml-fileutils/issues

---

### DIFF
--- a/packages/fileutils/fileutils.0.6.0/opam
+++ b/packages/fileutils/fileutils.0.6.0/opam
@@ -15,7 +15,6 @@ depends: [
   "oasis" {build}
   "ounit" {with-test & >= "2.0.0"}
 ]
-flags: light-uninstall
 build: [
   ["ocaml" "setup.ml" "-configure" "--prefix" prefix]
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
The `remove` field is unnecessary since opam 2.0. Your call to the configure script was also wrong (I think) since when `-t` is passed, it will call `./configure` twice.
Could you return the fixes if the opam file upstream?

gildor478:
853a37f86337ae626e314960d0c8a8591da34eca for the change to the "remove" field.

The call to the configure script is actually OK, it will keep former values (i.e. the --prefix on the line before) and enables the test when -t is passed.

kit-ty-kate:
Not it's totally fine, upstreaming just helps keep consistency and for future releases. Thanks a lot

PR #14740, opened 2019-08-26
frama-c: remove constraint on alt-ergo 2.3.0

maroneze:
Remove constraint to allow using alt-ergo 2.3.0 (which produces better results overall), despite the fact that in a few rare cases the result may be worse. This is a net gain overall.


kit-ty-kate:
Thanks

PR #14758, opened 2019-08-30
Package Pyms.0.2

jdrprod:
### `Pyms.0.2`
OCaml typedefinition to python3 transpiler
Pyms is a transpiler. It produces clean and reliable python modules from OCaml type definitions.

More than a transpiler, Pyms is also a language extension built on top of python allowing you to define OCaml-like variants and use them from a classical python source code.

---
* Homepage: https://github.com/jdrprod/Pym-s
* Source repo: git+https://github.com/jdrprod/Pym-s.git
* Bug tracker: https://github.com/jdrprod/Pym-s/issues

---

### DIFF
--- a/packages/Pyms/Pyms.0.2/opam
+++ b/packages/Pyms/Pyms.0.2/opam
@@ -1,0 +1,20 @@
+opam-version: "2.0"
+synopsis: "OCaml typedefinition to python3 transpiler"
+description: """
+Pyms is a transpiler. It produces clean and reliable python modules from OCaml type definitions.
+
+More than a transpiler, Pyms is also a language extension built on top of python allowing you to define OCaml-like variants and use them from a classical python source code.
+"""
+maintainer: "ARTHUR CORRENSON arthur.correnson@univ-tlse3.fr"
+authors: "ARTHUR CORRENSON arthur.correnson@univ-tlse3.fr"
+license: "MIT"
+homepage: "https://github.com/jdrprod/Pym-s"
+bug-reports: "https://github.com/jdrprod/Pym-s/issues"
+dev-repo: "git+https://github.com/jdrprod/Pym-s.git"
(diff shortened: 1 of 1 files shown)


mseri:
The `url` stanza is missing. Sometimes `opam-publish` omits it because it cannot find a tag on the repository. It should have printed a warning about virtual packages when called.

packages/Pyms/Pyms.0.2/opam
```diff
@@ -0,0 +1,20 @@
...
+
+More than a transpiler, Pyms is also a language extension built on top of python allowing you to define OCaml-like variants and use them from a classical python source code.
+"""
+maintainer: "ARTHUR CORRENSON arthur.correnson@univ-tlse3.fr"
+authors: "ARTHUR CORRENSON arthur.correnson@univ-tlse3.fr"
+license: "MIT"
+homepage: "https://github.com/jdrprod/Pym-s"
+bug-reports: "https://github.com/jdrprod/Pym-s/issues"
+dev-repo: "git+https://github.com/jdrprod/Pym-s.git"
+depends: [ "ocaml" "ocamlfind" "dune"]
```
mseri:
```suggestion
depends: [ "ocaml" "dune" {>= "1.9.0"} "menhir" "ocaml-compiler-libs"]
```
From a brief look at the code, I think you got the dependencies wrong. The CI will likely point out additional missing bounds

packages/Pyms/Pyms.0.2/opam
```diff
@@ -0,0 +1,20 @@
...
+build: [
```
mseri: You use vanilla `dune` from what I can see. I suggest to remove the install stanza, and replace the build one with the default dune rule:
```
build: [["dune" "build" "-p" name "-j" jobs]]
```

If this fails, it usually means that there is some cleanup needed in the dune files to fix the library and pacakges names.

PR #14759, opened 2019-08-30
Add package usb.1.3.1

letoh:


packages/usb/usb.1.3.1/opam
```diff
@@ -0,0 +1,50 @@
...
+
+build: [
+  ["./configure"]
+  ["./configure" "--enable-tests"] {with-test}
+  [make]
+  [make "doc"] {with-doc}
+  [make "test"] {with-test}
+]
+
+remove: [["ocamlfind" "remove" "usb"]]
```
mseri: This is not necessary with opam2 (and with it you can also remove the flags line)

packages/usb/usb.1.3.1/opam
```diff
@@ -0,0 +1,50 @@
...
+authors: "Jeremie Dimino <jeremie@dimino.org>"
+maintainer: "letoh"
+license: "BSD-3-Clause"
+homepage: "https://github.com/letoh/ocaml-usb"
+bug-reports: "https://github.com/letoh/ocaml-usb/issues"
+dev-repo: "git+https://github.com/letoh/ocaml-usb.git"
+
+depends: [
+  "ocaml" {>= "4.01.0"}
+  "ocamlfind"
```
mseri: Do you use this at runtime or is it a build dependency?

packages/usb/usb.1.3.1/opam
```diff
@@ -0,0 +1,50 @@
+opam-version: "2.0"
+name: "usb"
+version: "1.3.1"
```
mseri: Name and version are redundant, opam will override them with what comes from the path. Can you remove them?

letoh: I will remove the two lines in the next commit. Thank you!

packages/usb/usb.1.3.1/opam
letoh: Oasis needs it, so maybe I should leave it to oasis? or just add the build filter?

mseri:
Thanks. It looks good now, all green

letoh:
Thank you!

PR #14763, opened 2019-09-01
Add package glMLite version 0.03.52

fccm:


packages/glMLite/glMLite.0.03.52/opam
```diff
@@ -0,0 +1,34 @@
...
+  "ocamlfind" {build}
+]
+depexts: [
+  ["freeglut3-dev"] {os-family = "debian"}
+  ["libfreeglut-devel"] {os-family = "mageia"}
+  ["libjpeg-devel"] {os-family = "mageia"}
+  ["homebrew/x11/freeglut"] {os = "macos" & os-distribution = "homebrew"}
+]
+install: [make "install" "PREFIX=%{lib}%/glMLite"]
+remove: [["ocamlfind" "remove" "glMLite"]]
```
kit-ty-kate:
```suggestion
```
The `remove` section is unnecessary since opam 2.0

packages/glMLite/glMLite.0.03.52/opam
```diff
@@ -0,0 +1,34 @@
...
+dev-repo: "git+https://github.com/fccm/glMLite.git"
+bug-reports: "https://github.com/fccm/glMLite/issues"
+
+synopsis: "OpenGL bindings for OCaml"
+tags: [ "bindings" "opengl" "graphics" "3D" ]
+description: """
+Provide bindings for GL, Glu and Glut, GLE, FTGL, and also some
+small image loader modules for different image file formats.
+"""
+flags: light-uninstall
```
kit-ty-kate:
```suggestion
```
This becomes unnecessary if the `remove` section is deleted.

kit-ty-kate:
```
# ocamlmklib  -o  vertArray  varray.wrap.o -lGL -L"/home/opam/.opam/4.08/lib/ocaml" -lbigarray
# /usr/bin/ld: cannot find -lbigarray
# make[1]: *** [Makefile.VArray:64: dllvertArray.so] Error 2
```
This package doesn't seem to be compatible with the latest version of OCaml (4.08) as it assumes the `bigarray` library is still external to the standard library. Could you either restrict to older version of the OCaml compiler or fix this in your package?

fccm:
I'm still using ocaml 4.07.1 but I can update to 4.08.1 if needed.
If I remove the -lbigarray flag, will it be still compatible with ocaml 4.07.1?
Is there a way to make it compatible with both versions? Or should I make 2 packages: one for 4.07.1 and another one for 4.08.1?

kit-ty-kate:
> I'm still using ocaml 4.07.1 but I can update to 4.08.1 if needed.
> If I remove the -lbigarray flag, will it be still compatible with ocaml 4.07.1?
> Is there a way to make it compatible with both versions? Or should I make 2 packages: one for 4.07.1 and another one for 4.08.1?

I don't know you can try with 4.08 and 4.07 using opam switches. But it seems easy enough to do in make by just comparing the version number of OCaml to figure out if it needs a flag or not. Having two different packages (0.03.52 and 0.03.53 for instance) seem reasonable as well.

packages/glMLite/glMLite.0.03.52/opam
```diff
@@ -0,0 +1,32 @@
...
+synopsis: "OpenGL bindings for OCaml"
+tags: [ "bindings" "opengl" "graphics" "3D" ]
+description: """
+Provide bindings for GL, Glu and Glut, GLE, FTGL, and also some
+small image loader modules for different image file formats.
+"""
+build: make
+
+depends: [
+  "ocaml" {<= "4.07.1"}
```
kit-ty-kate:
```suggestion
  "ocaml" {< "4.08.0"}
```

kit-ty-kate:
Thanks!

PR #14767, opened 2019-09-02
Package dolmen.0.4.1

Gbury:
### `dolmen.0.4.1`
A parser library
Dolmen is a parser library. It currently targets languages used in automated theorem provers,
but may be extended ot other domains.

Dolmen provides functors that takes as arguments a representation of terms and statements,
and returns a module that can parse files (or streams of tokens) into the provided representation
of terms or statements. This is meant so that Dolmen can be used as a drop-in replacement of existing
parser, in order to factorize parsers among projects.

Additionally, Dolmen also provides a standard implementation of terms and statements that cna be
used ot instantiate its parsers.

---
* Homepage: https://github.com/Gbury/dolmen
* Source repo: git+https://github.com/Gbury/dolmen.git
* Bug tracker: https://github.com/Gbury/dolmen/issues

---
:camel: Pull-request generated by opam-publish v2.0.0

PR #14769, opened 2019-09-02
archsat.1.0: add missing lower bound on sequence

mseri:
Should fix the revdeps error in https://github.com/ocaml/opam-repository/pull/14767


kit-ty-kate:
Thanks!

PR #14773, opened 2019-09-02
Normalise licenses to point to valid SPDX ids

AltGr:
Ref. ocaml/opam#3976
This was done semi-manually. I didn't fix invalid licenses, just
normalised the ones that fit.


thomasblanc:
So, I think the CI should use xargs :-)

packages/conf-linux-libc-dev/conf-linux-libc-dev.0/opam
```diff
@@ -3,7 +3,7 @@ maintainer: "Markus W. Weissmann <markus.weissmann@in.tum.de>"
 authors: [ "Markus W. Weissmann <markus.weissmann@in.tum.de>" ]
 homepage: "https://kernel.org/"
 bug-reports: "https://github.com/ocaml/opam-repository/issues"
-license: [ "GPL 2.0" ]
+license: [ GPL-2.0-only ]
```
thomasblanc: You missed the quotes here

AltGr:
Probably not worth the work... but it would be interesting for PRs that _modify_ existing packages if Camelus could do a before/after check and diff that: it would ignore warnings/errors that were already there and you didn't touch, while pointing what you fixed or broke.

 @thomasblanc thanks for spotting it!

AltGr:
I feel sorry for the CI :D
It has a "replay" button, but no "cancel" :/

thomasblanc:
> Probably not worth the work... but it would be interesting for PRs that _modify_ existing packages if Camelus could do a before/after check and diff that: it would ignore warnings/errors that were already there and you didn't touch, while pointing what you fixed or broke.

I disagree. First, because it would add weight on Camelus. Also, and more importantly, it would remove an incentive to maintainers to fix their bad packages. This is only problematic on big prs by maintainers which don't happen every day.

AltGr:
Merging this, besides making licences much cleaner, would allow us to:
- gather some stats on the licenses used (because why not!)
- automatically link to the license text from the package pages on opam.ocaml.org/packages

AltGr:
If there is no objection, let's merge ?

kit-ty-kate:
Thanks a lot for this work

PR #14784, opened 2019-09-05
Package libsvm.0.10.0

pveber:
### `libsvm.0.10.0`
LIBSVM bindings for OCaml
LIBSVM bindings for OCaml

LIBSVM-OCaml is an OCaml library with bindings to the LIBSVM
library, which is a library for Support Vector Machines.
Support Vector Machines are used to create supervised
learning models for classification and regression problems
in machine learning.

---
* Homepage: https://github.com/pveber/libsvm-ocaml
* Source repo: git+https://github.com/pveber/libsvm-ocaml.git
* Bug tracker: https://github.com/pveber/libsvm-ocaml/issues

---

### DIFF
--- a/packages/libsvm/libsvm.0.10.0/opam
+++ b/packages/libsvm/libsvm.0.10.0/opam
@@ -32,6 +32,7 @@ depends: [
   "lacaml"
   "ocaml" {>= "4.03.0"}
   "stdio"
+  "core_kernel" {with-test}
 ]
 url {
   src: "https://github.com/pveber/libsvm-ocaml/archive/0.10.0.tar.gz"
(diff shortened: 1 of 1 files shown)


UnixJunkie:
are those new bindings?
Because the previous ones were terribly bad and crashing on me.
It would be super nice to have bindings to liblinear as well.

kit-ty-kate:
Thanks!

pveber:
Thanks a lot @kit-ty-kate!

@UnixJunkie I noticed like you that the bindings implemented in libsvm-ocaml would trigger segfaults. It was due to an incorrect interaction with the GC, so I rewrote a good portion of the bindings, trying my best to keep the feeling of the API. Hopefully the situation should be better now, although I can't say they have been tested much.

UnixJunkie:
Nice to have a SVM package in OCaml.
I am happy to read that someone is taking care of this package now.

PR #14787, opened 2019-09-06
added packages/genspir/genspir.1.0.0

UnixJunkie:
switched from oasis to dune

### DIFF
--- a/packages/genspir/genspir.1.0.0/opam
+++ b/packages/genspir/genspir.1.0.0/opam
@@ -23,6 +23,6 @@ The Mathematical Intelligencer, 19, 5-11,
 http://www.math.vanderbilt.edu/~esaff/texts/161.pdf
 """
 url {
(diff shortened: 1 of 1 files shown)


UnixJunkie:
I will use a proper tag name once all tests pass.

UnixJunkie:
This is the final tag.

kit-ty-kate:
Just an off topic general advice, I've seen a lot of your packages using dune but without any `dune-project` file in the repository. It really helps to know which minimal version of dune you require for your packages. Plus it's going to be mandatory for dune 2.0. I'd suggest to add them to your projects for future releases

kit-ty-kate:
Thanks!

UnixJunkie:
I really don't like putting automatically generated files under VC...

kit-ty-kate:
> I really don't like putting automatically generated files under VC...

it's not automatically generated. It's only generated once if it doesn't already exist. It's not going to change by itself.

UnixJunkie:
Ok, I'll consider it.

PR #14790, opened 2019-09-06
forgotten dependecies for qocamlbrowser

Kakadu:


Kakadu:
The only issue is about Qt installation on MacOS which is not related to these particular changes.

Ready to be merged

kit-ty-kate:
Thanks

PR #14792, opened 2019-09-08
Make unavailable uninstallable old versions

Kakadu:
They don't have reverse dependencies so it is safe.

Discussed at https://discuss.ocaml.org/t/question-when-it-is-ok-to-remove-some-packages-versions-from-opam/4321


kit-ty-kate:
Fair enough. Thanks

PR #14800, opened 2019-09-10
Tag some packages as accepting CI failures on Debian Sid

kit-ty-kate:
Some packages will always fail in CI because either they need some specific version of a package or because some external dependency is not installable in the CI.

My proposal (implemented in `opam-health-check` via the following two commits: https://github.com/kit-ty-kate/opam-health-check/commit/721699fe2beddcce21c830071f2ab25bbda4176c https://github.com/kit-ty-kate/opam-health-check/commit/3462fa92818e8d20151748712101f38f6050e345), is to tag the irredeemable packages to accept the failures, changing them from an hard failure to an internal failure. This change simplifies and cleans the view in [ci.ocamllabs.io](http://check.ocamllabs.io) and allows us to see which packages really fail. `opam-health-check` only uses Debian Sid so that's why only debian-sid is tagged.

For now this is only a handful of packages because checking whether or not the failure can be skipped is a time consuming process in some cases but I'll add more of them in the future.

Does anyone have any thoughts on this? cc @mseri @avsm @dra27 @samoht @AltGr @rjbou @hannesm

### DIFF
--- a/packages/aacplus/aacplus.0.2.1/opam
+++ b/packages/aacplus/aacplus.0.2.1/opam
@@ -12,6 +12,7 @@ depexts: [
   ["libaacplus"] {os-distribution = "gentoo"}
   ["libaacplus"] {os = "freebsd"}
 ]
+x-ci-accept-failures: ["debian-unstable"]
 install: [make "install"]
 synopsis:
   "Bindings for the aacplus library which provides functions for decoding AAC audio files"
--- a/packages/aacplus/aacplus.0.2.2/opam
+++ b/packages/aacplus/aacplus.0.2.2/opam
@@ -16,6 +16,7 @@ depexts: [
   ["libaacplus"] {os-distribution = "gentoo"}
   ["libaacplus"] {os = "freebsd"}
 ]
+x-ci-accept-failures: ["debian-unstable"]
 bug-reports: "https://github.com/savonet/ocaml-aacplus/issues"
 dev-repo: "git+https://github.com/savonet/ocaml-aacplus.git"
 synopsis:
--- a/packages/ahrocksdb/ahrocksdb.0.2.0/opam
+++ b/packages/ahrocksdb/ahrocksdb.0.2.0/opam
@@ -40,6 +40,7 @@ build: [
   ["dune" "build" "-p" name "-j" jobs]
   ["dune" "runtest"] {with-test}
 ]
+x-ci-accept-failures: ["debian-unstable"]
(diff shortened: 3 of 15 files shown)


avsm:
I'm fine with the idea of using tags for this (or even `x-` extension fields since they won't show up in the UI).  However, I would recommend using `-unstable` here to make it explicit that this is the fast moving branch.

A broader question though: why are we testing debian-unstable at all?  It seems more prudent to be regularly checking against debian-stable (the default target in the rest of the CI as the "first" check).

kit-ty-kate:
> I'm fine with the idea of using tags for this (or even `x-` extension fields since they won't show up in the UI). However, I would recommend using `-unstable` here to make it explicit that this is the fast moving branch.
> 
> A broader question though: why are we testing debian-unstable at all? It seems more prudent to be regularly checking against debian-stable (the default target in the rest of the CI as the "first" check).

`x-` extension fields seem indeed like a better idea. I'll do that shortly.

My reasoning behind checking debian-unstable is that it is a good middle point between all the distributions available that has up-to-date software but not too bleeding-edge. A copious amount of packages cannot be tested on debian-stable (or probably i should say couldn't because the new stable is "new" enough now) and this is better to test the latest versions of opam packages. Furthermore, ubuntu is based on unstable so we get that as well.

My intuition is that opam is mainly used by OCaml developers and that they usually prefer to have up-to-date software and use distributions which allow that. Maybe it's worth doing some regular run with debian-stable as well and diff between the two to improve quality for uses of opam in production, but since we already have mirage-ci to check that I feel like this can wait a bit more.

mseri:
I like the idea of tagging the packages with an `x-` field to cleanup the CI results. Also I think the idea of testing `debian-unstable`, or at least `debian-testing`, as a stable-ish bleeding-ish edge environment does not sound bad imho.

kit-ty-kate:
I've implemented both suggestions and reflected them in https://github.com/kit-ty-kate/opam-health-check/commit/af1fa328ad9c03d8cc6db8726654ef64d57e3d61

avsm:
How about just going with the latest Ubuntu point release? That reflects a snapshot of Debian's package, but at least a stable one. My experience with debian-unstable in CI is that, when it enters a period of breakage, it can stay broken for some time.  Which is fair enough, since it's their staging ground for major changes...

mseri:
That does not sound like a bad idea

kit-ty-kate:
I can switch to ubuntu later if it really is a problem. This new field will contain several distributions at the end anyway.

PR #14805, opened 2019-09-11
Fix some old warnings 24

AltGr:
Done with

    for f in $(opam admin lint 24 -l); do opam-ed -if "packages/${f%%.*}/$f/opam" 'prepend tags "org:ocamlpro"'; done

(and checking that all the packages started with `ocp`)
Note that this warning is now obsolete and could be removed (since no non-ocp packages hit it anymore).


kit-ty-kate:
Thanks!

PR #14807, opened 2019-09-11
irrlicht: add bound on OCaml version

damiendoligez:
`irrlicht.0.0.5` is not compatible with OCaml 4.09 because of the new `const` qualifier on the return type of `caml_named_value`. Upstream fix proposed: https://github.com/fccm/OCaml-Irrlicht/pulls

cc @fccm


fccm:
Isn't it possible to patch?
I will of course accept the pull-request, but it's quite annoying to make plenty of releases for very small changes, don't you think?

mseri:
I would not be against adding a `irrlicht.0.0.5-1` with the patch, in addition to this bound on the current package.

PR #14816, opened 2019-09-13
i3ipc 0.3

Armael:
cc @chimrod

### DIFF
--- a/packages/i3ipc/i3ipc.0.3/opam
+++ b/packages/i3ipc/i3ipc.0.3/opam
@@ -1,0 +1,31 @@
+opam-version: "2.0"
+maintainer: "armael@isomorphis.me"
+authors: "Armaël Guéneau"
+homepage: "https://github.com/Armael/ocaml-i3ipc"
+bug-reports: "https://github.com/Armael/ocaml-i3ipc/issues"
+license: "MIT"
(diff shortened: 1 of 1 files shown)


packages/i3ipc/i3ipc.0.3/opam
```diff
@@ -0,0 +1,31 @@
...
+dev-repo: "git+https://github.com/Armael/ocaml-i3ipc"
+doc: "https://armael.github.io/ocaml-i3ipc/0.3/"
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+]
+depends: [
+  "ocaml" {>= "4.03.0"}
+  "lwt"
+  "lwt_ppx"
+  "dune"
```
kit-ty-kate:
```suggestion
  "dune" {>= "1.8"}
```
The `dune-project` file in your sources says this package requires at least dune 1.8

Armael: Good catch. I've instead relaxed the constraint on the version in the dune-project; it now should work for dune >= 1.0, ie any version of dune.

Chimrod:
Merci !

I'll update my code with your new protocol_error printers as soon as your version will be publish on opam

kit-ty-kate:
Thanks a lot!

PR #14821, opened 2019-09-15
Make base available with ocaml 4.09.0

hhugo:
cc @xclerc


kit-ty-kate:
Same comment as with https://github.com/ocaml/opam-repository/pull/14822#issuecomment-531570409. Thanks!

PR #14822, opened 2019-09-15
ppx_deriving available with ocaml 4.09

hhugo:


kit-ty-kate:
cc @gasche

kit-ty-kate:
We don't have the CI infrastructure to test this but [ci.ocamllabs.io](http://check.ocamllabs.io) will in a few days.

kit-ty-kate:
In the meantime it should be fine. Thanks!

PR #14838, opened 2019-09-17
codept: 0.10.3 is fine with OCaml 4.09

Octachron:
This PR adjusts codept's upper bound for compatible OCaml version to 4.10 excluded .


kit-ty-kate:
The CI doesn't test against 4.09 for the moment but [ci.ocamllabs.io](http://check.ocamllabs.io) will in a few days. Thanks!

PR #14842, opened 2019-09-18
Release of OCaml 4.09.0

Octachron:
This is the full release of OCaml 4.09.0, on the accelerated track.
This release integrates one one-liner miscompilation fix compared to the second beta.

Some of the highlights in release 4.09 are:

- New optimisations, in particular for affine functions in matches, for instance
 ```ocaml
  type t = A | B | C
  let affine = function
    | A -> 4
    | B -> 3
    | C -> 2
```
- The `graphics` library was moved out of the compiler distribution.
- The `vmthread` library was removed.
- Support for compiler plugins was removed.
- Many bug fixes.

cc @gasche

### DIFF
--- a/packages/ocaml-variants/ocaml-variants.4.09.0+musl+flambda/opam
+++ b/packages/ocaml-variants/ocaml-variants.4.09.0+musl+flambda/opam
@@ -1,0 +1,51 @@
+opam-version: "2.0"
+synopsis:
+  "Official release 4.09.0, compiled with musl-gcc and with flambda activated"
+description:
+  "Requires musl-gcc to be installed (package musl on Arch Linux or musl-tools on Debian)"
+maintainer: "platform@lists.ocaml.org"
+authors: "Xavier Leroy and many contributors"
+homepage: "https://ocaml.org"
+bug-reports: "https://github.com/ocaml/ocaml/issues"
+dev-repo: "git://github.com/ocaml/ocaml"
+depends: [
+  "ocaml" {= "4.09.0" & post}
+  "base-unix" {post}
+  "base-bigarray" {post}
+  "base-threads" {post}
+]
+depexts: [
+  ["musl-tools"] {os-family = "debian"}
+  ["musl"] {os-distribution = "archlinux"}
+]
+conflict-class: "ocaml-core-compiler"
+flags: compiler
+setenv: CAML_LD_LIBRARY_PATH = "%{lib}%/stublibs"
+build: [
+  ["./configure" "--prefix=%{prefix}%"]
+    {os != "openbsd" & os != "freebsd" & os != "macos" & os != "linux"}
+  [
+    "./configure"
+    "--prefix=%{prefix}%"
+    "CC=musl-gcc -Os"
+    "ASPP=musl-gcc -c"
+    "--enable-flambda"
+    "--disable-graph-lib"
+  ] {os = "openbsd" | os = "freebsd" | os = "macos" | os = "linux"}
+  [make "-j%{jobs}%" {os != "cygwin"} "world"]
+  [make "-j%{jobs}%" {os != "cygwin"} "world.opt"]
+]
+install: [make "install"]
+url {
+  src: "https://github.com/ocaml/ocaml/archive/4.09.0.tar.gz"
+  checksum: "md5=76ac39570fc88b16fda2a94db7cd5cf3"
+}
+post-messages: [
+  "A failure in the middle of the build may be caused by build parallelism
+   (enabled by default).
+   Please file a bug report at https://github.com/ocaml/ocaml/issues"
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

**Comment by kit-ty-kate** [comment]
Merging this as soon as [check.ocamllabs.io](http://check.ocamllabs.io) finishes its latest run. It would be useful to see which packages are still missing and those two PRs got recently merged (https://github.com/ocaml/opam-repository/pull/14821 and https://github.com/ocaml/opam-repository/pull/14822) and probably unlock many of them. It should be over in a few hours.

**Comment by Octachron** [comment]
Thanks for the head-ups!

**Comment by kit-ty-kate** [comment]
It seems pretty good. If we choose to see at the largely used packages only `nocrypto` fails, and only because https://github.com/ocaml/ocaml/pull/8770 wasn't merged in beta1 or 2.

Concerning "key packages" not available on 4.09 yet, there are still a handful of them:
* `merlin` cc @let-def @trefis 
* `odoc` cc @aantron 
* `ocamlformat` (which should be unlocked once `odoc` is available but there is no ways to know for sure it works) cc @jberdine @gpetiot 
* `ocaml-freestanding` (for the mirage related packages depending on it) cc @mato 
* `utop` cc @diml 
* `batteries` cc @UnixJunkie @gasche 
* `js_of_ocaml` cc @hhugo 

Full detail of this run here: http://check.ocamllabs.io/run/1568558996-3394ff39238589fe9271e2531f6ae13e5f7aa2f6?comp=4.08&comp=4.09&available=4.08&available=4.09&show-diff-only=true&show-latest-only=true

**Comment by aantron** [comment]
We'll most likely release odoc 1.4.2 tomorrow, with 4.09 support (https://github.com/ocaml/odoc/issues/386). odoc `master` already has it, so it should be possible to test ocamlformat against odoc `master` even before the 1.4.2 release.

Thanks for the ping.

**Comment by Octachron** [comment]
Concerning `nocrypto`, I have just checked that it builds with the release version. But I also have no trouble building it with the 4.09.0+beta2: the fix was part of the second beta. So I am not sure what is going on?

The new version of utop released today is compatible with 4.09 .

**Comment by kit-ty-kate** [comment]
> Concerning `nocrypto`, I have just checked that it builds with the release version. But I also have no trouble building it with the 4.09.0+beta2: the fix was part of the second beta. So I am not sure what is going on?

Ah sorry I missed the PR introducing 4.09.0+beta2 (https://github.com/ocaml/opam-repository/pull/14809) and I couldn't see the tag for the beta on the commits that are supposed to fixed this problem. My bad.

> The new version of utop released today is compatible with 4.09 .

Oops, I missed that one, sorry for the noise

> We'll most likely release odoc 1.4.2 tomorrow, with 4.09 support ([ocaml/odoc#386](https://github.com/ocaml/odoc/issues/386)). odoc `master` already has it, so it should be possible to test ocamlformat against odoc `master` even before the 1.4.2 release.

Thanks a lot!

Merlin is probably the first package which will cause problems to users trying this new release right away and I'm not sure how long it's going to take for the next release but let's merge this now at least. Thanks a lot!

**Comment by pmetzger** [comment]
Didn't the fix for `utop` also go in a few hours ago?

**Comment by aantron** [comment]
The odoc release PR: https://github.com/ocaml/opam-repository/pull/14848

**Comment by UnixJunkie** [comment]
What is the opam switch I should install in order to test batteries with this upcoming OCaml release?
4.09.0+trunk?
I've never installed "experimental" OCaml compiler switches; sorry for the maybe stupid question.

**Comment by Octachron** [comment]
Since this PR was merged, you can simply use `opam switch create 4.09.0`. If you have any trouble with the migration, please feel free to ping me.

File: packages/ocaml-variants/ocaml-variants.4.09.0+musl+flambda/opam [hunk 1]
```diff
@@ -0,0 +1,51 @@
...
+build: [
+  ["./configure" "--prefix=%{prefix}%"]
+    {os != "openbsd" & os != "freebsd" & os != "macos" & os != "linux"}
+  [
+    "./configure"
+    "--prefix=%{prefix}%"
+    "CC=musl-gcc -Os"
+    "ASPP=musl-gcc -c"
+    "--enable-flambda"
+    "--disable-graph-lib"
```
copy: This is redundant since graph has been removed from OCaml 4.09.

copy: Fixed in https://github.com/ocaml/opam-repository/pull/14858

Octachron: Thanks!

UnixJunkie:
I think the problem with batteries was just the upper bound version constraint on ocaml. I sent a PR to update it.

PR #14843, opened 2019-09-18
[DO NOT MERGE] Move ocaml-variants.4.09.0+trunk to ocaml-variants.4.09.1+trunk

kit-ty-kate:
Following the release of OCaml 4.09.0, the package `ocaml-4.09.0+trunk` (and other +trunk variants for 4.09.0) is not valid anymore and needs to be moved to 4.09.1 as of https://github.com/ocaml/ocaml/commit/6207460e793f8c5538bfd8164fa3078f7300b57a

cc @talex5 @avsm

### DIFF
--- a/packages/ocaml-variants/ocaml-variants.4.09.1+trunk+afl/opam
+++ b/packages/ocaml-variants/ocaml-variants.4.09.1+trunk+afl/opam
@@ -1,8 +1,8 @@
 opam-version: "2.0"
-synopsis: "latest 4.09 development, with flambda activated"
+synopsis: "latest 4.09 development, with afl-fuzz instrumentation"
 maintainer: "platform@lists.ocaml.org"
 depends: [
-  "ocaml" {= "4.09.0" & post}
+  "ocaml" {= "4.09.1" & post}
   "base-unix" {post}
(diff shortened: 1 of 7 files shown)


kit-ty-kate:
@AltGr @rjbou is there a way to have a package versionned `4.09+trunk` instead or something like that, and not care about the minor part of the version? It seems to be a fragile pattern and maybe there is a way to fix it another way? I'd rather not add a package `ocaml.4.09.1` if it's not gonna be used by any stable release.

kit-ty-kate:
@thomasblanc could you take a look at the output from Camelus? It seems a bit buggy (the number of installable packages removed seems wrong and the ignored non-opam files as well)

thomasblanc:
> @thomasblanc could you take a look at the output from Camelus? It seems a bit buggy (the number of installable packages removed seems wrong and the ignored non-opam files as well)

I think this was an already existing bug that the "ignored files" section exposed. I'm checking tihs with @AltGr

thomasblanc:
After some investigation, it seems that Camelus is puzzled by moved files.

kit-ty-kate:
4.09.1 might be released at some point. If not the package can always be removed.

PR #14844, opened 2019-09-18
Travis: Use OCaml 4.09 and remove all redundant tests with mirage-ci

kit-ty-kate:
This PR makes use of OCaml 4.09 (added into ocaml-ci-scripts in https://github.com/ocaml/ocaml-ci-scripts/pull/311) and also removes every switches that are already tested by https://github.com/avsm/mirage-ci. Apart from MacOS, the only recent one that is not tested is the one with the ocaml compiler locally installed. Pre 4.02 compliers have also been removed.

This is a first step towards the complete removal of Travis from the CI of opam-repository.

cc @avsm

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,25 +8,9 @@ matrix:
   include:
   - os: osx
     osx_image: xcode10.1
-    env: OCAML_VERSION=4.07 OPAM_VERSION=2.0
-  - os: linux
     env: OCAML_VERSION=4.08 OPAM_VERSION=2.0
   - os: linux
-    env: OCAML_VERSION=4.07 OPAM_VERSION=2.0
-  - os: linux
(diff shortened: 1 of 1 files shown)


samoht:
Did `mirage-ci` run tests now? If I remember correctly that was one the differences between the two systems.

kit-ty-kate:
> Do `mirage-ci` run tests now? If I remember correctly that was one the differences between the two systems.

Yes it does since https://github.com/avsm/mirage-ci/commit/fd3b4562507ad9c38f519568717c05dab733e9a4

avsm:
I think it actually only runs them for revdeps and not for the actual package. Should fix that now that Travis is gone.

kit-ty-kate:
> I think it actually only runs them for revdeps and not for the actual package. Should fix that now that Travis is gone.

It's already in your commit I quoted. It is also ran for the actual package, just not in phase1, but for phase3 (test the modified/new packages on all the recent OCaml compilers)

PR #14850, opened 2019-09-19
Fixes for the duniverse.

xclerc:
This pull request contains new versions of existing
packages, expected to fix issues with the duniverse:

- `base_quickcheck.v0.12.1`;
- `core_kernel.v0.12.3`.

@avsm @diml

### DIFF
--- a/packages/base_quickcheck/base_quickcheck.v0.12.0/opam
+++ b/packages/base_quickcheck/base_quickcheck.v0.12.0/opam
@@ -17,7 +17,7 @@ depends: [
   "ppx_let"           {>= "v0.12" & < "v0.13"}
   "ppx_sexp_message"  {>= "v0.12" & < "v0.13"}
   "splittable_random" {>= "v0.12" & < "v0.13"}
-  "dune"              {>= "1.5.1"}
+  "dune"              {>= "1.5.1" & < "2.0.0"}
   "ppxlib"            {>= "0.5.0" & < "0.9.0"}
 ]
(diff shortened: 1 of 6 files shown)


diml:
Thanks! To be more precise, this is actually an incompatibility with Dune >= 2.0.0. This is related to cases where a file is both produced by a rule and present in the source tree. Jbuilder and Dune 1.x where not strict about this and it was often causing confusion. Dune 2.0.0 is strict about it and require the user to express the intent with `(mode fallback)` or `(mode promote)`

avsm:
Thanks! It might be worth upper bounding `dune < 2.0.0` for the older versions in this PR now while this is fresh in our heads, so that the older packages don't get selected for revdeps when dune 2.0 is released.

xclerc:
[These failures](https://ci.ocaml.org/ocaml/opam-repository/pr/14850?test=2%20Revdeps) are not directly related to this PR:
- `capnp.3.2.1` fails because of an unbound module (the fix is probably to add an upper bound on `core`);
- `tidy.1-4.9.30-0.1.1` fails because of a missing C header.

xclerc:
@avsm Sure; I just pushed https://github.com/ocaml/opam-repository/pull/14850/commits/2bc64f225b0409cc97326c8f36038a748232e0ac with these changes.

xclerc:
(The Travis error is actually a timeout.)

kit-ty-kate:
Thanks!

PR #14851, opened 2019-09-19
Add a dependency on dune-configurator for all packages that use dune.configurator

diml:
We are planning to split `dune.configurator` as a separate `dune-configurator` package in Dune 2.0.0. The `dune.configurator` name will still exist for backward compatibility reason, however packages that use it must declare an additional dependency on `dune-configurator` in order to be compatible with Dune >= 2.0.0.

In preparation of this, this PR adds an empty `dune-configurator.1.0.0` that depends on Dune < 2.0.0 and adds it as a dependency of all packages that use `dune.configurator`.

### DIFF
--- a/packages/fftw3/fftw3.0.8.2/opam
+++ b/packages/fftw3/fftw3.0.8.2/opam
@@ -14,7 +14,6 @@ build: [
     "env"
     "FFTW3_CFLAGS=-I /usr/local/include -L /usr/local/lib -I /usr/local/include -L /usr/local/lib"
     "dune"
-  "dune-configurator"
     "build"
     "-p"
     name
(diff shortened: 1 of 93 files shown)


diml:
The failures are unrelated.

packages/fftw3/fftw3.0.8.2/opam
```diff
@@ -14,6 +14,7 @@ build: [
     "env"
     "FFTW3_CFLAGS=-I /usr/local/include -L /usr/local/lib -I /usr/local/include -L /usr/local/lib"
     "dune"
+  "dune-configurator"
```
kit-ty-kate:
```suggestion
```
This shouldn't be there

diml: Actually, I believe it should:

```
$ opam source fft3.0.8.2
[fftw3.0.8.2] downloaded from cache at https://opam.ocaml.org/cache
Successfully extracted to /home/dim/code/fftw3.0.8.2
$ cd fftw3.0.8.2/
$ dune external-lib-deps @install
These are the external library dependencies in the default context:
- bigarray
- dune.configurator
- str
```

diml: Ah never-mind, I just realised that it was part of the build instructions...

kit-ty-kate:
Thanks!

PR #14853, opened 2019-09-19
Solo5 0.6.0

mato:
This adds Solo5 0.6.0 to OPAM. This release contains API breaking
changes, so we also add constraints on `{< "0.6.0"}` on existing
`mirage-solo5` releases. `ocaml-freestanding` is unaffected by the
changes.


mato:
Failure is due to the Genode bindings not being tested. Will have to cut an 0.6.1 :-(

PR #14854, opened 2019-09-19
Package tsdl_image.0.3.0

tokenrove:
### `tsdl_image.0.3.0`
SDL2_Image bindings to go with Tsdl
Tsdl_image provides bindings to SDL2_Image intended to be used with
Tsdl.

---
* Homepage: http://github.com/tokenrove/tsdl-image
* Source repo: git+https://github.com/tokenrove/tsdl-image.git
* Bug tracker: http://github.com/tokenrove/tsdl-image/issues

---

### DIFF
--- a/packages/tsdl_image/tsdl_image.0.3.0/opam
+++ b/packages/tsdl_image/tsdl_image.0.3.0/opam
@@ -1,0 +1,31 @@
+opam-version: "2.0"
+maintainer: "Julian Squires <julian@cipht.net>"
+authors: ["Julian Squires <julian@cipht.net>"]
+homepage: "http://github.com/tokenrove/tsdl-image"
+dev-repo: "git+https://github.com/tokenrove/tsdl-image.git"
+bug-reports: "http://github.com/tokenrove/tsdl-image/issues"
+tags: [ "bindings" "graphics" ]
+license: "BSD-3-Clause"
+depends: [
+  "ocaml" {>= "4.02"}
+  "ctypes" {>= "0.4.0"}
+  "ctypes-foreign"
+  "tsdl" {>= "0.9.0"}
+  "result"
+  "dune" {build & >= "1.11.0"}
+]
+depexts: [
+  ["libsdl2-image-dev"] {os-family = "debian"}
+  ["sdl2_image"] {os-distribution = "homebrew" & os = "macos"}
+  ["sdl2_image"] {os-distribution = "arch"}
+]
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "build" "test/what.png" "test/test.exe"] {with-test}
+  ["dune" "build" "@doc" "-p" name] {with-doc}
+]
+run-test: ["dune" "runtest" "-p" name]
+synopsis: "SDL2_Image bindings to go with Tsdl"
+description: """
+Tsdl_image provides bindings to SDL2_Image intended to be used with
+Tsdl."""


tokenrove:
Note that the package name has changed with this release, due to the switch to dune (https://github.com/tokenrove/tsdl-image/pull/10).

packages/tsdl_image/tsdl_image.0.3.0/opam
```diff
@@ -0,0 +1,31 @@
...
+bug-reports: "http://github.com/tokenrove/tsdl-image/issues"
+tags: [ "bindings" "graphics" ]
+license: "BSD-3-Clause"
+depends: [
+  "ocaml" {>= "4.02"}
+  "ctypes" {>= "0.4.0"}
+  "ctypes-foreign"
+  "tsdl" {>= "0.9.0"}
+  "result"
+  "dune" {build & >= "1.11.0"}
```
kit-ty-kate:
```suggestion
  "dune" {>= "1.11.0"}
```
Dune shouldn't be tagged as a build dependency anymore (see #14266)

packages/tsdl_image/tsdl_image.0.3.0/opam
```diff
@@ -0,0 +1,31 @@
...
+]
+depexts: [
+  ["libsdl2-image-dev"] {os-family = "debian"}
+  ["sdl2_image"] {os-distribution = "homebrew" & os = "macos"}
+  ["sdl2_image"] {os-distribution = "arch"}
+]
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "build" "test/what.png" "test/test.exe"] {with-test}
+  ["dune" "build" "@doc" "-p" name] {with-doc}
```
kit-ty-kate:
```suggestion
  ["dune" "build" "@doc" "-p" name "-j" jobs] {with-doc}
```

packages/tsdl_image/tsdl_image.0.3.0/opam
```diff
@@ -0,0 +1,31 @@
...
+]
+run-test: ["dune" "runtest" "-p" name]
```
kit-ty-kate:
```suggestion
run-test: ["dune" "runtest" "-p" name "-j" jobs]
```

kit-ty-kate:
> Note that the package name has changed with this release, due to the switch to dune ([tokenrove/tsdl-image#10](https://github.com/tokenrove/tsdl-image/pull/10)).

If that's the case, could you add a transition package (e.g. `tsdl-image.transition`, see `jbuilder.transition` for reference) so that people can switch from one package to the other more easily?

kit-ty-kate:
It seems like you forgot to call `opam publish prepare` after having pushed the tag. There is no `url` section and thus opam doesn't know where to pull the archive from.

kit-ty-kate:
Since you are using dune I think it might be easier to use `dune-publish` instead of `opam-publish`

tokenrove:
Thanks, I'll deal with all of this in a couple of days.  One thing though:

> It seems like you forgot to call opam publish prepare after having pushed the tag. There is no url section and thus opam doesn't know where to pull the archive from.

The version of opam-publish I installed does not seem to have a `prepare` subcommand and I was very confused.  Sorry about this.

packages/tsdl_image/tsdl_image.0.3.0/opam
tokenrove: This seems like a significant issue and I might cancel this transition as a consequence.

tokenrove:
Closing while I decide if the aforementioned issue with dune is the straw that breaks the caml's back.  Thanks for the comments, and sorry for the noise.

kit-ty-kate:
> Closing while I decide if the aforementioned issue with dune is the straw that breaks the caml's back.

What issue?

packages/tsdl_image/tsdl_image.0.3.0/opam
kit-ty-kate: I'm curious to know why it is an issue for you if you have time to tell?

PR #14855, opened 2019-09-19
ocaml-manual: 4.08 and 4.09

Octachron:
Distributing the OCaml manual as an opam packages sounds like a nice idea. This PR adds the last two versions of the manual to the list of available manuals.

cc @dbuenzli


dbuenzli:
Thanks !

kit-ty-kate:
Thanks!

PR #14856, opened 2019-09-19
Solo5 release 0.6.1

mato:
This adds Solo5 0.6.1 to OPAM. This release contains API breaking
changes, so we also add constraints on `{< "0.6.0"}` on existing
`mirage-solo5` releases. `ocaml-freestanding` is unaffected by the
changes.

### DIFF
--- a/packages/mirage-solo5/mirage-solo5.0.4.0/opam
+++ b/packages/mirage-solo5/mirage-solo5.0.4.0/opam
@@ -26,7 +26,7 @@ depends: [
   "lwt" {>= "2.4.3"}
   "ocaml-freestanding" {>= "0.4.0"}
   "logs"
-  "solo5-bindings-hvt" | "solo5-bindings-virtio" | "solo5-bindings-muen"
+  "solo5-bindings-hvt" {< "0.6.0"} | "solo5-bindings-virtio" {< "0.6.0"} | "solo5-bindings-muen" {< "0.6.0"}
 ]
 conflicts: [
   "io-page" {< "2.0.0"}
--- a/packages/mirage-solo5/mirage-solo5.0.5.0/opam
+++ b/packages/mirage-solo5/mirage-solo5.0.5.0/opam
@@ -26,7 +26,7 @@ depends: [
   "lwt" {>= "2.4.3"}
   "ocaml-freestanding" {>= "0.4.1"}
   "logs"
-  ("solo5-bindings-hvt" | "solo5-bindings-virtio" | "solo5-bindings-muen" | "solo5-bindings-genode")
+  ("solo5-bindings-hvt" {< "0.6.0"} | "solo5-bindings-virtio" {< "0.6.0"} | "solo5-bindings-muen" {< "0.6.0"} | "solo5-bindings-genode" {< "0.6.0"})
 ]
 conflicts: [
   "io-page" {< "2.0.0"}
--- a/packages/solo5-bindings-genode/solo5-bindings-genode.0.6.1/opam
+++ b/packages/solo5-bindings-genode/solo5-bindings-genode.0.6.1/opam
@@ -1,0 +1,46 @@
+opam-version: "2.0"
+maintainer: "ehmry@posteo.net"
+authors: [
+  "Emery Hemingway <ehmry@posteo.net>"
+]
+homepage: "https://github.com/solo5/solo5"
+bug-reports: "https://github.com/solo5/solo5/issues"
+license: "ISC"
+dev-repo: "git+https://github.com/solo5/solo5.git"
+build: [
+  ["./configure.sh"]
+  [make "V=1" "CONFIG_HVT=" "CONFIG_SPT=" "CONFIG_VIRTIO=" "CONFIG_MUEN="]
+]
+install: [make "V=1" "CONFIG_HVT=" "CONFIG_SPT=" "CONFIG_VIRTIO=" "CONFIG_MUEN=" "install-opam-genode" "PREFIX=%{prefix}%"]
+remove: [
+  ["touch" "./Makeconf"]
+  [make "V=1" "uninstall-opam-genode" "PREFIX=%{prefix}%"]
+]
+depends: "conf-pkg-config"
+conflicts: [
+  "solo5-bindings-hvt"
+  "solo5-bindings-spt"
+  "solo5-bindings-virtio"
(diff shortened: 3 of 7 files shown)


mato:
(supersedes #14853 which was broken)

mato:
The CentOS builds are failing due to this:

```
- configure.sh: ERROR: GCC 4.9.0 or newer is required for -mstack-protector-guard= support
```

Nothing I can do about that.

These two failures from revdeps I don't understand:

- https://ci.ocaml.org/log/saved/docker-run-5aece532af104ea9c810cf25d8cdd6f0/9a4f75f47f25078f6df40f47b826327b1dcd247e
- https://ci.ocaml.org/log/saved/docker-run-86bfd1b2ec7ea9cd83dee180d41c6d84/ed7b5db0b26daaa884ca7f3513df1199b9283471

They're both complaining about:

```
# src/e_jnf.c: In function 'ynf':
# src/e_jnf.c:193:17: error: comparison of integer expressions of different signedness: 'int32_t' {aka 'int'} and 'unsigned int' [-Werror=sign-compare]
#   for(i=1;i<n&&ib!=0xff800000;i++){
#                  ^~
# cc1: all warnings being treated as errors
```

However that code (part of ocaml-freestanding / vendored openlibm) the same as that compiled for the other -bindings variants, and those builds succeed.

samoht:
What is very strange is that the revdeps are actually failing before even trying to build the packages being tested. For instance:

- this ocaml-freestanding.0.4.6 is broken https://ci.ocaml.org/log/saved/docker-run-7e6dae962b01e49ca2f668a77fa9513a/c7ad545b0db3b1d981cca7cc06446efcdff9f3f8
- while this one works fine: https://ci.ocaml.org/log/saved/docker-run-5aece532af104ea9c810cf25d8cdd6f0/9a4f75f47f25078f6df40f47b826327b1dcd247e 

Maybe it's due to versions mismatch regarding the C compiler. In any case I think it's fine to merge this PR as is.

mato:
Note that as of 0.6.0, Solo5 requires a C11-compliant compiler. So it's entirely possible that there would be other failures on distros that don't have one. I'd still _like_ to see the build results for "All Distros" though.

samoht:
Actually the two runs (one successful the other not) are both using debian buster, so I don't understand why they differ.

The "all deps" runs are failing on CentOS with:
```
# configure.sh: ERROR: GCC 4.9.0 or newer is required for -mstack-protector-guard= support
```

samoht:
I don't get why they don't use the same build options either:

```
- gcc  -ffreestanding -fstack-protector-strong -nostdinc -mstack-protector-guard=global -isystem /home/opam/.opam/4.08/lib/pkgconfig/../../include/solo5-bindings-spt/crt -I/home/opam/.opam/4.08/lib/pkgconfig/../../include/solo5-bindings-spt/solo5 -isystem /home/opam/.opam/4.08/.opam-switch/build/ocaml-freestanding.0.4.6/nolibc/include -fno-gnu89-inline -fno-builtin -O3 -fPIC -m64 -std=c99 -Wall -I/home/opam/.opam/4.08/.opam-switch/build/ocaml-freestanding.0.4.6/build/openlibm -I/home/opam/.opam/4.08/.opam-switch/build/ocaml-freestanding.0.4.6/build/openlibm/include -I/home/opam/.opam/4.08/.opam-switch/build/ocaml-freestanding.0.4.6/build/openlibm/amd64 -I/home/opam/.opam/4.08/.opam-switch/build/ocaml-freestanding.0.4.6/build/openlibm/src -DASSEMBLER -D__BSD_VISIBLE -Wno-implicit-function-declaration -I/home/opam/.opam/4.08/.opam-switch/build/ocaml-freestanding.0.4.6/build/openlibm/ld80 -c src/common.c -o src/common.c.o
```

vs.

```
- gcc  -std=c11 -Wall -Wextra -Werror -O2 -g -ffreestanding -fstack-protector-strong -nostdinc -mstack-protector-guard=global -fPIC -isystem /home/opam/.opam/4.08/lib/pkgconfig/../../include/solo5-bindings-genode/crt -I/home/opam/.opam/4.08/lib/pkgconfig/../../include/solo5-bindings-genode/solo5 -isystem /home/opam/.opam/4.08/.opam-switch/build/ocaml-freestanding.0.4.6/nolibc/include -fno-gnu89-inline -fno-builtin -fPIC -m64 -std=c99 -Wall -I/home/opam/.opam/4.08/.opam-switch/build/ocaml-freestanding.0.4.6/build/openlibm -I/home/opam/.opam/4.08/.opam-switch/build/ocaml-freestanding.0.4.6/build/openlibm/include -I/home/opam/.opam/4.08/.opam-switch/build/ocaml-freestanding.0.4.6/build/openlibm/amd64 -I/home/opam/.opam/4.08/.opam-switch/build/ocaml-freestanding.0.4.6/build/openlibm/src -DASSEMBLER -D__BSD_VISIBLE -Wno-implicit-function-declaration -I/home/opam/.opam/4.08/.opam-switch/build/ocaml-freestanding.0.4.6/build/openlibm/ld80 -c src/common.c -o src/common.c.o
```

mato:
@samoht 

> The "all deps" runs are failing on CentOS with:

Presumably the CentOS GCC is too old. As I wrote above, nothing I can/will do about it, Solo5 needs newer compiler features and GCC < 4.9.0 is ancient.

> I don't get why they don't use the same build options either:

That's weird. Solo5 0.6.0+ always uses `-std=c11`.

samoht:
> That's weird. Solo5 0.6.0+ always uses -std=c11.

apparently not :-)

mato:
> > That's weird. Solo5 0.6.0+ always uses -std=c11.
> 
> apparently not :-)

https://github.com/Solo5/solo5/blob/master/Makefile.common#L33
https://github.com/Solo5/solo5/blob/master/Makefile.common#L67

No build-time detection going on.

mato:
Ah, though the _downstream_ CFLAGS, i.e. those that get passed to OPAM packages via pkg-config, come from:

https://github.com/Solo5/solo5/blob/master/Makefile.common#L90

which doesn't set any `-std=` option, since we can't force that on downstream code.

hannesm:
i agree with @samoht https://github.com/ocaml/opam-repository/pull/14856#issuecomment-533186690 "fine to merge this PR as is" -- the above issue(s) can be fixed in separate PRs

PR #14857, opened 2019-09-19
fixes bug in conf-ida configuration

gitoleg:
An awful bug was discovered in `conf-ida` configuration, thanks @hluwa for that.
The way we wrote OS kind inferring was absolutely wrong so that PR fixes it.


kit-ty-kate:
thanks

PR #14858, opened 2019-09-19
Fix 4.09+musl switches

copy:
Removes `--disable-graph-lib` on 4.09 switches since graph-lib has been removed from 4.09.


kit-ty-kate:
Thanks!

PR #14860, opened 2019-09-20
batteries: updated upper bound ocaml version constraint

UnixJunkie:


packages/batteries/batteries.2.10.0/opam
```diff
@@ -18,7 +18,7 @@ build: [
 install: [make "install"]
 remove: ["ocamlfind" "remove" "batteries"]
 depends: [
-  "ocaml" {>= "4.00.0" & < "4.09.0"}
+  "ocaml" {>= "4.00.0" & <= "4.09.0"}
```
mseri: Wouldn't be better to allow for bugfixes and bound on the next version instead?
```suggestion
  "ocaml" {>= "4.00.0" & < "4.10.0"}
```

kit-ty-kate:
Thanks!

UnixJunkie:
@mseri thanks, you were right.

PR #14861, opened 2019-09-20
ppx_deriving_yojson 3.5.1 is compatible with OCaml 4.09.0

anmonteiro:


mseri:
The keyparsers 0.9.2 failure does not depend on this. I think it may need a bound on ppx_deriving_yojson, coulds you have a look at it and see if you have an idea for the rigth fix at a glance? https://ci.ocaml.org/log/saved/docker-run-03fa63394c05dc0b93e5d1e4a1c6fbe0/75a00ffd640ddb84a811d4bfb47e6c4b8e1821c5

PR #14864, opened 2019-09-20
stdcompat: bump base ocaml version constraint

thierry-martinez:
Last version of stdcompat works as is with ocaml 4.09, so this is just a change in the declared constraint.


thierry-martinez:
The revdep failure is on the `cviode` package: I am not sure to understand the failure (the error message seems to say that an X display is missing) but it does not seem to be related to `stdcompat`.

mseri:
The error is due to the way the new Qt library works when invoked by the plplot backend. It can be ignore, I'll probably disable the tests to avoid further issues

PR #14866, opened 2019-09-20
Jane Street packages: support for OCaml 4.09

xclerc:
According to the latest report at http://check.ocamllabs.io,
only one package is not compatible with OCaml 4.09,
namely `toplevel_expect_test.v0.12.1`.

This pull request constrains this version to prior OCaml
versions, and releases a new version compatible with 4.09.

(This new version of `toplevel_expect_test` needs a new
version of `core`, also in this pull request.)

### DIFF
--- a/packages/core/core.v0.12.4/opam
+++ b/packages/core/core.v0.12.4/opam
@@ -1,0 +1,31 @@
+opam-version: "2.0"
+maintainer: "opensource@janestreet.com"
+authors: ["Jane Street Group, LLC <opensource@janestreet.com>"]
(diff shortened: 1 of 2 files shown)


xclerc:
The Travis failure is actually a timeout.
The revdep failure of "anycache" appears unrelated.
The revdep failure of "capnp" is because version 3.2.1
is not compatible with the latest versions of core ([tentative fix](https://github.com/ocaml/opam-repository/pull/14869)).

kit-ty-kate:
Thanks a lot!

PR #14867, opened 2019-09-20
Adds CPDF v2.3

johnwhitington:
I have set the dependency on CamlPDF to be exactly the same version, since CamlPDF and CPDF versions always move in lockstep.


johnwhitington:
It looks like the CI tries to install camlpdf twice, which results in an ocamlfind error.

kit-ty-kate:
> It looks like the CI tries to install camlpdf twice, which results in an ocamlfind error.

Do you know the reason of this? Is there a problem detecting whether or not camlpdf is already installed or something?

johnwhitington:
It's hard for me to understand the log, but it seems to try to build camlpdf when it's already installed. In fact, it seems to think it's building cpdf when actually it's building camlpdf again. Here is the snippet:

```
3957∗ installed conf-m4.1
3958∗ installed ocamlfind.1.8.1
3959∗ installed camlpdf.2.3
3960Done.
3961
3962====== Installing package ======
3963The following actions will be performed:
3964  ∗ install cpdf 2.3
3965
3966<><> Gathering sources ><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
3967[cpdf.2.3] found in cache
3968
3969<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
3970+ /home/travis/.opam/opam-init/hooks/sandbox.sh "build" "make" (CWD=/home/travis/.opam/ocaml-system/.opam-switch/build/cpdf.2.3)
3971- make[1]: Entering directory '/home/travis/.opam/ocaml-system/.opam-switch/build/cpdf.2.3'
3972- ocamlfind ocamldep -native pdfmerge.mli > ._ncdi/pdfmerge.di
3973- ocamlfind ocamldep -native pdftype0.mli > ._ncdi/pdftype0.di
3974- ocamlfind ocamldep -native pdftruetype.mli > ._ncdi/pdftruetype.di
```

(`pdfmerge.ml`, `pdftype0.ml`, etc are CamlPDF source files not CPDF source files, so it's trying to build camlPDF here, but thinks it's trying to build CPDF?)

mseri:
At a glance, it seems to me that the source archive of cpdf has the same hash as the source archive of camlpdf. Can it be an incorrect source tarball?

johnwhitington:
Thanks. I've fixed it now.

johnwhitington:
By the way, could this be a security problem? If I can get a package with the same md5 as someone else's, I can get it in the repository, and have it installed in place of the one with which it clashes?

mseri:
Opam 2 supports sha256 and sha512 now, you can add them to reduce the attack potential

PR #14869, opened 2019-09-20
capnp.3.2.1: fix core dependency uses  (remove in v0.12).

xclerc:
`capnp.3.2.1` uses `Core.Std`, but the module was removed in v0.12;
this pull request hence simply tighten the dependency on `core`.


kit-ty-kate:
Thanks a lot!

PR #14875, opened 2019-09-21
Add Unison 2.48.15

madroach:
with lablgtk dependency optional

### DIFF
--- a/packages/unison/unison.2.48.15v4/opam
+++ b/packages/unison/unison.2.48.15v4/opam
@@ -1,0 +1,29 @@
+opam-version: "2.0"
+maintainer: "juergen@hoetzel.info"
+authors: [
+  "Jürgen Hötzel <juergen@hoetzel.info>"
+  "Louis Gesbert <louis.gesbert@ocamlpro.com>"
+]
+homepage: "https://www.cis.upenn.edu/~bcpierce/unison/"
+bug-reports: "mailto:unison-users@yahoogroups.com "
+dev-repo: "git://github.com/bcpierce00/unison.git"
+depends: [ "ocaml" {< "4.10"} ]
+depopts: [ "lablgtk" ]
+patches: ["ocaml48.patch"]
+install: [make "install" "OCAMLLIBDIR=%{lib}%" "HOME=%{prefix}%"]
+synopsis: "File-synchronization tool for Unix and Windows"
+description: """
+Unison is a file-synchronization tool for Unix and Windows. It allows
+two replicas of a collection of files and directories to be stored on
+different hosts (or different disks on the same host), modified
+separately, and then brought up to date by propagating the changes in
+each replica to the other."""
+extra-files: [
+  ["unison.install" "md5=728111a6bced5914b04a18fc919cffb0"]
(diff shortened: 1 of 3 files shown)


packages/unison/unison.2.48.15v4/opam
```diff
@@ -0,0 +1,29 @@
+opam-version: "2.0"
+maintainer: "juergen@hoetzel.info"
+authors: [
+  "Jürgen Hötzel <juergen@hoetzel.info>"
+  "Louis Gesbert <louis.gesbert@ocamlpro.com>"
+]
+homepage: "https://www.cis.upenn.edu/~bcpierce/unison/"
+bug-reports: "mailto:unison-users@yahoogroups.com "
+# dev-repo: "https://github.com/bcpierce00/unison"
+depends: [ "ocaml" {<= "4.09"} ]
```
kit-ty-kate:
```suggestion
depends: [ "ocaml" {< "4.10"} ]
```
Is this package really incompatible with possible fix versions of the compiler?

kit-ty-kate: Also if this is supposed to include `ocaml.4.09`, this is not the right constraint. The version `4.09` doesn't exist and is strictly lower than `4.09.0`

packages/unison/unison.2.48.15v4/opam
```diff
@@ -0,0 +1,29 @@
...
+depopts: [ "lablgtk" ]
+patches: ["ocaml48.patch"]
```
kit-ty-kate: I don't understand why adding a patch when it's already a new version? Why not having the patch directly in the release? It would be less complicated.

packages/unison/unison.2.48.15v4/opam
```diff
@@ -0,0 +1,29 @@
+opam-version: "2.0"
+maintainer: "juergen@hoetzel.info"
+authors: [
+  "Jürgen Hötzel <juergen@hoetzel.info>"
+  "Louis Gesbert <louis.gesbert@ocamlpro.com>"
+]
+homepage: "https://www.cis.upenn.edu/~bcpierce/unison/"
+bug-reports: "mailto:unison-users@yahoogroups.com "
+# dev-repo: "https://github.com/bcpierce00/unison"
```
kit-ty-kate:
```suggestion
dev-repo: "git://github.com/bcpierce00/unison.git"
```
Why commenting this field?

kit-ty-kate:
There seems to be a more recent release of unison: https://github.com/bcpierce00/unison/tree/v2.51.2

madroach:
> There seems to be a more recent release of unison: https://github.com/bcpierce00/unison/tree/v2.51.2

Unison has a special versioning scheme. 2.4x is the stable branch, while 2.5x is the beta branch.

packages/unison/unison.2.48.15v4/opam
madroach: This patch is already included upstream, but not yet released.

kit-ty-kate:
Ok, then. I'm not sure why MacOS fails to build but this package is good enough as is. Thanks!

PR #14876, opened 2019-09-21
cpdf.2.2.1: update opam upper bound

mseri:
See https://discuss.ocaml.org/t/how-to-use-opam-to-compile-a-package-that-doesnt-with-opam-but-does-with-make


mseri:
Closing. It fails due to safe-strings, the bound is needed

PR #14886, opened 2019-09-23
Solo5 0.6.2

mato:
This should have less CI failures than #14856, but will still fail on distros with a GCC older than 4.9.0.

Please merge despite CI failures, unless something unusual compared to #14856 pops up.


avsm:
The CentOS failure is expected. I'll do a followup PR to remove 0.6.1 as it's known-broken due to adding Werror.

PR #14887, opened 2019-09-23
LablTk 8.06.7 and ocamlbrowser 4.08.1

garrigue:
New release fixing compatibility problems for ocaml 4.09.
* still compatible with ocaml 4.07
* ocamlbrowser compatible with ocaml 4.08 and 4.09

### DIFF
--- a/packages/ocamlbrowser/ocamlbrowser.4.08.0/opam
+++ b/packages/ocamlbrowser/ocamlbrowser.4.08.0/opam
@@ -15,7 +15,7 @@ install: [
 remove: [["rm" "-f" "%{bin}%/ocamlbrowser"]]
 depends: [
   "ocaml" {>= "4.08"}
-  "labltk" {>= "8.06.6"}
+  "labltk" {= "8.06.6"}
   "ocamlfind" {build}
   "conf-tcl"
   "conf-tk"
--- a/packages/ocamlbrowser/ocamlbrowser.4.08.0/opam
+++ b/packages/ocamlbrowser/ocamlbrowser.4.08.0/opam
@@ -15,7 +15,7 @@ install: [
 remove: [["rm" "-f" "%{bin}%/ocamlbrowser"]]
 depends: [
   "ocaml" {>= "4.08"}
-  "labltk" {= "8.06.6"}
+  "labltk" {>= "8.06.6"}
   "ocamlfind" {build}
   "conf-tcl"
   "conf-tk"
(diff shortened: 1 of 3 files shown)


garrigue:
This new version of labltk is not compatible with ocamlbrowser.4.08.0, but it is superseded by ocamlbrowser.4.08.1. Not sure how this should be fixed, as ocamlbrowser.4.08.0 seems to have other problems, and better be left alone. Or would it be fine to just overwrite ocamlbrowser.4.08.0 with the opam file of ocamlbrowser.4.08.1 ?
For me it is fine to just release the current state, but if you prefer another approach tell me.

kit-ty-kate:
> This new version of labltk is not compatible with ocamlbrowser.4.08.0, but it is superseded by ocamlbrowser.4.08.1. Not sure how this should be fixed, as ocamlbrowser.4.08.0 seems to have other problems, and better be left alone. Or would it be fine to just overwrite ocamlbrowser.4.08.0 with the opam file of ocamlbrowser.4.08.1 ?
> For me it is fine to just release the current state, but if you prefer another approach tell me.

For that, the fix is simple: just modify `ocamlbrowser.4.08.0/opam` and add the appropriate constraint to `labltk` like you did for `ocamlbrowser.4.08.1` in this PR (`"labltk" {= "8.06.7"}`)

garrigue:
I tried, but this doesn't work. Either a problem with ocaml 4.09.0, or something else changed since it was published. Since it is completely superseded, this seems not with investigating.

packages/ocamlbrowser/ocamlbrowser.4.08.0/opam
```diff
@@ -15,7 +15,7 @@ install: [
 remove: [["rm" "-f" "%{bin}%/ocamlbrowser"]]
 depends: [
   "ocaml" {>= "4.08"}
```
mseri: Why not add `< “4.09}` here if the problem is with 4.09?

garrigue: It's just that I don't know for sure what the problem is, and it is not directly related with this commit.

garrigue:
OK, it seems that I solved the problem by basically preventing ocamlbrowser.4.08.0 from building (i.e. restricting it only to 4.08.0). I'm not even sure that labltk.8.06.7 builds on 4.08.0, but at least this makes the CI happy...

garrigue:
That was hard, but after a few more configure tweaks, everybody send happy now.

mseri:
Thanks! Seems fine to merge, @kit-ty-kate do you agree?

kit-ty-kate:
I do. Thanks a lot!

PR #14891, opened 2019-09-23
Shrink.filter is only provided since qcheck 0.8.

gildor478:
Fix failure in #14890


mseri:
Thanks!

mseri:
I have no idea why it fails on osx, and succeeds everywhere else. I think we can merge it

PR #14892, opened 2019-09-24
added packages/dolog.4.0.0

UnixJunkie:
Switched from oasis to dune.
This introduces a Dolog top namespace, hence the new major version.


UnixJunkie:
I will fix all rev deps also.

mseri:
Seems that the failure depend on external factors. Can you also have a look?

UnixJunkie:
It would be nice to have a tool to manage constraints in opam-repository.
Here, I wanted to add an upper bound constraint, to all users of this library which
did not have one until now.

I'll have a look at the failures tomorrow.

packages/dolog/dolog.4.0.0/opam
```diff
@@ -0,0 +1,22 @@
...
+name: "dolog"
+authors: "Francois Berenger"
+maintainer: "unixjunkie@sdf.org"
+license: "BSD-3-Clause"
+homepage: "https://github.com/UnixJunkie/dolog"
+bug-reports: "https://github.com/UnixJunkie/dolog/issues"
+dev-repo: "git+https://github.com/UnixJunkie/dolog.git"
+depends: [
+  "ocaml"
+  "dune"
```
kit-ty-kate:
```suggestion
  "dune" {>= "1.11"}
```
The `dune-project` file in your repository says this package requires at least dune 1.11

packages/dolog/dolog.4.0.0/opam
```diff
@@ -0,0 +1,22 @@
...
+  "base-unix"
+]
+build: ["jbuilder" "build" "-p" name "-j" jobs]
```
kit-ty-kate:
```suggestion
build: ["dune" "build" "-p" name "-j" jobs]
```
I think you're using dune not jbuilder

kit-ty-kate:
> It would be nice to have a tool to manage constraints in opam-repository.
> Here, I wanted to add an upper bound constraint, to all users of this library which
> did not have one until now.

For simple things you can use `opam admin add-constraint`. In your case you could have done `opam admin add-constraint "dolog<4.0.0"`. However I've only seen a few people use it and i haven't used it myself until a few minutes ago just to try. In particular it seems to reformat all packages and it might be very annoying for review (or commit if you decide to clean it before)

avsm:
I use `opam admin add-constraint` fairly often. The trick is to `git add` just the packages you want, and then to revert all the other constraints it adds (since you can't select individual packages with it until opam 2.1)

UnixJunkie:
opam-admin is not in opam?! :disappointed:

kit-ty-kate:
Thanks!

PR #14894, opened 2019-09-24
delete solo5 0.6.1 bindings

avsm:
as a followon to #14886 there is a serious build issue with the
0.6.1 bindings, so they are pulled in favour of 0.6.2

cc @mato


kit-ty-kate:
Sounds fair. Thanks

PR #14897, opened 2019-09-24
mirage-solo5 0.6.0

mato:
This is an API breaking release, this PR also adds additional upper bounds to device driver packages.

Part of mirage/mirage#992.


mato:
The CentOS failure is expected due to Solo5 0.6.0+ requiring GCC >= 4.9.0.

kit-ty-kate:
Thanks!

PR #14899, opened 2019-09-24
matplotlib.0.1: Fix several version constraints

kit-ty-kate:
Detected using [check.ocamllabs.io](http://check.ocamllabs.io)

cc @LaurentMazare could you return those fixes into your repository so that they won't appear in future releases?


LaurentMazare:
Thanks for pointing that out @kit-ty-kate , I've upstreamed these changes.

PR #14888, opened 2019-09-23
[new release] ounit-lwt and ounit (2.1.2)

gildor478:
OUnit testing framework (Lwt)

- Project page: <a href="https://github.com/gildor478/ounit">https://github.com/gildor478/ounit</a>
- Documentation: <a href="https://gildor478.github.io/ounit/">https://gildor478.github.io/ounit/</a>

### Fixed                                                                       
- Fix assert_raises type, which has been inadvertently changed during the       
  migration to dune.                                                            
- Add unix dependency to ounit.advanced.

PR #14900, opened 2019-09-24
Disable zmq.3.2-* as it is not installable. 

andersfugmann:
This PR disables zmq.3.2-* packages, as these are no longer installable:
* C types have changed
* Requires libzmq = 3.2, which is no longer available in popular distributions. 

See #14888.


mseri:
I think this is a lesser evil. Thanks for the PR.

PR #14901, opened 2019-09-24
Fix reverse dependencies of ounit.

gildor478:
This is a followup of #14888 to fix reverse dependencies of ounit.

- Remove the package
  - ~~dtoa.0.1 -> no remote tarball~~ -> available: false
  - ~~lambdasoup.0.6.2 -> bad remote checksum~~ -> updated checksum but it would be better if @aantron can upload a tarball to the release of 0.6.2, 0.6.1 and 0.6
- Fix the package (simple fix):
  - ~~gapi-ocaml.0.3.5 -> monad-custom missing~~ -> added pa_monad_custom
  - gapi-ocaml.0.3.6 -> missing netsys modules -> not sure what the error is 
  - ~~hamt.1.0.0 -> missing core_bench package~~ -> removed the tests because it fails to build anyway
  - ~~markup.0.7.6 -> missing lwt.unix library~~ -> added lwt
  - ~~session.0.1.0 -> missing postgresql package~~ -> added postgresql  and removed the tests because it was failing
- Stop running tests:
  - ~~diet.0.2 -> deprecation warnings turned into errors~~
  - ~~osc.0.1.4 -> missing exec sclang~~ -> remove test-interop-sclang
  - ~~pgx_lwt.0.1 -> failing tests (running with ounit, but failing)~~
  - ~~qcow.0.10.0 -> failing tests (running with ounit, but failing)~~
  - ~~sanlock.0.0.9 -> sanlock daemon isn't running~~
  - ~~key-parsers.0.9.2 -> type error~~
  - ~~mecab.0.0.0 -> failing tests (running with ounit, but failing)~~
  - ~~yices2.0.0.2 -> fail to run /usr/bin/make ""~~
  - ~~yices2.0.0.3 -> fail to run /usr/bin/make ""~~
  - ~~zmq.5.0.0 -> failing tests (running with ounit, but failing)~~ maybe @andersfugmann want to have a look
- Limit OCaml version:
  - ~~mustache.3.0.0 -> type error (string vs byte)~~ -> added base-unsafe-string and removed the tests because it was failing
  - ~~mustache.3.0.2 -> type error (string vs byte)~~ -> added base-unsafe-string and removed the tests because it was failing
  - ~~pcap-format.0.5.0 -> missing Array1.map_file~~ -> ocaml < 4.08.0
  - ~~zmq.3.2-0 -> C stub fails to compile~~ -> #14900
  - ~~zmq.3.2-1 -> C stub fails to compile~~ -> #14900
  - ~~zmq.3.2-2 -> C stub fails to compile~~ -> #14900

Don't know section:
llvm.* -> version mismatch with installed C library
~~scid.1.0 -> missing file lib_test/suite.mly~~ -> removed test, @vbmithr please reactivate if needed
~~semver.0.1.0 -> bad oasis test (No variable test_semverh defined)~~ -> removed test

### DIFF
--- a/packages/zmq/zmq.5.0.0/opam
+++ b/packages/zmq/zmq.5.0.0/opam
@@ -9,12 +9,14 @@ bug-reports: "https://github.com/issuu/ocaml-zmq/issues"
 build: [
   ["jbuilder" "subst" "-p" name] {pinned}
   ["jbuilder" "build" "-p" name "-j" jobs]
+  ["jbuilder" "runtest" "-p" name "-j" jobs] {with-test}
 ]
 depends: [
   "ocaml"
   "conf-zmq"
   "jbuilder" {build & >= "1.0+beta17"}
   "configurator" {build & < "v0.13"}
+  "ounit" {with-test}
   "base-unix"
   "stdint" {>= "0.4.2"}
 ]
--- a/packages/zmq/zmq.5.0.0/opam
+++ b/packages/zmq/zmq.5.0.0/opam
@@ -9,14 +9,12 @@ bug-reports: "https://github.com/issuu/ocaml-zmq/issues"
 build: [
   ["jbuilder" "subst" "-p" name] {pinned}
   ["jbuilder" "build" "-p" name "-j" jobs]
-  ["jbuilder" "runtest" "-p" name "-j" jobs] {with-test}
 ]
 depends: [
   "ocaml"
   "conf-zmq"
   "jbuilder" {build & >= "1.0+beta17"}
   "configurator" {build & < "v0.13"}
-  "ounit" {with-test}
   "base-unix"
   "stdint" {>= "0.4.2"}
 ]
--- a/packages/diet/diet.0.2/opam
+++ b/packages/diet/diet.0.2/opam
@@ -10,7 +10,6 @@ doc: "https://djs55.github.io/ocaml-diet"
 build: [
   ["dune" "subst" ] {pinned}
   ["dune" "build" "-p" name "-j" jobs]
-  [make "test"] {with-test}
 ]
 depends: [
   "ocaml" {>= "4.03.0"}
@@ -18,7 +17,6 @@ depends: [
   "sexplib"
   "dune"
   "ppx_sexp_conv" {>= "v0.9.0"}
-  "ounit" {with-test}
 ]
 synopsis: "Discrete Interval Encoding Trees"
 description: """
--- a/packages/dtoa/dtoa.0.1/opam
+++ b/packages/dtoa/dtoa.0.1/opam
@@ -25,3 +25,4 @@ url {
   src: "https://github.com/flowtype/ocaml-dtoa/archive/v0.1.tar.gz"
   checksum: "md5=40cb786876812a7e70ab5a7cf4094829"
 }
+available: false
--- a/packages/gapi-ocaml/gapi-ocaml.0.3.5/opam
+++ b/packages/gapi-ocaml/gapi-ocaml.0.3.5/opam
@@ -26,7 +26,7 @@ depends: [
   "ocamlnet" {>= "4.1.4"}
   "ocurl"
   "ounit" {with-test}
-  "ppx_monadic" {with-test}
+  "pa_monad_custom" {with-test}
   "xmlm"
   "yojson"
 ]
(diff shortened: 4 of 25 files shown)


packages/zmq/zmq.5.0.0/opam
```diff
@@ -9,14 +9,12 @@ bug-reports: "https://github.com/issuu/ocaml-zmq/issues"
 build: [
   ["jbuilder" "subst" "-p" name] {pinned}
   ["jbuilder" "build" "-p" name "-j" jobs]
-  ["jbuilder" "runtest" "-p" name "-j" jobs] {with-test}
```
andersfugmann: The failing tests has nothing to do with oUnit, but rather that libzmq has changed semantics (There is a PR against the zmq repo to fix the issue, and a new release will be made soon). 

Is the general policy that we disable tests when/if they fail? It seems a bit contradictory to me.

gildor478: I am not a packager on OPAM, so this is just IMHO (and I do prefer to have tests in general -- I am the upstream author of ounit). 

My take here is that a failing test on OPAM is non-actionnable. The best is to open an issue against ocaml-zmq repository, disable the test on OPAM and to re-enable the test on OPAM when the upstream release  will be made. An issue in ocaml-zmq repository is more actionnable for the ocaml-zmq maintainers than for the people uploading packages on OPAM and getting "red flags" because a random reverse-dependency is failing.

Does it make sense to you?

gildor478:
@andersfugmann can you check packages/zmq-async/zmq-async.5.0.0/opam and packages/zmq-lwt/zmq-lwt.5.0.0/opam as well?

They were missing a dependency on ounit.

gildor478:
I had to disable zmq-async and zmq-lwt as well (kind of expected, I suppose given they are coming from the same source).

If I understand things correctly, even if you do a new release of ocaml-zmq, there are no ways that zmq.5.0.0 in the OPAM repository will work correctly?

The next version will work but the 5.0.0 version will always fail, AFAIU. Maybe once you have the new version, it is worth making this one uninstallable?

packages/zmq/zmq.5.0.0/opam
andersfugmann: I do understand that failing revdeps is annoying, but I think disabling tests here is a bit drastic.
It seems that there should be a general policy on what to do with failing revdeps, which seems slightly out of scope in this PR. 

In this particular case, its fine by me to disable tests on zmq 5.0, as a  new version will be arriving soon.

mseri:
If the issue is with libzmq, and it is going to be fixed, I would rather keep those tests failing and wait for the update libzmq to appear bundled in the failing distributions.

Tendentially, we disable the tests when there is the impossibility to run them (e.g. lack of library in some/many distributions, sandboxing issues, misconfiguration of the bundled tests, ...) and the package maintainers agree, but there is not a strict ruleon this.

andersfugmann:
Thanks for the pointers. I will make sure to update the opam dependencies of zmq-*.
As for disabling zmq-5.0, I'm somewhat hesitant as it builds and runs correctly (including tests) with libzmq < 4.3. I'll give it some thought.

gildor478:
@andersfugmann I understand your concerns and I have reverted the changes. We can disable the tests later if needed.

Just my 2 cents, as a non-expert, I would expect that ocaml-zmq v5.0.0 works with libzmq v5. This is just a question of versions matching. On the other hand the description is "OCaml bindings for ZeroMQ 4.x", so I am a bit confused.

aantron:
@gildor478 When following semantic versioning, it's usually impractical to keep version numbers synchronized with some underlying project. My observation has been that projects that try this eventually abandon it.

mseri:
Thanks a lot to you all for this cleanup!

Leonidas-from-XIV:
> Just my 2 cents, as a non-expert, I would expect that ocaml-zmq v5.0.0 works with libzmq v5. This is just a question of versions matching. On the other hand the description is "OCaml bindings for ZeroMQ 4.x", so I am a bit confused.

This was also the case for the ZMQ bindings, but after 4.0-1 to 4.0-7 which tell you next to nothing about the relative size of the change we decided to make 5.0 as a big major release since the async and lwt binding authors joined the project and contributed their bindings directly.

It is also somewhat tricky to match the version of libzmq since we support 4.0 to the current release in one release of ocaml-zmq so the only way I could see this potentially work is to have `4.0.major.minor.patch` which is somewhat surprising to all users of the OCaml library and even more confusing in case a potentially mostly compatible ZMQ 5 came along.

PR #13747, opened 2019-03-25
conf-libX11: remove osx script - execvp not allowed in osx sanbox

mseri:
This should solve https://github.com/ocaml/opam-repository/issues/13686

Signed-off-by: Marcello Seri <marcello.seri@gmail.com>

PR #14910, opened 2019-09-26
Fix conf-libX11 always recompiling

dra27:
Reported in https://github.com/ocaml/opam/issues/3995, this issue started with #13747. There's a bug in opam which is causing an extra newline to be added to the opam file which is installed to
`.opam-switch/packages/conf-libX11/opam` which is causing opam to think
the file has changed on every update.

The workaround is to use the triple quotes without a line-break, which I suggest is the best course of action here.


mseri:
Thanks!

PR #14912, opened 2019-09-26
Fix other packages affected by triple-quoting opam bug

dra27:
Following on from #14910, I checked for other packages.

The fix is to ensure the strings contain no newlines, which creates some hideously long lines (in some cases, these might better be moved to scripts?)

For now, `"""` should only be used in meta-data (so `messages`, `post-messages`, `synopsis` and `description` are OK, but commands are not)

### DIFF
--- a/packages/clangml/clangml.3.9.1.2/opam
+++ b/packages/clangml/clangml.3.9.1.2/opam
@@ -9,12 +9,7 @@ build: [
   [
     "sh"
     "-c"
-    """
-test -d ${HOME}/usr/clang39 \\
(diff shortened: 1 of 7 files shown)


mseri:
It's a pity, it makes the command practically unreadable. I'll wait for the CI, then merge

kit-ty-kate:
I think moving them to separate scripts is a better idea

mseri:
`torch` failures are unrelated. Tbh I'd merge them as they are for now, the only real bad one is clang, we could move it to a separate script later if it's really needed

kit-ty-kate:
ok fair enough

kit-ty-kate:
Thanks

PR #14913, opened 2019-09-26
conf-libev: try to build without any flags first

aantron:
Before adding -I and -L flags, try to build using just $CC test.c -lev.
This will succeed if libev is installed already in the compiler's
default search path. This improves libev detection with musl-gcc, as the
musl libev is typically installed in musl-gcc's search path. With
musl-gcc, immediately searching for ev.h and adding -I and -L is likely
to find a glibc (system) libev, rather than the one compiled for use
with musl-gcc.

See

  https://github.com/ocsigen/lwt/issues/718#issuecomment-535053644

### DIFF
--- a/packages/conf-libev/conf-libev.4-11/files/discover.ml
+++ b/packages/conf-libev/conf-libev.4-11/files/discover.ml
@@ -118,7 +118,7 @@ let c_args =
     | None -> ""
     | Some path -> flags path
 
-let compile args stub_file =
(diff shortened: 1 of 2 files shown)


mseri:
@aantron you need to update the hash of the modified file in the opam file

aantron:
Argh yes, I did see it, sorry. Updated now...

kit-ty-kate:
Thanks

PR #14915, opened 2019-09-26
[new release] decompress and rfc1951 (1.0.0)

dinosaure:
Implementation of Zlib in OCaml

##### CHANGES:

** breaking changes **

`decompress.1.0.0` is 3 times faster about decompression than before. A huge
[amount of work was done](https://tarides.com/blog/2019-08-26-decompress-the-new-decompress-api.html) to improve performance and coverage.

The main reason to update the API is to fix a bad design decision regarding split
compression and encoding. User is able to implement a new compression algorithm
and use it.

Release comes with regressions:
- `decompress` only supports `Bigarray` now, not `Bytes`
- GZIP layer does not exist anymore
- state of RFC1951 encoder/decoder is not referentially transparent anymore

Of course, v1.0.0 comes with fixes and improvements:
- `decompress` is able to compress/uncompress [Calgary corpus](https://en.wikipedia.org/wiki/Calgary_corpus)
- tests are improved and they include all coverage tests from `zlib`
- compression algorithm has a fuzzer
- encoder has a fuzzer
- performance about decoder is 3 times better than `decompress.v0.9.0` and 3
  times slower than `zlib`

`decompress` is split into 2 main modules:
- `dd` which implements RFC1951
- `zz` which implements ZLIB

API of them are pretty-close to what `decompress.v0.9.0` does with some
advantages on `dd`:
- User can use their own compression algorithm instead of `Dd.L`
- encoder exposes more granular control over what it emits (which block, when, where)
- Huffman tree generation is out of `dd`

As a response to mirage/decompress#25, `dd` provides a _higher_ level API resembling `camlzip`.

### DIFF
--- a/packages/decompress/decompress.1.0.0/opam
+++ b/packages/decompress/decompress.1.0.0/opam
@@ -1,0 +1,37 @@
+opam-version: "2.0"
+name:         "decompress"
+maintainer:   "Romain Calascibetta <romain.calascibetta@gmail.com>"
+authors:      "Romain Calascibetta <romain.calascibetta@gmail.com>"
+homepage:     "https://github.com/mirage/decompress"
+bug-reports:  "https://github.com/mirage/decompress/issues"
+dev-repo:     "git+https://github.com/mirage/decompress.git"
+doc:          "https://mirage.github.io/decompress/"
+license:      "MIT"
+synopsis:     "Implementation of Zlib in OCaml"
+description: """Decompress is an implementation of Zlib in OCaml
+
+It provides a pure non-blocking interface to inflate and deflate data flow.
+"""
+
+build: [ "dune" "build" "-p" name "-j" jobs ]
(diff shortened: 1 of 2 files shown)


dinosaure:
`crowbar` seems to lock any installation of `decompress` with `opam install -t`, I don't know how to fix it properly.

kit-ty-kate:
> `crowbar` seems to lock any installation of `decompress` with `opam install -t`, I don't know how to fix it properly.

that is due to https://github.com/ocaml/opam-depext/issues/121 and will be fixed in opam 2.1. However some thing can be fixed in mirage-ci so that we still test builds before tests. I'm on it.

kit-ty-kate:
`decompress` is now able to be tested for installation. It still won't be able to fix the issue with testing but it's good enough. Thanks!

dinosaure:
Ok, thanks :+1:

PR #14916, opened 2019-09-26
ZMQ 5.1.0 (including Lwt and Async)

Leonidas-from-XIV:
Since we got pinged in #14888 we implemented some fixes for the failing tests as well as merging some minor low hanging fruits.

### DIFF
--- a/packages/zmq-async/zmq-async.5.1.0/opam
+++ b/packages/zmq-async/zmq-async.5.1.0/opam
@@ -1,0 +1,31 @@
+opam-version: "2.0"
+maintainer: "anders@fugmann.net"
+authors: ["Rudi Grinberg"]
+license: "MIT"
+homepage: "https://github.com/issuu/ocaml-zmq"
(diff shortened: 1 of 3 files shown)


Leonidas-from-XIV:
* It looks like ZeroMQ is only on CentOS 7 when you enable EPEL, otherwise the package fails to install. I don't know what to do about this, the CI should possibly enable EPEL. @kit-ty-kate, what do you recommend doing in such cases?
* Core v0.9 on OCaml 4.03 uses `makedev` that doesn't compile anymore because glibc decided to break users. The [fix](https://github.com/janestreet/core/commit/deed3f6bc14e5f160a52b9e58144d3c4e541184e) is only available starting in v0.11 which requires OCaml 4.04.1.
* OpenSUSE fails to build because `conf-zmq` doesn't have a depext for it. This should be easy to fix.

kit-ty-kate:
Thanks

PR #14917, opened 2019-09-26
anycache-lwt 0.7.4 requires anycache >= 0.7.0

aantron:
`anycache-lwt` 0.7.4 [expects](https://ci.ocaml.org/log/saved/docker-run-67163e3725c0cf22cf043e5f3088816b/5bd5fa1b8c13bf2bb9e5e2f1ac54ca1e81cdaf87) a module `Anycache.PendingLimit`, which first [appeared](https://gitlab.com/edwintorok/ocaml-anycache/blob/v0.7.0/lib/anycache.mli#L125) in `anycache` 0.7.0.

The source repo is archived, so we probably don't need to cc anyone.


kit-ty-kate:
Thanks!

PR #14919, opened 2019-09-26
Add depext package name for ZMQ on openSUSE

Leonidas-from-XIV:
openSUSE can't install `zmq` because the `conf-zmq` does not specify a package name to install. This PR specifies the name to a name that the internet claims might be correct and matches the RPM tradition at least.

This should solve the openSUSE build issue in #14916.


kit-ty-kate:
Thanks

PR #14920, opened 2019-09-26
libzmq seems to only be in EPEL on CentOS

Leonidas-from-XIV:
Adding `epel-release` is what `conf-openblas` is doing so presumably this is the correct thing to do.

This should solve the openSUSE build issue in #14916.


kit-ty-kate:
Thanks

PR #14921, opened 2019-09-26
Dune >= 1.10.0 fails on caqti-lwt 0.11.0

aantron:
See [this log](https://ci.ocaml.org/log/saved/docker-run-ddf86665accb49c4cdcb25d64aa5816c/c276c00a549bf533cad9606cf32a832ead77631c). I reported the issue to Dune: https://github.com/ocaml/dune/issues/2674.

This doesn't affect the latest `caqti-lwt`, so I don't think we need to notify anyone else.


kit-ty-kate:
Thanks a lot!

PR #14922, opened 2019-09-26
exenum 0.86 is missing a dependency on lwt.unix

aantron:
It was possible to link with `lwt.unix` while only specifying `lwt` until Lwt 4.0.0. The project repo has a fix, lebotlan/ocaml-exenum@eab63e63a26d0a014c254183e66f236abe1529ce, but it is not released to opam.


yomimono:
Thanks for the fix :)

PR #14923, opened 2019-09-26
inotify 2.1 and 2.2 with-test requires OUnit

aantron:
See the builds logs for [2.1](https://ci.ocaml.org/log/saved/docker-run-f0214beb3618a0cc43be979fe654a978/b812ca31c5c04e43356df591fc0c28008eed0996) and [2.2](https://ci.ocaml.org/log/saved/docker-run-993b87ea8f3f46992097fd5c77ed7490/3efb551b4cc10149c75651e72938f01b7e786147). inotify 2.3 doesn't have `with-test` build commands, so I don't think we need to notify anyone.


kit-ty-kate:
Thanks!

PR #14925, opened 2019-09-26
release camels version 0.0.3

yomimono:


yomimono:
in a clear case of discrimination against bad software, my build has been cancelled :(

packages/camels/camels.0.0.3/opam
```diff
@@ -0,0 +1,39 @@
...
+license: "GPLv3"
+tags: ["game"]
+
+build: [
+  ["dune" "build"]
+]
+
+depends: [
+  "cmdliner"
+  "dune" {build}
```
kit-ty-kate:
```suggestion
  "dune" {>= "1.8"}
```
Dune shouldn't be tagged as a build dependency (see #14266) and the `dune-project` file in the repository says this package requires at least dune 1.8

packages/camels/camels.0.0.3/opam
```diff
@@ -0,0 +1,39 @@
...
+maintainer: ["maintenance@identity-function.com"]
+authors: ["yomimono"]
+homepage: "https://github.com/yomimono/camels"
+bug-reports: "https://github.com/yomimono/camels/issues"
+dev-repo: "git+https://github.com/yomimono/camels.git"
+license: "GPLv3"
+tags: ["game"]
+
+build: [
+  ["dune" "build"]
```
kit-ty-kate:
```suggestion
  ["dune" "build" "-p" name "-j" jobs]
```
`dune build` alone without `-p` or `--profile=release` is going to break at some point because it is meant for development and turns some warnings as errors.

yomimono:
Could we hint both of those things to users with Camelus or something similar? You've definitely had to tell me this more than once and a computer could be scolding me instead, way quicker, without me feeling like I'm wasting your time.

kit-ty-kate:
> Could we hint both of those things to users with Camelus or something similar? You've definitely had to tell me this more than once and a computer could be scolding me instead, way quicker, without me feeling like I'm wasting your time.
> 
> [ed: that read way harsher than I meant it to on reread; I just think it would be nice for me to not screw up package submission so much :) ]

That is on my todo list indeed, it would be a matter of adding more and more heuristics. But right now the state of Camelus is not the greatest. I'm trying to resolve that

kit-ty-kate:
Thanks!

PR #14926, opened 2019-09-26
mirage-block-ramdisk 0.3 expects real package io-page.unix

aantron:
See [log](https://ci.ocaml.org/log/saved/docker-run-04ce8ffd4887f755f65018bf021b4ba7/0591e1832eb6c89a9b1d42b11fbe905cca54ed9f). `io-page.unix` was a true package in the io-page repo until io-page 2.0.0, when it became an alias for io-page-unix, which has to be installed separately. mirage-block-ramdisk 0.3 expects `io-page.unix` not to be an alias.

This is not the latest release of mirage-block-ramdisk, so I don't think we have to ping anyone.


kit-ty-kate:
Thanks

PR #14929, opened 2019-09-26
pgx_lwt: don't run tests, they call getlogin

aantron:
Package `pgx_lwt`'s tests (transitively) run `getlogin`, which seems to fail in opam's containers. See [this log](https://ci.ocaml.org/log/saved/docker-run-cc33e62085841c5125f6a50112fa3c7c/8511d8abd1a6b269d3f6827e5b79ea1b8943f1d9). It seems best to simply not run them.

cc @brendanlong about removing that same line in the `pgx` repo, provided opam maintainers consider this change to be reasonable.

### DIFF
--- a/packages/pgx_lwt/pgx_lwt.0.1/opam
+++ b/packages/pgx_lwt/pgx_lwt.0.1/opam
@@ -8,7 +8,6 @@ doc: "https://arenadotio.github.io/pgx/doc"
 
 build: [
   ["jbuilder" "build" "-p" name "-j" jobs]
-  ["jbuilder" "runtest" "-p" name "-j" jobs] {with-test}
 ]
 depends: [
   "ocaml" {>= "4.04.2"}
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
The change seems reasonable indeed

kit-ty-kate:
I think it would be good to also remove the test dependencies

mseri:
Thanks

brendanlong:
@aantron I'd be fine with merging a fix for this upstream in pgx too. I'm not sure what exactly we should do though. Do we just need to wrap the call to `getlogin` and catch whatever exception gets thrown?

(It seems I looked at this too late and the build log is gone)

aantron:
I don't recall at this point, but likely yes. Either the test should be skipped if `getlogin` fails, or the `opam` file shouldn't try running tests at all. Probably the first is better.

PR #14930, opened 2019-09-26
Restrict zmq-async minimal OCaml/JST version

Leonidas-from-XIV:
Due to an issue with where glibc puts `makedev` and where Core < v0.11 (a dependency of Async) expects `makedev` this package will always fail to build. It is fixed in Core v0.11 but that requires at least OCaml 4.04.1 so we're restricting the minimal version.

A similar change is planned upstream, the next release of ZMQ will probably drop 4.03 support altogether (also in the base library and Lwt bindings): https://github.com/issuu/ocaml-zmq/pull/93

This PR is a more conservative approach to limit the disruption to users of the release on OPAM.

### DIFF
--- a/packages/zmq-async/zmq-async.5.1.0/opam
+++ b/packages/zmq-async/zmq-async.5.1.0/opam
@@ -12,12 +12,12 @@ build: [
   ["dune" "runtest" "-p" name "-j" jobs] {with-test}
 ]
 depends: [
-  "ocaml"
+  "ocaml" {>= "4.04.1"}
   "zmq" {>= "5.1.0"}
   "dune"
-  "async_unix" {>= "v0.9.0" & < "v0.13"}
-  "async_kernel" {>= "v0.9.0" & < "v0.13"}
-  "base" {>= "v0.9.0" & < "v0.13"}
+  "async_unix" {>= "v0.11.0" & < "v0.13"}
+  "async_kernel" {>= "v0.11.0" & < "v0.13"}
+  "base" {>= "v0.11.0" & < "v0.13"}
   "ounit" {with-test}
 ]
 synopsis: "Async aware bindings to zmq"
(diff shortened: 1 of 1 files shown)


mseri:
There is a ounit related failure, I think it may require an upper bound on ounit < 2.2.0:

```
#=== ERROR while compiling zmq-async.5.1.0 ====================================#
1574# context     2.0.5 | macos/x86_64 | ocaml-system.4.08.1 | file:///Users/travis/build/ocaml/opam-repository
1575# path        ~/.opam/ocaml-system/.opam-switch/build/zmq-async.5.1.0
1576# command     ~/.opam/opam-init/hooks/sandbox.sh build dune runtest -p zmq-async -j 1
1577# exit-code   1
1578# env-file    ~/.opam/log/zmq-async-93592-b7bf70.env
1579# output-file ~/.opam/log/zmq-async-93592-b7bf70.out
1580### output ###
1581# File "zmq-deferred/test/dune", line 4, characters 25-30:
1582# 4 |  (libraries zmq.deferred oUnit))
1583#                              ^^^^^
1584# Error: Library "oUnit" not found.
1585# Hint: try: dune external-lib-deps --missing -p zmq-async -j 1 @runtest
1586# File "zmq/test/dune", line 3, characters 12-17:
1587# 3 |  (libraries oUnit zmq threads))
1588#                 ^^^^^
1589# Error: Library "oUnit" not found.
1590# Hint: try: dune external-lib-deps --missing -p zmq-async -j 1 @runtest
```

I would prefer to not touch the current package but introduce `zmq-async.5.1.0-1` with these changes instead

mseri:
I just realised that this was just released. I am going to disable the tests for now, please have a look at the ounit error for future releases

Leonidas-from-XIV:
@mseri I am somewhat out of the loop with things. If I see correctly in ounit 2.1.1 a new version was released that renames `oUnit` to `ounit` while keeping both names around. That release does not seem to be on opam. Then there is 2.1.2, which doesn't seem to build on macOS. Then there is 2.2.0 which seems to introduce `ounit2` with presumably `ounit` as package name but that also doesn't seem to be on OPAM yet.

So, as a package maintainer I don't really know what the right thing to depend on is to get my tests working.

mseri:
I think either add an upper bound like I did in this PR or move to ounit2

PR #14931, opened 2019-09-26
protocol-9p 0.10.0 assumes io-page < 2.0.0

aantron:
Packages [protocol-9p 0.10.0](https://ci.ocaml.org/log/saved/docker-run-5430197842937898f6705692c5a97887/435d84467bd808f6229a90954758d52fcc3539ab) and [protocol-9p-unix 0.10.0](https://ci.ocaml.org/log/saved/docker-run-bda75b25b0c61c74d7fb179333694f1f/f62b991a5f20a2be2db1f4f8e3e30e5950e021b2) assume that io-page.unix is a real package, rather than an alias for io-page-unix, which they don't attempt to install. io-page.unix became an alias in io-page 2.0.0.

Newer versions of these packages build fine, so I don't think we need to notify anyone.

There are additional problems with some of the older protocol-9p packages. I will push in more commits after each build. I will mark this PR as ready for review once I think the problems are all addressed.

### DIFF
--- a/packages/protocol-9p-unix/protocol-9p-unix.0.10.0/opam
+++ b/packages/protocol-9p-unix/protocol-9p-unix.0.10.0/opam
@@ -37,6 +37,9 @@ depends: [
   "ppx_tools"
   "alcotest" {with-test & >= "0.4.0"}
 ]
+conflicts: [
+  "io-page" {>= "2.0.0"}
+]
 synopsis: "Unix clients and servers for the 9P protocol"
 url {
   src: "https://github.com/mirage/ocaml-9p/archive/v0.10.0.tar.gz"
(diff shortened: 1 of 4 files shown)


aantron:
The tests of [protocol-9p 0.11.0](https://ci.ocaml.org/log/saved/docker-run-b557c72f7de1f481f66ff80f9b128029/6bae46eee23fbd382f177bd9a15f3f6898323145) and [protocol-9p 0.11.1](https://ci.ocaml.org/log/saved/docker-run-2266df52aeef114ae7903c412e1979e8/7355a0909448f544bac71f12f24e0d8c6c6d3d7d) assume that protocol-9p-unix is also installed, but it isn't, when these packages are being installed, and can't be, because it depends on protocol-9p. The commit above disables the tests.

mseri:
Thanks

PR #14621, opened 2019-08-01
io-page: add bounds to io-page-unix and io-page-xen to prepare the re…

samoht:
…lease

PR #14932, opened 2019-09-26
Disable tests of qcow 0.10.0

aantron:
These tests are [failing](https://ci.ocaml.org/log/saved/docker-run-ce590c58a5d05af2dcabe12abcd10711/ea0a30c6c8ef81cb212828c8ad264c8cefc6d7e1). There are newer versions of qcow, so I don't think there is anything to report or do about these tests; they can only be disabled.

Interestingly, the newer versions, not affected by this PR, had their tests disabled in one of the commits in https://github.com/ocaml/opam-repository/pull/14621.


kit-ty-kate:
Thanks

PR #14933, opened 2019-09-26
Disable tests of session 0.1.0

aantron:
They [fail](https://ci.ocaml.org/log/saved/docker-run-663068e986527ca1811abad70d6b2399/0f16353d96c8538dc439a4ccb52575e7b447eb06) while trying to find package `postgresql`, which is not installed. This was fixed in the upstream repo the day after the 0.1.0 release in https://github.com/inhabitedtype/ocaml-session/commit/254b88e62f406632609ccbf82ee94309700c5810, which is included in every subsequent release, so we don't need to notify anyone.


mseri:
Thanks

PR #14934, opened 2019-09-26
zmq-lwt 5.0.0 tests require ounit

aantron:
Tests for package zmq-lwt 5.0.0 [require](https://ci.ocaml.org/log/saved/docker-run-65d4af0716a46a5d48a6bfe2cdfc3dc2/69d151c5465c4df4d61797e7751913cc186765d2) OUnit. I based the fix on the dependencies in zmq-lwt 5.1.0. Since there is a newer package, I don't think anyone needs to be notified.

### DIFF
--- a/packages/zmq-lwt/zmq-lwt.5.0.0/opam
+++ b/packages/zmq-lwt/zmq-lwt.5.0.0/opam
@@ -14,6 +14,7 @@ build: [
 depends: [
   "ocaml"
   "zmq" {>= "5.0.0"}
+  "ounit" {with-test}
   "jbuilder" {build & >= "1.0+beta17"}
   "lwt"
 ]


Leonidas-from-XIV:
The build failures are because newer versions of libzmq (>= 4.3.x) added events that the testsuite didn't anticipate, which is why we released 5.1.0 today that is compatible with both older and newer versions of libzmq.

aantron:
I can disable the tests of 5.0.0 now if required. Otherwise, myself, I would let this sit until the next Lwt release, and make any additional changes if there are any more problems in the revdeps build.

Leonidas-from-XIV:
I don't have a particular preference either way. For the next release we'll probably wait until the `ounit2`/`oUnit`/`ounit` situation has stabilized and pick the right dependency.

aantron:
The one commit for now is enough with respect to my interests in Lwt revdeps builds, so I suggest we merge if this is acceptable.

PR #14941, opened 2019-09-27
Release override 0.2.2

thierry-martinez:
- support for OCaml 4.09

- compatible with the last versions of `ppxlib` and `ppx_deriving`


kit-ty-kate:
Thanks!

PR #14942, opened 2019-09-27
Release pattern.0.1.1

thierry-martinez:
- compatible with OCaml 4.09.

- use ppxlib parsetree to be more independent from OCaml version.

- added Pattern_runtime.check function for a friendlier type inference
  in patterns


kit-ty-kate:
Thanks

PR #14944, opened 2019-09-27
Update for conf-libclang

thierry-martinez:
- support LLVM versions 8 and 9.

- add config variables `config` and `version` that can be used by packages depending on this one to get the path to `llvm-config` and the detected version number.


mseri:
Thanks

PR #14947, opened 2019-09-27
Add piqilib-0.6.15

alavrik:


alavrik:
should be good to go. Can you merge this?

the test failure on osx is due to missing  CXXFLAGS='-std=c++14' and only triggered when protobuf is installed. I'll disable running protobuf-related tests by default in future versions so that opam doesn't stumble on problems like this

packages/piqilib/piqilib.0.6.15/opam
```diff
@@ -0,0 +1,38 @@
...
+  ["./configure" "--prefix" prefix]
+  [make]
+  [make "ocaml"]
+  [make "test"] {with-test}
+]
+install: [
+  [make "install"]
+  [make "ocaml-install"]
+]
+remove: [
```
mseri: This and the relative `flags` Are no longer needed since opam2

alavrik: @mseri can you clarify? Do you mean the whole remove section is no longer needed? How does opam know what to remove without it?

For the flags, do you mean the `flags: light-uninstall` line?

mseri: Yes, you can remove the whole remove and flags sections. Opam2 tracks all installed packages in its folder. In the end, you need to keep the remove and flags sections only if you are making changes to files in other packages that need to be reverted.

alavrik: @mseri this is done now

mseri:
Before I merge can you check if the failure on Travis osx test is expected?

alavrik:
@mseri yes, the test failure is expected. Ok to merge with it. See my first comment in this pull request for details.

PR #14953, opened 2019-09-29
mccs requires a C++ compiler

kit-ty-kate:
This PR also gets ride of the dependency on `which` to check whether g++ (or other C++ compilers linked as g++ are present) and use `--version` instead (this should be supported by most C++ compilers)


mseri:
The two failures are unrelated. Merging

PR #14960, opened 2019-09-30
ocp-browser.1.1.9: lambda-term upper constraint

reynir:
It works fine with lambda-term.2.0.2 when I tested it. The binary compiles and runs just fine for me.


reynir:
Sorry, nevermind - it seems my opam environment wasn't set up correctly...

PR #14971, opened 2019-10-01
Remove tests directory from LLVM sources.

jfeser:
LLVM ships with a pretty large directory of tests (~350Mb right now). The ocaml bindings don't need them to build, so this patch removes them before building.

### DIFF
--- a/packages/llvm/llvm.9.0.0/files/install.sh
+++ b/packages/llvm/llvm.9.0.0/files/install.sh
@@ -10,6 +10,8 @@ function filter_experimental_targets {
 }
 
 function llvm_install {
+    rm -rf test
+
     mkdir build
     cd build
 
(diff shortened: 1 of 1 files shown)


mseri:
You need to update the checksum in the opam file

mseri:
We can only test in on osx, and there it works

mseri:
@kit-ty-kate should we merge or move the changes to llvm-9.0.0

kit-ty-kate:
I'm sorry but I don't understand this PR. The tests aren't built by default, there is no point in removing them

jfeser:
If you have multiple opam switches, paying 350Mb per switch for useless LLVM tests is painful.

kit-ty-kate:
> If you have multiple opam switches, paying 350Mb per switch for useless LLVM tests is painful.

To the best of my knowledge, post-build directories are not kept around by opam (unless asked explicitly using `opam install -b`), only the sources are kept around and they can't be changed by any extra build rules.

Have you tested this locally? Where on your filesystem does it give a difference between before and after this PR?

jfeser:
Ah, I see. I misunderstood what opam was doing. I thought it was building in the sources directory. Thanks for clarifying!

This patch won't do what I wanted it to, so I'm closing the PR.

It'd be nice to do something about LLVM though. On a fresh 4.09 switch with llvm, base, and core installed, the tests directory is nearly 25% of the size of the switch.

PR #14973, opened 2019-10-02
added packages/parmap/parmap.1.0-rc11

UnixJunkie:
should compile even with OCaml 4.09.0

### DIFF
--- a/packages/parmap/parmap.1.0-rc11/opam
+++ b/packages/parmap/parmap.1.0-rc11/opam
@@ -1,0 +1,35 @@
+opam-version: "2.0"
+maintainer: "Roberto Di Cosmo <roberto@dicosmo.org>"
+authors: "Roberto Di Cosmo <roberto@dicosmo.org>"
+homepage: "https://github.com/rdicosmo/parmap"
+dev-repo: "git+https://github.com/rdicosmo/parmap.git"
+bug-reports: "https://github.com/rdicosmo/parmap/issues"
+build: [
+  ["./configure"] 
+  [make "DESTDIR=%{prefix}%" "OCAMLLIBDIR=lib" ]
+]
+install: [
+  [make "install" "DESTDIR=%{prefix}%" "OCAMLLIBDIR=lib"]
+]
+remove: [
+  ["./configure"] 
+  [make "uninstall" "DESTDIR=%{prefix}%" "OCAMLLIBDIR=lib"]
+]
+depends: [
+  "ocaml" {< "4.08.0"}
+  "ocamlfind" {build}
+  "ocamlbuild" {build}
+]
+synopsis: "Minimalistic library allowing to exploit multicore architecture"
+description: """
+Parmap is a minimalistic library allowing to exploit multicore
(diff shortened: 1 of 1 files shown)


mseri:
You need use `~rc11` Otherwise it will not be considered newer than the final release

UnixJunkie:
@mseri What do you mean? Where should I use '~rc11'?

mseri:
`parmap.1.0~rc11`

UnixJunkie:
I still don't understand. Is this the git tag that should be renamed?

mseri:
No, it’s just the package name: 1.0.0~rc11 < 1.0.0 < 1.0.0-rc11

kit-ty-kate:
The version mistake has already been made and can't be fixed without renaming the previous ones. The current version is fine as is for now

packages/parmap/parmap.1.0-rc11/opam
```diff
@@ -0,0 +1,35 @@
...
+dev-repo: "git+https://github.com/rdicosmo/parmap.git"
+bug-reports: "https://github.com/rdicosmo/parmap/issues"
+build: [
+  ["./configure"] 
+  [make "DESTDIR=%{prefix}%" "OCAMLLIBDIR=lib" ]
+]
+install: [
+  [make "install" "DESTDIR=%{prefix}%" "OCAMLLIBDIR=lib"]
+]
+remove: [
```
kit-ty-kate: The `remove` field is unnecessary since opam 2.0

packages/parmap/parmap.1.0-rc11/opam
```diff
@@ -0,0 +1,35 @@
...
+  ["./configure"] 
+  [make "uninstall" "DESTDIR=%{prefix}%" "OCAMLLIBDIR=lib"]
+]
+depends: [
+  "ocaml" {< "4.08.0"}
```
kit-ty-kate:
```suggestion
  "ocaml" {< "4.08.0"}
```
If the big change of this version is `should compile even with OCaml 4.09.0`, why is the constraint on ocaml still the same?

packages/parmap/parmap.1.0-rc11/opam
UnixJunkie: ok, removed

packages/parmap/parmap.1.0-rc11/opam
UnixJunkie: I removed the constraint

UnixJunkie:
@mseri I just named the package as the previous ones.
Maybe, the renaming should be made in another PR.
From what I see in the opam-repository, people don't depend on parmap with a specific version constraint on it, so it should not break anything.

XVilka:
@glondu FYI, for your Debian OCaml 4.08+ [transition project](https://wiki.debian.org/Teams/OCamlTaskForce/TODO).

UnixJunkie:
I think the tests are OK.
The broken revdeps are packages of mine that I can fix later (probably by adding an upper bound on their parany library dependency).

kit-ty-kate:
Thanks!

PR #14907, opened 2019-09-25
[new release] ounit2-lwt, ounit2, ounit-lwt and ounit (2.2.0)

gildor478:
OUnit testing framework (Lwt)

- Project page: <a href="https://github.com/gildor478/ounit">https://github.com/gildor478/ounit</a>
- Documentation: <a href="https://gildor478.github.io/ounit/">https://gildor478.github.io/ounit/</a>

### Changed
- Rename ounit/ounit-lwt OPAM and library to ounit2/ounit2-lwt. The META file
  to rename oUnit to ounit was not working on Windows and MacOSX because their
  filesystems are case insensitive and the install directories were the same.
  The new ounit2/ounit2-lwt packages avoid name clash on Windows/MacOSX and
  we still have ounit/ounit-lwt to allow the transition to the new package
  name. (Closes: gildor478/ounit#8)

PR #14985, opened 2019-10-02
Prevent gildor478/ounit#10, don't compile on OCaml 4.03.

gildor478:


kit-ty-kate:
I don't understand how this error is triggered. https://github.com/ocaml/opam-repository/pull/14907 went fine regarding the CI. Is this error only appearing when built directly with `dune build` without `-p` from source? I suppose it should be the case since dune would promote some warnings as errors in the case and that case only

gildor478:
Your explanation is possible and it might not affect ounit on opam. Up to you if you want to accept the PR or not. I was able to reproduce the error locally.

kit-ty-kate:
I think this safe to ignore. Thanks anyway though.

PR #14988, opened 2019-10-03
added packages/lz4_chans/lz4_chans.3.0.0

UnixJunkie:
updated dolog dependency version, also, should compile on more ocaml versions

### DIFF
--- a/packages/lz4_chans/lz4_chans.3.0.0/opam
+++ b/packages/lz4_chans/lz4_chans.3.0.0/opam
@@ -7,7 +7,9 @@ dev-repo: "git+https://github.com/UnixJunkie/lz4-chans.git"
 license: "BSD-3-Clause"
 build: [
   ["dune" "build" "-p" name "-j" jobs]
-  [make "test"] {with-test}
+# the tests don't pass on OCaml 4.03.0, because of a marshalling behavior
+# that was working before 4.03.0 and put back in place after 4.03.0
+  [make "test"] {with-test & ocaml:version != "4.03.0"}
 ]
 depends: [
   "base-unix"
--- a/packages/lz4_chans/lz4_chans.3.0.0/opam
+++ b/packages/lz4_chans/lz4_chans.3.0.0/opam
@@ -1,0 +1,26 @@
+opam-version: "2.0"
+authors: "Francois Berenger"
+maintainer: "unixjunkie@sdf.org"
+homepage: "https://github.com/UnixJunkie/lz4-chans"
+bug-reports: "https://github.com/UnixJunkie/lz4-chans/issues"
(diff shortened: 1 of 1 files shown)


packages/lz4_chans/lz4_chans.3.0.0/opam
```diff
@@ -0,0 +1,26 @@
+opam-version: "2.0"
+authors: "Francois Berenger"
+maintainer: "unixjunkie@sdf.org"
+homepage: "https://github.com/UnixJunkie/lz4-chans"
+bug-reports: "https://github.com/UnixJunkie/lz4-chans/issues"
+dev-repo: "git+https://github.com/UnixJunkie/lz4-chans.git"
+license: "BSD-3-Clause"
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+  [make "test"] {with-test}
```
kit-ty-kate:
```suggestion
  ["dune" "exec" "-p" name "src/test.exe"] {with-test}
```
`dune build` without `-p` or `--profile=release` promotes some warnings as errors. This should be avoided in releases.

kit-ty-kate:
There is an error with ocaml 4.03:
```

#=== ERROR while compiling lz4_chans.3.0.0 ====================================#
# context              2.0.5 | linux/x86_64 | ocaml-base-compiler.4.03.0 | pinned(https://github.com/UnixJunkie/lz4-chans/archive/v3.0.0.tar.gz)
# path                 ~/.opam/4.03/.opam-switch/build/lz4_chans.3.0.0
# command              /usr/bin/make test
# exit-code            2
# env-file             ~/.opam/log/lz4_chans-29-e7966f.env
# output-file          ~/.opam/log/lz4_chans-29-e7966f.out
### output ###
# dune build src/test.exe
# _build/default/src/test.exe
# 2019-10-03 07:29:10.676 INFO : plain_fn: /tmp/lz4_chans_test_3d0189.bin
# 2019-10-03 07:29:15.871 INFO : plain output: 1925094.56 float/s
# Fatal error: exception Failure("input_value: truncated object")
# make: *** [Makefile:22: test] Error 2
```

packages/lz4_chans/lz4_chans.3.0.0/opam
```diff
@@ -0,0 +1,26 @@
...
+]
+depends: [
```
kit-ty-kate:
```suggestion
depends: [
  "ocaml"
```

UnixJunkie:
I'll check what happens with 4.03.0.

UnixJunkie:
Ok, this is due to a change in behavior of the Marshal module that was introduced in 4.03.0.
It works prior to this version, and after.
I'll disable the tests just for that version, if I manage to do it.

kit-ty-kate:
Thanks!

PR #14996, opened 2019-10-04
Add upper bounds to irmin revdeps

pascutto:
This will prevent breakage when irmin.2.0.0 is released.


kit-ty-kate:
Sounds good. Thanks!

PR #14997, opened 2019-10-04
Restrict zmq-lwt and zmq-async to their respective zmq release

Leonidas-from-XIV:
In the last PR I was confused why the older `-lwt` and `-async` bindings were reverse-dependencies of the current `zmq` package and then it dawned on me that we didn't restrict an upper version in the older releases.

We started doing so in version `5.1.1` and this PR just extends this to older releases, so new releases of ZMQ won't be tested against old versions of the bindings. In *theory* that might work, but since these packages are all released from the same repo, tarball and at the same time and only tested against their respective versions I don't think it makes sense to have them as revdeps to new releases and having to bother with them whenever we send a PR.

### DIFF
--- a/packages/zmq-async/zmq-async.5.0.0/opam
+++ b/packages/zmq-async/zmq-async.5.0.0/opam
@@ -13,7 +13,7 @@ build: [
 ]
 depends: [
   "ocaml"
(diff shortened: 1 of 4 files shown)


kit-ty-kate:
Thanks!

PR #15000, opened 2019-10-04
Release clangml-transforms.0.24

thierry-martinez:
New version compatible with Clangml 4

### DIFF
--- a/packages/clangml-transforms/clangml-transforms.0.24/opam
+++ b/packages/clangml-transforms/clangml-transforms.0.24/opam
@@ -16,8 +16,9 @@ synopsis: "Code transformers for clangml"
 depends: [
   "clangml" {>= "4.0.1"}
   "visitors" {>= "20190711"}
+  "dolog" {>= "4.0.0"}
 ]
 url {
   src: "https://gitlab.inria.fr/memcad/clangml-transforms/-/archive/0.24/clangml-transforms-0.24.tar.gz"
-  checksum: "sha512=d7abbdd0841a3eaa463820676d494e6aa833aa6e54794ac14521d0cb9ca979f1c579be35b55913d8fd97113af8c12ffec1e49d31d2267e3d9aa84890798d2665"
+  checksum: "sha512=f103b1b2e60436c1fcaa07cb0364ed9b07d8fd6840eec33a44adacfece998125737366613cb6ec6bd678d73a29c444b427596a7e6bc10073009a74f7cb496fd3"
 }
--- a/packages/clangml-transforms/clangml-transforms.0.24/opam
+++ b/packages/clangml-transforms/clangml-transforms.0.24/opam
@@ -14,6 +14,8 @@ license: "BSD"
 dev-repo: "git+https://gitlab.inria.fr/memcad/clangml-transforms"
 synopsis: "Code transformers for clangml"
 depends: [
+  "ocaml"
+  "dune" {>= "1.10"}
   "clangml" {>= "4.0.1"}
   "visitors" {>= "20190711"}
   "dolog" {>= "4.0.0"}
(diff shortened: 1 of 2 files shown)


packages/clangml-transforms/clangml-transforms.0.24/opam
```diff
@@ -0,0 +1,24 @@
...
+]
+maintainer: ["Thierry Martinez <thierry.martinez@inria.fr>"]
+authors: ["Thierry Martinez <thierry.martinez@inria.fr>"]
+bug-reports: "https://gitlab.inria.fr/memcad/clangml-transforms/issues"
+homepage: "https://gitlab.inria.fr/memcad/clangml-transforms"
+doc: "https://gitlab.inria.fr/memcad/clangml-transforms"
+license: "BSD"
+dev-repo: "git+https://gitlab.inria.fr/memcad/clangml-transforms"
+synopsis: "Code transformers for clangml"
+depends: [
```
kit-ty-kate:
```suggestion
depends: [
  "ocaml"
  "dune" {>= "1.10"}
```

packages/clangml-transforms/clangml-transforms.0.24/opam
```diff
@@ -0,0 +1,24 @@
+opam-version: "2.0"
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+  ["dune" "build" "-p" name "@doc"] {with-doc}
```
kit-ty-kate:
```suggestion
  ["dune" "build" "-p" name "@doc" "-j" jobs] {with-doc}
```

thierry-martinez:
@kit-ty-kate Thanks for your suggestions for the opam file. This opam file is generated by `dune` from the `dune-project` file: do you think it worth that I make a feature-request for `dune` to change how the opam file is generated (especially for the `"-j"` option with `with-doc`, but a lower-bound for `dune` version could be systematically added as well)?

kit-ty-kate:
I think it might be worth doing it yes: Everything regarding opam files generation are here I think: https://github.com/ocaml/dune/blob/master/src/dune/opam_create.ml

Do you want me to do it or are you motivated to do it?

kit-ty-kate:
oh, looking at the code, I think everything that you want (except `"-j" jobs`) is already done. But for some reason they require people to require at least dune 1.11 (marked in their dune-project file)

kit-ty-kate:
ok so the version requirement is here on purpose and the output shouldn't be changed so it's not worth spending time looking into it

thierry-martinez:
> ok so the version requirement is here on purpose and the output shouldn't be changed so it's not worth spending time looking into it

Thank you very much for having checked that: I will mark `dune-project` files with `1.11` from now on. Where did you find the purpose of the version requirement? (I was unable to find the explanation related to the commit https://github.com/ocaml/dune/commit/284dc0ba20c4bc9b036aab2c590f9a1bf43da898 ) Thanks again!

thierry-martinez:
The compiler failure on `memcad` seems to be due to the uninstallability of older versions of `clangml-transforms` which rely on old versions of LLVM that are no longer packaged. A new version of `memcad` should be released soon, which will use the new `clangml-transforms`.

kit-ty-kate:
Thanks a lot!

kit-ty-kate:
> Where did you find the purpose of the version requirement? (I was unable to find the explanation related to the commit [ocaml/dune@284dc0b](https://github.com/ocaml/dune/commit/284dc0ba20c4bc9b036aab2c590f9a1bf43da898) ) Thanks again!

I asked the authors ^^"

PR #14999, opened 2019-10-04
[new release] git-http, git, git-unix and git-mirage (2.1.1)

dinosaure:
Client implementation of the "Smart" HTTP Git protocol in pure OCaml

- Project page: <a href="https://github.com/mirage/ocaml-git">https://github.com/mirage/ocaml-git</a>
- Documentation: <a href="https://mirage.github.io/ocaml-git/">https://mirage.github.io/ocaml-git/</a>

##### CHANGES:

- Replace Pervasives by Stdlib and depends on `stdlib-shims` virtual package (@dinosaure, mirage/ocaml-git#372)
- Add `pp_fetch_one` and `pp_update_and_create` (@dinosaure, @pascutto, @samoht, mirage/ocaml-git#370)
- `git-mirage` depends on `conduit-mirage` instead `mirage-conduit` (@dinosaure, mirage/ocaml-git#371)
- Drop support of OCaml < 4.07.0
- Delete `mirage-fs-unix` useless dependency
- Constraint to use only `decompress.0.9.0`

PR #15009, opened 2019-10-07
git*: fix issues spotted by @kit-ty-kate on #14999

mseri:


kit-ty-kate:
Thanks a lot!

PR #15014, opened 2019-10-08
[new release] Solo5 0.6.3

mato:
This should fix the build problems seen on OpenSUSE CI, by using pkg-config to find libseccomp.

/cc @avsm

### DIFF
--- a/packages/solo5-bindings-genode/solo5-bindings-genode.0.6.3/opam
+++ b/packages/solo5-bindings-genode/solo5-bindings-genode.0.6.3/opam
@@ -1,0 +1,46 @@
+opam-version: "2.0"
+maintainer: "ehmry@posteo.net"
+authors: [
+  "Emery Hemingway <ehmry@posteo.net>"
+]
+homepage: "https://github.com/solo5/solo5"
+bug-reports: "https://github.com/solo5/solo5/issues"
+license: "ISC"
+dev-repo: "git+https://github.com/solo5/solo5.git"
+build: [
+  ["./configure.sh"]
+  [make "V=1" "CONFIG_HVT=" "CONFIG_SPT=" "CONFIG_VIRTIO=" "CONFIG_MUEN="]
+]
+install: [make "V=1" "CONFIG_HVT=" "CONFIG_SPT=" "CONFIG_VIRTIO=" "CONFIG_MUEN=" "install-opam-genode" "PREFIX=%{prefix}%"]
+remove: [
+  ["touch" "./Makeconf"]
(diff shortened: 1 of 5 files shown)


mato:
Ok, so the CI failure is due to https://github.com/Solo5/solo5/commit/1239f3eff2fbd6ee03e409c28d47dfa4d4fc1c5d#diff-735003ab6f5797c8b8bdecbbbb9322a8R182 where I effectively changed the logic from "Don't configure/compile spt if we have no libseccomp" to "Complain if we have no libseccomp".

Without going on a larger overhaul of the `configure.sh` / `Makefile` chain and adding something like `--disable-<target>`, the easiest solution would be to add the seccomp depexts to all the bindings variants. Any objections to this approach?

hannesm:
> add the seccomp depexts to all the bindings variants. Any objections to this approach?

it'd be good to have a `conf-libseccomp` package then, and the binding variants depending on that (instead of having the libseccomp depexts in each variant) (the `conf-libseccomp` would be similar to `conf-libcurl` and rarely being updated)

mato:
> it'd be good to have a conf-libseccomp package then

That's a good idea. Done in 21db5c4 and `depends:` updated to match.

mato:
Not sure what's up with the CI failures here, there are a lot of them and randomly clicking on some shows:

```
=== opam install -t solo5-bindings-spt.0.6.3
[2019-10-09 11:18.47] Starting...

=== Using base image sha256:818b29c754187962d2a0b2006a404f4f8bba59e1290d8ce73a2cac237ccd805d

=== docker run --rm -v opam2-archive:/home/opam/opam-repository/cache -e
OPAMWITHTEST=true
sha256:818b29c754187962d2a0b2006a404f4f8bba59e1290d8ce73a2cac237ccd805d opam
exec -- opam-ci-install
solo5-bindings-spt.0.6.3
[2019-10-09 11:21.02] Cancelled at user's request
[2019-10-09 11:21.02] Failed: Cancelled
```

kit-ty-kate:
All jobs should be restarted now. Sorry for that

mato:
The CentOS failures are expected and only affect `solo5-bindings-*`.

hannesm:
thanks!

hannesm:
FWIW, now that we have conf-libseccomp, this could be used by earlier (0.6.2) solo5-bindings-* releases

mato:
> FWIW, now that we have conf-libseccomp, this could be used by earlier (0.6.2) solo5-bindings-* releases

Personally I prefer to avoid retroactively changing OPAM metadata of previously released versions unless absolutely necessary.

hannesm:
fine with me as well

PR #15026, opened 2019-10-09
Remove constraint on digestif.0.7.4 with mirage.3.*

dinosaure:
digestif.0.7.4 is a special version to keep a compatibility package with MirageOS 3. Only digestif.0.7.3 and digestif.0.8.0 have a conflict with MirageOS 3.

### DIFF
--- a/packages/digestif/digestif.0.7.4/opam
+++ b/packages/digestif/digestif.0.7.4/opam
@@ -47,7 +47,7 @@ depopts: [
 conflicts: [
   "mirage-xen-posix" {< "3.1.0"}
   "ocaml-freestanding" {< "0.4.1"}
-  "mirage-runtime"{< "4.0.0"}
+  "mirage-runtime" {>= "4.0.0"}
 ]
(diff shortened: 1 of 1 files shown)


packages/digestif/digestif.0.7.4/opam
```diff
@@ -47,7 +47,6 @@ depopts: [
 conflicts: [
   "mirage-xen-posix" {< "3.1.0"}
   "ocaml-freestanding" {< "0.4.1"}
-  "mirage-runtime"{< "4.0.0"}
```
hannesm: shouldn't it be `"mirage-runtime" {>= "4.0.0"}` instead then? or is 0.7.4 working with both mirage3 and mirage4?

dinosaure: Hmmhmm, my workspace becomes some complex, I will follow your advise :+1: and be conservative.

kit-ty-kate:
Thanks

PR #15031, opened 2019-10-10
travis: extend opam timeout

Blaisorblade:
Let opam continue when running >10minutes without output, like it happens when a single package install runs that long.

This extends timeout to 20 minutes. `travis_wait $N` would give N minutes.
https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

This would help for instance with https://travis-ci.org/ocaml/opam-repository/jobs/596032476.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: c
 install: bash -e .travis-ci-install.sh
-script: bash -e .travis-ci.sh
+script: travis_wait bash -e .travis-ci.sh
 cache: apt
 dist: xenial
 sudo: required


kit-ty-kate:
Thanks a lot!

kit-ty-kate:
For the record the failing job you're pointing out is not going to be helped by this PR. The limit of 50 minutes is an hard limit for all travis processes and it doesn't have to do anything with no-output limit.
However I still think this PR is worth merging, we might even get rid of the hack for mac that is still somewhere in the code that was used to counter that

PR #15033, opened 2019-10-10
Disable tests of zmq-lwt 5.0.0

aantron:
`zmq-lwt` 5.0.0 tests fail with [an error](https://ci.ocaml.org/log/saved/docker-run-6b01f06e84cd18f5cc979bbd20653a12/1eed241baf1730c6e7c19bea05ce64cc2064349d) that we likely can't fix with a constraint, so this PR disables them. There are newer releases of `zmq-lwt`, and their tests pass.


kit-ty-kate:
Thanks

PR #15037, opened 2019-10-11
git-mirage.2.0 is not compatible with git.2.1.1

samoht:
See http://check.ocamllabs.io/log/1570453959-f412620d14e52a588975517bf940aea115523f44/4.04/bad/git-mirage.2.0.0


kit-ty-kate:
Thanks!

PR #15047, opened 2019-10-12
Package ocaml-protoc-plugin.1.0.0

andersfugmann:
### `ocaml-protoc-plugin.1.0.0`
- enums are mapped to adt's

---

---


andersfugmann:
Broken. I'll create a new PR, as I also need conf-protoc.

PR #15048, opened 2019-10-12
ocaml-protoc-plugin.1.0.0 and conf-protoc.1.0.0

andersfugmann:


packages/ocaml-protoc-plugin/ocaml-protoc-plugin.1.0.0/opam
```diff
@@ -0,0 +1,41 @@
...
+dev-repo: "git+https://github.com/issuu/ocaml-protoc-plugin"
+bug-reports: "https://github.com/issuu/ocaml-protoc-plugin/issues"
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+]
+depends: [
+  "conf-protoc"
+  "dune" {>= "1.10" & build}
```
kit-ty-kate:
```suggestion
  "dune" {>= "1.10"}
```
Dune should not be tagged as a build dependency (see #14266)

andersfugmann:
Thanks for fixing the conf package (and for the suggested change of dune not beeing a build time dependency)

kit-ty-kate:
Thanks!

PR #15051, opened 2019-10-13
restrict tcpip to mirage-protocols <3.1.0 (to-be-released)

hannesm:


hannesm:
NB: the CI failure is expected (tcpip tests on macOS -- I don't have such hardware, and it looks like others don't have time/interest to investigate) -- good to merge!

PR #15017, opened 2019-10-08
[new release] Add coq.8.10.0 and coqide.8.10.0

Blaisorblade:
Warning: I am not sure I know what I'm doing, but I took the contents of
https://coq.inria.fr/opam/core-dev/packages/coq/coq.8.10.dev/ and
https://coq.inria.fr/opam/core-dev/packages/coqide/coqide.8.10.dev/ (which
currently builds essentially the same version, successfully), and
updated URL and checksum following 8.9.1. I used sha512 for both checksums,
following *coqide.8.10+beta3/coq.8.10+beta3 (EDITED).

I already built coq locally, tho I was still using the md5 checksum — but IIUC PRs are checked by CI anyway? Or should I rebuild it locally from scratch?

PR #15054, opened 2019-10-14
coq.8.10: backport fix to parallel build

Blaisorblade:
Followup to #15017.
The backported patch matches https://github.com/vbgl/coq/commit/d25990e2da7e5e84034b927cdab9ea4019c56a30 from coq/coq#9963, backporting coq/coq#10881; I checked it makes sense also for 8.10 and I tested this locally as far as possible. But CI here is much better at finding issues with parallel builds.

/cc @kit-ty-kate.


kit-ty-kate:
Thanks a lot! Looks like it works!

PR #15057, opened 2019-10-14
[new release] functoria and functoria-runtime (2.2.5)

hannesm:
CHANGES:

* Functoria_runtime.with_argv now uses (mirage/functoria#179, by @hannesm)
  - exit 63 when `Help or `Version is requested (used to exit with 0)
  - exit 64 when Term.eval returns with an error (used to raise an exception)


kit-ty-kate:
Thanks!

PR #15069, opened 2019-10-17
simpler conf-rdkit package

UnixJunkie:


kit-ty-kate:
I think this requires python2 not python3
```
#=== ERROR while compiling conf-rdkit.1.0.0 ===================================#
# context              2.0.5 | linux/x86_64 | ocaml-base-compiler.4.09.0 | pinned
# path                 ~/.opam/4.09/.opam-switch/build/conf-rdkit.1.0.0
# command              /usr/bin/python3 test.py
# exit-code            1
# env-file             ~/.opam/log/conf-rdkit-23-6d07ae.env
# output-file          ~/.opam/log/conf-rdkit-23-6d07ae.out
### output ###
#   File "test.py", line 2, in <module>
#     import rdkit
# ModuleNotFoundError: No module named 'rdkit'

- Traceback (most recent call last):
-   File "test.py", line 2, in <module>
-     import rdkit
- ModuleNotFoundError: No module named 'rdkit'
```

UnixJunkie:
Note that this will never work on the Mac, since brew needs to use a special tap.
Also, the version of rdkit on Mac supports only Python3.
The rdkit packages on Debian and Centos are so old...

UnixJunkie:
Interesting. Still not working.
I think I should rename it conf-rdkit-python-2, maybe.
Because one day, there should be conf-rdkit-python-3.

UnixJunkie:
The problem is that many distros don't have yet the Python-3 ready version of rdkit.

UnixJunkie:
I think I could make this work only if the latest Ubuntu or Debian was in the test matrix.

UnixJunkie:
I will submit something better soon.

Issue #15055, opened 2019-10-14
Heartbeat issues on Travis


Blaisorblade:
Job https://travis-ci.org/ocaml/opam-repository/jobs/597614609#L1050-L1054
a3733e720f2fdf030c8bc4c89b18e97e926e7334 (and others) timed out for mysterious reasons.

Thanks to #11665, opam should output each executed command, so jobs shouldn't appear stuck. That should make unnecessary both  a3733e720f2fdf030c8bc4c89b18e97e926e7334 (a heartbeat) and possibly #15031 (tho that depends on how verbose build scripts are).
If a single command in a build script takes longer than 10 minutes, the build is probably truly stuck.

However, somehow, none of that works. Could we be running into https://docs.travis-ci.com/user/common-build-problems/#limitations-of-travis_wait?
EDIT: in fact, https://travis-ci.org/ocaml/opam-repository/jobs/596032476 got to 50 minutes (the max. time limit).

Separately, it seems the `opam install --deps-only` in `.travis-ci.sh` needs a `-v` option, or maybe we should just set `OPAMVERBOSE=y` ahead.
When building coqide, that command will build coq, so it will timeout if it's not verbose.

Blaisorblade:
Should be fixed by #15075.

PR #15072, opened 2019-10-17
Travis: Allocate more time per command

kit-ty-kate:
Extend time again. See https://github.com/ocaml/opam-repository/pull/15031
Let's see if that helps

cc @Blaisorblade

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: c
 install: bash -e .travis-ci-install.sh
-script: travis_wait bash -e .travis-ci.sh
+script: travis_wait 40 bash -e .travis-ci.sh
 cache: apt
 dist: xenial
 sudo: required


Blaisorblade:
See #15055 — something seems wrong right now, and I suspect my PR already introduced a regression.

I haven't analyzed the issue in depth, but using `export OPAMVERBOSE=y` before calling opam should be superior — if verbose opam gives no output for 10 minutes, that's probably a good reason to cause a timeout.
Right now, some opam calls get `-v`, but not all!

Blaisorblade:
#15075 is what I have in mind, but feel free to take over that PR and/or cherry-pick its commits, as needed. That might take extra testing and debugging, also on other troublesome packages, and I probably don't have time for that currently.

kit-ty-kate:
Closing this in favour of #15075

PR #15073, opened 2019-10-17
coqide8.10.0: use parallel build patch from coq8.10.0

### LINKED ISSUE #15070: coqide.8.10 should use the same patch as coq.8.10 to avoid parallel build errors
I hit the "gramlib" bug when trying to install coqide.8.10.0 using opam (cf. https://github.com/ocaml/opam-repository/pull/15017). The bug was circumvented for coq by adding `-j1`. The same fix should be applied to coqide. (And indeed the installation is successful if I add `OPAMJOBS=1` to the command line)

Blaisorblade:
Fix #15070.
Apparently, building coqide builds enough of coq that this patch is needed;
this is confirmed by the verbose build log.


Blaisorblade:
Sorry for the race @eponier!

eponier:
No problem. You won that one !

Blaisorblade:
The failure on Travis is a timeout, and is fixed by #15075 (as the tests there confirm).

kit-ty-kate:
Thanks a lot!

PR #15075, opened 2019-10-17
Make opam always verbose to prevent Travis timeouts

Blaisorblade:
Alternative to #15072, fixing #15055. I'm adding coqide changes from #15073 to
test this, but don't merge that yet.

### DIFF
--- a/.travis-ci.sh
+++ b/.travis-ci.sh
@@ -2,7 +2,7 @@
 source .travis-ocaml.env
 
 # Ensure build logs are printed live
-export OPAMVERBOSE=y
+export OPAMVERBOSE=yes
 
 # display info about OS distribution and version
 case $TRAVIS_OS_NAME in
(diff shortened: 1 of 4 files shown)


.travis-ci.sh
```diff
@@ -2,7 +2,7 @@
 source .travis-ocaml.env
 
 # Ensure build logs are printed live
-export OPAMVERBOSE=y
+export OPAMVERBOSE=yes
```
Blaisorblade: Apologies for not being paranoid enough.

Blaisorblade:
And it passed on Travis!

FWIW: if useful elsewhere, one could probably readd either the busy loop or `travis_wait` (better not both), but making things more verbose seems more robust. Your call!

kit-ty-kate:
It seems pretty good with verbose on. Thanks a lot for that!

PR #15077, opened 2019-10-17
Patch JS packages for upcoming dune 2

xclerc:
The upcoming dune 2 will feature a few breaking changes,
that will affect Jane Street packages. This pull request updates:

- `patdiff` by adding a `(modules_without_implementation ...)` stanza;
- `re2` and `zstandard` by adding a `(mode fallback)` stanza.

### DIFF
--- a/packages/re2/re2.v0.12.1/opam
+++ b/packages/re2/re2.v0.12.1/opam
@@ -14,6 +14,7 @@ depends: [
   "core_kernel" {>= "v0.12" & < "v0.13"}
   "ppx_jane"    {>= "v0.12" & < "v0.13"}
   "dune"        {>= "1.5.1"}
+  "conf-g++" {build}
 ]
 synopsis: "OCaml bindings for RE2, Google's regular expression library"
 description: "
--- a/packages/re2/re2.v0.12.1/opam
+++ b/packages/re2/re2.v0.12.1/opam
@@ -1,0 +1,28 @@
+opam-version: "2.0"
+maintainer: "opensource@janestreet.com"
+authors: ["Jane Street Group, LLC <opensource@janestreet.com>"]
+homepage: "https://github.com/janestreet/re2"
+bug-reports: "https://github.com/janestreet/re2/issues"
+dev-repo: "git+https://github.com/janestreet/re2.git"
+doc: "https://ocaml.janestreet.com/ocaml-core/latest/doc/re2/index.html"
+license: "MIT"
+build: [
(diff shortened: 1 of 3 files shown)


packages/re2/re2.v0.12.1/opam
```diff
@@ -0,0 +1,28 @@
...
+license: "MIT"
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+]
+depends: [
+  "ocaml"       {>= "4.07.0"}
+  "core_kernel" {>= "v0.12" & < "v0.13"}
+  "ppx_jane"    {>= "v0.12" & < "v0.13"}
+  "dune"        {>= "1.5.1"}
+]
```
kit-ty-kate:
```suggestion
  "conf-g++" {build}
]
```
It might be worth replacing the depexts with this

kit-ty-kate:
Also, if those packages are currently incompatible with dune 2, would it be possible for you to send a PR restricting the affected version to dune < 2.0.0. Or do you think it's too early to be sure that they are going to be incompatible? I don't know, as you wish.

xclerc:
> Also, if those packages are currently incompatible with dune 2, would it be possible for you to send a PR restricting the affected version to dune < 2.0.0. Or do you think it's too early to be sure that they are going to be incompatible? I don't know, as you wish.

I was not sure what was the best course of action there.
I have a handful of packages known to be incompatible,
but am not sure the ongoing work on dune will not create
more incompatibilities. Anyway, if that's OK with you, I
would rather do that through another PR.

kit-ty-kate:
Could you return the change I just made to the opam file upstream?

> Anyway, if that's OK with you, I would rather do that through another PR.

Sure!

xclerc:
> Could you return the change I just made to the opam file upstream?

Sorry, what do you mean?

kit-ty-kate:
> Sorry, what do you mean?

I committed something to this PR: https://github.com/ocaml/opam-repository/pull/15077/commits/5a179e5924e259d91d8e449eaf5f59c7da84be10

The depexts field is not necessary since `conf-g++` has been added to the dependency, in my comment on the suggestion I said it was to replace the depexts. And I also noticed that the `description` field was empty and I think it's better if only `synopsis` is taken.

It would be great if the opam file in the repository upstream is as close as possible to the one on the opam-repository to avoid asking for the same change everytime there is a new release.

kit-ty-kate:
Sorry for the delay

kit-ty-kate:
Thanks!

PR #15043, opened 2019-10-11
[new release] postgresql (4.5.1)

mmottl:
Bindings to the PostgreSQL library

- Project page: <a href="https://mmottl.github.io/postgresql-ocaml">https://mmottl.github.io/postgresql-ocaml</a>
- Documentation: <a href="https://mmottl.github.io/postgresql-ocaml/api">https://mmottl.github.io/postgresql-ocaml/api</a>

##### CHANGES:

* Fixed warnings in C-stubs

  * Support detection of release candidate version numbers

PR #15080, opened 2019-10-18
session <= 3.2: add upper bound on lwt

mseri:
Should fix some of the failures in https://github.com/ocaml/opam-repository/pull/15043


mseri:
Well spotted. Should we add the >= v0.13 bound in the conflicts then?

kit-ty-kate:
The `>= v0.13` bounds are not necessary, we'll see when a new version of those packages will actually be there

kit-ty-kate:
Thanks a lot!

PR #15084, opened 2019-10-20
Revert #14618, see if jsoo works in 4.09

hhugo:


hhugo:
@xclerc, do you remember why you created https://github.com/ocaml/opam-repository/pull/14618 ?

xclerc:
I do not remember the details, but I think I was unable to
get `js_of_ocaml` work on a 4.09 switch. It might have
been with a beta though; did you close the issue because
this version of `js_of_ocaml` is indeed incompatible with
4.09?

hhugo:
> did you close the issue because
this version of js_of_ocaml is indeed incompatible with
4.09?

Yes, there is an issue to compile the ppx.

PR #15082, opened 2019-10-19
prepare for next mirage-types release

hannesm:
prepare for upcoming and breaking mirage-types release (see mirage/mirage#1004):
    opam admin add-constraint "mirage-device<2.0.0"
    opam admin add-constraint "mirage-random<2.0.0"
    opam admin add-constraint "mirage-time<2.0.0"
    opam admin add-constraint "mirage-clock<3.0.0"
    opam admin add-constraint "mirage-flow<2.0.0"
    opam admin add-constraint "mirage-fs<3.0.0"
    opam admin add-constraint "mirage-console<3.0.0"
    opam admin add-constraint "mirage-protocols<4.0.0"
    opam admin add-constraint "mirage-stack<2.0.0"
    opam admin add-constraint "mirage-block<2.0.0"
    opam admin add-constraint "mirage-net<3.0.0"
    opam admin add-constraint "mirage-kv<3.0.0"
    opam admin add-constraint "mirage-channel<4.0.0"
    opam admin add-constraint "mirage-types<3.7.0"
    opam admin add-constraint "mirage-types-lwt<3.7.0"

PR #15085, opened 2019-10-20
some packages have optional dependencies on mirage-types packages

hannesm:
restrict them to released versions as well (not handled by opam admin add-constraint), continuation of #15082


hannesm:
it seems the CI is slightly broken "can't find image with sha yyy" -- I went through lots of failing logs, but couldn't find any legitimate ones (some of the packages touched in this PR do not compile due to missing bounds, but they did neither compile before this PR). It'd be good to have this merged, alternatively let me know which failure must be fixed to get this merged.

kit-ty-kate:
I think this can be merged safely. Errors are either unrelated or due to unused docker images being cleaned up in the middle of a CI run. @talex5 do you think docker-gc be a better solution for that?

kit-ty-kate:
thanks!

talex5:
@kit-ty-kate isn't docker gc what we're using already? I think the problem is that the ci machine just doesn't have enough disk space to build big PRs like this all in one go. @avsm mentioned something about getting an extra disk for it.

avsm:
@talex5 The problem here is that a `docker gc` will remove intermediate images which are depended on directly by a later mirage-ci job that assumes that sha256 hash will be around.

talex5:
Actually, I think this PR wasn't using too much disk space. It just had the bad luck to be running at midnight when the daily docker gc kicked in. I was thinking of a different PR earlier that day that consumed most of the disk space and required a manual GC.

PR #15086, opened 2019-10-20
Package ocaml-protoc-plugin.2.0.0

andersfugmann:
### `ocaml-protoc-plugin.2.0.0`
- enums are mapped to adt's

---

---


andersfugmann:
CentOS fails as its running centos7, which is not supported as the needed files are not distrubuted with this version. The solution would be to restrict centos to version >= 8, but I don't know how to do this. I've tried:

```
available: (os-distribution != "ubuntu" | os-version >= "18.04") & (os-distribution != "centos" | os-version >= "8")
```
but it does not seem to restrict centos to version >= 8. depext do not by default output the `os-version`, so its hard to see how the `os-version` variable is set, and what to match on.

andersfugmann:
AFAICS, depext sets `os-release`to the value of the key `VERSION_ID` in `/etc/os-release` ([src/state/opamSysPoll.ml](https://github.com/ocaml/opam/blob/master/src/state/opamSysPoll.ml#L119)).

For centos7, the value is `"7"` and for centos8 its `"8"`, so I'm a but puzzled as to why the package can be installed on cenos7.

andersfugmann:
```

andersfugmann:
Ahh. I think I found the problem. `ocaml-protoc-plugin` did not have a version constraint on conf-protoc, so the previous version (1.0.0) was installed.

andersfugmann:
All green and ready to merge.

kit-ty-kate:
Is there a reason why `conf-protoc` needs to have a new version? `1.0.0` and `1.0.1` seems worryingly identical.

andersfugmann:
Version 1.0.0 are supposed to have google's well know .proto files included along with the protoc compiler (its actually required by the spec that these are available). However, ocaml-protoc-plugin now relies on these protoc files, and I discovered that these were not available on centos7 and some depends were missing. 
To avoid revisionism, I elected to create a 1.0.1 (semantic versioning - fixing bugs). 

If you prefer, I can port the changes to conf-protoc 1.0.0.

andersfugmann:
@kit-ty-kate If you need further information, or would like me to change something, just let me know.

kit-ty-kate:
I'm really sorry for the time I took to reply. I ported the change to `conf-protoc.1.0.0`, that seems to be a really minimal change, I think it's safe to just port it. If it's really a problem in the future we can still split in two packages at some point in the future.

andersfugmann:
Look good to me. Thanks a lot! - and no worries about the delay.
I've added `flags: conf` to conf-protoc.0.0.9 also. Once green it should be ready to go.

kit-ty-kate:
Oh I missed that one. Thanks!

PR #15087, opened 2019-10-20
z3 opam package update

philzook58:
Updates to the opam z3 packages, including an experimental nightly package
Please look these over @c-cube


packages/z3/z3.4.8.6/opam
```diff
@@ -0,0 +1,30 @@
...
+license: "MIT"
+dev-repo: "git+https://github.com/Z3prover/z3.git"
+build: [
+  [ "python2.7" "scripts/mk_make.py" "--ml" "--staticlib" ]
+  [ make "-C" "build" "-j" jobs ]
+]
+install: [
+  [ "sh" "-c" "ocamlfind install z3 build/api/ml/META -nodll build/libz3* build/api/ml/*" ]
+]
+remove: ["ocamlfind" "remove" "z3"]
```
kit-ty-kate:
```suggestion
```
The `remove` field is optional since opam 2.0

packages/z3/z3.4.8.6/opam
```diff
@@ -0,0 +1,30 @@
...
+depends: [
+  "ocaml"
+  "ocamlfind" {build}
+  "num"
+  "conf-gmp"
+  "conf-python-2-7" {build}
+]
+synopsis: "Z3 solver"
+flags: light-uninstall
```
kit-ty-kate:
```suggestion
```
If the remove field is removed you won't need that

packages/z3/z3.~nightly/opam
```diff
@@ -0,0 +1,28 @@
...
+license: "MIT"
+dev-repo: "git+https://github.com/Z3prover/z3.git"
+build: [
+  [ "python2.7" "scripts/mk_make.py" "--ml" "--staticlib" ]
+  [ make "-C" "build" "-j" jobs ]
+]
+install: [
+  [ "sh" "-c" "ocamlfind install z3 build/api/ml/META -nodll build/libz3* build/api/ml/*" ]
+]
+remove: ["ocamlfind" "remove" "z3"]
```
kit-ty-kate:
```suggestion
```

packages/z3/z3.~nightly/opam
```diff
@@ -0,0 +1,28 @@
...
+depends: [
+  "ocaml"
+  "ocamlfind" {build}
+  "zarith"
+  "conf-gmp"
+  "conf-python-2-7" {build}
+]
+synopsis: "Z3 solver"
+flags: light-uninstall
```
kit-ty-kate:
```suggestion
```

kit-ty-kate:
I don't think having `z3.~nightly` is a good idea. I think we even have rules against that.
The exact same thing can be gotten with just `opam pin add --dev z3`. It'll take the current git master listed in `dev-repo`. This would be the more classical way of doing this.

philzook58:
I made the requested changes. The dependencies of the nightly version have changed though. It switch from num to zarith.

kit-ty-kate:
> I made the requested changes. The dependencies of the nightly version have changed though. It switch from num to zarith.

mh, and I see there is no opam file in the repository itself, this is indeed an issue. Usually the problem is fixed by just modifying the opam file in the repository accordingly, this way the changes would be picked up by opam. Is there no way of adding an opam file to the z3 repository?

philzook58:
I can ask. I rather doubt they would want an opam file at the top level of the project. Ocaml is not a  priority of the project. 

Would opam be able to find the opam file deeper in the hierarchy somehow? I think maybe they might accept one in this directory https://github.com/Z3Prover/z3/tree/master/src/api/ml

In any case, while I need the latest z3 fixes, not everyone does. So I think if the new z3.4.8.6 seems acceptable as is, it is still useful. And this particular dependency issue should be easily resolved when the next release of z3 is cut even if I can't get an opam file into z3 itself.

c-cube:
I made the suggestion to have a `z3-nightly` alternative package, for people who need it. That'd be less misleading to casual users.

kit-ty-kate:
Thanks!

PR #15100, opened 2019-10-22
camlzip version 1.09

leamingrad:
Version 1.09 of camlzip was released a few weeks ago

* Add a Gzip.flush_continue function that allows the user to flush
  the contents of the zip buffer all the way to disk but then continue
  writing to the zipped channel.


kit-ty-kate:
The directory should be named `camlzip.1.09` not `camlzip1.09`

leamingrad:
Fixed - thanks for spotting that

kit-ty-kate:
Errors are unrelated. Thanks!

PR #15107, opened 2019-10-24
async_udp.v0.12.0

Leonidas-from-XIV:
Please note that this package is meant to be replaced once the official async_udp package from Jane Street is released.

### DIFF
--- a/packages/async_udp/async_udp.v0.12.0/opam
+++ b/packages/async_udp/async_udp.v0.12.0/opam
@@ -1,0 +1,19 @@
+opam-version: "2.0"
+synopsis: "Async_udp shim"
+description: "
+Temporary shim to get back the Async Udp lost from async_extra in v0.12.
+"
+maintainer: "Michael Bacarella <michael.bacarella@gmail.com>"
+authors: ["Michael Bacarella <michael.bacarella@gmail.com>"]
+homepage: "https://github.com/mbacarella/async_udp"
+bug-reports: "Michael Bacarella <michael.bacarella@gmail.com>"
+build: ["dune" "build" "-p" name "-j" jobs]
+depends: [
+  "async" {>= "v0.12" & < "v0.13"}
+  "dune"  {>= "1.11"}
+  "ppx_jane" {>= "v0.12" & < "v0.13"}
+]
+url {
+  src: "https://github.com/mbacarella/async_udp/tarball/aa3607fc55dab9f36de7730622d55e48fc96a23f"
+  checksum: "md5=77898f384a7a31a227738a2c54228474"
+}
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
cc @xclerc ?

xclerc:
There is indeed an issue with UDP, which is now part
of [async_extra](https://github.com/janestreet/async_extra) (in master, not in the latest release).

As we expect to do a stable release (v0.13.0) of our
packages soon, we might want to avoid this intermediary
step of having a standalone `async_udp` package.

Leonidas-from-XIV:
The problem is that Udp has been missing ever since v0.12.0 was released ([ticket](https://github.com/janestreet/async_extra/issues/13)), and thus preventing everybody who uses async_udp from upgrading OCaml 4.07, since base v0.11 (a transitive dependency of async v0.11) is constrained to < OCaml 4.08.

xclerc:
One problem, I think is that in v0.12, `async` depends
on `async_extra`. It is no longer the case in master (and
thus upcoming v0.13).

If there is a standalone release of `async_udp`, could we
at least make it obvious from its name that it is a kind
of "transition" package.

(Also, I would like to have a second opinion on this PR,
thus cc @trefis.)

trefis:
What I don't understand is why we don't just release `async_extra.v0.12.X` which includes async_udp. I think that was the plan at some point, but that didn't happen.
Do you know why?

xclerc:
The problem is that `async_udp` (which is part of `async_extra`)
depends on `async`, which `include`s `async_extra`.

My understanding is that for the point release to happen, we would
need to remove the aforementioned `include`, thus breaking
compatibility in a point release. (I would also have to double check
for possible further consequences of the removal.)

trefis:
Thanks for the explanation!

I had a discussion offline with @xclerc, here are the highlights:
- we should have communicated the problem on https://github.com/janestreet/async_extra/issues/13
- `async_udp` in our repository is part of the `async_extra` package: that is because we didn't want to create opam packages too aggressively
- that is preventing us from releasing a v0.12: because of the cyclic-dependency problem described above; the problem will not exist in v0.13
- this PR is creating an `async_udp` package
- if this PR is merged, but the next version (v0.13) of `async_udp` is part of the `async_extra` package that's likely to be confusing for user (the package will seem to have disappeared)
- to avoid this confusion, we have decided to just split out `async_udp` into its own package
- since we're splitting `async_udp` out, we might as well release a v0.12 ourselves

tl;dr: We're going to create an `async_udp` repository and package, and we're going to make an opam released on v0.12. All of this should happen over the next few days.

Therefore, my opinion is that this PR should be closed without merging. (Because the ownership story is clearer if we do the release ourselves).

@kit-ty-kate: does that all make sense to you?

avsm:
Thanks for the detailed explanation @trefis. Regarding:

> Therefore, my opinion is that this PR should be closed without merging. (Because the ownership story is clearer if we do the release ourselves).

My strong preference is for ownership to remain with JS for all of these packages if possible, and to not create an exception for this sole release of async_udp. There are quite a few automated scripts that generate these Jane Street packages to opam-repository, so let's avoid exceptions if possible.

Leonidas-from-XIV:
@trefis Thanks for the summary. This sounds like the optimal outcome to me; I also would prefer for JS to keep ownership of it. My rationale about this PR was to make it possible to have v0.12 and UDP, but if a JS-owned `async_udp` package exists I would indeed prefer that.

PR #15110, opened 2019-10-25
Pyml compatible with 4.09

xclerc:
Local testing seems to indicate that the latest `pyml`
is actually compatible with OCaml 4.09, at least with
`stdcompat >= 11`.

(This is done in preparation of the upcoming v0.13
release of Jane Street packages, as we will try to
have all our packages -and their dependencies-
compatible with 4.09.)

@thierry-martinez @LaurentMazare


kit-ty-kate:
Thanks a lot!

PR #15119, opened 2019-10-27
Add Coq 8.10.1

palmskog:
Via @Blaisorblade, here is a package definition for the recently-release Coq 8.10.1. This includes only Coq and not CoqIDE, since the latter usually causes CI trouble and is less prioritized. If/when this package gets merged, I will open a PR for CoqIDE 8.10.1.


kit-ty-kate:
Thanks a lot!

PR #15121, opened 2019-10-28
[new release] Orsetto 1.0.2.

jhwoodyatt:
- Compatibility with OCaml 4.09.

### DIFF
--- a/packages/orsetto/orsetto.1.0.1/opam
+++ b/packages/orsetto/orsetto.1.0.1/opam
@@ -7,7 +7,7 @@ bug-reports: "https://bitbucket.org/jhw/orsetto/issues"
 dev-repo: "hg+https://bitbucket.org/jhw/orsetto"
 tags: [ "org:conjury.org" ]
 depends: [
-    ("ocaml" { >= "4.07" } | ("seq" & "ocaml" { >= "4.06.1" & < "4.07" }))
+    ("ocaml" { >= "4.07" & < "4.09" } | ("seq" & "ocaml" { >= "4.06.1" & < "4.07" }))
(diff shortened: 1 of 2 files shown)


packages/orsetto/orsetto.1.0.1/opam
```diff
@@ -7,7 +7,7 @@ bug-reports: "https://bitbucket.org/jhw/orsetto/issues"
 dev-repo: "hg+https://bitbucket.org/jhw/orsetto"
 tags: [ "org:conjury.org" ]
 depends: [
-    ("ocaml" { >= "4.07" } | ("seq" & "ocaml" { >= "4.06.1" & < "4.07" }))
+    ("ocaml" { >= "4.07" & < "4.09" } | ("seq" & "ocaml" { >= "4.06.1" & < "4.07" }))
```
kit-ty-kate: According to this [log](http://check.ocamllabs.io/log/1571397380-35766877199e05ba1d4d2dbd00f7cd14694ad244/4.09/good/orsetto.1.0.1), `orsetto.1.0.1` did compile on OCaml 4.09 (only showing a few warnings)

kit-ty-kate:
But overall I think this PR is good. Thanks a lot!

packages/orsetto/orsetto.1.0.1/opam
jhwoodyatt: Grmf. Well, in any case, in the 1.0.x series, there should be no reason to pin 1.0.1. The only change in 1.0.2 is to silence a warning on OCaml 4.09 that gets treated as an error in development builds on my work environment.

PR #15122, opened 2019-10-28
Why3 1.2.1

silene:


kit-ty-kate:
The failure on Fedora is unrelated (coq needs `find` to compile but is not installed on fedora by default). This might need to be fixed on our side

silene:
The `opam` package for Coq 8.9.1 does not depend on `conf-findutils`, unfortunately. (The one for Coq 8.10.1 properly does.)

kit-ty-kate:
Yes it should be fixed in https://github.com/ocaml/opam-repository/pull/15123

silene:
The Travis Xcode failure is due to a timeout. And the only other failure is the one about Coq 8.9.1 not building on Fedora.

kit-ty-kate:
Thanks!

PR #15124, opened 2019-10-28
push chase 1.2

ramsdell:
The chase program is a model finder for first order logic with
equality.  It finds minimal models of a theory expressed in finitary
geometric form, where functions in models may be partial.  A formula
is in finitary geometric form if it is a sentence consisting of a
single implication, the antecedent is a conjunction of atomic
formulas, and the consequent is a disjunction.  Each disjunct is a
possibly existentially quantified conjunction of atomic formulas.  A
function is partial if it is defined only on a proper subset of its
domain.


kit-ty-kate:
This PR seems similar to https://github.com/ocaml/opam-repository/pull/15128, could you choose either of them and integrate my requested changes?

PR #15125, opened 2019-10-28
Mark packages as incompatible with dune 2

xclerc:
Follow-up to https://github.com/ocaml/opam-repository/pull/15077, marking some Jane Street packages as
incompatible with the upcoming 2.0 release of dune.

(note: two other Jane Street packages are in theory affected,
namely `ppx_cold` and `ppx_quickcheck`, but they are not yet
in opam.)


xclerc:
The errors reported by @bactrian are consistent with the ones
observed in https://github.com/ocaml/opam-repository/pull/15077: `zstd` is either not available, or too old.

kit-ty-kate:
Thanks a lot!

PR #15126, opened 2019-10-28
git.2.1.1: lower lower checkseum bound to support mirage 3

hannesm:
//cc @dinosaure as done in https://github.com/mirage/ocaml-git/commit/62d581cb5fbe4109cbb19924d37aa808c2ab883a


kit-ty-kate:
Thanks!

PR #15127, opened 2019-10-28
Add coqide 8.10.1

palmskog:
Now that Coq 8.10.1 has been merged, here is the (more-complicated-to-build) package for CoqIDE 8.10.1. As for Coq, the heavy lifting was done by @Blaisorblade. Also as for Coq, I added a description in the opam file based on the opam file in the Coq repo.


kit-ty-kate:
Thanks!

PR #15128, opened 2019-10-28
Fix MD5 hash in Chase 1.1 OPAM file

ramsdell:


packages/chase/chase.1.2/opam
```diff
@@ -0,0 +1,26 @@
...
+quantified conjunction of atomic formulas.  A function is partial if
+it is defined only on a proper subset of its domain.
+"""
+maintainer: "John D. Ramsdell <ramsdell@mitre.org>"
+authors: "John D. Ramsdell <ramsdell@mitre.org>"
+license: "BSD"
+homepage: "https://github.com/ramsdell/chase"
+dev-repo: "git+https://github.com/ramsdell/chase.git"
+bug-reports: "https://github.com/ramsdell/chase/issues"
+depends: ["ocaml" { >= "4.05" } "ocamlfind" "getopt" "dune"]
```
kit-ty-kate:
```suggestion
depends: [
  "ocaml" { >= "4.05" }
  "getopt"
  "dune" {>= "1.1"}
]
```
I don't think ocamlfind is required

kit-ty-kate:
Once accepted, could you return this change to the opam file upstream?

packages/chase/chase.1.2/opam
ramsdell: I change the depends field as you suggest.

PR #15129, opened 2019-10-28
Untag jbuilder as a build dependency

kit-ty-kate:
Same as #14266 but for jbuilder as it affects jbuilder packages as well when `jbuilder.transition` is installed.

There were also a few missing packages for dune as well:
* cc @gasche for `ppx_deriving_protobuf`
* cc @rjbou for the opam packages
* cc @dinosaure for `ke`


gasche:
Apologies for the slipup; this is fine as far as I'm concerned -- and I suspect it is fine in general.

rjbou:
ok for me also!

PR #15131, opened 2019-10-29
[new release] elpi (1.8.0)

gares:
CHANGES:

- Bugfix:
  - `shorten foo.{ bar }.` when `foo.bar` is a builtin used to be miscompiled.
  - `elpi-typechecker.elpi` now correclty stops printing warnings after it
    printed 10 (used to stop after he processed 10, that may not be the same
    thing, since some warnings are suppressed).

- Parser:
  - Interpret `-2` (with no space) as the negative `2` not as the constant `-2`.
    This way `X is 3 - 2` and `Y is 3 + -2` are both valid.

- FFI:
  - `OpaqueData` now requires a ternary comparison, not just equality.

- Stdlib:
  - new data type `cmp` for ternary comparison.
  - `std.set` and `std.map` now based on ternary comparison.

- Builtin:
  - `cmp_term` giving an order on ground terms.
  - `ground_term` to check if a term is ground.

### DIFF
--- a/packages/elpi/elpi.1.8.0/opam
+++ b/packages/elpi/elpi.1.8.0/opam
@@ -22,7 +22,7 @@ depends: [
   "re" {>= "1.7.2"}
   "ANSITerminal" {with-test}
   "cmdliner" {with-test}
-  "dune"
+  "dune" {>= "1.6"}
   "conf-time" {with-test}
 ]
 synopsis: "ELPI - Embeddable λProlog Interpreter"
--- a/packages/elpi/elpi.1.8.0/opam
+++ b/packages/elpi/elpi.1.8.0/opam
@@ -10,9 +10,10 @@ bug-reports: "https://github.com/LPCIC/elpi/issues"
 build: [
   ["dune" "subst"] {pinned}
   ["dune" "build" "-p" name "-j" jobs]
-  [make "tests"] {with-test & os != "macos"}
+  [make "tests" "DUNE_OPTS=-p %{name}%"] {with-test & os != "macos"}
 ]
-
+patches: [ "0001-Makefile-pass-DUNE_OPTS-to-dune.patch" ]
+extra-files: [ ["0001-Makefile-pass-DUNE_OPTS-to-dune.patch" "810836aa85bdb52d6a02947e93353ced"] ]
 depends: [
   "ocaml" {>= "4.04.0"}
   "camlp5"
--- a/packages/elpi/elpi.1.8.0/opam
+++ b/packages/elpi/elpi.1.8.0/opam
@@ -1,0 +1,84 @@
+opam-version: "2.0"
+maintainer: "Enrico Tassi <enrico.tassi@inria.fr>"
+authors: [ "Claudio Sacerdoti Coen" "Enrico Tassi" ]
+license: "LGPL 2.1+"
+homepage: "https://github.com/LPCIC/elpi"
+doc: "https://LPCIC.github.io/elpi/"
+dev-repo: "git+https://github.com/LPCIC/elpi.git"
+bug-reports: "https://github.com/LPCIC/elpi/issues"
+
+build: [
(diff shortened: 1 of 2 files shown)


gares:
Alpine fails the test target because `time` is not the real `time` utility, but busybox's one I guess, even if I've a `conf-time` dependency.

packages/elpi/elpi.1.8.0/opam
```diff
@@ -0,0 +1,84 @@
...
+depends: [
+  "ocaml" {>= "4.04.0"}
+  "camlp5"
+  "ppx_tools_versioned"
+  "ppx_deriving"
+  "ocaml-migrate-parsetree"
+  "re" {>= "1.7.2"}
+  "ANSITerminal" {with-test}
+  "cmdliner" {with-test}
+  "dune"
```
kit-ty-kate:
```suggestion
  "dune" {>= "1.6"}
```
According to your `dune-project` file this project requires at least dune 1.6

packages/elpi/elpi.1.8.0/opam
```diff
@@ -0,0 +1,84 @@
...
+license: "LGPL 2.1+"
+homepage: "https://github.com/LPCIC/elpi"
+doc: "https://LPCIC.github.io/elpi/"
+dev-repo: "git+https://github.com/LPCIC/elpi.git"
+bug-reports: "https://github.com/LPCIC/elpi/issues"
+
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+  [make "tests"] {with-test & os != "macos"}
```
kit-ty-kate: I'm not too happy with this but I'm willing to let it pass if it's too annoying to do. The makefile target is calling `dune build` without either `-p` or `--profile` and is likely to break in the future since dune enables some warnings as errors by default. Would it be possible for you to either move the rules in dune and just call `dune runtest -p name -j jobs` or have settable variable in your makefile to trigger dev vs. release profile?

kit-ty-kate:
> Alpine fails the test target because `time` is not the real `time` utility, but busybox's one I guess, even if I've a `conf-time` dependency.

This is fine, the error will be ignored.

packages/elpi/elpi.1.8.0/opam
gares: Here the makefile snippet
```make
tests:
        dune build $(INSTALL)/bin/elpi
        dune build $(BUILD)/tests/test.exe
        ulimit -s $(STACK); \
                $(BUILD)/tests/test.exe \
                --seed $$RANDOM \
                --timeout $(TIMEOUT) \
                $(TIME) \
                --sources=$(shell pwd)/tests/sources/ \
                --plot=$(shell pwd)/tests/plot \
                $(addprefix --name-match ,$(ONLY)) \
                $(addprefix --runner , $(RUNNERS))
```
Where do you want me to pass `-p`?

I'd like to keep the target in make, since I need `ONLY` and `RUNNERS` to be overrideable via the command line (and computed via `wildcard` if not passed), as well as `TIME` & co.

But I'm happy to pass make something like `DUNE_OPTS` or `DUNE_PACKAGE` and use it where it fits.

packages/elpi/elpi.1.8.0/opam
gares: done, thanks

packages/elpi/elpi.1.8.0/opam
kit-ty-kate: > Where do you want me to pass `-p`?

The two first lines only (the two calls to dune if I'm not mistaken)

gares: OK thanks for clarifying. I made the patch to Makefile as upstream, and added the specific commit to the opam package via `patches:`

kit-ty-kate:
Thanks!

PR #15136, opened 2019-10-29
Cryptokit version 1.14

xavierleroy:
This matches the latest release of Cryptokit, a library of crypto primitives.  The release fixes some nasty bugs, such as:
* Zlib thinking that its data structures are corrupted just because the OCaml GC moved them;
* some AMD Ryzen 3000 processors having a broken RDRAND instruction that produces endless streams of ones as "random" data.

More info at https://github.com/xavierleroy/cryptokit/releases/tag/release114

### DIFF
--- a/packages/cryptokit/cryptokit.1.14/opam
+++ b/packages/cryptokit/cryptokit.1.14/opam
@@ -14,14 +14,12 @@ depends: [
 ]
 build: make
 install: [make "install"]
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
The `remove` field is unnecessary since opam 2.0 and since the opam file is not in the upstream repository I took the liberty to modify the PR directly.

Thanks!

xavierleroy:
Thanks for the fix and the merge.

PR #15140, opened 2019-10-30

xclerc:
Just like we did for the v0.12 release, we are preparing
the v0.13 release by adding upper bounds to reverse
dependencies of Jane Street packages.


xclerc:
(notes about the failures)

Regarding @camelus's report:
- only warnings, presumably out of the scope of this pull request.

Regarding "ci/datakit/1 Build":

- `idd.0.1`: the source archive cannot be fetched;
- `nice_parser.1.0.0`: the checksum of the archive is invalid.

mmottl:
@xclerc This is probably the more appropriate PR for my question: why do all reverse dependencies have to receive upper bounds?  I thought such constraints were only added automatically if the build process actually failed.  I haven't tested all my packages yet, but the ones I have should probably work fine with `base.v0.13.0`.  What is the suggested approach for that?  Should I just manually remove those constraints again?  Or is there a better way for the future to avoid such a situation in the first place?

xclerc:
After re-reading [1], I understand people have reservations about
this release process. Since it was marked as "experimental", it
might be changed if not satisfactory.

@avsm and @kit-ty-kate will probably have a better appreciation
of the big picture (Because I am biased: it has made our life easier,
although part of the improvement was also due to changes to our
internal tooling).

[1] https://discuss.ocaml.org/t/experimenting-with-a-new-opam-repository-release-strategy-for-large-libraries/2918/23

mmottl:
I don't mind if some automated or manual process modifies my OPAM files to prevent build problems due to known incompatibilities with updated dependencies.  But if the added constraints are unnecessary, having to remove them manually adds churn.  I'd certainly be happy if this could be handled automatically and predictably as much as possible.

PR #15152, opened 2019-10-30
Update package conf-glfw3

SylvainBoilard:
CHANGES:

* Add explicit dependency on mesa-libGL-dev on distributions where the GLFW package does not depend on a package that provides GL/gl.h .


kit-ty-kate:
Thanks!

PR #15173, opened 2019-11-01
release mirage-entropy 0.5.0

hannesm:
uses mirage-runtime provided hooks instead of mirage-os-shim dependency


kit-ty-kate:
Thanks!

PR #15192, opened 2019-11-02
nocrypto: build mirage sublibrary if mirage-entropy is installed

hannesm:
the `mirage` subdirectory / sublibrary depends on mirage-entropy (in sync with `_tags`) -- so if mirage-entropy is installed, install the mirage sublibrary (earlier, this was done if mirage-xen or mirage-solo5 was installed).


hannesm:
imho good to merge, there are some old revdeps which have unrelated failures

kit-ty-kate:
Thanks!

PR #15197, opened 2019-11-03
[new release] functoria and functoria-runtime (3.0.2)

hannesm:
CHANGES:

* Remove custom opam version comparison code, instead collect min and max as
  sets and output them all (mirage/functoria#183, @hannesm fixes mirage/functoria#143)
  for `package ~min:"1.0" ~max:"2.0" "a" ; package ~min:"1.5" ~max:"2.0" "a"`,
  the output is now `"a" {>= "1.0" & >= "1.5" & < "2.0"}`, it used to be
  `"a" {>= "1.5" & < "2.0"}`.

  The advantage of avoiding to parse version numbers is that it can't be
  incompatible with how opam works (functoria's approach used to not support
  "1.0~beta", "1.0-5", "v1.0"; and it used to handle "1.0" and "1.0.0"
  differently than opam).


hannesm:
CI is green, this is the last package needed for mirage release, I'll merge myself.

PR #15199, opened 2019-11-03
asn-combinators 0.2.0-1: leverage the cstruct constraint by using bigarray-compat

hannesm:
see mirleft/ocaml-asn1-combinators#27 for the origin of the patch (here applied to the most recent release). asn1-combinators is the only package that requires cstruct downgrade in various mirageos unikernels / universes. it is preferable to be compatible with 5.0.0 which introduced capabilities.


kit-ty-kate:
Thanks!

PR #15204, opened 2019-11-04
Package ocaml-protoc.2.0.0

mransan:
### `ocaml-protoc.2.0.0`
Protobuf compiler for OCaml
Protobuf compiler for OCaml

---
* Homepage: https://github.com/mransan/ocaml-protoc
* Source repo: git+https://github.com/mransan/ocaml-protoc.git
* Bug tracker: https://github.com/mransan/ocaml-protoc/issues

---
:camel: Pull-request generated by opam-publish v2.0.0

PR #15205, opened 2019-11-04
Package ocaml-protoc.2.0.1

mransan:
### `ocaml-protoc.2.0.1`
Protobuf compiler for OCaml
Protobuf compiler for OCaml

---
* Homepage: https://github.com/mransan/ocaml-protoc
* Source repo: git+https://github.com/mransan/ocaml-protoc.git
* Bug tracker: https://github.com/mransan/ocaml-protoc/issues

---

### DIFF
--- a/packages/ocaml-protoc/ocaml-protoc.2.0.1/opam
+++ b/packages/ocaml-protoc/ocaml-protoc.2.0.1/opam
@@ -1,0 +1,28 @@
+opam-version: "2.0"
+synopsis: "Protobuf compiler for OCaml"
+description: "Protobuf compiler for OCaml"
+maintainer: "Maxime Ransan <maxime.ransan@gmail.com>"
+authors:[
+  "Maxime Ransan <maxime.ransan@gmail.com>"
+]
+homepage: "https://github.com/mransan/ocaml-protoc"
+bug-reports:"https://github.com/mransan/ocaml-protoc/issues"
+dev-repo:"git+https://github.com/mransan/ocaml-protoc.git"
+license: "MIT"
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+]
+depends: [
+  "ocaml" {>="4.02.1"}
+  "dune"  {build}
+  "stdlib-shims"
+  "ocamlbuild" {build}
+]
+url {
+  src: "https://github.com/mransan/ocaml-protoc/archive/2.0.1.tar.gz"
+  checksum: [
+    "md5=40c8142a8ad00228dadeb8162d25bb64"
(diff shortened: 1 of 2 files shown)


packages/ocaml-protoc/ocaml-protoc.2.0.1/opam
```diff
@@ -0,0 +1,28 @@
...
+bug-reports:"https://github.com/mransan/ocaml-protoc/issues"
+dev-repo:"git+https://github.com/mransan/ocaml-protoc.git"
+license: "MIT"
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+]
+depends: [
+  "ocaml" {>="4.02.1"}
+  "dune"  {build}
```
kit-ty-kate:
```suggestion
  "dune"  {>= "1.11"}
```
Dune should not be tagged as a build dependency (see #14266). Moreover your `dune-project` file says this package requires at least dune 1.11

packages/ocaml-protoc/ocaml-protoc.2.0.1/opam
```diff
@@ -0,0 +1,28 @@
...
+  "stdlib-shims"
+  "ocamlbuild" {build}
```
kit-ty-kate:
```suggestion
```
This packages doesn't seem to depend on ocamlbuild anymore

mransan: unfortunately, the following [package](https://opam.ocaml.org/packages/ocamlbuild-protoc/) has a transitive dependency on ocamlbuild through ocaml-protoc. Here is the  [issue](https://github.com/Khady/ocamlbuild-protoc/issues/1). The short term fix is to leave the ocamlbuild dependency in this package pending the resolution. 

(see my previous publishing attempt [here](https://github.com/ocaml/opam-repository/pull/15204)

packages/ocaml-protoc/ocaml-protoc.2.0.1/opam
mransan: Thanks for the suggestion and info. I'll accept it.

mransan:
@kit-ty-kate thanks for looking at this. I accept your suggestion for the dune specification but due to an unfortunate issue with a reverse dependency I cannot get rid of the ocamlbuild build dependency like we both would like to :(

kit-ty-kate:
cc @Khady for the change in `ocamlbuild-protoc`

kit-ty-kate:
> @kit-ty-kate thanks for looking at this. I accept your suggestion for the dune specification but due to an unfortunate issue with a reverse dependency I cannot get rid of the ocamlbuild build dependency like we both would like to :(

yes it can be done. Other packages can be modified/fixed if needed. Anyone can request a change

mransan:
@kit-ty-kate  this is awesome. thanks for making the fix.

kit-ty-kate:
Thanks

PR #15211, opened 2019-11-04
Release mirage net 3.0.1

hannesm:
this provides a smooth transition from mirage 3.6 to 3.7 -- at the same time, remove the mirage-net 3.0.0 which does _not_ provide a smooth transition.


hannesm:
instead of removing packages, this marks mirage-net 3.0.0 as `available: false`.

kit-ty-kate:
> instead of removing packages, this marks mirage-net 3.0.0 as `available: false`.

That seems to be a better solution, thanks a lot!

PR #15221, opened 2019-11-04
Release mirage channel 4.0.1

hannesm:
this provides a smooth transition from mirage 3.6 to 3.7 -- at the same time, remove the mirage-channel 4.0.0 which does not provide a smooth transition.


hannesm:
instead of removing packages, this marks mirage-channel 4.0.0 as `available: false`.

PR #15222, opened 2019-11-04
Release mirage clock 3.0.1

hannesm:
this provides a smooth transition from mirage 3.6 to 3.7 -- at the same time, remove the mirage-clock*3.0.0 which does not provide a smooth transition.


kit-ty-kate:
Thanks for all the PRs

PR #15206, opened 2019-11-04
Constrain some package in anticipation of dune 2.0

kit-ty-kate:
I went through a custom run of opam-health-check (before and after https://github.com/ocaml/opam-repository/pull/15158) and added a few easy constraints where needed.

cc @rgrinberg (I also added a constraint in `jbuilder.transition` to be more complete even though the conflicts is already in `dune.2.0.0`)

* cc @raphael-proust for `blake2`, `hacl` and `uecc`
* cc @yomimono for `bun`
* cc @UnixJunkie for `bytearray`, `get_line`, `orsvm_e1071` and `setcore`
* cc @zshipko for `ezsqlite`
* cc @struktured for `gemini`
* cc @gildor478 for `gettext-stub`
* cc @mseri for `md2mld`
* cc @hannesm for `mirage-clock-unix`
* cc @smolkaj for `open`
* cc @smondet for `ppx_deriving_cmdliner`
* cc @charlesetc for `relit-reason`
* cc @dinosaure for `rosetta`
* cc @LaurentMazare for `tensorflow`
* cc @Leonidas-from-XIV for `zmq` 

The reason for each breakage is stated in each individual commit messages.

PR #15158, opened 2019-10-31
[new release] dune-build-info, dune, dune-configurator, dune-action-plugin, dune-private-libs and dune-glob (2.0.0+draft1)

rgrinberg:
Embed build informations inside executable

- Project page: <a href="https://github.com/ocaml/dune">https://github.com/ocaml/dune</a>
- Documentation: <a href="https://dune.readthedocs.io/">https://dune.readthedocs.io/</a>

##### CHANGES:

- Introduce `alias` and `package` fields to the `rule` stanza. This is the
  preferred way of attaching rules to aliases. (ocaml/dune#2744, @rgrinberg)

- Add field `(optional)` for executable stanzas (ocaml/dune#2463, fixes ocaml/dune#2433, @bobot)

- Infer targets for rule stanzas expressed in long form (ocaml/dune#2494, fixes ocaml/dune#2469,
  @NathanReb)

- Indicate the progress of the initial file tree loading (ocaml/dune#2459, fixes ocaml/dune#2374,
  @bobot)

- Build `.cm[ox]` files for executables more eagerly. This speeds up builds at
  the cost of building unnecessary artifacts in some cases. Some of these extra
  artifacts can fail to built, so this is a breaking change. (ocaml/dune#2268, @rgrinberg)

- Do not put the `<package>.install` files in the source tree unless `-p` or
  `--promote-install-files` is passed on the command line (ocaml/dune#2329, @diml)

- Change `implicit_transive_deps` to be false. Implicit transitive deps now must
  be manually enabled (ocaml/dune#2306, @rgrinberg)

- Compilation units of user defined executables are now mangled by default. This
  is done to prevent the accidental collision with library dependencies of the
  executable. (ocaml/dune#2364, fixes ocaml/dune#2292, @rgrinberg)

- Enable `(explicit_js_mode)` by default. (ocaml/dune#1941, @nojb)

- Add an option to clear the console in-between builds with
 `--terminal-persistence=clear-on-rebuild`

- Stop symlinking object files to main directory for stanzas defined `jbuild`
  files (ocaml/dune#2440, @rgrinberg)

- Library names are now validated in a strict fashion. Previously, invalid names
  would be allowed for unwrapped libraries (ocaml/dune#2442, @rgrinberg)

- mli only modules must now be explicitly declared. This was previously a
  warning and is now an error. (ocaml/dune#2442, @rgrinberg)

- Modules filtered out from the module list via the Ordered Set Language must
[...]

PR #15229, opened 2019-11-06
Constrain some package in anticipation of dune 2.0

kit-ty-kate:
Follow up to https://github.com/ocaml/opam-repository/pull/15206

* cc @talex5 for `0install`
* cc @maranget for `herdtools7`
* cc @samoht  for `irmin-watcher`
* cc @raphael-proust for `lwt-pipeline`
* cc @c-cube for `minisat`
* cc @LaurentMazare for `npy`

The reason for each breakage is stated in each individual commit messages and further information is available in #15158

### DIFF
--- a/packages/0install/0install.2.14.1/opam
+++ b/packages/0install/0install.2.14.1/opam
@@ -16,7 +16,7 @@ depends: [
   "lwt_react"
   "ocurl" {>= "0.7.9"}
   "sha" {>= "1.9"}
-  "dune"
+  "dune" {< "2.0"}
   "cppo" {build}
 ]
 depopts: ["obus" "lablgtk" "lwt_glib"]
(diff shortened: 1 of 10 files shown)


c-cube:
what's  the migration plan for projects using symlinks? :s

kit-ty-kate:
> what's the migration plan for projects using symlinks? :s

cc @rgrinberg

rgrinberg:
There is no recommendation yet because I'm still investigating the issue. As far as I understand so far, we should have never allowed this symlink even back in 1.11.

c-cube:
thanks @rgrinberg , sorry for making your life harder.

maranget:
Ok, thanks,

As far as I understand the constraint added to the opam file will no longer be useful for herdtools7 future versions as there is now a dune-project file committed upstream.

Thanks,

--Luc

PR #15230, opened 2019-11-06
Install sdl2.ba.

smimram:
This adds support for the `sdl2.ba` module.


kit-ty-kate:
cc @fccm

mseri:
Even if accepted, I'd suggest to release into a separate version: `ocamlsdl2.0.03-1`

kit-ty-kate:
Agreed

fccm:
To compile with BAs isn't it different with different OCaml versions?
In this case is it possible to patch the Makefile based on different OCaml versions?

kit-ty-kate:
Closing this for lack of response. Please reopen a PR when ready. Thanks!

PR #15233, opened 2019-11-07
dokeysto and dokeysto_lz4: Clean up dependencies and constraints

kit-ty-kate:
Just a test for now, let's see if this works

cc @UnixJunkie

### DIFF
--- a/packages/dokeysto/dokeysto.3.0.0/opam
+++ b/packages/dokeysto/dokeysto.3.0.0/opam
@@ -16,8 +16,6 @@ depends: [
   "base-bytes"
   "base-unix"
   "extunix"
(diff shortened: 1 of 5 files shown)


kit-ty-kate:
For the record the last version of `dokeysto` (3.0.2) removed the dependency towards `dolog`, however `dokeysto` itself didn't depend on it and only `dokeysto_lz4` depends on it. `dokeysto_lz4.3.0.2` hasn't been released and therefore this change hasn't had any effect in practice

UnixJunkie:
I will try to release 3.0.2.
The software is already tagged but I didn't update the opam package.
Thanks for having taken care of the existing dokeysto-related opam packages.

PR #15234, opened 2019-11-07
Mark dokeysto_lz4.3.0.1 as incompatible with dune 2.0

kit-ty-kate:
cc @UnixJunkie (see https://github.com/ocaml/opam-repository/pull/15158#issuecomment-551103739 for more details)

### DIFF
--- a/packages/dokeysto_lz4/dokeysto_lz4.3.0.1/opam
+++ b/packages/dokeysto_lz4/dokeysto_lz4.3.0.1/opam
@@ -12,7 +12,7 @@ build: [
 ]
 depends: [
   "ocaml"
-  "dune" {>= "1.0"}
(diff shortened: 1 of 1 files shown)


UnixJunkie:
Should I do something to make it compatible with dune 2.0?
Dune-2.0 is not in opam yet.
Ideally, I would like all my packages to install whatever the dune version.
A low version number, lower bound, on dune is acceptable, but I would
prefer not to have an upper bound.

kit-ty-kate:
As stated in https://github.com/ocaml/opam-repository/pull/15158#issuecomment-551081819 just adding a `dune-project` file fixes the issue so `3.0.2` will fix the issue already

UnixJunkie:
Cool, thanks for the details.

PR #15239, opened 2019-11-07
Switch mlmpfr package to dune

thvnx:
Hi, I switched the package to dune.

CI should fail if mpfr C library <= 4.0.0 is installed on test machine.

### DIFF
--- a/packages/mlmpfr/mlmpfr.4.0.2+dune/files/mlmpfr_compatibility_test.c
+++ b/packages/mlmpfr/mlmpfr.4.0.2+dune/files/mlmpfr_compatibility_test.c
@@ -1,0 +1,24 @@
+#include <string.h>
+#include <stdio.h>
+#include <mpfr.h>
+
+int main(int argc, char **argv)
+{
+  const char *version = mpfr_get_version();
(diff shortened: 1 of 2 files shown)


thvnx:
See also: https://github.com/thvnx/mlmpfr

kit-ty-kate:
I would rather get a new release (`mlmpfr.4.0.3` for instance) than changing the latest one.

thvnx:
> I would rather get a new release (`mlmpfr.4.0.3` for instance) than changing the latest one.

In fact, I stick to the original C library version (not very clever I agree...), it allows to easily see which version of mlmpfr is compatible with the mpfr C library.

kit-ty-kate:
Ok but is there no way of doing `4.0.2-1` for example?

thvnx:
Is something like `4.0.2+dune` possible? Will it be catched as the latest version?

kit-ty-kate:
+dune sounds better and yes it will

thvnx:
> +dune sounds better and yes it will

Done!

kit-ty-kate:
Thanks!

PR #15258, opened 2019-11-11

xclerc:
Follow-up to https://github.com/ocaml/opam-repository/pull/15140.


kit-ty-kate:
Thanks!

PR #15265, opened 2019-11-12
+ mysql 1.2.4

ygrek:
build is expected to fail due to depext (see explanation in post-messages)

### DIFF
--- a/packages/mysql/mysql.1.2.4/opam
+++ b/packages/mysql/mysql.1.2.4/opam
@@ -20,16 +20,7 @@ install: [make "install"]
 depends: [
   "ocaml"
   "ocamlfind" {build}
-]
-post-messages: [
-  "This package requires mysql client library installed. Due to abundance of mysql forks with different names and
-  opam not dealing very well with depext alternatives currently - they are not described as dependencies of this package.
-
-  Please ensure that mysql client library is installed and mysql_config or mariadb_config binary is available in PATH.
-  Two most popular ways to achieve that are described in conf-mysql and conf-mariadb packages.
-
-  See also discussion in https://github.com/ocaml/opam-repository/pull/10312
-" {failure}
+  ("conf-mariadb" | "conf-mysql")
 ]
 synopsis:
   "Bindings to C client library for interacting with Mysql/MariaDB/Percona databases"
--- a/packages/mysql/mysql.1.2.4/opam
+++ b/packages/mysql/mysql.1.2.4/opam
@@ -1,0 +1,45 @@
+opam-version: "2.0"
+maintainer: "ygrek@autistici.org"
+authors: [
+  "Christian Lindig <lindig@eecs.harvard.edu>"
+  "Shawn Wagner <shawnw@speakeasy.org>"
+  "ygrek <ygrek@autistici.org>"
+]
+homepage: "https://ygrek.org.ua/p/ocaml-mysql/"
+doc: "https://ygrek.org.ua/p/ocaml-mysql/api/index.html"
+dev-repo: "git://github.com/ygrek/ocaml-mysql.git"
+bug-reports: "https://github.com/ygrek/ocaml-mysql/issues"
+license: "LGPL-2.1"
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
I've been a bit puzzled for a few months about what to do with this package. I could see it fails in the CIs and I was unsure what to do about it.

I'd really like to have this issue fixed if possible as it might also be repulsive and confusing for users as well. I would have one questions: how many forks does this package support? If it's only a handful I think it would be good to split this package into a small number of sub-package (one for every fork supported) but pointing to the same code. It would be a bit less cryptic for users (if it is actually doable) I think

hcarty:
One positive aspect of the current name is that it is clear what the package provides - a library for communicating with mysql (compatible) servers.  It would be nice to keep that clarity if possible.

Would sub-packages help in practice?  The issue, as I understand it, is that there are four or more potential packages to rely on to provide the necessary C client library which are all external to opam: mysql, mysql c connector, mariadb, mariadb c connector.  Each OS may support and provide packages for multiple of those.  _And_ the specific packages available vary by OS/distribution version.  Do we have a clean way to represent that kind of open-ended dependency in opam?  Does it make things clearer if a user has to pick from `mysql-mysql`, `mysql-mariadb` and maybe more versus installing `mysql` directly?

I have found it useful that this single package works across different versions of the mysql/mariadb client libraries.  It makes it much easier to have a codebase which compiles and works across macos and various flavors of Linux which each have their own combination of C client libraries available.

Subpackages wouldn't break that if they all point to the same code.  I'm not sure which would be more confusing/concerning though - seeing that `mysql-mysql` is being installed by opam (and it works!) when some flavor of mariadb is installed locally.  Or always installing `mysql` with opam, regardless of the local client library being bound to, and having the opam CI never properly turn green.

hcarty:
We also may be able to improve the CI situation once opam 2.1.x lands as, if I recall correctly, it integrates depext with some improvements which should allow us to specify different system package requirements based on distribution version.

ygrek:
ftr the root of the problem originally was that opam didn't handle alternatives well (installed all of them, instead of being satisfied by one). Let me try pushing with conf-* alternatives dep, seems it works better now.

PS some rambling before trying to actually fix issue: 
I dont think splitting this will make it easier for users. Conceptually what this library does is binds to mysql api which can be implemented by many libraries, for today those are libmysqlclient from Oracle Mysql, old libmysqlclient from original mysql (some users still have it), mysql compatible layer provided by mariadb connector/c which was at different times shipped as libmysqlclient then libmariadbclient and now libmariadb in debian. This already causes enough troubles when upgrading, changing name of package to match that will not help.
Maybe we can solve this by having some opam package overlay for CI - ie ability to specify depext for CI. It could be x- field or a separate (partial) opam file or just a shell script to run before installing this package? The latter is easiest and hackiest from tooling effort I guess, would be my preferred method :)

ygrek:
travis now succeeds.
datakit shows that conf-mysql is not installable on new OS which is expected but it works for travis image which apparently has older distro?

kit-ty-kate:
Sorry for taking so much time. I think this is an acceptable partial fix for the moment. It doesn't work on debian stable but it works on debian unstable that I use on [check.ocamllabs.io](http://check.ocamllabs.io). Sorry again, thanks!

hcarty:
Adding the conf-* requirement has broken builds on macos + homebrew.  It should work with conf-mariadb but the pkg-config check should be `pkg-config libmariadb` rather than `pkg-config mariadb` at https://github.com/ocaml/opam-repository/blob/master/packages/conf-mariadb/conf-mariadb.1/opam#L8.

I'd like to unbreak this if at all possible.  Is it possible to select the build command with the same granularity as the depext fields are specified?

kit-ty-kate:
@hcarty yes, this should be fixable with `["pkg-config" "libmariadb" {os = "macos"} "mariadb" {os != "macos"}]` or something like that. I'm creating a PR right now.

hcarty:
Thank you @kit-ty-kate!

PR #15268, opened 2019-11-13
New version of ACGtk

Nyavlys:
This new version removes dependency to ocf (which currently does not compile with 4.08). Also introduces using cmdliner to handle command line arguments.

### DIFF
--- a/packages/acgtk/acgtk.1.5.1/opam
+++ b/packages/acgtk/acgtk.1.5.1/opam
@@ -12,7 +12,7 @@ build: [
 install: ["dune" "install"]
 
 depends: [
-  "ocaml" { >= "4.03.0"}
+  "ocaml" { >= "4.05.0"}
   "dune" {build}
   "menhir"
   "ANSITerminal"
@@ -38,5 +38,5 @@ description: "This toolkit provides a compiler and an interpreter for Abstract C
 
 url {
      src: "http://acg.loria.fr/software/acg-1.5.1-20191113.tar.gz"
-     checksum: "2ca67ad457bdfd036e942caf6d85164c"
+     checksum: "6d770c390d2b7870c4261dc2d1772696"
 }
--- a/packages/acgtk/acgtk.1.5.1/opam
+++ b/packages/acgtk/acgtk.1.5.1/opam
@@ -13,7 +13,7 @@ install: ["dune" "install"]
 
 depends: [
   "ocaml" { >= "4.05.0"}
-  "dune" {build}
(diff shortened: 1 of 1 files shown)


packages/acgtk/acgtk.1.5.1/opam
```diff
@@ -0,0 +1,42 @@
...
+authors: ["Sylvain Pogodalla"]
+bug-reports: "sylvain.pogodalla@inria.fr"
+
+synopsis: "Abstract Categorial Grammar development toolkit"
+
+description: "This toolkit provides a compiler and an interpreter for Abstract Categorial Grammars (ACGs). Grammars can be compiled and then used by the interpreter to parse (if the grammar is at most second-order) or to generate terms. See http://acg.loria.fr for more details and bibliographic references."
+
+url {
+     src: "http://acg.loria.fr/software/acg-1.5.1-20191113.tar.gz"
+     checksum: "f813f0ba2d4403667da4ac302af52990"
```
kit-ty-kate:
```suggestion
     checksum: "md5=f813f0ba2d4403667da4ac302af52990"
```

packages/acgtk/acgtk.1.5.1/opam
```diff
@@ -0,0 +1,42 @@
...
+
+build: [
+  ["dune" "subst"] {pinned}
+# remove the -p to also build the local libraries: conflict with the
+# fact that some libraries are also part of the acgtkLib package
+#  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "build" "-j" jobs]
+]
+
+install: ["dune" "install"]
```
kit-ty-kate:
```suggestion
```
Dune already support installs using the `<package>.install` file that opam uses.

kit-ty-kate: oh right nevermind, I should have seen the comment above

packages/acgtk/acgtk.1.5.1/opam
```diff
@@ -0,0 +1,42 @@
...
+
+depends: [
+  "ocaml" { >= "4.05.0"}
+  "dune" {build}
```
kit-ty-kate:
```suggestion
  "dune" {>= "1.1"}
```
Dune should not be tagged as a build dependency (see #14266) and your `dune-project` file says this package requires at least dune 1.1

Nyavlys:
There is a problem with the checksum. I can't see how to easily solve it, so I'll close this PR and start another one where I include the proposed changes.

Sorry.

PR #15276, opened 2019-11-14
Solo5 0.6.4 release

mato:


mato:
The CentOS failure is expected (GCC is too old).

kit-ty-kate:
Thanks!

PR #15282, opened 2019-11-14
Fix dune 2.0's revdeps

kit-ty-kate:
Broken packages with dune 2.0 discovered in https://github.com/ocaml/opam-repository/pull/15261

cc @rgrinberg 

* cc @UnixJunkie for `minicli` and `lz4_chans`
* cc @komamitsu for `fluent-logger`

Reasons for each fixes are noted in the commit message. More details here: https://github.com/ocaml/opam-repository/pull/15261#issuecomment-553979070

### DIFF
--- a/packages/fluent-logger/fluent-logger.1.1.0/opam
+++ b/packages/fluent-logger/fluent-logger.1.1.0/opam
@@ -13,7 +13,7 @@ build: [
 ]
 
 depends: [
-    "dune"
+    "dune" {< "2.0"}
     "msgpack"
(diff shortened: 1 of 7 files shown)


UnixJunkie:
Should I add a dune-project file in those packages so that they are future-proof regarding dune?

rgrinberg:
Yes, that is what we recommend.

UnixJunkie:
Ok, I'll do some bugfix releases for those packages then.

PR #15285, opened 2019-11-14
mariadb.1.1.4: ctypes-foreign dependency is not needed

ygrek:
https://github.com/andrenth/ocaml-mariadb/issues/26


kit-ty-kate:
Thanks

Issue #15281, opened 2019-11-14
bad checksum for ocaml-migrate-parsetree


aryx:
I use opam in a docker container to install some packages but I recently got this error in my CI:

[topkg.1.0.1] downloaded from http://erratique.ch/software/topkg/releases/topkg-1.0.1.tbz
[ERROR] The sources of the following couldn't be obtained, aborting:
          - ocaml-migrate-parsetree.1.4.0: Bad checksum

Anyway I can fix that?

Here is the line in the Dockerfile:
Step 6/18 : RUN git pull && opam update && opam switch 4.07 &&   opam install dune yaml &&   opam install pfff pfff.0.39.3

I run this in the docker image FROM ocaml/opam2:alpine

aryx:
Here is the relevant part of the Dockerfile:
FROM ocaml/opam2:alpine as build
USER root
RUN apk add --no-cache perl m4
USER opam
WORKDIR /home/opam/opam-repository
RUN git pull && opam update && opam switch 4.07 && \
  opam install dune yaml && \
  opam install pfff pfff.0.39.4
COPY --chown=opam src /build
WORKDIR /build/
RUN eval $(opam env); make

ygrek:
```
$ opam info ocaml-migrate-parsetree
...
url.src:      "https://github.com/ocaml-ppx/ocaml-migrate-parsetree/releases/download/v1.4.0/ocaml-migrate-parsetree-v1.4.0.tbz"
url.checksum: "sha256=231fbdc205187b3ee266b535d9cfe44b599067b2f6e97883c782ea7bb577d3b8"
              "sha512=61ee91d2d146cc2d2ff2d5dc4ef5dea4dc4d3c8dbd8b4c9586d64b6ad7302327ab35547aa0a5b0103c3f07b66b13d416a1bee6d4d117293cd3cabe44113ec6d4"
...
$ curl -L https://github.com/ocaml-ppx/ocaml-migrate-parsetree/releases/download/v1.4.0/ocaml-migrate-parsetree-v1.4.0.tbz | sha256sum 
dcd77cd7090ce181a87df08a910ab2935da466f37db8e674e5a966d49743de64 -
```

ygrek:
looks like upstream tarball was updated. cc @diml 
https://github.com/ocaml-ppx/ocaml-migrate-parsetree/releases/tag/v1.4.0
> v1.4.0
> @diml diml released this 6 hours ago

ygrek:
[...]

PR #15286, opened 2019-11-15
letsencrypt: sync version field with actual version

avsm:


mseri:
Is the version there needed for some compile-time reason (like for lwt) or we can actually drop it?

avsm:
no reason, but it's consistent with the other versions, so just fixed it rather than remove it.

mseri:
Unfortunately it cannot be tested until https://github.com/ocaml/opam-repository/issues/15281 is fixed (this is going to escalate fast, it affects a lot of packages)

PR #15288, opened 2019-11-15
added packages/lz4_chans/lz4_chans.3.0.1/opam

UnixJunkie:
added dune-project file in the source tree

### DIFF
--- a/packages/lz4_chans/lz4_chans.3.0.1/opam
+++ b/packages/lz4_chans/lz4_chans.3.0.1/opam
@@ -1,0 +1,29 @@
+opam-version: "2.0"
+authors: "Francois Berenger"
+maintainer: "unixjunkie@sdf.org"
+homepage: "https://github.com/UnixJunkie/lz4-chans"
+bug-reports: "https://github.com/UnixJunkie/lz4-chans/issues"
+dev-repo: "git+https://github.com/UnixJunkie/lz4-chans.git"
+license: "BSD-3-Clause"
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+# the tests don't pass on OCaml 4.03.0, because of a marshalling behavior
+# that was working before 4.03.0 and put back in place after 4.03.0
+  ["dune" "exec" "-p" name "src/test.exe"] {with-test & ocaml:version != "4.03.0"}
+]
+depends: [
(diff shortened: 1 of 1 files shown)


packages/lz4_chans/lz4_chans.3.0.1/opam
```diff
@@ -0,0 +1,29 @@
...
+# the tests don't pass on OCaml 4.03.0, because of a marshalling behavior
+# that was working before 4.03.0 and put back in place after 4.03.0
+  ["dune" "exec" "-p" name "src/test.exe"] {with-test & ocaml:version != "4.03.0"}
+]
+depends: [
+  "ocaml"
+  "base-unix"
+  "conf-lz4"
+  "dolog" {>= "4.0.0"}
+  "dune"
```
kit-ty-kate:
```suggestion
  "dune" {>= "1.6"}
```
According to your `dune-project` file, this package requires at least dune 1.6

kit-ty-kate:
Also I don't know if it was on purpose but your master branch doesn't contain the commits from `3.0.1`. You might want to do `git push origin master` in your repository.

packages/lz4_chans/lz4_chans.3.0.1/opam
UnixJunkie: Let's just hope that dune puts the lowest possible version number in there...
Because, the truth is the lowest dune version that can compile this project is OK.
Not just "the dune version that I was using when I generated this file".

UnixJunkie:
The git repository is supposed to be up to date.
Maybe github was lagging when you checked?

kit-ty-kate:
> The git repository is supposed to be up to date.
> Maybe github was lagging when you checked?

mmh maybe. Sorry about that.

packages/lz4_chans/lz4_chans.3.0.1/opam
kit-ty-kate: > Not just "the dune version that I was using when I generated this file".

Dune does that when you just do `dune build` or any dune command that generates it, except for `dune upgrade` which just uses dune 1.0 because dune 1.0 is the closest to jbuilder.

kit-ty-kate:
Thanks!

PR #15316, opened 2019-11-17
Add new package tsort.

dmbaturin:
tsort provides a topological sort function with a user-friendly interface that aims to be as easy to use as the classic UNIX tsort.

It also has distinct errors for cycles and non-existent dependencies.

Examples:
```
# Tsort.sort [("foundation", []); ("walls", ["foundation"]); ("roof", ["walls"])];;
 - : string Tsort.sort_result = Tsort.Sorted ["foundation"; "walls"; "roof"]

# Tsort.sort [("foundation", ["roof"]); ("walls", ["foundation"]); ("roof", ["walls"])];;
- : string Tsort.sort_result = Tsort.ErrorCycle ["roof"; "foundation"; "walls"]

# Tsort.sort [("foundation", ["building permit"]); ("walls", ["foundation"]); ("roof", ["walls"])];;
- : string Tsort.sort_result = Tsort.ErrorNonexistent ["building permit"]
```

### DIFF
--- a/packages/tsort/tsort.1.0.0/opam
+++ b/packages/tsort/tsort.1.0.0/opam
@@ -20,13 +20,13 @@ build: [
 depends: [
   "ocaml" {>= "4.03.0"}
   "containers"
-  "dune"
+  "dune" {>= "1.9"}
 ]
 
 url {
   src: "https://github.com/dmbaturin/ocaml-tsort/archive/1.0.0.tar.gz"
   checksum: [
-    "md5=63174cfcc61e7d765486cbec38da6c1b"
-    "sha512=02da7bbeb2b00718d5bec7f150be81a773f38ce9b898a20fcc9775890df5d6daeb1099d97b08644d966ac0a2ebfe1742eb559854a600150e63796a332d360773"
+    "md5=d7049eadc01cd641ff9e23ec08cc43e1"
+    "sha512=b4153c9f266b2b8f468e04e9c3823a9c2a14a92d463825a0b8f1d5777091c425b9327d4f5518fb72ba0b04fc9b54b03fc3a49eeb85e252e9c0cddb06adb7ffb6"
   ]
 }
--- a/packages/tsort/tsort.1.0.0/opam
+++ b/packages/tsort/tsort.1.0.0/opam
@@ -20,7 +20,7 @@ build: [
 depends: [
   "ocaml" {>= "4.03.0"}
   "containers"
-  "dune"     {>= "1.5.1"}
+  "dune"
 ]
 
 url {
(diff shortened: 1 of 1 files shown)


packages/tsort/tsort.1.0.0/opam
```diff
@@ -0,0 +1,32 @@
...
+homepage: "https://github.com/dmbaturin/ocaml-tsort"
+bug-reports: "https://github.com/dmbaturin/ocaml-tsort/issues"
+dev-repo: "git+https://github.com/dmbaturin/ocaml-tsort.git"
+license: "MIT"
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+]
+depends: [
+  "ocaml" {>= "4.03.0"}
+  "containers"
```
kit-ty-kate: I would be very surprised if any versions of containers works here. If you could check what is the minimal version required for this dependency that'd be perfect. Otherwise I understand that might be a bit annoying to do and I'm willing to pass on this. It's not mandatory for this PR to be merged.

packages/tsort/tsort.1.0.0/opam
```diff
@@ -0,0 +1,32 @@
...
+  "dune"     {>= "1.5.1"}
```
kit-ty-kate: I'm a bit puzzled by this constraint. Where does it come from? I don't see any `dune-project` file at the root of your repository. Could you add it? It makes our (opam-repository maintainers) life infinity easier as future version of dune might break this package.

kit-ty-kate:
Good to merge once those comments have been taken care of.

packages/tsort/tsort.1.0.0/opam
dmbaturin: The dependencies of containers 2.7 say 4.02. I've just checked it it works with 4.03 and it actually does.
Since tsort now uses custom a tri-state type for its results to indicate cycles and missing dependencies explicitly, it may work with even earlier versions, but I doubt anyone will want to try 4.02.

packages/tsort/tsort.1.0.0/opam
dmbaturin: I've just pushed a dune-project to dmbaturin/ocaml-tsort. Should I re-create a release with it in the tarball?

kit-ty-kate: If you can that'd be perfect (with the current `dune-project` file you have you will require to change the constraint to `{>= "1.9"}` though).

dmbaturin: Ok, I've re-created the release, and updated the opam file.
By future versions, you mean post-2.0? What's the best current practice in that regard now?

kit-ty-kate:
Thanks a lot!

PR #14935, opened 2019-09-27
added packages/molenc/molenc.5.0.0

UnixJunkie:
There is a Butina clustering implementation now.
Also, switched to dolog >= 4.0.0

PR #15318, opened 2019-11-18
added packages/conf-rdkit/conf-rdkit.1.0.0

UnixJunkie:


UnixJunkie:
Is there a way to make the following work?

On a mac, the following commands should be run so that the rdkit depext is installed correctly:
```
brew tap rdkit/rdkit
brew install rdkit --with-python3 --with-inchi
```
I.e. not only does brew need to install from a special tap, but also we must
pass several options to the brew install command for this specific package.

Thanks,
F.

UnixJunkie:
I confirm this package works on the mac I have access to.

UnixJunkie:
The CI tests fail on the Linux box because it does not have the required packages:
```
# The following system packages are needed:
librdkit-dev
python3
python3-rdkit
The following command needs to be run through "sudo":
    apt-get install -qq -yy librdkit-dev python3 python3-rdkit
E: Unable to locate package python3-rdkit
```
I think it exists in recent Debian:
https://packages.debian.org/sid/python3-rdkit

The tests fail on the mac because of this: 
```
### output ###
# Error: /usr/local/Cellar is not writable. You should change the
# ownership and permissions of /usr/local/Cellar back to your
# user account:#   sudo chown -R $(whoami) /usr/local/Cellar
```

If I can do anything to push this further, tell me.

Regards,
F.

kit-ty-kate:
I'm not sure how much of an improvement this new version is compared to the old one. At least `conf-rdkit.0.1` succeeds on debian & ubuntu, which makes packages using it testable. For instance see: https://github.com/ocaml/opam-repository/pull/14935. What advantages do we get from getting this new version?

UnixJunkie:
The problem is that many distributions ship a too old version of rdkit.
Recent versions don't support Python2 anymore.
So, this is a conf package for the current version of rdkit.
The package is simpler than the previous one, because now it is just a python script.

Maybe, this could be merged only once the CI distro images ship the python3-rdkit package.
Then all the CI tests should pass.

Also, this is annoying but the tests should pass on the mac.
I am one of the maintainers of the brew formula for rdkit, and it uses the latest
rdkit release.

kit-ty-kate:
Also on Debian Unstable the package exists but the test fail
```
#=== ERROR while compiling conf-rdkit.1.0.0 ===================================#
# context              2.0.5 | linux/x86_64 | ocaml-base-compiler.4.09.0 | pinned
# path                 ~/.opam/4.09/.opam-switch/build/conf-rdkit.1.0.0
# command              /usr/bin/python3 test.py
# exit-code            1
# env-file             ~/.opam/log/conf-rdkit-23-6d07ae.env
# output-file          ~/.opam/log/conf-rdkit-23-6d07ae.out
### output ###
#   File "test.py", line 4, in <module>
#     from rdkit import Chem
#   File "/usr/lib/python3/dist-packages/rdkit/Chem/__init__.py", line 23, in <module>
#     from rdkit.Chem.rdmolfiles import *
# ImportError: libmaeparser.so.1.2: cannot open shared object file: No such file or directory
```

UnixJunkie:
Ok, I'll check this one in a VM.

UnixJunkie:
I could reproduce the problem.
I sent an e-mail to the debichem team.
Probably this is a bug in the python3-rdkit package or one of its dependencies.

XVilka:
@UnixJunkie did they reply anything by the way?

UnixJunkie:
Yes, they acknowledged it.
So, some day, the package will get fixed.
Probably when they update the version they ship.
I cannot find back the debian bug id though.

mseri:
What is the state of this? Should we close it or should we consider merging it?

UnixJunkie:
Maybe on an up to date debian unstable, someone can test.
I confirm that with a recent rdkit installed (i.e. from sources, not via apt), the python script works.

UnixJunkie:
Can the tests be relaunched for this PR?
It should work on latest Debian / Ubuntu.
It should also work on some Macs.

kit-ty-kate:
Superseded by https://github.com/ocaml/opam-repository/pull/17909

PR #15322, opened 2019-11-18
Add a constraint on git about dune

dinosaure:
/cc @kit-ty-kate


kit-ty-kate:
Could you do it in the other packages (git-unix, git-mirage, git-http) as well? It's not "necessary" per say but it's a lot cleaner to see all the constraints for direct dependencies in one single file.

dinosaure:
Done!

kit-ty-kate:
Thanks a lot!

PR #15324, opened 2019-11-18
Jane Street packages: v0.13.0

xclerc:


kit-ty-kate:
Could you remove all the `descr` file from the PR?

kit-ty-kate:
And `url` files too. Mh I think something went wrong. I'm cancelling the CI

xclerc:
Sorry, there is apparently a problem with the export script.

xclerc:
Should be slightly better now; sorry for the false start.

xclerc:
The Travis builds failed on timeout.

The other failures seem to fall into two categories (I have to admit
I did not inspect every single log, though):

- `pull access denied for sha256, repository does not exist or may require 'docker login': denied: requested access to the resource is denied`;
- `docker: Error: No such image: sha256:...`.

kit-ty-kate:
It must have been running at the same time as the daily docker purge. I've relaunched the failing jobs.

xclerc:
It feels like the CI has not made any progress for several hours
(i.e. the message after  _Pending_ has not changed, and the
page on `ci.ocaml.org` is unchanged).

kit-ty-kate:
Seems alright. I'm launching Camelus manually just as a quick check but it seems good.

cc @thomasblanc could you check why so many PRs are left out by Camelus? More than half of them are not checked and it has been the case for a while now.

kit-ty-kate:
Mmh, the branch is too old to have a decent result. Oh well. Merging anyway. Could you check `zstandard` though. I don't think `ctypes.0.14.0+dynamicfunptrtype` exists in opam-repostiory.

kit-ty-kate:
Thanks

xclerc:
Thanks; there is indeed an issue with `zstandard`.

The latest `async_ssl` is based on a patch against `ctypes`,
and this ended up showing in `zstandard`. (We decided to
not release `async_ssl.v0.13.0` for the moment.)

A fix is in https://github.com/ocaml/opam-repository/pull/15336.

thomasblanc:
@kit-ty-kate we had a certificate issue that is now fixed. Camelus should be working. Please let me know in case of a problem because neither Louis or I have much time to babysit him atm.

PR #15328, opened 2019-11-18
[new release] printbox (v0.3)

c-cube:
Allows to print nested boxes, lists, arrays, tables in several formats

- Project page: <a href="https://github.com/c-cube/printbox/">https://github.com/c-cube/printbox/</a>

## 0.3

- improve code readability in text rendering
- add `align` and `center`
- add basic styling for text (ansi codes/html styles)
- add `printbox_unicode` for setting up proper unicode printing
- add `grid_l`, `grid_text_l`, and `record` helpers

- use a more accurate length estimate for unicode, add test
- remove mdx as a test dep
- fix rendering bugs related to align right, and padding

## 0.2

- make the box type opaque, with a view function
- require OCaml 4.03

- add `PrintBox_text.pp`
- expose a few new functions to build boxes
- change `Text` type, work on string slices when rendering

- automatic testing using dune and mdx
- migrate to dune and opam 2

## 0.1

initial release

### DIFF
--- a/packages/printbox/printbox.v0.3/opam
+++ b/packages/printbox/printbox.v0.3/opam
@@ -1,0 +1,35 @@
+opam-version: "2.0"
+version: "0.3"
+authors: ["Simon Cruanes" "Guillaume Bury"]
+maintainer: "simon.cruanes.2007@m4x.org"
+synopsis: "Allows to print nested boxes, lists, arrays, tables in several formats"
+build: [
+  ["dune" "build" "@install" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+  ["dune" "build" "@doc" "-p" name "-j" jobs] {with-doc}
+]
+depends: [
+  "dune" {build}
+  "base-bytes"
+  "odoc" {with-doc}
+  "ocaml" { >= "4.03" }
+  "uutf" {with-test}
+  "uucp" {with-test}
+]
+depopts: [
+  "tyxml"
+  "uutf"
+  "uucp"
+]
+tags: [ "print" "box" "table" "tree" ]
+homepage: "https://github.com/c-cube/printbox/"
+dev-repo: "git+https://github.com/c-cube/printbox.git"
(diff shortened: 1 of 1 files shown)


packages/printbox/printbox.v0.3/opam
```diff
@@ -0,0 +1,35 @@
+opam-version: "2.0"
+version: "0.3"
```
kit-ty-kate:
```suggestion
```
This field is unnecessary and doesn't match the package version. Could you move the package from `printbox.v0.3` to `printbox.0.3`?

packages/printbox/printbox.v0.3/opam
```diff
@@ -0,0 +1,35 @@
...
+authors: ["Simon Cruanes" "Guillaume Bury"]
+maintainer: "simon.cruanes.2007@m4x.org"
+synopsis: "Allows to print nested boxes, lists, arrays, tables in several formats"
+build: [
+  ["dune" "build" "@install" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+  ["dune" "build" "@doc" "-p" name "-j" jobs] {with-doc}
+]
+depends: [
+  "dune" {build}
```
kit-ty-kate:
```suggestion
  "dune" {>= "1.1"}
```
Dune should not be tagged as a build dependency (see #14266) and your `dune-project` file states that this package requires at least dune 1.1

kit-ty-kate:
Removing the `v` prefix is the priority

packages/printbox/printbox.v0.3/opam
c-cube: why does it matter what the git tag is? I'd rather not, really…

mseri: The version of the path is the one used by opam, it overrides the one in the opam file and shows a warning. The path should use the correct version, in this case it seems 0.3

c-cube:
subsumed by #15333, I found another flag in dune-release

PR #15337, opened 2019-11-20
Add dune-configurator dependency to base and jst-config v0.13

diml:
Signed-off-by: Jeremie Dimino <jeremie@dimino.org>


rgrinberg:
Looks good. I believe you need to make these changes in the upstream repos as well. Also, async_ssl#master needs to be updated as well.

diml:
Yh, I've made the change in our internal repo. It will appear on github in the next few days (/cc @xclerc).

PR #15338, opened 2019-11-20
Relax the uri-sexp -> ppx_sexp_conv constraint

xclerc:


mseri:
Should we use `v0.14` instead of fully removing the bound?

xclerc:
I don't have a strong opinion, and I guess it will not
make a difference from our perspective.

xclerc:
Thanks for merging; to expand on my previous answer,
I think it would make sense to always have upper bonds
on some designated packages but I have no idea how it
should happen (I mean: what the workflow would look like).

PR #15339, opened 2019-11-20
Package range.0.8

aldrikfr:
### `range.0.8`
Fold on integer range
Library to work on range of sequential integers with features like : split, map,
filter capacities.
Designed for making distributed computing easier with libraries like functory.

---
* Homepage: https://github.com/aldrikfr/range
* Source repo: git+https://github.com/aldrikfr/range
* Bug tracker: https://github.com/aldrikfr/range/issues

---

### DIFF
--- a/packages/range/range.0.8/opam
+++ b/packages/range/range.0.8/opam
@@ -1,0 +1,23 @@
+opam-version : "2.0"
+synopsis: "Fold on integer range"
+description: """
+Library to work on range of sequential integers with features like : split, map,
+filter capacities.
+Designed for making distributed computing easier with libraries like functory.
+"""
+maintainer : "Aldrik KLEBER <contact@aldrik.net>"
+authors : "Aldrik KLEBER"
+license : "GPL-3.0-only"
+homepage : "https://github.com/aldrikfr/range"
+dev-repo: "git+https://github.com/aldrikfr/range"
+bug-reports :"https://github.com/aldrikfr/range/issues"
+depends: [
+  "ocaml" {>= "4.03.0"}
+  "base" { >="v0.12" & <"v0.14"}
+  "dune" {>= "1.8"}
+]
+build: ["dune" "build" "-p" name "-j" jobs]
+url {
+   src: "https://github.com/aldrikfr/range/archive/0.8.tar.gz"
+   checksum: "md5=ceaf14c13f420e31b7275fcfbd3e4c6f"
+}


aldrikfr:
I'm not sure of the problem, travis isn't able to grab sources. I'm not sure of how I can fix that.

mseri:
The url stanza is missing in the opam file

mseri:
I am rerunning the test

kit-ty-kate:
I'm noticing a couple of issues in your project. First there is no `dune-project` file, without this your package might break with the release of dune 2.0 or later. Also your opam file in the source repository contains a `url` section, this shouldn't be the case. `url` sections are only there for releases, not for dev packages. If someone wants to try your package using `opam pin add range --dev` for instance or simply `opam pin add .` they will get your archive and not the dev repo.

Could you retag your project with those fixes?

packages/range/range.0.8/opam
```diff
@@ -0,0 +1,23 @@
...
+maintainer : "Aldrik KLEBER <contact@aldrik.net>"
+authors : "Aldrik KLEBER"
+license : "GPL-3.0-only"
+homepage : "https://github.com/aldrikfr/range"
+dev-repo: "git+https://github.com/aldrikfr/range"
+bug-reports :"https://github.com/aldrikfr/range/issues"
+depends: [
+  "ocaml" {>= "4.03.0"}
+  "base" { >="v0.12" & <"v0.14"}
+  "dune" {>= "1.8"}
```
kit-ty-kate: This should match the `dune-project` once it's there.

aldrikfr: > This should match the `dune-project` once it's there.

I'm not sure of what you mean

aldrikfr:
> I'm noticing a couple of issues in your project. First there is no `dune-project` file, without this your package might break with the release of dune 2.0 or later. Also your opam file in the source repository contains a `url` section, this shouldn't be the case. `url` sections are only there for releases, not for dev packages. If someone wants to try your package using `opam pin add range --dev` for instance or simply `opam pin add .` they will get your archive and not the dev repo.
> 
> Could you retag your project with those fixes?

Yes ! Done ;)

packages/range/range.0.8/opam
mseri: That the lower bound here reflects the version used in https://github.com/aldrikfr/range/blob/0.8/dune-project#L1

mseri: It's fine now

mseri:
Thanks

PR #15348, opened 2019-11-21
added packages/lbvs_consent/lbvs_consent.2.1.1

UnixJunkie:
compatible with latest dolog library

### DIFF
--- a/packages/lbvs_consent/lbvs_consent.2.1.1/opam
+++ b/packages/lbvs_consent/lbvs_consent.2.1.1/opam
@@ -13,10 +13,6 @@ install: [
   ["cp" "bin/lbvs_consent_ecfp4.py" "%{bin}%/lbvs_consent_ecfp4.py"]
   ["cp" "src/lbvs_consent_ob_maccs" "%{bin}%/lbvs_consent_ob_maccs"]
 ]
-remove: [
-  ["rm" "-f" "%{bin}%/lbvs_consent_ecfp4.py"
-             "%{bin}%/lbvs_consent_ob_maccs"]
-]
 depends: [
   "dune" {>= "1.6"}
   "batteries"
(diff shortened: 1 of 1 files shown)


packages/lbvs_consent/lbvs_consent.2.1.1/opam
```diff
@@ -0,0 +1,65 @@
...
+  ["dune" "build" "-p" name "-j" jobs]
+]
+install: [
+  ["cp" "bin/lbvs_consent_ecfp4.py" "%{bin}%/lbvs_consent_ecfp4.py"]
+  ["cp" "src/lbvs_consent_ob_maccs" "%{bin}%/lbvs_consent_ob_maccs"]
+]
+remove: [
+  ["rm" "-f" "%{bin}%/lbvs_consent_ecfp4.py"
+             "%{bin}%/lbvs_consent_ob_maccs"]
+]
```
kit-ty-kate: This field is unnecessary since opam 2.0

UnixJunkie: Ok, I removed it.
Sorry, you probably already told me once before.
Opam lint doesn't say anything about it though.

mseri: In practice, the remove field is useful now only if you modify some other files around (e.g. shared configurations) during install and you need to revert such changes in the remove phase. This means we cannot really deprecate it or raise a warning. See: https://opam.ocaml.org/doc/Manual.html#opamfield-remove

Can you please make the change upstream as well?

mseri:
The failures are all due to missing or wrong versions of the external dependencies

packages/lbvs_consent/lbvs_consent.2.1.1/opam
UnixJunkie: ok, corrected upstream too

PR #15350, opened 2019-11-21
Add an OCaml upper bound to cinaps.v0.9.1

xclerc:
According to http://check.ocamllabs.io/, `cinaps.v0.9.1`
is not compatible with OCaml 4.10.0.


kit-ty-kate:
(due to force-safe-string being the new default) Thanks!

PR #15352, opened 2019-11-21
Package eliom.6.9.2

jrochel:
### `eliom.6.9.2`
Client/server Web framework
Eliom is a framework for implementing client/server Web applications. It introduces new concepts to simplify the implementation of common behaviors, and uses advanced static typing features of OCaml to check many properties of the Web application at compile-time. Eliom allows implementing the whole application as a single program that includes both the client and the server code. We use a syntax extension to distinguish between the two sides. The client-side code is compiled to JS using Ocsigen Js_of_ocaml.

---
* Homepage: http://ocsigen.org/eliom/
* Source repo: git+https://github.com/ocsigen/eliom.git
* Bug tracker: https://github.com/ocsigen/eliom/issues/

---

### DIFF
--- a/packages/eliom/eliom.6.9.2/opam
+++ b/packages/eliom/eliom.6.9.2/opam
@@ -1,0 +1,38 @@
+opam-version: "2.0"
+maintainer: "dev@ocsigen.org"
+authors: "dev@ocsigen.org"
+synopsis: "Client/server Web framework"
+description: "Eliom is a framework for implementing client/server Web applications. It introduces new concepts to simplify the implementation of common behaviors, and uses advanced static typing features of OCaml to check many properties of the Web application at compile-time. Eliom allows implementing the whole application as a single program that includes both the client and the server code. We use a syntax extension to distinguish between the two sides. The client-side code is compiled to JS using Ocsigen Js_of_ocaml."
+homepage: "http://ocsigen.org/eliom/"
+bug-reports: "https://github.com/ocsigen/eliom/issues/"
+license: "LGPL-2.1 with OCaml linking exception"
+dev-repo: "git+https://github.com/ocsigen/eliom.git"
+build: [make]
+depends: [
+  "ocaml" {>= "4.07.1"}
+  "ocamlfind"
+  "ppx_deriving"
+  "ppx_tools" {>= "0.99.3"}
+  "js_of_ocaml-compiler" {>= "3.3" & < "3.5"}
+  "js_of_ocaml" {>= "3.3"}
+  "js_of_ocaml-lwt" {>= "3.3"}
+  "js_of_ocaml-ocamlbuild" {build}
+  "js_of_ocaml-ppx" {>= "3.3"}
+  "js_of_ocaml-ppx_deriving_json" {>= "3.3" & < "3.5"}
+  "js_of_ocaml-tyxml" {>= "3.3"}
+  "lwt_log"
+  "lwt_ppx"
+  "tyxml" {>= "4.3.0"}
+  "ocsigenserver" {>= "2.10"}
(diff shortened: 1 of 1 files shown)


jrochel:
@kit-ty-kate I'm having trouble with the revdeps. CI is trying to install ocsigen-toolkit.2.2.0 and ocsigen-toolkit.2.3.2 (and fails) even though there are fixed versions available: ocsigen-toolkit.2.2.1 and ocsigen-toolkit.2.3.3.

kit-ty-kate:
@jrochel as stated in one of your previous PR, the failing packages need to be fixed as their dependency constraints were not restrictive enough. Any packages that you add to opam-repository will always be around and available to install by users, they need to be fixed as well in occasion.

As opam-repository maintainers we can do this, however it takes a lot of time and it is usually better for the package maintainer themselves to do it as they usually are more knowledgable about what needs to be fixed in their packages.

jrochel:
Is the official opam policy really to overwrite existing versions with fixed ones? I don't like that approach at all because it might lead to situations where some user says: "I can't install ocsigen-toolkit.2.2.0" and I have to reply: "Which version of ocsigen-toolkit.2.2.0" were you trying to install, the one that you downloaded before or after the 11/10 13:44 UTC?"
In my mind the last component of the version number serves to distinguish exactly that: bugfixes and compatibility issues.

kit-ty-kate:
> Is the official opam policy really to overwrite existing versions with fixed ones?

Yes. The thing is that your package is broken **right now**, you will get users saying this in any case. Those packages are still installable and it is policy to not remove them. Opam does not have "revisions" and thus if one package is missing it might break things for people using `opam lock` for instance.

If packages are broken it is my problem and users' problem. For now we don't have a way of dealing with this. If you have some insights them please don't hesitate to open an issue on [opam's issue tracker](https://github.com/ocaml/opam/issues) or add your voice to this issue: https://github.com/ocaml/opam-repository/issues/10531

Please consider fixing your packages in the meantime though. If you don't have time I will do it but if you have I would probably better off doing something else.

jrochel:
https://github.com/ocaml/opam-repository/pull/15391
https://github.com/ocaml/opam-repository/pull/15390
https://github.com/ocaml/opam-repository/pull/15389
https://github.com/ocaml/opam-repository/pull/15388

jrochel:
@kit-ty-kate: I created an issue to discuss the above problem: https://github.com/ocaml/opam-repository/issues/15392

kit-ty-kate:
> #15391
> #15390
> #15389
> #15388

I'm sorry I don't understand the changes. Why did the archives change? This breaks even more the repository instead of fixing it. `opam publish` should not be used as a tool to fix packages.

avsm:
@jrochel wrote:
> Is the official opam policy really to overwrite existing versions with fixed ones?

To clarify, there is a difference between the package _contents_ and the opam _metadata_.  The contents are considered immutable for a given version, but the metadata (e.g. package constraints, description, conflicts, etc. in an opam file) are regularly updated to reflect the environment in which users install a particular package.  For example, it is common to add an upper bound to packages when an incompatible version of some dependency is added, so that already released packages do not break for users.

In the case of your PRs, it looks like you have also updated the package contents (the archive).  This is frowned upon in opam-repository, since it adversely affects caching among other things.

jrochel:
Thanks. Are these things documented somewhere?

avsm:
We're working on [new package submission infrastructure](https://ocurrent.org) that mechanises this policy (as the "camelus" bot does for some metadata at the moment). It's not explicitly written down anywhere else, but contributions welcome to kickstart an effort to do so.

kit-ty-kate:
All green, thanks

PR #15359, opened 2019-11-22
angstrom-async is compatible with async.v0.13.0

xclerc:
(We need to relax these constraints for an upcoming JS package.)


xclerc:
The Travis failures are actually timeouts.

kit-ty-kate:
Thanks

PR #15364, opened 2019-11-22
mirage 3.7.2: relax bounds on mirage-runtime

hannesm:
/cc @samoht I suspect this is why mirage 3.7.1 still gets installed in mirage-www travis job. I'm not entirely sure what to do in the mirage/mirage repository with this dependency -- a `=version` is for sure the safest since others may need to be modified upon release.. if we'd always release mirage, mirage-runtime, mirage-types, mirage-types-lwt at once, we would not run into this issue at all...

### DIFF
--- a/packages/mirage/mirage.3.7.2/opam
+++ b/packages/mirage/mirage.3.7.2/opam
@@ -25,7 +25,7 @@ depends: [
   "astring"
   "logs"
   "stdlib-shims"
-  "mirage-runtime"     {= version}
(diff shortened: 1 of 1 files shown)


samoht:
Thanks for the fix.

I feel that's wrong to release twice the same code with a different version number but if that helps I can try to do this next time.

hannesm:
@samoht I agree. maybe we should relax the constraints in the mirage/mirage repository as well, and have a release checklist (or remember) to adjust the bounds if necessary.

PR #15365, opened 2019-11-22
relaxes ezjsonm constraints for sexplib

gitoleg:
It looks like that the constraint is superfluous as we can easily install ezjsonm with the latest core, e.g.,
`opam install ezjsonm --ignore-constraints-on=sexplib`
works without any issues.

cc @samoht


kit-ty-kate:
Thanks

PR #15367, opened 2019-11-22
Fix conf-mariadb on MacOS

kit-ty-kate:
Per comment of @hcarty in https://github.com/ocaml/opam-repository/pull/15265#issuecomment-557730655


hcarty:
Thank you so much for fixing this!

kit-ty-kate:
Looks like it works. Thanks for the report!

PR #15384, opened 2019-11-25
switched oranger from jbuilder to dune

UnixJunkie:


packages/oranger/oranger.2.0.0/opam
```diff
@@ -0,0 +1,36 @@
...
+]
+install: [
+  ["sh" "install_ranger.sh"]
+]
+depends: [
+  "dune"
+  "conf-cmake" {build}
+  "conf-wget" {build}
+  "re"
+  "dolog"
```
kit-ty-kate:
```suggestion
  "dolog" {>= "4.0.0"}
```
I'm guessing this requires at least dolog 4.0.0 since anything would be incompatible otherwise?

packages/oranger/oranger.2.0.0/opam
```diff
@@ -0,0 +1,36 @@
...
+  "batteries"
+  "minicli" {with-test}
```
kit-ty-kate:
```suggestion
  "minicli" {with-test & >= "5.0.0"}
```
Same here?

kit-ty-kate:
One of the archive is failing due to an unmatching checksum
```
          - oranger.2.0.0:
            https://github.com/imbs-hl/ranger/archive/0.9.11.tar.gz (Bad checksum, expected md5=91e80fd65f21cf68fcdedd442ffdc901)
```

UnixJunkie:
I added a lower bound version constraint to dune and updated the MD5 for the external source tarball.

UnixJunkie:
I need to fix and retag the software.

packages/oranger/oranger.2.0.0/files/compile_ranger.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
kit-ty-kate: This file is in the wrong directory. It should be in `packages/oranger/oranger.2.0.1/files`

UnixJunkie: This file was left out of the previous commit by mistake; I just pushed it.

kit-ty-kate:
Thanks!

PR #15393, opened 2019-11-25
ocsigen-toolkit: upper version bound on js_of_ocaml(-lwt)

jrochel:


kit-ty-kate:
Checksums for ocsigen-toolkit 2.1.0, 2.2.0 and 2.3.1 are not matching, could you tag the right commit back to where there were before, if this works?

jrochel:
@kit-ty-kate Are you alright with me removing unsupported versions of the package?

jrochel:
@kit-ty-kate *bump* (Sorry, I'd really like to release eliom.6.9.2.)

kit-ty-kate:
Sorry to leave you hanging. We usually don't delete packages, what if we I take over the maintenance for those releases?

kit-ty-kate:
Alternatively they can be marked as `available: false` to effectively disable them but I'm not sure if it makes any sense if the archives are not available anymore

kit-ty-kate:
Ok, `available: false` will do for now. I'll merge this as soon as the CI is green, sorry for the delay.

kit-ty-kate:
@thomasblanc is Camelus not saying when a package is not available anymore? The `Installability check` is empty but packages were removed. I'm pretty sure it used to say `not installable anymore` or something like that.

kit-ty-kate:
Thanks, sorry again.

jrochel:
No worries, I don't have any sense of entitlement.
In any case, the archives are still available (or rather: available again).
Thanks. I'll stick to the `available: false` approach in the future.

PR #15394, opened 2019-11-25
Release ppx_regexp.0.4.3 and ppx_tyre.0.4.3.

paurkedal:
This a fix and compiler upgrade for two PPXes built from the same repo. From the change log:

- Fixed nested `[%pcre]` usage for `ppx_regexp`.
- Extended compiler support to 4.02.3 up to 4.09.0 (at least) for both PPXes.
- Upgrade to AST 4.09 to support newer compiler features.


kit-ty-kate:
Thanks a lot

PR #15397, opened 2019-11-26
added packages/orrandomForest/orrandomForest.2.0.1

UnixJunkie:
use dune instead of jbuilder plus latest version of dolog


UnixJunkie:
The installation of the R packages fails in the CI environment.
Well, it works fine on my test computers. :)

kit-ty-kate:
> The installation of the R packages fails in the CI environment.
> Well, it works fine on my test computers. :)

Have you disabled sandboxing or something like that? It seems that the R scripts are making a connexion somewhere.

In the meantime this can go in. Thanks!

UnixJunkie:
The R scripts are installing dependencies. I.e. using the R package manager to install required R libraries.

PR #15398, opened 2019-11-26
added packages/setcore/setcore.1.0.2

UnixJunkie:
remove the upper bound version constraint on dune and add a lower bound instead


kit-ty-kate:
Depends on https://github.com/ocaml/opam-repository/pull/15399

kit-ty-kate:
Thanks!

PR #15400, opened 2019-11-26
mirage-qubes 0.8.0: restrict to cstruct < 5.1.0 (since 5.1.0 @len 0 is forbidden)

hannesm:
reported upstream https://github.com/mirage/mirage-qubes/issues/47


kit-ty-kate:
Thanks

PR #15402, opened 2019-11-26
[new release] exit (0.0.1)

Cameleo:
Get exit status as declared in `stdlib.h`

- Project page: <a href="https://git.zapashcanon.fr/zapashcanon/exit">https://git.zapashcanon.fr/zapashcanon/exit</a>
- Documentation: <a href="https://doc.zapashcanon.fr/exit/">https://doc.zapashcanon.fr/exit/</a>

##### CHANGES:

First release

### DIFF
--- a/packages/exit/exit.0.0.1/opam
+++ b/packages/exit/exit.0.0.1/opam
@@ -1,0 +1,37 @@
+opam-version: "2.0"
+synopsis: "Get exit status as declared in `stdlib.h`"
+description:
+  "exit is an OCaml library to get exit status as declared in `stdlib.h`. It gives access to the values of `EXIT_SUCCESS` and `EXIT_FAILURE` macros and provides some functions built around them."
+maintainer: ["Léo Andrès <contact@ndrs.fr>"]
+authors: ["Léo Andrès <contact@ndrs.fr>"]
+license: "ISC"
+homepage: "https://git.zapashcanon.fr/zapashcanon/exit"
+doc: "https://doc.zapashcanon.fr/exit/"
+bug-reports: "https://git.zapashcanon.fr/zapashcanon/exit/issues"
+depends: [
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Hi, I'm not sure of the usefulness of this library to be honest. According to the C documentation (https://en.cppreference.com/w/c/program/EXIT_status):
```
Both EXIT_SUCCESS and the value zero indicate successful program execution status (see exit),
although it is not required that EXIT_SUCCESS equals zero.
```
I am unaware of real-life systems using values other than 0 for success and non-zero values for failures.

kit-ty-kate:
I'm not against merging your library in opam-repository, however as it is not practice to remove packages from opam-repository, I'd like to be convince that a library with such a recognisable name will not just be discarded after one version.

zapashcanon:
> I am unaware of real-life systems using values other than 0 for success and non-zero values for failures.

Well, there's at least [VMS](https://en.wikipedia.org/wiki/OpenVMS) which does weird things...

But I'm mostly using this library to avoid writing things like `exit 1` which is always less clear than `exit_failure ()`. It's also useful for teaching, meaningful names are always better than some weird historic numbers.

I'm also planning on adding stuff to the library to get some kind of framework for exit status. I recently got annoyed with a project where I wanted to have a unique exit status code for some specific error, so I picked 2, but then realized (after some painful debug) that it was the error code used by OCaml for uncaught exception...

> I'd like to be convince that a library with such a recognisable name will not just be discarded after one version

I'm going to maintain it, if that's what you're asking. About the name, I can change it if you want, I don't really care, I just took this one as it allows me to write `Exit.sucess ()` and `Exit.failure ()` directly.

kit-ty-kate:
Ok fair. I'll merge it once the MacOS CI is green.

kit-ty-kate:
Thanks

PR #15404, opened 2019-11-26
Add ocaml-secondary-compiler and ocamlfind-secondary

dra27:
This PR allows OCaml 4.08.1 to be installed in any non-4.08.x switch as the package `ocaml-secondary-compiler`. An additional package `ocamlfind-secondary` exposes this compiler as `-toolchain secondary` and additionally installs an `ocaml` `META` allowing the location of the compiler to be detected.

Dune 2.0.0 already has baked-in support to use this `ocamlfind` toolchain if it's found, so this PR also relaxes the constraint on the Dune 2.0.0 package.

Rather than effectively distributing the compiler inside the Dune 2.0.0 package (which would have the serious downside of rebuilding the compiler on any upgrade of Dune), this has been packaged in a way where the same mechanism can in principle be used by any program which requires a version of OCaml greater than the version of OCaml in the switch into which it wants to be installed. It's not expected that this will be of use to (m)any other projects, given that you need to vendor your dependencies. It's also intended as a stop-gap until a future version of `opam` permits making binaries globally available.


kit-ty-kate:
I like this PR very much for use in [check.ocamllabs.io](http://check.ocamllabs.io), however it is to be noted that without some sort of change in the CIs currently running, this PR is going to slow down pretty much every job testing packages < 4.06 (at 4 per packages modified/added per PRs) by around 4 minutes per jobs (at least 16 minutes per packages pr PRs) if the packages use dune:
```
-> installed conf-m4.1
-> installed ocamlfind.1.8.1
-> installed ocaml-secondary-compiler.4.08.1
-> installed ocamlfind-secondary.1.8.1
Done.

real    3m58.976s
```
I'd suggest waiting for a fix in `mirage-ci` first (I'm on it)

packages/dune/dune.2.0.0/opam
```diff
@@ -26,7 +26,7 @@ homepage: "https://github.com/ocaml/dune"
 doc: "https://dune.readthedocs.io/"
 bug-reports: "https://github.com/ocaml/dune/issues"
 depends: [
-  "ocaml" {>= "4.06"}
+  ("ocaml" {>= "4.06"} | "ocaml" {< "4.06~~"} & "ocamlfind-secondary")
```
kit-ty-kate:
```suggestion
  ("ocaml" {>= "4.06"} | ("ocaml" {< "4.06~~"} & "ocamlfind-secondary"))
```
Are you certain a parenthesis is not needed here? (I actually don't know)

dra27: I think it might have been in opam 1, but not in 2. I tested this with a 4.08.1 and 4.02.3 switch and only the latter tried to install ocamlfind-secondary, so it seemed to be working as I expected! Feel free to add it for clarity, of course

kit-ty-kate:
All good, thanks a lot!

PR #15408, opened 2019-11-27
[new release] owl-base, owl, owl-zoo and owl-top (0.7.1)

mseri:
OCaml Scientific and Engineering Computing - Base

- Project page: <a href="https://github.com/owlbarn/owl">https://github.com/owlbarn/owl</a>
- Documentation: <a href="https://owlbarn.github.io/">https://owlbarn.github.io/</a>

##### CHANGES:

* Add unit basis
* Fix issue #337 and #457 (#458)
* owl-base: drop seemingly unnecessary dependency on integers (#456)

### DIFF
--- a/packages/owl-top/owl-top.0.7.1/opam
+++ b/packages/owl-top/owl-top.0.7.1/opam
@@ -1,0 +1,30 @@
+opam-version: "2.0"
+maintainer: "Liang Wang <ryanrhymes@gmail.com>"
+authors: [ "Liang Wang" ]
+license: "MIT"
+homepage: "https://github.com/owlbarn/owl"
+dev-repo: "git+https://github.com/owlbarn/owl.git"
+bug-reports: "https://github.com/owlbarn/owl/issues"
+doc: "https://owlbarn.github.io/"
+synopsis: "OCaml Scientific and Engineering Computing - Top"
+description: "Owl is an OCaml scientific library."
+
+build: [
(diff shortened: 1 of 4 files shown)


mseri:
The link time issues in ubuntu and debian unstableare unfortunately expected

packages/owl-top/owl-top.0.7.1/opam
```diff
@@ -0,0 +1,30 @@
...
+build: [
+  [ "dune" "subst" ] {pinned}
+  [ "dune" "build" "-p" name "-j" jobs ]
+]
+
+depends: [
+  "ocaml" {>= "4.06.0"}
+  "dune" {>= "1.7.0"}
+  "ocaml-compiler-libs"
+  "owl" {>= version}
```
kit-ty-kate:
```suggestion
  "owl" {= version}
```
Wouldn't it be more error prone to use `=` instead of `>=`? (same for the other sub-packages)

mseri:
These packages use a microscopic stable part of the API, in fact we could use >= 0.4 and it would still be fine. We use the version lower bound to be over zealous. It may be worth doing it for owl-base but the api is only growing, not expected to shrink

mseri:
I have added the constraint to owl wrt owl-base. I’ll do the same upstream

kit-ty-kate:
Thanks!

PR #15411, opened 2019-11-27
jbuilder 1.11.4

diml:
This is a one of release. It is the jbuilder binary extracted from dune 1.11.4 as a separate package. This is to allow the following packages to be co-installed:

- dune >= 2.0.0
- ocaml >= 4.08
- jbuilder

In particular, this allows to co-install both dune >= 2.0.0 in conjunction with packages that haven't upgraded to dune when using a recent compiler.

### DIFF
--- a/packages/jbuilder/jbuilder.1.11.4/opam
+++ b/packages/jbuilder/jbuilder.1.11.4/opam
@@ -1,0 +1,28 @@
+opam-version: "2.0"
+synopsis: "Fast, portable and opinionated build system (deprecated)"
+description: """
+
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
To be totally co-installable, `"jbuilder" {!= "transition"}` should be removed from the `conflicts` section of `dune.2.0.0` first.

diml:
The conflict is actually `"jbuilder" {= transition}` for dune 2.0.0, so it should be good. That said, I was thinking more about the original problem and this PR might not completely solve the issue. I'm going to do more testing on my side.

PR #15416, opened 2019-11-29
[new release] Add Coq 8.10.2

Blaisorblade:
Adapted from 8.10.1.

CoqIDE will be a separate PR to avoid holding Coq up in case of issues.


kit-ty-kate:
All green, thanks!

PR #15423, opened 2019-12-01
Package lua-ml.0.9

dmbaturin:
### `lua-ml.0.9`
An embeddable Lua 2.5 interpreter implemented in OCaml

---
* Homepage: https://github.com/lindig/lua-ml
* Bug tracker: https://github.com/lindig/lua-ml/issues

---

### DIFF
--- a/packages/lua-ml/lua-ml.0.9/opam
+++ b/packages/lua-ml/lua-ml.0.9/opam
@@ -10,8 +10,7 @@ bug-reports: "https://github.com/lindig/lua-ml/issues"
 dev-repo: "git+https://github.com/lindig/lua-ml.git"
 depends: [
   "ocaml" {>= "4.07"}
-  "ocamlbuild"
-  "ocamlfind"
+  "ocamlbuild" {build}
 ]
 build: [make "lib"]
 url {
--- a/packages/lua-ml/lua-ml.0.9/opam
+++ b/packages/lua-ml/lua-ml.0.9/opam
@@ -7,6 +7,7 @@ authors: [
 license: "Two-clause BSD"
 homepage: "https://github.com/lindig/lua-ml"
 bug-reports: "https://github.com/lindig/lua-ml/issues"
+dev-repo: "git+https://github.com/lindig/lua-ml.git"
 depends: [
   "ocaml" {>= "4.07"}
   "ocamlbuild"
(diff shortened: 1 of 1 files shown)


dmbaturin:
This is the Lua interpreter from the defunct C-- project that Christian Lindig rescued and made into a standalone project, then Gabriel Radanne made opam packaging for it, and then I acccidentally ended up being its maintainer and hacked it to work with modern OCaml versions and cleared it from licensing issues.

The great thing about it is that it can interface with OCaml in a typesafe way and you the Lua library is assembled using functors—you can embed your own modules as if they always were there or replace parts of the standard library with your own modules.

The bad thing is that it requires heavy wizardry to use and that needs to be documented. Like https://github.com/dmbaturin/soupault/blob/master/src/plugin_api.ml#L148-L269

I and a few other people have been using it for writing soupault plugins for a while and it seems usable enough to bring to a wider audience (which, hopefully, will also bring more contributors in).

packages/lua-ml/lua-ml.0.9/opam
```diff
@@ -0,0 +1,23 @@
...
+authors: [
+  "Norman Ramsey <nr@cs.tufts.edu>" "Christian Lindig <lindig@gmail.com>"
+]
+license: "Two-clause BSD"
+homepage: "https://github.com/lindig/lua-ml"
+bug-reports: "https://github.com/lindig/lua-ml/issues"
+dev-repo: "git+https://github.com/lindig/lua-ml.git"
+depends: [
+  "ocaml" {>= "4.07"}
+  "ocamlbuild"
```
kit-ty-kate:
```suggestion
  "ocamlbuild" {build}
```
I think this dependency should be tagged as a build dependency

packages/lua-ml/lua-ml.0.9/opam
```diff
@@ -0,0 +1,23 @@
...
+  "ocamlfind"
```
kit-ty-kate:
```suggestion
```
I haven't found any use of ocamlfind anywhere in the code, I think is an artefact of a use of a different build system

kit-ty-kate:
This should be good to go once the two comments have been taken care of

kit-ty-kate:
Thanks!

PR #15424, opened 2019-12-02
added packages/parany/parany.7.0.3

UnixJunkie:
hardening against possible bug (Unix.select being interrupted at runtime).
Should be very rare.


kit-ty-kate:
Thanks!

PR #15431, opened 2019-12-03
Blacklist lablgtk3 3.0.beta7 also from Coq 8.10.{0,1}

Blaisorblade:
Workaround for coq/coq#11217; I've confirmed it affects those versions as well.

### DIFF
--- a/packages/coqide/coqide.8.10.0/opam
+++ b/packages/coqide/coqide.8.10.0/opam
@@ -9,7 +9,8 @@ synopsis: "IDE of the Coq formal proof management system"
 
 depends: [
   "coq" {= version}
-  "lablgtk3-sourceview3"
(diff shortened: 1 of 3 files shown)


mseri:
Does the issue affect only osx? In that casw should we use something linke
```
conflicts: ["lablgtk3-sourceview3" {= "3.0.beta7" & os-family = "macos"}
```
or it is not supported by opam?

CC @garrigue (fyi)

Blaisorblade:
AFAIK the bug shouldn't be mac-specific, but I haven't confirmed yet. Should I split the PR so that testing can pass on Travis?

ejgallego:
Indeed this bug is not mac-specific; note that 8.11 won't have this problem.

mseri:
Thanks for the clarification

PR #15436, opened 2019-12-03
Release rpclib and ppx_deriving_rpc 6.0.0 

anton-trunov:
I just hope @jonludlam and @mseri wouldn't mind.

### DIFF
--- a/packages/ppx_deriving_rpc/ppx_deriving_rpc.6.0.0/opam
+++ b/packages/ppx_deriving_rpc/ppx_deriving_rpc.6.0.0/opam
@@ -18,7 +18,7 @@ depends: [
   "jbuilder"
   "rpclib" {>= "6.0.0"}
   "rresult"
-  "ppxlib"
+  "ppxlib" {< 0.9.0}
   "rpclib-lwt" {with-test & >= "6.0.0"}
   "rpclib-async" {with-test & >= "6.0.0"}
   "lwt" {with-test & >= "3.0.0"}
--- a/packages/ppx_deriving_rpc/ppx_deriving_rpc.6.0.0/opam
+++ b/packages/ppx_deriving_rpc/ppx_deriving_rpc.6.0.0/opam
@@ -18,7 +18,7 @@ depends: [
   "jbuilder"
   "rpclib" {>= "6.0.0"}
   "rresult"
-  "ppxlib" {< 0.9.0}
+  "ppxlib" {< "0.9.0"}
   "rpclib-lwt" {with-test & >= "6.0.0"}
   "rpclib-async" {with-test & >= "6.0.0"}
   "lwt" {with-test & >= "3.0.0"}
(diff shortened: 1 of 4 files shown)


avsm:
This still uses jbuilder - any reason not to release something with dune instead?

anton-trunov:
@avsm I would also like that. `ocaml-rpc`'s authors already did port it to dune, but there is no official release with that support at the moment.

anton-trunov:
6.0.0 is out for about a year and I'd like this to be released because the project I'm working on needs to move on with its deps

mseri:
I've been wanting to do that for over a year, thanks. The reason it was not published is that we had not benchmarked the new version and the new monadic interface was a beta: there is lots of room for imprivement there but it will require some further breaking changes to be right. Also we needed to port the derivers to ppxlib.

By releasing I  hope we will get more users and feedback for that.

mseri:
You need to release the rpclibs with it, there were a number of breaking changes with it

mseri:
I can draft a release with the dune port later tomorrow if you want. And then we can release the one supprting newer compilers once all the PRs are merged

anton-trunov:
> I can draft a release with the dune port later tomorrow if you want. And then we can release the one supprting newer compilers once all the PRs are merged

That would be extremely helpful! thanks a lot in advance

> You need to release the rpclibs with it, there were a number of breaking changes with them
You are right, thanks, I'm going to add this.

mseri:
> Btw, it worked on my machine with OCaml 4.08.1. Can we please try it with `"ocaml" {>= "4.03.0" & < "4.09.0"}`? (after I add rpclib 6.0.0 -- I forgot that I installed it from my local copy of ocaml-rpc).

Done. Let's wait, maybe it is not needed afetr all

anton-trunov:
> Done. Let's wait, maybe it is not needed afetr all

Indeed, ppx_deriving_rpc 6.0.0 needs rpclib 6.0.0, I could also reproduce it locally.

mseri:
Then it probably also needs the async and lwt ones. We should probably do a full release

mseri:
If it's green, we will need to remove `{build}` from all jbuilders before merge

jonludlam:
:+1:

anton-trunov:
Everything was green, except the build on macOS timed out:

>The job exceeded the maximum time limit for jobs, and has been terminated.

anton-trunov:
I also managed to build locally with OCaml 4.09, it turned out that `ppxlib` should have been constrained to at most v0.8.
I'd try adding `< 4.10` constraint on OCaml to see if I could also use 4.09 in the project I'm working on.

anton-trunov:
@mseri macOS build timed out once again (but my local macOS builds succeed).

mseri:
Thanks!

anton-trunov:
Awesome, thank you!

PR #15437, opened 2019-12-03
base.v0.13.0 is not compatible with OCaml 4.10

xclerc:


xclerc:
For the record, base does *build* with 4.10 [1], but then
one cannot *link* against it (see [2]).

[1] http://check.ocamllabs.io/log/1574696611-eec32e8750d18eaf1eb0d416d0f03b9e0f1df40f/4.10/good/base.v0.13.0
[2] https://github.com/ocaml/ocaml/issues/9014

mseri:
Thanks

PR #15439, opened 2019-12-03
[new release] lablgtk3-gtkspell3, lablgtk3-sourceview3 and lablgtk3 (3.0.beta8)

garrigue:
OCaml interface to GTK+3

- Project page: <a href="https://github.com/garrigue/lablgtk">https://github.com/garrigue/lablgtk</a>
- Documentation: <a href="https://garrigue.github.io/lablgtk/lablgtk3-gtkspell3">https://garrigue.github.io/lablgtk/lablgtk3-gtkspell3</a>

##### CHANGES:

2019.12.03 [Jacques]
  * update toplevel loop instructions in README.md
  * keep Gtk{Th,}Init in tools for reference

2019.12.02 [Emilio]
  * remove Gtk{Th,}Init static initializers, applications should init
    Gtk manually instead of using a linking side-effect (garrigue/lablgtk#100)

2019.11.27 [Jacques]
  * remove declaration of young_start for ocaml 4.10 (garrigue/lablgtk#97)
  * add CI support for 4.10 [Emilio]

2019.11.26 [Emilio]
  * Require Dune 1.8, use improved configurator API (garrigue/lablgtk#95)

### DIFF
--- a/packages/lablgtk3-gtkspell3/lablgtk3-gtkspell3.3.0.beta8/opam
+++ b/packages/lablgtk3-gtkspell3/lablgtk3-gtkspell3.3.0.beta8/opam
@@ -1,0 +1,47 @@
+opam-version: "2.0"
+
+synopsis: "OCaml interface to GTK+3"
+description: """
+OCaml interface to GTK+3, gtkspell library
+
+See https://garrigue.github.io/lablgtk/ for more information.
+
+"""
+
+maintainer: "garrigue@math.nagoya-u.ac.jp"
+authors: ["Jacques Garrigue et al., Nagoya University"]
+homepage: "https://github.com/garrigue/lablgtk"
+bug-reports: "https://github.com/garrigue/lablgtk/issues"
+dev-repo: "git+https://github.com/garrigue/lablgtk.git"
+doc: "https://garrigue.github.io/lablgtk/lablgtk3-gtkspell3"
+license: "LGPL with linking exception"
+
+depends: [
+  "ocaml"                { >= "4.05.0" }
+  "dune"                 { >= "1.8.0"  }
+  "lablgtk3"             {  = version  }
+]
+depexts: [
+  ["gtkspell3-dev"] {os-distribution = "alpine"}
+  ["gtkspell3"] {os-distribution = "archlinux"}
+  ["epel-release" "gtkspell3-devel"] {os-distribution = "centos"}
+  ["libgtkspell3-3-dev"] {os-distribution = "debian"}
+  ["gtkspell3-devel"] {os-distribution = "fedora"}
+  ["gtkspell3"] {os = "freebsd"}
+  ["gtkspell3"] {os = "openbsd"}
+  ["gtkspell3-devel"] {os-family = "suse"}
+  ["libgtkspell3-3-dev"] {os-distribution = "ubuntu"}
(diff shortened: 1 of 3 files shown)


garrigue:
The failures on Alpine and CentOS were expected.
However the failure on framaC is more surprising, particularly as it comes from a failure on conf-gtksourceview3, which succeeded in the library build.
I'm trying to rebuild that one to see.

garrigue:
The failure on `frama-c` is stable. I think it comes from the
```
  ( ( "lablgtk" { >= "2.18.2" } & "conf-gnomecanvas" )
  | ( "lablgtk3" { >= "3.0.beta4" & os!="macos" } & "lablgtk3-sourceview3" ))
```
lines in the `depends` section of its `opam` file.
Namely, it causes `frama-c` to be detected as depending on `lablgtk3` rather than `lablgtk3-sourceview3`, so that it attempts to build it in an environment where only `lablgtk3` is installed.
It then attempts to install `lablgtk3-sourceview3`, which depends on `conf-gtksourceview3`, but the depexts for this last packages are not installed.

mseri:
ping @maroneze for the frama-c issue.

@garrigue you suggest to merge then?

garrigue:
I think the problem with frama-c is unrelated with changes in this release, so it should be fine to merge.

Correct dependencies in frama-c would probably be something like:
```
  ( ( "lablgtk" { >= "2.18.2" } & "conf-gnomecanvas" & "conf-gtksourceview" )
  | ( "lablgtk3-sourceview3" { >= "3.0.beta4" & os!="macos" }) )
```

vprevosto:
In fact, there is already a dependency in `lablgtk3-sourceview3` in the current Frama-C `opam` file. I think that the issue is more subtle: judging from the logs, `opam depext frama-c` considers that it should build the lablgtk2 external dependencies (namely  `libgnomecanvas2-dev`, `libgtk2.0-dev` and `libgtksourceview2.0-dev`), but not the one tied to `lablgtk3` and `lablgtk3-sourceview3` (the other part of the disjunction in Frama-C's dependencies). Afterwards, `opam install` seems to take the opposite direction, insisting on building `lablgtk3-sourceview3` (maybe because since `lablgtk3` is already installed, there's only one package to install in this branch, while for the other we still have `lablgtk` and `conf-gnomecanvas` left to install).

On the other hand, I agree that `conf-gtksourceview` is currently missing in the `lablgtk` branch of the dependencies.

garrigue:
What I meant is that revdeps are checked in an environment where the base library is already installed (with all its depexts). So this would be fine to compile frama-c with lablgtk3-gtksourceview3 installed. However, since frama-c has also an explicit dependency on lablgtk3, it is compiled with only lablgtk3 installed, which results in the failure.
I agree that the disjunctive dependency is another reason for this failure, but fixing that one would requiring creating two different packages, I'm afraid. You could add an explicit dependency on conf-gtksourceview3 in both cases, but it's going to break on some architectures.

kit-ty-kate:
The failure is a known problem in opam/opam-depext where opam-depext calls opam solver twice (once to get the depexts and the other to install) and the solve give different results each time. This should be resolved in opam 2.1.0 once https://github.com/ocaml/opam/pull/3975 gets merged.

This PR is good to go. Thanks!

PR #15443, opened 2019-12-04
Update conf-mariadb to work with recent alpine releases

hcarty:
The recent #15265 update broke an internal alpine-based CI build with mysql 1.2.4 because of the new `conf-*` dependency.  I think this should work, but we'll see how it does with the repository CI checks!

If the alpine CI tests here in opam-repository don't pass then this probably shouldn't be merged.

### DIFF
--- a/packages/conf-mariadb/conf-mariadb.1/opam
+++ b/packages/conf-mariadb/conf-mariadb.1/opam
@@ -6,8 +6,8 @@ bug-reports: "https://jira.mariadb.org/projects/MDEV/issues"
 dev-repo: "git+https://github.com/MariaDB/server.git"
 license: "GPL-2.0-only"
 build: [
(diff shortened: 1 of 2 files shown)


hcarty:
If I'm reading the CI results correctly I think the alpine check worked.

mseri:
Thanks. @hcarty can you please make this change in `conf-mariadb.2` or`conf-mariadb.1-1` instead? This way users of older version that currently have no issue with the package can pin it and keep having no issue

hcarty:
@mseri Certainly! I'll send an update a little later today

hcarty:
@mseri Hopefully addressed now, please let me know what you think

mseri:
Thanks

hcarty:
Thank you @mseri!

PR #15445, opened 2019-12-04
Fix rpclib's dependencies

anton-trunov:
Otherwise e.g. ppx_deriving_rpc 5.9.0 installs with rpclib 6.0.0, causing a build error.
This is a followup PR on #15436.


mseri:
I am always wary of `=version` being too strict but this is released so rarely that I think it is the best idea

mseri:
Networks issues on osx do not depend on this.

anton-trunov:
@mseri Thanks you so much!

PR #15451, opened 2019-12-04
tcpip: support dune-configurator so dune 2.0 works as well

avsm:
upstream at https://github.com/mirage/mirage-tcpip/pull/422
cc @dinosaure


mseri:
Thanks

PR #15452, opened 2019-12-04
relax ocamlformat constraint on base/stdio

avsm:
This is handy to have before the mass relaxation of other upper
bounds, since ocamlformat is needed early


mseri:
Thanks

PR #15456, opened 2019-12-05
vcaml.v0.13.0 - OCaml bindings for the Neovim API

xclerc:
The Neovim text editor comes with an RPC-based public API 
which can be used to control the editor.  This set of libraries implements 
an OCaml interface to those APIs, for the purpose of permitting neovim 
plugins to be written in OCaml.

### DIFF
--- a/packages/vcaml/vcaml.v0.13.0/opam
+++ b/packages/vcaml/vcaml.v0.13.0/opam
@@ -1,0 +1,34 @@
+opam-version: "2.0"
+maintainer: "opensource@janestreet.com"
+authors: ["Jane Street Group, LLC <opensource@janestreet.com>"]
+homepage: "https://github.com/janestreet/vcaml"
+bug-reports: "https://github.com/janestreet/vcaml/issues"
+dev-repo: "git+https://github.com/janestreet/vcaml.git"
+license: "MIT"
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+]
+depends: [
+  "ocaml" {>= "4.08.0"}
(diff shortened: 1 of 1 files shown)


packages/vcaml/vcaml.v0.13.0/opam
```diff
@@ -0,0 +1,34 @@
...
+  "ocaml" {>= "4.08.0"}
+  "async" {>= "v0.13" & < "v0.14"}
+  "async_extra" {>= "v0.13" & < "v0.14"}
+  "base" {>= "v0.13" & < "v0.14"}
+  "core" {>= "v0.13" & < "v0.14"}
+  "core_kernel" {>= "v0.13" & < "v0.14"}
+  "ppx_jane" {>= "v0.13" & < "v0.14"}
+  "angstrom"
+  "angstrom-async"
+  "dune"           {>= "2.0.0"}
```
kit-ty-kate:
```suggestion
  "dune"           {>= "1.5"}
```
According to the `dune-project` only at least dune 1.5 is needed.
It's not a blocker for to merge this but just if you want to fix this in the future

kit-ty-kate:
Thanks

packages/vcaml/vcaml.v0.13.0/opam
xclerc: This is indeed an inconsistency in our export process,
thanks for pointing it out.

We do want to depend on dune 2.0+ for configurator-related
issues, so it would make sense to bump the version of the
dune language. (The fix is in general not as simple as a mere
bump, because of deprecations.)

PR #15461, opened 2019-12-06
added packages/conf-haveged/conf-havaged.1.0.0

UnixJunkie:


packages/conf-haveged/conf-havaged.1.0.0/opam
```diff
@@ -0,0 +1,25 @@
+opam-version: "2.0"
+homepage: "https://linux.die.net/man/8/haveged"
+author: "A. Seznec and N. Sendrier"
+maintainer: "https://github.com/ocaml/opam-repository/issues"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: "GPL-1.0-or-later"
+depexts: [
+# ["havaged"] {os = "openbsd"} # not in openbsd?!
+# ["havaged"] {os-family = "debian"} # not in Debian?!
```
kit-ty-kate:
```suggestion
  ["havaged"] {os-family = "debian"}
```
It seems to exist https://packages.debian.org/buster/haveged

packages/conf-haveged/conf-havaged.1.0.0/opam
```diff
@@ -0,0 +1,25 @@
+opam-version: "2.0"
+homepage: "https://linux.die.net/man/8/haveged"
+author: "A. Seznec and N. Sendrier"
+maintainer: "https://github.com/ocaml/opam-repository/issues"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: "GPL-1.0-or-later"
+depexts: [
```
kit-ty-kate:
```suggestion
build: ["haveged" "--help"]
depexts: [
```
Maybe that could help detect if it is actually installed quicker

kit-ty-kate: Or maybe not, it seems installed in /usr/sbin on debian which might be inaccessible for non-privileged users

packages/conf-haveged/conf-havaged.1.0.0/opam
UnixJunkie: ok, thanks

kit-ty-kate:
ohh... Is it `haveged` or `havaged`? Both seem to be in the package (package name vs. homepage vs. depexts vs. debian package name). Could you clear that out?

UnixJunkie:
Sorry, there was a typo in many places effectively.

kit-ty-kate:
Thanks!

PR #15468, opened 2019-12-06
core_kernel.v0.13.0 is not compatible with OCaml 4.10

xclerc:
(The incompatibility cannot be observed right now because
`core_kernel` depends on `base` and we currently have no
`base >= v0.13 & < v0.14` compatible with OCaml 4.10.)


kit-ty-kate:
Thanks!

Issue #15371, opened 2019-11-23
coqide and gnome-icon-theme3


nobrowser:
After any `opam update` whatsoever, even after `opam update --development`, `opam upgrade` tells me that package `conf-gnome-icon-theme3` has 
"changed upstream" and, since `coqide` depends on it, I must rebuild `coqide`.  This is getting old, even though there is no error (the rebuild of `coqide` succeeds).

Here is the output of `opam config report`:

```
 matica!115 ~$ opam config report
# opam config report
# opam-version      2.0.4 
# self-upgrade      no
# system            arch=x86_64 os=linux os-distribution=gentoo os-version=2.6
# solver            builtin-mccs+glpk
# install-criteria  -removed,-count[version-lag,request],-count[version-lag,changed],-changed
# upgrade-criteria  -removed,-count[version-lag,solution],-new
# jobs              3
# repositories      2 (http) (default repo at a177b733)
# pinned            2 (git)
# current-switch    default
```

I suspect this is related to:

```
matica!112 ~$ cat !103:$
cat .local/packages/opam/repo/default/packages/conf-gnome-icon-theme3/conf-gnome-icon-theme3.0/opam
opam-version: "2.0"
homepage: "https://github.com/GNOME/adwaita-icon-theme"
bug-reports: "https://gitlab.gnome.org/GNOME/adwaita-icon-theme/issues"
authors: "GNU Project"
license: ["LGPL-3.0-only" "CC-BY-SA-3.0"]
build: []
depexts: [
  ["gnome-icon-theme"] {os-family = "debian"}
  ["gnome-icon-theme"] {os-family = "fedora" | os-family = "rhel"}
  ["gnome-icon-theme"] {os = "macos" & os-distribution = "homebrew"}
  ["gnome-icon-theme"] {os-family = "suse"}
  ["gnome-icon-theme"] {os-distribution = "arch"}
  ["gnome-icon-theme"] {os-distribution = "alpine"}
  ["gnome-icon-theme"] {os-distribution = "nixos"}
  ["gnome-icon-theme"] {os-distribution = "ol"}
]
synopsis: "Virtual package relying on gnome-icon-theme"
description:
  "This package can only install if the gnome-icon-theme package is installed on the system."
flags: conf
```

Note the total absence of gentoo in the above.

Gentoo includes the following two packages.  I have neither of these installed and I never noticed any problems that I could attribute to that fact, including no problems with `coqide`.

```
[...]

PR #15486, opened 2019-12-08
conf-gnome-icon-theme3: support gentoo (#15371)

Blaisorblade:
Note this does not cover the report in https://github.com/ocaml/opam-repository/issues/15371#issuecomment-560139229 as I have no insight in that issue.


kit-ty-kate:
Thanks!

PR #15491, opened 2019-12-09
ppxlib.0.{9,10}.0 are incompatible with OCaml 4.10

xclerc:
(See [this CI log](http://check.ocamllabs.io/log/1574696611-eec32e8750d18eaf1eb0d416d0f03b9e0f1df40f/4.10/partial/async.v0.13.0) for instance.)

(If people want to use OCaml 4.10, https://github.com/ocaml-ppx/ppxlib/pull/109 is expected to provide a fix.)


kit-ty-kate:
Thanks!

PR #15493, opened 2019-12-09
bit string v3.1.1: restore ocaml min version, add stdlib-shims dependency

xguerin:


kit-ty-kate:
Please don't change the archive next time. This already breaks the opam build for people who don't use the cache and might break this package anyway for people who are using it already. This is really bad practice. This should have been another release (3.1.2).

PR #15497, opened 2019-12-09
Markup.ml 0.8.2 - standards-compliant HTML parser

aantron:
- Parse `<noscript>` and `<script>` fragments in context of `<body>` by default, rather than `<head>` (aantron/markup.ml#48, prompted Daniil Baturin).
- Change license to MIT (aantron/markup.ml@89bee81).

cc @dmbaturin

### DIFF
--- a/packages/markup/markup.0.8.2/opam
+++ b/packages/markup/markup.0.8.2/opam
@@ -23,10 +23,6 @@ depends: [
 ]
 # Markup.ml implicitly requires OCaml 4.02.3, as this is a contraint of Dune.
 
-pin-depends: [
-  ["bisect_ppx.git" "git+https://github.com/aantron/bisect_ppx.git"]
-]
-
 build: [
   ["dune" "build" "-p" name "-j" jobs]
 ]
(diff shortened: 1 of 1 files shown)


packages/markup/markup.0.8.2/opam
```diff
@@ -0,0 +1,57 @@
...
+  "uutf" {>= "1.0.0"}
+
+  "bisect_ppx" {dev & >= "2.0.0"}
+  "ounit" {dev}
+]
+# Markup.ml implicitly requires OCaml 4.02.3, as this is a contraint of Dune.
+
+pin-depends: [
+  ["bisect_ppx.git" "git+https://github.com/aantron/bisect_ppx.git"]
+]
```
kit-ty-kate: I don't think a released package should have a pin-depends field. Users might pin this package if they just want this version, it doesn't have to be the dev version.

dmbaturin: I also found it odd when installing a dev version for testing. Why is it needed?

aantron: I've removed the field.

kit-ty-kate:
Thanks!

PR #15506, opened 2019-12-10
Install Stog binaries

Leonidas-from-XIV:
There was a complaint on IRC stog was missing its binaries. This is due to for some reason not installing them in the opam file.

This PR just adds the call to do so via the Makefile.

### DIFF
--- a/packages/stog/stog.0.18.0/opam
+++ b/packages/stog/stog.0.18.0/opam
@@ -11,7 +11,7 @@ build: [
   ["./configure" "--prefix" prefix]
   [make "all"]
 ]
-install: [make "install-lib" "install-share"]
+install: [make "install-lib" "install-share" "install-bin"]
 remove: ["ocamlfind" "remove" "stog"]
 depends: [
   "ocaml" {>= "4.03.0"}
--- a/packages/stog/stog.0.18.0/opam
+++ b/packages/stog/stog.0.18.0/opam
@@ -12,7 +12,6 @@ build: [
   [make "all"]
 ]
 install: [make "install-lib" "install-share" "install-bin"]
-remove: ["ocamlfind" "remove" "stog"]
 depends: [
   "ocaml" {>= "4.03.0"}
   "ocamlfind"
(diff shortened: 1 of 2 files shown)


packages/stog/stog.0.18.0/opam
```diff
@@ -11,7 +11,7 @@ build: [
   ["./configure" "--prefix" prefix]
   [make "all"]
 ]
-install: [make "install-lib" "install-share"]
+install: [make "install-lib" "install-share" "install-bin"]
 remove: ["ocamlfind" "remove" "stog"]
```
mseri: Given that we are here, and this line is not necessary, would you mind removing it?
```suggestion
```

mseri:
In this case I feel it is fine to just modify the opam file and not release a new 0.18.0-1 version, but maybe it is time we become stricter about it. @kit-ty-kate what is your opinion?

Leonidas-from-XIV:
I think it would be nice if there were a document with current best practices for opam-repository, so I wouldn't waste maintainer time doing the wrong thing. In this case, I didn't know there was a decision to use `-1` suffix versions.

mseri:
@Leonidas-from-XIV we usually do it if we add patches, so it would not cover this case. Also please do nothing for the moment, I was just pinging to have also another point of view.

But yours is a good point, once I have enough time I may draft a PR with the current practices and links to relevant discussions. If we decide to become stricter we will need it anyways.

kit-ty-kate:
> I think it would be nice if there were a document with current best practices for opam-repository, so I wouldn't waste maintainer time doing the wrong thing. In this case, I didn't know there was a decision to use `-1` suffix versions.

it really depends on the fix, there is no fixed rules for this. For this specific one I'd say probably yes, a `-1` suffix is adequate. I'll try to write something up tomorrow, add it to the wiki and add a automatic message when a PR is opened in which the wiki page would be linked (or simply put it in the automatic message if it's short enough)

mseri:
Thanks to both!

PR #15511, opened 2019-12-11
added packages/cpu/cpu.2.0.0

UnixJunkie:
The previous setcore package is deprecated because it could not coinstall with parmap.
This new package uses a different module name (Cpu instead of Setcore), so that
it can coexist with parmap.


UnixJunkie:
Parany will rely on this package in the very near future.

UnixJunkie:
I must send the PR for parany.8.0.0 once this one is merged

UnixJunkie:
The CI tests are all green, for once... :)

mseri:
Thanks

PR #15522, opened 2019-12-12
Add zarith-freestanding.1.7-3 which add zarith-freestanding.cm{,x}a

dinosaure:
This new version of `zarith-freestanding` adds a patch which provides:
- `zarith-freestanding.cmxa`
- `zarith-freestanding.cma`

From a new `mirage-install.sh` script and a new patch `fix-cma.patch`. The goal is to provide a new sub-library (according `ocamlfind`/`dune`) `zarith.freestanding` which points to `zarith-freestanding.cm{,x}a` with expected flags (in our case, `-lzarith-freestanding`).

This new version should not break anything when the `linkopts_freestanding` still is here - so, MirageOS pre-dune should continue to work.


dinosaure:
This is a WIP, we should not merge now but I would like to use the best CI ever to see if my change has an impact.

PR #15527, opened 2019-12-13
Package radare2.0.0.3

XVilka:
### `radare2.0.0.3`
OCaml interface to r2
Interact with radare2,
See the mli for documentation, example usage in utop:

#require "radare2";;
let result = R2.with_command_j ~cmd:"/j chown" "/bin/ls";;
val result : Yojson.Basic.json =
`List
  [`Assoc
    [("offset", `Int 4294987375); ("id:", `Int 0);
     ("data", `String "ywritesecuritychownfile_inheritdi")]]"

---
* Homepage: https://github.com/fxfactorial/ocaml-radare2
* Source repo: git+https://github.com/fxfactorial/ocaml-radare2.git
* Bug tracker: https://github.com/fxfactorial/ocaml-radare2/issues

---

### DIFF
--- a/packages/radare2/radare2.0.0.3/opam
+++ b/packages/radare2/radare2.0.0.3/opam
@@ -1,0 +1,60 @@
+opam-version: "2.0"
+maintainer: "Edgar Aroutiounian <edgar.factorial@gmail.com>"
+authors: [ "Edgar Aroutiounian <edgar.factorial@gmail.com>" ]
+license: "BSD-3-Clause"
+homepage: "https://github.com/fxfactorial/ocaml-radare2"
+dev-repo: "git+https://github.com/fxfactorial/ocaml-radare2.git"
+bug-reports: "https://github.com/fxfactorial/ocaml-radare2/issues"
+build: [
+	["dune" "subst"] {pinned}
+	["dune" "build" "-p" name "-j" jobs]
+]
+depexts: [
+  ["radare2"] {os-family = "debian"}
+  ["radare2"] {os-distribution = "fedora"}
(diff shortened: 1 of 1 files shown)


packages/radare2/radare2.0.0.3/opam
```diff
@@ -0,0 +1,60 @@
...
+	["dune" "subst"] {pinned}
+	["dune" "build" "-p" name "-j" jobs]
+]
+depexts: [
+  ["radare2"] {os-family = "debian"}
+  ["radare2"] {os-distribution = "fedora"}
+  ["radare2"] {os-distribution = "arch"}
+  ["radare2"] {os-distribution = "gentoo"}
+  ["radare2"] {os-distribution = "homebrew" & os = "macos"}
+]
```
kit-ty-kate: If you have the time I think the depexts could be better in their own `conf-radare2` package. It's fine as is for now but it might be easier for future releases

packages/radare2/radare2.0.0.3/opam
```diff
@@ -0,0 +1,60 @@
...
+depends: [
+  "ocaml" {>= "4.03.0"}
+  "dune"
```
kit-ty-kate:
```suggestion
  "dune" {>= "1.2"}
```
According to your `dune-project` file, this package requires at least dune 1.2

packages/radare2/radare2.0.0.3/opam
XVilka: Ok, will do that next time, created issue not to forget: https://github.com/fxfactorial/ocaml-radare2/issues/7

mseri:
Thanks

PR #15534, opened 2019-12-14
[new release] crunch (3.2.0)

hannesm:
CHANGES:

* Make crunch reproducible: use a Map.Make(String) instead of Hashtbl.
  Iterating over the former is guaranteed to be sorted over the keys.
  (mirage/ocaml-crunch#51 @hannesm)
* The state (Map.t) is passed explicit to `scan_file`, `output_implementation`
  and `walk_directory_tree`. The value `empty` is provided to construct an
  empty `t`. (mirage/ocaml-crunch#51 @hannesm)


mseri:
Thanks

PR #15537, opened 2019-12-15
Package pacomb.0.1

craff:
### `pacomb.0.1`
Parsing library based on combinators and ppx extension to write languages
Pacomb is a parsing library that compiles grammars to combinators prior to
parsing together with a PPX extension to write parsers inside OCaml files.

The advantages of Pacomb are

- Grammars as first class values defined in your OCaml files. This is
  an example from the distribution:

(* The three levels of priorities *)
type p = Atom | Prod | Sum
let%parser rec
              (* This includes each priority level in the next one *)
     expr p = Atom < Prod < Sum
            (* all other rule are selected by their priority level *)
            ; (p=Atom) (x::FLOAT)                        => x
            ; (p=Atom) '(' (e::expr Sum) ')'             => e
            ; (p=Prod) (x::expr Prod) '*' (y::expr Atom) => x*.y
            ; (p=Prod) (x::expr Prod) '/' (y::expr Atom) => x/.y
            ; (p=Sum ) (x::expr Sum ) '+' (y::expr Prod) => x+.y
            ; (p=Sum ) (x::expr Sum ) '-' (y::expr Prod) => x-.y

- Good performances:
  - on non ambiguous grammars, 2 to 3 time slower compared to ocamlyacc
  - on ambiguous grammars O(N^3 ln(N)) can be achieved.

- Parsing from left to right (despite the use of combinators) allowing not
  to keep the whole input in memory and allowing to parse streams.

- Dependant sequence allowing for self extensible grammars (like new infix
  with a given priority in a given example).

- Managing of blanks that for instance allows for nested language using
  different kind of comments or blanks.

- Support for cache and merge for ambiguous grammars (to get O(N^3 ln(N)))

- Enough support for utf8 to write parser for a language using utf8.

- Comes with documentation and various examples illustrating most possibilities.

All this makes Pacomb a promising solution to write languages in OCaml.

---
* Homepage: https://github.com/craff/pacomb
* Source repo: git+https://github.com/craff/pacomb.git
* Bug tracker: https://github.com/craff/pacomb/issues

---

### DIFF
--- a/packages/pacomb/pacomb.0.1/opam
+++ b/packages/pacomb/pacomb.0.1/opam
@@ -1,0 +1,64 @@
+synopsis: "Parsing library based on combinators and ppx extension to write languages"
+description:
+"""
+Pacomb is a parsing library that compiles grammars to combinators prior to
+parsing together with a PPX extension to write parsers inside OCaml files.
+
+The advantages of Pacomb are
+
+- Grammars as first class values defined in your OCaml files. This is
+  an example from the distribution:
+
+(* The three levels of priorities *)
+type p = Atom | Prod | Sum
+let%parser rec
+              (* This includes each priority level in the next one *)
+     expr p = Atom < Prod < Sum
+            (* all other rule are selected by their priority level *)
+            ; (p=Atom) (x::FLOAT)                        => x
+            ; (p=Atom) '(' (e::expr Sum) ')'             => e
+            ; (p=Prod) (x::expr Prod) '*' (y::expr Atom) => x*.y
+            ; (p=Prod) (x::expr Prod) '/' (y::expr Atom) => x/.y
+            ; (p=Sum ) (x::expr Sum ) '+' (y::expr Prod) => x+.y
+            ; (p=Sum ) (x::expr Sum ) '-' (y::expr Prod) => x-.y
+
+- Good performances:
+  - on non ambiguous grammars, 2 to 3 time slower compared to ocamlyacc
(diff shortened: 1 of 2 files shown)


mseri:
The `url` field is missing. This often happens if the tag for the submission is not pushed to the repository before running opam publish

packages/pacomb/pacomb.0.1/opam
```diff
@@ -0,0 +1,64 @@
...
+homepage: "https://github.com/craff/pacomb"
+dev-repo: "git+https://github.com/craff/pacomb.git"
+authors: [
+  "Christophe Raffalli <christophe@raffalli.eu>"
+  "Rodolphe Lepigre <lepigre@mpi-sws.org>" ]
+license: "MIT"
+
+depends: [
+  "ocaml" { >= "4.04.1" }
+  "dune"  { build & >= "1.2.0" }
```
mseri:
```suggestion
  "dune"  {>= "1.9.0" }
```
According to your dune-project file this should be 1.9. Also dune cannot be tagged as buil dependency due to forward compatibility reasons

craff:
Merci mseri pour les commentaires. J'espère avoir fait ce qu'il faut.
opam-publish devrait échouer non si il n'y a pas de section "url" ?

kit-ty-kate:
@craff some meta-packages don't have the url section, but in the general case yes probably. If you want you can write an issue for opam-publish here: https://github.com/ocaml/opam-publish/issues

kit-ty-kate:
thanks!

PR #15540, opened 2019-12-15
Disable tests of owork 0.1.0

aantron:
Its expect tests seem to [fail](https://ci.ocaml.org/log/saved/docker-run-dfeeeb949bdf125e9e6fdfced20165dd/db24badb7d1fb1b2970b6a505012eb7b53fc3cbd) due to a difference in formatting. There is a newer version of owork (0.1.1) in opam whose tests don't fail in this way.


mseri:
Thanks

PR #15541, opened 2019-12-15
Lwt 5.0.0 - promises and concurrent I/O; Lwt_ppx 2.0.0

aantron:
From the [changelog](https://github.com/ocsigen/lwt/releases/tag/5.0.0):

> Breaking
>
> *See ocsigen/lwt#584 for an extended summary and discussion of this release as a whole, or individual issues for each change specifically.*
>
> - The callback passed to [`Lwt.async`](http://ocsigen.org/lwt/dev/api/Lwt#VALasync) must now evaluate to `unit Lwt.t`, rather than `_ Lwt.t` (ocsigen/lwt#603, requested @cfcs).
> - [`Lwt.choose`](http://ocsigen.org/lwt/dev/api/Lwt#VALchoose), [`Lwt.nchoose`](http://ocsigen.org/lwt/dev/api/Lwt#VALnchoose), [`Lwt.nchoose_split`](http://ocsigen.org/lwt/dev/api/Lwt#VALnchoose_split), [`Lwt.pick`](http://ocsigen.org/lwt/dev/api/Lwt#VALpick), and [`Lwt.npick`](http://ocsigen.org/lwt/dev/api/Lwt#VALnpick) now raise `Invalid_argument` if given the empty list (ocsigen/lwt#557, Tim Reinke).
> - Catch nested calls to [`Lwt_main.run`](http://ocsigen.org/lwt/dev/api/Lwt_main#VALrun) (ocsigen/lwt#607, ocsigen/lwt#609, prompted François-René Rideau).
> - Use the new [`Lwt_unix.IO_vectors`](http://ocsigen.org/lwt/dev/api/Lwt_unix.IO_vectors) in [`Lwt_unix.recv_msg`](http://ocsigen.org/lwt/dev/api/Lwt_unix#VALrecv_msg) and [`Lwt_unix.send_msg`](http://ocsigen.org/lwt/dev/api/Lwt_unix#VALsend_msg) (ocsigen/lwt#594, prompted Marcello Seri).
> - Make [`Lwt_unix.Async_switch`](http://ocsigen.org/lwt/dev/api/Lwt_unix#TYPEasync_method) a synonym for `Lwt_unix.Async_detach` (ocsigen/lwt#572).
> - Remove the redundant `configure.ml` (ocsigen/lwt#700).
> - PPX: remove support for general `[%lwt ...]` expressions (ocsigen/lwt#527).
> - PPX: remove support for `Lwt_log` and the `-log` option (ocsigen/lwt#520).
[...]

PR #15545, opened 2019-12-16
Constraints for Lwt 5.0.0

aantron:
Release PR is https://github.com/ocaml/opam-repository/pull/15541.


aantron:
Details here: https://github.com/ocaml/opam-repository/pull/15541#issuecomment-566065188

talex5:
I'm not sure just removing all the failing tests is a good idea. How do we know these packages work correctly with the new Lwt if we're no longer testing them?

hannesm:
I agree with @talex5 and am in favour for packages I maintain to use an upper bound for lwt `< 5.0.0`, and fixing in upcoming releases, than to remove the tests.

aantron:
I agree that removing the tests is not optimal, but I am also hesitant about adding a constraint. The fact that it was the *tests* that failed indicates that the packages themselves compiled against Lwt, and given the conservative nature of breakage, there is plenty of reason to think that if they compiled, they still work if and only if they worked before. That would make new constraints excessive, and could be a burden on transitive users of the packages, for whom the *tests* are not in the dependency graph.

I don't know what the right thing to do here is, or how to properly express the constraints. Is it possible to add a partially redundant `"lwt" {with-test & < "5.0.0"}`?

hannesm:
@aantron I think adding `"lwt" {with-test & < "5.0.0"}` should work and would be fine with me. thanks.

talex5:
> The fact that it was the tests that failed to compile indicates that the packages themselves compiled against Lwt (and in some cases I verified this), and given the conservative nature of the breakage, there is plenty of reason to think that if they compiled, they still work if and only if they worked before.

But once the packages have been stripped of their tests, they are vulnerable to bugs in *any* new versions of their upstream packages, not just Lwt 5.0.0.

In any case, mirage-logs should be removed from this PR (I assume it works with the updated Lwt release).

aantron:
Ok, I updated the patch not to remove any tests, but only to add `with-test & < "5.0.0"` constraints or `with-test & >= "5.0.0"` conflicts for packages whose tests got broken. I tested that this kind of constraint works locally with a local opam repo.

`mirage-logs` is excluded from the patch. It will work with the retagged Lwt 5.0.0; I checked this locally. Thanks again @talex5. I also checked tests of `noise`, the other package whose tests triggered an exception due to nested `Lwt_main.run`. These tests still trigger that exception with the patched Lwt 5.0.0, so `noise` tests get a constraint.

kit-ty-kate:
Thanks a lot!

aantron:
Thanks! I was just checking the CI failures, but none of them seem related to these constraints, nor would they be, as the constraints should have no material effect on the repo at the moment.

I will fix the build of `markup-lwt` 0.5.0 against `markup` 0.7.5, one of the packages that failed, since I maintain that. It probably just needs a constraint, because `markup` 0.7.5 still included the code of `markup-lwt`, and the two were never meant to be installed together. I'll send a PR for that later.

PR #15546, opened 2019-12-16
Update checksum of brotli 1.3.0

aantron:
GitHub repacked the archive when the source repo name changed.

### DIFF
--- a/packages/brotli/brotli.1.3.0/opam
+++ b/packages/brotli/brotli.1.3.0/opam
@@ -32,5 +32,5 @@ http://www.ietf.org/id/draft-alakuijala-brotli"""
 flags: light-uninstall
 url {
   src: "https://github.com/fxfactorial/ocaml-brotli/archive/v1.3.0.tar.gz"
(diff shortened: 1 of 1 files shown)


aantron:
Example log with failure: https://ci.ocaml.org/log/saved/docker-run-72faa59f566a5f862d2e338233a369ec/7d50df61a7a53226106adae3d0633f798ba4e869.

The repo is now at https://github.com/fxfactorial/reasonml-brotli.

aantron:
Now the package is retrieved, but [fails to build](https://ci.ocaml.org/log/saved/docker-build-86a9d5775a255ee55cdf4b3f1589a248/2d75ba98c7c171a39b5071b4e7f19c4c3d466d15) because of a compilation error in its C stubs, so I think this PR is ready for merge.

kit-ty-kate:
I checked using `diff -r` and the two archives are exactly the same. Good to merge. Thanks!

aantron:
Thanks for the extra check! I hadn't done that.

PR #15549, opened 2019-12-16
Package cryptoverif.2.02

bruno-blanchet:
### `cryptoverif.2.02`
CryptoVerif: Cryptographic protocol verifier in the computational model
CryptoVerif is an automatic protocol prover sound in the computational model. It can prove

  - secrecy;
  - correspondences, which include in particular authentication;
  - indistinguishability between two games.

It provides a generic mechanism for specifying the security assumptions on cryptographic primitives, which can handle in particular symmetric encryption, message authentication codes, public-key encryption, signatures, hash functions.

The generated proofs are proofs by sequences of games, as used by cryptographers. These proofs are valid for a number of sessions polynomial in the security parameter, in the presence of an active adversary. CryptoVerif can also evaluate the probability of success of an attack against the protocol as a function of the probability of breaking each cryptographic primitive and of the number of sessions (exact security).

This software is under development; please use it at your own risk. Comments and bug reports welcome.

---
* Homepage: https://cryptoverif.inria.fr
* Bug tracker: Bruno.Blanchet@inria.fr

---

### DIFF
--- a/packages/cryptoverif/cryptoverif.2.02/opam
+++ b/packages/cryptoverif/cryptoverif.2.02/opam
@@ -1,0 +1,33 @@
+opam-version: "2.0"
+synopsis: "CryptoVerif: Cryptographic protocol verifier in the computational model"
+description: """
+CryptoVerif is an automatic protocol prover sound in the computational model. It can prove
+
+  - secrecy;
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Thanks!

PR #15552, opened 2019-12-16
[new release] arp and arp-mirage (2.2.1)

hannesm:
CHANGES:

* adapt to lwt 5.0.0 change (mirage/arp#18 @hannesm)


avsm:
tests failing on macOS, but this appears to not be a regression in this particular release

PR #15555, opened 2019-12-17
[new release] dune-build-info, dune, dune-configurator, dune-action-plugin, dune-private-libs and dune-glob (2.0.1)

rgrinberg:
Embed build informations inside executable

##### CHANGES:

- Delay errors raised by invalid `dune-package` files. The error is now raised
  only if the invalid package is treated as a library and used to build
  something. (ocaml/dune#2972, @rgrinberg)

### DIFF
--- a/packages/dune-private-libs/dune-private-libs.2.0.1/opam
+++ b/packages/dune-private-libs/dune-private-libs.2.0.1/opam
@@ -1,0 +1,42 @@
+opam-version: "2.0"
+synopsis: "Private libraries of Dune"
+description: """
+!!!!!!!!!!!!!!!!!!!!!!
+!!!!! DO NOT USE !!!!!
+!!!!!!!!!!!!!!!!!!!!!!
+
+This package contains code that is shared between various dune-xxx
+packages. However, it is not meant for public consumption and provides
+no stability guarantee.
+"""
+maintainer: ["Jane Street Group, LLC <opensource@janestreet.com>"]
+authors: ["Jane Street Group, LLC <opensource@janestreet.com>"]
+license: "MIT"
+homepage: "https://github.com/ocaml/dune"
+doc: "https://dune.readthedocs.io/"
+bug-reports: "https://github.com/ocaml/dune/issues"
+depends: [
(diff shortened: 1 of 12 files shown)


packages/dune-private-libs/dune-private-libs.2.0.1/opam
```diff
@@ -0,0 +1,42 @@
...
+"""
+maintainer: ["Jane Street Group, LLC <opensource@janestreet.com>"]
+authors: ["Jane Street Group, LLC <opensource@janestreet.com>"]
+license: "MIT"
+homepage: "https://github.com/ocaml/dune"
+doc: "https://dune.readthedocs.io/"
+bug-reports: "https://github.com/ocaml/dune/issues"
+depends: [
+  "dune" {>= "2.0"}
+  "ocaml" {>= "4.07"}
```
kit-ty-kate:
```suggestion
  "ocaml" {>= "4.06"}
```
I believe this should be 4.06, otherwise when dune.2.0.0 becomes unavailable, people on 4.06 won't be able to install dune2 anymore if they have packages relying on dune-configurator for instance, and they will revert to dune.1.11.4

kit-ty-kate:
@thomasblanc Camelus seems to have given up on several recent PRs, could you look into it?

thomasblanc:
@kit-ty-kate it seems there is a fd leak in Lwt, I restarted Camelus and am launching a recheck right now.

kit-ty-kate:
Thanks a lot!

avsm:
I don't think dune 2.0.0 should be marked uninstallable -- this has broken a bunch of previously working projects that installed it specifically (and that didn't use the downgrade functionality).  Isn't it enough to release a new version 2.0.1 that will be picked up by those upgrading?

rgrinberg:
Hmm, this is indeed a bit unfortunate and I did not foresee it. In an attempt to fix dune for users of a broken feature (downgrading), we've broken things further for a legitimate subset of users.

it's not possible to use duniverse yet for RWO?

PR #15559, opened 2019-12-18
added packages/pardi/pardi.2.0.3

UnixJunkie:
bug correction for macs, where the output of 'wc -l some_file' doesn't have
the same format than on Linux. String stripping is needed prior to parsing
the command output.


mseri:
Thanks

PR #15562, opened 2019-12-18
restrict to core<0.13.0 for packages that use polymorphic comparison

avsm:
found in revdeps for cohttp release #15561


packages/horned_worm/horned_worm.0.1.1/opam
```diff
@@ -15,6 +15,7 @@ depends: [
   "yojson" {>= "1.4.0"}
   "jbuilder" {>= "1.0+beta13"}
   "lwt" {>= "3.1.0"}
+  "core" {<"v0.13.0"}
```
kit-ty-kate:
```suggestion
```
This package does not depend on core or async

kit-ty-kate:
Thanks!

PR #15570, opened 2019-12-19
A couple more constraints for Lwt 5.0.0

aantron:
See

  https://github.com/ocaml/opam-repository/pull/15541#issuecomment-567579939.


kit-ty-kate:
Thanks!

PR #15577, opened 2019-12-20
Package binsec.0.3

rbonichon:
### `binsec.0.3`
Semantic analysis of binary executables
BINSEC aims at developing an open-source platform filling the gap between formal
methods over executable code and binary-level security analyses currently used
in the security industry.

The project targets the following applicative domains:

    vulnerability analyses
    malware comprehension
    code protection
    binary-level verification

BINSEC is developed at CEA List in scientfic collaboration with Verimag and LORIA.

An overview of some BINSEC features can be found in our SSPREW'17 tutorial.

---
* Homepage: https://binsec.github.io
* Bug tracker: https://github.com/binsec/binsec/issues

---

### DIFF
--- a/packages/binsec/binsec.0.3/opam
+++ b/packages/binsec/binsec.0.3/opam
@@ -1,0 +1,93 @@
+opam-version: "2.0"
+synopsis: "Semantic analysis of binary executables"
+description: "
+BINSEC aims at developing an open-source platform filling the gap between formal
+methods over executable code and binary-level security analyses currently used
+in the security industry.
+
+The project targets the following applicative domains:
+
+    vulnerability analyses
+    malware comprehension
+    code protection
+    binary-level verification
+
+BINSEC is developed at CEA List in scientfic collaboration with Verimag and LORIA.
+
+An overview of some BINSEC features can be found in our SSPREW'17 tutorial."
+maintainer: "BINSEC <binsec@saxifrage.saclay.cea.fr>"
+authors: [
+"Adel Djoudi"
+"Benjamin Farinier"
+"Frédéric Recoules"
+"Josselin Feist"
+"Lesly-Ann Daniel"
+"Manh-Dung Nguyen"
+"Mathilde Ollivier"
+"Matthieu Lemerre"
+"Olivier Nicole"
+"Richard Bonichon"
+"Robin David"
+"Ta Thanh Dinh"
+"Yaëlle Vinçont"
+]
+homepage: "https://binsec.github.io"
+license: "GNU Lesser General Public License version 2.1"
+doc: ["http://binsec.github.io/apiref/index.html"]
+bug-reports: "https://github.com/binsec/binsec/issues"
+tags: [
+  "binary code analysis"
+  "symbolic execution"
+  "deductive"
+  "program verification"
+  "formal specification"
+  "automated theorem prover"
+  "plugins"
+  "abstract interpretation"
+  "dataflow analysis"
+  "linking"
+  "disassembly"
+]
+
+build: [
+ ["autoconf"] {pinned}
+  ["./configure" "--prefix" prefix]
+  [make "-C" "src" "depend"]
+  [make "-C" "src" "-j%{jobs}%"]
+]
+
+install: [
+  [make "-C" "src" "install"]
+]
+
+depends: [
+  "ocaml" { >= "4.04.2" & <= "4.07.1" }
+  "ocamlfind" {build}
+  "menhir"
+  "ocamlgraph" { >= "1.8.5" & < "1.9~" }
+  "piqi"
+  "piqilib"
+  "zarith"
+  "zmq"
+  "llvm"
+]
+
+depexts: [
+  [ "z3" "protobuf-compiler" ] { os-family = "debian" }
+  [ "z3" "protobuf-compiler" ] { os-family = "fedora" }
+  [ "z3" "protobuf" ] { os-family = "archlinux" }
+  [ "z3" "protobuf" ] { os-family = "openbsd" }
+]
+
+messages: [
+  "BINSEC uses external automatic solvers, like boolector, z3, cvc4, or yices
+  for some functionalities. Only Z3 is listed in depexts.
+  install the others through your package manager."
+]
+url {
+  src: "https://github.com/binsec/binsec/archive/binsec-0.3.tar.gz"
+  checksum: [
+    "md5=337e588fe661ea007dd4da06d59dbda1"
+    "sha512=9283cf01564bc860fdab87aae2f73c8c8cc9a95110ec6c3cfaa981cab26b46ea2f35f093375427f072bd216e0341e2dffa3213ca3c0a0dd51cbc816561285d48"
+  ]
+}


rbonichon:
Hi @camelus, indeed this is. 

I'm not sure to understand how `ci/datakit/3 Compilers` fails.

mseri:
None of the tested compilers is tested because the `llvm-9-dev` depext is not on the system

mseri:
In practice the CI is not testing the installation of your package at all, because those failing ones are the only 4 containers with a compatible compiler

rbonichon:
@mseri Thanks for the explanation. 

I know that our tool builds on `debian:stable` for we build a Docker container on top of the distribution's image. The container is built with a fixed dependency on `llvm-6.0` (Debian) / `llvm=6.0.0` (opam). 

Is there a canonical way to proceed from the current state?

packages/binsec/binsec.0.3/opam
```diff
@@ -0,0 +1,78 @@
...
+  [make "-C" "src" "-j%{jobs}%"]
+]
+
+install: [
+  [make "-C" "src" "install"]
+]
+
+depends: [
+  "ocaml" { >= "4.04.2" & <= "4.07.1" }
+  "ocamlfind"
```
mseri: Is this just a build dependency or is it used at runtime?

mseri:
If it is supposed to work with all the llvm dependency versions, I would avoid restricting the version and merge this in good faith as it cannot be tested. Otherwise you can add a bound to the version of llvm

packages/binsec/binsec.0.3/opam
rbonichon: Both `ocamlfind` and `ocaml` are build dependencies.

packages/binsec/binsec.0.3/opam
```diff
@@ -0,0 +1,78 @@
...
+  "piqi"
+  "piqilib"
+  "zarith"
+  "zmq"
+  "llvm"
+]
+
+depexts: [
+  "make"
+]
```
kit-ty-kate: This is already a dependency of opam. Users will get a warning if make is not installed. Furthermore `make` is not the `make` package everywhere (e.g. macos). Could you remove this field?

packages/binsec/binsec.0.3/opam
```diff
@@ -0,0 +1,78 @@
...
+  [make "-C" "src" "depend"]
+  [make "-C" "src" "-j%{jobs}%"]
+]
+
+install: [
+  [make "-C" "src" "install"]
+]
+
+depends: [
+  "ocaml" { >= "4.04.2" & <= "4.07.1" }
```
kit-ty-kate:
```suggestion
  "ocaml" { >= "4.04.2" & < "4.08" }
```

kit-ty-kate:
In practice your package will be tested by [check.ocamllabs.io](http://check.ocamllabs.io) as it uses debian-unstable instead of debian-stable. But this will take a while as it takes a fairly long time to go through all the packages. In the meantime I'm testing locally your package using a docker container on debian-unstable. I'll get back to you once this is done

kit-ty-kate:
There is an error:
```
#=== ERROR while compiling binsec.0.3 =========================================#
"./configure": command not found.
```

packages/binsec/binsec.0.3/opam
rbonichon: I think `<= 4.07.1` is better since the `configure` script will reject `4.07.2`

rbonichon:
@kit-ty-kate The build section in the opam file starts with `autoconf`. Was it run ?

kit-ty-kate:
This seems to come from the fact that no archive has been defined thus opam doesn't know where to get it. Your tag seems to be there so something went wrong in the way opam-publish tried to get the tag. Maybe the `origin` git remote is not the github one or something like that

kit-ty-kate:
(oops my bad)

packages/binsec/binsec.0.3/opam
rbonichon: The actual dependency is on GNU make indeed. Will remove it.

rbonichon:
> This seems to come from the fact that no archive has been defined thus opam doesn't know where to get it. Your tag seems to be there so something went wrong in the way opam-publish tried to get the tag. Maybe the `origin` git remote is not the github one or something like that

Is this comment still valid ? I am a little lost in our interaction :thinking:

kit-ty-kate:
> @kit-ty-kate Is this comment still valid ? I am a little lost in our interaction thinking

Yes, I was just saying sorry about misclicking on the button to close your PR

rbonichon:
Indeed our github repo acts as a public secondary repo and the `origin` for me is an internal repo. If I read the doc correctly, it should be enough to re-run `opam-publish` with the rectified info, right?

packages/binsec/binsec.0.3/opam
kit-ty-kate: oh, in this case, fair.

kit-ty-kate:
It now starts the compilation but fails with:
```
#=== ERROR while compiling binsec.0.3 =========================================#
# context     2.0.5 | linux/x86_64 | ocaml-base-compiler.4.07.1 | file:///home/opam/opam-repository
# path        ~/.opam/4.07.1/.opam-switch/build/binsec.0.3
# command     /usr/bin/make -C src depend
# exit-code   2
# env-file    ~/.opam/log/binsec-90-71a8c9.env
# output-file ~/.opam/log/binsec-90-71a8c9.out
### output ###
# make: Entering directory '/home/opam/.opam/4.07.1/.opam-switch/build/binsec.0.3/src'
# PIQI piqi/common.piqi
# sh: 1: protoc: not found
# error: command execution failed: protoc -Iproto --include_imports --descriptor_set_out=proto/common.proto.pb proto/common.proto
# make: *** [../Piqi.mk:31: piqi/common.piqi] Error 1
# make: Leaving directory '/home/opam/.opam/4.07.1/.opam-switch/build/binsec.0.3/src'
```
There seems to be a missing dependency on some external binary called `protoc`

rbonichon:
Interesting. `protoc` is actually invoked by `piqi` here I guess. I'm not sure whose responsibility it is to expose this dependency.

rbonichon:
FWIW, I can confirm this comes from the `piqi` package. In `src/of_proto.ml`, the function below invokes `protoc` (which is not declared as a dependency of `piqi`)

```ocaml
let proto_to_wire ifile ofile =
  (* build -I parameters for protoc *)
  let paths = List.rev !paths in
  let includes =
    String.concat "" (List.map (fun x -> "-I" ^ x ^ " ") paths)
  in
  let cmd = Printf.sprintf
    "protoc %s--include_imports --descriptor_set_out=%s %s" includes ofile ifile
  in
  let res = Sys.command cmd in
  if res <> 0
  then piqi_error ("error: command execution failed: " ^ cmd)
```

On a Debian-like distribution, one needs to install `libprotobuf-dev libprotoc-dev protobuf-compiler`

kit-ty-kate:
cc @alavrik?

rbonichon:
@kit-ty-kate @alavrik I'd like to propose to contribute a patch the current `piqi` package to include a runtime dependency to `protoc`.

alavrik:
@rbonichon @kit-ty-kate piqi could be used without protoc installed. This is an optional dependency. You only need to install it if you need to work with .proto files.

rbonichon:
@alavrik My (limited) understanding is that there currently is no mode where you can somehow "protect" the parts that need `protoc`, e.g., through a configure script or a dedicated check when `protoc` is about to be called. *If* this is true, the safe overapproximation would be to make it a dependency (in case the user activates this part), would you agree? 

@kit-ty-kate In any case, I could make `protoc` a dependency for `binsec`

alavrik:
@rbonichon there is some logic along these lines in the test runner script: https://github.com/alavrik/piqi/blob/master/tests/Makefile#L7

My worry is that including it as a dependency would lead to bad experience for opam users and for me as a maintainer. Opam has an impressive platform coverage in its CI matrix. The unfortunate downside of this is things breaking in unpredictable way on obscure platforms during CI runs. I don't want to risk CI instability by adding protoc/libprotobuf which is a pretty big dependency.

rbonichon:
@kit-ty-kate How do you propose we proceed? Is adding `protoc` to runtime dependencies a good enough solution?

kit-ty-kate:
as long as it compiles, sure.

rbonichon:
I've updated depexts to reflect the `protoc` dependency above (and also `z3` at the same time)

kit-ty-kate:
I tested this package locally on debian-sid and it compiles fine out-of-the-box. Thanks a lot!

rbonichon:
Happy to see the package merged in. Thanks to everyone for the feedback and your time.

PR #15578, opened 2019-12-20
0install: add upper bound on Lwt < 5.0.0

talex5:
`0install search` doesn't work with 5.0.0. Now sure why the revdeps check didn't find this, as the unit-tests fail.


mseri:
Thanks

PR #15579, opened 2019-12-20
obus: add upper bounds on Lwt < 5.0.0

talex5:
Fails to install. Error is:

    # ocamlfind: [WARNING] Package `threads': Linking problems may arise because of the missing -thread or -vmthread switch
    # File "_none_", line 1:
    # Error: Required module `Mutex' is unavailable

/cc @Freyr666 @aantron

### DIFF
--- a/packages/obus/obus.1.1.8/opam
+++ b/packages/obus/obus.1.1.8/opam
@@ -11,7 +11,7 @@ remove: [["ocamlfind" "remove" "obus"]]
 depends: [
   "ocaml" {>= "4.02.3"}
   "ocamlfind" {build}
-  "lwt" {>= "2.7.0"}
+  "lwt" {>= "2.7.0" & < "5.0.0"}
   "lwt_react"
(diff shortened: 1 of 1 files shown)


mseri:
Thanks

Freyr666:
@talex5 I've pushed a [branch which should fix lwt.5.0.0](https://github.com/ocaml-community/obus/tree/lwt5_fixes). Would you mind to pin it and try, before I'll merge it tomorrow?

Also it works only with lwt.5.0.0+, so I had to update the deps unfortunately.

aantron:
@Freyr666 FYI you could make the branch require only Lwt >= 4.3.0 by using this API: https://github.com/ocsigen/lwt/blob/f38996ef37d19fd851c2931e27f116ce016fec8c/src/unix/lwt_unix.cppo.mli#L1496-L1504

talex5:
@Freyr666 : Yes, that branch works for me!

PR #15584, opened 2019-12-20
[new release] ANSITerminal (0.8.2)

Chris00:
CHANGES:

- Do not check the terminal at startup for the Windows version — so it
  can be used in a continuous integration environment.


kit-ty-kate:
Thanks!

PR #15588, opened 2019-12-20
remove orphaned files (not used by opam):

hannesm:
weberizer.0.7.2 refers to weberizer.install, which is present in weberizer.0.7.2/files/weberizer.install, the weberizer.0.7.2 /weberizer.install removed in this PR is unused (same content as the other copy)
cloudi.1.7.3/cloudi.install is not referenced at all


mseri:
Thanks

PR #15599, opened 2019-12-23
[new release] obus (1.2.1)

Freyr666:


packages/obus/obus.1.2.1/opam
```diff
@@ -0,0 +1,30 @@
...
+bug-reports: "https://github.com/ocaml-community/obus/issues"
+dev-repo: "git+https://github.com/ocaml-community/obus.git"
+license: "BSD-3-Clause"
+
+build: [
+  [ "dune" "build" "-p" name "-j" jobs ]
+]
+
+depends: [
+  "dune"
```
mseri:
```suggestion
  "dune" {>= "1.1"}
```
I am not sure if `lang menhir 2.0` requires a higher lower bound though

Freyr666: Do the previous releases need this too?

kit-ty-kate: No, previous releases did not use dune.

kit-ty-kate:
```suggestion
  "dune" {>= "1.4"}
```
Apparently `(using merhir 2.0)` requires dune 1.4 (see https://github.com/ocaml/dune/pull/863). Dune is not able to check which plugin require which version, see https://github.com/ocaml/dune/issues/2957

Freyr666: 1.2.0 released used dune as well, this release contains only trivial fixes for ppxlib and lwt compatibility. I'll add this constraint to both 1.2.0 and 1.2.1 then.

mseri:
Thanks!

PR #15601, opened 2019-12-25
trie 1.0.0

kandu:
There is nothing to say about such a common and simple data structure.

Changes:
  * port to dune
  * remove all external dependencies


kit-ty-kate:
Thanks

PR #15616, opened 2019-12-31
Package zed.2.0.4

kandu:
# zed
Abstract engine for text edition in OCaml

2.0.4 (2019-12-31)
------------------

* add wanted\_column support for wide width character
* Zed\_lines: `get_idx_by_width set row column_width` return the offset of the character at `[row, column_width]`


mseri:
Thanks

PR #15619, opened 2019-12-31
Mark various package as incompatible with OCaml 4.10

kit-ty-kate:
* `ppxlib` is already fixed
* `links` cc @SimonJF links is turning `warnings as errors` during releases. That should be avoided.


mseri:
Thanks

PR #15621, opened 2020-01-01
mirage-solo5 is not compatible with OCaml 4.10

kit-ty-kate:
Fix sent upstream in https://github.com/mirage/mirage-solo5/pull/57


mseri:
Thanks

PR #15626, opened 2020-01-03
[new release] minicaml (0.3.2)

0x0f0f0f:
A simple, didactical, purely functional programming language

- Project page: <a href="https://github.com/0x0f0f0f/minicaml">https://github.com/0x0f0f0f/minicaml</a>
- Documentation: <a href="https://github.com/0x0f0f0f/minicaml">https://github.com/0x0f0f0f/minicaml</a>

### Added 
- Reworking of language internals.
- More primitives and operators.

### DIFF
--- a/packages/minicaml/minicaml.0.3.2/opam
+++ b/packages/minicaml/minicaml.0.3.2/opam
@@ -1,0 +1,33 @@
+opam-version: "2.0"
+maintainer: "sudo-woodo3@protonmail.com"
+authors: ["Alessandro Cheli"]
+homepage: "https://github.com/0x0f0f0f/minicaml"
+bug-reports: "https://github.com/0x0f0f0f/minicaml/issues"
+dev-repo: "git+https://github.com/0x0f0f0f/minicaml.git"
+license: "MIT"
+synopsis: "A simple, didactical, purely functional programming language"
+description: "A simple, didactical, purely functional programming language written for the programming 2 course at the University of Pisa, extended with a simple parser made with Menhir and ocamllex"
+doc: "https://github.com/0x0f0f0f/minicaml"
+
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+  [make "test" "-j" jobs] {with-test}
+]
+depends: [
+    "dune" {>= "2.0"}
+    "ocaml" {>= "4.05.0"}
+    "ANSITerminal"
+    "menhir"
+    "ppx_deriving"
+    "cmdliner"
+    "alcotest" {with-test & >= "0.8.5"}
+    "bisect_ppx" {>= "1.4.1"}
+]
+url {
+  src:
+    "https://github.com/0x0f0f0f/minicaml/releases/download/0.3.2/minicaml-0.3.2.tbz"
+  checksum: [
+    "sha256=40e9ddad20d728339c8581d038fb26aec3b4f4310f06ccbafccf12d47aa11836"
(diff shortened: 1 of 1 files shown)


packages/minicaml/minicaml.0.3.2/opam
```diff
@@ -0,0 +1,32 @@
...
+authors: ["Alessandro Cheli"]
+homepage: "https://github.com/0x0f0f0f/minicaml"
+bug-reports: "https://github.com/0x0f0f0f/minicaml/issues"
+dev-repo: "git+https://github.com/0x0f0f0f/minicaml.git"
+license: "MIT"
+synopsis: "A simple, didactical, purely functional programming language"
+description: "A simple, didactical, purely functional programming language written for the programming 2 course at the University of Pisa, extended with a simple parser made with Menhir and ocamllex"
+doc: "https://github.com/0x0f0f0f/minicaml"
+build: [
+  [make "-j" jobs]
```
kit-ty-kate:
```suggestion
  ["dune" "build" "-p" name "-j" jobs]
```

kit-ty-kate: This command should be used in most cases for releases. The commands you are calling in your Makefile are failing on MacOS and also might break in the future (`dune` without `-p` or `--profile=release` enables some warnings as errors by default)

0x0f0f0f: I'm using sed in the Makefile to disable the bisect_ppx preprocessor when not running `make test`. I've seen there was a mess in Github issue discussion related to bisect_ppx usage with dune. Would you suggest any way of doing so by using only dune and without having to edit the file before build? So I may get rid of calling the makefile from the opam file. Thanks a lot.

kit-ty-kate: I was saying that because I tried running your make command and it did not change the dune file in any way. The sed command does nothing to the file and `bisect_ppx` is not used since it is already commented out behind the `;` (https://dune.readthedocs.io/en/stable/lexical-conventions.html#comments)

kit-ty-kate: >  Would you suggest any way of doing so by using only dune and without having to edit the file before build?

Yes, also I think that would be cleaner and better for you and everyone else.

0x0f0f0f: `bisect_ppx` is commented out in the dune file when running the default make target. it gets uncommented by sed when running `make test`. dune's documentation is quite messy and there is nothing written in there about enabling certain s-expressions only when tests are enabled (or how to emulate that using profiles). Please let me know if you know any workaround.  

https://github.com/ocaml/dune/issues/57

kit-ty-kate: From the documentation, using `bisect_ppx --conditional` should be enough: https://github.com/aantron/bisect-starter-dune/blob/master/dune#L4

packages/minicaml/minicaml.0.3.2/opam
```diff
@@ -0,0 +1,33 @@
...
+]
+depends: [
+    "dune" {>= "2.0"}
+    "ocaml" {>= "4.05.0"}
+    "ANSITerminal"
+    "menhir"
+    "ppx_deriving"
+    "cmdliner"
+    "alcotest" {with-test & >= "0.8.5"}
+    "bisect_ppx" {>= "1.4.1"}
```
kit-ty-kate:
```suggestion
    "bisect_ppx" {with-test & >= "1.4.1"}
```
I don't understand, why is this not a test dependency anymore? `--conditional` makes it a conditional dependency (so depending on whether or not the package is installed, the ppx is going to be used or not)

0x0f0f0f:
Closing, should be solved in new release. Thank you again.

PR #15631, opened 2020-01-06
Lwt 5.1.1 - promises and concurrent I/O

aantron:
Bugs fixed

- Exception raised by [`Lwt_main.run`](https://github.com/ocsigen/lwt/blob/8d8f63423b42ddd705918ae26ba3bb41b2582a82/src/unix/lwt_main.mli#L10-L43) upon call to `exit` (ocsigen/lwt#758, reported Gal Schlezinger).


aantron:
Revdeps failures seem to be the usual ones we expect.

kit-ty-kate:
Thanks!

PR #15637, opened 2020-01-07
Constrain various packages not compatible with OCaml 4.10

kit-ty-kate:
* cc @pdonadeo for `lens`
* cc @UnixJunkie for `linwrap` (it's actually just a constraint fix but might as well)
* cc @pelzlpj for `orpie`
* cc @smondet for `ppx_deriving_cmdliner`
* cc @nberth for `mlcuddidl`
* cc @thierry-martinez for `override`

The reason for each breakage is stated in each commit messages.

### DIFF
--- a/packages/lens/lens.1.2.0/opam
+++ b/packages/lens/lens.1.2.0/opam
@@ -14,7 +14,7 @@ remove: [
   ["ocamlfind" "remove" "lens"]
 ]
 depends: [
-  "ocaml" {>= "4.02.3"}
+  "ocaml" {>= "4.02.3" & < "4.10"}
   "ocamlfind" {build}
   "ounit" {build}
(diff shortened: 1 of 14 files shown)


UnixJunkie:
thanks for the heads-up

pdonadeo:
@kit-ty-kate thanks for the reminder. Out of curiosity, do you remember what API actually changed? In this moment I can't even compile `lens` because ppx_deriving is not available in a 4.10 switch.

kit-ty-kate:
@pdonadeo I have temporary fixes and PRs for all the major packages, and an opam-overlay to regroup them all for those willing to test the beta. See here for more info: https://discuss.ocaml.org/t/ocaml-4-10-0-first-beta/4989/5

PR #15638, opened 2020-01-08
added packages/linwrap/linwrap.2.0.0

UnixJunkie:
Bug correction (out of mem) on large datasets.

There is a new dependency (dokeysto_camltc) to circumvent the bug; hence a new major release.


mseri:
Thanks

PR #15641, opened 2020-01-09
Fix various packages

kit-ty-kate:
All the packages fixed here are fixed in their latest version so no notifications to their maintainers is needed.


kit-ty-kate:
@thomasblanc it seems like Camelus crashed

PR #15639, opened 2020-01-08
[new release] dune-build-info, dune, dune-configurator, dune-action-plugin, dune-private-libs and dune-glob (2.1.2)

rgrinberg:
Embed build informations inside executable

- Project page: <a href="https://github.com/ocaml/dune">https://github.com/ocaml/dune</a>
- Documentation: <a href="https://dune.readthedocs.io/">https://dune.readthedocs.io/</a>

##### CHANGES:

- Fix a bug in the `Fiber.finalize` function of the concurrency monad of Dune,
  causing a race condition at the user level (ocaml/dune#3009, fix ocaml/dune#2958, @diml)

PR #15642, opened 2020-01-09
Make dune.2.1.1 availble to OCaml < 4.07

NathanReb:
I basically reused the constraint from dune.2.0.1 as it seems it is working fine for this version but I might have gotten it wrong.

My understanding is that without this change opam doesn't know it has to fallback to building dune with a secondary compiler if you're using ocaml < 4.07.


kit-ty-kate:
This is fine (it requires one small change in the CI first though) but note that https://github.com/ocaml/opam-repository/pull/15639 added `dune.2.1.2` so it might be worth doing it on this version too

kit-ty-kate:
Superseded by #15820

PR #15645, opened 2020-01-09
Add lower bound on Logs in Index

CraigFe:
Index relies on the optional module `Logs_threaded`, which was introduced in `Logs.0.7.0`. The corresponding PR in `mirage/index` is https://github.com/mirage/index/pull/156.


kit-ty-kate:
Thanks!

PR #15646, opened 2020-01-09
[new release] dns, dns-certify, dns-mirage, dns-resolver, dns-client, dns-server, dns-tsig and dns-cli (4.3.0)

hannesm:
CHANGES:

* dns
  - BUGFIX Name_rr_map.remove_sub remove empty maps (mirage/ocaml-dns#205, @hannesm)
* server (mirage/ocaml-dns#205, @hannesm)
  - authentication refactoring: given a key by its Domain_name.t (name._op.zone),
    this is valid for operation `op` for `zone` and subdomains thereof. The
    operation may be one of `update`, `transfer`, and `notify`, with an `update`
    key being valid for any operation, and a `transfer` key valid for
    notifications as well
  - Primary.create has a new optional argument `unauthenticated_zone_transfer`
    to allow unsigned zone transfer requests
  - the type `Authentication.a` and value `Authentication.tsig_auth` are removed
    - Primary.create and Secondary.create no longer have the `a` argument
  - authentication uniformly uses `Authentication.access`
  - handle_update / handle_axfr_request / handle_ixfr_request are provided and
    under test
  - tests for authentication and handle_question
* client (mirage/ocaml-dns#204, @hannesm)
  - introduce get_resource_record which is the same as getaddrinfo, but returns
    the error as variant instead of [ `Msg of string ]
  - BUGFIX follow_cname handles replies with a cname and no data for the alias
    appropriately (and a regression test has been developed)

### DIFF
--- a/packages/dns-certify/dns-certify.4.3.0/opam
+++ b/packages/dns-certify/dns-certify.4.3.0/opam
@@ -1,0 +1,45 @@
+opam-version: "2.0"
+maintainer: "team AT robur dot io"
+authors: ["Hannes Mehnert <hannes@mehnert.org>"]
+homepage: "https://github.com/mirage/ocaml-dns"
+doc: "https://mirage.github.io/ocaml-dns/"
+dev-repo: "git+https://github.com/mirage/ocaml-dns.git"
+bug-reports: "https://github.com/mirage/ocaml-dns/issues"
+license: "BSD2"
+
+depends: [
+  "dune" {>= "1.2.0"}
+  "ocaml" {>= "4.07.0"}
+  "dns" {= version}
+  "dns-tsig" {= version}
+  "dns-mirage" {= version}
+  "randomconv" {>= "0.1.2"}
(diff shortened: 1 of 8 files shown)


kit-ty-kate:
@diml can we do something about https://github.com/ocaml/dune/issues/2957? It is getting rather exhausting on our side to check if the dune lower-bound is correct when packages use "extensions". In this case once again (see PRs linked to the issue for more examples) the lower-bound constraint is not the right one (`1.2` vs. `1.4` where `merlin 2.0` was added).

I really enjoyed this feature from dune as an opam-repository maintainer, being able to check if the constraints are correct with one quick look is great, but as the number of people using extensions grow it is getting less and less enjoyable.

I think getting at least a warning that we can search for in the logs would be a great help to us.

kit-ty-kate:
As for this PR I'll merge it without fix as it was already broken before. Thanks!

diml:
@kit-ty-kate sure, we are going to look at this shortly!

PR #15649, opened 2020-01-10
First beta release of ocaml 4.10.0

Octachron:
This is the usual set of ocaml-variant packages for the first beta release of 4.10.0, minus the `default-unsafe-string` variants that seemed not really relevant anymore. 

cc @kit-ty-kate .


kit-ty-kate:
Sounds good, thanks a lot!

PR #15650, opened 2020-01-10
Add ocaml-src.4.10.0~beta1

kit-ty-kate:
This is a draft for now as it would be the first time ocaml-src gets a beta version and I'm not its maintainer, @hannesm what do you think? This would unlock a lot of mirage packages without the use of this opam overlay: https://github.com/kit-ty-kate/opam-alpha-repository/ But I'm not too sure about adding a beta version to the repository so maybe it's not a good idea? Should we restrict to `ocaml-variants.4.10.0+beta1` or something like that?

### DIFF
--- a/packages/ocaml-src/ocaml-src.4.10.0~beta1/files/META
+++ b/packages/ocaml-src/ocaml-src.4.10.0~beta1/files/META
@@ -1,0 +1,1 @@
+version = "4.10.0~beta1"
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
Another solution would be to add it to https://github.com/ocaml/ocaml-beta-repository only? `4.10.0+trunk/beta1` should not be accessible for people who don't have this repository installed anyway. On second thought that seems cleaner

PR #15656, opened 2020-01-11
Fix license of fstar and kremlin.

omasanori:
They are incorrectly classified as Apache-1.0+ during SPDX License Identifier normalization.

References:
https://github.com/FStarLang/FStar/blob/master/LICENSE
https://github.com/FStarLang/kremlin/blob/master/LICENSE


mseri:
Thanks

PR #15660, opened 2020-01-11
lablqml.0.6.2: Fix installation (dune install uses ocamlfind)

kit-ty-kate:
cc @FardaleM @Kakadu 

The installation fails when ocamlfind is not installed in the current switch but can be found elsewhere. This is because `dune install` used by `make install` uses ocamlfind to detect the destination.


Kakadu:
The change seems to be OK but something is wrong again with Qt5 on alpine. I was working on this but don't remember the result.

PR #15665, opened 2020-01-13
LablTk 8.06.8 and OCamlBrowser 4.10.0

garrigue:
New version of OCamlBrowser for OCaml 4.10.0.

The changes are only in OCamlBrowser, but update also LablTk since they share the same archive.


mseri:
Thanks! Looks good

PR #15667, opened 2020-01-13
LablGtk 2.18.9

garrigue:
Fixes compatibility with ocaml 4.10.


garrigue:
The problem in postgresql is already fixed, and the one in 0install seems unrelated.

mseri:
Thanks!

PR #15670, opened 2020-01-14
[new release] elpi (1.9.1)

gares:
CHANGES:

- Tests:
  - Fix testing framework on runners pre 1.9.0.

- Parser:
  - Line nubers after quotations were wrong, since `\n` inside
    quotations were not taken into account.

- Typing:
  - Name alias abbreviations are not refolded in error messages.
    Eg. `typeabbrev x int` does not take over `int`, while
    `typeabbrev x (list int)` does over `list int`.
  - Fix type abbreviation refolding, was not working on some cases.

- Stdlib:
  - Fix `is` function `int_of_string` to do what it says
  - New `is` function `rhc : string -> int` computes the inverse of `chr`

### DIFF
--- a/packages/elpi/elpi.1.9.1/opam
+++ b/packages/elpi/elpi.1.9.1/opam
@@ -1,0 +1,84 @@
+opam-version: "2.0"
+maintainer: "Enrico Tassi <enrico.tassi@inria.fr>"
+authors: [ "Claudio Sacerdoti Coen" "Enrico Tassi" ]
+license: "LGPL 2.1+"
+homepage: "https://github.com/LPCIC/elpi"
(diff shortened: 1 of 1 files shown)


gares:
I could restart some CI jobs and they succeeded, but some other seem just to fail for unrelated reasons. Any ideas?

gares:
Can someone restart the whole ci/datakit thing, please.

gares:
Thanks!

As usual some tests fail on Alpine since `time` is present but not gnu time (I do have `"conf-time" {with-test}`, can I do better?)

kit-ty-kate:
It should be fine. Thanks!

PR #15671, opened 2020-01-14
Package asak.0.2

nobrakal:
### `asak.0.2`
Partitioning OCaml codes
Asak provides functions to parse, type-check and identify similar OCaml codes.
These functions are then used to partition codes implementing the same function and help to detect code duplication.

---
* Homepage: https://github.com/nobrakal/asak
* Source repo: git+https://github.com/nobrakal/asak.git
* Bug tracker: https://github.com/nobrakal/asak/issues

---
:camel: Pull-request generated by opam-publish v2.0.2

PR #15692, opened 2020-01-18
Fix asak

kit-ty-kate:
Fix issues introduced in https://github.com/ocaml/opam-repository/pull/15671

cc @nobrakal 

Dune should not be tagged as a build dependencies (see #14266).
In releases, dune should never be called without either `--profile=release` or `-p name`. By default dune turns some warnings as errors and this should never be the case in releases as it would break packages very easily.


nobrakal:
Thanks for the PR and sorry for incorrect orignal one !

kit-ty-kate:
No worries, it's not your fault

PR #15694, opened 2020-01-19
simplify cudf install/uninstall command

timbertson:
in [opam2nix](https://github.com/timbertson/opam2nix), the `lib` suffix is scoped per ocaml version to allow for multiple ocaml versions without conflict (e.g. `lib/ocaml/4.08.1/site-lib`). So the hardcoded `lib/` in cudf installs to the wrong location.

This PR leaves the install paths up to `ocamlfind` (which does the right thing in both contexts), but keeps `BINDIR` which is used by the Makefile itself.

I've verified this builds locally with opam and produces the same output (in lib/ and bin/)

/cc @rdicosmo

### DIFF
--- a/packages/cudf/cudf.0.9-1/opam
+++ b/packages/cudf/cudf.0.9-1/opam
@@ -5,14 +5,12 @@ homepage: "http://www.mancoosi.org/cudf/"
 bug-reports: "https://gforge.inria.fr/tracker/?atid=13811&group_id=4385&func=browse"
 dev-repo: "git+https://scm.gforge.inria.fr/anonscm/git/cudf/cudf.git"
 build: [make "all" "opt"]
-remove: [
- [make "uninstall" "BINDIR=%{bin}%"]
-]
 depends: [
   "ocaml"
   "conf-perl" {build}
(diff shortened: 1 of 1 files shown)


mseri:
Could you move this change to a different patched version: `cudf-0.9-1`?

kit-ty-kate:
> This PR leaves the install paths up to `ocamlfind` (which does the right thing in both contexts), but keeps `BINDIR` which is used by the Makefile itself.

If this is the case, a dependency on `ocamlfind` is missing. I'll add it.

timbertson:
Thanks @kit-ty-kate. I'm happy with that change, but surprised that you were able to push it into my repository. Is that because you have push permissions on this upstream repo, or some other reason?

kit-ty-kate:
@timbertson every PRs made through github give the push permission on the source branch to the maintainers of the repository the PR has been made to, by default. If you prefer not to, there is a checkbox to uncheck when you open a PR.

timbertson:
Huh, never knew that, thanks

kit-ty-kate:
Thanks!

PR #15695, opened 2020-01-19
Correct the missing URL and checksums.

dmbaturin:
As @fdopen rightfully pointed out in #15682, the URL and checksums were missing from the original pull request for 1.8.0.

This corrects it.

### DIFF
--- a/packages/soupault/soupault.1.8.0/opam
+++ b/packages/soupault/soupault.1.8.0/opam
@@ -21,7 +21,7 @@ bug-reports: "https://github.com/dmbaturin/soupault/issues"
 dev-repo: "git+https://github.com/dmbaturin/soupault"
 build: [
   ["dune" "subst"] {pinned}
-  ["dune" "build" "-p" name]
+  ["dune" "build" "-p" name "-j" jobs]
 ]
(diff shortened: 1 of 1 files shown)


packages/soupault/soupault.1.8.0/opam
```diff
@@ -42,3 +42,11 @@ depends: [
   "tsort"
   "lua-ml" {>= "0.9.1"}
```
kit-ty-kate: This packages version does not exist. This is why no errors showed on the original PR

kit-ty-kate:
Could you also add `"-j" jobs` to the build command. Otherwise dune will take the number of jobs by default which is not what users might want

packages/soupault/soupault.1.8.0/opam
dmbaturin: It didn't exist at the time of writing, which is why I left a note about it in #15682. It does exist now that #15668 is merged: https://opam.ocaml.org/packages/lua-ml/lua-ml.0.9.1/
Or I'm missing something?

kit-ty-kate:
Thanks

PR #15696, opened 2020-01-20
added packages/dokeysto/dokeysto.4.0.0

UnixJunkie:
Dokeysto is no more bound to only string keys and string values
generic keys ('k) and values ('v) are now supported by the newly introduced
*_gen modules.
From my little tests, dokeysto_camltc is the fastest backend.


kit-ty-kate:
Thanks!

PR #15701, opened 2020-01-20
ocsigen-toolkit & eliom: upper version bounds on js_of_ocaml-ppx_deriving_json

jrochel:
All the affected versions use the old [%derive.json: ...] syntax.


kit-ty-kate:
Sounds good, I've opened https://github.com/ocaml/opam-repository/pull/15724 to fix the remaining failures. Merging this in the meantime even if it's not perfect. Thanks!

PR #15704, opened 2020-01-21
Update Z3 to 4.8.7

philzook58:
Changed dependency of num to zarith as per this discussion #15087 
@c-cube


kit-ty-kate:
Thanks a lot!

PR #15707, opened 2020-01-21
Package menhir.20200121

fpottier:
### `menhir.20200121`
An LR(1) parser generator

---
* Homepage: http://gitlab.inria.fr/fpottier/menhir
* Source repo: git+https://gitlab.inria.fr/fpottier/menhir.git
* Bug tracker: menhir@inria.fr

---

### DIFF
--- a/packages/menhir/menhir.20200121/opam
+++ b/packages/menhir/menhir.20200121/opam
@@ -1,0 +1,28 @@
+opam-version: "2.0"
+maintainer: "francois.pottier@inria.fr"
+authors: [
+  "François Pottier <francois.pottier@inria.fr>"
+  "Yann Régis-Gianas <yrg@pps.univ-paris-diderot.fr>"
+]
+homepage: "http://gitlab.inria.fr/fpottier/menhir"
+dev-repo: "git+https://gitlab.inria.fr/fpottier/menhir.git"
+bug-reports: "menhir@inria.fr"
+build: [
+  ["dune" "build" "@install" "-j" jobs]
+]
+install: [
+  ["dune" "install" "-j" jobs]
+]
+depends: [
+  "ocaml" {>= "4.02.3"}
+  "dune" {>= "2.0.0"}
+]
+synopsis: "An LR(1) parser generator"
+url {
+  src:
+    "https://gitlab.inria.fr/fpottier/menhir/repository/20200121/archive.tar.gz"
+  checksum: [
+    "md5=ee92ec131eb493d55f433c37019a9e84"
+    "sha512=9aa7db5ceb00322f33cf08aff866b17352cb76011382d85e6962459abbe33d33d704c8f3918206edf622b48016cc05749b0852efd2cb42109d0216097d3d1281"
+  ]
+}


packages/menhir/menhir.20200121/opam
```diff
@@ -0,0 +1,28 @@
...
+maintainer: "francois.pottier@inria.fr"
+authors: [
+  "François Pottier <francois.pottier@inria.fr>"
+  "Yann Régis-Gianas <yrg@pps.univ-paris-diderot.fr>"
+]
+homepage: "http://gitlab.inria.fr/fpottier/menhir"
+dev-repo: "git+https://gitlab.inria.fr/fpottier/menhir.git"
+bug-reports: "menhir@inria.fr"
+build: [
+  ["dune" "build" "@install" "-j" jobs]
```
kit-ty-kate:
```suggestion
  ["dune" "build" "-p" name "-j" jobs]
```
`-p` or `--profile=release` should aways be used for releases.
However mid-way of writing this, I realized you used `-w @1..66` with any profile. Would it be possible to change this?
Apparently previously those were disabled when shipped (https://gitlab.inria.fr/fpottier/menhir/commit/21741c6b14f5ebb0ac5f19852a92d0d20ee7c513#b1842d9251c4973486ce53d049c606e0120f6421_2_0) but this not the case anymore if I understand correctly.

packages/menhir/menhir.20200121/opam
```diff
@@ -0,0 +1,28 @@
...
+]
+install: [
+  ["dune" "install" "-j" jobs]
+]
```
kit-ty-kate: dune install is a command only necessary when installing from source without opam. Opam already takes care of the installation. This field can be removed.

kit-ty-kate: Ah nevermind I see you're installing several packages at once. Mmh however this feature uses ocamlfind, so ocamlfind should be added to the dependencies (see https://github.com/ocaml/dune/issues/3025)

kit-ty-kate:
Also `why3` fails to bulid with this version, but I can change the constraints before merging this:
```
# Linking  lib/plugins/genequlin.cmxs
# File "_none_", line 1:
# Error: Cannot find file menhirLib.cmo
# make: *** [Makefile:2234: bin/why3wc.byte] Error 2
# make: *** Waiting for unfinished jobs....
```

packages/menhir/menhir.20200121/opam
kit-ty-kate: Another issue is that this command will also install some files in `<prefix>/lib/coq-menhirlib` even if coq is not installed or the library not built (an empty META/dune-package file is installed at the very least)

packages/menhir/menhir.20200121/opam
fpottier: Hi Kate, thanks for your comments.

Regarding `-p <name>`, as you noted, it doesn't work, because I install several packages.

Regarding the profile, indeed, as you noted, I do not rely on dune's notion of profile; I just enable a fixed set of warnings (which are non-fatal, I think). Is that a problem?

kit-ty-kate: > which are non-fatal, I think

they are fatal, `-w @<spec>` turns them into errors

packages/menhir/menhir.20200121/opam
fpottier: Regarding the `install` field in the opam file: it is indeed needed (I tried removing it, and `dune` installed nothing at all).

Regarding `ocamlfind`: I would rather avoid a dependency on `ocamlfind` if possible. Are you certain that installing several packages at once requires `ocamlfind`? (Where is this documented?) (Why should installing several packages at once be more complicated than installing one package at a time?)

packages/menhir/menhir.20200121/opam
kit-ty-kate: one way of fixing that would be to change `_` by `dev` on this line: https://gitlab.inria.fr/fpottier/menhir/blob/8f0d2b5f2f06762d0f592f753019f92462e7b9c8/src/dune#L4 and to add `--profile=release` to the command in opam-repository

packages/menhir/menhir.20200121/opam
fpottier: Regarding your remark that "this command will also install some files in `<prefix>/lib/coq-menhirlib`": I think I just need the remove the declaration `(package (name coq-menhirlib))` in the file `dune-project`, right?

fpottier: Following up on myself: if I remove the declaration `(package (name coq-menhirlib))` then `dune` complains about the presence of the file `coq-menhirlib.opam` when I try to build Menhir.

I do not see clearly how I should declare that I have both ocaml/opam packages and coq/opam packages in the same project tree...

kit-ty-kate: It's not that it is required for several packages, it is more like it is required when using `dune install` because dune install uses ocamlfind when available (e.g. https://github.com/ocaml/dune/blob/master/src/dune/context.ml#L247).

https://github.com/ocaml/dune/issues/3025#issuecomment-574149138 gives an answer from the dune team. ocamlfind is required to avoid crashes during installation in a local switch when ocamlfind is installed globally, so to avoid surprises for users it is best to install ocamlfind in every case.

------

One other possible fix would be to stop using dune install all together and instead split the project into 3, install them normally using opam and require them in the menhir package. Dune is meant to be used this way, using dune install is more of a "hack" when used in opam.

I can probably manage to provide a patch if that's a decent solution for you.

fpottier:
Thanks for pointing out the problem with Why3. I have reported it as a new issue (#436) in the Why3 repository.

packages/menhir/menhir.20200121/opam
kit-ty-kate: Posted in https://gitlab.inria.fr/fpottier/menhir/issues/32

packages/menhir/menhir.20200121/opam
fpottier: I just realized that the flags in `src/dune` were entirely ignored because they were overridden in `src/stage1/dune`, `src/stage2/dune` and `src/stage3/dune`. I need to change the stanzas `(-open MenhirSdk)` to `(flags :standard -open MenhirSdk)`.

fpottier: Done.

packages/menhir/menhir.20200121/opam
fpottier: > One other possible fix would be to stop using dune install all together and instead split the project into 3, install them normally using opam

Do you still recommend that I add a dependency on `ocamlfind`?

kit-ty-kate: with my patch there is no need for `dune install` thus no need for `ocamlfind`

fpottier:
I am closing this pull request and trying to make a new release today.

PR #15708, opened 2020-01-21
Fix various packages

kit-ty-kate:
* cc @mseri for `eigen` and `owl`. ctypes master has removed `CI.cptr` used in both of those packages. `0.17.0` is just a guess on what the next version number of ctypes is gonna be (cc @yallop). If it's not going to be the case I'll revert this change as fast as possible.
* cc @nobrakal for `asak`
* cc @kandu for `mmseg`


mseri:
Thanks for the heads-up

PR #15712, opened 2020-01-22
Release x509 0.9.0

hannesm:
CHANGES:
    
    * BREAKING add a whitelist of hash algorithms used for signatures. The default
      whitelist is the SHA-2 family (without SHA-224), Validation.valid_ca{,s} use
      all algorithms as default
      reported by @emillon in mirleft/ocaml-x509#123, fixed in mirleft/ocaml-x509#128
    * BREAKING Certificate.hostnames and Signing_request.hostnames (new) return a
      set of [`Wildcard|`Strict] * [`host] Domain_name.t (Certificate.Host_set.t)
      reported by @mmaker in mirleft/ocaml-x509#88, fixed in mirleft/ocaml-x509#127
    * BREAKING mirleft/ocaml-x509#127 Signing_request.sign returns a result type now, an error is
      returned if the signing request was not properly signed
    * BREAKING mirleft/ocaml-x509#127 Validation.{verify_chain_of_trust, trust_key_fingerprint,
      trust_cert_fingerptint} and the type Authenticator.t changed, no longer use
      of a Certificate.host, but instead a [`host] Domain_name.t (previously, it was
      a pair)
    * BUGFIX support AlgorithmIdentifier of RSA signature algorithms with parameter
      not present
      reported by @Ulrar in mirleft/ocaml-x509#108, fixed in mirleft/ocaml-x509#129
    * BUGFIX mirleft/ocaml-x509#127 preserve a signed signing request (Country in a DN sometimes uses
      a non-utf8 string encoding)
    * remove deprecation from Validation.trust_cert_fingerprint and
      Authenticator.server_cert_fingerprint
      requested by @mben-romdhane in mirleft/ocaml-x509#125, fixed in mirleft/ocaml-x509#126
    * Certificate.signature_algorithm, CRL.signature_algorithm, and
      Signing_request.signature_algorithm are now provided, returning a
      ([`RSA|`ECDSA] * Nocrypto.Hash.hash) option
      requested by @psafont in mirleft/ocaml-x509#123, fixed in mirleft/ocaml-x509#128


hannesm:
CI failures are about lwt 5.0.0 not being compatible with tls -- if lwt is enabled and tests as well -- the expression `conflicts: [ "lwt" {with-test & >= "5.0.0"} ]` does not seem to get properly picked up by CI.

from my point of view, this PR is good to be merged.

kit-ty-kate:
Sounds good indeed. Thanks a lot!

PR #15726, opened 2020-01-24
sqlgg: fails to build with sqlite3 >= 5.0.1

ygrek:
so disable sqlite3 traits in such case.
New release will fix it later


kit-ty-kate:
Could you add a conflict instead of disabling sqlite this way?
For instance:
```
conflicts: [
  "sqlite3" {>= "5.0.0"}
]
```
Otherwise there is no way for users of your library to enforce sqlite to be enabled.

ygrek:
Thanks for the hint

kit-ty-kate:
Thanks!

PR #15746, opened 2020-01-27
Ocaml 4.10.0: second beta

Octachron:
This new beta includes new fixes for:
* the internal runtime API compatibility.
* the analysis of recursive values
* some name collisions in C files under exotic condition

The set of variants is the same that in the first beta.


kit-ty-kate:
Thanks!

PR #15754, opened 2020-01-28
Digestif 0.8.0 and MirageOS

dinosaure:
/cc @hannesm 

This is an unconventional way to keep a compatibility with MirageOS 3.7. I'm not sure if it's ok and if we should go to this way. This patch unlocks `digestif.0.8.0` to be usable by MirageOS 3.7 (and keep the trick on the META file) by applying a patch at the end of the _build_ process.


hannesm:
hmm, I'm not entirely convinced. how is `extra_files` only applied after the build ensured (I was not aware opam would do such a thing)?

kit-ty-kate:
I'm also perplexed by this.

dinosaure:
@Armael confirms that patch was done __before__ the build command.

PR #15758, opened 2020-01-29
Package zed.2.0.5

kandu:
2.0.5 (2020-01-29)
------------------

* Zed\_rope.Zip: fix a bug in function `make_b`


kit-ty-kate:
Thanks!

PR #15767, opened 2020-01-31
Luv: binding to libuv — cross-platform async I/O and other system calls

aantron:
[**Luv**](https://github.com/aantron/luv) is a binding to [**libuv**](https://github.com/libuv/libuv), which is the library that does I/O and system calls in Node.js, and runs Node's main loop. libuv is roughly equivalent to `Lwt_unix`, but, due to usage in Node, it is more cross-platform and more intensively maintained.

libuv supports event-driven I/O, multithreading (including multithreaded event-driven I/O), and multiprocessing. It also exposes a lot of miscellaneous system calls. Luv [**binds**](https://aantron.github.io/luv/luv/index.html#api-reference) all of this in OCaml.

See:

- [Project page](https://github.com/aantron/luv)
- [User guide](https://aantron.github.io/luv/index.html)
- [API docs](https://aantron.github.io/luv/luv/index.html#api-reference)
- [libuv docs](http://docs.libuv.org/en/v1.x/)

The project is named after `-luv`, the linker flag for linking with libuv, which is probably the reason why libuv has its name in the first place (in spite of any mention of "unicorn velociraptors" or the logo).

### DIFF
--- a/packages/luv/luv.0.5.0/opam
+++ b/packages/luv/luv.0.5.0/opam
@@ -1,0 +1,41 @@
+opam-version: "2.0"
+
+synopsis: "Binding to libuv: cross-platform asynchronous I/O"
+
+version: "0.5.0"
+license: "MIT"
+homepage: "https://github.com/aantron/luv"
+doc: "https://aantron.github.io/luv"
+bug-reports: "https://github.com/aantron/luv/issues"
+
+authors: "Anton Bachin <antonbachin@yahoo.com>"
+maintainer: "Anton Bachin <antonbachin@yahoo.com>"
+dev-repo: "git+https://github.com/aantron/luv.git"
+
+depends: [
+  "base-unix" {build}
+  "conf-python-3" {build}
+  "ctypes" {>= "0.13.0"}
+  "dune" {>= "2.0.0"}
+  "ocaml" {>= "4.02.0"}
+  "result"
+
+  "alcotest" {dev & >= "0.8.1"}
+]
+
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+]
+
+description: "Luv is a binding to libuv, the cross-platform C library that does
+asynchronous I/O in Node.js and runs its main loop.
+
+Besides asynchronous I/O, libuv also supports multiprocessing and
+multithreading. Multiple event loops can be run in different threads. libuv also
+exposes a lot of other functionality, amounting to a full OS API, and an
+alternative to the standard module Unix."
+
+url {
+  src: "https://github.com/aantron/luv/releases/download/0.5.0/luv-0.5.0.tar.gz"
+  checksum: "md5=2b7caeeb0227e8854fd7cc1c0f22c9a4"
+}


aantron:
Please don't merge; I'm about to retag due to a bug spotted by @avsm (a stray debug print).

aantron:
Retagged and potentially ready for merge.

kit-ty-kate:
I think `python2` should be used here instead of the default version:
```
#=== ERROR while compiling luv.0.5.0 ==========================================#
# context              2.0.6 | linux/x86_64 | ocaml-base-compiler.4.09.0 | pinned(https://github.com/aantron/luv/releases/download/0.5.0/luv-0.5.0.tar.gz)
# path                 ~/.opam/4.09/.opam-switch/build/luv.0.5.0
# command              ~/.opam/4.09/bin/dune build -p luv -j 2
# exit-code            1
# env-file             ~/.opam/log/luv-23-6de86a.env
# output-file          ~/.opam/log/luv-23-6de86a.out
### output ###
#       gyp_uv src/c/dlluv.so,src/c/libuv.a (exit 127)
# (cd _build/default/src/c/vendor/libuv && ./gyp_uv.py -f make -Duv_library=static_library)
# /usr/bin/env: 'python': No such file or directory
```

aantron:
I can change the Luv build to detect whether `python` is present. However, this does not seem reasonable to me. Is it correct that the containers have a `python2` binary but no `python` binary? Is that the issue here, or is it possible that Python is simply missing?

1. It is gyp which is referring to `python`. I think the maintainers of gyp know well about the various environments gyp needs to run in, yet they chose to use `python`. I would prefer not to add detection code to Luv, to work around something in gyp, that is likely not an error in gyp anyway.
2. `python` is the standard name. Is there a reference about it being removed or unavailable when Python is installed?

aantron:
Looking at the logs, it may be conf-python-2-7 that needs updating. However, I'll look through gyp to see if anyone has raised a similar issue with them before.

aantron:
Ok, it looks like gyp became Python 3 compatible in late 2018 (https://groups.google.com/forum/#!topic/gyp-developer/B0dH9kd_2sU, https://bugs.chromium.org/p/gyp/issues/detail?id=36), and there have also been some recent commits to fix remaining issues (see log [here](https://chromium.googlesource.com/external/gyp)). I am upgrading the vendored gyp in Luv and switching to the `python3` command. I'll retag when all the tests are finished.

aantron:
Is the opam CI lagging/down?

kit-ty-kate:
> Is the opam CI lagging/down?

Yes, it has been broken for a few days. I have to relaunch it manually :/

kit-ty-kate:
I believe the `conf-python-2-7` dependency should be replaced by `conf-python-3`:
```
+ /home/opam/.opam/4.09/bin/dune "build" "-p" "luv" "-j" "2" (CWD=/home/opam/.opam/4.09/.opam-switch/build/luv.0.5.0)
- File "src/c/dune", line 23, characters 8-15:
- 23 |    (run python3 gyp_uv.py -f make -Duv_library=static_library))
-              ^^^^^^^
- Error: Program python3 not found in the tree or in PATH
-  (context: default)
```

aantron:
Yes, you're right. I did that locally, but forgot to update the file I'm sending to opam.

aantron:
Updated.

kit-ty-kate:
All green now (except an unrelated failure on macos). Thanks!

aantron:
Thank you!

PR #15768, opened 2020-01-31
Unlock digestif.0.8.0 and MirageOS.3.7.*

dinosaure:
This PR is on the way of #15754. The process is to manually call `patch` after the build process and unlock the usability of `digestif` with MirageOS 3.7.*. I completely agree that this solution is not great and we should do something else at the next release of `digestif` - an internal tool which trick on the META file (if the MirageOS support of C stubs does not change by then) but it seems to work.

/cc @hannesm


kit-ty-kate:
Thanks!

PR #15769, opened 2020-01-31
Package fix.20200131

fpottier:
### `fix.20200131`
Facilities for memoization and fixed points

---
* Homepage: https://gitlab.inria.fr/fpottier/fix
* Source repo: git+https://gitlab.inria.fr/fpottier/fix.git
* Bug tracker: francois.pottier@inria.fr

---

### DIFF
--- a/packages/fix/fix.20200131/opam
+++ b/packages/fix/fix.20200131/opam
@@ -1,0 +1,24 @@
+opam-version: "2.0"
+maintainer: "francois.pottier@inria.fr"
+authors: [
+  "François Pottier <francois.pottier@inria.fr>"
+]
+homepage: "https://gitlab.inria.fr/fpottier/fix"
+dev-repo: "git+https://gitlab.inria.fr/fpottier/fix.git"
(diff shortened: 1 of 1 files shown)


kit-ty-kate:

fpottier:
Hi Kate, could you explain the fix? I think that `dune` is needed only to build `fix` but is not needed by someone who wishes to use `fix`. Am I wrong?

kit-ty-kate:
> Hi Kate, could you explain the fix? I think that `dune` is needed only to build `fix` but is not needed by someone who wishes to use `fix`. Am I wrong?

It used to be the case but it's no longer the case since dune stores the version number of the dune binary used to install the package in a file dune later uses to extract metadata from. If somewhere down the line dune gets downgraded, dune will crash when trying to read the file it installed before.
Relevant issues:
* https://github.com/ocaml/opam-repository/pull/14266
* https://github.com/ocaml/opam-repository/issues/14185
* https://github.com/ocaml/opam/issues/3850
* https://github.com/ocaml/dune/issues/2147

PR #15770, opened 2020-02-01
Add pkg libdash

mgree:
A library with OCaml bindings to dash, a common, minimal POSIX shell. Provides access to the AST and a few other functions.

This is my first OPAM package, so I'm sure I've made some mistakes! Happy to receive any/all feedback.

One tiny note: I'm only able to install this on my Mac with `SDKROOT` unset and `TMPDIR=/tmp`.


packages/libdash/libdash.0.1/opam
```diff
@@ -0,0 +1,30 @@
...
+  "conf-autoconf" {build}
+]
+build: [
+  ["./autogen.sh"]
+  ["./configure" "--prefix=%{prefix}%" "--libdir=%{lib}%"]
+  [make]
+]
+install: [
+  [make "install"]
+  [make "-C" "ocaml"]
```
kit-ty-kate: Is this command building something? If so this should be in the `build` section, not in the `install` section

kit-ty-kate:
Could you return the fix I just pushed upstream?

packages/libdash/libdash.0.1/opam
kit-ty-kate: Looking at the makefile I think it is the case. I'll fix that

packages/libdash/libdash.0.1/opam
```diff
@@ -0,0 +1,31 @@
...
+version: "0.1"
+synopsis: "Bindings to the dash shell's parser"
+maintainer: "Michael Greenberg <michael.greenberg@pomona.edu>"
+authors: "Michael Greenberg <michael.greenberg@pomona.edu>"
+license: "BSD"
+homepage: "https://github.com/mgree/libdash"
+bug-reports: "https://github.com/mgree/libdash/issues"
+depends: [
+  "ocaml" {>= "4.0.7"}
+  "ctypes" {= "0.11.5"}
```
kit-ty-kate:
```suggestion
  "ctypes" {>= "0.11.5"}
```
Is there a reason you have a constraint so strict?

mgree: I haven't been able to get my whole system to build with later ctypes... but this library should be fine.

mgree:
The move from install to build breaks my build. I think the core problem here is that I don't know the right way to package a C library with its OCaml bindings. Is there a reference package I can look at that does this?

For background, my library has three relevant directories:

- src, which is a fork of the kernel.org dash repository with some added bindings. this builds a static and a shared library
- ocaml, which is a ctypes-based binding to the shared library
- test, which exercises the bindings in ocaml

Ideally, my OPAM script should install the src library where ocaml and test can see it. I'm not sure how to do that. In a truly ideal world, there'd be none of ctypes's dynamic linking at all, but I haven't found good documentation on how to statically link OCaml with C. Any/all help would be appreciated. (I'd post on discuss.ocaml.org, but the system put my account on hold for too many URLs and I've been waiting forever for that hold to be released.)

Sorry to unload, really not sure where to go/what to do here... :(

mgree:
I'm trying to get this to work in my local container-based CI, but the build is failing because it can't find `conf-libtool`. Searching doesn't seem to help... I'm using the Docker image `ocaml/opam2:debian-stable`.

kit-ty-kate:
I added the package just yesterday so it's not in the docker image yet. You can have it by doing `git pull origin master && opam update` first

mgree:
Gotcha. I can wait a minute or two, since the package is still unusable. :(

Any help on the build system in general would be appreciated: even in versions where I'm able to build the library, I've been unable to use it---I get linker errors no matter what. Are there any examples of bindings to C libraries where the library itself is bundled with the OPAM package? I have no idea where to put the .a/.so files or which incantations to make to the myriad OCaml tools (ocamlfind, ocamlmklib, ocamlopt, etc.).

avsm:
avsm/ocaml-yaml is a dune project that bundles the c library with the project

mgree:
The OPAM file is now in a shape where it ought to work---I can build it cleanly in docker if (a) I have libtool installed and (b) I drop the `conf-libtool` build dependency.

Should I fast forward this branch to include `conf-libtool`? The Travis logs seem to indicate that `conf-libtool` installed but `libtoolize` wasn't found.

kit-ty-kate:
> Should I fast forward this branch to include `conf-libtool`? The Travis logs seem to indicate that `conf-libtool` installed but `libtoolize` wasn't found.

Yes, that'd be perfect, thanks!
Other than that, an issue seems to be triggered during tests. Could you have a look at it?
```
#=== ERROR while installing libdash.0.1 =======================================#
# context              2.0.6 | linux/x86_64 | ocaml-base-compiler.4.08.1 | pinned(https://github.com/mgree/libdash/archive/0.1.tar.gz)
# path                 ~/.opam/4.08/.opam-switch/build/libdash.0.1
# command              /usr/bin/make -C test test
# exit-code            2
# env-file             ~/.opam/log/libdash-37-ad8886.env
# output-file          ~/.opam/log/libdash-37-ad8886.out
### output ###
# make: Entering directory '/home/opam/.opam/4.08/.opam-switch/build/libdash.0.1/test'
# ocamlfind ocamlopt -g -package dash,ctypes,ctypes.foreign -linkpkg test.ml -o test.native
# File "test.ml", line 1:
# Error: No implementations provided for the following modules:
#          Str referenced from /home/opam/.opam/4.08/lib/dash/dash.cmxa(Dash)
# make: *** [Makefile:17: test.native] Error 2
# make: Leaving directory '/home/opam/.opam/4.08/.opam-switch/build/libdash.0.1/test'
```

mgree:
I just realized that OPAM was pulling the release tarball and not from git `master`---which is quite reasonable, of course. Hopefully this revision will actually build...

mgree:
Alas. I see two core problems:

- On macOS, libtool isn't working (and yet `conf-libtool` installed... a bug?).
- On Linux, the tests need `Str`.

The solution to the former seems to be to make sure libtool is installed. I'm not certain how opam-repository should handle dependencies on tooling like this.

The solution to the latter is confusing me. What is different about the build setup here and the build setup in [libdash's own CI](https://travis-ci.com/mgree/libdash)? Why doesn't it pick up the `Str` dependency from the library itself, which is linked with `-linkpkg`? (I can probably solve this problem more easily by just replacing the call to `Str.split`, but I'm always reluctant to simply dodge issues.)

In other news, thank you for sticking with me in what is turning out to be a much longer and more involved process than I anticipated! 😅

mgree:
🎉 macOS builds and passes the tests!

The Linux builds are failing tests... and I suspect it's about [missing ldconfig](https://travis-ci.org/ocaml/opam-repository/jobs/652652200#L1988). Could it be the case that the "install" stage of the sandbox is being too aggressive here? Any advice on configuring that?

One possibility would be to configure the C library to install into a local `_build` directory; then the installation procedure would be a plain `ocamlfind install` call. I've been avoiding this because I've gotten away thus far not having to predict the precise library names/files that will be generated. Any thoughts? Am I missing something about OPAM's or the CI harness's behavior?

mgree:
I suspect there's a bug in `conf-autoconf` on [alpine](https://ci.ocaml.org/log/saved/docker-build-fc4782a77e1000803196ec12a270e73c/4c0317db210d97c3713e6dbd67f39e493eb87b8c), where `aclocal` isn't present. Though `aclocal` is usually part of `automake` and not `autoconf`, so perhaps we need a `conf-automake` dependency, too?

mgree:
Whew! We have a working build. The distro builds haven't been rerun yet, and I expect alpine will continue to fail because of missing `aclocal`.

I'm happy to try to create a `conf-automake` package that will install all of the automake tools, though I suspect that someone with more experience will find more success, more quickly.

kit-ty-kate:
Looks good. Thanks!

PR #15772, opened 2020-02-01
Remove jbuilder.transition

talex5:
This transition version was added back when dune was compatible with jbuilder, so it made sense to install dune in all cases. However, recent versions of dune no longer support jbuilder, so using this package just forces installation of an old version of dune, conflicting with packages that require a recent dune, which is generally undesirable.


kit-ty-kate:
Sounds good but could you set `available: false` instead of removing the file?

dra27:
Rather than unavailable, shouldn’t it be that it conflicts dune < 2.0.0? The old jbuilder packages are still not compatible with dune 1

talex5:
Ah. So any package that requires both `jbuilder` and `dune<2` will have no solution. For example:

`opam install --show-actions 'jbuilder<transition' csv.2.4 lwt.3.3.0 --unlock-base` fails, but succeeds if you change the `<` to `.`. How about changing the version of `jbuilder.transition` so that it is not the most-preferred, but could still be used?

talex5:
Given that dune 1 conflicts with jbuilder, I don't think this PR is an improvement.

PR #15779, opened 2020-02-03
[new release] elpi (1.10.0)

gares:
CHANGES:

- Compiler:
  - large refactoring to separate the table of global (statically initialized)
    symbols, the table of symbols of a program being compiled and the table
    of symbols used at runtime.
  - API for separate compilation.

- API:
  - Setup.init now returns a handle to an elpi instance to be passed to
    many other APIs.
  - New APIs Compile.unit and Compile.assemble for separate compilation.
  - Constants.from_stringc and Constants.mk*S API removed in favor for
    Constants.declare_global_symbol (to be used in module initialization code).

### DIFF
--- a/packages/elpi/elpi.1.10.0/opam
+++ b/packages/elpi/elpi.1.10.0/opam
@@ -10,7 +10,7 @@ bug-reports: "https://github.com/LPCIC/elpi/issues"
 build: [
   ["dune" "subst"] {pinned}
   ["dune" "build" "-p" name "-j" jobs]
-  [make "tests" "DUNE_OPTS=-p %{name}%"] {with-test & os != "macos"}
+  [make "tests" "DUNE_OPTS=-p %{name}%" "TIMEOUT=120"] {with-test & os != "macos"}
 ]
 
 depends: [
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
It seems that `conf-lua` is needed during tests:
```
# RUNNING  zebra (standard Prolog benchmark)           elpi
/usr/bin/env: 'lua5.1': No such file or directory
```

gares:
I've added conf-lua but this is not the cause. It seems one test failed by timeout (90s). Here it is indeed close to this limit, probably your VMs got a little slower than before and/or Elpi a bit less efficient. I've pushed a commit setting the timeout to 120s instead of 90s.

gares:
It turns out the VS does not have lua, so conf-lua failed... I'm removing that commit since it is not the real problem enyway

gares:
@kit-ty-kate it seems that the other CI has not triggered. Could you please activate it?

gares:
Thanks @kit-ty-kate , CI is green now!

kit-ty-kate:
Thanks!

PR #15789, opened 2020-02-05
os-distribution archlinux should be arch

avsm:
`docker run archlinux cat /usr/lib/os-release` shows that ID=arch.
Both are used in opam-repository, so correct archlinux to arch in
this PR.


mseri:
Some packages are using `os-family`, some `os-distribution`. Should we make them use the same? Is `arch` valid in both cases?

avsm:
Good question; I need to check os-family

kit-ty-kate:
It seems like having `os-family = "arch"` would be a good solution. For instance, Arch Linux ARM (used for Raspberry Pi) has:
```
ID=archarm
ID_LIKE=arch
```

kit-ty-kate:
In the meantime this PR is good to be merged anyway. Thanks!

PR #15792, opened 2020-02-06
OCaml 4.10, first release candidate

Octachron:
This PR adds the same variants as the second beta: normal, flambda, afl, frame-pointer, frame-pointer+flambda .
There are two bug fixes since the last beta:
-  the backtrace generation in the bytecode interpreter is no longer confused by large integers
- no more strange invalid module names errors when printing inline record types with a type/cmi name collision.


kit-ty-kate:
Thanks!

PR #15796, opened 2020-02-06
Opam publish camlp5.7.11

chetmurthy:
### `camlp5.7.11`
Preprocessor-pretty-printer of OCaml
Camlp5 is a preprocessor and pretty-printer for OCaml programs. It also provides parsing and printing tools.

As a preprocessor, it allows to:

extend the syntax of OCaml,
redefine the whole syntax of the language.
As a pretty printer, it allows to:

display OCaml programs in an elegant way,
convert from one syntax to another,
check the results of syntax extensions.
Camlp5 also provides some parsing and pretty printing tools:

extensible grammars
extensible printers
stream parsers and lexers
pretty print module
It works as a shell command and can also be used in the OCaml toplevel.

---
* Homepage: https://camlp5.github.io
* Source repo: git+https://github.com/camlp5/camlp5.git
* Bug tracker: https://github.com/camlp5/camlp5/issues

---
:camel: Pull-request manually copied from previous failed PR (which was generated by opam-publish v2.0.2)


packages/camlp5/camlp5.7.11/opam
```diff
@@ -0,0 +1,48 @@
...
+maintainer: "https://github.com/ocaml/opam-repository/issues"
+authors: ["Daniel de Rauglaudre"]
+homepage: "https://camlp5.github.io"
+license: "BSD-3-Clause"
+bug-reports: "https://github.com/camlp5/camlp5/issues"
+dev-repo: "git+https://github.com/camlp5/camlp5.git"
+doc: "https://camlp5.github.io/doc/html"
+
+depends: [
+  "ocaml"       { >= "4.02" & <= "4.10.0" }
```
kit-ty-kate:
```suggestion
  "ocaml"       { >= "4.02" & < "4.11" }
```
I'm pretty sure camlp5 could support a minor version of 4.10 such as 4.10.1. If this is indeed the case, could you apply this patch and return the fix upstream?

chetmurthy: (1) thank you for the suggestion!  I appreciate that you and your team are trying to keep things ... "clean".  It's  gotta be a little tedious, but I appreciate that you're doing it!
(2) Oof, if I'd only come here first, I could have just clicked that button, instead of implementing the change myself!  Next time!

In any case (I think!) all done.

kit-ty-kate:
Thanks!

PR #15778, opened 2020-02-03
Package sawja.1.5.8

anscoil:
### `sawja.1.5.8`
Sawja provides a high level representation of Java bytecode programs and static analysis tools
Sawja is a library written in OCaml, relying on Javalib to provide a high level representation of Java bytecode programs. Its name stands for Static Analysis Workshop for JAva. Whereas Javalib is dedicated to isolated classes, Sawja handles bytecode programs with their class hierarchy and control flow algorithms.
Moreover, Sawja provides some stackless intermediate representations of code, called JBir and A3Bir. The transformation algorithm, common to these representations, has been formalized and proved to be semantics-preserving.

---
* Homepage: http://sawja.inria.fr
* Source repo: git+https://github.com/javalib-team/sawja.git
* Bug tracker: https://github.com/javalib-team/sawja/issues

---
:camel: Pull-request generated by opam-publish v2.0.0

PR #15803, opened 2020-02-07
Package Sawja.1.5.8 (replaces previous attempt)

davidpichardie:
Please consider this PR instead of #15778


kit-ty-kate:
Thanks!

PR #15807, opened 2020-02-08
Lambda Soup 0.7.0 - HTML scraping using CSS

aantron:
From the [changelog](https://github.com/aantron/lambdasoup/releases/tag/0.7.0):

> - Raise a new exception `Soup.Parse_error` when CSS selectors can't be parsed, rather than `Failure` (aantron/lambdasoup#31, Daniil Baturin).
> - Add `with-test` command to `.opam` file (aantron/lambdasoup#30, Craig Ferguson).

cc @dmbaturin

### DIFF
--- a/packages/lambdasoup/lambdasoup.0.7.0/opam
+++ b/packages/lambdasoup/lambdasoup.0.7.0/opam
@@ -24,7 +24,7 @@ depends: [
 
 build: [
   ["dune" "build" "-p" name "-j" jobs]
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Thanks a lot! Could you return the fix upstream?

aantron:
Thanks! Yes, I will fix the upstream repo soon (perhaps tomorrow).

PR #15808, opened 2020-02-08
Release tcpip 4.1.0

hannesm:
[new release] tcpip (4.1.0)
    
CHANGES:
    
* Revert "Ipv4.Fragments use a Lru.M.t instead of Lru.F.t" (mirage/mirage-tcpip#423 by @hannesm)
      A Lru.M.t allocates a Hashtbl.t of size = capacity (= 256 * 1024 in our case),
      this leads to excessive ~2MB memory consumption for each Fragment cache,
      reported by @xaki23 in mirage/qubes-mirage-firewall#93
* use SOCK_RAW for an ICMP socket in the unix sockets API (previously used
      SOCK_DGRAM which did not work)
      reported by @justinc1 in mirage/mirage-tcpip#358, fixed in mirage/mirage-tcpip#424 by @hannesm
* tcp is now compatible with lwt >= 5.0.0 (where Lwt.async requires a function
      of (unit -> unit Lwt.t) (mirage/mirage-tcpip#370 mirage/mirage-tcpip#425 @cfcs @hannesm, issue mirage/mirage-tcpip#392 @emillon)
* Add a dependency on dune-configurator to support dune 2.0.0 (mirage/mirage-tcpip#421 @avsm)

### DIFF
--- a/packages/mirage-nat/mirage-nat.2.0.0/opam
+++ b/packages/mirage-nat/mirage-nat.2.0.0/opam
@@ -24,7 +24,7 @@ depends: [
   "lru" {>= "0.3.0"}
   "ppx_deriving" {>= "4.2" }
   "dune" {>= "1.0"}
-  "tcpip" { >= "3.7.8" }
+  "tcpip" { >= "3.7.9" & < "4.1.0" }
   "ethernet" { >= "2.0.0" }
   "arp"
   "alcotest" {with-test}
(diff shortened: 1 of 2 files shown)


hannesm:
the test failure on macOS is expected (been there since several releases, still unclear why it fails -- someone with a mac and interest in the tcpip package looking into that would be appreciated)

kit-ty-kate:
Thanks!

PR #15813, opened 2020-02-09
constrain jsoo-lwt to the same version as jsoo

avsm:
This is done in the latest releases, but not historical ones.
The lack of constraint leads to a failure in 4.03.0 due to a very
new jsoo-lwt being installed with an older jsoo.  Its just safer
to not mix these versions...


kit-ty-kate:
Already merged in https://github.com/ocaml/opam-repository/pull/15724

PR #15811, opened 2020-02-08
Links version 0.9.1

dhil:
A new minor release of the Links programming language. This release includes bug fixes and other improvements.

PR #15815, opened 2020-02-09
Sync links' sub-packages versions

kit-ty-kate:
Sync up versions of links' sub-packages with versions of links.
Fixes revdeps failure in #15811

cc @dhil is that ok? Could it also be the same thing with the 0.9.1 packages in #15811?

### DIFF
--- a/packages/links-postgresql/links-postgresql.0.7.2/opam
+++ b/packages/links-postgresql/links-postgresql.0.7.2/opam
@@ -14,7 +14,7 @@ depends: [
   "ocaml" {>= "4.04.0" & < "4.06"}
   "jbuilder"
   "postgresql"
-  "links"
+  "links" {= version}
 ]
 synopsis: "Linking Theory to Practice for the Web"
 description: """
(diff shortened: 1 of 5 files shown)


dhil:
> cc @dhil is that ok? Could it also be the same thing with the 0.9.1 packages in #15811?

I'd like to say 'yes', but I do not understand the rationale or the implications. I guess I am not sure what the variable `version` gets resolved to.

kit-ty-kate:
> > cc @dhil is that ok? Could it also be the same thing with the 0.9.1 packages in #15811?
> 
> I'd like to say 'yes', but I do not understand the rationale or the implications. I guess I am not sure what the variable `version` gets resolved to.

the `version` variable gets resolved to the version the current package. For instance `"links" {= version}` in `links-postgresql.0.7.2` is the same as `"links" {= "0.7.2"}`

dhil:
Thank you for the clarification. This change is fine by me.

kit-ty-kate:
The implication is that all the packages from the same "release bundle" will always be using the same version across all their packages.

PR #15823, opened 2020-02-11
New release of Menhir.

fpottier:
Including major bug fixes and a dependency on `dune` 2.2.0.


kit-ty-kate:
As explained in https://github.com/ocaml/opam-repository/pull/15769#issuecomment-581046715 dune should not be tagged as a build dependency. Could you return the fix upstream?

Thanks!

fpottier:
Done, thanks.

PR #15826, opened 2020-02-11
OCaml 4.10.0+rc2, second release candidate

Octachron:
This new release candidate integrates a bug fix for the new best-fit allocator on 32-bit system (https://github.com/ocaml/ocaml/pull/9292). Other users are not expected to be affected nor to update. This RC is mostly here to minimize the difference between the rc and final binaries.


kit-ty-kate:
Thanks!

PR #15821, opened 2020-02-10
ocaml-migrate-parsetree 1.6.0

aantron:
From the [changelog](https://github.com/ocaml-ppx/ocaml-migrate-parsetree/releases/tag/v1.6.0):

> CHANGES:
>
> - Preserve compiler version of binary ASTs across transformation (ocaml-ppx/ocaml-migrate-parsetree#79, @aantron)
>
> - Allow not exiting on error (ocaml-ppx/ocaml-migrate-parsetree#83, @aantron)

PR #15830, opened 2020-02-12
[new release] ppxfind (1.4)

diml:
Tool combining ocamlfind and ppx

- Project page: <a href="https://github.com/diml/ppxfind">https://github.com/diml/ppxfind</a>
- Documentation: <a href="https://diml.github.io/ppxfind/">https://diml.github.io/ppxfind/</a>

##### CHANGES:

- Fix compatibility with ocaml-migrate-parsetree >= 1.6.0

### DIFF
--- a/packages/ppxfind/ppxfind.1.4/opam
+++ b/packages/ppxfind/ppxfind.1.4/opam
@@ -1,0 +1,44 @@
+opam-version: "2.0"
+synopsis: "Tool combining ocamlfind and ppx"
+description: """
+Ppxfind is a small command line tool that among other things allows
+to use old style ppx rewriters with jbuilder.
+"""
+maintainer: ["Jérémie Dimino <jeremie@dimino.org>"]
+authors: ["Jérémie Dimino <jeremie@dimino.org>"]
+license: "BSD3"
+homepage: "https://github.com/diml/ppxfind"
+doc: "https://diml.github.io/ppxfind/"
+bug-reports: "https://github.com/diml/ppxfind/issues"
+depends: [
+  "dune" {>= "2.0"}
+  "ocaml-migrate-parsetree"
+  "ocamlfind"
+  "ocaml" {>= "4.02.3"}
+]
+conflicts: [
+  "dune" {= "1.2" | = "1.2.1"}
+]
+build: [
+  ["dune" "subst"] {pinned}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
+dev-repo: "git+https://github.com/diml/ppxfind.git"
+url {
+  src:
+    "https://github.com/diml/ppxfind/releases/download/1.4/ppxfind-1.4.tbz"
+  checksum: [
+    "sha256=98291c69f04f7f7b7cdad1b5d786c70fc595559d4663cc04cb711ac132db4971"
+    "sha512=f80b0ee09fb536aa9f154da80d06a1b68ba3b10605fb7338bd6449beb5c8d00e983bf66b4a63e12659ae1410fea56d0a2c4cfd43584616438504628035bcb981"
+  ]
+}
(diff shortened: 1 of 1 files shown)


packages/ppxfind/ppxfind.1.4/opam
```diff
@@ -0,0 +1,44 @@
...
+"""
+maintainer: ["Jérémie Dimino <jeremie@dimino.org>"]
+authors: ["Jérémie Dimino <jeremie@dimino.org>"]
+license: "BSD3"
+homepage: "https://github.com/diml/ppxfind"
+doc: "https://diml.github.io/ppxfind/"
+bug-reports: "https://github.com/diml/ppxfind/issues"
+depends: [
+  "dune" {>= "2.0"}
+  "ocaml-migrate-parsetree"
```
kit-ty-kate:
```suggestion
  "ocaml-migrate-parsetree" {>= "1.6.0"}
```
@aantron is that what you were saying in #15821 ?

packages/ppxfind/ppxfind.1.4/opam
```diff
@@ -0,0 +1,44 @@
...
+  "ocamlfind"
+  "ocaml" {>= "4.02.3"}
+]
+conflicts: [
+  "dune" {= "1.2" | = "1.2.1"}
+]
```
kit-ty-kate: This conflict is redundant since dune >= 2.0 is already required.

packages/ppxfind/ppxfind.1.4/opam
aantron: Yes, but I think we need to merge that PR first, before the change would make sense (the tests shouldn't be able to solve the constraints until then). See also https://github.com/diml/ppxfind/commit/14166274726ca24db097cbe18066ab9b2b76d9ea#diff-4f99a9f8e368c141259cd296f03c6049L15.

kit-ty-kate:
Thanks!

packages/ppxfind/ppxfind.1.4/opam
diml: Indeed, I fixed the upstream repo, thanks!

packages/ppxfind/ppxfind.1.4/opam
diml: BTW, I tested the build with omp 1.5.0 when I submitted the PR and it was building fine.

diml: So I'm not sure the constraint is necessary at all. Except if it's for some other reason I missed.

kit-ty-kate: The package itself compiled with omp 1.5.0 but the revdeps came back all negative: https://ci.ocaml.org/ocaml/opam-repository/pr/15830?history=5791b5da8eaa2ab041e18fd9eac2e1f22344341c&test=2%20Revdeps

aantron: It's necessary, because the `~argv` support is buggy before 1.6.0. I fixed it in https://github.com/ocaml-ppx/ocaml-migrate-parsetree/commit/0d28ffb24a3e555f280971fc6672ef39da439f5a, that's what parsing from the first argument refers to.

aantron: The "full" history is that I added `~argv` support for a specific use case in Bisect_ppx, that wasn't released anyway. The `~argv` support was released in 1.5.0, which was released for other reasons not related to Bisect_ppx or any other project that needed `~argv` at the time. It just happened to already be in `master` :) I realized that the support was buggy almost immediately, but the PR that fixed that remained unmerged (by me) for several months, across the 1.5.0 release, because, again, there was no pressing need for the feature anyway. I didn't even know there was a 1.5.0 release for a while :p 1.6.0 is the first release that properly includes the working version of this feature.

aantron: The release then broke ppxfind, which I wasn't aware of, but `~argv` was the way to fix ppxfind. The buggy version of `~argv` in 1.5.0 was buggy in exactly the right way to not break ppxfind.

PR #15831, opened 2020-02-12
Constrain existing releases of ppxfind

aantron:
See https://github.com/ocaml/opam-repository/pull/15821 for OMP 1.6.0 and https://github.com/ocaml/opam-repository/pull/15830 for a new release of ppxfind that is compatible with it.


kit-ty-kate:
Thanks!

PR #15836, opened 2020-02-13
Add Jsoo_storage 1.0.1 and Jsoo_broadcastchannel 2.0.0

xvw:


kit-ty-kate:
Could you return the fixes upstream? (dune should not be tagged as a build dependency, see #14266)

All green otherwise. Thanks!

PR #15838, opened 2020-02-13
Promise: native implementation of a JS promise binding

aantron:
[**Promise**](https://github.com/aantron/promise) is primarily a type-safe binding to JS promises on BuckleScript. However, it also offers a native implementation, for users who wish to write code that works on both BuckleScript and native. This release contains the packaging of that native implementation.


kit-ty-kate:
Thanks!

PR #15853, opened 2020-02-17
Result 1.5 - improved 4.08 compatibility

aantron:
This release mainly addresses https://github.com/janestreet/result/issues/8, which is that `Result` from `result` shadows `Result` from `Stdlib`, which is available starting from 4.08.

`Stdlib.Result` has more values declared and a different type name, and all of these become inaccessible because of the shadowing, unless one consistently writes `Stdlib.Result` instead of just `Result`. This was a great annoyance if `result` appeared anywhere in the dependency graph of a project, especially if it was introduced at a late stage in development, which could cause a large amount of code to break.

EDIT: This release makes `Result` from `result` include `Stdlib.Result` on OCaml >= 4.08.

cc @diml

### DIFF
--- a/packages/result/result.1.5/opam
+++ b/packages/result/result.1.5/opam
@@ -1,0 +1,22 @@
+opam-version: "2.0"
+maintainer: "opensource@janestreet.com"
+authors: ["Jane Street Group, LLC <opensource@janestreet.com>"]
+homepage: "https://github.com/janestreet/result"
+dev-repo: "git+https://github.com/janestreet/result.git"
+bug-reports: "https://github.com/janestreet/result/issues"
+license: "BSD-3-Clause"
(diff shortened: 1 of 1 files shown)


aantron:
Please consider this as WIP until we address https://github.com/janestreet/result/pull/9#issuecomment-586983275.

aantron:
Just repeating the comment from https://github.com/ocaml/opam-repository/pull/15893: I believe this can be merged, as only one revdep failure was due to `result`; that was the test cases of `containers.2.7`, which were constrained in https://github.com/ocaml/opam-repository/pull/15893. There are newer releases of `containers`, so nothing needs to be done upstream.

kit-ty-kate:
Thanks (sorry for the huge delay)

aantron:
Thanks!

PR #15856, opened 2020-02-17
nice_parser: fixed checksums

smolkaj:
Not sure why, but the checksum for this package were wrong.
@cakoch10


kit-ty-kate:
The change in checksum is due to the rename of the repository from ocaml-parsing to nice-parser.
No other difference has been spotted using:
```
diff -ru ocaml-parsing-1.0.0 nice-parser-1.0.0
```
This is good to merge. Thanks!

PR #15857, opened 2020-02-18
added packages/batteries/batteries.3.0.0

UnixJunkie:


UnixJunkie:
Users should note that this is a major release, so the changelog contains some (rather small) but potentially breaking changes.

kit-ty-kate:
All good. Thanks! Could you return the fix upstream as well?

UnixJunkie:
ok, thanks!

PR #15859, opened 2020-02-18
Constrain Alcotest to use Fmt >=0.8.6

CraigFe:
The corresponding change upstream is https://github.com/mirage/alcotest/pull/219.


kit-ty-kate:
Thanks!

PR #15865, opened 2020-02-18
[new release] letsencrypt (0.2.0)

hannesm:
CHANGES:

* support ACME as specified in RFC 8555 (letsencrypt v2 endpoints) mmaker/ocaml-letsencrypt#19
* support for the ALPN challenge as well mmaker/ocaml-letsencrypt#19


kit-ty-kate:
Thanks!

PR #5603, opened 2016-02-14
snappy: patch snappy to work on osx

dsheets:

PR #15866, opened 2020-02-18
+ snappy 0.1.2

ygrek:


ygrek:
ftr this removes patch from https://github.com/ocaml/opam-repository/pull/5603 which I consider wrong (host build environment configuration is outside of scope of opam)

kit-ty-kate:
Thanks!

PR #15869, opened 2020-02-19
js_of_ocaml-ppx_deriving_json: Add missing dependency

rixed:
Error was:
  ocamlfind: Package `ppx_deriving' not found - required by `js_of_ocaml-ppx_deriving_json'

### DIFF
--- a/packages/js_of_ocaml-ppx_deriving_json/js_of_ocaml-ppx_deriving_json.3.5.2/opam
+++ b/packages/js_of_ocaml-ppx_deriving_json/js_of_ocaml-ppx_deriving_json.3.5.2/opam
@@ -19,6 +19,7 @@ depends: [
   "js_of_ocaml" {= version}
   "ocaml-migrate-parsetree"
   "ppxlib" {>= "0.9.0"}
+  "ppx_deriving"
 ]
 url {
   src:


kit-ty-kate:
`ppx_deriving` does not seem to be used though: https://github.com/ocsigen/js_of_ocaml/blob/master/ppx/ppx_deriving_json/lib/dune

cc @hhugo

kit-ty-kate:
Where are you encountering the issue?

hhugo:
js_of_ocaml-ppx_deriving_json dropped its dependency on ppx_deriving starting with version 3.5

rixed:
Interesting. Here is the META file of js_of_ocaml-ppx_deriving_json that's installed:

```
version = "3.5.2"
description = "[@@deriving json]"
requires(ppx_driver) = "compiler-libs.common ocaml-migrate-parsetree ppxlib"
archive(ppx_driver,byte) = "ppx_deriving_json.cma"
archive(ppx_driver,native) = "ppx_deriving_json.cmxa"
plugin(ppx_driver,byte) = "ppx_deriving_json.cma"
plugin(ppx_driver,native) = "ppx_deriving_json.cmxs"
# This is what dune uses to find out the runtime dependencies of
# a preprocessor
ppx_runtime_deps = "js_of_ocaml.deriving"
# This line makes things transparent for people mixing preprocessors
# and normal dependencies
requires(-ppx_driver) = "js_of_ocaml.deriving"
requires(-ppx_driver,-custom_ppx) += "ppx_deriving"
ppxopt(-ppx_driver,-custom_ppx) = "ppx_deriving,package:js_of_ocaml-ppx_deriving_json"
library_kind = "ppx_deriver"
```

Notice at the end the conditional addition of ppx_deriving.

rixed:
I have no idea where those ppx_driver or custom_ppx predicates are supposed to come fro though.

hhugo:
This is added by dune. What build system do you use ?

hhugo:
cc @diml

rixed:
You mean the predicate is added by dune?
I'm using plain ocamlfind with `-package js_of_ocaml-ppx_deriving_json`.

hhugo:
I mean, dune is generating the META file that is installed with jsoo.

I believe things would work out of the box if you were to use dune to compile your project.

I don't think there is an issue in opam-repository, rather one should document the non dune workflow in jsoo.

rixed:
Looks like there is indeed a secret protocol between dune and ocamlfind involving those predicates, but that's still an opam bug as long as out of the box if one installs js_of_ocaml-ppx_deriving_json and try to use it, opam outputs that js_of_ocaml-ppx_deriving_json requires ppx_deriving and fails.

hhugo:
> opam outputs that js_of_ocaml-ppx_deriving_json requires ppx_deriving and fails

I assume you meant to write `ocamlfind`, not `opam`.


ppx_deriving_json needs a "driver" to work. I believe dune will create the driver using ppxlib. An alternative is to use ppx_deriving.

Do you think a more descriptive error message would address your issue ?

hhugo:
I believe this might affects all ppx derivers built with dune. @diml, any thoughts ?

diml:
Well, this is all super complicated and mostly to be backward compatible. `ppx_deriving` is indeed not needed at all if you build using dune, however it is required if you build with `ocamlformat`. i.e. if want to use ppx rewriter with `ocamlfind` via:

```
$ ocamlfind ocamlc -package js_of_ocaml-ppx_deriving_json file.ml
```

This is because:
- when using ppx rewriters with ocamlfind, ppx rewriters are executed as separate processes
- all `[@@deriving]` plugins must be executed as part of the same process
- as a result `[@@deriving]` plugins are distributed as plugins rather than executables, and loaded into a single process via a dynamic driver
- this driver is provided by `ppx_deriving`

On the other hand, `dune` statically links all the rewriters together into a single executable. So the dynamic driver provided by `ppx_deriving` is not needed.

rixed:
>> opam outputs that js_of_ocaml-ppx_deriving_json requires ppx_deriving and fails
> I assume you meant to write ocamlfind, not opam.

Indeed.

> Do you think a more descriptive error message would address your issue ?

A more descriptive error message from ocamlfind? Not really, from what I can tell ocamlfind is doing its job properly. According to my current understanding this is a dependency issue, and opam is the dependency manager.

If I understand correctly, a program using js_of_ocaml-ppx_deriving_json and dune will build properly because dune will do the required magic with no need for ppx_deriving, whereas a program using js_of_ocaml-ppx_deriving_json but that does not use dune for building will require ppx_deriving. If this is correct, then I can think of only two solutions:

1. include ppx_deriving in js_of_ocaml-ppx_deriving_json although it may not be actually needed if the user uses dune
2. add a variants of the package, say js_of_ocaml-ppx_deriving_json-dune without the dependency to ppx_deriving.

But given ppx_deriving is not such a large dependency I would favor the first of these options.

Surely, this is not that simple?

diml:
BTW, `META` files are being superseded by a new mechanism in the compiler directly. See [this accepted RFC](https://github.com/ocaml/RFCs/pull/7). On the Dune side, we plan to:

- stop reading `META` files from Dune 3
- stop generating `META` files from Dune 4

So all this stuff is basically going to go away. To help adapting to this change, we could provide a functionality in Dune to build static ppx drivers and list the runtime dependencies.

hhugo:
> include ppx_deriving in js_of_ocaml-ppx_deriving_json although it may not be actually needed if the user uses dune

I moved await from depending on ppx_deriving because it was a pain to maintain and was often the reason for the package to be uninstallable.

> add a variants of the package, say js_of_ocaml-ppx_deriving_json-dune without the dependency to ppx_deriving.

Given the popularity of dune, I would rather add a variant to support the non-dune workflow and have the current package work with dune out of the box.

> A more descriptive error message from ocamlfind? Not really, from what I can tell ocamlfind is doing its job properly. According to my current understanding this is a dependency issue, and opam is the dependency manager.

I was thinking about customizing the error message in the META saying that if you want to use a "plain ocamlfind workflow", you need to explicitly depend on ppx_deriving.

One could also remove this hack altogether and force users to deal with the mechanism used to apply derivers.

@rixed, can you say more about your setup and workflow ?

rixed:
> Given the popularity of dune, I would rather add a variant to support the non-dune workflow and have the current package work with dune out of the box.

That works as well of course, wtv.

> I was thinking about customizing the error message in the META saying that if you want to use a "plain ocamlfind workflow", you need to explicitly depend on ppx_deriving.

Not a big improvement I'm afraid.
Why not have an opam configuration variable that would trigger a conditional dependency? Mmm, but the same user could use dune for some projects and not others in the same switch (or why not in the same project).

> can you say more about your setup and workflow ?

Not sure I understand the question. My workflow is to use opam to install the libs that I'm directly using and count on it to transitively install all the libs and tools required by those libs that I'm directly using ;)

If the question was in fact "why don't you just use dune", then, first, that would not help other people who might hit the same issue, and second... well let's not get me started on the topic of build tools ;)

hhugo:
> use opam to install the libs that I'm directly using and count on it to transitively install all the libs and tools required by those libs that I'm directly using.

So removing the backward compatibility layer to force you to "directly" use ppx_deriving would work as well ? :D

I guess we could create yet another package (virtual?). But that will not solve anything for all other ppx derivers based on ppxlib.

kit-ty-kate:
Closing this. Feel free to propose something else in a future PR. In this state I don't think the PR should be merged in opam-repository as it is not strictly a dependency. I encourage to move the discussion to the `js_of_ocaml` or `dune` issue-trackers:
* https://github.com/ocsigen/js_of_ocaml/issues/
* https://github.com/ocaml/dune/issues/

thanks nonetheless.

PR #15872, opened 2020-02-19
[DRAFT] release preparation for ppx_deriving.4.4.1 with 4.10 support

gasche:


kit-ty-kate:
For the record, I already checked the revdeps (though without tests, just compiling), with OCaml 4.09 and 4.10. The CI here only checks OCaml 4.09 with tests and given the patch sets I don't think this check does anything but check for this change https://github.com/ocaml-ppx/ppx_deriving/commit/457d3916291e44528e46b6ac3e860a69d5448a31

gasche:
@kit-ty-kate so your advice is to release 4.4.1 right away?

kit-ty-kate:
With your refactoring I'm not so sure anymore. The semantics is not the same as my original patch. I'll comment on the original PR.

PR #15876, opened 2020-02-20
First set of constraints for result 1.5

aantron:
These are the constraints due to https://github.com/ocaml/opam-repository/pull/15853, for packages that already failed due to `open Result` on 4.08+ shadowing other identifiers. See the [revdeps logs](https://ci.ocaml.org/ocaml/opam-repository/pr/15853?test=2%20Revdeps) of each package from this PR for the error message.

Most packages that failed failed with a different error, which is [addressed](https://github.com/janestreet/result/commit/b0b9cd6e9cce8d2b7ad11127a6135540ec3f8a92) in `result` `master`, and `result` 1.5 will be retagged to include the fix. I'll do that after this PR is merged, to decrease noise in the revdeps logs.

Of the packages constrained by this PR, only `datakit-client`, `depyt`, `ocaml-version`, `promise`, `tsdl-image` need patching upstream. The rest of the packages have newer releases which are not affected, typically because they no longer support 4.02 and no longer depend on `result`. I will send PRs to the affected packages separately.

### DIFF
--- a/packages/alcotest/alcotest.0.8.0/opam
+++ b/packages/alcotest/alcotest.0.8.0/opam
@@ -17,7 +17,7 @@ depends: [
   "jbuilder" {>= "1.0+beta10"}
   "fmt" {>= "0.8.0"}
   "astring"
-  "result"
+  "result" {< "1.5"}
   "cmdliner"
 ]
 synopsis:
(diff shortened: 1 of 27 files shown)


kit-ty-kate:
Thanks!

PR #15877, opened 2020-02-20
Add {learn-ocaml, learn-ocaml-client} 0.12

erikmd:
[learn-ocaml](https://github.com/ocaml-sf/learn-ocaml) 0.12 has just been released by @yurug, and this PR adds the 2 related opam packages (server and CLI client).

### DIFF
--- a/packages/learn-ocaml-client/learn-ocaml-client.0.12/opam
+++ b/packages/learn-ocaml-client/learn-ocaml-client.0.12/opam
@@ -20,7 +20,7 @@ depends: [
   "cohttp" {>= "1.0.0"}
   "cohttp-lwt-unix" {>= "1.0.0"}
   "digestif" {>= "0.7.1"}
-  "dune" {build & >= "1.1.1"}
+  "dune" {>= "1.1.1"}
   "ezjsonm"
   "lwt" {>= "3.2.0" & < "4.0.0"}
   "lwt_ssl"
@@ -34,7 +34,7 @@ depends: [
   "ppx_tools"
 ]
 build: [
-  ["dune" "build" "@install" "-p" name]
+  ["dune" "build" "@install" "-p" name "-j" jobs]
 ]
 synopsis: "The learn-ocaml client"
 description: """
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
Is there a reason why these packages have such tight constraints? Using `= <version>` is not really good practice as there might be future bug fixes and it's annoying for people trying to use your package on other versions of OCaml (currently they cannot be installed with OCaml >= 4.07). Also this prevent other packages to be installed in the same switch.

Apart from that, `dune` should not be tagged as a build dependency (see #14266). Could you remove this? Also, could you add `"-j" jobs` to the dune build command. By default dune runs in parallel but users might want to reduce the number of jobs used.

erikmd:
@kit-ty-kate , thanks a lot for your your feedback.

Cc @yurug FYI

> `dune` should not be tagged as a build dependency (see #14266). Could you remove this? Also, could you add `"-j" jobs` to the dune build command. 

done, thanks.

> Is there a reason why these packages have such tight constraints?

Partly yes (see below) but indeed most of `= <version>` constraints seemed unnecessarily tight.
I've pushed a new relaxed version that locally compile on my machine and I'm also going to submit these changes upstream with a PR for learn-ocaml master.

> (currently they cannot be installed with OCaml >= 4.07)

AFAICT, learn-ocaml is currently tied with `ocaml 4.05.0`, notably because it incorporates some features that rely on `OCaml_405.Ast` (and the learn-ocaml CI test-build only runs with ocaml 4.05 anyway [cf. Dockerfile](https://github.com/ocaml-sf/learn-ocaml/blob/ffa4e54f0cbb47e28c8bb6569ca65f7795bae884/Dockerfile#L12) and [.ci-macosx.sh](https://github.com/ocaml-sf/learn-ocaml/blob/ffa4e54f0cbb47e28c8bb6569ca65f7795bae884/.ci-macosx.sh#L6)); maybe other ocaml versions could be supported later, but not for the current 0.12 release anyway…

So in the current version of this PR, apart from ocaml, only two `"="` dependencies remain:
```
"decompress" {= "0.8.1"}
"ocplib-json-typed" {= "0.6"}
```
(because of some non-backward compatibility with their next version…)

Kind regards

erikmd:
Hi @kit-ty-kate , do you believe you'll have the time to take another look at this PR before Monday 2019-02-24? (as we're going to use `learn-ocaml-client` in a lab session on Monday morning…)

kit-ty-kate:
I'm just waiting for the CI to finish and I'll have a final look by tonight

kit-ty-kate:
Looks good. Thanks!

PR #15879, opened 2020-02-20
[release] rpc metapackage (5.9.0, 6.0.0, 6.1.0, 7.0.0)

psafont:
Without the metapackage being available it's difficult to develop on the package while using this repo as the first one in the active opam switch, see https://github.com/mirage/ocaml-rpc/issues/150

- Project page: https://github.com/mirage/ocaml-rpc
- Documentation: 

Changes: 
7.0.0
- Add basic support for JSON-RPC notifications (@vycastor)

6.1.0
- opam: updated bounds on a more conservative basis
- travis: tests more compilers
- tests: disable useless-object-inheritance on pylint checks
- pythongen: generate python2-3 compatible bindings
- Add ISC license
- Incremented the upper bound for async's version.
- Added lower bound for js_of_ocaml in related .opam file.
- Fixed compilation issue with js_of_ocaml 3.5.0 and 3.5.1.
- opam: remove the 'build' directive on dune dependency
- opam: remove unnecessary flag
- port to dune

6.0.0
- Switch to ppxlib from ppx_deriving

5.9.0
- Optional named params fix

### DIFF
--- a/packages/rpc/rpc.5.9.0/opam
+++ b/packages/rpc/rpc.5.9.0/opam
@@ -19,8 +19,8 @@ depends: [
   "ppx_deriving_rpc" {= version}
   "async" {>= "v0.9.0" & <= "v0.11.0"}
   "lwt"
-  "alcotest" {with-test}
-  "alcotest-lwt" {with-test}
+  "alcotest" {with-test & < "1.0.0"}
+  "alcotest-lwt" {with-test & < "1.0.0"}
 ]
 conflicts: [ "core" {< "v0.9.0"} ]
 build: ["jbuilder" "runtest" "-p" name] {with-test}
--- a/packages/rpc/rpc.5.9.0/opam
+++ b/packages/rpc/rpc.5.9.0/opam
@@ -12,7 +12,7 @@ tags: [
 ]
 depends: [
   "ocaml" {>= "4.04.0"}
-  "jbuilder" {build}
+  "jbuilder"
   "rpclib" {= version}
   "rpclib-async" {= version}
   "rpclib-lwt" {= version}
@@ -23,7 +23,7 @@ depends: [
   "alcotest-lwt" {with-test & < "1.0.0"}
 ]
 conflicts: [ "core" {< "v0.9.0"} ]
-build: ["jbuilder" "runtest" "-p" name] {with-test}
+build: ["jbuilder" "build" "-p" name "-j" jobs]
 description: """
 `ocaml-rpc` is a library that provides remote procedure calls (RPC)
 using XML or JSON as transport encodings, and multiple generators
--- a/packages/rpc/rpc.5.9.0/opam
+++ b/packages/rpc/rpc.5.9.0/opam
@@ -1,0 +1,44 @@
+opam-version: "2.0"
+synopsis: "A library to deal with RPCs in OCaml - meta-package"
+maintainer: "thomas@gazagnaire.org"
+authors: "Thomas Gazagnaire, Jon Ludlam"
+homepage: "https://github.com/mirage/ocaml-rpc"
+bug-reports: "https://github.com/mirage/ocaml-rpc/issues"
+dev-repo: "git://github.com/mirage/ocaml-rpc"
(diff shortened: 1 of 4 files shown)


psafont:
@mseri travis is failing for 5.9.0 and 6.0.0 because dependencies are not compatible with the ocaml versions used for the tests, this is not surprising to me, but it'd be nice to have confirmation from you.

The tests for 6.1.0 are failing because `pylint` is missing, not much that can be done unless the environment is changed to include it.

mseri:
Thanks for the PR. Your analysis is correct. I think we can simply disable the tests in all of them. They have been tested already after all

kit-ty-kate:
@psafont the tests require `pylint`. We can add a new `conf-pylint` package so that you can require it during tests. I'll do that now

mseri:
It is not that easy I think. Some tests require the python2 version and the more recent ones the pyhon3

kit-ty-kate:
then I could add two packages instead of one (`python2-pylint` and `python3-pylint`), do you think that would fix the issue?

mseri:
Modulo distributions that only have one of the two, I believe it would. I am not sure it is worth the hassle, but let’s try

kit-ty-kate:
The build command was wrong and the packages didn't install anything. Could these fixes be returned upstream?

I also removed the tests as it seems too much of a hassle. This should be good to go once CI is done with it

mseri:
They were already good upstream https://github.com/mirage/ocaml-rpc/blob/master/rpc.opam I've just added the missing odoc dependency. Thanks again. I agree with the decision of turning off the tests, python dependencies are a bit of a hell still with the 2 to 3 transition

kit-ty-kate:
mh, it seems that `md2mld` is also needed all the time, not just for doc generation. Is this normal?
```
#=== ERROR while compiling rpc.6.1.0 ==========================================#
# context              2.0.6 | linux/x86_64 | ocaml-base-compiler.4.09.0 | pinned(https://github.com/mirage/ocaml-rpc/archive/v6.1.0.tar.gz)
# path                 ~/.opam/4.09/.opam-switch/build/rpc.6.1.0
# command              ~/.opam/4.09/bin/dune build -p rpc -j 72
# exit-code            1
# env-file             ~/.opam/log/rpc-23-2f89d0.env
# output-file          ~/.opam/log/rpc-23-2f89d0.out
### output ###
# File "docs/dune", line 7, characters 10-21:
# 7 |    (run %{bin:md2mld} -min-header 3 %{deps})
#               ^^^^^^^^^^^
# Error: Program md2mld not found in the tree or in PATH
#  (context: default)
```

mseri:
That's weird... it may be changed recently, it used to be called only to generate the documentation

mseri:
It is explicitly marked for documentation: https://github.com/mirage/ocaml-rpc/blob/master/docs/dune#L12

kit-ty-kate:
weird indeed. cc @rgrinberg @diml?

rgrinberg:
`.mld` files are installed as part of the package (see the .install file generated), so it's expected they're included here. I'm afraid you will need to add a dependency on the package that provides this binary or simply promote the generated artifact.

rgrinberg:
@avsm @diml this issue reminds me of our discussion at the retreat regarding the correct way to split packages so that they're as usable in as many contexts as possible. There was a remark that packages should be split along the lines laid out by Linux package managers, e.g `foo` to provide the binaries, `foo-dev` to provide the libraries. The issue here would be solved if documentation was packaged separately as well: `foo-doc`. Additionally, a `foo-test` package would help for cases when the test dependencies require additional dependencies.

I recall that in the retreat, we concluded that the user should factor their packages manually. Perhaps it's the wrong way after all as it will quickly get tedious and upstream packager will have to repeat this work. Maybe dune should be a little smarter do this factoring automatically to generate multiple packages?

I realize that this is in conflict with opam's `with-test` and `with-doc` features, but these have caused quite a few problems for me and other users. They seem to be anti features like optional dependencies at this point. Anyway, this is just something to think about. I don't think we should take any action here.

kit-ty-kate:
Looks good now. Thanks!

PR #15881, opened 2020-02-20
ppx_deriving_yojson.3.5.2 with support for OCaml 4.10

gasche:
This is a minor release with @kit-ty-kate's work for 4.10 support.


kit-ty-kate:
Thanks a lot!

PR #15882, opened 2020-02-21
ocaml-cgi v0.10

rixed:


kit-ty-kate:
Thanks! Could you return the fix upstream? (except probably for the version field, it's redundant for opam-repository but welcome in the source repository). The remove field however is unnecessary since opam 2.0.

PR #15884, opened 2020-02-21
OCaml 4.10.0 release

Octachron:
This release includes the usual (aka already existing variants) except force-safe-string.
The only difference with the second release candidate is some configuration for bsd using libbfd (that were not working at all previously).


kit-ty-kate:
Thanks a lot!

PR #15893, opened 2020-02-23
Last constraint for result 1.5

aantron:
See [this log](https://ci.ocaml.org/log/saved/docker-run-0d3f187e951ac8341a63ccef7be9ec36/527837989c4a5d7f3cba38e3e9a09de009073fdd). After this, I think #15853 can be merged without a rebase, as probably nothing new can be learned from another set of revdeps builds.


kit-ty-kate:
Thanks

PR #15909, opened 2020-02-27
'zed' 2.0.x doesn't support recent 'result' 1.5

mjambon:
Hello there,

This fixes an immediate incompatibility between zed 2.0 and the recently released result.1.5. This also adds an explicit dependency on the result package.

The error is with ocaml 4.08.0. I didn't check other versions.
```
# Error: The implementation src/zed_string.ml
# [...]
#          val compare_index :
#            t ->
#            ('a, 'b) result ->
#            ('a, 'b) result ->
#            ok:('a -> 'a -> int) -> error:('b -> 'b -> int) -> int
#        is not included in
#          val compare_index : t -> index -> index -> int
#        File "src/zed_string.mli", line 184, characters 0-46:
#          Expected declaration
#        File "src/zed_string.ml", line 454, characters 6-19:
#          Actual declaration
```

cc @diml
(I'm looking into a fix for the error)

### DIFF
--- a/packages/zed/zed.2.0.1/opam
+++ b/packages/zed/zed.2.0.1/opam
@@ -12,6 +12,7 @@ depends: [
   "camomile" {>= "1.0.1"}
   "react"
   "charInfo_width" {>= "1.1.0" & < "2.0~"}
+  "result" {< "1.5"}
 ]
 build: [
   ["dune" "subst"] {pinned}
--- a/packages/zed/zed.2.0.2/opam
+++ b/packages/zed/zed.2.0.2/opam
@@ -12,6 +12,7 @@ depends: [
   "camomile" {>= "1.0.1"}
   "react"
   "charInfo_width" {>= "1.1.0" & < "2.0~"}
+  "result" {< "1.5"}
 ]
 build: [
   ["dune" "build" "-p" name "-j" jobs]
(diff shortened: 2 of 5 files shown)


diml:
I have only been following this from far, I believe this PR is correct but I'll let @aantron or @kandu confirm.

aantron:
Looks good to me. It probably doesn't matter, but I would have used a constraint `< "1.5"` rather than `<= "1.4"`. We won't have a 1.4.1 release of `result`, with very high probability, but in case we ever did, the `<` constraint would be the one that remained correct.

mjambon:
@aantron I changed the constraint to "< 1.5" as you suggested. Waiting for the test results.

kandu:
LGTM

kandu:
Just noticed that #15906 which had been merged already constrained these opam files with upper-bound.

kit-ty-kate:
Thanks

kit-ty-kate:
oops sorry I didn't see the first comment, reverting to avoid duplicated constraints (I double-checked and all the constraints were already there)

eponier:
Actually, the fix in this PR was correct, while the one in the other seems not. Indeed, the constraint `< 1.5.0` seems to be passed by `result.1.5`, thus not resolving the problem. As a side note, the restriction does not seem necessary for `zed.1.6`.

kit-ty-kate:
arf, indeed. My bad. I un-reverted this PR and reverted the PR by anil. Thanks for spotting this.

PR #15913, opened 2020-02-27
New package metapp.0.1.0

thierry-martinez:
`metapp` is a PPX rewriter that provides a `[%meta ...]` extension, where the dots `...` are arbitrary OCaml expressions that are substituted at compile-time by the AST nodes they evaluate into.
These expressions build AST nodes either by quoting some code directly, or by using `compiler-libs` ([`Parsetree`], [`Ast_helper`], ...).


kit-ty-kate:
Thanks!

PR #15916, opened 2020-02-28
added packages/parany/parany.9.0.0

UnixJunkie:
There is now a Parany.Parmap module.
This module exposes an interface very
close to some of the functionalities offered by parmap.


kit-ty-kate:
Thanks!

Issue #15344, opened 2019-11-20
fail to build 4.0xxx+musl+static+flambda


aryx:
Here is the trace:
<pre><font color="#8AE234"><b>pad@yrax</b></font>:<font color="#729FCF"><b>~</b></font>$ opam switch create 4.07.1+musl+static+flambda

<font color="#06989A">&lt;&gt;&lt;&gt;</font> <b>Gathering sources</b> <font color="#06989A">&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;</font>
[<font color="#4E9A06">ocaml-variants.4.07.1+musl+static+flambda</font>] found in cache

<font color="#06989A">&lt;&gt;&lt;&gt;</font> <b>Processing actions</b> <font color="#06989A">&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;&lt;&gt;</font>
<font color="#4E9A06">∗</font> installed <b>base-bigarray</b>.base
<font color="#4E9A06">∗</font> installed <b>base-threads</b>.base
<font color="#4E9A06">∗</font> installed <b>base-unix</b>.base
<font color="#CC0000">[ERROR]</font> The compilation of ocaml-variants failed at &quot;/home/pad/.opam/opam-init/hooks/sandbox.sh build make -j7 world&quot;.

<font color="#CC0000">#===</font> <b>ERROR</b> <b>while compiling ocaml-variants.4.07.1+musl+static+flambda</b> <font color="#CC0000">==========#</font>
[...]

PR #15927, opened 2020-02-29
musl+static+flambba ocaml: disable shared libraries in compiler config

avsm:
this fixes relocation errors when building on some distributions
(ocaml/ocaml#9131 #15344)

### DIFF
--- a/packages/ocaml-variants/ocaml-variants.4.08.0+musl+static+flambda/opam
+++ b/packages/ocaml-variants/ocaml-variants.4.08.0+musl+static+flambda/opam
@@ -23,11 +23,13 @@ build: [
   [
     "./configure"
     "--prefix=%{prefix}%"
-    "CC=musl-gcc -Os"
+    "CC=musl-gcc"
+    "CFLAGS=-Os"
     "ASPP=musl-gcc -c"
     "--enable-flambda"
     "LIBS=-static"
     "--disable-graph-lib"
+    "--disable-shared"
   ] {os = "openbsd" | os = "freebsd" | os = "macos" | os = "linux"}
   [make "-j%{jobs}%" {os != "cygwin"} "world"]
   [make "-j%{jobs}%" {os != "cygwin"} "world.opt"]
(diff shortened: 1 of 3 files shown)


kit-ty-kate:
Thanks

copy:
@avsm This reverts https://github.com/ocaml/opam-repository/pull/14664 and thus breaks `ppx_deriving` on static switches again (https://github.com/ocaml/opam-repository/issues/14663). The problem is that Dynlink is disabled when `--disable-shared` is given and  `ppx_deriving` uses Dynlink.

Also note that musl+static+flambda doesn't create static executables by default (`-ccopt -static` needs to be passed to ocamlopt), so in https://github.com/ocaml/opam-repository/pull/14664 we created a musl+flambda switch, which may also be affected by this.

As a side note, I create my switches using `CC=/opt/x86_64-linux-musl-native/bin/gcc opam switch create 4.10.0+flambda` these days, which is not broken by this change.

mseri:
I think we shuld put a note on the image meaning/content and the hint about the switch shared by @copy somewhere reachable, like the switch itself or a page in ocamlverse. Any thought?

avsm:
Sorry about that @copy - I agree that putting a clearer marker on these description files would be very helpful @mseri.

But how do we fix this? There's a build error on PIE distributions if `.so` files are built and relocated currently.

copy:
> But how do we fix this? There's a build error on PIE distributions if .so files are built and relocated currently.

I don't really understand the issue, in particular why it happens with musl-gcc but not regular gcc. It seems to be independent of `-static`, i.e. `./configure CC=musl-gcc && make` fails on Ubuntu, so I think this should be fixed in the compiler. If not, we could at least add the `--disable-shared` flag only on PIE distributions.

avsm:
I tracked this down. Turns out that the version of musl in Ubuntu 18.04 (1.1.19) has a bug with static linking, so I backported 1.1.24 and created a PPA with it.

Doing

```
# add-apt-repository -y ppa:avsm/musl
# apt-get -y install musl-tools
# opam switch create foo ocaml-variants.4.10.0+musl+static+flambda
```

Now works on ubuntu 18.04. I'll test it a bit more and then add a messages to the opam file to point to these instructions on failure.  Incidentally ubuntu-18.04 is useful since its the default Linux VM for GitHub Actions, so now we can [build static binaries from there too](https://github.com/ocamllabs/duniverse/runs/512220645?check_suite_focus=true).

PR #15934, opened 2020-03-02
[new release] merlin (3.3.4-4.10preview1)

let-def:

##### CHANGES:

Mon Mar  2 14:26:32 CET 2020

This is a preview release that adds support for OCaml 4.10.
Short-path is disabled.  Other versions of OCaml are not supported.


kit-ty-kate:
Thanks a lot!!

let-def:
Thank you :)

Octachron:
@let-def , should I announce that there is a version of merlin available for 4.10.0 ? Or do you think it is better to wait?

let-def:
It is fine to announce the release, as long as it is clear that it is a preview and not a definitive version. (And maybe that short-path does not work?)

Octachron:
Ok, let's announce the preview then.

PR #15939, opened 2020-03-03
Fix the package description for cduce 0.5.5.

fpottier:
cduce does not compile with ocamlnet 4.0.1 or later.
It does compile with ocamlnet 3.7.7.
The error message is:
```
  File "parser/cduce_netclient.ml", line 13, characters 6-38:
  Error: Unbound module Http_client
```


kit-ty-kate:
Thanks!

PR #15942, opened 2020-03-04
plplot: update depexts

mseri:


mseri:
The failure I see are the ones I would expect

kit-ty-kate:
Thanks!

PR #15945, opened 2020-03-04
plplot: add also centos

mseri:
I had missed one


mseri:
I am going to merge and cleanup the pr list This works on centos < 8, for some reasons the plplot rpms are broken in the latest centos with epel

PR #15947, opened 2020-03-05
transept.0.1.0

d-plaindoux:
The first version of Transept an OCaml modular and generalised parser combinator library.

### DIFF
--- a/packages/transept/transept.0.1.0/opam
+++ b/packages/transept/transept.0.1.0/opam
@@ -1,5 +1,4 @@
 opam-version: "2.0"
-version: "0.1.0"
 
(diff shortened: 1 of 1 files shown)


packages/transept/transept.0.1.0/opam
```diff
@@ -0,0 +1,36 @@
+opam-version: "2.0"
+version: "0.1.0"
```
mseri:
```suggestion
```

This is redundant, opam will override it with what infers from the opam file path in the repo

mseri:
Looks good! Thank you very much

PR #15958, opened 2020-03-06
New release redirect 0.1.1

thierry-martinez:
- Fix backtrace handling


kit-ty-kate:

thierry-martinez:
> Thanks! Could you return the fix upstream?

Done! Thank you very much.

PR #15961, opened 2020-03-08
Liquidsoap: ignore errors while installing bash completion files

toots:
Hi there!

Apologize but it looks like the previous update on liquidsoap did not fix all cases of the issue. This one should hopefully put it to bed for good this time. I've tested it on Debian/testing where it was still occurring.

Thanks for your consideration!


kit-ty-kate:
Thanks

PR #15962, opened 2020-03-08
[new release] thrift (0.10.0)

vbmithr:
OCaml bindings for the Apache Thrift RPC system

- Project page: <a href="https://github.com/vbmithr/ocaml-thrift-lib">https://github.com/vbmithr/ocaml-thrift-lib</a>
- Documentation: <a href="https://vbmithr.github.io/ocaml-thrift-lib/doc">https://vbmithr.github.io/ocaml-thrift-lib/doc</a>

##### CHANGES:

* Sync with upstream
* Make the code safe-string ready (@c-cube)
* build: switch to dune (@c-cube)

### DIFF
--- a/packages/thrift/thrift.0.10.0/opam
+++ b/packages/thrift/thrift.0.10.0/opam
@@ -10,7 +10,7 @@ depends: [
   "dune" { >= "1.1" }
 ]
 depexts: [
-  ["thrift"] {os = "archlinux"}
+  ["thrift"] {os-family = "arch"}
   ["thrift-compiler"] {os-family = "debian"}
   ["devel/thrift"] {os-family = "bsd"}
 ]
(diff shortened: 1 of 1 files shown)


packages/thrift/thrift.0.10.0/opam
```diff
@@ -0,0 +1,32 @@
...
+doc: "https://vbmithr.github.io/ocaml-thrift-lib/doc"
+depends: [
+  "ocaml" { >= "4.03.0" }
+  "dune" { >= "1.1" }
+]
+depexts: [
+  ["thrift"] {os = "archlinux"}
+  ["thrift-compiler"] {os-family = "debian"}
+  ["devel/thrift"] {os-family = "bsd"}
+]
```
kit-ty-kate: The depexts don't seem to be strictly needed. I could only find use in the examples. Is this normal?

vbmithr: You are right, yes. I think it was necessary for CI to run. Anyway, it's probably not a bad idea to have thrift installed to use the library…

kit-ty-kate:
I guess it can always be modified later. Thanks!

kit-ty-kate:
Could you also return the fix upstream?

vbmithr:
Yes, will do it now.

PR #15963, opened 2020-03-08
Package glfw-ocaml (3.3~rc2)

SylvainBoilard:
CHANGES:

* Use a single type for window hints and attributes.


SylvainBoilard:
Hi,

The build fails because the libglfw3 version is only 3.1.2 on Ubuntu 16. The MacOS build has version 3.3.2 and passes. I think this can be merged.

kit-ty-kate:
Thanks

PR #15959, opened 2020-03-07
Reason 3.6.0

anmonteiro:

PR #15964, opened 2020-03-08
[new release] spin (0.5.0)

tmattio:
Project scaffolding tool and set of templates for Reason and OCaml

- Project page: <a href="https://github.com/tmattio/spin">https://github.com/tmattio/spin</a>
- Documentation: <a href="https://tmattio.github.io/spin">https://tmattio.github.io/spin</a>

## Added

- Before generating a template, Spin will check if the user has all the dependencies installed and exit gracefully if not. (by [@citizen428](https://github.com/citizen428))
- Add a `condition` field in the `postinstall` stanza for the templates.

## Changed

- Print a warning when the update of the official templates failed, but continue the execution.
- Do not update the official templates when running new with a local path or a git repository.
- Remove some runtime dependencies to reduce Spin's binary size.

## Fixed

- Trying to generate a file that already exists now raises an error instead of overwritting the file.
- Fix to output a proper message when no generator exist for the current project.
- Exit gracefully when failing to download a git repository or the initial templates for the first time. (by [@citizen428](https://github.com/citizen428))

### DIFF
--- a/packages/spin/spin.0.5.0/opam
+++ b/packages/spin/spin.0.5.0/opam
@@ -14,10 +14,6 @@ depends: [
   "ocaml" {>= "4.06.0" & < "4.10.0"}
   "dune" {>= "2.0"}
   "odoc" {with-doc}
-  "dune-release" {dev}
-  "merlin" {dev}
-  "ocamlformat" {dev}
-  "utop" {dev}
   "reason" {build}
   "base"
   "stdio"
--- a/packages/spin/spin.0.5.0/opam
+++ b/packages/spin/spin.0.5.0/opam
@@ -11,7 +11,7 @@ homepage: "https://github.com/tmattio/spin"
 doc: "https://tmattio.github.io/spin"
 bug-reports: "https://github.com/tmattio/spin/issues"
 depends: [
-  "ocaml" {>= "4.06.0" & < "4.10.0"}
+  "ocaml" {>= "4.06.0"}
   "dune" {>= "2.0"}
   "odoc" {with-doc}
   "reason" {build}
--- a/packages/spin/spin.0.5.0/opam
+++ b/packages/spin/spin.0.5.0/opam
@@ -1,0 +1,53 @@
+opam-version: "2.0"
+synopsis:
+  "Project scaffolding tool and set of templates for Reason and OCaml"
+description: """
+Project scaffolding tool and set of templates for Reason and OCaml
+"""
+maintainer: ["Thibaut Mattio"]
+authors: ["Thibaut Mattio"]
+license: "MIT"
+homepage: "https://github.com/tmattio/spin"
+doc: "https://tmattio.github.io/spin"
+bug-reports: "https://github.com/tmattio/spin/issues"
+depends: [
+  "ocaml" {>= "4.06.0" & < "4.10.0"}
+  "dune" {>= "2.0"}
+  "odoc" {with-doc}
+  "dune-release" {dev}
+  "merlin" {dev}
+  "ocamlformat" {dev}
+  "utop" {dev}
+  "reason" {build}
+  "base"
+  "stdio"
+  "cmdliner"
+  "fileutils"
+  "jingoo"
+  "lwt"
+  "ppx_sexp_conv"
+  "sexplib"
+]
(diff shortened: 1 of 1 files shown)


packages/spin/spin.0.5.0/opam
```diff
@@ -0,0 +1,49 @@
...
+authors: ["Thibaut Mattio"]
+license: "MIT"
+homepage: "https://github.com/tmattio/spin"
+doc: "https://tmattio.github.io/spin"
+bug-reports: "https://github.com/tmattio/spin/issues"
+depends: [
+  "ocaml" {>= "4.06.0" & < "4.10.0"}
+  "dune" {>= "2.0"}
+  "odoc" {with-doc}
+  "reason" {build}
```
kit-ty-kate: I can't find any use of `reason` in the source code. Is this really needed?

packages/spin/spin.0.5.0/opam
```diff
@@ -0,0 +1,49 @@
...
+Project scaffolding tool and set of templates for Reason and OCaml
+"""
+maintainer: ["Thibaut Mattio"]
+authors: ["Thibaut Mattio"]
+license: "MIT"
+homepage: "https://github.com/tmattio/spin"
+doc: "https://tmattio.github.io/spin"
+bug-reports: "https://github.com/tmattio/spin/issues"
+depends: [
+  "ocaml" {>= "4.06.0" & < "4.10.0"}
```
kit-ty-kate: I'm curious, why is this package not working on OCaml 4.10?

packages/spin/spin.0.5.0/opam
tmattio: The codebase is in Reason, so this is needed to build the project.

packages/spin/spin.0.5.0/opam
tmattio: The current version of Reason on Opam does not work with OCaml 4.10.
Once https://github.com/ocaml/opam-repository/pull/15959 is merged, I will be able to remove the constraint!

kit-ty-kate:
I removed the dependencies tagged as `{dev}` because they seem to be used for another purpose.
From the documentation (https://opam.ocaml.org/doc/Manual.html#Package-variables):
```
dev: true if this is a development package, i.e. it was not built from a release archive
```
In opam, this only happens if you pin the dev repository. I'm not sure to understand why users who want that would absolutely require merlin, utop, ocamlformat and dune-release.

If this is correct, could you return the fix upstream?

tmattio:
@kit-ty-kate Thanks! Is there any way to specify development dependencies with Opam? I would like to have merlin, utop, etc. installed when I work locally but don't want the users of the package to have to install them.

packages/spin/spin.0.5.0/opam
kit-ty-kate: If you want. Though for the record, opam packages usually avoids adding too many upper-bound constraints as that many back-and-forth between opam-repository and individual devs are not scalable. Our CI ensures that things don't break, we put the appropriate constraints when it does and ping the dev when necessary.

kit-ty-kate:
> @kit-ty-kate Thanks! Is there any way to specify development dependencies with Opam? I would like to have merlin, utop, etc. installed when I work locally but don't want the users of the package to have to install them.

If you want them, you can install them yourself using `opam install merlin utop etc.`.
If you really don't want to call this command for some reason then you can have a local opam file that won't be released in opam-repository called something like `spin+dev.opam` or something like that. Though, usually installing these packages are done once so I'm not sure it is time efficient to write an opam file just for that

packages/spin/spin.0.5.0/opam
kit-ty-kate: indeed. My bad ^^

packages/spin/spin.0.5.0/opam
tmattio: Didn't know, thanks for the explanation, I'll remove the upper bound!

kit-ty-kate: #15959 has now been merged

kit-ty-kate:
Thanks a lot!

PR #15986, opened 2020-03-11
Release x509 0.10.0

hannesm:
CHANGES:
    
* BREAKING mirleft/ocaml-x509#131 use mirage-crypto and mirage-crypto-pk instead of nocrypto
      raise lower OCaml bound to 4.07.0, test 4.10.0 in CI
* BREAKING mirleft/ocaml-x509#131 Certificate.Host_set is now Host.Set, and has pretty-printers
* BREAKING mirleft/ocaml-x509#131 Authenticator.null was removed (fixes mirleft/ocaml-x509#130, suggested by @emillon)
* BREAKING mirleft/ocaml-x509#131 Authenticator.t now requires (non-optional) ~host and ~time
      arguments which may return an option. This allows to more easily spot the
      cases where None is provided. Also, time is (unit -> Ptime.t option) instead
      of Ptime.t -- thus the timestamp at the time of certificate verification is
      used instead of the time of Authenticator construction. Similar changes in
      Validation for verify_chain, verify_chain_of_trust, trust_key_fingerprint,
      trust_cert_fingerprint.
      fixes mirleft/ocaml-x509#130, suggested by @emillon


hannesm:
the second commit adds upper bounds to reverse dependencies -- this PR is not responsible for the build failure of facteur, I'd appreciate if this PR would be merged without me having to fix all potential build issues. thanks.

dinosaure:
Thanks, this is on my current process release!

hannesm:
I looked through the CI failures, and they are unrelated to this PR. in my opinion, this is good to be merged.

kit-ty-kate:
Thanks a lot!

PR #15989, opened 2020-03-12
Add a conflicts clause to menhirSdk and menhirLib.

fpottier:
So that they will refuse to coexist with an old menhir package (which contains its own
versions of menhirLib and menhirSdk).


kit-ty-kate:
Thanks!

PR #15993, opened 2020-03-12
[new release] decompress and rfc1951 (1.1.0)

dinosaure:
Implementation of Zlib and GZip in OCaml

##### CHANGES:

- add GZip support (@dinosaure, @copy, @hcarty, mirage/decompress#79)
- **breaking changes**, `Higher` returns a `result` value instead to raise an exception (@dinosaure, @copy, mirage/decompress#80)
- protect Zlib decoder on multiple _no-op_ calls of `decode`
- test when we generate an empty zlib flow
- fix a bug on the DEFLATE layer when we must flush bits to avoid integer overflow
- really use the internal continuation of the Zlib state
- delete `fmt` depedency
- update fuzzer
- fix default level on `Zl.Higher`

### DIFF
--- a/packages/decompress/decompress.1.1.0/opam
+++ b/packages/decompress/decompress.1.1.0/opam
@@ -1,0 +1,39 @@
+opam-version: "2.0"
+name:         "decompress"
+maintainer:   "Romain Calascibetta <romain.calascibetta@gmail.com>"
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
Looks good. Could you return the fix upstream?

dinosaure:
Will do, thanks :+1:

PR #15998, opened 2020-03-13
rtop 3.6.0

adamsmd:


packages/rtop/rtop.3.6.0/opam
```diff
@@ -0,0 +1,26 @@
...
+dev-repo: "git://github.com/facebook/reason.git"
+tags: [ "syntax" ]
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+]
+depends: [
+  "ocaml"  {>= "4.03" & < "4.11"}
+  "dune"   {>= "1.4"}
+  "reason"
```
mseri: Does it work for all reason releases or needs to be synced or bounded?

adamsmd: I don't know if it needed to be synced or bounded.  I am just presenting it the same way as the `opam` file for `rtop` 3.5.2 does.

In fact, with the exception of 3.2.0, none of the `opam` files for the different versions of `rtop` contain a bound.

My plan was to open a separate issue addressing the syncing of all versions of `rtop` to their respective versions of `reason`, but I will happily defer to whatever the convention is for this sort of situation.

mseri: I’ll merge as is, please open the issue on the reason repository (or also there). It may be that, like utop works with multiple ocaml versions, rtop also works with multiple versions of reason. I hope that if upper or lower bounds are necessary, it will come up in the issue

PR #16015, opened 2020-03-14
[new release] conduit-mirage, conduit-lwt, conduit, conduit-lwt-unix and conduit-async (2.1.0)

hannesm:
CHANGES:

* port to tls.0.11.0 interfaces which also uses mirage-crypto (mirage/ocaml-conduit#309 @hannesm)
* do not use deprecated ppx sexplib declarations (mirage/ocaml-conduit#309 @avsm)
* replace Appveyor CI with GitHub Actions (mirage/ocaml-conduit#309 @avsm)

### DIFF
--- a/packages/conduit-async/conduit-async.2.1.0/opam
+++ b/packages/conduit-async/conduit-async.2.1.0/opam
@@ -1,0 +1,37 @@
+opam-version: "2.0"
+maintainer: "anil@recoil.org"
+authors: [
+  "Anil Madhavapeddy" "Thomas Leonard" "Thomas Gazagnaire" "Rudi Grinberg"
+]
+license: "ISC"
+tags: "org:mirage"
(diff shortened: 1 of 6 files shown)


hannesm:
the revdeps issue are a bit strange, the `depext` run does not seem to take `conf-pkg-config` into account (i.e. does not install `pkg-config` (which mirage-crypto has a build dependency on), and then complains about `pkg-config` not being present. any idea what's the underlying issue and how to fix this?

avsm:
possibly because the pkg-config is a `{build}` dep. I've pushed a change to see if it helps

hannesm:
@avsm hmm, if this fixes it, this looks like an issue in depext (there are btw various `conf-pkg-config {build}` dependencies in opam-repository)

avsm:
Tested manually and it compiled fine with those revdeps, so I've reverted my change and merging this.

PR #16019, opened 2020-03-15
Release otr 0.3.8

hannesm:
CHANGES:
    
* fix crypt function (0.3.7 used the counter wrong, the 0 should be the other half)


kit-ty-kate:
Thanks!

PR #16021, opened 2020-03-15
[new release] mirage, mirage-runtime, mirage-types-lwt and mirage-types (3.7.5)

hannesm:
CHANGES:

* use mirage-crypto (and mirage-crypto-entropy) instead of nocrypto, also
  tls-mirage and up-to-date conduit


kit-ty-kate:
Thanks!

PR #16024, opened 2020-03-15
Added lambda-streams

Risto-Stevcev:


packages/lambda_streams/lambda_streams.0.1.1/opam
```diff
@@ -0,0 +1,42 @@
...
+synopsis: "Lambda-based streaming library"
+description: "A simple and principled streaming library based on lambdas"
+maintainer: ["me@risto.codes"]
+authors: ["Risto Stevcev"]
+license: "BSD-3-Clause"
+homepage: "https://github.com/Risto-Stevcev/lambda-streams"
+doc: "https://risto-stevcev.github.io/lambda-streams"
+bug-reports: "https://github.com/Risto-Stevcev/lambda-streams/issues"
+depends: [
+  "ocaml" {>= "4.06.1" & <= "4.10.0"}
```
kit-ty-kate: Is there a reason to stop at OCaml 4.10.0 precisely? If you know you rely on something that will change in OCaml 4.11 it is a reasonable thing to add an upper-bound constraint on `< "4.11"`, but I find very surprising that this package wouldn't be compatible with OCaml 4.10.1 for instance. If this is not the case, would it be ok to remove the upper-bound constraint entirely?

kit-ty-kate:
Oh, looking at the other dependency constraint I can see something is wrong here, the constraints are too restrictive. On one particular switch opam does not allow multiple versions of the same package installed so restrictive constraint (using `=` in particular) is in general a bad idea. Our CI ensure that packages don't break when a new package is added so this check is down on our side.

Could you change your `=` constraints to `>=`?

packages/lambda_streams/lambda_streams.0.1.1/opam
Risto-Stevcev: What would happen if ocaml >= 4.11 introduces a change that would break this library though?

Risto-Stevcev:
I'm guess I'm not too familiar with the opam workflow. If I set the constraints to be for example something like `"dune" {>= "2.2.0"}`, what would happen if dune `3.0.0` introduces a breaking change? 
Would opam download the latest dune version (`3.0.0`) and the package would just break? in which case the end-user would have to manually fix everything. Or would opam get the lowest compatible version and download `2.2.0`? if it gets the latest version then that seems like the constraint would be lying because it might not necessarily work for all future versions.
 
I wanted to originally treat it like semver and do something like `"dune {>= "2.2.0" & < "3.0.0"}`, assuming there that dune would only introduce a breaking change on a major version update. But I think opam complained that it couldn't find `3.0.0` because it didn't exist yet, and I'm not sure if the ocaml compiler versioning follows semver anyway.

avsm:
We introduce the upper bounds at the point at which the incompatible package is merged into this repository.  Dune is a good example -- dune 2.0 could compile most dune 1.x packages, so an upper bound would have been unnecessarily restrictive.  If a future dune 3.0 does break this package, then our CI system will highlight it and we will edit the metadata for lambda-streams to have that upper bound. But for the purposes of this PR, keeping the upper bound empty is just fine.

Risto-Stevcev:
Oh that's cool! thanks for the clarification :slightly_smiling_face: I'll update the PR

packages/lambda_streams/lambda_streams.0.1.1/opam
```diff
@@ -0,0 +1,42 @@
+# This file is generated by dune, edit dune-project instead
+opam-version: "2.0"
+version: "0.1.2"
```
mseri: This os redundant, opam will override it with the value in the path, and in this case emit a warning as they don’t match

Risto-Stevcev: Can you clarify a bit more? do you mean the `opam-version` or the `version` or both? opam will override which value in the path? what doesn't match? Sorry for all the questions, I read the opam and dune docs but I didn't find anything written about the stuff mentioned here.

The `opam-version` is generated by `dune-project`. I added the project version because it was a main stanza for `dune-project` and I was just assuming it's added to the auto-generated `.opam` file for the same reasons as other package managers like npm's `package.json` or rust's `Cargo.toml`

mseri: The version field. Opam infers the version of the package from the path in the opam repository, in this case it gets 0.1.1, and uses that as the package version. Here you set 0.1.2 which should give rise to a warning. To change the version you need to git mv the opam file to `packages/lambda_streams/lambda_streams.0.1.2/opam`

```suggestion
```

mseri: I have double checked. If you open the build logs, currently you read
```
[ERROR] In the opam file for lambda_streams.0.1.1:
          - At /home/opam/.opam/repo/default/packages/lambda_streams/lambda_streams.0.1.1/opam:3:0:
            This file is for version '0.1.1' but its 'version:' field advertises '0.1.2'.
        'version' has been ignored.
```

Risto-Stevcev: Ah I forgot to update the folder name, thanks for catching that. So is it best practices to leave out the version field in `dune-project`/`opam`? or just that you don't necessarily need it for publishing to opam?

mseri: Unless you need it for some specific reason is best left out, otherwise we just delete it when pushing to the opam repo

packages/lambda_streams/lambda_streams.0.1.2/opam
```diff
@@ -0,0 +1,42 @@
+# This file is generated by dune, edit dune-project instead
+opam-version: "2.0"
+version: "0.1.2"
```
mseri:
```suggestion
```

mseri: It’s just metadata, can be deleted here without the need to retag. Opam relies on this file, not the one on the source repository

kit-ty-kate:
Thanks a lot!

PR #16027, opened 2020-03-17
Why3 1.3.0

silene:


silene:
The CentOS failure is as usual: "Error: Unable to find a match: gtksourceview2-devel".

Regarding the reverse dependencies, it is not clear to me what is the proper way to handle them:
1. do nothing;
2. modify Why3's opam file to mark it as conflicting with those packages;
3. modify those package's opam files so that their dependencies are stricter.

kit-ty-kate:
> Regarding the reverse dependencies, it is not clear to me what is the proper way to handle them:
> 
>     1. do nothing;
> 
>     2. modify Why3's opam file to mark it as conflicting with those packages;
> 
>     3. modify those package's opam files so that their dependencies are stricter.

The opam-repository policy is to do 3 unless the issue is on why3's side and the release has to be fixed. Looking at your reply I assume it's not so I'll open a PR to constrain those packages right away.

kit-ty-kate:
The fix is now merged so this one is good to go. Thanks!

silene:
Thanks. I think it would be worth mentioning this policy in `CONTRIBUTING.md`. (Or in some other place.) I looked for it and could not find it.

PR #16028, opened 2020-03-17
+secp256k1-internal.0.1.0

pirbo:
There is already a binding of secp256k1 in the opam repository but
this one is different and providing both is useful.

secp256k1 binds the public functions of libsecp256k1 (the one
exposed in secp256k1.h) and uses the dynamic library installed
system wise. This is very clean and allows an easy integration
in distribution that are piky with code duplication.

secp256k1-internal exposes the internal functions of libsecp256k1.
To do so, it copies the C code and build ocaml wrappers for low
level components. It is used by tezos codebase.


kit-ty-kate:
Thanks! Could you return the fixes upstream?

pirbo:
Done, thank you!

PR #16037, opened 2020-03-18
[new release] spin (0.5.1)

tmattio:
CHANGES:

## Added

- Add versionning for the official templates to ensure updates on the templates don't break old version of Spin.
- Sort `spin ls` result by name.

### Templates

- Rename react to bs-react. All the upcoming Bucklescript templates will be prefixed by bs-.
- Add a lib template that is releasable on Opam.
- Make the cli and ppx templates releasable on Opam.
- Add support for Alcotest as an alternative to Rely in the native templates.
- Add support for Opam as an alternative to Esy in the native templates.
- Stop using Pesy and the dune files manually in each sub-directories.
- Move the test_runner to a subdirectory support in test.
- Extract the Contributing section of the README to a CONTRIBUTING.md file.
- Add a CHANGES.md file that complies with dune-release requirements.
- Use dune-release in the release scripts.
- Remove homebrew Formula from the cli template that was a bad practice and didn’t allow users to update the generated projects.
- Enable Windows in CI/CD of generated projects

### DIFF
--- a/packages/spin/spin.0.5.1/opam
+++ b/packages/spin/spin.0.5.1/opam
@@ -1,0 +1,49 @@
+opam-version: "2.0"
+synopsis:
+  "Project scaffolding tool and set of templates for Reason and OCaml"
+description: """
+Project scaffolding tool and set of templates for Reason and OCaml
+"""
+maintainer: ["Thibaut Mattio"]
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Thanks!

PR #16046, opened 2020-03-19
An Ocamleditor preview release for OCaml 4.10.

vrotaru:
Ocamleditor is a pure OCaml editor originally written by Francesko Tovagliari, but the last release was for `4.01` version of OCaml

Here is a preview of ocamleditor with compiles and runs (with some issues) under OCaml 4.10.
I have Francesco's permission to go ahead and release a  new version on opam, received in a personal communication

### DIFF
--- a/packages/ocamleditor/ocamleditor.1.13.4~preview.4.10/opam
+++ b/packages/ocamleditor/ocamleditor.1.13.4~preview.4.10/opam
@@ -1,0 +1,32 @@
+opam-version: "2.0"
+name: "ocamleditor"
+version: "1.1.34~preview.4.10"
+
+synopsis:
+  "OCamlEditor is a GTK+ source code editor and build tool for OCaml"
+description:
+  """It provides many features to facilitate editing code, accessing API reference
+directly from the editor and compiling projects."""
+
+authors: "Francesco Tovagliari <ftovagliari@gmail.com>"
+maintainer: "Vasile Rotaru <vrotaru.md@gmail.com>"
+
+homepage: "https://github.com/vrotaru/ocamleditor"
+bug-reports: "https://github.com/vrotaru/ocamleditor/issues"
+
+build: [["ocaml" "build.ml" "ocamleditor"]]
+install: [["ocaml" "tools/install.ml" "-prefix" prefix]]
+remove: [["ocaml" "tools/uninstall.ml" "-prefix" prefix]]
+
+depends: [
+  "ocaml" {= "4.10.0"}
+  "ocamlfind" {>= "1.4.0"}
+  "lablgtk" {>= "2.18.0"}
+  "xml-light" {>= "2.2"}
+]
+depopts: [
+  "ocp-indent"
+  "ocurl"
+  "ocamldiff"
+]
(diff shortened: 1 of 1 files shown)


XVilka:
Maybe it worth porting to the [labgtk3](https://github.com/garrigue/lablgtk) as well? GTK 2 is being eliminated slowly from the modern distributions and abandoned for years.

vrotaru:
@XVilka GTK3 would be fine, as well as some king of integration with dune.

But it would have been a distraction, just now.

packages/ocamleditor/ocamleditor.1.13.4~preview.4.10/opam
```diff
@@ -0,0 +1,40 @@
+opam-version: "2.0"
+name: "ocamleditor"
+version: "1.1.34~preview.4.10"
```
mseri:
```suggestion
```
As above

packages/ocamleditor/ocamleditor.1.13.4~preview.4.10/opam
```diff
@@ -0,0 +1,40 @@
+opam-version: "2.0"
+name: "ocamleditor"
```
mseri:
```suggestion
```

This field is redundant. Opam will override it with what it infers from the path and may give rise to errors

packages/ocamleditor/ocamleditor.1.13.4~preview.4.10/opam
```diff
@@ -0,0 +1,40 @@
...
+
+authors: "Francesco Tovagliari <ftovagliari@gmail.com>"
+maintainer: "Vasile Rotaru <vrotaru.md@gmail.com>"
+
+homepage: "https://github.com/vrotaru/ocamleditor"
+bug-reports: "https://github.com/vrotaru/ocamleditor/issues"
+
+build: [["ocaml" "build.ml" "ocamleditor"]]
+install: [["ocaml" "tools/install.ml" "-prefix" prefix]]
+remove: [["ocaml" "tools/uninstall.ml" "-prefix" prefix]]
```
mseri:
```suggestion
```

This is not necessary with opam2. It keeps track of the installed files and uses the information  to uninstall the package

packages/ocamleditor/ocamleditor.1.13.4~preview.4.10/opam
```diff
@@ -0,0 +1,40 @@
...
+
+depends: [
+  "ocaml" {= "4.10.0"}
+  "ocamlfind" {>= "1.4.0"}
```
mseri:
```suggestion
  "ocamlfind" {build & >= "1.4.0"}
```

Is findlib used at runtime or this is just a build dependency?

vrotaru: For simple projects with no external dependencies is not used, but if uses the `ocamlfind` at runtime to find the paths/flags to pass to the ocaml compiler while compiling the changed files.

So, short version: it is not just an build dependency.

PR #9383, opened 2017-06-01
Add mirage-console.2.3.0

djs55:
Signed-off-by: David Scott <dave@recoil.org>

PR #16051, opened 2020-03-20
Back-port OCaml PR #9383

dra27:
This back-ports https://github.com/ocaml/ocaml/pull/9383 to `ocaml-base-compiler` 4.07.0-4.10.0 inclusive. I haven't done anything with `ocaml-variants`, but of course anyone else would be welcome to apply the same pattern to packages there, if needed.

This patch hardens the use of awk in OCaml's build system so that it is no longer affected by the user's  `AWKPATH`

I have tested each of the packages which are updated here, so it may be wise to terminate the CI run!

cc @nobrowser

### DIFF
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.4.07.0/files/12ef11225a5242f41bac8b47e5e1f6b578fd6f33.patch
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.4.07.0/files/12ef11225a5242f41bac8b47e5e1f6b578fd6f33.patch
@@ -1,0 +1,62 @@
+From 12ef11225a5242f41bac8b47e5e1f6b578fd6f33 Mon Sep 17 00:00:00 2001
+From: David Allsopp <david.allsopp@metastack.com>
+Date: Fri, 20 Mar 2020 13:05:44 +0000
+Subject: [PATCH] Merge pull request #9383 from dra27/explicit-awk
+
+Don't assume . in AWKPATH
+
+(cherry picked from commit d4ace8c347a9a4f8baa8f0b1738a5b4436fcd027)
+---
+ stdlib/Compflags   | 2 +-
+ stdlib/Makefile    | 2 +-
+ testsuite/Makefile | 4 ++--
+ 3 files changed, 4 insertions(+), 4 deletions(-)
+
+diff --git a/stdlib/Compflags b/stdlib/Compflags
+index 0bffcebefa4..b62e552aa6c 100755
+--- a/stdlib/Compflags
++++ b/stdlib/Compflags
+@@ -17,7 +17,7 @@
+ case $1 in
+   stdlib.cm[iox]|stdlib.p.cmx)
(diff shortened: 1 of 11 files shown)


kit-ty-kate:
Is this really necessary to backport it right away? Can't we just wait for new versions instead? Is this a security threat?

cc @Octachron

(I tagged it as "do not merge" for now because the CI can't test such PR)

avsm:
While it's an incongruous enough patch, I'd also prefer not modifying pristine OCaml releases with patches without a version bump somewhere to indicate it.

Octachron:
I am not sure that I grasp the urgency of the situation. Isn't the AWKPATH user controlled and only used only during the compiler compilation?

dra27:
@avsm - we have already modified pristine OCaml releases in the past, for similar reasons as this (i.e. where the build doesn't work in a particular environment).

@kit-ty-kate - the original poster of the problem reported that they could not build any switches for 4.07-4.10 as a result of this setting and they still can't unless the patch is backported. It's as important as that user's concerns!

@Octachron - indeed, `AWKPATH` it inadvertently affected the pre-processing of the module aliases in stdlib, because the `-f` parameter is not qualified.

I have no further opinion, I was just idly using my last Friday off before childcare becomes a 100% of the time activity!

Octachron:
My worries about fixing in urgency somewhat exotic configurations if that it can be hard to know if fixing one issue on one side does not create another issue in another somewhat exotic configuration. Moreover, it sounds like the issue can be solved by setting the AWKPATH to "." while building ocaml, which only requires trusting OCaml build process, or am I missing something?

dra27:
The only ‘urgency’ is that I fixed the bug in OCaml, but the OP’s problem was via opam, so I’ve pushed it here too. I agree with everything in principle, but the patch here adds `./` in a couple of places to OCaml’s build system. FWIW, in my experience of fighting awk, gawk and mawk, I’d be considerably more scared of a patch or change to the package which altered `AWKPATH`!

dra27:
I’m also wondering what opam-seasoning tastes like 🤔 (see Camelus’s comment)

avsm:
The patch seems very safe, but the alternative is simply to force `AWKPATH=.` and not modify the sources at all, isnt it? I'd strongly prefer pristine versions of historical releases of OCaml in opam-repository if at all practical.

dra27:
What do you mean by force, though? Alter the package, rather than patch it? Historic versions of Windows OCaml compilers is going to add a lot of patches, btw! None of the patches present in any of our packages alter the compiler, they just let you actually build it

avsm:
I mean through setting `AWKPATH` as an environment variable in the `opam` package description. My opinion of the historical Windows OCaml compilers is that they should definitely be in this repository, but with a opam (not OCaml) version variant that marks them as such.

dra27:
I think there "should" be a way of having things like this backported so that old compilers "just work" but I fully cede that opam-repository in its present guise is a risky place to do this for the base compilers... so closing this, to rise in a future ocaml-legacy way some day 🙂

PR #16060, opened 2020-03-23
Fix release version constraint on irmin-git

pascutto:
`irmin-git.2.0.0` is also compatible with `irmin.2.1.0`


kit-ty-kate:
Thanks

PR #16064, opened 2020-03-24
Fix musl+static switches for Alpine

XVilka:
Trying to fix https://github.com/ocaml/ocaml/issues/9390 in a "right way".
For now limited with only one compiler version, if everything is fine - I will add the rest.

### DIFF
--- a/packages/ocaml-variants/ocaml-variants.4.10.0+musl+static+flambda/opam
+++ b/packages/ocaml-variants/ocaml-variants.4.10.0+musl+static+flambda/opam
@@ -2,7 +2,7 @@ opam-version: "2.0"
 synopsis:
   "Official release 4.10.0, compiled with musl-gcc -static and with flambda activated"
 description:
-  "Requires musl-gcc to be installed (package musl on Arch Linux or musl-tools on Debian)"
+  "Requires musl-gcc to be installed (package musl on Arch Linux, musl-tools on Debian, musl-dev on Alpine)"
 maintainer: "platform@lists.ocaml.org"
 authors: "Xavier Leroy and many contributors"
 homepage: "https://ocaml.org"
@@ -17,6 +17,7 @@ depends: [
 depexts: [
   ["musl-tools"] {os-family = "debian"}
   ["musl"] {os-distribution = "arch"}
+  ["musl-dev"] {os-distribution = "alpine"}
 ]
 conflict-class: "ocaml-core-compiler"
 flags: compiler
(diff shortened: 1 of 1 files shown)


packages/ocaml-variants/ocaml-variants.4.10.0+musl+static+flambda/opam
```diff
@@ -32,7 +33,14 @@ build: [
...
     "LIBS=-static"
-  ] {os = "openbsd" | os = "freebsd" | os = "macos" | os = "linux"}
+  ] {os = "openbsd" | os = "freebsd" | os = "macos" | os = "linux" & os-distribution != "alpine" }
+  [
+    "./configure"
+    "--prefix=%{prefix}%"
+    "CFLAGS=-Os"
+    "--enable-flambda"
+    "LIBS=-static"
+  ] {os = "linux" & os-distribution != "alpine" }
```
gasche: You are checking `!= alpine` in both cases, so which configure is supposed to run for alpine then?

XVilka: Thanks, was obviously a typo, fixed.

mseri:
Thanks, Cc also @avsm

avsm:
This needs a manual test since CI won't run it (and the typo in the original PR indicates a lack of previous testing).

packages/ocaml-variants/ocaml-variants.4.10.0+musl+static+flambda/opam
```diff
@@ -17,6 +17,7 @@ depends: [
 depexts: [
   ["musl-tools"] {os-family = "debian"}
   ["musl"] {os-distribution = "arch"}
+  ["musl-dev"] {os-distribution = "alpine"}
```
avsm: Is this really needed on Alpine? musl is the default libc on Alpine, and it compiled before without this depext.

XVilka: musl-dev contains the *.a libc required for the static linking: https://pkgs.alpinelinux.org/contents?branch=edge&name=musl-dev&arch=x86_64&repo=main

mseri:
Any update on this?

XVilka:
I can proceed if this approach is good

avsm:
This broadly looks fine to me, but I'm a bit wary of adding `-Os` to the CFLAGS. It's yet another deviation that may cause unexpected behaviour. Is it necessary?

XVilka:
@avsm it was here before. It's not necessary, but for some reason it was here already.

avsm:
Indeed, we can fix that in a future PR.  This can be merged, or you can do one more simplification -- you don't need to duplicate the build stanzas, as you can put the constraint inline to just the portion of the configure script that is alpine-specific (see the build stanza just below it, where the -j flag is dropped on Cygwin alone.  That'll prevent any future cut-and-paste deviation from Alpine and the other distributions.

mseri:
@XVilka I have simplified the build command. Can you test this one? If it is ok I'll be happy to merge this and the subsequent PR for the musl+static variants.

mseri:
I made a couple of local tests that seem to work fine. I'll merge this one. @XVilka once you have time it will be good to also have the other ones. Thanks again all for the help

XVilka:
I am sorry, was (and still is) very busy with other things. Will send the rest ASAP once back to this.

mseri:
No hurry! Thanks for the help

PR #16065, opened 2020-03-24
Why3 1.3.1

silene:


silene:
The only issue is CentOS: "Error: Unable to find a match: gtksourceview2-devel". So, everything is fine.

kit-ty-kate:
Thanks!

PR #16070, opened 2020-03-24
Adding support for parsing [@alert ..] output.

vrotaru:
It fixes in-editor error reporting.

I hope re-using the same version is OK

### DIFF
--- a/packages/ocamleditor/ocamleditor.1.13.4~preview.4.10/opam
+++ b/packages/ocamleditor/ocamleditor.1.13.4~preview.4.10/opam
@@ -28,9 +28,9 @@ depopts: [
 ]
 dev-repo: "git+https://github.com/vrotaru/ocamleditor"
 url {
-    src: "https://github.com/vrotaru/ocamleditor/archive/1.1.34-preview.4.10.tar.gz"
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
I'd rather not reuse the same version. Also I'm noticing something a bit weird here. Why is it versioned `1.13.4~preview.4.10` but `1.13.4` already exists. Shouldn't it be `ocamleditor.1.13.5~preview.4.10` instead?

It's a bit out of the scope of this PR but I'm also surprise that the package accepts only OCaml 4.10.0. Doesn't it support any of the potential minor versions like 4.10.1?

vrotaru:
It is versioned `1.13.4~preview.4.10` because there are no extra features added compared with the latest released version.

I did not consider that "just" porting to `4.10` is worth a version bump. No actual new features were added.

And yes, it will not compile for OCaml `4.09` because it uses the `compiler-libs` to read the `.cmt/.cmti/.cmi` files which are used for indexing, displaying an outline of the code etc.

I do have branches which compile with versions `4.02`..`4.09` as well. Now that you've mentioned it is interesting it cleanly compiles and installs without a (build time) dependency on `compiler-libs`

vrotaru:
After further reflection I decided to close this pull request.

I have now more last minutes fixes on master, and something I need/want to fix. Hoping that this would has just a small fix was a bit naive of me.

PR #16076, opened 2020-03-25
ocaml-variants.4.10.0+default-unsafe-string: disable force-safe-string

ygrek:
Before:
```
$ opam sw show
4.10.0+default-unsafe-string

$ ocamlc -v
The OCaml compiler, version 4.10.0
Standard library directory: /home/ygrek/.opam/4.10.0+default-unsafe-string/lib/ocaml

$ ocaml
        OCaml version 4.10.0

# String.set "dsasd";;
Alert deprecated: Stdlib.String.set
Use Bytes.set instead.
Error: This expression has type string but an expression was expected of type
         bytes
```

After:
string type is equal to bytes as expected


kit-ty-kate:
oh good catch sorry for that. cc @Octachron

PR #16088, opened 2020-03-28
cry.0.6.5

toots:
This PR adds newly released `cry` version `0.6.5` with fixed support for `PUT` requests with icecast `2.5` and above.

### DIFF
--- a/packages/cry/cry.0.6.5/opam
+++ b/packages/cry/cry.0.6.5/opam
@@ -11,8 +11,11 @@ depends: [
   "dune" {> "2.0"}
 ]
 depopts: [
-  "ssl" {>= "0.5.9"}
+  "ssl"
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Constraints have a different meaning when placed in the `depopts` field. In `depopts`, if `ssl.0.5.0` is already installed for instance, opam will still install the package, the only thing that changes is whether or not the package gets rebuilt. `conflicts` on the contrary will forbid ssl < 0.5.9 to be installed alongside `cry.0.6.5` as it would detect that this package is installed and fail to build otherwise.

kit-ty-kate:
Thanks!

toots:
Thanks for the explanation. Fixed pushed upstream!

PR #16090, opened 2020-03-28
Add gufo as an opam package.

Piervit:
Gufo is a new shell language using functionnal programming. It is still early stage but using opam will help potentially beta-tester to take part.

Last weeks, I have tried a first pull request. It had several issues. I have taken in account the remarks that were given before closing the request. I tested the installation as stated in  https://opam.ocaml.org/doc/Packaging.html (opam install . from project directory). 

Thanks.


Piervit:
Sorry, checksum issue, will correct and submit a new request

PR #16091, opened 2020-03-28
+ mikmatch 1.0.9

ygrek:
https://github.com/mjambon/mikmatch/releases/tag/v1.0.9

Build fixes, including (but not limited to) :
* add separate ocamlfind subpackages for syntax and library
* fix build with ocaml-pcre 7.2.0 and later (api breakage)
* fix build with OCaml 4.10 (safe-string)

### DIFF
--- a/packages/mikmatch/mikmatch.1.0.9/opam
+++ b/packages/mikmatch/mikmatch.1.0.9/opam
@@ -7,11 +7,11 @@ dev-repo: "git://github.com/mjambon/mikmatch.git"
 authors: [ "Martin Jambon" ]
 build: [
   [make "str"]
-  [make] {"%{pcre:installed}%"}
+  [make] {pcre:installed}
 ]
 install: [
   [make "install-str"]
-  [make "install"] {"%{pcre:installed}%"}
+  [make "install"] {pcre:installed}
 ]
 depends: [
   "ocaml" {>= "4.07"}
--- a/packages/mikmatch/mikmatch.1.0.9/opam
+++ b/packages/mikmatch/mikmatch.1.0.9/opam
@@ -1,0 +1,40 @@
+opam-version: "2.0"
+maintainer: "ygrek@autistici.org"
+homepage: "https://mjambon.github.io/mjambon2016/micmatch.html"
+doc: ["https://mjambon.github.io/mjambon2016/mikmatch-manual.html"]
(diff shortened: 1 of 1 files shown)


packages/mikmatch/mikmatch.1.0.9/opam
```diff
@@ -0,0 +1,40 @@
...
+  [make "install"] {"%{pcre:installed}%"}
+]
+remove: [
+  [make "uninstall-str"]
+  [make "uninstall"] {"%{pcre:installed}%"}
+]
+depends: [
+  "ocaml" {>= "4.07"}
+  "ocamlfind" {build}
+  "camlp4" {build}
```
kit-ty-kate: Looking at https://github.com/mjambon/mikmatch/blob/23f78d273dbcc4c1b0a38beddffb0e0f3991d8a7/pcre/META it seems that camlp4 is used at runtime as well.

kit-ty-kate: Also know that camlp4 is not available for OCaml 4.09 and above and is deprecated.

packages/mikmatch/mikmatch.1.0.9/opam
```diff
@@ -0,0 +1,40 @@
...
+  [make] {"%{pcre:installed}%"}
+]
+install: [
+  [make "install-str"]
+  [make "install"] {"%{pcre:installed}%"}
+]
+remove: [
+  [make "uninstall-str"]
+  [make "uninstall"] {"%{pcre:installed}%"}
+]
```
kit-ty-kate: This field is unnecessary since opam 2.0

packages/mikmatch/mikmatch.1.0.9/opam
ygrek: https://github.com/ocaml/camlp4/pull/153

ygrek:
thanks for the fixes

kit-ty-kate:
It turns out the syntax for getting variables in conditional blocks was wrong (related issue https://github.com/ocaml/opam/issues/4000 for why it didn't trigger any warnings before)

packages/mikmatch/mikmatch.1.0.9/opam
kit-ty-kate: oh i see. Thanks for the info

kit-ty-kate:
Thanks!

PR #16092, opened 2020-03-29
[new release] 0install-gtk and 0install (2.15.2)

talex5:
Decentralised installation system - GTK UI

- Project page: <a href="https://0install.net/">https://0install.net/</a>

### Bug fixes

- Don't try to update distribution caches in `--dry-run` mode.
  If we needed to update a cache, the operation would fail with e.g.
  `Bug: 'with_open_out' called in --dry-run mode`.

### Opam package

- Mark obus as required (except on Windows and macOS).
  This is more explicit than just asking users to install it where it makes sense, and also means that the CI will test it.

- Split GTK UI into a separate opam package.
  This makes it easy to install it (using `opam install 0install-gtk`), and means that the CI will test it.

- Update opam metadata to match opam-repository.

### Code cleanups

- Rename "ocaml" directory to "src". There are no other languages here now.

- Make the solver return a `SOLVER_RESULT`.
  The old API was a bit confusing. The user of the solver had to wrap the solver's return to provide a `SOLVER_RESULT` for the diagnostics.
  Now, the user-required bits are in `SOLVER_INPUT` and the solver itself provides the rest.

- Split the solver into its own library. Since the refactoring in 2014
  (see [Simplifying the Solver With Functors](https://roscidus.com/blog/blog/2014/09/17/simplifying-the-solver-with-functors/)),
  the solver isn't really tied to 0install at all, and could be useful for other package managers.

- Generalise the solver's machine groups system.
  This removes the one remaining dependency from the solver to the rest of 0install.

- Rename option functions in solver to match OCaml 4.08.

- Split `Feed_metadata` and `Feed_import` out into their own modules.

- Rename `Feed.feed` to `Feed.t` and make the type abstract.

- Skip rpm2cpio unit test if cpio isn't available. It seems that recent Fedora images provide rpm2cpio, but not cpio.

### DIFF
--- a/packages/0install-gtk/0install-gtk.2.15.2/opam
+++ b/packages/0install-gtk/0install-gtk.2.15.2/opam
@@ -1,0 +1,29 @@
+opam-version: "2.0"
+synopsis: "Decentralised installation system - GTK UI"
+maintainer: "talex5@gmail.com"
+authors: "zero-install-devel@lists.sourceforge.net"
+homepage: "https://0install.net/"
+bug-reports: "https://github.com/0install/0install/issues"
+dev-repo: "git+https://github.com/0install/0install.git"
+build: [
+  ["dune" "build" "-p" name "-j" jobs "@install" "@runtest" {with-test}]
+]
+depends: [
+  "ocaml" {>= "4.05.0"}
+  "0install" {= version}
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
Thanks!

PR #16096, opened 2020-03-29
switch domain and use https

ygrek:
s/ygrek.org.ua/ygrek.org/


ygrek:
I don't really care to fix warnings in old versions of packages, I'd rather drop them if necessary.

kit-ty-kate:
Thanks a lot

PR #15933, opened 2020-03-02
[new release] ppx_inline_test (0.13.1)

diml:
Syntax extension for writing in-line tests in ocaml code

- Project page: <a href="https://github.com/janestreet/ppx_inline_test">https://github.com/janestreet/ppx_inline_test</a>
- Documentation: <a href="https://ocaml.janestreet.com/ocaml-core/latest/doc/ppx_inline_test/index.html">https://ocaml.janestreet.com/ocaml-core/latest/doc/ppx_inline_test/index.html</a>

##### CHANGES:

- Honor the `inline_tests` Dune variable so that inline tests are
  dropped in release builds

PR #16097, opened 2020-03-29
rename ppx_inline_test 0.13.1 to ppx_inline_test v0.13.1

hannesm:
otherwise it'll never be selected for installation //cc @diml //cc #15933


hannesm:
but unfortunately even with ppx_inline_test pinned to 0.13.1 I get in my _build/install/dns/META the following entries:
```bash
            ppx_expect.collector
            ppx_expect.config
            ppx_here.runtime-lib
            ppx_inline_test.runtime-lib
```

kit-ty-kate:
Sorry for not seeing that the first time. Thanks!

diml:
Thanks for the fix @hannesm! Doing a minor release of these packages is always a bit tedious because of the "v" prefix. We are currently missing the ability to tell dune-release to keep the "v" prefix once and for all for a given project.

PR #16104, opened 2020-03-30
apero-core.0.4.7

JEnoch:
Hi,
this PR adds [apero-core](https://github.com/atolab/apero-core).

### DIFF
--- a/packages/apero-core/apero-core.0.4.7/opam
+++ b/packages/apero-core/apero-core.0.4.7/opam
@@ -1,5 +1,4 @@
 opam-version: "2.0"
-version:"0.4.7"
 maintainer:   ["contact@adlink-labs.tech"]
 authors:      ["Angelo Corsaro" "Julien Enoch" "Oliver Hecart" "Gabriele Baldoni"]
 homepage:     "https://github.com/atolab"
@@ -10,7 +9,7 @@ license: "Apache-2.0"
 
 
 build: [
- ["dune" "build" "--only" "apero-core" "--root" "." "-j" jobs "@install"]
+ ["dune" "build" "-p" name "-j" "jobs"]
 ]
 
 depends: [
--- a/packages/apero-core/apero-core.0.4.7/opam
+++ b/packages/apero-core/apero-core.0.4.7/opam
@@ -9,7 +9,7 @@ license: "Apache-2.0"
 
 
 build: [
- ["dune" "build" "-p" name "-j" "jobs"]
+ ["dune" "build" "-p" name "-j" jobs]
 ]
 
 depends: [
--- a/packages/apero-core/apero-core.0.4.7/opam
+++ b/packages/apero-core/apero-core.0.4.7/opam
@@ -1,0 +1,45 @@
+opam-version: "2.0"
+version:"0.4.7"
+maintainer:   ["contact@adlink-labs.tech"]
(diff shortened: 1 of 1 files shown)


JEnoch:
Hi,
it seems that ci/datakit/3 Compilers job has some troubles (not sure why):
1. With OCaml 4.07.1:  The sources of the following couldn't be obtained, aborting: bigarray-compat.1.0.0
2. With OCaml 4.08.1: The sources of the following couldn't be obtained, aborting: biniou.1.2.1
3.  With OCaml 4.09.1: fatal: unable to connect to github.com: errno=Connection timed out

But I don't think it's caused by apero-core package.

packages/apero-core/apero-core.0.4.7/opam
```diff
@@ -0,0 +1,45 @@
...
+authors:      ["Angelo Corsaro" "Julien Enoch" "Oliver Hecart" "Gabriele Baldoni"]
+homepage:     "https://github.com/atolab"
+bug-reports:  "https://github.com/atolab/apero-core/issues/"
+dev-repo:     "git+https://github.com/atolab/apero-core.git"
+
+license: "Apache-2.0"
+
+
+build: [
+ ["dune" "build" "--only" "apero-core" "--root" "." "-j" jobs "@install"]
```
kit-ty-kate: The preferred command to use during releases is: `["dune" "build" "-p" name "-j" jobs"]`

packages/apero-core/apero-core.0.4.7/opam
```diff
@@ -0,0 +1,45 @@
+opam-version: "2.0"
+version:"0.4.7"
```
kit-ty-kate: This field is unnecessary

packages/apero-core/apero-core.0.4.7/opam
```diff
@@ -0,0 +1,45 @@
...
+  "cmdliner" {= "1.0.4" }
+  "fmt"  {= "0.8.8" }
+  "logs"  {= "0.7.0" }
+  "lwt" {= "5.1.1" }
+  "ppx_deriving"  {= "4.4" }
+  "uuidm" {= "0.9.7" }
+  "bigstringaf"  {= "0.6.1" }
+  "lwt_ppx"  {= "2.0.0" }
+  "yojson"  {= "1.7.0" }
+  "ppx_deriving_yojson"  {= "3.5.1" }
```
kit-ty-kate: Is there a reason for dependency constraints this strict? Our CI already checks when something breaks and strict constrains are very annoying for users

JEnoch: 2 reasons:

1. We wanted reproductible builds (similarly to what's achieved in Java using Maven). I.e.: be sure that the version 0.4.7 of the software always leads to the same set of codes and thus to the same binary.
2. apero-core is used in [Eclipse zenoh](https://projects.eclipse.org/projects/iot.zenoh). And the Eclipse rules force us to precisely (i.e. with version, sources...) identify each piece of software that is a dependency of an Eclipse project. The goal being to provide legal guarantees wrt. IP used by an Eclipse software.

kit-ty-kate:
Thanks

JEnoch:
Thanks to you!
Note that soon I'll do 3 other PRs for other related packages. I'll make sure to take your comments for this opam in account for the next ones. However, they'll also have strict dependency constraints for the same reasons.

PR #16102, opened 2020-03-29
[new release] salsa20-core (1.0.0)

abeaumont:
The Salsa20 core functions, in OCaml

- Project page: <a href="https://github.com/abeaumont/ocaml-salsa20-core">https://github.com/abeaumont/ocaml-salsa20-core</a>
- Documentation: <a href="https://abeaumont.github.io/ocaml-salsa20-core/">https://abeaumont.github.io/ocaml-salsa20-core/</a>

##### CHANGES:

* Port to dune (thanks to @hannesm)

PR #16106, opened 2020-03-31
restrict salsa20 to salsa20-core < 1.0.0, which removed some bindings

hannesm:
//cc @abeaumont in order to unblock #16102


hannesm:
CI failure is that there's a warning (and warn-error enabled) in the test cases. I fixed by removing the test instructions from salsa20.0.1.0.

kit-ty-kate:
Thanks a lot!

PR #16111, opened 2020-03-31
Lwt: dune exec: use --no-config

aantron:
See https://github.com/ocaml/dune/issues/3296.

### DIFF
--- a/packages/lwt/lwt.5.1.2/opam
+++ b/packages/lwt/lwt.5.1.2/opam
@@ -42,7 +42,7 @@ conflicts: [
 ]
 
 build: [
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
Would it not be even more simple with `"-p" name` instead?

kit-ty-kate:
> Would it not be even more simple with `"-p" name` instead?

(replacing the whole `"--no-config" "--root" "."`)

aantron:
Changed to `"-p" name`.

aantron:
The revdeps failures look unrelated to this PR. I'll submit separate PRs to fix some of them, though.

mseri:
Looks good. Thanks

Issue #15295, opened 2019-11-15
conf-libssl doesn't build on osx


Bronsa:
```
[~]> opam install conf-libssl
The following actions will be performed:
  ∗ install   conf-libssl      1
  ↻ recompile ssl              0.5.9 [uses conf-libssl]
  ↻ recompile conf-openssl     1     [uses conf-libssl]
  ↻ recompile lwt_ssl          1.1.2 [uses ssl]
  ↻ recompile conduit-lwt-unix 2.0.1 [uses lwt_ssl]
  ↻ recompile cohttp-lwt-unix  2.3.0 [uses conduit-lwt-unix]
===== ∗ 1   ↻ 5 =====
Do you want to continue? [Y/n] y

<><> Gathering sources ><><><><><><><><><><><><><><><><><><><><><><><><><><>  🐫
[lwt_ssl.1.1.2] found in cache
[ssl.0.5.9] downloaded from cache at https://opam.ocaml.org/cache
[cohttp-lwt-unix.2.3.0] downloaded from cache at https://opam.ocaml.org/cache
[conduit-lwt-unix.2.0.1] downloaded from cache at https://opam.ocaml.org/cache

<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><>  🐫

#=== ERROR while compiling conf-libssl.1 ======================================#
"./osx-build.sh": command not found.


<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><>  🐫
┌─ The following actions failed
│ λ build conf-libssl 1
└─
╶─ No changes have been performed

The packages you requested declare the following system dependencies. Please make sure they are installed before retrying:
    openssl
```

kit-ty-kate:
Is it fixed by https://github.com/ocaml/opam-repository/pull/15331?

Bronsa:
I'm not in front of a laptop right now, but https://github.com/ocaml/opam-repository/pull/15330 fixes conf-libssl on macos. 

Your change should make conf-openssl.1 installable again which is great, but conf-libssl should still be broken

kit-ty-kate:
I'm not too sure to understand how it worked before. The code to of `build` section didn't change. I'll merge it anyway as soon as the macos CI is green but I'm a bit puzzled.

Bronsa:
Maybe some recent change to opam? it looks like the +x flag is cleared
after the file is installed locally

reykjalin:
[...]

PR #16112, opened 2020-03-31
Fix conf-libssl on macos with homebrew

kit-ty-kate:
Issue encountered in https://github.com/ocaml/opam-repository/issues/15295#issuecomment-606186272 and in the CI


mseri:
Looks good and works on my laptop

PR #16117, opened 2020-04-01
Update META for metaocaml package

ale64bit:
TL;DR: to make metaocaml work with dune.

See full discussion: https://discuss.ocaml.org/t/how-to-use-ber-metaocaml-with-dune-cleanly/5419


kit-ty-kate:
Sounds good. cc @yallop

kit-ty-kate:
Thanks!

PR #16119, opened 2020-04-01
Disable tests of rpclib-lwt

aantron:
`rpclib-lwt`'s tests produce failures like [this one](https://ci.ocaml.org/log/saved/docker-run-41be14d28eeb8ecf1eb98587cb889cbf/453ebae17e8028b0c921e8b171504b01470310dc):

```
- File "tests/lwt/suite.ml", line 3, characters 28-52:
- 3 |     [("Client_server_test", Client_server_test.tests)]
-                                 ^^^^^^^^^^^^^^^^^^^^^^^^
- Error: This expression has type unit Alcotest_lwt.test_case list
-        but an expression was expected of type unit Alcotest.test_case list
-        Type
-          unit Alcotest_lwt.test_case =
-            string * Alcotest_lwt.speed_level * (unit -> Alcotest_lwt.return)
-        is not compatible with type
-          unit Alcotest.test_case =
-            string * Alcotest.speed_level * (unit -> Alcotest.return) 
-        Type Alcotest_lwt.return = unit Lwt.t is not compatible with type
-          Alcotest.return = unit
```

...so this PR disables them. This applies to all versions of the package.

@samoht, @jonludlam, is there a more intelligent solution to this, perhaps some constraints on Alcotest?

### DIFF
--- a/packages/rpclib-lwt/rpclib-lwt.5.9.0/opam
+++ b/packages/rpclib-lwt/rpclib-lwt.5.9.0/opam
@@ -15,11 +15,11 @@ build: [
 ]
 depends: [
   "ocaml"
-  "alcotest" {with-test}
+  "alcotest" {with-test & < "1.0.0"}
   "jbuilder"
   "rpclib" {= version}
   "lwt" {>= "3.0.0"}
-  "alcotest-lwt" {with-test}
+  "alcotest-lwt" {with-test & < "1.0.0"}
 ]
 synopsis: "Driver for rpclib using lwt"
 url {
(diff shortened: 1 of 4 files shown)


jonludlam:
Looks to me like @mseri has fixed this in upstream rpclib - https://github.com/mirage/ocaml-rpc/commit/51110e7e326026aa5f4ab02a4cb120413e20fcd6 - so when we get a new release we can turn the tests back on.

kit-ty-kate:
I believe this can be fixed with a constraint `alcotest < 1.0.0`

jonludlam:
Even better then!

mseri:
Thanks! Yes the upper bounds should be enough

kit-ty-kate:
Thanks a lot!

PR #16126, opened 2020-04-01
Remove erroneous freebsd constraint

web-vertalo:
Package name for freebsd must match output of pkg query %n (https://github.com/ocaml/opam-depext/blob/14a9aa6093bfb510b83f2ba07205b899b7e89989/depext.ml#L343) and should not use the category prefix.

The correct constraint already appears in each file.


kit-ty-kate:
Thanks

PR #16128, opened 2020-04-01
fixes a typo in bap-constant-tracker

gitoleg:
This PR fixes a typo that causes a failure during the reinstalling `bap-constant-tracker` package.


kit-ty-kate:
Thanks

PR #16138, opened 2020-04-03
Create ocaml-variants.4.09.1+afl+flambda.

fpottier:
Hello! I don't know whether the combination of `afl` and `flambda` is supposed to work and has been tested, but I would be curious to try it out. Hence this PR. I have a library that has been well tested using `afl` under regular OCaml and would like to do the same under flambda; I figure that this might be an interesting way of finding bugs in my code or in flambda. (@mshinwell, @stedolan, what do you think?)


kit-ty-kate:
Sounds good. Thanks! cc @Octachron

PR #16142, opened 2020-04-04
Package ocplib-endian.1.1

chambart:
After, hum quite some time. I do release a new version of ocplib-endian.
The minimal OCaml version has been bumped to 4.02.3

### DIFF
--- a/packages/ocplib-endian/ocplib-endian.1.1/opam
+++ b/packages/ocplib-endian/ocplib-endian.1.1/opam
@@ -1,0 +1,38 @@
+opam-version: "2.0"
+synopsis:
+  "Optimised functions to read and write int16/32/64 from strings and bigarrays"
+description: """
+The library implements three modules:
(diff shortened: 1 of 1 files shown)


chambart:
There are two build failures. msgpack is a bad checksum, so completely unrelated, we can ignore it here.

But the other is more interesting. The sequence of installs for the test produced a situation where dune 2.4.0 was installed then ocplib-endian was installed because it was required by opam-depext. Then the package requested a dune version < 2 hence it downgraded to dune 1.11.4.

But ocplib-endian was installed by dune, which produced a dune-package file with `(lang dune 2.4)` which of course is rejected by dune 1.11.4.

Is there a way to tell dune to produce dune-package files with version that are not higher than the one in the dune-project file (hence corresponding to the version constraint on the package) ? Or dune should just not be a build only dependency ?

chambart:
I just changed it to a non-build dependency. If this is the right solution then it will be ready, otherwise, let me know and I will change this.

kit-ty-kate:
> Or dune should just not be a build only dependency ?

Yes. See #14266 and https://github.com/ocaml/dune/issues/2147

Thanks a lot!

PR #16144, opened 2020-04-04
New project, new package: dune-deps 1.0.0

mjambon:
The project is at https://github.com/mjambon/dune-deps

### DIFF
--- a/packages/dune-deps/dune-deps.1.0.0/opam
+++ b/packages/dune-deps/dune-deps.1.0.0/opam
@@ -1,0 +1,34 @@
+opam-version: "2.0"
+maintainer: "martin@mjambon.com"
+authors: ["Martin Jambon"]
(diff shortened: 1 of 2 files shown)


mjambon:
I see `Error: Unbound value Hashtbl.find_opt` on older ocaml versions. I'm on it.

kit-ty-kate:
Your `dune-project` tells us that this project requires at least dune 2.1, moreover the use of `dune install` (previously called by `make install` in not necessary and should be avoided as much as possible as it uses ocamlfind internally amongst other issues).

This should be good to merge once the CI finishes. Could you return the fixes upstream?

kit-ty-kate:
All good now. Thanks!

mjambon:
Great. Thanks for the fixes!

PR #16145, opened 2020-04-05
[new release] 0install-gtk and 0install (2.16)

talex5:
CHANGES:

- Update to GTK 3, because Debian is removing GTK 2 support now. Note that the systray icon no longer blinks to indicate that action is required, as GTK removed this feature.

- Upgrade to dune 2.1.

### DIFF
--- a/packages/0install-gtk/0install-gtk.2.16/opam
+++ b/packages/0install-gtk/0install-gtk.2.16/opam
@@ -1,0 +1,28 @@
+opam-version: "2.0"
+synopsis: "Decentralised installation system - GTK UI"
+maintainer: "talex5@gmail.com"
+authors: "zero-install-devel@lists.sourceforge.net"
+homepage: "https://0install.net/"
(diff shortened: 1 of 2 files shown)


talex5:
macOS failed with:

```
ERROR while compiling conf-gtk3.18
# command     ~/.opam/opam-init/hooks/sandbox.sh build pkg-config --short-errors --print-errors --atleast-version 3.18 gtk+-3.0
# No package 'gtk+-3.0' found
```

@garrigue is this a known problem? I don't use a mac, but it looks like the package is called `gtk+3` (not `gtk`) from this page https://formulae.brew.sh/formula/gtk+3

kit-ty-kate:
Attempt at fixing the issue in https://github.com/ocaml/opam-repository/pull/16147

kit-ty-kate:
The fix also has an issue so we can ignore the error for now. Thanks!

PR #16147, opened 2020-04-05
Fix conf-gtk3 on macos with homebrew

kit-ty-kate:


kit-ty-kate:
I'm not sure why this fails, I've tried 3 times to restart the job and it fails the same way. It seems to be an issue with Homebrew that fails to install both python3.7 and 3.8:
```
==> Pouring python-3.7.7.high_sierra.bottle.tar.gz

Error: The `brew link` step did not complete successfully
```
It should be fine to merge anyway

PR #16155, opened 2020-04-07
Release x509 0.11.0

hannesm:
[new release] x509 (0.11.0)

CHANGES:

* BREAKING Validation.validate_raw_signature results in a
  `(unit, signature_error) result` and logs (Logs.warn) if a weak (non-SHA2)
  hash algorithm was used. This function is used for verifying signatures
  on certificates, signing requests, and certificate revocation lists.
* The `` `CAInvalidSelfSignature `` constructor (Validation.ca_error) and
  `` `ChainInvalidSignature `` constructor (Validation.chain_validation_error)
  have been removed.
* BREAKING The polymorphic variant `Validation.chain_error` is now flat (the
  tags `` `Chain `` and `` `Leaf `` have been removed)
* BREAKING Adapted return type of CRL.validate and CRL.verify
* The pretty-printer Public_key.pp is now provided
* All implemented by @hannesm in mirleft/ocaml-x509#132 based on private conversation with @cfcs
  (who kindly reviewed the changes)

### DIFF
--- a/packages/letsencrypt/letsencrypt.0.2.1/opam
+++ b/packages/letsencrypt/letsencrypt.0.2.1/opam
@@ -25,7 +25,7 @@ depends: [
   "mirage-crypto"
   "mirage-crypto-pk"
   "mirage-crypto-rng"
-  "x509" {>= "0.10.0"}
+  "x509" {>= "0.10.0" & < "0.11.0"}
   "yojson" {>= "1.6.0"}
   "ounit" {with-test}
   "dns"
--- a/packages/tls/tls.0.11.0/opam
+++ b/packages/tls/tls.0.11.0/opam
@@ -24,7 +24,7 @@ depends: [
   "mirage-crypto"
   "mirage-crypto-pk"
   "mirage-crypto-rng"
-  "x509" {>= "0.10.0"}
+  "x509" {>= "0.10.0" & < "0.11.0"}
   "domain-name" {>= "0.3.0"}
   "fmt"
   "cstruct-unix" {with-test & >= "3.0.0"}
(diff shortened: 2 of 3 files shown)


hannesm:
revdep failure is unrelated (in irc-client, all versions):
```
# File "_none_", line 1:
# Error: Required module `Mutex' is unavailable
# Command exited with code 2.
```
maybe @c-cube @johnelse could fix this (likely inserting an upper bound (not sure on what -- I suspect there's need to pass `thread` to fix it)?)

johnelse:
Is there a policy for removing old versions of packages? `irc-client.0.5.*` and earlier are from before jbuilder/dune and also from before the opam package was split up into separate lwt/unix/tls packages, so I'd be happy for them to be removed at this point - I can't imagine anyone wanting to use them now.

I'll look into fixing them if removing old packages is not the done thing.

hannesm:
@johnelse ah, ok. what I do instead of removing is adding `available: false` to the opam file, which seems to be accepted by the opam-repository maintainers. :)

johnelse:
@hannesm ah that sounds like a good solution :) PR is here: https://github.com/ocaml/opam-repository/pull/16156

kit-ty-kate:
The issues with `irc-client` is sadly not fixable (I think there should be an opam issue on this somewhere, I'll try to link it later).

kit-ty-kate:
> The issues with `irc-client` is sadly not fixable (I think there should be an opam issue on this somewhere, I'll try to link it later).

I found it. It's this one: https://github.com/ocaml/opam/issues/4048

kit-ty-kate:
Oops sorry I forgot to merge this one. Thanks!

PR #16156, opened 2020-04-07
Make old versions of irc-client unavailable

johnelse:
These are all from before the switch to jbuilder/dune, and before this
package was split up into lwt/tls/unix variants. They shouldn't be used
by anyone at this point.

Signed-off-by: John Else <john.else@gmail.com>

### DIFF
--- a/packages/irc-client/irc-client.0.1.1/opam
+++ b/packages/irc-client/irc-client.0.1.1/opam
@@ -22,3 +22,4 @@ url {
   src: "https://github.com/johnelse/ocaml-irc-client/archive/0.1.1.tar.gz"
   checksum: "md5=90ac5b939c3d1e473afc6a9b9cb0e33a"
 }
+available: false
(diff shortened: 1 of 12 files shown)


avsm:
We'd prefer not to make them unavailable unless there's a strong reason for doing so -- we keep all historical versions of packages in the opam-repository.  Users will automatically get the latest release of irc-client with the newer build system, or is there some other reason that the older releases are problematic to keep around?

johnelse:
@avsm The problem is they sometimes break revdeps for tls and x509 (e.g. https://github.com/ocaml/opam-repository/pull/16155). Normally a version restriction would fix it but I don't think that's the right fix here. I'll try to reproduce the same failure locally and see if I can up with something better than removing these packages outright.

avsm:
thanks; no need to fix the older versions and give yourself more work. The easiest thing to do is simply to add an upper bound to the x509 package versions, so that future releases will not even be selected for older irc-client.  If in the future some enterprising soul really wants to make older irc-client work with newer x509, they can relax those bounds. In the meanwhile, you won't be bothered with revdeps failures, and older irc-client continues to install using the old packages :)

johnelse:
Thanks all!

PR #16157, opened 2020-04-07
remove 'with-test' in conflict of irc-client (0.4.0, 0.5.*) to avoid the failing issues

hannesm:
see ocaml/opam#4048 for a description, #16155 for initial discussion. this is meant as a (partial) alternative to #16156 //cc @kit-ty-kate @johnelse


hannesm:
I won't dig the revdeps failure rabbit hole further -- to me it looks like the irc-client-* packages should include a dependency on `irc-client {=version}`.

kit-ty-kate:
Looks fine. Thanks!

PR #16158, opened 2020-04-07
added ProVerif version 2.01

bruno-blanchet:


bruno-blanchet:
I do not really understand why camelus says "new broken packages (1): proverif.2.01". The message is not very explicit, and the other tests work, so is the package really broken? If you can help me, that would be great. Thanks a lot.

kit-ty-kate:
"broken packages" for Camelus means "package that cannot be installed (anymore)"

kit-ty-kate:
however in this case the CI could check that it indeed works and Camelus just took this branch alone without merging our master in to do the check. Your branch is very old (the last commit before this one was from 2017) so some packages might have been missing then.

It looks fine otherwise. I'll open an issue on Camelus to also take into account master not just the branch

kit-ty-kate:
Thanks

PR #16159, opened 2020-04-07
Release dune-deps 1.1.0. Adds dependency on cmdliner.

mjambon:
`make test` also depends on alcotest but I didn't specify it in the opam file since it's not required for the build.


kit-ty-kate:
Thanks

PR #16160, opened 2020-04-07
3 packages from ryanslade/influxdb-ocaml at 0.3.0

ryanslade:
This pull-request concerns:
-`influxdb.0.3.0`: InfluxDB client library
-`influxdb-async.0.3.0`: InfluxDB client library using async for concurrency
-`influxdb-lwt.0.3.0`: InfluxDB client library using lwt for concurrency

---
* Homepage: https://github.com/ryanslade/influxdb-ocaml
* Source repo: git://github.com/ryanslade/influxdb-ocaml.git
* Bug tracker: https://github.com/ryanslade/influxdb-ocaml/issues

---

### DIFF
--- a/packages/influxdb-async/influxdb-async.0.3.0/opam
+++ b/packages/influxdb-async/influxdb-async.0.3.0/opam
@@ -1,0 +1,26 @@
+opam-version: "2.0"
+synopsis: "InfluxDB client library using async for concurrency"
+description: "A client library for writing time series data to InfluxDB"
+maintainer: "Ryan Slade <ryanslade@gmail.com>"
+authors: "Ryan Slade <ryanslade@gmail.com>"
+homepage: "https://github.com/ryanslade/influxdb-ocaml"
+bug-reports: "https://github.com/ryanslade/influxdb-ocaml/issues"
+dev-repo: "git://github.com/ryanslade/influxdb-ocaml.git"
+build: [["dune" "build" "-p" name "-j" jobs]]
+depends: [
+  "dune" {>= "2.0"}
+  "base"
+  "alcotest" {with-test}
+  "influxdb" {= "0.3.0"}
+  "async"
+  "cohttp"
+  "cohttp-async"
+  "ocaml" {>= "4.04.0"}
+]
+url {
+  src: "https://github.com/ryanslade/influxdb-ocaml/archive/0.3.0.tar.gz"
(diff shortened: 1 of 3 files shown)


kit-ty-kate:
Dune should not be tagged as a build dependency (see https://github.com/ocaml/dune/issues/2147) and according to your `dune-project` file, these packages require at least dune 2.0.

Could you return those fixes upstream?

packages/influxdb-async/influxdb-async.0.3.0/opam
```diff
@@ -0,0 +1,26 @@
...
+authors: "Ryan Slade <ryanslade@gmail.com>"
+homepage: "https://github.com/ryanslade/influxdb-ocaml"
+bug-reports: "https://github.com/ryanslade/influxdb-ocaml/issues"
+dev-repo: "git://github.com/ryanslade/influxdb-ocaml.git"
+build: [["dune" "build" "-p" name "-j" jobs]]
+depends: [
+  "dune" {>= "2.0"}
+  "base"
+  "alcotest" {with-test}
+  "influxdb"
```
kit-ty-kate, line 14: I would also strongly suggest to add `{= version}` to not have to maintain each previous packages everytime the API breaks. It is not mandatory but it is a friendly advice.

packages/influxdb-lwt/influxdb-lwt.0.3.0/opam
```diff
@@ -0,0 +1,26 @@
...
+authors: "Ryan Slade <ryanslade@gmail.com>"
+homepage: "https://github.com/ryanslade/influxdb-ocaml"
+bug-reports: "https://github.com/ryanslade/influxdb-ocaml/issues"
+dev-repo: "git://github.com/ryanslade/influxdb-ocaml.git"
+build: [["dune" "build" "-p" name "-j" jobs]]
+depends: [
+  "dune" {>= "2.0"}
+  "base"
+  "alcotest" {with-test}
+  "influxdb"
```
kit-ty-kate, line 14: same comment/suggestion here.

ryanslade:
@kit-ty-kate Thank you for the tips, PTAL

kit-ty-kate:
Thanks! I left a comment on your latest commit in the source repository btw (just in case you don't get notified for that)

PR #16163, opened 2020-04-08
Genprint v0.3

progman1:


kit-ty-kate:

progman1:
hi
I get a bad checksum error when installing from the default repository.
I had a horrible time originally getting this into the repo,
repeatedly releasing,
generating a checksum and incorporating it into the opam file, pushing it....
20-30 times!
this time I'm assuming this circularity is accounted for somewhere so
I generate a release, calc a checksum, update opam file, push it, and
pull-request.

what's wrong with that?!
one spanner in the works that didn't seem to be there originally was
having made a release and finding I had to change something, I deleted
the release and regnerated it.
or so I thought!  it seems to be the case that github just re-instates
the old release bundle
unless the name is changed.  so tarball is now 0.3a rather than 0.3.
would that make a difference?

thanks.

ps. opam-publish doesn't work for me - apart from creating another
local copy of the 250Mb repo it creates a branch <xxx>.ccc where the
ccc is literal I expect ought to be the version number.  so I'm doing
things
piecemeal though it be ever so complicated....

progman1:
fix upstream? am I 'upstream'?

kit-ty-kate:
> fix upstream? am I 'upstream'?

I'm talking about the upstream repository (this file: https://github.com/progman1/genprintlib/blob/master/genprint.opam)

Also not that having the `url` field in this file in the source repository is usually unwanted. Usually when pinning a package locally you want the version that is actually in the source repository, not the latest stable version. I'd advice you to remove it, the `url` field is only meant to be used in opam-repository for stable releases.

kit-ty-kate:
Reading your previous comment it seems you were confused about the `url` section. As said above it is not meant to be in the source repository. The opam file in the archive is not used when using opam-repository so they don't have to be synchronised (it is impossible anyway as you mentionned).

> or so I thought!  it seems to be the case that github just re-instates the old release bundle

I'm a bit confused about that. As long as the tag is pushed-force the archive linked to the tag changes with it.

> ps. opam-publish doesn't work for me

Which version of `opam-publish` are you using?
Something in the vain of:
```
$ git tag v0.3
$ git push origin master
$ git push origin v0.3
$ opam publish
```
should work. If not maybe with `--tag=v0.3 -v 0.3` or something like that might help for next time.

progman1:
so no need to download the archive and generate a checksum?
it's inserted into the opam file by the build checking process?

opam-publish, a standalone binary, gives 2.0.0
when I tried 'opam publish' in the source directory it somehow picks up the 0.3.
but from where? the dev-repo line in the opam file?

thanks.

kit-ty-kate:
> so no need to download the archive and generate a checksum? it's inserted into the opam file by the build checking process?

If you are using `opam-publish` then it will be inserted by `opam-publish`. If you're not using `opam-publish` or similar tools then you will have to insert it by hand when doing the PR here.

> opam-publish, a standalone binary, gives 2.0.0 when I tried 'opam publish' in the source directory it somehow picks up the 0.3. but from where? the dev-repo line in the opam file? thanks.

`opam-publish` detects it using git. It will take the `origin` branch, the latest tag in the current branch and will deduce the path to the archive on github, download it, take the checksum and open the PR with those informations.

PR #16164, opened 2020-04-08
Genprint v0.3

progman1:


progman1:
I took out the url, corrected the dune dep....
and looking forward to plain sailing.

kit-ty-kate:
Sorry if I wasn't clear earlier, I did say that the `url` section was not necessary in the **source repository** (https://github.com/progman1/genprintlib/) but it is necessary for **opam-repository**. Otherwise how would we guess where to get the archive from?

PR #16171, opened 2020-04-10
[ocamlformat] Fix dependency on dune

ejgallego:
`ocamlformat` depends on `(using menhir 2.1)` which in turn requires
Dune >= 2.2.0


kit-ty-kate:
cc @gpetiot @Julow

kit-ty-kate:
Thanks!

Julow:
Thanks !

PR #16179, opened 2020-04-10
camlidl: fix bytecode link

ygrek:


ygrek:
```
￼ 2 ignored non-opam files:
packages/camlidl/camlidl.1.07/files/META
packages/camlidl/camlidl.1.07/files/disable-fatal-warn-31.diff
```
I am not sure what that means, not clean state on CI?

kit-ty-kate:
one of the revdeps (`lutin.2.70.4`) fails with this error message. Is it expected?
```
# Error: Error on dynamically loaded library: /home/opam/.opam/4.10/lib/stublibs/dllcamlidl.so: /home/opam/.opam/4.10/lib/stublibs/dllcamlidl.so: undefined symbol: IID_IUnknown
# Command exited with code 2.
```

ygrek:
It was a bug in camlidl, it is now fixed in [git](https://github.com/xavierleroy/camlidl/commit/875fbe94429c884247c425a0190ca28ed08d9f50). Technically it means that camlidl in bytecode is not useable in current release (and so the change in META in this PR is not effective, but it will help with pins to dev-depo).

kit-ty-kate:
Should we close this PR then?

ygrek:
The current fix is not making situation worse and can be postponed till next camlidl release, true. OTOH it helps people doing `pin --dev-repo camlidl`.
Your call :)

kit-ty-kate:
opam-repository is not really made to encourage use `opam pin --dev`, I'd much rather not break anything and wait for the release

PR #16181, opened 2020-04-11
[new release] bimage, bimage-unix, bimage-gtk and bimage-sdl (0.2.1)

zshipko:
A simple, efficient image-processing library

- Project page: <a href="https://github.com/zshipko/ocaml-bimage">https://github.com/zshipko/ocaml-bimage</a>
- Documentation: <a href="https://zshipko.github.io/ocaml-bimage/doc">https://zshipko.github.io/ocaml-bimage/doc</a>

##### CHANGES:

- Add Func Expr
- Rewrite ops as exprs
- Update stb-image/stb-image-write

PR #16182, opened 2020-04-11
Sync up versions of bimage subpackages

kit-ty-kate:
Fixes revdep failures encountered in https://github.com/ocaml/opam-repository/pull/16181 as well as potential future breakage.

@zshipko is that ok?


zshipko:
Yep, looks good! I appreciate the help :smiley:

PR #16183, opened 2020-04-11
session-upper-bound: add upper bound to session-related libraries

seliopou:
This is in preparation for an upcoming release that will cause these libraries to break (or at least to behave in non-obvious ways) if they depend on the new session library.


seliopou:
Related to inhabitedtype/session#22

kit-ty-kate:
If you usually do your releases in batches (every sublibraries from this repository at once), I'd suggest you sync them up by adding the following constraint `{= version}` to packages from the same repository they depend on, in your future releases.
For instance:
```
  "session" {= version}
```
In the meantime this is good. Thanks!

PR #16185, opened 2020-04-12
Add llvm.10.0.0 and conf-llvm.10.0.0

vaivaswatha:
Mimicing https://github.com/ocaml/opam-repository/pull/14954 with appropriate modifications (version number, links and checksums).

### DIFF
--- a/packages/conf-llvm/conf-llvm.10.0.0/files/configure.sh
+++ b/packages/conf-llvm/conf-llvm.10.0.0/files/configure.sh
@@ -1,0 +1,24 @@
+#!/bin/bash -ex
+
+version=${1/%.?.?/}
+
+if hash brew 2>/dev/null; then
+    brew_llvm_config="$(brew --cellar)"/llvm*/${version}*/bin/llvm-config
+fi
+
+shopt -s nullglob
+for llvm_config in llvm-config-$version llvm-config-${version}.0 llvm-config${version}0 llvm-config-mp-$version llvm-config-mp-${version}.0 llvm${version}-config $brew_llvm_config llvm-config; do
+    llvm_version="`$llvm_config --version`" || true
+    case $llvm_version in
+    $version*)
+        echo "config: \"$llvm_config\"" >> conf-llvm.config
+        echo "version: \"$llvm_version\"" >> conf-llvm.config
+        exit 0;;
+    *)
+        echo "Note: '$llvm_config' doesn't match the required version. Got '$llvm_version' but required '$version'."
+        continue;;
+    esac
+done
+
+echo "Error: LLVM ${version} is not installed."
+exit 1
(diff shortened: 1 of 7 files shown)


kit-ty-kate:
Arf sorry I missed the release. I'll check in details your PR tomorrow as it cannot be tested by the CI (not available on debian stable yet)

kit-ty-kate:
Sorry for taking so long but the patch to link to LLVM dynamically does not apply anymore because the CMake code changed by https://github.com/llvm-mirror/llvm/commit/ab916900b90299735cf58eb3f0f7b6e79de7e837. This whole system using findlib predicate is not really usable anymore (e.g. can't be used when using dune) and poorly documented (by my fault).

I need to spend a bit more time on this to change the layout a bit (unless you'd want to do it instead))

vaivaswatha:
> Sorry for taking so long but the patch to link to LLVM dynamically does not apply anymore because the CMake code changed by [llvm-mirror/llvm@ab91690](https://github.com/llvm-mirror/llvm/commit/ab916900b90299735cf58eb3f0f7b6e79de7e837). This whole system using findlib predicate is not really usable anymore (e.g. can't be used when using dune) and poorly documented (by my fault).
> 
> I need to spend a bit more time on this to change the layout a bit (unless you'd want to do it instead))

Thank you for trying this out @kit-ty-kate . While I understand CMake, LLVM and OCaml, my knowledge on opam (or how it works internally) is almost zero.

I can possibly attempt this with some guidance from you, but I'm not particularly in a hurry for this, So if you think it won't take a lot of your time, I'll just wait it out.

vaivaswatha:
@kit-ty-kate . I've updated the patch and its hash. It worked for me locally, on Ubuntu. Can you test once? Thanks a lot.

kit-ty-kate:
Sorry for the time that it took to deal with this PR. Thanks a lot!

kit-ty-kate:
I thought I would have some time to fix the layout issue too but I guess that'll wait for next release or something. Sorry again

vaivaswatha:
>Sorry for the time that it took to deal with this PR. Thanks a lot!

Thank you @kit-ty-kate for testing and merging this PR.

>I thought I would have some time to fix the layout issue too but I guess that'll wait for next release or something. Sorry again

Not a problem. I'm not sure I even understand the issue :smile: . Thank you very much again.

vaivaswatha:
My `opam` still only lists 9.0.0 (as does this page https://opam.ocaml.org/packages/llvm/). When will I see 10.0.0 there? I assumed that merging this PR should make 10.0.0 available.

kit-ty-kate:
It usually takes an hour or two before the next sync between the official opam repo (also used for caching purpose) and the official git repo

kit-ty-kate:
You can see when the last sync happened here: https://opam.ocaml.org/lastlog.txt

vaivaswatha:
Thank you very much.

Issue #16174, opened 2020-04-10
The compilation of liquidsoap failed


diaverso:
I am using a Centova Cast installer
I get an error when compiling liquidsoap 1.3.7
log:
https://pastebin.com/DJJfRVWu

kit-ty-kate:
cc @toots

PR #16188, opened 2020-04-13
fdkaac.0.3.2: Add conflict on liquidsoap < 1.4.0

toots:
This was forgotten during the last release and fixes: #16174


kit-ty-kate:
I'd rather keep the constraints in the related packages (here `liquidsoap`), would it be ok to do it there instead? It allows a much clearer description of packages and easier to modify later without having to remember where to modify the constraints.

If you don't have time I'll do it at some point.

toots:
> I'd rather keep the constraints in the related packages (here `liquidsoap`), would it be ok to do it there instead? It allows a much clearer description of packages and easier to modify later without having to remember where to modify the constraints.
> 
> If you don't have time I'll do it at some point.

Sure no problem, I'll do it right now.

toots:
Should be good now @kit-ty-kate

kit-ty-kate:
Thanks a lot!

PR #16189, opened 2020-04-13
Package lp.0.0.1

ktahar:
### `lp.0.0.1`
LP and MIP modeling in OCaml
This library is a modeling tool for Linear Programming (LP) and Mixed Integer Programming (MIP).
The model can be exported to CPLEX LP file format, which can be loaded by various solvers.
Importing models from LP file is also supported.

---
* Homepage: https://github.com/ktahar/ocaml-lp
* Bug tracker: https://github.com/ktahar/ocaml-lp/issues

---


kit-ty-kate:
the dune extension `(using menhir 2.1)` you are using, is only available since dune 2.2. Could you return the fix upstream?

kit-ty-kate:
Thanks!

ktahar:
Thank you, too :+1:

PR #16196, opened 2020-04-14
Genprint v0.3

progman1:


kit-ty-kate:
The checksum does not correspond.
```
          - genprint.0.3:
            https://github.com/progman1/genprintlib/archive/v0.3.tar.gz (Bad checksum, expected md5=d5b58cc0aa7e1dd745d94fcc405d9d27)
```

PR #16197, opened 2020-04-14
Mew,mew vi

kandu:
This PR contains the core modal editing engine and its vi specialized complement.

They are pre-requirements for the upcoming vi-mode of utop and lambda-term.


packages/mew_vi/mew_vi.0.1.0/opam
```diff
@@ -0,0 +1,24 @@
...
+homepage: "https://github.com/kandu/mew_vi"
+bug-reports: "https://github.com/kandu/mew_vi/issues"
+license: "MIT"
+dev-repo: "git+https://github.com/kandu/mew_vi.git"
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+]
+depends: [
+  "ocaml" {>= "4.02.3"}
+  "mew"
```
kit-ty-kate, line 13: Would it be ok if all the packages from this repository were to sync all their versions using the `{= version}` constraint?
It allows packages to be more easily maintainable.

kandu: Thanks, I added a version constraint.  :)

kit-ty-kate:
Thanks!

PR #16206, opened 2020-04-15
Package conf-nauty.1.0

zajer:
### `conf-nauty.1.0`
Virtual package relying on nauty
This package can only install if the libnauty library is installed on the system.

---
* Homepage: http://pallini.di.uniroma1.it/
* Bug tracker: https://github.com/zajer/conf-nauty/issues

---

### DIFF
--- a/packages/conf-nauty/conf-nauty.1.0/opam
+++ b/packages/conf-nauty/conf-nauty.1.0/opam
@@ -4,8 +4,10 @@ homepage: "http://pallini.di.uniroma1.it/"
 authors: "http://pallini.di.uniroma1.it/Links.html#contact"
 bug-reports: "https://github.com/zajer/conf-nauty/issues"
 license: "Apache-2.0"
-build: ["pkg-config" "nauty"] {os!="macos"}
-depends: ["conf-pkg-config" {build & os!="macos"}]
+build: ["pkg-config" "nauty"] {os != "macos"}
+depends: [
+  "conf-pkg-config" {build & os != "macos"}
+]
 depexts: [
   ["libnauty2-dev"] {os-family = "debian"}
   ["libnauty-devel"] {os-distribution = "fedora"}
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Meta-packages have no need for archives or sources, they simply live in opam-repository.

I'd suggest you to remove your https://github.com/zajer/conf-nauty git repository unless you want to be voluntary long-term maintainer of this package and use its `issues` for this purpose. Though, this kind of package usually gets modified on the go in opam-repository directly, as directed by people who use them on their own distributions, I suspect the top-down approach of maintenance here won't really work.

zajer:
Thx for taking the time to look at my pull request.

Since I am a complete newbie with opam I didn't know that `opam publish` adds the `url{}` tag to an opam file. Having said that, may I ask you how did you get rid of it? I mean manually or does `opam publish` allow to publish without adding anything to .opam file?

About the maintenance, sure I will remove it. Didn't know I can point someone else to maintain my package without prior asking :) 

Finally, should I just publish a newer version of this package or try to fix the current pull request and, when it pass all checks, do some updates (if any)? I know it's a stupid question by I am not yet familiar with the git workflow.

kit-ty-kate:
> Having said that, may I ask you how did you get rid of it? I mean manually or does `opam publish` allow to publish without adding anything to .opam file?

I simply modified the file from Github and pushed into your branch directly.

> About the maintenance, sure I will remove it. Didn't know I can point someone else to maintain my package without prior asking :)

It's ok no worries.

> Finally, should I just publish a newer version of this package or try to fix the current pull request and, when it pass all checks, do some updates (if any)? I know it's a stupid question by I am not yet familiar with the git workflow.

I've taken care of everything, you don't need to do anything anymore for now. I'll merge this PR now.
In the future if you want to make some adjustments (like define the depexts for other distributions), you simply need to fork this repository (opam-repository) – I believe this is already done – modify the file, push it in a custom branch, click the link given by github to open a pull-request here and it'll be reviewed the same way.
I suppose opam-publish might work but I wouldn't advise for it for conf packages as you usually want to modify current packages, not publish a new version.

Thanks a lot!

PR #16208, opened 2020-04-16
new file:   packages/oranger/oranger.2.0.2

UnixJunkie:


UnixJunkie:
bug correction: some temp files were created but never removed

packages/oranger/oranger.2.0.2/files/compile_ranger.sh
```diff
@@ -0,0 +1,12 @@
...
+set -x
+
+tar xzf 0.9.11.tar.gz
+
+cd ranger-0.9.11/cpp_version
+mkdir build
+cd build
+cmake ../
+make
+cp ranger `opam config var bin`/ml_rf_ranger
```
kit-ty-kate, line 12: I think it would be better if `%{bin}%` was passed as parameter instead

packages/oranger/oranger.2.0.2/opam
```diff
@@ -0,0 +1,36 @@
...
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+]
+install: [
+  ["sh" "install_ranger.sh"]
+]
+depends: [
+  "dune" {>= "1.0.0"}
+  "conf-cmake" {build}
+  "conf-wget" {build}
```
kit-ty-kate, line 17: I don't see any use of wget (and it would be refused by the sandbox anyway if it's used at build time). Is this some sort of runtime dependency or is it just a reminiscent of your local `install_ranger.sh` script (in which case it would be safe to remove)

UnixJunkie: yes, probably no more used

packages/oranger/oranger.2.0.2/files/compile_ranger.sh
UnixJunkie: done

UnixJunkie:
Thanks for the suggestions, all resolved.

kit-ty-kate:
Looks good now. Thanks a lot! Could you return my last fix upstream?

UnixJunkie:
yes, they were up-streamed too.

PR #16215, opened 2020-04-16
Tezos release 7.0~rc1

pirbo:
We can finally release software to participate to the Tezos blockchain in opam "central" repository (up to now, binaries were specific to a specific "chain"). :tada: 

For now, we follow the policy to add the constraint `{ = version }` in between all the packages part of the same tar.bz2 but we're not sure about that and we welcome feedback...

An other question: is it OK to have "meta-package" in the opam-repository? In that case, we would add a package `tezos` with nothing but dependencies to all the binaries (tezos-node, tezos-client, tezos-{accuser,baker,endorser}-006-PsCARTHA) and optional dependencies.

### DIFF
--- a/packages/tezos-accuser-006-PsCARTHA/tezos-accuser-006-PsCARTHA.7.0~rc1/opam
+++ b/packages/tezos-accuser-006-PsCARTHA/tezos-accuser-006-PsCARTHA.7.0~rc1/opam
@@ -1,0 +1,27 @@
+opam-version: "2.0"
+maintainer: "contact@tezos.com"
+authors: [ "Tezos devteam" ]
+homepage: "https://www.tezos.com/"
+bug-reports: "https://gitlab.com/tezos/tezos/issues"
+dev-repo: "git+https://gitlab.com/tezos/tezos.git"
+license: "MIT"
+depends: [
+  "tezos-tooling" { with-test & = version }
+  "dune" { >= "1.11" }
+  "tezos-baking-006-PsCARTHA-commands" { = version }
+  "tezos-client-base-unix" { = version }
+]
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+  ["mv" "src/proto_006_PsCARTHA/bin_accuser/%{name}%.install" "./"]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+]
+synopsis: "Tezos/Protocol: 006_PsCARTHA accuser binary"
+
+url {
+  src: "https://gitlab.com/tezos/tezos/-/archive/v7.0-rc1/tezos-v7.0-rc1.tar.bz2"
+  checksum: [
+    "sha256=89385ff254582d64b909bcd33f10068f1ba5fe8a47802bb0bcf983f30f7eb49b"
+    "sha512=413bb72b556de80fadfdf3278a6c674a471a614af6adf19c9b383cbc2ee0d081f0421156c92fa917808d47eccaed4a369fc9613a267d5f77a6c3b8e928e5f244"
+  ]
+}
--- a/packages/tezos-tooling/tezos-tooling.7.0~rc1/opam
+++ b/packages/tezos-tooling/tezos-tooling.7.0~rc1/opam
@@ -1,0 +1,25 @@
+opam-version: "2.0"
+maintainer: "contact@tezos.com"
+authors: [ "Tezos devteam" ]
+homepage: "https://www.tezos.com/"
+bug-reports: "https://gitlab.com/tezos/tezos/issues"
+dev-repo: "git+https://gitlab.com/tezos/tezos.git"
+license: "MIT"
+depends: [
+  "dune" { >= "1.11" }
+  "ocamlformat" { = "0.10" }
+]
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+  ["mv" "src/tooling/%{name}%.install" "./"]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+]
+synopsis: "Tezos: tooling for the project"
+
+url {
+  src: "https://gitlab.com/tezos/tezos/-/archive/v7.0-rc1/tezos-v7.0-rc1.tar.bz2"
+  checksum: [
+    "sha256=89385ff254582d64b909bcd33f10068f1ba5fe8a47802bb0bcf983f30f7eb49b"
+    "sha512=413bb72b556de80fadfdf3278a6c674a471a614af6adf19c9b383cbc2ee0d081f0421156c92fa917808d47eccaed4a369fc9613a267d5f77a6c3b8e928e5f244"
+  ]
+}
--- a/packages/tezos-workers/tezos-workers.7.0~rc1/opam
+++ b/packages/tezos-workers/tezos-workers.7.0~rc1/opam
@@ -1,0 +1,26 @@
+opam-version: "2.0"
+maintainer: "contact@tezos.com"
+authors: [ "Tezos devteam" ]
+homepage: "https://www.tezos.com/"
+bug-reports: "https://gitlab.com/tezos/tezos/issues"
+dev-repo: "git+https://gitlab.com/tezos/tezos.git"
+license: "MIT"
+depends: [
+  "tezos-tooling" { with-test & = version }
+  "dune" { >= "1.11" }
+  "tezos-base" { = version }
+]
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+  ["mv" "src/lib_workers/%{name}%.install" "./"]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+]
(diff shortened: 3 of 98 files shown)


avsm:
Glad to have your here! 

There's a small bug in the generation scripts I think:

`tezos-accuser-006-PsCARTHA/tezos-accuser-006-PsCARTHA.7.0~rc1/tezos-accuser-006-PsCARTHA.7.0~beta/`

You should only have one level of versioned directory there.

It's fine (and encouraged) to set =version for the packages generated from the same repo, as that's how they will be tested together.

A tezos metapackage is just fine, and a good user experience (`opam install tezos`). Feel free to add that to this PR.

packages/tezos-accuser-006-PsCARTHA/tezos-accuser-006-PsCARTHA.7.0~rc1/opam
```diff
@@ -0,0 +1,27 @@
...
+license: "MIT"
+depends: [
+  "tezos-tooling" { with-test & = version }
+  "dune" { >= "1.11" }
+  "tezos-baking-006-PsCARTHA-commands" { = version }
+  "tezos-client-base-unix" { = version }
+]
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+  ["mv" "src/proto_006_PsCARTHA/bin_accuser/%{name}%.install" "./"]
```
avsm, line 16: It would be good if you could get rid of this line in the packages. In normal operation, `dune build -p` will build only the install file for the package you have requested.  Do you need this explicit install override for some other reason?  We may be able to just fix that directly in dune if there's an underlying limitation.

packages/tezos-tooling/tezos-tooling.7.0~rc1/opam
```diff
@@ -0,0 +1,25 @@
+opam-version: "2.0"
+maintainer: "contact@tezos.com"
+authors: [ "Tezos devteam" ]
+homepage: "https://www.tezos.com/"
+bug-reports: "https://gitlab.com/tezos/tezos/issues"
+dev-repo: "git+https://gitlab.com/tezos/tezos.git"
+license: "MIT"
+depends: [
+  "dune" { >= "1.11" }
+  "ocamlformat" { = "0.10" }
```
avsm, line 10: Do you need to propagate this ocamlformat constraint to all users of the package? It may be better to add `{dev}` here so that only pinned packages (which can then be edited) get the ocamlformat version constraint.

packages/tezos-workers/tezos-workers.7.0~rc1/opam
```diff
@@ -0,0 +1,27 @@
+opam-version: "2.0"
+maintainer: "contact@tezos.com"
+authors: [ "Tezos devteam" ]
+homepage: "https://www.tezos.com/"
+bug-reports: "https://gitlab.com/tezos/tezos/issues"
+dev-repo: "git+https://gitlab.com/tezos/tezos.git"
+license: "MIT"
+depends: [
+  "tezos-tooling" { with-test & = version }
+  "ocamlfind" { build }
```
avsm, line 10: Are these ocamlfind dependencies actually necessary?  (i.e. is there a Findlib or `ocamlfind` invocation from a dune rule?).  Dune includes the required ocamlfind functionality for the common case, so you may not need this.

pirbo: That's clearly a remaining of the past, thank you.

packages/tezos-tooling/tezos-tooling.7.0~rc1/opam
pirbo: `tezos-tooling` is only a `{ with-test }` dependency of other packages so "normal users" should never have it installed.
On the contrary, checking that the code follows ocamlformat rules is part of the tests and ocamlformat rules changes with version so yes, sadly, the tests fails as soon as you relax this constraint.
I'm not the expert here so if there's a better solution, I'll implement it

packages/tezos-accuser-006-PsCARTHA/tezos-accuser-006-PsCARTHA.7.0~rc1/opam
pirbo: Very good question about general organization of the repository. I always wanted to discuss it more broadly with the community without daring :-)

These `mv` are necessary because opam files are not at the root of the repository and dune generates `.install` files next to their corresponding opam file.

In addition of requiring these ugly `mv`, I think (but I'm not sure I wanted to try) that not having the opam files at the root of the repository breaks `opam pin add --dev` which is also unfortunate.

On the contrary, as you see the tezos repository defines 80+ opam packages so having 80+ more files at the root would quite pollute it . Moreover, some part of the code ("economic protocols" in our terminology corresponding to the folders `src/proto_*`) are supposed to be (un)pluggable (self contained) plugins, if they are there, software is able to understand/follow the corresponding protocol, if they aren't, you have valid binaries but that don't know about these specific rulesets (but can ask them to their "peers"/compile them/dynamically load them on necessity... an other story). If the opam files corresponding to the libraries of the protocol "foobar" has now to be at the root of the repo, adding/removing protocol "foobar" is not as simple as adding/removing folder `src/proto_foobar` anymore and that's annoying.

So I'm confused about what direction to follow...

pirbo:
For the record. I'm investigating the test failure of `tezos-lmdb` on MacOS in travisCI but on the MacOS computer I have access to, these tests work perfectly so I'm confused and tempted to use the joker "travis-ci specific issue"...

packages/tezos-accuser-006-PsCARTHA/tezos-accuser-006-PsCARTHA.7.0~rc1/opam
hnrgrgr: Hi @pirbo, glad you were able to release tezos here :-)

The next version of opam should add better support for non-toplevel `.opam`. See: https://github.com/ocaml/opam/pull/3499 Basically, it allow to add a field `subpath` in the package description, meaning: this package only requires this fragment of the `.tar.gz` (or the repository) in order to build. So `opam` only extracts (or fetch) the required `subpath` and build directly in the subdirectory. So, from the dune perspective, both`.opam` and `.install` are at toplevel.

And, to fix the daily workflow, it adds `opam pin add --dev --rec`: il will lookup for `.opam` in sub-directories.

I have no idea whenever the next (feature-)release of opam might happen but, meanwhile, maybe this ugly `mv` is acceptable. And, maybe, you are the perfect beta-tester for this new feature :-)

pirbo: Fantastic! 
To hear from you first and foremost but also this feature looks to be exactly what we want indeed. No surprise given its original author. :-D
I agree that the `mv` are acceptable in the meantime.

avsm: I'm not quite convinced that the behaviour of dune is quite correct here, though. If the `dune build -p` works, then the semantics of `.install` files are that they should be generated at the root of the project to be useful.  If you could open up an issue over at https://github.com/ocaml/dune/issues about this, we can take a look at supporting this properly.

We'll be kicking off the opam 2.1 release process in the next weeks, to answer @hnrgrgr's question above.  But it would be good to find a solution that'll work with opam 2.0.x as well, since it is widely packaged on OS distributions.

packages/tezos-tooling/tezos-tooling.7.0~rc1/opam
avsm: Making it a `with-test` dependency is just fine.  The majority of opam installations (aside from CI) will not use that.

avsm:
This Mac failure is probably sandboxing-related. Are the lmdb tests trying to write somewhere external to the ~/.opam area?

pirbo:
Yes, MacOS failure was related to an hardwired reference to `/tmp`. I've added a commit that fixes it by replacing "/tmp" by Filename.get_temp_dir.

avsm:
Sounds good. I've just triggered a rebuild of this PR to look at the revdeps more closely, as some failed to build due to the volume causing the Docker GC to kick in.

avsm:
This broadly all looks good to go now. There are quite a few spurious CI failures, but I double-checked that ledgerwallet-tezos will install on CentOS (I suspect conf-hiadpi needs OpenSUSE depext metadata to install on OpenSUSE, but this isn't a blocker to merge).

Do you want to push a non-draft tezos.7 set of packages @pirbo?

pirbo:
Thank you very much for all your help.
It has been decided to not merge the rc1 in opam-repository in the end. See you in 2 weeks (hopefully :-) ) for the final v7.0...

avsm:
Look forward to it. Good luck with the release!

PR #16217, opened 2020-04-17
Add conflicts on flac >= 0.1.5 for liquidsoap 1.3.x

toots:
_When it rains it pours..._

Apparently, we also missed some conflicts when releasing `flac.0.1.6`: savonet/liquidsoap#1165

This PR adds the required version conflicts on `liquidsoap` and not on `flac` as it was done previously.


kit-ty-kate:
Thanks a lot!

PR #16218, opened 2020-04-17
Package onauty.0.5

zajer:
### `onauty.0.5`
OCaml bindings to nauty library
Library for creation and basic manipulation of graphs and digraps. Library uses

---
* Homepage: https://github.com/zajer/onauty
* Source repo: git+https://github.com/zajer/onauty.git
* Bug tracker: https://github.com/zajer/onauty/issues

---

### DIFF
--- a/packages/onauty/onauty.0.5/opam
+++ b/packages/onauty/onauty.0.5/opam
@@ -1,0 +1,37 @@
+opam-version: "2.0"
+synopsis: "OCaml bindings to nauty library"
+description:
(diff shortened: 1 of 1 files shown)


zajer:
Can somebody please tell me how can I fix this package so that it passes all checks?
So far I understood that `ounit2` library was not found during linux test and the source code could not be downloaded in case of macos test.
I guess the former can be fixed by declaring dependency to ounit with variable `with-test`.
Unfortunately I have no clue how to fix the latter, I just checked and had no problems downloading the source code.

PR #16221, opened 2020-04-18
Package conf-nauty.1.1

zajer:
Restricted availability to ubuntu,debian and fedora distros.


zajer:
I guess 7 hours of tests for a conf-package is indicating that something is wrong. 
Can someone please point out what should I do to make it work ?

Or maybe everything is fine ?

PR #16223, opened 2020-04-18
update lmdb to 1.0

madroach:


madroach:
For RHEL and CentOS lmdb-devel would need to be installed from [EPEL](https://fedoraproject.org/wiki/EPEL). Is this possible via ``depext`` ?
The build failure on Ubuntu xenial is due to a bug in the outdated backend lmdb 0.9.17.

packages/lmdb/lmdb.1.0/url
```diff
@@ -0,0 +1,5 @@
+src: "https://github.com/Drup/ocaml-lmdb/archive/1.0.tar.gz"
```
kit-ty-kate, line 1: `url` files are deprecated by opam 2.0. Could you remove it and insert it in a `url` field instead? (see any other packages for the syntax)

kit-ty-kate:
Could you also return my fix upstream?

madroach:
Sure I will merge this back upstream. Why did you remove the build tag on dune ?

kit-ty-kate:
> Sure I will merge this back upstream. Why did you remove the build tag on dune ?

dune is only partially backward compatible. If it were to be downgraded for any reason, all the packages that were installed with a newer version would be totally unusable and the whole switch would have to be rebuilt. See #14266 and https://github.com/ocaml/dune/issues/2147 for more informations.

kit-ty-kate:
Thanks a lot!

madroach:
Ok. I get it about the build tag on dune. But why remove it for dune-configurator ?

kit-ty-kate:
mmh, good point, I guess `dune-configurator` can still be a `{build}` dependency but in the general case it doesn't usually matter as `dune` and `dune-configurator` usually upgrades in pair.

PR #16224, opened 2020-04-19
Package conf-nauty.1.1

zajer:
Added restrictions on distros.


kit-ty-kate:
Looking at the difference between conf-nauty.1.0 and conf-nauty.1.1 I'm not so sure this deserve to be a new version:
```
--- conf-nauty/conf-nauty.1.0/opam	2020-04-20 16:28:35.581826498 +0100
+++ conf-nauty/conf-nauty.1.1/opam	2020-04-21 13:29:14.878319418 +0100
@@ -1,17 +1,17 @@
 opam-version: "2.0"
-maintainer: "pikus3@list.pl"
+version: "1.1"
+maintainer: "Piotr Cybulski <pikus3@list.pl>"
 homepage: "http://pallini.di.uniroma1.it/"
 authors: "http://pallini.di.uniroma1.it/Links.html#contact"
 license: "Apache-2.0"
-build: ["pkg-config" "nauty"] {os != "macos"}
-depends: [
-  "conf-pkg-config" {build & os != "macos"}
-]
+build: ["pkg-config" "nauty"] 
+depends: ["conf-pkg-config"]
 depexts: [
-  ["libnauty2-dev"] {os-family = "debian"}
+  ["libnauty2-dev"] {os-distribution = "ubuntu" | os-distribution = "debian" }
   ["libnauty-devel"] {os-distribution = "fedora"}
 ]
+available: [ os-distribution = "ubuntu" | os-distribution = "fedora" | os-distribution = "debian" ]
 synopsis: "Virtual package relying on nauty"
 description:
   "This package can only install if the libnauty library is installed on the system."
```
The change in the `available` field seems a little excessive. We want for people who install libnauty from source to also be able to install your packages.

I'm not sure what was the goal of these changes.

zajer:
The 1.0 version can be installed on systems that do not actually have libnauty on them.
(I noticed it when pull requested `onauty`, `conf-nauty` was installed on osx system and yet c stubs couldn't be compiled)
The 1.1 actually restricts its availability to distros that I know for sure have libnauty available from the distribution's default package repository.
In case of new systems providing libnauty library , this version can be updated. 
In case someone build libnauty from source then they can still use version 1.0 of this package.

Disclaimer: I don't know where too look for packages available on osx so the `available` list might be incomplete.
For ubuntu, debian and fedora I found this:
https://packages.ubuntu.com/eoan/libnauty2-dev
https://packages.debian.org/search?keywords=libnauty2-dev
https://apps.fedoraproject.org/packages/libnauty-devel

kit-ty-kate:
>  In case someone build libnauty from source then they can still use version 1.0 of this package.

Well, the thing is that given it's unavailable to them they will use `conf-naughty.1.0` anyway. The 1.1 version doesn't really bring anything new.
I'd rather update 1.0 by adding a warning in case one of the distributions that it cannot check is the current one.

zajer:
Could you please tell me how can I add such a warning to a package? 
And how does it impact Travis tests? 
I mean specifically tests on osx. 
I want them passed, but because I don't know how to check availability of the library on certain systems all packages depending on conf-nauty will faill their own tests.

mseri:
@zajer you could add something like:

```
post-messages: [
  "Unable to automatically install libnauty for your system. Please make sure that libnauty is installed before relying on this package or in case of linking errors.. " { os-distribution != "ubuntu" & os-distribution != "fedora" & os-distribution != "debian" ]}
]
```

Then it will successfully install everywhere like version 1.0 but warn those users with other systems

zajer:
@mseri 
Thx for the answer!

In regard to adding the warning as a solution.
How will checks on osx pass for a library that uses _conf-nauty_ ?
Last time I checked they didn't and while the library, that I was trying to publish, was working fine I couldn't get it merged because of it.

mseri:
Oh sorry, it was an oversight. Since the package is checked by using pkg-config, I would tell the users to set the right PKG_CONFIG_PATH before proceeding with the installation. Something like:

```
  "Unable to automatically install libnauty for your system. Please make sure that libnauty is installed and set PKG_CONFIG_PATH accordingly when relying on it." { os-distribution != "ubuntu" & os-distribution != "fedora" & os-distribution != "debian" ]}
```

It will fail on osx Travis CI, but it will tell the users how to fix the issue and then try again

zajer:
Ok I will modify the existing version of _conf-nauty_ and make another pull request with suggested changes.

Thank you both for support.

PR #16227, opened 2020-04-19
Changes required to publish tezos-node

pirbo:
- conf-hidapi may still fail on alpine as hidapi is only available in the most recent alpine 3.11
- the compatibility of ledgerwallet with dune 1.11 won't be test by the CI I guess but I checked it and you can see it in CI logs of !16215 (meta packages `tezos` installs at the same time ledgerwallet-tezos and tezos-node which transitively imposes dune < 2)


kit-ty-kate:
Thanks!

avsm:
We have an override in the CI to add edge packages to Alpine, as several other packages need it too.

PR #16199, opened 2020-04-14
Genprint v0.3

progman1:
library for printing of values vis a vis the ocaml interpreter, but from within compilation units.

PR #16231, opened 2020-04-19
Genprint v0.4

progman1:

### DIFF
--- a/packages/genprint/genprint.0.3/opam
+++ b/packages/genprint/genprint.0.3/opam
@@ -26,7 +26,7 @@ depends: [
 ]
 
 url {
-  src: "https://github.com/progman1/genprintlib/archive/v0.3.tar.gz"
+  src: "https://github.com/progman1/genprintlib/archive/v0.3a.tar.gz"
 
-  checksum: "md5=0f577477a139c27de8acbc9d5a878562"
+  checksum: "md5=d5b58cc0aa7e1dd745d94fcc405d9d27"
 }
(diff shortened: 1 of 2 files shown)


packages/genprint/genprint.0.3/opam
```diff
@@ -26,6 +26,7 @@ depends: [
 ]
 
 url {
-  src: "https://github.com/progman1/genprintlib/archive/v0.3a.tar.gz"
-  checksum: "md5=d5b58cc0aa7e1dd745d94fcc405d9d27"
+  src: "https://github.com/progman1/genprintlib/archive/v0.3.tar.gz"
+
+  checksum: "md5=0f577477a139c27de8acbc9d5a878562"
```
kit-ty-kate, line 31: I'm sorry but we already talked about this in https://github.com/ocaml/opam-repository/pull/16199. Could you please not change previous releases? Your new `genprint.0.4` is fine though, why change the previous one?

progman1: I updated the checksum but forgot to bump the src line release file number so opam file pointed to wrong tarball.

progman1: the process is confusing - that was back when I tried to correct the 0.3 release thinking it was not even installable.
If I revert the details, is that safe to do?!

kit-ty-kate: yes

kit-ty-kate:
Thanks

PR #16233, opened 2020-04-19
Release of Alba 0.4.2

hbr:
new file:   packages/alba/alba.0.4.2/opam


kit-ty-kate:
According to your `dune-project` file, this package requires at least dune 1.10

kit-ty-kate:

hbr:
Thanks for fixing the dependency constraint and merging.

Is there a way for me to make these checks before I generate the pull request? I have built my project locally with `opam pin add .` and the build process did not report any inconsistency. Can I run some checks similar to the checks done by camelus?

Thanks for any hints
Helmut

PR #16236, opened 2020-04-19
ocaml-secondary-compiler: --disable-graph-lib (disable Graphics)

### LINKED ISSUE #16226: ERROR while compiling ocaml-secondary-compiler.4.08.1
Hi, I am currently trying to install `ocaml-secondary-compiler.4.08.1`, and opam gave an error while building with the following error message. 

```
[ERROR] The compilation of ocaml-secondary-compiler failed at "/Users/irene/.opam/opam-init/hooks/sandbox.sh build make -j3 world.opt".

#=== ERROR while compiling ocaml-secondary-compiler.4.08.1 ====================#
# context     2.0.5 | macos/x86_64 | ocaml-variants.4.02.2+multicore | https://opam.ocaml.org#8e32fa53
# path        ~/.opam/4.02.2+multicore/.opam-switch/build/ocaml-secondary-compiler.4.08.1
# command     ~/.opam/opam-init/hooks/sandbox.sh build make -j3 world.opt
# exit-code   2
# env-file    ~/.opam/log/ocaml-secondary-compiler-77298-1114b2.env
# output-file ~/.opam/log/ocaml-secondary-compiler-77298-1114b2.out
### output ###
# [...]
# ../../boot/ocamlrun ../../ocamlc -nostdlib -I ../../stdlib -c -absname -w +a-4-9-41-42-44-45-48 -warn-error A -bin-annot -g -safe-string -strict-sequence -strict-formats  graphicsX11.mli
# ../../boot/ocamlrun ../../ocamlc -nostdlib -I ../../stdlib -c -absname -w +a-4-9-41-42-44-45-48 -warn-error A -bin-annot -g -safe-string -strict-sequence -strict-formats  graphics.ml
# ../../boot/ocamlrun ../../ocamlc -nostdlib -I ../../stdlib -c -absname -w +a-4-9-41-42-44-45-48 -warn-error A -bin-annot -g -safe-string -strict-sequence -strict-formats  graphicsX11.ml
# ../../boot/ocamlrun ../../tools/ocamlmklib -oc graphics open.o draw.o fill.o color.o text.o image.o make_img.o dump_img.o point_col.o sound.o events.o subwindow.o -ldopt "-L/usr/local/lib -lX11"
# ld: library not found for -lX11
# clang: error: linker command failed with exit code 1 (use -v to see invocation)
# make[4]: *** [libgraphics.a] Error 2
# make[4]: *** Waiting for unfinished jobs....
# make[3]: *** [all] Error 2
# make[2]: *** [otherlibraries] Error 2
# make[1]: *** [opt.opt] Error 2
# make: *** [world.opt] Error 2
```

```
<><> ocaml-secondary-compiler.4.08.1 troubleshooting ><><><><><><><><><><><>  🐫
=> A failure in the middle of the build may be caused by build parallelism
      (enabled by default).
Please file a bug report at https://github.com/ocaml/ocaml/issues
```

I tried installing `reagents` again using `--jobs=1` to force a sequential build, and it still gave me the same error message. I'm currently on `ocaml` version 4.02.2 with support for Multicore OCaml (`ocaml-variants` 4.02.2+multicore).

gasche:
Fixes #16226

cc @dra27 @diml: do you need ocamldoc in the secondary compiler, or can it also be disabled? (I sort of assume that you do, given that @dra27 made the effort of backporting the manpages patch?)

### DIFF
--- a/packages/ocaml-secondary-compiler/ocaml-secondary-compiler.4.08.1/opam
+++ b/packages/ocaml-secondary-compiler/ocaml-secondary-compiler.4.08.1/opam
@@ -15,6 +15,7 @@ build: [
       "--disable-installing-bytecode-programs"
       "--disable-debug-runtime"
       "--disable-instrumented-runtime"
+      "--disable-graph-lib"
       "CC=cc" {os = "openbsd" | os = "freebsd" | os = "macos"}
       "ASPP=cc -c" {os = "openbsd" | os = "freebsd" | os = "macos"}
   ]


diml:
No, we only need unix

dra27:
The packaging of `ocaml-secondary-compiler` was done to be generic enough that it could be useful for other projects which needs to bootstrap themselves with a compiler newer than the switch's (for example, I know ocamlformat has been looking at having a mode with bundled dependencies which can build this way using ocaml-secondary-compiler). So I left `ocamldoc` in because I had no justification in general to disable the tool.

My intention was that when `ocaml-secondary-compiler` adds a version supporting `--disable-stdlib-manpages` that we'd add that.

This fix is fine - there's obviously a bug in 4.08, because `configure` should have automatically disabled the graphics library for the OP, but I see no harm in this PR, therefore.

Only out of an abundance of caution, though, I was put this in `ocaml-secondary-compiler.4.08.1-1` (cf `ocamlfind.1.7.3-1`) just so it's clear that the packaging changed.

dra27:
@AltGr - is that Camelus warning correct?!

gasche:
I am timing out on this PR: I thought it would be simple but it looks like it is going to involve more work, and I'm not available enough right now to do it. So I will close for now; anyone, please feel free to adopt the change (it's very simple).

PR #16238, opened 2020-04-20
new file:   packages/hts_shrink/hts_shrink.2.1.1

UnixJunkie:


UnixJunkie:
made compatible with latest versions of dependencies

packages/hts_shrink/hts_shrink.2.1.1/opam
```diff
@@ -0,0 +1,38 @@
...
+dev-repo: "git+https://github.com/UnixJunkie/hts_shrink.git"
+license: "BSD-3-Clause"
+build: ["dune" "build" "-p" name "-j" jobs]
+depends: [
+  "dune"
+  "batteries"
+  "dolog"
+  "parmap"
+  "parany"
+  "minicli"
```
kit-ty-kate, line 15: I believe these dependencies need a lower-bound check as they became incompatible somewhat recently (e.g. since 5.0.0 for `minicli`)

packages/hts_shrink/hts_shrink.2.1.1/opam
```diff
@@ -0,0 +1,38 @@
+opam-version: "2.0"
+authors: "Francois Berenger"
+maintainer: "unixjunkie@sdf.org"
+homepage: "https://github.com/UnixJunkie/hts_shrink"
+bug-reports: "https://github.com/UnixJunkie/hts_shrink/issues"
+dev-repo: "git+https://github.com/UnixJunkie/hts_shrink.git"
+license: "BSD-3-Clause"
+build: ["dune" "build" "-p" name "-j" jobs]
+depends: [
+  "dune"
```
kit-ty-kate, line 10: Could the `dune-project` file be included in the source for the next release? I think dune 1.0 is fine here but it might change in the future

packages/hts_shrink/hts_shrink.2.1.1/opam
UnixJunkie: Probably, but unless there is a tool that allows to discover those bounds automatically,
I am not going to look for them.

packages/hts_shrink/hts_shrink.2.1.1/opam
UnixJunkie: Ok, the dune-project file was added to the repository upstream.

kit-ty-kate:
Could you return the fix upstream? Also your minimal required version of dune upstream is set to dune 2.5. This might not be what you want (it's fine for me either way, but just a heads up)

UnixJunkie:
I'll add those upstream.

Did you use a tool to find the bounds automatically?
This requires at least one O(log(n)) search for each dependency; with n the number of versions
of this dependency. I hope you did not do this by hand.

About the dune version, this is a dune problem:  dune should not report a version higher
than what is strictly needed to parse the dune file...

UnixJunkie:
Related to https://github.com/ocaml/dune/issues/3411

kit-ty-kate:
Oops sorry I forgot to merge this PR. Thanks!

kit-ty-kate:
> Did you use a tool to find the bounds automatically?
> This requires at least one O(log(n)) search for each dependency; with n the number of versions
> of this dependency. I hope you did not do this by hand.

Well no, I just did them by memory. I just remembered those packages changed their whole interface when they switched to jbuilder/dune so I just put the first version to be switched to jbuilder/dune for each, no test. It's not very accurate but it's better than nothing.

In the future there will be a CI check to check for lower-bounds anyway, we'll be sure at that point.

PR #5898, opened 2016-03-07
conf-ncurses: Do not depend on `conf-pkg-config` when not required

OCamlPro-Henry:
This is a fix for:

  https://github.com/ocaml/opam-repository/issues/5880#issuecomment-193480687

PR #16242, opened 2020-04-20
merge conf-ncurses.1 and ncurses.1+system

avsm:
It's unnecessary to have the solver for this, when one package can express the right dependencies fine.  This in turns helps with static analysis of a packages dependencies when we run opam on a host other than the distribution target.


avsm:
@dra27 does this look ok to you?

dra27:
The existence of these two packages was a hangover from opam 1.x days - see #5898

dra27:
LGTM, yes

avsm:
Thanks!

PR #16243, opened 2020-04-20
2 packages from project-everest/hacl-star at 0.1

victor-dumitrescu:
This pull-request concerns:
-`hacl-star.0.1`: OCaml API for EverCrypt/HACL*
-`hacl-star-raw.0.1`: Auto-generated low-level OCaml bindings for EverCrypt/HACL*

---
* Homepage: https://hacl-star.github.io/
* Source repo: git+https://github.com/project-everest/hacl-star.git
* Bug tracker: https://github.com/project-everest/hacl-star/issues

---


packages/hacl-star/hacl-star.0.1/opam
```diff
@@ -0,0 +1,23 @@
...
+  "ocaml" {>= "4.05.0"}
+  "hacl-star-raw"
+  "zarith"
+]
+build: ["dune" "build" "-p" name "-j" jobs]
+run-test: ["dune" "runtest"]
+dev-repo: "git+https://github.com/project-everest/hacl-star.git"
+url {
+  src:
+    "https://github.com/project-everest/hacl-star/raw/master/bindings/ocaml/archive/hacl-star.0.1.tar.gz"
```
kit-ty-kate, line 18: This is a little sketchy. Could you ensure this archive won't ever change? Could it not be a normal git tag / extra download instead?

kit-ty-kate:
I really don't recommand to place archives in your git repository, github or git itself already generate it for you, it increases the size of your repository by a good factor each time you want to release something and it makes it too easy to change it and that would break the opam package.
I don't think I would be comfortable merging this as is.

Otherwise I've made a few changes to your packages (a few missing dependencies, parameters, ...) if you agree on this change, can these be returned upstream?
Also I've made so that `hacl-star.0.1` always depends on `hacl-star-raw.0.1` so the two packages are linked. It avoids mistakes later down the line for future releases, but it's optional if you want to explicitly control your dependencies you can revert my change.

kit-ty-kate:
oh also there are a couple of failure on MacOS:
```
#=== ERROR while compiling hacl-star-raw.0.1 ==================================#
# context     2.0.6 | macos/x86_64 | ocaml-system.4.09.0 | file:///Users/travis/build/ocaml/opam-repository
# path        ~/.opam/ocaml-system/.opam-switch/build/hacl-star-raw.0.1
# command     ~/.opam/opam-init/hooks/sandbox.sh build make -C raw
# exit-code   2
# env-file    ~/.opam/log/hacl-star-raw-60549-865888.env
# output-file ~/.opam/log/hacl-star-raw-60549-865888.out
### output ###
# ./curve25519-inline.h:595:5: error: inline assembly requires more registers than available
# [...]
#     ^
# ./curve25519-inline.h:595:5: error: inline assembly requires more registers than available
# ./curve25519-inline.h:221:5: error: inline assembly requires more registers than available
#     "  movq 0(%1), %%rdx;"
#     ^
# ./curve25519-inline.h:595:5: error: inline assembly requires more registers than available
#     "  movq 0(%1), %%rdx;"                                       // f[0]
#     ^
# ./curve25519-inline.h:595:5: error: inline assembly requires more registers than available
# 5 errors generated.
# make: *** [Hacl_Curve25519_64.o] Error 1
```
and on ocaml 4.09 with local install:
```
- (cd _build/default/tests && ./ecdsa_test.exe)
- Fatal error: exception Failure("[Hacl.ECDSA Test 1 ] Failure: Signing")
-    aead_test alias tests/runtest
```

victor-dumitrescu:
Thanks for your review and changes! We're still trying to figure out what's causing the build issues. Once we do, I'll also update the package description to use a Git tag instead.

msprotz:
@kit-ty-kate do you know why OPAM uses specifically XCode 10.1 in Travis CI? We seem to be hitting a compiler failure that's only reproducible with this exact version of XCode -- this requires me to add extra configure checks and probes to detect this particular failure, and seems to be gone for later versions of XCode

I'm adding them nonetheless, but just wondering how important XCode 10.1 is vs, say, XCode 10.3 or XCode 11

Thanks!

kit-ty-kate:
> @kit-ty-kate do you know why OPAM uses specifically XCode 10.1 in Travis CI? We seem to be hitting a compiler failure that's only reproducible with this exact version of XCode -- this requires me to add extra configure checks and probes to detect this particular failure, and seems to be gone for later versions of XCode

Mh, probably no other reasons other than "we never had any issue with this and I didn't even know this version was outdated or that we had to select the version of xcode in travis". Sorry for that. I'll propose a PR to change this.

avsm:
The choice of xcode version was due to travis not having uptodate versions for a long period of time, and then deploying updates more regularly. I last updated it in https://github.com/ocaml/opam-repository/commit/977a15ba1f74aa3942535b0ba42dca4905f1ff98, and it's certainly good to keep this matching the latest macOS release as best as we can.

mseri:
Travis is using a newer xcode now afaik. @victor-dumitrescu can you update this PR?

victor-dumitrescu:
All the issues should now be fixed, but before being merged I will resubmit to point to a tag on the main HACL* repository (currently it's from a fork).

msprotz:
All green! Congratulations @victor-dumitrescu

kit-ty-kate:
Thanks a lot!

PR #16245, opened 2020-04-20
fixed spotlib.4.0.3

camlspotter:


camlspotter:
Thanks for pointing the diffs.  For this, I think I should retrieve the original version.

PR #16257, opened 2020-04-22
Lwt 5.3.0

aantron:
Changelog:

> - Add `let*` and `and*` in `Lwt.Syntax` and `Lwt_result.Syntax` (ocsigen/lwt#775, Rahul Kumar).
> - Add `Lwt_result.both` (ocsigen/lwt#775, Rahul Kumar).
> - Always use libev if detected (ocsigen/lwt#601).


aantron:
Closing, because https://github.com/ocsigen/lwt/pull/776 will also add `let+`. I will retag and send 5.3.0 again. cc @CraigFe

PR #16258, opened 2020-04-22
Lambda Soup 0.7.1: HTML scraping with CSS

aantron:
- Prepend `<!DOCTYPE html>` to full documents (aantron/lambdasoup#32, prompted Fabian Hemmer).


kit-ty-kate:
Thanks!

PR #16259, opened 2020-04-22
fix for spotlib.4.0.3

camlspotter:


kit-ty-kate:
The archive had no difference apart from the traditional hg file. Thanks!

PR #16265, opened 2020-04-22
Update the libtorch package to 1.5

LaurentMazare:
PyTorch 1.5 has recently been released, update the libtorch package to support this version.


kit-ty-kate:
Thanks

PR #16268, opened 2020-04-23
Lwt 5.3.0

aantron:
Changelog:

> - Add `let*` and `and*` in `Lwt.Syntax` and `Lwt_result.Syntax` (ocsigen/lwt#775, Rahul Kumar).
> - Also add `let+` and `and+` (ocsigen/lwt#776, Craig Ferguson).
> - Add `Lwt_result.both` (ocsigen/lwt#775, Rahul Kumar).
> - Always use libev if detected (ocsigen/lwt#601).


kit-ty-kate:
Thanks a lot!

PR #16272, opened 2020-04-23
Development packages for OCaml 4.10.1 and 4.12.0

Octachron:
and update the synopsis for 4.11.0 development packages after #16261 .

cc @damiendoligez .


kit-ty-kate:
Thanks a lot!

PR #16278, opened 2020-04-23
Remove {build} flags from all ppx_* dependencies.

paurkedal:
This was suggested by @avsm in #11852, which had a more narrow scope addressing the specific issue with having `ppx_sexp_conv` as a build dependency. I agree with extending the scope, though we may want to discuss the approach before merging.  I am not familiar with most of the packages involved, so what I have done here is to remove the `{build}` flag on every dependency on a package starting with `ppx_*`, as well as `bisect_ppx` and `lwt_ppx`, and hoping that the CI will catch any errors.

### DIFF
--- a/packages/alba/alba.0.4.1/opam
+++ b/packages/alba/alba.0.4.1/opam
@@ -20,8 +20,8 @@ depends: [
   "odoc" {with-doc}
   "js_of_ocaml" {build}
   "js_of_ocaml-ppx"
-  "ppx_inline_test" {build} (* is needed for building the libraries
-                               otherwise it does not compile *)
+  "ppx_inline_test" (* is needed for building the libraries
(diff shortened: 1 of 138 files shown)


kit-ty-kate:
It seems a bit too excessive to remove it for all ppx packages. It makes sense for ppx_sexp_lib however.

avsm:
Why is it excessive? It’s strictly correct to do so — we simply don’t know what has changed in the library that may affect the code generation phase before.

avsm:
I've started a discussion among the opam-repo maintainer team about this, and will update when we reach consensus.

mseri:
I am going to close this for now, we will reopen and update it as soon as there is a consensus. Currently, it has got stale and would need rebasing anyways

PR #16280, opened 2020-04-23
Add css-parser 0.2.2

astrada:


kit-ty-kate:
According to your `dune-project` file this package requires at least dune 2.4. Could you return this fix upstream?

kit-ty-kate:
Thanks

astrada:
> Could you return this fix upstream?

Sure, thanks!

PR #16283, opened 2020-04-24
+ extlib 1.7.7

ygrek:
* sync with OCaml 4.10
* mark ExtList.find_map deprecated in anticipation of type breakage in next release to match OCaml 4.10 function with same name
* String: fold_left and fold_right without allocations


kit-ty-kate:
Thanks!

PR #16295, opened 2020-04-26
[new release] js_of_ocaml, js_of_ocaml-ocamlbuild, js_of_ocaml-ppx_deriving_json, js_of_ocaml-tyxml, js_of_ocaml-ppx, js_of_ocaml-lwt, js_of_ocaml-compiler and js_of_ocaml-toplevel (3.6.0)

hhugo:
Compiler from OCaml bytecode to Javascript

- Project page: <a href="http://ocsigen.github.io/js_of_ocaml">http://ocsigen.github.io/js_of_ocaml</a>

PR #16298, opened 2020-04-26
Fix vg constraint regarding jsoo

hhugo:
PR to fix this upstream https://github.com/dbuenzli/vg/pull/25
cc @dbuenzli 

required for #16295

### DIFF
--- a/packages/vg/vg.0.9.3/opam
+++ b/packages/vg/vg.0.9.3/opam
@@ -16,7 +16,7 @@ depends: [
   "ocamlbuild" {build}
   "topkg" {build}
   "gg" {>= "0.9.0"}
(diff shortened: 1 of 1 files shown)


avsm:
Thanks!

joelburget:
What was the problem here? I'm on jsoo 3.6.0 and am trying to use vg.

hhugo:
See the change in https://github.com/dbuenzli/vg/pull/25

joelburget:
Thanks @hhugo. Does that mean vg just needs to cut a new release including your patch?

hhugo:
yes. I believe so. cc @dbuenzli

dbuenzli:
Will do. Thanks @hhugo for you patch.

PR #16299, opened 2020-04-27
github.com/diml -> github.com/jeremiedimino

jeremiedimino:
I changed my github username from "diml" to "jeremiedimino". According to github the old URLs should still work, but we might as well use the new ones.

This PR replaces all references to "diml" by "jeremiedimino" in the repo.

### DIFF
--- a/packages/csexp/csexp.1.0.0/opam
+++ b/packages/csexp/csexp.1.0.0/opam
@@ -24,9 +24,9 @@ authors: [
   "Jeremie Dimino <jeremie@dimino.org>"
 ]
 license: "MIT"
-homepage: "https://github.com/diml/csexp"
-doc: "https://diml.github.io/csexp/"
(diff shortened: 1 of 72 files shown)


avsm:
Thanks @jeremiedimino!

One limitation of this username migration: if you had previously transferred something to another org, then the original `diml -> <org>` redirection no longer works.

So some packages like `pp` are broken, since they use:

```
https://github.com/jeremiedimino/pp/releases/download/1.0.1/pp-1.0.1.tbz
```

but only

```
https://github.com/ocaml-dune/pp/releases/download/1.0.1/pp-1.0.1.tbz
```

works. If you could push changes for those (the CI jobs are failing), then this is good to merge.

jeremiedimino:
Sure, I'm going to update these!

jeremiedimino:
Alright, I think I got them all

kit-ty-kate:
Looks good. Thanks a lot!

PR #16288, opened 2020-04-25
+angstrom.0.14.0

seliopou:

PR #16307, opened 2020-04-28
Fix angstrom revdeps

kit-ty-kate:


kit-ty-kate:
Cherry-picking commits from @seliopou from https://github.com/ocaml/opam-repository/pull/16288 to avoid bloating the CI

PR #16310, opened 2020-04-28
[new release] jose (0.3.0)

ulrikstrid:
JOSE implementation for OCaml and ReasonML

- Project page: <a href="https://ulrikstrid.github.io/reason-jose">https://ulrikstrid.github.io/reason-jose</a>
- Documentation: <a href="https://ulrikstrid.github.io/reason-jose">https://ulrikstrid.github.io/reason-jose</a>

##### CHANGES:

- Change the JWT representation to be based on a GADT, this allows us to use a private JWT for anything where a public JWT is enough.
- Add JWE encryption and decryption


kit-ty-kate:
I believe what you wanted is strict less than 0.7 for mirage-crypto. Allowing just for 0.7.0 would make no sense as it is not available yet and minor versions might fix some things.

Could you return the fix upstream? Thanks!

ulrikstrid:
Thank you @kit-ty-kate, will do!

PR #16311, opened 2020-04-29
+dispatch.0.5.0

seliopou:


kit-ty-kate:
Dune should not be tagged as a build dependency (see #14266 and https://github.com/ocaml/dune/issues/2147). Could you return the fix upstream?

Thanks!

PR #16312, opened 2020-04-29
Package hack_parallel.1.0.0

rvantonder:
:wave: 

I'm using the same constraints as added in the [previous release](https://github.com/ocaml/opam-repository/blob/master/packages/hack_parallel/hack_parallel.0.1.1/opam) and dropping `core` for `core_kernel` so this should be fine 🤞

### DIFF
--- a/packages/hack_parallel/hack_parallel.1.0.0/opam
+++ b/packages/hack_parallel/hack_parallel.1.0.0/opam
@@ -9,10 +9,11 @@ build: [
   ["dune" "build" "-p" name "-j" jobs "@install"]
 ]
 depends: [
+  "ocaml" {>= "4.04.1"}
+  "dune" {>= "1.11"}
   "conf-pkg-config"
   "conf-sqlite3"
   "core_kernel" {< "v0.14"}
-  "ocaml" {>= "4.04.1"}
   "ppx_deriving"
   "ppxlib" {< "0.9.0"}
   "sexplib" {< "v0.14"}
--- a/packages/hack_parallel/hack_parallel.1.0.0/opam
+++ b/packages/hack_parallel/hack_parallel.1.0.0/opam
@@ -9,16 +9,13 @@ build: [
   ["dune" "build" "-p" name "-j" jobs "@install"]
 ]
 depends: [
+  "conf-pkg-config"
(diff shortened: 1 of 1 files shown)


rvantonder:
> /home/opam/.opam/4.09/bin/dune "build" "-p" "hack_parallel" "-j" "72" "@install" (CWD=/home/opam/.opam/4.09/.opam-switch/build/hack_parallel.1.0.0)
[ERROR] The compilation of hack_parallel failed at "/home/opam/.opam/4.09/bin/dune build -p hack_parallel -j 72 @install".

> Error: I don't know about package hack_parallel (passed through -p)

Based on some [other issues](https://github.com/vbrankov/hdf5-ocaml/issues/22#issuecomment-507292302), it looks like opam expects the package name to be the same as the GH project name since some time in 2019? In this case the GitHub project name is hyphenated (no underscore). I'll try rename the GH project rather than moving the opam file here first.

rvantonder:
Fixed -- think the real issue was that I forgot to uncomment the `url` part :)

packages/hack_parallel/hack_parallel.1.0.0/opam
```diff
@@ -0,0 +1,31 @@
...
+depends: [
+  "conf-sqlite3"
+  "core_kernel" {< "v0.14"}
+  "ocaml" {>= "4.04.1"}
+  "ppx_deriving"
+  "ppxlib" {< "0.9.0"}
+  "sexplib" {< "v0.14"}
+]
+depexts: [
+  "pkg-config"
```
avsm, line 20: These depext's aren't quite the right format: you can just depend on `conf-pkg-config` above in the `depends` field  and drop the `depexts` clause entirely. See the `conf-sqlite3` and `conf-pkg-config` opam files for the opam 2.0 depext format if curious.

rvantonder: Handy, thanks! Updated.

kit-ty-kate:
There was a missing dependency on dune. Could you return the fix upstream?

Thanks!

rvantonder:
@kit-ty-kate thanks for catching!

PR #16315, opened 2020-04-29
+httpaf.0.6.6

seliopou:


seliopou:
It's unfortunate but for the CI to go green an upper bound needs to be added to angstrom so that the tests pass. Not sure if that's what should live in OPAM or not.

kit-ty-kate:
This should fix the issue. Does that sound ok for you?

seliopou:
I didn’t know you could do that. Looks good!

kit-ty-kate:
Thanks!

PR #16316, opened 2020-04-29
loosen-coin-deps: no upper bound on angstrom necessary

seliopou:
It looks like the upper bound on angstrom wasn't necessary after all.


kit-ty-kate:
Thanks!

PR #16318, opened 2020-04-30
new file:   packages/molenc/molenc.8.0.0/opam

UnixJunkie:
Added an atom pairs encoder.
Trigger with:
```
molenc.sh --pairs ...
```
Might be a less lossy molecular encoding than circular fingerprints with a fixed (maximum) radius.


kit-ty-kate:
Thanks!

PR #16327, opened 2020-05-01
new file:   packages/linwrap/linwrap.5.0.0/opam

UnixJunkie:


kit-ty-kate:
I added a missing constraint on `minicli`. Could you return this upstream? Thanks!

UnixJunkie:
ok, done.

PR #13284, opened 2019-01-17
conf-openssl: macos: look for openssl until it is found

cmr:
Right now, on a machine with both nix and homebrew configured, the conf-openssl osx script will only look at nix for the openssl pkg-config files. This changes it to look in nix, and if it isn't found, also look in homebrew, and if it isn't found there, finally look in macports.

PR #10503, opened 2017-10-19
Add support for macport openssl package.

toots:
This is a complement to #10453. The other provider for `openssl` on OSX is MacPorts and, of course, they use a different location. Please refer to #10453 for technical details.. Thanks for your consideration!

PR #16337, opened 2020-05-02
Add conf-libssl.2: Early failure and helpful error messages

kit-ty-kate:
Here is the difference between the two versions:
```
--- conf-libssl.1/opam	2020-04-03 15:22:16.796495179 +0100
+++ conf-libssl.2/opam	2020-05-02 13:41:51.890023963 +0100
@@ -4,12 +4,10 @@
 homepage: "https://www.openssl.org/"
 license: "Apache-1.0"
-build: [
-  ["pkg-config" "openssl"]
-    {os != "macos" & os != "freebsd" & os != "openbsd" & os != "netbsd"}
-  ["sh" "./osx-build.sh"] {os = "macos"}
+build: ["pkg-config" "openssl"]
+depends: [
+  "conf-pkg-config" {build}
 ]
-depends: ["conf-pkg-config" {build}]
 depexts: [
   ["libssl-dev"] {os-family = "debian"}
   ["openssl-devel"] {os-distribution = "centos"}
@@ -21,9 +19,16 @@
   ["openssl"] {os-distribution = "nixos"}
   ["openssl"] {os-distribution = "arch"}
   ["libopenssl-devel"] {os-family = "suse"}
+  ["openssl"] {os = "freebsd"}
+]
+post-messages: [
+  "Make sure libssl/openssl is installed and accessible using pkg-config." {failure}
+  "Set the PKG_CONFIG_PATH environment variable if necessary." {failure}
+  "For example: export PKG_CONFIG_PATH=$(brew --prefix openssl)/lib/pkgconfig" {failure & os-distribution = "homebrew"}
+  "For example: export PKG_CONFIG_PATH=$HOME/.nix-profile/lib/pkgconfig" {failure & os-distribution = "nixos"}
+  "For example: export PKG_CONFIG_PATH=/opt/local/lib/pkgconfig" {failure & os-distribution = "macports"}
 ]
 synopsis: "Virtual package relying on an OpenSSL library system installation"
 description:
   "This package can only install if the OpenSSL library is installed on the system."
-extra-files: ["osx-build.sh" "md5=faf81a6bf30e346abe94610f136c6193"]
 flags: conf
```
With this change, the package should not succeed and let users try to install `ssl` for example without having a reachable libssl library (see https://github.com/ocaml/opam-repository/issues/15295#issuecomment-606789054). cc @reykjalin does that help?

The `osx-build.sh` script didn't seem to make much sense, I'm not too sure why it was added in the first place. I didn't understand it myself and seemed to have made things worst. It seems to me that instead of looking for the right library, we should display helpful error messages instead of ignoring a potential future error, as if people don't set their `PKG_CONFIG_PATH` environment variable, nothing will tell them something is missing until they run into issues later like:
```
discover.c:2:10: fatal error: 'openssl/ssl.h' file not found
#include <openssl/ssl.h>
```
cc @toots it seems that the original special case for macOS was added in https://github.com/ocaml/opam-repository/pull/10453 and https://github.com/ocaml/opam-repository/pull/10503. Is this distinction still relevant? How can we help users and not run into issues later if it is the case?

The Nix special case was added in https://github.com/ocaml/opam-repository/pull/13135 and fixed in https://github.com/ocaml/opam-repository/pull/13284. cc @anmonteiro does my change make sense?

cc @hannesm @rneswold I was looking at old commits (https://github.com/ocaml/opam-repository/commit/b96dc1f095593c9089ee3edf52cda44658fbd169) to understand why there was restrictions on BSDs. Should those restrictions still exists? I was able to get openssl installed and working with pkg-config first try with FreeBSD 12.1. (let's see if the experimental freebsd travis check confirms it)

This PR is just a draft for the moment, a place for discussion.

### DIFF
--- a/packages/conf-libssl/conf-libssl.2/opam
+++ b/packages/conf-libssl/conf-libssl.2/opam
@@ -4,7 +4,10 @@ authors: ["The OpenSSL Project"]
 bug-reports: "https://github.com/ocaml/opam-repository/issues"
 homepage: "https://www.openssl.org/"
 license: "Apache-1.0"
-build: ["pkg-config" "openssl"]
+build: [
+  ["pkg-config" "openssl"]
+    {os != "freebsd" & os != "openbsd" & os != "netbsd"} # libssl is provided by base system on BSDs
+]
 depends: [
   "conf-pkg-config" {build}
 ]
@@ -19,7 +22,6 @@ depexts: [
   ["openssl"] {os-distribution = "nixos"}
   ["openssl"] {os-distribution = "arch"}
   ["libopenssl-devel"] {os-family = "suse"}
-  ["openssl"] {os = "freebsd"}
 ]
 post-messages: [
   "Make sure libssl/openssl is installed and accessible using pkg-config." {failure}
--- a/packages/conf-libssl/conf-libssl.2/opam
+++ b/packages/conf-libssl/conf-libssl.2/opam
@@ -1,0 +1,34 @@
+opam-version: "2.0"
+maintainer: "David Sheets <sheets@alum.mit.edu>"
+authors: ["The OpenSSL Project"]
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+homepage: "https://www.openssl.org/"
+license: "Apache-1.0"
+build: ["pkg-config" "openssl"]
+depends: [
+  "conf-pkg-config" {build}
+]
+depexts: [
+  ["libssl-dev"] {os-family = "debian"}
+  ["openssl-devel"] {os-distribution = "centos"}
+  ["openssl-devel"] {os-distribution = "ol"}
+  ["openssl-devel"] {os-distribution = "fedora"}
+  ["openssl"] {os = "macos" & os-distribution = "homebrew"}
+  ["openssl"] {os = "macos" & os-distribution = "macports"}
+  ["libressl-dev"] {os-distribution = "alpine"}
+  ["openssl"] {os-distribution = "nixos"}
+  ["openssl"] {os-distribution = "arch"}
+  ["libopenssl-devel"] {os-family = "suse"}
+  ["openssl"] {os = "freebsd"}
+]
+post-messages: [
+  "Make sure libssl/openssl is installed and accessible using pkg-config." {failure}
+  "Set the PKG_CONFIG_PATH environment variable if necessary." {failure}
+  "For example: export PKG_CONFIG_PATH=$(brew --prefix openssl)/lib/pkgconfig" {failure & os-distribution = "homebrew"}
+  "For example: export PKG_CONFIG_PATH=$HOME/.nix-profile/lib/pkgconfig" {failure & os-distribution = "nixos"}
+  "For example: export PKG_CONFIG_PATH=/opt/local/lib/pkgconfig" {failure & os-distribution = "macports"}
+]
+synopsis: "Virtual package relying on an OpenSSL library system installation"
+description:
+  "This package can only install if the OpenSSL library is installed on the system."
+flags: conf


hannesm:
@kit-ty-kate I'm not sure I understand what is the issue to be fixed here. On FreeBSD, libssl is provided by the base system, as is a C compiler, a C library, etc. there's usually no need to install "openssl" from ports. which opam package fails to install on FreeBSD with `conf-libssl.1`?

toots:
@kit-ty-kate As far as I can tell, OSX needs a special treatment because the library can be installed by multiple sources and some of these installs are purposely "build only", i.e. invisible for the global `pkg-config` view. Typically, from homebrew:
```
openssl@1.1 is keg-only, which means it was not symlinked into /usr/local,
because openssl/libressl is provided by macOS so don't link an incompatible version.

If you need to have openssl@1.1 first in your PATH run:
  echo 'export PATH="/usr/local/opt/openssl@1.1/bin:$PATH"' >> ~/.zshrc

For compilers to find openssl@1.1 you may need to set:
  export LDFLAGS="-L/usr/local/opt/openssl@1.1/lib"
  export CPPFLAGS="-I/usr/local/opt/openssl@1.1/include"

For pkg-config to find openssl@1.1 you may need to set:
  export PKG_CONFIG_PATH="/usr/local/opt/openssl@1.1/lib/pkgconfig"
```

The script seems to be trying each different possible location for the library on macOS. I believe that one assumption about it is that we never want to link to the system-provided library _and_ that the system-provided library does not support `pkg-config`.

mseri:
The post message on OSX seems to cover most (if not all) of the issues we saw raised for osx and libssl. I think that part will reduce a lot future complaints

kit-ty-kate:
> The script seems to be trying each different possible location for the library on macOS. I believe that one assumption about it is that we never want to link to the system-provided library _and_ that the system-provided library does not support `pkg-config`.

@toots It seems to be that it is exactly why this PR makes sense.

Homebrew for instance tells you explicitly to add the path to your `PKG_CONFIG_PATH` but if you don't the script in `conf-libssl.1` does not say anything and lets you fail later.
Is there a reasonably used source for libssl that does not provide any pkg-config? Homebrew does, macports maybe? If so maybe we should just filter on the package system used.

toots:
> > The script seems to be trying each different possible location for the library on macOS. I believe that one assumption about it is that we never want to link to the system-provided library _and_ that the system-provided library does not support `pkg-config`.
> 
> @toots It seems to be that it is exactly why this PR makes sense.
> 
> Homebrew for instance tells you explicitly to add the path to your `PKG_CONFIG_PATH` but if you don't the script in `conf-libssl.1` does not say anything and lets you fail later.
> Is there a reasonably used source for libssl that does not provide any pkg-config? Homebrew does, macports maybe? If so maybe we should just filter on the package system used.

Yes, you're right. It would be nice if `opam` could handle the simple situations w/o having to require external input from the user. Perhaps the script could test for the presence of the `brew` command only then use brew's `PKG_CONFIG` path, and same with the other ones?

kit-ty-kate:
> Yes, you're right. It would be nice if `opam` could handle the simple situations w/o having to require external input from the user. Perhaps the script could test for the presence of the `brew` command only then use brew's `PKG_CONFIG` path, and same with the other ones?

I'm not sure to understand, isn't it what the script from `conf-libssl.1` was doing? An user input is required in any case I'm afraid.

reykjalin:
> With this change, the package should not succeed and let users try to install ssl for example without having a reachable libssl library (see https://github.com/ocaml/opam-repository/issues/15295#issuecomment-606789054)

That helps a lot. I think the main part of my issue was that I was using [esy](https://esy.sh), which uses `opam` in the background, but removes any custom environment setups (like modifications to `PKG_CONFIG_PATH`).
As a result when I applied a `conf-libssl` specific fix that just transferred the problem to a different package (see https://github.com/ocaml/opam-repository/issues/15295#issuecomment-606789054).

Because of this I think this really should be a warning displayed for the user so they can fix it in their environment. My specific case requires that I overwrite the system provided openssl library with a symlink to `openssl@1.1` from Homebrew.
That's not an ideal case, and it should be up to the user to make that decision, which is why I think @kit-ty-kate's fix is the correct approach.

mseri:
What is preventing this from being merged? Tbh I think it's an improvement over the current state of things

kit-ty-kate:
Sorry this took so long. Thanks a lot for your inputs everyone.

PR #16342, opened 2020-05-04
Add wasm-1.1.1

rossberg:
Includes new multi values, bit extension, and saturating conversions features.

### DIFF
--- a/packages/wasm/wasm.1.1.1/opam
+++ b/packages/wasm/wasm.1.1.1/opam
@@ -1,0 +1,22 @@
+opam-version: "2.0"
+maintainer: "Andreas Rossberg <rossberg@mpi-sws.org>"
+authors: "Andreas Rossberg <rossberg@mpi-sws.org>"
+homepage: "https://github.com/WebAssembly/spec"
+bug-reports: "https://github.com/WebAssembly/spec/issues"
+license: "Apache-2.0"
+dev-repo: "git+https://github.com/WebAssembly/spec.git"
+build: [
+  [make "-C" "interpreter" "opt" "unopt"]
+]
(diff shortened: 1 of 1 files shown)


rossberg:
Btw, is it possible to update existing opam packages in a purely administrative manner? We'd like to clean up our git labels in the Wasm repo, which means that the zip file names and MD5's would have to change in this repo (though not their src content). Is that considered okay?

kit-ty-kate:
Thanks!

> Btw, is it possible to update existing opam packages in a purely administrative manner? We'd like to clean up our git labels in the Wasm repo, which means that the zip file names and MD5's would have to change in this repo (though not their src content). Is that considered okay?

It is possible. If the checksums are the same this should be an easy PR to merge. However if the checksums are not the same we have to check that the changes are not major.

In any case the tags used by opam today should only be removed **after** (ideally give it a few days) the PR is being merged.

rossberg:
@kit-ty-kate, the only thing that would change _inside_ the zip (and affects the checksum) is the name of the top-level directory everything is zipped into, because of the way GitHub creates these zip files. All the actual files inside that would remain identical. It sounds like that would be okay?

kit-ty-kate:
Unless zip files generation are different in this regard, the top-level directory should not change name and the checksum should thus not be affected. They are only based on the name of the repository/organisation as far as I know. But even if they were, as long as the original archive/tag are not removed before the PR is merged I don't see any issues with this.

rossberg:
Thanks, I'll create a PR then. (Except for `master`, the repo zip file generated by GH mangles the tag/branch name into the dir name, for example `xyz-abc` for tag `abc` in repo `xyz`.)

PR #16346, opened 2020-05-04
[new release] 0install-gtk, 0install and 0install-solver (2.17)

talex5:
Decentralised installation system - GTK UI

- Project page: <a href="https://github.com/0install/0install">https://github.com/0install/0install</a>
- Documentation: <a href="https://0install.github.io/0install/">https://0install.github.io/0install/</a>

##### CHANGES:

Solver:

- Add 0install-solver.opam. This makes the solver into a separate opam package.

- Improve display of implementations in diagnostics. Instead of `sha1=3ce644dc725f1d21cfcf02562c76f375944b266a (1)`, show: `v1 (sha1=3ce644dc725f...)`.

- Only report restrictions that affected the result. If a restriction didn't remove any candidates (either because it matched all of them or because another restriction already removed them) then don't bother reporting it. Also, don't bother reporting a restriction that only removed candidates worse than the selected one (if any). For user-provided restrictions, filter the rejects to show only the version the user asked for (unless that would remove all of them). Since we only show the first 5 rejects, this would often mean that the interesting candidates weren't even shown.

- Expose more of the solver diagnostics API. This allows users to format the results in other ways (for example, as a list in a GUI). Also, use formatting boxes instead of manual indentation.

- Add `conflict_class`. This isn't used by 0install, but the new opam backend that uses the 0install solver needs it.

Build:

- Update minimum OCaml version to 4.08. Remove our custom `Option` module and use the new stdlib one instead. Use the new `List.filter_map` from the stdlib instead of our one. Also, rename our `first_match` to `find_map`, to match the name in 4.10.

- Depend on ounit2 for unit-tests. The old `oUnit` is just a transition package that depends on `ounit2` now.

### DIFF
--- a/packages/0install-gtk/0install-gtk.2.17/opam
+++ b/packages/0install-gtk/0install-gtk.2.17/opam
@@ -1,0 +1,30 @@
+opam-version: "2.0"
+synopsis: "Decentralised installation system - GTK UI"
+maintainer: "talex5@gmail.com"
+authors: "zero-install-devel@lists.sourceforge.net"
+homepage: "https://github.com/0install/0install"
+bug-reports: "https://github.com/0install/0install/issues"
+dev-repo: "git+https://github.com/0install/0install.git"
+doc: "https://0install.github.io/0install/"
+build: [
+  ["dune" "build" "-p" name "-j" jobs "@install" "@runtest" {with-test}]
+]
+depends: [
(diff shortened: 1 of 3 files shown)


talex5:
(Travis failed on mac as usual, due to CI sandboxing blocking the unit-tests)

kit-ty-kate:
Thanks!

PR #16347, opened 2020-05-04
Fix depext on pkgconfig in alpine

AltGr:


kit-ty-kate:
oh this is surprising. The issue was that pkg-config is installed by default in our CI images for Alpine (but not on other distributions I think). So we never ran into this issue in CI.

kit-ty-kate:
Thanks!

PR #16350, opened 2020-05-05
rename "ocaml-flac"

fdopen:
Previous versions had proper names; probably just a misconfigured publish script...
CC @toots


toots:
Yes, my bad.

kit-ty-kate:
Indeed, I should've realized in https://github.com/ocaml/opam-repository/pull/16338. Thanks!

PR #16351, opened 2020-05-05
Redirect Wasm packages to new tag names

rossberg:
In order to complete a clean-up of the tagging scheme in the Wasm spec repo, this PR changes the existing Wasm packages to point to the new tag names.

Note: Since GH's zip file generator mangles the tag name into the toplevel dir name inside the zip, this also changes the MD5's, even though the actual source contents aren't changed.


kit-ty-kate:
For our record I double-checked and none of the archive changed. Thanks a lot!

rossberg:
Thanks @kit-ty-kate!

PR #16357, opened 2020-05-05
add ocaml.4.10.0+nnpcheck to dynamically check for naked pointers

avsm:
see https://github.com/ocaml/ocaml/pull/9534 for more discussion

### DIFF
--- a/packages/ocaml-variants/ocaml-variants.4.10.0+nnpcheck/opam
+++ b/packages/ocaml-variants/ocaml-variants.4.10.0+nnpcheck/opam
@@ -1,0 +1,41 @@
+opam-version: "2.0"
+synopsis: "Official release 4.10.0, with a dynamic check for naked pointers"
+maintainer: "platform@lists.ocaml.org"
+authors: "Xavier Leroy and many contributors"
+homepage: "https://ocaml.org"
+bug-reports: "https://github.com/ocaml/ocaml/issues"
+dev-repo: "git://github.com/ocaml/ocaml"
+depends: [
+  "ocaml" {= "4.10.0" & post}
+  "base-unix" {post}
+  "base-bigarray" {post}
+  "base-threads" {post}
+]
+conflict-class: "ocaml-core-compiler"
+flags: compiler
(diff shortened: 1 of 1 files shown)


kayceesrk:
Great. Thanks!

packages/ocaml-variants/ocaml-variants.4.10.0+nnpcheck/opam
```diff
@@ -0,0 +1,41 @@
...
+    "--prefix=%{prefix}%"
+    "CC=cc"
+    "ASPP=cc -c"
+  ] {os = "openbsd" | os = "freebsd" | os = "macos"}
+  [make "-j%{jobs}%" {os != "cygwin"} "world"]
+  [make "-j%{jobs}%" {os != "cygwin"} "world.opt"]
+]
+install: [make "install"]
+url {
+  src: "git+https://github.com/kayceesrk/ocaml.git#4.10.0+nnp+check"
```
kit-ty-kate, line 31:
```suggestion
  src: "https://github.com/kayceesrk/ocaml/archive/4.10.0+nnp+check.tar.gz"
```
We can avoid git here

kayceesrk:
There will be a separate branch for `4.10.1` if there will be one.

kit-ty-kate:
Sweet! Then once the git requirement is removed this should be good to go.

avsm:
The reason I made this git is that it's much more efficient as a pinned repository.  This particular variant is intended for a very specific development use, so I think that's fine. (e.g. if @kayceesrk pushes more changes to this branch as the OCaml PR develops).  There's not too much utility in making this a http archive without a corresponding checksum, which won't work in this case.

(I realise that this is a departure from the normal opam-repository policy, but we've done this in the past for other cases where it's useful to the community, such as with early ppx_tools).

kit-ty-kate:
Ok then. Thanks!

kayceesrk:
Thanks @kit-ty-kate!

avsm:
Thanks @kit-ty-kate!

PR #16360, opened 2020-05-05
First alpha for OCaml 4.11.0

Octachron:
For this release cycle, we are testing publishing alpha versions of OCaml 4.11.0 starting from the feature freeze.

The idea is that the alpha version would be updated regularly, and then morph into a beta once the opam ecosystem is in shape for the new version.

I am using the same set of variant as usual vanilla, flambda, fp, flambda+fp, afl.


kit-ty-kate:
Thanks a lot!

PR #16361, opened 2020-05-05
Package batch_jaro_winkler.0.1.0

dbousque:
### `batch_jaro_winkler.0.1.0`
Fast batch jaro winkler distance implementation in C99
This project gets its performance from the pre-calculation of an optimized model in advance of the actual runtime calculations. Supports any encoding.

---
* Homepage: https://github.com/dbousque/batch_jaro_winkler

---

### DIFF
--- a/packages/batch_jaro_winkler/batch_jaro_winkler.0.1.0/opam
+++ b/packages/batch_jaro_winkler/batch_jaro_winkler.0.1.0/opam
@@ -18,7 +18,7 @@ depends: [ "ocaml" {>= "4.0"} "dune" {>= "1.11.4"} ]
 url {
   src: "https://github.com/dbousque/batch_jaro_winkler/archive/0.1.0.tar.gz"
   checksum: [
-    "md5=3efee6018441a468d3d95ddad5954fbf"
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Your dune setup does not seem quite right. The `dune-project` file should almost always be in the top-level directory of your project, dune should always be called with `-p <name of your project>` (otherwise it adds flags to the compiler that are meant for development, not for releases) and `dune install` should almost never be used in opam.

I'll try to hot-fix your opam file for the moment.

kit-ty-kate:
CI is green. Do my changes seem good to you? If so this PR should be good to go. Could you also return them to the opam file upstream so that the next release does not have to go through the same changes?

dbousque:
@kit-ty-kate Thanks, looks good. Yes I'll update the upstream with the changes. How much of a problem could it be that `dune-project` is not in the top-level directory? As this is a multi-language project, I wouldn't like putting OCaml files outside of the `ocaml` folder.

kit-ty-kate:
I understand, it's not mandatory (it's fine with my changes). However know that you might encounter an issue later if someone tries to pin your repository using opam. I've talked with the opam devs and created an issue describing the problem in https://github.com/ocaml/opam/issues/4185

For now I think this is fine. Thanks a lot!

dbousque:
Thanks @kit-ty-kate!

PR #16362, opened 2020-05-06
Package: lambda-term.3.0.1

kandu:
3.0.1 (2020-05-06)
------------------

* `LTerm_read_line`: fix synchronization bug


kit-ty-kate:
Thanks!

PR #16368, opened 2020-05-06
ocaml-freestanding 0.6.0

mato:
## v0.6.0 (2020-05-06)

* Drop support for OCaml 4.06 and 4.07. 
* Define an `__ocaml_freestanding__` preprocessor macro, undefine leaky host toolchain preprocessor macros (`__linux__`, `__FreeBSD__`, `__OpenBSD__` et al.), provide an `<endian.h>`.
* Install all OCaml runtime headers.
* Provide a minimal `<inttypes.h>`.
* Various build system improvements and cleanups.

Note that mirage/ocaml-freestanding#74 may break downstream C code that uses `#if...#elif...#endif` chains to detect the system it is built on. Such code should be explicitly adapted to detect the presence of ocaml-freestanding by testing for the `__ocaml_freestanding__` preprocessor macro.

Further, downstream OPAM packages with C code that wish to use the interfaces/headers added in this release should depend on `ocaml-freestanding { >= 0.6.0 }`.

### DIFF
--- a/packages/ocaml-freestanding/ocaml-freestanding.0.6.0/opam
+++ b/packages/ocaml-freestanding/ocaml-freestanding.0.6.0/opam
@@ -1,0 +1,40 @@
+opam-version: "2.0"
+maintainer: "Martin Lucina <martin@lucina.net>"
+authors: "Martin Lucina <martin@lucina.net>"
+homepage: "https://github.com/mirage/ocaml-freestanding"
+bug-reports: "https://github.com/mirage/ocaml-freestanding/issues/"
+license: "MIT"
+tags: "org:mirage"
+dev-repo: "git+https://github.com/mirage/ocaml-freestanding.git"
+build: [make]
+install: [make "install" "PREFIX=%{prefix}%"]
+remove: [make "uninstall" "PREFIX=%{prefix}%"]
+depends: [
+  "conf-pkg-config"
+  "ocamlfind" {build}
+  "ocaml-src" {build}
+  ("solo5-bindings-hvt" | "solo5-bindings-spt" | "solo5-bindings-virtio" | "solo5-bindings-muen" | "solo5-bindings-genode")
+  "ocaml" {>= "4.08.0" & < "4.11.0"}
+]
+substs: [
+  "flags/cflags.tmp"
+  "flags/libs.tmp"
+]
+conflicts: [
+  "sexplib" {= "v0.9.0"}
+  "solo5-kernel-ukvm"
+  "solo5-kernel-virtio"
+  "solo5-kernel-muen"
+]
+available: [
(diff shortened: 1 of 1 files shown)


mato:
So, most (all? would need to go through them) failures here are due to OPAM picking `solo5-bindings-genode` out of ocaml-freestanding's possible dependencies on Solo5 variants:

    ("solo5-bindings-hvt" | "solo5-bindings-spt" | "solo5-bindings-virtio" | "solo5-bindings-muen" | "solo5-bindings-genode")

I'm not sure if it's deterministic and is just picking `-genode` as first when sorted alphabetically. This is unfortunate, since the Solo5 Genode bindings are broken on _some_ distros, and not really fixable (see https://github.com/Solo5/solo5/pull/446).

Some ideas on what to do about it:

1. Ignore such failures every time we deal with a new upload of `ocaml-freestanding` or `solo5-bindings-*` to OPAM. This seems dangerous, as we may be papering over/missing other problems.
2. Somehow force the CI / OPAM to always pick a particular Solo5 variant from the list. I'm not sure if this is possible.
3. Declare the Genode variant unmaintainable, removing it from OPAM in this & future releases of ocaml-freestanding and Solo5.

Thoughts? /cc @hannesm @ehmry

hannesm:
I'm not sure what a sensible solution is. I suspect not publishing `solo5-bindings-genode` to opam-repository (and marking the existing ones as `available: false`) would be a good way forward. Then we'd need to document in the solo5 repository (and/or MirageOS website) "to use genode, please manually `opam pin add solo5-bindings-genode https://github.com/solo5/solo5.git#latest_release_tag`".

For the time being, I'd be in favour to get this PR merged (the package does not fail, I've packages which depend on ocaml-freestanding 0.6.0 that I'd like to release.

mato:
> I suspect not publishing solo5-bindings-genode to opam-repository (and marking the existing ones as available: false) would be a good way forward. [...]

That seems like a good way forward.

> For the time being, I'd be in favour to get this PR merged (the package does not fail, I've packages which depend on ocaml-freestanding 0.6.0 that I'd like to release.

Yes, please. The package has been extensively tested both manually and by other CIs. So, if the opam-repository maintainers are not against, then let's merge it.

ehmry:
I'm fine with deprecating the genode bindings for the time being. If someone else wants to assume maintainership then I'm happy to help.

avsm:
I'm fine with merging this, as it's in line with previous PRs for earlier versions. However, we would appreciate a followup PR deprecating the earlier genode bindings, so that bulk builds of opam-repo will select buildable targets.  Will merge this once CI gets through its backlog.

kit-ty-kate:
Good for me. Thanks!

PR #16372, opened 2020-05-06
[new release] mssql (2.0.1)

brendanlong:

### Added

- Streaming `execute_` helpers: `execute_map`, `execute_iter`, `execute_fold`, and `execute_pipe`.
- `Param.Array` now supports lists, which is useful for `IN ($1)` clauses.

### Changed

- Make `connect`'s `port` argument optional
- Support Core v0.13
- Result sets that don't contain row data aren't returned. For example, `INSERT ...; SELECT ...` now returns one
  result set instead of two.

### Fixed

- Correctly use `port` when provided
- Various [upstream fixes in `ocaml-freetds`](https://github.com/kennknowles/ocaml-freetds/releases/tag/0.7)
  - Exceptions shouldn't break the connection handle
  - Runtime lock released during queries
- Logging always occurs in an Async context
- Logging occurs in the same Async context as the caller and not a random one
- We depend on Async_extra

### Removed

- `Mssql.Test`. This module was for testing and shouldn't have been part of the public API. We recommend adding a
  module like this to your own code if you want it.
- Semi-broken connection pool (`Mssql.Pool`) removed. Doing this safely requires setting the
  [`RESETCONNECTION` bit](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/ce398f9a-7d47-4ede-8f36-9dd6fc21ca43),
  which doesn't seem to be possible in FreeTDS.

### DIFF
--- a/packages/mssql/mssql.2.0.1/opam
+++ b/packages/mssql/mssql.2.0.1/opam
@@ -1,0 +1,46 @@
+opam-version: "2.0"
+synopsis: "Async SQL Server client using FreeTDS"
+description:
+  "Mssql wraps FreeTDS in a nicer and safer interface, with support for parameterized queries, thread-based async IO, and a thread pool."
+maintainer: ["Arena Developers <silver-snakes@arena.io>"]
+authors: ["Arena Developers <silver-snakes@arena.io>"]
+license: "Apache-2.0"
+homepage: "https://github.com/arenadotio/ocaml-mssql"
+bug-reports: "https://github.com/arenadotio/ocaml-mssql/issues"
+depends: [
+  "alcotest" {with-test & >= "1.0.1"}
+  "alcotest-async" {with-test & >= "1.0.1"}
+  "async_unix"
+  "bignum"
+  "ppx_jane"
+  "iter" {>= "1.2"}
+  "ocaml" {>= "4.06.1"}
+  "logs"
+  "text" {>= "0.8.0"}
+  "freetds" {>= "0.7"}
+  "bisect_ppx" {dev & >= "2.0.0"}
+  "dune" {>= "1.11"}
+]
+build: [
+  ["dune" "subst"] {pinned}
+  [
+    "dune"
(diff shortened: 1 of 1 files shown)


brendanlong:
I'm not sure what we should do about that caml4p build error. We opened https://github.com/vbmithr/ocaml-text/issues/4 since we don't use caml4p, but it seems like this should work anyway, since caml4p claims to support 4.09 here: https://opam.ocaml.org/packages/camlp4/

The `+system` and `+1` versions of caml4p are suspicious though. I wonder if opam should be installing the `+1` version if `+system` doesn't work?

brendanlong:
I'm also not sure what to do about the test failures. The tests require a SQL Server instance so most people shouldn't run them.

packages/mssql/mssql.2.0.1/opam
```diff
@@ -0,0 +1,46 @@
...
+  ["dune" "subst"] {pinned}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
```
mseri, line 34: For the tests we can probably disable them
```suggestion
```

packages/mssql/mssql.2.0.1/opam
```diff
@@ -0,0 +1,46 @@
...
+    "@doc" {with-doc}
```
mseri, line 35: If you have this line you also need to add `odoc {with-doc}` as dependency

brendanlong:
See #16378

If there's a better way to update PR's I can do that but I couldn't convince GitHub to let me replace a release.

kit-ty-kate:
> If there's a better way to update PR's I can do that but I couldn't convince GitHub to let me replace a release.

It seems you're using `dune-release`, I don't use it but you can ask them here: https://github.com/ocamllabs/dune-release

Should this PR be closed?

PR #16376, opened 2020-05-07
[new release] ocaml-migrate-parsetree (1.7.3)

jeremiedimino:
Convert OCaml parsetrees between different versions

- Project page: <a href="https://github.com/ocaml-ppx/ocaml-migrate-parsetree">https://github.com/ocaml-ppx/ocaml-migrate-parsetree</a>
- Documentation: <a href="https://ocaml-ppx.github.io/ocaml-migrate-parsetree/">https://ocaml-ppx.github.io/ocaml-migrate-parsetree/</a>

##### CHANGES:

- Fix magic numbers for the 4.11 ast (ocaml-ppx/ocaml-migrate-parsetree#96, @hhugo)

### DIFF
--- a/packages/ocaml-migrate-parsetree/ocaml-migrate-parsetree.1.7.3/opam
+++ b/packages/ocaml-migrate-parsetree/ocaml-migrate-parsetree.1.7.3/opam
@@ -17,7 +17,7 @@ depends: [
   "result"
   "ppx_derivers"
   "dune" {>= "1.9.0"}
-  "ocaml" {>= "4.02.3"}
+  "ocaml" {>= "4.02.3" & < "4.12"}
 ]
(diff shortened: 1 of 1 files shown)


avsm:
thanks @jeremiedimino! Do earlier versions of omp need to be upper-bound restricted to `<4.11.0` as well to avoid them being selected?

jeremiedimino:
That seems right. In fact, now that you say it I wonder why omp doesn't systematically have a upper bound on the version of OCaml.

Well, at least `1.7.0` to `1.7.2` should be marked as incompatible. Should I do this in this PR?

kit-ty-kate:
I'll do a PR to restrict previous versions as well right now. Thanks a lot!

PR #16392, opened 2020-05-09
[new release] lp (0.1.0)

ktahar:
new release of [lp](https://github.com/ktahar/ocaml-lp).

- added interface to GLPK (new package named lp-glpk)
- various API enhancements / changes


ktahar:
Since building tests of this package requires [GLPK](https://www.gnu.org/software/glpk/),
I put dependency on [conf-glpk](https://github.com/ocaml/opam-repository/blob/master/packages/conf-glpk/conf-glpk.1/opam) in opam file.
It seems OK on most of the distros, but failing on Alpine.
Quick googling tells me that Alpine doesn't have a pre-compiled GLPK package.
What can I do?

kit-ty-kate:
Thanks!

PR #16394, opened 2020-05-09
minios-xen: drop ocaml depedency

hannesm:
it does not depend on ocaml, the dependency got introduced in the big https://github.com/ocaml/opam-repository/commit/1bab4537a2f3b2328d771e1bb50b0d0268b0798a "restart 2.0 with a clean repo" commit for unknown reasons.


kit-ty-kate:
Thanks!

PR #16397, opened 2020-05-10
[new release] stdcompat.14

thierry-martinez:
- Support for OCaml 4.11.0 with
  - `Array.{for_all2, exists2}`
  - `Lexing.{set_position, set_filename}`
  - `List.{filteri, fold_left_map}`
  - `Printexc.{default_uncaught_exception_handler, Slot.name}`
  - `Seq.{cons, append, unfold}`
  - `{Set,Map}.filter_map`
  - `Printf.{ibprintf, ikbprintf}`

- More efficient implementation of `Set` functions

- Support for version mismatch between `ocamlc` and `ocamlfind` packages


kit-ty-kate:
Thanks!

PR #16403, opened 2020-05-10
Package memprof-limits.0.1

gadmm:
### `memprof-limits.0.1`
Memory limits based on memprof
Memprof-limits provides per-thread global memory limits, and
per-thread allocation limits à la Haskell, compatible with systhreads.

Per-thread global memory limits let you bound the memory consumption
of specific parts of your program, in terms of memory used by the
whole program.

Per-thread allocation limits let you bound the execution of parts of
the program measured in number of allocation, analogous to the same
feature in Haskell. Allocation limits count allocations but not
deallocations, and is therefore a measure of the work done which can
be more suitable than execution time.

The implementation uses memprof with a low sampling rate that does not
affect performance. A reimplementation of the memprof interface
compatible with memprof-limits running at the same time is provided
for profiling needs.

THIS PACKAGE IS EXPERIMENTAL

---
* Homepage: https://gitlab.com/gadmm/memprof-limits
* Bug tracker: https://gitlab.com/gadmm/memprof-limits/issues

---

### DIFF
--- a/packages/memprof-limits/memprof-limits.0.1/opam
+++ b/packages/memprof-limits/memprof-limits.0.1/opam
@@ -1,0 +1,45 @@
+opam-version: "2.0"
+maintainer: "Guillaume Munch-Maccagnoni <Guillaume.Munch-Maccagnoni@inria.fr>"
+authors: [
+  "Guillaume Munch-Maccagnoni <Guillaume.Munch-Maccagnoni@inria.fr>"
+]
+license: "LGPL-3.0-only"
+homepage: "https://gitlab.com/gadmm/memprof-limits"
+bug-reports: "https://gitlab.com/gadmm/memprof-limits/issues"
+depends: [
+  "ocaml"
+  "ocaml-variants" {>="4.11.0"}
+  "dune" {build & >= "1.0"}
+]
+build: ["dune" "build" "-p" name "-j" jobs]
+dev-repo: "git+https://gitlab.com/gadmm/memprof-limits.git"
+synopsis: "Memory limits based on memprof"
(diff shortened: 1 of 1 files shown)


gadmm:
Thanks @kit-ty-kate, I have already upstreamed your fix!

kit-ty-kate:
thanks!

gadmm:
Hi @kit-ty-kate, I see many opam files where the dune dependency is tagged with the `build` variable like I did initially. I removed it based on your correction but I am wondering if I should add it back. What was the reason to remove it?

kit-ty-kate:
The reason behind this is a long standing bug in dune: https://github.com/ocaml/dune/issues/2147.

What packages did you see where dune is tagged as a build dependency? I can't find any in opam-repository (if there is this is something that we should fix)

gadmm:
I grepped my local download of the opam repository, which I usually use instead for auditing OCaml code.
- I downloaded it up to two years ago, so maybe it was in the meanwhile that you removed all the build tags?
- These are the upstream opam files, not the ones on opam-repository, so maybe the fixes were not upstreamed?

If this does not explain the difference then I can send the list of what I have found.

kit-ty-kate:
> I downloaded it up to two years ago, so maybe it was in the meanwhile that you removed all the build tags?

Yes indeed, this was done two years ago in https://github.com/ocaml/opam-repository/pull/14266

> These are the upstream opam files, not the ones on opam-repository, so maybe the fixes were not upstreamed?

Most likely

gadmm:
Ok thanks!

PR #16412, opened 2020-05-11
Luv 0.5.3: libuv binding

aantron:
- Interpret port numbers correctly as unsigned short integers (aantron/luv@23d2ac6).

### DIFF
--- a/packages/luv/luv.0.5.3/opam
+++ b/packages/luv/luv.0.5.3/opam
@@ -1,0 +1,41 @@
+opam-version: "2.0"
+
+synopsis: "Binding to libuv: cross-platform asynchronous I/O"
+
+version: "0.5.3"
+license: "MIT"
+homepage: "https://github.com/aantron/luv"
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
The bash script seems to fail somewhere in the middle when OCaml is also installed system wide but the current one is the one built by opam:
```
-         bash src/c/dlluv.so,src/c/libuv.a (exit 2)
```
Could you have a look at it? https://travis-ci.org/github/ocaml/opam-repository/jobs/685765509

aantron:
libuv has had a release in the meantime (yesterday), so I'll retag in addition to looking at the Travis issue.

aantron:
I wasn't able to reproduce the problem in Travis locally, by installing a system OCaml (from APT). I suggest that I'll debug the Travis problem in a separate dummy PR with a "release" that has all the libuv build verbosity options enabled.

aantron:
I updated the release.

avsm:
Thanks!

PR #16415, opened 2020-05-11
soupault: Specify dependency on tsort < 2.0 to prevent breaking builds.

dmbaturin:
[tsort](https://opam.ocaml.org/packages/tsort/) is about to make a about to make a [breaking change](https://github.com/dmbaturin/ocaml-tsort/pull/1).

This will prevent buiilds of older soupault versions from breaking due to that change.


dmbaturin:
Travis build fails because the CI script wants to install camlp4 on OS X and can't.

kit-ty-kate:
Thanks!

PR #16416, opened 2020-05-11
crowbar.0.2: Untag dune as a build dependency

kit-ty-kate:
cc @stedolan could you return the fix upstream?


stedolan:
Thanks for the fix! Upstreamed now.

PR #16419, opened 2020-05-12
Solo5: Disable Genode bindings

mato:
This is a followup to #16368 where the consensus was to disable the
existing solo5-bindings-genode due to these being unsupported upstream
except on very specific toolchains. See also Solo5/solo5#446.


mato:
Note the camelus failure is intentional, since we're trying to disable/de-publish the Genode bindings here.

mato:
Ping @avsm / @kit-ty-kate would like to get this merged so that we are in a position to proceed with Solo5 releases without CI failures.

PR #16425, opened 2020-05-12
[new release] conduit-mirage (2.2.0)

hannesm:
CHANGES:

* conduit-mirage adapt to dns-client 4.5.0 (mirage/ocaml-conduit#314 @hannesm)


hannesm:
I added a conflict with existing mirage packages, since this will otherwise fail. The mirage package hardcodes the functor parameters for conduit, which have changed in this release. An upcoming mirage release will make use of this conduit-mirage package.

kit-ty-kate:
Thanks a lot!

PR #16427, opened 2020-05-12
[new release] mssql (2.1.0)

brendanlong:

- Documentation: <a href="https://arenadotio.github.io/ocaml-mssql">https://arenadotio.github.io/ocaml-mssql</a>

### Fixed

- Fix `execute_pipe`'s exception handling. Previously, exceptions could cause the pipe to close and appear to finish
  successfully. Now, the pipe will be left open and an exception will be thrown.
- Correctly parser SQL queries containing single quotes in comments. Previously, a $ parameter surrounded by
  comments containing quotes could fail to be replaced and would be treated as a literal currency value.
- Fix hang if `execute_map`'s `~f` throws an exception. This was caused by `f` being executed in a different
  Async context, so the exception wouldn't be caught by the correct context.
- Ensure result sets can't be skipped when using `execute_pipe`. This makes us finish reading the previous result
  set before continuing to the next one.
- Cancel query results immediately after an exception instead of waiting for the next query. This can prevent
  holding onto resources for an excessive amount of time.

### DIFF
--- a/packages/mssql/mssql.2.1.0/opam
+++ b/packages/mssql/mssql.2.1.0/opam
@@ -1,0 +1,48 @@
+opam-version: "2.0"
+synopsis: "Async SQL Server client using FreeTDS"
+description:
+  "Mssql wraps FreeTDS in a nicer and safer interface, with support for parameterized queries, thread-based async IO, and a thread pool."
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Thanks a lot!

PR #4099, opened 2015-05-15
irmin 0.9.[0-4] are not compatible with OCaml 4.02.2

planar:
because of `-warn-error A`

PR #16430, opened 2020-05-12
Package libdash.0.1.1

mgree:
### `libdash.0.1.1`
Bindings to the dash shell's parser

---
* Homepage: https://github.com/mgree/libdash
* Source repo: git+https:///github.com/mgree/libdash
* Bug tracker: https://github.com/mgree/libdash/issues

---

### DIFF
--- a/packages/libdash/libdash.0.1.1/opam
+++ b/packages/libdash/libdash.0.1.1/opam
@@ -10,7 +10,6 @@ depends: [
   "ocamlfind" {>= "1.8.0"}
   "ctypes" {>= "0.11.5"}
   "ctypes-foreign" {>= "0.4.0"}
-  "opam-installer" {>= "2.0.0"}
   "conf-autoconf" {build}
   "conf-aclocal" {build}
   "conf-libtool" {build}
@@ -30,9 +29,6 @@ build: [
   ["./mk_dot_install.sh"]
   ["./ldconfig.sh"] # fix up .so files if ldconfig didn't do it
   [make "-C" "ocaml" "test"] {with-test}
-]
-install: [
-  ["opam-installer" "--prefix=%{prefix}%" "libdash.install"]
 ]
 dev-repo: "git+https:///github.com/mgree/libdash"
 url {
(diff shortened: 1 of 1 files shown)


mgree:
Can anyone explain what the INSTALL_LOCAL flag is doing? I couldn't find documentation anywhere and the Travis CI seems different from the last time I looked at it...

mgree:
I would appreciate any insight into why [my own tests](https://travis-ci.com/github/mgree/libdash/builds/165758236) succeed on FreeBSD but not the [opam-repository tests](https://travis-ci.com/github/mgree/opam-repository/builds/165762131)... I suspect it's a linking issue, but the error message is uninformative.

avsm:
Your invocation of the test seems different in your version.

```
 LD_LIBRARY_PATH=$(opam var libdash:lib) opam exec -- make -C test test
```

mgree:
Gah! I thought I had taken that out. Thank you.

mgree:
Looks like `conf-aclocal` was the trick for Alpine. So: all the critical stuff is working.

My current best guess about the FreeBSD issue is the `#!/bin/bash` in my testing script---that would explain error 127 in `make` (program not found). Why that script would work in my CI but not yours is baffling, but... FreeBSD keeps `bash` in `/usr/local/bin/bash`, I believe. If the tests don't work this time, I'll call it for now... 🤷‍♀️

mgree:
Well, some transient error broke the OCaml 4.11.0+trunk build and the OpenSUSE build, but it looks like that fixed FreeBSD.

If you want an all green build, I suspect that simply retriggering the builds should do it.

I'm hopeful this is the end of this already-too-long PR. Maybe the next version bump will be simpler? 🤞

mgree:
Thank you, good catch!

I needed opam-installer when I had a different resting regime in place (which meant I had more actions in the install phase and couldn't use the default; see [ocaml/opam  #4099](https://github.com/ocaml/opam/issues/4099).

kit-ty-kate:
Thanks!

PR #16433, opened 2020-05-13
Package nmea.0.1.5

dakk:
### `nmea.0.1.5`
Nmea parser
NMEA sentence parser library for OCaml

- Compass sentences
- ZDA sentence
- Extended test suite
- Fixing parsing error with real old GPS devices

Sorry for the new release so near to the first, but I tested the library with other real instruments, fixing parsing errors (there is no fixed standard for empty fields, so every instrument adopt a different approach that have to be handled via software)

---
* Homepage: https://github.com/dakk/nmea
* Source repo: git+https://github.com/dakk/nmea.git
* Bug tracker: https://github.com/dakk/nmea/issues

---


kit-ty-kate:
Thanks!

PR #16440, opened 2020-05-14
[new release] elpi (1.11.1)

gares:
CHANGES:

- Fix to the opam file, ppxlib is required to be >= 0.12.0 and
  ocaml-migrate-parsetree >= 1.6.0. Moreover we disable tests on Alpine linux
- Print locations in such a way that VScode can understand then, and click jump
  to type errors


kit-ty-kate:
Could you return the fix to the opam and dune-project file upstream? Maybe with a small comment describing why it is this way if you think that would help

gares:
ok

kit-ty-kate:
Thanks a lot!

Issue #451, opened 2013-02-25
problem with ocurl package


UnixJunkie:
On 02/25/2013 11:09 AM, oliver wrote:> On Mon, Feb 25, 2013 at 10:26:34AM +0900, Francois Berenger wrote:

> > On 02/25/2013 09:50 AM, oliver wrote:
> > 
> > > Hello Roberto,
> > > 
> > > On Mon, Feb 18, 2013 at 08:34:35AM +0100, Roberto Di Cosmo wrote:
> > > 
> > > > Dear Oliver,
> > > >      just use the Opam package manager for OCaml and you will be fine.
> > > > [...]
> > > 
> > > Do you mean something like ths following?!
> > > # 
> > > 
> > > oliver@siouxsie:~$ . ocamlbrew/ocaml-4.00.1/etc/ocamlbrew.bashrc
> > > oliver@siouxsie:~$ opam install ocurl
> > > The following actions will be performed:
> > > - install ocurl.0.5.4
> > >   1 to install | 0 to reinstall | 0 to upgrade | 0 to downgrade | 0 to remove
> > > 
> > > =-=-= ocurl.0.5.4 =-=-=
> > > The archive for ocurl.0.5.4 is in the local cache.
> > > Extracting /home/oliver/ocamlbrew/ocaml-4.00.1/.opam/archives/ocurl.0.5.4+opam.tar.gz.
> > > Building ocurl.0.5.4:
> > >    ./configure
> > >    make
> > >    make install
> > > The compilation of ocurl.0.5.4 failed.
> > > Uninstalling ocurl.0.5.4:
> > >    ocamlfind remove curl
> > > 
> > > ==== ERROR [while installing ocurl.0.5.4] ====
> > > # command         make
> > > # path            /home/oliver/ocamlbrew/ocaml-4.00.1/.opam/4.00.1/build/ocurl.0.5.4
> > > # exit-code       2
> > > # env-file        /home/oliver/ocamlbrew/ocaml-4.00.1/.opam/4.00.1/build/ocurl.0.5.4/ocurl-1872ec.env
> > > # stdout-file     /home/oliver/ocamlbrew/ocaml-4.00.1/.opam/4.00.1/build/ocurl.0.5.4/ocurl-1872ec.out
> > > # stderr-file     /home/oliver/ocamlbrew/ocaml-4.00.1/.opam/4.00.1/build/ocurl.0.5.4/ocurl-1872ec.err
> > > ### stdout
> > > 
[...]

PR #16443, opened 2020-05-14
Solo5 0.6.5

mato:
New features:

* Stop host kernels from attempting to execute Solo5 binaries. This improves
  both the user experience on some hosts (e.g. "No such file or directory" vs.
  "Segmentation fault" on Linux) and overall security posture by forcing the
  host kernel ELF loader to bail out earlier rather than actually jumping to
  the unikernel code. (#442)
* hvt: Full support for W^X and correct guest-side page protections on OpenBSD
  6.7+ systems with EPT. (#447)
* hvt: capsicum(4) sandbox for the hvt tender on FreeBSD 12+. (#366)

Bug fixes:

* hvt: Fix hang in `HVT_HYPERCALL_POLL`. On Linux hosts, if `solo5_yield()` was
  called with a deadline that has already passed and the unikernel was not using
  any network devices then the underlying hypercall would hang forever. Not
  known to affect any existing code in production. (#460)

Other notable changes:

* muen: Muen ABI updates, now uses ABI version 2 on the Solo5 side. Muen
  commit f10bd6b or later is required. (#454, #448)
* genode: Support for Genode is limited by toolchain issues and Genode bindings
  are no longer built by default. (#446, see also ocaml/opam-repository#16368)
* Improvements to the build system on BSD/clang hosts. System headers
  (sys/endian.h, osreldate.h) that were mistakenly being installed into the
  Solo5-provided include paths have been removed. For OCaml/MirageOS users,
  ocaml-freestanding 0.6.0 or later is now required. (#453, #455, #457, #461,
  see also mirage/ocaml-freestanding#77)
* Improvements to built-in self tests. (#451)
* Fix build failures with GCC >= 10. (#459)

Known issues:

* Full W^X support / correct guest-side page protections are currently only
  available on the "spt" target on Linux, and the "hvt" target on OpenBSD 6.7
  or later. (#303)
* On OpenBSD, "hvt" operation with multiple network devices results in packet
  loss. This appears to be a bug in kqueue(2) but we have no confirmation from
  upstream. (#374)
* virtio-net is not functional on at least QEMU 5.0 and possibly earlier
  versions. QEMU versions up to and including 3.1.0 are known to work. (#463)

Acknowledgements:

* Thanks to Adam Steen (@adamsteen) for pushing for OpenBSD kernel support for
  manipulating guest EPT mappings, bringing full W^X to hvt on OpenBSD 6.7 or
  later.
* Thanks to Adrian-Ken Rueegsegger (@kensan) for the Muen updates.
* Thanks to Anurag Soni (@anuragsoni) for diagnosing and fixing the build on
  systems with GCC >= 10.
* Thanks to Hannes Mehnert (@hannesm) for diagnosing #460 and for help with
  testing BSD/clang build system changes and generally helping out.
* Thanks to Stefan Grundmann (@sg2342) for the capsicum(4) hvt tender sandbox
  on FreeBSD.

### DIFF
--- a/packages/solo5-bindings-hvt/solo5-bindings-hvt.0.6.5/opam
+++ b/packages/solo5-bindings-hvt/solo5-bindings-hvt.0.6.5/opam
@@ -14,20 +14,15 @@ build: [
   [make "V=1" "CONFIG_SPT=" "CONFIG_VIRTIO=" "CONFIG_MUEN=" "CONFIG_GENODE="]
 ]
 install: [make "V=1" "CONFIG_SPT=" "CONFIG_VIRTIO=" "CONFIG_MUEN=" "CONFIG_GENODE=" "install-opam-hvt" "PREFIX=%{prefix}%"]
-remove: [
-  ["touch" "./Makeconf"]
-  [make "V=1" "uninstall-opam-hvt" "PREFIX=%{prefix}%"]
-]
 depends: [
   "conf-pkg-config"
   "conf-libseccomp" {build & os = "linux"}
 ]
 depexts: [
   ["linux-headers"] {os-distribution = "alpine"}
-  ["linux-libc-dev"] {os-distribution = "debian"}
   ["kernel-headers"] {os-distribution = "fedora"}
   ["kernel-headers"] {os-distribution = "rhel"}
-  ["linux-libc-dev"] {os-distribution = "ubuntu"}
+  ["linux-libc-dev"] {os-family = "debian"}
 ]
 conflicts: [
   "ocaml-freestanding" {< "0.6.0"}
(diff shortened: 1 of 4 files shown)


kit-ty-kate:
Would it be ok to remove the `remove` fields and change it in the source repository? Since opam 2.0 they are not necessary unless it is to remove a configuration file or something that has been created after installation.

Also,
```
["linux-libc-dev"] {os-distribution = "debian"}
["linux-libc-dev"] {os-distribution = "ubuntu"}
```
can be reduced to:
```
["linux-libc-dev"] {os-family = "debian"}
```
Could you also change this? This would allow to handle Raspbian and similar derived distributions.

mato:
@kit-ty-kate Done here, I've filed an issue to do it in the source repository so I don't forget.

kit-ty-kate:
Thanks a lot!

PR #16444, opened 2020-05-14
vcardgen v1.2

vzaliva:


kit-ty-kate:
Dune does not support the build tag (see https://github.com/ocaml/dune/issues/2147) and the constraint on it was incorrect as you `dune-project` file says this package requires at least dune 2.0.

Could you return the fix to the opam file upstream?

kit-ty-kate:
Thanks!

PR #16446, opened 2020-05-15
[new release] clangml.4.1.0, clangml-transforms.0.25, conf-libclang.10, pattern.0.2.0

thierry-martinez:
This pull-request releases clangml.4.1.0, compatible with LLVM 10.0.0 and OCaml 4.11, and its dependencies.


thierry-martinez:
I believe that `pkg install devel/clang devel/llvm` used to work on FreeBSD but I don't manage to find which packages to install in order to have `libclang*.so` on FreeBSD now. I think we can wait for proper FreeBSD support in a future PR.

packages/clangml/clangml.4.1.0/opam
```diff
@@ -0,0 +1,35 @@
...
+authors: ["Thierry Martinez <thierry.martinez@inria.fr>"]
+license: "BSD-2-Clause"
+homepage: "https://memcad.gitlabpages.inria.fr/clangml/"
+doc: "https://memcad.gitlabpages.inria.fr/clangml/doc/clangml/index.html"
+bug-reports: "https://gitlab.inria.fr/memcad/clangml/issues"
+depends: [
+  "conf-libclang"
+  "conf-ncurses"
+  "conf-zlib"
+  "dune" {>= "1.10.0"}
```
mseri, line 17: According to the dune-project file, you are using lang 1.11.
Can you update the bound also upstream?

packages/clangml-transforms/clangml-transforms.0.25/opam
```diff
@@ -0,0 +1,35 @@
...
+synopsis: "Code transformers for clangml"
+maintainer: ["Thierry Martinez <thierry.martinez@inria.fr>"]
+authors: ["Thierry Martinez <thierry.martinez@inria.fr>"]
+license: "BSD-2-Clause"
+homepage: "https://gitlab.inria.fr/memcad/clangml-transforms"
+doc: "https://gitlab.inria.fr/memcad/clangml-transforms"
+bug-reports: "https://gitlab.inria.fr/memcad/clangml-transforms/issues"
+depends: [
+  "ocaml" {>= "4.03.0" & < "4.12.0"}
+  "dune" {>= "2.2.0"}
```
mseri, line 11: According to the dune-project file, you are using lang 1.11. This is stricter, so it is not a problem, but maybe you want to relax it.

packages/clangml/clangml.4.1.0/opam
thierry-martinez: Fixed in https://github.com/ocaml/opam-repository/pull/16446/commits/5d0e453624df25718384b54691295e442158be4c (and upstream: https://gitlab.inria.fr/memcad/clangml/-/commit/003002d33d2fe55f7dd80a9ad24470018eb7e6d8 )
Thanks!

packages/clangml-transforms/clangml-transforms.0.25/opam
thierry-martinez: Fixed in https://github.com/ocaml/opam-repository/pull/16446/commits/c5f23b5623cd7e084a3e675870ccc187a74ac275 (and upstream: https://gitlab.inria.fr/memcad/clangml-transforms/-/commit/98d2e9b0aca914fca30f8ba0084f7a51e35fb87e )
Thanks!

mseri:
Thanks!

thierry-martinez:
Last commits allow clangml.4.0.1(the previous release) to be compiled with Clang 9.0.1: it was made necessary to satisfy the CI because otherwise clangml.4.0.1 failed to compile on Debian testing, which has just switched to LLVM 9.0.1.

In the future, I think that `conf-libclang.1.0.0` (the "old" version) will have to set explicitly more versions for packages in the depexts section to keep Clang version below 10. However, I think it is not opportune to do that now, because it will prevent depext to succeed on old systems that do not know the fixed versions.

mseri:
Thanks for all the fixes!

PR #16451, opened 2020-05-15
mirage-crypto-entropy does not support 32 bit

hannesm:
see report in https://github.com/mirage/mirage-crypto/issues/60, and the next release will have 32 bit report.


kit-ty-kate:
Thanks!

PR #16452, opened 2020-05-16
new file:   packages/parany/parany.11.0.2/opam

UnixJunkie:
Major release; existing client software that is not yet updated should use parany < 11.0.0.

```
val Parany.run ?preserve:bool -> ?core_pin:bool -> ?csize:int -> int ->
  demux:(unit -> 'a) -> work:('a -> 'b) -> mux:('b -> unit) -> unit                                                                                                                                   
```
ncores is no more a labeled paramete. Rather, it is the only anonymous parameter now.
New optional parameters with default values:
~preserve (default = false); to reflect input order in output. Off by default for performance reasons.
Also note that while the output will be in the same order as the input. Computations might still
happen out of order for performance reasons.
~csize (default=1); to tweak the granularity of a computation.
~core_pin (default=false); pin processes to core (off by default because powerful computers are usually shared by several users; also, you could run several parany computations on one computer
and you don't want that they all stay pinned to the same cores).

Thanks to @darrenldl for some non trivial code factorization.

### DIFF
--- a/packages/hts_shrink/hts_shrink.2.1.1/opam
+++ b/packages/hts_shrink/hts_shrink.2.1.1/opam
@@ -11,7 +11,7 @@ depends: [
   "batteries"
   "dolog" {>= "4.0.0"}
   "parmap"
-  "parany" {>= "5.0.0"}
+  "parany" {>= "5.0.0" & < "11.0.0"}
   "minicli" {>= "5.0.0"}
   "bitv"
   "bst"
--- a/packages/molenc/molenc.7.0.1/opam
+++ b/packages/molenc/molenc.7.0.1/opam
@@ -25,7 +25,7 @@ depends: [
   "dune" {>= "1.11"}
   "minicli"
   "ocaml"
-  "parany" {>= "9.0.0"} # for the Parmap compatibility module
+  "parany" {>= "9.0.0" & < "11.0.0"}
 ]
 synopsis: "Molecular encoder/featurizer using rdkit and OCaml"
 description: """Chemical fingerprints are lossy encodings of molecules.
--- a/packages/molenc/molenc.8.0.2/opam
+++ b/packages/molenc/molenc.8.0.2/opam
@@ -25,7 +25,7 @@ depends: [
   "dune" {>= "1.11"}
   "minicli"
   "ocaml"
-  "parany" {>= "9.0.0"} # for the Parmap compatibility module
+  "parany" {>= "9.0.0" & < "11.0.0"}
 ]
 synopsis: "Molecular encoder/featurizer using rdkit and OCaml"
 description: """Chemical fingerprints are lossy encodings of molecules.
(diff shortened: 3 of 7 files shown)


kit-ty-kate:
This package fails on OCaml 4.02:
```

#=== ERROR while compiling parany.11.0.1 ======================================#
# context              2.0.7 | linux/x86_64 | ocaml-base-compiler.4.02.3 | pinned(https://github.com/UnixJunkie/parany/archive/v11.0.1.tar.gz)
# path                 ~/.opam/4.02/.opam-switch/build/parany.11.0.1
# command              ~/.opam/4.02/bin/dune build -p parany -j 72
# exit-code            1
# env-file             ~/.opam/log/parany-23-b57320.env
# output-file          ~/.opam/log/parany-23-b57320.out
### output ###
#     ocamldep src/.parany.objs/parany.ml.d (exit 2)
# (cd _build/default && /home/opam/.opam/4.02/bin/ocamldep.opt -modules -impl src/parany.ml) > _build/default/src/.parany.objs/parany.ml.d
# File "src/parany.ml", line 166, characters 29-30:
# Error: Syntax error: operator expected.
```
It should be fine to constrain the ocaml version.

Also, all of the packages depending on this now fail:
* https://ci.ocaml.org/log/saved/docker-run-a1ad993d7b70c3822c6367e97d875a91/c8bf0f7dfd939cb505b15719c922be0dc85c3b5f
* https://ci.ocaml.org/log/saved/docker-run-9a16b80a73c76f24ac920301d42883df/aa8b10b9f9f0db66ec7d85fd83473470158d23c1
* https://ci.ocaml.org/log/saved/docker-run-dc2dbca076ca7dcb6ff0f575255d1b09/76db4a82c04ff3d6fd33b211590577c42aada36b
* https://ci.ocaml.org/log/saved/docker-run-334023ea73ee6b81356e95c702fd2b9d/f647ab3584f1ff16f369ab90fbd9d899a4ca36b2
* https://ci.ocaml.org/log/saved/docker-run-2cc9113be59bcb013b437a9429823301/629c429e5b1c2c02fd47ca8bdb5804c1f02e6bc0

I suppose this is expected? If it is, I'll make a PR to constrain every packages depending on parany.

UnixJunkie:
Line 166 is:
```
  let demux_work_mux (type a b)
```
@darrenldl is your contribution only for OCaml >= 4.02.0 ?
If there is a way to make it compile on more OCaml versions, that would be nice.

darrenldl:
@UnixJunkie Moving `demux_work_mux` to outside of run and remove `(type a b)` should work - I'll make a PR shortly.

UnixJunkie:
@darrenldl don't bother with a PR, I'll try your suggested fix.

darrenldl:
@UnixJunkie Sounds good. I can confirm the fix works in 4.02.0 - https://github.com/darrenldl/parany/blob/master/src/parany.ml

UnixJunkie:
fixed. I retagged to 11.0.2 then.

kit-ty-kate:
4.02 now works. So, are the revdeps failures expected?

UnixJunkie:
Yes, current users should all use parany < 11.0.0, unless they update their code.

UnixJunkie:
I just added some constraints to fix the revdeps.

mseri:
All the remaining failures seem to be due to r or rdkit missing

PR #16455, opened 2020-05-16
Update torch.

LaurentMazare:
Update the torch package to version 0.9 which handles PyTorch 1.5.

### DIFF
--- a/packages/torch/torch.0.9/opam
+++ b/packages/torch/torch.0.9/opam
@@ -1,0 +1,43 @@
+opam-version: "2.0"
+bug-reports:  "https://github.com/LaurentMazare/ocaml-torch/issues"
+homepage:     "https://github.com/LaurentMazare/ocaml-torch"
+dev-repo:     "git+https://github.com/LaurentMazare/ocaml-torch.git"
+maintainer:   "Laurent Mazare <lmazare@gmail.com>"
+authors:      [ "Laurent Mazare" ]
+
+build: [["dune" "build" "-p" name "-j" jobs]]
+
+depends: [
+  "base" {>= "v0.11.0" & < "v0.14"}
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
This package fails to build on MacOS and old ubuntu with:
```
#=== ERROR while compiling torch.0.9 ==========================================#
# context     2.0.7 | macos/x86_64 | ocaml-system.4.09.0 | file:///Users/travis/build/ocaml/opam-repository
# path        ~/.opam/ocaml-system/.opam-switch/build/torch.0.9
# command     ~/.opam/opam-init/hooks/sandbox.sh build dune build -p torch -j 1
# exit-code   1
# env-file    ~/.opam/log/torch-54045-865888.env
# output-file ~/.opam/log/torch-54045-865888.out
### output ###
#   ocamlmklib src/wrapper/dlltorch_core_stubs.so,src/wrapper/libtorch_core_stubs.a
# Unknown option -Wl,--no-as-needed
#     ocamlopt src/wrapper/torch_core.cmxs (exit 2)
# (cd _build/default && /usr/local/bin/ocamlopt.opt -w -40 -g -shared -linkall -I src/wrapper -o src/wrapper/torch_core.cmxs src/wrapper/torch_core.cmxa)
# ld: unknown option: --no-as-needed
# clang: error: linker command failed with exit code 1 (use -v to see invocation)
# Error: Error during linking
```

LaurentMazare:
Thanks for pointing this out @kit-ty-kate I've pushed some fixes for this and it now works on macos in the CI environment.
It still fails on xenial but from what I understand of the logs, this is caused by running out of memory when compiling some large generated file.

kit-ty-kate:
Thanks!

PR #16461, opened 2020-05-17
Add webtest.2.0.0 and webtest-js.2.0.0

johnelse:


johnelse:
Looks like the travis failure is `js_of_ocaml` failing to build with OCaml 4.11

mseri:
Do you know if the js version compiles with 4.11 using maybe a dev version of js_of_ocaml or if we need an upper bound?

packages/webtest-js/webtest-js.2.0.0/opam
```diff
@@ -0,0 +1,25 @@
...
+  [ "dune" "build" "-p" name "-j" jobs ]
+  [ "dune" "runtest" ] {with-test}
+]
+
+depends: [
+  "dune" {>= "1.1"}
+  "js_of_ocaml" {>= "3.4.0"}
+  "js_of_ocaml-ppx" {>= "3.4.0"}
+  "ocaml" {>= "4.02.1"}
+  "webtest"
```
mseri, line 18: We tend to use stricter versioning when packages are released together. Would you mind if we add the following constraint? Could you also port it upstream?

```suggestion
  "webtest" {=version}
```

johnelse: Sure! I've added this change.

mseri:
Thanks!

johnelse:
Thank you!

PR #16465, opened 2020-05-18
first ocp_reveal release

zapashcanon:
Hi,

Here's the first release of [ocp_reveal](https://github.com/OCamlPro/ocp-reveal).

Cheers !

### DIFF
--- a/packages/ocp_reveal/ocp_reveal.0.1/opam
+++ b/packages/ocp_reveal/ocp_reveal.0.1/opam
@@ -1,0 +1,39 @@
+opam-version: "2.0"
+synopsis: "OCaml bindings for Reveal.js, an HTML presentation framework"
+description: "OCaml bindings for Reveal.js, an HTML presentation framework"
+maintainer: ["OCamlPro <contact@ocamlpro.com>"]
+authors: ["Çagdas Bozman"]
+license: "LGPL"
+homepage: "https://github.com/OCamlPro/ocp-reveal"
+doc: "github.com/OCamlPro/ocp-reveal"
+bug-reports: "https://github.com/OCamlPro/ocp-reveal/issues"
+depends: [
+  "ocaml" {>= "4.05"}
+  "dune" {>= "2.5"}
+  "omd" {>= "1.3"}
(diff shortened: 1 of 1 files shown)


zapashcanon:
The Travis failure on trunk seems to be unrelated.

packages/ocp_reveal/ocp_reveal.0.1/opam
```diff
@@ -0,0 +1,39 @@
...
+synopsis: "OCaml bindings for Reveal.js, an HTML presentation framework"
+description: "OCaml bindings for Reveal.js, an HTML presentation framework"
+maintainer: ["OCamlPro <contact@ocamlpro.com>"]
+authors: ["Çagdas Bozman"]
+license: "LGPL"
+homepage: "https://github.com/OCamlPro/ocp-reveal"
+doc: "github.com/OCamlPro/ocp-reveal"
+bug-reports: "https://github.com/OCamlPro/ocp-reveal/issues"
+depends: [
+  "ocaml" {>= "4.05"}
```
mseri, line 11:
```suggestion
  "ocaml" {>= "4.05" & < "4.11"}
```
The failure seems to indicate incompatibility with ocaml 4.11

mseri: Because in that case I would merge directly, without changes

zapashcanon: I don't see why it wouldn't work with 4.11.

zapashcanon:
Thanks !

avsm:
Just one minor point - did you intend to call the package `ocp_reveal` instead of `ocp-reveal`? The convention is to use dashes and not underscores in opam package names.

zapashcanon:
It was intended. I think `-` are confusing because you can't use them everywhere. Whereas `_` can be used in opam packages, libraries and modules names. Ideally I'll try to rename the GitHub repository to ocp_indent but later.

```sh
cd opam-repository
ls | grep _ -c
332
ls | grep - -c
1167
```

The conventions is not completely clear. There's still hope for `_`. :D

PR #16467, opened 2020-05-18
fix for ppx_meta_conv.4.0.0

camlspotter:


mseri:
I have manually checked the contents of the tarballs. This is good to go

PR #16469, opened 2020-05-18
[new release] mirage (3.7.7)

hannesm:
CHANGES:

* handle errors from Bos.OS.Cmd.run_out
* use PREFIX if defined (no need to call "opam config var prefix")
* adapt to conduit 2.2.0, tls 0.12, mirage-crypto 0.7.0 changes


kit-ty-kate:
Thanks!

Issue #15867, opened 2020-02-19
imagelib.20160413 should depend on ocaml  < 4.06.0


ddeclerck:
Since imagelib.20160413 uses mutable strings, it should depend on ocaml < 4.06.0 (just like imagelib.20170118).

mseri:
I had missed this issue. I'll send a fix in the afternoon. Thanks for reporting it

PR #16477, opened 2020-05-19
imagelib.20160413: add missing upper bound

mseri:
Reported in https://github.com/ocaml/opam-repository/issues/15867

### DIFF
--- a/packages/imagelib/imagelib.20160413/opam
+++ b/packages/imagelib/imagelib.20160413/opam
@@ -9,7 +9,7 @@ build: [make]
 install: [make "install"]
 remove: [make "uninstall"]
 depends: [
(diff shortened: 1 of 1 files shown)


mseri:
Looks like this is no longer downloadable. Should I point it to the opam cache?

kit-ty-kate:
> Looks like this is no longer downloadable. Should I point it to the opam cache?

I think it might be the only thing to do. @craff is there any archive available elsewhere? The Patoline website died.

kit-ty-kate:
Fixed in bulk PR https://github.com/ocaml/opam-repository/pull/16534

mseri:
Thanks

PR #16484, opened 2020-05-20
Add upper bound constraints to all reverse dependencies of Jane Street packages

xclerc:
In preparation of the upcoming v0.14 release of all our packages,
we are adding upper bound constraints to reverse dependencies
so that the release does not break anything.


kit-ty-kate:
Thanks a lot!

PR #16485, opened 2020-05-20
Release conduit mirage 2.2.1

hannesm:
which only change is that it has its dependencies in both opam and dune corrected (this is the reason for marking 2.2.0 as unavailable)


kit-ty-kate:
Thanks!

PR #16495, opened 2020-05-22
+ sqlgg 20200521

ygrek:


kit-ty-kate:
Your `dune-project` file, says this package requires at least dune 2.0. Could you return the fix in the opam file upstream?

kit-ty-kate:
Also, the directory layout is not the one used by opam-repository and our CI does not realize this is a new package. Could you move the `sqlgg.20200521` directory to the `sqlgg` directory?

ygrek:
Done. Thanks for spotting, not sure how it happened :/

kit-ty-kate:
Thanks!

Issue #12681, opened 2018-09-23
conf-mpfr fails on Mac OS with MacPorts


examachine:
This happens because the Mac Ports prefix is /opt/local. /usr is hardcoded in build rule which isn't portable:
```
build: [
  ["cc" "test.c" "-I/usr/local/include" "-L/usr/local/lib" "-lgmp" "-lmpfr"]
]
```
pkg-config ought to be used instead:
```
$ pkg-config --libs mpfr
-L/opt/local/lib -lmpfr -lgmp
```

```
$ opam install conf-mpfr
The following actions will be performed:
  ∗ install conf-mpfr 1

<><> Gathering sources ><><><><><><><><><><><><><><><><><><><><><><><><><><>  🐫 

<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><>  🐫 
[ERROR] The compilation of conf-mpfr failed at
        "/Volumes/Centauri/Users/malfunct/.opam/opam-init/hooks/sandbox.sh
        build cc test.c -I/usr/local/include -L/usr/local/lib -lgmp -lmpfr".

#=== ERROR while compiling conf-mpfr.1 ========================================#
# context     2.0.0 | macos/x86_64 | ocaml-system.4.05.0 | https://opam.ocaml.org#7cd43695
# path        /Volumes/Centauri/Users/malfunct/.opam/default/.opam-switch/build/conf-mpfr.1
# command     /Volumes/Centauri/Users/malfunct/.opam/opam-init/hooks/sandbox.sh build cc test.c -I/usr/local/include -L/usr/local/lib -lgmp -lmpfr
# exit-code   1
# env-file    /Volumes/Centauri/Users/malfunct/.opam/log/conf-mpfr-7786-9b8556.env
# output-file /Volumes/Centauri/Users/malfunct/.opam/log/conf-mpfr-7786-9b8556.out
### output ###
# test.c:2:10: fatal error: 'mpfr.h' file not found
# #include <mpfr.h>
#          ^~~~~~~~
# 1 error generated.



<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><>  🐫 
┌─ The following actions failed
│ λ build conf-mpfr 1
└─ 
╶─ No changes have been performed
```

examachine:
stale? was I supposed to submit a patch?

PR #16499, opened 2020-05-22
conf-mpfr: fix macports build

mseri:
This should fix https://github.com/ocaml/opam-repository/issues/12681
Tested locally. I think for the future it would be a good idea to make
these conf packages use pkgconfig instead.

### DIFF
--- a/packages/conf-mpfr/conf-mpfr.2/files/test.c
+++ b/packages/conf-mpfr/conf-mpfr.2/files/test.c
@@ -1,0 +1,11 @@
+#include <stdio.h>
+#include <mpfr.h>
+
+// compile with: gcc test.c -lmpfr -lgmp
+
+int main (void) {
+  printf ("MPFR library: %-12s\nMPFR header:  %s (based on %d.%d.%d)\n",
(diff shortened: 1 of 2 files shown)


pmetzger:
Hardcoding does indeed seem inferior to using `pkg-config` if it is available. Dune's Configurator library has built-in support for running pkg-config and generating build flags for C programs and libraries based on the `pkg-config` output.

mseri:
@pmetzger can you open an issue on the issue tracker? It is a bit more subtle than using pkg-config in that there are some pkgconfig related  naming issues and some packages that refuse to provide the pkgconfig configuration file. But I think we should improve the conf situation and probably there will be a semi-universal way of doing it if we think hard enough

pmetzger:
@mseri The trick usually is to default to something sane if pkg-config returns nothing. FWIW, the Dune Configurator library has nice tools for this.

mseri:
Yes I agree with you. I am familiar with it. I can point out a number of rather standard libraries whose defaults are heavily distribution dependant, that is my main worry at the moment. I’d like to have a solution for as many packages as possible. A configurable exectuable which could be used for most conf packages would be great

PR #16505, opened 2020-05-23
add conf-mbedtls

andyli:
It's a new dependent of haxe (4.1.0).

### DIFF
--- a/packages/conf-mbedtls/conf-mbedtls.1/opam
+++ b/packages/conf-mbedtls/conf-mbedtls.1/opam
@@ -6,8 +6,7 @@ homepage: "https://tls.mbed.org/"
 bug-reports: "https://github.com/ARMmbed/mbedtls/issues"
 license: "Apache-2.0"
 build: [
-  ["ls" "/usr/include/mbedtls/version.h"] {os != "macos"}
-  ["brew" "ls" "--versions" "mbedtls"] {os = "macos"}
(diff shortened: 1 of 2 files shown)


packages/conf-mbedtls/conf-mbedtls.1/opam
```diff
@@ -0,0 +1,23 @@
+opam-version: "2.0"
+name: "conf-mbedtls"
+authors: "ARM Limited"
+maintainer: "Andy Li <andy@onthewings.net>"
+homepage: "https://tls.mbed.org/"
+bug-reports: "https://github.com/ARMmbed/mbedtls/issues"
+license: "Apache-2.0"
+build: [
+  ["ls" "/usr/include/mbedtls/version.h"] {os != "macos"}
+  ["brew" "ls" "--versions" "mbedtls"] {os = "macos"}
```
kit-ty-kate, line 10: I don't think this is a good solution to test whether this package is available or not.
Could you change this using a test file? For instance:
```
$ cat files/test.c
#include <mbedtls/version.h>
$ cat opam
[...]
build: ["cc" "test.c"]
```

andyli: Thanks for the review. Just updated the PR as suggested.

kit-ty-kate:
Thanks a lot!

PR #16508, opened 2020-05-23
Tighten the dependencies of irmin-unix.1.3.3

erikmd:
This small PR is a follow-up of a suggestion by @avsm - https://github.com/ocaml/opam/issues/4203#issuecomment-631493823

This patch makes the installation of irmin-unix.1.3.3 tractable (instead of raising a timeout due to a required time of around 10mn with opam's default solver options).
 
I know that `=` dependencies are not recommended, but:

* The constraint on ocaml `>= "4.01.0"` is kept as is
* The chosen dependencies versions correspond to the latest compatible versions
* Version 1.3.3 is not the bleeding-edge version of irmin-unix so "flexibility impact" is limited

### DIFF
--- a/packages/irmin-unix/irmin-unix.1.3.3/opam
+++ b/packages/irmin-unix/irmin-unix.1.3.3/opam
@@ -14,14 +14,14 @@ build: [
 ]
 depends: [
   "ocaml" {>= "4.01.0"}
-  "jbuilder" {>= "1.0+beta10"}
-  "irmin" {>= "1.3.0" & < "2.0.0"}
-  "irmin-mem" {>= "1.3.0" & < "2.0.0"}
-  "irmin-git" {>= "1.3.0" & < "2.0.0"}
-  "irmin-http" {>= "1.3.0" & < "2.0.0"}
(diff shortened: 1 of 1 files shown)


avsm:
Given the solver timeouts, this is fine for me for historical versions of irmin; cc @samoht

erikmd:
Hi, it seems `irmin-unix.1.3.3` installs fine (cf. e.g. [this 4.05.0 log](https://ci.ocaml.org/log/saved/docker-build-1607e26b52523aaa35ce948200a5ca33/facd4f4f8e10e728f188992f4875ea13e78d11a0))
while ` opam install -t irmin-unix.1.3.3` just fails because of a type error in the testsuite (cf. [this log](https://ci.ocaml.org/log/saved/docker-run-ed8ceb1858500edba9f7f50502b838aa/ac581cad2c9412e4061caaf4cce18f6e30abd692))
So do you agree to merge this PR? given that this test failure couldn't be fixed anyway.

avsm:
Thanks for the review. Merging this. Let us know if you have any other solver timeouts, @erikmd

PR #16513, opened 2020-05-24
Relocate distributions for Conjury and Orsetto.

jhwoodyatt:
- Atlassian has announced end of life for their Mercurial services.
- Moved the repositories to their Git services. All history preserved.


avsm:
Thanks!

PR #16516, opened 2020-05-25
Package template_engine.0.1

EmileTrotignon:
### `template_engine.0.1`
Simple template engine for OCaml
Simple template engine for OCaml

---
* Homepage: https://emiletrotignon.github.io/ocaml_template_engine/
* Source repo: git+https://github.com/EmileTrotignon/ocaml_template_engine.git
* Bug tracker: https://github.com/EmileTrotignon/ocaml_template_engine/issues

---

### DIFF
--- a/packages/template_engine/template_engine.0.1/opam
+++ b/packages/template_engine/template_engine.0.1/opam
@@ -1,0 +1,16 @@
+opam-version: "2.0"
+synopsis: "Simple template engine for OCaml"
+description: """
+Simple template engine for OCaml
+"""
+maintainer: "Emile Trotignon emile.trotignon@gmail.com"
+authors: "Emile Trotignon emile.trotignon@gmail.com"
+license: "MIT"
+homepage: "https://emiletrotignon.github.io/ocaml_template_engine/"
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
The archive is not defined so opam doesn't know where to get it. It seems that you haven't pushed the release tag. Could you push it then call opam-publish again?

Also Could you indent the `depends` fields this way, it's easier for us if we need to maintain/change things in the future?
```
depends: [
  "dep1"
  "dep2"
  ...
]
```

EmileTrotignon:
> The archive is not defined so opam doesn't know where to get it. It seems that you haven't pushed the release tag. Could you push it then call opam-publish again?
> 
> Also Could you indent the `depends` fields this way, it's easier for us if we need to maintain/change things in the future?
> 
> ```
> depends: [
>   "dep1"
>   "dep2"
>   ...
> ]
> ```

Since I commited this, I found out that what I did already existed in some way that makes this useless. I will therefore close this pull request, but thank you a lot for the advice, I will remember it if I make another contribution

PR #16518, opened 2020-05-25
zenon: conflict with Coq >= 8.9

rafoo:
At installation time, the zenon package runs Coq on a few files that rely on the `Implicit Arguments` syntax that has been deprecated in Coq 8.9. The following issue is about updating the files to support a more recent version of Coq: https://github.com/zenon-prover/zenon/issues/1.


kit-ty-kate:
Thanks a lot!

PR #16517, opened 2020-05-25
Package gen_js_api.1.0.6

mlasson:
### `gen_js_api.1.0.6`
Easy OCaml bindings for Javascript libraries
gen_js_api aims at simplifying the creation of OCaml bindings for
Javascript libraries.  Authors of bindings write OCaml signatures for
Javascript libraries and the tool generates the actual binding code
with a combination of implicit conventions and explicit annotations.

gen_js_api is to be used with the js_of_ocaml compiler.

---
* Homepage: https://github.com/LexiFi/gen_js_api
* Source repo: git+https://github.com/LexiFi/gen_js_api.git
* Bug tracker: https://github.com/LexiFi/gen_js_api/issues

---
:camel: Pull-request generated by opam-publish v2.0.2

PR #16519, opened 2020-05-25
Package ocaml-vdom.0.2

mlasson:
### `ocaml-vdom.0.2`
DOM and VDOM for OCaml
This package contains:
- OCaml bindings to DOM and other client-side Javascript APIs (using gen_js_api).
- An implementation of the Elm architecture, where the UI is specified as a functional "view" on the current state.

---
* Homepage: https://github.com/LexiFi/ocaml-vdom
* Source repo: git+https://github.com/LexiFi/ocaml-vdom.git
* Bug tracker: https://github.com/LexiFi/ocaml-vdom/issues

---

### DIFF
--- a/packages/ocaml-vdom/ocaml-vdom.0.2/opam
+++ b/packages/ocaml-vdom/ocaml-vdom.0.2/opam
@@ -1,0 +1,29 @@
+opam-version: "2.0"
+maintainer: "Alain Frisch <alain.frisch@lexifi.com>"
+authors: [
+  "Alain Frisch <alain.frisch@lexifi.com>"
+]
+homepage: "https://github.com/LexiFi/ocaml-vdom"
+bug-reports: "https://github.com/LexiFi/ocaml-vdom/issues"
+license: "MIT"
+dev-repo: "git+https://github.com/LexiFi/ocaml-vdom.git"
+build: [["dune" "build" "-p" name "-j" jobs]]
+depends: [
+  "ocaml" {>= "4.08.0"}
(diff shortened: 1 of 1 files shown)


mlasson:
This will not be mergeable before #16519 as gen_js_api 1.0.6 is a dependency to this package !

kit-ty-kate:
Could you return the fix upstream? (same thing as in https://github.com/ocaml/opam-repository/pull/16517)

Also, I have one extra question: Given that the package's main module is called `Vdom`, wouldn't it be simpler to just call this new package `vdom` (dropping the `ocaml-` prefix) instead? The opam-repository tried to avoid having the `ocaml-` prefix/suffix everywhere when possible and relevant. If you think the ocaml prefix is relevant here, I'll go ahead with the merge.

mlasson:
Your question makes a lot of sense. I'll have to check with the authors and come back to you as soon as possible.

alainfrisch:
I think it makes sense to keep the ocaml- prefix for the official name of the project (if only to make its name searchable). As for the name on OPAM, I've no opinion!

mlasson:
I guess that it would probably better to keep the official on opam to be consistent with the name and the github repository. So let's merge it like that if it is not too damageable for opam  !

mlasson:
What's the remaining error ?

kit-ty-kate:
Sorry for the time this took. Thanks!

mlasson:
> Sorry for the time this took. Thanks!
This is fine !
Thank YOU for taking care of that.

PR #16524, opened 2020-05-26
Package radare2.0.0.4

XVilka:
* Migrated to Dune 2.x
* Migrated to `Yojson.Basic.t` from `Yojson.Basic.json`
* Fixed the example in README and opam file


XVilka:
The failure is because Debian removed radare2 from stable "due to vulnerabilities":
image
See https://repology.org/project/radare2/versions

kit-ty-kate:
Looks good. Could you return the fix upstream? Thanks!

PR #16525, opened 2020-05-26
Add upper bound constraints to all reverse dependencies of Jane Street packages

xclerc:
Small follow-up to https://github.com/ocaml/opam-repository/pull/16484.


kit-ty-kate:
Thanks!

PR #16531, opened 2020-05-26
Package camelot.1.3.0

KeenWill:


packages/camelot/camelot.1.3.0/opam
```diff
@@ -0,0 +1,37 @@
+opam-version: "2.0"
+synopsis: "An OCaml Linter / Style Checker"
+maintainer: ["William Goeller <william@williamgoeller.com>"]
+authors: ["Vighnesh Vijay" "Daniel Like" "William Goeller"]
+license: "Apache License 2.0"
+homepage: "https://github.com/upenn-cis1xx/camelot"
+bug-reports: "https://github.com/upenn-cis1xx/camelot/issues"
+depends: [
+  "dune" {>= "2.5"}
+  "ocaml" {>= "4.10.0"}
```
avsm, line 10:
```suggestion
  "ocaml" {>= "4.10.0" & < "4.11.0"}
```
I notice you say that this is only expected to work with OCaml 4.10 in the README. You may want add this constraint to prevent future versions of OCaml from being picked up.

KeenWill:
Thanks for the suggestion!

kit-ty-kate:
Thanks a lot!

PR #16533, opened 2020-05-27
Spotlib.4.1.0

camlspotter:


kit-ty-kate:
Thanks! Could you return the fix upstream? (except for the version field removal, as it's only redundant in opam-repository)

PR #16535, opened 2020-05-27
fixes z3 compilation for z3.4.8.8

ivg:
This PR includes the https://github.com/Z3Prover/z3/pull/4468 patch so that we can benefit from using z3 without any hassle right now and do not wait until the next version is released. It also adds the `remove` stanza as well as enables the installation (and de-installation) of the z3 binary (so that we don't need to install the system-wide z3 in order to use the binary). See the upstream PR to z3 for the details as well as the commit messages.


avsm:
Thanks for such a detailed description and upstream PR. This looks like the right fix.

However, could you please modify this PR with a z3 package with a new version (e.g. z3.4.8.8-1) with your patch added, rather than modifying the original upstream? This allows opam users to fall back to the unpatched version via normal constraints if it has issues for any reason.  You can just duplicate the z3.4.8.8 metadata into a new directory to do this.

ivg:
Sounds, very reasonable, done!

avsm:
That "needs reporter action" label sure didn't last long :-)

packages/z3/z3.4.8.8-1/opam
```diff
@@ -0,0 +1,42 @@
...
+synopsis: "Z3 solver"
+extra-files: [
+  ["4468.patch" "md5=05a00fdd98aecba2198393bff8484efa"]
+]
+
+remove: [
+  ["ocamlfind" "remove" "z3"]
+  ["rm" "-f" "%{bin}%/z3"]
+]
+
```
kit-ty-kate, line 34: In your first comment you said you introduced the remove field. However this field should not be necessary since opam 2.0 unless you specifically require to remove something that has been installed at runtime (after installation), and it does not seem to be the case here. I'm going to remove it and this PR should be good to go

kit-ty-kate:
Thanks a lot!

PR #16544, opened 2020-05-29
async_unix.v0.13.1

xclerc:
This pull request makes `async_unix` compatible with OCaml 4.11,
and marks previous versions as not compatible.


xclerc:
(The Travis error is a timeout.)

kit-ty-kate:
Thanks a lot for this fast fix!

PR #16547, opened 2020-05-29
[new release] ppx_factory (0.1.1)

bbc2:
PPX to derive factories and default values

- Project page: <a href="https://github.com/cryptosense/ppx_factory">https://github.com/cryptosense/ppx_factory</a>
- Documentation: <a href="https://cryptosense.github.io/ppx_factory/doc">https://cryptosense.github.io/ppx_factory/doc</a>

##### CHANGES:

All notable changes to this project will be documented in this file.

The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this
project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).

## Unreleased

_No unreleased changes at the moment_

### Added

- Add support for OCaml 4.09 and 4.10

### Fixes

- Raise an error when used with an abstract type, except when invoked from `ocamldep`, @NathanReb
  [cryptosense/ppx_factory#20](https://github.com/cryptosense/ppx_factory/pull/20)

### Added

- Initial release

### DIFF
--- a/packages/ppx_factory/ppx_factory.0.1.1/opam
+++ b/packages/ppx_factory/ppx_factory.0.1.1/opam
@@ -1,0 +1,41 @@
+opam-version: "2.0"
+maintainer: "Cryptosense <opensource@cryptosense.com>"
+authors: [
+    "Cryptosense <opensource@cryptosense.com>"
+    "Nathan Rebours <nathan.p.rebours@gmail.com>"
+]
+homepage: "https://github.com/cryptosense/ppx_factory"
+bug-reports: "https://github.com/cryptosense/ppx_factory/issues"
+license: "BSD-2"
+dev-repo: "git+https://github.com/cryptosense/ppx_factory.git"
+doc: "https://cryptosense.github.io/ppx_factory/doc"
+build: [
+  [ "dune" "build" "-p" name "-j" jobs ]
+]
+run-test: [
+  [ "dune" "runtest" "-p" name "-j" jobs ]
+]
+depends: [
+  "dune"
+  "ocaml" {>= "4.07.0" & < "4.11.0"}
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
According to your `dune-project` file this package requires at least dune 1.1. Could you return the fix upstream?

I also have a question, I noticed you had the `ocaml < 4.11` constraint but this project compile perfectly fine with OCaml 4.11, was there another reason why the constraint was there in the first place?

bbc2:
> According to your dune-project file this package requires at least dune 1.1. Could you return the fix upstream?

Done. Thanks for reporting that.

> I also have a question, I noticed you had the ocaml < 4.11 constraint but this project compile perfectly fine with OCaml 4.11, was there another reason why the constraint was there in the first place?

I think the idea of the upper bound for a PPX comes from the fact that new versions of OCaml are very likely to break it.  To avoid likely failures for users, we maintain the constraint until we've checked that it seems to work.

If it works well with 4.11, I'll weaken the constraint. Is it easy to test that version in CI until it is released?

What should the weakened constraint be?  "< 4.12" or no upper bound?  I guess that both can work but maybe one is part of good practices or conventions that I'm not aware of?

kit-ty-kate:
unless you're sure it will break no upper bound constraint is better. Before each new release of the compiler we check that every packages work correctly with it and add the correct upper bound if it does not.

Depends which CI you have but with travis you just need to add:
```
- OCAML_VERSION=4.11.0+trunk OCAML_BETA=enable
```
to your build matrix. For instance: https://github.com/ocaml-community/utop/commit/aa832e280cce7c5a7cb92e65609285932650bf49

Thanks a lot!

PR #16548, opened 2020-05-29
BAP 2.1.0 release

gitoleg:
The Carnegie Mellon University Binary Analysis Platform ([CMU BAP][3]) is a suite of utilities and libraries that enables analysis of programs that are represented as machine code (aka binaries). CMU BAP is written in OCaml and uses plugin-based architecture to enable extensibility. We also have a domain-specific language, called Primus Lisp, that we use to write analysis, specify verification conditions, interact with the built-in SMT solver, and model the semantics of machine instructions and functions.

The 2.1.0 Release is very rich in [new features][1] but the most prominent addition is the new [symbolic executor][2] mode for the Primus framework. We also significantly updated the Primus framework, integrated it with our new Knowledge Base, which was introduced in the BAP 2.0 release; we made our interpreter much faster; we added the systems and components facilities, inspired by Common Lisp; and we implemented a gradual type checker for Primus Lisp with type inference. We also added an ability to represent machine instructions as intrinsic functions so now it is possible to express their semantics using Primus Lisp since we added IEEE754 primitives to the Lisp interpreter.

As usual, we upgraded BAP to the newer versions of the Core library and OCaml (we now support OCaml versions from 4.07 to 4.09). We also significantly improved our build times and added an optional omake backend, which we are using in-house.

From the user perspective, one of the key features of BAP as an analysis platform is that you can run BAP on binaries that you can't run otherwise, either because they need special hardware or software, or need to interact with the outside world. In the past couple of months, we have run BAP on various firmware and found numerous zero-day vulnerabilities, particular, we found two vulnerabilities in VxWorks operating system that runs on, potentially, billions of devices including mission-critical and military appliances.

[1]: https://github.com/BinaryAnalysisPlatform/bap/releases/tag/v2.1.0
[2]: https://github.com/BinaryAnalysisPlatform/bap/pull/1105
[3]: https://github.com/BinaryAnalysisPlatform/bap


ivg:
I've fixes camelus warnings (the synopsis capitalization and dots at the end), also made conf-bap-llvm a little bit more portable (it was using String.equal now it is using the plain polymorphic compare), I also backported this change to the older versions. 

Looked carefully at other failures, didn't find anything suspicious on our side. Below is the summary of failures. 

Summary of failures:
1) Old conf-bap-llvm that I have retrospectively fixed to enable compatibility with venerable versions of OCaml are not buildable because there is no appropriate versions of LLVM in the CI. (Versions 1.5 and 1.6 are passing the tests)
2) A few packages are failing for CI reasons, at least I don't [see any errors][1] relevant to the installation process, e.g.,
```
Processing 144/218: [bitvec-order: ./configure] [lambda-term: dune build] [ppx_bench: dune build] [ppx_fail: dune build]
Error removing intermediate container 8df78428b4f9: No such container: 8df78428b4f9ff5d026d4b84d4132b4cfba3883d582b4f457b1bd0b7e98830ad
The command '/bin/sh -c opam exec -- opam-ci-install bap-abi.2.1.0' returned a non-zero code: 137
[2020-06-01 18:36.03] Failed: Command "sh" "-c" "docker build --label com.docker.datakit.digest=a6318314467c225a7606090388514483 --label com.docker.datakit.builton=1591035295.77 --rm --force-rm - < /tmp/tmp-d2c04c4/Dockerfile" exited with status 137
```
3) As usual there are a lot of connectivity issues, e.g.,
```
ERROR] The sources of the following couldn't be obtained, aborting:
          - stringext.1.6.0:
            https://github.com/rgrinberg/stringext/releases/download/1.6.0/stringext-1.6.0.tbz (Curl failed: "/usr/bin/curl --write-out %{http_code}\\n --retry 3 --retry-delay 2 --user-agent opam/2.0.7 -L -o /home/opam/.opam/4.07/.opam-switch/sources/stringext.1.6.0/stringext-1.6.0.tbz.part https://github.com/rgrinberg/stringext/releases/download/1.6.0/stringext-1.6.0.tbz" exited with code 52)
'opam install --verbose --yes --jobs=72 bap-api.2.1.0' failed.
```
I also wonder is `--jobs=72` a good idea?




[1]: https://ci.ocaml.org/log/saved/docker-build-a6318314467c225a7606090388514483/fa77fb683cf92788c45bfba86d46ae83e35d0e0e
[2]: https://ci.ocaml.org/log/saved/docker-build-8ff2cc1bebf0c48df3ef2ba4e924347d/9ccc8bc21d98f24398a903a7b9d7d0f1cb10cb87

ivg:
@kit-ty-kate, can you guys please merge this PR, we have to finish the release and do public announces and we're currently blocked here. There are no failures relevant to BAP (and as you know we have our own pretty descent CI system) so I think that we can move forward on it.

kit-ty-kate:
Sorry for the wait.

For the future, several things can be improved:
* `{= "2.1.0"}` can be changed to `{= version}` which is usually more easy to maintain
* The `remove` fields can probably be removed unless the remove section also deals with extra config files written after the installation. Since opam 2.0 every files installed by a package are recorded and automatically removed when uninstalled.

Thanks!

PR #16549, opened 2020-05-29
Package camlp5.rel712

chetmurthy:
### `camlp5.rel712`
Preprocessor-pretty-printer of OCaml
Camlp5 is a preprocessor and pretty-printer for OCaml programs. It also provides parsing and printing tools.

As a preprocessor, it allows to:

extend the syntax of OCaml,
redefine the whole syntax of the language.
As a pretty printer, it allows to:

display OCaml programs in an elegant way,
convert from one syntax to another,
check the results of syntax extensions.
Camlp5 also provides some parsing and pretty printing tools:

extensible grammars
extensible printers
stream parsers and lexers
pretty print module
It works as a shell command and can also be used in the OCaml toplevel.

---
* Homepage: https://camlp5.github.io
* Source repo: git+https://github.com/camlp5/camlp5.git
* Bug tracker: https://github.com/camlp5/camlp5/issues

---


kit-ty-kate:
I fixed the version number that didn't correspond to the previous ones and removed the `remove` field as it is not necessary since opam 2.0 (could you return the fix to the opam file upstream?)

kit-ty-kate:
Thanks a lot!

chetmurthy:
Oh, that's my mistake.  I'm sorry, I thought I'd removed that "remove"
stanza.  I did so now, and will make a note to check that, as well as the
version-number, for next time.  Sorry sorry!

PR #16555, opened 2020-05-31
fix for ppx_meta_conv.4.0.0

camlspotter:


kit-ty-kate:
Thanks. This change is quite weird from gitlab's point of vu. The only difference between the two archives are the unix modification time for the files (i.e. what day the archive has been created). Do you know if they re-create all archive once a month or something? This is the first time I see such a thing.

PR #16556, opened 2020-05-31
[soupault] Pin lambdasoup dependency to 0.7.0 due to backwards-incompatile changes in 0.7.1

dmbaturin:
lambdasoup, unfortunately, made a stealthy backwards-incompatible behaviour change (it now automatically adds an HTML5 doctype, while older versions never produced a doctype so code that used the formatting function had to compensate for it—now that will produce a duplicate doctype).

I will work with @aantron to work out something.

This issue also affects minima-theme.


kit-ty-kate:
Thanks

PR #16557, opened 2020-05-31
ppx_meta_conv.4.0.1

camlspotter:


camlspotter:
revdep issue of msgpack.1.3.0.  It does not compile with ppx_meta_conv.4.0.0 already.  It only supports older versions of ppx_meta_conv I guess.

kit-ty-kate:
Thanks!

PR #16559, opened 2020-06-01
haxe 4.1.1

andyli:


kit-ty-kate:
Two constraints were missing – for `ptmap` (according to build failures on 4.11) and `dune` (according to your `dune-project` file). Also, almost all dependencies were also tagged as `{build}` dependencies which makes it vulnerable to being out of sync regarding eventual security fixes or things of the sort.

Could you return the fixes in the opam file upstream?

Thanks a lot!

PR #16561, opened 2020-06-02
Release Caqti 1.2.3, core and drivers updated.

paurkedal:
Details from the change log:

- Silence PostgreSQL notifications which were printed to stderr.  They can
  be re-enabled with a parameter on the connection URI (paurkedal/ocaml-caqti#33).
- Improve details in exceptions in two places.
- Fix info about connection caching for sqlite in README (paurkedal/ocaml-caqti#35, dgllghr).


paurkedal:
I adjusted a few dependencies to fix revdeps tests. Now, hopefully there will only be the expected failures related to the availability of the MariaDB client library.

paurkedal:
I can only see expected failures now from the CI.

kit-ty-kate:
Thanks!

PR #16563, opened 2020-06-02
[new release] dns, dns-stub, dns-certify, dns-mirage, dns-resolver, dns-client, dns-server, dns-tsig and dns-cli (4.6.0)

hannesm:
CHANGES:

* dns: bugfix for name compression when encoding names at offset > (2 ^ 14) - 1
  (mirage/ocaml-dns#225 @hannesm)
* dns: allow unknown DNSKEY algorithm, TLSA certificate usage, selector,
  matching type, SSHFP algorithm and typ. This makes the DNS library
  future-proof for when new values are assigned (mirage/ocaml-dns#228 @hannesm)
* dns: enforce a max_rdata_length for all resource records. This ensures that
  when a resource record is loaded into the server, it can be extracted via a
  DNS query and transferred via IXFR/AXFR
  (mirage/ocaml-dns#230 @hannesm, reported in mirage/ocaml-dns#229 via mirage/ocaml-dns#225)
* AXFR: encode and decode support for AXFR transfers spanning multiple messages
  (mirage/ocaml-dns#225 @hannesm)
* client: do not initialize the Mirage_crypto_rng in the library, initialize
  the RNG in applications (mirage/ocaml-dns#227 @hannesm)
* certify: provide cert_matches_csr function and use it (cleans up partial
  ad-hoc matches which did not verify that all hostnames of the CSR are present
  in the certificate) (mirage/ocaml-dns#226 @hannesm, reported in mirage/ocaml-dns#224)

### DIFF
--- a/packages/dns-certify/dns-certify.4.6.0/opam
+++ b/packages/dns-certify/dns-certify.4.6.0/opam
@@ -1,0 +1,46 @@
+opam-version: "2.0"
+maintainer: "team AT robur dot io"
+authors: ["Hannes Mehnert <hannes@mehnert.org>"]
+homepage: "https://github.com/mirage/ocaml-dns"
+doc: "https://mirage.github.io/ocaml-dns/"
+dev-repo: "git+https://github.com/mirage/ocaml-dns.git"
+bug-reports: "https://github.com/mirage/ocaml-dns/issues"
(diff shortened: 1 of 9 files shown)


kit-ty-kate:
Thanks a lot!

PR #16567, opened 2020-06-03
Package lem.2020-06-03

CuiCui66:
### `lem.2020-06-03`
Lem is a tool for lightweight executable mathematics
Lem is a tool for lightweight executable mathematics, for writing,
managing, and publishing large-scale portable semantic definitions,
with export to LaTeX, executable code (currently OCaml) and
interactive theorem provers (currently Coq, HOL4, and Isabelle/HOL).

It is also intended as an intermediate language for generating
definitions from domain-specific tools, and for porting definitions
between interactive theorem proving systems.

---
* Homepage: http://www.cl.cam.ac.uk/~pes20/lem/
* Source repo: git+https://github.com/rems-project/lem.git
* Bug tracker: https://github.com/rems-project/lem/issues

---

### DIFF
--- a/packages/lem/lem.2020-06-03/opam
+++ b/packages/lem/lem.2020-06-03/opam
@@ -1,0 +1,50 @@
+opam-version: "2.0"
+maintainer: "Lem Devs <cl-lem-dev@lists.cam.ac.uk>"
+authors: [
+  "Dominic Mulligan"
+  "Francesco Zappa Nardelli"
+  "Gabriel Kerneis"
+  "Kathy Gray"
+  "Peter Boehm"
+  "Peter Sewell"
+  "Scott Owens"
+  "Thomas Tuerk"
+  "Brian Campbell"
+  "Shaked Flur"
+  "Thomas Bauereiss"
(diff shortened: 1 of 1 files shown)


CuiCui66:
Do not merge

Actually I discovered some tests, (I'm just packaging, I did not write the software) and they don't pass, So I need to make them pass and then add them to the the target "with-test" in the opam file.

Should I close the PR and open another one when this is done, or should I leave it open?

kit-ty-kate:
if the version number gets change it might be easier to close it, otherwise you can let it open as opam-publish should be able to update the PR.

CuiCui66:
After discussing with an author, this was a legacy test suite that was not updated for several years and is not valid anymore. It's normal they fail. They are not even mentioned in the README or anything else. The main test suite is already invoked by the main `make` invocation and pass. We'll keep them in case someone updates them later but they will not be run by the opam `with-test` variable as they are not correct anymore.

@kit-ty-kate In the end, we can merge the PR as-is

avsm:

PR #16572, opened 2020-06-04
[test] Split packages by two-letter prefix

AltGr:
This is a test with two goals in mind:
- see how it affects performance
- see how tools cope: opam itself should handle it naturally, but other tools (opam publish, opam2web, conex...) might need adjustment.

Basically, I just wondered if it would improve performance in some aspects, and since opam should already support it...


AltGr:
Sorry to all CI bots out there :grin:

kit-ty-kate:
By some luck the CI wasn't triggered :D 

What kind of performance improvement are we talking about here? Solver?

AltGr:
Just file scanning perf... like `opam update` or stuff like that

AltGr:
Either the CI was clever enough to detect that actually no package was modified, or it failed to find the packages at their new location and therefore decided that there were no packages to check :grin:

avsm:
A number of tools do assume the 2-level hierarchy right now, although technically anything under packages should just work.  I'll close this PR though; let's not hold open PRs that cause potential rebuilds of every package, as there is a large PR backlog at the moment.

PR #16574, opened 2020-06-04
Package linksem.0.7

CuiCui66:
### `linksem.0.7`
A formalisation of the core ELF file format written in Lem
A formalisation of the core ELF file format written in Lem.
ELF is the de facto standard executable and linkable file format
on Linux and related systems.
This formalisation has been tested against approximately 5,000 ELF binaries
found "in the wild" on various different platforms.

---
* Homepage: https://github.com/rems-project/linksem
* Source repo: git+https://github.com/rems-project/linksem.git
* Bug tracker: https://github.com/rems-project/linksem/issues

---

### DIFF
--- a/packages/linksem/linksem.0.7/opam
+++ b/packages/linksem/linksem.0.7/opam
@@ -7,7 +7,6 @@ license: "BSD-2"
 dev-repo: "git+https://github.com/rems-project/linksem.git"
 build: [make]
 install: [make "install"]
-remove: [make "uninstall"]
 depends: [
   "ocaml" {>= "4.06.1"}
   "ocamlfind" {build}
--- a/packages/linksem/linksem.0.7/opam
+++ b/packages/linksem/linksem.0.7/opam
@@ -1,0 +1,32 @@
+opam-version: "2.0"
+maintainer: "Linksem Devs <cl-linksem-dev@lists.cam.ac.uk>"
+authors: ["Stephen Kell" "Dominic Mulligan" "Peter Sewell"]
+homepage: "https://github.com/rems-project/linksem"
+bug-reports: "https://github.com/rems-project/linksem/issues"
+license: "BSD-2"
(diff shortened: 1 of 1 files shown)


CuiCui66:
This depends on #16567

avsm:
Could you please rebase/merge this against the opam-repository master branch, so that CI picks up #16567?

CuiCui66:
@avsm So I rebased and now the CI failed specifically failed on fedora and I have no idea why. We have someone who had a fedora and it worked. Especially it seems to be an issue of ocamlfind not finding a module and I don't see why this kind of thing would be distribution-dependent

avsm:
It's possibly because you need to also depend on the `num` package. That's been pulled out since OCaml 4.06 into a separate package.

CuiCui66:
`linksem` has a dependency `lem` that depends on `num` and it does not use `num` without going through `lem`, so I think we should be fine on that side.

However, I think we found it out: It's because the `find` command is not installed on your fedora install so our build system for `lem` do not find it's own file and so don't install them (it does a find over all cmi, cmx and mli), then linksem cannot link against it. I'll add `conf-findutils` to our dependencies list

avsm:
Yes, that'll work. Fedora no longer includes `findutils` as part of its base, so that conf package addition will let `opam depext -i linksem` work on Fedora as well.

CuiCui66:
We now require #16594, for the bug to be fixed

kit-ty-kate:
Thanks! Sorry for the wait

kit-ty-kate:
Could you also remove the `remove` field from the opam file upstream? This field is usually unnecessary since opam 2.0 (unless it removes files created after installation as well)

CuiCui66:
@kit-ty-kate I've removed the `remove` field upstream. It was just there from opam 1 legacy

PR #16577, opened 2020-06-05
0install-solver: require dune 2.5 to fix installation

talex5:
https://github.com/ocaml/dune/issues/3252 causes 0install-solver to fail to install, with:

    # File "src/tests/dune", line 3, characters 10-23:
    # 3 |   (deps %{bin:0install})
    #               ^^^^^^^^^^^^^
    # Error: Program 0install not found in the tree or in PATH


kit-ty-kate:
Thanks a lot!

PR #16579, opened 2020-06-05
bap-primus-systems: fix incorrect name in opam file

avsm:
it needs to match the name of the directory around it.
cc @ivg


ivg:
Ooops! Thanks for noticing that :)

kit-ty-kate:
Thanks!

PR #16582, opened 2020-06-05
Package sedlex.2.2

toots:
### `sedlex.2.2`
An OCaml lexer generator for Unicode
sedlex is a lexer generator for OCaml. It is similar to ocamllex, but supports
Unicode. Unlike ocamllex, sedlex allows lexer specifications within regular
OCaml source files. Lexing specific constructs are provided via a ppx syntax
extension.

---
* Homepage: https://github.com/ocaml-community/sedlex
* Source repo: git+https://github.com/ocaml-community/sedlex.git
* Bug tracker: https://github.com/ocaml-community/sedlex/issues

---

### DIFF
--- a/packages/sedlex/sedlex.2.2/opam
+++ b/packages/sedlex/sedlex.2.2/opam
@@ -1,0 +1,41 @@
+opam-version: "2.0"
+synopsis: "An OCaml lexer generator for Unicode"
+description: "
+sedlex is a lexer generator for OCaml. It is similar to ocamllex, but supports
+Unicode. Unlike ocamllex, sedlex allows lexer specifications within regular
+OCaml source files. Lexing specific constructs are provided via a ppx syntax
+extension.
+"
+license: "MIT"
+doc: "https://ocaml-community.github.io/sedlex/index.html"
+maintainer: "Alain Frisch <alain.frisch@lexifi.com>"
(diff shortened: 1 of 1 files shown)


avsm:
Thanks for this! Do previous releases of sedlex need to have an upper bound added of `< "4.08.0"`, since this release adds support for that? Otherwise looks good to merge.

toots:
Good call, adding it.

kit-ty-kate:
I'm not sure the upper-bounds for the previous releases are necessary. The changelog says this version adds "support for OCaml 4.08", however looking at the actual commits, all I can find is this version adding "support for OCaml 4.08 (specific) syntax" (https://github.com/ocaml-community/sedlex/commit/6c15cc420b46d13c2f725bd02543642ed45698d2), previous versions and its reverse dependencies were compiling fine with OCaml 4.08 and above.

Unless I'm missing something from how sedlex works I think it is to the charge of the packages depending on sedlex to get the right lower-bound constraint if they really use OCaml 4.08 syntaxes.

toots:
You are totally right @kit-ty-kate support for `4.08` is for the AST support. I've removed my latest commit.

kit-ty-kate:
Thanks a lot!

PR #16584, opened 2020-06-05
package Coq 8.11.2

RalfJung:


avsm:
Looks good, thanks! CentOS failing due to the gtk binding packages not being found, but that's occurred before, so not a blocker.

PR #16590, opened 2020-06-06
9 packages from savonet/ocaml-posix at 2.0.0

toots:
This pull-request introduces external posix bindings based on `ocaml-ctypes`. The plan is to remove/deprecate `ocaml-ctypes`'s internal `PosixTypes` in favor of this module's `posix-types`. See: https://github.com/ocamllabs/ocaml-ctypes/issues/624 and https://github.com/ocamllabs/ocaml-ctypes/pull/648

The bindings for these modules are light-weight and follow the pattern detailed here: https://medium.com/@romain.beauxis/advanced-c-binding-using-ocaml-ctypes-and-dune-cc3f4cbab302

In particular, they allow proper export of both low-level ctypes API (struct, etc) and high-level bindings suitable for general consumption.

This pull-request contains the following packages:
-`posix-base.2.0.0`: Base module for the posix bindings
-`posix-bindings.2.0.0`: POSIX bindings
-`posix-getopt.2.0.0`: Bindings for posix getopt/getopt_long
-`posix-signal.2.0.0`: Bindings for the types defined in <signal.h>
-`posix-socket.2.0.0`: Bindings for posix sockets
-`posix-socket-unix.2.0.0`: Bindings for posix sockets
-`posix-time2.2.0.0`: Bindings for posix time functions
-`posix-types.2.0.0`: Bindings for the types defined in <sys/types.h>
-`posix-uname.2.0.0`: Bindings for posix uname

---
* Homepage: https://github.com/savonet/ocaml-posix
* Source repo: git+https://github.com/savonet/ocaml-posix.git
* Bug tracker: https://github.com/savonet/ocaml-posix/issues

---

### DIFF
--- a/packages/posix-base/posix-base.2.0.0/opam
+++ b/packages/posix-base/posix-base.2.0.0/opam
@@ -1,0 +1,35 @@
+opam-version: "2.0"
+synopsis: "Base module for the posix bindings"
+description: "posix-base provides base tools for the posix binding modules."
+maintainer: ["romain.beauxis@gmail.com"]
+authors: ["Romain Beauxis"]
+license: "MIT"
+homepage: "https://github.com/savonet/ocaml-posix"
+bug-reports: "https://github.com/savonet/ocaml-posix/issues"
+depends: [
+  "dune" {> "2.5"}
+  "integers"
+  "ctypes"
+]
+build: [
+  ["dune" "subst"] {pinned}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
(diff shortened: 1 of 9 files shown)


toots:
It looks like tests are failing on `alpine`, which is not entirely surprising since the libc is a complete rewrite there. I'll have a look at it but I don't think that this should prevent the packages from being accepted.

kit-ty-kate:
I've opened https://github.com/savonet/ocaml-posix/issues/2 to track the issue. In the meantime this should be good to go.

Also not that `os = "bsd"` is not correct. The `os` variable will have the value `freebsd`, `openbsd` or `netbsd` depending on which one it is. I'll let as is for now but it's also something to keep in mind.

Thanks! Sorry for the wait

toots:
Thanks!

PR #16592, opened 2020-06-07
Release caqti-driver-postgresql.1.2.4.

paurkedal:
I forgot to merge a patch promised for the 1.2.3 release last weekend, so this is an update to the affected package:

- Switch to TEXT format for PostgreSQL populate implementation.


kit-ty-kate:
Thanks

PR #16595, opened 2020-06-08
Release camelot.1.4.2

KeenWill:


kit-ty-kate:
Thanks! Could your return the fix to the opam file upstream?

PR #16601, opened 2020-06-09
archetype requires at least digestif 0.7.2 (with dune variants)

kit-ty-kate:
cc @guillaumeduhamel could you return the fix in your opam file upstream?


guillaumeduhamel:
> cc @guillaumeduhamel could you return the fix in your opam file upstream?

done, https://github.com/edukera/archetype-lang/commit/e0d934b893374776121d2cb55d7abe33d63ee498

PR #16605, opened 2020-06-09
Package yuujinchou.0.9

favonia:
### `yuujinchou.0.9`
Generic name manipulation combinators
This library implements a generic library for selecting names. It intends to facilitate the implementation of the "import" statement or any feature that allows users to select a group of names by patterns.

---
* Homepage: https://github.com/favonia/yuujinchou
* Source repo: git+https://github.com/favonia/yuujinchou.git
* Bug tracker: https://github.com/favonia/yuujinchou/issues

---

### DIFF
--- a/packages/yuujinchou/yuujinchou.0.9/opam
+++ b/packages/yuujinchou/yuujinchou.0.9/opam
@@ -1,0 +1,27 @@
+opam-version: "2.0"
+synopsis: "Generic name manipulation combinators"
+description: """
+This library implements a generic library for selecting names. It intends to facilitate the implementation of the "import" statement or any feature that allows users to select a group of names by patterns.
+"""
+maintainer: "favonia <favonia@gmail.com>"
+authors: "favonia <favonia@gmail.com>"
+license: "Apache 2.0"
+homepage: "https://github.com/favonia/yuujinchou"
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Your `dune-project` file indicates that this package requires at least dune 2.5.
Also you usage of `ppx_deriving` does not seem to involve any of its internals to I'm assuming the upper-bound should be safe to remove (if I'm wrong please say so and I'll put it again)
Also, dune should always be called with the `-p %{name}%` arguments. Otherwise other sub-packages might be compiled or tested and files would be compiled dev only (warnings as errors) parameters.

Could you return the changes in your opam file in the source repository upstream?

kit-ty-kate:
Thanks!

favonia:
Thank you for your review. The current opam file in the archive `yuujinchou-0.9.tar.gz` is still "wrong." Should I worry about it? (I assume moving the tag would be bad because that would also change the hash values.) Thank you again for helping beginners like me.

kit-ty-kate:
by "wrong" you mean "does not have the opam file fixed" as does opam-repository? If yes, it does not matter, the opam file inside the archive are not used by opam unless you build it by hand.

PR #16607, opened 2020-06-10
added scaml.1.2.0pre7

camlspotter:


kit-ty-kate:
Sorry for taking so much time. I've fixed a few issues, could you return them upstream? (notably the version field was not equal to the version you're pushing here and dune is not compatible with the `{build}` tag).

There is also an issue concerning the version number. The `pre7` prefix isn't right as `1.2.0` isn't superior to `1.2.0pre7`. Opam will take `1.2.0pre7` over `1.2.0` and I don't think that's what you want. See https://opam.ocaml.org/doc/Manual.html#Package-Formulas. If that's not what you want, could you rename the version to `1.2.0~pre7` instead? (you don't have to change the archive, just the directory added in this PR)

camlspotter:
Thanks, I will fix issues.

camlspotter:
@kit-ty-kate Could you check the fix?  Thanks!

avsm:

PR #16610, opened 2020-06-11
obandit: Fixing tarball URLs for old versions.

freuk:
This commit fixes the release tarball URLs for older
versions of the "obandit" package. Github tarballs are used in
place of self-hosted ones, which should be more reliable.


avsm:
Thanks for this!

PR #16612, opened 2020-06-11
mdx: add bound on cmdliner

emillon:
Versions 1.5.0 and 1.6.0 use `Arg.conv` which has been added in 1.0.0.

See realworldocaml/mdx#268.


kit-ty-kate:
Thanks!

PR #16619, opened 2020-06-12
[RFC] Add opam config to set OCAMLPARAM per switch

jberdine:
It would be helpful to have an easier way to set up an opam switch where everything is compiled with flambda -O3 than manually setting OCAMLPARAM, etc. While this does not achieve anything that cannot be done with careful setting of OCAMLPARAM, it seems clear that doing so is not broadly known or common practice, and so flambda would have higher impact across the community with a switch such as this.  See https://discuss.ocaml.org/t/ann-bap-2-1-0-release/5906 for discussion and some benchmarks that indicate this is useful.

As a test, I checked the file sizes of e.g. `lib/base/base__Map.cmx` with this compiler variant vs the existing 4.10.0+flambda one, which indicates that -O3 has taken effect and led to larger files.

~~On the other hand, the same does not appear to hold for the stdlib, as the sizes of the `lib/ocaml/stdlib*.cmx` files are exactly the same. This is unexpected, right? Does anyone know how to adjust the build so that the stdlib is built with -O3? I expected the `build-env` clause to be enough.~~ Resolved, see comment below.

Pinging @ivg who might be interested.

For convenience, here is the diff with the existing flambda variant:
```diff
$ diff ocaml-variants.4.10.0+flambda/opam ocaml-variants.4.10.0+flambda+opt
2c2
< synopsis: "Official release 4.10.0, with flambda activated"
---
> synopsis: "Official release 4.10.0, with flambda -O3 activated"
16c16,20
< setenv: CAML_LD_LIBRARY_PATH = "%{lib}%/stublibs"
---
> setenv: [
>  [OCAMLPARAM = "_,O3=1"]
>  [CAML_LD_LIBRARY_PATH = "%{lib}%/stublibs"]
> ]
> build-env: [OCAMLPARAM = "_,O3=1"]
```

### DIFF
--- a/packages/ocaml-variants/ocaml-variants.4.12.0+trunk+flambda/opam
+++ b/packages/ocaml-variants/ocaml-variants.4.12.0+trunk+flambda/opam
@@ -10,7 +10,10 @@ depends: [
 ]
 conflict-class: "ocaml-core-compiler"
 flags: compiler
-setenv: CAML_LD_LIBRARY_PATH = "%{lib}%/stublibs"
+setenv: [
+  [OCAMLPARAM = "%{ocamlparam}%"]
+  [CAML_LD_LIBRARY_PATH = "%{lib}%/stublibs"]
+]
 build: [
   [
     "./configure"


jberdine:
Update: the situation with the stdlib is ok and understood: `Makefile.common` contains
```
ifeq "$(FLAMBDA)" "true"
OPTCOMPFLAGS += -O3
endif
```
so the stdlib is compiled with -O3 with or without this PR.

avsm:
Is there any reason not to just make -O3 the default with the flambda variant, instead of adding a new one?  Most people using `+flambda` do so with the intention of their binaries running as fast as possible.

*Your friendly local opam-repository maintainer, trying to bring the proliferation of compiler variants under control.*

jberdine:
That's a good point.

I didn't propose to change the existing switch definition because just jamming things into an environment variable is a pretty blunt approach. I don't know whether this will break or significantly inconvenience any packages, and I don't really know how to test that. There is also the point that if the globally best default is -O3, then probably that default should be set in the compiler itself instead of by the package manager. But I probably have the minority view here since the current state is that e.g. dune does not keep the compiler defaults either.

Another question is whether, with the switch approach, it should set `OCAMLPARAM = "_,O3=1"` or `OCAMLPARAM = "O3=1,_"` or something else. I'm unsure of how to answer that.

Is there a way to make a switch like this available widely for the community to test / get some experience with without expanding the set of compiler variants? Perhaps this ought to go with the switches for PRs.

ivg:
A possible workflow that I would imagine, is to rely on the opam 2.0 configuration variables for that, e.g.,
```
opam config set optimization-level 3
opam install bap
```

This will give a level of abstraction over those drastic `_,-O3=1` and give some level of documentation,
```
opam config list | grep optimization
optimization-level 3
```

And yes, this configuration option is switch-local. A small drawback is that you have to set it after you install the compiler (so the compiler itself is not built with O3, but the standard library is built with O3, so I think it is ok). 

The only problem is that it is not obvious is it possible to it work, maybe adding 
```
setenv {
  OCAMLPARAM = "_,O%{optimization-level}%=1"
}
```
will work, I don't know, but most likely it shouldn't. I didn't test.

XVilka:
It should be added also to the other >=4.10 flambda switches as well, including 4.11 alpha and beta releases, I think.

Blaisorblade:
Uh, I didn’t know that trick either, and we have the same problem in Coq, see https://github.com/ocaml/opam-repository/pull/16779. I’ll look at your proposals.

Blaisorblade:
@avsm 
> Your friendly local opam-repository maintainer, trying to bring the proliferation of compiler variants under control.

Package flags, like in Ubuntu, would be nice to avoid that. Copy-pasting opam files and adding +foo is a very poor alternative; I‘d be interested in at least 3 such flags for coq, but I don’t plan to write by hand 8 opam files for each Coq release.

Blaisorblade:
> I don't know whether this will break or significantly inconvenience any packages, and I don't really know how to test that. 

I know one example: 
flambda seems to exacerbate some performance problems in Coq, especially on some platforms, unless you disable native compute *at compile-time*. So flambda and native compute are two of the flags I’d add to Coq packages.

Blaisorblade:
> And yes, this configuration option is switch-local. A small drawback is that you have to set it after you install the compiler (so the compiler itself is not built with O3, but the standard library is built with O3, so I think it is ok).

But there's `set-global`! You can
```
opam config set-global optimization-level 3
```
So, even if you want this just for one switch, you can set it globally, create your switch, and then if you want unset it globally and set it only locally.* Using mutable state for that is not ideal, but it is much better than the current scenario.

*You can also set it globally and locally, but not independently it seems?

dra27:
Re https://github.com/ocaml/opam-repository/pull/16619#issuecomment-644727604, note that building `ocamlopt` _itself_ with `-O3` is not with its own risks at the moment (see https://github.com/ocaml/ocaml/issues/9839)

jberdine:
Now that the PR enabling the compiler to treat env vars set to empty the same as unset has been merged, I retargeted this PR to ocaml-variants.4.12.0+trunk+flambda.

For 4.12, as far as I understand this change could be included in all the official switch opam files, with no additional variants, enabling users who wish to set default values of OCAMLPARAM for an entire switch.

As a first test, I did the following:

Set up a standard 4.12.0+trunk+flambda switch with dune installed:
```
$ opam switch create --empty param-test
$ opam install ocaml-variants.4.12.0+trunk+flambda dune
```
And set up a 4.12.0+trunk+flambda switch where OCAMLPARAM is set to `O3=1,_`:
```
$ opam repo add local --kind=local ~/src/opam-repository
$ opam switch create --empty param_test --repositories=local,beta
$ opam config set ocamlparam "O3=1,_"
$ opam install ocaml-variants.4.12.0+trunk+flambda dune
```
(Note the local repo business is to test my clone of opam-repository, and beta is needed since this targets a `+trunk` variant.)

The test is first to check that the switch creation goes through without error, which it did, and second that the OCAMLPARAM setting took effect. This second part is very coarsely done by comparing the size of the compiled dune binaries:
```
$ ls -l ~/.opam/param-test/bin/dune ~/.opam/param_test/bin/dune
```
On my test machine, this resulted in a difference of 16 vs 23 MB, which is consistent with -O3 inlining more leading to larger binaries.

Would it make sense to include this in the opam files of switches for the upcoming test cycle for 4.12?

dra27:
I'm not sure that ocaml/ocaml#9854 is sufficient to do this in opam-repository... when you switch from a switch which has set `OCAMLPARAM`, `opam env` will remove the setting, but not unset `OCAMLPARAM` so other switches will get warnings/errors about the empty `OCAMLPARAM`.

I think if one switch is going to set it, all switches needed to put _something_ in it?

jberdine:
Oh, wow. I was working under the assumption that `opam switch foo; opam switch bar; opam switch foo` was equivalent to `opam switch foo`, but if opam doesn't implement that, then these setenv clauses are risky. For whatever reason, I was thinking of these setenv settings as being applied to the child processes of opam, rather than changing the global state.

So it seems that the options are:

1. Apply the diff of this PR to *all* of packages/ocaml-variants/*/opam and hope that no one mixes one of these switches with one specified elsewhere. This approach would require an additional change in opam to treat e.g. `setenv: OCAMLPARAM = "%{ocamlparam}%"` as `unsetenv OCAMLPARAM` when `%{ocamlparam}%` is empty. It would seem wrong for such a change to be specific to OCAMLPARAM, but I don't know if there are other uses of opam set env variables where the distinction between set-but-empty and unset is significant.

2. Give up trying to automate this and make sure outside of ocaml/opam that OCAMLPARAM is set correctly before any invocation of `opam`.

@AltGr, can I ask you if, based on opam's handling of env vars, you see any other possible approaches to be able to set OCAMLPARAM to the value of an opam config var in a way that does not lead to cross-switch environment pollution?

AltGr:
> Oh, wow. I was working under the assumption that `opam switch foo; opam switch bar; opam switch foo` was equivalent to `opam switch foo`, but if opam doesn't implement that, then these setenv clauses are risky. 

It is partially implemented, but not perfect: opam will check the modification it did previously and attempt to _revert_ them, but:
1. it doesn't store the previous value of env variables, so destructive updates (like affectation) can not be reverted
2. it sets variable to the empty string rather than undef them (affectation was already quite tricky to get working reliably on all the shells out there, it didn't seem worth it to add so many cases for de-affectation)
These are a problem for the case of `OCAMLPARAM`, esp. 2. with ocaml < 4.12

The only way to more reliably handle reverting would be to store more information _within the env itself_, but that seems quite involved, and would amount to one huge variable definition; we would still have to handle cases where the variables were manually changed in the meantime.

As a related matter, I have just been running a small experiment on this branch for compiler variants which are just configuration flags for `ocaml`: https://github.com/ocaml/opam-repository/commits/recompilo
This approach could be scaled to defining environment variables that affect `ocaml` at runtime, but that seems a bit dangerous to me ; the upside is that, although it's a bit heavyweight to use virtual packages for configuration options, you get tracking and proper recompilations of reverse dependencies which you wouldn't get with opam variable

Blaisorblade:
What do you mean by “affectation”? After a few Googling attempts, I suspect a false friend with  French for *“variable assignment”, but affectation means > behaviour, speech, or writing that is pretentious and designed to impress.

* https://www.linguee.fr/francais-anglais/traduction/affectation+de+variable.html

jberdine:
@AltGr I guess it would be too large and invasive a change, but I wonder if a design was ever considered where opam did not change environment variables but instead executed everything in a subshell / child process? That way changing a switch would only change which environment variable settings would be applied to the future child processes, and those changes would not persist. I expect that this design could support uses such as `opam install ...` with the current user interface, but that the current user interface for uses such as `opam switch ...; ocamlopt` would not be possible. Essentially, instead of `eval $(opam config)` we would need to fork a subshell. I don't know how much people would revolt over such an interface.

PR #16620, opened 2020-06-12
Package np.np1.18-0.3.0

lehy:
### `np.np1.18-0.3.0`
Fundamental scientific computing with Numpy for OCaml
These are OCaml bindings to Numpy, the fundamental package for scientific computing with Python:
- powerful n-dimensional arrays
- numerical computing tools
- interoperable
- performant
- easy to use
- open source.

---
* Homepage: https://github.com/lehy/ocaml-sklearn
* Source repo: git+https://github.com/lehy/ocaml-sklearn.git
* Bug tracker: https://github.com/lehy/ocaml-sklearn/issues

---

### DIFF
--- a/packages/np/np.np1.18-0.3.0/opam
+++ b/packages/np/np.np1.18-0.3.0/opam
@@ -1,0 +1,41 @@
+opam-version: "2.0"
+synopsis: "Fundamental scientific computing with Numpy for OCaml"
+description: """
+These are OCaml bindings to Numpy, the fundamental package for scientific computing with Python:
+- powerful n-dimensional arrays
+- numerical computing tools
+- interoperable
+- performant
+- easy to use
+- open source.
+"""
+maintainer: ["Ronan Le Hy <ronan.lehy@gmail.com>"]
+authors: ["Ronan Le Hy"]
+license: "BSD-3-Clause"
+homepage: "https://github.com/lehy/ocaml-sklearn"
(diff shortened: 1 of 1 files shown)


lehy:
This is a new package, on which sklearn now depends. All three of np, scipy and sklearn are produced from the same git repo (the ocaml-sklearn one), and documented in the same place.

hannesm:
same comment as #16622: is the version `np1.18-0.3.0` intentional? I'm surprised by the `np1.` prefix.

lehy:
> same comment as #16622: is the version `np1.18-0.3.0` intentional? I'm surprised by the `np1.` prefix.

Yes, it is intentional. I am releasing the three packages from a single git repo, and it allows the git tags to match package versions without clashing (so my repo has tags np1.18-0.3.0 for package numpy, Python version 1.18, generator version 0.3.0 and sp1.4-0.3.0 for package scipy, Python version 1.4, generator version 0.3.0). 

These OCaml packages are tied to a fixed version of Python packages, and are produced using a common code generator, hence the two versions appearing in the OCaml package version.

I am not sure this is the simplest way to do things, but it seemed it would work (looking at how opam parses versions).

lehy:
I have little experience packaging, and welcome suggestions or pointers if this way of versioning/publishing is suboptimal, by the way :).

avsm:
looks good. If they're from the same repository, you can consider using `dune-release` to publish them. This will generate them in as one PR with multiple packages, and let them all be tested as a batch.

PR #16624, opened 2020-06-12
Add smart-print.0.3.0

clarus:
Add a new release for smart-print. We also modify `atd2cconv.0.0.0` as it requires `ocamlbuild` which was previously provided by smart-print.

### DIFF
--- a/packages/smart-print/smart-print.0.3.0/opam
+++ b/packages/smart-print/smart-print.0.3.0/opam
@@ -1,0 +1,25 @@
+opam-version: "2.0"
+
+maintainer: "dev@clarus.me"
+authors: ["Guillaume Claret <dev@clarus.me>"]
+homepage: "https://github.com/clarus/smart-print"
+bug-reports: "https://github.com/clarus/smart-print/issues"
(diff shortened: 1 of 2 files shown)


avsm:

clarus:
Thanks! This is mostly a maintenance release, to use Dune as a build system.

packages/smart-print/smart-print.0.3.0/opam
```diff
@@ -0,0 +1,25 @@
...
+authors: ["Guillaume Claret <dev@clarus.me>"]
+homepage: "https://github.com/clarus/smart-print"
+bug-reports: "https://github.com/clarus/smart-print/issues"
+license: "BSD-3-Clause"
+
+depends: [
+  "ocaml" {>= "4.00.0"}
+  "dune" {>= "2.0"}
+  "odoc" {with-doc}
+]
```
kit-ty-kate, line 13: Could you add a dependency on `ocamlfind`? Calling `dune install` requires `ocamlfind` (see https://github.com/ocaml/dune/issues/3025)

PR #16628, opened 2020-06-12
Fix ocsigenserver for GNU Make 4.3

kit-ty-kate:
cc @vouillon as told in https://github.com/ocsigen/ocsigenserver/pull/182#issuecomment-643302846 I believe users are going to experience this pretty soon. GNU Make 4.3 has only been around in Debian Sid for 3 weeks so far but if it's not a bug that is going to be fixed soon on their side I would rather see it fixed here.

For the record for everyone interested here is the change of behaviour. Using the following Makefile:
```
TEST :=
TEST += test
AAA := echo aaa${TEST}

all:
	$(AAA)
```
With GNU Make 4.2.1:
```
$ make --version
GNU Make 4.2.1
$ make
echo aaa test
aaa test
```
With GNU Make 4.3 on the same machine:
```
$ sudo apt-get upgrade make
[...]
$ make --version
GNU Make 4.3
$ make
echo aaatest
aaatest
```

### DIFF
--- a/packages/ocsigenserver/ocsigenserver.2.1/files/fix-gmake-4.3.patch
+++ b/packages/ocsigenserver/ocsigenserver.2.1/files/fix-gmake-4.3.patch
@@ -1,0 +1,48 @@
+commit 014aefc4e460686a361b974f16ebb7e0c993b36b (from e9ab14a4437640656117aed30a1ee5b6227f07de)
+Merge: e9ab14a4 5cc446bd
+Author: Jerome Vouillon <jvouillon@besport.com>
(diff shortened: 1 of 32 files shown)


avsm:
I'm ok with this change. When opam 2.1 is out, we can consider having a more sophisticated conf-gmake package and make older packages uninstallable when they hit this bug (its reasonable to have to upgrade in the usual case, rather than holding all the patches in opam-repository)

kit-ty-kate:
I've tested them locally, the patches all apply.

PR #16648, opened 2020-06-16
conduit-lwt-unix.2.2.2: disable package

avsm:
This causes a number of upstream users of cohttp-lwt-unix to
dynamically fail. Fix being discussed in https://github.com/mirage/ocaml-conduit/issues/318
so disabling package for now.


hcarty:
Thank you @avsm!

PR #16649, opened 2020-06-16
Fix ocamlfind dep for smart-print.0.3.0

clarus:
For the remark https://github.com/ocaml/opam-repository/pull/16624#discussion_r440955237 of @kit-ty-kate


kit-ty-kate:
Thanks a lot!

PR #16650, opened 2020-06-16
smart-print.0.3.0: dune install is not necessary when using deprecated_package_name

kit-ty-kate:
@clarus hi, sorry to make you change the opam description again. I thought the `(deprecate_package/library_name smart_print)` you're having was the reason you had a `dune install` in the first place. However after trying it, the resulting install file cleverly has the deprecate library in a separate section `lib_root` so everything is installed using the install file as it is the usual. So no need for calling `dune install` in the end.


clarus:
OK thanks! I upstream it.

PR #16651, opened 2020-06-17
added camlimages.5.0.3

camlspotter:


packages/camlimages/camlimages.5.0.3/opam
```diff
@@ -0,0 +1,21 @@
+opam-version: "1.2"
```
avsm, line 1: This needs to be upgraded to opam 2.0 format.  This can be done by `opam package-upgrade <opam file>` and then run `opam lint` to correct any errors.

camlspotter:
Thanks will do!

camlspotter:
Still some fixes required.

packages/camlimages/camlimages.5.0.3/opam
camlspotter: done!

kit-ty-kate:
Looks good now. Thanks!

PR #16656, opened 2020-06-17
Update asli upper bound on ott

CuiCui66:
ASLI versions `<= 0.2.0` are not compatible with ott version `0.31` so I'm adding an upper bound. I will soon release ASLI `0.2.1` that builds on top of ott version `0.31` once both this PR and the ott PR are merged.


kit-ty-kate:
I fixed the syntax and it's usually better to use a strict comparison operator when possible. A version `0.30.1` or a patch version `0.30-1` could be added for instance.

kit-ty-kate:
thanks!

PR #16660, opened 2020-06-17
Package embedded_ocaml_templates.0.1.3

EmileTrotignon:
### `embedded_ocaml_templates.0.1.3`
EML is a simple templating language that lets you generate text with plain OCaml
EML is a simple templating language that lets you generate text with plain OCaml
Inspired by EJS templates

---
* Homepage: https://github.com/EmileTrotignon/embedded_ocaml_templates
* Source repo: git+https://github.com/EmileTrotignon/embedded_ocaml_templates.git
* Bug tracker: https://github.com/EmileTrotignon/embedded_ocaml_templates/issues

---

### DIFF
--- a/packages/embedded_ocaml_templates/embedded_ocaml_templates.0.1.3/opam
+++ b/packages/embedded_ocaml_templates/embedded_ocaml_templates.0.1.3/opam
@@ -1,0 +1,34 @@
+opam-version: "2.0"
+synopsis: "EML is a simple templating language that lets you generate text with plain OCaml"
+description: """
+EML is a simple templating language that lets you generate text with plain OCaml
+Inspired by EJS templates
+"""
+maintainer: "Emile Trotignon emile.trotignon@gmail.com"
+authors: "Emile Trotignon emile.trotignon@gmail.com"
+license: "MIT"
+homepage: "https://github.com/EmileTrotignon/embedded_ocaml_templates"
+bug-reports: "https://github.com/EmileTrotignon/embedded_ocaml_templates/issues"
+dev-repo: "git+https://github.com/EmileTrotignon/embedded_ocaml_templates.git"
+depends: [ 
+    "ocaml"
+    "ocamlfind" 
+    "dune" {build} 
+    "sedlex" 
+    "core" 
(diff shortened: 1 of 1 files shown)


avsm:
Looks like this breaks as on older ocaml as it selects an ancient ppx_jane (v0.11) that doesnt work with dune:

```
# File "dune", line 6, characters 18-26:
# 6 |   (pps sedlex.ppx ppx_jane))
#                       ^^^^^^^^
# Error: Ppx dependency on a non-ppx library "ppx_jane". If "ppx_jane" is in
# fact a ppx rewriter library, it should have (kind ppx_rewriter) in its dune
# file.
```

As that version of ppx_jane was using jbuilder.

@jeremiedimino @rgrinberg, should we conflict all the pre-dune JS ppx packages with dune, so that they only get selected if jbuilder is used?

avsm:
@EmileTrotignon if you don't care about older versions of OCaml, you can also advance this PR by restricting your packages to OCaml >= 4.07 which bypasses the problem :-)

EmileTrotignon:
> @EmileTrotignon if you don't care about older versions of OCaml, you can also advance this PR by restricting your packages to OCaml >= 4.07 which bypasses the problem :-)

@avsm
Thanks a lot. I had it figured out, I did another PR (https://github.com/ocaml/opam-repository/pull/16664), but the tests still dont pass, this time for reasons I dont understand. I would really appreciate it if you take a look.
Should I close this PR ?

rgrinberg:
> should we conflict all the pre-dune JS ppx packages with dune, so that they only get selected if jbuilder is used?

Most versions of dune should be usable in fact. I think this issue is actually a subtle breaking change we introduced in dune 2.x. @snowleopard is this perhaps caused by the validation for ppx libraries we introduced?

snowleopard:
@rgrinberg You are right, this error is caused by the validation for ppx libraries. As far as I remember, the error is only active starting from version 2.2 of the Dune language, so one way to avoid it is to use an earlier language (not sure it's an option here).

mseri:
I am closing this, since it has been superseded by #16664

PR #16668, opened 2020-06-18
Release tuntap 2.0.0

hannesm:
CHANGES:
    
* Adapt to ipaddr 5.0.0 interface (mirage/ocaml-tuntap#35 @hannesm)
* Since Ipaddr.Prefix.t now contain both IP address and netmask, the functions
      `getifaddrs{,_v4,_v6}`, `addrs_of_ifname`, and `{v4,v6}_of_ifname` no longer
      return Ipaddr.{V4,V6}.t * Ipaddr.{V4,V6}.Prefix.t, but only the latter part
      (mirage/ocaml-tuntap#35 @hannesm)

### DIFF
--- a/packages/charrua-unix/charrua-unix.0.1/opam
+++ b/packages/charrua-unix/charrua-unix.0.1/opam
@@ -20,7 +20,7 @@ depends: [
   "sexplib" {< "113.09.00"}
   "camlp4" {build}
   "ocamlbuild" {build}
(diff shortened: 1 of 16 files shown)


hannesm:
the remaining CI issues are solver timeouts.

kit-ty-kate:
Thanks a lot!

PR #15727, opened 2020-01-24
mlcuddidl.3.0.6

nberth:
Supports OCaml 4.10

PR #16670, opened 2020-06-18
mlcuddidl.3.0.6

nberth:
This brings a fix in the API.

It also constitutes a new attempt after https://github.com/ocaml/opam-repository/pull/15727


kit-ty-kate:
Looking into the code, it seems that `camlidl` is a runtime dependency of this package, so I removed the `{build}` tag.

Otherwise there was no issue. Thanks a lot!

nberth:
@kit-ty-kate thanks a lot for merging!

For the build flag about the dependency to camlidl, normally it's runtime is statically linked within the libraries that mlcuddidl provides. But maybe there's a glitch in the Makefile; I don't think it matters much to enforce having camidl around anyways.

PR #16674, opened 2020-06-19
conf-python-2.7: fix depext for alpine

AltGr:
seems to be 'python2' and not 'python', at least in alpine 3.11.
    
('python' is a virtual package, so 'apk add python' would work, but opam would still detect the depext as uninstalled)
    
Q: why are there two versions of this package, with apparently just one depext fixed between the two ?


kit-ty-kate:
Thanks!

PR #16657, opened 2020-06-17
Add coq-of-ocaml.2.2.0

clarus:
New release of coq-of-ocaml. The previous releases were on the Coq opam repository. The rationale for doing a release here is that it does not necessarily depend on Coq anymore (could be used as a lint tool for a CI for example).

PR #16675, opened 2020-06-19
Add coq-of-ocaml.2.2.1

clarus:
Should be a cleaner version of the previous attempt with coq-of-ocaml.2.2.0 https://github.com/ocaml/opam-repository/pull/16657 with a more standard build.


kit-ty-kate:
I'm not sure if coq-of-ocaml uses this part of merlin, but just in case to avoid issues I've removed the `{build}` tag from the ocamlfind dependency and added the same constraint as there is in the merlin package.

Thanks!

clarus:
OK, thanks for the review work!

PR #16679, opened 2020-06-19
[new release] tcpip (5.0.0)

hannesm:
CHANGES:

* Static_ipv4.connect API change: takes a cidr:Ipaddr.V4.Prefix.t instead of
  ip:Ipaddr.V4.t and network:Ipaddr.V4.Prefix.t (mirage/mirage-tcpip#426 @hannesm)
* Adapt to ipaddr 5.0.0 API changes (mirage/mirage-tcpip#426 @hannesm)


hannesm:
@kit-ty-kate thanks for the label. I'd like to wait for a revdeps run -- I expect there will be some failures which I'm happy to fix (constrain the upper bounds) before a merge.

kit-ty-kate:
Don't worry, the label is only there to remind myself that I've checked the package description, I would not have merged it without waiting for the revdeps first ofc ^^

PR #16682, opened 2020-06-20
Release tls 0.12.2

hannesm:
CHANGES:
    
* tls_lwt again calls Mirage_crypto_rng_lwt.initialize () -- which is since
      mirage-crypto-rng 0.8 no longer inside the lwt monad, and safe to be called
      multiple times and on top level (mirleft/ocaml-tls#415 by @hannesm)

this also marks 0.12.1 as `available: false` due to semantics change in respect to RNG initialization between 0.12.0 and 0.12.1, which has been reverted in 0.12.2


hannesm:
not sure what to do with the CI reporting an internal failure "(Failure "9p error: connection disconnected")" with no possibility for me to "restart" it.

hannesm:
to me this looks good to merge, remaining failures are colombe (broken since ages) and solver timeouts.

kit-ty-kate:
Thanks!

PR #16684, opened 2020-06-20
allow OCaml < 4.10 for fstar dev-repo

tahina-pro:
Requested by @catalin-hritcu 

The current opam packages for F* do not allow OCaml >= 4.08. However, the latest F* `master` does compile with OCaml < 4.10.
Thus, this pull request is meant to make `opam pin fstar --dev-repo` work for OCaml < 4.10, as documented at https://github.com/FStarLang/FStar/blob/3025f90e008f5793e3755ab115aecb4c17ab3714/INSTALL.md#opam-package

### DIFF
--- a/packages/fstar/fstar.0.9.7.0-alpha1/opam
+++ b/packages/fstar/fstar.0.9.7.0-alpha1/opam
@@ -5,7 +5,7 @@ authors: "Nik Swamy <nswamy@microsoft.com>,Jonathan Protzenko <protz@microsoft.c
 homepage: "http://fstar-lang.org"
 license: "Apache-2.0"
 depends: [
-  "ocaml" {>= "4.04.0" & < "4.08.0"}
+  "ocaml" {>= "4.04.0" & ( < "4.08.0" | (dev & < "4.10.0")) }
   "ocamlfind"
   "batteries"
   "zarith"


avsm:
This won't quite work, unfortunately, as it'll fail if you pin the released version of fstar with 4.10.

The solution to document for this is to commit the correct opam file into the f* repository, and then `opam pin add fstar --dev` should work, as opam will pick up that file and use its constraints.

tahina-pro:
Sorry, I don't quite understand the first part of your comment. Could you please clarify? In fact, I do intend both the released version and `--dev-repo` to fail with OCaml >= 4.10, but only `--dev-repo` to succeed for OCaml >= 4.08 and < 4.10.

XVilka:
@tahina-pro shouldn't be https://github.com/FStarLang/FStar/issues/1792 fixed first? I think ulex was abandoned and will not work with newer OCaml.

catalin-hritcu:
@XVilka The purpose of this patch is *not* to support OCaml 4.10. It's to support 4.9.1, which is currently wrongly rejected for the dev repo, even if it would be the version we would normally ask people to use these days.

XVilka:
Oh, sorry, I misread.

avsm:
Right; this is what you need to do to support the master branch of F*:

- add a `fstar.opam` file to your repository, with the correct metadata (that is `"ocaml" {<"4.10.0"}` for example)
- run `opam pin add fsharp --dev` to install it.

That's all.  The `--dev` option will query the dev-repo field in the latest known version of fsharp in the opam-repo, use that to retrieve the master branch opam file, and subsequently use that for the metadata calculation.

This way, you don't need to backpatch existing fsharp metadata versions. When the new fsharp is ready to release, `opam-publish` or `dune-release` will just grab that latest metadata file in your trunk and use that for the new release.

catalin-hritcu:
Anil's suggestion indeed does the trick, so this pull request is no longer needed and can be safely closed. Many thanks.

mseri:
Thanks!

PR #16708, opened 2020-06-24
[new release] index (1.2.1)

pascutto:
CHANGES:

## Added

- Added `Index_unix.Syscalls`, a module exposing various Unix bindings for
  interacting with file-systems.

## Fixed

- Fail when `Index_unix.IO` file version number is not as expected.
- Fixed creation of an index when an empty `data` file exists.


avsm:

PR #16712, opened 2020-06-24
Add ocamlfuse 2.7.1-cvs7

astrada:


kit-ty-kate:
The `dune-project` file in your source repository indicates this package requires at least dune 1.6 but no lower-bound constraint was set for dune. Also the `light-uninstall` flag was unnecessary as Camelus suggests. Could you return the fix to the opam file upstream?

astrada:
OK, thanks!

kit-ty-kate:
Thanks!

PR #16717, opened 2020-06-25
ezjs_min 0.1

zapashcanon:


kit-ty-kate:
Thanks! There was a missing dependency in the description. Could you return the fix in the opam file upstream so that this does not happen for next release?

zapashcanon:
Done. :)

PR #16583, opened 2020-06-05
ocaml: compile with -fcommon to support gcc 10

avsm:
gcc 10 and upwards alters the defaults such that duplicate C symbols results in a linker error.  This PR forces compilation with the `-fcommon` flag, which preserves the pre-gcc-10 behaviour and thus compatibility with OCaml <4.09.1.

`-fcommon` has been supported from gcc 4.4 upwards, which is a reasonable base default (checked by @Octachron)

A PR to do this for ocaml-variants will follow once this is merged.

PR #16722, opened 2020-06-26
ocaml-variants: fix gcc10 compilation with -fcommon

avsm:
As #16583 explains, gcc 10 and upwards alters the defaults such that duplicate C symbols results in a linker error. This PR forces compilation with the -fcommon flag, which preserves the pre-gcc-10 behaviour and thus compatibility with OCaml <4.09.1.
    
-fcommon has been supported from gcc 4.4 upwards, which is a reasonable base default.
    
This patch skips a few compiler variants that already have patches, such as termux and the bucklescript variants. Those can deal with gcc10 on their own if they still need it.
    
 The patches are copies of the respective base patch for that compiler, found from the ocaml-base-compiler package from #16583.

Testing by running `opam source` over every affected package, which applies the patch.  Not fully build tested yet, but I'll be adding this PR to the CI container builds scripts to do so before merging live.


mseri:
I randomly picked a dozen green tests, but they all seem to failed with
```
$ opam depext -uivyj 72 ocaml-variants.4.06.0+fp
# Detecting depexts using vars: arch=x86_64, os=linux, os-distribution=opensuse-leap, os-family=suse
[ERROR] No solution for ocaml-variants.4.06.0+fp: Your request can't be satisfied:
          - no matching version


Command failed: opam list --readonly --external  --resolve=ocaml-variants.4.06.0+fp returned 20
```

avsm:
Thanks for checking that @mseri. It looks like the datakit-ci doesn't understand ocaml base packages, as I never added a special case for them. So what it's trying to do is to solve for the changed ocaml package without an `--unlock-base`, and hence all the resolutions will fail.

What I've done is to manually merge in my branch into the CI containers on ocaml/opam2, and [that build](https://gitlab.com/ocaml-platform/ci/-/pipelines/160845619) has finished and works successfully. That build matrix also now includes Fedora 32 and gcc10, as well as the older distros, and with build successes there I think this good to merge. 

That just leaves the secondary compiler to fix now, which should be pretty easy, and we have gcc10 support again. Phew!

avsm:
In it goes! Will be on the lookout for any regressions...

mseri:
:shipit: 🤞

Blaisorblade:
Any reason why 4.07.1+flambda is not included in this patch? That's still the recommended compiler for Coq.

Blaisorblade:
I'm sending a PR for that.

PR #16727, opened 2020-06-28
Package tdigest.0.1.0

SGrondin:
### `tdigest.0.1.0`
OCaml implementation of the T-Digest algorithm
OCaml implementation of the T-Digest algorithm.

The T-Digest is a data structure and algorithm for constructing an approximate distribution for a collection of real numbers presented as a stream.

The T-Digest can estimate percentiles or quantiles extremely accurately even at the tails, while using a fraction of the space.

Additionally, the T-Digest is concatenable, making a good fit for distributed systems. The internal state of a T-Digest can be exported as a binary string, and the concatenation of any number of those strings can then be imported to form a new T-Digest.

---
* Homepage: https://github.com/SGrondin/tdigest
* Bug tracker: https://github.com/SGrondin/tdigest/issues

---

### DIFF
--- a/packages/tdigest/tdigest.0.1.0/opam
+++ b/packages/tdigest/tdigest.0.1.0/opam
@@ -1,0 +1,36 @@
+opam-version: "2.0"
+maintainer: "Simon Grondin"
+authors: [
+  "Simon Grondin"
+  "Will Welch"
+]
+synopsis: "OCaml implementation of the T-Digest algorithm"
+description: """
+The T-Digest is a data structure and algorithm for constructing an approximate distribution for a collection of real numbers presented as a stream.
+
+The T-Digest can estimate percentiles or quantiles extremely accurately even at the tails, while using a fraction of the space.
+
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Dune is not compatible with the `{build}` flag (see https://github.com/ocaml/dune/issues/2147) and you `dune-project` file indicates that this package requires at least dune 2.4. I've made the required fixes. Could you return these fixes to the opam file upstream so that this won't happen again for next release?

Also, I had a look at your project and there didn't seem to be anything actually installed by it. None of the binaries or libraries were tagged as being part of the package or being a public library so nothing should be installed. Could you fix this? (see https://github.com/ocaml/dune/issues/2882)

SGrondin:
Thank you, I'll make the changes shortly.

SGrondin:
It should be ready to go now. Thanks for reviewing it

avsm:
Force pushes are just fine to your branch, no apologies required. Thanks for the package contribution. You may want to announce this on <https://discuss.ocaml.org>, where we have a Community category and an `announce` tag for this purpose.

PR #16728, opened 2020-06-29
ocaml-secondary-compiler: fix with gcc10

avsm:
as with #16583 and #16722, this fixes the secondary compiler
with gcc10, by passing -fcommon to the compiler


packages/ocaml-secondary-compiler/ocaml-secondary-compiler.4.08.1-1/files/fix-gcc10.patch
```diff
@@ -0,0 +1,43 @@
...
+diff --git a/configure b/configure
+index 9a78a4554..0c54b560b 100755
+--- a/configure
++++ b/configure
+@@ -2688,7 +2688,7 @@ mkexe="\$(CC) \$(OC_CFLAGS) \$(OC_CPPFLAGS) \$(OC_LDFLAGS)"
+ 
+ # Flags for building executable files with debugging symbols
+ mkexedebugflag="-g"
+-common_cflags=""
++common_cflags="-fcommon"
```
dra27, line 18: This isn't in the `configure.ac` diff (and FWIW would break MSVC)?

avsm: that snuck in! removed in the base compiler packages too

dra27:
That stray hunk seems to be present in all the ocaml-variants patches for 4.08.0 and 4.08.1 as well, I'm afraid!

avsm:
Fixed, this should be good for merge after running a few more tests.

PR #16733, opened 2020-06-29
Package SZXX.0.1.0

SGrondin:
### `SZXX.0.1.0`
Streaming ZIP XML XLSX parser
SZXX is a streaming, non-seeking and efficient XLSX parser built from ground up for low memory usage.
SZXX is able to output XLSX rows while a file is being read from the file descriptor without buffering any part of the file.

---

---

### DIFF
--- a/packages/SZXX/SZXX.0.1.0/opam
+++ b/packages/SZXX/SZXX.0.1.0/opam
@@ -8,7 +8,7 @@ description: """
 SZXX is a streaming, non-seeking and efficient XLSX parser built from ground up for low memory usage.
 SZXX is able to output XLSX rows while a file is being read from the file descriptor without buffering any part of the file.
 """
-license: "Asemio"
+license: "MIT"
 tags: ["Stream" "ZIP" "XML" "XLSX"]
 homepage: "https://github.com/asemio/SZXX"
 dev-repo: "git://github.com/asemio/SZXX"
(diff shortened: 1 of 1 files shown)


SGrondin:
It seems like the build script is having issues? I don't think it's an issue with the package, is it?
I followed the same process I did with another package yesterday, the only significant difference I see is this one relies on `camlzip` for its C bindings.

packages/SZXX/SZXX.0.1.0/opam
```diff
@@ -0,0 +1,38 @@
...
+maintainer: "Asemio"
+authors: [
+  "Simon Grondin"
+]
+synopsis: "Streaming ZIP XML XLSX parser"
+description: """
+SZXX is a streaming, non-seeking and efficient XLSX parser built from ground up for low memory usage.
+SZXX is able to output XLSX rows while a file is being read from the file descriptor without buffering any part of the file.
+"""
+license: "Asemio"
```
mseri, line 11: According to the repository this should be MIT. Is that ok if I update it? (If so, please upstream that change)

mseri:
Yes, something weird was going on in the CI. I was going to merge but I noticed the licensing issue: once that is resolved I'll merge. I can make the change here before merging if you confirm that it is correct

SGrondin:
@mseri  Yes please change it to MIT and merge, that was an oversight on my part! I'll fix it upstream. Thank you so much!

PR #16737, opened 2020-07-01
[missing release] z3 (4.8.5)

floriangru:
This pull request adds a missing release to the Z3 prover (4.8.5). This version of Z3 built with opam has been successfully been used to build [FStar](https://github.com/FStarLang/FStar), [kremlin](https://github.com/FStarLang/kremlin) and [hacl-star](https://github.com/project-everest/hacl-star). 
Opam files are based upon the 4.8.8-1 release, so opam-publish hasn't been used in the process.

### DIFF
--- a/packages/z3/z3.4.8.5/files/4468.patch
+++ b/packages/z3/z3.4.8.5/files/4468.patch
@@ -1,0 +1,267 @@
+From 8179ae62817d46c7a3df82aded1ff105fa2c3646 Mon Sep 17 00:00:00 2001
+From: ivg <ivg@ieee.org>
+Date: Wed, 27 May 2020 07:52:52 -0400
+Subject: [PATCH] fixes numerous issues in OCaml bindings building process
+
+It now works both in dynamic and static mode and the compiled
+libraries can be used by all linkers in the OCaml system, without
+any specificy instructions other than specifying the dependency on
+the z3 library.
+
+Using the libraries
+===================
+
+Compiling binaries
+------------------
+
+The libraries can be linked statically with both ocamlc and ocamlopt
+compilers, e.g.,
+
+```
+ocamlfind ocamlc -thread -package z3 -linkpkg run.ml -o run
+```
+or
+```
+ocamlfind ocamlopt -thread -package z3 -linkpkg run.ml -o run
+```
+
+When bindings compiled with the `--staticlib` the produced binary will
+not have any dependencies on z3
+```
+$ ldd ./run
+        linux-vdso.so.1 (0x00007fff9c9ed000)
+        libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007fb56f09c000)
+        libgmp.so.10 => /usr/lib/x86_64-linux-gnu/libgmp.so.10 (0x00007fb56ee1b000)
+        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fb56ebfc000)
+        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fb56e85e000)
+        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fb56e65a000)
+        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fb56e442000)
+        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fb56e051000)
+        /lib64/ld-linux-x86-64.so.2 (0x00007fb570de9000)
+```
+
+The bytecode version will have a depedency on z3 and other external
+libraries (packed as dlls and usually installed in opam switch):
+```
+$ ocamlobjinfo run | grep 'Used DLL' -A5
+Used DLLs:
+        dllz3ml
+        dllzarith
+        dllthreads
+        dllunix
+```
+
+But it is possible to compile a portable self-contained version of the
+bytecode executable using the `-custom` switch:
+
+```
+ocamlfind ocamlc -custom -thread -package z3 -linkpkg run.ml -o run
+```
+
+The build binary is now quite large but doesn't have any external
+dependencies (modulo the system dependencies):
+```
+$ du -h run
+27M     run
+$ ocamlobjinfo run | grep 'Used DLL' | wc -l
+0
+$ ldd run
+        linux-vdso.so.1 (0x00007ffee42c2000)
+        libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007fdbdc415000)
+        libgmp.so.10 => /usr/lib/x86_64-linux-gnu/libgmp.so.10 (0x00007fdbdc194000)
+        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fdbdbf75000)
+        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fdbdbbd7000)
+        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fdbdb9d3000)
+        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fdbdb7bb000)
+        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fdbdb3ca000)
+        /lib64/ld-linux-x86-64.so.2 (0x00007fdbde026000)
+```
+
+Loading in toplevel
+-------------------
+
+It is also possible to use the built libraries in toplevel and use
+them in ocaml scripts, e.g.,
+```
+$ ocaml
+        OCaml version 4.09.0
+
+ # #use "topfind";;
+ - : unit = ()
+ Findlib has been successfully loaded. Additional directives:
+  #require "package";;      to load a package
+  #list;;                   to list the available packages
+  #camlp4o;;                to load camlp4 (standard syntax)
+  #camlp4r;;                to load camlp4 (revised syntax)
+  #predicates "p,q,...";;   to set these predicates
+  Topfind.reset();;         to force that packages will be reloaded
+  #thread;;                 to enable threads
+
+- : unit = ()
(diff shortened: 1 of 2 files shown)


ivg:
Looks like that all the failure are CI related so I had to test on mine machine. And while the patch was applied successfully it looks like that there is a missing dependency to `libgomp`,

```
$ ocaml
        OCaml version 4.03.0

# #use "topfind";;
- : unit = ()
Findlib has been successfully loaded. Additional directives:
  #require "package";;      to load a package
  #list;;                   to list the available packages
  #camlp4o;;                to load camlp4 (standard syntax)
  #camlp4r;;                to load camlp4 (revised syntax)
  #predicates "p,q,...";;   to set these predicates
  Topfind.reset();;         to force that packages will be reloaded
  #thread;;                 to enable threads

- : unit = ()
# #require "z3";;
/home/ivg/.opam/4.03.0/lib/ocaml/nums.cma: loaded
/home/ivg/.opam/4.03.0/lib/num-top: added to search path
/home/ivg/.opam/4.03.0/lib/num-top/num_top.cma: loaded
/home/ivg/.opam/4.03.0/lib/num: added to search path
/home/ivg/.opam/4.03.0/lib/z3: added to search path
/home/ivg/.opam/4.03.0/lib/z3/z3ml.cma: loaded
Cannot load required shared library dllz3ml.
Reason: /home/ivg/.opam/4.03.0/lib/stublibs/dllz3ml.so: /home/ivg/.opam/4.03.0/lib/stublibs/dllz3ml.so: undefined symbol: omp_unset_nest_lock.
# 
```

kit-ty-kate:
I can confirm that even if the CI is green, loading the `z3` package in the top-level ends up failing as @ivg points out.

packages/z3/z3.4.8.5/files/4468.patch
```diff
@@ -0,0 +1,267 @@
...
+diff --git a/scripts/mk_util.py b/scripts/mk_util.py
+index 1a95ee20c3..156be9ca2f 100644
+--- a/scripts/mk_util.py
++++ b/scripts/mk_util.py
+@@ -1957,7 +1957,7 @@ def mk_makefile(self, out):
+ 
+             OCAMLMKLIB = 'ocamlmklib'
+ 
+-            LIBZ3 = '-cclib -l' + z3link
++            LIBZ3 = '-l' + z3link + ' -lstdc++'
```
ivg, line 252: @floriangru, we need to add the missing libgomp here, i.e., `-lgomp`.

ivg (changes requested):
A quick search revealed that at z3 4.8.5  was dependent on GOMP, the dependency that they have dropped later. Therefore, to make this package buildable and runnable we need to 

- [x] add the missing GOMP dependency to the build script to enable proper LDD entries
- [ ] add the conf-gomp package that will enable the installation of GOMP on the supported operating systems

Probably, we can skip the second step, as we are already passing the build step on CI. That is because libgomp is a part of the toolchain, 
```
$ aptitude why libgomp1
i   libstdc++-6-dev Depends libgcc-6-dev (= 6.5.0-2ubuntu1~18.04)
i A libgcc-6-dev    Depends libgomp1 (>= 6.5.0-2ubuntu1~18.04)
```

However, this might not be true on Mac OS X and some other rare systems, @kit-ty-kate what do you think, do we need conf-gomp package?

avsm:
@floriangru have you had a chance to review the suggested changes above?

mseri:
@floriangru any update on this?

catalin-hritcu:
Wouldn't wait for @floriangru, who's taking care of family at the moment, especially since you already know much better what to do here. I've also contacted one of the Z3 developers (@wintersteiger), who also confirmed this:
> What they said in the discussion is right, Z3 depends on OpenMP, so the packaging scripts need to know about it. I think Nikolaj added a flag to disable OpenMP (and everything that's parallel) at some point, but I'm not sure how well that would work for OCaml users. Also, at least on Windows it's impossible to link OpenMP statically (by design), so one way or another this has to be handled by the packaging scripts.

packages/z3/z3.4.8.5/files/4468.patch
mseri: @catalin-hritcu do you know about this issue?

catalin-hritcu: @mseri: Not really, sorry. I'm one of the developers of F* who originally encouraged @floriangru to submit this pull request, because for us exactly this version of Z3 is a dependency. Neither me nor @floriangru knows much about Z3 packing, but we were hoping it would be easy :)

mseri: Ok, thanks for letting us know. Let's try to unlock this PR.

@ivg, since you seem to know what is missing, would you have time to take over the PR and add the missing `-lgomp` where needed?

mseri: > Neither me nor @floriangru knows much about Z3 packing, but we were hoping it would be easy :)

Normally it is, but it depends very much on how complex is the package and its set of external dependencies.

ivg: @mseri, sure, I will take a loot into it.

ivg, line 252:
```suggestion
+            LIBZ3 = '-l' + z3link + ' -lstdc++' + '-lgomp'
```

ivg: @mseri, I can't really edit this PR (other than creating a PR to @floriangru master branch) so I just made the suggested change and any maintaner should be able to accep it). This should fix the linking error. 

Concerting the `conf-gomp` package, I am for skipping this step, as it is a part of the C++ toolchain and it is already in the dependencies of the z3 package.

mseri: Thank you very much

packages/z3/z3.4.8.5/files/4468.patch
```diff
@@ -0,0 +1,267 @@
...
+diff --git a/scripts/mk_util.py b/scripts/mk_util.py
+index 1a95ee20c3..156be9ca2f 100644
+--- a/scripts/mk_util.py
++++ b/scripts/mk_util.py
+@@ -1957,7 +1957,7 @@ def mk_makefile(self, out):
+ 
+             OCAMLMKLIB = 'ocamlmklib'
+ 
+-            LIBZ3 = '-cclib -l' + z3link
++            LIBZ3 = '-l' + z3link + ' -lstdc++' + '-lgomp'
```
mseri, line 252:
```suggestion
+            LIBZ3 = '-l' + z3link + ' -lstdc++' + ' -lgomp'
```

I think this needs an extra space here (@ivg correct me if I am wrong)

packages/z3/z3.4.8.5/opam
```diff
@@ -0,0 +1,32 @@
...
+depends: [
+  "ocaml"
+  "ocamlfind" {build}
+  "num"
+  "conf-gmp"
+  "conf-python-2-7" {build}
+]
+synopsis: "Z3 solver"
+extra-files: [
+  ["4468.patch" "md5=05a00fdd98aecba2198393bff8484efa"]
```
mseri, line 27:
```suggestion
  ["4468.patch" "sha512=d4cf90a4b3a9e9f1d07b1e6a04f89fb2c74e429d7e151bbc91841b54e34c8adbcd74183d79ed95c4df5d33b064dc715125523c7842cc3dfcc2b5b42ac83a244c"]
```

packages/z3/z3.4.8.5/files/4468.patch
ivg: the space is definitely needed and I was pretty sure that I put one... but indeed I missed it

mseri: No problem. I have added it and the new hash, let's see if it makes the CI happy and resolve this PR.
Thanks for the help!

packages/z3/z3.4.8.5/files/4468.patch
mseri: Mmh, it fails on macos due to missing gomp

```
#=== ERROR while compiling z3.4.8.5 ===========================================#
2100# context     2.0.7 | macos/x86_64 | ocaml-system.4.10.0 | file:///Users/travis/build/ocaml/opam-repository
2101# path        ~/.opam/ocaml-system/.opam-switch/build/z3.4.8.5
2102# command     ~/.opam/opam-init/hooks/sandbox.sh build make -C build -j 1
2103# exit-code   2
2104# env-file    ~/.opam/log/z3-93566-60459a.env
2105# output-file ~/.opam/log/z3-93566-60459a.out
2106### output ###
2107# 206 |   let compare = Pervasives.compare
2108# [...]
2109# Alert deprecated: module Stdlib.Pervasives
2110# Use Stdlib instead.
2111# 
2112# If you need to stay compatible with OCaml < 4.07, you can use the 
2113# stdlib-shims library: https://github.com/ocaml/stdlib-shims
2114# cp libz3.a libz3-static.a
2115# ocamlfind ocamlc -package num  -ccopt "-D_MP_INTERNAL -DNDEBUG -D_EXTERNAL_RELEASE   -fvisibility=hidden -c -mfpmath=sse -msse -msse2 -D_NO_OMP_ -O3 -Wno-unknown-pragmas -Wno-overloaded-virtual -Wno-unused-value -fPIC -I /usr/local/lib/ocaml -I ../src/api -I ../src/api/ml -o api/ml/z3native_stubs.o" -c ../src/api/ml/z3native_stubs.c
2116# ocamlmklib -o api/ml/z3ml -I api/ml -L. api/ml/z3native_stubs.o api/ml/z3enums.cmo api/ml/z3native.cmo api/ml/z3.cmo  -lz3-static -lstdc++ -lgomp 
2117# ld: library not found for -lgomp
2118# clang: error: linker command failed with exit code 1 (use -v to see invocation)
2119# make: *** [api/ml/z3ml.cma] Error 2
```

mseri:
I don’t know what is the right way to deal with macosx, it seems that it would need manual intervention no matter what: https://stackoverflow.com/questions/64116617/libgomp-and-macosx

mseri:
With macports I need to install gcc10 and then can compile successfully a test by calling the custom macports gcc: `gcc-mp-10 -lgomp test.c`. With homebrew it is a similar story. I am inclined to merge, there isn't much we can do anyway

PR #16738, opened 2020-07-01
[new release] mirage-net-xen and netchannel (1.13.1)

hannesm:
CHANGES:

* MirageOS (mirage-net) defines the MTU as the link-level payload size, adjust from 1514 to 1500 (@hannesm, mirage/mirage-net-xen#98)
* Only pass the sub-buffer of requested size to the fill function (solves mirage/qubes-mirage-firewall#111, @hannesm, mirage/mirage-net-xen#98)
* listen: do not catch out of memory exception (@hannesm, mirage/mirage-net-xen#97)


kit-ty-kate:
Thanks!

PR #16739, opened 2020-07-01
Package jhupllib.0.2.2

zepalmer:
### `jhupllib.0.2.2`
A collection of OCaml utilities used by the JHU PL lab
A collection of OCaml utilities used by the JHU PL lab.  These are miscellaneous
utilities which did not appear readily in standard libraries when they were
written.

---
* Homepage: http://pl.cs.jhu.edu/
* Source repo: git+https://github.com/JHU-PL-Lab/jhu-pl-lib.git
* Bug tracker: https://github.com/JHU-PL-Lab/jhu-pl-lib/issues

---

### DIFF
--- a/packages/jhupllib/jhupllib.0.2.2/opam
+++ b/packages/jhupllib/jhupllib.0.2.2/opam
@@ -18,12 +18,10 @@ depends: [
   "ocaml" {>= "4.07.0"}
   "base-threads"
   "batteries"
-  "dune" {build & >= "1.0+beta17"}
+  "dune" {>= "1.0"}
   "monadlib"
   "ocaml-monadic" {>= "0.4.1"}
-  "ocamlbuild" {build}
-  "ocamlfind" {build}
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
I removed the dependencies towards `ocamlfind` and `ocamlbuild` as they are not used anymore.
`ounit` was tagged as a build dependency, however it seems to be rather a test dependency according to the `test/dune` file.
Dune was tagged as a build dependency, however dune does not support this tag (see https://github.com/ocaml/dune/issues/2147).
Dune was also given a lower-bound constraint that did not match its requirement.

Could you return these changes to the opam file upstream so that they are fixed for the next release as well?

zepalmer:
I have done so.  Thank you both for the effort and for teaching me about the ecosystem.  I'm definitely an OCaml packaging novice.

kit-ty-kate:
Thanks a lot!

PR #16740, opened 2020-07-02
[new release] mirage-crypto-pk, mirage-crypto, mirage-crypto-rng and mirage-crypto-rng-mirage (0.8.1)

hannesm:
CHANGES:
    
* Add Chacha20 implementation (based on abeaumont/ocaml-chacha), supporting
      both DJB's original specification (nonce 64 bit, counter 64 bit) and IETF
      (RFC 8439: nonce 96 bit, counter 32 bit)
      (mirage/mirage-crypto#72 @hannesm)
* Add Poly1305 implementation based on floodyberry/poly1305-donna (mirage/mirage-crypto#72 @hannesm)
* Unified AEAD module type, implemented by CCM, GCM, and Chacha20/Poly1305
      The functions "authenticate_encrypt" and "authenticate_decrypt" are defined,
      which append (encrypt) and check equality (decrypt, using Eqaf for
      constant-time comparison) the authentication tag directly.
      Breaking changes:
      - GCM "~iv" is now "~nonce"
      - GCM encrypt returns the encrypted data and authentication tag appended
      - GCM decrypt returns the plaintext as option (None on authentication failure)
      (mirage/mirage-crypto#73 @hannesm)

### DIFF
--- a/packages/mirage-crypto-pk/mirage-crypto-pk.0.8.1/opam
+++ b/packages/mirage-crypto-pk/mirage-crypto-pk.0.8.1/opam
@@ -1,0 +1,42 @@
+opam-version: "2.0"
+homepage:     "https://github.com/mirage/mirage-crypto"
+dev-repo:     "git+https://github.com/mirage/mirage-crypto.git"
+bug-reports:  "https://github.com/mirage/mirage-crypto/issues"
+doc:          "https://mirage.github.io/mirage-crypto/doc"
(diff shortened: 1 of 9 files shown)


hannesm:
a reverse depedency run would be appreciated since there may be some CCM/GCM users which don't compile anymore.

kit-ty-kate:
Everything seems good. Thanks a lot!

PR #16753, opened 2020-07-04
[Discussion] New compiler flag scheme

kit-ty-kate:
This draft PR attacks several issues in how OCaml compilers are defined in opam-repository:
1) There is currently no ways of saying "this package isn't compatible with this or that compiler variant (e.g. flambda)". To fix that, the meta-package `base-<flag>` (e.g. `base-flambda`) is added. Which allows people to either request or conflict with particular variant of the compiler. cc @emillon 
2) There is currently no ways of combining different flags together other than to define a new `ocaml-variant.<ver>+<flag1>+<flag2>` package. Given the number of available combinations, this can be tedious and unmaintainable. To fix that, the new `base-<flag>` meta-packages become a optional dependencies of the `ocaml-base-compiler` package. This way, one could request a new custom variant this way: `opam switch create --packages=ocaml-base-compiler.4.10.0,base-flambda,base-spacetime 4.10+flambda+spacetime`. This could be improved in opam itself by adding an optional argument of some kind; something of the sort `opam switch create --extra-packages=base-flambda 4.10 4.10.0`.
3) The number and different nature of `ocaml-variants` packages is hard to maintain, this PR or similar would make them obsolete and only to be used for bigger compiler variants/changes such as metaocaml, betas or 32bits variants.

Currently this PR is only a place for discussion, in no way something ready to be merged.
The current issues with this approach is that I couldn't find a way to add the lack of an optional dependency to the switch invariant, so with this one could install `base-flambda` without realizing and the next thing they know their whole switch is being recompiled... This is less than ideal. I'd be happy to know if there is any ways of doing something like this while keeping the current behaviour of opam. Maybe this could be done by creating a `ocaml-base-compiler.config` file, then make the new `base-<flag>` packages available only if `ocaml-base-compiler:<flag-name>` is set to `true`. I'm not sure this works, I'll investigate later. cc @rjbou @AltGr @thomasblanc

### DIFF
--- a/packages/base-afl/base-afl.base/opam
+++ b/packages/base-afl/base-afl.base/opam
@@ -1,0 +1,2 @@
+opam-version: "2.0"
+synopsis: "A pseudo-package stating that the OCaml compiler has been compiled with afl support"
--- a/packages/base-flambda/base-flambda.base/opam
+++ b/packages/base-flambda/base-flambda.base/opam
@@ -1,0 +1,2 @@
+opam-version: "2.0"
+synopsis: "A pseudo-package stating that the OCaml compiler has been compiled with flambda support"
--- a/packages/base-frame-pointers/base-frame-pointers.base/opam
+++ b/packages/base-frame-pointers/base-frame-pointers.base/opam
@@ -1,0 +1,2 @@
+opam-version: "2.0"
+synopsis: "A pseudo-package stating that the OCaml compiler has been compiled with frame-pointers support"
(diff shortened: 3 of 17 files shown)


Blaisorblade:
@kit-ty-kate can one abuse opam configuration variables as package flags, instead of abusing packages for the job? That's one of the options discussed in https://github.com/ocaml/opam-repository/pull/16619... Conversely, I'd like some way to build variants of the Coq package.

dra27:
This is a nice idea. I would possibly word the descriptions slightly differently (perhaps even avoiding the prefix `base-`) - if you depend on `base-afl`, you don't state the compiler needs to have been built with AFL support, you'll cause opam to recompile it with AFL support.

`base-no-nativecode` sticks out a bit. Are there packages which would want to _conflict_ `base-no-nativecode` as opposed to _depend_ on `base-ocamlopt` or `base-native`?

kit-ty-kate:
Superseded by #17541

PR #16758, opened 2020-07-04
fix colombe 0.1.0

hannesm:
I got a bit tired of the CI issue:
```
         test alias test/runtest (exit 125)
 (cd _build/default/test && ./test.exe)
 test.exe: internal error, uncaught exception:
           Alcotest__Core.Registration_error("Duplicate test name: requests")
```

so, it is not compatible with the newest and shiniest alcotest

//cc @dinosaure (but I don't really expect this to be of much interest, since there are newer releases of colombe -- I just repeatedly saw it in tls releases)


kit-ty-kate:
I was on my way to fixing it, I got tired as well ^^"

Thanks a lot!

PR #16765, opened 2020-07-05
Package vue-jsoo.0.1

ocpmax:
### `vue-jsoo.0.1`
Binding of Vue_js

---
* Homepage: https://gitlab.com/o-labs/vue-jsoo
* Source repo: git://gitlab.com/o-labs/vue-jsoo
* Bug tracker: https://gitlab.com/o-labs/vue-jsoo/-/issues

---

### DIFF
--- a/packages/vue-jsoo/vue-jsoo.0.1/opam
+++ b/packages/vue-jsoo/vue-jsoo.0.1/opam
@@ -12,7 +12,10 @@ depends: [
   "lwt"
 ]
 depopts: [
-  "js_of_ocaml-tyxml" {>= "3.1.0"}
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Looks good, thanks a lot! I removed the constraints away from the depopts field because it doesn't have the meaning people usually want (i.e. the package with be detected as an optional dependency only if js_of_ocaml-tyxml >= 3.1.0 is installed, if < 3.1.0 is installed then your package will still be installed but will fail since it is not compatible with this version)

Could you return the fix in the opam file in the source repository?

ocpmax:
Thanks for the review and the fix! I pushed it also in the source repository!

PR #16767, opened 2020-07-05
Package letters.0.1.0

jerben:
### `letters.0.1.0`
Client library for sending email over SMTP
Simple to use SMTP client implementation for ocaml

---
* Homepage: https://github.com/oxidizing/letters/
* Source repo: git+https://github.com/oxidizing/letters.git
* Bug tracker: https://github.com/oxidizing/letters/issues

---

### DIFF
--- a/packages/letters/letters.0.1.0/opam
+++ b/packages/letters/letters.0.1.0/opam
@@ -21,6 +21,7 @@ depends: [
   "alcotest" {>= "1.1.0" & with-test}
   "alcotest-lwt" {>= "1.1.0" & with-test}
   "yojson" {>= "1.7.0" & with-test}
+  "odoc" {with-doc}
 ]
 build: [
   ["dune" "subst"] {pinned}
@@ -32,7 +33,6 @@ build: [
     "-j"
     jobs
     "@install"
-    "@runtest" {with-test}
     "@doc" {with-doc}
   ]
 ]
--- a/packages/letters/letters.0.1.0/opam
+++ b/packages/letters/letters.0.1.0/opam
@@ -1,0 +1,46 @@
+opam-version: "2.0"
+synopsis: "Client library for sending email over SMTP"
+description: "Simple to use SMTP client implementation for ocaml"
+maintainer: ["Miko Nieminen <miko.nieminen@iki.fi>"]
+authors: ["Miko Nieminen"]
+license: "MIT"
+homepage: "https://github.com/oxidizing/letters/"
+doc: "https://oxidizing.github.io/letters/"
+bug-reports: "https://github.com/oxidizing/letters/issues"
+depends: [
+  "ocaml" {>= "4.03.0"}
+  "dune" {>= "2.3"}
+  "mrmime" {= "0.3.0"}
+  "colombe" {= "0.3.0"}
+  "sendmail-lwt" {= "0.3.0"}
+  "fmt" {>= "0.8.8"}
+  "x509" {>= "0.9.0"}
+  "ptime" {>= "0.8.5"}
+  "lwt" {>= "5.2.0"}
+  "fpath" {>= "0.7.0"}
+  "alcotest" {>= "1.1.0" & with-test}
+  "alcotest-lwt" {>= "1.1.0" & with-test}
+  "yojson" {>= "1.7.0" & with-test}
+]
+build: [
+  ["dune" "subst"] {pinned}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
+dev-repo: "git+https://github.com/oxidizing/letters.git"
+url {
+  src: "https://github.com/oxidizing/letters/archive/0.1.0.tar.gz"
+  checksum: [
+    "md5=37c789a5ff830b38b48a76b3bb416718"
+    "sha512=192016aedd336844524c2f8964aac1cb909442fd79ce74561994941dfaa0efb1c0ad8b63c272a4846a8607678be08b1d6660288826272ce11f376c91cce51237"
+  ]
+}


packages/letters/letters.0.1.0/opam
```diff
@@ -0,0 +1,42 @@
+opam-version: "2.0"
+synopsis: "Client library for sending email over SMTP"
+description: "Simple to use SMTP client implementation for ocaml"
+maintainer: ["Miko Nieminen <miko.nieminen@iki.fi>"]
+authors: ["Miko Nieminen <miko.nieminen@iki.fi>"]
+homepage: "https://github.com/oxidizing/letters/"
+bug-reports: "https://github.com/oxidizing/letters/issues"
+depends: [
+  "ocaml" {>= "4.03.0"}
+  "dune" {>= "2.0"}
```
kit-ty-kate, line 10: Your `dune-project` file indicates this package requires at least dune 2.3

packages/letters/letters.0.1.0/opam
```diff
@@ -0,0 +1,42 @@
...
+  "mrmime" {= "0.3.0"}
+  "colombe" {= "0.3.0"}
+  "sendmail-lwt" {= "0.3.0"}
+  "fmt" {= "0.8.8"}
```
kit-ty-kate, line 14: These constraints seem rather strict. Your project seems rather useful, but potential users might get a bit annoyed if they can't have their up-to-date library and install your project. Opam only allows one version to be installable in a switch.

Could you consider changing all the `=` constraints to a `>=`? Our CI will check your package on future releases of the dependencies. We modify the constraints accordingly it is effectively broken with a new release.

packages/letters/letters.0.1.0/opam
```diff
@@ -0,0 +1,42 @@
...
+  "x509" {>= "0.9.0"}
+  "ptime" {>= "0.8.5"}
+  "lwt" {>= "5.2.0" & < "6.0.0"}
```
kit-ty-kate, line 17: Are you sure this is going to break with lwt 6.0? Could you remove the upper-bound constraint if you are not sure? Major versions of Lwt tends to be well tested.

packages/letters/letters.0.1.0/opam
```diff
@@ -0,0 +1,42 @@
...
+  "fpath"
+  "alcotest" {>= "1.1.0" & with-test}
+  "alcotest-lwt" {>= "1.1.0" & with-test}
+  "yojson" {>= "1.7.0" & < "2.0.0" & with-test}
```
kit-ty-kate, line 21: Same comment here I suppose

packages/letters/letters.0.1.0/opam
mikonieminen: I didn't realize that, thanks for pointing out.

packages/letters/letters.0.1.0/opam
mikonieminen: I was playing a bit and testing how dependency constraints work. I'll remove all others except `mrmime`, `colombe` and `sendmail-lwt`. APIs of those three aren't very stable state so fixing versions like this is pretty much needed.

packages/letters/letters.0.1.0/opam
mikonieminen: Fixing with next time we push.

packages/letters/letters.0.1.0/opam
mikonieminen: Fixing with next time we push.

mikonieminen:
I need to wait for @jerben to push my fixes, because I haven't been able to make `opam publish` work. For some reasone, I cannot give 2FA to it, it simply keeps asking for the code again and again.

packages/letters/letters.0.1.0/opam
mikonieminen: Fixed.

packages/letters/letters.0.1.0/opam
mikonieminen: Fixed with the exceptions I commented above.

packages/letters/letters.0.1.0/opam
mikonieminen: Fixed.

packages/letters/letters.0.1.0/opam
mikonieminen: Fixed.

SGrondin:
@mikonieminen I had the same issue with `opam-publish`. Installing LastPass Authenticator on my phone, then it worked once I set up 2FA with an app (in the Github settings) rather than SMS.

mseri:
Now the failure is because the package does not provide the `url` field.
Since you use dune, you could also try to release with `dune-release` that does lots of the piping and safety checks for you

mikonieminen:
@SGrondin Originally I was using Google Authenticator for 2FA, but that didn't work. Then I switched to SMS based 2FA and that does not seem to work either. I need to investigate a bit closer once I'm done with this PR.

@mseri I wanted to use just dune, but unfortunately executing tests are failing because I would need non-released version of `colombe` for running tests. The problem with tests is that I run those against *ethereal.email* service (the SMTP server is actually broken, but `colombe`'s maintainer was nice enough to fix the implementation to allow this kind of broker server, unfortunately this fix is not released yet). Additionally, I wasn't sure if it is OK to run test that make requests to outside world and I wasn't sure what to add in `dune-project` so that the opam file's build would not run tests.

mikonieminen:
If I manage to get my 2FA to work, can I somehow update this PR even when it was created by @jerben?

mikonieminen:
I'm starting to be a bit lost here. What ever I change, the build fails. Any hints what exactly is causing this? The failure is:
```
Processing 169/170: [letters: dune subst]
+ /home/opam/.opam/4.10/bin/dune "subst" (CWD=/home/opam/.opam/4.10/.opam-switch/build/letters.0.1.0)
Processing 169/170: [letters: dune build]
+ /home/opam/.opam/4.10/bin/dune "build" "-p" "letters" "-j" "72" "@install" (CWD=/home/opam/.opam/4.10/.opam-switch/build/letters.0.1.0)
- Error: I don't know about package letters (passed through -p)
[ERROR] The compilation of letters failed at "/home/opam/.opam/4.10/bin/dune build -p letters -j 72 @install".

#=== ERROR while compiling letters.0.1.0 ======================================#
# context              2.0.7 | linux/x86_64 | ocaml-base-compiler.4.10.0 | pinned
# path                 ~/.opam/4.10/.opam-switch/build/letters.0.1.0
# command              ~/.opam/4.10/bin/dune build -p letters -j 72 @install
# exit-code            1
# env-file             ~/.opam/log/letters-23-2e0e88.env
# output-file          ~/.opam/log/letters-23-2e0e88.out
### output ###
# Error: I don't know about package letters (passed through -p)
```

mseri:
Yes, your opam file is missing the url field. You should see a warning from opam-publish mentioning a virtual package when you try to submit. Ofter that happens when the repository does not contain a tag for the release

mikonieminen:
Thanks, for the explanation. I ran locally `dune-release` and `opam lint` and every other thing I managed to think of and didn't see any issues. I have also tagged the repo as expected, but since I've been modifying the release, I've been moving the tag. I know this is bad thing to do, but not sure what would be the recommended process when releasing opam packages and fixing issues raised during the review. Additionally, I tagged the repo using annotated tag, in some cases this might also be an issue, depending on how you check for tags.

@jerben will check his env that he has fetched/pulled tags correctly. He cannot do it right now, but I would expect in an hour or so.

mseri:
It is fine to modify the tag when it is not yet released, don’t worry, it happens all the time. I don’t know why it is not picking it properly. The easiest thing is to add it by hand if you have the url and the checksum.

It is worth disabling the tests on the release if they cannot succeed anyways. In that way you are likely able to use dune-release which will also upload an archive for you in the release

packages/letters/letters.0.1.0/opam
mseri: Even though it looks correct on the branch (https://github.com/oxidizing/letters/blob/0.1.0/dune-project#L21), In the submitted version it is back to 2.0.

packages/letters/letters.0.1.0/opam
```diff
@@ -0,0 +1,42 @@
...
+description: "Simple to use SMTP client implementation for ocaml"
+maintainer: ["Miko Nieminen <miko.nieminen@iki.fi>"]
+authors: ["Miko Nieminen <miko.nieminen@iki.fi>"]
+homepage: "https://github.com/oxidizing/letters/"
+bug-reports: "https://github.com/oxidizing/letters/issues"
+depends: [
+  "ocaml" {>= "4.03.0"}
+  "dune" {>= "2.0"}
+  "mrmime" {= "0.3.0"}
+  "colombe" {= "0.3.0"}
```
mseri, line 12: Sorry, one last question. Can we relax these three bounds with a `>=`? You mentioned that there are incoming fixes for example to colombine which you will benefit from/

mseri: Ignore my comment, I have just seen your response in the other thread

mseri:
The url field is there now. Let's wait for the CI!

jerben:
After the tag was moved I didn't move it locally as well so I ran `opam publish` on the wrong commit. Should be fixed now!

packages/letters/letters.0.1.0/opam
mikonieminen: The API of `colombe` and I think also `mrmime` changes with every release as it is very early release and I had some head scratching why for example `facteur` (example app using `colombe`) didn't work until I realized that it needed old version of `colombe`. That's why I fixed these version now, but I suppose I could say something like `{ >= "0.3.0" & < "0.4.0" }`.

packages/letters/letters.0.1.0/opam
mikonieminen: @jerben pushed first wrong version and then moment after the correct one that should match with the release branch. So now you should have the correct one.

mikonieminen:
Finally all tests passed. There was also a small glitch with SUSE package installation, but rebuilding the step seemed to fix it.

mseri:
Thanks for the patience. Feel free to announce it on discuss.ocaml.org, we have community category and announcements tag for this purpose

jerben:
Thanks a lot for the help releasing the package :slightly_smiling_face:

mikonieminen:
Likewise, thanks for the patience and especially big thanks for all the help.

packages/letters/letters.0.1.0/opam
kit-ty-kate: `{ >= "0.3.0" & < "0.4.0" }` would have been a bit better.
In any case, please note that strict requirements are just going to annoy your users and you in the long run. New `mrmime`/`colombe`/... packages are checked in our CI for breakage, and we or the package maintainers (in this case @dinosaure) do the job of updating the constraints if needed.

PR #16779, opened 2020-07-07
Coq 8.11.2: enable flambda if the compiler supports it

Blaisorblade:
Code taken from
https://github.com/Blaisorblade/opam-overlay/blob/master/packages/coq/coq.8.11.2%2Bflambda-byte/opam#L36

Discussion on this change is on
https://coq.discourse.group/t/install-notes-on-coq-and-ocaml-versions-configuration/713/7.

- I'd like to wait for @ejgallego's feedback on merging this. To make this clear, I'm marking it as a draft PR.
- If approved, I'm happy to backport it as far back as 8.10.1 (or as appropriate), as long as we can also disable native-compute on Mac on those versions (to avoid risking regressions due to the flambda/native-compute interaction).

https://github.com/Blaisorblade/opam-overlay/tree/master/packages/coq
- [ ] Add `ocaml:version >= "4.07"` as proposed by @erikmd:
https://github.com/coq-community/docker-coq/issues/10#issuecomment-655793711

### DIFF
--- a/packages/coq/coq.8.11.2/opam
+++ b/packages/coq/coq.8.11.2/opam
@@ -33,6 +33,7 @@ build: [
     "-libdir" "%{lib}%/coq"
     "-datadir" "%{share}%/coq"
     "-coqide" "no"
+    "-flambda-opts" "-O3 -unbox-closures"
     "-native-compiler" {os = "macos"} "no" {os = "macos"}
   ]
   [make "-j%{jobs}%"]


ejgallego:
The options are ignored when no-flambda compiler is used, so they are relatively safe. On the other hand, they will have the compilation of Coq eat a lot of memory due to some OCaml bugs in OCaml < 4.07 , that's the only risk I can see.

AFAIU the problem with flambda + native should appear equally in all platforms.

Blaisorblade:
> OCaml bugs in OCaml < 4.07 , that's the only risk I can see.

You'd have to select explicitly an flambda variant of OCaml < 4.07. That seems an acceptable risk.

In the future, configure could detect that combination and ignore `-flambda` options.
 
> AFAIU the problem with flambda + native should appear equally in all platforms.

Right, says https://github.com/coq/coq/issues/11107 (I just fixed the misleading title).

Without this, we'd want another fix for https://coq.zulipchat.com/#narrow/stream/237663-coq-community-devs.20.26.20users/topic/flambda-using.20Docker-Coq.20images.

So I asked on https://github.com/coq/coq/issues/12564 if we can disable native-compute by default.

Blaisorblade:
Closing for now; we decided to use "dummy packages" as flags on Zulip. Thread (login required):

https://coq.zulipchat.com/#narrow/stream/237663-coq-community-devs.20.26.20users/topic/flambda-using.20Docker-Coq.20images

PR #16788, opened 2020-07-09
Fix ppxlib.0.13.0 dune's constraint

NathanReb:
The dune-project requires dune 1.11 but the opam file wasn't properly updated!


mseri:
Thanks

PR #16789, opened 2020-07-09
Package yices2_bindings.0.1

disteph:
### `yices2_bindings.0.1`
Ocaml bindings for yices2
Ocaml bindings for yices2

---
* Homepage: https://github.com/SRI-CSL/yices2_ocaml_bindings
* Bug tracker: https://github.com/SRI-CSL/yices2_ocaml_bindings/issues

---

### DIFF
--- a/packages/yices2_bindings/yices2_bindings.0.1/opam
+++ b/packages/yices2_bindings/yices2_bindings.0.1/opam
@@ -29,7 +29,7 @@ install: [
 url {
   src: "https://github.com/SRI-CSL/yices2_ocaml_bindings/archive/0.1.tar.gz"
   checksum: [
-    "md5=69c2bb4868c1b5d774f14115b4e9f649"
-    "sha512=9fbd04b9354090d8145f5342a407d231313b9f2dfc64be0406332a4decd2fa7bb09ec9e4a7d77ceeadaf0255ebac0459914a4f50ad4857d205281fb14505409c"
+    "md5=c5097d0fd1e22b601594313a1ed85238"
+    "sha512=89310ced285fdea4bd3111bb83f5384b7cf1bed0e739e067fec7c133f480275b9e872b9910bfbbfb44f7a5f3122e49cc1764ccacf56c75427c3471c837d322bc"
   ]
 }
--- a/packages/yices2_bindings/yices2_bindings.0.1/opam
+++ b/packages/yices2_bindings/yices2_bindings.0.1/opam
@@ -1,0 +1,38 @@
+opam-version: "2.0"
+synopsis: "Ocaml bindings for yices2"
+description: "Ocaml bindings for yices2"
+maintainer: "Stephane Graham-Lengrand <stephane.graham-lengrand@csl.sri.com>"
+authors: "Stephane Graham-Lengrand <stephane.graham-lengrand@csl.sri.com>"
+license: "GPLv3"
+homepage: "https://github.com/SRI-CSL/yices2_ocaml_bindings"
+bug-reports: "https://github.com/SRI-CSL/yices2_ocaml_bindings/issues"
+depends: [
+  "ocaml" {>= "4.08" & < "4.11"}
+  "ocamlbuild"
+  "ocamlfind"
+  "ctypes"
+  "ctypes-foreign"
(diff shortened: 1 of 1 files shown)


packages/yices2_bindings/yices2_bindings.0.1/opam
```diff
@@ -0,0 +1,38 @@
+opam-version: "2.0"
+synopsis: "Ocaml bindings for yices2"
+description: "Ocaml bindings for yices2"
+maintainer: "Stephane Graham-Lengrand <stephane.graham-lengrand@csl.sri.com>"
+authors: "Stephane Graham-Lengrand <stephane.graham-lengrand@csl.sri.com>"
+license: "GPLv3"
+homepage: "https://github.com/SRI-CSL/yices2_ocaml_bindings"
+bug-reports: "https://github.com/SRI-CSL/yices2_ocaml_bindings/issues"
+depends: [
+  "ocaml" {>= "4.08" & < "4.11"}
```
kit-ty-kate, line 10: are you certain this package is not compilable with OCaml 4.11? If you're not, could you remove the upper-bound constraint? Our CI will test it for you.

disteph: I haven't tested with OCaml 4.11; I'm just being conservative when it hasn't been tested.
And I haven't tested with 4.11 because I thought it wasn't out yet. Is it? Opam only shows me alpha/beta/trunk versions among the available switches. Should I use one of those? Which?

kit-ty-kate: The latest beta is fine, but we have it running in our CI already. I loosen up the constraint, we'll see what it says.

kit-ty-kate:
I also removed the `remove` field which is unnecessary since opam 2.0 (unless you want to remove files that came after the installation), and tagged `ocamlfind` and `ocamlbuild` as build dependency since you seem to use them only during build time and these are compatible with opam's build tag.

I'm also wondering what is the difference with the already existing `yices2` package. I'm unsure of the difference between the two, but maybe you might want to merge the two projects or something? I don't know, just putting that out-there, it's not mandatory for the merge.

kit-ty-kate:
Oh.. It seems that you have ocamlbuild as a runtime dependency. Is that expected? It is listed in your META files but I can't find any other use elsewhere.

disteph:
> Oh.. It seems that you have ocamlbuild as a runtime dependency. Is that expected? It is listed in your META files but I can't find any other use elsewhere.

Ah, no. I just didn't distinguish build dependencies and runtime dependencies. You're right, ocamlbuild shouldn't be in the META files.

> I also removed the `remove` field which is unnecessary since opam 2.0 (unless you want to remove files that came after the installation), and tagged `ocamlfind` and `ocamlbuild` as build dependency since you seem to use them only during build time and these are compatible with opam's build tag.

That's right.

> I'm also wondering what is the difference with the already existing `yices2` package. I'm unsure of the difference between the two, but maybe you might want to merge the two projects or something? I don't know, just putting that out-there, it's not mandatory for the merge.

`yices2_bindings` is developed by the Yices team to closely follow the Yices development, and for instance contains bindings for the latest Yices API functions (e.g. term traversals) that the `yices2` package doesn't have. I don't know to what extent that package is maintained. I suppose we could have done a fork, but we just wrote those bindings from scratch.

disteph:
> > Oh.. It seems that you have ocamlbuild as a runtime dependency. Is that expected? It is listed in your META files but I can't find any other use elsewhere.
> 
> Ah, no. I just didn't distinguish build dependencies and runtime dependencies. You're right, ocamlbuild shouldn't be in the META files.

Actually, now I don't know whether it should or should not appear in the ocamlfind's META file.

The ocamlfind manual, when describing the META file, says
> The variable "requires" specifies the list of required packages.
without saying whether this is for build time dependencies or runtime dependencies; in doubt, I listed ocamlfind. Should I? should I not?

kit-ty-kate:
Findlib (META files) only deals with libraries required at runtime in order to use the library. ocamlbuild is not one of them.

kit-ty-kate:
Thanks a lot!

PR #16791, opened 2020-07-09
Update the GSL depext for Debian

Chris00:
libgsl0-dev is a virtual package provided by libgsl-dev and [only available in Debian stable](https://packages.debian.org/search?keywords=libgsl0&searchon=names&suite=stable&section=all).


mseri:
Can you move this change to conf-gsl.1-1 or conf-gsl.2? This way it will not break for users of older versions

Chris00:
@mseri Done.

mseri:
Thanks!

PR #16795, opened 2020-07-10
Release slap v4.1.1

akabe:
Slap (https://github.com/akabe/slap) now supports ocaml 4.11 thanks to @kit-ty-kate.

### DIFF
--- a/packages/slap/slap.4.1.1/opam
+++ b/packages/slap/slap.4.1.1/opam
@@ -18,10 +18,6 @@ build: [
   ["ocaml" "setup.ml" "-doc"] {with-doc}
 ]
 install: ["ocaml" "setup.ml" "-install"]
-remove: [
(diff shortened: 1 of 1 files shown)


packages/slap/slap.4.1.1/opam
```diff
@@ -0,0 +1,49 @@
...
+    "-configure"
+    "--prefix"
+    prefix
+    "--disable-ppx" {ocaml:version < "4.02"}
+  ]
+  ["ocaml" "setup.ml" "-build"]
+  ["ocaml" "setup.ml" "-doc"] {with-doc}
+]
+install: ["ocaml" "setup.ml" "-install"]
+remove: [
```
mseri, line 21: This is no longer needed with opam2, it keeps track of installed files on its own. I will remove it before merging, can you upstream the change?

PR #16800, opened 2020-07-11
Add nixos distribution to sqlite3 depexts

akotulski:
opam2nix tool uses `os-distribution = "nixos"` for depexts:
https://github.com/timbertson/opam2nix#how-do-opam-depexts-work


kit-ty-kate:
Thanks!

PR #16808, opened 2020-07-13
irmin-indexeddb: add a conflict with irmin's master

samoht:
/cc @talex5 

Happy to fold that library into `irmin` repository to ease the maintenance if you like.


samoht:
https://github.com/mirage/irmin/pull/1030 will break irmin-indexeddb compilation with

```
# File "lib/branch_store.ml", line 43, characters 10-34:
# Error: This function has type
#          'a Irmin.Type.t ->
#          (string -> ('a, [ `Msg of string ]) result) Irmin.Type.staged
#        It is applied to too many arguments; maybe you forgot a `;'.
```

These serialisation functions are now staged and should be using `Irmin.Type.{stage,unstage}`

talex5:
> Happy to fold that library into irmin repository to ease the maintenance if you like.

That would be great!

mseri:
Thanks

PR #16811, opened 2020-07-14
Coq 8.9: fix camlp5 dependency

RalfJung:
When camlp5 8.00~alpha01 got added to the repo, it got picked up automatically by our CI through `opam upgrade` (seems a bit weird for alpha versions to be released like that), breaking the build of Coq 8.9. This adjusts the Coq 8.9 package to fix that.

(Newer versions of Coq lucky enough to not depend on camlp5 any more.)

PR #16812, opened 2020-07-14
also fix Coq 8.8 camlp5 dependency

RalfJung:
Same as https://github.com/ocaml/opam-repository/pull/16811 but for Coq 8.8 (turns out we need that one, too).


mseri:
I am surprised that coq was not picked up as a revdep by the CI when we tested camlp5. I wonder if it depends from the ocaml upper bounds...

mseri:
I think we will have to add it to all the old versions of coq by the way

PR #16814, opened 2020-07-14
Fix camlp5 also for coq < 8.8

mseri:
This should fix potential incompatibilities with the latest, breaking, camlp5. Ping @RalfJung 

### DIFF
--- a/packages/coq/coq.8.3/opam
+++ b/packages/coq/coq.8.3/opam
@@ -14,7 +14,7 @@ build: [
 ]
 depends: [
   "ocaml" {>= "3.10.2" & < "4.02.0"}
-  "camlp5" {!= "5.00"}
+  "camlp5" {!= "5.00" & < "8"}
   "ocamlbuild" {build}
   "num"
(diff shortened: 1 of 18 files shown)


packages/coq/coq.8.3/opam
```diff
@@ -14,7 +14,7 @@ build: [
 ]
 depends: [
   "ocaml" {>= "3.10.2" & < "4.02.0"}
-  "camlp5" {!= "5.00"}
+  "camlp5" {!= "5.00" & < "8"}
```
RalfJung, line 17: Wouldn't `>= 5` be better?

mseri: Not with `>=`, as it would accept `5.00`. If it is compatible with versions `< 5`, then I'd rather keep the upper bound, otherwise I am happy to change to `> "5"`

RalfJung: Ah right... maybe best to just not change it.^^

RalfJung:
LGTM.

PR #16815, opened 2020-07-14
efl: add version 1.24.0

axiles:


kit-ty-kate:
The `remove` field is unnecessary since opam 2.0. Sorry for the time it took to handle this PR. Thanks!

PR #16816, opened 2020-07-14
ocaml-variant.4.12.0+trunk*: Use --disable-warn-error + Factorize the build rule

kit-ty-kate:
https://github.com/ocaml/ocaml/pull/9625 brings the ability for the compiler during its dev phase, to not pass `-Werror` to the C linker, which is not enabled on releases. This option should not be given to the C linker by default.

This PR also reorganize the build rules to avoid copy/pasting.


kit-ty-kate:
cc @Octachron what do you think?

Octachron:
The factorization change looks good and I agree that the `+trunk` packages should not fail due to C compiler warnings.

PR #16824, opened 2020-07-15
Add bitmasks.1.2.0

dra27:


packages/bitmasks/bitmasks.1.2.0/opam
```diff
@@ -0,0 +1,28 @@
...
+license: "BSD-3-clause with OCaml linking exception"
+homepage: "https://metastack.github.io/bitmasks"
+dev-repo: "git+https://github.com/metastack/bitmasks.git"
+bug-reports: "https://github.com/metastack/bitmasks/issues"
+build: [
+  [ "dune" "build" "-p" name "-j" jobs "@install" "@runtest" {with-test}]
+  [ make "doc" ] {with-doc}
+]
+depends: [
+  "ocaml" {< "4.11.0"}
```
kit-ty-kate, line 13: I'm a bit surprised to see this restriction. Previous versions worked fine on OCaml 4.11 and the restriction isn't in the source repository either. Could it be removed?

dra27: It’s the different between “builds with OCaml 4.11” and “works with 4.11” - you can build it, but you can’t do anything with it. In fact, I should probably push a constraint for 1.1.1 to < 4.07.0 for the same reason!

(it has to match all the functions of Set.S)

dra27: (https://github.com/metastack/bitmasks/pull/4 is ready to lift the restriction)

kit-ty-kate:
Thanks!

PR #16827, opened 2020-07-16
[new release] clangml.4.2.0, refl 0.2.1, clangml-transforms 0.26

thierry-martinez:
This PR provides a new clangml release, and an update for its dependency refl and its rev-dependency clangml-transforms.

- Support for Clang/LLVM 10.0.1

- Support for all C/C++ Clang attributes (including `_Alignas` (C) and `alignas` (C++) attributes, that have been requested by Damien Rouhling).  See the auto-generated `bootstrap/attributes.ml`.

- class base specifiers now expose a full `qual_type` instead of a mere `ident` string.

- `ident_ref` now carries a `template_arguments` parameter.

- `function_decl` now exposes `inline_specified` and `inlined` fields.

- `function_type` now exposes `ref_qualifier` field.

- `TypeLoc` are used broaderly to compute `qual_type`: in particular, parameters in function types have now correct names.  The type `type_loc` and the functions `Clang.Decl.get_type_loc` and `Clang.Parameter.get_type_loc` are kept for compatibility, but `qual_type` should contain all the informations.

- `Clang.Lazy` exposes a lazy AST, that is to say that each `desc` field is a lazy value that is computed on demand. This is useful to explore large ASTs efficiently (note that Clang parsing itself can still be slow; the lazy part only concerns the conversion into the `Clang.Lazy.Ast` datatypes). Modules `Clang.Ast`, `Clang.Expr`, `Clang.Stmt`, `Clang.Type`, ..., have their lazy counter-parts: `Clang.Lazy.Ast`, `Clang.Lazy.Expr`, `Clang.Lazy.Stmt`, `Clang.Lazy.Type`, etc.

- The common signature exposed by every AST node module (`Clang.Expr`, `Clang.Stmt`, ...) now declares a `hash` function and and a `Hashtbl` module.

- `Clang.format_diagnostics` now prints presumed locations by default and the behavior can be changed through the optional `?options` argument. The implementation now relies on the new `Clang.pp_diagnostic` which is an OCaml reimplementation of `clang_formatDiagnostic` from libclang.

- `Clang.Expr.parse_string` parses strings as C expressions.  (suggested by Damien Rouhling)

- `Clangml_printer` has now moved to `Clang.Printer`. The package `clangml.printer` still exists for compatibility and reexposes the `Clang.Printer` interface.

### DIFF
--- a/packages/clangml-transforms/clangml-transforms.0.26/opam
+++ b/packages/clangml-transforms/clangml-transforms.0.26/opam
@@ -9,10 +9,11 @@ bug-reports: "https://gitlab.inria.fr/memcad/clangml-transforms/issues"
 depends: [
   "ocaml" {>= "4.03.0" & < "4.12.0"}
   "dune" {>= "1.11.0"}
-  "clangml" {>= "4.1.0" & < "4.2.0"}
+  "clangml" {>= "4.2.0"}
   "dolog" {>= "4.0.0"}
   "traverse" {>= "0.2.0"}
-  "refl" {>= "0.1.0"}
+  "refl" {>= "0.2.1"}
+  "odoc" {with-doc & >= "1.5.1"}
 ]
 build: [
   ["dune" "subst"] {pinned}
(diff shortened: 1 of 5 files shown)


thierry-martinez:
There is still a failure on freebsd because `conf-libclang` does not know how to install libclang on freebsd. I don't see which package to install now that `devel/clang` has been removed. I think this can wait a future PR.

mseri:
I think it is now integrated with llvm, at least `devel/llvm10` includes `clang`

thierry-martinez:
Thank you, @mseri . I looked at this package for the previous release of clangml: I don't know how but I missed that the library files were present indeed! I give it a try. Thanks again !

thierry-martinez:
FreeBSD is green as well now!

mseri:
Thanks!

PR #16830, opened 2020-07-16
package "funfields" version 0.04

fccm:
Rename package "funbits" to "funfields", and add new version 0.04 of this package.


avsm:
In general, we do not accept the deletion of older packages in the opam-repository, as this may break dependent packages. Funbits.0.02 has been in the opam-repository for over a year, so this isn't a short time window either.

Would you consider modifying this PR to just marking the older funbits packages with a `messages` field to indicate it has been renamed to funfields? It's fine to modify the older package metadata to reflect the new homepage and so on, to avoid broken links -- it's just dropping the `funbits` opam package name or modifying its distfile that is problematic.

fccm:
This is not a big package. Only 3 persons installed it, so I think IMHO that we don't have to worry that much about renaming it.

avsm:
I'm afraid the "3 people installing it" number isn't accurate -- we just don't keep these kind of statistics.  However, we have had several instances of projects breaking since they take a dependency on a small package, which subsequently disappears from the opam-repository.

The point of publishing multiple versions in opam-repo of a library is that those versions actually stick around -- if you rename something, then the only way the project depending on it can continue to work is by pinning to an older git revision of opam-repo, and missing out on other metadata fixes.

So yes, please do put back the funbits package. I only ask because time is indeed precious -- and that's why it's best not to remove packages without very good reason and risk breaking other people's code (e.g. for serious security or licensing violations).

avsm:
I've discussed this offline with @fccm and there is a good reason for the renaming, so this can be allowed.

fccm:
OK, I just added back an empty package `funbits.0.0.0` as yo suggested, with `synopsis` and `messages` fields defined as:
"Deprecated: This package was renamed into \"funfields\""

mseri:
Thanks for looking into it, and thanks for the release

Issue #16689, opened 2020-06-22
Checking for external dependency installed via coreutils


maroneze:
I'd like to add a package `conf-timeout` to check that the `timeout` command is installed in the machine, so that we can use it as a depext in our OCaml packages.

However, when using `conf-time` as inspiration for writing `conf-timeout`, I realized that in all systems I know of, `timeout` is installed via the `coreutils` package.

Therefore I wonder: is it useful to add such a package to opam-repository? Or can we assume that `coreutils` is always installed in the system? For instance, is it already an implicit dependency of installing opam? I think any VM/Docker having `opam` should probably already have `coreutils` installed, but I'm not entirely sure. From what I could gather, even FreeBSD and NetBSD have a `timeout` command, even if it is not POSIX; but I couldn't confirm whether OpenBSD does.

However, it seems that macOS does not contain a `timeout` command by default. In this case, a `conf-*` package might make sense. If you have more experience about the issue, could you please share it?

maroneze:
By the way, a related question: when creating such a small `conf-` package, should I assign myself as maintainer or someone else?

mseri:
If also in *BSD is part of some utils package, it could more generally be a `conf-coreutils` package, or something similar. Let's wait a few days, maybe somebody with OpenBSD knowledge can tell you about it.

As maintainer you can put yourself. For conf packages it's not really requiring much (if any) effort

maroneze:
Ok, I'll wait and create if it makes sense. My macOS colleague told me that Homebrew also has a `coreutils` package, so it should help macOS users.

maroneze:
Well, I was preparing a package, and I stumbled upon a very basic error: how to detect that `coreutils` is installed? Other than using a distribution-specific process (e.g. `dpkg` on Debian, `brew list coreutils` on Homebrew, etc), it seems that there is no generic way to check for them (e.g. `coreutils --version`). I could test a single utility (`timeout`, in my case), but it would not strictly guarantee that it is coreutils that is installed. In practice, I couldn't find any distribution where `timeout` was its own standalone package, so it might work, but then, would it be better than simply writing a `conf-timeout` package which uses `coreutils` as depexts?

mseri:
[...]

PR #16832, opened 2020-07-17
[conf-timeout] package to detect if timeout is available

maroneze:


mseri:
Cfr https://github.com/ocaml/opam-repository/issues/16689

mseri:
Not sure about the centos fail. I'll wait for @kit-ty-kate to have a look before doing anything

maroneze:
I just tried a centos Docker and got the same error. But what's worse, is that `which` is not available by default. I'll change the package to not require `conf-which`, using instead `command -v`, as recommended by [SO](https://stackoverflow.com/questions/592620/how-can-i-check-if-a-program-exists-from-a-bash-script).

maroneze:
It seems CentOS is a bit finnicky: it has both packages `coreutils` and `coreutils-single`, which conflict with each other. I don't think opam has a way to say "check if either of these is installed, otherwise try installing one", so I'm tempted to simply not check at all whether coreutils is available in CentOS (which should be the case most of the time), deferring the check to `command` itself. For instance, I just tried uninstalling `coreutils-single`, and I got an error saying that this would remove `systemd`, a protected package. So there's very little chance of that happening.

mseri:
Thanks, this has waited far too long

PR #16834, opened 2020-07-17
conf-binutils 0.3

gitoleg:
This PR introduces a new version of `conf-binutils` and constrains all `bap` packages to the old ones. The only package beyond `bap` universe that uses `conf-binutils` is `clangml`, and looks like it doesn't depend on our interface, so we leave it untouched.

What's new:

### Improved Search
We search for files that have `cxxfilt`, `objdump`, and `readelf` suffixes in the `brew` cellar, if one available, and in the `PATH` constituents. Instead of using tools from `findutils`, `which`, and `mdfind`, as we did before, we search for the matching files manually, using the OCaml `Sys.readdir` function.

The problem is that on macOS with `brew` we can't use which utility for the search: `binutils` formula is marked as `keg-only` and also can shadow system files, so even `brew link --force` doesn't link `binutils` somewhere in the `$PATH`. That's why we have problems with installing bap on macOS: `which gobjdump` doesn't return a path to the file anymore. With this PR this problem will be solved.

### Updated Opam Variables
Previously, we were creating two `opam` config variables per each tool, e.g. `objdump` and `objdump-targets`. But since all that we do with them on the bap side is concatenating them in one list, there is no need in several variables for one tool. Now we add only one, e.g `objdumps` or `cxxfilts` that contains a list of absolute paths to all discovered files for the given tool. The preferred tool still can be set via `opam` config or environment variable. In this case, it will be first in the paths list.

See https://github.com/BinaryAnalysisPlatform/opam-repository/pull/51 for more information and discussion.

### DIFF
--- a/packages/bap-cxxfilt/bap-cxxfilt.1.0.0/opam
+++ b/packages/bap-cxxfilt/bap-cxxfilt.1.0.0/opam
@@ -25,7 +25,7 @@ depends: [
   "ocaml"
   "bap-std" {= "1.0.0"}
   "bap-demangle" {= "1.0.0"}
-  "conf-binutils" {>= "0.2"}
+  "conf-binutils" {= "0.2"}
 ]
 synopsis: "A demangler that relies on a c++filt utility"
 url {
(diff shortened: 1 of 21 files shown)


avsm:
Firstly, thank you for linking to the discussion upstream in BAP -- very useful indeed. The opam repository abhors being the home of original OCaml code, but in the case of the binutils-finding script I can see why it's useful here and can make an exception to that guideline.  We're considering various options to split out the depext configuration packages from the rest of the opam repository in the future, and the binutils package provides a useful source of configuration complexity.

In an aside unrelated to this PR, it looks like a couple of BAP packages fail to build on OCaml 4.03 and 4.04. You may want a new PR to constrain those to prevent the build errors.

PR #16835, opened 2020-07-17
Package pa_ppx.0.01

chetmurthy:
### `pa_ppx.0.01`
PPX Rewriters for Ocaml, written using Camlp5
This is a collection of PPX rewriters, re-implementing those based on ppxlib
and other libraries, but instead based on Camlp5.  Included is also a collection
of support libraries for writing new PPX rewriters.  Included are:

pa_assert: ppx_assert
pa_ppx.deriving, pa_ppx.deriving_plugins (enum, eq, fold, iter, make, map, ord, sexp, show, yojson):
  ppx_deriving, plugins, ppx_sexp_conv, ppx_deriving_yojson
pa_ppx.expect_test: ppx_expect_test
pa_ppx.here: ppx_here
pa_ppx.import: ppx_import
pa_ppx.inline_test: ppx_inline_test

pa_ppx.undo_deriving: pa_ppx.deriving expands [@@deriving ...] into code; this rewriter undoes that.
pa_ppx.unmatched_vala: expands to match-cases (support library for camlp5-based PPX rewriters)
pa_ppx.hashrecons: support for writing AST rewriters that automatically fills in hash-consing boilerplate
pa_dock: implements doc-comment extraction for camlp5 preprocessors

Many of the reimplementations in fact offer significant enhanced
function, described in the pa_ppx documentation.  In addition, there
is an extensive test-suite, much of it slightly modified versions of
the tests for the respective PPX rewriters.

---
* Homepage: https://github.com/chetmurthy/pa_ppx
* Source repo: git+https://github.com/chetmurthy/pa_ppx.git
* Bug tracker: https://github.com/chetmurthy/pa_ppx/issues

---

### DIFF
--- a/packages/pa_ppx/pa_ppx.0.01/opam
+++ b/packages/pa_ppx/pa_ppx.0.01/opam
@@ -58,6 +58,10 @@ depexts: [
     "perl-string-shellquote"
     "perl-ipc-system-simple"
   ] {os-distribution = "alpine"}
+  [
+    "perl-string-shellquote"
+    "perl-ipc-system-simple"
+  ] {os-distribution = "centos"}
 ]
 
 build: [
--- a/packages/pa_ppx/pa_ppx.0.01/opam
+++ b/packages/pa_ppx/pa_ppx.0.01/opam
@@ -1,0 +1,69 @@
+synopsis: "PPX Rewriters for Ocaml, written using Camlp5"
+description:
+"""
+This is a collection of PPX rewriters, re-implementing those based on ppxlib
+and other libraries, but instead based on Camlp5.  Included is also a collection
+of support libraries for writing new PPX rewriters.  Included are:
+
+pa_assert: ppx_assert
+pa_ppx.deriving, pa_ppx.deriving_plugins (enum, eq, fold, iter, make, map, ord, sexp, show, yojson):
+  ppx_deriving, plugins, ppx_sexp_conv, ppx_deriving_yojson
+pa_ppx.expect_test: ppx_expect_test
+pa_ppx.here: ppx_here
+pa_ppx.import: ppx_import
+pa_ppx.inline_test: ppx_inline_test
+
+pa_ppx.undo_deriving: pa_ppx.deriving expands [@@deriving ...] into code; this rewriter undoes that.
+pa_ppx.unmatched_vala: expands to match-cases (support library for camlp5-based PPX rewriters)
+pa_ppx.hashrecons: support for writing AST rewriters that automatically fills in hash-consing boilerplate
+pa_dock: implements doc-comment extraction for camlp5 preprocessors
+
+Many of the reimplementations in fact offer significant enhanced
+function, described in the pa_ppx documentation.  In addition, there
+is an extensive test-suite, much of it slightly modified versions of
+the tests for the respective PPX rewriters.
+
+"""
+opam-version: "2.0"
+maintainer: "Chet Murthy <chetsky@gmail.com>"
+authors: ["Chet Murthy"]
+homepage: "https://github.com/chetmurthy/pa_ppx"
+license: "BSD-3-Clause"
+bug-reports: "https://github.com/chetmurthy/pa_ppx/issues"
+dev-repo: "git+https://github.com/chetmurthy/pa_ppx.git"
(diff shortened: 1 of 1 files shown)


avsm:
The perl depexts need to be settled before merging this, on Fedora at least (and OpenSUSE ideally).  Would you consider just rewriting the META generation in OCaml to avoid this depext on a new package?

chetmurthy:
I'll make another round of trying to get the depexts right.  I left it to run overnight, only b/c it wasn't making progress towards completion.  I'll add a comment when it's all working, or I've reached a dead-end.

META.pl mentions the two Perl modules, but doesn't actually use them (sheer laziness).  They -are- used in tools/LAUNCH, and are somewhat useful there.  I could eschew them, but would prefer to try to figure out how to get the depexts working, simply because these modules actually make writing perl-scripts that construct complex shell-commands much, much safer.

avsm:
You may find it easier to locally test by using Docker. You can `docker run -it ocaml/opam2:fedora` or `docker run -it ocaml/opam2:opensuse` to get the relevant environment to find the packages in.

chetmurthy:
Ohh yeah, I was gonna ask you that question after I got things working.  That's exactly what I had wished for!  Thank you!

chetmurthy:
All working, I think!

kit-ty-kate:
All JaneStreet packages have the `v` prefix to their version numbers. Could you return the fix in the opam file upstream?

packages/pa_ppx/pa_ppx.0.01/opam
```diff
@@ -0,0 +1,85 @@
...
+maintainer: "Chet Murthy <chetsky@gmail.com>"
+authors: ["Chet Murthy"]
+homepage: "https://github.com/chetmurthy/pa_ppx"
+license: "BSD-3-Clause"
+bug-reports: "https://github.com/chetmurthy/pa_ppx/issues"
+dev-repo: "git+https://github.com/chetmurthy/pa_ppx.git"
+doc: "https://github.com/chetmurthy/pa_ppx/doc"
+
+depends: [
+  "ocaml"       { = "4.10.0" }
```
kit-ty-kate, line 37: This constraint seems a little tight. Could this be changed to `>= "4.10.0" & < "4.11.0"` so that it also takes any future minor versions too? The internal API does not change in these minor versions.

chetmurthy: It's an excellent question!  So ... yes, the code should work for higher-numbered ocaml versions.  Heck, it should probably work for lower-numbered ones (though I haven't tested it, and since I don't really want to test camlp5 8.00 on lower-numbered ocaml versions, that's a non-starter).  But the problem is, I can't build (yet) pa_ppx against ocaml 4.11.0, b/c some of the packages it depends upon (to be fair, most of those deps are for testing purposes, but at least some of them are libraries that actually are -used- by pa_ppx) aren't yet ported to ocaml 4.11.0.

So as soon as pa_ppx can build agaiinst ocaml 4.11.0, I'll be able to put out a new release that bumps that version-number.

But also: there are some files that actually -do- depend on the Ocaml version.  Specifically, I apply pa_ppx (in the manner of `ppx_import`) against the types in `parsing/parsetree.mli` in order to generate pretty-printers and such for the Ocaml AST types.  Since those change from ocaml version to ocaml version, there is a workflow for generating versions of those types for a new version of ocaml.  But that workflow depends on pa_ppx being build-able (with a sort of "dummy" used instead of the generated code for the Ocaml AST types).

Hope that's clear.  Once those supporting packages all work with Ocaml 4.11.0, it will be a straightforward "build, run this, build, run this, build" sequence to update pa_ppx to support it.

chetmurthy: Oh ha, I see now what your'e saying, and will fix.  Sorry sorry.

chetmurthy:
I'm not sure what to do here.  The three failing tests appear to have failed even though all their subsidiary steps have succeeded?

chetmurthy:
Hmmm ... is this what's causing the problem?
```
(Failure "9p error: connection disconnected")
```

kit-ty-kate:
There was an issue in our CI. Sorry for that. Thanks!

PR #16837, opened 2020-07-18
Various fixes

kit-ty-kate:
cc @pirbo for tezos packages (tl;dr: lots of missing test dependencies in the latest versions)
cc @dinosaure I simply removed the tests for eqaf, they're not deterministic
cc @c-cube for the tests in containers


pirbo:
Thank you for all these, I'm taking them upstream...

PR #16838, opened 2020-07-19
Packages js_of_ocaml-webgpu.0.1 and js_of_ocaml-webidl.0.1

tari3x:
This pull-request concerns:
-`js_of_ocaml-webgpu.0.1`: Js_of_ocaml bindings for webgpu
-`js_of_ocaml-webidl.0.1`: Generate js_of_ocaml bindings from webidl definitions

---
* Homepage: https://github.com/tari3x/webgpu
* Bug tracker: https://github.com/tari3x/webgpu

---

### DIFF
--- a/packages/js_of_ocaml-webgpu/js_of_ocaml-webgpu.0.1/opam
+++ b/packages/js_of_ocaml-webgpu/js_of_ocaml-webgpu.0.1/opam
@@ -1,0 +1,25 @@
+opam-version: "2.0"
+synopsis: "Js_of_ocaml bindings for webgpu"
+maintainer: ["Misha Aizatulin <mihhail.aizatulin@gmail.com>"]
+authors: ["Misha Aizatulin <mihhail.aizatulin@gmail.com>"]
+license: "BSD-3-Clause"
+homepage: "https://github.com/tari3x/webgpu"
+bug-reports: "https://github.com/tari3x/webgpu"
+depends: [
+  "dune" {>= "2.2"}
+  "js_of_ocaml-webidl"
+]
+build: [
+  ["dune" "subst"] {pinned}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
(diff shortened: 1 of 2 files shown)


tari3x:
So this adds two packages and one of them depends on the other. Looks like this is not supported? Should I get them published in turn?

mseri:
No, that is supported. There are a number of other issues.

The failures you see are due to the fact that the `opam` files you submitted do not contain the `source` field, that points to the source tarballs and provides the checksums. It seems that `opam-publish` is not picking up the tag `0.1`: I can't help much with that since I usually use `dune-release`. However, you can see if it picked the tag from the output. When `opam-publish` does cannot infer the sources location, it emits a warning about "virtual packages".

Once you add the sources, though, there build will still fail.
From a brief look at the code, `webidl` will fail at least due to the following missing dependencies
```
core async webidl ppx_deriving.runtime js_of_ocaml js_of_ocaml-lwt js_of_ocaml-ppx ppx_jane
```
a number of which seems to be required by `webgpu` as well.

Once those are added, I think we can look at the CI and see if there are further fixes needed.

tari3x:
So I made changes and crated tag 0.2 for that, but that means opam publish created a new pull request: https://github.com/ocaml/opam-repository/pull/16861

Should I drop that one and move tag 0.1 instead?

More questions: 
1. Are dependencies transitive? Is it enough for me to just add them to webidl and make webgpu depend on webidl only?
2. I'm not attached to opam publish, and could try dune-release. Is that going to mess things up since I already started with opam publish?

mseri:
> So I made changes and crated tag 0.2 for that, but that means opam publish created a new pull request: #16861
> 
> Should I drop that one and move tag 0.1 instead?

Either way is fine, it is not a problem for us to close this PR and directly release 0.2. You can do what you prefer

> Are dependencies transitive? Is it enough for me to just add them to webidl and make webgpu depend on webidl only?

They are in principle, unless specifically disable in dune, however is a good practice to list all the required dependencies in each package.

> I'm not attached to opam publish, and could try dune-release. Is that going to mess things up since I already started with opam publish?

No, they play along nicely. Since I see that also 0.2 has the same problems (lack of url for the sources and missing dependencies) I would suggest that you try to use dune-release and see how it goes. If the tag is already there you can do `dune-release -t yourtag` and should pick it up.

tari3x:
Trying dune-release, and it seems to not like dots in package names:
error  3: File format error in 'depends' at line 17, column 23: while expecting pkg-name: Invalid character in package name "ppx_deriving.runtime"

I'll try to go back to "opam release" and see why it's not creating a tarball.

tari3x:
Oh, I guess the thing in the deps should just be ppx_deriving...

tari3x:
All right, dune-release worked: created a new PR: https://github.com/ocaml/opam-repository/pull/16885

mseri:
I am going to close this since there is the new one.

PR #16842, opened 2020-07-20
Patch cohttp-2.5.2, revert semantic change in Header.replace

mseri:
See https://github.com/mirage/ocaml-cohttp/issues/701

Note that 2.5.3 is not being patched since a new minor release will be submitted once https://github.com/mirage/ocaml-cohttp/pull/702 is merged.


hannesm:
it may be worth to mark 2.5.2 (and 2.5.3) as `available: false` to avoid users installing these versions which semantics were changed

mseri:
Sure, I'll send a separate pr for those

PR #16846, opened 2020-07-20
Use pixel_pusher release 1.1, which is compatible with both jane 13 and 14

mbacarella:
The opam file for pixel_pusher 1.0 was modified to hard-code jane <14.


kit-ty-kate:
Thanks!

PR #16847, opened 2020-07-21
new file: packages/linwrap/linwrap.7.0.1/opam

UnixJunkie:
Can train a linear SVR now.
This regressor needs the optimization of only two parameters:
C and e "epsilon".
Previously, linwrap could only train a linear SVC, optionally with bagging.


kit-ty-kate:
Thanks!

PR #16852, opened 2020-07-21
ocaml-freestanding 0.6.1

mato:
* Add support for Solo5/xen bindings, `posix_memalign()`. (#81, @mato)
* nolibc: Implement `assert()` which was previously a no-op. (#80, @mato)

Downstream OPAM packages with C code that wish to use the interfaces/headers added in this release should depend on `ocaml-freestanding { >= 0.6.1 }`.

Part of mirage/mirage#1159.

### DIFF
--- a/packages/ocaml-freestanding/ocaml-freestanding.0.6.1/opam
+++ b/packages/ocaml-freestanding/ocaml-freestanding.0.6.1/opam
@@ -8,7 +8,6 @@ tags: "org:mirage"
 dev-repo: "git+https://github.com/mirage/ocaml-freestanding.git"
 build: [make]
 install: [make "install" "PREFIX=%{prefix}%"]
-remove: [make "uninstall" "PREFIX=%{prefix}%"]
 depends: [
   "conf-pkg-config"
   "ocamlfind" {build}
--- a/packages/ocaml-freestanding/ocaml-freestanding.0.6.1/opam
+++ b/packages/ocaml-freestanding/ocaml-freestanding.0.6.1/opam
@@ -1,0 +1,40 @@
+opam-version: "2.0"
+maintainer: "Martin Lucina <martin@lucina.net>"
+authors: "Martin Lucina <martin@lucina.net>"
+homepage: "https://github.com/mirage/ocaml-freestanding"
+bug-reports: "https://github.com/mirage/ocaml-freestanding/issues/"
+license: "MIT"
+tags: "org:mirage"
+dev-repo: "git+https://github.com/mirage/ocaml-freestanding.git"
+build: [make]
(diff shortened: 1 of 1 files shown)


mato:
@hannesm Any idea what's up with this FreeBSD Travis failure?

https://travis-ci.org/github/ocaml/opam-repository/jobs/710461114#L1847

It's inside the OCaml runtime; while it's possible that there's something in ocaml-freestanding's nolibc that's triggering this, it looks like a red herring since if this was an actual relevant failure we'd have picked it up ages ago...

hannesm:
@mato the failure is not relevant. though, on my development machine, I do not have `-Werror` in the CFLAGS of solo5-bindings-hvt and ocaml-freestanding -- so I get the 3 issues in io.c as warnings (and not as errors).

Reading the CI output, it looks to me that the configure script of OCaml embeds the `-Wall -Werror`:
`configure: Configuring OCaml version 4.09.1+dev1-2020-03-13`

and in `configure`:
```
  *+dev*) :
    gcc_warnings="-Wall -Werror" ;; #(
  *) :
    gcc_warnings="-Wall"
```

-> for some reason the 4.09.1 ocaml-src refers to a tarball which version is not 4.09.1 but 4.09.1+dev1-2020-03-13 (according to configure PACKAGE_VERSION -- @Octachron is this intentional? could the OCaml release process ensure that the file VERSION and PACKAGE_VERSION in configure are the same?)

Octachron:
@hannesm , the VERSION and PACKAGE_VERSION should be the same. Unfortunately, I messed up and indeed configure's PACKAGE_VERSION is not the right one for 4.09.1 .

kit-ty-kate:
I just removed the `remove` field which is unnecessary since opam 2.0. Could you return the fix in the opam file upstream?

mato:
@kit-ty-kate Done.

mato:
Actually, don't merge this. I just found a bug in `posix_memalign()` that was introduced here, will push an 0.6.2 once I've fixed it.

hannesm:
@Octachron thanks for your comment. Would it be worth to streamline VERSION & configure in the ocaml/ocaml repository so that this will not happen (I'd think configure could read the string from VERSION?) if you think this is a good idea, I can try to come up with a suitable PR.

for the 4.09.1, I'll PR later a patch to opam-repository (ocaml-src package) which fixes the version in configure (so that the build succeeds, and -Werror is not used).

mato:
FWIW #16859 is the ocaml-freestanding 0.6.2 PR which supersedes this.

hannesm:
please disregard my previous comment (@Octachron), I figured autoconf.ac reads the VERSION file; for 4.09.1 I opened #16862

PR #16856, opened 2020-07-22
Make dune >= 1.7 incompatible with OCaml 4.12

kit-ty-kate:
Should be fixed in the next major release most probably https://github.com/ocaml/dune/pull/3585

I suggest to merge this PR at the same time as the next release with a fix so that users won't have to rebuild their whole switch for nothing. cc @rgrinberg are there any release planned soon?

### DIFF
--- a/packages/dune/dune.1.11.0/opam
+++ b/packages/dune/dune.1.11.0/opam
@@ -26,7 +26,7 @@ homepage: "https://github.com/ocaml/dune"
 doc: "https://dune.readthedocs.io/"
 bug-reports: "https://github.com/ocaml/dune/issues"
 depends: [
-  "ocaml" {>= "4.02"}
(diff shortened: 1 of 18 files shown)


rgrinberg:
> are there any release planned soon?

We could make a point release with that fix if you'd like

kit-ty-kate:
> We could make a point release with that fix if you'd like

If you can that'd be amazing. I'm sure @ejgallego and other dev testers would be very happy.

avsm:
I don't think adding an upper bound will cause a package recompile -- that's only if the archive or build/install instructions change. It would be good to get this in to make opam-repository more consistent as soon as possible @kit-ty-kate, and merge the dune point release subsequently

kit-ty-kate:
> I don't think adding an upper bound will cause a package recompile

Sadly it does. I tested it with opam 2.0.7 before saying that

PR #16857, opened 2020-07-22
new file:   packages/linwrap/linwrap.7.1.3/opam

UnixJunkie:


UnixJunkie:
the package has a known incompatibility with mac

UnixJunkie:
I think this can be merged.
It would be nice if I can declare the conf-rdkit dependency as optional/nice-to-have.

UnixJunkie:
I'll send another version, which works on Mac.
I am still interested by this though:
" It would be nice if I can declare the conf-rdkit dependency as optional/nice-to-have.".
While not having this dependency will make some things not work at run-time, some things are still running and potentially useful to users.

PR #16859, opened 2020-07-22
ocaml-freestanding 0.6.2

mato:


mato:
Not sure what's up with the CI here. @kit-ty-kate?

mseri:
The CI should be back online. I merged the `ocaml-src` patch for `4.09.1` so that should also work fine now

kit-ty-kate:
Thanks a lot!

Issue #15643, opened 2020-01-09
conf-ida: make the search script a bit less aggressive


Armael:
cc @ivg 

I was trying to understand why my build was stuck trying to install `conf-ida`. Turns out it was frantically exploring the 8TB storage server mounted in `/tmp/nas` (a bit of a scary thing to discover).

Could the discovery script be made a bit less aggressive, and for instance avoid crossing filesystem boundaries?

ivg:
So here is the story, before opam 2.0 we were relying on locate on Linux and on mdfind on Mac. Both are not performing any searches, but make queries to the file database, which is fast. However, opam 2.0 is sandboxed so we can't access the locatedb anymore in opam 2.0 and we had to fall back to the `find` utility. We can tweak the search a bit and do not follow the tmp folder and other folders that are well-known for not storing user data, e.g., `/proc` or `/sys`. We could also add a timeout and stop searching after some number of seconds.

However, the general idea is that search is the last resort and we provide a few options for a user to set the IDA Pro folder. The very first option is to tell opam explicitly where it is located using `opam config`, e.g.,
```
opam config set ida-path /path/to/ida/folder
```
If this global variable is set no searches will be executed.

The second option is to specify the `IDA_PATH` environment variable. The third option is to add a folder with idaq executable to  `PATH`, so that `which idaq` returns the path. Only after that, the fourth attempt would be to extensively search the file system. 

cc @gitoleg

Armael:
Thank you for the details. For the record, I've hit this issue not because I wanted to install this package in particular, but because we've been hacking with @gasche on some tool that tries to build "all opam packages".

In that setting, we would simply like for the search to behave "nicely enough" and timeout without blocking the build (since most likely, the CI machine won't have IDA).

> We can tweak the search a bit and do not follow the tmp folder and other folders that are well-known for not storing user data, e.g., /proc or /sys. We could also add a timeout and stop searching after some number of seconds.

That would be very good :).

ivg:
Ok, so the simple and easy to implement solution would be wrapping the find into `timeout`, e.g, `timeout 10 find ....` and checking the exit status. 

Any objections? If not, then I will push the fix soonish (I'm currently working on Ghidra support (an open source alternative for IDA from NSA), so I will update both the newly comming `conf-ghidra` and the existing `conf-ida` scripts (both versions), to be more respectful to the user time).

Armael:
[...]

PR #16860, opened 2020-07-22
conf-ida 0.3

gitoleg:
The main outcome is that we're no longer using the `find`/`mdfind` utilities to find the IDA Pro installation, meaning that this PR makes our search less time-consuming (fixes https://github.com/ocaml/opam-repository/issues/15643)

## The Search

Both for Linux and macOS we start searching for IDA 64-bit executables in `PATH` using `which`. Then, if it fails, we continue searching in the OS-specific directories. We search until the first success, no deeper than 4 subdirectories, and examine no more than 1024 paths. 

## The macOS details

On macOS, IDA Pro 6.x and below is installed into two folders, `idaq.app` and `idaq64.app`, and each of them containing all the binaries we need (at `Contents/MacOS`). After a quick search on the Internet, we figured out that `IDA-7` has changed the naming scheme to `ida.app` and `ida64.app`.

We examine the following folders, `/Applications/`, `$HOME/Applications`, and `$HOME` for the `IDA` for the `*.app` folders in the following order
1. `ida64`
2. `ida`
3. `idaq64`
4. `idaq`

This gives priority to the more modern and 64-bit versions.

## The Linux details
On Linux, we try to find IDA 64 bit executables in the `$HOME` folder.

## Customization

It is possible to disable searching and set the concrete path to the IDA Pro installation using either the `IDA_PATH` environment variable or the `ida-path` opam variable.

See https://github.com/gitoleg/opam-repository/pull/2  for more information and discussion.

### DIFF
--- a/packages/conf-ida/conf-ida.0.2/files/find-ida.ml.in
+++ b/packages/conf-ida/conf-ida.0.2/files/find-ida.ml.in
@@ -1,41 +1,6 @@
+#load "unix.cma"
+
 open Printf
-open StdLabels
-
-module Array = struct
-  include Array
-
-  let find t ~f =
-    let length = length t in
-    let rec loop i =
-      if i >= length then None
-      else if f t.(i) then Some t.(i)
-      else loop (i + 1) in
-    loop 0
-
-  let find_map t ~f =
-    let length = length t in
-    let rec loop i =
-      if i >= length then None
-      else
-        match f t.(i) with
-        | None -> loop (i + 1)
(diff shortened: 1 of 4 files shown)


gitoleg:
@kit-ty-kate  Could you please merge this PR? The only reason we have fails is that IDA is absent on all the runners. I can add more distros in `x-ci-accept-failures` if you want.

kit-ty-kate:
Sorry for that, thanks!

PR #16862, opened 2020-07-23
add ocaml-src 4.09.1-1 with a patch for the configure script using the correct version; mark 4.09.1 as unavailable

hannesm:


mseri:
This breaks the CIs since now the 4.09.1 compiler is no longer available: 
```
[ERROR] No solution for ocaml-src.4.09.1: Your request can't be satisfied:

          - no matching version
```
I think, in this case, it is better to patch 4.09.1 itself to minimise breakage

hannesm:
@mseri I don't quite understand (and fail to see the CI log you pasted above).

The purpose of the ocaml-src package is to retrieve the source tarball of ocaml, and "cross-compile" the runtime system. This is used in ocaml-freestanding and mirage-xen-ocaml. It is not a real compiler, but sources only. This change (a) adds 4.09.1-1 which requires `"ocaml" {= "4.09.1"}` and removes/disables the 4.09.1 (which configure has the wrong version string, leading to `-Wall -Werror` being added to CFLAGS for the runtime, which breaks ocaml-freestanding on FreeBSD).

The compiler packages use the link to the source tarball directly (i.e. the reverse dependencies of ocaml-src are only those two mirage packages listed above).

mseri:
> @mseri I don't quite understand (and fail to see the CI log you pasted above).

It is the Travis job for 4.09: https://travis-ci.org/github/ocaml/opam-repository/jobs/711030428
It looks for 4.09.1 but that is no-longer available, however it seems it goes on to recompile 4.09.1-1, so it is not as bad as I thought.

The other failure is in the CI, even though it is green:
```
<><> Repository configuration for switch 4.10 <><><><><><><><><><><><><><><><><>
 1 default file:///home/opam/opam-repository
$ opam list
# Packages matching: installed
# Name              # Installed # Synopsis
base-bigarray       base
base-threads        base
base-unix           base
depext              transition  opam-depext transition package
ocaml               4.10.0      The OCaml compiler (virtual package)
ocaml-base-compiler 4.10.0      Official release 4.10.0
ocaml-config        1           OCaml Switch Configuration
opam-ci-scripts     0.1         pinned to version 0.1 at git+https://github.com/kit-ty-kate/opam-ci-scripts.git
opam-depext         1.1.3       Query and install external dependencies of OPAM packages
$ opam remove -y ocaml-src.4.09.1-1
Nothing to do.
[NOTE] ocaml-src.4.09.1-1 is not installed.

$ opam depext -uivyj 72 ocaml-src.4.09.1-1
[ERROR] No solution for ocaml-src.4.09.1-1: The following dependencies couldn't be met:
          - ocaml-src -> ocaml = 4.09.1
              base of this switch (use `--unlock-base' to force)
```
https://ci.ocaml.org/log/saved/docker-build-93b952e028b1ff5b82461f634b9580a6/916a423148ac82833f992c9c1e03c3cc12e71161

mseri:
Also, it looks like the base compiler is going to pick the broken sources anyways: it links tehem directly, see https://github.com/ocaml/opam-repository/blob/master/packages/ocaml-base-compiler/ocaml-base-compiler.4.09.1/opam

mseri:
Maybe it is just because this is not the real compiler, and it is sort of an unhortodox package. Thanks for the explanation

mseri:
The patch applied correctly in the tests that were able to run. I think this can be merged

PR #16863, opened 2020-07-23
disable cohttp: 2.5.2 and 2.5.3

mseri:
This is due to a breaking change in semantics in `Headers.replace` which is fixed in 2.5.2-1 and 2.5.4


mseri:
@hannesm here we go

mseri:
The uninstallability of all the dependent packages is expected and on purpose

PR #16864, opened 2020-07-23
xenctrl: add depext for opensuse

avsm:
will fix ci for https://github.com/mirage/ocaml-vchan/pull/135

### DIFF
--- a/packages/xen-evtchn-unix/xen-evtchn-unix.2.0.0/opam
+++ b/packages/xen-evtchn-unix/xen-evtchn-unix.2.0.0/opam
@@ -25,6 +25,7 @@ depexts: [
   ["libxen-dev"] {os-family = "debian"}
   ["xen-devel"] {os-distribution = "centos"}
   ["xen-devel"] {os-distribution = "fedora"}
+  ["xen-devel"] {os-distribution = "opensuse"}
   ["xenstore"] {os-distribution = "arch"}
 ]
 synopsis: "Xen event channel bindings."
--- a/packages/xen-evtchn-unix/xen-evtchn-unix.2.1.0/opam
+++ b/packages/xen-evtchn-unix/xen-evtchn-unix.2.1.0/opam
@@ -25,6 +25,7 @@ depexts: [
   ["libxen-dev"] {os-family = "debian"}
   ["xen-devel"] {os-distribution = "centos"}
   ["xen-devel"] {os-distribution = "fedora"}
+  ["xen-devel"] {os-distribution = "opensuse"}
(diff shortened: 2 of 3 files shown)


mseri:
Thanks

kit-ty-kate:
I believe this should be `os-family = "suse"` instead of `os-distribution = "opensuse"`. I don't see the package installed in the CI logs:
```
# Detecting depexts using vars: arch=x86_64, os=linux, os-distribution=opensuse-leap, os-family=suse
m4
# The following system packages are needed:
# All required OS packages found.
```

anuragsoni:
FWIW opensuse is a little annoying here. The value for ID_LIKE that populates the os_family is different for opensuse leap (which is what's run on the CI) and opensuse tumbleweed (their rolling release distribution).

Leap defines it as `suse opensuse` which would translate to `os-family=suse` and Opensuse tumbleweed defines it as `os-family=opensuse suse` which maps to `os-family=opensuse` :disappointed: 

I recently switched to tumbleweed and as a result depext doesn't work for since the package definitions talk about os-family=suse. I'm not sure what's the best way to deal with this, and I don't wan to add extra maintenance burden to support an OS that not a lot of OCaml users might be using. So far for my use i've been looking up the depext value for `suse` and installing the same packages manually.

mseri:
We could duplicate the Suse line for opensuse in the conf packages. If the dependencies names are the same it should not be a problem

avsm:
I agree, duplicating it seems just fine to me.  Should we be building a CI distro for tumbleweed as well then, @anuragsoni ?

anuragsoni:
> We could duplicate the Suse line for opensuse in the conf packages. If the dependencies names are the same it should not be a problem
I can't say for all the packages, but the ones i've tried so far have been the same for both opensuse-leap and opensuse-tumbleweed, so this scheme should be fine.

> I agree, duplicating it seems just fine to me. Should we be building a CI distro for tumbleweed as well then, @anuragsoni ?
I'd definitely appreciate a working depext for tumbleweed. its a fast moving distro (but i guess there is CI for arch so its not too different). They do publish official docker images which can be useful. I'd be very happy to work on contributing this to the opam CI if someone points me to the correct repository to look at.

mseri:
@anuragsoni do you mind sending a PR with the depext duplication?

anuragsoni:
> @anuragsoni do you mind sending a PR with the depext duplication?

Sure. I'll be able to make a PR later in the week.

mseri:
Tis can be improved separately in @anuragsoni pr if needed. Let’s merge in the meantime

PR #16876, opened 2020-07-25
package Coq 8.12.0

RalfJung:


packages/coqide/coqide.8.12.0/opam
```diff
@@ -0,0 +1,50 @@
...
+synopsis: "IDE of the Coq formal proof management system"
+description: """
+CoqIDE is a graphical user interface for interactive development
+of mathematical definitions, executable algorithms, and proofs of theorems
+using the Coq proof assistant.
+"""
+
+depends: [
+  "coq" {= version}
+  "lablgtk3-sourceview3"
```
Blaisorblade, line 17: Did the problem with beta7 get fixed?

packages/coq/coq.8.12.0/opam
```diff
@@ -0,0 +1,53 @@
...
+  [
+    "./configure"
+    "-configdir" "%{lib}%/coq/config"
+    "-prefix" prefix
+    "-mandir" man
+    "-docdir" doc
+    "-libdir" "%{lib}%/coq"
+    "-datadir" "%{share}%/coq"
+    "-coqide" "no"
+    "-flambda-opts" "-O3 -unbox-closures"
```
Blaisorblade, line 36: Having this on by default seems a bit too dangerous — there's serious enough bugs that 8.12.0 doesn't fix — some with bad ocaml versions, others in combo with native-compiler.
The best way forward seems to make native-compiler depend on a "package flag", encoded as a dummy package.

For the last discussion see https://github.com/ocaml/opam-repository/pull/16779 and https://coq.zulipchat.com/#narrow/stream/237663-coq-community-devs.20.26.20users/topic/flambda-using.20Docker-Coq.20images, but I think that's the summary.

packages/coqide/coqide.8.12.0/opam
```diff
@@ -0,0 +1,50 @@
+opam-version: "2.0"
```
Blaisorblade, line 1: Let's remember to split CoqIDE out if there are problems in CI.

packages/coqide/coqide.8.12.0/opam
RalfJung: I don't know, but the latest version 3.3.1 works. I don't think it is worth testing each intermediate version for incompatibilities, nor is it worth keeping a list of incompatible versions forever.

packages/coqide/coqide.8.12.0/opam
RalfJung: I don't follow... we want a CoqIDE package, no?
I consider creating two PRs and tracking their mutual dependencies to be just unnecessary work.

packages/coq/coq.8.12.0/opam
RalfJung: Sure I can remove this. I took it from the core-dev packages.

RalfJung: If those flags cause problems, IMO they should be removed from the core-dev packages as well.

packages/coqide/coqide.8.12.0/opam
Blaisorblade: Your call; I was just mentioning @palmskog's suggestion to avoid holding up Coq packages (IIRC because coqide held up some 8.11.* version for a long while). No strong opinion myself.

Blaisorblade: But again, no action needed if there are no problems in CI 🙂

packages/coqide/coqide.8.12.0/opam
avsm: In that case, just stick a constraint in to indicate the supported envelope.
```suggestion
  "lablgtk3-sourceview3" {>="3.1.1"}
```

kit-ty-kate: I used ` {>= "3.1.0"}` so that it at least gets rid of the 3.0beta versions, which seems sensible. This can always be modified later if you prefer.

kit-ty-kate:
Thanks a lot!

packages/coqide/coqide.8.12.0/opam
RalfJung: I have no idea what the future-compat policy is for these packages.

packages/coqide/coqide.8.12.0/opam
```diff
@@ -0,0 +1,49 @@
...
+synopsis: "IDE of the Coq formal proof management system"
+description: """
+CoqIDE is a graphical user interface for interactive development
+of mathematical definitions, executable algorithms, and proofs of theorems
+using the Coq proof assistant.
+"""
+
+depends: [
+  "coq" {= version}
+  "lablgtk3-sourceview3" {>= "3.1.0"}
```
MSoegtropIMC, line 17: @RalfJung @ejgallego : Is there a good reason for the restriction to lablgtk 3.1.0? For a reason yet to be investigated this version does not build on windows, but 3.0.beta5 does and it seems to work fine.

Blaisorblade: That's from https://github.com/ocaml/opam-repository/pull/16876#discussion_r460383730; there was a specific reason we banned beta7, but beta8 was fine again, and I suspect it was fixed on the Coq side as well.

Blaisorblade: Sorry: that's to say "no, there's no strong reason for the restriction to be that strict".

Blaisorblade: In fact, I should have searched the original issue. Indeed, beta7 needn't be blacklisted since 
https://github.com/coq/coq/pull/11234, which was merged in 8.11.0; only 8.10.* need the exclusion, as I noted in https://github.com/garrigue/lablgtk/pull/100#issuecomment-560587640.

Blaisorblade: Summing up: no, beta7 had to be banned only for coqide 8.10.*. No comment on potential other problems.
Coq 8.11.0 should be compatible with beta7 since https://github.com/coq/coq/pull/11234.

=== Below the former comments which I dropped.

1. That's from https://github.com/ocaml/opam-repository/pull/16876#discussion_r460383730; there was a specific reason we banned beta7, but beta8 was fine again, and I suspect it was fixed on the Coq side as well.

2. Sorry: that's to say "no, there's no strong reason for the restriction to be that strict".

3. In fact, I should have searched the original issue. Indeed, beta7 needn't be blacklisted since 
https://github.com/coq/coq/pull/11234, which was merged in 8.11.0; only 8.10.* need the exclusion, as I noted in https://github.com/garrigue/lablgtk/pull/100#issuecomment-560587640.

RalfJung: Okay, I opened https://github.com/ocaml/opam-repository/pull/16881 to relax the version bound.

MSoegtropIMC:
Thanks, this should work! I didn't try it but hearsay is that beta6 still works on windows but beta7 doesn't (I resorted to beta5 which we have been using on Windows for a while). Before I release the 8.12.0 platform, I will find out what the issue is (a strange link and/or flexlink issue).

Zimmi48:
Is there a reason why native compute was not disabled on macOS for this version contrary to what was done for versions 8.11.1 and 8.11.2?

Blaisorblade:
> Is there a reason why native compute was not disabled on macOS for this version contrary to what was done for versions 8.11.1 and 8.11.2?

No good reason. And apologies, you'd think I'd notice, wouldn't you 🤦?

But we know how to avoid that: fix the core-dev packages. https://github.com/ocaml/opam-repository/pull/16876#discussion_r460384284
@RalfJung used those as templates, and I'd guess that's the *correct* workflow (I used it too) — if the new release needs different opam packaging, how else would you know?
OTOH, there's also an opam file in-tree, but that's for devs and uses dune so it's not directly applicable.

Blaisorblade:
Just submitted https://github.com/ocaml/opam-repository/pull/16908 with the (untested) fix — I took a diff with the 8.11.2 packages.

RalfJung:
Ah, I wasn't aware it is still needed -- indeed I assumed it was not since the core-dev packages do not have this.

PR #16881, opened 2020-07-26
coq: relax lablgtk3-sourceview3 dependency

RalfJung:
Cc https://github.com/ocaml/opam-repository/pull/16876#discussion_r460521690

### DIFF
--- a/packages/coqide/coqide.8.12.0/opam
+++ b/packages/coqide/coqide.8.12.0/opam
@@ -14,7 +14,7 @@ using the Coq proof assistant.
 
 depends: [
   "coq" {= version}
-  "lablgtk3-sourceview3" {>= "3.1.0"}
+  "lablgtk3-sourceview3"
   "conf-findutils" {build}
   "conf-gnome-icon-theme3"
 ]


packages/coqide/coqide.8.12.0/opam
```diff
@@ -14,7 +14,7 @@ using the Coq proof assistant.
 
 depends: [
   "coq" {= version}
-  "lablgtk3-sourceview3" {>= "3.1.0"}
+  "lablgtk3-sourceview3" {>= "3.0"}
```
Blaisorblade, line 17: I'd have thought this bound was trivially satisfied?

Blaisorblade: See e.g. https://github.com/ocaml/opam-repository/tree/master/packages/lablgtk3-sourceview3

RalfJung: Coming from Rust, an unconstrained dependency is a complete no-go... but sure.

Blaisorblade: Thanks. And sorry that we looped back here... (My initial question was still a valid question, but I could have spent more looking up the answer 😅).

kit-ty-kate:
Thanks

PR #16883, opened 2020-07-26
Package ppx_cstubs.0.4.3

fdopen:
### `ppx_cstubs.0.4.3`
Preprocessor for easier stub generation with ctypes
ppx_cstubs is a ppx-based preprocessor for stub generation with
ctypes. ppx_cstubs creates two files from a single ml file: a file
with c stub code and an OCaml file with all additional boilerplate
code.

---
* Homepage: https://fdopen.github.io/ppx_cstubs/
* Source repo: git+https://github.com/fdopen/ppx_cstubs.git
* Bug tracker: https://github.com/fdopen/ppx_cstubs/issues

---

### DIFF
--- a/packages/ppx_cstubs/ppx_cstubs.0.4.3/opam
+++ b/packages/ppx_cstubs/ppx_cstubs.0.4.3/opam
@@ -36,9 +36,9 @@ with c stub code and an OCaml file with all additional boilerplate
 code.
 """
 url {
-  src: "https://github.com/fdopen/ppx_cstubs/archive/0.4.2.tar.gz"
+  src: "https://github.com/fdopen/ppx_cstubs/archive/0.4.3.tar.gz"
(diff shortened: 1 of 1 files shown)


avsm:
Thanks for the package contribution! We do keep all versions of opam packages published, so please don't delete the old version -- you simply need to add the new version.  This way, users depending on the old version can stick to it if they want, but by default `opam upgrade` will upgrade them to the freshest one.

fdopen:
This is a hotfix of a version that I've published one hour earlier; nobody depends on the "deleted" version or ever used it. 
Marking the other version as unavailable would just add more noise to the repo.

avsm:
That's fine; please try to specify that in the future in the PR so we know. It's possible to make exceptions like this, but we have to examine every case individually.

PR #16894, opened 2020-07-27
Update systemverilog repo information

xguerin:


mseri:
Is it possible to upload the old archive as a github release artifact instead of changing the checksums?

xguerin:
Sadly no. I did not anticipate it to be a problem and closed the original repos. I will keep that in mind for next time.

mseri:
We have the artefacts available in the opam cache:

https://opam.ocaml.org/cache/md5/c0/c088ec4d1a8fddd0d4eb14d797beaaac <- systemverilog.0.0.1
https://opam.ocaml.org/cache/md5/a2/a24fe433fd21a1ff4764caf10b013b65 <- systemverilog.0.0.2

If you don't mind, can you download those and replace the artefacts so that we can keep the checksum? Otherwise I need to diff their contents before merging, and I will not be able to do it before a couple of days

xguerin:
Perfect, thanks. I'll take care of this right away.

mseri:
Thank you very much. I'll wait for the CI, and if all goes well I'll merge

mseri:
Looks good, thanks

PR #16900, opened 2020-07-28
[new release] ezjsonm-lwt and ezjsonm (1.2.0)

avsm:
Simple Lwt-based interface to the Jsonm JSON library

- Project page: <a href="https://github.com/mirage/ezjsonm">https://github.com/mirage/ezjsonm</a>
- Documentation: <a href="https://mirage.github.io/ezjsonm">https://mirage.github.io/ezjsonm</a>

##### CHANGES:

* Add `find_opt` to provide an exception-less version of `find` (mirage/ezjsonm#39 @avsm)
* Raise `Parse_error` instead of assert failure if the input
  to `from_string` is not a valid JSON array or object (mirage/ezjsonm#39 @avsm).
* Upgrade build rules to dune 2.0 (mirage/ezjsonm#38 @avsm)
* Depend on Sexplib0 instead of Sexplib since we only need
  the type definition. This reduces the dependency cone of
  Ezjsonm (and skips Base). (mirage/ezjsonm#38 @avsm)

### DIFF
--- a/packages/ezjsonm-lwt/ezjsonm-lwt.1.2.0/opam
+++ b/packages/ezjsonm-lwt/ezjsonm-lwt.1.2.0/opam
@@ -8,10 +8,10 @@ doc: "https://mirage.github.io/ezjsonm"
 bug-reports: "https://github.com/mirage/ezjsonm/issues"
 depends: [
   "ocaml"
-  "ezjsonm"
+  "ezjsonm" {=version}
   "dune" {>= "1.0"}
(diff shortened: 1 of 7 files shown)


mseri:
I think opium_kernel was getting sexplib as a transitive dependency. Can you check if it is the case and add an upper bound on them? Thanks

avsm:
thanks for the traige, upper bounded mirage-tc and opium_kernel

mseri:
Further opium failures can be dealt with separately. Thanks again

PR #16910, opened 2020-07-29
2 packages from OCamlPro/ocp-index at 1.2.1

AltGr:
This pull-request concerns:
-`ocp-browser.1.2.1`: Console browser for the documentation of installed OCaml libraries
-`ocp-index.1.2.1`: Lightweight completion and documentation browsing for OCaml libraries

---
* Homepage: http://www.typerex.org/ocp-index.html
* Source repo: git+https://github.com/OCamlPro/ocp-index.git
* Bug tracker: https://github.com/OCamlPro/ocp-index/issues

---


mseri:
Tested locally. It seems that `ocp-browser` is compiled using `-alert` which has only been introduced in ocaml `4.08`, I would expect the ci to fail as well if and when it will be able to run the job. Since the rest is green I am going to add a lower bound on `ocaml >= 4.08` to `ocp-browser`, please either upstream the change or reconsidert the usage of `-alert` in the next version.

kit-ty-kate:
Thanks a lot!

PR #16911, opened 2020-07-30
[new release] postgresql (4.6.1)

mmottl:

##### CHANGES:

* Fixed a bug in `request_cancel` that turned errors into success and
    success into an error.  Thanks to Dmitry Astapov for this patch!

  * Added support for const char strings in stubs due to stricter handling
    in newer OCaml runtimes.  This eliminates C-compiler warnings.

### DIFF
--- a/packages/postgresql/postgresql.4.6.1/opam
+++ b/packages/postgresql/postgresql.4.6.1/opam
@@ -1,0 +1,38 @@
+opam-version: "2.0"
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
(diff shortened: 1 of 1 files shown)


avsm:
It looks like the freebsd depext is incorrect now, but not a blocker to merge:

```
pkg: No packages available to install matching 'database/postgresql96-client' have been found in the repositories
361
```

You may want to announce this on <https://discuss.ocaml.org>, where we have a Community category and an `announce` tag for this purpose.

mmottl:
It's difficult for developers to keep track of system dependencies for all platforms.  I assume PostgreSQL 9.6 has become deprecated on FreeBSD.  It would be great to know what the next best alternative on the test server is, which would presumably work for most users, too.  I'll then update the `conf-*` package accordingly.

PR #16912, opened 2020-07-30
Added MacPorts solution for depext for a few conf packages

MSoegtropIMC:
This PR adds a MacPorts solution for 4 conf packages which only had a homebrew solution on macOS.

What I am unsure about are the rules if an existing package should be modified or a package with a new / incremented version should be created. I followed what seems to be common practice for each individual package but ended up to do it differently. I would appreciate some general advice on this and some special advice for the case of adding a depext MacPorts solution.

Note: the handling of variants in MacPorts (like "gtk3 +quartz") doesn't look solid. It seems to work but depext treats this partly as separate packages (e.g. when listing what to install) and partly as one package (e.g. when sorting packages).

### DIFF
--- a/packages/conf-gnome-icon-theme3/conf-gnome-icon-theme3.0/opam
+++ b/packages/conf-gnome-icon-theme3/conf-gnome-icon-theme3.0/opam
@@ -7,6 +7,7 @@ depexts: [
   ["gnome-icon-theme"] {os-family = "debian"}
   ["gnome-icon-theme"] {os-family = "fedora" | os-family = "rhel"}
   ["gnome-icon-theme"] {os = "macos" & os-distribution = "homebrew"}
+  ["adwaita-icon-theme"] {os = "macos" & os-distribution = "macports"}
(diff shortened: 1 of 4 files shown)


MSoegtropIMC:
Can someone tell me what to do about the errors?

> * **`conf-gnome-icon-theme3.0`** has errors:
>   * **error 23:** Missing field 'maintainer'

This was so before

> * **`conf-pkg-config.1.3`** has errors:
>   * **error 46:** Package is flagged "conf" but has source, install or remove instructions

The existing conf-pkg-config files should have the same issue

> * **`conf-gtk3.18`** has some warnings:
>   * **warning 54**: External dependencies should not contain spaces nor empty string: "gtk3 +quartz"

For macports there is no way around this

> * **`conf-gtksourceview3.0+2`** has some warnings:
>   * **warning 54**: External dependencies should not contain spaces nor empty string: "gtksourceview3 +quartz"


For macports there is no way around this

mseri:
Thanks

PR #16916, opened 2020-07-30
qcheck, qcheck-ounit, qcheck-core and qcheck-alcotest 0.14

c-cube:


c-cube:
revdeps fail because of some breaking changes in dependencies (depend on `ounit2` instead of `oUnit`)

avsm:
would you mind adding upper bounds on those packages that are breaking in the revdeps? `opam admin add-constraint` will help, and then just `git add` the affected packages and push them to this PR.

kit-ty-kate:
I'm fixing the formatting issues brought by `opam admin add-constraint`. @avsm while waiting for the first stable release opam 2.1 which should fix this issue, I'd rather personally not advise people to use this command as it adds too many unnecessary changes and is annoying when using `git blame` or similar commands when debugging.

kit-ty-kate:
Thanks!

PR #16931, opened 2020-07-30
Package lens.1.2.4

pdonadeo:


kit-ty-kate:
Thanks! I untagged `ppx_deriving` as build dependency as it is rather clear from the source code (https://github.com/pdonadeo/ocaml-lens/blob/master/src/dune#L11) it is not. Could you return the fix to the opam file upstream so it is also change for future releases?

pdonadeo:
Thank you @kit-ty-kate!

PR #16933, opened 2020-07-31
added packages/oplsr/oplsr.8.0.1

UnixJunkie:


UnixJunkie:
I think this can be merged.
You need to interact with R so that the package can be installed properly (it is just a wrapper on top of some R package).

kit-ty-kate:
As usual indeed. Thanks!

PR #16934, opened 2020-07-31
new file:   packages/pardi/pardi.3.2.0/opam

UnixJunkie:
added -s option


UnixJunkie:
Cannot install on freebsd (who cares?) because the cpu and lz4 opam package doesn't compile on freebsd (who would have known?). :D
I opened issues for both.

kit-ty-kate:
Some people use freebsd, why make fun of it?

Anyway, thanks a lot for opening those issues.

UnixJunkie:
@kit-ty-kate I was just joking. But, fixing those and testing those is annoying (you need knowledge of freebsd and access to one...).

PR #16945, opened 2020-08-01
Add torch 0.10.

LaurentMazare:
Torch 0.10 switches to the recently released pytorch/libtorch 1.6.


LaurentMazare:
The xenial failure on travis seems to be caused by the virtual machine running out of memory. There are some large generated files in this package that take lots of memory to compile, hence running with '-j 1' but this is not sufficient here (not sure how much memory these VMs have).

kit-ty-kate:
Thanks

PR #16957, opened 2020-08-04
Adds conf-radare2

gitoleg:
This PR adds `conf-radare2` package
See https://github.com/BinaryAnalysisPlatform/bap/pull/1195 for more information

Fixes https://github.com/fxfactorial/ocaml-radare2/issues/7

### DIFF
--- a/packages/conf-radare2/conf-radare2.0.1/opam
+++ b/packages/conf-radare2/conf-radare2.0.1/opam
@@ -1,0 +1,21 @@
+opam-version: "2.0"
+name: "conf-radare2"
+version: "0.3"
+maintainer: "Ivan Gotovchits <ivg@ieee.org>"
+authors: "BAP Team"
+homepage: "https://github.com/BinaryAnalysisPlatform/bap/"
+bug-reports: "https://github.com/BinaryAnalysisPlatform/bap/issues"
+dev-repo: "git://github.com/BinaryAnalysisPlatform/bap/"
+license: "MIT"
+build: [
+  ["ocaml" "unix.cma" "find-radare2.ml"]
+]
+depends: [
+   "ocaml"
+   "base-unix"
+   "conf-which" {build}
+]
+substs: [ "find-radare2.ml" ]
+flags: [ conf ]
+
+synopsis: "Checks that radare2 is installed"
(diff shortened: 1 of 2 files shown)


ivg:
The failures are due to missing radare2 executable on some distributions. Also looping in @XVilka for the situational awareness.

XVilka:
Yes, already subscribed. I will use it as a dependency for "radare2"
package as well once merged.

ivg:
@kit-ty-kate, a friendly ping :) Could you guys merge this, we have two issues and two PR in two different repositories blocked on it :)

packages/conf-radare2/conf-radare2.0.1/opam
```diff
@@ -0,0 +1,23 @@
...
+]
+depends: [
+   "ocaml"
+   "base-unix"
+   "conf-which" {build}
+]
+substs: [ "find-radare2.ml" ]
+flags: [ conf ]
+
+depexts: [ "radare2" ]
```
kit-ty-kate, line 21: is the name of the package `radare2` on **every single** distributions?
Could it be replaced by:
```
depexts: [
  ["radare2"] {os-family = "debian"}
  ["radare2"] {os-distribution = "fedora"}
  ["radare2"] {os-distribution = "arch"}
  ["radare2"] {os-distribution = "gentoo"}
  ["radare2"] {os-distribution = "homebrew" & os = "macos"}
]
```
(taken from the `radare2` package) instead?

packages/conf-radare2/conf-radare2.0.1/opam
```diff
@@ -0,0 +1,23 @@
...
+authors: "BAP Team"
+homepage: "https://github.com/BinaryAnalysisPlatform/bap/"
+bug-reports: "https://github.com/BinaryAnalysisPlatform/bap/issues"
+dev-repo: "git://github.com/BinaryAnalysisPlatform/bap/"
+license: "MIT"
+build: [
+  ["ocaml" "unix.cma" "find-radare2.ml"]
+]
+depends: [
+   "ocaml"
```
kit-ty-kate, line 14: if I remember correctly the `{| .. |}` type of string came in OCaml 4.02 so I think this should be constrained.

packages/conf-radare2/conf-radare2.0.1/opam
gitoleg: done!

packages/conf-radare2/conf-radare2.0.1/opam
gitoleg: done! :)

kit-ty-kate:
Thanks!

PR #16963, opened 2020-08-05
new file:   packages/molenc/molenc.11.4.0/opam

UnixJunkie:


kit-ty-kate:
I added a missing constraint on minicli as its interface changed entirely starting from minicli 5.0.0. Could you return this upstream?
Thanks a lot!

UnixJunkie:
ok, I'll do

PR #16965, opened 2020-08-05
[new release] ppxlib (0.15.0)

pitag-ha:
Standard library for ppx rewriters

- Project page: <a href="https://github.com/ocaml-ppx/ppxlib">https://github.com/ocaml-ppx/ppxlib</a>
- Documentation: <a href="https://ocaml-ppx.github.io/ppxlib/">https://ocaml-ppx.github.io/ppxlib/</a>

##### CHANGES:

- Remove `base` and `stdio` dependencies (ocaml-ppx/ppxlib#151, @ceastlund)

- Update README and opam description (ocaml-ppx/ppxlib#155, @jeremiedimino)

- Fix `Driver.partition_transformation` (ocaml-ppx/ppxlib#156, @NathanReb)

- Implement name mangling for `ppxlib_traverse` (ocaml-ppx/ppxlib#159, @ceastlund)

### DIFF
--- a/packages/ppxlib/ppxlib.0.15.0/opam
+++ b/packages/ppxlib/ppxlib.0.15.0/opam
@@ -1,0 +1,50 @@
+opam-version: "2.0"
+maintainer: "opensource@janestreet.com"
+authors: ["Jane Street Group, LLC <opensource@janestreet.com>"]
+homepage: "https://github.com/ocaml-ppx/ppxlib"
+bug-reports: "https://github.com/ocaml-ppx/ppxlib/issues"
+dev-repo: "git+https://github.com/ocaml-ppx/ppxlib.git"
+doc: "https://ocaml-ppx.github.io/ppxlib/"
+license: "MIT"
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+]
+run-test: [
+  ["dune" "runtest" "-p" name "-j" jobs] { ocaml:version >= "4.10" }
+]
+depends: [
+  "ocaml"                   {>= "4.04.1"}
+  "dune"                    {>= "1.11"}
+  "ocaml-compiler-libs"     {>= "v0.11.0"}
+  "ocaml-migrate-parsetree" {>= "1.5.0"}
+  "ppx_derivers"            {>= "1.0"}
+  "sexplib0"
+  "stdlib-shims"
+  "ocamlfind"               {with-test}
+  "cinaps"                  {with-test & >= "v0.12.1"}
+  "base"                    {with-test}
+  "stdio"                   {with-test}
+]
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Many revdeps fail with this version. I've constrained them in #16967

NathanReb:
What were the revdeps failures? I haven't had a chance to look at them.

NathanReb:
We weren't really expecting this release to cause breakage so I'm curious to know what happened, aside from the packages that transitively depended on base rather than explicitly declaring it I mean.

kit-ty-kate:
> aside from the packages that transitively depended on base rather than explicitly declaring it I mean.

precisely :sweat_smile: Most of the packages that broke were in that case.

However, the main culprit was `ppx_sexp_conv.v0.14.0` which broke the packages depending on it. Our CI server is down at the moment so I can't show the logs but I remember this one being about using `Ppxlib.Longident` and not finding one of the required function to make a JaneStreet-based Map (i think it was a map). I'm guessing it comes from this change: https://github.com/ocaml-ppx/ppxlib/pull/151/files#diff-2cb2cf1b2a500c6b66ea9b2e127bc4bcL35
This breakage is a bit different from the others so maybe you want to do something about it instead. Sorry I should not have acted so fast for this package in retrospect.

pitag-ha:
I've found two problems: 

- as both of you have mentioned, several packages transitively depended on base without declaring that in the dune file. Several of them only indirectly by using `Caml` and nothing else in `base` (e.g. [here](https://github.com/progman1/genprintlib/blob/master/src/ppx_genprint.cppo.ml#L2)). In those packages folks might actually not want to depend on `base`. So a `module Caml = Stdlib` in there would do (as done in `ppxlib` now).

- more importantly (I think): some packages break, because of the change of record [Parsetree.module_binding](https://github.com/NathanReb/ppxlib/blob/f655de4e273c5447b918ed8d1bc749d638aa20f6/ast/ast.ml#L1004) due to [bumping the internal AST to 4.10](https://github.com/ocaml-ppx/ppxlib/pull/130/files#diff-b59beea0826752ba2e5b051c131b7716R859). That change for example breaks the build of [ppx_variants_conv](https://github.com/janestreet/ppx_variants_conv/blob/master/src/ppx_variants_conv.ml#L232), which some of the libraries in #16967 seem to depend on.

NathanReb:
> This breakage is a bit different from the others so maybe you want to do something about it instead. Sorry I should not have acted so fast for this package in retrospect.

No no worries, just wanted to make sure we didn't unexpectedly break something but all seems in order and entirely related to the removal of the base dependency so it's :ok_hand: 

Thanks for adding the upper bounds!

kit-ty-kate:
Ok so is it good to merge then?

kit-ty-kate:
Thanks!

PR #16974, opened 2020-08-06
[new release] merlin-extend (0.6)

let-def:
A protocol to provide custom frontend to Merlin

- Project page: <a href="https://github.com/let-def/merlin-extend">https://github.com/let-def/merlin-extend</a>
- Documentation: <a href="https://let-def.github.io/merlin-extend">https://let-def.github.io/merlin-extend</a>

##### CHANGES:

Remove META.transition (defining `merlin_extend`) because reasons (no pun
intended, fix let-def/merlin-extend#11).


kit-ty-kate:
The OCaml 4.11 support is missing from the Changelog but it's fine ^^ Thanks a lot!

PR #16979, opened 2020-08-07
Package ff.0.2.0

dannywillems:
### `ff.0.2.0`
OCaml implementation of Finite Field operations
OCaml implementation of Finite Field operations

---
* Homepage: https://gitlab.com/dannywillems/ocaml-ff
* Source repo: git+https://gitlab.com/dannywillems/ocaml-ff.git
* Bug tracker: https://gitlab.com/dannywillems/ocaml-ff/issues

---

### DIFF
--- a/packages/ff/ff.0.2.0/opam
+++ b/packages/ff/ff.0.2.0/opam
@@ -1,0 +1,18 @@
+opam-version: "2.0"
+authors: [ "Danny Willems <be.danny.willems@gmail.com>" ]
+maintainer: [ "Danny Willems <be.danny.willems@gmail.com>" ]
+synopsis: "OCaml implementation of Finite Field operations"
+description: "OCaml implementation of Finite Field operations"
+homepage: "https://gitlab.com/dannywillems/ocaml-ff"
+dev-repo: "git+https://gitlab.com/dannywillems/ocaml-ff.git"
+bug-reports: "https://gitlab.com/dannywillems/ocaml-ff/issues"
+license: "MIT"
+depends: [
+  "dune" {build & >= "2.0"}
+  # Has not been tested with other Zarith versions.
+  # The CI should verify it and it can then be extent
+  "zarith" {build & = "1.9.1"}
+  "alcotest" {with-test}
+]
+build: [[ "dune" "build" "-j" jobs "-p" name "@install" ]]
+run-test: [[ "dune" "runtest" "-p" name "-j" jobs ]]


kit-ty-kate:
opam-publish does not support automatic url fetching from gitlab. Instead you have to pass the url to the archive as follow:
```
$ opam publish -v 0.2.0 https://gitlab.com/dannywillems/ocaml-ff/-/archive/0.2.0/ocaml-ff-0.2.0.tar.gz
```

packages/ff/ff.0.2.0/opam
```diff
@@ -0,0 +1,18 @@
...
+authors: [ "Danny Willems <be.danny.willems@gmail.com>" ]
+maintainer: [ "Danny Willems <be.danny.willems@gmail.com>" ]
+synopsis: "OCaml implementation of Finite Field operations"
+description: "OCaml implementation of Finite Field operations"
+homepage: "https://gitlab.com/dannywillems/ocaml-ff"
+dev-repo: "git+https://gitlab.com/dannywillems/ocaml-ff.git"
+bug-reports: "https://gitlab.com/dannywillems/ocaml-ff/issues"
+license: "MIT"
+depends: [
+  "dune" {build & >= "2.0"}
```
kit-ty-kate, line 11: Dune does not support the `{build}` tag. See https://github.com/ocaml/dune/issues/2147

packages/ff/ff.0.2.0/opam
```diff
@@ -0,0 +1,18 @@
...
+  # Has not been tested with other Zarith versions.
+  # The CI should verify it and it can then be extent
+  "zarith" {build & = "1.9.1"}
```
kit-ty-kate, line 14: This dependency is not a build dependency. In your own `src/dune` file:
```
(library
 (public_name ff)
 (libraries zarith)
 (name ff))
```
it is used at runtime.

kit-ty-kate: Also the dependency is too strict. This should be at least `>= "1.9.1"` if you really want to be prudent, and maybe add `& < "2.0"` if you want to be extra careful if you think the next major version might break your code but otherwise the upper-bound is not necessarily needed.
A lower-bound constraint is nice to have though, otherwise if the dependency is too strict people won't be able to install your package alongside other packages that might require at least a different version.

packages/ff/ff.0.2.0/opam
dannywillems: TBH, I am not sure I understand the consequence of this. Do you mean I should just remove the `build` tag?

kit-ty-kate: yes

packages/ff/ff.0.2.0/opam
dannywillems: I agree with the upper-bound version. I prefer to have a lower-bound (and leave at 1.9.1 for the moment). I'll update the dep if someone does require 1.9.0 or 1.x. I prefer to be strict first and increase the version interval when required.

dannywillems:
I close this MR for https://github.com/ocaml/opam-repository/pull/16997/

PR #16981, opened 2020-08-08
Package radare2.0.0.7

XVilka:
- Rewritten IPC with `radare2` process to reduce amount of used descriptors and quit gracefully by @ivg 
- Removed depext and depend on `conf-radare2` package instead.


kit-ty-kate:
Thanks!

PR #16983, opened 2020-08-08
stdlib-shims: fails to build on 4.12.0+trunk

avsm:
upstream fix proposed at https://github.com/ocaml/stdlib-shims/pull/17

### DIFF
--- a/packages/stdlib-shims/stdlib-shims.0.1.0/opam
+++ b/packages/stdlib-shims/stdlib-shims.0.1.0/opam
@@ -10,7 +10,7 @@ license: ["typeof OCaml system"]
 depends: [
   "ocaml" {>="4.02.3"}
   "dune"
- ("dune" {> "2.7.0"} | "ocaml" {<"4.12.0~~"})
+ ("dune" {>= "2.7.0"} | "ocaml" {<"4.12.0~~"})
 ]
 build: [ "dune" "build" "-p" name "-j" jobs ]
 synopsis: "Backport some of the new stdlib features to older compiler"
--- a/packages/stdlib-shims/stdlib-shims.0.1.0/opam
+++ b/packages/stdlib-shims/stdlib-shims.0.1.0/opam
@@ -8,8 +8,9 @@ bug-reports: "https://github.com/ocaml/stdlib-shims/issues"
 tags: ["stdlib" "compatibility" "org:ocaml"]
 license: ["typeof OCaml system"]
 depends: [
+  "ocaml" {>="4.02.3"}
   "dune"
-  "ocaml" {>= "4.02.3" & < "4.12.0"}
+ ("dune" {> "2.7.0"} | "ocaml" {<"4.12.0~~"})
 ]
 build: [ "dune" "build" "-p" name "-j" jobs ]
 synopsis: "Backport some of the new stdlib features to older compiler"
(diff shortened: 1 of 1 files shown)


dra27:
This was fixed in Dune in https://github.com/ocaml/dune/pull/3576

avsm:
Sounds like we need to conflict this with dune < 2.7 and ocaml <4.12 then

avsm:
does https://github.com/ocaml/opam-repository/pull/16983/commits/241ff77171fde471ef7951c27bd6ec27199a5a3a look right @dra27?

packages/stdlib-shims/stdlib-shims.0.1.0/opam
```diff
@@ -8,8 +8,9 @@ bug-reports: "https://github.com/ocaml/stdlib-shims/issues"
 tags: ["stdlib" "compatibility" "org:ocaml"]
 license: ["typeof OCaml system"]
 depends: [
+  "ocaml" {>="4.02.3"}
   "dune"
-  "ocaml" {>= "4.02.3"}
+ ("dune" {> "2.7.0"} | "ocaml" {<"4.12.0~~"})
```
kit-ty-kate, line 13: why `>` instead of `>=`? I don't think dune 2.7.0 has been released yet.

avsm: it should be `>=` indeed! good catch

avsm, line 13:
```suggestion
 ("dune" {>= "2.7.0"} | "ocaml" {<"4.12.0~~"})
```

hannesm:
I'm curious whether dune < 2.7.0 should require ocaml < 4.12.0, and not bother with constraining stdlib-shims at all... since it looks to me that more packages (all with empty modules) will encounter this issue...

packages/stdlib-shims/stdlib-shims.0.1.0/opam
```diff
@@ -8,8 +8,9 @@ bug-reports: "https://github.com/ocaml/stdlib-shims/issues"
 tags: ["stdlib" "compatibility" "org:ocaml"]
 license: ["typeof OCaml system"]
 depends: [
+  "ocaml" {>="4.02.3"}
   "dune"
-  "ocaml" {>= "4.02.3"}
+ ("dune" {>= "2.7.0"} | "ocaml" {<"4.12.0~~"})
```
dra27, line 13: Sorry, I should not answer on my phone at the weekend. This isn't right!
```suggestion
 ("dune" {>= "2.7.0"} | "dune" & "ocaml" {<"4.12.0~~"})
```

dra27:
@hannesm is correct! Are you aware of any packages which would be affected by this change, though - it's not empty modules, but empty _libraries_ (i.e. a library with no modules)

avsm:
We'll find this out soon enough when we start doing bulk tests on 4.12. I suspect the current restriction is fine, and we can apply the big dune restriction if there turn out to be a non-trivial number of affected packages.

mseri:
Should we merge this or you prefer to wait? In the end it is a small change to keep the issue under control, it seems a reasonable start

avsm:
Agreed. I'll merge go ahead and merge this, and it can be tested with #17005

PR #16994, opened 2020-08-10
[new release] kafka_async, kafka_lwt and kafka (0.5)

didier-wenzek:
OCaml bindings for Kafka, Async bindings

- Project page: <a href="https://github.com/didier-wenzek/ocaml-kafka">https://github.com/didier-wenzek/ocaml-kafka</a>
- Documentation: <a href="https://didier-wenzek.github.io/ocaml-kafka">https://didier-wenzek.github.io/ocaml-kafka</a>

##### CHANGES:

* Add Async bindings
* Split sync and Lwt variants into separate OPAM packages
* Fix compatibility with OCaml 4.09
* Register callbacks as GC roots
* Port to `dune`
* Port to OPAM 2 format
* Make `dune` generate `opam` files automatically
* Add Changelog to facilitate releases with `dune-release`
* Depend ob `zlib` depext instead of explicit package names

### DIFF
--- a/packages/kafka/kafka.0.5/opam
+++ b/packages/kafka/kafka.0.5/opam
@@ -30,9 +30,9 @@ depexts: [
   ["librdkafka-devel"] {os-distribution = "fedora"}
   ["librdkafka"] {os-distribution = "homebrew" & os = "macos"}
   ["librdkafka-dev"] {os-distribution = "ubuntu"}
-  ["librdkafka"] {os = "freebsd"}
-  ["librdkafka"] {os = "netbsd"}
-  ["librdkafka"] {os = "dragonfly"}
+  ["librdkafka"] {os-distribution = "freebsd"}
+  ["librdkafka"] {os-distribution = "netbsd"}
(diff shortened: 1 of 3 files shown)


Leonidas-from-XIV:
The failure on FreeBSD is due to a required C library missing, possibly fixable by this PR: https://github.com/didier-wenzek/ocaml-kafka/pull/24

Leonidas-from-XIV:
I think the issue is that FreeBSD does not install ports to a location from where

```c
#include <librdkafka/rdkafka.h>
```

can read it. I guess this is a bigger can of worms (figuring out what the right `-L` and `-I` incantations are) that we should tackle upstream and we should just consider it uninstallable on FreeBSD for now.

didier-wenzek:
Agree.

The issue on Ubuntu is not easy to tackle either. The latter is due to Ubuntu Xenial coming with an old version of librdkafka (0.8.6).

mseri:
Thanks for the explanations

didier-wenzek:
Thank you.

PR #16998, opened 2020-08-11
add jsoo version contraints on menhir

hhugo:
fixing https://github.com/ocsigen/js_of_ocaml/issues/1050


hhugo:
note that we only add contraints to old js_of_ocaml versions. All `2.*` ones

mseri:
Thanks

PR #17014, opened 2020-08-17
[new release] user-agent-parser (0.1.3)

martinslota:
OCaml implementation of [user agent parse rules](https://github.com/ua-parser/uap-core)

### DIFF
--- a/packages/user-agent-parser/user-agent-parser.0.1.3/opam
+++ b/packages/user-agent-parser/user-agent-parser.0.1.3/opam
@@ -1,0 +1,32 @@
+opam-version: "2.0"
+maintainer: "Issuu Raccoons <racoon@issuu.com>"
+synopsis: "OCaml implementation of the user agent parse rules of uap-core"
+homepage: "https://github.com/issuu/uap-ocaml"
+doc: "https://issuu.github.io/uap-ocaml/"
+dev-repo: "git+https://github.com/issuu/uap-ocaml.git"
+bug-reports: "https://github.com/issuu/uap-ocaml/issues"
+build: [["dune" "build" "-p" name "-j" jobs]]
+run-test: [["dune" "runtest" "-p" name "-j" jobs]]
+depends: [
+  "alcotest" {with-test & >= "0.8.3"}
+  "base" {>= "v0.11"}
+  "dune"
+  "ppx_deriving" {build}
+  "ppx_let" {build}
+  "stdio" {>= "v0.11"}
+  "re" {>= "1.8.0"}
+  "yaml" {build & >= "0.2.1"}
+  "yojson" {>= "1.6.0"}
+]
+description: """
+OCaml implementation of the user agent parse rules from https://github.com/ua-parser/uap-core"""
+x-commit-hash: "aa0c038b0e968cccb62f90420a2296ee5a580824"
+authors: "Issuu Raccoon"
+url {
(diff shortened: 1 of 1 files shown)


packages/user-agent-parser/user-agent-parser.0.1.3/opam
```diff
@@ -0,0 +1,32 @@
...
+homepage: "https://github.com/issuu/uap-ocaml"
+doc: "https://issuu.github.io/uap-ocaml/"
+dev-repo: "git+https://github.com/issuu/uap-ocaml.git"
+bug-reports: "https://github.com/issuu/uap-ocaml/issues"
+build: [["dune" "build" "-p" name "-j" jobs]]
+run-test: [["dune" "runtest" "-p" name "-j" jobs]]
+depends: [
+  "alcotest" {with-test & >= "0.8.3"}
+  "base" {>= "v0.11"}
+  "dune"
```
avsm, line 13:
```suggestion
  "dune" {>="1.3"}
```

The dune-project indicates you need at least this version of dune, so adding bound.

packages/user-agent-parser/user-agent-parser.0.1.3/opam
```diff
@@ -0,0 +1,32 @@
...
+maintainer: "Issuu Raccoons <racoon@issuu.com>"
+synopsis: "OCaml implementation of the user agent parse rules of uap-core"
+homepage: "https://github.com/issuu/uap-ocaml"
+doc: "https://issuu.github.io/uap-ocaml/"
+dev-repo: "git+https://github.com/issuu/uap-ocaml.git"
+bug-reports: "https://github.com/issuu/uap-ocaml/issues"
+build: [["dune" "build" "-p" name "-j" jobs]]
+run-test: [["dune" "runtest" "-p" name "-j" jobs]]
+depends: [
+  "alcotest" {with-test & >= "0.8.3"}
```
avsm, line 11:
```suggestion
  "ocaml" {>="4.05.0"}
  "alcotest" {with-test & >= "0.8.3"}
```

This is missing an OCaml dependency. It looks like tests fail below 4.05, so that's a reasonable limit.

packages/user-agent-parser/user-agent-parser.0.1.3/opam
```diff
@@ -0,0 +1,33 @@
...
+bug-reports: "https://github.com/issuu/uap-ocaml/issues"
+build: [["dune" "build" "-p" name "-j" jobs]]
+run-test: [["dune" "runtest" "-p" name "-j" jobs]]
+depends: [
+  "ocaml" {>="4.05.0"}
+  "alcotest" {with-test & >= "0.8.3"}
+  "base" {>= "v0.11"}
+  "dune" {>="1.3"}
+  "ppx_deriving" {build}
+  "ppx_let" {build}
```
mseri, line 16: One last thing: `ppx_let` and `ppx_deriving` are used by the library. The first may have or may introduce a runtime, the second has a runtime, so it is customary to no longer tag them as build dependency.

I will make the change myself in this PR and merge, but please make this change (as well as the others) also on the upstream repository.

martinslota: The corresponding changes [have now been made upstream](https://github.com/issuu/uap-ocaml/pull/6).

PR #17022, opened 2020-08-19
update doc and description fields of ocamlsdl2.0.03

fccm:
Update doc field with new URL.
Update description field with versions information.


avsm:
Thanks!

PR #17023, opened 2020-08-19
suggested improvement of conf-libffi

UnixJunkie:
stolen from the ctypes-foreign opam package, which should depend on this new conf-libffi probably

### DIFF
--- a/packages/conf-libffi/conf-libffi.1/opam
+++ b/packages/conf-libffi/conf-libffi.1/opam
@@ -16,6 +16,5 @@ depexts: [
 ]
 synopsis: "Virtual package relying on a libffi system installation"
 description: "This package can only install if the libffi is installed on the system."
-depends: ["conf-pkg-config" {build}]
 bug-reports: "https://github.com/ocaml/opam-repository/issues"
 flags: conf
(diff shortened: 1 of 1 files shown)


UnixJunkie:
I also propose to remove the build line in the opam file because pkg-config is probably a linux-specific thing.

mseri:
I would prefer keeping pkg-config though and disable it (With the build string) only in unsupported systems

mseri:
Also please move the changes to conf-libffi.2

fccm:
Hi, maybe you could update the `mageia` line at the same time:
```
-  ["libffi5-devel"] {os-distribution = "mageia"}
+  ["libffi-devel"] {os-distribution = "mageia"}
```
(I'm using Mageia 7)

UnixJunkie:
Ok both, I will send another, updated PR.
@mseri: what is the way to do what you want, i.e. what is the constraint that should be added to the build string
so that it is tried only on systems with pkg-config?

mseri:
If the issue is only opne/freebsd, I would do
```
build: [["pkg-config" "libffi"]] {os != "freebsd" & os != "openbsd"}
```

mseri:
Although, at least freebsd, has https://www.freshports.org/devel/libffi/ and https://www.freshports.org/devel/pkg-config/

UnixJunkie:
@mseri apparently, even openbsd has pkg-config (cf. conf-pkg-config)

UnixJunkie:
superseded by https://github.com/ocaml/opam-repository/pull/17037

PR #17029, opened 2020-08-19
Attempt to fix PostgreSQL BSD package dependencies

mmottl:


avsm:
And that passes: https://travis-ci.org/github/ocaml/opam-repository/jobs/719380080

Thanks!

PR #17033, opened 2020-08-20
[new release] elpi (1.11.3)

gares:
CHANGES:

- ppxfindcache: use `shasum` instead of `sha1sum`
- Parser: print file names in a VScode friendly way
- Fix opam package dependency on camlp5 and ppxlib
- Fix output of `-print*` options to the `elpi` command line utility
- New builtin `rex_split` (like OCaml's `Str.split`)


gares:
test suite fails on freebsd because there is no `/bin/bash`, is it a blocker?

gares:
Thanks

gares:
I looked at the logs of alpine and centos, but I can't really find out the reason of failure. Fedora, which is not so different than centos, works... any ideas?

PR #17044, opened 2020-08-22
duplicate suse deps for opensuse

### LINKED ISSUE #17004: Duplicate depexts for suse/opensuse
See https://github.com/ocaml/opam-repository/pull/16864
@anuragsoni I have created this issue as a reminder, no pressure. There is no hurry

anuragsoni:
diff created via

```
rg 'os-family = "suse"' -l | xargs sed -i.'bak' -n 'p; s/os-family = "suse"/os-family = "opensuse"/p'
git clean -f
```
* ripgrep to search for files that contain os details for suse
* -n in sed prevents lines from printing unless it is explicitly printed
* p in the sed pattern forces printing if there is a match
* `git clean -f` to remove the backup files created by sed

closes #17004

### DIFF
--- a/packages/alsa/alsa.0.2.3/opam
+++ b/packages/alsa/alsa.0.2.3/opam
@@ -16,6 +16,7 @@ depexts: [
   ["alsa-lib-devel"] {os-distribution = "centos"}
   ["alsa-lib-devel"] {os-distribution = "fedora"}
   ["alsa-lib-devel"] {os-family = "suse"}
+  ["alsa-lib-devel"] {os-family = "opensuse"}
   ["libasound2-dev"] {os-family = "debian"}
 ]
 bug-reports: "https://github.com/savonet/ocaml-alsa/issues"
--- a/packages/alsa/alsa.0.3.0/opam
+++ b/packages/alsa/alsa.0.3.0/opam
@@ -34,6 +34,7 @@ depexts: [
   ["alsa-lib-devel"] {os-distribution = "centos"}
   ["alsa-lib-devel"] {os-distribution = "fedora"}
   ["alsa-lib-devel"] {os-family = "suse"}
+  ["alsa-lib-devel"] {os-family = "opensuse"}
   ["libasound2-dev"] {os-family = "debian"}
 ]
 available: [os = "linux"]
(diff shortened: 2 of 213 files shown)


mseri:
I am trying to understand why it says “ removed installable packages (1): pa_ppx.0.01”. 
@kit-ty-kate who has access to camelus logs?

XVilka:
What about just making it a synonym or the same family in opam. I believe, there is a minimum difference between them, like in many distributions in "debian" family.

mseri:
@rjbou is that feasible or you think it would be better to duplicate?

XVilka:
Nevermind, it should be already the same: https://github.com/ocaml/opam/blob/6aefe95e60084e63d01b1c7c028b6b77de1f839f/src/state/opamSysInteract.ml#L117

rjbou:
@mseri Yes, in current implementation it is the same (code pointed by @XVilka). If in the future, there is a need to have  different depext treatment, it can be split.

anuragsoni:
@XVilka @rjbou Thanks for pointing out the alias handling. I hadn't tried opam 2.1.0's alpha version where this aliasing seems to be present, and for some reason when i was searching in the codebase i only looked under the 2.0.7 tag 😄 

I'll close this PR, as the alias should be good. From what i've seen so far the package names have been the same on opensuse tumbleweed and opensuse leap (I think leap is what eventually becomes the commercial suse distribution so the names must be similar there as well).

If i run into any issues about this i'll open issues and work on fixing them at the opam repository.

Thanks!

rjbou:
@anuragsoni opam 2.1.0~alpha integrated external dependencies handling opam 2.0.x, it still rely on `opam-depext` plugin, it also handles `suse` & `opensuse` the [same way](https://github.com/ocaml-opam/opam-depext/blob/2.0/depext.ml#L168).

rjbou:
If you have the opportunity, please give a try to 2.1.0 (beta just released) & test the opensuse/suse depext handling in opam. Open an [issue](https://github.com/ocaml/opam/issues) if there things to fix. Thanks!

anuragsoni:
Thanks @rjbou 
I'll switch to the 2.1.0 beta. I was looking forward to the newer lockfile and depext handling so this is a good time to try it :)

PR #17049, opened 2020-08-22
Release conjury 2.0.1 (An OMake library).

jhwoodyatt:
This corrects some compatibility errors in the not very widely used Conjury library for OMake.

### DIFF
--- a/packages/conjury/conjury.2.0.1/opam
+++ b/packages/conjury/conjury.2.0.1/opam
@@ -20,5 +20,5 @@ build: [
 install: [ "omake" "install" ]
 url {
     src: "https://bitbucket.org/jhw/conjury/get/r2.0.1.tar.gz"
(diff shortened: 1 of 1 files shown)


avsm:
It looks like the ounit2 deps are being built too early

```
- ocamlfind: Package `oUnit2' not found
1915- *** omake: 47/59 targets are up to date
1916- *** omake: failed (0.24 sec, 5/5 scans, 19/19 rules, 25/91 digests)
1917- *** omake: targets were not rebuilt because of errors:
1918-    <scanner stage/develop/lib/main.cmo.scan>
1919-       depends on: tests/ocaml/main.ml
```

jhwoodyatt:
More likely, the problem is I develop on a case-insensitive filesystem, and ocamlfind has interesting behavior for matching package names that it inherits from the filesystem. I think I can fix this.

kit-ty-kate:
Thanks

PR #17054, opened 2020-08-24
Update LablTk's hasgot for clang 12

garrigue:
clang 12 seems to requires function to have prototypes, which breaks the `hasgot.c` generated during configuration.
This PR should allow to compile labltk and ocamlbrowser for ocaml 4.10 on MacOS.

### DIFF
--- a/packages/conf-tcl/conf-tcl.1/files/check.sh
+++ b/packages/conf-tcl/conf-tcl.1/files/check.sh
@@ -2,4 +2,5 @@ pkg-config tcl || \
 cc -o /dev/null \
   -I/usr/include/tcl \
   -I/usr/local/include/tcl \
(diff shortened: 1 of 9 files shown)


garrigue:
The last commit is just to make the CI happy (I'm all for adding freebsd depexts!)

garrigue:
Note that the revdep failure on `conf-tcl` and `conf-tk` is unrelated with this PR, as they succeed and produce no output (they can only succeed or fail).
It is just that `labltk.8.06.6` is incompatible with recent versions of ocaml.
(changes in 4.10 about how to access interpreter state?)
Is it worth fixing, knowing that there is no reason to install it with ocaml 4.10?

kit-ty-kate:
Thanks!

PR #17056, opened 2020-08-24
Release LablTk 8.06.9 and OCamlBrowser 4.11.0

garrigue:
Required for compatibility for ocaml 4.11.


garrigue:
The fix in `configure` for freebsd relies on merging #17054 first.
It should compile on other systems without it.

garrigue:
As written above, #17054 should fix the freebsd failure.
All other tests succeeded, in particular ocamlbrowser-4.11 on ocaml-4.11.0.

kit-ty-kate:
Thanks a lot!

PR #17077, opened 2020-08-27
add ez_file.0.1.2

lefessan:


packages/ez_file/ez_file.0.1.2/opam
```diff
@@ -0,0 +1,44 @@
+# This file was generated by `drom` from `drom.toml`.
+# Do not modify or add to the `skip` field of `drom.toml`.
+opam-version: "2.0"
+name: "ez_file"
```
mseri, line 4: Can you please remove name and version? They are inferred from the opam file path in any case

mseri (commented):
I would like to understand why Camelus thinks it is a broken package

packages/ez_file/ez_file.0.1.2/opam
lefessan: What is the problem with having them in the `opam` file ? It's quite useful to have a file which is self-contained, instead of relying on the fact that it is in a directory NAME.VERSION.

lefessan:
Louis just told me that it is because one of its dependencies is missing in its GIT branch (but the dependency is present in opam-repository `master` branch). I just rebased on `master`, let's see if it works...

packages/ez_file/ez_file.0.1.2/opam
mseri: No particular reason, mostly cleaning redundant information. In the past we had a few cases in which versions between opam files, opam metadata and dune did not match and created warnings all over the place

mseri:
Looks good now

PR #17083, opened 2020-08-28
Travis, let’s have also one test for 4.10

mseri:


kit-ty-kate:
I'm not sure this is necessary. We hope to get rid of Travis at some point and all the tests in Travis right now are the ones we don't have in datakit-ci. `ci/datakit/1 Build` does exactly the same with but with debian-stable instead of ubuntu-16.04.

mseri:
We have some packages compatible only with 4.10 in the PR pipeline, and they are currently not tested.
Since the CI was not always running, I wanted to add this to have a test in the meantime.

I don't think that adding it does much harm, and we can always remove it later if we see that it is only an annoyance.

PR #17085, opened 2020-08-29
Package ppx_const.2.0

mcclure:
### `ppx_const.2.0`
Compile-time "if" statement for conditional inclusion of code
This is a ppx extension which adds `if#const` and `match#const` constructs to
OCaml. They behave like normal `if` and `const`, but conditions are evaluated
at compile time and AST sections not selected are excluded from the program
completely. In conjunction with ppx_getenv, this can be used for conditional
compilation of code.

---
* Homepage: https://github.com/mcclure/ppx_const
* Source repo: git+https://github.com/mcclure/ppx_const.git
* Bug tracker: https://github.com/mcclure/ppx_const/issues

---

### DIFF
--- a/packages/ppx_const/ppx_const.2.0/opam
+++ b/packages/ppx_const/ppx_const.2.0/opam
@@ -1,0 +1,45 @@
+opam-version: "2.0"
+synopsis: "Compile-time \"if\" statement for conditional inclusion of code"
+description: """
+This is a ppx extension which adds `if#const` and `match#const` constructs to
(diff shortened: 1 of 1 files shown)


mcclure:
Version 2.0 incorporates a pull request that adds support for newer OCaml versions, plus Dune

mcclure:
Actually, I just realized a typo in my summary (from 2005); "They behave like normal if and const" should be "They behave like normal if and match". Gonna nuke and recreate this request.

PR #17096, opened 2020-08-31
Add opam-bin 0.9.4 and its companion library

lefessan:
opam-bin is a framework to enable creation and sharing of binary packages,
for faster recompilation and reuse of packages.


mseri:
Thanks

PR #17097, opened 2020-09-01
+faraday.0.7.2

seliopou:


kit-ty-kate:
This should fix all the revdeps failures triggered by this release.
As for the rest could you return 9878a58 to the opam files upstream? Your dune-project indicates this package requires at least dune 1.11.

I'll merge this tomorrow if all the revdeps are fine.

PR #17112, opened 2020-09-03
added packages/omlr/omlr.1.0.2/opam

UnixJunkie:


UnixJunkie:
The failure on FreeBSD is due to conf-r not supporting FreeBSD.
I sent a trial fix here: https://github.com/ocaml/opam-repository/pull/17121

kit-ty-kate:
Thanks

PR #17116, opened 2020-09-03
Fix ocaml-variants.4.10.0+nnpcheck

dra27:
The opam package didn't actually disable no-naked-pointers!


avsm:
Um, this might explain the lack of positive hits, except for @nojb who didn't use opam for this!

PR #17117, opened 2020-09-03
[new release] sihl (0.1.0)

jerben:
The modular functional web framework

- Project page: <a href="https://github.com/oxidizing/sihl">https://github.com/oxidizing/sihl</a>
- Documentation: <a href="https://oxidizing.github.io/sihl/">https://oxidizing.github.io/sihl/</a>

### Fixed
- DB connection leaks caused deadlocks
- Provide all service dependencies using functors
- Move Opium & Cohttp specific stuff into the web server service implementation to allow for swappable implementation based on something like httpaf
- Inject log service to all other services by default

### Added
- Support letters 0.2.0 for SMTP emailing
- Switch to exception based service API
- HTTP Response API to respond with file `Sihl.Web.Res.file`

### Fixed
- Stop running integration tests during OPAM release

### Added
- Initial release of Sihl

### DIFF
--- a/packages/sihl/sihl.0.1.0/opam
+++ b/packages/sihl/sihl.0.1.0/opam
@@ -1,0 +1,70 @@
+opam-version: "2.0"
+synopsis: "The modular functional web framework"
+description: "Build web apps fast with long-term maintainability in mind."
+maintainer: ["josef@oxidizing.io"]
+authors: ["Josef Erben"]
+license: "MIT"
+homepage: "https://github.com/oxidizing/sihl"
+doc: "https://oxidizing.github.io/sihl/"
+bug-reports: "https://github.com/oxidizing/sihl/issues"
+depends: [
+  "dune" {>= "2.4"}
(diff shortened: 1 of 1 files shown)


jerben:
Travis fails on Mac OS:

```
The packages you requested declare the following system dependencies. Please
make sure they are installed before retrying:
    openssl
```

kit-ty-kate:
I've fixed a few tiny things in your package description:
* `{dev}` dependencies are dependencies required when the user wants to get the master branch of your package. This is quite different from a short cut for people to install their favourite development software. I would encourage you to instead add a custom Makefile rule to install them or something similar.
* Some version numbers were missing their `v` prefix, required for JaneStreet packages.
* `alcotest` was required twice. Once as a normal dependency and the other as a test dependency. Since the later is redundant I removed it.

Could your return these fixes in the opam file upstream? Otherwise this seems good, thanks a lot!

jerben:
Thanks a for the review, opam file is fixed upstream!

PR #17118, opened 2020-09-03
conf-secp256k1: macOS package is in domt4/crypto tap

anton-trunov:
`secp256k1` is not part of the Homebrew core anymore and instead is available as a *tap*. This is the tap's current address: https://github.com/DomT4/homebrew-crypto. `opam depext` fails to install `secp256k1` because of this.


kit-ty-kate:
Sounds good. Thanks a lot!

PR #17119, opened 2020-09-03
core: does not build on ppc64

avsm:
upstream at https://github.com/janestreet/core/issues/140
Reported originally at https://github.com/mirage/ocaml-cohttp/pull/709


kit-ty-kate:
Thanks a lot!

PR #17122, opened 2020-09-04
[new release] daypack-lib.0.0.4

darrenldl:
Daypack-lib is a schedule, time and time slots handling library

Homepage: https://github.com/daypack-dev/daypack-lib
Repo: git+https://gitlab.com/daypack-dev/daypack-lib.git
Issue tracker: https://github.com/daypack-dev/daypack-lib/issues

### DIFF
--- a/packages/daypack-lib/daypack-lib.0.0.4/opam
+++ b/packages/daypack-lib/daypack-lib.0.0.4/opam
@@ -1,6 +1,4 @@
 opam-version: "2.0"
-name: "daypack-lib"
-version: "0.0.4"
 synopsis: "A schedule, time and time slots handling library"
 description: """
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
`dune install` is unnecessary and not recommended when using opam. `dune build` already creates a `<pkg>.install` file that opam is going to use to install everything.

Could you return the fix in your opam file upstream so that it is taken into account for the future?

Otherwise this looks fine. Thanks.

darrenldl:
Aha, thanks for letting me know.

Is it desirable for me to patch the opam files of the previous published versions (in another PR)?

kit-ty-kate:
> Is it desirable for me to patch the opam files of the previous published versions (in another PR)?

That'd be really helpful yes :relaxed:

PR #17128, opened 2020-09-04
Correct OCaml lowerbound for old PCRE

dra27:
I was doing some archaeology and discovered a missing lowerbound on PCRE.


kit-ty-kate:
Thanks!

PR #17130, opened 2020-09-04
ppx string.v0.14.1

xclerc:
Should properly fix https://github.com/janestreet/ppx_string/issues/1 and https://github.com/janestreet/ppx_string/issues/2.


mseri:
Looks good. Thanks

PR #17134, opened 2020-09-04
Tezos node.7.4

pirbo:
This is the first part of the Tezos v7.4 release (the client part)

This release is only about adding the "economic protocol" under vote on the mainnet chain and giving access to its corresponding test-network. Nothing else changed.


kit-ty-kate:
Thanks!

PR #17136, opened 2020-09-05
Why3 1.3.2

silene:


silene:
The CI error is on FreeBSD, where the `lablgtk` dependency of `why3-ide` failed to install. (Package `why3`, however, was successful.)

kit-ty-kate:
Thanks!

PR #17147, opened 2020-09-07
Compile dune 2.7.1 without interbranch propagation

garrigue:
This applies a patch to dune 2.7.1 so that it can be compiled with the `garrigue:remove_interbranch_propagation_410` branch.
Note that it only adds (correct) type annotations, so it cannot break anything :-)

(This is for testing purposes)

### DIFF
--- a/packages/dune/dune.2.7.1/opam
+++ b/packages/dune/dune.2.7.1/opam
@@ -46,6 +46,9 @@ depends: [
   "base-unix"
   "base-threads"
 ]
+patches: [
+  ["dune_no_propagation.diff" "md5=24d098fcb1f0c141c2d32a55986e5311"]
+]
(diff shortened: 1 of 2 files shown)


emillon:
If that does the right thing for you, no problem to add this to dune master of course.

packages/dune/dune.2.7.1/opam
```diff
@@ -46,6 +46,9 @@ depends: [
   "base-unix"
   "base-threads"
 ]
+patches: [
+  ["dune_no_propagation.diff" "md5=24d098fcb1f0c141c2d32a55986e5311"]
```
mseri, line 50:
```suggestion
  "dune_no_propagation.diff"
```

Afaik extra-files require the hash but not the patches

mseri: This should make camelus happy

PR #17156, opened 2020-09-08
[conf-gmp] fix test to avoid warning about undeclared function

maroneze:
A user of a package depending on `conf-gmp` reported having issues on macOS, possibly due to a non-standard `CFLAGS` in their environment (containing `-Werror`). We noticed the `test.c` file in the `conf-gmp` package always generates a warning due to usage of a function which is not declared (the function is actually not exported, so the user is never supposed to call it directly). I modified the code to use some basic GMP features, but without triggering the warning.

For information, before this PR, simply installing `conf-gmp` with `--verbose` results in:

```
- + cc -c -I/usr/local/include test.c
- test.c: In function ‘test’:
- test.c:7:2: warning: implicit declaration of function ‘__gmp_init’; did you mean ‘__gmpf_init’? [-Wimplicit-function-declaration]
-     7 |  __gmp_init();
-       |  ^~~~~~~~~~
```

So, in practice, the package did correctly detect whether gmp was installed (otherwise the `#include <gmp.h>` would fail), but generated a warning that could in some cases lead to errors even when GMP is installed.

(Explanation and suggestions by @zilbuz)

### DIFF
--- a/packages/conf-gmp/conf-gmp.2/files/test.c
+++ b/packages/conf-gmp/conf-gmp.2/files/test.c
@@ -1,0 +1,10 @@
+#include <gmp.h>
+#ifndef __GMP_H__
+#error "No GMP header"
+#endif
+
+void test(void) {
+  mpz_t n;
+  mpz_init(n);
+  mpz_clear(n);
+}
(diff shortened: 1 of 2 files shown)


mseri:
Thanks, could you move the change to a new `conf-gmp.2`? You can copy the current folder and make the changes there. You also need to change the hash of the `test.c` file in the `opam` file: https://github.com/ocaml/opam-repository/blob/bb56bd0f2ae570e00f23afa483950e68e8b1e3f6/packages/conf-gmp/conf-gmp.1/opam

maroneze:
Thanks for the remarks, and indeed I had forgot to update the checksum.

mseri:
Thanks!

PR #17155, opened 2020-09-08
Package fpath.0.7.3

dbuenzli:
### `fpath.0.7.3`
File system paths for OCaml
Fpath is an OCaml module for handling file system paths with POSIX or
Windows conventions. Fpath processes paths without accessing the file
system and is independent from any system library.

Fpath depends on [Astring][astring] and is distributed under the ISC
license.

[astring]: http://erratique.ch/software/astring

---
* Homepage: https://erratique.ch/software/fpath
* Source repo: git+https://erratique.ch/repos/fpath.git
* Bug tracker: https://github.com/dbuenzli/fpath/issues

---
v0.7.3 2020-09-08 Zagreb
------------------------

- Require OCaml 4.03 and drop `result` compatibility package
- Support OCaml 4.12 injectiviy annotation of Map.S (#18).
  Thanks to Kate for the patch. 

---
:camel: Pull-request generated by opam-publish v2.0.2

PR #17161, opened 2020-09-09
Temporarily fix esy: Avoid using the %{dev}% value as esy doesn't know about it

kit-ty-kate:
Hopefully this should fix your issue @phated (https://github.com/ocaml/opam-repository/pull/17155#issuecomment-689251506). Could you check if this work? (I don't know if it's possible to use a special commit of opam-repository with esy)

cc @dbuenzli I'm not sure how quickly the esy issue https://github.com/esy/esy/issues/1165 is going to be fixed and released but just in case you're planning to release some other packages soon it might be worth using `=true" {dev}` instead, idk.

### DIFF
--- a/packages/b0/b0.0.0.0/opam
+++ b/packages/b0/b0.0.0.0/opam
@@ -17,7 +17,7 @@ depends:
 ]
 build:
 [[
-  "ocaml" "pkg/pkg.ml" "build" "--dev-pkg" "%{dev}%"
+  "ocaml" "pkg/pkg.ml" "build" "--dev-pkg=true" {dev}
 ]]
 
 synopsis: """Software construction care"""
--- a/packages/b0/b0.0.0.1/opam
+++ b/packages/b0/b0.0.0.1/opam
@@ -17,7 +17,7 @@ depends:
 ]
 build:
 [[
-  "ocaml" "pkg/pkg.ml" "build" "--dev-pkg" "%{dev}%"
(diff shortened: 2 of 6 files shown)


phated:
Oh, I thought that `{dev}` was supposed to be providing a boolean value to `--dev-pkg`, but that assumption seems to be incorrect with this workaround? What exactly does `{dev}` do there? I want to make sure the upstream esy fix is correct 😄

kit-ty-kate:
> Oh, I thought that `{dev}` was supposed to be providing a boolean value to `--dev-pkg`, but that assumption seems to be incorrect with this workaround? What exactly does `{dev}` do there? I want to make sure the upstream esy fix is correct smile

Mh, I'm not sure to understand what you are saying. `%{dev}%` does provide a boolean value.
It can be used both as a filter or as a value directly. When used as a filter, the parameter given before is used only if the value is `true` and ignored otherwise. When used as a value directly, it is replaced by either `true` or `false`.

The upstream fix seems to be correct (if I understand what it does correctly): https://github.com/esy/esy/pull/1166

phated:
Perfect, thanks for the explanation and for looking at the proposed fix!

ManasJayanth:
Thank you so much for this!

I have raised a PR to fix it at esy https://github.com/esy/esy/pull/1166

Will report back once a bug fix release it made.

ManasJayanth:
Hi, Esy 0.6.7 has been released with the fix for missing %{dev} in the build steps. 
https://github.com/esy/esy/releases/tag/v0.6.7

PR #17172, opened 2020-09-10
conf-libev depends on ocaml

hcarty:
A dependency on `ocaml` is missing in `conf-libev`.  I found this when trying to install lwt with libev support in an empty switch.

The 3.11.0 constraint on OCaml is a somewhat conservative guess.  Much older compilers would probably work as well since `discover.ml` doesn't seem to use any remotely new language features or stdlib functions.


mseri:
Thanks. I’ll wait for the ocaml CI to run and if all is well merge

kit-ty-kate:
Thanks a lot!

hcarty:
Thank you @mseri and @kit-ty-kate!

PR #17174, opened 2020-09-11
[new release] opam-compiler (0.1.0)

emillon:
Plugin to create switches using custom compilers

- Project page: <a href="https://github.com/ocaml-opam/opam-compiler">https://github.com/ocaml-opam/opam-compiler</a>

##### CHANGES:

- Initial release
- Add "create" command

### DIFF
--- a/packages/opam-compiler/opam-compiler.0.1.0/opam
+++ b/packages/opam-compiler/opam-compiler.0.1.0/opam
@@ -5,7 +5,6 @@ description:
 maintainer: ["Etienne Millon <me@emillon.org>"]
 authors: ["Etienne Millon <me@emillon.org>"]
 license: "ISC"
-tags: ["flags:plugin"]
 homepage: "https://github.com/ocaml-opam/opam-compiler"
 bug-reports: "https://github.com/ocaml-opam/opam-compiler/issues"
 depends: [
@@ -33,6 +32,7 @@ build: [
   ]
 ]
 dev-repo: "git+https://github.com/ocaml-opam/opam-compiler.git"
+flags: plugin
 x-commit-hash: "ca6d450733363c38ca206afd9ac0cf2a783fdf8f"
 url {
   src:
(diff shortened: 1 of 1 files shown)


packages/opam-compiler/opam-compiler.0.1.0/opam
```diff
@@ -0,0 +1,44 @@
+opam-version: "2.0"
+synopsis: "Plugin to create switches using custom compilers"
+description:
+  "This plugin can manage switches using various sources for compilers, such as git branches, github PRs, etc"
+maintainer: ["Etienne Millon <me@emillon.org>"]
+authors: ["Etienne Millon <me@emillon.org>"]
+license: "ISC"
+tags: ["flags:plugin"]
```
avsm, line 8: this should be `flags: plugin` to be explicit I think.

avsm: (that is, a direct field, not just the tags entry)

emillon: hmm I stole that from opam-publish:

https://github.com/ocaml/opam-repository/blob/54b0223dcacb6311aa12f5342c34e1be684a79a3/packages/opam-publish/opam-publish.2.0.1/opam#L9

but publish uses just flags:

https://github.com/ocaml/opam-repository/blob/54b0223dcacb6311aa12f5342c34e1be684a79a3/packages/publish/publish.0.3.4/opam#L23

I think this was correctly picked, but I'll double check the opam docs and sources

emillon: indeed, opam just looks at `flags:`. I missed this because it's only considered when opam knows about this plugin, but it's not installed. So I didn't test this case. Anyway, that should be ok now.

kit-ty-kate:
Thanks a lot!

PR #17178, opened 2020-09-11
2 packages from Ninjapouet/ezcmdliner at 0.1.1

Ninjapouet:
This pull-request concerns:
-`ezcmdliner.0.1.1`: An easy interface to cmdliner
-`ezcmdliner-ppx.0.1.1`: Derivates cmdliner terms from type definitions

---
* Homepage: https://github.com/Ninjapouet/ezcmdliner
* Source repo: git+https://github.com/Ninjapouet/ezcmdliner.git
* Bug tracker: https://github.com/Ninjapouet/ezcmdliner/issues

---

### DIFF
--- a/packages/ezcmdliner-ppx/ezcmdliner-ppx.0.1.1/opam
+++ b/packages/ezcmdliner-ppx/ezcmdliner-ppx.0.1.1/opam
@@ -1,0 +1,37 @@
+opam-version: "2.0"
+synopsis: "Derivates cmdliner terms from type definitions"
+description:
+  "This PPX gives to your program a nice CLI using simple type annotations"
+maintainer: "Julien Blond <julien.blond@gmail.com>"
+authors: "Julien Blond <julien.blond@gmail.com>"
+license: "Apache 2.0"
+homepage: "https://github.com/Ninjapouet/ezcmdliner"
+bug-reports: "https://github.com/Ninjapouet/ezcmdliner/issues"
+depends: [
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
Thanks!

kit-ty-kate:
I hadn't realized this but there was already a package with a very similar name `ez_cmdliner` from @OCamlPro used internally since 2017 and released in opam-repository a month ago (https://github.com/ocamlpro/ez_cmdliner).
Would it be possible to discuss some kind of renaming to avoid confusion? Possibly with the author of the other library if needed?

Ninjapouet:
Erf. I missed that ez_cmdliner release... I won't interfere with Fabrice's work since it's older and probably used by more people. My lib was published for few people only six months ago. I'll try to find another name or find a way to merge the two libs if it makes sense and see with Fabrice then. By the way, how can i rename the package in opam publishing workflow ?

kit-ty-kate:
Thanks! This is a bit of an exceptional process but since the package has been added to opam-repository not too long ago a simple pull request on the git repository (this one) moving the file from `ezcmdliner/ezcmdliner.0.1.1` to `<new_name>/<new_name>.0.1.1` should be enough.

kit-ty-kate:
oh and some of the data will have to change too of course, at least the url to the archive and its checksums. To get the new checksums just download the new archive (with the changed library name etc.) and call `md5sum` and `sha512sum` on it.

PR #17180, opened 2020-09-11
Why3 1.3.3

silene:


silene:
For `why3-ide`, the failures on CentOS and Debian Testing are caused by missing external packages needed for building the `lablgtk` dependency. Nothing new here.

For `why3-coq`, the failures on Fedora and Alpine happen during the build of the `coq` dependency (abrupt termination due to timeout? out of memory?).

Also, 4 `opam install` jobs seem to be stuck at initialization time (server failure?). The other ~30 installation tests succeeded.

kit-ty-kate:
Thanks a lot!

Issue #17152, opened 2020-09-07
csexp with-test dependency cycle


leviroth:
```
$ opam install csexp --with-test
The actions to process have cyclic dependencies:
  - ∗ jst-config.v0.14.0 → ∗ time_now.v0.14.0 → ∗ ppx_inline_test.v0.14.1 → ∗ ppx_expect.v0.14.0 → ∗ csexp.1.3.1 → ∗ dune-configurator.2.7.1 → ∗ jst-config.v0.14.0
  - ∗ base.v0.14.0 → ∗ jst-config.v0.14.0 → ∗ time_now.v0.14.0 → ∗ ppx_inline_test.v0.14.1 → ∗ ppx_expect.v0.14.0 → ∗ csexp.1.3.1 → ∗ dune-configurator.2.7.1 → ∗ base.v0.14.0

No solution found, exiting
```

I believe this was introduced with dune-configurator.2.7.1's dependency on csexp.

mseri:
Thanks for the report. It is a bug in opam trying to traverse too many layers for the test dependencies. It is currently fixed in the beta of opam 2.1 if you need a fast fix and don’t want to pin an older version.

I don’t yet know if there are plans to backport the fix or ideas for other ways to solve it, but I know that what is the best thing to do is currently being discussed

PR #17186, opened 2020-09-12
Mitigate test dependency cycle (ocaml-opam/opam-depext#121) between base, dune-configurator and csexp

kit-ty-kate:
Fixes https://github.com/ocaml/opam-repository/issues/17152
Error also encountered by @emillon, @voodoos, @c-cube 

This is only a mitigation until opam 2.1 comes out and is wildly used in CIs.

@jeremiedimino @mefyl Could this be ported to csexp upstream temporarily so that it does not happen in the future until opam 2.1 comes out?


jeremiedimino:
Sure, @voodoos upstreamed the patch

PR #17189, opened 2020-09-13
Fix dune lower-bound for headache

Et7f3:
It is unlikely that someone will use sur lower dune but it wasn't valid.

CC: @pbaudin if you want to update your opam file https://github.com/Frama-C/headache/blob/master/headache.opam dune shouldn't be a build deps


pbaudin:
> It is unlikely that someone will use sur lower dune but it wasn't valid.
> 
> CC: @pbaudin if you want to update your opam file https://github.com/Frama-C/headache/blob/master/headache.opam dune shouldn't be a build deps

Thanks @Et7f3, the file is updated as suggested.

PR #17196, opened 2020-09-14
Add opam-bin.0.9.5

lefessan:


kit-ty-kate:
Completely removing version from the repository is not in our usual policy. Would it be possible to add `available: false` to the that version instead?

lefessan:
> Completely removing version from the repository is not in our usual policy. 

It is a very bad policy: solvers (even aspcud and other champions) don't perform well when the fan (the number of versions par package) is too big, resulting in queries to opam taking minutes to solve. You should really ask the team behind `Dose` about this. Removing packages from the repository used to be a problem when `opam` needed to redownload them to properly remove/uninstall them, but now, `opam` uses its own meta-information to remove packages, so removing a package from the repository hardly breaks anything, no ?

Also, the other version is buggy, is it really in your policy to keep buggy versions in the repository ?

Finally, it is a program, published for the first time a few days ago, it is unlikely it is going to break any other package or installation.

> Would it be possible to add available: false to the that version instead?

It is possible, but I still think it is a bad idea and the package should be removed.

avsm:
That's our policy. When you publish publicly to opam-repo, your users choose when to up/grade, not the author of the software.  We can debate changing it, but not in the middle of a PR please.

lefessan:
I pushed a version with `available: false`.

> That's our policy. When you publish publicly to opam-repo, your users choose when to up/grade, not the author of the software. 
I think making every user wait longer when calling `opam` is a bigger problem than forcing occasionally a few users to upgrade a buggy version.

> We can debate changing it, but not in the middle of a PR please.
Indeed, but it's a policy, not a law, so a PR is a good place to discuss whether the policy should be strictly applied to the PR or not. In this case, the likelihood of a user noticing the removal of `opam-bin.0.9.4` was minimal, so it was a good candidate for an exception. Anyway, the discussion is closed since the PR now keeps the buggy version.

avsm:
We consider the impact of multiple versions on the opam repository holistically (that is, across the full package set), and not on the basis of individual versions.  So we are fixing any such solver issues across the board, and removing one package will make no difference whatsoever.  You can track the development of opam 2.1 in the [dev wiki](https://github.com/ocaml/opam/wiki) to read more on those efforts.

> the likelihood of a user noticing the removal of opam-bin.0.9.4 was minimal

The point of this package repository is that the _user_ gets to make that decision, not you (the package publisher).  Even marking it as `available: false` without a clear reason (that is, a security issue or other serious flaw or build issue) goes against the conventions we apply to every other package in this repository.  Every new version submitted here fixes bugs or adds features or changes interfaces -- the opam client ensures that users get the freshest version available, and can also pin back if they choose to.

So... this PR can go in without question if it just added opam-bin.0.9.5.  But as it modifies the availability of another package, I have to ask: why `available: false` and not just let the user update as normal to the latest issue?

lefessan:
> So... this PR can go in without question if it just added opam-bin.0.9.5.

Ok, I removed the part on opam-bin.0.9.4, the PR is now just adding opam-bin.0.9.5 and its library.

> But as it modifies the availability of another package, I have to ask: why available: false and not just let the user update as normal to the latest issue?

opam-bin.0.9.4 had a bug that prevents creation of binary packages in the default config. Given that it's the only goal of opam-bin, it would be a waste of precious time for the user to try to use it.

kit-ty-kate:
Thanks

avsm:
Thanks @lefessan. If you judge that earlier versions of your package really shouldn't be users, feel free to open a PR to mark them as unavailable.  A note in the PR as to why this is the case would let us know (and a short comment in the opam file can also be helpful for future maintainers to see). If it's a security issue and you do not want the reason to be public, you can email any of us directly; we've had several of those in the past.

PR #17197, opened 2020-09-14
new package: zarith-freestanding 1.10

hannesm:


kit-ty-kate:
The package sounds good. I made a few improvements to the shell scripts, is that ok? If not I can always revert them.

hannesm:
@kit-ty-kate they're fine with me, thanks!

Issue #16780, opened 2020-07-07
conf-cmake: Uses bash instead of system shell; potentially incorrect on macOS 10.15


ospencer:
The `conf-cmake` package [uses bash to detect the presence of cmake](https://github.com/ocaml/opam-repository/blob/master/packages/conf-cmake/conf-cmake.1/opam#L7). macOS 10.15 changed the default shell from `bash` to `zsh`, so it's possible that a user may have a `.zshrc` that `bash` would not pick up, with alterations to the `PATH`.

This means that `cmake` can exist on the `PATH` for the system's default shell and not for `bash`, causing this script to fail even though `cmake` is installed.

Let me know if there's any additional info I can provide or let me know if there's anything I can do to help!

PR #17203, opened 2020-09-15
conf-cmake: Use /bin/sh instead of bash

kit-ty-kate:
Should fix https://github.com/ocaml/opam-repository/issues/16780


mseri:
The hash of configure.sh needs to be updated

kit-ty-kate:
Indeed. All done and green now.

PR #17207, opened 2020-09-15
Memtrace viewer (with dependencies)

xclerc:
This pull request releases `memtrace_viewer`.

Something perhaps noteworthy about this package is that
it vendors most of its dependencies (in particular to other
Jane Street packages). It is unconventional, but since releasing
this package implied a number of yet unpublished changes
to the packages, vendoring them allows us to publish the
memtrace viewer significantly earlier.

### DIFF
--- a/packages/memtrace_viewer_with_deps/memtrace_viewer_with_deps.v0.14.0/opam
+++ b/packages/memtrace_viewer_with_deps/memtrace_viewer_with_deps.v0.14.0/opam
@@ -17,6 +17,7 @@ remove: [
 ]
 depends: [
   "ocaml" {>= "4.11.0"}
+  "conf-libssl"
   "astring" {>= "0.8.5"}
   "base64" {>= "3.4.0"}
   "cryptokit" {>= "1.14"}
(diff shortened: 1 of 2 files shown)


lpw25:
@xclerc I think this package should be made available as just memtrace-viewer the with deps part is an implementation detail from the user's perspective.

packages/memtrace_viewer_with_deps/memtrace_viewer_with_deps.v0.14.0/opam
```diff
@@ -0,0 +1,56 @@
...
+license: "MIT"
+build: [
+  [make]
+]
+install: [
+  [make "install"]
+]
+remove: [
+  [make "uninstall"]
+]
```
kit-ty-kate, line 17: Could `dune build --profile=release @memtrace_viewer/install` be used directly for the build phase? Also could you remove the `install` and `remove` phase and instead add
```
  ["cp" "memtrace_viewer.install" "."]
```
to the build phase?
`dune install` requires ocamlfind at runtime and is not recommended in opam and `dune uninstall` is unnecessary with opam 2.0.

xclerc:
@lpw25 @kit-ty-kate https://github.com/ocaml/opam-repository/pull/17207/commits/2e2a8757a041f44ebd380451a8703c141d89b160 should address your concerns.

kit-ty-kate:
Thanks!

PR #17208, opened 2020-09-15
Kappa Tool Suite release 4.1

pirbo:


pirbo:
I've disabled the test-suite as the CI makes me realized it was way too fragile (it suppose float are always exactly computed/printed the same way which we know isn't true) (thanks (again) for providing this infra :-))

kit-ty-kate:
Looks good now. Could you return the missing constraint to dune? Your `dune-project` file indicates that at least dune 1.1 is required.

Thanks!

PR #17212, opened 2020-09-16
2 packages from project-everest/hacl-star at 0.2.2

victor-dumitrescu:
This pull-request concerns:
-`hacl-star.0.2.2`: OCaml API for EverCrypt/HACL*
-`hacl-star-raw.0.2.2`: Auto-generated low-level OCaml bindings for EverCrypt/HACL*

---
* Homepage: https://hacl-star.github.io/
* Source repo: git+https://github.com/project-everest/hacl-star.git
* Bug tracker: https://github.com/project-everest/hacl-star/issues

---

### DIFF
--- a/packages/hacl-star-raw/hacl-star-raw.0.2.2/opam
+++ b/packages/hacl-star-raw/hacl-star-raw.0.2.2/opam
@@ -1,0 +1,32 @@
+opam-version: "2.0"
+synopsis: "Auto-generated low-level OCaml bindings for EverCrypt/HACL*"
+description: """
+This package contains a snapshot of the EverCrypt crypto provider and
+the HACL* library, along with automatically generated Ctypes bindings.
+For a higher-level idiomatic API see the `hacl-star` package, of
+which `hacl-star-raw` is a dependency."""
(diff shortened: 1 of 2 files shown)


victor-dumitrescu:
Hi, is it no longer the case that the FreeBSD build is allowed to fail?

kit-ty-kate:
> Hi, is it no longer the case that the FreeBSD build is allowed to fail?

It makes very little sense for us to mark it as "allowed to fail" as it would just hide the failure if it's a genuine one. CI failures are always a case-by-case analysis anyway. In this case if you know it fails for a good reason it's fine to merge anyway. There is no hard blocks on CI failures (CIs are quite unstable beasts anyway)

Thanks!

victor-dumitrescu:
Thanks for merging! Indeed, the FreeBSD failure was expected.

kit-ty-kate:
When it is a structural issue with the package (e.g. use of linux-only syscalls, something that's not fixable by the user in any way, …) you can add `available: os = "linux" & os = "macos"` to your opam file to make it more explicit for instance.

PR #17214, opened 2020-09-16
[elpi] [win] use Digest rather than shasum

gares:
This makes the package work on windows with no change, CC @fdopen @msoegtropIMC
The patch is already applied upstream (in master)

### DIFF
--- a/packages/elpi/elpi.1.11.4-1/files/use-digest.patch
+++ b/packages/elpi/elpi.1.11.4-1/files/use-digest.patch
@@ -1,0 +1,132 @@
+commit 652e8a892963949d6d9245c3e5cb252010b14b78
+Author: Enrico Tassi <Enrico.Tassi@Inria.fr>
+Date:   Tue Sep 15 21:50:01 2020 +0200
+
+    backport to the released version
+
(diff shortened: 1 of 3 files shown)


MSoegtropIMC:
Hmm - this is lacking the temp file patch, so as is it wouldn't work on Windows. This version does work (in first tests) on macOS, Linux and Windows: (https://github.com/MSoegtropIMC/coq-platform/tree/v8.12/opam/packages/elpi/elpi.1.11.4)

gares:
Oh right, I'm adding that

gares:
The temp file patch is OK also on linux and OSX, I've tested it in CI and merged it upstream already

MSoegtropIMC:
Perfect, thanks!

kit-ty-kate:
```
- TIMEOUT  grundlagen (type checker for λδ)          elpi
```
Just so you know, I've seen this error quite a bit. Is there any way to get deterministic tests? It would be really good for us if it was possible.

Otherwise it's good to merge. Thanks

gares:
I guess the best would be to separate simple test from benchmarks. I'll try something.

PR #17215, opened 2020-09-16
[new release] merlin and dot-merlin-reader (3.4.0)

voodoos:

##### CHANGES:

Wed Sep 16 15:00:42 CEST 2020

  + merlin binary
    - fix completion of pattern matchings with exception patterns (ocaml/merlin#1169)
    - delegate configuration reading to external programs via a simple protocol
    and create a new package `dot-merlin-reader` with a binary that reads
    `.merlin` files. (ocaml/merlin#1123, ocaml/merlin#1152)

### DIFF
--- a/packages/merlin/merlin.3.4.0/opam
+++ b/packages/merlin/merlin.3.4.0/opam
@@ -1,0 +1,74 @@
+opam-version: "2.0"
+maintainer:   "defree@gmail.com"
+authors:      "The Merlin team"
+homepage:     "https://github.com/ocaml/merlin"
+bug-reports:  "https://github.com/ocaml/merlin/issues"
+dev-repo: "git+https://github.com/ocaml/merlin.git"
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" "1"] {with-test & ocaml:version >= "4.03"}
+]
+depends: [
+  "ocaml" {>= "4.02.3" & < "4.12"}
+  "dune" {>= "1.8.0"}
+  "dot-merlin-reader" {= "3.4.0"}
+  "yojson" {>= "1.6.0"}
+  "mdx" {with-test & >= "1.3.0"}
+  "conf-jq" {with-test}
+  "csexp" {>= "1.2.3"}
+  "result" {>= "1.5"}
+]
+synopsis:
+  "Editor helper, provides completion, typing and source browsing in Vim and Emacs"
+description:
+  "Merlin is an assistant for editing OCaml code. It aims to provide the features available in modern IDEs: error reporting, auto completion, source browsing and much more."
+post-messages: [
+  "merlin installed.
+
+Quick setup for VIM
+-------------------
+Append this to your .vimrc to add merlin to vim's runtime-path:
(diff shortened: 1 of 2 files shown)


voodoos:
This release is a bit peculiar as it introduced a new package `dot-merlin-reader` and `merlin` depends on it.
I am not sure how this will play with the CI ?

For your information, here is the anoucement I plan to make when the release do happen:

> # [ANN] Merlin 3.4.0 : introducing external configuration readers
> I am glad to announce, on behalf of the Merlin team, the release of Merlin
> `3.4.0` which bring some major changes in the way configuration is handled.
> As you might know, Merlin reads its configuration from the closest `.merlin`
> file to the source file being edited. These files tell merlin where to find
> other source files and build artifacts but also which flags should be passed to
> the compiler, which syntax extensions are enabled and which packages are used by
> the project.
> In this setting the configuration is the same for all the source files of a
> folder, regardless of their specificities. In other words, the configuration
> loaded for a single source file contains the union of the dependencies of this
> file and of all its siblings which is not an optimal behavior.
> Starting with version `3.4.0` merlin will ship with two packages: `merlin`
> and `dot-merlin-reader` which, as the name suggest, reads configuration from
> `.merlin` files. Both are necessary for proper function.
> When a `.merlin` file is present in the source folder the Merlin server will
> start a `dot-merlin-reader` process and communicate with it via standard input
> and output following a simple protocol. These processes are halted with the server.
> *This change should not have any visible impact on users' workflows as long as
> the `dot-merlin-reader` package is correctly installed and in the path*. (which
> should be the case in opam-based setups)
> This change in itself will not solve the granularity problem mentioned earlier,
> but it paves the way to such improvements: in a near-future Dune will stop
> generating `.merlin` files and Merlin will obtain its file-based configuration
> directly from the build system using the same protocol as the on used by
> `dot-merlin-reader`.

voodoos:
Do not merge yet, we just noticed that the csexp dependency force an upgrade to dune 2.7.

I am going to have a look if this constraint can be relaxed.

voodoos:
Thi can be released once `csexp 1.3.2` is: https://github.com/ocaml/opam-repository/pull/17225

packages/merlin/merlin.3.4.0/opam
```diff
@@ -0,0 +1,74 @@
...
+dev-repo: "git+https://github.com/ocaml/merlin.git"
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" "1"] {with-test & ocaml:version >= "4.03"}
+]
+depends: [
+  "ocaml" {>= "4.02.3" & < "4.12"}
+  "dune" {>= "1.8.0"}
+  "dot-merlin-reader" {= "3.4.0"}
```
kit-ty-kate, line 15: Note for next time that this can be replaced by `{= version}`. This is probably gonna be easier to maintain this way (unless you're planning to release both separately).

kit-ty-kate:
Thanks!

packages/merlin/merlin.3.4.0/opam
voodoos: We did not decide yet if we are going to release separately or not.
Hopefully `dot-merlin-reader` should change very little...

PR #17217, opened 2020-09-16
Constrain version of GT because of updated ppxlib

Kakadu:


Kakadu:
At the moment I see issues only related to camlp5...



@chetmurthy Can you add a constraint that camlp5 7.10 doesn't suit OCaml 4.09.1?

https://travis-ci.org/github/ocaml/opam-repository/jobs/727765324

kit-ty-kate:
thanks

chetmurthy:
Kakadu, I'll go thru and add the extra constraint to all the camlp5 versions in the 7.xx stream, >= 7.10.  And ensure that the next 7.13 release contains support for ocaml 4.09.1.

In any case, I've modified the config process so that when a bugfix version (e.g. 4.09.12) isn't supported, it defaults to the base minor version (4.09.0).

PR #17227, opened 2020-09-17
Add bitstring and ppx_bitstring v4.0.0

xguerin:


xguerin:
From what I can tell the failures are not related to `bitstring`.

kit-ty-kate:
> From what I can tell the failures are not related to `bitstring`.

All the failures are related:
* Here is a fix for FreeBSD: https://github.com/xguerin/bitstring/pull/23
* The revdep failures are all related to the fact that `bitstring.ppx` is not provided anymore. I assume this is an expected change, I'll add a commit to fix this in a minute.

kit-ty-kate:
>     * The revdep failures are all related to the fact that `bitstring.ppx` is not provided anymore. I assume this is an expected change, I'll add a commit to fix this in a minute.

Actually, after re-reading the result, this breaks every single packages that is currently using `bitstring`. Is that expected or do you prefer to re-add this for compatibility? It's not a blocker but I just want to be sure before.

xguerin:
I see. Thanks. Let me see what I can do.

kit-ty-kate:
Hm, I'm not entirely sure what happened here. Why has my commit been removed from your branch?

xguerin:
Not quite sure. It's on my master. I suspected a wrong hash/package path so I repushed everything.

xguerin:
@kit-ty-kate It does not look like the FreeBSD Travis worker is picking your change. Would the old package be cached somehow?

kit-ty-kate:
FreeBSD fails because it also tests the other packages changed in this PR and these now require bitstring < 4.0.0 which isn't compatible with FreeBSD.

Sorry for all the time this took to handle this PR. Thanks a lot for your work!

xguerin:
Thank you !

PR #17232, opened 2020-09-18
2 packages from Ninjapouet/slurp at 0.1.0

Ninjapouet:
This pull-request concerns:
-`slurp.0.1.0`: Sinatra Like URL Route Processing
-`slurp-cohttp.0.1.0`: Cohttp definition for Slurp

---
* Homepage: https://github.com/Ninjapouet/slurp
* Source repo: git+https://github.com/Ninjapouet/slurp.git
* Bug tracker: https://github.com/Ninjapouet/slurp/issues

---

### DIFF
--- a/packages/slurp-cohttp/slurp-cohttp.0.1.0/opam
+++ b/packages/slurp-cohttp/slurp-cohttp.0.1.0/opam
@@ -1,0 +1,37 @@
+opam-version: "2.0"
+synopsis: "Cohttp definition for Slurp"
+maintainer: "Julien Blond <julien.blond@gmail.com>"
+authors: "Julien Blond <julien.blond@gmail.com>"
+license: "Apache 2.0"
+tags: ["sinatra" "route" "http" "server" "cohttp"]
+homepage: "https://github.com/Ninjapouet/slurp"
+bug-reports: "https://github.com/Ninjapouet/slurp/issues"
+depends: [
+  "dune" {>= "2.7"}
+  "slurp" {>= "0.1.0"}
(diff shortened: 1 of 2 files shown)


packages/slurp-cohttp/slurp-cohttp.0.1.0/opam
```diff
@@ -0,0 +1,37 @@
...
+synopsis: "Cohttp definition for Slurp"
+maintainer: "Julien Blond <julien.blond@gmail.com>"
+authors: "Julien Blond <julien.blond@gmail.com>"
+license: "Apache 2.0"
+tags: ["sinatra" "route" "http" "server" "cohttp"]
+homepage: "https://github.com/Ninjapouet/slurp"
+bug-reports: "https://github.com/Ninjapouet/slurp/issues"
+depends: [
+  "dune" {>= "2.7"}
+  "slurp" {>= "0.1.0"}
```
kit-ty-kate, line 11: If you intend in releasing `slurp` and `slup-cohttp` together, I would suggest to change this constraint to `{= version}` as it will simplify future releases and avoid breakage of the previous versions.

Ninjapouet: Why being so restrictive the breakage will most likely append by other packages introduced by (>=) property; why being so restrictive ? Although the two packages are effectively intended to be released together, i often encountered the case where the binary was kept at version X but library at version Y > X  for the lib Y fixed some security issue but the binary could not be changed because of other lib dependency reasons.

PR #17236, opened 2020-09-19
Package shapefile.0.2.0

cyril-allignol:
### `shapefile.0.2.0`
A small library to read ESRI shapefiles

---
* Homepage: https://github.com/cyril-allignol/ocaml-shapefile
* Source repo: git+https://github.com/cyril-allignol/ocaml-shapefile.git
* Bug tracker: https://github.com/cyril-allignol/ocaml-shapefile/issues

---

### DIFF
--- a/packages/shapefile/shapefile.0.2.0/opam
+++ b/packages/shapefile/shapefile.0.2.0/opam
@@ -1,0 +1,14 @@
+opam-version: "2.0"
+synopsis: "A small library to read ESRI shapefiles"
+maintainer: "Cyril Allignol <cyril@allignol.com>"
+authors: ["Cyril Allignol <cyril@allignol.com>"]
+homepage: "https://github.com/cyril-allignol/ocaml-shapefile"
+bug-reports: "https://github.com/cyril-allignol/ocaml-shapefile/issues"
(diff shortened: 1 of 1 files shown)


cyril-allignol:
The Travis build does not seem to gather the sources for my package. I thought it was a URL error, thus I changed it to use git+https (which seems to be in use by many packages), but it does not fix the build issue.
In the log, the "Gathering sources" section is empty, then in the "Processing actions" section, the build system fails by telling it does not know about my "shapefile" package.
What am I missing?

kit-ty-kate:
The `url` section containing archive is missing from your opam file. `opam-publish` normally adds it automatically, however it somehow did not work here. Had you pushed the tag before you called opam publish? This can happen when the tag exists locally but had not been pushed upstream before calling opam publish?

Looking at your repository, the tag seems to be there now. Could you try to call opam publish again?

cyril-allignol:
Thanks for your help. opam-publish still does not add the url field, for some reason. I added the field manually, it now seems to work properly, except for the FreeBSD build, where an error on the bitstring package prevents the build of my package.

cyril-allignol:
Hum, OK, bitstring has changed back again to a separate package for syntax extension. I have to update my dune and opam files to depend on the recent version of bitstring. I close this PR and will open a new one with those fixes.

PR #17241, opened 2020-09-21
[new release] ocaml-compiler-libs (v0.12.3)

jeremiedimino:
OCaml compiler libraries repackaged

- Project page: <a href="https://github.com/janestreet/ocaml-compiler-libs">https://github.com/janestreet/ocaml-compiler-libs</a>

##### CHANGES:

- Update opam file


kit-ty-kate:
Dune is not compatible with the `{build}` tag. Could you return the fix into the opam file upstream?

Thanks!

jeremiedimino:
> Dune is not compatible with the {build} tag. Could you return the fix into the opam file upstream?

Sure, done

PR #17254, opened 2020-09-22
Renamed the ezcmdliner package to clim.

Ninjapouet:
Updated the version by the way for it's a significant change in the
package history.

### DIFF
--- a/packages/clim-ppx/clim-ppx.0.2.0/opam
+++ b/packages/clim-ppx/clim-ppx.0.2.0/opam
@@ -1,0 +1,39 @@
+opam-version: "2.0"
+synopsis: "Derivates cmdliner terms from type definitions"
+description:
+  "This PPX gives to your program a nice CLI using simple type annotations"
+maintainer: "Julien Blond <julien.blond@gmail.com>"
+authors: "Julien Blond <julien.blond@gmail.com>"
+license: "Apache 2.0"
+homepage: "https://github.com/Ninjapouet/clim"
+bug-reports: "https://github.com/Ninjapouet/clim/issues"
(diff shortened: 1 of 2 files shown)


avsm:
Thanks for the contribution!  We prefer not to delete packages from the opam-repo without a good reason, since other people may be depending on it. To do the rename, you can mark ezcmdliner as deprecated by adding a `messages` field to point users at the new package (and editing the `descr`).  If you could modify this PR to not delete ezcmdliner, we can go ahead and merge the new package.

Ninjapouet:
Well, the problem is that `ezcmdliner` is too confusing with `ez_cmdliner`. The two packages provide different layout above `cmdliner` and when I started to work on `ezcmdliner` the other wasn't released to opam and I though the name was free. When I PR the package, i didn't notice that OCamlPro also pushed `ez_cmdliner` few weeks ago. As the `ezcmdliner` package was just merged kit-ty-kate asked me to exceptionally do that [here](https://github.com/ocaml/opam-repository/pull/17178#issuecomment-696809638). I'm agree that deleting package is quite ugly but in this case, the solution would avoid a weird situation...

kit-ty-kate:
@avsm I had agreed on exceptionally removing this one. It was merged just 10 days ago, does not have any and was source of reasonable complains from the ocamlpro team.

avsm:
That sounds fine. Let's avoid that confusion then -- no objection from me to this renaming.

Ninjapouet:
No need to reopen this PR. While discussing, i move forward on the package with a 0.3.0 which is already in PR (with the old name package deletion).

PR #17269, opened 2020-09-24
Package shapefile.0.2.1

cyril-allignol:
### `shapefile.0.2.1`
A small library to read ESRI shapefiles

---
* Homepage: https://github.com/cyril-allignol/ocaml-shapefile
* Source repo: git+https://github.com/cyril-allignol/ocaml-shapefile.git
* Bug tracker: https://github.com/cyril-allignol/ocaml-shapefile/issues

---

### DIFF
--- a/packages/shapefile/shapefile.0.2.1/opam
+++ b/packages/shapefile/shapefile.0.2.1/opam
@@ -13,3 +13,7 @@ depends: [
   "ppx_bitstring" {>= "4.0.0"}
   "menhir" {build}
 ]
+url {
+  src: "https://github.com/cyril-allignol/ocaml-shapefile/archive/v0.2.1.tar.gz"
+  checksum: "md5=73b1c3a0d124e12d114a67164113953e"
+}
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Dune is not compatible with the `{build}` tag (see https://github.com/ocaml/dune/issues/2147) and `ocamlfind` was not used anywhere so I removed it.
I also explicitly added the minimal required version of dune and ocaml in the dependencies, it'll be more simple to change things in the future if it's more explicit.

Could you return these changes in the opam file upstream so it can be taken for future releases?

Thanks!

cyril-allignol:
> Dune is not compatible with the `{build}` tag (see [ocaml/dune#2147](https://github.com/ocaml/dune/issues/2147)) and `ocamlfind` was not used anywhere so I removed it.
> I also explicitly added the minimal required version of dune and ocaml in the dependencies, it'll be more simple to change things in the future if it's more explicit.
> 
> Could you return these changes in the opam file upstream so it can be taken for future releases?

Done in the master branch. Thanks again for your help.

PR #17272, opened 2020-09-24
[conf-gmp-powm-sec] fix test to avoid warning about undeclared function

blandinw:
Same issue as https://github.com/ocaml/opam-repository/pull/17156 and same fix.
The raison d'être of this package is to make sure the installed `libgmp` has `mpz_powm_sec()`, so the test also checks for it for completeness.


blandinw:
Tested with:
```
opam remote add local .

```

kit-ty-kate:
Thanks!

PR #17274, opened 2020-09-24
[new release] zlist (0.3.3)

hakuch:


kit-ty-kate:
This package fails with OCaml < 4.07:
```
# File "typing/reactivity_check.ml", line 75, characters 20-34:
# Error: Unbound module Stdlib
# make[1]: *** [../configure-tools/Makefile.common:54: typing/reactivity_check.cmx] Error 2
# make[1]: Leaving directory '/home/opam/.opam/4.04/.opam-switch/build/rml.1.09.06/compiler'
# make: *** [Makefile:9: all] Error 2
```
You can fix this by either adding a restriction e.g. `"ocaml" {>= "4.07"}` or depending on `stdlib-shims` and resubmit the release.

hakuch:
@kit-ty-kate, it's not clear to me that anything in the package would require OCaml >= 4.0.7.

I tried to install the package locally (by pinning) on a 4.03.0 switch:

- With `opam install ~/src/zlist`, the installation succeeded
- With `opam install -t ~/src/zlist`, the installation failed with this error, which is also the error I see in the failing CI log:

```
# File "test/dune", line 6, characters 7-17:
# 6 |   (pps ppx_expect)))
#            ^^^^^^^^^^
# Error: Ppx dependency on a non-ppx library "ppx_expect". If "ppx_expect" is
# in fact a ppx rewriter library, it should have (kind ppx_rewriter) in its
# dune file.

- File "test/dune", line 6, characters 7-17:
- 6 |   (pps ppx_expect)))
-            ^^^^^^^^^^
- Error: Ppx dependency on a non-ppx library "ppx_expect". If "ppx_expect" is
- in fact a ppx rewriter library, it should have (kind ppx_rewriter) in its
- dune file.
```

I'm experimenting in the mean time with different constraints on OCaml, but I'd appreciate more explanation on the origin on the 4.07 constraint that you're suggesting.

hakuch:
To expand on those experiments:

Speculating that an older version of `ppx_expect` was incompatible with the way Dune handled ppx libraries, I introduced a constraint

```
"ppx_expect" {with-test & >= "v0.11"}
```

but then opam complained that the version of OCaml expected by that version of `ppx_expect` was >= 4.04.2, so I added

```
"ocaml" {>= "4.04"}
```

resulting in:

```
depends: [
  "dune" {>= "2.7"}
  # Because of the version constraint on ppx_expect.
  "ocaml" {>= "4.04"}
  "odoc" {with-doc}
  "ppx_expect" {with-test & >= "v0.11"}
  "ppx_inline_test" {with-test}
]
```

and now `opam pin install -t ~/src/zlist` works on a 4.04 switch.

What do you think?

hakuch:
Thanks, @mseri! I appreciate your revision. I'm aiming for releases to be immutable, so I'd sooner publish a PR for 0.3.4 with the changes I inquired  about rather than modifying 0.3.3.

PR #17278, opened 2020-09-25
Add package for gappa - a proof engine for numeric (floating point) proofs

MSoegtropIMC:
Gappa is a tool to automatically generate proofs for numerical goals, e.g. error bounds for floating point approximations of a real expression.

This PR might be a bit controversial - gappa is written in C++. It is used in Coq, so it is convenient to have an opam package for it. I don't know how agreeable it is to add an opam packages for C++ code to the opam main repo, but since there is also a package for Z3, I thought I at least ask.

In case you don't want it, I can also add it to the Coq opam repo, but maybe you want some of the prerequisite conf packages.

### DIFF
--- a/packages/conf-automake/conf-automake.1/opam
+++ b/packages/conf-automake/conf-automake.1/opam
@@ -1,0 +1,33 @@
+opam-version: "2.0"
+maintainer: "7895506+MSoegtropIMC@users.noreply.github.com"
+homepage: "https://www.gnu.org/software/automake"
+authors: ["Jim Meyering" "David J. MacKenzie" "https://git.savannah.gnu.org/cgit/automake.git/tree/THANKS"]
+bug-reports: "https://www.gnu.org/software/automake/manual/html_node/Reporting-Bugs.html"
+license: "GPL-2.0+"
+build: [ "automake" "--help" ]
+depends: "conf-autoconf"
+depexts: [
+  # The list of supported os-family names can be derived from the opam sources (thanks Enrico)
+  #   https://github.com/ocaml/opam/blob/6aefe95e60084e63d01b1c7c028b6b77de1f839f/src/state/opamSysInteract.ml#L84-L123
+  # The package name for many Linux/BSD/macOS/cygwin/msys platforms can be searched here (thanks Théo)
+  #   https://repology.org specifically https://repology.org/project/automake/versions
+  #   Not listed above but supported by opam are suse, oraclelinix, ol, rhel
+  [ "automake" ] {os-family = "alpine"}
+  [ "automake" ] {os-family = "amzn"}
+  [ "automake" ] {os-family = "arch"}
+  [ "automake" ] {os-family = "archlinux"}
+  [ "automake" ] {os-family = "centos"}
+  # Note: although https://repology.org/project/automake/versions states debian has only automake packages
+  # with explicit versions, there is an "automake" package, at least on Ubuntu 18.04
+  [ "automake" ] {os-family = "debian"}
+  [ "automake" ] {os-family = "fedora"}
+  [ "sys-devel/automake" ] {os-family = "gentoo"}
+  [ "automake" ] {os-family = "homebrew"}
+  [ "automake" ] {os-family = "macports"}
+  [ "automake" ] {os-family = "mageia"}
+  [ "automake" ] {os-family = "opensuse"}
+  [ "devel/automake" ] {os-family = "bsd"}
+]
+synopsis: "Virtual package relying on GNU automake"
+description: "This package can only install if GNU automake is installed on the system."
+flags: conf
--- a/packages/conf-bison/conf-bison.1/opam
+++ b/packages/conf-bison/conf-bison.1/opam
@@ -1,0 +1,41 @@
+opam-version: "2.0"
+maintainer: "7895506+MSoegtropIMC@users.noreply.github.com"
+homepage: "https://www.gnu.org/software/bison/"
+bug-reports: "https://lists.gnu.org/mailman/listinfo/bug-bison"
+license: "GPL-3.0-or-later"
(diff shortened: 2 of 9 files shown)


mseri:
The old ubuntus are failing with

```
error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
#  #error This file requires compiler and library support \
```

I am not sure how easy/worth is to address it.

MSoegtropIMC:
@mseri : Since the user base for gappa is not that large, I would expect that the overlap between gappa users and Ubuntu 16 users is empty and won't increase in the future. So unless it is terribly difficult for you to give a waiver for this, I would indeed prefer to keep it like this.

What is a bit more disturbing is the failure on Fedora. To me this looks like the autoconf package on Fedora is broken - autoconf needs aclocal and this does not seem to be installed, although autoconf is. Should I add `aclocal` as depext to `conf-autoconf` and see what happens?

mseri:
> @mseri : Since the user base for gappa is not that large, I would expect that the overlap between gappa users and Ubuntu 16 users is empty and won't increase in the future. So unless it is terribly difficult for you to give a waiver for this, I would indeed prefer to keep it like this.

This is not a problem. I wanted to hear your opinion before deciding on the merge

> What is a bit more disturbing is the failure on Fedora. To me this looks like the autoconf package on Fedora is broken - autoconf needs aclocal and this does not seem to be installed, although autoconf is. Should I add `aclocal` as depext to `conf-autoconf` and see what happens?

If you don't mind we could have a `conf-autoconf.2` which adds `aclocal` to the depexts for Fedora. Let's see if that will fix it

MSoegtropIMC:
> If you don't mind to add it, we could have a conf-autoconf.2 which adds aclocal to the depexts for Fedora. Let's see if that will fix it. Thanks for the help.

Sure I will try that - I anyway wanted to setup a Fedora VM for my release tests.

I also though about Ubuntu 16 - actually the Coq platform project might make the installation base of gappa much larger - a Windows only predecessor project had more than 10.000 downloads a year. I will discuss with the Coq team how to handle this and either fix it or explicitly state that Ubuntu 16 or older is not supported.

I set this to work in progress until it is decided.

MSoegtropIMC:
@mseri : I analyzed the situation on Ubuntu and Fedora and I would say I should add an opam package for automake. On Ubuntu automake seems to be installed - possibly via the binutils package - on Fedora not. The opam conf-autoconf package I require installs just autoconf also on Ubuntu. The autoconf package recommends but does not require automake on Ubuntu. See:
(https://packages.ubuntu.com/bionic/autoconf). So in summary it is just luck that it works on Ubuntu - the prerequisites I gave are incomplete.

mseri:
Thanks for looking into it

MSoegtropIMC:
@mseri : I am still working on the Ubuntu16 fix (currently installing a VM).

mseri:
Wow, thank you very much for this!

MSoegtropIMC:
@mseri : I am thinking about dropping Ubuntu 16 (Xenial) support for these reasons:

- installing opam requires adding a non standard single package PPA to install the sandbox (bubble) or to disable the sandbox - I took the extra PPA approach with sandbox

- OCaml 4.07.1 does not compile once I get opam installed (switch creation fails)

But possibly I messed something up - how does the opam CI get as far as it does? Are these hand crafted images grown and patched over years? I used a plain fresh Ubuntu 16.04.7 with just build-essentials and bubble installed and would say opam is not supported at all on this platform, unless someone can give instructions on how get at least opam and OCaml 4.07.1 running there.

Possibly disabling the sandbox is the way to go?

mseri:
Not sure about the way the images are built, sorry. I will have to ping @kit-ty-kate for that. 
You are free to not support Xenial, any interested user can either install manually bubblewrap or use unsandboxed builds. This is not a problem

It is weird that ocaml 4.07.1 does not build though.

MSoegtropIMC:
> It is weird that ocaml 4.07.1 does not build though.

I didn't look into why OCaml 4.07.1 fails - opam is not terribly informative - e.g. does not provide a link to the logs. It just says one could try --jobs=1, but this doesn't help.

But since OCaml 4.07.1 was common when Xenial was common, it is indeed weird that it fails. If this helps, I can give the exact commands I issues on a fresh installed Ubuntu 16.

I would like to support gappa if Xenial is supported at all - something which is unclear to me.

mseri:
Xenial is supported, but you need to disable the sandbox or install bubblewrap. The error seems to be due to the old gcc version, see https://ci.ocaml.org/log/saved/docker-build-d15454418c83e6142f8d1d3ba9e4f84c/817d0b96f4d83b95c2dc1ce35f3e607e466bc2c6

Fwiw, I think this PR is ready to merge if you want

MSoegtropIMC:
> Fwiw, I think this PR is ready to merge if you want

Let me look into it today evening once more and then I will decide.

MSoegtropIMC:
I think I can fix it for Xenial until Monday. Please wait until then.

MSoegtropIMC:
A quick status note: I got opam installed on Ubuntu 16.04. I tried to install bubblewrap - it is not included in the Ubuntu package repo for 16.04, but there is an addon repo - but the version of bubblewrap got this way is too old and results in strange errors. I think minimum required is 0.2.1.

The tests for this PR on Ubuntu 16.04 are ongoing ...

MSoegtropIMC:
@mseri : this should work now - at least locally it does build now on Ubuntu 16.04.

mseri:
Thank you very much for this! It is good to merge

MSoegtropIMC:
> Thank you very much for this! It is good to merge

I say thank your for your openness to take this foreign language stuff in!

PR #17279, opened 2020-09-25
Added an existence check to conf-gnome-icon-theme3

MSoegtropIMC:
The PR adds a check if the gnome icon theme is actually installed, so that the package fails in case it is not installed (as is usually advertised for a conf package). As is the package has just depext dependencies but happily continues in case the icon theme is not installed.

### DIFF
--- a/packages/conf-adwaita-icon-theme/conf-adwaita-icon-theme.1/opam
+++ b/packages/conf-adwaita-icon-theme/conf-adwaita-icon-theme.1/opam
@@ -1,0 +1,36 @@
+opam-version: "2.0"
+maintainer: "7895506+MSoegtropIMC@users.noreply.github.com"
+homepage: "https://github.com/adwaita/adwaita-icon-theme"
+bug-reports: "https://gitlab.adwaita.org/adwaita/adwaita-icon-theme/issues"
+authors: "GNU Project"
+license: ["LGPL-3.0-only" "CC-BY-SA-3.0"]
+# E.g. on Fedora the system package does not contain a .pc file.
+# Testing for this folder seems to be more portable than using system packager commands as fallback.
+build: [ "sh" "-c" "pkg-config --short-errors --print-errors adwaita-icon-theme || [ -d /usr/share/icons/Adwaita ]"]
+depends: [
+  "conf-pkg-config" {build}
+]
+depexts: [
+  # The list of supported os-family names can be derived from the opam sources
+  #   https://github.com/ocaml/opam/blob/6aefe95e60084e63d01b1c7c028b6b77de1f839f/src/state/opamSysInteract.ml#L84-L123
+  # The package name for many Linux/BSD/macOS/cygwin/msys platforms can be searched here
+  #   https://repology.org specifically https://repology.org/project/adwaita-icon-theme/versions
+  #   Not listed above but supported by opam are suse, oraclelinix, ol, rhel
+  [ "adwaita-icon-theme" "adwaita-icon-theme-dev" ] {os-family = "alpine"}
+  [ "adwaita-icon-theme" ] {os-family = "amzn"}
+  [ "adwaita-icon-theme" ] {os-family = "arch"}
+  [ "adwaita-icon-theme" ] {os-family = "archlinux"}
+  [ "adwaita-icon-theme" ] {os-family = "centos"}
+  [ "adwaita-icon-theme" ] {os-family = "debian"}
+  [ "adwaita-icon-theme" ] {os-family = "fedora"}
+  [ "x11-themes/adwaita-icon-theme" ] {os-family = "gentoo"}
+  [ "adwaita-icon-theme" ] {os-family = "homebrew"}
+  [ "adwaita-icon-theme" ] {os-family = "macports"}
+  [ "adwaita-icon-theme" ] {os-family = "mageia"}
+  [ "adwaita-icon-theme" ] {os-family = "opensuse"}
+  [ "x11/gnome/adwaita-icon-theme" ] {os-family = "bsd"}
+]
+synopsis: "Virtual package relying on adwaita-icon-theme"
+description:
+  "This package can only install if the adwaita-icon-theme package is installed on the system."
+flags: conf
--- a/packages/conf-gnome-icon-theme3/conf-gnome-icon-theme3.0/opam
+++ b/packages/conf-gnome-icon-theme3/conf-gnome-icon-theme3.0/opam
@@ -3,6 +3,8 @@ homepage: "https://github.com/GNOME/adwaita-icon-theme"
 bug-reports: "https://gitlab.gnome.org/GNOME/adwaita-icon-theme/issues"
 authors: "GNU Project"
 license: ["LGPL-3.0-only" "CC-BY-SA-3.0"]
+build: [["pkg-config" "--short-errors" "--print-errors" "adwaita-icon-theme"]]
+depends: ["conf-pkg-config" {build}]
 depexts: [
   ["gnome-icon-theme"] {os-family = "debian"}
   ["gnome-icon-theme"] {os-family = "fedora" | os-family = "rhel"}
--- a/packages/conf-gnome-icon-theme3/conf-gnome-icon-theme3.0/opam
+++ b/packages/conf-gnome-icon-theme3/conf-gnome-icon-theme3.0/opam
@@ -9,7 +9,7 @@ depends: [
 ]
 depexts: [
   ["gnome-icon-theme"] {os-family = "debian"}
-  ["gnome-icon-theme"] {os-family = "fedora" | os-family = "rhel"}
+  ["gnome-icon-theme-devel"] {os-family = "fedora" | os-family = "rhel"}
   ["gnome-icon-theme"] {os = "macos" & os-distribution = "homebrew"}
   ["adwaita-icon-theme"] {os = "macos" & os-distribution = "macports"}
   ["gnome-icon-theme"] {os-family = "suse"}
--- a/packages/conf-gnome-icon-theme3/conf-gnome-icon-theme3.0/opam
+++ b/packages/conf-gnome-icon-theme3/conf-gnome-icon-theme3.0/opam
@@ -4,14 +4,10 @@ bug-reports: "https://gitlab.gnome.org/GNOME/adwaita-icon-theme/issues"
 authors: "GNU Project"
 license: ["LGPL-3.0-only" "CC-BY-SA-3.0"]
 build: ["pkg-config" "--short-errors" "--print-errors" "gnome-icon-theme"]
-depends: [
-  "conf-pkg-config" {build}
-]
+depends: "conf-pkg-config" {build}
 depexts: [
   ["gnome-icon-theme"] {os-family = "debian"}
   ["gnome-icon-theme-devel"] {os-family = "fedora" | os-family = "rhel"}
-  ["gnome-icon-theme"] {os = "macos" & os-distribution = "homebrew"}
-  ["adwaita-icon-theme"] {os = "macos" & os-distribution = "macports"}
   ["gnome-icon-theme"] {os-family = "suse"}
   ["gnome-icon-theme"] {os-distribution = "arch"}
   ["gnome-icon-theme"] {os-distribution = "alpine"}
@@ -23,4 +19,6 @@ depexts: [
 synopsis: "Virtual package relying on gnome-icon-theme"
 description:
   "This package can only install if the gnome-icon-theme package is installed on the system."
+post-messages:
+  "The package 'conf-gnome-icon-theme3' is deprecated. Please use 'conf-adwaita-icon-theme' instead!"
 flags: conf
--- a/packages/conf-gnome-icon-theme3/conf-gnome-icon-theme3.0/opam
+++ b/packages/conf-gnome-icon-theme3/conf-gnome-icon-theme3.0/opam
@@ -11,12 +11,14 @@ depexts: [
   ["gnome-icon-theme"] {os-family = "debian"}
   ["gnome-icon-theme"] {os-family = "fedora" | os-family = "rhel"}
   ["gnome-icon-theme"] {os = "macos" & os-distribution = "homebrew"}
+  ["adwaita-icon-theme"] {os = "macos" & os-distribution = "macports"}
   ["gnome-icon-theme"] {os-family = "suse"}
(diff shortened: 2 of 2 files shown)


MSoegtropIMC:
I need a little help with your CI. On Ubuntu 18.04, MacOS+MacPorts and Cygwin+MinGW this works fine. The logs say that the icon theme package is installed using depext but then:
```
Processing  3/4: [conf-gnome-icon-theme3: pkg-config adwaita-icon-theme]
+ /usr/bin/pkg-config "--short-errors" "--print-errors" "adwaita-icon-theme" (CWD=/home/opam/.opam/4.02/.opam-switch/build/conf-gnome-icon-theme3.0)
- No package 'adwaita-icon-theme' found
```
fails. This is odd. Can you tell me on what Linux distro and version the CI runs, so that I can check why this fails? Possibly the pkg-config name of the icon theme is different on different distros (which would also be odd for a basic gnome package) or the icon theme does not have a pkg-config meta file on some distros.

mseri:
I can see it fail in debian buster, centos and freebsd (which may not be supported?)

kit-ty-kate:
@MSoegtropIMC the issue was that apparently `gnome-icon-theme` and `adwaita-icon-theme` are not the same thing. On all the distros expect ubuntu 16.04 one does not install the other. I changed to the right name.

I also removed the macports package as `gnome-icon-theme` does not exist there.

MSoegtropIMC:
@kit-ty-kate : thanks! I will check what is going on on free-bsd and MacOS (I am mostly working on 10.15 with MacPorts and this works - I guess the test uses homebrew).

kit-ty-kate:
This should fix all the issue. I cannot test MacPorts but if you say it works this it good!

MSoegtropIMC:
@kit-ty-kate : let me merge this back to my coq platform opam patch repo and trigger a build on macOS with MacPorts. Please hold back the merge until I report the results here (will take a few hours).

kit-ty-kate:
sure, we need to wait for our CI as well anyway (should be done tomorrow)

MSoegtropIMC:
@kit-ty-kate : looking into this I am not sure I understand all implications. Here are a few links:

from this opam file: bug-reports: (https://gitlab.gnome.org/GNOME/adwaita-icon-theme/issues)

macOS packages - a gnome-icon-theme exists on neither
(https://ports.macports.org/port/adwaita-icon-theme/summary)
(https://formulae.brew.sh/formula/adwaita-icon-theme)

Repology overview for both packages:
(https://repology.org/project/adwaita-icon-theme/versions)
(https://repology.org/project/gnome-icon-theme/versions)

Do you understand what the respective packages are? Is it so that "adwaita-icon-theme" is a possible instance for a "gnome-icon-theme" and "gnome-icon-theme" is a meta package which installs some gnome icon theme - usually "adwaite-icon-theme" (afaik there is only adwaita and hicolor"). But then I don't understand the pkg-config part of it ...

Or is there really something which is the "gnome-icon-theme"? Afaik Coq explicitly needs the Adwaita icon theme because it uses icons which don't exist in the hicolor scheme.

MSoegtropIMC:
P.S.: I have meanwhile VMs of the major Linux distros where I can look up and test things - except for FreeBSD where I failed to setup a system with GTK (which is somehow required to test this). At first glance FreeBSD looks like a console only system so one might say this is not available on FreeBSD. But I saw that there is a FreeBSD derivate, which seems to have a Gnome UI - unfortunately there image download server is quite slow.

MSoegtropIMC:
@kit-ty-kate : on Mac with MacPorts this fails because the pkg-config name is "adwaita-icon-theme" and on Homebrew (I guess) it fails because the given package does not exist.

How shall we proceed? How about having separate packages for "gnome-icon-theme" and "adwaita-icon-theme"?

mseri:
> How shall we proceed? How about having separate packages for "gnome-icon-theme" and "adwaita-icon-theme"?

You can have different build commands for homebrew and for macports in an opam file, that is not a problem

MSoegtropIMC:
@kit-ty-kate : thinking about this I would say we should move this package to "adwaita-icon-theme". The header of the opam package says this is what it does. As far as I can tell this exists on all major platforms. On Gentoo the package name should be "x11-themes/adwaita-icon-theme".

MSoegtropIMC:
> You can have different build commands for homebrew and for macports in an opam file, that is not a problem

They should be the same - they both call it "adwaita-icon-theme". Right now the package names are different, so they fail for different reasons.

kit-ty-kate:
I had a look in the changelog of the debian package and it seems `adwaita-icon-theme` is simply a change of name of `gnome-icon-theme` that occured in gnome 3.13 (3.14's dev cycle). I had a look in coqide (master) and it seems that only `dialog-warning`, `cancel`, `bold`, `italic`, `underline`, `strikethrough`, `save`, `new`, `close` and `ok` are required and all of them seem to be in gnome-icon-theme.

However it might be an issue later as it is not up-to-date. I propose we make `conf-gnome-icon-theme3` deprecated (a simple post-message alerting of the deprecation should be enough) and create a new `conf-adwaita-icon-theme` package that future versions of coqide should use instead.

What do you think? Is there something I missed?

MSoegtropIMC:
> I had a look in coqide (master) and it seems that only dialog-warning, cancel, bold, italic, underline, strikethrough, save, new, close and ok

@kit-ty-kate : coqide uses quite a few more icons, partly via pre defined dialogs (say file type icons in file open dialogs) but also its main toolbar has quite a few navigation icons.

> What do you think? Is there something I missed?

I would say we should silently move this package to adwaita. I believe this leads to less maintenance effort overall - projects build fine with the wrong icon scheme but don't run or have missing icons. Since Adwaita is a superset of previous icon sets it should not hurt and afaik GTK selects adwaita if it is there. Since the icons don't need header files or the like, I don't think many projects have explicit tests for the icon scheme. I have this because I am too lazy to handle issues about missing icons.

But I am also fine with your suggestion - deprecate this package and have a new one. For me it doesn't matter.

So which way do we go? Do you want to do it or shall I do it? Either way I will help with testing.

MSoegtropIMC:
In order to get forward on this I will create a new package names "conf-adwaita-icon-theme" as suggested by @kit-ty-kate and mark the existing package as deprecated.

packages/conf-adwaita-icon-theme/conf-adwaita-icon-theme.1/opam
```diff
@@ -0,0 +1,36 @@
...
+  [ "adwaita-icon-theme" "adwaita-icon-theme-dev" ] {os-family = "alpine"}
+  [ "adwaita-icon-theme" ] {os-family = "amzn"}
+  [ "adwaita-icon-theme" ] {os-family = "arch"}
+  [ "adwaita-icon-theme" ] {os-family = "archlinux"}
+  [ "adwaita-icon-theme" ] {os-family = "centos"}
+  [ "adwaita-icon-theme" ] {os-family = "debian"}
+  [ "adwaita-icon-theme" ] {os-family = "fedora"}
+  [ "x11-themes/adwaita-icon-theme" ] {os-family = "gentoo"}
+  [ "adwaita-icon-theme" ] {os-family = "homebrew"}
+  [ "adwaita-icon-theme" ] {os-family = "macports"}
```
mseri, line 28: Don't you need `{os = "macos" & os-distribution = "macports"}` and `{os = "macos" & os-distribution = "homebrew"}` here?

MSoegtropIMC:
@kit-ty-kate : I added a new conf package "conf-adwaita-icon-theme" and deprecated "conf-gnome-icon-theme3" via post message. I also removed homebrew and Macports from the latter since the package exists on neither.

On some platforms (e.g. Fedora) there is no pkg-config file for the icon theme (on most platforms there is), so I added a fallback check for the literal path where it usually is. If you have a better idea, please let me know, but I think the combination of both is unlikely to fail since the folder is quite standard.

packages/conf-adwaita-icon-theme/conf-adwaita-icon-theme.1/opam
```diff
@@ -0,0 +1,36 @@
...
+  "conf-pkg-config" {build}
+]
+depexts: [
+  # The list of supported os-family names can be derived from the opam sources
+  #   https://github.com/ocaml/opam/blob/6aefe95e60084e63d01b1c7c028b6b77de1f839f/src/state/opamSysInteract.ml#L84-L123
+  # The package name for many Linux/BSD/macOS/cygwin/msys platforms can be searched here
+  #   https://repology.org specifically https://repology.org/project/adwaita-icon-theme/versions
+  #   Not listed above but supported by opam are suse, oraclelinix, ol, rhel
+  [ "adwaita-icon-theme" "adwaita-icon-theme-dev" ] {os-family = "alpine"}
+  [ "adwaita-icon-theme" ] {os-family = "amzn"}
```
mseri, line 20: I see that there is a new `amzon` and there is no longer `ol`. Just curious, I have no idea what either of these is :P

packages/conf-adwaita-icon-theme/conf-adwaita-icon-theme.1/opam
```diff
@@ -0,0 +1,36 @@
...
+  [ "adwaita-icon-theme" ] {os-family = "arch"}
+  [ "adwaita-icon-theme" ] {os-family = "archlinux"}
```
mseri, line 22: Do you need both these lines? Isn't `["gnome-icon-theme"] {os-distribution = "arch"}` enough?

packages/conf-adwaita-icon-theme/conf-adwaita-icon-theme.1/opam
```diff
@@ -0,0 +1,36 @@
+opam-version: "2.0"
+maintainer: "7895506+MSoegtropIMC@users.noreply.github.com"
+homepage: "https://github.com/adwaita/adwaita-icon-theme"
+bug-reports: "https://gitlab.adwaita.org/adwaita/adwaita-icon-theme/issues"
+authors: "GNU Project"
+license: ["LGPL-3.0-only" "CC-BY-SA-3.0"]
+# E.g. on Fedora the system package does not contain a .pc file.
+# Testing for this folder seems to be more portable than using system packager commands as fallback.
+build: [ "sh" "-c" "pkg-config --short-errors --print-errors adwaita-icon-theme || [ -d /usr/share/icons/Adwaita ]"]
```
mseri, line 9: Neat! Does `[ -d /usr/share/icons/Adwaita ]` work also on `sh` (when it is not bash)?

packages/conf-adwaita-icon-theme/conf-adwaita-icon-theme.1/opam
MSoegtropIMC: I am not sure. I got the OS-family values systematically from the opam sources, and this specific one from:

(https://github.com/ocaml/opam/blob/6aefe95e60084e63d01b1c7c028b6b77de1f839f/src/state/opamSysInteract.ml#L108)

But maybe I should look for the back translation of `type families` - I am not sure how this works.

packages/conf-adwaita-icon-theme/conf-adwaita-icon-theme.1/opam
MSoegtropIMC: Well I call "sh" explicitly here, so it should be sh and not bash - the right part is tested on Fedora and works. But to make sure I started a sh and tried it and it seems to work.

packages/conf-adwaita-icon-theme/conf-adwaita-icon-theme.1/opam
MSoegtropIMC: I got this from:
(https://github.com/ocaml/opam/blob/a44bdd5e12585179167ba62602f537484c04c268/src/state/opamSysInteract.ml#L103-L117)
but I am not sure if I should take the strings which go in or if there is a back translation of `type families`.

The `suse, oraclelinix, ol, rhel` variants I did not list because (https://repology.org/project/adwaita-icon-theme/versions) does not list package names for these variants, possibly because they are commercial.

MSoegtropIMC:
I see that `camelus` CI has errors, but there is no details button ...

packages/conf-adwaita-icon-theme/conf-adwaita-icon-theme.1/opam
MSoegtropIMC: I would think that the os-families homebrew and macports are unique. I think one should use the os-families as most generic classes unless one needs to further break it down - I guess this is the reason d'être for the OS families. So I looked up in the opam sources what exists and went from there. But quite possibly this is a missunderstading on my side.

The package names I get from repology - as stated in the comment. I actually test on Macports, Ubuntu 18 and 16 and Fedore 32.

MSoegtropIMC:
This is still failing on CentOS, OpenSuse and FreeBSD - I can install a VM and try to fix this.

With FreeBSD I must say I am not sure how to install a version with UI - the usual setup packages seem to be console only servers. Obviosuly a console only system doesn't have packages for icons. Should I just remove the support for FreeBSD?

For CentOS and OpenSuse I don't have a VM as yet, but I can create one.

MSoegtropIMC:
> Should I just remove the support for FreeBSD?

The alternative is: are you aware of a FreeBSD like distro with gnome UI I can try?

mseri:
No idea about freebsd. The other failures I think are due to the versions of the respective OS: for example I could find `adwaita-icon-theme` in the latest opensuse

mseri:
I don't know what is wrong with camelus, but it could have to do with one of those packages missing from its environment.
Thank you very much for the effort! I think this is fine to merge, @kit-ty-kate do you agree?

MSoegtropIMC:
@mseri : I am open to fixing this on OpenSuse and CentOS - I think these are sufficiently important distros. As I mentioned I am working on an opam based setup procedure for Coq and this is a dependency for its IDE. This setup procedure is expected to be widely used e.g. in teaching so wide distro support is a good thing and I do appreciate that opam CI tests a lot of distros. I must admit I was a bit naive regarding Linux compatibility.

The issue might be minor details in the package names. I got these from repology.

Let me look into this today and tomorrow. It would not hurt to learn how he package name repology reports relate to the package names I have to give in depext. After that I might write some wiki page on how to write depext packages effectively.

mseri:
I am going to merge. We can make any extra tweak as a separate PR

mseri:
Thanks again for the help!

PR #17282, opened 2020-09-25
Update lablgtk3.3.0.beta5 (still required for Windows compatibility)

MSoegtropIMC:
This PR does a few minor updates to lablgtk3.3.0.beta5:
- update required dune version
- make it dependent on conf-gtksourceview3 (otherwise it fails e.g. on Mac where this is not commonly installed)

The old beta5 package is still required for Windows compatibility, because newer packages have still unresolved issues with flexlink (or the system linker).

### DIFF
--- a/packages/lablgtk3/lablgtk3.3.0.beta5/opam
+++ b/packages/lablgtk3/lablgtk3.3.0.beta5/opam
@@ -17,11 +17,10 @@ doc: "https://garrigue.github.io/lablgtk/lablgtk3"
 
 depends: [
   "ocaml"     {         >= "4.05.0" & < "4.10" }
-  "dune"      {         >= "1.4.0"
-                      & != "1.7.0"
-                      & != "1.7.1"  } # Due to dune/dune#1833
+  "dune"      {         >= "1.8.0"  }
   "dune-configurator"
   "conf-gtk3" { build & >= "18"     }
+  "conf-gtksourceview3" { build     }
   "cairo2"    {         >= "0.6"    }
 ]
 


mseri:
Can we move the changes to a new `beta5-1` Version instead?

packages/lablgtk3/lablgtk3.3.0.beta5/opam
```diff
@@ -17,11 +17,10 @@ doc: "https://garrigue.github.io/lablgtk/lablgtk3"
 
 depends: [
   "ocaml"     {         >= "4.05.0" & < "4.10" }
-  "dune"      {         >= "1.4.0"
-                      & != "1.7.0"
-                      & != "1.7.1"  } # Due to dune/dune#1833
+  "dune"      {         >= "1.8.0"  }
   "dune-configurator"
   "conf-gtk3" { build & >= "18"     }
+  "conf-gtksourceview3" { build     }
```
kit-ty-kate, line 23: I'm unsure why this is required. Could you elaborate?

MSoegtropIMC: lablgtk also contains an OCaml wrapper for gtksourceview3. If this is not installed, it will fail. I am not so sure why I put in the "build", though - likely I just copied it from the conf-gtk3. Obviously both are required for build and run, unless one assumes that these packages just install the dev headers, which is not the case on Mac and Windows. On Linux the runtime is installed by default, so it would suffice to install the dev headers. Maybe this is why "build" is written here.

Is there a policy if a conf package for a library is the runtime or dev version?

MSoegtropIMC:
> Can we move the changes to a new beta5-1 Version instead?

Not sure what beta5-1 is - I don't see this in master. There is a beta6..beta8 though. As far as I remember, the Windows build works up to including beta6 but not for beta7 and also not for the release versions, but I can try if beta6 works. I kept beta5 in my dependencies because this is well tested (unlike beta6). Some day I try to fix the issues with the release version - it looks non trivial (some bizarre linker issue).

mseri:
What I mean is, can you copy lablgtk3.3.0.beta5 Into lablgtk3.3.0.beta5-1 and make those changes in this new lablgtk3.3.0.beta5-1 version instead?

mseri:
I have manually moved it to a patch version and merged

PR #17288, opened 2020-09-26
[new release] odepack (0.7.1)

Chris00:
CHANGES:

- Compile the FORTRAN code in "legacy" mode.
- Upgrade to dune-configurator.

### DIFF
--- a/packages/odepack/odepack.0.7.1/opam
+++ b/packages/odepack/odepack.0.7.1/opam
@@ -10,13 +10,13 @@ tags: [ "ODE" "scientific-computing"  ]
 build: [
   ["dune" "subst"] {pinned}
   ["dune" "build" "-p" name "-j" jobs]
-  ["dune" "build" "@doc"] {with-doc}
+  ["dune" "build" "@doc" "-p" name "-j" jobs] {with-doc}
   ["dune" "runtest" "-p" name "-j" jobs] {with-test}
 ]
 depends: [
   "ocaml" {>= "4.02"}
   "base-bigarray"
-  "dune"
+  "dune" {>= "1.1"}
   "dune-configurator"
   "base-bytes" {build}
   "conf-gfortran" {build}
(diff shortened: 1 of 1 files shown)


Chris00:
Does [conf-gfortran](https://github.com/ocaml/opam-repository/blob/master/packages/conf-gfortran/conf-gfortran.0/opam) need to be modified for FreeBSD?

mseri:
It shouldn’t. `Lang/gcc` is a metapackge including all of c, c+c and gfortran compilers. The conf package installs fine, I wonder if odepack need some extra includes at compile time to work — this I don’t know

mseri:
@kit-ty-kate you have a test freebsd machine, right? Could you have a look?

kit-ty-kate:
It seems you already fixed the issue. One other way of fixing this would have also been to use `gcc` directly instead of the default compiler (`cc`) as it already contains the path to the gcc library `libgfortran`. This way you would not even have to use `gfortran --print-file-name`. But it sounds good now anyway, thanks ^_^

Chris00:
Thanks or looking at it @kit-ty-kate!

Chris00:
BTW, if my memory serves, we should use `dune-configurator` now but I can't find a link to the information.  Can you confirm?

Chris00:
Found [this commit](https://github.com/ocaml/opam-repository/commit/d7e7f626e6a8c6bd43d03115fe465e66cad3cc75), so I guess it is OK.

mseri:
Sorry for the late replay. That is correct. Since dune 2.0 configurator is released separately with each dune release

Chris00:
@mseri Thanks for the confirmation.  Do not know why the [manual](https://dune.readthedocs.io/en/stable/) was [not updated](https://github.com/ocaml/dune/pull/3822).

kit-ty-kate:
Thanks a lot! I fixed your opam file here as your `dune-project` file indicates this package requires at least dune 1.1 and the `-p <name> -j <jobs>` was missing from the `dune build @doc` command.
Could you return those fixes upstream for next time?

Chris00:
Thanks.  Your changes are [merged in the main repo](https://github.com/Chris00/ocaml-odepack/commit/29a554ffc713226f78c6b1d83e11a248ede58fcf).

PR #17289, opened 2020-09-26
ppx_deriving_rpc: fix compatibility with recent ppxlib

mseri:
There was a transitive dependency on base


mseri:
There was already an upper bound on ppxlib. I will release a fixed version soon

PR #17303, opened 2020-09-28
pa_ppx Version 0.03

chetmurthy:
* [27 Sep 2020] added TRIP-TEST-ALL to generate and save
  generated-files (and run tests) for all supported ocaml versions.

* [24 Sep 2020] new way of selecting generated-src versions.
  hopefully this will make pa_ppx immune to "new camlp5 release, we
  can't build" problems.

  pa_ppx.import now support "mli-only" and "redeclare" options:

  "-pa_import-mli-only": to ensure that only MLI files are consulting during import
  (e.g. to pull in OCaml AST from an older OCaml version, and avoid
  doing so from the stdlib)

  "-pa_import-redeclare": to import types and *not* add the type-equation that
  binds them to the existing type.  For instance, to combine with
  mli-only (since there isn't any existing type that matches).

  cache files as they're read in pa_import, so we don't need to reread them.

  the [%%import ...] item-extension (for str-items and sig-items) now takes a
  [@add ...] attribute, viz.

  [%%import: MLast.expr
    [@add type loc = [%import: MLast.loc]
          and type_var = [%import: MLast.type_var]
          and 'a vala = [%import: 'a Ploc.vala]
    ]
  ]

  This says to import the entire typedecl-group in which MLast.expr is
  declared, and to add to that the types MLast.type_var and Ploc.vala,
  all in a single grouped typedef (which will be recursive, if the
  initial typedef is).  The "@add" attribute payload is a str-item, so
  it can itself have a payload that is an item-extension ("[%%import
  ...]") in order to import-and-add entire typedecl groups, thus:

  [%%import: MLast.expr
    [@add [%%import: MLast.loc]]
    [@add [%%import: MLast.type_var]]
    [@add [%%import: 'a Ploc.vala]]
    [@with Ploc.vala := vala]
  ] [@@deriving show]

  The "@with" directive applies to all types imported.  If we wanted
  to apply it to one of the types in "@add" directives (say, for
  "MLast.type_var"), we'd just put it in that "%%import" attribute in
  the normal way.

* [12 Sep 2020] added a bunch of support for matching & substitution over types,
  to support pa_ppx_migrate (which is now in a separate project).

  also add generated files for camlp5 8.00~alpha04

* [22 Aug 2020] cleanup opam file, bump version for compat with camlp5 8.00~alpha03.

  Change versioning process.

### DIFF
--- a/packages/pa_ppx/pa_ppx.0.03/opam
+++ b/packages/pa_ppx/pa_ppx.0.03/opam
@@ -1,0 +1,87 @@
+version: "0.03"
+synopsis: "PPX Rewriters for Ocaml, written using Camlp5"
+description:
+"""
+This is a collection of PPX rewriters, re-implementing those based on ppxlib
+and other libraries, but instead based on Camlp5.  Included is also a collection
+of support libraries for writing new PPX rewriters.  Included are:
+
+pa_assert: ppx_assert
+pa_ppx.deriving, pa_ppx.deriving_plugins (enum, eq, fold, iter, make, map, ord, sexp, show, yojson):
+  ppx_deriving, plugins, ppx_sexp_conv, ppx_deriving_yojson
+pa_ppx.expect_test: ppx_expect_test
+pa_ppx.here: ppx_here
+pa_ppx.import: ppx_import
+pa_ppx.inline_test: ppx_inline_test
+
+pa_ppx.undo_deriving: pa_ppx.deriving expands [@@deriving ...] into code; this rewriter undoes that.
+pa_ppx.unmatched_vala: expands to match-cases (support library for camlp5-based PPX rewriters)
+pa_ppx.hashrecons: support for writing AST rewriters that automatically fills in hash-consing boilerplate
+pa_dock: implements doc-comment extraction for camlp5 preprocessors
+
+Many of the reimplementations in fact offer significant enhanced
+function, described in the pa_ppx documentation.  In addition, there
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
The tests on this package fails on OCaml 4.10. Could you have a look at it?

Full log: https://ci.ocaml.org/log/saved/docker-run-333c9ada8c83e0593ffb9031db5d0187/4d2d1573e7a4488cedd0cf78c163196e168e3bc9

chetmurthy:
How odd.  The failures are test-failures that are known to happen, and have been that way for several releases now.  I'm surprised that they're failing on only this one test-run, rather than every time pa_ppx is installed during CI.  Is it the case that sometimes CI installs packages -with- "-t" and sometimes -without- "-t" ?

I will (of course) disable running these tests during "-t" -- that's not a problem at all.

kit-ty-kate:
> Is it the case that sometimes CI installs packages -with- "-t" and sometimes -without- "-t" ?

Our CI will always try to install both with and without tests whenever possible, if that's what you're asking.

chetmurthy:
Excellent!  That's what I was wondering.  I have another package that I haven't yet released, that has a quite voluminous test-suite.  I've moved it so it doesn't get run by "install -t".  Thanks!

chetmurthy:
Hi, is there anything else I need to do, for this PR?

mseri:
Thanks!

PR #17316, opened 2020-09-29
+angstrom.0.15.0

seliopou:


seliopou:
Looks like the macOS build just needs a kick.

kit-ty-kate:
Could you return these fixes upstream? The constraint did not match the requirement indicated in the dune-project file

mseri:
Looks good now, thanks

PR #17322, opened 2020-09-29
[new release] conformist (0.1.0)

jerben:
Conformist allows you to define schemas to decode, validate and sanitize input data declaratively

- Project page: <a href="https://github.com/oxidizing/conformist">https://github.com/oxidizing/conformist</a>
- Documentation: <a href="https://oxidizing.github.io/conformist/">https://oxidizing.github.io/conformist/</a>

### Added
- `Conformist.Field.type_` can be used to retrieve a string representation of the type
- Support for custom `encoders` for custom types
- Support for encoding default values if present, this can be used to print schemas

### Fixed
- Move mandatory field name parameter to the end of create functions in order to get rid of the trailing () argument

### Fixed
- Move `fold_left` to top level module and accept `Conformist.t` as input

### Added
- Initial release supporting `int`, `float`, `string`, `bool`, `Ptime.date` and custom types

### DIFF
--- a/packages/conformist/conformist.0.1.0/opam
+++ b/packages/conformist/conformist.0.1.0/opam
@@ -1,0 +1,44 @@
+opam-version: "2.0"
+synopsis:
+  "Conformist allows you to define schemas to decode, validate and sanitize input data declaratively"
+description: """
+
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
I just added some line-breaks in the description, this makes it a bit easier to read and to review on github. This should not have any impact on the output of opam.ocaml.org nor of `opam show conformist`

Could you return this fix upstream so that it gets picked up in future releases?

kit-ty-kate:
Thanks!

PR #17323, opened 2020-09-30
Fix letters dependency to mrmime

mikonieminen:
Fix mrmime dependency to 0.3.0 in letters.


mseri:
It is not compatible with 0.3.1?

mseri:
Sorry two PRs with same fix, I merged the new one instead fi this by mistake. It should be fixed now

mikonieminen:
Yep, Josef didn't realize that I had created this one. Both did the same thing in a bit different way.

PR #17341, opened 2020-10-02
Fix to compile ocaml 4.07.1 on Xcode 12

garrigue:
Following a discussion at [Discuss](https://discuss.ocaml.org/t/ocaml-4-07-1-fails-to-build-with-apple-xcode-12/6441/20), here is a PR that allows to configure and compile ocaml-4.07.1 on recent MacOS.
Note that `-Wno-implicit-function-declaration` is only used during configuration.

### DIFF
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.4.07.1/files/configure-xcode12.patch
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.4.07.1/files/configure-xcode12.patch
@@ -1,0 +1,25 @@
+diff --git a/configure b/configure
+index 1316b3c1e..052a20876 100755
+--- a/configure
++++ b/configure
+@@ -454,8 +454,11 @@ case "$ocamlversion" in
+        *) gcc_warnings="-Wall";;
+ esac
+ 
++origcc="$cc"
++
+ case "$ccfamily" in
+   clang-*)
++    cc="$cc -Wno-implicit-function-declaration";
+     common_cflags="-O2 -fno-strict-aliasing -fwrapv";
+     internal_cflags="$gcc_warnings";;
+   gcc-[012]-*)
+@@ -2054,7 +2057,7 @@ fi
+ 
+ cclibs="$cclibs $mathlib"
+ 
+-config CC "$cc"
++config CC "$origcc"
+ config CPP "$cpp"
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
Does this only happens on OCaml 4.07.1? Looking at the discuss post people seem to say it won't happen with OCaml >= 4.08. Should we patch every compiler < 4.08?

An-Tu:
I am getting the same error when trying to install 4.06.1
MacOS. Error - `Fatal error: exception Invalid_argument("Sys.getcwd not implemented")`

garrigue:
> Does this only happens on OCaml 4.07.1? Looking at the discuss post people seem to say it won't happen with OCaml >= 4.08. Should we patch every compiler < 4.08?

I suppose yes. Hopefully, the exact same patch should work. But this is a lot of opam files to update.

dra27:
I would add this command before the `configure` command to the build recipe of _all_ compilers for 4.07 and earlier:

```
  ["sed" "-ib" "-e" "s/opts=\"\"/opts=\"-Wno-implicit-function-declaration\"/" "config/auto-aux/hasgot"] {os = "macos"}
```

(The `-ib` in the `sed` instructs `sed` to create a backup file, which is mandatory with the BSD-style `sed` installed on macOS and, mercifully, is compatible with GNU sed.)

This fixes the configure step all the way back to 3.07 (which is the oldest compiler in opam-repository) although for other reasons 3.11 is the oldest compiler you can then actually build. That change has the benefit of not requiring patch files, so is a bit less of a maintenance problem! Certainly on 4.07 (and possibly others), @garrigue's patch also "leaked" the `-Wno-implicit-function-declaration` into `MKDLL` and `MKMAINDLL` (although that almost certainly doesn't matter).

mattdamon108:
@dra27 Is there any way to use this command during `opam switch create` process?

dra27:
@mattdamon108 - no, not at present. The repository needs patching. See https://github.com/dra27/opam-repository/commit/67b0914532927b2a9f442992feb3b88bb12066cc

mseri:
@dra27 can you send a PR with the change?

kit-ty-kate:
Fixed in https://github.com/ocaml/opam-repository/pull/17379. Thanks everyone!

PR #17342, opened 2020-10-02
added packages/cpm/cpm.11.0.0/opam

UnixJunkie:
Some things are faster.
Some existing functions can be called on arrays, instead of only supporting lists previously.
But lists are still supported, for compatibility.


mseri:
Thanks!

PR #17344, opened 2020-10-02
[new release] ocamlgraph_gtk and ocamlgraph (2.0.0)

backtracking:
CHANGES:

- port to dune and opam 2.0
  - :exclamation: opam package now split into two packages: ocamlgraph and ocamlgraph_gtk
  - [WeakTopological] fixed incorrect use of generic hash tables
    (backtracking/ocamlgraph#99, Tomáš Dacík)
  - [Oper] fixed transitive_reduction (backtracking/ocamlgraph#91)
  - fix incorrect uses of polymorphic equality (Steffen Smolka, Boris Yakobowski)
  - [Coloring] fixed generation of OCamlDoc documentation
    (contributed by Earnestly)
  - :exclamation: [Coloring] functions now fail if the graph is directed
  - :exclamation: [Coloring] now uses a single, global exception [NoColoring]
  - [Coloring] new function two_color to 2-color a graph (or fail)
  - :exclamation: [Fixpoint] Take initial labeling of nodes into account (Johannes Kloos)

PR #17346, opened 2020-10-02
conf-gnomecanvas: Add support for FreeBSD

kit-ty-kate:
Per comment in https://github.com/ocaml/opam-repository/pull/17344


mseri:
Thanks!

PR #17349, opened 2020-10-04
[new release] cviode (0.0.4)

mseri:
Contact variational integrators - native ocaml version

- Project page: <a href="https://github.com/mseri/ocaml-cviode">https://github.com/mseri/ocaml-cviode</a>
- Documentation: <a href="https://mseri.github.io/ocaml-cviode/">https://mseri.github.io/ocaml-cviode/</a>

##### CHANGES:

* Update for compatibility with ode 0.3.0 and owl 0.8.0

PR #17352, opened 2020-10-04
Add openblas depext for freebsd

mseri:
Spotted on https://github.com/ocaml/opam-repository/pull/17349

### DIFF
--- a/packages/conf-openblas/conf-openblas.0.2.0/opam
+++ b/packages/conf-openblas/conf-openblas.0.2.0/opam
@@ -25,6 +25,7 @@ depexts: [
   ["openblas-devel"] {os-family = "suse"}
   ["openblas" "lapacke" "cblas"] {os-distribution = "arch"}
   ["openblas"] {os = "macos" & os-distribution = "homebrew"}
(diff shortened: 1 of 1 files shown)


mseri:
I don't have access to freeBSD to check if I need to specify a specific include\library path

mseri:
Thanks! Let's see how it goes. Unless they have changed policy, Fedora, RH and Centos had removed the pkg-config configuration for openblas a few releases back.

kit-ty-kate:
> Thanks! Let's see how it goes. Unless they have changed policy, Fedora, RH and Centos had removed the pkg-config configuration for openblas a few releases back.

oh :/ arf, I see. Looking at the package content on fedora it indeed doesn't have any pkg-config file :( I'll put back the file with some tweaks

mseri:
Yeah, the portability of blas/lapack is a nightmare... If one day I will have the time, I plan to modify owl to bundle its own version

mseri:
It's green!!!!

kit-ty-kate:
Thanks!

PR #17281, opened 2020-09-25
[new release] metapp, metaquot, traverse, refl, pattern 0.3.0

thierry-martinez:
These new releases support the new ppxlib.
This PR replaces https://github.com/ocaml/opam-repository/pull/17256 by removing clangml that will be published in a subsequent PR.

PR #17353, opened 2020-10-05
[new release] clangml.4.3.0

thierry-martinez:
This PR publishes the new release clangml.4.3.0:

- Port to ppxlib 0.16 / ocaml-migrate-parsetree 2.0.0

- Support for Clang/LLVM 11.0.0

- New module `Clang.Cursor`, exposing the signature
  `Hashtbl.HashedType with type t = cxcursor`
  and `Clang.Cursor.Hashtbl`.

- New function `Clang.Decl.annotate_access_specifier`, to enumerate
  class fields with their access specifiers.

- Fix segmentation fault when parsing the redeclaration of a function decl
  without prototype when the previous declarations has some parameters.
  (reported by Kihong Heo)

- New field `has_written_prototype` in `function_decl` to distinguish
  between function declarations where the prototype is actually provided
  and declarations where the prototype is inherited.

### DIFF
--- a/packages/clangml/clangml.4.3.0/opam
+++ b/packages/clangml/clangml.4.3.0/opam
@@ -1,0 +1,36 @@
+opam-version: "2.0"
+synopsis: "OCaml bindings for Clang API"
+description: """
+clangml provides bindings to call the Clang API from OCaml.
+"""
+maintainer: ["Thierry Martinez <thierry.martinez@inria.fr>"]
+authors: ["Thierry Martinez <thierry.martinez@inria.fr>"]
+license: "BSD-2-Clause"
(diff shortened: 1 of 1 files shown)


thierry-martinez:
We observe the same `Stack overflow` errors as in https://github.com/ocaml/opam-repository/pull/17281 due to incompatibility with spacetime. I will try to understand better this problem that I suspect is linked with `refl` generated code, but I would prefer to do that for a future version.

mseri:
I would not worry too much about that: spacetime is being removed from the compiler and it is picked incorrectly by the CI in place of the vanilla compiler.

thierry-martinez:
Thank you very much, @mseri !

PR #17355, opened 2020-10-05
Declare package cache in git-controlled repo file

AltGr:
The repository served over HTTP has it enabled anyway (enabled by
opam-admin), but we want clients to use the cache even when they use the
git repository directly.

NOTE: the repository scripts will need to be updated ASAP after this
is merged, to avoid a double-definition of the repository cache.

### DIFF
--- a/repo
+++ b/repo
@@ -1,3 +1,4 @@
 opam-version: "2.0"
-browse: "https://opam.ocaml.org/pkg/"
+browse: "https://opam.ocaml.org/packages/"
 upstream: "https://github.com/ocaml/opam-repository/tree/master/"
+archive-mirrors: "https://opam.ocaml.org/cache"


hannesm:
imho it would be nice to have this merged (and opam.ocaml.org adapted to not magically include that url)

mseri:
I think there needs to be coordination. What does
```
NOTE: the repository scripts will need to be updated ASAP after this
is merged, to avoid a double-definition of the repository cache.
```
imply? What do we or somebody else need to do?

AltGr:
It just means that until they do, there will be a double definition of the cache. I'll do the change, just need to remember about it.

The only consequence is that, in case the initial download from cache fails, it will be attempted a second time before falling back to the upstream URL. So not really a terrible problem :)

mseri:
@AltGr thanks. Then I am going to merge. Is that ok?

mseri:
@AltGr has the companion change happened?

Lupus:
I noticed a regression when `archive-mirrors` is configured in `$OPAMROOT/.config` to point to other mirror. Our CI pipelines stopped using local cache, and started using the cache at `opam.ocaml.org/cache` instead. Opam version in use is 2.0.7.
After I deleted `archive-mirrors` from `repo` file of our mirror of the official repo, local caches started getting used again.

mseri:
@hannesm @AltGr ⬆️ 
@Lupus thanks for the report

AltGr:
@Lupus hm, indeed opam processes the caches in the other way around: it will first try the "more specific", i.e. the one defined for just one repo, before trying the global one.
That's ignoring that users have control over the global one, but it's much more difficult to customise the repository one, so it should probably done the other way around.

PR #17375, opened 2020-10-08
Package flac.0.2.0

toots:
### `flac.0.2.0`
Interface for the Free Lossless Audio Codec otherwise known as FLAC

---
* Homepage: https://github.com/savonet/ocaml-flac
* Source repo: git+https://github.com/savonet/ocaml-flac.git
* Bug tracker: https://github.com/savonet/ocaml-flac/issues

---

### DIFF
--- a/packages/flac/flac.0.2.0/opam
+++ b/packages/flac/flac.0.2.0/opam
@@ -1,0 +1,49 @@
+opam-version: "2.0"
+maintainer: "Romain Beauxis <toots@rastageeks.org>"
+authors: "The Savonet Team <savonet-users@lists.sourceforge.net>"
+homepage: "https://github.com/savonet/ocaml-flac"
+build: [
+  ["./bootstrap"] {dev}
+  ["./configure" "--prefix" prefix] {os != "macos"}
+  [
+    "./configure"
+    "CFLAGS=-I/usr/local/include"
+    "LDFLAGS=-L/usr/local/lib"
+    "OCAMLFLAGS=-ccopt -I/usr/local/include -cclib -L/usr/local/lib"
+    "--prefix"
+    prefix
+  ] {os = "macos"}
+  [make "clean"] {dev}
+  [make]
+]
+install: [
+  [make "install"]
+]
+depends: [
+  "ocaml" {>= "4.03"}
+  "ocamlfind" {build}
+  "ogg"
+]
+depexts: [
+  ["flac-dev"] {os-distribution = "alpine"}
+  ["flac"] {os-distribution = "arch"}
+  ["flac-devel"] {os-distribution = "centos"}
+  ["flac-devel"] {os-distribution = "fedora"}
+  ["flac-devel"] {os-family = "suse"}
+  ["libflac-dev"] {os-family = "debian"}
+  ["flac"] {os-distribution = "nixos"}
+  ["flac"] {os = "macos" & os-distribution = "homebrew"}
+  ["flac"] {os = "freebsd"}
+]
+bug-reports: "https://github.com/savonet/ocaml-flac/issues"
+dev-repo: "git+https://github.com/savonet/ocaml-flac.git"
+synopsis:
+  "Interface for the Free Lossless Audio Codec otherwise known as FLAC"
+url {
+  src:
+    "https://github.com/savonet/ocaml-flac/releases/download/v0.2.0/ocaml-flac-0.2.0.tar.gz"
+  checksum: [
+    "md5=72acb898a4e313b3e73e5c04b0664d8e"
+    "sha512=27b3e8e9a08b241b89e7d06ce816916518b213c81a79cf2e087a0236eac0df9c05e53254959c2a2ad7860cb6972a0e274e2cb18a19f86b7f2fe26bf4e9e789c5"
+  ]
+}


kit-ty-kate:
This package was failing with:
```
# checking whether pkg-config knows about flac... /home/opam/.opam/4.11/.opam-switch/build/flac.0.2.0/./configure: line 5674: --exists: command not found
# configure: error: flac.pc not found.. Do you need to set PKG_CONFIG_PATH?
```
Telling that `pkg-config` was missing.
I added the missing constraint on top of creating a separate `conf-libflac` package to centralize any change to the depexts (this makes it was more maintainable).

Would it be ok to return the change upstream?

packages/flac/flac.0.1.1/opam
```diff
@@ -4,18 +4,15 @@ build: [
...
-  ["flac-devel"] {os-family = "suse"}
-  ["libflac-dev"] {os-family = "debian"}
-  ["flac"] {os-distribution = "nixos"}
-  ["flac"] {os = "macos" & os-distribution = "homebrew"}
+depends: [
+  "ocaml"
+  "ocamlfind"
+  "ogg"
+  "base-unsafe-string"
+  "conf-pkg-config" {build}
```
toots, line 12: Should `conf-pkg-config` be removed since it's already required by `conf-libflac`?
```suggestion
```

packages/flac/flac.0.1.2/opam
```diff
@@ -17,24 +17,17 @@ build: [
...
   "ocamlfind"
   "ogg"
-]
-depexts: [
-  ["flac-devel"] {os-distribution = "centos"}
-  ["flac-devel"] {os-distribution = "fedora"}
-  ["flac-devel"] {os-family = "suse"}
-  ["libflac-dev"] {os-family = "debian"}
-  ["flac"] {os = "macos" & os-distribution = "homebrew"}
+  "conf-pkg-config" {build}
```
toots, line 24: Same here.
```suggestion
```

packages/flac/flac.0.1.3/opam
```diff
@@ -17,27 +17,17 @@ build: [
...
-depexts: [
-  ["flac-dev"] {os-distribution = "alpine"}
-  ["flac"] {os-distribution = "arch"}
-  ["flac-devel"] {os-distribution = "centos"}
-  ["flac-devel"] {os-distribution = "fedora"}
-  ["flac-devel"] {os-family = "suse"}
-  ["libflac-dev"] {os-family = "debian"}
-  ["flac"] {os-distribution = "nixos"}
-  ["flac"] {os = "macos" & os-distribution = "homebrew"}
+  "conf-pkg-config" {build}
```
toots, line 24: Same here.
```suggestion
```

packages/flac/flac.0.1.4/opam
```diff
@@ -17,27 +17,17 @@ build: [
...
-depexts: [
-  ["flac-dev"] {os-distribution = "alpine"}
-  ["flac"] {os-distribution = "arch"}
-  ["flac-devel"] {os-distribution = "centos"}
-  ["flac-devel"] {os-distribution = "fedora"}
-  ["flac-devel"] {os-family = "suse"}
-  ["libflac-dev"] {os-family = "debian"}
-  ["flac"] {os-distribution = "nixos"}
-  ["flac"] {os = "macos" & os-distribution = "homebrew"}
+  "conf-pkg-config" {build}
```
toots, line 24: Same here.
```suggestion
```

packages/flac/flac.0.1.6/opam
```diff
@@ -21,16 +21,8 @@ depends: [
...
-depexts: [
-  ["flac-dev"] {os-distribution = "alpine"}
-  ["flac"] {os-distribution = "archlinux"}
-  ["flac-devel"] {os-distribution = "centos"}
-  ["flac-devel"] {os-distribution = "fedora"}
-  ["flac-devel"] {os-family = "suse"}
-  ["libflac-dev"] {os-family = "debian"}
-  ["flac"] {os-distribution = "nixos"}
-  ["flac"] {os = "macos" & os-distribution = "homebrew"}
+  "conf-pkg-config" {build}
```
toots, line 24: Same here.
```suggestion
```

packages/flac/flac.0.1.7/opam
```diff
@@ -23,17 +23,8 @@ depends: [
...
-  ["flac-dev"] {os-distribution = "alpine"}
-  ["flac"] {os-distribution = "archlinux"}
-  ["flac-devel"] {os-distribution = "centos"}
-  ["flac-devel"] {os-distribution = "fedora"}
-  ["flac-devel"] {os-family = "suse"}
-  ["libflac-dev"] {os-family = "debian"}
-  ["flac"] {os-distribution = "nixos"}
-  ["flac"] {os = "macos" & os-distribution = "homebrew"}
-  ["flac"] {os = "freebsd"}
+  "conf-pkg-config" {build}
```
toots, line 26: Same here.
```suggestion
```

packages/flac/flac.0.2.0/opam
```diff
@@ -0,0 +1,40 @@
...
+  [make]
+]
+install: [
+  [make "install"]
+]
+depends: [
+  "ocaml" {>= "4.03"}
+  "ocamlfind" {build}
+  "ogg"
+  "conf-pkg-config" {build}
```
toots, line 26: Same here.
```suggestion
  "conf-pkg-config" {build}
```

toots (commented):
Thanks for splitting the configuration package, this is a great improvement. I'm definitely down to merge it upstream & will try to follow that pattern in the future. Made a couple of comments..

packages/flac/flac.0.1.1/opam
kit-ty-kate: It's always better to be explicit about things the package use explicitly, `conf-libflac` might use a different method of detection in the future, we never know.

Unless the old packages did not use `pkg-config` I would suggest to keep `conf-pkg-config`. Is that the case?

toots: That makes total sense. The packages does indeed use `pkg-config` so it makes sense to keep it there.

toots:
Backport done: https://github.com/savonet/ocaml-flac/commit/568e788ed9d83c92283242544df5b1149e16b750 🙂

kit-ty-kate:
Thanks a lot!

PR #17379, opened 2020-10-08
Fix OCaml < 4.08.0 for XCode 12

dra27:
As requested in https://github.com/ocaml/opam-repository/pull/17341#issuecomment-705623176 🙂

This PR patches `config/auto-aux/hasgot` in all pre-autoconf versions of OCaml to pass `-Wno-implicit-function-declaration` when building on macOS. Apparently prior to GCC 5, `-Wno-implicit-function-declaration` didn't exist - I don't know what the story would have been for clang back then, but I imagine it's quite difficult to be sat at a Mac running that old a version of GCC.

### DIFF
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.3.07+1/opam
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.3.07+1/opam
@@ -11,6 +11,13 @@ conflict-class: "ocaml-core-compiler"
 flags: compiler
 setenv: CAML_LD_LIBRARY_PATH = "%{lib}%/stublibs"
 build: [
+  [
+    "sed"
+    "-ib"
+    "-e"
+    "s/opts=\"\"/opts=\"-Wno-implicit-function-declaration\"/"
+    "config/auto-aux/hasgot"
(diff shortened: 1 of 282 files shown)


kit-ty-kate:
Thanks a lot!

mattdamon108:
Is it supposed to be built successfully after this PR merged? I'm afraid to say that I have still a build failure with the same error message before.

```
# Fatal error: exception Invalid_argument("Sys.getcwd not implemented")
```

dra27:
I haven't tested, so I haven't messed it up, but did you do `opam update` and did you use the http repo or the git repo? I _hope_ it's just that you've tried before the http repo was re-sync'd. Try `opam switch create testing-4.06.1 ocaml-base-compiler.4.06.1 --repos=git=git+https://github.com/ocaml/opam-repository.git`

dra27:
This is definitely working for me with both the Git and HTTP remotes (I double-checked before updating the HTTP that the old one was indeed not working)

mattdamon108:
My bad. `opam update` does fix. It works perfectly. Thank you!

PR #17391, opened 2020-10-11
[new release] ca-certs (0.1.1)

hannesm:
Detect root CA certificates from the operating system

- Project page: <a href="https://github.com/mirage/ca-certs">https://github.com/mirage/ca-certs</a>
- Documentation: <a href="https://mirage.github.io/ca-certs/doc">https://mirage.github.io/ca-certs/doc</a>

##### CHANGES:

* Revise test suite to not connect to the network (to please opam's sandbox),
  instead use hardcoded certificate chains.

### DIFF
--- a/packages/ca-certs/ca-certs.0.1.1/opam
+++ b/packages/ca-certs/ca-certs.0.1.1/opam
@@ -1,0 +1,52 @@
+opam-version: "2.0"
+synopsis: "Detect root CA certificates from the operating system"
+description: """
+TLS requires a set of root anchors (Certificate Authorities) to
+authenticate servers. This library exposes this list so that it can be
+registered with ocaml-tls.
+"""
+maintainer: ["Etienne Millon <me@emillon.org>"]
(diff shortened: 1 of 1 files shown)


hannesm:
hmm, anyone with a macos 10.14 around who's an idea where their root certificates are kept? on a GitHub action CI with macos 10.15.7 it works fine, travis uses 10.14.6 -- and there it looks like "/System/Library/Keychains/SystemRootCertificates.keychain" is not the right path...

in contrast to some stack overflow threads, running only `security find-certificate -a -p` does _not_ work on macos 10.15 -- it does not return the trust anchors, but only some system internal certificates...

kit-ty-kate:
@hannesm I just tested on my test server running Catalina (10.15) and I got the same error when ran inside opam. However when calling dune directly it works. Maybe something to do with sandbox-exec?

hannesm:
let me prepare a 0.1.2 -- whch unlikely fixes the issue, but avoids the test run on macos

PR #17397, opened 2020-10-12
ocaml-freestanding 0.6.3

mato:
* nolibc: dlmalloc: Improve security posture and robustness by always
enabling assertions, initialising magic value for heap canaries using
monotonic time and enabling FOOTERs. (#87, @mato)


kit-ty-kate:
Thanks!

PR #17410, opened 2020-10-14
[new release] repr-bench, repr, ppx_repr and repr-fuzz (0.1.0)

CraigFe:
Dynamic type representations. Provides no stability guarantee

- Project page: <a href="https://github.com/mirage/repr">https://github.com/mirage/repr</a>

##### CHANGES:

Initial release.

### DIFF
--- a/packages/repr/repr.0.1.0/opam
+++ b/packages/repr/repr.0.1.0/opam
@@ -23,6 +23,7 @@ depends: [
   "hex" {with-test}
   "alcotest" {>= "1.1.0" & with-test}
   "odoc" {with-doc}
+  "ppx_repr" {= version & post & with-test}
 ]
 build: [
   ["dune" "subst"] {dev}
--- a/packages/repr/repr.0.1.0/opam
+++ b/packages/repr/repr.0.1.0/opam
@@ -23,7 +23,7 @@ depends: [
   "hex" {with-test}
   "alcotest" {>= "1.1.0" & with-test}
   "odoc" {with-doc}
-  "ppx_repr" {= version & post & with-test}
+  "ppx_repr" {= version & with-test}
 ]
 build: [
   ["dune" "subst"] {dev}
--- a/packages/repr/repr.0.1.0/opam
+++ b/packages/repr/repr.0.1.0/opam
@@ -1,0 +1,50 @@
+opam-version: "2.0"
+synopsis: "Dynamic type representations. Provides no stability guarantee"
+description: """
+This package defines a library of combinators for building dynamic type
+representations and a set of generic operations over representable types, used
+in the implementation of Irmin and related packages.
+
+It is not yet intended for public consumption and provides no stability
+guarantee.
+"""
+maintainer: ["thomas@gazagnaire.org"]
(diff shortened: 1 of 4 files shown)


CraigFe:
The CI is failing because of a missing internal dependency. I'll fix that upstream and then report back.

packages/repr/repr.0.1.0/opam
```diff
@@ -23,6 +23,7 @@ depends: [
   "hex" {with-test}
   "alcotest" {>= "1.1.0" & with-test}
   "odoc" {with-doc}
+  "ppx_repr" {= version & post & with-test}
```
mseri, line 26: Looks like that by using post, it is installed after running the tests...
```suggestion
  "ppx_repr" {= version & with-test & post}
```
Do you think ordering may be an issue here? (I have no idea how post work)

Otherwise we could just disable the tests (or you could move these tests to `ppx_repr` instead)

kit-ty-kate: ah nah nevermind I hadn't realized this was a dependency from the same package. @CraigFe is there any way you could not use this dependency recursively? If the base library is not usuable without the ppx then the ppx must be merged in the main library.

CraigFe: Hmm, `post` was not intended as a CI hack, and I think it should be possible to use this dependency set, since we were doing the same thing with `irmin` and `ppx_irmin` shortly before these packages were extracted to their own repository.

I'll take a look at it tomorrow.

mseri:
Without `post` it does work. Should we merge it?

CraigFe:
I'm a bit suspicious of _why_ it works, given the cycle dependency. See https://travis-ci.org/github/mirage/repr/jobs/736333035, where making this change upstream causes `ocaml-ci-scripts` to fail; it'd be nice to find a solution that doesn't require us to maintain a different set of constraints during development.

mseri:
Why merging `repr` and `ppx_repr` is not an option?

CraigFe:
Merging them would require considering certain test dependencies of `repr` as real dependencies via `ppx_repr` (in particular, `ppxlib`), so it has a non-trivial maintenance cost.

Re. moving `repr` tests requiring `ppx_repr` into `ppx_repr`, we could do that too, but that is currently _all_ of the tests. (Those tests are not actually tests of the PPX, but just tests that use the PPX for convenience.) It seems best to keep code where it belongs semantically unless there is no other option (and I would like to get to the bottom of `post`, since this dependency situation seems likely to arise again with other PPX workflows.)

mseri:
> Re. moving `repr` tests requiring `ppx_repr` into `ppx_repr`, we could do that too, but that is currently _all_ of the tests.

They are in the same repo, it does not seem such a terrible problem. Also semantically they are sort of testing both libraries anyways.

> I would like to get to the bottom of `post`, since this dependency situation seems likely to arise again with other PPX workflows

Absolutely!

CraigFe:
Having discussed this a bit with @emillon offline, it seems that `post` is indeed not the correct solution here. I'm closing this PR for now while we work out a less cyclic test configuration upstream.

Thanks for the help!

PR #17414, opened 2020-10-15
frama-c: do not force disable GUI (fix lablgtk3 issue)

maroneze:
The Frama-C configure should be able to automatically disable the GUI when the required packages are not available, so we don't need to test them again at the opam level. Also, this avoids an issue when installing lablgtk3-specific packages (namely conf-gtksourceview3), which would force disabling the GUI unnecessarily.

### DIFF
--- a/packages/frama-c/frama-c.21.1/opam
+++ b/packages/frama-c/frama-c.21.1/opam
@@ -86,9 +86,6 @@ tags: [
 build: [
   ["autoconf"] {pinned}
   ["./configure" "--prefix" prefix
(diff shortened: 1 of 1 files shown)


mseri:
Can you please move this change from `frama-c.21.1` to a patch version `frama-c.21.1-1`?

maroneze:
After discussing with the Frama-C team, we believe that the bug is not that severe, a new major release will be available soon, and adding a patch version to opam without referencing it from our side would be unwise. We'll also try fixing the issues with the tests (related to Python dependencies which I don't know how to express precisely).

Thank you, I'll be closing this for now.

PR #17417, opened 2020-10-15
[new release] conduit-lwt-tls, conduit-mirage, conduit-tls, conduit-lwt, conduit-async-tls, conduit-async, conduit, conduit-async-ssl and conduit-lwt-ssl (3.0.0)

dinosaure:
A portable network connection establishment library using Lwt

- Project page: <a href="https://github.com/mirage/ocaml-conduit">https://github.com/mirage/ocaml-conduit</a>

##### CHANGES:

* **breaking change** - New version of `conduit`. Most of the new API has a
  description on the pull-request
  [mirage/ocaml-conduit#311](https://github.com/mirage/ocaml-conduit/pull/311). Documentation is
  updated as well with a small HOW-TO (available
  [here](https://mirage.github.io/ocaml-conduit/conduit/howto.html)) which
  describes the new API. An other document exists to understand the goal
  of Conduit [here](https://mirage.github.io/ocaml-conduit/conduit/readme.html).

### DIFF
--- a/packages/conduit-async-ssl/conduit-async-ssl.3.0.0/opam
+++ b/packages/conduit-async-ssl/conduit-async-ssl.3.0.0/opam
@@ -1,0 +1,38 @@
+opam-version: "2.0"
+maintainer: "anil@recoil.org"
+authors: [
+  "Anil Madhavapeddy"
+  "Thomas Leonard"
+  "Thomas Gazagnaire"
+  "Rudi Grinberg"
+]
+license: "ISC"
+tags: "org:mirage"
+homepage: "https://github.com/mirage/ocaml-conduit"
+bug-reports: "https://github.com/mirage/ocaml-conduit/issues"
+dev-repo: "git+https://github.com/mirage/ocaml-conduit.git"
+synopsis: "A network connection establishment library using Async and OpenSSL"
+
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j1"] {with-test}
+]
+
+depends: [
+  "ocaml"         {>= "4.07.0"}
+  "dune"          {>= "2.0.0"}
+  "core"
+  "conduit-async" {= version}
+  "async"         {>= "v0.12.0"}
+  "async_ssl"
+]
+x-commit-hash: "f5ec040e47bf878c17c247a8e18bed7a0d60d2af"
+url {
+  src:
+    "https://github.com/mirage/ocaml-conduit/releases/download/v3.0.0/conduit-v3.0.0.tbz"
+  checksum: [
+    "sha256=ac41681b50107c96bcae92fc93be4faa3d4d2b6c2095e79b963c142a0f055fd8"
+    "sha512=b427c2f37908b662f98360410dcaa18b31554024ddf2480abd9195acc83f427df44428642d2b028faa15c99f4a538d15fcd89d9b8dca399ee684864633963738"
+  ]
+}
--- a/packages/conduit-async-tls/conduit-async-tls.3.0.0/opam
+++ b/packages/conduit-async-tls/conduit-async-tls.3.0.0/opam
@@ -1,0 +1,38 @@
+opam-version: "2.0"
+maintainer: "anil@recoil.org"
+authors: [
+  "Anil Madhavapeddy"
+  "Thomas Leonard"
+  "Thomas Gazagnaire"
+  "Rudi Grinberg"
+]
+license: "ISC"
+tags: "org:mirage"
+homepage: "https://github.com/mirage/ocaml-conduit"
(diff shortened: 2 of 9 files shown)


dinosaure:
I will be available tomorrow to fix revdeps on this PR - I know that it will break many things but most important upper packages are updated as well on this new version.

kit-ty-kate:
Could you also restrict every packages that is using conduit? The new interface is incompatible with all of them. Only `github-hooks.0.5.0` and the conduit packages are compatible.

This should be easily done using: `opam admin add-constraint '< 3.0.0'`

dinosaure:
I did what you told me to do and fix some minor issues :+1:. I think it should be ready to test it now!

mseri:
@kit-ty-kate do you think the CI will be able to cope with such an amount of mainstream packages before the nightly reset?

If it is too stressed by the size of the test, I think we could merge https://github.com/ocaml/opam-repository/pull/17417/commits/9351effdc644661e3cc7ac571be1297077de7408 as a separate PR and rebase this on top of it

mseri:
All the errors are unrelated to this pr, mostly about solver timeouts. There is a github-unix one that seems to be due to a missing js_of_ocaml upper or lower bound

kit-ty-kate:
@mseri your first intuition was right. I created #17428 and it should be merged first. Then we'll be able to re-test this one (sorry for that)

hannesm:
looks like a missing dependency (in travis runs) for conduit-async {with-test}:
```bash
=== ERROR while compiling conduit-async.3.0.0 ================================#
# context     2.0.6 | freebsd/x86_64 | ocaml-base-compiler.4.09.1 | file:///usr/home/travis/build/ocaml/opam-repository
# path        /usr/home/travis/.opam/ocaml-base-compiler.4.09.1/.opam-switch/build/conduit-async.3.0.0
# command     /usr/home/travis/.opam/ocaml-base-compiler.4.09.1/bin/dune runtest -p conduit-async -j1
# exit-code   1
# env-file    /usr/home/travis/.opam/log/conduit-async-89772-7ee6e6.env
# output-file /usr/home/travis/.opam/log/conduit-async-89772-7ee6e6.out
### output ###
# File "tests/ping-pong/dune", line 35, characters 21-38:
# 35 |    conduit-async-tls conduit-async-ssl))
#                           ^^^^^^^^^^^^^^^^^
# Error: Library "conduit-async-ssl" not found.
# Hint: try:
#   dune external-lib-deps --missing -p conduit-async -j 1 @runtest
```

kit-ty-kate:
mh, that sounds like a cyclic dependency. I'm guessing what's happening is that this test case isn't linked to any package and thus executed for any packages. In this case, on top of it, this test case uses both `conduit-async-tls` and `conduit-async-ssl`, which doesn't really makes sense. Could you re-specify those tests? Looking at them it looks like they'd be defined a lot more concisely using the `(test ...)` stanza as well.

dinosaure:
> mh, that sounds like a cyclic dependency. I'm guessing what's happening is that this test case isn't linked to any package and thus executed for any packages. In this case, on top of it, this test case uses both conduit-async-tls and conduit-async-ssl, which doesn't really makes sense. Could you re-specify those tests? Looking at them it looks like they'd be defined a lot more concisely using the (test ...) stanza as well.

These tests should be a part of `conduit-async` and it's why I added the `post` tag on `conduit-async-ssl` and `conduit-async-tls`. I'm not sure that this is the way to go however...

mseri:
> These tests should be a part of `conduit-async` and it's why I added the `post` tag on `conduit-async-ssl` and `conduit-async-tls`. I'm not sure that this is the way to go however...

There was recently a similar issue for `repr/ppx_repr` (see https://github.com/ocaml/opam-repository/pull/17423 and the referenced PR). In that case I suggested moving the tests as the simplest solution (as I did in the past for `ocaml-rpc`). I don't know what is the best thing to do to be honest.

CraigFe:
@dinosaure: As @mseri says, we ran into this problem last week, during which I discovered that `{ post & with-test }` dependencies are not guaranteed to be installed before running the tests. As far as I'm aware, `opam` has no direct support for this workflow so it's necessary to do some package adjustment to make it happy. (See related `opam` issue https://github.com/ocaml/opam/issues/4267.)

dinosaure:
>  In that case I suggested moving the tests as the simplest solution

What you mean by moving tests? We should delete the `[ dune runtest ] {with-test}`?

mseri:
That is a possibility but, no, I mean run the tests that require `conduit-async-ssl` as tests for `conduit-async-ssl` (even if they are meant as tests for `conduit-async`) and similarly for `tls`.

dinosaure:
Ah ok, I see... I need to tweak a bit the distribution so (when the test for async requires `conduit-async-tls` and `conduit-async-ssl`). Thanks, I will come back with another push.

PR #17429, opened 2020-10-18
Fix tezos compatibility with uri 4.0

pirbo:
Uri 4.0 dropped (well separate into a new package) its dependency to re. 

That revealed that, in 2 places in tezos, re is used without being explicitly declared in the `dune` files which now led build failure. These patches fix that.

### DIFF
--- a/packages/tezos-base/tezos-base.7.4/opam
+++ b/packages/tezos-base/tezos-base.7.4/opam
@@ -14,7 +14,6 @@ depends: [
   "ptime" { >= "0.8.4" }
   "ezjsonm" { >= "0.5.0" }
   "ipaddr" { >= "4.0.0" & < "5.0.0" }
-  "re" { >= "1.7.2" }
 ]
 build: [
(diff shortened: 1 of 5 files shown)


avsm:
Interesting, thanks for the fix. This sort of transitive dependency issue is difficult to spot in testing, but we've got https://github.com/ocurrent/ocaml-ci/issues/257 open to see if we can simulate it better.  Dune also supports `implicit_transitive_deps` and there are a few open PRs to improve the usability of this https://github.com/ocaml/dune/issues/2733.

avsm:
Note that tezos does not show up in the reverse dep list for uri at all: https://ci.ocaml.org/ocaml/opam-repository/pr/17411?history=7282de3706ca1661102d7ef80353bfe9db5a3822&test=2%20Revdeps

mseri:
In general I would have asked to move the change to a patch release but the packages are all version-linked and that would require a full release. So in this case I am in favour of making an exception an merge the patch. @kit-ty-kate do you agree?

PR #17438, opened 2020-10-19
[new release] conduit-lwt-tls, conduit-mirage, conduit-tls, conduit-lwt, conduit-async-tls, conduit-async, conduit, conduit-async-ssl and conduit-lwt-ssl (3.0.0)

dinosaure:
A portable network connection establishment library using Lwt and ocaml-tls

- Project page: <a href="https://github.com/mirage/ocaml-conduit">https://github.com/mirage/ocaml-conduit</a>

##### CHANGES:

* **breaking change** - New version of `conduit`. Most of the new API has a
  description on the pull-request
  [mirage/ocaml-conduit#311](https://github.com/mirage/ocaml-conduit/pull/311). Documentation is
  updated as well with a small HOW-TO (available
  [here](https://mirage.github.io/ocaml-conduit/conduit/howto.html)) which
  describes the new API.

### DIFF
--- a/packages/conduit-async-ssl/conduit-async-ssl.3.0.0/opam
+++ b/packages/conduit-async-ssl/conduit-async-ssl.3.0.0/opam
@@ -1,0 +1,38 @@
+opam-version: "2.0"
+maintainer: "anil@recoil.org"
+authors: [
+  "Anil Madhavapeddy" 
+  "Thomas Leonard" 
+  "Thomas Gazagnaire" 
+  "Rudi Grinberg"
+]
+license: "ISC"
+tags: "org:mirage"
+homepage: "https://github.com/mirage/ocaml-conduit"
+bug-reports: "https://github.com/mirage/ocaml-conduit/issues"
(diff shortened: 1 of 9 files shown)


dinosaure:
Redo #17417.

mseri:
Now that there are no-longer revdeps, due to the upper bounds everywhere, the only thing that needs to be checked is that `conduit-*` builds also on ocaml `4.07` (which is not tested in the conduit CI). If that is the case (I have not tried, and cannot at the moment) should we merge and release the ocaml CI?

dinosaure:
> Now that there are no-longer revdeps, due to the upper bounds everywhere, the only thing that needs to be checked is that conduit-* builds also on ocaml 4.07 (which is not tested in the conduit CI). If that is the case (I have not tried, and cannot at the moment) should we merge and release the ocaml CI?

I'm not sure, as far as I can tell, all should works - build & tests work. So I will happy to definitely merge it (and unlock several PR). May be @kit-ty-kate has some clue about the state of the CI?

mseri:
It’s all green!

PR #17445, opened 2020-10-20
coqide cannot be installed anymore due to upgrade of conf-gnome-icon-…

MarcPantel:
…theme3

Message from spam during the installation: "The package 'conf-gnome-icon-theme3' is deprecated. Please use 'conf-adwaita-icon-theme' instead!"

### DIFF
--- a/packages/coqide/coqide.8.12.0-1/files/coqide.install
+++ b/packages/coqide/coqide.8.12.0-1/files/coqide.install
@@ -1,0 +1,9 @@
+bin: [
+  "bin/coqide"
+]
+share_root: [
+  "ide/coq.lang" {"coq/coq.lang"}
+  "ide/coq-ssreflect.lang" {"coq/coq-ssreflect.lang"}
+  "ide/coq.png" {"coq/coq.png"}
+  "ide/coq_style.xml" {"coq/coq_style.xml"}
+]
--- a/packages/coqide/coqide.8.12.0-1/opam
+++ b/packages/coqide/coqide.8.12.0-1/opam
@@ -17,7 +17,7 @@ depends: [
   "ocamlfind" {build}
   "conf-findutils" {build}
   "lablgtk3-sourceview3"
-  "conf-gnome-icon-theme3"
+  "conf-adwaita-icon-theme"
 ]
 build: [
   [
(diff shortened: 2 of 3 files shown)


mseri:
Thanks! Can you make this change to a new patch release `packages/coqide/coqide.8.12.0-1/opam`?

mseri:
Even though it was deprecated, did the package install or the installation failed?

MarcPantel:
The installation fails because conf-gnome-icon-theme3 refuse to be installed. It occurred when I upgraded my OPAM today. It is a recent upgrade of the conf-gnome-icon-theme3 package.

Zimmi48:
FTR the fix contained in this PR was exactly the same as the one in #17463 but limited to the latest version of CoqIDE. The move to a new package version instead of the originally proposed fix resulted in the conflict being missed. But as pointed out by @yannl35133, this new version serves no purpose whatsoever, so should simply be removed.

mseri:
Yes, that is correct. At the time I had misunderstood the nature of the change. Marking the package as unavailable would likely break the switch on upgrade for the users with this version installed. Since keeping it makes no harm, I would recommend against touching it again

Zimmi48:
>for the users with this version installed

With which version? `8.12.0-1` cannot be installed at all, unless users manually edit the opam configuration.

mseri:
Aha, I see. Good point, this was never installable indeed

MarcPantel:
Yes, I forgot to comment when I detected it and the next day, the other patch was committed so I though you had done the change.

MSoegtropIMC:
Also please note that the adwaita icon theme is the correct package and the gnome icon theme is the wrong one. It is not helpful to install the wrong package just because this happens to be installable on some systems. If the adwaita icon theme conf package does not work on some system, the conf package should be fixed for that system.

The only reason CoqIDE works with installing the wrong icon theme is that the correct (adwaita) icon theme is installed anyway on most systems.

mseri:
Sure, this should be fixed now: https://github.com/ocaml/opam-repository/pull/17520

MSoegtropIMC:
Sorry, I was confused by looking at:

(https://github.com/ocaml/opam-repository/commit/a037ee74ee61dc319e756b6212728e5c7017b78e)

which changes the icon theme of the original 8.12.0 package which is not fixed in #17520, but then this commit is not what has been merged. So in master it is indeed fine now.

PR #17458, opened 2020-10-21
[new release] vchan-unix, vchan and vchan-xen (6.0.0)

hannesm:
Xen Vchan implementation

- Project page: <a href="https://github.com/mirage/ocaml-vchan">https://github.com/mirage/ocaml-vchan</a>
- Documentation: <a href="https://mirage.github.io/ocaml-vchan">https://mirage.github.io/ocaml-vchan</a>

##### CHANGES:

* Adapt to mirage-xen 6.0.0 API changes (Solo5 based Xen PVH, mirage/ocaml-vchan#136 @mato)

PR #17461, opened 2020-10-22
xen-devel: use on os-family = suse

hannesm:
this should fix the failure observed in https://github.com/ocaml/opam-repository/pull/17458 (and https://github.com/mirage/ocaml-vchan/pull/136#issuecomment-713448337)


mseri:
There is a big upcoming PR that will move everything to use family, which is more generic

PR #17463, opened 2020-10-22
Changed GTK icon theme to adwaita in all GTK3 based CoqIDE packages

MSoegtropIMC:
This is a follow up on my PR #17279 which replaces the dependency from the now deprecated conf-gnome-icon-theme3 to conf-adwaita-icon-theme for all versions of CoqIDE based on GTK3.

Note: on Mac CoqIDE does no longer build since #17279 because the hack to install the adwaita icon theme in conf-gnome-icon-theme3 on Mac has been removed with #17279.

See also (https://github.com/coq/coq/issues/13229).

### DIFF
--- a/packages/coqide/coqide.8.10.0/opam
+++ b/packages/coqide/coqide.8.10.0/opam
@@ -11,7 +11,7 @@ depends: [
   "coq" {= version}
   "lablgtk3-sourceview3" {!= "3.0.beta7"}
   "conf-findutils" {build}
-  "conf-gnome-icon-theme3"
+  "conf-adwaita-icon-theme"
 ]
 build: [
   [
--- a/packages/coqide/coqide.8.10.1/opam
+++ b/packages/coqide/coqide.8.10.1/opam
@@ -16,7 +16,7 @@ depends: [
   "coq" {= version}
   "lablgtk3-sourceview3" {!= "3.0.beta7"}
   "conf-findutils" {build}
-  "conf-gnome-icon-theme3"
+  "conf-adwaita-icon-theme"
 ]
 build: [
   [
--- a/packages/coqide/coqide.8.10.2/opam
+++ b/packages/coqide/coqide.8.10.2/opam
@@ -16,7 +16,7 @@ depends: [
   "coq" {= version}
   "lablgtk3-sourceview3" {!= "3.0.beta7"}
   "conf-findutils" {build}
-  "conf-gnome-icon-theme3"
+  "conf-adwaita-icon-theme"
 ]
 build: [
   [
--- a/packages/coqide/coqide.8.11.0/opam
+++ b/packages/coqide/coqide.8.11.0/opam
@@ -16,7 +16,7 @@ depends: [
   "coq" {= version}
   "lablgtk3-sourceview3" {!= "3.0.beta7"}
   "conf-findutils" {build}
-  "conf-gnome-icon-theme3"
+  "conf-adwaita-icon-theme"
 ]
 build: [
   [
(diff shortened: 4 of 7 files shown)


mseri:
Please do not modify the packages in place, instead create patch releases by appending `-1` to the version

MSoegtropIMC:
> Please do not modify the packages in place, instead create patch releases by appending -1 to the version (see a037ee7).

I thought about this but decided against it for these reasons:

- none of the old packages still builds on Mac
- on all other platforms it will note that it has deprecated dependencies
- it actually installs the wrong thing - CoqIDE needs the adwaita icon theme. The only reason why this did work before is because the conf-gnome-icon-theme3 package had no check if the icon theme is actually installed and most people have the adwaita icon theme installed anyway
- the actual content of the package (the software it builds) is not changed - this is a pure adjustment to system dependencies and changes in other packages.

So I don't see a reason for keeping the old packages. There have been similar decision for other packages. As far as I can tell the reasoning so far was: "keep the old package if there might be a reason for someone to install the old package". This is clearly not the case here. So for consistency I changed the existing packages.

But this is just my interpretation of what the policy is based on precedence cases - not sure what the official policy is. It might also be that I mix things with PRs on the Coq opam repo and here. I am happy to change it if you say this is really the policy even if the old packages don't make sense any more.

mseri:
That is correct. I misunderstood that the gnome icon theme would be the package needed on legacy systems

MSoegtropIMC:
> I misunderstood that the gnome icon theme would be the package needed on legacy systems.

As far as I can tell all systems supporting GTK3 have the adwaita icon theme pre installed except Mac, where the conf-gnome-icon-theme3 package previously installed the adwaita icon theme.

So I would say that this PR changes the situation to what it was before #17279 on all systems. On Mac it does the same thing again and on all other systems the previous dependency was a nop since the icon packages are pre-installed and the conf package didn't check anything.

mseri:
But then let’s keep the nop there. It does no harm to move it to the patch version for extra safety

MSoegtropIMC:
> But then let’s keep the nop there. 

It no longer is a nop after #17279 . The conf package now does check if the icon theme is there - as it should. I would prefer to keep the number of Coq and CoqIDE packages as small as possible.

mseri:
Ok, that makes sense. You convinced me

mseri:
I'll wait for the CI and merge.

MSoegtropIMC:
> I'll wait for the CI and merge.

Thanks!

CI is failing - 3 failed cause of a timeout (the prerequisite Coq proofs a good part of known math which takes a while) but FreeBSD failed because there is no icon theme package. I created a VM with FreeBSD last week but couldn't figure out how to install FreeBSD with UI. I can further research this, though. Also open are OpenSuse and CentOS, afair.

MSoegtropIMC:
To be sure let me do a local test on all systems / VMs I have - it will take until today evening.

mseri:
Don't worry about freebsd, I think users would know what to do and can send us a PR if they stumble on it

MSoegtropIMC:
I just realized that I did test since 10 days with this PR in - I run my tests with a local patch repo and this change is in there - at least for CoqIDE 8.12.0. See https://github.com/coq/platform/blob/v8.12/opam/packages/coqide/coqide.8.12.0/opam. So this is locally tested on Ubuntu 16 & 18, Fedora 32, Mac and Windows (but known to fail on OpenSuse and CentOS cause of the wrong package name or test for adwaita - on my todo list).

So good to go from my side.

MSoegtropIMC:
Thanks for the quick merge - I appreciate it!

PR #17464, opened 2020-10-22
[new release] conf-openblas (0.2.1)

kkirstein:
fixes installation on Windows for `os = "win32" & os-distribution = "cygwinports"`

### DIFF
--- a/packages/conf-openblas/conf-openblas.0.2.1/opam
+++ b/packages/conf-openblas/conf-openblas.0.2.1/opam
@@ -1,0 +1,38 @@
+opam-version: "2.0"
+maintainer: "Liang Wang <ryanrhymes@gmail.com>"
+authors: [ "Liang Wang" ]
+homepage: "https://github.com/xianyi/OpenBLAS"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: "BSD-3-Clause"
+build: [
+  ["sh" "-exc" "cc $CFLAGS -I/usr/include/openblas test.c -lopenblas"]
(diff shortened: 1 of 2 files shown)


packages/conf-openblas/conf-openblas.0.2.1/opam
```diff
@@ -0,0 +1,38 @@
...
+  ["sh" "-exc" "cc $CFLAGS test.c -lcblas"]
+    {os-distribution = "arch"}
+  ["sh" "-exc" "cc $CFLAGS -I/usr/local/include -L/usr/local/lib test.c -lopenblas"]
+    {os = "freebsd"}
+  ["sh" "-exc" "x86_64-w64-mingw32-gcc $CFLAGS test.c -lopenblas"]
+    {os = "win32" & os-distribution = "cygwinports"}
+  ["sh" "-exc" "cc $CFLAGS test.c -lopenblas"]
+    {os-distribution != "fedora" & os-distribution != "centos" & os-family != "suse" & os != "macos" & os-distribution != "arch" & os != "freebsd" & os != "win32"}
+]
+depexts: [
```
mseri, line 24: Is openblas already part of cygwinports or it needs also a depext?

kkirstein: No, to my knowledge there is only a Cygwin-based openblas package, but no for native Windows (mingw), as needed by OCaml for Windows. Current status is to manually install openblas (I have instructions for that, but don't know where to put them).

mseri: It would be useful to have them, at least, on a section in the readme of owl, where we mention `common issues`

PR #17467, opened 2020-10-22
new conf-xen package to uniquify the xen headers, add FreeBSD depext

hannesm:
//cc @gaborigloi for xenctrl
once merged I'll PR to upstream about conf-xen package

### DIFF
--- a/packages/conf-xen/conf-xen.1/opam
+++ b/packages/conf-xen/conf-xen.1/opam
@@ -19,6 +19,7 @@ depexts: [
   ["xenstore"] {os-distribution = "arch"}
   ["xen-tools"] {os = "freebsd"}
 ]
+available: os != "macos"
 synopsis:
   "Virtual package relying on Xen headers"
 description: """
--- a/packages/conf-xen/conf-xen.1/opam
+++ b/packages/conf-xen/conf-xen.1/opam
@@ -1,0 +1,27 @@
+opam-version: "2.0"
+maintainer: "hannes@mehnert.org"
+authors: [ "hannes@mehnert.org" ]
+homepage: "https://xenproject.org/"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: [ "GPLv2" "LGPL3" ]
+build: [
+  ["sh" "-exc" "cc -c $CFLAGS -I/usr/local/include test.c"]
+]
+depexts: [
+  ["xen-dev"] {os-distribution = "alpine"}
+  ["libxen-dev"] {os-family = "debian"}
+  ["xen-devel"] {os-distribution = "centos"}
+  ["xen-devel"] {os-distribution = "fedora"}
+  ["xen-devel"] {os-distribution = "rhel"}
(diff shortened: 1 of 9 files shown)


packages/conf-xen/conf-xen.1/opam
```diff
@@ -0,0 +1,27 @@
...
+  ["xen-dev"] {os-distribution = "alpine"}
+  ["libxen-dev"] {os-family = "debian"}
+  ["xen-devel"] {os-distribution = "centos"}
+  ["xen-devel"] {os-distribution = "fedora"}
+  ["xen-devel"] {os-distribution = "rhel"}
+  ["xen-devel"] {os-distribution = "ol"}
+  ["xen-devel"] {os-family = "suse"}
+  ["xen-dom0-libs-devel" "xen-libs-devel"] {os = "xenserver"}
+  ["xenstore"] {os-distribution = "arch"}
+  ["xen-tools"] {os = "freebsd"}
```
mseri, line 20: oh nice! Also freebsd

hannesm:
FreeBSD installs headers into /usr/local, thus in order to fix the packages (xen-gnt-unix, xen-evtchn-unix) a `-I/usr/local/include` needs to be passed around. won't fix this for released versions, but eventually upstream. in any case, the introduction of a `conf-xen` package is imho an improvement.

hannesm:
CI failures:
- xenctrl on some platforms (that errored before as well)
- FreeBSD (as described above)
- solver timeout
- centos-8 "cannot find xen-devel package" (according to google, this should be the name of the rpm -- anyone using centos who may help out how the xen-devel package name is? and why the CI can't install it)?

imho this is good to be merged, and improves the current state.

mseri:
I agree. It is a good move to enable future improvements

psafont:
Just saw this. Thanks for the heads-up, we'll change xenctrl to use conf-xen.

packages/conf-xen/conf-xen.1/opam
```diff
@@ -0,0 +1,28 @@
+opam-version: "2.0"
+maintainer: "hannes@mehnert.org"
+authors: [ "hannes@mehnert.org" ]
+homepage: "https://xenproject.org/"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: [ "GPLv2" "LGPL3" ]
```
psafont, line 6: @hannesm I'm reviewing licenses used in xapi, and I see most of the fiels are licenses with `LGPL-2.1-only`, `LGPL-2.1-or-later` in xenctrl and the other C client libraries in xen, do you re member why you added only both these licenses? (there are other licenses as all files have define their own licenses and all files must be reviewed individually, unfortunately)

hannesm: It is likely I did not check any license by myself, but just copied the licenses from the FreeBSD port (https://www.freshports.org/sysutils/xen-tools/).

psafont: That makes sense, thanks!

PR #17480, opened 2020-10-23
add package for newly-released Zenon 0.8.5

damiendoligez:


kit-ty-kate:
Thanks! The `remove` field is unnecessary since opam 2.0. Could you return the fix upstream?

damiendoligez:
Sure. Next release, I'll just copy the `opam` file from 0.8.5 anyway...

PR #17483, opened 2020-10-24
mirage-kv uses Fmt.failwith, available only since 0.8.4

hannesm:


hannesm:
upstream PR https://github.com/mirage/mirage-kv/pull/22

kit-ty-kate:
Thanks!

PR #17499, opened 2020-10-25
[new release] waylaunch (0.2.0)

kit-ty-kate:
Waylaunch is a program launcher for Wayland

- Project page: <a href="https://github.com/kit-ty-kate/waylaunch">https://github.com/kit-ty-kate/waylaunch</a>

##### CHANGES:

* Get rid of the client code from bemenu and use bemenu's library directly
* Disable use of implicit dependencies


kit-ty-kate:
The package fails on Ubuntu 16.04 because the wayland library is too old

PR #17431, opened 2020-10-18
Markup.ml 1.0.0: standards-compliant HTML5 parser

aantron:
[Changelog](https://github.com/aantron/markup.ml/releases/tag/1.0.0):

> Breaking
>
> - Include tag attributes in [`` `Misnested_tag``](https://github.com/aantron/markup.ml/blob/27c4b85fdf5c27949f54e6cbe0d5adf1bee87af6/src/markup.mli#L132) errors, to assist in error recovery (aantron/markup.ml#55, Albert Peschar).
>
> Bugs fixed
>
> - Parser hangs on `<rb>` and other ruby annotation tags (aantron/markup.ml#56, reported Julien Sagot).

PR #17500, opened 2020-10-25
Re-release Markup.ml 1.0.0

aantron:
I'd like to amend the Markup 1.0.0 release from last week (#17431) with a new breaking change. cc @dmbaturin.

[Changelog](https://github.com/aantron/markup.ml/releases/tag/1.0.0):

> Breaking
>
> - Include tag attributes in [`` `Misnested_tag``](https://github.com/aantron/markup.ml/blob/27c4b85fdf5c27949f54e6cbe0d5adf1bee87af6/src/markup.mli#L132) errors, to assist in error recovery (aantron/markup.ml#55, Albert Peschar).
> - Delete type `content_signal` (aantron/markup.ml#59).
>
> Bugs fixed
>
> - Parser hangs on `<rb>` and other ruby annotation tags (aantron/markup.ml#56, reported Julien Sagot).
> - Preserve the original doctype when outputting HTML (aantron/markup.ml@b4b59ae, reported Daniil Baturin).
> - Pretty-printer: emit newline after doctype (aantron/markup.ml#58, reported Daniil Baturin).


dmbaturin:
@aantron Your opam file has a wrong checksum. 

You have this:

```
url {
  src: "https://github.com/aantron/markup.ml/archive/1.0.0.tar.gz"
  checksum: "md5=d41d8cd98f00b204e9800998ecf8427e"
}
```

But actual tarball has:

```
$  wget https://github.com/aantron/markup.ml/archive/1.0.0.tar.gz
$ md5sum ./1.0.0.tar.gz 
d9964f196850af5507b66b7a141bff11  ./1.0.0.tar.gz
```

aantron:
@dmbaturin Thanks, fixed. Was a firewall issue.

mseri:
Do we have to re-release? There have been a lot of complaints about mutability of releases, is there no other option in this case?

I will also ping @kit-ty-kate for an opinion

avsm:
If this is re-released, please bump the epoch (so `markup.1.0.0-1`) to indicate the checksum has changed for a known reason. This seems like a reasonable convention in the event that upstream really wants a re-release.

aantron:
I changed the version number to `1.0.0-1`.

I still think this is unnecessary in this particular case, as a fast merge is likely to avoid causing problems for downstream projects.

avsm:
> I still think this is unnecessary in this particular case, as a fast merge is likely to avoid causing problems

Unfortunately, we can't predict when the people using opam-repository take their snapshots for their own deployments, so this hasn't been true in the past. (I caused great drama in the Tezos downstream opam-repo with a re-release of uri last year, for instance).

dmbaturin:
How should I specify that version as a dependency? Will `markup { >= 1.0.0-1 }` work as expected?

avsm:
Yes, if you add quotes around the version.

packages/markup/markup.1.0.0/opam
```diff
@@ -2,7 +2,7 @@ opam-version: "2.0"
 
 synopsis: "Error-recovering functional HTML5 and XML parsers and writers"
 
-version: "1.0.0"
+version: "1.0.0-1"
```
kit-ty-kate, line 5: this will not work as the version in the directory name is going to be taken instead.

kit-ty-kate:
Since it is already changing versions, could it be 1.1.0 or 2.0.0 to reflect the API break?
Also the tag changed. Could you add `available: false` to the `markup.1.0.0` package?

aantron:
I created a `markup.1.0.0-1` directory with the new `opam` file, reverted the previous `markup.1.0.0/opam`, and added `available: false` to it.

aantron:
Thank you!

PR #17520, opened 2020-10-28
Revert "Move change to coqide.8.12.0-1"

mseri:
This reverts commit a037ee74ee61dc319e756b6212728e5c7017b78e.

The package was never installable, the conf package in 8.12.0 was broken
and needs to be update to conf-adwaita as pointed oyt by @MarkPantel and
@Zimmi48


mseri:
The package has never been installable due to the `=version` constraint

PR #17531, opened 2020-10-29
Revised detection for conf-capnproto

dra27:
See also https://github.com/capnproto/capnp-ocaml/pull/75

`capnpc` was replaced with `capnp` when the compiler was switched from Haskell to C++ (see [the 0.2 news item](https://capnproto.org/news/2013-08-12-capnproto-0.2-no-more-haskell.html)). Ubuntu certainly, and I guess others, symlink `capnpc` to `capnp`, but the Windows binary distribution only has `capnp.exe`.

I've put this in a new package, rather than updating the `.0` package, since if this does break anything it should be clear that it was a package upgrade which did it!

### DIFF
--- a/packages/conf-capnproto/conf-capnproto.1/opam
+++ b/packages/conf-capnproto/conf-capnproto.1/opam
@@ -1,0 +1,28 @@
+opam-version: "2.0"
+maintainer: "Kate <kit.ty.kate@disroot.org>"
+authors: [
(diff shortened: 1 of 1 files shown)


dra27:
cc @talex5

mseri:
Thanks!

mseri:
Thanks!

PR #17534, opened 2020-10-29
Improve conf-gles2

kit-ty-kate:
This should hopefully handle every systems opam can handle


mseri:
For some reason it is still failing on Travis' freebsd. The library is correctly installed and should come with the glesv2.pc file from what I can see from the port informations

kit-ty-kate:
Yeah it works for me on my local FreeBSD... I'm not sure what's wrong :/

mseri:
Maybe it is something with Travis environment or Travis version of freeBSD 🤷 
Let's merge, if somebody has issues with it, we can look more deeply into it

PR #17535, opened 2020-10-30
Add dependencies and bounds to memtrace_viewer

lpw25:
memtrace_viewer doesn't currently install in a fresh opam switch. I installed some versions of packages manually and then it built successfully. This PR adds those packages as dependencies and bounds in the opam files for memtrace_viewer.


mseri:
Thanks

PR #17537, opened 2020-10-30
conf-gles2: Fix FreeBSD

kit-ty-kate:
Fixes issue discovered in https://github.com/ocaml/opam-repository/pull/17534
cc @mseri somehow the files prefixed by `@comment` in FreshPorts are not installed by the package and I already had `mesa-libs` installed locally..


mseri:
Do we need to install both? In the port description I thought that `mesa-devel` was the one installing the pkgconfig file. 
EDIT: No, looks like `libs` is the good one: https://svnweb.freebsd.org/ports/head/graphics/mesa-libs/pkg-plist?view=markup#l50

kit-ty-kate:
FreeBSD CI is Green!

PR #17545, opened 2020-10-31
[new release] erlang (0.0.13)

ostera:
CHANGES:

* erlang: prepare erlang library for initial release to opam.

### DIFF
--- a/packages/erlang/erlang.0.0.13/opam
+++ b/packages/erlang/erlang.0.0.13/opam
@@ -1,0 +1,47 @@
+opam-version: "2.0"
+synopsis: "Libraries to manipulate Erlang sources"
+description: """
+erlang is a  a set of libraries designed to facilitate manipulating
+ Standard Erlang and Core Erlang sources.
+
+ It provides a lexer/parser, a concrete AST, and a printer for Standard Erlang
+ in its current version.
+"""
+maintainer: ["Leandro Ostera <leandro@ostera.io>"]
+authors: ["Leandro Ostera <leandro@ostera.io>"]
+license: "Apache-2.0"
+homepage: "https://github.com/AbstractMachinesLab/caramel"
+bug-reports: "https://github.com/AbstractMachinesLab/caramel/issues"
+depends: [
+  "dune" {>= "2.7" & >= "2.7"}
+  "ocaml" {>= "4.11.1"}
+  "cmdliner"
+  "menhir"
+  "ppx_sexp_conv"
+  "sexplib"
+  "odoc" {with-doc}
+]
+build: [
+  ["dune" "subst"] {dev}
+  [
+    "dune"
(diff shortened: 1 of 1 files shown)


mseri:
Not sure what happened with the CI. The failure on arm looks weird, I would have not expected it

ostera:
I think I'm gonna mark it as a draft until I figure out how to get the `caramel` tests out of the source tarball for `erlang` -- the failure makes sense, since those tests require binaries that are not being built as part of the `erlang` package.

mseri:
There are similar issues with ppx packages. In that case we tend to move the tests on the package that is required to run them, even if they happen to be tests of a parent package. Alternative is to create a separate test package (this happens in ocaml-rpc, where the dummy rpc package is there both for compatibility and to run some tests that require the whole bundle)

mseri:
In think opam 2.1 allows you to use caramel as a test dependency of erlang without creating a dependency cycle, but it is still in beta

ostera:
Here's a relevant issue: https://github.com/ocaml/dune/issues/3792

It also looks like I can't just remove these sources _before_ calling `dune-release distrib -n erlang -p erlang` because the tag is already there and _the tagged sources_ will be used to create the tarball.

But I need the tag to trigger the release process 😅 

I may need to do some restructuring.

ostera:
I think I found a workaround.

The problem seems to be that running `dune build .... @runtest` will compile all of the tests in the sources, which includes also the ones from caramel.

We don't want that, so instead we can use `dune runtest ./erlang/tests -p erlang` and let _that_ compile the relevant tests based on that path filtering.

It is running now, lets see how it goes.

ostera:
Alright, that seemed to work! Let's see what camelus thinks.

packages/erlang/erlang.0.0.13/opam
```diff
@@ -0,0 +1,47 @@
...
+ It provides a lexer/parser, a concrete AST, and a printer for Standard Erlang
+ in its current version.
+"""
+maintainer: ["Leandro Ostera <leandro@ostera.io>"]
+authors: ["Leandro Ostera <leandro@ostera.io>"]
+license: "Apache-2.0"
+homepage: "https://github.com/AbstractMachinesLab/caramel"
+bug-reports: "https://github.com/AbstractMachinesLab/caramel/issues"
+depends: [
+  "dune" {>= "2.7" & >= "2.7"}
```
mseri, line 16: Looks good overall, I think this is dune trying to be smart but not enough with constraints 
```suggestion
  "dune" {>= "2.7"}
```

ostera:
Alright, this looks good. I'll make one last change so there's no trace of the Caramel compiler in this lib and I'll resubmit.

ostera:
Closing this in favor of #17553

PR #17554, opened 2020-11-02
[new release] elpi (1.12.0)

gares:
CHANGES:

- FFI:
  - `RawOpaqueData.cin` now returns a term and takes `constants` into account.
    Whenever a value is represented with a named constant, the API grants that:
    - the value is embedded using that constant
    - that constant is read back to the original value
- API:
  - New `Compile.extend` to (quickly) add clauses to a program
  - New argument `?follows:program` to `Compile.unit` to have the unit
    be compiled using the very same symbol table (no new symbols can be
    declared by the unit in this case)
- Library:
  - rename `map2_filter` -> `map2-filter`
  - new `map-filter`
- Build:
  - use md5 (OCaml's digest) instead of sha1 (external utility)

### DIFF
--- a/packages/elpi/elpi.1.12.0/opam
+++ b/packages/elpi/elpi.1.12.0/opam
@@ -1,0 +1,84 @@
+opam-version: "2.0"
+maintainer: "Enrico Tassi <enrico.tassi@inria.fr>"
+authors: [ "Claudio Sacerdoti Coen" "Enrico Tassi" ]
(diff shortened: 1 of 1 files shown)


mseri:
Thanks. I will wait for the big CI before merging, assuming it survives

gares:
It frankly looks... stuck

mseri:
Looks good!

gares:
thanks

PR #17558, opened 2020-11-02
add z3 4.8.9

c-cube:


mseri:
Looks good. Should we merge or it has to be considered still a draft?

c-cube:
I think it's good, I also compiled and tested it locally :)

PR #15970, opened 2020-03-10
[new release] checkseum (0.2.0)

dinosaure:
Adler-32, CRC32 and CRC32-C implementation in C and OCaml

- Project page: <a href="https://github.com/mirage/checkseum">https://github.com/mirage/checkseum</a>
- Documentation: <a href="https://mirage.github.io/checkseum/">https://mirage.github.io/checkseum/</a>

##### CHANGES:

- fix cross-compilation with `dune -x windows` (mirage/checkseum#45, @dinosaure, @pirbo)
- add CRC-24 (mirage/checkseum#43, @dinosaure, @cfcs)
- factorize C stubs (as digestif)
- avoid clash of names when we use `checkseum.c`
  Any functions are prefixed by `checkseum_`
- fix META file (mirage/checkseum#39 & mirage/checkseum#41, @hannesm, @dinosaure)
  A test was added to see if runes (static C libraries) are available for
  MirageOS targets (freestanding & xen)
- provide a binary `checkseum` to _digest_ standard input or file
  `checkseum.checkseum` is available to compute check-sum of standard input
  or file. The tool is used only for debugging.
- clean distribution (mirage/checkseum#38, @dinosaure)
  `checkseum` depends only on `bigarray-compat`, `base-bytes` & `optint`
- `limits.h` is available on any targets (mirage/checkseum#37, @dinosaure, @pirbo)

PR #17566, opened 2020-11-03
[new release] checkseum (0.3.0)

dinosaure:
Adler-32, CRC32 and CRC32-C implementation in C and OCaml

- Project page: <a href="https://github.com/mirage/checkseum">https://github.com/mirage/checkseum</a>
- Documentation: <a href="https://mirage.github.io/checkseum/">https://mirage.github.io/checkseum/</a>

##### CHANGES:

- Upgrade C artifacts with MirageOS 3.9 (mirage/checkseum#50, @dinosaure, @hannesm)
- Fix `esy` installation (mirage/checkseum#49, @dinosaure, @jordwalke, reported by @Faliszek)

### DIFF
--- a/packages/checkseum/checkseum.0.3.0/opam
+++ b/packages/checkseum/checkseum.0.3.0/opam
@@ -1,0 +1,62 @@
+opam-version: "2.0"
+maintainer:   [ "Romain Calascibetta <romain.calascibetta@gmail.com>" ]
+authors:      [ "Romain Calascibetta <romain.calascibetta@gmail.com>" ]
+homepage:     "https://github.com/mirage/checkseum"
+bug-reports:  "https://github.com/mirage/checkseum/issues"
+dev-repo:     "git+https://github.com/mirage/checkseum.git"
+doc:          "https://mirage.github.io/checkseum/"
+license:      "MIT"
+synopsis:     "Adler-32, CRC32 and CRC32-C implementation in C and OCaml"
+description: """
+Checkseum is a library to provide implementation of Adler-32, CRC32 and CRC32-C
+in C and OCaml.
+
+This library use the linking trick to choose between the C implementation
+(checkseum.c) or the OCaml implementation (checkseum.ocaml). This library is on
+top of optint to get the best representation of an int32. """
+
+build: [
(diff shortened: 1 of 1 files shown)


packages/checkseum/checkseum.0.3.0/opam
```diff
@@ -0,0 +1,62 @@
...
+top of optint to get the best representation of an int32. """
+
+build: [
+  [ "dune" "build" "-p" name "-j" jobs ]
+  [ "./install/install.ml" ]
+  [ "dune" "runtest" "-p" name "-j" jobs ] {with-test}
+]
+
+install: [
+  [ "dune" "install" "-p" name ] {with-test}
```
mseri, line 25:
```suggestion
  [ "dune" "install" "-p" name ]
```

Does it actually install the package if you have `install: [ ]`?

dinosaure: If I remember correctly, I tagged `dune install` to be able to launch correctly the next line (`test_runes.ml`) (which needs an installation of `checkseum`) - the idea is to launch `test_runes.ml` **after** the `dune`'s installation (which, I think, does not exactly follow the `opam` installation). To be sure to wrap such trick only into my tests, I added `{with-test}` on both.

May be, for a release, the best is to delete the whole `install: [ ]`.

mseri: Yes, that is clear, my worry was that in the case without tests, install becomes empty, and I don't know what opam will do. Does it still install the library or it considers it empty?

If you remove the `with-test` flag on the first line, it will always install the package both before running that test or in the case tests should not run.

I prefer not to risk the `install: [ ]` case, maybe now it works but could change behaviour in the future, and either remove those tests or make the change that I suggested above.

dinosaure: So, with a search with an old release of `checkseum`: https://github.com/ocaml/opam-repository/pull/15970#discussion_r390649256
It seems that the best is to remove all the `install: [ ]` field as I did [here](https://github.com/ocaml/opam-repository/pull/15970/commits/7c6d5a5571f1f2ab89bd42c4c4b9493e53042894).

PR #17581, opened 2020-11-05
Various fixes (OCaml 4.12 edition, round 2)

kit-ty-kate:
Patch for `extlib` proposed in https://github.com/ygrek/ocaml-extlib/pull/55


kit-ty-kate:
Patch for `tls` proposed in https://github.com/mirleft/ocaml-tls/pull/421

PR #17588, opened 2020-11-06
introduces ppx_bap, a ppx rewriter for BAP needs

ivg:
This package is merely a copy of the ppx_jane with an amended set of
ppx rewriters, namely with the ppx_js_style removed, which makes it
impossible to use ppx_jane (and ppx_base) especially when the code is
generated, e.g., when the `findlib.dynload` library is used together
with ocamlfind, which requires the inclusion of a generated code that
triggers a fatal warning in ppx_js_style. We provided [a fix][1] for
that, which was ignored. Another example, is code generated by various
protocol compilers, namely piqi-ocaml, for which we also provided a
[fix][2] that got merged but was never released. In other words,
ppx_js_style is applying itself to the code that neither belongs to
Jane Street nor even to a user of the ppx_jane rewriter.

We have [been asking Jane Street][5] to remove ppx_js_style for a long
time but we were just ignored (no replies in 5 months). Therefore we
decided to create our own package with the set of ppx rewriters that
we use in BAP (which also solves another issue associated with
[ppx_jane][4]).

Once this package is merged we can move BAP to the newest versions of
OCaml, up to 4.11.1, see [the PR to BAP][5].

[1]: https://github.com/ocaml/ocamlfind/pull/9
[2]: https://github.com/alavrik/piqi-ocaml/pull/15
[3]: https://github.com/janestreet/ppx_jane/issues/5
[4]: https://github.com/janestreet/ppx_jane/issues/6
[5]: https://github.com/BinaryAnalysisPlatform/bap/pull/1116

### DIFF
--- a/packages/ppx_bap/ppx_bap.v0.14.0/opam
+++ b/packages/ppx_bap/ppx_bap.v0.14.0/opam
@@ -1,0 +1,39 @@
+opam-version: "2.0"
+maintainer: "ivg@cmu.edu"
+authors: ["Carnegie Mellon University BAP Team"]
+homepage: "https://github.com/BinaryAnalysisPlatform/ppx_bap"
+bug-reports: "https://github.com/BinaryAnalysisPlatform/ppx_bap/issues"
+dev-repo: "git+https://github.com/BinaryAnalysisPlatform/ppx_bap.git"
+license: "MIT"
+build: [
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
No worries at all. Thanks a lot!

PR #17589, opened 2020-11-07
Pull release of cohttp 3 (available = false)

mseri:
We will release a 3.0.0-1 with some more breaking changes. See https://github.com/mirage/ocaml-cohttp/pull/734 and https://github.com/mirage/ocaml-cohttp/issues/733


kit-ty-kate:
Thanks!

PR #17598, opened 2020-11-09
tezos-p2p: Relax constraints on Lwt

kit-ty-kate:
@pirbo the package compiles. Is there a particular reason why the constraint is there?

### DIFF
--- a/packages/tezos-p2p/tezos-p2p.7.4/opam
+++ b/packages/tezos-p2p/tezos-p2p.7.4/opam
@@ -10,7 +10,7 @@ depends: [
   "dune" { >= "1.11" }
   "tezos-p2p-services" { = version }
   "alcotest-lwt" { with-test & = "0.8.5" }
-  "lwt" { >= "4.0" & < "4.3" }
+  "lwt" { >= "4.0" }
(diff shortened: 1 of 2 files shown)


pirbo:
Tests are not compatible with Lwt 5 (because of the deprecation of Lwt_unix.set_default_async_method).

I'm OK with deactivating the tests from the opam package and remove that constraint. Tezos binaries works indeed well when compiled with lwt 5.

kit-ty-kate:
@pirbo ok, does the current patch look ok for you? I changed it so it will effectively only use lwt < 4.3 when testing, and any version >= 4.0 otherwise (in most cases)

kit-ty-kate:
if this looks good, could it be ported upstream? (if compatibility with Lwt >= 4.3 is not already fixed)

pirbo:
Seems good to me, yes.
Uptream and version 8.0 (to be released next week the latest) are fully compatible with lwt >= 4.3

kit-ty-kate:
> Uptream and version 8.0 (to be released next week the latest) are fully compatible with lwt >= 4.3

Oh, in that case I'll just wait for the next release and keep my patches local for testing purpose. Thanks!

PR #17600, opened 2020-11-09
Enlarge OCaml compatibility of bigarray-compat (4.02.3)

dinosaure:
I think it's fine for `bigarray-compat` to support OCaml 4.02.3 (`dune.1.0.0` seems works for OCaml 4.02.3).


mseri:
Looks like it is building, the revdeps are run with recent compilers, so they will just test what was already tested

PR #17621, opened 2020-11-14
conf-blas fails to install on Manjaro linux

code-ghalib:
Seems like a repeat of https://github.com/ocaml/opam-repository/issues/12989

`opam config list` shows:
```
opam-version      2.0.7             # The currently running opam version
os                linux             # Inferred from system
os-distribution   manjaro           # Inferred from system
os-family         arch              # Inferred from system
os-version        20.2              # Inferred from system
```
I suspect `os-family` may be a more appropriate fix.


mseri:
Thanks, indeed `os-family` is broader. I think there is a plan to fix this over all conf-packages soon

PR #17623, opened 2020-11-14
bibtex2html 1.99: do not compile doc

backtracking:


mseri:
Please can you make this change in a new patch package `bibtex2html.1.99-1`: practically you would just need to copy the `opam` file to `packages/bibtex2html/bibtex2html.1.99-1/opam` and making the change there

backtracking:
Sure, no problem. Do I need to include `files/` from `bitex2html-1.99` as well?

mseri:
You can avoid doing that if you remove lines 20-23 in the file, that is you drop both `patches` and `remove`. 

This should be fine since with opam2 the `remove` section is in most cases redundant: the new opam tracks changes at install time and uses this information to know what to remove

backtracking:
Thanks! Then I guess I should also remove lines 25--29 about extra files, right?

mseri:
Yes, indeed!

backtracking:
Thanks. Done.

mseri:
Thanks!

PR #17630, opened 2020-11-15
Various fixes (OCaml 4.12 edition, round 5)

kit-ty-kate:


kit-ty-kate:
@thierry-martinez is that ok to remove the upper-bound for `pyml`? Was there a specific reason why it was there?

PR #17551, opened 2020-11-01
[new release] ocaml-protoc-plugin (4.1.0)

andersfugmann:
Plugin for protoc protobuf compiler to generate ocaml definitions from a .proto file

- Project page: <a href="https://github.com/issuu/ocaml-protoc-plugin">https://github.com/issuu/ocaml-protoc-plugin</a>

##### CHANGES:

- [x] Fix bug with Proto2 default integer arguments for Int32 and
      Int64 types
- [x] Add function to construct messages with default values
- [x] Add missing includes for google well known types

PR #16401, opened 2020-05-10
ocaml-protoc-plugin.4.0.0

andersfugmann:
This pull-request concerns:
-`ocaml-protoc-plugin.4.0.0`: Plugin for protoc protobuf compiler to generate ocaml definitions from a .proto file

---

---
:camel: Pull-request generated by opam-publish v2.0.2

PR #17632, opened 2020-11-15
Try making a conf-oniguruma package

alan-j-hu:
I am writing bindings for the oniguruma library, so I need a conf-oniguruma package. However, I am not sure if I am making it properly, so I want to test the CI and see what happens.

I searched up the Oniguruma development packages on https://pkgs.org. I do not actually run all of these operating systems. So, I might have made a mistake.

### DIFF
--- a/packages/conf-oniguruma/conf-oniguruma.1/opam
+++ b/packages/conf-oniguruma/conf-oniguruma.1/opam
@@ -7,14 +7,19 @@ license: "BSD-2-Clause"
 build: [["pkg-config" "oniguruma"]]
 depends: ["conf-pkg-config" {build}]
 depexts: [
-  ["oniguruma-dev"] {os-family = "alpine"}
-  ["oniguruma-devel"] {os-distribution = "centos"}
   ["libonig-dev"] {os-family = "debian"}
+  ["libonig-dev"] {os-family = "ubuntu"}
   ["oniguruma-devel"] {os-distribution = "fedora"}
-  ["oniguruma"] {os-distribution = "homebrew"}
+  ["oniguruma-devel"] {os-distribution = "centos"}
+  ["oniguruma-devel"] {os-distribution = "ol"}
   ["oniguruma-devel"] {os-family = "suse"}
+  ["oniguruma-dev"] {os-family = "alpine"}
+  ["oniguruma"] {os-family = "arch"}
+  ["oniguruma"] {os-distribution = "homebrew"}
+  ["oniguruma"] {os = "freebsd"}
+  ["oniguruma"] {os = "netbsd"}
 ]
-available: os-family != "bsd"
+available:
 synopsis: "Virtual package relying on an Oniguruma system installation"
 description:
   "This package can only install if the Oniguruma development package is installed on the system."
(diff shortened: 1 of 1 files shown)


alan-j-hu:
Looks like the FreeBSD build failed, since I didn't specify a system package for FreeBSD.

I searched up a FreeBSD package on https://pkgs.org/search/?q=oniguruma.h&on=provides, but there were no results.

alan-j-hu:
Looks like the MacOS Travis build failed because of some error generating the build script.

alan-j-hu:
Looks like the CentOS build failed because oniguruma-devel is in the Powertools repository. This has happened for one other package, but the package was still allowed to be published: https://github.com/ocaml/opam-repository/pull/16401

I need to figure out what happened with OpenSUSE.

alan-j-hu:
Okay, I did `os-distribution = opensuse`, but the CI's `os-distribution` reported `opensuse-leap`. I should have done `os-family = suse`.

alan-j-hu:
The MacOS build failed again due to an error generating the build script.

alan-j-hu:
This PR should be good to merge...

1. The only reason the Mac build failed in the most recent Travis CI was an error on Travis CI's part.
2. The reason why the CentOS build failed in the big CI is that the powertools repository is needed. There is prior precedent for accepting packages with external dependencies from the powertools repository (https://github.com/ocaml/opam-repository/pull/16401, https://github.com/ocaml/opam-repository/pull/17551).

alan-j-hu:
Are the `oniguruma` packages for the BSDs correct? At least according to https://freebsd.pkgs.org/13/freebsd-aarch64/oniguruma-6.9.5.r1_1.txz.html, on `FreeBSD` the `oniguruma` package only installs the `.so` file, but the `.h` file is also needed for bindings to compile.

Maybe I will add a test `.c` file to compile to make sure that the necessary files are installed,

alan-j-hu (commented):
Are the oniguruma packages for the BSDs correct? At least according to https://freebsd.pkgs.org/13/freebsd-aarch64/oniguruma-6.9.5.r1_1.txz.html, on FreeBSD the oniguruma package only installs the `.so` file, but the `.h` file is also needed for bindings to compile.

Maybe I will add a test `.c` file to compile to make sure that the necessary files are installed.

kit-ty-kate:
Are you sure? The freshports page (https://www.freshports.org/devel/oniguruma/) is listing the `.h`, `.so` and `.pc`

alan-j-hu:
I am not an expert on BSD, but according to my cursory research, a BSD "port" is how you build packages from source, and it's separate from the package manager.

https://www.freebsd.org/ports/

https://www.freebsd.org/doc/handbook/ports.html

alan-j-hu:
Okay, so the development package should also install the `.pc` file. Since the build step uses `pkg-config`, if any of the packages listed in the opam file only installs the shared object file, the build should fail.

kit-ty-kate:
Thanks!

alan-j-hu:
(To clarify what I meant: By "development package," I meant the packages (usually named `libname-dev` or `libname-devel`) that install the `.h` and `.pc` files, among other files and which are distinct from `libname`. `libname` usually only installs the `.so` file. When I wrote the opam file, I checked to make sure I was specifying the development package. I suspect that the BSD packages you added will only install the `.so` file, which will lead to a build failure if a user tries to install the dependency with `opam depext`. The libraries you found that contained the header file were BSD "ports," which are a different concept from BSD packages installable with the package manager.)

kit-ty-kate:
> The libraries you found that contained the header file were BSD "ports," which are a different concept from BSD packages installable with the package manager.

Mh, are you sure? I'm a FreeBSD user and I'm a bit confused here.

alan-j-hu:
I don't use FreeBSD, so I will trust your knowledge.

PR #17633, opened 2020-11-16
Relax some dependency constraints for key-parsers and zxcvbn

bbc2:
I don't know why those higher bounds where there for ppx_deriving and ppx_deriving_yojson and not the other dependencies.  Additionally, the packages seem to work fine with ppx_deriving 5.

I figured this is the easiest way to get the dependencies fixed.  Let me know if I should release new versions of key-parsers and zxcvbn instead.


kit-ty-kate:
I'm guessing the constraints were there because it copied some packages doing the same thing (notably `ppx_deriving_yojson` before 3.6.0) as some of them relied on the ppx_deriving API and not just the plugins (ppx_deriving.std, ppx_deriving.show, …)

CI still picked ppx_deriving 4.5 but if you say it works that's good enough for me (seeing as it doesn't use the api). Thanks!

PR #17635, opened 2020-11-16
PARTIALLD not set for ocaml-option-32bit

dra27:
Small regression from #17541


dra27:
cf. https://github.com/ocaml/opam-repository/blob/a6b26bd5c5d69de16b6471a29624f5441030252d/packages/ocaml-variants/ocaml-variants.4.11.1%2B32bit/opam#L29

kit-ty-kate:
Just waiting for Camelus to double-check syntax and merging right after.

kit-ty-kate:
Thanks!

PR #17644, opened 2020-11-17
[new release] ppx_deriving_cmdliner (0.5.1)

trepetti:
CHANGES:

- Add support for OCaml 4.10.0, 4.11.0, and 4.11.1. (Fix for https://github.com/hammerlab/ppx_deriving_cmdliner/issues/40)

### DIFF
--- a/packages/ppx_deriving_cmdliner/ppx_deriving_cmdliner.0.5.1/opam
+++ b/packages/ppx_deriving_cmdliner/ppx_deriving_cmdliner.0.5.1/opam
@@ -1,0 +1,45 @@
+opam-version: "2.0"
+maintainer: "Seb Mondet <seb@mondet.org>"
+authors: [
+  "Isaac Hodes <isaachodes@gmail.com>"
+  "B. Arman Aksoy <arman@aksoy.org>"
+  "Seb Mondet <seb@mondet.org>"
+  "Nick Zalutskiy <nick@const.fun>"
+  "Armaël Guéneau <armael.gueneau@ens-lyon.fr>"
+]
+homepage: "https://github.com/hammerlab/ppx_deriving_cmdliner"
+bug-reports: "https://github.com/hammerlab/ppx_deriving_cmdliner/issues"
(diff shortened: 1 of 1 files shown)


packages/ppx_deriving_cmdliner/ppx_deriving_cmdliner.0.5.1/opam
```diff
@@ -0,0 +1,45 @@
...
+]
+run-test: [
+  ["jbuilder" "runtest" "-p" name "-j" jobs]
+]
+depends: [
+  "ocaml" {>= "4.03"}
+  "cmdliner" {>= "1.0.0"}
+  "result"
+  "ppx_deriving" {>= "4.0" & < "5.0"}
+  "dune" {< "2.0"}
```
kit-ty-kate, line 29: Is this expected? Looking at the code it seems to use dune, not jbuilder as the build/test step imply.

kit-ty-kate: I fixed it directly. If it doesn't work, our CI will tell us.

trepetti: Thank you @kit-ty-kate. Yes, I should have updated the project opam here and upstream. My next contribution upstream will be moving to dune 2.0+, so I will modify the opam template in the upstream repo then. I do not believe there are any incompatible features from dune 1.0 in use, so it should be straightforward.

trepetti: Looks like the x86_64 Linux test failed to start. It errored out without a log after 1 second. Others seem to have worked fine.

kit-ty-kate:
Thanks a lot! Note also that ppx_deriving 5.0 is out and that it switched to ppxlib instead of ppx_tools. The translation might take an hour or two if you have the time. If you don't I'll look into it.

PR #17649, opened 2020-11-17
core_kernel v0.14.1

cwong-ocaml:

### DIFF
--- a/packages/core_kernel/core_kernel.v0.14.1/opam
+++ b/packages/core_kernel/core_kernel.v0.14.1/opam
@@ -1,0 +1,47 @@
+opam-version: "2.0"
+maintainer: "opensource@janestreet.com"
+authors: ["Jane Street Group, LLC <opensource@janestreet.com>"]
+homepage: "https://github.com/janestreet/core_kernel"
+bug-reports: "https://github.com/janestreet/core_kernel/issues"
+dev-repo: "git+https://github.com/janestreet/core_kernel.git"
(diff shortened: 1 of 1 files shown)


packages/core_kernel/core_kernel.v0.14.1/opam
```diff
@@ -0,0 +1,47 @@
...
+homepage: "https://github.com/janestreet/core_kernel"
+bug-reports: "https://github.com/janestreet/core_kernel/issues"
+dev-repo: "git+https://github.com/janestreet/core_kernel.git"
+doc: "https://ocaml.janestreet.com/ocaml-core/latest/doc/core_kernel/index.html"
+license: "MIT"
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+]
+depends: [
+  "ocaml" {>= "4.12.0"}
```
kit-ty-kate, line 13: Is this expected? I would have expected `>= "4.08"` instead. It worked pretty well in https://github.com/janestreet/core_kernel/pull/101

I would rather avoid fragmentation of the ecosystem if possible.

kit-ty-kate (changes requested):

packages/core_kernel/core_kernel.v0.14.1/opam
cwong-ocaml: My mistake, pushed a fix.

kit-ty-kate:
Thanks!

PR #17659, opened 2020-11-18
Fix primes package build instructions

lefessan:
While building documentation, @thomasblanc noticed that the `primes` package files are broken, the .cma/.cmxa are executable install of libraries. This PR should fix these problems.


mseri:
can we move these changes to `1.3.x-1` patch versions?

kit-ty-kate:
> can we move these changes to `1.3.x-1` patch versions?

In this particular case I don't think it's needed. The library is not used anywhere and couldn't be used anyway.

I'll merge once Travis succeeds.

kit-ty-kate:
Travis seems rather dead. It looks fine within our CI so let's merge this. Thanks!

PR #17670, opened 2020-11-20
Frama-C: new release (22.0)

AllanBlanchard:


packages/frama-c/frama-c.22.0/opam
```diff
@@ -0,0 +1,146 @@
...
+run-test: [
+  [make "-j%{jobs}%" "PTESTS_OPTS=-error-code" "tests"]
+]
+
+depends: [
+  "ocaml" { >= "4.08.1" }
+  "ocamlgraph" { >= "1.8.8" }
+  "ocamlfind" # needed beyond build stage, used by -load-module
+  "zarith"
+  "conf-autoconf" { build }
```
mseri, line 109:
```suggestion
  "conf-autoconf" { build }
  "conf-time" {with-test}
```

Builds are currently failing with
```
+ /usr/bin/make "-j47" "PTESTS_OPTS=-error-code" "tests" (CWD=/home/opam/.opam/4.08/.opam-switch/build/frama-c.22.0)
- Running      ptests
- /bin/sh: 1: time: not found
- make: *** [Makefile:1354: tests] Error 127
[ERROR] The compilation of frama-c failed at "/usr/bin/make -j47 PTESTS_OPTS=-error-code tests".
```

AllanBlanchard:
I do not understand the remaining failure, de are supposer to use lablgtk3 (without canvas) when a package forme lablgtk2 isn't available 🤔

kit-ty-kate:
@AllanBlanchard you can ignore the issues with libgnomecanvas being unavailable for now. However I encourage you to find alternatives because most distributions are migrating away from gtk2. For instance debian-testing – in line for the next release of Debian – doesn't have gnomecanvas.

However there are still failures to be looked at:
* Some of your tests fail on ARM64.
```
# --- tests/syntax/oracle/cpp-command.0.res.oracle	2020-11-20 00:00:00.000000000 +0000
# +++ tests/syntax/result/cpp-command.0.res.log	2020-11-20 23:31:37.955251981 +0000
# @@ -1,2 +1,2 @@
#  [kernel] Parsing tests/syntax/cpp-command.c (with preprocessing)
# -[cpp-command.c cpp-command.c cpp-command.c cpp-command.c] [TMPDIR/FILE.i TMPDIR/FILE.i TMPDIR/FILE.i TMPDIR/FILE.i] [ -I./share/libc -D__FRAMAC__ -D__FC_MACHDEP_X86_32 -dD -nostdinc -m32]
# +[cpp-command.c cpp-command.c cpp-command.c cpp-command.c] [TMPDIR/FILE.i TMPDIR/FILE.i TMPDIR/FILE.i TMPDIR/FILE.i] [ -I./share/libc -D__FRAMAC__ -D__FC_MACHDEP_X86_32 -dD -nostdinc]
# Env:
# FRAMAC_LIB (set from outside) = "/home/opam/.opam/4.11/.opam-switch/build/frama-c.22.0/lib/fc"
# FRAMAC_PLUGIN_GUI = "./lib/plugins/gui"
# FRAMAC_PLUGIN = "./lib/plugins"
# FRAMAC_SHARE = "./share"
# FRAMAC_SESSION = "."
# OCAMLRUNPARAM = ""
# Command:
# bin/toplevel.opt -check tests/syntax/cpp-command.c   -no-autoload-plugins -cpp-frama-c-compliant -cpp-command "echo %%1 = \$(basename '%1') %%2 = '%2' %%args = '%args'"  2>/tmp/cpp-command.1e1addf.err.log | /bin/sed "s:/[^ ]*/cpp-command\.[^ ]*\.i:TMPDIR/FILE.i:g; s:$PWD/::" >tests/syntax/result/cpp-command.1.res.log && /bin/sed "s:/[^ ]*/cpp-command\.[^ ]*\.i:TMPDIR/FILE.i:g; s:$PWD/::" < /tmp/cpp-command.1e1addf.err.log >tests/syntax/result/cpp-command.1.err.log && rm -f /tmp/cpp-command.1e1addf.err.log
# --- tests/syntax/oracle/cpp-command.1.res.oracle	2020-11-20 00:00:00.000000000 +0000
# +++ tests/syntax/result/cpp-command.1.res.log	2020-11-20 23:31:38.007251614 +0000
# @@ -1,2 +1,2 @@
#  [kernel] Parsing tests/syntax/cpp-command.c (with preprocessing)
# -%1 = cpp-command.c %2 = TMPDIR/FILE.i %args =  -I./share/libc -D__FRAMAC__ -D__FC_MACHDEP_X86_32 -dD -nostdinc -m32
# +%1 = cpp-command.c %2 = TMPDIR/FILE.i %args =  -I./share/libc -D__FRAMAC__ -D__FC_MACHDEP_X86_32 -dD -nostdinc
# Ocamlc       tests/syntax/ast_init.cmo
# Ocamlopt     tests/syntax/ast_init.cmx
# Env:
# FRAMAC_LIB (set from outside) = "/home/opam/.opam/4.11/.opam-switch/build/frama-c.22.0/lib/fc"
# FRAMAC_PLUGIN_GUI = "./lib/plugins/gui"
# FRAMAC_PLUGIN = "./lib/plugins"
# FRAMAC_SHARE = "./share"
# FRAMAC_SESSION = "."
# OCAMLRUNPARAM = ""
# Command:
# bin/toplevel.opt -check tests/syntax/cpp-command.c   -no-autoload-plugins -print-cpp-commands  2>/tmp/cpp-command.47deb55.err.log | /bin/sed "s:/[^ ]*/cpp-command\.[^ ]*\.i:TMPDIR/FILE.i:g; s:$PWD/::" >tests/syntax/result/cpp-command.4.res.log && /bin/sed "s:/[^ ]*/cpp-command\.[^ ]*\.i:TMPDIR/FILE.i:g; s:$PWD/::" < /tmp/cpp-command.47deb55.err.log >tests/syntax/result/cpp-command.4.err.log && rm -f /tmp/cpp-command.47deb55.err.log
# --- tests/syntax/oracle/cpp-command.4.res.oracle	2020-11-20 00:00:00.000000000 +0000
# +++ tests/syntax/result/cpp-command.4.res.log	2020-11-20 23:31:38.107250907 +0000
# @@ -1,2 +1,2 @@
#  [kernel] Preprocessing command:
# -  gcc -E -C -I.  -I./share/libc -D__FRAMAC__ -D__FC_MACHDEP_X86_32 -dD -nostdinc -m32 'tests/syntax/cpp-command.c' -o 'TMPDIR/FILE.i'
# +  gcc -E -C -I.  -I./share/libc -D__FRAMAC__ -D__FC_MACHDEP_X86_32 -dD -nostdinc 'tests/syntax/cpp-command.c' -o 'TMPDIR/FILE.i'
```
Full log: https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/c850fa9cde508ba7ee69143d5580b2e23f66498f/variant/extras,arm64,frama-c.22.0,tests
* Similar diff issues in the tests on PPC64. Full log: https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/c850fa9cde508ba7ee69143d5580b2e23f66498f/variant/extras,ppc64,frama-c.22.0,tests
* Issue while testing on Fedora:
```
# Error creating flamegraph, aborting.
# Can't locate open.pm in @INC (you may need to install the open module) (@INC contains: /usr/local/lib64/perl5/5.30 /usr/local/share/perl5/5.30 /usr/lib64/perl5/vendor_perl /usr/share/perl5/vendor_perl /usr/lib64/perl5 /usr/share/perl5) at /home/opam/.opam/4.11/.opam-switch/build/frama-c.22.0/bin/../share/analysis-scripts/flamegraph.pl line 109.
# BEGIN failed--compilation aborted at /home/opam/.opam/4.11/.opam-switch/build/frama-c.22.0/bin/../share/analysis-scripts/flamegraph.pl line 109.
```
It seems that `open.pm` is not distributed in Fedora since Fedora 27.
Full log: https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/c850fa9cde508ba7ee69143d5580b2e23f66498f/variant/distributions,fedora-32,frama-c.22.0,tests

AllanBlanchard:
> @AllanBlanchard you can ignore the issues with libgnomecanvas being unavailable for now. However I encourage you to find alternatives because most distributions are migrating away from gtk2. For instance debian-testing – in line for the next release of Debian – doesn't have gnomecanvas.

We just select GTK2 by default as all GUI features are supported with these packages. If some of the lablgtk2 packages aren't availalble, the opam package (and the Frama-C configure) are supposed to select GTK3 and disable a minor feature related to Gnome Canvas.

We will investigate the other issues.

maroneze:
The first test that fails is indeed platform-dependent, so we'll add a small patch to change the oracle (the test works fine, it's just that our oracle is a bit too verbose and spits some non-essential architecture-dependent information). For the second test, indeed there is a separate `perl-open` package in Fedora, but it is not even checked by `conf-perl`, and we never asked for the dependence anyway. So the best solution is to disable the test (it's a very minor optional feature, we don't want to create a new `conf-perl-open` package and add all the necessary checks for this test only).

Would it be ok if we add two `patches` to the `opam` directory for this release (since the issues are only related to minor test configurations and not bugs in the code itself)? We did it in the past, but I don't know if we can still operate that way.

kit-ty-kate:
> Would it be ok if we add two `patches` to the `opam` directory for this release (since the issues are only related to minor test configurations and not bugs in the code itself)? We did it in the past, but I don't know if we can still operate that way.

Sure, no problem.

It's still failing though, is that ok if I simply disable the tests on all architectures but `x86_64`? Since it seems to be only in the tests and not the software itself it seems safe to do so for now. I will not delay this release any further.
If these issues can be fixed for next release though that'd be perfect ^_^

maroneze:
Ideally, keeping the tests on arm64 would be useful, so there's another patch I tried, but I agree that for now disabling the tests would be the fastest solution.

On ppc64, we'll disable the tests since we cannot easily reproduce them anyway.

Anyway, for now I think your patch will be fine, thanks! We hope the next release will go smoother.

AllanBlanchard:
Thanks a lot :)

kit-ty-kate:
Thanks!

maroneze:
By the way, do you know if a simple Raspberry Pi 4 would be enough to locally test arm64? We'd like to be able to perform such tests ourselves without having to use opam's CI infrastructure, and if we could avoid buying a new Mac for that, it would be great...

kit-ty-kate:
> By the way, do you know if a simple Raspberry Pi 4 would be enough to locally test arm64? We'd like to be able to perform such tests ourselves without having to use opam's CI infrastructure, and if we could avoid buying a new Mac for that, it would be great...

A Raspberry Pi 4 would be more than enough I think. The new ARM mac would also be good, although automated CI in those is more tricky to setup due to the lack of macOS docker images. But if it's a one-off thing, sure.
Ideally having both would be perfect since they both (linux and macOS) have different environments (e.g. gcc vs. clang)

PR #17676, opened 2020-11-21
conf-gnomecanvas: Add support for alpine and ubuntu/arch-derived distributions

kit-ty-kate:
Issue detected in https://github.com/ocaml/opam-repository/pull/17670


kit-ty-kate:
The alpine is missing a dependency on x86_64. Ignoring for now.. (hopefully it should be fixed on their side soon)

PR #17687, opened 2020-11-23
Fix drom.0.1.0 checksum

lefessan:
There was a modification of the release commit, so it changed the checksum.

### DIFF
--- a/packages/drom/drom.0.2.0/opam
+++ b/packages/drom/drom.0.2.0/opam
@@ -2,7 +2,7 @@
 # Do not modify or add to the `skip` field of `drom.toml`.
 opam-version: "2.0"
 name: "drom"
-version: "0.1.0"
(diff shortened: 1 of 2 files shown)


avsm:
Could you please rename this package to drom.0.1.0-1 then, and remove the previous? We're trying hard to make the released packages immutable from an archive perspective.

lefessan:
I directly switched to 0.2.0, and removed the former (incorrect) package.

lefessan:
The CI failed because of download problems ?

avsm:
Thanks @lefessan; it looks like it was a transient GitHub problem (they returned a 404 on the archive), so I've restarted the affected jobs.

lefessan:
Thanks ! All the lights are green now :-)

kit-ty-kate:
Thanks!

PR #17692, opened 2020-11-24
Fix revdeps for the next release of lwt_ppx (switch to ppxlib)

kit-ty-kate:
cc @aantron @raphael-proust, you'll be happy to hear that apart from the minor failures on these packages everything ran smoothly with the new lwt_ppx (using ppxlib), no regression detected :tada:


aantron:
Thank you!

PR #17697, opened 2020-11-24
Fix ocaml-variants with ocaml-option-bytecode-only enabled

kit-ty-kate:
Current error otherwise:
```
#=== ERROR while compiling ocaml-variants.4.12.0+trunk ========================#
# context              2.1.0~beta3 | linux/x86_64 |  | file:///home/opam/opam-repository
# path                 ~/.opam/4.12.0+trunk/.opam-switch/build/ocaml-variants.4.12.0+trunk
# command              /usr/bin/make -j47 world.opt
# exit-code            2
# env-file             ~/.opam/log/ocaml-variants-7-9fcb43.env
# output-file          ~/.opam/log/ocaml-variants-7-9fcb43.out
### output ###
# Makefile:920: *** The native-code compiler is not supported on this platform.  Stop.
```

### DIFF
--- a/packages/ocaml-variants/ocaml-variants.4.12.0+trunk/opam
+++ b/packages/ocaml-variants/ocaml-variants.4.12.0+trunk/opam
@@ -47,7 +47,7 @@ build: [
     "--disable-warn-error"
   ]
   [make "-j%{jobs}%" {os != "cygwin"} "world"]
-  [make "-j%{jobs}%" {os != "cygwin"} "world.opt"]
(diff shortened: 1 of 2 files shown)


dra27:
Oops - yes, this is obviously correct. We can go further with 4.12 and just drop the two commands - since either 4.10 or 4.11 you only need to do `make` and it will run `world` or `world.opt` as appropriate. That will be quicker for two reasons:
1. The stdlib manpages will be compiled with `ocamldoc.opt`
2. `world.opt` won't waste a teeny bit of time rescanning directories which have already been built

unless there's a deeper reason for building `world` and then `world.opt`?

PR #17616, opened 2020-11-13
Package llvm.11.0.0

kit-ty-kate:
### `llvm.11.0.0`
The OCaml bindings distributed with LLVM
Note: LLVM should be installed first.

---
* Homepage: http://llvm.moe
* Source repo: git+https://github.com/kit-ty-kate/llvm-dune.git
* Bug tracker: http://llvm.org/bugs/

---
:camel: Pull-request generated by opam-publish v2.0.2

PR #17703, opened 2020-11-25
[new release] conf-libclang.11

thierry-martinez:
This PR adds support for detecting libclang 11.

### DIFF
--- a/packages/conf-libclang/conf-libclang.11/files/configure.sh
+++ b/packages/conf-libclang/conf-libclang.11/files/configure.sh
@@ -1,0 +1,79 @@
+#!/bin/bash -ex
+
+find_llvm_config () {
+    # Locate llvm-config (taken from conf-llvm's configure.sh file)
+
+    shopt -s nullglob
+    for version in 11 10 9 8 7 6 5 4 3; do
+        if hash brew 2>/dev/null; then
+           brew_llvm_config="$(brew --cellar)"/llvm*/${version}*/bin/llvm-config
+        fi
+        for llvm_config in \
+            llvm-config-${version} llvm-config-${version}.0 \
+            llvm-config${version}0 llvm-config${version} \
+            llvm-config-mp-$version \
+            llvm-config-mp-${version}.0 $brew_llvm_config \
+            /usr/lib64/llvm/${version}/bin/llvm-config \
+            /usr/lib/llvm/${version}/bin/llvm-config \
+            llvm-config; do
+            llvm_version="`$llvm_config --version`" || continue
+            return
+        done
+    done
+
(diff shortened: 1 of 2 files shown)


packages/conf-libclang/conf-libclang.11/files/configure.sh
```diff
@@ -0,0 +1,79 @@
...
+        fi
+        for llvm_config in \
+            llvm-config-${version} llvm-config-${version}.0 \
+            llvm-config${version}0 llvm-config${version} \
+            llvm-config-mp-$version \
+            llvm-config-mp-${version}.0 $brew_llvm_config \
+            /usr/lib64/llvm/${version}/bin/llvm-config \
+            /usr/lib/llvm/${version}/bin/llvm-config \
+            llvm-config; do
+            llvm_version="`$llvm_config --version`" || continue
```
kit-ty-kate, line 19: it is to be noted that on fedora, centos and similar distributions, `llvm-config` is named `llvm-config-64` (on 64bits arch) and `llvm-config-32` (on 32bits arch), see https://github.com/ocaml/opam-repository/pull/17616/files#diff-ec9c8838b3b6ff2cbb890f64386e483d40736ad8040844ce98c71a949930fc3fR10

kit-ty-kate:
Sorry I haven't had time to look at this PR, I don't think there are any blockers, though I have a question. Is there an opportunity of merging `conf-llvm` and `conf-libclang`?

thierry-martinez:
Thank you very much for the review, @kit-ty-kate ! I see two differences between `conf-llvm` and `conf-libclang` that forces the distinction:

- on most distributions, clang is packaged is an additional package, in addition to LLVM, and opam packages that only depend on `conf-llvm`do not have to require the clang package to be installed;

- `conf-llvm` and `conf-libclang` have different semantics with respect to versioning: `conf-llvm` looks for a particular version of libllvm, whereas `conf-libclang` seeks for any version of llvm/clang below a given upper bound. This behavior of `conf-libclang` is linked to the fact that the `clangml` library supports any version of llvm/clang. I admit that it is not very clean, but I don't see any other mean to express to opam the fact that `clangml` package is installable as soon as any particular version of `conf-llvm` is installable, automatically (I believe that opam will just try to install the last version of `conf-llvm`, and possibly fail if it is not available, even if a lower version would have been installable).

packages/conf-libclang/conf-libclang.11/files/configure.sh
thierry-martinez: Thanks for the suggestion! Fixed in 19972619bff15e2eb97ae4767e5cb1fe6baa3ac6

PR #17704, opened 2020-11-25
Package gmp.6.2.1

TheLortex:
### `gmp.6.2.1
The GNU Multiple Precision Arithmetic Library
Dune packaging of the GMP library, suitable for 
cross-compilation.

---
* Homepage: https://github.com/TheLortex/ocaml-gmp
* Source repo: git+https://github.com/TheLortex/ocaml-gmp.git
* Bug tracker: https://github.com/TheLortex/ocaml-gmp/issues

---

### DIFF
--- a/packages/gmp/gmp.6.2.0/opam
+++ b/packages/gmp/gmp.6.2.0/opam
@@ -1,0 +1,29 @@
+opam-version: "2.0"
+maintainer: "Lucas Pluvinage <lucas@tarides.com>"
+license:      ["LGPL-3.0-only" "LGPL-2.0-only"]
+authors: "Torbjörn Granlund and contributors"
+homepage: "https://github.com/TheLortex/ocaml-gmp"
+bug-reports: "https://github.com/TheLortex/ocaml-gmp/issues"
+dev-repo: "git+https://github.com/TheLortex/ocaml-gmp.git"
+build: [
+ [ "dune" "build" "-p" name "-j" jobs ]
+ [ "dune" "runtest" {with-test} ]
+]
+depends: [
+  "ocaml" {>="4.05.0"}
+  "dune" {>="2.6"}
+]
+synopsis: "The GNU Multiple Precision Arithmetic Library"
+description: """Dune packaging of the GMP library, suitable for 
+cross-compilation."""
+extra-source "src/gmp-6.2.0.tar.xz" {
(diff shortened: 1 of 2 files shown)


hannesm:
thanks for your work on this package -- as I understand this will be a dependency of MirageOS4 (replacing gmp-freestanding) -- I'd be in favor to have it under the mirage organization, would this suit you well?

TheLortex:
Yes sure, it's a better idea. I'll transfer the repository and update the PR.

TheLortex:
Seems good to go !

I'm not sure why that error occurs on 4.12 though.

mseri:
I agree. @kit-ty-kate do you know why it has that failure with 4.12?

kit-ty-kate:
That's a dune issue which does not currently support OCaml 4.12 fully: https://github.com/ocaml/dune/issues/3766 / https://github.com/ocaml/dune/pull/3793 / https://github.com/ocaml/dune/pull/3904

I think it can be ignored for now. If there is an issue I'll see it sooner or later and will fix it before the 4.12 release.

kit-ty-kate:
I've made a few fixes to the opam file, could you return those fixes upstream?
I've also noticed the underlying `build.sh` script uses `make -j8`, I have 2 minor issues with this:
* we can't test on freebsd anymore, are you sure this is compatible with bmake?
* is it possible to pass the number of jobs given by opam to dune instead of the `8` magic number? For instance if someone would like to avoid their computer to take 200% of their CPU and would like to do something while opam is doing its thing by passing `opam install -j1` it would be a bit sad to suddenly have that happen in the middle. Or on the contrary if we have 100+ cores available (in CI for instance) it's a bit sad to not use them and make the build faster.

What is the state of this PR otherwise? The 4.12 failure is not a blocker (the minor issues above might be though) but PR is still a draft.

TheLortex:
Hi, it should be good now. 
I think @hannesm tested it on FreeBSD. And I changed the build script to take the opam `jobs` variable into account. 
I also upgraded GMP in the meantime so it's `6.2.1` instead of `6.2.0`.

hannesm:
indeed I ensured (with https://github.com/mirage/ocaml-gmp/pull/1 -- merged and part of the above 6.2.1 release) that it works on FreeBSD.

kit-ty-kate:
thanks a lot!

PR #17705, opened 2020-11-25
[new release] charrua-unix, charrua, charrua-server and charrua-client (1.3.0)

hannesm:
CHANGES:
    
* Revise packaging: charrua-client-lwt and charrua-client-mirage are now available as charrua-client.lwt and charrua-client.mirage (mirage/charrua#110 @hannesm)
* Dhcp_ipv4 directly uses Dhcp_client_mirage (instead of an abstract module interface being passed) (mirage/charrua#110 @hannesm)
* Fix sending of client_identifier with appropriate type as sent by the client (mirage/charrua#98 @hannesm, reported in mirage/charrua#84 by @lynxis)


hannesm:
the failures are CI running out of disk space as far as I can tell...

mseri:
Seems to be fine now

PR #17716, opened 2020-11-26
Add support for OCaml 4.10.1 on macOS/arm64

avsm:
This adds the backported compiler from https://github.com/ocaml/ocaml/pull/10026 as a ocaml-variants.4.10.1+macos-arm64 package.

The PR also adds constraints to all the other OCaml base-compiler and variants packages pre-4.12 to restrict them from being selected on macos/arm64, so that the default compiler will be 4.10.1 (and 4.12.0~alpha1 if the beta repository is active).

Still testing this, just got my M1 delivered...

### DIFF
--- a/packages/conf-gmp-powm-sec/conf-gmp-powm-sec.3/files/test.c
+++ b/packages/conf-gmp-powm-sec/conf-gmp-powm-sec.3/files/test.c
@@ -1,0 +1,26 @@
+#include <gmp.h>
+#ifndef __GMP_H__
+#error "No GMP header"
+#endif
+#if __GNU_MP_VERSION < 5
+#error "GMP >= 5 is required to support mpz_powm_sec"
+#endif
+
+void test(void) {
+  mpz_t base;
+  mpz_t exp;
+  mpz_t mod;
+  mpz_t rop;
+
+  mpz_init_set_ui(base, 2u);
+  mpz_init_set_ui(exp, 4u);
+  mpz_init_set_ui(mod, 3u);
+  mpz_init(rop);
+
+  mpz_powm_sec(rop, base, exp, mod);
+
+  mpz_clear(base);
+  mpz_clear(exp);
+  mpz_clear(mod);
+  mpz_clear(rop);
+}
--- a/packages/conf-gmp-powm-sec/conf-gmp-powm-sec.3/opam
+++ b/packages/conf-gmp-powm-sec/conf-gmp-powm-sec.3/opam
@@ -1,0 +1,22 @@
+opam-version: "2.0"
+maintainer: "Etienne Millon <etienne@cryptosense.com>"
+homepage: "http://gmplib.org/"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: "GPL-1.0-or-later"
+build: [
+  ["sh" "-exc" "cc -c $CFLAGS -I/usr/local/include test.c"] {os != "macos"}
+  [
+    "sh"
+    "-exc"
+    "cc -c $CFLAGS -I/opt/homebrew/include -I/opt/local/include -I/usr/local/include test.c"
+  ] {os = "macos"}
+]
+depends: ["conf-gmp"]
+synopsis:
+  "Virtual package relying on a GMP lib with constant-time modular exponentiation"
+description: """
+This package can only install if the GMP lib is installed on the system and
+corresponds to a version that has the mpz_powm_sec function."""
+authors: "Etienne Millon <etienne@cryptosense.com>"
+extra-files: ["test.c" "md5=29317f477fa828e18428660ef31064fb"]
+flags: conf
--- a/packages/conf-gmp/conf-gmp.3/files/test.c
+++ b/packages/conf-gmp/conf-gmp.3/files/test.c
@@ -1,0 +1,10 @@
+#include <gmp.h>
+#ifndef __GMP_H__
+#error "No GMP header"
+#endif
+
+void test(void) {
+  mpz_t n;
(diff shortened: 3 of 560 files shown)


packages/ocaml-base-compiler/ocaml-base-compiler.4.10.1+macos-arm64/opam
```diff
@@ -0,0 +1,42 @@
...
+    "-C"
+    "CC=cc" {os = "openbsd" | os = "freebsd" | os = "macos"}
+    "ASPP=cc -c" {os = "openbsd" | os = "freebsd" | os = "macos"}
+  ]
+  [make "-j%{jobs}%" {os != "cygwin"} "world"]
+  [make "-j%{jobs}%" {os != "cygwin"} "world.opt"]
+]
+install: [make "install"]
+url {
+  src: "git+https://github.com/avsm/ocaml.git#4.10.1+macos-arm64"
```
mseri, line 30: There is no corresponding path on ocaml’s main compiler branch?

packages/ocaml-base-compiler/ocaml-base-compiler.4.10.1+macos-arm64/opam
```diff
@@ -0,0 +1,42 @@
...
+}
+extra-files: ["ocaml-base-compiler.install" "md5=3e969b841df1f51ca448e6e6295cb451"]
+post-messages: [
+  "A failure in the middle of the build may be caused by build parallelism
+   (enabled by default).
+   Please file a bug report at https://github.com/ocaml/ocaml/issues"
```
mseri, line 36: Is this still the right place for issues if we point at the branch on your fork?

packages/ocaml-base-compiler/ocaml-base-compiler.4.10.1+macos-arm64/opam
mseri: Should we also add the md5/sha?

packages/ocaml-base-compiler/ocaml-base-compiler.4.10.1+macos-arm64/opam
kit-ty-kate: This just sits on top of the 4.10 branch, I'm pretty sure this is fine to point to this.

packages/ocaml-base-compiler/ocaml-base-compiler.4.10.1+macos-arm64/opam
kit-ty-kate: I believe this is the same sort of issue as with https://github.com/ocaml/opam-repository/pull/16357

packages/ocaml/ocaml.4.10.1/opam
```diff
@@ -6,7 +6,7 @@ and polls it to initialise specific variables like `ocaml:native-dynlink`"""
 maintainer: "platform@lists.ocaml.org"
 depends: [
   "ocaml-config"
-  "ocaml-base-compiler" {= "4.10.1"} |
+  "ocaml-base-compiler" {= "4.10.1" | = "4.10.1+macos-arm64"} |
```
kit-ty-kate, line 9: Could it be moved into `ocaml-variants`, to my knowledge the distinction between `ocaml-base-compiler` and `ocaml-variants` is that the first only gathers the official release. This is not an official release so I would feel better not having it in `ocaml-base-compiler`. Does this would also get rid of this hack.

packages/ocaml-base-compiler/ocaml-base-compiler.4.10.1+macos-arm64/opam
avsm: Yeah, git remote is fine for this sort of "temporary" package (since 4.12 will supplant it with an official release). Good point about moving it to ocaml/ though rather than my fork -- I'll talk to @Octachron about that.

packages/ocaml/ocaml.4.10.1/opam
avsm: I had an odd problem with `opam init` not working with ocaml-variants for some reason. I'll dig into it more later today.

avsm:
Just waiting on a decision upstream and then will make all the changes requested.

packages/ocaml/ocaml.4.10.1/opam
avsm: This does need to be in base-compiler since `opam init` specifically tries to initialise ocaml-system or ocaml-base-compiler (and not ocaml-variants).  This is close enough to an official release due to the unusual situation of having a new tier1 architecture show up, so I'll wait on @Octachron to take a decision on where the branch should live and then unblock macOS/ARM users.

kit-ty-kate: That seems really surprising to me. To my knowledge there is nothing in opam code that favours `opam-base-compiler` over `ocaml-variants`. Are you sure you didn't forget to remove the `ocaml-beta` dependency?

avsm: https://github.com/ocaml/opam/blob/2.0.7/src/client/opamInitDefaults.ml#L26

kit-ty-kate: oh I see. Sorry I missed that.

kit-ty-kate:
Feel free to merge whenever you're ready.

avsm:
@Octachron is going to release 4.10.2, so I'll hold off merging this until that release is available (which avoids the base-compiler change in this PR as well).

PR #17724, opened 2020-11-27
[new release] reparse and reparse-unix (2.0.0)

bikallem:
Recursive descent parsing library for ocaml

- Project page: <a href="https://github.com/lemaetech/reparse">https://github.com/lemaetech/reparse</a>

##### CHANGES:

- Rewrite the whole package to use exceptions rather than `result` type
- Adds many more parsing combinators
- Adds comprehensive unit tests
- Adds comprehensive documentation, host documentation and add links in repo home page
- Adds abstraction for input source
- Provides unix file source and string input source
- Adds separate package `reparse-unix` for unix file input
- Adds calc.ml and json.ml in examples.

### DIFF
--- a/packages/reparse-unix/reparse-unix.2.0.0/opam
+++ b/packages/reparse-unix/reparse-unix.2.0.0/opam
@@ -1,0 +1,42 @@
+opam-version: "2.0"
+synopsis:
+  "Provides support for parsing files as source of input for reparse library "
+description:
+  "Provides support for parsing files as source of input for reparse library "
+maintainer: ["Bikal Lem"]
+authors: ["Bikal Lem <gbikal@gmail.com>"]
+license: "MPL-2.0"
+homepage: "https://github.com/lemaetech/reparse"
+bug-reports: "https://github.com/lemaetech/reparse/issues"
+depends: [
+  "dune" {>= "2.7"}
+  "ocaml" {>= "4.10.0"}
+  "reparse" {= "2.0.0"}
+  "base-unix"
+  "alcotest" {with-test}
+  "odoc" {with-doc}
+]
+build: [
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
+dev-repo: "git+https://github.com/lemaetech/reparse.git"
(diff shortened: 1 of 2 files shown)


packages/reparse-unix/reparse-unix.2.0.0/opam
```diff
@@ -0,0 +1,41 @@
...
+  "Provides support for parsing files as source of input for reparse library "
+maintainer: ["Bikal Lem"]
+authors: ["Bikal Lem <gbikal@gmail.com>"]
+license: "MPL-2.0"
+homepage: "https://github.com/lemaetech/reparse"
+bug-reports: "https://github.com/lemaetech/reparse/issues"
+depends: [
+  "dune" {>= "2.7"}
+  "ocaml" {>= "4.10.0"}
+  "reparse"
```
mseri, line 14:
```suggestion
  "reparse" {=version}
```
We tend to strictly tie packages that are released from the same repository. This usually simplifies tagging releases and prevents issues when new releases are published with api changes

packages/reparse-unix/reparse-unix.2.0.0/opam
```diff
@@ -0,0 +1,41 @@
...
+  "base-unix"
+  "odoc" {with-doc}
+]
```
mseri, line 17:
```suggestion
  "alcotest" {with-test}
]
```

packages/reparse/reparse.2.0.0/opam
```diff
@@ -0,0 +1,39 @@
...
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
```
mseri, line 26: Some tests for this package require `release-unix`. This would create a cyclic dependency, one solution could be to make these also tests in reparse-unix. Otherwise we can disable them, in the future opam 2.1 will be able to deal with this kind of cycles

mseri (changes requested):

packages/reparse/reparse.2.0.0/opam
bikallem: Could we disable the test, please. Thanks. The opam file is generated by dune. How do I disable tests in dune-project?

bikallem: Done.

PR #17729, opened 2020-11-28
Package gobba.0.4.2

0x0f0f0f:
### `gobba.0.4.2`
A simple, didactical, purely functional programming language
A simple, didactical, purely functional programming language written for the programming 2 course at the University of Pisa, extended with a simple parser made with Menhir and ocamllex

---
* Homepage: https://github.com/0x0f0f0f/gobba
* Source repo: git+https://github.com/0x0f0f0f/gobba.git
* Bug tracker: https://github.com/0x0f0f0f/gobba/issues

---

### DIFF
--- a/packages/gobba/gobba.0.4.2/opam
+++ b/packages/gobba/gobba.0.4.2/opam
@@ -10,12 +10,10 @@ description: "A simple, didactical, purely functional programming language writt
 doc: "https://0x0f0f0f.github.io/gobba"
 
 build: [
-  ["dune" "subst"] {pinned}
+  ["dune" "subst"] {dev}
   ["dune" "build" "-p" name "-j" jobs]
   ["sh" "-c" "GOBBA_EXAMPLES=$(realpath ./examples/) dune runtest -p %{name}% -j %{jobs}%"] {with-test}
 ]
-
-conflicts: "gobba" {!= "transition"}
 
 depends: [
     "dune" {>= "2.0"}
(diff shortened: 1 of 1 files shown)


packages/gobba/gobba.0.4.2/opam
```diff
@@ -0,0 +1,40 @@
...
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["sh" "-c" "GOBBA_EXAMPLES=$(realpath ./examples/) dune runtest -p %{name}% -j %{jobs}%"] {with-test}
+]
+
+conflicts: "gobba" {!= "transition"}
+
+depends: [
+    "dune" {>= "2.0"}
+    "ocaml" {>= "4.05.0"}
```
mseri, line 22:
```suggestion
    "ocaml" {>= "4.07.0"}
```
`Set.add_seq` was added on ocaml 4.07.0

mseri: @0x0f0f0f Can you make this change also upstream?

mseri:
Has the package changed upstream? I see the following failure
```
https://github.com/0x0f0f0f/gobba/archive/0.4.2.tar.gz (Bad checksum, expected md5=10ba85585b0be85888b38a4eb88918cb)
``

mseri:
I think to use bisect only as test dependency with dune, one needs to use the new bisect support in the latest dune, otherwise bisect needs to be there on release (or you need an autogenerated file hack that I would not recommend)

mseri:
Looks like this can be merged now (there are some issues on 32 bit systems and PPC, maybe we should just mark them not supported). @kit-ty-kate what do you think?

kit-ty-kate:
@mseri we could. However we haven't heard back from the the author for a month and a half so maybe they want to fix that first. cc @0x0f0f0f ?

For `bisect_ppx` that changed in recent versions so it could be done differently. See: https://github.com/aantron/bisect_ppx/pull/331

0x0f0f0f:
Thanks for the fixes. I have not worked on gobba recently. Are the fails reported by opam-ci negligible? I remember from a few months ago it used to build correctly on 32 bit and 64 bit ARM (manually tried on a Pi). Nothing has changed much in gobba since then. Probably changes in owl.

mseri:
Indeed, looks like something weird has happened to owl flags? I will open a bug report on owl repo and merge this.

PR #17744, opened 2020-12-02
[new release] decimal (0.2.0)

yawaramin:
Arbitrary-precision floating-point decimal library

- Project page: <a href="https://github.com/yawaramin/ocaml-decimal">https://github.com/yawaramin/ocaml-decimal</a>
- Documentation: <a href="https://yawaramin.github.io/ocaml-decimal/api">https://yawaramin.github.io/ocaml-decimal/api</a>

### DIFF
--- a/packages/decimal/decimal.0.2.0/opam
+++ b/packages/decimal/decimal.0.2.0/opam
@@ -1,0 +1,42 @@
+opam-version: "2.0"
+synopsis: "Arbitrary-precision floating-point decimal library"
+description: """
+Arbitrary-precision floating-point decimal library ported from
+the Python decimal module."""
+maintainer: ["Yawar Amin <yawar.amin@gmail.com>"]
+authors: ["Yawar Amin <yawar.amin@gmail.com>"]
+license: "PSF-2.0"
+homepage: "https://github.com/yawaramin/ocaml-decimal"
+doc: "https://yawaramin.github.io/ocaml-decimal/api"
+bug-reports: "https://github.com/yawaramin/ocaml-decimal/issues"
+depends: [
+  "dune" {>= "2.7"}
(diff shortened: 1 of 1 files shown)


mseri:
Tests for long integers are failing on 32 bits systems. Is that expected?

mseri:
```
# FAIL: 1E-958594760
# context: { prec = 9; round = half_up; e_max = 999999999; e_min = -999999999; capitals = true; clamp = false; traps = { clamped = false; invalid_operation = false; conversion_syntax = false; div_by_zero = false; div_impossible = false; div_undefined = false; inexact = false; rounded = false; subnormal = false; overflow = false; underflow = false }; flags = { clamped = false; invalid_operation = false; conversion_syntax = false; div_by_zero = false; div_impossible = false; div_undefined = false; inexact = false; rounded = false; subnormal = false; overflow = false; underflow = false } }
# 
# Fatal error: exception Assert_failure("test/decimal_test.ml", 58, 6)
```

yawaramin:
Oh–looks like an overflow. I'll look into it–thanks!

yawaramin:
Hi @mseri --it is indeed an integer overflow issue. Unfortunately I don't have the resources to test/develop/support 32-bit systems. I'll add a note that makes it clear that only 64-bit is supported in the opam file.

mseri:
You could also disable it there if you think it is better: like in https://github.com/ocaml/opam-repository/blob/a78d7fa13785cc738a8dab82014f49ec6040608c/packages/mirage-entropy/mirage-entropy.0.4.1/opam#L36 but with `!=`
(using https://opam.ocaml.org/doc/Manual.html#opamvar-arch )

yawaramin:
Thanks, that's a good suggestion. However I'm not sure how it would get generated from the `dune-project` file. There's no documentation about that.

PR #17742, opened 2020-12-01
[new release] albatross (1.0.0)

hannesm:
Albatross - orchestrate and manage MirageOS unikernels with Solo5

- Project page: <a href="https://github.com/roburio/albatross">https://github.com/roburio/albatross</a>

##### CHANGES:

Initial release

PR #17745, opened 2020-12-02
Add conf-libnl3 virtual package

reynir:
In debian the libnl libraries are packaged in a handful of different packages while on centos family they seem to be all packaged in one package. This depends on both libnl-3 and libnl-route-3 since that's what we need in #17742.


reynir:
Alright, I believe I found all the OS packages that has libnl3, at least the ones in CI. In the case of Oracle Linux you need to add the optional repository (`ol7_optional_latest`). I don't know where to report that.

hannesm:
thanks @reynir -- I agree this is good to be merged :)

kit-ty-kate:
Thanks a lot!

PR #17749, opened 2020-12-02
[new release] mirage, mirage-runtime, mirage-types-lwt and mirage-types (3.10.0)

hannesm:
CHANGES:

IPv6 and dual (IPv4 and IPv6) stack support mirage/mirage#1187

Since a long time, IPv6 code was around in our TCP/IP stack (thanks to @nojb
who developed it in 2014). Some months ago, @hannesm and @MagnusS got excited
to use it. After we managed to fix some bugs and add some test cases, and
writing more code to setup IPv6-only and dual stacks, we are eager to share
this support for MirageOS in a released version. We expect there to be bugs
lingering around, but duplicate address detection (neighbour solicitation and
advertisements) has been implemented, and (unless
"--accept-router-advertisement=false") router advertisements are decoded and
used to configure the IPv6 part of the stack. Configuring a static IPv6 address
is also possible (with "--ipv6=2001::42/64").

While at it, we unified the boot arguments between the different targets:
namely, on Unix (when using the socket stack), you can now pass
"--ipv4=127.0.0.1/24" to the same effect as the direct stack: only listen
on 127.0.0.1 (the subnet mask is ignored for the Unix socket stack).

A dual stack unikernel has "--ipv4-only=BOOL" and "--ipv6-only=BOOL" parameters,
so a unikernel binary could support both Internet Protocol versions, while the
operator can decide which protocol version to use.

Please also note that the default IPv4 network configuration no longer uses
10.0.0.1 as default gateway (since there was no way to unset the default
gateway mirage/mirage#1147).

For unikernel developers, there are some API changes in the Mirage module
- New "v4v6" types for IP protocols and stacks
- The ipv6_config record was adjusted in the same fashion as the ipv4_config
  type: it is now a record of a network (V6.Prefix.t) and gateway (V6.t option)

Some parts of the Mirage_key module were unified as well:
- Arp.ip_address is available (for a dual Ipaddr.t)
- Arg.ipv6_address replaces Arg.ipv6 (for an Ipaddr.V6.t)
- Arg.ipv6 replaces Arg.ipv6_prefix (for a Ipaddr.V6.Prefix.t)
- V6.network and V6.gateway are available, mirroring the V4 submodule

If you're ready to experiment with the dual stack, here's a diff for our basic
network example (from mirage-skeleton/device-usage/network) replacing IPv4
with a dual stack:

```
diff --git a/device-usage/network/config.ml b/device-usage/network/config.ml

### DIFF
--- a/packages/capnp-rpc-mirage/capnp-rpc-mirage.0.8.0/opam
+++ b/packages/capnp-rpc-mirage/capnp-rpc-mirage.0.8.0/opam
@@ -22,7 +22,7 @@ depends: [
   "arp-mirage" {with-test}
   "alcotest-lwt" {>= "1.0.1" & with-test}
   "io-page-unix" {with-test}
-  "tcpip" {>= "5.0.0" & with-test}
+  "tcpip" {>= "5.0.0" & < "6.0.0" & with-test}
   "mirage-vnetif" {with-test}
   "dune" {>= "2.0"}
 ]
--- a/packages/mirage-runtime/mirage-runtime.3.10.0/opam
+++ b/packages/mirage-runtime/mirage-runtime.3.10.0/opam
@@ -1,0 +1,39 @@
+opam-version: "2.0"
+maintainer:   ["anil@recoil.org" "thomas@gazagnaire.org"]
+authors:      ["Thomas Gazagnaire" "Anil Madhavapeddy" "Gabriel Radanne"
+               "Mindy Preston" "Thomas Leonard" "Nicolas Ojeda Bar"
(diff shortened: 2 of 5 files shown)


mseri:
Failures seems fine, can you have a look as well before we merge?

hannesm:
thanks @mseri, I looked through the failures and fixed the capnp-rpc-mirage testrun (submitted a fix upstream, in this PR an upper bound is introduced): the issue was breaking changes in tcpip 6.0.0. this PR is ready to be merged.

mseri:
Thanks!

PR #17755, opened 2020-12-04
Package ladspa.0.2.0

toots:
### `ladspa.0.2.0`
Bindings for the LADSPA API which provides audio effects

---
* Homepage: https://github.com/savonet/ocaml-ladspa
* Source repo: git+https://github.com/savonet/ocaml-ladspa.git
* Bug tracker: https://github.com/savonet/ocaml-ladspa/issues

---

### DIFF
--- a/packages/conf-ladspa/conf-ladspa.1/opam
+++ b/packages/conf-ladspa/conf-ladspa.1/opam
@@ -1,0 +1,17 @@
+opam-version: "2.0"
+maintainer: "https://github.com/ocaml/opam-repository/issues"
+homepage: "https://www.ladspa.org/"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+authors: "ladspa dev team"
+license: "BSD"
+depexts: [
+  ["ladspa-dev"] {os-distribution = "alpine"}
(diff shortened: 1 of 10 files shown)


toots:
This one is a bit tricky: LADSPA is basically just a header so I'm not sure how to compile `conf-ladspa`. Couple of notes:
* We could embed the header but it is versioned so it's better not to.
* We could detect manually the header but that implies a lot of assumption regarding the compiler
* We could re-implement the `dune-configurator` detection in `conf-ladspa`
* We could just do nothing

`ladspa` is kind of a niche package so I'm tempted to leave it as-is (last option). I might consider embedding a default version of the header in a future version of the module, to be used in case no system header can be detected and call it a day.

Let me know what y'all think..

toots:
Hmm old on, I didn't see that we need to export a head for `dssi` when migrating to `dune`. I'll do a re-released with the default `ladspa.h` header embedded as proposed above.

toots:
Ok, the release now installs the right headers, including a local copy of `ladspa.h`, used when no header could be found in the system. Also, `dssi` needs a proper release to support this new version.

mseri:
Thanks

PR #17758, opened 2020-12-04
Mark memtrace as incompatible with `arm32` and `x86_32`

talex5:
See https://github.com/janestreet/memtrace/issues/7.

This should fix the build problem shown at https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/741d81be6463c185927ec3db3a465db4a43bb887.

/cc @stedolan


talex5:
Note: the arm32 restriction (copied from albatross) doesn't work yet due to https://github.com/ocaml/opam/pull/4462.

mseri:
Thanks!

PR #17766, opened 2020-12-04
package Frama-C plugin MetAcsl

vprevosto:
This actually creates two packages:

- frama-c-metacsl for the brand new [MetAcsl](https://www.frama-c.com/fc-plugins/metacsl.html) package of Frama-C
- conf-swi-prolog, as MetAcsl optionally uses the SWI-prolog interpreter.

### DIFF
--- a/packages/conf-swi-prolog/conf-swi-prolog.1/opam
+++ b/packages/conf-swi-prolog/conf-swi-prolog.1/opam
@@ -1,0 +1,15 @@
+opam-version: "2.0"
+maintainer: "virgile.prevosto@m4x.org"
+homepage: "https://www.swi-prolog.org/"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: "BSD-2-Clause"
+build: ["swipl" "--version"]
+depexts: [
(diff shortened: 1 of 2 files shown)


mseri:
It would be nice if we could add support for more distributions in the conf prolog package. The libgnomecanvas failure is also annoying, but Inwould not block a merge on that.

I will wait to hear @kit-ty-kate opinion before merging

kit-ty-kate:
yeah many distributions don't ship with GTK2 anymore. I see that frama-c allows both gtk2 and gtk3 so our CI should get better once we switch to opam 2.1.

Otherwise, could you return the last fix I made to the opam file upstream? The `{pinned}` tag shouldn't really be used in this context, `{dev}` is what you're looking for.

Thanks!

vprevosto:
> Otherwise, could you return the last fix I made to the opam file upstream? The {pinned} tag shouldn't really be used in this context, {dev} is what you're looking for.

Thanks for the tip (and for having extended the list of distributions supporting `opam depext conf-swi-prolog`). I'll take care of that.

PR #17768, opened 2020-12-04
Add conf-tzdata

emillon:
This new conf package refers to the time zone database. These files are necessary at run time when using for example [the timezone package](https://github.com/janestreet/timezone).

The path is hardcoded in the build step but the test can be extended if necessary.


emillon:
This is defined by hand in the Real World OCaml repository (realworldocaml/book#3419), but opam-repository is probably a better place for this piece of information.

mseri:
Thanks

Issue #17680, opened 2020-11-22
opam-file-format 2.1.1 conflicts with core_kernel


na4zagin3:
I found opam-file-format 2.1.1 conflicts with core_kernel.

Issue https://github.com/ocaml/opam-file-format/issues/33 has the reproducible example.

I was about to send a MR to add conflicts section to opam-file-format 2.1.1 to tell OPAM they conflict with core_kernel but I found the problem did not reproduced without installing `core.v0.14.0`.

Could you tell me what I should do?

```
...
$ opam install --yes dune.2.7.1 "core.v0.14.0" core_kernel.v0.14.1 opam-format.2.0.7 opam-file-format.2.1.1
...
```

PR #17774, opened 2020-12-05
Make opam-file-format 2.1.1 conflict with core_kernel

na4zagin3:
See Issue ocaml/opam-file-format#33 for the reproducible example.

Closes https://github.com/ocaml/opam-repository/issues/17680.


kit-ty-kate:
Usually I would not mark this as incompatible since they're only incompatible when used together but in this case it makes sense since it's a bug that is going to be fixed in 2.1.2. Thanks!

PR #16568, opened 2020-06-03
Second alpha for OCaml 4.11.0

Octachron:
This is the new opam package for the second alpha of OCaml 4.11.0
The change compared to the first alpha are:
* [*additional fixes*] [6673](https://github.com/ocaml/ocaml/issues/6673), [1132](https://github.com/ocaml/ocaml/issues/1132), [9617](https://github.com/ocaml/ocaml/issues/9617): Relax the handling of explicit polymorphic types
   (Leo White, review by Jacques Garrigue and Gabriel Scherer)

* [*additional fixes*] [7364](https://github.com/ocaml/ocaml/issues/7364), [2188](https://github.com/ocaml/ocaml/issues/2188), [9592](https://github.com/ocaml/ocaml/issues/9592), [9609](https://github.com/ocaml/ocaml/issues/9609): improvement of the unboxability check for types
   with a single constructor. Mutually-recursive type declarations can
   now contain unboxed types. This is based on the paper
     https://arxiv.org/abs/1811.02300

- [9549](https://github.com/ocaml/ocaml/issues/9549), [9557](https://github.com/ocaml/ocaml/issues/9557): Make -flarge-toc the default for PowerPC and introduce
  -fsmall-toc to enable the previous behaviour.
  (David Allsopp, report by Nathaniel Wesley Filardo, review by Xavier Leroy)

- [9320](https://github.com/ocaml/ocaml/issues/9320), [9550](https://github.com/ocaml/ocaml/issues/9550): under Windows, make sure that the Unix.exec* functions
  properly quote their argument lists.
  (Xavier Leroy, report by André Maroneze, review by Nicolás Ojeda Bär
   and David Allsopp)

- [9490](https://github.com/ocaml/ocaml/issues/9490), [9505](https://github.com/ocaml/ocaml/issues/9505): ensure proper rounding of file times returned by
  Unix.stat, Unix.lstat, Unix.fstat.
  (Xavier Leroy and Guillaume Melquiond, report by David Brown,
   review by Gabriel Scherer and David Allsopp)

- [8676](https://github.com/ocaml/ocaml/issues/8676), [9594](https://github.com/ocaml/ocaml/issues/9594): turn debugger off in programs launched by the program
  being debugged
  (Xavier Leroy, report by Michael Soegtrop, review by Gabriel Scherer)

- [9552](https://github.com/ocaml/ocaml/issues/9552): restore ocamloptp build and installation
  (Florian Angeletti, review by David Allsopp and Xavier Leroy)

[...]

PR #17775, opened 2020-12-05
ocaml-base-compiler 4.11*, ocaml-variants 4.11*: do not pass CC=cc on FreeBSD

hannesm:
As done earlier (in #16575) and mentioned in
https://github.com/ocaml/opam-repository/pull/16568#commitcomment-39656969
this is no longer necessary since https://github.com/ocaml/ocaml/pull/9437 -
which was part of 4.11.0.

NB: I'm not a big fan of modifying existing packages, but imho this is sane
and safe in this case, and looking at #16575 it seemed to be fine with
maintainers back then.

I did not touch:
- 32bit
- flambda+musl
- flambda+musl+static

Since I don't understand how they are supposed to work in the first place.
Eventually it'd make sense to restrict them to supported platforms via
(available = [ os = "linux" & os-distribution = ... ])?

//cc @Octachron @kit-ty-kate @avsm -- AFAICT the 4.12 release is fine in this respect. Thanks.

### DIFF
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.4.11.0/opam
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.4.11.0/opam
@@ -19,8 +19,8 @@ build: [
     "./configure"
     "--prefix=%{prefix}%"
     "-C"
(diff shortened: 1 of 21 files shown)


kit-ty-kate:
Thanks!

PR #17778, opened 2020-12-06
Lens' ppx_deriving dependency is not build-only

mrmr1993:
Upgrading `ppx_deriving` should recompile the lens package's ppx library, otherwise it cannot be linked into ppx executables etc. after the upgrade.

This had already been fixed for 1.2.4; this PR applies the fix for versions `1.2.1`, `1.2.2`, `1.2.3` so that OCaml versions `< 4.10.0` work correctly too.


kit-ty-kate:
Thanks!

PR #17780, opened 2020-12-06
Package dssi.0.1.3

toots:
### `dssi.0.1.3`
Bindings for the DSSI API which provides audio synthesizers

This PR also includes some extra changes:
* dssi relies on the alsa headers so I added `conf-alsa` and switched the latest `alsa` to it.
* There is a compat header for alsa on OSX but I don't think that it can be used to compile a working alsa library on OSX so I limited this depext to `conf-dssi` and made `conf-alsa` only available on linux systems.
* While working on it, I noticed a little glitch on `conf-soundtouch` for `oraclelinux` and tried to fix it, pushing an update to the latest release. Alas, it didn't fix it but that's still a slight enhancement (`soundtouch` has historically multiple `pkg-config` packages names, which is annoying. I added a new one to the supported list).

Let me know if anything here needs fixing. Thanks!

### DIFF
--- a/packages/alsa/alsa.0.3.0/opam
+++ b/packages/alsa/alsa.0.3.0/opam
@@ -7,6 +7,7 @@ license: "GPL-2.0"
 homepage: "https://github.com/savonet/ocaml-alsa"
 bug-reports: "https://github.com/savonet/ocaml-alsa/issues"
 depends: [
+  "conf-alsa"
   "ocaml" {>= "4.02.0"}
   "dune" {> "2.0"}
   "dune-configurator"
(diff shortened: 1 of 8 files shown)


packages/alsa/alsa.0.3.0/opam
```diff
@@ -7,6 +7,7 @@ license: "GPL-2.0"
 homepage: "https://github.com/savonet/ocaml-alsa"
 bug-reports: "https://github.com/savonet/ocaml-alsa/issues"
 depends: [
+  "conf-alsa"
```
toots, line 10:
```suggestion
  "conf-alsa"
  "conf-pkg-config"
```

mseri:
Fails on arm64 and x86_32 but it seems on purpose. Do you need to have a look or it can be merged?

toots:
I think it's good to go. As far as I'm concerned, those are bugs with the build system, not us.

PR #17802, opened 2020-12-09
macOS 11 / XCode 12 fix for remake (used by gappa)

MSoegtropIMC:
This PR fixes an issue with the remake tool which came up on macOS 11 / XCode 12.

@silene : FYI.

### DIFF
--- a/packages/gappa/gappa.1.3.5/files/remake.patch
+++ b/packages/gappa/gappa.1.3.5/files/remake.patch
@@ -1,7 +1,16 @@
 diff --git a/remake.cpp b/remake.cpp
-index 3a6af80..47d5be8 100644
+index 3a6af80..d5ac424 100644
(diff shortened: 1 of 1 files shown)


mseri:
Thanks, can you move this to a patch release? Simply copy the whole folder (with the changes) into `gappa.1.3.5-1` and reset the current folder.

MSoegtropIMC:
@mseri : I don't think a patch release is required. The change doesn't change anything on platforms on which it does compile without the patch. On platforms where it did compile before, there is no need to recompile and on platforms on which it did not compile before it shouldn't be installed yet. So the only effect of a patch release would be that gappa gets recompiled needlessly. As far as I understand the policy is to not create a patch release in such a case.

mseri:
I agree

PR #17799, opened 2020-12-08
BAP 2.2.0 release

ivg:
The 2.2.0 release is focusing on stability and usability. It brings,
however, a few new features, some of them highlighted here (see the
full list at the [release][1] page):

- a floating-point lifter, contributed to us by ForAllSecure,
- a new thumb/thumbv2 lifter (thanks to @Phosphorus15 and @XVilka)
- support for interworking and new representation
- new representation of machine targets (instead of the old arch)
- completely rewritten llvm loader backend
- integration with radare2
- improved symbolization facilities (makes bap more autonomous)
- significant performance improvements
- a REPL powered by ocaml-linenoise, see the demo below

[1]: https://github.com/BinaryAnalysisPlatform/bap/releases/tag/v2.2.0

[asciicast](https://asciinema.org/a/358996)

PR #17805, opened 2020-12-09
constraints the version of piqilib on sedlex

ivg:
It was [detected][1] in #17799 build that piqilib.0.6.15 fails to
compile with sedlex.1.99.  The version 2.2.0 works for me, it might be
that others are also working, but I decided to be more conservative
and select 2.2.0 and above only.

CC @alavrik for situational awareness. And if you know better which
version works (and whether it affects the older versions of piqilib)
then please tell.

[1]: https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/c56ed99592368ed1bc27bb36062b6dce4ad23888/variant/compilers,4.08,bap-piqi.2.2.0


mseri:
Thanks!

PR #17811, opened 2020-12-11
opam-monorepo.0.1.0 requires dune >= 2.7

emillon:
It vendors its dependencies, including dune 2.7 (for dune libraries), so it needs a dune that can interpret `(lang dune 2.7)`.


emillon:
cc @NathanReb, you'll want to sync the opam file in the project as well.

mseri:
@emilion thanks! `dune` should probably also fail in this case. Can you open an issue there?

emillon:
Do you mean, it should generate a bound compatible with all subprojects? sure!

mseri:
Yes! Thank you very much

emillon:
@mseri done in ocaml/dune#4051

PR #17817, opened 2020-12-12
Update bitstring and ppx_bitstring to 4.1.0

xguerin:


kit-ty-kate:
The tag seems to have been modified since:
```
          - bitstring.4.1.0:
            https://github.com/xguerin/bitstring/archive/v4.1.0.tar.gz (Bad checksum, expected md5=5ef783b736837f6f3d0ef0af8eef6582)
```

xguerin:
Thanks, that should be addressed now.

kit-ty-kate:
Thanks! I spotted an issue: `ppxlib` shouldn't be tagged as a build dependency as it is used in the runtime and having it unsync would result in a failure. Could you return the fix upstream?

xguerin:
Sure, will do. Thanks !

PR #17818, opened 2020-12-12
Added opam packages for Coq 8.12.2 and CoqIDE 8.12.2

MSoegtropIMC:
This PR adds opam packages for the new Coq 8.12.2 release.

@Zimmi48 (the Coq 8.12 release manager) : FYI

### DIFF
--- a/packages/coq/coq.8.12.2/files/coq.install
+++ b/packages/coq/coq.8.12.2/files/coq.install
@@ -1,0 +1,12 @@
+bin: [
+  "bin/coqwc"
+  "bin/coqtop"
+  "bin/coqtop.byte"
+  "bin/coqdoc"
+  "bin/coqdep"
+  "bin/coqchk"
+  "bin/coqc"
+  "bin/coq_makefile"
+  "bin/coq-tex"
+  "bin/coqworkmgr"
+]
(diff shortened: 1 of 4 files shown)


RalfJung:
Great, I was about to do this but now I don't have to. :-) Thanks @MSoegtropIMC!

MSoegtropIMC:
A note on the CI failures on CentOS and OracleLinux: I guess the PowerTools system repo is not enabled on these, which is required for some depext packages (e.g. gtk-sourceview3). This can be enabled with:
```
sudo dnf config-manager --set-enabled PowerTools
```
On CentOS 8 I tested that it doesn't work without this and that it does work with this. Since OracleLinux is a related distribution, I guess it is the same there, but I don't know for sure (I can test this on a VM in case it fixes it on CentOS and not on OracleLinux).

I guess this needs to be done by a CI maintainer.

But I think neither of this should stop the merge of this PR - I would assume it was always like this for Coq and CoqIDE.

Zimmi48:
@kit-ty-kate If you agree with the assessment regarding CI, could you merge this?

kit-ty-kate:
Thanks! Sorry for the wait

MSoegtropIMC:
@kit-ty-kate : thanks for merging! Should I create an issue for CI on CentOS and OracleLinux?

kit-ty-kate:
@MSoegtropIMC it shouldn't be necessary. The issue on CentOS is fixed upstream (https://github.com/avsm/ocaml-dockerfile/pull/26) but has not yet trickled down to opam-repo-ci. Oraclelinux can be ignored though as there doesn't seem to be any similar repository for it and this particular distro breaks quite often anyway.

PR #17819, opened 2020-12-12
Add the packages for pytorch 1.7.

LaurentMazare:
This also splits the libtorch package in some specific macos and linux versions to avoid the extra downloads where not needed.


LaurentMazare:
The two remaining CI failures (centos-7, oraclelinux-7) seems to be caused by the gcc version not supporting c++14 which the main torch library requires. We could disable these two distributions in the opam file but this would feel a bit sad as users may be using a more recent centos-8 for example.

mseri:
Looks good! Thanks

PR #17826, opened 2020-12-15
Fix the libtorch macos checksum

LaurentMazare:
The md5 checksum for the macos libtorch version was wrong, I'm only discovering it now as it fails in the workflow tests of the ocaml-torch repo, sorry about that.
I'm not sure if it's ok to fix in place or if this requires a new version to be crafted (this PR fixes it in place).


mseri:
It is the official tarball, I think it is fine to change it in place in this case. Patch release would have been for a change in the ocaml library tarball, patches or for changes in the build

mseri:
Hopefully we get will get macos (and bsd) test runners soon

LaurentMazare:
Thanks @mseri !

PR #17827, opened 2020-12-15
+tezos-rust-libs.1.0

pirbo:
This package is a pragmatic solution to unlock the release of tezos version 8.0 in opam.

In order to work inside opam sandbox, we generated an archive where are vendored all the rust dependencies of our rust dependencies. This way we can `cargo build` locally the `.a` we've done binding for in the tezos-crypto package...


mseri:
Neat. And there is the install file to deal with the libs. Seems to work fine. LGTM

mseri:
I will wait a bit to see if there is any comment from other maintainers

kit-ty-kate:
That looks good. Thanks a lot!

avsm:
I'm taking a look at updating the base images to allow conf-rust to work on centos7/8 as well.

mseri:
Nice!

PR #17835, opened 2020-12-16
toml.6.0.0 and toml_cconv.6.0.0

zapashcanon:


packages/toml/toml.6.0.0/opam
```diff
@@ -0,0 +1,26 @@
...
+authors: [ "Julien Sagot" "Emmanuel Surleau" "mackwic" "Andrew Rudenko" "orbifx" "c-cube"]
+maintainer: "OCamlPro <contact@ocamlpro.com>"
+homepage: "https://ocaml-toml.github.io/To.ml/"
+synopsis: "TOML parser"
+description: "The toml library provides a parser and serializer for Tom's Obvious Minimal Language v0.4.0, a minimal configuration file format. Helpful getters to retrieve data as OCaml primitive types are also supplied"
+build: [ "dune" "build" "-p" name]
+license: "LGPL3"
+depends: [
+  "ocaml" {>= "4.02"}
+  "dune"
```
mseri, line 11:
```suggestion
  "dune" { >="1.2"}
```

zapashcanon:
@mseri, is this OK for you ? The failures are because of other packages missing lower bounds on OCaml or unrelated errors.

kit-ty-kate:
Thanks a lot! Sorry for the delay. Could you return the fixes upstream. The syntax clean is just so we don't have to deal with oneline descriptions (it's easier to deal with when we edit files) and passing `-j %{jobs}%` to dune is rather important as it allows users to controle how much resources they want opam to take on their system. If it wasn't there dune assume "take all the cores".

Could you return this in the opam files upstream?

zapashcanon:
Thanks ! I fixed the files in the upstream repository.

PR #17842, opened 2020-12-17
Package beluga.1.0

Ailrun:
### `beluga.1.0`

---
* Homepage: https://github.com/Beluga-lang/Beluga
* Source repo: git+https://github.com/Beluga-lang/Beluga.git
* Bug tracker: https://github.com/Beluga-lang/Beluga/issues

---

### DIFF
--- a/packages/beluga/beluga.1.0/opam
+++ b/packages/beluga/beluga.1.0/opam
@@ -1,0 +1,37 @@
+opam-version: "2.0"
+synopsis:
+  "Implementation of contextual modal logic for reasoning with higher-order abstract syntax"
+maintainer: "jacob.errington@mail.mcgill.ca"
+authors: [
+  "Brigitte Pientka"
+  "Joshua Dunfield"
+  "Andrew Cave"
+  "Jacob Thomas Errington"
+  "Junyoung Clare Jang"
+]
+license: "GPLv3"
+homepage: "https://github.com/Beluga-lang/Beluga"
+bug-reports: "https://github.com/Beluga-lang/Beluga/issues"
+depends: [
+  "dune" {>= "1.1"}
+  "dune-build-info"
+  "extlib"
+  "gen"
+  "linenoise"
+  "ocaml" {>= "4.08"}
+  "sedlex"
+]
+build: ["dune" "build" "-p" name "-j" jobs]
+run-test: [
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Hi, it seems that you called `opam publish` before the tag was pushed on the repository (at least that's my guess), and thus does not have the required `url` section describing where to get the archive.

Could you re-run `opam publish`? Sorry for the wait.

Ailrun:
Sure, I will try that asap.

Ailrun:
It looks like CI still gives the same error; did I miss something?

mseri:
Indeed, the opam file is stukk missing the `url` stanza. It should have something like
```
url {
  src: "https://..."
  checksum: [
    "md5=..."
    "sha512=..."
  ]
}
```
Usually `opam-publish` and `dune-release` populate it on their own.
Maybe it has some issues with the tag, you should be able to work around it by running
```
opam-publish https://github.com/Beluga-lang/Beluga/archive/v1.0.tar.gz
```

packages/beluga/beluga.1.0/opam
```diff
@@ -0,0 +1,41 @@
...
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "exec" "./TEST"] {with-test}
+  ["dune" "exec" "./TEST" "--" "--" "+htmltest"] {with-test}
+  ["dune" "exec" "./TEST" "--" "--" "+sexp"] {with-test}
+]
+run-test: [
+  ["./TEST"]
+  ["./TEST" "--" "+htmltest"]
+]
```
kit-ty-kate, line 33: The `./TEST` calls seem to repeat themselves twice (in build and in run-test). Is that expected?

kit-ty-kate:
I've fixed your constraints (CI showed that you don't support OCaml < 4.08) and your `dune-project` file shows that you require at least dune 1.1. If I made wrong guesses please correct me. If that's alright, could you return the fix in the opam file upstream?

Ailrun:
Upstream has updated.

kit-ty-kate:
Your `./TEST -- +sexp` test seems to have some issues. Could you have a look at it?

kit-ty-kate:
Thanks a lot!

PR #17851, opened 2020-12-18
Add package ocamlsdl2 version 0.04

fccm:
Various enhancements and naked pointers removal.

### DIFF
--- a/packages/ocamlsdl2/ocamlsdl2.0.04/opam
+++ b/packages/ocamlsdl2/ocamlsdl2.0.04/opam
@@ -1,0 +1,50 @@
+opam-version: "2.0"
+maintainer: "monnier.florent@gmail.com"
+authors: [
+  "Florent Monnier"
+  "Piotr Mardziel"
+  "David Cadé"
+  "Guillaume Munch-Maccagnoni"
+]
+license: "restrictionless Zlib"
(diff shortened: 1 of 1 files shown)


fccm:
Hi @kit-ty-kate,
The error that you mentioned about `ocamlsdl2.0.03` on the PR about `ocamlsdl2-ttf` seems to be there too with this version 0.04 on `ubuntu-20.10`:
```
# /usr/bin/ld: cannot find -ldrm
# /usr/bin/ld: cannot find -lgbm
```
The libs parameters to link with in the Makefile are taken from the result of the command:
```
sdl2-config --static-libs
```
This utility `sdl2-config` is provided by the lib `SDL2` (this package is an ocaml binding for this lib).
On my computer I get this:
```
$ sdl2-config --static-libs
-L/usr/lib -lSDL2 -Wl,--no-undefined -lm -ldl -lpthread -lrt
```
It look like a dependency issue in `ubuntu-20.10` and Debian.

kit-ty-kate:
@fccm hi, sorry I forgot to reply to you. Yes indeed that sounds like a bug in Debian. Would you be able to report it?

kit-ty-kate:
Looking into the errors again it seems that the previous release of ocamlsdl2 is also affected so I should not have waited on this bug to be fixed. Sorry for all this useless wait, merging now.

PR #17857, opened 2020-12-19
Add soupault.2.3.1 and adjust dependency of soupault.2.3.0 for a breaking change in To.ml

dmbaturin:
To.ml 6.0.0 released just after soupault 2.3.0 made a [breaking change](https://github.com/ocaml-toml/To.ml/issues/68).
Now soupault 2.3.0 doesn't build anymore. This requires both updating the dependency of 2.3.0 and a maintenance release.


kit-ty-kate:
Thanks!

PR #17846, opened 2020-12-18
new file:   packages/dolog/dolog.6.0.0/opam

UnixJunkie:

PR #17863, opened 2020-12-21
dolog {>= "4.0.0"} --> dolog {= "4.0.0"}

UnixJunkie:
because dolog 5.0.0 has a new API that is incompatible, in some use cases


UnixJunkie:
related to https://github.com/ocaml/opam-repository/pull/17846

kit-ty-kate:
equality should be avoided here in case a patch release is made. Could you add `& < "5.0.0"` instead?

UnixJunkie:
ok, I'll do that.

UnixJunkie:
superseded by https://github.com/ocaml/opam-repository/pull/17870

PR #17878, opened 2020-12-23
Package kind2.v1.3.0

daniel-larraz:
### `kind2.v1.3.0`
Multi-engine, parallel, SMT-based automatic model checker for safety properties of Lustre programs
Kind 2 is an open-source, multi-engine, SMT-based automatic model checker for safety properties of finite-state or infinite-state synchronous reactive systems expressed as in an extension of the Lustre language. In its basic configuration it takes as input a Lustre file annotated with properties to be proven invariant, and outputs for each property either a confirmation or a counterexample, i.e., a sequence inputs that falsifies the property. More advanced features include contract-based compositional verification, proof generation for proven properties, and contract-based test generation.

---
* Homepage: https://kind2-mc.github.io/kind2
* Source repo: git+https://github.com/kind2-mc/kind2.git
* Bug tracker: https://github.com/kind2-mc/kind2/issues

---


daniel-larraz:
Hi @kit-ty-kate. Thank you for your help. It seems that the latest z3 package doesn’t include a z3 binary. Version **4.8.8-1** includes it. So a possible solution would be to set `z3 4.8.8-1` as a test dependency. Does this sound reasonable?

kit-ty-kate:
Sure, that sounds reasonable. Would it be possible to change the version number on your side by dropping the `v` prefix btw? The `v` prefix should be avoided wherever possible (only the janestreet packages do, because of a change in version scheme a few years ago).

All this need is to do `git mv kind2.v1.3.0 kind2.1.3.0`, the rest can all stay the same, there are no issues there. Did you use `opam publish --keep-v` to have that effect btw?

daniel-larraz:
> Sure, that sounds reasonable. Would it be possible to change the version number on your side by dropping the `v` prefix btw? 

Done!

> All this need is to do `git mv kind2.v1.3.0 kind2.1.3.0`, the rest can all stay the same, there are no issues there. Did you use `opam publish --keep-v` to have that effect btw?

No, I didn’t use `--keep-v`. I just ran `opam publish` (opam v2.0.1).

kit-ty-kate:
Thanks a lot!

PR #17886, opened 2020-12-27
camlp5 8.00, not-ocamlfind 0.07.02, pa_ppx 0.07

chetmurthy:
This is a release of camlp5 8.00, and its accompanying pa_ppx 0.07 (still not really ready for 1.00)

camlp5 8.00 has many improvements over camlp5. 7.xx, most notably full support for all new syntax, including PPX attributes/extensions.  pa_ppx is a package of PPX rewriter infrastructure and re-implementations of many of the rewriters from the standard collection, but using camlp5 instead of the current standard infrastructure.  That said, many of the re-implementations are pretty much equivalent in function (though some have more) and many/most of the unit-tests are taken from those standard PPX rewriters.

This exposed problems with supporting Perl (and libpcre) packages in centos, alpine, oraclelinux.  I refactored perl support into a conf package (conf-perl-packages), updated camlp5 8.00~* packages to reflect this, and constrained GT 0.3.0 so that it would not build with much-older camlp5 (which did not support all the features required for GT 0.3.0).


chetmurthy:
FYI, I enabled make parallelism here; let's see how it goes (I of course also *tested* it, ha!)

chetmurthy:
Once all checks complete successfully, I'll squash this branch

chetmurthy:
Hum .... the "camelus" check has failed, but there are no logs available to know why?

Kakadu:
> Hum .... the "camelus" check has failed, but there are no logs available to know why?

@chetmurthy  I think that you pressed an extra `q` in checksum

https://github.com/ocaml/opam-repository/pull/17886/commits/0fdb5fa9268063c817192ec17040d2f00e735dcf

chetmurthy:
Oops, fat-fingered the checksum (thanks Kakadu@ !)

chetmurthy:
OK, I think I'm stuck.  If you can advise how I can debug these CI infra bugs, I'm happy to go further.  After I get it all cleaned-up, I'll squash all of this.  I'd like to add the "conf-diffutils" and "conf-perl-packages", just because it makes things cleaner,, but will remove them and just inline into my opam files if you'd prefer.  I think, at least, conf-diffutils ought to remain, though.

Kakadu:
> OK, I think I'm stuck. If you can advise how I can debug these CI infra bugs, I'm happy to go further. After I get it all cleaned-up, I'll squash all of this. I'd like to add the "conf-diffutils" and "conf-perl-packages", just because it makes things cleaner,, but will remove them and just inline into my opam files if you'd prefer. I think, at least, conf-diffutils ought to remain, though.

I performed some digging and it seems that we can't install package `perl-IPC-System-Simple` because it  exists only in Centos 8 but CI has centos 7. I see three ways to workaround this
1) ignore this issue
2) upgrade CI either to use Centos8 or add extra repository EPEL with a required package
3) change something to make this dependency optional

chetmurthy:
OK, I'll change previous versions of camlp5 so they don't install on centos-7 and oraclelinux-7.

chetmurthy:
OK, I think this is good-to-go now.  Still don't answers on how to make alpine, centos, oraclelinux work (all those depexts) but I guess it doesn't really matter so much.

Kakadu:
It will be great to merge it in this year....

chetmurthy:
Ehhh .... I'd expect sometime after the New Year: the opam-repo maintainers should have a Christmas/New Year's break, too, after all.

packages/camlp5/camlp5.8.00/opam
```diff
@@ -0,0 +1,58 @@
...
+]
+build: [
+  ["./configure" "--prefix" prefix "-libdir" lib "-mandir" man]
+  [make "-j%{jobs}%" "DEBUG=-g" "world.opt"]
+  [make "-j%{jobs}%" "DEBUG=-g" "all"]
+  [make "-C" "testsuite" "clean" "all-tests"] { with-test & ocaml:version >= "4.10.0" }
+  [make "-C" "test" "clean" "all"] {with-test & ocaml:version >= "4.10.0" }
+]
+install: [make "install"]
+available: [ os-distribution != "alpine" ]
```
kit-ty-kate, line 47: What's the reason why alpine is not supported? I couldn't see any reference to alpine in the discussion above

packages/conf-diffutils/conf-diffutils.1/opam
```diff
@@ -0,0 +1,22 @@
+opam-version: "2.0"
+homepage: "https://www.gnu.org/software/diffutils/"
+bug-reports: "chetsky@gmail.com"
+maintainer: "chetsky@gmail.com"
+authors: "GNU Project"
+license: "GPL-3.0-or-later"
+build: [["sh" "-exc" "find . -name ."]]
```
kit-ty-kate, line 7:
```suggestion
build: ["diff" "--help"]
```

packages/conf-diffutils/conf-diffutils.1/opam
```diff
@@ -0,0 +1,22 @@
+opam-version: "2.0"
```
kit-ty-kate, line 1: I'm a bit surprised by the addition of this package. Why do you need it? Which distributions failed without it? Do you require a specific flavour of `diff` (GNU?) ?

packages/conf-perl-packages/conf-perl-packages.1/opam
```diff
@@ -0,0 +1,45 @@
+opam-version: "2.0"
```
kit-ty-kate, line 1: would it be possible to split this into two packages: `conf-perl-string-shellquote` and `conf-perl-ipc-system-simple` ?
This way the name of the new package correctly reflects what it installs (which is rather vague with the current name/setup)

packages/conf-perl-packages/conf-perl-packages.1/opam
```diff
@@ -0,0 +1,45 @@
...
+    "libipc-system-simple-perl"
+  ] {os-family = "debian"}
+  [
+    "perl-string-shellquote"
+    "perl-ipc-system-simple"
+  ] {os-distribution = "alpine"}
+  [
+    "perl-String-ShellQuote"
+    "perl-IPC-System-Simple"
+  ] {os-distribution = "centos"}
```
kit-ty-kate, line 27:
```suggestion
  ] {os-distribution = "centos" | os-distribution = "ol"}
```
I've tested it using the `ocaml/opam:oraclelinux` docker image and both centos and oraclelinux have these two packages

packages/conf-perl-packages/conf-perl-packages.1/opam
```diff
@@ -0,0 +1,45 @@
...
+  [
+    "perl-String-ShellQuote"
+    "perl-IPC-System-Simple"
+  ] {os-family = "fedora"}
+  [
+    "perl-string-shellquote"
+    "perl-ipc-system-simple"
+  ] {os-family = "arch"}
+]
+available: [ os-distribution != "centos" & os-distribution != "ol" ]
```
kit-ty-kate, line 41: Per suggestion above this should not be necessary.

packages/not-ocamlfind/not-ocamlfind.0.07.02/opam
```diff
@@ -0,0 +1,35 @@
...
+  "ocamlfind" {>= "1.8.0"}
+  "fmt" {>= "0.8.8"}
+  "rresult" {>= "0.6.0"}
+  "ocamlgraph" {>= "2.0.0"}
+  "conf-which"
+]
+depexts: [
+  [
+    "xdot"
+  ] {os-family = "debian"}
```
kit-ty-kate, line 22: I'm curious (not a blocker), why is it needed only on debian?

kit-ty-kate (changes requested):

packages/conf-diffutils/conf-diffutils.1/opam
chetmurthy: arrgh, that was stupid!  Thank you for catching it!

packages/camlp5/camlp5.8.00/opam
chetmurthy: Oof, I -thought- that I had depext (perl package) problems on alpine, but now that I think about it, maybe not.  I'll remove it and see if the build breaks.arr

packages/conf-diffutils/conf-diffutils.1/opam
chetmurthy: OK, it's clear I made too many changes in a single PR.  Again, you caught my oversight!

Camlp5's tests (which I want to enable) (specifically, the last test that is run) needs a diff that knows "-Bwiu".

packages/conf-perl-packages/conf-perl-packages.1/opam
chetmurthy: 100% with you, will do.  I didn't do so, only b/c .... well, so many packages (I figured maybe it'd be less troublesome to the maintainers with fewer packages).

packages/conf-perl-packages/conf-perl-packages.1/opam
kit-ty-kate: Ah nevermind, it works with oraclelinux 8 but not oraclelinux 7. I'll ignore the error on oraclelinux 7

chetmurthy: How odd.  OK, I'll enable that and either it goes thru, or we'll have a testcase.  How odd.  I am -very- sure that both centos and ol had CI failures with these packages.

packages/not-ocamlfind/not-ocamlfind.0.07.02/opam
chetmurthy: there's function (to draw a graph of findlib packages starting from some initial set) that needs xdot.  I was too lazy to figure out what the right package would be on each distro (and at the time, I only knew the answer for debian/ubuntu).  Since it was
1. neither needed for the build
2. nor for testing,
3.  figured I wouldn't bother figuring out how to get it working on all the other distros, until somebody asked.

The useful part of not-ocamlfind is "reinstall-if-diff" and "preprocess"; this subcommand  (`package-graph`) is .... well, it's nice, but really, I don't expect anybody to install not-ocamlfind for that functionality.

Would you prefer that I figure out how to enable it on all the other platforms?

packages/conf-perl-packages/conf-perl-packages.1/opam
chetmurthy: Well, I also saw errors on centos-7.  Since I've already made the changes, I'll let it go thru CI (which will be -fast-, b/c early failures) and we'll see).

packages/camlp5/camlp5.8.00/opam
```diff
@@ -31,6 +31,7 @@ depends: [
   "ocaml" {>= "4.02" & < "4.13.0"}
   "conf-perl"
   "conf-perl-packages"
```
kit-ty-kate, line 33: This (and similar in several places) needs to change now that the packages have been renamed

chetmurthy: Yep, I noticed and pushed just before your comment came in.

packages/conf-diffutils/conf-diffutils.1/opam
kit-ty-kate: These options seem to be available on MacOS, FreeBSD and NetBSD so that sounds ok portability wise.
Opam already requires a posix `diff` command (see https://github.com/ocaml/opam/blob/master/src/client/opamInitDefaults.ml#L126) so I believe this package is not required

packages/not-ocamlfind/not-ocamlfind.0.07.02/opam
kit-ty-kate: Ideally yes, but it's ok for now given previous versions also did the same. Do it when you are bored and have some spare time.

packages/conf-diffutils/conf-diffutils.1/opam
chetmurthy: Right, it was on a platform that ran everything in busybox, where I saw the error.  I can remove this prereq, and if it fails, put it back in.

kit-ty-kate: Ah! I forgot to check busybox :D I see! So that's why you marked Alpine as unavailable in the first place.
In that case could you advertise this package as installing GNU diff specifically and change the camlp5 package the following way?:
```
"conf-diffutils" { with-test & os-distribution = "alpine" }
```

chetmurthy:
OK, the CI run has completed: https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/7fb4adf27fdeac822c9b5af5e76ca7fee08d2200
and there is one problem left:
(1) oraclelinux-7 doesn't support perl-IPC-System-Simple

I believe I've applied every other suggestion you've made.  I'll check again tomorrow AM, just to be sure.

kit-ty-kate:
as said in #17897 this is expected and fine. I'll just ignore the error and go ahead with the merge. Thanks!

chetmurthy:
I'm consistently impressed with the quality of this .... cultivation of the package-collection.  Thank you and your colleagues, for doing this.  On New Year's Eve, no less!

packages/pa_ppx/pa_ppx.0.07/opam
```diff
@@ -0,0 +1,67 @@
...
+  "ocaml"       { >= "4.10.0" & < "4.13.0" }
+  "conf-perl"
+  "conf-perl-ipc-system-simple"
+  "conf-perl-string-shellquote"
+  "camlp5"      { >= "8.00" }
+  "not-ocamlfind" { >= "0.01" }
+  "pcre" { >= "7.4.3" }
+  "yojson" { >= "1.7.0" }
+  "sexplib0"
+  "ounit"
```
Kakadu, line 47: @chetmurthy Are you sure that `ounit` is required not only when performing tests?

chetmurthy: Good point; I've made the change, and that'll go out with the next bugfix release.  Thanks!

[I grepped to check this, and yeah, it seems you're right.  Maybe there's some odd reason that ounit is needed; if so, I'll find out when preparing the next bugfix release.]

packages/camlp5/camlp5.8.00/opam
```diff
@@ -0,0 +1,59 @@
...
+pretty print module
+It works as a shell command and can also be used in the OCaml toplevel."""
+maintainer: "https://github.com/ocaml/opam-repository/issues"
+authors: ["Daniel de Rauglaudre" "Chet Murthy"]
+license: "BSD-3-Clause"
+homepage: "https://camlp5.github.io"
+doc: "https://camlp5.github.io/doc/html"
+bug-reports: "https://github.com/camlp5/camlp5/issues"
+depends: [
+  "ocaml" {>= "4.02" & < "4.13.0"}
```
Kakadu, line 31: `>= "4.10"`
@chetmurthy

Issue #10925, opened 2017-12-05
Confusing state of num in 4.05.0 and ocamlfind 1.7.3


rleonid:
I updated my `4.05.0` switch to ocamlfind 1.7.3 and now
```
$ ocamlfind query num
ocamlfind: Package `num' not found
```
but
```
$ opam info num

=-=- num: information on all versions -=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=  🐫
name                   num
all-installed-versions 0 [4.05.0+trunk+flambda 4.05.0+flambda 4.05.0+spacetime 4.05.0+flambda+spacetime 4.04.0 4.05.0]  1.1 [4.06.0]
all-versions           0  1.0  1.1

=-=- Version-specific details -=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=  🐫
version      0
repository   default
homepage:    https://github.com/ocaml/ocaml
bug-reports: http://caml.inria.fr/mantis/
dev-repo:    git+https://github.com/ocaml/ocaml.git
authors:     "Valérie Ménissier-Morain" "Pierre Weis" "Xavier Leroy"
license:     LGPL-2 with OCaml linking exception
depends:     "ocaml" {< "4.06.0"} "base-num"
synopsis     The Num library for arbitrary-precision integer and rational arithmetic
```
and
```
$ opam list | grep num
base-num                base         Num library distributed with the OCaml compiler
num                     0            The Num library for arbitrary-precision integer and rational arithmetic
parany                  3.0.0        Parallelize any computation over independent items, even if there is an infinite number of them.
ppx_enumerate           v0.9.0       Generate a list containing all values of a finite type
```

dbuenzli:
Thanks for the report, maybe related to https://github.com/ocaml/opam-repository/issues/10857 didn't investigate further.  /cc @xavierleroy the maintainer of the `num` and `base-num` packages.

xavierleroy:
I think the problem doesn't come from the `num` and `base-num` packages, but comes from ocamlfind 1.7.3.  There is a heuristics in the configure script for ocamlfind that determines whether
1. the Num library is provided by the core OCaml installation, in which case ocamlfind should provide its own META files for it, or
2. the Num library is no longer in the core OCaml installation and is handled separately (e.g. by the OPAM `num` package version >= 1), in which case someone else will install the appropriate META file, and ocamlfind itself should not.

[...]

PR #17891, opened 2020-12-28
ocamlfind: Fix num's META file installation when reinstalled

kit-ty-kate:
Fixes https://github.com/ocaml/opam-repository/issues/10925

### DIFF
--- a/packages/ocamlfind/ocamlfind.1.3.1/opam
+++ b/packages/ocamlfind/ocamlfind.1.3.1/opam
@@ -25,8 +25,7 @@ depends: [
   "conf-m4" {build}
   "num" {= "0"}
 ]
-install: [make "install"]
-remove: [
+install: [
   [
     "./configure"
     "-bindir"
@@ -37,10 +36,9 @@ remove: [
     man
     "-config"
     "%{lib}%/findlib.conf"
-    "-no-custom"
     "-no-topfind" {ocaml:preinstalled}
   ]
-  [make "uninstall"]
+  [make "install"]
 ]
 synopsis: "A library manager for OCaml."
 description: """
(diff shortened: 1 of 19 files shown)


kit-ty-kate:
Waiting a few days for review.

avsm:
Great to see this fixed. I think it really needs an ok from @dra27, as he looked at the original problem report.

dra27:
This is a very good catch! It feels as though there should be fixes in two places: it would be good if findlib's build system supported deferring the detection of the `META` files so that it could be run in `install`, but it would also be good if opam had a flag which indicated a package which doesn't support being built when it's already installed.

For this PR, the duplication of the full `configure` line is unfortunate (it always was) and it also reveals past inconsistency with use of flags like `-no-custom`. Looking both at 1.3.1 (the oldest packaged version of `ocamlfind`) and 1.8.1, I would propose this simplification of the `install` stage:

```
install: [
  # Redetect currently-installed META files, see ocaml/opam-repository#17891
  ["mv" "Makefile.config" "build-config"]
  ["./configure" "-sitelib" lib]
  ["mv" "build-config" "Makefile.config"]
  [make "install"]
]
```

The re-run of `configure` will generate a nonsense `Makefile.config`, but it will succeed and it will generate the correct set of `META` files. All of those missing flags only affect `Makefile.config`, they don't affect any tests in `configure`, beyond some unnecessary probing of `ocamlc -config`.

kit-ty-kate:
While I agree this duplication is unfortunate, I'm really not convinced by the alternative proposal. This issue is going to be fixed upstream at some point anyway (via https://github.com/ocaml/ocamlfind/pull/12) so I'd rather make sure we keep those two duplicate equal rather than relying on knowing what every options to `./configure` does in every versions. (from my point of view)

It can still be improved in a future PR but for now I'd rather fix the original issue as I keep being busy with other things and don't really have time to look at that. It's not a huge deal if we have this intermediate solution in the meantime.

dra27:
I completely agree! As we'd discussed offline, if/when the upstream patch is merged, there's a minimum set of backports possibly worth doing, but the problem will be permanently fixed with the latest version

PR #17892, opened 2020-12-28
[new release] capnp-rpc-net, capnp-rpc-lwt, capnp-rpc-mirage, capnp-rpc and capnp-rpc-unix (0.9.0)

talex5:

##### CHANGES:

- Add connection progress indicator (@talex5 mirage/capnp-rpc#220).  
  Clients can now use `Capnp_rpc_unix.with_cap_exn` to show a progress indicator while waiting to connect.
  It will use the log if enabled.
  If not, but stderr is a tty, it shows a message while connecting and then erases it (if connecting takes longer than 0.5s).
  If stderr is not a tty, it just prints a message there.

- Use `Mirage_crypto_rng_lwt.initialize` instead of `Mirage_crypto_rng_unix.initialize` (@hannesm mirage/capnp-rpc#217).  
  The former periodically feeds entropy to the RNG, while the latter seeds the RNG only once.

- Set TCP_KEEPIDLE=60 if possible (@talex5 mirage/capnp-rpc#214).  
  Cap'n Proto connections tend to be long lived and we therefore turn on the `SO_KEEPALIVE` option.
  However, the default keepalive timeout of 2 hours is much too long for some networks.
  In particular, Docker's libnetwork silently drops idle connections after about 10 minutes.

Windows support:

- Prevent crash if ExtUnix wasn't built with sockopt (@MisterDA mirage/capnp-rpc#218).  
  If ExtUnix isn't built with sockopt then the symbol `have_sockopt_int` isn't available and an exception is raised. 
  This may be the case on some environments (e.g., ExtUnix built with mingw-w64).
  It is a workaround until support for `TCP_KEEPIDLE` is added for these environments (e.g., using Winsock instead).

- There is no SIGPIPE on Windows (@dra27 mirage/capnp-rpc#212).

- Switch from "capnpc" to "capnp compile" (@talex5 mirage/capnp-rpc#213).  
  `capnpc` seems to be the old name, and isn't present on Windows.

Build changes:

- Remove Travis and Dockerfile (@talex5 mirage/capnp-rpc#220).  
  Travis is very slow and error-prone now, and ocaml-ci generates a working Dockerfile automatically for people who want it.

- Include transitive dependencies in opam files (@talex5 mirage/capnp-rpc#219).  
  Patch generated by https://github.com/ocurrent/opam-dune-lint/

- Require `tcpip.unix` specifically to get the checksum stubs linked (@hannesm mirage/capnp-rpc#216).

### DIFF
--- a/packages/capnp-rpc-lwt/capnp-rpc-lwt.0.9.0/opam
+++ b/packages/capnp-rpc-lwt/capnp-rpc-lwt.0.9.0/opam
@@ -1,0 +1,39 @@
+opam-version: "2.0"
+synopsis:
+  "Cap'n Proto is a capability-based RPC system with bindings for many languages"
+description: """
+This package provides a version of the Cap'n Proto RPC system using the Cap'n
+Proto serialisation format and Lwt for concurrency."""
+maintainer: "Thomas Leonard <talex5@gmail.com>"
+authors: "Thomas Leonard <talex5@gmail.com>"
+license: "Apache"
+homepage: "https://github.com/mirage/capnp-rpc"
+bug-reports: "https://github.com/mirage/capnp-rpc/issues"
+doc: "https://mirage.github.io/capnp-rpc/"
+depends: [
+  "ocaml" {>= "4.03.0"}
+  "conf-capnproto" {build}
+  "capnp" {>= "3.4.0"}
+  "capnp-rpc" {= version}
+  "lwt"
+  "astring"
+  "fmt"
+  "logs"
+  "asetmap"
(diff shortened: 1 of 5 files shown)


talex5:
The only CI errors are from distributions where the dependencies don't install (`hacl_x25519` fails to install on CentOS, and `conf-capnproto` doesn't install on Oracle).

kit-ty-kate:
I tried to add support for capnproto on oraclelinux in #17902 but the package does not exists.

Looks good otherwise, thanks!

PR #17898, opened 2020-12-29
[new release] ocaml_portaudio (0.19.6)

wlitwin:
CHANGES:

First release of ocaml_portaudio bindings. These bindings are made for portaudio v19.6, may work with other versions. The bindings are more or less 1:1. The Ocaml_portaudio.Portaudio.C module contains the raw C bindings. The Ocaml_portaudio.Portaudio module contains a higher-level OCaml interface to the portaudio library. *Warning* The Portaudio.View module wraps plain C arrays, so be careful with them.

### DIFF
--- a/packages/ocaml_portaudio/ocaml_portaudio.0.19.6/opam
+++ b/packages/ocaml_portaudio/ocaml_portaudio.0.19.6/opam
@@ -14,6 +14,7 @@ depends: [
   "dune" {>= "2.7"}
   "conf-portaudio"
   "ctypes"
+  "ctypes-foreign"
   "odoc" {with-doc}
 ]
 build: [
--- a/packages/ocaml_portaudio/ocaml_portaudio.0.19.6/opam
+++ b/packages/ocaml_portaudio/ocaml_portaudio.0.19.6/opam
@@ -1,0 +1,42 @@
+opam-version: "2.0"
+synopsis: "Bindings to the C portaudio library"
+description: """
+Bindings to the C portaudio library."
+Exposes low-level C bindings and a higher level OCaml interface."
+Version corresponds to the portaudio version this was built to."
+"""
+maintainer: ["walter@litwinczyk.com"]
+authors: ["Walter Litwinczyk"]
+license: "MIT"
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
You missed a dependency on ctypes-foreign to be able to use the foreign interface:
```
#=== ERROR while compiling ocaml_portaudio.0.19.6 =============================#
# context              2.0.7 | linux/x86_64 | ocaml-base-compiler.4.02.3 | pinned(https://github.com/wlitwin/ocaml_portaudio/releases/download/0.19.6/ocaml_portaudio-0.19.6.tbz)
# path                 ~/.opam/4.02/.opam-switch/build/ocaml_portaudio.0.19.6
# command              ~/.opam/4.02/bin/dune build -p ocaml_portaudio -j 47 @install
# exit-code            1
# env-file             ~/.opam/log/ocaml_portaudio-1-dcf56b.env
# output-file          ~/.opam/log/ocaml_portaudio-1-dcf56b.out
### output ###
# File "/home/opam/.opam/4.02/lib/ctypes/META", line 1, characters 0-0:
# Error: Library "ctypes.foreign.threaded" in /home/opam/.opam/4.02/lib/ctypes
# is hidden (unsatisfied 'exist_if').
# -> required by library "ctypes.foreign" in /home/opam/.opam/4.02/lib/ctypes
# Hint: try:
#   dune external-lib-deps --missing -p ocaml_portaudio -j 47 @install
```
I fixed it in the above commit. Could you return it in the opam file upstream as well?

Apart from that would it be possible to rename this package to something else? We try to avoid the ocaml prefix on new package when not necessary. The obvious choice would have been to simply name it `portaudio` however it seems that it is already taken by https://github.com/savonet/ocaml-portaudio (which is also a good reason to rename the project to something else since they're both named `ocaml{_,-)portaudio`)

wlitwin:
Thanks for the fix @kit-ty-kate. I'll update the repo and think of a different name to use.

wlitwin:
Closing because I renamed the package to `portaudio_c_bindings` https://github.com/ocaml/opam-repository/pull/17916

PR #17902, opened 2020-12-31
conf-capnproto: Add support for oraclelinux

kit-ty-kate:


kit-ty-kate:
Interesting... oh well. No capnproto on this plateform then.

PR #17909, opened 2020-12-31
conf-rdkit: handle newer distributions

kit-ty-kate:


kit-ty-kate:
@UnixJunkie what do you think about this instead of #15318? This seems to work on Debian < 10 and also above #15318 only works above stable

UnixJunkie:
thanks a lot for those improvements

kit-ty-kate:
Thanks!

Issue #17897, opened 2020-12-29
CI infra bug? depexts for centos, oraclelinux seem busted?


chetmurthy:
I have some depexts that have worked for many months, and they're failing now.  In order to both make it simpler to deal with busted distributions, and also to debug this, I created a conf package `conf-perl-packages` and checked this behaviour.  Here's the PR I created and then closed, for reproduction.

https://github.com/ocaml/opam-repository/pull/17896

I don't know precisely how to go further in debugging, but would be happy to take advice and continue.

Part of what seems to confirm that this is a new bug in the CI infrastructure, is that `GT.0.3.0` no longer builds.  That is, it depends on `camlp5.8.00~alpha06`, and that package depends on external Perl packages.  This camlp5 package has already been released, and built fine previously.  Now it's failing on trying to install these perl packages, on centos and oraclelinux.

You can see this in the following PR: https://github.com/ocaml/opam-repository/pull/17886

I reproduced  using the instructions in the CI logs, and found the following error:
```
The following command needs to be run through "sudo":
    rpm -q --whatprovides perl-IPC-System-Simple perl-String-ShellQuote
no package provides perl-IPC-System-Simple
perl-String-ShellQuote-1.04-10.el7.noarch
OS package installation failed
The command '/bin/sh -c opam depext -uivy GT.0.3.0' returned a non-zero code: 1
```

kit-ty-kate:
The error simply comes from new distributions that we are testing.

The two issues that you've seen on oraclelinux and centos come from the fact that the `perl-IPC-System-Simple` package does not exist in oraclelinux 7 and requires the EPEL repository in centos-7. For centos this would be simply fixed by doing:
```
depexts: [
  ["epel-release" "perl-IPC-System-Simple"] {os-distribution = "centos"}
]
```
For oraclelinux 7 it's a bit trickier but once https://github.com/ocurrent/opam-repo-ci/pull/62 is merge we should be able to remove it from the tests for your particular needs. In the general case we simply ignore these errors.

PR #17911, opened 2020-12-31
OCanren

Kakadu:


Kakadu:
There are issues related to oracle linux that should be ignored because of #17897

kit-ty-kate:
Dune and Camlp5 should not be tagged as build dependencies and the requires lowest version of dune is 2.5. Could you return the fix upstream?

I also have one question about this package. It seems to be a fork from Jetbrain, is there an agreement to release this written somewhere? Is upstream ok with this?

Kakadu:
> Dune and Camlp5 should not be tagged as build dependencies and the requires lowest version of dune is 2.5. Could you return the fix upstream?
> 
> I also have one question about this package. It seems to be a fork from Jetbrain, is there an agreement to release this written somewhere? Is upstream ok with this?

Everything is fixed now. @kit-ty-kate, please recheck

kit-ty-kate:
Thanks!

PR #17914, opened 2021-01-02
Package bjack.0.1.6

toots:
### `bjack.0.1.6`
Bindings for the Jack library which provides functions for linking audio programs

---
* Homepage: https://github.com/savonet/ocaml-bjack
* Source repo: git+https://github.com/savonet/ocaml-bjack.git
* Bug tracker: https://github.com/savonet/ocaml-bjack/issues

---


packages/conf-bjack/conf-bjack.1/opam
```diff
@@ -0,0 +1,21 @@
+opam-version: "2.0"
```
kit-ty-kate, line 1: it seems that this file is attempting to do two things at once. `conf-samplerate` already exists. Could you use this one and create `conf-jack` separately?

toots:
Excellent idea, thanks! Just pushed it and also renamed `conf-bjack` into `conf-jack` to that the package clearly indicates if `jack` is installed (`bjack` has a wrapper around the C API to create blocking calls)

kit-ty-kate:
Thanks a lot!

PR #17918, opened 2021-01-04
Package mm.0.7.0

toots:
### `mm.0.7.0`
The mm library contains high-level to create and manipulate multimedia streams (audio, video, MIDI)

---
* Homepage: https://github.com/savonet/ocaml-mm
* Source repo: git+https://github.com/savonet/ocaml-mm.git
* Bug tracker: https://github.com/savonet/ocaml-mm/issues

---

### DIFF
--- a/packages/mkaudio/mkaudio.1.1.1/opam
+++ b/packages/mkaudio/mkaudio.1.1.1/opam
@@ -11,7 +11,7 @@ depends: [
   "ocaml"
   "dune" {>= "2.0"}
   "cmdliner"
-  "mm" {>= "0.6.0"}
+  "mm" {>= "0.6.0" & < "0.7.0"}
   "result"
   "ounit" {with-test}
 ]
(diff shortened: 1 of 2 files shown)


toots:
As far as I can tell, the missing header in `alpine` is part of `musl-dev`. I don't think `mm` should be in charge of pulling this depext. Maybe something missing the test environment setup?

kit-ty-kate:
> As far as I can tell, the missing header in `alpine` is part of `musl-dev`. I don't think `mm` should be in charge of pulling this depext. Maybe something missing the test environment setup?

I've fixed the issue. Other packages (core, …) fix this by adding this depexts only on alpine.
On top of that there was an issue with the depopts. Constraints should almost never placed in the depopts field and placed in the conflicts field instead (the behaviour is different)

kit-ty-kate:
All green now. Thanks!

toots:
Fixes backported, thanks so much for your help on this!

PR #17922, opened 2021-01-05
[new release] index and index-bench (1.3.0)

CraigFe:
A platform-agnostic multi-level index for OCaml

- Project page: <a href="https://github.com/mirage/index">https://github.com/mirage/index</a>
- Documentation: <a href="https://mirage.github.io/index/">https://mirage.github.io/index/</a>

## Added

- Added `flush_callback` parameter to the creation of a store, to register
  a callback before a flush. This callback can be temporarily disabled by
  `~no_callback:()` to `flush`. (mirage/index#189, mirage/index#216)

- Added `Stats.merge_durations` to list the duration of the last 10 merges.
  (mirage/index#193)

- Added `is_merging` to detect if a merge is running. (mirage/index#192)

- New `IO.Header.{get,set}` functions to read and write the file headers
  atomically (mirage/index#175, mirage/index#204, @icristescu, @CraigFe, @samoht)

- Added a `throttle` configuration option to select the strategy to use
  when the cache are full and an async merge is already in progress. The
  current behavior is the (default) `` `Block_writes`` strategy. The new
  `` `Overcommit_memory`` does not block but continue to fill the cache instead.
  (mirage/index#209, @samoht)

- Add `IO.exists` obligation for IO implementations, to be used for lazy
  creation of IO instances. (mirage/index#233, @CraigFe)

- `Index.close` now takes an `~immediately:()` argument. When passed, this
  causes `close` to terminate any ongoing asynchronous merge operation, rather
  than waiting for it to finish. (mirage/index#185, mirage/index#234)

- Added `Index.Checks.cli`, which provides offline integrity checking of Index
  stores. (mirage/index#236)

- `Index.replace` now takes a `~overcommit` argument to postpone a merge. (mirage/index#253)

- `Index.merge` is now part of the public API. (mirage/index#253)

- `Index.try_merge` is now part of the public API. `try_merge' is a no-op if
  the number of entries in the write-ahead log is smaller than `log_size`,
  otherwise it's `merge'. (mirage/index#253 @samoht)

## Changed

- `sync` has to be called by the read-only instance to synchronise with the
  files on disk. (mirage/index#175)
- Caching of `Index` instances is now explicit: `Index.Make` requires a cache
  implementation, and `Index.v` may be passed a cache to be used for instance
  sharing. The default behaviour is _not_ to share instances. (mirage/index#188)

## Fixed

- Added values after a clear are found by read-only instances. (mirage/index#168)
- Fix a race between `merge` and `sync` (mirage/index#203, @samoht, @CraigFe)
- Fix a potential loss of data if a crash occurs at the end of a merge (mirage/index#232)
- Fix `Index.iter` to only iterate once over elements persisted on the disk
  (mirage/index#260, @samoht, @icristescu)

### DIFF
--- a/packages/index-bench/index-bench.1.3.0/opam
+++ b/packages/index-bench/index-bench.1.3.0/opam
@@ -1,0 +1,39 @@
+opam-version: "2.0"
+maintainer:   "Clement Pascutto"
+authors:      ["Clement Pascutto" "Thomas Gazagnaire" "Ioana Cristescu"]
+license:      "MIT"
+homepage:     "https://github.com/mirage/index"
+bug-reports:  "https://github.com/mirage/index/issues"
+dev-repo:     "git+https://github.com/mirage/index.git"
+
+build: [
+ ["dune" "subst"] {pinned}
+ ["dune" "build" "-p" name "-j" jobs]
+ ["dune" "runtest" "-p" name] {with-test}
+]
+
+depends: [
+  "ocaml"   {>= "4.03.0"}
+  "cmdliner"
+  "dune"    {>= "2.7.0"}
+  "fmt"
+  "index"
+  "metrics"
+  "metrics-unix"
+  "ppx_deriving_yojson"
+  "re"
+  "stdlib-shims"
+  "yojson"
+  "ppx_repr"
+]
+
+synopsis: "Index benchmarking suite"
+x-commit-hash: "70bd80e40e0320075e80ddf443f7f462684cbdd3"
+url {
+  src:
+    "https://github.com/mirage/index/releases/download/1.3.0/index-1.3.0.tbz"
+  checksum: [
+    "sha256=d6fc288801f05962203dd820e358843e696cc0be775974df2a9ae9741e871c03"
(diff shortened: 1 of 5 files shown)


kit-ty-kate:
cc @pirbo for the constraint on tezos-storage

kit-ty-kate:
@CraigFe I had a look at the failure in irmin-unix. I think this is an issue with the fact that all the irmin subpackages are not synchronized with each other and break when used because the API do not match.
In this case `type endpoint` is missing a constraint that exists in 2.2.0 but not before.

Would it be possible to synchronize all the irmin packages to fix this sort of issue? I would gladly do the PR for that.

CraigFe:
@kit-ty-kate: Yes, let's synchronise the packages. We've actually [already done part of this](https://github.com/mirage/irmin/pull/1101) upstream in Irmin, but the final step is to make 2.3.0-era packages incompatible with 2.2.0.

Thanks for your help.

kit-ty-kate:
All good now. Thanks a lot!

PR #17927, opened 2021-01-06
Make opam-repository compatible with opam 2.1.0~beta4

kit-ty-kate:
There are many other packages which now require `OPAMCLI=2.0` but let's start with these ones.
We will also need to to a thorough search of the logs as errors might be ignored in some cases and the only discovered when actually using the package.

cc @rjbou @AltGr

### DIFF
--- a/packages/asli/asli.0.1/opam
+++ b/packages/asli/asli.0.1/opam
@@ -23,6 +23,7 @@ depends: [
   "zarith"
   "z3" {<= "4.7.1"}
 ]
+build-env: [OPAMCLI = "2.0"]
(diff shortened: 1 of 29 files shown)


kit-ty-kate:
oh actually, now that I think of it... Would it be possible to add this `OPAMCLI=2.0` automatically when the `opam-version` field of the package is `2.0`? I'll open an issue

dra27:
I had thought it was mirage packages affected by this, but it looks like there are oasis-based things too?

kit-ty-kate:
There isn't really any ways of getting the opam prefix (useful for compilers and other tools) in dune, for example. There are many small cases like this that require opam on whichever build-system the developer is using.

Closing in favour of https://github.com/ocaml/opam/pull/4492

PR #17932, opened 2021-01-06
adds bap-thumb and bap-systemz to the bap meta package

ivg:
We forgot to add those packages during the BAP 2.2.0 release. The packages implement support for systemz (very minimal) and thumb/thumb2 instruction sets.


kit-ty-kate:
Thanks!

PR #17935, opened 2021-01-07
camlbz2 changed upstream to gitlab.com

josch:
Hi,

with the blessings of its former maintainer @zacchiro the camlbz2 sources moved from the abandoned forge.ocamlcore to gitlab.com/irill/camlbz2.

Unfortunately, the tarball hashes are different because the new tarball is generated directly from git. As such, some files are now missing:

```
$ tar xf camlbz2-camlbz2-0.6.0.tar.gz
$ tar xf camlbz2-0.6.0.tar.gz
$ diff -ru camlbz2-0.6.0 camlbz2-camlbz2-0.6.0
Only in camlbz2-0.6.0: aclocal.m4
Only in camlbz2-0.6.0: configure
Only in camlbz2-0.6.0: doc
Only in camlbz2-camlbz2-0.6.0: .gitignore
Only in camlbz2-camlbz2-0.6.0: ocaml.m4
```

So before calling `./configure`, opam has to call `autoreconf` to generate `configure`. I am not an opam user, so I leave this change up to you.

Thanks!

### DIFF
--- a/packages/camlbz2/camlbz2.0.6.0/opam
+++ b/packages/camlbz2/camlbz2.0.6.0/opam
@@ -14,16 +14,15 @@ build: [
   ["./configure" "--prefix" prefix]
   [make]
 ]
-remove: [["ocamlfind" "remove" "bz2"]]
-depends: ["ocaml" "ocamlfind" "conf-autoconf" "conf-aclocal"]
-depexts: [
-  ["libbz2-dev" "autoconf"] {os-family = "debian"}
-  ["bzip2-dev"] {os-family = "alpine"}
-  ["bzip2-devel"] {os-distribution = "centos" | os-distribution = "fedora" | os-family = "suse" | os = "oraclelinux"}
+depends: [
+  "ocaml"
+  "ocamlfind"
+  "conf-autoconf"
+  "conf-aclocal"
+  "conf-libbz2"
 ]
 install: [make "install"]
 synopsis: "Bindings for bzip2"
-flags: light-uninstall
 url {
   src: "https://gitlab.com/irill/camlbz2/-/archive/0.6.1/camlbz2-0.6.1.tar.gz"
   checksum: "md5=cfb844c53833f9c9726a79fe093490ad"
--- a/packages/camlbz2/camlbz2.0.6.0/opam
+++ b/packages/camlbz2/camlbz2.0.6.0/opam
@@ -9,21 +9,17 @@ license: "LGPL-2.1-or-later with OCaml-LGPL-linking-exception"
 bug-reports: "https://gitlab.com/irill/camlbz2/issues"
 dev-repo: "git+https://gitlab.com/irill/camlbz2.git"
 build: [
-  ["aclocal" "-I" "."]
-  ["autoconf"]
   ["./configure" "--prefix" prefix]
   [make]
 ]
 depends: [
   "ocaml"
   "ocamlfind"
-  "conf-autoconf"
-  "conf-aclocal"
   "conf-libbz2"
 ]
 install: [make "install"]
 synopsis: "Bindings for bzip2"
 url {
-  src: "https://gitlab.com/irill/camlbz2/-/archive/0.6.1/camlbz2-0.6.1.tar.gz"
-  checksum: "md5=cfb844c53833f9c9726a79fe093490ad"
+  src: "https://gitlab.com/irill/camlbz2/uploads/77de1eae9411b4927b6849a3562341a1/camlbz2-0.6.0.tar.gz"
+  checksum: "md5=7a1cf822b3fe0ef57df4f8ebd86cac99"
 }
--- a/packages/camlbz2/camlbz2.0.6.0/opam
+++ b/packages/camlbz2/camlbz2.0.6.0/opam
@@ -4,21 +4,27 @@ authors: [
   "Olivier Andrieu"
   "Stefano Zacchiroli"
 ]
-homepage: "http://camlbz2.forge.ocamlcore.org/"
+homepage: "https://gitlab.com/irill/camlbz2"
 license: "LGPL-2.1-or-later with OCaml-LGPL-linking-exception"
+bug-reports: "https://gitlab.com/irill/camlbz2/issues"
+dev-repo: "git+https://gitlab.com/irill/camlbz2.git"
 build: [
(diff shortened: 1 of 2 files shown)


josch:
Thanks, @camelus! I added the missing fields and also tried to run aclocal and autoconf based on what I saw in opam files of other projects. It seems to work for Debian and Ubuntu but on on other distros bzlib.h is not available. Sadly, I don't know my way around in these other distributions.

kit-ty-kate:
I'm not against this change but Gitlab allows custom tarballs to be uploaded. Since the original tarball still exists could it be maybe downloaded then re-uploaded on Gitlab on the correct tag? This way users would not have to have to install autoconf & aclocal.

If the author is not available to do that I'll go ahead and merge this but ideally I would prefer to only merge the addition of the conf-libbz2 package.

josch:
Hi @kit-ty-kate, thanks for your help! :+1: 

I'm quite hesitant about shipping the autoconf output with the release tarball. Please convince me that it is a good idea.

In Debian, we have many upstream projects that ship the autotools results in their release tarballs. Before building the package, we remove those and then regenerate them so that:

 - we benefit from improvements to autotools
 - we can make sure that the existing configure.ac still works
 - we definitely ignore any possible manual patches to configure
 - we make sure that we have built from source instead of relying on non-source artifacts

Are those not also goals shared by opam as a software distribution? What would definitely convince me to add the generated files is, if there was some opam policy that requires projects to ship generated files but I see that other packages in opam are also relying on `conf-autoconf`.

So this is a question of what actually belongs into a release tarball. From my point of view, only the sources belong in it and not the generated build results and configure files are not source but the results of parts of the build process. If the argument is "we want to reduce the dependencies" then why do we not also ship the ELF libraries which are also produced by the build? But we would not call the result a source tarball.

As for the "dependency reduction" argument one can also counter, that for example in Debian, [28% of our users](https://qa.debian.org/popcon.php?package=autoconf) already have autoconf installed.

avsm:
> So this is a question of what actually belongs into a release tarball.  

From a traditional "UNIX packaging perspective", it should include the autogenerated files so that "./configure && make && make install" just work. The OCaml compiler, for example, does this.  It used to be the case that this was more portable since the output of autotools has minimal requirements, whereas finding a specific version of autotools to generate the right runes on (e.g. Solaris) was quite a painful process. With the low pace of releases of autotools these days, that matters less.

Your point above about "why not ship the ELF libraries also produced' is a false equivalence to the autoconf situation. The output of autoconf is _designed_ to be included for portable bootstrap on some unusual systems with just a C compiler and a POSIX-ish shell installed, whereas binary outputs are obviously not portable (many of the systems it will compile on do not even use ELF).  Autotools are at the very beginning of a compilation cycle, and binaries are near the end.

From an opam repository perspective though, we don't have an opinion -- do as you will, as long as it works for a source-based build with external dependencies installed.  Our CI tests all those cases so you'll know soon enough if all the various `autoconf` packages on different distros are sufficiently to interpret the config.ac.

josch:
Hi @avsm and thanks for your input! So this comment clearly has nothing to do any more with camlbz2 and the situation in this pull request but maybe this is interesting for the reader anyways. Yes, the "traditional UNIX packaging perspective" included `configure` and yes, this used to be because it increased portability. Unfortunately, these days it's the other way round and that's why the ELF example I used is not that wrong, I think. So let me explain this a bit more. I'm working in the bootstrapping team in Debian. In Debian we bootstrap roughly one new architecture per year. On every port we do, the biggest blocker of why a new package does not compile on the new architecture is `configure` files shipped by upstream. The last port for which I have hard numbers is the Debian port to arm64. Out ouf about 10k packages, 574 failed to build on arm64. And out of those build failures, 60% were because `configure` was not regenerated during the build but the existing version shipped by the project was used which was using an outdated config.guess and other bits the new architecture needed. So coming back to my analogy with shipping ELF binaries: when a new architecture comes along, then obviously the existing ELF binaries don't help us. We need to build new ones from source. But the same holds for `./configure`. When a new architecture comes along, we need to run `autoconf` again and produce a new `./configure`. So yeah, in the past (way before I got involved in all of this) shipping a pre-compiled configure was used for portability. But for modern bootstrapping purposes to new architectures, the existing `./configure` is just as useless as pre-built ELF binaries would be. I hope this makes my comparison more clear. :)

Now because we do lots of ports in Debian (I think porting all our 32k packages to a new architecture once per year is a lot) we adjusted more and more packages to just throw any possibly existing `configure` away and run autoreconf on every build. This helps bootstrapping new architectures plus has all the benefits I listed in my post above. So these days, when we do new ports, things look much better than 60% of our packages failing due to autoconf. But this brings me back to camlbz2. I will also be maintaining camlbz2 in Debian and if I let the camlbz2 upstream package ship a pre-generated `configure`, then I will have to do extra work in Debian to patch it out, which is not ideal.

I'll talk to my colleagues and sleep over this. Thank you @avsm and @kit-ty-kate for your help and input! It is very much appreciated. :)

josch:
Hi,

so I thought more about the idea of uploading custom tarballs to releases on gitlab.com as suggested by @kit-ty-kate. I think the custom tarballs can serve a valid purpose, which is not to destroy history. What this pull request did until recently was to "change the past" by uploading a past release as a tarball with different contents. This is not nice. I now downloaded the old tarballs from forge.ocamlcore.org and uploaded them to gitlab.com as attachments to the release tags. That way, the old release should be bit-by-bit preserved now.

Additionally, I made a new release. I decided against adding the generated configure into the tarball. I must admit maintaining the package in Debian also influenced my decision. I hope the opam world is not angry at me about this.

Since there is a new release now, this pull request is not anymore adding a different hash for an old release (which would not be ideal) but instead directly refers to the new release in addition to updating the URLs in the metadata to gitlab.com. Would you like me to split up the commit into the part that changes the upstream data and the part that updates the version or is it fine now as things are?

Also, sorry for the botched rebase and force push as you can see in the history of this pull request.

Thanks!

kit-ty-kate:
Thanks a lot! Could you release 0.6.1 separately?

kit-ty-kate:
(as in, in a separate PR, sorry if that wasn't clear)

josch:
Thanks a lot for your help @kit-ty-kate! I just opened #17990

PR #17942, opened 2021-01-08
Add zelus and  zelus-gtk 2.1 packages

gbdrt:
These packages contains the Zelus compiler (package zelus) and optional gtk dependent librairies (zelus-gtk)
https://zelus.di.ens.fr/

### DIFF
--- a/packages/zelus-gtk/zelus-gtk.2.1/opam
+++ b/packages/zelus-gtk/zelus-gtk.2.1/opam
@@ -1,6 +1,9 @@
 opam-version: "2.0"
-synopsis: "A synchronous language with ODEs"
-authors: ["Marc Pouzet <marc.pouzet@ens.fr>"]
+synopsis: "Zelus GTK library"
+authors: [
+  "Timothy Bourke <timothy.bourke@inria.fr>"
+  "Marc Pouzet <marc.pouzet@ens.fr>"
+  ]
 maintainer: ["Marc Pouzet <marc.pouzet@ens.fr>"]
 homepage: "http://zelus.di.ens.fr"
 doc: "http://zelus.di.ens.fr/man/"
@@ -10,7 +13,6 @@ depends: [
   "dune" {>= "2.6"}
   "zelus" {= version}
   "lablgtk"
-  "conf-gtksourceview"
 ]
 build: [
   ["dune" "subst"] {dev}
--- a/packages/zelus-gtk/zelus-gtk.2.1/opam
+++ b/packages/zelus-gtk/zelus-gtk.2.1/opam
@@ -11,6 +11,13 @@ depends: [
   "zelus" "lablgtk"
 ]
 depopts: ["glmlite"]
+depexts: [
+  [
+    "libx11"
+    "m4"
+    "pkgconf"
+  ] {os-distribution = "arch"}
+]
 build: [
   ["dune" "subst"] {pinned}
   [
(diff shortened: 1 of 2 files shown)


gbdrt:
I don't understand why the build keeps failing on archlinux. I tried to add the missing dependencies, but:
- I don't understand why these packages are required
- The build is still failing with these explicit depexts.

Am I missing something?

gbdrt:
Ok, making progress...
I removed the graphics dependencies, but now it is complaining about a missing gtk+.
Do I need to add it as a depexts? It looks like other packages depending on lablgtk do not need this.

Thanks for you help.

packages/zelus/zelus.2.1/opam
```diff
@@ -0,0 +1,31 @@
+opam-version: "2.0"
+name: "zelus"
+version: "2.1"
+synopsis: "A synchronous language with ODEs"
+authors: ["Marc Pouzet <marc.pouzet@ens.fr>"]
+maintainer: ["Marc Pouzet <marc.pouzet@ens.fr>"]
+homepage: "http://zelus.di.ens.fr"
+doc: "http://zelus.di.ens.fr/man/"
+bug-reports: "https://gitlab.inria.fr/parkas/zelus/-/issues"
```
kit-ty-kate, line 9: this link is not accessible

packages/zelus/zelus.2.1/opam
```diff
@@ -0,0 +1,31 @@
...
+depends: [
+  "ocaml" {>= "4.08.1"}
+  "dune" {>= "2.6" & build}
```
kit-ty-kate: dune should not be tagged as a build dependency (see https://github.com/ocaml/dune/issues/2147)

packages/zelus-gtk/zelus-gtk.2.1/opam
```diff
@@ -0,0 +1,26 @@
...
+synopsis: "A synchronous language with ODEs"
+authors: ["Marc Pouzet <marc.pouzet@ens.fr>"]
+maintainer: ["Marc Pouzet <marc.pouzet@ens.fr>"]
+homepage: "http://zelus.di.ens.fr"
+doc: "http://zelus.di.ens.fr/man/"
+bug-reports: "https://github.com/INRIA/zelus/issues"
+depends: [
+  "ocaml" {>= "4.08.1"}
+  "dune" {>= "2.6"}
+  "zelus" {= version}
```
kit-ty-kate, line 11: If you are planning to release the two packages together, I would highly encourage you to use `{= version}`. If not, feel free to revert this change.

packages/zelus-gtk/zelus-gtk.2.1/opam
```diff
@@ -0,0 +1,26 @@
...
+  "lablgtk"
+  "conf-gtksourceview"
```
kit-ty-kate, line 13: gtksourceview-2.0 is getting really old and is not supported by an increasing number of distributions, it might be more future-proof to switch to gtksourceview3 in next release

kit-ty-kate:
The PR should be good to go after that. Could you review my change and return them in the opam file upstream if they seem correct to you?

gbdrt:
Thanks a lot!
I just pushed the updated version (with small fixes in the descriptions).
I also removed the dependencies to gtksourceview. We do not really need it, I wanted to see if it could solve the build failure on archlinux.

kit-ty-kate:
Looks good now, thanks! (note that gtk2 itself is also quite old, though better supported across distributions than its sublibraries)

PR #17947, opened 2021-01-08
Tezos 8.1

pirbo:
This is a bugfix release. Nothing has changed in the packaging since v8.0

### DIFF
--- a/packages/tezos-client/tezos-client.8.1/opam
+++ b/packages/tezos-client/tezos-client.8.1/opam
@@ -21,17 +21,30 @@ depends: [
   "tezos-baking-008-PtEdoTez-commands" { = version }
 ]
 depopts: [
-  "tezos-client-genesis" { = version }
-  "tezos-client-demo-counter" { = version }
-  "tezos-client-alpha-commands-registration" { = version }
-  "tezos-baking-alpha-commands" { = version }
+  "tezos-client-genesis"
+  "tezos-client-demo-counter"
+  "tezos-client-alpha-commands-registration"
+  "tezos-baking-alpha-commands"
 
-  "tezos-client-001-PtCJ7pwo-commands" { = version }
-  "tezos-client-002-PsYLVpVv-commands" { = version }
-  "tezos-client-003-PsddFKi3-commands" { = version }
-  "tezos-client-004-Pt24m4xi-commands" { = version }
-  "tezos-client-005-PsBabyM1-commands" { = version }
-  "tezos-client-006-PsCARTHA-commands" { = version }
+  "tezos-client-001-PtCJ7pwo-commands"
+  "tezos-client-002-PsYLVpVv-commands"
+  "tezos-client-003-PsddFKi3-commands"
+  "tezos-client-004-Pt24m4xi-commands"
+  "tezos-client-005-PsBabyM1-commands"
+  "tezos-client-006-PsCARTHA-commands"
+]
+conflicts: [
+  "tezos-client-genesis" { != version }
+  "tezos-client-demo-counter" { != version }
+  "tezos-client-alpha-commands-registration" { != version }
+  "tezos-baking-alpha-commands" { != version }
+
+  "tezos-client-001-PtCJ7pwo-commands" { != version }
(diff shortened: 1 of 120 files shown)


avsm:
One possible new test failure; rest all looks fine:

```
# context              2.0.7 | linux/x86_64 | ocaml-base-compiler.4.11.1 | pinned(https://gitlab.com/tezos/tezos/-/archive/v8.1/tezos-v8.1.tar.bz2)
# path                 ~/.opam/4.11/.opam-switch/build/tezos-stdlib-unix.8.1
# command              ~/.opam/4.11/bin/dune runtest -p tezos-stdlib-unix -j 31
# exit-code            1
# env-file             ~/.opam/log/tezos-stdlib-unix-18-d8a4ae.env
# output-file          ~/.opam/log/tezos-stdlib-unix-18-d8a4ae.out
### output ###
# test_lwt_exit_range_of_signals alias src/lib_stdlib_unix/test/runtest_test_lwt_exit_range_of_signals (exit 2)
# (cd _build/default/src/lib_stdlib_unix/test && ./test_lwt_exit_range_of_signals.exe)
# Fatal error: exception Failure("WSIGNALED(-2) on ALRM")
```

pirbo:
Well no actually, we noticed that one last time and I forgot about it :-/
It is nondeterministic (and looked related to "the load" of the machine) I clicked retry and it worked... As it did last time... I need to not forget to do something about it for next time though.

packages/tezos-client/tezos-client.8.1/opam
```diff
@@ -0,0 +1,50 @@
...
+  "tezos-client-alpha-commands-registration" { = version }
+  "tezos-baking-alpha-commands" { = version }
+
+  "tezos-client-001-PtCJ7pwo-commands" { = version }
+  "tezos-client-002-PsYLVpVv-commands" { = version }
+  "tezos-client-003-PsddFKi3-commands" { = version }
+  "tezos-client-004-Pt24m4xi-commands" { = version }
+  "tezos-client-005-PsBabyM1-commands" { = version }
+  "tezos-client-006-PsCARTHA-commands" { = version }
+]
```
kit-ty-kate, line 35: Constraints in the `depopts` field have a different behaviour compared to the `conflicts` field.
The former allows the depopts to be installed alongside this package if they don't have the same version. This is obviously wrong in most cases.

In this particular case it might be held up together by one thread as everything is constrained down to some tezos base package

Could you use the `conflicts` field instead for this package and others like it which do the same thing?

packages/tezos-lwt-result-stdlib/tezos-lwt-result-stdlib.8.1/opam
```diff
@@ -0,0 +1,28 @@
+opam-version: "2.0"
+maintainer: "contact@tezos.com"
+authors: [ "Tezos devteam" ]
+homepage: "https://www.tezos.com/"
+bug-reports: "https://gitlab.com/tezos/tezos/issues"
+dev-repo: "git+https://gitlab.com/tezos/tezos.git"
+license: "MIT"
+depends: [
+  "ocaml" { < "4.12" }
```
kit-ty-kate, line 9: Quick question: The release of OCaml 4.12 is getting close. Is there any chance of seeing this unlocked?

packages/tezos-client/tezos-client.8.1/opam
pirbo: Done, Thank you

packages/tezos-lwt-result-stdlib/tezos-lwt-result-stdlib.8.1/opam
pirbo: `tezos-protocol-compiler` has strong constraint on the version of the OCaml compiler it uses and for now, development is still in 4.09!
We are already compatible with 4.10 and we'll work in the next month to move all the dev to OCaml 4.11. Maybe, once this done and we don't have anymore to be compatible with 4.09, we will be able to enable 4.12 compatibility here yes.

kit-ty-kate:
Thanks! (sorry for the wait)

PR #17776, opened 2020-12-06
Add the wasmtime package

LaurentMazare:
This PR adds two new packages:
- wasmtime: a package wrapping the [wasmtime c library](https://github.com/bytecodealliance/wasmtime) so that it can be used from OCaml. This makes it possible to run wasm files, including system calls via wasi, callbacks, etc.
- libwasmtime: this ships the c library (as it is not included in distributions at the moment).

PR #17948, opened 2021-01-08
Add libwasmtime 0.22

LaurentMazare:
Add the newly released libwasmtime 0.22 packages.

### DIFF
--- a/packages/libwasmtime/libwasmtime.0.22.0+linux-x86_64/opam
+++ b/packages/libwasmtime/libwasmtime.0.22.0+linux-x86_64/opam
@@ -1,0 +1,23 @@
+opam-version: "2.0"
+homepage: "https://github.com/LaurentMazare/ocaml-wasmtime"
+maintainer: "lmazare@gmail.com"
+bug-reports: "https://github.com/LaurentMazare/ocaml-wasmtime/issues"
+authors: [
+  "Laurent Mazare"
+]
+install: [
+  [
+    "sh"
+    "-c"
+    "test -d %{lib}%/libwasmtime/lib/libwasmtime.a || ( mkdir -p %{lib}%/libwasmtime && tar xf wasmtime-linux.tar.xz && mv -f wasmtime-*linux*/* %{lib}%/libwasmtime )"
+  ]
+]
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
Hi, it might have missed that in https://github.com/ocaml/opam-repository/pull/17776, but what is the reason to ship binary packages? Opam tries to be a source-based package manager as much as possible and shipping binaries might cause more issues. The `libtorch` package already fails on major distributions because of link-time incompatibilities.

Is it because rust-based packages are hard to port? If so, recently there was an example of a successful rust port setup in opam by the tezos team: https://github.com/ocaml/opam-repository/blob/master/packages/tezos-rust-libs/tezos-rust-libs.1.0/opam

It's not a blocker but I'd like to make sure before merging.

LaurentMazare:
@kit-ty-kate the main reason for the binary packages is simplicity on my side. I imagine that it may be be possible to create packages that would build the library using rust/cargo but I'm not sure how much work this is compared to leveraging the already bundled binaries (and it would mean that users have to download the whole rust toolchain for this I imagine).
It's hard for me to tell whether it's worth it, in my case these solve my use case, but maybe that creates a lot of hassle to others.

kit-ty-kate:
> @kit-ty-kate the main reason for the binary packages is simplicity on my side. I imagine that it may be be possible to create packages that would build the library using rust/cargo but I'm not sure how much work this is compared to leveraging the already bundled binaries (and it would mean that users have to download the whole rust toolchain for this I imagine).
> It's hard for me to tell whether it's worth it, in my case these solve my use case, but maybe that creates a lot of hassle to others.

I've tested locally and it seems feasible for personal projects, less for big outside libraries.
In the case of `tezos-rust-libs` they:
* vendor all of their libraries
* use the vendor directory as their main repository using a local cargo config (to avoid having cargo pull the main repository on the network)

In your case, there are few vendored libraries (34) and 80 more would have to be vendored. In terms of build time, my local development station compiled everything (including all the rust dependencies) in under 4 minutes.

So all in all, this is possible but would require a light-fork to be viable.
Sorry for the wait, this is good.

PR #17956, opened 2021-01-11
[new release] qbf (0.3)

maelvls:
CHANGES:

### Added

- Moved to [Dune](https://github.com/ocaml/dune) which has become the
  de-facto OCaml build tool. We were previously using Oasis, which was
  using Ocamlbuild under the hood and made the builds slightly slower.

### Removed

- `depqbf` support has been temporarily removed due to the move to Dune. A
  PR to re-introduce `depqbf` in Dune would be welcome!

### Changed

- OCaml 4.08.x is now the minimum version required to run this library.

### Fixed

- Fixed the `x86_64` build on Windows + Cygwin64 + the mingw64 toolchain.
- Fixed building ocaml-qbf using opam 2. Opam 2 is sandboxing builds and
  do not accept messing with /tmp.

[Unreleased]: https://github.com/c-cube/ocaml-qbf/compare/0.3...HEAD
[0.3]: https://github.com/c-cube/ocaml-qbf/compare/0.2..0.3
[0.2]: https://github.com/c-cube/ocaml-qbf/releases/0.2


maelvls:
Build failed for an older version of `touist` that apparently did not specify `cppo` as a build dependency 😓

Should I send a PR to fix the 3.4.1 opam file of `touist`?

packages/qbf/qbf.0.3/opam
```diff
@@ -0,0 +1,41 @@
...
+authors: [ "Simon Cruanes" ]
+license: "BSD-3-clause"
+homepage: "https://github.com/c-cube/ocaml-qbf"
+dev-repo: "git+https://github.com/c-cube/ocaml-qbf.git"
+bug-reports: "https://github.com/c-cube/ocaml-qbf/issues"
+tags: [ "clib:quantor" "clib:qdpll" "clib:picosat"  ]
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "build" "@doc" "-p" name ] {with-doc}
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test & ocaml:version < "4.11"}
```
kit-ty-kate, line 12: What's the reason behind not testing on 4.11?

kit-ty-kate (commented):
> Should have I put instead:
> 
> ```
>  "cppo_ocamlbuild" {build & >= "1.6.0"} 
>  "cppo" {build & >= "0.9.4" & <= "1.5.0"}
> ```
> 
> ?

Indeed. I just fixed that.

packages/qbf/qbf.0.3/opam
maelvls: Not sure, @c-cube might know more about that 😅

kit-ty-kate: ok, let's try. I'll remove my change if there is an issue

kit-ty-kate: It seems to be passing without any issues. Could you return the fix upstream?

kit-ty-kate:
All green. Thanks!

packages/qbf/qbf.0.3/opam
maelvls: Will do!

Issue #1224, opened 2013-10-20
Unable to install Lwt after having pinned it.


vbmithr:
```
[vb@nerenyi ~/code/opam-repository/packages/lwt.2.4.3]% opam install lwt                                          
The following actions will be performed:
 - install lwt.2.4.3
1 to install | 0 to reinstall | 0 to upgrade | 0 to downgrade | 0 to remove
=-=-= Installing lwt.2.4.3 =-=-=
Building lwt.2.4.3:
  patch -p1 -i 4.01cloexec.diff
  ./configure --enable-libev --enable-react --enable-ssl --enable-unix --enable-extra --enable-preemptive --disabl
e-glib --disable-text
  make build
  make install
The compilation of lwt.2.4.3 failed.
Removing lwt.2.4.3.
  ocamlfind remove lwt
==== ERROR [while installing lwt.2.4.3] ====
# opam-version    1.0.1 (4231ef3f9078199633942325436ecca2e805f26f)
# os              linux
# command         patch -p1 -i 4.01cloexec.diff
# path            /home/vb/.opam/4.01.0/build/lwt.2.4.3
# compiler        4.01.0
# exit-code       2
# env-file        /home/vb/.opam/4.01.0/build/lwt.2.4.3/lwt-10198-7afd23.env
# stdout-file     /home/vb/.opam/4.01.0/build/lwt.2.4.3/lwt-10198-7afd23.out
# stderr-file     /home/vb/.opam/4.01.0/build/lwt.2.4.3/lwt-10198-7afd23.err
### stderr ###
patch: **** Can't open patch file 4.01cloexec.diff : No such file or directory
'opam install lwt' failed.
```

Issue #1218, opened 2013-10-16
ocaml-zmq won't build


hcarty:
The ocaml-zmq.0 package is currently just a pointer to the master branch of the upstream repository.  Some commits have been made to the repository which change the download checksum and break the build.

The package should probably be changed to point to the recently created 1.0.0 branch of the repository.

Issue #1189, opened 2013-10-08
add bindings to berkeley DB


UnixJunkie:

samoht:
Please see with upstream.

Issue #1169, opened 2013-10-06
ocamlnet.3.7.3 errors


kindnation:
I'm trying to install google drive ocamlfuse with opam, and it requires ocamlnet 3.7.3, but it exits with errors in the file:

/.opam/system/build/ocamlnet.3.7.3/src/netstring/netencoding.mli

File "netencoding.mli", line 117, characters 1-3
Error: This comment contains an unterminated string literal
make[1]: *** [depend] Error 1
make: *** [all] Error 2

I tried compiling ocamlnet version 3.7.3 manually and can confirm the bug.  There must be some hidden characters because it looks like a normal comment so I just removed the comments from 117 and a few lines below it from netencoding.mli and then it is able to compile successfully.

samoht:
Please re-open if newer version of ocamlnet didn't fix your problem.

PR #17960, opened 2021-01-12
[new release] irmin et al. (2.3.0)

CraigFe:
- Project page: <a href="https://github.com/mirage/irmin">https://github.com/mirage/irmin</a>

### Fixed

- **irmin-git**
  - Update `irmin` to the last version of `ocaml-git` (#1065)
    It fixes an issue on serialization/deserialization of big tree object
    (see #1001) 

- **irmin-pack***
  - Fix a major bug in the LRU which was never used (#1035, @samoht)

- **irmin***
  - Improve performance of `last_modified` (#948, @pascutto)

  - Changed the pattern matching of the function `last_modified`. The case of a
    created key is now considered a modification by the function. (#1167,
    @clecat)

  - Make Tree.clear tail-recursive (#1171, @samoht)

  - Fix `Tree.fold ~force:(False f)` where results where partially skipped
    (#1174, @Ngoguey42, @samoht and @CraigFe )

  - Fix `Tree.kind`. Empty path on a tree used to return a None instead of a
	`` `Node``. (#1218, @Ngoguey42)

- **ppx_irmin**
  - Fix a bug causing certain type derivations to be incorrect due to unsound
    namespacing. (#1083, @CraigFe)

- **irmin-unix**
  - Update irmin config path to respect `XDG_CONFIG_HOME`. (#1168, @zshipko)

#### Added

- **irmin-layers** (_new_):
  - Created a new package, `irmin-layers` that includes common signatures for
    layered stores. It contains a stub `Make_layers` functor (#882, @icristescu)

- **irmin-bench** (_new_):
  - Created a new package to contain benchmarks for Irmin and its various
    backends. (#1142, @CraigFe)
  - Added ability to get json output and a make target to run layers benchmark.
    (#1146, @gs0510)

- **irmin**
  - Added `Tree.{Contents,Node}` modules exposing operations over lazy tree
    contents and nodes respectively. (#1022, @CraigFe)

  - Added `Type.Unboxed.{encode_bin,decode_bin,size_of}` to work with unboxed
    values (#1030, @samoht)

  - Remove the `headers` option in `Type.{encode_bin,decode_bin,size_of}`. Use
    `Type.Unboxed.<fn>` instead (#1030, @samoht)

  - `Type.v` now takes an extra mandatory `unit` argument (#1030, @samoht)

  - Added `Type.pp_dump`, which provides a way to pretty-print values with a
    syntax almost identical to native OCaml syntax, so that they can easily be
    copy-pasted into an OCaml REPL for inspection. (#1046, @liautaud)

  - Generic functions in `Irmin.Type` are now more efficient when a partial
    closure is constructed to the type representation (#1030 #1093, @samoht
    @CraigFe).  To make this even more explicit, these functions are now staged
    and `Type.{unstage,stage}` can manipulate these. The goal is to encourage
    users to write this kind of (efficent) pattern:
    ```ocaml
    let encode_bin = Type.(unstage (encode_bin ty))
    let _ = <begin loop> ... encode_bin foo ... <end loop>
    ```
  - Added a `clear` function for stores (#1071, @icristescu, @CraigFe)

  - Requires digestif>=0.9 to use digestif's default variants
    (#873, @pascutto, @samoht)

  - Added `iter_commits` and `iter_nodes` functions to traverse the commits and
    nodes graphs (#1077, @icristescu)

  - Added `Repo.iter` to traverse object graphs (#1128, @samoht)

- **irmin-pack**:
  - Added `index_throttle` option to `Irmin_pack.config`, which exposes the
    memory throttle feature of `Index` in `Irmin-Pack`. (#1049, @icristescu)

  - Added `Pack.clear` and `Dict.clear` (#1047, @icristescu, @CraigFe, @samoht)

  - Added a `migrate` function for upgrading stores with old formats (#1070,
    @icristescu, @CraigFe)

  - Added a `flush` function for a repo (#1092, @icristescu)

  - Added `Layered.Make functor, to construct layered stores from irmin-pack.
    (#882, @icristescu)

  - Added `Checks.Make which provides some offline checks for irmin-pack
    stores. (#1117, @icristescu, @CraigFe)

  - Added `reconstruct_index` to reconstruct an index from a pack file. (#1097,
    @icristescu)

  - Added `reconstruct-index` command to `irmin-fsck` for reconstructing an index from
    the command line (#1189, @zshipko)

  - Added `integrity-check` command to `irmin-fsck` for checking the integrity of
    an `irmin-pack` store (#1196, @zshipko)

- **ppx_irmin**:

  - Added support for deriving type representations for types with type
    parameters. Type `'a t` generates a representation of type
    `'a Type.t -> 'a t Type.t` (#1085, @CraigFe)

  - Added a `--lib` command-line option which has the same behaviour as the
    `lib` run-time argument (i.e. `--lib Foo` will cause `ppx_irmin` to derive
    type representations using combinators in the `Foo` module). (#1086,
    @CraigFe)

  - Added an extension point `[typ: <core-type>]` for deriving type
    representations inline. (#1087, @CraigFe)

#### Changed

- **irmin**
  - Renamed the `Tree.tree` type to `Tree.t`. (#1022, @CraigFe)

  - Replaced `Tree.pp_stats` with the type representation `Tree.stats_t`. (#TODO, @CraigFe)

  - Changed the JSON encoding of special floats. `Float.nan`, `Float.infinity`
    and `Float.neg_infinity` are now encoded as `"nan"`, `"inf"` and `"-inf"`
    respectively. (#979, @liautaud)

  - The functions `Type.{v,like,map}` no longer take a `~cli` argument, and now
    take separate `~pp` and `~of_string` arguments instead. (#1103, @CraigFe)

  - The `Irmin.Type` combinators are now supplied by the `repr` package. The
    API of `Irmin.Type` is not changed. (#1106, @CraigFe)

  - `Irmin.Type` uses staging for `equal`, `short_hash` and `compare` to
    speed-up generic operations (#1130, #1131, #1132, @samoht)

  - Make `Tree.fold` more expressive and ensure it uses a bounded memory
    (#1169, @samoht)

- **irmin-pack**:
  - `sync` has to be called by the read-only instance to synchronise with the
    files on disk. (#1008, @icristescu)

  - Renamed `sync` to `flush` for the operation that flushes to disk all buffers
    of a read-write instance. (#1008, @icristescu)

  - Changed the format of headers for the files on disk to include a generation
    number. Version 1 of irmin-pack was used for the previous format, version 2
    is used with the new format. (#1047, @icristescu, @CraigFe, @samoht)

  - Use `Repo.iter` to speed-up copies between layers (#1149, #1150 @samoht)

  - Add an option to bypass data integrity checks on reads (#1154, @samoht)

  - Add `heads` parameter to `check-self-contained` command in `Checks` (#1224, @zshipko)

- **ppx_irmin**:

  - The `[@generic ...]` attribute has been renamed to `[@repr ...]`. (#1082,
    @CraigFe)

### DIFF
--- a/packages/irmin/irmin.2.3.0/opam
+++ b/packages/irmin/irmin.2.3.0/opam
@@ -1,0 +1,55 @@
+opam-version: "2.0"
+maintainer:   "thomas@gazagnaire.org"
+authors:      ["Thomas Gazagnaire" "Thomas Leonard"]
+license:      "ISC"
+homepage:     "https://github.com/mirage/irmin"
+bug-reports:  "https://github.com/mirage/irmin/issues"
+dev-repo:     "git+https://github.com/mirage/irmin.git"
+doc:          "https://mirage.github.io/irmin/"
+
+build: [
+ ["dune" "subst"] {pinned}
+ ["dune" "build" "-p" name "-j" jobs]
+ ["dune" "runtest" "-p" name] {with-test}
+]
+
+depends: [
+  "ocaml"   {>= "4.07.0"}
+  "dune"    {>= "2.7.0"}
+  "repr"    {>= "0.2.0"}
+  "fmt"     {>= "0.8.0"}
+  "uri"     {>= "1.3.12"}
+  "uutf"
+  "jsonm"   {>= "1.0.0"}
+  "lwt"     {>= "2.4.7"}
+  "base64"  {>= "2.0.0"}
+  "digestif" {>= "0.9.0"}
+  "ocamlgraph"
+  "logs"    {>= "0.5.0"}
+  "bheap" {>= "2.0.0"}
+  "astring"
+  "ppx_irmin" {= version}
+  "hex"      {with-test}
+  "alcotest" {>= "1.1.0" & with-test}
+  "alcotest-lwt" {with-test}
+  "irmin-mem" {with-test & post & = version}
+]
+synopsis: """
+Irmin, a distributed database that follows the same design principles as Git
+"""
+description: """
+Irmin is a library for persistent stores with built-in snapshot,
+branching and reverting mechanisms. It is designed to use a large
+variety of backends. Irmin is written in pure OCaml and does not
+depend on external C stubs; it aims to run everywhere, from Linux,
+to browsers and Xen unikernels.
+"""
+x-commit-hash: "3e22238e86b6bac0c4eddff5b899efeaebe49dc9"
+url {
+  src:
+    "https://github.com/mirage/irmin/releases/download/2.3.0/irmin-2.3.0.tbz"
+  checksum: [
+    "sha256=41d1d08cedb19d17935b7e840f18265ebd8b8a50cc3797d031aee992125070a6"
+    "sha512=ff527303850c2e36853ae0857f2ecbc22b5096f87197521dd049f80a234228c74f840aadef6d7c826acceda7b0a49e1dc2b5c105bb65d7c9551f0f8f377b2fc6"
+  ]
+}
--- a/packages/irmin/irmin.2.3.0/opam
+++ b/packages/irmin/irmin.2.3.0/opam
@@ -8,9 +8,9 @@ dev-repo:     "git+https://github.com/mirage/irmin.git"
 doc:          "https://mirage.github.io/irmin/"
 
 build: [
- ["dune" "subst"] {pinned}
+ ["dune" "subst"] {dev}
  ["dune" "build" "-p" name "-j" jobs]
- ["dune" "runtest" "-p" name] {with-test}
+ ["dune" "runtest" "-p" name "-j" jobs] {with-test}
 ]
 
 depends: [
--- a/packages/irmin/irmin.2.3.0/opam
+++ b/packages/irmin/irmin.2.3.0/opam
@@ -10,7 +10,6 @@ doc:          "https://mirage.github.io/irmin/"
 build: [
  ["dune" "subst"] {dev}
  ["dune" "build" "-p" name "-j" jobs]
- ["dune" "runtest" "-p" name "-j" jobs] {with-test}
 ]
 
 depends: [
@@ -29,10 +28,6 @@ depends: [
   "bheap" {>= "2.0.0"}
   "astring"
   "ppx_irmin" {= version}
-  "hex"      {with-test}
-  "alcotest" {>= "1.1.0" & with-test}
-  "alcotest-lwt" {with-test}
-  "irmin-mem" {with-test & post & = version}
 ]
 synopsis: """
 Irmin, a distributed database that follows the same design principles as Git
--- a/packages/ppx_irmin/ppx_irmin.2.3.0/opam
+++ b/packages/ppx_irmin/ppx_irmin.2.3.0/opam
@@ -1,0 +1,32 @@
+opam-version: "2.0"
+maintainer: "Craig Ferguson <craig@tarides.com>"
+homepage: "https://github.com/mirage/irmin"
+bug-reports: "https://github.com/mirage/irmin/issues"
+license: "ISC"
+dev-repo: "git+https://github.com/mirage/irmin.git"
+
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+]
+
+depends: [
+  "dune" {>= "2.7.0"}
+  "ocaml" {>= "4.08.0"}
+  "ppxlib" {>= "0.12.0" & < "0.18.0"}
+  "ppx_repr" {>= "0.2.0"}
+  "irmin" {with-test & post & = version}
+]
+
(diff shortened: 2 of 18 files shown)


kit-ty-kate:
The tests failed with:
```
#=== ERROR while compiling irmin-chunk.2.3.0 ==================================#
# context              2.1.0~beta4 | linux/arm32 | ocaml-base-compiler.4.11.1 | pinned(https://github.com/mirage/irmin/releases/download/2.3.0/irmin-2.3.0.tbz)
# path                 ~/.opam/4.11/.opam-switch/build/irmin-chunk.2.3.0
# command              ~/.opam/4.11/bin/dune runtest -p irmin-chunk
# exit-code            1
# env-file             ~/.opam/log/irmin-chunk-24-4652d0.env
# output-file          ~/.opam/log/irmin-chunk-24-4652d0.out
### output ###
# File "test/irmin-pack/cli/dune", line 4, characters 23-41:
# 4 |  (libraries irmin-pack irmin-pack.layered))
#                            ^^^^^^^^^^^^^^^^^^
# Error: Library "irmin-pack.layered" not found.
# Hint: try:
#   dune external-lib-deps --missing -p irmin-chunk @runtest
# File "test/irmin-pack/cli/dune", line 9, characters 23-41:
# 9 |  (libraries irmin-pack irmin-pack.layered))
#                            ^^^^^^^^^^^^^^^^^^
# Error: Library "irmin-pack.layered" not found.
# Hint: try:
#   dune external-lib-deps --missing -p irmin-chunk @runtest
#         test alias test/irmin-chunk/runtest
# Testing `irmin-chunk'.
```

packages/ppx_irmin/ppx_irmin.2.3.0/opam
```diff
@@ -0,0 +1,32 @@
...
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+]
+
+depends: [
+  "dune" {>= "2.7.0"}
+  "ocaml" {>= "4.08.0"}
+  "ppxlib" {>= "0.12.0" & < "0.18.0"}
```
kit-ty-kate, line 17: I don't remember seeing the previous version of ppx_irmin incompatible with ppxlib >= 0.18.0. Is that expected? This is required for a proper compatibility with OCaml 4.12.

packages/irmin/irmin.2.3.0/opam
```diff
@@ -0,0 +1,55 @@
...
+license:      "ISC"
+homepage:     "https://github.com/mirage/irmin"
+bug-reports:  "https://github.com/mirage/irmin/issues"
+dev-repo:     "git+https://github.com/mirage/irmin.git"
+doc:          "https://mirage.github.io/irmin/"
+
+build: [
+ ["dune" "subst"] {pinned}
+ ["dune" "build" "-p" name "-j" jobs]
+ ["dune" "runtest" "-p" name] {with-test}
```
kit-ty-kate, line 13: If there is a change to make, would it be possible to change the `{pinned}` to `{dev}` and to add `-j jobs` to the tests in all of the opam files?

packages/ppx_irmin/ppx_irmin.2.3.0/opam
CraigFe: Probably just because `ppxlib.0.18` wasn't out at the last Irmin release; this constraint was added to prior versions during the `0.18` Ppxlib release https://github.com/ocaml/opam-repository/pull/17360/files.

We'd require a cleverer fix to drop this upper bound, since we can't afford to lose compatibility with pre-0.18 versions of ppxlib at the moment.

kit-ty-kate: > We'd require a cleverer fix to drop this upper bound, since we can't afford to lose compatibility with pre-0.18 versions of ppxlib at the moment.

In the past this was done using `cppo`, however since Dune does not support pulling package version this is harder but not impossible (I've done it for the switch of dose3 to dune using ocamlfind + cppo)

packages/irmin/irmin.2.3.0/opam
CraigFe: Done :+1:

packages/ppx_irmin/ppx_irmin.2.3.0/opam
CraigFe: I'm happy to make a change like that, but it would require a re-release of `repr` since `ppx_repr` has the same issue. Perhaps we can save that for `2.3.1`?

kit-ty-kate: Sure, sure! ^_^

CraigFe:
Updated to a new commit hash which should contain a passing test suite :crossed_fingers:

CraigFe:
There are still issues reported by the CI; currently working on fixing them.

samoht:
The errors are unrelated:

Screenshot 2021-01-12 at 22 00 33

So I'm merging this one.

PR #17963, opened 2021-01-13
Package lwt-canceler.0.3

raphael-proust:
### `lwt-canceler.0.3`
Cancellation synchronization object

---
* Homepage: https://gitlab.com/nomadic-labs/lwt-canceler
* Source repo: git+https://gitlab.com/nomadic-labs/lwt-canceler.git
* Bug tracker: https://gitlab.com/nomadic-labs/lwt-canceler/issues

---

### DIFF
--- a/packages/lwt-canceler/lwt-canceler.0.3/opam
+++ b/packages/lwt-canceler/lwt-canceler.0.3/opam
@@ -1,0 +1,25 @@
+opam-version: "2.0"
+maintainer: "contact@nomadic-labs.com"
+authors: [ "Nomadic Labs" ]
+homepage: "https://gitlab.com/nomadic-labs/lwt-canceler"
+bug-reports: "https://gitlab.com/nomadic-labs/lwt-canceler/issues"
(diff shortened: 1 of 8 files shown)


raphael-proust:
I just retagged and pushed to set a lower bound to the `ocaml` dependency following CI failure.

For the revdeps, this is an issue with `tezos` packages that should depend on exact version for packages such as lwt-canceler that have a strange versioning scheme.

raphael-proust:
Oh! It makes sense to add the constraints to tezos packages in this MR! Thanks! Next time, I'll do that.

kit-ty-kate:
> should depend on exact version for packages such as lwt-canceler that have a strange versioning scheme.

For the record I don't think this should be the case. Strict version dependencies are a nightmare for opam users and should be avoided when possible (especially `=` constraints as they do not allow revisions or unexpected fix micro versions to be used)

I've fixed another issue in your opam file. Could you return the fix upstream? Thanks a lot!

PR #17965, opened 2021-01-13
[new release] opam-monorepo (0.2.1)

NathanReb:

### Fixed

- Fix `--recurse-opam` option for the monorepo lock phase: correctly perform special directory
  filtering, add an error message when two versions of the same package opam file exist in the
  source tree, perform package name filtering before checking for uniqueness (ocamllabs/opam-monorepo#151, @TheLortex)

### DIFF
--- a/packages/opam-monorepo/opam-monorepo.0.2.1/opam
+++ b/packages/opam-monorepo/opam-monorepo.0.2.1/opam
@@ -1,0 +1,47 @@
+opam-version: "2.0"
+synopsis: "Assemble and manage fully vendored Dune repositories"
+description: """
+The opam monorepo plugin provides a convenient interface to bridge the
+opam package manager with having a local copy of all the source
+code required to build a project using the dune build tool."""
+maintainer: ["anil@recoil.org"]
+authors: [
+  "Anil Madhavapeddy" "Nathan Rebours" "Lucas Pluvinage" "Jules Aguillon"
+]
+license: "ISC"
+homepage: "https://github.com/ocamllabs/opam-monorepo"
+bug-reports: "https://github.com/ocamllabs/opam-monorepo/issues"
+depends: [
+  "dune" {>= "2.6"}
+  "ocaml" {>= "4.08.0" & < "4.12"}
+]
+dev-repo: "git+https://github.com/ocamllabs/opam-monorepo.git"
+build: [ "dune" "build" "-p" name "-j" jobs "@install" "@runtest" {with-test} ]
+flags: [ plugin ]
+depexts: [
+  ["devel/pkgconf"] {os = "openbsd"}
+  ["pkg-config"] {os-distribution = "cygwinports"}
+  ["pkg-config"] {os-family = "debian"}
+  ["pkg-config"] {os = "macos" & os-distribution = "homebrew"}
+  ["pkgconf"] {os = "freebsd"}
+  ["pkgconf"] {os-distribution = "alpine"}
+  ["pkgconf"] {os-distribution = "arch"}
+  ["pkgconf-pkg-config"] {os-distribution = "centos" & os-version >= "8"}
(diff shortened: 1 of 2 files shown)


NathanReb:
I also updated `opam-monorepo.0.2.0` for which I forgot to add the depexts!

packages/opam-monorepo/opam-monorepo.0.2.1/opam
```diff
@@ -0,0 +1,47 @@
...
+  ["pkgconf"] {os-distribution = "arch"}
+  ["pkgconf-pkg-config"] {os-distribution = "centos" & os-version >= "8"}
+  ["pkgconf-pkg-config"] {os-distribution = "rhel" & os-version >= "8"}
+  ["pkgconfig"] {os-distribution = "fedora"}
+  ["pkgconfig"] {os-distribution = "mageia"}
+  ["pkgconfig"] {os-distribution = "nixos"}
+  ["pkgconfig"] {os-distribution = "ol"}
+  ["pkgconfig"] {os = "macos" & os-distribution = "macports"}
+  ["pkgconfig"] {os-distribution = "centos" & os-version <= "7"}
+  ["pkgconfig"] {os-distribution = "rhel" & os-version <= "7"}
```
kit-ty-kate, line 37: I remember the meeting we had on friday but at the time I couldn't really grasp what this would mean for opam-repository (it's difficult for me to reason about something if I don't have something to look at).

Is there any chance opam-monorepo could detect the difference between `conf-*` packages, just depend on them and vendor the rest? Duplicating all this is annoying on a maintenance point-of-vue

NathanReb: For now I think we'll have to settle with this. Those are actual dependencies of `opam-monorepo` since it vendors its opam deps.

The future plan is for opam-monorepo to generate a separate virtual package with only the depext so that mean we would have `opam-monorepo -> conf-opam-monorepo` or whatever an approriate name would be and `conf-opam-monorepo` would just have the full set of depexts.

If that's not satisfying we could think about an improvement so that the conf package depends on conf packages itself but I'm not sure we can always assume such a package exist for every depext out there. Direct depexts are probably more reliable.

NathanReb: Out of curiosity and to help me bette fix this in the future, could you explain what exactly is the problem with the depexts expressed that way?

Also a quick note, to avoid having some fragile logic in opam-monorepo, it might be nice to have a flag to mark `conf-*` packages so they're easy to identify. We can work without it at first by looking for packages with no url field, no depends and just depexts but I'm guessing there might be some value into quickly be able to identify them even for other uses.

kit-ty-kate: > it might be nice to have a flag to mark conf-* packages so they're easy to identify

It's already the case. They all have `flags: conf`

> Out of curiosity and to help me bette fix this in the future, could you explain what exactly is the problem with the depexts expressed that way?

The issue is that with both the current method and the `conf-opam-monorepo` solution us opam-repo maintainers have to maintains a duplicated sets of depexts (x² times in the case of the current method, +1 time per package using opam-monorepo in the second case). In this case if we need to change something in the way pkg-config is fetched we have to do it x times and that's not sustainable for us.

I've fixed that in this PR for now. Is that alright for you?

NathanReb: Yeah that looks good!

NathanReb:
I'll file an issue on opam-monorepo so we can improve the situation in the future but to clarify, we should make it so it always depends on `conf-*` packages when possible, only falling back to depexts when there is no such package.

All these dependencies would be in a separate virtual opam package.

Does that sound right to you?

kit-ty-kate:
yes that sounds perfect! Thanks a lot!

kit-ty-kate:
Thanks a lot!

PR #17971, opened 2021-01-14
Patch dune 2.8.0: Fix missing output from dune --version

kit-ty-kate:


jeremiedimino (approved):
LGTM

avsm:
This needs to be in a 2.0.8-p1 patch release ideally in opam-repo now that the earlier one has been merged :-/  It's bad form to mutate an existing release if at all possible -- what are the implications of not having the patch if someone takes the earlier changeset of opam-repo?

Issue #17975, opened 2021-01-14
Irmin 2.3.0 Vs Tezos(-storage) 8.1: missing constraint


smondet:
It seems that both releases happened close enough that the CI didn't catch this(?)

```
#=== ERROR while compiling tezos-storage.8.1 ==================================#
# context     2.0.3 | linux/x86_64 | ocaml-base-compiler.4.09.1 | https://opam.ocaml.org#e0239c66
# path        ~/.opam/for_tezos/.opam-switch/build/tezos-storage.8.1
# command     ~/.opam/opam-init/hooks/sandbox.sh build dune build -p tezos-storage -j 1
# exit-code   1
# env-file    ~/.opam/log/tezos-storage-18811-25a199.env
# output-file ~/.opam/log/tezos-storage-18811-25a199.out
### output ###
# 109 |   let short_hash t = Irmin.Type.(short_hash string (H.to_raw_string t))
# [...]
# Error: This function has type
#          'a Irmin.Type.t -> (?seed:int -> 'a -> int) Irmin.Type.staged
#        It is applied to too many arguments; maybe you forgot a `;'.
#     ocamlopt src/lib_storage/.tezos_storage.objs/native/tezos_storage__Context.{cmx,o} (exit 2)
# (cd _build/default && /home/admin/.opam/for_tezos/bin/ocamlopt.opt -w -40 -open Tezos_shell_services -open Tezos_base__TzPervasives -open Tezos_stdlib_unix -open Tezos_stdlib -g -I src/lib_storage/.tezos_storage.objs/byte -I src/lib_storage/.tezos_storage.objs/native -I /home/admin/.opam/for_tezos/lib/angstrom -I /home/admin/.opam/for_tezos/lib/astring -I /home/admin/.opam/for_tezos/lib/base6[...]
# File "src/lib_storage/context.ml", line 109, characters 33-43:
# 109 |   let short_hash t = Irmin.Type.(short_hash string (H.to_raw_string t))
#                                        ^^^^^^^^^^
# Error: This function has type
#          'a Irmin.Type.t -> (?seed:int -> 'a -> int) Irmin.Type.staged
#        It is applied to too many arguments; maybe you forgot a `;'.
```

PR #17976, opened 2021-01-14
Fix version constraint: tezos-storage → irmin

smondet:
Attempt at fixing <https://github.com/ocaml/opam-repository/issues/17975>.


samoht:
woops thanks for the fix!

kit-ty-kate:
My bad. Thanks!

PR #17986, opened 2021-01-17
Release caqti 1.3.0.

paurkedal:
All sub-packages are updated except caqti-type-calendar.

From the release notes:

- Implement `affected_count` for sqlite backend (GPR#46 jakob).
- Add method `exec_with_affected_count` to `Caqti_connection_sig.S` (GPR#45
  jakob).
- Add `?max_idle_size` to pool creation functions.
- Dropped dependency on `ppx_deriving` due to issue with static compilation
  (GPR#50 Ulrik Strid).
- Pass through `$<var>$` in query strings and deprecate `$$`.
- Log statements to be executed at debug level.
- Add COPYING.OCAML and fix license expression in opam files.
- Misc improvements to tests and documentation (GPR#51 Philippe Wang, GPR#54
  Reynir Björnsson, etc.).

### DIFF
--- a/packages/caqti-async/caqti-async.1.3.0/opam
+++ b/packages/caqti-async/caqti-async.1.3.0/opam
@@ -1,0 +1,32 @@
+opam-version: "2.0"
+maintainer: "Petter A. Urkedal <paurkedal@gmail.com>"
+authors: "Petter A. Urkedal <paurkedal@gmail.com>"
+license: "LGPL-3.0-or-later WITH OCaml-LGPL-linking-exception"
+homepage: "https://github.com/paurkedal/ocaml-caqti/"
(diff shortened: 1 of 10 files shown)


paurkedal:
I think the errors from opam-ci are not be be solved here, as they are due to missing client libraries on some OSes and a 32 bit issue in the mariadb package.

kit-ty-kate:
I've fixed 32bits support in https://github.com/andrenth/ocaml-mariadb/pull/37 and I'll fix the `mariadb` package here in a future PR.

Otherwise everything is green now. Thanks a lot!

paurkedal:
Great work! I was just looking at that 32 bit issue.

PR #17987, opened 2021-01-17
release camlp5, (full) pa_ppx suite, compat ocaml 4.12.0

chetmurthy:
This releases

* camlp5 compat with ocaml 4.12.0, along with some bugfixes
* pa_ppx, same as above
* new PPX rewriters for:
  * migrating ASTs
  * hashconsing types automatically (and generating memoizing functions)
  * adding unique-ids to nodes in an AST
  * automatically generating Camlp5-style quotations for new AST types
  * the start of an attribute-grammar evaluator-generator: currently support topological and ordered evaluation

Lot of code here, and I broke it up b/c they really are independent, and some are big enough that someone might not want to install them all in order to just get one bit of function.

### DIFF
--- a/packages/camlp5/camlp5.8.00.01/opam
+++ b/packages/camlp5/camlp5.8.00.01/opam
@@ -1,0 +1,59 @@
+name: "camlp5"
+version: "8.00.01"
+opam-version: "2.0"
+synopsis: "Preprocessor-pretty-printer of OCaml"
+description: """
+Camlp5 is a preprocessor and pretty-printer for OCaml programs. It also provides parsing and printing tools.
+
+As a preprocessor, it allows to:
+
+extend the syntax of OCaml,
+redefine the whole syntax of the language.
+As a pretty printer, it allows to:
+
+display OCaml programs in an elegant way,
+convert from one syntax to another,
+check the results of syntax extensions.
+Camlp5 also provides some parsing and pretty printing tools:
+
+extensible grammars
+extensible printers
+stream parsers and lexers
+pretty print module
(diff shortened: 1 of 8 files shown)


chetmurthy:
OK, it looks like everything passed except for oraclelinux-7, which doesn't have some required Perl packages.

kit-ty-kate:
All green. Thanks!

packages/camlp5/camlp5.8.00.01/opam
```diff
@@ -0,0 +1,59 @@
...
+pretty print module
+It works as a shell command and can also be used in the OCaml toplevel."""
+maintainer: "https://github.com/ocaml/opam-repository/issues"
+authors: ["Daniel de Rauglaudre" "Chet Murthy"]
+license: "BSD-3-Clause"
+homepage: "https://camlp5.github.io"
+doc: "https://camlp5.github.io/doc/html"
+bug-reports: "https://github.com/camlp5/camlp5/issues"
+depends: [
+  "ocaml" {>= "4.02" & < "4.13.0"}
```
Kakadu, line 31: I think that it should me `>= "4.10"`
@chetmurthy

chetmurthy: um, camlp5 8.00 should work fine with OCaml 4.02 -- it just won't have any of the new syntax available.

chetmurthy: I'll go further: it should be a -bug- if Camlp5 8.xx does not work with Ocaml >= 4.02 .

PR #17989, opened 2021-01-18
Add xquartz depext to conf-libX11 for macOS

smorimoto:
See https://github.com/ocaml/graphics/pull/31 for details.


kit-ty-kate:
I've tested it locally and this seems to work just fine. Thanks a lot!

smorimoto:
No! Thanks for your hard work on this very massive repository and project!

PR #17990, opened 2021-01-18
camlbz2: upgrade to 0.7.0

josch:


kit-ty-kate:
Mh, I'm not sure to understand. Why is it not possible to simply add `camlbz2.0.6.1` instead of modifying 0.6.0?

Also just as a personal note (nothing blocking) this should be released as 0.7.0 given the interface changes quite significantly.
```
diff -ru camlbz2-0.6.0/bz2.mli camlbz2-0.6.1/bz2.mli
--- camlbz2-0.6.0/bz2.mli	2008-11-08 11:05:52.000000000 +0000
+++ camlbz2-0.6.1/bz2.mli	2021-01-11 11:11:49.000000000 +0000
@@ -51,9 +51,8 @@
     decompressing that is slower but uses less memory. Defaults:
     [false]
 *)
-external open_in : ?small:bool -> ?unused:string -> Pervasives.in_channel ->
-  in_channel 
-  = "mlbz_readopen"
+val open_in : ?small:bool -> ?unused:bytes -> Pervasives.in_channel ->
+  in_channel
 
 (** [read buf pos len] reads up to [len] characters and store them in
     the string buffer [buf], starting at position [pos].
@@ -62,17 +61,14 @@
     [len] means end of stream).
 
     @raise End_of_file if end of stream was already reached. *)
-external read : in_channel -> string -> int -> int -> int
-  = "mlbz_read"
+val read : in_channel -> bytes -> int -> int -> int
 
 (** If there's some data after the compressed stream that you want to
     read from the same [Pervasives] [in_channel], use
     [read_get_unused]. *)
-external read_get_unused : in_channel -> string
-  = "mlbz_readgetunused"
+val read_get_unused : in_channel -> bytes
 
-external close_in : in_channel -> unit
-  = "mlbz_readclose"
+val close_in : in_channel -> unit
 
 (** {3 File output} *)
 
@@ -85,16 +81,13 @@
     @param block block size to use for compresion. It is a value
     between 1 and 9 inclusive. 9 is the default and provides best
     compression but takes most memory. *)
-external open_out : ?block:int -> Pervasives.out_channel -> out_channel
-  = "mlbz_writeopen"
+val open_out : ?block:int -> Pervasives.out_channel -> out_channel
 
 (** [write oc buf pos len] writes [len] characters, coming from [buf]
     and starting at position [pos], to [oc] *)
-external write : out_channel -> string -> int -> int -> unit
-  = "mlbz_write"
+val write : out_channel -> bytes -> int -> int -> unit
 
-external close_out : out_channel -> unit
-  = "mlbz_writeclose"
+val close_out : out_channel -> unit
 
 
 (** {2 In-memory compression} *)
@@ -105,13 +98,11 @@
     [len] character long, and starting at [pos].
 
     @return compressed data chunk as string *)
-external compress : ?block:int -> string -> int -> int -> string
-  = "mlbz_compress"
+val compress : ?block:int -> bytes -> int -> int -> bytes
 
 (** [uncompress buf pos len] uncompress a data chunk comfing from
     [buf], [len] character long, and starting at [pos].
 
     @param small see [Bz2.open_in] above
     @return uncompressed data chunk as string *)
-external uncompress : ?small:bool -> string -> int -> int -> string
-  = "mlbz_uncompress"
+val uncompress : ?small:bool -> bytes -> int -> int -> bytes
```

josch:
> Mh, I'm not sure to understand. Why is it not possible to simply add camlbz2.0.6.1 instead of modifying 0.6.0?

Whoops indeed. Thanks! I have no clue about opam, so I didn't even think about the option of having multiple versions at once.

> Also just as a personal note (nothing blocking) this should be released as 0.7.0 given the interface changes quite significantly.

You are absolutely right. I botched that release. I fixed it now by indeed making a 0.7.0 release with no changes other than in the changelog, explaining the situation. Thanks!

kit-ty-kate:
Thanks a lot!

PR #17997, opened 2021-01-19
Unlock the whole bap stack for OCaml 4.12

kit-ty-kate:
cc @ivg is that ok for you? I've compiled the whole stack (or at least the path used by `bap-pheonix.2.2.0`) locally on OCaml 4.12 and it works perfectly

### DIFF
--- a/packages/bap-relation/bap-relation.2.2.0/opam
+++ b/packages/bap-relation/bap-relation.2.2.0/opam
@@ -14,7 +14,7 @@ build: [
 install: [[make "install"]]
 remove: [["ocamlfind" "remove" "bap-relation"]]
 depends: [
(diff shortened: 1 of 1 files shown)


ivg:
wow, that's awesome, thanks! And yes, it is totally OK for me)

ivg:
the only caveat is that this patch will only unlock one library not the whole bap stack :)

kit-ty-kate:
> the only caveat is that this patch will only unlock one library not the whole bap stack :)

As far at I tested this was the only package that would prevent the bap packages from being installed on 4.12.
I'm running a test on the whole opam-repository to confirm this and will merge this right after.

kit-ty-kate:
Everything seems good! Thanks!

PR #17998, opened 2021-01-20
+ extlib 1.7.8

ygrek:
1.7.8 (2021-01-19)
* sync with OCaml 4.12
* breaking change: ExtList.find_map type updated to match stdlib (following deprecation in previous release)
* breaking change: minimal (recommended) build of extlib now excludes Base64 module

### DIFF
--- a/packages/dose3/dose3.4.3/opam
+++ b/packages/dose3/dose3.4.3/opam
@@ -27,7 +27,7 @@ depends: [
   "ocaml"
   "ocamlgraph" {= "1.8.6"}
   "cudf" {>= "0.7"}
-  ("extlib" {>= "1.7.0"} | "extlib-compat" {>= "1.7.0"})
+  ("extlib" {>= "1.7.0" & < "1.7.8"} | "extlib-compat" {>= "1.7.0"})
   "re" {>= "1.2.2"}
   "ocamlbuild" {build}
   "cppo" {build & >= "1.1.2"}
--- a/packages/dose3/dose3.5.0.1-1/opam
+++ b/packages/dose3/dose3.5.0.1-1/opam
@@ -24,7 +24,7 @@ depends: [
   "ocaml"
   "ocamlgraph" {>= "1.8.6"}
   "cudf" {>= "0.7"}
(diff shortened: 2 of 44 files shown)


kit-ty-kate:
Some packages were broken by the breaking changes.

cc @abate for `dose3` (I wrote a comment already in !2)
cc @astrada for `google-drive-ocamlfuse`
cc @andyli for `haxe`
cc @mfp for `extprot`

Fixing these issues is required for get the compatibility with OCaml 4.12.

Thanks!

ygrek:
> implicitly uses Base64, previously pulled by extlib

to clarify: extlib has implementation of Base64 module. There is also a separate opam package base64 (unrelated to extlib) which has much more elaborate implementation of base64. Due to all extlib modules being at top-level of the namespace - there was a name conflict which prevented using extlib and base64 opam packages in the same project. In this release extlib drops Base64 from the minimal build (but it is still available in the full build, opam package extlib-compat) to avoid the above problem, recommended solution is to switch to standalone base64 implementation. Short-term solution may be to use extlib-compat (but I stopped uploading it to opam for the lack of interest, let me know if it is needed).

kit-ty-kate:
ah.. Looks like I made a bad judgement then.

This is a rather large breakage for a point release, I'm rather surprised here.

kit-ty-kate:
Mh, thinking about it a bit more, it makes sense if taken from the point of vue of extlib + extlib-compat I guess. I hadn't realized this was the goal of extlib-compat.

Could you also send a PR for extlib-compat.1.7.8? that'd be wonderful. I'll handle the addition of `|  extlib-compat` everywhere if you don't have the time to do that

ygrek:
extlib has inherent namespacing problem, and reducing number of modules (ideally leaving only Ext* ones) is one way to solve it eventually. Hence, I think it is best for ecosystem if dependent projects eventually migrated off these modules. I will upload extlib-compat.

ygrek:
actually I just realized that probably easiest way to solve packaging/conflicts issues may be to provide one-module extlib-base64 extlib-utf8 etc libraries (best in the sense that namespacing problem remains, but extlib choices will not create new problems for existing code)

PR #17999, opened 2021-01-20
+ extlib-compat 1.7.8

ygrek:
followup to https://github.com/ocaml/opam-repository/pull/17998


kit-ty-kate:
I like the `extlib-base64` proposal from https://github.com/ocaml/opam-repository/pull/17998#issuecomment-763325276 better. `extlib-compat` add more complexities than it solves compared to it.

Having `extlib-base64` would simply allow every packages that I constrained in #17998 to be compatible without having to install an incompatible package with `extlib` (aka `extlib-compat`).

Would it be ok to just add this new package instead? (and just make it clear for the users that the library is deprecated)

kit-ty-kate:
Superseded by https://github.com/ocaml/opam-repository/pull/18041

PR #18000, opened 2021-01-20
+ devkit 1.20210120

ygrek:


kit-ty-kate:
This package seems to fail on 32bit architectures. Is that expected?
```
#=== ERROR while compiling devkit.1.20210120 ==================================#
# context              2.1.0~beta4 | linux/x86_32 | ocaml-base-compiler.4.11.1 | pinned(https://github.com/ahrefs/devkit/releases/download/1.20210120/devkit-1.20210120.tbz)
# path                 ~/.opam/4.11/.opam-switch/build/devkit.1.20210120
# command              ~/.opam/4.11/bin/dune build -p devkit -j 31 @install @runtest
# exit-code            1
# env-file             ~/.opam/log/devkit-23-b0e90f.env
# output-file          ~/.opam/log/devkit-23-b0e90f.out
### output ###
#         test alias runtest (exit 2)
# (cd _build/default && ./test.exe)
# Fatal error: exception File "network.ml", line 3, characters 9-15: Assertion failed
```

ygrek:
Will see if we can lift it for the next release
For now will just add a constraint
Sorry for the fixes requires in opam file

ygrek:
actually it was never buildable on 32-bit, nobody complained except CI, so not going to look to fix :)

packages/devkit/devkit.1.20210120/opam
```diff
@@ -7,12 +7,12 @@ dev-repo: "git+https://github.com/ahrefs/devkit.git"
 bug-reports: "https://github.com/ahrefs/devkit/issues"
 synopsis: "Development kit - general purpose library"
 build: [
-  ["dune" "subst"] {pinned}
+  ["dune" "subst"] {dev}
```
ygrek, line 10: I understand `dev` subsumes `pinned` so I should replace in opam file in devkit git repo, right?

kit-ty-kate:
The `ocaml-option` package is not the way to do this. This is only a package to enable 32bit-only support in the ocaml compiler, not to signify a 32bit architecture. I've removed the commit and fixed it

packages/devkit/devkit.1.20210120/opam
kit-ty-kate: yes please!

kit-ty-kate:
Looks good now. Thanks!

PR #18003, opened 2021-01-20
Package boltzgen.0.9

barbot:
### `boltzgen.0.9`
Generate test using boltzman sampling
BoltzGen is a tool to generate tests. Given a function signature it generates a random set of call of this function on generated random input. Inputs are generated using a Boltzman sampler for unlabeled structure.

---
* Homepage: https://git.lacl.fr/barbot/boltzgen
* Bug tracker: https://git.lacl.fr/barbot/boltzgen/-/issues

---

### DIFF
--- a/packages/boltzgen/boltzgen.0.9/opam
+++ b/packages/boltzgen/boltzgen.0.9/opam
@@ -1,0 +1,37 @@
+opam-version: "2.0"
+synopsis: "Generate test using boltzman sampling"
+description:
+  "BoltzGen is a tool to generate tests. Given a function signature it generates a random set of call of this function on generated random input. Inputs are generated using a Boltzman sampler for unlabeled structure."
+maintainer: ["barbot@lacl.fr"]
+authors: ["Benoît Barbot"]
+license: "GPLv3"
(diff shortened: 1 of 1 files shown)


packages/boltzgen/boltzgen.0.9/opam
```diff
@@ -0,0 +1,37 @@
...
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
+url {
+  src:
+    "https://git.lacl.fr/barbot/boltzgen/-/archive/master/boltzgen-master.tar.gz"
```
kit-ty-kate, line 32: I don't think this is the archive you want to release. Unless I'm mistaken, this archive will change overtime and the installation of this package will fail because the checksum is not going to be correct.

Could you use the archive of the tag instead?

barbot: Of course you are right, 
It is the first time I publish a package, should'I rerun the opam publish tool or should'I modify this submission ?

kit-ty-kate:
Thanks!

PR #18021, opened 2021-01-22
Add ocaml-option-nnpchecker

dra27:
This PR adds `ocaml-option-nnpchecker` to trigger the `--enable-naked-pointers-checker` option in OCaml 4.12+

I haven't updated ocaml-config to emulate a switch name for this - I'm not sure we particularly want one, but it can be added if required.

I have intentionally edited the existing options as they only apply to a beta release - in future we'd expect to bump the versions.

cc @Octachron

### DIFF
--- a/packages/ocaml-option-nnpchecker/ocaml-option-nnpchecker.1/opam
+++ b/packages/ocaml-option-nnpchecker/ocaml-option-nnpchecker.1/opam
@@ -1,0 +1,9 @@
+opam-version: "2.0"
+synopsis: "Set OCaml to be compiled with --enable-naked-pointers-checker"
+depends: [
+  "ocaml-variants" {post & >= "4.12.0~"}
+]
+available: arch = "x86_64" & (os = "linux" | os = "macos" | os = "openbsd" | os = "freebsd" | os = "sunos")
+conflicts: [ "ocaml-option-nnp" "ocaml-option-32bit" ]
+maintainer: "platform@lists.ocaml.org"
+flags: compiler
--- a/packages/ocaml-options-only-afl/ocaml-options-only-afl.1/opam
+++ b/packages/ocaml-options-only-afl/ocaml-options-only-afl.1/opam
@@ -12,6 +12,7 @@ conflicts: [
   "ocaml-option-spacetime"
   "ocaml-option-static"
   "ocaml-option-nnp"
+  "ocaml-option-nnpchecker"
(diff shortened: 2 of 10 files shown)


packages/ocaml-option-nnpchecker/ocaml-option-nnpchecker.1/opam
```diff
@@ -0,0 +1,9 @@
+opam-version: "2.0"
+synopsis: "Set OCaml to be compiled with --enable-naked-pointers-checker"
+depends: [
+  "ocaml-variants" {post & >= "4.12.0~"}
+]
+available: arch = "x86_64" & (os = "linux" | os = "macos" | os = "openbsd" | os = "freebsd")
```
kit-ty-kate, line 6: is the option really specific to those platforms and architecture?

dra27: It is indeed, see https://github.com/ocaml/ocaml/blob/trunk/configure.ac#L1727-L1735. Technically msvc64 and solaris are also included, but I know we don't have an encoding for msvc64 and I haven't got access to Solaris to see what `uname -s` does (I think it's `SunOS`, but I have no way to check!)

dra27: Perhaps @mtelvers still has a Solaris box gathering dust somewhere? 😉

mtelvers: @dra27 I have recovered my SparcStation 20 from deep storage and even found a Sun 13W3-M to VGA adapter but sadly none of my monitors accept the resolution/timing.

avsm: @dra27 It's SunOS -- the Wikipedia page has a convenient list: https://en.wikipedia.org/wiki/Uname

dra27: Ta, @avsm! A handy list (if less fun than persuading Mark to try building OCaml on a 1994 Sun Workstation 😉). I've pushed an amended version with `os = "sunos"` too (opam normalises with `lowercase_ascii`)

kit-ty-kate:
Thanks!

PR #18022, opened 2021-01-22
conf-linux-libc-dev: Remove absolute path to header

Julow:
Absolute paths can't work on some systems (for example, NixOS, opam2nix, termux).
This PR looks up `linux/stddef.h` using the C compiler.

If the header still can't be found, the error will look like this:

```
# + echo '#include <doesnt_exist.h>'
# + cc -E -
# <stdin>:1:10: fatal error: doesnt_exist.h: No such file or directory
# compilation terminated.
```


kit-ty-kate:
Thanks a lot!

PR #18027, opened 2021-01-23
Relax version constraint for caqti-async.

paurkedal:
I have overlooked that the async constraint had been relaxed since the
previous release.  Thanks to Aaron Zeng (@bcc32) for reporting the
issue.


kit-ty-kate:
Looks good but just a slight comment: you might want to remove the upper-bound entirely in your repository. We check and constrain the packages that are not compatible ourselves at each releases of the janestreet packages. This might relive you of some maintenance strain.

Of course the choice is yours but that's just my opinion.

paurkedal:
Yes, that is much better from my point of view. I though it was mandatory. Will update the PR in a moment.

kit-ty-kate:
Thanks!

PR #18028, opened 2021-01-24
Package timere.0.1.1

darrenldl:
OCaml time handling and reasoning library

Homepage: https://github.com/daypack-dev/timere
Repo: git+https://gitlab.com/daypack-dev/timere.git
Issue tracker: https://github.com/daypack-dev/timere/issues


darrenldl:
An improvement is in the works that would superseded the fix applied here. Will submit that version instead.

PR #18032, opened 2021-01-25
ppx_repr: add ppx_deriving dependency (for non-dune users)

hannesm:
/cc @CraigFe @dinosaure 

see https://github.com/roburio/dns-primary-git/pull/2 https://github.com/mirage/repr/pull/36


hannesm:
sorry, I'm not sure about the revdeps failure and won't investigate.

dinosaure:
> sorry, I'm not sure about the revdeps failure and won't investigate.

I think it's a missing constraint between `index` and `index-bench` (which should require the `{= version}` constraint). Any idea @CraigFe?

CraigFe:
@dinosaure: agreed. PR here: https://github.com/ocaml/opam-repository/pull/18036

kit-ty-kate:
Thanks!

PR #18033, opened 2021-01-25
Fix incompatibility constraint with dune 2.8

clarus:
This incompatibility is due to the fact that dune >2.8 does not generate a `.merlin` file anymore. We will do a new release of coq-of-ocaml to work with the new dune version.


kit-ty-kate:
thanks!

PR #18036, opened 2021-01-25
index-bench: add missing version constraint on index

CraigFe:
c.f. https://github.com/ocaml/opam-repository/pull/18032#issuecomment-766868849.


kit-ty-kate:
Thanks!

PR #18041, opened 2021-01-26
extlib.1.7.7-1: Add OCaml 4.12 compatibility to extlib < 1.7.8

kit-ty-kate:
Patches taken from upstream.

Should solve issues from https://github.com/ocaml/opam-repository/pull/17998 / https://github.com/ocaml/opam-repository/pull/17999 without having to rely on extlib-compat.

cc @ygrek is that fine for you? I'll close your PR for `extlib-compat.1.7.8` if yes.


kit-ty-kate:
Unless you want to release `extlib.1.7.7.1`?

ygrek:
Sure, thanks.
 I understand this is a quickfix until I add extlib-base64.

kit-ty-kate:
> I understand this is a quickfix until I add extlib-base64.

I've tried doing the extlib-base64 package myself but it seems not possible to do as I previously thought, as the module was previously in the `extlib` library itself. Adding an extlib-base64 library would not be very useful as it would require packages to be patched and use it explicitly which would defeat the purpose of adding it in the first place.

Sorry for failing to understand how extlib worked in the first place. Don't worry about making extlib-base64 (unless users struggle to switch to the `base64` package)

PR #18049, opened 2021-01-27
upgrade various Jane Street libraries

cwong-ocaml:


kit-ty-kate:
hardcaml does not compile on 32bit platforms. I'm going to assume this is expected and will make the package unavailable on those platforms when CI will be done to avoid overloading it.

kit-ty-kate:
Thanks a lot! Could you return the fix upstream?

PR #18057, opened 2021-01-28
[new release] earlybird-411 (1.0.0-beta.0)

hackwaly:
Debug adapter for OCaml 4.11

- Project page: <a href="https://github.com/hackwaly/ocamlearlybird">https://github.com/hackwaly/ocamlearlybird</a>

##### CHANGES:

Initial release.

### DIFF
--- a/packages/earlybird-411/earlybird-411.1.0.0-beta.0/opam
+++ b/packages/earlybird-411/earlybird-411.1.0.0-beta.0/opam
@@ -1,0 +1,49 @@
+opam-version: "2.0"
+synopsis: "Debug adapter for OCaml 4.11"
+description: "Debug adapter for OCaml 4.11."
+maintainer: ["hackwaly@qq.com"]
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
`~` should always be used as a separator instead of `-` for beta versions. See: https://opam.ocaml.org/doc/Manual.html#version-ordering.
Otherwise `1.0.0-beta.0` will be prioritied before `1.0.0`.

Also the package seems to have changed its name. Adding a new package for each version of OCaml seems rather inconvenient for everyone. I think it would be better to stick with the `earlybird` package name and let the opam constraints do their job. What do you think?

hackwaly:
> I think it would be better to stick with the earlybird package name and let the opam constraints do their job. What do you think?

How do I do? Different implementation for each minor OCaml versions is required.

hackwaly:
It seems dune-release do not support `~` symbols in version. I consider to release 1.0.0 directly

PR #18059, opened 2021-01-28
Create symlink for conf-bison and conf-flex in Mac homebrew

MSoegtropIMC:
Mac Homebrew installs bison and flex to folders which are not in the usual PATH, because they would hide the macOS supplied (old) versions of bison and flex. Since packages depending on these conf packages can be assumed to require the provided tools and not the system supplied tools, the conf packages are changed to include an install step which creates a symlink in the switch bin folder.

I don't know how agreeable this method is - technically the package is no longer a conf package for homebrew, but I think this is the most reasonable solution. I also thought about using `setenv`, but then there is no way to call brew to get the proper path with `setenv`. Also cluttering PATH is not that nice. So in the end I found the symlink solution most reasonable.

I increased the package version because for macOS with homebrew it does make a difference - the package did install before - without the symlinks.

Other OSes (notably macOS MacPorts) don't have a similar issue - this is homebrew specific.

This is required to fix (https://gitlab.inria.fr/gappa/gappa/-/issues/7).

### DIFF
--- a/packages/conf-bison/conf-bison.2/opam
+++ b/packages/conf-bison/conf-bison.2/opam
@@ -1,0 +1,48 @@
+opam-version: "2.0"
+maintainer: "7895506+MSoegtropIMC@users.noreply.github.com"
+homepage: "https://www.gnu.org/software/bison/"
+bug-reports: "https://lists.gnu.org/mailman/listinfo/bug-bison"
+license: "GPL-3.0-or-later"
+authors: [
+  "Robert Corbett" 
+  "Richard Stallman"
+  "Wilfred Hansen"
+  "Akim Demaille"
+  "Paul Hilfinger" 
+  "Joel E. Denny"
+  "Paolo Bonzini"
+  "Alex Rozenman"
+  "Paul Eggert"
+]
+build: [ "bison" "--help" ]
+install: [
+  # Note: Mac homebrew installs bison to a path not in PATH, so symlink it in the opam switch bin folder.
+  # This affects only the current opam switch and is thus least invasive.
+  ["sh" "-c" "ln -s \"$(brew --prefix bison)/bin/bison\" \"%{bin}%/bison\""] {os-family = "homebrew"}
+  ["sh" "-c" "ln -s \"$(brew --prefix bison)/bin/yacc\" \"%{bin}%/yacc\""] {os-family = "homebrew"}
+]
+depexts: [
+  # The list of supported os-family names can be derived from the opam sources (thanks Enrico)
+  #   https://github.com/ocaml/opam/blob/6aefe95e60084e63d01b1c7c028b6b77de1f839f/src/state/opamSysInteract.ml#L84-L123
+  # The package name for many Linux/BSD/macOS/cygwin/msys platforms can be searched here (thanks Théo)
+  #   https://repology.org specifically https://repology.org/project/bison/versions
+  #   Not listed above but supported by opam are suse, oraclelinix, ol, rhel
+  # On macOS bison seems to be installed by default, but the homebrew and macports package is listed anyway
+  [ "bison" ] {os-family = "alpine"}
+  [ "bison" ] {os-family = "amzn"}
+  [ "bison" ] {os-family = "arch"}
+  [ "bison" ] {os-family = "archlinux"}
+  [ "bison" ] {os-family = "centos"}
+  [ "bison" ] {os-family = "debian"}
+  [ "bison" ] {os-family = "fedora"}
+  [ "sys-devel/bison" ] {os-family = "gentoo"}
+  [ "bison" ] {os-family = "homebrew"}
+  [ "bison" ] {os-family = "macports"}
+  [ "bison" ] {os-family = "mageia"}
+  [ "bison" ] {os-family = "opensuse"}
+  [ "devel/bison" ] {os-family = "bsd"}
+]
+synopsis: "Virtual package relying on GNU bison"
+description: "This package can only install if GNU bison is installed on the system."
+# Not technically a conf package for homebrew (has an install setp)
+# flags: conf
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
oh, interesting. I like the symlink idea. I wonder if the same thing can be done to improve the `conf-libssl` package on homebrew/macports/nixos.

packages/conf-bison/conf-bison.2/opam
```diff
@@ -0,0 +1,48 @@
...
+  "Richard Stallman"
+  "Wilfred Hansen"
+  "Akim Demaille"
+  "Paul Hilfinger" 
+  "Joel E. Denny"
+  "Paolo Bonzini"
+  "Alex Rozenman"
+  "Paul Eggert"
+]
+build: [ "bison" "--help" ]
```
kit-ty-kate, line 17: This should be more something like:
```
build: [
  ["bison" "--help"] {os-family != "homebrew"}
  ["sh" "-c" "\"$(brew --prefix bison)/bin/bison\""] {os-family = "homebrew"}
]
```
same for flex

MSoegtropIMC: Indeed, yes. Well spotted!

MSoegtropIMC:
> oh, interesting. I like the symlink idea. I wonder if the same thing can be done to improve the conf-libssl package on homebrew/macports/nixos.

I guess so, but one would need more symlinks and a few more tricks. With homebrew I get:
```
brew list openssl
/usr/local/Cellar/openssl@1.1/1.1.1i/.bottle/etc/ (7 files)
/usr/local/Cellar/openssl@1.1/1.1.1i/bin/c_rehash
/usr/local/Cellar/openssl@1.1/1.1.1i/bin/openssl
/usr/local/Cellar/openssl@1.1/1.1.1i/include/openssl/ (104 files)
/usr/local/Cellar/openssl@1.1/1.1.1i/lib/libcrypto.1.1.dylib
/usr/local/Cellar/openssl@1.1/1.1.1i/lib/libssl.1.1.dylib
/usr/local/Cellar/openssl@1.1/1.1.1i/lib/engines-1.1/ (2 files)
/usr/local/Cellar/openssl@1.1/1.1.1i/lib/pkgconfig/ (3 files)
/usr/local/Cellar/openssl@1.1/1.1.1i/lib/ (4 other files)
/usr/local/Cellar/openssl@1.1/1.1.1i/share/doc/ (3968 files)
/usr/local/Cellar/openssl@1.1/1.1.1i/share/man/ (3968 files)
```
Most of these can be symlinked as individual files / folders into the opam switch. The pkfconfig files might be an issue, though. I don't know if opam maintains its own pkg-config database - probably not. But one could change this in an opam package, say create a pkgconfig folder under lib in the switch and set the environment variable so that pkgconfig will look there - this can be done with `setenv` since the destination path is an opam path and accessible with e.g. `%{lib}%`. I would probably add a script in `files` to do the linking for openssl.

Btw.: I wouldn't expect any issues with MacPorts. If one uses MacPorts one almost always has something in the shell profile to give MacPorts stuff precedence over system stuff. MacPorts doesn't take care not to hide system provided tools as Homebrew does. I know nothing about Nix.

So if you have issues there I think it would be worth a try. I am available for discussions.

MSoegtropIMC:
P.S.: I had a look at conf-pkg-config. It doesn't seem to create an opam local database. So I would add an additional opam package which creates a switch local pkg-config database and sets the environment variable to point there. Maybe I would call it local-pkg-config. It is not problem to have several pkg-config databases.

kit-ty-kate:
yeah I was thinking of modifying `conf-pkg-config` by adding the local database with `setenv` and simply symlink the .pc file for openssl. Though I'm not sure if anyone uses libssl without pkg-config, I'd have to check first i guess

MSoegtropIMC:
I would have two different packages, the conf-pkg-config as is and another package which refers to it and in addition creates a local pkg.config data base. As far as I understand the logic of setenv, then only packages which depend on the local database package will actually get it. But I don't have a very strong opinion about this.

Btw.: such packages will break if the installed version changes. This is not that bad since if one uninstalls a system package the conf package also breaks (it remains installed). I guess one should have a `check` command which replaces the `build` in conf packages and is run every time a package is built which depends on it and one also might want to have an `opam check` command which checks if the dependencies of all conf packages are still there. Or does this exist already?

MSoegtropIMC:
@kit-ty-kate : I did the changes you suggested, so this should be good to go now.

mseri:
What happens if you reinstall the packages, say after an update from homebrew. Will it fail the relinking or silently do nothing?

MSoegtropIMC:
@mseri: I tested that uninstalling the conf package does remove the symlink. Since opam always removes packages before installing them new, I expect this to work. It is a bit hard to test this - I would have to create a Homebrew fake package and unfortunately I am more familiar with MacPorts.

I could do a `ln -sf` to be sure, but as I said, by design it should work without it, and I prefer no to give the `-f` option unless there is a good reason.

mseri:
Thanks for checking it out, it’s ok as is then. I think it is a useful improvement

MSoegtropIMC:
Should I create a feature request to opam for the check mechanism I outline above - that is replace the build stanza common in conf packages with a check stanza which is always run when a package is build requiring the conf package to check if the system didn't change since the conf package has been installed. It could also provide some easy to use mechanism for parsing the result of `xyz --version`.

kit-ty-kate:
I just wanted to make sure the distributions were correct. The only thing I could find is that the `os-family = "bsd"` shouldn't be there and should almost always be separated between freebsd, openbsd, netbsd, dragonfly but since the previous version does that too, it's fine.

Don't worry about it, I'll fix it someday.

Thanks a lot!

PR #18064, opened 2021-01-30
Improve pkg-config and improve the conf-libssl experience with Homebrew

kit-ty-kate:
This PR tries to improve the opam experience for people using macOS and NixOS and well as the general `pkg-config` experience.
The `conf-libssl` package would previously ask the user to change their `PKG_CONFIG_PATH` by hand to be able to locate openssl.pc.
The change to `conf-pkg-config` makes it so that `%{lib}%/pkgconfig` is always created and prepended to the global `PKG_CONFIG_PATH`.

This is a draft for the moment as it is just a PoC for the moment and a few questions remain unanswered, such as:
* Is the implementation for NixOS correct? Does it not have the path in the environment by default? (cc @timbertson)
* Is it correct for Macports? (cc @mseri)
* Should we symlink all the `.pc` files from those directories? With homebrew the directory contains:
  - libcrypto.pc
  - libssl.pc
  - openssl.pc

*Big thanks to @MSoegtropIMC for the original idea for bison & flex in https://github.com/ocaml/opam-repository/pull/18059*

### DIFF
--- a/packages/conf-libssl/conf-libssl.3/files/homebrew.sh
+++ b/packages/conf-libssl/conf-libssl.3/files/homebrew.sh
@@ -16,7 +16,7 @@ install)
     exit 1
   fi
   for file in $(ls "$brew_pkg_config"/*.pc); do
-    ln -s "$brew_pkg_config/$file" "$2/pkgconfig/$file"
+    ln -s "$file" "$2/pkgconfig/$file"
   done;;
 *)
   echo "Usage: $0 <check|install>"
--- a/packages/conf-libssl/conf-libssl.3/files/homebrew.sh
+++ b/packages/conf-libssl/conf-libssl.3/files/homebrew.sh
@@ -1,0 +1,24 @@
+#!/bin/sh -ex
+
+brew_pkg_config=$(brew --prefix openssl)/lib/pkgconfig
+
+case "$1" in
+check)
+  if test "$#" != 1; then
+    echo "Usage: $0 check"
+    exit 1
+  fi
+  export PKG_CONFIG_PATH=$brew_pkg_config:$PKG_CONFIG_PATH
+  pkg-config --print-errors --exists openssl;;
+install)
+  if test "$#" != 2; then
+    echo "Usage: $0 install <libdir>"
+    exit 1
+  fi
+  for file in $(ls "$brew_pkg_config"/*.pc); do
+    ln -s "$brew_pkg_config/$file" "$2/pkgconfig/$file"
+  done;;
+*)
+  echo "Usage: $0 <check|install>"
+  exit 1;;
+esac
--- a/packages/conf-libssl/conf-libssl.3/files/homebrew.sh
+++ b/packages/conf-libssl/conf-libssl.3/files/homebrew.sh
@@ -15,8 +15,11 @@ install)
     echo "Usage: $0 install <libdir>"
     exit 1
   fi
-  for file in $(ls "$brew_pkg_config"/*.pc); do
-    ln -s "$file" "$2/pkgconfig/$file"
+  cd "$brew_pkg_config"
+  for file in $(ls *.pc); do
+    if test -f "$file"; then
+      ln -s "$brew_pkg_config/$file" "$2/pkgconfig/$file"
+    fi
   done;;
 *)
   echo "Usage: $0 <check|install>"
--- a/packages/conf-libssl/conf-libssl.3/files/homebrew.sh
+++ b/packages/conf-libssl/conf-libssl.3/files/homebrew.sh
@@ -1,5 +1,8 @@
 #!/bin/sh -ex
 
+# check the openssl installation
+# and symlink the homebrew pkg-config files for openssl to the opam local pkgconfig directory
+
 brew_pkg_config=$(brew --prefix openssl)/lib/pkgconfig
 
 case "$1" in
--- a/packages/conf-libssl/conf-libssl.3/opam
+++ b/packages/conf-libssl/conf-libssl.3/opam
@@ -4,6 +4,7 @@ authors: ["The OpenSSL Project"]
 bug-reports: "https://github.com/ocaml/opam-repository/issues"
 homepage: "https://www.openssl.org/"
 license: "Apache-1.0"
+setenv: [PKG_CONFIG_PATH = "%{lib}%/pkgconfig"]
 build: [
   ["pkg-config" "--print-errors" "--exists" "openssl"]
     {os != "freebsd" & os != "openbsd" & os != "netbsd" # libssl is provided by base system on BSDs
--- a/packages/conf-libssl/conf-libssl.3/opam
+++ b/packages/conf-libssl/conf-libssl.3/opam
@@ -1,0 +1,40 @@
+opam-version: "2.0"
+maintainer: "David Sheets <sheets@alum.mit.edu>"
+authors: ["The OpenSSL Project"]
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+homepage: "https://www.openssl.org/"
+license: "Apache-1.0"
+build: [
+  ["pkg-config" "--print-errors" "--exists" "openssl"]
+    {os != "freebsd" & os != "openbsd" & os != "netbsd" # libssl is provided by base system on BSDs
+     os-distribution != "homebrew" & os-distribution != "macports" & os-distribution != "nixos"}
+  ["sh" "-c" "env \"PKG_CONFIG_PATH=$(brew --prefix openssl)/lib/pkgconfig:$PKG_CONFIG_PATH" pkg-config --print-errors --exists openssl"] {os-distribution = "homebrew"}
+  ["sh" "-c" "env \"PKG_CONFIG_PATH=$HOME/.nix-profile/lib/pkgconfig:$PKG_CONFIG_PATH" pkg-config --print-errors --exists openssl"] {os-distribution = "nixos"}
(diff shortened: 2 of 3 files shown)


kit-ty-kate:
cc @hannesm and @mato the pkg-config part might be of interest of you

packages/conf-libssl/conf-libssl.3/opam
```diff
@@ -0,0 +1,39 @@
...
+  ["pkg-config" "--print-errors" "--exists" "openssl"]
+    {os != "freebsd" & os != "openbsd" & os != "netbsd" & # libssl is provided by base system on BSDs
+     os-distribution != "homebrew" & os-distribution != "macports" & os-distribution != "nixos"}
+  ["sh" "-c" "env \"PKG_CONFIG_PATH=$(brew --prefix openssl)/lib/pkgconfig:$PKG_CONFIG_PATH\" pkg-config --print-errors --exists openssl"] {os-distribution = "homebrew"}
+  ["sh" "-c" "env \"PKG_CONFIG_PATH=$HOME/.nix-profile/lib/pkgconfig:$PKG_CONFIG_PATH\" pkg-config --print-errors --exists openssl"] {os-distribution = "nixos"}
+  ["sh" "-c" "env \"PKG_CONFIG_PATH=/opt/local/lib/pkgconfig:$PKG_CONFIG_PATH\" pkg-config --print-errors --exists openssl"] {os-distribution = "macports"}
+]
+install: [
+  ["sh" "-c" "ln -s \"$(brew --prefix openssl)/lib/pkgconfig/openssl.pc\" '%{lib}%/pkgconfig/openssl.pc'"] {os-distribution = "homebrew"}
+  ["sh" "-c" "ln -s \"$HOME/.nix-profile/lib/pkgconfig/openssl.pc\" '%{lib}%/pkgconfig/openssl.pc'"] {os-distribution = "nixos"}
```
timbertson, line 17: Is this intended to be useful under `opam2nix`, or when running vanilla `opam` on NixOS?

If it's the former, I'm pretty sure it will fail - `$HOME` points to a dummy directory when building a derivation. But I also don't think it's necessary: the transitive dependency on `openssl` and pkg-config should result in [this setup hook](https://github.com/NixOS/nixpkgs/blob/ac6ac61e3f4255ec7ebd9fad36f410dddb0fde7a/pkgs/build-support/pkg-config-wrapper/setup-hook.sh) giving you the right PKG_CONFIG_PATH.

If you're talking about regular opam on NixOS, then I think this would work, if you have already run `nix-env -i openssl` or whatever first.

I see you're using `setenv` in conf-pkg-config, can you just add `setenv: [PKG_CONFIG_PATH += "$HOME/.nix-profile/lib/pkgconfig"]` there for nixos? That would be harmless if the path doesn't exist, and feels cleaner anyway imo.

kit-ty-kate: It's the later. I'm guessing the conf-* packages are ignored by `opam2nix`.

> I see you're using setenv in conf-pkg-config, can you just add setenv: [PKG_CONFIG_PATH += "$HOME/.nix-profile/lib/pkgconfig"] there for nixos? That would be harmless if the path doesn't exist, and feels cleaner anyway imo.

The thing here is that I cannot use shell variable inside `setenv` so sadly I don't think I can do that.

> If you're talking about regular opam on NixOS, then I think this would work, if you have already run nix-env -i openssl or whatever first.

Ok that seems to be what we want in nixos I'm guessing

timbertson: > It's the later. I'm guessing the conf-* packages are ignored by opam2nix.

Nope conf-* aren't treated specially. To `opam2nix`, they look like any other package with `depexts`.

> The thing here is that I cannot use shell variable inside setenv so sadly I don't think I can do that.

Ah, fair enough. FWIW Nix [discourages installing libraries profile-wide in `nix-env`](https://nixos.wiki/wiki/FAQ/I_installed_a_library_but_my_compiler_is_not_finding_it._Why%3F), so I'd suggest dropping this - it's going to make these `conf-*` packages work differently to other libraries, which will be a lot of work to apply uniformly and might be a debugging headache in the future. If you are running `opam` and end up depending on `ssl`, the recommendation is to do so within a `nix-shell -p openssl`, whether it's a literal invocation like that or putting it in a `shell.nix`.

If you do want to keep it, it probably needs to be more tolerant of errors:
 - add `-f` to `-ln` so it doesn't care if the destination path exists (it won't in opam2nix)
 - `mkdir -p '%{lib}%/pkgconfig'` first? You're doing a `mkdir` in `conf-pkg-config`, but the value of `%{lib}%` isn't the same between packages, is it? Even if they are the same path in `opam`, they're not in `opam2nix` so a `mkdir -p` would make this less brittle.

packages/conf-libssl/conf-libssl.3/opam
```diff
@@ -0,0 +1,39 @@
...
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+homepage: "https://www.openssl.org/"
+license: "Apache-1.0"
+build: [
+  ["pkg-config" "--print-errors" "--exists" "openssl"]
+    {os != "freebsd" & os != "openbsd" & os != "netbsd" & # libssl is provided by base system on BSDs
+     os-distribution != "homebrew" & os-distribution != "macports" & os-distribution != "nixos"}
+  ["sh" "-c" "env \"PKG_CONFIG_PATH=$(brew --prefix openssl)/lib/pkgconfig:$PKG_CONFIG_PATH\" pkg-config --print-errors --exists openssl"] {os-distribution = "homebrew"}
+  ["sh" "-c" "env \"PKG_CONFIG_PATH=$HOME/.nix-profile/lib/pkgconfig:$PKG_CONFIG_PATH\" pkg-config --print-errors --exists openssl"] {os-distribution = "nixos"}
+  ["sh" "-c" "env \"PKG_CONFIG_PATH=/opt/local/lib/pkgconfig:$PKG_CONFIG_PATH\" pkg-config --print-errors --exists openssl"] {os-distribution = "macports"}
```
mseri, line 13: I never needed to add this to compile against libraries installed by macports, I think PKG_CONFIG_PATH is set up correctly by macports when you install the distribution

mseri: I am creating a fresh environment to check this. It will take a while :P

MSoegtropIMC: That's also my experience. MacPorts doesn't care to hide system tools if MacPorts binaries are in the PATH. Homebrew installs to a path which is in PATH by default, so it needs to take care to not hide system tools. The MacPorts pkg-config takes the packages installed by MacPorts into account.

mato:
I don't have any strong opinions on how this should work, but looking at how it's implemented I'd suggest using a shell script instead of a series of mutually-exclusive `build:` clauses. The latter is quite hard to follow and looks like a potential maintenance nightmare.

kit-ty-kate:
Ok thanks a lot for all your comments. This PR is now good for a final review and merge.
To summarize, this PR adds:
- `conf-pkg-config.2`: which automatically adds `%{prefix}%/lib/pkgconfig` to your `PKG_CONFIG_PATH`.
  This is useful for anyone using `conf-pkg-config`, I've seen a few packages using the `env OPAM_PKG_CONFIG_PATH=... cmd` to build and this should get rid of that.
- `conf-libssl.3`: which will now automatically symlink the .pc files from libssl to `%{prefix}%/lib/pkgconfig` when using homebrew.
  This improves the user experience of people using homebrew who had to set this environment variable by hand.

timbertson:
👍  LGTM from a NixOS perspective, thanks

packages/conf-pkg-config/conf-pkg-config.2/opam
```diff
@@ -0,0 +1,41 @@
...
+  ["pkgconfig"] {os-distribution = "nixos"}
+  ["devel/pkgconf"] {os = "openbsd"}
+  ["pkg-config"] {os = "macos" & os-distribution = "homebrew"}
+  ["pkgconfig"] {os = "macos" & os-distribution = "macports"}
+  ["pkgconf"] {os = "freebsd"}
+  ["pkgconf-pkg-config"] {os-distribution = "rhel" & os-version >= "8"}
+  ["pkgconf-pkg-config"] {os-distribution = "centos" & os-version >= "8"}
+  ["pkg-config"] {os-distribution = "cygwinports"}
+]
+synopsis: "Virtual package relying on pkg-config installation"
```
MSoegtropIMC, line 37: It is not really a virtual package any more - it does create an opam local pkg-config library. The synopsis / description should reflect this. E.g.

`Check if pkg-config is installed and create an opam switch local pkg-config folder.`

packages/conf-libssl/conf-libssl.3/files/homebrew.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/sh -ex
+
```
MSoegtropIMC, line 2: It would make sense to add a line of comment what this script does, e.g.:

`# symlink the homebrew pkg-config files for openssl to the opam local pkg-config library`

MSoegtropIMC (approved):
Except for my editorial comments, this looks good to me.

mseri (approved):
Good imho, but I agree with @MSoegtropIMC improvements

packages/conf-libssl/conf-libssl.3/files/homebrew.sh
kit-ty-kate: done. Thanks!

packages/conf-pkg-config/conf-pkg-config.2/opam
kit-ty-kate: done thanks!

kit-ty-kate:
Thanks everyone for your inputs!

PR #18065, opened 2021-01-31
Update source hash code for gappa

MSoegtropIMC:
Interestingly the source hash code for gappa changed (the source URL is a commit reference, so the commit did for sure not change).

We had a discussion on Coq Zulip how this can be possible - apparently the contents of a .gitattributes file on the master branch can influence the contents of tar balls for an old commit with gitlab. The difference between the old and new tar ball for the same commit are that the .gitignore files are removed - in line with a .giattributes file modification in a later commit.

Anyway - the hash code needs to be adjusted.

@silene : FYI.

### DIFF
--- a/packages/gappa/gappa.1.3.5/opam
+++ b/packages/gappa/gappa.1.3.5/opam
@@ -36,6 +36,6 @@ depends: [
 ]
 synopsis: "Tool intended for formally proving properties on numerical programs dealing with floating-point or fixed-point arithmetic"
 url {
-  src: "https://gitlab.inria.fr/gappa/gappa/-/archive/f53e105cd73484fc76eb58ba24ead73be502c608.tar.gz"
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Indeed the only difference between the two archives are:
```
Only in old-gappa/doc: .gitignore
Only in old-gappa: .gitignore
Only in old-gappa/src: .gitignore
Only in old-gappa/src/parser: .gitignore
```
For me this PR is good to go. Feel free to mark it as ready to review whenever

MSoegtropIMC:
Sorry, I forgot to remove the "work in progress flag". This is ready to merge.

MSoegtropIMC:
Sorry I forgot why I set this to draft - I wanted to change the commit hash with a tag. Will do so today.

PR #18067, opened 2021-01-31
Package ocaml-protoc-plugin.4.2.0

andersfugmann:
### `ocaml-protoc-plugin.4.2.0`
- enums are mapped to adt's

---

---


andersfugmann:
Oracle Linux 7/8 fails, since they do not supply google protobuf packages containing the compiler `protoc` nor the include files 
(or at least I cannot find the correct packages).

The tests does not run with 32 bit machine words, and I'm not a big fan of conditional compilation.

kit-ty-kate:
> The tests does not run with 32 bit machine words, and I'm not a big fan of conditional compilation.

Do you expect the library itself to be usable on 32bit architecture?

andersfugmann:
Hi @kit-ty-kate,

Thanks for the changes to the PR. Very much appreciated. 
The package will work on a 32 bit machine also. The tests are tailored to test encoding of 63 bit long integers and will not work on 32 bit machines (or compile for that matter).

kit-ty-kate:
Thanks!

PR #18069, opened 2021-02-01
Package lilac.v0.1.1

shnewto:
### `lilac.v0.1.1`
Get the value of any field in a YAML file as a string

---
* Homepage: https://github.com/shnewto/lilac
* Source repo: git+https://github.com/shnewto/lilac
* Bug tracker: https://github.com/shnewto/lilac/issues

---

### DIFF
--- a/packages/lilac/lilac.0.1.1/opam
+++ b/packages/lilac/lilac.0.1.1/opam
@@ -7,21 +7,21 @@ dev-repo: "git+https://github.com/shnewto/lilac"
 bug-reports: "https://github.com/shnewto/lilac/issues"
 tags: [ "library" "lib" "lilac" "yaml" ]
 license: "MIT"
-depends: [ 
-  "cmdliner" 
+depends: [
+  "cmdliner" {with-test & >= "1.0.4"}
   "stdio" {>= "v0.9.0"}
   "base" {>= "v0.9.0"}
   "ounit" {with-test}
-  "yaml" 
+  "yaml" {>= "2.1.0"}
   "dune" {>= "2.7.0"}
   "ocaml" {>= "4.10.0"}
-  "bisect_ppx" {dev & >= "2.5.0"} 
+  "bisect_ppx" {with-test & >= "2.5.0"}
 ]
 build: ["dune" "build" "-p" name "-j" jobs]
 url {
-  src: "https://github.com/shnewto/lilac/archive/v0.1.1.tar.gz"
+  src: "https://github.com/shnewto/lilac/archive/0.1.1.tar.gz"
   checksum: [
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
The original file has issues:
- `dune build` was call without `-p` which should never be the case for releases (warnings are turned into errors)
- `dune install` was used which should almost never be the case as opam installs itself files according to the `<pkg>.install` file generated by dune during the compilation step
- Some constraints were missing. In this case I have no idea which lower-bound constraints you require I'm just making an educated guess knowing that any versions before v0.9.0 are essentially unusable with dune.
- ounit was not marked as a test-only dependency

Could you return those fixes to the opam file upstream?

Aside from that there is also another issue. The version number is not correct for opam. Package don't typically have a `v` prefix, only janestreet packages do as a hack to reset compared to the older versions whose version number were getting out of hand (e.g. 113.13.56). Could you rename the version to `0.1.1` instead of `v0.1.1`?

shnewto:
Oh wow, thanks for the notes and edits @kit-ty-kate 🙌 😄 

I took a swing at addressing your requests and continuing the work you did. Filled in the rest of the version constraints, marked a couple more test-only, and changed the version.

(in case it'd be an issue, I force pushed the c1149ab commit to reword the message, I'd mistyped this pr's number so the generated link in my first commit message was nonsensical)

kit-ty-kate:
`bisect_ppx` is not meant to be used during tests automatically, rather by hand by the developer (you ^^)
Also `cmdliner` was used by the `lilac` binary.

kit-ty-kate:
All good now. Could you return the fix in the opam file upstream? Thanks a lot!

shnewto:
awesome! thanks again @kit-ty-kate! I do have a question about returning the fix to the upstream opam file, do you mean incorporate the latest revision into lilac's 0.1.1 tag/release? I'm unclear on how the checksum is calculated, and whether, in that case, the differences in the opam file would effect the artifact's checksum recorded in this PR's changes.

Sorry if that's a naive question 😅, thanks for your guidance.

kit-ty-kate:
> do you mean incorporate the latest revision into lilac's 0.1.1 tag/release? I'm unclear on how the checksum is calculated, and whether, in that case, the differences in the opam file would effect the artifact's checksum recorded in this PR's changes.

the checksum is simply the md5/sha512 checksum of the archive (which, in your case, is simply an archive created by github using `git archive` for a given tag). The opam file in the archive is relatively useless, so no need to retag or rerelease anything.
By "returning the fix upstream" I mean incorporate the changes to the opam file on the master branch so that the next release won't have the same issues.

shnewto:
Oh okay perfect, thank you!

PR #18047, opened 2021-01-27
Package bls12-381.0.3.15

dannywillems:
### `bls12-381.0.3.15`
OCaml binding for bls12-381 from librustzcash

---
* Homepage: https://gitlab.com/dannywillems/ocaml-bls12-381
* Source repo: git+https://gitlab.com/dannywillems/ocaml-bls12-381.git
* Bug tracker: https://gitlab.com/dannywillems/ocaml-bls12-381/issues

---
:camel: Pull-request generated by opam-publish v2.0.2

PR #18073, opened 2021-02-02
Tezos-crypto should depend on 0.3.15, not 0.3.14.

dannywillems:
Contains https://github.com/ocaml/opam-repository/pull/18047.
The previous version of tezos-crypto does not use the correct version of bls12-381. @pirbo can confirm.


pirbo:
Yes, I confirm that this PR is important for tezos as bls12-381.0.3.14 is (subtlety which make it worse :-) ) broken.

dannywillems:
Also
```
opam repository set-url default https://github.com/dannywillems/opam-repository.git\#8315e03465ca0d8e0838aab2af0e283e590fc1e7
opam install tezos -y
```
should work (in case of requiring a test).

PR #18082, opened 2021-02-03
Hacl-star 0.3.0 1

pirbo:
There were a problem with hacl-star-raw.0.3.0 build system that introduced big performance issues that we fix here.

This change is again important for Tezos :-)


kit-ty-kate:
Thanks

PR #18087, opened 2021-02-04
Flag deprecated packages as deprecated

kit-ty-kate:
Discussed in:
- https://github.com/ocurrent/opam-health-check/issues/8
- https://github.com/ocaml/opam/issues/4522

Implemented as a builtin flag in opam in: https://github.com/ocaml/opam/pull/4523

This PR should not have any effect on any of the packages from opam 2.0 to opam 2.1.0~beta4 and is just there for external tools support for the moment. If https://github.com/ocaml/opam/pull/4523 gets merged, these packages will still be available, just avoided, if possible, by the solver.


kit-ty-kate:
accepted on principal by the opam devs. Can always be changed/reverted if need be.

PR #18071, opened 2021-02-02
Package ssl.0.5.10

smimram:
### `ssl.0.5.10`
Bindings for OpenSSL

---
* Homepage: https://github.com/savonet/ocaml-ssl
* Source repo: git+https://github.com/savonet/ocaml-ssl.git
* Bug tracker: https://github.com/savonet/ocaml-ssl/issues

---
:camel: Pull-request generated by opam-publish v2.0.3

PR #18090, opened 2021-02-04
Fix alpine depext for the latest openssl release

samoht:
Trying to fix https://github.com/ocaml/opam-repository/pull/18071#issuecomment-773495454


kit-ty-kate:
Thanks!

PR #18092, opened 2021-02-04
+ hdfs 0.3

ygrek:
this release fixes build with OCaml >= 4.10 (no unsafe-string)


ygrek:
build failure (`(failed: Fatal: hdfs.h: No such file or directory)`) is expected, because hadoop is not packaged in official debian repositories (idk for other distros)

mseri:
I looked for it (unsuccessfully) for the other distros, I think we cannot do much there

PR #18094, opened 2021-02-05
Add avoid-version to existing packages with hidden-version

dra27:
See ocaml/opam#4527 - this PR would be merged after that one.

We'd want to do the same pattern of `avoid-version` and `hidden-version` just until opam 2.1 is released. After that - i.e. for OCaml 4.13 - we'd just use `avoid-version` since there'd be no version of opam around using `hidden-version` (recall that opam ignores flags it doesn't understand).


kit-ty-kate:
Thanks!

PR #18101, opened 2021-02-06
Bisect_ppx 2.6.0: code coverage analysis

aantron:
[Changelog](https://github.com/aantron/bisect_ppx/releases/tag/2.6.0):

<br/>

> Changes
>
> - Port to ppxlib (aantron/bisect_ppx#327, Sonja Heinze).
> - CSS: use ligature fonts, if installed (aantron/bisect_ppx#339).
> - Internal: switch to Dune cram tests, finding lots of bugs in the process (aantron/bisect_ppx@beb80b3, aantron/bisect_ppx@593a7c2, aantron/bisect_ppx#352).
>
> Bugs fixed
>
> - Workaround for typing issue with instrumentation of or-patterns under GADTs; includes a new `match` instrumenter (aantron/bisect_ppx#325, reported by Mehdi Bouaziz).
> - The second subexpression of `(||)` and `(&&)` is in tail position iff the whole expression is in tail position; Bisect was unconditionally wrapping the second expression in successor code (aantron/bisect_ppx#359, reported by Benjamin Monate).
> - HTML: words that have an instrumentation point on one of their characters were not searchable (aantron/bisect_ppx#346).
> - Don't clobber the global random number generator (aantron/bisect_ppx#344).
> - `(or)` nested inside `(||)` not properly instrumented (aantron/bisect_ppx#347).
> - Subexpressions of `Pexp_newtype` were always treated as if they were in tail position (aantron/bisect_ppx#348).
> - `assert` instrumentation was placed at the wrong point in the generated code (aantron/bisect_ppx#349).
> - `assert false` should not be instrumented; the point is unreachable (aantron/bisect_ppx#355).
> - Instrumentation of or-patterns inside local `open` patterns could trigger unused `open` warning (aantron/bisect_ppx#350).
>
> Removed
>
> *These removals were planned and announced in the 2.0.0 release.*
>
> - PPX option `--mode` (aantron/bisect_ppx#200).
> - PPX option `--no-comment-parsing` (aantron/bisect_ppx#202).
> - PPX option `--exclude` (aantron/bisect_ppx#244).
> - PPX option `--exclude-file`, which has been renamed to `--exclusions` (aantron/bisect_ppx#245).
> - Reporter options `--html`, `--text`, `--coveralls`, which have been replaced by `html`, `summary`, `send-to` subcommands, respectively (aantron/bisect_ppx#145).
> - Support for discovering `.out` files. Bisect switched to extension `.coverage` in 2.0.0 (aantron/bisect_ppx#110).
> - `--dump`, `--csv`, and the old-style text report (aantron/bisect_ppx#251).
> - The old command line, including single-dashed (`-`) multi-character options, and usage of `bisect-ppx-report` without a subcommand (such as `bisect-ppx-report html`) (aantron/bisect_ppx#145).

### DIFF
--- a/packages/bisect_ppx/bisect_ppx.2.6.0/opam
+++ b/packages/bisect_ppx/bisect_ppx.2.6.0/opam
@@ -1,0 +1,47 @@
+opam-version: "2.0"
+
+synopsis: "Code coverage for OCaml"
+version: "2.6.0"
+license: "MIT"
+homepage: "https://github.com/aantron/bisect_ppx"
+doc: "https://github.com/aantron/bisect_ppx"
+bug-reports: "https://github.com/aantron/bisect_ppx/issues"
+
+dev-repo: "git+https://github.com/aantron/bisect_ppx.git"
+authors: [
+  "Xavier Clerc <bisect@x9c.fr>"
+  "Leonid Rozenberg <leonidr@gmail.com>"
+  "Anton Bachin <antonbachin@yahoo.com>"
+]
+maintainer: [
+  "Anton Bachin <antonbachin@yahoo.com>"
+  "Leonid Rozenberg <leonidr@gmail.com>"
+]
+
+depends: [
+  "base-unix"
+  "cmdliner" {>= "1.0.0"}
+  "dune" {>= "2.7.0"}
+  "ocaml" {>= "4.02.0"}
+  "ppxlib" {>= "0.21.0"}
+
+  "ocamlformat" {with-test & = "0.16.0"}
+]
+
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "build" "-p" name "-j" jobs "@compatible"] {with-test}
+]
+
+description: "Bisect_ppx helps you test thoroughly. It is a small preprocessor
+that inserts instrumentation at places in your code, such as if-then-else and
+match expressions. After you run tests, Bisect_ppx gives a nice HTML report
+showing which places were visited and which were missed.
+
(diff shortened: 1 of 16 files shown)


mseri:
There are some failures due to
```
# .ppx/47e007fefa63695257b4d9874e3138b7/ppx.exe: unknown option '-conditional'.
```
This used to be a flag for bisect_ppx, right? Can you have a look?

aantron:
Thanks for the notice, @mseri. I've added constraints.

The `-conditional` flag was [deprecated in 2.0.0](https://github.com/aantron/bisect_ppx/releases/tag/2.0.0) in March 2020, in favor of `--conditional`. Adding that extra leading dash is probably the quickest way to fix each repo.

However, `--conditional` itself has been replaced by a better Dune integration, available since Dune 2.7.0. See the [new instructions](https://github.com/aantron/bisect_ppx#Dune). The new integration makes it easy to turn Bisect into a proper `dev` dependency, so that downstream users do not depend on Bisect.

I'm ccing the maintainers of the affected packages:

- @marcoonroad &mdash; `cuid`
- @zapashcanon &mdash; `exit`, `hc`, `memo`, `omg`, `opazl`, `so`
- @0x0f0f0f &mdash; `gobba`
- @darrenldl &mdash; `reed-solomon-erasure`

For packages that had older releases, I checked the repos for whether they use `-conditional` at the appropriate tags, and constrained those releases accordingly.

kit-ty-kate:
Looks alright now. Thanks a lot!

PR #18110, opened 2021-02-08
2 packages from LexiFi/gen_js_api at 1.0.7

mlasson:
This pull-request concerns:
-`gen_js_api.1.0.7`: Easy OCaml bindings for JavaScript libraries
-`ojs.1.0.7`: Runtime Library for gen_js_api generated libraries

---
* Homepage: https://github.com/LexiFi/gen_js_api
* Source repo: git+https://github.com/LexiFi/gen_js_api.git
* Bug tracker: https://github.com/LexiFi/gen_js_api/issues

---

### DIFF
--- a/packages/conf-npm/conf-npm.1/opam
+++ b/packages/conf-npm/conf-npm.1/opam
@@ -10,7 +10,8 @@ build: [
 depexts: [
   ["npm"] {os-distribution = "alpine"}
   ["npm"] {os-distribution = "arch"}
-  ["npm"] {os-distribution = "centos"}
+  ["epel-release" "npm"] {os-distribution = "centos"}
+  ["npm"] {os-distribution = "ol" & os-version >= "8"}
   ["npm"] {os-family = "debian"}
   ["npm"] {os-distribution = "fedora"}
   ["npm"] {os = "freebsd"}
@@ -20,6 +21,9 @@ depexts: [
   ["nodejs"] {os = "netbsd"}
   ["node"] {os = "openbsd"}
   ["npm"] {os-family = "suse"}
+]
+x-ci-accept-failures: [
+  "oraclelinux-7" # not available
 ]
 synopsis: "Virtual package relying on npm installation"
 flags: conf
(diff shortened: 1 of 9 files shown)


mlasson:
I think the failure reported by opam-ci should be acceptable.

kit-ty-kate:
i had a look at the remaining failures and the failure in `ocaml-js-stdlib` looks like a bug on `gen_js_api`'s side.
The code that fails is generated by `gen_js_api`, worked fine with `gen_js_api.1.0.6` but not with this new release.

The code is generated through a GADT https://github.com/dannywillems/ocaml-js-stdlib/blob/master/js_dom.mli#L203, are there things that changed between 1.0.6 and 1.0.7 in this part of the code?

kit-ty-kate:
I've also sent a PR to fix the tests on Ubuntu 16.04 (and Centos 7 probably) https://github.com/LexiFi/gen_js_api/pull/141

mlasson:
> i had a look at the remaining failures and the failure in `ocaml-js-stdlib` looks like a bug on `gen_js_api`'s side.
> The code that fails is generated by `gen_js_api`, worked fine with `gen_js_api.1.0.6` but not with this new release.
> The code is generated through a GADT https://github.com/dannywillems/ocaml-js-stdlib/blob/master/js_dom.mli#L203, are there things that changed between 1.0.6 and 1.0.7 in this part of the code?

Thanks for the detective work. 
I was aware of this issue ; I should have tell it to spare you the work. 
It is actually more a "Bug Fix" than a "Bug" ; the generated code was wrong even when it was compiling. 
This type should be removed from js_stdlib. 

Would it be possible to add a conflict with "gen_js_api > 1.0.6" in ocaml-js-stdlib ?

There's no support for GADT in gen_js_api. 

> I've also sent a PR to fix the tests on Ubuntu 16.04 (and Centos 7 probably): LexiFi/gen_js_api#141

For some reason, it breaks "macos-latest" CI: 
```
 dune build @runtest
File "node-test/test1/dune", line 12, characters 7-13:
12 |   (run nodejs %{dep:./test.bc.js})))
            ^^^^^^
Error: Program nodejs not found in the tree or in PATH
 (context: default)
```

Note that the node test requires a very recent version 
of node anyway ; so you will only allow the tests to 
fail (with an error message telling to upgrade to node 
> 12 ).

mseri:
What is the state of this?Should we add an upper bound to `ocaml-js-stdlib` and merge? @kit-ty-kate would it be ok for you?

kit-ty-kate:
Thanks! Sorry for the delay

PR #18112, opened 2021-02-08
5 packages from gitlab.com/dannywillems/ocaml-bls12-381/-/archive/0.4.1/ocaml-bls12-381-0.4.1.tar.bz2

dannywillems:
This pull-request concerns:
-`bls12-381.0.4.1`: Virtual package for BLS12-381 primitives
-`bls12-381-gen.0.4.1`: Functors to generate BLS12-381 primitives based on stubs
-`bls12-381-js.0.4.1`: JavaScript version of BLS12-381 primitives implementing the virtual
 package bls12-381
-`bls12-381-js-gen.0.4.1`: Functors to generate BLS12-381 JavaScript primitives based on stubs
-`bls12-381-unix.0.4.1`: UNIX version of BLS12-381 primitives implementing the virtual package
 bls12-381

This new release brings JavaScript support using js_of_ocaml. More information can be found in the [README](https://gitlab.com/dannywillems/ocaml-bls12-381/-/blob/c35d14caa7217931ad47c9434a8ed42cd4aa3d12/README.md).

For the UNIX version, nothing changes compared to before, except adding `bls12-381-unix` when writing an executable.
For the JavaScript version, the integration does rely on a communication with a wasm module, generated by [wasm-bindgen](https://github.com/rustwasm/wasm-bindgen) from the same source code the UNIX package uses ([rustc-bls12-381](https://gitlab.com/dannywillems/rustc-bls12-381/-/tree/bc34f9d806d5b94bb0e3158bd01cc5598e83adcb), integrated in tezos-rust-libs, see https://gitlab.com/tezos/tezos-rust-libs/-/merge_requests/1). [js_of_ocaml](https://github.com/ocsigen/js_of_ocaml) is used on the OCaml side to generate JavaScript code which is going to call the wasm module and read/write the wasm memory.

I am not aware of any previous attempt on releasing OPAM packages using Rust code with a JavaScript support (using the same Rust dependency). This solution does work, but a bunch of improvements can be made. In particular, the JavaScript binding could be improved by writing a library « à la ctypes », see https://gitlab.com/nomadic-labs/tezos/-/issues/230 for the idea. If there is any other attempt, please share. There is an ongoing work to do the same for Sapling in Tezos, and feedbacks are welcome.

I also started (but never finished, even if the most interesting parts, like pointers are not written but used in these packages...) a bunch of README and code to write web applications using OCaml and Rust: https://gitlab.com/dannywillems/web-ocaml-rust-tuto.

---
* Homepage: https://gitlab.com/dannywillems/ocaml-bls12-381
* Source repo: git+https://gitlab.com/dannywillems/ocaml-bls12-381.git
* Bug tracker: https://gitlab.com/dannywillems/ocaml-bls12-381/issues

---

### DIFF
--- a/packages/bls12-381-gen/bls12-381-gen.0.4.1/opam
+++ b/packages/bls12-381-gen/bls12-381-gen.0.4.1/opam
@@ -7,11 +7,11 @@ license: "MIT"
 homepage: "https://gitlab.com/dannywillems/ocaml-bls12-381"
 bug-reports: "https://gitlab.com/dannywillems/ocaml-bls12-381/issues"
 depends: [
-  "ocaml" {>= "4.08" & < "4.12"}
+  "ocaml" {>= "4.08"}
   "dune" {>= "2.7"}
   "ff-sig" {>= "0.6.1" & < "0.7.0"}
   "zarith" {>= "1.10" & < "2.0"}
-  "bisect_ppx" {with-test & >= "2.5" & < "2.6.0"}
+  "bisect_ppx" {with-test & >= "2.5"}
 ]
 build: ["dune" "build" "-j" jobs "-p" name "@install"]
 dev-repo: "git+https://gitlab.com/dannywillems/ocaml-bls12-381.git"
--- a/packages/bls12-381-gen/bls12-381-gen.0.4.1/opam
+++ b/packages/bls12-381-gen/bls12-381-gen.0.4.1/opam
@@ -1,0 +1,25 @@
+opam-version: "2.0"
+synopsis: "Functors to generate BLS12-381 primitives based on stubs"
+description: "Functors to generate BLS12-381 primitives based on stubs"
+maintainer: "Danny Willems <be.danny.willems@gmail.com>"
+authors: "Danny Willems <be.danny.willems@gmail.com>"
+license: "MIT"
+homepage: "https://gitlab.com/dannywillems/ocaml-bls12-381"
+bug-reports: "https://gitlab.com/dannywillems/ocaml-bls12-381/issues"
+depends: [
+  "ocaml" {>= "4.08" & < "4.12"}
+  "dune" {>= "2.7"}
+  "ff-sig" {>= "0.6.1" & < "0.7.0"}
+  "zarith" {>= "1.10" & < "2.0"}
+  "bisect_ppx" {with-test & >= "2.5" & < "2.6.0"}
+]
+build: ["dune" "build" "-j" jobs "-p" name "@install"]
+dev-repo: "git+https://gitlab.com/dannywillems/ocaml-bls12-381.git"
+url {
+  src:
+    "https://gitlab.com/dannywillems/ocaml-bls12-381/-/archive/0.4.1/ocaml-bls12-381-0.4.1.tar.bz2"
+  checksum: [
+    "md5=8e9aa7459420c090c2f664554357adf4"
+    "sha512=d3d11088fe3a338a7c1e6f62aed8590edccabdde47f0d49cc564bc803d9a5e6aef9917b6f07629f0f831defa7c24294b234a8788661c80a4948dd956b35037b8"
+  ]
+}
--- a/packages/bls12-381-js-gen/bls12-381-js-gen.0.4.1/opam
+++ b/packages/bls12-381-js-gen/bls12-381-js-gen.0.4.1/opam
@@ -1,0 +1,37 @@
+opam-version: "2.0"
+synopsis:
+  "Functors to generate BLS12-381 JavaScript primitives based on stubs"
+description:
+  "Functors to generate BLS12-381 JavaScript primitives based on stubs"
+maintainer: "Danny Willems <be.danny.willems@gmail.com>"
+authors: "Danny Willems <be.danny.willems@gmail.com>"
+license: "MIT"
+homepage: "https://gitlab.com/dannywillems/ocaml-bls12-381"
+bug-reports: "https://gitlab.com/dannywillems/ocaml-bls12-381/issues"
+depends: [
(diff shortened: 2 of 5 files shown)


dannywillems:
Looks like there is an issue with ocamlformat.0.16.0 :confused:

dannywillems:
Also, looks like npm is missing. I am not sure using `npm` from the package manager is the best. Should I use anyway? Tests require it as it does run in node.

kit-ty-kate:
> Looks like there is an issue with ocamlformat.0.16.0 confused

what was the issue? I couldn't see any failures related to ocamlformat

kit-ty-kate:
I pushed a fix, could you return the fix upstream?

dannywillems:
> I pushed a fix, could you return the fix upstream?

Done in https://gitlab.com/dannywillems/ocaml-bls12-381/-/commit/fe9bfca313955a00e7cd11f725e7b4bac8eee1e5

dannywillems:
> > Looks like there is an issue with ocamlformat.0.16.0 confused
> 
> what was the issue? I couldn't see any failures related to ocamlformat

Builds failed with a cyclic deps error for bisect_ppx 2.6 and ocamlformat 0.16.0.

dannywillems:
Like [here](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/ac0e7a2112c7706aff21c67abede5ce9bf90fd7c/variant/compilers,4.08,bls12-381-js-gen.0.4.1,tests)

dannywillems:
It looks like more packages are affected by the circular dependencies. The CI says it is fine, but the commands fail with an error. It is introduced by bisect_ppx 2.6.0. bisect_ppx is a dependency of `ff` and `ff-pbt`, with a version ">= 2.5". I will add a temporary fix here with `bisect_ppx < "2.6.0"`.

dannywillems:
Some jobs are failing because the node version is old, see `- node: bad option: --experimental-wasm-modules
` in [here](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/cff08c2978074e67f461d28494d578dd201f214b/variant/compilers,4.08,bls12-381-js-gen.0.4.1,tests). Tests require the option `--experimental-wasm-modules`, introduced in `Node 12.3.0`, see https://nodejs.org/tr/blog/uncategorized/10-lts-to-12-lts/#notable-changes-in-node-js-12-3-0.
And I suppose an old version in the tests (by using `conf-npm` which uses the system packages, not recommended for NPM/Node).

For OCaml 4.12, should I add a constraint on the OCaml version as jsoo does? https://github.com/ocaml/opam-repository/blob/master/packages/js_of_ocaml-compiler/js_of_ocaml-compiler.3.8.0/opam#L17 and https://github.com/ocaml/opam-repository/blob/master/packages/js_of_ocaml-compiler/js_of_ocaml-compiler.3.7.1/opam.

kit-ty-kate:
Is there no way of having tests without `npm`? If not I'll simply disable them.

dannywillems:
> Is there no way of having tests without `npm`? If not I'll simply disable them.

No it is not possible.

packages/bls12-381-gen/bls12-381-gen.0.4.1/opam
```diff
@@ -0,0 +1,25 @@
+opam-version: "2.0"
+synopsis: "Functors to generate BLS12-381 primitives based on stubs"
+description: "Functors to generate BLS12-381 primitives based on stubs"
+maintainer: "Danny Willems <be.danny.willems@gmail.com>"
+authors: "Danny Willems <be.danny.willems@gmail.com>"
+license: "MIT"
+homepage: "https://gitlab.com/dannywillems/ocaml-bls12-381"
+bug-reports: "https://gitlab.com/dannywillems/ocaml-bls12-381/issues"
+depends: [
+  "ocaml" {>= "4.08" & < "4.12"}
```
kit-ty-kate, line 10: One last question and it should be good to go: why is it not available on 4.12? If there is any issue I'll fix your package but if that's "just in case" I'd rather not to have that. That would add work to both of us:
- you by making another release for 4.12
- me by having to understand why the package was constrained in the first place and pushing the change anyway.

If it's a real incompatibility however that's totally fine.

dannywillems: It is compatible with 4.12, but the CI here was failing because jsoo was not available for 4.12. I thought adding the dep was the way to go.

kit-ty-kate: I'm surprised, was it really failing or just skipped? It's not supposed to fail.

packages/bls12-381-gen/bls12-381-gen.0.4.1/opam
```diff
@@ -0,0 +1,25 @@
...
+  "dune" {>= "2.7"}
+  "ff-sig" {>= "0.6.1" & < "0.7.0"}
+  "zarith" {>= "1.10" & < "2.0"}
+  "bisect_ppx" {with-test & >= "2.5" & < "2.6.0"}
```
kit-ty-kate, line 14: Same question with that actually. I believe this is just a bug fixed by https://github.com/ocaml/opam-repository/pull/18141

dannywillems: Yes I do confirm `< "2.6.0"` is just a temporary workaround because the tests were failing in the CI here.

kit-ty-kate:
Thanks!

dannywillems:
Thanks for merging.
Looks like https://github.com/ocaml/dune/issues/4195 does also happen when running `opam install bls12-381-unix`. I was a bit confident it passed the CI here.

PR #18114, opened 2021-02-09
dns: rework the conflicts a bit

samoht:


packages/dns/dns.0.6.1/opam
```diff
@@ -29,12 +39,11 @@ depends: [
...
 ]
-depopts: ["mirage-net"]
+depopts: ["mirage-net" "base-unix"]
 conflicts: [
-  "mirage-net" {> "0.9.4"}
-  "mirage-net" {< "0.5.2"}
+  "mirage-net" {< "0.5.2" | > "0.9.4"}
 ]
-dev-repo: "git://github.com/mirage/ocaml-dns"
+available: false # need a constraint on conf-libssl (only in opam 2.1)
```
mseri, line 46: You mean that these were not installable before?

avsm: sort of -- the underlying OS has upgraded, so it _used_ to be build on an older openssl, but no longer does. When we can express "version of depexts" constraints, this will be buildable again. It's a pity to mark it unavailable when it would techncially build on an older debian, but that cant be helped I think.

packages/dns/dns.0.5.0/opam
```diff
@@ -27,10 +37,9 @@ depends: [
   "uri"
   "ocamlbuild" {build}
 ]
-depopts: ["lwt" "mirage-net"]
+depopts: ["lwt" "mirage-net" "async"]
 conflicts: [
-  "lwt" {< "2.4.1"}
-  "lwt" {>= "3.0.0"}
+  "lwt" {< "2.4.1" | >= "3.0.0"}
```
kit-ty-kate, line 42: why change this? Does this impact the solver? Oo

packages/dns/dns.0.6.1/opam
kit-ty-kate: I don't understand the `(only in opam 2.1)` bit of the comment. What was possible in opam 2.1 that is not in 2.0 that would end up in marking it as available false?

PR #18115, opened 2021-02-09
irmin: fix base64 constraint

samoht:


samoht:
so no more time outs for `irmin-git` 🎉 

(but still time outs for `irmin-unix` and `irmin-indexeddb` -- will fix that later if I manage to understand what happens)

kit-ty-kate:
what was this fix for? Correctness or "tweak until the solver doesn't timeout" ?

PR #18118, opened 2021-02-09
[release] elpi 1.13.0

gares:


gares:
It fails on a handful of linux distributions which done have `time` installed...

packages/conf-time/conf-time.1/opam
```diff
@@ -4,13 +4,16 @@ authors: ["POSIX"]
...
 depexts: [
   ["time"] {os-family = "debian"}
+  ["time"] {os-family = "ubuntu"}
   ["time"] {os-distribution = "centos"}
   ["time"] {os-distribution = "fedora"}
-  ["time"] {os-distribution = "arch"}
+  ["time"] {os-distribution = "ol"}
+  ["time"] {os-family = "suse"}
+  ["time"] {os-family = "arch"}
+  # busybox time is installed by default on Alpine, GNU-time is time@testing
```
gares, line 16: Busybox time is nothing like gnu time. If there is a depext for alpine, it should be used IMO.

gares: This is also why elpi tests are disabled on alpine

gares:
I checked a couple of failures, and they are timeouts in the test suite.
If this new CI is using slowed VMs we either increase the timeout (currently 120s set in the opam file) or disable some tests (which requires me to add support for it, the is currently a way to only run the tests that match a name, but not to exclude the the ones testing for performance regression)

packages/conf-time/conf-time.1/opam
kit-ty-kate: true, but the `@testing` packages are rather flacky. I'll let it as is for now, we can think about it for next time

kit-ty-kate:
Thanks!

gares:
thanks to you!

PR #18124, opened 2021-02-10
Remove dependency cycles

samoht:


samoht:
This makes `opam admin check --cycle --ignore-test-doc` return without error.

samoht:
It's probably not worth testing the revdpes on the entiere universe. @dra27 and @kit-ty-kate are you ok to merge this?

packages/graphics/graphics.5.1.1/opam
```diff
@@ -22,6 +22,7 @@ depends: [
   "conf-pkg-config" {os != "win32"}
   "ocaml" {>= "4.09.0~~"}
 ]
+conflicts: ["ocamlfind-secondary"]
```
kit-ty-kate, line 25: same question here. Something tells me the constraint should be in ocamlfind-secondary.

samoht:
@kit-ty-kate there is a cycle between these packages which confuses the solver:

```
dune (>= 2.0.0 & < 2.1.1) -> ocamlfind-secondary = 1.8.1 -> ocamlfind = 1.8.1 -> graphics = 5.0.0
```

This PR breaks that cycle explicitly by adding a conflict in the middle of the dependency chain. That same conflict can be added either in `dune` or `ocamlfind` or `grapics` but not in `ocamlfind-secondary` as it's not part of the chain. I can add a comment in the file to make that clearer if you think that's useful.

samoht:
@kit-ty-kate I've fixed the location of the `graphics` constraint ; for the result one it's easier to patch the old `result.1.3.` than all the existing (and future) release of `dune-configurator`. Is that ok with you?

samoht:
The revdep testing are killing the CI. @kit-ty-kate is this ok if I merge this quickly?

kit-ty-kate:
This PR might not be necessary as `opam admin check --cycle` might not take coinstallability into account according to @AltGr. I suggest to not merge until we know more.
Opam issue: https://github.com/ocaml/opam/issues/4541

samoht:
Closing as it needs caml/opam#4541 to be resolved first.

PR #18125, opened 2021-02-10
Remove doc and test dependency cycles

samoht:
And here the remaining cycle fixes to complement https://github.com/ocaml/opam-repository/pull/18124.

With these 2 PRs, `opam admin check --cycle` returns no errors.

### DIFF
--- a/packages/alcotest/alcotest.0.4.6/opam
+++ b/packages/alcotest/alcotest.0.4.6/opam
@@ -19,7 +19,8 @@ depends: [
   "result"
   "cmdliner"
 ]
-synopsis: "Alcotest is a lightweight and colourful test framework."
+conflicts:["odoc"]
+synopsis: "Alcotest is a lightweight and colourful test framework"
 description: """
 Alcotest exposes simple interface to perform unit tests. It exposes
 a simple `TESTABLE` module type, a `check` function to assert test
--- a/packages/alcotest/alcotest.0.4.7/opam
+++ b/packages/alcotest/alcotest.0.4.7/opam
@@ -19,7 +19,8 @@ depends: [
   "result"
   "cmdliner"
 ]
-synopsis: "Alcotest is a lightweight and colourful test framework."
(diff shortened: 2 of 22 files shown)


samoht:
@rgrinberg I've removed `odoc` from the `with-doc` dependencies of `dune` as it creates a cycle between `odoc` and `dune`.

@c-cube same for `qtest` and `odoc`.

@AltGr reckons it's an actual issues as if you try to install these packages with `opam install foo --with-doc` you'll get an error. I suspect the way forward is to have docs.ocaml.org anyway, which is actively worked on by @jonludlam

c-cube:
Sorry, what's the cycle involving qtest?

samoht:
@c-cube `odoc` is using `tyxml` which is using `alcotest` which is using `stringext` which is using `qtest` which is using `odoc` (for some versions of these packages):

```
tyxml (>= 4.0.0 & < 4.3.0) -> alcotest (>= 0.4.3 & < 0.4.8) -> stringext >= 1.4.3 -> qtest (>= 2.5 & < 2.11.1) ->
    qcheck (>= 0.9 & < 0.11) -> qcheck-ounit < 0.11 -> odoc = 1.3.0
```

samoht:
(or check the output of `opam admin check --cycle`)

samoht:
Note that I can also just add a conflict in the middle of the cycle. Will try this to avoid changing all the qcheck packages.

samoht:
Ok I've reverted the changes to `dune-configurator` and `qheck` : now only old versions of `alcotest` holds that conflict. That'll be easier to maintain (as it just fixes old packages).

kit-ty-kate:
This PR might not be necessary as `opam admin check --cycle` might not take coinstallability into account according to @AltGr. I suggest to not merge until we know more.
Opam issue: https://github.com/ocaml/opam/issues/4541

samoht:
Closing until caml/opam#4541 is resolved.

PR #18127, opened 2021-02-10
Add delimcc.2020.10.08

yallop:
Older versions of `delimcc` (2018.03.16 and earlier) don't work with recent versions of OCaml (4.10.0 onwards).  There's a [bug report on discuss.ocaml.org](https://discuss.ocaml.org/t/bug-in-delimcc-with-ocaml-4-10/6534) with a program that illustrates the problem.

This PR adds the latest version of `delimcc` (`2020.10.08`) and constrains older versions to OCaml versions below 4.10.0.

/cc @gasche, @rgrinberg (listed as maintainers of the earlier versions)

### DIFF
--- a/packages/delimcc/delimcc.2017.03.02/opam
+++ b/packages/delimcc/delimcc.2017.03.02/opam
@@ -9,7 +9,7 @@ build: [
 ]
 remove: [["ocamlfind" "remove" "delimcc"]]
 depends: [
-  "ocaml" {>= "4.04.0" & < "4.12.0"}
+  "ocaml" {>= "4.04.0" & < "4.10.0"}
   "ocamlfind"
 ]
 install: [make "findlib-install"]
--- a/packages/delimcc/delimcc.2018.03.16/opam
+++ b/packages/delimcc/delimcc.2018.03.16/opam
@@ -8,7 +8,7 @@ build: [
   [make "opt"]
 ]
 depends: [
-  "ocaml" {>= "4.04.0" & < "4.12.0"}
+  "ocaml" {>= "4.04.0" & < "4.10.0"}
   "ocamlfind" {build}
 ]
 install: [make "findlib-install"]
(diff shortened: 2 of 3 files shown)


gasche:
In case anyone is curious, the new release (stops using `caml_fatal_error`, using its own `printf + exit(2)` instead, and) adds the following diff:

```diff
+ /* Prior to 4.10 we used the following function */
+ #if 0
extern void caml_raise_exception (const value bucket) Noreturn;
+ #endif
+ /* but now it takes the extra argument, the caml_state. Luckily,
+   there is a new function with the old interface, which also
+   does some clean-up.
+   So we should use it.
+ */
+ extern void caml_raise (const value bucket) Noreturn;
+ #define RAISE(exc) caml_raise(exc)
```

(I don't understand what the RAISE macro is for, to more easily change the name in the future?)

If I understand correctly, the upstream change that affected Oleg's code is https://github.com/ocaml/ocaml/commit/fc6f0284922621fc8717033a8d73b1134f1506ac .

`caml_raise_exception` is an internal function that was not part of the OCaml public API. `caml_raise` was available for a long time to do this (which is why the change is backward-compatible with <4.10 versions). But the usage of `extern void <function>` is not such a good practice, because it lets you use any (possibly-internal) runtime function, it would be nicer to include `fail.h` and use `caml_raise` exported there directly.

All this to say: it would make sense to have a global look at the delimcc code with the purpose of using non-internal APIs to improve forward-compatibility, but the change itself sounds fine and of course we should release the next version on OPAM anyway.

kit-ty-kate:
Thanks a lot!

Btw if any of you have some spare time to look at the delimcc port to OCaml 4.12 (https://github.com/kit-ty-kate/delimcc/pull/1) that would make me very happy.

gasche:
I may try to have a look. Note that there was a previous delimcc mirror on github:
  https://github.com/zinid/delimcc

Issue #18128, opened 2021-02-10
conf-gmp build fails on NetBSD


azarens-git:
## Report

opam install conf-gmp fails as follows:

```
<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
[ERROR] The compilation of conf-gmp failed at "/bin/sh -exc cc -c $CFLAGS -I/usr/local/include test.c".

#=== ERROR while compiling conf-gmp.3 =========================================#
# context     2.0.6 | netbsd/x86_64 | ocaml-base-compiler.4.11.1 | https://opam.ocaml.org#4acae195
# path        ~/.opam/ocaml-base-compiler.4.11.1/.opam-switch/build/conf-gmp.3
# command     /bin/sh -exc cc -c $CFLAGS -I/usr/local/include test.c
# exit-code   1
# env-file    ~/.opam/log/conf-gmp-9842-ffb3fd.env
# output-file ~/.opam/log/conf-gmp-9842-ffb3fd.out
### output ###
# + cc -c -I/usr/local/include test.c
# test.c:1:10: fatal error: gmp.h: No such file or directory
#  #include <gmp.h>
#           ^~~~~~~
# compilation terminated.

```

## Analysis

Looking at the corresponding opam definitions:
```
build: [
  ["sh" "-exc" "cc -c $CFLAGS -I/usr/local/include test.c"] {os != "macos"}
  [
    "sh"
    "-exc"
    "cc -c $CFLAGS -I/opt/homebrew/include -I/opt/local/include -I/usr/local/include test.c"
  ] {os = "macos"}
]
```

It is rather plain that the author had decided to hardcode include paths, instead of using pkg-config.  The pkg-config on this NetBSD host produces the following correct output:

```
$ pkg-config --cflags gmp
-I/usr/pkg/include
```

Please fix the issue.

azarens-git:
I would like to particularly encourage the use of pkg-config approach with this package instead of hardcoding `/usr/pkg/include` for another reason as well.  It is not unheard of and is certainly not unreasonable for a NetBSD system administrator to change the location of **LOCALBASE** from `/usr/pkg` to another value.  In fact, if pkgsrc is used to build an environment in some unprivileged location, like e.g. a home directory, then the value of **LOCALBASE** would certainly be distinct from `/usr/pkg`.

kit-ty-kate:
> It is rather plain that the author had decided to hardcode include paths, instead of using pkg-config.

[...]

PR #18129, opened 2021-02-11
Improve conf-gmp: Use pkg-config when available

kit-ty-kate:
Fixes issue with pkgsrc (NetBSD)
Fixes https://github.com/ocaml/opam-repository/issues/18128

cc @azarens-git

### DIFF
--- a/packages/conf-gmp/conf-gmp.4/files/test.c
+++ b/packages/conf-gmp/conf-gmp.4/files/test.c
@@ -1,0 +1,10 @@
+#include <gmp.h>
+#ifndef __GMP_H__
+#error "No GMP header"
+#endif
+
+void test(void) {
+  mpz_t n;
+  mpz_init(n);
+  mpz_clear(n);
+}
--- a/packages/conf-gmp/conf-gmp.4/opam
+++ b/packages/conf-gmp/conf-gmp.4/opam
@@ -1,0 +1,32 @@
+opam-version: "2.0"
+maintainer: "https://github.com/ocaml/opam-repository/issues"
+authors: "GMP dev team"
+homepage: "http://gmplib.org/"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: "GPL-1.0-or-later"
+build: [
+  ["sh" "-c" "pkg-config --print-errors --exists gmp || cc -c $CFLAGS -I/usr/local/include test.c"]
+]
+depends: [
+  "conf-pkg-config" {build}
+]
+depexts: [
+  ["libgmp-dev"] {os-family = "debian"}
+  ["libgmp-dev"] {os-family = "ubuntu"}
+  ["gmp-dev"] {os-family = "alpine"}
+  ["gmp" "gmp-devel"] {os-distribution = "centos"}
+  ["gmp" "gmp-devel"] {os-distribution = "fedora"}
+  ["gmp" "gmp-devel"] {os-distribution = "ol"}
+  ["gmp-devel"] {os-family = "suse"}
+  ["gmp"] {os-family = "arch"}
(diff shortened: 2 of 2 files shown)


kit-ty-kate:
Underlying packages (at least most notably `zarith`) do not themselves use `pkg-config`. Some more testing is required.

azarens-git:
## Analysis

> Underlying packages (at least most notably `zarith`) do not themselves use `pkg-config`. Some more testing is required.

As far as ocaml builds are concerned, the ocaml-specific packages are usually robustly managed (either by ocamlfind or opam), so the general challenge is with interfacing with system library dependencies from OCaml.  In this case, the dependency has been **by** zarith **upon** system gmp.

#### Further considerations

Now, it might be plausible to use the `/usr/pkg/include` and `/usr/pkg/lib` directories as fallbacks for NetBSD case whence `pkg-config` is not present.  If it would be possible to add a conditional native dependency on **pkg-config** (corresponding native package name is **pkgconf**) for builds on NetBSD, that would be even more general and robust.  I would strongly advise against the default fallback directories, and would strongly advise to conditionally add depopt on pkgconf.  The latter is never a burden to install on a pkgsrc-based system, and the native **gmp** package alway has _.pc_ files available, so all it needs is just a small binary.

## Tests

Also, I could confirm that zarith builds successfully on NetBSD 9.1_STABLE box with switches 4.11.1, 4.11.1+flambda, and 4.11.1+spacetime with the following opam configuration:

```
opam-version: "2.0"
maintainer: "Xavier Leroy <xavier.leroy@inria.fr>"
authors: [
  "Antoine Miné"
  "Xavier Leroy"
  "Pascal Cuoq"
]
homepage: "https://github.com/ocaml/Zarith"
bug-reports: "https://github.com/ocaml/Zarith/issues"
dev-repo: "git+https://github.com/ocaml/Zarith.git"
build: [
  ["./configure"] {os != "netbsd" & os != "openbsd" & os != "freebsd" & os != "macos"}
  [
    "sh"
    "-exc"
    "LDFLAGS=\"$LDFLAGS $(pkg-config --libs-only-L gmp)\" CFLAGS=\"$CFLAGS $(pkg-config --cflags-only-I gmp)\" ./configure"
  ] {os = "netbsd"}
  [
    "sh"
    "-exc"
    "LDFLAGS=\"$LDFLAGS -L/usr/local/lib\" CFLAGS=\"$CFLAGS -I/usr/local/include\" ./configure"
  ] {os = "openbsd" | os = "freebsd"}
  [
    "sh"
    "-exc"
    "LDFLAGS=\"$LDFLAGS -L/opt/homebrew/lib -L/opt/local/lib -L/usr/local/lib\" CFLAGS=\"$CFLAGS -I/opt/homebrew/include -I/opt/local/include -I/usr/local/include\" ./configure"
  ] {os = "macos"}
  [make]
]
install: [
  [make "install"]
]
depends: [
  "ocaml"
  "ocamlfind" {build}
  "conf-gmp"
  "conf-perl" {build}
]
synopsis:
  "Implements arithmetic and logical operations over arbitrary-precision integers"
description: """
The Zarith library implements arithmetic and logical operations over
arbitrary-precision integers. It uses GMP to efficiently implement
arithmetic over big integers. Small integers are represented as Caml
unboxed integers, for speed and space economy."""
url {
  src: "https://github.com/ocaml/Zarith/archive/release-1.11.tar.gz"
  checksum: [
    "md5=95e8cb358b70f156328c552f1bde8d81"
    "sha512=7c0645d63511d42b763cd0a33b3fb139c26d23ae0cf59cd651388644cb8e7c43a83265d55d69ff3320d8c4586c4f8dbd9c3412197bab2e110c38e6e7dac88b83"
  ]
}
```

PR #18132, opened 2021-02-11
[new release] mdx (1.8.1)

NathanReb:
Executable code blocks inside markdown files

- Project page: <a href="https://github.com/realworldocaml/mdx">https://github.com/realworldocaml/mdx</a>

#### Changed

- Dropped OMP dependency and use handwritten compat layers instead
  (realworldocaml/mdx#317, @NathanReb)

### DIFF
--- a/packages/mdx/mdx.1.8.1/opam
+++ b/packages/mdx/mdx.1.8.1/opam
@@ -37,7 +37,7 @@ depends: [
   "alcotest" {with-test}
 ]
 build: [
-  ["dune" "subst"] {pinned}
+  ["dune" "subst"] {dev}
   [
     "dune"
     "build"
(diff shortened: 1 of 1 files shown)


packages/mdx/mdx.1.8.1/opam
```diff
@@ -37,7 +37,7 @@ depends: [
   "alcotest" {with-test}
 ]
 build: [
-  ["dune" "subst"] {pinned}
+  ["dune" "subst"] {dev}
```
NathanReb, line 40: This will need to be patched in dune's opam file generation as this is was the auto-generated build command!

kit-ty-kate: This is already patched since dune 2.7 but this requires at least `(lang dune 2.7)`

kit-ty-kate: This commit was also mostly to retrigger CI as most of the revdeps failed due to dependency cycles: https://github.com/ocaml/opam-repository/pull/18141
Usually I ignore those unless I need to modify opam file already.

kit-ty-kate:
Thanks a lot!

Issue #18134, opened 2021-02-11
Cyclic dependencies involving bisect_ppx 2.6.0


MattWindsor91:
I noticed that some of my GitHub Actions builds for [travesty](https://github.com/MattWindsor91/travesty) using [avsm/setup-ocaml@v1](https://github.com/avsm/setup-ocaml) were failing with the following invocation and issue:

```
opam pin add travesty.dev . --no-action
opam depext travesty --yes --with-doc --with-test
# Detecting depexts using vars: arch=x86_64, os=linux, os-distribution=ubuntu, os-family=debian
Error:  No solution for travesty: The actions to process have cyclic dependencies:
          - ∗ ocamlformat.0.16.0 → ∗ bisect_ppx.2.6.0 → ∗ odoc.1.5.2 → ∗ ocamlformat.0.16.0
          - ∗ ocamlformat.0.16.0 → ∗ bisect_ppx.2.6.0 → ∗ odoc.1.5.2 → ∗ dune-configurator.2.8.2 → ∗ base.v0.14.1 → ∗ stdio.v0.14.0 → ∗ ocamlformat.0.16.0
          - ∗ ocamlformat.0.16.0 → ∗ bisect_ppx.2.6.0 → ∗ odoc.1.5.2 → ∗ dune-configurator.2.8.2 → ∗ base.v0.14.1 → ∗ ppxlib.0.21.0 → ∗ ocamlformat.0.16.0
          - ∗ bisect_ppx.2.6.0 → ∗ odoc.1.5.2 → ∗ dune-configurator.2.8.2 → ∗ base.v0.14.1 → ∗ ppxlib.0.21.0 → ∗ bisect_ppx.2.6.0
          - ∗ ocamlformat.0.16.0 → ∗ bisect_ppx.2.6.0 → ∗ odoc.1.5.2 → ∗ dune-configurator.2.8.2 → ∗ base.v0.14.1 → ∗ ocamlformat.0.16.0
          - ∗ ocamlformat.0.16.0 → ∗ bisect_ppx.2.6.0 → ∗ odoc.1.5.2 → ∗ dune-build-info.2.8.2 → ∗ ocamlformat.0.16.0
```

I'm not sure where `bisect_ppx` has turned up in the dependency graph here, as it wasn't installed on my local machine at any point while working on the library.  Nonetheless, it broke every run that I did, on ocaml version 4.08.1 through 4.11 and on both ubuntu and macOS.

[...]

PR #18135, opened 2021-02-11
[new release] travesty (0.7.0)

MattWindsor91:
**Note:** `travesty` currently has a `conflicts` against `bisect_ppx`, which is a quick-fix for issue
#18134 - this will likely become unnecessary when PR #18125 is merged.

CHANGES:

Major release with large breaking changes, and added and removed features.

## Breaking changes

- `Traversable` and `Bi_traversable` now encode applicative traversals, not
  monadic traversals.  This means that every use of a `Make` functor needs
  changing such that, instead of an `On_monad` functor over `Monad.S`, they
  expose an `On` functor over `Applicative.S`.  No function names have changed
  (they remain `map_m`, `iter_m` etc. for backwards compatibility), and the
  various `Make` functors generate both `On` (applicative) and `On_monad`
  (monadic) functors, so most code that _consumes_ traversals should be fine.
- Some traversal implementations in `Base_exts` have changed; this may subtly
  affect side effect processing order.

## Added

- `Monad_exts.App` is `Base.Applicative.Of_monad`, but keeps the original
  monad type around; this is useful for passing monads to things that expect
  applicatives.
- `Monad_exts.Let` produces OCaml 4.08-style `let+` and `let*` bindings for
  monads.  It is included in `Monad_exts.Make`.

## Removed

- The `Traversable.Helpers` module; in practice this doesn't seem to be very
  useful compared to writing the traversable boilerplate manually, and it
  doesn't gel particularly well with the planned monad to applicative change.
- Support for OCaml v4.07.  This has effectively been broken since v0.6.2 when
  building on Jane Street libraries v0.14.

### DIFF
--- a/packages/travesty/travesty.0.7.0/opam
+++ b/packages/travesty/travesty.0.7.0/opam
@@ -1,0 +1,48 @@
+opam-version: "2.0"
+synopsis: "Traversable containers, monad extensions, and more"
+description: """
+'Travesty' is a library for defining containers with applicative
+or monadic traversals, inspired by Haskell's Traversable typeclass; it also
+contains various helpers for monadic code, such as state transformers and
+extension functions for common monads and containers.  It sits on top of Jane
+Street's Base library and ecosystem."""
+maintainer: ["Matt Windsor <mattwindsor91@gmail.com>"]
+authors: ["Matt Windsor <mattwindsor91@gmail.com>"]
+license: "MIT"
+homepage: "https://MattWindsor91.github.io/travesty/"
+doc: "https://MattWindsor91.github.io/travesty/"
+bug-reports: "https://github.com/MattWindsor91/travesty/issues"
+depends: [
+  "dune" {>= "2.0"}
+  "ocaml" {>= "4.08"}
+  "ppx_jane" {>= "v0.12.0" & < "v0.15.0"}
+  "ppx_expect" {with-test & >= "v0.12.0" & < "v0.15.0"}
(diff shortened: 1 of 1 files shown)


packages/travesty/travesty.0.7.0/opam
```diff
@@ -0,0 +1,48 @@
...
+bug-reports: "https://github.com/MattWindsor91/travesty/issues"
+depends: [
+  "dune" {>= "2.0"}
+  "ocaml" {>= "4.08"}
+  "ppx_jane" {>= "v0.12.0" & < "v0.15.0"}
+  "ppx_expect" {with-test & >= "v0.12.0" & < "v0.15.0"}
+  "base" {>= "v0.12.0" & < "v0.15.0"}
+]
+conflicts: [
+  "bisect_ppx" {>= "2.6.0"}
```
kit-ty-kate, line 23: why is there a conflict with bisect_ppx 2.6.0?

MattWindsor91: brute-force quickfix to stop dependency cycles involving `odoc` and `bisect_ppx` (which were causing my CI to fail and possibly would've caused opam CI to fail too?), but https://github.com/ocaml/opam-repository/pull/18141 _should_ fix them and make the conflict unnecessary.

Besides, now I come to think of it, there's probably a way of specifying a build-time-only conflict, in which case this should've been one of those.

kit-ty-kate:
I'll let CI run again and this should be good to go

kit-ty-kate:
Thanks!

PR #18139, opened 2021-02-11
Package portia.1.5

rixed:


packages/portia/portia.1.5/opam
```diff
@@ -0,0 +1,42 @@
...
+  "ocamlfind" {build}
+  "batteries"
+  "qtest" {with-test}
+]
+depexts: [
+  ["asciidoc"] {os-family = "debian"}
+]
+post-messages: [
+  "This package requires asciidoc to build the doc." {failure}
+]
```
kit-ty-kate, line 36: Do you have time to create a new `conf-asciidoc` package? This way the documentation would be available everywhere, maintainance would be reduced and the dependency would be removed during normal execution (non-doc).

If you don't have time I'll do it at some point

rixed: I can not look into this in the near future but maybe in a few weeks.

kit-ty-kate: That's ok, I've just done it ^^

kit-ty-kate:
Thanks!

PR #18141, opened 2021-02-12
odoc: bisect_ppx is not used during tests

kit-ty-kate:
Should fix https://github.com/ocaml/opam-repository/issues/18134


kit-ty-kate:
@jonludlam is that ok? If so would it be possible to return this fix upstream in the 1.x branch (if there is one) and master?

PR #18144, opened 2021-02-12
[new release] postgresql (5.0.0)

mmottl:

##### CHANGES:

* Added support for parameter types to `exec`, `prepare`, `send_query`,
    and `send_prepare`.

    Thanks to Petter A. Urkedal for the patch!


kit-ty-kate:
Thanks! Could you return the fix in the opam file upstream? (I know you're using an auto-generated opam file, so at least the change in `license`, the rest can be fixed if you upgrade to `(lang dune 2.7)`, and I'm not sure for the comment though since you're using dune-release, it's supposed to do it automatically)

mmottl:
@kit-ty-kate Thanks, I'll do that for all my packages.

PR #18145, opened 2021-02-12
conf-aclocal: Make it an alias to conf-automake (more complete, duplicated package)

kit-ty-kate:
Fixes duplicating as well as issues on macOS and other distributions


packages/conf-aclocal/conf-aclocal.2/opam
```diff
@@ -0,0 +1,12 @@
+opam-version: "2.0"
+maintainer: "unixjunkie@sdf.org"
+authors: "https://git.savannah.gnu.org/cgit/automake.git/tree/AUTHORS"
+homepage: "http://www.gnu.org/software/automake/"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: "GPL-1.0-or-later"
+build: ["aclocal" "--version"]
```
toots, line 7: This should probably be removed too.

PR #18154, opened 2021-02-15
Remove package rebranded to funfields

fccm:
Removed the old deprecated package.
Renamed into "funfields".


kit-ty-kate:
Usually we don't accept straight up deletion without discussion (rather `available: false` in most cases) but given the history and content of this package this is fine. Thanks!

PR #18159, opened 2021-02-15
[new release] opasswd (1.3.1)

psafont:
CHANGES:
  + Port to dune.
  + Fixed quite a few memory safety issues
  + Make the types of to_*_t and of_*_t be more symmetrical
  + Update string fields to char ptrs

This version is a couple years old and has some memory safety fixes, I thought it was about time it got published here.

### DIFF
--- a/packages/opasswd/opasswd.1.3.1/opam
+++ b/packages/opasswd/opasswd.1.3.1/opam
@@ -16,7 +16,6 @@ depends: [
 ]
 build: [
   ["dune" "build" "-p" name "-j" jobs]
-  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
 ]
 dev-repo: "git+https://github.com/xapi-project/ocaml-opasswd.git"
 url {
--- a/packages/opasswd/opasswd.1.3.1/opam
+++ b/packages/opasswd/opasswd.1.3.1/opam
@@ -1,0 +1,26 @@
+opam-version: "2.0"
+synopsis:
+  "OCaml bindings to the glibc passwd file and shadow password file interface"
+maintainer: "xen-api@lists.xen.org"
+authors: "Mike McClurg"
+license: "ISC"
+tags: "org:xapi-project"
+homepage: "https://github.com/xapi-project/ocaml-opasswd"
+bug-reports: "https://github.com/xapi-project/ocaml-opasswd/issues"
+depends: [
+  "ocaml" {>= "4.01.0"}
+  "dune" {>= "1.4"}
+  "ctypes" {>= "0.2.2"}
+  "ctypes-foreign"
+  "odoc" {with-doc}
+]
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+]
+dev-repo: "git+https://github.com/xapi-project/ocaml-opasswd.git"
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
The archive checksum doesn't match the archive:
```
          - opasswd.1.3.1:
            https://github.com/xapi-project/ocaml-opasswd/archive/v1.0.1/ocaml-opasswd-1.0.1.tar.gz (Bad checksum, expected sha256=18a24a9b1467796fa73bc939f14d15807cc6dc8995f56530f6b908b174837235)
```

packages/opasswd/opasswd.1.3.1/opam
```diff
@@ -0,0 +1,26 @@
...
+  "OCaml bindings to the glibc passwd file and shadow password file interface"
+maintainer: "xen-api@lists.xen.org"
+authors: "Mike McClurg"
+license: "ISC"
+tags: "org:xapi-project"
+homepage: "https://github.com/xapi-project/ocaml-opasswd"
+bug-reports: "https://github.com/xapi-project/ocaml-opasswd/issues"
+depends: [
+  "ocaml" {>= "4.01.0"}
+  "dune" {build}
```
kit-ty-kate, line 12:
```suggestion
  "dune" {>= "1.4"}
```

psafont:
The errors in the tests are due to failing to acquire the lock for the shadow file:
```
- Getting password for root
- Lock acquired? false
- chspwd_test couldn't acquire shadow lock
- * finished test_chspwd
- Couldn't acquire lock, must be root
- Fatal error: exception Unix.Unix_error(Unix.EINVAL, "getpwent", "")
```
and on alpine:
```
# Getting password for root
# Lock acquired? true
# Fatal error: exception Unix.Unix_error(Unix.EACCES, "getspnam", "")
```

kit-ty-kate:
> The errors in the tests are due to failing to acquire the lock for the shadow file:
> 
> ```
> - Getting password for root
> - Lock acquired? false
> - chspwd_test couldn't acquire shadow lock
> - * finished test_chspwd
> - Couldn't acquire lock, must be root
> - Fatal error: exception Unix.Unix_error(Unix.EINVAL, "getpwent", "")
> ```
> 
> and on alpine:
> 
> ```
> # Getting password for root
> # Lock acquired? true
> # Fatal error: exception Unix.Unix_error(Unix.EACCES, "getspnam", "")
> ```

Any chance you could remove that part of your default testsuite? Having to run the testsuite in root doesn't seem very practical.

psafont:
The whole point of the library is to read the /etc/shadow, which needs root access. I think it's easier to drop tests altogether, if root is not a given.

I'm a bit conflicted since this library was dropped because of constant issues on the C bindings and was replaced by another with handwritten one.

kit-ty-kate:
> The whole point of the library is to read the /etc/shadow, which needs root access. I think it's easier to drop tests altogether, if root is not a given.

I had a look at the tests and the library in general. I agree the tests should be dropped in the current state of things.

However I think it would be an improvement to ship the tests with a custom made, user accessible, shadow/passwd file and to have an alternative to `Passwd.get_db` and `Shadow.get_db` which makes it able to take an files, not just the system one.

In the meantime I'll remove the tests and will merge this afterwards.

kit-ty-kate:
Thanks!

PR #18163, opened 2021-02-16
added packages/ocamlscript/ocamlscript.3.0.0

UnixJunkie:
doesn't depend anymore on camlp4


kit-ty-kate:

UnixJunkie:
ok, I'll update the opam file accordingly, thanks for the fixes.

PR #18164, opened 2021-02-16
dot-merlin-reader 4.0, merlin 4.1

trefis:
Our opam files should be in order now.
Also, we're including some extra bugfixes.

(I'll also be doing a bugfix release of our 3.4 branch shorty, as a separate PR)


trefis:
Ah, still not quite right, dot-merlin-reader's opam file is incorrect. :facepalm: 
I'll fix that now.

kit-ty-kate:
Looks perfect now. Thanks a lot!! :tada:

trefis:
Thank *you*!

PR #18166, opened 2021-02-16
OCaml 4.12.0, first release candidate

Octachron:
This should be the last prerelease version of OCaml 4.12.0.

The only small change is a small runtime posix compatibility fix for Solaris and OmniOs.


kit-ty-kate:
Thanks a lot! :tada:

PR #18171, opened 2021-02-17
Package bnfgen.3.0.0

dmbaturin:
### `bnfgen.3.0.0`
Random text generator that takes context-free grammars from BNF files
BNFGen generates random texts based on user-defined context-free grammars
specified in a BNF-like syntax. There are descriptive syntax error messages
and tracing options.

You can specify "weight" for rules with alternation to influence their probabilities.
For example, in `<foo> ::= 10 <foo> "foo" | "foo";` the first (recursive) option will be
taken ten times more often.

You can also specify deterministic repetition ranges, like `<foo>{4}` (exactly four of `<foo>`)
or `<foo>{1,5}` (from one to five of `<foo>`).

This package includes both a library and a CLI tool based on it.

---
* Homepage: https://baturin.org/tools/bnfgen
* Source repo: git+https://github.com/dmbaturin/bnfgen
* Bug tracker: https://github.com/dmbaturin/bnfgen/issues

---

### DIFF
--- a/packages/bnfgen/bnfgen.3.0.0/opam
+++ b/packages/bnfgen/bnfgen.3.0.0/opam
@@ -1,0 +1,38 @@
+opam-version: "2.0"
+synopsis:
+  "Random text generator that takes context-free grammars from BNF files"
+description: """\
+BNFGen generates random texts based on user-defined context-free grammars
+specified in a BNF-like syntax. There are descriptive syntax error messages
+and tracing options.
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Thanks! Could you return the fixes in the opam file upstream?

dmbaturin:
Done. Thanks for the suggestions!

PR #17360, opened 2020-10-06
[new release] ppxlib (0.18.0)

NathanReb:
Standard library for ppx rewriters

- Project page: <a href="https://github.com/ocaml-ppx/ppxlib">https://github.com/ocaml-ppx/ppxlib</a>
- Documentation: <a href="https://ocaml-ppx.github.io/ppxlib/">https://ocaml-ppx.github.io/ppxlib/</a>

##### CHANGES:

- Bump ppxlib's AST to 4.11 (ocaml-ppx/ppxlib#180, @NathanReb)

PR #18172, opened 2021-02-17
ppx_protocol_conv: Works with ppxlib > 0.18.0

copy:
The constraint `ppxlib < 0.18` was added in https://github.com/ocaml/opam-repository/pull/17360, however ppx_protocol_conv is already compatible with ppxlib 0.21. It's unclear whether the problem was spurious. I opted to change the constraint to `ppxlib != 0.18`.


kit-ty-kate:
Mmh, indeed I'm not sure why it was added in the first place. I tried locally and it compiled fine so I simply removed the constraint entirerly.

kit-ty-kate:
Thanks!

Issue #18174, opened 2021-02-18
conf-pkg-config.2 fails on windows 


tjammer:
Hi,

I am noticing failures during the installation of conf-pkg-config on windows, see here https://github.com/tjammer/raylib-ocaml/runs/1924692048?check_suite_focus=true . These failures coincide with the update of conf-pkg-config to version 2. 

With version 1.3, depext not only installed `pkg-config` but also `libpkgconf3`, see https://github.com/tjammer/raylib-ocaml/runs/1633951783?check_suite_focus=true . This is not installed with version 2  

edit: Sorry, you probably couldn't see what's inside the logs

<details><summary>conf-pkg-config.2 log</summary>
<p>

```bash
Run opam depext raylib
# Detecting depexts using vars: arch=x86_64, os=win32, os-distribution=cygwinports, os-family=windows
# The following system packages are needed:
pkg-config
system:libX11-devel
system:libXcursor-devel
system:libXi-devel
system:libXinerama-devel
system:libXrandr-devel
Starting cygwin install, version 2.905
User has backup/restore rights
Current Directory: D:\cygwin\packages
Could not open service McShield for query, start and stop. McAfee may not be installed, or we don't have access.
root: D:\cygwin system
Selected local directory: D:\cygwin\packages
net: Preconfig
site: http://cygwin.mirror.constant.com/
solving: 6 tasks, update: no, use test packages: no
Augmented Transaction List:
   0 install mingw64-x86_64-pkg-config 1.6.3-1   
   1 install gsettings-desktop-schemas 3.24.1-1  
   2 install libxml2                   2.9.10-2  
   3 install libXau6                   1.0.9-1   
   4 install libXdmcp6                 1.1.3-1   
   5 install libglib2.0_0              2.54.3-1  
   6 install libxcb1                   1.14-1    
   7 install desktop-file-utils        0.23-1    
   8 install libX11_6                  1.7.0-1   
   9 install gamin                     0.1.10-15 
  10 install libXrender1               0.9.10-1  
  11 install libXext6                  1.3.4-1   
  12 install libXfixes3                5.0.3-1   
  13 install libfam0                   0.1.10-15 
  14 install libXi6                    1.7.10-1  
  15 install libXinerama1              1.1.4-1   
  16 install libXrandr2                1.5.2-1   
  17 install libXcursor1               1.2.0-1   
  18 install shared-mime-info          2.1-1     
  19 install pkg-config                0.29.1-1  
[...]

PR #18177, opened 2021-02-18
conf-pkg-config.2: Fix depexts for windows

kit-ty-kate:
Fixes https://github.com/ocaml/opam-repository/issues/18174

### DIFF
--- a/packages/conf-pkg-config/conf-pkg-config.2/opam
+++ b/packages/conf-pkg-config/conf-pkg-config.2/opam
@@ -32,7 +32,7 @@ depexts: [
   ["pkgconf"] {os = "freebsd"}
   ["pkgconf-pkg-config"] {os-distribution = "rhel" & os-version >= "8"}
   ["pkgconf-pkg-config"] {os-distribution = "centos" & os-version >= "8"}
-  ["pkg-config"] {os-distribution = "cygwinports"}
+  ["system:pkgconf"] {os = "win32" & os-distribution = "cygwinports"}
 ]
 synopsis: "Check if pkg-config is installed and create an opam switch local pkgconfig folder"
 description: """


packages/conf-pkg-config/conf-pkg-config.2/opam
```diff
@@ -32,7 +32,7 @@ depexts: [
   ["pkgconf"] {os = "freebsd"}
   ["pkgconf-pkg-config"] {os-distribution = "rhel" & os-version >= "8"}
   ["pkgconf-pkg-config"] {os-distribution = "centos" & os-version >= "8"}
-  ["pkg-config"] {os-distribution = "cygwinports"}
+  ["system:pkgconf"] {os = "win32" & os-distribution = "cygwinports"}
```
mseri, line 35: So only win64 uses the other repo?

MSoegtropIMC: @mseri : confusingly 64 bit windows als has os=win32. Here is what `opam var` gives on 64 bit cygwin:
```
arch              x86_64                                        # Inferred from system
opam-version      2.0.8                                         # The currently running opam version
os                win32                                         # Inferred from system
os-distribution   cygwinports                                   # Inferred from system
os-family         windows                                       # Inferred from system
```
Note the value of the arch variable. This i686 on 32 bit cygwin.

Also note that this is the MinGW patched opam, not the plain cygwin opam, but I guess this would be the same.

Anyway I would also recommend to remove the `os = "win32"` test since it is confusing and superflous. `os-distribution = "cygwinports"` implies `os = "win32"`.

MSoegtropIMC:
I was about to make the same pull request :-)

@mseri: some background info: on WIndows many people use this opam repo (https://github.com/fdopen/opam-repository-mingw) which always had the package name "system:pkgconf". But the new opam version was merged over without changing this.

I would say long term it makes more sense to fix things here than to patch them in the MinGW special repo, so that long term we can remove the need for this repo.

PR #18179, opened 2021-02-18
merlin: fix ocaml version contraints

MisterDA:
If working in an OCaml 4.11.2 switch (release forthcoming) merlin
would be upgraded from 411 to 412.
This also helps with beta switches, if my calculations are correct we have 4.11 <= 4.11.0~x.
@trefis


trefis:
> if my calculations are correct we have 4.11 <= 4.11.0~x.

Ah !

kit-ty-kate:
Oops, I missed that in my review. Sorry about that. Thanks a lot!

PR #18180, opened 2021-02-18
[new release] git, git-cohttp, git-cohttp-unix, git-cohttp-mirage, git-unix and git-mirage (3.3.0)

dinosaure:
Git format and protocol in pure OCaml

- Project page: <a href="https://github.com/mirage/ocaml-git">https://github.com/mirage/ocaml-git</a>
- Documentation: <a href="https://mirage.github.io/ocaml-git/">https://mirage.github.io/ocaml-git/</a>

##### CHANGES:

- Fix tests when we push to an empty repository (mirage/ocaml-git#462, @dinosaure, @ulugbekna)
- Fix new smart tests without a global git config (mirage/ocaml-git#463, @sternenseemann, @dinosaure)
- Refactor tests (mirage/ocaml-git#464, @ulugbekna, @dinosaure)
- Refactor `find_common.ml` (mirage/ocaml-git#465, @ulugbekna, @dinosaure)
- Some preparation about Git protocol v2 (mirage/ocaml-git#466, @ulugbekna, @dinosaure)
- Add HTTP/HTTPS support on the unikernel example (mirage/ocaml-git#467, @dinosaure)
- Fix bug about push and capabilities (mirage/ocaml-git#468, @dinosaure, @hannesm, @ulugbekna)
- Implement HTTP push and handle username & password in the given `Uri.t` (mirage/ocaml-git#469, @dinosaure)
- **breaking changes**
  Commit object must have a double LF to separate the header and the body. This format changes
  hashes of commit and invalid old commits generated by `ocaml-git`. It seems that on the Git
  side, the format is not fully respected - and it's why Git did not complain about that on our
  tests for a long time.

  However, `git fsck` does this check and an HTTP push to GitHub run `git fsck` at the end. To
  be able to push (or run our _unikernel_ over HTTP/S), we must generate right commits.

  So, we advise users to make a fake commit with Git (`git commit --allow-empty -m.`) on
  repositories used by `ocaml-git` and fetch it with `depth:1`. Otherwise, this new version
  of `ocaml-git` will refute older commits - due to the inherent isomorphism between
  decoder/encoder in `ocaml-git`.

  For Irmin users, this breaking changes does not change anything when `irmin` reformats correctly
  the message to put the second LF. However, a breaking change exists on the API level when the
  Git commit (and a Git tag) expects a `string option` now (instead of a simple `string).

  For more details and tests, see mirage/ocaml-git#470 (@dinosaure)

### DIFF
--- a/packages/git-cohttp-mirage/git-cohttp-mirage.3.3.0/opam
+++ b/packages/git-cohttp-mirage/git-cohttp-mirage.3.3.0/opam
@@ -1,0 +1,43 @@
+opam-version: "2.0"
+synopsis: "A package to use HTTP-based ocaml-git with MirageOS backend"
+maintainer: ["thomas@gazagnaire.org" "romain.calascibetta@gmail.com"]
+authors: "Thomas Gazagnaire"
+license: "ISC"
+homepage: "https://github.com/mirage/ocaml-git"
+doc: "https://mirage.github.io/ocaml-git/"
+bug-reports: "https://github.com/mirage/ocaml-git/issues"
+depends: [
+  "ocaml" {>= "4.08.0"}
+  "dune" {>= "2.6.0"}
+  "git" {= version}
+  "mimic"
+  "cohttp-mirage"
+  "cohttp" {>= "2.5.4"}
+  "cohttp-lwt" {>= "2.5.4"}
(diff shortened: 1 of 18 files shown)


kit-ty-kate:
Thanks!

PR #18181, opened 2021-02-19
pardi.3.2.0: upper bound on dolog is not needed

UnixJunkie:
pardi compiles fine with dolog.6.0.0, for example


kit-ty-kate:
Thanks!

PR #18187, opened 2021-02-19
Release clangml.4.4.0, stdcompat.15, refl.0.4.0 and dependencies

thierry-martinez:
This pull request adds new releases for clangml, stdcompat, refl and
dependencies with support for OCaml 4.12. In addition, clangml
supports new and next releases of Clang/LLVM: 11.0.1, 11.1.0, 12.0.0.

Packages traverse and pattern are unchanged but the OCaml version
constraint is removed since they work as is with OCaml 4.12.


thierry-martinez:
I think that CI error `Error in conf-libclang.12: error 60: Upstream check failed: "No url defined")` is OK for a `conf-*` package.

kit-ty-kate:
> I think that CI error `Error in conf-libclang.12: error 60: Upstream check failed: "No url defined")` is OK for a `conf-*` package.

yes it's been fixed in the `opam` library on master, it'll be fixed in the CI in a week or two

packages/clangml/clangml.4.4.0/opam
```diff
@@ -0,0 +1,41 @@
...
+  "pattern" {with-test & >= "0.2.0"}
+  "metapp" {>= "0.4.0"}
+  "metaquot" {>= "0.4.0"}
+  "refl" {>= "0.4.0"}
+  "odoc" {with-doc & >= "1.5.1"}
+]
+x-ci-accept-failures: [
+   "oraclelinux-7" "oraclelinux-8" # llvm-config not available
+   "centos-7" # GCC does not support --std=c++14
+   "debian-10" # Extra distributions fail
```
kit-ty-kate, line 30: I don't think this should be there. There is a genuine segfault:
```
#     ocamlopt clangml/.clang.objs/native/clang__Clang__bindings.{cmx,o} (got signal SEGV)
```
Those platforms (ppc64 and arm32 either have to be forbidden using the `available` field or this issue should be debugged)

thierry-martinez: You are right: I forgot about the `available` field, but it is definitively better than `x-ci-accept-failures` in this case (and I would prefer not to debug these issues for this release).

thierry-martinez:
Thank you very much, @kit-ty-kate ! I managed to fix the bug revealed by the tests on `x86_32`, so the only remaining failure is on `oraclelinux-7`, which misses `clang-devel` package. I do not know how to exclude this specific version with the `available` field without excluding `oraclelinux-8` as well (and maybe it is even better to let the user install it in the case she has installed `clang` manually). Is there still any blocker before merging?

kit-ty-kate:
Thanks!

PR #18202, opened 2021-02-22
Fmlib Release 0.1.0

hbr:


kit-ty-kate:
I changed your opam file quite a bit, could you check that all the changes I've made (apart from the reformat) are correct? If so, could you return them to your opam file upstream?

hbr:
@kit-ty-kate

I don't see any problems with the changes you have made. I am not an expert in opam files. Could you explain to me the reason for the changes? Having understood them, I am glad to integrate them into my opam file.

Thanks

kit-ty-kate:
There was several issues:
* the `homepage` was pointing to the documentation, so I simply changed it to the `doc` field and added a new `homepage` field pointing to the github repository
* the `bug-report` field was pointing to the repository but usually points directly to the /issues page
* `dune subst {pinned}` is a bug that got slipped in dune until dune 2.7. `{dev}` should be used instead
* The `@install` parameter to `dune build` was present twice so i simply removed one of them
* `ocamdfind`, `js_of_ocaml` and `js_of_ocaml-ppx` wasn't needed anywhere in the code so I simply removed them
* `odoc` was a documentation dependency, not a build dependency
* `ppx_inline_test` has a runtime so it's not a build dependency. The added constraint is here to pick a reasonable lower-bound constraint so it does not fail with users using older versions. janestreet packages can change quite a bit between versions.
* `dune` is not a build dependency (see the bug in dune https://github.com/ocaml/dune/issues/2147). Also your `dune-project` file states that this package requires at least dune 1.8 so i added the constraint.

hbr:
@kit-ty-kate
Thanks for the precise explanation of the issues.

kit-ty-kate:
Thanks

PR #18209, opened 2021-02-23
Add coq-of-ocaml version 2.4.0

clarus:
Add `coq-of-ocaml` version 2.4.0.

### DIFF
--- a/packages/coq-of-ocaml/coq-of-ocaml.2.4.0/opam
+++ b/packages/coq-of-ocaml/coq-of-ocaml.2.4.0/opam
@@ -1,0 +1,42 @@
+opam-version: "2.0"
+maintainer: "dev@clarus.me"
+homepage: "https://github.com/clarus/coq-of-ocaml"
+dev-repo: "git+https://github.com/clarus/coq-of-ocaml.git"
+bug-reports: "https://github.com/clarus/coq-of-ocaml/issues"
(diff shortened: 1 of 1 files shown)


packages/coq-of-ocaml/coq-of-ocaml.2.4.0/opam
```diff
@@ -0,0 +1,42 @@
...
+build: [
+  ["sh" "-c" "cd proofs && ./configure.sh"] {coq:installed}
+  [make "-C" "proofs" "-j%{jobs}%"] {coq:installed}
+  ["dune" "build" "-p" name "-j" jobs]
+]
+install: [
+  [make "-C" "proofs" "install"] {coq:installed}
+]
+depends: [
+  "dune" {>= "2.0" & < "2.8"}
```
kit-ty-kate, line 17: this constraint is surprising. What's the cause of this?

clarus: This is because the version of Merlin which we use internally is not compatible with the new connection system to Dune, and must use `.merlin` files which are not generated anymore by Dune. At some point we will probably update our vendored version of Merlin to support the new Dune.

kit-ty-kate:
Fair enough. Thanks!

PR #18215, opened 2021-02-24
new file:   packages/parany/parany.12.0.2/opam

UnixJunkie:


kit-ty-kate:
The tests failed with:
```
#=== ERROR while compiling parany.12.0.2 ======================================#
# context              2.0.8 | linux/x86_64 | ocaml-base-compiler.4.03.0 | pinned(https://github.com/UnixJunkie/parany/archive/v12.0.2.tar.gz)
# path                 ~/.opam/4.03/.opam-switch/build/parany.12.0.2
# command              ~/.opam/4.03/bin/dune runtest -p parany -j 47
# exit-code            1
# env-file             ~/.opam/log/parany-19-e7966f.env
# output-file          ~/.opam/log/parany-19-e7966f.out
### output ###
# File "src/dune", line 16, characters 20-25:
# 16 |   (run ./test.exe %{jobs} 800 0)))
#                          ^^^^^
# Error: Unknown variable "jobs"
```
In the previous PR i was talking about environment variables (as in unix env), not dune variables. AFAIK there is no builtin way in dune to get the `jobs` opam variable.

kit-ty-kate:
for the purpose of tests the hard-coded number is mostly fine as long as it's not too high.

UnixJunkie:
I told you, we should have left those tests commented out. :)
Now, I have to look into the dune documentation how to run some shell code...

kit-ty-kate:
my fixed as-is worked fine. I'm telling you, it's not worth the hassle, hardcoded numbers are fine.

UnixJunkie:
I'm trying to run the regression test suite via the opam file.
I don't like the fact that the dune file let people think that it is _the_ way to regression test this library.
Dune should just do one thing well (build), I will take care of doing the other things well. :)

kit-ty-kate:

UnixJunkie:
Ok, thanks!

PR #18218, opened 2021-02-24
[new release] mssql (2.2.0)

brendanlong:

- Documentation: <a href="https://arenadotio.github.io/ocaml-mssql">https://arenadotio.github.io/ocaml-mssql</a>

### Fixed

- `Row.datetime` and `Row.datetime_exn` now work correctly for the `DATETIME2` type
- Queries should be slightly less likely to return errors due to previous queries not being cleaned up properly.


avsm:
Looks good, thanks! There are missing depexts for OL7/8 and Arch, but that's not a blocker. You may want to announce this on <https://discuss.ocaml.org>, where we have a Community category and an `announce` tag for this purpose.

PR #18220, opened 2021-02-24
Cygwin builds in parallel since 4.09.0

dra27:
#14257 enabled parallel builds for the compiler. It was necessary to exclude OCaml 4.07 and 4.08 on Cygwin, but in the discussion and rebases/re-runs it somehow got extended to all of the versions and has continued ever since!

Parallel builds on Cygwin should work in the same way as every other distro since 4.09.0 (https://github.com/ocaml/ocaml/pull/2267).


kit-ty-kate:
Thanks!

PR #18223, opened 2021-02-25
new file:   packages/parany/parany.12.0.3/opam

UnixJunkie:
corrected the duplicated I/O problem.
Not a parany bug per se, just something that might annoy/surprise some users.


kit-ty-kate:
Thanks!

PR #18225, opened 2021-02-25
Fix OCaml development branch ("+trunk") packages

dra27:
All (!!) the `+trunk` packages for 4.02 onwards are broken, except for 4.13 (i.e. the actual trunk!).

OCaml always bumps the `VERSION` file after every tag so, for example, as soon as 4.12.0 is released, the +trunk package for 4.12 needs to become `ocaml-variants.4.12.1+trunk` since `ocaml.4.12.0` will not install with a compiler returning `4.12.1+dev0-2021-02-24`.

These packages have been haphazardly renamed in the past - there are (at least) two other possible schemes we should be considering for this, but the aim of this PR is to make the present behaviour correct.

Therefore:
- `ocaml` gains the required versions (that is, for every released version x.y.z of OCaml, there is definitely `ocaml.x.y.(z+1)`)
- The existing `+trunk` packages for 4.02+ are all renamed to the correct version. **Because that version is necessarily later than `ocaml-base-compiler` in the same release, I have tagged them all with `ocaml-beta`/`avoid-version` flags**. This should be a safe-change given that we transition from packages which were broken to packages which are unavailable.
- The 4.10 trunk packages disappeared at some point - I think by accident.
- There were never 4.03 and 4.04 packages, so I added them! The gcc10 patches are exact copies of ones from other 4.03 and 4.04 compiler packages.
- Since 4.12, configure now supports `--disable-warn-error`. Builds of OCaml with `+dev` in the version number compile with C warnings turned into errors. This is obviously a problem for opam-repository where an old compiler may necessarily be installed with a new C compiler. The 4.12 and 4.13 +trunk packages already use this new flag - for the old packages, the additional of `-Werror` is sed-d out of the `configure` script. In particular, 4.03 and 4.04 don't build without it on at least gcc 9.


avsm:
Looks right to me. Will not merge when tired at the end of the day though :)

dra27:
One of the commits accidentally touched the misnamed `packages/ocaml-variants/ocaml-variants.4.04.0+trunk+forced_lto/opam`, which I've now removed!

PR #18226, opened 2021-02-25
Various fixes (OCaml 4.13 edition)

kit-ty-kate:
cc @jzstark for `owl-symbolic`
cc @KeenWill for `camelot` (oops i forgot to specify why in the commit message. It's the type of the pattern matching parsetree that changed in compiler-libs)

The rest will be fixed in due time


jzstark:
Thanks for the fixing Kate! Looks good to me.

KeenWill:
Thank you Kate, I'll push these changes upstream to my repo too!

kit-ty-kate:
Oh sorry I should specify. When I cc the maintainers here, it's usually less to upstream the changes and more to encourage people to fix their packages early so the next release can be as smooth as possible

PR #18230, opened 2021-02-26
LablTk 8.06.10 and OCamlBrowser 4.12.0

garrigue:
This is a bug fix release of LablTk and OCamlBrowser.
It also makes OCamlBrowser work on OCaml 4.12 (and only this version).


garrigue:
I do not understand the error message for ocaml 4.12.
Locally, I could install both packages using those opam files and ocaml 4.12.0.

kit-ty-kate:
The errors are fixed now. Thanks a lot!

PR #18233, opened 2021-02-26
upgrade janestreet packages

cwong-ocaml:
upgrades to base_quickcheck and ppx_typerep_conv


cwong-ocaml:
Thanks for catching my errors! Looks like some stuff fell through with my script; I'll make sure to look into that.

kit-ty-kate:
Thanks!

PR #18235, opened 2021-02-27
pa_ppx requires exact versions of the compiler

kit-ty-kate:
@chetmurthy pa_ppx broke with the release of ocaml 4.11.2.
Could you return this fix to the opam file upstream? It's not the first time this type of error appears in your software stack.


chetmurthy:
arrgh, sorry, I'll get on this immediately

kit-ty-kate:
Thanks!

chetmurthy:
Not that it's of any help, but I've fixed the underlying problem in the latest code, but just haven't released it.  I'll make that release now, after checking that (ahem) actually does deal with this problem.

chetmurthy:
BTW, in case you're still working [and I hope not, on a Friday] the new releases 4.12.0 and 4.11.2 don't seem to be available in opam.

PR #18222, opened 2021-02-25
Package hack_parallel.1.0.1

rvantonder:
### `hack_parallel.1.0.1`
Parallel and shared memory library
Parallel and shared memory components used in Facebook's Hack, Flow, and Pyre
projects.

---
* Homepage: https://github.com/rvantonder/hack_parallel
* Source repo: git+https://github.com/rvantonder/hack_parallel.git
* Bug tracker: https://github.com/rvantonder/hack_parallel/issues

---
:camel: Pull-request generated by opam-publish v2.0.3

PR #18237, opened 2021-02-27
relax ppxlib constraint on hack_parallel.1.0.1

rvantonder:
I think this constraint can be relaxed but missed removing it in https://github.com/ocaml/opam-repository/pull/18222. Let's see.


kit-ty-kate:
Thanks!

rvantonder:
Ah, thank you @kit-ty-kate for the `available: arch != "arm32"` addition!

PR #18238, opened 2021-02-28
Remove GT.0.2 -- it is not compilable on any version of CI

Kakadu:
And I don't want to repair it :)

Signed-off-by: Kakadu <Kakadu@pm.me>


avsm:
Please bear in mind that packages on opam should be as immutable as possible, so we generally do not permit deletion but prefer `available = false` so that downstream users can enable it if they take a dependency on the package.

However, in this case, it appears we should never have merged this version of the package in the first place as it doesn't have a stable `url` (just a git remote pointing to the master branch of the repository). Given that it's unclear what the intended original sources were, I'll take this deletion.

PR #18243, opened 2021-03-01
Package camelot.1.7.0

KeenWill:


kit-ty-kate:
Thanks! Could you return the change of license name to the opam file upstream?

KeenWill:
Sure! Thanks Kate!

PR #18240, opened 2021-02-28
Revive the pre 4.02.0 compilers

dra27:
The old OCaml compilers are of interest to ~a few people~at least me and @herbelin (see #5260)! I recall noticing several years ago that 3.07 and 3.08 had stopped building on anything later than Ubuntu 14.04 (IIRC!), but my lazy solution on the very rare occasions that I've "needed" them is to fire up that old VM. It's been gradually spreading, though as gcc and, more relevantly, the linker have evolved. I can't visit actual care homes at the moment, so I spent a little time with some old compilers instead.

As always, the patches carried here are strictly ones required to build an old OCaml with a new C compiler. For 3.07-3.09, this necessitates back-porting the 3.09 addition of `-fPIC` to ocamlopt to 3.07 and 3.08 and the 3.10.0 adjustment to use it by default to all three of those.

- It turns out OCaml 3 is not affected by the `-fcommon` fix (I wasn't interested enough to check why!) so the patches in #16583 are extended to 4.00.0, 4.00.1 and 4.01.0
- `ocaml-base-compiler.3.11.2` contained an external reference to a patch on the OCamlPRO website which is no longer there. This patch is required for any modern binutils and so I've replaced it with the actual commit from the OCaml repo and extended it to 3.11.0 and 3.11.1
- `ocaml-config.1` has never worked on 3.07... I've simply inserted `ocaml-config.0` to handle the change in version number scheme 😂 Note that the change to `ocaml-config.1` will trigger a switch rebuild in opam 2.1 but because `ocaml-config` is a base package, it won't trigger that in opam 2.0. On that basis, I've amended its constraints, rather than a slightly more weird adjustment to `ocaml-base-compiler.3.07*/opam`.

If we're going to have these packages (and, even as we end up splitting opam-repository, they should still be kept) then we should periodically check that they still build. To that end, I've adjusted 3 packages:
- `ocamlfind.1.0.4` is the last version which supported 3.07, so adding it allows `ocamlfind` to be installed in _any_ ocaml switch
- `pcre.6.2.5` allows a library with C stubs to be installed in 3.08+
- `calendar.1.10` - at least at present - allows a single package at a single version to be installed on _all_ supported versions of OCaml

As usual, while touching `ocaml-base-compiler` packages and others, I've updated metadata.

PR #18245, opened 2021-03-01
Restore ARM64 support for 4.02-4.05

dra27:
Further to https://github.com/ocaml/opam-repository/pull/18240#issuecomment-788052388

I haven't constrained the patches at all, as they only apply to files in the ARM64 backend.

### DIFF
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.4.02.0/files/gpr1330.patch
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.4.02.0/files/gpr1330.patch
@@ -1,0 +1,58 @@
+From 475c2dc0c7ae197772bfcdf0dd5a17ffc5d4fb3b Mon Sep 17 00:00:00 2001
+From: Mark Shinwell <mshinwell@gmail.com>
+Date: Wed, 13 Sep 2017 10:23:16 +0100
+Subject: [PATCH] AArch64 GOT fixed
+
+---
+ asmcomp/arm64/emit.mlp     | 12 ++++++++++--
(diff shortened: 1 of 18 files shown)


mseri:
The ppc failures are addressed in https://github.com/ocaml/opam-repository/pull/18240 if I am not mistaken. This seems good to merge

avsm:
A provenance question before I merge this @dra27 -- where did those arm64 patches originate from? (another Git repo, or through direct correspondence, or something else?)

dra27:
It's from https://github.com/ocaml/ocaml/pull/1330/commits/4274dfe22ae37c4c5589a6fe7fe9d53d62013df5 (OCaml 4.06.0) with hunks relating to `Changes` and three assertions deleted (to make the rebase easier).

I think (apropos this morning's meeting) that there are two kinds of patches to keep the old compilers going:
1. The tools OCaml calls change, necessitating flags or build system tweaks (e.g. GCC 10 fcommon, macOS function detection in `configure` in 4.07 and earlier, etc.)
2. The code OCaml generates becomes wrong, necessitating back-porting of patches (e.g. this one, and the PIC-related ones for the previous PR for OCaml 3.x)

My feeling is that the first kind can safely go in OCaml maintenance branches, since they change how OCaml is built, not what it is. I don't think applies to the second kind here, because OCaml itself is altered (if only so that it actually builds!).

avsm:
Thanks; I'm in total agreement. It's just the lack of provenance to the patches to their origins that is sometimes a problem, but it could easily be tracked back to this PR from the merge point if we need to in the future.

PR #18247, opened 2021-03-01
Package liquidsoap.1.4.4

toots:
Hi!

We had a bad regression with the `1.4.4` release and had to re-release it. This PR updates the tarball checksums.

Sorry about that..


kit-ty-kate:
I'll merge since it's broken already due to your force-push, but next time please consider releasing 1.4.5 instead, it's much simpler for everyone.

toots:
Thanks. I'll do it this way next time.

PR #18261, opened 2021-03-03
camlp4.system: remove brew depext

avsm:
Homebrew removed it so this causes issues in opam 2.1 with depexts
on by default; https://github.com/Homebrew/homebrew-core/issues/49661


kit-ty-kate:
> this causes issues in opam 2.1 with depexts on by default

The thing is that removing it will not really fix the issue either. See https://github.com/ocaml/opam/pull/4548
If you want users to not install these packages on macos with ocaml-system, `available: os != "macos"` should be given instead.

However this is also an issue for people using this package with manually installed `camlp4` so it's not that clear that it should be the solution

avsm:
> If you want users to not install these packages on macos with ocaml-system, available: os != "macos" should be given instead.

This PR is still correct as-is. We shouldn't specify a depext for Homebrew when the package doesn't exist.  Available != macos will not deal with the situation where the user has put a `camlp4` binary manually onto their system in the PATH...

PR #18284, opened 2021-03-07
[new release] omd (2.0.0+alpha1)

nojb:
A Markdown frontend in pure OCaml

- Project page: <a href="https://github.com/ocaml/omd">https://github.com/ocaml/omd</a>

##### CHANGES:

- Lower OCaml requirement to 4.04.2  (ocaml/omd#213, @jfrolich)

- Big refactoring by @nojb. Changes in interface (simplification of code, might
  affect performance a little). To be tested!

### DIFF
--- a/packages/omd/omd.2.0.0+alpha1/opam
+++ b/packages/omd/omd.2.0.0+alpha1/opam
@@ -1,0 +1,45 @@
+opam-version: "2.0"
+synopsis: "A Markdown frontend in pure OCaml"
+description: """
+This Markdown library is implemented using only pure OCaml (including
+I/O operations provided by the standard OCaml compiler distribution).
+OMD is meant to be as faithful as possible to the original Markdown.
+Additionally, OMD implements a few Github markdown features, an
+extension mechanism, and some other features. Note that the opam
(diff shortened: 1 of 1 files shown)


nojb:
What's the standard way to deal with packages which are incompatible with the new version?

aantron:
@nojb should it be `+alpha` or `~alpha`? Looking [here](https://opam.ocaml.org/doc/Manual.html#Package-Formulas) at Version Ordering, it seems that `2.0.0~` would come before `2.0.0`. I don't see any special meaning for `+` here, so would it be considered greater than `2.0.0`? Or is this intentional?

nojb:
> @nojb should it be `+alpha` or `~alpha`? Looking [here](https://opam.ocaml.org/doc/Manual.html#Package-Formulas) at Version Ordering, it seems that `2.0.0~` would come before `2.0.0`. I don't see any special meaning for `+` here, so would it be considered greater than `2.0.0`? Or is this intentional?

Hi, thanks, you are right, it should be `~alpha`... something got mixed up along the way, I'll look into it.

PR #18300, opened 2021-03-08
Add the libtorch 1.8 packages.

LaurentMazare:
Update the libtorch linux and macos packages to use the newly released PyTorch 1.8 binaries.


mseri:
Just a question before merging, @kit-ty-kate do you know if we need to add the license to conf- packages that download and install sources from external tarballs instead of using the package managers?

mseri:
The licence was also not in the previous packages. I am going to merge for now, we can add the information for all packages at the same time if needed

LaurentMazare:
Thanks!

PR #18304, opened 2021-03-09
Package binaryen.0.8.1

grainbot:
### `binaryen.0.8.1`
OCaml bindings for Binaryen

---
* Homepage: https://github.com/grain-lang/binaryen.ml
* Source repo: git+https://github.com/grain-lang/binaryen.ml.git
* Bug tracker: https://github.com/grain-lang/binaryen.ml/issues

---

### Bug Fixes

* Module.optimize on macOS ([#77](https://www.github.com/grain-lang/binaryen.ml/issues/77)) ([0b92043](https://www.github.com/grain-lang/binaryen.ml/commit/0b920431372119d92b0077f51b91ca745fbfde87))

### Continuous Integration

* write opam CHANGES from environment var ([#78](https://www.github.com/grain-lang/binaryen.ml/issues/78)) ([4f25e0e](https://www.github.com/grain-lang/binaryen.ml/commit/4f25e0e96c3b1d2ad22d3b037695960cd6b2722a))

---

### DIFF
--- a/packages/binaryen/binaryen.0.8.1/opam
+++ b/packages/binaryen/binaryen.0.8.1/opam
@@ -1,0 +1,29 @@
+opam-version: "2.0"
+synopsis: "OCaml bindings for Binaryen"
+maintainer: "oscar@grain-lang.org"
+authors: "Oscar Spencer"
+license: " Apache-2.0"
+homepage: "https://github.com/grain-lang/binaryen.ml"
+bug-reports: "https://github.com/grain-lang/binaryen.ml/issues"
+depends: [
+  "ocaml" {>= "4.09"}
+  "dune" {>= "2.7.1"}
+  "dune-configurator" {>= "2.7.1"}
+  "js_of_ocaml" {>= "3.6.0"}
+  "js_of_ocaml-ppx" {>= "3.6.0"}
+  "js_of_ocaml-compiler" {>= "3.6.0"}
+]
+available: arch = "x86_64" & (os = "linux" | os = "macos" | os = "win32")
+build: [
+  ["dune" "subst"] {dev}
+  ["dune" "build" "-p" name "-j" jobs]
+]
+dev-repo: "git+https://github.com/grain-lang/binaryen.ml.git"
+url {
+  src:
+    "https://github.com/grain-lang/binaryen.ml/releases/download/v0.8.1/binaryen-archive-v0.8.1.tar.gz"
+  checksum: [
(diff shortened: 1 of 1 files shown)


phated:
Good bot! 🎊

phated:
Any idea why it ran on ppc64 if our arch says we only run on x86_64 linux, windows, mac?

packages/binaryen/binaryen.0.8.1/opam
```diff
@@ -0,0 +1,29 @@
+opam-version: "2.0"
+synopsis: "OCaml bindings for Binaryen"
+maintainer: "oscar@grain-lang.org"
+authors: "Oscar Spencer"
+license: " Apache-2.0"
+homepage: "https://github.com/grain-lang/binaryen.ml"
+bug-reports: "https://github.com/grain-lang/binaryen.ml/issues"
+depends: [
+  "ocaml" {>= "4.09"}
+  "dune" {>= "2.7.1"}
```
mseri, line 10: The dune-project file requires dune 2.6. I am fine to merge as is, but if you prefer I can lower this to 2.6 before merging

mseri:
> Any idea why it ran on ppc64 if our arch says we only run on x86_64 linux, windows, mac?

The error is unrelated. It fails to create the VM, before realizing that it does not need to build the project

packages/binaryen/binaryen.0.8.1/opam
phated: @mseri I was under the impression that the `dune` version and `dune lang` version were separate and that we could upgrade dune without upgrading the language in our dune files?

mseri: Yes, you can. What I meant is that if you are using dune lang version 2.6, the package is automatically compatible with dune 2.6, even if you use newer versions of dune

phated: I see. We actually had dune pinned at 2.6.1 for a long time so this project would build with esy and then we were finally able to upgrade to >=2.7.1. Maybe it was 2.7 that didn't work (?) so we didn't want to change to >=2.6 and instead upgraded to >=2.7.1. This change was also applied to the opam file in order to keep those dependencies in sync.

mseri:
```suggestion
  "dune" {>= "2.6"}
```
I leave it up to you. I will merge tomorrow, if you decide you want this change feel free to accept the suggestion, otherwise I merge as is

PR #18315, opened 2021-03-10
camllib.1.3.3

nberth:


kit-ty-kate:
Thanks! If the opam file is also stored somewhere in your source repository, could you return the fix there so that the issue won't be there next time?

PR #18319, opened 2021-03-12
new file:   packages/conf-sed/conf-sed.1.0/opam [do not merge; must be changed to conf-gsed]

UnixJunkie:


packages/conf-sed/conf-sed.1.0/opam
```diff
@@ -0,0 +1,23 @@
+opam-version: "2.0"
+maintainer: "unixjunkie@sdf.org"
+homepage: "https://www.gnu.org/software/sed/"
+authors: "https://www.gnu.org/software/sed/#maintainers"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: "GPLv3"
```
mseri, line 6:
```suggestion
license: "GPL-3.0-or-later"
```

kit-ty-kate:
as far as i know sed is a posix standard tool and is installed as part of all the distributions listed in the package.
if you specifically need GNU sed or something like that, could you reflect the presence test accordingly? Currently it only check if any sed binary is present.

UnixJunkie:
If it is always installed, we are fine indeed and we don't need this conf package.
However, I prefer to test with a conf package that sed is here rather than see something fail at runtime.
On Ubuntu, the sed package needs to be installed.
I am not interested in GNU sed especially.
Because on BSDs you would need to use gsed instead of sed then.

packages/conf-sed/conf-sed.1.0/opam
```diff
@@ -0,0 +1,23 @@
...
+depexts: [
+  ["sed"] {os-family = "debian"}
+  ["sed"] {os-distribution = "centos"}
+  ["sed"] {os-distribution = "fedora"}
+  ["sed"] {os-distribution = "ol"}
+  ["sed"] {os = "macos" & os-distribution = "homebrew"}
+  ["sed"] {os-distribution = "alpine"}
+  ["sed"] {os-family = "suse"}
+  ["sed"] {os-family = "arch"}
+  ["sed"] {os = "freebsd"}
```
hannesm, line 18: there's no such package on FreeBSD. The BSD `sed` is shipped with the base system, and there is a package named `gsed` which installs a `GNU sed` (the installed binary is `gsed`) -- I'd recommend to remove the above line from this conf package.

UnixJunkie: fixed

UnixJunkie:
All the requested changes were done.

UnixJunkie:
Still not merged?
Should I do something more about it?

kit-ty-kate:
>  On Ubuntu, the sed package needs to be installed.

I'm curious of that, where did you see that? I just tried, the `sed` package is installed by default and trying to remove it results in a rather angry looking warning and gatekeeper:
```
root@da08d32bdb60:/# apt remove sed
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following packages will be REMOVED:
  sed
WARNING: The following essential packages will be removed.
This should NOT be done unless you know exactly what you are doing!
  sed
0 upgraded, 0 newly installed, 1 to remove and 0 not upgraded.
After this operation, 344 kB disk space will be freed.
You are about to do something potentially harmful.
To continue type in the phrase 'Yes, do as I say!'
```
If you really need it we can merge it but my concern is that if we accept this one, why not have a conf package for every POSIX tools like `conf-echo`, `conf-cat`, `conf-ls`, `conf-grep`. Maybe what you want is for a `conf-posix-1` (containing e.g. grep) and `conf-posix-2` (containing e.g. sed) instead, that doesn't seem too bad to have instead.

avsm:
This package isn't precise enough to be a `conf` package indeed. The homepage entry points to GNU sed, and so the rest of the package should test for GNU sed (e.g. `gsed` on FreeBSD and OpenBSD). There's no point testing for base `sed` as that's part of the POSIX base, but GNU sed is useful to test for due to extensions such as `sed -i` which are commonly used.

UnixJunkie:
Yes, I will change this to conf-gsed and update the test.
Indeed I want to be sure gsed is here rather than just sed.
I tried to use the BSD sed, but it's just too incompatible with the GNU one;
I don't want to learn it.

kit-ty-kate:
Closing while waiting for a new PR for conf-gsed, there is little point in leaving this one open.

PR #18318, opened 2021-03-11
Release Caqti 1.4.0.

paurkedal:
This release is first of all motivated by a bug in the sqlite3 driver (paurkedal/ocaml-caqti#63). The full change log:

- Fix infinite loop when deserializing an optional tuple (GPR#63 mefyl).
- Add `Caqti_connect_sig.S.with_connection` (GPR#61 Anton Bachin).
- Pass parameter types to PostgreSQL prepare and query functions. This
  avoids the need to CAST parameters on the SQL side in some cases.
- Add `?post_connect` callback to `connect_pool`.
- Documentation fixes and improvements (Aaron L. Zeng, Anton Bachin, Petter
  A. Urkedal).

PR #18320, opened 2021-03-12
Fix package constraints after release Caqti 1.4.0.

paurkedal:
I forgot to update the constraints for subpackages which were not updated in #18318.  And thanks for fixing the mariadb dependencies.


mseri:
Thanks

PR #18333, opened 2021-03-14
Sundials/ML: fix build under OCaml 4.12.0

tbrk:
Minor patches to fix build under OCaml 4.12.0. Still no major upgrade to the supported version of Sundials.

### DIFF
--- a/packages/conf-sundials/conf-sundials.1/files/test.c
+++ b/packages/conf-sundials/conf-sundials.1/files/test.c
@@ -1,0 +1,1 @@
+#include <sundials/sundials_config.h>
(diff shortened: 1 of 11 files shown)


tbrk:
The failures occur on platforms where the underlying sundials package is at version > 3.x. We haven't yet released our library for these later versions.

tbrk:
Thank you @kit-ty-kate for your patches!

mseri:
I don't think we can really do much better. People using the older version will likely have their custom compiled library. I think this can be merged

tbrk:
Hi @mseri, I spent a few weeks on Sundials/ML last summer but only managed to update it for Sundials 4.x. Rather than perfect this branch, I tried to go all the way to Sundials 5.x, but in the end I ran out of time. I prefer not to release the 4.x branch into opam because it is not as solid as the 3.x one. The 5.x branch is not in a bad state, it just needs another two weeks or so, but I've been too busy on other things. I'm going to have another shot this summer!

mseri:
No worries. Thanks for the hard work on it and these patches. The comment was about the fact that it is nearly impossible to have perfect depexts for certain packages, so I think this is good to merge

PR #18340, opened 2021-03-15
[new release] carton, carton-lwt and carton-git (0.4.0)

dinosaure:
Implementation of PACKv2 file in OCaml

- Project page: <a href="https://github.com/mirage/ocaml-git">https://github.com/mirage/ocaml-git</a>
- Documentation: <a href="https://mirage.github.io/ocaml-git/">https://mirage.github.io/ocaml-git/</a>

##### CHANGES:

- Handle `trunc` argument when we process a `thin` PACK file
  **breaking changes**
  An optional argument is added on the record which abstract the file-system.
  It should be correctly handled by underlying implementation of the
  file-system. It appears that, at top, we need to figure out such option,
  specially for Git and `Cstruct_append` to correctly access to memories.

### DIFF
--- a/packages/carton-git/carton-git.0.4.0/opam
+++ b/packages/carton-git/carton-git.0.4.0/opam
@@ -1,0 +1,49 @@
+opam-version: "2.0"
+synopsis: "Implementation of PACK file in OCaml"
+description: """\
+Carton is an implementation of the PACK file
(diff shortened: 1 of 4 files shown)


kit-ty-kate:
Thanks!

PR #18342, opened 2021-03-15
Add coq-of-ocaml version 2.4.1

clarus:
Add the version 2.4.1 of `coq-of-ocaml`, with compatibility with Dune 2.8.


kit-ty-kate:
Thanks!

PR #18345, opened 2021-03-15
ocamlfind.1.9.0: Fix BSD support

kit-ty-kate:


dra27:
Slightly simpler fix: add `["sed" "-ib" "-e" "s/</?/" "tools/extract_args/Makefile"]`

dra27:
(my bad - I neither realised that `$<` in that meaning is a GNUism, nor remembered that while the build requires GNU make, configuration doesn't!)

kit-ty-kate:
Tested successfully on FreeBSD and macOS. Merging.

hannesm:
Thanks for your quick patch and merge!

PR #18346, opened 2021-03-16
All cohttp-*.3.0.0 packages need to be unavailable

mseri:
otherwise pinning is messing up the versions.


samoht:
Thanks!

PR #18348, opened 2021-03-16
Move ocamlfind.1.9.0 changes to ocamlfind.1.9.0-1

dra27:
This PR reverts several changes to ocamlfind.1.9.0 in #18345:
- ocamlfind.1.9.0 is marked as unavailable on the BSDs (any other platform which has `make` not being GNU make needs adding)
- The tarball for ocamlfind.1.9.0 is reverted to the main upstream one - the missing script from it is retrieved directly from GitHub (the GitHub tarball and the release tarball differ in _many_ other ways)
- ocamlfind.1.9.0-1 is added restoring support for the BSDs via a simple `sed` call (cf. a considerable number of the ocaml packages)

Since this only tidies up the state of the package, it could be delayed for merging after ocamlfind.1.9.1 is released to opam.


packages/ocamlfind/ocamlfind.1.9.0/opam
```diff
@@ -37,14 +36,19 @@ install: [
...
-  ["fix-bsd.patch" "md5=00e4db61fceee51af794c3a36e4cef0e"]
-]
 dev-repo: "git+https://gitlab.camlcity.org/gerd/lib-findlib.git"
+extra-source "ocaml-stub" {
+  src: "https://raw.githubusercontent.com/ocaml/ocamlfind/findlib-1.9/ocaml-stub"
+  checksum: [
+    "md5=181f259c9e0bad9ef523e7d4abfdf87a"
+    "sha512=d838b025311fe3f3dc887884a741749eebd2878a7b04562a216c8a6a7b435a35bd178cd8b4dab8abf117e4709a722368ec2960198423da6018a8bdb30f590b7a"
+  ]
+}
```
kit-ty-kate, line 46: The `ocamlfind.install` file is also to be downloaded.

dra27: Good point!

dra27: That of course didn't fail on my FreeBSD 11 VM!

hannesm:
There are some reasons I left being a maintainer of the opam-repository. And this PR resonates with me, since it attempts to fix (modify content) a package ("revert") for no obvious good reason. After this PR is merged, there are 3 different versions of ocamlfind 1.9.0 out in the wild.

I (think I) understand what the reason for this PR is, and it looks like this closes the stable door after the horse has bolted. My suggestion for improving the current repository:
- mark ocamlfind 1.9.0 as unavailable (`available: false`) -- once the patches are settled (AFAICT the current ocamlfind 1.9.0 works fine everywhere)
- provide a new ocamlfind.1.9.0-1 with all the desired patches (and using the upstream tarball)

Both approaches (this PR and the above suggestion) lead to reinstallation / upgrade of ocamlfind on any client machine.

kit-ty-kate:
ocamlfind.1.9.1 just got released (https://github.com/ocaml/opam-repository/pull/18349), I think it might be reasonable to mark 1.9.0 as `available: false` and switch directly to 1.9.1 without adding more patches

mseri:
At this stage, why not just release the new 1.9.1 and leave the current 1.9.0 as is instead?

kit-ty-kate:
> At this stage, why not just release the new 1.9.1 and leave the current 1.9.0 as is instead?

that would be fine by me too

hannesm:
Oh, nevermind then. Also @dra27 commented:

> Since this only tidies up the state of the package, it could be delayed for merging after ocamlfind.1.9.1 is released to opam

Thus please disregard my comment above.

avsm:
Closing in favour of #18349

PR #18351, opened 2021-03-16
BLS12-381 packages depend on dune >= 2.8.4 since 0.4.1

dannywillems:
Related to issues when installing implementation of virtual packages in a sandoxed opam environment, see https://github.com/ocaml/dune/pull/4233

It is fixed upstream for future versions


kit-ty-kate:
Thanks!

PR #18356, opened 2021-03-17
Liquidsoap prometheus optional dependency.

toots:
This PR ads a missing optional dependency on prometheus for liquidsoap. Since the dependency actually requires two packages, we introduce a virtual package that installs both of them.

### DIFF
--- a/packages/liquidsoap/liquidsoap.1.4.0/opam
+++ b/packages/liquidsoap/liquidsoap.1.4.0/opam
@@ -76,6 +76,7 @@ depopts: [
   "opus"
   "osx-secure-transport"
   "portaudio"
+  "prometheus-liquidsoap"
   "pulseaudio"
   "samplerate"
   "shine"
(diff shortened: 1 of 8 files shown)


mseri:
Why not just add both dependencies directly?

kit-ty-kate:
sadly the virtual package used in this way does introduce an issue. For instance:
```
opam install liquidsoap prometheus-app cohttp-lwt-unix
```
could make the compilation of liquidsoap fail as the two other package would be installed in parallel and a partial installation could be reached during the compilation of liquidsoap.
The only way to fix this would be to add `prometheus-app` and `cohttp-lwt-unix` separately to the list of depopts and this would make the meta package a bit less relevant.
This issue already exists currently if i understand it correctly so it wouldn't break things per-say.

toots:
Hi @kit-ty-kate and @mseri !

The liquidsoap configure script knows to test for both packages. It actually tests first for `cohttp-lwt-unix` and then, only if it is detected, tests for `prometheus-app`. This way, I believe that race conditions cannot happen as the optional prometheus modules in liquidsoap will only be compiled if both packages are effectively present at time of configuration.

avsm:
Thanks for the clarification @toots -- your analysis looks sound to me, so merging.

PR #18359, opened 2021-03-18
bls12-381-unix can use 0.18.0

dannywillems:
Some warnings appear with ctypes 0.17.1 (see https://github.com/ocamllabs/ocaml-ctypes/pull/646 and https://github.com/ocamllabs/ocaml-ctypes/pull/638). There is a fix upstream in master for future releases (https://gitlab.com/dannywillems/ocaml-bls12-381/-/commit/50c2416b4e1ea12e7ab01c2529cf5d8d8a424998). 0.17.1 is still allowed in this version in case of already some users require 0.17.1. However, tezos uses ctypes 0.18.0 and require this version.

### DIFF
--- a/packages/bls12-381-unix/bls12-381-unix.0.4.1/opam
+++ b/packages/bls12-381-unix/bls12-381-unix.0.4.1/opam
@@ -17,7 +17,7 @@ depends: [
   "dune-configurator" {build}
   "ff-sig" {>= "0.6.1" & < "0.7.0"}
   "zarith" {>= "1.10" & < "2.0"}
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
The commit/PR name doesn't seem to to the same thing as its implementation. Is that expected?

dannywillems:
> The commit/PR name doesn't seem to to the same thing as its implementation. Is that expected?

Yes, I motivate it in my first message.

kit-ty-kate:
oh sorry nevermind I was seeing the change inverted. Thanks!

PR #18363, opened 2021-03-18
Enable WebAssembly for LLVM 8, 9, 10, and 11

alan-j-hu:
Hi all,

I want to use LLVM to emit WebAssembly, but when I install the OCaml package, I find that the WebAssembly backend module isn't installed. [LLVM moved WebAssembly from experimental to being a default backend in version 8](https://releases.llvm.org/8.0.0/docs/ReleaseNotes.html#changes-to-the-webassembly-target), so it shouldn't be filtered out in versions 8+. Could you please enable WebAssembly? Thanks.


kit-ty-kate:
Thanks!

PR #18366, opened 2021-03-19
[new release] doi2bib (0.3.0)

mseri:

##### Release 0.3.0


mseri:
The failure on ARM32 is unrelated. For some reason it reliably fails to resolve `www.ncbi.nlm.nih.gov`, no idea why.
In principle, I could weaken tls requirements and require `("tls" | "ssl")` but I don't know if it makes any difference here.

PR #18376, opened 2021-03-22
Various fixes

kit-ty-kate:
cc @tmattio for `spin`


tmattio:
Noted, thanks for the ping @kit-ty-kate 🙂

PR #18382, opened 2021-03-23
Various fixes

kit-ty-kate:
All old versions. Maintainers notification not needed.

### DIFF
--- a/packages/angstrom/angstrom.0.1.0/opam
+++ b/packages/angstrom/angstrom.0.1.0/opam
@@ -10,26 +10,22 @@ build: [
     "ocaml"
     "setup.ml"
     "-configure"
-    "--prefix"
-    prefix
+    "--prefix" prefix
     "--%{base-unix:enable}%-unix"
     "--%{lwt:enable}%-lwt"
     "--%{async:enable}%-async"
+    "--enable-tests" {with-test}
   ]
   ["ocaml" "setup.ml" "-build"]
-  ["ocaml" "setup.ml" "-configure" "--enable-tests"] {with-test}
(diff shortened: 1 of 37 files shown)


packages/angstrom/angstrom.0.1.0/opam
```diff
@@ -10,26 +10,22 @@ build: [
...
-    "--prefix"
-    prefix
+    "--prefix" prefix
     "--%{base-unix:enable}%-unix"
     "--%{lwt:enable}%-lwt"
     "--%{async:enable}%-async"
+    "--enable-tests" {with-test}
   ]
   ["ocaml" "setup.ml" "-build"]
-  ["ocaml" "setup.ml" "-configure" "--enable-tests"] {with-test}
```
avsm, line 20: what's this removal fixing? If I remember oasis right, the tests had to be explicitly enabled to work

avsm:
It looks like there are systematic fixes to tests that you've applied here. A brief explanation would be useful about the failure class and what your changes do. The `remove` field removal looks good to me.

packages/angstrom/angstrom.0.1.0/opam
kit-ty-kate: yes this is just move to the line above. The opam file has redundant instructions for configure and build which came from a wrong assumptions about how the `with-test` variable works. This is just a simplification/ci-optimisation

mseri:
The test change does not fix an issue but cleans up the build instruction by moving the necessary flags into a single configure call instead of duplicating the lines. This is in line with what we are doing to all other packages and how also dune behaves when generating the opam files. I think this should be enough of a clarification for the kind of change.

PR #18395, opened 2021-03-25
No spacetime in 4.12

dra27:
Slightly academic, but the spacetime option remained in the package. The `ocaml-option-spacetime` package won't install with 4.12, so this is a belt-and-braces "remove an option which shouldn't be there".

cf. https://github.com/ocaml/ocaml/pull/10310. It's probably best to hold this off until either the pinning updates (when I finish those) or 4.12.1, given that the change in the opam file will trigger rebuilds.


mseri:
Thanks.  I’ll tag it as do not merge for now. I’ll un tag once 4.12.1 or the pinning changes are ready

PR #18398, opened 2021-03-26
Remove bls12-381*.0.4.1 because of wrong usage of virtual pkg def

dannywillems:
See https://github.com/ocaml/dune/issues/4372.


mseri:
Can you mark it `available:false` instead of removing it?

kit-ty-kate:
I'm not seeing the issue described in https://github.com/ocaml/dune/issues/4372 inside these packages. Could you provide a test case? I commented on the dune issue with a fix to use it outside the project

kit-ty-kate:
Closing for now. Feel free to reopen if needed

PR #18400, opened 2021-03-26
Two minor tweaks

dra27:
Shedding weight from a very old windows branch... a profound correction to an old ocamlfind package and a tweak to the OCaml 3.12.0 package


mseri:
Looks good. Seems that the CI cannot test these packages but looks correct

mseri:
Tested localy

PR #18408, opened 2021-03-30
[New package] ez_pgocaml.1.0.0

lefessan:
This is the first release of ez_pgocaml, a library that we have been using for a few years now in projects such as TzScan and other ones. It has some helpers for pgocaml, plus a migration system, that can ne used upgrade/downgrade a database in production (a bit like what Ruby on Rails provides).

### DIFF
--- a/packages/ez_pgocaml/ez_pgocaml.1.0.0/opam
+++ b/packages/ez_pgocaml/ez_pgocaml.1.0.0/opam
@@ -1,5 +1,5 @@
 opam-version: "2.0"
-synopsis: "a simple library to work with pgocaml"
+synopsis: "A simple library to work with pgocaml"
 maintainer: ["Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>"]
 authors: ["Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>"]
 license: "GPL-3.0-or-later"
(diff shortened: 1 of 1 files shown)


packages/ez_pgocaml/ez_pgocaml.1.0.0/opam
```diff
@@ -0,0 +1,30 @@
+opam-version: "2.0"
+synopsis: "A simple library to work with pgocaml"
+maintainer: ["Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>"]
+authors: ["Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>"]
+license: "GPL-3.0-or-later"
```
avsm, line 5: This says GPLv3, but the repo says LGPLv2.1+linking exception. https://github.com/OCamlPro/ez_pgocaml/blob/master/LICENSE.md

packages/ez_pgocaml/ez_pgocaml.1.0.0/opam
```diff
@@ -0,0 +1,30 @@
...
+homepage: "https://github.com/ocamlpro/ez_pgocaml"
+bug-reports: "https://github.com/ocamlpro/ez_pgocaml/issues"
+depends: [
+  "dune" {>= "2.0"}
```
avsm, line 9: This should also have the minimum version of OCaml that you expect to support; e.g. `"ocaml" {>="4.06.0"}`

lefessan: Thanks, I should have used `drom` to package it, to avoid all these problems...

lefessan:
Oups, I will fix some other issues I just noticed, please do not merge if CI succeeds now.

lefessan:
I removed the WIP status, it's now ok for merging.

kit-ty-kate:
Thanks!

PR #18418, opened 2021-03-30
+httpaf.0.7.1

seliopou:


packages/httpaf-async/httpaf-async.0.7.1/opam
```diff
@@ -0,0 +1,25 @@
...
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name] {with-test}
+]
+depends: [
+  "ocaml" {>= "4.03.0"}
+  "dune" {>= "1.5.0"}
+  "faraday-async"
+  "async"
```
mseri, line 18:
```suggestion
  "async" {< "v0.14"}
```

Can you upstream this change?

kit-ty-kate:
Thanks! Could you return the fixes in the opam files upstream so that it can be taken for the next releases?

PR #18425, opened 2021-03-31
directories.0.3

zapashcanon:
Hi,

This release is just because we dropped the dependency on ctypes-foreign (we now use only ctyps.stubs).


zapashcanon:
I don't understand what happens with flambda and centos-8 but everything else seems OK.

kit-ty-kate:
Apart from that all seems good. Thanks!

zapashcanon:
Thanks ! I'll upstream things and update the dune version.

PR #18432, opened 2021-04-01
Add coq-of-ocaml version 2.5.0

clarus:
Add the version 2.5.0 of `coq-of-ocaml`.


kit-ty-kate:

clarus:
OK thanks, I have not seen that I missed the `dune` dependency. I did the back-port. A pull-request to precise the dune version: https://github.com/ocaml/opam-repository/pull/18437

PR #18437, opened 2021-04-02
Precise the dune version for coq-of-ocaml

clarus:
Precise the version of dune due to the change of the dune <-> Merlin protocol.


mseri:
Thanks

PR #18440, opened 2021-04-03
malfunction 0.3 is not compatible with omd 2.0

anton-trunov:
I see the following compilation error with omd 2.0.0~alpha1
```
File "test/test.ml", line 160, characters 11-29:
160 |   let open Omd_representation in
                 ^^^^^^^^^^^^^^^^^^
Error: Unbound module Omd_representation
```


mseri:
Thanks!

anton-trunov:
Thanks for fixing it -- I totally forgot about the tilda :)

mseri:
This should make also the linter happy given that we are there

Issue #17185, opened 2020-09-12
Version not set correctly in "lwt" version 4.2.1 (required by stog)


fccm:
While trying to install "stog" with opam I got an error at configure time with stog checking which version of lwt.unix is available:
```
eval $(opam env --switch 4.06.0 --set-switch)
opam install stog.0.18.0-1
# checking for lwt.unix... Package lwt.unix found with version dev, but wanted >= 2.5
```
Indeed in `~/.opam/4.06.0/lib/lwt/META` there is:
```
version = "dev"
```
When I edit this `META` file to set a correct version:
```
version = "4.2.1"
```
(and same for `package "unix"` sub-section)
Then when I run `opam install stog.0.18.0-1` the install proceeds correctly as expected.

A patch for the package `lwt.4.2.1` is required to install `stog` correctly.

mseri:
Can you send a PR, copying `lwt.4.2.1` into `lwt.4.2.1-1` and adding the required patch?

fccm:
Yes, sure!
I expected to do so when I have a moment.

Why creating a new package version? I thought in opam we fix existing packages without renaming it.

mseri:
We are trying to have patch versions when the change involves patches or functional modifications. We try to modify in place only if we need to fix the metadata (constraints, urls, maintainer names, ...).

PR #18455, opened 2021-04-06
Remove all unnecessary name and version fields

kit-ty-kate:
I checked and this does not make opam rebuild any installed packages, so this change should be painless for users.


mseri:
There was some reason for which `lwt` actually needed the version field in the past. There should be a discussion about it somewhere

mseri:
Do you know why this funny failure?
```
2021-04-06 13:53.14: Job failed: "Assert_failure src/repository/opamDownload.ml:139:2"
```

avsm:
That line is 
```
  assert (url.OpamUrl.backend = `http);
```

in opam trunk -- looks like a real bug as it should never be triggered in opam 2.1 (cc @altgr @rjbou @dra27)

rjbou:
Is it reproducible ?

dra27:
opam-repo-ci is calling `OpamDownload.download` with the source URL which isn't guaranteed to be downloadable (see [here](https://github.com/ocurrent/opam-repo-ci/blob/8bedf7791d5fe8daff526211a7cd557fa480684d/lib/lint.ml#L93)). Various of the `camltc` packages have git URLs.

It would be easier to do be doing the equivalent of `opam source` here I think (I can't remember the API). The only change is opam is that arguably (pun intended) it should be an `Invalid_arg` not an `assert`.

kit-ty-kate:
I tried to use the API used by opam source on Sunday, however this is inside the opam-client library which is currently incompatible with opam-repo-ci due to its transitive dependency on extlib (extlib < 1.7.8 and base64 cannot be used in the same project)

dra27:
Is there any reason not to shell out to `opam source` directly? The API's not providing much benefit here - the download job shells out to curl (or git!) and the extract job shells out to tar.

rjbou:
issue opened in opam-repo-ci bugtracker, better continue the discussion there

dra27:
The issue in lwt.4.2.0 and lwt.4.2.1 is with the sources themselves. Dune looks at the opam files _in the source distribution_ as the primary source of a packages version (although cf. https://github.com/ocaml/dune/issues/4457) which is what was wrong in #17185. The opam file from the repository (or, in general, the one opam used to determine what to do to build and install the package) is not directly available to the build, although in opam 2.1 the `opamfile` package variable contains its location (cf. https://github.com/ocaml/opam/pull/4402).

`name` and `version` in an opam repository are unnecessary - indeed, they're confusing because they're ignored by opam (they're always determined from the directory in which the `opam` file was found).

Having removed these unnecessary fields, do we have issues/PRs open for linting to keep them removed?

kit-ty-kate:
> Having removed these unnecessary fields, do we have issues/PRs open for linting to keep them removed?

This is already in opam-repo-ci since https://github.com/ocurrent/opam-repo-ci/pull/70

mseri:
> `name` and `version` in an opam repository are unnecessary - indeed, they're confusing because they're ignored by opam (they're always determined from the directory in which the `opam` file was found).

Then I think we should go on and merge

PR #18465, opened 2021-04-07
[new release] http-cookie (3.0.0)

bikallem:
HTTP cookie library for OCaml

- Project page: <a href="https://github.com/lemaetech/http-cookie">https://github.com/lemaetech/http-cookie</a>

##### CHANGES:

- Backwards incompatible change: remove `base-unix` dependency. Uses own `date_time` instead of `Unix.tm`.

### DIFF
--- a/packages/http-cookie/http-cookie.3.0.0/opam
+++ b/packages/http-cookie/http-cookie.3.0.0/opam
@@ -26,12 +26,12 @@ build: [
   ]
 ]
 dev-repo: "git+https://github.com/lemaetech/http-cookie.git"
-x-commit-hash: "8d9e601e994ab43ba610d46856b1ea939671041e"
+x-commit-hash: "063f3160517f3c8bfe848d0f0d80b76450a9034c"
 url {
   src:
     "https://github.com/lemaetech/http-cookie/releases/download/v3.0.0/http-cookie-v3.0.0.tbz"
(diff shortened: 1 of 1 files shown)


mseri:
This was released already three months ago: https://github.com/ocaml/opam-repository/commit/34561af01b64b35456d524f559325d995fcfaa19

Is it possible to make this new release with a new version instead of changing the sources?

bikallem:
I accidentally deleted the old release 3.0 artifacts in github and had to recreate them again. Is this okay?

mseri:
Oh, ok. Indeed I was looking at the diff and it looks like it is just a reformat.
By the way, you can get the old release artifacts from the opam cache, I you replace the tbz in the release with what you get from
```
wget -O http-cookie-v3.0.0.tbz https://opam.ocaml.org/cache/sha256/66/66b131ca6e63bc46ef859fc8924c815377377abfb2c52324c658d9c000130475
```
then you can also avoid doing the new release and changing anything.

bikallem:
Ah ok. In that case then we can close this pull request.

mseri:
Thanks for the very fast reaction and the clarifications.

Also, if you have not yet done it, feel free to announce your releases on discuss.ocaml.org. We have a community category and an announcement tag for this purpose

PR #18476, opened 2021-04-09
Add new package freeton_ocaml_sdk, version 0.2.0

lefessan:
OCaml SDK for the Free TON blockchain.


lefessan:
I don't really understand why the CI fails. If I compare the logs of the CI with the ones of the Github CI for freeton_ocaml_sdk (https://pipelines.actions.githubusercontent.com/q2LTcSEqFOb9JfpG5qCzsBKT4voha0awLb4N8qRhdO8U3HGonX/_apis/pipelines/1/runs/126/signedlogcontent/3?urlExpires=2021-04-10T09%3A06%3A52.5438527Z&urlSigningMethod=HMACV1&urlSignature=ZEbwUylVytPWOCvKGjRTQ8t6%2BUaHvs3AQktzwkpMe5E%3D for example), the same Rust packages and versions are compiled, but OCaml CI fails where Gihub CI succeeds.

lefessan:
Looks like it builds correctly on Debian unstable and testing, but not on Debian stable. I use Debian stable myself, and it works, but I have installed Rust separately. It's probably a problem with the default (obsolete) Rust version on most systems.
What's the policy of opam-repository in this case ?

mseri:
I think in this case we can just mark it as failable on the ci for debian-stable, I'll wait for @kit-ty-kate opinion

mseri:
The problem, to me, is that this seems to fail in practically all builds, not just debian stable

lefessan:
Yes, but I have no control over these environments, how could I ask for a different version of rustc ? I am already asking for conf-rust-2018, but even with that, it still fails because of indirect dependencies of Rust packages (by the way, some of which come from ocaml binding packages like ocaml-interop).

avsm:
This will break for users with sandboxing active as well, since you are downloading from within the job. Have you considered vendoring the rust dependencies into an archive and publishing that on opam?

We'll need a separate set of `conf-rust` packages that do a setenv to the right version of Cargo, if you need to discriminate to a higher version of Rust. That doesn't exist right now.

kit-ty-kate:
If you need an example of a working package that is using cargo, there is one here: https://gitlab.com/tezos/tezos-rust-libs/

lefessan:
> Have you considered vendoring the rust dependencies into an archive and publishing that on opam?
Ok, I can do that.

> We'll need a separate set of `conf-rust` packages that do a setenv to the right version of Cargo, if you need to discriminate to a higher version of Rust. That doesn't exist right now.
I don't understand this part. If there is only one version of cargo on the system, how is setenv going to solve the problem ?

lefessan:
I fixed the sandboxing problem (all Rust packages are now included in the verdor/ directory), but still no clue on how to fix the Rust version problem.
It succeeds on debian-testing and debian-unstable, it sounds good enough for me. Users on other systems will have to upgrade their Rust version, it's the same as for a binding to a C-library for which no packaged version exists, no ? 
The only other solution I see would be to create a "rustup" package that would install Rust as an opam package, but ending up with one Rust installation per switch doesn't seem like a good idea either.

lefessan:
(I moved 2 packages out of the PR, they are now in a different Github repository)

avsm:
@lefessan wrote:
> If there is only one version of cargo on the system, how is setenv going to solve the problem ?

By extending the PATH to cover the installation location of that Rust toolchain. I haven't looked into the specifics of how that might work.

I would highly recommend spending some time thinking about how to make this pass CI on at least debian-stable, since then your packages will benefit from the other benefits of bulk opam builds.  It's not necessary to make it pass on every distro supported of course.

lefessan:
> By extending the PATH to cover the installation location of that Rust toolchain. I haven't looked into the specifics of how that might work.

The current problem is not that it cannot find the Rust toolchain, but that the only available toolchain on the system is out-of-date. 

> I would highly recommend spending some time thinking about how to make this pass CI on at least debian-stable, since then your packages will benefit from the other benefits of bulk opam builds.

A new Debian stable (Bullseye ) is expected soon, so the problem might disappear by itself soon. Also, I don't have a clue on how I could change the CI from an opam package (there is depext, but depext only works with existing system packages, which is not the case for the new Rust version).

Also, my package is using Github Actions for CI, and it doesn't fail with ubuntu-latest there. Why are all ubuntu versions on OCaml CI failing, I would have expected at least Ubuntu 20.4 to succeed.

avsm:
> Also, my package is using Github Actions for CI, and it doesn't fail with ubuntu-latest there. Why are all ubuntu versions on OCaml CI failing, I would have expected at least Ubuntu 20.4 to succeed.

Because OCaml CI uses upstream Ubuntu, whereas GitHub Actions installs newer versions of toolchains in their environment.

mseri:
I think the best we can do with rust-based packages that need more recent version than just 2018 edition is to add a post-message of failure type i forming the user that a version of rust > x is needed, install it manually and then try again. If we add this message I would be inclined to say that this could be merged. Then in the future when the various distros move to more recent versions of rust this will become lesser and lesser of an issue

lefessan:
I don't know how the opam-repository should deal with such packages. As this PR was stuck here, we decided to create our own small repository to distribute these packages ( https://github.com/OCamlPro/ocp-opam-repository ), and later to dockerize the application. Using docker is now the advised way to install freeton_wallet, as it brings cross-platform portability (linux-macosx-windows).
Anyway, if you decide to accept such packages in the future, keep me informed and I will be submit the most recent version of these packages.

mseri:
The docker approach using an overlay repo seems more robust and easier also to manage indeed.

We have merged another package, also using cargo vendor and not accessing the network for installation, and providing the post-install fail message pointing out the required rust version. I think that is the best we can do for the moment, and makes more sense for small packages imo. I leave it up to you if you want to release here or not.

In any case I am going to close this since a new release would require a new PR in any case.

lefessan:
What is the name of that package ? It might be useful to look at what they have done...

mseri:
Still ongoing discussion: https://github.com/ocaml/opam-repository/pull/19057
Merged (but works fine with older versions of rust): https://github.com/ocaml/opam-repository/pull/19065

lefessan:
Thanks for the links !

PR #18482, opened 2021-04-11
Release Caqti 1.5.0.

paurkedal:
There is only two changes since the last release,

- Request the full UTF-8 character for the MariaDB connection.
- Support int16 and enum types for parameters and rows.

but some people hit some limitations due to the increased DB type safety introduced by 1.4.0, which is addressed by the second point.


mseri:
Looks good, thanks

PR #18485, opened 2021-04-12
redirect ocaml/ocaml issues to ocaml/opam-repository

### LINKED ISSUE #18479: In opam-base-compiler packages, do not suggest to file a bug report in ocaml/ocaml
Currently, a failure in `opam switch create` prints a suggestion to file a bug in the ocaml/ocaml repository, e.g. https://github.com/ocaml/ocaml/issues/10331 .  This seems to come from the opam-base-compiler OPAM files, e.g. https://github.com/ocaml/opam-repository/blob/e248948afd8b084309f2db5165d753f2b056fc16/packages/ocaml-base-compiler/ocaml-base-compiler.4.11.1/opam#L34-L37

I can't remember the last time when the failure was an issue with the OCaml core package.  Recent "bug" reports were due to missing software components (e.g. XCode command-line tools not installed on macOS) or an unsupported platform (e.g. macOS-ARM64 with 4.11.1).  This is just noise in the ocaml/ocaml repository.

Could we please remove this suggestion from the ocaml-base-compiler packages?

avsm:
Most issues with OCaml compilation these days stem from some packaging problem (such as disaggregation from the compiler). This PR changes the default location for bug reports to the opam-repository issue tracker.  The opam-repo team can triage these and transfer relevant issues to ocaml/ocaml.

Fixes #18479

### DIFF
--- a/packages/graphics/graphics.3.09.1/opam
+++ b/packages/graphics/graphics.3.09.1/opam
@@ -1,6 +1,6 @@
 opam-version: "2.0"
 maintainer: "David Allsopp <david.allsopp@metastack.com>"
-bug-reports: "https://github.com/ocaml/ocaml/issues"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
 dev-repo: "git+https://github.com/ocaml/ocaml.git"
 authors: [ "Xavier Leroy"
            "Jun Furuse"
--- a/packages/graphics/graphics.3.07+1/opam
+++ b/packages/graphics/graphics.3.07+1/opam
@@ -1,6 +1,6 @@
 opam-version: "2.0"
 maintainer: "David Allsopp <david.allsopp@metastack.com>"
-bug-reports: "https://github.com/ocaml/ocaml/issues"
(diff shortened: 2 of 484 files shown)


packages/graphics/graphics.3.09.1/opam
```diff
@@ -1,6 +1,6 @@
 opam-version: "2.0"
 maintainer: "David Allsopp <david.allsopp@metastack.com>"
-bug-reports: "https://github.com/ocaml/ocaml/issues"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
```
avsm, line 3: @dra27 should all these graphics ones redirect to https://github.com/ocaml/graphics/issues instead?

dra27: For the 5.0.0+ ones, yes. The others are special packages which compile the graphics package from the main distribution and add it, so issues are packaging issues and should be reported to opam-repository. (slightly pedantically, given that I'm a maintainer of both repos........)

avsm: Thanks! I think this is good to merge then.

PR #18489, opened 2021-04-12
Add conf-tree-sitter

joseeMDS:
Tree-sitter is a parser generator tool for building a concrete syntax tree for a source file and efficiently update it.

This PR add a package to check if it is installed in the system

### DIFF
--- a/packages/conf-tree-sitter/conf-tree-sitter.0.19.4/opam
+++ b/packages/conf-tree-sitter/conf-tree-sitter.0.19.4/opam
@@ -1,0 +1,24 @@
+opam-version: "2.0"
+maintainer: "José Eduardo <josee.edumont@gmail.com>"
+authors: ["Max Brunsfeld"]
+homepage: "https://tree-sitter.github.io/"
+bug-reports: "https://github.com/tree-sitter/tree-sitter/issues"
+license: "MIT"
+build: ["tree-sitter" "--version"]
+depexts: [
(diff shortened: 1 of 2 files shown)


joseeMDS:
I squashed the commits since it were just small fix commits, but im not sure what im doing wrong here

mseri:
The failures are due to missing tree-sitter in most of the distributions. We could allow this to fail in all of them, is there a tree sitter package for debian/Ubuntu/fedora... that you know of?

joseeMDS:
For debian i found [this](https://packages.debian.org/experimental/libtree-sitter0), but since it is experimental i was not sure if i should add it, i followed [opam-depext](https://github.com/ocaml-opam/opam-depext) table but i only found the ones that i added

joseeMDS:
How should i add the OS that dont have tree-sitter by default to opam file?

mseri:
@kit-ty-kate can we merge this in the current state, or you think we can improve it further?

kit-ty-kate:
On Alpine the tree-sitter binary isn't available but it's marked as TODO so I'll leave it like this for the moment. I also moved the version number to be just `1` as giving it the version number of the latest know version of the tool doesn't make much sense unless the version number is going to be checked to be >= version. For the `conf-*` packages we usually prefer using natural numbers starting from `1` and increasing the version number by one on major changes only (adding distributions does *not* count as major change).

Other than that this is now ready. Thanks!

PR #18501, opened 2021-04-14
Package soupault.2.6.0

dmbaturin:
### `soupault.2.6.0`
Static website generator based on HTML rewriting
A website generator that works with page element tree rather than text
and allows you to manipulate pages and retrieve metadata from
existing HTML using arbitrary CSS selectors.

With soupault you can:

* Generate ToC and footnotes.
* Insert file content or an HTML snippet in any element.
* Preprocess element content with external programs (e.g. run `<pre>` tags through a highlighter)
* Extract page metadata (think microformats) and render it using a Jingoo template or an external script.
* Export extracted metadata to JSON.

Soupault is extensible with Lua (2.5) plugins and provides an API for element tree manipulation,
similar to web browsers.

The website generator mode is optional, you can use it as post-processor for existing sites.

---
* Homepage: https://www.soupault.app
* Source repo: git+https://github.com/dmbaturin/soupault
* Bug tracker: https://github.com/dmbaturin/soupault/issues

---

### DIFF
--- a/packages/soupault/soupault.2.6.0/opam
+++ b/packages/soupault/soupault.2.6.0/opam
@@ -1,0 +1,55 @@
+opam-version: "2.0"
+synopsis: "Static website generator based on HTML rewriting"
+description: """\
+A website generator that works with page element tree rather than text
+and allows you to manipulate pages and retrieve metadata from
+existing HTML using arbitrary CSS selectors.
+
+With soupault you can:
+
+* Generate ToC and footnotes.
+* Insert file content or an HTML snippet in any element.
+* Preprocess element content with external programs (e.g. run `<pre>` tags through a highlighter)
+* Extract page metadata (think microformats) and render it using a Jingoo template or an external script.
+* Export extracted metadata to JSON.
+
+Soupault is extensible with Lua (2.5) plugins and provides an API for element tree manipulation,
+similar to web browsers.
+
+The website generator mode is optional, you can use it as post-processor for existing sites."""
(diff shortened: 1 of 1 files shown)


packages/soupault/soupault.2.6.0/opam
```diff
@@ -0,0 +1,55 @@
...
+  "containers"
+  "odate"
+  "spelll"
+  "base64"
+  "jingoo" {>= "1.4.2"}
+  "tsort" {>= "2.0.0"}
+  "lua-ml" {>= "0.9.2"}
+]
+build: [
+  ["dune" "subst"] {pinned}
```
mseri, line 45:
```suggestion
  ["dune" "subst"] {dev}
```

mseri: Fixed in dune 2.8, we can fix it manually here

mseri:
Thanks

packages/soupault/soupault.2.6.0/opam
dmbaturin: Should I make this change in my repo? I have to admit I don't understand the effect of your suggestion.

mseri: Yes please, I thought it was an autogenerated opam file when I commented.
You can see the explanation here: https://github.com/ocaml/dune/pull/3647
In short, dune docs and autogeneration were using pinned instead of dev by mistake.

I just noticed that dune documentation has not been updated and sent a PR also for that.

PR #18502, opened 2021-04-14
mark conduit*3 as unavailable

hannesm:
I think this is what we concluded -- that conduit 3 should be avoided (and will be revised in the future)

//cc @avsm @samoht @dinosaure


dinosaure:
Agree with this where it seems that nobody really use `conduit.3.0.0`. As `cohttp.3.0.0`, the best is to remove and release `conduit.4.0.0` may be, I'm not really sure.

avsm:
I'm generally reluctant to remove releases from opam-repository without an exceptional reason, as you dont know which downstream repositories are using this package.  @samoht @hannesm would you consider releasing a conduit.4.0.0 that is the same as the conduit.2.x series so it can continue forwards from a versioning perspective?  A note similar to the one in cohttp's CHANGES file that @mseri did would suffice to explain the reasoning.

hannesm:
> would you consider releasing a conduit.4.0.0 that is the same as the conduit.2.x series so it can continue forwards from a versioning perspective?

This is a great idea, we should do that. I can work on that with changes for tls 0.13 and dns 5.0 (IPv6) included.

avsm:
Many thanks @hannesm! I'll close this PR for now then.

hannesm:
I don't understand why closing this PR is a way forward. There is CI trouble with conduit*3, and it looks like nobody has interest to support & develop conduit 3. Why should it stay as available in opam-repository?

mseri:
I agree with @hannesm, we should probably mark it unavailable sooner rather than later, then push to have conduit 4.0 out immediately. Otherwise, we risk that people start updating their code or using the new library which will be de-facto unmaintained.

PR #18515, opened 2021-04-16
new file:   packages/pardi/pardi.3.2.1/opam

UnixJunkie:
easier to install, because less deps


UnixJunkie:
finally, I will also ship another feature while I am at it

PR #18519, opened 2021-04-16
Tezos 9.0rc2

pirbo:
This is the hijack of your CI I do each time I'm about to do a major release of the tezos/tezos packages in order to ensure that packaging is ready for the real release (expected 2021-04-21) :-)


pirbo:
Thank you. see your in a few days...

PR #18520, opened 2021-04-16
Various fixes

kit-ty-kate:
most of them are old versions and do not require their maintainers to be pinged. Those that are latest versions are due to the upcoming OCaml 4.13 and will have time to upgrade anyway

### DIFF
--- a/packages/obytelib/obytelib.1.1/opam
+++ b/packages/obytelib/obytelib.1.1/opam
@@ -14,6 +14,7 @@ remove: [
 ]
 depends: [
   "ocaml" {>= "4.03.0" & < "4.13"}
+  "conf-which"
   "ocamlbuild"
 ]
 install: [make "install"]
--- a/packages/obytelib/obytelib.1.1/opam
+++ b/packages/obytelib/obytelib.1.1/opam
@@ -13,7 +13,7 @@ remove: [
   [make "uninstall"]
 ]
 depends: [
-  "ocaml" {>= "4.03.0"}
+  "ocaml" {>= "4.03.0" & < "4.13"}
   "ocamlbuild"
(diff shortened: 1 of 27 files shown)


packages/obytelib/obytelib.1.1/opam
```diff
@@ -13,7 +13,7 @@ remove: [
   [make "uninstall"]
 ]
 depends: [
-  "ocaml" {>= "4.03.0"}
+  "ocaml" {>= "4.03.0" & < "4.13"}
```
mseri, line 16:
```suggestion
  "ocaml" {>= "4.03.0" & < "4.13"}
  "conf-which"
```

packages/obytelib/obytelib.1.2/opam
```diff
@@ -13,7 +13,7 @@ remove: [
   [make "uninstall"]
 ]
 depends: [
-  "ocaml" {>= "4.04.0"}
+  "ocaml" {>= "4.04.0" & < "4.13"}
```
mseri, line 16:
```suggestion
  "ocaml" {>= "4.04.0" & < "4.13"}
  "conf-which"
```

packages/ocamlpp/ocamlpp.1.1/opam
```diff
@@ -13,7 +13,7 @@ remove: [
   [make "uninstall"]
 ]
 depends: [
-  "ocaml" {>= "4.03.0"}
+  "ocaml" {>= "4.03.0" & < "4.13"}
```
mseri, line 16:
```suggestion
  "ocaml" {>= "4.03.0" & < "4.13"}
  "conf-which"
```

PR #18524, opened 2021-04-18
[new release] vercel and lambda-runtime (0.1.0)

anmonteiro:
A custom runtime for Vercel.com (Now v2) written in OCaml

- Project page: <a href="https://github.com/anmonteiro/aws-lambda-ocaml-runtime">https://github.com/anmonteiro/aws-lambda-ocaml-runtime</a>

##### CHANGES:

- Initial public release

### DIFF
--- a/packages/vercel/vercel.0.1.0/opam
+++ b/packages/vercel/vercel.0.1.0/opam
@@ -17,7 +17,7 @@ depends: [
   "base64"
   "piaf"
   "ppx_deriving_yojson"
-  "lambda-runtime"
+  "lambda-runtime" {=version}
   "alcotest" {with-test}
 ]
 synopsis:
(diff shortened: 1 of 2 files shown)


packages/vercel/vercel.0.1.0/opam
```diff
@@ -0,0 +1,39 @@
...
+depends: [
+  "ocaml" {>= "4.08"}
+  "dune" {>= "1.7"}
+  "result"
+  "yojson"
+  "lwt"
+  "base64"
+  "piaf"
+  "ppx_deriving_yojson"
+  "lambda-runtime"
```
mseri, line 20:
```suggestion
  "lambda-runtime" {=version}
```

we tend to enforce strict bounds on packages that come together from the same repository. Before we did this, bounds were getting messy and metadata was getting hard to manage. Would this be a problem for you? Do you prefer keeping this unconstrained for some specific reason?

anmonteiro: This should be OK. I'll fix it going forward in the upstream repo, if that works.

kit-ty-kate:
Looks good. Thanks!

PR #18526, opened 2021-04-18
Conf-qt: repair depexts for new Ubuntu 

Kakadu:


Kakadu:
It complains about missing `url` but I think that this should not be fixed

kit-ty-kate:
> Hm, our CI is not doing the right thing here. A depexts change should be considered significant and therefore trigger a distro run, shouldn't it @kit-ty-kate ?

uh, indeed. https://github.com/ocurrent/opam-repo-ci/pull/100 should fix this

Kakadu:
@kit-ty-kate Can we revisit this PR? The last issue we have is `(lint) (failed: Error in conf-qt.5.2.1: error 60: Upstream check failed: "No url defined")` but I hope we can close eyes on this.

kit-ty-kate:
Thanks!

PR #18528, opened 2021-04-18
geoip: Fix compilation on mac os

mebsout:
Add include paths and library paths where lib is installed with either macports or homebrew.
(Not tested with homebrew but should be ok.)

### DIFF
--- a/packages/geoip/geoip.0.0.3/opam
+++ b/packages/geoip/geoip.0.0.3/opam
@@ -26,9 +26,10 @@ depexts: [
   ["GeoIP-devel"] {os-distribution = "centos"}
 ]
 install: ["ocaml" "setup.ml" "-install"]
-synopsis: "Bindings to GeoIP database library."
+synopsis: "Bindings to GeoIP database library"
 description:
   "GeoIP lets you discover information about a specific IP address."
+license: "LGPL-2.1 with OCaml linking exception"
 flags: light-uninstall
 url {
   src: "https://ygrek.org/p/release/ocaml-geoip/ocaml-geoip-0.0.3.tar.gz"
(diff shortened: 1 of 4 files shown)


packages/geoip/geoip.0.0.3/opam
```diff
@@ -5,29 +5,23 @@ homepage: "https://ygrek.org/p/ocaml-geoip/"
 doc: ["https://ygrek.org/p/ocaml-geoip/api/index.html"]
 bug-reports: "https://github.com/ygrek/ocaml-geoip/issues"
 dev-repo: "git://github.com/ygrek/ocaml-geoip.git"
+license: "LGPL-2.1-only with OCaml-LGPL-linking-exception"
 tags: ["org:ygrek"]
 build: [
   ["ocaml" "setup.ml" "-configure" "--prefix" prefix]
-  ["ocaml" "setup.ml" "-build"]
+  ["ocaml" "setup.ml" "-build" "-cflags" conf-libgeoip:cflags]
```
kit-ty-kate, line 12: are you sure the `-cflags -I<path>` worked in the first place?

mebsout: Indeed, it does not work with this. It needs `-cflags "-I <path>"`, not sure if this is possible with pkg-config. Also I did not find a way to pass -L options to the linker with ocamlbuild, but it works by setting the env variable `LIBRARY_PATH` (with `build-env:`). Again, not sure how to do this with a conf package.

avsm: best off porting geoip to dune and using a configure.ml as other packages do in order to do the directory probing to find out where geoip is installed. This one's not the best thing to fix from the opam packaging unfortunately...

mseri:
I have opened an issue on geoip. I will close this PR

Issue #18475, opened 2021-04-09
unable to install opam graphics package


AntoineDevFr:
Hey, I'm trying to install graphics package but I have always the same message:

```
Last login: Fri Apr  9 21:57:05 on ttys000
antoinebanchet@MacBook-Pro-de-Antoine ~ % opam install graphics
The following actions will be performed:
  ∗ install   conf-libX11 1     [required by graphics]
  ∗ install   graphics    5.1.1
  ↻ recompile ocamlfind   1.9.1 [uses graphics]
===== ∗ 2   ↻ 1 =====
Do you want to continue? [Y/n] Y

<><> Gathering sources ><><><><><><><><><><><><><><><><><><><><><><><><><><>  🐫 
[graphics.5.1.1] found in cache
[ocamlfind.1.9.1] found in cache

<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><>  🐫 
[ERROR] The compilation of conf-libX11 failed at
        "/Users/antoinebanchet/.opam/opam-init/hooks/sandbox.sh build sh -exc
        PKG_CONFIG_PATH=\"$PKG_CONFIG_PATH:/usr/local/lib/pkgconfig:/opt/X11/lib/pkgconfig\"
        pkg-config --libs x11".

#=== ERROR while compiling conf-libX11.1 ======================================#
# context     2.0.8 | macos/arm64 | ocaml-base-compiler.4.12.0 | https://opam.ocaml.org#323145ae
# path        ~/.opam/default/.opam-switch/build/conf-libX11.1
# command     ~/.opam/opam-init/hooks/sandbox.sh build sh -exc PKG_CONFIG_PATH="$PKG_CONFIG_PATH:/usr/local/lib/pkgconfig:/opt/X11/lib/pkgconfig" pkg-config --libs x11
# exit-code   1
# env-file    ~/.opam/log/conf-libX11-1098-d8a4ae.env
# output-file ~/.opam/log/conf-libX11-1098-d8a4ae.out
### output ###
# + PKG_CONFIG_PATH=/Users/antoinebanchet/.opam/default/lib/pkgconfig:/usr/local/lib/pkgconfig:/opt/X11/lib/pkgconfig
# + pkg-config --libs x11
# Package xproto was not found in the pkg-config search path.
# Perhaps you should add the directory containing `xproto.pc'
# to the PKG_CONFIG_PATH environment variable
# Package 'xproto', required by 'x11', not found



[...]

PR #18530, opened 2021-04-19
conf-libx11: brew now installs in /opt

mseri:
See https://github.com/ocaml/opam-repository/issues/18475

Should we add a post-install (failure?) message mentioning to be careful with the path? An alternative could be to update this as we did in `conf-libssl.3`

### DIFF
--- a/packages/conf-libX11/conf-libX11.1/opam
+++ b/packages/conf-libX11/conf-libX11.1/opam
@@ -8,7 +8,7 @@ build: [
   ["pkg-config" "x11"] {os != "macos"}
   [
     "sh" "-exc"
-    """PKG_CONFIG_PATH="$PKG_CONFIG_PATH:/usr/local/lib/pkgconfig:/opt/X11/lib/pkgconfig" pkg-config --libs x11"""
+    """PKG_CONFIG_PATH="$PKG_CONFIG_PATH:/usr/local/lib/pkgconfig:/opt/X11/lib/pkgconfig:/opt/X11/share/pkgconfig" pkg-config --libs x11"""
   ] {os = "macos"}
 ]
 depends: ["conf-pkg-config" {build}]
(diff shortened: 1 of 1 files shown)


mseri:
I moved to macports a while ago, so I cannot test this. Don’t merge until somebody can tell us if it works. MAybe ping @smorimoto

kit-ty-kate:
I can confirm this works. Thanks a lot!

smorimoto:
This change caused the same error again…
https://github.com/ocsigen/js_of_ocaml/runs/2597248226

mseri:
@kit-ty-kate should we revert this, or re-add `"xorgproto"`?

kit-ty-kate:
@smorimoto are you sure this is the same error message? `conf-libX11` successfully installed. Now what's going on is that `graphics` doesn't find the library (see: https://github.com/ocaml/graphics/blob/master/src/discover/discover.ml#L30)

So it looks to be like `graphics` needs to be patch now.

kit-ty-kate:
The alternative would be to copy all the needed `.pc` file to the local opam `PKG_CONFIG_PATH` as suggested by @mseri above

mseri:
> Now what's going on is that `graphics` doesn't find the library (see: https://github.com/ocaml/graphics/blob/master/src/discover/discover.ml#L30)

It should though, in the end it checks the env variable before putting its own path: https://github.com/ocaml/graphics/blob/7643bef3d5748dc83b380203b424085f121b69ef/src/discover/discover.ml#L35

kit-ty-kate:
sure but the actual path of xquartz is never set before.

Looking a bit deaper into it, it seems that xorgproto installs the `libx11` package on my test machine at least (somehow??, it's not even a dependency...). Does anyone have any clue what's the difference between it and xquartz?

smorimoto:
I think https://github.com/ocaml/opam-repository/pull/18729 will fix the problem.

smorimoto:
The problem is that the environment variable is not exported, since at least after making changes to the discover everything worked fine: https://github.com/ocaml/graphics/pull/36

PR #18532, opened 2021-04-19
Add a post-message to Merlin regarding completion in Emacs

voodoos:
This PR adds the following post-install message to Merlin packages. 
Sadly there is no way to have this message only show when user upgrade from a specific version, so I guess it will be best not to include it in future release assuming that people will have done the update by then.

```
"Since version 4.2, Merlin integration with completion packages company
and auto-complete has moved to separate modules. Make sure to add
`(require 'merlin-company)` or `(require 'merlin-ac)` to your emacs
configuration if you were using one of these."
  {success}
  ```

@trefis, @let-def


kit-ty-kate:
Thanks a lot!

PR #18536, opened 2021-04-20
Dream — tidy, feature-complete Web framework

aantron:
Dream is a Web framework with a [simple programming model](https://aantron.github.io/dream/#types). It features...

- HTTP/1.1, [HTTP/2, and TLS](https://github.com/aantron/dream/tree/master/example/l-https#files),
- [WebSockets](https://github.com/aantron/dream/tree/master/example/k-websocket#files),
- [GraphQL](https://github.com/aantron/dream/tree/master/example/w-graphql-subscription#files), including subscriptions,
- [HTML templates](https://github.com/aantron/dream/tree/master/example/7-template#files),
- a composable [router](https://github.com/aantron/dream/tree/master/example/3-router#files),
- [sessions](https://github.com/aantron/dream/tree/master/example/b-session#files) with pluggable [back ends](https://aantron.github.io/dream/#back-ends),
- secure helpers for [cookies](https://github.com/aantron/dream/tree/master/example/c-cookie#files) and [forms](https://github.com/aantron/dream/tree/master/example/d-form#files), and
- internationalization-friendly [error handling](https://github.com/aantron/dream/tree/master/example/9-error#files).

Dream is presented as a single, flat module. It does not assume any boilerplate. Indeed, depending on the use case, a Dream app can be packaged as a [single executable](https://github.com/aantron/dream/tree/master/example/w-one-binary#files).

See

- [Project page](https://github.com/aantron/dream)
- [API documentation](https://aantron.github.io/dream/)
- [Example index](https://github.com/aantron/dream/tree/master/example#readme)

Included, in particular, are full-stack (server and client in ML) examples with...

- [js_of_ocaml](https://github.com/aantron/dream/tree/master/example/w-fullstack-jsoo#files)
- [ReScript](https://github.com/aantron/dream/tree/master/example/w-fullstack-rescript#files)
- [Melange](https://github.com/aantron/dream/tree/master/example/r-fullstack-melange#files)

### DIFF
--- a/packages/dream/dream.1.0.0~alpha1/opam
+++ b/packages/dream/dream.1.0.0~alpha1/opam
@@ -1,0 +1,128 @@
+opam-version: "2.0"
+
+synopsis: "Tidy, feature-complete Web framework"
+tags: ["http" "web" "framework" "websocket" "graphql" "server" "http2" "tls"]
+description: """
+Dream is a feature-complete Web framework with a simple programming
+model and no boilerplate. It provides only two data types, request and
+response.
+
+Almost everything else is either a built-in OCaml type, or an
+abbreviation for a bare function. For example, a Web app, known in
+Dream as a handler, is just an ordinary function from requests to
+responses. And a middleware is then just a function from handlers to
+handlers.
+
+Within this model, Dream adds:
+
+- Session management with pluggable back ends.
+- A fully composable router.
+- Support for HTTP/1.1, HTTP/2, and HTTPS.
+- WebSockets.
+- GraphQL, including subscriptions and a built-in GraphiQL editor.
+- SQL connection pool helpers.
+- Server-side HTML templates.
+- Automatic secure handling of cookies and forms.
+- Unified, internationalization-friendly error handling.
+- A neat log, and OCaml runtime configuration.
+- Helpers for Web formats, such as Base64url, and a modern cipher.
+
+Because of the simple programming model, everything is optional and
+composable. It is trivailly possible to strip Dream down to just a
+bare driver of the various HTTP protocols.
+
+Dream is presented as a single module, whose API is documented on one
+page. In addition, Dream comes with a large number of examples.
+Security topics are introduced throughout, wherever they are
+applicable."""
+
+license: "MIT"
+homepage: "https://github.com/aantron/dream"
+doc: "https://aantron.github.io/dream"
+bug-reports: "https://github.com/aantron/dream/issues"
+dev-repo: "git+https://github.com/aantron/dream.git"
+
+author: "Anton Bachin <antonbachin@yahoo.com>"
+maintainer: "Anton Bachin <antonbachin@yahoo.com>"
+
+depends: [
+  "base-unix"
+  "base64" {>= "3.1.0"}  # Base64.encode_string.
+  "bigarray-compat"
+  "caqti-lwt"
+  "conf-libev"
+  "cstruct"
+  "dune" {>= "2.7.0"}  # --instrument-with.
+  "fmt" {>= "0.8.7"}  # `Italic.
+  "graphql_parser"
+  "graphql-lwt"
+  "hmap"
+  "lwt"
+  "lwt_ppx"
+  "lwt_ssl"
+  "logs" {>= "0.5.0"}
+  "magic-mime"
+  "mirage-crypto" {>= "0.8.1"}  # AES-256-GCM.
+  "mirage-crypto-rng" {>= "0.8.0"}  # Signature of initialize.
+  "multipart-form-data" {>= "0.3.0"}
+  "ocaml" {>= "4.08.0"}
+  "opam-installer" {build}
+  "uri"
+  "yojson"  # ...
+
+  # Currently vendored.
+  # "gluten"
+  # "gluten-lwt-unix"
+  # "httpaf"
+  # "httpaf-lwt-unix"
+  # "h2"
+  # "h2-lwt-unix"
+  # "websocketaf"
+
+  # Dependencies of vendored packages.
+  "angstrom" {>= "0.14.0"}
+  "bigstringaf" {>= "0.4.0"}
+  "digestif" {>= "0.7"}  # websocket/af, sha1.
+  "faraday" {>= "0.6.1"}
+  "faraday-lwt-unix"
+  "psq"  # h2.
+  "result"  # http/af, websocket/af.
+
+  # Testing, development.
+  "alcotest" {with-test}
+  "bisect_ppx" {dev & >= "2.5.0"}  # --instrument-with.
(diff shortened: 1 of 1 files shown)


aantron:
CI failures are on older distributions:

```
(failed: ssl_stubs.c: 'TLS1_3_VERSION' undeclared (first use in this function); did you mean 'TLS1_2_VERSION'?)
```

I will look into this for future releases, but for alpha1, from Dream's point of view, it's fine not to support building on older distributions.

packages/dream/dream.1.0.0~alpha1/opam
```diff
@@ -0,0 +1,128 @@
...
+]
+
+build: [
+  ["dune" "build" "-p"
+    "dream,gluten,gluten-lwt,gluten-lwt-unix,websocketaf,httpaf,httpaf-lwt,httpaf-lwt-unix,hpack,h2,h2-lwt,h2-lwt-unix"
+    "-j" jobs]
+]
+
+install: [
+  ["opam-installer" "--prefix" prefix "dream.install"]
```
kit-ty-kate, line 111:
```suggestion
```
This should be unnecessary as opam already installs `<pkg>.install` by default (on top of the `install` section)

aantron: Thanks. I just tested removing this line with esy, and esy does not install `dream.install` by default, causing a dependency error later if the line is removed. This may be a bug/oversight, but it's non-obvious and I think it's better not to rely on the repository client interpreting it correctly. So, I prefer to keep this line.

Hopefully, this whole `install:` section will become unnecessary relatively soon, and I'll remove it completely in a future Dream release.

dinosaure:
@aantron the failure is due to: https://github.com/savonet/ocaml-ssl/issues/70

aantron:
@dinosaure Thanks, subscribed to that issue!

packages/dream/dream.1.0.0~alpha1/opam
mseri: One last question before merging. This may be a good opportunity to try how the vendoring behaves. Have you checked what happens if you install dream and then install the opam version of any of its vendored dependencies separately (or if you have some of the vendored dependencies preinstalled when you compile it)?

mseri: > Thanks. I just tested removing this line with esy, and esy does not install `dream.install` by default, causing a dependency error later if the line is removed.

Sorry, dumb question, have you tried removing the whole `install` section or just that line?

aantron: > One last question (from me) before merging. This may be a good opportunity to try how the vendoring behaves. Have you checked what happens if you install dream and then install the opam version of any of its vendored dependencies separately (or if you have some of the vendored dependencies preinstalled when you compile it)?

Yes, assuming you are asking about `opam` as the client. Because of the `conflicts:` section, opam shows prompts like this:

```
  ⊘ remove  dream  ~dev* [conflicts with httpaf]
  ∗ install httpaf 0.7.1
===== ∗ 1   ⊘ 1 =====
Do you want to continue? [Y/n]
```

> have you tried removing the whole `install` section or just that line?

I removed just that line during this PR. The section itself exists because depending on `dream` fails without it (when the vendored packages are missing), which I experienced when I first tried to install `dream` after vendoring these deps. I also tried variations of `dune install -p ...` (as suggested by @kit-ty-kate, thank you) and [`(vendored_dirs)`](https://dune.readthedocs.io/en/stable/dune-files.html?highlight=vendored_dirs#vendored-dirs-since-1-11). However, `dune install -p` did not seem to be able to actually find the vendored packages' `.install` files when used on its own. `(vendored_dirs)` disallows passing `-p` options for  selecting subpackages of vendored packages, which forces an overestimation of their deps that are needed for `dream`, since they are monorepos, pulling in e.g. `async` via `httpaf-async`.

So, AFAIK, `opam-installer` is the only thing that works right now (apart from hacky manual scripts). I've been implicitly testing with `opam-installer` for several weeks now.

aantron: > The section itself exists because depending on `dream` fails without it (when the vendored packages are missing), which I experienced when I first tried to install `dream` after vendoring these deps.

To clarify, building a package that depends *on* `dream` fails because of missing transitive dependencies like `httpaf-lwt-unix`.

aantron: > `(vendored_dirs)` disallows passing `-p` options for selecting subpackages of vendored packages, which forces an overestimation of their deps that are needed for `dream`, since they are monorepos, pulling in e.g. `async` via `httpaf-async`.

And to clarify this, the failure is during the build of `dream` itself, since it picks up false transitive dependencies like `async`, which are not installed (and shouldn't be, on a request for `dream` only). So I don't know if `dune install` together with `(vendored_dirs)` would work, because `(vendored_dirs)` forces a failure during build, before the install step.

anmonteiro: BTW, since you made most of the [Piaf](https://github.com/anmonteiro/piaf) dependency choices (my forks of http/af and websocket/af, lwt_ssl, etc), you could depend on Piaf and its vendored libs (`piaf.httpaf`, `piaf.h2`, etc).

Then you'd just need to vendor websocket/af and e.g. use a `subdir` stanza like [I do in Piaf](https://github.com/anmonteiro/piaf/blob/master/vendor/dune)

aantron: I may well do that eventually. However, I understand that then Dream will be "locked" to whatever exact commits Piaf is vendoring, is that right? At this early point, I still need the flexibility to choose other commits, but I hope to streamline that away soon.

Thanks for the `subdir` stanza tip.

mseri: Thanks for the explanations, indeed one reason I was asking was due the presence of those conflicts. Thanks for clarifying the issue. It is a pity that it does not install cleanly due to the vendoring

aantron: Do you mean in the context of a switch with those packages? It seems to install cleanly otherwise. The vendoring is a temporary situation. My goal is to eliminate it relatively quickly. I think it's fine for a first release &mdash; this isn't affecting any users of Dream, because it mostly doesn't have any. It will just be a temporary annoyance for early adopters.

mseri:
@aantron I guess it's time to announce the release :)

aantron:
@mseri Thanks!

I'm going to hold off announcing it *immediately*, because now, with the package in opam, I need to do some deployment testing, etc. I may end up making some ~alpha2 release based on any experience from that (hopefully not too soon!), so there's a good chance that's the one that will get an intentional announcement.

PR #18537, opened 2021-04-20
[new release] mimic (0.0.3)

dinosaure:
A simple protocol dispatcher

- Project page: <a href="https://github.com/dinosaure/mimic">https://github.com/dinosaure/mimic</a>
- Documentation: <a href="https://dinosaure.github.io/mimic/">https://dinosaure.github.io/mimic/</a>

##### CHANGES:

- Move the project to https://github.com/dinosaure/mimic (@dinosaure)
  Old distributions of `mimic` are still available on
  https://github.com/mirage/ocaml-git but `mimic` starts to be
  used by others projects than `ocaml-git`. We decided to make
  its own repository.
- Take the most recent value in the `ctx` instead of the older one
  **breaking changes**
  When `mimic` wants to instantiate a transmission protocol, if
  a value `'a Mimic.value` was inserted multiple times, `mimic`
  took the older one to instance the transmission protocol.

  Now, `mimic` takes the newer one. It useful when we want to
  implement the rediction in HTTP where we need to "replace" values
  by the new destination.

### DIFF
--- a/packages/mimic/mimic.0.0.3/opam
+++ b/packages/mimic/mimic.0.0.3/opam
@@ -1,0 +1,39 @@
+opam-version: "2.0"
+synopsis: "A simple protocol dispatcher"
+description: "A middleware to dispatch protocols"
+maintainer: ["romain.calascibetta@gmail.com"]
+authors: "Romain Calascibetta"
+license: "ISC"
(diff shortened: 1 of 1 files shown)


mseri:
Looks good, thanks

PR #18541, opened 2021-04-21
bytearray-v102: bug correction on 32 bits systems

UnixJunkie:


kit-ty-kate:
Thanks! Could you return the missing field upstream?

UnixJunkie:
Done, thanks!

PR #18542, opened 2021-04-21
Tezos 9.0

pirbo:
This is a new major release of the tezos/tezos codebase. I'll announce it on OCaml discuss at the time of announcements on all medium.

### DIFF
--- a/packages/tezos-protocol-compiler/tezos-protocol-compiler.9.0/opam
+++ b/packages/tezos-protocol-compiler/tezos-protocol-compiler.9.0/opam
@@ -1,0 +1,28 @@
+opam-version: "2.0"
+maintainer: "contact@tezos.com"
+authors: [ "Tezos devteam" ]
+homepage: "https://www.tezos.com/"
+bug-reports: "https://gitlab.com/tezos/tezos/issues"
+dev-repo: "git+https://gitlab.com/tezos/tezos.git"
+license: "MIT"
+depends: [
+  "ocaml" { >= "4.10.2" & < "4.11" }
+  "dune" { >= "2.0" }
+  "tezos-protocol-environment" { = version }
+  "ocp-ocamlres" { >= "0.4" }
+]
+build: [
+  ["rm" "-r" "vendors"]
+  ["dune" "build" "-p" name "-j" jobs]
+  ["mv" "src/lib_protocol_compiler/%{name}%.install" "./"]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+]
+synopsis: "Tezos: economic-protocol compiler"
+
+url {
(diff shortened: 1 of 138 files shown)


packages/tezos-protocol-compiler/tezos-protocol-compiler.9.0/opam
```diff
@@ -0,0 +1,28 @@
+opam-version: "2.0"
+maintainer: "contact@tezos.com"
+authors: [ "Tezos devteam" ]
+homepage: "https://www.tezos.com/"
+bug-reports: "https://gitlab.com/tezos/tezos/issues"
+dev-repo: "git+https://gitlab.com/tezos/tezos.git"
+license: "MIT"
+depends: [
+  "ocaml" { >= "4.10.2" & < "4.11" }
```
kit-ty-kate, line 9: is there a particular reason why tezos has been stuck at OCaml 4.10 for a while?

pirbo: Sorry long answer :-)

The tezos blockchain has the feature to self amend. This means that the community votes upgrades that are automatically deployed if approved. Technically, that means that when an upgrade passes, the new code of the "economic protocol" is downloaded through the peer2peer network, compiled thanks to the OCaml compiler embedded in  tezos-node, dynamically loaded and executed to validate future blocks...
This fancy mechanism works only if the requirement to compile "economic protocols" stay stable over time. So tezos is not likely to use the cutting edges version of the compiler (Changing these per-requirement and therefore requiring an explicit manual upgrades from all participants of the network has a name in tezos terminology: "changing the protocol environment", we do that but not too often, with care and it has a cost). 

Now, more practically, in the present case, why did we bump to 4.10 during our last "environment upgrade": 4.11 were not an option as "it is not available for Mac M1" (I repeat what I've been told). Changes in the signatures of the stdlib modules made using 4.12 non trivial. (Ah yes, I forgot to mention: To ensure that everybody uses the same rule, a released economic protocol is characterize by the hash of its .ml. So, no way to touch them at all after the fact! Now, as everybody has to be able to resync the chain from scratch re-validating all the block, you have to build compatibility layers around "old protocols" so that their very exact code from way back when you wrote them compiles with the stack you are using today!) 4.12 were also incompatible with being stuck with old version of dependencies (ocamlformat and ocaml-tls if I remember correctly) for "stupid" reasons.
Anyway, we are nearly done removing all the barrier and, as the next "protocol upgrade" will require a "protocol environment bump" to add extra crypto primitives, tezos version 10 will probably use OCaml 4.12 :-)

kit-ty-kate:
Thanks!

PR #18559, opened 2021-04-22
adds FrontC.3.4.3

ivg:
This is a small update that adds support for pure-asm bodies and provies better diagnostic messages in case of parsing errors. It also cleans a bit the installation procedure (moves META to the upstream repository and applies the opam.patch). It also builds and installs the cmxs file (and updates, correpsondingly, the META file). 

Changelog:

* BinaryAnalysisPlatform/FrontC#7 raises parse errors rather than asserting false
* BinaryAnalysisPlatform/FrontC#11 supports functions with pure-asm bodies
* BinaryAnalysisPlatform/FrontC#13 builds and installs cmxs

### DIFF
--- a/packages/FrontC/FrontC.3.4.3/opam
+++ b/packages/FrontC/FrontC.3.4.3/opam
@@ -1,0 +1,31 @@
+opam-version: "2.0"
+authors: "Hugues Cassé <casse@irit.fr>"
+homepage: "https://github.com/BinaryAnalysisPlatform/FrontC"
+bug-reports: "https://github.com/BinaryAnalysisPlatform/FrontC/issues"
+dev-repo: "git+https://github.com/BinaryAnalysisPlatform/FrontC"
+maintainer: "Ivan Gotovchits <ivg@ieee.org>"
+depends: [
+  "ocaml"
+]
+install: [
+  [make "install" "PREFIX=%{lib}%" "OCAML_SITE=%{lib}%"]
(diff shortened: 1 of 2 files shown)


packages/FrontC/FrontC.3.4.3/opam
```diff
@@ -0,0 +1,32 @@
...
+homepage: "https://github.com/BinaryAnalysisPlatform/FrontC"
+bug-reports: "https://github.com/BinaryAnalysisPlatform/FrontC/issues"
+dev-repo: "git+https://github.com/BinaryAnalysisPlatform/FrontC"
+maintainer: "Ivan Gotovchits <ivg@ieee.org>"
+license: "LGPL-2.1-only"
+depends: [
+  "ocaml"
+]
+install: [
+  [make "install" "PREFIX=%{lib}%" "OCAML_SITE=%{lib}%"]
```
kit-ty-kate, line 12: I'm not seeing any `build` section in this package. Are there nothing built? If so, please use it, putting everything in the `install` section will slow down concurrent installation for your users'

ivg: good point! ALso, is there any sense in `FrontC.install` file? I have inherited it from the previous maintainers of FrontC and have no idea what it is meant for. Any guesses?

ivg:
@kit-ty-kate, the reporter has acted :)

kit-ty-kate:
Sorry I missed it. Thanks!

PR #18564, opened 2021-04-23
z3: fixes the installation of dynamic libraries

ivg:
@c-cube accidentaly reverted (in #17558) the installation command back to what was used in 4.8.8 instead of using one from 4.8.8-1, so the dll files are again not installed, which prevents us from packing z3 into a plugin (cmxs) or loading it in a toplevel (ocaml or utop).

It also restored the installation of the z3 binary, which is also extremely useful.

### DIFF
--- a/packages/z3/z3.4.8.9-1/opam
+++ b/packages/z3/z3.4.8.9-1/opam
@@ -9,9 +9,12 @@ build: [
   [ "python2.7" "scripts/mk_make.py" "--ml" "--staticlib" ]
   [ make "-C" "build" "-j" jobs ]
 ]
+
 install: [
-  [ "sh" "-c" "ocamlfind install z3 build/api/ml/META -nodll build/libz3* build/api/ml/*" ]
+  [ "sh" "-c" "ocamlfind install z3 build/api/ml/* build/libz3-static.a"]
+  [ "cp" "build/z3" "%{bin}%/z3"]
 ]
+
 depends: [
   "ocaml"
   "ocamlfind" {build}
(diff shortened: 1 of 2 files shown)


mseri:
Thanks!

mseri:
Could you move this change to 0.4.9-1? I will keep an eye out for this in the next release

ivg:
> Could you move this change to 0.4.9-1? I will keep an eye out for this in the next release

Then I will have a problem, as I need to specify that bap works with 4.8.8-1 and 4.8.9-1, it is doable, (I can use conflicts) but confusing. Also, the problem is very subtle, in fact, it is only triggered by toplevels as normal linking and dynamic linking works fine.

Basically, my point is that it is an error in the installation script that should be fixed. But I can imagine that this point could be debunked. There is no rush, right now, let's wait until Monday and have some discussion when @c-cube will come and all other interested parties.

avsm:
Please bump the version and just mark this existing one as unavailable in the constraints so it won't get selected (so you don't need constraints).  The discipline of treating versions as immutable is important as it will be enforced eventually, so we need to get used to doing that.  Note that the amount of _time_ between an error entering the repository is immaterial, since downstream users of opam-repo use git commits -- and we don't want anyone observing the same version of a package materially changing its contents when they advance back and forth on the opam-repo commit history.

ivg:
>  The discipline of treating versions as immutable is important as it will be enforced eventually, so we need to get used to doing that. 

Okay, let's stick to this rule then. I am all for it. 

> mark this existing one as unavailable in the constraints

Do you mean to mark it in the current 4.8.9 opam, right?

mseri:
> Do you mean to mark it in the current 4.8.9 opam, right?

Yes, correct! Thanks

c-cube:
Sorry for the unintensional regression. I have no particular opinion on the new patch :)

mseri:
Thanks!

PR #18565, opened 2021-04-23
Package prbnmcn-mcts.0.0.1

igarnier:
### `prbnmcn-mcts.0.0.1`
Monte-Carlo tree search based on UCB1 bandits
Monte-Carlo tree search based on UCB1 bandits

---
* Homepage: http://github.com/igarnier/prbnmcn-mcts
* Source repo: git+https://gitlab.com/igarnier/monorepo.git
* Bug tracker: http://github.com/igarnier/prbnmcn-mcts

---

### DIFF
--- a/packages/prbnmcn-mcts/prbnmcn-mcts.0.0.1/opam
+++ b/packages/prbnmcn-mcts/prbnmcn-mcts.0.0.1/opam
@@ -1,0 +1,36 @@
+opam-version: "2.0"
+synopsis: "Monte-Carlo tree search based on UCB1 bandits"
+description: "Monte-Carlo tree search based on UCB1 bandits"
+maintainer: ["igarnier@protonmail.com"]
+authors: ["Ilias Garnier"]
+license: "MIT"
+homepage: "http://github.com/igarnier/prbnmcn-mcts"
+bug-reports: "http://github.com/igarnier/prbnmcn-mcts"
+depends: [
+  "ocaml" {>= "4.08.0"}
+  "dune" {>= "2.8"}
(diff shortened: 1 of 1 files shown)


igarnier:
This package builds a basic Monte-Carlo tree search library on top of UCB1 bandits. A version constraint was missing on the OCaml compiler version, which I force-pushed.

packages/prbnmcn-mcts/prbnmcn-mcts.0.0.1/opam
```diff
@@ -0,0 +1,36 @@
...
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
+dev-repo: "git+https://gitlab.com/igarnier/monorepo.git"
```
mseri, line 29: Is this line correct or we should replace it with `https://github.com/igarnier/prbnmcn-mcts`?

igarnier: Hi! I asked myself the same question. This is where development of this (and other) package(s) happen. Is there another intended semantics for that field?

igarnier: I now see that this is the field `opam` uses to interpret the `--dev-repo` option when pinning: indeed, the particular value set for `dev-reo` in this PR will make this command fail. (Note that I used the same value in the recently merged `prbnmcn-ucb1` but this can maybe be fixed in a separate PR).

I'll push a fix.

kit-ty-kate:
Thanks!

PR #18572, opened 2021-04-25
[new release] parmap (1.2.1)

rdicosmo:
Minimalistic library allowing to exploit multicore architecture

- Project page: <a href="https://github.com/rdicosmo/parmap">https://github.com/rdicosmo/parmap</a>

##### Fix issue with 32bits architectures, bump up release to 1.2.1

### DIFF
--- a/packages/parmap/parmap.1.2.1/opam
+++ b/packages/parmap/parmap.1.2.1/opam
@@ -1,0 +1,38 @@
+opam-version: "2.0"
+build: [
+  ["dune" "subst"] {dev}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+  ["dune" "build" "-p" name "@doc"] {with-doc}
+]
+maintainer: ["Roberto Di Cosmo <roberto@dicosmo.org>"]
+authors: ["Roberto Di Cosmo <roberto@dicosmo.org>"]
+bug-reports: "https://github.com/rdicosmo/parmap/issues"
+homepage: "https://github.com/rdicosmo/parmap"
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
The linter failed on this package:
```
2021-04-25 15:45.46: Error in parmap.1.2.1: Your dune-project file indicates that this package requires at least dune 1.10 but the dependencies only requires dune >= 1.0.
```
Could you fix this?

rdicosmo:
> The linter failed on this package:
> 
> ```
> 2021-04-25 15:45.46: Error in parmap.1.2.1: Your dune-project file indicates that this package requires at least dune 1.10 but the dependencies only requires dune >= 1.0.
> ```
> 
> Could you fix this?

Well, I looked into that, and it seems the linter is wrong: build fails for the package if we only require dune >= 1.0

kit-ty-kate:
> Well, I looked into that, and it seems the linter is wrong: build fails for the package if we only require dune >= 1.0

well, precisely. The package already requires >= 1.0 but needs >= 1.10.

If you think the message is too confusing, could you help us by providing a message that's less confusing?

rdicosmo:
> well, precisely. The package already requires >= 1.0 but needs >= 1.10.

Ok, I see, my bad: I had read the message as a suggestion to change the requirement in dune-project, instead of in the dependencies, thanks for clarifying and fixing this!

> If you think the message is too confusing, could you help us by providing a message that's less confusing?

Maybe something like the following (if it is not too verbose)?

Your dune-project file indicates that this package requires at least dune 1.10 but the dependencies only requires dune >= 1.0: please check which requirement is the right one, and fix the other.

> edit: oops, sorry i edited your comment by mistake, i tried to revert it quicky
No problem, and thank you for your precious work!

PR #18575, opened 2021-04-25
Package lens.1.2.5: Migration to ppxlib

pdonadeo:


packages/lens/lens.1.2.5/opam
```diff
@@ -0,0 +1,32 @@
...
+bug-reports: "https://github.com/pdonadeo/ocaml-lens/issues"
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+]
+depends: [
+  "ocaml"           {>= "4.10"}
+  "ppx_deriving"    {< "5.0"}
+  "ppx_tools"       {build}
+  "ppxfind"         {build}
```
kit-ty-kate, line 16: Are you sure this is still the case? It seems to conflict with your changelog

pdonadeo: Of course you are right, I forgot to update the opam file. I'm going to submit again. @kit-ty-kate

kit-ty-kate:
@pdonadeo could you return the fix in the opam upstream? I'm not sure `ppxlib` should be tagged as `{build}` because i don't know if it has a runtime (or if not, will always have none).
Also constraints were missing. The switch to ppxlib occurred in ppx_deriving 5.0 and you're using features from ppxlib 0.14 so i added the missing constraints.

kit-ty-kate:
Also, I'm gonna put my ppx_deriving maintainer hat on for a second: the ppx_deriving API is deprecated and if you have time to switch it would be better to use `ppxlib` directly. See https://github.com/ocaml-ppx/ppx_deriving/pull/250
If you don't have the time to do this, don't worry the API will still be there for a reasonable amount of time (until it gets prohibitively expensive to maintain or something like that)

kit-ty-kate:
Good to go now. Thanks!

pdonadeo:
Thanks to you @kit-ty-kate for your job 😃

PR #18580, opened 2021-04-26
Fix dependency about git.3.4.0 (it requires at least carton.0.4.1)

dinosaure:
A simple fix about `git.3.4.0` and `carton.0.4.1` spotted by @samoht.


mseri:
Thanks

PR #18583, opened 2021-04-26
[NEW] New package ocaml-solidity.0.1.0

lefessan:


kit-ty-kate:
Given these are new packages, are you sure you want to name them `ocam-solidity-*` and not just `solidity-*` ? We usually try to avoid this for new packages for obvious reasons

packages/ocaml-solidity-alcotest/ocaml-solidity-alcotest.0.1.0/opam
```diff
@@ -0,0 +1,56 @@
...
+  ]
+  ["sh" "-c" "./scripts/after.sh build '%{name}%'"]
+]
+install: [
+  ["sh" "-c" "./scripts/before.sh install '%{name}%'"]
+]
+depends: [
+  "ocaml" {>= "4.08.0"}
+  "dune" {>= "2.6.0"}
+  "ocaml-solidity-parser" {}
```
avsm, line 43: do you want these batched packages to be `= version` so that they are all installed together? Since these are new packages, it won't matter now, but it will begin to as you issue new versions.

packages/ez_hash/ez_hash.0.1.0/opam
```diff
@@ -0,0 +1,50 @@
...
+"""
+authors: ["Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>"]
+maintainer: [
+  "David Declerck <david.declerck@ocamlpro.com>"
+  "Steven De Oliveira <steven.de-oliveira@ocamlpro.com>"
+]
+homepage: "https://github.com/OcamlPro/ocaml-solidity"
+doc: "https://OCamlPro.github.io/ocaml-solidity/sphinx"
+bug-reports: "https://github.com/OCamlPro/ocaml-solidity/issues"
+dev-repo: "git+git+https://github.com/OcamlPro/ocaml-solidity.git"
```
hannesm, line 16: this looks suspicious (curious that the linter does not complain about `git+git+https`)

lefessan:
Thanks for the feedback, I think all these issues have been fixed in this version.

lefessan:
It looks like the only problem is with OCaml 4.13... and the error is triggered by dune. Probably not up to me ?

kit-ty-kate:
Indeed, this is due to https://github.com/ocaml/ocaml/pull/10277 which broke dune and a number of other packages. We're still trying to figure things out with the compiler devs. OCaml 4.13 is still very much in an unstable phase so this can be ignored.

Thanks!

Issue #18586, opened 2021-04-26
Failed compilation of ocaml-secondary-compiler during utop installation in new machine at ocaml-base-compiler.4.07.0


yonoteam:
Hello, my computer suggested filing a bug report and I found nothing related in previous issues. Below I enumerate the process that generated the error message appearing below. Any help/explanation is welcomed.

Using `opam config report` after step 14 below, outputs the following message.
```
    # opam config report
    # opam-version      2.0.8 
    # self-upgrade      no
    # system            arch=arm64 os=macos os-distribution=homebrew os-version=11.2.3
    # solver            builtin-mccs+glpk
    # install-criteria  -removed,-count[version-lag,request],-count[version-lag,changed],-changed
    # upgrade-criteria  -removed,-count[version-lag,solution],-new
    # jobs              7
    # repositories      1 (http) (default repo at 5e01cfa3)
    # pinned            0
    # current-switch    ocaml-base-compiler.4.07.0
```
 
Process:
0. `opam switch list` reports I am using `ocaml-base-compiler.4.07.0`
1. did `eval $(opam env)`
2. did `opam init`
3. did `opam install zarith`
4. tried `opam install graphics` but failed, reported missing xorgproto and xquartz
5. did `opam install ocamlbuild`
6. did `opam install ocamlfind`
7. retried `opam install graphics` but failed again for same reason
8. brew installed xorgproto and xquartz
9. retried `opam install graphics` and worked
10. did `opam install zarith`
11. did `opam install utop` but error message appeared suggesting filing a bug report here, trying installing again sequentially or restore to former state with suggested command.
12. restored to former state with suggested command
13. retried `opam install utop`  but failed again suggesting same fixes
14. tried `opam install --jobs=1 utop` but failed with following message appearing so decided to file bug-report:
```
    The following actions will be performed:
      - install ocaml-secondary-compiler 4.08.1-1 [required by ocamlfind-secondary]
      - install ocamlfind-secondary      1.8.1    [required by dune]
      - install dune                     2.8.5    [required by utop]
      - install trie                     1.0.0    [required by mew]
      - install result                   1.5      [required by lwt]
      - install ocaml-syntax-shims       1.0.0    [required by lwt]
      - install mmap                     1.1.0    [required by lwt]
      - install csexp                    1.5.1    [required by dune-configurator]
      - install cppo                     1.6.7    [required by utop]
      - install camomile                 1.0.2    [required by utop]
[...]

PR #18587, opened 2021-04-26
Forbid ocaml-secondary-compiler on macOS/arm64

kit-ty-kate:
Noticed in https://github.com/ocaml/opam-repository/issues/18586


avsm:
lgtm

PR #18588, opened 2021-04-26
[new release] jose (0.6.0)

ulrikstrid:
CHANGES:

- JWT/JWS/JWK: Add support for ES256 and ES512 signing via the updated mirage-crypto and x509 (by @ulrikstrid)
- JWT: [BREAKING] JWT will not validate `exp` by default anymore (by @ulrikstrid)


kit-ty-kate:
Thanks! Could you return the fix in the opam file upstream (most probably the dune-project file) ?

ulrikstrid:
My current setup is broken with dune-release so that's why I had to change it. This will be fixed next time since I'm getting a new computer and will make sure it works.

PR #18598, opened 2021-04-27
[new release] irmin et al. (2.6.0)

CraigFe:
### Fixed

- **irmin**
  - Fix stack overflow exception when working with wide trees (mirage/irmin#1313, @zshipko)

  - `Tree.of_concrete` now prunes empty subdirectories, and raises
    `Invalid_argument` if the input contains duplicate bindings. (mirage/irmin#1385,
    @CraigFe)

- **irmin-chunk**
  - Use the pre_hash function to compute entry keys instead of
    their raw binary representation (mirage/irmin#1308, @samoht)

### Changed

- **irmin-git**
  - Upgrade `irmin-git` with `git.3.4.0`. (mirage/irmin#1392, @dinosaure)

### DIFF
--- a/packages/irmin-bench/irmin-bench.2.6.0/opam
+++ b/packages/irmin-bench/irmin-bench.2.6.0/opam
@@ -1,0 +1,49 @@
+opam-version: "2.0"
+maintainer:   "thomas@gazagnaire.org"
+authors:      ["Thomas Gazagnaire"]
+license:      "ISC"
+homepage:     "https://github.com/mirage/irmin"
+bug-reports:  "https://github.com/mirage/irmin/issues"
+dev-repo:     "git+https://github.com/mirage/irmin.git"
+doc:          "https://mirage.github.io/irmin/"
+
+build: [
+ ["dune" "subst"] {dev}
+ ["dune" "build" "-p" name "-j" jobs]
+]
+
+depends: [
+  "dune"         {>= "2.7.0"}
+  "irmin-pack"   {= version}
+  "irmin-layers" {= version}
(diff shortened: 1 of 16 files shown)


dinosaure:
Hmmhmm it's strange that we did not notice the error about `irmin-git` 😕 .

hannesm:
hm, there are test failures (irmin-git) on 32 bit platforms. the upstream CI (https://ci.ocamllabs.io/github/mirage/irmin/commit/5066df4b9e56d74b66def5b98300470c216db193) did not run on 32 bit because some irmin packages have the arch <> 32 bit constraint.

maybe all irmin packages should be made unavailable on 32 bit? or the bug being found and fixed? I'm interested in this PR being merged (for some 64 bit unikernels)...

dinosaure:
> maybe all irmin packages should be made unavailable on 32 bit? or the bug being found and fixed? I'm interested in this PR being merged (for some 64 bit unikernels)...

With a discussion with @CraigFe, _a priori_, `irmin-git` should not work on 32bits - however, I should try to find why on my side. I think this PR should be merge and I will make a specific issue about the support of `irmin-git` on 32bits arch.

CraigFe:
To be clear, it's not that I don't think `irmin-git` should support 32-bit architectures: it's just difficult to ensure that it does when we don't get 32-bit CI due to incompatibilities in unrelated packages in the same repository, as @hannesm points out.

I'm happy to set incompatibility with 32-bits for now and make a patch release at 2.6.1 when the 32-bit issue is fixed.

kit-ty-kate:
Thanks!

PR #18599, opened 2021-04-28
new file:   packages/oranger/oranger.2.0.3/

UnixJunkie:
compatible with latest dolog lib


packages/oranger/oranger.2.0.3/opam
```diff
@@ -0,0 +1,39 @@
...
+  [conf-cmake:cmd "-S" "ranger-0.9.11/cpp_version" "-B" "ranger-build"]
+  [make "-C" "ranger-build"]
+]
+install: [
+  ["cp" "ranger-build/ranger" "%{bin}%/ml_rf_ranger"]
+]
+depends: [
+  "dune" {>= "1.0.0"}
+  "conf-cmake" {build}
+  "conf-wget" {build}
```
kit-ty-kate, line 20:
```suggestion
  "conf-wget" {dev}
```
Is this only needed in the dev repository?

UnixJunkie: I don't think, because the opam file mentions a script called install_ranger.sh.
This script dowloads the ranger tarball in case it is not already here.
So, each new install / build of oranger needs wget.

avsm: That's going to run afoul of opam sandboxing.  You should specify any additional sources in `extra-sources` (with checksums) rather than fetching them from within the opam install script.

UnixJunkie:
Does this mean I should put the tarball I need under files/ ?
I can certainly do that and I guess it makes the package easier / more
robust to install.

packages/oranger/oranger.2.0.3/opam
kit-ty-kate: > I don't think, because the opam file mentions a script called install_ranger.sh.
> This script dowloads the ranger tarball in case it is not already here.
> So, each new install / build of oranger needs wget.

That's not true, opam always get the tarball and will fail anyway if it cannot be downloaded, before it can reach the build phase.
I've made the necessary adjustments anyway so the script is not required

kit-ty-kate:
> Does this mean I should put the tarball I need under files/ ? I can certainly do that and I guess it makes the package easier / more robust to install.

No, we cannot take tarball ourselves for licensing reasons. However if it can be vendored in your source repository that would be a good solution.

UnixJunkie:
Ok, I will look at vendoring it with my sources.

UnixJunkie:
On a second thought, including the ranger tarball is not a good idea, because this would force me to change the license
of oranger.
So, I think this current version can be merged.

packages/oranger/oranger.2.0.3/opam
UnixJunkie: conf-wget is not required anymore if opam can download the tarball

UnixJunkie:
I will send corrections to the opam file; locally the current one does not work for me.

UnixJunkie:
superseded by https://github.com/ocaml/opam-repository/pull/18639

UnixJunkie:
Sorry, I use git like a caveman, I know.

PR #18611, opened 2021-04-30
[new release] dolmen_type, dolmen_bin, dolmen_loop, dolmen_lsp and dolmen (0.6)

Gbury:

- Documentation: <a href="http://gbury.github.io/dolmen">http://gbury.github.io/dolmen</a>

### Release

- The official github release now provides access to
  already built binaries for `dolmen` and `dolmenls`,
  for linux (ubuntu) and macos
- The LSP server has been updated to depend on
  `linol~0.2`

### Bugfixes

- Smtlib2 let-bindings were treated as sequential, but are
  now treated as parrallel as specified by the spec;
  i.e. the following is now correctly rejected:
  `(let (x 0) (y x) (...))`

### Features

- Added support for higher order, including tptp's THF and Zf
- Optimized some corner cases of the typechecker to avoid
  exponential blowups

### API

- The interface of the `Expr` module has changed to support
  higher-order
- Additionally, there is now proper support for type aliases
  (which are expanded on demand as necessary), in `Expr`
- There is now a new typechecker module exposed as Thf for
  typing higher order expressions

### DIFF
--- a/packages/dolmen/dolmen.0.6/opam
+++ b/packages/dolmen/dolmen.0.6/opam
@@ -3,12 +3,13 @@ maintainer: "Guillaume Bury <guillaume.bury@gmail.com>"
 authors: "Guillaume Bury <guillaume.bury@gmail.com>"
 license: "BSD-2-Clause"
 build: [
-  ["dune" "subst"] {pinned}
+  ["dune" "subst"] {dev}
   ["dune" "build" "-p" name "-j" jobs]
 ]
 depends: [
   "ocaml" {>= "4.08"}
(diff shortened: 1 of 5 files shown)


Gbury:
As far as I can tell, the CI failures are because `dune` fails to compile on `4.13`.

kit-ty-kate:
Thanks! Could you return the fix to the opam files upstream?

Gbury:
Thanks a lot !
The fix to the opam files is merged upstream (in https://github.com/Gbury/dolmen/commit/5ca14273a4a4d54f2c8e073daaaa2ed16a14b13c ).

PR #18618, opened 2021-05-01
[new release] ocaml-lsp-server, lsp and jsonrpc (1.6.0)

rgrinberg:
LSP Server for OCaml

## Features

- Code action to annotate a value with its type (ocaml/ocaml-lsp#397)

## Fixes

- Fix interface/implementation switching on Windows (ocaml/ocaml-lsp#427)

- Correctly parse project paths with spaces and other special characters that
  must be escaped.

- Print types with `-short-paths` even if the project wasn't built yet


kit-ty-kate:
Thanks!

PR #18619, opened 2021-05-01
[new release] elpi (1.13.2)

gares:
CHANGES:

- Build:
  - drop ppxfindcache
  - relax dependency on ocaml-migrate-parsetree


gares:
It's weird that my CI works on 4.04 while here it seems there is a problem with ppx_deriving.
Other than that, it looks good to me.

kit-ty-kate:
thanks! Could you return the constraint fix to the opam file upstream?

gares:
done thanks!

PR #18626, opened 2021-05-03
FrontC: fixes building cmxs files and installs cmx files

ivg:
This PR disables the recently released FrontC 3.4.3 (#18559), as it
was installing incorrectly linked cmxs files that were missing code.
In addition it enables the installation of cmx files.

Since no code was changed but Makefiles only, I decided to pack this
is a patch rather then making a new release. The corresponding patch
is merged into FrontC so in the next release there would not be a need
for such patch.


ivg:
why is camelus reports errors on 3.4.3? (and what are the errors?)

kit-ty-kate:
Looks good. Thanks! The "errors" were simply the removed packages as it is not standard to remove them.

PR #18628, opened 2021-05-03
Tezos 9.1

pirbo:
This is a minor release of the tezos codebase fixing issues about the "Florence economic protocol" which will go live on the main network on May 11th (as well as imposing a version of irmin that does not suffer from known inefficiency).

I know I didn't publish on Ocaml discuss the previous major release because I was busy dealing with problems raised by 9.0 and finally fixed here but I'll do :-)


kit-ty-kate:
Thanks

PR #18632, opened 2021-05-04
releases FrontC 4.0.0

ivg:
This release doesn't really introduce any new features but it switches
FrontC to menhir and dune. The switch to menir was required because
the parser has a circular dependency on the lexer and only menhir is
capable of refactoring tokens from the parser. The circular
dependency, on its own, prevented us from upgrading FrontC to dune,
which in turn prevented us from using FrontC in duniverse.

Switching to menhir required some number of tweaks to the grammar,
which might change the behavior of the parser, hence the new
version. We will still support the old version of FrontC (aka stable)
and release bug fixes, but only the 4.x version will receive new
feautures, such as support for the more modern C syntax.

We also added tests and split the package into three subpackages.
The FrontC package is the library with the same contents as
before. And there are also two packages with executables:

- calipso - a program analysis tool that reduces non-structural
  control flow in C programs;

- ctoxml - a program that will dump the parser C AST in XML.

The `printc` executable is no longer installed as its main purpose is
testing and doesn't provide any useful features for the end users, it
just works as `cat`.

### DIFF
--- a/packages/FrontC/FrontC.4.0.0/opam
+++ b/packages/FrontC/FrontC.4.0.0/opam
@@ -1,0 +1,37 @@
+opam-version: "2.0"
+version: "4.0.0"
+synopsis: "Parses C programs to an abstract syntax tree"
+description:
+  "FrontC provides a C parser and an OCaml definition of an abstract syntax treee for the C language. It also includes AST pretty-printers in plain and XML formats."
+maintainer: ["Ivan Gotovchits <ivg@ieee.org>"]
+authors: ["Hugues Cassé <casse@irit.fr> et al"]
+license: "LGPL-2.0-only"
+tags: ["FrontC" "C" "parser" "XML"]
+homepage: "https://github.com/BinaryAnalysisPlatform/FrontC"
+bug-reports: "https://github.com/BinaryAnalysisPlatform/FrontC/issues"
+depends: [
+  "dune" {>= "2.7" & >= "2.7" & build}
+  "menhir"
+  "odoc" {with-doc}
+]
+build: [
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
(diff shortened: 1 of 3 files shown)


ivg:
it looks like that all recent PRs are failing on 4.13 when they try bootstrapping dune with
````
# File "vendor/cmdliner/src/cmdliner.ml", line 195, characters 25-47:
# Warning 6 [labels-omitted]: label errs was omitted in the application of this function.
# cd _boot && /home/opam/.opam/4.13/bin/ocamlopt.opt -c -g -no-alias-deps -w -49 -I +threads memo.ml
# File "src/memo/memo.ml", line 1002, characters 73-74:
# Error: This expression has type 'a but an expression was expected of type i 
#        This instance of i is ambiguous:
#        it would escape the scope of its equation
```

ivg (commented):
Thanks, @kit-ty-kate! But can you please explain why is dune not a build dependency? What am I missing here?

Concerning the version field it is actually coming from the `dune-project` file, which in turn, requires this field.

ivg:
Hmm.. I am experiencing some issues downstream, so I would close this issue until they are resolved. Will be back soon :)

ivg:
OK, I am opening it again, false alarm. Was confused with various versions (I wish at some time we will switch to duniverse)

ivg:
Any updates on the `This instance of i is ambiguous:` error, I see it strikes all recent PR? 
Also, any plans to merge us?

kit-ty-kate:
> Any updates on the `This instance of i is ambiguous:` error, I see it strikes all recent PR?

This is an error in dune with OCaml 4.13 (currently trunk). It has been fixed in OCaml, and it should stop showing up in a few days

> Thanks, @kit-ty-kate! But can you please explain why is dune not a build dependency? What am I missing here?

This is an issue in dune which breaks the whole opam switch in case of a downgrade if everything is not reinstalled. See https://github.com/ocaml/dune/issues/2147 for more details.

Sorry it took so long to merge. Thanks!

PR #18638, opened 2021-05-05
[new release] raygui and raylib (0.3)

tjammer:
CHANGES:

* Update bindings to raylib 3.5
* Add WIP raygui bindings
* Add bindings for working with models
* Fix / Make more conveniant 3D and rlgl part
* Update dune and ocamlformat to recent versions


packages/raylib/raylib.0.3/opam
```diff
@@ -0,0 +1,54 @@
...
+dev-repo: "git+https://github.com/tjammer/raylib-ocaml.git"
+depexts: [
+  ["mesa-common-dev" "libx11-dev" "libxrandr-dev" "libxi-dev" "xorg-dev" "libgl1-mesa-dev" "libglu1-mesa-dev"] {os-family = "debian"}
+  ["mesa-libGL-devel" "libX11-devel" "libXrandr-devel" "libXi-devel" "libXcursor-devel" "libXinerama-devel"] {os-family = "fedora"}
+  ["mesa-libGL-devel" "libX11-devel" "libXrandr-devel" "libXi-devel" "libXcursor-devel" "libXinerama-devel"] {os-family = "rhel"}
+  ["mesa-dev" "libx11-dev" "libxrandr-dev" "libxi-dev" "libxcursor-dev" "libxinerama-dev"] {os-distribution = "alpine"}
+  ["Mesa-libGL-devel" "libX11-devel" "libXrandr-devel" "libXi-devel" "libXcursor-devel" "libXinerama-devel"] {os-family = "suse"}
+  ["system:libX11-devel" "system:libXrandr-devel" "system:libXi-devel" "system:libXcursor-devel" "system:libXinerama-devel"] {os-distribution = "cygwinports"}
+  ["mesa-libs" "libX11" "libXrandr" "libXi" "libXcursor" "libXinerama"] {os-distribution = "freebsd"}
+  ["libxcursor" "libxrandr" "libxinerama" "libxi" "mesa"] {os-family = "arch"}
```
kit-ty-kate, line 39: is there any chance you could split these dependencies into their own `conf-*` packages (see `conf-libX11` for example) ? I know this is a lot but it would help you and us maintain this package and others in the long term

packages/raylib/raylib.0.3/opam
```diff
@@ -0,0 +1,54 @@
...
+]
+available: [arch != "arm32" & arch != "ppc64"]
+x-ci-accept-failures: [
+  "centos-7"
+  "oraclelinux-7"
+]
```
kit-ty-kate, line 45: what is the reason why these distributions are not supported?

packages/raylib/raylib.0.3/opam
tjammer: Yes, this sounds reasonable, will do.

packages/raylib/raylib.0.3.1/opam
```diff
@@ -0,0 +1,49 @@
...
+depends: [
+  "dune" {>= "2.8"}
+  "dune-configurator"
+  "ctypes" {>= "0.14"}
+  "ppx_cstubs" {>= "0.6"}
+  "conf-mesa" {os = "linux" | os = "freebsd"}
+  "conf-libxcursor" {os = "linux" | os = "freebsd"}
+  "conf-libxi" {os = "linux" | os = "freebsd"}
+  "conf-libxinerama" {os = "linux" | os = "freebsd"}
+  "conf-libxrandr" {os = "linux" | os = "freebsd"}
```
kit-ty-kate, line 18:
```suggestion
  "conf-mesa" {build}
  "conf-libxcursor" {build}
  "conf-libxi" {build}
  "conf-libxinerama" {build}
  "conf-libxrandr" {build}
```
are the `{os = "linux" | os = "freebsd"}` here for a particular reason? Are they really not needed on other OSs? Support for other OSs could you be added in the future in those packages.

packages/raylib/raylib.0.3.1/opam
```diff
@@ -0,0 +1,49 @@
...
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
+dev-repo: "git+https://github.com/tjammer/raylib-ocaml.git"
+available: [arch != "arm32" & arch != "ppc64"]
+x-ci-accept-failures: [
+  "centos-7"
+  "oraclelinux-7"
```
kit-ty-kate, line 39:
```suggestion
  "centos-7" # C compiler is too old
  "oraclelinux-7" # C compiler is too old
```
Note that this could probably be fixed upstream by adding `-std=c99` to the CFLAGS

packages/raygui/raygui.0.3.1/opam
```diff
@@ -0,0 +1,45 @@
...
+    "@doc" {with-doc}
+  ]
+]
+dev-repo: "git+https://github.com/tjammer/raylib-ocaml.git"
+available: [arch != "arm32" & arch != "ppc64"]
+x-commit-hash: "d04c2066294c95fab7d5fe2f1f795502e988fd92"
+x-ci-accept-failures: [
+  "centos-7"
+  "oraclelinux-7"
+]
```
kit-ty-kate, line 37:
```suggestion
```
This is not needed.Only raylib will fail. CI will detect this.

packages/raylib/raylib.0.3.1/opam
tjammer: Yes, they are. The lib builds fine on other OS without depext, eg windows and macos https://github.com/tjammer/raylib-ocaml/actions/runs/837001056. But on these OS other (native) libs are linked instead of the X libs which are typical for linux. I don't want to stop people from building on other systems, where I don't know the exact dependencies.

I thought about using optional dependencies, but that doesn't really fit either. On linux/bsd these conf-packages are needed, and depexts facilitate easy installing. And for other systems other conf-packages might be needed.

packages/raygui/raygui.0.3.1/opam
tjammer: Raygui also builds some c stubs, so the build of raygui will fail on these distros for the same reason, independant of raylib

packages/raylib/raylib.0.3.1/opam
kit-ty-kate: Oh ok I see, i understand. In which case, could you do:
```suggestion
  "conf-mesa" {os = "linux" | os-family = "bsd"}
  "conf-libxcursor" {os = "linux" | os-family = "bsd"}
  "conf-libxi" {os = "linux" | os-family = "bsd"}
  "conf-libxinerama" {os = "linux" | os-family = "bsd"}
  "conf-libxrandr" {os = "linux" | os-family = "bsd"}
```
instead, so that other BSDs can be compatible?

packages/raygui/raygui.0.3.1/opam
kit-ty-kate: Ok fair enough. Could you add the same comment than in raylib though (just so that it is easier to maintain/understand why this is here in the future)

kit-ty-kate:
Thanks a lot!

PR #18642, opened 2021-05-06
support conf-libev on Homebrew/arm which is in /opt/homebrew

avsm:
only difference from previous version of discover.ml is adding /opt/homebrew to search path


anuragsoni:
I tested this change on my m1 macbook and this seems to work well! I was able to install `conf-libev` and able to use it with lwt.

aantron:
My understanding is that the underlying issue is a usability problem for people installing Dream using opam on m1 macs.

PR #18652, opened 2021-05-08
Package embedded_ocaml_templates.0.6

EmileTrotignon:
### `embedded_ocaml_templates.0.6`
EML is a simple templating language that lets you generate text with plain OCaml
Inspired by EJS templates, it does currently implements all of its functionnality.
I plan to implement everything eventually, especially if someone actually want to use this.
Please contact me if you find this interesting but there is a missing feature that you need !

---
* Homepage: https://github.com/EmileTrotignon/embedded_ocaml_templates
* Source repo: git+https://github.com/EmileTrotignon/embedded_ocaml_templates.git
* Bug tracker: https://github.com/EmileTrotignon/embedded_ocaml_templates/issues

---

### DIFF
--- a/packages/embedded_ocaml_templates/embedded_ocaml_templates.0.6/opam
+++ b/packages/embedded_ocaml_templates/embedded_ocaml_templates.0.6/opam
@@ -1,0 +1,35 @@
+opam-version: "2.0"
+synopsis: "EML is a simple templating language that lets you generate text with plain OCaml"
+description: """
+Inspired by EJS templates, it does currently implements all of its functionnality.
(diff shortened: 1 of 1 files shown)


EmileTrotignon:
I have no understanding of why this fails. I do not have this error when I `opam lint` on my machine. It looks like I need to replace `subst` by `dev`, but I do not know what either of them does.

kit-ty-kate:
Thanks!

kit-ty-kate:
Oh I forgot to ask, could you return the fix to the opam upstream so that it can be taken into account next time you make a release?

PR #18655, opened 2021-05-09
Xlibs conf packages

tjammer:
This adds the (linux + freebsd) dependencies of raylib and raygui packages as conf-packages, as suggested by @kit-ty-kate .
See https://github.com/ocaml/opam-repository/pull/18638


kit-ty-kate:
Thanks a lot for your work!

PR #18659, opened 2021-05-10
Package xmldiff.0.6.0

zoggy:
### `xmldiff.0.6.0`
Computing and applying diffs on XML trees

---
* Homepage: https://zoggy.frama.io/xmldiff/
* Source repo: git+https://framagit.org/zoggy/xmldiff.git
* Bug tracker: https://framagit.org/zoggy/xmldiff/-/issues

---

### DIFF
--- a/packages/xmldiff/xmldiff.0.6.0/opam
+++ b/packages/xmldiff/xmldiff.0.6.0/opam
@@ -1,0 +1,33 @@
+opam-version: "2.0"
+synopsis: "Computing and applying diffs on XML trees"
+maintainer: "Zoggy <zoggy@bat8.org>"
+authors: "Zoggy <zoggy@bat8.org>"
+license: "LGPL-3.0-only"
+homepage: "https://zoggy.frama.io/xmldiff/"
+doc: "https://zoggy.frama.io/xmldiff/refdoc/"
+bug-reports: "https://framagit.org/zoggy/xmldiff/-/issues"
+depends: [
+  "ocaml" {>= "4.12.0"}
+  "ocamlfind" {build}
+  "xmlm" {>= "1.3.0"}
+]
+depopts: [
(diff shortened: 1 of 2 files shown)


packages/xmldiff/xmldiff.0.6.0/opam
```diff
@@ -0,0 +1,33 @@
...
+homepage: "https://zoggy.frama.io/xmldiff/"
+doc: "https://zoggy.frama.io/xmldiff/refdoc/"
+bug-reports: "https://framagit.org/zoggy/xmldiff/-/issues"
+depends: [
+  "ocaml" {>= "4.12.0"}
+  "ocamlfind" {build}
+  "xmlm" {>= "1.3.0"}
+]
+depopts: [
+  "js_of_ocaml" {>= "3.9.0"}
```
kit-ty-kate, line 15: The semantics of constraints inside `depopts` is different compared to `conflicts`. Are you sure you want to have it here?

Currently you're saying that `js_of_ocaml.3.8.0` can be installed concurrently with `xmldiff.0.6.0` without any issues. Is that really the case?

zoggy: Indeed this is not the semantics I expected. But in this case this is not a problem since the configuration step will check the version number of jsoo and behave like it is not available if installed with a version < 3.9.0 .

kit-ty-kate: Mh, the issue here is that they would then be installed concurrently so the installation might be in a state where js_of_ocaml is only partially installed. I wouldn't trust the installation so much.
However I was checking your code and realized that `js_of_ocaml-ppx` is also needed. I believe it should also be listed in the `depopts`. I'll fix both at the same time

kit-ty-kate:

zoggy:
Sure. Thanks to you !

PR #18668, opened 2021-05-12
Package soupault.2.7.0

dmbaturin:
### `soupault.2.7.0`
Static website generator based on HTML rewriting
A website generator that works with page element tree rather than text
and allows you to manipulate pages and retrieve metadata from
existing HTML using arbitrary CSS selectors.

With soupault you can:

* Generate ToC and footnotes.
* Insert file content or an HTML snippet in any element.
* Preprocess element content with external programs (e.g. run `<pre>` tags through a highlighter)
* Extract page metadata (think microformats) and render it using a Jingoo template or an external script.
* Export extracted metadata to JSON.

Soupault is extensible with Lua (2.5) plugins and provides an API for element tree manipulation,
similar to web browsers.

The website generator mode is optional, you can use it as post-processor for existing sites.

---
* Homepage: https://www.soupault.app
* Source repo: git+https://github.com/dmbaturin/soupault
* Bug tracker: https://github.com/dmbaturin/soupault/issues

---

### DIFF
--- a/packages/soupault/soupault.2.7.0/opam
+++ b/packages/soupault/soupault.2.7.0/opam
@@ -1,0 +1,55 @@
+opam-version: "2.0"
+synopsis: "Static website generator based on HTML rewriting"
+description: """\
+A website generator that works with page element tree rather than text
+and allows you to manipulate pages and retrieve metadata from
+existing HTML using arbitrary CSS selectors.
+
+With soupault you can:
+
+* Generate ToC and footnotes.
+* Insert file content or an HTML snippet in any element.
+* Preprocess element content with external programs (e.g. run `<pre>` tags through a highlighter)
+* Extract page metadata (think microformats) and render it using a Jingoo template or an external script.
+* Export extracted metadata to JSON.
+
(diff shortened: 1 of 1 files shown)


mseri:
For `toml`, we need to add a lower bound on the iso library `"0.2.0"`: https://github.com/ocaml/opam-repository/blob/4ff0fe2aa6e75b48062f0455ed0957162e871aaa/packages/toml/toml.6.0.0/opam#L17

kit-ty-kate:
https://github.com/ocaml/opam-repository/pull/18681 should fix this

dmbaturin:
Well, at least it helped uncover some missing constraints... I wonder why I never had these issues with my own builds though, I always got latest versions of both ISO8601 and Menhir.

mseri:
We have recently enabled a CI check to also test packages with the lowest possible bounds for the dependencies, so for some time we will have a number of broken builds and bounds to add

mseri:
This should be fine now, sorry for the wait

dmbaturin:
Oh, it's a great idea actually. I'm ready to fix the lower bounds in my packages.

PR #18666, opened 2021-05-11
[new release] otr (1.0.0)

bikallem:
A typed router for OCaml web applications

- Project page: <a href="https://github.com/lemaetech/otr">https://github.com/lemaetech/otr</a>

##### CHANGES:

- First release.

PR #18672, opened 2021-05-12
Revert "[new release] otr (1.0.0)"

kit-ty-kate:
Reverts ocaml/opam-repository#18666 as it clashes with the `otr` package previously available.


avsm:
This can be an exception to "not deleting packages", as it walks over another package's namespace.

PR #18675, opened 2021-05-12
[new release] pgx_lwt_mirage, pgx_value_core, pgx_async, pgx_unix, pgx_lwt_unix, pgx_lwt and pgx (2.0)

brendanlong:
Pgx using Lwt on Mirage for IO

- Project page: <a href="https://github.com/arenadotio/pgx">https://github.com/arenadotio/pgx</a>
- Documentation: <a href="https://arenadotio.github.io/pgx">https://arenadotio.github.io/pgx</a>

### Breaking changes

* The Pgx module is now wrapped, which means `Pgx_aux`, `Types`, `Access`, etc. aren't added to the global scope.
  The main result of this is that `Pgx_value` now needs to be accessed as `Pgx.Value`.
  (https://github.com/arenadotio/pgx/pull/103)
* `Pgx_async.connect` and `with_conn` now have an additional optional `?ssl` argument (see below).

### Added

* Pgx_async now supports TLS connections using Conduit_async. This is enabled by default and can be controlled with the
  new `?ssl` argument to `connect` and `with_conn`.
  (https://github.com/arenadotio/pgx/pull/108)

### Fixed

* Improved message for authentication errors. Previously these raised `Pgx_eof`, and now they raise
  `PostgreSQL_Error("Failed to authenticate with postgres server", additional details ...)`.
  (https://github.com/arenadotio/pgx/pull/105)

### Changed

* Support new Mirage-conduit timeout argument (https://github.com/arenadotio/pgx/pull/95).

### DIFF
--- a/packages/pgx/pgx.0.1/opam
+++ b/packages/pgx/pgx.0.1/opam
@@ -1,6 +1,7 @@
 opam-version: "2.0"
 authors: ["Arena Developers <silver-snakes@arena.io>"]
 maintainer: "silver-snakes@arena.io"
+license: "LGPL-2.1 with OCaml-LGPL-linking-exception"
 homepage: "https://github.com/arenadotio/pgx"
 dev-repo: "git+https://github.com/arenadotio/pgx.git"
 bug-reports: "https://github.com/arenadotio/pgx/issues"
--- a/packages/pgx/pgx.1.0/opam
+++ b/packages/pgx/pgx.1.0/opam
@@ -24,7 +24,7 @@ depends: [
   "uuidm"
 ]
 build: [
-  ["dune" "subst"] {pinned}
+  ["dune" "subst"] {dev}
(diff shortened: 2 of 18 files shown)


kit-ty-kate:
Thanks! Could you return the fixes to the opam files upstream?

brendanlong:
> Thanks! Could you return the fixes to the opam files upstream?

I added most of these fixes to https://github.com/arenadotio/pgx/pull/113

For our license, I think the fix incorrectly assumed we're using LGPLv2.1 but we're stuck on LGPLv2.0 (since we have to match PG'OCaml: https://github.com/darioteixeira/pgocaml/blob/master/LICENSE.txt).

For the pinned/dev thing, I think that section is generated by dune so I'm not sure how to fix it. Do we need a newer version of dune or something?

PR #18689, opened 2021-05-15
Release i3ipc 0.4

Armael:


Armael:
cc @pdonadeo @mreppen : thank you for the fixes included in this release!

kit-ty-kate:

Armael:
Thanks a lot for the fix! Will do.

PR #18690, opened 2021-05-15
Package bio_io.0.1.1

mooreryan:
### `bio_io.0.1.1`
A library for reading and writing common file formats used in bioinformatics like FASTA files

---
* Homepage: https://github.com/mooreryan/bio_io
* Source repo: git+https://github.com/mooreryan/bio_io.git
* Bug tracker: https://github.com/mooreryan/bio_io/issues

---

### DIFF
--- a/packages/bio_io/bio_io.0.1.1/opam
+++ b/packages/bio_io/bio_io.0.1.1/opam
@@ -9,9 +9,11 @@ doc: "https://mooreryan.github.io/bio_io/"
 bug-reports: "https://github.com/mooreryan/bio_io/issues"
 depends: [
   "dune" {>= "2.8"}
-  "bisect_ppx" {with-test}
-  "core_kernel"
-  "ocaml" {>= "4.08.0"}
+  "bisect_ppx" {dev}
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
I made a few improvements to your list of dependencies:
* `ppx_bisect` doesn't do anything during the tests in CI. It's only useful when called by hand so I moved it from a test dependency to a dev dependency (only installed when installed from the source repository)
* I added the lower-bound constraints for the janestreet packages (your package requires at least v0.12)
* Explicited the dependencies to ppx_sexp_conv (used explicitly in your package)
* Relaxed the constraint for OCaml to >= 4.07 and >= 4.08 for the tests only (based on the CI result of the previous PR for this package)

Could you return the fixes to the opam file upstream so that it can be taken into account the next time you make a release? You can of course change it as you wish if you don't quite agree with my changes.

mooreryan:
Thanks for the help! If you don't mind me asking, how did you find the min required versions for the core kernel and ppx libraries? Is there an easy way to find that info out? Thanks!!

PR #18693, opened 2021-05-16
[new release] elpi (1.13.5)

gares:
CHANGES:

- API:
  - New `prune` which can be used to prune a unification variable
  - New `distinct_names` which checks if a list of arguments is in the pattern
    fragment


kit-ty-kate:
Thanks!

PR #18694, opened 2021-05-17
Higlo moving

zoggy:
Update URLs of higlo after migrating to Framagit. Urls of already released archives are fixed to refer to github.io instead of github.com.


zoggy:
Sorry, dit not want to make a PR containing also the new package conf-goocanvas.

PR #18698, opened 2021-05-17
Various fixes (lower-bounds edition, round 3)

kit-ty-kate:
cc @dra27 for the changes in camlp4.4.02+{1,2,3,4} and ocaml-secondary-compiler.4.08.1


dra27:
Hah, looking at it again, the patch in https://github.com/camlp4/camlp4/pull/126 is more ruthless than it should be... `C.ocamlnat` should in fact be `C.natdynlink` (which needs probing in `configure`) so nothing has been using native plugins since that was merged, given that ocamlnat isn't built by the opam compiler packages.

In the interests of not having any risk of regressing https://github.com/ocaml/opam/issues/2276, could you mark those packages where the patch is removed as `available: os != "cygwin"` (it's the one platform where it's very easy to have native code but not native dynlink) but that's obviously just for belt and braces!

dra27:
The ocaml-secondary-compiler change reverts the suggestion in https://github.com/ocaml/opam-repository/pull/16728#pullrequestreview-439094475, but the lower-bounds checker reveals that was a bad suggestion of mine (and you have obviously fixed the patch, as noted in the review!) 🙂

PR #18699, opened 2021-05-18
Add conf-capnproto.2.

ewanmellor:
This is the same as conf-capnproto.1, except the Alpine package has
moved from @testing to @edgecommunity.

This will require a line in /etc/apk/repositories like:
@edgecommunity https://dl-cdn.alpinelinux.org/alpine/edge/community

Note that we are preserving conf-capnproto.1 (even though it no longer
builds with the current Alpine package repos) so that users with an old
mirror can continue to use that.

Signed-off-by: Ewan Mellor <ewan@tarides.com>

### DIFF
--- a/packages/conf-capnproto/conf-capnproto.2/opam
+++ b/packages/conf-capnproto/conf-capnproto.2/opam
@@ -1,0 +1,28 @@
+opam-version: "2.0"
+maintainer: "Kate <kit.ty.kate@disroot.org>"
+authors: [
+  "Sandstorm Development Group, Inc."
+  "Cloudflare, Inc."
+  "other contributors"
+]
+homepage: "https://capnproto.org/"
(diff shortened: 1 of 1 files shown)


talex5:
I didn't notice this PR when I filed #18704. Feel free to close mine if you prefer.

> Note that we are preserving conf-capnproto.1 (even though it no longer builds with the current Alpine package repos) so that users with an old mirror can continue to use that.

I'm not sure, but it seems a bit unlikely that someone would use a stale mirror of the testing repository, but continue to update opam-repository. I think it would be better to treat this like a URL that has moved, and update it in-place.

avsm:
I did ask @ewanmellor to duplicate the version in a private conversation, but it does seem unlikely given how Alpine organises their repos that this anything more than a URL change.  So I merged #18704 instead, but thanks for the PR @ewanmellor!

PR #18702, opened 2021-05-18
Package goblint-cil.1.8.0

michael-schwarz:
### `goblint-cil.1.8.0`
A front-end for the C programming language that facilitates program analysis and transformation
This is a fork of the 'cil' package needed to build 'goblint'.
Changes:
- Proper support for C99, (#9) and VLAs in particular (#5, #7)
- It uses Zarith instead of the deprecated Num
- Support for more recent OCaml versions (≥ 4.06)
- Large integer constants that do not fit in an OCaml int are represented as a string instead of getting truncated
- Syntactic search extension (#21)
- Some warnings were made optional
- Unmaintained extensions (#30) were removed
- Many bug fixes

---
* Homepage: https://cil-project.github.io/cil/
* Source repo: git+https://github.com/goblint/cil.git
* Bug tracker: https://github.com/goblint/cil/issues

---

### DIFF
--- a/packages/goblint-cil/goblint-cil.1.8.0/opam
+++ b/packages/goblint-cil/goblint-cil.1.8.0/opam
@@ -1,0 +1,64 @@
+opam-version: "2.0"
+synopsis:
+  "A front-end for the C programming language that facilitates program analysis and transformation"
+description: """
+This is a fork of the 'cil' package needed to build 'goblint'.
+Changes:
+- Proper support for C99, (#9) and VLAs in particular (#5, #7)
+- It uses Zarith instead of the deprecated Num
+- Support for more recent OCaml versions (≥ 4.06)
+- Large integer constants that do not fit in an OCaml int are represented as a string instead of getting truncated
+- Syntactic search extension (#21)
+- Some warnings were made optional
+- Unmaintained extensions (#30) were removed
+- Many bug fixes"""
+maintainer: [
+  "Michael Schwarz <michael.schwarz93@gmail.com>"
+  "Ralf Vogler <ralf.vogler@gmail.com>"
+]
+authors: ["gabriel@kerneis.info"]
+license: "BSD-3-Clause"
+homepage: "https://cil-project.github.io/cil/"
+bug-reports: "https://github.com/goblint/cil/issues"
+depends: [
+  "ocaml" {>= "4.04.2"}
+  "ocamlfind"
+  "zarith"
+  "hevea" {with-doc}
+  "dune" {>= "2.7"}
+  "odoc" {with-doc}
+  "stdlib-shims"
+  "ppx_deriving_yojson" {>= "3.2"}
+  "yojson"
+  "batteries" {>= "3.2.0"}
+  "conf-perl"
+]
+conflicts: ["cil"]
+build: [
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
+dev-repo: "git+https://github.com/goblint/cil.git"
+depexts: [
+  ["perl-ExtUtils-MakeMaker"] {os-distribution = "centos" | os-distribution = "fedora" | os-distribution = "ol"}
+  ["perl-FindBin"] {os-distribution = "fedora"}
+  ["build-base"] {os-distribution = "alpine"}
+]
+available: arch != "ppc32" & arch != "ppc64"
+url {
+  src: "https://github.com/goblint/cil/archive/1.8.0.tar.gz"
+  checksum: [
+    "md5=796ad26120b5c6b939a57e8623088aef"
+    "sha512=01a58ac6d928afead21c8a97af5865715114cd0562234d1d4aef9e4ac5d91415d040a15927c52cb896dbb39a53e915627f498ebe2d026a548c3ff597682041b2"
+  ]
+}


michael-schwarz:
Sorry for the spurious references. The references in the description refer to issues/PRs in the `goblint-cil` issue tracker, I did not remember to prefix them so GitHub also understands that here.

michael-schwarz:
Most of the failing tests are caused by https://github.com/goblint/cil/issues/37.
I will update this PR when these are fixed.

packages/goblint-cil/goblint-cil.1.8.0/opam
```diff
@@ -0,0 +1,67 @@
...
+authors: "gabriel@kerneis.info"
+license: "BSD-3-Clause"
+homepage: "https://cil-project.github.io/cil/"
+bug-reports: "https://github.com/goblint/cil/issues"
+depends: [
+  "ocaml" {>= "4.04.2"}
+  "ocamlfind"
+  "zarith"
+  "hevea" {with-doc}
+  "dune" {>= "2.7" & >= "2.7"}
```
kit-ty-kate, line 28: could you remove the redundant conflict here? This should be easily done by removing the `(dune (>= 2.7))` line from your dune-project file.

packages/goblint-cil/goblint-cil.1.8.0/opam
```diff
@@ -0,0 +1,67 @@
...
+  "odoc" {with-doc}
+  "stdlib-shims"
+  "ppx_deriving_yojson"
```
kit-ty-kate, line 31:
```suggestion
  "ppx_deriving_yojson" {>= "3.2"}
```
You are using dune and dune requires ppxs to be built with dune when used using the built-in method

packages/goblint-cil/goblint-cil.1.8.0/opam
```diff
@@ -0,0 +1,67 @@
...
+  "Michael Schwarz <michael.schwarz93@gmail.com>"
+  "Ralf Vogler <ralf.vogler@gmail.com>"
+]
+authors: "gabriel@kerneis.info"
+license: "BSD-3-Clause"
+homepage: "https://cil-project.github.io/cil/"
+bug-reports: "https://github.com/goblint/cil/issues"
+depends: [
+  "ocaml" {>= "4.04.2"}
+  "ocamlfind"
```
kit-ty-kate, line 25:
```suggestion
```
This dependency doesn't seem to be used anymore

packages/goblint-cil/goblint-cil.1.8.0/opam
```diff
@@ -0,0 +1,67 @@
...
+  ]
+]
+depexts: [
+  ["libc6-dev-i386"] {os-family = "debian" | os-family = "ubuntu"}
+  ["glibc-devel-32bit"] {os-family = "suse"}
+  ["glibc-devel.i686"]
+    {os-distribution = "fedora" | os-distribution = "rhel" |
+     os-distribution = "centos" |
+     os-distribution = "ol"}
+  ["lib32-glibc-devel"] {os-family = "arch"}
```
kit-ty-kate, line 58: I'm curious, why is this required? If you require the user to use x86_32 you can simply say so by replacing this by:
```
available: arch = "x86_32"
```
Though I would advise against it and would recommend a more portable approach (which would be neither requiring x86_32 glibc or requiring the package to be installed on x86_32 only

packages/goblint-cil/goblint-cil.1.8.0/opam
michael-schwarz: :+1:

packages/goblint-cil/goblint-cil.1.8.0/opam
michael-schwarz: :+1:

packages/goblint-cil/goblint-cil.1.8.0/opam
michael-schwarz: These packages are the 32bit headers of glibc for a 64bit architecture. `goblint-cil` attempts to build a [machine model](https://people.eecs.berkeley.edu/~necula/cil/cil007.html) also for the case where one is compiling for a 32bit target during the testing phase. However, the tests still succeed if this x86 machine model can not be built, so I will remove this requirement again.

packages/goblint-cil/goblint-cil.1.8.0/opam
michael-schwarz: The configure script from the old days that we still use relies on `ocamlfind` being available, and at least one of the tests actively needs it. In the long run we probably want to drop that dependency, but right now it does not seem to be worth the extra effort.

michael-schwarz:
@kit-ty-kate: Thanks for your suggestions! I fixed most of the issues. 3 things are still failing 

- The builds for Alpine still fail because it uses `musl` as the standard library and that has some specific quirks we do not support in CIL (yet). Most things work though. **What is the policy here? Should I exclude Alpine completely or can we merge packages where the CI is not 100% green?**

- This `lower bounds` test has me really stumped. Any ideas how one could go about determining what is at fault? https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/4fa13d094cf669f00bbd4653e5aaeb520cf66988/variant/opam-2.1,compilers,4.07,goblint-cil.1.8.0,lower-bounds

michael-schwarz:
From our side we could go ahead with the merge, what is the policy on merging PRs that are not 100% green?

kit-ty-kate:
Sorry this took so long, I forgot about. The remaining failures don't seem related. Thanks!

PR #18704, opened 2021-05-18
Update conf-capnproto for Alpine changes

talex5:
Alpine has moved the capnproto package from `testing` to `community`.
Attempting to install it now fails with:

    $ docker run --rm -it ocaml/opam:alpine-3.13-ocaml-4.12
    bash-5.1$ opam depext conf-capnproto
    capnproto-dev@testing
    The following command needs to be run through "sudo":
        apk add capnproto-dev@testing
    ERROR: unable to select packages:
      capnproto-0.8.0-r1:
        masked in: @edgecommunity

### DIFF
--- a/packages/conf-capnproto/conf-capnproto.1/opam
+++ b/packages/conf-capnproto/conf-capnproto.1/opam
@@ -13,7 +13,7 @@ build: [
   ["capnp" "--version"]
 ]
 depexts: [
-  ["capnproto-dev@testing"] {os-distribution = "alpine"}
+  ["capnproto-dev@edgecommunity"] {os-distribution = "alpine"}
   ["capnproto"] {os-distribution = "arch"}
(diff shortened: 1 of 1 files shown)


talex5:
Not sure why some distributions are failing with opam 2.1, but are OK with opam 2.0.

kit-ty-kate:
mmh, indeed, there is an issue when EPEL is involved. I opened a ticket in https://github.com/ocaml/opam/issues/4669

talex5:
Thanks. It's also failing on Alpine with opam 2.1.

kit-ty-kate:
Yes, sorry i just realized. I'm writing another ticket. This one seems easier to deal with so I'm looking into the code as well

avsm:
Let's unblock opam 2.0!

PR #18712, opened 2021-05-19
conf-capnproto: switch to edgecommunity on alpine

avsm:
Since https://github.com/ocurrent/docker-base-images/pull/109
and https://github.com/avsm/ocaml-dockerfile/pull/48, we can
use the edgecommunity repository, which is where the depext for
capnproto now lives


talex5:
Duplicate of #18704 and #18699... perhaps someone could merge one of these?

avsm:
dup of #18704

PR #18713, opened 2021-05-19
Release Fmlib version 0.3.0

hbr:
new file:   fmlib/fmlib.0.3.0/opam
	new file:   fmlib_js/fmlib_js.0.3.0/opam
	new file:   fmlib_parse/fmlib_parse.0.3.0/opam
	new file:   fmlib_pretty/fmlib_pretty.0.3.0/opam
	new file:   fmlib_std/fmlib_std.0.3.0/opam

### DIFF
--- a/packages/fmlib/fmlib.0.3.0/opam
+++ b/packages/fmlib/fmlib.0.3.0/opam
@@ -1,0 +1,46 @@
+opam-version: "2.0"
+
+synopsis: "Functional monadic library"
+
+description: """
+
+Library supporting functional programming with managed effects. It features
+
+- Fmlib_std: A lot of standard data types.
+
+- Fmlib_parse: Monadic parsing functions with combinators suitable for indentation sensitive
+  parsing.
+
+- Fmlib_pretty: Pretty printing functions.
+
+- Fmlib_js: Interface to javscript via js_of_ocaml.
+
+"""
+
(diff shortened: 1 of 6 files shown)


hbr:
I don't understand why the ci-tests are trying to build version 0.2.0. This release is 0.3.0 and all tests with 0.3.0 passed.

Between 0.2.0 and 0.3.0 I have splitted the package `fmlib` into the 3 packages `fmlib_std`, `fmlib_pretty` and `fmlib_parse`. The package `fmlib_js` has already been present in 0.2.0 as a separate package but it has used in 0.2.0 the library `Fmlib_std` as part of `fmlib`.

Has anybody an idea what's going on here? Thanks for any hint.

kit-ty-kate:
> I don't understand why the ci-tests are trying to build version 0.2.0

It does because I fixed a constraint in the previous version as well as the new one. (`ppx_inline_test`, see my previous commits)
It also showed another missing constraint in dune. That was the reason why this package was failing here. I've removed your previous commit as well since `{= version}` is equivalent to `{= "0.2.0"}` in this case.

These failures were detected by our new lower-bounds constraint checker which wasn't there last time you released fmlib, that is why.

It should now be ready to go once CI finishes.

kit-ty-kate:
Thanks! Could you return the `fmlib_pretty` missing-depedency fix and the `ppx_inline_test` constraint fix to the opam file upstream so it can be taken into account for future releases?

PR #18714, opened 2021-05-19
[new release] inquire (0.3.0)

tmattio:
CHANGES:

## Added

- Improved all prompts prompt to handle escape sequences (e.g. "Ctrl-L" will clear the screen while reading from answer).
- Improved `select` to allow selection using indexes.
- Improved prompts to clear defaults and other tooltip on a valid answer and print the answer after the prompt.

## Changed

- Stripped out `lambda-term` dependency
- Removed `lwt` integration


kit-ty-kate:
Thanks! Sorry for the noise. Could you return the changes I made to inquire.0.3.0 to the opam file upstream?

Also `spin` might need a new release (if you can also return the changes to the constraints I made to spin.0.7.0 too that would be perfect.

tmattio:
Wow, I opened a can of worms, didn't I? 😅 
Thanks a lot for the constraint work, I'll upstream to `inquire` and cut a `spin` release soon!

PR #18716, opened 2021-05-20
Package kind2.v1.4.0

daniel-larraz:
### `kind2.v1.4.0`
Multi-engine, parallel, SMT-based automatic model checker for safety properties of Lustre programs
Kind 2 is an open-source, multi-engine, SMT-based automatic model checker for safety properties of finite-state or infinite-state synchronous reactive systems expressed as in an extension of the Lustre language. In its basic configuration it takes as input a Lustre file annotated with properties to be proven invariant, and outputs for each property either a confirmation or a counterexample, i.e., a sequence inputs that falsifies the property. More advanced features include contract-based compositional verification, proof generation for proven properties, and contract-based test generation.

---
* Homepage: https://kind2-mc.github.io/kind2
* Source repo: git+https://github.com/kind2-mc/kind2.git
* Bug tracker: https://github.com/kind2-mc/kind2/issues

---

### DIFF
--- a/packages/kind2/kind2.v1.4.0/opam
+++ b/packages/kind2/kind2.v1.4.0/opam
@@ -1,0 +1,45 @@
+opam-version: "2.0"
+synopsis:
+  "Multi-engine, parallel, SMT-based automatic model checker for safety properties of Lustre programs"
+description:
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Thanks! Could you return the fixes to the opam file upstream so they can be taken into account for the next release?

daniel-larraz:
> Thanks! Could you return the fixes to the opam file upstream so they can be taken into account for the next release?

Sure, will do! But I have a question about one of your changes. I use `dune` to generate automatically the opam file, and dune generates the following field `["dune" "subst"] {pinned}` instead of `["dune" "subst"] {dev}` (it does not seem to be customizable). Could you please explain the differences between the two values? Would it be make sense to suggest changing the value to dune's developers?

kit-ty-kate:
This was fixed in dune 2.7.0, however dune does not change the automatic generation of the opam file without changing the appropriate `(lang dune ...)` to `2.7` or above in your dune-project file.

See more details in https://github.com/ocaml/dune/pull/3647

daniel-larraz:
Thanks for the info!

PR #18722, opened 2021-05-21
[new release] bitwuzla and bitwuzla-bin (0.0.0)

recoules:
SMT solver for QF_AUFBVFP

- Project page: <a href="https://bitwuzla.github.io">https://bitwuzla.github.io</a>
- Documentation: <a href="https://bitwuzla.github.io/docs/ocaml/">https://bitwuzla.github.io/docs/ocaml/</a>

##### CHANGES:

Initial release.

OCaml binding for the SMT solver [Bitwuzla](https://bitwuzla.github.io/).

- `Bitwuzla_c` library exposes low level functions from the
  [C API](https://bitwuzla.github.io/docs/c/interface.html);
- `Bitwuzla_z` library converts bitvector value to
  [Zarith](https://github.com/ocaml/Zarith) integer;
- `bitwuzla-bin` package installs Bitwuzla executable.

Vendor submodules:
- [Cadical](https://github.com/arminbiere/cadical) tag:sc2021
- [SymFPU](https://github.com/martin-cs/symfpu) commit:8fbe139
- [Btor2Tools](https://github.com/Boolector/btor2tools) commit:1df768d
- [Bitwuzla](https://github.com/bitwuzla/bitwuzla) commit:2f4dad6

### DIFF
--- a/packages/bitwuzla-bin/bitwuzla-bin.0.0.0/opam
+++ b/packages/bitwuzla-bin/bitwuzla-bin.0.0.0/opam
@@ -19,9 +19,6 @@ depends: [
   "conf-g++" {build}
   "conf-gmp" {build}
 ]
-depexts: [
-  ["libstdc++-static" "libc-static" "gcc-static"] {os-family = "fedora"}
-]
 build: [
   ["dune" "subst"] {dev}
   [
@@ -44,4 +41,4 @@ url {
     "sha512=a94a466ef8b299c18cde51dd82e1be61f357de0e5efe617320ee7a888aeb8f7cd53d278f1e70e80b24c43003afe9595b8ad97c91c07a866616003e39642b35e8"
   ]
 }
-available: [ os-family = "debian" | os-family = "alpine" | os-family = "fedora" ]
+available: [ os-family = "debian" | os-family = "alpine" ]
--- a/packages/bitwuzla-bin/bitwuzla-bin.0.0.0/opam
+++ b/packages/bitwuzla-bin/bitwuzla-bin.0.0.0/opam
@@ -1,0 +1,45 @@
+opam-version: "2.0"
+synopsis: "Bitwuzla SMT solver executable"
+description: """
+
+Standalone installer for the SMT solver Bitwuzla.
+
+Bitwuzla is a Satisfiability Modulo Theories (SMT) solver for the theories of fixed-size bit-vectors, arrays and uninterpreted functions and their combinations. Its name is derived from an Austrian dialect expression that can be translated as “someone who tinkers with bits”."""
+maintainer: ["Frédéric Recoules <frederic.recoules@cea.fr>"]
+authors: ["Frédéric Recoules"]
+license: "MIT"
+tags: ["SMT solver" "SMT-COMP 2020" "QF_AUFBVFP"]
+homepage: "https://bitwuzla.github.io"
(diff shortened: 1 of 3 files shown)


packages/bitwuzla-bin/bitwuzla-bin.0.0.0/opam
```diff
@@ -0,0 +1,44 @@
...
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+  ]
```
kit-ty-kate, line 32: why has the tests and doc been removed from there? It seem to be only a matter of adding a missing test dependency

packages/bitwuzla-bin/bitwuzla-bin.0.0.0/opam
```diff
@@ -0,0 +1,44 @@
...
+x-commit-hash: "e1f77a4a4b12c9a8b624c6d9f11b0409f9b092fa"
+url {
+  src:
+    "https://github.com/bitwuzla/ocaml-bitwuzla/releases/download/0.0.0/bitwuzla-bin-0.0.0.tbz"
+  checksum: [
+    "sha256=d4226ee5d1972caae12ea7bfde7ffd20a04fa81abb7a85a7432fb71be34b6207"
+    "sha512=a94a466ef8b299c18cde51dd82e1be61f357de0e5efe617320ee7a888aeb8f7cd53d278f1e70e80b24c43003afe9595b8ad97c91c07a866616003e39642b35e8"
+  ]
+}
+available: [ os-family = "debian" | os-distribution = "alpine" ]
```
kit-ty-kate, line 44: This seems a bit excessive. Is there any chance you could find the reason why?

packages/bitwuzla-bin/bitwuzla-bin.0.0.0/opam
recoules: Because this package is just a convenient way to install bitwuzla in `bin`. There is no test associated to it and so, no reason to force to install useless dependencies.

packages/bitwuzla-bin/bitwuzla-bin.0.0.0/opam
recoules: Well, I will investigate, but not quite soon (yet the errors are on the build of vendor foreign libraries).

Since it is only the first release, I think it is better to release it for people for which it can work right now and release new patches/versions to relax the constraints.

recoules:
I can no longer check the details of the `opam-ci`. Is there a problem with the server?

avsm:
The CI server was just offline briefly; its back now. However, you seem to have bitwuzla-bin.0.0.0 and bitwuzla.0.0.1 in this PR, but the other two variants are missing (bitwuzla-bin.0.0.1 and bitwuzla.0.0.0).  Could you please confirm that matches your intent?

recoules:
> The CI server was just offline briefly; its back now. However, you seem to have bitwuzla-bin.0.0.0 and bitwuzla.0.0.1 in this PR, but the other two variants are missing (bitwuzla-bin.0.0.1 and bitwuzla.0.0.0). Could you please confirm that matches your intent?

Yes, in fact, both are version `0.0`. The `bitwuzla-bin` will be discontinued, but yet, it can still be used as is on `debian-family` distributions. The `bitwuzla-0.0.1` is just a fix over the initial release `0.0.0` that do not make sense to publish anymore.

PR #18727, opened 2021-05-22
Package feather.b1fb1b724cc0

charlesetc:
### `feather.b1fb1b724cc0`
A minimal shell interface

---
* Homepage: https://github.com/charlesetc/feather
* Bug tracker: https://github.com/charlesetc/feather/issues

---

### DIFF
--- a/packages/feather/feather.b1fb1b724cc0/opam
+++ b/packages/feather/feather.b1fb1b724cc0/opam
@@ -1,0 +1,33 @@
+opam-version: "2.0"
+synopsis: "A minimal shell interface"
+maintainer: ["charles@nrwhl.xyz"]
+authors: ["Charles Chamberlain"]
+license: "MIT"
+homepage: "https://github.com/charlesetc/feather"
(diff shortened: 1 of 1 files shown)


charlesetc:
Does this look right? I'm a little confused why it's taking the commit hash instead of the tag when I run `opam publish --tag 0.2.0`. Thanks for the help!

kit-ty-kate:
opam-publish has detected the version number based on the version number in the opam file in your repository:
https://github.com/charlesetc/feather/blob/8314ea30d535f29dbdb6d91d927951b00e730605/feather.opam#L3

To fix that you can either re-run `opam publish -v 0.2.0 --tag 0.2.0` or change the version field in the opam file directly and re-run opam publish (you probably need to de this anyway, otherwise the same thing will happen next time you do a release)

Side note: you might have to change the dune-project file instead, which seems to be the one that has autogenerated the opam file for you.

charlesetc:
Thank you!

PR #18729, opened 2021-05-23
Make sure conf-libX11 exports PKG_CONFIG_PATH as expected

smorimoto:


kit-ty-kate:
I'm not sure to understand what's the difference between the two here. Could you detail this a little?

smorimoto:
The problem is that we don't export `PKG_CONFIG_PATH`. The graphics library is checking the existence of `PKG_CONFIG_PATH` and manipulating the `PKG_CONFIG_PATH`, which complicated the problem. In other words, we didn't actually use any of the environment variables we had set up in this package so far, only the discover module in the graphics library handled that.

smorimoto:
The local test tells me that this works as expected.

kit-ty-kate:
> The problem is that we don't export `PKG_CONFIG_PATH`. The graphics library is checking the existence of `PKG_CONFIG_PATH` and manipulating the `PKG_CONFIG_PATH`, which complicated the problem. In other words, we didn't actually use any of the environment variables we had set up in this package so far, only the discover module in the graphics library handled that.

I'm confused at how this change make opam export the `PKG_CONFIG_PATH` variable outside of this particular command inside the `conf-libx11` package.
I would have expected @mseri's proposal in https://github.com/ocaml/opam-repository/pull/18530#issue-617750877 to be the necessary fix here if that was the case.

smorimoto:
Oh, certainly it might be better to add some notes in `post-messages`. At first, I thought it would be better to centralize these things in the opam repository to absorb the complexity here, but on second thought it doesn't seem that ideal.

PR #18730, opened 2021-05-24
new file:   packages/line_oriented/line_oriented.1.0.0/opam

UnixJunkie:


kit-ty-kate:
This package does not seem to contain anything (nothing is built)

UnixJunkie:
How come? If I do this in a fresh git checkout:
```
opam pin add line_oriented $PWD
```
The package is correctly built and installed.

UnixJunkie:
Maybe a problem with "line-oriented" instead of "line_oriented" somewhere?

kit-ty-kate:
There is simply no `build` or `install` field in this opam file so nothing is executed

UnixJunkie:
ok, corrected

kit-ty-kate:
Thanks!

PR #18270, opened 2021-03-05
[new release] asn1-combinators (0.2.5)

hannesm:
Embed typed ASN.1 grammars in OCaml

- Project page: <a href="https://github.com/mirleft/ocaml-asn1-combinators">https://github.com/mirleft/ocaml-asn1-combinators</a>
- Documentation: <a href="https://mirleft.github.io/ocaml-asn1-combinators/doc">https://mirleft.github.io/ocaml-asn1-combinators/doc</a>

##### CHANGES:

* Fix an integer overflow in the length field on 32 bit architectures
  (mirleft/ocaml-asn1-combinators#36 by @hannesm)

PR #18119, opened 2021-02-09
8 packages from ocaml/opam at 2.0.8

rjbou:
This pull-request concerns:
-`opam-client.2.0.8`: opam 2.0 development libraries (client)
-`opam-core.2.0.8`: opam 2.0 development libraries (core)
-`opam-devel.2.0.8`: opam 2.0 bootstrapped binary
-`opam-format.2.0.8`: opam 2.0 development libraries (format)
-`opam-installer.2.0.8`: Installation of files to a prefix, following opam conventions
-`opam-repository.2.0.8`: opam 2.0 development libraries (repository)
-`opam-solver.2.0.8`: opam 2.0 development libraries (solver)
-`opam-state.2.0.8`: opam 2.0 development libraries (state)

---
* Source repo: git+https://github.com/ocaml/opam.git
* Bug tracker: https://github.com/ocaml/opam/issues

---
:camel: Pull-request generated by opam-publish v2.0.3

PR #18731, opened 2021-05-24
Constrain older opam-format to older opam-file-format

dra27:
#18119 introduced a constraint on opam-file-format for opam-format.2.0.8 - this should either be present on the earlier versions too or not present at all. The same constraint applies to opam-format.2.1.0~beta2 (failure seen in [revdeps for #18270](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/9528af89646f39e2f463821ce0e167d89decf024/variant/opam-2.1,compilers,4.12,opam-file-format.2.1.3,revdeps,opam-devel.2.1.0~beta2).

The caveat for this is that it's seeking to avoid deprecation warnings, so it's not _completely_ clear to me if these constraints should be present at all.

### DIFF
--- a/packages/opam-format/opam-format.2.0.0/opam
+++ b/packages/opam-format/opam-format.2.0.0/opam
@@ -22,7 +22,7 @@ build: [
 depends: [
   "ocaml" {>= "4.02.3"}
   "opam-core" {= "2.0.0"}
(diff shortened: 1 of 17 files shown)


avsm:
Applying this consistently with #18119 to all versions makes sense to me.

PR #18732, opened 2021-05-24
Change all dune subst {pinned} into dune subst {dev}

kit-ty-kate:
This was a mistaken guideline/autogeneration in jbuilder/dune until dune 2.7. See https://github.com/ocaml/dune/pull/3647

This is a draft for now as it will make everyone rebuild their packages but it might be worth doing it at some point if only to give a good example.


avsm:
This looks fine to just merge, rather than holding it open and the conflicts growing. It also doesn't introduce a semantic change into the packages beyond fixing a bug (the specific case of a pinned archive will change), so no version bump for the packages is required.

mseri:
I vote for merging this fast

avsm:
I've eyeballed the diff and it looks fine. Merging, thanks @kit-ty-kate !

PR #18733, opened 2021-05-24
Various fixes (lower-bounds checker edition, round 4)

kit-ty-kate:


packages/conf-rust-2018/conf-rust-2018.1/opam
```diff
@@ -6,21 +6,25 @@ bug-reports: "https://github.com/ocaml/opam-repository/issues"
 license: "MIT"
 extra-files: ["test.rs" "md5=639e04c270fef8589636e0416761a67b"]
```
avsm, line 7: This change to conf-rust looks unrelated to lower bounds.

kit-ty-kate: I moved this into its own PR

PR #18744, opened 2021-05-26
Remove opensource@janestreet.com email address

jeremiedimino:
This PR removes the opensource@janestreet.com email address from Jane Street packages. This is because we are making this list internal to Jane Street only to match the way it is currently being used.

We minted a new email address to use as a point of contact. However, I didn't use it here as it is not meant for opam related queries. Github or slack are the preferred communication channels for such queries.

### DIFF
--- a/packages/accessor/accessor.v0.14.0/opam
+++ b/packages/accessor/accessor.v0.14.0/opam
@@ -1,6 +1,6 @@
 opam-version: "2.0"
-maintainer: "opensource@janestreet.com"
-authors: ["Jane Street Group, LLC <opensource@janestreet.com>"]
+maintainer: "Jane Street developers"
+authors: ["Jane Street Group, LLC"]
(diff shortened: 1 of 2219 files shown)


avsm:
Eyeballed the patch to confirm it's only changing the maintainers and authors fields.  Looks fine, thanks!

kit-ty-kate:
This broke `dune-configurator.1.11.4`. The sed command used also modified the patch file without changing its checksum.

kit-ty-kate:
> This broke `dune-configurator.1.11.4`. The sed command used also modified the patch file without changing its checksum.

Fixed in https://github.com/ocaml/opam-repository/pull/18749

jeremiedimino:
Thanks @kit-ty-kate for the fix!

avsm:
good catch @kit-ty-kate !

PR #18748, opened 2021-05-26
ocaml-base-compiler.4.02.3: use the git release tarball

avsm:
caml.inria.fr is currently down, so this swaps the 4.02.3 release archive to the one found off the git release tag. Oddly, this is a different checksum due to the following extra files being found in the ocaml git release:

Only in ocaml-4.02.3/compilerlibs: .gitignore
Only in ocaml-4.02.3: experimental
Only in ocaml-4.02.3/ocamlbuild: test

These all look like harmless extra directories that appear to have been stripped out of the caml.inria.fr 4.02.3 release tarball, so this change should be safe enough in opam-repo

(this change is fairly urgent to unbreak the docker-base-images build used in our CI)

### DIFF
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.4.02.3/opam
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.4.02.3/opam
@@ -39,8 +39,8 @@ build: [
 ]
 install: [make "install"]
 url {
-  src: "http://caml.inria.fr/pub/distrib/ocaml-4.02/ocaml-4.02.3.tar.gz"
-  checksum: "md5=ef1a324608c97031cbd92a442d685ab7"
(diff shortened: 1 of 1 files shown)


avsm:
cc @Octachron regarding the discrepancy between caml.inria.fr and the git release tag

dra27 (approved):
4.03.0 onwards already use the GitHub tarballs.

Octachron:
This sounds fine. However, caml.inria.fr doesn't seem globally down? At least, I have just that I can reach the 4.02.3 archive.

Octachron:
It looks like it is the http address which is down and not the https one?

avsm:
Aha, it could be fixed by switching to the `https` url too!  For consistency with all the other base packages though, I'll go ahead and merge this to use the git tagged sources if there's nothing much in it between the two choices.

PR #18749, opened 2021-05-26
dune-configurator.1.11.4: Fix patch checksum (broken by #18744)

kit-ty-kate:


kit-ty-kate:
Lint test passed. Merging.

PR #18756, opened 2021-05-29
Package testu01.1.2.3-0.1

Niols:
### `testu01.1.2.3-0.1`
OCaml bindings for TestU01 1.2.3
This package provides OCaml bindings for TestU01 1.2.3 TestU01 is
  a software library, implemented in C, and offering a collection of utilities
  for the empirical statistical testing of uniform random number generators.
  The OCaml bindings allow for easy testing of random number generators written
  in OCaml and that claim to be uniform.

---
* Homepage: https://github.com/LesBoloss-es/ocaml-testu01
* Source repo: git+https://github.com/LesBoloss-es/ocaml-testu01.git
* Bug tracker: https://github.com/LesBoloss-es/ocaml-testu01/issues

---

### DIFF
--- a/packages/testu01/testu01.1.2.3-0.1/opam
+++ b/packages/testu01/testu01.1.2.3-0.1/opam
@@ -1,0 +1,46 @@
+opam-version: "2.0"
+synopsis: "OCaml bindings for TestU01 1.2.3"
+description: """
+This package provides OCaml bindings for TestU01 1.2.3 TestU01 is
+  a software library, implemented in C, and offering a collection of utilities
+  for the empirical statistical testing of uniform random number generators.
+  The OCaml bindings allow for easy testing of random number generators written
+  in OCaml and that claim to be uniform."""
+maintainer: ["Niols “Niols” Jeannerod <niols@niols.fr>"]
+authors: [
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
The version number seems rather uncommon, especially for a first version. Is `1.2.3-0.1` really expected to be the version number for this release?

Niols:
That is a question that bothered me actually quite a bit. Since those are bindings, shouldn't the version number of the C library (here, 1.2.3) show up in the version number of the OPAM package (hence 1.2.3 and 0.1 giving 1.2.3-0.1)? I've searched on OPAM and seen both bindings that do not include the original version and bindings that do. I think I prefer the latter, but it is not a strong opinion. What do you think?

kit-ty-kate:
I do not have strong opinions either. OCaml bindings are usually based on system installed libraries rather than vendored ones, so the version number is usually the one of the binding.

I didn't know that yours was vendored so in this case that makes sense. Thanks!

Niols:
I don't like the vendoring so much myself, but TestU01 is actually quite hard to find in system packages, hence my choice. Thanks for the merge!

PR #18760, opened 2021-05-30
mirage-crypto: Add missing lower-bound constraint

kit-ty-kate:
cpuid < 0.1.2 did not link with result and thus Result.result was an abstract type unless result is loaded)

Detected in https://github.com/ocaml/opam-repository/pull/18758


mseri:
Unless you think we should do something for cstruct.6.0.0 on x86_32 and cpuid on arm32, I think this can be merged

PR #18772, opened 2021-06-01
x509: Fix constraint (uses Base64.encode_string)

kit-ty-kate:
Detected in https://github.com/ocaml/opam-repository/pull/18769
cc @hannesm


hannesm:
thanks, I pushed this change upstream.

kit-ty-kate:
Thanks a lot!

PR #18777, opened 2021-06-02
Package rdf.0.12.0

zoggy:
### `rdf.0.12.0`
OCaml library to manipulate RDF graphs; implements SPARQL

---
* Homepage: https://www.good-eris.net/ocaml-rdf/
* Source repo: git+https://framagit.org/zoggy/ocaml-rdf.git
* Bug tracker: https://framagit.org/zoggy/ocaml-rdf/-/issues

---
:camel: Pull-request generated by opam-publish v2.0.3

PR #18779, opened 2021-06-02
genet.0.6 isn't compatible with OCaml >= 4.06 (uses unsafe strings)

kit-ty-kate:
Detected in https://github.com/ocaml/opam-repository/pull/18777
cc @zoggy


zoggy:
thanks !

PR #18787, opened 2021-06-03
Release dune-deps 1.3.0

mjambon:
Nothing special to report, only the tarball changed.


kit-ty-kate:
Thanks. Could you return the fix to the opam file upstream so that it can be taken into account for the next releases?

mjambon:
Nice catch, thank you! I'll propagate the change.

PR #18793, opened 2021-06-04
OCaml 4.13.0, first alpha release

Octachron:
This PR adds the ocaml-base-compiler package and ocaml-variants...+options package for the first alpha release of OCaml 4.13.0 .
I had to relax the constraints on the virtual ocaml package for 4.13.0 to make those two package installable.


kit-ty-kate:
Thanks!

Issue #18801, opened 2021-06-05
Error while installing 


MohammedRehabS:
Facing the below error while installing liquidsoap 1 by opam. Please provide a solution or workaround to get the latest version.

```
The following actions will be performed:
  ∗ install mm         0.7.1   [required by liquidsoap]
  ∗ install liquidsoap 1.3.3-1
===== ∗ 2 =====
Do you want to continue? [Y/n] Y

<><> Gathering sources ><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
[liquidsoap.1.3.3-1] found in cache
[mm.0.7.1] found in cache

<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
[ERROR] The compilation of mm failed at
        "/home/liquidsoap/.opam/opam-init/hooks/sandbox.sh build dune build -p
        mm -j 3 @install".

\#=== ERROR while compiling mm.0.7.1 ===========================================#
\# context     2.0.8 | linux/arm32 | ocaml-system.4.02.3 | https://opam.ocaml.org#7f5f038c
\# path        ~/.opam/default/.opam-switch/build/mm.0.7.1
\# command     ~/.opam/opam-init/hooks/sandbox.sh build dune build -p mm -j 3 @install
\# exit-code   1
\# env-file    \~/.opam/log/mm-18064-a71bb4.env
\# output-file \~/.opam/log/mm-18064-a71bb4.out
\### output ###
\# image_rgb.c:1111:22: warning: variable ‘aa’ set but not used [-Wunused-but-set-variable]
\# [...]
\#                       \^\~
\#       ocamlc external/.mm_alsa.objs/byte/mm_alsa.{cmo,cmt} (exit 2)
\# (cd _build/default && /usr/bin/ocamlc -w -40 -g -bin-annot -I external/.mm_alsa.objs/byte -I /home/liquidsoap/.opam/default/lib/alsa -I src/.mm_audio.objs/byte -I src/.mm_audio.objs/native -I src/.mm_base.objs/byte -I src/.mm_base.objs/native -intf-suffix .ml -no-alias-deps -o external/.mm_alsa.objs/byte/mm_alsa.cmo -c -impl external/mm_alsa.ml)
\# File "external/mm_alsa.ml", line 69, characters 22-45:
\# Error: Unbound value Alsa.Pcm.readn_float_ba
\# Hint: Did you mean readn_float64 or readn_float?
[...]

PR #18802, opened 2021-06-05
mm >= 0.6.0: Add missing constraint

kit-ty-kate:
Should fix https://github.com/ocaml/opam-repository/issues/18801
cc @toots @smimram


toots:
Thanks! This has been backported upstream!

PR #18804, opened 2021-06-05
[new release] current_slack, current_rpc, current_examples, current_docker, current_incr, current_git, current_web, current_github and current (0.5)

talex5:
Slack plugin for OCurrent

- Project page: <a href="https://github.com/ocurrent/ocurrent">https://github.com/ocurrent/ocurrent</a>

##### CHANGES:

Web UI:

- Space and color rows in query page (@MisterDA ocurrent/ocurrent#265).

- Duplicate the Rebuild selected button on top (@MisterDA ocurrent/ocurrent#265).

- Add a 'select-all' checkbox to the query page (@MisterDA ocurrent/ocurrent#265).

- Better printing of job duration (@MisterDA ocurrent/ocurrent#252, @ewanmellor ocurrent/ocurrent#259).

- Speed up rendering of the Query page (@talex5 ocurrent/ocurrent#262).

Windows support:

- Remove RO bit on Windows file when deleting tmpdir (@MisterDA ocurrent/ocurrent#270).

- Use `open_in_bin` instead of `open_in` to open log files (@MisterDA ocurrent/ocurrent#246).

- Support Windows file separators (@MisterDA ocurrent/ocurrent#246).

API:

- GitHub: add `staleness` argument to allow ignoring inactive branches (@patricoferris ocurrent/ocurrent#231).

- GitHub: add `default_ref` to get the default branch name (@patricoferris ocurrent/ocurrent#231).

- Git: expose `Current_git.Local.repo` (@ewanmellor ocurrent/ocurrent#257).
  This means that callers can use this for debugging or labelling.

Other:

- Log a message when cancelling a job finishes (@talex5 ocurrent/ocurrent#261).
  This avoids the confusing situation where the job finishes as you cancel, and you see the success message from the build as the last thing in the log, even though the job actually failed.

- Move ANSI parser to its own repository (@talex5 ocurrent/ocurrent#267, @MisterDA ocurrent/ocurrent#268).
  It's now at https://github.com/ocurrent/ansi

- `current_git` depends on `conf-git` (@MisterDA ocurrent/ocurrent#253).

- Update Dockerfile to latest opam-repository (@ewanmellor ocurrent/ocurrent#251).

- Fix typo in doc-strings (@tatchi ocurrent/ocurrent#250).

- Replace deprecated `Fmt` functions (@MisterDA ocurrent/ocurrent#245).

- Update to latest x509 API changes (@talex5 ocurrent/ocurrent#264).

- Use mirage-crypto for id-of-repo digests (@talex5 ocurrent/ocurrent#243).

### DIFF
--- a/packages/current/current.0.5/opam
+++ b/packages/current/current.0.5/opam
@@ -1,0 +1,58 @@
+opam-version: "2.0"
+synopsis: "Pipeline language for keeping things up-to-date"
+description: """\
+OCurrent provides an OCaml eDSL for writing CI/CD pipelines.
+
+It is used in ocaml-ci (which provides CI for OCaml projects on GitHub),
+and in docker-base-images (a pipeline that builds Docker images for various
+Linux distributions, OCaml compiler versions and CPU types, and pushes them
+to Docker Hub).
+
+A pipeline is written much like you would write a one-shot sequential script,
+but OCurrent will automatically re-run steps when the inputs change, and will
+run steps in parallel where possible."""
+maintainer: "talex5@gmail.com"
+authors: "talex5@gmail.com"
+homepage: "https://github.com/ocurrent/ocurrent"
+doc: "https://ocurrent.github.io/ocurrent/"
+bug-reports: "https://github.com/ocurrent/ocurrent/issues"
+depends: [
+  "ocaml" {>= "4.08.0"}
+  "current_incr" {= version}
+  "fmt" {>= "0.8.9"}
+  "bos"
+  "ppx_deriving"
+  "lwt" {>= "4.3.0"}
+  "cmdliner"
+  "sqlite3"
+  "duration"
+  "prometheus"
+  "dune" {>= "2.0"}
+  "re" {>= "1.9.0"}
+  "lwt-dllist"
+  "alcotest" {>= "1.2.0" & with-test}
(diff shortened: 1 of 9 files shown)


talex5:
The lower-bounds checker did:
```
downgrade seq                     base to 0.2.2
```
I think that caused the failures:
```
#        Type Re.Group.t Stdlib.Seq.t = unit -> Re.Group.t Stdlib.Seq.node
#        is not compatible with type Re__Core.Group.t Seq/1.t
```
Should non-base versions of seq conflict with OCaml >= 4.07?

kit-ty-kate:
The implementation of the `seq` library makes it an alias of `Stdlib.Seq` with OCaml >= 4.07, so this should be correct.
This is failing here because ocurrent uses `(implicit_transitive_deps false)` and `seq` is not part of the explicit dependencies so that causes the compiler to transform `Seq.t` to an abstract type.

I'm not sure what's the best thing to do here. Probably adding:
```
conflicts: [
  "seq" {< "base"} # enforced by (implicit_transitive_deps false)
]
```
should be good. Though maybe the `seq.base` package should install a META file itself

talex5:
OK, I added that. I'll remove the transitive dependencies thing upstream for next time.

kit-ty-kate:
Looks good now. Thanks!

PR #18810, opened 2021-06-06
Package gpiod.0.7

jawline:
### `gpiod.0.7`
A wrapper around the C libgpiod library for GPIO on recent (>4.8) Linux kernels

---
* Homepage: https://github.com/jawline/ocamlGpiod/
* Source repo: git+https://github.com/jawline/ocamlGpiod.git
* Bug tracker: https://github.com/jawline/ocamlGpiod/

---

### DIFF
--- a/packages/gpiod/gpiod.0.7/opam
+++ b/packages/gpiod/gpiod.0.7/opam
@@ -1,0 +1,46 @@
+opam-version: "2.0"
+synopsis:
+  "A wrapper around the C libgpiod library for GPIO on recent (>4.8) Linux kernels"
+maintainer: ["Blake Loring <blake@parsed.uk>"]
+authors: ["Blake Loring"]
+license: "BSD-3-Clause"
+homepage: "https://github.com/jawline/ocamlGpiod/"
+bug-reports: "https://github.com/jawline/ocamlGpiod/"
+depends: [
+  "dune" {>= "2.8"}
+  "re" {>= "1.9.0"}
+  "core" {>= "v0.14.1"}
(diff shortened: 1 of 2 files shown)


packages/gpiod/gpiod.0.7/opam
```diff
@@ -0,0 +1,41 @@
...
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
+dev-repo: "git+https://github.com/jawline/ocamlGpiod.git"
+depexts: [
+  "libgpiod-dev"
+]
+available: [ os = "linux" & os-family = "debian"  ]
```
mseri, line 34: Is gpiod debian specific or we could drop this line and add the relevant depexts?

jawline: Ah I really should - sorry I'm still getting the hang of depexts. I'll do some citation chasing for other distributions when I have a moment.

jawline: Not all distributions include it, I've marked all those I could fine with packages but I'm having some trouble eliminating oraclelinux from the build. Is there any table of all distributions os, os-distribution and os-family data anywhere? I can't install opam in docker on oraclelinux so it's not easy for me to do an adhoc check.

mseri:
@jawline, I moved the depexts on a conf-gpiod package, in this way it will be easier to update and can be used separately. Is that ok with you?

jawline:
That's great - thanks you for all your help.

mseri:
Thanks. I will update the depexts later on if I figure out what need to be added

PR #18814, opened 2021-06-07
lablqml.0.7

Kakadu:
CI should become less red after dealing with https://github.com/ocaml/opam-repository/pull/18526


Kakadu:
@kit-ty-kate We need your opinion about sedlex's CI failures.

kit-ty-kate:
https://github.com/ocaml/opam-repository/pull/18850 should fix that

packages/lablqml/lablqml.0.6.2/opam
```diff
@@ -1,11 +1,15 @@
 opam-version: "2.0"
 maintainer:   "kakadu.hafanana@gmail.com"
 authors:      "kakadu.hafanana@gmail.com"
+license: "LGPL-2.1-or-later with OCaml-LGPL-linking-exception"
 homepage:     "http://kakadu.github.io/lablqt/"
 bug-reports:  "https://github.com/kakadu/lablqml/issues"
 dev-repo: "git+https://github.com/Kakadu/lablqml.git"
 tags: [ "gui" "ui" "qt" ]
 
+# Marking as unavailable because on CI file system is readonly and configure fails
```
kit-ty-kate, line 10: Note that this is not the CI system but the sandbox in opam 2.1 that's not making it fail. The only reason why this worked in CI before is that the sandbox was not enabled and that we were testing only opam 2.0, but for users using 2.1 it was failing too.

packages/qocamlbrowser/qocamlbrowser.0.2.12/opam
```diff
@@ -0,0 +1,33 @@
...
+synopsis: "OCamlBrowser clone written with OCaml and QtQuick 2"
+
+build: [
+  ["sh" "-exc" "PATH=/usr/lib64/qt5/bin:/usr/lib/qt5/bin:$PATH ./configure"]
+    {os != "macos"}
+  ["sh" "-exc" "PATH=/usr/lib64/qt5/bin:/usr/lib/qt5/bin:$PATH make"]
+    {os != "macos"}
+  ["sh" "-exc" "PATH=/usr/local/opt/qt/bin:$PATH ./configure"] {os = "macos"}
+  ["sh" "-exc" "PATH=/usr/local/opt/qt/bin:$PATH make"] {os = "macos"}
+]
```
kit-ty-kate, line 17: I think it would be simpler and more maintainable to detect where the binaries you need are, in `conf-qt` and to use the configuration variable directly. (see the `conf-cmake` as a simple example and `conf-llvm` as a more developed one)

packages/lablqml/lablqml.0.6.2/opam
Kakadu: It looks like you don't insist on fixing this, but I rephrased that anyway.

packages/qocamlbrowser/qocamlbrowser.0.2.12/opam
Kakadu: I did it.

Kakadu: OKay, I should probably do the same here too.

Kakadu: Done!

Kakadu:
@kit-ty-kate It's ready for the 2nd round of review.

Kakadu:
An issue appeared for OCaml 4.13 and ppx_js_style. 
@aalekseyev, could you check is it mine or yours?

aalekseyev:
@Kakadu, looks like it's "ours" in the sense that our code needs a fix for compatibility with 4.13, but fixing all ppxes usually takes time.
Maybe this check is expected to be red  the current state of OPAM and you can ignore it for the purpose of your PR?

Kakadu:
> @Kakadu, looks like it's "ours" in the sense that our code needs a fix for compatibility with 4.13, but fixing all ppxes usually takes time.
> Maybe this check is expected to be red the current state of OPAM and you can ignore it for the purpose of your PR?

I will ignore it then

packages/lablqml/lablqml.0.6.2/opam
```diff
@@ -1,11 +1,16 @@
...
 authors:      "kakadu.hafanana@gmail.com"
+license: "LGPL-2.1-or-later with OCaml-LGPL-linking-exception"
 homepage:     "http://kakadu.github.io/lablqt/"
 bug-reports:  "https://github.com/kakadu/lablqml/issues"
 dev-repo: "git+https://github.com/Kakadu/lablqml.git"
 tags: [ "gui" "ui" "qt" ]
 
+# Marking as unavailable because opam 2.1 sandbox doesn't allow creating new files this way
+# As we have version of lablqml newer then 0.6.2, this issue is not going to be fixed.
+available: false
```
mseri, line 12: This is going to work with older opam versions right? If so, I would keep it available. @kit-ty-kate is there a way to set availability depending on opam?

packages/qocamlbrowser/qocamlbrowser.0.2.10/opam
```diff
@@ -3,9 +3,13 @@ maintainer: "kakadu.hafanana@gmail.com"
 homepage: "https://github.com/Kakadu/QOcamlBrowser_quick/"
 bug-reports: "https://github.com/Kakadu/QOcamlBrowser_quick/issues"
 dev-repo: "git://github.com/Kakadu/QOcamlBrowser_quick"
+license: "LGPL-2.1-or-later with OCaml-LGPL-linking-exception"
 authors: "Kakadu"
 synopsis: "OCamlBrowser clone written with OCaml and QtQuick 2"
 
+# Marking as unavailable because on CI file system is readonly and configure fails
+available: false
```
mseri, line 11: Is this just a CI issue? The package would become unavailable for everybody in this way

packages/qocamlbrowser/qocamlbrowser.0.2.11/opam
```diff
@@ -3,9 +3,13 @@ maintainer: "kakadu.hafanana@gmail.com"
 homepage: "https://github.com/Kakadu/QOcamlBrowser_quick/"
 bug-reports: "https://github.com/Kakadu/QOcamlBrowser_quick/issues"
 dev-repo: "git://github.com/Kakadu/QOcamlBrowser_quick"
+license: "LGPL-2.1-or-later with OCaml-LGPL-linking-exception"
 authors: "Kakadu"
 synopsis: "OCamlBrowser clone written with OCaml and QtQuick 2"
 
+# Marking as unavailable because on CI file system is readonly and configure fails
+available: false
```
mseri, line 11: As above

packages/qocamlbrowser/qocamlbrowser.0.2.9/opam
```diff
@@ -1,8 +1,13 @@
 opam-version: "2.0"
 maintainer: "kakadu.hafanana@gmail.com"
+license: "LGPL-2.1-or-later with OCaml-LGPL-linking-exception"
 homepage: "https://github.com/Kakadu/QOcamlBrowser_quick/"
 bug-reports: "https://github.com/Kakadu/QOcamlBrowser_quick/issues"
 authors: "Kakadu"
+
+# Marking as unavailable because on CI file system is readonly and configure fails
+available: false
```
mseri, line 9: As above

packages/lablqml/lablqml.0.6.2/opam
Kakadu: Yes, AFAIU it should work with older opam. The old release definitely passed CI some time ago. 
But because of old version is not widely used (never got any feedback about this), I think it is OK to mark it as unavailable.

mseri: To be frank, I would keep it available but add a postinstall-failure message saying that you need to disable the sandbox for this version to work. Same for qocamlbrowser

Kakadu: I applied changes that you requested.

mseri:
Looks good! Thanks a lot

PR #18264, opened 2021-03-04
Package comby.1.2.2

rvantonder:
### `comby.1.2.2`
A tool for structural code search and replace that supports ~every language
Comby is a tool designed to make it easy to match change syntax for basically
every language. You can think of it like grep or sed, but with added ability to
recognize code structures like blocks and expressions (e.g., delimited by
braces or parentheses), strings (e.g., delimited by quotes), and so on.

---
* Homepage: https://github.com/comby-tools/comby
* Source repo: git+https://github.com/comby-tools/comby.git
* Bug tracker: https://github.com/comby-tools/comby/issues

---
:camel: Pull-request generated by opam-publish v2.0.3

PR #18816, opened 2021-06-07
3 packages from comby-tools/comby at 1.7.0

rvantonder:
This pull-request concerns:
-`comby.1.7.0`: A tool for structural code search and replace that supports ~every language
-`comby-kernel.1.7.0`: A match engine for structural code search and replace that supports ~every language
-`comby-semantic.1.7.0`: A match engine for structural code search and replace that supports ~every language

---
* Homepage: https://github.com/comby-tools/comby
* Source repo: git+https://github.com/comby-tools/comby.git
* Bug tracker: https://github.com/comby-tools/comby/issues

---

### DIFF
--- a/packages/comby/comby.1.7.0/opam
+++ b/packages/comby/comby.1.7.0/opam
@@ -1,0 +1,62 @@
+opam-version: "2.0"
+maintainer: "rvantonder@gmail.com"
+authors: "Rijnard van Tonder"
+homepage: "https://github.com/comby-tools/comby"
+bug-reports: "https://github.com/comby-tools/comby/issues"
+dev-repo: "git+https://github.com/comby-tools/comby.git"
+license: "Apache-2.0"
+build: [
+    [
+      "dune"
+      "build"
+      "-p"
+      name
+      "-j"
+      jobs
+      "@install"
+      "@runtest" {with-test}
+    ]
+]
+depends: [
+  "dune" {>= "2.8.0"}
+  "ocaml" {>= "4.08.1"}
+  "conf-autoconf"
+  "conf-gmp"
+  "conf-libev" {os-distribution != "ol"}
+  "conf-libpcre"
+  "conf-m4" {os = "linux"}
+  "conf-pkg-config"
+  "conf-sqlite3"
+  "conf-zlib" {os = "linux"}
+  "cohttp-lwt-unix"
+  "comby-kernel" {= "1.7.0"}
+  "comby-semantic" {= "1.7.0"}
+  "core"
+  "hack_parallel" {arch != "arm32" & arch != "arm64"}
+  "lwt"
+  "lwt_react"
+  "mparser" {>= "1.3"}
+  "mparser-pcre"
+  "parany" {>= "12.0.3"}
+  "patience_diff" {>= "v0.14" & < "v0.15"}
+  "ppx_deriving"
+  "ppx_deriving_yojson" {>= "3.6.0"}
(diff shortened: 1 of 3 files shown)


rvantonder:
Messed up a vendored dependency, will update and try again.

rvantonder:
@kit-ty-kate not urgent, more of a question: only `opam-2.1` builds fail and `opam-2.0` succeeds without problems. I must be missing something that changed there? Should I be trying to reproduce locally with `opam 2.1`?

This PR also has a failing build for `conf-libev` on `oraclelinux-7`. Not sure why that's happening, it reminds me of this previous comment: https://github.com/ocaml/opam-repository/pull/18264#issuecomment-791592346. I'm fine with adding a constraint that doesn't support `oraclelinux`, if that helps.

rvantonder:
> Should I be trying to reproduce locally with opam 2.1?

I installed `opam 2.1` and had no problems locally. I see in the build output there is some dependency conflicts (dune?) that cascades, but it's not obvious what's going wrong here 😅 . [example](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/5db9f22e5a515806ab913621b41adb0bd1e4739b/variant/opam-2.1,compilers,4.08,comby.1.7.0,lower-bounds). I will try do a fresh push.

rvantonder:
OK I think I noticed an issue. The lower-bounds build is triggering:

>  - install   jbuilder                1.0+beta20.1         [required by ppx_derivers, cppo, bigstringaf, etc.]

Definitely don't want a dependency on `jbuilder`, and it looks like that's coming from `conduit-lwt` and possibly others. Will test and add some constraints.

kit-ty-kate:
I don't have full access to a computer to do the correct change right now but the issue comes from a missing lower-bound constraint on parany in comby. You just need to add the correct one. jbuilder has nothing to do with it, what made you think this?

rvantonder:
OK thanks! I do see now that the lower bound for parany does a downgrade which probably causes that issue. I'll give that a shot. I find a lot of the other unconstrained lower bounds suspicious though, and `jbuilder` just popped up as something where I know about as being deprecated and went "uh, that's really old I'm not sure I should depend on a lower bound that relies on `jbuilder`".

rvantonder:
@kit-ty-kate things look good now, I worked through the packages giving issues, thanks. I also added 

>   "conf-libev" {os-distribution != "ol"}

because the build kept breaking on oraclelinux 7 due to a libev issue, and I don't mind not supporting `oraclelinux` here.

packages/comby/comby.1.7.0/opam
```diff
@@ -0,0 +1,62 @@
...
+  "conf-autoconf"
+  "conf-gmp"
+  "conf-libev" {os-distribution != "ol"}
+  "conf-libpcre"
+  "conf-m4" {os = "linux"}
+  "conf-pkg-config"
+  "conf-sqlite3"
+  "conf-zlib" {os = "linux"}
+  "cohttp-lwt-unix"
+  "comby-kernel" {= "1.7.0"}
```
kit-ty-kate, line 32: For next time know that you can replace `{= "1.7.0"}` by `{= version}` for simplicity.

kit-ty-kate:
Thanks!

packages/comby/comby.1.7.0/opam
rvantonder: Niiice, will do!

PR #18828, opened 2021-06-08
Add a version constraint on ppxlib in the previous release of visitors.

fpottier:
The next release of `ppxlib` makes the type `Parser.token` abstract and is not compatible with visitors.20210316.


avsm:
Looks good, thanks.

PR #18838, opened 2021-06-09
[new release] fromager (0.4.0)

mimoo:
A CLI to format an ocaml codebase

- Project page: <a href="https://github.com/mimoo/fromager">https://github.com/mimoo/fromager</a>

##### CHANGES:

Remove version constraint on ocamlformat so that one can use whatever ocamlformat version they want.


kit-ty-kate:
Thanks!

PR #18842, opened 2021-06-10
[new release] capnp-rpc-mirage, capnp-rpc-net, capnp-rpc, capnp-rpc-unix and capnp-rpc-lwt (1.2)

talex5:

##### CHANGES:

- Don't crash if the peer disconnects before the bootstrap reply is ready (@talex5 mirage/capnp-rpc#234).
  When replying to a normal question we checked that the answer was still needed before trying to reply, but didn't for bootstrap requests.

- Replace `wait_until_settled` with safer `await_settled` (@talex5 mirage/capnp-rpc#233).
  `wait_until_settled` makes it too easy to ignore errors.

- Require fmt >= 0.8.7 for `kstr` and set OCaml 4.08 as the minimum version everywhere (@talex5 mirage/capnp-rpc#232).
  Older versions aren't tested in CI because the unix and mirage packages require at least OCaml 4.08.

### DIFF
--- a/packages/capnp-rpc-lwt/capnp-rpc-lwt.1.2/opam
+++ b/packages/capnp-rpc-lwt/capnp-rpc-lwt.1.2/opam
@@ -1,0 +1,39 @@
+opam-version: "2.0"
+synopsis:
+  "Cap'n Proto is a capability-based RPC system with bindings for many languages"
+description: """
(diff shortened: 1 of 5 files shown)


kit-ty-kate:
Thanks a lot!

PR #18846, opened 2021-06-10
+faraday.0.7.3

seliopou:


kit-ty-kate:
Thanks! Could you return the fixes to the opam files upstream so that they get taken into account for the future releases?

PR #18849, opened 2021-06-10
Release Caqti 1.6.0.

paurkedal:
I have accumulated some bug fixes and a few minor improvements in need of a release. From the change log:

- Set the time zone of PostgreSQL connections to UTC to mitigate an
  undesirable implicit conversion to the local time zone for `timestamp`.
  This issue was exposed by the specification of field types introduced in
  version 1.4.0.  Earlier versions worked as expected, if only accidentally,
  since the time zone is ignored when a string is converted to a
  `timestamp`.  While this change makes `timestamp` more usable again for
  storing UTC time stamps, I strongly recommend using `timestamp with time
  zone` since it's interpretation is unambiguous.  The API reference is now
  updated with details about how the `ptime` OCaml type is mapped for
  different database systems.
- Drop specification of OCaml `string` as SQL `text` for PostgreSQL. This is
  due to issues with implicit conversions and function overloading when the
  desired type on the SQL side is `char`, `varchar`, or `jsonb`.
- Add `Caqti_type.redact` to protect sensitive information from being
  logged.
- Only log parameters if `$CAQTI_DEBUG_PARAM` is set to "`true`".
- When logging requests, show underlying values for custom types.
- Reject multi-row response in `find_opt` implementation for sqlite3.
- Tolerate Lwt promise rejections in `Pool.use`.

### DIFF
--- a/packages/caqti-async/caqti-async.1.2.0/opam
+++ b/packages/caqti-async/caqti-async.1.2.0/opam
@@ -1,7 +1,7 @@
 opam-version: "2.0"
 maintainer: "Petter A. Urkedal <paurkedal@gmail.com>"
 authors: "Petter A. Urkedal <paurkedal@gmail.com>"
-license: "LGPL-3 with OCaml linking exception"
+license: "LGPL-3.0-or-later with OCaml-LGPL-linking-exception"
 homepage: "https://github.com/paurkedal/ocaml-caqti/"
 doc: "http://paurkedal.github.io/ocaml-caqti/index.html"
 bug-reports: "https://github.com/paurkedal/ocaml-caqti/issues"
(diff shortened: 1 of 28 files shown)


paurkedal:
I added a version constraint to cppo to all caqti.opam files. Seems to have missed it earlier.

paurkedal:
Thanks for the help with fixing reverse deps.  Shall I rebase on master?

kit-ty-kate:
Thanks! Sorry for the delay

kit-ty-kate:
Could you return the fixes to the opam file upstream if needed so that it can be taken into account for next time?

paurkedal:
Thanks for also updating the license lines in 1.2.*.  The ones in the Caqti repo are fixed already.

PR #18855, opened 2021-06-11
ocamlformat: Add missing constraints

kit-ty-kate:
Detected in https://github.com/ocaml/opam-repository/pull/18838

PR #18856, opened 2021-06-12
uucp: Add missing constraint

kit-ty-kate:
Uses --dev-pkg
Detected in https://github.com/ocaml/opam-repository/pull/18855
cc @dbuenzli


dbuenzli:
Thanks for the ping @kit-ty-kate.

PR #18859, opened 2021-06-13
add two ocaml variants 4.06.1+rescript/4.10.2+rescript for upgrade

bobzhang:
Tested on my local machine, works fine


packages/ocaml-variants/ocaml-variants.4.10.2+rescript/opam
```diff
@@ -0,0 +1,45 @@
+opam-version: "2.0"
+synopsis: "4.10.2 release with specific featres for ReScript development"
```
avsm, line 2:
```suggestion
synopsis: "4.10.2 release with specific features for ReScript development"
```

avsm:
This PR looks fine, but in general for compilers we prefer the variant to describe the feature being changed (e.g. whatever the "conditional compilation" is). If you could do that in future PRs with a description of the changes you are introducing it'll make the compiler variant a bit more easily reusable by others.

PR #18861, opened 2021-06-15
Package unison.2.51.4~rc3

kit-ty-kate:
### `unison.2.51.4~rc3`
File-synchronization tool for Unix and Windows
Unison is a file-synchronization tool for Unix and Windows. It allows
two replicas of a collection of files and directories to be stored on
different hosts (or different disks on the same host), modified
separately, and then brought up to date by propagating the changes in
each replica to the other.

---
* Homepage: https://www.cis.upenn.edu/~bcpierce/unison/
* Source repo: git://github.com/bcpierce00/unison.git
* Bug tracker: https://github.com/bcpierce00/unison/issues

---

### DIFF
--- a/packages/unison/unison.2.51.4~rc3/opam
+++ b/packages/unison/unison.2.51.4~rc3/opam
@@ -1,0 +1,30 @@
+opam-version: "2.0"
+synopsis: "File-synchronization tool for Unix and Windows"
+description: """\
+Unison is a file-synchronization tool for Unix and Windows. It allows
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
@gdt if CI succeeds would you be interested to have unison in opam-repository? For the moment this is only a draft (given this is only an RC and all) so it does not have to be merged if you think this should not be.

kit-ty-kate:
Nevermind: https://github.com/bcpierce00/unison/issues/538

PR #18866, opened 2021-06-15
Package ocaml-monadic.0.5

zepalmer:
### `ocaml-monadic.0.5`
A PPX extension to provide an OCaml-friendly monadic syntax

---
* Homepage: http://github.com/zepalmer/ocaml-monadic
* Source repo: git+https://github.com/zepalmer/ocaml-monadic.git
* Bug tracker: https://github.com/zepalmer/ocaml-monadic/issues

---

### DIFF
--- a/packages/ocaml-monadic/ocaml-monadic.0.5/opam
+++ b/packages/ocaml-monadic/ocaml-monadic.0.5/opam
@@ -11,7 +11,7 @@ depends: [
   "ppxlib" {>= "0.18.0"}
 ]
 build: [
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
To fix the `"dune" "subst" {pinned}` bit you need to upgrade to `(lang dune 2.7)` as dune will only generate the correct opam file with this prerequisite.

kit-ty-kate:
Apart from that everything looks good. Thanks!

PR #18867, opened 2021-06-15
Dream alpha2: drop ppxlib constraints

aantron:
Now that ppxlib 0.21.1 and 0.22.1 are out, the constraints are no longer necessary. They were meant to exclude 0.21.0 and 0.22.0, but it should be enough to just rely on the .1 versions being installed, in most cases.


kit-ty-kate:
Thanks!

PR #18872, opened 2021-06-16
opam-devel < 2.1.0~rc: Disable tests

kit-ty-kate:
non-deterministic based on environment variables
Detected in https://github.com/ocaml/opam-repository/pull/18871
cc @rjbou


rjbou:
I confirm, part of the test uses an archive from opam repository.

PR #18876, opened 2021-06-16
[new release] repr-bench, repr, ppx_repr and repr-fuzz (0.4.0)

CraigFe:
(Apologies for the PR noise; `dune-release` did something weird the first time...)

- Project page: <a href="https://github.com/mirage/repr">https://github.com/mirage/repr</a>
- Documentation: <a href="https://mirage.github.io/repr">https://mirage.github.io/repr</a>

##### CHANGES:

- Add `Repr.{random,random_state}`, a pair of generic functions for sampling
  random instances of representable types. (mirage/repr#58, @CraigFe)

- Add `Repr.Size`, which provides sizing functions for binary codecs that are
  more informative than the existing `Repr.size_of`. Types built using `Repr.v`
  and `Repr.like` must now pass a sizer built using `Repr.Size.custom_*`. (mirage/repr#69,
  @CraigFe)


CraigFe:
The failures here are unrelated to Repr, but happen via Irmin:

 - Irmin uses Lwt, which pulls Result for pre-4.03 compatibility
 - when on the "lower bounds" mode, the CI tries to pull `result.1.0`;
 - this shadows Irmin's usage of the stdlib's `Result` module, which is otherwise fine as Irmin requires 4.08. `result.1.5` fixed this issue by making `Result` a conditional alias.

I'm not sure what the correct solution is for that: I'd rather not add an explicit lower bound on `result` in Irmin, since we don't ever require it. Perhaps Lwt's usage of `result` should be conditional: https://github.com/ocsigen/lwt/issues/794.

kit-ty-kate:
All green! :tada:

PR #18885, opened 2021-06-18
new package orf v1.0.0

UnixJunkie:


packages/orf/orf.1.0.0/opam
```diff
@@ -0,0 +1,60 @@
...
+homepage: "https://github.com/UnixJunkie/orf"
+bug-reports: "https://github.com/UnixJunkie/orf/issues"
+dev-repo: "git+https://github.com/UnixJunkie/orf.git"
+license: "LGPL w/ linking exception"
+build: ["dune" "build" "-p" name "-j" jobs]
+depends: [
+  "batteries"
+  "cpm"
+  "dolog"
+  "dune" {>= "1.10"}
```
mseri, line 13:
```suggestion
  "dune" {>= "2.8"}
```
The dune-project file is requiring 2.8.
I think a lower bound is needed in cpm or minicli, at least the lower bound checker is failing to compile in some cases.

UnixJunkie:
@mseri Isn't there a tool to automatically find the lower version constraints bounds? That would help a lot.

mseri:
I don’t think so, it would be very useful indeed but quite hard to implement imho. Often I have to go digging in repos history to figure that out, but sometimes some authors have good changelogs

UnixJunkie:
I can hack a shell script, but I will be slow to run I guess.

UnixJunkie:
I added two version constraints, maybe enough to go further in the CI tests.

packages/orf/orf.1.0.0/opam
```diff
@@ -0,0 +1,60 @@
+opam-version: "2.0"
+authors: "Francois Berenger"
+maintainer: "unixjunkie@sdf.org"
+homepage: "https://github.com/UnixJunkie/orf"
+bug-reports: "https://github.com/UnixJunkie/orf/issues"
+dev-repo: "git+https://github.com/UnixJunkie/orf.git"
+license: "LGPL w/ linking exception"
```
mseri, line 7: Would
```suggestion
license: "LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception"
```
Be correct? I could not find the license file in the repo (did not search too hard though)

mseri:
Almost there. I don’t know where the A.fold comes from, but hopefully that is the last lower bound to add

UnixJunkie:
A.fold is very probably BatArray.fold (alias for fold_left, since batteries 3.2.0).

packages/orf/orf.1.0.0/opam
```diff
@@ -0,0 +1,60 @@
...
+authors: "Francois Berenger"
+maintainer: "unixjunkie@sdf.org"
+homepage: "https://github.com/UnixJunkie/orf"
+bug-reports: "https://github.com/UnixJunkie/orf/issues"
+dev-repo: "git+https://github.com/UnixJunkie/orf.git"
+license: "LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception"
+build: ["dune" "build" "-p" name "-j" jobs]
+depends: [
+  "batteries" {>= "3.2.0"}
+  "cpm"
```
mseri, line 11:
```suggestion
  "cpm" { >= "5.0.0"}
```

UnixJunkie:
Using opam binary packages, a tool looking automatically at the lowest compatible bound for each dep would be feasible.
From sources and recompiling many things each time, the performance of such tool is unacceptable.

mseri:
Good to go now! Can you please upstream these lower bounds?

UnixJunkie:
ok, thanks for the help.

PR #18886, opened 2021-06-18
bump dolog version constraint in oplsr.8.0.1

UnixJunkie:
so that it can co-install with other recent things

PS: I tested locally; it can still compile

### DIFF
--- a/packages/oplsr/oplsr.8.0.1/opam
+++ b/packages/oplsr/oplsr.8.0.1/opam
@@ -12,7 +12,7 @@ depends: [
   "conf-gnuplot"
   "conf-r"
   "cpm"
-  "dolog" {>= "6.0.0"}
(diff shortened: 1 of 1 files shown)


mseri:
Thanks! Can you move this change to `opslr.8.0.1-1` and make it less restrictive, like
```
"dolog" {(>= "4.0.0" & < "5.0.0") | (>= "6.0.0")}
```
?

UnixJunkie:
I weakened the version constraint, after some local testing.

mseri:
Thanks. The installation failure is the usual issue with R packages, right?

UnixJunkie:
Probably:
```
[ERROR] The installation of oplsr failed at "R CMD BATCH install_pls.r".
```
I tested locally though, I could install the package and R dependencies.

PR #18888, opened 2021-06-18
[new release] elpi (1.13.6)

gares:
CHANGES:

- API:
  - Fix `std.findall` is now calling a builtin which is able to produce
    solutions which contain eigenvariables and uvars as well.
  - `loc` is now printed using `/` as a path separator also on Windows
  - `loc.fields` to project a `loc` into the file, line, etc...


kit-ty-kate:
Thanks!

PR #18901, opened 2021-06-21
Adds cpdf v2.4

johnwhitington:


packages/cpdf/cpdf.2.4/opam
```diff
@@ -0,0 +1,19 @@
+opam-version: "2.0"
+maintainer: "contact@coherentgraphics.co.uk"
+build: [[make]]
+depends: [
+  "ocaml" {>= "4.10.0"}
+  "ocamlfind" {build}
+  "camlpdf" {= "2.4"}
```
mseri, line 7: If they always have to have the same version you can also use 
```suggestion
  "camlpdf" {= "version}
```

johnwhitington:
Thanks. That's a nice feature.

packages/cpdf/cpdf.2.4/opam
```diff
@@ -0,0 +1,19 @@
+opam-version: "2.0"
+maintainer: "contact@coherentgraphics.co.uk"
```
mseri, line 2:
```suggestion
maintainer: "contact@coherentgraphics.co.uk"
license: "Coherent Graphics Ltd Non-Commercial Use License Agreement"
```

mseri:
I am not sure why it is failing on archlinux. Do you have any idea?

mseri:
You can find more details here: https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/2cd56a1c630a4fe853e7581ebd932324b8644cdf/variant/opam-2.1,distributions,archlinux,cpdf.2.4

johnwhitington:
It's bizarre. The warning 58s should not be there at all - the cmx files from CamlPDF are all installed.

The final error is:

```
- Error: Cannot find file /home/opam/.opam/4.12/lib/camlpdf/camlpdf.cmxa
```

If that's true, CamlPDF hasn't been installed at all.

I assume it works on other linux configurations?

johnwhitington:
Your license change is fine, by the way.

mseri:
Yes, it works fine on the other configurations. @kit-ty-kate do you by chance know what is wrong in archlinux?

avsm:
Arch tends to use newer version of GNU tools, so I suspect something around a GNU make upgrade or similar.

mseri:
Thanks. I am going to merge, this seems more an issue between arch and campldf than cpdf problem

PR #18903, opened 2021-06-21
opam-grep: Refine availibility

kit-ty-kate:
opam-grep needs some sort of posix environment. MacOS should work and Cygwin should probably work too


mseri:
There are spurious failures due to CI overload I guess

PR #18905, opened 2021-06-21
[new release] voqc (0.2.1)

khieta:
A verified optimizer for quantum circuits (VOQC)

- Project page: <a href="https://github.com/inQWIRE/mlvoqc">https://github.com/inQWIRE/mlvoqc</a>
- Documentation: <a href="https://inQWIRE.github.io/mlvoqc">https://inQWIRE.github.io/mlvoqc</a>

##### CHANGES:

- Updated to the most recent version of SQIR & extracted using Coq v8.13.2.
- The only change in SQIR that impacts the behavior of mlvoqc is that Optimize1qGates now merges as many gates as possible instead of just two at a time.

### DIFF
--- a/packages/voqc/voqc.0.2.1/opam
+++ b/packages/voqc/voqc.0.2.1/opam
@@ -1,0 +1,41 @@
+opam-version: "2.0"
+synopsis: "A verified optimizer for quantum circuits (VOQC)"
+description:
+  "OCaml library for VOQC, a Verified Optimizer for Quantum Circuits"
+maintainer: ["kesha@cs.umd.edu"]
+authors: ["Kesha Hietala"]
+license: "MIT"
+homepage: "https://github.com/inQWIRE/mlvoqc"
+doc: "https://inQWIRE.github.io/mlvoqc"
+bug-reports: "https://github.com/inQWIRE/mlvoqc/issues"
+depends: [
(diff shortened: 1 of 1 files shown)


packages/voqc/voqc.0.2.1/opam
```diff
@@ -0,0 +1,41 @@
...
+authors: ["Kesha Hietala"]
+license: "MIT"
+homepage: "https://github.com/inQWIRE/mlvoqc"
+doc: "https://inQWIRE.github.io/mlvoqc"
+bug-reports: "https://github.com/inQWIRE/mlvoqc/issues"
+depends: [
+  "dune" {>= "2.7"}
+  "ocaml" {>= "4.08"}
+  "openQASM"
+  "zarith"
```
mseri, line 15:
```suggestion
  "zarith" {>= "1.5"}
```

kit-ty-kate: For posterity: this package uses `Q.to_float` which was added in 1.5

kit-ty-kate:
Thanks! Could you return the fix to the opam file upstream so this can be taken into account for the future releases?

khieta:
Thanks! I've fixed the version requirement upstream.

PR #18909, opened 2021-06-22
[new release] codept (0.11.0)

Octachron:
This is a new version of codept, an alternative dependency analyzer for OCaml projects.

This new version restores the support for recent compiler (up to 4.13.0). 

It also extends codept to handle every module-level features of OCaml (including abstract modules types). 

Finally, the internal core of codept (the interruptible outliner for OCaml module system) has been entirely switched to the zipper-based engine. 

## Features

  * Support for abstract module types:
	Codept is now aware that `Y` is `X.M.Y` and not an external dependency in

```ocaml
  module F(X:sig module type t module M:t end) = X
  module X = struct
    module M = struct
      module Y = struct end
    end
    module type t = module type of M
  end
  open F(X)
  open M
  open Y
```

  * Support for OCaml 4.09, 4.10, 4.11, 4.12, 4.13

  * Support for split compilation units: the interface and  implementation files do not need to share a filesystem path to be identified as part of the same compilation unit

## Bug fix

 * Nested with constraints `with A.B. ...` triggers now alias dependencies in -no-alias-deps mode
In other words, in this example
```ocaml
module type s = sig module Alias = Deps end
module E = struct end
module type s2 = s with module Deps.Sub = E
```
`Deps` should be counted as an external dependency due to left hand side of the substitution.

* The signature of dune wrapped library can now be imported properly: libraries are now read in the `-no-alias-deps` mode by default.

## Internal

* Switch to the new zipper-based outliner
* Fully typed recursive definition for schema
* Extended testsuite
* More detailled AST for the simplified value level (no more encoding)

### DIFF
--- a/packages/codept/codept.0.11.0/opam
+++ b/packages/codept/codept.0.11.0/opam
@@ -1,0 +1,35 @@
+opam-version: "2.0"
+maintainer: "Florian Angeletti <octa@polychoron.fr>"
+homepage: "https://github.com/Octachron/codept"
+bug-reports: "https://github.com/Octachron/codept/issues"
+license: "GPL-3.0-or-later"
+dev-repo: "git+https://github.com/Octachron/codept.git"
+build: [
+  ["dune" "build" "-p" name]
+]
+run-test: [
+  ["dune" "runtest"]
+]
+depends: ["dune" {>="2.5"} "menhir" {build & >="20180523" }  "ocaml" {>="4.03" & < "4.14~"}]
+synopsis: "Alternative ocaml dependency analyzer"
+description:"""
+Codept intends to be a dependency solver for OCaml project and an alternative to ocamldep. Compared to ocamldep, codept major features are:
+
+ * whole project analysis
+ * exhaustive warning and error messages
+ * structured format (s-expression or json) for dependencies
+ * uniform handling of delayed alias dependencies
+ * (experimental) full dependencies,
+  when dependencies up to transitive closure are not enough
+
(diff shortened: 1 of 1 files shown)


Octachron:
A type error that only appears with `dune -p` and without conjunctive types? That is unexpected.

packages/codept/codept.0.11.0/opam
```diff
@@ -0,0 +1,35 @@
...
+bug-reports: "https://github.com/Octachron/codept/issues"
+license: "gpl-3"
+dev-repo: "git+https://github.com/Octachron/codept.git"
+build: [
+  ["dune" "build" "-p" name]
+]
+run-test: [
+  ["dune" "runtest"]
+]
+depends: ["dune" "menhir" {build} "ocaml" {>="4.03" & < "4.14~"}]
```
mseri, line 13:
```suggestion
depends: [
  "ocaml" {>="4.03" & < "4.14~"}
  "dune" {>= "2.7"}
  "menhir" {build}
  "ocamlfind" {with-test}
]
```
If I looked correctly, the dune-project file requires dune 2.7 and the tests call ocamlfind

mseri:
It can happen when you vendor libraries, which does not seem the case here. I don't know.

packages/codept/codept.0.11.0/opam
Octachron: I am going to relax the dune requirement to 2.5 . Similarly, I will update the test to call `ocamlc -where` rather than ocamlfind to locate `compiler-libs`.

Octachron:
Here, it is the GADT-heavy code for the dynamic type representation that is hitting a pre-4.06 bug ... which is avoided when the `unused-module` warning is enabled (??!). I am thus just going to enable the warning in the release profile.

mseri:
Wow! 🙄

mseri:
Looks good now, thanks!

PR #18913, opened 2021-06-23
yaml: fix lower bounds on ctypes

samoht:


samoht:
The ezjsonm errors should be fixed by https://github.com/ocaml/opam-repository/pull/18931

mseri:
This one seems fine indeed. 
We need to have a look at `yices2_bindings` which fails with
```
File "src/yices2_low.ml", line 13, characters 19-29:
# 13 |   let equal_sint = SInt.equal
#                         ^^^^^^^^^^
# Error: Unbound value SInt.equal
```
(no idea where it comes from), but this can be dealt with separately

packages/integers/integers.0.1.0/opam
```diff
@@ -13,7 +13,7 @@ build:
            "--pinned" "%{pinned}%"]]
 depends: [
   "ocaml"
-  "ocamlbuild" {build}
+  "ocamlbuild" {build & >= "0.9.1"}
```
kit-ty-kate, line 16: This should be `& != "0.9.0"` as `ocamlbuild.0` is the only version available in OCaml < 4.03

mseri: oh, sorry

kit-ty-kate: There is no easy button to revert one commit quickly and the whole PR could do with a few explanation of why they were failing in the first place. I'll split them all in several PRs coming right up.

mseri: Fair point, thanks

packages/jekyll-format/jekyll-format.0.2.0/opam
```diff
@@ -18,7 +18,7 @@ depends: [
   "rresult"
   "ptime"
   "fpath"
-  "yaml"
+  "yaml" {>= "2.0.0"}
```
kit-ty-kate, line 21: This was only required for the tests, it was perfectly fine before without any constraints. This should've been:
```
  "yaml"
  "yaml" {with-test & >= "2.0.0"}
```

packages/ppx_deriving_yaml/ppx_deriving_yaml.0.1.0/opam
```diff
@@ -14,7 +14,7 @@ depends: [
   "ezjsonm" {with-test}
   "ocaml" {>= "4.08.1"}
   "ppxlib" {>= "0.14.0"}
-  "yaml"
+  "yaml" {>= "2.0.0"}
```
kit-ty-kate, line 17: same for this one

packages/user-agent-parser/user-agent-parser.0.2.0/opam
```diff
@@ -16,7 +16,7 @@ depends: [
   "ppx_let"
   "stdio" {>= "v0.11"}
   "re" {>= "1.8.0"}
-  "yaml" {build & >= "0.2.1"}
+  "yaml" {build & >= "2.0.0"}
```
kit-ty-kate, line 19: same for the two above

PR #18914, opened 2021-06-23
graphql-cohttp: use a recent version of yojson

samoht:


mseri:
The remaining issue needs to be fixed separately: `graphql_parser 0.9.0` uses menhir 2.0 but requires dune 1.2.0. However menhir 2.0 is supported only since dune 1.4:
I am going to open a separate PR

PR #18932, opened 2021-06-25
update graphql_parser lower bounds on dune

mseri:
It uses menhir lang 2.0 which was introduced only in dune 1.4


mseri:
Seen in https://github.com/ocaml/opam-repository/pull/18914

mseri:
Looks like there is something wrong with result, maybe due to
`  - downgrade result                   1.5 to 1.0`
@kit-ty-kate you mentioned a solution for this in the past, or I recall it wrong?

kit-ty-kate:
> Looks like there is something wrong with result, maybe due to
> ` - downgrade result 1.5 to 1.0`
> @kit-ty-kate you mentioned a solution for this in the past, or I recall it wrong?

You're probably thinking of the `conflicts: ["result" {< "1.5"}]`. However here this is a different issue: https://github.com/dbuenzli/fmt/compare/v0.8.1...v0.8.2 ^^"

Hopefully my last commit fixed the issue

mseri:
Indeed, thanks for looking into it and the fix

kit-ty-kate:
thanks a lot!

PR #18943, opened 2021-06-25
Tezos 9.3

pirbo:
Third bugfix release of the version 9 of the to be renamed `octez` software suite :-)


avsm:
This mostly all looks fine. One package uses Uri_re and has a lower bound failure but that can be fixed in a subsequent PR. Are you happy with all the CI results @pirbo?

kit-ty-kate:
> One package uses Uri_re and has a lower bound failure but that can be fixed in a subsequent PR

The `Uri_re` issue is a known non-deterministic failure in older versions of re. I'm not sure this can be fixed by `-j1` but maybe.

Looks good. Thanks!

PR #18944, opened 2021-06-25
[new release] uring (0.1)

talex5:
OCaml bindings for Linux io_uring

- Project page: <a href="https://github.com/ocaml-multicore/ocaml-uring">https://github.com/ocaml-multicore/ocaml-uring</a>

##### CHANGES:

- Initial release.

### DIFF
--- a/packages/uring/uring.0.1/opam
+++ b/packages/uring/uring.0.1/opam
@@ -1,0 +1,54 @@
+opam-version: "2.0"
+synopsis: "OCaml bindings for Linux io_uring"
+description:
+  "Bindings to the Linux io_uring kernel IO interfaces. See https://github.com/ocaml-multicore/eio for a higher-level API using this."
+maintainer: ["anil@recoil.org"]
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
thanks! Could you return the fix upstream?

talex5:
OK. Is there some documentation on the format of the license field? I had an idea we were using SPDX, so I used that syntax (https://spdx.github.io/spdx-spec/appendix-IV-SPDX-license-expressions/).

I would be good to add something to https://opam.ocaml.org/doc/Packaging.html#Publishing and https://github.com/ocaml/opam-repository/blob/master/CONTRIBUTING.md about the format of the license field.

talex5:
Also, this format doesn't seem to be accepted by dune:

- `(license "ISC" "MIT")` gives `Error: Too many argument for license`.
- `(license ("ISC" "MIT"))` gives `Error: Atom or quoted string expected`.

PR #18945, opened 2021-06-25
yaml: use the right versions of ctypes

samoht:
otherwise it fails with:

```
 File "types/bindings/yaml_bindings_types.mli", line 52, characters 13-24:
 52 | module M(F : Ctypes.TYPE) : sig
                   ^^^^^^^^^^^
 Error: Unbound module type Ctypes.TYPE
```
and:

```
 File "lib/stream.ml", line 215, characters 12-35:
 215 |   let buf = Ctypes.CArray.of_string str in
                   ^^^^^^^^^^^^^^^^^^^^^^^
 Error: Unbound value Ctypes.CArray.of_string
```


kit-ty-kate:
Thanks a lot!

PR #18947, opened 2021-06-26
Add menhir lowerbound for liquidsoap 1.4.x

toots:


mseri:
Thanks a lot! Can you also add the license field?

toots:
> Thanks a lot! Can you also add the license field? It should be
> 
> ```
> license: "GPL-2.0-or-later"
> ```
> 
> Right? I can also add it, if you confirm that it is correct.

Yes, correct. Just added it!

mseri:
Thanks a lot!

PR #18970, opened 2021-06-29
Package camlp-streams.5.0

xavierleroy:
### `camlp-streams.5.0`
The Stream and Genlex libraries for use with Camlp4 and Camlp5
This package provides two library modules:
- Stream: imperative streams, with in-place update and memoization
  of the latest element produced.
- Genlex: a small parameterized lexical analyzer producing streams
  of tokens from streams of characters.

The two modules are designed for use with Camlp4 and Camlp5:
- The stream patterns and stream expressions of Camlp4/Camlp5 consume
  and produce data of type 'a Stream.t.
- The Genlex tokenizer can be used as a simple lexical analyzer for
  Camlp4/Camlp5-generated parsers.

The Stream module can also be used by hand-written recursive-descent
parsers, but is not very convenient for this purpose.

The Stream and Genlex modules have been part of the OCaml standard library
for a long time, and have been distributed as part of the core OCaml system.
They will be removed from the OCaml standard library at some future point,
but will be maintained and distributed separately in this camlpstreams package.

---
* Homepage: https://github.com/ocaml/camlp-streams
* Source repo: git+https://github.com/ocaml/camlp-streams.git
* Bug tracker: https://github.com/ocaml/camlp-streams/issues

---

### DIFF
--- a/packages/camlp-streams/camlp-streams.5.0/opam
+++ b/packages/camlp-streams/camlp-streams.5.0/opam
@@ -29,6 +29,7 @@ maintainer: [
 authors: ["Daniel de Rauglaudre" "Xavier Leroy"]
 homepage: "https://github.com/ocaml/camlp-streams"
 bug-reports: "https://github.com/ocaml/camlp-streams/issues"
+license: "LGPL-2.1-only with OCaml-LGPL-linking-exception"
 depends: [
   "dune" {>= "2.5"}
   "ocaml" {>= "4.05.0"}
--- a/packages/camlp-streams/camlp-streams.5.0/opam
+++ b/packages/camlp-streams/camlp-streams.5.0/opam
@@ -34,7 +34,7 @@ depends: [
   "ocaml" {>= "4.05.0"}
 ]
 build: [
-  ["dune" "subst"] {pinned}
+  ["dune" "subst"] {dev}
   [
     "dune"
     "build"
--- a/packages/camlp-streams/camlp-streams.5.0/opam
+++ b/packages/camlp-streams/camlp-streams.5.0/opam
@@ -1,0 +1,57 @@
+opam-version: "2.0"
+synopsis: "The Stream and Genlex libraries for use with Camlp4 and Camlp5"
+description: """
+
+This package provides two library modules:
+- Stream: imperative streams, with in-place update and memoization
+  of the latest element produced.
+- Genlex: a small parameterized lexical analyzer producing streams
+  of tokens from streams of characters.
+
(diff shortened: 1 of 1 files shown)


packages/camlp-streams/camlp-streams.5.0/opam
```diff
@@ -0,0 +1,57 @@
...
+They will be removed from the OCaml standard library at some future point,
+but will be maintained and distributed separately in this camlpstreams package.
+"""
+maintainer: [
+  "Florian Angeletti <florian.angeletti@inria.fr>"
+  "Xavier Leroy <xavier.leroy@college-de-france.fr>"
+]
+authors: ["Daniel de Rauglaudre" "Xavier Leroy"]
+homepage: "https://github.com/ocaml/camlp-streams"
+bug-reports: "https://github.com/ocaml/camlp-streams/issues"
```
avsm, line 31:
```suggestion
bug-reports: "https://github.com/ocaml/camlp-streams/issues"
license: "LGPL-2.1-only with OCaml-LGPL-linking-exception"
```

(I'll commit this to your PR and submit it upstream once you confirm this is the right license @xavierleroy).  We definitely need to improve the "start a project on dune/opam" to avoid all these iterations. @tmattio has https://github.com/tmattio/spin which should hopefully improve this in the coming months.

packages/camlp-streams/camlp-streams.5.0/opam
```diff
@@ -0,0 +1,57 @@
...
+depends: [
+  "dune" {>= "2.5"}
+  "ocaml" {>= "4.05.0"}
+]
+build: [
+  ["dune" "subst"] {pinned}
```
avsm, line 37:
```suggestion
  ["dune" "subst"] {dev}
```

This is because dune opam file generation before dune 2.7 generated the incorrect opam file. The version in `dune-project` need to be bumped to 2.7 and it'll generate this fix.

packages/camlp-streams/camlp-streams.5.0/opam
xavierleroy: Yes, this is the right license.  Please update this PR while I add the license to the sources.

avsm:
All green, thanks!

xavierleroy:
When can we expect to see this package reach the opam.ocaml.org collection?  The most recent package on opam.ocaml.org is from June 27th, yet there was much activity on ocaml/opam-repository since then.

xavierleroy:
Never mind, I understand there were some hardware failures lately, and the new packages are there on opam.ocaml.org.

PR #18972, opened 2021-06-29
[new release] opam-monorepo (0.2.2)

NathanReb:

### Added

- Add `--ocaml-version` argument to `lock`: it allows to determine the ocaml version in the
  lockfile that's being generated (ocamllabs/opam-monorepo#161, @pitag-ha)

### Changed

- Exclude packages depending on `jbuilder` from the lock step. Since dune 2.0, `jbuild` files are
  not supported. A new `--allow-jbuilder` option have been added to enable the old behavior.
- Recognize packages with an optional dependency on dune as building with dune. This allows
  opam-monorepo to rightfully recognize `opam-file-format` latest versions as building with
  dune. (ocamllabs/opam-monorepo#176, @NathanReb)
- Only print the full list of selected root packages once and only in verbose mode, simply printing
  the number in the default logs. (ocamllabs/opam-monorepo#173, @NathanReb)
- Improve the solving process so it only accepts base-compilers unless one explicitly requires
  a compiler variant, either directly or using `ocaml-option-*` packages. (ocamllabs/opam-monorepo#178, @NathanReb)

### Fixed

- Fix the default branch mechanism when the opam remote starts with `git+https` (ocamllabs/opam-monorepo#166, @TheLortex)
- Fix a log that was still refering to the old tool name `duniverse` (ocamllabs/opam-monorepo#158, @rizo)
- Improve how the default branch for a git repository is queried, fixing a bug
  where opam-monorepo wouldn't work outside of of git repo and a bug where it wouldn't
  work on non-english systems. (ocamllabs/opam-monorepo#157, fixes ocamllabs/opam-monorepo#114, @TheLortex)

### DIFF
--- a/packages/opam-monorepo/opam-monorepo.0.2.2/opam
+++ b/packages/opam-monorepo/opam-monorepo.0.2.2/opam
@@ -13,7 +13,7 @@ homepage: "https://github.com/ocamllabs/opam-monorepo"
 bug-reports: "https://github.com/ocamllabs/opam-monorepo/issues"
 depends: [
   "dune" {>= "2.8"}
-  "ocaml" {>= "4.08.0"}
+  "ocaml" {>= "4.08.0" & < "4.13.0"}
   "conf-pkg-config" {build}
 ]
 dev-repo: "git+https://github.com/ocamllabs/opam-monorepo.git"
--- a/packages/opam-monorepo/opam-monorepo.0.2.2/opam
+++ b/packages/opam-monorepo/opam-monorepo.0.2.2/opam
@@ -14,7 +14,6 @@ bug-reports: "https://github.com/ocamllabs/opam-monorepo/issues"
 depends: [
   "dune" {>= "2.8"}
   "ocaml" {>= "4.08.0"}
-  "odoc" {with-doc}
 ]
 dev-repo: "git+https://github.com/ocamllabs/opam-monorepo.git"
 build: [ "dune" "build" "-p" name "-j" jobs "@install" "@runtest" {with-test} ]
--- a/packages/opam-monorepo/opam-monorepo.0.2.2/opam
+++ b/packages/opam-monorepo/opam-monorepo.0.2.2/opam
@@ -14,28 +14,11 @@ bug-reports: "https://github.com/ocamllabs/opam-monorepo/issues"
 depends: [
   "dune" {>= "2.8"}
   "ocaml" {>= "4.08.0"}
+  "conf-pkg-config" {build}
 ]
 dev-repo: "git+https://github.com/ocamllabs/opam-monorepo.git"
 build: [ "dune" "build" "-p" name "-j" jobs "@install" "@runtest" {with-test} ]
 flags: [ plugin ]
-depexts: [
-  ["devel/pkgconf"] {os = "openbsd"}
-  ["pkg-config"] {os-family = "debian"}
-  ["pkg-config"] {os = "macos" & os-distribution = "homebrew"}
-  ["pkgconf"] {os = "freebsd"}
-  ["pkgconf"] {os-distribution = "alpine"}
-  ["pkgconf"] {os-distribution = "arch"}
-  ["pkgconf-pkg-config"] {os-distribution = "centos" & os-version >= "8"}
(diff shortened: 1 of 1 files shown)


packages/opam-monorepo/opam-monorepo.0.2.2/opam
```diff
@@ -0,0 +1,47 @@
...
+  ["pkgconf-pkg-config"] {os-distribution = "rhel" & os-version >= "8"}
+  ["pkgconfig"] {os-distribution = "fedora"}
+  ["pkgconfig"] {os-distribution = "mageia"}
+  ["pkgconfig"] {os-distribution = "nixos"}
+  ["pkgconfig"] {os-distribution = "ol"}
+  ["pkgconfig"] {os = "macos" & os-distribution = "macports"}
+  ["pkgconfig"] {os-distribution = "centos" & os-version <= "7"}
+  ["pkgconfig"] {os-distribution = "rhel" & os-version <= "7"}
+  ["system:pkgconf"] {os = "win32" & os-distribution = "cygwinports"}
+]
```
kit-ty-kate, line 38:
```suggestion
  "conf-pkg-config" {build}
]
dev-repo: "git+https://github.com/ocamllabs/opam-monorepo.git"
build: [ "dune" "build" "-p" name "-j" jobs "@install" "@runtest" {with-test} ]
flags: [ plugin ]
```
I was under the impression that this alternative way of handling depexts was more or less agreed in the discussion leading to https://github.com/ocamllabs/opam-monorepo/issues/152

avsm: I'm fine with this diff being in the opam-repository, and also with @NathanReb's version in the upstream repository lockfile.

NathanReb: Having further thought about this I'm not sure how well it fits with opam-monorepo. One of the reason to use lockfiles is to shield itself from upstream changes in opam-repository to ensure reproducibility.

I understand that in this case there are few chances that changes to `conf-*` packages actually result in breaking something but I learned to adopt a defensive posture with those things.

That being said, opam-monorepo is using itself mostly to get direct feedback on the tool and not because we really care about reproducibility so I'm happy with this change here.

avsm: Agreed @NathanReb. Let's treat this as an experiment: start by vendoring the OCaml dependencies here with monorepo and see how it goes, and keep an eye on depexts and see if it becomes a problem.  We'll learn with every new monorepo-using release. Perhaps its worth adding a `tag` field to mark the package as a monorepo user?

packages/opam-monorepo/opam-monorepo.0.2.2/opam
```diff
@@ -0,0 +1,30 @@
...
+maintainer: ["anil@recoil.org"]
+authors: [
+  "Anil Madhavapeddy" "Nathan Rebours" "Lucas Pluvinage" "Jules Aguillon"
+]
+license: "ISC"
+homepage: "https://github.com/ocamllabs/opam-monorepo"
+bug-reports: "https://github.com/ocamllabs/opam-monorepo/issues"
+depends: [
+  "dune" {>= "2.8"}
+  "ocaml" {>= "4.08.0"}
```
avsm, line 16:
```suggestion
  "ocaml" {>= "4.08.0" & < "4.13.0"}
```

The CI fails to build the vendored libraries on 4.13, so constraining it down.

packages/opam-monorepo/opam-monorepo.0.2.2/opam
NathanReb: We could do that indeed! I'm also still strongly considering using a variable to control which dependencies are vendored so that the opam metadata reflect what the actual dependencies are while still not installing them by default and relying on the vendoring instead.

The presence of this variable in an opam file would act as a tag as it would easily allow us to track down which packages use opam-monorepo.

You're suggesting the use of tag for that purpose right? Since we're actively trying to make the releases compatible with default opam, I don't expect opam itself to interpret this tag.

NathanReb:
Restarted the failed centos build, it all seems good now!

avsm:

PR #18781, opened 2021-06-02
Release Links 0.9.3

dhil:
Minor release for Links (this time without links-mysql8 included). Various stability and improvement fixes.

PR #18975, opened 2021-06-30
SPDX compliant license field for Links packages.

dhil:
Follow up on #18781.

This commit makes the license fields of previously released Links packages SPDX compliant.


avsm:
Thanks, we appreciate the followup!

dhil:
My pleasure!

PR #18985, opened 2021-07-01
[new release] doi2bib (0.5.1)

mseri:

##### CHANGES:

- Fix for transitive dependency in cuz
- Improved lower bounds on the opam file

### DIFF
--- a/packages/doi2bib/doi2bib.0.5.1/opam
+++ b/packages/doi2bib/doi2bib.0.5.1/opam
@@ -30,7 +30,6 @@ build: [
     "-j"
     jobs
     "@install"
-    "@runtest" {with-test}
     "@doc" {with-doc}
   ]
 ]
--- a/packages/doi2bib/doi2bib.0.5.1/opam
+++ b/packages/doi2bib/doi2bib.0.5.1/opam
@@ -20,6 +20,9 @@ depends: [
   "re" {>= "1.0.0"}
   "odoc" {with-doc}
 ]
+conflicts: [
+  "result" {< "1.5"} # uses Result.map but result can be pulled via lwt and takes over the Result module
+]
 build: [
   ["dune" "subst"] {dev}
   [
--- a/packages/doi2bib/doi2bib.0.5.1/opam
+++ b/packages/doi2bib/doi2bib.0.5.1/opam
@@ -12,9 +12,9 @@ depends: [
   "astring" {>= "0.8.0"}
   "cohttp-lwt-unix" {>= "2.5.0"}
   "cmdliner" {>= "1.0.0"}
-  "decompress" {>= "1.3.0"}
+  "decompress" {>= "1.4.0"}
   "ezxmlm" {>= "1.1.0"}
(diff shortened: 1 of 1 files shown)


packages/doi2bib/doi2bib.0.5.1/opam
```diff
@@ -0,0 +1,46 @@
...
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
```
mseri, line 33:
```suggestion
```
Disabled since they cannot run in the sandbox

packages/doi2bib/doi2bib.0.5.1/opam
```diff
@@ -0,0 +1,45 @@
...
+license: "MIT"
+homepage: "https://github.com/mseri/doi2bib"
+bug-reports: "https://github.com/mseri/doi2bib/issues"
+depends: [
+  "dune" {>= "2.7"}
+  "ocaml" {>= "4.08"}
+  "astring" {>= "0.8.0"}
+  "cohttp-lwt-unix" {>= "2.5.0"}
+  "cmdliner" {>= "1.0.0"}
+  "decompress" {>= "1.3.0"}
```
mseri, line 15:
```suggestion
  "decompress" {>= "1.4.0"}
```

packages/doi2bib/doi2bib.0.5.1/opam
```diff
@@ -0,0 +1,45 @@
...
+  "ezxmlm" {>= "1.1.0"}
+  "lwt" {>= "5.2.0"}
```
mseri, line 17:
```suggestion
  "lwt" {>= "5.3.0"}
```

mseri:
The failure is weird. Isn't `Result.map` available since ocaml 4.08? Why is it failing with `- Error: Unbound value Result.map`?

kit-ty-kate:
> The failure is weird. Isn't `Result.map` available since ocaml 4.08? Why is it failing with `- Error: Unbound value Result.map`?

The usual culprit: result < 1.5 :sweat_smile:

mseri:
Thanks!

kit-ty-kate:
To fix this without having to add a conflict you might want to add `(env (_ (flags :standard -open Stdlib)))` to your dune file, though it's probably not a really future proof way to do. Adding `module Result = Stdlib.Result` at the top of your files is uglier but probably more future proof.

mseri:
Thanks. For now I added the conflict as you did here. When I update the code I will override result on the top module just in case…

kit-ty-kate:
> Thanks. For now I added the conflict as you did here. When I update the code I will override result on the top module just in case…

The issue with using `module Result = ...` is that it has to be systematically added to every modules (even signatures) as even mentioning `Result.t` is not supported by result.1.0 :/

kit-ty-kate:
Thanks a lot!

mseri:
Thank you!

PR #18994, opened 2021-07-02
Package tezt.1.0.0

jtcoolen:
This PR packages Tezos' integration testing framework. @rbardou suggested using the version number `1.0.0` since it is Tezt's first release.

More information about Tezt at: [https://tezos.gitlab.io/developer/tezt.html](https://tezos.gitlab.io/developer/tezt.html).

### DIFF
--- a/packages/tezt/tezt.1.0.0/opam
+++ b/packages/tezt/tezt.1.0.0/opam
@@ -6,7 +6,7 @@ bug-reports: "https://gitlab.com/tezos/tezos/issues"
 dev-repo: "git+https://gitlab.com/tezos/tezos.git"
 license: "MIT"
 depends: [
-  "dune" {>= "2.1.1"}
+  "dune" {>= "2.5"}
   "ocaml" {>= "4.08"}
   "re" {>= "1.9.0"}
   "lwt" {>= "5.4.1"}
--- a/packages/tezt/tezt.1.0.0/opam
+++ b/packages/tezt/tezt.1.0.0/opam
@@ -1,0 +1,26 @@
+opam-version: "2.0"
+maintainer: "contact@tezos.com"
+authors: [ "Tezos devteam" ]
+homepage: "https://www.tezos.com/"
+bug-reports: "https://gitlab.com/tezos/tezos/issues"
+dev-repo: "git+https://gitlab.com/tezos/tezos.git"
+license: "MIT"
+depends: [
+  "dune" {>= "2.1.1"}
+  "ocaml" {>= "4.08"}
+  "re" {>= "1.9.0"}
+  "lwt" {>= "5.4.1"}
+  "ezjsonm" {>= "1.2.0"}
+]
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+]
+synopsis: "Framework for integration tests with external processes"
+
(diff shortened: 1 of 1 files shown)


packages/tezt/tezt.1.0.0/opam
```diff
@@ -0,0 +1,25 @@
+opam-version: "2.0"
+maintainer: "contact@tezos.com"
+authors: [ "Tezos devteam" ]
+homepage: "https://www.tezos.com/"
+bug-reports: "https://gitlab.com/tezos/tezos/issues"
+dev-repo: "git+https://gitlab.com/tezos/tezos.git"
+license: "MIT"
+depends: [
+  "dune" { >= "2.0" }
```
mseri, line 9:
```suggestion
  "ocaml" {>= "4.08"}
  "dune" { >= "2.0" }
```

jtcoolen: Thanks, I came to the same conclusion.

mseri: It may also need a lower bound on dune. The lower bound tests, downgrading to dune 2.0.1 fails at https://gitlab.com/tezos/tezos/-/blob/master/src/lib_micheline/test/dune#L10 with
```
#=== ERROR while compiling tezt.1.0.0 =========================================#
# context              2.1.0~rc2 | linux/x86_64 | ocaml-base-compiler.4.08.1 | pinned(https://gitlab.com/tezos/tezos/-/archive/v10.0-rc1/tezos-v10.0-rc1.tar.bz2)
# path                 ~/.opam/4.08/.opam-switch/build/tezt.1.0.0
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p tezt -j 47
# exit-code            1
# env-file             ~/.opam/log/tezt-6473-4bfab6.env
# output-file          ~/.opam/log/tezt-6473-4bfab6.out
### output ###
# File "src/lib_micheline/test/dune", line 10, characters 2-16:
# 10 |   (action progn))
#        ^^^^^^^^^^^^^^
# Error: if you meant for this to be executed with bash, write (bash "...")
# instead
```

jtcoolen: We might want to disable tests from other Tezos modules.

jtcoolen: In fact, there's a missing `(package ...)` in the dune file, I'll open a merge request on Tezos first.

jtcoolen:
The Tezos toolchain was upgraded to OCaml 4.12, but not 4.13 yet.

kit-ty-kate:
I removed your last commit. The error on 4.13 should be fixed by #18997 and the constraint used was wrong. `<= "4.12"` does not include 4.12 (as `4.12 < 4.12.0`, see the [opam manual](https://opam.ocaml.org/doc/Manual.html#Package-Formulas)) and constraints such as `<=` should almost never be used in practice.

kit-ty-kate:
This package seems compatible with OCaml 4.13 so there is no further change to do. Thanks! Could you return the fixes to the opam file upstream?

jtcoolen:
Thanks, that was quick! Sure, will look after it.

PR #19006, opened 2021-07-06
Package lutils.1.51.0

jahierwan:
### `lutils.1.51.0`
Tools and libs shared by Verimag/synchronous tools (lustre, lutin, rdbg)
The lutils library contains various modules shared between tools developped at
Verimag in the synchronous group. Those modules deal with:
- generate and parse RIF files
- generate dro files (to call luciole)

The gnuplot-rif tool vizualise RIF files using gnuplot.

---
* Homepage: https://gricad-gitlab.univ-grenoble-alpes.fr/verimag/synchrone/lutils/
* Source repo: git+https://gricad-gitlab.univ-grenoble-alpes.fr/verimag/synchrone/lutils.git
* Bug tracker: https://gricad-gitlab.univ-grenoble-alpes.fr/verimag/synchrone/lutils/issues

---

### DIFF
--- a/packages/lutils/lutils.1.51.0/opam
+++ b/packages/lutils/lutils.1.51.0/opam
@@ -16,9 +16,9 @@ depends: [
   "num"
 ]
 build: [
-  [make "build"]
+  [make "lib/lutilsVersion.ml"]
+  ["dune" "build" "-p" name "-j" jobs]
 ]
-install: [make "install"]
 post-messages:
   "The last version can be obtained via (opam repo add) http://www-verimag.imag.fr/DIST-TOOLS/SYNCHRONE/opam-repository/"
 dev-repo:
--- a/packages/lutils/lutils.1.51.0/opam
+++ b/packages/lutils/lutils.1.51.0/opam
@@ -1,0 +1,42 @@
+opam-version: "2.0"
+maintainer: "erwan.jahier@imag.fr"
+authors: "Erwan Jahier"
+license: "CeCILL"
+homepage:
+  "https://gricad-gitlab.univ-grenoble-alpes.fr/verimag/synchrone/lutils/"
+bug-reports:
+  "https://gricad-gitlab.univ-grenoble-alpes.fr/verimag/synchrone/lutils/issues"
+doc: "http://www-verimag.imag.fr/DIST-TOOLS/SYNCHRONE/lustre-v6"
+
+depends: [
+  "ocaml" {>= "4.02"}
+  "dune"  {>= "1.2"}
+  "base-unix"
+  "ocamlfind"
+  "num"
+]
+build: [
+  [make "build"]
+]
+install: [make "install"]
+post-messages:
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
It looks like this package uses dune under the hood. However there is no `dune-project` file present in the archive. Furthermore dune build without `-p` should never be used in releases as it is way too brittle.
Looking at the source it looks like `lib/lutilsVersion.ml` could be generated using dune itself, this would simplify the package.
The call to `dune install` is also not necessary since `dune build` already generates a `<pkg>.install` file used by opam.

Could you return the fixes to the upstream file upstream?

jahierwan:
> It looks like this package uses dune under the hood. However there is no `dune-project` file present in the archive. 

it is in the git(lab) repo, but not in the archive indeed. I will add it.

> Furthermore dune build without `-p` should never be used in releases as it is way too brittle.

ok, thanks for the tip.

> Looking at the source it looks like `lib/lutilsVersion.ml` could be generated using dune itself, this would simplify the package.

how so?

> The call to `dune install` is also not necessary since `dune build` already generates a `<pkg>.install` file used by opam.

ok

> Could you return the fixes to the upstream file upstream?

yep.

kit-ty-kate:
> how so?

This should do the job:
```
diff -ru lutils.1.51.0/lib/dune lutils.1.51.0-new/lib/dune
--- lutils.1.51.0/lib/dune	2020-07-23 10:41:33.000000000 +0100
+++ lutils.1.51.0-new/lib/dune	2021-07-06 13:48:45.303758136 +0100
@@ -5,3 +5,8 @@
   (wrapped false)
   (synopsis "A lib shared by Verimag/synchronous tools (lustre-v6, lutin, rdbg).")
 )
+
+(rule
+  (targets lutilsVersion.ml)
+  (deps generate-version.sh)
+  (action (run ./generate-version.sh %{version:lutils})))
```
With `(version 1.51.0)` in your dune-project file and the following `lib/generate-version.sh`:
```
#!/bin/sh

set -e

SHA=$(git log -1 --pretty=format:"%h" || echo "opam")
VERSION=$(git describe --tags || echo "${1}")
echo "let str=\"${VERSION}\"" > lutilsVersion.ml
echo "let sha=\"${SHA}\"" >> lutilsVersion.ml 
```

jahierwan:
ok, thank you very much !
[...]
Still, I'm not sure to understand why you want me to write the version in the dune-project (which I'd to prefer to avoid, as in my current workflow, the version number is always automatically generated). 

My guess is that it is necessary to get %{version:lutils} instanciated, but your generate-version.sh script does not use any argument anyway. Beside,
```
(rule
  (targets lutilsVersion.ml)
  (deps generate-version.sh)
  (action (run ./generate-version.sh)))
 ```
works fine actually.

jahierwan:
is it ok if I opam publish a 1.15.1 version instead of this one ?
(sorry to be such a newbie)

jahierwan:
oh, I see the $1 in your script now...

jahierwan:
I've taken (hopefully) all your remarks into accout here:  https://github.com/ocaml/opam-repository/pull/19011

PR #19009, opened 2021-07-06
Tezos 9.4

pirbo:
This is an emergency update of the octez distribution to workaround spam that occurred last Friday. Emergency is now over but it is still useful to everybody to take the patch :-)


pirbo:
I have side question : The diff between version 9.3 and 9.4 is in fact 1 patch in 1 opam package of the distribution so instead of repushing more than an hundred new packages because it is a """new version""", would it be better to "replace in place" 1 package by adding a patch?

avsm:
Right now, it's better to keep all the packages published the way they are in this PR. The reason is that they share the same distfile and have all been tested together.  Mixing and matching versions if possible if there is a testing strategy to ensure all the various combinations are guaranteed to work.

avsm:

PR #19010, opened 2021-07-06
Package memprof-limits.0.2.0

gadmm:
### `memprof-limits.0.2.0`
Memory limits, allocation limits, and thread cancellation
Memprof-limits provides per-thread global memory limits, per-thread
allocation limits, and cancellation of threads, with ways to
ensure resource-safety after interruption.

Global memory limits let you bound the memory consumption of a task,
in terms of the major heap size.

Allocation limits let you bound the execution time of a task measured
in number of allocations. Allocation limits do not count
deallocations, and are therefore a measure of the work done, which can
be more suitable (reliable, portable, deterministic) than wall-clock
time.

Token limits lets you cancel a (CPU-bound) task preemptively and at a
distance.

Tasks are interrupted by raising an asynchronous exception.
Memprof-limits provides features and guidance for reasoning about the
consistency of state in the presence of such interrupts.

The implementation uses OCaml's Memprof engine with a low sampling
rate that does not affect performance. A reimplementation of the
Memprof interface compatible with memprof-limits running at the same
time is provided for profiling needs.

---
* Homepage: https://gitlab.com/gadmm/memprof-limits/
* Bug tracker: https://gitlab.com/gadmm/memprof-limits/issues/

---

### DIFF
--- a/packages/memprof-limits/memprof-limits.0.2.0/opam
+++ b/packages/memprof-limits/memprof-limits.0.2.0/opam
@@ -1,0 +1,48 @@
+opam-version: "2.0"
+synopsis: "Memory limits, allocation limits, and thread cancellation"
+description: """\
+Memprof-limits provides per-thread global memory limits, per-thread
+allocation limits, and cancellation of threads, with ways to
+ensure resource-safety after interruption.
+
+Global memory limits let you bound the memory consumption of a task,
+in terms of the major heap size.
+
+Allocation limits let you bound the execution time of a task measured
+in number of allocations. Allocation limits do not count
+deallocations, and are therefore a measure of the work done, which can
+be more suitable (reliable, portable, deterministic) than wall-clock
+time.
+
+Token limits lets you cancel a (CPU-bound) task preemptively and at a
+distance.
+
+Tasks are interrupted by raising an asynchronous exception.
+Memprof-limits provides features and guidance for reasoning about the
+consistency of state in the presence of such interrupts.
+
+The implementation uses OCaml's Memprof engine with a low sampling
+rate that does not affect performance. A reimplementation of the
+Memprof interface compatible with memprof-limits running at the same
+time is provided for profiling needs."""
+maintainer:
+  "Guillaume Munch-Maccagnoni <Guillaume.Munch-Maccagnoni@inria.fr>"
+authors: "Guillaume Munch-Maccagnoni <Guillaume.Munch-Maccagnoni@inria.fr>"
+license: "LGPL-3.0-linking-exception"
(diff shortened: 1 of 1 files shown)


packages/memprof-limits/memprof-limits.0.2.0/opam
```diff
@@ -0,0 +1,48 @@
...
+consistency of state in the presence of such interrupts.
+
+The implementation uses OCaml's Memprof engine with a low sampling
+rate that does not affect performance. A reimplementation of the
+Memprof interface compatible with memprof-limits running at the same
+time is provided for profiling needs."""
+maintainer:
+  "Guillaume Munch-Maccagnoni <Guillaume.Munch-Maccagnoni@inria.fr>"
+authors: "Guillaume Munch-Maccagnoni <Guillaume.Munch-Maccagnoni@inria.fr>"
+license: "LGPL-3.0-linking-exception"
```
mseri, line 31:
```suggestion
license: "LGPL-3-only WITH OCaml-LGPL-linking-exception"
```
is that correct or is the `or-later` version?

gadmm: It is fixed in my repo already but I thought this CI check could be overridden :) (I do not imagine opam forcing a choice of licenses among a list from third party, and this field has no legal value anyway.) 
Note the appropriate fix below. The OCaml linking exception is for LGPL 2 and is meaningless for LGPL 3, and I have seen many opam packages apply a meaningless linking exception to LGPL 3.
```suggestion
license: "LGPL-3.0-only WITH LGPL-3.0-linking-exception"
```

gadmm: The CI still chokes on this. I do not see what is wrong nor do I have time to read this "SPDX" specification to try to guess what is wrong. I think it would be nice to disable this lint from the CI since this seems to cause delays for no useful reasons.

gadmm: The bug seems to be at <https://github.com/ocaml/opam/blob/074df6b6d87d4114116ea41311892b342cfad3de/src/state/opamSpdxList.ml>, which misses the LGPL 3 linking exception.

mseri: That's not a problem. Thanks for looking into it, I will send a PR to add it. This lint is not so strict, but since many packages are missing licenses or have wrong ones (like the one I was suggesting), I think it is important to keep it.

gadmm: ok! thanks.

PR #19014, opened 2021-07-07
conf-git: update version to not register a homebrew depext

avsm:
The homebrew version of git is a newer version, but is not
necessary to explicitly specify in conf-git since there is
always a base `git` binary in macOS.  This matters when using
opam 2.1, which automatically tries to `brew install git`
unnecessarily.

### DIFF
--- a/packages/conf-git/conf-git.1.1/opam
+++ b/packages/conf-git/conf-git.1.1/opam
@@ -4,17 +4,17 @@ homepage: "https://git-scm.com"
 authors: "Linus Torvalds"
 bug-reports: "https://github.com/ocaml/opam-repository/issues"
 license: "GPL-2.0-or-later"
-build: [["which" "git"]]
-depends: [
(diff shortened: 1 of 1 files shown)


packages/conf-git/conf-git.1.1/opam
```diff
@@ -0,0 +1,22 @@
+opam-version: "2.0"
+maintainer: "unixjunkie@sdf.org"
+homepage: "https://git-scm.com"
+authors: "Linus Torvalds"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: "GPL-2.0-or-later"
+build: [["which" "git"]]
```
mseri, line 7: What about
```suggestion
build: [["git" "--version"]]
```
completely dropping the depends field?
This way this can be compatible with windows

kit-ty-kate:
looks good now. Thanks!

PR #19022, opened 2021-07-08
[new release] randoml (0.1.0)

mimoo:
Generating cryptographically-secure random numbers

- Project page: <a href="https://github.com/mimoo/randoml">https://github.com/mimoo/randoml</a>

##### CHANGES:

First release

### DIFF
--- a/packages/randoml/randoml.0.1.0/opam
+++ b/packages/randoml/randoml.0.1.0/opam
@@ -10,6 +10,7 @@ bug-reports: "https://github.com/mimoo/randoml/issues"
 depends: [
   "bignum" {>= "v0.14.0"}
   "dune" {>= "2.8" & >= "2.8.0"}
+  "conf-rust-2018"
   "ppx_inline_test" {with-test & >= "v0.14.1"}
   "core_kernel" {> "v0.14.0"}
   "odoc" {with-doc}
--- a/packages/randoml/randoml.0.1.0/opam
+++ b/packages/randoml/randoml.0.1.0/opam
@@ -1,0 +1,40 @@
+opam-version: "2.0"
+synopsis: "Generating cryptographically-secure random numbers"
+description:
+  "A library relying on Rust rand to generate random numbers. The random numbers are cryptographically secure."
+maintainer: ["david@o1labs.org"]
+authors: ["David Wong"]
+license: "ISC"
+homepage: "https://github.com/mimoo/randoml"
+bug-reports: "https://github.com/mimoo/randoml/issues"
+depends: [
+  "bignum" {>= "v0.14.0"}
+  "dune" {>= "2.8" & >= "2.8.0"}
+  "ppx_inline_test" {with-test & >= "v0.14.1"}
+  "core_kernel" {> "v0.14.0"}
+  "odoc" {with-doc}
+]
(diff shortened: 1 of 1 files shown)


packages/randoml/randoml.0.1.0/opam
```diff
@@ -0,0 +1,40 @@
...
+  "A library relying on Rust rand to generate random numbers. The random numbers are cryptographically secure."
+maintainer: ["david@o1labs.org"]
+authors: ["David Wong"]
+license: "ISC"
+homepage: "https://github.com/mimoo/randoml"
+bug-reports: "https://github.com/mimoo/randoml/issues"
+depends: [
+  "bignum" {>= "v0.14.0"}
+  "dune" {>= "2.8" & >= "2.8.0"}
+  "ppx_inline_test" {with-test & >= "v0.14.1"}
```
mseri, line 13:
```suggestion
  "conf-rust-2018"
  "ppx_inline_test" {with-test & >= "v0.14.1"}
```
There is also `conf-rust` if you don't need the specific edition

packages/randoml/randoml.0.1.0/opam
```diff
@@ -0,0 +1,40 @@
...
+description:
+  "A library relying on Rust rand to generate random numbers. The random numbers are cryptographically secure."
+maintainer: ["david@o1labs.org"]
+authors: ["David Wong"]
+license: "ISC"
+homepage: "https://github.com/mimoo/randoml"
+bug-reports: "https://github.com/mimoo/randoml/issues"
+depends: [
+  "bignum" {>= "v0.14.0"}
+  "dune" {>= "2.8" & >= "2.8.0"}
```
mseri, line 12:
```suggestion
  "dune" {>= "2.8.0"}
```

packages/randoml/randoml.0.1.0/opam
mseri: It will not be able to pull `rand` though, you need to also vendor that, for example this is how is done by https://gitlab.com/tezos/tezos-rust-libs/

packages/randoml/randoml.0.1.0/opam
mimoo: [opam-dune-lint](https://opam.ocaml.org/packages/opam-dune-lint/) does weird things o.o

mseri: Don’t worry about that. It is a know bug. You can avoid it by removing dune from the dependencies in dune-project. Dune automatically adds itself to the list in any case

packages/randoml/randoml.0.1.0/opam
mimoo: because I also named my rust wrapper "rand" I'm not sure which rand you're talking about :D

[my rand](https://github.com/mimoo/randoml/blob/main/lib/rand/dune#L24) should be fine if I understand correctly because the public name is `randoml.rand`

intuitively the rust dependencies should be fine as well as cargo will download them during build. But now I think I understand that since opam install runs in a sandbox this won't happen... OK I see

mimoo:
version 0.1.1 with fixes here: https://github.com/ocaml/opam-repository/pull/19036

PR #19023, opened 2021-07-08
[new release] opam-dune-lint (0.2)

talex5:
Ensure dune and opam dependencies are consistent

- Project page: <a href="https://github.com/ocurrent/opam-dune-lint">https://github.com/ocurrent/opam-dune-lint</a>

##### CHANGES:

- Cope with missing `(depends ...)` in `dune-project` (@talex5 ocurrent/opam-dune-lint#33). We tried to add the missing packages to an existing depends field, but if there wasn't one at all then we did nothing.

- Use quoted versions in the fix suggestion string (@tmcgilchrist ocurrent/opam-dune-lint#32). Makes copy-and-paste easier for people using it via a web UI.

- Support older versions of OCaml back to 4.10 (@tmcgilchrist ocurrent/opam-dune-lint#31).

- Ignore dependencies on sub-packages (@dra27 ocurrent/opam-dune-lint#27). Library `foo` may depend on library `foo.bar` but this cannot introduce an opam dependency on `foo` in `foo.opam`.

- Require opam libraries compatible with the client (@dra27 ocurrent/opam-dune-lint#26).

- Add support for multiple dependency clauses for the same package (@kit-ty-kate ocurrent/opam-dune-lint#25).

- Upgrade to dune-private-libs 2.8.0 (@kit-ty-kate ocurrent/opam-dune-lint#20).

- Remove dependency on ocamlfind, as we don't use it for anything now (@talex5 ocurrent/opam-dune-lint#18).

### DIFF
--- a/packages/opam-dune-lint/opam-dune-lint.0.2/opam
+++ b/packages/opam-dune-lint/opam-dune-lint.0.2/opam
@@ -1,0 +1,47 @@
+opam-version: "2.0"
+synopsis: "Ensure dune and opam dependencies are consistent"
+description:
+  "opam-dune-lint checks that all ocamlfind libraries listed as dune dependencies have corresponding opam dependencies listed in the opam files. If not, it offers to add them (either to your opam files, or to your dune-project if you're generating your opam files from that)."
+maintainer: ["talex5@gmail.com"]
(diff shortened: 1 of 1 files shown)


mseri:
Looks good, thanks

PR #19026, opened 2021-07-08
add some missing metadata to various packages

zapashcanon:
Hi,

I noticed some stuff was missing while doing some linting on the whole opam repository and had a go at fixing some of the packages.


kit-ty-kate:
Thanks a lot!

PR #19030, opened 2021-07-09
[new release] non_empty_list (0.1)

johnyob:
A list that is known, statically, to be nonempty. Removes `option`, `result` types, or exceptions from many list functions that rely on non-empty lists. 
- https://github.com/johnyob/ocaml-non-empty-list 

**CHANGES:**

Initial release.

### DIFF
--- a/packages/non_empty_list/non_empty_list.0.1/opam
+++ b/packages/non_empty_list/non_empty_list.0.1/opam
@@ -9,7 +9,6 @@ depends: [
   "dune" { >= "2.8" }
   "base" { >= "v0.14.1" }
   "ppx_deriving" { >= "4.5" }
-  "ocamlformat"
   "alcotest" {with-test}
 ]
 license: "MIT"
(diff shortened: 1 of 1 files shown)


packages/non_empty_list/non_empty_list.0.1/opam
```diff
@@ -0,0 +1,36 @@
+opam-version: "2.0"
+maintainer: "ajo41@cam.ac.uk"
+authors: ["Alistair O'Brien"]
+homepage: "https://github.com/johnyob/ocaml-non-empty-list"
+bug-reports: "https://github.com/johnyob/ocaml-non-empty-list/issues"
+dev-repo: "git+https://github.com/johnyob/ocaml-non-empty-list.git"
+depends: [ 
+  "ocaml" { >= "4.08.0" }
+  "dune" { >= "2.8" }
+  "base" { >= "0.14.1" }
```
mseri, line 10:
```suggestion
  "base" { >= "v0.14.1" }
```

mseri: Can you make this change also on your repo?

johnyob: Sure! Thank you so much for your help. It's my first opam package so, thanks for fixing the lower bound tests :)

kit-ty-kate:
ocamlformat was not a dependency so I removed it. I also added missing -j to allow users to make opam take less resources if they need. Otherwise it’s good now. Could you return the fixes to the opam file upstream so the changes are taken into account for future releases?

PR #19031, opened 2021-07-09
5 packages from www-verimag.imag.fr/DIST-TOOLS/SYNCHRONE/pool/lutin.2.71.10.tgz

jahierwan:
This pull-request concerns:
-`bddrand.2.71.10`: A simple front-end to the lutin Random toss machinary
-`ezdl.2.71.10`: Easy dynamic linking of C functions from ocaml
-`gbddml.2.71.10`: The Verimag bdd library
-`lutin.2.71.10`: Lutin: modeling stochastic reactive systems
-`polka.2.71.10`: Polka: convex polyhedron library by Bertrand Jeannet (now part of apron)

---

---

### DIFF
--- a/packages/lutin/lutin.2.56/opam
+++ b/packages/lutin/lutin.2.56/opam
@@ -25,7 +25,7 @@ remove: [
   ["ocaml" "%{etc}%/Lutin/setup.ml" "-C" "%{etc}%/Lutin" "-uninstall"]
 ]
 depends: [
-  "ocaml" {>= "4.02" & < "4.06" }
+  "ocaml" {>= "4.04" & < "4.06" }
   "conf-gmp"
   "conf-mpfr"
   "conf-m4" {build}
--- a/packages/lutin/lutin.2.56/opam
+++ b/packages/lutin/lutin.2.56/opam
@@ -25,7 +25,7 @@ remove: [
   ["ocaml" "%{etc}%/Lutin/setup.ml" "-C" "%{etc}%/Lutin" "-uninstall"]
 ]
 depends: [
-  "ocaml" {>= "4.02" & < "4.06" | >= "4.11"  }
+  "ocaml" {>= "4.02" & < "4.06" }
   "conf-gmp"
   "conf-mpfr"
   "conf-m4" {build}
--- a/packages/lutin/lutin.2.56/opam
+++ b/packages/lutin/lutin.2.56/opam
@@ -25,10 +25,8 @@ remove: [
   ["ocaml" "%{etc}%/Lutin/setup.ml" "-C" "%{etc}%/Lutin" "-uninstall"]
 ]
 depends: [
-  "ocaml" {>= "4.04" & < "4.06" }
+  "ocaml" {>= "4.02" & < "4.12"}
   "conf-gmp"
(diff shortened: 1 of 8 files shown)


jahierwan:
I have 5 packages defined in the same git repo. It works well from dune as they are all compiled at the same time. But opam install them in turn and dependencies issues arises.

I'll will fix them (here or in another PR).

jahierwan:
One external dep, the gmp-dev package, fails to install on alpine with opam 2.1 (while it works with opam 2.0).

jahierwan:
ok, many thanks for the fix! I'll do more tgz rebuilding the afternoon.

jahierwan:
thank for your help ktk

The lutin.2.71.10 actually install nothing ; it seems I need to put back the build (and the install) rules. I
11f4889b183d869a5cc3ea4878e2bb9d599e76d0 was actually useless;  I will undo it.

I will refine the constraints for 2.56 and 2.70.4 
    "ocaml" {>= "4.02" & < "4.06" | >= "4.11"}

jahierwan:
I still have occurrences of useless Stdlib references. I will update the lutin.2.71.10.tgz

packages/lutin/lutin.2.71.10/opam
```diff
@@ -0,0 +1,47 @@
...
+  "extlib" {build} | "extlib-compat" {build}
+  "lustre-v6"
+  "lutils" {>= "1.49"}
+  "mlgmpidl"
+  "dune" {>= "2.0"}
+  "ocamlfind"
+  "rdbg" {>= "1.196.9"}
+]
+build: ["dune" "build" "@install" "-j" jobs]
+install: [ make "install" ]
```
kit-ty-kate, line 37: Why has the `-p` argument been removed? make install shouldn't be required either.

packages/lutin/lutin.2.56/opam
```diff
@@ -25,8 +25,10 @@ remove: [
   ["ocaml" "%{etc}%/Lutin/setup.ml" "-C" "%{etc}%/Lutin" "-uninstall"]
 ]
 depends: [
-  "ocaml" {>= "4.02" & < "4.12"}
+  "ocaml" {>= "4.04" & < "4.06" }
```
kit-ty-kate, line 28: these new constraints seem highly suspisious. Failures on those specific already established packages can be ignored for now. Could you move this to another PR? (same for 2.70.4)

packages/lutin/lutin.2.71.10/opam
jahierwan: I've removed the -p argument to fix the build but you fix in 0a48d3a8e7c24a127a33650db54902fc6b2c8cba is better indeed (I tough I've tried something like that!), thanks.

Without the install, the ci is happy but nothing is actually installed

kit-ty-kate:
Thanks!

kit-ty-kate:
Could you return the fix in the opam files upstream?

jahierwan:
Sure !

Thanks for your help 
ps: your changes break the ci on my local repo, but if it works here, i guess this is fine. 
I should be to figure out what's happening.

PR #19039, opened 2021-07-12
[new release] elpi (1.13.7)

gares:
CHANGES:

- Compiler:
  - Fix bug in spilling when the spilled expression generates more than one
    argument.


kit-ty-kate:
Thanks a lot!

PR #19040, opened 2021-07-12
New version: Fix few bugs

Nyavlys:
ACGtk new version. A few bug fix and ocaml 4.12.0 compatibility fix.

### DIFF
--- a/packages/acgtk/acgtk.1.5.3/opam
+++ b/packages/acgtk/acgtk.1.5.3/opam
@@ -14,7 +14,7 @@ install: ["dune" "install"]
 depends: [
   "ocaml" { >= "4.05.0" }
   "dune" { >= "1.4" }
-  "menhir"
+  "menhir" {>= "20181113"}
   "ANSITerminal"
   "fmt"
   "logs"
@@ -24,7 +24,7 @@ depends: [
   "conf-pkg-config"
   "conf-cairo"
   "cairo2"
-  "yojson"
+  "yojson" {>= "1.6.0"}
   "easy-format"
 ]
 
--- a/packages/acgtk/acgtk.1.5.3/opam
+++ b/packages/acgtk/acgtk.1.5.3/opam
@@ -19,7 +19,7 @@ depends: [
   "fmt"
   "logs"
   "mtime" {>= "1.0.0"}
-  "cmdliner"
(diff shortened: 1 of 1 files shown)


packages/acgtk/acgtk.1.5.3/opam
```diff
@@ -0,0 +1,45 @@
+opam-version: "2.0"
+maintainer: "sylvain.pogodalla@inria.fr"
+
+build: [
+  ["dune" "subst"] {dev}
+# remove the -p to also build the local libraries: conflict with the
+# fact that some libraries are also part of the acgtkLib package
```
mseri, line 7: Does this require to add a `conflicts: ["acgtklib"]` section?

Nyavlys: I don't think so. The acgtkLib package was only meant for local dev usage and is not meant to be distributed with opam anyway, but I never dig deeply enough in opam+dune to clearly understand what the problem was. Anyway, it is not used anymore.

packages/acgtk/acgtk.1.5.3/opam
```diff
@@ -0,0 +1,45 @@
...
+install: ["dune" "install"]
+
+depends: [
+  "ocaml" { >= "4.05.0" }
+  "dune" { >= "1.4" }
+  "menhir"
+  "ANSITerminal"
+  "fmt"
+  "logs"
+  "mtime"
```
mseri, line 21:
```suggestion
  "mtime" {>= "1.0.0"}
```
This is where `mtime.clock.os` was introduced

packages/acgtk/acgtk.1.5.3/opam
```diff
@@ -0,0 +1,45 @@
...
+
+depends: [
+  "ocaml" { >= "4.05.0" }
+  "dune" { >= "1.4" }
+  "menhir" {>= "20181113"}
+  "ANSITerminal"
+  "fmt"
+  "logs"
+  "mtime" {>= "1.0.0"}
+  "cmdliner"
```
mseri, line 22:
```suggestion
  "cmdliner" {>= "1.0.0"}
```
This is wheer `Manpage.s_description` should have been introduced

mseri:
Looks good now, sorry for the delay

PR #19043, opened 2021-07-12
Add conf-mpfr.3: Use pkg-config to detect the library

### LINKED ISSUE #19042: conf-mpfr fails on Apple Silicon M1 macOS due to /opt/homebrew instead of /usr/local
[Only found #18167 searching for similar issues. Tried to add some keywords in the title.]

`brew install gmp && opam install apron` fails with `fatal error: 'mpfr.h' file not found`.
`C_INCLUDE_PATH=/opt/homebrew/include LIBRARY_PATH=/opt/homebrew/lib opam install apron` installs `conf-mpfr` but then fails when compiling apron with `fatal error: 'gmp.h' file not found`.
(`opam install apron` only succeeded after `ln -s /opt/homebrew/{include,lib} /usr/local/`)

https://docs.brew.sh/FAQ#why-should-i-install-homebrew-in-the-default-location
`echo $HOMEBREW_PREFIX` is `/usr/local` on Intel, but changed to `/opt/homebrew` on ARM and they advise against changing it.

The paths are set here:
https://github.com/ocaml/opam-repository/blob/c0e9300f14d3570da85aad7e6e0ae47484a597a9/packages/conf-mpfr/conf-mpfr.2/opam#L8-L9

Should we add another case for `arch = "arm64" & os = "macos" & os-distribution = "homebrew"`?

Does it make sense for opam to provide some [variable](https://opam.ocaml.org/doc/Manual.html#Global-variables) that can be used for `-I` and `-L`? Otherwise this switching would have to be duplicated in each opam file that does something like this.

kit-ty-kate:
Fixes #19042

### DIFF
--- a/packages/conf-mpfr/conf-mpfr.3/files/test.c
+++ b/packages/conf-mpfr/conf-mpfr.3/files/test.c
@@ -1,0 +1,11 @@
+#include <stdio.h>
+#include <mpfr.h>
+
+// compile with: gcc test.c -lmpfr -lgmp
+
+int main (void) {
+  printf ("MPFR library: %-12s\nMPFR header:  %s (based on %d.%d.%d)\n",
+          mpfr_get_version (), MPFR_VERSION_STRING, MPFR_VERSION_MAJOR,
+          MPFR_VERSION_MINOR, MPFR_VERSION_PATCHLEVEL);
+  return 0;
+}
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
cc @UnixJunkie

UnixJunkie:
thanks for the fix!

PR #19050, opened 2021-07-13
[new release] randoml (0.1.3)

mimoo:
Generating cryptographically-secure random numbers

- Project page: <a href="https://github.com/mimoo/randoml">https://github.com/mimoo/randoml</a>

##### CHANGES:

Fix vendoring of rust dependencies for release (bis)

### DIFF
--- a/packages/randoml/randoml.0.1.3/opam
+++ b/packages/randoml/randoml.0.1.3/opam
@@ -1,0 +1,41 @@
+opam-version: "2.0"
+synopsis: "Generating cryptographically-secure random numbers"
+description:
+  "A library relying on Rust rand to generate random numbers. The random numbers are cryptographically secure."
+maintainer: ["david@o1labs.org"]
+authors: ["David Wong"]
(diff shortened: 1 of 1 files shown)


mseri:
https://github.com/mimoo/randoml/blob/583d86b55d63ad6ed5fce097df4d105918ee7a35/lib/rand/dune#L3: I think you need to copy `vendor` into _build, otherwise you are running `cargo build` without the vendored sources present.
Also, you don't need the quotes there in the dune file.

mseri:
To test it, you can run `opam install -t .` from the root of the project

mimoo:
ok thanks for the pointers! I'll get it right before submitting here :) thanks a lot

mimoo:
I think the issue is that it's not copying the `.cargo` folder in `_build`

mimoo:
and I need `vendor` in `(dirs :standard \ "target" "vendor")` because it has some broken dune files in it

PR #19051, opened 2021-07-13
js_of_ocaml.3.4.0: require ppx_tools_versioned 5.1

emillon:
It uses `ppx_tools_versioned.metaquot_406` that has been added in that version.


emillon:
I don't think I can fix missing license easily since the license is not that straightforward and there is no valid version to copy from, so I prefer to leave this one to the opam-repository maintainers.

mseri:
@hhugo would it be possible to send a PR adding the missing licenses from older js_of_ocaml packages?

mseri:
Thanks @emillon

PR #19052, opened 2021-07-14
Remove references to the OCaml forge

garrigue:
Old lablgtk packages still contained references to the OCaml forge.
While there is a redirection to github from there, it goes to repository page rather than the project home page.
This fixes these links, and homogenizes the packages.
The only remaining references to ocamlcore are the download links, which are still supported.


kit-ty-kate:
Thanks a lot!

PR #19054, opened 2021-07-14
[new release] reparse, reparse-lwt and reparse-lwt-unix (3.1.0)

bikallem:
Recursive descent parsing library for ocaml

- Project page: <a href="https://github.com/lemaetech/reparse">https://github.com/lemaetech/reparse</a>

##### CHANGES:

- Add `unsafe_take_cstruct_exact` and `unsafe_take_cstruct_ne` parsers.
- `PARSER.parse` function now takes `?pos` parameter and returns `pos` along with the parsed value. This allows creating pull based parsers.
- Expose `Promise` module.

PR #19056, opened 2021-07-14
Http multipart formdata upper bound

bikallem:
Add upper bounds to package http-multipart-formdata v2.0.0 and v2.0.1 in preparation for reparse v3.1.0 release.
https://github.com/ocaml/opam-repository/pull/19054


mseri:
Thanks

PR #19060, opened 2021-07-15
Add opam-bin.1.1.0

lefessan:


kit-ty-kate:
Thanks!

smorimoto:
I lost quite a bit of this boat, but I think we should have set the "plugin" flag.

PR #19061, opened 2021-07-15
[new release] conf-rust-1-53 (0.1)

mimoo:
Virtual package relying on cargo (rust build system)

- Project page: <a href="https://github.com/mimoo/conf-rust-1-53">https://github.com/mimoo/conf-rust-1-53</a>

##### CHANGES:

First version.

### DIFF
--- a/packages/conf-rust-1-53/conf-rust-1-53.0.1/opam
+++ b/packages/conf-rust-1-53/conf-rust-1-53.0.1/opam
@@ -1,0 +1,27 @@
+opam-version: "2.0"
+maintainer: "david@o1labs.org"
+authors: "David Wong"
+homepage: "https://github.com/mimoo/conf-rust-1-53"
+bug-reports: "https://github.com/mimoo/conf-rust-1-53/issues"
+dev-repo: "git+https://github.com/mimoo/conf-rust-1-53.git"
+license: "MIT"
+build: [
+  ["cargo" "--version"]
+  ["cargo" "+1.53.0" "build"]
+]
+depends: [
+  "conf-rust-2018"
+]
+synopsis: "Virtual package relying on cargo (rust build system)"
+description:
+  "This package can only install if cargo (edition=2018) is installed on the system with rust 1.53.0-x86_64-apple-darwin."
+flags: conf
+url {
+  src:
+    "https://github.com/mimoo/conf-rust-1-53/releases/download/0.1/conf-rust-1-53-0.1.tbz"
+  checksum: [
+    "sha256=0301a76554739cf82486a76c6ba15ddd397284f97e4376aef4c74c2a698e8350"
+    "sha512=d1e68341e03f559d7688a068c7ea93b64641896a8569594bd73f79dcf6c5a504ecce4e6db22054cdd7177e72be58567095ae4a043133cb4702e2e93fd9e6e801"
+  ]
+}
+x-commit-hash: "f003328b66ecf965d88502b6af15c27a35f14aa9"


mimoo:
So I'm assuming it is normal that the job fails everywhere, as no system has that specific version of rust installed. I think I'd be in the camp of "opam should let cargo access the interwebs"

mseri:
Sorry, I think I misled you with my comment on the other PR.

If I remember correctly, we ended up using the "editions" because it would be too much to have a large number of conf-rust packages for each version. I would suggest, on your other PR, to use conf-rust and add a post-install failure message explaining that you need at least rust 1.53 for the package

mseri:
I am going to pink @kit-ty-kate, she maybe recalls the discussion and what was decided then

packages/conf-rust-1-53/conf-rust-1-53.0.1/opam
```diff
@@ -0,0 +1,27 @@
...
+flags: conf
+url {
+  src:
+    "https://github.com/mimoo/conf-rust-1-53/releases/download/0.1/conf-rust-1-53-0.1.tbz"
+  checksum: [
+    "sha256=0301a76554739cf82486a76c6ba15ddd397284f97e4376aef4c74c2a698e8350"
+    "sha512=d1e68341e03f559d7688a068c7ea93b64641896a8569594bd73f79dcf6c5a504ecce4e6db22054cdd7177e72be58567095ae4a043133cb4702e2e93fd9e6e801"
+  ]
+}
+x-commit-hash: "f003328b66ecf965d88502b6af15c27a35f14aa9"
```
kit-ty-kate, line 27: I’m quite confused as to why your meta package has an archive associated with it.

mseri: It contains the dummy sources for cargo to build.

My question was more related to how "fine" should conf-rust packaging be, I thought we had come to the middle ground to package a generic "rust" and the editions

mseri: Found it, I think nothing was really decided: https://github.com/ocaml/opam-repository/pull/13351

packages/conf-rust-1-53/conf-rust-1-53.0.1/opam
```diff
@@ -0,0 +1,27 @@
+opam-version: "2.0"
+maintainer: "david@o1labs.org"
+authors: "David Wong"
+homepage: "https://github.com/mimoo/conf-rust-1-53"
+bug-reports: "https://github.com/mimoo/conf-rust-1-53/issues"
+dev-repo: "git+https://github.com/mimoo/conf-rust-1-53.git"
+license: "MIT"
+build: [
+  ["cargo" "--version"]
+  ["cargo" "+1.53.0" "build"]
```
kit-ty-kate, line 10: This package can basically be summerised by this command. If it’s a requirement of your package, why not add this check to your configure/build step early and add a meaningful error there? If opam-repository had a package for all the possible preconditions (bash version, gcc version, autoconf version, …) it wouldn’t be manageable in the current state.

packages/conf-rust-1-53/conf-rust-1-53.0.1/opam
kit-ty-kate: yeah, idk, adding conf-rust-2018 was probably a mistake in hinsight

mseri:
I have added a suggestion with post-message in case of failure in the package that would have used this. I am closing the PR since there is no consensus on what to do and we feel that having a package for each rust version would be too much

PR #19075, opened 2021-07-17
Slug 1.0.1

thangngoc89:
Patch releases that makes Slug's behavior consistent with node-slugify.
String is normalized with NFC instead of NFD

### DIFF
--- a/packages/slug/slug.1.0.1/opam
+++ b/packages/slug/slug.1.0.1/opam
@@ -1,0 +1,34 @@
+opam-version: "2.0"
+authors: "Khoa Nguyen"
+homepage: "https://github.com/thangngoc89/ocaml-slug"
+maintainer: "hi@khoanguyen.me"
+dev-repo: "git+ssh://git@github.com:thangngoc89/ocaml-slug.git"
+bug-reports: "https://github.com/thangngoc89/ocaml-slug/issues"
+license: "MIT"
+build: [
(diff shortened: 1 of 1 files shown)


packages/slug/slug.1.0.1/opam
```diff
@@ -0,0 +1,34 @@
...
+bug-reports: "https://github.com/thangngoc89/ocaml-slug/issues"
+license: "MIT"
+build: [
+  [ "dune" "subst" ] {dev}
+  [ "dune" "build" "-p" name "-j" jobs ]
+]
+depends: [
+  "ocaml" {>= "4.05"}
+  "dune" {>= "2.0"}
+  "opam-lock" {dev}
```
kit-ty-kate, line 15: why is opam-lock required in dev mode? I can’t find any mention of it in the source code

kit-ty-kate:
The failures can be ignored btw.

packages/slug/slug.1.0.1/opam
thangngoc89: I have no ideas. I copied this from 1.0.0. Let’s me delete this

kit-ty-kate:
Thanks! Could you return the fix in the opam file upstream so that it can be taken into account for next time?

thangngoc89:
Sure. Thank you very much

PR #19082, opened 2021-07-19
[new release] ppx_deriving_cmdliner (0.6.1)

trepetti:
**CHANGES:**

- Bump minimum ppxlib version to 0.18.0 to fix https://github.com/hammerlab/ppx_deriving_cmdliner/issues/49.


mseri:
Is the change just the lower bound on ppxlib or there are other changes? In the first case we should only update the opam metadata of the 0.6.0 release

mseri:
I went to check and when we released 0.6.0, we did not yet have the lower bounds check on the CI. Now that this kind of checks are part of the CI, it will not happen again :)

trepetti:
Thanks for fixing the previous release!

PR #19083, opened 2021-07-19
[new release] biocaml (0.11.2)

pveber:
The OCaml Bioinformatics Library

- Project page: <a href="https://github.com/biocaml/biocaml/">https://github.com/biocaml/biocaml/</a>

##### CHANGES:

* fixed incompatibilities with version 4.13 of the compiler

### DIFF
--- a/packages/biocaml/biocaml.0.11.2/opam
+++ b/packages/biocaml/biocaml.0.11.2/opam
@@ -1,0 +1,55 @@
+opam-version: "2.0"
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+  ["dune" "build" "-p" name "@doc"] {with-doc}
+]
+maintainer: ["philippe.veber@gmail.com"]
+authors: [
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
I just saw the failure on biocaml.0.11.1 today and was going to look into it.
Are you sure this is not a compiler bug? The fix seems suspicious

kit-ty-kate:
It looks like this was fixed by https://github.com/ocaml/ocaml/pull/10529. Could you confirm?

pveber:
Thanks a lot for looking into this @kit-ty-kate! I was too shy when I saw the compile error and thought this was the result of a recent change I had missed. Ok then, so I have to work on a new release for biocaml excluding the changes related to this. BTW I've recently started to use `dune-release` and sometimes it reopens a new PR when I force-push a new version of a release sometimes it doesn't. I couldn't find how to make sure I was using the same PR so sorry the same release jumps to another PR!

pveber:
Actually, if this problem is fixed in the compiler, releasing now is a lot less useful. Let's save that for another time. Thanks again!

kit-ty-kate:
No worries, i should've looked into it a month ago but i was busy with something else and didn't have the time

PR #19099, opened 2021-07-21
stale.yml: allow persistent issues

mseri:
There are some issues that we continuously mark as unstale and that we want to keep around for longer. Recent stale bots have the feature to ignore certain issues using some specific label.

I suggest to enable this on the opam-repository


kit-ty-kate:
Oh there is a config file for that! I'm glad to know! Nice catch, thanks a lot!

PR #19100, opened 2021-07-21
frama-c: new release (23.1)

AllanBlanchard:


packages/frama-c/frama-c.23.1/opam
```diff
@@ -0,0 +1,150 @@
...
+  "abstract interpretation"
+  "slicing"
+  "weakest precondition"
+  "ACSL"
+  "dataflow analysis"
+  "runtime verification"
+]
+
+build: [
+  ["autoconf"] {pinned}
```
mseri, line 86:
```suggestion
  ["autoconf"] {dev}
```

mseri: Besides this, all the failures are the usual one

AllanBlanchard: Is this something new? (we have not changed this line for years) I tried to find some documentation on this parameter but without success.

mseri: No, it has never been correct. I don't think it works as expected if you try. I am not sure how it started, dune opam file generation had the same bug for a long time, it was fixed only recently. The correct flag for pinned packages is `dev`: https://opam.ocaml.org/doc/Manual.html#pkgvar-dev

mseri: Apologies for not spotting it before

mseri: I edited my first comment since it was not completely correct what I was saying

mseri:
Since my comment relates to something that is there since `frama-c.17.1`, we can fix them all in bulk in a separate PR if needed. I am going to ping @kit-ty-kate that knows better the internals of opam

packages/frama-c/frama-c.23.1/opam
AllanBlanchard: In fact, the simplest fix we can do is probably to entirely remove the `autoconf` dependency on the Opam repository :thinking: . We provide the `configure` file when we release Frama-C, and our development version (that can be pinned) uses a different `opam` file.

PR #19101, opened 2021-07-21
Sync opam's opam files with upstream

dra27:
cf. https://github.com/ocaml/opam/pull/4757. Linting passes and dependencies tweaks.


packages/opam-devel/opam-devel.2.0.8/opam
```diff
@@ -17,15 +17,19 @@ authors: [
...
-  "cmdliner" {>= "0.9.8"}
+  "cmdliner" {>= "1.0.0"}
   "dune" {>= "1.2.1"}
+  "conf-openssl" {with-test}
+  "conf-diffutils" {with-test}
 ]
 build: [
   ["./configure" "--disable-checks" "--prefix" prefix]
   [make "%{name}%.install"]
+  [make "tests"] {with-test}
```
kit-ty-kate, line 32: This was removed on purpose. See https://github.com/ocaml/opam/pull/4500

dra27: I thought it was removed from 2.1 on purpose - does it affect 2.0 as well (different testing system)?

rjbou: yes, it is the old testing system in 2.0

dra27: It still requires a system compiler, even if no network. There's no totally principled way in opam-repository at the moment of depending on the presence of a _system_ compiler itself (it's not that we need to depend on `ocaml-system`, it's that we actually need the depext packages for ocaml installed) so given that tests in opam-devel 2.1 later can't be run in opam anyway, I've removed this.

kit-ty-kate:
Thanks!

PR #18392, opened 2021-03-25
Partly revert #17577

mseri:
This should allow us to test for compatibility and eventual further breakages in the [cohttp 4.0.0 release](https://github.com/ocaml/opam-repository/pull/18385).

Signed-off-by: Marcello Seri <marcello.seri@gmail.com>

PR #19104, opened 2021-07-22
Prometheus-app also works with cohttp 4

talex5:
Same fix as for prometheus.1.0 (and now fixed upstream too):
https://github.com/ocaml/opam-repository/pull/18392/files#diff-60879fe2781d9550236f9fdbe6d36f97b30c6563e42b76355e57c9c3770b2419


kit-ty-kate:
Thanks!

PR #19110, opened 2021-07-23
[new release] pyre-ast (0.1.3)

grievejia:
CHANGES:

- Fix crashes when parsing non-UTF8-decodable string literals.
- \N escape sequence in string literals will no longer cause a syntax error.
- Module parsing APIs now accept an additional `enable_type_comment` argument, controlling whether to parse type comments or not.


mseri:
Thanks

PR #19111, opened 2021-07-23
Mark mirage-dns as deprecated (new name is dns-mirage)

kit-ty-kate:
cc @hannesm


hannesm:
it is indeed deprecated, the successor is a variety of packages, mainly dns-server.mirage and dns-client.mirage. I also have no clue what "lags: deprecated" actually does.

kit-ty-kate:
Currently it does nothing apart from making opam-health-check ignore those packages.
In opam 2.2.0 it will hopefully be used by opam to avoid the non-explicit installation of the package and print a warning if it is explicitly installed. See the https://github.com/ocaml/opam/issues/4522 proposal

Issue #18167, opened 2021-02-16
conf-libev.4-1 finding libev on macOS M1 with /opt/homebrew installation


lindig:
I'm not sure how to report this properly. The opam file for conf-libev.4-1 asks to file issues here:

`conf-libev.4-1` fails to find libev on macOS with a homebrew installation of libev, which is located under `/opt/homebrew/` for the new ARM-based Macs :

```
lindig@mini conf-libev.4-11 % sh build.sh 
testing for libev: ............................. unavailable
```

However, with a bit of direction it does find it:

```
lindig@mini conf-libev.4-11 % C_INCLUDE_PATH=/opt/homebrew/include LIBRARY_PATH=/opt/homebrew/lib sh build.sh
testing for libev: ............................. available
```

Is there a way to include this location in `conf-libev.4-11`?

lindig:
I suspect all that is needed is adding `/opt/homebrew` here:

```
let search_paths = uniq (List.append [
  "/usr";
  "/usr/local";
  "/opt";
  "/opt/local";
  "/sw";
  "/mingw";] env_paths)
```

mseri:
Can you try it locally and see if it fixes the issue for you? I'd be happy to merge the PR with the fix

mseri:
I think this needs also some extra change somewhere.
The `discover.ml` file in there mentions:
```ocaml
(* Keep that in sync with the list in myocamlbuild.ml *)
```
I wonder to what package that belongs/belonged

lindig:
I'll try it locally later today. Obviously this is just a test that look for the header files. Any actual package still needs to find the library to link against it. I'm not familiar with this distribution of labour between `conf-libev` and packages using `libev`.

lindig:
The change (adding `/opt/homebrew`) makes it work for me and consequently I could install packages that depend on this one. I am still a bit mystified how all of this fits together because any code using `libev` must look in the correct place to begin with.

mseri:
@kit-ty-kate has attempted a pkg-config based approach using conf-ssl which may simplify things moving forward. Of course, it depends a lot on what packages are doing to then link the libraries, if they use pkg-comfig or not, if they override the search paths, ...

in this case I don’t know, I would make a PR for a new version of conf-libev which includes the new search path

zapashcanon:
I believe that [leveldb#15](https://github.com/mfp/ocaml-leveldb/issues/15) is exactly the same but for conf-leveldb. @charlesetc

[...]

PR #19114, opened 2021-07-23
Conf-leveldb for leveldb >= 1.21

thangngoc89:
[Since LevelDb >= 1.21, C++ 11 is required](https://github.com/google/leveldb/releases/tag/1.21) 

This add `-std=c++11` to build command .

More details could be see here: https://github.com/ocaml/opam-repository/issues/18167#issuecomment-885459238


kit-ty-kate:
Looks good. Thanks!

thangngoc89:
@kit-ty-kate Thanks. This only fixes `conf-leveldb`. The same patch must be applied to `leveldb` package itself. I opened a PR here https://github.com/mfp/ocaml-leveldb/pull/21

PR #19116, opened 2021-07-24
Package asak.0.3

nobrakal:
### `asak.0.3`
Partitioning OCaml codes
Asak provides functions to parse, type-check and identify similar OCaml codes.
These functions are then used to partition codes implementing the same function and help to detect code duplication.

---
* Homepage: https://github.com/nobrakal/asak
* Source repo: git+https://github.com/nobrakal/asak.git
* Bug tracker: https://github.com/nobrakal/asak/issues

---

### DIFF
--- a/packages/asak/asak.0.3/opam
+++ b/packages/asak/asak.0.3/opam
@@ -26,7 +26,7 @@ depends: [
 ]
 
 build: [
-  ["dune" "subst"] {pinned}
+  ["dune" "subst"] {dev}
   ["dune" "build" "-p" name "-j" jobs]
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
`dune subst` should only be used when using the dev version, not when pinned. See https://github.com/ocaml/dune/pull/3647

Could you return the fix in the opam file upstream so that it can be taken into account for next time? (no need to change this PR or its archive, everything is fixed already)

kit-ty-kate:
Thanks!

nobrakal:
I backported the change to upstream, thanks for the quick merge :)

PR #19119, opened 2021-07-25
Package pla.1.4

modlfo:
### `pla.1.4`

---

---


kit-ty-kate:
Thanks! Could you return the fixes upstream? (dune has a bug that makes it not work with the build tag, see: https://github.com/ocaml/dune/issues/2147, and OMP is a runtime dependency of pla.ppx)

PR #16244, opened 2020-04-20
ocaml-secondary-compiler: --disable-graph-lib (disable Graphics)

dra27:
Continuation of #16236 in a bid to fix #16226.

@AltGr / @rjbou - https://github.com/ocaml/opam-repository/commit/33e22b6f3dc30101bd35f0003c86716cd248e47f expresses the same requirement using `depends` instead of `conflicts` but that's not ideal, since that would attempt to change the installed compiler, rather than reporting the conflict (I think)? What the best way to express the conflict with all 4.08 versions of `ocaml`?

PR #19130, opened 2021-07-26
Use conflicts instead of depends for ocaml-secondary-compiler

dra27:
In https://github.com/ocaml/opam-repository/pull/16244, I fixed an incorrect use of `&` in `conflicts` for `ocaml` by using `depends` for the `ocaml-secondary-compiler` package.

It was a slightly dirty change (it's not great to depend on something you actually don't depend on) and I've hit an annoyance with it  in CI. On the basis that OCaml 4.08 is not going to change, I have instead restored the use of `conflicts` but this time just listing all releases of 4.08 instead.

Prevents unnecessary rebuild on system compiler upgrade. I was trying to use this in CI in order to cache a build of ocaml-secondary-compiler in a switch using ocaml-system and then upgrade the system compiler  to the required version for each item in a build matrix.


kit-ty-kate:
I'm unsure why there is such a conflict in the first place. Can they not be removed all together?

dra27:
I think the reasoning of whoever it was wrote it (!!) was that having 4.08.1 in a 4.08.0/4.08.2 in the same switch was a very strange thing to do, so it was meant to stop it being unnecessarily installed by accident.

To be honest, I'd happily see no constraint - it's actually been a nuisance in opam CI, as I have to download a cache of a 4.03 system compiler to install `ocaml-secondary-compiler` rather than more quickly installing the OS packager's 4.08 compiler.

kit-ty-kate:
I've cancelled the revdeps to avoid CI constraints

PR #19128, opened 2021-07-26
[new release] merlin 4.3.1-412/411 and 3.6.1

voodoos:
## Changes for 4.3.1-412 / 411

  + merlin binary
    - recover ill-typed patterns (ocaml/merlin#1317, ocaml/merlin#1342)
    - more accurate type-enclosing for methods (ocaml/merlin#1328, fixes ocaml/merlin#1124)
    - fix location of patterns in Occurrences (ocaml/merlin#1324, fixes ocaml/ocaml-lsp#375)
    - fix location of module definitions done via functors (ocaml/merlin#1329, fixes ocaml/merlin#1199)
    - fix -cmt-path dirs mistakenly added to build path (ocaml/merlin#1330)
    - add new module holes that can replace module expressions (ocaml/merlin#1333)
    - add a new command `construct` that builds a list of possible terms when
      called on a typed hole (ocaml/merlin#1318)
    - `refactor-open` improvements (ocaml/merlin#1313, ocaml/merlin#1314, ocaml/merlin#1366, ocaml/merlin#1372)
      - do not make paths absolute, simply prefix with the identifier under
      the cursor
        ```ocaml
        open Foo (* calling refactor-open qualify on this open *)
        let _ = Foo.bar (* previously could result in [Dune__exe.Foo.bar] *)
        ```
      - do not return identical (duplicate) edits
      - do not return unnecessary edits that when applied do not change
        the document
      - handle record fields properly
      - handle multi-line paths
      - `unqualify` should not qualify
    - Handle `Persistent_env.Error` in `Typemod.initial_env` (ocaml/merlin#1355)
    - locate: reset global state from all entry points (ocaml/merlin#1364)
    - Windows: replace user name by its SID in socketnames (ocaml/merlin#1345, @ttamttam)
  + editor modes
    - vim: add a simple interface to the new `construct` command:
      `MerlinConstruct`. When several results are suggested, `<c-i>`
      and `<c-u>` can be use to change the depth of the recursive
      construction. (ocaml/merlin#1318)
    - vim: add support for the `merlin-locate-type` command:
      `MerlinLocateType` (ocaml/merlin#1359)
    - emacs: add a simple interface to the new `construct` command:
      `merlin-construct`. (ocaml/merlin#1352)
    - emacs: add support for the `merlin-locate-type` command. (ocaml/merlin#1359)
    - emacs: fix issue with `merlin--highlight` and  various minor improvements
        (ocaml/merlin#1367, @mattiase)
  + test suite
[...]

PR #19131, opened 2021-07-26
mdx: Add missing constraints

kit-ty-kate:
Spotted in #19128


avsm:
Just tests failing now so this an improvement; merging.

PR #18940, opened 2021-06-25
[new release] cstruct-sexp, cstruct, cstruct-unix, ppx_cstruct, cstruct-lwt and cstruct-async (6.0.1)

dinosaure:
S-expression serialisers for C-like structures

- Project page: <a href="https://github.com/mirage/ocaml-cstruct">https://github.com/mirage/ocaml-cstruct</a>
- Documentation: <a href="https://mirage.github.io/ocaml-cstruct/">https://mirage.github.io/ocaml-cstruct/</a>

##### CHANGES:

**breaking changes**
- `Cstruct.len` is deprecated, it will be deleted at the next
  release. You should use `Cstruct.length` instead.
  (@dinosaure, @hannesm, mirage/ocaml-cstruct#284)

- Remove color from ppx's binary to replicate diff of errors in
  any contexts (@dinosaure, @sternenseemann, mirage/ocaml-cstruct#285)
- Add `shiftv` (@talex5, @avsm, @dinosaure, mirage/ocaml-cstruct#287)
- Use `Bytes_val` if available (@hannesm, @avsm, @dinosaure, mirage/ocaml-cstruct#286)

PR #19142, opened 2021-07-27
mirage-block-lwt isn't compatible with cstruct 6.0.1

kit-ty-kate:
Fix revdeps for https://github.com/ocaml/opam-repository/pull/18940


kit-ty-kate:
cc @hannesm

PR #19143, opened 2021-07-27
Fix cstruct.6.0.1's revdeps (use -warn-error and Cstruct.len is now deprecated)

kit-ty-kate:
Fix the remaining failures from https://github.com/ocaml/opam-repository/pull/18940


kit-ty-kate:
cc @hannesm for the mirage packages
cc @abeaumont for `salsa20-core`, `scrypt-kdf` and `pbkdf`

PR #19147, opened 2021-07-27
Disable tests of some packages

kit-ty-kate:
the sandbox now disables writes into /tmp since opam 2.0.9


kit-ty-kate:
Will do separately later

PR #17642, opened 2020-11-16
Package opam-depext.1.1.5

rjbou:
### `opam-depext.1.1.5`
install OS distribution packages
opam-depext is a simple program intended to facilitate the interaction between
OPAM packages and the host package management system. It can query OPAM for the
right external dependencies on a set of packages, depending on the host OS, and
call the OS's package manager in the appropriate way to install them.

---
* Homepage: https://github.com/ocaml/opam-depext
* Source repo: git+https://github.com/ocaml/opam-depext.git#2.0
* Bug tracker: https://github.com/ocaml/opam-depext/issues

---
:camel: Pull-request generated by opam-publish v2.0.3

PR #17435, opened 2020-10-19
Package opam-depext.1.1.4

rjbou:
### `opam-depext.1.1.4`
install OS distribution packages
opam-depext is a simple program intended to facilitate the interaction between
OPAM packages and the host package management system. It can query OPAM for the
right external dependencies on a set of packages, depending on the host OS, and
call the OS's package manager in the appropriate way to install them.

---
* Homepage: https://github.com/ocaml/opam-depext
* Source repo: git+https://github.com/ocaml/opam-depext.git#2.0
* Bug tracker: https://github.com/ocaml/opam-depext/issues

---
:camel: Pull-request generated by opam-publish v2.0.3

PR #19160, opened 2021-07-29
Add missing opam constraint to recent depext releases

dra27:
#17435 & #17642 are missing the constraint away from opam 2.1 added in #16325


mseri:
Thanks

PR #19168, opened 2021-07-31
ELINA Octagon domain support with OCaml

GgnDpSngh:


packages/elina/elina.1.1/opam
```diff
@@ -1,22 +1,24 @@
 opam-version: "2.0"
 authors: ["Gagandeep Singh" "Markus Püschel" "Martin Vechev"]
 homepage: "http://elina.ethz.ch/"
-maintainer: "Gagandeep Singh <gsingh@inf.ethz.ch>"
+maintainer: "Gagandeep Singh <ggnds@illinois.edu>"
 dev-repo: "git+https://github.com/eth-srl/ELINA.git"
 bug-reports: "https://github.com/eth-srl/ELINA/issues"
+version: "1.3"
```
mseri, line 7: You are overwriting version 1.1, please move the changes to elina.1.3/opam

packages/elina/elina.1.1/opam
```diff
@@ -34,7 +36,9 @@ depends: [
   "conf-perl"
   "apron"
 ]
+flags: light-uninstall
```
mseri, line 39: This flag as well as the “remove” section are no longer necessary with opam 2

GgnDpSngh: Thanks for the feedback, let m try to fix it and send another request.

PR #19173, opened 2021-08-01
Package tsort.2.1.0

dmbaturin:
### `tsort.2.1.0`
Easy to use and user-friendly topological sort
Easy to use and user-friendly topological sort.

Example:
```
Tsort.sort [("foundation", []); ("walls", ["foundation"]); ("roof", ["walls"])]
```

---
* Homepage: https://github.com/dmbaturin/ocaml-tsort
* Source repo: git+https://github.com/dmbaturin/ocaml-tsort.git
* Bug tracker: https://github.com/dmbaturin/ocaml-tsort/issues

---


dmbaturin:
The reason why d2b7115 didn't fix anything seems to be that sihl lacks containers dependency not only in its opam file, but also in the dune file, so containers were both installed and linked only as a transitive dependency of tsort. I hope locking it to 2.0.0 will help.

mseri:
Ping @jerben

dmbaturin:
@mseri @jerben Looking at the current CI run, there are still quite a few lower bounds test failures in sihl, but they are related to other dependencies, not to tsort and containers.

dmbaturin:
Ok, I've just done a full review of the last CI run and I couldn't find any failure where tsort would be the cause. I consider the issue solved from my side (even if the issue on hand was never my fault, to begin with ;).

mseri:
Thanks, I agree. We can fix sihl lower bounds separately

jerben:
Thanks for the fix. I will have a look regarding Sihl lower bounds.

PR #19178, opened 2021-08-02
Add ocamlfind-secondary.1.9.1

kit-ty-kate:
Just to be able to compile it with OCaml 4.13 if need be on day and to avoid reliance on m4 (removed in ocamlfind 1.9.0)


dra27:
Thanks, @kit-ty-kate - while looking at opam's CI updates, I'd also put getting ocamlfind-secondary.1.9.1 on my todo list 🙂

Using `"ocamlfind" {= version}` is much clearer than the hard-coded version I did in 1.8.1 - I've pushed a commit changing that in the 1.8.1 version as well and pushed commits to pass the lint checks. In practice, ocamlfind-secondary won't pin (because of the extra-files), but it may as well point to upstream ocamlfind.

PR #19184, opened 2021-08-03
Package samplerate.0.1.6

toots:
### `samplerate.0.1.6`
Samplerate audio conversion library
Bindings for the samplerate library which provides functions for changing samplerate of audio data

---
* Homepage: https://github.com/savonet/ocaml-samplerate
* Source repo: git+https://github.com/savonet/ocaml-samplerate.git
* Bug tracker: https://github.com/savonet/ocaml-samplerate/issues

---


mseri:
The failures seems to be due to to an incompatible version of the depexts, is that right? In that case I think we can merge

toots:
Yeah i think the package is too old on those OSes.

PR #19189, opened 2021-08-03
nnpchecker available on 4.14 for ARM64

dra27:
Mergeable after https://github.com/ocaml/ocaml/pull/10549.


dra27:
The upstream ocaml PR has been merged.

PR #19207, opened 2021-08-04
4 packages from c-cube/qcheck at 0.18

c-cube:
This pull-request concerns:
-`qcheck.0.18`: Compatibility package for qcheck
-`qcheck-alcotest.0.18`: Alcotest backend for qcheck
-`qcheck-core.0.18`: Core qcheck library
-`qcheck-ounit.0.18`: OUnit backend for qcheck

---
* Homepage: https://github.com/c-cube/qcheck/
* Source repo: git+https://github.com/c-cube/qcheck.git
* Bug tracker: https://github.com/c-cube/qcheck/issues

---

### DIFF
--- a/packages/alg_structs_qcheck/alg_structs_qcheck.0.1.3/opam
+++ b/packages/alg_structs_qcheck/alg_structs_qcheck.0.1.3/opam
@@ -16,7 +16,7 @@ depends: [
   "dune" {>= "1.11.3"}
   "ocaml" {>= "4.08.0"}
   "alg_structs" {>= "0.1.3" & < "0.2.0"}
-  "qcheck" {>= "0.11"}
+  "qcheck" {>= "0.11" & < "0.18"}
   "alcotest" {with-test & >= "0.8.5"}
   "qcheck-alcotest" {with-test & >= "0.11"}
 ]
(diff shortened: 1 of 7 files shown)


avsm:
Quite a few lint failures and a missing alcotest dependency.

mseri:
I think you need to update the tarball with the new fixes, right?

mseri:
Looks almost good now. I need to send a separate PR with upper bounds for

fadbadml.0.1.2 (failed: Unbound record field QCheck.TestResult.state)
alg_structs_qcheck.0.1.3 (failed: Signature mismatch:)
preface.0.1.0 (failed: The implementation test/preface_qcheck/arbitrary.ml)

There is a segfault in the tests for all architectures different from x86_64, is this expected?

c-cube:
not really expected but at this point I'm tired of this, I don't even know how we'd test on 32 bits. is there a constraint for "64 bits only"?

mseri:
Yes, we can just set available flag for specific architectures, however the failure happens only in qcheck-core and in some cases, are you sure you want to flag it unavailable for all non-x86_64 architectures?

If it is not something too worrisome, I would suggest that we leave it as is, I add the upper bounds on monday and then merge. If the issue starts popping up, we could add the flag later on

c-cube:
ok, sounds good!

mseri:
Sorry for the delay!

PR #19209, opened 2021-08-05
New bugfix release of not-ocamlfind with better syntax predicate support

chetmurthy:
formerly, a syntax predicate gets added to "syntax_preds".
with this patch, "syntax_" prepended to that predicate
also gets added to "predicates".

### DIFF
--- a/packages/not-ocamlfind/not-ocamlfind.0.08/opam
+++ b/packages/not-ocamlfind/not-ocamlfind.0.08/opam
@@ -1,5 +1,6 @@
 opam-version: "2.0"
 synopsis: "A small frontend for ocamlfind that adds a few useful commands"
+license: "MIT"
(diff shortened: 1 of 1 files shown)


packages/not-ocamlfind/not-ocamlfind.0.08/opam
```diff
@@ -0,0 +1,36 @@
+opam-version: "2.0"
+synopsis: "A small frontend for ocamlfind that adds a few useful commands"
+license: "MIT"
+maintainer: "Chet Murthy <chetsky@gmail.com>"
+
+(* Gerd wrote most of this code; I just modified it (and probably
+introduced bugs.  This is to silence opam *)
+
+authors: "Chet Murthy <chetsky@gmail.com>"
```
avsm, line 9: Regarding the comment, the authors field can also be a list of authors so you could include Gerd as well in a future PR.

PR #19210, opened 2021-08-05
camlp5 rel8.00.02: updates to support OCaml 4.13.0

chetmurthy:
New OCaml 4.13 syntax supported, with tests.


kit-ty-kate:
Thanks!

PR #19215, opened 2021-08-05
Luv 0.5.10 - binding to libuv, the core of Node.js

aantron:
This release contains one bugfix, which is critical for 32-bit platforms:

> Bugs fixed
>
> - Shim error constant values out of range on 32-bit platforms (aantron/luv#123, reported by Jerry James).

### DIFF
--- a/packages/luv/luv.0.5.10/opam
+++ b/packages/luv/luv.0.5.10/opam
@@ -1,0 +1,41 @@
+opam-version: "2.0"
+
+synopsis: "Binding to libuv: cross-platform asynchronous I/O"
+
+license: "MIT"
+homepage: "https://github.com/aantron/luv"
+doc: "https://aantron.github.io/luv"
+bug-reports: "https://github.com/aantron/luv/issues"
+
+authors: "Anton Bachin <antonbachin@yahoo.com>"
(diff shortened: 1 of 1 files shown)


packages/luv/luv.0.5.10/opam
```diff
@@ -0,0 +1,41 @@
...
+depends: [
+  "base-unix" {build}
+  "ctypes" {>= "0.14.0"}
+  "dune" {>= "2.0.0"}
+  "ocaml" {>= "4.02.0"}
+  "result"
+
+  "alcotest" {with-test & >= "0.8.1"}
+  "base-unix" {with-test}
+  "odoc" {with-doc & = "1.5.2"}
```
kit-ty-kate, line 23:
```suggestion
  "odoc" {with-doc & >= "1.5.2"}
```
Why does this requests only 1.5.2? Does it not work with 1.5.3 for instance?

aantron: Because I want fully reproducible doc builds, as I carefully review the docs. I manually upgrade the odoc version when I do major docs work.

aantron:
Thanks!

PR #19216, opened 2021-08-05
pa_ppx release 0.08

chetmurthy:
OCaml 4.13.0 support


chetmurthy:
Looks like macos-homebrew has problems with some perl packages that we created a while ago.  Should I modify them so they exclude macos-homebrew?  I guess that would fix these Ci failures?  Or just leave it alone?

kit-ty-kate:
> Looks like macos-homebrew has problems with some perl packages that we created a while ago. Should I modify them so they exclude macos-homebrew? I guess that would fix these Ci failures? Or just leave it alone?

I was only able to reproduce it on macOS < 11.0 so maybe it's fine for now. I think we can just ignore the error for now.
Thanks!

PR #19223, opened 2021-08-07
Update depext conf-postgresql.1 to use updated postgresql-devel packages on rhel, centos and fedora

bikallem:
This PR udpates the conf-postgresql.1 depext to use update postgresql-devel packages. The package names are detailed here. https://www.postgresql.org/download/linux/redhat/


packages/conf-postgresql/conf-postgresql.1/opam
```diff
@@ -11,9 +11,24 @@ depexts: [
...
+  ["postgresql10-devel"] {os-distribution = "centos"}
+  ["postgresql11-devel"] {os-distribution = "centos"}
+  ["postgresql12-devel"] {os-distribution = "centos"}
+  ["postgresql13-devel"] {os-distribution = "centos"}
+  ["postgresql-devel"]   {os-distribution = "rhel"}
+  ["postgresql96-devel"] {os-distribution = "rhel"}
+  ["postgresql10-devel"] {os-distribution = "rhel"}
+  ["postgresql11-devel"] {os-distribution = "rhel"}
+  ["postgresql12-devel"] {os-distribution = "rhel"}
+  ["postgresql13-devel"] {os-distribution = "rhel"}
```
kit-ty-kate, line 25: I don't understand how it's supposed to work. How can we install all of these incompatible packages at once?
Is this versions of rhel/centos/... dependent?

bikallem: Basically on rhel/centos/rockylinux 8.4 and above, there is no package called `postgresql-devel`. It seems the best way to install postgres on these OSes is to use the yum repo as listed on https://www.postgresql.org/download/linux/redhat/ ,i.e. the package names are postgresql version specific. What I meant to specify was if one of the listed packages exists then `conf-postgresql.1` depext is satisfied. Is there a correct way to specify such dependencies?

bikallem: Also how do I install this opam depext locally in my switch, i.e. can I do `opam pin conf-postgresql.1` locally?

kit-ty-kate: you can do `opam remote add tmp git://github.com/bikallem/opam-repository.git#conf-postgresql` and then try `opam install yourpkg` to test locally

kit-ty-kate: > Is there a correct way to specify such dependencies?

Sadly if it's not OS version dependent, no. Is there really no virtual package for it? If different versions are required you can do like the `conf-llvm` does and have one package per version but this isn't ideal.

bikallem: > Is there really no virtual package for it? 
Unfortunately, no. IIUC it is so to allow different versions of posgresql to be installed in the same machine. 
> 
If different versions are required you can do like the conf-llvm does and have one package per version but this isn't ideal.

From a quick glance it seems that is the way to go if one wants to avoid vendoring/pinning `postgresql` when using the `postgresql` package in the affected OSes.

I will report progress here but it will probably be a bit slow due to time constraints.

bikallem: @mmottl as you are the maintainer of `postgresql` what do you think regarding creating separate versions of `postgresql` that matches the postgresql version?

mmottl: @bikallem Not sure what a good solution would be.  The `postgresql-ocaml` package is supposed to compile with all versions, possibly by disabling unsupported or undetected features.  I always found the PostgreSQL packaging scheme a little odd and don't quite understand why it should be necessary to install different versions simultaneously.  It shouldn't be terribly difficult for newer versions to support older database formats.

I think this would be best solved by OPAM, because this is apparently not the only package with this kind of problem.  It should be possible to specify multiple ways of satisfying a dependency.  They could be specified in order of preference.  If no dependency is installed, the system could try installing them in order until it finds one that works.  I'd rather not confuse people with multiple OPAM packages, each of which would need to be updated each time I make a new release.

bikallem: @mmottl I am not sure if opam can solve this issue yet, perhaps @kit-ty-kate knows of a solution without requiring `postgresql` package changes. 
However, for now since `conf-postgresql` dependency seems more of a hindrance than help, 
perhaps `postgresql` package can remove the dependency on `conf-postgresql`. At the moment, `postgresql` package is unusuable with opam 2.1 since it checks for package dependency every time we issue `opam update`, `opam install` and the command seems to fail or we have to give it `--no-depext` every time we use opam.

kit-ty-kate: For now, you can bypass the check by calling: `opam option --global depext-bypass=postgresql-devel`

kit-ty-kate: Actually `--no-depext` is supposed to record the exception/depext-bypass as well. This looks like a bug in opam if not. Could you make an issue in https://github.com/ocaml/opam/issues ?

mmottl: @bikallem This is not just an rhel problem, some other platforms (freebsd, openbsd, macports) just keep requiring a specific older version.  This is not a big problem, a client that was built for an older protocol should typically still be able to communicate with newer servers.  That said, maybe we can just require a version on rhel and on other platforms that ideally has all features currently supported by the bindings?

It seems that requiring at least version 12 of PostgreSQL should be fine, the latest stable one is already 13.  I'm not sure about available package versions on your platforms, could you please update `conf-postgresql` accordingly?

kit-ty-kate:
Has https://github.com/ocaml/opam-repository/pull/19254 fix this issue?

bikallem:
I haven't tested it but it should. Closing this.

PR #19226, opened 2021-08-09
Add GT v0.4.2

Kakadu:
Ready for review. The only CI left is for homebrew, I think it should be OK.


Kakadu:
Ready for review. The only CI left is for homebrew, I think it should be OK.

kit-ty-kate:
Thanks

PR #19229, opened 2021-08-09
Record the absolute path of llvm-config when found in PATH

jberdine:
When conf-llvm finds the llvm-config executable in PATH, currently the config var only records the basename of the executable. If the PATH later changes, dependent packages can fail to build, or be built with the wrong `llvm-config`. This patch adds the absolute path the the found llvm-config executable to the config var to avoid this instability.

Cc: @kit-ty-kate

### DIFF
--- a/packages/conf-llvm/conf-llvm.10.0.0/files/configure.sh
+++ b/packages/conf-llvm/conf-llvm.10.0.0/files/configure.sh
@@ -11,7 +11,7 @@ for llvm_config in llvm-config-$version llvm-config-${version}.0 llvm-config${ve
     llvm_version="`$llvm_config --version`" || true
     case $llvm_version in
     $version*)
-        echo "config: \"$llvm_config\"" >> conf-llvm.config
+        echo "config: \"$(command -v '$llvm_config')\"" >> conf-llvm.config
         echo "version: \"$llvm_version\"" >> conf-llvm.config
         exit 0;;
     *)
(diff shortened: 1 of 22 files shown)


jberdine:
Thanks for the update @kit-ty-kate! This LGTM, though I haven't tested the old versions.

jberdine:
@mseri if you wanted to test to see if this resolves the failure you observed, it would be very helpful.

avsm:
@jberdine might you have a view on how to progress this?  The opam-repo maintainers are all a bit overloaded with PRs at the moment...

jberdine:
In the projects I am involved in that use llvm, a vendored version of conf-llvm using this change is used with no trouble. As far as I understand, @mseri saw some issues with just the first commit [ea11b7e](https://github.com/ocaml/opam-repository/commit/ea11b7e1b58486355749c42586533ea694d99df6) of this PR after it was force-pushed, but I am unaware of any issue with the PR including both commits.

For how to proceed, I would merge unless someone volunteers more testing.

mseri:
Shouldn’t the quotes be removed from all the configuree.sh files? I would imagine the same failure may happen with the others on certain systems (cannot check the logs anynlonger unfortunately)

jberdine:
Oh yes, you are right. I forgot that I only removed them in the latest version to get feedback from @kit-ty-kate as I was changing her code. This makes me realize that a few versions have been released in the meantime that also need updating.

shonfeder:
Hello! I hope this isn't an unwelcome ping, but I wonder if we can either mover this PR forward or close it. WDYT?

mseri:
I’d be fine for merging it if it is fixed and the new patches are moved to the opam-source-archive in agreement with the new policy

shonfeder:
I am going to close this issue, provisionally. Given the accrued merge conflicts, and the fact that it has been stalled for 2 years, it seems pretty clear no one has the motivation or bandwidth to land this as is.

If I am in error, please re-open, open will an alternative fix, or open an issue to track the unmet need here.

Thanks!

PR #19231, opened 2021-08-09
Tezos 9.7

pirbo:
This a minor yet critical update of the octez distribution (packages are not renamed yet and still called `tezos` :-) ) that allows to run the tezos blockchain with its latest "economic protocol" smoothly.

### DIFF
--- a/packages/tezos-accuser-008-PtEdo2Zk/tezos-accuser-008-PtEdo2Zk.9.7/opam
+++ b/packages/tezos-accuser-008-PtEdo2Zk/tezos-accuser-008-PtEdo2Zk.9.7/opam
@@ -1,0 +1,27 @@
+opam-version: "2.0"
+maintainer: "contact@tezos.com"
+authors: [ "Tezos devteam" ]
+homepage: "https://www.tezos.com/"
+bug-reports: "https://gitlab.com/tezos/tezos/issues"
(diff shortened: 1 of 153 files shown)


avsm:
```
2021-08-09 21:50.27: Warning in tezos-client-001-PtCJ7pwo.9.7: The package seems to use dune but the dune-project file is missing.
```

This seems like a new lint error. It can probably wait until a subsequent PR though, if merging this is urgent.

pirbo:
The problem is that I don't understand the message... 
This opam package has the same structure as all the 152 other packages and there is a dune-project file in the archive for it at https://gitlab.com/tezos/tezos/-/blob/v9.7/src/proto_001_PtCJ7pwo/lib_client/dune-project, so why complaining for this one in particular (or it is the first tested and only the first error is reported...)
As https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/22fc8e409470afd3094ee97c201f97443529f310/variant/(lint) only talks about `git` commands, I'm not helped to reproduce locally in order to try to understand better
In any case, yes, if it is OK for you I would be in favor of sorting that out after the merge of the PR :-)

kit-ty-kate:
This was due to a race-condition in opam-repo-ci. This is now fixed in https://github.com/ocurrent/opam-repo-ci/pull/120. Sorry about that.

Thanks!

PR #19249, opened 2021-08-12
Make the license names match the SPDX standard

kit-ty-kate:
opam 2.2.0 will bring a full SPDX format support (see https://github.com/ocaml/opam/pull/4768).
This PR normalizes the license names to match the standard:
* The `WITH` keyword has to be uppercase according to the standard. opam < 2.2.0 was too permissive.
* Several license ids (e.g. `LGPL-2.1` instead of `LGPL-2.1-only` or `LGPL-2.1-or-later`) are deprecated since SPDX 3.0
* This PR also fixes various license name mistakes from older packages (I've only done the obvious ones, there are still 392 packages to fix manually)


avsm (approved):
I've not verified each license obviously, but eyeballed the diff and checked that it only touches the `license` field and the changes themselves are normalisations. Thanks Kate!

avsm:
Merging before the conflicts kick in :)

PR #19250, opened 2021-08-13
iso639: Disable tests (non-deterministic due to uninitialized bytes)

kit-ty-kate:
```
#=== ERROR while compiling iso639.0.0.4 =======================================#
# context              2.1.0 | linux/x86_64 | ocaml-base-compiler.4.12.0 | file:///home/opam/opam-repository
# path                 ~/.opam/4.12.0/.opam-switch/build/iso639.0.0.4
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune runtest -p iso639 -j 31
# exit-code            1
# env-file             ~/.opam/log/iso639-1796-5b354b.env
# output-file          ~/.opam/log/iso639-1796-5b354b.out
### output ###
#  test_iso639 alias tests/runtest (exit 2)
# (cd _build/default/tests && ./test_iso639.exe)
# Fatal error: exception Assert_failure("tests/test_iso639.ml", 56, 10)
```
cc @paurkedal

### DIFF
--- a/packages/iso639/iso639.0.0.4/opam
+++ b/packages/iso639/iso639.0.0.4/opam
@@ -11,7 +11,7 @@ depends: [
 ]
 build: [
   ["dune" "build" "-p" name "-j" jobs]
-  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+#  ["dune" "runtest" "-p" name "-j" jobs] {with-test} # Non-deterministic (uninitialized bytes)
 ]
 dev-repo: "git+https://github.com/paurkedal/ocaml-iso639.git"
 synopsis: "Language Codes for OCaml"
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
It doesn't seem to be a parallelism problem. @paurkedal would you know what going on?

paurkedal:
The problem is an uninitialized bytes in the code which generates a bitmap to determine whether a three-letter combination is an ISO-639-3 code. I'll release a new version with the fix and updated tables.

paurkedal:
Unfortunately, the bug is in the library itself, so disabling the test will only hide it. I sent a PR for a new release.

kit-ty-kate:
ah i see. Well that's ok we can hide it for this version so we don't have to wonder why it's failing in our revdeps test.
Thanks for looking into it

PR #19237, opened 2021-08-11
Problem with tezos-lmbd.7.4 checksum

pirbo:
In between yesterday where we tested and merged tezos.9.7 and today, the sha*sum of the tarball generated by https://gitlab.com/tezos/tezos/-/archive/v7.4/tezos-v7.4.tar.bz2 has changed! In don't really understand this question of stability of the tarball generated by git* but I suspect that our mistake was to not declare on gitlab.com the tag v7.4 as a "release" in the gitlab interface... (I'm reluctant to do it now as it would be displyed on gitlab as a new release which it is not.) 

This MR deals with the emergency: allowing people to install the latest version of tezos. To enable that I packaged tezos-lmbd.7.5 (`v7.5` being a tag declared as a "release" on gitlab so hopefully the tarball will be more stable (tezos.7.5 were not packaged in opam because there are not changes in ocaml code in between 7.4 and 7.5...)) and I updated the reverse dependency to use this new version.

After that will come an other question: what do we do with all the now uninstallable all version of tezos packages because of `Bad checksum`? Do we remove them? Do we update the checksum but it looks like a never ending jobs? ....

PR #19252, opened 2021-08-13
Update tezos tarball checksum

pirbo:
As explained in #19237, we updated (for good reasons :-)) the `.git-attributes` file in our master branch of https://gitlab.com/tezos/tezos and (for reason I don't see the rational of :-), ) it updated all the tarball generated for all the old versions of tezos!

We'll be of course extremely careful in the future to not touch our git attributes files in ways that may impact the old tarballs (or we'll move out of relying on gitlab git archive to obtain stable tarballs :-p ) but for this time, unless you desagree, we think that the """cheapest""" hotfix is to update the checksums in the opam-repository...

I have an favor to ask, can this PR be exonerated to pass the CI? ;-) I'm pretty sure some versions constraint in these old opam files are outdated (especially as the cool feature "pick oldest possible version of deps" didn't existed at that time) and I'm asked to not spend my work time on updating them :-|


kit-ty-kate:
Thanks! dune-release does something to help this issue: uploading a fixed archive to github linked to the release and use this as archive.
You won't be able to use dune-release because it's still github centric but maybe it might be worth doing something like that for next releases?

PR #19254, opened 2021-08-14
Correct conf-postgresql for EL 8.

paurkedal:
On CentOS 8, the PostgreSQL client library package is called libpq-devel. I presume the same is the case for RHEL and for the distribution with the identifier "os", which seems to be in the family judging from other conf-* packages.


paurkedal:
The second commit makes the same fix for fedora. According to https://docs.fedoraproject.org/en-US/quick-docs/fedora-and-red-hat-enterprise-linux/ I would expect the cut to have been before Fedora 28, but inspecting Docker containers, I found the library changed in Fedora 30.

kit-ty-kate:
Thanks a lot!

PR #19255, opened 2021-08-14
Libtorch1.9

LaurentMazare:
Update the libtorch binary package to the 1.9 version.


kit-ty-kate:
The license field is missing from these two new packages. Could you add it?

LaurentMazare:
Sure, I've added it to both packages.

kit-ty-kate:
Thanks a lot!

PR #19258, opened 2021-08-15
python3: update to fix #18540

### LINKED ISSUE #18540: Archlinux depext for conf-python-3
The depexts for `conf-python-3` on archlinux seems incorrect. From what I can tell, the `python` package now provides `python3` (and there is an explicit `python2` package, see https://wiki.archlinux.org/index.php/python ), but is simply named `python` and the `3` is in the version of the package.

In fact trying to `pacman -S python3` will install the `python` package (version `3.9.3-1` currently it seems), but crucially, I presumes that the `python3` package will not appear in the lists of installed packages that opam depexts looks into, but instead will be listed as simply the `python` package.

As can be seen below, although python3 is tecnically installed on my machine, opam complains about it not being installed.
```
$ yay -Qi python3
Name            : python
Version         : 3.9.3-1
Description     : Next generation of the python high-level scripting language
Architecture    : x86_64
URL             : https://www.python.org/
Licenses        : custom
Groups          : None
Provides        : python3
Depends On      : bzip2  expat  gdbm  libffi  libnsl  libxcrypt  openssl  zlib
Optional Deps   : python-setuptools [installed]
                  python-pip
                  sqlite [installed]
                  mpdecimal: for decimal
                  xz: for lzma [installed]
                  tk: for tkinter [installed]
Required By     : afl-utils  asciidoc  bpytop  gdb  gdb-common  glslang  gstreamer  iotop  perf  python-appdirs  python-cairo  python-distro  python-evdev  python-gobject  python-greenlet  python-idna
                  python-lxml  python-msgpack  python-ordered-set  python-pexpect  python-pillow  python-psutil  python-ptyprocess  python-pynvim  python-pyparsing  python-pyxdg  python-six  python-urllib3
                  python-yaml  reflector  steam  subdl  youtube-dl
Optional For    : botan  git  glib2  graphviz  gtk2  lensfun  libevent  libproxy  lilv  obs-studio  openexr  tdb  texlive-core  texlive-music  usbutils  util-linux  vim  vim-runtime
Conflicts With  : None
Replaces        : python3
Installed Size  : 79.29 MiB
Packager        : Felix Yan <felixonmars@archlinux.org>
Build Date      : Fri 09 Apr 2021 01:35:02 AM CEST
Install Date    : Wed 14 Apr 2021 07:27:35 AM CEST
Install Reason  : Installed as a dependency for another package
Install Script  : No
Validated By    : Signature

guigui@space-guigui ~ $ opam install conf-python-3
[ERROR] Package conf-python-3 depends on the unavailable system package 'python3'. You can use `--no-depexts' to attempt installation anyway.
```

I guess the solution would simply be replacing `python3` by `python` in the depext for archlinux ? (is it possible to try and add a lower bound on version of external packages ?)

mseri:
Should resolve #18540


kit-ty-kate:
I'm not really convinced this change requires it to be a new package version. Is there any other difference than `python3 --> python` for archlinux?

mseri:
I don’t think so. I made a new package to make sure that there is the old version to pin for older versions of arch. I don’t have a strong opinion on having a new package or not

kit-ty-kate:
To me it doesn't make sense since archlinux does not support old versions of their repositories. It is quite renoune that a few months old out-of-date archlinux install is a dead install, and the switch to python3 by default was made over 10 years ago (2011) ^^"

mseri:
Perfect. I merge it back to the previous version then

kit-ty-kate:
Thanks!

PR #19269, opened 2021-08-16
core_kernel v0.14.2 and ppx_optcomp v0.14.3

cwong-ocaml:
These add support for some flambda stuff.


mshinwell:
In particular this ensures that `Core_kernel` uses the new primitives `%field{0,1}_immut` for `fst` and `snd` that are re-exported from the stdlib directly as primitives.  This is required for compilation with compilers built using the Flambda backend build system (https://github.com/ocaml-flambda/flambda-backend), which includes Flambda 2.

kit-ty-kate:
Thanks! Could you return the fixes to the opam file upstream in the v0.14 branch?

PR #19277, opened 2021-08-18
[new release] 0.0.3 release of GopCaml-mode

Gopiandcode:
- Version 0.0.3 - Polishing release
  - *Added support for customisable verbosity*
    - Customise the Emacs variable `gopcaml-messaging-level` to change
      the level of messages that are output by GopCaml. Set it to
      `'none` to disable messages entirely.
  - *Fixed bug when starting zipper mode at the start of a file.*
    - Zipper mode selects the immediately prior byte position to avoid
      inconsistencies when the cursor is just on the edge of an
      expression, but when the cursor is at position 1, this causes an
      error as 0 is not a valid point.
  - *Special casing of shebangs*
    - Added support for handling shebangs at the start of a buffer.
    - Implemented as part of a larger library for preprocessing buffer
      text before running the parser on it - could be extended to
      support additional preprocessing in the future.
    - Another possible direction for extension is to use an Emacs
      callback to modify the text, although this may not be ideal, as
      the parsing has to be as fast as possible.


Gopiandcode:
Not sure why the lower bound tests keep failing - it looks like when uninstalling the package, ocp-build fails to build? Is that a problem on my side?

mseri:
No, sorry for the delay. I will look into that issue separately, looks more some error in the dependencies of ocp-build or ocp-indent

PR #19278, opened 2021-08-18
[conf-zmq] Conforms to macOS M1 with Homebrew

correnson:
Disclaimer : I didn't find any special issue tracker for conf-zmq.

On macOS+M1, Homebrew installs to `/opt/homebrew`.
This fix is a _proposal_ to keep previous behaviors except for this target.
A fix for zmq package is probably also needed.


kit-ty-kate:
Subsequent change tested successfully on macOS/arm64 and FreeBSD. Using pkg-config when available is more much reliable in the long run than having magic strings. Thanks for bringing up the issue

PR #19280, opened 2021-08-18
[new release] unix-errno (0.6.0)

psafont:
Unix errno types, maps, and support

- Project page: <a href="https://github.com/xapi-project/ocaml-unix-errno">https://github.com/xapi-project/ocaml-unix-errno</a>

##### CHANGES:

* Port to dune

### DIFF
--- a/packages/dlm/dlm.0.3.0/opam
+++ b/packages/dlm/dlm.0.3.0/opam
@@ -13,11 +13,12 @@ build: [
 depends: [
   "ocaml"
   "jbuilder" {>= "1.0+beta10"}
-  "ctypes" {>= "0.11.1"}
-  "lwt"
-  "unix-errno"
-  "ounit"
-  "uuidm"
+  "conf-pkg-config"
+  "ctypes" {>= "0.11.1" & < "0.18.0"}
+  "lwt" {>= "4.2.0"}
+  "unix-errno" {< "0.6.0"}
+  "ounit" {>= "2.0.0"}
+  "uuidm" {>= "0.9.6"}
 ]
 depexts: [
   ["libdlm-dev"] {os-family = "debian"}
@@ -25,10 +26,9 @@ depexts: [
   ["dlm-devel"] {os-distribution = "centos"}
   ["dlm-devel"] {os-distribution = "fedora"}
   ["dlm-devel"] {os-distribution = "ol"}
-  ["dlm-git"] {os-distribution = "arch"}
   ["libdlm-devel"] {os-family = "suse"}
 ]
-available: [ os = "linux" & os != "alpine" ]
+available: [ os = "linux" & os != "alpine" & os != "archlinux"]
 synopsis: "Libdlm bindings"
 description: """
 [![Build Status](https://travis-ci.org/xapi-project/ocaml-dlm.svg?branch=master)](https://travis-ci.org/xapi-project/ocaml-dlm)
--- a/packages/dlm/dlm.0.3.0/opam
+++ b/packages/dlm/dlm.0.3.0/opam
@@ -28,7 +28,7 @@ depexts: [
   ["dlm-devel"] {os-distribution = "ol"}
   ["libdlm-devel"] {os-family = "suse"}
 ]
-available: [ os = "linux" & os != "alpine" & os != "archlinux"]
+available: [ os = "linux" ]
 synopsis: "Libdlm bindings"
 description: """
 [![Build Status](https://travis-ci.org/xapi-project/ocaml-dlm.svg?branch=master)](https://travis-ci.org/xapi-project/ocaml-dlm)
--- a/packages/dlm/dlm.0.3.1/opam
+++ b/packages/dlm/dlm.0.3.1/opam
@@ -13,10 +13,11 @@ run-test: ["dune" "runtest" "-p" name "-j" jobs]
 depends: [
   "ocaml"
   "dune" {>= "1.4"}
-  "ctypes" {>= "0.11.1"}
-  "lwt"
-  "unix-errno"
-  "ounit"
+  "conf-pkg-config"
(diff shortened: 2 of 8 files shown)


psafont:
Failing revdeps seems to be produced because ocamlfind is unable to find a private library to build, I assume because the build expects it to be present for the public one (unix-errno.unix):
```
- cc -I /home/opam/.opam/4.12/lib/ctypes -I /home/opam/.opam/4.12/lib/ocaml -o lib_gen/unix_time_types_detect lib_gen/unix_time_types_detect.c
- lib_gen/unix_time_types_detect > unix/unix_time_types_detected.ml
- ocamlfind ocamldep -package ctypes.stubs -package ctypes -package unix-errno.unix -modules unix/unix_time_types_detected.ml > unix/unix_time_types_detected.ml.depends
- + ocamlfind ocamldep -package ctypes.stubs -package ctypes -package unix-errno.unix -modules unix/unix_time_types_detected.ml > unix/unix_time_types_detected.ml.depends
- ocamlfind: Package `errno_bindings' not found - required by `unix-errno.unix'
- Command exited with code 2.
- make: *** [Makefile:47: build] Error 10
```

Is this something I can solve by changing how dune is used? Is this a problem with the files that Dune produces?

I'll add an upper bound to the revdeps but it'd be nice to know why they fail.

psafont:
There are two types of failures:
- The one in arm32 about negative values not fitting 2-bytes values (I guess the error codes)
- Revdeps failing because their build systems don't like what dune is doing to keep some libraries "private": not installing the .cmi files. This happend both with jbuilder and ocamlbuild:
```
- lib_gen/unix_sys_resource_types_detect > unix/unix_sys_resource_types_detected.ml
- ocamlfind ocamldep -package ctypes.stubs -package unix-errno.unix -modules unix/unix_sys_resource_types_detected.ml > unix/unix_sys_resource_types_detected.ml.depends
- + ocamlfind ocamldep -package ctypes.stubs -package unix-errno.unix -modules unix/unix_sys_resource_types_detected.ml > unix/unix_sys_resource_types_detected.ml.depends
- ocamlfind: Package `errno_bindings' not found - required by `unix-errno.unix'
- Command exited with code 2.
- make: *** [Makefile:47: build] Error 10

```
```
# File "/home/opam/.opam/4.12/lib/unix-errno/META", line 1, characters 0-0:
# Error: Library "errno_bindings" not found.
# -> required by library "unix-errno.unix" in /home/opam/.opam/4.12/lib/unix-errno/unix
```
Is this something that should be reported to dune developers?

Aside from this dlm packages have problems finding pkg-config

psafont:
Most of the DLM issues are about depexts and distributions supporting dlmlib. I don't think it should block the release.

The revdeps issues stand

mseri:
I think `unix-sys-stat` requires an upper bound on `unix-errno`, or you mean that it has never worked?

psafont:
> I think unix-sys-stat requires an upper bound on unix-errno, or you mean that it has never worked?

I think it worked previously. The suspicion I have is that the failure should not happen after using the private lib mechanism in dune. I do not know whether this is a bug in the mechanism, somewhere else or I'm wrong and failure is expeccted and should add the upper bound to the revdeps.

mseri:
Ok, then we need to add upper bounds since the code needs to be updated to work with the new version. Thanks for the clarification

psafont:
Let me know if more changes are needed

packages/dlm/dlm.0.3.1/opam
```diff
@@ -26,10 +27,9 @@ depexts: [
   ["dlm-devel"] {os-distribution = "centos"}
   ["dlm-devel"] {os-distribution = "fedora"}
   ["dlm-devel"] {os-distribution = "oraclelinux"}
-  ["dlm-git"] {os-distribution = "archlinux"}
   ["libdlm-devel"] {os-distribution = "opensuse"}
 ]
-available: [ os = "linux" ]
+available: [ os = "linux" & os != "alpine" & os != "archlinux"]
```
mseri, line 32:
```suggestion
available: [ os = "linux"]
```
Is it possible to install the package by manually compiling the appropriate library?

packages/dlm/dlm.0.3.0/opam
```diff
@@ -13,22 +13,22 @@ build: [
...
   ["libdlm-dev"] {os-family = "debian"}
   ["dlm-devel"] {os-distribution = "rhel"}
   ["dlm-devel"] {os-distribution = "centos"}
   ["dlm-devel"] {os-distribution = "fedora"}
   ["dlm-devel"] {os-distribution = "ol"}
-  ["dlm-git"] {os-distribution = "arch"}
   ["libdlm-devel"] {os-family = "suse"}
 ]
-available: [ os = "linux" & os != "alpine" ]
+available: [ os = "linux" & os != "alpine" & os != "archlinux"]
```
mseri, line 31:
```suggestion
available: [ os = "linux" ]
```
Is it possible to install the package by manually compiling the appropriate library?

mseri:
Thanks a lot. I have just two more questions, than it is good to go

packages/dlm/dlm.0.3.1/opam
psafont: The source code can be found in https://pagure.io/dlm/blob/master/f/libdlm

I guess it's possible to manually compile it, although I haven't tried

packages/dlm/dlm.0.3.0/opam
psafont: See above

mseri:
Let's make it available then, if it screws up something I will take care to revert the commits. No more action needed, I am waiting the CI to make sure I did not mess up and then merge

mseri:
Thanks a lot!

PR #19282, opened 2021-08-18
swhid.0.1

RadioPotin:
Hello again,

@kit-ty-kate this time this should be okay, again, sorry for previous noise


zapashcanon:
@kit-ty-kate I found that the correct minimal menhir version is `menhir.20181113`.

zapashcanon:
@kit-ty-kate, it's failing only on macOS and I don't understand why there's a timeout. It works fine on our CI.

PR #19285, opened 2021-08-19
conf-zmq.0.1: depend on conf-pkg-config

fdopen:
missing in 0c7ecee2b12b1f50cbc7aab572ca1f58c61d0651


kit-ty-kate:
oops I'm really sorry about that. Turns out most package manager make zmq depend on pkg-config so that's why our CI didn't catch that.
Thanks!

PR #19291, opened 2021-08-20
added packages/fasmifra/fasmifra.1.1.0/opam

UnixJunkie:
faster than previous version in case there are _many_ molecular fragments
(i.e. many molecules were fragmented).
A binary cache of indexed fragments is stored on disk.
Significantly accelerates molecular generation in case the same
collection of fragments is used to generate molecules several times.


mseri:
Thanks

PR #19299, opened 2021-08-22
Package prbnmcn-proptest.0.0.1

igarnier:
### `prbnmcn-proptest.0.0.1`
Property-based test helpers for prbnmcn packages

---
* Homepage: http://github.com/igarnier/prbnmcn-proptest
* Source repo: git+https://github.com/igarnier/prbnmcn-proptest
* Bug tracker: http://github.com/igarnier/prbnmcn-proptest

---


igarnier:
It seems the CI previously failed because the `crowbar` package doesn't declare an adequate lower bound on the version of `cmdliner`, I added a dummy constraint in my package.

kit-ty-kate:
> It seems the CI previously failed because the `crowbar` package doesn't declare an adequate lower bound on the version of `cmdliner`, I added a dummy constraint in my package.

Thanks! I've fixed the crowbar package directly so the dummy constraint is no longer needed and I removed it.

Thanks a lot!

PR #19310, opened 2021-08-24
Package ppx_make.0.2.1

bn-d:
### `ppx_make.0.2.1`
Ppxlib based make deriver

---
* Homepage: https://github.com/erebuxy/ppx_make
* Source repo: git+https://github.com/erebuxy/ppx_make.git
* Bug tracker: https://github.com/erebuxy/ppx_make/issues

---


bn-d:
I changed my username, so this would avoid github redirect

kit-ty-kate:
Interestingly enough, the two archives decompress to exactly the same files (no diff) but the archives differ on a few gzip metadata.. This is a first one ^^"

Thanks!

bn-d:
> Interestingly enough, the two archives decompress to exactly the same files (no diff) but the archives differ on a few gzip metadata.. This is a first one ^^"
> 
> Thanks!

I rebased and re signed all the commits recently. So I also need to re tag all the release. That probably is the reason.

PR #19318, opened 2021-08-24
Add gapi-ocaml 0.4.2, google-drive-ocamlfuse 0.7.27

astrada:


kit-ty-kate:
Could you have a look at the failures? There seems to be missing constraints you can add.

astrada:
Do you mean the failures that report: Version 1.6 of dune is not supported?

Actually the error is caused by the ocamlfuse dependency, but in the opam file of ocamlfuse 2.7.1-cvs7 the constraint on dune (>= 1.6) is already specified (https://opam.ocaml.org/packages/ocamlfuse/). Is it expected that packages that use a dependency with a constraint on dune should replicate that constraint?

kit-ty-kate:
ah my bad I missed that it was a dependency. I've added a commit to fix that. Don't worry you don't have to do anything regarding the dune constraint. Thanks for spotting my mistake.

astrada:
Thanks!
I missed the following line in the logs, and I thought it was still using 2.7.1-cvs7 that has the correct constraint specified.
  - downgrade ocamlfuse                2.7.1-cvs7 to 2.7.1-cvs6 [uses dune]

mseri:
Looks good, sorry for the delay

astrada:
no problem!

PR #19325, opened 2021-08-25
fix dev-repo for binbin

zapashcanon:
Hi,

Just so we don't need ssh when running `opam source` on this package.


kit-ty-kate:
Thanks!

PR #19326, opened 2021-08-25
Make patdiff.v0.14.0 unavailable on s390x until janestreet/core#145 gets fixed

kit-ty-kate:
I'm not sure this is the right fix so I'm making it a draft for the moment until we know for sure it's an issue in core and not down the chain (e.g. `jst-config`)


kit-ty-kate:
Closing in favour of #19412

Issue #18343, opened 2021-03-15
Various basic OCaml core packages don't have a license field


MSoegtropIMC:
I maintain a project called "Coq Platform" which is essentially a set of opam packages. I would like to drag the license information for the ReadMe from opam. Almost all packages do have a license field. A notable exception are various OCaml core packages as the following output shows (packages with license field stripped):

```
Coq$ opam list --columns=name,version,license:
# Packages matching: installed
# Name                   # Version  # License
base-bigarray            base
base-threads             base
base-unix                base
depext                   transition
ocaml                    4.07.1
ocaml-base-compiler      4.07.1
ocaml-config             1
ocaml-secondary-compiler 4.08.1-1
ocamlfind                1.8.1
ocamlfind-secondary      1.8.1
seq                      base
```

Would it be possible to fix this, also for older compilers (for Windows compatibility reasons I am currently stuck with 4.07.1)?

In a private communication with @xavierleroy he suggested that `opam lint` should flag missing license fields - provided we agree that opam packages should have a license field.

For complex projects like Coq Platform, a consistent license field would be of great help for the maintainers and users.

kit-ty-kate:
> In a private communication with @xavierleroy he suggested that `opam lint` should flag missing license fields - provided we agree that opam packages should have a license field.

I raised this issue separately upstream in https://github.com/ocaml/opam/issues/4598

avsm:
Thanks for bringing this up @MSoegtropIMC -- I've had this on my todo list for some years now, as establishing the license of a set of opam packages is important for almost any industrial use.  The answer isn't quite as simple as just following the `license` field in the package, so here follows my thoughts:

There are two uses for the licensing metadata:
- a "shallow" check (for example to ensure there is no AGPL code in a product intended to be closed source)
- a  "deep" check to satisfy the needs of documentation and acknowledgements (e.g. the BSD advertising clauses and similar)

For the shallow check, the licensing fields are _not_ checked closely by the opam-repository team when we merge PRs, and also doesn't cover the case where a library is multilicensed due to incorporating some modules under another license (not common, but not zero either).  In this case, you need to drop into the `opam source` to verify each package. 

[...]

PR #19327, opened 2021-08-25
Add missing license field on the OCaml compiler packages

kit-ty-kate:
cc @Octachron 
Partly solves https://github.com/ocaml/opam-repository/issues/18343

### DIFF
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.4.03.0/opam
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.4.03.0/opam
@@ -1,6 +1,7 @@
 opam-version: "2.0"
 synopsis: "Official 4.03.0 release"
 maintainer: "platform@lists.ocaml.org"
+license: "QPL-1.0 AND LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception"
 authors: ["Xavier Leroy" "Damien Doligez" "Alain Frisch" "Jacques Garrigue" "Didier Rémy" "Jérôme Vouillon"]
 homepage: "https://ocaml.org"
 bug-reports: "https://github.com/ocaml/opam-repository/issues"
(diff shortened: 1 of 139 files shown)


packages/ocaml-base-compiler/ocaml-base-compiler.4.03.0/opam
```diff
@@ -1,6 +1,7 @@
 opam-version: "2.0"
 synopsis: "Official 4.03.0 release"
 maintainer: "platform@lists.ocaml.org"
+license: "QPL-1.0 AND LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception"
```
Octachron, line 4: 4.03 was the first release to be distributed under the LGPL with linkink exceptions and not QPL.

Octachron:
Normalizing the compiler opam files is a good idea.

packages/ocaml-base-compiler/ocaml-base-compiler.4.03.0/opam
kit-ty-kate: are you sure? According to https://github.com/ocaml/ocaml/commit/7fc2d214c4ee8853615fc3e5649f45e11d85378d it was only done in 4.04.0. The 4.03 branch has it technically but it was only added after the release of 4.03.0

Octachron: Hm, and according to ocaml/ocaml@0ca4d6a74 the plan was to update the license for 4.03 . Keeping the license is probably more precise indeed. Sorry for the noise.

xavierleroy: I agree with @kit-ty-kate : 4.03.0 was released before the switch to LGPL.

xavierleroy (approved):
Thanks @kit-ty-kate for taking care of this!  I checked the main packages and it looks good to me.  Didn't check all the variant packages, but I trust you.

Side note: the line
```
authors: "Xavier Leroy and many contributors"
```
that appears in some of the `opam` files doesn't give enough justice to the "many contributors", but that's unrelated to this PR.

PR #19331, opened 2021-08-25
[new release] re (1.10.0)

rgrinberg:
RE is a regular expression library for OCaml

- Project page: <a href="https://github.com/ocaml/ocaml-re">https://github.com/ocaml/ocaml-re</a>

##### CHANGES:

* Add the `[:alpha:]` character class in `Re.Perl` (ocaml/ocaml-re#169)
* Double asterisk (`**`) in `Re.Glob` (ocaml/ocaml-re#172)
  Like `*` but also match `/` characters when `pathname` is set.
* Double asterisk should match 0 or more directories unless in trailing
  position. (ocaml/ocaml-re#192, fixes ocaml/ocaml-re#185)

### DIFF
--- a/packages/re/re.1.10.0/opam
+++ b/packages/re/re.1.10.0/opam
@@ -1,0 +1,46 @@
+opam-version: "2.0"
+
+maintainer: "rudi.grinberg@gmail.com"
+authors: [
+  "Jerome Vouillon"
+  "Thomas Gazagnaire"
+  "Anil Madhavapeddy"
+  "Rudi Grinberg"
+  "Gabriel Radanne"
+]
+license: "LGPL-2.0 with OCaml linking exception"
+homepage: "https://github.com/ocaml/ocaml-re"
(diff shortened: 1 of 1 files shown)


mseri:
The failures seem to be independent

kit-ty-kate:
> The failures seem to be independent

These ones don't seem to be:
```
# File "src/tyre.ml", line 332, characters 19-29:
# 332 |     Seq.map aux @@ Re.all_seq ~pos ~len re original
#                          ^^^^^^^^^^
# Error: Unbound value Re.all_seq
# Hint: Did you mean all_gen?
```
I double checked and it worked with the previous version of `re` (1.9.0). However I can't find the commit that removed it or any mention of it in the changelog, which i find very weird. @rgrinberg was this meant to happen?

rgrinberg:
Good catch. This was indeed an accidental regression. I released 1.10.1 with the fixe.

PR #19338, opened 2021-08-26
Add conf-rustup as an alternative package to conf-rust

lthms:
Most of the time, Rust developers are using `rustup` to manage their installation of Rust. We propose to add a new package, `conf-rustup`, as an alternative of `conf-rust`, that would use `rustup` instead of `rust`.

I have never implemented such a package. Feedback is welcome!

### DIFF
--- a/packages/conf-rustup/conf-rustup.0.1/opam
+++ b/packages/conf-rustup/conf-rustup.0.1/opam
@@ -1,0 +1,22 @@
+opam-version: "2.0"
+maintainer: "lthms@nomadic-labs.com"
+authors: "Thomas Letan"
+homepage: "https://github.com/ocaml/opam-repository"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: "MIT"
+build: [
+  ["rustup-init" "--version"] {os-distribution = "alpine"}
+  ["rustup" "--version"] {os-distribution = "arch"}
+  ["rustup" "--version"] {os-distribution = "nixos"}
+  ["rustup" "--version"] {os = "macos" & os-distribution = "homebrew"}
+]
+depexts: [
+  ["rustup"] {os-distribution = "nixos"}
+  ["rustup"] {os-distribution = "arch"}
+  ["rustup"] {os-distribution = "alpine"}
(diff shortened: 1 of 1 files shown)


lthms:
It actually does more than that. It allows you to set-up a rust version to use _per directory_ (*e.g.*, you can include a `rustup-toolchain` file in the root of your project, and you know thanks to rustup that when you use `rust` in your repo, the expected version of rust is used).

The idea here is not to use `rustup` to install `rust`. The idea is to rely on `rustup` to know that the expected version of rust is used.

lthms:
Out of curiosity, @kit-ty-kate, what is the best way to move forward here?

kit-ty-kate:
sorry i have many other things to do (see the number of PRs pilling up) and thus slightly overwhelmed.

Mh, could you give me an example usage for this? Like a repository opam ready that makes use of rustup

lthms:
As of now, Tezos can only be built using a specific Rust version, and rely on `rustup` (through a `rust-toolchain` file) to make said version available.

Admittedly, you _can_ build Tezos without `rustup`, but not without the specific version of `rust`.

mseri:
The depext mechanism is designed to install external system dependencies, at the moment it cannot be used to install extra package-managers in order to install extra compiler toolchains.

In fact, rustup would not be able install anything while running in the opam sandbox (it does not have network access nor disk access outside the local folder) and its presence does not guarantee that the correct version of the toolchain is installed.

I think that what you are doing right now, installing rustup and the correct toolchain version in advance locally or in the CI is the correct thing to do. I would suggest, additionally, to add a post-install failure message in your rust-dependent packages that specifies the minimal rust version that needs to be installed.
 
Currently we are looking at some workarounds that would allow to at least detect early the presence of the right installed version and fail with an informative message otherwise, I hope this will work out and tell you more in few weeks.
There is also a long-term plan to solve this kind of issue more generally by allowing special interfacing to external package/toolchain managers, but this will not see the light any time soon.

I hope my reasoning above clarifies why I am closing this PR. Thanks again for proposing it and stimulating the discussion.

PR #19064, opened 2021-07-15
[new release] textmate-language (0.3.1)

alan-j-hu:
Tokenizing code with TextMate grammars for syntax highlighting

- Project page: <a href="https://github.com/alan-j-hu/ocaml-textmate-language">https://github.com/alan-j-hu/ocaml-textmate-language</a>
- Documentation: <a href="https://alan-j-hu.github.io/ocaml-textmate-language/">https://alan-j-hu.github.io/ocaml-textmate-language/</a>

##### CHANGES:

- Add `find_by_filetype` function.

PR #19354, opened 2021-08-27
Accept failures in conf-oniguruma

alan-j-hu:
This should ignore some failures in which the Oniguruma package isn't available for certain distributions. See the conversation at https://github.com/ocaml/opam-repository/pull/19064.

Note that the CI failure for Oracle Linux 7 is not because of conf-oniguruma. A package exists, but the version is too old.


mseri:
This will not make a difference on the CI for the dependent packages but may be making a difference in the future

PR #19370, opened 2021-08-30
[new release] Initial release of Ego (0.0.5) - EGraphs in OCaml

Gopiandcode:
# Initial Release of Ego

- EGraph resolution works (but search space can blow up)
- Search procedure needs some more tuning + some optimisations can be stolen from Egg.
- Fully documented interface


kit-ty-kate:
Thanks!

PR #19375, opened 2021-08-31
[new release] graphv_webgl_impl, graphv_webgl, graphv_gles2_native_impl, graphv_gles2_native, graphv_gles2, graphv_font_stb_truetype, graphv_font_js, graphv_font, graphv_core_lib, graphv_core and graphv (0.1.1)

wlitwin:
WebGL implementation of the backend renderer for the Graphv library

- Project page: <a href="https://github.com/wlitwin/graphv">https://github.com/wlitwin/graphv</a>
- Documentation: <a href="https://wlitwin.github.io/docs/graphv/graphv">https://wlitwin.github.io/docs/graphv/graphv</a>

##### CHANGES:

* Fix stencil path bug when rendering a fill (9ba038d)
* Remove debug assert

### DIFF
--- a/packages/graphv/graphv.0.1.1/opam
+++ b/packages/graphv/graphv.0.1.1/opam
@@ -1,0 +1,39 @@
+opam-version: "2.0"
+synopsis: "Top_level graphv package, includes all dependencies"
+maintainer: ["walter@litwinczyk.com"]
+authors: ["Walter Litwinczyk"]
+license: "MIT"
+homepage: "https://github.com/wlitwin/graphv"
+doc: "https://wlitwin.github.io/docs/graphv/graphv"
+bug-reports: "https://github.com/wlitwin/graphv/issues"
+depends: [
+  "dune" {>= "2.8"}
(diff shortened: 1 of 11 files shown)


wlitwin:
What is the best way to exclude Mac/OSX? It seems Apple has deprecated OpenGL on their platform. Also, I'm unsure what to do about the Oracle Linux, as it seems it can't find the package it thinks it has.

mseri:
For ease of maintenance of the packages, I would recommend to use `{=version}` to tie together the versions of the packages released in bundle. In this way you will not need to add lots of upper bound if a breaking change happens, and you can simplify the release and tag by re-releasing all packages together.

Of course you are free to keep it as it is, but in the past we saw that this can make a big difference both for us and the maintainers of the package

wlitwin:
No worries, if {= version} makes it easier I'll do that! I have no preference.

wlitwin:
I have changed it to {=version} and it looks like it didn't break anything. (Same Oracle7 issue as before).

mseri:
Thanks a lot

PR #19378, opened 2021-08-31
fix zbar dev-repo

zapashcanon:


kit-ty-kate:
What's the motivation behind this change? `git://` should work the same way as `git+https://` regardless of the `.git` suffix.

zapashcanon:
Hmmmm, I believe I made a mistake, sorry...

PR #19389, opened 2021-09-01
fix dev-repo for vrt

zapashcanon:


zapashcanon:
@afiniate, do you think you could upstream the change ?

kit-ty-kate:
Thanks

PR #19153, opened 2021-07-28
zoggy.github.com > zoggy.github.io

sagotch:
I did not test every url, but every url I tested is broken before this fix.

See [GitHub Pages will stop redirecting Pages sites from *.github.com after April 15, 2021 ](https://github.blog/changelog/2021-01-29-github-pages-will-stop-redirecting-pages-sites-from-github-com-after-april-15-2021/)

cc @zoggy 

i.e. currently, `ocamldiff` fails to install on my CI script

```
[ERROR] The sources of the following couldn't be obtained, aborting:
          - ocamldiff.1.1:
            http://zoggy.github.com/ocamldiff/ocamldiff-1.1.tar.gz (curl: code 404 while downloading http://zoggy.github.com/ocamldiff/ocamldiff-1.1.tar.gz)
```

PR #19392, opened 2021-09-02
zoggy.github.com > zoggy.github.io

mseri:
Same as #19153 with the merge conflicts resolved


mseri:
The xtmpl and ocaml-iri packages have moved to https://framagit.org/zoggy and are no-longer on github. So they are going to be broken independently of this change :-/

mseri:
Being addressed in a different way

PR #19396, opened 2021-09-03
Package ocsigen-toolkit.2.13.0

jrochel:
### `ocsigen-toolkit.2.13.0`
Reusable UI components for Eliom applications (client only, or client-server)
The Ocsigen Toolkit is a set of user interface widgets that facilitate the development of Eliom applications.

---
* Homepage: http://www.ocsigen.org
* Source repo: git+https://github.com/ocsigen/ocsigen-toolkit.git
* Bug tracker: https://github.com/ocsigen/ocsigen-toolkit/issues/

---

### DIFF
--- a/packages/ocsigen-toolkit/ocsigen-toolkit.2.13.0/opam
+++ b/packages/ocsigen-toolkit/ocsigen-toolkit.2.13.0/opam
@@ -1,0 +1,23 @@
+opam-version: "2.0"
+maintainer: "dev@ocsigen.org"
+synopsis: "Reusable UI components for Eliom applications (client only, or client-server)"
+description: "The Ocsigen Toolkit is a set of user interface widgets that facilitate the development of Eliom applications."
+authors: "dev@ocsigen.org"
+homepage: "http://www.ocsigen.org"
+bug-reports: "https://github.com/ocsigen/ocsigen-toolkit/issues/"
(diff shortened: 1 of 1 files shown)


packages/ocsigen-toolkit/ocsigen-toolkit.2.13.0/opam
```diff
@@ -0,0 +1,26 @@
...
+license: "LGPL-2.1-only with OCaml-LGPL-linking-exception"
+build: [ make "-j%{jobs}%" ]
+install: [ make "install" ]
+depends: [
+  "ocaml" {>= "4.08.0"}
+  "eliom" {>= "6.12.1"}
+  "calendar" {>= "2.0.0"}
+]
+depexts: [
+  ["libgdbm-dev"] {os-family = "debian"}
```
kit-ty-kate, line 18: Is that a hack to help opam 2.0's depext plugin?
2.1 works perfectly and it doesn't seem to help opam 2.0 either.

I would suggest to remove this or if you need dbm explicitly (i'm guessing it's not the case) use the `conf-dbm` package.

jrochel: Thanks. It seems that this dependency was either added by accident or to resolve some obscure error with GitHub Actions. Something like that...

kit-ty-kate:
Thanks!

PR #19428, opened 2021-09-08
add missing conf flag for ocaml

zapashcanon:
:)

### DIFF
--- a/packages/ocaml/ocaml.3.07+1/opam
+++ b/packages/ocaml/ocaml.3.07+1/opam
@@ -27,3 +27,4 @@ authors: [
   "Didier Rémy"
   "Jérôme Vouillon"
 ]
+flags: conf
--- a/packages/ocaml/ocaml.3.07+2/opam
+++ b/packages/ocaml/ocaml.3.07+2/opam
@@ -27,3 +27,4 @@ authors: [
   "Didier Rémy"
   "Jérôme Vouillon"
 ]
+flags: conf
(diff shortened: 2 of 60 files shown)


kit-ty-kate:
For this particular package, this is debatable given this is basically a meta-package here to abstract over `ocaml-base-compiler` and `ocaml-variants`. It has no `url` and thus `opam pin ocaml --dev` would give the wrong result.

zapashcanon:
Shouldn't we add a flag like `compiler` or something to the package then ? I'd like to be able to filter out this kind of package easily and without hard-coding them.

kit-ty-kate:
> Shouldn't we add a flag like `compiler` or something to the package then ?

no, `compiler` isn't the right flag for this. `conf` could be ok though. @rjbou @dra27  ?

zapashcanon:
Yes, conf would be OK for me (I already filter them out :-)).

rjbou:
Yes indeed, it fits `conf` requirements.

dra27:
Ideally no dev-repo for this package and _definitely_ not pointing to ocaml/ocaml, please 🙂

dra27:
In case you have one in the pipeline, please don't open a PR for ocaml-base-compiler or ocaml-variants - there's work in progress on the pinning workflow for the compiler which upstream changes as well (discussed back in March, I just haven't got to it yet)

zapashcanon:
@kit-ty-kate I usef `flag: conf` everywhere, OK for you ?

dra27:
This just needs similar updates to ocaml.4.12.2, ocaml.4.13.1, ocaml.4.13.2, ocaml.5.00.0 and I think it's good to go, thanks!

kit-ty-kate:
Thanks!

PR #19430, opened 2021-09-08
add missing dev-repo to wiringpi

zapashcanon:
@Leonidas-from-XIV, could you upstream the fix ?


Leonidas-from-XIV:
@zapashcanon Thanks. The project is very old and didn't have an OPAM file so I added this. Good start to get it modernized.

kit-ty-kate:
Thanks

PR #19431, opened 2021-09-08
add dev-repo for wget

zapashcanon:
@samoht, could you upstream the fix ?


kit-ty-kate:
Thanks!

samoht:
@zapashcanon there is no opam file upstream, I'll add one with your fix. Thanks!

PR #19437, opened 2021-09-08
Hide dune.2.9.0 as much as possible (generates broken dune files)

kit-ty-kate:
We can't remove it (as it would break lock files) but we can "hide" it for opam >= 2.1 users. It will still be available but will only be taken on the very last resort.
Now that 2.9.1 is release this won't change the users' installation too much.

cc @ejgallego


ejgallego:
Thanks a lot!

Et7f3:
branch can be deleted ?

PR #19438, opened 2021-09-08
add dev-repo to rtime

zapashcanon:
@dbuenzli if you ever care to upstream the change (I saw the project is deprecated)

also, I noticed on cmdliner that you use `http` instead of `https` which should work, it's probably the case for other of your packages


kit-ty-kate:
Thanks

PR #19441, opened 2021-09-08
add dev-repo for pa_qualified and update other fields

zapashcanon:
@murmour could you upstream the fix ?


zapashcanon:
@kit-ty-kate what's wrong with this one ?

zapashcanon:
I believe @cakeplus changed its name to @murmour and the change is correct. GitHub still redirects [cakeplus/pa_qualified](https://github.com/cakeplus/pa_qualified) to `murmour/pa_qualified` but that may not last forever. :)

mseri:
Thanks

PR #19449, opened 2021-09-08
add missing dev-repo to ocaml-arg

zapashcanon:
@samoht could you upstream the fix ?


kit-ty-kate:
Thanks

samoht:
I've added an opam file to the dev repo. These are ooooold packages...

PR #19450, opened 2021-09-08
+flow_parser.0.159.0

alsoijw:
New package version

### DIFF
--- a/packages/flow_parser/flow_parser.0.159.0/opam
+++ b/packages/flow_parser/flow_parser.0.159.0/opam
@@ -1,6 +1,4 @@
 opam-version: "2.0"
-name: "flow_parser"
-version: "0.159.0"
 maintainer: "flow@fb.com"
 authors: ["Flow Team <flow@fb.com>"]
 homepage: "https://github.com/facebook/flow/tree/master/src/parser"
--- a/packages/flow_parser/flow_parser.0.159.0/opam
+++ b/packages/flow_parser/flow_parser.0.159.0/opam
@@ -1,0 +1,34 @@
+opam-version: "2.0"
+name: "flow_parser"
+version: "0.159.0"
+maintainer: "flow@fb.com"
+authors: ["Flow Team <flow@fb.com>"]
+homepage: "https://github.com/facebook/flow/tree/master/src/parser"
+bug-reports: "https://github.com/facebook/flow/issues"
+license: "MIT"
+
+build: [ make "-C" "tmp/flow-0.159.0/src/parser" "build-parser" ]
+install: [ make "-C" "tmp/flow-0.159.0/src/parser" "ocamlfind-install"]
+depends: [
+  "base" {>= "v0.14.1"}
+  "core_kernel" {>= "v0.14.1"}
+  "dtoa" {>= "0.3.2"}
+  "ocaml" {>= "4.10.2"}
+  "ocamlbuild" {build}
+  "ocamlfind" {build}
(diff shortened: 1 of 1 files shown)


packages/flow_parser/flow_parser.0.159.0/opam
```diff
@@ -0,0 +1,34 @@
+opam-version: "2.0"
+name: "flow_parser"
```
mseri, line 2:
```suggestion
```

packages/flow_parser/flow_parser.0.159.0/opam
```diff
@@ -0,0 +1,34 @@
...
+version: "0.159.0"
```
mseri, line 3:
```suggestion
```

mseri:
I think you cannot access the `/tmp/` folder from the sandbox. Is this something that changed with opam 2.1 @kit-ty-kate ? Or I recall it wrong?

packages/flow_parser/flow_parser.0.159.0/opam
```diff
@@ -0,0 +1,34 @@
...
+maintainer: "flow@fb.com"
+authors: ["Flow Team <flow@fb.com>"]
+homepage: "https://github.com/facebook/flow/tree/master/src/parser"
+bug-reports: "https://github.com/facebook/flow/issues"
+license: "MIT"
+
+build: [ make "-C" "tmp/flow-0.159.0/src/parser" "build-parser" ]
```
mseri, line 8:
```suggestion
build: [ make "-C" "src/parser" "build-parser" ]
```

packages/flow_parser/flow_parser.0.159.0/opam
```diff
@@ -0,0 +1,34 @@
...
+install: [ make "-C" "tmp/flow-0.159.0/src/parser" "ocamlfind-install"]
```
mseri, line 9:
```suggestion
install: [ make "-C" "src/parser" "ocamlfind-install"]
```

mseri:
Forget what I said, I think I was completely off-track

mseri:
@kit-ty-kate is this a new warning of ocaml 4.13? Do we need an upper bound on ocaml here?

kit-ty-kate:
> @kit-ty-kate is this a new warning of ocaml 4.13? Do we need an upper bound on ocaml here?

Yes. It's supposed to be a warning by default but this package uses `-warn-error` which is not recommended in releases.

So either the author wants to fix the package or we have to add the `"ocaml" {< "4.13"}` constraint. @alsoijw what do you think?

alsoijw:
@kit-ty-kate, I think it's better to limit the version now.

packages/flow_parser/flow_parser.0.159.0/opam
```diff
@@ -0,0 +1,32 @@
...
+bug-reports: "https://github.com/facebook/flow/issues"
+license: "MIT"
+
+build: [ make "-C" "src/parser" "build-parser" ]
+install: [ make "-C" "src/parser" "ocamlfind-install"]
+depends: [
+  "base" {>= "v0.14.1"}
+  "core_kernel" {>= "v0.14.1"}
+  "dtoa" {>= "0.3.2"}
+  "ocaml" {>= "4.10.2"}
```
mseri, line 14:
```suggestion
  "ocaml" {>= "4.10.2" & < "4.13~"}
```

mseri:
Perfect, thanks a lot

PR #19453, opened 2021-09-09
add dev-repo to mpg123

zapashcanon:
@mbacarella could you upstream the fix ?


mbacarella:
should be done. will cut a new release while i'm at it.

kit-ty-kate:
Thanks

PR #19463, opened 2021-09-09
add missing dev-repo to frag

zapashcanon:
@pw374 could you upstream the fix ?


pw374:
> @pw374 could you upstream the fix ?

Hi, thank you for your interest in `frag`. I had totally forgotten about `frag`.
What do you mean by "upstream the fix"? Do you mean adding an opam file to the repo?

zapashcanon:
Yes. I made a lot of PR to add dev-repo to a lot of repository so I didn't check all of them. In case where there's an .opam file, it would be good to add the missing field so it's not forgotten again.

But in case there's no opam file, yes, you could just add an opam file. :)

kit-ty-kate:
Thanks!

PR #19465, opened 2021-09-09
add missing dev-repo to erssical

zapashcanon:
@zoggy could you upstream the fix ? :)

### DIFF
--- a/packages/erssical/erssical.0.1/opam
+++ b/packages/erssical/erssical.0.1/opam
@@ -1,9 +1,10 @@
 opam-version: "2.0"
 maintainer: "zoggy@bat8.org"
 authors: ["Maxence Guesdon"]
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Thanks

zoggy:
In fact erssical has moved to Framagit and opam file should be ok: https://framagit.org/zoggy/erssical/-/blob/master/erssical.opam

Do you mind if I destroy its github repository ?

zapashcanon:
Please don't destroy it now. If you want to do so, please update all opam files for each version of the package to use the new framagit-hosted version. I didn't do it because the checksum are not the same and I don't have the time to check that the content is exactly the same... But if you want to check yourself, please do it and update the .opam files here before deleting the github repo.

PR #19470, opened 2021-09-09
Some depexts update

rjbou:
Found some missing or inconsistent depext:
* `qrencode.0.2`: add `conf-pkg-config`
* `mccs.1.1+13`: add `conf-g++`


UnixJunkie:
@hannesm Yes, at least on Ubuntu, g++ created a c++ soft link that works.

UnixJunkie:
Apparently, on FreeBSD we need to know which version is supposed to be installed:
```bash
pkg install lang/gcc49
```
I cannot test though.
Maybe a FreeBSD user can update the conf-g++ package to add the correct missing line.
Maybe: 'pkg install lang/gcc' ?

PR #19488, opened 2021-09-11
[new release] opam-grep (0.2.0)

kit-ty-kate:
An opam plugin that greps anything in the sources of every opam packages

- Project page: <a href="https://github.com/kit-ty-kate/opam-grep">https://github.com/kit-ty-kate/opam-grep</a>

##### CHANGES:

- Complete rewrite from POSIX shell script to OCaml, making it more portable
- Use the faster `ripgrep` and `ugrep` over `grep` when available (suggestion from @Engil)
- Ignore read failures from grep (symlinks to unreachable files can make it fail)
- Use `$XDG_CACHE_HOME` when available to store the archives instead of just `~/.cache`
- Use the `progress` library to show progress instead of a non-portable/DIY spinner
- Use the `cmdliner` library to handle arguments and produce a manpage via --help

### DIFF
--- a/packages/opam-grep/opam-grep.0.2.0/opam
+++ b/packages/opam-grep/opam-grep.0.2.0/opam
@@ -1,0 +1,28 @@
+opam-version: "2.0"
+synopsis: "An opam plugin that greps anything in the sources of every opam packages"
+license: "MIT"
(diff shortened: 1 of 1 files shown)


mseri:
Looks good, thanks

PR #19494, opened 2021-09-13
[new release] opam-monorepo (0.2.4)

NathanReb:

### Added

- When querying the solver for the local packages, if no explicit version was
  provided, use the value of the version field in the opam files instead of the
  default `zdev` if it is defined (ocamllabs/opam-monorepo#183, @emillon)

- Add a `-l`/`--lockfile` command line option to explicitly set the lockfile
  to use or generate in `pull` or `lock` (ocamllabs/opam-monorepo#163, @NathanReb)

- Honor `pin-depends` field in opam files. When present, these will be used by
  the solver (ocamllabs/opam-monorepo#153, ocamllabs/opam-monorepo#159, ocamllabs/opam-monorepo#167, @rizo, @TheLortex, @NathanReb).

### Fixed

- Improve `lock` performance (about 2x faster) by loading the repository state
  only once (ocamllabs/opam-monorepo#188, ocamllabs/opam-monorepo#192, @emillon)

- Fix a bug where the dune-project parsing in the `pull` command would fail
  if it used CRLF for new lines. (ocamllabs/opam-monorepo#191, @NathanReb)

- Simply warn instead of exiting when the dune-project file can't be parsed
  by `pull` as it only use it to suggest updating the lang version for
  convenience (ocamllabs/opam-monorepo#191, @NathanReb)

- Fix a bug where `pull` and `lock` would expect the lockfile to sit in a
  different place and pull would fail. `pull` now simply looks for a
  `.opam.locked` file and pulls it unless there are multiple matching files in
  the repository's root. (ocamllabs/opam-monorepo#163, @NathanReb)

- Fix failure when a package is pinned to a specific commit. `lock` now skips
  resolution when the ref is actually a commit pointed by a remote branch or
  when it looks like a commit (hexadecimal characters only, at least 7
  characters-long). (ocamllabs/opam-monorepo#195, fixes ocamllabs/opam-monorepo#127, @TheLortex)

### DIFF
--- a/packages/opam-monorepo/opam-monorepo.0.2.4/opam
+++ b/packages/opam-monorepo/opam-monorepo.0.2.4/opam
@@ -14,7 +14,7 @@ bug-reports: "https://github.com/ocamllabs/opam-monorepo/issues"
 depends: [
   "dune" {>= "2.9"}
   "ocaml" {>= "4.08.0"}
-  "odoc" {with-doc}
+  "conf-pkg-config" {build}
 ]
 conflicts: [
   "dune-build-info" {= "2.7.0" | = "2.7.1"}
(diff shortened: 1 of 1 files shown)


mseri:
Looks good, thanks

PR #19501, opened 2021-09-13
fix url.src for chamo

zapashcanon:
@zoggy I tried to use the tarball from gitlab but the md5sum wasn't matching (at least for 2.01) so I went with opam's cache


zoggy:
As you wish :-)

zapashcanon:
If you ever want the repo to point to framagit, you'll have to check that the content is the same for each archive and update the checksums. But that's also OK like this. :)

mseri:
Thanks. We pointed it to framagit and manually checked that the contents where unchanged in a separate PR.

PR #19511, opened 2021-09-13
make ezjs_min.0.2.2 not available

zapashcanon:
the tarball is not even in the cache so there's nothing we can do here...

@maxtori


mseri:
Yes, looking at the commit history I think 0.2.2 is what is now 0.2.3

PR #19518, opened 2021-09-13
[new release] re (1.10.3)

rgrinberg:
RE is a regular expression library for OCaml

- Project page: <a href="https://github.com/ocaml/ocaml-re">https://github.com/ocaml/ocaml-re</a>

##### CHANGES:

* Glob: change optional argument `?backslash_escapes` to `?match_backslashes`.
  The interpretation of backslashes in the glob pattern remains unchanged with
  the new option, but forward slashes match backslashes when activated (ocaml/ocaml-re#199)

### DIFF
--- a/packages/re/re.1.10.3/opam
+++ b/packages/re/re.1.10.3/opam
@@ -1,0 +1,46 @@
+opam-version: "2.0"
+
+maintainer: "rudi.grinberg@gmail.com"
+authors: [
+  "Jerome Vouillon"
+  "Thomas Gazagnaire"
+  "Anil Madhavapeddy"
+  "Rudi Grinberg"
+  "Gabriel Radanne"
+]
+license: "LGPL-2.0 with OCaml linking exception"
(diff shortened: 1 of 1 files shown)


mseri:
I went through the pain of checking all the failures. The only suspicious one is
https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/6f47bdb08bc513f7de48241824b9ef2d9411878a/variant/opam-2.0,compilers,4.12,re.1.10.3,revdeps,owork.0.1.1
Can you please check it and confirm that is not an issue with re? Then I think we can merge

rgrinberg:
I'm not really qualified to answer, but It doesn't seem related. The failure is in the package's test suite, and I have no idea if/how they use re.

mseri:
I had some time to look at the code. It looks like they don't even use Re... It must be a transitive dependency. Let's merge, thanks for looking into it

PR #19523, opened 2021-09-14
fix url.src for imagemagick

zapashcanon:
@kit-ty-kate there's something weird with this one: 0.34 and 0.34-1 are exactly the same ! Should we remove 0.34-1 ?

@fccm fyi, your pages on `linux-nantes.org` are down so we can't acess the dev-repo anymore nor the tarballs.

EDIT: OK actually, 0.34-1 seems to contain a patch in `extra-files`, my bad. Yet I'm surprised: why hasn't the patch been added to 0.34 directly ?


kit-ty-kate:
> Yet I'm surprised: why hasn't the patch been added to 0.34 directly ?

In some cases, when the patch might break current installation (or simply out of an excess of caution) we create a new revision for it. With some luck https://github.com/ocaml/opam/issues/4566 should avoid that all together.

Thanks

zapashcanon:
Makes sense. :)

PR #19524, opened 2021-09-14
fix url.src for iri

zapashcanon:
@zoggy fyi


kit-ty-kate:
Thanks!

zoggy:
iri has moved to Framagit: https://framagit.org/zoggy/ocaml-iri
I think the `iri.opam` file in its repo is correct.

PR #19531, opened 2021-09-14
Octez (Tezos) 10.2

pirbo:
At last, here is the packaging of the last major release of Octez distribution of tezos.

Replace #19330


kit-ty-kate:
Looks ok. Thanks!

PR #19533, opened 2021-09-14
8 packages from savonet/ocaml-ffmpeg at 1.0.0

toots:
This pull-request concerns:
-`ffmpeg.1.0.0`: Bindings for the ffmpeg libraries
-`ffmpeg-av.1.0.0`: Bindings for the ffmpeg libraries -- top-level helpers
-`ffmpeg-avcodec.1.0.0`: Bindings for the ffmpeg avcodec library
-`ffmpeg-avdevice.1.0.0`: Bindings for the ffmpeg avdevice library
-`ffmpeg-avfilter.1.0.0`: Bindings for the ffmpeg avfilter library
-`ffmpeg-avutil.1.0.0`: Bindings for the ffmpeg avutil libraries
-`ffmpeg-swresample.1.0.0`: Bindings for the ffmpeg swresample library
-`ffmpeg-swscale.1.0.0`: Bindings for the ffmpeg swscale library

---
* Homepage: https://github.com/savonet/ocaml-ffmpeg
* Source repo: git+https://github.com/savonet/ocaml-ffmpeg.git
* Bug tracker: https://github.com/savonet/ocaml-ffmpeg/issues

---
:camel: Pull-request generated by opam-publish v2.0.3

PR #19534, opened 2021-09-14
Package liquidsoap.2.0.0~rc1

toots:
### `liquidsoap.2.0.0~rc1`
Swiss-army knife for multimedia streaming
Liquidsoap is a powerful and flexible language for describing your
streams. It offers a rich collection of operators that you can combine
at will, giving you more power than you need for creating or
transforming streams. But liquidsoap is still very light and easy to
use, in the Unix tradition of simple strong components working
together.

---
* Homepage: https://github.com/savonet/liquidsoap
* Source repo: git+https://github.com/savonet/liquidsoap.git
* Bug tracker: https://github.com/savonet/liquidsoap/issues

---

### DIFF
--- a/packages/liquidsoap/liquidsoap.2.0.0~rc1/opam
+++ b/packages/liquidsoap/liquidsoap.2.0.0~rc1/opam
@@ -9,9 +9,9 @@ use, in the Unix tradition of simple strong components working
 together."""
 maintainer: "romain.beauxis@gmail.com"
 authors: "The Savonet Team <savonet-users@lists.sourceforge.net>"
+license: "GPL-2.0-or-later"
 homepage: "https://github.com/savonet/liquidsoap"
 bug-reports: "https://github.com/savonet/liquidsoap/issues"
-license: "GPL-2.0-or-later"
 depends: [
   "conf-autoconf" {dev}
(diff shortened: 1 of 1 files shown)


toots:
This is the companion PR for https://github.com/ocaml/opam-repository/pull/19533, fixing all dependency issues with `ffmpeg`. We would really like to avoid doing a new release that close to the final one so hopefully this'll work for y'all as well..

Thanks for y'all help!

toots:
> I compared the two tarballs with pkgdiff, the differences are about printing the version in the config script and the opam file dependency update. Being just a metadata fix it does not warrant a new release, it would have been enough to update the meetadata of the opam-repository, without the need to re-relase the package

The issue is with our own `configure` script. It does not understand opam's pre-release versions, i.e. `1.0.0~rc2` so it kept rejecting `ffmpeg` based on its version. 

I had to re-relase a new tarball with an updated `configure` script, unfortunately, because this is an issue that affects anyone using the `configure` script, not just `opam` users..

mseri:
Makes sense now, thanks

mseri:
Looks good

toots:
Thank you!

PR #19536, opened 2021-09-14
fix url.src for mirage-http

zapashcanon:
@samoht @avsm fyi


kit-ty-kate:
Thanks!

samoht:
I suspect that the correct fix for these is to use https://github.com/mirage/ocaml-cohttp /cc @dinosaure

PR #19537, opened 2021-09-14
fix url.src for mmseg

zapashcanon:
@kandu fyi

@kit-ty-kate this one was not available in the cache so I used the one on github (before it was renamed to wseg) which mean the checksum is different. I couldn't check the content was the same as we don't have the original anywhere...


kandu:
@zapashcanon Thanks for your recent PRs

I can confirm that this fix is proper. Though the checksum differs, which is caused by the different packaging scheme between bitbucket and github, the content is the same.

PR #19538, opened 2021-09-14
fix url.src for nebula

zapashcanon:
@hakuch fyi. Also, I wonder why you deleted the repository, has it been moved somewhere else ?

### DIFF
--- a/packages/nebula/nebula.0.2.1/opam
+++ b/packages/nebula/nebula.0.2.1/opam
@@ -21,7 +21,7 @@ depends: [
   "utop" {>= "1.18" & < "2.0"}
   "ocamlbuild" {build}
 ]
-synopsis: "DCPU-16 emulator."
+synopsis: "DCPU-16 emulator"
 description:
   "Nebula is a complete DCPU-16 emulator including simulated hardware devices."
 url {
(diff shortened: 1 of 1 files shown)


hakuch:
@zapashcanon: I appreciate your interest in the project, but I'm no longer maintaining it. It's not hosted elsewhere. My preference would be for the opam package to be deleted: I've learned a great deal since then and I don't look back on it especially fondly :stuck_out_tongue:.

 I certainly wouldn't use it as any kind of model for programming with monads, though it may (or may not!) be a useful reference for implementing the DCPU-16 specification (I can't vouch for its correctness, since it's been a long time).

zapashcanon:
I think that the policy for the opam-repository is to try to be immutable as much as possible, which means not deleting packages. Thus I don't think removal is likely to happen... But @kit-ty-kate or @mseri are probably more knowledgeable than me about it. :)

kit-ty-kate:
> I think that the policy for the opam-repository is to try to be immutable as much as possible, which means not deleting packages. Thus I don't think removal is likely to happen... But @kit-ty-kate or @mseri are probably more knowledgeable than me about it. :)

That's right. We try to keep every versions of every packages as much as possible to avoid breaking users who might depend on it

Thanks!

PR #19554, opened 2021-09-14
fix url.src for opencc0

zapashcanon:
@kandu  fyi


mseri:
Can't be tested anywhere, but at least lints fine :)

PR #19569, opened 2021-09-16
fix url.src for ppx_monadic

zapashcanon:
@camlspotter fyi

@kit-ty-kate I made 2.2.1 `available: false` because it has no hash and thus I can't find it in the cache.


zapashcanon:
I've been able to find the tarball for 2.2.1 thanks to @rjbou, so no more `available: false` is needed. :)

PR #19586, opened 2021-09-16
fix url.src for stog-rdf

zapashcanon:
@zoggy fyi

### DIFF
--- a/packages/stog-rdf/stog-rdf.0.16.1/opam
+++ b/packages/stog-rdf/stog-rdf.0.16.1/opam
@@ -28,6 +28,6 @@ documents and to generate RDF graphs from elements in
 documents."""
 flags: light-uninstall
 url {
-  src: "https://github.io/zoggy/stog-rdf/archive/0.16.1.tar.gz"
+  src: "https://framagit.org/zoggy/stog-rdf/-/archive/0.16.1/stog-rdf-0.16.1.tar.gz"
   checksum: "md5=c5ceda5190eb8c0333e795cc7d1b7d31"
 }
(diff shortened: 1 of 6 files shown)


packages/stog-rdf/stog-rdf.0.16.1/opam
```diff
@@ -28,6 +28,6 @@ documents and to generate RDF graphs from elements in
 documents."""
 flags: light-uninstall
 url {
-  src: "https://github.com/zoggy/stog-rdf/archive/0.16.1.tar.gz"
+  src: "https://github.io/zoggy/stog-rdf/archive/0.16.1.tar.gz"
```
mseri, line 31:
```
2021-09-16 19:32.16: Error in stog-rdf.0.16.1: Failed to download the archive. Details: https://github.io/zoggy/stog-rdf/archive/0.16.1.tar.gz (Bad checksum, expected md5=c5ceda5190eb8c0333e795cc7d1b7d31)
```

zapashcanon: I can update the checksum but I can't check that the content is the same as `https://opam.ocaml.org/cache/c5/c5ceda5190eb8c0333e795cc7d1b7d31` is not in cache and leads to a 404... Is it OK for you ?

mseri: I don't know what is the best thing to do. Let's wait for @kit-ty-kate opinion, we have the same issue in at least another PR I think

mseri:
Closing for now

zoggy:
This should be fixed in https://github.com/ocaml/opam-repository/pull/20186 .

PR #19590, opened 2021-09-16
fix url.src for tcx

zapashcanon:
@a12n fyi

### DIFF
--- a/packages/tcx/tcx.0.25.0/opam
+++ b/packages/tcx/tcx.0.25.0/opam
@@ -15,6 +15,6 @@ depends: [
 synopsis:
   "OCaml library for parsing and formatting Training Center XML files."
 url {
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Thanks

a12n:
@zapashcanon Thanks for the fix. Bitbucket was removing all the Mercurial repos, so I've moved it to [GitHub](https://github.com/a12n/ocaml-tcx). The `tar.gz` file may be fetched from https://github.com/a12n/ocaml-tcx/archive/refs/tags/0.25.0.tar.gz.

zapashcanon:
@a12n thanks ! But I believe that the checksum will be different, so I would have to check the content of the two archives and I don't have the time to do it. If you're willing to do it, then you could change the link to point to github instead ?

PR #19606, opened 2021-09-17
a newer version of elina 1.3.2 is proposed,

GgnDpSngh:
can 1.3.1 be deleted?


kit-ty-kate:
Thanks!

> can 1.3.1 be deleted?

Is there a particular reason? We try to avoid this situation as much as possible to avoid breaking people's lock files or setup

PR #19609, opened 2021-09-19
OCaml 4.13 preview for ocaml-lsp-server

kit-ty-kate:
Same as #19408 but for `ocaml-lsp-server`
Code mirrors https://github.com/ocaml/ocaml-lsp/pull/506

cc @ulugbekna @rgrinberg @voodoos


voodoos:
If I am not mistaken this PR does not disable `-short-paths` which is always-on in OCaml-LSP and will produce issues when the flag `principal` is also used. See https://github.com/ocaml/merlin/issues/1389

@ulugbekna

kit-ty-kate:
That's ok, it doesn't happen often from my testing and the package will automatically get fixed as soon as it is fixed in the parent branch. It's only a preview

PR #19627, opened 2021-09-23
Fix dependencies on `tar.1.{0.1,1.0}`

dinosaure:
This PR is not about `tar.2.0.0` but about dependencies on `tar.1.{0.1,1.0}` to prepare then the release of `tar.2.0.0`.


MisterDA:
Instead of the changes in https://github.com/ocaml/opam-repository/pull/19627/commits/9f4f5c2e90e6efe85a3b0eb048266ad8a9ec058a, I think you should replace the `< 2.0.0` with `= version`, and I'm not sure that `>= 2.0.0` will work (but the CI will tell) considering the changes to `Tar.Header`. Maybe `= version` would be better here too?

dinosaure:
> Instead of the changes in 9f4f5c2, I think you should replace the < 2.0.0 with = version

Such change should be applied only on new packages I think, @kit-ty-kate what do you think about that?

PR #19629, opened 2021-09-23
coq is not compatible with no-naked-pointers

kit-ty-kate:
cc @ejgallego 
```
#=== ERROR while compiling coq.8.13.2 =========================================#
# context              2.1.0 | linux/x86_64 | ocaml-options-only-nnp.1 | file:///home/opam/opam-repository
# path                 ~/.opam/4.12/.opam-switch/build/coq.8.13.2
# command              ~/.opam/opam-init/hooks/sandbox.sh build make -j31
# exit-code            2
# env-file             ~/.opam/log/coq-19-b76bba.env
# output-file          ~/.opam/log/coq-19-b76bba.out
### output ###
# COQC -noinit theories/Init/Notations.v
# make[1]: *** [Makefile.build:863: theories/Init/Notations.vo] Segmentation fault
```

### DIFF
--- a/packages/coq/coq.8.13.0/opam
+++ b/packages/coq/coq.8.13.0/opam
@@ -27,6 +27,9 @@ depends: [
   "conf-findutils" {build}
   "zarith" {>= "1.10"}
 ]
+conflicts: [
+  "ocaml-option-nnp"
+]
 build: [
   [
     "./configure"
--- a/packages/coq/coq.8.13.1/opam
+++ b/packages/coq/coq.8.13.1/opam
@@ -27,6 +27,9 @@ depends: [
   "conf-findutils" {build}
   "zarith" {>= "1.10"}
 ]
+conflicts: [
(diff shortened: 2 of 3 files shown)


ejgallego:
Thanks @kit-ty-kate , this is a known problem and we were discussing about it recently (@ppedrot @skyskimmer)  [in the context of multicore]

Not sure if we have folks an upstream issue about this? But indeed, we should try to adopt a solution.

Maybe for now, we can have a configure flag that disables native and the vm when required?

avsm:
A config option sounds good @ejgallego -- you may want to adopt a mechanism similar to the `ocaml-options` packages to control their setting. For now I'll merge this as the released Coq will be incompatible.

ejgallego:
Thanks! We will try indeed to detect if the compiler is in no-naked-pointers and disable the incompatible flags; upstream issue is coq/coq#13940 for those interested.

PR #19630, opened 2021-09-23
Add opam package for gappa 1.4.0

MSoegtropIMC:
This PR adds an opam package for the new version 1.4.0 of gappa (a generator for arithmetic precision proofs).

FYI: @silene 

Btw: for building on platforms where C++11 is not the default, I still need the attached patch to the configure script. You might want to merge this into gappa upstream, but I have no issue keeping it as opam patch.


mseri:

PR #19633, opened 2021-09-24
OpenCC 1.1

kandu:
OpenCC had updated its SOVERSION to 1.1, so we need new bindings for the new version.

1. Only the C++ interface of opencc was changed. This package only depends on the C interface, so the ocaml interface is still.
2. opencc in the FreeBSD ports is not updated, so we drop FreeBSD depext item for now.


mseri:
It works everywhere it can, thanks!

PR #19636, opened 2021-09-24
OCaml 4.13.0 release

Octachron:
This is the package update for the joint release of OCaml 4.13.0 and 4.12.1 .
For both version, this PR updates the following packages
- `ocaml-variants*+options`
- `ocaml-base-system`
- `ocaml-system`
- `ocaml-src`

For OCaml 4.13.0, this PR also updates the packages
- `ocaml.4.13.1` to unlock trunk version of OCaml on the 4.13 branch
- `ocaml-variants*+trunk`
- `ocaml-manual`

I have not updated the trunk version of 4.12.1 to avoid creating a 4.12.2 metapackage.


dra27:
> I have not updated the trunk version of 4.12.1 to avoid creating a 4.12.2 metapackage.

`ocaml.4.12.2` became necessary as soon as `VERSION` on the 4.12 branch changed to `4.12.2+dev...` - at the moment, `ocaml-variants.4.12.1+trunk` doesn't install completely (because the `ocaml.4.12.1` package can't install when the compiler reports a version of 4.12.2)

Octachron:
I added a fake `4.12.2` package. It seems somewhat debatable to have to have such a package when there is no 4.12.2 release expected.

kit-ty-kate:
Thanks a lot!

PR #19649, opened 2021-09-25
Add ez_subst.0.2.0

lefessan:
`Ez_subst` is a simple library to replace variables within strings by other strings. It can be configured to use different characters ('%', '$', etc.), and use different substitutions depending on the format used (`$X`, `${X}`, `$(X)` and `$[X]` can be processed by different substitutions), and recursive ones.

### DIFF
--- a/packages/ez_subst/ez_subst.0.2.0/opam
+++ b/packages/ez_subst/ez_subst.0.2.0/opam
@@ -1,0 +1,47 @@
+opam-version: "2.0"
+license: "LGPL-2.1-only with OCaml-LGPL-linking-exception"
+synopsis: "Ez_subst is a simple module to perform string substitutions"
+description: """
+Ez_subst is a simple module to perform string substitutions, like
+${brace}, $(paren), $[bracket] or $alpha. It provides many options
+to configure the substitutions.
+"""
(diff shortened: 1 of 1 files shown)


lefessan:
> [ERROR] No solution for ez_subst.0.2.0: The actions to process have cyclic dependencies:
>         - install ocamlformat.0.16.0 -> install bisect_ppx.2.6.0 -> install odoc.1.5.3 -> install ocamlformat.0.16.0

lefessan:
> ez_subst.0.2.0 (failed: Error: Could not update repository "beta": "/usr/bin/git fetch -q" exited with code 128 "fatal: unable to look up github.com (port 9418) (Temporary failure in name resolution)")

lefessan:
I think I can't do anything to solve these issues on my side, can I ?

dinosaure:
It seems that you have a cycle on `4.07` because you depends on `ocamlformat` for testing. I did a quick look on your repository and it seems that `ocamlformat` is only for the development purpose, do you really need `ocamlformat` to execute tests?

lefessan:
So, I just removed support for 4.07, it looks like it was enough to remove the cycle in dependencies.

PR #19655, opened 2021-09-27
release ocaml-freestanding 0.6.5

hannesm:
* Add support for OCaml 4.13 (mirage/ocaml-freestanding#95 mirage/ocaml-freestanding#94, @kit-ty-kate @dra27)
* Remove cflags and libs flags that are no longer used in the MirageOS-with-dune development (mirage/ocaml-freestanding#91 @hannesm, reverts mirage/ocaml-freestanding#50 @TheLortex). These were introduced in version 0.4.4.

### DIFF
--- a/packages/ocaml-freestanding/ocaml-freestanding.0.6.5/opam
+++ b/packages/ocaml-freestanding/ocaml-freestanding.0.6.5/opam
@@ -1,0 +1,35 @@
+opam-version: "2.0"
+maintainer: "Martin Lucina <martin@lucina.net>"
+authors: "Martin Lucina <martin@lucina.net>"
+homepage: "https://github.com/mirage/ocaml-freestanding"
+bug-reports: "https://github.com/mirage/ocaml-freestanding/issues/"
+license: "MIT"
+tags: "org:mirage"
+dev-repo: "git+https://github.com/mirage/ocaml-freestanding.git"
+build: [make]
+install: [make "install" "PREFIX=%{prefix}%"]
+depends: [
+  "conf-pkg-config"
+  "ocamlfind" {build}
+  "ocaml-src" {build}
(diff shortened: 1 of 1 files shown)


hannesm:
I am not sure what to do with the lower-bounds failures: any solo5-bindings* package before 0.6.5 (see https://github.com/Solo5/solo5/pull/459) does not work nicely with gcc 10. Now, restricting ocaml-freestanding to solo5-bindings* >= 0.6.5 feels wrong. Marking solo5-bindings* < 0.6.5 as not available also feels wrong.

A good solution would be to depend in solo5-bindings* on "if gcc is used as cc, then require gcc < 10", but I suspect this is hard in opam.

On the other hand, already having spend 5 minutes on this, I would be fine to mark these old solo5 packages as unavailable. Development is moving forward, no need to support and spend time on arcane versions that are unlikely being used by anyone.

What do you think (esp. @mseri @kit-ty-kate)?

hannesm:
see https://github.com/ocaml/opam-repository/pull/19656 as my proposal on this topic.

hannesm:
I looked through the failures, they are unrelated to this release. The only action item are the solo5-bindings-* lower bounds (gcc), as outlined in the above comment, and being addressed in a separate PR.

hannesm:
is there anything I can do to move forward with this PR?

mseri:
In this case, I think we should just let the lower bound check to fail. Once it will be possible to encode:

> A good solution would be to depend in solo5-bindings* on "if gcc is used as cc, then require gcc < 10", but I suspect this is hard in opam.

we can think of doing that

mseri:
I just came back from holiday and Kate is also in holiday, it will take some time before we restart merging at the usual pace

PR #19656, opened 2021-09-27
solo5-bindings-* < 0.6.5: mark unavailable since they need gcc < 10

hannesm:
as suggested in https://github.com/ocaml/opam-repository/pull/19655#issuecomment-927702506


mseri:
I think we should leave them available. If anything we could have a post-install failure message saying that in case of failure the user should make sure that gcc < 10 is installed and selected as compiler.

hannesm:
agreed

Issue #18836, opened 2021-06-09
An env var affects the build of coq


ana-borges:
If `COQ_USE_DUNE` is set, then the opam installation of coq will try to use dune and end up not installing anything.

Steps to reproduce:
```shell
export COQ_USE_DUNE=true
opam switch create test 4.12.0
opam update
opam repo add coq-released https://coq.inria.fr/opam/released
opam install coq -vv # observe dune is being used and an error is produced
```

Cf. https://github.com/coq/coq/issues/14469, https://github.com/coq/opam-coq-archive/issues/1736, and the [zulip discussion](https://coq.zulipchat.com/#narrow/stream/237656-Coq-devs.20.26.20plugin.20devs/topic/Installing.20coq.2Edev.20through.20opam.2C.20missing.20binaries).

avsm:
We do need a coq package maintainer to submit these fixes to the opam-repository by forcing the variable to be unset in the build rules.

PR #19663, opened 2021-09-28
Unset COQ_USE_DUNE during coq build

ana-borges:
Fix for #18836.


ejgallego:
Thanks @ana-borges , the fix should only be needed for Coq 8.10-8.13 tho, let me actually check when this problem was introduced.

ejgallego (changes requested):
Thanks for the patch @ana-borges , however this fix is only needed for Coq 8.12.x and Coq 8.13.x ; I thus suggest you amend the PR.

ejgallego:
Also, I'd suggest you update the PR / commit message to `fixes #18836` , so Github will link this patch to the issue it is solving

ana-borges:
> this fix is only needed for Coq 8.12.x and Coq 8.13.x ; I thus suggest you amend the PR.

Thank you for checking @ejgallego.

Did you also check CoqIDE? I tried checking again and couldn't install it with or without `COQ_USE_DUNE`, probably for an unrelated reason (I remember not being able to install last year either). So I may easily have been wrong yesterday when I concluded this problem affected it as well.

ejgallego:
Thanks @ana-borges , indeed coqide packages need the same fix.

packages/coq/coq.8.13.2/opam
```diff
@@ -30,6 +30,9 @@ depends: [
 conflicts: [
   "ocaml-option-nnp"
 ]
+build-env: [
+  [COQ_USE_DUNE="false"]
```
ejgallego, line 34: Actually, I'm not sure this will work; the test is coq's makefile is `ifndef COQ_USE_DUNE` , thus, I think the right fix is  unset COQ_USE_DUNE` before the `make world` call in the script itself.

ana-borges: Currently it indeed doesn't work (as you can see in the CI). Frustratingly, I can't get anything to work locally at all, meaning that even when I checkout `master`, `cd`ing to, say, `packages/coq/coq.8.13.2` and doing `opam install .` fails (with `bwrap: execvp ./configure: No such file or directory`). However, the env log does show that `COQ_USE_DUNE` is set to `false` in this branch, which I take as a minor win.

Am I just testing things wrong?

ejgallego: I think the way to test things is indeed a bit different; I'm not an expert, but the way I test this kind of stuff, is to do `opam add local $path_to_my_repos` 

I guess you testing is failing due to `opam install .` expecting the sources of Coq to be there too.

ana-borges: Ok! I can now properly test things, thank you very much.

As you predicted, my "fix" doesn't fix anything. The env log shows `COQ_USE_DUNE` set to false, but `dune` is clearly being used anyway.

ejgallego: Yup, you need to do `unset COQ_USE_DUNE && make` in the build field.

ejgallego: Umm, no idea; does `make "COQ_USE_DUNE=" "-j%{jobs"` work locally? If so, that should may be the preferred form.

ana-borges: It... almost works? Your example made me realize that `[make "-j%{jobs}%" "COQ_USE_DUNE=false"]` works as well, I was just too pessimistic to actually read the error message and realize it was a new, different error. I'm now stuck at `make install-byte`, for some as yet unknown reason.

ejgallego: You need to add that line to all calls to make unfortunately, but now that I think, maybe setting in the `env` part as you did first `[ COQ_USE_DUNE= ]` would actually work too.

ana-borges: Yes, adding it to the `make install byte` line made everything actually work! For some reason it wasn't needed in the `make install` line (???).

> maybe setting in the env part as you did first [ COQ_USE_DUNE= ] would actually work too.

I will try it for curiosity's sake, but is that really different from `[COQ_USE_DUNE=false]`?

ejgallego: Yes, because the makefile has `ifdef COQ_USE_DUNE`

ejgallego: By the way I'm sure it is needed in the `install` call too, if you that route; likely omitting it will cause the install target to turn into a no-op

ana-borges: > maybe setting in the env part as you did first [ COQ_USE_DUNE= ] would actually work too.

It seems that `opam` does not allow this; it gives a syntax error, even with `[ COQ_USE_DUNE="" ]`. So I guess we're going with the other option.

ejgallego: Pity, that would have been clearer. So indeed we need undefine the var in every call to make; annoying ;S

ana-borges:
@mseri: There are a bunch of CI failures, which I guess should be fixed before merging, right?

I failed to reproduce the errors due to not being able to install all dependencies, and then I got busy with other stuff and couldn't pick this up again. Next week I will have more time. If you or anyone has some insight on the errors, I would appreciate it.

PR #19673, opened 2021-09-30
[new release] domainslib (0.3.1)

Sudha247:
Changes:

* https://github.com/ocaml-multicore/domainslib/pull/45 adds support for named pools. This is a breaking change with setup_pool taking an optional name parameter and an extra unit parameter.

* A minor bug fix in `parallel_for_reduce`.


Sudha247:
The macos jobs have timed out, any idea what needs to be done there?

dinosaure:
> The macos jobs have timed out, any idea what needs to be done there?

It seems common that the MacOS timed out for many tests.

PR #19674, opened 2021-09-30
Avoid previews when possible

kit-ty-kate:
Prevents upgrades to OCaml 4.13 while the previews are still the only versions available.
This does not prevent people from installing those preview explicitly nor does this prevent people from installing those packages on 4.13 switches.

Fixes https://discuss.ocaml.org/t/opam-prevent-preview-in-upgrade/8552

### DIFF
--- a/packages/merlin/merlin.4.3.2~4.13preview/opam
+++ b/packages/merlin/merlin.4.3.2~4.13preview/opam
@@ -10,6 +10,7 @@ build: [
   ["dune" "build" "-p" name "-j" jobs]
   ["dune" "runtest" "-p" "merlin,dot-merlin-reader" "-j" jobs] {with-test}
 ]
+flags: avoid-version
 depends: [
   "ocaml" {>= "4.13" & < "4.14"}
   "dune" {>= "2.9.0"}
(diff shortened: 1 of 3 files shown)


dra27:
Just to expand with examples:

- If you are in a 4.13 switch (e.g. `opam switch create 4.13 ocaml-base-compiler.4.13.0`) and you `opam install merlin`, you will get the 4.13 preview, but it will upgrade when the upstream release happens
- If you are in a 4.12 switch (e.g. `opam switch create 4.12 ocaml-base-compiler.4.12.1`) and you `opam install merlin`, you will get 4.3.1-412 (i.e. latest stable) as before
- opam will not upgrade to OCaml 4.13 if you have installed merlin in the switch and, when doing `opam switch create .` opam will not select 4.13 if merlin (or lsp-server) are included in the dependencies of the opam files which get pinned.

So, for example:

```
opam switch create my-project ocaml-base-compiler.4.12.1
opam switch set-invariant ocaml-base-compiler
opam install merlin
opam upgrade
```
will stick with OCaml 4.12.1, but if you:
```
opam remove merlin
opam upgrade
```
you'll get upgraded to OCaml 4.13.0. NB in opam 2.0.x, none of this applies - you'll get the preview version, but opam 2.0.x doesn't tend to upgrade compilers.

PR #19680, opened 2021-10-01
Fix opus depext on macOS

toots:
Sidenote: `opam` `2.1.0` will keep failing even if the depext is installed with the fixed name. This is not really robust as depext do change over OS release. Maybe `--assume-depext` should be implicitly tried at least once?


mseri:
There should be an open issue about this in opam, can you check? If so, please add a comment there, otherwise open a new issue. It would be quite useful to have a brooader discussion on this

toots:
Done:  https://github.com/ocaml/opam/issues/4860. Thks!

mseri:
Thanks a lot

PR #19657, opened 2021-09-27
[new release] x509 (0.15.0)

hannesm:
Public Key Infrastructure (RFC 5280, PKCS) purely in OCaml

- Project page: <a href="https://github.com/mirleft/ocaml-x509">https://github.com/mirleft/ocaml-x509</a>
- Documentation: <a href="https://mirleft.github.io/ocaml-x509/doc">https://mirleft.github.io/ocaml-x509/doc</a>

##### CHANGES:

* FEATURE support validation of an IP address in the leaf certificate
  (mirleft/ocaml-x509#152 mirleft/ocaml-x509#153 @reynir @hannesm)
* FEATURE provide Certificate.ips and Certificate.supports_ip
  (mirleft/ocaml-x509#152 @reynir @hannesm)
* BREAKING revise certificate and public key fingerprint authenticators API:
  now a single fingerprint is supported, previously a list of pairs of
  hostname and fingerprint was used (mirleft/ocaml-x509#153 @hannesm)
* BREAKING The Authenticator.t type has been extended with ?ip:Ipaddr.t
  (mirleft/ocaml-x509#153 @hannesm)

PR #19689, opened 2021-10-04
`mirage-crypto` is available on s390x since 0.10.3

dinosaure:
According to #19657, this PR wants to lint `mirage-crypto` package with the `s390x` architecture (/cc @hannesm).


hannesm:
thanks, looks good to me.

dinosaure:
I would like to merge this PR but it seems that `mirage-crypto` (for older versions) has some issues - when the CI got some errors from tests in extra architectures. See specially:
https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/5a734e13b74379838c2011a6bae645cee9494552/variant/opam-2.1,extras,arm32,mirage-crypto.0.10.2,tests

It can be fixed by an other PR but we should look on it. Then, I want to have a double check from someone else to merge it.

hannesm:
Thanks @dinosaure for looking into this. I agree this should be merged since it improves the current state. About the error, `Worker stops running: Killed by signal -22`, is there a way to get more output? Is there a way to figure out what "signal -22" is on arm32? Is there an easy way to reproduce? May it be some overcommit of resources on the arm32 runner?

dinosaure:
> About the error, Worker stops running: Killed by signal -22, is there a way to get more output? Is there a way to figure out what "signal -22" is on arm32? Is there an easy way to reproduce? May it be some overcommit of resources on the arm32 runner?

Currently we should be able to reproduce it with `docker` and `qemu`. Let's say that this PR does not want to solve this specific error. I will try to reproduce it as soon as I can on my side and come back with better outcomes directly on the `mirage-crypto` repository.

PR #19703, opened 2021-10-06
 [new package] pyast.0.1.0 and [new release] metapp.0.4.1 and metaquot.0.5.0 and pattern.0.3.1

thierry-martinez:
This PR adds the new package pyast.0.1.0 and new release metapp.0.4.1, metaquot.0.5.0 and pattern.0.3.1 to add support for OCaml 4.13. This library provides versioned abstract syntax tree for all Python versions.

### DIFF
--- a/packages/metapp/metapp.0.4.1/opam
+++ b/packages/metapp/metapp.0.4.1/opam
@@ -1,0 +1,40 @@
+opam-version: "2.0"
+synopsis: "Meta-preprocessor for OCaml"
+description: """
+Meta-preprocessor for OCaml: extends the language with [%meta ... ]
+construction where ... stands for OCaml code evaluated at
+compile-time.
+"""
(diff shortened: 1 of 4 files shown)


thierry-martinez:
It looks like macos runners are stuck (in particular, one job is cancelled and I believe I can't do anything to resurrect it: https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/c81cb04b30125f830fd8f9b096300bff0968ec08/variant/opam-2.0,distributions,macos-homebrew,pattern.0.3.1 ). Other jobs look good. I had to disable clangml in many architectures because it looks broken now, but I plan to make a new release for clangml very soon (but I would prefer if this would not block the release of pyast).

thierry-martinez:
There seems also to be a bug with metaquot and ocaml domains variant (PPX appears to segfault as soon as the input file is long enough), but I would prefer to fix it after the release.

PR #19710, opened 2021-10-06
Constrain older odigs on odoc < 2.0.0

dbuenzli:
As it should be. https://github.com/ocaml/odoc/issues/748


mseri:
Then I think this is good to go in. We will fix them once we figure them out, I have seen smiliar kind of failures in a few other packages, I will start collecting them to check if I can see a pattern

PR #19711, opened 2021-10-07
new file:   packages/conf-libsvm-tools/conf-libsvm-tools.1.0.0/opam

UnixJunkie:


UnixJunkie:
Yes, certainly. That would remove the conf-which dependency. I'll try to do that.

UnixJunkie:
Taking inspiration from https://stackoverflow.com/questions/592620/how-can-i-check-if-a-program-exists-from-a-bash-script#677212

UnixJunkie:
done

PR #19719, opened 2021-10-07
reverse dependency of tls: prepare those failing for upcoming release

hannesm:
upper bounds for the failures observed in https://github.com/ocaml/opam-repository/pull/19716
- capnp-rpc-mirage/net @talex5 
- conduit-mirage @talex5 @samoht @avsm 
- dream (via vendor of gluten) @aantron
- gluten-lwt-unix @anmonteiro 
- irc-client-tls @c-cube @johnelse 
- paf / paf-cohttp @dinosaure 

The change is actually in X509 0.15.0 that modified the type Authenticator.t (used to be:
```OCaml
type t = host:[`host] Domain_name.t option -> Certificate.t list -> Validation.r
```

now it is:
```OCaml
type t = ?ip:Ipaddr.t -> host:[`host] Domain_name.t option -> Certificate.t list -> Validation.r
```

This means the leaf certificate can be validated based on an IP address in the subjectAlternativeName extension instead of a hostname (this is crucial for DNS-over-TLS). In all the above packages you seem to have copied over the type for an authenticator -- the easy fix is to add `?ip:_` to your function.

### DIFF
--- a/packages/capnp-rpc-mirage/capnp-rpc-mirage.1.1/opam
+++ b/packages/capnp-rpc-mirage/capnp-rpc-mirage.1.1/opam
@@ -17,7 +17,7 @@ depends: [
   "fmt"
   "logs"
   "dns-client" {>= "5.0.0"}
-  "tls-mirage"
+  "tls-mirage" {< "0.15.0"}
   "mirage-stack" {>= "2.0.0"}
   "arp" {>= "2.3.0" & with-test}
   "alcotest" {>= "1.0.1" & with-test}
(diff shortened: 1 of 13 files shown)


aantron:
The constraint added to Dream introduces a false dependency on `tls`. Dream doesn't actually depend on `tls`, just existing releases *conflict* with it. I will open a new PR in a few days to fix this.

hannesm:
@aantron indeed - sorry about that. it should be the same as gluten-lwt-unix: an optional dependency on tls, and a conflict with tls < 0.15.0.

kit-ty-kate:
I opened https://github.com/ocaml/opam-repository/pull/19973 to fix that. Sorry about that.

aantron:
@kit-ty-kate thank you!!

PR #19741, opened 2021-10-10
+ cbor 0.5

ygrek:


dinosaure:
It seems that your package needs the `Float` module which is available only since 4.07. You can put a constraint such as your package needs, at least, 4.07 or you can use `stdlib-shims` to keep the availability on older version if you want.

smorimoto:
I just opened a PR to address them: https://github.com/ygrek/ocaml-cbor/pull/17

PR #19745, opened 2021-10-11
[release] gmp.6.2.1-1

TheLortex:
## `gmp.6.2.1-1`

`gmp` is a small package wrapping the `gmp` library build system inside the `dune` build system. 

It statically builds `gmp` from source and allows, when built using the _dune workspace_ feature, ✨cross-compilation✨.
See https://github.com/mirage/ocaml-gmp

**CHANGES:**
This release incorporates a set of patches in the build system to fix cross-compilation issues.


dinosaure:
> I think you could make it work using gcc in principle

Can we get `gcc` on Mac OS? I'm not a Mac OS user 😕

Issue #19700, opened 2021-10-06
[lower bound] mdx 1.8.0, 1.8.1 and ocamlformat-rpc 0.18.0


mseri:
See: https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/c1ab98721ec2efc4c65d0c5a65a85c826925db6c
```
4.02

    mdx.1.8.0
        lower-bounds (failed: Library "logs.fmt" not found.)

4.03

    mdx.1.8.0
        lower-bounds (failed: Library "logs.fmt" not found.)
    mdx.1.8.1
        lower-bounds (failed: Library "logs.fmt" not found.)

4.04

    mdx.1.8.0
        lower-bounds (failed: Unbound module Migrate_parsetree.Ast_408)
    mdx.1.8.1
        lower-bounds (failed: Unbound module Odoc_model)

4.05

    mdx.1.8.0
        lower-bounds (failed: Unbound module Migrate_parsetree.Ast_408)
    mdx.1.8.1
        lower-bounds (failed: Unbound module Odoc_model)

4.06

    mdx.1.8.0
        lower-bounds (failed: Unbound module Migrate_parsetree.Ast_408)
    mdx.1.8.1
        lower-bounds (failed: Unbound module Odoc_model)

4.07

    mdx.1.8.0
        lower-bounds (failed: Unbound module Migrate_parsetree.Ast_408)
    mdx.1.8.1
        lower-bounds (failed: Unbound module Odoc_model)

4.08

    mdx.1.8.0
        lower-bounds (failed: Unbound module Odoc_model)
    mdx.1.8.1
        lower-bounds (failed: Unbound module Odoc_model)
    ocamlformat-rpc.0.18.0
        lower-bounds (failed: Unbound module Glob)

4.09

    ocamlformat-rpc.0.18.0
        lower-bounds (failed: Unbound module Glob)

4.10

    ocamlformat-rpc.0.18.0
        lower-bounds (failed: Unbound module Glob)

4.11

    ocamlformat-rpc.0.18.0
        lower-bounds (failed: Unbound module Glob)
```

PR #19752, opened 2021-10-12
OCamlformat and Odoc don't depend on `bisect_ppx {with-test}`

Julow:
Might fix some errors in https://github.com/ocaml/opam-repository/issues/19700

`bisect_ppx` occurs in build script but is never called by the build and the test commands.


smorimoto:
I fixed the CI. It should be all green soon.

PR #19755, opened 2021-10-12
solo5-bindings-* need conf-gcc

samoht:


dra27:
There is a PR for opam-repo-ci to improve the error message, but what it's trying to tell you is that `opam-version: "2.1"` is incorrect - in all files where that's been changed it should be `opam-version: "2.0"` (there is no version 2.1 of the `opam` format).

The `avoid-version` flag is ignored by opam 2.0.x (as that's an opam 2.1 feature). If that's a problem then you either need to do a trick like the [`ocaml-beta`](https://opam.ocaml.org/packages/ocaml-beta/) package, add `available: opam-version >= "2.1"`, or just live with it for opam 2.0.x (I expect that it's actually a case of just living with it?!)

dra27:
> > In retrospect, the opam-version field is extremely unfortunately named
> 
> I wonder who choose this name :p

One who didn’t have the benefit of hindsight!! 🙇

PR #19749, opened 2021-10-12
bap-print.2.0.0: missing dependency

mseri:
Add dependency on re to fix
```
#=== ERROR while compiling bap-print.2.0.0 ====================================#
# context              2.0.9 | linux/x86_64 | ocaml-base-compiler.4.06.1 | file:///home/opam/opam-repository
# path                 ~/.opam/4.06/.opam-switch/build/bap-print.2.0.0
# command              ~/.opam/opam-init/hooks/sandbox.sh build make
# exit-code            2
# env-file             ~/.opam/log/bap-print-947-03b880.env
# output-file          ~/.opam/log/bap-print-947-03b880.out
### output ###
# ./setup.exe -quiet -build 
# ocamlfind ocamlopt -package unix -package ocamlbuild -linkpkg -package findlib myocamlbuild.ml /home/opam/.opam/4.06/lib/ocamlbuild/ocamlbuild.cmx -o myocamlbuild
# /home/opam/.opam/4.06/bin/ocamlfind ocamldep -package text-tags -package bap-demangle -package bap -modules plugins/print/print_main.ml > plugins/print/print_main.ml.depends
# /home/opam/.opam/4.06/bin/ocamlfind ocamlc -c -g -annot -bin-annot -short-paths -w +a-4-6-7-9-27-29-32..42-44-45-48-50-60 -package text-tags -package bap-demangle -package bap -I plugins/print -o plugins/print/print_main.cmo plugins/print/print_main.ml
# + /home/opam/.opam/4.06/bin/ocamlfind ocamlc -c -g -annot -bin-annot -short-paths -w +a-4-6-7-9-27-29-32..42-44-45-48-50-60 -package text-tags -package bap-demangle -package bap -I plugins/print -o plugins/print/print_main.cmo plugins/print/print_main.ml
# File "plugins/print/print_main.ml", line 60, characters 24-32:
# Error: Unbound module Re
# Command exited with code 2.
# make: *** [Makefile:8: build] Error 1
```

See on https://github.com/ocaml/opam-repository/pull/19651

PR #19756, opened 2021-10-12
adds a proper constraint to the bap-std package (continues #19749)

ivg:
In BAP 2.0.0 and 2.1.0 some of the packages were not specifying
correctly the set of dependencies neither in the oasis file nor in the
opam file and were building succesfully because of transitive
depedencies that were brought from other packages, sometimes even
third-party.

In this case `bap-print` is an example of such package, it is failing
with the missing dependency on `re`, which was previously brought by a
dependency on `bap-std`, which, in turn, was depending on the uri
package, which was finally responsible for getting the re module into
scope. The uri package dropped the `re` dependency in versions bigger
than 3.1.0 (by splitting into two subpackages).

I know that the right fix would be to properly specify all depedencies
and do not depend on them, and this is implemented since 2.2.0, more
than a year ago. This fix, albeit quite indirect, is minimal and is a
good showcase that transitive dependencies is a very bad thing.

CC @mseri

### DIFF
--- a/packages/bap-std/bap-std.2.0.0/opam
+++ b/packages/bap-std/bap-std.2.0.0/opam
@@ -43,7 +43,7 @@ depends: [
   "ocamlfind" {>= "1.5.6" & < "2.0"}
   "ppx_jane" {>= "v0.11" & < "v0.12"}
   "regular"
-  "uri"
+  "uri" {<= "3.1.0"}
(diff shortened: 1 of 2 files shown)


mseri:
We can look into the lower bound fixes for bap-std separately in another PR. Thanks again

PR #19773, opened 2021-10-14
Coq 8.14.0

silene:


silene:
Except for one, all the failures are due to GTK failing to install/compile, and thus are completely unrelated to these two new packages.

The only exception is the `nnpchecker` failure, which is known. (Package `ocaml-option-nnp` was already marked as conflicting, but that does not seem to encompass `ocaml-option-nnpchecker`, so I need to mark the latter as conflicting too.)

As of now, the tests for `macos-homebrew` are still missing, by lack of workers.

smorimoto:
Is it not compatible with `nnpchecker` only from 8.14.0? If it's the same for other versions, I think we should add that to all versions that are not compatible with `nnp`.

Issue #19757, opened 2021-10-12
gforge.inria.fr was shut down


mseri:
Many packages some also used by opam are now unavailable. For example mccs is uninstallable because cudf is no longer downloadable.

I think they may have moved to gitlab.inria.fr but I didn’t check. We should probably prioritise this issue

avsm:
I've created a new https://github.com/ocam/opam-source-archives to host selective archives that are important and prone to disappearance from elsewhere. This should tide us along until the Software Heritage archive is available next year for use with opam 2.2

smorimoto:
I see. And it's good to know that swh already has them!

smorimoto:
Hmm? Does GitLab actually generate files that are not the same? That is quite weird.

hannesm:
I'm closing this issue: the opam repository archiving process has identified various missing archives and bad checksums, and has taken measures (if the source tarball couldn't be found, the package has been marked as unavailable). now most packages have been moved to the archive, and some more will follow on february 1st. thanks for the issue and heads up - I hope we have dealt with those packages that we could.

if there are remaining issues or someone finds tarballs with matching checksums, please don't hesitate to open a specific issue.

PR #19776, opened 2021-10-14
Update package overview of camlp5 to move from GForge to GitHub

smorimoto:
Some versions are not yet available on GitHub, so this PR is still WIP.

Related issue: #19757


mseri:
Any update on this?

chetmurthy:
what update?  Last I looked at this, it appears that this is for very, very old versions of camlp5.  is there a reason these should be supported?  Are there users who still rely on these versions ?

chetmurthy:
Ah, ok, maybe the thing that needs to be done, is to remove the 6.xx versions of camlp5 from opam-repository?  If so, I can do that.

mseri:
We should not remove them. If there is no way to obtain the tarball, we should mark them as unavailable imo
Ping @smorimoto who opened the PR, can you check/ask if the tarball is in the opam cache backup?

chetmurthy:
OK, looking deeper into this issue, I see that release 614 has a tag ("rel614") but the other releases don't even have tags in github.  I don't know what happened to them: this was all 8 years ago, long before I took over maintenance.  Given how old these releases are, I don't understand the utility of keeping them around.  There are no log-comments for releases earlier than 6.12, but there are these:

6.12: bd79ec5b5b47c4f118ed020d341ea1152574a511
6.13: c2fa935b7dcfb164c8b87e5c4d4954720695aa16
6.15: 524b6b93e829cf91b378d53c6178c4c7ca4ee07b
6.16: 059c6046d717de2e88249995d44ce2f9e6632167

It isn't even clear that these commits correspond to the named releases.

So I'm not sure what to do here.  If you can explain to me how to mark them as unavailable, I'm happy to do so and submit a PR.  Honestly, why anybody would use any camlp5 prior to the 7.xx series is ..... a mystery to me.

Kakadu:
> So I'm not sure what to do here. If you can explain to me how to mark them as unavailable,

You could try to add `available: false` to the recipe

mseri:
Sorry update ping was meant to @smorimoto. In any case, I am checking if there is a cache backup with the missing tarballs, otherwise I will mark them as unavailable and merge

mseri:
They are all there now

PR #19792, opened 2021-10-16
[new package] DAGaml.0.01

JoanThibault:
Canonical DAG representation and manipulation in OCaml.
Exposes several variants of BDD (Binary Decision Diagrams) under a common framework allowing a simple and fair comparisons

Thank you very much for your time considering this new package =)


JoanThibault:
I dot not understand why : "# make[1]: /bin/ocamlbuild: No such file or directory"
When [ ocamlbuild {build}] is specified in dependencies

mseri:
The binary is called as `/bin/ocamlbuild` But rarely that is the correct path. You should call just `ocamlbuild`

JoanThibault:
ok fixed =)
new error : "undefined reference to `caml_hash_univ_param'" which seems to interverne when linking with GuaCaml.0.03 ....

PR #19808, opened 2021-10-19
[coq] Fix install bug in local opam switches and ld setup in OSX

ejgallego:
Due to our makefiles not setting `--release` correctly on the call to
`dune install`

We also include a patch to send more strict include dirs in OSX so the linker doesn't warn about non-existing dirs.

c.f.
- https://github.com/coq/coq/pull/15045
- https://github.com/coq/coq/pull/15050


rgrinberg:
Merging as this bug is very serious. The remaining CI errors do not seem to be related.

PR #19824, opened 2021-10-20
[updated] conf-libclang.12 and [new release] conf-libclang.13

thierry-martinez:
This PR adds the version conf-libclang.13 to detect libclang <=13.0.x.

The logic of conf-libclang.12 is modified such that if a given version of clang is installed but cannot be compiled with, the script continues finding if an older version is available (in Ubuntu 21.10, clang is packaged such a way that headers of clang 10, clang 11 and clang 12 are missing). conf-libclang.13 implements the same logic.

Moreover, conf-libclang.12 is modified such that clangml.4.4.0 can be used on systems where llvm 12.0.1 is installed even if clangml.4.4.0 does only recognize llvm 12.0.0.

### DIFF
--- a/packages/clangml/clangml.4.4.0/opam
+++ b/packages/clangml/clangml.4.4.0/opam
@@ -10,7 +10,7 @@ homepage: "https://memcad.gitlabpages.inria.fr/clangml/"
 doc: "https://memcad.gitlabpages.inria.fr/clangml/doc/clangml/index.html"
 bug-reports: "https://gitlab.inria.fr/memcad/clangml/issues"
 depends: [
-  "conf-libclang" {= "12"}
+  "conf-libclang" {<= "12"}
   "conf-ncurses"
   "conf-zlib"
   "dune" {>= "1.11.0"}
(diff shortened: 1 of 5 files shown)


thierry-martinez:
What can I do for this PR to be merged?

packages/clangml/clangml.4.4.0/opam
```diff
@@ -10,7 +10,7 @@ homepage: "https://memcad.gitlabpages.inria.fr/clangml/"
 doc: "https://memcad.gitlabpages.inria.fr/clangml/doc/clangml/index.html"
 bug-reports: "https://gitlab.inria.fr/memcad/clangml/issues"
 depends: [
-  "conf-libclang"
+  "conf-libclang" {= "12"}
```
kit-ty-kate, line 13: Is there a reason why this is now required here? Does it not work with 10 or 11 ?

kit-ty-kate: ah nevermind I just saw the new configure option. Mmh, the annoying thing is that this change would break people with lock files with both clangml.4.4.0 and conf-libclang.11. Could this be in a new revision `clangml.4.4.0-1`?

thierry-martinez: I just made a try with `<= 12`: I hope that if the variable `conf-libclang:clangml440_configure_options` is not defined, it will be replaced by the empty string, which will in turn be ignored by `./configure` (with just a warning).

thierry-martinez: Indeed, failures and successes are the same with `<=`!

kit-ty-kate:
Thanks. Sorry for the long wait

PR #19903, opened 2021-10-29
[coq] Fix the compilation of many Coq plugins for OCaml 4.13

ejgallego:
Coq imposes a very strict set of default warnings for plugins plus
`-warn-error`. This was meant to be done only for Coq's CI and
development setup, but this setting wrongly leaked into the release
setup of Coq Opam archive.

We fixup things for OCaml 4.13 while a more complete solution is being
worked on.


ejgallego:
> @ejgallego can we move these changes to `coq.8.14.0-1` or it breaks some `=version` invariants?

IMHO this doesn't warrant a release, we would have to add a lot of constraints to coq-opam-archive to avoid 8.14.0 then.

Note that indeed I'm not a fan of the current "imperative" package update, however, if the opam repos policy still does allow it, IMHO this kind of fixup is the right use case.

Issue #19880, opened 2021-10-27
The result story


hannesm:
Hello,

- Since OCaml 4.03 a `type ('a, 'b) result = Ok of 'a | Error of 'b` is provided by the OCaml Stdlib (back then called Pervasives).
- Since OCaml 4.08 a `Result` module (with combinators and functions) is provided by the OCaml Stdlib
- There is a opam package called "result" which intention was initially to provide the result type for older OCaml versions. Before 1.5 this was the case. Since 1.5, it additionally provides the functionality from the OCaml 4.08 Stdlib module.

Now, I've been using in quite some packages the OCaml Stdlib Result module heavily (all of them are only supported on OCaml 4.08+ - namly ca-certs, ca-certs-nss, dns, ethernet, http-lwt-client, otr, x509, more to come -- a "git grep '"result" {< "1.5"}'" shows many more packages whose conflicts may be lifted), and the lower-bounds checks are telling me that I need to do something in respect to result < 1.5 (either add a dependency to result >= 1.5 or a conflict with result < 1.5). I don't think this is a good solution. The reasoning is that some libraries depend on result, and thus implicitly the result package is used instead of the Stdlib one.

My preferred solution would be to have result < 1.5 depend on "ocaml" {< "4.08.0"} -- than we'd be in an environment where if result is required (e.g. by lwt) and we have OCaml >= 4.08 (as required by e.g. ca-certs-nss), we'd be sure to have the whole Result module (as expected). I'm happy to PR such a change to opam-repository, but would first be interested (to avoid unneeded CI work) if there are arguments against an upper bound for early (< 1.5) result versions?

//cc @kit-ty-kate (as opam-repo hat) @aantron (who did the last result release) @jeremiedimino @xclerc  (who have been releasing earlier result versions) @avsm @dra27 (from the platform team)

hannesm:
Just as a heads up: the `seq` package is in a similar state as `result 1.0` used to be -- with all current releases won't provide the Seq interface of OCaml 4.14. Since 4.14 is not released yet, eventually you could assign some upper bounds to avoid history repeating.

mseri:
Should we think of releasing seq with the new interface and make all previous version incompatible with 4.14 before it is released?

dra27:
[...]

PR #19904, opened 2021-10-29
Only permit result 1.5+ on OCaml 4.14 and later

dra27:
Apropos https://github.com/ocaml/opam-repository/issues/19880#issuecomment-954121920, this takes @hannesm's original suggestion and shifts it from rewriting the past to prewriting the future 🙂

This is a slightly unusual constraining, because we're prohibiting packages which _may_ compile on 4.14, but this stops a bug from 1.0-1.4 from continuing to permeate the ecosystem.


kit-ty-kate:
Looks good to me. It can always be reverted later if need be (but i doubt it)

Thanks a lot!

avsm:
The 'frozen out' packages don't look too important to me either.

PR #19905, opened 2021-10-29
All current versions of kind2 require zmq < 5.1.4

daniel-larraz:
This PR updates version constraints for kind2.

### DIFF
--- a/packages/kind2/kind2.1.4.0/opam
+++ b/packages/kind2/kind2.1.4.0/opam
@@ -24,7 +24,7 @@ depends: [
   "ounit2" {with-test}
   "z3" {< "4.8.9" & with-test}
   "yojson"
-  "zmq" {>= "5.0.0"}
+  "zmq" {>= "5.1.0" & < "5.1.4"}
 ]
 build: [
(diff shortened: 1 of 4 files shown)


packages/kind2/kind2.1.4.0/opam
```diff
@@ -24,7 +24,7 @@ depends: [
   "ounit2" {with-test}
   "z3" {< "4.8.9" & with-test}
   "yojson"
-  "zmq" {>= "5.0.0"}
```
mseri, line 27: Why was 5 wrong?

daniel-larraz: zmq 5.0.0 requires ocaml < 4.06.0 (zmq → configurator < v0.15 → base < v0.14 → ocaml < 4.06.0). All versions of Kind 2 requires at least ocaml 4.07, so zmq 5.1.0 is the smallest version I could install and use to build Kind 2.

mseri: Thanks

PR #19917, opened 2021-10-31
Otoml 0.9.2

dmbaturin:
Due to a breaking change in this release that affects soupault, I also adjusted dependencies in all affected soupault versions to use `{< "0.9.2"}`. The other dependent package, [dirsift](https://opam.ocaml.org/packages/dirsift/) isn't using the affected function and shouldn't need any adjustments.

I'll make a maintenance release to make soupault buildable with the latest version once this PR is merged.

### Breaking changes

`Otoml.get_array` now has type `?strict:bool -> (t -> 'a) -> t -> 'a list`,
that is, it requires an accessor function that will be applied to every item of the array.

For example, you can use `Otoml.find t (Otoml.get_array Otoml.get_string) ["foo"]` to retrieve
an array of strings from a TOML document's key `foo`.

The motivation for the change is that it allows retrieving arrays of unwrapped OCaml values in one step.
The old behaviour can still be emulated using an identify function for the accessor,
for example the built-in `Otoml.get_value : 'a -> 'a`.

### New features

New `Otoml.path_exists t ["some"; "table"; "key"]` allows checking if a key path exists in a TOML document.

`Otoml.Printer.to_string/to_channel` functions now provide `~force_table_array` option. When set to true,
it forces every array that contains nothing but tables to be rendered using the `[[...]]`` table array syntax.

### Bug fixes

Unicode escape sequences are now printed correctly.

If a table has subtables and non-table items, the non-table items are forcibly moved before the first subtable
for printing. This way the output parses correctly, otherwise the non-table items would be mistakenly treated
as subtable members. This way hand-constructed TOML tables are always formatted correctly even if the user
inserts non-table items after a subtable.

### Testing

Added a minimal test suite for the read-write interface, and a TOML test suite client executable for encoder testing.

### DIFF
--- a/packages/otoml/otoml.0.9.2/opam
+++ b/packages/otoml/otoml.0.9.2/opam
@@ -1,0 +1,33 @@
+opam-version: "2.0"
+synopsis:
+  "TOML parsing, manipulation, and pretty-printing library (1.0.0-compliant)"
+description: """\
+OTOML is a library for parsing, manipulating, and pretty-printing TOML files.
+
+* Fully 1.0.0-compliant.
+* No extra dependencies: default implementation uses native numbers and represents dates as strings.
+* Provides a functor for building alternative implementations: plug your own bignum and calendar libraries if required.
+* Informative parse error reporting.
+* Pretty-printer offers flexible indentation options."""
+maintainer: "daniil+opam@baturin.org"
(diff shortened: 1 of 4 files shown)


kit-ty-kate:
Thanks

PR #19922, opened 2021-10-31
Package pyml_bindgen.0.1.0

mooreryan:
### `pyml_bindgen.0.1.0`

---

---

### DIFF
--- a/packages/pyml_bindgen/pyml_bindgen.0.1.0/opam
+++ b/packages/pyml_bindgen/pyml_bindgen.0.1.0/opam
@@ -1,0 +1,48 @@
+opam-version: "2.0"
+synopsis: "Generate pyml bindings from OCaml value specifications"
+maintainer: "Ryan M. Moore"
+authors: "Ryan M. Moore"
+license: "MIT"
+homepage: "https://github.com/mooreryan/ocaml_python_bindgen"
+doc: "https://mooreryan.github.io/ocaml_python_bindgen/"
+bug-reports: "https://github.com/mooreryan/ocaml_python_bindgen/issues"
+depends: [
+  "dune" {>= "2.8"}
+  "angstrom"
+  "base"
+  "cmdliner"
+  "ppx_jane"
+  "re2"
+  "stdio"
+  "ocaml"
+  "core_kernel" {with-test}
+  "ocamlformat" {with-test}
+  "ppx_inline_test" {with-test}
+  "ppx_expect" {with-test}
+  "pyml" {with-test}
+  "bisect_ppx" {dev}
+  "odoc" {with-doc}
+]
+build: [
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
+dev-repo: "git+https://github.com/mooreryan/ocaml_python_bindgen.git"
+url {
+  src:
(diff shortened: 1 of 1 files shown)


mooreryan:
Looks like I need to set the lower bounds on some of the packages. In my other opam packages I have the Jane street libs as ">= v0.12.0" and ocaml compiler ">= 4.08.0". 

I'm not sure about the angstrom Let syntax errors. Is there a way to find the correct lower bounds for that?

kit-ty-kate:
> I'm not sure about the angstrom Let syntax errors. Is there a way to find the correct lower bounds for that?

Sadly apart from trial-and-error there is no easy way that i know of

kit-ty-kate:
in this particular case reading the code source was faster though. Git tells us that the feature is only present since 0.15.0: https://github.com/inhabitedtype/angstrom/commit/f29438e004361b4a2ab61cbc854b1bf6f05aea0d

There is also an issue with the tests on platforms with non-GNU grep. Do you want to fix that first? We can disable the tests but it’s probably better if you fix it on your side if you have the time
```
- (cd _build/.sandbox/42cc81eab5899373147dedd26b3fc310/default && /usr/bin/git --no-pager diff --no-index --color=always -u ../../../default/test/of_pyo_option.t/run.t test/of_pyo_option.t/run.t.corrected)
- diff --git a/../../../default/test/of_pyo_option.t/run.t b/test/of_pyo_option.t/run.t.corrected
- index d8f92e2..7bd6372 100644
- --- a/../../../default/test/of_pyo_option.t/run.t
- +++ b/test/of_pyo_option.t/run.t.corrected
- @@ -18,6 +18,37 @@ error if you try to run the code.
-    $ dune exec ./hi.exe 2> err
-    [1]
-    $ grep -i --silent 'signature mismatch' err
- +  grep: unrecognized option: silent
- +  BusyBox v1.32.1 () multi-call binary.
- +  
- +  Usage: grep [-HhnlLoqvsrRiwFE] [-m N] [-A/B/C N] PATTERN/-e PATTERN.../-f FILE [FILE]...
- +  
- +  Search for PATTERN in FILEs (or stdin)
- +  
- +  	-H	Add 'filename:' prefix
- +  	-h	Do not add 'filename:' prefix
- +  	-n	Add 'line_no:' prefix
- +  	-l	Show only names of files that match
- +  	-L	Show only names of files that don't match
- +  	-c	Show only count of matching lines
- +  	-o	Show only the matching part of line
- +  	-q	Quiet. Return 0 if PATTERN is found, 1 otherwise
- +  	-v	Select non-matching lines
- +  	-s	Suppress open and read errors
- +  	-r	Recurse
- +  	-R	Recurse and dereference symlinks
- +  	-i	Ignore case
- +  	-w	Match whole words only
- +  	-x	Match whole lines only
- +  	-F	PATTERN is a literal (not regexp)
- +  	-E	PATTERN is an extended regexp
- +  	-m N	Match up to N times per file
- +  	-A N	Print N lines of trailing context
- +  	-B N	Print N lines of leading context
- +  	-C N	Same as '-A N -B N'
- +  	-e PTRN	Pattern to match
- +  	-f FILE	Read pattern from file
- +  [2]
```

mooreryan:
Yeah I will fix the non-GNU grep issues, and the bounds, thanks!

Should I open a new pull request when I do?

mseri:
You can also modify this PR, either way is fine

mooreryan:
I opened a new pull request (https://github.com/ocaml/opam-repository/pull/19947).

Thanks for the help/suggestions!

PR #19926, opened 2021-11-01
[new release] ocaml-r (OCaml-R)

pveber:
Objective Caml bindings for the R interpreter

- Project page: <a href="https://github.com/pveber/ocaml-r/">https://github.com/pveber/ocaml-r/</a>
- Documentation: <a href="https://pveber.github.io/ocaml-r/ocaml-r/index.html">https://pveber.github.io/ocaml-r/ocaml-r/index.html</a>

##### CHANGES:

ocaml-r-0.5.0 2021-11-01
------------------------

- added graphics functions (axis, text, points, smooth_scatter,
  boxplot)
- added base function (c, list creation)
- fixed interaction with pkg-config (by @drjdn)

ocaml-r-0.4.0 2020-10-15
------------------------

This version implements a massive API refactoring, which generalizes
and amplifies the design experimented in version 0.2.0. Each R
datatype is represented as an abstract type of a dedicated module. The
`t` type of the module is a bare SEXP (or at rather a custom block
wrapping a bare SEXP). S3 inheritance is represented by module (type)
inclusion.

- addition of an (alco)test suite
- (typed) matrices for logical, character and int types
- row and column access for data.frame and matrix
- access to factor values
- safer interface for list and data.frame components
- modular interface for statistical tests

ocaml-r-0.3.1 2020-07-29
------------------------

- replaced configurator by dune.configurator
- fixed compilation warnings

ocaml-r-0.3.0 2020-05-07
------------------------

- base: added readRDS, saveRDS, cbind, rbind
- graphics: abline
- stats: ecdf
- fixed compilation bug

ocaml-r-0.2.0 2019-06-07
------------------------

- experimented modular (instead of object) interfaces
- a few more wrappers in base, stats and graphics
- improved stub generation
- API documentation

ocaml-r-0.1.1 2018-11-18
------------------------

jbuilder-to-dune transition

ocaml-r-0.1.0 2018-06-06
------------------------

First formal release

### DIFF
--- a/packages/ocaml-r/ocaml-r.0.5.0/opam
+++ b/packages/ocaml-r/ocaml-r.0.5.0/opam
@@ -1,0 +1,48 @@
+opam-version: "2.0"
+synopsis: "Objective Caml bindings for the R interpreter"
+description: """
+
+OCaml-R is a library that can be used to construct R values in memory,
+convert them to OCaml values, and build clean wrappers to R
+functions. It provide a simple means to develop bindings to any R
+package."""
+maintainer: ["philippe.veber@gmail.com"]
+authors: ["Guillaume Yzyquel" "Maxence Guesdon" "Philippe Veber"]
+license: "GPL-3.0-only"
+tags: ["R" "statistics"]
+homepage: "https://github.com/pveber/ocaml-r/"
+doc: "https://pveber.github.io/ocaml-r/ocaml-r/index.html"
+bug-reports: "https://github.com/pveber/ocaml-r/issues"
+depends: [
+  "alcotest" {with-test}
+  "base" {build & >= "v0.14"}
+  "conf-r" {build}
+  "conf-r-mathlib" {build}
+  "dune" {>= "2.5"}
+  "ocaml" {>= "4.08"}
+  "stdio" {>= "v0.14" & build}
+]
+build: [
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
+dev-repo: "git+https://github.com/pveber/ocaml-r.git"
+url {
+  src:
+    "https://github.com/pveber/ocaml-r/releases/download/v0.5.0/ocaml-r-0.5.0.tbz"
+  checksum: [
+    "sha256=35fc05e7852e18309df75caaeb61d466d2eeaeee13bc31e509576da040b3606e"
+    "sha512=5b4ad4e76fe2239acc50edd8b6a3636434111af78d2279fa1f818c8127ebe72262d421d8488adfd15466d1893d9e5a05a5aa754bcbe68a32ecd8b0e4a4b2c19a"
+  ]
+}
+x-commit-hash: "d9c80f09f6f029398bff6164dfb541a6acdc8c49"


packages/ocaml-r/ocaml-r.OCaml-R/opam
```diff
@@ -0,0 +1,48 @@
...
+synopsis: "Objective Caml bindings for the R interpreter"
+description: """
+
+OCaml-R is a library that can be used to construct R values in memory,
+convert them to OCaml values, and build clean wrappers to R
+functions. It provide a simple means to develop bindings to any R
+package."""
+maintainer: ["philippe.veber@gmail.com"]
+authors: ["Guillaume Yzyquel" "Maxence Guesdon" "Philippe Veber"]
+license: "GPL3"
```
mseri, line 11:
```suggestion
license: "GPL-3.0-only"
```
If I am not mistaken this is the correct SPDX identifier

pveber:
Thanks for the review! I've tried to reproduce the errors (e.g. for alpine) using docker, but failed and instead got another error when trying to run (in the container):
```sh
opam repository set-url --strict default "file://$HOME/opam-repository"
```
I get:
```
Step 13/13 : RUN opam repository set-url --strict default "file://$HOME/opam-repository"
 ---> Running in 2f3e1e517b0e
[default] Initialised

<><> Upgrading repositories from older opam format ><><><><><><><><><><><><><><>
Upgrading repository "default"...
Compiler 3.07 successfully converted to package ocaml-base-compiler.3.07
Fetching patches of ocaml-variants.3.07+1 to check their hashes...
[...]
Compiler 4.03.0+beta1-no-debug successfully converted to package ocaml-variants.4.03.0+beta1-no-debug
Compiler 4.03.0+beta2 successfully converted to package ocaml-variants.4.03.0+beta2
[ERROR] In /home/opam/.opam/repo/default/compilers/4.03.0/4.03.0+beta2+flambda/4.03.0+beta2+flambda.comp:
        Mismatching file name and 'version:' field
[ERROR] Strict mode: aborting
```
Don't know what I'm doing wrong, maybe I'm not cloning the right opam-repository? (I cloned pveber:release-ocaml-r-v0.5.0)

mseri:
Oh, that is a strange error. Let me ping @kit-ty-kate that is more acquainted with the Docker images

pveber:
Ok, got it. I just forgot to checkout the correct branch after cloning... In any case, the problem seems related to `conf-r` so in case I find a solution, I'll send it in a different PR.

pveber:
After investigation it seems that the compilation fails on alpine because the R-mathlib package doesn't install pkg-config files which the conf-r and conf-r-mathlib opam packages depend on. So I suppose I can just declare conf-r-mathlib is simply not available on alpine right?

pveber:
Same thing for suse, that doesn't have a separate library for libRmath

mseri:
> So I suppose I can just declare conf-r-mathlib is simply not available on alpine right?

Don’t declare anything like that, people can install manually or peovide the lkgco fig files and then it would work. It is not a problem if it fails in the ci when system packages are missing or incorrect versions. Thanks for looking into it though

pveber:
Errr, sorry, I've just noticed something weird, the release seems to be called "OCaml-R" instead of "v0.5.0", including in Github's release page. Maybe the syntax in my CHANGES file is not compatible with `dune-release` or something, but I should change that. If you're ok with it, I'm going to close this PR and re-submit.

mseri:
Indeed the version is wrong. You can also force push to this PR, as you prefer. Either way isne

mseri:
Looks good now, thanks

pveber:
Great, thanks (again) for your patience!

PR #19928, opened 2021-11-01
Tezos-stdlib-unix needs fmt (used to be transitively provided)

pirbo:
We forgot to declare the dependency toward `fmt` of `tezos-stdlib-unix` in the dune file (but not in the opam file!) because up to `domain-name.0.3.0`, `domain-name` brought `fmt` and we were saved by the dependency chain `domain-name` -> `ipaddr` -> `tezos-stdlib-unix`...

Of course, there is an alternative fix which is to declare `"domain-name" { < "0.3.1" }` as a dependency constraint but is it really less ugly then the patch?

### DIFF
--- a/packages/tezos-stdlib-unix/tezos-stdlib-unix.10.2/opam
+++ b/packages/tezos-stdlib-unix/tezos-stdlib-unix.10.2/opam
@@ -13,6 +13,7 @@ depends: [
   "ptime" { >= "0.8.4" }
   "mtime" { >= "1.0.0" }
   "conf-libev"
+  "domain-name" { < "0.3.1" }
   "ipaddr" { >= "4.0.0" }
   "ezjsonm" { >= "1.1.0" }
   "fmt" { >= "0.8.7" }


mseri:
Can you move the latch to version 10.2-1 and add the upper bound to version 10.2?

pirbo:
Is `10.2-1` = `10.2` for opam? Else, not really because we declare dependencies in between tezos packages with ` { = version }` so I would need to do make a """release""" 10.2-1 of all tezos packages....

Actually, there is a 10.3 that I didn't package here by lack of time, let's just add the constraint for 10.2 and I'll put the patch in 10.3 when I have a minute to push it here.

mseri:
> Is `10.2-1` = `10.2` for opam? Else, not really because we declare dependencies in between tezos packages with ` { = version }` so I would need to do make a """release""" 10.2-1 of all tezos packages....

Oh I see the problem, they are seen as different afaik, which would then be an issue. @kit-ty-kate am I correct?

> Actually, there is a 10.3 that I didn't package here by lack of time, let's just add the constraint for 10.2 and I'll put the patch in 10.3 when I have a minute to push it here.

Sounds great, thanks!

kit-ty-kate:
Thanks

kit-ty-kate:
> Oh I see the problem, they are seen as different afaik, which would then be an issue. @kit-ty-kate am I correct?

yeah, hopefully this should be fixed by some sort of combination of https://github.com/ocaml/opam/issues/4566 and https://github.com/ocaml/opam/pull/4532

PR #19938, opened 2021-11-02
[new release] prc (0.1.0)

pveber:
Utilities for precision-recall curves

- Project page: <a href="https://github.com/pveber/prc/">https://github.com/pveber/prc/</a>

##### CHANGES:

- documentation improvements
- added bootstrap procedure to compute confidence intervals
- added demo code to check the procedures perform ok.

### DIFF
--- a/packages/prc/prc.0.1.0/opam
+++ b/packages/prc/prc.0.1.0/opam
@@ -1,0 +1,48 @@
+opam-version: "2.0"
+synopsis: "Utilities for precision-recall curves"
+description: """
+
+This library provides functions to compute precision-recall curves, as
+well as several methods to compute their AUC (area under curve)
+"""
+maintainer: ["philippe.veber@gmail.com"]
+authors: ["Philippe Veber"]
+license: "CeCILL-B"
+tags: ["precision" "recall" "statistics" "machine-learning"]
+homepage: "https://github.com/pveber/prc/"
+bug-reports: "https://github.com/pveber/prc/issues"
+depends: [
+  "alcotest" {with-test}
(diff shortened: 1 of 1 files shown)


pveber:
I reapplied a few changes that we made for v0.0.0, not sure how to keep them accross versions...

mseri:
Disabling the tests needs to be done by hand afaik and the dev thing is fixed in dune 2.7 (so need to update the version-also in the dune-project file-) but we can do it manually when releasing in general

mseri:
The lower bound failure is due to incompatibility between gsl and the installed system gsl library, Indon’t think we should to do anything for that

packages/prc/prc.0.1.0/opam
```diff
@@ -0,0 +1,48 @@
...
+  "core_kernel" {>= "v0.12.0"}
+  "dune" {>= "2.4"}
+  "gsl"
+  "ppx_jane"
+  "ppx_deriving" {with-test}
+  "sklearn" {with-test}
+]
+depopts: [
+  "ocaml-r" {>= "v0.5.0"}
+]
```
kit-ty-kate, line 25:
```suggestion
depopts: [
  "ocaml-r"
]
conflicts: [
  "ocaml-r" {< "0.5.0"}
]
```
Unless the package has some kind of weird setup conflict should almost always be used. Also there was a v prefix which made the version never attainable.

kit-ty-kate:
Thanks!

kit-ty-kate:
Could you return the fixes in the opam file upstream?

pveber:
Done in pveber/prc@f2b218f
Thanks a lot you both for the careful review!

PR #19940, opened 2021-11-03
[new release] gospel (0.1.0)

pascutto:
Gospel - A tool-agnostic formal specification language for OCaml

- Project page: https://ocaml-gospel.github.io/gospel/


kit-ty-kate:

pascutto:
> Thanks! Could you return the fix in the opam file upstream?

Sure, thanks a lot!

PR #19952, opened 2021-11-05
[new release] dune-release (1.5.2)

Leonidas-from-XIV:
Release dune packages in opam

- Project page: <a href="https://github.com/ocamllabs/dune-release">https://github.com/ocamllabs/dune-release</a>

### Fixed

- Fixed the release asset URL for projects with multiple opam packages. Before,
  the packages would attempt to infer their URL and fail in rare cases where
  the project uses `v` as prefix for tags but the project version omits it. Now
  they share the same URL. (ocamllabs/dune-release#402, ocamllabs/dune-release#404, @Leonidas-from-XIV)


packages/dune-release/dune-release.1.5.2/opam
```diff
@@ -0,0 +1,61 @@
...
+  "dune" {>= "2.7"}
+  "ocaml" {>= "4.06.0"}
+  "curly"
+  "fmt" {>= "0.8.7"}
+  "fpath" {>= "0.7.3"}
+  "bos"
+  "cmdliner"
+  "re" {>= "1.7.2"}
+  "astring"
+  "opam-format" {>= "2.1.0~beta"}
```
mseri, line 29:
```suggestion
  "opam-file-format" {>= "2.1.1"}
  "opam-format" {>= "2.1.0~beta"}
```

uses OpamParserTypes.FullPos

mseri: Is this the right thing to do, or we should instead add
```
conflicts: [
  "opam-file-format" {< "2.1.1"}
]
```
Ping @kit-ty-kate

Leonidas-from-XIV: I created [this PR](https://github.com/ocamllabs/dune-release/pull/412) upstream, and I think given we use `OpamParserTypes` directly, we should declare a dependency on `opam-file-format.2.1.2` (I think it was added in 2.1.2).

mseri: 2.1.1 seems fine, at least the lower bound check is not failing now.
The remaining failures are due to the CI environment for certain distributions, e.g.
```
$ git commit -m "" --allow-empty-message --quiet
# +  error: Terminal is dumb, but EDITOR unset
# +  Please supply the message using either -m or -F option.
# +  [1]
```

Leonidas-from-XIV: Ah yes, I assume that's because `OpamParserTypes` is now coming from... somewhere else. `opam-format` probably.

Looks like the git version in CentOS is so old, that it doesn't accept empty commit messages. I've [made a PR upstream](https://github.com/ocamllabs/dune-release/pull/413) but given we don't test on CentOS 7 upstream there is the question on whether supporting this release is worth the effort.

mseri:
Looks fine to me. Can you have a brief look at the error in https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/e384af9797bb7153323acb4782d0b34960be0f7e/variant/opam-2.0,distributions,centos-7,dune-release.1.5.2,tests ? I think it is due to an old version of git or something similar, and should not block the release, but better to double check

Leonidas-from-XIV:
@mseri Yes, this is due to CentOS 7 and OracleLinux 7 having old git versions (like really old, `1.8.x.y` era) and fixing up some of the more obvious tests only helped a little, only to run into the next issue. It is probably not worth making sure the tests works on such old versions (the tool itself probably still does run just fine, it is just the tests that expect very specific output) therefore it is better to just ignore test failures on these two distributions.

We do run tests on CentOS 8 and there the tests work, so this shouldn't in practice be a problematic situation for users.

mseri:
Thanks

PR #19953, opened 2021-11-05
Octez 11rc2

pirbo:
A new major release is around the corner and once again we check on an rc that we didn't messed up with packaging...


kit-ty-kate:
Some kind of upper-bound constraint is probably missing on older tezos packages:
```

kit-ty-kate:
(pushing fixes here directly to clear the job queue and allow the other PRs be able to be tested again and not be blocked by this one)

pirbo:
Now #20044

PR #19957, opened 2021-11-05
OCaml 4.14: dev packages

Octachron:
This PR adds the development package for the 4.14 branch of the compiler, update the trunk package to 5.0.0+trunk and add a new ocaml-5.0.0 metapackage as a dependency on the new trunk package.

### DIFF
--- a/packages/ocaml-variants/ocaml-variants.5.0.0+trunk/opam
+++ b/packages/ocaml-variants/ocaml-variants.5.0.0+trunk/opam
@@ -1,13 +1,13 @@
 opam-version: "2.0"
 license: "LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception"
-synopsis: "Latest 4.14.0 development"
+synopsis: "Current trunk"
 maintainer: "platform@lists.ocaml.org"
 authors: ["Xavier Leroy" "Damien Doligez" "Alain Frisch" "Jacques Garrigue" "Didier Rémy" "Jérôme Vouillon"]
 homepage: "https://ocaml.org"
 bug-reports: "https://github.com/ocaml/opam-repository/issues"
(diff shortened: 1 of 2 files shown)


packages/ocaml-variants/ocaml-variants.5.0.0+trunk/opam
```diff
@@ -0,0 +1,79 @@
...
+  "ocaml-option-bytecode-only"
+  "ocaml-option-default-unsafe-string"
+  "ocaml-option-no-flat-float-array"
+  "ocaml-option-flambda"
+  "ocaml-option-fp"
+  "ocaml-option-musl"
+  "ocaml-option-static"
+  "ocaml-option-spacetime"
+  "ocaml-option-nnp"
+  "ocaml-option-nnpchecker"
```
kit-ty-kate, line 78:
```suggestion
  "ocaml-option-32bit"
  "ocaml-option-afl"
  "ocaml-option-bytecode-only"
  "ocaml-option-default-unsafe-string"
  "ocaml-option-no-flat-float-array"
  "ocaml-option-flambda"
  "ocaml-option-fp"
  "ocaml-option-musl"
  "ocaml-option-static"
  "ocaml-option-spacetime"
  "ocaml-option-nnp"
  "ocaml-option-nnpchecker"
```
There are probably several things that might become straight dependencies (e.g. nnp) or conflicts (32bit) in the future. I don’t know if it’s worth doing it now, but I suppose this is fine for now.

kit-ty-kate:
Thanks!

kit-ty-kate:
cc @dra27 regarding https://github.com/ocaml/ocaml/pull/10757. I missed the missing 00. If it is accepted could you also rename it here?

PR #19959, opened 2021-11-06
OCaml 5.00.0 rather than 5.0.0

dra27:
Apropos https://github.com/ocaml/ocaml/pull/10757 and https://github.com/ocaml/opam-repository/pull/19957#issuecomment-962501629.


dra27:
If merged, as this only affects trunk releases, I'm guessing we'd live with a rename on this occasion (especially as the decision will probably happen within the next 48 hours)

kit-ty-kate:
https://github.com/ocaml/ocaml/pull/10757 has been merged. Let’s merge this one too. Thanks!

PR #19962, opened 2021-11-08
Remove ppx_deriving_qcheck and ppx_pbt

vch9:
Hi,

I'm removing `ppx_deriving_qcheck` as the development is now part of [QCheck](https://github.com/c-cube/qcheck) and this early release can lead to confusion in potential users.

I'm also removing `ppx_pbt` which was an experimental work that needs `ppx_deriving_qcheck`. However, as I would not have time to improve `ppx_pbt`, I'd rather remove it and come back with a better version in the future :)


kit-ty-kate:
Straight up removing packages is against our policy (rare exceptions apply) as it would break users looking for those packages or who have lock files requiring them.

Would it be ok to instead print a warnings to users who are installing it like "This package is deprecated, please use qcheck >= <version> instead" ?

kit-ty-kate:
Another suggestion on top of it from @dra27 is: we could also restrict the packages to ocaml < 4.14 as an artificial constraint so they are not available for future users.

vch9:
Sorry about the delay and thank you for your answer !

> Would it be ok to instead print a warnings to users who are installing it like "This package is deprecated, please use qcheck >= instead" ?

The issue is that I've made this early release but there is not yet a "stable" release to propose to users. Therefore, I can not yet say "please use ... instead"

> Another suggestion on top of it from @dra27 is: we could also restrict the packages to ocaml < 4.14 as an artificial constraint so they are not available for future users.

This seem reasonable based on what I said above.

mseri:
I have impemented the suggestions here in https://github.com/ocaml/opam-repository/pull/20094. I am closing this in favour of the other PR.
Please @vch9 let me know if you agree with the other PR, and I'll merge

PR #19963, opened 2021-11-08
Update coccinelle 1.1.0 checksums

evdenis:
For some reason checksums changed. I created a ticket for this:
https://github.com/coccinelle/coccinelle/issues/268

But I think it's preferable to update checksums anyway.

Signed-off-by: Denis Efremov <efremov@linux.com>

### DIFF
--- a/packages/coccinelle/coccinelle.1.1.0/opam
+++ b/packages/coccinelle/coccinelle.1.1.0/opam
@@ -16,7 +16,7 @@ install: [
 ]
 depends: [
   "ocaml"
-  "menhir"
+  "menhir" {>= "20181113"}
   "ocamlfind"
   "pcre"
   "stdcompat"
@@ -26,9 +26,10 @@ depends: [
   "conf-python-3-dev"
   "conf-aclocal"
   "conf-autoconf"
-  "parmap"
+  "parmap" {>= "1.0-rc9"}
   "num"
 ]
+available: os != "macos"
 synopsis: "Coccinelle is a C source code matching and transformation engine"
 description: """
 Coccinelle provides the language SmPL (Semantic Patch Language) for specifying
--- a/packages/coccinelle/coccinelle.1.1.0/opam
+++ b/packages/coccinelle/coccinelle.1.1.0/opam
@@ -36,7 +36,7 @@ desired matches and transformations in C code."""
 url {
   src: "https://github.com/coccinelle/coccinelle/archive/1.1.0.tar.gz"
   checksum: [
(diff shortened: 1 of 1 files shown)


packages/coccinelle/coccinelle.1.1.0/opam
```diff
@@ -26,17 +26,18 @@ depends: [
   "conf-python-3-dev"
   "conf-aclocal"
   "conf-autoconf"
-  "parmap"
+  "parmap" {>= "1.0-rc9"}
   "num"
 ]
+available: os != "macos"
```
kit-ty-kate, line 32: For posterity could you describe why this constraint has been added?

evdenis: I added this to pass the CI tests, I don't think that coccinelle officially supported on macos (at least I didn't find mentions about it). If it's too strict, I'll drop it.

evdenis:
I'm not sure about `opensuse-15.3` test. Seems like external dependency problem.

packages/coccinelle/coccinelle.1.1.0/opam
mseri: This seems in line with what was done in the 1.1.1 release: https://github.com/ocaml/opam-repository/pull/19762

mseri: I also tested it locally and can confirm that it does not compile on macos:
```
.../.opam/4.13.1/bin/ocamlc.opt -unsafe -I parmap  -ccopt -D_GNU_SOURCE -ccopt -fPIC parmap/setcore_stubs.c -o parmap/setcore_stubs.o
parmap/setcore_stubs.c:47:17: error: implicit declaration of function 'thread_policy_set' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
      retcode = thread_policy_set(mach_thread_self(),
                ^
1 error generated.
make[2]: *** [parmap/setcore_stubs.o] Error 2
make[1]: *** [all] Error 2
make: *** [.../code/ocaml/coccinelle/bundles/parmap/parmap.cmxa] Error 2
```

mseri: Incidentally, I can build parmap on mac, but it is probably a more recent version than the one vendored by coccinelle

mseri:
Yes, opensuse is an independent failure

mseri:
I could not find the package in the cache, so I am going to accept the change since the maintainers are aware and happy with it

PR #19964, opened 2021-11-08
Package ocplib-endian.1.2

chambart:
### `ocplib-endian.1.2`
Optimised functions to read and write int16/32/64 from strings and bigarrays
The library implements three modules:
* [EndianString](https://github.com/OCamlPro/ocplib-endian/blob/master/src/endianString.mli) works directly on strings, and provides submodules BigEndian and LittleEndian, with their unsafe counter-parts;
* [EndianBytes](https://github.com/OCamlPro/ocplib-endian/blob/master/src/endianBytes.mli) works directly on bytes, and provides submodules BigEndian and LittleEndian, with their unsafe counter-parts;
* [EndianBigstring](https://github.com/OCamlPro/ocplib-endian/blob/master/src/endianBigstring.mli) works on bigstrings (Bigarrays of chars), and provides submodules BigEndian and LittleEndian, with their unsafe counter-parts.

---
* Homepage: https://github.com/OCamlPro/ocplib-endian
* Source repo: git+https://github.com/OCamlPro/ocplib-endian.git
* Bug tracker: https://github.com/OCamlPro/ocplib-endian/issues

---

### DIFF
--- a/packages/ocplib-endian/ocplib-endian.1.2/opam
+++ b/packages/ocplib-endian/ocplib-endian.1.2/opam
@@ -1,0 +1,39 @@
+opam-version: "2.0"
+synopsis:
+  "Optimised functions to read and write int16/32/64 from strings and bigarrays"
+description: """\
+The library implements three modules:
+* [EndianString](https://github.com/OCamlPro/ocplib-endian/blob/master/src/endianString.mli) works directly on strings, and provides submodules BigEndian and LittleEndian, with their unsafe counter-parts;
+* [EndianBytes](https://github.com/OCamlPro/ocplib-endian/blob/master/src/endianBytes.mli) works directly on bytes, and provides submodules BigEndian and LittleEndian, with their unsafe counter-parts;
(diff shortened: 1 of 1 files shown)


packages/ocplib-endian/ocplib-endian.1.2/opam
```diff
@@ -0,0 +1,34 @@
...
+]
+build: [
+  ["dune" "build" "-p" name "-j" jobs
+   "@install"
+   "@runtest" {with-test}
+   "@doc" {with-doc}]
+]
+dev-repo: "git+https://github.com/OCamlPro/ocplib-endian.git"
+url {
+  src: "https://github.com/OCamlPro/ocplib-endian/archive/1.1.tar.gz"
```
kit-ty-kate, line 29:
```suggestion
  src: "https://github.com/OCamlPro/ocplib-endian/archive/1.2.tar.gz"
```
Shouldn’t this be 1.2 instead? However the 1.2 tag seems to be missing

chambart: Good catch. I didn't even look at it, it was produced by opam-publish so I had supposed that it would be ok. I probably forgot to push the tag. opam-publish took the version from the local git for some things and from the remote for the archive !
@AltGr Is that a bug ?

mseri:
Looks good. Can you please add the license field also upstream? (no need to re-release)

PR #19965, opened 2021-11-08
[new release] redirect.0.2.1

thierry-martinez:
This PR adds the new release for redirect.0.2.1:

- Fix `with_channel_from_string` for long strings that do not fit in
  pipe, by using a thread for outputing the string.

- Some minor optimizations, using the fact that re-raising an exception
  preserves the backtrace and the constraint on OCaml version (>=4.03)
  ensures that match-with-exception is available.


kit-ty-kate:
Thanks!

PR #19971, opened 2021-11-09
[NEW RELEASE] DAGaml.0.02

JoanThibault:
- **NEW** RBTF, pRBTF, CRBTF, RBTF+max/int
- Upgraded interfaces : OOPS ( cmp, cofactor, id ) and QEOPS ( support )
- **NEW** new module : Support ( interfaced with QEOPS )
- Imported `RBTF_ArgMax_vUInt` from `Snowflake`
- Improved performances of LDD-O-NUCX
- **NEW** Added AQEOPS' library, including : `sat_select_A`, `exists_projA`, `argmaxA`
- **FORMAT** Fixed 'Modele' into 'Model' (spellcheck)

### DIFF
--- a/packages/DAGaml/DAGaml.0.02/opam
+++ b/packages/DAGaml/DAGaml.0.02/opam
@@ -1,0 +1,21 @@
+opam-version: "2.0"
+synopsis: "DAGaml : Abstract DAG manipulation in OCaml"
+maintainer: "Joan Thibault <joan.thibault@ens-rennes.fr>"
+authors: "Joan Thibault <joan.thibault@ens-rennes.fr>"
+license: "LGPL-3.0-only with OCaml-LGPL-linking-exception"
+homepage: "https://gitlab.com/boreal-ldd/dagaml"
+bug-reports: "https://gitlab.com/boreal-ldd/dagaml"
+depends: [
+	"ocaml" {>= "4.08.1" & < "4.12"}
+	"GuaCaml" {>= "0.03"}
+	"Snowflake" {>= "0.02.02"}
+	"ocamlbuild" {build}
+	"ocamlfind" {build}
+]
+build: [make "lib"]
+install: [make "install"]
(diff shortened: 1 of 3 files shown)


packages/DAGaml/DAGaml.0.02/opam
```diff
@@ -0,0 +1,21 @@
+opam-version: "2.0"
+synopsis: "DAGaml : Abstract DAG manipulation in OCaml"
+maintainer: "Joan Thibault <joan.thibault@ens-rennes.fr>"
+authors: "Joan Thibault <joan.thibault@ens-rennes.fr>"
+license: "LGPL-3.0-only with OCaml-LGPL-linking-exception"
+homepage: "https://gitlab.com/boreal-ldd/dagaml"
+bug-reports: "https://gitlab.com/boreal-ldd/dagaml"
+depends: [
+	"ocaml" {>= "4.08.1" & < "4.12"}
```
kit-ty-kate, line 9: Just curious, what is the reason 4.12 and up aren’t supported?

kit-ty-kate: Testing it locally in 4.13 I’m not seeing any issues
```suggestion
	"ocaml" {>= "4.08.1"}
```

JoanThibault: Actually, there is a linking-time error with `GuaCaml.0.03` due to the lack of some hash-related function starting from ocaml 4.12.
So, I added a patch to GuaCaml ( `GuaCaml.0.03.01` ) and fixed DAGaml's dependencies accordingly. =)

JoanThibault:
With regards to the compilation error, it is true that DAGaml only support 64bits architecture (only because of low-level implementation details, its not a global requirement) would it be possible to specify so in the opam file ?

mseri:
Thanks a lot

PR #19986, opened 2021-11-11
Tweak depext.1.2.0 metadata

dra27:
I didn't review the opam-depext.1.2.0 PR in time to make these as comments there. The development mode of opam-depext only needs to exist in the upstream repo - there's no value in the pinning instructions being in opam-repository (cf. the previous versions).

I tweaked slightly the post-message - at a major version bump of opam the plugin will finally not be supported, but the plugin is not is not itself _deprecated_, it's just _unnecessary_.


kit-ty-kate:
Thanks

PR #19990, opened 2021-11-12
Support installing gappa on Apple Silicon with Homebrew

k4rtik:
This change is needed for Coq Platform to enable support for Apple's arm64 / M1 chips. See: https://github.com/coq/platform/issues/168

### DIFF
--- a/packages/gappa/gappa.1.4.0/opam
+++ b/packages/gappa/gappa.1.4.0/opam
@@ -15,6 +15,8 @@ build: [
   [ "./configure"
     # If someone knows how to ask MacPorts for the usual include and lib path, please tell me
     "CXXFLAGS=-I/opt/local/include" "LDFLAGS=-L/opt/local/lib" { os-distribution = "macports" & os = "macos" }
+    # Support installing on Apple Silicon with Homebrew
+    "CXXFLAGS=-I/opt/homebrew/include" "LDFLAGS=-L/opt/homebrew/lib" { os-distribution = "homebrew" & os = "macos" }
     "--build=%{arch}%-pc-cygwin" { os = "win32" & os-distribution = "cygwinports" }
     "--host=%{arch}%-w64-mingw32" { os = "win32" & os-distribution = "cygwinports" }
     "--target=%{arch}%-w64-mingw32" { os = "win32" & os-distribution = "cygwinports" }


packages/gappa/gappa.1.4.0/opam
```diff
@@ -15,6 +15,8 @@ build: [
   [ "./configure"
     # If someone knows how to ask MacPorts for the usual include and lib path, please tell me
     "CXXFLAGS=-I/opt/local/include" "LDFLAGS=-L/opt/local/lib" { os-distribution = "macports" & os = "macos" }
+    # Support installing on Apple Silicon with Homebrew
+    "CXXFLAGS=-I/opt/homebrew/include" "LDFLAGS=-L/opt/homebrew/lib" { os-distribution = "homebrew" & os = "macos" }
```
kit-ty-kate, line 19:
```suggestion
    "CXXFLAGS=-I/opt/local/include" "LDFLAGS=-L/opt/local/lib" { os-distribution = "macports" & os = "macos" & arch = "x86_64" }
    # Support installing on Apple Silicon with Homebrew
    "CXXFLAGS=-I/opt/homebrew/include" "LDFLAGS=-L/opt/homebrew/lib" { os-distribution = "homebrew" & os = "macos" & arch = "arm64" }
```
Otherwise it both would always be set and I’m not sure both would be concatenated. I think they would be always overwritten

k4rtik: Thanks for the suggestion. I was under the impression that `os-distribution` value might distinguish these, but arch is certainly helpful and descriptive.

kit-ty-kate: oh.. Nevermind I read it wrong. I thought it was homebrew twice. Could you remove the last commit?

k4rtik: Ah, done.

mseri: Wait, I was too fast at reading the change.
Wouldn't this make `"CXXFLAGS=-I/opt/local/include"` always set and always be overridden by `"CXXFLAGS=-I/opt/local/include"`? Am I mistaken that the filter only applies to the LDFLAGS string?

MSoegtropIMC:
Is `/opt/homebrew` the standard path of homebrew? it used to install things under `/opt/local`, but I am a MacPorts user, so I am not up do date with Homebrew.

mseri:
Looks like that is correct but only for arm: https://docs.brew.sh/FAQ#why-is-the-default-installation-prefix-opthomebrew-on-apple-silicon

MSoegtropIMC:
Thanks -looks good to me then (I am the maintainer of this package).

mseri:
I have opened a separate PR because the way it is now would fail on x86 homebrew (and I think the flags are passed incorrectly): https://github.com/ocaml/opam-repository/pull/19993

PR #19993, opened 2021-11-12
gappa 1.4.0: fix os-dependent flags

mseri:
The filter was only applied to a part of the flags.
See #19990

### DIFF
--- a/packages/gappa/gappa.1.4.0/opam
+++ b/packages/gappa/gappa.1.4.0/opam
@@ -16,9 +16,12 @@ build: [
     # If someone knows how to ask MacPorts for the usual include and lib path, please tell me
     "CXXFLAGS=-I/opt/local/include" { os-distribution = "macports" & os = "macos" }
     "LDFLAGS=-L/opt/local/lib" { os-distribution = "macports" & os = "macos" }
+    # Support installing on x86 with Homebrew
+    "CXXFLAGS=-I/usr/local/include" { os-distribution = "homebrew" & os = "macos" & arch = "x86_64"}
+    "LDFLAGS=-L/usr/local/lib" { os-distribution = "homebrew" & os = "macos" & arch = "x86_64"}
     # Support installing on Apple Silicon with Homebrew
-    "CXXFLAGS=-I/opt/homebrew/include" { os-distribution = "homebrew" & os = "macos" }
-    "LDFLAGS=-L/opt/homebrew/lib" { os-distribution = "homebrew" & os = "macos" }
+    "CXXFLAGS=-I/opt/homebrew/include" { os-distribution = "homebrew" & os = "macos" & arch = "arm64"}
+    "LDFLAGS=-L/opt/homebrew/lib" { os-distribution = "homebrew" & os = "macos" & arch = "arm64"}
     # Support for windows
     "--build=%{arch}%-pc-cygwin" { os = "win32" & os-distribution = "cygwinports" }
     "--host=%{arch}%-w64-mingw32" { os = "win32" & os-distribution = "cygwinports" }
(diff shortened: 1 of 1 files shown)


mseri:
cc @kit-ty-kate @k4rtik

mseri:
cc @MSoegtropIMC

k4rtik:
I do not think this change is needed. The `xxxFLAGS` user variables only add to the locations that the compiler or linker already search for. Wouldn't the CI run fail for my PR #19990 on macOS with homebrew if the FLAGS were being overridden? 

I see https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/5984852b39b053472b17e1797058cab535d26bf2 that they had passed.

k4rtik:
The only issue that I see is instead of CXXFLAGS (C++ compiler flags), we should be passing CPPFLAGS (C/C++ Preprocessor flags) to include additional locations, but I followed the example of Macports in the same file to select CXXFLAGS.

mseri:
Without the filter the extra flags would override the macports flag, that I am reasonably confident is a bug.

It is maybe overkill to add the x86_64 flags if the build already searches in the /usr/local folder, but I think it is harmless to add.
This raises another question, though: what happens on an M1 mac with both rosetta and arm homebrew installed? Does it actually work?

k4rtik:
> This raises another question, though: what happens on an M1 mac with both rosetta and arm homebrew installed? Does it actually work?

This is worth testing for sure.

In my understanding most users (and I have been an occasional one for the past 11 months or so using a friend's MB Air) tend to use arm homebrew by default and invoke rosetta version explicitly when needed, but indeed if gappa finds the rosetta version of the library first, then it may choke. On my current M1 Pro machine, I am avoiding installing the Rosetta homebrew to see what software doesn't work at all.

k4rtik:
Perhaps [this doc](https://gcc.gnu.org/onlinedocs/gcc-4.9.3/cpp/Invocation.html#Invocation) has the answer:

> -I dir
Add the directory _dir_ to the list of directories to be searched for header files. See Search Path. Directories named by -I are searched before the standard system include directories. If the directory dir is a standard system include directory, the option is ignored to ensure that the default search order for system directories and the special treatment of system headers are not defeated

k4rtik:
Great. I will let @MSoegtropIMC chip in once as the maintainer, but in my understanding this change is indeed not needed.

MSoegtropIMC:
I agree that "/usr/local/include" or "/usr/local/lib" should not be added - on macOS these are system search folders:
```
$ clang++ --version
Apple clang version 13.0.0 (clang-1300.0.29.3)
Target: x86_64-apple-darwin20.6.0

$clang++ -E -x c++ -v /dev/null

#include <...> search starts here:
 /usr/local/include
 /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1
 /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/13.0.0/include
 /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include
 /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include
 /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks (framework directory)
End of search list.
```
And I agree that for the includes it would be cleaner to use CPPFLAGS instead of CXXFLAGS, but it wouldn't make a difference here since it only uses C++ and doesn't call the preprocessor explicitly.

MSoegtropIMC:
@mseri : thanks for spotting the issues with the conditions!

MSoegtropIMC:
Just a note: the currently active version of the gappa package fails on MacPorts.

MSoegtropIMC:
I would guess the CI failures on Alpine where always there, so I wouldn't take that as a reason to block on MacPorts. We should create an issue for Alpine.

mseri:
I use macports, I will test it as as on as I have a minute

MSoegtropIMC:
I just tested it locally and your latest commit works fine for me.

mseri:
Is working fine for me as well

PR #19996, opened 2021-11-12
[new release] builder-web (0.1.0)

hannesm:
Web interface for builder

- Project page: <a href="https://github.com/roburio/builder-web">https://github.com/roburio/builder-web</a>

##### CHANGES:

* Initial public release

PR #19998, opened 2021-11-12
metrics-lwt, metrics-unix: metrics {= version}

reynir:
Discovered this missing version constraint in CI lower-bounds check for #19996.

### DIFF
--- a/packages/metrics-lwt/metrics-lwt.0.1.0/opam
+++ b/packages/metrics-lwt/metrics-lwt.0.1.0/opam
@@ -15,7 +15,7 @@ build: [
 
 depends: [
   "dune"
-  "metrics"
+  "metrics" {= version}
   "lwt"
 ]
 synopsis: "Lwt backend for the Metrics library"
--- a/packages/metrics-lwt/metrics-lwt.0.1.0/opam
+++ b/packages/metrics-lwt/metrics-lwt.0.1.0/opam
@@ -14,7 +14,7 @@ build: [
 ]
 
 depends: [
-  "dune"
+  "dune" {>= "1.1"}
   "metrics" {= version}
   "lwt"
 ]
--- a/packages/metrics-lwt/metrics-lwt.0.1.0/opam
+++ b/packages/metrics-lwt/metrics-lwt.0.1.0/opam
@@ -10,12 +10,12 @@ doc:          "https://mirage.github.io/metrics/"
 build: [
   ["dune" "subst"] {dev}
   ["dune" "build" "-p" name "-j" jobs]
-  ["dune" "runtest" "-p" name] {with-test}
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
 ]
 
 depends: [
   "dune" {>= "1.4"}
-  "metrics" {= version}
+  "metrics" {< "0.5"} # should have been "= version" but to avoid breaking lock files and stuff..
   "lwt"
 ]
 synopsis: "Lwt backend for the Metrics library"
--- a/packages/metrics-lwt/metrics-lwt.0.1.0/opam
+++ b/packages/metrics-lwt/metrics-lwt.0.1.0/opam
@@ -14,7 +14,7 @@ build: [
 ]
 
 depends: [
-  "dune" {>= "1.1"}
+  "dune" {>= "1.4"}
   "metrics" {= version}
   "lwt"
 ]
--- a/packages/metrics-lwt/metrics-lwt.0.1.0/opam
+++ b/packages/metrics-lwt/metrics-lwt.0.1.0/opam
@@ -16,7 +16,7 @@ build: [
 depends: [
   "dune" {>= "1.4"}
   "metrics" {< "0.5"} # should have been "= version" but to avoid breaking lock files and stuff..
-  "lwt"
+  "lwt" {>= "2.4.7"}
 ]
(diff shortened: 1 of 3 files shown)


packages/metrics-unix/metrics-unix.0.1.0/opam
```diff
@@ -16,8 +16,8 @@ build: [
 depends: [
   "dune"
```
hannesm, line 17:
```suggestion
  "dune" {>= "1.1"}
```

packages/metrics-lwt/metrics-lwt.0.1.0/opam
```diff
@@ -15,7 +15,7 @@ build: [
 
 depends: [
   "dune"
```
hannesm, line 17:
```suggestion
  "dune" {>= "1.1"}
```

reynir:
Thanks!

hannesm:
the remaining CI failures are due to "lower-bounds (failed: Unknown extension "fmt".)" -- I was not able to figure out from the documentation https://dune.readthedocs.io/en/stable/formatting.html#using-fmt-1-0 or CHANGES.md in dune since when that `using fmt 1.0` is supported. It looks like the `lang dune 1.1` (in the dune-project) leads to dune (recent versions) not complain about the fmt.

dinosaure:
That's weird, we should make an issue on `dune` about that. I would like a double check from someone else but it seems fine to merge this PR.

reynir:
I grepped for `using fmt` in the dune git history, and found a commit that introduced test cases for this: https://github.com/ocaml/dune/commit/928024b8137032cc112649743bed44008b06c64f

It also added a line in CHANGES.md for 1.4.0 : https://github.com/ocaml/dune/blob/main/CHANGES.md#140-10102018

> Integration with automatic formatters (#1252, fix #1201, @emillon)

emillon:
Yes, bumping to `(lang dune 1.1)` is the right solution.

There is a check that `(using)` is consistent with `(lang dune)` but it has been added later (around dune 1.4 IIRC). So there is no check in the lower-bounds run.

packages/metrics-lwt/metrics-lwt.0.1.0/opam
```diff
@@ -14,8 +14,8 @@ build: [
 ]
 
 depends: [
-  "dune"
-  "metrics"
+  "dune" {>= "1.4"}
+  "metrics" {= version}
```
kit-ty-kate, line 18: I’m a bit thorn on this:
- on one hand this makes a lot of sense to keep it synched with metrics
- but on the other, they currently compile fine even with metrics 0.4.0 and i wouldn’t want to break existing users (e.g. lock files with both metrics-lwt.0.1.0 and metrics.0.2.0)

In practice i don’t think it matters too much for this one but i still feel conflicted.

mseri: Should we just do 
```suggestion
  "metrics" {< "0.5"}
```
Then we would not break the current compatibility but ensure the syncing from 0.5 onwards

packages/metrics-lwt/metrics-lwt.0.1.0/opam
```diff
@@ -10,12 +10,12 @@ doc:          "https://mirage.github.io/metrics/"
...
-  ["dune" "runtest" "-p" name] {with-test}
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
 ]
 
 depends: [
-  "dune"
-  "metrics"
+  "dune" {>= "1.4"}
+  "metrics" {< "0.5"} # should have been "= version" but to avoid breaking lock files and stuff..
   "lwt"
```
mseri, line 19:
```suggestion
  "lwt" {>= "2.4.7"}
```

Uses Lwt.Infix

packages/metrics-unix/metrics-unix.0.1.0/opam
```diff
@@ -10,14 +10,15 @@ doc:          "https://mirage.github.io/metrics/"
...
   ["dune" "build" "-p" name "-j" jobs]
-  ["dune" "runtest" "-p" name] {with-test}
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
 ]
 
 depends: [
-  "dune"
+  "dune" {>= "1.4"}
+  "fmt" {>= "0.8.5"}
   "uuidm"
```
mseri, line 19:
```suggestion
  "uuidm" {>= "0.9.6"}
```

Uses Uuidm.v

packages/metrics-unix/metrics-unix.0.1.0/opam
```diff
@@ -10,14 +10,15 @@ doc:          "https://mirage.github.io/metrics/"
...
-  "dune"
-  "uuidm"
-  "metrics"
-  "mtime"
+  "dune" {>= "1.4"}
+  "fmt" {>= "0.8.5"}
+  "uuidm" {>= "0.9.6"}
+  "metrics" {< "0.5"} # should have been "= version" but to avoid breaking lock files and stuff..
+  "mtime" {>= "1.0.0"}
   "lwt"
```
hannesm, line 22:
```suggestion
  "lwt" {>= "2.4.7"}
```
uses Lwt.Infix

mseri:
🎉

PR #20001, opened 2021-11-12
[new release] pyre-ast (0.1.6)

grievejia:
Full-fidelity Python parser in OCaml

- Project page: <a href="https://github.com/grievejia/pyre-ast">https://github.com/grievejia/pyre-ast</a>
- Documentation: <a href="https://grievejia.github.io/pyre-ast/doc">https://grievejia.github.io/pyre-ast/doc</a>

##### CHANGES:

- Expose 2 additional fields `end_lineno` and `end_offset` from CPython3.10 `SyntaxError`. Fix an error in documentation where column numbers should start from 1 instead of 0.
- Remove the optional `filename` argument from the `parse_module` API. It turns out that this argument is actually dropped when computing error messages so it does not serve any purpose at the moment.

### DIFF
--- a/packages/pyre-ast/pyre-ast.0.1.6/opam
+++ b/packages/pyre-ast/pyre-ast.0.1.6/opam
@@ -1,0 +1,47 @@
+opam-version: "2.0"
+synopsis: "Full-fidelity Python parser in OCaml"
+description:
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Thanks!

PR #20004, opened 2021-11-13
Fix gappa support on Apple Silicon for 1.3.5 as well

k4rtik:
gappa-1.3.5 is included in __coq-platform.2021.09.0~8.12 and fails with the same problem as #19990


kit-ty-kate:
Thanks

PR #20008, opened 2021-11-13
bistro: add missing dep on ppxlib

pveber:
pveber/bistro#47 mentionned there's an incompatibility with version 0.23.0 of ppxlib, which made me realize that the dep on ppxlib was missing.


mseri:
Thanks

PR #20026, opened 2021-11-15
Package ocamlog.0.1

paulpatault:
### `ocamlog.0.1`
Simple Logger for OCaml
Able to print with different colors, levels, trace out caller, etc.

---
* Homepage: https://www.github.com/paulpatault/OcamLog
* Source repo: git+https://github.com/paulpatault/OcamLog
* Bug tracker: https://www.github.com/paulpatault/OcamLog/issues

---
:camel: Pull-request generated by opam-publish v2.1.0

PR #20037, opened 2021-11-16
Package ocamlog.0.11

paulpatault:
### `ocamlog.0.11`
Simple Logger for OCaml
Able to print with different colors, levels, trace out caller, etc.

---
* Homepage: https://www.github.com/paulpatault/OcamLog
* Source repo: git+https://github.com/paulpatault/OcamLog
* Bug tracker: https://www.github.com/paulpatault/OcamLog/issues

---

### DIFF
--- a/packages/ocamlog/ocamlog.0.11/opam
+++ b/packages/ocamlog/ocamlog.0.11/opam
@@ -1,0 +1,22 @@
+
+opam-version: "2.0"
+synopsis: "Simple Logger for OCaml"
+description: "Able to print with different colors, levels, trace out caller, etc."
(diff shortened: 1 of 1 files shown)


dinosaure:
According to #20026, what is the diff between this release and the older one? Should we tag this release as `0.2` instead of `0.11`?

paulpatault:
Yes I should have change the tag to `0.2` instead of `0.11`, my bad.
The diff is just a small change in the Ocamlog code.

mseri:
@paulpatault can you rename the tag and the package to 0.2 or are you happy releasing 0.11 instead? In that case, the next update will have to be > 0.13

PR #20039, opened 2021-11-16
js_of_ocaml: conflict with base-domains

avsm:
Currently the multicore compilers have a different bytecode
format which results in an exception:
Js_of_ocaml_compiler.Instr.Bad_instruction(152))

This PR prevents the existing jsoo being selected as an installable
option with the multicore compiler and erroring out on CI.

### DIFF
--- a/packages/js_of_ocaml-compiler/js_of_ocaml-compiler.3.0.1/opam
+++ b/packages/js_of_ocaml-compiler/js_of_ocaml-compiler.3.0.1/opam
@@ -19,6 +19,7 @@ depends: [
 conflicts: [
   "ocamlfind"   {< "1.5.1"}
   "js_of_ocaml" {< "3.0"}
+  "base-domains"
 ]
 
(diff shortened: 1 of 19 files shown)


kit-ty-kate:
I believe that the correct fix is to add the conflict to the `js_of_ocaml-compiler` package not `js_of_ocaml` (which only includes the library)

mseri:
I think we want to prevent the use of jsoo completely until it actually knows about the new bytecode. Can we actually use jsoo in a meaningful way without also using the compiler? (Genuinely asking, I don't know)

kit-ty-kate:
jsoo requires its compiler to compile. As far as I know `js_of_ocaml` itself only implements the javascript API (DOM manipulation, …), `js_of_ocaml-compiler` handles all the compilation from OCaml so it’s this part that in incompatible not the javascript API

kit-ty-kate:
Thanks

PR #20040, opened 2021-11-16
Fix conf-python-2-7-1.1 for Fedora

fangyi-zhou:
The package name for Fedora should be `python2.7`, see https://fedoraproject.org/wiki/Changes/RetirePython2


mseri:
Can you move the change to a new version `conf-python-2-7-1.2/opam`, so that users of older releases can still pin this version and ensure it works?

mseri:
Thanks a lot

PR #20036, opened 2021-11-16
Tezos mockup missing version constraint

pirbo:

PR #20044, opened 2021-11-17
Tezos node 11.0

pirbo:
A new major release of Octez (spitted in 2 to be nicer with the CI).

Includes #20036


samoht:
The tests do not pass on 32-bits but otherwise the rest of the errors are due to the CI system (homebrew and oraclelinux).

Screenshot 2021-11-22 at 16 04 30

This is good to merge.

PR #20051, opened 2021-11-17
Coq.8.13.0/1 disable warn 70

clarus:
Follow up of https://github.com/ocaml/opam-repository/pull/20025

Removes one warning from Coq to fix many Coq plugins with OCaml 4.13.


clarus:
@ejgallego if you want to have a look. I think we should be good after that. I tested that both Coq 8.13.0 and 8.13.1 can install `coq-bignums` with this change.

ejgallego (approved):

mseri:
Thanks

PR #20065, opened 2021-11-19
Add Z3 4.8.13 and 4.8.14 with dynamic loading of libraries

k4rtik:
See https://github.com/Z3Prover/z3/issues/5578#issuecomment-947082861 for the decision to split Z3 into two packages. Currently, on macOS static linking (the default before this change) does not work. This PR makes dynamic linking the default and adds another package for statically linked variant of the package.

CC'ing: @arbipher who implemented the change to review this PR.


k4rtik:
So the failures seem to be for `z3-static` on macOS where it is already expected to fail.

packages/z3-static/z3-static.4.8.13/opam
```diff
@@ -0,0 +1,37 @@
...
+  "ocamlfind" {build}
+  "zarith"
+  "conf-gmp"
+  "conf-python-2-7" {build}
+]
+x-ci-accept-failures: [
+  "centos-7"  "oraclelinux-7" # C compiler is too old
+  "oraclelinux-8" # Does not ship python 2.7
+]
+synopsis: "Z3 solver"
```
mseri, line 29:
```suggestion
available: os != "macos"
synopsis: "Z3 solver"
```

k4rtik: Thanks, that's a better solution than what I was trying before.

k4rtik:
It seems now the cancelled tests are all timeouts.

packages/z3/z3.4.8.13/opam
```diff
@@ -0,0 +1,37 @@
...
+bug-reports: "https://github.com/Z3prover/z3/issues"
+license: "MIT"
+dev-repo: "git+https://github.com/Z3prover/z3.git"
+build: [
+  [ "python2.7" "scripts/mk_make.py" "--ml"]
+  [ make "-C" "build" "-j" jobs ]
+]
+
+install: [
+  [ "sh" "-c" "ocamlfind install z3 build/api/ml/* build/libz3.*"]
```
arbipher, line 14: `[ "sh" "-c" "ocamlfind install z3 build/api/ml/* -dll build/libz3.*"]`

Here needs a `-dll` to guarantee either `libz3.so` or `libz3.dylib` to be installed into `<opam-sitelib>/stublibs`.

arbipher:
The opam files look good to me.

btw, the opam-ci for z3 packages tests the installation but no post-installation checking yet. My plan was also to test the opam installation locally on simple programs to use the packages. It should be work without any problems, but just to prevent z3 makes some changes between the time when I fixed the previous problem and when this new version was released.

Would you also help to check it? I may not be free today to do the testing on both ubuntu and macos.

packages/z3/z3.4.8.13/opam
k4rtik: Thanks for taking a look. I see in the man page of `ocamlfind install` that:

> If there are files ending in the suffixes .so or .dll, the package directory will be added to the DLL configuration file ld.conf, such that the dynamic loader can find the DLL. [...]
> 
> However, if there is a stublibs directory in site-lib, the DLLs are not installed in the package directory, but in this directory that is shared by all packages that are installed at the same location. [...]
> 
> The options -dll and -nodll can be used to control exactly which files are considered as DLLs and which not. By default, the mentioned suffix rule is in effect: files ending in ".so" (Unix) or ".dll" (Windows) are DLLs. The switch -dll changes this, and all following files are considered as DLLs, regardless of their suffix. The switch  -nodll expresses that the following files are not DLLs, even if they have a DLL-like suffix.

It seems it should do the right thing if we do not specify the `-dll` option. Do you still think we need it?

arbipher: I think it's the right thing to do with the `-dll`.
There is some inconsistency between z3, opamfind, and macos.
z3 itself will make `libz3.so` on ubuntu and `libz3.dylib` on macos.
z3's ocamlbinding, which uses `ocamlmklib`, will make `dllz3ml.so` on both platforms.

We need (it makes things much simpler) to install `dllz3ml.so` and `libz3.{so|dylib}` into the `stublibs` explicitly.

I don't think `ocamlfind` handles it correctly for macos. I made this [PR](https://github.com/ocaml/ocamlfind/pull/25) there but no response yet. We need to guard it here on our own.

As for macos, it's more complex and subtle. (Please forgive my laziness to not post the reference here, but TLDR;) both `.so` and `.dylib` can be used.

kit-ty-kate:
cc @c-cube

packages/z3/z3.4.8.13/opam
k4rtik: Thanks, @arbipher, I just tried running a simple test after adding the -dll option to the opam file. It works fine on macOS but on Ubuntu, it still fails with:

```
File "Z3Example.ml", line 1:
Error: Error on dynamically loaded library: /home/kartik/.opam/default/lib/stublibs/dllz3ml.so: libz3.so: cannot open shared object file: No such file or directory
```

even though the file exists in the stublibs directory:
```
~/.opam/default/lib/stublibs$ ls
dllz3ml.so  dllz3ml.so.owner  dllzarith.so  dllzarith.so.owner  libz3.so  libz3.so.owner
```

It is as if the search path is different on linux than on macOS.

k4rtik:
Never mind [my comment above](https://github.com/ocaml/opam-repository/pull/20065#discussion_r753694381). Turns out on my linux machine, I had a system ocaml installed as well, so the ld.conf at the system level was taking precedence over the one installed opam under ~/.opam.

packages/z3-static/z3-static.4.8.13/opam
```diff
@@ -0,0 +1,38 @@
...
+license: "MIT"
+dev-repo: "git+https://github.com/Z3prover/z3.git"
+build: [
+  [ "python2.7" "scripts/mk_make.py" "--ml" "--staticlib" ]
+  [ make "-C" "build" "-j" jobs ]
+]
+
+install: [
+  [ "sh" "-c" "ocamlfind install z3 build/api/ml/* build/libz3-static.a"]
+  [ "cp" "build/z3" "%{bin}%/z3-static"]
```
k4rtik, line 15: Now, the final question with this PR is whether the name of the binary for statically-linked installation of z3 should be called `z3-static` or just `z3`? Note that the package is still called z3 here.

cc: @arbipher @c-cube

mseri: Should there be a mutual conflict between the packages?

arbipher: oh, @k4rtik you are right. they should be `z3` and `z3-static`. This was also discussed in that same [z3 issue](https://github.com/Z3Prover/z3/issues/5578).

@mseri these two packages are consistent and they can be both installed on ubuntu. `z3` and `z3-static` relies on no system-level z3, both are built from the source and there should be no name conflict based on my last fix.

k4rtik: I guess I am confused about whether the package name in `opam list` should be `z3` or `z3-static` after doing the static install. In my testing, it came out to be `z3` but that may have been because of the way I am doing testing (manually downloading the tarball, copying the opam file in the extracted directory whose name is `z3.4.8.13` and then issuing `opam install ./opam`). But if this is not what we want, then do the users need to use the statically linked z3 using a different command:

```
ocamlfind ocamlc -o ml_example.byte -thread -package z3-static -linkpkg ml_example.ml
```
instead of the [usual](https://github.com/Z3Prover/z3/tree/master/examples/ml):
```
ocamlfind ocamlc -o ml_example.byte -thread -package z3 -linkpkg ml_example.ml
```
If this is the case, we need to document it properly.

Can someone else verify the static install on Linux and report back?

arbipher: There will be two distinct packages `z3` and `z3-static`.

`opam list` should have `z3` if you install `z3`, `z3-static` if you install `z3-static`. I guess your manual installation may leave some files in the opam site path.

This is my testing [file](https://github.com/arbipher/ocaml-build-examples/blob/main/Makefile), which tests the combination of using `{z3,z3-static}` * `{bytecode, native, bytecode-with-custom}`. Every use of the z3 static should use `z3-static`. `z3` and `z3-static` shares nothing.

btw, after changing line 14 to 
`[ "sh" "-c" "ocamlfind install z3-static build/api/ml/* build/libz3-static.a"]`,
it could pass all my tests with ocaml 4.13.1 on my ubuntu.

arbipher:
Besides testing `z3-static` on ubuntu, I also tested using `z3 4.8.13` (the shared library version) with ocaml 4.13.1 on ubuntu and with ocaml 4.12.0 on macos. They both work for me.

arbipher:
I have a question for your opam maintainers

The [root case](https://discuss.ocaml.org/t/my-fix-of-the-linking-problem-on-z3-ocaml-binding/8691/3?u=arbipher) of the problem on using z3-static on macos is c++ ABI mismatch. However, this problem doesn't always appear and it seems to depend on which z3 api are used. (disclaimer: I cannot use it)

Is it possible to bypass `os != "macos"` if the _advanced_ users want to use it? What is the opam's philosophy for partial working packages?

k4rtik:
Alright, thanks for sharing the tests, @arbipher. I feel more comfortable about these opam files now after testing the installs using the method provided at https://coq.inria.fr/opam-packaging.html under **Test your new package**. Everything seems to pass correctly on Ubuntu as well.

Now this PR only waits for a final review from maintainers and also perhaps answer to your question above.

packages/z3-static/z3-static.4.8.13/opam
```diff
@@ -0,0 +1,38 @@
+opam-version: "2.0"
+maintainer: "simon@aestheticintegration.com"
+authors: "MSR"
+homepage: "https://github.com/Z3prover/z3"
+bug-reports: "https://github.com/Z3prover/z3/issues"
+license: "MIT"
+dev-repo: "git+https://github.com/Z3prover/z3.git"
+build: [
+  [ "python2.7" "scripts/mk_make.py" "--ml" "--staticlib" ]
```
copy, line 9: Could this be changed to python3 (and `conf-python-2-7` to `conf-python-3` below)?

arbipher: I think it would be fine.

I test this update locally (my opam [change](https://github.com/arbipher/opam-repository/commit/58dbab6ccf0ecfb446d20326adecf5fd6a8a3c4a)) and redo all my previous tests. All pass.

c-cube:
Feel more than free to change the 'maintainer' field, btw.

k4rtik:
> Feel more than free to change the 'maintainer' field, btw.

Thanks, but not planning to. :) One of my projects is kinda stuck because of lack of Z3 installation on macOS, hence putting in this much effort. (Of course, the major work was done by @arbipher earlier.) But if anyone else wants to take maintainership, please feel free to let me know, and I can update the field.

arbipher:
I can be one maintainer in maybe my phd time.
My work email is weng@cs.jhu.edu.

k4rtik:
So with the switch to Python 3, the CI fails on [all debian-based distributions](https://askubuntu.com/a/1260519) because "distutils.sysconfig" is required but not installed by default in those distros.

I don't know an easy way to solve this issue.

`conf-python-3` should perhaps ensure this, [but doesn't](https://github.com/ocaml/opam-repository/blob/master/packages/conf-python-3/conf-python-3.1.0.0/opam). @kit-ty-kate @mseri any suggestions?

k4rtik:
Alright, the CI seems happy. Ready for merge from my end. @kit-ty-kate 

(what a whirlwind of issues with this PR)

MSoegtropIMC:
My 5 cents: for me the static link version works just fine on macOS. What is the specific issue you have?

I see this a bit critical, e.g. I wanted to add a package for the tptp front end variant of z3 and this change would make this quite a bit more complicated because I might need to select the dependency depending on the OS.

I would suggest we try to find out what the issue with static linking on macOS is. "Doesn't work" is IMHO not an acceptable statement.

Also I think the package should simply install both, the static and the dynamic variant - then everybody can choose what is more appropriate.

k4rtik:
Hi @MSoegtropIMC, several people [including me](https://github.com/Z3Prover/z3/issues/5517#issuecomment-926600948) had reported seeing the following error on macOS as reported on the Z3 repo. See the linked issue which has the most detail: https://github.com/Z3Prover/z3/issues/5578

> libc++abi: terminating with uncaught exception of type is_non_propositional_predicate::found

What version of Z3 and macOS you are on?

arbipher:
@MSoegtropIMC 
I use both incremental solver in z3 and its datatype APIs. On macos, one of these features uses exceptions that should be caught (as an implementation technique) but it incorrectly [raises to users](https://github.com/Z3Prover/z3/issues/5578#issuecomment-931571446). It is the start point of these issues.

It's OK to me to allow users to use the static version on mac on their own choice instead of disabling it at all. You can see my question on _partial working pakcage_ in the above reply. I don't think it will be easy to fix it in a short time. Though I have some guessing on the reason, it takes time to take experiments and may be related to the z3 source code.

> Also I think the package should simply install both, the static and the dynamic variant - then everybody can choose what is more appropriate.

It's the same freedom to me whether `one z3 package install both shared and static binding`  or `two packages z3 and z3-static with its own binding`. The current choice (the latter) makes the installing and building easier.

MSoegtropIMC:
@arbipher : my argument against separate packages for static and dynamic linking is that opam packages are usually logical entities and not technical entities. Z3 is Z3, independent of the link method. Also such details tend to be heavily system dependent and IMHO wan doesn't want to move such system dependent properties to the package level.

So yes, technical two separate packages are equivalent to one package, but it doesn't fit well opam's philosophy.

@k4rtik : I am currently using Z3 4.8.11 on BigSur (Intel) with MacPorts.

One thought: did someone try to use a different compiler? I guess the XCode C compiler was used. One could check if the issue goes away with gcc. It wouldn't be difficult to tell opam to use gcc on macOS.

MSoegtropIMC:
@k4rtik : the error `ld: warning: directory not found for option '-L/opt/local/lib'` looks like a solveable configuration problem, although I must say I find the python based configuration of Z3 difficult, because it is non standard.

The problem described by @arbipher looks more serious, but it looks like a problem which might go away with a different compiler. It doesn't look like a problem which depends on static vs dynamic linking. Does it?

Also is this a problem in the ML bindings or on the C++ level?

arbipher:
> One thought: did someone try to use a different compiler? I guess the XCode C compiler was used. One could check if the issue goes away with gcc. It wouldn't be difficult to tell opam to use gcc on macOS.

In macos, gcc is an alias of clang. It may not be easy to require gcc installed (then users will have a fake-gcc-which-is-clang and a true-gcc). (See the link of root cause above). I don't know if we shall blame which part but at least they(Z3 and ocaml) don't align well.

MSoegtropIMC:
@arbipher : It still might make a difference if you have XCode clang or another clang. With MacPorts this should be possible and I would think with homebrew too.

Also did you try to debug this? I don't think we can find a solution without knowing what the root cause is.

Anyway: can you explain what this has to do with the static vs. dynamic change? I find it rather unfortunate to discuss two apparently unrelated problems in one issue.

arbipher:
I have xcode clang and other clang. In my vague memory of reading the installation log, both z3 and its ocaml bindings use the xcode clang (system level alias of gcc).

I don't know the ultimate **reason** why static binding doesn't work on macos for me and @k4rtik. However, the behavior **fact** is it doesn't work. We have to use z3 4.8.1 for a long time. There are a few z3 issues on this problem. If we revisit the history of z3 binding, in z3 4.8.1, [all](https://github.com/ocaml/opam-repository/blob/e3825c34dc7243090d6a911821b9681418e601b0/packages/z3/z3.4.8.1/opam?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L10) uses the dynamic bindings. In the next release 4.8.4, another severe bug is fix but the default [changes](https://github.com/ocaml/opam-repository/blob/e3825c34dc7243090d6a911821b9681418e601b0/packages/z3/z3.4.8.4/opam?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L14) to the static bindings. The current latest before this release on macos, the dynamic binding has errors on linking and the static binding doesn't work on us.

I tried but am not able to debug and fix this root cause yet. This release is mainly for solidating the fix of the dynamic link on mac, to at least allow some mac users (who have to use 4.8.1 till now) to use the latest version of z3.

arbipher:
As for the naming, would `z3` and `z3.static` be better than `z3` and `z3-static`? Given some other packages also use `blabla` `blabla.toplevel` `blabla.ppx`.

c-cube:
You cannot name an opam package `z3.static`. It's a thing on the ocamlfind level (`z3.static` would be a sub-library in the package `z3` typically). I think `z3` and `z3-static` are adequate for separate opam packages.

MSoegtropIMC:
@arbipher : I read your write up in https://discuss.ocaml.org/t/my-fix-of-the-linking-problem-on-z3-ocaml-binding/8691/3?u=arbipher.

A few notes: I don't have issues with finding shared libraries, but I link dependencies to shared libraries to the **installed** shared library. I think the problem with the z3 ocaml package is that it builds the z3 shared library and the ML wrapper for z3 shared library in one go, so that the latter remembers the build location rather than the install location of the shared library. So a better split (instead of static and dynamic) should be to split it into C++ z3 and the z3 ML binding. Then when the z3 ML binding is built, it finds the z3 library in its final installation location. The z3 C++ package should simply install both, a static and a dynamic library.

Regarding the ABI issues: I don't experience these either, but of course have different uses cases. But IMHO this is something which should be root cause analyzed and fixed properly. This static / dynamic think is IMHO a hack which is likely to produce more issues down the road than it fixes.

arbipher:
@MSoegtropIMC 
> I don't have issues with finding shared libraries, but I link dependencies to shared libraries to the installed shared library. 
This was a problem on macos before this fix (release). If you didn't suffer from this, the reason may be you have another global z3 installed out of opam. The dynamic binding of z3 itself on macos was incorrect (as I said in the writeup). 

> So a better split (instead of static and dynamic) should be to split it into C++ z3 and the z3 ML binding.
I see. You mean we can have packages `conf-z3` which installs `z3` and `z3` which installs the ocaml binding. This sounds good to me. I know `llvm` and `clangml` uses the similar approach. As for z3, we merely follow the previous practice and this may not be optimal.

However, even with this, the static binding on mac is still a problem on macos. I don't fully understand your advice on what I shall do next. macos users like me want any working binding (whether dynamic or static). If the users can used static binding before, they can also use it now (after removing `os != "macos"`). After split the installation of `z3` lib and `z3` ocaml, it will still be necessary if we can have to choice to use which binding before the bug is fixed.

Would you mind if I make another post on the forum for our discussion? We can have more points and ideas on people interested in the topic but may miss this issue.

MSoegtropIMC:
@arbipher : Let me try your specific example given here (https://github.com/Z3Prover/z3/issues/5578) tomorrow on my setup (I am also on Mac (BigSur / MacPorts / Intel). I can also test it on freshly installed MacOS with Homebrew.

I just have a feeling that you try to fix something which is an effect of your very specific setup and that the fix you propose causes more trouble than good for mostly everybody else.

E.g. what you describe can easily happen by changing the default C++ compiler between setting up the opam switch and compiling Z3.

arbipher:
@MSoegtropIMC 
I have some more [tests](https://github.com/JHU-PL-Lab/ddse/blob/test-generation-dbmc/test/sudu/test_sudu_z3.ml) but it is not in a standalone repo yet. You can run test in the `test/sudu` path if you would like to use it.

> I just have a feeling that you try to fix something which is an effect of your very specific setup and that the fix you propose causes more trouble than good for most everybody else.

I am not sure about this. The dynamic binding on macos was definitely fixed. As for the static binding, it's not only me having this problem. It will be nice if it could be solved. However, I don't think providing two bindings is overkill but maybe there are better ways to arrange it which we don't have a conclusion yet.

MSoegtropIMC:
Thanks for test tests! I will see under which conditions I can reproduce your issues.

MSoegtropIMC:
@arbipher : I had a look at this and can reproduce this issue in clean environments.

To get forward let me explain why I am against a split. The main issue I see is that the two packages won't be compatible, because they will partly install the same file (I guess the z3 executable e.g.). People not on MacOS will likely randomly choose one or the other package for z3. This way we will quickly end up with a situation where people create packages which are not compatible with macOS. Worse is that we will have packages which cannot be installed together, because one uses the static and one the non static version, and these can't be installed together. I maintain a largish opam meta project (Coq Platform) and I have a lot of work with such issues. People think they make things flexible by offering variants, but by this they in the end reduce flexibility cause they introduce unnecessary incompatibility points.

So if, as it looks, on Mac one needs dynamic binding just create one package which does whatever it did before on other platforms and uses shared libraries on MacOS. Or use shared libraries everywhere (I would think platforms which don't support shared libraries are extinct for > 30 years).

I looked a bit into this but didn't find something obvious. Apparently on macOS a static version of the C/C++ library does not exist (as does on Windows), so that one cannot easily get this messed up (as on Windows). See e.g. https://developer.apple.com/library/archive/qa/qa1118/_index.html. So I think one can assume that this is indeed a bug in /usr/lib/libc++.1.dylib or /usr/lib/libc++abi.dylib.

k4rtik:
It seems like this change should be enough and I can delete `z3-static` file in this PR:

```
--- a/packages/z3/z3.4.8.13/opam
+++ b/packages/z3/z3.4.8.13/opam
@@ -9,12 +9,14 @@ depexts: [
   ["python3-distutils"] {os-family = "debian"}
 ]
 build: [
-  [ "python3" "scripts/mk_make.py" "--ml"]
+  [ "python3" "scripts/mk_make.py" "--ml" "--staticlib" ] {os != "macos"}
+  [ "python3" "scripts/mk_make.py" "--ml" ]                {os = "macos"}
   [ make "-C" "build" "-j" jobs ]
 ]

 install: [
-  [ "sh" "-c" "ocamlfind install z3 build/api/ml/* -dll build/libz3.*"]
+  [ "sh" "-c" "ocamlfind install z3 build/api/ml/* build/libz3.a"]     {os != "macos"}
+  [ "sh" "-c" "ocamlfind install z3 build/api/ml/* -dll build/libz3.*"] {os = "macos"}
   [ "cp" "build/z3" "%{bin}%/z3"]
 ]
```

How does that sound?

Let me push it to test over all environments (it works fine on my local macOS).

MSoegtropIMC:
@k4rtik : I would recommend to use dynamic on all platforms because it is the default of Z3 (static needs an extra option), it might complicate the design of platform independent dependent packages if this is different and I think for a large library like z3 dynamic linking is anyway the first choice option and one shouldn't use static unless there is a good reason. I would see if this breaks anything - one could check the opam repo for dependencies.

I will check the currently one z3 dependencies I have in Coq Platform.

k4rtik:
Sounds fair and we already tested that the dynamic package works on all platforms in previous iterations of this PR.

I think it will be worth getting input from @NikolajBjorner and perhaps [OCaml community](https://discuss.ocaml.org/t/my-fix-of-the-linking-problem-on-z3-ocaml-binding/8691/3?u=arbipher) as @arbipher mentioned:

> Would you mind if I make another post on the forum for our discussion? We can have more points and ideas on people interested in the topic but may miss this issue.

@arbipher if you have not already, could you help move this decision forward with input from the forum?

arbipher:
> To get forward let me explain why I am against a split. The main issue I see is that the two packages won't be compatible, because they will partly install the same file (I guess the z3 executable e.g.). People not on MacOS will likely randomly choose one or the other package for z3. This way we will quickly end up with a situation where people create packages which are not compatible with macOS. Worse is that we will have packages which cannot be installed together, because one uses the static and one the non static version, and these can't be installed together. I maintain a largish opam meta project (Coq Platform) and I have a lot of work with such issues. People think they make things flexible by offering variants, but by this they in the end reduce flexibility cause they introduce unnecessary incompatibility points.
> 
> So if, as it looks, on Mac one needs dynamic binding just create one package which does whatever it did before on other platforms and uses shared libraries on MacOS. Or use shared libraries everywhere (I would think platforms which don't support shared libraries are extinct for > 30 years).

@MSoegtropIMC 
I appreciate this argument. If not providing static library could be a choice, I am with it.

> @arbipher : I had a look at this and can reproduce this issue in clean environments.
> 
> I looked a bit into this but didn't find something obvious. Apparently on macOS a static version of the C/C++ library does not exist (as does on Windows), so that one cannot easily get this messed up (as on Windows). See e.g. https://developer.apple.com/library/archive/qa/qa1118/_index.html. So I think one can assume that this is indeed a bug in /usr/lib/libc++.1.dylib or /usr/lib/libc++abi.dylib.

Yep, there is no obvious reason or fix on the problem. Thanks for your time for this. (I am willing to spend some time to finally solve it but not at this moment.)

@k4rtik 
I am not done with the posting yet. I will do it today.

arbipher:
I just make this [post](https://discuss.ocaml.org/t/ask-for-suggestions-for-the-next-z3-package-release/8922) on the ocaml forum. I don't do another editing round with you before posting it (for saving time), so please reply to the post with your suggestions and ideas.

MSoegtropIMC:
> I just make this post on the ocaml forum. I don't do another editing round with you before posting it (for saving time), so please reply to the post with the correction, suggestions and ideas.

Thanks! I would say we wait two days. I take the package as is now as patch into Coq Platform (it has its own opam patch repo) and do tests and run the Coq Platform CI with it. If all goes well and no one complains in the forum we go with the the shared library only solution.

k4rtik:
Hi, @MSoegtropIMC did you get a chance to test the dynamically-linked version of Z3 (as the current PR) provides? If nothing is broken, we could perhaps get it merged.

MSoegtropIMC:
Sorry was busy this week. I will check it over the weekend.

MSoegtropIMC:
What about the CI status?

k4rtik:
I believe most tests are passing, some are canceled because of timeouts and a couple have some internal error/powered off status. I do not think they are caused by this PR.

@kit-ty-kate does that sound right?

MSoegtropIMC:
Well in my experience opam tests usually all pass, and for this PR CI did look much better before (just macos static was failing). It might make sense to just restart CI. I don't think it will be merged if CI looks like this.

kit-ty-kate:
It looks like the tests fail due to them taking too much RAM:
```
# virtual memory exhausted: Cannot allocate memory
# make: *** [Makefile:3565: muz/base/dl_context.o] Error 1
```

kit-ty-kate:
ah nevermind there is no tests... mh that’s a weird one

k4rtik:
Alright, all tests are a-go!

I like that Opam CI lets one just hit on the rebuild button for specific tests, instead of needing to rerun the whole CI again. I had to do that for some of the canceled tests.

MSoegtropIMC:
So just my tests missing ... on the way

MSoegtropIMC:
@k4rtik : sorry I was busy with other things, but think I can do the z3 testing until tomorrow.

I had a look at
```
opam-coq-archive$ opam list --depends-on z3
# Packages matching: (installed | available) & depends-on(z3)
# Name                       # Installed # Synopsis
asli                         --          Interpreter for Arm's Architecture Specification Language (ASL)
bap-primus-symbolic-executor --          Primus Symbolic Executor
fstar                        --          Verification system for effectful programs
```

Did you check if these packages still work with your change? I think this should be tested. Afaik Opam CI doesn't check if dependencies still build.

k4rtik:
@MSoegtropIMC `$ opam list --depends-on z3` does not show `fstar` on my system. To make sure I was using the most recent version of opam-repository, I merged upstream into my fork (without realizing that it will restart the CI), but even then `fstar` does not appear in the list that depends on z3.

In any case, with this z3, `asli.0.2.0` and `bap-primus-symbolic-executor.2.3.0` install just fine on my system and AFAIK [`fstar` uses a custom version of z3](https://github.com/FStarLang/FStar/blob/master/INSTALL.md#runtime-dependency-particular-version-of-z3), so we should be good.

mseri:
This is interesting since the CI does run the revdeps, and in this case picks
```
/home/opam: (run (shell "echo '@@@OUTPUT' && opam list -s --color=never --depends-on 'z3.4.8.13' --coinstallable-with 'z3.4.8.13' --installable --all-versions --recursive --depopts && opam list -s --color=never --depends-on 'z3.4.8.13' --coinstallable-with 'z3.4.8.13' --installable --all-versions --with-test --depopts && echo '@@@OUTPUT'"))
@@@OUTPUT
asli.0.2.0
bap-primus-symbolic-executor.2.2.0
bap-primus-symbolic-executor.2.3.0
opam-client.2.1.0~beta2
opam-client.2.1.0~beta4
opam-client.2.1.0~rc2
opam-client.2.1.0
opam-client.2.1.1
opam-custom-install.0.1
opam-devel.2.1.0~beta2
opam-devel.2.1.0~beta4
opam-devel.2.1.0~rc2
opam-devel.2.1.0
opam-devel.2.1.1
opam-lock.0.1
opam-lock.0.2
opam-solver.2.1.0~beta2
opam-solver.2.1.0~beta4
opam-solver.2.1.0~rc2
opam-solver.2.1.0
opam-solver.2.1.1
z3.4.8.13
asli.0.2.0
bap-primus-symbolic-executor.2.2.0
bap-primus-symbolic-executor.2.3.0
nuscr.2.0.0
opam-solver.2.1.0~beta2
opam-solver.2.1.0~beta4
opam-solver.2.1.0~rc2
opam-solver.2.1.0
opam-solver.2.1.1
@@@OUTPUT
```
But the test does then not seem to be run!

mseri:
In the previous commit they were all succeeding. Look down under compilers -> 4.13 -> revdeps in 
https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/161e47619e167b37f885a5a462128cba5bc92e22

mseri:
> @MSoegtropIMC `$ opam list --depends-on z3` does not show `fstar` on my system. To make sure I was using the most recent version of opam-repository, I merged upstream into my fork (without realizing that it will restart the CI), but even then `fstar` does not appear in the list that depends on z3.

Are you using ocaml 4.13? fstar is not compatible with 4.12 and would not appear in that case

kit-ty-kate:
Given this library switched to dynamic loading by default, we should probably constrain all the packages depending on it as it changes the expected behaviour. For example for `opam-solver` and `opam-devel` it is expected that the user can just take the opam binary generated and remove any opam installation afterward, but doing that would result in a runtime failure.

k4rtik:
> Are you using ocaml 4.13? fstar is not compatible with 4.12 and would not appear in that case

I was, I have three switches installed, one on 4.13 and two on 4.10.2, it does not appear in any:

```
~/Hub/opam-repository master ❯ opam switch                                                                                                         07:52:57
#  switch                                 compiler                    description
   __coq-platform.2021.11.0~8.13~2021.11  ocaml-base-compiler.4.10.2  __coq-platform.2021.11.0~8.13~2021.11
   __coq-platform.2021.11.0~8.14~2021.11  ocaml-base-compiler.4.10.2  __coq-platform.2021.11.0~8.14~2021.11
→  default                                ocaml.4.13.1                default

[WARNING] The environment is not in sync with the current switch.
          You should run: eval $(opam env)
~/Hub/opam-repository master ❯ eval $(opam env)                                                                                                    08:09:06
~/Hub/opam-repository master ❯ opam switch __coq-platform.2021.11.0~8.14~2021.11                                                                   08:09:08
~/Hub/opam-repository master ❯ eval $(opam env)                                                                                                    08:09:22
~/Hub/opam-repository master ❯ opam list --depends-on z3                                                                                           08:09:24
# Packages matching: (installed | available) & depends-on(z3)
# Name                       # Installed # Synopsis
asli                         --          Interpreter for Arm's Architecture Specification Language (ASL)
bap-primus-symbolic-executor --          Primus Symbolic Executor
~/Hub/opam-repository master ❯ opam switch __coq-platform.2021.11.0~8.13~2021.11                                                                   08:09:29
~/Hub/opam-repository master ❯ eval $(opam env)                                                                                                    08:09:48
~/Hub/opam-repository master ❯ opam list --depends-on z3                                                                                           08:09:50
# Packages matching: (installed | available) & depends-on(z3)
# Name                       # Installed # Synopsis
asli                         --          Interpreter for Arm's Architecture Specification Language (ASL)
bap-primus-symbolic-executor --          Primus Symbolic Executor
```

Do you need me to test with 4.11 manually?

k4rtik:
Here's the constraint on z3 from the most recent `fstar` package here: https://github.com/ocaml/opam-repository/blob/b0c660f8e5a8712b84c8d1802c1e742b4358cecf/packages/fstar/fstar.2021.06.06/opam#L28

I don't think we need to worry about `fstar` here, but I don't know what needs to be done for what @kit-ty-kate said above.

MSoegtropIMC:
> @MSoegtropIMC `$ opam list --depends-on z3` does not show `fstar` on my system. 

Hmm - what does `opam show fstar` say? I get `"z3" {= "4.8.5"}` . Possibly it depends on the opam version if the dependencies show up if you have a different version installed (I have 4.8.11).

MSoegtropIMC:
@kit-ty-kate : so are you saying that opam has a strong use case for a statically linked Z3 - and maybe this is the reason why the Z3 package was statically linked in the first place?

MSoegtropIMC:
@kit-ty-kate : for my understanding: so opam uses opam to build opam?

k4rtik:
> maybe this is the reason why the Z3 package was statically linked in the first place?

The original PR that switched Z3 to static build in v4.8.4 is https://github.com/ocaml/opam-repository/pull/14173 I am unsure if that should affect opam directly.

k4rtik:
> maybe this is the reason why the Z3 package was statically linked in the first place?

The original PR that switched Z3 to static build in v4.8.4 is https://github.com/ocaml/opam-repository/pull/14173 I am unsure if that should affect opam directly.

AltGr:
It hasn't been in a release yet, but indeed for opam we're interested in having a static z3 lib; for now I've been meddling with it with manually tweaked buids of z3.
Another related issue is that, before #14173, the dynamic lib wouldn't work without manually setting `LD_LIBRARY_PATH` ; at least on Linux.

But I don't think this release will break anything important; [generating static builds](https://www.ocamlpro.com/blog/2021_09_02_generating_static_and_portable_executables_with_ocaml) requires several layers of hacks anyway :)

Maybe the question we should ask is: do we want to accept dynamic deps to stuff that remains within an opam switch ? Dynamic deps are normal at the moment, but I think all of them are expected to be present on the system, outside of opam's realm.

All this said, if this makes the lives of most Z3 users easier, I have no objection to merging!

MSoegtropIMC:
IMHO one should try to live with the dynamic libraries. It usually requires a bit of overhead, but I don't think it makes sense to statically link several libs of the complexity of z3. And then there are the reported problems ...

I would more think about if opam env shouldn't setup the environment accordingly in case the current switch contains shared library packages.

arbipher:
@MSoegtropIMC 
> I would more think about if opam env shouldn't setup the environment accordingly in case the current switch contains shared library packages.

May you explain the _opam env_ thing in more details. I don't get it.

k4rtik:
BTW, [the last test](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/d5c0b60f205505d7cd148f1ad8084c2039eeb947/variant/opam-2.1,extras,nnpchecker,z3.4.8.13,tests) on the CI is stuck at this message since morning:
`2021-12-06 14:45.27: Cancelling: Timeout (60.0 minutes)`

Unsure, why it isn't actually canceled yet.

k4rtik:
> May you explain the _opam env_ thing in more details.

I believe Michael was referring to the need for setting `LD_LIBRARY_PATH` manually as reported in the previous comment. Ideally `opam env` should take care of that.

MSoegtropIMC:
> I believe Michael was referring to the need for setting LD_LIBRARY_PATH manually as reported in the previous comment. 

Yes, that's what I meant. I don't see a good reason for not doing this.

Of course a use of Z3 by opam itself - in its solver - would be a different story, but as was mentioned, one might have special purpose packages for this.

arbipher:
I got your point. I am also curious on the opam's idea on this.

One problem here is on macos both `LD_LIBRARY_PATH` and `DYLD_LIBRARY_PATH` are ignored due to [SIP](https://en.wikipedia.org/wiki/System_Integrity_Protection).

MSoegtropIMC:
I thought about this a bit more. Actually I am using a few shared libraries without issues and I think the reason this works is that I install the shared libraries before I link them to executables, so that the executable knows the absolute path of the library. I wonder what happens during installation of a z3 executable using a z3 shared library. Would the install process patch the executable to point it to the new installation path of the shared library, or would it relay on that the destination path is in LD_LIBRARY_PATH?

k4rtik:
I believe @arbipher patched Z3 to work with opam-installed shared libraries as described at https://github.com/Z3Prover/z3/issues/5578#issuecomment-947082861 before the release of z3-4.8.13.

k4rtik:
Just an update on the CI status, there are 6 tests stuck on:
```
[time]: Waiting for resource in pool OCluster
[time]: Waiting for worker...
```
since yesterday. They all have passed the tests before, so they should not be considered a blocker for merge. I do not plan to rerun the CI as it seems unpredictable as I [noticed in the last run](https://github.com/ocaml/opam-repository/pull/20065#issuecomment-987125027).

k4rtik:
@MSoegtropIMC are there still any tests that you wanted to do before this PR can be considered ready to merge?

MSoegtropIMC:
Yes, but unfortunately I have 0 time right now and as it looks wont have before early next week. Would this be OK with you?

k4rtik:
I have a commit ready for Z3 4.8.14 at https://github.com/k4rtik/opam-repository/commit/442415f367be37462fe036d7be161c1a29b5db59 and will make a PR once the discussion resolves here. cc: @MSoegtropIMC

k4rtik:
@arbipher as you [mentioned on OCaml Discuss](https://discuss.ocaml.org/t/ask-for-suggestions-for-the-next-z3-package-release/8922/10?u=kartik), did you get a chance to port the install script to cmake?

arbipher:
@k4rtik Not yet. It is in my plan but I even haven't finished other tech debts during the break. Feel free if you have plans to do that.

packages/z3/z3.4.8.13/opam
```diff
@@ -0,0 +1,39 @@
...
+  "ocaml"
+  "ocamlfind" {build}
+  "zarith"
+  "conf-gmp"
+  "conf-python-3" {build}
+]
+x-ci-accept-failures: [
+  "centos-7" "oraclelinux-7" # C compiler is too old
+]
+synopsis: "Z3 solver"
```
kit-ty-kate, line 31: If this is to be merged in the current state, I believe it at least needs to be a message to tell the users about this silent breaking change. Something like this:
```suggestion
synopsis: "Z3 solver"
post-messages: [
  """
This version of Z3 changed the linking mode from static to dynamic.
This change is silent and potentially breaking.
Please make sure that you don’t require static binaries in the programs using Z3.
""" {success}
]
```

MSoegtropIMC:
Just wanted to note that I was busy with other tings the last 6 weeks or so. I will now look into if the suggested packages work for my applications. Should be finished soon.

MSoegtropIMC:
I put the package from this PR into my local patch repo and adjusted my packages which use the z3 library and all works fine. So I have no issues with this.

Two notes:
- I have a local package for the TPTP front end for Z3  here [z3_tptp](https://github.com/coq/platform/tree/main/opam/opam-repository/packages/z3_tptp/z3_tptp.4.8.13). TPTP is a standard exchange format more common with ATPs than with SMTs. It might make sense to put it into the main opam repo - I would appreciate your opinion on this. My package is a bit ugly (calls cpp directly), but I figured that this is more robust and faster than using Z3's make facilities. opam knows much better where the libraries are and the front end has only 2 C files.
- There is a Z3 4.8.14 meanwhile.

Do you need help with CI?

MSoegtropIMC:
P.S.: I hope you don't mind that I put your package into the Coq Platform opam patch repo for the time being - I plan to do a release next week and want to keep it like this.

k4rtik:
@MSoegtropIMC I don't have any opinion on the inclusion of Z3 TPTP frontend, but I think it is orthogonal to this PR, so I wouldn't want to include it here.

I included Z3 4.8.14 as well here.

Unsure if we need any more help with CI than what opam-ci does here. I will let the maintainers chip in on that question.

arbipher:
I agree with kk4rtik that we can discuss it after this PR.
Would you mind discussing it on the forum [post](https://discuss.ocaml.org/t/ask-for-suggestions-for-the-next-z3-package-release/8922), or just have a new issue here?

I don't know too much about tptp. I am thinking tptp is just an official [binary](https://github.com/Z3Prover/z3/blob/master/examples/tptp/README). If in the future as proposed in that post, z3 is split into a system-level library and an opam binding, tptp may be also better as a system-level package. Personally, I am fully convinced by the splitting idea because I don't think (e.g.) ubuntu provides the latest z3 while currently this ocaml z3 bindings are definitely built with the (latest) corresponding z3.

NikolajBjorner:
The tptp front-end is an example and it is an executable. I was not aware that it was used until now.
The VampireZ3 executable should subsume the functionality, except it may have complementary 
capabilities when it comes to quantifiers (in some contexts it is a joke, in others it makes sense: 
have a pass-through mode for VampireZ3 where it just calls z3).
Geoff Sutcliffe will be very happy to hear about your use; I can schedule a sailing trip with him in good faith.
My take is that a dependency "z3 - tptp - ocaml" may be asking for more work than reasonable, but I don't know how your tool chains work (everything goes over opam, including executables..).

MSoegtropIMC:
Regarding the tptp front end: it is used by a Coq plugin which tries to prove lemmas automatically by throwing them at various ATPs (eprover, z3-tptp, cvc4, vampire). As far as I understand this, the plugin communicates with all these solvers in tptp format and it uses stand alone executables of the solvers, so no need for OCaml bindings. I didn't implement this - I just do system integration (Coq Platform).

The Coq plugin works fine with the executable created by the z3_tptp opam package I linked above.

@NikolajBjorner : Thanks for pointing me to VampireZ3. Would you say it makes sense to use both front ends for Z3 in parallel?

I am perfectly fine to do the TPTP front end as a separate opam package, which then depends on the z3 opam package for the shared library - this is how it is done currently.

NikolajBjorner:
I have not benchmarked vampirez3 vs z3 so have no idea how they compare. The internals are different, vampirez3 calls z3 as part of Avatar loop and the base architecture is quantifier free. While it can in principle do many other different things I have not code-reviewed the current state or the state you use. Vampire can be quite effective at quantifier reasoning, while model construction is not taking theories into account as first class. A pass-through mode from Vampire would be more sustainable because TPTP is maintained as a first-class front-end to Vampire. But you give me a superb excuse to mingle with Geoff to bring the TPTP front-end up to date for features that may be of interest.

kit-ty-kate:
Just for the record, this PR is not being merge because this is a heavy silent breaking change and I want to make sure everyone agrees before commiting to such change (CI does not matter here).

I’ve pinged the Fstar devs as well but they haven’t replied yet.

kit-ty-kate:
Ok, let’s merge it now then and deal with the fallouts later. Thanks!

> For me adjusting to this package only required to change the library name from z3-static to z3 and to change the library folder (the static lib is in "lib", the shared library is in "stublibs").

What do you mean by this? How does that work for users of the library?

PR #20069, opened 2021-11-20
Package otoml.0.9.3

dmbaturin:
### `otoml.0.9.3`
TOML parsing, manipulation, and pretty-printing library (1.0.0-compliant)
OTOML is a library for parsing, manipulating, and pretty-printing TOML files.

* Fully 1.0.0-compliant.
* No extra dependencies: default implementation uses native numbers and represents dates as strings.
* Provides a functor for building alternative implementations: plug your own bignum and calendar libraries if required.
* Informative parse error reporting.
* Pretty-printer offers flexible indentation options.

---
* Homepage: https://github.com/dmbaturin/otoml
* Source repo: git+https://github.com/dmbaturin/otoml.git
* Bug tracker: https://github.com/dmbaturin/otoml/issues

---

### DIFF
--- a/packages/otoml/otoml.0.9.3/opam
+++ b/packages/otoml/otoml.0.9.3/opam
@@ -1,0 +1,44 @@
+opam-version: "2.0"
+synopsis:
+  "TOML parsing, manipulation, and pretty-printing library (1.0.0-compliant)"
+description: """\
+OTOML is a library for parsing, manipulating, and pretty-printing TOML files.
+
+* Fully 1.0.0-compliant.
+* No extra dependencies: default implementation uses native numbers and represents dates as strings.
+* Provides a functor for building alternative implementations: plug your own bignum and calendar libraries if required.
+* Informative parse error reporting.
+* Pretty-printer offers flexible indentation options."""
+maintainer: "daniil+opam@baturin.org"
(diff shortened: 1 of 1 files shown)


dmbaturin:
##  Changelog

### Breaking changes

The functor now takes a single `TomlNumber` module insteaf of independent `TomlInteger` and `TomlFloat`
modules.
The reason for that change is that for integer/float conversions to work, the module needs to know
both int and float types and provide conversion functions that involve both types.
That change has no effect on the default implementation interface.

### New features

* `get_opt` and `get_result` combinators.
* `get_table_values` for quickly retrieving unboxed values from homogemous tables.
* `string_of_path : `
* New `*_opt` and `*_result` versions of the high level interface functions and `*_exn` aliases.

### Bug fixes

* Accessors: fixed incorrect float to boolean conversion and added int/float conversions.
* Exceptions are now correctly exposed in the functorial interface.
* `path_exists` is correctly exposed now as well.
* Integer and float implicit conversions in non-strict mode are now supported.

### Internal changes

* The lexer no longer uses internal mutable state so it's now re-entrant and (hopefully) thread-safe.
* Modules signatures now use `include` to avoid copying large chunks, so the library is much easier to contribute to.

mseri:
Thanks

PR #20080, opened 2021-11-22
PRINGO version 1.3

xavierleroy:
- Add `uniform` function producing a random floating-point number in (0.0, 1.0) 
- Add the LXM generator (the L64X128 variant)
- New test infrastructure based on TestU01
- Add Schaathun's "split sequences" to the tests


kit-ty-kate:
Thanks! Could you return the fix in the opam file upstream so that the issue doesn’t happen in the next release?

xavierleroy:
> Could you return the fix in the opam file upstream so that the issue doesn’t happen in the next release?

Will do ! Thanks for the emergency fix and for the merge.

PR #20081, opened 2021-11-22
base v0.14.2

cwong-ocaml:
Includes [this fix](https://github.com/janestreet/base/pull/124)

### DIFF
--- a/packages/base/base.v0.14.2/opam
+++ b/packages/base/base.v0.14.2/opam
@@ -10,7 +10,7 @@ build: [
   ["dune" "build" "-p" name "-j" jobs]
 ]
 depends: [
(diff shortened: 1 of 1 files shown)


cwong-ocaml:
Looks like there's some incompatibility with ocaml-variants >= 0.14.0? I'm unsure how to proceed.

kit-ty-kate:
I’m not sure to understand why the requirement was bumped to a compiler that is not released yet. My PR was compatible with previous versions of the compiler as well, as far as I know.

cwong-ocaml:
looks like a miscommunication, let's try again

kit-ty-kate:
Thanks a lot!

PR #20102, opened 2021-11-25
Tezt: fix checksums and install script

rbardou:
The checksums for Octez (the Tezos implementation) tarballs changed a while ago. We are working on making them more stable, but in the meantime, here is a patch that fixes them for the `tezt` package (which is part of the source code of Octez).

Additionally, the `opam` file for `tezt` did not have the `["mv" "tezt/lib/%{name}%.install" "./"]` line which is required because the `.opam` file (and thus the `.install`) file is not at the root, so `opam install tezt` did not actually install anything. This patch also fixes that.

I assume there is a policy to not replace existing packages, but is it also the case for this kind of fixes? If so, should I instead push a new version?

### DIFF
--- a/packages/tezt/tezt.1.0.0/opam
+++ b/packages/tezt/tezt.1.0.0/opam
@@ -14,13 +14,14 @@ depends: [
 ]
 build: [
   ["dune" "build" "-p" name "-j" jobs]
+  ["mv" "tezt/lib/%{name}%.install" "./"]
 ]
 synopsis: "Framework for integration tests with external processes"
 
 url {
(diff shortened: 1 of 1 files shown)


samoht:
It's fine to merge this change without bumping the version number if:
1. you, as the maintainer of that package are happy with this change (I guess that's the case as you are the one proposing the change :p)
2. you have actually checked that contents of the archive is similar (for instance by checking what is in the opam cache) and/or you are sure that this is pointing to the right git tag.

rbardou:
Thank you for providing the original tarball. I was able to check the diff in the `tezt/lib` folder and all files are identical in the new tarball of this MR.

So to answer @samoht I confirm that:
1) I am happy with this change;
2) I have checked the contents is the same, at least the part that is relevant for the `tezt` library.

PR #20109, opened 2021-11-26
[new release] goblint (1.1.0)

sim642:
Static analysis framework for C

- Project page: <a href="https://goblint.in.tum.de">https://goblint.in.tum.de</a>
- Documentation: <a href="https://goblint.readthedocs.io/en/latest/">https://goblint.readthedocs.io/en/latest/</a>

##### CHANGES:

Goblint "fat" release containing 4 years of development.
This version is functionally equivalent to Goblint in SV-COMP 2022.

### DIFF
--- a/packages/goblint/goblint.1.1.0/opam
+++ b/packages/goblint/goblint.1.1.0/opam
@@ -1,0 +1,73 @@
+opam-version: "2.0"
+synopsis: "Static analysis framework for C"
+maintainer: [
+  "Michael Schwarz <michael.schwarz93@gmail.com>"
+  "Simmo Saan <simmo.saan@gmail.com>"
+  "Ralf Vogler <ralf.vogler@gmail.com>"
+]
+authors: [
+  "Vesal Vojdani"
+  "Kalmer Apinis"
+  "Ralf Vogler"
+  "Michael Schwarz"
+  "Julian Erhard"
+  "Simmo Saan"
+]
+license: "MIT"
+homepage: "https://goblint.in.tum.de"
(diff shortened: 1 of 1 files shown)


sim642:
If I understand the lower-bounds CI correctly, I should add a newer lower bound, e.g. `>= 0.5` for ocaml-monadic, right?

For the `oUnit` library, I guess it should be `ounit2` in the dune file. I'm a bit confused how that didn't come up locally.

Another problem I already see in the testing log is `Could not resolve host: github.com`, where the tests want to download some custom Linux headers. While not listed as the problem, it should fail `runtest`, so the CI wouldn't pass. Is there anything to do about that other than simply removing those tests from `runtest`?

dinosaure:
> If I understand the lower-bounds CI correctly, I should add a newer lower bound, e.g. >= 0.5 for ocaml-monadic, right?

Yes :+1:.

> For the oUnit library, I guess it should be ounit2 in the dune file. I'm a bit confused how that didn't come up locally.

Yeah the module name does not match the package name.

> Another problem I already see in the testing log is Could not resolve host: github.com, where the tests want to download some custom Linux headers. While not listed as the problem, it should fail runtest, so the CI wouldn't pass. Is there anything to do about that other than simply removing those tests from runtest?

An installation _via_ `opam` is done with `bwrap` which disallow you to use extra-resources during installation or tests. By this way, something like download extra resources from internet is not possible - and it's dangerous for the end-user. Can you provide a test which does not need to download something elsewhere?

sim642:
Subsumed by #20132.

PR #20112, opened 2021-11-26
Add plugin flag to opam-0install

dra27:


kit-ty-kate:
cc @talex5. Does it really make sense to have a command `opam 0install`? I’m not sure opam-0install is meant to be that user facing

dra27:
I only hit this from the repro case in https://github.com/ocaml/opam/issues/4554#issue-811269537 and got bitten because opam 2.1 is more fastidious about plugins being tagged than 2.0 is (I think).

dra27:
(I therefore have no strong opinion either way!)

dra27:
xref https://github.com/ocaml-opam/opam-0install-solver/pull/34

kit-ty-kate:
The issue is that `opam-0install` does not do any installation so it’s a bit weird to have such a command that only calls a solver.

Maybe if it was called `opam-solve`?

dra27:
This appears to have been a typo in the original issue leading my slightly down the wrong garden path! Sorry for the noise...

Issue #19975, opened 2021-11-10
[CI macos-homebrew (experimental)] suspicious failures with cmdliner


mseri:
Seen on https://github.com/ocaml/opam-repository/pull/19969 and https://github.com/ocaml/opam-repository/pull/19973
The failures read:
```
#=== ERROR while compiling opam-installer.2.1.0 ===============================#
# context              2.1.0 | macos/x86_64 | ocaml-system.4.13.1 | file:///Volumes/tank/result/05776a68916ac0fe591e74b46fac18c5d9e5aa2d3e7bda717d718c54966f514e/opam-repository
# path                 ~/.opam/default/.opam-switch/build/opam-installer.2.1.0
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p opam-installer -j 11
# exit-code            1
# env-file             ~/.opam/log/opam-installer-21580-076bba.env
# output-file          ~/.opam/log/opam-installer-21580-076bba.out
### output ###
#       ocamlc src/tools/.opam_installer.eobjs/byte/opam_installer.{cmi,cmo,cmt}
# File "src/tools/opam_installer.ml", line 1:
# Warning 70 [missing-mli]: Cannot find interface file.
#     ocamlopt src/tools/opam_installer.exe (exit 2)
[...]

PR #20113, opened 2021-11-26
Package SZXX.2.0.0

SGrondin:
### `SZXX.2.0.0`
Streaming ZIP XML XLSX parser
SZXX is a streaming, non-seeking and efficient XLSX parser built from ground up for low memory usage.
SZXX is able to output XLSX rows while a file is being read from the file descriptor without buffering any part of the file.

---

---

### DIFF
--- a/packages/SZXX/SZXX.2.0.0/opam
+++ b/packages/SZXX/SZXX.2.0.0/opam
@@ -1,0 +1,38 @@
+opam-version: "2.0"
+maintainer: "Asemio"
+authors: [
+  "Simon Grondin"
+]
+synopsis: "Streaming ZIP XML XLSX parser"
+description: """
+SZXX is a streaming, non-seeking and efficient XLSX parser built from ground up for low memory usage.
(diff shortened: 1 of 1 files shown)


SGrondin:
I'll take another look tomorrow, but right now I'm not seeing how `decompress.de` is unavailable in some but not all of the tests.

Any help is appreciated.

SGrondin:
The only failures are the "Experimental" macOS builds. I cannot reproduce it using the docker instructions because:
image

However:
- everything works on an actual macOS computer
- this library does not rely on any Unix/system symbols (it supports jsoo)
- the bug seems to be in `decompress`

Is there anything I can do to unblock this PR?

patricoferris:
Apologies for the macOS errors, this looks like a slightly different version of #19975 which is most likely the builder's fault and not yours.

SGrondin:
Thank you @patricoferris ! I hope it means it can be merged in :)

dinosaure:
Thanks!

PR #10462, opened 2017-10-13
ZArith version 1.7

xavierleroy:
The modern library for arbitrary-precision integer and rational arithmetic.

New in release 1.7:
* Issue #14, pull request #15: ARM assembly code was broken.
* Fix tests so that they work even if the legacy Num library is unavailable, as is the case in a base install of OCaml 4.06.

PR #10441, opened 2017-10-10
Fix msgpack.1.3.0 dependencies

jpdeplaix:
Is there a rule that distinguishes between ```base-x``` and ```x``` packages ? It seems to have different rules between ```bytes```, ```num```, ```camlp4``` and ```ocamlbuild```.

PR #10437, opened 2017-10-10
Package pcre.7.3.2

mmottl:
### `pcre.7.3.2`

Bindings to the Perl Compatibility Regular Expressions library

pcre-ocaml offers library functions for string pattern matching and
substitution, similar to the functionality offered by the Perl language.

---
* Homepage: https://mmottl.github.io/pcre-ocaml
* Source repo: https://github.com/mmottl/pcre-ocaml.git
* Bug tracker: https://github.com/mmottl/pcre-ocaml/issues

---

---
### 7.3.2 (2017-10-10)

  * Improved compatibility with MSVC
:camel: Pull-request generated by opam-publish v0.3.5

Issue #181, opened 2012-11-13
bisect 1.1: missing dependency on str, and bisectThread* not installed


edwintorok:
Problem:
  The META file is missing the dependency on str.cma for bisect_pp:
  bisectThread.cmx is not installed

Possible fix:
  Upgrade to bisect 1.3, and apply Makefile patch from http://bugs.x9c.fr/view.php?id=108

The META file is missing the dependency on str.cma for bisect_pp in bisect 1.1 (fixed in 1.3):
Camlp4: Uncaught exception: DynLoader.Error ("/home/edwin/.opam/4.00.1/lib/bisect/bisect_pp.cmo", "error while linking /home/edwin/.opam/4.00.1/lib/bisect/bisect_pp.cmo.\nReference to undefined global `Str'")

When trying to use bisect on a program compiled with threads I get an error that bisectThread.cmx is missing, and indeed it was not installed. This is a problem with bisect 1.3 too, see here for a patch:
http://bugs.x9c.fr/view.php?id=108

PR #20114, opened 2021-11-26
Allow ocamlformat.0.19.0 to be compiled on OCaml 4.14

kit-ty-kate:
cc @gpetiot does that looks ok to you? I’m not seeing any errors or warnings when compiling and the [changelog](https://github.com/ocaml/ocaml/blob/trunk/Changes) only lists things that should not require special code in ocamlformat (at least from what I can guess):
```
### Language features:

- #10462: Add attribute to produce a compiler error for polls.

- #10437: Allow explicit binders for type variables.

- #10441: Remove unnecessary parentheses surrounding immediate objects.
  Allow 'object ... end # f', 'f object ... end', etc.

- #181, #9760: opt-in tail-modulo-cons (TMC) transformation
    let[@tail_mod_cons] rec map f li = ...
 ```

### DIFF
--- a/packages/ocamlformat/ocamlformat.0.19.0/opam
+++ b/packages/ocamlformat/ocamlformat.0.19.0/opam
@@ -9,7 +9,7 @@ homepage: "https://github.com/ocaml-ppx/ocamlformat"
 bug-reports: "https://github.com/ocaml-ppx/ocamlformat/issues"
 depends: [
   "dune" {>= "2.8"}
-  "ocaml" {>= "4.08" & < "4.14"}
+  "ocaml" {>= "4.08" & < "4.15"}
   "alcotest" {with-test}
   "base" {>= "v0.12.0" & < "v0.15"}
   "base-unix"


kit-ty-kate:
cc @emillon

gpetiot:
> - #10462: Add attribute to produce a compiler error for polls.

There is nothing to do


> - #10437: Allow explicit binders for type variables.

This should work with https://github.com/ocaml-ppx/ocamlformat/pull/1757


> - #10441: Remove unnecessary parentheses surrounding immediate objects.
>   Allow 'object ... end # f', 'f object ... end', etc.

This should work with https://github.com/ocaml-ppx/ocamlformat/pull/1934


> - #181, #9760: opt-in tail-modulo-cons (TMC) transformation
>    let[@tail_mod_cons] rec map f li = ...

There is nothing to do

kit-ty-kate:
Ok fair enough. I'll close this then. Could you make a new release of ocamlformat? It looks like everything is ready.

Thanks a lot!

emillon:
In addition, I believe it's fine to let 0.19.0 build with ocaml 4.14. It won't be able to format certain 4.14 files but that's expected. Having 4.14 build more than 0.20+ can possibly help ocaml-ci with lint-fmt jobs too.

PR #20129, opened 2021-11-29
[new release] camyll (0.4.0)

alan-j-hu:
A static site generator

- Project page: <a href="https://alan-j-hu.github.io/camyll">https://alan-j-hu.github.io/camyll</a>

##### CHANGES:

- Switch from To.ml to OTOML library (alan-j-hu/camyll#1, Daniil Baturin).
- Close a leaked file handle.
- Use `slug` library instead of handrolled function for generating slugs.


alan-j-hu:
I was not aware of the stack overflow. It looks like a problem with the uunf transitive dependency.

PR #20130, opened 2021-11-29
Use the archive cache on opam 2.2

kit-ty-kate:
To merge when https://github.com/ocaml/opam/pull/4830 is merged. See linked issues for previous attempts at doing this


kit-ty-kate:
Sadly per https://opam.ocaml.org/doc/Manual.html#repofield-archive-mirrors this syntax is forbidden... Return to square one..

PR #20137, opened 2021-11-30
Alpine capnproto-dev is now in the main repository

talex5:
The old `capnproto-dev@edgecommunity` now fails with:

    command              /usr/bin/capnp --version
    Error relocating /usr/lib/libcapnpc.so.0.9.1: _ZSt28__throw_bad_array_new_lengthv: symbol not found
    Error relocating /usr/bin/capnp: _ZSt28__throw_bad_array_new_lengthv: symbol not found

### DIFF
--- a/packages/conf-capnproto/conf-capnproto.2/opam
+++ b/packages/conf-capnproto/conf-capnproto.2/opam
@@ -1,0 +1,33 @@
+opam-version: "2.0"
+maintainer: "Kate <kit.ty.kate@disroot.org>"
+authors: [
+  "Sandstorm Development Group, Inc."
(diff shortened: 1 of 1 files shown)


talex5:
Can I indicate that it is now expected to fail on the old alpine-3.13? I tried this, but it still said failed:
```
 x-ci-accept-failures: [
   "alpine-3.13"   # Does not have capnproto in the main repository
```

mseri:
I don't understand why it is failing to filter out alpine, but this should not block the merge I think

mseri:
Ping @kit-ty-kate: should we open an issue somewhere for this?

PR #20147, opened 2021-12-01
Package core_compat.0.14.0

aalekseyev:
### `core_compat.v0.14.0`
Compatibility for core v0.14
Compatibility wrapper to make it possible to have code compatible with both Core v0.14 and v0.15.

---
* Homepage: https://github.com/janestreet/core_compat
* Source repo: git+https://github.com/janestreet/core_compat.git
* Bug tracker: https://github.com/janestreet/core_compat/issues

---

### DIFF
--- a/packages/core_compat/core_compat.0.14.0/opam
+++ b/packages/core_compat/core_compat.0.14.0/opam
@@ -1,0 +1,28 @@
+opam-version: "2.0"
+maintainer: "opensource@janestreet.com"
+authors: ["Jane Street Group, LLC <opensource@janestreet.com>"]
+homepage: "https://github.com/janestreet/core_compat"
+bug-reports: "https://github.com/janestreet/core_compat/issues"
+dev-repo: "git+https://github.com/janestreet/core_compat.git"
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
The version number does not match your usual scheme. Did you mean `v0.14.0` instead?

aalekseyev:
Yes, I did, thanks. Fixing.

aalekseyev:
Pushed the fix. I wonder where `opam publish` took the "0.14" number from.

kit-ty-kate:
> Pushed the fix. I wonder where `opam publish` took the "0.14" number from.

from the tag. By default if the `version:` field is not present or if the `-v <version>` argument is not given, opam-publish will try to guess based on the tag. By default it will ignore the `v` from the tag as it is a common thing to do to separate between tags and branches for example.
dune-release also has the same behaviour (though has a `--keep-v` argument as a more direct workaround)

aalekseyev:
Actually the #20154 achieves this better. I don't think we need `core_compat` then (at least not in its current state).

PR #20151, opened 2021-12-02
Remove reference to non-existing archive for unavailable ezjs_min.0.2.2

AltGr:


mseri:
Let’s first check if we can find it in a cache backup

avsm:
It's not in the older opam-0.ocaml.org cache.

packages/ezjs_min/ezjs_min.0.2.2/opam
```diff
@@ -32,11 +32,4 @@ build: [
   ]
 ]
 dev-repo: "git+https://github.com/ocamlpro/ezjs_min.git"
```
mseri, line 34:
```suggestion
dev-repo: "git+https://github.com/ocamlpro/ezjs_min.git"
available: false # The tarball is no longer available
```

mseri:
Thanks, then I guess we don't have an alternative

PR #20153, opened 2021-12-02
dune 1.11: add conflict with num =1.0

mseri:
This is a test to check if the error
```

#=== ERROR while compiling dune.1.11.0 ========================================#
# context              2.1.1 | linux/x86_64 | ocaml-base-compiler.4.10.2 | file:///home/opam/opam-repository
# path                 ~/.opam/4.10/.opam-switch/build/dune.1.11.0
# command              ~/.opam/opam-init/hooks/sandbox.sh build ./boot.exe --release -j 31
# exit-code            1
# env-file             ~/.opam/log/dune-99-bf8fcb.env
# output-file          ~/.opam/log/dune-99-bf8fcb.out
### output ###
# Error: stat: /home/opam/.opam/4.10/lib/ocaml/arith_flags.cmx: No such file or directory
# -> required by src/stdune/.stdune.objs/native/stdune__Appendable_list.cmx
# -> required by src/catapult/.catapult.objs/native/catapult.cmx
# -> required by src/.dune.objs/native/dune__Action.cmx
# -> required by bin/.main.objs/native/main.cmx
# -> required by bin/main.a
# -> required by bin/main/main_dune.exe
# -> required by install bin/dune
# -> required by dune.install
```
seen on Frama-C PR is due to a problem in num =1.0

### DIFF
--- a/packages/dune/dune.1.11.0/opam
+++ b/packages/dune/dune.1.11.0/opam
@@ -34,6 +34,7 @@ conflicts: [
   "jbuilder" {!= "transition"}
   "odoc" {< "1.3.0"}
   "dune-release" {< "1.3.0"}
+  "num" {= "1.0"}
 ]
 dev-repo: "git+https://github.com/ocaml/dune.git"
 build: [
--- a/packages/dune/dune.1.11.0/opam
+++ b/packages/dune/dune.1.11.0/opam
@@ -33,6 +33,7 @@ conflicts: [
   "jbuilder" {!= "transition"}
   "odoc" {< "1.3.0"}
   "dune-release" {< "1.3.0"}
+  "num"
 ]
 dev-repo: "git+https://github.com/ocaml/dune.git"
 build: [
--- a/packages/dune/dune.1.11.0/opam
+++ b/packages/dune/dune.1.11.0/opam
@@ -33,7 +33,6 @@ conflicts: [
   "jbuilder" {!= "transition"}
   "odoc" {< "1.3.0"}
   "dune-release" {< "1.3.0"}
-  "num" {= "1.0"}
 ]
 dev-repo: "git+https://github.com/ocaml/dune.git"
 build: [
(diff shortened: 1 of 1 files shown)


dra27:
The conflict doesn't feel right, and marking num.1.0 as unavailable definitely doesn't (lockfiles, etc.). Reproducing locally, say, https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/73d98347c3c6abab26619e872fabd5648cca833d/variant/opam-2.1,compilers,4.11,frama-c.24.0,lower-bounds, I don't have any problems installing num.1.0 and dune.1.11.0 (in fact, the build failed because of C compiler problems with lablgtk).

dra27:
I'm not sure what the underlying issue is without being able to reproduce it with an entire log - I was wondering if it's that `num` is only partially installed, but that seems surprising, given the number of instances of it in the original PR.

packages/dune/dune.1.11.0/opam
```diff
@@ -34,6 +34,7 @@ conflicts: [
   "jbuilder" {!= "transition"}
   "odoc" {< "1.3.0"}
   "dune-release" {< "1.3.0"}
+  "num" {= "1.0"}
```
mseri, line 37:
```suggestion
```

Let's make an experiment and see if removing this lets the issue appear

mseri:
> The conflict doesn't feel right, 

I agree, it was just to see if the issue goes away with it.

> and marking num.1.0 as unavailable definitely doesn't (lockfiles, etc.).

Then I'd like to understand why it happens. It was clearly a problem with num, some files where missing

mseri:
Maybe the issue is here:
```
-> installed base-bytes.base
[ERROR] The compilation of dune.1.11.0 failed at "./boot.exe --release -j 31".
-> installed camlzip.1.06
-> installed num.1.0
```

is it possible that it is trying to compile dune before num is installed and some dependencies of dune, or dune itself, actually need num?

mseri:
In that case, dune should depend on num because `stdune` and `catapult` do

mseri:
That's also not it, since without num is working fine. I don't know. I'll close this, it was to check if it was related or not, and I agree that the issue must be hidden somewhere else

PR #20154, opened 2021-12-02
Package core_unix.v0.14.0

aalekseyev:
In Jane Street v0.15 we're about to split up core library and move most of it into package `core_unix`.

This PR adds a compatibility package `core_unix.v0.14.0` which should be largely compatible with the upcoming `core_unix.v0.15.0`, while being implemented on top of `core.v0.14.*`.

cc @kit-ty-kate 

Signed-off-by: Arseniy Alekseyev <aalekseyev@janestreet.com>

### DIFF
--- a/packages/core_unix/core_unix.v0.14.0/opam
+++ b/packages/core_unix/core_unix.v0.14.0/opam
@@ -1,5 +1,4 @@
 opam-version: "2.0"
-version: "v0.14.0"
 maintainer: "opensource@janestreet.com"
 authors: ["Jane Street Group, LLC <opensource@janestreet.com>"]
 homepage: "https://github.com/janestreet/core_unix"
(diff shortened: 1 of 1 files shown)


packages/core_unix/core_unix.v0.14.0/opam
```diff
@@ -0,0 +1,29 @@
+opam-version: "2.0"
+version: "v0.14.0"
```
mseri, line 2:
```suggestion
```

aalekseyev: Thanks. I wasn't sure why or whether this field was needed, but I cargo-culted it from somewhere.

mseri:
No worries. Name and version are inferred from the path in the opam repository, so we have started removing them when they are left over in the linting stage

mseri:
Looks good, thanks

PR #20158, opened 2021-12-02
Inital draft of OpenCog CogUtil package

linas:
Base package for all other OpenCog projects. Provides core build tools and general infrastructure.

### DIFF
--- a/packages/cogutil/cogutil.opam-1.0/opam
+++ b/packages/cogutil/cogutil.opam-1.0/opam
@@ -9,8 +9,6 @@
 (** Finally, say `eval $(opam env)` *)
 
 opam-version: "2.0"
-name: "cogutil"
-version: "opam-1.0"
 synopsis: "OpenCog CogUtil base package"
 description: """
 The CogUtil package provides base infrastructure for other OpenCog
--- a/packages/cogutil/cogutil.opam-1.0/opam
+++ b/packages/cogutil/cogutil.opam-1.0/opam
@@ -23,7 +23,7 @@ license: "AGPL-3.0-or-later"
 homepage: "https://opencog.org"
 bug-reports: "https://github.com/opencog/cogutil/issues"
 dev-repo: "git+https://github.com/opencog/cogutil/"
-depends: [ "ocaml" "ocamlfind" ]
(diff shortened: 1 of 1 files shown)


mseri:
This cannot work. The C dependencies should be passed using the [depexts](https://opam.ocaml.org/doc/Manual.html#opamfield-depexts) field – look at `conf-` packages in this repository to get an idea. 

Sudo should not be necessary, and cannot be used in the opam sandbox anyway. It seems to me that you need it, because your package wants to install libraries in the system folders, which is something that is no longer allowed by opam (unless these are prepackaged and installable via depexts).

It seems to me that this package does not contain any ocaml code, so the dependency on ocaml and ocamlbuild seems incorrect. Are you sure you want to publish this on the opam repository?

packages/cogutil/cogutil.opam-1.0/opam
```diff
@@ -0,0 +1,29 @@
+(** OPAM (OCaml Package Manager) module for CogUtil *)
+(** *)
+(** This is in the root directory because I cannot figure out *)
```
mseri, line 3: That is correct, the opam files are in the root of ocaml packages distributed via opam. But this seems to be a C/C++ package, so I am not sure why you need the opam file at all

packages/cogutil/cogutil.opam-1.0/opam
```diff
@@ -0,0 +1,29 @@
...
+license: "AGPL-3.0-or-later"
+homepage: "https://opencog.org"
+bug-reports: "https://github.com/opencog/cogutil/issues"
+dev-repo: "git+https://github.com/opencog/cogutil/"
+depends: [ "g++" "cmake" "libboost-dev" "libboost-program-options-dev" "libboost-system-dev" "libboost-thread-dev" "binutils-dev" "libiberty-dev" "ocaml" "ocamlfind" ]
+build: [
+  ["cmake" "."]
+  [make]
+]
+install: ["sudo" "make" "install"]
```
mseri, line 29: You cannot run sudo in the opam sandbox, commands cannot access any folder outside the sources folder

linas:
Thanks, I am digesting...

mseri:
I am going to close this for now. Feel free to reopen once it is ready

linas:
Yes, thanks, the project hit a speed-bump. I see what to fix here, but there unrelated issues blocking forward progress.

PR #20167, opened 2021-12-08
Package swipl.0.5

Gopiandcode:
### `swipl.0.5`
Bindings to SWI-Prolog for OCaml
Swipl are Free Software high-level OCaml bindings to SWI Prolog 8.5, supporting easy interop between OCaml and Prolog.

---
* Homepage: https://gitlab.com/gopiandcode/swipl-ocaml
* Source repo: git+https://gitlab.com/gopiandcode/swipl-ocaml.git
* Bug tracker: https://gitlab.com/gopiandcode/swipl-ocaml/issues

---


Gopiandcode:
First time publishing a package with bindings to C, so I'm not sure if I've done this correctly.

My bindings require a specific version of Prolog (8.5 or higher - i.e the latest) - this is the version shipped on the Debian bullseye (which is now stable). My generated C stubs have a check to make sure the version is correct and fails if they have an older ffi version and so the build will fail, rather than silently going through. However, there's nothing in the opam dependencies that tells a user about this - is that fine? should I instead write my header to make the code silently go through, even though many of the symbols and functions the user expects to work will not behave as expected?

mseri:
Looks good to me. The lower bounds are fine, and it is compiling wherever it can (and fail with incorrect version when an incorrect version is detected). I'll wait for a double check by one of the other maintainers but I think this is good to merge

mseri:
Thanks

PR #20168, opened 2021-12-08
add ez_search 0.2.0

RadioPotin:


RadioPotin:
Hi,

This is the first release for `ez_search` I was unsure on how to constraint the package to linux only so feel free to indicate how I could do that if I've messed up something :)

Until next time !

PR #20182, opened 2021-12-08
Package p5scm.0.1.0

drjdn:
### `p5scm.0.1.0`
Scheme via camlp5
This a simple scheme implementation using pa_schemer from camlp5

---
* Homepage: https://github.com/drjdn/p5scm
* Source repo: git+https://github.com/drjdn/p5scm.git
* Bug tracker: https://github.com/drjdn/p5scm/issues

---
:camel: Pull-request generated by opam-publish v2.1.0

PR #20188, opened 2021-12-09
Package p5scm.0.1.0

drjdn:
### `p5scm.0.1.0`
Scheme via camlp5
This a simple scheme implementation using pa_schemer from camlp5

---
* Homepage: https://github.com/drjdn/p5scm
* Source repo: git+https://github.com/drjdn/p5scm.git
* Bug tracker: https://github.com/drjdn/p5scm/issues

---


kit-ty-kate:
I had a look at your fixes and the bug mentioned in https://github.com/ocaml/opam-repository/pull/20182#issuecomment-989372702 is still present. This won’t work with subsequent minor releases of the compiler

drjdn:
Got it.  If I change to upper bound on ocaml version to 4.13.1 would that be acceptable?

kit-ty-kate:
@drjdn no not really. The only correct way to define this package would be to set the dependency to ocaml to be exact versions of `ocaml-base-compiler`. e.g.
```
"ocaml-base-compiler" {= "4.10.2" | = "4.10.1" | = "4.10.2" | = "4.09.1" | ...}
```
this is neither sustainable for your nor for your users so I’d encourage you to look into a solution. It should be quite easy to replace the `%{ocaml_version}` variable by something more reasonable

drjdn:
@kit-ty-kate I've looked at the failing tests.  The `lower-bounds` for opam-2.1 isn't clear to me so I can't replicate.

kit-ty-kate:
Thanks!

PR #20202, opened 2021-12-11
2 packages from c-cube/ocaml-avro at 0.1

c-cube:
This pull-request concerns:
-`avro.0.1`: Runtime library for encoding/decoding Avro
-`avro-compiler.0.1`: Schema compiler for Avro

---
* Homepage: https://github.com/c-cube/ocaml-avro
* Source repo: git+https://github.com/c-cube/ocaml-avro.git
* Bug tracker: https://github.com/c-cube/ocaml-avro/issues/

---


kit-ty-kate:
Thanks! Could you return the fix in the opam file upstream? (if you want to keep the opam file autogeneration from dune, you just have to upgrade to dune >= 2.7)

c-cube:
hmm I use dune 2.9.1 locally, how comes?

kit-ty-kate:
yes sorry that's what i meant, you need at least (lang dune 2.7)

c-cube:
if it's a bug in dune, why isn't it fixed regardless of `(lang …)`? :/

kit-ty-kate:
because it would be a nightmare if different people with different versions of dune were to commit different conflicting changes to the same opam file since it is autogenerated and promoted every time `dune build` is called, sadly.

PR #20211, opened 2021-12-12
Package nanoid.1.0.0

mefyl:
### `nanoid.1.0.0`
Nano ID implementation for OCaml

---
* Homepage: https://github.com/routineco/ocaml-nanoid
* Source repo: git+https://github.com/routineco/ocaml-nanoid.git
* Bug tracker: https://github.com/routineco/ocaml-nanoid/issues

---


mefyl:
First submission ever of a package here, let me know what I did wrong 🙈

mefyl:
The package depends on `npm-conf` to have node installed in the environment, yet the tests also need a run of `npm` to install the nanoid javascript package. How can I get the CI to run it?

kit-ty-kate:
Would it be possible to vendor the required npm packages for the tests?
You can test if it passes the sandbox by running: `~/.opam/opam-init/hooks/sandbox.sh build <your command>`

mefyl:
> Would it be possible to vendor the required npm packages for the tests?

I'll try that thanks!

Wouldn't it make sense to run it automatically when there is a dependency on `npm`? In the present case `node_modules` is rather small, but it can get really huge.

kit-ty-kate:
> Wouldn't it make sense to run it automatically when there is a dependency on `npm`?

I’m not sure to understand. What do you mean by "run"?

mefyl:
I mean, automatically run `npm` if the package depends on `npm-conf` and has a `package.json` at its root. Just like it will automatically run `opam install --deps-only .` or something similar to pull your Opam dependencies. That would make any package that supports js_of_ocaml and imports javascript work out of the box in the CI instead of having to vendor all javascript dependencies.

But that is just an humble suggestion, I did vendor the dependencies.

kit-ty-kate:
Could you open a feature request on https://github.com/ocaml/opam/issues ?

I’m still not entirely sure what you are suggesting though.
- Is it for releases? In which case opening a feature request in opam-publish/dune-release seems more appropriate.
- Or do you mean always run npm before building? In which case please feel free to detail the feature request on the opam bugtracker (linked above). The same thing could be helpful for users of cargo but I’m not familiar with either of them so I’m not sure.

But in any case, making the feature depend on `conf-npm` would not be possible in opam given opam doesn’t know about any of the packages in opam-repository. A flag could be used instead though.

kit-ty-kate:
For the last commit see https://github.com/ocaml/opam/issues/4959

mefyl:
Maybe it's me having a misconception about how the Opam CI works. My understanding is that it has a series of step that are out of my control and that build and check any well-defined Opam package. 

One of those steps are (from the output of the jobs):

* `opam remove nanoid.1.0.0 && opam depext nanoid.1.0.0 && opam install --deps-only nanoid.1.0.0 && opam install -v nanoid.1.0.0`

My suggestion would be to insert a step before that one akin to `which -q npm && test -f package.json && npm`. That would have no effect on the CI of most packages, and would make packages such as nanoid build out of the box without having to vendor npm dependencies since I suppose packages also targeting js_of_ocaml might have js dependencies. In my understanding, this modification would be part of the CI of Opam, not Opam itself.

Hopefully this makes more sense; if so, happy to open a feature request.

kit-ty-kate:
opam-repo-ci cannot do anything here as it does not have access to the directory where the package it built ahead of time. Also the CI is here to check that everything builds correctly out of the box. If we start adding adhoc rules that users don’t know about it looses its purpose.

opam-repo-ci only does the equivalent of `opam install --with-test <pkg>`, which anyone can do and should be able to replicate with any setup. You definitly want this feature in opam, opam-publish or dune-release (depending on what you actually want), not in opam-repo-ci.

kit-ty-kate:
Thanks! Could you return the fixes in the opam file upstream so that it can be taken into account next time?

mefyl:
Makes sense. I'll reflect on it a bit.

> Thanks! Could you return the fixes in the opam file upstream so that it can be taken into account next time?

Done. Thanks for your help!

PR #20216, opened 2021-12-13
Update archive links of old codes

zoggy:
On my way to remove old repositories, I archived git histories of these codes to a single git repo and added tags so that old archives can now be retrieved from this new repo. For example, dbforge 2.0.1 archive can now be retrieved from https://framagit.org/zoggy/old-codes/-/archive/dbforge-2.0.1/old-codes-dbforge-2.0.1.tar.gz 

I did not updated any other information (except checksums of course). I doubt all these codes will be missing to someone.

### DIFF
--- a/packages/config-file/config-file.1.2/opam
+++ b/packages/config-file/config-file.1.2/opam
@@ -19,7 +19,6 @@ install: [make "install"]
 synopsis: "Small library to define, load and save options files."
 flags: light-uninstall
 url {
-  src:
-    "https://download.ocamlcore.org/config-file/config_file/1.2/config-file-1.2.tar.gz"
-  checksum: "md5=fece1f143285fb5fddf17d5d36a577c5"
+  src: "https://framagit.org/zoggy/old-codes/-/archive/config-file-1.2/old-codes-config-file-1.2.tar.gz"
+  checksum: "md5=2f3679799ae8e3bc67847a204aae71c9"
 }
--- a/packages/config-file/config-file.1.1/opam
+++ b/packages/config-file/config-file.1.1/opam
@@ -14,7 +14,6 @@ depends: ["ocaml" "ocamlfind" "camlp4"]
 install: [make "install"]
 synopsis: "Small library to define, load and save options files."
 url {
-  src:
-    "https://download.ocamlcore.org/config-file/config_file/1.1/config-file-1.1.tar.gz"
-  checksum: "md5=7bc051234ceb29ffd5823ee6bcffe19c"
+  src: "https://framagit.org/zoggy/old-codes/-/archive/config-file-1.1/old-codes-config-file-1.1.tar.gz"
+  checksum: "md5=77b0e72a15fe22868c6eceb8ba1f7325"
 }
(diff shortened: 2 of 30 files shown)


mseri:
According to pkgdiff these are all unchanged archives, thanks

mseri:
I don't think we should address all failures as part of this PR

packages/config-file/config-file.1.2/opam
```diff
@@ -19,7 +19,6 @@ install: [make "install"]
 synopsis: "Small library to define, load and save options files."
 flags: light-uninstall
 url {
-  src:
-    "https://download.ocamlcore.org/config-file/config_file/1.2/config-file-1.2.tar.gz"
-  checksum: "md5=fece1f143285fb5fddf17d5d36a577c5"
+  src: "https://framagit.org/zoggy/old-codes/-/archive/config-file-1.2/old-codes-config-file-1.2.tar.gz"
+  checksum: "md5=2f3679799ae8e3bc67847a204aae71c9"
```
dra27, line 23: These two tarballs differ:

```diff
diff -Naur config-file-1.2.orig/configure old-codes-config-file-1.2/configure
--- a/configure      2014-03-20 09:44:28.000000000 +0000
+++ b/configure 2014-03-20 09:48:40.000000000 +0000
@@ -1673,7 +1673,7 @@

 # Check for Ocaml compilers

-OCAML_NEEDED=3.08.0
+OCAML_NEEDED=4.00

 # we first look for ocamlfind in the path; if not present, we fail
 # Extract the first word of "ocamlfind", so it can be a program name with args.
```
which is causing a patch in opam-repository-mingw to cease applying (cf. https://github.com/ocaml/setup-ocaml/issues/361). @zoggy - is it possible to shift the commit so the tarball is the same?

mseri: 👀 I thought I had checked them all…

smorimoto: As Windows builds are continually failing, it seems best to address this as soon as possible.

mseri: I have sent a pr to upload the old files to opam-source-archives. With that we can revert to the old files, but that needs to be merged. Alternatively, you can update the patches on the windows opam repo

dra27: It feels cleaner to be reverting to tarballs which are identical - which also doesn't require a patch in opam-repository-mingw.

mseri: As I mentioned on slack it is fine with me. I think the fastest way is to merge the opam repo source PR, then I can make an update on the fly

dra27: Can you transfer the pull request over to ocaml/opam-source-archives and I'll merge 🙂

mseri: Lol I got the wrong repo?

mseri: Done, sorry 😁

PR #20218, opened 2021-12-13
Solo5 0.6.9

dinosaure:
* virtio: add missing reset for net and blk devices (Solo5/solo5#491)
* CI: disable genode (Solo5/solo5#493)
  Genode bindings are failing to build on Debian testing / GCC 10.x.
* Avoid "-" in section names (in the ld scripts) since some GNU ld versions
  reject that (alpine 3.15, Fefora 35) (Solo5/solo5#502)
* On OpenBSD, OpenBSD 6.7 is not supported anymore, all support releases use
  ld.lld (Solo5/solo5#495)

### DIFF
--- a/packages/solo5-bindings-hvt/solo5-bindings-hvt.0.6.9/opam
+++ b/packages/solo5-bindings-hvt/solo5-bindings-hvt.0.6.9/opam
@@ -36,6 +36,9 @@ available: [
   (arch = "x86_64" | arch = "arm64") &
   (os = "linux" | os = "freebsd" | os = "openbsd")
 ]
+x-ci-accept-failures: [
+  "centos-7"
+]
 synopsis: "Solo5 sandboxed execution environment (hvt target)"
 description: """
 Solo5 is a sandboxed execution environment primarily intended
--- a/packages/solo5-bindings-hvt/solo5-bindings-hvt.0.6.9/opam
+++ b/packages/solo5-bindings-hvt/solo5-bindings-hvt.0.6.9/opam
@@ -1,0 +1,55 @@
+opam-version: "2.0"
+maintainer: "martin@lucina.net"
+authors: [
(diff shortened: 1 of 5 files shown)


packages/solo5-bindings-hvt/solo5-bindings-hvt.0.6.9/opam
```diff
@@ -0,0 +1,55 @@
...
+  "Martin Lucina <martin@lucina.net>"
+  "Ricardo Koller <kollerr@us.ibm.com>"
+]
+homepage: "https://github.com/solo5/solo5"
+bug-reports: "https://github.com/solo5/solo5/issues"
+license: "ISC"
+dev-repo: "git+https://github.com/solo5/solo5.git"
+build: [
+  ["./configure.sh"]
+  [make "V=1" "CONFIG_SPT=" "CONFIG_VIRTIO=" "CONFIG_MUEN=" "CONFIG_GENODE=" "CONFIG_XEN="]
```
mseri, line 14: Are all the flags white on purpose?

mseri:
Looks good

packages/solo5-bindings-hvt/solo5-bindings-hvt.0.6.9/opam
dinosaure: Yes, this is due to the process release which pick pre-existing OPAM files.

dinosaure:
It seems that Solo5 does not work on CentOS 7 (/cc @hannesm), should we add a constraint? Then, failures seems unrelated to the release (but it's nice to see all of these failures!).

hannesm:
@dinosaure yes, maybe mark it unavailable on centos7, or allow ci failures on that platform. maybe that should be done for earlier releases as well.

dinosaure:
I added `x-ci-failures` on packages :+1:.

PR #20224, opened 2021-12-13
Add the merlin preview for OCaml 4.14

kit-ty-kate:
Same as #19408 but for OCaml 4.14

Upstream PR: https://github.com/ocaml/merlin/pull/1415
cc @voodoos


mseri:
We cannot test this right?

kit-ty-kate:
not automatically yet no

PR #20275, opened 2021-12-17
[new release] doi2bib (0.5.2)

mseri:

##### CHANGES:

- Move from cuz to the published clz
- Move from dx.doi.org to crossref rest api service,
  the latter gives better and more consistent results and
  does not seem to require a fallback service any longer
- Update arxiv generated bibtex accordingly
- Update ocamlformat

### DIFF
--- a/packages/doi2bib/doi2bib.0.5.2/opam
+++ b/packages/doi2bib/doi2bib.0.5.2/opam
@@ -1,0 +1,46 @@
+opam-version: "2.0"
+synopsis:
+  "Small CLI to get a bibtex entry from a DOI, an arXiv ID or a PubMed ID"
+maintainer: ["marcello.seri@gmail.com"]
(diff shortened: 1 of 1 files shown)


packages/doi2bib/doi2bib.0.5.2/opam
```diff
@@ -0,0 +1,46 @@
...
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
```
mseri, line 33:
```suggestion
#    "@runtest" {with-test}
```

mseri: Tests require the network. Cannot run in the sandbox

PR #20295, opened 2021-12-22
Package opam-dune-expand.v0.0.1

Gopiandcode:
### `opam-dune-expand.v0.0.1`
Plugin to view ppx-expanded OCaml source files
opam-dune-expand is an opam plugin that allows users to view the expanded output of PPXs on their source files.

---
* Homepage: https://gitlab.com/gopiandcode/dune-expand
* Source repo: git+https://gitlab.com/gopiandcode/dune-expand.git
* Bug tracker: https://gitlab.com/gopiandcode/dune-expand/issues

---

### DIFF
--- a/packages/opam-dune-expand/opam-dune-expand.v0.0.1/opam
+++ b/packages/opam-dune-expand/opam-dune-expand.v0.0.1/opam
@@ -1,0 +1,43 @@
+opam-version: "2.0"
+synopsis: "Plugin to view ppx-expanded OCaml source files"
+description:
+  "opam-dune-expand is an opam plugin that allows users to view the expanded output of PPXs on their source files."
+maintainer: "kirang@comp.nus.edu.sg"
+authors: "Kiran Gopinathan"
+license: "GPL-3.0+"
+homepage: "https://gitlab.com/gopiandcode/dune-expand"
+bug-reports: "https://gitlab.com/gopiandcode/dune-expand/issues"
+depends: [
+  "dune" {>= "2.8"}
+  "sexplib" {>= "~"}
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
I’ve had a glance on the source code and I have a question: why the choice to make it an opam plugin? Opam doesn’t seem to be used anywhere in this project. Why not keep the upstream repository name: dune-expand ?

More related the the PR itself, the `v` prefix shouldn’t be there, in opam-repository it is only used as a trick for janestreet packages (as `v0.9.0` > `120.0.0`). Could you rename the release to `0.0.1` instead of `v0.0.1` ?

Gopiandcode:
> Opam doesn’t seem to be used anywhere in this project. Why not keep the upstream repository name: dune-expand ?

I defined it as an opam-package following the style of `dune-lint`, primarily as it is intended as a tool to help debug projects, and thus felt it would make sense as an executable launched from opam, but if that doesn't make sense then I can remove the plugin part.

Regarding the version, sure, can do.

kit-ty-kate:
`opam-dune-lint` makes sense because it uses the opam library to check that the opam file is in sync with the dune files. In that sense it can be regarded as an extension of `opam lint`.

`dune-expend` doesn’t make sense to use as a plugin to me. If we go this way we can argue for tools like `dune`, `merlin` or even basically any packages installing executable to become plugins.

To me it even makes your users less happy because they would have to type more characters to lunch your binary for no obvious purpose.

Gopiandcode:
I see, thanks for the explanation, that makes sense. I'll close this request and open another with the suggested changes.

Thanks.

PR #20256, opened 2021-12-15
3 packages from gitlab.inria.fr/fpottier/menhir/-/archive/20211215/archive.tar.gz

fpottier:
This pull-request concerns:
-`menhir.20211215`: An LR(1) parser generator
-`menhirLib.20211215`: Runtime support library for parsers generated by Menhir
-`menhirSdk.20211215`: Compile-time library for auxiliary tools related to
 Menhir

---
* Homepage: http://gitlab.inria.fr/fpottier/menhir
* Source repo: git+https://gitlab.inria.fr/fpottier/menhir.git
* Bug tracker: https://gitlab.inria.fr/fpottier/menhir/-/issues

---
:camel: Pull-request generated by opam-publish v2.1.0

PR #20303, opened 2021-12-23
Add further menhir upper bounds

mseri:
Relates to failures in https://github.com/ocaml/opam-repository/pull/20256

I think these are all the ones appearing, the bounds were added with `opam admin add-constraint 'menhir<20211215'`, I only cherrypicked which packages to add


mseri:
I am not 100% sure about rfsm (feel free to take it out) and I am not going to fix the lower bounds for this PR :P

kit-ty-kate:
It doesn’t seem related but we can’t test further somehow. But it doesn’t hurt anything given the new constraint is on a new package. We can always unlock it later. Thanks!

PR #20310, opened 2021-12-24
LablGtk 2.18.12

garrigue:
New release of lablgtk2, with new features and better compatibility with 5.00

PR #20317, opened 2021-12-24
lablgtk3.3.1.2 isn’t compatible with OCaml < 4.09

kit-ty-kate:
Uses caml_sys_modify_argv as pointed out in https://github.com/ocaml/opam-repository/pull/20310#discussion_r775039077

cc @garrigue could you add some tests to at least build some example binaries to detect this kind of issues early? (one of the binaries in `examples/` would do I guess)

### DIFF
--- a/packages/lablgtk3/lablgtk3.3.1.2/opam
+++ b/packages/lablgtk3/lablgtk3.3.1.2/opam
@@ -16,7 +16,7 @@ license: "LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception"
 doc: "https://garrigue.github.io/lablgtk/lablgtk3"
 
 depends: [
-  "ocaml"     {         >= "4.05.0" }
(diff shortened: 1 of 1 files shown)


mseri:
Looks good

garrigue:
Thanks for the fix, I missed that one.
As for testing earlier in the future, we will try to do something. I'm not a dune expert, but @ejgallego is.

ejgallego:
The problem we have is that labgltk ci did shut down and we didn't have time to resurrect it. This, releases and merges go unchecked. A working ci would have captured this issue.

garrigue:
But did the CI check for older releases of ocaml ?
This is actually one difficult part to test by hand, because this is extremely tedious.

ejgallego:
> But did the CI check for older releases of ocaml ? This is actually one difficult part to test by hand, because this is extremely tedious.

That was almost the whole point of it

PR #20271, opened 2021-12-16
Package lame.0.3.5

toots:
### `lame.0.3.5`
MP3 encoding library
Bindings for the lame library which provides functions for encoding mp3 files

---
* Homepage: https://github.com/savonet/ocaml-lame
* Source repo: git+https://github.com/savonet/ocaml-lame.git
* Bug tracker: https://github.com/savonet/ocaml-lame/issues

---
:camel: Pull-request generated by opam-publish v2.1.0

PR #20322, opened 2021-12-25
Add support for Github actions-based CI

toots:
This PR adds support for CI using github actions, following a discussion in https://github.com/ocaml/opam-repository/pull/20271#issuecomment-999170463

Example of a full run: https://github.com/toots/opam-repository/actions/runs/1620485627

### Implementation details
* The builds are made from scratch, using official docker images for distributions or github runners and the standard `ocaml/setup-ocaml@v2` and etc. No need to maintain a matrix of docker images
* OCaml builds are cached and take about `1m40s`
* Most features from the current CI are supported except:
  * The `opensus` image was buggy and had to be dropped
  * lowerbound tests aren't implemented as it seems that the required solver is not (yet) available with the `opam` build shipped by `ocaml/setup-ocaml@v2`
  * Architectures other than the one provided by the runners (`linux`/`mac`, `amd64`) are not supported. Windows could be but the dance between official `opam` packages and `windows` packages does seem to make this painful.
  * `extra` are supported based on available `ocaml-option-*` packages.
 
 Overall, this seems to be an interesting option as this is free and scales well. It could be nice to have it along the official CI to see how it behaves and, perhaps, use it to alleviate some of the pressure on the current CI.

### DIFF
--- a/.github/check-installable/action.yml
+++ b/.github/check-installable/action.yml
@@ -1,0 +1,35 @@
+name: Check if package is installable
+inputs:
+  opam:
+    required: true
+  package:
+    required: true
+outputs:
+  installable:
+    description: "Is the package installable?"
+    value: ${{ steps.installable.outputs.installable }}
+runs:
+  using: composite
+  steps:
+    - name: Get opam cli option
+      id: cli
+      shell: sh
+      run: |
+        if [ "${{ inputs.opam }}" = "opam-2.1" ]; then
+          echo '##[set-output name=cli]--cli=2.1'
+        fi
+    - name: Check if package is installable
+      id: installable
+      shell: sh
+      run: |
+        if opam ${{ steps.cli.outputs.cli }} list --readonly --external --resolve=${{ inputs.package }} >/dev/null 2>&1; then
+          echo '##[set-output name=installable]true';
+        else
+          echo "Package is not installable";
+        fi
+    - name: Annotate result
+      if: ${{ !steps.installable.outputs.installable }}
+      uses: actions/github-script@v5
+      with:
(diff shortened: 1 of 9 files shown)


kit-ty-kate:
As explained in the "discussion" you linked above, we are not looking to use github actions.
I feel a bit bad to close this but I hope you didn’t wasted too much time on this PR on this christmas day.

Github actions only supports up to 20 concurrent jobs, our cluster supports up to a 1000, a simple revdeps run on dune or other heavily used package would simply kill github actions and for simple packages this would result in the same behaviour as opam-repo-ci.

As said previously, if you have ideas on how to improve opam-repo-ci you can open feature requests at https://github.com/ocurrent/opam-repo-ci/issues.
If you want to have a voice chat with us feel free to hit me up on Discord or other messaging system of your choice.

toots:
Haha. No, no problem. It was a good exercise for me to understand both github actions and how the CI works here.

I still think this could be added at no cost and provide and alternative to the main CI, I don't see why would be lost by adding it.

Happy holidays!

kit-ty-kate:
> I still think this could be added at no cost and provide and alternative to the main CI, I don't see what would be lost by adding it.

It still brings quite a heavy maintenance and duplication cost as well as confusion for users and a difficult to navigate interface for us maintainers (having to scroll a list of 1000s of PR checks that we would not use just to get the results from opam-repo-ci would be quite painful).
I would much rather focus our efforts into improving what we have. For instance here is a proposal to improve the UI and split core checks and revdeps checks to improve readability for users: https://github.com/ocurrent/opam-repo-ci/issues/127

PR #20325, opened 2021-12-26
8 packages from savonet/ocaml-ffmpeg at 1.1.1

toots:
This pull-request concerns:
-`ffmpeg.1.1.1`: Bindings for the ffmpeg libraries
-`ffmpeg-av.1.1.1`: Bindings for the ffmpeg libraries -- top-level helpers
-`ffmpeg-avcodec.1.1.1`: Bindings for the ffmpeg avcodec library
-`ffmpeg-avdevice.1.1.1`: Bindings for the ffmpeg avdevice library
-`ffmpeg-avfilter.1.1.1`: Bindings for the ffmpeg avfilter library
-`ffmpeg-avutil.1.1.1`: Bindings for the ffmpeg avutil libraries
-`ffmpeg-swresample.1.1.1`: Bindings for the ffmpeg swresample library
-`ffmpeg-swscale.1.1.1`: Bindings for the ffmpeg swscale library

---
* Homepage: https://github.com/savonet/ocaml-ffmpeg
* Source repo: git+https://github.com/savonet/ocaml-ffmpeg.git
* Bug tracker: https://github.com/savonet/ocaml-ffmpeg/issues

---


kit-ty-kate:
Would you be able to add the appropriate constraints to these packages if the changes are expected?
Screenshot 2021-12-26 at 16 51 43

toots:
@kit-ty-kate I'm not quite sure how to fix this one. The released liquidsoap configure is not smart enough to detect which submodules it exactly needs. I have set the depopt to the top-level `ffmpeg` but it still picks up any installed modules when running the configure and fails with them.

Is there a way to declare more finer-grained depopts like needs `ffmpeg` if `ffmpeg-avutils` is installed?

toots:
@kit-ty-kate I've had another review. In fact, all liquidsoap 2.x packages in opam now only optionally depend on top-level `ffmpeg` so I'm not sure why they are pulled as revdeps for the subpackages?

toots:
@kit-ty-kate I've tightened up the dependencies, this should be passing now.

mseri:
Looks green where it can be (almost everywhere)

packages/liquidsoap/liquidsoap.2.0.1/opam
```diff
@@ -36,13 +36,7 @@ depopts: [
   "dssi"
   "faad"
   "fdkaac"
-  "ffmpeg-avutil"
-  "ffmpeg-av"
-  "ffmpeg-avcodec"
-  "ffmpeg-avdevice"
-  "ffmpeg-avfilter"
-  "ffmpeg-swresample"
-  "ffmpeg-swscale"
```
kit-ty-kate, line 45: According to https://github.com/savonet/liquidsoap/commit/d21649a6c3594dbae8a702b99b53fb52163d3401 2.0.1 still has optional dependencies toward these packages.

This does not seem right. For example this unlocks opam to compile both ffmpeg-av at the same time as liquidsoap, which could cause random failures/race-conditions.

packages/liquidsoap/liquidsoap.2.0.1/opam
```diff
@@ -89,13 +83,9 @@ conflicts: [
   "faad" {< "0.5.0"}
   "fdkaac" {< "0.3.1"}
   "ffmpeg" {< "1.1.0"}
+  "ffmpeg" {> "1.1.0"}
```
kit-ty-kate, line 86:
```suggestion
  "ffmpeg" {>= "1.1.1"}
```
Using `>=` is usally always better than its stricter version as it would not allow for an eventual `1.1.0-1`

packages/liquidsoap/liquidsoap.2.0.1/opam
```diff
@@ -89,13 +83,9 @@ conflicts: [
...
   "ffmpeg-avutil" {< "1.1.0"}
-  "ffmpeg-av" {< "1.1.0"}
-  "ffmpeg-avcodec" {< "1.1.0"}
-  "ffmpeg-avdevice" {< "1.1.0"}
-  "ffmpeg-avfilter" {< "1.1.0"}
-  "ffmpeg-swresample" {< "1.1.0"}
-  "ffmpeg-swscale" {< "1.1.0"}
```
kit-ty-kate, line 98: Couldn’t they not have `{>= "1.1.1"}` like the two other ones above?

packages/liquidsoap/liquidsoap.2.0.1/opam
toots: I see you point. Yes, the underlying problem is that the build system for liquidsoap `2.0.x` makes the wrong assumptions about which compbinations of ffmpeg subpackages it needs for its files.. 

I have fixed that in the upcoming `2.0.2` bugfix release, which directly depends on the top-level `ffmpeg` library. It will come out as soon as this package is accepted, after which I believe these issues will matter much less.

packages/liquidsoap/liquidsoap.2.0.0~rc1/opam
```diff
@@ -83,6 +83,8 @@ conflicts: [
   "fdkaac" {< "0.3.1"}
   "ffmpeg" {< "1.0.1"}
   "ffmpeg" {> "1.1.0"}
+  "ffmpeg-avutil" {< "1.0.1"}
+  "ffmpeg-avutil" {> "1.1.0"}
```
toots, line 87:
```suggestion
  "ffmpeg-avutil" {>= "1.1.1"}
```

packages/liquidsoap/liquidsoap.2.0.0/opam
```diff
@@ -83,6 +83,8 @@ conflicts: [
   "fdkaac" {< "0.3.1"}
   "ffmpeg" {< "1.0.1"}
   "ffmpeg" {> "1.1.0"}
+  "ffmpeg-avutil" {< "1.0.1"}
+  "ffmpeg-avutil" {> "1.1.0"}
```
toots, line 87:
```suggestion
  "ffmpeg-avutil" {>= "1.1.1"}
```

packages/liquidsoap/liquidsoap.2.0.1/opam
```diff
@@ -89,13 +83,9 @@ conflicts: [
...
   "ffmpeg" {< "1.1.0"}
+  "ffmpeg" {>= "1.1.1"}
   "ffmpeg-avutil" {< "1.1.0"}
-  "ffmpeg-av" {< "1.1.0"}
-  "ffmpeg-avcodec" {< "1.1.0"}
-  "ffmpeg-avdevice" {< "1.1.0"}
-  "ffmpeg-avfilter" {< "1.1.0"}
-  "ffmpeg-swresample" {< "1.1.0"}
-  "ffmpeg-swscale" {< "1.1.0"}
+  "ffmpeg-avutil" {> "1.1.0"}
```
toots, line 88:
```suggestion
  "ffmpeg-avutil" {>= "1.1.0"}
```

packages/liquidsoap/liquidsoap.2.0.0~rc1/opam
```diff
@@ -83,6 +83,8 @@ conflicts: [
   "fdkaac" {< "0.3.1"}
   "ffmpeg" {< "1.0.1"}
   "ffmpeg" {> "1.1.0"}
```
toots, line 85:
```suggestion
  "ffmpeg" {>= "1.1.1"}
```

packages/liquidsoap/liquidsoap.2.0.0/opam
```diff
@@ -83,6 +83,8 @@ conflicts: [
   "fdkaac" {< "0.3.1"}
   "ffmpeg" {< "1.0.1"}
   "ffmpeg" {> "1.1.0"}
```
toots, line 85:
```suggestion
  "ffmpeg" {>= "1.1.1"}
```

packages/liquidsoap/liquidsoap.2.0.1/opam
toots: I'm not sure if there's a great solution. As I said, I believe the correct solution is to let this one in and then add liquidsoap `2.0.2`, which fixes the underlying issue and would be the default pick for the `2.0.x` branch.

packages/liquidsoap/liquidsoap.2.0.1/opam
kit-ty-kate: In the future, it might be easier to split the package/libraries into one per optional dependency that you have right now and transform the liquidsoap package into a package providing the binary (with optional dependencies to the sublibraries but with `= version` constraints).
Opam isn’t really clever enough to handle the truck load of issues that comes with optional dependencies and it might be easier for your users.

toots: Definitely! This is part of the plan with the `dune` migration but, liquidsoap is a large codebase with historically a lot of finer-tunes compilation details.

packages/liquidsoap/liquidsoap.2.0.1/opam
kit-ty-kate: But it changes the behaviour of this package quite heavily and enables race-condition during the compilation. What is so bad about adding `{>= "1.1.1"} ?

toots: I'm happy to implement either changes. If I add these back, the CI revdeps will start failing again, which is what triggered those changes in the first place.

toots: @kit-ty-kate Any preference here?

kit-ty-kate: I’d rather keep the depopts as they are currently and add a new conflict for each

packages/liquidsoap/liquidsoap.2.0.0~rc1/opam
```diff
@@ -83,6 +90,20 @@ conflicts: [
...
```
toots, line 92:
```suggestion
  "ffmpeg" {>= "1.1.0"}
```

packages/liquidsoap/liquidsoap.2.0.1/opam
toots: @kit-ty-kate You got it. Done! Thanks for your time with this.

kit-ty-kate:
Looks fine now. Thanks!

For next time, it looks like these the ffmpeg library contains naked pointers which makes it incompatible with OCaml 5.0 coming out very soon next year.

toots:
> Looks fine now. Thanks!
> 
> For next time, it looks like these the ffmpeg library contains naked pointers which makes it incompatible with OCaml 5.0 coming out very soon next year.

Sounds good thks for the pointer.

PR #20331, opened 2021-12-29
Make latest soupault depend on otoml >= 1.0.1 to avoid the inline table lookup bug

dmbaturin:
Soupault is affected by the bug fixed in OTOML 1.0.1 and that makes certain valid configurations unusable if they use inline tables instead of "normal" tables for nested values.


mseri:
Thanks

PR #20338, opened 2021-12-31
Package res_tailwindcss.0.1.2

mattdamon108:
### `res_tailwindcss.0.1.2`
PPX validates the tailwindcss class names
ppx_tailwindcss validates the tailwindcss class names in compile time.

---
* Homepage: https://github.com/green-labs/res_tailwindcss
* Source repo: git+https://github.com/green-labs/res_tailwindcss.git
* Bug tracker: https://github.com/green-labs/res_tailwindcss/issues

---

### DIFF
--- a/packages/res_tailwindcss/res_tailwindcss.0.1.2/opam
+++ b/packages/res_tailwindcss/res_tailwindcss.0.1.2/opam
@@ -1,0 +1,32 @@
+opam-version: "2.0"
+synopsis: "PPX validates the tailwindcss class names"
+description: """
+ppx_tailwindcss validates the tailwindcss class names in compile time.
+"""
+maintainer: "Greenlabs Dev <developer@greenlabs.co.kr>"
+authors: "Greenlabs Dev <developer@greenlabs.co.kr>"
+license: "MIT"
+homepage: "https://github.com/green-labs/res_tailwindcss"
+bug-reports: "https://github.com/green-labs/res_tailwindcss/issues"
+dev-repo: "git+https://github.com/green-labs/res_tailwindcss.git"
+depends: [
+  "ocaml" {>= "4.12.1"}
+  "dune" { >= "2.8"}
+  "ppxlib"
+  "core"
+  "ppx_inline_test"
+  "ppx_expect"
+  "ppx_deriving"
+  "menhir" { >= "20211230"}
+]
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
What is the reason behind not having the `dune-project` and `res_tailwindcss.opam` files in the root directory? Could you change the release and move them there? Otherwise this makes this will confuse dune in local opam switch for example

mattdamon108:
@kit-ty-kate Frankly, I don't know the dune system and how *.opam works pretty much. It is my first try to publish my work. I'll move the `dune-project` and `res_tailwindcss.opam` files into the project root and change the PR.

mattdamon108:
@kit-ty-kate One question, is there any proper way to package the sub directory? If I can `opam publish` the `src` directory only, then I think everything would be fine.
```
root - src (dune-project, res_tailwindcss.opam files are here)
     |
     - test (rescript related)
```

kit-ty-kate:
> @kit-ty-kate One question, is there any proper way to package the sub directory? If I can `opam publish` the `src` directory only, then I think everything would be fine.
> 
> ```
> root - src (dune-project, res_tailwindcss.opam files are here)
>      |
>      - test (rescript related)
> ```

tezos packages does it by copying opam files around before calling dune but it’s not really pretty.
the alternative would be to make sure that the archive publish to opam-repository packages only the src directory (as root dir) but this is not an option in either dune-release or opam publish and it complicates things quite a bit.

Furthermore having it in subdirectories makes it very annoying for users to easily use the dev version of your package (e.g. `opam pin res_tailwindcss --dev`) as the opam file would not be found and package information would get out-of-sync. The same thing is true if users wants to opam pin locally (`opam install .`) as it is quite uncanny to have to go to the `src` directory first.

mattdamon108:
@kit-ty-kate Thank you for your detailed explanation. I change the structure of my project and re-publish it.

kit-ty-kate:
Looks good now. Thanks!

For next time, please consider merging https://github.com/green-labs/res_tailwindcss/pull/6, install files really shouldn’t be in the repository (though technically it does not matter since it will be overwritten anyway by dune)

mattdamon108:
@kit-ty-kate Thanks! I'll merge your PR.

PR #20340, opened 2021-12-31
Solo5 0.7.0

dinosaure:
## v0.7.0 (2021-12-31)

* Remove Genode bindings (#494, #506)
  The Genode bindings becomes incompatible with the upcoming LLVM/Clang based
  toolchain. It can be reinstated in the future if there is interest.

  `GENODE_ABI_TARGET` is not removed to disallow to re-use it for another
  purpose than the Genode binding.
* New packaging, toolchains, cross-compilation (#494, #506)
  + Make Solo5 installable system-wide and be able to packaging Solo5 for the
    BSDs and Linux distributions
  + Replace Solo5-internal use of `pkg-config` by generated toolchain wrappers
    installed as "PREFIX/bin/ARCH-solo5-none-static-{cc,ld,objcopy}"
  + Solo5 bindings (ABIs) are co-installable under the same PREFIX
  + OPAM packaging is simplified (only one package `solo5` remains)
  + Experimental: support for the cross-compilation (specially `aarch64`)
* virtio: FreeBSD requires the grub-bhyve command (#506)
* virtio: FreeBSD: wait a bit for `cat` to create the nmdm device (#506)
* OpenBSD: all supported releases of OpenBSD use `ld.ldd` (#495, #506)
* Be able to release our cross toolchain via our opam-release.sh script (#504)

**NOTE**: `pkg-config` still is required by Solo5 to get flags from
`libseccomp`.

### DIFF
--- a/packages/solo5-cross-aarch64/solo5-cross-aarch64.0.7.0/opam
+++ b/packages/solo5-cross-aarch64/solo5-cross-aarch64.0.7.0/opam
@@ -28,6 +28,9 @@ depexts: [
 ]
 available: [
   (arch != "arm64") &
+  (arch != "arm32") &
+  (arch != "ppc64") &
+  (arch != "s390x") &
   (os = "linux" & os-family = "debian")
 ]
 synopsis: "Solo5 sandboxed execution environment"
--- a/packages/solo5-cross-aarch64/solo5-cross-aarch64.0.7.0/opam
+++ b/packages/solo5-cross-aarch64/solo5-cross-aarch64.0.7.0/opam
@@ -1,0 +1,45 @@
+opam-version: "2.0"
+maintainer: "martin@lucina.net"
+authors: [
+  "Dan Williams <djwillia@us.ibm.com>"
+  "Martin Lucina <martin@lucina.net>"
+  "Ricardo Koller <kollerr@us.ibm.com>"
+]
+homepage: "https://github.com/solo5/solo5"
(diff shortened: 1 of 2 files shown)


packages/solo5-cross-aarch64/solo5-cross-aarch64.0.7.0/opam
```diff
@@ -0,0 +1,48 @@
...
+  ["kernel-headers"] {os-distribution = "rhel"}
+  ["linux-libc-dev"] {os-family = "debian"}
+  ["gcc-aarch64-linux-gnu"] {os-family = "debian"}
+]
+available: [
+  (arch != "arm64") &
+  (arch != "arm32") &
+  (arch != "ppc64") &
+  (arch != "s390x") &
+  (os = "linux" & os-family = "debian")
```
kit-ty-kate, line 34: Is it really available only on Debian? If so why even add depexts for other distributions?

dinosaure: I'm not the author of this choice and currently, `solo5-cross-aarch64` works on `ubuntu` and ArchLinux (/cc @TheLortex) but I think Martin Lucina was pretty conservative about this new package. As an experimental release (see the Changelog), I think we should keep these constraints and then check if everything works for others distributions :+1:.

kit-ty-kate:
ok then. Thanks!

PR #20341, opened 2021-12-31
Release conjury-2.1.

jhwoodyatt:
The Conjury package is currently used only by Orsetto, whose forthcoming next release depends on features in this new minor version.

### DIFF
--- a/packages/conjury/conjury.2.1/opam
+++ b/packages/conjury/conjury.2.1/opam
@@ -1,0 +1,25 @@
+opam-version: "2.0"
+synopsis: "Conjury library for OMake"
+maintainer: "james woodyatt <jhw@conjury.org>"
+authors: "james woodyatt <jhw@conjury.org>"
+license: "BSD-2-Clause"
+tags: "org:conjury.org"
+homepage: "https://bitbucket.org/jhw/conjury/"
+bug-reports: "https://bitbucket.org/jhw/conjury/issues"
+depends: [
+  "base-unix"
+  "ocaml" {>= "4.08"}
+  "ocamlfind" {>= "1.7.3"}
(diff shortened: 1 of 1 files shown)


jhwoodyatt:
I believe the failing checks are infrastructure isses.

- The OMake package needs a version update to be ready for OCaml effects.
- The macOS `homebrew` configurations do not seem to have the auxiliary `unix.a` libraries properly installed.

kit-ty-kate:
```

**Comment by kit-ty-kate** [comment]
It looks like `conjury`’s build system is not compatible with system compilers (`ocaml-system`)
```
-    File Conjury/Basis/Driver.om: line 75, characters 16-78
-    Do not know how to build "/Users/mac705/.opam/default/lib/ocaml/unix.a" required for "stage/develop/tests/ocaml/ounit/alfa_test.component/alfa_test.cmx"
```
With such setup `opam var ocaml:lib` (used in https://bitbucket.org/jhw/conjury/src/3e1f9d4cc131b5f579470351ccb30d2635bc902f/Conjury/V3/Language/OCaml/Testing/OUnit.om?at=principal#lines-48) does not exist. `ocamlc -where` should be used instead.

jhwoodyatt:
> With such setup opam var ocaml:lib [...] does not exist. ocamlc -where should be used instead.

Ah! I see. Thanks for finding that!!

jhwoodyatt:
Looks like it's just a transient failures to download source distributions from Bitbucket now.

jhwoodyatt:
So at this point, the only failure remaining is the one caused by OMake not being ready for the `effect` keyword. I think this can be safely merged, and I hope you'll agree.

kit-ty-kate:
yes indeed. Thanks!

PR #20343, opened 2021-12-31
ocamlnet: request more paths from pkg-config

madroach:
this is a work-around for the "broken" ports pkgconf, which is forced
inte the PATH by conf-pkg-config.

OpenBSD has a custom-tailored /usr/bin/pkg-config.

The full-featured ports /usr/local/bin/pkgconf is installed via ports. But this thing is too smart. It strips -L/usr/local/lib from its recommended options.

Adding the --keep-system option will make it less smart and more helpful.

This whole problem occurs only because conf-pkg-config adds a symlink to the opam path to enforce use of the smart and full-featured ports pkgconf. The full featured pkgconf seems to be required by mirage.

### DIFF
--- a/packages/ocamlnet/ocamlnet.4.1.9-1/files/pkg-config.patch
+++ b/packages/ocamlnet/ocamlnet.4.1.9-1/files/pkg-config.patch
@@ -1,0 +1,13 @@
+diff --git code/src/nettls-gnutls/configure code/src/nettls-gnutls/configure
+index 247ee47..341231a 100755
+--- code/src/nettls-gnutls/configure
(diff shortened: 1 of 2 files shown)


mseri:
Thanks

gerdstolpmann:
This breaks ocamlnet on macos!

mseri:
Really? It worked fine when I tested on my machine, do you use macports or homebrew? Maybe we need to make it more specific

gerdstolpmann:
see full report on https://github.com/ocaml/opam-repository/issues/20387

PR #20357, opened 2022-01-02
Package mad.0.5.1

toots:
### `mad.0.5.1`
Mad decoding library
Bindings for the mad library which provides functions for encoding wave audio files into mp3

---
* Homepage: https://github.com/savonet/ocaml-mad
* Source repo: git+https://github.com/savonet/ocaml-mad.git
* Bug tracker: https://github.com/savonet/ocaml-mad/issues

---


mseri:
It is weird that it fails on archlinux, since it ships the pc files, but it can be fixed separately if we find a fix

PR #20358, opened 2022-01-02
Add liquidsoap.2.0.2-1 with bugfix patch, switch to s3 hosted tarball.

toots:


packages/liquidsoap/liquidsoap.2.0.2-1/opam
```diff
@@ -0,0 +1,183 @@
...
+authors: "The Savonet Team <savonet-users@lists.sourceforge.net>"
+license: "GPL-2.0-or-later"
+homepage: "https://github.com/savonet/liquidsoap"
+bug-reports: "https://github.com/savonet/liquidsoap/issues"
+patches: [
+  "patches/fix-last-metadata.diff"
+]
+extra-files: [
+  ["patches/fix-last-metadata.diff" "md5=bcd8f7659e1e935233e025bd39728670"]
+]
```
mseri, line 20:
```
Error in liquidsoap.2.0.2-1:              error 53: Mismatching 'extra-files:' field: "patches/fix-last-metadata.diff", "patches"
```
Afaik, the [extra-files](https://opam.ocaml.org/doc/Manual.html#opamfield-extra-files) and [patches](https://opam.ocaml.org/doc/Manual.html#opamfield-patches) must be all in the files folder, subfolders are not allowed.

There is also a permissions issue:
```
2022-01-02 17:04.22: Error in liquidsoap.2.0.2-1: Forbidden permission for file packages/liquidsoap/liquidsoap.2.0.2-1/files/patches. All files should have permissions 644.
```

toots: Thanks for looking into this. Changes applied!

mseri:
No idea why macos-homebrew fails to find the patch, but I think is a CI problem, not a problem of this package. Let me ping @kit-ty-kate, she may know more

I think this can be merged

kit-ty-kate:
no idea either. cc @patricoferris ?

toots:
@kit-ty-kate @mseri I can confirm that the package works on macos.

kit-ty-kate:
Thanks

patricoferris:
Hmm strange, nothing immediately obvious but it appears this kind of bug (macOS builders not finding anything in `files`) has happened before (e.g. https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/b8381d465c42be77848641b202c0aaa58e607d08/variant/opam-2.1,distributions,macos-homebrew,lablgtk3.3.1.2 -- both failures using `opam.2.1`?) -- I'll keep digging, thanks for reporting it :))

PR #20366, opened 2022-01-03
new release ocaml-freestanding.0.7.0

dinosaure:
## v0.7.0 (2022-01-03)

* **NOTE**: This release is a part of the MirageOS 4.0 release and, at this stage, `ocaml-freestanding.0.7.0` will **not** continue to support MirageOS 3 unikernels. `ocaml-freestanding` becomes a real cross-compiler used then by `dune` (with the cross-compilation option) to compile libraries including C stubs. For the MirageOS 4.0 perspective, libraries which include C stubs does not need anymore to compile multiple times C artifacts depending on what `ocaml-freestanding` provides. The `dune` context will help the compilation of these C stubs according the target chosen by the end-user (with the `mirage` tool).

* Add compatibility with the solo5 0.7.0 package split. (mirage/ocaml-freestanding#104)
* Build a freestanding cross-compiler to use with the ocamlfind toolchain feature. This cross-compiler is able to build partial executables to link with a solo5 bindings library. (mirage/ocaml-freestanding#104)
* Fix the OpenBSD 7 support (mirage/ocaml-freestanding#104)
* Remove `pkg-config` (mirage/ocaml-freestanding#104)

### DIFF
--- a/packages/ocaml-freestanding-cross-aarch64/ocaml-freestanding-cross-aarch64.0.7.0/opam
+++ b/packages/ocaml-freestanding-cross-aarch64/ocaml-freestanding-cross-aarch64.0.7.0/opam
@@ -10,12 +10,9 @@ build: [
   ["./configure.sh"
    "--prefix=%{prefix}%"
    "--target=aarch64-solo5-none-static"
-   "--ocaml-configure-option=--with-afl" {ocaml-option-afl:installed}
    "--ocaml-configure-option=--disable-flat-float-array" {ocaml-option-no-flat-float-array:installed}
    "--ocaml-configure-option=--enable-flambda" {ocaml-option-flambda:installed}
-   "--ocaml-configure-option=--enable-frame-pointers" {ocaml-option-fp:installed}
    "--ocaml-configure-option=--disable-naked-pointers" {ocaml-option-nnp:installed}
-   "--ocaml-configure-option=--enable-naked-pointers-checker" {ocaml-option-nnpchecker:installed}
   ]
   [make "-j%{jobs}%"]
 ]
--- a/packages/ocaml-freestanding-cross-aarch64/ocaml-freestanding-cross-aarch64.0.7.0/opam
+++ b/packages/ocaml-freestanding-cross-aarch64/ocaml-freestanding-cross-aarch64.0.7.0/opam
@@ -1,0 +1,48 @@
+opam-version: "2.0"
+maintainer: "Martin Lucina <martin@lucina.net>"
+authors: "Martin Lucina <martin@lucina.net>"
+homepage: "https://github.com/mirage/ocaml-freestanding"
+bug-reports: "https://github.com/mirage/ocaml-freestanding/issues/"
+license: "MIT"
+tags: "org:mirage"
+dev-repo: "git+https://github.com/mirage/ocaml-freestanding.git"
+build: [
+  ["./configure.sh"
+   "--prefix=%{prefix}%"
+   "--target=aarch64-solo5-none-static"
+   "--ocaml-configure-option=--with-afl" {ocaml-option-afl:installed}
+   "--ocaml-configure-option=--disable-flat-float-array" {ocaml-option-no-flat-float-array:installed}
+   "--ocaml-configure-option=--enable-flambda" {ocaml-option-flambda:installed}
+   "--ocaml-configure-option=--enable-frame-pointers" {ocaml-option-fp:installed}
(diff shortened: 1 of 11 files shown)


dinosaure:
We probably need to add constraints on:
- `gmp-freestanding`
- `zarith-freestanding`

/cc @TheLortex, @hannesm

TheLortex:
What about the other revdeps, such as `bigstringaf`, `checkseum`, `digestif` ? They should also get the constraints

dinosaure:
> What about the other revdeps, such as bigstringaf, checkseum, digestif ? They should also get the constraints

These packages are more smart than `zarith-freestanding`/`gmp-freestanding`. If `pkg-config` does not exists, they just `touch libfreestanding_*.a` and do an usual installation. Then, put a constraint on them will require a release of them then without their C artifacts (which is much more work). I think the situation is good enough for them - and we will just waiting another release of them to delete their `freestanding` supports.

hannesm:
yes, ti me this looks good -- including upper bounds on gmp-freestanding and zarith-freestanding.

Maybe worth to investigate the extras:
```
flambda-fp

    ocaml-freestanding-cross-aarch64.0.7.0 (failed: frame pointers not supported on this platform)
    ocaml-freestanding.0.7.0 (failed: frame pointers not supported on this platform)

fp

    ocaml-freestanding-cross-aarch64.0.7.0 (failed: frame pointers not supported on this platform)
    ocaml-freestanding.0.7.0 (failed: frame pointers not supported on this platform)

nnpchecker

    ocaml-freestanding-cross-aarch64.0.7.0 (failed: naked pointers checker not supported on this platform)
    ocaml-freestanding.0.7.0 (failed: naked pointers checker not supported on this platform)
```

dinosaure:
`solo5-bindings-*` requires at least GCC 4.0 but I don't know how to check that - if we have a `conf-gcc` package or something like that. Then, as @hannesm said, `flambda-fp`, `fp` and `nnpchecker` seems accurate errors about this release - others errors seems unrelated.

TheLortex:
> It seems however that we have a problem on Archlinux:

It's a bit strange because it's an exception that comes from `obuilder`

dinosaure:
The current test to check if we handle frame-pointer is:
```sh
  case "$host,$CC" in #(
  x86_64-*-linux*,gcc*|x86_64-*-linux*,clang*) :
    common_cflags="$common_cflags -g  -fno-omit-frame-pointer"
      frame_pointers=true
      $as_echo "#define WITH_FRAME_POINTERS 1" >>confdefs.h

      { $as_echo "$as_me:${as_lineno-$LINENO}: using frame pointers" >&5
$as_echo "$as_me: using frame pointers" >&6;} ;; #(
  *) :
    as_fn_error $? "frame pointers not supported on this platform" "$LINENO" 5

```

Which obviously fails because the compiler is: `x86_64-solo5-none-static`. Unfortunately, we can not lie on this side - we must keep our toolchain. The best will be to say that `ocaml-freestanding` is not allowed with a compiler with frame-pointers/naked-pointers.

hannesm:
what we could do is to add `-fno-omit-frame-pointer` to the cflags if desired. but until anyone requests this feature, we can also just remove the `fp` (see https://github.com/hannesm/ocaml-freestanding/commit/c47deb49bba0cb59b4fbe2b105bdb359ecda8f4e). the afl option isn't worth it, since with freestanding we're not supporting the C runtime (runtime/afl.c from OCaml) part for afl communication.

packages/ocaml-freestanding-cross-aarch64/ocaml-freestanding-cross-aarch64.0.7.0/opam
```diff
@@ -33,6 +33,9 @@ conflicts: [
   "solo5-kernel-ukvm"
   "solo5-kernel-virtio"
   "solo5-kernel-muen"
+  "ocaml-option-fp"
+  "ocaml-option-nnp"
```
hannesm, line 37: the nnp option is fine afaict. the nnpchecker is not fine.

dinosaure:
This PR seems ready to merge. About `zarith-freestanding`/`gmp-freestanding`:
1) we don't use them anymore for MirageOS 4.0
2) I don't know how to constraint them (and specially `solo5-bindings-*` packages) to require, at least, `gcc.4.*`

hannesm:
I looked through the failures, and indeed they are unrelated -- some sandbox issues (9p gets EROFS), some other issues in packages -- but nothing related to the new ocaml-freestanding. I fully support a merge of this PR.

PR #20356, opened 2022-01-02
Package opus.0.2.1

toots:
### `opus.0.2.1`
Bindings to libopus

---
* Homepage: https://github.com/savonet/ocaml-opus
* Source repo: git+https://github.com/savonet/ocaml-opus.git
* Bug tracker: https://github.com/savonet/ocaml-opus/issues

---
:camel: Pull-request generated by opam-publish v2.1.0

PR #20379, opened 2022-01-04
conf-libopus: update suse library name

mseri:
Should fix one of the failures in https://github.com/ocaml/opam-repository/pull/20356/files


mseri:
Still failing but at least the library is correctly installed now

Issue #20387, opened 2022-01-05
ocamlnet cannot be installed anymore on Linux and macos


gerdstolpmann:
Looks like #20343 broke it. The symptom is that any attempt to use `nettls-gnutls` results in lots of errors like
```
/usr/bin/ld: /home/bb/.opam/4.12.0/lib/nettls-gnutls/libnettls-gnutls.a(nettls_gnutls_bindings_stubs.o): in function `net_gnutls_pubkey_verify_data2':
/home/bb/.opam/4.12.0/.opam-switch/build/ocamlnet.4.1.9-1/src/nettls-gnutls/nettls_gnutls_bindings_stubs.c:7536: undefined reference to `gnutls_pubkey_verify_data2
```
which is because `-lgnutls` is not passed to the system linker anymore.

`pkg-config` doesn't have `--keep-system-cflags --keep-system-libs`:

```
$ pkg-config --keep-system-cflags gnutls
Unknown option --keep-system-cflags
$ pkg-config --keep-system-libs gnutls
Unknown option --keep-system-libs
```

Can you please fix by checking whether the installed version of `pkg-config` supports these flags? Thanks.

mseri:
Thanks for the report, both my test on macos+macports and the CI checks had successfully completed so I did not expect this failure when I merged.

I have reverted the patch for now and asked the author to submit a new one with more careful checks on these flags

PR #20388, opened 2022-01-05
Revert ocamlnet patch and publish new patched version for openbsd

mseri:
This reverts commit e9a8729495a4f4a2094f8c0bf602fd92d46ab0e8.

The fix breaks some configurations: https://github.com/ocaml/opam-repository/pull/20343#issuecomment-1005652283
Should fix https://github.com/ocaml/opam-repository/issues/20387


mseri:
@madroach would you mind submitting a new PR where the patch checks first that the flags can be used?

mseri:
Thanks for the report, both my test on macos+macports and the CI checks had successfully completed so I did not expect this failure when I merged.

I have reverted the patch for now and asked the author to submit a new one with more careful checks on these flags

mseri:
Also, please add the patch on `ocamlnet.4.1.9-2`.
I mistakenly thought that the `-1` was already a new patched version when I revised the PR, apologies

mseri:
@madroach I have added here a patch that should work. Can you double check if it fixes your issue?

@gerdstolpmann Can you check that with the patch I have added here it does work on your system?

madroach:
Hi, thanks for improving my patch.
But this is getting out of hand. conf-pkg-config caters for special needs of mirage by forcing a ported pkgconf into my `$PATH`, thereby breaking ocamlnet. Now ocamlnet needs to specially instruct this non-default ported pkgconf so that it works again.
I would very much prefer removing all those OpenBSD "fixes" and moving them to mirage.

madroach:
otherwise setting `PKG_CONFIG_ALLOW_SYSTEM_LIBS` and `PKG_CONFIG_ALLOW_SYSTEM_CFLAGS` would do the job, too.

mseri:
Thanks, I like this alternative more.
As per mirage, can you please open an issue in the corresponding repo?

mseri:
@kit-ty-kate do you know why the patch does not apply? If I run `patch -p1 -i pkg-config.patch` manually, it works without problems

madroach:
```
#=== ERROR while compiling ocamlnet.4.1.9-2 ===================================#
These patches didn't apply at /home/madroach/opam/4.13.1/.opam-switch/build/ocamlnet.4.1.9-2:
  - pkg-config.patch: "/usr/local/bin/gpatch -p1 -i /home/madroach/opam/log/processed-patch-86368-015123" exited with code 1
```

this does apply and does build fine on OpenBSD:
https://github.com/madroach/opam-repository/commit/3e2729a785da6106df9f35f2385676c7dbe4c367

madroach:
You are missing an additional directory level in your diff. It should apply without the `-p1`.
```
$git diff master..mseri/ocamlnet -- . |cat
diff --git packages/ocamlnet/ocamlnet.4.1.9-2/files/pkg-config.patch packages/ocamlnet/ocamlnet.4.1.9-2/files/pkg-config.patch
index c4ba014c60..afe79fdd0b 100644
--- packages/ocamlnet/ocamlnet.4.1.9-2/files/pkg-config.patch
+++ packages/ocamlnet/ocamlnet.4.1.9-2/files/pkg-config.patch
@@ -1,14 +1,17 @@
-diff --git code/src/nettls-gnutls/configure code/src/nettls-gnutls/configure
-index 247ee47..35025f7 100755
---- code/src/nettls-gnutls/configure
-+++ code/src/nettls-gnutls/configure
-@@ -10,6 +10,9 @@ have_gnutls=0
-
+diff --git a/code/src/nettls-gnutls/configure b/code/src/nettls-gnutls/configure
+index 247ee47..65833db 100755
+--- a/code/src/nettls-gnutls/configure
++++ b/code/src/nettls-gnutls/configure
+@@ -9,7 +9,10 @@
+ have_gnutls=0
+
  if [ -z "$PKG_CONFIG" ]; then
-     PKG_CONFIG="pkg-config"
-+    # Workaround issue on OpenBSD
-+    export PKG_CONFIG_ALLOW_SYSTEM_LIBS=1
-+    export PKG_CONFIG_ALLOW_SYSTEM_CFLAGS=1
+-    PKG_CONFIG="pkg-config"
++   PKG_CONFIG="pkg-config"
++   # Workaround issue on OpenBSD
++   export PKG_CONFIG_ALLOW_SYSTEM_LIBS=1
++   export PKG_CONFIG_ALLOW_SYSTEM_CFLAGS=1
  fi
-
+
  if [ -z "$GNUTLS_LIBS" ]; then
+
diff --git packages/ocamlnet/ocamlnet.4.1.9-2/opam packages/ocamlnet/ocamlnet.4.1.9-2/opam
index 3a33b3a204..264c61c8fa 100644
--- packages/ocamlnet/ocamlnet.4.1.9-2/opam
+++ packages/ocamlnet/ocamlnet.4.1.9-2/opam
@@ -57,7 +57,7 @@ conflicts: [
 extra-files: [
   ["ocamlnet.install" "md5=ed54a9f3d6382ccc01ea1cf1af8f2c38"]
   ["fix-version-number.patch" "md5=23c41def93bfc8bdbd9805a36849fe74"]
-  ["pkg-config.patch" "md5=2dd96f09350d398ed9798791174a9685"]
+  ["pkg-config.patch" "md5=4c1352c8a4d9c34595dc914c5d0f15fe"]
 ]
 url {
   src: "http://download.camlcity.org/download/ocamlnet-4.1.9.tar.gz"
```

packages/ocamlnet/ocamlnet.4.1.9-2/files/pkg-config.patch
```diff
@@ -0,0 +1,17 @@
+diff --git a/code/src/nettls-gnutls/configure b/code/src/nettls-gnutls/configure
+index 247ee47..65833db 100755
+--- a/code/src/nettls-gnutls/configure
++++ b/code/src/nettls-gnutls/configure
```
kit-ty-kate, line 4:
```suggestion
--- a/src/nettls-gnutls/configure
+++ b/src/nettls-gnutls/configure
```
easy. This patch has been made from the ocamlnet source repository and it does not have the same directory structure as the archive. The archive is basically the `code/` subdirectory from the source repository

madroach:
Aha, so I just got lucky and the `-p1` stripped the extra `code/` my diff had. funny.

madroach:
anyways please also note your diff adds trailing whitespace and changes indentation.

madroach:
but when applied it allows ocamlnet to build. So I'm ok with it.

packages/ocamlnet/ocamlnet.4.1.9-2/files/pkg-config.patch
mseri: Aaah, thanks a lot!

mseri:
New version should have the same indentation and no trailing spaces. Sorry for the mess

mseri:
Logs are looking fine. I am going to release some pressure from the CI

madroach:
Thank you :)

hannesm:
@madroach wrote
> But this is getting out of hand. conf-pkg-config caters for special needs of mirage by forcing a ported pkgconf into my $PATH, thereby breaking ocamlnet.

This has been the case, indeed. Due to some trouble with the system provided one. Fortunately with MirageOS 4 we don't rely on pkg-config anymore.

What could be done:
- develop a conf-pkg-config.3 which doesn't install any symlinks
- add upper bounds "conf-pkg-config" {< "3"} to existing solo5* and ocaml-freestanding packages
- require conf-pkg-config >= 3 in e.g. ocamlnet

Unfortunately I don't have the time to do the above, but indeed getting rid of the symlink would be great for normal pkg-config use on OpenBSD of opam packages.

gerdstolpmann:
I can confirm that 4.1.9-2 works for me. Thanks for the quick reaction!

PR #20396, opened 2022-01-06
add packge sd_logic

zevbo:


packages/sd_logic/sd_logic.0.1/opam
```diff
@@ -0,0 +1,38 @@
+# This file is generated by dune, edit dune-project instead
+opam-version: "2.0"
+synopsis: "Robot State!"
+description: """
+
+Longer description: https://github.com/zevbo/RobotState
```
kit-ty-kate, line 6: Would it be possible to have more meaningful `synopsis` and `description` here? It is hard to understand what the package does from either one of them.

zevbo: Yeah sure no problem

zevbo: I just added a synopsis and description. Let me know if it's good or it needs more/to be editted :)

kit-ty-kate:
Thanks! Could you return the improvements to the opam file in the source repository so they can be present for the next release?

PR #20397, opened 2022-01-07
2 packages from pdarragh/camlrack at 0.9.0

pdarragh:
This pull-request concerns:
-`camlrack.0.9.0`: S-Expression parsing for OCaml
-`ppx_camlrack.0.9.0`: PPX for matching S-Expressions

---
* Homepage: https://github.com/pdarragh/camlrack
* Source repo: git+https://github.com/pdarragh/camlrack.git
* Bug tracker: https://github.com/pdarragh/camlrack/issues

---


packages/ppx_camlrack/ppx_camlrack.0.9.0/opam
```diff
@@ -0,0 +1,45 @@
...
+license: "MIT"
+tags: ["S-Expressions" "parsing" "matching" "ppx" "extension"]
+homepage: "https://github.com/pdarragh/camlrack"
+doc: "https://github.com/pdarragh/camlrack"
+bug-reports: "https://github.com/pdarragh/camlrack/issues"
+depends: [
+  "dune" {>= "2.9"}
+  "ocaml" {>= "4.12"}
+  "ppxlib" {>= "0.23.0"}
+  "camlrack" {>= "0.9.0"}
```
mseri, line 19:
```suggestion
  "camlrack" {= version}
```
For ease of maintenance, I would suggest to bind the two version numbers and always release the two packages together

packages/ppx_camlrack/ppx_camlrack.0.9.0/opam
```diff
@@ -0,0 +1,45 @@
...
+  "odoc" {with-doc}
+]
```
mseri, line 21:
```suggestion
  "ounit"
]
```

packages/ppx_camlrack/ppx_camlrack.0.9.0/opam
```diff
@@ -0,0 +1,45 @@
...
+tags: ["S-Expressions" "parsing" "matching" "ppx" "extension"]
+homepage: "https://github.com/pdarragh/camlrack"
+doc: "https://github.com/pdarragh/camlrack"
+bug-reports: "https://github.com/pdarragh/camlrack/issues"
+depends: [
+  "dune" {>= "2.9"}
+  "ocaml" {>= "4.12"}
+  "ppxlib" {>= "0.23.0"}
+  "camlrack" {>= "0.9.0"}
+  "odoc" {with-doc}
```
mseri, line 20:
```suggestion
  "ounit" {with-test}
  "odoc" {with-doc}
```

packages/camlrack/camlrack.0.9.0/opam
```diff
@@ -0,0 +1,43 @@
...
+maintainer: ["Pierce Darragh <pierce.darragh@gmail.com>"]
+authors: ["Pierce Darragh <pierce.darragh@gmail.com>"]
+license: "MIT"
+tags: ["S-Expressions" "parsing"]
+homepage: "https://github.com/pdarragh/camlrack"
+doc: "https://github.com/pdarragh/camlrack"
+bug-reports: "https://github.com/pdarragh/camlrack/issues"
+depends: [
+  "dune" {>= "2.9"}
+  "ocaml" {>= "4.12"}
```
mseri, line 17:
```suggestion
  "ocaml" {>= "4.12"}
  "ppx_camlrack" {= version & with-test & post}
```

mseri: otherwise you need to move all the tests on the ppx package

mseri: @kit-ty-kate it fails still with post, do we have a better solution?

mseri:
You need to add `(package ppx_camlrack)` to the runtest aliases in the ppx_camlrack tests, otherwise it tries to build them for camlrack as well

packages/camlrack/camlrack.0.9.0/opam
kit-ty-kate: `post` is not the solution here. It will just make the dependency install after this package (which is going to make it fail anyway).

The solution here is to modify the dune files in the source and add the correct `(package ...)` field to each `(test ...)`. By default all of them will be run regardless of the `-p` given.

pdarragh: So if I add the `(package ...)` to the tests, should I also do `"ppx_camlrack" {= version & with-test & post}`, or is that no longer necessary? (Sorry, I've never distributed a package on opam before.)

mseri: It depends. If it is `(package ppx_camlrack)` no, otherwise yes

pdarragh:
I believe I've addressed the testing issues, and I also made some other changes to the code in the meantime. Is the recommended course of action to run `opam publish` from my local repository again? I assume that will create a new pull request, in which case I would come here and manually close this one?

(Again, sorry for not knowing the system better, but the instructions [here](https://opam.ocaml.org/doc/Packaging.html) are not the most fully featured.)

mseri:
If you re-tag and run `opam-publish`, it will force push over this PR.
If for some reason it opens another PR, we can just close this one and keep going there, it is not a problem in either case

mseri:
Ping @pdarragh

pdarragh:
Hi @mseri, apologies for lack of communication. The semester picked up pretty quick and I lost track.

Looks like issuing a new tag followed by `opam publish` caused a new pull request at #20704, so I'll close this one!

PR #20433, opened 2022-01-12
ppx_factory.0.2.0: Remove upper bound on OCaml version

bbc2:
I don't think there was a good reason to put an upper bound in the first place and tests pass with OCaml 4.12 and 4.13 (https://github.com/cryptosense/ppx_factory).


kit-ty-kate:
thanks

PR #20438, opened 2022-01-12
release lambdapi-2.1.0

fblanqui:


packages/lambdapi/lambdapi.2.1.0/opam
```diff
@@ -0,0 +1,82 @@
...
+the Why3 platform."""
+maintainer: ["dedukti-dev@inria.fr"]
+authors: ["Deducteam"]
+license: "CECILL-2.1"
+homepage: "https://github.com/Deducteam/lambdapi"
+bug-reports: "https://github.com/Deducteam/lambdapi/issues"
+depends: [
+  "dune" {>= "2.7"}
+  "ocaml" {>= "4.08.0"}
+  "menhir" {>= "20200624" & < "20211223"}
```
kit-ty-kate, line 47:
```suggestion
  "menhir" {>= "20200624"}
```
Is there a reason why this constraint is still there? I see you are using `(using menhir 2.0)` so it should be compatible with 20211223

fblanqui: I am not sure that why3 is compatible with the last version of menhir.
And why3 is not compatible with the last version of alt-ergo (used in some lambdapi tests).
See https://gitlab.inria.fr/why3/why3/-/issues/601.
and https://github.com/OCamlPro/alt-ergo/issues/492. It seems to have been fixed on the master branch of alt-ergo (https://github.com/OCamlPro/alt-ergo/pull/495) but not released yet.

mseri: The latest version of menhir came with some breaking changes that required us to add lots of upper bounds. All versions of alt-ergo currently have the upper bound, so for the moment you would get it implicitly.

My suggestion is to remove it from here, and once alt-ergo and why3 are updated, if we see that it breaks lambdapi, we will re-add it and inform you. What do you think?

fblanqui:
Ok, let's do this. Do you want me to commit the change?

fblanqui:
Wait, I am going to make a last-minute change in my release. So, I am going to update my PR.

fblanqui:
Hi @kit-ty-kate . Everything is ready now. :-)

kit-ty-kate:
Thanks!

PR #20439, opened 2022-01-12
uecc.0.2: conflict with dune 3

emillon:
This package does not have a `dune-project` file, so it will default to the latest one. Since there are changes to C flags in dune 3, this breaks uecc.0.2. This could be fixed upstream by adding a dune-project file with a dune-lang 2.0, but there is also a fixed version already.

See ocaml/dune#5268.


kit-ty-kate:
Thanks!

PR #20443, opened 2022-01-13
Dune3: add implicit "result" dependencies

emillon:
Dune < 3.0 ships with META files that implicitly define a "result" library. However, it is not the same as the one from opam (see https://github.com/ocaml/dune/pull/4946#issuecomment-928031670).

So an explicit dependency is still necessary for packages that refer to the result library.

See ocaml/dune#5263.


mseri:
The lower bounds succeeded as well as installation and test

PR #20454, opened 2022-01-14
Introduce base-nnp

dra27:
OCaml 5.00.0 makes `--disable-naked-pointers` mode the default. #20420 prevents the `ocaml-option-nnp` package from being specified for > 5.00.0. In some ways (as was done in the ocaml-multicore project's repo), the `ocaml-option-nnp` should be forced, but this isn't consistent with how the other `ocaml-option-` packages are used, or the base packages.

However, in changing this, packages which didn't support no-naked-pointers mode, and so which _conflicted_ `ocaml-option-nnp` are now installable on OCaml 5.00.0 (e.g. `llvm`) even though they presumably don't work.

This PR introduces `base-nnp.base` which is installed automatically by `ocaml-variants.5.00.0+trunk` (along with `base-unix.base`, etc.) and which is also pulled in if `ocaml-option-nnp` is selected for 4.12-4.14. Packages which conflicted `ocaml-option-nnp` are updated to conflict `base-nnp` instead.

### DIFF
--- a/packages/base-nnp/base-nnp.base/opam
+++ b/packages/base-nnp/base-nnp.base/opam
@@ -1,0 +1,10 @@
+opam-version: "2.0"
+maintainer: "https://github.com/ocaml/opam-repository/issues"
+synopsis: "Naked pointers prohibited in the OCaml heap"
+description: """
+Installed when the compiler does not permit naked pointers in
+the heap. Prior to OCaml 5.00.0, this mode was expressly selected
+by configuring with --disable-naked-pointers. The shared memory
+parallelism added in OCaml 5.00.0 requires this mode.
+"""
+depends: "ocaml" {>= "5.00"} | "ocaml-option-nnp" | "ocaml-variants" {= "4.06.1+no-naked-pointers+flambda"}
--- a/packages/base-nnp/base-nnp.base/opam
+++ b/packages/base-nnp/base-nnp.base/opam
@@ -7,4 +7,4 @@ the heap. Prior to OCaml 5.00.0, this mode was expressly selected
 by configuring with --disable-naked-pointers. The shared memory
 parallelism added in OCaml 5.00.0 requires this mode.
 """
-depends: "ocaml" {>= "5.00"} | "ocaml-option-nnp" | "ocaml-variants" {= "4.06.1+no-naked-pointers+flambda"}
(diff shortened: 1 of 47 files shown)


packages/base-nnp/base-nnp.base/opam
```diff
@@ -0,0 +1,10 @@
+opam-version: "2.0"
+maintainer: "https://github.com/ocaml/opam-repository/issues"
+synopsis: "Naked pointers prohibited in the OCaml heap"
+description: """
+Installed when the compiler does not permit naked pointers in
+the heap. Prior to OCaml 5.00.0, this mode was expressly selected
+by configuring with --disable-naked-pointers. The shared memory
+parallelism added in OCaml 5.00.0 requires this mode.
+"""
+depends: "ocaml" {>= "5.00"} | "ocaml-option-nnp" | "ocaml-variants" {= "4.06.1+no-naked-pointers+flambda"}
```
kit-ty-kate, line 10:
```suggestion
depends: "base-domains" | "ocaml-option-nnp" | "ocaml-variants" {= "4.06.1+no-naked-pointers+flambda"}
```

kit-ty-kate: What about adding `ocaml-option-nnpchecker` as well. I know it’s not really "requiring" no-naked-pointers but if you want to constrain against one, then the other should also probably come with it.

dra27: I don't understand what you're suggesting here?

kit-ty-kate: Something like that:
```suggestion
depends: "base-domains" | "ocaml-option-nnp" | "ocaml-option-nnpchecker" | "ocaml-variants" {= "4.06.1+no-naked-pointers+flambda"}
```

dra27: That's a contradiction - `ocaml-option-nnpchecker` and `ocaml-option-nnp` can't be installed at the same time? The NNP checker only works if you're _in_ no-naked-pointers mode.

kit-ty-kate: yes, I know. Internally it’s not in no-naked-pointers mode, but for users it’s the same thing given if it doesn’t work with one, it won’t work with the other either.

kit-ty-kate:
Thanks!

PR #20461, opened 2022-01-15
Fix z3 4.8.5 package

R1kM:
cf issues https://github.com/Z3Prover/z3/issues/5437 and https://github.com/FStarLang/FStar/issues/2344


R1kM:
I am a bit confused by the error messages returned by CI.
The MacOS failure seems to be because the hash of the patch does not match, is the hash computed differently on MacOS than on other platforms?
The Alpine failures seem to be related to tests taking too long and being cancelled, is there a way to extend the timeout for this?

patricoferris:
Re:macOS failure -- we've seen the macOS builders fail before relating to patches not working so I think this is a bug in the builder and not with your package, apologies for the confusion and noise.

PR #20447, opened 2022-01-14
Package binaryen.0.11.0

phated:
### `binaryen.0.11.0`
OCaml bindings for Binaryen

---
* Homepage: https://github.com/grain-lang/binaryen.ml
* Source repo: git+https://github.com/grain-lang/binaryen.ml.git
* Bug tracker: https://github.com/grain-lang/binaryen.ml/issues

---
### ⚠ BREAKING CHANGES

- Switch to libbinaryen dependency (#113)
- Remove MacOS-specific library flags (#111)
- Require OCaml 4.12 (#108)

### Features

- Depend on libbinaryen ([59874ce](https://www.github.com/grain-lang/binaryen.ml/commit/59874ce785288d4a33ed5408ac4a0a33e16ff4c1))
- Remove MacOS-specific library flags ([#111](https://www.github.com/grain-lang/binaryen.ml/issues/111)) ([5a67f9f](https://www.github.com/grain-lang/binaryen.ml/commit/5a67f9f330332f8957976841819707d09124dc94))
- Require OCaml 4.12 ([#108](https://www.github.com/grain-lang/binaryen.ml/issues/108)) ([36eedea](https://www.github.com/grain-lang/binaryen.ml/commit/36eedea279fb19ee6131b72b21c24ba0a9d7f993))
- Switch to libbinaryen dependency ([#113](https://www.github.com/grain-lang/binaryen.ml/issues/113)) ([59874ce](https://www.github.com/grain-lang/binaryen.ml/commit/59874ce785288d4a33ed5408ac4a0a33e16ff4c1))

---
:camel: Pull-request generated by opam-publish v2.0.3

PR #20463, opened 2022-01-15
Lock binaryen bindings to 4.12

phated:
In #20447, @kit-ty-kate "unlocked" ocaml 4.13; however, I had thought I locked this down for a specific reason.

When trying to apply this [same patch upstream](https://github.com/grain-lang/binaryen.ml/pull/121), it was apparent that the reason for this is that ocaml 4.13 + windows explodes in CI: https://github.com/grain-lang/binaryen.ml/runs/4829181505?check_suite_focus=true

Attached is the log if you aren't allowed to see it in the GitHub Actions UI.
[error.log](https://github.com/ocaml/opam-repository/files/7876105/error.log)

If I'm doing something wrong, please let me know, but I think we should lock ocaml to 4.12 for the time being.

### DIFF
--- a/packages/binaryen/binaryen.0.11.0/opam
+++ b/packages/binaryen/binaryen.0.11.0/opam
@@ -6,7 +6,7 @@ license: " Apache-2.0"
 homepage: "https://github.com/grain-lang/binaryen.ml"
 bug-reports: "https://github.com/grain-lang/binaryen.ml/issues"
 depends: [
-  "ocaml" {>= "4.12"}
+  "ocaml" {>= "4.12" < "4.13"}
   "dune" {>= "2.9.1"}
   "dune-configurator" {>= "2.9.1"}
   "js_of_ocaml" {>= "3.10.0"}


packages/binaryen/binaryen.0.11.0/opam
```diff
@@ -6,7 +6,7 @@ license: " Apache-2.0"
 homepage: "https://github.com/grain-lang/binaryen.ml"
 bug-reports: "https://github.com/grain-lang/binaryen.ml/issues"
 depends: [
-  "ocaml" {>= "4.12"}
+  "ocaml" {>= "4.12" < "4.13"}
```
kit-ty-kate, line 9: If it’s windows specific what about this?
```suggestion
  "ocaml" {>= "4.12"}
  "ocaml" {< "4.13" & os = "win32"}
```

phated: I actually don't think esy supports that syntax 🤔 Do you know if it does?

kit-ty-kate:
also cc @dra27 who might have an idea of what’s going on windows

packages/binaryen/binaryen.0.11.0/opam
kit-ty-kate: it adds the `< 4.13` constraint if the current OS is windows

phated: Esy doesn't apply the OS to the constraint system. I had a PR open for it, but it can't be accepted because those can't be written into the lockfile. https://github.com/esy/esy/pull/1349

dra27: From opam's perspective, this package doesn't work properly on Windows/4.12 either as far as I can tell - it's only working at all in setup-ocaml because it's pulling in tools from unrelated distros? I'd confirm that 4.12 works on a stable set-up first on Windows before digging into the 4.12/4.13 differences.

phated: That's not really useful help. We definitely work with esy on Windows.

phated:
Since the only help I generally receive regarding Windows is "figure it out", I'm just dropping Windows support from libbinaryen and binaryen.ml. These bindings work inside esy, which is the way that Grain consumes them and we were just trying to contribute back to the community. I'm removing the CI now and will add `available: os != win32` in the next major release.

I'd suggest that setup-ocaml be the happy-path for opam/ocaml Windows integration and not "some weird distro" as they aren't crazily setup machines and that CI driven development is going to be how most people contribute their Windows support.

packages/binaryen/binaryen.0.11.0/opam
dra27: Where are you pulling cmake and python 3 from?

dra27:
setup-ocaml _is_ the “some weird distro”?

phated:
> setup-ocaml is the “some weird distro”?

That's what I'm taking away from your message. We are using setup-ocaml as directed. There's nothing out of the ordinary in https://github.com/grain-lang/binaryen.ml/blob/master/.github/workflows/opam.yml

dra27:
I'm puzzled as to where this is being "used" - is the GitHub Actions Windows Runner your sole Windows user?

I have a hunch about what might be affecting the 4.13 linking. I know what changed in OCaml because I maintain that part of OCaml, but I'm not willing to debug a problem I've never seen before where I have to submit everything to a GitHub Actions runner to investigate it - life's too short! So I ensured my Docker images were up-to-date (while coffee was brewing) and attempted to install your repo, which didn't work and at which point I found https://github.com/fdopen/opam-repository-mingw/issues/101 which made me suspicious that 4.12 wasn't technically "working" either. All I was after is a local environment which worked and so asked for confirmation of a known 4.12 Windows environment which works.

> as they aren't crazily setup machines

Oh yes they are - no developer on earth should need this number of tools installed on their local machine 🙂 https://github.com/actions/virtual-environments/blob/main/images/win/Windows2022-Readme.md

phated:
> is the GitHub Actions Windows Runner your sole Windows user?

This is the only way I use to test it. I don't have a Windows machine with opam and I am the only person on the team that works on Windows support (or opam support, for that matter).

> which didn't work and at which point I found fdopen/opam-repository-mingw#101 which made me suspicious that 4.12 wasn't technically "working" either

Yes, fdopen said something about the mingw CMake but would never elaborate. The experience between the forked opam repository has been pretty miserable.

phated:
I've spent a few hours diving deeper into this and it seems that depexts aren't specified for `cygwinports` in `conf-python` nor `conf-cmake` (not even on the mingw repository fork). I'm looking into adding those. Additionally, the mingw fork doesn't install a C++ compiler by default, so I'm looking at that too. I think with those changes, it'll mostly work (just trying to solve a filename resolution issue right now).

phated:
@dra27 the `conf-cmake` and `conf-python-3` changes aren't published on the mingw fork yet, but this commit adds the depexts directly in the meantime: https://github.com/grain-lang/libbinaryen/commit/3b0e11acd2ffc10f82020a470ede6b2f0e3f8f09

phated:
And it looks like a similar issue with OCaml 4.13.1
```
/usr/lib/gcc/x86_64-w64-mingw32/11/../../../../x86_64-w
64-mingw32/bin/ld: test/libbinaryen_stubs.a(binaryen_stubs_types.o): in function
 `caml_binaryen_type_none':
/home/grain/libbinaryen/_build/default/test/binaryen_stubs_types.c:33: undefined
 reference to `__imp_Caml_state'
/usr/lib/gcc/x86_64-w64-mingw32/11/../../../../x86_64-w64-mingw32/bin/ld: test/l
ibbinaryen_stubs.a(binaryen_stubs_types.o): in function `alloc_BinaryenType':
/home/grain/libbinaryen/_build/default/test/binaryen_stubs_types.c:26: undefined
 reference to `__imp_caml_alloc_custom'
collect2: error: ld returned 1 exit status
** Fatal error: Error during linking
```
I keep seeing this `__imp_Caml_state` thing

PR #20464, opened 2022-01-16
conf-cmake: Add cygwinports depext installation

phated:
I'm not sure what the policy is for adding `cygwinports` depext. One of the problems with the libbinaryen package on Windows is that this isn't specified in the depext list and no cmake gets installed on Windows.

I noticed that `cygwinports` is at least specified in `conf-pkg-config`. Why doesn't it exist in more conf packages?

Ref discussion in #20463

### DIFF
--- a/packages/conf-cmake/conf-cmake.1/opam
+++ b/packages/conf-cmake/conf-cmake.1/opam
@@ -20,6 +20,7 @@ depexts: [
   ["devel/cmake"] {os = "openbsd"}
   ["devel/cmake"] {os = "netbsd"}
   ["devel/cmake"] {os = "dragonfly"}
+  ["cmake"] {os-distribution = "cygwinports"}
 ]
 synopsis: "Virtual package relying on cmake"
 description:


kit-ty-kate (approved):

kit-ty-kate:
> I noticed that `cygwinports` is at least specified in `conf-pkg-config`. Why doesn't it exist in more conf packages?

simply because nobody had done it before ^^' also upstream opam-depexts does not support any windows distribution and we don't have the CI infrastructure to test it yet (though very soon, see https://github.com/ocurrent/obuilder/pull/75 for part of the progress being done)

All the windows specific stuff are done in fdopen's forks and afaik very few upstreaming work was attempted.

Thanks a lot! ❤️

phated:
I'm testing this more thoroughly. It seems like like cygwinports is doing something weird with this, as it is trying to lookup `mingw64-x86_64-cmake` instead of just `cmake`. I get this error: `Package 'mingw64-x86_64-cmake' not found.`

phated:
@kit-ty-kate Okay, I figured it out. This needs to be prefixed with `system:` in order to avoid it adding a prefix.

PR #20467, opened 2022-01-16
conf-python-3: Add cygwinports depext installation

phated:
This adds a cygwinports depext for python3 in the `conf-python-3` package. It also uses the `system:` prefix (as mentioned in #20466).

I'm adding these depexts directly to my package to test them out before they are upstreamed into the mingw fork.


mseri:
Thanks

PR #20477, opened 2022-01-18
added packages/oranger/oranger.4.3.3/opam

UnixJunkie:
small bug correction: there was no default mtry value
while earlier versions or oranger had this feature
(you don't need to choose an mtry value in order to train
a model; however, for a production model it is
advized to try optimizing the mtry fraction).


mseri:
Thanks

PR #20488, opened 2022-01-19
Adds cpdf v2.5

johnwhitington:


johnwhitington:
Then the CamlPDF dependency has not been installed at all on those machines, it seems. The OPAM file uses the "version" idiom:

```
depends: [
   "ocaml" {>= "4.10.0"}
   "ocamlfind" {build}
   "camlpdf" {= version}
 ]
```

Perhaps something is going wrong with 'version'?

kit-ty-kate:
It’s camlpdf.cmxa that is missing. I found the issue. camlpdf’s Makefile uses `which`, which is not a standard posix tool and not available on those distributions: https://github.com/johnwhitington/camlpdf/blob/master/Makefile#L24

kit-ty-kate:
I’ve fixed it in https://github.com/ocaml/opam-repository/pull/20488/commits/9d4282269f5f061e0fcc898ae6b12006cdf54c8e

If you could avoid using `which` altogether in future releases that would be awesome as well (simply calling `ocamlopt -version` is an equally simple test to detect if it is there)

johnwhitington:
Thanks. Fixed for the future in  https://github.com/johnwhitington/camlpdf/commit/b56d0103ca863fad26668d1dc30027b54887d2c7

kit-ty-kate:
Thanks!

PR #20492, opened 2022-01-19
camlp5: disable testing for macos/homebrew

chetmurthy:
disable testing (and deps on conf-perl-* packages) on macos & homebrew, so camlp5 will install there.

### DIFF
--- a/packages/camlp5/camlp5.8.00.02/opam
+++ b/packages/camlp5/camlp5.8.00.02/opam
@@ -29,10 +29,10 @@ depends: [
   "ocaml" {>= "4.02" & < "4.14.0"}
   "conf-perl"
   "conf-perl-ipc-system-simple" {
-    os != "macos"
+    os-distribution != "homebrew"
   }
   "conf-perl-string-shellquote" {
-    os != "macos"
+    os-distribution != "homebrew"
   }
   "conf-diffutils" { with-test & os-distribution = "alpine" }
   "pcre" { with-test }
@@ -47,12 +47,12 @@ build: [
   [make "-C" "testsuite" "clean" "all-tests"] {
     with-test
   & ocaml:version >= "4.10.0"
-  & os != "macos"
+  & os-distribution != "homebrew"
   }
   [make "-C" "test" "clean" "all"] {
     with-test
   & ocaml:version >= "4.10.0"
-  & os != "macos"
+  & os-distribution != "homebrew"
   }
 ]
 install: [make "install"]
--- a/packages/camlp5/camlp5.8.00.02/opam
+++ b/packages/camlp5/camlp5.8.00.02/opam
@@ -29,12 +29,10 @@ depends: [
   "ocaml" {>= "4.02" & < "4.14.0"}
   "conf-perl"
   "conf-perl-ipc-system-simple" {
-    with-test
-  & os != "macos"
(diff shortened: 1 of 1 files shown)


mseri:
I suggest to use `os != "macos"` rather than `os-distribution != "homebrew" & os-distribution != "macports"`, otherwise if I have neither installed and try `opam install -t` it will also fail

packages/camlp5/camlp5.8.00.02/opam
```diff
@@ -28,8 +28,12 @@ bug-reports: "https://github.com/camlp5/camlp5/issues"
 depends: [
   "ocaml" {>= "4.02" & < "4.14.0"}
   "conf-perl"
-  "conf-perl-ipc-system-simple"
-  "conf-perl-string-shellquote"
+  "conf-perl-ipc-system-simple" {
+    os != "macos"
```
mseri, line 32:
```suggestion
    os-distribution != "homebrew"
```
And the same below. You have sent the fix for macports in the other pr

chetmurthy: Fair cop.  I had thought I'd just disable it for macos, but hey, it's worth seeing if tests will pass with macports.  Fair cop.

mseri: Is that only needed fr tests? If so, the with-test flag was correct and was fixing an incorrect opam file

chetmurthy: No, these two conf-perl-* packages are used in a script, "mkcamlp5", that is used to build executables.  That script isn't invoked during the build, so once installed, if these conf-perl-* packages aren't available, the script will fail.  For instance the package "GT" uses this script.

chetmurthy: There might be something wrong with the macos-homebrew machine(s) ?  The test for that platform has sat waiting for resources for a long time -- hours and hours.

mseri: Yeah. Feel feee to ignore them for now

mseri:
Ok but then remove the flag altogether on those two packages and just disable the tests on mac with homebrew. If you want you can add a post install message for homebrew users saying that they need to install those Perl packages manually for the package to work

chetmurthy:
> Ok but then remove the flag altogether on those two packages and just disable the tests on mac with homebrew. If you want you can add a post install message for homebrew users saying that they need to install those Perl packages manually for the package to work

I do have a post-install message (have had one for a long time) about this. I don't quite understand what you're suggesting.  The flag is on the two conf-perl-* packages b/c they're actually used by mkcamlp5, so they need to be installed.  On homebrew, they cannot be installed.

mseri:
Sorry, I had not noticed it. I am suggesting it to leave
```
  "conf-perl-ipc-system-simple" {
     os-distribution != "homebrew"
   }
   "conf-perl-string-shellquote" {
     os-distribution != "homebrew"
   }
```
As it was
```
  "conf-perl-ipc-system-simple"
   "conf-perl-string-shellquote"
```

chetmurthy:
> as it was

Oh, I see now what you mean.  fair enough.  Done.

mseri:
Thanks. I have seen your comment on the other Pr. This one is good

PR #20495, opened 2022-01-20
OCaml 4.14.0, first alpha release

Octachron:
This PR adds the base and variants+option package for the first alpha release of OCaml 4.14.0.

As discussed, I have added the `--docdir=%{doc}%/ocaml"` configure option to account for ocaml/ocaml#10669 .

Like with the 4.13.0~alpha1 packages, I have updated the `ocaml` meta-package constraints to allow installation.

### DIFF
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.4.14.0~alpha1/files/ocaml-base-compiler.install
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.4.14.0~alpha1/files/ocaml-base-compiler.install
@@ -1,0 +1,1 @@
+share_root: ["config.cache" {"ocaml/config.cache"}]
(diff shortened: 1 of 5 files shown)


kit-ty-kate:
Successfully tested locally. Thanks a lot! :tada:

dbuenzli:
What happend to the `--docdir` ? I wanted to `odig` the 4.14 release notes but got nothing: 
```
> ls $(opam var ocaml:doc)
ls: /Users/dbuenzli/.opam/4.14.0/doc/ocaml: No such file or directory
```

kit-ty-kate:
arf, mmh, it looks like the option disappeared from the opam file between the last rc and the final release :/
Could you open a PR to add it to the 4.14.0 packages and cc @Octachron and @dra27 in it for double-check?

dbuenzli:
> Could you open a PR to add it to the 4.14.0 packages

It's not obvious to me were that's supposed to happen (and a bit in a hurry right now): 

```
> opam info --raw ocaml.4.14.0 | grep build: -C 2
  [OCAML_TOPLEVEL_PATH = "%{toplevel}%"]
]
build: [
  "ocaml" "%{ocaml-config:share}%/gen_ocaml_config.ml" _:version _:name
]
```

dra27:
See https://github.com/ocaml/opam-repository/pull/21890.

PR #20499, opened 2022-01-20
add proper perl depext for macos/macports

chetmurthy:
Let's see if we can get "conf-perl" working correctly on macos.

I tried this as part of another PR, and somehow it broke tons of other things.  So trying this in isolation, maybe I can figure out what's wrong.  Or maybe it'll just work, b/c geez, it's so simple.


mseri:
Thanks

PR #20503, opened 2022-01-21
fix: add ocaml < 5.00.

fpottier:
These releases of fix use the old ephemeron API.


kit-ty-kate (approved):

fpottier:
Same constraint for some releases of Menhir (because Menhir includes fix).

kit-ty-kate:
Thanks!

PR #20511, opened 2022-01-21
Add new package for the 'E Theorem Prover'

MSoegtropIMC:
This is yet another of my "why not use opam for useful C++ software" packages.

It adds an opam package for the E Theorem prover, an automated theorem prover. This compiles and installs an executable, which communicates via a standard text file format for theorem provers.

Addition of this package as well as the meta information has been agreed on with the authors via email.

This is used in the Coq ecosystem. In case you believe this is not useful for the OCaml ecosystem, I can also add it to Coq's opam repo.


kit-ty-kate:
Thanks!

MSoegtropIMC:
Thanks for the quick merge!

PR #20515, opened 2022-01-22
Package camlp4.4.14+1

kit-ty-kate:
I’m thinking about not adding it’s `+system` counter-part this time. Are there any known reasons why the `+system` package would still be useful? cc @ygrek


ygrek:
idk really
I can see debian has 4.13 camlp4 in unstable, I imagine they will have 4.14 as well, but afaics they don't have any -camlp4-dev anymore in current stable. I guess it is reasonable to drop it and see if anyone complains

PR #20519, opened 2022-01-23
Remove use of the outdated hidden-version flag

kit-ty-kate:
`hidden-verison` was only available in some beta versions of opam 2.1.0.
It has been renamed into `avoid-version` in the actual release.

The trick here is to make those packages go through ocaml-beta instead if some users still use beta versions of opam 2.1 as `2.1 < 2.1.0~beta1 < 2.1.0`


dra27 (approved):
LGTM - I particularly like the `< "2.1.0"` tweak 🙂

PR #20520, opened 2022-01-23
cppffigen: an FFI generator for C++ libraries based on STL

chetmurthy:
This is the first release of cppffigen, a C++ FFI generator for C++ libraries that expose their APIs using STL collection classes and following the Google C++ API style guide.

### DIFF
--- a/packages/cppffigen/cppffigen.0.002/opam
+++ b/packages/cppffigen/cppffigen.0.002/opam
@@ -1,0 +1,24 @@
+opam-version: "2.0"
+version: "0.002"
+synopsis: "A C++ foreign-function-interface generator for Ocaml based on C++ STL Containers"
+maintainer: "Chet Murthy <chetsky@gmail.com>"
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Thanks

chetmurthy:
I noticed that this package got a version "~dev".  Is this because I didn't put a "Version:" entry in the opam file ?

chetmurthy:
Also, thank you for merging this so quickly!

kit-ty-kate:
> got a version "~dev"

what do you mean, where did you get that?

chetmurthy:
Oh hm, now that I think about it, just b/c it's merged, doesn't mean that it's been released, and I'd done an "install -t ." yesterday.  So maybe it's that.  Sorry, I suspect this was a false alarm!

chetmurthy:
yeah, that was it.  Sorry sorry.

PR #20525, opened 2022-01-24
Specify `--docdir` for the +trunk packages for 4.14 and 5.00

dra27:
Enables #10669 for the 4.14 and 5.00 trunk packages (already the specified for the 4.14.0 alpha)


kit-ty-kate:
Thanks!

PR #20537, opened 2022-01-25
Fixes #19243: There should be a conf package that installs adwaita-icon-theme-full

### LINKED ISSUE #19243: There should be a conf package that installs adwaita-icon-theme-full on Ubuntu
CoqIDE, at least, requires `adwaita-icon-theme-full`.  I'm not sure whether this means that we need a separate conf package, or if the existing conf package should be changed, but certainly one of these needs to happen.

_Originally posted by @JasonGross in https://github.com/coq/coq/issues/14769#issuecomment-896963037_

MSoegtropIMC:
This PR fixes #19243 following the discussion in the issue.

Please note that the original package has a bug - there is no "os-family ubuntu". The os-family of ubuntu is debian.

### DIFF
--- a/packages/conf-adwaita-icon-theme/conf-adwaita-icon-theme.2/opam
+++ b/packages/conf-adwaita-icon-theme/conf-adwaita-icon-theme.2/opam
@@ -1,0 +1,37 @@
+opam-version: "2.0"
+maintainer: "https://github.com/ocaml/opam-repository/issues"
+homepage: "https://github.com/GNOME/adwaita-icon-theme"
+bug-reports: "https://gitlab.gnome.org/GNOME/adwaita-icon-theme/issues"
+authors: "GNOME devs"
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
> Please note that the original package has a bug - there is no "os-family ubuntu". The os-family of ubuntu is debian.

No this is on purpose. `os-family = ubuntu` will be the case for distributions such as `Pop! OS`, `Mint`, …

MSoegtropIMC:
I see, so ubuntu is not ubuntu.

The original idea was that on ubuntu one installs both, adwaita-icon-theme and adwaita-icon-theme-full, but I guess the latter included the former. So I guess that your change in logic is OK.

As far as I know this works on homebrew (didn't change and I didn't hear complaints in the past 2 years) - shall I investigate the Ci failure?

kit-ty-kate:
This is due to an internal CI error (the macOS target is still experimental). cc @patricoferris

PR #20540, opened 2022-01-25
Add conf flag to more recent ocaml packages

dra27:
Follows up https://github.com/ocaml/opam-repository/pull/19428#issuecomment-1019811905


kit-ty-kate:
Thanks

PR #20480, opened 2022-01-18
3 packages from c-cube/printbox at 0.6

c-cube:
This pull-request concerns:
-`printbox.0.6`: Allows to print nested boxes, lists, arrays, tables in several formats
-`printbox-html.0.6`: Printbox unicode handling
-`printbox-text.0.6`: Text renderer for printbox, using unicode edges

---
* Homepage: https://github.com/c-cube/printbox/
* Source repo: git+https://github.com/c-cube/printbox.git
* Bug tracker: https://github.com/c-cube/printbox/issues/

---
:camel: Pull-request generated by opam-publish v2.1.0

PR #20551, opened 2022-01-26
add upper bounds for printbox 0.6

mseri:
Seen on https://github.com/ocaml/opam-repository/pull/20480
In all cases the error is
```
# Error: Unbound module PrintBox_text
```


mseri:
I don't plan to fix any other upper or lower bound as part of this PR

PR #20578, opened 2022-01-28
Update Camlp5 for compatibility with OCaml 4.14.0~alpha1.

chetmurthy:
There might need to be a further release if 4.14.0 changes
in significant ways, but this should cover all the changes
I'm aware of in the alpha.


kit-ty-kate:
> There might need to be a further release if 4.14.0 changes

The api camlp5 uses is not supposed to change after the first alpha (even the branching)

PR #20579, opened 2022-01-28
Update ppx_seq.0.2.0

hyphenrf:
Update features lower OCaml version bounds and perf annotations for compilers that support it.


kit-ty-kate:
Dune is not a build dependency (see https://github.com/ocaml/dune/issues/2147)
Ppxlib is not a build dependency either as it is used as library to create this library (ppxs are libraries) and thus will need to be rebuilt everytime ppxlib is.

kit-ty-kate:
Thanks! Could you return the fix in the opam file upstream so it can be taken into account for next time?

hyphenrf:
Thanks for the pointers @kit-ty-kate, I'm on it!

PR #20558, opened 2022-01-26
[new release] mirage, mirage-runtime, functoria and functoria-runtime (4.0.0+beta1)

dinosaure:
The MirageOS library operating system

- Project page: <a href="https://github.com/mirage/mirage">https://github.com/mirage/mirage</a>
- Documentation: <a href="https://mirage.github.io/mirage/">https://mirage.github.io/mirage/</a>

##### CHANGES:

Refactor build process to use [Dune](https://dune.build/) build system. The
motivation is to drop `ocamlbuild`-induced technical debt and to obtain
first-class support for _cross-compilation_. To learn more about how Dune is
able to perform cross-compilation, please refer to the
[documentation](https://dune.readthedocs.io/en/stable/cross-compilation.html).

Main changes:

* Two opam files are generated when running `mirage configure`:
  - `<unikernel>-switch.opam`: for dependencies that are meant to be installed
    in the user's opam switch. It comprises of build tools such as
    `ocaml-freestanding` for Solo5 targets.
  - `<unikernel>-monorepo.opam`: for unikernel dependencies, they are locally
    fetched to compose a _dune workspace_.

* Unikernel dependencies are fetched in the source project using the
  `opam-monorepo` tool. This tool reads the `<unikernel>-monorepo.opam` file and
  make use of the opam solver to compute the transitive dependency set, saves
  that as a _lockfile_, and fetch sources in the `duniverse/` subfolder.
  More info on the [Github repository](https://github.com/ocamllabs/opam-monorepo).

* The compilation scheme use `dune`'s concept of a _workspace_: it's a set of
  libraries that are built together using the same _context_. For each
  compilation target, the Mirage tool is able to generate a _context_ definition
  able to compile for that target. A _context_ is defined by an OCaml compiler
  (or cross-compiler) toolchain, as defined by `findlib`, it can be tuned with
  environment variables and custom flags for the OCaml or underlying C compiler.

* The usual workflow: `mirage configure && make depends && mirage build` does
  not change however. Generated files are located into the new directory
  `./mirage` (OPAM files, `main.ml`, `key_gen.ml` or `manifest.json`) and
  the final artifact is created into the `./dist` directory.

Breaking changes:

* Unikernel dependencies need to use `dune` as a build system. Other build
  systems may be sandboxed in a Dune package, but the recommended way is to
  switch to `dune`. A set of packages have been ported to dune but the changes
  are not upstreamed, they are available in the `dune-universe`
[...]

PR #20581, opened 2022-01-29
functoria-runtime: add upper bounds to prepare for the MirageOS 4

samoht:
Extracted from #20558

### DIFF
--- a/packages/functoria-runtime/functoria-runtime.3.0.1/opam
+++ b/packages/functoria-runtime/functoria-runtime.3.0.1/opam
@@ -22,7 +22,7 @@ depends: [
   "dune" {>= "1.1"}
   "cmdliner" {>= "0.9.8"}
   "fmt"
-  "alcotest"  {with-test}
+  "alcotest"  {with-test & < "1.4.0"}
 ]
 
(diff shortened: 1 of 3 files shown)


samoht:
The revdeps tests are getting a bit crazy here (I am not sure to understand why the full universe is getting tested as revdeps of functoria-runtime.3.0.1...) so I'll merge this quickly.

hannesm:
this looks good to me.

kit-ty-kate:
> I am not sure to understand why the full universe is getting tested as revdeps of functoria-runtime.3.0.1...

functoria-runtime.3.0.1 --> mirage-runtime --> mirage-crypto-rng.0.7.0 --> {tls,x509,*} (explosion)

kit-ty-kate:
@hannesm is there a reason why mirage-crypto-rng.0.7.0 is the only version of mirage-crypto-rng to depend on mirage-runtime?

hannesm:
@kit-ty-kate yes, that one bundled the package that is nowadays separate as mirage-crypto-rng-mirage. since it is mostly of historic interest, we can also remove it (or mark it unavailable if that is sufficient).

PR #20582, opened 2022-01-29
mirage-crypto-pk: use mirage-runtime >= "4.0" (instead of "4.0.0") to support mirage 4.0.0~beta1 release

hannesm:
extracted from #20558 /cc @samoht @dinosaure


hannesm:
CI looks good, I don't think there's much value to have the revdeps finish.

PR #20584, opened 2022-01-29
mirage 3: restrict mirage-runtime and functoria to < 4.0

hannesm:
extracted from #20558 //cc @samoht @dinosaure


hannesm:
rebased on top of #20585 (to avoid CI failures requiring newer Fmt)

kit-ty-kate:
Thanks

PR #20588, opened 2022-01-29
[new release] mirage & all (4.0.0~beta1)

samoht:
The MirageOS library operating system

- Project page: <a href="https://github.com/mirage/mirage">https://github.com/mirage/mirage</a>
- Documentation: <a href="https://mirage.github.io/mirage/">https://mirage.github.io/mirage/</a>

##### CHANGES:

Refactor build process to use [Dune](https://dune.build/) build system. The
motivation is to drop `ocamlbuild`-induced technical debt and to obtain
first-class support for _cross-compilation_. To learn more about how Dune is
able to perform cross-compilation, please refer to the
[documentation](https://dune.readthedocs.io/en/stable/cross-compilation.html).

Main changes:

* Two opam files are generated when running `mirage configure`:
  - `<unikernel>-switch.opam`: for dependencies that are meant to be installed
    in the user's opam switch. It comprises of build tools such as
    `ocaml-freestanding` for Solo5 targets.
  - `<unikernel>-monorepo.opam`: for unikernel dependencies, they are locally
    fetched to compose a _dune workspace_.

* Unikernel dependencies are fetched in the source project using the
  `opam-monorepo` tool. This tool reads the `<unikernel>-monorepo.opam` file and
  make use of the opam solver to compute the transitive dependency set, saves
  that as a _lockfile_, and fetch sources in the `duniverse/` subfolder.
  More info on the
  [Github repository](https://github.com/ocamllabs/opam-monorepo).

* The compilation scheme use `dune`'s concept of a _workspace_: it's a set of
  libraries that are built together using the same _context_. For each
  compilation target, the Mirage tool is able to generate a _context_ definition
  able to compile for that target. A _context_ is defined by an OCaml compiler
  (or cross-compiler) toolchain, as defined by `findlib`, it can be tuned with
  environment variables and custom flags for the OCaml or underlying C compiler.

* The usual workflow `mirage configure && make depends && mirage build` does
  not change. However, files are now generated in the `./mirage` directory
  (OPAM files, `main.ml`, `key_gen.ml` or `manifest.json`), and
  the final artefact is created in the `./dist` directory.

Breaking changes:

* Unikernel dependencies need to use `dune` as a build system. Other build
  systems can be sandboxed, but the recommended way is to switch to `dune`.
  Many packages not compiling with dune yet have been ported  and are available
  as an additional
  [opam repository](https://github.com/dune-universe/opam-overlays) overlay.
  In addition, a few packages not supporting cross-compilation have been fixed
  and are available in another
  [opam repository](https://github.com/dune-universe/mirage-opam-overlays)
  overlay. The mirage tool uses these two opam overlays by default. To only
  use the default packages provided by Opam,
  use `mirage configure --no-extra-repo`.
* `Functoria_runtime.info` and `Mirage_runtime.info` now list all the libraries
  that are statically linked against the unikernel. The `packages` field have
  been removed and the `libraries` field is now accurate and contains the
  versions computed by `dune-build-info`.

* Update the DSL to describe devices into the `config.ml`.  We don't use
  objects anymore, and we replace it with the usage of `Mirage.impl` that
  expects the same _fields_ as before.


samoht:
The irmin-git error should be fixed by https://github.com/ocaml/opam-repository/pull/20589

samoht:
Docteur-unix should be fixed by #20591

kit-ty-kate:
Thanks a lot!

PR #20590, opened 2022-01-29
Release camlp5 for ocaml 4.14.0 compatibility

chetmurthy:
This is just to get camlp5 updated; there will be releases for the pa_ppx* packages presently to get those up-to-date also.


chetmurthy:
I think this CI run only failed for:
1. homebrew (still doesn't work automatically, user has to install two perl packages)
2. packages that I've marked as unavailable with this version of camlp5

    --> p5scm: sent PR to author
    --> pa_ppx: working on upgrade right now

kit-ty-kate:
Thanks

PR #20592, opened 2022-01-29
release pa_ppx 0.09 for compat with ocaml 4.14.0

chetmurthy:


kit-ty-kate:
Thanks. Could you return the fix in the opam file upstream so it can be taken into account for next time?

chetmurthy:
Roger that, have already put it into the opam file in the repo.

chetmurthy:
Thank you for merging so quickly!  On a weekend!

PR #20599, opened 2022-01-31
Release of `mirage.4.0.0~beta2`

dinosaure:
It's an hot-fix of the MirageOS distribution which allows us to emit right
constraints about dependencies on unikernels. The initial bug is about the
`4.0.0~beta1` release where OPAM considers `4.0.0~beta1 < 4.0.0`. To ensure
that unikernels downloads the right mirage{,-runtime} version, we must emit a
new constraint:

```
"mirage" {>= "4.0"}
```

from the mirage tool. This new distribution of `mirage.4.0.0~beta1` includes
this hot-fix.

PS: I know that it's an usual request so tell me if I'm doing right or not - and if an other better solution exists.


packages/patch.diff
```diff
@@ -0,0 +1,64 @@
+diff --git a/packages/functoria-runtime/functoria-runtime.4.0.0~beta1/opam b/packages/functoria-runtime/functoria-runtime.4.0.0~beta1/opam
```
kit-ty-kate, line 1: i believe this was added by mistake

dinosaure: Yes, I just force-pushed without it.

dinosaure:
@kit-ty-kate it's ok to merge this PR as is?

PR #20622, opened 2022-02-02
Patch world breaking gitlab-unix.0.1.0 installation

kit-ty-kate:
Installing `diff` will break anyone (including opam itself) trying to do anything after installing `gitlab-unix.0.1.0`

Reported by @dinosaure 
cc @tmcgilchrist


mseri:
Tests are good

PR #20621, opened 2022-02-02
Release mirage v4.0.0 beta3

dinosaure:
This is the final one which work perfectly!

PR #20623, opened 2022-02-02
Fix constraints about the new release of MirageOS 4.0.0

dinosaure:
Related to #20621 which fixes few constraints. Only `git`, `vchan-xen` and `mirage-entropy` probably needs a release.


dinosaure:
The solver timeout on some packages because they need (transitively) `opam.2.1`.

PR #20632, opened 2022-02-03
Fix constraints about mirage and docteur-{unix,solo5}

dinosaure:
Related to #20621

### DIFF
--- a/packages/docteur-unix/docteur-unix.0.0.2/opam
+++ b/packages/docteur-unix/docteur-unix.0.0.2/opam
@@ -11,7 +11,7 @@ depends: [
   "ocaml" {>= "4.07.0"}
   "dune" {>= "2.8.0"}
   "docteur" {= version}
-  "mirage-unix" {>= "4.0.0"}
+  "mirage-unix" {>= "4.0.0" & < "5.0.0"}
   "art" {>= "0.1.1"}
   "bigstringaf" {>= "0.7.0"}
   "carton" {>= "0.4.1"}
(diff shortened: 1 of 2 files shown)


packages/docteur-unix/docteur-unix.0.0.2/opam
```diff
@@ -11,7 +11,7 @@ depends: [
   "ocaml" {>= "4.07.0"}
   "dune" {>= "2.8.0"}
   "docteur" {= version}
-  "mirage-unix" {>= "4.0.0"}
+  "mirage-unix" {>= "4.0.0" & < "5.0.0"}
```
samoht, line 14: I still don't understand why we need an upper bound here. What error is this fixing?

samoht: ok got it, mirage3 needs mirage-unix.4 and mirag4 needs mirage-unix.5

hannesm:
my impression from https://github.com/ocaml/opam-repository/pull/20621 is that docteur and docteur-solo5 use "OS.Solo5", and their `mirage-solo5` dependency needs an upper bound (i.e. the docteur-solo5 fix here is fine, but docteur still needs an upper bound, and I can't see anything about docteur-unix).

dinosaure:
Again, all of these constraints is due to our rename of `OS` to `<target>_os`. I will cut a release of `docteur` after the release of `mirage`. It's fine to merge this PR and lint our world between MirageOS 3 and MirageOS 4.

kit-ty-kate:
Thanks!

PR #20542, opened 2022-01-25
add secp256k1-internal-0.3

hhugo:

PR #20637, opened 2022-02-03
secp256k1-internal.0.3 - Fix archive

hhugo:
Folllowing #20542
The archive was incorrect. I fixed it and bump the minor version.
Also "marked" secp256k1-internal.0.3.0 as broken


mseri:
Thanks

Issue #14646, opened 2019-08-06
Use system-num, if available


dra27:
cf. https://github.com/ocaml/opam-repository/pull/11300#issuecomment-518722338

Given that `num` is stable, it should be possible to do the same trick in `num` as `graphics` - so in a system switch, if the system-installed version of num is the correct version (e.g. from homebrew's `ocaml-num` package), make installing `num` a no-op, otherwise build it and override the `META` in `$SWITCHROOT/lib/num`.

Happy to be assigned this, assuming no one has a strong objection to the principle of the change - I just haven't to time right now to amend the package and propose a PR.

dbuenzli:
I would rather be against this. These tricks significantly hamper the ability for tools like `odig` or `omod` to correctly do their job as you'd need to make them aware of these install directories and how to look them up and bind a package name to them.

System compilers and their non opam managed `ocamlc -where` are already annoying enough, I constantly have to special case for them. In fact something I wanted to raise at some point was whether it would be feasible for system compilers to actually symlink the contents of `ocaml -where` in `$(opam var lib)/ocaml`. Note that it's the actual contents that need to be symlinked as opam assumes it can write that directory.

Well in fact I could be for this if the same symlinking strategy is workable and applied to such a package. One problem I see though is how to keep the symlinks in sync with the system packages if those update.

One will tell me this is because you are not using `META` files, yes that's true but I thought there was a consensus at some point that the eco-system should get rid of that indirection and have something reasonable and simple to be upstreamed to do library lookup (the more complicated version being Leo's namespaces). The tools I mention actually make this a reality in the current eco-system.

dra27:
@dbuenzli - even with the symlinking, don't `odig`/`omod` already/still have to cope with the files for `num` potentially being installed in two different places (either as a pure findlib package or in the OCaml stdlib dir)?

dbuenzli:
The problem is that there is no convention at the moment. Things are not entirely settled and ad-hoc both in `omod` and `odig`, as I was waiting on Leo's namespace proposal but maybe we should try to move on regardless.

Also [it seems](https://github.com/ocaml/ocaml/pull/1065#issuecomment-282796690) upstream is not entirely cold about it but I think we can also move towards the model described below without too much upstream involvement and/or implementing new lookups in the compiler.

So it was not clear to me that `system-num` would install in `ocamlc -where`. Regarding what happens in `odig`/`omod` in that case:

[...]

PR #20645, opened 2022-02-03
Add a post-message for num's known install failure

shonfeder:
This is meant to help users workaround the known installation failure with num reported in #14646 (and the various linked issues). 

The failure message was first suggested at https://github.com/ocaml/opam-repository/issues/14646#issuecomment-933022088, and pointers to the configuration change needed was provided by @dbuenzli in a reply.

This PR is to propose the workaround, but I honestly don't know which versions it should be applied to or what the optimal messaging is here. All corrections and suggestions for improvement welcome.

### DIFF
--- a/packages/num/num.1.0/opam
+++ b/packages/num/num.1.0/opam
@@ -39,3 +39,14 @@ url {
   src: "https://github.com/ocaml/num/archive/v1.0.tar.gz"
   checksum: "md5=a4c359108d0b885bf43909e705b21291"
 }
+post-messages: [
+  "Installation failure with the message 'The installation of num failed at \"make findlib-install\"'
+is due to a known packaging bug with the num package. As a workaround run the following two commands,
+then try to reinstall the target package again:
+
+  ocamlfind remove num
+  ocamlfind remove num-top
+
+See https://github.com/ocaml/opam-repository/issues/14646 for details."
+  {failure}
+]
--- a/packages/num/num.1.0/opam
+++ b/packages/num/num.1.0/opam
@@ -14,9 +14,13 @@ build: [
   [make]
 ]
 install: [
-  make
-  "install" {!ocaml:preinstalled}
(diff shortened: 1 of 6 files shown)


packages/num/num.1.0/opam
```diff
@@ -39,3 +39,14 @@ url {
   src: "https://github.com/ocaml/num/archive/v1.0.tar.gz"
   checksum: "md5=a4c359108d0b885bf43909e705b21291"
 }
+post-messages: [
+  "Installation failure with the message 'The installation of num failed at \"make findlib-install\"'
+is due to a known packaging bug with the num package. As a workaround run the following two commands,
+then try to reinstall the target package again:
+
+  ocamlfind remove num
+  ocamlfind remove num-top
```
kit-ty-kate, line 48: If this is a common issue why not run these before each install instead?

shonfeder: I'm not sure how frequent the problem is. It only hits me about once every couple of months. I'm happy with that tho, if it seems sensible! I'll wait until the additional reviewer you've tagged in weighs in :)

Thanks for the suggestion!

dra27:
The inclusion of a message looks fine - does this issue remain macOS-specific? Can you remind why it's `ocamlfind remove` rather than `brew remove`? I had thought the problem here was the system package manager's installed copy of num, which you wouldn't normally remove using `ocamlfind` itself?

shonfeder:
I hit this periodically on arch and Ubuntu, and afaik, have no installation of num other than via opam. I’m not clear on the underlying cause, unfortunately, and I am just cargo cutting the fix proposed in the linked issue :)

mseri:
I hit the same bug both without homebrew and on linux in the past a few times, I think adding the comment is harmless and help people incurring in this in the future

shonfeder:
Thanks for the better fix! :)

PR #20647, opened 2022-02-04
[new release] yaml and yaml-sexp (3.0.1)

avsm:
Parse and generate YAML 1.1/1.2 files

- Project page: <a href="https://github.com/avsm/ocaml-yaml">https://github.com/avsm/ocaml-yaml</a>
- Documentation: <a href="https://avsm.github.io/ocaml-yaml/">https://avsm.github.io/ocaml-yaml/</a>

##### CHANGES:

* Remove dependency on Rresult (@maxtori avsm/ocaml-yaml#50).

PR #20666, opened 2022-02-06
ppx_deriving_yaml 0.1.0: rresult was linked implicitly from yaml

mseri:


packages/ppx_deriving_yaml/ppx_deriving_yaml.0.1.0/opam
```diff
@@ -15,7 +15,7 @@ depends: [
   "ocaml" {>= "4.08.1"}
   "ppxlib" {>= "0.14.0"}
   "yaml"
```
kit-ty-kate, line 17:
```suggestion
  "yaml" {< "3.0.1"}
```
As i understand it, the ppx itself generates code that uses Rresult, so this should be on the dependency overall, not just for tests

mseri: Fair enough. In principle the ppx works with newer versions, you just need to manually link agains rresult

mseri:
Seen on https://github.com/ocaml/opam-repository/pull/20647

PR #20685, opened 2022-02-08
[new release] ounit2, ounit2-lwt, ounit and ounit-lwt (2.2.6)

gildor478:
OUnit testing framework

- Project page: <a href="https://github.com/gildor478/ounit">https://github.com/gildor478/ounit</a>
- Documentation: <a href="https://gildor478.github.io/ounit/">https://gildor478.github.io/ounit/</a>

### Fixed
- Use package seq to prepare for OCaml 5.00. Thanks to kit-ty-kate.

PR #20691, opened 2022-02-08
Add upper bounds to new ounit release

mseri:
Should fix 'failed: This expression has type char Stream.t -> unit'
Seen on #20685


kit-ty-kate:
Thanks!

PR #20694, opened 2022-02-09
extunix.0.1.4: only enable test if with-test

mseri:
Otherwise this enables tests even for more recent compilers,
making it fail due to the use of unsafe-string. Seen on #20685

### DIFF
--- a/packages/extunix/extunix.0.1.4/opam
+++ b/packages/extunix/extunix.0.1.4/opam
@@ -49,7 +49,6 @@ depends: [
   "ounit" {with-test & >= "1.0.3"}
   "base-bigarray"
   "base-unix"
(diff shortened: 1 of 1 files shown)


packages/extunix/extunix.0.1.4/opam
```diff
@@ -49,6 +49,7 @@ depends: [
   "ounit" {with-test & >= "1.0.3"}
   "base-bigarray"
   "base-unix"
+  "base-unsafe-string"
```
mseri, line 52:
```suggestion
```

mseri: I had forgotten to remove this, sorry

PR #20695, opened 2022-02-09
[new release] eio_main, eio_luv, eio_linux and eio (0.1)

talex5:
Effect-based direct-style IO mainloop for OCaml

- Project page: <a href="https://github.com/ocaml-multicore/eio">https://github.com/ocaml-multicore/eio</a>
- Documentation: <a href="https://ocaml-multicore.github.io/eio/">https://ocaml-multicore.github.io/eio/</a>

##### CHANGES:

- Initial release.

### DIFF
--- a/packages/eio/eio.0.1/opam
+++ b/packages/eio/eio.0.1/opam
@@ -1,0 +1,50 @@
+opam-version: "2.0"
+synopsis: "Effect-based direct-style IO API for OCaml"
+description: "An effect-based IO API for multicore OCaml with fibres."
+maintainer: ["anil@recoil.org"]
+authors: ["Anil Madhavapeddy" "Thomas Leonard"]
+license: "ISC"
+homepage: "https://github.com/ocaml-multicore/eio"
+doc: "https://ocaml-multicore.github.io/eio/"
+bug-reports: "https://github.com/ocaml-multicore/eio/issues"
+depends: [
+  "dune" {>= "2.9"}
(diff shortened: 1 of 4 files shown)


talex5:
Note the `Unix.Unix_error(Unix.ENOMEM, "io_uring_register_buffers", "")` error means that the CI worker decided not to let it lock enough memory. This often happens when the cluster is busy.

I see one of the tests wants to create a socket in /tmp, which fails because that's not mounted in CI. I'll change that upstream - does it need a new release?

The nnp test says:
```
  - downgrade ocaml-variants    4.13.1+options to 4.12.0+domains+effects
```
Is that expected?

kit-ty-kate:
> Is that expected?

it’s not. I saw that some time ago (it’s only switching when it is not available, so it’s not a huge issue) but I was kind of hoping for https://github.com/ocaml/opam/pull/5002 to be merged and backported a few weeks ago but obviously it wasn’t. I’ll try to hotfix opam-repo-ci but for now we can ignore them.

kit-ty-kate:
> I see one of the tests wants to create a socket in /tmp, which fails because that's not mounted in CI. I'll change that upstream - does it need a new release?

if you can, that’d be much appreciated. `/tmp` is not available in the sandbox, you need to use `$TMPDIR` whenever available

talex5:
I think I might just disable the eio_main tests until we have the ENOMEM thing fixed. Otherwise, it'll generate lots of random failures in opam-health-check.

PR #20703, opened 2022-02-09
[new release] stdcompat.18

thierry-martinez:
This PR adds the new release of `stdcompat.18`.

* Support for OCaml 4.14 with
  - Lot of new functions in `Seq`
  - `Uchar.utf_decode` and co
  - `In_channel` and `Out_channel` modules
  - `Sys.{development_version, ocaml_release}`

* Add `Stdlib.{acosh,asinh,atanh}` missing from 4.13 and `Stdlib.__FUNCTION__` missing from 4.12

* Add module `Unit` missing from 4.08

* Add module `Random`, with new functions introduced from 4.13

* `Filename.chop_suffix` checks that suffixes match and fails otherwise
  (behavior introduced in 4.14)

* `Buffer.add_channel` adds data read from the channel even if `End_of_file` has been reached
  (behavior introduced in 4.03)

* Add dependency for generating `.cmt` files
  (reported by Sabyrzhan Tasbolatov)

### DIFF
--- a/packages/stdcompat/stdcompat.18/opam
+++ b/packages/stdcompat/stdcompat.18/opam
@@ -1,0 +1,26 @@
+opam-version: "2.0"
+synopsis: "Compatibility module for OCaml standard library"
+description:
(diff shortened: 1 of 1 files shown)


mseri:
Looks good, thanks

PR #20533, opened 2022-01-24
[new release] js_of_ocaml, js_of_ocaml-tyxml, js_of_ocaml-toplevel, js_of_ocaml-ppx_deriving_json, js_of_ocaml-ppx, js_of_ocaml-lwt and js_of_ocaml-compiler (4.0.0)

hhugo:
Compiler from OCaml bytecode to JavaScript

- Project page: <a href="https://ocsigen.org/js_of_ocaml/latest/manual/overview">https://ocsigen.org/js_of_ocaml/latest/manual/overview</a>
- Documentation: <a href="https://ocsigen.org/js_of_ocaml/latest/manual/overview">https://ocsigen.org/js_of_ocaml/latest/manual/overview</a>

## Features/Changes
* Compiler: add --target-env flag, for JS runtime specific compilation targets (ocsigen/js_of_ocaml#1160).
* Compiler: static evaluation of backend_type (ocsigen/js_of_ocaml#1166)
* Compiler: speedup emitting js files (ocsigen/js_of_ocaml#1174)
* Compiler: simplify (a | 0) >>> 0 into (a >>> 0) (ocsigen/js_of_ocaml#1177)
* Compiler: improve static evaluation of cond (ocsigen/js_of_ocaml#1178)
* Compiler: be more consistent dealing with js vs ocaml strings (ocsigen/js_of_ocaml#984)
* Compiler: Compiler: add BigInt to provided symbols (fix ocsigen/js_of_ocaml#1168) (ocsigen/js_of_ocaml#1191)
* Compiler: use globalThis, drop joo_global_object ocsigen/js_of_ocaml#1193
* Compiler: new -Werror flag to turn wanrings into errors (ocsigen/js_of_ocaml#1222)
* Compiler: make the inlining less agressive, reduce size, improve pref (ocsigen/js_of_ocaml#1220)
* Compiler: rename internal library js_of_ocaml-compiler.runtime to js_of_ocaml-compiler.runtime-files
* Lib: new runtime library to improve compatibility with Brr and gen_js_api
* Lib: add messageEvent to Dom_html (ocsigen/js_of_ocaml#1164)
* Lib: add PerformanceObserver API (ocsigen/js_of_ocaml#1164)
* Lib: add CSSStyleDeclaration.{setProperty, getPropertyValue, getPropertyPriority, removeProperty} (ocsigen/js_of_ocaml#1170)
* Lib: make window.{inner,outer}{Width,Height} non-optional
[...]

PR #20716, opened 2022-02-10
2 packages from LexiFi/gen_js_api at 1.1.0

mlasson:
This pull-request concerns:
-`gen_js_api.1.1.0`: Easy OCaml bindings for JavaScript libraries
-`ojs.1.1.0`: Runtime Library for gen_js_api generated libraries

---
* Homepage: https://github.com/LexiFi/gen_js_api
* Source repo: git+https://github.com/LexiFi/gen_js_api.git
* Bug tracker: https://github.com/LexiFi/gen_js_api/issues

---


smorimoto:
Now it's failing on the domain, but I think this is a jsoo issue. Should we just add constraints or wait for a new jsoo release to be published?

mlasson:
Thanks for the feedback !

It seems that this PR https://github.com/ocaml/opam-repository/pull/20039 added a conflict with "base-domain" directly in js_of_ocaml-compiler. Wasn't this conflict forgotten in https://github.com/ocaml/opam-repository/pull/20533 ? @kit-ty-kate @hhugo ?

I don't mind adding the conflict directly in gen_js_api but it seems that other libraries may have the same problem.

mseri:
You are right. I thought the bound on ocaml < 4.14 would be enough, but that does not account for the specific switch. I’ll fix it separately, thanks

PR #20718, opened 2022-02-10
[new release] ocaml-lsp-server, lsp and jsonrpc (1.10.0)

rgrinberg:
LSP Server for OCaml

## Features

- Add better support for code folding: more folds and more precise folds

## Fixes

- Fix infer interface code action crash when implementation source does not
  exist (ocaml/ocaml-lsp#597)

- Improve error message when the reason plugin for merlin is absent (ocaml/ocaml-lsp#608)

- Fix `chdir` races when running ppx (ocaml/ocaml-lsp#550)

- More accurate completion kinds.
  New completion kinds for variants and fields. Removed inaccurate completion
  kinds for constructors and types. (ocaml/ocaml-lsp#510)

- Fix handling request cancellation (ocaml/ocaml-lsp#616)

### DIFF
--- a/packages/jsonrpc/jsonrpc.1.10.0/opam
+++ b/packages/jsonrpc/jsonrpc.1.10.0/opam
@@ -1,0 +1,48 @@
+opam-version: "2.0"
+synopsis: "Jsonrpc protocol implemenation"
+description: "See https://www.jsonrpc.org/specification"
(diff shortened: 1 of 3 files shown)


mseri:
I don’t think we can make such a fine availability check, let’s merge as is. I think we can figure it out if somebody has issues with it

PR #20724, opened 2022-02-11
updates conf-bap-llvm to reflect versions supported by modern bap

ivg:
It looks like we didn't update this package for a long time. BAP supports llvm versions from 6 to 12 (and up to 14 in the 2.5.0-alpha).

### DIFF
--- a/packages/conf-bap-llvm/conf-bap-llvm.1.7/files/find-llvm.ml.in
+++ b/packages/conf-bap-llvm/conf-bap-llvm.1.7/files/find-llvm.ml.in
@@ -1,0 +1,183 @@
+open Printf
+open StdLabels
+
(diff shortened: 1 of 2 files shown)


mseri:
This one goes up to 12, but from the description it seems expected. Can you confirm?

ivg:
Yes, this version is capped with 12.0 because it is the upper bound for the released 2.4.0. 

The current alpha version (which is not available in the upstream, read this, opam repository) supports 13 and  14.  The alpha version could be installed using our testing opam-repository (which is the rolling release bap repository), e.g.,
```
opam repo add  bap-testing git+https://github.com/BinaryAnalysisPlatform/opam-repository#testing
```

PR #20745, opened 2022-02-16
Add gcc to depexts at ocaml-mariadb for nixos distribution

Nymphium:


mseri:
You should probably add it to conf-gcc if it is needed by some package

Nymphium:
Sorry, I misunderstood.`mariadb` needs `gcc` and this package doesn't.

> You should probably add it to conf-gcc if it is needed by some package
Okay, I close this and create another PR. Thank you.

PR #20753, opened 2022-02-17
Add conf-gcc depext for mariadb

Nymphium:


mseri:
Thanks. @kit-ty-kate to fix the errors we need to modify the upper bound on `4.08` to `4.07` in mariadb <= 1.1.2 ?

kit-ty-kate:
@mseri sure, sounds good.

@Nymphium does it need gcc specifically or any `cc` command will do?

kit-ty-kate:
Thanks a lot!

PR #20754, opened 2022-02-17
[new release] domainslib (0.4.1)

Sudha247:
This release fixes compatibility with OCaml `5.00.0+trunk` in https://github.com/ocaml-multicore/domainslib/pull/61. Breaks compatibility with older Multicore variants `4.12.0+domains` and `4.12.0+domains+effects`.

### DIFF
--- a/packages/domainslib/domainslib.0.4.1/opam
+++ b/packages/domainslib/domainslib.0.4.1/opam
@@ -9,12 +9,12 @@ dev-repo: "git+https://github.com/ocaml-multicore/domainslib.git"
 bug-reports: "https://github.com/ocaml-multicore/domainslib/issues"
 tags: ["org:ocamllabs"]
 depends: [
-  "ocaml" {>= 5.00}
+  "ocaml" {>= "5.00"}
   "dune" {>= "1.8"}
-  "base-domains"
   "mirage-clock-unix" {with-test}
 ]
 build: ["dune" "build" "-p" name "-j" jobs]
+run-test: ["dune" "runtest" "-p" name "-j" jobs]
 url {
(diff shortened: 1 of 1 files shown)


packages/domainslib/domainslib.0.4.1/opam
```diff
@@ -0,0 +1,20 @@
...
+authors: ["KC Sivaramakrishnan <kc@kcsrk.info>"]
+homepage: "https://github.com/ocaml-multicore/domainslib"
+doc: "https://ocaml-multicore.github.io/domainslib/"
+synopsis: "Parallel Structures over Domains for Multicore OCaml"
+license: "ISC"
+dev-repo: "git+https://github.com/ocaml-multicore/domainslib.git"
+bug-reports: "https://github.com/ocaml-multicore/domainslib/issues"
+tags: ["org:ocamllabs"]
+depends: [
+  "dune" {>= "1.8"}
```
mseri, line 12:
```suggestion
  "ocaml" {>= "5"}
  "dune" {>= "1.8"}
```
We need to make it explicit that it is incompatible with 4.12+domains

mseri: Don't know if "5" is the right number there though

Sudha247: Thanks! I've followed the `5.00` convention used here https://github.com/ocaml/opam-repository/blob/master/packages/base-domains/base-domains.base/opam#L7. Might probably need an update later if the version numbering were to change.

kit-ty-kate:
Could you return the fixes in the opam file upstream so that they can be taken into account for next time?

PR #20755, opened 2022-02-17
Add License to conf-libev and conf-pam packages

psafont:
Conf-libev actually lets the user choose between the BSD and the GPL license, but the base is the BSD license.
Conf-pam has a similar license but doesn't specify which GPL version to choose.

I didn't see a way to express this kind of dual license using SPDX, so I opted to pick the base license.


kit-ty-kate:
> I didn't see a way to express this kind of dual license using SPDX, so I opted to pick the base license.

The SPDX format has `LICENSE1 OR LICENSE2` and `LICENSE1 AND LICENSE2`. See the spec here: https://spdx.github.io/spdx-spec/SPDX-license-expressions/

In any case I think this is fine as-is for opam-repository. Thanks!

PR #20762, opened 2022-02-18
Why3 1.4.1

silene:


silene:
Of the 39 failed jobs, 37 are caused by a failure to install Gtk2 and 1 is caused by a network error. The remaining failure is with the compiler `4.12.0+domains+effects`, as it defines extra OCaml keywords which conflict with Why3 source code.

packages/why3/why3.1.4.1/opam
```diff
@@ -0,0 +1,78 @@
...
+  "ppx_deriving" {build}
+  "ppx_sexp_conv" {build}
+  "mlmpfr"
+]
+
+conflicts: [
+  "why3-base"
+  "ocamlgraph" {< "1.8.2"}
+  "mlmpfr" {< "4.0.0"}
+]
```
mseri, line 66:
```suggestion
  "base-effects"
]
```

silene:
Thanks for suggesting `base-effects`. I was looking for something akin to `ocaml-option-effects` and did not find anything relevant.

(The network error "Could not resolve host: github.com" is still there, so it seems like it is not transient but specific to the 4.14 job.)

mseri:
Yes, don't worry about that. The rest looks fine

PR #20763, opened 2022-02-18
Replace 'typeof OCaml system' with the equivalent valid SPDX

psafont:
i.e. LGPL-2.1-only WITH OCaml-LGPL-linking-exception

https://github.com/ocaml/uchar/pull/4
https://github.com/ocaml/stdlib-shims/pull/21


kit-ty-kate:
Thanks a lot

PR #20781, opened 2022-02-22
Lambdapi: update dependencies

fblanqui:


packages/lambdapi/lambdapi.2.1.0/opam
```diff
@@ -48,13 +48,12 @@ depends: [
...
   "alcotest" {with-test}
   "alt-ergo" {with-test}
-  "alt-ergo" {<= "2.4.0"}
   "bindlib" {>= "5.0.1"}
   "timed" {>= "1.0"}
   "pratter" {>= "1.2"}
   "why3" {>= "1.4.0"}
   "yojson" {>= "1.6.0"}
-  "cmdliner" {>= "1.0.3"}
+  "cmdliner" {>= "1.0.3" & <= "1.0.4"}
```
kit-ty-kate, line 56: Are you sure? It compiled just fine with cmdliner.1.1.0
see for example: http://check.ocamllabs.io/log/1645246063-968ab5c21b6019c3e65d6b6cbee96fcc8ca43e39/4.14/good/lambdapi.2.1.0

packages/lambdapi/lambdapi.2.1.0/opam
```diff
@@ -48,13 +48,12 @@ depends: [
   "sedlex" {>= "2.2"}
   "alcotest" {with-test}
   "alt-ergo" {with-test}
-  "alt-ergo" {<= "2.4.0"}
```
kit-ty-kate, line 51: what’s the reason for removing this dependency?

fblanqui: The new release of Why3 is now compatible with Alt-ergo 2.4.1.

packages/lambdapi/lambdapi.2.1.0/opam
fblanqui: Ah, strange, I recently got CI problems with this new version of cmdliner. I am going to check again.

fblanqui: It does not work:
```
09:29 ~/src/lambdapi (master) make
File "src/cli/init.ml", line 73, characters 2-11:
73 |   Term.info "init" ~doc
       ^^^^^^^^^
Error (alert deprecated): Cmdliner.Term.info
Use Cmd.info instead.
File "src/cli/install.ml", line 116, characters 2-11:
116 |   Term.info "install" ~doc
        ^^^^^^^^^
Error (alert deprecated): Cmdliner.Term.info
Use Cmd.info instead.
File "src/cli/install.ml", line 121, characters 2-11:
121 |   Term.info "uninstall" ~doc
        ^^^^^^^^^
Error (alert deprecated): Cmdliner.Term.info
Use Cmd.info instead.
make: *** [Makefile:11 : bin] Erreur 1
```

kit-ty-kate: these are deprecations. Not real errors. This means you use dune directly, not the build instruction from your opam file

packages/lambdapi/lambdapi.2.1.0/opam
kit-ty-kate: Then the fix is to remove the constraint, not the dependency entirely

fblanqui: Lambdapi does not actually depends on Alt-ergo. Alt-ergo is only used for a few tests (hence the dependency on line 50) and tests are not included in the build of Lambdapi.

packages/lambdapi/lambdapi.2.1.0/opam
fblanqui: I see but it may well be the case that in a few months or years, this version of lambdapi won't compile anymore with cmdliner. So I prefer to anticipate. A PR is already ready for moving to cmdliner 1.1.0: https://github.com/Deducteam/lambdapi/pull/851. So the next release of Lambdapi in a few weeks or months will be for cmdliner >= 1.1.0.

fblanqui:
Hello @kit-ty-kate . Is it ok or do you want me to change something and what?

packages/lambdapi/lambdapi.2.1.0/opam
```diff
@@ -48,13 +48,12 @@ depends: [
...
   "bindlib" {>= "5.0.1"}
```
rlepigre, line 51: @fblanqui You'll also need to change that to an equality. The release of Bindlib 6.0.0 (see https://github.com/ocaml/opam-repository/pull/20808) is gonna break Lambdapi otherwise. In general, it is not guaranteed that libraries are backward compatible, so it is a good idea to keep dependencies strict.

rlepigre: I actually did the constraining myself in https://github.com/ocaml/opam-repository/pull/20808. (Also see https://github.com/Deducteam/lambdapi/issues/852.)

fblanqui: Thanks @rlepigre . This change is fine for me.

mseri:
The changes have been justified enough imo and the new version of lambdapi is on the way from what I read here

fblanqui:
Thank you!

PR #20794, opened 2022-02-24
Add batteries 3.5.1

kit-ty-kate:
cc @UnixJunkie 

Just opening that as a draft. I believe everything is ready for another release once CI is green


UnixJunkie:
Are we supposed to be compatible against 4.14.0~beta1?
That's the one I am testing against right now.

kit-ty-kate:
> Are we supposed to be compatible against 4.14.0~beta1? That's the one I am testing against right now.

yes that’s the goal

UnixJunkie:
I have tagged batteries v3.5.1.

PR #20797, opened 2022-02-24
Add google-drive-ocamlfuse 0.7.28

astrada:


kit-ty-kate:
Thanks! Could you return the fix to the opam file upstream so that it can be taken into account for next time?

astrada:
Sure, thanks!

PR #20805, opened 2022-02-25
Rename OCaml compiler from 5.00.0 to 5.0.0

patricoferris:
After https://github.com/ocaml/ocaml/pull/11049 trying to install a trunk compiler would lead to:

```
$ opam switch create 5.00.0+trunk
OCaml version mismatch: 5.0.0, expected 5.00.0
```

This seems to fix it locally for me, I don't know if all of the various version constraints (like `< "5.00"`) also need changed across the repository?


kit-ty-kate:
Thanks!

kit-ty-kate:
> This seems to fix it locally for me, I don't know if all of the various version constraints (like `< "5.00"`) also need changed across the repository?

They don’t as `5.00 = 5.0` per opam version ordering: https://opam.ocaml.org/doc/Manual.html#Package-Formulas

PR #20818, opened 2022-02-28
[new release] links, links-postgresql, links-sqlite3 and links-mysql (0.9.6)

dhil:

##### CHANGES:

This release extends the core features of Links and resolves various
minor bugs.

* Links now supports System F-style explicit type abstractions:
  For instance, writing `/\ [a, e::Row] { foo }` abstracts the expression `foo`
  over type variable `a` and row variable `e`. Here, `foo` must have a unique
  type and must be pure (to satisfy the value restriction).
* Fixed a bug in "mixing" query normalisation, which prevented certain queries using
  concatenation inside `for` statements from being correctly converted to SQL.
* Links now has basic support for temporal database operations. More information
  can be found on the [Wiki](https://github.com/links-lang/links/wiki/Temporal-Databases).
  There are new keywords: `valid`, `to`, `vt_insert`, `tt_insert`, and `TemporalTable`.
* A new commandline option `--compile` (shorthand `-c`) has been
  added, which runs Links in a "compile only" mode. In this mode the
  JavaScript compilation artefact can be saved to a file (the naming
  of this file is controlled via the commandline option `-o`). Note
  that the generated file may not be directly runnable without linking
  the runtime system first. Currently, the runtime system must be
  linked manually.
* Fixed a bug where calling either of `newAP`, `newClientAP`, and
  `newServerAP` on the client-side would crash the client.
* It is now possible to dispatch an MVU message from outside of the
  event loop. This is particularly useful, for example, when dealing
  with a persistent, stateful thread which is receiving messages from
  a server. New things include:
  + A new type alias `MvuHandle(msg)`.
  + A family of runners: `runHandle`, `runCmdHandle`,
    `runSimpleHandle` which return an `MvuHandle(msg)` rather than the
    unit value.
  + A new dispatcher `Mvu.dispatch : (msg, MvuHandle(msg)) ~> ()`,
    which directly dispatches a message to the MVU loop.

* The built-in webserver now supports SSL connections.  To enable
  secure connections, you must first obtain an adequate certificate
  and key, e.g. via Let's Encrypt or a self-signed certificate. The
  latter can be useful for testing, e.g. the following command starts
  an interactive process to create a self-signed certificate (that
  uses 4096 bits RSA encryption and is valid for 365 days):
```shell
openssl req -x509 -newkey rsa:4096 -keyout server.key -out server.crt -days 365 -nodes
```
  After obtaining a valid certificate, you must tell Links to run in SSL
  mode and you must also tell it how to locate the `key` and `crt`
  file. This can be done via a configuration file, e.g.

```
# ssl.config
ssl=true
ssl_cert_file=server.crt
ssl_key_file=server.key
```
  Then running `./links --config=ssl.config <file.links>` will cause
  the webserver to only serve requests via https.

  When a webpage is served via https, then the websocket layer will
  automatically communicate via the wss protocol.
* Other various bug fixes.

### DIFF
--- a/packages/links-mysql/links-mysql.0.9.6/opam
+++ b/packages/links-mysql/links-mysql.0.9.6/opam
@@ -1,0 +1,31 @@
+opam-version: "2.0"
+maintainer: "James Cheney <jcheney@inf.ed.ac.uk>"
+authors: "The Links Team <links-dev@inf.ed.ac.uk>"
+synopsis: "MySQL database driver for the Links Programming Language"
+description: "MySQL database driver for the Links Programming Language"
+homepage: "https://github.com/links-lang/links"
+dev-repo: "git+https://github.com/links-lang/links.git"
+bug-reports: "https://github.com/links-lang/links/issues"
+license: "GPL-3.0-only"
+
+build: [
+  [ "dune" "subst" ] {dev}
+  [ "dune" "build" "-p" name "-j" jobs ]
+]
+
+depends: [
+  "ocaml" {>= "4.08.0"}
+  "dune" {>= "2.7"}
+  "conf-mysql"
+  "mysql"
+  "links" {= version}
+]
+url {
+  src:
+    "https://github.com/links-lang/links/releases/download/0.9.6/links-0.9.6.tbz"
+  checksum: [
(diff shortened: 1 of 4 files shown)


kit-ty-kate:
Thanks!

PR #20826, opened 2022-03-01
[new-release] domainslib (0.4.2)

Sudha247:
Includes `Effect.eff` -> `Effect.t` change from OCaml trunk. (https://github.com/ocaml-multicore/domainslib/pull/65)


kit-ty-kate:
Thanks

PR #20842, opened 2022-03-03
Multicont version 1.0.0-rc.1

dhil:
(Disclaimer: this is my first attempt at releasing a package that depends on an unstable switch version, so chances are this package might require some tweaking before it can go into this repository.)

This is release candidate 1 (and initial release) of the multicont library for OCaml.

This release is compatible with the [OCaml 5.0 trunk @ 7f7c0f5](https://github.com/ocaml/ocaml/commit/7f7c0f521b65874f5d102b5a4da14ae116203def) and x64 architectures.

**Full Changelog**: https://github.com/dhil/ocaml-multicont/commits/v1.0.0-rc.1


packages/multicont/multicont.1.0.0-rc.1/opam
```diff
@@ -0,0 +1,36 @@
...
+description: "This library provides facilities for programming with multi-shot continuations in OCaml"
+homepage: "https://github.com/dhil/ocaml-multicont"
+dev-repo: "git+https://github.com/dhil/ocaml-multicont.git"
+bug-reports: "https://github.com/dhil/ocaml-multicont"
+license: "MIT"
+version: "1.0.0-rc.1"
+
+
+build: [
+  [ make "all" ] {dev}
```
mseri, line 14:
```suggestion
```
It is a duplicate of the next line

packages/multicont/multicont.1.0.0-rc.1/opam
```diff
@@ -0,0 +1,36 @@
...
+  [ make "all" ]
+]
+
+install: [
+  [ make "install" ]
+  [ make "VERSION=%{version}%" "dune-package" ]
```
mseri, line 20: Why you want the `dune-project` file if it is compiled and installed via a Makefile?
Also, you could generate a `multicont.install` file and `opam` would automatically use that to make the installation. In this case, you could statically drop the file in the repository

mseri: I see. But why don’t just use dune to build all the project then?

kit-ty-kate:
The version of this package should be changed to `1.0.0~rc1` (~ instead of -) as `1.0.0~rc1 < 1.0.0 < 1.0.0-rc1` per https://opam.ocaml.org/doc/Manual.html#version-ordering

packages/multicont/multicont.1.0.0-rc.1/opam
```diff
@@ -7,7 +7,7 @@ homepage: "https://github.com/dhil/ocaml-multicont"
 dev-repo: "git+https://github.com/dhil/ocaml-multicont.git"
 bug-reports: "https://github.com/dhil/ocaml-multicont"
 license: "MIT"
-version: "1.0.0-rc.1"
+version: "1.0.0~rc.1"
```
kit-ty-kate, line 10:
```suggestion
```
The version field is not used in opam-repository. The version number is already present in the directory structure. You need to `git mv` the directory instead and remove this field

packages/multicont/multicont.1.0.0-rc.1/opam
dhil: I was fighting a bit with dune initially, as I need to compile the library differently for byte and native modes. The culprit is how to compile the `multicont_stubs.c` and `fiber_primitives.c` with the correct flags.

packages/multicont/multicont.1.0.0-rc.1/opam
dhil: I use the version available during the install atm. Although, I could potentially make a script to generate the .install as suggested above.

packages/multicont/multicont.1.0.0-rc.1/opam
kit-ty-kate: I’m not sure why you need to create a `dune-package` file though. Dune already understands META file so the dune-package file isn’t necessary and puts more work for you.

Also installing a `dune-package` file would require to add a conflict with `dune < 3.0`. But yeah i don’t think it’s worth having it anyway.

kit-ty-kate: > I was fighting a bit with dune initially, as I need to compile the library differently for byte and native modes. The culprit is how to compile the `multicont_stubs.c` and `fiber_primitives.c` with the correct flags.

On the other topic: if you just need to pass different arguments to ocamlc and ocamlopt (If I’m reading your makefile correctly) using `ocamlc_flags`/`ocamlopt_flags` should be enough. See for example:

https://github.com/ocaml/dune/blob/d75bd885262b98075299e9145a2aa3bd1a54a35d/test/blackbox-tests/test-cases/env/envs-and-contexts.t/run.t

dhil: > I’m not sure why you need to create a `dune-package` file though. Dune already understands META file so the dune-package file isn’t necessary and puts more work for you.
> 
> Also installing a `dune-package` file would require to add a conflict with `dune < 3.0`. But yeah i don’t think it’s worth having it anyway.

I built the dune-package file because dune wouldn't load/link the library correctly otherwise (when I compiled the examples). It should be noted that this library tinkers with the runtime primitives of OCaml. It is also very possible that I failed to get it to work properly before due to my lack of knowledge of dune. I can give it another try at building the project with dune.

packages/multicont/multicont.1.0.0-rc.1/opam
kit-ty-kate: The opam file is not copied in the build directory IIRC so removing it wouldn’t do anything to your package

packages/multicont/multicont.1.0.0-rc.1/opam
dhil: I have rediscovered the problem that led me away from dune. The problem is that I need to generate different foreign archives for byte and native code. It is not clear to me how I can instruct dune to build, e.g. "libmulticont.a" to link with code produced by ocamlc, and "libmulticontopt.a" to link with code produced by ocamlopt.

mseri:
Sounds good!

dhil:
I am happy to give it one more try to move to `dune`. I will try using the hint you gave me in your previous comment.

packages/multicont/multicont.1.0.0~rc.1/opam
```diff
@@ -0,0 +1,32 @@
...
+
+install: [
+  [ make "install" ]
+  [ make "VERSION=%{version}%" "META" ]
+]
+
+depends: [
+  "ocaml" {>= "5.0"}
+]
+
```
mseri, line 23:
```suggestion
available: arch = "x86_64" & arch = "arm64"
```

packages/multicont/multicont.1.0.0~rc.1/opam
```diff
@@ -0,0 +1,32 @@
...
+install: [
+  [ make "install" ]
+  [ make "VERSION=%{version}%" "META" ]
+]
+
+depends: [
+  "ocaml" {>= "5.0"}
+]
+available: arch = "x86_64" & arch = "arm64"
```
mseri, line 23:
```suggestion

available: arch = "x86_64" | arch = "arm64"

```

Apologies, wrote to fast

mseri:
Thanks

PR #20853, opened 2022-03-07
[new release] eio_main, eio_luv, eio_linux and eio (0.2)

talex5:
Effect-based direct-style IO mainloop for OCaml

- Project page: <a href="https://github.com/ocaml-multicore/eio">https://github.com/ocaml-multicore/eio</a>
- Documentation: <a href="https://ocaml-multicore.github.io/eio/">https://ocaml-multicore.github.io/eio/</a>

##### CHANGES:

- Add support for UDP (@patricoferris ocaml-multicore/eio#171).

- Rename Fibre to Fiber (@talex5 ocaml-multicore/eio#195). This is to match the compiler's spelling.

- Switch to luv backend if uring can't be used (@talex5 ocaml-multicore/eio#203).
  Useful on Windows with WSL, and also in Docker containers on older systems.

- Eio_linux: cope with lack of fixed chunks (@talex5 ocaml-multicore/eio#200).
  - If we run out of fixed memory, just use regular memory instead of waiting (which might deadlock).
  - If we try to allocate a fixed buffer and fail, we now just log a warning and continue without one.

- Add support for FD passing with Eio_linux (@talex5 ocaml-multicore/eio#199).

- Add `Eio_unix.FD.as_socket` (@talex5 ocaml-multicore/eio#193).
  Useful for working with existing libraries that provide a `Unix.file_descr`, or for receiving FDs from elsewhere (e.g. socket activation).
  Also, the `Luv.{File,Handle}.of_luv` functions now allow controlling whether to close the wrapped FD.

- Add `Eio_unix.sleep` (@talex5 ocaml-multicore/eio#188). Based on feedback that some people don't want to treat time as a capability. Possibly also useful for debugging race conditions.

- Tidy up forking API (@talex5 ocaml-multicore/eio#192). Moves some common code out the the individual backends.

- Improve documentation (@talex5 ocaml-multicore/eio#197 ocaml-multicore/eio#194 ocaml-multicore/eio#186 ocaml-multicore/eio#185). In particular, explain more low-level details about how cancellation works.

- Add an example `Eio_null` backend (@talex5 ocaml-multicore/eio#189). This supports creating fibers, promises and cancellation, but provides no IO operations.

- `Effect.eff` is now `Effect.t` in OCaml trunk (@talex5 ocaml-multicore/eio#201).

### DIFF
--- a/packages/eio/eio.0.2/opam
+++ b/packages/eio/eio.0.2/opam
@@ -1,0 +1,50 @@
+opam-version: "2.0"
+synopsis: "Effect-based direct-style IO API for OCaml"
+description: "An effect-based IO API for multicore OCaml with fibers."
+maintainer: ["anil@recoil.org"]
+authors: ["Anil Madhavapeddy" "Thomas Leonard"]
+license: "ISC"
+homepage: "https://github.com/ocaml-multicore/eio"
+doc: "https://ocaml-multicore.github.io/eio/"
+bug-reports: "https://github.com/ocaml-multicore/eio/issues"
+depends: [
+  "dune" {>= "2.9"}
+  "ocaml" {>= "4.12.0"}
+  "base-domains"
+  "cstruct" {>= "6.0.1"}
+  "lwt-dllist"
+  "optint" {>= "0.1.0"}
+  "psq" {>= "0.2.0"}
+  "fmt" {>= "0.8.9"}
+  "astring" {>= "0.8.5" & with-test}
+  "crowbar" {>= "0.2" & with-test}
+  "mtime" {>= "1.2.0"}
+  "alcotest" {>= "1.4.0" & with-test}
+  "odoc" {with-doc}
+]
+build: [
+  ["dune" "subst"] {dev}
+  [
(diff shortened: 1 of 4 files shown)


talex5:
Builds on `cree` seem to be having some strange problem:
```
/home/opam: (copy (src .) (dst opam-repository/))
tar: Skipping to next header
tar: Skipping to next header
```
Since all packages depend on base-domains, the 4.12+domains tests are the only ones that matter (and they all passed). The other failures seem to be compilers getting "upgraded" to different versions, e.g.

debian-11-ocaml-4.13-no-flat-float-array/amd64 says:
```
  - upgrade   ocaml-variants  4.13.1+options to 5.0.0+trunk
```
(and then 5.0.0+trunk fails due to https://github.com/ocaml-doc/odoc-parser/pull/6)

talex5:
The 4.12+domains tests are no longer present in the results for some reason.

kit-ty-kate:
Ah sigh it’s still needed. Ok I’ll put it back.

PR #20859, opened 2022-03-07
Package yuujinchou.2.0.0

favonia:
### `yuujinchou.2.0.0`
Name pattern combinators
This package implements a pattern language for selecting names. It intends to facilitate the implementation of the "import" statement or any feature that allows users to select a group of names by patterns.

---
* Homepage: https://github.com/RedPRL/yuujinchou
* Source repo: git+https://github.com/RedPRL/yuujinchou.git
* Bug tracker: https://github.com/RedPRL/yuujinchou/issues

---

### DIFF
--- a/packages/yuujinchou/yuujinchou.2.0.0/opam
+++ b/packages/yuujinchou/yuujinchou.2.0.0/opam
@@ -1,0 +1,32 @@
+opam-version: "2.0"
+synopsis: "Name pattern combinators"
+description: """
+This package implements a pattern language for selecting names. It intends to facilitate the implementation of the "import" statement or any feature that allows users to select a group of names by patterns.
+"""
+maintainer: "favonia <favonia@gmail.com>"
+authors: "favonia <favonia@gmail.com>"
+license: "Apache-2.0"
+homepage: "https://github.com/RedPRL/yuujinchou"
+bug-reports: "https://github.com/RedPRL/yuujinchou/issues"
+dev-repo: "git+https://github.com/RedPRL/yuujinchou.git"
+depends: [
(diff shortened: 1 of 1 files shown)


packages/yuujinchou/yuujinchou.2.0.0/opam
```diff
@@ -0,0 +1,32 @@
...
+"""
+maintainer: "favonia <favonia@gmail.com>"
+authors: "favonia <favonia@gmail.com>"
+license: "Apache-2.0"
+homepage: "https://github.com/RedPRL/yuujinchou"
+bug-reports: "https://github.com/RedPRL/yuujinchou/issues"
+dev-repo: "git+https://github.com/RedPRL/yuujinchou.git"
+depends: [
+  "dune" {>= "2.7"}
+  "ocaml" {>= "4.12.0" & < "5"}
```
kit-ty-kate, line 14: Is there a specific reason why this isn’t compatible with OCaml 5.0?

kit-ty-kate: I tested locally and this was compiling just fine with OCaml 5.0

favonia: No, but I also don't know whether it would be compatible. I suppose you want me to remove the upper bound?

kit-ty-kate:
Thanks

kit-ty-kate:
Could you return the fix to the opam file upstream so that it can be taken into account for next time?

favonia:
> Could you return the fix to the opam file upstream so that it can be taken into account for next time?

Yes, it was already done. Thanks!

PR #20871, opened 2022-03-09
Relax dependency version constraint in resto-cohttp-server

raphael-proust:
One of the dependency constraint for resto-cohttp-server is tighter than necessary. The project's CI is all green with this change: https://gitlab.com/nomadic-labs/resto/-/pipelines/487831799


pirbo (approved):

kit-ty-kate:
Thanks

PR #20888, opened 2022-03-10
ojs_base_ppx.0.7.0: Remove an unused dependency

kit-ty-kate:
@zoggy could you return this upstream?


zoggy:
Shouldn't I add a version number in constraint, like >= 0.25.0 ?

mseri:
In this case it is not needed since requiring ocaml 4.12 enforces a new enough version of ppxlib

zoggy:
Ok, this is fixed upstream: https://framagit.org/zoggy/ojs-base/-/commit/369ea0f9ff4026a9de490acdceee6495ca781368
Thanks :-)

PR #20889, opened 2022-03-10
Add revised camlimages.5.0.4

reynir:
The revision adds optional dependencies on system packages. Three libraries did not have conf packages and I left them in commented out. Does it make sense to have them in the file even if they don't exist (yet)?

As discussed with @kit-ty-kate on discord.


mseri:
Yes, it is still useful as a hint imo. And can be uncommented later on if these are added

PR #20891, opened 2022-03-10
pcap-format.0.5.2: Remove unused dependency

kit-ty-kate:
Upstreaming in https://github.com/mirage/ocaml-pcap/pull/35


dinosaure:
It seems that we have an error about `bigarray`, I will try to propose a fix about that.

PR #20899, opened 2022-03-11
[new release] let-if (0.3.0)

reynir:
let%if syntax inspired by Rust's if let syntax

- Project page: <a href="https://github.com/reynir/let-if/">https://github.com/reynir/let-if/</a>

##### CHANGES:

* Add new if%true that is a if expression that returns the value of the condition.

* Refactor let%if. Error locations are slightly different.


reynir:
I am not eager to dig up what is the earliest version of dune that ppxlib.0.6.0 supports. I intentionally ignored the opam lint warning since I don't want to uppercase OCaml syntax. I will think how I might rephrase it so it doesn't begin with `let%if`...

kit-ty-kate:
Thanks!

kit-ty-kate:
Could you return the fixes in the opam file upstream so that they can be taken into account next time?

PR #20886, opened 2022-03-10
lablqml.0.7: Unlock ppxlib >= 0.23.0

kit-ty-kate:
Upstreaming in: https://github.com/Kakadu/lablqml/pull/66

Necessary for OCaml 4.14 support
cc @Kakadu

PR #20908, opened 2022-03-12
Fix installation of conf-qt on Mac

Kakadu:
Related to https://github.com/ocaml/opam-repository/pull/20886

MacOS uses weird installation locations, and we need to patch the environment
using `setenv`. Issues:

* Opam warns that this feature should be used only by compilers
* There is no way to apply setenv conitionally (guards are not allowed here).

````
setenv: [
  [ PKG_CONFIG_PATH += "/usr/local/opt/qt@5/lib/pkgconfig" ]
      # { os="macos" } #not allowed here
] # { os="macos" } #not allowed here too
````

### DIFF
--- a/packages/conf-qt/conf-qt.5.2.1/opam
+++ b/packages/conf-qt/conf-qt.5.2.1/opam
@@ -10,11 +10,11 @@ flags: conf
 
 # TODO: There is a `setenv` syntax that should patch the environment for dependencies of this package
 # At the moment it is reserved only for compiler packages, and it's usage gives a warning.
-# But in principle it could simplyfy the life of user of installed Qt
+# But in principle it could simplify the life of the user of installed Qt
 #setenv: [ PATH += "/usr/lib64/qt5/bin:/usr/lib/qt5/bin"] #{ os-distribution = "alpine" | os-distribution = "centos" | os-distribution = "fedora" }
-setenv: [
-  [ PKG_CONFIG_PATH += "/usr/local/opt/qt@5/lib/pkgconfig" ] 
-] #{ os="macos" }
+#setenv: [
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/conf-qt/conf-qt.5.2.1/opam [hunk 1]
```diff
@@ -9,6 +9,13 @@ dev-repo: "git+https://github.com/Kakadu/lablqml.git"
 
 #setenv: [ PATH += "/usr/lib64/qt5/bin:/usr/lib/qt5/bin"] #{ os-distribution = "alpine" | os-distribution = "centos" | os-distribution = "fedora" }
 
+# MacOS uses weird installation locations, and we need to patch the environment. Issues:
+# 1) Opam warns that this feature should be used only by compilers
+# 2) There is no way to apply setenv conitionally (guards are not allowed here).
+setenv: [
+  [ PKG_CONFIG_PATH += "/usr/local/opt/qt@5/lib/pkgconfig" ] # { os="macos" } #not allowed here
```
mseri, line 16: This is just for homebrew though, right?

Kakadu: Maybe. I have never seriously used a Mac, so I apply (kind of random) changes are just to make CI happy

Kakadu:
@mseri ^^

mseri:
The macos CI is not really reliable right now. It is going to be disabled until we are able to fix it.

This change fixes only homebrew on x86_64. Would not work on M1 or for users using macports. I would rather add a `post-install {failure}` message explaining that users may need to extend their paths accordingly, for example on intel homewbrew one needs `PATH=.... opam install ...`, instead.

PR #20911, opened 2022-03-13
Fix conf-gmp to play nice with Nix

vapourismo:
1. Copies `conf-gmp.4.1.0` from `conf-gmp.4`
2. Adds a depexts entry for Nix (`os-distribution = "nix"`)
3. Switches the checksum algo for extra files to SHA256 

Ultimately version 4.1.0 aims to make it easier to use this package programatically with Nix.


mseri:
I ported the change to the current conf file in a separate PR, thanks!

PR #20912, opened 2022-03-13
Add Nix depext entry for conf-libffi

vapourismo:
1. Copies conf-libffi.2.1.0 from conf-libffi.2.0.0
2. Adds a depexts entry for Nix (os-distribution = "nix")

The goal is to make this package easier to integrate on Nix.

### DIFF
--- a/packages/conf-libffi/conf-libffi.2.1.0/opam
+++ b/packages/conf-libffi/conf-libffi.2.1.0/opam
@@ -1,0 +1,24 @@
+opam-version: "2.0"
+maintainer: "blue-prawn"
+authors: ["Anthony Green"]
+homepage: "http://sourceware.org/libffi/"
+license: "MIT"
+build: ["pkg-config" "libffi"]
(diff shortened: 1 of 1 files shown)


samoht:
@vapourismo do you know how we could test this in our CI? Is there a way to setup nix easily on our CI cluster?

samoht:
Also: why do you need a new package and not modifying the existing package?

vapourismo:
I am currently not aware of robust tooling that makes use of the `depexts` annotation which would allow you to test in CI.

In fact, I am currently working on something to integrate .opam files and the opam-repository more easily into Nix - so I've tested it locally. However, I don't think it's completely ready yet.

> Also: why do you need a new package and not just modify the existing package metadata?

I wasn't sure what's the policy about modifying existing manifests. Happy to integrate into existing versions if needed.

mseri:
I ported the change to the current conf file in a separate PR, thanks!

PR #20914, opened 2022-03-14
Bisect_ppx 2.8.1: code coverage for OCaml and Reason

aantron:
This is a release of aantron/bisect_ppx#399, in which @kit-ty-kate adapted Bisect_ppx to OCaml 5.00.


kit-ty-kate:
Thanks!

PR #20916, opened 2022-03-14
Markup.ml 1.0.3: standards-compliant HTML5 parser

aantron:
This is a release of aantron/markup.ml#68, in which @kit-ty-kate adapted Markup.ml to OCaml 5.00.


kit-ty-kate:
Thanks!

PR #20929, opened 2022-03-15
Package baguette_sharp.2.0.4

coco33920:
### `baguette_sharp.2.0.4`
The Baguette# Interpreter REPL
The REPL for Baguette#

---
* Homepage: https://baguettesharp.fr
* Source repo: git+https://github.com/coco33920/ocaml-baguettesharp-interpreter.git
* Bug tracker: https://github.com/coco33920/ocaml-baguettesharp-interpreter/issues

---

Forgot the message sorry! 
*Finally* with some help I managed to redefine the work tree and have only one library which install the language and the interpreter (under baguette_sharp.repl)!
It should pass the CI now!
Thank you,
Charlotte.


coco33920:
I wonder why it fails at linenoise ? It is present in the opam repository :thinking:

coco33920:
It seemed to be that I forgot the linenoise dep in the dependencies now the builds returns clear :)

mseri:
Thanks

packages/baguette_sharp/baguette_sharp.2.0.4/opam
```diff
@@ -0,0 +1,42 @@
...
+maintainer: ["Charlotte Thomas"]
+authors: ["Charlotte Thomas"]
+license: "MIT"
+homepage: "https://baguettesharp.fr"
+doc: "https://github.com/coco33920/ocaml-baguettesharp-interpreter/wiki"
+bug-reports:
+  "https://github.com/coco33920/ocaml-baguettesharp-interpreter/issues"
+depends: [
+  "dune" {>= "3.0"}
+  "ocaml" {>="4.13.1"}
```
mseri, line 14:
```suggestion
  "ocaml" {>="4.13.1"}
```
Do you actually need such a struct bound or the package can work on older ocaml versions? Looking at the code I could not see any obvious reason for this

coco33920: I use `String.starts_with` that was added in OCaml 4.13 according to the changelog.
If its necessary I can however implement them locally to support OCaml 4.12 maybe 4.11 I don't know if i use much 4.12 only features

coco33920: yeah just tested I can easily lower to 4.12 i'm downloading 4.11.1 to see if that's easily doable too (not checked the changelog yet). Is that preferable ?

mseri: Don’t worry it is nit a problem, it was just to check if we could relax the bound here or not. If you use functionalities from 4.13 I am happy to merge as is

coco33920: I'm happy to merge as it is now too, I do not see the urge to implement backward compatibility now. Maybe I will in an eventual future update after the concours when i'll have time to expand this language :)
Thanks you!
Charlotte

PR #20963, opened 2022-03-17
fix conf-perl-XX build for macos homebrew

chetmurthy:
The build scripts in conf-perl-XX packages were wrong for homebrew.  This should fix.

### DIFF
--- a/packages/conf-perl-ipc-system-simple/conf-perl-ipc-system-simple.3/opam
+++ b/packages/conf-perl-ipc-system-simple/conf-perl-ipc-system-simple.3/opam
@@ -1,0 +1,30 @@
+opam-version: "2.0"
+maintainer: "chetsky@gmail.com"
+homepage: "https://www.perl.org/"
+bug-reports: "chesky@gmail.com"
+license: "GPL-1.0-or-later"
+authors: "Larry Wall et. al."
+depends: [
+  "conf-perl"
+]
+build: [
+  ["perl" "--version"]
+  ["perl" "-MIPC::System::Simple" "-e" "1"] { os-distribution != "homebrew" }
+]
+depexts: [
+  ["libipc-system-simple-perl"] {os-family = "debian"}
+  ["perl-ipc-system-simple"] {os-distribution = "alpine"}
+  ["epel-release" "perl-IPC-System-Simple"] {os-distribution = "centos"}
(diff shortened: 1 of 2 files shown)


mseri:
Thanks

Kakadu:
@chetmurthy Shouldn't this PR specify some depexts when os-distribution is 'homebrew'? I got a student that can't install perl-related camlp5 dependencies on MacOS.

chetmurthy:
It doesn't because we don't know what to do there, and in the case of camlp5, these dependencies are only used for tests and mkcamlp5[.opt] .

And by not specifying anything, if you're on MacOS and can install these dependencies, then the build will just work.  Though obviously the right thing for a homebrew user to do, is one of:
1. switch to MacPorts
2. fix homebrew: this foolishness of "you can't have a brew package that installs Perl modules" is ridiculous.

chetmurthy:
I will note that I used to reject building on homebrew, and this caused problems for others.  So there's no solution that is acceptable to all.

Kakadu:
I understood. 
It would be great if opam recipe will mention explicitly that homebrew is weird, and we should use Macports. Something like 

````
post-messages:  " blah "
    {os = "macos" && os-distribution = "homebrew" }
````

chetmurthy:
Um, as-is, it works though, right?  Your student can install campl5, right?

chetmurthy:
I'm asking b/c I don't like messing with OS-specific details on OSes that I can neither run nor test.  If you can test out the changes you desire and submit a PR, that would work though.

Kakadu:
> Um, as-is, it works though, right? Your student can install campl5, right?
Yes, but mkcamlp5.opt doesn't work, but is needed

> I'm asking b/c I don't like messing with OS-specific details on OSes that I can neither run nor test. If you can test out the changes you desire and submit a PR, that would work though.
I don't have Mac too. It's highly likely that today is a first day I hear about MacPorts.

chetmurthy:
Truthfully, I don't know much about MacPorts, either.  All I know is, that it seems to be better-supported than homebrew.  I'm old enough that the rule I adopt is always "if I can't support the variant operating system I'm using, then I should switch to the standard one".  That has always meant that I figure out how to support the variant, b/c when I started in commercial work, the "standard" was Windows.  And there's no way I was using Winders.  Today, the standard is Linux.

If your student doesn't want to make that investment, there is always another choice: install Virtualbox and run Linux inside that.  It isn't hard.

PR #20971, opened 2022-03-20
Package atomic.0.1

c-cube:
### `atomic.0.1`
Compatibility package for OCaml's Atomic module, from 4.12

---
* Homepage: https://github.com/c-cube/ocaml-atomic/
* Source repo: git+https://github.com/c-cube/ocaml-atomic.git
* Bug tracker: https://github.com/c-cube/ocaml-atomic/issues

---

### DIFF
--- a/packages/atomic/atomic.0.1/opam
+++ b/packages/atomic/atomic.0.1/opam
@@ -1,0 +1,24 @@
+opam-version: "2.0"
+synopsis:
+  "Compatibility package for OCaml's Atomic module, from 4.12"
+maintainer: "simon.cruanes.2007@m4x.org"
+license: "LGPL-2.1"
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+]
+depends: [
+  "ocaml" {>= "4.03" & < "4.12"}
+  "dune" {>= "1.1.0"}
+]
+tags: [ "atomic" "threads" "compatibility" ]
+homepage: "https://github.com/c-cube/ocaml-atomic/"
+bug-reports: "https://github.com/c-cube/ocaml-atomic/issues"
+dev-repo: "git+https://github.com/c-cube/ocaml-atomic.git"
+authors: "OCaml maintainers"
+url {
+  src: "https://github.com/c-cube/ocaml-atomic/archive/v0.1.tar.gz"
+  checksum: [
+    "md5=ef7764c9932e403faaec545372e54cb2"
+    "sha512=61746b050033f700f9a41ce0c68efa8d007abc9ef490001a1108a646c1b7ffdaa08f0df659b28f2a396cee011cb25aa52a7a65be3d217697f70b66b634cbe956"
+  ]
+}
--- a/packages/atomic/atomic.base/files/META.atomic
+++ b/packages/atomic/atomic.base/files/META.atomic
@@ -1,0 +1,4 @@
+name="atomic"
+version="[distributed with OCaml 4.12 or above]"
+description="dummy backward-compatibility package for Atomic"
+requires=""
--- a/packages/atomic/atomic.base/files/atomic.install
+++ b/packages/atomic/atomic.base/files/atomic.install
@@ -1,0 +1,3 @@
+lib:[
+  "META.atomic" {"META"}
+]
--- a/packages/atomic/atomic.base/opam
+++ b/packages/atomic/atomic.base/opam
@@ -1,0 +1,15 @@
+opam-version: "2.0"
+maintainer: " "
+authors: " "
+homepage: " "
+depends: [
+  "ocaml" {>= "4.12.0"}
+]
+dev-repo: "git+https://github.com/ocaml/ocaml.git"
+bug-reports: "https://caml.inria.fr/mantis/main_page.php"
+synopsis:
+  "Compatibility package for OCaml's Atomic module starting from 4.12"
+extra-files: [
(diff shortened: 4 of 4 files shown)


c-cube:
cc @gasche on that one, I think it's fine if it's compatible with ocaml >= 4.12, but we could also have an empty package from there up, like for `seq`.

mseri:
@gasche is it ok to merge this?

gasche:
I'm not sure why you are asking me!

I looked at [the code](https://github.com/c-cube/ocaml-atomic/blob/0174918/src/gen.ml) and I have some questions:
- Wouldn't it be better to have `atomic.mli` be defined with `type 'a t = 'a Atomic.t` under >=4.12? It sounds problematic to introduce two incompatible `Atomic.t` types.
- What is the compatibility guarantee in the future, if new Atomic operations are added? Will this module also make them available to older OCaml versions? (This is what the phrasing of the description suggests.)
- Why are you using `[@inline]` everywhere? It looks useless in some cases and possibly harmful in others (the upstream versions use `[@inline never]` for `exchange`, `compare_and_set`, `fetch_and_add` with [a comment](https://github.com/ocaml/ocaml/blob/4.12/stdlib/camlinternalAtomic.ml#L31-L33), see https://github.com/ocaml/ocaml/pull/10035).

c-cube:
@gasche you're unlucky enough that you're the OCaml maintainer whose name pops first in my head :-).

- ideally on >= 4.12 one would just drop this module. I just don't know how it's done for `Seq`. I'll definitely do what you say in the mean time
- compatibility: I imagine I'd update this module so it's a complete stub.
-  for the inline, interesting. Would it be valid to wrap the critical section in `Sys.opaque_identity`?

gasche:
> Would it be valid to wrap the critical section in `Sys.opaque_identity` instead?

I'm not completely sure, you probably need an flambda expert to tell. I think that sticking with the upstream version is safer -- and it's not like these functions are performance-critical anyway.

gasche:
But here we are talking about the sequential emulation of the atomics, which is going to be miles faster than the real atomic operations anyway, but also used rarely as we expect people in practice to migrate to the multicore runtime. Having a direct function call is perfectly fine, the overall cost will still be noticeably lower than with the "real" version -- even when the multicore version is measured in uncontended scenarios, which could be very common.

gasche:
Re. `Sys.opaque_identity`: I think that it may be the case that `Sys.opaque_identity (let x = a in b)` is understood as equivalent to `let x = a in Sys.opaque_identity b`, this is what call-by-value suggests and also validated by semantics of `Sys.opaque_identity` I've read, and this suggests that `Sys.opaque_identity` is *not* safe here.

Gbury:
I'm not an expert, but indeed, in this case, the `[@inline never]` seems to be the best way to ensure proper semantics.

I guess that adding something like `let x = opaque_identity x in..` at the start of `exchange` might prevent flambda from sinking the allocation inside the critical section but I'm not completely sure, so for the moment I'd say keeping the inline annotation is safer. I'll try and discuss that some more with other member of the flambda team (but that might take a while).

mseri:
Please let me know once the implementation is agreed upon and this can be merged

gasche:
> ideally on >= 4.12 one would just drop this module. I just don't know how it's done for Seq.

Wait @c-cube, you were the one who designed the `seq` compatibility package, right?

- [seq.base/opam](https://github.com/ocaml/opam-repository/blob/master/packages/seq/seq.base/opam)
- [seq.base/files/META.seq](https://github.com/ocaml/opam-repository/blob/master/packages/seq/seq.base/files/META.seq)
- [seq.base/files/seq.install](https://github.com/ocaml/opam-repository/blob/master/packages/seq/seq.base/files/seq.install)

mseri:
The macos runner are unreliable, don’t worry about that

gasche:
Note (possibly for 0.2): now that the package is restricted to (<4.12), you can simplify the implementation by removing all the metaprogramming stuff.

mseri:
Do you know why the nnp and nnpchecker tests are failing with
```
[NOTE] atomic.0.1 is not installed.

Nothing to do.
[ERROR] Package conflict!
```
?

c-cube:
I don't know what nnp and nnpchecker even are :/

mseri:
They denote the no-naked-pointers and no-naked-pointers checker versions of the compiler afaik. Sorry for being cryptic. I don't think this is blocking the release, just wanted to double check if you have any idea

c-cube:
sorry, I have indeed no idea :sweat_smile: . I might remove the metaprogramming stuff before you merge, like @gasche suggested.

mseri:
Perfect. I’ll wait for that then

c-cube:
idk why it fails :shrug:

mseri:
@kit-ty-kate do you have any clue?

kit-ty-kate:
This looks quite strange. It looks like the package makes opam infinit looks in some conditions :/ I’m really not sure why.

mseri:
I think it may have to do with the fact that the library version 0.1 is only compatible with ocaml < 4.12 and in all the failing switches we use a special version of 4.13. Something there may get confused and produces the wrong error. In any case the correct base package is installed fine in those switches, so this is not a real issue

mseri:
I am going to merge. Thanks for the patience!

c-cube:
Thank you @mseri , and thanks @gasche for answering my prompt :)

PR #20974, opened 2022-03-21
mirage-crypto-rng 0.7.0: mark unavailable (to reduce revdeps)

hannesm:
//cc @kit-ty-kate as noticed in https://github.com/ocaml/opam-repository/pull/20581#issuecomment-1024919336

I hope this will help the CI system to have less load. The 0.7.0 release is very old (and not maintained anyways) --> it'd as wel be fine with me to completely remove it from opam-repository.

Yes, some will talk about "opam lock" -- but tbh the lock feature is ill-designed if it does not include the entire metadata from opam-repo or a commit of the repositories.


samoht:
Let's remove that package from now - I don't know how `avoid-version` works exactly (does it trim the package universe before solving the dependencies) but it's clearly annoying to have a package whose reverse dependency is the whole universe.

PR #20630, opened 2022-02-03
Package pyml_bindgen.0.2.0

mooreryan:
### `pyml_bindgen.0.2.0`
Generate pyml bindings from OCaml value specifications

---
* Homepage: https://github.com/mooreryan/ocaml_python_bindgen
* Source repo: git+https://github.com/mooreryan/ocaml_python_bindgen.git
* Bug tracker: https://github.com/mooreryan/ocaml_python_bindgen/issues

---
:camel: Pull-request generated by opam-publish v2.1.0

PR #20979, opened 2022-03-21
Package pyml_bindgen.0.3.0

mooreryan:
### `pyml_bindgen.0.3.0`

---

---

### DIFF
--- a/packages/pyml_bindgen/pyml_bindgen.0.3.0/opam
+++ b/packages/pyml_bindgen/pyml_bindgen.0.3.0/opam
@@ -26,9 +26,6 @@ depends: [
   "ppx_expect" {>= "v0.12" & with-test}
   "pyml" {with-test}
   "shexp" {>= "v0.14" & with-test}
-  "bisect_ppx" {dev}
-  "core" {>= "v0.12" & dev}
-  "core_bench" {>= "v0.12" & dev}
   "odoc" {with-doc}
 ]
 build: [
@@ -40,11 +37,12 @@ build: [
     name
     "-j"
     jobs
-    "@install" {arch != "x86_32"}
-    "@runtest" {with-test & arch != "x86_32"}
+    "@install"
+    "@runtest" {with-test}
     "@doc" {with-doc}
   ]
 ]
+available: arch != "x86_32"
 dev-repo: "git+https://github.com/mooreryan/ocaml_python_bindgen.git"
 url {
   src:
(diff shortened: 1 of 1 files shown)


mooreryan:
Note that [these issues](https://github.com/ocaml/opam-repository/pull/20630#issuecomment-1029744502) probably still apply....

packages/pyml_bindgen/pyml_bindgen.0.3.0/opam
```diff
@@ -0,0 +1,56 @@
...
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install" {arch != "x86_32"}
+    "@runtest" {with-test & arch != "x86_32"}
+    "@doc" {with-doc}
+  ]
+]
```
kit-ty-kate, line 47: This packages does not install anything on x86_32, it should be disabled instead not empty.
```suggestion
    "@install"
    "@runtest" {with-test}
    "@doc" {with-doc}
  ]
]
available: arch != "x86_32"
```

packages/pyml_bindgen/pyml_bindgen.0.3.0/opam
```diff
@@ -0,0 +1,56 @@
...
+  "core_kernel" {>= "v0.12" & with-test}
+  "ocamlformat" {>= "0.20.0" & < "0.21.0" & with-test}
+  "ppx_assert" {>= "v0.12" & with-test}
+  "ppx_inline_test" {>= "v0.12" & with-test}
+  "ppx_expect" {>= "v0.12" & with-test}
+  "pyml" {with-test}
+  "shexp" {>= "v0.14" & with-test}
+  "bisect_ppx" {dev}
+  "core" {>= "v0.12" & dev}
+  "core_bench" {>= "v0.12" & dev}
```
kit-ty-kate, line 31: These dependencies don’t seem to be needed anywhere but for bench runs which aren’t in opam.
```suggestion
```

mooreryan: I appreciate the review, kit-ty-kate.  I was under the impression that marking these deps as `dev` was how to signal that you only need them if e.g., installing from github or something like that.  As in dependencies used for development (eg bisect_ppx for testing, the other two for benching).  If they are removed, and I try to install it from github in a fresh switch with `opam install . --deps-only --with-doc --with-test`, then the instrumented testing and the benching won't work without installing them by hand.  It's not that big of a deal, as I could put an extra message in the readme about it.   

I did check on bisect_ppx and core_bench and there are quite a few other repositories that have them either as `with-test` or `dev` dependencies (ke and dream are two for example).  It is just a bit unclear to me given the docs and the other examples of packages in opam as to when to include and not include these types of dependencies in `dev` or `with-test` if that is the case.

All that being said, I am fine with removing them if you think it best.

PR #20991, opened 2022-03-22
[new package] ocaml-in-python.0.1.0

thierry-martinez:
This pull request publishes the first release of `ocaml-in-python`, a
Python package for automatic bindings of OCaml modules.


thierry-martinez:
I think remaining errors are ok: lint complains about the license "PSF-2.0", despite the license is listed in https://spdx.org/licenses/ (reported in #21017), and `macos-homebrew` has a failure on open that seems to be related to opam itself (the same kind of failure that I observed in #20992, but I don't know what is going on).

mseri:
Thanks!

PR #20986, opened 2022-03-22
Add ocaml-jupyter v2.7.9 and jupyter-archimedes v2.7.3

akabe:
Release notes:
https://github.com/akabe/ocaml-jupyter/releases/tag/v2.7.9
https://github.com/akabe/jupyter-archimedes/releases/tag/v2.7.3

PR #20994, opened 2022-03-23
Fix ocaml-jupyter lower bounds and related packages

mseri:
Should resolve many of the failures in #20986


mseri:
There are still some issues pending on old releases, but I have not idea how to fix them for the moment

PR #21001, opened 2022-03-24
add coq.8.15.1

palmskog:
New 8.15 minor release of Coq.

cc: @MSoegtropIMC @skyskimmer


palmskog:
As far as I can see, the only failures are:
- skipping due to too old OCaml version
- reached a timeout on running tests on flambda-fp
- a failure to compile GMP on "macos-homebrew (experimental)" (unrelated to Coq itself)

In my view, these failures are not fatal and the package could still be merged as-is.

PR #21002, opened 2022-03-24
Release Caqti 1.8.0.

paurkedal:
I would like to release [Caqti 1.8.0](https://github.com/paurkedal/ocaml-caqti/releases/tag/v1.8.0).


paurkedal:
Though I don't see which version is used, the warnings from the linter must be due to a limitation fixed in https://github.com/ocaml/opam/commit/57014721d9862a6bd81fbbb56cf4caa72c99c448.

mseri:
I think the new spdx parser is not yet used in the opam of the CI. Looks good, thanks

PR #21010, opened 2022-03-24
[new release] atds, atdpy, atdj, atdgen, atdgen-runtime, atdgen-codec-runtime and atd (2.4.0)

mjambon:
ATD Code generator for Scala

- Project page: <a href="https://github.com/ahrefs/atd">https://github.com/ahrefs/atd</a>

##### CHANGES:

* atdpy: allow custom imports and class decorators to be added to the
  generated Python code.


mjambon:
The tests for `atdpy` are expected to fail because some python dependencies will be missing (mypy, pytest). I don't know why `atdpy.opam` contains `    "@runtest" {with-test}` in the dune install command but the other packages don't. For the previous release (2.3.3), I removed it so that no test would fail (!).

I don't particularly mind, though. We have our own CI where all the tests run (on just x86_64/Linux).

mjambon:
I removed the `@runtest` target from atdpy's opam file so as to get only green checkmarks.

mseri:
Looks good, thanks!

PR #21015, opened 2022-03-25
mariadb.1.1.4: constraint known build failure

ygrek:


samoht:
Could you also copy the error message (either in the commit message or in a GH comment)? That's useful if we hit that for other packages too.

ygrek:
it  was `Stream` deprecation and apparently warnings as errors in test code :
```
File "examples/blocking/blocking_example.ml", line 60, characters 2-13:
60 |   Stream.iter print_row s;
       ^^^^^^^^^^^
Error (alert deprecated): module Stdlib.Stream
Use the camlp-streams library instead.
Command exited with code 2.
```

kit-ty-kate:
Thanks!

PR #21022, opened 2022-03-27
Update ocaml-freestanding repository

samoht:
We have renamed ocaml-freestanding into ocaml-solo5 and that broke the packages hashes...


samoht:
I'm merging this quickly as lint step is passing (and so the checshum are correct now).

samoht:
I'm merging this quickly as lint step is passing (and so the checshum are correct now).

PR #21040, opened 2022-03-28
OCaml 4.14.0 release

Octachron:
This is the package for OCaml 4.14.0 release.
We have:

* ocaml-base-compiler 4.14.0
* ocaml-variants 4.14.0+options
* ocaml-system 4.14.0
* ocaml-src 4.14.0

I have not added the 4.14.0 `ocaml-manual` package since this require to update v2.ocaml.org. I will send the corresponding package once the manual is available online.

Since we expect more patch versions on 4.14 (as a LTS version), I have also added the 4.14.1 development packages:

* ocaml 4.14.1 meta package
* ocaml-variants.4.14.1+trunk


kit-ty-kate:
Thanks a lot!

PR #21011, opened 2022-03-25
Fix various bounds

mseri:
These issues appear often in the CI

PR #21047, opened 2022-03-29
Further bounds cleanup

mseri:
This removes the upper bound on menhir in webidl 1.0-1.2, added erroneously by #21011

1.0, 1.1 should succeed
1.2 we need to check (UPDATE: it is fine as well)

Once this is checked, I will add the commit adding those bounds to `js_of_ocaml-webidl` which depends on webidl but also vendors a version of webidl which needs the bounds.

### DIFF
--- a/packages/js_of_ocaml-webidl/js_of_ocaml-webidl.0.1/opam
+++ b/packages/js_of_ocaml-webidl/js_of_ocaml-webidl.0.1/opam
@@ -18,7 +18,7 @@ depends: [
   "js_of_ocaml-lwt"
   "js_of_ocaml-ppx"
   "ppx_jane"
-  "menhir" {build & >= "20170607" & < "20211215"}
+  "menhir" {build & >= "20180523" & < "20211215"}
 ]
 build: [
   ["dune" "subst"] {dev}
--- a/packages/js_of_ocaml-webidl/js_of_ocaml-webidl.0.1/opam
+++ b/packages/js_of_ocaml-webidl/js_of_ocaml-webidl.0.1/opam
@@ -18,6 +18,7 @@ depends: [
   "js_of_ocaml-lwt"
   "js_of_ocaml-ppx"
   "ppx_jane"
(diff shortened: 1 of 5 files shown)


packages/webidl/webidl.1.2/opam
```diff
@@ -11,7 +11,7 @@ depends: [
   "ocaml" {>= "4.03.0"}
   "ocamlfind" {build & >= "1.7.1"}
   "ocamlbuild" {build & >= "0.9.3"}
-  "menhir" {build & >= "20170607" & < "20211215"}
+  "menhir" {build & >= "20170607"}
```
kit-ty-kate, line 14: `webidl.1.2` does not use `menhir` at all.
```suggestion
```

mseri: Thanks

packages/js_of_ocaml-webidl/js_of_ocaml-webidl.0.1/opam
```diff
@@ -18,6 +18,7 @@ depends: [
   "js_of_ocaml-lwt"
   "js_of_ocaml-ppx"
   "ppx_jane"
+  "menhir" {build & >= "20170607" & < "20211215"}
```
mseri, line 21:
```suggestion
  "menhir" {build & >= "20180523" & < "20211215"}
```
Uses `--infer-write-query`

packages/js_of_ocaml-webidl/js_of_ocaml-webidl.0.2/opam
```diff
@@ -18,6 +18,7 @@ depends: [
   "js_of_ocaml-lwt"
   "js_of_ocaml-ppx"
   "ppx_jane"
+  "menhir" {build & >= "20170607" & < "20211215"}
```
mseri, line 21:
```suggestion
  "menhir" {build & >= "20180523" & < "20211215"}
```
Uses `--infer-write-query`

PR #21048, opened 2022-03-29
Add new package for Z3 TPTP

MSoegtropIMC:
This PR adds a new package which contains the TPTP front end for Z3. It uses the existing opam z3 package and z3 shared library.

TPTP is a front end common in automatic theorem provers like Eprover. It is used by the Coq Hammer tactic.

The build instructions are not as clean as I would like it, but I couldn't find a way to build it with Z3's python based build system with an existing opam Z3 and the front end is a single C++ file + the Z3 library, so I guess this explicit compiler call might be more robust than hacking into the Z3 build system.

This is yet another of my "why not build interesting C++ software with opam" PRs and quite Coq centric. If you think it is not so useful for the wider OCaml world, I can also keep it in Coq's own opam repo, but I always ask here first.

@k4rtik : FYI.

### DIFF
--- a/packages/z3_tptp/z3_tptp.4.8.13/opam
+++ b/packages/z3_tptp/z3_tptp.4.8.13/opam
@@ -1,0 +1,36 @@
+opam-version: "2.0"
+maintainer: "7895506+MSoegtropIMC@users.noreply.github.com"
+authors: "MSR"
+homepage: "https://github.com/Z3prover/z3"
+bug-reports: "https://github.com/Z3prover/z3/issues"
+license: "MIT"
+dev-repo: "git+https://github.com/Z3prover/z3.git"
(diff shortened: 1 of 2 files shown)


MSoegtropIMC:
Can someone help me to figure out what is wrong with the homebrew builds? It actually works fine with Mac Homebrew in my own CI and I don't see any error message, besides a failed in the end.

k4rtik:
I just tried your package (4.8.14) locally and it works fine. There seems to be a general issue with the macos/homebrew CI, a quick scan of recent PRs in this repo suggests all of them are failing with homebrew.

MSoegtropIMC:
@k4rtik : thanks for the confirmation.

IMHO this is good for merge then (except homebrew all tests passed).

kit-ty-kate:
Thanks!

PR #21050, opened 2022-03-29
Add new version 3.9 for conf-python-3

MSoegtropIMC:
This PR adds a new version for the python 3 conf package (the latest available is 3.1 which is way too old).

I get btw a lint warning:
```
warning 62: License doesn't adhere to the SPDX standard, see https://spdx.org/licenses/: "PSF-2.0"
```
which I would say is a false positive. See (https://spdx.org/licenses/PSF-2.0.html).

Since this package hasn't been updated by the written maintainer, please feel free to put in my Github ID / email.

### DIFF
--- a/packages/conf-python-3/conf-python-3.9.0.0/files/test.py
+++ b/packages/conf-python-3/conf-python-3.9.0.0/files/test.py
@@ -1,0 +1,1 @@
+print('python-3 OK')
--- a/packages/conf-python-3/conf-python-3.9.0.0/opam
+++ b/packages/conf-python-3/conf-python-3.9.0.0/opam
@@ -1,0 +1,33 @@
+opam-version: "2.0"
+maintainer: "unixjunkie@sdf.org"
+homepage: "https://www.python.org/downloads/release/python-3910/"
+authors: "Python Software Foundation"
+license: "PSF-2.0"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+build: ["python3" "test.py"]
+depexts: [
+  ["python3"] {os-family = "debian"}
+  ["python3"] {os-distribution = "nixos"}
+  ["python3"] {os-distribution = "alpine"}
+  ["python39" "epel-release"] {os-distribution = "centos"}
+  ["python3"] {os-distribution = "fedora"}
+  ["python3"] {os-distribution = "ol"}
+  ["python"] {os-distribution = "arch"}
+  ["python3"] {os-family = "suse"}
+  ["dev-lang/python:3.6"] {os-distribution = "gentoo"}
+  ["python3"] {os = "openbsd"}
+  ["lang/python39"] {os = "netbsd"}
+  ["lang/python39"] {os = "freebsd"}
+  ["python39"] {os-distribution = "macports" & os = "macos"}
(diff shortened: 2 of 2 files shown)


mseri:
Thanks. I wonder two things:

1. if we follow `conf-python-3-7` from https://github.com/ocaml/opam-repository/pull/20991/files, we can still use this package with `ocaml-in-python`, so I would suggest to do that.

2. right now `conf-python-3-7` follows `conf-python-2-7` and I thought this was the idea also for `conf-python-3`, by decoupling python version from the `conf-python` package. This would suggest that this package is called `conf-python-3-9`. I wonder if it would be better to deprecate everything and instead have `conf-python` with packages whose version is python's version lower bound. This is going to be imprecise in many cases, but some package managers want the python version specified, so this packages are actually needed from time to time

@kit-ty-kate what do you think?

MSoegtropIMC:
Thanks, I somehow didn't see the con-python-3-7 package. If this is the standard, I will follow it. And yes, I think it might make sense to have a `conf-python` package which just gives a recent python (if I understood you right, this is what you suggested).

mseri:
Thanks. Please do not modify this for the moment, until we have an agreement with the other maintainers of the opam-repository

MSoegtropIMC:
OK, i will wait for the conclusion. It is not urgent for me since my main projects has its own opam repos, and I keep it there until this is resolved here.

MSoegtropIMC:
@mseri : thanks!

Btw.: I should mention that most distributions do not have a way to select specific versions of python, you can just select python2 or python3. Other distros have separate packages for python 3.x.

This is not really satisfactory, but I don't see a way around it.

mseri:
Yes, I know unfortunately. We do what we can :)

MSoegtropIMC:
> We do what we can :)

Which is a lot! I really don't know how I would get my job done without opam.

MSoegtropIMC:
@mseri : some time went by since we last discussed this - as far as I can see there is still no conf-python package. Should I create this?

MSoegtropIMC:
Btw.: the conf-python-3-7 package does not work any more for homebrew - one has to say `python@3.7`- otherwise opam 2.1.x tries to reinstall it again and again since `brew install python@3` installs `python@3.10` and opam looks for `python@3`. I guess this came with the switch to new depext.

So it would be a good time to redo the python packages ...

mseri:
Thanks @MSoegtropIMC for pinging me. I could not free enough time for this yet due to some issues at home and then I went in holiday, planning to get back to it this week or the next. If you want to attempt a PR on the line of what I was saying before, I'd be grateful. Otherwise I should be able to do it in a week or so.

Please, do not remove the old packages: we should add a new version to transition to the new `conf-python`, something like https://github.com/ocaml/opam-repository/pull/18145

MSoegtropIMC:
@mseri : I will adjust my PR to what we discussed (I would include the 3 latest python 3 and the latest python 2 variants (so 4 in total).

I won't remove the old packages but won't add to them either (that is no old style 3.9).

mseri:
We don't have a good solution yet for python packages, so let's merge this as is and sorry for the long wait

PR #21056, opened 2022-03-29
zarith_stubs_js: add conflicts with `zarith` versions

smondet:
I've been bitten by this a couple of times until I figured it out, hope it helps other users; case of builds/CI working and JSOO client code failing at startup.

screenshot-20220329-142252


mseri:
Thanks

PR #21065, opened 2022-03-30
Fix Sihl 0.3.0 tarball src

joseferben:
Related to https://github.com/oxidizing/sihl/issues/533

I don't know why this stopped working, but the correct tarball src can be found under releases.


mseri:
Thanks!

PR #21066, opened 2022-03-30
Package bwd.1.1.0

favonia:
### `bwd.1.1.0`
Backward lists
This OCaml package defines backward lists that are isomorphic to lists.
They are useful when one wishes to give a different type to the lists that are semantically in reverse.
In our experience, it is easy to miss List.rev or misuse List.rev_append when both semantically forward and backward lists are present.
With backward lists having a different type, it is impossible to make these mistakes.

---
* Homepage: https://github.com/RedPRL/ocaml-bwd
* Source repo: git+https://github.com/RedPRL/ocaml-bwd.git
* Bug tracker: https://github.com/RedPRL/ocaml-bwd/issues

---

### DIFF
--- a/packages/bwd/bwd.1.1.0/opam
+++ b/packages/bwd/bwd.1.1.0/opam
@@ -1,0 +1,31 @@
+opam-version: "2.0"
+synopsis: "Backward lists"
+description: """
+This OCaml package defines backward lists that are isomorphic to lists.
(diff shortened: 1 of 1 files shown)


favonia:
The only failure is from "macos-homebrew (experimental)". I might need help in understanding the error (if there's something I should fix).

mseri:
You can safely ignore that. Thanks

PR #21081, opened 2022-04-01
Remove spacetime option from 4.12+

dra27:
Reported by @jmid - the package lists are reporting `ocaml-option-spacetime` as a depopt of the options packages, which is confusing as it's not installable on 4.12+!


kit-ty-kate:
Thanks

PR #21091, opened 2022-04-01
ometrics: remove the dependency to dot-merlin-reader

vch9:
Having dot-merlin-reader and vendoring merlin in my repo causes issues with an opam-mono-repo including ometrics.

Thus I'm trying to remove the dependency to dot-merlin-reader and use the executable from my vendor.
(If I remember correctly, I added the dependency because the opam ci complained about the missing executable)

### DIFF
--- a/packages/ometrics/ometrics.0.1.3/opam
+++ b/packages/ometrics/ometrics.0.1.3/opam
@@ -13,7 +13,6 @@ depends: [
   "menhirSdk"
   "menhirLib"
   "menhir"
-  "dot-merlin-reader" {>= "4.1"}
   "csexp" {>= "1.5.1"}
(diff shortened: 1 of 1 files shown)


vch9:
Apparently the dependency is completely useless. I tried locally to:
1. pin `ometrics` to a commit where `dot-merlin-reader` is removed from the `.opam`
2. ensure that `dot-merlin-reader` isn't available in my path
3. run the `ometrics` executable installed through the pin

mseri:
I tried locally and seems fine, I would just like another pair of eyes on this before merging

vch9:
If it's fine I'll probably remove the dependency for the 3 versions I published

vch9:
Thanks both of you

PR #21093, opened 2022-04-02
The atd library won't build with menhir 20211230

mjambon:
See the upstream issue: https://github.com/ahrefs/atd/pull/272

Sorry for the back and forth on this!


mjambon:
Note that we're having [build performance problems with the latest menhir](https://github.com/returntocorp/pfff/issues/512), so it's important to allow the use of earlier versions as much as possible.

mseri:
Is it just this version that creates problems?

mjambon:
Yes, only version `20211230` has the problem. The versions immediately before and after that are fine.

PR #21095, opened 2022-04-02
[new release] http_async (0.0.2)

anuragsoni:
Reasonably performant http 1.1 server for async

- Project page: <a href="https://github.com/anuragsoni/http_async">https://github.com/anuragsoni/http_async</a>

##### CHANGES:

* Initial release

PR #21100, opened 2022-04-04
http_async can build on arm64 as well

anuragsoni:
I missed `arm64` when proposing the original opam file. The package builds and installs without issues on arm. I can cut a new release instead of modifying the existing opam file if that's preferable.

Reference: https://github.com/ocaml/opam-repository/pull/21095


anuragsoni:
I've fixed the opam file upstream as part of https://github.com/anuragsoni/http_async/commit/a0311ae4cefcc45502365b9d21589ca59eab5402 so the next release will have the proper platform constraint setup.

PR #21108, opened 2022-04-04
[new release] fontforge-of-ocaml (1.1.0)

pbaudin:
OCaml binding of FontForge

- Project page: <a href="http://github.com/pbaudin/fontforge-of-ocaml">http://github.com/pbaudin/fontforge-of-ocaml</a>

### DIFF
--- a/packages/fontforge-of-ocaml/fontforge-of-ocaml.1.1.0/opam
+++ b/packages/fontforge-of-ocaml/fontforge-of-ocaml.1.1.0/opam
@@ -21,7 +21,6 @@ build: [
     "-j"
     jobs
     "@install"
-    "@runtest" {with-test}
     "@doc" {with-doc}
   ]
 ]
--- a/packages/fontforge-of-ocaml/fontforge-of-ocaml.1.1.0/opam
+++ b/packages/fontforge-of-ocaml/fontforge-of-ocaml.1.1.0/opam
@@ -1,0 +1,49 @@
+tags: ["font" "fontforge" "binding"]
+depends: [
+  "dune" {>= "2.8"}
+  "ocaml" {>= "4.09.0"}
+  "base" {>= "v0.12.2"}
+  "fmt" {>= "0.8.9"}
+  "pyml" {>= "20200518"}
+  "ppx_python" {>= "v0.12.0"}
+  "odoc" {with-doc}
+]
+depopts: [
+  "headache" {>= "1.05"}
+]
+build: [
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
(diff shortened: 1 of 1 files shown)


packages/fontforge-of-ocaml/fontforge-of-ocaml.1.1.0/opam
```diff
@@ -0,0 +1,51 @@
+opam-version: "2.0"
```
kit-ty-kate, line 1: We try to at least get `opam-version:` be the first field. I’m not sure how your opam file got several others before

packages/fontforge-of-ocaml/fontforge-of-ocaml.1.1.0/opam
```diff
@@ -0,0 +1,51 @@
...
+  "pyml" {>= "20200518"}
+  "ppx_python" {>= "v0.12.0"}
+  "odoc" {with-doc}
+]
+depopts: [
+  "headache"
+]
+conflicts: [
+  "headache" {< "1.05"}
+]
```
kit-ty-kate, line 29: The semantics of version constraints inside `depopts` is different from the one you are looking for.
The usual semantics that users want is the one provided by adding the constraint in `conflicts`.

In `depopts` it means: "this package optionally depends on this, but only if the version matches the constraint" so it would still be possible for `headache < 1.05` to be installed, it would just not rebuild the package and would break at the next reinstall

kit-ty-kate:
Thanks. Could you return the fixes upstream so that they can be taken into account for next time?

packages/fontforge-of-ocaml/fontforge-of-ocaml.1.1.0/opam
pbaudin: That file was generated automatically by `dune` from a template. In that template, `opam-version: "2.0"` is the first line. 
In fact, `dune` generated automatically all the lines before from the `dune-project` file.

packages/fontforge-of-ocaml/fontforge-of-ocaml.1.1.0/opam
pbaudin: I see. Thank you for the explanation.

packages/fontforge-of-ocaml/fontforge-of-ocaml.1.1.0/opam
kit-ty-kate: Why do you need this template file? As far as I can see everything can be written in the dune-project file (or even better imo, autogeneration is not needed)

PR #21114, opened 2022-04-05
Add upper bound on sexplib0 to memtrace_viewer

lpw25:
It seems that the v0.14 versions of memtrace_viewer do not work with version v0.15 of sexplib0. I'm not sure why the relationship between these packages is expressed as a conflict rather than a dependency, but I'd rather not change it.


kit-ty-kate:
Thanks

kit-ty-kate:
Thanks

PR #21115, opened 2022-04-05
lambdapi: fix install command to install lambdapi libraries too

fblanqui:


fblanqui:
Wait. I am perhaps going to use another solution.

PR #21116, opened 2022-04-05
Package conformist.0.8.1

joseferben:
### `conformist.0.8.1`
Conformist allows you to define schemas to decode, validate and sanitize input data declaratively
Conformist allows you to define schemas to decode, validate and sanitize input data declaratively.
It comes with runtime types for primitive OCaml types such as `int`, `string`, `bool` and `float` but also `Ptime.t`, `option` and custom types.
Re-use business rules in validators and run it on the client side with js_of_ocaml.
Arbitrary meta data can be stored in schemas which is useful to build functionality on top of conformist.

---
* Homepage: https://github.com/oxidizing/conformist
* Source repo: git+https://github.com/oxidizing/conformist.git
* Bug tracker: https://github.com/oxidizing/conformist/issues

---


kit-ty-kate:
Thanks. Could you return the fix upstream so that it can be taken into account for next time?

joseferben:
@kit-ty-kate Thanks! I guess the redundant dune lower limit comes from `(lang dune 2.7)` and `(depends (dune (>= 2.4))` in the dune file. Do you have an idea on how to fix it?

kit-ty-kate:
removing the dune line in `(depends ...)` will work for that

PR #21126, opened 2022-04-06
new package frama-c-lannotate

vprevosto:
opam package for the newly released LAnnotate Frama-C plugin


mseri:
The lower bound errors are all due to frama-c, mergin. Thanks

PR #21128, opened 2022-04-06
new package frama-c-luncov

vprevosto:
opam package for the new Frama-C plugin LUncov


mseri:
The lower bound errors are all due to frama-c, mergin. Thanks

PR #21129, opened 2022-04-06
An hot-fix for git.3.7.0 about the push process

dinosaure:
This fix permits to us to still use `git.3.7.0` (due to `irmin.3.2.0`) and be able to correctly push without a bug. It includes a really small patch which fix the initial issue (which will be surely re-applied into 3.8.0 and 3.9.0).


dinosaure:
From a @kit-ty-kate's advise, it's better to release a `3.{7,8,9}.1`, I will do that tomorrow.

dinosaure:
See #21136

mseri:
Should we close this then?

dinosaure:
Yes, it's fine :+1:.

PR #21139, opened 2022-04-07
Remove three previous releases of feat.

fpottier:
These three releases contain a serious bug in `Enum.sample` so I would like to remove them. Today's release fixes the bug.


mseri:
To be more consistent with past requests, I think it would be better to leave the opam files but mark them `available: false # serious bug in Enum.sample`

fpottier:
Thanks, I have done that.

mseri:
Thanks!

PR #21150, opened 2022-04-08
use zip archive as source of ocaml-lsp-server.1.9.2~4.14preview

Khady:
With the source being `git+https://github.com/kit-ty-kate/ocaml-lsp.git#414` every `opam update` will have to do a git fetch which is kind of slow and adds some noise. So I replaced the url to point to the zip archive generated by github for the same commit.


Khady:
actually gh archives don't contain submodules :(

kit-ty-kate:
The preview will cesse to exist as soon as https://github.com/ocaml/opam-repository/pull/21149 is merged anyway so it’s only a few minutes from being fixed.

PR #21164, opened 2022-04-09
Package diskuvbox.0.1.0

jonahbeckford:
### `diskuvbox.0.1.0`
Cross-platform basic set of script commands
A cross-platform basic set of script commands. Available as a single binary (`diskuvbox`, or `diskuvbox.exe` on Windows) and as an OCaml library.

---
* Homepage: https://github.com/diskuv/diskuvbox
* Source repo: git+https://github.com/diskuv/diskuvbox.git
* Bug tracker: https://github.com/diskuv/diskuvbox/issues

---


packages/diskuvbox/diskuvbox.0.1.0/opam
```diff
@@ -0,0 +1,47 @@
...
+  "dune" {>= "2.9"}
+  "odoc" {>= "1.5.3" & with-doc}
+  "ocaml" {>= "4.09.0"}
+  "ppx_deriving" {>= "5.2.1"}
+  "bos" {>= "0.2.0"}
+  "fmt" {>= "0.8.9"}
+  "logs" {>= "0.7.0"}
+  "mdx" {>= "2.0.0"}
+  "headache" {>= "1.05" & build}
+  "ocamlformat" {>= "0.19.0" & build}
```
mseri, line 21: Is this actually needed for the build?

mseri: Yes: https://github.com/diskuv/diskuvbox/blob/b6b7e6fa65ad53e55dc077cb071ee300fb931fae/src/bin/dune#L16

packages/diskuvbox/diskuvbox.0.1.0/opam
```diff
@@ -0,0 +1,47 @@
...
+  "cmdliner" {<= "1.2.0"}
```
mseri, line 22: Do you already know that the next cmdliner breaks it? If not, I suggest to remove this bound. If a future release of cmdliner breaks the package we will add the bound and inform you
```suggestion
  "cmdliner"
```

jonahbeckford: I'll replace the bound in dune-project so it doesn't get overwritten

mseri:
Thanks

PR #21191, opened 2022-04-12
new file:   packages/line_oriented/line_oriented.1.3.0/opam

UnixJunkie:


mseri:
No, but this error is not due to lower bounds in the packages, is more likely something we need to fix in the num library management, which is a bit special. 

You don’t need to do anything for this one

mseri:
Trying lower bound by hand is a mix of reiterated attempts at

https://github.com/ocaml/opam-repository/wiki/Lower-bound-check---DYI

and looking at changelogs or v3.ocaml.org package documentations, to see when specific api or signature changes happens

PR #21202, opened 2022-04-13
Nix depexts for conf-hidapi, conf-libev and conf-postgresql

vapourismo:
This adds `depexts` entries so that these may be built on Nix.


kit-ty-kate:
Thanks

PR #21204, opened 2022-04-14
Add conf-fts.1

emillon:
This is a conf package for fts.h, which is packaged separately in Alpine for example.


kit-ty-kate:
Thanks

PR #21210, opened 2022-04-14
opam-file-format: Pass %{lib}% var explicitly to Makefile

vapourismo:
The Makefile derives `LIBDIR=%{prefix}%/lib` which isn't always correct.

This only affects `opam-file-format` prior to version 2.1. Does this need to be backported?

### DIFF
--- a/packages/opam-file-format/opam-file-format.2.0.0/opam
+++ b/packages/opam-file-format/opam-file-format.2.0.0/opam
@@ -10,7 +10,7 @@ build: [
   "byte" {!ocaml:native}
   "all" {ocaml:native}
 ]
-install: [make "install" "PREFIX=%{prefix}%"]
+install: [make "install" "PREFIX=%{prefix}%" "LIBDIR=%{lib}%"]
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
I’m not sure to follow. How is it not always correct in opam? It should be

vapourismo:
When using opam (the executable) `%{lib}%` happens to be exactly `%{prefix}%/lib`.

However, when not using `opam`, for instance when instrumenting `opam-installer` independently, this condition no longer holds. `dune`, `ocamlbuild` and `topkg` (and `opam-installer` for that matter) seem to respect that difference AFAICT. It is only custom build systems like the one for `opam-file-format` prior to 2.1 that don't.

To put this into context, in Nix, the variables are populated like this:
  - `prefix=$out`
  - `lib=$out/lib/ocaml/$ocaml_version/site-lib` (ocamlfind/findlib dictates this)

I personally think that variable setting is a bit unfortunate but changing this upstream is incredibly involved. So changing the OCaml packages with custom installation mechanisms seems easier for now :smirk:

PR #21211, opened 2022-04-14
conf-rust: Require 'rustc' depext on Nix

vapourismo:
Cargo does not ship with the Rust compiler itself on Nix.


kit-ty-kate:
Thanks!

PR #21227, opened 2022-04-18
magic-trace 1.0.1

cgaebel:
New version of magic-trace: 1.0 (and 1.0.1 to fix a small error in the dune file). Release notes are here:

https://github.com/janestreet/magic-trace/releases/tag/v1.0.0

### DIFF
--- a/packages/magic-trace/magic-trace.1.0.1/opam
+++ b/packages/magic-trace/magic-trace.1.0.1/opam
@@ -1,0 +1,35 @@
+opam-version: "2.0"
+maintainer: "Jane Street developers"
+authors: ["Jane Street Group, LLC"]
+homepage: "https://magic-trace.org"
+bug-reports: "https://github.com/janestreet/magic-trace/issues"
+dev-repo: "git+https://github.com/janestreet/magic-trace.git"
+doc: "https://github.com/janestreet/magic-trace/wiki"
+license: "MIT"
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+]
+depends: [
+  "ocaml"        {>= "4.12"}
+  "async"
+  "cohttp"
+  "cohttp_static_handler"
+  "core"
+  "core_unix"
+  "expect_test_helpers_async"
+  "ppx_jane"
+  "shell"
+  "dune"         {>= "2.0.0"}
+  "owee"         {>= "0.6"}
+  "re"           {>= "1.8.0"}
+]
+synopsis: "Collects and displays high-resolution traces of what a process is doing"
+description: "https://github.com/janestreet/magic-trace"
+url {
+  src: "https://github.com/janestreet/magic-trace/archive/refs/tags/v1.0.1.tar.gz"
+  checksum: [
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
```
#=== ERROR while compiling magic-trace.v1.0.0 =================================#
# context              2.1.2 | linux/x86_64 | ocaml-base-compiler.4.14.0 | pinned(https://github.com/janestreet/magic-trace/archive/refs/tags/v1.0.0.tar.gz)
# path                 ~/.opam/4.14/.opam-switch/build/magic-trace.v1.0.0
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p magic-trace -j 31
# exit-code            1
# env-file             ~/.opam/log/magic-trace-16777-ed4a61.env
# output-file          ~/.opam/log/magic-trace-16777-ed4a61.out
### output ###
# File "src/dune", line 8, characters 29-50:
# 8 |    core_unix.sys_unix cohttp cohttp_static_handler core_unix.signal_unix
#                                  ^^^^^^^^^^^^^^^^^^^^^
# Error: Library "cohttp_static_handler" not found.
# -> required by library "magic-trace.magic_trace_lib" in _build/default/src
# -> required by _build/default/META.magic-trace
# -> required by _build/install/default/lib/magic-trace/META
# -> required by _build/default/magic-trace.install
# (cd _build/default/bin && ./gen-versions.sh version_stubs.c)
# fatal: not a git repository (or any parent up to mount point /)
# Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```

packages/magic-trace/magic-trace.v1.0.0/opam
```diff
@@ -0,0 +1,35 @@
+opam-version: "2.0"
```
kit-ty-kate, line 1: The version number should not have the `v` prefix. Could you rename it to `1.0.0` instead?

cgaebel:
We've vendored a copy of the `tracing` and `fzf` packages, and I think that's breaking CI. What's the standard procedure for dealing with that?

kit-ty-kate:
Thanks

gasche:
I think there is a small issue with the new magic-trace release: the previous version has version number `v0.15.0`, and the new release has version number `1.0.1`, and it looks like opam orders them `1.0.1 < v0.15.0`, making v0.15.0 understood as the latest version (see https://opam.ocaml.org/packages/magic-trace/ for example). Users are still able to select 1.0.1 explicitly (marked as a "downgrade"), but ideally this ordering issue would be fixed.

I'm not sure what's the best fix, but I suppose @kit-ty-kate may have good advice here.

kit-ty-kate:
I'm very confused why a package from the janestreet suite suddenly broke out of the "version glue" they have together. What's the story behind this? Why wasn't it released as v0.15.1?

kit-ty-kate:
cc @aalekseyev

cgaebel:
We're developing magic-trace in the open, separate from the rest of the public Jane Street release process.

Would it be possible to delete magic-trace v0.15? It has few, if any users, and anyone who uses it certainly wants 1.0.1 instead.

PR #21266, opened 2022-04-20
Add a constraint on mirage.4.0.0 about opam-monorepo.0.3.0

dinosaure:
It's not obvious but the usage of `opam monorepo` by `mirage` is not compatible with `opam-monorepo.3.0.0`.

### DIFF
--- a/packages/mirage/mirage.4.0.0/opam
+++ b/packages/mirage/mirage.4.0.0/opam
@@ -26,7 +26,7 @@ depends: [
   "astring"
   "logs"
   "mirage-runtime" {= version}
-  "opam-monorepo" {>= "0.2.6" & < "3.0.0"}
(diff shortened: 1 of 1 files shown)


packages/mirage/mirage.4.0.0/opam
```diff
@@ -26,7 +26,7 @@ depends: [
   "astring"
   "logs"
   "mirage-runtime" {= version}
-  "opam-monorepo" {>= "0.2.6"}
+  "opam-monorepo" {>= "0.2.6" & < "3.0.0"}
```
kit-ty-kate, line 29:
```suggestion
  "opam-monorepo" {>= "0.2.6" & < "0.3.0"}
```

kit-ty-kate:
cc @NathanReb

PR #21270, opened 2022-04-21
Remove magic-trace v15

cgaebel:
That version is impossibly out of date. Anyone who uses it should
instead use magic-trace 1.0.0.

Normally, I wouldn't yank this. If only for the historically curious.
But, per the discussion at #21227, it's causing issues for people
because opam sorts v15 as > 1.0.0.


kit-ty-kate:
This is very exceptional but the package has only been added last month, isn’t used by anyone and is obviously a mistake so let’s go with it. Thanks.

PR #21278, opened 2022-04-24
misuja: update url, the old one is no longer accessible

mseri:
Ping @smondet @toots 
Unfortunately I don't have a way to compare the contents (after logging in, the old url fails with 404 not found), is one of you able to do that, or to make the old tarball available again?

### DIFF
--- a/packages/misuja/misuja.0.0.0/opam
+++ b/packages/misuja/misuja.0.0.0/opam
@@ -31,6 +31,6 @@ which is manipulated with an OCaml API (communicating through
 ring-buffers provided by the Jack API)."""
 url {
   src:
-    "https://gitlab.com/smondet/misuja/repository/misuja.0.0.0/archive.tar.gz"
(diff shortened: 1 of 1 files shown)


toots:
I don't that I am related to this package.. 🙂

mseri:
Sorry, I had seen your name as the last one that modified the package (https://github.com/ocaml/opam-repository/commit/3173fdee31574f021df182e50b796bc3f0f546bd) :D

smondet:
I had it in my ./opam/download-cache

[a30de68e28d4edd1073b1bb6110e613b.tar.gz](https://github.com/ocaml/opam-repository/files/8552772/a30de68e28d4edd1073b1bb6110e613b.tar.gz)

The tag has not moved: https://gitlab.com/smondet/misuja/-/tags/misuja.0.0.0

mseri:
Thanks a lot for checking! Indeed the content is identical

PR #21279, opened 2022-04-25
Package ott.0.32

tperami:
# `ott.0.32`

Homepage: http://www.cl.cam.ac.uk/~pes20/ott/
Source repo: https://github.com/ott-lang/ott
Bugs: https://github.com/ott-lang/ott/issues
Changelog: https://github.com/ott-lang/ott/blob/master/revision_history.txt

### DIFF
--- a/packages/ott/ott.0.32/opam
+++ b/packages/ott/ott.0.32/opam
@@ -1,7 +1,7 @@
 opam-version: "2.0"
 maintainer: "Thibaut Pérami <Thibaut.Perami@cl.cam.ac.uk>"
 authors: ["Peter Sewell" "Francesco Zappa Nardelli" "Scott Owens"]
-license: "part BSD3, part LGPL 2.1"
+license: ["BSD-3-Clause" "LGPL-2.1-only"]
 homepage: "http://www.cl.cam.ac.uk/~pes20/ott/"
 bug-reports: "https://github.com/ott-lang/ott/issues"
 depends: [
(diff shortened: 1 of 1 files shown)


packages/ott/ott.0.32/opam
```diff
@@ -0,0 +1,47 @@
+opam-version: "2.0"
+maintainer: "Thibaut Pérami <Thibaut.Perami@cl.cam.ac.uk>"
+authors: ["Peter Sewell" "Francesco Zappa Nardelli" "Scott Owens"]
+license: "part BSD3, part LGPL 2.1"
```
kit-ty-kate, line 4:
```suggestion
license: ["BSD-3-Clause" "LGPL-2.1-only"]
```

tperami:
Just checking, but is having two licence in the licence field like `license: ["BSD-3-Clause" "LGPL-2.1-only"]` representative that some parts of the code are under one licence and other parts of the code are under the other licence? I was just afraid that it would implicitely mean that all the code can be available under any of those two licences

kit-ty-kate:
It would be equivalent to `BSD-3-Clause AND LGPL-2.1-only` using the SPDX description format (see https://github.com/ocaml/opam/blob/9974973ece9c8fb20afa0cb5283328dc4bf92af1/doc/pages/Manual.md#opam for the the opam specific documentation)

tperami:
I need to check with the authors if they are fine with that, don't merge now, please

kit-ty-kate:
Thanks. Could you return the fix to the opam file upstream so it can be taken into account for next time?

tperami:
Yes, I will do that, no problem

PR #21290, opened 2022-04-28
Add package aws-config-0.0.1

Nymphium:
This PR adds new package aws-config which reads aws configuration from config file / environment variables.
https://github.com/Nymphium/aws-config

### DIFF
--- a/packages/aws-config/aws-config.0.0.1/opam
+++ b/packages/aws-config/aws-config.0.0.1/opam
@@ -9,10 +9,9 @@ dev-repo: "git+https://github.com/Nymphium/aws-config.git"
 depends: [
   "ocaml" {>= "4.08"}
   "dune" {>= "3.0"}
-  "re"
-  "yojson"
+  "re" {>= "1.7.2"}
+  "yojson" {>= "1.6.0"}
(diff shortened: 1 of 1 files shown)


packages/aws-config/aws-config.0.0.1/opam
```diff
@@ -0,0 +1,25 @@
...
+maintainer: ["Satoru Kawahara <s1311350@gmail.com>"]
+authors: ["Satoru Kawahara <s1311350@gmail.com>"]
+license: "MIT"
+homepage: "https://github.com/Nymphium/aws-config"
+bug-reports: "https://github.com/Nymphium/aws-config/issues"
+dev-repo: "git+https://github.com/Nymphium/aws-config.git"
+depends: [
+  "ocaml" {>= "4.08"}
+  "dune" {>= "3.0"}
+  "re"
```
kit-ty-kate, line 12: Uses `Re.Pcre`
```suggestion
  "re" {>= "1.7.2"}
```

kit-ty-kate:
Thanks. Could you return the fixes to the opam file upstream so that they can be taken into account for next time?

PR #21294, opened 2022-04-28
new file:   packages/parany/parany.12.2.0/opam

UnixJunkie:


UnixJunkie:
There is a new function in the interface: Parany.get_rank(); it allows a worker process to check its rank.
This can be useful when a specific worker need to do something special.
Probably most users will not need this; while advanced users might.

kit-ty-kate:
Thanks

PR #21282, opened 2022-04-25
Re-enable ocaml_intrinsics on macos

gretay-js:
I'd like to re-enable `ocaml_intrinsics`  after the fix in https://github.com/janestreet/ocaml_intrinsics/pull/5 is merged.

PR #21296, opened 2022-04-28
jst-config-v0.15.1

aalekseyev:
Signed-off-by: Arseniy Alekseyev <aalekseyev@janestreet.com>


aalekseyev:
This should fix a core_unix compilation. It is only going to be properly tested when rebased on top of 
#21282. (and I should remember to remove the constraint that PR is adding)

kit-ty-kate:
Thanks

PR #21303, opened 2022-04-29
Why3 1.5.0

silene:


silene:
All the "lower-bounds" failures are independent of Why3 and thus irrelevant.

The "revdeps" failures, however, are genuine and I have added a commit to mark Lambdapi as incompatible with Why3 1.5.

silene:
The only remaining failures are now completely unrelated. That is, Dune 1.11.0, which Why3 does not use, fails to build.

PR #21312, opened 2022-05-02
Cryptokit version 1.17

xavierleroy:
Release 1.17:
- Add interfaces for authenticated encryption (AEAD) and two implementations: AES-GCM and Chacha20-Poly1305.
- Use `getentropy()` for `system_rng` when available (Linux, macOS, BSD).
- Removed support for EGD (the Entropy Gathering Daemon).
- Added compile-time alerts on uses of broken or weak ciphers and hashes. (Can be silenced with "-alert -crypto".)
- Add the hmac_sha384 MAC.
- Add the SipHash MAC.
- Set file descriptor to close-on-exec in `device_rng`.
- Improve compatibility with OCaml 5.0.
- Make sure CryptokitBignum is installed like before the switch to Dune.

### DIFF
--- a/packages/cryptokit/cryptokit.1.17/opam
+++ b/packages/cryptokit/cryptokit.1.17/opam
@@ -2,13 +2,14 @@ opam-version: "2.0"
 synopsis: "A library of cryptographic primitives"
 description:
   "Cryptokit includes authenticated encryption (AES-GCM, Chacha20-Poly1305), block ciphers (AES, DES, 3DES), stream ciphers (Chacha20, ARCfour), public-key cryptography (RSA, DH), hashes (SHA-256, SHA-512, SHA-3, Blake2), MACs, compression, random number generation -- all presented with a compositional, extensible interface."
+license="GNU Library General Public License v2 or later with OCaml LGPL Linking Exception"
 maintainer: ["Xavier Leroy <xavier.leroy@college-de-france.fr>"]
 authors: ["Xavier Leroy"]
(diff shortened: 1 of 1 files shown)


packages/cryptokit/cryptokit.1.17/opam
```diff
@@ -0,0 +1,39 @@
+opam-version: "2.0"
+synopsis: "A library of cryptographic primitives"
+description:
+  "Cryptokit includes authenticated encryption (AES-GCM, Chacha20-Poly1305), block ciphers (AES, DES, 3DES), stream ciphers (Chacha20, ARCfour), public-key cryptography (RSA, DH), hashes (SHA-256, SHA-512, SHA-3, Blake2), MACs, compression, random number generation -- all presented with a compositional, extensible interface."
+license: "GNU Library General Public License v2 or later with OCaml LGPL Linking Exception"
```
mseri, line 5:
```suggestion
license: "LGPL-2.0-or-later WITH OCaml-LGPL-linking-exception"
```

xavierleroy: Makes sense, thanks.  Before writing the license field I searched the Web for documentation on the license field in OPAM files and didn't find anything.

mseri: You are perfectly right, thanks for pointing it out. I have opened an issue on the opam repository to improve the license field description and include an example with the `WITH` keyword.

mseri:
Thanks

PR #21318, opened 2022-05-02
Fix ppx_deriving.4.5 with dune >= 3

balsoft:
Patch ppx_deriving 4.5 to correctly specify the dep, to fix it on dune 3.

- Fix ppx_deriving.4.5 on dune >= 3
- Revert "ppx_deriving.4.5 is not compatible with dune 3.0"

### DIFF
--- a/packages/ppx_deriving/ppx_deriving.4.5-1/files/fix_dune_3.patch
+++ b/packages/ppx_deriving/ppx_deriving.4.5-1/files/fix_dune_3.patch
@@ -1,0 +1,19 @@
+diff --git a/src_plugins/map/dune b/src_plugins/map/dune
+--- a/src_plugins/map/dune
++++ b/src_plugins/map/dune
+@@ -1,5 +1,5 @@
+ (rule
+- (deps ppx_deriving_map.cppo.ml)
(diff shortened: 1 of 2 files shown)


mseri:
Thanks. Can you move these changes to a new parched version `ppx_deriving.4.5+1`?

balsoft:
Done

mseri:
Not sure how to fix the lower bounds issues but those were there already

mseri:
Thanks

balsoft:
> Not sure how to fix the lower bounds issues but those were there already

I suppose we could bump up the lower bounds on cppo and such, not sure if it really makes sense to do here though.

mseri:
Oh I am really really sorry. I just realised I wrote `+1` Instead of `-1` as the new version. Would it be possible to update it? Again apologies 🙏

mseri:
> I suppose we could bump up the lower bounds on cppo and such, not sure if it really makes sense to do here though.

Don’t worry about that. I will look into it later on and send a separate PR

balsoft:
> Oh I am really really sorry. I just realised I wrote +1 Instead of -1 as the new version. Would it be possible to update it

Sure, fixed

mseri:
Thanks a lot!

PR #21327, opened 2022-05-04
Package herdtools7.7.56.1

maranget:
Hi, package created by hand. I hop nothing is wrong.
Thanks.

### DIFF
--- a/packages/herdtools7/herdtools7.7.56.1/opam
+++ b/packages/herdtools7/herdtools7.7.56.1/opam
@@ -1,0 +1,31 @@
+opam-version: "2.0"
+name: "herdtools7"
+version: "7.56.1"
+synopsis: "The herdtools suite for simulating and studying weak memory models"
+description: ""
+maintainer: "Luc Maranget <Luc.Maranget@inria.fr>"
+authors: [
+  "Luc Maranget <Luc.Maranget@inria.fr>"
+  "Jade Alglave <j.alglave@ucl.ac.uk>"
+]
+homepage: "http://diy.inria.fr/"
+bug-reports: "http://github.com/herd/herdtools7/issues/"
+doc: "http://diy.inria.fr/doc/index.html"
+dev-repo: "git+https://github.com/herd/herdtools7.git"
+build: ["sh" "./dune-build.sh" "%{prefix}%"]
+install: ["sh" "./dune-install.sh" "%{prefix}%"]
+# @todo Add "build-doc" field
+# @todo Add "build-test" field
+depends: [
+  "ocaml" {>= "4.05.0"}
+  "dune"  {>= "1.4" }
+  "ocamlfind" { build }
+  "menhir" {>= "20180530"}
+]
+url {
+  src: "https://github.com/herd/herdtools7/archive/7.56.1.tar.gz"
+  checksum: [
+    "md5=e56d543c3a7b312003f56c143ea9d202"
+    "sha512=c1f4786d3161a0f374b27c7576a33c50cc09123c3b54d1b192e420499333606076faba3a631e97d6d0186dbf411100cc0379b8de12ddcf8182444b67965e661a"
+  ]
+}


packages/herdtools7/herdtools7.7.56.1/opam
```diff
@@ -0,0 +1,31 @@
+opam-version: "2.0"
+name: "herdtools7"
```
mseri, line 2:
```suggestio
```
This is inferred from the opam file path

packages/herdtools7/herdtools7.7.56.1/opam
```diff
@@ -0,0 +1,31 @@
...
+version: "7.56.1"
```
mseri, line 3:
```suggestion
```
This as well

packages/herdtools7/herdtools7.7.56.1/opam
```diff
@@ -0,0 +1,31 @@
...
+synopsis: "The herdtools suite for simulating and studying weak memory models"
+description: ""
```
mseri, line 5:
```suggestion
```
No need for this if it is empty

packages/herdtools7/herdtools7.7.56.1/opam
```diff
@@ -0,0 +1,31 @@
...
+authors: [
+  "Luc Maranget <Luc.Maranget@inria.fr>"
+  "Jade Alglave <j.alglave@ucl.ac.uk>"
+]
+homepage: "http://diy.inria.fr/"
+bug-reports: "http://github.com/herd/herdtools7/issues/"
+doc: "http://diy.inria.fr/doc/index.html"
+dev-repo: "git+https://github.com/herd/herdtools7.git"
+build: ["sh" "./dune-build.sh" "%{prefix}%"]
+install: ["sh" "./dune-install.sh" "%{prefix}%"]
```
mseri, line 16: Let’s see how it goes, but I expect this to fail. The sandbox does not allow to modify external folders. Also there is no library/executable specified by `public_name` for installation. I would recommend to use that and let dune generate the right `.install` file.

packages/herdtools7/herdtools7.7.56.1/opam
```diff
@@ -0,0 +1,31 @@
...
+maintainer: "Luc Maranget <Luc.Maranget@inria.fr>"
+authors: [
+  "Luc Maranget <Luc.Maranget@inria.fr>"
+  "Jade Alglave <j.alglave@ucl.ac.uk>"
+]
+homepage: "http://diy.inria.fr/"
+bug-reports: "http://github.com/herd/herdtools7/issues/"
+doc: "http://diy.inria.fr/doc/index.html"
+dev-repo: "git+https://github.com/herd/herdtools7.git"
+build: ["sh" "./dune-build.sh" "%{prefix}%"]
```
mseri, line 15: Dune has a subst feature to have automatic development and release versioning of the packages. It would greatly simplify the build and installation

mseri: Comsider also not using `wrapped false` In the huild, it is very name conflict prone

packages/herdtools7/herdtools7.7.56.1/opam
```diff
@@ -0,0 +1,31 @@
...
+install: ["sh" "./dune-install.sh" "%{prefix}%"]
+# @todo Add "build-doc" field
+# @todo Add "build-test" field
+depends: [
+  "ocaml" {>= "4.05.0"}
+  "dune"  {>= "1.4" }
+  "ocamlfind" { build }
```
mseri, line 22: Is this actually used? From a brief search it did not seem the case

mseri: The installation seems fine. No library is installed, just executables, so there is no need to address the last two points if this is expected

packages/herdtools7/herdtools7.7.56.1/opam
mseri: The installation seems fine. No library is installed, just executables, so there is no need to address the build and install points if this is what is expected

maranget:
> Looks good, thanks. Just a couple of lines to remove and a lower bound to adjust

Thanks, I am working on suppressing warnings and rising lower bound, I'll submit again soon.

packages/herdtools7/herdtools7.7.56.1/opam
maranget: Ok, thanks, won't fix.

packages/herdtools7/herdtools7.7.56.1/opam
maranget: If you mean is `ocamlfind` used ? The answer is that `ocamlfind` is used to build a component that is not compiled by default. As a consequence, teh dependency on `ocamlfind` can probably be removed. I'll try

packages/herdtools7/herdtools7.7.56.1/opam
```diff
@@ -0,0 +1,31 @@
...
+  "menhir" {>= "20180530"}
```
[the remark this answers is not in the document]
maranget: Will do, thanks.

mseri:
Thanks!

PR #21332, opened 2022-05-06
Package herdtools7, version 7.56.2

maranget:
Option `-version` was not working properly in previous 7.56.1. Version 7.56.2 backports a change that fixes the problem.


maranget:
Thanks.

PR #21333, opened 2022-05-06
Do not depend on `ocamlfind` in `base-bytes`

Leonidas-from-XIV:
The current status is probably because `ocamlfind` before 1.5.3 did not know about bytes, but this leads to `ocamlfind` always being installed when `base-bytes` is installed - even when `ocamlfind` is not used, e.g. in setups where everything is built using dune and vendored dependencies like with opam-monorepo.

This PR changes the dependency into a conflict. Another option could be to just conflict all `ocamlfind` versions prior to 1.5.3 with `base-bytes`.

### DIFF
--- a/packages/ANSITerminal/ANSITerminal.0.8/opam
+++ b/packages/ANSITerminal/ANSITerminal.0.8/opam
@@ -15,7 +15,7 @@ build: [
 ]
 depends: [
   "ocaml"
-  "jbuilder" {>= "1.0+beta7"}
+  "dune" {>= "1.4" & < "2.0"}
   "base-bytes"
   "base-unix"
 ]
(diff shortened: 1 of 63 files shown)


dra27:
Sorry for the delay coming to this - I'm just slowly paging this, and the related PRs for OCaml 5 shipping its own `META` files.

I haven't yet done a thorough review, but there are definitely some issues with what's being done here, even at a glance.

I'll aim to review this thoroughly this week, but I think this actually springs from a misunderstanding from the flurry of jbuilder ports back in 2018. No package built with Jbuilder or Dune should ever include `bytes` in a libraries stanza, and no package built using them should therefore depend on base-bytes because they can only target OCaml 4.02+.

From an opam-monorepo perspective, I think that's the better fix - remove the unnecessary dependencies from packages? It would be sensible once https://github.com/ocaml/dune/pull/5526 is merged to look at doing similar for the bytes package (i.e. have dune start to emit a warning about its use).

mseri:
We have discussed about this. Unfortunately this is a quite complicated situation and we cannot merge this as a fix

PR #21345, opened 2022-05-09
CSV-Lwt needs at least Lwt 2.4.7

Leonidas-from-XIV:
While working on #21333 I ran across an issue where `Lwt_io.read_into` needs to be `bytes` not `string`. This is the case starting from Lwt 2.4.7.


mseri:
Thanks!

PR #21346, opened 2022-05-10
Require Lwt 2.4.7 for `Lwt.Infix`

Leonidas-from-XIV:
The error message when using an older version is:

```
File "src/tube.ml", line 12, characters 7-16:
Error: Unbound module Lwt.Infix
```


mseri:
Thanks

PR #21350, opened 2022-05-10
Require older mdx version that installs the `mdx` binary

Leonidas-from-XIV:
The error I ran across in #21333 is the following:
```
File "dune", line 6, characters 15-18:
6 |           (run mdx test %{deps})
                   ^^^
Error: Program mdx not found in the tree or in PATH
 (context: default)
```

It has nothing to do with the original PR so this is a separate fix.


mseri:
Thanks

PR #21353, opened 2022-05-10
Do not run tests of `docker-api.0.1`

Leonidas-from-XIV:
`docker-api.0.1` seems to be failing to build. Upon checking it is due to the tests not being included in the tarball. This PR removes the calls to build and run tests.


mseri:
Thanks

PR #21354, opened 2022-05-10
Do not allow co-installation with ounit1 with different version

Leonidas-from-XIV:
While building
https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/aa898f09e7ffe95265651b4b2dda4d21d91baa19/variant/opam-2.1,compilers,4.14,ounit.2.1.2,revdeps,bencode.2.0 it turns out that ounit 2.1.2 was installed via one dependency and ounit2 2.2.6 via another. This fails to link since the symbols exist twice.

So while ounit1 should depend on ounit2 after the transition, ounit2 should fail to install if version of ounit1 does not match its own version (that is, these versions are known to be compatible).

### DIFF
--- a/packages/ounit2/ounit2.2.2.6/opam
+++ b/packages/ounit2/ounit2.2.2.6/opam
@@ -14,6 +14,9 @@ depends: [
   "seq"
   "stdlib-shims"
 ]
+conflicts: [
+  "ounit" {!= version}
+]
 build: [
   ["dune" "build" "-p" name "-j" jobs]
   ["dune" "runtest" "-p" name "-j" jobs] {with-test}
--- a/packages/ounit2/ounit2.2.2.0/opam
+++ b/packages/ounit2/ounit2.2.2.0/opam
@@ -12,6 +12,9 @@ depends: [
   "base-unix"
   "stdlib-shims"
 ]
+conflicts: [
+  "ounit" {!= version}
+]
 build: [
   ["dune" "build" "-p" name "-j" jobs]
   ["dune" "runtest" "-p" name "-j" jobs] {with-test & ocaml:version < "4.11"}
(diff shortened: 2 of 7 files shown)


Leonidas-from-XIV:
Another case of failure is this: https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/50fa2a30a43b41fbac2717478936578a4bc45b07/variant/opam-2.1,compilers,4.14,ounit.2.1.2,revdeps,stringext.1.4.3

Where qtest depends on ounit2 and stringext on ounit1 and when linked together the compiler refuses to build because the interfaces don't match.

mseri:
I am going to discuss this at the next opam-repo maintainers meeting. Thanks for the PR

packages/ounit2/ounit2.2.2.6/opam
```diff
@@ -14,6 +14,9 @@ depends: [
   "seq"
   "stdlib-shims"
 ]
+conflicts: [
+  "ounit" {!= version}
```
mseri, line 18: This would not work. In your second example `ounit 2.1.2` is installed allongside ounit2 2.2.6.
So the `!=version` would not prevent this to happen

mseri: Ah no, sorry, it is in the conflicts, so in principle would be fine

mseri:
I don't know if this is the right thing or if it would be better to add a convlict in all ounit < 2.2 with ounit2...  I think that would avoid having to bring this conflict to all new versions of ounit2

mseri:
We need to do the opposite. Add the constraint to `ounit1` instead so that we don't need to update future releases of `ounit2`

mseri:
Superseded by https://github.com/ocaml/opam-repository/pull/22255

mseri:
Thanks again for this PR and sorry for the slow response

PR #21358, opened 2022-05-11
Add conf package to require LAPACKE for `owl`

Leonidas-from-XIV:
As I was going through my failing revdeps in #21333, one of them was `owl`, which was failing because LAPACKE did not exist. It's always a bit of a hassle with C libraries on the CI so this PR attempts to pull the dependency out into `conf-lapacke`.

LAPACKE does not seem to be commonly packaged, but Debian has its separate package, so does FreeBSD. Alpine seems to lump it into the regular LAPACK package.

### DIFF
--- a/packages/conf-lapacke/conf-lapacke.1/opam
+++ b/packages/conf-lapacke/conf-lapacke.1/opam
@@ -1,0 +1,16 @@
+opam-version: "2.0"
+maintainer: "Marek Kubica <marek@xivilization.net>"
+authors: "Marek Kubica <marek@xivilization.net>"
+homepage: "http://www.netlib.org/lapack"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: "LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception"
+depexts: [
+  ["liblapacke-dev"] {os-family = "debian"}
+  ["lapack-dev"] {os-distribution = "alpine"}
+  ["lapacke" "gcc"] {os = "freebsd"}
+]
+synopsis: "Virtual package for LAPACKE configuration"
+description: """
+Virtual package relying on a LAPACKE (Linear Algebra) library installation.
+This package requires a LAPACKE installation on the system."""
+flags: conf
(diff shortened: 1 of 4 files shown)


mseri:
This is a bit tricky, there is a conflict between openblas (which normally provides also lapacke and cblas and is used by owl) and blas/lapacke which can be provided by many packages. It used to work on debian because they were providing a good pkg-config package, but this was modified to follow ubuntu recently (which instead is somewhat broken, there are a number of open issues due to the bad packaging if you use lapacke)

mseri:
It does seem to improve the situation for older owl though. I am now not sure why it fails for owl 1.0.2.

Leonidas-from-XIV:
Ah, good to know. Guess we'll have to wait until Debian fixes this and can ignore the failures in the meantime.

Actually, owl 1.0.2 is the only one that creates issues on my other PRs so this PR might not be fixing anything, because owl 1.0.2 still seems to continue to fail. I would've expected all 1.0.x packages to work the same but possibly something changed between 1.0.1 and 1.0.2 that makes it not pick up any LAPACKE implementation when linking.

mseri:
There does not seem to be any difference in the build and link instructions looking at the diff 😕

mseri:
The build has not changed, they use the same exact dune and configurator scripts.
I am updating my M1 PR to take into account the new debian lapacke package though, I think your PR is a strict improvement with recent debians

mseri:
In any case, looks like owl 1.0.2 was the only one failing to link but not the other ones... I am really puzzled: https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/cd341914e20b9a61888bd50ddb1b68c6957c9ddc

Leonidas-from-XIV:
I saw the same here: https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/aa898f09e7ffe95265651b4b2dda4d21d91baa19 (the other failing owl packages also picked 1.0.2 and failed mostly due to that), but I assumed that this was only because that is the most current version and no other versions were attempted to build.

mseri:
Lapacke is installed already as part of conf-openblas: https://github.com/ocaml/opam-repository/blob/master/packages/conf-openblas/conf-openblas.0.2.1/opam

I suggest to close this and add
```
 ["lapack-dev"] {os-distribution = "alpine"}
  ["lapacke" "gcc"] {os = "freebsd"}
```
with also the openblas package to that conf- package? Can I ask you to do that?

PR #21360, opened 2022-05-11
Add LAPACKE-containing packages of Alpine and FreeBSD

Leonidas-from-XIV:
As discussed in #21358 this adds the packages that contain LAPACKE in Alpine and FreeBSD. Debian already had `liblapacke-dev`, so no change required there.


mseri:
Thanks a lot

mseri:
Thanks a lot

PR #21365, opened 2022-05-13
Fix minimal version of `irc-client-lwt`

Leonidas-from-XIV:
`irc-client-lwt` fails on building `irc-client` 0.4.0 which fails because building the latter fails on an OASIS error.

This PR attempts to fix the build by using a newer version of OASIS, but it is unclear whether that will fix the build or the constraint will have to be raised.

### DIFF
--- a/packages/irc-client/irc-client.0.4.0/opam
+++ b/packages/irc-client/irc-client.0.4.0/opam
@@ -29,7 +29,7 @@ remove: [
 ]
 depends: [
   "ocaml" {>= "4.00.1"}
-  "oasis" {build & >= "0.4.11"}
+  "oasis" {build & >= "0.4.2"}
   "ocamlbuild" {build}
   "ocamlfind" {build}
   "result"
--- a/packages/irc-client/irc-client.0.5.0/opam
+++ b/packages/irc-client/irc-client.0.5.0/opam
@@ -29,7 +29,7 @@ remove: [
 ]
 depends: [
   "ocaml" {>= "4.00.1"}
-  "oasis" {build & >= "0.4.11"}
+  "oasis" {build & >= "0.4.2"}
   "ocamlbuild" {build}
   "ocamlfind" {build}
   "result"
--- a/packages/irc-client/irc-client.0.5.1/opam
+++ b/packages/irc-client/irc-client.0.5.1/opam
@@ -29,7 +29,7 @@ remove: [
 ]
 depends: [
   "ocaml" {>= "4.00.1"}
-  "oasis" {build & >= "0.4.11"}
+  "oasis" {build & >= "0.4.2"}
   "ocamlbuild" {build}
   "ocamlfind" {build}
   "result"
--- a/packages/irc-client/irc-client.0.5.2/opam
+++ b/packages/irc-client/irc-client.0.5.2/opam
@@ -29,7 +29,7 @@ remove: [
 ]
 depends: [
   "ocaml" {>= "4.00.1"}
-  "oasis" {build & >= "0.4.11"}
+  "oasis" {build & >= "0.4.2"}
   "ocamlbuild" {build}
   "ocamlfind" {build}
   "result"
--- a/packages/irc-client/irc-client.0.5.4/opam
+++ b/packages/irc-client/irc-client.0.5.4/opam
@@ -29,7 +29,7 @@ remove: [
 ]
 depends: [
   "ocaml" {>= "4.00.1"}
-  "oasis" {build & >= "0.4.11"}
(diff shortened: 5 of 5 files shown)


Leonidas-from-XIV:
Example of previously failing build: https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/50fa2a30a43b41fbac2717478936578a4bc45b07/variant/opam-2.1,compilers,4.02,irc-client-lwt.0.6.0,lower-bounds

Looks like this:
```
Exception: PropList.Unknown_field ("XOCamlbuildExtraArgs", "Document").
File "setup.ml", line 1:
Error: Reference to undefined global `OASISDynRun'
```

OASIS 0.4.0 should include `DynRun` but somehow it doesn't get picked up. It seems to work with OASIS 0.4.11 though (newest version at the time of writing) so something was fixed in the meantime - neither the changelog nor a cursory exploration of the OASIS repo provides a simple way to determine the exact release since when this was solved, but 0.4.11 is the same series so that constraint should be acceptable anyway.

Leonidas-from-XIV:
No. It takes a lot of time waiting on the CI and I could not find anything in the changelog that would point at this being fixed in some release, nor did the source code provide any obvious hints, so instead of spending a day bisecting different versions to find the absolute minimum version, I picked `0.4.11` given that all packages that depend on `oasis.0.4.0` should be okay with `0.4.11`.

I find that to be an acceptable time vs correctness compromise, given that probably other packages that have minimum constraints written by the maintainers might similarly overestimate the required minimal version (e.g. packages that require dune often build with older dune just fine)

mseri:
Sure, that is a reasonable approach. I'll take over from this and do the bisecting locally before merging.

packages/irc-client/irc-client.0.4.0/opam
```diff
@@ -28,7 +29,7 @@ remove: [
 ]
 depends: [
   "ocaml" {>= "4.00.1"}
-  "oasis" {build & >= "0.4.0"}
+  "oasis" {build & >= "0.4.11"}
```
mseri, line 32:
```suggestion
  "oasis" {build & >= "0.4.2"}
```

packages/irc-client/irc-client.0.5.0/opam
```diff
@@ -28,7 +29,7 @@ remove: [
 ]
 depends: [
   "ocaml" {>= "4.00.1"}
-  "oasis" {build & >= "0.4.0"}
+  "oasis" {build & >= "0.4.11"}
```
mseri, line 32:
```suggestion
  "oasis" {build & >= "0.4.2"}
```

packages/irc-client/irc-client.0.5.2/opam
```diff
@@ -28,7 +29,7 @@ remove: [
 ]
 depends: [
   "ocaml" {>= "4.00.1"}
-  "oasis" {build & >= "0.4.0"}
+  "oasis" {build & >= "0.4.11"}
```
mseri, line 32:
```suggestion
  "oasis" {build & >= "0.4.2"}
```

packages/irc-client/irc-client.0.5.4/opam
```diff
@@ -28,7 +29,7 @@ remove: [
 ]
 depends: [
   "ocaml" {>= "4.00.1"}
-  "oasis" {build & >= "0.4.0"}
+  "oasis" {build & >= "0.4.11"}
```
mseri, line 32:
```suggestion
  "oasis" {build & >= "0.4.2"}
```

mseri (commented):
The first release of oasis fixing
```
# Exception: PropList.Unknown_field ("XOCamlbuildExtraArgs", "Document").
# File "setup.ml", line 1:
# Error: Reference to undefined global `OASISDynRun'
```
is 0.4.2. I tested this locally

packages/irc-client/irc-client.0.5.1/opam
```diff
@@ -28,7 +29,7 @@ remove: [
 ]
 depends: [
   "ocaml" {>= "4.00.1"}
-  "oasis" {build & >= "0.4.0"}
+  "oasis" {build & >= "0.4.11"}
```
mseri, line 32:
```suggestion
  "oasis" {build & >= "0.4.2"}
```
Missed this

Leonidas-from-XIV:
Looks good. Maybe it would make sense to collect such minimum version failures to working version mappings somewhere:

 * Objects from C-stubs not found: blacklist `ocamlbuild.0.9.0`
 * OASISDynRun not found: depend on `oasis {>= 0.4.2}
 * etc.`

PR #21366, opened 2022-05-13
Fix minimum version of SSL dependency for Lwt_ssl

Leonidas-from-XIV:
It currently fails to compile with `invalid use of incomplete typedef 'X509_STORE_CTX'` and `ssl.0.5.3` seems to solve this differently, so upgrading the dependency might solve the issue at hand.

### DIFF
--- a/packages/lwt_ssl/lwt_ssl.1.1.1/opam
+++ b/packages/lwt_ssl/lwt_ssl.1.1.1/opam
@@ -23,7 +23,7 @@ depends: [
   "base-unix"
   "jbuilder" {>= "1.0+beta10"}
   "lwt" {>= "3.0.0"}
-  "ssl" {>= "0.5.0"}
+  "ssl" {>= "0.5.3"}
 ]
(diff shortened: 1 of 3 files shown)


mseri:
If this is due to a too recent version of libssl in the various oses, I would leave it as is

Leonidas-from-XIV:
I am not sure it ever worked, but it is hard to say. It seems like `ssl.0.5.3` fixed the issue by having special accessor functions so I assume it works now both on newer and older SSL but one would need to ask the `ssl` authors.

Maybe @toots or @smimram  can explain the change and how compatible it is.

toots:
I have talked with @smimram and we're pretty overwhelmed. Do y'all think that there would be community interest in taking over maintenance for `ocaml-ssl`?

anmonteiro:
I’d be happy to step up as OCaml-ssl maintainer. I’ve contributed a few PRs and am fairly familiar with the codebase.

mseri:
I am going to close this for now. I think the error is due to the C library, in which case the library may still install in older systems without this updated bound

PR #21367, opened 2022-05-13
Fix lower-version `dokeysto` packaging issues

Leonidas-from-XIV:
With extunix.0.1.1 if fails with dozens of messages like

```
gcc: error: src/eventfd.o: No such file or directory
```

### DIFF
--- a/packages/conf-liblz4/conf-liblz4.1/opam
+++ b/packages/conf-liblz4/conf-liblz4.1/opam
@@ -11,6 +11,7 @@ depexts: [
   ["lz4-devel"] {os-distribution = "centos"}
   ["lz4-devel"] {os-distribution = "rhel"}
   ["lz4-devel"] {os-distribution = "fedora"}
+  ["lz4-devel"] {os-distribution = "oraclelinux-8"}
   ["lz4-dev"] {os-distribution = "alpine"}
   ["liblz4-devel"] {os-family = "suse"}
   ["lz4"] {os = "macos" & os-distribution = "homebrew"}
--- a/packages/conf-liblz4/conf-liblz4.1/opam
+++ b/packages/conf-liblz4/conf-liblz4.1/opam
@@ -11,7 +11,7 @@ depexts: [
   ["lz4-devel"] {os-distribution = "centos"}
   ["lz4-devel"] {os-distribution = "rhel"}
   ["lz4-devel"] {os-distribution = "fedora"}
-  ["lz4-devel"] {os-distribution = "oraclelinux-8"}
+  ["lz4-devel"] {os-distribution = "ol"}
   ["lz4-dev"] {os-distribution = "alpine"}
   ["liblz4-devel"] {os-family = "suse"}
   ["lz4"] {os = "macos" & os-distribution = "homebrew"}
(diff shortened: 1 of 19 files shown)


Leonidas-from-XIV:
Hmm, no, this looks more like an issue with the build system that extunix uses.

packages/conf-liblz4/conf-liblz4.1/opam
```diff
@@ -11,6 +11,7 @@ depexts: [
   ["lz4-devel"] {os-distribution = "centos"}
   ["lz4-devel"] {os-distribution = "rhel"}
   ["lz4-devel"] {os-distribution = "fedora"}
+  ["lz4-devel"] {os-distribution = "oraclelinux-8"}
```
mseri, line 14: Does this work? I think the distribution is "ol"

mseri:
Thanks. Can you please document what error is related to the various lower bound added in this or?

packages/conf-liblz4/conf-liblz4.1/opam
Leonidas-from-XIV: It didn't. I tried `oraclelinux` and `oraclelinux-8` before. Thanks for the pointer for `ol`, I'll give that one a try.

Sidenote: how do I learn about the possible names? Especially if I am not running that distribution. I can often get package names from online search interfaces but the OS names/distribution names are difficult to figure out.

mseri: Kate added them here: https://github.com/ocaml/opam-repository/wiki/Depexts-os-distribution---os-family-values
I used to search among the conf-package in the opam-repository before :D

packages/dokeysto_camltc/dokeysto_camltc.4.0.0/opam
```diff
@@ -12,8 +12,8 @@ build: [
 depends: [
   "ocaml"
   "dune" {>= "1.11"}
-  "dokeysto"
-  "camltc"
+  "dokeysto" {>= "4.0.0"}
```
Leonidas-from-XIV, line 15:
```
7 | module RO: functor (KV: Dokeysto.Db_gen.Key_val) -> sig
                            ^^^^^^^^^^^^^^^^^^^^^^^
Error: Unbound module Dokeysto.Db_gen
```

packages/dokeysto_lz4/dokeysto_lz4.4.0.0/opam
```diff
@@ -12,7 +12,7 @@ build: [
 depends: [
   "ocaml"
   "dune" {>= "1.11"}
-  "dokeysto"
+  "dokeysto" {>= "4.0.0"}
```
Leonidas-from-XIV, line 15:
```
7 | module RO: functor (KV: Dokeysto.Db_gen.Key_val) -> sig
                            ^^^^^^^^^^^^^^^^^^^^^^^
Error: Unbound module Dokeysto.Db_gen
```

Leonidas-from-XIV (commented):
I'll have to hunt down the other errors from the CI but I force-pushed so it is hard to find the older builds.

mseri:
Thanks

PR #21368, opened 2022-05-13
Use newer `ocamlbuild` to build `cstruct` versions that use it

Leonidas-from-XIV:
`cstruct.0.7.0` FTBFS because the stubs can't be found by `ocamlbuild.0.9.0`. This PR raises the minimum version to one where the stubs are found.

### DIFF
--- a/packages/angstrom/angstrom.0.1.0/opam
+++ b/packages/angstrom/angstrom.0.1.0/opam
@@ -25,7 +25,7 @@ depends: [
   "alcotest" {with-test & >= "0.4.1" & < "0.8.0"}
   "cstruct" {>= "1.0.0" & < "3.2.0"}
   "ocamlfind" {build}
-  "ocamlbuild" {build}
+  "ocamlbuild" {build & >= "0.9.1"}
   "result"
   "ocplib-endian" {>= "0.6"}
 ]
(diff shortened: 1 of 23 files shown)


Leonidas-from-XIV:
The lower-bound builds fail with

```
ocamlfind ocamlc -g -ccopt -O3 -c -o lib/cstruct_stubs.o lib/cstruct_stubs.c
/home/opam/.opam/4.03/bin/ocamlmklib -o lib/cstruct_stubs lib/cstruct_stubs.o
+ /home/opam/.opam/4.03/bin/ocamlmklib -o lib/cstruct_stubs lib/cstruct_stubs.o
gcc: error: lib/cstruct_stubs.o: No such file or directory
gcc: fatal error: no input files
compilation terminated.
Command exited with code 2.
```

when picking `ocamlbuild.0.9.0`. Turns out builds succeed when using `ocamlbuild.0.9.1`, so this PR adds the constraints to all versions of `cstruct` that use `ocamlbuild`

mseri:
I think This was a known bug of 0.9.0. It should work if you use `!= "0.9.0"`. Are you seeing failures with older versions?

Leonidas-from-XIV:
Ah, that's a good idea. I assumed it was all versions but depending on `>= 0.9.1` disallows all versions that ship with a built-in `ocamlbuild`. I'll give it a try and see if that works too.

Leonidas-from-XIV:
Yes, this seems to have worked, I've adjusted all the `opam` files to avoid `ocamlbuild.0.9.0`.

mseri:
Thanks!

PR #21369, opened 2022-05-13
Package ometrics.0.2.1

vch9:
### `ometrics.0.2.1`
OCaml analysis in a merge request changes

---
* Homepage: https://gitlab.com/nomadic-labs/ometrics
* Source repo: git+https://gitlab.com/nomadic-labs/ometrics.git
* Bug tracker: https://gitlab.com/nomadic-labs/ometrics/-/issues

---

### DIFF
--- a/packages/ometrics/ometrics.0.2.1/opam
+++ b/packages/ometrics/ometrics.0.2.1/opam
@@ -1,0 +1,32 @@
+opam-version: "2.0"
+license: "MIT"
+maintainer: "Valentin Chaboche <valentin.chaboche@lambda-coins.com>"
+homepage: "https://gitlab.com/nomadic-labs/ometrics"
+dev-repo: "git+https://gitlab.com/nomadic-labs/ometrics.git"
+bug-reports: "https://gitlab.com/nomadic-labs/ometrics/-/issues"
+synopsis: "OCaml analysis in a merge request changes"
+
+depends: [
+  "ocaml" {>= "4.12.0"}
+  "dune" {>= "2.9.1"}
+  "ppxlib" {>= "0.24.0" & < "0.26.0"}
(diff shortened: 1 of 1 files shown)


vch9:
In the last version `0.2.0` I required a `ppxlib.0.26.0`. The constraint is too hard, it is not compatible with some projects and I do not use the features introduced in `>= 0.26.0`. Unfortunately, I can not only change the constraint since there are needed changes in the AST between 0.25 and 0.26.

mseri:
OpamSolution.Fetch_fail("https://github.com/vch9/ometrics/archive/0.2.1.tar.gz (curl: code 404 while downloading https://github.com/vch9/ometrics/archive/0.2.1.tar.gz)")

vch9:
> OpamSolution.Fetch_fail("https://github.com/vch9/ometrics/archive/0.2.1.tar.gz (curl: code 404 while downloading https://github.com/vch9/ometrics/archive/0.2.1.tar.gz)")

Sorry about that, I drafted the release but did not publish it.
However I could restart one specific job (not the whole pipeline), and it seems that I need to restrict ppxlib even more, it apparently build only with ppxlib.0.25.0

I recall how [coqffi](https://github.com/coq-community/coqffi) handles different versions of OCaml: it uses a shell script to apply a git patch for a given ocaml version. Would you think that's a good idea here for ppxlib? With this setup I would require ppxlib to be at least a given version and for every version since then I'd write a patch.

I'm afraid that this will become tedious to maintain but  `ppxlib {= 0.25.0}` seems really complicated with

vch9:
For now I've restricted the version, I'll see if git patches are necessary in a next release

mseri:
Looks good, thanks

PR #21371, opened 2022-05-13
Package opsian.0.1

RichardWarburton:
### `opsian.0.1`
Low overhead profiling

---
* Homepage: https://github.com/Opsian/opsian-ocaml/wiki
* Source repo: git://git@github.com/Opsian/opsian-ocaml.git
* Bug tracker: https://github.com/Opsian/opsian-ocaml/issues

---

### DIFF
--- a/packages/opsian/opsian.0.1/opam
+++ b/packages/opsian/opsian.0.1/opam
@@ -13,6 +13,7 @@ depends: [
   "conf-automake"
   "conf-perl"
   "conf-liblzma"
+  "conf-libtool"
 ]
 available:
   arch != "arm32" & arch != "x86_32" & arch != "s390x" & os-family != "arch" &
--- a/packages/opsian/opsian.0.1/opam
+++ b/packages/opsian/opsian.0.1/opam
@@ -2,7 +2,7 @@ opam-version: "2.0"
 synopsis: "Low overhead profiling"
 maintainer: "Opsian <admin@opsian.com>"
 authors:
-  "Richard Warburton <richard@opsian.com>, Sadiq Jaffer <sadiq@opsian.com>"
+  ["Richard Warburton <richard@opsian.com>" "Sadiq Jaffer <sadiq@opsian.com>"]
 license: "MIT"
 homepage: "https://github.com/Opsian/opsian-ocaml/wiki"
 bug-reports: "https://github.com/Opsian/opsian-ocaml/issues"
@@ -10,43 +10,15 @@ depends: [
   "dune" {>= "2.8"}
   ("ocaml" {>= "4.12.0" & < "5.0.0"} |
    ("ocaml" {>= "5.0.0"} & "runtime_events_tools"))
+  "conf-automake"
+  "conf-perl"
+  "conf-liblzma"
 ]
 available:
   arch != "arm32" & arch != "x86_32" & arch != "s390x" & os-family != "arch" &
-  os-family != "alpine"
+  os-family != "alpine" & os != "macos"
 build: ["dune" "build" "--release" "-j" jobs]
-depexts: [
-  ["libc-dev"] {os-family = "debian" | os-family = "ubuntu"}
-  ["automake"] {os-family = "debian" | os-family = "ubuntu"}
-  ["libtool"] {os-family = "debian" | os-family = "ubuntu"}
-  ["liblzma-dev"] {os-family = "debian" | os-family = "ubuntu"}
-  ["perl"] {os-family = "debian" | os-family = "ubuntu"}
-  ["libc-dev"] {os-family = "alpine"}
-  ["autoconf"] {os-family = "alpine"}
-  ["automake"] {os-family = "alpine"}
-  ["libtool"] {os-family = "alpine"}
-  ["xz-dev"] {os-family = "alpine"}
-  ["automake"] {os-family = "arch"}
-  ["libtool"] {os-family = "arch"}
-  ["xz"] {os-family = "arch"}
-  ["glibc-devel"]
-    {os-family = "fedora" | os-distribution = "rhel" | os-family = "centos"}
-  ["automake"]
-    {os-family = "fedora" | os-distribution = "rhel" | os-family = "centos"}
-  ["libtool"]
-    {os-family = "fedora" | os-distribution = "rhel" | os-family = "centos"}
-  ["xz-devel"]
-    {os-family = "fedora" | os-distribution = "rhel" | os-family = "centos"}
-  ["perl-core"]
-    {os-family = "fedora" | os-distribution = "rhel" | os-family = "centos"}
-  ["glibc-devel"] {os-family = "suse" | os-family = "opensuse"}
-  ["automake"] {os-family = "suse" | os-family = "opensuse"}
-  ["libtool"] {os-family = "suse" | os-family = "opensuse"}
-  ["xz-devel"] {os-family = "suse" | os-family = "opensuse"}
-  ["perl-core"]
-    {os-family = "fedora" | os-distribution = "rhel" | os-family = "centos"}
-]
-dev-repo: "git://git@github.com/Opsian/opsian-ocaml.git"
+dev-repo: "git+https://github.com/Opsian/opsian-ocaml.git"
 url {
   src:
     "https://github.com/Opsian/opsian-ocaml/releases/download/0.1/0.1.tar.gz"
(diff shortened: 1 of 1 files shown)


packages/opsian/opsian.0.1/opam
```diff
@@ -0,0 +1,35 @@
...
+# for building native deps
+depexts: [
+  [ "libpthread-stubs0-dev" ] {os-distribution = "ubuntu" }
+  [ "libc6-dev" ] {os-distribution = "ubuntu" } # -ldl -lrt
+  [ "autoconf" ] {os-distribution = "ubuntu" }
+  [ "libtool" ] {os-distribution = "ubuntu" } # libunwind
+  [ "liblzma-dev" ] {os-distribution = "ubuntu" }
+]
+build: [
+  [ "git" "submodule" "update" "--init" "--recursive" ]
```
avsm, line 23: You shouldn't need a submodule init here, since this package is using a release archive.

packages/opsian/opsian.0.1/opam
```diff
@@ -0,0 +1,35 @@
...
+authors:
+  "Richard Warburton <richard@opsian.com>, Sadiq Jaffer <sadiq@opsian.com>"
+license: "MIT"
+homepage: "https://github.com/Opsian/opsian-ocaml/wiki"
+bug-reports: "https://github.com/Opsian/opsian-ocaml/issues"
+depends: [
+  "dune" {>= "2.8"}
+  "ocaml" {>= "4.12.0"}
+]
+flags: [ verbose ]
```
avsm, line 13: Do you need to force verbose builds for this package? In general, the opam-repo prefers to let the user set this when installing.

RichardWarburton:
> Nothing is exposed by this package. Neither of the library or the examples exposes a `public_name`d artifact, so nothing would be installed if I am not mistaken

Thanks for pointing that out - that got accidentally removed last week.

packages/opsian/opsian.0.1/opam
RichardWarburton: Will be fixed in latest push.

RichardWarburton:
All reported issues resolved and CI passing. Can I request a merge?

packages/opsian/opsian.0.1/opam
```diff
@@ -0,0 +1,57 @@
...
+homepage: "https://github.com/Opsian/opsian-ocaml/wiki"
+bug-reports: "https://github.com/Opsian/opsian-ocaml/issues"
+depends: [
+  "dune" {>= "2.8"}
+  ("ocaml" {>= "4.12.0" & < "5.0.0"} |
+   ("ocaml" {>= "5.0.0"} & "runtime_events_tools"))
+]
+available:
+  arch != "arm32" & arch != "x86_32" & arch != "s390x" & os-family != "arch" &
+  os-family != "alpine"
```
kit-ty-kate, line 16:
```suggestion
  os-family != "alpine" & os != "macos"
```

packages/opsian/opsian.0.1/opam
```diff
@@ -0,0 +1,57 @@
...
+  ["automake"] {os-family = "suse" | os-family = "opensuse"}
+  ["libtool"] {os-family = "suse" | os-family = "opensuse"}
+  ["xz-devel"] {os-family = "suse" | os-family = "opensuse"}
+  ["perl-core"]
+    {os-family = "fedora" | os-distribution = "rhel" | os-family = "centos"}
+]
+dev-repo: "git://git@github.com/Opsian/opsian-ocaml.git"
+url {
+  src:
+    "https://github.com/Opsian/opsian-ocaml/releases/download/0.1/0.1.tar.gz"
```
kit-ty-kate, line 52: Note for next time that this archive contains the whole `.git` directory which shouldn’t be required for building and makes the archive bigger than it needs to

packages/opsian/opsian.0.1/opam
```diff
@@ -0,0 +1,57 @@
...
+maintainer: "Opsian <admin@opsian.com>"
+authors:
+  "Richard Warburton <richard@opsian.com>, Sadiq Jaffer <sadiq@opsian.com>"
+license: "MIT"
+homepage: "https://github.com/Opsian/opsian-ocaml/wiki"
+bug-reports: "https://github.com/Opsian/opsian-ocaml/issues"
+depends: [
+  "dune" {>= "2.8"}
+  ("ocaml" {>= "4.12.0" & < "5.0.0"} |
+   ("ocaml" {>= "5.0.0"} & "runtime_events_tools"))
```
kit-ty-kate, line 12:
```suggestion
   ("ocaml" {>= "5.0.0"} & "runtime_events_tools"))
  "conf-automake"
  "conf-perl"
  "conf-liblzma"
```

packages/opsian/opsian.0.1/opam
```diff
@@ -0,0 +1,57 @@
...
+    {os-family = "fedora" | os-distribution = "rhel" | os-family = "centos"}
+  ["perl-core"]
+    {os-family = "fedora" | os-distribution = "rhel" | os-family = "centos"}
+  ["glibc-devel"] {os-family = "suse" | os-family = "opensuse"}
+  ["automake"] {os-family = "suse" | os-family = "opensuse"}
+  ["libtool"] {os-family = "suse" | os-family = "opensuse"}
+  ["xz-devel"] {os-family = "suse" | os-family = "opensuse"}
+  ["perl-core"]
+    {os-family = "fedora" | os-distribution = "rhel" | os-family = "centos"}
+]
```
kit-ty-kate, line 48: Replaced by the suggested dependencies above. This makes everything much more maintainable.
```suggestion
```

packages/opsian/opsian.0.1/opam
```diff
@@ -0,0 +1,57 @@
+opam-version: "2.0"
+synopsis: "Low overhead profiling"
+maintainer: "Opsian <admin@opsian.com>"
+authors:
+  "Richard Warburton <richard@opsian.com>, Sadiq Jaffer <sadiq@opsian.com>"
```
kit-ty-kate, line 5:
```suggestion
  ["Richard Warburton <richard@opsian.com>" "Sadiq Jaffer <sadiq@opsian.com>"]
```

packages/opsian/opsian.0.1/opam
kit-ty-kate: Uses of `sed` isn’t compatible with macOS’s

packages/opsian/opsian.0.1/opam
RichardWarburton: Noted and understood. We've been using git submodules to manage dependencies here. The only other option appears to be committing the sources of those dependencies into the repository. Is that preferred?

packages/opsian/opsian.0.1/opam
```diff
@@ -0,0 +1,29 @@
...
+license: "MIT"
+homepage: "https://github.com/Opsian/opsian-ocaml/wiki"
+bug-reports: "https://github.com/Opsian/opsian-ocaml/issues"
+depends: [
+  "dune" {>= "2.8"}
+  ("ocaml" {>= "4.12.0" & < "5.0.0"} |
+   ("ocaml" {>= "5.0.0"} & "runtime_events_tools"))
+  "conf-automake"
+  "conf-perl"
+  "conf-liblzma"
```
kit-ty-kate, line 15:
```suggestion
  "conf-liblzma"
  "conf-libtool"
```

RichardWarburton:
@kit-ty-kate I think the replacement of the dep-exts with the virtual opam packages has resulted in an error in the vendored openssl compiling. When I tried this before it needed `perl-core` on RHEL/Fedora, whilst the opam `conf-perl` virtual package seems to depend upon: `perl-Pod-Html`

https://github.com/ocaml/opam-repository/blob/master/packages/conf-perl/conf-perl.2/opam

packages/opsian/opsian.0.1/opam
kit-ty-kate: why is that? The archive already contains all the submodules. Unless you somehow use `git` during the compilation process (which shouldn’t be needed), there is no need for the git directory to be present in that particular archive

kit-ty-kate:
That looks like a bug in `conf-perl` to me. We can fix that later, it’s not a blocker for this PR.

Thanks! Could you return the fixes to the opam file upstream so that it can be taken into account for next time?

RichardWarburton:
Thanks @kit-ty-kate ! I've pushed the changes into our repo.

PR #21379, opened 2022-05-16
owl.1.0.2: release owl 1.0.2-1 with patch to fix systems providing liblapacke separately

mseri:
The patch reverts a spurious commits that break all systems providing liblapacke separately. This should fix the build on debian and recent ubuntus, while we wait for the support to M1 and the new build configuration to land


mseri:
Thanks @Leonidas-from-XIV for spotting the issue: https://github.com/ocaml/opam-repository/pull/21358

PR #21381, opened 2022-05-17
Prevent jbuilder from reading config files

emillon:
## The issue

The issue we're trying to fix is that `jbuilder` tries to read dune's configuration file. The symptom is an error message like:

    File "/path/to/.config/dune/config", line 1, characters 0-16:
    Error: S-expression of the form (_ _) expected

(tracked as #15943 and dune#3232)

## A bit of history

When introduced, `jbuilder` did not have any configuration file. Instructions in opam files were not very precise, basically just `jbuild build`.
In `1.0+beta7`, the `-p` option got added, with the same semantics that `dune` has today: something to attach various options to release packages.
In `1.0+beta18`, a config file was added, and with it the option `--no-config` (implied by `-p`). The path to that config file is `~/.config/dune/config`.

So we're seeing opam files that use not precise instructions (no `-p`, no `--no-config` - because these were not supported, or because instructions were not generated by `jbuilder`).

## What can we do about this?

The precise fix is to pass `--no-config` whenever it is supported by `jbuilder`. That means adding `"--no-config" {jbuilder:version >= "1.0+beta18"}` to all affected build instructions. This works before `1.0+beta18` (no config reading so nothing to fix) and after (`jbuilder` might read the config file, but the option will prevent it).

Let's explore the alternatives:
- we could add a lower bound `"jbuilder" {>= "1.0+beta18"}` and add `--no-config`. This would work is upgrading `jbuilder` was always an option (like it is for `dune`) but unfortunately some packages only build with old versions of `jbuilder` (see #10479 for an example). So this would make these uninstallable.
- we could add an upper bound `"jbuilder" {< "1.0+beta18"}` to prevent config reading. But that's problematic because that restricts the build plans a lot (for example it prevents using them with `dune` `1.x` through `jbuilder.transition`).
- we could add `-p` to the affected packages and add a bound `"jbuilder" {>= "1.0+beta7"}` - this will automatically pass `--no-config` after `1.0+beta18`. But this change is a bit more invasive as it will change what is being built.

## Finding affected packages

To find the affected packages, I ran the following program to check for packages that depend on `jbuilder` but whose build instructions do not contain `-p` or `--no-config`.

<details>
<summary>find.ml</summary>

```ocaml
let depends_on package (formula : OpamTypes.filtered_formula) =
  OpamFormula.fold_left
    (fun acc (name, _) -> acc || OpamPackage.Name.compare package name = 0)
    false formula

let commands_to_list commands =
  List.concat_map
    (fun (args, _) ->
[...]

Issue #15943, opened 2020-03-04
Remaining jbuilder-using packages should use --no-config-file


smondet:
Cf. https://github.com/ocaml/dune/issues/3232

Is there a tool to edit all build commands in an opam-repo?

XVilka:
List of those (removed the ones that had earlier versions depending on it):
```
    ago>=0.4
    ascii85>=0.4
    asl>=0.11
    atd>=1.2.1
    atdgen>=1.10.2
    atdgen-runtime
    balancer
    binbin
    bip32
    bisect_ppx-ocamlbuild
    bitmasks>=1.1.0
    brotli>=2.0.3
    build_path_prefix_map
    callipyge>=0.2
    camlimages>=5.0.0
    camlon>=2.0.1
    cdrom>=0.9.3
    cmdtui>=0.4.3
    cmdtui-lambda-term
    configurator
    craml
    crc>=2.0.0
    crowbar
    cryptodbm>=0.84.2
    curly
    depyt>=0.2.0
    dlm
    dnssd
    doc-ock>=1.1.0
    doc-ock-html>=1.1.0
    doc-ock-xml>=1.1.0
    dryunit
    dune_watch
    easy_xlsx
    ezxenstore>=0.1.2
    facile>=1.1.4
    faraday>=0.3.0
    faraday-async
    faraday-lwt
    faraday-lwt-unix
    fd-send-recv>=1.0.5
    febusy
    findlib_top
    frenetic>=5.0.0
    gammu>=0.9.4
    gdal>=0.10.0
    gdbprofiler>=0.2
    gen=0.5.1
    genspio>=0.0.1
    hiredis>=0.8
    hiredis-value
    hvsock>=1.0.0
    incremental_kernel>=v0.9.0
    inuit>=0.4.1
    ipaddr=2.8.0
    ipv6-multicast>=0.9
    ipv6-multicast-lwt
    jane-street-tests
    jenga>=v0.9.0
    json-derivers
    json-wheel_jane_street_overlay
    json_of_jsonm
    kafka>=0.3
    kubecaml
    kyotocabinet
    let-if
    levenshtein>=1.1.3
    llopt
    lwt_camlp4
    mecab
    mesh-display
    milter>=1.0.4
    minimal
    modular-arithmetic
    monomorphic>=1.5
    mstruct>=1.3.3
    mvar
    nonstd>=0.0.3
    nullable-array
    numalib
    nunchaku>=0.5.1
    obi
    oc45
[...]

PR #21383, opened 2022-05-17
jbuilder: -p requires >= 1.0+beta7

emillon:
The -p option was only added in 1.0+beta7, so this adjusts the constraints of all the packages that use it.

### DIFF
--- a/packages/ANSITerminal/ANSITerminal.0.8/opam
+++ b/packages/ANSITerminal/ANSITerminal.0.8/opam
@@ -15,7 +15,7 @@ build: [
 ]
 depends: [
   "ocaml"
-  "jbuilder"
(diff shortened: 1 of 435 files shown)


dra27:
These days, opam-repo-ci should survive being asked to do more revdeps 💪

emillon:
All right, let's see how this goes!

emillon:
Some context on this PR: I've started fixing #15943 in #21381 and saw some failures on the lower bounds runs, so here's a fix. There are probably more related ones so that depends how deep we want to fix jbuilder packages but these two ones seem important to fix.

mseri:
Thanks, looks fine. Fixing all the lower bounds is out of scope for this PR

PR #21396, opened 2022-05-19
Package sedlex.3.0

toots:
### `sedlex.3.0`
An OCaml lexer generator for Unicode
sedlex is a lexer generator for OCaml. It is similar to ocamllex, but supports
Unicode. Unlike ocamllex, sedlex allows lexer specifications within regular
OCaml source files. Lexing specific constructs are provided via a ppx syntax
extension.

---
* Homepage: https://github.com/ocaml-community/sedlex
* Source repo: git+https://github.com/ocaml-community/sedlex.git
* Bug tracker: https://github.com/ocaml-community/sedlex/issues

---


mseri:
Do the failing revdeps require upper bounds on the new sedlex?

mseri:
The upper bound have been added to #21401 as well as the incorrect lower bound in the gen dependency

toots:
Thanks for checking this out!

jberdine:
Is the changed bound on ocaml due to the removal of Genlex and Stream? If so, it would be better to still allow 4.14.0 since they are still there, but won't be in 4.14.1.

Atry:
This PR is a regression. I think it's better to revert it. I created #21405.

PR #21399, opened 2022-05-20
2 packages from ocaml-community/cppo at 1.6.9

liyishuai:
This pull-request concerns:
-`cppo.1.6.9`: Code preprocessor like cpp for OCaml
-`cppo_ocamlbuild.1.6.9`: Plugin to use cppo with ocamlbuild

---
* Homepage: https://github.com/ocaml-community/cppo
* Source repo: git+https://github.com/ocaml-community/cppo.git
* Bug tracker: https://github.com/ocaml-community/cppo/issues

---
:camel: Pull-request generated by opam-publish v2.1.0

PR #21401, opened 2022-05-20
Fix various bounds

mseri:
Seen on #21399 and #21396

### DIFF
--- a/packages/bytearray/bytearray.1.0.1/opam
+++ b/packages/bytearray/bytearray.1.0.1/opam
@@ -9,7 +9,7 @@ build: [
 ]
 depends: [
   "base-bigarray"
-  "cppo"
+  "cppo" {>= "1.1.2"}
   "dune"
   "ocaml" {< "4.14~" }
 ]
--- a/packages/bytearray/bytearray.1.0.1/opam
+++ b/packages/bytearray/bytearray.1.0.1/opam
@@ -11,7 +11,7 @@ depends: [
   "base-bigarray"
   "cppo"
   "dune"
-  "ocaml"
+  "ocaml" {< "4.14~" }
 ]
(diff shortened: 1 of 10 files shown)


mseri:
Failures are described in the commit messages

packages/bytearray/bytearray.1.0.1/opam
```diff
@@ -11,7 +11,7 @@ depends: [
   "base-bigarray"
   "cppo"
```
mseri, line 12:
```suggestion
  "cppo" {>= "1.1.2"}
```
Uses `cppo -V`

packages/bytearray/bytearray.1.0.2/opam
```diff
@@ -12,7 +12,7 @@ depends: [
   "base-bigarray"
   "cppo"
```
mseri, line 13:
```suggestion
  "cppo" {>= "1.1.2"}
```

packages/gufo/gufo.0.1.2/opam
```diff
@@ -26,7 +26,7 @@ depends: [
   "lwt"
   "lambda-term" {>= "2.0.2"}
   "base-unix"
-  "sedlex" {>= "2.0.0"}
+  "sedlex" {>= "2.0.0" & < "3.0.0"}
```
mseri, line 29:
```suggestion
  "sedlex" {>= "2.0.0" & < "3.0"}
```

jberdine:
Aside: Packages in opam-repository should not be allowed to specify `-Werror`. Is there some way to override such things globally in opam? Perhaps opam could set `OCAMLPARAM=_,warn-error=-a`.

mseri:
I agree, but there is no global mechanism afaik. You can change them only creating manual patches and releasing a patched version

PR #21405, opened 2022-05-20
Allow Sedlex 2.x to be used with OCaml 4.14

Atry:
#21396 is a regression because it breaks existing Sedlex 2.x users on OCaml 4.14. This PR fixes the regression by setting the upper bound OCaml version to 5.0.

Note that OCaml 4.14 deprecated the `Stream` API, but didn't remove it, therefore OCaml 4.14 is still compatible with Sedlex 2.x. Setting the upper bound to 4.14.0 was a mistake in #21396 .

### DIFF
--- a/packages/sedlex/sedlex.1.99.4/opam
+++ b/packages/sedlex/sedlex.1.99.4/opam
@@ -7,7 +7,7 @@ build: [ [make "all"]  [make "opt"] ]
 install: [make "install"]
 remove: [["ocamlfind" "remove" "sedlex"]]
 depends: [
-  "ocaml" {>= "4.03.0" & < "4.14.0"}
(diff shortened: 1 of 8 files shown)


toots:
Thanks for the quick fix

Atry:
The CI failure is preexisting in Sedlex.1.99.4 
- https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/b5b56724d6afc623121d4c152f5e50563382faeb
- https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/0c62fb569e3ea034f3279a65513a6424be6791a1

mseri:
Thanks a lot. Yes I saw that in the previous PRs as well but I did not know how to fix them

PR #21424, opened 2022-05-25
Update ocaml-config.3 for Windows readiness and compatibility with latest OCaml trunk

dra27:
Trunk OCaml changed the installation layout of the Unix, Dynlink and Str libraries yesterday evening. As this only affects the test ocaml-variants.5.0.0+trunk, ocaml-config.3 is updated directly, as we we can't constrain it.

While it's being updated, I widened the constraint for `ocaml-config.3` ready for the Windows version of ocaml-base-compiler - Windows requires ocaml-config.3. Technically, ocaml-config.1 and ocaml-config.2 should be updated to constrain away Windows, but this will trigger mass switch rebuilds, so I suggest leaving it be for now.


dra27:
The extras failures in opam-repo-ci don't matter - the rest confirms the skipping on all released versions.

PR #21426, opened 2022-05-25
Package secp256k1.0.4.4

dakk:
### `secp256k1.0.4.4`
Elliptic curve library secp256k1 wrapper for Ocaml
This library wrap the secp256k1 EC(DSA) library into an OCaml library. At the moment only a subset of functionalities are available:
- Context: create, clone, destroy, randomize
- Elliptic curve: public key creation
- ECDSA: verify, sign, recover

All exchanged data (pubkey, signature, seckey) are represented as hex strings.

---
* Homepage: https://github.com/dakk/secp256k1-ml
* Source repo: git+https://github.com/dakk/secp256k1-ml
* Bug tracker: https://github.com/dakk/secp256k1-ml/issues

---

### DIFF
--- a/packages/secp256k1/secp256k1.0.4.4/opam
+++ b/packages/secp256k1/secp256k1.0.4.4/opam
@@ -1,0 +1,40 @@
+opam-version: "2.0"
+synopsis: "Elliptic curve library secp256k1 wrapper for Ocaml"
+description: """\
+This library wrap the secp256k1 EC(DSA) library into an OCaml library. At the moment only a subset of functionalities are available:
+- Context: create, clone, destroy, randomize
+- Elliptic curve: public key creation
+- ECDSA: verify, sign, recover
+
+All exchanged data (pubkey, signature, seckey) are represented as hex strings."""
+maintainer: "Davide Gessa <gessadavide@gmail.com>"
+authors: [
+  "Davide Gessa <gessadavide@gmail.com>"
+  "Vincent Bernardoff <vb@luminar.eu.org>"
+  "Yoichi Hirai <i@yoichihirai.com>"
+  "Anton Trunov"
+]
+license: "MIT"
+homepage: "https://github.com/dakk/secp256k1-ml"
+bug-reports: "https://github.com/dakk/secp256k1-ml/issues"
+depends: [
+  "ocaml"
+  "dune" {>= "1.0"}
+  "base" {build & >= "v0.9.2"}
+  "stdio" {build & >= "v0.9.0"}
+  "dune-configurator" {build & >= "1.0"}
+  "hex" {with-test & >= "1.1.1"}
+  "alcotest" {with-test & >= "0.8.1"}
+  "odoc" {with-test & >= "1.3.0"}
+  "base-bigarray"
+  "conf-secp256k1"
+]
+build: ["dune" "build" "-p" name "-j" jobs]
+dev-repo: "git+https://github.com/dakk/secp256k1-ml"
+url {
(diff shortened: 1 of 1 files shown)


packages/secp256k1/secp256k1.0.4.4/opam
```diff
@@ -0,0 +1,40 @@
...
+  "Vincent Bernardoff <vb@luminar.eu.org>"
+  "Yoichi Hirai <i@yoichihirai.com>"
+  "Anton Trunov"
+]
+license: "MIT"
+homepage: "https://github.com/dakk/secp256k1-ml"
+bug-reports: "https://github.com/dakk/secp256k1-ml/issues"
+depends: [
+  "ocaml"
+  "dune" {>= "1.0"}
```
anton-trunov, line 22:
```suggestion
  "dune" {>= "1.3"}
```

I think with this change and given that #21428 has been merged CI may succeed this time

anton-trunov, line 22:
```suggestion
  "dune" {>= "1.3"}
```

I think with this change and given that #21428 has been merged CI may succeed this time

mseri: Thanks, I was just looking at this

packages/secp256k1/secp256k1.0.4.4/opam
```diff
@@ -0,0 +1,40 @@
...
+  "Vincent Bernardoff <vb@luminar.eu.org>"
+  "Yoichi Hirai <i@yoichihirai.com>"
+  "Anton Trunov"
+]
+license: "MIT"
+homepage: "https://github.com/dakk/secp256k1-ml"
+bug-reports: "https://github.com/dakk/secp256k1-ml/issues"
+depends: [
+  "ocaml"
+  "dune" {>= "1.3"}
```
mseri, line 22:
```suggestion
  "dune" {>= "1.0"}
```

packages/secp256k1/secp256k1.0.4.4/opam
```diff
@@ -0,0 +1,40 @@
...
+  "base" {build & >= "v0.9.2"}
```
mseri, line 23: I think requiring `base >= "v1.12.0"` should also fix the issue, without requiring a more restrictive bound on dune

mseri: In any case I sent your way a PR to simplify the build dependencies

mseri: Sorry, there was a typo, I meant "v0.12.0". In any case, that drops ocaml 4.02, I think your original suggestion was better

mseri:
I think https://github.com/dakk/secp256k1-ml/pull/21/ is a strictly better approach from this point of view

packages/secp256k1/secp256k1.0.4.4/opam
mseri: I think https://github.com/dakk/secp256k1-ml/pull/21/ is a strictly better approach from this point of view

dakk:
> I think [dakk/secp256k1-ml#21](https://github.com/dakk/secp256k1-ml/pull/21) is a strictly better approach from this point of view

thanks, btw I can check and try that not before the next week

dakk:
> If you are fine with the current solution, I am happy to merge this in the meantime

sure, it's ok for me

anton-trunov:
Thanks a bunch, @dakk and @mseri!

PR #21430, opened 2022-05-26
conf-secp256k1: update oraclelinux

mseri:
The correct id is ol


mseri:
Let’s leave it as is, so that it will start working once oraclelinux adds the package

PR #21437, opened 2022-05-30
Add a new release of mlmpfr.4.1.0 renaming the main module

thvnx:
This patch remanes the main mlmpfr module from Mpfr to Mlmpfr to avoid
name clashes with the mlgmpidl's mpfr module.

Warning: it breaks the backward compatibility.

See also https://github.com/thvnx/mlmpfr/issues/21.


mseri:
The revdeps seem to be all working

thvnx:
Yes, the breaking change is in the release, I fixed the `conflicts` bound for the only one dependency (why3) to not use the latest version. (They are aware of this change).

PR #21447, opened 2022-05-31
pgx_*.2.1: fix use of dune subst

mseri:
Ping @gtrak 

This is a bug in dune, it is fixed in `dune language` >= 2.7. It is fine to fix it when releasing just in the opam repository


gtrak:
thanks, I'll update our things.

PR #21457, opened 2022-06-02
Package algaeff.0.2.0

favonia:
### `algaeff.0.2.0`
Reusable Effects-Based Components
This OCaml library collects reusable effects-based components we have identified while developing our proof assistants based on algebraic effects.

---
* Homepage: https://github.com/RedPRL/algaeff
* Source repo: git+https://github.com/RedPRL/algaeff.git
* Bug tracker: https://github.com/RedPRL/algaeff/issues

---

### DIFF
--- a/packages/algaeff/algaeff.0.2.0/opam
+++ b/packages/algaeff/algaeff.0.2.0/opam
@@ -1,0 +1,30 @@
+opam-version: "2.0"
+synopsis: "Reusable Effects-Based Components"
+description: """
+This OCaml library collects reusable effects-based components we have identified while developing our proof assistants based on algebraic effects.
+"""
+maintainer: "favonia <favonia@gmail.com>"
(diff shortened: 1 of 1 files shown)


mseri:
Does this require a patched version of ocamlfind from the alpha repository to run the tests?

favonia:
> Does this require a patched version of ocamlfind from the alpha repository to run the tests?

Ah, yes. I have been using the alpha repository and did not notice that (https://github.com/ocaml/ocamlfind/pull/41). What should I do now? To be clear, this package does not directly require ocamlfind and should compile without changes once the toolchain (compiler/ocamlfind/...) is fixed.

mseri:
I think we should just merge it, once the new ocamlfind is release it should also come with upper bounds for the previous versions. I think this was not yet done to avoid triggering massive rebuilds.

PR #21471, opened 2022-06-05
[new release] textrazor (0.1.1)

Richard-Degenne:
An OCaml wrapper for the TextRazor API

- Project page: <a href="https://github.com/Richard-Degenne/ocaml-textrazor">https://github.com/Richard-Degenne/ocaml-textrazor</a>

### Fixed
- Added support for Yojson 2


Richard-Degenne:
I see errors for 4.04 through 4.07 about a request timing out, but I'm not sure what they mean.

mseri:
It means that opam was nit able to find a combination of compatible packages in 60 seconds for those compilers. It is fine to merge as is for me

mseri:
Not necessarily. Just means that opam is unable to find a compatible set of dependencies in a short time. It probably does not exist but not necessarily

PR #21460, opened 2022-06-02
[new release] yojson and yojson-bench (2.0.0)

panglesd:
Yojson is an optimized parsing and printing library for the JSON format

- Project page: <a href="https://github.com/ocaml-community/yojson">https://github.com/ocaml-community/yojson</a>
- Documentation: <a href="https://ocaml-community.github.io/yojson/">https://ocaml-community.github.io/yojson/</a>

### Removed

- Removed dependency on easy-format and removed `pretty_format` from
  `Yojson`, `Yojson.Basic`, `Yojson.Safe` and `Yojson.Raw`. (@c-cube, ocaml-community/yojson#90)
- Removed dependency on `biniou`, simplifying the chain of dependencies. This
  changes some APIs:
  * `Bi_outbuf.t` in signatures is replaced with `Buffer.t`
  * `to_outbuf` becomes `to_buffer` and `stream_to_outbuf` becomes
    `stream_to_buffer`
  (@Leonidas-from-XIV, ocaml-community/yojson#74, and @gasche, ocaml-community/yojson#132)
- Removed `yojson-biniou` library
- Removed deprecated `json` type aliasing type `t` which has been available
  since 1.6.0 (@Leonidas-from-XIV, ocaml-community/yojson#100).
- Removed `json_max` type (@Leonidas-from-XIV, ocaml-community/yojson#103)
- Removed constraint that the "root" value being rendered (via either
  `pretty_print` or `to_string`) must be an object or array. (@cemerick, ocaml-community/yojson#121)
- Removed `validate_json` as it only made sense if the type was called `json`.
  (@Leonidas-from-XIV, ocaml-community/yojson#137)

### Add

- Add an opam package `yojson-bench` to deal with benchmarks dependency
  (@tmcgilchrist, ocaml-community/yojson#117)
- Add a benchmark to judge the respective performance of providing a buffer vs
  letting Yojson create an internal (ocaml-community/yojson#134, @Leonidas-from-XIV)
- Add an optional `suf` keyword argument was added to functions that write
  serialized JSON, thus allowing NDJSON output. Most functions default to not
  adding any suffix except for `to_file` (ocaml-community/yojson#124, @panglesd) and functions
  writing sequences of values where the default is `\n` (ocaml-community/yojson#135,
  @Leonidas-from-XIV)

### Change

- The `stream_from_*` and `stream_to_*` functions now use a `Seq.t` instead of a
[...]

PR #21475, opened 2022-06-06
Yojson 2 constraints

Leonidas-from-XIV:
This PR is (collecting) the Yojson 2.0.0 maximum constraints that are being revealed by the revdeps of #21460. I'll try to open PRs to the upstream projects which are affected so future releases will support newer Yojson versions again.

### DIFF
--- a/packages/0install/0install.2.15.1/opam
+++ b/packages/0install/0install.2.15.1/opam
@@ -10,7 +10,7 @@ build: [
 ]
 depends: [
   "ocaml" {>= "4.05.0"}
-  "yojson"
+  "yojson" {< "2.0.0"}
   "xmlm"
   "ounit" {with-test}
   "lwt"
--- a/packages/0install/0install.2.15.2/opam
+++ b/packages/0install/0install.2.15.2/opam
@@ -10,7 +10,7 @@ build: [
 ]
 depends: [
   "ocaml" {>= "4.05.0"}
-  "yojson"
+  "yojson" {< "2.0.0"}
   "xmlm"
   "ounit" {with-test}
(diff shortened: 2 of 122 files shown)


Leonidas-from-XIV:
@c-cube I know it is archived but it looks like the tarball of `cconv.0.2` is messed up. Wonder whether to replace it with the one from the OPAM cache?

mseri:
I think this may be full of lower-bound issues. I suggest not to bother with them now unless there is something trivial to fix. I will try to solve them separately in subsequent PRs

c-cube:
@Leonidas-from-XIV I think the opam-repo should be updated to use https://github.com/c-cube/cconv/archive/refs/tags/0.2.tar.gz if it's not in sync.

Leonidas-from-XIV:
@mseri Ok, I fixed the whole rat-tail of minimum version failures, but there's some leftover failures like `obus.1.1.8` which is failing on… unclear what. Maybe ocamlfind? Maybe oasis? Do you have an idea?

mseri:
Thanks a lot. I will have a look at this PR tomorrow and most likely merge it. After that we can try the yojson PR again

Leonidas-from-XIV:
Great, let me know when to rebase and squash all the fixup commits. I let it like this for now so the links to the CI would work.

mseri (approved):
Looks great. Thanks a lot

Leonidas-from-XIV:
Ok, I rebased and squashed the constraint-adding commits into one as well as the respective `fixup` commits, to the list of commits should be reasonably clean. No further changes.

PR #21486, opened 2022-06-07
Release bin_prot v0.14.1

tov:

PR #21491, opened 2022-06-08
Fix some “bad checksum” errors

mseri:
Seen on https://github.com/ocaml/opam-repository/pull/21486


mseri:
The linter seems to be unhappy with the use of `msat` in
```
install: [
    [make "DOCDIR=%{msat:doc}%" "install"]
]
remove: [
    [make "DOCDIR=%{msat:doc}%" "uninstall"]
]
```
but this has been there forever. Should it be fixed?

PR #21493, opened 2022-06-09
Restrict maximum ocamlformat version for OCanren

Leonidas-from-XIV:
The error message looks like this:

```
ocamlformat: option '--profile': value `compact` has been removed in version
             0.22.
```

So this makes sure that OCanren will pick a version that has `compact`.


mseri:
Thanks

PR #21494, opened 2022-06-09
Svmwrap nonlopt 400 [do not merge]

UnixJunkie:


UnixJunkie:
I am not sure this is the right way to do; but in case one of the dependencies cannot install; I might still want to be able to install the software; so there is a git branch where we can compile even without the said dependency (ocaml-nlopt).

packages/svmwrap_nonlopt/svmwrap_nonlopt.4.0.0/opam
```diff
@@ -0,0 +1,74 @@
+opam-version: "2.0"
+authors: "Francois Berenger"
+maintainer: "unixjunkie@sdf.org"
+homepage: "https://github.com/UnixJunkie/svmwrap"
+bug-reports: "https://github.com/UnixJunkie/svmwrap/issues"
+dev-repo: "git+https://github.com/UnixJunkie/svmwrap.git"
+license: "BSD-3-Clause"
+build: ["dune" "build" "-p" name "-j" jobs]
```
mseri, line 8:
```suggestion
build: ["dune" "build" "-p" "svmwrap" "-j" jobs]
```
This should fix the build, but it would be saner to make nlwrap a depopt or to separate from `svmwrap` a `svmwrap-nlopt` subpackage

UnixJunkie: maybe I should just rename the opam file in the sources

UnixJunkie: good idea

kit-ty-kate:
@UnixJunkie what's left to do on this PR?

UnixJunkie:
I have a no_nlopt branch; let's just use this one and install from sources if users need it.

PR #21495, opened 2022-06-09
OCaml 5.0: create 5.1.0+trunk package and update 5.0.0+trunk

Octachron:
This PR creates a new `ocaml-variants.5.1.0+trunk` and `ocaml.5.1.0` packages and update the `ocaml-variant.5.0.0+trunk` package to track the new OCaml 5.0 branch.

### DIFF
--- a/packages/ocaml/ocaml.5.1.0/opam
+++ b/packages/ocaml/ocaml.5.1.0/opam
@@ -1,0 +1,31 @@
+opam-version: "2.0"
+license: "LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception"
+synopsis: "The OCaml compiler (virtual package)"
+description: """
+This package requires a matching implementation of OCaml,
+and polls it to initialise specific variables like `ocaml:native-dynlink`"""
+maintainer: "platform@lists.ocaml.org"
+depends: [
+  "ocaml-config" {>= "2"}
+  "ocaml-base-compiler" {= "5.1.0"} |
+  "ocaml-variants" {>= "5.1.0" & < "5.1.1~"} |
+  "ocaml-system" {>= "5.1.0" & < "5.1.1~"}
+]
+setenv: [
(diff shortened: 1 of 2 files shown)


packages/ocaml-variants/ocaml-variants.5.0.0+trunk/opam
```diff
@@ -52,7 +52,7 @@ build: [
 ]
 install: [make "install"]
 url {
-  src: "https://github.com/ocaml/ocaml/archive/trunk.tar.gz"
+  src: "https://github.com/ocaml/ocaml/archive/5.0.tar.gz"
```
mseri, line 55: wow, the archive is actually called 5.0? No alpha/beta/...?

Octachron: It is the archive for the 5.0 branch (the version itself is 5.0.0+dev2-2022-06-08).

mseri:
The installation works in the CI cases where it can happen. We can't test it better than that right now

packages/ocaml/ocaml.5.1.0/opam
```diff
@@ -0,0 +1,31 @@
+opam-version: "2.0"
+license: "LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception"
+synopsis: "The OCaml compiler (virtual package)"
+description: """
+This package requires a matching implementation of OCaml,
+and polls it to initialise specific variables like `ocaml:native-dynlink`"""
+maintainer: "platform@lists.ocaml.org"
+depends: [
+  "ocaml-config" {>= "2"}
```
dra27, line 9:
```suggestion
  "ocaml-config" {>= "3"}
```

and the ocaml.5.0.0 package needs this update as well

Octachron: Updated.

dra27 (approved):
Good to go!

PR #21499, opened 2022-06-09
Release ppx_regexp 0.5.1 (bugfix).

paurkedal:
There was an unfortunate bug in the latest version released a few days ago (which prevents preprocessing files which do not actually use the extension). [Change log](https://github.com/paurkedal/ppx_regexp/releases/tag/v0.5.1):

- Fix invalid AST due to empty binding list in `ppx_regexp`.


mseri:
If you want/think is important to do it, you can mark the previous version with `available: false` to make sure it is not selected by chance

paurkedal:
Yes, that's a good idea. I guess we'll do that in a new PR.

PR #21502, opened 2022-06-10
Remove unnecessary bound on cmdliner

NathanReb:
In a recent change to index, the maintainers made it compatible with both cmdliner.1.1.0 and older versions. This upper bound was a leftover and is not necessary anymore.


NathanReb:
We pushed those changes upstream too: https://github.com/mirage/index/pull/391

PR #21509, opened 2022-06-11
Update OCaml 5.x version number format

dra27:
More important PR for 5.0-readiness to follow, but this gets the tedious bit out of the way!


dra27:
No upstream changes are needed for this PR - they're all either old releases or upstream doesn't require editing.

kit-ty-kate:
Thanks!

PR #21510, opened 2022-06-11
Fix architecture support for OCaml 5.x

dra27:
OCaml 5.x is only available in native code for amd64 and arm64; all other configurations are bytecode-only.

- `ocaml-option-32bit` therefore pulls in `ocaml-option-bytecode-only` for 5.x
- `ocaml-option-afl`, `ocaml-option-fp` and `ocaml-option-flambda` are native compiler options, so they should always have conflicted `ocaml-option-bytecode-only`
- Frame pointers has only ever been available on linux amd64 which is now reflected in `ocaml-option-fp`
- Frame pointer support hasn't yet been restored to OCaml 5.x (the PR is under review), so for now the 5.x packages conflict `ocaml-option-fp`
- The 5.x packages depend on `ocaml-option-bytecode-only` if the architecture is not amd64 or arm64
- It turns out that `opam-depext` doesn't install correctly in a bytecode-only switch (ironically, as it _always_ build the bytecode version of the plugin, even in a full compiler switch). This is addressed in https://github.com/ocaml-opam/opam-depext/pull/149, but this PR also updates the build instructions for 1.2.1 (as a 1.2.1-1 package). There will be a 1.2.2 release with the upstream fix, at which point we can constrain the older packages away from 5.x, but this fix is critical for the CI base images.

### DIFF
--- a/packages/ocaml-option-32bit/ocaml-option-32bit.1/opam
+++ b/packages/ocaml-option-32bit/ocaml-option-32bit.1/opam
@@ -4,7 +4,7 @@ depexts: [
   ["gcc-multilib" "g++-multilib"] {os-family = "debian"}
 ]
 depends: [
-  "ocaml-variants" {post & >= "4.12.0~"}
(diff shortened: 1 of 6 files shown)


avsm (approved):
This all looks good to me.

dra27:
Thanks!

PR #21511, opened 2022-06-11
[new release] knights_tour (0.0.3)

kdvolder:
Solves the Knight's tour puzzle; and others

- Project page: <a href="https://github.com/kdvolder/knights_tour">https://github.com/kdvolder/knights_tour</a>

##### CHANGES:

Add some randomization to the Hexo puzzle solver to make it more
interesting. It still does a 'complete' search however by
randomizing the order in which puzzle pieces and variants are
tried means it will find different solutions on each run.
Randomizations also more often than not find a solution much 
faster than the default ordering.

Remove the 'withUndo' operator from Searchspace because it
is hard to support properly and hard to use properly without
introducing bugs in a searchspace definition (hard to contain
side-effects).

Add a 'breadth_search' searchspace exploration algorithm that
gradually regresses back to dept-first search to limit memory
usage.

Add a human readable textual de/serialization format
for specifying polyomino puzzles. 

Make graphical drawing of polyomino board more pleasing to the eye:
- each piece has a different color.
- borders around the pieces to make them stand out more.

Add hexo_solve executable which solves an instance of Hexominos.

Add random_hexo_puzzle executable which generates and saves a
randomized hexo_puzzle to a text file.

Add a `solve_file` executable that reads a polyomino puzzle 
specification from a text file and starts solving it.

### DIFF
--- a/packages/knights_tour/knights_tour.0.0.3/opam
+++ b/packages/knights_tour/knights_tour.0.0.3/opam
@@ -1,0 +1,42 @@
+opam-version: "2.0"
+synopsis: "Solves the Knight's tour puzzle; and others"
+description: "See https://en.wikipedia.org/wiki/Knight%27s_tour"
+maintainer: ["kris.de.volder@gmail.com"]
+authors: ["Kris De Volder"]
+license: "MIT"
+homepage: "https://github.com/kdvolder/knights_tour"
+bug-reports: "https://github.com/kdvolder/knights_tour/issues"
+depends: [
+  "dune" {>= "3.1"}
+  "ocaml" {>= "4.14.0"}
+  "graphics"
+  "stdio"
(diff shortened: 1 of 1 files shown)


mseri:
Thanks

kdvolder:
There were some failures in the CI checks. Message like this: 
```
2022-06-12 01:29.09: Job failed: Disconnected: Switch turned off
```

@mseri The fact that you merged the PR regardless probably means this isn't a real issue.  But I'm curious what this error really means. And if there's something I can/should do to avoid it in the future.

mseri:
Yes those are CI problems unrelated to this PR. In this case something happened to the cluster probably due to very high load, and some jobs where interrupted/workers disconnected. It is more and more rare to see them but from time to time they still happen

PR #21514, opened 2022-06-12
Maximum constraint on Yojson for Merlin 4.5-413

ode79:
As discussed in the comments at https://github.com/ocaml/opam-repository/commit/63cbf66956bbd16de6ff7ad951d4f06aa64758d4

PR #21520, opened 2022-06-13
Apply Yojson 2.0 constraint on all Merlin versions

Leonidas-from-XIV:
They probably all use the removed `Stream` API so unless inspecting them all individually, a general Yojson 2.0 constraint is probably the best course of action.

Related to #21514.


mseri:
Sounds reasonable, I was planning on doing the same later this evening. Thanks a lot

PR #21521, opened 2022-06-13
oseq 0.4.0: add upper bound on ocaml 4.14

mseri:
The api is no longer compatible


kit-ty-kate:
What was the build failure for this? It built fine here with both 4.14 and 5.0: http://check.ocamllabs.io/log/1654728944-506b5ecc1e701d138834708697cf5deded61cf4e/4.14/good/oseq.0.4

mseri:
Maybe it is just the tests then? Does it work when you use it in another package?

PR #21523, opened 2022-06-13
Remove ppx_deriving_yojson constraint on Yojson 2.0

Leonidas-from-XIV:
The tests seem to run with Yojson 2.0 just fine, so to allow people to use Yojson when they use ppx_deriving removing this constraint might be sensible.

Opening as a draft first to see whether this works with the latest release and might be removing the constraints from more releases if it works.

### DIFF
--- a/packages/ppx_deriving_yojson/ppx_deriving_yojson.3.5.1/opam
+++ b/packages/ppx_deriving_yojson/ppx_deriving_yojson.3.5.1/opam
@@ -13,7 +13,7 @@ build: [
 ]
 depends: [
   "ocaml" {>= "4.04.0" & < "4.10.0"}
-  "yojson" {>= "1.6.0" & < "2.0.0"}
+  "yojson" {>= "1.6.0"}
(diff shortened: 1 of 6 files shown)


mseri:
Looks like it worked, we just need to send a PR with a few upper bounds

Leonidas-from-XIV:
Looking at the revdeps, the failures with `Bi_outbuf.t` make me a bit sad because the packages depend on Biniou transitively, without declaring a dependency on it and with Yojson 2 dropping that dependency makes it completely disappear from their dependency cone and thus fail the build.

mseri:
I suugest to add `conflicts: [ "yojson" {>= "2.0.0"} ]` in that case

mseri:
Should be fine now. Can you rebase on master so that we double check?

mseri:
Looks like we only need to fix key-parsers. I will add archetype tarball in a separate PR over the next few days

mseri:
Thanks

PR #21524, opened 2022-06-13
Restore ocaml-option-bytecode-only as a depopt for 5.x packages

dra27:
I made a small mistake in https://github.com/ocaml/opam-repository/pull/21510 - `ocaml-option-bytecode-only` should have remained as a depopt for the two 5.x trunk packages (otherwise if you have a normal switch and then `opam install ocaml-option-bytecode-only` it doesn't rebuild the switch).


kit-ty-kate:
Sounds good. Thanks!

kit-ty-kate:
This PR is in a bugged state on Github (it’s merged in master as https://github.com/ocaml/opam-repository/commit/13c823a5c47d2cf50bb3ef4c7d5429069ec803f8 but was not automatically marked as merged and closed)

Waiting does not seem to help so closing manually

PR #21525, opened 2022-06-14
[new release] base32 (1.0.0)

pukkamustard:
>  Base32 is a binary-to-text encoding that represents binary data in an ASCII string format by translating it into a radix-32 representation.  It is specified in RFC 4648.  This library provides an OCaml implementation of Base32.


packages/base32/base32.1.0.0/opam
```diff
@@ -0,0 +1,41 @@
+# This file is generated by dune, edit dune-project instead
+opam-version: "2.0"
+version: "1.0.0"
```
kit-ty-kate, line 3:
```suggestion
```

packages/base32/base32.1.0.0/opam
```diff
@@ -0,0 +1,40 @@
+# This file is generated by dune, edit dune-project instead
+opam-version: "2.0"
+synopsis: "Base32 encoding for OCaml"
+description:
+  "Base32 is a binary-to-text encoding that represents binary data in an ASCII string format by translating it into a radix-32 representation.  It is specified in RFC 4648.  This library provides an OCaml implementation of Base32."
```
kit-ty-kate, line 5: It might be more readable by doing something like that:
```suggestion
description: """
Base32 is a binary-to-text encoding that represents binary data in an ASCII string format by translating it into a radix-32 representation.  It is specified in RFC 4648.
This library provides an OCaml implementation of Base32.
"""
```

kit-ty-kate:
Thanks

pukkamustard:
> Thanks

Thank you!

PR #21528, opened 2022-06-14
release ocaml-solo5 0.8.1

hannesm:
Keep track of memory allocation and releasing (fast than mallinfo, more accurate than footprint) https://github.com/mirage/ocaml-solo5/pull/120 @winux138 @palainp


kit-ty-kate:
Thanks!

PR #21530, opened 2022-06-14
[new release] ppx_import (1.10.0)

ejgallego:
A syntax extension for importing declarations from interface files

- Project page: <a href="https://github.com/ocaml-ppx/ppx_import">https://github.com/ocaml-ppx/ppx_import</a>
- Documentation: <a href="https://ocaml-ppx.github.io/ppx_import/">https://ocaml-ppx.github.io/ppx_import/</a>

##### CHANGES:

  * Update ppxlib to 0.26.0 (ocaml-ppx/ppx_import#69, @pitag-ha)

### DIFF
--- a/packages/ppx_import/ppx_import.1.10.0/opam
+++ b/packages/ppx_import/ppx_import.1.10.0/opam
@@ -1,0 +1,42 @@
+opam-version: "2.0"
+synopsis: "A syntax extension for importing declarations from interface files"
+maintainer: "whitequark <whitequark@whitequark.org>"
+authors: [ "whitequark <whitequark@whitequark.org>" ]
+homepage: "https://github.com/ocaml-ppx/ppx_import"
+doc: "https://ocaml-ppx.github.io/ppx_import/"
+license: "MIT"
+bug-reports: "https://github.com/ocaml-ppx/ppx_import/issues"
+dev-repo: "git+https://github.com/ocaml-ppx/ppx_import.git"
+tags: [ "syntax" ]
+
+depends: [
+  (
+  "ocaml"                   {>= "4.05.0" &  < "4.10.0"  }
+  "dune"                    {              >= "1.11.0"  }
+  "ppxlib"                  {              >= "0.26.0"  }
+  "ounit"                   { with-test                 }
+  "ppx_deriving"            { with-test  & >= "4.2.1"   }
+  )
+  |
+  (
+  "ocaml"                   { >= "4.10.0"               }
+  "ppx_sexp_conv"           { with-test  & >= "v0.13.0" }
+  "dune"                    {              >= "1.11.0"  }
+  "ppxlib"                  {              >= "0.26.0"  }
+  "ounit"                   { with-test                 }
+  "ppx_deriving"            { with-test  & >= "4.2.1"   }
+  )
+]
+
+build:      [["dune" "build"   "-p" name "-j" jobs]
+             ["dune" "runtest" "-p" name "-j" jobs] { with-test }
+            ]
+url {
+  src:
+    "https://github.com/ocaml-ppx/ppx_import/releases/download/1.10.0/ppx_import-1.10.0.tbz"
+  checksum: [
+    "sha256=300f2c7f417b0a1d702432fc13ce3bd2e90ac7b2a2796ca35899c942ca81556f"
+    "sha512=835d5abff0f1eba28313f5925beaeb8c27a4458c91cf395fdd28b984b0745aad8725b4a8a921517ad09ed567ff8586f48a4afa4127eab6c2f773efc5d50c40fb"
+  ]
+}
+x-commit-hash: "bba10ee113166982fda75687f31439d63271b560"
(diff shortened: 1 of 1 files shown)


packages/ppx_import/ppx_import.1.10.0/opam
```diff
@@ -0,0 +1,42 @@
...
+  )
+  |
+  (
+  "ocaml"                   { >= "4.10.0"               }
+  "ppx_sexp_conv"           { with-test  & >= "v0.13.0" }
+  "dune"                    {              >= "1.11.0"  }
+  "ppxlib"                  {              >= "0.26.0"  }
+  "ounit"                   { with-test                 }
+  "ppx_deriving"            { with-test  & >= "4.2.1"   }
+  )
```
kit-ty-kate, line 28: I’m not sure to understand why the dependencies has to be so complicated. Both sides of the disjunction is basically the same except for `ppx_sexp_conv`

This can be simplified into:
```suggestion
  "ocaml"                  {>= "4.05.0" &  < "4.10.0"  } | ("ocaml" {>= "4.10.0"} "ppx_sexp_conv" {with-test & >= "v0.13.0"})
  "dune"                    {              >= "1.11.0"  }
  "ppxlib"                  {              >= "0.26.0"  }
  "ounit"                   { with-test                 }
  "ppx_deriving"            { with-test  & >= "4.2.1"   }
```

kit-ty-kate: I’m also unsure why avoiding installing ppx_sexp_conv (especially with such broad constraints) in the first place.
Looking at https://github.com/ocaml-ppx/ppx_import/pull/69 it seems to be because it’s not used but in this case it wouldn’t really matter if it’s installed or not and would simplify everything.

Could we remove the disjunction all together?

kit-ty-kate:
```suggestion
  "ocaml"                   {              >= "4.05.0"  }
  "dune"                    {              >= "1.11.0"  }
  "ppxlib"                  {              >= "0.26.0"  }
  "ounit"                   { with-test                 }
  "ppx_deriving"            { with-test  & >= "4.2.1"   }
  "ppx_sexp_conv"           { with-test  & >= "v0.13.0" }
```

ejgallego: > I’m also unsure why avoiding installing ppx_sexp_conv (especially with such broad constraints) in the first place. Looking at [ocaml-ppx/ppx_import#69](https://github.com/ocaml-ppx/ppx_import/pull/69) it seems to be because it’s not used but in this case it wouldn’t really matter if it’s installed or not and would simplify everything.
> 
> Could we remove the disjunction all together?

I'm afraid the constraint is needed for ppx_import test-suite, as it is not possible to install ppxlib gt 0.26 than ppx_sexp_conv in OCaml 4.09 or less

kit-ty-kate: ah I understand the issue now. Fair enough. I’ll commit the simplified disjunction instead then

packages/ppx_import/ppx_import.1.10.0/opam
```diff
@@ -0,0 +1,31 @@
...
+bug-reports: "https://github.com/ocaml-ppx/ppx_import/issues"
+dev-repo: "git+https://github.com/ocaml-ppx/ppx_import.git"
+tags: [ "syntax" ]
+
+depends: [
+  "ocaml"                   {>= "4.05.0" &  < "4.10.0"  } | ("ocaml" {>= "4.10.0"} "ppx_sexp_conv" {with-test & >= "v0.13.0"})
+  "dune"                    {              >= "1.11.0"  }
+  "ppxlib"                  {              >= "0.26.0"  }
+  "ounit"                   { with-test                 }
+  "ppx_deriving"            { with-test  & >= "4.2.1"   }
```
ejgallego, line 17: I think this will make ppx_import `-with-test` uninstallable on OCaml 4.09 downwards.

kit-ty-kate: I’m not seeing it. How so? (also this was already there in the original formula)

kit-ty-kate: ppx_deriving < 5.0 does not use ppxlib

kit-ty-kate: but also 5.2.1 is compatible: https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/5adf43152e79f493b22f5029fb5152e38cab1e9c/variant/opam-2.1,compilers,4.05,ppx_import.1.10.0,tests

ejgallego: ppx_deriving is fine, but note you are not installing ppx_sexp_conv (and the dune file disable these tests)

I'm a bit lost on what you mean here.

ejgallego: The change is already pushed upstream.

> I’m lost as well. CI shows everything working on 4.05 upwards so I’m merging. Thanks a lot!

It is working because you placed the or constraint, my comment was for the version without the or constraint, which can't possible work.

Thanks!

ejgallego: In particular note this https://github.com/ocaml-ppx/ppx_import/blob/master/src_test/ppx_deriving_sexp/dune#L3

PR #21533, opened 2022-06-15
`ppx_deriving_yojson` Yojson 2.0 conflicts

Leonidas-from-XIV:
This fixes the failures of #21523.

### DIFF
--- a/packages/comby-kernel/comby-kernel.1.4.1/opam
+++ b/packages/comby-kernel/comby-kernel.1.4.1/opam
@@ -28,6 +28,7 @@ depends: [
   "ppx_deriving_yojson" {>= "3.6.0"}
   "pcre"
   "bisect_ppx" {with-test & dev & >= "2.5.0"}
+  "yojson" {< "2.0.0"}
 ]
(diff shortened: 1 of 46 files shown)


Leonidas-from-XIV:
@mseri Looks like `mparser.1.2` has broken checksums, can you replace the tarball by the archived one?

mseri:
It is uploaded here: https://github.com/ocaml/opam-source-archives/raw/main/mparser-1.2.tar.gz
Can you update the URL, I cannot make the PR right now?

mseri:
Do we really want to fix it further? Ping @kit-ty-kate for an extra opinion, imho it is a strict improvement and it is good to be merged, we can keep fixing thing incrementally separately

Leonidas-from-XIV:
No, I think I'm good for now. The remaining minimum-version failures are a rather complicated field of ppx-rewriters like omp, ppxlib, ppx_tools_versioned and their transitive dependencies which is probably a substantial amount of work to get right and I think this PR already represents a decent improvement over the status quo.

mseri:
Thanks

PR #21547, opened 2022-06-16
new file:   packages/vector3/vector3.1.0.0/opam

UnixJunkie:
switched to dune


UnixJunkie:
just to make it clear: vector3 is the name of the opam package. The rest is a version number.

UnixJunkie:
done; thanks for the fixes

PR #21544, opened 2022-06-16
[new release] dlm (0.3.3)

psafont:
Libdlm bindings

- Project page: <a href="https://github.com/xapi-project/ocaml-dlm/">https://github.com/xapi-project/ocaml-dlm/</a>
- Documentation: <a href="https://xapi-project.github.io/ocaml-dlm/doc">https://xapi-project.github.io/ocaml-dlm/doc</a>

##### CHANGES:

Remove unused "primitive"

The metadata hasn't been maintained in quite some time, so I expect some tweaks

PR #21550, opened 2022-06-16
unix-errno < 0.6: don't use ocamlbuild 0.9.0

mseri:
It has issues with stubs. The error was seen in the lower bound tests of #21544 
The error is
```
+ ocamlfind ocamlmklib -o unix/unix_errno_stubs -L/usr/local/lib unix/unix_errno_util.o unix/unix_errno_stubs.o
# gcc: error: unix/unix_errno_util.o: No such file or directory
# gcc: error: unix/unix_errno_stubs.o: No such file or directory
```
but is not very informative, the problem has been seen in other instances of packages with stubs when compiling using ocamlbuild 0.9.0

### DIFF
--- a/packages/unix-errno/unix-errno.0.2.0/opam
+++ b/packages/unix-errno/unix-errno.0.2.0/opam
@@ -13,7 +13,7 @@ depends: [
   "ocamlfind" {build}
   "ctypes" {>= "0.4.0" & < "0.6.0"}
   "rresult"
-  "ocamlbuild" {build}
+  "ocamlbuild" {build & != "0.9.0"}
 ]
 synopsis: "Unix errno types, maps, and support"
 description: """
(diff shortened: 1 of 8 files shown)


mseri:
I don’t know what to do with the assembler errors. Should we merge it as is or make it unavailable there?

packages/unix-errno/unix-errno.0.2.0/opam
```diff
@@ -15,6 +15,9 @@ depends: [
   "rresult"
   "ocamlbuild" {build}
 ]
+conflicts: [
+  "ocamlbuild" {= "0.9.0"}
+]
```
kit-ty-kate, line 20: Doing this would be simpler and easier to maintain
```suggestion
  "ocamlbuild" {build & != "0.9.0"}
]
```

PR #21554, opened 2022-06-17
[new release] biniou (1.2.2)

tmattio:
Binary data format designed for speed, safety, ease of use and backward compatibility as protocols evolve

- Project page: <a href="https://github.com/mjambon/biniou">https://github.com/mjambon/biniou</a>
- Documentation: <a href="https://mjambon.github.io/biniou/">https://mjambon.github.io/biniou/</a>

##### CHANGES:

- 5.0 compatibilty. Depend on camlp-streams explicitly (mjambon/biniou#31)

### DIFF
--- a/packages/biniou/biniou.1.2.2/opam
+++ b/packages/biniou/biniou.1.2.2/opam
@@ -1,6 +1,6 @@
 opam-version: "2.0"
 build: [
-  ["dune" "subst"] {pinned}
+  ["dune" "subst"] {dev}
   ["dune" "build" "-p" name "-j" jobs]
   ["dune" "runtest" "-p" name "-j" jobs] {with-test}
   ["dune" "build" "-p" name "@doc"] {with-doc}
--- a/packages/biniou/biniou.1.2.2/opam
+++ b/packages/biniou/biniou.1.2.2/opam
@@ -1,0 +1,47 @@
+opam-version: "2.0"
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+  ["dune" "build" "-p" name "@doc"] {with-doc}
+]
+maintainer: ["martin@mjambon.com"]
+authors: ["Martin Jambon"]
+bug-reports: "https://github.com/mjambon/biniou/issues"
+homepage: "https://github.com/mjambon/biniou"
+doc: "https://mjambon.github.io/biniou/"
+license: "BSD-3-Clause"
+dev-repo: "git+https://github.com/mjambon/biniou.git"
+synopsis:
+  "Binary data format designed for speed, safety, ease of use and backward compatibility as protocols evolve"
+description: """
+
+Biniou (pronounced "be new") is a binary data format designed for speed, safety,
+ease of use and backward compatibility as protocols evolve. Biniou is vastly
+equivalent to JSON in terms of functionality but allows implementations several
+times faster (4 times faster than yojson), with 25-35% space savings.
+
+Biniou data can be decoded into human-readable form without knowledge of type
+definitions except for field and variant names which are represented by 31-bit
+hashes. A program named bdump is provided for routine visualization of biniou
+data files.
+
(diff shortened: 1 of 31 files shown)


tmattio:
cc @rgrinberg

kit-ty-kate:
- `merlin` is using `(implicit_transitive_deps false)` and thus failing to pick up the right `Stream` module.. cc @trefis @voodoos 
- in `coq-serapi`: There is an incompatibility between `Stream.t` from `camlp-streams` coming from `binou` and `Stdlib.Stream.t` required by `Pcoq.Parsable.make`. `coq` master replaces it with yet another different type in https://github.com/coq/coq/pull/15867 which will continue to cause issues. cc @ejgallego

ejgallego:
Hi Kate, how come Stdlib.Stream.t and camlp-stremas Stream.t are not the same type? That will cause issues in every project actually?

Coq's master forked Stream.t but it is under Gramlib.Stream.t so it should be ok I think, but none of that code is in opal yet so I'll continue discussion in Coq's PR

kit-ty-kate:
> I am not sure about what we should do concerning the breakage of the current version of Merlin due to this new version of `biniou`. Should we add a upper-bound on biniou's version ?

yes, I’m doing it right now.

Leonidas-from-XIV:
I'm confused why camlp-streams chose to go this way. Wouldn't it make sense that in the case of older OCaml versions the package would pass through the `Stream` module (in the case of OCaml 4.14 dropping the deprecation warnings)? Thus packages can use it as drop-in implementation of `Stream` for both OCaml 4 and 5? Somewhat similar to how the `seq` package adds `Seq` for _older_ versions of OCaml and is a no-op on newer versions?

This also solves the issue of "accidental poisoning" via transitive dependencies as in the Merlin case since it doesn't matter as the types are identical in cases where the compiler comes with streams.

I'm probably missing something, so I'd appreciate being enlightened:-)

ejgallego:
I share @Leonidas-from-XIV view, the current setup is otherwise very brittle not even CI can catch problems, as these depend on build ordering. For example, on a machine with enough jobs, installing coq-serapi may trigger the build of camlp-streams before Coq, thus thinks would work.

kit-ty-kate:
> I share @Leonidas-from-XIV view, the current setup is otherwise very brittle not even CI can catch problems, as these depend on build ordering. For example, on a machine with enough jobs, installing coq-serapi may trigger the build of camlp-streams before Coq, thus thinks would work.

`camlp-streams` is a normal findlib package, it does *not* install modules in `%{lib}%/ocaml/` so there is no issue with build ordering.

Leonidas-from-XIV:
No, but on 4.14 (and older) it is possible to mix `Stream` from the compiler (by packages that predate the move out of the Stdlib, they get deprecation warnings but compile) with `Stream` from `camlp-streams` (like biniou in this case, but I also considered using `camlp-streams` in Yojson to be compatible across 4.x and 5.x) so programs might end up having two identical but incompatible `Stream.t` types?

dra27:
#21656 releases camlp-streams 5.0.1. In OCaml terms, this should be a drop-in 5.x compatibility replacement for anything which uses `Stream.t` - for 4.08-4.14, the library re-exports `Stdlib.Stream`/`Stdlib.Genlex` as `Stream`/`Genlex` (with strengthened types), so the mixing of camlp-streams and non-camlp-streams libraries is transparent. For 4.07 and earlier it is definitely a drop-in replacement, as it's an empty library.

The only "gotcha" is the limitations of `implicit_transitive_deps false` in Dune, but this is more down to the usability of `implicit_transitive_deps` than a problem with the library approach. The 4.14 version of camlp-streams _has_ to shadow  `Stdlib.Stream` and `Stdlib.Genlex` in order to suppress the deprecation warnings.

Leonidas-from-XIV:
Thanks @dra27, this looks excellent! I appreciate how the older version of camlp-streams is restricted to OCaml 5.0+ where it is still safe to use while resolving the problems for most packages on older versions :+1:

kit-ty-kate:
Thanks, sorry for the delay

PR #21557, opened 2022-06-17
[new release] ppxlib (0.25.1)

pitag-ha:
Standard library for ppx rewriters

- Project page: <a href="https://github.com/ocaml-ppx/ppxlib">https://github.com/ocaml-ppx/ppxlib</a>
- Documentation: <a href="https://ocaml-ppx.github.io/ppxlib/">https://ocaml-ppx.github.io/ppxlib/</a>

##### CHANGES:

- Add support for OCaml 5.0 (ocaml-ppx/ppxlib#355, @pitag-ha)

PR #21559, opened 2022-06-17
Remove package `ppxlib.0.25.0~5.00preview`

pitag-ha:
Since the magic numbers have been bumped, the preview doesn't even compile anymore on 5.0.


kit-ty-kate:
Could you instead add `available: false` to this file directly in #21557 ?

pitag-ha:
Sounds good, yes. Thanks!

PR #21581, opened 2022-06-20
swhid_core.0.1

zapashcanon:


kit-ty-kate:
thanks. Could you return the fixes to the opam file in the repository upstream?

PR #21583, opened 2022-06-20
odoc-parser uses `Stream.t` which is gone in OCaml 5.0

Leonidas-from-XIV:
Upstream has already merged a fix, this is just to prevent old releases from being picked up in OCaml 5.0.


kit-ty-kate:
Thanks

Issue #21578, opened 2022-06-20
lwt 5.5.0 is not compatible with ocaml 5.0


mseri:
The build fails with
```
#=== ERROR while compiling lwt.5.5.0 ==========================================#
# context              2.1.2 | linux/x86_64 | ocaml-variants.5.0.0+trunk | file:///home/opam/opam-repository
# path                 ~/.opam/5.0/.opam-switch/build/lwt.5.5.0
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune exec -p lwt src/unix/config/discover.exe -- --save --use-libev false
# exit-code            1
# env-file             ~/.opam/log/lwt-19-20d2e0.env
# output-file          ~/.opam/log/lwt-19-20d2e0.out
### output ###
# (cd _build/default && /home/opam/.opam/5.0/bin/ocamlc.opt -w -40 -g -bin-annot -I src/unix/config/.discover.eobjs/byte -I /home/opam/.opam/5.0/lib/csexp -I /home/opam/.opam/5.0/lib/dune-configurator -I /home/opam/.opam/5.0/lib/ocaml/unix -no-alias-deps -o src/unix/config/.discover.eobjs/byte/discover.cmo -c -impl src/unix/config/discover.ml)
# File "src/unix/config/discover.ml", line 67, characters 16-32:
# 67 | let uppercase = String.uppercase [@ocaml.warning "-3"]
#                      ^^^^^^^^^^^^^^^^
# Error: Unbound value String.uppercase
```
Seen on [#21571](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/0287ea3926ee890307fd8e8881790606021d0507/variant/opam-2.1,compilers,5.0,hvsock.3.0.1) CI

PR #21589, opened 2022-06-20
lwt < 5.6.0 is not compatible with OCaml 5.0

kit-ty-kate:
fixes https://github.com/ocaml/opam-repository/issues/21578


mseri:
Thanks!

PR #21595, opened 2022-06-21
swhid.0.2

zapashcanon:
Hi,

I updated `swhid` to use the latest `swhid_core`.

Now, both `swhid_compute` and `swhid_types` are deprecated. The policy is still the same (not removing anything) ? I believe they've never been used by anyone so it would be simpler to delete them. But if you prefer I can mark them as deprecated.

PR #21598, opened 2022-06-21
deprecate swhid_compute and swhid_types

redianthus:
see  #21595


kit-ty-kate:
Could you add `flags: deprecated` to them instead of removing them?

zapashcanon:
done

kit-ty-kate:
Thanks

PR #21611, opened 2022-06-22
updated failure message in packages/conf-rdkit/conf-rdkit.1/opam

UnixJunkie:


UnixJunkie:
previously the error message was just for OS X.
Now, I think it is valid for all operating systems.

kit-ty-kate:
Thanks! Sorry for the wait

PR #21612, opened 2022-06-22
Remove dependency on `base-bytes` for packages that use jbuilder/dune

Leonidas-from-XIV:
This package is never required since jbuilder/dune require OCaml 4.02 and 4.02 has Bytes out of the box.

This PR is an exploration of the explanation given by @dra27 in https://github.com/ocaml/opam-repository/pull/21333#issuecomment-1153276739 to see how much breakage will happen if these extraneous dependencies are removed.

For now a draft, will require a few more iterations to capture all affected packages and possibly add patches to them.


mseri:
Unfortunately removing the dependency makes it fail all over the place

PR #21614, opened 2022-06-22
kmt v0.1; initial add to the repo

mgree:
This is the initial add of KMT to OPAM. 😁 

Tests are green on our end: https://github.com/mgree/kmt/actions/runs/2543196473.

### DIFF
--- a/packages/kmt/kmt.0.1/opam
+++ b/packages/kmt/kmt.0.1/opam
@@ -1,10 +1,8 @@
 opam-version: "2.0"
-name: "kmt"
-version: "0.1"
 synopsis: "Framework for deriving Kleene Algebras with Tests (KAT)"
 maintainer: "Michael Greenberg <michael@greenberg.science>"
 authors: "Michael Greenberg <michael@greenberg.science>"
-license: "BSD"
+license: "MIT"
 homepage: "https://github.com/mgree/kmt"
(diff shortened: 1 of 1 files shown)


packages/kmt/kmt.0.1/opam
```diff
@@ -0,0 +1,26 @@
...
+  "ocaml" {>= "4.12"}
+  "z3" {>= "4.8"}
+  "batteries" {>= "3.5"}
+  "ANSITerminal" {>= "0.8"}
+  "fmt" {>= "0.9"}
+  "alcotest" {>= "1.4"}
+  "logs" {>= "0.7"}
+  "cmdliner" {>= "1.0.4"}
+  "ppx_deriving" {>= "5.2"}
+  "dune" {>= "2.9"}
```
kit-ty-kate, line 18:
```suggestion
  "dune" {>= "3.0"}
```
The linter still fails with:
```
Your dune-project file indicates that this package requires at least dune 3.0 but your opam file only requires dune >= 2.9. Please check which requirement is the right one, and fix the other.
```

mgree: Oops---thank you! Fooled by a local switch.

mgree:
Looking at https://github.com/dbuenzli/cmdliner/blob/master/CHANGES.md, I've gotten the minimum version wrong.

PR #21651, opened 2022-06-24
Update ppx_seq to 0.3.2

hyphenrf:
upstream urls changed, should i change them retroactively in the older versions?


kit-ty-kate:
Thanks!

> upstream urls changed, should i change them retroactively in the older versions?

You mean `dev-repo` ? Sure, that’d be great.

PR #21656, opened 2022-06-26
Package camlp-streams.5.0.1

dra27:
### `camlp-streams.5.0.1`
The Stream and Genlex libraries for use with Camlp4 and Camlp5
This package provides two library modules:
- Stream: imperative streams, with in-place update and memoization
  of the latest element produced.
- Genlex: a small parameterized lexical analyzer producing streams
  of tokens from streams of characters.

The two modules are designed for use with Camlp4 and Camlp5:
- The stream patterns and stream expressions of Camlp4/Camlp5 consume
  and produce data of type 'a Stream.t.
- The Genlex tokenizer can be used as a simple lexical analyzer for
  Camlp4/Camlp5-generated parsers.

The Stream module can also be used by hand-written recursive-descent
parsers, but is not very convenient for this purpose.

The Stream and Genlex modules have been part of the OCaml standard library
for a long time, and have been distributed as part of the core OCaml system.
They will be removed from the OCaml standard library at some future point,
but will be maintained and distributed separately in this camlpstreams package.

---
* Homepage: https://github.com/ocaml/camlp-streams
* Source repo: git+https://github.com/ocaml/camlp-streams.git
* Bug tracker: https://github.com/ocaml/camlp-streams/issues

---

### DIFF
--- a/packages/camlp-streams/camlp-streams.5.0/opam
+++ b/packages/camlp-streams/camlp-streams.5.0/opam
@@ -32,7 +32,7 @@ bug-reports: "https://github.com/ocaml/camlp-streams/issues"
 license: "LGPL-2.1-only WITH OCaml-LGPL-linking-exception"
 depends: [
   "dune" {>= "2.5"}
-  "ocaml" {>= "4.02.3"}
+  "ocaml" {>= "5.0"}
 ]
 build: [
   ["dune" "subst"] {dev}
--- a/packages/camlp-streams/camlp-streams.5.0.1/opam
+++ b/packages/camlp-streams/camlp-streams.5.0.1/opam
@@ -1,0 +1,59 @@
+opam-version: "2.0"
+synopsis: "The Stream and Genlex libraries for use with Camlp4 and Camlp5"
+description: """
+
+This package provides two library modules:
+- Stream: imperative streams, with in-place update and memoization
+  of the latest element produced.
+- Genlex: a small parameterized lexical analyzer producing streams
+  of tokens from streams of characters.
+
+The two modules are designed for use with Camlp4 and Camlp5:
+- The stream patterns and stream expressions of Camlp4/Camlp5 consume
+  and produce data of type 'a Stream.t.
+- The Genlex tokenizer can be used as a simple lexical analyzer for
+  Camlp4/Camlp5-generated parsers.
+
+The Stream module can also be used by hand-written recursive-descent
+parsers, but is not very convenient for this purpose.
+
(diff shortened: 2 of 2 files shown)


dinosaure:
Thanks!

packages/camlp-streams/camlp-streams.5.0/opam
```diff
@@ -32,7 +32,7 @@ bug-reports: "https://github.com/ocaml/camlp-streams/issues"
 license: "LGPL-2.1-only WITH OCaml-LGPL-linking-exception"
 depends: [
   "dune" {>= "2.5"}
-  "ocaml" {>= "4.02.3"}
+  "ocaml" {>= "5.0"}
```
jvillard, line 35: Was this change intended?

dra27: Yes, as noted in the commit message: the 5.0 release doesn't work correctly with OCaml 4.x

Lupus: atdgen-runtime depends on camlp-streams, you just broke CI for a lot of folks out there.

dra27: Sorry it's caused hassle, but what you had before was a broken library passing in CI!

jvillard: @dra27 oops, it wasn't in the PR message and I didn't see the commit message. Like @Lupus, I'm here because our CI broke following this package mutation that didn't change the package version, which defeated `opam lock`ing our dependencies to specific package versions.

jvillard: Thanks @samoht, I'll give `--ignore-constraints` a try!

Lupus: Our CI setup is somewhat custom, we used to do `opam install . --locked`, but for our large dependency tree it's too slow to build everything from scratch, played around with opam bin - didn't like it, too complex and fragile infrastructure. So we wrote a small tool that creates two switch export files - one with the most heavy dependencies (like core) and their transitive dependencies, and the other includes all project dependencies and their transitive dependencies as well. Projects are build in Docker with 2 distinct layers for importing switch states: base set of heavy dependencies, that change infrequently, and importing full set of dependencies on top of already imported base set. This way docker can effectively cache large amount of our projects OCaml depedencies. Works really well for us, fast and reliable.

Opam monorepo does not work with all packages in opam, and fails on some of our non-dune dependencies.

PR #21658, opened 2022-06-27
Add jasmin.2022.04.0

eponier:
This adds the Jasmin compiler, version 2022.04.0.

We have interrogations regarding the "license" field, since we distribute not only our own code (MIT), but also code extracted from Coq libraries : the coq stdlib (LGPL 2.1), Coqword (MIT) and mathcomp (CeCILL-B). For now, we just wrote "MIT", I don't know if we should list everything.


eponier:
I fixed some problems, but I don't know what to do with the errors about "ppl" (probably to be fixed "conf-ppl") and "Z.compare" (no idea what the problem is).

kit-ty-kate:
Thanks

PR #21665, opened 2022-06-27
Remove dependency on `base-bytes` from res which is unused

Leonidas-from-XIV:
Upstream PR: https://github.com/mmottl/res/pull/5 As can be seen the `bytes` library is not used in the codebase.


kit-ty-kate:
Thanks!

PR #21667, opened 2022-06-27
Add option to compile ocaml with leak sanitizer.

toots:
This PR adds a configuration package to compile ocaml with `-fsanitize=address` (linux & macos) or `-fsanitize=leak` (linux).

These options provide tools to inspect and detect issues related to memory allocation and can be very useful to developers. See: https://github.com/google/sanitizers/wiki/ for more details.

Because compilation builds and executes its own binary, the checks must disabled during compilation, which is is documented in the package descriptions.

### DIFF
--- a/packages/ocaml-option-address-sanitizer/ocaml-option-address-sanitizer.1/opam
+++ b/packages/ocaml-option-address-sanitizer/ocaml-option-address-sanitizer.1/opam
@@ -2,9 +2,11 @@ opam-version: "2.0"
 synopsis: "Set OCaml to be compiled with address sanitizer"
 description: """
 This configuration package enables memory leak sanitzation using
-the -fsanitize=leak gcc and clang option. Make sure to set
-ASAN_OPTIONS=detect_leaks=0,exitcode=0 while compiling and to
-unset it before running compiled binaries.
+the -fsanitize=leak gcc and clang option.
+"""
+post-messages: """
+Make sure to set ASAN_OPTIONS=detect_leaks=0,exitcode=0
+while compiling and to unset it before running compiled binaries.
 """
 depends: [
   "ocaml-variants" {post & >= "5.0"}
--- a/packages/ocaml-option-address-sanitizer/ocaml-option-address-sanitizer.1/opam
+++ b/packages/ocaml-option-address-sanitizer/ocaml-option-address-sanitizer.1/opam
@@ -1,0 +1,15 @@
+opam-version: "2.0"
+synopsis: "Set OCaml to be compiled with address sanitizer"
+description: """
+This configuration package enables memory leak sanitzation using
+the -fsanitize=leak gcc and clang option. Make sure to set
+ASAN_OPTIONS=detect_leaks=0,exitcode=0 while compiling and to
+unset it before running compiled binaries.
+"""
+depends: [
+  "ocaml-variants" {post & >= "4.14.0"}
+]
+conflicts: [ "ocaml-option-leak-sanitizer" ]
+available: [ os = "macos" | os = "linux" ]
+maintainer: "platform@lists.ocaml.org"
+flags: compiler
--- a/packages/ocaml-option-address-sanitizer/ocaml-option-address-sanitizer.1/opam
+++ b/packages/ocaml-option-address-sanitizer/ocaml-option-address-sanitizer.1/opam
@@ -7,7 +7,7 @@ ASAN_OPTIONS=detect_leaks=0,exitcode=0 while compiling and to
 unset it before running compiled binaries.
 """
 depends: [
-  "ocaml-variants" {post & >= "4.14.0"}
+  "ocaml-variants" {post & >= "5.0"}
 ]
 conflicts: [
   "ocaml-option-32bit"
--- a/packages/ocaml-option-address-sanitizer/ocaml-option-address-sanitizer.1/opam
+++ b/packages/ocaml-option-address-sanitizer/ocaml-option-address-sanitizer.1/opam
@@ -21,8 +21,8 @@ conflicts: [
   "ocaml-option-static"
   "ocaml-option-nnp"
   "ocaml-option-nnpchecker"
-  "ocaml-options-fp"
-  "ocaml-options-only-leak-sanitizer"
+  "ocaml-option-fp"
+  "ocaml-option-leak-sanitizer"
 ]
 available: [ os = "macos" | os = "linux" ]
 maintainer: "platform@lists.ocaml.org"
--- a/packages/ocaml-option-leak-sanitizer/ocaml-option-leak-sanitizer.1/opam
+++ b/packages/ocaml-option-leak-sanitizer/ocaml-option-leak-sanitizer.1/opam
@@ -22,7 +22,7 @@ conflicts: [
   "ocaml-option-nnp"
   "ocaml-option-nnpchecker"
   "ocaml-option-fp"
-  "ocaml-options-only-memory-sanitizer"
+  "ocaml-option-address-sanitizer"
 ]
 available: [ os = "linux" ]
 maintainer: "platform@lists.ocaml.org"
--- a/packages/ocaml-option-leak-sanitizer/ocaml-option-leak-sanitizer.1/opam
+++ b/packages/ocaml-option-leak-sanitizer/ocaml-option-leak-sanitizer.1/opam
@@ -2,9 +2,11 @@ opam-version: "2.0"
 synopsis: "Set OCaml to be compiled with leak sanitizer"
 description: """
 This configuration package enables memory address sanitzation using
-the -fsanitize=address gcc and clang option. Make sure to set
-LSAN_OPTIONS=detect_leaks=0,exitcode=0 while compiling and to
-unset it before running compiled binaries.
+the -fsanitize=address gcc and clang option.
+"""
+post-messages: """
+Make sure to set LSAN_OPTIONS=detect_leaks=0,exitcode=0
+while compiling and to unset it before running compiled binaries.
 """
 depends: [
   "ocaml-variants" {post & >= "5.0"}
--- a/packages/ocaml-option-leak-sanitizer/ocaml-option-leak-sanitizer.1/opam
+++ b/packages/ocaml-option-leak-sanitizer/ocaml-option-leak-sanitizer.1/opam
@@ -7,7 +7,7 @@ LSAN_OPTIONS=detect_leaks=0,exitcode=0 while compiling and to
 unset it before running compiled binaries.
 """
 depends: [
-  "ocaml-variants" {post & >= "4.14.0"}
+  "ocaml-variants" {post & >= "5.0"}
 ]
 conflicts: [
   "ocaml-option-32bit"
--- a/packages/ocaml-option-leak-sanitizer/ocaml-option-leak-sanitizer.1/opam
+++ b/packages/ocaml-option-leak-sanitizer/ocaml-option-leak-sanitizer.1/opam
@@ -21,7 +21,7 @@ conflicts: [
   "ocaml-option-static"
   "ocaml-option-nnp"
   "ocaml-option-nnpchecker"
-  "ocaml-options-fp"
+  "ocaml-option-fp"
   "ocaml-options-only-memory-sanitizer"
 ]
 available: [ os = "linux" ]
--- a/packages/ocaml-option-leak-sanitizer/ocaml-option-leak-sanitizer.1/opam
+++ b/packages/ocaml-option-leak-sanitizer/ocaml-option-leak-sanitizer.1/opam
@@ -1,0 +1,15 @@
+opam-version: "2.0"
+synopsis: "Set OCaml to be compiled with leak sanitizer"
+description: """
+This configuration package enables memory address sanitzation using
+the -fsanitize=address gcc and clang option. Make sure to set
+LSAN_OPTIONS=detect_leaks=0,exitcode=0 while compiling and to
+unset it before running compiled binaries.
+"""
+depends: [
+  "ocaml-variants" {post & >= "4.14.0"}
+]
+conflicts: [ "ocaml-option-address-sanitizer" ]
+available: [ os = "linux" ]
+maintainer: "platform@lists.ocaml.org"
+flags: compiler
--- a/packages/ocaml-options-only-leak-sanitizer/ocaml-options-only-leak-sanitizer.1/opam
+++ b/packages/ocaml-options-only-leak-sanitizer/ocaml-options-only-leak-sanitizer.1/opam
@@ -9,7 +9,21 @@ unset it before running compiled binaries.
 depends: [
   "ocaml-variants" {post & >= "4.14.0"}
 ]
-conflicts: [ "ocaml-option-address-sanitizer" ]
+conflicts: [
+  "ocaml-option-32bit"
+  "ocaml-option-afl"
+  "ocaml-option-bytecode-only"
+  "ocaml-option-default-unsafe-string"
+  "ocaml-option-flambda"
+  "ocaml-option-musl"
+  "ocaml-option-no-flat-float-array"
+  "ocaml-option-spacetime"
+  "ocaml-option-static"
+  "ocaml-option-nnp"
+  "ocaml-option-nnpchecker"
+  "ocaml-options-fp"
+  "ocaml-options-only-memory-sanitizer"
+]
 available: [ os = "linux" ]
 maintainer: "platform@lists.ocaml.org"
 flags: compiler
--- a/packages/ocaml-variants/ocaml-variants.4.12.0+options/opam
+++ b/packages/ocaml-variants/ocaml-variants.4.12.0+options/opam
@@ -15,6 +15,10 @@ depends: [
 conflict-class: "ocaml-core-compiler"
 flags: compiler
 setenv: CAML_LD_LIBRARY_PATH = "%{lib}%/stublibs"
+build-env: [
+  [LSAN_OPTIONS = "detect_leaks=0,exitcode=0"]
+  [ASAN_OPTIONS = "detect_leaks=0,exitcode=0"]
+]
 build: [
   [
     "./configure"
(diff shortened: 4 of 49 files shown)


toots:
I'm not sure why the CI tests are failing w/o more logs. I have tested the config successfully on linux and macos.

dra27:
Sorry, I hadn't seen this one. This looks great, in principle, there are a few things which need doing:
- I think this should work with OCaml 4.12 and 4.13 - can we update the +options packages for those versions as well?
- The passage of time means that `ocaml-variants.5.0.0~alpha1+options` now needs updating, too
- All of the `ocaml-options-only-*` packages need these two packages added to their `conflicts` list (which I think may aid some of the CI failures)
- Does the address sanitizer work on macos arm64 as well as amd64?
- Rather than advising the user to build the switch with particular environment variables set, it would be much better to use `build-env` in the packages themselves. Unfortunately, it's not possible to gate `build-env` changes with a formula as you can in `build`, but I think this would be fine for `ocaml.4.14.0` and `ocaml-variants.4.14.0+options`, etc.
```diff
--- a/packages/ocaml-variants/ocaml-variants.4.14.0+options/opam
+++ b/packages/ocaml-variants/ocaml-variants.4.14.0+options/opam
@@ -15,6 +15,10 @@ depends: [
 conflict-class: "ocaml-core-compiler"
 flags: compiler
 setenv: CAML_LD_LIBRARY_PATH = "%{lib}%/stublibs"
+build-env: [
+  [LSAN_OPTIONS = "detect_leaks=0,exitcode=0"]
+  [ASAN_OPTIONS = "detect_leaks=0,exitcode=0"]
+]
 build: [
   [
     "./configure"
diff --git a/packages/ocaml/ocaml.4.14.0/opam b/packages/ocaml/ocaml.4.14.0/opam
index 91ef323cd5..81b186f1ac 100644
--- a/packages/ocaml/ocaml.4.14.0/opam
+++ b/packages/ocaml/ocaml.4.14.0/opam
@@ -17,7 +17,11 @@ setenv: [
   [OCAML_TOPLEVEL_PATH = "%{toplevel}%"]
 ]
 build: ["ocaml" "%{ocaml-config:share}%/gen_ocaml_config.ml" _:version _:name]
-build-env: CAML_LD_LIBRARY_PATH = ""
+build-env: [
+  [CAML_LD_LIBRARY_PATH = ""]
+  [LSAN_OPTIONS = "detect_leaks=0,exitcode=0"]
+  [ASAN_OPTIONS = "detect_leaks=0,exitcode=0"]
+]
 homepage: "https://ocaml.org"
 bug-reports: "https://github.com/ocaml/opam-repository/issues"
 authors: [
```
That said, I may be misunderstanding this, but is the aim here that the programs built and linked by the compiler want this, but not the compiler itself? In other words, do we want `ocamlopt.opt` to have been compiled completely without sanitizer options but to have _the installed_ `libasmrun.a` compiled with with sanitizer options?
cc @fabbing and @oliviernicole who've both been working TSAN in OCaml and have a better understanding of the mechanics of this than I do.

OlivierNicole:
Just to clarify, is this to help compiler developers, or developers of regular OCaml programs? I would assume the former, as OCaml manages memory for the programmer. Or is this for programmers who develop OCaml libraries that contain C code?

toots:
@OlivierNicole it's a bit of both I believe. Enabling this option can help tracking memory leaks in both the internal OCaml code (see: https://github.com/ocaml/ocaml/pull/11406) but also any code compiled when building, including dependencies.

I ended up going that route for a large project where rebuilding individual libraries with this enabled was not practical. We're still tracking memory usage issues, too.

Sorry I haven't had much time to go back to this. Will do my best to find some time soon. Thanks for the pointed advices @dra27

toots:
Thanks for the great review @dra27 I've applied most changes. I say the compiler should be built w/o it but the library and default compiler have it enabled. This would make it possible to compile programs using `ocamlopt` without having to set env variables. I'm not sure how to do this w/o changing a lot of the compiler's code so, perhaps as a first support this would be fine like this?

OlivierNicole:
> That said, I may be misunderstanding this, but is the aim here that the programs built and linked by the compiler want this, but not the compiler itself? In other words, do we want `ocamlopt.opt` to have been compiled completely without sanitizer options but to have _the installed_ `libasmrun.a` compiled with with sanitizer options?



> I say the compiler should be built w/o it but the library and default compiler have it enabled. This would make it possible to compile programs using `ocamlopt` without having to set env variables. I'm not sure how to do this w/o changing a lot of the compiler's code so, perhaps as a first support this would be fine like this?

Yes, ideally `ocamlc.opt` and `ocamlopt.opt` should not be compiled instrumented, but the libraries should be. However, this is not possible in the current build process as the Stdlib compiled by `ocamlopt` is used to create `ocamlc.opt` and `ocamlopt.opt`. Ideally we would need to build two versions of the stdlib, one instrumented and one non instrumented. But in a meeting @dra27 you seemed to imply that this would be painful.

If that's any motivation, we have exactly the same problem with TSan, and for now I don't see how to do better than telling the user to set different env variables during build and during execution of user programs.

toots:
Hey there! Any opinion on this? Would be nice to merge it before it gets outdated again. I think that the current implementation is a good start, I've been using it on different systems already and it's proven useful.

packages/ocaml-variants/ocaml-variants.4.12.0+options/opam
```diff
@@ -15,6 +15,10 @@ depends: [
 conflict-class: "ocaml-core-compiler"
 flags: compiler
 setenv: CAML_LD_LIBRARY_PATH = "%{lib}%/stublibs"
+build-env: [
+  [LSAN_OPTIONS = "detect_leaks=0,exitcode=0"]
```
mseri, line 19: Shouldn't these be added only if one of the two options packages is installed?

raphael-proust:
These options should probably be exposed as an `ocaml-options-only-*` packages with conflicts against all other options. That's because the added environment variables will probably not survive the combination of two different options.

@toots Can you make that change? (Or maybe I missed something?)

packages/ocaml-variants/ocaml-variants.4.12.0+options/opam
toots: Yeah but this is not possible at the moment.

toots:
@raphael-proust packages converted to exclusive packages.

Y'all could we consider a decision here? This PR is 6 months old now. I believe that it's pretty useful but I cannot keep chasing new compilers and options forever.

raphael-proust:
There are failures in the CI but they are just revdeps and the few I checked seem unrelated to these options being set. (Not sure it's the case for all of them.)

I'm ok with the PR as it stands but I'm pretty new here and I'm likely to have missed things.

It might be useful for the `options-only` packages to be able to use `options-*` in the `conflicts` field (obviously this shouldn't indicate a conflict with itself). It'd simplify the many `options-only` packages and PRs like this one. Of course, this is independent on merging this PR.

packages/ocaml/ocaml.4.12.0/opam
```diff
@@ -17,7 +17,11 @@ setenv: [
   [OCAML_TOPLEVEL_PATH = "%{toplevel}%"]
 ]
 build: ["ocaml" "%{ocaml-config:share}%/gen_ocaml_config.ml" _:version _:name]
-build-env: CAML_LD_LIBRARY_PATH = ""
```
mseri, line 20: This should not be changed in the vanilla compilers.

packages/ocaml/ocaml.4.14.1/opam
```diff
@@ -17,7 +17,11 @@ setenv: [
   [OCAML_TOPLEVEL_PATH = "%{toplevel}%"]
 ]
 build: ["ocaml" "%{ocaml-config:share}%/gen_ocaml_config.ml" _:version _:name]
-build-env: CAML_LD_LIBRARY_PATH = ""
+build-env: [
+  [CAML_LD_LIBRARY_PATH = ""]
+  [LSAN_OPTIONS = "detect_leaks=0,exitcode=0"]
+  [ASAN_OPTIONS = "detect_leaks=0,exitcode=0"]
+]
```
kit-ty-kate, line 24: this will trigger a rebuild for all the compiler and the addition seems rather weird. Could you avoid changing the `ocaml` packages?

toots: I don't think so. This is following @dra27's advice here: https://github.com/ocaml/opam-repository/pull/21667#issuecomment-1205234468

If that's too much trouble, we can restrict support to the latest ocaml 5 beta release and move on from there with future versions.

packages/ocaml-options-only-leak-sanitizer/ocaml-options-only-leak-sanitizer.1/opam
```diff
@@ -0,0 +1,29 @@
+opam-version: "2.0"
```
dra27, line 1: Same for this one - `ocaml-option-leak-sanitizer`

packages/ocaml/ocaml.4.14.1/opam
kit-ty-kate: > If that's too much trouble, we can restrict support to the latest ocaml 5 beta release and move on from there with future versions.

in this case I'd rather we do that

packages/ocaml-options-only-leak-sanitizer/ocaml-options-only-leak-sanitizer.1/opam
```diff
@@ -0,0 +1,29 @@
...
+  "ocaml-option-default-unsafe-string"
+  "ocaml-option-flambda"
+  "ocaml-option-musl"
+  "ocaml-option-no-flat-float-array"
+  "ocaml-option-spacetime"
+  "ocaml-option-static"
+  "ocaml-option-nnp"
+  "ocaml-option-nnpchecker"
+  "ocaml-options-fp"
+  "ocaml-options-only-memory-sanitizer"
```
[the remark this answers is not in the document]
toots: Thanks for catching that typo.

toots:
Thanks for the review y'all. @dra27 rename suggestion have been applied and support has been drop for all `4.x` versions. Hopefully this is the one!

packages/ocaml-option-address-sanitizer/ocaml-option-address-sanitizer.1/opam
```diff
@@ -0,0 +1,29 @@
+opam-version: "2.0"
+synopsis: "Set OCaml to be compiled with address sanitizer"
+description: """
+This configuration package enables memory leak sanitzation using
+the -fsanitize=leak gcc and clang option. Make sure to set
+ASAN_OPTIONS=detect_leaks=0,exitcode=0 while compiling and to
+unset it before running compiled binaries.
+"""
+depends: [
+  "ocaml-variants" {post & >= "4.14.0"}
```
kit-ty-kate, line 10:
```suggestion
  "ocaml-variants" {post & >= "5.0"}
```

packages/ocaml-option-leak-sanitizer/ocaml-option-leak-sanitizer.1/opam
```diff
@@ -0,0 +1,29 @@
+opam-version: "2.0"
+synopsis: "Set OCaml to be compiled with leak sanitizer"
+description: """
+This configuration package enables memory address sanitzation using
+the -fsanitize=address gcc and clang option. Make sure to set
+LSAN_OPTIONS=detect_leaks=0,exitcode=0 while compiling and to
+unset it before running compiled binaries.
+"""
+depends: [
+  "ocaml-variants" {post & >= "4.14.0"}
```
kit-ty-kate, line 10:
```suggestion
  "ocaml-variants" {post & >= "5.0"}
```

kit-ty-kate (commented):
I think i forgot to keep the change in `ocaml.5.0.0` as well.
Could you also add the depopt and everything to the 5.1.0 packages as well?

packages/ocaml-option-leak-sanitizer/ocaml-option-leak-sanitizer.1/opam
toots: Done!

toots:
> I think i forgot to keep the change in `ocaml.5.0.0` as well. Could you also add the depopt and everything to the 5.1.0 packages as well?

Done!

packages/ocaml-option-leak-sanitizer/ocaml-option-leak-sanitizer.1/opam
```diff
@@ -0,0 +1,29 @@
...
+  "ocaml-option-default-unsafe-string"
+  "ocaml-option-flambda"
+  "ocaml-option-musl"
+  "ocaml-option-no-flat-float-array"
+  "ocaml-option-spacetime"
+  "ocaml-option-static"
+  "ocaml-option-nnp"
+  "ocaml-option-nnpchecker"
+  "ocaml-option-fp"
+  "ocaml-options-only-memory-sanitizer"
```
kit-ty-kate, line 25:
```suggestion
  "ocaml-option-address-sanitizer"
```

packages/ocaml-option-address-sanitizer/ocaml-option-address-sanitizer.1/opam
```diff
@@ -0,0 +1,29 @@
+opam-version: "2.0"
+synopsis: "Set OCaml to be compiled with address sanitizer"
+description: """
+This configuration package enables memory leak sanitzation using
+the -fsanitize=leak gcc and clang option. Make sure to set
+ASAN_OPTIONS=detect_leaks=0,exitcode=0 while compiling and to
+unset it before running compiled binaries.
```
kit-ty-kate, line 7: Maybe this should a post-message to make sure the users of this package see it
```suggestion
the -fsanitize=leak gcc and clang option.
"""
post-messages: """
Make sure to set ASAN_OPTIONS=detect_leaks=0,exitcode=0
while compiling and to unset it before running compiled binaries.
```

toots: That's a great idea.

packages/ocaml-option-leak-sanitizer/ocaml-option-leak-sanitizer.1/opam
```diff
@@ -0,0 +1,29 @@
+opam-version: "2.0"
+synopsis: "Set OCaml to be compiled with leak sanitizer"
+description: """
+This configuration package enables memory address sanitzation using
+the -fsanitize=address gcc and clang option. Make sure to set
+LSAN_OPTIONS=detect_leaks=0,exitcode=0 while compiling and to
+unset it before running compiled binaries.
```
kit-ty-kate, line 7:
```suggestion
the -fsanitize=address gcc and clang option.
"""
post-messages: """
Make sure to set LSAN_OPTIONS=detect_leaks=0,exitcode=0
while compiling and to unset it before running compiled binaries.
```

kit-ty-kate (approved):
LGTM. @dra27 does that looks good to you too?

kit-ty-kate:
`ocaml-option-address-sanitizer` seems to be failing:
```
#=== ERROR while compiling ocaml-variants.5.0.0+trunk =========================#
# context              2.2.0~alpha~dev | linux/x86_64 | ocaml-variants.5.0.0+trunk | pinned(https://github.com/ocaml/ocaml/archive/5.0.tar.gz)
# path                 ~/.opam/5.0/.opam-switch/build/ocaml-variants.5.0.0+trunk
# command              ~/.opam/opam-init/hooks/sandbox.sh build ./configure --prefix=/home/opam/.opam/5.0 --docdir=/home/opam/.opam/5.0/doc/ocaml -C LDFLAGS=-Wl,--no-as-needed,-ldl CC=gcc -ldl -fsanitize=address -fno-omit-frame-pointer -O1 -g --disable-warn-error
# exit-code            77
# env-file             ~/.opam/log/ocaml-variants-7-7ca6b7.env
# output-file          ~/.opam/log/ocaml-variants-7-7ca6b7.out
### output ###
# configure: creating cache config.cache
# configure: Configuring OCaml version 5.0.0+dev8-2022-10-12
# checking build system type... x86_64-pc-linux-gnu
# checking host system type... x86_64-pc-linux-gnu
# checking target system type... x86_64-pc-linux-gnu
# checking for ld... ld
# checking how to print strings... printf
# checking for gcc... gcc -ldl -fsanitize=address -fno-omit-frame-pointer -O1 -g
# checking whether the C compiler works... yes
# checking for C compiler default output file name... a.out
# checking for suffix of executables... 
# checking whether we are cross compiling... configure: error: in `/home/opam/.opam/5.0/.opam-switch/build/ocaml-variants.5.0.0+trunk':
# configure: error: cannot run C compiled programs.
# If you meant to cross compile, use `--host'.
# See `config.log' for more details
```

toots:
> `ocaml-option-address-sanitizer` seems to be failing:
> 
> ```
> #=== ERROR while compiling ocaml-variants.5.0.0+trunk =========================#
> # context              2.2.0~alpha~dev | linux/x86_64 | ocaml-variants.5.0.0+trunk | pinned(https://github.com/ocaml/ocaml/archive/5.0.tar.gz)
> # path                 ~/.opam/5.0/.opam-switch/build/ocaml-variants.5.0.0+trunk
> # command              ~/.opam/opam-init/hooks/sandbox.sh build ./configure --prefix=/home/opam/.opam/5.0 --docdir=/home/opam/.opam/5.0/doc/ocaml -C LDFLAGS=-Wl,--no-as-needed,-ldl CC=gcc -ldl -fsanitize=address -fno-omit-frame-pointer -O1 -g --disable-warn-error
> # exit-code            77
> # env-file             ~/.opam/log/ocaml-variants-7-7ca6b7.env
> # output-file          ~/.opam/log/ocaml-variants-7-7ca6b7.out
> ### output ###
> # configure: creating cache config.cache
> # configure: Configuring OCaml version 5.0.0+dev8-2022-10-12
> # checking build system type... x86_64-pc-linux-gnu
> # checking host system type... x86_64-pc-linux-gnu
> # checking target system type... x86_64-pc-linux-gnu
> # checking for ld... ld
> # checking how to print strings... printf
> # checking for gcc... gcc -ldl -fsanitize=address -fno-omit-frame-pointer -O1 -g
> # checking whether the C compiler works... yes
> # checking for C compiler default output file name... a.out
> # checking for suffix of executables... 
> # checking whether we are cross compiling... configure: error: in `/home/opam/.opam/5.0/.opam-switch/build/ocaml-variants.5.0.0+trunk':
> # configure: error: cannot run C compiled programs.
> # If you meant to cross compile, use `--host'.
> # See `config.log' for more details
> ```

This is odd. I can't reproduce locally and the invocation hasn't changed much. I'm temped to consider it a CI issue. I've rebased the PR, let's see if a new CI run works.

toots:
@kit-ty-kate I can confirm that manually running:
```
opam switch create asan  --packages ocaml-option-address-sanitizer.1,ocaml-variants.5.0.0+trunk
```
does work on a ubuntu system. 

My guess is that something changed in the docker CI config. These debugging tools might need specific OS package/versions, I'm not surprised that they wouldn't work in all cases.

kit-ty-kate:
ping @dra27 for a review

(also, at this point let’s wait for the next OCaml 5.0 beta to get this merged and avoid unnecessary recompilation for the users)

toots:
> ping @dra27 for a review
> 
> (also, at this point let’s wait for the next OCaml 5.0 beta to get this merged and avoid unnecessary recompilation for the users)

Ok. Thanks for following-up!

raphael-proust:
There is a merge conflict that's not syntactical so git doesn't complain.

Waiting for the beta to fix this.

kit-ty-kate:
This PR got merged as part #22566. Thanks and sorry for the wait.

toots:
Awesome thanks for taking care of this!

PR #21668, opened 2022-06-28
Remove dependency on `base-bytes` on stringext versions using dune/jbuilder

Leonidas-from-XIV:
They do not use the `bytes` package in the codebase itself.

Upstream PR: https://github.com/rgrinberg/stringext/pull/12


kit-ty-kate:
Thanks

PR #21629, opened 2022-06-24
Update the merlin preview for OCaml 5.0

kit-ty-kate:
cc @voodoos

PR #21671, opened 2022-06-28
Update `dot-merlin-reader`'s ocaml upper bounds

voodoos:
The package does not build with OCaml 5 since it uses the now-removed `String.capitalize` function.
See #21629


kit-ty-kate:
Thanks!

PR #21675, opened 2022-06-28
Package bsmi_ocaml.0.1.5

ghost:
### `bsmi_ocaml.0.1.5`
Bsmi common ocaml library
Provided common tools deal with time,date, strings, html, filters.

---
* Homepage: https://github.com/infobsmi/bsmi_ocaml
* Source repo: git+https://github.com/infobsmi/bsmi_ocaml.git
* Bug tracker: https://github.com/infobsmi/bsmi_ocaml/issues

---


fccm:
Some comments on your lib.
It would help if you use English to contribute to OSS.
If you also want to put Chinese I think everyone will be happy with it, but then I would suggest maybe some kind of bilingualism in your project: You can put Chinese and English side by side.
Google translate gives good results on your Chinese comments, but you will notice that every package in Opam uses English.
For git logs you can for example put several `-m` parameter when you commit, I would suggest the first one to be English, and the second one Chinese if you really want to use your born language. For example:
```
git commit -m 'Update build scripts' -m '更新构建脚本'
```

English or bilingualism would also be appreciated in ocamldoc comments.

Also the name of a file in OCaml will become the module name to call the functions that it contains, so you don't need to create a module with the same name in a file:

`Phoneutils.Phoneutils.isValidCnPhone` is redundant.

If you watch the pull request page of opam-repository you will notice that there are 33 packages there waiting, some are there since 3 weeks, so just be patient for a review.

In my opinion the only function that could be worth it is the `isValidCnPhone`, but if you wish to make this function easy to find for someone else, a README or an ocamldoc containing the right key words will be enough, if you want someone can find it with google, or the search engine of github.

In the opam file the synopsis should contain a short description of what the lib provides, not only its name.
The description advertises for html and filters, there is nothing about this in you lib yet. Maybe you plan to add it later, so only advertise it when it will be there.
For string utilities there is only a `String.sub` with a zero parameter. Netrc only contains one function that reads `/etc/resolv.conf`.

In my opinion this lib doesn't contain enough yet to be a candidate for inclusion in opam.

But the first version of Vector3 that was included in Opam was also very small as you can see:
[Vector3 v0.2](https://github.com/UnixJunkie/vector3/blob/v0.2/vector3.ml)

mseri:
I agree with @kit-ty-kate that it would be better to change the package name (for opam, not the repository name, that can easily remain what it is now) into `bsmi` or `bsmi_common`. You can only install ocaml packages with opam, so this will not conflict with other packages. I also agree that synpsis and descriptions should be made clearer.

Please don't take this personally. We make requests like this all packages, especially when it is the first submission, so that all future submissions will be fast and smooth.

As a final aside, neither me or @kit-ty-kate know how to block users (I did not even know it was possible) and we can't find any blocked users, so we don't really know why you get such error.

PR #21697, opened 2022-07-01
Package camlot.0.0.1

igarnier:
### `camlot.0.0.1`
Solve optimal transportation problems using the FastTransport library
This library provides binding to the FastTransport library by Nicolas Bonneel, available here: https://perso.liris.cnrs.fr/nicolas.bonneel/FastTransport/.
Part of the C++ wrapper is copied from https://github.com/PythonOT/POT

---
* Homepage: http://github.com/igarnier/camlot
* Source repo: git+https://github.com/igarnier/camlot
* Bug tracker: http://github.com/igarnier/camlot

---
# 0.0.1
- Initial release

---


igarnier:
Added a version constraint on OCaml version

kit-ty-kate:
The package itself seems good, however since it’s the first version of this package I’m wondering if it could be renamed to something else?

This seems a bit too close to https://github.com/upenn-cis1xx/camelot (which has been around in opam for a bit more than two years) and it might cause issues with users installing one thinking it’s the other and getting confused.

I’m open to discussion of course, what do you think?

igarnier:
> I’m open to discussion of course, what do you think?

I think you're right, renaming this package is the reasonable thing to do. I guess I'll do things properly and rename the repo as well.

PR #21699, opened 2022-07-01
Fix dot-merlin-reader version: 4.6-414 -> 4.6

voodoos:
I made a mistake in yesterday release: dot-merlin-reader is not specific to a version of ocaml.
This PR changes the version by renaming the directory.

I don't know if I am allowed to do this ? If not then it is not an issue to keep the current version.


kit-ty-kate:
Since it’s only been a day and it’s unlikely to be stored in lock files this should be fine to merge. Thanks.

PR #21704, opened 2022-07-03
Package dkml-base-compiler.4.12.1~v1.0.0

jonahbeckford:
### `dkml-base-compiler.4.12.1~v1.0.0`
OCaml cross-compiler and libraries from the DKML distribution that works with at least Win32 and macOS
The DKML distribution of the OCaml bytecode and native compiler, Stdlib and the other OCaml libraries (str, unix, bigarray, etc.).
A cross-compiler for macOS x86_64 to macOS arm64 is included; for build consistency the regular OCaml compiler will be for x86_64 regardless of whether the build machine is Apple Silicon.
Install with something like: opam switch create dkml-4.12.1 '--formula="dkml-base-compiler" {>= "4.12.1~" & < "4.13.0~"}'

---
* Homepage: https://github.com/diskuv/dkml-compiler
* Source repo: git+https://github.com/diskuv/dkml-component-ocamlcompiler.git
* Bug tracker: https://github.com/diskuv/dkml-compiler/issues

---

### DIFF
--- a/packages/dkml-base-compiler/dkml-base-compiler.4.12.1~v1.0.0/opam
+++ b/packages/dkml-base-compiler/dkml-base-compiler.4.12.1~v1.0.0/opam
@@ -1,0 +1,106 @@
+opam-version: "2.0"
+synopsis: "OCaml cross-compiler and libraries from the DKML distribution that works with at least Win32 and macOS"
+description:
+  """The DKML distribution of the OCaml bytecode and native compiler, Stdlib and the other OCaml libraries (str, unix, bigarray, etc.).
+A cross-compiler for macOS x86_64 to macOS arm64 is included; for build consistency the regular OCaml compiler will be for x86_64 regardless of whether the build machine is Apple Silicon.
+Install with something like: opam switch create dkml-4.12.1 '--formula="dkml-base-compiler" {>= "4.12.1~" & < "4.13.0~"}'"""
+maintainer: ["opensource+diskuv-ocaml@support.diskuv.com"]
+authors: ["Diskuv, Inc. <opensource+diskuv-ocaml@support.diskuv.com>"]
+license: "Apache-2.0"
+homepage: "https://github.com/diskuv/dkml-compiler"
+bug-reports: "https://github.com/diskuv/dkml-compiler/issues"
+depends: [
+  "ocaml" {= "4.12.1" & post}
+
+  "base-unix" {post}
+  "base-bigarray" {post}
+  "base-threads" {post}
+]
+conflict-class: "ocaml-core-compiler"
+flags: [ compiler avoid-version ]
+build: [
+  # OCaml source code
+  ["install" "-d" "dl/ocaml/flexdll"]
+  ["tar" "xCfz" "dl/ocaml"          "dl/ocaml.tar.gz"   "--strip-components=1"]
+  ["tar" "xCfz" "dl/ocaml/flexdll"  "dl/flexdll.tar.gz" "--strip-components=1"]
+
+  # Create a DKMLDIR. Its structure mimics a git submodule setup.
+  #   <dkmldir>/.dkmlroot
+  ["install" "-d" "dkmldir"]
+  #     4.12.1~v0.4.1          --dkml-semver--> 0.4.1
+  #     4.12.1~v0.4.1~prerel69 --dkml-semver--> 0.4.1-prerel69
+  ["sh" "-c" "printf 'dkml_root_version=%s\\n' '%{version}%' | sed 's/[0-9.]*~v//; s/~/-/' > dkmldir/.dkmlroot"]
+
+  #   <dkmldir>/vendor/drc/
+  ["install" "-d" "dkmldir/vendor/drc"]
+  ["tar" "xCfz" "dkmldir/vendor/drc" "dl/dkml-runtime-common.tar.gz" "--strip-components=1"]
+
+  #   <dkmldir>/vendor/dkml-compiler/
+  ["install" "-d" "dkmldir/vendor/dkml-compiler/src"]
+  ["sh" "-eufc" "tar cCf src/ - . | tar xCf dkmldir/vendor/dkml-compiler/src/ -"]
+]
+install: [
+  # Run r-c-ocaml-1-setup.sh
+  [
+    "env" "TOPDIR=dkmldir/vendor/drc/all/emptytop"
+      "dkmldir/vendor/dkml-compiler/src/r-c-ocaml-1-setup.sh"
+      "-d" "dkmldir"
+      "-t" "%{prefix}%"
+      "-f" "src-ocaml"
+      "-g" "%{_:share}%/mlcross"
+      "-v" "dl/ocaml"
+      #   Host architectures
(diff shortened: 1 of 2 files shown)


jonahbeckford:
CI explanation: All tests were skipped because ocaml-config.3, which introduces dkml-base-compiler, builds only OCaml 5+ except for Windows. The reasoning for the OCaml 5+ constraint was to not force a recompile of all packages.

https://github.com/ocaml/opam-repository/blob/97da9a1b68b824a65a09e5f7d071fcf2da35bd1b/packages/ocaml-config/ocaml-config.3/opam#L13-L18

And there are no Windows machines in the CI fleet so nothing gets tested.

I'm not sure how I feel about nothing being tested! But it is what it is.

---

Somewhat related: dkml-base-compiler.4.12.1 works with macOS and is useful because it comes bundled with a x86_64 -> arm64 cross-compiler. I would like to relax the constraints on ocaml-config.3 to include macOS at 4.12.1. Since macOS is also not in the CI fleet, it should not force a recompile of all the packages.

jonahbeckford:
> I think this needs the change to [packages/ocaml/ocaml.4.12.1/opam](https://github.com/ocaml/opam-repository/pull/20536/files#diff-afb505da8001852a85d05b9fb78eb13dfd47d327f0c3a74ef14fadc10c910166) from #20536?

Yep. Added!

jonahbeckford:
Removed ocaml-option-32bit as per talk with @kit-ty-kate . Also updated version to `4.12.1-v1.0.0` to reflect release of Diskuv OCaml 1.0.0.

kit-ty-kate:
is the tilde in the version number expected or did you mean a dash instead?
It does changes the version ordering:
`dkml-base-compiler.4.12.1~v1.0.0` < `dkml-base-compiler.4.12.1` < `dkml-base-compiler.4.12.1-v1.0.0`

jonahbeckford:
> is the tilde in the version number expected or did you mean a dash instead? It does changes the version ordering: `dkml-base-compiler.4.12.1~v1.0.0` < `dkml-base-compiler.4.12.1` < `dkml-base-compiler.4.12.1-v1.0.0`

I do mean tilde (`~`) for Opam versioning, and I was aiming for the `4.12.1~v1 < 4.12.1 < 4.12.1-v1` you wrote. Semantically I was trying to say that the DKML compiler is not past version OCaml 4.12.1, which in hindsight was a bit arbitrary.

Sadly I couldn't be consistent with the `~` in tag/branch name ... git does not allow `~` in branch names. So I just kept with the formal semver notation for branch names where `-` means "not past" or prerelease.

kit-ty-kate:
Same error on macOS as with the other two PRs

jonahbeckford:
This has been fixed as well. The lint warning has nothing to do with the PR.

packages/dkml-base-compiler/dkml-base-compiler.4.12.1~v1.0.0/opam
```diff
@@ -0,0 +1,163 @@
...
+bug-reports: "https://github.com/diskuv/dkml-compiler/issues"
+depends: [
+  "ocaml" {= "4.12.1" & post}
+
+  "base-unix" {post}
+  "base-bigarray" {post}
+  "base-threads" {post}
+]
+conflict-class: "ocaml-core-compiler"
+flags: [ compiler avoid-version ]
```
kit-ty-kate, line 20: Since it is relying on `avoid-version` (only available since 2.1) it would be wise to constrain the package to opam 2.1 and above only
```suggestion
flags: [ compiler avoid-version ]
available: opam-version >= "2.1"
```

packages/dkml-base-compiler/dkml-base-compiler.4.12.1~v1.0.0/opam
```diff
@@ -0,0 +1,164 @@
...
+available: opam-version >= "2.1"
```
kit-ty-kate, line 21:
```suggestion
available: opam-version >= "2.1.0"
```

packages/dkml-base-compiler/dkml-base-compiler.4.12.1~v1.0.0/opam
```diff
@@ -0,0 +1,164 @@
...
+      "dkmldir/vendor/dkml-compiler/src/r-c-ocaml-1-setup.sh"
+      "-d" "dkmldir"
+      "-t" "%{prefix}%"
+      "-f" "src-ocaml"
+      "-g" "%{_:share}%/mlcross"
+      "-v" "dl/ocaml"
+      #   Host architectures
+      "-ewindows_x86_64"  { os = "win32" }
+      "-elinux_x86_64"    { os = "linux" }
+      "-edarwin_x86_64"   { os = "macos" }
```
kit-ty-kate, line 105: Does this mean that only x86_64 as a host is supported?

kit-ty-kate: > For Apple Silicon an ARM64 host is fine.

As long as Rosetta2 is available i presume. So in the future when it inevitably gets removed from the standard distribution we'll probably have to restrict against versions of macos.

packages/dkml-base-compiler/dkml-base-compiler.4.12.1~v1.0.0/opam
```diff
@@ -0,0 +1,164 @@
...
+depends: [
+  "ocaml" {= "4.12.1" & post}
+
+  "base-unix" {post}
+  "base-bigarray" {post}
+  "base-threads" {post}
+]
+conflict-class: "ocaml-core-compiler"
+flags: [ compiler avoid-version ]
+available: opam-version >= "2.1.0"
```
kit-ty-kate, line 21:
```suggestion
available: [
  opam-version >= "2.1.0" &
  ( (arch = "x86_64" & (os = "linux" | os = "win32" | os = "macos") |
    (arch = "arm64" & os = "macos") )
]
```

jonahbeckford: Oh that makes sense. Thanks!

packages/dkml-base-compiler/dkml-base-compiler.4.12.1~v1.0.0/opam
jonahbeckford: Yep.

packages/dkml-base-compiler/dkml-base-compiler.4.12.1~v1.0.0/opam
```diff
@@ -0,0 +1,168 @@
...
+
+  "base-unix" {post}
+  "base-bigarray" {post}
+  "base-threads" {post}
+]
+conflict-class: "ocaml-core-compiler"
+flags: [ compiler avoid-version ]
+available: [
+  opam-version >= "2.1.0" &
+  ( (arch = "x86_64" & (os = "linux" | os = "win32" | os = "macos") |
```
kit-ty-kate, line 23:
```suggestion
  ( (arch = "x86_64" & (os = "linux" | os = "win32" | os = "macos")) |
```

PR #21727, opened 2022-07-05
opam file for frama-c-metacsl 0.3

vprevosto:
New MetAcsl release following Frama-C 25.0 Manganese


kit-ty-kate:
Thanks, could you return the fix to the opam file upstream so that it is taken into account for next time?

PR #21732, opened 2022-07-06
tezos-bls12-381-polynomial, tezos-plompiler, tezos-plonk 0.1.2

victor-dumitrescu:
Release of tezos-bls12-381-polynomial, tezos-plompiler, and tezos-plonk 0.1.2


victor-dumitrescu:
Hi, from what I could gather it seems to me that the macos failures are unrelated to the packages themselves? Let me know if that's not the case. Thanks!

kit-ty-kate:
Indeed, we’re trying to fix the issue, for the moment they can be ignored.

Thanks!

PR #21736, opened 2022-07-06
Package dkml-package-console.0.2.0

jonahbeckford:
### `dkml-package-console.0.2.0`
Console setup and uninstall executables for Diskuv OCaml (DKML) installation
The setup and uninstall executables are responsible for launching the DKML runners.

---
* Homepage: https://github.com/diskuv/dkml-install-api
* Source repo: git+https://github.com/diskuv/dkml-install-api.git
* Bug tracker: https://github.com/diskuv/dkml-install-api/issues

---

### DIFF
--- a/packages/dkml-package-console/dkml-package-console.0.2.0/opam
+++ b/packages/dkml-package-console/dkml-package-console.0.2.0/opam
@@ -1,0 +1,49 @@
+opam-version: "2.0"
+synopsis:
+  "Console setup and uninstall executables for Diskuv OCaml (DKML) installation"
+description:
+  "The setup and uninstall executables are responsible for launching the DKML runners."
+maintainer: ["opensource+diskuv-ocaml@support.diskuv.com"]
+authors: ["Diskuv, Inc. <opensource+diskuv-ocaml@support.diskuv.com>"]
+license: "Apache-2.0"
+homepage: "https://github.com/diskuv/dkml-install-api"
+bug-reports: "https://github.com/diskuv/dkml-install-api/issues"
+#   Components can be assembled on any "host" build machine defined in
+#   https://github.com/diskuv/dkml-install-api/blob/5cfd7b57c79d990c76a9bdc8f8f0fa9f6fd5346f/runner/src/host_abi.ml
+#   into a installer that will run on any end-user machine defined in
+#   https://github.com/diskuv/dkml-install-api/blob/5cfd7b57c79d990c76a9bdc8f8f0fa9f6fd5346f/runner/src/ocaml_abi.ml
+available: os = "win32" | os = "linux" | os = "macos"
+depends: [
+  "alcotest" {>= "1.4.0" & with-test}
+  "odoc" {>= "1.5.3" & with-doc}
+  "dkml-install" {= version}
+  "dkml-install-runner" {= version}
+  "dune" {>= "2.9"}
+  "diskuvbox" {>= "0.1.0"}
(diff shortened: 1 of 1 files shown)


packages/dkml-package-console/dkml-package-console.0.1.0/opam
```diff
@@ -0,0 +1,44 @@
...
+bug-reports: "https://github.com/diskuv/dkml-install-api/issues"
+depends: [
+  "alcotest" {>= "1.4.0" & with-test}
+  "odoc" {>= "1.5.3" & with-doc}
+  "dkml-install" {= version}
+  "dkml-install-runner" {= version}
+  "dune" {>= "2.9"}
+  "diskuvbox" {>= "0.1.0"}
+  "crunch" {>= "3.2.0"}
+  "dkml-component-xx-console" {>= "0.1.0"}
```
kit-ty-kate, line 19: since `dkml-component-xx-console` is only available on Windows (per #21711), do you need this package only on windows (like is it some kind of runtime dependency) or is `dkml-package-console` also only available on Windows?
```suggestion
  "dkml-component-xx-console" {>= "0.1.0" & os = "win32" & (arch = "x86_64" | arch = "x86_32")}
```

jonahbeckford: The dkml-component-* packages are slightly different from ordinary OCaml packages. They are packages that build bytecode binaries and/or download native assets. Several of these assets, plus ocamlrun, make a setup.exe on Windows (or .rpm/.deb on Linux, etc.).

Building bytecode binaries and downloading native assets can be done on any build machine (by design). The only reason why I can't build a Windows setup.exe on a Linux or macOS machine today is because there are two packaging steps that I haven't made cross-platform yet.

So technically the availability constraint for Windows is too limiting. I was initially fine with the availability constraint, but now I realize that a) these availability constraints are viral and must be attached to any indirect consumers and b) eventually we'll have to undo the availability constraints when I support all packaging steps on Unix.

What do you suggest?

jonahbeckford: After further thought, I think explicitly setting the Opam availability to match the installer API's supported ABIs:
* `Linux_(arm32v7|arm32v6|arm64|x86|x86_64)`, `Darwin_(x86_64|arm64)` from https://github.com/diskuv/dkml-install-api/blob/484694c11eb697f58a747e488dd506b963f29a0d/runner/src/host_abi.ml#L20-L37
* `Windows_(x86|x86_64|arm32|arm64)` from https://github.com/diskuv/dkml-install-api/blob/484694c11eb697f58a747e488dd506b963f29a0d/runner/src/host_abi.ml#L49-L83

makes the most sense. Those ABIs constrain the build machines that are used to generate an installer, and the end-user machines that run the installer. All dkml-components and dkml-packages rely on that API, so the Opam availability should naturally be viral.

If I don't hear any objections I'll resubmit dkml-install-api and any existing dkml-component-* and dkml-package-* packages to a) use that Opam availability, and b) document the link to the authoritative ABIs.

jonahbeckford: - [ ] Pending PR #21768. I'll be adding some `available:` constraints common to all components.

jonahbeckford:
This is ready to merge. Thanks!

kit-ty-kate:
Thanks

PR #21743, opened 2022-07-07
2 packages from diskuv/dkml-component-opam at 2.2.0~dkml20220801

jonahbeckford:
This pull-request concerns:
- ~~`dkml-component-staging-opam32.2.2.0~dkml20220707T014024Z`: DKML component for 32-bit versions of opam~~
- ~~`dkml-component-staging-opam64.2.2.0~dkml20220707T014024Z`: DKML component for 64-bit versions of opam~~

-`dkml-component-staging-opam32.2.2.0~dkml20220801`: DKML component for 32-bit versions of opam
-`dkml-component-staging-opam64.2.2.0~dkml20220801`: DKML component for 64-bit versions of opam

---
* Homepage: https://github.com/diskuv/dkml-component-opam
* Source repo: git+https://github.com/diskuv/dkml-component-ocamlcompiler.git
* Bug tracker: https://github.com/diskuv/dkml-component-opam/issues

---

### DIFF
--- a/packages/dkml-component-staging-opam32/dkml-component-staging-opam32.2.2.0~dkml20220801/opam
+++ b/packages/dkml-component-staging-opam32/dkml-component-staging-opam32.2.2.0~dkml20220801/opam
@@ -179,7 +179,7 @@ dev-repo: "git+https://github.com/diskuv/dkml-component-ocamlcompiler.git"
 extra-source "dl/dkml-compiler.tar.gz" {
   src: "https://github.com/diskuv/dkml-compiler/archive/refs/tags/4.12.1-v1.0.0.tar.gz"
   checksum: [
-    "sha256=b4850314ea64d357ce814faebc1f1ef54052e76ea048602401f481cf6bf40455"
+    "sha256=8beda92f97cde6d4a55a836ca6dc9f860bb5f1a6b765b80be4594943288571cf"
   ]
 }
 extra-source "dl/opam.tar.gz" {
--- a/packages/dkml-component-staging-opam32/dkml-component-staging-opam32.2.2.0~dkml20220801/opam
+++ b/packages/dkml-component-staging-opam32/dkml-component-staging-opam32.2.2.0~dkml20220801/opam
@@ -1,0 +1,365 @@
+opam-version: "2.0"
+synopsis: "DKML component for 32-bit versions of opam"
+description: """For 32-bit capable platforms, opam, opam-putenv and opam-installer will be in <share>/staging-files/<platform>.
+But for any platform that does not support 32-bit, this package will install nothing (aka. be a no-op).
+Consumers of the component should place both tools/opam64 and tools/opam32 into the PATH, so that whichever is available can be used."""
+maintainer: ["opensource+diskuv-ocaml@support.diskuv.com"]
+authors: ["Diskuv, Inc. <opensource+diskuv-ocaml@support.diskuv.com>"]
+license: "Apache-2.0"
+homepage: "https://github.com/diskuv/dkml-component-opam"
+bug-reports: "https://github.com/diskuv/dkml-component-opam/issues"
+depends: [
+  "dkml-install"            {>= "0.2.0"}
+  "dune"                    {>= "2.9"}
+  "diskuvbox"               {>= "0.1.0" & build}
+]
+depopts: [
+  "ocaml-system"
+  "dkml-base-compiler"
+  "ocaml-base-compiler"
+  "ocaml-variants"
+  "ocaml-option-32bit"
+]
+build: [
+  # Homebrew's bundle tap, needed for reproducible build auditing in drc's crossplatform-functions.sh.
+  #
+  # The Opam sandbox will stop Homebrew from auto-installing its own tap. In particular 'brew bundle ...'
+  # is automatically installed per https://github.com/Homebrew/homebrew-bundle/tree/4756e4c4cf95485c5ea4da27375946c1dac2c71d#installation,
+  # and it is an "official" tap per
+  # https://github.com/Homebrew/brew/blob/master/Library/Homebrew/official_taps.rb#L11-L18 .
+  #
+  # So our solution is to:
+  # a) Use extra-source:[] to download a Git tarball without tripping over the sandbox firewall
+  # b) Recreate a local file git repository
+  # c) Use the next build:[] steps to let Homebrew know about the bundle tap
+  #
+  # Test with: brew untap homebrew/bundle
+  ["install" "-d" "dl/homebrew-bundle"] {os = "macos"}
+  ["tar" "xCfz" "dl/homebrew-bundle" "dl/homebrew-bundle.tar.gz" "--strip-components=1"] {os = "macos"}
+  ["sh" "-c" """#!/bin/sh
+    set -eufx
+    if command -v brew; then
+      eval $(brew shellenv)
+      HOMEBREW_BREW_FILE="$(command -v brew)"
+      HOMEBREW_LIBRARY="${HOMEBREW_REPOSITORY}/Library"
+      HOMEBREW_PATH="$PATH"
+      HOMEBREW_CACHE="$PWD/dl/homebrew-cache"
+      export HOMEBREW_BREW_FILE HOMEBREW_PATH HOMEBREW_CACHE
+
+      # Clone the Library which contains the Taps/
+      # (and also Homebrew/ which is all the core scripts).
+      install -d dl/homebrew
+      cp -a "$HOMEBREW_LIBRARY" dl/homebrew/ || true
+      HOMEBREW_LIBRARY="$PWD/dl/homebrew/Library"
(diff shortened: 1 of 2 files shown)


mseri:
Macos failure is infrastructural and does not depend on this PR

jonahbeckford:
> Is such a massive version number expected?

My expectations of version number lengths comes from https://semver.org/ (see the examples of prerelease and build metadata version numbers). Nothing in those examples would lead me to expect that the version number here is *massive*.

The intention is to look at the version number and know, at a glance, which Opam commits are present and which are not in the prerelease. This knowledge is important when something goes wrong and a Windows user has to look at `opam --version` to file a bug or to verify that they have a specific fix. An accurate reflection of what is running on their machine is, IMHO, essential until we have full Windows support in Opam.

jonahbeckford:
(One last change to fix the homepage and bug-reports link was just submitted)

jonahbeckford:
- [ ] Pending PR #21768. I'll be adding some `available:` constraints common to all components.

jonahbeckford:
Updated version to `2.2.0~dkml20220801` which matches what is deployed in Diskuv OCaml 1.0.0.

I used a shorter version number because there are very few open Opam PRs left for Windows; it doesn't need to be that precise anymore.

jonahbeckford:
This is ready to merge.

FYI: Even using the Docker reproduction commands at the top of the Opam CI logs ... I could not recreate the network failures I saw in this PR. I could work back from the logs what was happening and use trial and error, but I think there is some magic networking (iptables, etc.) done in Opam CI that isn't in the Dockerfile.

kit-ty-kate:
Same issue on macOS as with https://github.com/ocaml/opam-repository/pull/21756#issuecomment-1210597334

packages/dkml-component-staging-opam32/dkml-component-staging-opam32.2.2.0~dkml20220801/opam
```diff
@@ -288,9 +288,11 @@ extra-source "dl/opam/src_ext/archives/sha-1.15.2.tbz" {
 # END OPAM ARCHIVES
 # -------------------
 extra-source "dl/dkml-runtime-common.tar.gz" {
-  src: "https://github.com/diskuv/dkml-runtime-common/archive/refs/tags/v1.0.0.tar.gz"
+  # This violates the v1.0.0 versioning, but this future patch is needed to resolve build
+  # issues on macOS in Opam CI.
```
kit-ty-kate, line 292: This is not just an opam issue. Regular users will encounter this issue. My guess for why you didn't encounter that yourself is that you ran `brew install` in the past couple of days before testing.
```suggestion
  # issues on macOS in Opam.
```

jonahbeckford:
Thanks; good catch! I've been ignoring macos failures; won't anymore.

jonahbeckford:
```
# + /usr/bin/env HOMEBREW_NO_AUTO_UPDATE=1 HOMEBREW_NO_ANALYTICS=1 HOMEBREW_NO_COLOR=1 HOMEBREW_NO_EMOJI=1 brew bundle dump --force
# ==> Tapping homebrew/bundle
# fatal: could not create work tree dir '/usr/local/Homebrew/Library/Taps/homebrew/homebrew-bundle': Operation not permitted
# Error: Failure while executing; `git clone https://github.com/Homebrew/homebrew-bundle /usr/local/Homebrew/Library/Taps/homebrew/homebrew-bundle --origin=origin --template=` exited with 128.
# Error: Failure while executing; `/usr/local/bin/brew tap homebrew/bundle` exited with 1.
```

I'll use a macOS VM or something similar to see what is happening.

kit-ty-kate:
From `brew bundle`'s documentation:
```
You can skip the installation of dependencies by adding space-separated values to one or more of the following environment variables: HOMEBREW_BUNDLE_BREW_SKIP, HOMEBREW_BUNDLE_CASK_SKIP, HOMEBREW_BUNDLE_MAS_SKIP, HOMEBREW_BUNDLE_WHALEBREW_SKIP, HOMEBREW_BUNDLE_TAP_SKIP.
```
I'm not sure what it refers to when saying "installation of dependencies" but that's my guess

Source: https://docs.brew.sh/Manpage

jonahbeckford:
No problem. I'll relook at that manpage while I'm using a MacVM machine.

---

I'm quite puzzled why this problem does not occur in my own GitHub Actions CI on macOS (ex. https://github.com/diskuv/dkml-component-ocamlrun/actions/runs/2827173045 ). I realize I do a `brew install opam` in my own CI (so I won't see any Homebrew update problems as you mentioned earlier) but what would make a user not able to write into their own brew directory? That's sounds like a [bad Homebrew installation](https://docs.brew.sh/FAQ#why-does-homebrew-say-sudo-is-bad).

Regardless, my intention is not to perturb the brew installation at all, so as long as Homebrew makes a read-only mode possible I'll try to fix it. But I'm suggesting the Opam CI setup isn't realistic for a macOS user (or I'm missing something fundamental).

kit-ty-kate:
> but what would make a user not able to write into their own brew directory?

the opam sandbox would. The sandbox only allows writes in `$TMPDIR` or the current opam switch directory during installation.

kit-ty-kate:
> But I'm suggesting the Opam CI setup isn't realistic for a macOS user (or I'm missing something fundamental).

I disagree. Myself as a Brew user would have hit those issues (I don't install packages every day)

jonahbeckford:
Root Cause: The `brew bundle dump` commands installs the `homebrew/bundle` tap if it is not already installed. In particular, Opam CI does not have the `homebrew/bundle` tap pre-installed (and it does not need to be pre-installed), while GitHub Actions CI does have it pre-installed. Logs from a MacVM are at https://gist.github.com/jonahbeckford/477b231668966ba91a1255c52dce3aaf . 

I can just install `if command -v brew; then brew tap homebrew/bundle; fi` as part of the `.opam` files with a `os = macos` condition. I feel like there is a cleaner way in Opam to express a brew dependency; is there?

Thanks!

kit-ty-kate:
> I can just install `if command -v brew; then brew tap homebrew/bundle; fi` as part of the `.opam` files with a `os = macos` condition. I feel like there is a cleaner way in Opam to express a brew dependency; is there?

Mmh, I'm not sure to understand how it would fix this issue as brew is available in both cases. Are `HOMEBREW_BUNDLE_TAP_SKIP=1` and the others environment variables not helping?

jonahbeckford:
> Mmh, I'm not sure to understand how it would fix this issue as brew is available in both cases.
> Are `HOMEBREW_BUNDLE_TAP_SKIP=1` and the others environment variables not helping?

`HOMEBREW_BUNDLE_TAP_SKIP=1` is not helping and it would be weird if that worked because doing `brew bundle dump` is explicitly telling Homebrew that I *want* the `homebrew/bundle` tap. That is sane Homebrew behavior. And I had used the other env vars; no effect (again, this makes sense).

I'll do the `brew tap homebrew/bundle` command in the opam files, and undo the environment variable changes, and see what happens.

jonahbeckford:
FYI, all of what I am doing can be avoided if somehow we pre-installed Homebrew's `bundle` tap and any other tap that is owned by Homebrew (anything in the https://github.com/Homebrew/homebrew-* namespace).

https://github.com/Homebrew/homebrew-bundle/tree/master#installation says:
> `brew bundle` is automatically installed when first run.

But because of the Opam sandbox, Homebrew is blocked from installing its own tap! I feel it would be weird if an end-user would not let Homebrew install Homebrew's own taps like `homebrew/core`, `homebrew/bundle`, etc.

I also realize the usefulness of the Opam sandbox, so I'm just venting ....

Once I've got a good solution (which involves `extra-source` because I have to work around the network firewall in the sandbox) other people can copy it.

jonahbeckford:
It's done. I'll apply the change to the other PRs as well.

packages/dkml-component-staging-opam32/dkml-component-staging-opam32.2.2.0~dkml20220801/opam
```diff
@@ -0,0 +1,365 @@
...
+    set -eufx
+    if command -v brew; then
+      eval $(brew shellenv)
+      HOMEBREW_BREW_FILE="$(command -v brew)"
+      HOMEBREW_LIBRARY="${HOMEBREW_REPOSITORY}/Library"
+      HOMEBREW_PATH="$PATH"
+      HOMEBREW_CACHE="$PWD/dl/homebrew-cache"
+      export HOMEBREW_BREW_FILE HOMEBREW_PATH HOMEBREW_CACHE
+
+      # Clone the Library which contains the Taps/
```
kit-ty-kate, line 49: ... and so on (couldn't drag my cursor down far enough to englobe everything)

Just a small note for yourself: these would probably be easier to maintain/read if it was in a separate `setup-homebrew.sh` file

kit-ty-kate:
Thanks!

PR #21749, opened 2022-07-07
Revert "Add stdcompat upper bounds"

mseri:
This reverts commit 567f637fcb2f5630df33a14f51a7bc39b5127fff.
First we need it to compare the errors and understand a few things.

PR #21537, opened 2022-06-15
[new release] stdcompat.19

thierry-martinez:
This PR publishes the new release stdcompat.19, compatible with OCaml 5.0, and mark older versions of stdcompat incompatible with OCaml 5.0. Here is stdcompat.19 change log:

- Compatibility with OCaml 5.0

- Add String.`{hash, seeded_hash}`

- Restore compatibility with OCaml 3.07, and fix order execution for
  `Set` and `Map.{iter, fold, filter_map}` on OCaml 3.07

- Updated port to `dune`
  (Marek Kubica, https://github.com/thierry-martinez/stdcompat/pull/16,
   https://github.com/thierry-martinez/stdcompat/pull/19)

- Add support for `flambda2`
  (Guillaume Bury, https://github.com/thierry-martinez/stdcompat/pull/14)

- Prevent replacing `/dev/null/ by a regular file when `./configure` is
  run as root
  (reported by Marc Chevalier, https://github.com/ocaml/ocaml/issues/11302)

PR #21753, opened 2022-07-07
Revert "[new release] stdcompat.19"

kit-ty-kate:
Reverts ocaml/opam-repository#21537

This new release seems to be having a lot of issues (see #21749). Since the release is rather recent it seems safe to revert until the issues are fixed

cc @thierry-martinez


thierry-martinez:
I agree! Sorry for all this noise and thank you for noticing this.

PR #21765, opened 2022-07-08
[new release] stdcompat.19 metapp.0.4.4 metaquot.0.5.2

thierry-martinez:
This is a new tentative for releasing stdcompat.19 (after previously
reverted attempt:
https://github.com/ocaml/opam-repository/pull/21753).

This commit ships new versions of metapp and metaquot compatible with
stdcompat.19 and add version bounds to incompatible package versions.


kit-ty-kate:
I’m trying to patch stdcompat locally but the compilation fails with
```
configure: OCaml version is 5.0.0+dev4-2022-06-14
configure: Obj.magic will be used in redefinition when possible
    (add --disable-magic for pure redefinitions).
File "dune", line 1, characters 0-2363:
  1 | (library
  2 |  (name stdcompat)
  3 |  (public_name stdcompat)
.....
106 |    stdcompat__weak
107 |    stdcompat__weak_s
108 |    stdcompat))
ld: archive has no table of contents file './libstdcompat_stubs.a' for architecture arm64
File "caml_startup", line 1:
Error: Error during linking (exit code 1)
```
on my macOS/arm64 machine with OCaml 5.0. I tried with OCaml 4.14 and it seems to work. Is that expected?

thierry-martinez:
Thank you, Kate, for your feedback! I appreciate it very much, but note
that I marked this PR as draft for you don't waste too much of your time
debugging it before I polished it! Thank you for looking at it yet.

Indeed, I didn't expect the error you encounter and I was not aware of
it because I don't work with macOS/arm64 regularly. I was able to
reproduce the failure: the failure happens when `ocamlmklib` is used to
make an empty C library (the stdcompat C library is empty on OCaml 5.0
since there is nothing more recent to declare!), and it seems that this
is a change in the behavior of `ocamlmklib` since OCaml 5.0. I am not
sure if I should and how to report this yet, because it is not really a
bug and it is quite a corner case.

I fix the error by adding a dummy symbol in the `stdcompat` C library
when compiled with OCaml 5.0.

There is still a strange error with `refl` making segfaulting the
compiler with OCaml 5.0: I put an upperbound on `refl` to exclude OCaml
5.0 but I will need to investigate this.

There is also a strange error with `ocaml-in-python` on `ppc64`, since
there is no really error message in the log...

-- 
Thierry.

mseri:
Should be fine. I have restarted some fo the cancelled tests just to double check

mseri:
Thanks and sorry for the wait

PR #21775, opened 2022-07-12
[new release] gopcaml-mode 0.0.6 and gopcaml-mode-merlin 0.0.6

Gopiandcode:
- Updates gopcaml-mode and gopcaml-mode-merlin to work with OCaml 4.14.0


kit-ty-kate:
Thanks

PR #21776, opened 2022-07-12
[new release] chamelon and chamelon-unix (0.0.10)

yomimono:
Subset of littlefs filesystem fulfilling MirageOS KV

- Project page: <a href="https://github.com/yomimono/chamelon">https://github.com/yomimono/chamelon</a>

##### CHANGES:

* detect simple cycles in the metadata tree at connect time (@yomimono)
* check block size in the superblock at connect time, and fail if it doesn't match block device (@yomimono)


yomimono:
Test failures are in the macos-homebrew target, where the provided `dd` is failing on the `bs` argument. I can fix this for future releases but would prefer if this version can be released as-is.

kit-ty-kate:
macos is an unstable CI target for now anyway so it’s ok. Thanks!

PR #21778, opened 2022-07-12
lambdapi: fix pratter constraints

fblanqui:
Lambdapi is not compatible with pratter >= 2.0.0. This PR fixes contraints on pratter for Lambdapi.
See https://github.com/Deducteam/lambdapi/issues/880 and https://github.com/Deducteam/lambdapi/pull/879.


kit-ty-kate:
Thanks

PR #21784, opened 2022-07-13
Package data-encoding.0.6

raphael-proust:
### `data-encoding.0.6`
Library of JSON and binary encoding combinators

---
* Homepage: https://gitlab.com/nomadic-labs/data-encoding
* Source repo: git+https://gitlab.com/nomadic-labs/data-encoding.git
* Bug tracker: https://gitlab.com/nomadic-labs/data-encoding/issues

---

### DIFF
--- a/packages/data-encoding/data-encoding.0.6/opam
+++ b/packages/data-encoding/data-encoding.0.6/opam
@@ -1,0 +1,39 @@
+opam-version: "2.0"
+maintainer: "contact@nomadic-labs.com"
+authors: [ "Nomadic Labs" ]
+homepage: "https://gitlab.com/nomadic-labs/data-encoding"
+bug-reports: "https://gitlab.com/nomadic-labs/data-encoding/issues"
+dev-repo: "git+https://gitlab.com/nomadic-labs/data-encoding.git"
+license: "MIT"
+depends: [
+  "ocaml" { >= "4.10" }
+  "dune" { >= "2.0" }
+  "ezjsonm"
+  "zarith" {>= "1.4"}
+  "zarith_stubs_js"
+  "hex" {>= "1.3.0"}
+  "json-data-encoding" { = "0.11" }
+  "json-data-encoding-bson" { = "0.11" }
+  "alcotest" { with-test }
+  "crowbar" { >= "0.2" & with-test }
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
> * The `md2mld` package should only be a documentation dependency. Currently it is set as a test dependency to circumvent

I’m not sure to understand, why is it building the documentation during normal compilation of the package?

raphael-proust:
I guess I don't know `dune` well enough. It turns out that the `documentation` stanza causes the mld files to be installed. In my case the mld file is generated.

I'll change the build or something.

raphael-proust:
Fixed.

In case it is useful for someone else:
I used the `(mode promote)` of `dune` and committed the generated file into the repository. My use-case (avoiding additional dependencies at release time) is given as a typical use-case for `(mode promote)` in the dune documentation. https://dune.readthedocs.io/en/stable/dune-files.html?highlight=promote#modes

packages/data-encoding/data-encoding.0.6/opam
```diff
@@ -0,0 +1,39 @@
...
+  "json-data-encoding" { = "0.11" }
+  "json-data-encoding-bson" { = "0.11" }
+  "alcotest" { with-test }
+  "crowbar" { >= "0.2" & with-test }
+  "ppx_expect" { with-test }
+  "either"
+  "ppx_hash"
+  "ocamlformat" { = "0.20.1" & with-doc } # not technically a doc dep; modify when with-dev becomes available
+  "odoc" { with-doc }
+  "md2mld" { with-test } # not technically a test dep; modify when https://github.com/ocurrent/ocaml-ci/issues/264 is fixed
```
kit-ty-kate, line 24: I’m guessing those depedendencies aren’t actually required outside of a personal dev setup?
```suggestion
  "odoc" { with-doc }
```

raphael-proust: Is that what `with-tools` is for? From the opam doc

>     `with-test`: only true if tests have been enabled for this specific package
>     `with-doc`: similarly for documentation
>     `with-tools`: similarly for developer tools

That's very terse documentation so I'm not 100% sure. Not sure what it means to _enable_ developer tools.

PR #21785, opened 2022-07-13
Add patches to ocaml-base-compiler 3.09.1 to fix compilation with gcc 11.2

AltGr:
Original patch by @damiendoligez.
I had no issue with ocaml>=4.10 though, although the indentation fix is much more recent than that. Probably the cflags have been tweaked to avoid treating it as an error ?


dra27:
Cross-reference #21630, although it's a different warning being triggered

PR #21786, opened 2022-07-13
Add patches to ocaml-base-compiler 3.09.1 to fix compilation with gcc 11.2

AltGr:
Reopening as #21785 was inadvertly closed


AltGr:
CI doesn't like compiler packages, but relevant parts of it are OK and this seems good to go.

Octachron:
The patch looks fine to me (however this concerns 4.09.1 and not 3.09.1).

kit-ty-kate:
Thanks!

PR #21781, opened 2022-07-12
releases BAP 2.5.0

ivg:
### Features                     

- BinaryAnalysisPlatform/bap#1390
- BinaryAnalysisPlatform/bap#1389
- BinaryAnalysisPlatform/bap#1394
- BinaryAnalysisPlatform/bap#1395
- BinaryAnalysisPlatform/bap#1408
- BinaryAnalysisPlatform/bap#1409
- BinaryAnalysisPlatform/bap#1410
- BinaryAnalysisPlatform/bap#1428
- BinaryAnalysisPlatform/bap#1429
- BinaryAnalysisPlatform/bap#1434
- BinaryAnalysisPlatform/bap#1444
- BinaryAnalysisPlatform/bap#1445
- BinaryAnalysisPlatform/bap#1446
- BinaryAnalysisPlatform/bap#1452
- BinaryAnalysisPlatform/bap#1457
- BinaryAnalysisPlatform/bap#1461
- BinaryAnalysisPlatform/bap#1464
- BinaryAnalysisPlatform/bap#1460
- BinaryAnalysisPlatform/bap#1466
- BinaryAnalysisPlatform/bap#1469
- BinaryAnalysisPlatform/bap#1458
- BinaryAnalysisPlatform/bap#1473
- BinaryAnalysisPlatform/bap#1476
- BinaryAnalysisPlatform/bap#1479
- BinaryAnalysisPlatform/bap#1482
- BinaryAnalysisPlatform/bap#1481
- BinaryAnalysisPlatform/bap#1483
- BinaryAnalysisPlatform/bap#1488
- BinaryAnalysisPlatform/bap#1489
- BinaryAnalysisPlatform/bap#1490
- BinaryAnalysisPlatform/bap#1492
- BinaryAnalysisPlatform/bap#1493
- BinaryAnalysisPlatform/bap#1491
- BinaryAnalysisPlatform/bap#1497
- BinaryAnalysisPlatform/bap#1498
- BinaryAnalysisPlatform/bap#1499
- BinaryAnalysisPlatform/bap#1500
- BinaryAnalysisPlatform/bap#1503
- BinaryAnalysisPlatform/bap#1504
- BinaryAnalysisPlatform/bap#1505
- BinaryAnalysisPlatform/bap#1511
- BinaryAnalysisPlatform/bap#1513
- BinaryAnalysisPlatform/bap#1515
- BinaryAnalysisPlatform/bap#1516
- BinaryAnalysisPlatform/bap#1517
[...]

PR #21791, opened 2022-07-13
adds to uuidm a lower bound constraint on the cmdliner version

ivg:
The uuidm.0.9.7 version is missing a conflict that defines the lower bound constraint on the cmdliner version, which results in the following error,

```
File "test/uuidtrip.ml", line 85, characters 8-13:
85 |   Term.(const gen $ version $ ns $ name_ $ upper $ binary),
```

when uuidm.0.9.7 [is built][1] with cmdliner.0.9.4.

Discovered in PR #21781, fixed in a separate PR for cleaner history.

CC Daniel @dbuenzli for the situational awareness.

[1]: https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/bb37510c2bf1d1e78cfa6c25a14fe0c9627562f8/variant/opam-2.1,compilers,4.08,bap-core-theory.2.5.0,lower-bounds

### DIFF
--- a/packages/uuidm/uuidm.0.9.7/opam
+++ b/packages/uuidm/uuidm.0.9.7/opam
@@ -13,6 +13,9 @@ depends: [
   "ocamlbuild" {build}
   "topkg" {build} ]
 depopts: [ "cmdliner" ]
+conflicts: [
+  "cmdliner" {< "0.9.8"}
+]
 build:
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
It’s already been added in #21790

ivg:
That was a pretty synchronous finish))

PR #21795, opened 2022-07-13
Add patch to ocamlfind 1.9.5 for installation without coreutils

AltGr:
Not sure if we want to patch it this way or wait for a proper release, but although the failure is on a corner case it does block the whole ecosystem.
So I thought I would propose this PR anyway.
Sad to have to trigger so many recompilations while this only patches the install without changing the code, though, but well.


AltGr:
Anyone ? https://github.com/ocaml/ocamlfind/pull/54 is still not merged so it might be a while before there is an upstream release fixing this.

kit-ty-kate:
yolo. thanks, sorry for the wait

PR #21821, opened 2022-07-15
removes a spurious dependency on `str` from the bap-x86 package

ivg:
There is no such package as `str`, not sure how it managed to slip in and pass through the CI, probably some release script error on our side. Obviously, it prevents us from installing bap.2.5.0 with
```
$ opam install bap.2.5.0
[ERROR] Package conflict!
  * Missing dependency:
    - str
    unknown package
```

### DIFF
--- a/packages/bap-x86/bap-x86.2.5.0/opam
+++ b/packages/bap-x86/bap-x86.2.5.0/opam
@@ -38,7 +38,6 @@ depends: [
   "bap-primus" {= "2.5.0"}
   "bap-main" {= "2.5.0"}
   "bap-std" {= "2.5.0"}
-  "str"
 ]
 synopsis: "BAP x86 lifter"
 


ivg:
In fact, it wasn't a script or anything, it was my [deliberate action][1] :( Not sure what I was thinking of...

All the CI failures on this are either from the distributions that we do not support (mostly missing llvm) or from 32-bit systems, this one will be fixed in the next release of BAP. 

I would really appreciate if this PR will be merged ASAP, as it blocks us and we can't make a press release)

[1]: https://github.com/ocaml/opam-repository/pull/21781/commits/2ff6a0ffee4fe9bbf5df2ccbabaa83ae0d3d7f30

kit-ty-kate:
no worries, I’m just waiting for the revdeps to finish and it should be good to go

ivg:
Looks like that the revdeps finished (with revdep bap finished successfully, which is a good indicator). Could you please merge it?

PR #21822, opened 2022-07-15
Update charset constraints to support new ocaml_intrinsics release.

yallop:


dinosaure:
Seems ok for me but I would like an expert review :+1: (/cc @kit-ty-kate)

packages/charset/charset.0.1.0/opam
```diff
@@ -13,7 +13,7 @@ build: [
 depends: [
   "ocaml" {>= "4.11.0"}
   "dune" {>= "2.3"}
-  "ocaml_intrinsics" {os = "linux" & arch = "x86_64"}
+  "ocaml_intrinsics" {>= "v0.15.1" & os != "win32"}
```
mseri, line 16:
```suggestion
  "ocaml_intrinsics" {>= "v0.15.0" & os != "win32"}
```
otherwise it is not equivalent to the previous version

yallop: It's not intended to be equivalent; it's supposed to enable intrinsics support to `charset` on macos now that `ocaml_intrinsics` supports macos.

[`ocaml_intrinsics.v0.15.1`](https://github.com/ocaml/opam-repository/blob/master/packages/ocaml_intrinsics/ocaml_intrinsics.v0.15.1/opam) supports `macos`, but [`ocaml_intrinsics.v0.15.0`](https://github.com/ocaml/opam-repository/blob/master/packages/ocaml_intrinsics/ocaml_intrinsics.v0.15.1/opam) does not, hence the version constraint in this PR.

mseri: Yes, but opam will deal with that depending on the system. So will automatically select 0.15.1 as lower bound on macos

mseri: If you look at the CI, it is working fine on macos

PR #21830, opened 2022-07-18
[new release] ocamlformat and ocamlformat-rpc-lib (0.24.1)

gpetiot:
Auto-formatter for OCaml code

- Project page: <a href="https://github.com/ocaml-ppx/ocamlformat">https://github.com/ocaml-ppx/ocamlformat</a>

### New features

- Support `odoc-parser.2.0.0` (ocaml-ppx/ocamlformat#2123, @gpetiot)
  * Breaking change: incompatible with earlier versions of `odoc-parser`
  * New inline math elements `{m ...}` available in doc-comments
  * New block math elements `{math ...}` available in doc-comments

### DIFF
--- a/packages/ocamlformat/ocamlformat.0.24.1/opam
+++ b/packages/ocamlformat/ocamlformat.0.24.1/opam
@@ -1,0 +1,57 @@
+opam-version: "2.0"
+synopsis: "Auto-formatter for OCaml code"
+description:
+  "OCamlFormat is a tool to automatically format OCaml code in a uniform style."
+maintainer: ["OCamlFormat Team <ocamlformat-dev@lists.ocaml.org>"]
+authors: ["Josh Berdine <jjb@fb.com>"]
+homepage: "https://github.com/ocaml-ppx/ocamlformat"
+bug-reports: "https://github.com/ocaml-ppx/ocamlformat/issues"
+depends: [
+  "ocaml" {>= "4.08"}
+  "alcotest" {with-test}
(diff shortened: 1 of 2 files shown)


packages/ocamlformat/ocamlformat.0.24.1/opam
```diff
@@ -0,0 +1,57 @@
...
+  "dune" {>= "2.8"}
+  "dune-build-info"
+  "either"
+  "fix"
+  "fpath"
+  "menhir" {>= "20201216"}
+  "menhirLib" {>= "20201216"}
+  "menhirSdk" {>= "20201216"}
+  "ocaml-version" {>= "3.3.0"}
+  "ocamlformat-rpc-lib" {with-test & post & = version}
```
kit-ty-kate, line 23: the `{post}` flag here seems suspicious

gpetiot: It's been the case for a few versions now, should we just get rid of it for the next ones?

kit-ty-kate: what’s its purpose?

gpetiot: We don't have post deps so I guess it's useless, I will remove this tag.

kit-ty-kate:

gpetiot:
> Thanks! Could you return the fix upstream?

Here it is: https://github.com/ocaml-ppx/ocamlformat/pull/2130

PR #21834, opened 2022-07-18
post-messages for OCaml pre-release packages

dra27:
This PR seeks both to thank users for trying out OCaml pre-releases and also to advertise more loudly that alpha releases may require the addition of a testing repository.

The intention is that the message about the alpha repository would disappear either with the beta or rc release. At the actual release, the message for the trunk package would change to reflect the fact it's no longer a pre-release.

It is possible to include a package in opam-alpha-repository to customise the message based on whether the repository has been enabled or not (using exactly the same mechanism as we do for ocaml-beta). This adds a small bit of complexity (one extra package appears in the switch) but it does have the benefit that we could _warn_ users to **remove** opam-alpha-repository when they test beta and rc releases, etc. We could even go further and _prevent_ the use of opam-alpha-repository with a released compiler (if we wanted).

This PR also proposes removing the parallelism failure message, at least from pre-releases, where that should be reported as a bug and arguably from all new releases, as it's a bug which should be reported and fixed.

Note that changing post-messages doesn't trigger a switch rebuild.

### DIFF
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.5.0.0~alpha0/opam
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.5.0.0~alpha0/opam
@@ -36,3 +36,18 @@ url {
   checksum: "sha256=70f5353001ba95dff96f71680a435e1e7d3708eddb1fd8c74b9e18a292fe40ce"
 }
 extra-files: ["ocaml-base-compiler.install" "md5=3e969b841df1f51ca448e6e6295cb451"]
+post-messages: [
+  "Thank you for testing OCaml 5.0!
+
+   During the release cycle, there can be a lag between packages being fixed for
+   compatibility with the latest release of OCaml and an actual package-release
+   being made to opam-repository. You may wish to add opam-alpha-repository to
+   your switch by issuing:
+
+   opam repo add alpha git+https://github.com/ocaml/opam-alpha-repository
+
+   This will enable patched, but unreleased, versions of some packages which are
+   used for automated testing and made available to aid development and testing of
+   the new release. These packages are automatically upgraded when a new upstream
+   release is made to the official opam-repository."
+]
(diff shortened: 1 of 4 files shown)


kit-ty-kate:
LGTM. One question before merging: are you sure the line breaks are taken into account with the simple double quotes `"` (instead of `"""`)?

dra27:
We need to move opam-alpha-repository (or update the URL in this one) also before merging (sorry, I should have said that in the PR!)

Yeah - the triple quoted version just allows you to use `"` marks unquoted in the text, it doesn't affect newlines.

packages/ocaml-base-compiler/ocaml-base-compiler.5.0.0~alpha0/opam
```diff
@@ -37,11 +37,17 @@ url {
...
-   to force a sequential build instead."
-  {failure & jobs > 1 & opam-version >= "2.0.5"}
+  "Thank you for testing OCaml 5.0!
+
+   During the release cycle, there can be a lag between packages being fixed for
+   compatibility with the latest release of OCaml and an actual package-release
+   being made to opam-repository. You may wish to add opam-alpha-repository to
+   your switch by issuing:
+
+   opam repo add alpha git+https://github.com/ocaml/opam-alpha-repository
```
mseri, line 47: The url is wrong. The alpha repository is at: https://github.com/kit-ty-kate/opam-alpha-repository
Maybe it should be moved under the ocaml umbrella though

mseri: Same comment for all the other opam files

dra27: That was the intention - sorry, I should have been explicit about that in the PR's description!

mseri: Now that the beta is out and the official guidelines ask to use @kit-ty-kate alpha repo, I think it may be worth just creating its doppleganger in the ocaml org and make this PR mergeable. What do you think?

kit-ty-kate:
This PR isn't relevant anymore since the release of 5.0.0. We could discuss it again for 5.1 but I don't think the opam-alpha-repository will be required the same way it was during the 5.0 alpha cycle.

dra27:
Agreed - let's remember to bring this up when 5.1 is branched in April (it might be worth having a post-messages for the releases regardless, even if it doesn't need to advertise the alpha repo)

PR #21841, opened 2022-07-19
Package liquidsoap.2.1.0-1

toots:
We have encountered minor issues with the original release:
* Build would fail in some situations when `pandoc` binary is in the `PATH` as this triggered a broken doc build
* `mem_usage` optional features were not being compiled.

This release fixes both. Since the `2.1.0` release is not installable in some instance, it is also marked as not available.

### DIFF
--- a/packages/liquidsoap/liquidsoap.2.1.0-1/opam
+++ b/packages/liquidsoap/liquidsoap.2.1.0-1/opam
@@ -1,4 +1,3 @@
-available: false
 opam-version: "2.0"
 synopsis: "Swiss-army knife for multimedia streaming"
 description: """\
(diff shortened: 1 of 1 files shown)


packages/liquidsoap/liquidsoap.2.1.0-1/opam
```diff
@@ -0,0 +1,186 @@
...
+together."""
+maintainer: "romain.beauxis@gmail.com"
+authors: "The Savonet Team <savonet-users@lists.sourceforge.net>"
+license: "GPL-2.0-or-later"
+homepage: "https://github.com/savonet/liquidsoap"
+bug-reports: "https://github.com/savonet/liquidsoap/issues"
+depends: [
+  "conf-autoconf" {dev}
+  "conf-automake" {dev}
+  "ocaml" {>= "4.12.0" & <= "5.0.0"}
```
kit-ty-kate, line 18:
```suggestion
  "ocaml" {>= "4.12.0" & < "5.0.0"}
```

PR #21842, opened 2022-07-20
ndbm is already present in the base system on Macs

UnixJunkie:
No need to install gdbm.
Tested on a Mac M1.


kit-ty-kate:
Thanks

PR #21844, opened 2022-07-20
[new release] links, links-postgresql, links-sqlite3 and links-mysql (0.9.7)

dhil:

##### CHANGES:

This release includes minor bug fixes, improvements, and a breaking
change.

* **Breaking change**: The Links standard library has adopted `Maybe`
  as it optional type. Previously, the standard library made
  inconsistent use of both `Option` and `Maybe` types. As a
  consequence of this change, every instance of `Option` has been
  turned into a `Maybe` type.
* Uniform paths: The file paths used by Links internally are now
  uniformly computed at compile time from the base of the current OPAM
  switch.
* Support for record extension in the presence of temporal projections.
* It is now to use `default` as a setting value in the REPL.
* Fixed an assertion failure in the relational lens code base.
* Effect aliases: It is now possible to define aliases for effect
  signatures, e.g.

```links
effectname State(s, e::Eff) = {Get:s, Put:(s) -> () |e};
```

defines an alias named `State` for an effect row with two operations
`Get`, and `Put`. This alias is analogous to aliases introduced via
`typename`, except that `effectname` aliases may be used in places
where an effect row is expected, e.g.

```links
sig f : () ~State(Int, {})~> ()
```

is equivalent to

```links
sig f : () {Get:Int, Put:(Int) -> ()}~> ()
```

* Client/server annotations are now permitted on local and anonymous
  functions.
* Fixed a bug where remote calls involving a client closure would
  cause the server side to crash.

### DIFF
--- a/packages/links-mysql/links-mysql.0.9.7/opam
+++ b/packages/links-mysql/links-mysql.0.9.7/opam
@@ -1,0 +1,31 @@
+opam-version: "2.0"
+maintainer: "James Cheney <jcheney@inf.ed.ac.uk>"
+authors: "The Links Team <links-dev@inf.ed.ac.uk>"
+synopsis: "MySQL database driver for the Links Programming Language"
+description: "MySQL database driver for the Links Programming Language"
+homepage: "https://github.com/links-lang/links"
+dev-repo: "git+https://github.com/links-lang/links.git"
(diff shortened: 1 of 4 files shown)


kit-ty-kate:
Thanks

PR #21890, opened 2022-07-27
Set `--docdir` in ocaml-base-compiler.4.14.0

dra27:
Unintended difference between the rc2 and release packages. The 5.0.0 alpha packages are still setting `--docdir` correctly, so we'll just be careful that this doesn't happen with that release package too.

### DIFF
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.4.14.0/opam
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.4.14.0/opam
@@ -20,6 +20,7 @@ build: [
   [
     "./configure"
     "--prefix=%{prefix}%"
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
Could you add it to `ocaml-variants.4.14.0+options` as well ?

dra27:
Oops - I was rushing... fix pushed for the +options package too!

kit-ty-kate:
Thanks

dbuenzli:
I missed that ! Thank you @dra27 you can't imagine the sheer joy of being able to do `odig changes ocaml` (even more so now that `ocaml.org` became entirely unusable as far as tracking when a changed occured is concerned).

PR #21905, opened 2022-07-27
Package seq.0.3.1

c-cube:
### `seq.0.3.1`
Compatibility package for OCaml's standard iterator type starting from 4.07

---
* Homepage: https://github.com/c-cube/seq/
* Source repo: git+https://github.com/c-cube/seq.git
* Bug tracker: https://github.com/c-cube/seq/issues

---

### DIFF
--- a/packages/seq/seq.0.3.1/opam
+++ b/packages/seq/seq.0.3.1/opam
@@ -1,0 +1,24 @@
+opam-version: "2.0"
+synopsis:
+  "Compatibility package for OCaml's standard iterator type starting from 4.07"
+maintainer: "simon.cruanes.2007@m4x.org"
+license:  "LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception"
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+]
+depends: [
+  "dune" {>= "2.5"}
+  "ocaml"
+]
+tags: [ "iterator" "seq" "pure" "list" "compatibility" "cascade" ]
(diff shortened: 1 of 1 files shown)


sim642:
Since this is in response to a regression (https://github.com/c-cube/seq/issues/14), I wonder if for good measure seq 0.3 should get a conflicts clause with OCaml < 4.07 or the avoid-version flag, since it broke the sole feature it aims to deliver?

c-cube:
just to clarify, it only breaks in conjunction with `-linkall`, correct? Or did I misunderstand that?

sim642:
Yes, but that can show up in surprising places like dune's mdx. I think the consequences of that aren't visible from opam-health-check because that doesn't run tests (during which dune runs mdx).

c-cube:
Can it be another PR and not delay this one? :/

kit-ty-kate:
Thanks

kit-ty-kate:
> Since this is in response to a regression ([c-cube/seq#14](https://github.com/c-cube/seq/issues/14)), I wonder if for good measure seq 0.3 should get a conflicts clause with OCaml < 4.07 or the avoid-version flag, since it broke the sole feature it aims to deliver?

Since 0.3 is only a week old it should be fine to do i suppose (older it would have a greater chance to break users using it in lock files). Feel free to open a PR

c-cube:
https://github.com/ocaml/opam-repository/pull/21913 @sim642  would that do?

PR #21912, opened 2022-07-29
[new release] chamelon and chamelon-unix (0.1.1)

yomimono:
CHANGES:

* bugfix: remove dependency cycle between chamelon and chamelon-unix
* bugfix: tag src/runtest with `(package chamelon-unix)`


kit-ty-kate:
The tests fail on macOS with:
```
# (cd _build/default/mirage_test && /bin/dd if=/dev/zero of=emptyfile bs=4K count=10)
# dd: bs: illegal numeric value
```
Do you want to fix it or disable tests on macOS?

kit-ty-kate:
Actually i can’t reproduce this locally on the latest version of macOS (12.5) so let’s just ignore it.

kit-ty-kate:
Thanks!

PR #21913, opened 2022-07-29
make seq.0.3 incompatible with ocaml < 4.07

c-cube:


packages/seq/seq.0.3/opam
```diff
@@ -10,6 +10,9 @@ depends: [
   "dune" {>= "2.5"}
   "ocaml"
 ]
+conflicts: [
+  "ocaml" { < "4.07.0" }
+]
```
kit-ty-kate, line 15:
```suggestion
  "ocaml" {>= "4.07.0"}
]
```

kit-ty-kate:
Thanks

PR #21923, opened 2022-07-31
Package pyml_bindgen.0.4.0

mooreryan:
### `pyml_bindgen.0.4.0`

---

---


mooreryan:
Regarding the fedora-35 fail...I think it may be related to the fact that I was missing some deps in the cram stanza of the tests (eg, `%{bin:gen_multi}` was missing, and the test on the fedora-35 was saying that `gen_multi`, which is from this package too, was missing).  If you think this is it, I will close and resubmit an updated version with that fixed.

Regarding the x86_32 tests, I have no idea what the problem is in that case....

kit-ty-kate:
It feels like it’s indeed related to a missing dependency in your tests (and it only fails on one of them just by luck, as gen_multi is probably being compiled at the same time).

For x86_32 i think this is a bug in the compiler: https://github.com/ocaml/ocaml/pull/10835. It will be fixed in 4.14.1 and 5.0 so we can just ignore it for now. It’s not a blocker.

mooreryan:
Thanks, I will close this and fix the issue.

PR #21948, opened 2022-08-05
Package timed.1.1

rlepigre:
### `timed.1.1`
Timed references for imperative state
Timed references for imperative state. This module provides an alternative
   type for references (or mutable cells) supporting undo/redo operations. In
   particular, an abstract notion of time is used to capture the state of the
   references at any given point, so that it can be restored. Note that usual
   reference operations only have a constant time / memory overhead (compared
   to those of the standard library).

   Moreover, we provide an alternative implementation based on the references
   of the standard library (Pervasives module). However, it is less efficient
   than the first one.

---
* Homepage: https://github.com/rlepigre/ocaml-timed
* Source repo: git+https://github.com/rlepigre/ocaml-timed.git
* Bug tracker: https://github.com/rlepigre/ocaml-timed/issues

---


kit-ty-kate:
Thanks!

PR #21878, opened 2022-07-26
Just a minor update, to compile "ocaml-sfml" with ocaml 4.14.0

fccm:

PR #21949, opened 2022-08-06
conf-sfml2: Add support for most distributions

kit-ty-kate:
To help merge https://github.com/ocaml/opam-repository/pull/21878

### DIFF
--- a/packages/conf-sfml2/conf-sfml2.1/opam
+++ b/packages/conf-sfml2/conf-sfml2.1/opam
@@ -6,7 +6,15 @@ license: "Zlib"
 build: [["pkg-config" "sfml-all"]]
 depexts: [
   ["libsfml-dev"] {os-family = "debian"}
+  ["libsfml-dev"] {os-family = "ubuntu"}
+  ["sfml"] {os-family = "arch"}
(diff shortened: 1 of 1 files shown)


fccm:
On a ChromeBook, the command `uname -a` tells me that the Linux provided is called "penguin".
The command used to install something is an "apt-get" and the name of the package is the same than on Debian:
`"libsfml-dev"`
I don't know if it should be included with `os-family` of `os-distribution`.

kit-ty-kate:
what does `opam config report` give you?

fccm:
```
# opam config report
# opam-version      2.0.3 
# self-upgrade      no
# system            arch=x86_64 os=linux os-distribution=debian os-version=10
# solver            mccs
# install-criteria  -removed,-count[version-lag:,true],-changed,-count[version-lag:,false],-new
# upgrade-criteria  -removed,-count[version-lag:,false],-new
# jobs              1
# repositories      1 (http) (default repo at e3a26c21)
# pinned            0
# current-switch    4.14.0
```

kit-ty-kate:
Then no, it will be correctly detected as a Debian-based distribution.

PR #21744, opened 2022-07-07
new file:   packages/nlopt/nlopt.0.0.6.1/opam

UnixJunkie:

PR #21955, opened 2022-08-08
Nlopt.0.6.1

UnixJunkie:


UnixJunkie:
Cc @mkur.

UnixJunkie:
The previous package version was wrong (extra leading 0.).
The tarball of the package is now hosted by the package creator since a PR was merged (we can now opam pin the official repository).
Sorry for any misunderstanding.
And indeed, there was some package renaming (not due to me): previously this package was nlopt-ocaml, but it became nlopt recently.

UnixJunkie:
related to https://github.com/ocaml/opam-repository/pull/21744

kit-ty-kate:
Looks good. Could you add a `nlopt-ocaml.transition` package (similar to `jbuilder.transition`) so that the move from the `nlopt-ocaml` and `nlopt` package can be seamless for users?

UnixJunkie:
Ok, I'll have a look at that; probably today.

UnixJunkie:
Ok, I added a transition package in there.
If someone can read its opam file, that would be nice.
Especially, do we really need a depends on OCaml line?

packages/nlopt-ocaml/nlopt-ocaml.transition/opam
```diff
@@ -0,0 +1,14 @@
+opam-version: "2.0"
+maintainer: "Francois Berenger"
+authors: ["Francois Berenger"]
+homepage: "https://github.com/ocaml/dune"
+bug-reports: "https://github.com/ocaml/dune/issues"
+dev-repo: "git+https://github.com/ocaml/dune.git"
+license: "MIT"
+depends: [ "ocaml" ]
```
mseri, line 8:
```suggestion
depends: [
  "ocaml"
  "nlopt"
]
```

mseri:
Thanks!

UnixJunkie:
Thanks for maintaining opam-repository, you are both doing a great job.

mseri:
Thanks!

PR #21979, opened 2022-08-13
Move dkml-compiler from tag tarball to release

jonahbeckford:
My stupidity. I deleted the dkml-compiler tag, realized my mistake, and then re-applied the tag. But GitHub tag archives don't seem to be reproducible after some time elapses.

This PR uses a GitHub release to store the archive. I recovered the original tarball from an Opam download-cache.


kit-ty-kate:
Thanks!

PR #21982, opened 2022-08-14
Remove conf-libsamplerate in favor of conf-samplerate.

toots:
`conf-samplerate` and `conf-libsamplerate` serve the same purpose!


kit-ty-kate:
Removing the package outright might break some users. Could you instead do as in https://github.com/ocaml/opam-repository/pull/18145 and create a new `conf-libsamplerate.2` package that aliases to `conf-samplerate`?

Also note that the two packages have different definitions. In particular nixos and alpine (os-family vs. os-distribution) are different. Could you move the relevant changes to conf-samplerate (especially the latter)?

toots:
@kit-ty-kate @mseri PR review suggestions applied!

kit-ty-kate:
Thanks a lot!

PR #21985, opened 2022-08-15
Remove constraints on opam < 2.0

dra27:
While working on some upstreaming:
- All the files in the repository are headed `opam-version: "2.0"` and can't be lexed by OPAM 1.x, so remove `available` constraints when less than 2.0
- Dune 1.x and 2.x contain a workaround for OPAM 1.x which can consequently never be evaluated
- #12457 is now equivalent to `available: false`


kit-ty-kate:
Thanks!

PR #21987, opened 2022-08-16
Split conf-srt into conf-srt-openssl and conf-srt-gnutls when relevant.

toots:
The depext for `conf-srt` have some issues due to the fact that some OS/distributions are specializing them between `openssl` and `gnutls`:
* `homebrew` expects `openssl` but does not install it with `srt` (seems like a bug frankly)
* `debian` has been specializing into `gnutls`/`openssl` for a while
* `ubuntu` has started picking up those changes since the `22.04`/`jammy` LTS release

Thus, this PR bumps `conf-srt` and adds `conf-srt-openssl` and `conf-srt-gnutls` and adds relevant dependencies.

This will break compatibility with older versions of ubuntu but I'm not sure if there is a better way to do and this is clearly the solution moving forward.


toots:
The build failure on debian is a fixed bug: https://bugs.debian.org/1001529

mseri:
Thanks and sorry for the wait

PR #21988, opened 2022-08-16
Update alt-ergo's version with specific patch

Stevendeo:
Just a simple patch for updating alt-ergo to display the correct version number at calls to `--version`.


kit-ty-kate:
Thanks

PR #21996, opened 2022-08-17
catala.0.7.0: remvove references to dev variables

AltGr:
There is apparently a bad interaction with opam's current --strict option. These variables are used as a workaround for the
lack of the forthcoming --for-dev flag. We'll need to remember to filter out these variables from the in-source opam file when publishing the next versions of the package.


AltGr:
it's weird that it's `--strict` and not `opam lint` that is failing on this though. Probably worth having a look into how it's done in opam.

kit-ty-kate:
> There is apparently a bad interaction with opam's current --strict option

I find this rather weird. We're also using `--strict` with opam 2.0, 2.1 and 2.2 in CI on the whole repository and never had any issue about those variables

na4zagin3:
Thank you all for addressing the issue!

PR #21998, opened 2022-08-18
[new release] goblint (2.0.0)

sim642:
Static analysis framework for C

- Project page: <a href="https://goblint.in.tum.de">https://goblint.in.tum.de</a>
- Documentation: <a href="https://goblint.readthedocs.io/en/latest/">https://goblint.readthedocs.io/en/latest/</a>

##### CHANGES:

Goblint "lean" release after a lot of cleanup.

* Remove unmaintained analyses: OSEK, ARINC, shapes, containment, deadlocksByRaces (goblint/analyzer#460, goblint/analyzer#736, goblint/analyzer#812, goblint/analyzer#474).
* Add interactive analysis (goblint/analyzer#391).
* Add server mode (goblint/analyzer#522).
* Add Compilation Database support (goblint/analyzer#406, goblint/analyzer#448).
* Add floating point domain, unrolled array domain and improved struct domains (goblint/analyzer#734, goblint/analyzer#761, goblint/analyzer#577, goblint/analyzer#419).
* Add static loop unrolling and heap variable unrolling (goblint/analyzer#563, goblint/analyzer#722).
* Improve race detection with may-happen-in-parallel analysis (goblint/analyzer#529, goblint/analyzer#518, goblint/analyzer#595).
* Reimplement lockset and deadlock analyses (goblint/analyzer#659, goblint/analyzer#662, goblint/analyzer#650, goblint/analyzer#655).
* Add pthread extraction to Promela (goblint/analyzer#220).
* Add location spans to output (goblint/analyzer#428, goblint/analyzer#449).
* Improve race reporting (goblint/analyzer#550, goblint/analyzer#551).
* Improve dead code reporting (goblint/analyzer#94, goblint/analyzer#353, goblint/analyzer#785).
* Refactor warnings (goblint/analyzer#55, goblint/analyzer#783).
* Add JSON schema for configuration (goblint/analyzer#476, goblint/analyzer#499).
* Refactor option names (goblint/analyzer#28, goblint/analyzer#192, goblint/analyzer#516, goblint/analyzer#675).
* Add bash completion (goblint/analyzer#669).
* Add OCaml 4.13 and 4.14 support, remove OCaml 4.09 support (goblint/analyzer#503, goblint/analyzer#672).

PR #22003, opened 2022-08-19
conf-ruby: add rubygems depext on oraclelinux

sim642:
In https://github.com/ocaml/opam-repository/pull/21998#issuecomment-1219594066 it turned out that ruby doesn't work on oraclelinux as installed by conf-ruby.

The problem can be reproduced directly in `docker run -it --rm ocaml/opam:oraclelinux-8-ocaml-4.14`:
1. Install ruby as per conf-ruby: `sudo yum install -y ruby`.
2. conf-ruby build command works fine: `ruby --version`
3. `ruby` itself crashes with
    ```
    Traceback (most recent call last):
    	1: from <internal:gem_prelude>:2:in `<internal:gem_prelude>'
    <internal:gem_prelude>:2:in `require': cannot load such file -- rubygems.rb (LoadError)
    ```
4. Install rubygems also: `sudo yum install -y rubygems`
5. Now `ruby` at least starts without crashing.

centos and fedora don't have this problem: there `ruby` seems to have `rubygems` as a dependency. I guess this is some sort of oraclelinux screw-up that it doesn't.

### DIFF
--- a/packages/conf-ruby/conf-ruby.1.0.0/opam
+++ b/packages/conf-ruby/conf-ruby.1.0.0/opam
@@ -19,7 +19,7 @@ depexts: [
   ["ruby"] {os-distribution = "nixos"}
   ["ruby"] {os = "openbsd"}
   ["ruby"] {os-family = "suse"}
-  ["ruby"] {os-distribution = "ol"}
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Thanks!

PR #22004, opened 2022-08-19
[new release] 0install, 0install-solver and 0install-gtk (2.18)

talex5:
Decentralised installation system

- Project page: <a href="https://0install.net/">https://0install.net/</a>
- Documentation: <a href="https://0install.github.io/0install/">https://0install.github.io/0install/</a>

##### CHANGES:

Changes:

- Add support for ARM64 (Bastian Eicher).

- Add `+` to list of allowed characters in "extract" (Bastian Eicher, 0install/0install#185).

Bug fixes:

- Don't abort if chmod of launcher directory fails.
  We try to make the directory read-only, but this isn't very important.
  It can sometimes fail if running inside a sandbox that prevents modifications to the filesystem.

- Handle implementations that directly conflict with themselves (0install/0install#180).

- Improve SIGPIPE handling (0install/0install#162).
  Disable the SIGPIPE signal, so that we handle it as an exception instead, not by killing the process.

Updates for upstream changes:

- Make compatible with Yojson `json` type deprecation (Marek Kubica).
  Yojson 2.0 removes the `json` type, so this code switches to using `t`.

- Replace uses of deprecated `Stdlib.Stream` and `Lwt_unix.yield`.

- Fix new compiler warnings from OCaml 4.12 and 4.13.

- Update .NET Framework detection logic (Bastian Eicher, 0install/0install#178).

- Simplify GUI build using dune's new plugin support (0install/0install#161).

- Move feed to apps.0install.net (Bastian Eicher).

### DIFF
--- a/packages/0install-gtk/0install-gtk.2.18/opam
+++ b/packages/0install-gtk/0install-gtk.2.18/opam
@@ -1,0 +1,32 @@
+opam-version: "2.0"
+synopsis: "Decentralised installation system - GTK UI"
+maintainer: "talex5@gmail.com"
+license: "LGPL-2.1-or-later"
+authors: "zero-install-devel@lists.sourceforge.net"
+homepage: "https://0install.net/"
+bug-reports: "https://github.com/0install/0install/issues"
+dev-repo: "git+https://github.com/0install/0install.git"
+doc: "https://0install.github.io/0install/"
+build: [
+  ["dune" "build" "-p" name "-j" jobs "@install" "@runtest" {with-test}]
+]
+depends: [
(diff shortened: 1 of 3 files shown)


talex5:
- fedora-34 failed because it has a buggy version of RPM (https://github.com/rpm-software-management/rpm/issues/1635), which was fixed in Fedora 35.
- macos is failing because of the sandbox causing the network tests to fail.
- ubuntu 18.04 failed to install ocurl.

kit-ty-kate:
fair enough. Thanks!

PR #22005, opened 2022-08-20
Package z3.4.11.0

smuenzel:
### `z3.4.11.0`
Z3 solver

---
* Homepage: https://github.com/Z3prover/z3
* Source repo: git+https://github.com/Z3prover/z3.git
* Bug tracker: https://github.com/Z3prover/z3/issues

---

### DIFF
--- a/packages/goblint/goblint.2.0.0/opam
+++ b/packages/goblint/goblint.2.0.0/opam
@@ -46,7 +46,7 @@ depends: [
   "benchmark" {with-test}
   "conf-gcc"
 ]
-depopts: ["apron" "z3"]
+depopts: ["apron"]
 conflicts: [
   "result" {< "1.5"}
   "z3"
--- a/packages/goblint/goblint.2.0.0/opam
+++ b/packages/goblint/goblint.2.0.0/opam
@@ -49,6 +49,7 @@ depends: [
 depopts: ["apron" "z3"]
 conflicts: [
   "result" {< "1.5"}
+  "z3"
 ]
 build: [
   ["dune" "subst"] {dev}
--- a/packages/z3/z3.4.11.0/opam
+++ b/packages/z3/z3.4.11.0/opam
@@ -1,0 +1,40 @@
+opam-version: "2.0"
+maintainer: "weng@cs.jhu.edu"
+authors: "MSR"
+homepage: "https://github.com/Z3prover/z3"
+bug-reports: "https://github.com/Z3prover/z3/issues"
+license: "MIT"
+dev-repo: "git+https://github.com/Z3prover/z3.git"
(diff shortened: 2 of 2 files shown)


smuenzel:
I removed the message about static linking, since it's been the case for versions 4.9.X and 4.10.X, so everyone should have had time to adjust

smuenzel:
The goblint failure seems unrelated to the z3 change, it fails without this change as well.

smuenzel:
@sim642, it seems like that when z3 is installed, goblint detects it and the compilation fails. I believe this is a new issue in the current goblint version, but I'm not sure

```
#=== ERROR while compiling goblint.2.0.0 ======================================#
# context     2.1.2 | linux/x86_64 |  | git+file:///home/stefan/development/opam-repository
# path        ~/.opam/4.14.0/.opam-switch/build/goblint.2.0.0
# command     ~/.opam/opam-init/hooks/sandbox.sh build dune build -p goblint -j 15 --promote-install-files=false @install
# exit-code   1
# env-file    ~/.opam/log/goblint-128398-43c72b.env
# output-file ~/.opam/log/goblint-128398-43c72b.out
### output ###
# (cd _build/default/src && /usr/bin/bash -e -u -o pipefail -c 'git describe --all --long --dirty || echo "n/a"') > /tmp/build_0f8b3f_dune/dune-pipe-action-_26aca4_.stdout
# fatal: not a git repository (or any parent up to mount point /)
# Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
# (cd _build/default && /home/stefan/.opam/4.14.0/bin/ocamlc.opt -w -40 -g -bin-annot -I src/.goblint_lib.objs/byte -I /home/stefan/.opam/4.14.0/lib/angstrom -I /home/stefan/.opam/4.14.0/lib/arg-complete -I /home/stefan/.opam/4.14.0/lib/astring -I /home/stefan/.opam/4.14.0/lib/batteries -I /home/stefan/.opam/4.14.0/lib/bigarray-compat -I /home/stefan/.opam/4.14.0/lib/bigstringaf -I /home/stefan[...]
# File "src/witness/z3/violationZ3.z3.ml", line 8, characters 7-10:
# Error: Unbound module Cil

```

sim642:
Thanks for the mention! It would be better to add a conflicts with `z3` (any version) to goblint.2.0.0 instead of the other way around (as you've done right now), because any version of `z3` installed would trigger that error.

Looks like the CI does trigger revdeps via `depopts` but not in the "forward" direction: this incompatibility didn't show in #21998 because the CI doesn't try different `depopts` combinations.

kit-ty-kate:
Thanks!

PR #22011, opened 2022-08-22
Fix base implicits

chetmurthy:
Fix a small version-number typo in `base-implicits` that prevents installing the modular-implicits variant.


kit-ty-kate:
Thanks

PR #22012, opened 2022-08-23
new file:   packages/bst/bst.7.0.1/opam

UnixJunkie:
added a dune-project file, so that we don't have a dune upper-bound version constraint

### DIFF
--- a/packages/bst/bst.7.0.1/opam
+++ b/packages/bst/bst.7.0.1/opam
@@ -9,7 +9,7 @@ build: ["dune" "build" "-p" name "-j" jobs]
 depends: [
   "ocaml"
   "batteries"
-  "dune"
+  "dune" {>= "1.11"}
   "base-unix" {with-test}
   "dolog" {with-test}
   "minicli" {with-test}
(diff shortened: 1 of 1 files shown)


packages/bst/bst.7.0.1/opam
```diff
@@ -0,0 +1,32 @@
...
+authors: ["Francois Berenger"]
+homepage: "https://github.com/UnixJunkie/bisec-tree"
+bug-reports: "https://github.com/UnixJunkie/bisec-tree/issues"
+dev-repo: "git+https://github.com/UnixJunkie/bisec-tree.git"
+license: "BSD-3-Clause"
+build: ["dune" "build" "-p" name "-j" jobs]
+depends: [
+  "ocaml"
+  "batteries"
+  "dune"
```
mseri, line 12:
```suggestion
  "dune" {>= "1.11"}
```

kit-ty-kate:
Thanks. Could you return the fix in the opam file upstream?

UnixJunkie:
Done, thanks!

PR #22017, opened 2022-08-23
Package: Wasmer

rajdakin:
This PR adds the Wasmer library support.
Note that this requires the latest version of OCaml 4.14.1+trunk, as there is an issue with the alternate signal stack in older versions.


packages/libwasmer/libwasmer.3.0.0~beta+macos-x86_64/opam
```diff
@@ -0,0 +1,23 @@
...
+dev-repo: "git+https://github.com/wasmerio/wasmer-ocaml.git"
+tags: ["org:wasmer"]
+
+synopsis: "The official Wasmer library"
+description: """
+This package is used by the wasmer package to trigger the installation
+of the Wasmer library."""
+
+extra-source "libwasmer.tar.gz" {
+	src: "https://github.com/wasmerio/wasmer/releases/download/3.0.0-beta/wasmer-darwin-amd64.tar.gz"
```
kit-ty-kate, line 19: The use of binary packages in opam is discouraged. You can see for example `tezos-rust-libs.1.0.0` as an example of a rust library built in opam by simply vendoring its dependencies and overriding the config file.

packages/wasmer/wasmer.1.0.0/opam
```diff
@@ -0,0 +1,29 @@
...
+depends: ["ocaml" {>= "4.14.1"} "ocamlfind" "ctypes" "ctypes-foreign" "libwasmer"]
+build: [[make "all"]]
+run-test: [[make "check"]]
+build-env: [
+	[ECHO = "true"]
+]
+setenv: [ # Hacky, but should work; prioritizes the opam library over the system one
+	[LIBRARY_PATH := "%{libwasmer:lib}%"]
+	[LD_LIBRARY_PATH := "%{libwasmer:lib}%"]
+	[DYLD_LIBRARY_PATH := "%{libwasmer:lib}%"]
```
kit-ty-kate, line 27: I'm not sure what you are trying to do with that but this is going to be refused. You are resetting the user environment. You might want to use `build-env` instead of `setenv` if this is just to build it.

packages/wasmer/wasmer.1.0.0/opam
```diff
@@ -0,0 +1,29 @@
...
+maintainer: "engineering@wasmer.io"
+license: "MIT"
+homepage: "https://github.com/wasmerio/wasmer-ocaml"
+doc: "https://github.com/wasmerio/wasmer-ocaml"
+bug-reports: "https://github.com/wasmerio/wasmer-ocaml/issues"
+dev-repo: "git+https://github.com/wasmerio/wasmer-ocaml.git"
+tags: ["org:wasmer"]
+url {
+	src: "https://github.com/wasmerio/wasmer-ocaml/archive/refs/tags/1.0.0.zip"
+	checksum: [] # Placeholder, TODO
```
kit-ty-kate, line 13: Every release archives have to have an accompagnied checksum to be merged.

kit-ty-kate (changes requested):

packages/wasmer/wasmer.1.0.0/opam
rajdakin: This is indeed used during build (which is why I then added it back to `build-env`), but the library is also required to be available at run time.
Also, I've read that `:=` is supposed to prepend the argument and adding a separator if necessary in the docs (instead of resetting the variable entirely)...
If there is a solution (like installing it by the `libwasmer` package), please tell me.

packages/libwasmer/libwasmer.3.0.0~beta+macos-x86_64/opam
rajdakin: This should work now.

packages/wasmer/wasmer.1.0.0/opam
rajdakin: Fixed.

packages/libwasmer/libwasmer.3.0.0~beta/opam
```diff
@@ -0,0 +1,46 @@
...
+		"md5=fd33f0fe59ce78d636b1a4178e4ef2f7"
+		"sha512=b0bee5168db59a7b629c4090cbd63adddf29931013b47e46050b1c30fc40937c886cf805eeacf43ab8d83d36204c5fb5abf17e90e178a09ebd224af598bfa9b2"
+	]
+}
+build: [
+	["mkdir" ".cargo"]
+	["mv" "cargo-config" ".cargo/config.toml"]
+	["tar" "xf" "vendor.tar.gz"]
+	["make" "build-capi"]
+	["make" "package-capi"]
```
kit-ty-kate, line 39:
```suggestion
	[make "build-capi"]
	[make "package-capi"]
```

packages/libwasmer/libwasmer.3.0.0~beta/opam
```diff
@@ -0,0 +1,46 @@
...
+]
+run-test: [
+	["make" "test-capi"]
```
kit-ty-kate, line 42:
```suggestion
	[make "test-capi"]
```

packages/libwasmer/libwasmer.3.0.0~beta/opam
```diff
@@ -0,0 +1,46 @@
...
+]
+install: [
+	["make" "install-capi"]
```
kit-ty-kate, line 45:
```suggestion
	[make "install-capi"]
```

packages/wasmer/wasmer.1.0.0/opam
```diff
@@ -0,0 +1,25 @@
...
+	src: "https://github.com/wasmerio/wasmer-ocaml/archive/refs/tags/1.0.0+api-fix.tar.gz"
+	checksum: [
+		"md5=7bc01f59449024de88857fc15d0f12ee"
+		"sha512=efa6ef01459ed5def7c9ead773147bc875c97a75c5321c46a18cc2891d22f3109718717078a43549e342e3f04242ba1a95b3218060cb0e9c6d3b7ca4bf398078"
+	]
+}
+
+synopsis: "The official Wasmer bindings for OCaml"
+
+depends: ["ocaml" {>= "4.14.1"} "ocamlfind" "ctypes" "ctypes-foreign" "libwasmer"]
```
kit-ty-kate, line 21: It's easier to modify later this way
```suggestion
depends: [
  "ocaml" {>= "4.14.1"}
  "ocamlfind"
  "ctypes"
  "ctypes-foreign"
  "libwasmer"
]
```

packages/wasmer/wasmer.1.0.0/opam
```diff
@@ -0,0 +1,25 @@
...
+build: [["make" "all"]]
+run-test: [["make" "check"]]
+build-env: [[ECHO = "true"]]
+install: ["make" "install"]
```
kit-ty-kate, line 25:
```suggestion
build: [make "all"]
run-test: [make "check"]
build-env: [ECHO = "true"]
install: [make "install"]
```

packages/libwasmer/libwasmer.3.0.0~beta/opam
kit-ty-kate: "make" vs. `%{make}%` where the former is always `make` (whatever the platform) and the latter is always GNU Make (e.g. `gmake` on BSDs)

packages/libwasmer/libwasmer.3.0.0~beta+macos-x86_64/opam
rajdakin: Update: it seems the system's Cargo and Rust are not up to date (Wasmer requires Rust 1.59 for now, while the system installs cargo 0.47.0 with rustc version 1.48.0), which makes the compilation fail ("feature `resolver` is required")... So I either need to execute the `rustup` script from rust-lang.org (`curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh`) -- which also requires an online connection during build, use a precompiled Rust binary (though I don't even know where to find the required version, let alone use it), or use the precompiled binaries.
I'm going back to using the precompiled binaries for now, but I'm open for alternatives.

(PS: this is how wasmtime does it as well, see #17776 and #17948.)

kit-ty-kate: > the system's Cargo and Rust are not up to date

which system? If it is only some of the distribution we're testing in CI but some worked it's totally fine.

rajdakin: I think it was the 4.02 compiler in the CI? I'm not sure, but it used Debian 11.
In any case, some in the company want to push toward even more recent `rustc` ("the Rust ecosystem makes it so that using an outdated Rust version is near useless"), so using the package manager's `rustc` would be very unstable depending on the version.

packages/libwasmer/libwasmer.3.0.0~beta/opam
rajdakin: Fixed.

packages/libwasmer/libwasmer.3.0.0~beta/opam
rajdakin: Fixed.

packages/libwasmer/libwasmer.3.0.0~beta/opam
rajdakin: Fixed.

packages/wasmer/wasmer.1.0.0/opam
rajdakin: Fixed.

packages/wasmer/wasmer.1.0.0/opam
rajdakin: Fixed.

packages/libwasmer/libwasmer.3.0.0~beta+macos-x86_64/opam
kit-ty-kate: That's ok, some packages do work this way

kit-ty-kate:
Please consider fixing https://github.com/ocaml/opam-repository/pull/22017/commits/84e142ce50872d9f4c07fa9f9ea3689206dbe9f0 instead. The compatible distributions were only failing because of wrongly defined lock file:
```
# RUSTFLAGS="" cargo build  --manifest-path lib/c-api/Cargo.toml --release \
# 	--no-default-features --features wat,compiler,wasi,middlewares --features cranelift,singlepass,wasmer-artifact-create,static-artifact-create,wasmer-artifact-load,static-artifact-load
# error: failed to select a version for the requirement `wat = "^1.0"` (locked to 1.0.47)
# candidate versions found which didn't match: 1.0.48
# location searched: directory source `/home/opam/.opam/4.14/.opam-switch/build/libwasmer.3.0.0~beta/vendor` (which is replacing registry `crates-io`)
# required by package `wasmer v3.0.0-beta (/home/opam/.opam/4.14/.opam-switch/build/libwasmer.3.0.0~beta/lib/api)`
# perhaps a crate was updated and forgotten to be re-vendored?
# make: *** [Makefile:428: build-capi] Error 101
```

packages/wasmer/wasmer.1.0.0/opam
rajdakin: I fixed the issue in release 1.1.0.

kit-ty-kate:
Hi, could you stop using our CI resources for something that is not going to be merged in this state. Please see my multiple pleading above

rajdakin:
Hi, I'm sorry for not noticing your comment.
I'm still not understanding why the CI fails where the tests on my machine works... Also, the local build of the library shouldn't change the outcome (the issue being that the library is not required by the executable, even though it is linked against in the command line), which is why I didn't try to change it.

kit-ty-kate:
The CI can be emulated locally with e.g.:
```
cd $(mktemp -d)
git clone --recursive "https://github.com/ocaml/opam-repository.git" && cd "opam-repository" && git fetch origin "refs/pull/22017/head" && git reset --hard 0f69e3f8
git fetch origin master
git merge --no-edit 42a177d7ac37cd347aab366a90d20469203fc926
cat > ../Dockerfile <<'END-OF-DOCKERFILE'
FROM ocaml/opam:debian-11-ocaml-4.14@sha256:d861cc4efa66cefb01b2eca0fd6b7246b935b47386983cd6399b482c82626343
USER 1000:1000
WORKDIR /home/opam
RUN sudo ln -f /usr/bin/opam-dev /usr/bin/opam
RUN opam init --reinit -ni
ENV OPAMDOWNLOADJOBS="1"
ENV OPAMERRLOGLEN="0"
ENV OPAMSOLVERTIMEOUT="500"
ENV OPAMPRECISETRACKING="1"
RUN rm -rf opam-repository/
COPY --chown=1000:1000 . opam-repository/
ENV OPAMCRITERIA="-removed,-count[avoid-version,changed],-count[version-lag,request],-count[version-lag,changed],-count[missing-depexts,changed],-changed"
ENV OPAMFIXUPCRITERIA="-removed,-count[avoid-version,changed],-count[version-lag,request],-count[version-lag,changed],-count[missing-depexts,changed],-changed"
ENV OPAMUPGRADECRITERIA="-removed,-count[avoid-version,changed],-count[version-lag,request],-count[version-lag,changed],-count[missing-depexts,changed],-changed"
RUN opam pin add -k version -yn wasmer.1.1.0 1.1.0
RUN opam update --depexts
RUN opam reinstall wasmer.1.1.0; \
    res=$?; \
    test "$res" != 31 && exit "$res"; \
    export OPAMCLI=2.0; \
    build_dir=$(opam var prefix)/.opam-switch/build; \
    failed=$(ls "$build_dir"); \
    for pkg in $failed; do \
    if opam show -f x-ci-accept-failures: "$pkg" | grep -qF "\"debian-11\""; then \
    echo "A package failed and has been disabled for CI using the 'x-ci-accept-failures' field."; \
    fi; \
    done; \
    exit 1

END-OF-DOCKERFILE
docker build -f ../Dockerfile .
```

PR #22030, opened 2022-08-26
[new release] eris (1.0.0)

pukkamustard:
This adds the `eris` and `eris-lwt` packages (https://codeberg.org/eris/ocaml-eris/). This is an OCaml implementation of the [Encoding for Robust and Immutable Storage (ERIS)](http://purl.org/eris).

Two crypto implementations (`eris.crypto-zig` and `eris.crypto-wasm`) require the [Zig](https://ziglang.org/) compiler to be installed.

Tests are run using the `monocypher` crypto library.


pukkamustard:
CI fails due to missing Zig compiler.

Tests (`dune test -p eris`) would run even if `dune build -p eris` fails as by default `monocypher` is used as crypto implementation. But the CI does not continue with tests if build fails.

Any suggestions on what could be improved? Add a virtual package for zig? Merge as is?

kit-ty-kate:
Could you add a conf-zig package? See https://github.com/ocaml/opam-repository/blob/master/packages/conf-jq/conf-jq.1/opam as an example

pukkamustard:
Compilation with Zig now fails (on distros where Zig can be installed) with a ReadOnlyFileSystem error. I suspect it has something to do with a missing `XDG_CACHE_HOME` directory. Zig seems to require this set to something writeable (see https://github.com/ziglang/zig/issues/6810).

Anyway we can set the `XDG_CACHE_HOME` environment variable from the opam file?

mseri:
Is there a way to make zig do all its stuff locally in the _build folder? The sandbox will prevent it to access any other path

pukkamustard:
> Is there a way to make zig do all its stuff locally in the _build folder? The sandbox will prevent it to access any other path

Yes, by setting the `XDG_CACHE_HOME` directory.

> Anyway we can set the XDG_CACHE_HOME environment variable from the opam file?

Sorry for being lazy and not reading the manual first. I figure this can be done with `build-env` in the opam file. Trying it now.

pukkamustard:
Setting `build-env: [XDG_CACHE_HOME = "$TMP"]` fails as `$TMP` is no expanded and for some reason Zig needs a shared cache between the two directories `lib/crypto/zig` and `lib/crypto/wasm`. Trying with `build-env: [XDG_CACHE_HOME = "../cache"]`.

pukkamustard:
Seems to work now!

Some remaining failures seem to be due to the distributions/OSs:
- oraclelinux gets stuck in a loop when trying to install zig (package does not yet exist). Build finally times-out after two hours.
- macos and debian (used by extras/opam-2.0) fails as maybe some x-ci-accepts-failures field is set (?).

Sorry for putting so much work on the CI.

mseri:
macos failure is strange, unfortunately I cannot test it locally since the current zig on macos arm with macports is broken. I will merge and revisit the availability later it in case I find that it cannot compile on mac

mseri:
Feel free to announce the release on discuss.ocaml.org. We have a community category and an announcement tag for this purpose

mseri:
I can confirm that it works on mac (tried on arm with latest zig installed manually)

PR #22043, opened 2022-08-29
frama-c: add macOS-specific patch

maroneze:
To avoid issues when users do not have GNU bash installed (e.g. via Homebrew).

### DIFF
--- a/packages/frama-c/frama-c.25.0/files/macos.patch
+++ b/packages/frama-c/frama-c.25.0/files/macos.patch
@@ -1,0 +1,14 @@
+diff --git a/src/plugins/value/gen-api.sh b/src/plugins/value/gen-api.sh
+index d00c16fdd4a..a64e6268c2d 100755
+--- a/src/plugins/value/gen-api.sh
++++ b/src/plugins/value/gen-api.sh
+@@ -22,7 +22,8 @@ for i in "$@"
+ do
+     file=$(basename $i)
+     module=${file%.*}
+-    printf '\nmodule %s: sig\n' ${module^}
++    Module="$(echo "${module:0:1}" | tr '[:lower:]' '[:upper:]')${module:1}"
++    printf '\nmodule %s: sig\n' $Module
+     awk '/\[@@@ api_start\]/{flag=1;next} /\[@@@ api_end\]/{flag=0} flag{ print (NF ? "  ":"") $0 }' $i
+     printf 'end\n'
+ done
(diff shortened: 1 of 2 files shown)


maroneze:
Hum, I forgot about this PR, until a new user reported the same issue with macOS that made me created the PR in the first place.

It seems there are additional CI failures w.r.t. the latest merged PR, but I'm not sure why. For instance, in the macos-homebrew, we have this error:

    # configure: error: unsupported version; GNU Make version 4.0

Indeed, Frama-C's configure needs either `gmake` available (via Homebrew, which usually is > 4.0), or a `make` command that is also GNU make, and > 4.0. In our test machines, our `make` command calls the Homebrew-installed version. In both cases, it _is_ compatible with homebrew, but requires the gmake package.

What I don't understand, is why the simple addition of a patch to another file (that is unrelated to `configure`) would make the CI fail this time, but not previously. Unless the macOS-Homebrew test configuration is new? In this case, it would be expected to fail, due to the absence of gmake. In practice, however, this patched version does allow Frama-C to compile when gmake is installed, while the version currently available in opam fails even in that case.

The other "new" failure is `extras,opam-2.0`, which fails due to the absence of package `libgtksourceview2.0-dev`, but that is a known issue (Debian 11).

Is there something I could do to improve this PR? Maybe add some configurations to `x-ci-accept-failures`?

maroneze:
Could someone please take a look at this PR and tell us if we need to change something to get it merged?

From what I could tell from the CI logs:
- Most failures are due to the lack of lablgtk2 or conf-gtksourceview packages in several distros;
- For homebrew, our script fails when GNU Make 4 is not detected. This is intentional: our makefile uses some features unavailable in macOS' default GNU Make.

Still, users with Homebrew can install `gmake` to get a recent GNU make, and thus use this on macOS. `gmake` is checked by our configure script and used if available.

The overall bug fixed by this PR is that, on macOS *with gmake*, compilation will fail due to use of a Bash 4 feature in a script. The patch in this PR fixes that by not making use of Bash 4 features. Thus macOS users with GNU make are able to compile and use Frama-C 25.

From what we tested, this PR strictly improves the situation on macOS, and does not affect anything on other systems, so it should be safe to be merged.

AllanBlanchard:
This is still an issue. Could someone tell us if there is something wrong here?

mseri:
No, sorry. I just wanted to test it locally

PR #22045, opened 2022-08-30
[new release] rfc1951 and decompress (1.5.1)

dinosaure:
Implementation of RFC1951 in OCaml

##### CHANGES:

- Fix the stream of gzip inflation. If the user wants to know how many bytes
  are available into the output buffer, he/she must be in the DEFLATE internal
  state. Otherwise, we raise an exception. However, such information is not
  available into the API so we decided to say that the full output buffer is
  free when we are into the GZip header state.

  It ensures a real full stream API. (@dinosaure, mirage/decompress#144)


dinosaure:
Same as #22040, see https://github.com/ocaml-multicore/ocaml-multicore/issues/602 for more details about the failure.

PR #22049, opened 2022-08-31
relaxes piqilib's upper bound

ivg:
It builds fine with 4.14.0 on my machine, but let's try 5.0.0, if it won't work, then I will add `& < "5.0.0"`.

cc @alavrik


ivg:
okay, looks like it works fine 4.14.x and is automatically disabled on 5.0.0 (because it requires sedlex < 3.0) There are a couple of revdep failures (all are byts vs strings errors), let me work on the real quick by adding proper constraints to them.

kit-ty-kate:
Thanks a lot!

PR #22050, opened 2022-08-31
Package: Wasmer

rajdakin:
This PR adds the Wasmer package.
Note: this PR ~has been tested to work on the Docker container, and~ makes use of the precompiled Wasmer libraries to allow its usage on every system (since Wasmer uses features incompatible with older versions of Rust that may be installed by the system's package manager).

EDIT: turns out the tests also fail on my machine, but only using the provided script in the CI. I'll investigate and fix it (but I can reproduce the issue).


kit-ty-kate (changes requested):
Same comments as in #22017

rajdakin:
I've just attempted to build the Wasmer library from source in the Docker container, but the system `cargo` (cargo 1.46.0) is too old to support the required features ("feature `resolver` is required", then if I remove the problematic line, "failed to parse the `edition` key" because "this version of Cargo is older than the `2021` edition, and only supports `2015` and `2018` editions.").

rajdakin:
As a side note, I can't use the provided script to build the MacOS version as the Docker image seems to not exist (docker fails with the error "pull access denied for macos-homebrew-ocaml-4.14, repository does not exist or may require 'docker login': denied: requested access to the resource is denied").

kit-ty-kate:
https://github.com/ocaml/opam-repository/pull/22204 added a `conf-rust-2022` package. Is there any chance you could take example from this PR to have the wasmer built from source instead?

rajdakin:
I've looked at that package, but the system `cargo` for the "main tests" (Debian 11, which is the image used by the 5.0 OCaml compiler version) is only 1.46.0, which, as I mentioned previously, is too old. If you read the `opam` file, you can see that the `centos-7`, `oraclelinux-7`, `debian-10`, `debian-11` and `alpine-3.14` distributions are all expected to fail (with the additional comment of "rust 2021 is not in the default repository (version is too old)" or "does not have cargo by default")...

kit-ty-kate:
Yeah, that's fine by us.

kit-ty-kate:
in practice these are all old distributions and it shouldn't be too much of a big deal for users as they would be notified by the `conf-rust-2021` package, that their rust version is too old. This is only a (temporary) problem for CI (we're looking at fixing that)

packages/libwasmer/libwasmer.3.0.0~beta+build/opam
```diff
@@ -0,0 +1,48 @@
...
+	["cargo"] {os-distribution = "ol" & os-version >= "8"}
+	["cargo"] {os-distribution = "fedora"}
+	["cargo"] {os-family = "suse"}
+	["cargo"] {os-family = "debian"}
+	["cargo"] {os-family = "ubuntu"}
+	["cargo" "rustc"] {os-distribution = "nixos"}
+	["cargo"] {os-family = "alpine"}
+	["rust"] {os-family = "arch"}
+	["rust"] {os = "macos" & os-distribution = "homebrew"}
+]
```
kit-ty-kate, line 29: This can be simplified
```suggestion
depends: [
  "conf-rust-2021"
]
```

packages/libwasmer/libwasmer.3.0.0~beta+build/opam
```diff
@@ -0,0 +1,48 @@
...
+	src: "https://github.com/wasmerio/wasmer-ocaml/releases/download/v1.2.1%2Bdunefix/libwasmer-vendor.tar.gz"
+}
+build: [
+	["mkdir" ".cargo"]
+	["mv" "config.toml" ".cargo"]
+	["sh" "-c" "gunzip vendor.tar.gz && tar xf vendor.tar"]
+	[make "build-capi"]
+	[make "package-capi"]
+]
+available: (os-distribution = "centos" & os-version >= "8") | (os-distribution = "ol" & os-version >= "8") | (os-distribution = "fedora") | (os-family = "suse") | (os-family = "debian" & os-version >= "12") | (os-family = "ubuntu") | (os-distribution = "nixos") | (os-family = "alpine" & os-version > "3.14") | (os-family = "arch") | (os = "macos" & os-distribution = "homebrew")
```
kit-ty-kate, line 48: This shouldn't be needed with the dependency above 
```suggestion
```

rajdakin: I was just about to ask for confirmation: if the `conf-rust-2021` package fails to install, it will instead try to use one of the precompiled binaries right? Also, can I tell `opam` that this package should be used in priority or should I mark the other ones with the `avoid-version` flag?

kit-ty-kate: no but do you really need support for all the older distributions in CI? (regular users can just install rust manually)

packages/libwasmer/libwasmer.3.0.0~beta+build/opam
rajdakin: Fixed

packages/libwasmer/libwasmer.3.0.0~beta+build/opam
rajdakin: Alright, I didn't think of that. I've removed the other ways to install the package.

packages/wasmer/wasmer.1.2.1/opam
```diff
@@ -0,0 +1,36 @@
...
+		"md5=be37208390097ed6b7c7d9ede3c1a2a7"
+		"sha512=2422c3b32ae053a9f39336b1c9829ec9a508dd7d303e9a68853287ab7dea90b7cbc57f54a04b20a9fb6217527e1c89e8f806349407db5f8679d90fdf7dd3038e"
+	]
+}
+
+synopsis: "OCaml bindings for Wasmer"
+description: "The official Wasmer bindings for OCaml"
+
+depends: [
+  "ocaml" {>= "4.14.1"}
```
kit-ty-kate, line 23: OCaml 4.14.1 hasn't been released yet
```suggestion
  "ocaml" {>= "4.14.0"}
```

rajdakin:
EDIT: I've reverted the change to use `conf-rust-2021` as it only requires Rust edition 2021, which only requires Rust to be at least 1.56, but a Rust package (`corosensei v0.1.3`) has Rust version 1.5**9** as a requirement (which is therefore not guaranteed by this package).

kit-ty-kate:
> EDIT: I've reverted the change to use `conf-rust-2021` as it only requires Rust edition 2021, which only requires Rust to be at least 1.5**6**, but a Rust package used by `libwasmer` (`corosensei v0.1.3`) has Rust version 1.5**9** as a requirement (which is therefore not guaranteed by this package).

I would've said "close enough". We can't really have one package per version (that'd be a bit too much, though we could if you really need), but the error message seemed to be precise enough to indicate what to do
```
# error: package `corosensei v0.1.3` cannot be built because it requires rustc 1.59.0 or newer, while the currently active rustc version is 1.58.1
```
To me, modulo the suggestion i made just above, it looked fine to merge before.

packages/wasmer/wasmer.1.2.1/opam
rajdakin: Yes, but OCaml 4.14.0 has a bug that makes the program crash when exiting (see ocaml/ocaml#11489). It has been fixed only for OCaml 5.0 and 4.14.1, but 4.14.0 still has this bug.

kit-ty-kate:
Should be good to go once https://github.com/ocaml/opam-repository/pull/22050#discussion_r997396463 is fixed

packages/wasmer/wasmer.1.2.1/opam
kit-ty-kate: ah my bad. Good to go then.

packages/libwasmer/libwasmer.3.0.0~beta/files/libwasmer.install
```diff
@@ -0,0 +1,3 @@
+lib: [
+	"package/lib/libwasmer.so"
```
kit-ty-kate, line 2: Ah, nevermind there seems to be an issue on macOS where this file doesn't exist. Shouldn't it be `libwasmer.dylib` there? Or maybe you'd prefer to take the static archive?

kit-ty-kate: if you really need the dynamic library over the static one:
```suggestion
	"?package/lib/libwasmer.so"
	"?package/lib/libwasmer.dylib"
	"?package/lib/libwasmer.dll"
```

kit-ty-kate: If your package works with the static archive only the following should be necessary:
```suggestion
	"package/lib/libwasmer.a"
```

rajdakin: It should be, but this file is OS-independent and Dune **only** supports the `.so` extension on MacOS. So I need to rename the library after building it. (Also the package doesn't work or generate massive binaries with the static library, so I need to use the dynamic one.)
I've added the Windows extension for future-proofing though.

kit-ty-kate: Actually, according to the Wasmer Makefile:
```
if [ -f target/release/wasmer.dll ]; then \
	cp target/release/wasmer.dll package/lib/wasmer.dll ;\
fi
if [ -f target/release/wasmer.dll.lib ]; then \
	cp target/release/wasmer.dll.lib package/lib/wasmer.dll.lib ;\
fi
if [ -f target/release/wasmer.lib ]; then \
	cp target/release/wasmer.lib package/lib/wasmer.lib ;\
fi
if [ -f target/release/libwasmer.dylib ]; then \
	cp target/release/libwasmer.dylib package/lib/libwasmer.dylib ;\
fi
if [ -f target/release/libwasmer.so ]; then \
	cp target/release/libwasmer.so package/lib/libwasmer.so ;\
fi
if [ -f target/release/libwasmer.a ]; then \
	cp target/release/libwasmer.a package/lib/libwasmer.a ;\
fi
```
so to be more complete it should be:
```suggestion
	"?package/lib/libwasmer.dll"
	"?package/lib/libwasmer.dll.lib"
	"?package/lib/libwasmer.dylib"
	"?package/lib/libwasmer.so"
	"?package/lib/libwasmer.a"
```

kit-ty-kate: > It should be, but this file is OS-independent and Dune **only** supports the `.so` extension on MacOS. So I need to rename the library after building it.

really? It should normally. It's only `.so` on all platforms for OCaml libraries but the system libraries it stays with their builtin names AFAIK

rajdakin: > so to be more complete it should [include the `.dll`, `.dll.lib`, `.dylib`, `.so` and `.a` libraries]

That's true, but I only need the `.so` (or the `.dll` on Windows probably).

> It's only .so on all platforms for OCaml libraries but the system libraries stay with their builtin names AFAIK

That is probably correct if it can find it (I think it links using the `-l` linker option), but I force the linker to use the *exact library* in the `~/.opam` directory (which may or may not be the best way to do it, but it is the only way I managed to make it work on Linux and Mac at the same time). Also, `libwasmer` is an OCaml library in this context since it is built and installed by `opam`, inside the `~/.opam` directory.

kit-ty-kate: ah i see you've used the OCaml extension manually https://github.com/wasmerio/wasmer-ocaml/blob/d3b2b5721a4013fa4ba942097c3435418c13f095/lib/dune#L5

Mmh, there is an issue though. I've tried to use your library locally to see if it worked and i get this error in the OCaml toplevel:
```
# #use_output "dune top";;
Exception:
Dl.DL_error
 "/home/kit_ty_kate/.opam/default/lib/stublibs/dllctypes-foreign_stubs.so: undefined symbol: wat2wasm".
```
This is because the dynamic library is not present in the default LD_LIBRARY_PATH. So either your users will have to set it manually or I would rather suggest to use static libraries instead even if it makes the resulting binary bigger.

rajdakin: I didn't think of the toplevel. However, I did fix the issue with the executables by setting the `rpath` as well, which is added in the linking step for executables (but not the `wasmer` library itself as it is a `-cclib` option). I'm not really sure how to fix the toplevel though...

kit-ty-kate: ah i see. If it's only the toplevel then it's fine.

kit-ty-kate:
Thanks! Sorry for the wait.

kit-ty-kate (approved):

PR #22051, opened 2022-09-01
[ppx_import] [rev-deps] Add ppx_import < 2.0 constraint to all its deps.

ejgallego:
Syntax for ppx_import 2.0 will change in an incompatible way; this PR
is thus a preventive action to avoid breakage for all the packages
that use it.

Note that most users depend on ppx_import only for tests.


kit-ty-kate:
Letting the linter double-check that everything is good, and good to merge after that.

kit-ty-kate:
Thanks a lot!

ejgallego:
Thanks to you @kit-ty-kate !

Issue #19840, opened 2021-10-21
llvm distrib package is unavailable on archlinux, alpine and opensuse


thierry-martinez:
While trying to update `conf-libclang` and make it available on recent distributions (#19824), I am facing problems with the package managers of ~~archlinux,~~ alpine and opensuse inside the dockers of Opam CI. For these three distributions, I am not able to reproduce these problems when I run a docker locally, so I suspect that this is due to some caching problems in opam CI.

In particular:
- ~~with *archlinux*, in opam CI, running `llvm-config` after the installation of the package `llvm` produces the following error:~~
```
llvm-config: error while loading shared libraries: libffi.so.8: cannot open shared object file: No such file or directory
```
~~(see https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/30f4971f7357c8c73ce37001155ccd3e7a5ab502/variant/opam-2.0,distributions,archlinux,conf-libclang.13 )~~
~~while running `llvm-config` after `sudo pacman -Syu llvm` in a fresh `docker run --rm -it ocaml/opam:archlinux-ocaml-4.12` goes fine.~~

- with *alpine*, in opam CI, `opam` reports
```
Package conf-libclang.13 depends on the unavailable system package 'llvm-dev'. You can use `--no-depexts' to attempt installation anyway
```
while running `sudo apk install llvm-dev` in a fresh `docker run --rm -it ocaml/opam:alpine-3.13-ocaml-4.12` succeeds.
(see https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/30f4971f7357c8c73ce37001155ccd3e7a5ab502/variant/opam-2.1,distributions,alpine-3.13,conf-libclang.13 )

- with *opensuse*, in opam CI, `opam` reports
```
Package conf-libclang.13 depends on the unavailable system package 'llvm-clang-devel'. You can use `--no-depexts' to attempt installation anyway.
```
while running `sudo zypper install llvm-clang-devel` in a fresh `docker run -it --rm ocaml/opam:opensuse-15.3-ocaml-4.12` succeeds.
(see https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/30f4971f7357c8c73ce37001155ccd3e7a5ab502/variant/opam-2.1,distributions,opensuse-15.3,conf-libclang.13 )

thierry-martinez:
[...]

PR #22062, opened 2022-09-05
[new release] clangml.4.8.0

thierry-martinez:
This commit pushes a new release of clangml.4.8.0, compatible with
the upcoming LLVM/Clang 15. Changes are:

- Support for Clang/LLVM 15.0.0

- `Ast.character_kind` and `Ast.string_kind` are now distinct
  types (aliases for `Clang.clang_ext_characterkind` and
  `Clang.clang_ext_stringkind` respectively.
  The constructor `Ordinary` replaced the former constructor `Ascii`
  for `string_kind`, to match the new convention used by
  `clang::StringLiteral::StringKind` from Clang 15.0.0.
  The constructor `Ascii` for `character_kind` is left unchanged.

- https://github.com/thierry-martinez/clangml/issues/1, https://github.com/thierry-martinez/clangml/pull/2: `Ast.Var` and `Ast.Function` constructors now have a `storage`
  field in addition to the computed `linkage`, exposing the value
  previously accessible via `cursor_get_storage_class`. The
  storage classes are now correctly printed by the printer.
  (reported by Ronan, rsaill/n47, https://github.com/thierry-martinez/clangml/issues/1)

- https://github.com/thierry-martinez/clangml/issues/3, https://github.com/thierry-martinez/clangml/pull/4: fix `Clang.Expr.radix_of_integer_literal` when the literal
  comes from a macro expansion, and fix printing of unsigned/long
  integer literals with `Clang.Printer`.
  (reported by Ronan, rsaill/n47, https://github.com/thierry-martinez/clangml/issues/3)

### DIFF
--- a/packages/clangml/clangml.4.7.0/opam
+++ b/packages/clangml/clangml.4.7.0/opam
@@ -11,7 +11,7 @@ homepage: "https://github.com/thierry-martinez/clangml/"
 doc: "https://github.com/thierry-martinez/clangml/blob/v4.7.0/README.md"
 bug-reports: "https://github.com/thierry-martinez/clangml/issues"
 depends: [
-  "conf-libclang" {<= "14"}
+  "conf-libclang" {< "15"}
   "conf-ncurses"
   "conf-zlib"
   "dune" {>= "1.11.0"}
--- a/packages/clangml/clangml.4.7.0/opam
+++ b/packages/clangml/clangml.4.7.0/opam
@@ -10,7 +10,6 @@ license: "BSD-2-Clause"
 homepage: "https://github.com/thierry-martinez/clangml/"
 doc: "https://github.com/thierry-martinez/clangml/blob/v4.7.0/README.md"
 bug-reports: "https://github.com/thierry-martinez/clangml/issues"
-available: arch != "s390x"
 depends: [
   "conf-libclang" {<= "14"}
   "conf-ncurses"
--- a/packages/clangml/clangml.4.7.0/opam
+++ b/packages/clangml/clangml.4.7.0/opam
@@ -10,6 +10,7 @@ license: "BSD-2-Clause"
 homepage: "https://github.com/thierry-martinez/clangml/"
 doc: "https://github.com/thierry-martinez/clangml/blob/v4.7.0/README.md"
 bug-reports: "https://github.com/thierry-martinez/clangml/issues"
+available: arch != "s390x"
 depends: [
   "conf-libclang" {<= "14"}
   "conf-ncurses"
--- a/packages/clangml/clangml.4.8.0/opam
+++ b/packages/clangml/clangml.4.8.0/opam
@@ -10,6 +10,7 @@ license: "BSD-2-Clause"
 homepage: "https://github.com/thierry-martinez/clangml/"
 doc: "https://github.com/thierry-martinez/clangml/blob/v4.8.0/README.md"
 bug-reports: "https://github.com/thierry-martinez/clangml/issues"
+available: arch != "s390x"
 depends: [
   "conf-libclang" {<= "15"}
   "conf-ncurses"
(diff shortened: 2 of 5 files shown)


thierry-martinez:
`s390x` looks unsupported by clangml (registry allocation failure in `ocamlopt`, and I don't see easy workaround), but I don't see how to declare this fact in `x-ci-accept-failures` (only `debian-11` is grepped, which is not specific enough). The error with `alpine-3.15` looks transient (no space left on device), and the failures with this distribution should be ignored anyway (the package `clang-dev` is not available in opam CI, whereas it looks available otherwise: this is probably related to https://github.com/ocaml/opam-repository/issues/19840).

packages/conf-libclang/conf-libclang.15/files/configure.sh
```diff
@@ -0,0 +1,125 @@
+#!/bin/bash -ex
```
kit-ty-kate, line 1: What's the difference between conf-libclang.14 and conf-libclang.15 ?

packages/clangml/clangml.4.8.0/opam
```diff
@@ -0,0 +1,41 @@
...
+"""
+maintainer: ["Thierry Martinez <thierry.martinez@inria.fr>"]
+authors: ["Thierry Martinez <thierry.martinez@inria.fr>"]
+license: "BSD-2-Clause"
+homepage: "https://github.com/thierry-martinez/clangml/"
+doc: "https://github.com/thierry-martinez/clangml/blob/v4.8.0/README.md"
+bug-reports: "https://github.com/thierry-martinez/clangml/issues"
+available: arch != "s390x"
+depends: [
+  "conf-libclang" {<= "15"}
```
kit-ty-kate, line 15:
```suggestion
  "conf-libclang" {< "16"}
```

packages/clangml/clangml.4.7.0/opam
```diff
@@ -11,7 +11,7 @@ homepage: "https://github.com/thierry-martinez/clangml/"
 doc: "https://github.com/thierry-martinez/clangml/blob/v4.7.0/README.md"
 bug-reports: "https://github.com/thierry-martinez/clangml/issues"
 depends: [
-  "conf-libclang"
+  "conf-libclang" {<= "14"}
```
kit-ty-kate, line 14:
```suggestion
  "conf-libclang" {< "15"}
```

kit-ty-kate (commented):
<= should be avoided

packages/conf-libclang/conf-libclang.15/files/configure.sh
thierry-martinez: They differ in `files/configure.sh`: `conf-libclang.14` searches for LLVM libraries up to version 14 (included), whereas `conf-libclang.15` searches for LLVM libraries up to version 15 (included). The purpose is to be able to express that `clangml.4.7.0` for instance requires `conf-libclang<15` (because the bindings only support clang 14 and below) and `clangml.4.8.0` requires `conf-libclang<16` (because the bindings only support clang 15 and below).

thierry-martinez: Full diff:
```diff
diff --color -r conf-libclang.14/files/configure.sh conf-libclang.15/files/configure.sh
9c9
< maximum_version=14
---
> maximum_version=15
diff --color -r conf-libclang.14/opam conf-libclang.15/opam
32c32
<   "configure.sh" "sha512=901562da7c7d5016b2b348ce4bf05ea2173b78e66d64f962a5cef0686b8df33c850ee26093ca8c00163043fc2506681fc014681ff75f7fcfdc5a57a67a48e162"
---
>   "configure.sh" "sha512=495b1915a845c75033449e8a86b35e399320a50d63cee959f5685dedb085c5efc2a27d5c2672c90c38444a5af4edd212cf2e23bb2b34aaedc5cf271073dce087"
34c34
< synopsis: "Virtual package relying on the installation of llvm and clang libraries (<= 14.0.x)"
---
> synopsis: "Virtual package relying on the installation of llvm and clang libraries (<= 15.0.x)"
```

thierry-martinez:
> <= should be avoided

Thanks!

kit-ty-kate:
Thanks

PR #22066, opened 2022-09-06
Add new Coq version 8.16.0

palmskog:
New release of Coq.

cc: @MSoegtropIMC @ppedrot

### DIFF
--- a/packages/coq/coq.8.16.0/opam
+++ b/packages/coq/coq.8.16.0/opam
@@ -1,0 +1,55 @@
+opam-version: "2.0"
+maintainer: "coqdev@inria.fr"
+authors: "The Coq development team, INRIA, CNRS, and contributors."
+homepage: "https://coq.inria.fr/"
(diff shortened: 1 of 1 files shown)


palmskog:
Looks like the only CI failure is a reverse dependency (`frama-c.24.0`). I don't think they are expected to work.

MSoegtropIMC:
Coq is a depopt of frama-c. Can one put in version restrictions in depopts (anyway it would require a PR to the main opam repo, we don't want to wait for).

MSoegtropIMC:
@palmskog : since I need coq and coqide in the Coq Platform patch repo: do you plan to change anything in the coqide opam file beyond the URL and SHA?

palmskog:
@MSoegtropIMC the version of the coqide package I'll submit (after this PR is merged) only has URL and SHA changed from 8.16+rc1. But as we've seen before, an opam file may have to be substantially rewritten to pass CI.

kit-ty-kate:
Thanks!

PR #22069, opened 2022-09-06
Add a warning shown on "opam update" to inform users who are using an out-of-date opam binary

kit-ty-kate:
This will show the following output for people using opam < 2.1.3
```
$ opam update tmp

<><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>
[tmp] synchronised from git+file:///home/kit_ty_kate/work/opam-repository
tmp (at git+file:///home/kit_ty_kate/work/opam-repository): 
    [WARNING] Your opam version is out-of-date. Please consider upgrading: https://opam.ocaml.org/doc/Install.html

```

### DIFF
--- a/repo
+++ b/repo
@@ -1,3 +1,8 @@
 opam-version: "2.0"
 browse: "https://opam.ocaml.org/pkg/"
 upstream: "https://github.com/ocaml/opam-repository/tree/master/"
+announce: [
+"""
+[WARNING] Your opam version is out-of-date. Please consider upgrading: https://opam.ocaml.org/doc/Install.html
+""" {opam-version < "2.1.3"}
+]


dra27 (commented):
This is an excellent change to be making - thank you for doing it!

While I of course strongly recommend all our users to update to the very latest release, we should differentiate changing from the latest release in an old series, which is a recommended _upgrade_ (i.e. 2.0.10 → 2.1.3) vs installing a security/bug-fix release, which is a "mandatory" _update_[^1]

[^1]: cf. an LTS Ubuntu, which _advises_ when a new LTS release is available, but _warns_ when updates to your system aren't installed

kit-ty-kate:
@dra27 thanks for the review. I modified slightly the suggestion to have a slightly more explicit filter (i never remember which has priority between `&` and `|`) and filled the missing gap between 2.0.10 and 2.1.0 (we never know… :sweat_smile:)

Personally i feel like `Please update it` is a bit too strong to direct at users and would prefer to go back to `Please consider updating it`

mseri:
> Personally i feel like `Please update it` is a bit too strong to direct at users and would prefer to go back to `Please consider updating it`

I agree, this sounds nicer

repo
```diff
@@ -1,3 +1,11 @@
 opam-version: "2.0"
 browse: "https://opam.ocaml.org/pkg/"
 upstream: "https://github.com/ocaml/opam-repository/tree/master/"
+announce: [
+"""
+[WARNING] opam is out-of-date. Please update it (https://opam.ocaml.org/doc/Install.html)
```
kit-ty-kate, line 6:
```suggestion
[WARNING] opam is out-of-date. Please consider updating it (https://opam.ocaml.org/doc/Install.html)
```

dra27:
This all looks good to me, thanks! The "missing gap" shouldn't happen (cf. the linked PR), but I agree it's better that if a mistake made we display _both_ the "please update" and "please upgrade" messages for 2.0.10 rather than neither for a 2.0.11!

PR #22070, opened 2022-09-06
Release Caqti 1.9.0.

paurkedal:
I would like to make a new Caqti release including a few feature and bug fixes ([release notes](https://github.com/paurkedal/ocaml-caqti/releases/tag/v1.9.0)).

### DIFF
--- a/packages/caqti-async/caqti-async.1.9.0/opam
+++ b/packages/caqti-async/caqti-async.1.9.0/opam
@@ -1,0 +1,37 @@
+opam-version: "2.0"
+maintainer: "Petter A. Urkedal <paurkedal@gmail.com>"
+authors: "Petter A. Urkedal <paurkedal@gmail.com>"
+license: "LGPL-3.0-or-later WITH LGPL-3.0-linking-exception"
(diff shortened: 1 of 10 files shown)


paurkedal:
I can see the lower bounds (at least) need some work. I'll take a closer look in the evening.

paurkedal:
Here is my take at the remaining CI errors:

  - (compilers 4.05) There is some build error for cppo on ocaml-4.05. Not sure, but could be it needs a newer dune/jbuild.
  - (compilers 4.14) `ppx_rapper.3.1.0` (and later) should depend on `caqti {>= "1.7.0"}`. The other Caqti dependencies should not need such a constraint. (Reported upstream: roddyyaga/ppx_rapper#26).
  - (compilers 5.0) The `mariadb` is caused by the use of `Stream` in the `setup.ml` script, so I guess it really as oasis issue.
  - (compilers 5.0) Old versions of `sqlite3` will need an upper compiler bound due to `String.lowercase` usage.
  - The remaining looks like CI problems.

mseri:
Thanks

PR #22072, opened 2022-09-06
Add new CoqIDE version 8.16.0

palmskog:
This is a followup to #22066, adding CoqIDE 8.16.0 served from the same archive.

cc: @MSoegtropIMC @ppedrot


mseri:
Thanks

PR #22080, opened 2022-09-07
[new release] irmin project (3.4.1)

icristescu:
### Added

- **irmin**
  - Add `Storage` module for creating custom storage layers (mirage/irmin#2047, @metanivek)

- **irmin-pack**
  - Add `Gc.is_allowed` (mirage/irmin#2076, @icristescu)
  - Add a `weight` parameter in the LRU implementation to bound
    memory usage (mirage/irmin#2050, @samoht)

### Changed

- **irmin**
  - Removed `Irmin_unix.set_listen_dir_hook` (mirage/irmin#2071, @zshipko)

### Fixed

- **irmin-pack**
  - Fix the behaviour of irmin-pack regarding hashes and keys to GCed objects.
    It used to not correctly ignore these entries, which could have resulted in
    various bugs. E.g. the impossibility to append an object that used to be
    dead and that has its hash in index. (mirage/irmin#2070, @Ngoguey42)

### DIFF
--- a/packages/irmin-bench/irmin-bench.3.4.1/opam
+++ b/packages/irmin-bench/irmin-bench.3.4.1/opam
@@ -1,0 +1,59 @@
+opam-version: "2.0"
+maintainer:   "thomas@gazagnaire.org"
+authors:      ["Thomas Gazagnaire"]
+license:      "ISC"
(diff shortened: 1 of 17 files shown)


kit-ty-kate:
Thanks

PR #22085, opened 2022-09-08
Update ocaml-protoc-plugin to 4.3.0

AndreasDahl:
* Use pkg-config to locate google well known types (thanks @vprevosto)
* Support proto3 optional fields


AndreasDahl:
We have a last-minute contribution that we want in this release. So I will close this PR now and open another one at a later time.

PR #22088, opened 2022-09-09
new file:   genspir/genspir.1.0.1/opam

UnixJunkie:
added dune-project file so that no dune upper bound is needed


kit-ty-kate:
Thanks

PR #22090, opened 2022-09-09
Package ocaml-protoc-plugin.4.3.0

andersfugmann:
### `ocaml-protoc-plugin.4.3.0`
- enums are mapped to adt's

---

---

### DIFF
--- a/packages/ocaml-protoc-plugin/ocaml-protoc-plugin.4.3.0/opam
+++ b/packages/ocaml-protoc-plugin/ocaml-protoc-plugin.4.3.0/opam
@@ -20,7 +20,7 @@ depends: [
   "ppx_inline_test" {with-test}
   "ppx_deriving" {with-test}
   "conf-pkg-config" {build}
-  "dune-configurator" {with-test}
+  "dune-configurator"
 ]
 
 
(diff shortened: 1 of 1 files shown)


andersfugmann:
CI Failed on s390x which is a big-endian. This is a bug in the code (Assuming little endian when converting floats), but is not a regression so I dont think it should be blocking a release.

packages/ocaml-protoc-plugin/ocaml-protoc-plugin.4.3.0/opam
```diff
@@ -0,0 +1,46 @@
...
+  "dune" {>= "3.2"}
+  "ocaml" {>= "4.08.0"}
+  "ppx_expect" {with-test}
+  "ppx_inline_test" {with-test}
+  "ppx_deriving" {with-test}
+  "conf-pkg-config" {build}
+  "dune-configurator"
+]
+
+
```
mseri, line 26:
```suggestion
available: arch != "s390x"
```

mseri: If it is broken there, I suggest to mark it unavailable. Or it is a minor incompatibility and most functionalities are expected to work?

andersfugmann: Its a real bug on all big endian architectures. which has not been discovered due to lack of testing on said architectures. 
I already created a fix for this, and its pending review. I will close this PR, and create a new release once the fix is in.

andersfugmann:
Superseded by https://github.com/ocaml/opam-repository/pull/22095

PR #22092, opened 2022-09-12
Why3 1.5.1

silene:


silene:
There is a single failure (no idea why it states that "2 jobs failed"), which is due to `opam depext` trying to install `libgtksourceview2.0-dev` and failing. The same job when using Opam 2.1 instead of 2.0 succeeds, as `opam depext` then installs a package that actually exists. So, this pull request seems ready.

AllanBlanchard:
> There is a single failure (no idea why it states that "2 jobs failed"), which is due to `opam depext` trying to install `libgtksourceview2.0-dev` and failing. The same job when using Opam 2.1 instead of 2.0 succeeds, as `opam depext` then installs a package that actually exists. So, this pull request seems ready.

Same behavior as we have in Frama-C : when both gtk2 and gtk3 are available in the `opam` file, Opam selects GTK2 and then depext fails to find GTK2 on the distribution.

kit-ty-kate:
Yes, that's an inherent flaw of opam 2.0 (fixed in 2.1). Thanks.

PR #22097, opened 2022-09-12
add missing *.a and *.cmx files

nguermond:
Dune cannot find agrep because agrep.a and agrep.cmx files are missing


nguermond:
Owner has fixed the issue, this fix is no longer necessary

PR #22098, opened 2022-09-13
fix linking problems with z3_tptp program

palmskog:
Currently, after installing a `z3_tptp` package, the `z3_tptp` program gives error messages like this whenever it is run:
```
z3_tptp: error while loading shared libraries: libz3.so: cannot open shared object file: No such file or directory
```
This update fixes the problem with the packages by modifying compilation options. The background and motivation can be seen in https://github.com/coq/platform/issues/248

cc: @MSoegtropIMC

### DIFF
--- a/packages/z3_tptp/z3_tptp.4.8.13/opam
+++ b/packages/z3_tptp/z3_tptp.4.8.13/opam
@@ -18,6 +18,8 @@ build: [
     "-o" "z3_tptp"
     "examples/tptp/tptp5.cpp" "examples/tptp/tptp5.lex.cpp" 
     "-lz3"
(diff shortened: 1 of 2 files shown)


MSoegtropIMC:
See also https://github.com/ocaml/opam-repository/pull/22100 and https://github.com/ocaml/opam-repository/pull/22099

mseri:
Thanks! Can we also merge https://github.com/ocaml/opam-repository/pull/22100?

MSoegtropIMC:
@mseri : #22100 first needs to get the same changes as @palmskog did here. I try to update today. Please also see my comment there on the rather strange CI issues (make called with -j 255).

PR #22122, opened 2022-09-16
vhd-format-lwt: solve conflicting deps on io-page

psafont:
io-page-unix is only available for 2.3.0 and lower, this conflicts with the requirement of io-page >= 2.4.0.

Remove the former so tests can be run.

### DIFF
--- a/packages/vhd-format-lwt/vhd-format-lwt.0.12.3/opam
+++ b/packages/vhd-format-lwt/vhd-format-lwt.0.12.3/opam
@@ -13,7 +13,6 @@ depends: [
   "mirage-block" {>= "2.0.1"}
   "ounit2" {with-test}
   "vhd-format"
-  "io-page-unix" {with-test}
   "dune" {>= "1.0"}
   "io-page" {with-test & >= "2.4.0"}
 ]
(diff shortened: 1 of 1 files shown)


packages/vhd-format-lwt/vhd-format-lwt.0.12.3/opam
```diff
@@ -7,15 +7,14 @@ homepage: "https://github.com/mirage/ocaml-vhd"
...
+  "ocaml" {>= "4.02.3" & < "5.0.0"}
   "cstruct" {< "6.1.0"}
   "lwt" {>= "3.2.0"}
   "mirage-block" {>= "2.0.1"}
   "ounit2" {with-test}
   "vhd-format"
-  "io-page-unix" {with-test}
   "dune" {>= "1.0"}
-  "io-page" {with-test & >= "2.4.0"}
+  "io-page" {with-test & = "2.4.0"}
```
kit-ty-kate, line 17: We try to avoid equality constraints when it is not a coherent set of packages (e.g. from the same repository or same developer). Would that work instead?
```suggestion
  "io-page" {with-test & >= "2.4.0" & < "3.0.0"}
```

mseri:
Thanks

PR #22124, opened 2022-09-17
[new release] http-lwt-client (0.0.8)

hannesm:
A simple HTTP client using http/af, h2, and lwt

- Project page: <a href="https://github.com/roburio/http-lwt-client">https://github.com/roburio/http-lwt-client</a>

##### CHANGES:

* use Status.is_redirection instead of a match on the polymorphic variant, to
  support unknown 3xx codes (such as 308) (roburio/http-lwt-client#11 @hannesm)
* add some yield in the main loop to allow concurrency with other fibers (issue
  roburio/http-lwt-client#12 by @kit-ty-kate, fixed in roburio/http-lwt-client#13 by @dinosaure)
* hurl: add a no-follow flag (@hannesm)


kit-ty-kate:
Thanks

PR #22125, opened 2022-09-19
Conjury 2.0, 2.0.1 and 2.1 are incompatible with OMake > 0.10.3.

jhwoodyatt:
See [CON-21](https://conjury.atlassian.net/browse/CON-21) for details.


jhwoodyatt:
The OCaml 5.0 build of Conjury 2.1 is failing because it requires OMake 0.10.3 (because 0.10.5 made a backward incompatible change to its standard library), and OMake 0.10.3 is not compatible with OCaml 5.0. Seems to me that the OMake 0.10.3 dependency on OCaml needs to be updated.

jhwoodyatt:
Gentle ping to @kit-ty-kate about my previous comment. Any advice on how to proceed? I think you may want to take this pull request anyway, because it does help to keep Conjury from failing because it's broken under OMake 0.10.5.

mseri:
Thanks and sorry for the wait.
I wanted to add a constraint on omake before merging

jhwoodyatt:
Thanks! I have a couple more updates incoming: another release of Conjury, followed by another release of Orsetto.

PR #22131, opened 2022-09-20
Release version 2.1.1 of conjury

jhwoodyatt:


jhwoodyatt:
And alas, it looks like `omake.0.10.5` is _also_ incompatible with `ocaml.5.0`. I filed an issue with `omake` for it.

packages/conjury/conjury.2.1.1/opam
```diff
@@ -0,0 +1,25 @@
...
+authors: "james woodyatt <jhw@conjury.org>"
+license: "BSD-2-Clause"
+tags: "org:conjury.org"
+homepage: "https://bitbucket.org/jhw/conjury/"
+bug-reports: "https://bitbucket.org/jhw/conjury/issues"
+depends: [
+  "base-unix"
+  "ocaml" { >= "4.08"}
+  "ocamlfind" { >= "1.7.3"}
+  "omake" { = "0.10.3"}
```
mseri, line 13:
```suggestion
  "omake" { = "0.10.5"}
```

jhwoodyatt: Ah. I think I see my mistake. I inadvertently changed the `>=` to `=` and I need to change it back. I think OMake 0.10.5 is still broken on OCaml 5.0 anyway.

jhwoodyatt:
And now the failure in CI is with OMake 0.10.5 (same as with 0.10.3).

kit-ty-kate:
Thanks!

PR #22139, opened 2022-09-21
Release caqti-driver-postgresql 1.9.1.

paurkedal:
We have fixed two bugs in the caqti-driver-postgresql sub-package which I think warrants a release:

  - PostgreSQL expects "=" in query values to be URI-encoded (paurkedal/ocaml-caqti#95).

  - The validation logic for PostgreSQL was missing a call to consume
    available input, meaning that broken pooled connections were reused.


mseri:

PR #22142, opened 2022-09-21
Package llvm.15.0.1

kit-ty-kate:
### `llvm.15.0.1`
The OCaml bindings distributed with LLVM
Note: LLVM should be installed first.

---
* Homepage: http://llvm.moe
* Source repo: git+https://github.com/kit-ty-kate/llvm-dune.git
* Bug tracker: http://llvm.org/bugs/

---

### DIFF
--- a/packages/llvm/llvm.15.0.1/opam
+++ b/packages/llvm/llvm.15.0.1/opam
@@ -7,7 +7,7 @@ authors: [
   "whitequark <whitequark@whitequark.org>"
   "The LLVM team"
 ]
-license: "Apache-2.0 with LLVM-exception"
+license: "Apache-2.0 WITH LLVM-exception"
 homepage: "http://llvm.moe"
 doc: "http://llvm.moe/ocaml"
 bug-reports: "http://llvm.org/bugs/"
--- a/packages/llvm/llvm.15.0.1/opam
+++ b/packages/llvm/llvm.15.0.1/opam
@@ -12,7 +12,7 @@ homepage: "http://llvm.moe"
 doc: "http://llvm.moe/ocaml"
 bug-reports: "http://llvm.org/bugs/"
 depends: [
-  "ocaml" {>= "4.00.0"}
+  "ocaml" {>= "4.00.0" & < "5.0.0"}
   "dune" {>= "2.7"}
   "ctypes" {>= "0.4"}
   "ctypes-foreign" {>= "0.4"}
--- a/packages/llvm/llvm.15.0.1/opam
+++ b/packages/llvm/llvm.15.0.1/opam
@@ -15,6 +15,7 @@ depends: [
   "ocaml" {>= "4.00.0"}
   "dune" {>= "2.7"}
   "ctypes" {>= "0.4"}
+  "ctypes-foreign" {>= "0.4"}
   "ounit" {with-test}
   "conf-llvm" {build & = version}
 ]
--- a/packages/llvm/llvm.15.0.1/opam
+++ b/packages/llvm/llvm.15.0.1/opam
@@ -1,0 +1,35 @@
+opam-version: "2.0"
+synopsis: "The OCaml bindings distributed with LLVM"
+description: "Note: LLVM should be installed first."
+maintainer: "Kate <kit.ty.kate@disroot.org>"
+authors: [
+  "Gordon Henriksen <gordonhenriksen@mac.com>"
+  "whitequark <whitequark@whitequark.org>"
+  "The LLVM team"
+]
+license: "Apache-2.0 with LLVM-exception"
+homepage: "http://llvm.moe"
+doc: "http://llvm.moe/ocaml"
+bug-reports: "http://llvm.org/bugs/"
+depends: [
+  "ocaml" {>= "4.00.0"}
+  "dune" {>= "2.7"}
+  "ctypes" {>= "0.4"}
+  "ounit" {with-test}
+  "conf-llvm" {build & = version}
+]
+build: [
+  ["./setup.sh" conf-llvm:config]
+  ["dune" "build" "--release" "-j" jobs]
+  ["rm" "%{name}%.install"]
+]
+install: ["./install.sh" prefix]
+dev-repo: "git+https://github.com/kit-ty-kate/llvm-dune.git"
+url {
+  src:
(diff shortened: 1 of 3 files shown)


Kakadu:
@kit-ty-kate Was the change of a license intentional?

packages/llvm/llvm.15.0.1/opam
```diff
@@ -0,0 +1,36 @@
...
+  "Gordon Henriksen <gordonhenriksen@mac.com>"
+  "whitequark <whitequark@whitequark.org>"
+  "The LLVM team"
+]
+license: "Apache-2.0 WITH LLVM-exception"
+homepage: "http://llvm.moe"
+doc: "http://llvm.moe/ocaml"
+bug-reports: "http://llvm.org/bugs/"
+depends: [
+  "ocaml" {>= "4.00.0"}
```
kit-ty-kate, line 15:
```suggestion
  "ocaml" {>= "4.00.0" & < "5.0.0"}
```

alan-j-hu:
I have been hired by the OCaml Software Foundation to ensure that the LLVM OCaml bindings continue working with OCaml 5 and the LLVM changes. @jberdine is supervising me, and he decided that my first action should be to backport the OCaml 5 fixes to previous releases and have the code downloaded from opam have the changes applied. I have a fork of LLVM where I have fixed versions of the 15 and 16 releases:

- https://github.com/alan-j-hu/llvm-project/tree/release/15.x+nnp
- https://github.com/alan-j-hu/llvm-project/tree/release/16.x+nnp

I would like to incorporate these fixes into the LLVM 15 and 16 packages to be published on opam. I see that for this PR, you have a repository for building the bindings with Dune which has the LLVM repository as a Git submodule. Would pointing the submodule URL to a fork be the best solution?

alan-j-hu:
At @jberdine's request, I've added changes to ease the typed to opaque pointer transition. In the C API, the opaque pointer equivalents of deprecated typed pointer functions have been suffixed with 2. For example, an old deprecated C API function is `LLVMConstGEP` and its replacement is `LLVMConstGEP2`. Upstream, the OCaml bindings for LLVM 15 have `const_gep` (old) and `const_gep2` (new), but the bindings for LLVM 16 just has `const_gep` (which now binds to `LLVMConstGEP2`. This makes the transition hard because there is no OCaml code using these functions that would compile on two consecutive LLVM versions. To fix this problem, I patched the LLVM 15 bindings to add a deprecation alert for `const_gep` in favor of `const_gep2`, then patched LLVM 16 to add `const_gep2` as a deprecated alias of `const_gep` (which therefore both bind `LLVMConstGEP2` in the C API). Right now, these changes are on separate branches. Do you think that adding this patch, which changes the API from the upstream bindings, are a good idea? (For example, it would cause the opam package's API to differ from the API of the version of the bindings in Linux package repositories, which use the upstream code.)

Kakadu:
Hey, @alan-j-hu 

It looks like original author of this PR is occupied in the real life. Did you consider creating a separate PR for LLVM 16 + your patch? (I think, it will be OK to skip release 15 for now). I could imagine this PR as 
1. A copy of previous opam recipe with updated tarball URL
2. An extra patch with your changes as part of PR too.

alan-j-hu:
So far, @jberdine has not replied to my email, so I decided to go ahead and experiment with your Dune additions. I made a fork at https://github.com/alan-j-hu/llvm-project/tree/release/15.x%2Bnnp%2Bdepr%2Bdune that's based on your work, but with a bunch of changes. I created `dune` files for each sublibrary instead of generating them all with a shell script. The main difference between your original work and my version is that I don't use virtual libraries. As a result, with your version someone could choose between static and dynamic LLVM libraries by depending on an implementation module, but with my version, people can't. I want to better understand the situation with static and dynamic linking. Do users generally want to be able to choose between static and dynamic linking to external libraries? Is having a virtual module and two implementation modules with the exact same code but different flags the only way to choose between them?

I didn't write the code to generate each backend module yet because I want to know how you feel about my version.

alan-j-hu:
I found [your Discuss post](https://discuss.ocaml.org/t/ask-for-suggestions-for-the-next-z3-package-release/8922/4) that goes over the same information you told me over the call, about how the old approach was hacky and so you wanted to move to Dune. Having the details down in writing helped me better understand what is going on. If I understand correctly:

- The "hack" to choose static linking is the `llvm.static` flag in the patched META file. This is what you can't replicate with Dune.
- Allowing users to choose between static and dynamic linking LLVM is still a requirement with the Dune version, so the virtual modules feature is necessary. So, my version would not be acceptable.

I feel that the appeal of using Dune is to have a standardized way of describing builds. Before the widespread adoption of Dune, OCaml packages would use a wide variety of build systems and scripts, but with the adoption of Dune, OCaml packages may have a more uniform build file format and filesystem organization. I was hoping that LLVM could use Dune in the standard way, i.e. by having files that look like

```
(library
 (name blah)
 (libraries ...)
```

I feel that having a script that generates Dune files twice for statically and dynamically linked versions and also copies the .ml files defeats the main appeal of Dune, the simplicity and uniformity, due to how unconventional it is. I also think that this Dune approach (i.e. generating `dune` files and copying the .ml file into static and shared directories) is uglier than the META file patch because the former heavily reorganizes the directory layout.

I tried finding a better way to allow a choice between static and dynamic linking, but could not find one. I feel that Dune is often opinionated, which is part of its appeal, but also a disadvantage here. My opinion is that Dune is not fit for the use case of building the LLVM bindings.

alan-j-hu:
I'm reading the install.sh file in the opam repository to better understand how the old build system works, and I'm confused by these lines:

- https://github.com/ocaml/opam-repository/blob/eb733d35a0a83a2635d25cd85e905661d145aead/packages/llvm/llvm.14.0.6/files/install.sh#L26

According to the CMake docs, `BUILD_SHARED_LIBS` changes the behavior of CMake's `add_library`, so wouldn't this option switch between building static or shared object files for the OCaml bindings, not whether to link static or shared to the pre-installed LLVM library?

- https://github.com/ocaml/opam-repository/blob/eb733d35a0a83a2635d25cd85e905661d145aead/packages/llvm/llvm.14.0.6/files/install.sh#L29

I thought that you need to pass `--link-static` or `--link-shared` when using `llvm-config --libs` because depending on whether the linking is static or shared, the command will output different flags?

If the purpose of the install.sh file is to link to static or shared object files for a pre-installed LLVM, is it doing the right thing?

PR #22149, opened 2022-09-22
Add package grpc-0.0.1

Nymphium:
This PR adds a new package grpc which is a wrapper of gRPC-Core with ctypes.
https://github.com/Nymphium/grpc-ocaml


raphael-proust:
The package build seems to be broken: some module is not found. Check the CI for details.

ansiwen:
I agree. Since this is a wrapper, how about `libgrpc` instead?

Nymphium:
I'll remake PR after fixing problems, thank you!

PR #9728, opened 2017-07-04
Fix gmp/zarith configuration on macOS.

dbuenzli:
In particular this allows to use again LDFLAGS and CFLAGS
environment variables to choose your location and and automatically
checks for the two locations where gmp will be found in 99% of
the cases.

Closes #3000 and #9389.

See [this comment](https://github.com/ocaml/opam-repository/issues/3000#issuecomment-312752454) for details.

PR #22150, opened 2022-09-23
[new release] malfunction (0.4)

stedolan:
Compiler back-end for functional languages, based on OCaml

- Project page: <a href="https://github.com/stedolan/malfunction">https://github.com/stedolan/malfunction</a>

##### CHANGES:

Support for OCaml 4.09 to 4.14
Recursive lazy bindings now supported

### DIFF
--- a/packages/malfunction/malfunction.0.4/opam
+++ b/packages/malfunction/malfunction.0.4/opam
@@ -13,7 +13,7 @@ depends: [
   "ocaml" {>= "4.04" & < "5.0.0"}
   "ocamlfind"
   "dune" {>= "1.2"}
-  "cppo" {build}
+  "cppo" {build & >= "1.1.0"}
   "omd" {with-test & < "2.0.0~"}
   "craml" {with-test}
   "zarith"
--- a/packages/malfunction/malfunction.0.4/opam
+++ b/packages/malfunction/malfunction.0.4/opam
@@ -1,0 +1,34 @@
+opam-version: "2.0"
+maintainer: "stephen.dolan@cl.cam.ac.uk"
+authors: ["Stephen Dolan"]
+homepage: "https://github.com/stedolan/malfunction"
+bug-reports: "https://github.com/stedolan/malfunction/issues"
+dev-repo: "git+https://github.com/stedolan/malfunction.git"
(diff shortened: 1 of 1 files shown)


packages/malfunction/malfunction.0.4/opam
```diff
@@ -0,0 +1,34 @@
...
+dev-repo: "git+https://github.com/stedolan/malfunction.git"
+license: "LGPL-2.0-or-later"
+build: [
+  [ "dune" "build" "-p" name "-j" jobs ]
+  [ "dune" "runtest" "-p" name "-j" jobs ] {with-test}
+]
+depends: [
+  "ocaml" {>= "4.04" & < "5.0.0"}
+  "ocamlfind"
+  "dune"
```
mseri, line 15:
```suggestion
  "dune" {>= "1.2"}
```
According to the dune-project file

packages/malfunction/malfunction.0.4/opam
```diff
@@ -0,0 +1,34 @@
...
+license: "LGPL-2.0-or-later"
+build: [
+  [ "dune" "build" "-p" name "-j" jobs ]
+  [ "dune" "runtest" "-p" name "-j" jobs ] {with-test}
+]
+depends: [
+  "ocaml" {>= "4.04" & < "5.0.0"}
+  "ocamlfind"
+  "dune" {>= "1.2"}
+  "cppo" {build}
```
kit-ty-kate, line 16: Uses `cppo -V`
```suggestion
  "cppo" {build & >= "1.1.0"}
```

stedolan:
> There are some issues with the tests on macOS:

This is due to ocamlopt producing a spurious warning when invoked by malfunction and linking with zarith (see #3000, #9728). The issue is marked as resolved, so I'm not sure why we're still seeing it. Perhaps malfunction invokes ocamlopt in an unusual way? I'll try and work it out if I can reproduce it on a mac, but since it all still works except for an extra line of warning-spam when using the CLI, it's not a high priority for me.

> and on x86_32:

Thanks, this is a bug! I had an off-by-one in the range of permissible shift counts, but it was working essentially by coincidence on the platforms I had tested. I'll close this and upload a v0.4.1 with the fix.

PR #22153, opened 2022-09-25
Add depexts for package xxhash

LasseBlaauwbroek:
Note: This is not my package

### DIFF
--- a/packages/conf-xxhash/conf-xxhash.1/opam
+++ b/packages/conf-xxhash/conf-xxhash.1/opam
@@ -12,6 +12,7 @@ depexts: [
   ["xxhash-devel"] {os-family = "rhel"}
   ["xxhash-devel"] {os-distribution = "fedora"}
   ["libxxhash-dev"] {os-family = "debian"}
+  ["libxxhash-dev"] {os-family = "ubuntu"}
   ["xxhash"] {os-family = "gentoo"}
   ["xxhash"] {os-distribution = "homebrew" & os = "macos"}
   ["xxhash"] {os-distribution = "macports" & os = "macos"}
--- a/packages/conf-xxhash/conf-xxhash.1/opam
+++ b/packages/conf-xxhash/conf-xxhash.1/opam
@@ -1,0 +1,30 @@
+opam-version: "2.0"
+maintainer: "Lasse Blaauwbroek"
+authors: [
+  "Yann Collet"
+]
+homepage: "http://www.xxhash.com/"
+license: "BSD-2-Clause"
+build: [["pkg-config" "libxxhash"]]
+depexts: [
+  ["xxhash-dev"] {os-family = "alpine"}
+  ["xxhash"] {os-family = "arch"}
+  ["xxhash-devel"] {os-family = "rhel"}
+  ["xxhash-devel"] {os-distribution = "fedora"}
+  ["libxxhash-dev"] {os-family = "debian"}
+  ["xxhash"] {os-family = "gentoo"}
+  ["xxhash"] {os-distribution = "homebrew" & os = "macos"}
+  ["xxhash"] {os-distribution = "macports" & os = "macos"}
+  ["xxhash-devel"] {os-family = "suse"}
+  ["xxhash"] {os = "freebsd"}
+  ["xxhash-devel"] {os-distribution = "ol"}
+]
+x-ci-accept-failures: [
+  "oraclelinux-7" "oraclelinux-8" # Requires epel-release
+]
+synopsis: "Virtual package relying on a xxhash system installation"
+description:
+  "This package can only install if the xxhash is installed on the system."
+depends: ["conf-pkg-config" {build}]
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+flags: conf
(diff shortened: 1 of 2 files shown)


LasseBlaauwbroek:
I added a `x-ci-accept-failures` for oraclelinux, but the ci is still red. Shouldn't this make ci green?

LasseBlaauwbroek:
Probably (but this is not my package, so I chose to be conservative). But a conf package should probably include a check that the dev headers are actually present? Is there an established guidance on how to do that?

packages/conf-xxhash/conf-xxhash.1/opam
```diff
@@ -0,0 +1,30 @@
...
+]
+homepage: "http://www.xxhash.com/"
+license: "BSD-2-Clause"
+build: [["pkg-config" "libxxhash"]]
+depexts: [
+  ["xxhash-dev"] {os-family = "alpine"}
+  ["xxhash"] {os-family = "arch"}
+  ["xxhash-devel"] {os-family = "rhel"}
+  ["xxhash-devel"] {os-distribution = "fedora"}
+  ["libxxhash-dev"] {os-family = "debian"}
```
kit-ty-kate, line 14: For ubuntu-based distributions
```suggestion
  ["libxxhash-dev"] {os-family = "debian"}
  ["libxxhash-dev"] {os-family = "ubuntu"}
```

kit-ty-kate (commented):
> I added a conf-xxhash package. It fails on Debian 10. Not entirely sure why, but I think because the xxhash version is too old.

Yeah that's ok. It's simply because Debian 10 does not ship a version that installs the .pc file: https://packages.debian.org/buster/arm64/libxxhash-dev/filelist

packages/xxhash/xxhash.0.1/opam
```diff
@@ -15,14 +15,12 @@ install: ["ocaml" "setup.ml" "-install"]
 remove: ["ocamlfind" "remove" "xxhash"]
 depends: [
   "ocaml"
-  "oasis" {build}
-  "ctypes"
+  "oasis" {>= "0.4.7" & build}
+  "ocamlbuild" {(> "0.9.0" | < "0.9.0") & build} # Does not work on 0.9.0
```
kit-ty-kate, line 19:
```suggestion
  "ocamlbuild" {!= "0.9.0" & build}
```

packages/xxhash/xxhash.0.1/opam
```diff
@@ -15,14 +15,12 @@ install: ["ocaml" "setup.ml" "-install"]
 remove: ["ocamlfind" "remove" "xxhash"]
 depends: [
   "ocaml"
-  "oasis" {build}
-  "ctypes"
+  "oasis" {>= "0.4.7" & build}
+  "ocamlbuild" {!= "0.9.0" & build}
+  "ctypes" {>= "0.6.0"}
```
kit-ty-kate, line 20: Do you have the logs for these new constraints for our record?

LasseBlaauwbroek: I used the CI to determine this. But I force pushed over old commits so I cannot find them any more. But in summary:
- OCamlbuild 0,9.0 does not work for unknown reasons. Version 0 and version 0.9.1 do work https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/6d7deffe78d9ee9d1c220cac83c9a82969174fc1/variant/compilers,4.03,xxhash.0.1,lower-bounds
- Ctypes 0.6.0 introduces `Unsigned.UInt.to_int64` needed by this package: https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/76f25ec3012a7834e3d10c40819fba6cda8cc389/variant/compilers,4.04,xxhash.0.1,lower-bounds
- I've already forgotten what the issue with older versions of Oasis was, but I think it may have been this log: https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/fcf9f161fc8c67d08665fbcff0dd3c3ff4ce4297/variant/compilers,4.02,xxhash.0.1,lower-bounds

kit-ty-kate:
```suggestion
  "oasis" {build}
  "ocamlbuild" {!= "0.9.0" & build}
  "ctypes"
```

kit-ty-kate (approved):

packages/xxhash/xxhash.0.1/opam
LasseBlaauwbroek: Pretty sure that the ctypes constraint is correct. The `integers` package was split off from `ctypes`, but for older versions it was still part of `ctypes`, and that is the case for 0.6.0

LasseBlaauwbroek:
@kit-ty-kate I don't understand why you removed the ctypes constraint. Unsigned.UInt.64 was clearly introduced here: https://github.com/ocamllabs/ocaml-ctypes/commit/29e65cf8d2c18f3e9e0e68dda6bc588dbad7c122 which was first shipped in 0.6.0.

LasseBlaauwbroek:
Same goes for Oasis. I tested these older versions and they clearly don't work. Why would you then remove those constraints.

kit-ty-kate:
It was to unlock this particular PR. The package was already in this state when you opened it and further fixes can be always be done later. Having several things done in one PR makes it hard to review

LasseBlaauwbroek:
Well, okay, but this is getting a bit bureacratic. I spent several hours nailing down the exact requirements for this package, and upon request I provided the logs showing the failures. I'm certainly not going to open another PR and go through the same exercise again.

kit-ty-kate:
That's ok, thanks for that. I've cherry-picked your commit to https://github.com/ocaml/opam-repository/pull/22166 and will try to dig the logs for oasis later.

PR #22159, opened 2022-09-26
Add a note if the installation of esperanto-cosmopolitan fails on linux

dinosaure:
The `esperanto-cosmopolitan` distribution tries to execute a Cosmopolitan executable to install/unzip the Cosmopolitan distribution. If the user has Wine and `binfmt_misc`, Wine completes `binfmt_misc` with a description of MS/DOS executable. However, Cosmopolitan binaries start with such header and, then, they are recognized as Wine binaries. Finally, our Cosmopolitan binary fails. This patch informs the user how to fix such issue by a script provived by the Cosmopolitan's author.


dra27:
Would it be better to wrap this logic into a `conf-` package? For example, the `conf` package could download the Cosmopolitan hello world example and ensure that it runs? On failure, it could even inspect the binfmt_misc configuration to try to identify known problems (for example, Mono on Ubuntu installs a binfmt_misc filter which gets in the way of of αpε as well).

dinosaure:
> Would it be better to wrap this logic into a conf- package? For example, the conf package could download the Cosmopolitan hello world example and ensure that it runs?

As you want 🙂 I will propose a `conf-cosmopolitan` package this afternoon.

dinosaure:
**NOTE**: I have to add `hello2.com` as an executable (so I set the permission to 0o755) even though the `opam-this` linter complains about it. I'm not sure how to resolve this in a reasonable way.

dinosaure:
I just fixed the Alpine support. It seems that the default shell on Alpine is `zsh` which has a bug about Cosmopolitan (fixed since `zsh.5.9.0`).

kit-ty-kate:
> I just fixed the Alpine support. It seems that the default shell on Alpine is `zsh` which has a bug about Cosmopolitan (fixed since `zsh.5.9.0`).

it is not zsh but simply `sh`. However when ran directly by opam, opam calls `Unix.create_process_env`, is that an issue for cosmopolitan?

kit-ty-kate:
> EDIT: It seems that the default shell of Alpine is `ash` (https://wiki.alpinelinux.org/wiki/Change_default_shell). As far as I can say, no Cosmopolitan's users noticed an error with it.

`ash` is the default `sh` on Alpine
```
/ # ls -l /bin/sh
lrwxrwxrwx    1 root     root            12 May 23 16:51 /bin/sh -> /bin/busybox
/ # ls -l /bin/ash 
lrwxrwxrwx    1 root     root            12 May 23 16:51 /bin/ash -> /bin/busybox
```

dinosaure:
Currently, the problem seems about the underlying call of `posix_spawn` with the Musl C library. The issue is so directly related to Cosmopolitan and/or Musl. We have different possibilities here:
1) we can, as we currently do, use `sh -c` which work everywhere
2) waiting a possible fix from one or the other side before merging this PR

WDYT?

packages/conf-cosmopolitan/conf-cosmopolitan.1/opam
```diff
@@ -0,0 +1,23 @@
+opam-version: "2.0"
+maintainer: "Romain Calascibetta <romain.calascibetta@gmail.com>"
+authors: "Romain Calascibetta <romain.calascibetta@gmail.com>"
+homepage: "https://github.com/dinosaure/esperanto"
+bug-reports: "https://github.com/dinosaure/esperanto/issues/"
+license: "MIT"
+tags: "org:mirage"
+dev-repo: "git+https://github.com/dinosaure/esperanto.git"
+build: [
+  [ "echo" "$SHELL" ]
```
kit-ty-kate, line 10:
```suggestion
```

kit-ty-kate:
Looks fine to be merge as is. No need to wait

dinosaure:
Thanks!

kit-ty-kate:
Sadly i made a mistake when reviewing this PR and didn't see it was adding a 256k binary in opam-repository. I had to remove the commit as it would cause issues to the opam infrastructure and makes the repository bigger for no reason.

Would you be able to reopen a PR with the binary fetched from an external source using the `extra-source` field?

PR #22168, opened 2022-09-26
Package ppx_tools.6.6

kit-ty-kate:
### `ppx_tools.6.6`
Tools for authors of ppx rewriters and other syntactic tools

---
* Homepage: https://github.com/ocaml-ppx/ppx_tools
* Source repo: git+https://github.com/ocaml-ppx/ppx_tools.git
* Bug tracker: https://github.com/ocaml-ppx/ppx_tools/issues

---


mseri:
I don't think the failures depend on this. 
There are a couple of packages that probably need updated bounds in the opam file and we need to recover archetype 1.3.3 and mparser 1.2.1 (failing with bad checksum), but this could be done separately

PR #22173, opened 2022-09-27
[New release] Tuareg 3.0.0

Chris00:
Tuareg 3.0.0


Chris00:
The check fails because Emacs is not being found.

mseri:
Thanks

PR #22176, opened 2022-09-27
Release mbr-format.1.0.0 and mirage-block-partition

reynir:
`mbr-format` lives at https://github.com/mirage/ocaml-mbr and `mirage-block-partition` lives at https://github.com/reynir/mirage-block-partition.

PR #22177, opened 2022-09-27
[new release] oneffs (0.1.0)

reynir:
One-file filesystem is a filesystem for storing a single unnamed file

- Project page: <a href="https://git.robur.io/reynir/oneffs">https://git.robur.io/reynir/oneffs</a>

##### CHANGES:

* Initial release


reynir:
> Should I mark it as "don't merge" given your comment on the other PR?

Thank you for asking. After a discussion I have decided to go ahead with releasing it as is (also #22176). They are new packages and an old package I don't think sees any or much use today. If anyone should complain I can do a point release.

PR #22183, opened 2022-09-28
[new release] wayland (1.1)

talex5:
Pure OCaml Wayland protocol library

- Project page: <a href="https://github.com/talex5/ocaml-wayland">https://github.com/talex5/ocaml-wayland</a>
- Documentation: <a href="https://talex5.github.io/ocaml-wayland/">https://talex5.github.io/ocaml-wayland/</a>

##### CHANGES:

- Add update.sh script and update protocols (@MisterDA talex5/ocaml-wayland#25, @talex5 talex5/ocaml-wayland#33).

- Improve proxy module documentation (@talex5 talex5/ocaml-wayland#31).

- Improve "No such object" error (@talex5 talex5/ocaml-wayland#28).

- Make `Handler.cast_version` more general (@talex5 talex5/ocaml-wayland#27).
  Avoids the need to cast to `Handler.t` before using `cast_version`.

- Update to cmdliner.1.1.1 (@MisterDA talex5/ocaml-wayland#26).

### DIFF
--- a/packages/wayland/wayland.1.1/opam
+++ b/packages/wayland/wayland.1.1/opam
@@ -1,0 +1,46 @@
+opam-version: "2.0"
+synopsis: "Pure OCaml Wayland protocol library"
+description:
+  "Wayland is a communications protocol intended for use between processes on a single computer. It is mainly used by graphical applications (clients) to talk to display servers, but nothing about the protocol is specific to graphics and it could be used for other things. This library can be used to write Wayland clients, servers and proxies."
+maintainer: ["talex5@gmail.com"]
+authors: ["talex5@gmail.com"]
+license: "Apache-2.0 (excluding schema files)"
+homepage: "https://github.com/talex5/ocaml-wayland"
+doc: "https://talex5.github.io/ocaml-wayland/"
+bug-reports: "https://github.com/talex5/ocaml-wayland/issues"
+depends: [
+  "dune" {>= "2.8"}
+  "alcotest-lwt" {>= "1.2.3" & with-test}
+  "ocaml" {>= "4.08.0"}
+  "xmlm" {>= "1.3.0"}
+  "lwt" {>= "5.4.0"}
+  "logs" {>= "0.7.0"}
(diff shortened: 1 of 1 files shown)


dinosaure:
The `excluding schema files` does not respect the SPDX format of license. What is this exception about? Could you put this exception in the distribution and consider the project under the Apache 2.0 license only?

talex5:
It's the same license as the previous versions in opam-repository. The XML schema files describing the protocols aren't mine, so I can't relicense them (but they're all permissive BSD-type licenses).

mseri:
Do you know the license of the schema files?
We could have
```
license: [
  "Apache-2.0" # library excluding schema files
  "whatever" # schema files
]
```
or, maybe better,
````
license: [
  "Apache-2.0 (excluding schema files)";
  "whatever (schema files)"
]
```

mseri:
If there are too many involved, we can merge as is, like in the previous versions

packages/wayland/wayland.1.1/opam
```diff
@@ -0,0 +1,46 @@
+opam-version: "2.0"
+synopsis: "Pure OCaml Wayland protocol library"
+description:
+  "Wayland is a communications protocol intended for use between processes on a single computer. It is mainly used by graphical applications (clients) to talk to display servers, but nothing about the protocol is specific to graphics and it could be used for other things. This library can be used to write Wayland clients, servers and proxies."
+maintainer: ["talex5@gmail.com"]
+authors: ["talex5@gmail.com"]
+license: "Apache-2.0 (excluding schema files)"
```
kit-ty-kate, line 7: Let's try that for now (that's SPDX compatible)
```suggestion
license: "Apache-2.0 AND LicenseRef-various-licenses-for-the-schema-files"
```

kit-ty-kate:
Thanks! Could you return the fix to the opam file upstream (or equivalent).

*Side note: the full SPDX syntax is only supported with opam >= 2.2.0 (master), in case you're wondering why `opam lint` still complains*

PR #22204, opened 2022-10-04
Package tezos-rust-libs.1.2

vapourismo:
Package `tezos-rust-libs` version 1.2.

/cc @pirbo @dannywillems @emturner


vapourismo:
The `opam-ci` appears to fail because it runs rustc 1.56.1, and this package requires at least 1.59.

packages/tezos-rust-libs/tezos-rust-libs.1.2/opam
```diff
@@ -0,0 +1,36 @@
+opam-version: "2.0"
+synopsis: "Tezos: all rust dependencies and their dependencies"
+maintainer: "contact@tezos.com"
+authors: "Tezos devteam"
+license: "multiple"
+homepage: "https://www.tezos.com/"
+bug-reports: "https://gitlab.com/tezos/tezos-rust-libs/issues"
+depends: ["conf-rust"]
```
raphael-proust, line 8: I think we need to add some constraints on the os/distributions/versions. Probably something like

```
available: [
  (os != "debian" | os-version > "10")
  | (os != "alpine" | os-version > "3.15")
  | …
]
```

I'm not entirely sure of the exact entry that needs to be added.
- Not sure what `os`, `os-family`, `os-distribution` combinations to use.
- Not entirely sure how the "testing" and "unsatble" versions of debians work, I assume there'll be a special case in the comparison function because that seems like something that would have been useful to other packages.
- etc.

I'd recommend you add those constraints locally based on the result of the CI, especially the "distributions" section. Then lint your opam files locally to make sure that it's parsable. And then push an amended version.

raphael-proust: Also, you can submit a `conf-rust-2021` package (see `conf-rust-2018` package for an example) and then you can depend on this versioned one instead. That might actually be a better place to set the os constraints!

vapourismo: I added the `conf-rust-2021` package, but CI looks only marginally greener.

packages/conf-rust-2021/conf-rust-2021.1/opam
```diff
@@ -0,0 +1,34 @@
...
+build: [
+  ["cargo" "--version"]
+  ["rustc" "--edition" "2021" "test.rs"]
+]
+depexts: [
+  ["cargo"] {os-distribution = "centos" & os-version >= "8"}
+  ["cargo"] {os-distribution = "ol" & os-version >= "8"}
+  ["cargo"] {os-distribution = "fedora"}
+  ["cargo"] {os-family = "suse"}
+  ["cargo"] {os-family = "debian"}
```
raphael-proust, line 20: This is not available on all debians. I think this line should be

```suggestion
  ["cargo"] {os-family = "debian" & os-version > "10"}
```

packages/conf-rust-2021/conf-rust-2021.1/opam
```diff
@@ -0,0 +1,34 @@
...
+  ["cargo"] {os-family = "ubuntu"}
+  ["cargo" "rustc"] {os-distribution = "nixos"}
+  ["cargo"] {os-family = "alpine"}
```
raphael-proust, line 23:
```suggestion
  ["cargo"] {os-family = "alpine" & os-version > "3.15"}
```

packages/conf-rust-2021/conf-rust-2021.1/opam
```diff
@@ -0,0 +1,34 @@
...
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: "MIT"
+extra-files: ["test.rs" "sha256=536e506bb90914c243a12b397b9a998f85ae2cbd9ba02dfd03a9e155ca5ca0f4"]
+build: [
+  ["cargo" "--version"]
+  ["rustc" "--edition" "2021" "test.rs"]
+]
+depexts: [
+  ["cargo"] {os-distribution = "centos" & os-version >= "8"}
+  ["cargo"] {os-distribution = "ol" & os-version >= "8"}
```
raphael-proust, line 17:
```suggestion
  ["cargo"] {os-distribution = "ol" & os-version > "8"}
```

although Idk what version of cargo/rust is avialable in what version of oraclelinux so you need to check the specifics.

packages/tezos-rust-libs/tezos-rust-libs.1.2/opam
raphael-proust: I think the errors are better though, so that's already something.

Mostly, I think that the constraints on the `conf-rust-2021` need to be tighter. See other thread.

packages/conf-rust-2021/conf-rust-2021.1/opam
vapourismo: I am not sure about these. This package [compiles fine for Alpine 3.15](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/a5fc3b58a46d47eb8a9e75ad2ac3d03f0819e565/variant/distributions,alpine-3.15,conf-rust-2021.1). However, the change you're proposing means that no depext will be suggested for Alpine prior to 3.16. Is that what you want to achieve?

raphael-proust: If the package is available on 3.15 then my suggestion is wrong indeed.

And indeed, looking at the errors, it's a different one for 3.14 and 3.15:

> alpine-3.14
> 
>     [conf-rust-2021.1 (failed: edition 2021 is unstable and only available with -Z unstable-options.)](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/a5fc3b58a46d47eb8a9e75ad2ac3d03f0819e565/variant/distributions,alpine-3.14,conf-rust-2021.1)
>     [tezos-rust-libs.1.2 (failed: edition 2021 is unstable and only available with -Z unstable-options.)](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/a5fc3b58a46d47eb8a9e75ad2ac3d03f0819e565/variant/distributions,alpine-3.14,tezos-rust-libs.1.2)
> 
> alpine-3.15
> 
>     [tezos-rust-libs.1.2 (failed: package `corosensei v0.1.3` cannot be built because it requires rustc 1.59.0 or newer, while the currently active rustc version is 1.56.1)](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/a5fc3b58a46d47eb8a9e75ad2ac3d03f0819e565/variant/distributions,alpine-3.15,tezos-rust-libs.1.2)

So it should be `>= "3.15"`

Same remark for `"ol"`.

packages/conf-rust-2021/conf-rust-2021.1/opam
raphael-proust: This suggestion is wrong. See thread about alpine.

packages/conf-rust-2021/conf-rust-2021.1/opam
vapourismo: I am not sure that'll do much good though. The default Rust version that ships with Debian 11 (bullseye) is also too old (1.48). However, if one were to amend the constraint to `>= 12`, that means no `cargo` depext is suggested at all.

vapourismo: To put into perspective how the error looks: https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/edb4003792ecd6df75f29618ab0d3d63aaac66d1/variant/distributions,debian-10,conf-rust-2021.1

packages/conf-rust-2021/conf-rust-2021.1/opam
raphael-proust: You might need to use an additional field in the `opam` file: `available:`. That additional field basically makes the package available/not for different platforms/constraints. That field should be used if e.g., the underlying system package cannot be installed (automatically by opam? at all?) on the specified platform.

I was trying to suggest fixes so that the package is simply not available on those platform that it is incompatible with. But I might have gotten wrong advice. The `available:` field is probably more important here.

I'm not very knowledgeable on `conf-`/system packages so we might want some additional info from maintainers.

packages/conf-rust-2021/conf-rust-2021.1/opam
```diff
@@ -0,0 +1,34 @@
...
+  ["cargo"] {os-family = "debian" & os-version >= "12"}
+  ["cargo"] {os-family = "ubuntu"}
+  ["cargo" "rustc"] {os-distribution = "nixos"}
+  ["cargo"] {os-family = "alpine" & os-version >= "3.15"}
+  ["rust"] {os-family = "arch"}
+  ["rust"] {os = "macos" & os-distribution = "homebrew"}
+]
+x-ci-accept-failures: [
+  "centos-7" # does not have cargo by default
+  "oraclelinux-7" # does not have cargo by default
```
kit-ty-kate, line 29:
```suggestion
  "oraclelinux-7" # does not have cargo by default
  "debian-10" # rust 2021 is not in the default repository (version is too old)
  "debian-11" # rust 2021 is not in the default repository (version is too old)
```

vapourismo:
> The `opam-ci` appears to fail because it runs rustc 1.56.1, and this package requires at least 1.59.

I applied @kit-ty-kate's suggestion to both packages. Hopefully that satisfies the package building part of the CI now.

avsm:
Does this package definitely work with older versions of tezos-sapling?  A number of those have no upper bounds on tezos-rust-libs:

```
/Users/avsm/src/git/ocaml/opam-repository/packages/tezos-sapling/tezos-sapling.11.1/opam:  "tezos-rust-libs" { >= "1.1" }
/Users/avsm/src/git/ocaml/opam-repository/packages/tezos-sapling/tezos-sapling.11.0/opam:  "tezos-rust-libs" { >= "1.1" }
/Users/avsm/src/git/ocaml/opam-repository/packages/tezos-sapling/tezos-sapling.10.2/opam:  "tezos-rust-libs" { >= "1.1" }
/Users/avsm/src/git/ocaml/opam-repository/packages/tezos-sapling/tezos-sapling.12.0/opam:  "tezos-rust-libs" { >= "1.1" }
/Users/avsm/src/git/ocaml/opam-repository/packages/tezos-sapling/tezos-sapling.13.0/opam:  "tezos-rust-libs" { = "1.1" }
/Users/avsm/src/git/ocaml/opam-repository/packages/tezos-sapling/tezos-sapling.13.0/opam:  "tezos-rust-libs"
/Users/avsm/src/git/ocaml/opam-repository/packages/tezos-sapling/tezos-sapling.14.0/opam:  "tezos-rust-libs" { = "1.1" }
/Users/avsm/src/git/ocaml/opam-repository/packages/tezos-sapling/tezos-sapling.14.0/opam:  "tezos-rust-libs"
/Users/avsm/src/git/ocaml/opam-repository/packages/tezos-sapling/tezos-sapling.12.3/opam:  "tezos-rust-libs" { >= "1.1" }
```

avsm:
@vapourismo note that the ci-accept-failures are a workaround, but not a solution. Ideally, we'd like an opam-repo PR that will allow cargo/rust to be available at the right version for the package on all distros.  Bear in mind that the CI replicates the end-user experience, and if you simply disable the failure in CI you do not remove user errors that will be encountered.  Much better to define a depext for Cargo that installs the right version of Rust required.

vapourismo:
> @vapourismo note that the ci-accept-failures are a workaround, but not a solution. Ideally, we'd like an opam-repo PR that will allow cargo/rust to be available at the right version for the package on all distros.  Bear in mind that the CI replicates the end-user experience, and if you simply disable the failure in CI you do not remove user errors that will be encountered.  Much better to define a depext for Cargo that installs the right version of Rust required.

Mostly agree, but it is not obvious to me how I can amend the CI reflect a modern Rust user experience. However, irrespective of the CI, I don't see a good way to convey Rust version constraints through the Opam format - this is ultimately the problem at hand.
The end-user experience you mention is super niche in respect to this package.

Maybe we can talk about this in more depth if you visit us at Trili again :)

vapourismo:
> Does this package definitely work with older versions of tezos-sapling?  A number of those have no upper bounds on tezos-rust-libs:
> 
> 
> 
> ```
> 
> /Users/avsm/src/git/ocaml/opam-repository/packages/tezos-sapling/tezos-sapling.11.1/opam:  "tezos-rust-libs" { >= "1.1" }
> 
> /Users/avsm/src/git/ocaml/opam-repository/packages/tezos-sapling/tezos-sapling.11.0/opam:  "tezos-rust-libs" { >= "1.1" }
> 
> /Users/avsm/src/git/ocaml/opam-repository/packages/tezos-sapling/tezos-sapling.10.2/opam:  "tezos-rust-libs" { >= "1.1" }
> 
> /Users/avsm/src/git/ocaml/opam-repository/packages/tezos-sapling/tezos-sapling.12.0/opam:  "tezos-rust-libs" { >= "1.1" }
> 
> /Users/avsm/src/git/ocaml/opam-repository/packages/tezos-sapling/tezos-sapling.13.0/opam:  "tezos-rust-libs" { = "1.1" }
> 
> /Users/avsm/src/git/ocaml/opam-repository/packages/tezos-sapling/tezos-sapling.13.0/opam:  "tezos-rust-libs"
> 
> /Users/avsm/src/git/ocaml/opam-repository/packages/tezos-sapling/tezos-sapling.14.0/opam:  "tezos-rust-libs" { = "1.1" }
> 
> /Users/avsm/src/git/ocaml/opam-repository/packages/tezos-sapling/tezos-sapling.14.0/opam:  "tezos-rust-libs"
> 
> /Users/avsm/src/git/ocaml/opam-repository/packages/tezos-sapling/tezos-sapling.12.3/opam:  "tezos-rust-libs" { >= "1.1" }
> 
> ```

It will once the Tezos release cycle is completed. This will also update the constraint on that particular package.

avsm:
> It will once the Tezos release cycle is completed. This will also update the constraint on that particular package.

I can't parse this at all.  If this PR is merged, it can result in this package being selected on those older tezos-sapling packages. We don't know if it actually works with them since there's no CI.  What's the Tezos release cycle got to do with already-released versions of tezos-sapling?

Regarding the rust mapping to opam package, it seems to be a matter of creating an opam package per version of Rust that you want supported.  Has anyone tried that?

vapourismo:
> I can't parse this at all.  If this PR is merged, it can result in this package being selected on those older tezos-sapling packages. We don't know if it actually works with them since there's no CI.  What's the Tezos release cycle got to do with already-released versions of tezos-sapling?

I misunderstood what you meant. The 1.2 release is a minor version bump from 1.1 hence backwards compatible. The exact-version constraint for 1.2 will apply to new versions of Tezos/Octez. Romain, the Octez release manager, might be able to help make the older version constraints tighter if needed. 

> Regarding the rust mapping to opam package, it seems to be a matter of creating an opam package per version of Rust that you want supported.  Has anyone tried that?

Unfortunately it can't work exactly like that. You can switch to a minimum-version-based approach, which avoids dependency hell and retains composability but requires a check on every package that would depend on such conf package in order to ensure it does so using a `>=` constraint. However, you would push the CI workaround to those packages roughly times the number of Rust versions. I believe this approach has virtually no upsides apart from a slightly nicer error message for when the user doesn't have the right Rust version available.

kit-ty-kate:
Looks good to me. Thanks!

PR #22213, opened 2022-10-05
conf-ppl: add fedora and suse

eponier:
Hope I get the names right. There seems to be no package for alpine.

### DIFF
--- a/packages/conf-ppl/conf-ppl.1/opam
+++ b/packages/conf-ppl/conf-ppl.1/opam
@@ -16,8 +16,8 @@ depexts: [
   ["ppl"] {os = "freebsd"}
   ["ppl"] {os = "openbsd"}
   ["ppl"] {os = "macos" & os-distribution = "homebrew"}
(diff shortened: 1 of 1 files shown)


packages/conf-ppl/conf-ppl.1/opam
```diff
@@ -16,6 +16,8 @@ depexts: [
   ["ppl"] {os = "freebsd"}
   ["ppl"] {os = "openbsd"}
   ["ppl"] {os = "macos" & os-distribution = "homebrew"}
+  ["ppl"] {os-distribution = "fedora"}
+  ["ppl"] {os-family = "suse"}
```
kit-ty-kate, line 20:
```suggestion
  ["ppl-devel"] {os-distribution = "fedora"}
  ["ppl-devel"] {os-family = "suse"}
```

eponier:
Not sure there is a package for these platforms, but I don't manage to find a definitive answer on the Internet.

kit-ty-kate:
Thanks! Sorry for the wait

PR #22215, opened 2022-10-05
release odds.1.1, odds.1.2

raphael-proust:
1.1: use dune to build (to support monorepo)
1.2: bump ocaml version dependency and use effects


kit-ty-kate:
Thanks

PR #22218, opened 2022-10-06
LablGtk 2.18.13

garrigue:
Make LablGtk2 compatible with OCaml 5.0.

### DIFF
--- a/packages/lablgtk/lablgtk.2.18.13/files/lablgldir.patch
+++ b/packages/lablgtk/lablgtk.2.18.13/files/lablgldir.patch
@@ -1,0 +1,13 @@
+diff --git a/configure b/configure
+index 177e9547..f8a56fe0 100755
+--- a/configure
++++ b/configure
+@@ -4925,7 +4925,7 @@ printf %s "checking lablGL directory... " >&6; }
+       cat > conftest.ml << EOF
+       open Raw
+ EOF
+-      if $CAMLC -c -I "${LABLGLDIR:=+lablGL}" conftest.ml > /dev/null 2>&1 ; then
++      if $CAMLC -c -I "$LABLGLDIR" conftest.ml > /dev/null 2>&1 ; then
+         { printf "%s\n" "$as_me:${as_lineno-$LINENO}: result: $LABLGLDIR" >&5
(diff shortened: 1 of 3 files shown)


packages/lablgtk/lablgtk.2.18.13/files/lablgldir.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/configure b/configure
+index 177e9547..f8a56fe0 100755
+--- a/configure
++++ b/configure
+@@ -4925,7 +4925,7 @@ printf %s "checking lablGL directory... " >&6; }
+       cat > conftest.ml << EOF
+       open Raw
+ EOF
+-      if $CAMLC -c -I "${LABLGLDIR:=+lablGL}" conftest.ml > /dev/null 2>&1 ; then
++      if $CAMLC -c -I "$LABLGLDIR" conftest.ml > /dev/null 2>&1 ; then
```
kit-ty-kate, line 10: Is there a reason why this patch is not upstream?

packages/lablgtk/lablgtk.2.18.13/opam
```diff
@@ -0,0 +1,48 @@
...
+]
+depopts: [
+  "conf-gtksourceview"
+  "conf-gnomecanvas"
+  "conf-glade"
+  "lablgl"
+]
+conflicts: [
+  "base-domains"
+]
```
kit-ty-kate, line 29: If this release is really adding OCaml 5.0 support then this should be safe to remove
```suggestion
```

packages/lablgtk/lablgtk.2.18.13/files/lablgldir.patch
garrigue: Originally, it was because the original form was more comfortable for non-opam install.
Will think about that for the next release.

packages/lablgtk/lablgtk.2.18.13/opam
garrigue: Thank you for the fix. Done.

kit-ty-kate:
Thanks!

PR #22222, opened 2022-10-07
Add Frama-Clang package

vprevosto:
Introducing at last an opam package for the Frama-Clang plug-in of Frama-C

### DIFF
--- a/packages/frama-clang/frama-clang.0.0.13/opam
+++ b/packages/frama-clang/frama-clang.0.0.13/opam
@@ -1,0 +1,44 @@
+opam-version: "2.0"
+synopsis: "Frama-Clang, Frama-C's C++ front-end"
+description:"""
+Frama-Clang is a plug-in of Frama-C that allows parsing C++ files.
+It is based on clang/LLVM
+"""
+maintainer: "virgile.prevosto@cea.fr"
+authors: [
+  "David Cok"
+  "Virgile Prevosto"
+  "Franck Védrine"
+]
+homepage: "https://frama-c.com/frama-clang.html"
+license: "LGPL-2.1-only"
+dev-repo: "git+https://git.frama-c.com/pub/frama-clang.git"
+bug-reports: "https://git.frama-c.com/pub/frama-clang/-/issues"
+tags: [
+  "Frama-C"
+  "C++"
+]
+
+url {
+  src: "https://frama-c.com/download/frama-clang-0.0.13.tar.gz"
+  checksum: [ "sha512=d4fbfc170b43e7486fd6920363fe5ed73cfd6e1d537e6f64790f5cf9384e6d0c73eb626ee8c79d5492ffc559eec8ad649ed6ed2e5caedb519ab74c781f34f6ff" ]
+}
+
+build: [
+  ["./configure"]
+  [make "-j%{jobs}%"]
+]
+
+install: [
+  [make "install"]
+]
+
+depends: [
+  "ocaml" { >= "4.08.1" }
+  "frama-c" { >= "25.0" & < "26.0~" }
+  "camlp5"
+  "camlp-streams"
+  "conf-llvm" { >= "10.0.0" }
+  "conf-libclang" { >= "10" }
+  "conf-cmake"
+]
--- a/packages/frama-clang/frama-clang.0.0.13/opam
+++ b/packages/frama-clang/frama-clang.0.0.13/opam
@@ -25,7 +25,7 @@ url {
 }
 
 build: [
-  ["sed" "-i" "-e" "s|llvm-config|%{conf-llvm:config}%|g" "configure"]
+  ["sed" "-i.bak" "-e" "s|llvm-config|%{conf-llvm:config}%|g" "configure"]
   ["./configure"]
   [make "-j%{jobs}%"]
 ]
--- a/packages/frama-clang/frama-clang.0.0.13/opam
+++ b/packages/frama-clang/frama-clang.0.0.13/opam
@@ -25,6 +25,7 @@ url {
 }
 
 build: [
+  ["sed" "-i" "-e" "s|llvm-config|%{conf-llvm:config}%|g" "configure"]
   ["./configure"]
   [make "-j%{jobs}%"]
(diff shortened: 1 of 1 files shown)


vprevosto:
In order to have the package be installed in the debian-10 docker image, I have manually installed `libclang-cpp13-dev` and `libclang-13-dev` @thierry-martinez I was wondering why `conf-libclang.13`  only installs the latter for Ubuntu 22.04 and only `libclang-dev` for other distributions of the Debian family? [side question: would it make sense to include the former as well?]

packages/frama-clang/frama-clang.0.0.13/opam
```diff
@@ -0,0 +1,46 @@
...
+  "C++"
+]
+
+url {
+  src: "https://frama-c.com/download/frama-clang-0.0.13.tar.gz"
+  checksum: [ "sha512=d4fbfc170b43e7486fd6920363fe5ed73cfd6e1d537e6f64790f5cf9384e6d0c73eb626ee8c79d5492ffc559eec8ad649ed6ed2e5caedb519ab74c781f34f6ff" ]
+}
+
+build: [
+  ["sed" "-i" "-e" "s|llvm-config|%{conf-llvm:config}%|g" "configure"]
```
kit-ty-kate, line 28: `sed -i` works differently one GNU, Busybox (Alpine) and macOS. Only `-i.<suffix>` is supported on all platforms
```suggestion
  ["sed" "-i.bak" "-e" "s|llvm-config|%{conf-llvm:config}%|g" "configure"]
```

raphael-proust:
> In order to have the package be installed in the debian-10 docker image, I have manually installed `libclang-cpp13-dev`

You probably need to create a `conf-libclang-cpp13` package and add a dependency to it. Alternatively, you could include the clang-cpp13-dev library to the existing `conf-libclang` but that would make coarser packages and you should check with @thierry-martinez  about that.

packages/frama-clang/frama-clang.0.0.13/opam
```diff
@@ -0,0 +1,46 @@
...
+  [make "install"]
+]
+
+depends: [
+  "ocaml" { >= "4.08.1" }
+  "frama-c" { >= "25.0" & < "26.0~" }
+  "camlp5"
+  "camlp-streams"
+  "conf-llvm" { >= "10.0.0" }
+  "conf-libclang" { >= "10" }
```
kit-ty-kate, line 43: older versions of this package is not consistent. For example on debian stable, `conf-libclang.15` will install `libclang-dev` which is libclang 11 instead of 15. I'm not sure how much of a problem it is but it's worth mentioning.

packages/frama-clang/frama-clang.0.0.13/opam
vprevosto: You're right, thanks for reminding me about sed's portability issues.

packages/frama-clang/frama-clang.0.0.13/opam
vprevosto: Actually, the main point is to have the same version than the llvm library, but  as far as I can tell from reading the opam manual, this is not really possible to express with package formulas

kit-ty-kate: It is possible (see the conf-llvm package) but rather annoying to maintain

vprevosto:
> > In order to have the package be installed in the debian-10 docker image, I have manually installed `libclang-cpp13-dev`
> 
> You probably need to create a `conf-libclang-cpp13` package and add a dependency to it. Alternatively, you could include the clang-cpp13-dev library to the existing `conf-libclang` but that would make coarser packages and you should check with @thierry-martinez about that.

Well, I can't create a `conf-libclang-cpp13`. At best `conf-libclang-cpp.13`, with the same versioning issues as libclang itself.  Moreover, this will be very specific to debian and its derivatives (arch installs everything in a single package for instance, and this is the distribution I use daily. I'm not extremely familiar with debian, and their packaging policy with respect to llvm/clang seems quite hard to follow from the outside).

raphael-proust:
> Moreover, this will be very specific to debian and its derivatives (arch installs everything in a single package for instance, and this is the distribution I use daily

In that case I think that you could modify the `conf-libclang`'s `debian` entry. Currently it has these  `depexts` packages: `["libclang-dev" "llvm-dev"]` and it could have the cpp one.

packages/frama-clang/frama-clang.0.0.13/opam
vprevosto: Sorry I meant that ideally the package formula should check that `conf-llvm` and `conf-libclang` share the same major version number. But you're right, the `depexts` in the various `conf-libclang.*/opam` are a bit odd. Furthermore, the `configure.sh` of `conf-libclang` is really testing the C bindings of clang. Maybe the suggestion of @raphael-proust to create a `conf-libclang-cpp` is the way to go. I'll give it a try.

thierry-martinez:
Sorry for answering so late: for all version `V`, `conf-libclang.V` ensures that `libclang` is installed with version at most _V_. The situation depends on the OS: on some OSes, the `libclang-dev` default package is more recent than `V` (typically, on recent Ubuntus), so `conf-libclang.V` installs `libclang-V-dev`; on other OSes, the `libclang-dev` package is older than `V` (typically, on Debian stable), so `conf-libclang.V` installs `libclang-dev` to have the last version distributed by the OS. (Of course, this is supposed to be adapted to follow distribution updates; I don't know a perfect solution.)
This behavior is motivated by the fact that a given version of `clangml` works with all versions of `libclang` up to a given upperbound.

raphael-proust:
I restarted the debian job now that #22442 has been merged.

vprevosto:
Please accept my apologies for not having followed up with this PR as I should have (especially after the merge of #22694).

Anyway, the point of this PR is rather moot right now, as two versions of Frama-C have been released since, and Frama-Clang 0.0.13 won't compile with Frama-C 27. I'm closing this PR and will focus (hoping to have more time to devote on that than last year) on making #24344 fit for merging.

PR #22224, opened 2022-10-08
Update conf-xkbcommon

ddeclerck:
Summary of changes :
- Windows: add the Cygwin package
- Debian & Ubuntu: add `os-family = "ubuntu"` because some (many ?) Ubuntu derivatives (Linux Mint, elementaryOS, ...) only list Ubuntu in `os-release:ID_LIKE` (or list both, but Ubuntu first)
- Mandriva and derivatives: add the package for this family
- Arch Linux: add `os-family = "archlinux"` in addition of `os-family = "arch"` (I have seen both in Arch and its derivatives)
- Gentoo and derivatives: add the package for this family
- *BSD: use `os-family = "bsd"` (all BSD variants seem to use the same name for this package)
- MacOS: add the MacPorts package

### DIFF
--- a/packages/conf-xkbcommon/conf-xkbcommon.1/opam
+++ b/packages/conf-xkbcommon/conf-xkbcommon.1/opam
@@ -17,7 +17,7 @@ depexts: [
   ["libxkbcommon-dev"] {os-family = "alpine"}
   ["libxkbcommon"] {os-family = "arch" | os-family = "archlinux"}
   ["x11-libs/libxkbcommon"] {os-family = "gentoo"}
-  ["libxkbcommon"] {os-family = "bsd" & os != "openbsd"}
+  ["libxkbcommon"] {os = "freebsd" | os = "dragonfly" | os = "netbsd"}
   ["xkbcommon"] {os = "openbsd"}
   ["libxkbcommon"] {os = "macos" & os-distribution = "homebrew"}
   ["libxkbcommon"] {os = "macos" & os-distribution = "macports"}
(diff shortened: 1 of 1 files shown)


packages/conf-xkbcommon/conf-xkbcommon.1/opam
```diff
@@ -9,15 +9,18 @@ depends: [
...
+  ["libxkbcommon-devel"] {os = "cygwin" | os-distribution = "cygwinports"}
+  ["libxkbcommon-dev"] {os-family = "debian" | os-family = "ubuntu"}
+  ["libxkbcommon-devel"] {os-family = "fedora" | os-family = "rhel" | os-family = "centos"}
+  ["libxkbcommon-devel"] {os-family = "mandriva" | os-family = "openmandriva" | os-family = "mageia"}
+  ["libxkbcommon-devel"] {os-family = "suse" | os-family = "opensuse" | os-family = "sles"}
+  ["libxkbcommon-dev"] {os-family = "alpine"}
+  ["libxkbcommon"] {os-family = "arch" | os-family = "archlinux"}
+  ["x11-libs/libxkbcommon"] {os-family = "gentoo"}
+  ["libxkbcommon"] {os-family = "bsd" & os != "openbsd"}
+  ["xkbcommon"] {os = "openbsd"}
```
kit-ty-kate, line 21: i prefer be more explicit about what is supported (white-list vs black-list) in this case
```suggestion
  ["libxkbcommon"] {os = "freebsd" | os = "dragonfly" | os = "netbsd"}
  ["xkbcommon"] {os = "openbsd"}
```

PR #22225, opened 2022-10-08
Update conf-libpng

ddeclerck:
Summary of changes :
- Windows: use package `libpng-devel` instead of `libpng`, add `os = "cygwin"`
- Debian & Ubuntu: regroup Debian and Ubuntu, use `libpng-dev` for both, add `os-family = "ubuntu"` because some (many ?) Ubuntu derivatives (Linux Mint, elementaryOS, ...) only list Ubuntu in os-release:ID_LIKE (or list both, but Ubuntu first)

Add packages for :
- Mandriva and derivatives (same as Mageia)
- SUSE/OpenSUSE/SLES (note: no generic `libpng-devel` package, so use `libpng16-devel`)
- Fedora/RHEL/CentOS
- Alpine
- Arch Linux
- Gentoo and derivatives
- *BSD
- macOS

### DIFF
--- a/packages/conf-libpng/conf-libpng.1/opam
+++ b/packages/conf-libpng/conf-libpng.1/opam
@@ -21,6 +21,9 @@ depexts: [
   ["libpng"] {os = "macos" & os-distribution = "homebrew"}
   ["libpng"] {os = "macos" & os-distribution = "macports"}
 ]
+x-ci-accept-failures: [
+  "opensuse-15.3" # does not have libpng
+]
 synopsis: "Virtual package relying on a libpng system installation"
 description:
   "This package can only install if the libpng is installed on the system."
--- a/packages/conf-libpng/conf-libpng.1/opam
+++ b/packages/conf-libpng/conf-libpng.1/opam
@@ -10,10 +10,17 @@ homepage: "http://www.libpng.org/pub/png/libpng.html"
 license: "Zlib"
 build: [["pkg-config" "libpng"]]
 depexts: [
-  ["libpng-dev"] {os-family = "debian" & os-distribution != "ubuntu"}
-  ["libpng-devel"] {os-distribution = "mageia"}
-  ["libpng12-dev"] {os-distribution = "ubuntu"}
-  ["libpng"] {os = "win32" & os-distribution = "cygwinports"}
+  ["libpng-devel"] {os = "cygwin" | os-distribution = "cygwinports"}
+  ["libpng-dev"] {os-family = "debian" | os-family = "ubuntu"}
+  ["libpng-devel"] {os-family = "mandriva" | os-family = "openmandriva" | os-family = "mageia"}
+  ["libpng16-devel"] {os-family = "suse" | os-family = "opensuse" | os-family = "sles"}
+  ["libpng-devel"] {os-family = "fedora" | os-family = "rhel" | os-family = "centos"}
+  ["libpng-dev"] {os-family = "alpine"}
+  ["libpng"] {os-family = "arch" | os-family = "archlinux"}
+  ["media-libs/libpng"] {os-family = "gentoo"}
+  ["png"] {os-family = "bsd"}
+  ["libpng"] {os = "macos" & os-distribution = "homebrew"}
+  ["libpng"] {os = "macos" & os-distribution = "macports"}
 ]
 synopsis: "Virtual package relying on a libpng system installation"
 description:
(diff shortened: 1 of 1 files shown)


ddeclerck:
I don't get why the CI fails on OpenSUSE 15.3. Works just fine with a fresh install in a VM.

packages/conf-libpng/conf-libpng.1/opam
```diff
@@ -10,10 +10,17 @@ homepage: "http://www.libpng.org/pub/png/libpng.html"
...
-  ["libpng"] {os = "win32" & os-distribution = "cygwinports"}
+  ["libpng-devel"] {os = "cygwin" | os-distribution = "cygwinports"}
+  ["libpng-dev"] {os-family = "debian" | os-family = "ubuntu"}
+  ["libpng-devel"] {os-family = "mandriva" | os-family = "openmandriva" | os-family = "mageia"}
+  ["libpng16-devel"] {os-family = "suse" | os-family = "opensuse" | os-family = "sles"}
+  ["libpng-devel"] {os-family = "fedora" | os-family = "rhel" | os-family = "centos"}
+  ["libpng-dev"] {os-family = "alpine"}
+  ["libpng"] {os-family = "arch" | os-family = "archlinux"}
+  ["media-libs/libpng"] {os-family = "gentoo"}
+  ["png"] {os-family = "bsd"}
```
kit-ty-kate, line 21:
```suggestion
  ["png"] {os = "freebsd" | os = "dragonfly" | os = "netbsd" | os = "openbsd"}
```

packages/conf-libpng/conf-libpng.1/opam
```diff
@@ -10,10 +10,17 @@ homepage: "http://www.libpng.org/pub/png/libpng.html"
...
 build: [["pkg-config" "libpng"]]
 depexts: [
-  ["libpng-dev"] {os-family = "debian" & os-distribution != "ubuntu"}
-  ["libpng-devel"] {os-distribution = "mageia"}
-  ["libpng12-dev"] {os-distribution = "ubuntu"}
-  ["libpng"] {os = "win32" & os-distribution = "cygwinports"}
+  ["libpng-devel"] {os = "cygwin" | os-distribution = "cygwinports"}
+  ["libpng-dev"] {os-family = "debian" | os-family = "ubuntu"}
+  ["libpng-devel"] {os-family = "mandriva" | os-family = "openmandriva" | os-family = "mageia"}
+  ["libpng16-devel"] {os-family = "suse" | os-family = "opensuse" | os-family = "sles"}
```
kit-ty-kate, line 16:
```suggestion
```
The `libpng16-devel` package does not install libpng. It only installs libpng16 (everything is suffixed with 16 for some reason and opensuse does not seem to have a proper libpng

packages/conf-libpng/conf-libpng.1/opam
```diff
@@ -10,10 +10,16 @@ homepage: "http://www.libpng.org/pub/png/libpng.html"
...
+  ["libpng-dev"] {os-family = "debian" | os-family = "ubuntu"}
+  ["libpng-devel"] {os-family = "mandriva" | os-family = "openmandriva" | os-family = "mageia"}
+  ["libpng-devel"] {os-family = "fedora" | os-family = "rhel" | os-family = "centos"}
+  ["libpng-dev"] {os-family = "alpine"}
+  ["libpng"] {os-family = "arch" | os-family = "archlinux"}
+  ["media-libs/libpng"] {os-family = "gentoo"}
+  ["png"] {os = "freebsd" | os = "dragonfly" | os = "netbsd" | os = "openbsd"}
+  ["libpng"] {os = "macos" & os-distribution = "homebrew"}
+  ["libpng"] {os = "macos" & os-distribution = "macports"}
 ]
```
kit-ty-kate, line 23:
```suggestion
]
x-ci-accept-failures: [
  "opensuse-15.3" # does not have libpng
]
```

packages/conf-libpng/conf-libpng.1/opam
ddeclerck: I have a VM with OpenSUSE Leap 15.3 ; here's what I found :
- libpng16-16 provides libpng16.so.16
- libpng16-devel depends on libpng16-16, and symlinks libpng16.so to libpng16.so.16
- libpng16-compat-devel (which depends on libpng-16-devel) symlinks libpng.so to libpng16.so
So I guess we could just install libpng16-compat-devel and let the dependency magic handle the rest.

ddeclerck: Oh, actually, `libpng16-compat-devel` specifies that it provides `libpng-devel`; so you can just `zypper install libpng-devel` (tested right now, works fine).

kit-ty-kate:
Thanks. Sorry for the wait

ddeclerck:
Ah, I think you missed my previous comment about `libpng-devel` being provided by `libpng16-compat-devel` on OpenSUSE (i.e. `zypper install libpng-devel` works just fine).

PR #22226, opened 2022-10-09
Update conf-libwayland and conf-wayland-protocols

ddeclerck:
Summary of changes :
- (conf-libwayland only) change build command to both test for the `wayland-scanner` tool and the `libwayland-client` library: while most OSes/distributions have these two artifacts in the same package, some (Debian, Mandriva and Gentoo families) actually use two packages, so only testing for `wayland-scanner` would result in the library not being installed if `wayland-scanner` is installed before (or, we could also have a separate package for `wayland-scanner`...)
- Debian & Ubuntu: add `os-family = "ubuntu"` because some (many ?) Ubuntu derivatives (Linux Mint, elementaryOS, ...) only list Ubuntu in os-release:ID_LIKE (or list both, but Ubuntu first)
- Mandriva and derivatives: add the packages for this family
- Arch Linux: add `os-family = "archlinux"` in addition of `os-family = "arch"` (I have seen both in Arch and its derivatives)
- Gentoo and derivatives: add the packages for this family (with an explicit `wayland-scanner`, as it is only considered a build dependency on this system)
- *BSD: use `os-family = "bsd"` (all BSD variants seem to use the same name for these two packages)

### DIFF
--- a/packages/conf-libwayland/conf-libwayland.1/opam
+++ b/packages/conf-libwayland/conf-libwayland.1/opam
@@ -4,16 +4,24 @@ homepage: "https://wayland.freedesktop.org/"
 bug-reports: "https://github.com/ocaml/opam-repository/issues"
 authors: "Wayland dev team"
 license: "MIT"
-build: ["wayland-scanner" "--version"]
+available: [ os != "macos" & os != "win32" & os != "cygwin" ]
+build: [
+  ["wayland-scanner" "--version"]
+  ["pkg-config" "--exists" "wayland-client"]
+]
+depends: [
+  "conf-pkg-config" {build}
+]
 depexts: [
-  ["libwayland-dev"] {os-family = "debian"}
-  ["wayland-devel"] {os-distribution = "fedora"}
-  ["wayland-devel"] {os-distribution = "rhel"}
-  ["wayland-devel"] {os-distribution = "centos"}
-  ["wayland-devel"] {os-family = "suse"}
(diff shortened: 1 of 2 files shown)


packages/conf-libwayland/conf-libwayland.1/opam
```diff
@@ -4,16 +4,24 @@ homepage: "https://wayland.freedesktop.org/"
...
-  ["wayland"] {os-distribution = "arch"}
-  ["wayland"] {os = "freebsd"}
+  ["libwayland-dev"] {os-family = "debian" | os-family = "ubuntu"}
+  ["wayland-devel"] {os-family = "mandriva" | os-family = "openmandriva" | os-family = "mageia"}
+  ["wayland-devel"] {os-family = "fedora" | os-family = "rhel" | os-family = "centos"}
+  ["wayland-devel"] {os-family = "suse" | os-family = "opensuse" | os-family = "sles"}
+  ["wayland-dev"] {os-family = "alpine"}
+  ["wayland"] {os-family = "arch" | os-family = "archlinux"}
+  ["dev-libs/wayland"] {os-family = "gentoo"}
+  ["dev-util/wayland-scanner"] {os-family = "gentoo"}
```
kit-ty-kate, line 23:
```suggestion
  ["dev-libs/wayland" "dev-util/wayland-scanner"] {os-family = "gentoo"}
```

kit-ty-kate:
Thanks. Sorry for the wait.

PR #22227, opened 2022-10-09
Update conf-freetype

ddeclerck:
Summary of changes :
- Windows: add the Cygwin package, use `libfreetype-devel` instead of `freetype2` (unmaintained)
- Debian & Ubuntu: add `os-family = "ubuntu"` because some (many ?) Ubuntu derivatives (Linux Mint, elementaryOS, ...) only list Ubuntu in os-release:ID_LIKE (or list both, but Ubuntu first)
- Mandriva and derivatives: add `os-family = "mandriva"` and `os-family = "openmandriva"`
- Arch Linux and derivatives: add the package for this family
- Gentoo and derivatives: add the package for this family
- *BSD: use `os-family = "bsd"` (all BSD variants seem to use the same name for this package)
- MacOS: add the MacPorts package

### DIFF
--- a/packages/conf-freetype/conf-freetype.1/opam
+++ b/packages/conf-freetype/conf-freetype.1/opam
@@ -7,17 +7,19 @@ license: "GPL-1.0-or-later"
 build: [["pkg-config" "freetype2"]]
 depends: ["conf-pkg-config" {build}]
 depexts: [
-  ["libfreetype6-dev"] {os-family = "debian"}
-  ["freetype-dev"] {os-distribution = "alpine"}
-  ["freetype-devel"] {os-distribution = "centos"}
-  ["freetype-devel"] {os-distribution = "fedora"}
-  ["freetype-devel"] {os-distribution = "rhel"}
-  ["libfreetype2-devel"] {os-distribution = "mageia"}
-  ["freetype2-devel"] {os-family = "suse"}
-  ["print/freetype2"] {os = "freebsd"}
-  ["print/freetype2"] {os = "openbsd"}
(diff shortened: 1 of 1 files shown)


ddeclerck:
I believe we can ignore the CI failure on Frama-C 24 (seems unrelated to this update).

packages/conf-freetype/conf-freetype.1/opam
```diff
@@ -7,17 +7,17 @@ license: "GPL-1.0-or-later"
...
-  ["print/freetype2"] {os = "freebsd"}
-  ["print/freetype2"] {os = "openbsd"}
+  ["libfreetype6-dev"] {os-family = "debian" | os-family = "ubuntu"}
+  ["freetype-dev"] {os-family = "alpine"}
+  ["freetype-devel"] {os-family = "fedora" | os-family = "rhel" | os-family = "centos"}
+  ["freetype2-devel"] {os-family = "mandriva" | os-family = "openmandriva" | os-family = "mageia"}
+  ["freetype2-devel"] {os-family = "suse" | os-family = "opensuse" | os-family = "sles"}
+  ["freetype2"] {os-family = "arch" | os-family = "archlinux"}
+  ["media-libs/freetype"] {os-family = "gentoo"}
+  ["freetype2"] {os-family = "bsd" & os != "openbsd"}
```
kit-ty-kate, line 17: I’d rather avoid using `os-family = "bsd"` whenever possible as it is easy to introduce bugs this way. Having one line per OS makes it also easier to maintain

kit-ty-kate:
Thanks

PR #22228, opened 2022-10-09
Update conf-libfontconfig

ddeclerck:
Regroup all BSD variants with `os-family = "bsd"`, remove the package prefix (appears to be optional).


kit-ty-kate:
I’m not sure i understand the reasoning behind this change.
For example this particular change breaks OpenBSD where the `fontconfig` package does not exist.
In my opinion keeping all variants explicitly listed makes it much easier to maintain and reason about.

What was the original issue you were trying to fix?

kit-ty-kate:
I’m still not too happy with this change but let’s merge this anyway. Sorry for the wait

PR #22229, opened 2022-10-09
Update conf-libxcb*

ddeclerck:
Regroup all BSD variants with `os-family = "bsd"`, remove the package prefix (appears to be optional).


ddeclerck:
Regarding the CI failure on Oracle Linux 8 for `conf-libxcb-keysyms` and `conf-libxcb-image` : the corresponding system packages (`xcb-util-keysyms-devel` and `xcb-util-image-devel`) are in the CodeReady Builder repository, which is not active in the CI environment. We can probably overlook that.

kit-ty-kate:
Same question as in https://github.com/ocaml/opam-repository/pull/22228

kit-ty-kate:
Thanks

PR #22235, opened 2022-10-11
[new release] routes (2.0.0)

anuragsoni:
Typed routing for OCaml applications

- Project page: <a href="https://github.com/anuragsoni/routes">https://github.com/anuragsoni/routes</a>
- Documentation: <a href="https://anuragsoni.github.io/routes/">https://anuragsoni.github.io/routes/</a>

##### CHANGES:

* use `ppx_expect` for tests
* add `route` function which is an alias of `@-->` and is used to connect a route pattern to a handler

## Breaking changes

* Drop support for OCaml 4.05-4.07
* Switch to a new model for trailing slash handling. In routes 1.0.0 users needed to be careful about using `/?` and `//?` as the former would only match routes without a trailing slash, and the latter would enforce a trailing slash.
  - Users only need to use `/?` to end routes, and it will cover both routes ending with trailing slashes and without
  - The type used for representing match results has more information about whether it was an exact match, or if it was a match but the input target had a trailing slash at the end.
  - `MatchWithTrailingSlash` informs the user that the current target was considered a match, but that the target has an additional trailing slash

### DIFF
--- a/packages/current_web/current_web.0.3/opam
+++ b/packages/current_web/current_web.0.3/opam
@@ -31,7 +31,7 @@ depends: [
   "prometheus-app" {< "1.2"}
   "cohttp-lwt-unix" {>= "2.2.0"}
   "tyxml"
-  "routes" {>= "0.8.0"}
+  "routes" {>= "0.8.0" & < "2.0.0"}
   "dune" {>= "2.0"}
 ]
 url {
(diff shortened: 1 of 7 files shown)


raphael-proust:
There are some errors in the CI: the direct dependency `current { = version }` fails to install in the lower-bound test for all compiler versions.

I'm not sure what the fix is. Maybe `current` should set an explicit dependency to the `seq` compatibility package with a minimal version?

kit-ty-kate:
The lower-bounds only fails for packages that are already there and the failures are unrelated to the release so we can simply ignore them for now (though feel free to dig into it if you want to fix it later @raphael-proust)

Thanks!

anuragsoni:
Thank you!

PR #22249, opened 2022-10-13
Added cloudi.2.0.5

okeuday:


packages/cloudi/cloudi.2.0.5/opam
```diff
@@ -0,0 +1,20 @@
+opam-version: "2.0"
+maintainer: "Michael Truog <mjtruog at protonmail dot com>"
+authors: "Michael Truog <mjtruog at protonmail dot com>"
+homepage: "https://cloudi.org/"
+bug-reports: "https://github.com/CloudI/CloudI/issues"
```
kit-ty-kate, line 5:
```suggestion
bug-reports: "https://github.com/CloudI/cloudi_api_ocaml/issues"
```

kit-ty-kate:
Thanks. Could you return the fix to the opam file upstream so that it can be taken into account for next time?

okeuday:
Ok, I will.  Thanks!

PR #22257, opened 2022-10-13
Add a conf-cosmopolitan package which checks if we can execute a Cosmopolitan application with OPAM

dinosaure:
I redo my PR #22159 with `extra-source` (/cc @kit-ty-kate)

### DIFF
--- a/packages/conf-cosmopolitan/conf-cosmopolitan.1/opam
+++ b/packages/conf-cosmopolitan/conf-cosmopolitan.1/opam
@@ -7,7 +7,7 @@ license: "MIT"
 tags: "org:mirage"
 dev-repo: "git+https://github.com/dinosaure/esperanto.git"
 build: [
-  [ "sh" "-c" "chmod +x ./hello.com" ]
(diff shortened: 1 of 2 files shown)


packages/conf-cosmopolitan/conf-cosmopolitan.1/opam
```diff
@@ -0,0 +1,26 @@
+opam-version: "2.0"
+maintainer: "Romain Calascibetta <romain.calascibetta@gmail.com>"
+authors: "Romain Calascibetta <romain.calascibetta@gmail.com>"
+homepage: "https://github.com/dinosaure/esperanto"
+bug-reports: "https://github.com/dinosaure/esperanto/issues/"
+license: "MIT"
+tags: "org:mirage"
+dev-repo: "git+https://github.com/dinosaure/esperanto.git"
+build: [
+  [ "sh" "-c" "chmod +x ./hello.com" ]
```
kit-ty-kate, line 10: For this we don't have to use sh
```suggestion
  [ "chmod" "+x" "./hello.com" ]
```

PR #22271, opened 2022-10-14
Re-add h2-async 0.9.0, gluten-async 0.3.0

anmonteiro:
Should be possible after https://github.com/ocaml/opam-repository/pull/22260

### DIFF
--- a/packages/h2-async/h2-async.0.9.0/opam
+++ b/packages/h2-async/h2-async.0.9.0/opam
@@ -13,9 +13,9 @@ depends: [
   "ocaml" {>= "4.06"}
   "dune" {>= "1.7"}
   "faraday-async" {>= "0.8.2"}
-  "async"
+  "async" {>= "v0.15.0"}
   "gluten-async" {>= "0.3.0"}
-  "h2" {>= version}
+  "h2" {= version}
 ]
 depopts: ["async_ssl"]
 synopsis: "Async support for h2"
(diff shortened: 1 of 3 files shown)


anmonteiro:
Not sure what the linking errors are on `macos-homebrew-ocaml-5.0`

packages/h2-async/h2-async.0.9.0/opam
```diff
@@ -0,0 +1,34 @@
...
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+]
+depends: [
+  "ocaml" {>= "4.06"}
+  "dune" {>= "1.7"}
+  "faraday-async" {>= "0.8.2"}
+  "async"
+  "gluten-async" {>= "0.3.0"}
+  "h2" {>= version}
```
kit-ty-kate, line 18: Given you're doing the same change above, did you mean equality instead?
```suggestion
  "h2" {= version}
```

packages/h2-async/h2-async.0.9.0/opam
```diff
@@ -0,0 +1,34 @@
...
+dev-repo: "git+https://github.com/anmonteiro/ocaml-h2.git"
+doc: "https://anmonteiro.github.io/ocaml-h2/"
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+]
+depends: [
+  "ocaml" {>= "4.06"}
+  "dune" {>= "1.7"}
+  "faraday-async" {>= "0.8.2"}
+  "async"
```
kit-ty-kate, line 16: Optional since it's already in as part of gluten-async but i suppose it's a bit better to be more explicit
```suggestion
  "async" {>= "v0.15.0"}
```

packages/h2-async/h2-async.0.9.0/opam
anmonteiro: Makes sense to start doing this. I tried doing `>=` before but it's only caused trouble so far.

kit-ty-kate:
Thanks

PR #22273, opened 2022-10-14
core_unix.v0.15.2

tov:
Sorry about the v0.15.1 misfire. Here's a new release that should work for OCaml 5.

PR #22284, opened 2022-10-17
[new release] http_async (0.1.0)

anuragsoni:
Async library for HTTP/1.1 servers

- Project page: <a href="https://github.com/anuragsoni/http_async">https://github.com/anuragsoni/http_async</a>
- Documentation: <a href="https://anuragsoni.github.io/http_async/">https://anuragsoni.github.io/http_async/</a>

##### CHANGES:

* Don't use Eager_deferred
* Increase lower-bound on shuttle to version 0.6.0

### DIFF
--- a/packages/http_async/http_async.0.1.0/opam
+++ b/packages/http_async/http_async.0.1.0/opam
@@ -1,0 +1,45 @@
+opam-version: "2.0"
+synopsis: "Async library for HTTP/1.1 servers"
+description:
+  "http_async implements an efficient HTTP/1.1 server. It uses the shuttle library for network IO, and provides an easy-to-use interface for writing async http services while minimal overhead on top of the user provided http handler."
+maintainer: ["Anurag Soni <anurag@sonianurag.com>"]
+authors: ["Anurag Soni <anurag@sonianurag.com>"]
+license: "MIT"
+tags: ["http-server" "http" "http1.1" "async"]
+homepage: "https://github.com/anuragsoni/http_async"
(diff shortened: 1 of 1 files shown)


anuragsoni:
The failures on 5.0 should be resolved once core_unix is available on that compiler switch. See https://github.com/ocaml/opam-repository/pull/22273 for the in-progress work for core_unix.

packages/http_async/http_async.0.1.0/opam
```diff
@@ -0,0 +1,45 @@
...
+    jobs
+    "--promote-install-files=false"
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+  ["dune" "install" "-p" name "--create-install-files" name]
+]
+dev-repo: "git+https://github.com/anuragsoni/http_async.git"
+available: [ arch = "x86_64" | arch = "arm64" ]
```
kit-ty-kate, line 36: is there a particular reason for this restriction?

anuragsoni: In an older submission to the opam-repository there were some test failures on `s390x`. I haven't had a chance to spend time debugging those yet so I'm locking the architectures to the platform I have an easy way to use and test on. as for only supporting 64 bit archictures, I made some assumptions to represent http body lengths as an `int` instead of `int64` that will simplifies things on 64 bit systems, but will cause unintended issues on 32 bit archiectures.

PR #16250, opened 2020-04-21
Add a GitHub Action to close stale issues

avsm:
At the OCaml core developers meeting in April 2020, we decided
to experiment with closing stale issues that have no activity for
a long period of time. This is simply to keep the working set of
issues at mangeable levels.

This PR experiments with adding this to opam-repository, where
stale issues are particularly prone to being closable as many
have been fixed by subsequent metadata improvements. Feedback
welcome on the timeouts, and the message itself.

PR #22288, opened 2022-10-17
Remove the stale bot (not used)

kit-ty-kate:
The stale bot was added back in https://github.com/ocaml/opam-repository/pull/16250, however I don't think the maintainer team every really used it (we just keep removing the bot messages and tag the issues as `persist` to get rid of the bot actions).

In my opinion the bot doesn't seem useful and rather unfriendly for users reporting issues.


dinosaure:
Thanks!

PR #22295, opened 2022-10-17
pythonlib v0.15.1

tov:
OCaml 5 support

### DIFF
--- a/packages/pythonlib/pythonlib.v0.15.1/opam
+++ b/packages/pythonlib/pythonlib.v0.15.1/opam
@@ -26,5 +26,5 @@ This library helps exposing ocaml functions to python. The python runtime intera
 "
 url {
 src: "https://github.com/janestreet/pythonlib/archive/refs/tags/v0.15.1.tar.gz"
-checksum: "sha256=849f5a1b7e4535d517f076feaf92a17378c3ab29ebda63b25175fcb03f7e1f13"
+checksum: "sha256=ed28b98f8aacafb5948d8dc0b7014524f7f8a30356db7af5b0d098e67a390440"
 }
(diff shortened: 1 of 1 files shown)


packages/pythonlib/pythonlib.v0.15.1/opam
```diff
@@ -0,0 +1,30 @@
...
+homepage: "https://github.com/janestreet/pythonlib"
+bug-reports: "https://github.com/janestreet/pythonlib/issues"
+dev-repo: "git+https://github.com/janestreet/pythonlib.git"
+doc: "https://ocaml.janestreet.com/ocaml-core/latest/doc/pythonlib/index.html"
+license: "MIT"
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+]
+depends: [
+  "ocaml"      {>= "4.11.0"}
```
kit-ty-kate, line 13:
```suggestion
  "ocaml"      {>= "4.14.0"}
```

kit-ty-kate: Actually maybe i could’ve used `[%%if ocaml_version ...` in my PR. Could you add those where necessary if you think this is needed? (in the current state this release skips OCaml 4.13)

tov: Nevermind, I figured it out.

tov:
Okay, let's see how this does.

tov:
It looks like the problem here is that `pyml < 20220322` doesn't work on OCaml 5.0. Would it make sense to add an upper bound of `< "5.0"` to the older `pyml` versions, rather than raising the lower bound for `pythonlib`?

kit-ty-kate:
Done in #22315. Thanks.

PR #22308, opened 2022-10-18
Fix bad checksums for archetype-1.3.3 and plebeia.2.0.2

mseri:
Recovered from the opam cache and added to opam-source-archives.


kit-ty-kate:
Thanks a lot

Issue #22309, opened 2022-10-18
wrong checksum for core_unix.v0.15.1


emillon:
Hi,
We're seeing checksum errors for core_unix.v0.15.1:

```
OpamSolution.Fetch_fail("https://github.com/janestreet/core_unix/archive/refs/tags/v0.15.1.tar.gz (Bad checksum, expected sha256=b467c7c64616d5cc88f42f9660b447247e7a848148336e0ad0de6bd35edce9b3)")
```

(the hash I'm seeing locally is `b0fe5650d9fd2d7d0efe0560ae8da8a90dc47eee20ca5c3283843d4522d51e5d`)

tov:
Sorry, my fault. This is fixed in #22310.

PR #22310, opened 2022-10-18
Fix checksum for core_unix.v0.15.1

tov:
Addresses #22309


kit-ty-kate:
Could you not force-push the tag back to the original instead? This would avoid everyone recompiling their packages for nothing.

kit-ty-kate:
Superseded by #22314

PR #22314, opened 2022-10-19
Fix core_unix.v0.15.1 source (upstream broke the checksum)

kit-ty-kate:
Fixes https://github.com/ocaml/opam-repository/issues/22309


emillon:
FTR, this is not just a repack that happened to change the hash. The tarball hosted at https://github.com/janestreet/core_unix/archive/refs/tags/v0.15.1.tar.gz with sha256sum b0fe5650d9fd2d7d0efe0560ae8da8a90dc47eee20ca5c3283843d4522d51e5d is a v0.15.2 tarball (at least according to its opam file).

PR #22322, opened 2022-10-19
[new release] tar, tar-unix and tar-mirage (2.2.0)

reynir:

##### CHANGES:

- `tar-mirage` requires and implements `mirage-kv.5.0.0` (@hannesm, mirage/ocaml-tar#96)
- `tar-mirage` implements `Mirage_kv.RW` (append-only) (@hannesm, @reynir, @dinosaure, review by @MisterDA, mirage/ocaml-tar#93)
- Update usage of `cstruct` in `tar`: unnecessary memsets removed, use `Cstruct.of_string` (@hannesm, mirage/ocaml-tar#93)
- Fix `tar-mirage` read buffer allocation error (@reynir, review by @hannesm, mirage/ocaml-tar#94)
- `tar` and `tar-mirage` do not require `re` anymore, `tar-mirage` doesn't depend on `iopage` and works with solo5 and other improvements (@hannesm, review by @reynir, mirage/ocaml-tar#90)

### DIFF
--- a/packages/tar-mirage/tar-mirage.2.2.0/opam
+++ b/packages/tar-mirage/tar-mirage.2.2.0/opam
@@ -29,6 +29,9 @@ depends: [
   "tar-unix" {with-test & = version}
   "odoc" {with-doc}
 ]
+conflicts: [
+  "result" {< "1.5"}
+]
 build: [
   ["dune" "subst"] {dev}
   [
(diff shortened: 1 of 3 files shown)


raphael-proust:
The CI fails with in the lower bound checks for `tar-mirage` because of some `Unbound value Result.bind`. It seems the `result` package is pulled at version 1.0 on that failed test but at version 1.5 on the non-lower-bound tests. That's probably where the issue comes from.

dinosaure:
I would like to have a review from @kit-ty-kate who is the expert about the `result` package. I'm not sure that a `conflict` is the good solution.

PR #22368, opened 2022-10-25
[new release] psq (0.2.1)

pqwy:
Functional Priority Search Queues

- Project page: <a href="https://github.com/pqwy/psq">https://github.com/pqwy/psq</a>
- Documentation: <a href="https://pqwy.github.io/psq/doc">https://pqwy.github.io/psq/doc</a>

##### CHANGES:

- added `push` to bump priorities
- added `split_at`
- changed `++`, `of_list` to select the lowest, not the last/rightmost priority

PR #22369, opened 2022-10-25
fix lru.0.3.0 dep constraint

pqwy:
Tighten the version constraint, in the presence of https://github.com/ocaml/opam-repository/pull/22368.


mseri:
Thanks

PR #22363, opened 2022-10-25
Package azure-cosmos-db.0.2.0

mknaack:
### `azure-cosmos-db.0.2.0`
Azure cosmos db interface
Provides an interface to the Azure Cosmos db.

---
* Homepage: https://bitbucket.org/knaack/azure-cosmos-db
* Source repo: git+https://bitbucket.org/knaack/azure-cosmos-db.git
* Bug tracker: https://bitbucket.org/knaack/azure-cosmos-db/issues

---
:camel: Pull-request generated by opam-publish v2.1.0

PR #22376, opened 2022-10-27
cryptokit is not OCaml 5 / NNP ready

kit-ty-kate:
See failures in https://github.com/ocaml/opam-repository/pull/22363#issuecomment-1293435408
and current C stubs which looks suspicious. e.g. https://github.com/xavierleroy/cryptokit/blob/d79833fd52f0bd3bfba5b389d436da20bcba92ff/src/stubs-blake3.c#L41


kit-ty-kate:
Nevermind it's from ocaml-ssl in the end: https://github.com/savonet/ocaml-ssl/pull/98

PR #22385, opened 2022-10-27
Dose3.5.0.1-1: Patch to grep more careful for the ocaml system

LasseBlaauwbroek:
Dose3's configure system reports on the OCaml system by grepping for `system` in `ocamlc -config`. However, on some systems, the output of `ocamlc -config` also has the word `system` on unrelated lines. Here is an example:

```
version: 4.14.0
standard_library_default: /tmp/tmp.PADAUCpkbY/opam-root/default/lib/ocaml
standard_library: /tmp/tmp.PADAUCpkbY/opam-root/default/lib/ocaml
ccomp_type: cc
c_compiler: /tmp/tmp.PADAUCpkbY/test-env/bin/x86_64-conda-linux-gnu-cc
ocamlc_cflags: -O2 -fno-strict-aliasing -fwrapv -pthread -fPIC -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem /tmp/tmp.PADAUCpkbY/test-env/include -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem /tmp/tmp.PADAUCpkbY/test-env/include -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem /tmp/tmp.PADAUCpkbY/test-env/include
ocamlc_cppflags: -D_FILE_OFFSET_BITS=64 -DNDEBUG -D_FORTIFY_SOURCE=2 -O2 -isystem /tmp/tmp.PADAUCpkbY/test-env/include -DNDEBUG -D_FORTIFY_SOURCE=2 -O2 -isystem /tmp/tmp.PADAUCpkbY/test-env/include -DNDEBUG -D_FORTIFY_SOURCE=2 -O2 -isystem /tmp/tmp.PADAUCpkbY/test-env/include
[...]

PR #22395, opened 2022-10-28
Revert "Dose3.5.0.1-1: Patch to grep more careful for the ocaml system"

kit-ty-kate:
Reverts ocaml/opam-repository#22385 as it breaks the package on BSDs


kit-ty-kate:
cc @LasseBlaauwbroek Could you reopen a PR without the `-P`? (it should probably also be done as a new `dose3.5.0.1-2` package)

PR #22400, opened 2022-10-28
codept 0.11: extend compatibility to 5.0

Octachron:
Since OCaml 5.0 shares the same AST as 4.14.0 and codept is using is own compatibility shims for avoiding deprecated functions, the  `codept.0.11.1` version is accidentally compatible with OCaml 5.0.

This PR thus relaxes the OCaml upper-bound constraints in the `codept.0.11.1` opam package.


mseri:
Was green. Thanks

PR #22401, opened 2022-10-28
Allow use of pkcs11 and pkcs11-driver on macOS

bbc2:
Out of the 4 libraries provided by https://github.com/cryptosense/pkcs11, only `pkcs11-rev` is currently incompatible with macOS. We declared `pkcs11` and `pkcs11-driver` (1.0.0+) as incompatible by mistake.


mseri:
Thanks

PR #22405, opened 2022-10-28
[new release] phantom-algebra (1.0.1)

Octachron:
This PR decreases the number of packages still relying on `jbuilder` by one, and restore the compatibility of the `phantom-algebra` package with recent versions of OCaml.

Note that the main feature of this package is that it tests a very pathological use of the type system... And it might be theoretically useful when bridging GLSL and OCaml.

### DIFF
--- a/packages/phantom-algebra/phantom-algebra.1.0.1/opam
+++ b/packages/phantom-algebra/phantom-algebra.1.0.1/opam
@@ -15,7 +15,7 @@ depends: [
   "ocaml" {>= "4.06.0"}
 ]
 build: [
-  ["dune" "subst"] {pinned}
+  ["dune" "subst"] {dev}
   [
     "dune"
     "build"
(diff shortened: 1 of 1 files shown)


packages/phantom-algebra/phantom-algebra.1.0.1/opam
```diff
@@ -0,0 +1,40 @@
...
+authors: ["Florian Angeletti <octa@polychoron.fr>"]
+license: "MIT"
+homepage: "https://github.com/Octachron/phantom_algebra"
+bug-reports: "https://github.com/Octachron/phantom_algebra/issues"
+depends: [
+  "dune" {>= "2.5"}
+  "ocaml" {>= "4.06.0"}
+]
+build: [
+  ["dune" "subst"] {pinned}
```
mseri, line 18:
```suggestion
  ["dune" "subst"] {dev}
```

kit-ty-kate:
Thanks! Could you return the fix to the opam file upstream so it can be taken into account for next time?

PR #22414, opened 2022-10-30
[new release] camyll (0.4.1)

alan-j-hu:
A static site generator

- Project page: <a href="https://alan-j-hu.github.io/camyll">https://alan-j-hu.github.io/camyll</a>

##### CHANGES:

- Upgrade `calendar`, `cmdliner`, and `textmate-language` dependencies and
  replace usage of deprecated `cmdliner` API.
- Catch stray exceptions.
- Support JSON and YAML grammar files.

### DIFF
--- a/packages/camyll/camyll.0.4.1/opam
+++ b/packages/camyll/camyll.0.4.1/opam
@@ -1,0 +1,63 @@
+opam-version: "2.0"
+synopsis: "A static site generator"
+description: """
+Camyll is a static site generator.
+
+Features:
+
+- Conversion from Markdown to HTML
+- Syntax highlighting of any language via user-provided TextMate grammars
+- Post tagging
+- Processing of literate Agda"""
+maintainer: ["Alan Hu <alanh@ccs.neu.edu>"]
+authors: ["Alan Hu <alanh@ccs.neu.edu>"]
+license: "MIT"
+tags: ["blog" "web" "website"]
+homepage: "https://alan-j-hu.github.io/camyll"
+bug-reports: "https://github.com/alan-j-hu/camyll/issues"
+depends: [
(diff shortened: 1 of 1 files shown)


alan-j-hu:
I'm concerned because one of my dependencies, lambdasoup, still has not been updated for OCaml 5, which is why the CI is failing. There is a PR to do so at https://github.com/aantron/lambdasoup/pull/51, but the author of the library has not been maintaining his libraries because of a loss of business due to world events: https://discuss.ocaml.org/t/the-future-of-dream/10602/5.

kit-ty-kate:
This should be now fixed in #22417.
I'm concerned about the amount of strict constraints in this package. Is there any chance the upper-bounds and other `=` could be removed?

alan-j-hu:
I am following the SemVer convention that a major version increment means a breaking change, except where the major version is 0, in which a minor version increment means a breaking change.

You can get rid of the upper bounds, provided that you take responsibility for adding the appropriate bound back if any of the dependencies have a major release that causes breakage for my package...

The packages `plist-xml` and `textmate-language` on my own, and since they are major version 0, I plan on incrementing the minor version upon a breaking change. Therefore, their upper bounds should be kept.

From what I can tell, the `omd` package is going through major churn right now. The working version on GitHub already appears to have a very different API than the 2.0 beta version on Opam. The `=` should be kept.

kit-ty-kate:
Thanks. Sorry for the wait. For upper-bound we can see that for next time but SemVer doesn’t mean much for OCaml due to its module system for instance. In opam-repository we do take responsability for all the revdeps when new versions of the packages are released and usually prefer if the new released packages do not have upper-bound constraints so users are not limited in the versions they have to use.

PR #22419, opened 2022-10-31
Package pla.2.1

modlfo:
### `pla.2.1`

---

---

### DIFF
--- a/packages/pla/pla.2.1/opam
+++ b/packages/pla/pla.2.1/opam
@@ -1,0 +1,23 @@
+opam-version: "2.0"
+maintainer: "Leonardo Laguna Ruiz <modlfo@gmail.com>"
+synopsis: "Pla is a simple library and ppx syntax extension to create composable templates based on verbatim strings"
+homepage: "https://modlfo.github.io/pla/"
+bug-reports: "https://github.com/modlfo/pla/issues"
+license: "MIT"
+dev-repo: "git+https://github.com/modlfo/pla.git"
+
+build: [["dune" "build" "-p" name "-j" jobs]]
+
+depends: [
+  "ocaml" { >= "4.07.0" }
+  "dune" { >= "2.5" }
+  "ppxlib" { >= "0.22.0" }
+]
+authors: "Leonardo Laguna Ruiz <modlfo@gmail.com>"
+url {
+  src: "https://github.com/modlfo/pla/archive/v2.1.tar.gz"
+  checksum: [
+    "md5=48d8a9c4f3476375bcc685bf23864d62"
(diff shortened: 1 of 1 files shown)


packages/pla/pla.2.1/opam
```diff
@@ -0,0 +1,24 @@
...
+license: "MIT"
+dev-repo: "git+https://github.com/modlfo/pla.git"
+
+build: [["dune" "build" "-p" name "-j" jobs]]
+
+depends: [
+  "ocaml" { >= "4.07.0" }
+  "dune" { build & >= "2.5" }
+  "ppxlib" { >= "0.22.0" }
+  "ocaml-compiler-libs"
```
kit-ty-kate, line 15: Are you really using `ocaml-compiler-libs` directly? I can only see uses of `compiler-libs` (those are two different libraries, one is a janestreet library, the other is a package included in OCaml itself)

packages/pla/pla.2.1/opam
```diff
@@ -0,0 +1,24 @@
...
+homepage: "https://modlfo.github.io/pla/"
+bug-reports: "https://github.com/modlfo/pla/issues"
+license: "MIT"
+dev-repo: "git+https://github.com/modlfo/pla.git"
+
+build: [["dune" "build" "-p" name "-j" jobs]]
+
+depends: [
+  "ocaml" { >= "4.07.0" }
+  "dune" { build & >= "2.5" }
```
kit-ty-kate, line 13: Dune cannot be tagged as a build dependency (see https://github.com/ocaml/dune/issues/2147)
```suggestion
  "dune" { >= "2.5" }
```

packages/pla/pla.2.1/opam
kit-ty-kate:
```suggestion
```

modlfo: It is the `ocaml-compiler-libs`. I use it to get `Location.t`. However, when linking the library I need to use the name `compiler-libs` in the dune file. I don't know why.

kit-ty-kate: That's because `ppxlib` uses `ocaml-compiler-libs` under the hood already so that makes it a transitive dependency

packages/pla/pla.2.1/opam
kit-ty-kate: This change got removed in your latest push-force

modlfo: I have applied the suggested change.

packages/pla/pla.2.1/opam
kit-ty-kate: Do you want to change your usages of `compiler-libs` by `ocaml-compiler-libs` in your dune files and re-release? Or do that later (the dependency would be a bit misleading here though)

modlfo: I can remove it in the next revison I make to the library.

kit-ty-kate:
Thanks a lot!

PR #22422, opened 2022-10-31
camlidl.1.09.1

nberth:
Provides an OCaml 5-compatible version of camlidl, without the breaking changes introduced in camlidl.1.10

### DIFF
--- a/packages/camlidl/camlidl.1.09.1/opam
+++ b/packages/camlidl/camlidl.1.09.1/opam
@@ -1,0 +1,33 @@
+opam-version: "2.0"
+synopsis: "Stub code generator for OCaml"
+description: """
+CamlIDL is a stub code generator for OCaml. It automates the
+generation of C stub code required for the Caml/C interface, based on
+an MIDL specification. Also provides some support for Microsoft's COM
+software components."""
+maintainer: "Nicolas Berthier <m@nberth.space>"
+authors: ["Xavier Leroy"]
+license: [
+  "QPL-1.0 WITH OCaml-LGPL-linking-exception"
+  "LGPL-2.0-or-later WITH OCaml-LGPL-linking-exception"
+]
+homepage: "https://github.com/xavierleroy/camlidl"
+dev-repo: "git+https://github.com/xavierleroy/camlidl.git"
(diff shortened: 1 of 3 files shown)


packages/camlidl/camlidl.1.09.1/opam
```diff
@@ -0,0 +1,33 @@
...
+depends: [
+  "ocaml" {>= "4.03"}
+]
+extra-files: [
+  ["camlidl.install" "md5=cf56e14faed046880b7c9d0f4cd737f1"]
+  ["META" "md5=f12cc59517db6c176c1ec3809dc12bac"]
+]
+url {
+  src:
+    "https://github.com/kit-ty-kate/camlidl/archive/500.tar.gz"
```
kit-ty-kate, line 31: Why is my hotfix branch in this package?

kit-ty-kate (changes requested):

packages/camlidl/camlidl.1.09.1/opam
nberth: At the beginning that was mostly to see whether the CI would pass. I looked at the diff w.r.t official 1.09 and thought it was legit enough for a quick release of a minor version.
Is there a branch/commit/PR in @xavierleroy's camlidl repository so we could do an official release for a 1.09-like version for ocaml 5.0?

kit-ty-kate: Would would not accept versions that are not official. If upstream does not want/have time to release 1.09.1 then a patch release can be done instead (1.09-1, where the official 1.09 tarball would be used but you can add a patch to the `files` directory) but I'd rather avoid it as a default when possible

xavierleroy:
Nobody reported that the changes in CamlIDL 1.10 break something.  That would be a good starting point.

nberth:
> Nobody reported that the changes in CamlIDL 1.10 break something. That would be a good starting point.

My bad. I did not read the `Changes` file carefully enough to notice there was no breaking change explicitly mentioned. I've opened an issue on what I think is either a breaking change or a bug on the `camlidl` repository.

nberth:
Closing in favor of camlidl1.11 (#22486) that brings a bugfig.

PR #22428, opened 2022-11-01
pfff needs older Menhir

Leonidas-from-XIV:
Bug reported on Discuss https://discuss.ocaml.org/t/10737

It might be reasonable to add the same constraint on all versions, since I assume the issue is valid for all releases.

### DIFF
--- a/packages/pfff/pfff.0.25.1/opam
+++ b/packages/pfff/pfff.0.25.1/opam
@@ -15,6 +15,7 @@ depends: [
   "num"
 ]
 dev-repo: "git+https://github.com/facebook/pfff"
+license: "LGPL-2.1-only WITH OCaml-LGPL-linking-exception"
 install: [make "install-findlib"]
 synopsis:
   "Tools and APIs for program analysis, code visualization, refactoring"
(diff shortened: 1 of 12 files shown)


Leonidas-from-XIV:
Looks like it also needs some lower bound on Menhir but from the error message I can't tell what the lower bound should be.

packages/pfff/pfff.0.25.1/opam
```diff
@@ -14,7 +15,9 @@ depends: [
   "camlp4"
   "num"
 ]
+bug-reports: "https://github.com/facebook/pfff/issues"
 dev-repo: "git+https://github.com/facebook/pfff"
+license: "LGPL-2.1-only WITH OCaml-LGPL-linking-exception"
```
kit-ty-kate, line 20: reading the `license.txt` file it seems to be the (default) `-or-later` variant
```suggestion
license: "LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception"
```

kit-ty-kate: ah actually nevermind I'm dumb

kit-ty-kate:
Thanks!

PR #22431, opened 2022-11-02
[new release] http-mirage-client (0.0.1)

dinosaure:
HTTP client for MirageOS

- Project page: <a href="https://github.com/roburio/http-mirage-client">https://github.com/roburio/http-mirage-client</a>

##### CHANGES:

* First release

### DIFF
--- a/packages/http-mirage-client/http-mirage-client.0.0.1/opam
+++ b/packages/http-mirage-client/http-mirage-client.0.0.1/opam
@@ -1,0 +1,49 @@
+opam-version: "2.0"
+synopsis: "HTTP client for MirageOS"
+maintainer: ["team@robur.coop"]
+authors: [
+  "Robur Team <team@robur.coop>"
+]
+license: "MIT"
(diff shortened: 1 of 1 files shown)


dinosaure:
The error (as `paf`) on MacOS is due to the unability to launch `listen` and launch an HTTP server. Otherwise, I think it's fine.

packages/http-mirage-client/http-mirage-client.0.0.1/opam
```diff
@@ -0,0 +1,50 @@
...
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
```
kit-ty-kate, line 37:
```suggestion
    "@runtest" {with-test & os != "macos"} # macOS is disabled due to restrictions in sandbox-exec
```

kit-ty-kate:
Thanks. Could you return the fix in the opam file upstream?

PR #22432, opened 2022-11-02
Frama-C: new release (26.0~beta-Iron)

AllanBlanchard:


AllanBlanchard:
The problem might be related to https://git.frama-c.com/pub/frama-c/-/issues/2634.

packages/frama-c/frama-c.26.0~beta/opam
```diff
@@ -0,0 +1,162 @@
...
+
+depopts: [
+  # cannot use {build}: Frama-C must be recompiled when Coq and libraries changes.
+  # Coq: because .vo would would not be loadable by another version of Coq
+  # libraries: because we use dynamic linking
+  "apron"
+  "coq"
+  "mlmpfr" { >= "4.1.0-bugfix2"}
+  "zmq"
+]
```
kit-ty-kate, line 140:
```suggestion
]

conflicts: [
  "result" {< "1.5"}
]
```

packages/frama-c/frama-c.26.0~beta/opam
```diff
@@ -0,0 +1,162 @@
...
+  "ppx_import"
+]
+
+depopts: [
+  # cannot use {build}: Frama-C must be recompiled when Coq and libraries changes.
+  # Coq: because .vo would would not be loadable by another version of Coq
+  # libraries: because we use dynamic linking
+  "apron"
+  "coq"
+  "mlmpfr" { >= "4.1.0-bugfix2"}
```
kit-ty-kate, line 138: Depopts should not contain constraints (with some exceptions). This should be moved to the conflicts field below

AllanBlanchard:
We have two problems to deal with by the final release but that are non-blocking for the beta relase.

Minor: The tests are disabled on macos because of the realpath command that does not exist on macos but is used in our tests.

Major: If both GTK2 and GTK3 are available, Frama-C will use GTK2 by default (wrong order of rules in our Dune files ...). We do not want to block the beta for this. But we have a decide what we do about. Entirely remove GTK2 is an option, sadly for the few users that might want to use Frama-C GUI on Macos (the subset that can get a running GTK2 GUI of Frama-C). One more motivation to finish our work on our new GUI and entirely remove GTK.

kit-ty-kate:
LGTM. I have a question before merging: this package is a beta. Is this meant to be used by default for users who want to install frama-c or just as an option while waiting for the stable release?

If it's the former, there is nothing to do and I'll be happy to merge.
If it's the later, we might want to add the following two lines to the opam file:
```
flags: avoid-version
```
This will ensure that people who are just here to install whatever the latest stable version of frama-c is will still get the latest stable version and not the beta version. For users who want the beta they'll have to use `opam install 'frama-c>=26'` or similar

AllanBlanchard:
Indeed, the later applies. I thought that the tag `~beta` was enough, I did not know about this flag. Is this a change related to precisely Opam 2.1? (in particular, why do we need the `available: opam-version >= "2.1.0"`?

mseri:
Because previous version of opam don’t understand the tag

kit-ty-kate:
Thanks a lot!

PR #22436, opened 2022-11-02
conf-tk: Add support for macOS

kit-ty-kate:


packages/conf-tk/conf-tk.1/opam
```diff
@@ -4,7 +4,10 @@ authors: "Tcl Core Team"
 homepage: "http://www.tcl.tk"
 bug-reports: "https://github.com/ocaml/opam-repository/issues"
 build: [["sh" "check.sh"]]
-depends: ["conf-pkg-config" {build}]
+depends: [
+  "conf-pkg-config" {build}
+  "conf-X11" {os = "macos"}
```
mseri, line 9: That should be libX11. I meant to comment and merged instead. I'll fix in a new PR

PR #22442, opened 2022-11-03
Add lib-clang-cpp to conf-libclang debian depexts

raphael-proust:
Attempting to improve things re: https://github.com/ocaml/opam-repository/pull/22222


kit-ty-kate:
Thanks!

PR #22443, opened 2022-11-03
Remove unnecessary upper bounds for camyll

alan-j-hu:
I realized that I forgot to update the upper bound for `calendar` when I made my new release, forcing me to downgrade it... So I'm taking your advice and removing the upper bounds (except for my own packages).


raphael-proust:
The CI errors seem to be:

- oniguruma availability on oraclelinux (not an issue)
- stack-overflow during compilation of the uunf dependency with flambda (I don't know what to do about that)
- some other issue with uunf compilation on ppc64 (also don't know about that)

All the errors are for dependencies, none for the packages themselves.

mseri:
These are all known issues, thanks!

PR #22460, opened 2022-11-07
[new release] dkml-component-staging-opam64, dkml-component-staging-opam32 and dkml-component-offline-opam (2.2.0-alpha-20221104)

jonahbeckford:
DKML component for 64-bit versions of opam

- Project page: <a href="https://github.com/diskuv/dkml-component-opam">https://github.com/diskuv/dkml-component-opam</a>

## `2.2.0~alpha~20221104`

First winget release of opam. Is the trunk (`master`) branch of opam
up to and include 2022-11-04.

## `2.2.0~dkml20220801`

opam used in DKML 1.0.0. Contains the following patches for DKML:
* https://github.com/jonahbeckford/opam/commit/b6ba2e113d32045f51bc37c53e949126133f3d3a .
  Use LOCALAPPDATA for initial Opam root

### DIFF
--- a/packages/dkml-component-offline-opam/dkml-component-offline-opam.2.2.0-alpha-20221104/opam
+++ b/packages/dkml-component-offline-opam/dkml-component-offline-opam.2.2.0-alpha-20221104/opam
@@ -1,0 +1,44 @@
+opam-version: "2.0"
+synopsis: "Offline install of Opam"
+description:
+  "Offline install that places opam in the installation directory. Depending on the end-user's integer size, as reported by Sys.int_size, either 32-bit or 64-bit opam will be installed"
+maintainer: ["opensource+diskuv-ocaml@support.diskuv.com"]
+authors: ["Diskuv, Inc. <opensource+diskuv-ocaml@support.diskuv.com>"]
+license: "Apache-2.0"
+homepage: "https://github.com/diskuv/dkml-component-opam"
+bug-reports: "https://github.com/diskuv/dkml-component-opam/issues"
+depends: [
+  "dune" {>= "2.9"}
+  "dkml-component-staging-opam32" {= version}
+  "dkml-component-staging-opam64" {= version}
+  "dkml-component-staging-ocamlrun" {>= "4.12.1~"}
+  "dkml-install" {>= "0.2.0"}
+  "diskuvbox" {>= "0.1.0"}
+  "odoc" {with-doc}
+]
+build: [
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "--promote-install-files=false"
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+  ["dune" "install" "-p" name "--create-install-files" name]
+]
+dev-repo: "git+https://github.com/diskuv/dkml-component-opam.git"
+url {
+  src:
+    "https://github.com/diskuv/dkml-component-opam/releases/download/v2.2.0-alpha-20221104/dkml-component-offline-opam-2.2.0-alpha-20221104.tbz"
+  checksum: [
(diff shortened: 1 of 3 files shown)


jonahbeckford:
I was confused about why this package would be failing on macOS arm64 (even if it is experimental). I used macOS arm64 for most of the development of this package, and the same code builds on other CI systems (GitHub Actions + GitLab CI). Then I saw the logs:

```
# @config.log+| configure:4031: checking whether we are cross compiling
# @config.log+| configure:4039: /usr/bin/clang -arch x86_64 -o conftest -arch x86_64   conftest.c  >&5
# @config.log+| configure:4043: $? = 0
# @config.log+| configure:4050: ./conftest
# @config.log+| ./configure: line 4052: ./conftest: Bad CPU type in executable
```

Is Rosetta disabled on `macos-homebrew-ocaml-4.14-arm64`? If so, what is the rationale and how did you disable it?

If the rationale makes sense, I may disable Rosetta in the GitHub/GitLab CI testing I vend through https://github.com/diskuv/dkml-workflows#dkml-workflows

kit-ty-kate:
are the versions not meant to be `~alpha` instead of `-alpha` ?

jonahbeckford:
> are the versions not meant to be `~alpha` instead of `-alpha` ?

Are we talking about branch names or the Opam versions? `git` does not allow `~` in branch and tag names, so I always use semver for branch names:

```bash
% git tag mytag~alpha~dev
fatal: 'mytag~alpha~dev' is not a valid tag name.
```

kit-ty-kate:
> Is Rosetta disabled on `macos-homebrew-ocaml-4.14-arm64`? If so, what is the rationale and how did you disable it?

I think so, at least i don't think any effort was put into enabling it and Rosetta2 is an opt-in program installed the first time you execute an x86_64 program in GUI (or maybe that has changed in 11.0 I can't remember)

cc @mtelvers do you know?

kit-ty-kate:
> Are we talking about branch names or the Opam versions? `git` does not allow `~` in branch and tag names, so I always use semver for branch names:

no i was talking about the package versions. Your changelog says `~` but the package version you submitted says `-`

jonahbeckford:
> > Are we talking about branch names or the Opam versions? `git` does not allow `~` in branch and tag names, so I always use semver for branch names:
> 
> no i was talking about the package versions. Your changelog says `~` but the package version you submitted says `-`

Oh interesting. I guess `dune-release` uses the branch name rather than the Opam package version that is inside `dune-project` and the `.opam` files. Hopefully there is an option to override it, and then I'll resubmit after figuring out about the Rosetta2 stuff.

mtelvers:
Rosetta2 is not enabled on the `macOS-arm64` machines.

jonahbeckford:
Rosetta2: The reason why the `dkml` packages assume x86_64 is always present on Apple Silicon is because it makes it easier for the end-user to do cross-compilation, and I've honestly never heard of any macOS/Silicon that didn't have Rosetta on. For cross-compilation with Dune, you have to choose what architecture is the `default` and which is the cross-compiler (ie. `default.darwin_arm64`). I've chosen that `default` is always x86_64 so to cross-compile for both macOS/Intel and macOS/Silicon you always do `dune -x darwin_arm64`. That works regardless if you are on x86_64 or on arm64, but it requires the convention that Rosetta2 is enabled.

I'm not sure what the right answer is. If we say that Rosetta2 is disabled, then it means the end-user has to differentiate when doing cross-compiling (ie. if you are on Silicon you do `dune -x darwin_x86_64`, and if you are on x86_64 you do `dune -x darwin_arm64`). And you need to also look at those differences when you look inside your `_build/` directory for the right folder or when you look inside your `<opamroot>` for the cross-compiled sysroot folder.

But there is a counter-argument, especially if there are macOS developer machines that don't have Rosetta2 enabled.

jonahbeckford:
This is not a one-way door decision; it is relatively easy to switch between the two. So I'll close this and resubmit one later in the week that does not use Rosetta2.

PR #22462, opened 2022-11-07
Package inotify.2.4.1

c-cube:
### `inotify.2.4.1`
Inotify bindings for OCaml
Inotify bindings for OCaml

---
* Homepage: https://github.com/whitequark/ocaml-inotify
* Source repo: git+https://github.com/whitequark/ocaml-inotify.git
* Bug tracker: https://github.com/whitequark/ocaml-inotify/issues

---


packages/inotify/inotify.2.4.1/opam
```diff
@@ -0,0 +1,45 @@
+# This file is generated by dune, edit dune-project instead
+opam-version: "2.0"
+synopsis: "Inotify bindings for OCaml"
+description: "Inotify bindings for OCaml"
+maintainer: ["whitequark <whitequark@whitequark.org>"]
+authors: ["whitequark <whitequark@whitequark.org>"]
+license: "LGPL-2.1"
```
kit-ty-kate, line 7:
```suggestion
license: "LGPL-2.1-only WITH OCaml-LGPL-linking-exception"
```

packages/inotify/inotify.2.4.1/opam
```diff
@@ -0,0 +1,45 @@
...
+  "dune" {>= "2.9"}
+  "base-unix"
+  "base-bytes"
+  "fileutils" {with-test & >= "0.4.4"}
+  "ounit2" {with-test & >= "2.0"}
+  "lwt" {with-test}
+  "ocaml" {>= "4.03"}
+  "odoc" {with-doc}
+]
+depopts: ["lwt"]
```
kit-ty-kate, line 21: I belive inotify is only available on linux (at least based on the failures on macOS)
```suggestion
depopts: ["lwt"]
available: os = "linux"
```

packages/inotify/inotify.2.4.1/opam
c-cube: will upstream

packages/inotify/inotify.2.4.1/opam
c-cube: how do I write that in a `dune-project` file?

kit-ty-kate: do you have to keep dune generating the opam file? There is probably a way with a template file but I'd highly recommend just stopping dune from autogenerating it.

c-cube: well that's disappointing :/
cc @rgrinberg would that be something we can add to dune? maybe arbitrary raw fields to be copied as-is into the opam file?

c-cube: Ah, I just tried a template with _only_ that field, and it works. Excellent, thank you :100:

kit-ty-kate:
Thanks

PR #22466, opened 2022-11-07
Update the torch package to 0.17.

LaurentMazare:
This makes ocaml-torch compatible with the latest libtorch 1.13 version.


kit-ty-kate:
Thanks

PR #22468, opened 2022-11-08
[new release] diskuvbox (0.1.1)

jonahbeckford:
Cross-platform basic set of script commands

- Project page: <a href="https://github.com/diskuv/diskuvbox">https://github.com/diskuv/diskuvbox</a>

##### CHANGES:

- Use memory buffering to copy files. Removes 16 MiB max file limitation on
  32-bit OCaml.
- Validate and document a bytecode guarantee that only standard stublibs are used
- Distribute binaries with setup-dkml.yml@v1
- Fix Dune build steps so works under cross-compiler
- Code working with Cmdliner.1.1.1
- Increase minimum OCaml to 4.10 to work on macOS
- Cross-compile `darwin_arm64` on `darwin_x86_64`

### DIFF
--- a/packages/diskuvbox/diskuvbox.0.1.1/opam
+++ b/packages/diskuvbox/diskuvbox.0.1.1/opam
@@ -1,0 +1,39 @@
+opam-version: "2.0"
+synopsis: "Cross-platform basic set of script commands"
+description:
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Thanks

PR #22482, opened 2022-11-09
Octez 15.0 packages

rbardou:
Octez 15.0 has been released recently. Octez is an implementation of the Tezos blockchain. Note that in this release, the executables have been renamed (from `tezos-*` to `octez-*` in particular), and the corresponding opam packages have been renamed too.

This patch was generated from a new automatic release process so I expect there to be some issues, but `opam install octez` worked on my laptop.

### DIFF
--- a/packages/tezos-protocol-environment/tezos-protocol-environment.15.0/opam
+++ b/packages/tezos-protocol-environment/tezos-protocol-environment.15.0/opam
@@ -7,7 +7,7 @@ dev-repo: "git+https://gitlab.com/tezos/tezos.git"
 license: "MIT"
 depends: [
   "dune" { >= "3.0" }
-  "ocaml" { >= "4.12" }
+  "ocaml" { >= "4.14.0" & < "4.15" }
   "tezos-base" { = version }
   "alcotest-lwt" { with-test & >= "1.5.0" }
   "tezos-test-helpers" { with-test & = version }
--- a/packages/tezos-protocol-environment/tezos-protocol-environment.15.0/opam
+++ b/packages/tezos-protocol-environment/tezos-protocol-environment.15.0/opam
@@ -1,0 +1,57 @@
+opam-version: "2.0"
+maintainer: "contact@tezos.com"
+authors: ["Tezos devteam"]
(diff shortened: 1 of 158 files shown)


rbardou:
Looks like all current CI failures are due to CI timeouts. I'm not sure what I should do now? Should we just merge?

samoht:
I've relaunched the CI for the failing builds (with timeouts and internal errors).

The archlinux error is odd, maybe a missing depext for hidapi? (as it works locally for @raphael-proust)

packages/tezos-protocol-environment/tezos-protocol-environment.15.0/opam
```diff
@@ -0,0 +1,57 @@
+opam-version: "2.0"
+maintainer: "contact@tezos.com"
+authors: ["Tezos devteam"]
+homepage: "https://www.tezos.com/"
+bug-reports: "https://gitlab.com/tezos/tezos/issues"
+dev-repo: "git+https://gitlab.com/tezos/tezos.git"
+license: "MIT"
+depends: [
+  "dune" { >= "3.0" }
+  "ocaml" { >= "4.14.0" & < "4.15" }
```
kit-ty-kate, line 10: Is there any reason why this is still not compatible with OCaml 5.0?

raphael-proust: The octez suite as a whole is not compatibly because it depends on `ringo` which we are still in the process of updating. (The release still depends on ringo<1.0.0 which is incompatible with ocaml5 because it iterates over an ephemeron table which is not domain-safe and thus removed in ocaml5.)

rbardou: Perhaps more fundamentally, we want to be very careful not to change the semantics of the economic protocol of Tezos, since that would introduce a fork in the blockchain. To this end, the protocol environment (i.e. this package) specifies exactly which major version of OCaml to use. Before bumping this major version we run tests to check that the chain still behaves the same.

PR #22499, opened 2022-11-14
conf-hidapi: install optional dep in arch, mark CI as failing in ol

raphael-proust:
Fixes for the `conf-hidapi` package following some CI failures in https://github.com/ocaml/opam-repository/pull/22482


kit-ty-kate:
Thanks

PR #22505, opened 2022-11-15
Fix conf-npm system dependency on OpenSUSE.

mefyl:
This is a blind fix based on [this ocaml-ci issue](https://github.com/ocurrent/ocaml-ci/issues/602).


kit-ty-kate:
Thanks!

PR #22507, opened 2022-11-15
Octez: add octez-proxy-server

rbardou:
This was missing from https://github.com/ocaml/opam-repository/pull/22482. I noticed it after it was merged unfortunately. Luckily the difference is small and should strain the CI much less :)


kit-ty-kate:
Thanks

PR #22512, opened 2022-11-16
ocaml 5.0.0~alpha0: conflict with dune >= 3.4

emillon:
This is related to the META files that the compiler now ship. This feature has been added in `5.0.0~alpha1`. Dune is able to determine that, but only has access to the (major, minor) numbers to do this. So recent versions of dune can not be used with `5.0.0~alpha0`.

Historically we've added the conflict on the dune side, but this means that we need to keep adding the conflict on new releases. This commit puts in on the other side, which is easier to maintain since no new incompatible versions of the ocaml compiler are expected.

(We could also remove the existing conflicts on the dune side, but doing so would cause many existing switches to be rebuilt even for users that are not on alpha0).

### DIFF
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.5.0.0~alpha0/opam
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.5.0.0~alpha0/opam
@@ -16,6 +16,9 @@ depends: [
   "ocaml-options-vanilla" {post}
   "ocaml-beta" {opam-version < "2.1.0"}
 ]
+conflicts: [
+  "dune" {>= "3.4.0"}
+]
 conflict-class: "ocaml-core-compiler"
 flags: [ compiler avoid-version ]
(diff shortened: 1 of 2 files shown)


emillon:
cc @dra27 @kit-ty-kate

mseri:
Looks reasonable. Thanks!

hannesm:
FWIW, I think the constraints should be in the dune opam package. This reduces the work on the solver (AFAIK). The maintenance issue shouldn't be too hard, you can once and for all put a `"ocaml" {!= "5.0.0~alpha0" & ...}` in your opam file(s) - as other packages do. With this change, it is very cumbersome to figure out why some packages are selected.

If such a methodology becomes mainstream, I wouldn't know where to annotate upper and lower bounds, at package X or its dependency Y.

mseri:
In general I agree with you, but this is a one-off for the first alpha of a the new compiler with lots of breaking changes. I think it is a fair exception to the norm

PR #22513, opened 2022-11-16
toml.7.0.0 does allow recent versions of menhir

nberth:


mseri:
Can you do this in a new `toml.7.0.0-1` or a new `7.0.1` release instead?

nberth:
> Can you do this in a new `toml.7.0.0-1` or a new `7.0.1` release instead?

Ok no problem.

kit-ty-kate:
Could you also use a patch file instead of taking a github commit directly? It makes it clearer to understand what has changed.

nberth:
> Could you also use a patch file instead of taking a github commit directly? It makes it clearer to understand what has changed.

There will actually be a new release for the package anyways, so that PR (which is a draft) will probably just be closed soon.

nberth:
Followed up in #22530

PR #22510, opened 2022-11-16
[new release] confero (0.1.0)

paurkedal:
Unicode Collation

- Project page: <a href="https://paurkedal.github.com/paurkedal/confero/">https://paurkedal.github.com/paurkedal/confero/</a>

##### CHANGES:

  - Initial implementation.

PR #22518, opened 2022-11-16
[new release] confero (0.1.1)

paurkedal:
I'm making a new attempt to publish this new package after the first version failed to compile with OCaml 4 on the CI due to stack overflows (#22510).  The stack overflows were caused by very large arrays in the generated source code.

Confero implements the Unicode 15.0.0 Collation Algorithm and ships with the DUCET language-neutral mapping.

CHANGES:

  - Make it possible to compile the library with OCaml 4 using a small stack size (like 4096 KiB).


paurkedal:
I think this is ready for review now that I got the lower bounds right for the remaining dependencies. The two remaining CI failure are stack overflows when building the uunf dependency under "extras".

mseri:
Thanks

PR #22527, opened 2022-11-17
+ledgerwallet.0.3.0

pirbo:
This version allows you to use the Speculos emulator of LedgerHQ hardware wallet.

### DIFF
--- a/packages/tezos-signer-backends/tezos-signer-backends.13.0/opam
+++ b/packages/tezos-signer-backends/tezos-signer-backends.13.0/opam
@@ -26,7 +26,7 @@ depopts: [
   "ledgerwallet-tezos"
 ]
 conflicts: [
-  "ledgerwallet-tezos" { < "0.2.1" }
+  "ledgerwallet-tezos" { < "0.2.1" | >= "0.3.0" }
 ]
 build: [
   ["rm" "-r" "vendors"]
--- a/packages/tezos-signer-backends/tezos-signer-backends.14.0/opam
+++ b/packages/tezos-signer-backends/tezos-signer-backends.14.0/opam
@@ -27,7 +27,7 @@ depopts: [
   "ledgerwallet-tezos"
 ]
 conflicts: [
-  "ledgerwallet-tezos" { < "0.2.1" }
+  "ledgerwallet-tezos" { < "0.2.1" | >= "0.3.0" }
 ]
 build: [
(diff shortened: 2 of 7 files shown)


packages/tezos-signer-backends/tezos-signer-backends.13.0/opam
```diff
@@ -26,7 +26,7 @@ depopts: [
   "ledgerwallet-tezos"
 ]
 conflicts: [
-  "ledgerwallet-tezos" { < "0.2.1" }
+  "ledgerwallet-tezos" { < "0.2.1" & >= "0.3.0" }
```
mseri, line 29:
```suggestion
  "ledgerwallet-tezos" { < "0.2.1" | >= "0.3.0" }
```

packages/tezos-signer-backends/tezos-signer-backends.14.0/opam
```diff
@@ -27,7 +27,7 @@ depopts: [
   "ledgerwallet-tezos"
 ]
 conflicts: [
-  "ledgerwallet-tezos" { < "0.2.1" }
+  "ledgerwallet-tezos" { < "0.2.1" & >= "0.3.0" }
```
mseri, line 30:
```suggestion
  "ledgerwallet-tezos" { < "0.2.1" | >= "0.3.0" }
```

raphael-proust:
I'm monitoring the CI and I'll push a commit to fix the tezos-related revdeps when I have a complete list.

raphael-proust:
There are CI errors still, but they are acceptable to nomadic-labs. (Although we can amend some packages with `available`/`conflicts`/etc.) if need be.)

- Linking errors (multiple definitions of the same symbols) for some binaries (codec and the test executable for client-base-unix) on arm32 only. We do not support 32bit architectures in most of our binaries anyway, so we can live with those errors whilst we check if our build system has one too many `linkall` or some such.

- An internal CI error on macos.

- An error related to ocaml5 and macos but seemingly unrelated to the packages.

pirbo:
I confirm that there is nothing I con do about the remaining MacOS only failures

raphael-proust:
https://github.com/ocaml/opam-repository/pull/22558 ← this PR should the failing test in the CI. The other errors in the CI are wontfix.

kit-ty-kate:
Thanks!

PR #22531, opened 2022-11-17
Some fixes of base-domains and js_of_ocaml-compiler.4.1.0

rjbou:
* @kayceesrk   `base-domains`  depends on non-existent `ocaml-variants.4.14.0+variants`
* @hhugo `js_of_ocaml-compiler.4.1.0` no more conflict with `base-domains` (as ocaml 5 compatible), but doesn't work with `ocaml-variants.4.12.0+domains`.

### DIFF
--- a/packages/base-domains/base-domains.base/opam
+++ b/packages/base-domains/base-domains.base/opam
@@ -6,7 +6,6 @@ Domains-based parallelism distributed with the Multicore OCaml compiler"
 depends: [
   "ocaml" {>= "5.0"} |
   "ocaml-variants" {
-    = "4.14.0+domains" |
     = "4.12.0+domains+effects" |
     = "4.12.0+domains" |
     = "4.10.0+multicore" |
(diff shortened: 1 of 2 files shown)


packages/base-domains/base-domains.base/opam
```diff
@@ -6,7 +6,6 @@ Domains-based parallelism distributed with the Multicore OCaml compiler"
 depends: [
   "ocaml" {>= "5.0"} |
   "ocaml-variants" {
-    = "4.14.0+domains" |
     = "4.12.0+domains+effects" |
```
mseri, line 9:
```suggestion
    = "4.14.0+domains" |
    = "4.12.0+domains+effects" |
```

mseri: We can't remove this since, although it is deprecated, it still exists in the multicore overlay and removing it could break things. It is harmless anyway

kit-ty-kate:
Thanks

PR #22532, opened 2022-11-18
Package guardian.0.0.4

mabiede:
### `guardian.0.0.4`
Role-based access control for OCaml
Defined actors can access targets according to specified rules (allowed actions: CRUD).

---
* Homepage: https://github.com/uzh/guardian
* Bug tracker: https://github.com/uzh/guardian/issues

---


packages/guardian/guardian.0.0.4/opam
```diff
@@ -0,0 +1,60 @@
...
+homepage: "https://github.com/uzh/guardian"
+doc: "https://github.com/uzh/guardian"
+bug-reports: "https://github.com/uzh/guardian/issues"
+depends: [
+  "base64" {>= "3.5.0"}
+  "caqti-driver-mariadb" {>= "1.8.0"}
+  "caqti-lwt" {>= "1.8.0"}
+  "conformist" {>= "0.8.1"}
+  "containers" {>= "3.6"}
+  "dune" {>= "2.9" & > "2.9.0" & < "3.0.0"}
```
kit-ty-kate, line 20: Is there a reason why this is not compatible with dune 3?
```suggestion
  "dune" {>= "2.9"}
```

packages/guardian/guardian.0.0.4/opam
```diff
@@ -0,0 +1,60 @@
...
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "--promote-install-files=false"
+    "@install"
+    "@runtest" {with-test}
```
kit-ty-kate, line 48:
```suggestion
    # "@runtest" {with-test} # The tests do not work as they require mariadb to be setup and running already.
```

kit-ty-kate:
Thanks. Could you return the fixes to the opam file upstream so that they can be taken into account for next time?

packages/guardian/guardian.0.0.4/opam
```diff
@@ -0,0 +1,60 @@
...
+  "base64" {>= "3.5.0"}
+  "caqti-driver-mariadb" {>= "1.8.0"}
+  "caqti-lwt" {>= "1.8.0"}
+  "conformist" {>= "0.8.1"}
+  "containers" {>= "3.6"}
+  "dune" {>= "2.9"}
+  "logs" {>= "0.7.0"}
+  "lwt" {>= "5.6.1"}
+  "lwt_ppx" {>= "2.1.0"}
+  "mariadb" {= "1.1.4"}
```
kit-ty-kate, line 24: oh i missed that constraint. Could you open a PR to change the constraint to something less strict like `>= 1.1.4` ?

mabiede: updated here: https://github.com/ocaml/opam-repository/pull/22567

PR #22536, opened 2022-11-20
ocaml-systemd 1.3

juergenhoetzel:
- Add compatibility with OCaml 5
- Build library using dune

### DIFF
--- a/packages/ocaml-systemd/ocaml-systemd.1.3/opam
+++ b/packages/ocaml-systemd/ocaml-systemd.1.3/opam
@@ -1,0 +1,37 @@
+opam-version: "2.0"
+maintainer: "Juergen Hoetzel <juergen@archlinux.org>"
+authors: "Juergen Hoetzel <juergen@archlinux.org>"
+homepage: "https://github.com/juergenhoetzel/ocaml-systemd/"
+bug-reports:  "https://github.com/juergenhoetzel/ocaml-systemd/issues"
+dev-repo: "git+https://github.com/juergenhoetzel/ocaml-systemd.git"
+license: "LGPL-3.0-only WITH OCaml-LGPL-linking-exception"
+
+build: [
+  "dune"
+  "build"
+  "-j"
+  jobs
+  "@install"
+]
+depends: [
+  "ocaml"
+  "dune" {build & >= "2.0"}
+]
+depexts: [
+  ["libsystemd-dev"] {os-family = "debian"}
+  ["systemd-devel"] { os-family = "suse" | os-family = "opensuse" | os-family = "fedora"}
+]
+synopsis: "OCaml module for native access to the systemd facilities"
+description: """
+* Logging to the Journal
+* Socket activation
+* Watchdog
+* Notifications"""
+url {
+  src: "https://github.com/juergenhoetzel/ocaml-systemd/archive/1.3.tar.gz"
+  checksum: [
+    "sha256=717acee7f71bc04ff3bb53c0af1231bd5c8112d4f2c1830562d5087cf1d4b152"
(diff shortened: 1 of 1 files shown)


packages/ocaml-systemd/ocaml-systemd.1.3/opam
```diff
@@ -0,0 +1,37 @@
...
+build: [
+  "dune"
+  "build"
+  "-j"
+  jobs
+  "@install"
+]
+depends: [
+  "ocaml"
+  "dune" {build & >= "2.0"}
```
kit-ty-kate, line 18:
```suggestion
  "dune" {>= "2.0"}
```
Dune cannot be tagged as a build dependency due to a bug in dune: https://github.com/ocaml/dune/issues/2147

packages/ocaml-systemd/ocaml-systemd.1.3/opam
```diff
@@ -0,0 +1,37 @@
...
+]
+depexts: [
+  ["libsystemd-dev"] {os-family = "debian"}
+  ["systemd-devel"] { os-family = "suse" | os-family = "opensuse" | os-family = "fedora"}
+]
```
kit-ty-kate, line 23: In the future it would be nice to replace this by a `conf-libsystemd` package to make it easier to maintain.

kit-ty-kate:
```
#=== ERROR while compiling ocaml-systemd.1.3 ==================================#
# context              2.2.0~alpha~dev | linux/x86_64 | ocaml-base-compiler.4.02.3 | pinned(https://github.com/juergenhoetzel/ocaml-systemd/archive/1.3.tar.gz)
# path                 ~/.opam/4.02/.opam-switch/build/ocaml-systemd.1.3
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p ocaml-systemd -j 31
# exit-code            1
# env-file             ~/.opam/log/ocaml-systemd-7-f3820a.env
# output-file          ~/.opam/log/ocaml-systemd-7-f3820a.out
### output ###
# Error: I don't know about package ocaml-systemd (passed through
# --only-packages)
```
There is a mismatch between the name of the package and the name of the library. Is this expected?

Note that before my addition of `-p name` the package would have installed nothing at all.

juergenhoetzel:
> There is a mismatch between the name of the package and the name of the library. Is this expected?

No, i guess is should have omitted the prefix when submitting the first version of ocaml-systemd back in 2015.

I just hardcoded the "library" name in the amended commit.

packages/ocaml-systemd/ocaml-systemd.1.3/opam
```diff
@@ -0,0 +1,33 @@
+opam-version: "2.0"
+maintainer: "Juergen Hoetzel <juergen@archlinux.org>"
+authors: "Juergen Hoetzel <juergen@archlinux.org>"
+homepage: "https://github.com/juergenhoetzel/ocaml-systemd/"
+bug-reports:  "https://github.com/juergenhoetzel/ocaml-systemd/issues"
+dev-repo: "git+https://github.com/juergenhoetzel/ocaml-systemd.git"
+license: "LGPL-3.0-only WITH OCaml-LGPL-linking-exception"
+
+build: ["dune" "build" "-p" "systemd" "-j" jobs]
+install: ["dune" "install" "-p" "systemd"]
```
kit-ty-kate, line 10:
```suggestion
install: ["dune" "install" "-p" "systemd" "systemd"]
```
See https://github.com/ocaml/dune/issues/4814

kit-ty-kate:
LGTM once CI finishes. Could you return the fixes in the opam file upstream to that they can be taken into account for next time?

Note also, another solution for future releases could also be to create a new `systemd` package and create a `ocaml-systemd.transition` package that points to it. Several packages made similar moves in the past and it avoids the extra annoying `dune install` (slower than what opam does)

kit-ty-kate:
Thanks

PR #22548, opened 2022-11-23
Update conf-libpng

ddeclerck:
Following the discussion in #22225, add back the SUSE packages `libpng-devel`.

### DIFF
--- a/packages/conf-libpng/conf-libpng.1/opam
+++ b/packages/conf-libpng/conf-libpng.1/opam
@@ -14,15 +14,13 @@ depexts: [
   ["libpng-dev"] {os-family = "debian" | os-family = "ubuntu"}
   ["libpng-devel"] {os-family = "mandriva" | os-family = "openmandriva" | os-family = "mageia"}
   ["libpng-devel"] {os-family = "fedora" | os-family = "rhel" | os-family = "centos"}
+  ["libpng-devel"] {os-family = "suse" | os-family = "opensuse" | os-family = "sles"}
   ["libpng-dev"] {os-family = "alpine"}
   ["libpng"] {os-family = "arch" | os-family = "archlinux"}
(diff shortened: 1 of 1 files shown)


kit-ty-kate (changes requested):
The installation just fails with
```
[ERROR] Package conf-libpng.1 depends on the unavailable system package 'libpng-devel'. You can use `--no-depexts' to attempt installation anyway.
```
on OpenSUSE

ddeclerck:
> The installation just fails with
> 
> ```
> [ERROR] Package conf-libpng.1 depends on the unavailable system package 'libpng-devel'. You can use `--no-depexts' to attempt installation anyway.
> ```
> 
> on OpenSUSE

Ah, I see. I checked in opam source code, and realized that when using zypper, it does not consider "virtual" packages when **searching** packages, although it does when installing packages, i.e. `zypper install libpng-devel` works fine, but `zypper search libpng-devel` does not (tested right now in an OpenSUSE 15.3 VM).

For this to work, that would require adding the `--provides` (and maybe `--match-exact` ?) argument to the `zypper` command in opam... Should I open an issue / feature request on opam for this ?

kit-ty-kate:
> Should I open an issue / feature request on opam for this ?

The overall problem is tracked in https://github.com/ocaml/opam/issues/4759 but could you do a PR to add the missing argument if that fixes the issue on OpenSUSE? (if you have the time, otherwise you can just comment on that issue)

ddeclerck:
> > Should I open an issue / feature request on opam for this ?
> 
> The overall problem is tracked in [ocaml/opam#4759](https://github.com/ocaml/opam/issues/4759) but could you do a PR to add the missing argument if that fixes the issue on OpenSUSE? (if you have the time, otherwise you can just comment on that issue)

Well I'll try to do make the fix and check that it doesn't break anything.

PR #22552, opened 2022-11-23
Frama-C: new release (26.0-Iron)

AllanBlanchard:


packages/frama-c/frama-c.26.0/opam
```diff
@@ -0,0 +1,171 @@
...
+  "dune-site"
+
+  ( "alt-ergo-free" | "alt-ergo" )
+  "conf-graphviz" { post }
+  "conf-time" { with-test }
+  "ocaml" { >= "4.11.1" }
+  "ocamlfind" # needed beyond build stage, used by -load-module
+  "ocamlgraph" { >= "1.8.8" }
+  "why3" { >= "1.5.1" }
+  "yojson" { >= "1.6.0" & < "2.0.0" }
```
kit-ty-kate, line 122: Is that really the case? yojson 2 was released almost 6 months ago at this point so I’m just wondering if this is just not a remanent of an outdated state of frama-c
```suggestion
  "yojson" { >= "1.6.0" }
```

AllanBlanchard: In fact it is related to the tests. The pretty printer has been changed in Yojson 2.0.0, and it breaks some of our tests. Maybe we could indicate this restriction for the test only :thinking: . Does it make sense?

kit-ty-kate: ah indeed. In this case I'll change that for all the previous versions as well. It seems to have been a problem since frama-c 22.0. In `frama-c.25.0` the package added with `"yojson" {>= "1.6.0" & ( < "2.0.0" | ! with-test) }` but for some reason subsequent releases removed it. Do you know the reason behind it?

kit-ty-kate:
I've opened https://github.com/ocaml/opam-repository/pull/22570 to fix the previous versions as well.

packages/frama-c/frama-c.26.0/opam
kit-ty-kate:
```suggestion
  "yojson" {>= "1.6.0" & (< "2.0.0" | !with-test)}
```
I don't have write access to your branch. Could you apply this suggestion?

AllanBlanchard: > Do you know the reason behind it?

Hum, no, I don't know. Migrating Frama-C to Dune forced us to change a lot of things and maybe we just removed this constraint so that pinning dependencies for working on the dune development branch was faster. I'm not sure.

kit-ty-kate:
Thanks

PR #22562, opened 2022-11-24
guardian: Remove restrictive constraints

kit-ty-kate:
Follow up to https://github.com/ocaml/opam-repository/pull/22532#discussion_r1031802151
@mabiede Is there a reason why this constraint was there in the first place?


mabiede:
@kit-ty-kate not specifically, thanks for spotting it. I'll fix the dependency constraints and release a new version.

PR #22564, opened 2022-11-25
add coq.8.16.1

palmskog:
New release of Coq.

cc: @MSoegtropIMC @ppedrot

PR #22568, opened 2022-11-25
add coqide.8.16.1

palmskog:
This is a followup to #22564 adding CoqIDE 8.16.1 served from the same archive.

cc: @MSoegtropIMC @ppedrot


kit-ty-kate:
Thanks

PR #22569, opened 2022-11-25
frama-c.24.0: restrict coq to < 8.16.1

artagnon:
coq.8.16.1 has been released, but the CI fails in #22564, due to which coq.8.16.1 is not in opam. Restrict the version for now, and unblock the Coq opam release.

### DIFF
--- a/packages/frama-c/frama-c.24.0/opam
+++ b/packages/frama-c/frama-c.24.0/opam
@@ -128,7 +128,7 @@ depopts: [
   # Coq: because .vo would would not be loadable by another version of Coq
   # libraries: because we use dynamic linking
   "apron"
-  "coq" { < "8.16.1" }
+  "coq" { < "8.16" }
   "mlgmpidl"
   "ppx_deriving"
   "ppx_deriving_yojson"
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
What error does this fixes? I'm not seeing the (related) failure

packages/frama-c/frama-c.24.0/opam
```diff
@@ -128,7 +128,7 @@ depopts: [
   # Coq: because .vo would would not be loadable by another version of Coq
   # libraries: because we use dynamic linking
   "apron"
-  "coq"
+  "coq" { < "8.16.1" }
```
palmskog, line 131:
```suggestion
  "coq" { < "8.16" }
```
We can see in CI runs for https://github.com/ocaml/opam-repository/pull/22564 and https://github.com/ocaml/opam-repository/pull/22066 that `frama-c.24.0` fails as a reverse dependency. However, this could also be due to a general problem with the package (and not with Coq version specifically) .

kit-ty-kate: precisely, so I'm not sure why the author added an unrelated constraint?

kit-ty-kate:
Closing this as the constraint is unnecessary and unrelated (as mentioned above). Feel free to open a new PR when you find the actual issue (most likely a change in yojson or something)

kit-ty-kate:
I've opened https://github.com/ocaml/opam-repository/pull/22570 instead. This should be sufficient. Turns out it was a known issue by the frama-c developers.

PR #22574, opened 2022-11-28
new file:   packages/parany/parany.12.2.1/opam [do not merge]

UnixJunkie:
Bug correction for Mac OS X.
you need to use [Filename.temp_file ~temp_dir:"/tmp"] or Mac OS will put temp files under some too longly named temporary directory. Of course, on Linux there is no such problem and you are not forced to specify ~temp_dir.

Bug reported and fix tested by Nils Becker @nilsbecker. Big thanks to him!

### DIFF
--- a/packages/parany/parany.12.2.1/opam
+++ b/packages/parany/parany.12.2.1/opam
@@ -11,6 +11,7 @@ depends: [
   "dune" {>= "1.6.0"}
   "ocaml" {>= "4.03.0"}
 ]
+available: [ os != "windows" ]
 build: [
   ["dune" "build" "-p" name "-j" jobs]
   ["dune" "build" "-p" name "-j" jobs "src/test.exe"] {with-test}
--- a/packages/parany/parany.12.2.1/opam
+++ b/packages/parany/parany.12.2.1/opam
@@ -1,0 +1,47 @@
+opam-version: "2.0"
+maintainer: "unixjunkie@sdf.org"
+authors: "Francois Berenger"
+license: "LGPL-2.0-or-later"
+homepage: "https://github.com/UnixJunkie/parany"
+bug-reports: "https://github.com/UnixJunkie/parany/issues"
+dev-repo: "git+https://github.com/UnixJunkie/parany.git"
+depends: [
+  "base-unix"
+  "cpu"
+  "dune" {>= "1.6.0"}
+  "ocaml" {>= "4.03.0"}
+]
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "build" "-p" name "-j" jobs "src/test.exe"] {with-test}
+  ["./test.sh"] {with-test}
+]
+synopsis: "Parallelize any computation"
+description: """
+Generalized map reduce for parallel computers (not distributed computing).
+Can process in parallel an infinite stream of elements.
+
+Can process a very large file in parallel on a multicore computer;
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
It looks like the tests for this package infinit loops on Linux. Maybe this error is related too?
```
#=== ERROR while compiling parany.12.2.1 ======================================#
# context              2.2.0~alpha~dev | linux/x86_64 | ocaml-base-compiler.4.14.0 | pinned(https://github.com/UnixJunkie/parany/archive/v12.2.1.tar.gz)
# path                 ~/.opam/4.14/.opam-switch/build/parany.12.2.1
# command              ~/.opam/opam-init/hooks/sandbox.sh build ./test.sh
# exit-code            1
# env-file             ~/.opam/log/parany-7-4a8d4b.env
# output-file          ~/.opam/log/parany-7-4a8d4b.out
### output ###
# ./test.sh: line 7: _build/default/src/test_parmap.exe: No such file or directory
# diff: can't create temp file '/tmp/difXXXXXX': Read-only file system
# diff: can't create temp file '/tmp/difXXXXXX': Read-only file system
# diff: can't create temp file '/tmp/difXXXXXX': Read-only file system
# diff: can't create temp file '/tmp/difXXXXXX': Read-only file system
```

UnixJunkie:
I would not release this software if the regression test did not pass on my machine.
The error is obviously related to the CI test environment: /tmp is a read-only file system ?!
Well, that's not what the library is expecting.

kit-ty-kate:
> The error is obviously related to the CI test environment: /tmp is a read-only file system ?!

Yes, in a sandboxed environment `/tmp` is readonly. You need to use the standard `TMPDIR` environment variable instead (the OCaml `Filename.*temp*` functions or any POSIX tools do support that for instance)

raphael-proust:
The code seem to hardcode the `/tmp` prefix. Does this work on windows? Should the package have an `avaiable` field to exclude windows?

UnixJunkie:
Since this is fork-based, this is not supposed to work on windows.
When ocaml5 is out, I will release a version that is not fork-based and so that should also work on windows.

UnixJunkie:
We could probably use TMPDIR instead of a hard-coded "/tmp", but I am not sure the OSX bug we just corrected would not reappear then.

mseri:
It worked fine for me using `get_temp_dir`: https://github.com/UnixJunkie/parany/pull/61

packages/parany/parany.12.2.1/opam
```diff
@@ -0,0 +1,47 @@
...
+license: "LGPL-2.0-or-later"
+homepage: "https://github.com/UnixJunkie/parany"
+bug-reports: "https://github.com/UnixJunkie/parany/issues"
+dev-repo: "git+https://github.com/UnixJunkie/parany.git"
+depends: [
+  "base-unix"
+  "cpu"
+  "dune" {>= "1.6.0"}
+  "ocaml" {>= "4.03.0"}
+]
```
raphael-proust, line 13:
```suggestion
]
available: [ os != "windows" ]
```

Something like that might be appropriate if the package is not intended to be usable on windows. Maybe the same change should be applied to other versions of the same package?

UnixJunkie:
I'll integrate some further improvement by @mseri, then issue a new tag.

UnixJunkie:
superseded by https://github.com/ocaml/opam-repository/pull/22605

PR #22576, opened 2022-11-28
Package parmap.1.2.5

kit-ty-kate:
### `parmap.1.2.5`
Minimalistic library allowing to exploit multicore architecture
Parmap is a minimalistic library allowing to exploit multicore
architecture for OCaml programs with minimal modifications: if you
want to use your many cores to accelerate an operation which happens
to be a map, fold or map/fold (map-reduce), just use Parmap’s parmap,
parfold and parmapfold primitives in place of the standard List.map
and friends, and specify the number of subprocesses to use by the
optional parameter ~ncores.

---
* Homepage: https://github.com/rdicosmo/parmap
* Source repo: git+https://github.com/rdicosmo/parmap.git
* Bug tracker: https://github.com/rdicosmo/parmap/issues

---

### DIFF
--- a/packages/parmap/parmap.1.2.5/opam
+++ b/packages/parmap/parmap.1.2.5/opam
@@ -32,7 +32,7 @@ build: [
     "-j"
     jobs
     "@install"
-    "@runtest" {with-test}
+    "@runtest" {with-test & arch != "arm32" & arch != "x86_32"}
     "@doc" {with-doc}
   ]
 ]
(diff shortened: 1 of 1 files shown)


packages/parmap/parmap.1.2.5/opam
```diff
@@ -0,0 +1,46 @@
...
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
```
kit-ty-kate, line 35:
```suggestion
    "@runtest" {with-test & arch != "arm32" & arch != "x86_32"}
```

PR #22578, opened 2022-11-29
Allow the use of rustup in `conf-rust` packages

raphael-proust:
`rustup` is a tool to manage rust installations. Its functionalities overlap with `opam`. For example, similarly to opam switches, it is possible to have multiple versions of rust installed on the same machine, and have different shells (for different software projects) use different versions. Unfortunately, there is interference in the way that rustup dispatches to different installations and the way opam sandboxes the building of packages.

This PR attempts to fix that. It does so by making the install step try both `rustc` (the binary that is available in the PATH in the sandboxed environment) and `rustup rustc` (the binary that `rustup` dispatches to from the sandboxed environment).

I'm not sure this is the best way to make it all work. It worked on my local machine. And I think it's ok. But maybe there is a better approach. Suggestions welcome.

### DIFF
--- a/packages/conf-rust-2018/conf-rust-2018.1/files/rustc.sh
+++ b/packages/conf-rust-2018/conf-rust-2018.1/files/rustc.sh
@@ -1,0 +1,3 @@
+!/bin/sh
+
+rustc --edition 2018 test.rs || rustup rustc --edition 2018 test.rs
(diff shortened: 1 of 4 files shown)


packages/conf-rust-2018/conf-rust-2018.1/files/rustc.sh
```diff
@@ -0,0 +1,3 @@
+!/bin/sh
```
kit-ty-kate, line 1:
```suggestion
#!/bin/sh
```

packages/conf-rust-2021/conf-rust-2021.1/files/rustc.sh
```diff
@@ -0,0 +1,3 @@
+!/bin/sh
```
kit-ty-kate, line 1:
```suggestion
#!/bin/sh
```

raphael-proust:
Closing, I'll reopen if I can reproduce and I'll give better explanations.

PR #22586, opened 2022-11-29
extunix: fix OCaml 5.0 incompatibilites and libexecinfo-dev Alpine depext availability

MisterDA:
Commit removing the package:
https://git.alpinelinux.org/aports/commit/?id=50795a14dee6

There is no out-of-the box replacement. Apparently it was all broken anyway.

### DIFF
--- a/packages/extunix/extunix.0.4.0/opam
+++ b/packages/extunix/extunix.0.4.0/opam
@@ -33,7 +33,7 @@ homepage: "https://github.com/ygrek/extunix"
 bug-reports: "https://github.com/ygrek/extunix/issues"
 depends: [
   "dune" {>= "2.8"}
-  "ocaml" {>= "4.06"}
+  "ocaml" {>= "4.06" & < "5.0"}
   "dune-configurator" {>= "2.9" & build}
   "ppxlib" {>= "0.18" & build}
   "ounit2" {with-test}
--- a/packages/extunix/extunix.0.4.0/opam
+++ b/packages/extunix/extunix.0.4.0/opam
@@ -57,7 +57,7 @@ build: [
   ]
 ]
 dev-repo: "git+https://github.com/ygrek/extunix.git"
-depexts: [["libexecinfo-dev"] {os = "alpine" & "3.5" <= os-version & os-version < "3.17"}]
+depexts: [["libexecinfo-dev"] {os = "alpine" & os-version >= "3.5" & os-version < "3.17"}]
 url {
   src:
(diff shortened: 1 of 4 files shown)


mseri:
Looks like extunix 0.3.2 and 0.4 are not compatible with ocaml 5. Do we need to add an upper bound?

MisterDA:
extunix.0.4.1 fixes OCaml 5 build. I've added an upper bound to the 0.3 and 0.4 series, do I need to go further back in time?

mseri:
We can take care of older versions separately

packages/extunix/extunix.0.4.0/opam
```diff
@@ -57,7 +57,7 @@ build: [
   ]
 ]
 dev-repo: "git+https://github.com/ygrek/extunix.git"
-depexts: [["libexecinfo-dev"] {os = "alpine"}]
+depexts: [["libexecinfo-dev"] {os = "alpine" & "3.5" <= os-version & os-version < "3.17"}]
```
kit-ty-kate, line 60:
```suggestion
depexts: [["libexecinfo-dev"] {os = "alpine" & os-version >= "3.5" & os-version < "3.17"}]
```

packages/extunix/extunix.0.4.1/opam
```diff
@@ -57,7 +57,7 @@ build: [
   ]
 ]
 dev-repo: "git+https://github.com/ygrek/extunix.git"
-depexts: [["libexecinfo-dev"] {os = "alpine"}]
+depexts: [["libexecinfo-dev"] {os = "alpine" & "3.5" <= os-version & os-version < "3.17"}]
```
kit-ty-kate, line 60:
```suggestion
depexts: [["libexecinfo-dev"] {os = "alpine" & os-version >= "3.5" & os-version < "3.17"}]
```

kit-ty-kate (commented):
That’s the more standard way of reading the opam syntax

kit-ty-kate:
I’m just waiting for the CI to actually test Alpine 3.17. I’ll do that tomorrow.

kit-ty-kate:
It works just fine on Alpine 3.17 (I'm not sure if it's expected though). Thanks

MisterDA:
> It works just fine on Alpine 3.17 (I'm not sure if it's expected though).

That is expected: extunix has a configure-like step detecting the presence of libexec. I had to add it as a depext so that opam and the CI would install it, but it's now removed in Alpine, so we can remove the depext too and extunix will not detect it.

PR #22503, opened 2022-11-15
[new release] js_of_ocaml, js_of_ocaml-tyxml, js_of_ocaml-toplevel, js_of_ocaml-ppx_deriving_json, js_of_ocaml-ppx, js_of_ocaml-lwt and js_of_ocaml-compiler (4.1.0)

hhugo:
Compiler from OCaml bytecode to JavaScript

- Project page: <a href="https://ocsigen.org/js_of_ocaml/latest/manual/overview">https://ocsigen.org/js_of_ocaml/latest/manual/overview</a>
- Documentation: <a href="https://ocsigen.org/js_of_ocaml/latest/manual/overview">https://ocsigen.org/js_of_ocaml/latest/manual/overview</a>

## Features/Changes
* Compiler: initial support for OCaml 5 (ocsigen/js_of_ocaml#1265,ocsigen/js_of_ocaml#1303)
* Compiler: bump magic number to match the 5.0.0~alpha0 release (ocsigen/js_of_ocaml#1288)
* Compiler: complain when runtime and compiler built-in primitives disagree (ocsigen/js_of_ocaml#1312)
* Compiler: more efficient implementation of Js_traverse.freevar
* Compiler: more efficient implementation of Js_traverse.rename_variable
* Compiler: --linkall now export all compilation units in addition to primitives (ocsigen/js_of_ocaml#1324)
* Compiler: improve --dynlink, one no longer need to pass --toplevel to use Dynlink (ocsigen/js_of_ocaml#1324)
* Compiler: toplevel runtime files "+toplevel.js" and "+dynlink.js" are added automatically (ocsigen/js_of_ocaml#1324)
* Misc: switch to cmdliner.1.1.0
* Misc: remove old binaries jsoo_link, jsoo_fs
* Misc: remove uchar dep
* Misc: use 4.14 in the CI
* Misc: switch to dune 3
* Lib: add missing options for Intl.DateTimeFormat
* Lib: add missing options for Intl.NumberFormat
* Lib: wheel event binding
* Lib: add normalize in js_string (ES6)
* Lib: more complete transition event bindings
* Lib: remove support for old browser-specific transition events
* Runtime: Implement weak semantic for weak and ephemeron
* Runtime: Implement Gc.finalise_last
* Runtime: Implement buffer for in_channels
* Runtime: add support for unix_opendir, unix_readdir, unix_closedir, win_findfirst, win_findnext, win_findclose
* Runtime: Dont use require when target-env is browser
[...]

PR #22600, opened 2022-12-02
Remove upper bound of brr on  jsoo 

dbuenzli:
This is a test. It was introduced here https://github.com/ocaml/opam-repository/pull/22503 by @hhugo but I don't think that's necessary (at least things seem to compile file on my side).


dbuenzli:
I think I understood now. It must be possible to make the package compile by adding a dependency on `js_of_ocaml` – which the last commit does.

PR #22602, opened 2022-12-02
dns < 6.0.0 is not compatible with OCaml 5.0

kit-ty-kate:
```
#=== ERROR while compiling dns.4.6.3 ==========================================#
# context              2.2.0~alpha~dev | linux/x86_64 | ocaml-variants.5.0.0+trunk | file:///home/opam/opam-repository
# path                 ~/.opam/5.0/.opam-switch/build/dns.4.6.3
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p dns -j 31
# exit-code            1
# env-file             ~/.opam/log/dns-8-bfb0e7.env
# output-file          ~/.opam/log/dns-8-bfb0e7.out
### output ###
# File "cache/dns_cache.ml", line 98, characters 13-41:
# 98 | module LRU = Lru.M.MakeSeeded(Key)(Entry)
#                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
# Error: The functor application is ill-typed.
#        These arguments:
#          Key Entry
#        do not match these parameters:
#          functor (K : Hashtbl.SeededHashedType) (V : Lru.Weighted) -> ...
#        1. Modules do not match:
#             Key :
#             sig
#               type t = [ `raw ] Domain_name.t
#               val equal : 'a Domain_name.t -> 'b Domain_name.t -> bool
#               val hash : int -> 'a -> int
#             end
#           is not included in
#             Hashtbl.SeededHashedType
#           The value `seeded_hash' is required but not provided
#           File "hashtbl.mli", line 411, characters 4-36: Expected declaration
#        2. Module Entry matches the expected module type Lru.Weighted
```

### DIFF
--- a/packages/dns/dns.0.10.0/opam
+++ b/packages/dns/dns.0.10.0/opam
@@ -30,7 +30,7 @@ build: [
 ]
 remove: [["ocamlfind" "remove" "dns"]]
 depends: [
-  "ocaml" {>= "4.00.0"}
+  "ocaml" {>= "4.00.0" & < "5.0"}
   "lwt" {>= "2.4.3" & < "3.0.0"}
   "cstruct" {>= "1.0.1" & < "1.6.0"}
   "ocamlfind"
--- a/packages/dns/dns.0.11.0/opam
+++ b/packages/dns/dns.0.11.0/opam
@@ -30,7 +30,7 @@ build: [
 ]
 remove: [["ocamlfind" "remove" "dns"]]
 depends: [
-  "ocaml" {>= "4.00.0"}
+  "ocaml" {>= "4.00.0" & < "5.0"}
   "lwt" {>= "2.4.3" & < "3.0.0"}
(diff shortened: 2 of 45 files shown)


hannesm:
Thanks for this PR, but this is only using the lru package -- I wonder why there should be a restriction on dns? (I'd think with a modern version of lru that supports ocaml 5, dns should build fine!?).

But in the end, I don't care much of these arcane packages.

kit-ty-kate:
@hannesm This is not an issue in lru though. lru < 0.3.0 are already marked as incompatible. The issue comes from the module passed in parameter to Lru which has changed type.

See this [4.14 (working)](http://check.ocamllabs.io/log/1669744335-db38a406384370363a352c142c72890a2354f1b0/4.14/good/dns.4.6.3) run vs. this [5.0 (broken)](http://check.ocamllabs.io/log/1669744335-db38a406384370363a352c142c72890a2354f1b0/5.0/bad/dns.4.6.3)

PR #22605, opened 2022-12-05
new file:   packages/parany/parany.12.2.2/opam

UnixJunkie:


mseri:
The alpine failures are due to alpine’s workers configuration since the cose is using the correct ocaml function to locate the tmp dir

PR #22607, opened 2022-12-05
Octez 15.1 packages

rbardou:
This is a bug fix release of Octez.

In theory I included all the fixes in `.opam` files I had to make for the 15.0 release, so it shouldn't require as many back and forth.

### DIFF
--- a/packages/tezos-signer-backends/tezos-signer-backends.15.1/opam
+++ b/packages/tezos-signer-backends/tezos-signer-backends.15.1/opam
@@ -1,0 +1,45 @@
+opam-version: "2.0"
+maintainer: "contact@tezos.com"
+authors: ["Tezos devteam"]
+homepage: "https://www.tezos.com/"
+bug-reports: "https://gitlab.com/tezos/tezos/issues"
+dev-repo: "git+https://gitlab.com/tezos/tezos.git"
+license: "MIT"
+depends: [
+  "dune" { >= "3.0" }
+  "tezos-base" { = version }
+  "tezos-stdlib" { = version }
+  "tezos-client-base" { = version }
+  "tezos-rpc-http" { = version }
+  "tezos-rpc-http-client" { = version }
+  "tezos-signer-services" { = version }
+  "tezos-shell-services" { = version }
+  "uri" { >= "2.2.0" }
+  "tezos-error-monad" { with-test & = version }
+  "tezos-crypto" { with-test & = version }
+  "alcotest-lwt" { with-test & >= "1.5.0" }
+  "ocplib-endian"
+  "fmt" { >= "0.8.7" }
+  "tezos-clic" { = version }
+  "tezos-stdlib-unix" { = version }
+  "tezos-rpc-http-client-unix" { = version }
+]
+depopts: [
+  "ledgerwallet-tezos"
+]
+conflicts: [
+  "ledgerwallet-tezos" { < "0.2.1" }
+]
+build: [
+  ["rm" "-r" "vendors"]
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+]
+synopsis: "Tezos: remote-signature backends for `tezos-client`"
+url {
+  src: "https://gitlab.com/tezos/tezos/-/archive/v15.1/tezos-15.1.tar.gz"
+  checksum: [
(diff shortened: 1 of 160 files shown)


packages/tezos-signer-backends/tezos-signer-backends.15.1/opam
```diff
@@ -0,0 +1,45 @@
...
+  "fmt" { >= "0.8.7" }
+  "tezos-clic" { = version }
+  "tezos-stdlib-unix" { = version }
+  "tezos-rpc-http-client-unix" { = version }
+]
+depopts: [
+  "ledgerwallet-tezos"
+]
+conflicts: [
+  "ledgerwallet-tezos" { < "0.2.1" }
```
raphael-proust, line 31:
```suggestion
  "ledgerwallet-tezos" { < "0.2.1" | >= "0.3.0" }
```

There was a breaking-change in the API going into the 0.3.x versions. This is causing the CI to fail.

raphael-proust:
Looking at all the failures of the CI:

- [lower-bounds (failed: /home/opam/.opam/4.14/lib/zarith/META: No such file or directory)](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/24913c645f2b75b3ba1e561fbd1c18f21139d639/variant/compilers,4.14,tezos-client-014-PtKathma.15.1,lower-bounds) fails whilst installing `zarith_stubs_js.v0.12.0`. Looking at some other jobs, they seem to install this same package correctly (e.g., [this other job](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/24913c645f2b75b3ba1e561fbd1c18f21139d639/variant/compilers,4.14,octez.15.1,lower-bounds). I'm not sure what the issue is. The two jobs use the same low `dune` version.

- (archlinux) [tests (failed: The compilation of tezos-baking-alpha.15.1 failed at "dune runtest -p tezos-baking-alpha -j 71".)](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/24913c645f2b75b3ba1e561fbd1c18f21139d639/variant/distributions,archlinux,tezos-baking-alpha.15.1,tests) fails with a test error because of a timeout inside the test. I think the timeout need to be changed upstream, it's probably too specialised to upstream's CI.

- (debian-unstable) [tests (failed: The compilation of tezos-baking-alpha.15.1 failed at "dune runtest -p tezos-baking-alpha -j 71".)](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/24913c645f2b75b3ba1e561fbd1c18f21139d639/variant/distributions,debian-unstable,tezos-baking-alpha.15.1,tests) fails with a similar error (overspecialised timeout in upstream test).

- (fedora-35) [tests (failed: The compilation of tezos-baking-015-PtLimaPt.15.1 failed at "dune runtest -p tezos-baking-015-PtLimaPt -j 31".)](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/24913c645f2b75b3ba1e561fbd1c18f21139d639/variant/distributions,fedora-35,tezos-baking-015-PtLimaPt.15.1,tests) fails with a similar error (overspecialised timeout in upstream test).

- (ubuntu-20.04) [tests (failed: The compilation of tezos-baking-015-PtLimaPt.15.1 failed at "dune runtest -p tezos-baking-015-PtLimaPt -j 31".)](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/24913c645f2b75b3ba1e561fbd1c18f21139d639/variant/distributions,ubuntu-20.04,tezos-baking-015-PtLimaPt.15.1,tests) fails with a similar error (overspecialised timeout in upstream test).

- There is a lot of macos failures that seem due to runner errors.

raphael-proust:
Status update:
- There are some revdeps job that fail with some runner error (`Canceled`), but it's for the deps-only package to install them all, and this one succeeds on its own. So I think we can ignore this.
- There are some timeout failures in tests. I don't want to spam the cluster by retrying those until we get a lucky break and it passes. This needs to be fixed upstream. But in the meantime we can mostly ignore them: those tests work on some platforms, on some versions, etc.
- Fedora and open-suse have some failures related to fetching system packages. I think we can ignore.
- s390x has runner failures: the disk is full. I don't know what to do (or even if I can do anything) about full disks in the runners.
- macos runners seem to have run in some kind of internal issue. They complain about a file already existing. I guess there's some synchronisation/parallelism issue somewhere. I think we can ignore because upstream doesn't provide macos support.

kit-ty-kate:
Thanks

PR #22611, opened 2022-12-05
releases lwt-parallel.1.0.0

ivg:
The [lwt-parallel][1] library allows running lwt computations in different OS processes. In general, Lwt and Unix fork do not play nicely but this library makes it possible. See the README and implementation for more details. The library works on Unix systems, such as GNU/Linux and macOS.

The bumped to 1.0.0 version indicates the stability of the interface, i.e., we do not plan to break it without chaning the major number. It also indicates that the library is ready to use and is supported.

The 1.0.0 release includes some new features, such as an explicit notation for process snapshots (and support for running multiple snapshots in parallel). It is also now possible to specify custom serialization protocols for data that is passed between processes.

[1]: https://github.com/ivg/lwt-parallel

### DIFF
--- a/packages/lwt-parallel/lwt-parallel.1.0.0/opam
+++ b/packages/lwt-parallel/lwt-parallel.1.0.0/opam
@@ -1,0 +1,29 @@
+opam-version: "2.0"
+maintainer: "Ivan Gotovchits <ivg@ieee.org>"
+authors: "Ivan Gotovchits <ivg@ieee.org>"
+homepage: "https://github.com/ivg/parallel"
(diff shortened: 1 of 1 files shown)


ivg:
Okay, looks like that we have passed all the tests, except the three experimental jobs on macOS that didn't even start. Their pipleines failed with messages like,
```
fatal: Unable to create '/private/var/lib/ocluster-worker/git/opam-repository.git-b322f7b18d276db68487afc3dfc5a5befdd45818/.git/index.lock': File exists.
```

mseri:
Looks good, thanks

PR #22614, opened 2022-12-06
[new release] fd-send-recv (v2.0.2)

psafont:
CHANGES:

- OCaml 5 compatibility

Signed-off-by: Pau Ruiz Safont <pau.ruizsafont@cloud.com>

### DIFF
--- a/packages/fd-send-recv/fd-send-recv.2.0.2/opam
+++ b/packages/fd-send-recv/fd-send-recv.2.0.2/opam
@@ -1,0 +1,34 @@
+opam-version: "2.0"
+maintainer: "xen-api@lists.xensource.com"
+authors: [
+  "David Scott"
+  "David Sheets"
+  "Euan Harris"
+  "Vincent Bernardoff"
+]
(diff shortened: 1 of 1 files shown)


packages/fd-send-recv/fd-send-recv.2.0.2/opam
```diff
@@ -0,0 +1,34 @@
...
+  "org:mirage"
+  "org:xapi-project"
+]
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+  ["dune" "build" "@doc" "-p" name] {with-doc}
+]
+depends: [
+  "ocaml" {>= "5.0"}
```
mseri, line 24: Is it just ocaml5 or it supports also older versions? Unless you use effects it should at least work in 4.14.
```suggestion
  "ocaml" {>= "4.02"}
```
I'd remove the bound and check

mseri: I see apologies. I think that may be the only one

PR #22625, opened 2022-12-07
New release of Solo5.0.7.5

dinosaure:
* Since MirageOS moved from PV mode to PVH on Xen, and thus replacing Mini-OS with solo5, there was an issue in the solo5 code which failed to properly account the already written bytes on the console. This only occurs if the output to be performed does not fit in a single output buffer (2048 bytes on Xen).

  The code in question set the number of bytes written to the last written count `written = output_some(buf)`, instead of increasing the written count `written += output_some(buf)`.

  Thanks to Krzysztof Burghardt, Pierre Alain, Thomas Leonard & Hannes Mehnert
  for the issue, review and proposal. The fix is available here: #538.

### DIFF
--- a/packages/solo5-cross-aarch64/solo5-cross-aarch64.0.7.5/opam
+++ b/packages/solo5-cross-aarch64/solo5-cross-aarch64.0.7.5/opam
@@ -1,0 +1,45 @@
+opam-version: "2.0"
+maintainer: "martin@lucina.net"
+authors: [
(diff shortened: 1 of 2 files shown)


hannesm:
the CI failures are unrelated - would it be possible to merge this (since it is security-related, and we would like to announce an advisory). thanks! :)

kit-ty-kate:
Thanks

PR #22626, opened 2022-12-07
[new release] dolmen_type, dolmen_model, dolmen_lsp, dolmen_loop, dolmen_bin and dolmen (0.8)

Gbury:

### UI

- Add a minimal reporting style accessible via the `--report-style`
  option. When used, the `dolmen` binary will use at most one line
  to output the result of processing the input file
- Add an option to the `dolmen` binary to force a specific smtlib2
  logic, overriding the one given in the file. This is accessible
  via the `--force-smtlib2-logic` option
- Add some documentation for setting up the lsp with neovim (PR#114)

### Model verification

- Added model verification. This currently supports all
  builtins, except for String/Regular expressions.

### Parsing

- Fix long compilation time of tptp parser due to flambda (PR#111)
- Replace some `assert false` by proper error messages when
  there is not the same number of function signatures as function
  definitions in a `define-funs-rec` command in smtlib2
- Accept all reserved words in s-exprs in smtlib (mainly affects
  parsing of attributes)
- Added a parser for the smtlib model specification language
- Fix doc comments mentionning removed parameters
  (PR Gbury/dolmen#107, issue Gbury/dolmen#106)
- Add an option to print syntax error identifiers (mainly to be
  used for debug)
- Register a printer for the `Uncaught_exn` exception (mainly useful
  for library users)
- Add a tag to differentiate predicates from functions in alt-ergo
  (PR#104)

### Typing

- Properly typecheck s-expressions in attributes for smtlib2
  (most notably in :patterns attributes for psmt2)
- Cleaned up handling of definitions: instead of using
  the functors in `Def`, definitions are now simply declared
  using the functions exposed by the typechecker
- Stop emitting unused warnings for type wildcards
- Expose term constants in the `Std.Expr` module (PR#112)

### Loop

- Changed the state type from a record to an heterogeneous
  map. This simplifies interfaces for all Loop modules,
  and makes it much more extensible.
- Added initialization functions for each pipe in order to
  correctly init the expecteds keys in the state
- Allow users to better control the interactive prompt when
  parsing from stdin (PR#113)


Gbury:
> Yes it should be adde. The next release of Colibri2 will support the new version.

I added the upper bound

Gbury:
In order to make the CI happy, I had to add quite a few constraints to deps/rev-deps (e.g. `spelll`, `colibrics`) of dolmen that could not compile on ocaml.5.0 (due to removal of some features, such as `Stream`, of string tags for formatters), is that okay to do in this PR, or should these changes be part of another PR ?

mseri:
Thanks, fixes can go here no problem (especially since they re just a few packages). We can fix the lints separately. Thanks for explaining the bounds reasons in the commits

PR #22638, opened 2022-12-09
Eio: add an upper-bound on mtime

talex5:
It currently fails to install, with:
```
# File "lib_eio/time.ml", line 91, characters 14-29:
# 91 |       let d = Mtime.Span.to_s d in
#                    ^^^^^^^^^^^^^^^
```
/cc @bikallem


mseri:
Thanks

PR #22653, opened 2022-12-12
[new release] OCADml (0.2.0)

geoffder:
Types and functions for building CAD packages in OCaml

- Project page: <a href="https://github.com/OCADml/OCADml">https://github.com/OCADml/OCADml</a>
- Documentation: <a href="https://OCADml.github.io/OCADml">https://OCADml.github.io/OCADml</a>

##### CHANGES:

- Use v4 type for `Plane.t` abstract it and add type conversions
- `Mesh.extrude` altered to return mid-sectionless shape when `~height` is less
  than the combined height of `~caps` (rather than breaking with `nans`)
- Use `Mesh.Cap.t` and sub-variants more consistently for `~caps` specification in
  sweeping functions of `Mesh` (looping restricted using sub-types)
- Add `Mesh.revolve`

### DIFF
--- a/packages/OCADml/OCADml.0.2.0/opam
+++ b/packages/OCADml/OCADml.0.2.0/opam
@@ -1,0 +1,40 @@
+opam-version: "2.0"
+synopsis: "Types and functions for building CAD packages in OCaml"
+description: "Types and functions for building CAD packages in OCaml"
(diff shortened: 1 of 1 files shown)


mseri:
Thanks

PR #22654, opened 2022-12-12
[new release] omd (2.0.0.alpha3)

shonfeder:
A Markdown frontend in pure OCaml

- Project page: <a href="https://github.com/ocaml/omd">https://github.com/ocaml/omd</a>

##### CHANGES:

- Expose the HTML escape function `htmlentities` (ocaml/omd#295 @cuihtlauac)

- Support generation of identifiers in headers (ocaml/omd#294, @tatchi)

- Support GitHub-Flavoured Markdown tables (ocaml/omd#292, @bobatkey)

- Update parser to support CommonMark Spec 0.30 (ocaml/omd#266, @SquidDev)

- Preserve the order of input files in the HTML output to stdout (ocaml/omd#258,
  @patricoferris)

- Fix all deviations from CommonMark Spec 0.30 (ocaml/omd#284, ocaml/omd#283, ocaml/omd#278, ocaml/omd#277, ocaml/omd#269,
  @tatchi)

PR #22659, opened 2022-12-13
Release omd.2.0.0~alpha3

mseri:
See discussion in https://github.com/ocaml/opam-repository/pull/22654#issuecomment-1349595997

The tagging is the same as the old alpha releases, I think we can simply recover the release but use the correct versioning scheme


kit-ty-kate:
Thanks

PR #22662, opened 2022-12-14
[conf-clang-format] add package to test for clang-format

maroneze:
Some distributions do not install clang-format along with clang; others have a specific package that does not depend on the 'clang' package itself. In both cases, it is useful to have a specific package for this command.

### DIFF
--- a/packages/conf-clang-format/conf-clang-format.1/opam
+++ b/packages/conf-clang-format/conf-clang-format.1/opam
@@ -8,7 +8,8 @@ build: ["clang-format" "--version"]
 depexts: [
   ["clang-format"] {os-family = "debian"}
   ["clang-tools-extra"] {os-family = "fedora"}
-  ["clang15-extra-tools"] {os-distribution = "alpine"}
+  ["clang-extra-tools"] {os-distribution = "alpine" & os-version < "3.17" }
+  ["clang15-extra-tools"] {os-distribution = "alpine" & os-version >= "3.17" }
   ["clang-tools"] {os-family = "suse"}
   ["clang"] {os-distribution = "arch"}
   ["clang-format"] {os-distribution = "homebrew" & os = "macos"}
--- a/packages/conf-clang-format/conf-clang-format.1/opam
+++ b/packages/conf-clang-format/conf-clang-format.1/opam
@@ -1,0 +1,19 @@
+opam-version: "2.0"
+maintainer: "https://github.com/ocaml/opam-repository/issues"
+homepage: "https://clang.llvm.org/docs/ClangFormat.html"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+authors: "The Clang Team"
+license: "NCSA"
+build: ["clang-format" "--version"]
+depexts: [
+  ["clang-format"] {os-family = "debian"}
+  ["clang-tools-extra"] {os-family = "fedora"}
+  ["clang15-extra-tools"] {os-distribution = "alpine"}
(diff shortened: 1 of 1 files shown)


mseri:
In macports is included in `clang-XX` where `XX` denotes the version, so we don't have a way to add a specific package

packages/conf-clang-format/conf-clang-format.1/opam
```diff
@@ -0,0 +1,19 @@
+opam-version: "2.0"
+maintainer: "https://github.com/ocaml/opam-repository/issues"
+homepage: "https://clang.llvm.org/docs/ClangFormat.html"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+authors: "The Clang Team"
+license: "NCSA"
+build: ["clang-format" "--version"]
+depexts: [
+  ["clang-format"] {os-family = "debian"}
+  ["clang-tools-extra"] {os-family = "fedora"}
```
mseri, line 10:
```suggestion
  ["clang-tools-extra"] {os-distribution = "fedora"}
```

packages/conf-clang-format/conf-clang-format.1/opam
```diff
@@ -0,0 +1,19 @@
...
+  ["clang15-extra-tools"] {os-distribution = "alpine"}
```
mseri, line 11: I am not convinced by this, in the sense that previous alpine would work fine if we were selecting the right versions: `alpine-3.16` provides `clang-extra-tools` which installs version 13.0.1.
As per macports, I think it may be better to omit this, but I'll wait to see what the other maintainers think before doing anything

maroneze: I agree the `alpine` match was not ideal, but I was not sure how to proceed. Should I specify a specific version of alpine, such as this?

```
  ["clang-extra-tools"] {os-distribution = "alpine" & os-version < "3.17" }
  ["clang15-extra-tools"] {os-distribution = "alpine" & os-version >= "3.17" }
```

There is something similar in the conf-postgresql.1 package (postgresql-dev/postgresql14-dev).

I have a Mac with homebrew for testing, so I'm afraid I cannot easily test MacPorts without risking breaking things. But if there is a way, I can try it.

packages/conf-clang-format/conf-clang-format.1/opam
mseri: Thanks. Maybe then this change is not needed, I was just trying to uniform with other conf packages. Interesting to see that the family name is actually fedora now.

packages/conf-clang-format/conf-clang-format.1/opam
mseri:
```suggestion
  ["clang-extra-tools"] {os-distribution = "alpine" & os-version < "3.17" }
  ["clang15-extra-tools"] {os-distribution = "alpine" & os-version >= "3.17" }
```
Let's try your suggestion here

packages/conf-clang-format/conf-clang-format.1/opam
maroneze: I think the "real" family name (as in the OCaml constructor) might be `Centos`, so it might make sense to use that instead. However, if I grep `os-family` in `opam-repository`, I think `fedora` appears more often than `centos` and `rhel`. In any case, if there is a will to normalize it, no problem, just tell me which one to use and I'll change it.

mseri:
It's good as is. Thanks!

PR #22667, opened 2022-12-14
coq-native currently conflicts with OCaml 5.0

palmskog:
cc: @ejgallego @erikmd


ejgallego:
The conflict situation is as follows:
- coq-native conflicts with any NNP compiler (including those of the 4.x series)
- Coq 8.16 doesn't support OCaml 5 at all, that could change if someone backports a few changes
- Coq 8.17 will hopefully support OCaml 5 as long as `-native-compiler no` is used in configure.

PR #22677, opened 2022-12-15
Update version bounds: Camyll also works with 2.0.0~alpha3

alan-j-hu:


alan-j-hu:
Okay, I did something wrong. Apparently there are breaking changes in the update from alpha1 to alpha2.

kit-ty-kate:
Thanks!

PR #22678, opened 2022-12-15
Parany for OCaml-5

UnixJunkie:
Slight changes in the interface so major version.

Adapting to this new version should be straightforward. Expect even faster parallel OCaml programs.
Exchanging values between parallel workers is now way faster than what any previous parany version was doing (relying on Marshal).

new file:   packages/parany/parany.13.0.0/opam

### DIFF
--- a/packages/parany/parany.13.0.0/opam
+++ b/packages/parany/parany.13.0.0/opam
@@ -6,7 +6,7 @@ homepage: "https://github.com/UnixJunkie/parany"
 bug-reports: "https://github.com/UnixJunkie/parany/issues"
 dev-repo: "git+https://github.com/UnixJunkie/parany.git"
 depends: [
-  "domainslib"
+  "domainslib" {>= "0.5.0"}
   "dune" {>= "1.6.0"}
   "ocaml" {>= "5.0.0"}
 ]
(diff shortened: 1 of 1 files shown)


packages/parany/parany.13.0.0/opam
```diff
@@ -0,0 +1,41 @@
+opam-version: "2.0"
+maintainer: "unixjunkie@sdf.org"
+authors: "Francois Berenger"
+license: "LGPL-2.0-or-later"
+homepage: "https://github.com/UnixJunkie/parany"
+bug-reports: "https://github.com/UnixJunkie/parany/issues"
+dev-repo: "git+https://github.com/UnixJunkie/parany.git"
+depends: [
+  "domainslib"
```
raphael-proust, line 9: There's a lower-bound failure which suggests that there should be a constraint here:

```suggestion
  "domainslib" {>= "0.5.0"}
```

kit-ty-kate:
Thanks

UnixJunkie:
@raphael-proust thanks for the opam file fixes, I have reflected them in the corresponding branch.
Apparently I still need to update the master branch so that it reflects this new version.

PR #22682, opened 2022-12-15
Mark octez.15.0 as unavailable

raphael-proust:
ping @rbardou


raphael-proust:
@rbardou Can you confirm these are the packages that need to be marked?

raphael-proust:
confirmed to me offband

Issue #22457, opened 2022-11-06
Add conf-bash for BSDs


kit-ty-kate:
zarith and gmp could be easy target to add a dependency for them
```
File "duniverse/Zarith/dune", line 79, characters 0-146:
79 | (rule
80 |  (deps META)
81 |  (action
82 |   (with-stdout-to
83 |    zarith_version.ml
84 |    (progn
85 |     (run echo "let")
86 |     (bash "grep \"version\" META | head -1")))))
Error: I need bash to interpret (bash ...) actions but I couldn't find it :(
```

raphael-proust:
https://github.com/ocaml/opam-repository/pull/22516 ← This pull request introduced some packages that could use a build-time dependency to `conf-bash`

jideelgh:
I'm using OpenBSD 7.4 with ksh, nothing fancy.
conf-bash keeps asking to install bash every time i launch "opam upgrade" while it's already installed.
I tried to let opam install bash or to install it manually.
Bash is installed in /usr/local/bin/bash, I tried to symlink to /bin/bash, no success.
bash --version returns 
"GNU bash, version 5.2.15(1)-release (x86_64-unknown-openbsd7.4)
Copyright (C) 2022 Free Software Foundation, Inc.
[...]"
I used ktrace to investigate some pledge/unveil issue, couldn't identify one.
I also tried to start opam w/ the --debug flag & adjusting the level with --debug-level flags, no luck either.
Any idea ?

PR #22683, opened 2022-12-15
Bash for bsd

raphael-proust:
Starting to work on https://github.com/ocaml/opam-repository/issues/22457

### DIFF
--- a/packages/conf-bash/conf-bash.1/opam
+++ b/packages/conf-bash/conf-bash.1/opam
@@ -1,0 +1,13 @@
+opam-version: "2.0"
+maintainer: "code@bnwr.net"
+homepage: "https://www.gnu.org/software/bash/"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: "GPL-3.0-or-later"
+build: ["bash" "--version"]
+depexts: [
+  ["bash"] {os-family = "bsd"}
+]
+synopsis: "Virtual package to install the Bash shell"
+description: "This package will install a system bash"
+authors: "code@bnwr.net"
+flags: conf
--- a/packages/conf-bash/conf-bash.1/opam
+++ b/packages/conf-bash/conf-bash.1/opam
@@ -6,6 +6,7 @@ license: "GPL-3.0-or-later"
 build: ["bash" "--version"]
 depexts: [
   ["bash"] {os-family = "bsd"}
+  ["bash"] {os-distribution = "alpine"}
 ]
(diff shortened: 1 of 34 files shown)


packages/conf-bash/conf-bash.1/opam
```diff
@@ -0,0 +1,13 @@
+opam-version: "2.0"
+maintainer: "code@bnwr.net"
+homepage: "https://www.gnu.org/software/bash/"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: "GPL-3.0-or-later"
+build: ["bash" "--version"]
+depexts: [
+  ["bash"] {os-family = "bsd"}
```
avsm, line 8: Also isn't zsh default on homebrew now, so bash would be needed there?

mseri: For the moment macos keeps shipping bash though
```
$ /bin/bash --version
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin22)
Copyright (C) 2007 Free Software Foundation, Inc.
```
I am not sure I would like an unnecessary extra install of bash in that case

raphael-proust: @avsm the aim is to allow packages that call `bash` explicitly to specify that system dependency.

avsm: Fine by me.

packages/oranger/oranger.2.0.1/opam
```diff
@@ -21,6 +21,7 @@ depends: [
   "minicli" {with-test & >= "5.0.0"}
   "cpm" {with-test}
   "conf-gnuplot" {with-test}
+  "conf-bash" {build}
```
kit-ty-kate, line 24: is this package really using bash? I can only see `sh` used here

raphael-proust: The file `packages/oranger/oranger.2.0.1/files/compile_ranger.sh` has a `#!/bin/bash` shebang. And in the repository the file `install_ranger.sh` also has the bash shebang.

kit-ty-kate: `compile_ranger.sh` is not used anywhere as far as i can see, and `install_ranger.sh` is called directly using `sh` which bypasser the shebang

raphael-proust: I removed this for this package and the other `oranger` packages.

PR #22689, opened 2022-12-16
num is does not handle not having ocamlopt

kit-ty-kate:
```
#=== ERROR while compiling num.1.4 ============================================#
# context              2.2.0~alpha~dev | linux/arm32 | ocaml-variants.5.0.0+trunk | file:///home/opam/opam-repository
# path                 ~/.opam/5.0/.opam-switch/build/num.1.4
# command              ~/.opam/opam-init/hooks/sandbox.sh build make
# exit-code            2
# env-file             ~/.opam/log/num-7-813787.env
# output-file          ~/.opam/log/num-7-813787.out
### output ###
# make -C src all
# make[1]: Entering directory '/home/opam/.opam/5.0/.opam-switch/build/num.1.4/src'
# ocamlc -ccopt -DBNG_ARCH_generic -c bng.c
# ocamlc -ccopt -DBNG_ARCH_generic -c nat_stubs.c
# ocamlmklib -oc nums bng.o nat_stubs.o
# ocamlc -w +a-4-9-41-42-44-45-48 -warn-error A -bin-annot -g -safe-string -strict-sequence -strict-formats -c int_misc.mli
# ocamlc -w +a-4-9-41-42-44-45-48 -warn-error A -bin-annot -g -safe-string -strict-sequence -strict-formats -c int_misc.ml
# ocamlc -w +a-4-9-41-42-44-45-48 -warn-error A -bin-annot -g -safe-string -strict-sequence -strict-formats -c nat.mli
# ocamlc -w +a-4-9-41-42-44-45-48 -warn-error A -bin-annot -g -safe-string -strict-sequence -strict-formats -c nat.ml
# ocamlc -w +a-4-9-41-42-44-45-48 -warn-error A -bin-annot -g -safe-string -strict-sequence -strict-formats -c big_int.mli
# ocamlc -w +a-4-9-41-42-44-45-48 -warn-error A -bin-annot -g -safe-string -strict-sequence -strict-formats -c big_int.ml
# ocamlc -w +a-4-9-41-42-44-45-48 -warn-error A -bin-annot -g -safe-string -strict-sequence -strict-formats -c arith_flags.mli
# ocamlc -w +a-4-9-41-42-44-45-48 -warn-error A -bin-annot -g -safe-string -strict-sequence -strict-formats -c arith_flags.ml
# ocamlc -w +a-4-9-41-42-44-45-48 -warn-error A -bin-annot -g -safe-string -strict-sequence -strict-formats -c ratio.mli
# ocamlc -w +a-4-9-41-42-44-45-48 -warn-error A -bin-annot -g -safe-string -strict-sequence -strict-formats -c ratio.ml
# ocamlc -w +a-4-9-41-42-44-45-48 -warn-error A -bin-annot -g -safe-string -strict-sequence -strict-formats -c num.mli
# ocamlc -w +a-4-9-41-42-44-45-48 -warn-error A -bin-annot -g -safe-string -strict-sequence -strict-formats -c num.ml
# ocamlc -w +a-4-9-41-42-44-45-48 -warn-error A -bin-annot -g -safe-string -strict-sequence -strict-formats -c arith_status.mli
# ocamlc -w +a-4-9-41-42-44-45-48 -warn-error A -bin-annot -g -safe-string -strict-sequence -strict-formats -c arith_status.ml
# ocamlmklib -o nums -oc nums -linkall int_misc.cmo nat.cmo big_int.cmo arith_flags.cmo ratio.cmo num.cmo arith_status.cmo
# ocamlopt -w +a-4-9-41-42-44-45-48 -warn-error A -bin-annot -g -safe-string -strict-sequence -strict-formats -c int_misc.ml
# make[1]: ocamlopt: No such file or directory
# make[1]: *** [Makefile:66: int_misc.cmx] Error 127
# make[1]: Leaving directory '/home/opam/.opam/5.0/.opam-switch/build/num.1.4/src'
# make: *** [Makefile:2: all] Error 2
```

### DIFF
--- a/packages/num/num.1.0/opam
+++ b/packages/num/num.1.0/opam
@@ -31,7 +31,10 @@ depends: [
   "ocaml" {>= "4.06.0"}
   "ocamlfind" {build & >= "1.7.3"}
 ]
-conflicts: [ "base-num" ]
+conflicts: [
+  "base-num"
+  "ocaml-option-bytecode-only"
+]
 patches: [ "findlib-install.patch" "installation-warning.patch" ]
 synopsis:
   "The legacy Num library for arbitrary-precision integer and rational arithmetic"
--- a/packages/num/num.1.1/opam
+++ b/packages/num/num.1.1/opam
@@ -31,7 +31,10 @@ depends: [
   "ocaml" {>= "4.06.0"}
   "ocamlfind" {build & >= "1.7.3"}
 ]
-conflicts: [ "base-num" ]
+conflicts: [
+  "base-num"
+  "ocaml-option-bytecode-only"
+]
 patches: [ "findlib-install.patch" "installation-warning.patch" ]
 synopsis:
   "The legacy Num library for arbitrary-precision integer and rational arithmetic"
--- a/packages/num/num.1.2/opam
+++ b/packages/num/num.1.2/opam
@@ -26,7 +26,10 @@ depends: [
   "ocaml" {>= "4.06.0"}
   "ocamlfind" {build & >= "1.7.3"}
 ]
-conflicts: [ "base-num" ]
+conflicts: [
+  "base-num"
+  "ocaml-option-bytecode-only"
+]
 patches: [ "installation-warning.patch" ]
 synopsis:
   "The legacy Num library for arbitrary-precision integer and rational arithmetic"
--- a/packages/num/num.1.3/opam
+++ b/packages/num/num.1.3/opam
@@ -25,7 +25,10 @@ depends: [
   "ocaml" {>= "4.06.0"}
   "ocamlfind" {build & >= "1.7.3"}
 ]
-conflicts: [ "base-num" ]
+conflicts: [
+  "base-num"
+  "ocaml-option-bytecode-only"
+]
 patches: [ "installation-warning.patch" ]
 synopsis:
   "The legacy Num library for arbitrary-precision integer and rational arithmetic"
--- a/packages/num/num.1.4/opam
+++ b/packages/num/num.1.4/opam
@@ -10,7 +10,10 @@ depends: [
   "ocaml" {>= "4.06.0"}
   "ocamlfind" {build & >= "1.7.3"}
 ]
-conflicts: ["base-num"]
+conflicts: [
+  "base-num"
+  "ocaml-option-bytecode-only"
+]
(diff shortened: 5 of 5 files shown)


kit-ty-kate:
cc @xavierleroy

xavierleroy:
This is an OCaml 5 problem.  With OCaml 4, num detects the absence of ocamlopt just fine.  The problem is that the meaning of the `ARCH` variable in OCaml's `Makefile.config` changed in OCaml 5: now it can be different from `none` even if the native-code compiler is not supported.  @dra27 might remember the reason for this change (I cannot) and suggest a fix to the num library.  I wonder how many other packages are broken because of this change.

xavierleroy:
@kit-ty-kate : I think you should have a conflict for OCaml >= 5 only, as no change is required with OCaml 4.

dra27:
It affects 4.12+ (it's https://github.com/ocaml/ocaml/pull/10044), we're just likely to see this more now with some architectures being _always_ bytecode only. You can see it with:

```Dockerfile
FROM ocaml/opam:debian-11-opam
ENV OPAMCONFIRMLEVEL=unsafe-yes
RUN opam-2.1 switch create 4.12 ocaml.4.12.1 ocaml-option-bytecode-only
RUN opam-2.1 switch create 4.11 ocaml-variants.4.11.2+bytecode-only
RUN opam-2.1 install --switch=4.11 num
RUN opam-2.1 install --switch=4.12 num
```

The 4.11 installation succeeds where the 4.12 one doesn't.

Dune detects the executable rather than reading the configuration (IIRC it only reads the configuration files for a couple of things it can't get from `ocaml{c,opt} -config`), so most other packages should be fine.

@kit-ty-kate - up to you if you want to do an opam-health-run to compare a 4.14 with 4.14-bytecode-only switch to pre-emptively catch any more of these?

@xavierleroy - I'm happy to do a fix for num (unless you're already doing one?), but it'll be tomorrow.

xavierleroy:
There's no need to rush a fix in `num`, but it should be fixed eventually.  Not sure what's the best way to go about it.  (I must say I'm tired of so much back-and-forth on configuration stuff.)  In 5.00 hindsight, https://github.com/ocaml/ocaml/pull/10044 should not have been merged as-is.

PR #22695, opened 2022-12-16
[new release] git-kv (0.0.3)

hannesm:
A Mirage_kv implementation using git

- Project page: <a href="https://github.com/roburio/git-kv">https://github.com/roburio/git-kv</a>

##### CHANGES:

- Allow author, author_email, and message being specified in `change_and_push`
  (roburio/git-kv#28 @hannes)

PR #22697, opened 2022-12-16
[new release] caldav (0.2.1)

hannesm:
A CalDAV server

- Project page: <a href="https://github.com/roburio/caldav">https://github.com/roburio/caldav</a>
- Documentation: <a href="https://roburio.github.io/caldav/">https://roburio.github.io/caldav/</a>

##### CHANGES:

* Use Mirage-kv 6.0.0 API (roburio/caldav#32 @hannesm)
* Log more write errors (roburio/caldav#32 @hannesm)
* Refine Webdav_fs.batch to return a ('a, [> `Msg of string ]) result io, so
  it can fail (roburio/caldav#32 @hannesm)
* unikernel: use git-kv instead of irmin-mirage-git (roburio/caldav#32 @hannesm)

### DIFF
--- a/packages/caldav/caldav.0.2.1/opam
+++ b/packages/caldav/caldav.0.2.1/opam
@@ -1,0 +1,72 @@
+opam-version: "2.0"
+maintainer: [
+  "Stefanie Schirmer @linse"
+  "Hannes Mehnert"
+]
+authors: [
+  "Stefanie Schirmer @linse"
+  "Hannes Mehnert"
+]
+homepage: "https://github.com/roburio/caldav"
+bug-reports: "https://github.com/roburio/caldav/issues"
+dev-repo: "git+https://github.com/roburio/caldav.git"
+tags: ["org:mirage" "org:robur"]
+doc: "https://roburio.github.io/caldav/"
+license: "ISC"
+
+build: [
+  ["dune" "subst"] {dev}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+]
+
+depends: [
+  "ocaml" {>= "4.08.0"}
+  "dune" {>= "1.3"}
+  "alcotest" {with-test & >= "0.8.5"}
+  "ounit" {with-test & >= "2.0.0"}
+  "mirage-random-test" {with-test}
+  "tcpip" {with-test & >= "3.7.0"}
+  "mirage-clock-unix" {with-test & >= "2.0.0"}
+  "mirage-kv-mem" {with-test & >= "2.0.0"}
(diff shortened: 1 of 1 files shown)


hannesm:
CI failures are unrelated (same lwt issue as #22695), fine to merge :)

packages/caldav/caldav.0.2.1/opam
```diff
@@ -0,0 +1,72 @@
...
+bug-reports: "https://github.com/roburio/caldav/issues"
+dev-repo: "git+https://github.com/roburio/caldav.git"
+tags: ["org:mirage" "org:robur"]
+doc: "https://roburio.github.io/caldav/"
+license: "ISC"
+
+build: [
+  ["dune" "subst"] {dev}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
```
kit-ty-kate, line 20:
```suggestion
  ["dune" "runtest" "-p" name "-j" jobs] {with-test & os != "macos"} # Local network access is forbidden in the macos sandbox
```

kit-ty-kate:
Thanks. Could you return the fix to the opam file upstream?

hannesm:
Thanks for merging.

> Could you return the fix to the opam file upstream?

To be honest, I'm not sure. What is the state of "opam sandbox" and macOS? Should we really annotate random packages with "test not available on macos", just to remove that again in the future (once opam sandbox on macOS is working nicely)?

By the time, the opam sandbox on macOS is in better shape, will there be a cleanup and some added "&& opam < 2.y.z" added? Any guideline would be helpful.

Would a change also lead OCaml-CI to not execute the tests on macOS (if the answer is yes, I'd be in favour to keep the tests to be executed for the development repository).

kit-ty-kate:
> To be honest, I'm not sure. What is the state of "opam sandbox" and macOS? Should we really annotate random packages with "test not available on macos", just to remove that again in the future (once opam sandbox on macOS is working nicely)?
> 
> By the time, the opam sandbox on macOS is in better shape, will there be a cleanup and some added "&& opam < 2.y.z" added? Any guideline would be helpful.

This is just the state of sandboxing on macOS in general. I don’t see the networking magically working in the next 10 years personally (Apple doesn’t care about bugs or missing features in their CLI tools and already told me they won’t do anything about it)

> Would a change also lead OCaml-CI to not execute the tests on macOS (if the answer is yes, I'd be in favour to keep the tests to be executed for the development repository).

No that’s only when the tests are ran through opam. OCaml-CI only runs dune.

hannesm:
Thanks for your reply @kit-ty-kate. I'll push the change upstream.

PR #22699, opened 2022-12-16
OCaml 5 manual package

Octachron:
This PR adds the OCaml manual package for 5.0.0 now that I found the time to generate the manual and opam file.


kit-ty-kate:
Thanks!

PR #22703, opened 2022-12-16
Correct the OCaml 5 bytecode no-flat-float-array packages

dra27:
Corrects the previous PR, which didn't encode the version/architecture constraints correctly


kit-ty-kate:
Thanks a lot!

PR #22704, opened 2022-12-16
Mark all versions of llvm as not compatible with OCaml 5.0

kit-ty-kate:
For some reason only llvm >= 10 was marked as such.

cc @alan-j-hu do you have any updates on your LLVM for OCaml 5.0 work by any chance?

### DIFF
--- a/packages/llvm/llvm.3.4/opam
+++ b/packages/llvm/llvm.3.4/opam
@@ -22,6 +22,10 @@ depends: [
   "ocaml"
   "ocamlbuild" {build}
 ]
+conflicts: [
+  "base-nnp"
+  "ocaml-option-nnpchecker"
+]
 install: ["sh" "./compile.sh" "install" "3.4" make prefix lib]
 synopsis: "The official LLVM binding"
 description: "Note: LLVM should be installed first."
--- a/packages/llvm/llvm.3.5/opam
+++ b/packages/llvm/llvm.3.5/opam
@@ -22,6 +22,10 @@ depends: [
   "ocaml"
   "ocamlbuild" {build}
 ]
+conflicts: [
+  "base-nnp"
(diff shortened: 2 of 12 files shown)


alan-j-hu:
I wrote a patch and got it passing the tests with the OCaml 5 beta. I added a bunch of people as reviewers and some of them resigned. @jberdine emailed me in October and let me know that he was busy but would hopefully find time to check out my changes "next week," saying that the existing tests weren't good for such a massive change. However, he hasn't touched the patch at all. I sent an email reply reminding him on December 4, and pinged him on Phabricator a few days ago.

I'm wondering who else could review my code, find any issues, and get it accepted?

kit-ty-kate:
@alan-j-hu if there is a way to backport your patch to LLVM 14 i could test it right away with what https://github.com/kit-ty-kate/labrys uses from the llvm binding

alan-j-hu:
What's the proper way of installing the LLVM bindings from source?

I ran `cmake -P bindings/ocaml/cmake_install.cmake` https://github.com/llvm/llvm-project/tree/main/llvm/bindings/ocaml, which installed them under `.opam/5.0.0/lib/ocaml`. I moved the `llvm` directory one directory up, and moved the `META` files such as `META.llvm` there as well. However, OCaml still can't find the LLVM package.

kit-ty-kate:
The META files are in the wrong directory. They should be in the parent directory from this one

alan-j-hu:
Thanks! It looks like Labrys depends on `msgpack`, which is not available for OCaml 5? Ideally I'd be testing this with OCaml 5.

Also, for some reason, attempting to download the `llvm` package from opam (NOT the one that I built locally) hangs forever on the download step... I don't know if this is just happening for me or if it's reproducible.

PR #22718, opened 2022-12-18
lua-ml: add a warning message about parallelism problems for OCaml 5.0.0

dmbaturin:
This PR may be misguided, please close it if it is and let me know.

Lua-ML does build with OCaml 5.0.0. However, it only works with it as long as the code is single-threaded. If someone tries to run Lua code in more than one domain, that will blow up due to hidden mutable state in the interpreter that causes race conditions and random, weird failures.

I wonder if it's better to make the package unavailable for 5.0.0 until those issues are fixed to avoid misleading people into thinking that it's good for multicore (until it's actually good for it).

Or is there another way to mark a package multicore-unsafe?

### DIFF
--- a/packages/lua-ml/lua-ml.0.9.1/opam
+++ b/packages/lua-ml/lua-ml.0.9.1/opam
@@ -22,3 +22,7 @@ url {
     "sha512=5cb56cd96dbfae42835b3f10f275753685575472ace518e351e62e50a01c07a510203be21644fd3429680b926ecd9905d4dca1e101b762c442f7c6dc00b0f0ae"
   ]
 }
+post-messages: [
+ "Lua-ML does build with OCaml 5.0.0, however, it only works with it as long as the code is single-threaded: \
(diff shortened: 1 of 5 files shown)


kit-ty-kate:
I think it's ok to leave it as-is. Thread-safety is usually a per-function/module property, not really something that a whole package carries or even OCaml 5.0 as a whole for that matter. For example `Queue`, `Stack` and `Hashtbl` from the standard library are not thread-safe, even in OCaml 5.0.

mseri:
Would it be worth to mention it in the description or as a post-install message?

dmbaturin:
@mseri I think it's a good idea. Any pointers how to implement that, and if it can be done only for >5.0.0?

mseri:
For the description it is just about updating the description in the opam file, for the post-install messages you need to add to the file something like
```
post-messages: [
 "Lua-ML does build with OCaml 5.0.0, however, it only works with it as long as the code is single-threaded: trying to execute Lua code in more than one domain may incur in confusing, unpredictable failures." {ocaml:version >= "5.0"}
]
```

mseri:
Thanks

PR #22719, opened 2022-12-19
Package ocaml.4.14.0

jonahbeckford:
### `ocaml.4.14.0`
The OCaml compiler (virtual package)
This package requires a matching implementation of OCaml,
and polls it to initialise specific variables like `ocaml:native-dynlink`

---
* Homepage: https://ocaml.org
* Bug tracker: https://github.com/ocaml/opam-repository/issues

---

### DIFF
--- a/packages/ocaml/ocaml.4.14.0/opam
+++ b/packages/ocaml/ocaml.4.14.0/opam
@@ -9,7 +9,8 @@ depends: [
   "ocaml-config" {>= "2"}
   "ocaml-base-compiler" {>= "4.14.0~" & < "4.14.1~"} |
   "ocaml-variants" {>= "4.14.0~" & < "4.14.1~"} |
-  "ocaml-system" {>= "4.14.0" & < "4.14.1~"}
+  "ocaml-system" {>= "4.14.0" & < "4.14.1~"} |
+  "dkml-base-compiler" {>= "4.14.0~" & < "4.14.1~"}
 ]
 setenv: [
   [CAML_LD_LIBRARY_PATH = "%{_:stubsdir}%"]


packages/ocaml/ocaml.4.14.0/opam
```diff
@@ -9,7 +9,8 @@ depends: [
   "ocaml-config" {>= "2"}
   "ocaml-base-compiler" {>= "4.14.0~" & < "4.14.1~"} |
   "ocaml-variants" {>= "4.14.0~" & < "4.14.1~"} |
-  "ocaml-system" {>= "4.14.0" & < "4.14.1~"}
+  "ocaml-system" {>= "4.14.0" & < "4.14.1~"} |
+  "dkml-base-compiler" {>= "4.14.0~" & < "4.14.1~"}
```
mseri, line 13: I don’t know about this, it would trigger a recompilation of all switches. Also why this and not something on the line of ocaml-variant-dkml?

jonahbeckford: The `dkml-base-compiler` is already present for `ocaml.4.12.1` in https://github.com/ocaml/opam-repository/pull/21704 .

There were prior discussions about how to package it up (ie. a new base-compiler rather than variants) that aren't part of the PR. I followed the suggestions given to me ... and sorry I don't remember the exact details!

jonahbeckford: Whatever happens, it would be good going forward for all new ocaml versions to just include `dkml-base-compiler`, to avoid the recompilation cost.

mseri: Thanks for the clarification. I generally wait for @kit-ty-kate and other compiler maintainers to have a look before doing anything on compiler-related PRs, so I'll wait for their thumbs up

kit-ty-kate: recompilation only happens with opam < 2.2 but in any case it should be fine now that 4.14.1 was released.
@jonahbeckford would it be ok to wait a couple more days for the docker images to have picked up the 4.14.1 change and for ocaml-ci (which still uses opam 2.1, as opposed to opam-repo-ci which uses 2.2) to pick up this change as well? Otherwise it'll recompile 4.14 on all its jobs and that'll hurt the cluster.

This is dependent on https://github.com/ocurrent/ocaml-version/pull/56 cc @dra27

jonahbeckford: Yep, this can wait!

dra27: The base images are rebuilt, so I think this is good to go (possibly modulo the Emacs-removal of the EOL-at-EOF 😉)

Having already added this for 4.12.1, I'm wondering if `dkml-base-compiler` should simply be being added to all new `ocaml` packages at this point. The way this is implemented at the moment (where the link between the virtual `ocaml` package and `dkml-base-compiler` has to be specified in `ocaml` instead of via a `provides`-style field in `dkml-base-compiler`), we obviously have to pause to consider whether a new package gets added to the disjunction. But we've done that already in 4.12.1! If they're added pre-emptively to others, that means that when @jonahbeckford updates the compiler in Diskuv, there's no churn or concern needed in opam-repository.

Concretely, that means I'd suggest we add the package to `ocaml.5.0.1` and `ocaml.5.1.0` now (that only affects trunk builds). We _could_ go further and also add it to `ocaml.5.0.0` and `ocaml.4.14.1` now - in terms of the cluster, that just means that we should trigger a manual rebuild of the base images as soon as the PR is merged (and then manually pull new images in ocaml-ci once they're built). That would obviously trigger a lot of rebuilds for users, so it's not as obvious a thing to do.

@Octachron and @kit-ty-kate - what do you think?

kit-ty-kate:
Thanks!

packages/ocaml/ocaml.4.14.0/opam
kit-ty-kate: Sounds good to me

PR #22721, opened 2022-12-19
Package kkmarkdown.1.0.0

kkeundotnet:
kkmarkdown is (yet another) markdown engine that is,

* to avoid XSS attack, by supporting limited markdown syntax,
* to run the same markdown engine in both server and client sides, thanks to js_of_ocaml.

### DIFF
--- a/packages/kkmarkdown/kkmarkdown.1.0.0/opam
+++ b/packages/kkmarkdown/kkmarkdown.1.0.0/opam
@@ -3,8 +3,8 @@ synopsis: "A safe markdown engine"
 description: """
 The goals of kkmarkdown are,
 
-* To avoid XSS attack, by supporting limited markdown syntax.
-* To run the same markdown engine in both server and client sides, thanks to js_of_ocaml.
+* to avoid XSS attack, by supporting limited markdown syntax,
+* to run the same markdown engine in both server and client sides, thanks to js_of_ocaml.
 """
 maintainer: "kkeundotnet <kkeundotnet@gmail.com>"
 authors: "kkeundotnet <kkeundotnet@gmail.com>"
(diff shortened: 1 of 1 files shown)


packages/kkmarkdown/kkmarkdown.1.0.0/opam
```diff
@@ -0,0 +1,33 @@
...
+* To run the same markdown engine in both server and client sides, thanks to js_of_ocaml.
+"""
+maintainer: "kkeundotnet <kkeundotnet@gmail.com>"
+authors: "kkeundotnet <kkeundotnet@gmail.com>"
+license: "Unlicense"
+homepage: "https://github.com/kkeundotnet/kkmarkdown"
+bug-reports: "https://github.com/kkeundotnet/kkmarkdown/issues"
+dev-repo: "git+https://github.com/kkeundotnet/kkmarkdown.git"
+depends: [
+  "dune" {build}
```
mseri, line 18:
```suggestion
  "dune" {>= "3.6"}
```
According to the dune-project file. Note that due to a bug in dune (https://github.com/ocaml/dune/issues/2147), it should never be tagged as a build dependency.

packages/kkmarkdown/kkmarkdown.1.0.0/opam
```diff
@@ -0,0 +1,33 @@
+opam-version: "2.0"
+name: "kkmarkdown"
```
mseri, line 2:
```suggestion
```

packages/kkmarkdown/kkmarkdown.1.0.0/opam
```diff
@@ -0,0 +1,33 @@
...
+version: "1.0.0"
```
mseri, line 3:
```suggestion
```

packages/kkmarkdown/kkmarkdown.1.0.0/opam
kkeundotnet: Thank you. Good to know that.

kit-ty-kate:
Thanks

Issue #22723, opened 2022-12-19
5.0.0+trunk package broken


jmid:
The version bump to 5.0.1 in #22687 broke the 5.0.0+trunk package.

To recreate:
```
$ opam update
$ opam switch create ocaml-variants.5.0.0+trunk

<><> Installing new switch packages <><><><><><><><><><><><><><><><><><><><><><>
Switch invariant: ["ocaml-variants" {= "5.0.0+trunk"}]

<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
∗ installed base-bigarray.base
∗ installed base-threads.base
∗ installed base-unix.base
⬇ retrieved ocaml-variants.5.0.0+trunk  (https://github.com/ocaml/ocaml/archive/5.0.tar.gz)
∗ installed ocaml-variants.5.0.0+trunk
∗ installed ocaml-config.3
[ERROR] The compilation of ocaml.5.0.0 failed at "ocaml
        /home/jmi/.opam/ocaml-variants.5.0.0+trunk/share/ocaml-config/gen_ocaml_config.ml 5.0.0 ocaml".

#=== ERROR while compiling ocaml.5.0.0 ========================================#
# context     2.1.2 | linux/x86_64 | ocaml-variants.5.0.0+trunk | https://opam.ocaml.org#789990a7
# path        ~/.opam/ocaml-variants.5.0.0+trunk/.opam-switch/build/ocaml.5.0.0
# command     ~/.opam/opam-init/hooks/sandbox.sh build ocaml /home/jmi/.opam/ocaml-variants.5.0.0+trunk/share/ocaml-config/gen_ocaml_config.ml 5.0.0 ocaml
# exit-code   1
# env-file    ~/.opam/log/ocaml-3717-84ccb7.env
# output-file ~/.opam/log/ocaml-3717-84ccb7.out
### output ###
# OCaml version mismatch: 5.0.1, expected 5.0.0



<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
┌─ The following actions failed
│ λ build ocaml 5.0.0
└─ 
┌─ The following changes have been performed (the rest was aborted)
│ ∗ install base-bigarray  base
│ ∗ install base-threads   base
│ ∗ install base-unix      base
│ ∗ install ocaml-config   3
│ ∗ install ocaml-variants 5.0.0+trunk
└─ 
[...]

PR #22724, opened 2022-12-19
opam packages for OCaml 4.14.1

Octachron:
This PR adds the package for tomorrow release of OCaml 4.14.:

- ocaml-base-compiler.4.14.1
- ocaml-system.4.14.1
- ocaml-variants.4.14.1+options
- ocaml-src.4.14.1

and the `ocaml.4.14.2` meta-package.


kit-ty-kate:
I think we also need to move the `ocaml-variants.4.14.1+trunk` package to `ocaml-variants.4.14.2+trunk`

kit-ty-kate:
ah, apparently it was also missing from the 5.0.0 release #22723

Octachron:
The OCaml version should be changed in the `synopsis` and `depends` fields.

kit-ty-kate:
Thanks a lot!

PR #22728, opened 2022-12-20
new file:   packages/parany/parany.13.0.1/opam

UnixJunkie:
Only textual changes: ocamldoc was out of date; same for some text in the opam file.

### DIFF
--- a/packages/parany/parany.13.0.1/opam
+++ b/packages/parany/parany.13.0.1/opam
@@ -1,0 +1,39 @@
+opam-version: "2.0"
+maintainer: "unixjunkie@sdf.org"
+authors: "Francois Berenger"
+license: "LGPL-2.0-or-later"
+homepage: "https://github.com/UnixJunkie/parany"
+bug-reports: "https://github.com/UnixJunkie/parany/issues"
+dev-repo: "git+https://github.com/UnixJunkie/parany.git"
+depends: [
+  "domainslib" {>= "0.5.0"}
+  "dune" {>= "1.6.0"}
+  "ocaml" {>= "5.0.0"}
+]
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "build" "-p" name "-j" jobs "src/test.exe"] {with-test}
+  ["./test.sh"] {with-test}
+]
+synopsis: "Parallelize any computation"
+description: """
+Generalized map reduce for parallel computers (not distributed computing).
+Can process in parallel an infinite stream of elements.
+
+Can process a very large file in parallel on a multicore computer;
+provided there is a way to cut your file into independent blocks
+(the 'demux' function).
+The processing function is called 'work'.
+The function gathering the results is called 'mux'.
+The chunk size (number of items) processed by one call to the 'work' function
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
The tests fail on Alpine:
```
#=== ERROR while compiling parany.13.0.1 ======================================#
# context              2.2.0~alpha~dev | linux/x86_64 | ocaml-base-compiler.5.0.0 | pinned(https://github.com/UnixJunkie/parany/archive/v13.0.1.tar.gz)
# path                 ~/.opam/5.0/.opam-switch/build/parany.13.0.1
# command              ~/.opam/opam-init/hooks/sandbox.sh build ./test.sh
# exit-code            1
# env-file             ~/.opam/log/parany-7-3abaeb.env
# output-file          ~/.opam/log/parany-7-3abaeb.out
### output ###
# ./test.sh: line 7: _build/default/src/test_parmap.exe: No such file or directory
# diff: can't create temp file '/tmp/difXXXXXX': Read-only file system
# diff: can't create temp file '/tmp/difXXXXXX': Read-only file system
# diff: can't create temp file '/tmp/difXXXXXX': Read-only file system
# diff: can't create temp file '/tmp/difXXXXXX': Read-only file system
```
`src/test_parmap.exe` is indeed never built. Is that expected?

Also it’s out of your reach but it seems that Busybox’s implementation of `diff` isn’t POSIX-complient.... 😵‍💫 
* https://git.busybox.net/busybox/tree/editors/diff.c#n733
* https://en.wikipedia.org/wiki/TMPDIR

Is there any chance you could use dune’s builtin tests to replace this bash script? Dune has a diff primitive that you can use instead: https://dune.readthedocs.io/en/stable/concepts.html?highlight=diff#diffing-and-promotion

UnixJunkie:
My suggestion: we remove the two {with-test} conditioned lines in the opam file.
Since I would not release this software unless I am ready to bet my house on it.

UnixJunkie:
The long explanation is that there is a Makefile, then there is a test.sh script.
https://github.com/UnixJunkie/parany/blob/master/Makefile
https://github.com/UnixJunkie/parany/blob/master/test.sh
I don't do everything with dune and ocaml, and I don't plan to.

packages/parany/parany.13.0.1/opam
```diff
@@ -0,0 +1,39 @@
...
+dev-repo: "git+https://github.com/UnixJunkie/parany.git"
+depends: [
+  "domainslib" {>= "0.5.0"}
+  "dune" {>= "1.6.0"}
+  "ocaml" {>= "5.0.0"}
+]
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "build" "-p" name "-j" jobs "src/test.exe"] {with-test}
+  ["./test.sh"] {with-test}
```
mseri, line 16:
```suggestion
  ["./test.sh"] {with-test & os-distribution != "alpine"}
```

mseri:
It just fails on alpine, we'll just disable it

UnixJunkie:
thanks, changed reflected upstream

PR #22736, opened 2022-12-20
Datakit-client needs an old version of Lwt

samoht:
Otherwise it fails with:

```
 File "api/ocaml/datakit_client.mli", line 141, characters 32-45:
 141 |   type +'a result = ('a, error) Result.result Lwt.t
                                       ^^^^^^^^^^^^^
 Error: Unbound type constructor Result.result
```

### DIFF
--- a/packages/datakit-client/datakit-client.0.12.0/opam
+++ b/packages/datakit-client/datakit-client.0.12.0/opam
@@ -16,7 +16,7 @@ depends: [
   "astring"
   "result"  {< "1.5"}
   "fmt"
-  "lwt"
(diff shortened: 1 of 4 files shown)


samoht:
I'm not super happy with dde464a - the package compiles fine on old version of OCaml, there's just a package in the chain that brings `Result` in scope but I cannot test it on my mac m1 as I cannot compile old versions of OCaml there.

mseri:
Seems to be resolved with your latest change withot a need to restrict ocaml. Feel free to merge if you are happy with it

PR #22737, opened 2022-12-20
Old version of depyt needs an old version of cstruct

samoht:
Otherwise it fails with:

```
  File "src/depyt.ml", line 756, characters 15-37:
  756 |     | C buf -> Cstruct.blit_to_string buf pos str 0 len
                       ^^^^^^^^^^^^^^^^^^^^^^
  Error: Unbound value Cstruct.blit_to_string
  Hint: Did you mean blit_from_string?
```


kit-ty-kate:
Thanks

PR #22752, opened 2022-12-20
Fix missing patch for ocaml-base-compiler 4.01.0

balsoft:
A patch on the URL [1] seems to return 404, as the account of jeremiedimino has been removed from GitHub. This means that ocaml-base-compiler.4.01.0 is failing, and so do all the packages which still depend on it (a transitive total of 29 packages, if my counting is correct).

https://github.com/Niols helped me locate the missing commits on Software Heritage's website [2]. There doesn't seem to be an easy way to just download the total diff from there, so instead we put the diff here directly.

The md5 checksum appears to be different. I'm not sure why that is. Potentially, the format of the diff is different.

It seems that the ocaml-base-compiler.4.01.0 now builds on both Linux and macOS.

[1]: https://github.com/jeremiedimino/ocaml/compare/bd7fa181cb64742c3b6cbb8ee13436554eb18cd7...fix-clang-build.diff
[2]: https://archive.softwareheritage.org/browse/revision/f2fd13b96528323270c35c4a221032beb7c4e91f

### DIFF
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.4.01.0/files/fix-clang-build.patch
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.4.01.0/files/fix-clang-build.patch
@@ -1,0 +1,20 @@
+diff --git a/configure b/configure
+index d45e88f..25d872b 100755
+--- a/configure
++++ b/configure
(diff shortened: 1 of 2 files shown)


balsoft:
Sorry, didn't notice this was already fixed in master :)

PR #22755, opened 2022-12-21
Package stdlib-random.1.0.0

Octachron:
### `stdlib-random.1.0.0`
Versioned Random module from the OCaml standard library
The stdlib-random package provides a stable and compiler-independent
implementation of all the PRNGs used in the Random module.
Those PRNGs are available in the various libraries:
- stdlib-random.v3: OCaml 3.07 to 3.11 PRNG
- stdlib-random.v4: OCaml 3.12 to 4.14 PRNG
- stdlib-random.v5: current OCaml 5.0 PRNG
- stdlib-random.v5o: pure OCaml version of the OCaml 5 PRNG
All those libraries can be used together and the signature of their
Random$n module has been extended to the latest signature whenever possible.

---
* Homepage: https://github.com/ocaml/stdlib-random
* Source repo: git+https://github.com/ocaml/stdlib-random.git
* Bug tracker: https://github.com/ocaml/stdlib-random/issues

---


packages/stdlib-random/stdlib-random.1.0.0/opam
```diff
@@ -0,0 +1,48 @@
...
+Random$n module has been extended to the latest signature whenever possible.
+"""
+maintainer: ["Florian Angeletti, <octa@polychoron.fr>"]
+authors: ["Damien Doligez" "Xavier Leroy"]
+license: "LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception"
+homepage: "https://github.com/ocaml/stdlib-random"
+bug-reports: "https://github.com/ocaml/stdlib-random/issues"
+depends: [
+  "dune" {>= "2.7"}
+  "cppo" {dev}
```
kit-ty-kate, line 22: It seems to be needed all the time. Not just for the dev builds
```suggestion
  "cppo" {>= "1.1.0"}
```

Octachron: You are right, I forgot to update the dependency at some point. Thanks for the fix!

kit-ty-kate:

PR #22765, opened 2022-12-21
[new release] opam-monorepo (0.3.5)

Leonidas-from-XIV:

- Project page: <a href="https://github.com/tarides/opam-monorepo">https://github.com/tarides/opam-monorepo</a>
- Documentation: <a href="https://tarides.github.io/opam-monorepo">https://tarides.github.io/opam-monorepo</a>

### Changed

- Treat packages with no build commands as if they can be built with dune (tarides/opam-monorepo#355,
  @gridbugs)

### Fixed

- Fix resolving refs of locally pinned repositories (tarides/opam-monorepo#326, tarides/opam-monorepo#332, @hannesm,
  @Leonidas-from-XIV)
- Read the `compiler` flag from OPAM metadata thus classifying more packages
  correctly as base packages (tarides/opam-monorepo#328, @Leonidas-from-XIV)
- Fix bug where dev repo urls ending with a "/" would result in
  `opam monorepo pull` placing package source code directly inside the duniverse
  directory instead of in a subdirectory of the duniverse directory (tarides/opam-monorepo#359,
  @gridbugs)

### DIFF
--- a/packages/mirage/mirage.4.2.0/opam
+++ b/packages/mirage/mirage.4.2.0/opam
@@ -27,6 +27,7 @@ depends: [
   "logs"
   "mirage-runtime" {= version}
   "opam-monorepo" {>= "0.3.2"}
+  "opam-monorepo" {< "0.3.5" & with-test}
   "alcotest" {with-test}
   "fmt" {>= "0.8.7" & with-test}
 ]
--- a/packages/mirage/mirage.4.2.1/opam
+++ b/packages/mirage/mirage.4.2.1/opam
@@ -28,6 +28,7 @@ depends: [
   "logs"
   "mirage-runtime" {= version}
   "opam-monorepo" {>= "0.3.2"}
+  "opam-monorepo" {< "0.3.5" & with-test}
   "alcotest" {with-test}
   "fmt" {>= "0.8.7" & with-test}
 ]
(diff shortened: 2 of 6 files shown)


raphael-proust:
CI errors:
- Some revdeps: mirage on 4.14 and 5.0. I don't know if it's related to this PR.
- Archlinux system package installation errors: should be fixed when updating the base image which should include some new maintainer keys. Not blocking IMO.
- opensuse system package installation error: I don't know opensuse well enough to tell, but that seems to be a transient network issue or maybe a docker image issue.

Leonidas-from-XIV:
Looks like the Arch Linux and OpenSUSE tests are unrelated issues.

The mirage failures are indeed unexpected and hint at the fact that the test-suite of opam-monorepo is not exercising the case that mirage is running into because I don't see how the changes between 0.3.4 and 0.3.5 would break it. It's possible something went wrong in this release.

Leonidas-from-XIV:
Ok, after investigating @kit-ty-kate's solution of restricting the version is correct. It's not really a bug, rather a conscious change in `opam-monorepo.0.3.5` that packages without build instructions are considered "virtual" and thus skipped. Mirage ran into this issue because the dummy packages it tests opam-monorepo against do not contain `build` stanzas.

I've made a PR to `mirage` to fix it in the future: https://github.com/mirage/mirage/pull/1383

Leonidas-from-XIV:
The conan failures are due to OCaml 5 rejecting the formatters in the conan test, so I've added a constraint: https://github.com/ocaml/opam-repository/pull/22790

PR #22690, opened 2022-12-16
[new release] optint (0.3.0)

dinosaure:
Efficient integer types on 64-bit architectures

- Project page: <a href="https://github.com/mirage/optint">https://github.com/mirage/optint</a>
- Documentation: <a href="https://mirage.github.io/optint/">https://mirage.github.io/optint/</a>

##### CHANGES:

- Add infix operators for bitwise operations (@reynir, mirage/optint#23)
- Add a deprecation about old infix operators
  They will be removed at the next minor release

PR #22787, opened 2022-12-22
atd 1.x is not compatible with ocaml 5

mseri:
Explicitly uses the Pervasives module.

Seen in the revdeps of https://github.com/ocaml/opam-repository/pull/22690


mseri:
1.13 had already the upper bound

PR #22793, opened 2022-12-22
[new release] xcursor (0.1.0)

yilinwei:
Xcursor

- Project page: <a href="https://github.com/yilinwei/ocaml-xcursor">https://github.com/yilinwei/ocaml-xcursor</a>
- Documentation: <a href="https://yilinwei.github.io/ocaml-xcursor">https://yilinwei.github.io/ocaml-xcursor</a>

##### CHANGES:

Initial release.

### DIFF
--- a/packages/xcursor/xcursor.0.1.0/opam
+++ b/packages/xcursor/xcursor.0.1.0/opam
@@ -4,14 +4,14 @@ description: "A pure implementation of Xcursor in Ocaml"
 maintainer: ["Yilin Wei"]
 authors: ["Yilin Wei"]
 license: "MIT"
-available: [ arch != "arm32" ]
+available: [ arch != "arm32" & arch != "x86_32" ]
 tags: ["desktop" "gui" "cursor" "xcursor"]
 homepage: "https://github.com/yilinwei/ocaml-xcursor"
 doc: "https://yilinwei.github.io/ocaml-xcursor"
 bug-reports: "https://github.com/yilinwei/ocaml-xcursor/issues"
 depends: [
   "dune" {>= "3.6"}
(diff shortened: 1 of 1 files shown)


yilinwei:
Would it be possible to wait until I've regenerated the `.opam` file in the main repository and artifacts to be the same as this one please? Otherwise it will be confusing.

Should I open it as a separate PR?

mseri:
No need to create a new PR, you can even force push here if you want

mseri:
Let me know when I can merge, this looks good to me.

By the way, opam will use the opam file from this repository as source of truth, in general with time upper bounds are added and they go out of sync. As long as the dev branch is kept updated, you don't need to regenerate the release unless you need to modify the code

yilinwei:
@mseri 

Should be good to go now, thanks!

packages/xcursor/xcursor.0.1.0/opam
```diff
@@ -0,0 +1,42 @@
+opam-version: "2.0"
+synopsis: "Xcursor"
+description: "A pure implementation of Xcursor in Ocaml"
```
kit-ty-kate, line 3: a synopsis that doesn’t tell anything isn’t really useful. The description you had looks more like a synopsis than a description so you can just make this one the synopsis:
```suggestion
synopsis: "A pure implementation of Xcursor in Ocaml"
```

yilinwei: Thanks! I will take a look at some better descriptions and update accordingly.

yilinwei: Does the new description fit the criteria?

kit-ty-kate:
Thanks!

PR #22801, opened 2022-12-25
minimal support for ocaml 5.0.0

chetmurthy:
Mostly just updating the worked example of OCaml AST migrations.


chetmurthy:
This looks like corrupt packages on archlinux, and known prereq problems on macos-homebrew.

mseri:
thanks

PR #22806, opened 2022-12-27
Constrain DKML to < ocaml.5.0.0~

jonahbeckford:
Nothing in DKML is ready for 5.0.0, especially since the MSVC and Android ARM32 ports of OCaml 5.0.0 are not available and none of the cross-compiling bits have been ported.

Fixes revdeps in #22805


mseri:
Thanks

PR #22809, opened 2022-12-29
Package ulex-camlp5.1.4

sacerdot:


sacerdot:
All the failures are due to camlp5 (a dependency) that fails to build even if the package is in opam.
I have no idea what to do.

PR #22810, opened 2022-12-29
Add conf-assimp package (checks system for assimp)

geoffder:
This is my first time making a `conf` package, so I'm not sure that I've done everything just right. I noticed that there is an `assimp` bindings package, but for my current purposes, I'm trying to make sure the library is installed for linking during compilation of a vendored C++ dependency.

### DIFF
--- a/packages/conf-assimp/conf-assimp.1/opam
+++ b/packages/conf-assimp/conf-assimp.1/opam
@@ -1,0 +1,28 @@
+opam-version: "2.0"
+maintainer: "geoffderosenroll@gmail.com"
+authors: ["Geoff deRosenroll"]
+homepage: "https://github.com/assimp/assimp"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: "BSD-3-Clause"
+build: [
+  ["assimp" "--help"] {os != "openbsd"}
+  ["/usr/local/bin/assimp" "--help"] {os = "openbsd"}
+]
+depexts: [
+  ["libassimp-dev"] {os-family = "debian"}
+  ["assimp"] {os-distribution = "arch"}
+  ["assimp-devel"] {os-distribution = "fedora"}
+  ["lib64assimp-devel"] {os-distribution = "mageia"}
+  ["assimp"] {os-distribution = "alpine"}
+  ["assimp"] {os-distribution = "nixos"}
+  ["assimp-5.0.0"] {os = "openbsd"}
+  ["assimp"] {os = "macos" & os-distribution = "homebrew"}
+  ["assimp"] {os = "macos" & os-distribution = "macports"}
+  ["assimp"] {os = "freebsd"}
+  ["assimp"] {os = "win32" & os-distribution = "vcpkg"}
+]
+synopsis: "Check if assimp (Open Asset Import Library) is installed"
+description: """
+This package can only install if the assimp package (Open Asset Import Library)
+is installed on the system."""
+flags: conf
--- a/packages/conf-assimp/conf-assimp.1/opam
+++ b/packages/conf-assimp/conf-assimp.1/opam
@@ -13,7 +13,7 @@ depexts: [
   ["libassimp-dev"] {os-family = "debian"}
   ["assimp"] {os-distribution = "arch"}
   ["assimp-devel"] {os-distribution = "fedora"}
-  ["assimp-devel.x86_64"] {os-distribution = "ol"}
+  ["assimp-devel"] {os-distribution = "ol"}
   ["assimp"] {os-distribution = "suse"}
   ["libassimp-dev"] {os-family = "suse"}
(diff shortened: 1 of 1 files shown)


geoffder:
It seems like the arch 5.0 issue is probably a CI thing based on the other PRs, but the Oracle one I either don't have the right name, or it isn't available in the repository (I tried a couple based on what I could find on package search websites but no luck). Beyond removing the **ol** entries, is there anything I can do to get this opam ready?

packages/conf-assimp/conf-assimp.1/opam
```diff
@@ -0,0 +1,33 @@
+opam-version: "2.0"
+maintainer: "geoffderosenroll@gmail.com"
+authors: ["Geoff deRosenroll"]
+homepage: "https://github.com/assimp/assimp"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: "BSD-3-Clause"
+build: [
+  ["assimp" "--help"] {os != "openbsd"}
+  ["/usr/local/bin/assimp" "--help"] {os = "openbsd"}
```
kit-ty-kate, line 9: why is that different on openbsd?

geoffder: I have to admit that I copied that quirk from [conf-pkg-config](https://github.com/ocaml/opam-repository/blob/master/packages/conf-pkg-config/conf-pkg-config.2/opam), which I used as a reference when writing this / learning about *conf* packages.

packages/conf-assimp/conf-assimp.1/opam
```diff
@@ -0,0 +1,33 @@
...
+  ["assimp"] {os-distribution = "suse"}
+  ["libassimp-dev"] {os-family = "suse"}
+  ["lib64assimp-devel"] {os-distribution = "mageia"}
+  ["assimp"] {os-distribution = "alpine"}
+  ["assimp"] {os-distribution = "nixos"}
+  ["assimp-5.0.0"] {os = "openbsd"}
+  ["assimp"] {os = "macos" & os-distribution = "homebrew"}
+  ["assimp"] {os = "macos" & os-distribution = "macports"}
+  ["assimp"] {os = "freebsd"}
+  ["assimp"] {os = "win32" & os-distribution = "vcpkg"}
```
kit-ty-kate, line 26: This can be simplified this way. Also `assimp` is the package name for openbsd. There is no package named `assimp-5.0.0`
```suggestion
  ["assimp-utils" "libassimp-dev"] {os-family = "debian"}
  ["assimp-utils" "libassimp-dev"] {os-family = "ubuntu"}
  ["assimp"] {os-family = "arch"}
  ["assimp-devel"] {os-distribution = "fedora"}
  ["assimp-devel"] {os-distribution = "ol"}
  ["assimp" "libassimp-dev"] {os-family = "suse"}
  ["lib64assimp-devel"] {os-distribution = "mageia"}
  ["assimp"] {os-family = "alpine"}
  ["assimp"] {os-distribution = "nixos"}
  ["assimp"] {os = "openbsd"}
```

packages/conf-assimp/conf-assimp.1/opam
kit-ty-kate, line 9:
```suggestion
  ["assimp" "--help"]
```

packages/conf-assimp/conf-assimp.1/opam
```diff
@@ -0,0 +1,31 @@
...
+  ["assimp-devel"] {os-distribution = "ol"}
+  ["assimp" "libassimp-dev"] {os-family = "suse"}
+  ["lib64assimp-devel"] {os-distribution = "mageia"}
+  ["assimp"] {os-family = "alpine"}
+  ["assimp"] {os-distribution = "nixos"}
+  ["assimp"] {os = "openbsd"}
+  ["assimp"] {os = "macos" & os-distribution = "homebrew"}
+  ["assimp"] {os = "macos" & os-distribution = "macports"}
+  ["assimp"] {os = "freebsd"}
+  ["assimp"] {os = "win32" & os-distribution = "vcpkg"}
```
kit-ty-kate, line 24: Side note: opam doesn't support vcpkg currently so this part isn't used.

kit-ty-kate:
Looks good to me once the CI finishes. Sorry for the wait

packages/conf-assimp/conf-assimp.1/opam
geoffder: I suspected that might be the case, but thought the message might be useful to the user anyway. Thanks for confirming.

packages/conf-assimp/conf-assimp.1/opam
geoffder: Ah TIL thanks, I should have tried that / read the docs better.

geoffder:
No problem, thanks for the improvements!

kit-ty-kate:
Thanks! Sorry for the wait

PR #22815, opened 2022-12-31
Package ulex-camlp5.1.3

sacerdot:
Porting to ocaml 5.

NOTE: the package depends on camlp5 that fails to build in CI on some opam architectures, even if the camlp5 package has been accepted before. Please, force it somehow.


sacerdot:
As expected, the failing CI checks are due to camlp5 and not ulex-camlp5.
Can you please force ulex-camlp5 in?
Thanks

PR #22818, opened 2023-01-02
Add drom.0.8.0 and dependencies

lefessan:


packages/drom_lib/drom_lib.0.8.0/opam
```diff
@@ -0,0 +1,63 @@
...
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+  ["sh" "-c" "./scripts/after.sh build '%{name}%'"]
```
mseri, line 38: This does nothing

packages/drom_lib/drom_lib.0.8.0/opam
```diff
@@ -0,0 +1,63 @@
...
+]
+install: [
```
mseri, line 40: This does nothing

packages/drom_lib/drom_lib.0.8.0/opam
```diff
@@ -0,0 +1,63 @@
...
+  "Léo Andrès <leo.andres@ocamlpro.com>"
+]
+homepage: "https://ocamlpro.github.io/drom"
+doc: "https://ocamlpro.github.io/drom/sphinx"
+bug-reports: "https://github.com/ocamlpro/drom/issues"
+dev-repo: "git+https://github.com/ocamlpro/drom.git"
+tags: "org:ocamlpro"
+build: [
+  ["dune" "subst"] {dev}
+  ["sh" "-c" "./scripts/before.sh build '%{name}%'"]
```
mseri, line 26: This does nothing

mseri:
Can I remove those sh calls from the opam files? Unless I am mistaken, they do nothing

packages/drom_lib/drom_lib.0.8.0/opam
lefessan: Yes, but these `opam` files are automatically generated, so this line is always put in case there is a post-install script to run after installation (the idea is that such a script can be run both by `opam install`, but also by `make install` or any other installation tool).

lefessan:
You can remove these calls if you want, but they will always appear in projects generated by `drom`, so I wouldn't care too much about them...

PR #22819, opened 2023-01-02
hevea 2.36

gasche:
This version is the first to implement support for OCaml 5.

cc @maranget @Octachron (nothing to do on your side)

### DIFF
--- a/packages/hevea/hevea.2.36/opam
+++ b/packages/hevea/hevea.2.36/opam
@@ -5,7 +5,7 @@ homepage: "http://hevea.inria.fr/"
 bug-reports: "http://github.com/herd/hevea/issues/"
 doc: "http://hevea.inria.fr/doc/index.html"
 dev-repo: "git+https://github.com/maranget/hevea.git"
-license: ["QPL-1.0 WITH OCaml-LGPL-linking-exception" "LGPL-2.0-only"]
+license: ["QPL-1.0 WITH OCaml-LGPL-linking-exception"]
 build: [make "PREFIX=%{prefix}%"]
 install: [make "PREFIX=%{prefix}%" "install"]
 post-messages: [
--- a/packages/hevea/hevea.2.36/opam
+++ b/packages/hevea/hevea.2.36/opam
@@ -1,0 +1,25 @@
+opam-version: "2.0"
+maintainer: "Luc Maranget <Luc.Maranget@inria.fr>"
+authors: "Luc Maranget"
+homepage: "http://hevea.inria.fr/"
+bug-reports: "http://github.com/herd/hevea/issues/"
+doc: "http://hevea.inria.fr/doc/index.html"
+dev-repo: "git+https://github.com/maranget/hevea.git"
+license: ["QPL-1.0 WITH OCaml-LGPL-linking-exception" "LGPL-2.0-only"]
+build: [make "PREFIX=%{prefix}%"]
+install: [make "PREFIX=%{prefix}%" "install"]
+post-messages: [
+  "The file 'hevea.sty' has been installed in %{lib}%/hevea but latex won't see it by itself" {success}
+]
+depends: [
+  "ocaml" {>= "4.08.0"}
+  "ocamlbuild" {build}
+]
+conflicts: [
+  "ocaml-option-bytecode-only"
+]
+synopsis: "A quite complete and fast LATEX to HTML translator"
+url {
+  src: "https://github.com/maranget/hevea/archive/v2.36.tar.gz"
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
hevea doesn't seem to support bytecode-only (e.g. x86_32) mode:
```
# /home/opam/.opam/5.0/bin/ocamldep -modules myStack.ml > myStack.ml.depends
# ocamlopt -c -w +a-4-9-41-45-67-70 -annot -safe-string -o myLexing.cmx myLexing.ml
# ocamlopt -c -w +a-4-9-41-45-67-70 -annot -safe-string -o myStack.cmx myStack.ml
# + ocamlopt -c -w +a-4-9-41-45-67-70 -annot -safe-string -o myLexing.cmx myLexing.ml
# /bin/sh: 1: ocamlopt: not found
# Command exited with code 127.
# make: *** [Makefile:60: ocb-opt] Error 10
```

packages/hevea/hevea.2.36/opam
```diff
@@ -0,0 +1,22 @@
+opam-version: "2.0"
+maintainer: "Luc Maranget <Luc.Maranget@inria.fr>"
+authors: "Luc Maranget"
+homepage: "http://hevea.inria.fr/"
+bug-reports: "http://github.com/herd/hevea/issues/"
+doc: "http://hevea.inria.fr/doc/index.html"
+dev-repo: "git+https://github.com/maranget/hevea.git"
+license: ["QPL-1.0" "LGPL-2 with exceptions"]
```
kit-ty-kate, line 8: I'm not seeing the exception in the license for the LGPL, only for the QPL part.
```suggestion
license: ["QPL-1.0 WITH OCaml-LGPL-linking-exception" "LGPL-2.0-or-later"]
```

gasche:
Do you have a recommendation on what to do in this case? (Is there a way to say so in the opam metadata, can you point me to an example?) Or must we tweak the hevea build system? I agree that bytecode-only systems are slightly more relevant to OCaml 5 than OCaml 4, but as a hobbyist packager that only needs hevea to build the OCaml manual I would still be happy with a native-systems-only hevea package for now.

packages/hevea/hevea.2.36/opam
```diff
@@ -0,0 +1,22 @@
...
+build: [make "PREFIX=%{prefix}%"]
+install: [make "PREFIX=%{prefix}%" "install"]
+post-messages: [
+  "The file 'hevea.sty' has been installed in %{lib}%/hevea but latex won't see it by itself" {success}
+]
+depends: [
+  "ocaml" {>= "4.08.0"}
+  "ocamlbuild" {build}
+]
```
kit-ty-kate, line 17: That's fine, I was just saying that in case you wanted to fix that issue. In that case adding this conflict should work:
```suggestion
]
conflicts: [
  "ocaml-option-bytecode-only"
]
```

gasche:
Thanks @kate for the review and help. I will look at the license thing, rebase the PR, and let you know if I'm happy with the final state.

kit-ty-kate:
Ah sorry i had already pushed the code so your commit is a duplicate change that makes the opam file not work. I've reverted your commit to make it work

packages/hevea/hevea.2.36/opam
gasche: Note: after reviewing the LICENSE terms, I believe that it is actually LGPL-2.0-only, not -or-later, so I changed the package accordingly. The LGPL license only applies to "binary distributions of the package" anyway, so it is unclear to me that they actually apply to distribution through opam, but oh well, better be safe than sorry.

kit-ty-kate: >  I believe that it is actually LGPL-2.0-only, not -or-later, so I changed the package accordingly.

Are you sure? The license file does say:
```
    This library is free software; you can redistribute it and/or
    modify it under the terms of the GNU Library General Public
    License as published by the Free Software Foundation; either
    version 2 of the License, or (at your option) any later version.
```

kit-ty-kate: Ah indeed. In that case let's remove mentions of the LGPL entirely given it doesn't apply to any of the opam versions.

gasche:
Fine with me -- please feel free to merge whenever.

I some people are worried by the Hevea license (QPL is not a very good license), I think it would be fairly easy to convince the author to do a new release following more recent OCaml licensing terms (for example LGPL-2.1-or-later-with-OCaml-linking-exception).

kit-ty-kate:
Thanks!

PR #22820, opened 2023-01-02
[new release] lambdapi (2.3.0)

fblanqui:

### Added

- Export to Coq.
- (API) the rewrite engine can match on the constant `TYPE`.
- Automatic coercion insertion mechanism.
  For example, the command `coerce_rule coerce Int Float $x ↪ FloatOfInt $x;`
  can be used to instruct Lambdapi to automatically coerce integers to floats
  using the function `FloatOfInt`.

### Fixed

- Generation of metavariables through the rewriting engine.
- Application of pattern variables in rewrite rules RHS in the Dedukti
  export.
- Dedukti export: invalid Dedukti module name were not brace-quoted,
  for instance, `#REQUIRE module-name.` could be exported, while `module-name`
  is not recognised by Dedukti2. It is now exported as `#REQUIRE {|module-name|}`,
  and symbols are exported as `{|module-name|}.foo`.
- HRS and XTC exports.

### DIFF
--- a/packages/lambdapi/lambdapi.2.3.0/opam
+++ b/packages/lambdapi/lambdapi.2.3.0/opam
@@ -1,0 +1,80 @@
+opam-version: "2.0"
+synopsis: "Proof assistant for the λΠ-calculus modulo rewriting"
+description: """
+This package provides:
+- A lambdapi command for checking .lp or .dk files,
+translating .dk files to .lp files and vice versa,
+or launching an LSP server for editing .lp files.
+- A library of logic definitions and basic definitions and proofs
+on natural numbers and polymorphic lists.
+- A rich Emacs mode based on LSP (available on MELPA too).
+- A basic mode for Vim.
+- OCaml libraries.
+A VSCode extension is also available on the VSCode Marketplace.
+
+Find Lambdapi user manual on https://lambdapi.readthedocs.io/.
+
+Lambdapi provides a rich type system with dependent types.
+In Lambdapi, one can define both type and function symbols
+by using rewriting rules (oriented equations).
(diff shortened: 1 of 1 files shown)


fblanqui:
Hi. This PR is ready to be merged.

packages/lambdapi/lambdapi.2.3.0/opam
```diff
@@ -0,0 +1,80 @@
...
+  "cmdliner" {>= "1.1.0"}
+  "stdlib-shims" {>= "0.1.0"}
+  "odoc" {with-doc}
+]
+build: [
+  ["dune" "subst"] {dev}
+  [make]
+]
+install: [
+  [make "install"]
```
kit-ty-kate, line 70: I've had a look at your Makefile and I'm not sure to understand why the emacs mode has its own dune package. Why not simply integrate it into lambdapi and add an optional installation of the .elc (using `(install (files (include foo.sexp) ...`) ? This way you get the benefit of avoiding the manual installation (which is more resource intensive) and a simpler build system that doesn't rely on GNU Make or `which` (which is actually missing from the dependencies)

If you don't want/don't have time to do it right now, it's ok i can merge it, but if you can do it it would be better for the longer term maintainability of the package (non release build in dune will inevitably fail at some point due to new compiler warnings-as-error or similar things)

fblanqui: Hi. Thank you very much for your suggestions. But I am not familiar with dune enough to do that right now. So I would appreciate very much if you could merge this PR as it is now. And I will try to improve the scripts later.

PR #19855, opened 2021-10-25
Backport glibc 2.34 fix to 3.07-4.12

dra27:
With the release of Ubuntu 21.10 (Impish), we're starting to get (more) distros with Glibc 2.34+ and therefore https://github.com/ocaml/ocaml/pull/10250 is biting. I think this is as with the GCC 10 changes - we need to back-port in opam-repository in order to keep old compilers building with newer distros.

I propose pushing https://github.com/ocaml/ocaml/issues/10266 to the maintenance tips of the 4.10, 4.11 and 4.12 branches. The back-port to these versions is trivial. I'll aim to confirm this with @Octachron later and, assuming they get pushed upstream I'll amend this PR to point to ocaml/ocaml instead of dra27/ocaml.

I'm opening the draft for now in order to allow the cluster to test these packages.

I have not updated:
- alpha, beta, rc packages
- multicore 4.12 packages (cc. @ctk21 - this won't be a priority right now of course, but back-porting ocaml/ocaml#10266 looked non-trivial when I just tried... it may of course be relatively fresh in your mind from the other direction for OCaml 5.00!)

The back-port to 4.09 and earlier is not as trivial: the machinery was changed in https://github.com/ocaml/ocaml/pull/8670 to fix stack overflow in systhreads, so malloc'ing the alternate stack needs extricating from that. I intend to open a PR against 4.09 for that, just to allow (hopefully brief) review. Once that's merged to the 4.09 maintenance branch, I'd then aim to push that to previous branches and will then open another opam-repository PR like this one.

Patches prepared. Remaining tasks:
- [x] Test `opam source` on all packages
- [x] Test `opam switch create` on all the ocaml-base-compiler packages and the `+options` variants`
- [x] Submit and review patches for pushing to ocaml/ocaml release branches
- [ ] Update packages to point to ocaml/ocaml instead of dra27/ocaml

PR #22825, opened 2023-01-03
Further ARM64 fixes for OCaml 4.02-4.05

dra27:
This is an upgrade and continuation of #18245. It's a low-priority PR following on from getting the Docker base images for Debian 11 to build correctly with OCaml 5. With this PR, the last failing jobs for [Debian 11 on the base image builder](https://images.ci.ocaml.org/?distro=debian-11&) should succeed.

Since #18245 was merged, the patching mechanism for supporting building _old OCaml_ compilers using _new C_ compilers has changed, and the patches live as Git commits in the maintenance branches in [ocaml/ocaml](https://github.com/ocaml/ocaml). The patches needed for ARM64 support were pushed in December (see https://github.com/ocaml/ocaml/pull/11821).

- The first commit upgrades #18245 to reference the patches directly from the repository, as for #19855.
- The second commit refines the availability added in #17716 for +fp variants. These are not available on _any_ arm64 (or non-Linux) build.
- The third commit then fixes ARM64 support for a whole series of ocaml-variants packages for 4.02-4.05. I have tested the compilation of all of these both on arm64 and amd64 - a few of the musl ones still technically need more work on arm64 (if someone were ever to care), but in principle they would require the patch for ocaml/ocaml#1330, so it's added.
- The fourth commit adds the GCC 10 fix from #16722 to 3 further compiler forks (which allows them to be compiled again).
- The fifth and sixth commits fix the installation of `ocaml-variants.4.04.1+copatterns` - this has a `+dev` version number and so compiles with `-Werror` by default (cf. #18225) and the package was mis-named with a `4.04.0` version number. This didn't matter in OPAM 1.x when it was first merged, but it does matter in opam 2.x.
- Finally, the location of the tarball for `ocaml-variants.4.02.1+BER` is corrected.

This PR is slightly easier to review commit-by-commit, as similar changes are then grouped together.

[Why does the GCC 10 fix remains as a patch file? Glad you asked! There are commits for these in the maintenance branches as well (cf. ocaml/ocaml#11811), but that patch fixes the actual problem in the source code. The patch in opam-repository is less invasive, since it just changes the gcc flags to restore the pre-GCC 10 behaviour. The ARM64 and glibc patches cannot be fixed just by tweaking flags, hence the code changes in the maintenance branches are adopted instead.]


kit-ty-kate:
Thanks

PR #22838, opened 2023-01-04
coq-of-ocaml: update the urls of the repo

clarus:
Update the urls to the GitHub repository of `coq-of-ocaml`. Hopefully this does not break anything as GitHub was already doing redirects.


kit-ty-kate:
Thanks!

PR #22839, opened 2023-01-04
coq-of-ocaml: add version 2.5.3

clarus:
There are versions for OCaml 4.12, 4.13, and 4.14. Thanks!

### DIFF
--- a/packages/coq-of-ocaml/coq-of-ocaml.2.5.3+4.12/opam
+++ b/packages/coq-of-ocaml/coq-of-ocaml.2.5.3+4.12/opam
@@ -14,7 +14,7 @@ install: [
   [make "-C" "proofs" "install"] {coq:installed}
 ]
 depends: [
-  "angstrom"
+  "angstrom" {>= "0.15.0"}
   "csexp"
   "dune" {>= "2.8"}
   "ocaml" {>= "4.12" & < "4.13"}
--- a/packages/coq-of-ocaml/coq-of-ocaml.2.5.3+4.12/opam
+++ b/packages/coq-of-ocaml/coq-of-ocaml.2.5.3+4.12/opam
@@ -1,0 +1,45 @@
+opam-version: "2.0"
+maintainer: "dev@clarus.me"
+homepage: "https://github.com/formal-land/coq-of-ocaml"
+dev-repo: "git+https://github.com/formal-land/coq-of-ocaml.git"
+bug-reports: "https://github.com/formal-land/coq-of-ocaml/issues"
+authors: ["Guillaume Claret"]
+license: "MIT"
+build: [
+  ["sh" "-c" "cd proofs && ./configure.sh"] {coq:installed}
+  [make "-C" "proofs" "-j%{jobs}%"] {coq:installed}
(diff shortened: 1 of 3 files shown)


packages/coq-of-ocaml/coq-of-ocaml.2.5.3+4.12/opam
```diff
@@ -0,0 +1,45 @@
...
+build: [
+  ["sh" "-c" "cd proofs && ./configure.sh"] {coq:installed}
+  [make "-C" "proofs" "-j%{jobs}%"] {coq:installed}
+  ["dune" "build" "-p" name "-j" jobs]
+]
+install: [
+  [make "-C" "proofs" "install"] {coq:installed}
+]
+depends: [
+  "angstrom"
```
raphael-proust, line 17: You are missing a lower-bound for this constraint. This is causing the CI error:

```
# File "src/constant.ml", line 55, characters 13-25:
# 55 | let dquote = Angstrom.map ~f:(fun _ -> PDQuote) (Angstrom.char '"')
#                   ^^^^^^^^^^^^
# Error: Unbound value Angstrom.map
```

I think the `map` function was introduced in 0.15.0: https://github.com/inhabitedtype/angstrom/commit/93f29c2343491ef773efc89d2fa4e695a5bd1202

So I'd suggest

```suggestion
  "angstrom" {>= "0.15.0"}
```

If this is correct, then you should also upstream this change.

raphael-proust: Note: the same change is needed on all three package files.

clarus: Thanks! I have added the constraint for all three packages.

packages/coq-of-ocaml/coq-of-ocaml.2.5.3+4.14/opam
```diff
@@ -0,0 +1,45 @@
...
+  ["sh" "-c" "cd proofs && ./configure.sh"] {coq:installed}
+  [make "-C" "proofs" "-j%{jobs}%"] {coq:installed}
+  ["dune" "build" "-p" name "-j" jobs]
+]
+install: [
+  [make "-C" "proofs" "install"] {coq:installed}
+]
+depends: [
+  "angstrom" {>= "0.15.0"}
+  "csexp"
```
raphael-proust, line 18: There seems to be a lower-bound constraint missing here:
[lower-bounds (failed: Unbound type constructor Csexp.t)](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/5114feaf36e3735c6a5def87485d58a854d4077e/variant/compilers,4.14,coq-of-ocaml.2.5.3+4.14,lower-bounds)

I'm not sure what exact version you need, but the downgrade which reveals the error is:
```
downgrade csexp              1.5.1 to 1.0.0      [required by coq-of-ocaml]
```

clarus: OK, thanks! I added a lower bound constraint.

kit-ty-kate:
Thanks

PR #22840, opened 2023-01-05
new file:   packages/molenc/molenc.16.15.0/opam

UnixJunkie:
molenc for ocaml-5.
Forced to switch to lower performance dokeysto instead of dokeysto_camltc, because camltc is not compatible with ocaml-5 yet (complained upstream).


kit-ty-kate:
Thanks

PR #22848, opened 2023-01-05
[new release] jose (0.8.0)

ulrikstrid:
CHANGES:

- Make `use` and `alg` optional
- Correct thumbprint generation on all algs
- Add getters for claims
- Thumbprint is now a Cstruct.t instead of string which is less ambigious
- Make `header` argument optional when signing which simplifies the normal usecase

https://github.com/ulrikstrid/reason-jose/issues/54


ulrikstrid:
Is the failure that I need `yojson < 2.0.0`?

ulrikstrid:
Closing in favour of #22849

PR #22836, opened 2023-01-04
Add ocamlfind-secondary.1.9.6

dra27:
Fixes #22817

PR #22852, opened 2023-01-05
Fix upper bounds on ocaml 5 (first 1/3)

mseri:
Seen on https://github.com/ocaml/opam-repository/pull/22836

This is about a third of the packages that need fixing from the list of obvious ocaml 5 failures in https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/0fa072f75cea976a57094a6a545875b4c4c77446

### DIFF
--- a/packages/gsl/gsl.1.21.0/opam
+++ b/packages/gsl/gsl.1.21.0/opam
@@ -16,7 +16,7 @@ build: [
 ]
 
 depends: [
-  "ocaml" {>= "4.04"&| < "5.0"}
+  "ocaml" {>= "4.04" & < "5.0"}
   "conf-gsl" {build}
   "conf-pkg-config" {build}
   "base" {build & < "v0.13"}
--- a/packages/gsl/gsl.1.21.0/opam
+++ b/packages/gsl/gsl.1.21.0/opam
@@ -16,7 +16,7 @@ build: [
 ]
 
 depends: [
-  "ocaml" {>= "4.04"}
+  "ocaml" {>= "4.04"&| < "5.0"}
   "conf-gsl" {build}
   "conf-pkg-config" {build}
   "base" {build & < "v0.13"}
--- a/packages/jingoo/jingoo.1.2.19/opam
+++ b/packages/jingoo/jingoo.1.2.19/opam
@@ -12,7 +12,7 @@ build: [
 install: [make "install"]
 remove: ["ocamlfind" "remove" "jingoo"]
 depends: [
-  "ocaml" {>= "4.02.3" & <"5.0"}
+  "ocaml" {>= "4.02.3"}
   "ocamlfind" {build}
   "pcre"
   "uutf"
--- a/packages/jingoo/jingoo.1.2.19/opam
+++ b/packages/jingoo/jingoo.1.2.19/opam
@@ -12,7 +12,7 @@ build: [
 install: [make "install"]
 remove: ["ocamlfind" "remove" "jingoo"]
 depends: [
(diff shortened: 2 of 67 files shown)


kit-ty-kate:
We should avoid doing this kind of massive PRs, they're harder to review and more annoying to update it.
Have you tried using my `git_pr` script instead?

mseri:
No, which `git_pr` script?

kit-ty-kate:
This one: https://github.com/kit-ty-kate/dotfiles/blob/master/.local/bin/git_pr (it's tailored for my username and remote name but if you modify them use yours it'll work just fine)

mseri:
Thanks! How it would have helped? I thought that making a single PR instead of 37 separate ones would be better. In the end, this is just about upper bound on ocaml 5.0 (errors in the commit messages) and not more complex bounds or changes.

packages/gsl/gsl.1.21.0/opam
```diff
@@ -16,7 +16,7 @@ build: [
 ]
 
 depends: [
-  "ocaml" {>= "4.04"}
+  "ocaml" {>= "4.04"&| < "5.0"}
```
mseri, line 19:
```suggestion
  "ocaml" {>= "4.04" & < "5.0"}
```

kit-ty-kate:
> Thanks! How it would have helped? I thought that making a single PR instead of 37 separate ones would be better. In the end, this is just about upper bound on ocaml 5.0 on quite old packages (errors in the commit messages) and not more complex bounds or changes.

If this kind of huge PR was done before a release maybe it would be better but since it’s done after we have to be extra careful about doing changes in existing packages.
And even in this case, try to double-check your own work (going commit by commit, looking at the failing logs in the commit message, looking at the changes to check which packages have been modified) and you will see it’s a lot more work than checking 37 PRs with one semantically equivalent change per PR.

Furthermore if you need to make a change (e.g. the syntax error), the CI doesn’t have to run multiple times with half a million runs for nothing, only for the actual PR that has an issue.

> What I am trying to understand is why the 37 separate PRs are less intensive on the CI.

Mostly for the reasons stated above, but as for CI reason, it’s much less intensive to render several small packages than one very large one, on top of being harder to read the result of.

packages/jingoo/jingoo.1.2.19/opam
```diff
@@ -12,7 +12,7 @@ build: [
 install: [make "install"]
 remove: ["ocamlfind" "remove" "jingoo"]
 depends: [
-  "ocaml" {>= "4.02.3"}
+  "ocaml" {>= "4.02.3" & <"5.0"}
```
kit-ty-kate, line 15: jingoo 1.2.19 works fine on 5.0: http://check.ocamllabs.io/log/1672677267-c44c6653bf257a9e3493815b1a509330ada60e1c/5.0/good/jingoo.1.2.19

kit-ty-kate:
Also makeing multiple PRs is easier to detect if a maintainer of a package whose latest version is broken, needs pinging.
After reviewing this, I’m too exausted to go back and look them up one by one.

packages/jingoo/jingoo.1.2.19/opam
kit-ty-kate:
```suggestion
  "ocaml" {>= "4.02.3"}
```

kit-ty-kate:
Thanks!

mseri:
Makes perfect sense, thanks. Making small PRs is not a problem in any case, I will do that from the next time!

thanks for the review

PR #22869, opened 2023-01-06
Restore ocaml-variants.5.0.0+trunk and ocaml-variants.4.14.1+trunk

dra27:
The version numbering requirements of the `ocaml` package mean that the `+trunk` development variant becomes obsolete as soon as a release is made - i.e. once 5.0.0 and 4.14.1 were released, the `+trunk` packages need to be `ocaml-variants.5.0.1+trunk` and `ocaml-variants.4.14.2+trunk`. This was done in #22724, but the renaming is causing problems on some platforms because of the way `opam update` works internally. There's work in progress on the opam side, with linked issues at https://github.com/ocaml/opam/pull/5400. There's also failures like https://github.com/ocaml/opam/issues/5408. On top of that, fairly soon, the aim is to be able to get "trunk" OCaml via a pin, rather than a specific package. So in future this problem hopefully disappears.

In the meantime, I think it may help just to restore dummy versions of the old packages (and potentially do this for any new OCaml releases)?

cc @Octachron (and, tautologically, @kit-ty-kate!)

### DIFF
--- a/packages/ocaml-variants/ocaml-variants.4.14.1+trunk/opam
+++ b/packages/ocaml-variants/ocaml-variants.4.14.1+trunk/opam
@@ -1,0 +1,11 @@
+opam-version: "2.0"
+license: "LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception"
+synopsis: "4.14.1 obsolete development package"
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
Fine by me. Thanks!

PR #22901, opened 2023-01-10
Add dkml-base-compiler to 4.14 & 5.x trunk packages

dra27:
Apropos https://github.com/ocaml/opam-repository/pull/22719#discussion_r1062400286, this adds `dkml-base-compiler` to the "trunk" `ocaml` packages for 4.14 and 5.x (i.e. the unreleased versions: 4.14.2, 5.0.1 and 5.1.0).

This PR is low-impact as it only triggers for development builds (and the compiler itself is not recompiled as a result of changes in the `ocaml` package).

PR #22902, opened 2023-01-10
Add dkml-base-compiler to ocaml.4.14.1 & ocaml.5.0.0

dra27:
Also apropos https://github.com/ocaml/opam-repository/pull/22719#discussion_r1062400286, this adds `dkml-base-compiler` to the latest releases in 4.14 and 5.0.

Unlike #22901, this is higher impact, as it means `opam update` for users will trigger a rebuild of all installed packages (apart from the compiler itself).

### DIFF
--- a/packages/ocaml/ocaml.4.14.1/opam
+++ b/packages/ocaml/ocaml.4.14.1/opam
@@ -9,7 +9,8 @@ depends: [
   "ocaml-config" {>= "2"}
   "ocaml-base-compiler" {>= "4.14.1~" & < "4.14.2~"} |
   "ocaml-variants" {>= "4.14.1~" & < "4.14.2~"} |
-  "ocaml-system" {>= "4.14.1" & < "4.14.2~"}
+  "ocaml-system" {>= "4.14.1" & < "4.14.2~"} |
+  "dkml-base-compiler" {>= "4.14.1~" & < "4.14.2~"}
 ]
 setenv: [
   [CAML_LD_LIBRARY_PATH = "%{_:stubsdir}%"]
--- a/packages/ocaml/ocaml.5.0.0/opam
+++ b/packages/ocaml/ocaml.5.0.0/opam
@@ -9,7 +9,8 @@ depends: [
   "ocaml-config" {>= "3"}
   "ocaml-base-compiler" {>= "5.0.0~" & < "5.0.1~" } |
   "ocaml-variants" {>= "5.0.0~" & < "5.0.1~"} |
-  "ocaml-system" {>= "5.0.0" & < "5.0.1~"}
+  "ocaml-system" {>= "5.0.0" & < "5.0.1~"} |
+  "dkml-base-compiler" {>= "5.0.0~" & < "5.0.1~"}
 ]
 setenv: [
   [CAML_LD_LIBRARY_PATH = "%{_:stubsdir}%"]


kit-ty-kate:
I’m not so sure about this one compared to https://github.com/ocaml/opam-repository/pull/22901. I’d prefer not to do it if possible

jonahbeckford:
DKML can skip point releases, but and so …

5.0.0 is much less necessary. 4.14.1 is pretty important; the recent release of OCaml LSP, for example, disables Dune watch mode tracking on Windows for 4.14.0. (Yes, DKML supports Dune watch mode today).
When is 4.14.2? If “soon” we can skip 4.14.1.

dra27:
I imagine there may be 4.14.2 somewhere around the time of 5.1.0 - so 6(-9) months' time. I think the only question is whether these packages get added _before_ dkml-base-compiler exists at the same version or _when_ they exist (i.e. if `dkml-base-compiler.4.14.1` exists then `ocaml.4.14.1` ought to support it and likewise for `ocaml.5.0.0`).

I emphasise _think_, as I presume another possibility is to carry a patched `ocaml.4.14.1` package in `diskuv-opam-repository` (is that right, @jonahbeckford?)

jonahbeckford:
I’m sure you don’t want that stuck in the DKML repository. DKML operates on rolling releases, including the setup-dkml CI which has some adoption today. Ripple effects of never releasing 4.14.1 to the main repository are hard to enumerate: winget install opam would not be reproducible from the main repository; not sure how just-released opam-monorepo will behave now that they added some support for dkml-base-compiler (untested by me so far, but today it’s the only simple option to get Arm64 macOS builds on GitHub/GitLab x64 hardware); etc.

It is the unknown unknowns (hopefully that term is understood outside of America) that worries me.

I would be less worried if 4.14.2 was soon (weeks) because LSP forced my hand … I have to bump to 4.14.1 soon or Windows users will get crippled on my next release. Perhaps we can redirect LSP to have a better solution than https://github.com/ocaml/ocaml-lsp/pull/935? (Which did not realize that dune build —watch works today for DKML users, and that recently there has been a large increase in Windows users).

dra27:
Hmm, in which case I wonder if the thing to do is leave `ocaml.5.0.0` completely alone and update `ocaml.4.14.1` when there's `dkml-base-compiler.4.14.1` to go with it? There isn't an easy solution for ocaml-lsp - the compiler either needs that patch or they'd need a custom C stub to avoid using `Unix.stat` (neither of which is particularly nice!)

jonahbeckford:
There was a third solution; do nothing. The problem went from high productivity with occasional crashes (forcing the user occasionally to restart `dune build --watch` or LSP) to no crashes but low productivity (forcing the user often to restart OCaml LSP).

*Solution?*

For now I can take that do-nothing solution. No upgrades to OCaml LSP until OCaml 4.14.2 is released. And DKML will skip upstreaming 4.14.1 to ocaml/opam-repository since there is no longer a pressing need for it.

Also, ignoring LSP/Dune for the moment, backporting https://github.com/ocaml/ocaml/issues/11737 into `dkml-base-compiler.4.14.0` would be good regardless. So I'll do that backport before upstreaming 4.14.0.

kit-ty-kate:
@jonahbeckford what do you think we should do? Do you have plans to use 4.14.1 ?
@Octachron is there a timeframe for the release of 4.14.2 ?

jonahbeckford:
Don't really have a choice. On Windows we'll skip 4.14.1, so we can close this ticket.

Octachron:
I am aiming for a 4.14.2 release  in the beginning of November.

kit-ty-kate:
Awesome, thank you both!

kit-ty-kate:
oops.... muscle memory made me click on the green button instead of the close button.. Emergency reverted in https://github.com/ocaml/opam-repository/pull/24488

PR #22905, opened 2023-01-10
[new release] raylib (0.6.0)

tjammer:
CHANGES:

* Support OCaml 5
* Upgrade raylib to 4.2.0


tjammer:
The breakage for old versions of raygui with the new raylib version is to be expected, unfortunately

kit-ty-kate:
> The breakage for old versions of raygui with the new raylib version is to be expected, unfortunately

Fixed in #22923

kit-ty-kate:
Thanks

PR #22906, opened 2023-01-11
new file:   packages/acpc/acpc.2.0.0/opam

UnixJunkie:
compiles with ocaml-5 and dune (instead of obuild previously)

### DIFF
--- a/packages/acpc/acpc.2.0.0/opam
+++ b/packages/acpc/acpc.2.0.0/opam
@@ -1,0 +1,38 @@
+opam-version: "2.0"
+name: "acpc"
+authors: "Francois Berenger"
+maintainer: "unixjunkie@sdf.org"
+homepage: "https://github.com/UnixJunkie/ACPC"
+bug-reports: "https://github.com/UnixJunkie/ACPC/issues"
+license: "BSD-3-Clause"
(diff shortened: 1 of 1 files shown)


packages/acpc/acpc.2.0.0/opam
```diff
@@ -0,0 +1,38 @@
...
+depends: [
+  "base-unix"
+  "batteries"
+  "bitv"
+  "conf-autoconf"
+  "conf-gnuplot"
+  "dolog"
+  "dune" {>= "1.11"}
+  "ocaml" {>= "5.0.0"}
+  "parany"
```
kit-ty-kate, line 19: Does not expect `Parmap.parmap` to have a `~ncores` argument
```suggestion
  "parany" {>= "11"}
```

UnixJunkie:
Can't the solver find a proper parany version?
There are only few which are OK with ocaml>=5.0.0.

packages/acpc/acpc.2.0.0/opam
```diff
@@ -0,0 +1,38 @@
+opam-version: "2.0"
+name: "acpc"
```
mseri, line 2:
```suggestion
```

mseri:
The remaining failures are unrelated

PR #22946, opened 2023-01-12
Package petrol.1.0.0

Gopiandcode:
### `petrol.1.0.0`
Petrol's an OCaml SQL API made to go FAST
Petrol is an typed embedding of SQL into OCaml integrated with Caqti and support for versioning.

---
* Homepage: https://github.com/gopiandcode/petrol
* Bug tracker: https://github.com/gopiandcode/petrol/issues

---


kit-ty-kate:
Thanks. Could you return the fix to the opam file upstream so that it can be taken into account for next release?

Gopiandcode:
Noted, will do! Thanks for fixing the deps!

PR #22947, opened 2023-01-13
new file:   packages/parany/parany.13.1.2/opam

UnixJunkie:
even safer Parany.get_rank().
There is also now a Parany.DLS module.


mseri:
Thanks

PR #22961, opened 2023-01-14
owi.0.1

zapashcanon:
Hi,

This is the first release of [owi](https://github.com/ocamlpro/owi).

Cheers,


zapashcanon:
The openSUSE failure seems unrelated.

The i32 failure caught at least two bugs in my code, I fixed one but the other will take more time. The bug is really happening in a corner case that is unlikely to happen to most people. Could it be possible to disable tests only for the i32 arch ? If not, I'm fine with disabling the package for i32.

kit-ty-kate:
Thanks

zapashcanon:
Thanks !

PR #22963, opened 2023-01-15
update pa_ppx_{hashcons,q_ast,unique} for ocaml 5.0.0

chetmurthy:
minor changes to make these work with ocaml 5.0.0


chetmurthy:
Looks like the test-failures are all the same as always: macos-homebrew (b/c of missing conf-perl-* packages)

chetmurthy:
Hi, is there anything else I need to do here?  This fails the macos-homebrew tests, but everything else passes, and the failures are for the same reasons as in the past.

kit-ty-kate:
Thanks, sorry for the wait

chetmurthy:
Gracias!

PR #16743, opened 2020-07-03
Various fixes

kit-ty-kate:
cc @pirbo for the tezos packages

PR #22983, opened 2023-01-17
ffmpeg: disable CI on Debian 11

emillon:
This is a followup of #16743 - the version that was unstable at that
time is now debian 11.


emillon:
We could go further:
- restrict it to more distros
- add patches to change the header file to read (`avcodec.h` -> `codec_id.h` in `gen_code.ml`)
But just removing debian 11 which is used for pretty much all CI runs seems like it's an effortless tradeoff.

mseri:
Sounds good as is, the ffmpeg package is well mantained and has new periodic releases anyway

kit-ty-kate:
Thanks!

PR #22985, opened 2023-01-17
aacplus: do not test in Debian 11

emillon:
Follow-up to #14800. These changes have propagated to Debian 11.


kit-ty-kate:
Thanks

PR #22986, opened 2023-01-17
ahrocksdb: ignore CI failures

emillon:
The error message is:

> installed RocksDB installation is too old: found 6.11, expected 5.14 minimum

Which is actually fine! There's a logic error which has been fixed upstream in ahrefs/ocaml-ahrocksdb@d74dc6f8337557c23b66c2a2488957d7ce3c278c.

This PR ignores CI errors caused by this.

### DIFF
--- a/packages/ahrocksdb/ahrocksdb.0.2.0/opam
+++ b/packages/ahrocksdb/ahrocksdb.0.2.0/opam
@@ -39,7 +39,7 @@ depends: [
 build: [
   ["dune" "subst"] {dev}
   ["dune" "build" "-p" name "-j" jobs]
-  ["dune" "runtest"] {with-test}
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
 ]
 x-ci-accept-failures: ["debian-unstable"]
 dev-repo: "git+https://github.com/ahrefs/ocaml-ahrocksdb.git"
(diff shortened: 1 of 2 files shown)


packages/ahrocksdb/ahrocksdb.0.2.0/opam
```diff
@@ -42,6 +42,13 @@ build: [
   ["dune" "runtest"] {with-test}
```
mseri, line 42:
```suggestion
  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
```

packages/ahrocksdb/ahrocksdb.0.2.1/opam
```diff
@@ -40,6 +40,13 @@ build: [
   ["dune" "runtest"] {with-test}
```
mseri, line 40:
```suggestion
  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
```

packages/ahrocksdb/ahrocksdb.0.2.0/opam
```diff
@@ -39,9 +39,16 @@ depends: [
 build: [
   ["dune" "subst"] {dev}
   ["dune" "build" "-p" name "-j" jobs]
-  ["dune" "runtest"] {with-test}
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
 ]
 x-ci-accept-failures: ["debian-unstable"]
```
kit-ty-kate, line 44: I'm not sure it's worth patching packages especially if the latest version fixed the issue already. I'd recommend just ignoring the failure instead
```suggestion
x-ci-accept-failures: ["debian-11" "debian-unstable"]
```

emillon:
@kit-ty-kate I switched this to just ignore CI errors.

mseri:
Since the fix was already there I don’t see why not keeping it. Anyway I am fine to merge it as is or with the fix

kit-ty-kate:
Thanks. Sorry for the wait

PR #22989, opened 2023-01-17
Release of esperanto v0.0.2

dinosaure:
- Delete `conf-gcc` (we currently are able to use `clang`)
- The current toolchain does not work on MacOS
- Add conf-cosmopolitan package
- Upgrade to Cosmopolitan v2.2 (#28)
- Update the provided `extract.com` (#30)
- Fix the support of OCaml 4.14.1
- Delete `pth` (Cosmopolitan provides `pthread`)

### DIFF
--- a/packages/esperanto-cosmopolitan/esperanto-cosmopolitan.0.0.2/opam
+++ b/packages/esperanto-cosmopolitan/esperanto-cosmopolitan.0.0.2/opam
@@ -15,7 +15,7 @@ build: [
 # depends: [
 #   "conf-binutils"
 # ]
-install: [ "sh" "-c" "cd toolchain ; %{make}% install" ]
+install: [ "make" "-C" "toolchain" "install" ]
 synopsis: "Cosmopolitan toolchain for OCaml compiler"
 description: "A little toolchain for OCaml with Cosmopolitan"
 available: [ arch = "x86_64" & (os = "linux" | os = "freebsd" | os = "openbsd") ]
--- a/packages/esperanto-cosmopolitan/esperanto-cosmopolitan.0.0.2/opam
+++ b/packages/esperanto-cosmopolitan/esperanto-cosmopolitan.0.0.2/opam
@@ -15,7 +15,7 @@ build: [
 # depends: [
 #   "conf-binutils"
 # ]
-install: [ "make" "-C" "toolchain" "install" ]
+install: [ make "-C" "toolchain" "install" ]
 synopsis: "Cosmopolitan toolchain for OCaml compiler"
 description: "A little toolchain for OCaml with Cosmopolitan"
 available: [ arch = "x86_64" & (os = "linux" | os = "freebsd" | os = "openbsd") ]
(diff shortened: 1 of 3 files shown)


packages/esperanto/esperanto.0.0.2/opam
```diff
@@ -0,0 +1,27 @@
...
+homepage: "https://github.com/dinosaure/gilbraltar"
+bug-reports: "https://github.com/dinosaure/gilbraltar/issues/"
+license: "MIT"
+tags: "org:mirage"
+dev-repo: "git+https://github.com/dinosaure/gilbraltar.git"
+build: [
+  ["sh" "-c" "cd caml && ./configure.sh --prefix=%{prefix}% --target=x86_64-esperanto-none-static"]
+  [make "-C" "caml" "-j%{jobs}%"]
+]
+install: ["sh" "-c" "cd caml ; %{make}% install" ]
```
kit-ty-kate, line 13:
```suggestion
install: [make "-C" "caml" "install" ]
```

packages/esperanto-cosmopolitan/esperanto-cosmopolitan.0.0.2/opam
```diff
@@ -0,0 +1,25 @@
...
+build: [
+  [ "sh" "-c" "cd toolchain && ./configure.sh --prefix=%{prefix}%" ]
+  [ make "-C" "toolchain" "all" "V=1" ]
+]
+# We should depend on conf-binutils but the package does not work on FreeBSD
+# even if we can do [pkg install binutils] on FreeBSD 12
+# depends: [
+#   "conf-binutils"
+# ]
+install: [ "sh" "-c" "cd toolchain ; %{make}% install" ]
```
kit-ty-kate, line 18:
```suggestion
install: [ make "-C" "toolchain" "install" ]
```

kit-ty-kate:
> Fix the support of OCaml 4.14.1

I was kinda surprised by this changelog entry so I looked into it, and I'm really not keen on merging such a brittle package in its current state if it's going to break for each version of the compiler

https://github.com/dinosaure/esperanto/blob/03e6822204f7961bbeb8723b93fefc37d989f11a/caml/GNUmakefile#L106

Is there any chance you could replace this with something of the sort?
```
	@if [ "$(MAKECONF_OCAML_VERSION_MINOR)" -ge 14 ] ; then \
```
or/and a more robust way to guard against unwanted change than `sed -e "56,58 s/...`. For example by grepping for the entire multi-line string instead of just using the line number

dinosaure:
> Is there any chance you could replace this with something of the sort?
> 
>     @if [ "$(MAKECONF_OCAML_VERSION_MINOR)" -ge 14 ] ; then \

I integrated your requested change into my last tarball :+1:.

dinosaure:
Thanks!

kit-ty-kate:
Thanks. Note that `"$(OCAML_VERSION_MAJOR)" -ge "4" -a "$(OCAML_VERSION_MINOR)" -ge "14"` is not equivalent to `OCaml >= 4.14`. For this particular package it's fine given it only works on 4.x but this won't be correct once you add 5.x support. I would also still encourage you to try a more robust way compared to using the line numbers.

Could you also bring this fix to your `main` branch? It seems to be only present on your `prepare-v0.0.2` branch

dinosaure:
Yes, I will do that as soon as I can :+1:.

PR #22993, opened 2023-01-18
conf-srt-openssl: allow failures on Debian 11

emillon:
This package is affected by <https://bugs.debian.org/1001529> in Debian 11: the wrong `.pc` file is installed, so `pkg-config --exists srt` exits with an error (since `gnutls` is not installed).


emillon:
(not sure what to do about non-debian OSes. this conf- package probably only makes sense under debian, so maybe it should be `available:` only on debian. but this can be fixed separately if needed)

mseri:
Let’s eave it as is, we don’t need to fix this types of issues

PR #22995, opened 2023-01-18
conf-trexio: ignore CI failures on Debian 11

emillon:
This package is not packaged in Debian 11, so all CI builds will fail.


kit-ty-kate:
Thanks

PR #22996, opened 2023-01-18
Octez 16.0~rc1 packages

killian-delarue:
Packages for the first release candidate for Octez v16.0


packages/tezos-stdlib-unix/tezos-stdlib-unix.16.0~rc1/opam
```diff
@@ -0,0 +1,38 @@
...
+  "tezos-lwt-result-stdlib" { = version }
+  "tezos-event-logging" { = version }
+  "tezos-stdlib" { = version }
+  "data-encoding" { >= "0.7.1" & < "1.0.0" }
+  "lwt" { >= "5.6.0" }
+  "ipaddr" { >= "5.0.0" & < "6.0.0" }
+  "re" { >= "1.9.0" }
+  "ezjsonm" { >= "1.1.0" }
+  "ptime" { >= "1.0.0" }
+  "mtime" { >= "1.4.0" }
```
raphael-proust, line 21:
```suggestion
  "mtime" { >= "1.4.0" & < "2.0.0" }
```

The conversion functions were removed in 2.0.0: https://github.com/dbuenzli/mtime/blob/master/CHANGES.md

killian-delarue: Applied in 06703111c68508673b5208e659144a57ad1eb046

raphael-proust:
Amongst the errors of the CI are a few `Error: No rule found for <some .exe>`. I've checked the dune files and the `(rule (alias runtest) …)` stanzas that trigger this error have corresponding `(executables …)` stanzas. I'm not sure what could be causing this; it seems to happen on ocaml5 on more unusual archs.

emillon:
> I'm not sure what could be causing this; it seems to happen on ocaml5 on more unusual archs.

These archs do not have `ocamlopt` and the dune file is asking for a `.exe` (native executable), so it fails with this error message. A related error message is that `zarith.cmxa` is missing.

(please let this build go through (it's almost over) before pushing fixes as I think it'll allow the cluster to pick other PRs in the meantime)

raphael-proust:
Thanks for the explanations. I can't find a way to mark a package as available only when ocamlopt is available. I'm looking at https://opam.ocaml.org/doc/Packaging.html and I can't see an appropriate variable.

Am I missing something?  
Should we just ignore these issues?

kit-ty-kate:
@raphael-proust adding a conflict to the `ocaml-option-bytecode-only` package should be enough:
```
conflicts: [
  "ocaml-option-bytecode-only"
]
```

raphael-proust:
I've opened an issue upstream regarding the assembler errors on arm32 and ppc64: https://github.com/ocaml/ocaml/issues/11948

The other errors are all acceptable to the NL release team (runner errors, timeout errors in some tests, etc.) on some architectures we do not officially support.

PR #22998, opened 2023-01-18
bls12-381: disable tests on ocaml 5

emillon:
The random number generator has changed in OCaml 5.0.0 and the test suite has different expected values depending on the version.

There is a patch upstream that has been applied in later versions.

### DIFF
--- a/packages/bls12-381-unix/bls12-381-unix.1.1.0/opam
+++ b/packages/bls12-381-unix/bls12-381-unix.1.1.0/opam
@@ -25,7 +25,7 @@ x-ci-accept-failures: [
   "oraclelinux-7" # default C compiler too old?
 ]
 build: ["dune" "build" "-j" jobs "-p" name "@install"]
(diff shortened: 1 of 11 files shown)


kit-ty-kate:
I'm not sure it's worth patching packages compared to just disabling the tests on ocaml5

emillon:
Ah I thought of depending on `"ocaml" {< "5.0.0" & with-test}` but that seemed to restrictive. What's the recommended technique to skip tests on >=5 ?

kit-ty-kate:
Either that or:
```
run-test: ["dune" "runtest" "-p" name "-j" jobs] {ocaml:version < "5.0"}
```

emillon:
Something like this?

    run-test: ["dune" "runtest" "-p" name "-j" jobs] { ocaml:version < "5.0.0" }

emillon:
Let's go with that

emillon:
Done. This looks better indeed.

mseri:
thanks

PR #23013, opened 2023-01-19
ocaml 5: patch.1.0.0 uses bytes

emillon:
Depend on base-bytes to avoid this failure:

    #=== ERROR while compiling patch.1.0.0 ========================================#
    # context              2.2.0~alpha~dev | linux/x86_64 | ocaml-base-compiler.5.0.0 | file:///home/opam/opam-repository
    # path                 ~/.opam/5.0/.opam-switch/build/patch.1.0.0
    # command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p patch -j 47
    # exit-code            1
    # env-file             ~/.opam/log/patch-10-53f31d.env
    # output-file          ~/.opam/log/patch-10-53f31d.out
    ### output ###
    # File "src/dune", line 7, characters 12-17:
    # 7 |  (libraries bytes))
    #                 ^^^^^

### DIFF
--- a/packages/patch/patch.1.0.0/opam
+++ b/packages/patch/patch.1.0.0/opam
@@ -9,7 +9,7 @@ license: "ISC"
 
 depends: [
   ("ocaml" {>= "4.04.2" & < "5.0"} | "ocaml" {>= "5.0"} & "base-bytes")
-  "dune" {>= "1.4.0"}
+  ("dune" {>= "1.4.0"} | "dune" {< "1.4.0"} & "base-bytes")
   "alcotest" {with-test}
   "crowbar" {with-test}
 ]
--- a/packages/patch/patch.1.0.0/opam
+++ b/packages/patch/patch.1.0.0/opam
@@ -12,7 +12,7 @@ depends: [
   "dune"
   "alcotest" {with-test}
   "crowbar" {with-test}
-  "base-bytes"
+  "base-bytes" {ocaml:version >= "5.0"}
 ]
 build: [
   ["dune" "subst"] {dev}
--- a/packages/patch/patch.1.0.0/opam
+++ b/packages/patch/patch.1.0.0/opam
@@ -8,11 +8,10 @@ bug-reports: "https://github.com/hannesm/patch/issues"
 license: "ISC"
 
 depends: [
-  "ocaml" {>= "4.04.2"}
+  ("ocaml" {>= "4.04.2" & < "5.0"} | "ocaml" {>= "5.0"} & "base-bytes")
   "dune"
   "alcotest" {with-test}
   "crowbar" {with-test}
-  ("ocaml" {>= "4.04.2" & < "5.0"} | "ocaml" {>= "5.0"} & "base-bytes")
 ]
 build: [
   ["dune" "subst"] {dev}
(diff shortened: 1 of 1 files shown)


packages/patch/patch.1.0.0/opam
```diff
@@ -12,6 +12,7 @@ depends: [
   "dune"
   "alcotest" {with-test}
   "crowbar" {with-test}
+  "base-bytes"
```
kit-ty-kate, line 15: The issue with this is that the package now suddenly depends on ocamlfind which might not be the intent of the developers. `patch.1.0.1` works fine with ocaml5, why not simply add the `< "5.0"` constraint?

emillon: I asked what was the right fix for that and @dra27 told me to use `base-bytes`. But as you prefer, I don't think it matters in that case.

kit-ty-kate: The current consensus is to go half way:
```suggestion
  "base-bytes" {ocaml:version >= "5.0"}
```

packages/patch/patch.1.0.0/opam
```diff
@@ -12,6 +12,7 @@ depends: [
   "dune"
   "alcotest" {with-test}
   "crowbar" {with-test}
+  "base-bytes" {ocaml:version >= "5.0"}
```
dra27, line 15: Package variables can't be used here (technically for `:version`, opam _could_ rewrite it to a package formula, but it doesn't). I can't edit L11, but this wants to be:
```suggestion
  ("ocaml" {>= "4.04.2" & < "5.0"} | "ocaml" {>= "5.0"} & "base-bytes")
```
without L11 above.

packages/patch/patch.1.0.0/opam
```diff
@@ -8,7 +8,7 @@ bug-reports: "https://github.com/hannesm/patch/issues"
 license: "ISC"
 
 depends: [
-  "ocaml" {>= "4.04.2"}
+  ("ocaml" {>= "4.04.2" & < "5.0"} | "ocaml" {>= "5.0"} & "base-bytes")
   "dune"
```
dra27, line 12: Should fix the lowerbounds jobs: 1.4.0 includes https://github.com/ocaml/dune/pull/1391
```suggestion
  "dune" {>= "1.4.0"}
```
(the addition of the lowerbound is independent of the ocaml 5 support, obviously)

packages/patch/patch.1.0.0/opam
```diff
@@ -8,8 +8,8 @@ bug-reports: "https://github.com/hannesm/patch/issues"
 license: "ISC"
 
 depends: [
-  "ocaml" {>= "4.04.2"}
-  "dune"
+  ("ocaml" {>= "4.04.2" & < "5.0"} | "ocaml" {>= "5.0"} & "base-bytes")
+  "dune" {>= "1.4.0"}
```
kit-ty-kate, line 12: Otherwise we're removing perfectly valid setup that existed before:
```suggestion
  ("dune" {>= "1.4.0"} | "dune" {< "1.4.0"} & "base-bytes")
```

dra27: That's true - I hadn't considered transitive dependencies (i.e. `base-bytes` or `ocamlfind` being installed for other reasons)

kit-ty-kate:
Thanks!

PR #23017, opened 2023-01-19
Mark topkg as incompatible with the bytecode-only mode

kit-ty-kate:
See https://github.com/dbuenzli/topkg/issues/140
Wait for the next release of topkg before merging.

This might be a bit overkill as only part of the feature set is broken in bytecode-only mode but it makes it less tedious to maintain this way. If you prefer to only mark the revdeps that actually have C bindings (e.g. mtime) i can do that too but meh.

cc @dbuenzli

### DIFF
--- a/packages/topkg/topkg.0.7.5/opam
+++ b/packages/topkg/topkg.0.7.5/opam
@@ -13,6 +13,9 @@ depends: [
   "ocamlbuild" {build}
   "result"
 ]
+conflicts: [
+  "ocaml-option-bytecode-only"
+]
 build:[[
   "ocaml" "pkg/pkg.ml" "build"
           "--pkg-name" "%{name}%"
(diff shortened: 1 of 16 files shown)


dbuenzli:
I think it's fine that way. However before making a new release should I maybe see if you uncover other peculiarities ?

kit-ty-kate:
I'm not sure what i would be able to test over time. I've tested using mtime and tsdl and build a large portion of your packages in a bytecode-only switch. If something is uncovered by this it would either be by opam-repo-ci here during the release or after by users with some corner cases which is hard to check anyway.

dbuenzli:
Ok will proceed to release in the upcoming days then.

mseri:
Should this be merged?

kit-ty-kate:
Now that topkg.1.0.7 was merged, indeed.

PR #23020, opened 2023-01-20
[new release] curly (0.3.0)

emillon:
Curly is a brain dead wrapper around the curl command line utility

- Project page: <a href="https://github.com/rgrinberg/curly">https://github.com/rgrinberg/curly</a>

##### CHANGES:

* Passthrough case-insensitive PATH and SYSTEMROOT on Windows (@emillon,
  @jonahbeckford, rgrinberg/curly#6, rgrinberg/curly#8)
* Passthrough PATH to Unixes too. (@bikal, rgrinberg/curly#12)
* Add `?follow_redirects` argument to `run` and related functions
  (@rawleyfowler, rgrinberg/curly#5).

### DIFF
--- a/packages/curly/curly.0.3.0/opam
+++ b/packages/curly/curly.0.3.0/opam
@@ -1,0 +1,41 @@
+opam-version: "2.0"
+synopsis:
+  "Curly is a brain dead wrapper around the curl command line utility"
+maintainer: ["rudi.grinberg@gmail.com"]
+authors: ["Rudi Grinberg"]
+license: "ISC"
+homepage: "https://github.com/rgrinberg/curly"
+bug-reports: "https://github.com/rgrinberg/curly/issues"
+depends: [
+  "dune" {>= "2.7"}
+  "ocaml" {>= "4.03.0"}
+  "base-unix"
+  "result"
+  "alcotest" {with-test}
+  "cohttp-lwt-unix" {with-test}
+  "odoc" {with-doc}
+]
+build: [
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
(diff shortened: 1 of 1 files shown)


packages/curly/curly.0.3.0/opam
```diff
@@ -0,0 +1,41 @@
...
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
```
emillon, line 28:
```suggestion
    "@runtest" {with-test & ocaml:version >= "4.08"}
```

emillon: Let's skip tests unless >= 4.08

kit-ty-kate: was that because of this?
```
        Try to specify a more precise request, use a different solver, or increase the allowed time by setting OPAMSOLVERTIMEOUT to a bigger value (currently, it is set to 500.0 seconds).
```
If so this is a known opam bug I'm currently debugging (https://github.com/ocaml/opam/issues/5422) and this change won't really do anything about it (the solver is called before running the tests) and I'd rather have them around just in case

emillon: there was an actual failure on 4.02 and 4.03. I can change that to 4.04+ (the next version will be 4.08+ anyway)

packages/curly/curly.0.3.0/opam
```diff
@@ -0,0 +1,41 @@
...
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test & ocaml:version >= "4.08"}
```
emillon, line 28:
```suggestion
    "@runtest" {with-test & ocaml:version >= "4.04"}
```

PR #23021, opened 2023-01-20
Constrain jupyter releases away from 4.14.1

dra27:
Law of unintended consequences - see https://github.com/akabe/ocaml-jupyter/pull/196 for details.


mseri:
Thanks

PR #23027, opened 2023-01-22
Package commons.1.5.4

aryx:
### `commons.1.5.4`
Yet another set of common utilities
This is a small library of utilities used by Semgrep and
a few other projects developed at r2c.

---
* Homepage: https://semgrep.dev
* Source repo: git+https://github.com/returntocorp/semgrep
* Bug tracker: https://github.com/returntocorp/semgrep/issues

---

### DIFF
--- a/packages/commons/commons.1.5.4/opam
+++ b/packages/commons/commons.1.5.4/opam
@@ -25,8 +25,7 @@ depends: [
   "ppx_deriving" {>= "5.2.1"}
 ]
 
-build: ["dune" "build" "./_build/default/commons.install"]
-install: ["dune" "install" "commons"]
+build: ["dune" "build" "-p" name "-j" jobs]
 url {
   src: "https://github.com/returntocorp/sgrep/archive/refs/tags/1.5.4.tar.gz"
   checksum: [
--- a/packages/commons/commons.1.5.4/opam
+++ b/packages/commons/commons.1.5.4/opam
@@ -1,0 +1,36 @@
+opam-version: "2.0"
+synopsis: "Yet another set of common utilities"
+description: """
+This is a small library of utilities used by Semgrep and
+a few other projects developed at r2c.
+"""
+
+maintainer: "Yoann Padioleau <pad@r2c.dev>"
+authors: [ "Yoann Padioleau <pad@r2c.dev>" ]
+license: "LGPL-2.1-only"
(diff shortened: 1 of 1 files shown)


aryx:
I plan to release other packages developed by r2c in the next few weeks, including one package
per language parsed by semgrep, and ultimately an OPAM package for semgrep.

packages/commons/commons.1.5.4/opam
```diff
@@ -0,0 +1,36 @@
...
+dev-repo: "git+https://github.com/returntocorp/semgrep"
+bug-reports: "https://github.com/returntocorp/semgrep/issues"
+
+depends: [
+  "ocaml" {>= "4.12.0"}
+  "dune" {>= "2.7.0" }
+  "alcotest" {>= "1.5.0"}
+  "ANSITerminal" {>= "0.8.4"}
+  "easy_logging" { = "0.8.1" }
+  "easy_logging_yojson" { = "0.8.1" }
```
kit-ty-kate, line 21: Is there a particular reason for those strict constraints?
```suggestion
  "easy_logging" {>= "0.8.1"}
  "easy_logging_yojson" {>= "0.8.1"}
```

packages/commons/commons.1.5.4/opam
```diff
@@ -0,0 +1,36 @@
...
+  "yojson" {>= "1.7.0"}
+  "re" {>= "1.10.4"}
+  "ppxlib" {>= "0.25.0"}
+  "ppx_deriving" {>= "5.2.1"}
+]
+
+build: ["dune" "build" "./_build/default/commons.install"]
+install: ["dune" "install" "commons"]
```
kit-ty-kate, line 29: The `dune install` command is not needed as opam uses the install file to install the correct files. Furthermore by default dune runs on dev mode not release mode which will break in the future. Using `-p` and removing the install field fixes those two problems:
```suggestion
build: ["dune" "build" "-p" name "-j" jobs]
```

aryx:
@kit-ty-kate I've committed your 2 fixes but now everything fails ... Can I revert and can we merge what was working?

kit-ty-kate:
I would prefer if the package was fixed instead. I've sent a PR upstream: https://github.com/returntocorp/semgrep/pull/6922

We can apply it as a patch here if you don't want to re-release

PR #23028, opened 2023-01-22
[new release] odep (0.1.0)

sim642:
Dependency graphs for OCaml modules, libraries and packages

- Project page: <a href="https://github.com/sim642/odep">https://github.com/sim642/odep</a>

##### CHANGES:

Initial release.

### DIFF
--- a/packages/odep/odep.0.1.0/opam
+++ b/packages/odep/odep.0.1.0/opam
@@ -1,0 +1,48 @@
+opam-version: "2.0"
+synopsis: "Dependency graphs for OCaml modules, libraries and packages"
+maintainer: ["Simmo Saan <simmo.saan@gmail.com>"]
+authors: ["Simmo Saan"]
+license: "MIT"
+homepage: "https://github.com/sim642/odep"
+bug-reports: "https://github.com/sim642/odep/issues"
+depends: [
+  "dune" {>= "2.5"}
+  "ocaml" {>= "4.08"}
+  "sexplib"
+  "ppx_sexp_conv" {>= "v0.13"}
+  "parsexp"
+  "opam-core" {>= "2.1.0"}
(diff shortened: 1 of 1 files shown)


packages/odep/odep.0.1.0/opam
```diff
@@ -0,0 +1,48 @@
...
+bug-reports: "https://github.com/sim642/odep/issues"
+depends: [
+  "dune" {>= "2.5"}
+  "ocaml" {>= "4.08"}
+  "sexplib"
+  "ppx_sexp_conv" {>= "v0.13"}
+  "parsexp"
+  "opam-core" {>= "2.1.0"}
+  "opam-state" {>= "2.1.0"}
+  "opam-format"
```
kit-ty-kate, line 16: I was curious where these packages were used as they need an extra step in `available` and i can't find any use in the package itself outside of its tests. Should they instead marked as tests only? If so are there any more dependencies like this that aren't tagged?
```suggestion
  "opam-core" {>= "2.1.0" & with-test}
  "opam-state" {>= "2.1.0" & with-test}
  "opam-format" {>= "2.1.0" & with-test}
```

sim642: The tests for this package are currently disabled by default because they are too fragile as-is, depending on the set of installed opam packages.

These opam libraries do actually get used for normal usage, e.g. here https://github.com/sim642/odep/blob/bb7419a7077dc5c834c06aa229164a7e1b0b4497/src/depgraph/opam_installed_graph.ml#L99-L107.

kit-ty-kate: Ah weird, the github search didn't show that to me for some reason. Sorry for the wait

kit-ty-kate:
Thanks

PR #23033, opened 2023-01-23
scfg.0.2

zapashcanon:
Hi,

This is the `0.2` release of [scfg](https://git.zapashcanon.fr/zapashcanon/scfg).

Changes are:

- add `get_param_bool`, `get_param_int`, `get_param_pos_int` to the `Query` module
- fix bug with files starting with newlines
- use a proper Format box instead of an int to indent

Cheers,


kit-ty-kate:
Thanks

PR #23015, opened 2023-01-19
Package curses.1.0.11

mbacarella:
### `curses.1.0.11`
Bindings to ncurses
Tools for building terminal-based user interfaces

---
* Homepage: https://github.com/mbacarella/curses
* Source repo: git+https://github.com/mbacarella/curses.git
* Bug tracker: https://github.com/mbacarella/curses/issues

---
:camel: Pull-request generated by opam-publish v2.2.0

PR #23036, opened 2023-01-23
Constrain older versions of curses to OCaml < 5.0.0

mbacarella:
curses doesn't build on 5.0.0 due to C stub symbol changes, which is fixed in this PR
https://github.com/ocaml/opam-repository/pull/23015

This new PR is to constrain older versions of curses to OCaml <5.0.0.


kit-ty-kate:
Thanks a lot!

PR #23038, opened 2023-01-24
Initial release (pa_ppx_string)

chetmurthy:
A PPX rewriter like ppx_string, but based on Camlp5 so compatible with all the other Camlp5-based rewriters.

### DIFF
--- a/packages/pa_ppx_string/pa_ppx_string.0.01/opam
+++ b/packages/pa_ppx_string/pa_ppx_string.0.01/opam
@@ -24,7 +24,7 @@ depends: [
   "not-ocamlfind" { >= "0.01" }
   "ounit" {with-test}
   "fmt"
-  "re"
+  "re" { >= "1.10.4" }
 ]
 build: [
   [make "sys"]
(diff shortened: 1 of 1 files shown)


chetmurthy:
fixed minor bug

chetmurthy:
Looks like every thing passed except for macos-homebrew -- which fails b/c of lack of some conf-perl-* packages.  Can  you release this package ?

packages/pa_ppx_string/pa_ppx_string.0.01/opam
```diff
@@ -0,0 +1,39 @@
...
+maintainer: "Chet Murthy <chetsky@gmail.com>"
+authors: ["Chet Murthy"]
+homepage: "https://github.com/camlp5/pa_ppx_string"
+license: "BSD-3-Clause"
+bug-reports: "https://github.com/camlp5/pa_ppx_string/issues"
+dev-repo: "git+https://github.com/camlp5/pa_ppx_string.git"
+doc: "https://github.com/camlp5/pa_ppx_string/doc"
+
+depends: [
+  "ocaml"       { >= "4.10.0" & < "5.01.0" }
```
mseri, line 18:
```suggestion
  "ocaml"       { >= "4.10.0" & < "5.1.0" }
```
Is the upper bound on "5.1" really needed? We could add it later if something breaks

chetmurthy: Good point, removed the upper bound.

kit-ty-kate:
Thanks

PR #23053, opened 2023-01-25
Add max version bounds to OCADml dependent libraries

geoffder:
Protect against upcoming breaking changes to both OCADml and OSCADml by making the version cutoffs explicit.


mseri:
Thanks

PR #23063, opened 2023-01-25
Add pkg chase

ramsdell:
Removed name and version in opam


kit-ty-kate:
Thanks

ramsdell:
Thank you Kate.  `opam install chase` provides the new version.

PR #23064, opened 2023-01-25
libssh: disable tests

emillon:
The only test in the package tries to hit a private server, which does
not exist anymore (and probably requires authentication):

https://github.com/fxfactorial/ocaml-libssh/blob/v0.1/tests/r_test.ml


kit-ty-kate:
Thanks

kit-ty-kate:
Oh actually i think it might be better to keep the "build" part of the test (especially since it's a C binding) and remove the "run" part of the test

kit-ty-kate:
Thanks

PR #23118, opened 2023-01-25
Package syncweb.0.5

aryx:
### `syncweb.0.5`
Syncweb, Literate Programming meets Unison
Syncweb is a command-line tool enabling programmers to use the
literate programming[1] development methodology, using the noweb[2]
tool, while still being able to modify the generated files
from the literate document. syncweb provides a way to
"synchronize" the possibly modified original document with its
possibly modified views with an interface similar to unison[3]. In
addition, syncweb synchronizes data at a fine grained level by
computing and storing md5sum of the different chunks.

[1] http://en.wikipedia.org/wiki/Literate_programming
[2] http://www.cs.tufts.edu/~nr/noweb/
[3] http://www.seas.upenn.edu/~bcpierce/unison/

---
* Homepage: https://github.com/aryx/syncweb
* Source repo: git+https://github.com/aryx/syncweb
* Bug tracker: https://github.com/aryx/syncweb/issues

---

### DIFF
--- a/packages/syncweb/syncweb.0.5/opam
+++ b/packages/syncweb/syncweb.0.5/opam
@@ -1,0 +1,38 @@
+opam-version: "2.0"
+synopsis: "Syncweb, Literate Programming meets Unison"
+description: """
+Syncweb is a command-line tool enabling programmers to use the
+literate programming[1] development methodology, using the noweb[2]
+tool, while still being able to modify the generated files
+from the literate document. syncweb provides a way to
+"synchronize" the possibly modified original document with its
+possibly modified views with an interface similar to unison[3]. In
+addition, syncweb synchronizes data at a fine grained level by
+computing and storing md5sum of the different chunks.
+
+[1] http://en.wikipedia.org/wiki/Literate_programming
+[2] http://www.cs.tufts.edu/~nr/noweb/
+[3] http://www.seas.upenn.edu/~bcpierce/unison/
+"""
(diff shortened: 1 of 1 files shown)


aryx:
@kit-ty-kate what do I need to put in my dune file to make it work for the other architecture where opam-ci seems to currently fail?
```
Error: No rule found for main/Main.exe
```

aryx:
This is the main/dune file:
```
(executables
 (names Main)
 (libraries
    commons
    frontend
    backend
 )
 (preprocess (pps ppx_deriving.show )) ; LATER: ppx_profiling
 ; byte for ocamldebug
 (modes native byte)
)

(install
 (package syncweb)
 (section bin)
 (files (Main.exe as syncweb))
)
```

kit-ty-kate:
`s/native/best/` should work (requiring native mode means it will fail in bytecode-only mode)

aryx:
Is it availably only in recent version of dune?
I've tried ` (modes best byte)`
and I get ```
File "main/dune", line 10, characters 8-12:
10 |  (modes best byte)
             ^^^^
Error: Unknown value best
```

kit-ty-kate:
Ah nevermind according to the dune documentation it should be `(best exe)` not just `best`. `exe` seems to also be an alias to `(best exe)` if you prefer.

aryx:
ok, will make a new release then

PR #23166, opened 2023-01-26
Disable tests for mrmime.0.4.0 and .0.5.0 on ocaml5

raphael-proust:
These tests are dependent on the PRNG which is *differently predictible* on ocaml5 than on previous ocaml versions.

See https://github.com/mirage/mrmime/issues/91 for additional context.

ping @dinosaure


kit-ty-kate:
Thanks

PR #23172, opened 2023-01-27
Fixes Alt-ergo links

AllanBlanchard:
The link provided for the Alt-Ergo packages is ambiguous on GitHub (2.4.2 both correspond to a branch and a tag) it seems that for a few days, the archive that is obtained is the one that correspond to the branch and not to the tag, leading to checksum failures. This MR disambiguate the link so that it corresponds to the tag.

FYI @Stevendeo

### DIFF
--- a/packages/alt-ergo-lib/alt-ergo-lib.2.4.2/opam
+++ b/packages/alt-ergo-lib/alt-ergo-lib.2.4.2/opam
@@ -37,7 +37,7 @@ build: [
   ["dune" "build" "-p" name "-j" jobs]
 ]
 url {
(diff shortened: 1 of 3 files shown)


mseri:
Thanks. Although the tags on github are unreliable. Would it be possible to download the archive from the tag, add it as a release artifact in the corresponding github releases, and link that?

AllanBlanchard:
I think this a question for OCamlPro :smile:

mseri:
I am going to merge since this fixes installability, I hoe ocamlpro will send a new PR with more stable release artifacts urls

PR #23178, opened 2023-01-27
[hotfix] OCADml (0.3.0)

geoffder:
Types and functions for building CAD packages in OCaml

- Project page: <a href="https://github.com/OCADml/OCADml">https://github.com/OCADml/OCADml</a>
- Documentation: <a href="https://OCADml.github.io/OCADml">https://OCADml.github.io/OCADml</a>

##### CHANGES:

- add gg library dependency
- replaced vector (V{2,3,4}.t), matrix (Affine{2,3}), and bounding box types
 with those from gg
- protect bezier memoization with mutex (OCaml 5 compatibility)

### DIFF
--- a/packages/OCADml/OCADml.0.3.0/opam
+++ b/packages/OCADml/OCADml.0.3.0/opam
@@ -34,8 +34,8 @@ url {
   src:
     "https://github.com/OCADml/OCADml/releases/download/v0.3.0/OCADml-0.3.0.tbz"
   checksum: [
-    "sha256=418f066c06a5abbae7e2751eef1f2bb707637f1cd7b4f7d6413678ac11a390bb"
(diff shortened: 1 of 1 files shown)


geoffder:
Unfortunately found an correctness issue that I thought was worthy of a hotfix, sorry for the extra trouble!

mseri:
can you please release it as 0.3.1 and, if you really want to remove it, mark this version `available: false`?

geoffder:
Ok I'll do that instead. I'd really rather have this one unavailable.

geoffder:
Added `availability: false`, I'll make a PR for v0.3.1.

kit-ty-kate:
Thanks. Please remember in the future that tags should never change after a release as it breaks your end users and ourselves. If something really bad happens you can either remove it if it's really bad, patch it using opam's `patches:` field if you can't do otherwise or simply make a new release (preferred solution)

geoffder:
Roger, will (not) do. I should have asked first before assuming it would be fine just because it was soon after the merge.

PR #23052, opened 2023-01-25
gotd.0.1

zapashcanon:
Hi,

This is the first release of [gotd](https://git.zapashcanon.fr/zapashcanon/gotd), a tool to quickly start an OCaml project.

Cheers,

PR #23048, opened 2023-01-24
sun.0.1

zapashcanon:
Hi,

This is the first release of [sun](https://git.zapashcanon.fr/zapashcanon/sun), a command line tool to take screenshots under Wayland using `grim` and `slurp`.

Cheers,

PR #23180, opened 2023-01-28
fix checksum for scfg.0.1

zapashcanon:
Hi,

In #23048 and #23052 I noticed that something was wrong with `scfg.0.1`'s checksum.

After investigating a bit, it turns out `gitea` is deleting archives automatically after some times using a cronjob. I disabled it so it should not happen again.

I checked that the new archive content is OK.


kit-ty-kate:
Instead of changing the checksum I think it would be better to upload the archive to https://github.com/ocaml/opam-source-archives/ and change the url here. This way people don't have to reinstall this package if this is merged as-is.

I can do it later if you don't have time

kit-ty-kate:
Fair enough. Thanks

PR #21692, opened 2022-06-30
5 packages from c-cube/qcheck

jmid:
This pull-request concerns:
-`ppx_deriving_qcheck.0.2.0`: PPX Deriver for QCheck
-`qcheck.0.19`: Compatibility package for qcheck
-`qcheck-alcotest.0.19`: Alcotest backend for qcheck
-`qcheck-core.0.19`: Core qcheck library
-`qcheck-ounit.0.19`: OUnit backend for qcheck

---
* Homepage: https://github.com/c-cube/qcheck/

---
:camel: Pull-request generated by opam-publish v2.1.0

PR #23191, opened 2023-01-30
Do not run qcheck tests on OCaml 5

Leonidas-from-XIV:
Due to the changes in the RNG the amount of shrink steps and the values that it produces are different.

One could argue that this means that injecting a seed doesn't work, as the seed value to reproduce a run will differ between OCaml 4 and OCaml 5 and this could be considered a bug. I'll leave the decision up to opam-repository developers and @c-cube.

Build failure messages are part of the respective commits, since there is a number of different failures and some are very long.


packages/qcheck-alcotest/qcheck-alcotest.0.19/opam
```diff
@@ -28,4 +28,4 @@ url {
     "md5=6913306eb3065da1dee749d67732cb01"
     "sha512=ab77384e64887c4c88d6e641f433709a83bafc7cd2d05d548fdfee3593beb3bc5c4345db03e324e033ecd4fd2ae600fa92cea3e5753d5e322d13a01ba787eeb6"
   ]
-}
\ No newline at end of file
+}
```
raphael-proust, line 31: There's some end-of-line-at-end-of-file artifact here

```suggestion
}
```

packages/qcheck-core/qcheck-core.0.19/opam
```diff
@@ -30,4 +30,4 @@ url {
     "md5=6913306eb3065da1dee749d67732cb01"
     "sha512=ab77384e64887c4c88d6e641f433709a83bafc7cd2d05d548fdfee3593beb3bc5c4345db03e324e033ecd4fd2ae600fa92cea3e5753d5e322d13a01ba787eeb6"
   ]
-}
\ No newline at end of file
+}
```
raphael-proust, line 33: Same here

```suggestion
}
```

packages/qcheck-ounit/qcheck-ounit.0.19/opam
```diff
@@ -28,4 +28,4 @@ url {
     "md5=6913306eb3065da1dee749d67732cb01"
     "sha512=ab77384e64887c4c88d6e641f433709a83bafc7cd2d05d548fdfee3593beb3bc5c4345db03e324e033ecd4fd2ae600fa92cea3e5753d5e322d13a01ba787eeb6"
   ]
-}
\ No newline at end of file
+}
```
raphael-proust, line 31: Same here

```suggestion
}
```

packages/qcheck/qcheck.0.19/opam
```diff
@@ -32,4 +32,4 @@ url {
     "md5=6913306eb3065da1dee749d67732cb01"
     "sha512=ab77384e64887c4c88d6e641f433709a83bafc7cd2d05d548fdfee3593beb3bc5c4345db03e324e033ecd4fd2ae600fa92cea3e5753d5e322d13a01ba787eeb6"
   ]
-}
\ No newline at end of file
+}
```
raphael-proust, line 35: Here too

```suggestion
}
```

raphael-proust (commented):
Apart from the minor remarks in the comments I left, this looks ok.

jmid:
Expect test author and occasional QCheck maintainer here! :wave: PR looks good to me too!
I recently added separate expect output for OCaml 5 - but that went into version 0.20: https://github.com/c-cube/qcheck/pull/262

There's now also https://github.com/ocaml/stdlib-random that we might consider switching to (this is not the first time the RNG changed)

packages/qcheck-alcotest/qcheck-alcotest.0.19/opam
Leonidas-from-XIV: Yes, my vim automatically adds newlines at end of text files where they are missing. Pushed a commit to remove them, because for some reason Github can't apply your suggestions.

dinosaure:
That's expected that `qcheck-core` (`0.18`) fails on some platformes (like `arm32` or `ppc64`)?

jmid:
Yes:
- it is expected that 0.18 fails on 32-bit architectures due to expect test failures against 64-bit `.expected` files. This was solved in https://github.com/c-cube/qcheck/pull/250 that went into 0.19
- ppc64 failures are due to missing stack overflow protection on that architecture - see https://github.com/ocaml/opam-repository/pull/21692#issuecomment-1179095063 for more details and pointers

dinosaure:
Ok, so let's merge. Thanks!

PR #23197, opened 2023-01-31
Update cstruct bounds of vhd-format

psafont:
6.0.0 is too low of an upper bound, relax it to be 6.1.0, like vhd-format-lwt


dinosaure:
Thanks!

PR #23199, opened 2023-01-31
[new release] jose (0.8.2)

ulrikstrid:
CHANGES:

- JWS now properly checks the signature. Reported by @nankeen and fixed by @ulrikstrid. CVE-2023-23928

This is a critical bugfix release that fixes a bug that has been present in all(?) previous versions

https://github.com/ulrikstrid/reason-jose/security/advisories/GHSA-7jj9-6qwv-wpm7

PR #23201, opened 2023-01-31
jose: mark unavailable

anmonteiro:
Mark all versions of JOSE prior to 0.8.2 unavailable, due to https://github.com/ulrikstrid/reason--jose/security/advisories/GHSA-7jj9-6qwv-wpm7

This should probably be merged after #23199


kit-ty-kate:
Thanks. cc @ulrikstrid

PR #23213, opened 2023-02-01
[new release] dolmen_type, dolmen_model, dolmen_lsp, dolmen_loop, dolmen_bin and dolmen (0.8.1)

Gbury:

### UI

- Fix handling of size/time limits on windows (PR#117)
- Fix spurious printing of backtraces (PR#118)
- Add release binaries for windows

### LSP

- Add option for the lsp to read preludes before checking
  each file (PR#116)
- The LSP now sends an empty list of diagnostics upon closing
  a file (PR#116)

### Parsing

- Fix a bug related to alt-ergos function definition, which
  were previously alwyas non-recursive. Now, alt-ergo's function
  definitions are always recursive (PR#123)
- Add `parse_raw_lazy` to parse a string into a lazy list of
  statements (PR#125)

### Typing

- Properly add binding locations for implicit type variables
  (PR#123)
- Ensure that type of recursively defined symbols are freshened
  to avoid type variables sharing between declaration and
  definition (PR#123)

### Loop

- New module to implement Alarms (size/time limits) (PR#117)
- Add optional argument to `Pipeline.run` to specify
  an alarm implementation (PR#117)
- Add a `bt` key to the state to record whether we should print
  backtraces (PR#118)
- Use `parse_raw_lazy` to parse raw contents in full mode if/when
  necessary (PR#125)

### DIFF
--- a/packages/dolmen/dolmen.0.8.1/opam
+++ b/packages/dolmen/dolmen.0.8.1/opam
@@ -1,0 +1,46 @@
+opam-version: "2.0"
+maintainer: "Guillaume Bury <guillaume.bury@gmail.com>"
+authors: "Guillaume Bury <guillaume.bury@gmail.com>"
+license: "BSD-2-Clause"
+build: [
+  ["dune" "subst"] {dev}
+  ["dune" "build" "-p" name "-j" jobs "@install" "@runtest" {with-test} "@doc" {with-doc}]
+]
+depends: [
+  "ocaml" {>= "4.08"}
+  "menhir" {>= "20211230" }
+  "dune" { >= "3.0" }
+  "fmt" { >= "0.8.7" }
+  "seq"
+  "odoc" { with-doc }
+  "qcheck" { with-test }
+  "mdx" { with-test }
+]
+
+tags: [ "parser" "logic" "tptp" "smtlib" "dimacs" ]
+homepage: "https://github.com/Gbury/dolmen"
+dev-repo: "git+https://github.com/Gbury/dolmen.git"
+bug-reports: "https://github.com/Gbury/dolmen/issues"
+
+doc: "https://gbury.github.io/dolmen"
+synopsis: "A parser library for automated deduction"
+description:
+"Dolmen is a parser library. It currently targets languages used in automated theorem provers,
+but may be extended to other domains.
+
+Dolmen provides functors that takes as arguments a representation of terms and statements,
+and returns a module that can parse files (or streams of tokens) into the provided representation
+of terms or statements. This is meant so that Dolmen can be used as a drop-in replacement of existing
+parser, in order to factorize parsers among projects.
+
+Additionally, Dolmen also provides a standard implementation of terms and statements that cna be
+used ot instantiate its parsers."
+url {
+  src:
+    "https://github.com/Gbury/dolmen/releases/download/v0.8.1/dolmen-0.8.1.tbz"
+  checksum: [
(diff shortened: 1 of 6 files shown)


packages/dolmen/dolmen.0.8.1/opam
```diff
@@ -0,0 +1,46 @@
+opam-version: "2.0"
+maintainer: "Guillaume Bury <guillaume.bury@gmail.com>"
+authors: "Guillaume Bury <guillaume.bury@gmail.com>"
+license: "BSD-2-Clause"
+build: [
+  ["dune" "subst"] {dev}
+  ["dune" "build" "-p" name "-j" jobs]
```
raphael-proust, line 7: Any reason not to allow the installation to run the tests of the package?

```suggestion
  ["dune" "build" "-p" name "-j" jobs "@runtest" {with-test}]
```

packages/dolmen_bin/dolmen_bin.0.8.1/opam
```diff
@@ -0,0 +1,42 @@
+opam-version: "2.0"
+maintainer: "Guillaume Bury <guillaume.bury@gmail.com>"
+authors: "Guillaume Bury <guillaume.bury@gmail.com>"
+license: "BSD-2-Clause"
+build: [
+  ["dune" "subst"] {dev}
+  ["dune" "build" "-p" name "-j" jobs]
```
raphael-proust, line 7: Same remark here and in other packages.

packages/dolmen/dolmen.0.8.1/opam
Gbury: there's no reason, just that I wrote the file by hand some time ago, and didn't update it in a while. I'll add the test part (and probably also the doc part I suppose).

Gbury:
I updated the build instructions to also build tests and doc when appropriate (matching commit in the dolmen repo : https://github.com/Gbury/dolmen/commit/18b02c1c23598df5c09601667093e1ff5f42ab43 )

packages/dolmen_bin/dolmen_bin.0.8.1/opam
Gbury: fixed

Gbury:
So, the addition of the tests showed that some tests that should only be run on the main package (`dolmen`) are also run on other packages (because I forgot a `package` stanza in the corresponding dune file), so as a workaround I added `mdx` as a test-only dependency on all of the `dolmen_*` packages (which is not a problem, since `dolmen` already has such a dependency, and all other packages depend on `dolmen`).
I also committed the correct fix in the dolmen repo (see https://github.com/Gbury/dolmen/commit/2cddfe2d24e45e58d05696f18e9af984fe3e41f7 )

Gbury:
I'm investigating the 5.0 failures

Gbury:
The 5.0 failures are related to the following bug in ocaml 5.0 : https://github.com/ocaml/ocaml/issues/11995
I have a way to fix the code of dolmen to make it compatible (and bug-free) for all versions of the compiler, so I'll redo the 0.8.1 release with the fix, and add conflicts between ocaml `5.0` and older releases of dolmen.

Gbury:
So, I re-did the `0.8.1` release, including a fix that should make dolmen work on ocaml `5.0.0`. Hopefully, the CI should now be happy, ^^

kit-ty-kate:
Thanks

PR #23224, opened 2023-02-02
initial release of ocaml buildscripts for camlp5 and pa_ppx packages

chetmurthy:
These are build-scripts used by camlp5 and pa_ppx, to replace the ones that are currently used, and that rely on conf-perl-* packages.  Once these scripts are released, I can push releases to camlp5 and the pa_ppx_* packages that no longer use these Perl packages, so they should all build without a hitch on macos-homebrew.


chetmurthy:
Hi, is there anything else I need to do, to be ready for merging ?

kit-ty-kate:
Thanks

PR #23229, opened 2023-02-04
Package ulex-camlp5.1.3

sacerdot:
### `ulex-camlp5.1.3`
A lexer generator for Unicode (backported to camlp5)

---
* Homepage: https://github.com/whitequark/ulex
* Source repo: git+https://github.com/whitequark/ulex.git
* Bug tracker: https://github.com/whitequark/ulex/issues

---

### DIFF
--- a/packages/ulex-camlp5/ulex-camlp5.1.3/opam
+++ b/packages/ulex-camlp5/ulex-camlp5.1.3/opam
@@ -12,7 +12,7 @@ build: [
 ]
 install: [make "install"]
 depends: [
-  "ocaml" {>="5.0.0"} 
+  "ocaml" {>="4.14.1"}
   "ocamlfind" {build}
   "camlp5" {>= "8.00.04"}
   "camlp-streams"


sacerdot:
The only change of this PR is to allow compilation on ocaml 4.14.1 too (it was allowed for ocaml 5.0 only)

kit-ty-kate:
The status of this package seems very confusing. To me it looks like a fork given the url is in you fork repository but a 1.3 version pointing to the camlp5 branch upstream also exists? What are the relations between the two?

sacerdot:
@kit-ty-kate Indeed. The two are completely in sync. You can make it point to your repository, but you will need to make also releases that track camlp5 in parallel with releases that track camlp4. The previous opam versions used already to point to your main for camlp4 and to my fork for camlp5, but I always kept your camlp5 branch update.

sacerdot:
@kit-ty-kate I think that the ball is in your court.
(Note that I have issued a PR upstream to be completely in sync even on the opam/META files)

kit-ty-kate:
Thanks, I’m very sorry for the wait.

Sorry we didn’t notice the discrepancy before, for next release(s) it would be less confusing if the tarball came from the upstream repository even if your proposed tarball is morally the same. @mseri is going to open a PR to switch the archives to point to the upstream ones (we checked they are equivalent), if that’s ok with you.

PR #23235, opened 2023-02-05
tsdl.0.9.7+: not available on OpenSUSE Leap

edwintorok:
OpenSUSE Leap 15.4 has SDL2 version 2.0.8.
However tsdl 0.9.7 [requires SDL_DISPLAYEVENT](https://github.com/dbuenzli/tsdl/blob/master/CHANGES.md#v097-2019-07-19-zagreb) which got introduced in SDL2 [version 2.0.9](https://github.com/libsdl-org/SDL/blob/264da8c127f8bf8930fa62f106763855efbf2a93/WhatsNew.txt#L439)

tsdl.0.9.6 is latest version that installs on OpenSUSE Leap 15.x. However OpenSUSE Tumbleweed has a newer version, so assume that version will have it, and make the package unavailable only on 'Leap < 16'.

The package is unavailable on: OpenSUSE-Leap <= 15, therefore it is available if distro != OpenSUSE-Leap, or distro == OpenSUSE-Leap && version > 16.

Without this change there is an 'ocaml-ci' failure building packages downstream of 'tsdl': https://ci.ocamllabs.io/github/sanette/bogue/commit/c05d874d5bc217625abdf1fe8c73dc8acd0a8986/variant/opensuse-15.3-5.0_opam-2.1

### DIFF
--- a/packages/conf-sdl2/conf-sdl2.1/opam
+++ b/packages/conf-sdl2/conf-sdl2.1/opam
@@ -5,6 +5,7 @@ homepage: "http://libsdl.org/"
 bug-reports: "https://github.com/ocaml/opam-repository/issues"
 license: "Zlib"
 build: [["pkg-config" "sdl2"]]
+available: [!((os-distribution = "oracelinux" | os-distribution = "ol") & os-version < 9)]
 depexts: [
   ["sdl2-dev"] {os-distribution = "alpine"}
   ["sdl2"] {os-distribution = "arch"}
--- a/packages/tsdl/tsdl.0.9.7/opam
+++ b/packages/tsdl/tsdl.0.9.7/opam
@@ -16,7 +16,11 @@ depends: [
   "conf-sdl2"
   "ctypes" {>= "0.14.0"}
   "ctypes-foreign" ]
-available: [ (os-distribution != "opensuse-leap" | os-version >= 16) ]
+available: [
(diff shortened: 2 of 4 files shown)


kit-ty-kate:
The issue with making packages unavailable on specific platforms for this kind of errors is that it doesn't take what users have installed or available at all. It's only a guess based on the default version of SDL on those platforms' repositories.

For example this will break users that use custom apt repository on ubuntu 18.04 or who installed sdl by hand, and the only way to fix that on their end is extremely convoluted and annoying.

In my opinion, if this is CI specific, `x-ci-accept-failures` should be used instead.

edwintorok:
it'd be better if dependencies could be expressed based on package versions, but it doesn't look like opam would support that currently.
Is x-ci-accept-failures transitive? i.e. will it make reverse dependencies unavailable in the CI too?

kit-ty-kate:
> Is x-ci-accept-failures transitive? i.e. will it make reverse dependencies unavailable in the CI too?

Yes and yes. However only opam-repo-ci supports the field currently. There is a PR for OCaml-CI here if you want to add some weight to it: https://github.com/ocurrent/ocaml-ci/pull/753

edwintorok:
Lets close this for now until something better is available then for use in both CIs. The immediate CI failure in bogue got worked around by making it unavailable on opensuse-leap, and the problem will likely sort itself out the next time opensuse does a major release (it works on tumbleweed).

PR #23236, opened 2023-02-05
[new release] eio_main, eio_luv, eio_linux and eio (0.8.1)

talex5:
CHANGES:

Some build fixes:

- Fix build on various architectures (@talex5 ocaml-multicore/eio#432).
  - Work around dune `%{system}` bug.
  - eio_luv: fix `max_luv_buffer_size` on 32-bit platforms.

- Add missing test-dependency on MDX (@talex5 ocaml-multicore/eio#430).

### DIFF
--- a/packages/eio/eio.0.8.1/opam
+++ b/packages/eio/eio.0.8.1/opam
@@ -1,0 +1,55 @@
+opam-version: "2.0"
+synopsis: "Effect-based direct-style IO API for OCaml"
+description: "An effect-based IO API for multicore OCaml with fibers."
+maintainer: ["anil@recoil.org"]
+authors: ["Anil Madhavapeddy" "Thomas Leonard"]
+license: "ISC"
+homepage: "https://github.com/ocaml-multicore/eio"
+doc: "https://ocaml-multicore.github.io/eio/"
+bug-reports: "https://github.com/ocaml-multicore/eio/issues"
+depends: [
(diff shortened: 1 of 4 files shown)


talex5:
- `s390x-ocaml-5.0` is expected to fail for `eio_luv` (until https://github.com/aantron/luv/issues/144 is fixed). However, s390x should pass for `eio_main`, which uses `eio_linux` by default instead, so this isn't a big problem.
- The `lwt_eio` revdep tests are likely to time out (until https://github.com/ocsigen/lwt/issues/981 is fixed).

talex5:
I think this is ready to merge now. The two expected failures correspond to real bugs in other libraries, and the mac failure (`Uncaught exception: Failure("\"rsync\"`) is unrelated.

raphael-proust:
I made https://github.com/ocaml/opam-repository/pull/23271 to mark `luv` as not available on s390x.

@talex5 do you think that's the appropriate change to make? Do you know if the error exists also for ocaml≠5?

talex5:
> I made https://github.com/ocaml/opam-repository/pull/23271 to mark luv as not available on s390x.

I'd rather not do that. `eio_main` depends on `eio_luv` (as a fallback) and making `luv` uninstallable will therefore prevent `eio_main` from working in all cases.

kit-ty-kate:
Thanks

PR #23241, opened 2023-02-05
doculib 1.3.4 release

nguermond:
Changes:
* Remove tag column from metadata search
* bug fix: existing metadata was overwritten when metadata was searched


kit-ty-kate:
Thanks

PR #23252, opened 2023-02-07
Prevent gd and magic from building with ocaml 5.

toots:
The packages will build but won't link because they are using deprecated OCaml C symbols.

### DIFF
--- a/packages/magic/magic.0.7.3/opam
+++ b/packages/magic/magic.0.7.3/opam
@@ -8,7 +8,10 @@ remove: [
   ["./configure" "--prefix" prefix "--datarootdir" prefix]
   [make "uninstall"]
 ]
-depends: ["ocaml" "ocamlfind"]
+depends: [
+  "ocaml" {< "5.0"}
+  "ocamlfind"
+]
 depexts: [
   ["file-dev"] {os-distribution = "alpine"}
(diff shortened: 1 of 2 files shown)


packages/magic/magic.0.7.3/opam
```diff
@@ -8,7 +8,10 @@ remove: [
   ["./configure" "--prefix" prefix "--datarootdir" prefix]
   [make "uninstall"]
 ]
-depends: ["ocaml" "ocamlfind"]
+depends: [
+  "ocaml" {< "5.0"}
+  "ocamlfind"
```
raphael-proust, line 13: There seem to be some packaging error on Archlinux where the configure script fails because it can't find `which`: `./configure: line 1708: which: command not found`.

```suggestion
  "ocamlfind"
  "conf-which" {build}
```

raphael-proust:
The remaining errors seem to be system issues installing system dependencies. There are also some harder to understand errors on oracle8: one of them seem to be related to system packages, the other has no output.

I think it's ok to merge now. Or more precisely, I don't know how to improve the installability (but others might have  suggestions).

toots:
Thanks for the review effort! I think at the very least not allowing these packages to build when we know the build will fail is definitely an improvement.

mseri:
Thanks

PR #23254, opened 2023-02-07
conf-expat upgrade: more distributions considered

sacerdot:


sacerdot:
- the oraclelinux failure is strange because ocaml-expat compiles on oraclelinux
- the homebrew failure is strange (interrupted system call). Maybe a transient one?

sacerdot:
It works. Thanks to @kit-ty-kate for the oraclelinux spell. The homebrew failures are indeed transients (probably one machine in the pool has wrong permissions on installation dir).

kit-ty-kate:
Thanks a lot

PR #23256, opened 2023-02-08
Camlp5: Rewrote last scripts that depended on conf-perl-* packages

chetmurthy:
viz, conf-perl-ipc-system-simple,conf-perl-string-shellquote

So now camlp5 should install cleanly on macos-homebrew


chetmurthy:
Huzzah!  No more build-breaks on macos-homebrew!

PR #23258, opened 2023-02-08
Package opine.0.1.3

ArulselvanMadhavan:
### `opine.0.1.3`
Python AST unparse implementation in OCaml
Opine is an OCaml library that unparses the python AST back to python code.

---
* Homepage: https://github.com/ArulselvanMadhavan/opine
* Source repo: git+https://github.com/ArulselvanMadhavan/opine.git
* Bug tracker: https://github.com/ArulselvanMadhavan/opine/issues

---


packages/opine/opine.0.1.2/opam
```diff
@@ -0,0 +1,51 @@
+# This file is generated by dune, edit dune-project instead
+opam-version: "2.0"
+synopsis: "Python AST unparse implementation in OCaml"
+description:
+  "Opine is an OCaml library that unparses the python AST back to python code."
+maintainer: ["arulselvan1234@gmail.com"]
+authors: ["Arulselvan Madhavan"]
+license: "MIT"
```
raphael-proust, line 8: There doesn't seem to be any license info in your repo (no `LICENSE` file and no license headers). You should probably add some info over there too.

packages/opine/opine.0.1.2/opam
```diff
@@ -0,0 +1,51 @@
...
+tags: ["topics" "ocaml" "python" "ast"]
+homepage: "https://github.com/ArulselvanMadhavan/opine"
+doc: "https://url/to/documentation"
+bug-reports: "https://github.com/ArulselvanMadhavan/opine/issues"
+depends: [
+  "ocaml"
```
raphael-proust, line 14:
```suggestion
  "ocaml" {>= "4.13"}
```

You are using the `Option.value_exn @@ …` and `Option.value` takes optional parameters. There was a change in the type-checker in 4.13 to make `@@` more transparent. This makes your usage valid on 4.13 and later.

packages/opine/opine.0.1.2/opam
```diff
@@ -0,0 +1,51 @@
...
+  "dune" {>= "3.6"}
+  "ppx_hash"
+  "ppx_compare"
+  "ppx_sexp_conv"
+  "ppx_make"
+  "ppx_deriving"
+  "pyre-ast"
+  "stdio"
+  "base"
```
raphael-proust, line 23: I think there is a missing lower-bound on this as well. Specifically there is an issue during the lower-bound check that has to do with the interface of the `Result` module. I'm guessing you inherit the `Result` module from `base` (although I am not sure). And so I suspect you need to add the following lower-bound constraint:

```suggestion
  "base" {>= "v0.15.0"}
```

packages/opine/opine.0.1.2/opam
ArulselvanMadhavan: Added a LICENSE file

packages/opine/opine.0.1.2/opam
ArulselvanMadhavan: Updated

ArulselvanMadhavan:
@raphael-proust Thanks for taking a look! I have made the suggested changes

PR #23259, opened 2023-02-08
ocamlformat.0.24.1: require ocaml-version < 3.6.0

emillon:
This OCamlformat release base has a comparison against `v4_14` for a feature that has been added in `4.14.0`. So it breaks when using `ocaml-version.3.6.0`, which moves `v4_14` (the logical latest release in 4.14.x) to `4.14.1`.

The check has been fixed in ocaml-ppx/ocamlformat#2240.

(cc @gpetiot)

### DIFF
--- a/packages/ocamlformat/ocamlformat.0.24.1/opam
+++ b/packages/ocamlformat/ocamlformat.0.24.1/opam
@@ -19,7 +19,7 @@ depends: [
   "menhir" {>= "20201216"}
   "menhirLib" {>= "20201216"}
   "menhirSdk" {>= "20201216"}
-  "ocaml-version" {>= "3.3.0"}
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Thanks

kit-ty-kate:
Actually wouldn’t the old versions require that change too?

emillon:
hmm, let me check which versions are affected

emillon:
The offending test is present in versions 0.20.0 to 0.24.1 included, but for all the versions except 0.24.1:
- ocamformat depends on `dune {< 3 & with-test}`
- ocaml-version >= 3.6.0 (necessary for the bug to trigger) requires dune >= 3

So there are affected versions but which we can't test. I'll open a followup PR.

PR #23260, opened 2023-02-08
current_git: skip tests that hang

emillon:
These releases trigger an infinite loop in certain versions of git.

This is discussed in ocurrent/ocurrent#386 and ocurrent/ocurrent#405.

cc @maiste


kit-ty-kate:
Thanks

PR #23271, opened 2023-02-09
Mark luv as not available on s390x

raphael-proust:
Issue upstream: https://github.com/aantron/luv/issues/144


kit-ty-kate:
Thanks

PR #23276, opened 2023-02-09
ocamlformat: add more bounds against ocaml-version

emillon:
Followup to #23259:

The offending test is present in versions 0.20.0 to 0.24.1 included, but
for all the versions except 0.24.1 the problematic combination can not
exist when `{with-test}` is set since ocamformat depends on dune < 3
and ocaml-version requires dune >= 3. Still these need to be fixed.


kit-ty-kate:
Thanks!

PR #23287, opened 2023-02-10
Pa ppx remove conf perl packages dependency

chetmurthy:
This should make everything build on macos-homebrew. [retrying to avoid build-failure that seems to have jammed GIT]


chetmurthy:
This is still busted.  closing.

PR #23309, opened 2023-02-14
new file:   packages/odnnr/odnnr.2.0.3/opam

UnixJunkie:


packages/odnnr/odnnr.2.0.3/opam
```diff
@@ -0,0 +1,46 @@
...
+"""
+post-messages:
+  """
+Please interact with pip3 then R to install required dependencies in user-space:
+pip3 install tensorflow keras
+R
+install.packages('keras', repos='http://cran.r-project.org')
+library(keras)""" {failure}
+extra-files: [
+  ["install_keras.r" "md5=ad1bb71c78db06e7a38b24ec6b34273d"]
```
kit-ty-kate, line 41: This file is missing

packages/odnnr/odnnr.2.0.3/files/install_keras.r
```diff
@@ -0,0 +1,8 @@
+# for keras and tensorflow, let's assume the user has installed them in
+# user-space as a prerequisite using 'pip3 install tensorflow keras'
+# here, we just test that the keras library can be loaded in R
+pkg <- 'keras'
+if (!require(pkg, character.only = TRUE)) {
+  install.packages(pkg, dependencies = TRUE, repos='http://cran.r-project.org')
```
kit-ty-kate, line 6: I know this pattern is everywhere in your R-related packages, and I’ve kinda let that slip under the rug for all those years but how is this supposed to work without network access? (sandbox)

UnixJunkie: Isn't the sandbox too tight, if we cannot download listed things at install time?
Of course, R will also download all the required dependencies...

PR #23311, opened 2023-02-14
[new release] tls, tls-mirage, tls-lwt, tls-eio and tls-async (0.16.0)

hannesm:
Transport Layer Security purely in OCaml

- Project page: <a href="https://github.com/mirleft/ocaml-tls">https://github.com/mirleft/ocaml-tls</a>
- Documentation: <a href="https://mirleft.github.io/ocaml-tls/doc">https://mirleft.github.io/ocaml-tls/doc</a>

##### CHANGES:

* BREAKING: new opam package tls-lwt (formerly tls.lwt), in dune:
  (libraries tls.lwt) should now be libraries (tls-lwt)
  (mirleft/ocaml-tls#468 @hannesm, reported mirleft/ocaml-tls#449 by @mbacarella)
* tls: update to mirage-crypto 0.11 API (mirleft/ocaml-tls#468 @hannesm)
* tls: relax SignatureAlgorithms extension handling to allow OpenSSL
  interoperability tests with TLS 1.0 and TLS 1.1 (mirleft/ocaml-tls#469 @hannesm)
* tls: remove Utils.filter_map and and Utils.option, use Stdlib instead (mirleft/ocaml-tls#455
  @hannesm)
* tls: do not globally open Utils (mirleft/ocaml-tls#455 @hannesm)
* tls: export log source of Tracing module (mirleft/ocaml-tls#461 @bikallem)
* tls: remove unused ciphersuites to reduce binary size (mirleft/ocaml-tls#467 @hannesm)
* tls-lwt: do not catch out of memory exception (mirleft/ocaml-tls#469 @hannesm)
* tls-eio: add fuzz testing using crowbar (mirleft/ocaml-tls#456 mirleft/ocaml-tls#463 @talex5)
* tls-eio: update to eio 0.7 (mirleft/ocaml-tls#456 @talex5)
* tls-eio: fix test for develop with vendoring (mirleft/ocaml-tls#462 @bikallem)

PR #23315, opened 2023-02-14
Tls 016 revdeps2

hannesm:
third part of fixing revdeps #23311


packages/sihl/sihl.0.1.10/opam
```diff
@@ -37,6 +37,7 @@ depends: [
   "alcotest-lwt" {>= "1.2.0" & < "3.0.0" & with-test}
   "cohttp-lwt-unix" {>= "2.5.1"& with-test}
   "odoc" {with-doc}
+  "mirage-crypto"
```
kit-ty-kate, line 40: I don't think this is the right fix here. `sihl` uses the `Mirage_crypto` module but only implicitly through `tls` and does not require it in its dune file.

The only fix possible here I think is to add `< "0.16.0"` to the `tls` dependency above

PR #23332, opened 2023-02-16
[new release] dns (7.0.0) happy-eyeballs (0.5.0)

hannesm:
DNSSec support for OCaml-DNS

- Project page: <a href="https://github.com/mirage/ocaml-dns">https://github.com/mirage/ocaml-dns</a>
- Documentation: <a href="https://mirage.github.io/ocaml-dns/">https://mirage.github.io/ocaml-dns/</a>

##### CHANGES:

* BREAKING: dns-client is split into 3 packages: dns-client-lwt,
  dns-client-mirage. If your dune file contains dns-client.lwt, use
  dns-client-lwt now. If your dune file contains dns-client.mirage, use
  dns-client-mirage now (mirage/ocaml-dns#331 @hannesm)
* update to mirage-crypto 0.11.0 API changes and tls 0.16.0 packaging changes
  (mirage/ocaml-dns#331 @hannesm)
* dns-client.resolvconf: add line number to parser (mirage/ocaml-dns#334 @hannesm, inspired by
  mirage/ocaml-dns#328 @bikallem)
* dns-client.resolvconf: allow zone idx (RFC 4007) for IPv6 entries
  (mirage/ocaml-dns#334 @hannesm, inspired by mirage/ocaml-dns#328 @bikallem)
* dns-server.zone: allow zone files without final newline (add a newline to the
  buffer if the last character is not \n) (mirage/ocaml-dns#333 @hannesm)
* dns-client-{lwt,mirage}: do not log when the resolver closed the connection,
  but there are no pending requests (mirage/ocaml-dns#332 @reynir)
* dns-certify: in Dns_certify_mirage use X509.Private_key.of_string, the
  behaviour when both key_data and key_seed is provided changed, and leads to
  an exception now (mirage/ocaml-dns#330 @hannesm)

PR #23334, opened 2023-02-16
Dns7 revdeps part one

hannesm:
see #23332


hannesm:
CI looks fine, failures are unrelated. Would be great to merge this to unblock #23332

dinosaure:
Regardless MacOS errors, only `equinox` remains with an error (/cc @maiste) which is unrelated to this PR. Thanks for the revdeps work.

PR #23368, opened 2023-02-20
dune 3.7: fix revdeps

emillon:
Two packages got broken by dune 3.7: commons and chamelon. They were relying on behavior in dune's dependency cycle detection that we classified as a bug (see ocaml/dune#7018).

Since the latest versions are affected, we sent patches upstream: yomimono/chamelon#18, returntocorp/semgrep#7173.


emillon:
Marking as draft since 3.7 is not merged (#23349).

kit-ty-kate:
Thanks

PR #23369, opened 2023-02-20
Add `build` field to XML-Light 2.4

Leonidas-from-XIV:
The current instructions are missing a `build` field (that `xml-light.2.3` has). It works because the `install` field uses the `install_ocamlfind` target which depends on `all` and `opt` but it seems a bit inelegant to build in the `install` step.

This was detected as opam-monorepo classifies packages without `build` instructions as packages without a source, which is not the case for `xml-light`. A PR exists
https://github.com/tarides/opam-monorepo/pull/376 but there is the question whether missing `build` steps is a packaging mistake or not.

In any case, this PR is a proposal to fix the issue and I'm looking forward to hear the input of opam-repository maintainers on this question.

### DIFF
--- a/packages/xml-light/xml-light.2.4/opam
+++ b/packages/xml-light/xml-light.2.4/opam
@@ -6,6 +6,10 @@ depends: [
   "ocamlfind"
 ]
 dev-repo: "git+https://github.com/ncannasse/xml-light"
+build: [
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Thanks

PR #23380, opened 2023-02-21
Add irc-client.0.7.1

johnelse:


raphael-proust:
CI errors are either on experimental runners (macos) or some system package installation failures which seem to be network related (Operation too slow. Less than 1 bytes/sec transferred the last 10 seconds).

mseri:
Thanks

PR #23376, opened 2023-02-20
Octez 16.0~rc2 packages

killian-delarue:
Packages for the second release candidate for Octez v16.0

PR #23384, opened 2023-02-21
LablGL 1.07

garrigue:
New version of LablGL for compatibility with OCaml 5.0.0.


garrigue:
I see that compilation fails on architectures that do not have a native code compiler.
Is there a way in the opam file to make part of the compilation/installation dependent on its availability?

raphael-proust:
@garrigue According to https://github.com/ocaml/opam-repository/pull/23376#issuecomment-1439950764 you can use

```
conflicts: ["ocaml-option-bytecode-only"]
```

to mark a whole package as needing ocamlopt

Or you can use `!ocaml:native` in your build rule filters.

garrigue:
Ok, it is not exactly what I was asking for, but it was hidden in your answer.
By writing `[make "glutopt"] {ocaml:native}` , the package also install on bytecode-only architectures.
So the package itself looks fine.

There is still something strange: I see a huge number of spurious revdeps, i.e. packages that do not depend on lablgl but are tested by the CI. Any idea what this comes from ?

mseri:
It is acutally due to lablgtk. If you run `opam list --depends-on lablgl --recursive --all --depopts` you will see the full list (thanks @kit-ty-kate)

mseri:
Non of the failures depend on this. I have sent a PR to fix all the one that needed upper or lower bounds

PR #23391, opened 2023-02-22
[new release] plist-xml (0.4.0)

alan-j-hu:
Reading and writing of plist files in the XML format in pure OCaml

- Project page: <a href="https://github.com/alan-j-hu/ocaml-plist-xml">https://github.com/alan-j-hu/ocaml-plist-xml</a>
- Documentation: <a href="https://alan-j-hu.github.io/ocaml-plist-xml/">https://alan-j-hu.github.io/ocaml-plist-xml/</a>

##### CHANGES:

Version 0.4.0 is a major rewrite with an API redesign that is not
backwards-compatible. `plist-xml` now uses `xmlm` instead of `markup` for the
underlying XML codec. The package `plist-xml-lwt` no longer exists. In the
new API, `plist-xml` has both a high-level tree interface and a low-level
streaming interface.


kit-ty-kate:
I had already added the missing dependency in https://github.com/ocaml/opam-repository/pull/23395

alan-j-hu:
Ok, thanks. However, because release 0.4.0 is an API redesign, the revdep build should still fail. I went and fixed the merge conflict, added a lower bound for markup, and alphabetized the test dependencies.

alan-j-hu:
That's not what we were talking about. My `textmate-language` has an implicit test dependency on `markup` and an explicit test dependency on `plist-xml`. This new version of `plist-xml` uses `xmlm` instead of `markup`. So, `textmate-language` failed to build because `markup` was missing, but adding `markup` as an explicit dependency is not enough to make the package build. An upper bound on `plist-xml` is also necessary. `markup` is not a system package.

mseri:
Thanks

PR #23378, opened 2023-02-20
[new release] OCADml (0.4.0)

geoffder:
Types and functions for building CAD packages in OCaml

- Project page: <a href="https://github.com/OCADml/OCADml">https://github.com/OCADml/OCADml</a>
- Documentation: <a href="https://OCADml.github.io/OCADml">https://OCADml.github.io/OCADml</a>

##### CHANGES:

- make cairo2 a depopt and break `PolyText` into optional sub-library
 (now free of non-OCaml dependencies)
- fix bug in `Path2.point_inside` (always returning outside)
- abstract `Poly{2,3}.t` (from private) free up for potential implementation change
- add missing `{of,to}_{seq,list,array}` to `Path` and `Poly` modules for
 convenience

PR #23396, opened 2023-02-23
Fix OCADml dependency vers in OSCADml

geoffder:
Fixes the rev-dep breakage causing the CI to fail on the new OCADml release (https://github.com/ocaml/opam-repository/pull/23378).


raphael-proust:
CI errors are system package installation issues on uncommon platforms (macos+homebrew). No other errors.

geoffder:
OCADml itself is building, but OSCADml fails to build with the latest version in the linked PR, so this change locks the dependency to avoid it.

packages/OSCADml/OSCADml.0.2.0/opam
```diff
@@ -17,7 +17,7 @@ depends: [
   "ocaml" {>= "4.14.0"}
   "gg" {>= "1.0.0"}
   "cairo2" {>= "0.6.2"}
-  "OCADml" {>= "0.3.0"}
+  "OCADml" {= "0.3.1" | = "0.3.2"}
```
mseri, line 20: Why not
```suggestion
  "OCADml" {>= "0.3.0" & < "0.3.3"}
```
It was working on 0.3.0 right? And this allows for possible bugfix backports (no matter how improbable)

geoffder: Updated with this suggestion, except with a lower bound of `0.3.1`, as `0.3.0` has been marked unavailable since my attempted hotfixing debacle.

mseri:

PR #23398, opened 2023-02-23
bump conf-pkg-config to 3: remove OpenBSD hack to install a port version

hannesm:
See https://github.com/ocaml/opam-repository/pull/20343 and https://github.com/mirage/mirage/issues/1262 for the background story:

Mirage 3.x used for "cross-compilation" a pkg-config universe (for opam packages that contained C libraries). Now, on OpenBSD the system pkg-config was not powerful enough, and we needed the pkgconfig from ports, and pkg-config < 3 installed a symbolic link to have the pkg-config binary linked to ports/pkgconfig (which install a binary called pkgconf) in PATH.

Since MirageOS 4 is out of the door since ~1 year, we can finally remove this hack and to enable forensics / archaeology we put upper bounds to conf-pkg-config on the relevant packages.

//cc @madroach who reported #20343 -- and of course everyone is welcome to provide feedback and test on OpenBSD machines.

### DIFF
--- a/packages/conf-pkg-config/conf-pkg-config.3/opam
+++ b/packages/conf-pkg-config/conf-pkg-config.3/opam
@@ -1,0 +1,32 @@
+opam-version: "2.0"
+maintainer: "unixjunkie@sdf.org"
+authors: ["Francois Berenger"]
+homepage: "http://www.freedesktop.org/wiki/Software/pkg-config/"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: "GPL-1.0-or-later"
+build: [
+  ["pkg-config" "--help"] {os != "openbsd"}
+]
+depexts: [
+  ["pkg-config"] {os-family = "debian"}
+  ["pkgconf"] {os-distribution = "arch"}
+  ["pkgconf-pkg-config"] {os-distribution = "fedora"}
+  ["pkgconfig"] {os-distribution = "centos" & os-version <= "7"}
+  ["pkgconf-pkg-config"] {os-distribution = "mageia"}
+  ["pkgconfig"] {os-distribution = "rhel" & os-version <= "7"}
+  ["pkgconfig"] {os-distribution = "ol" & os-version <= "7"}
+  ["pkgconf"] {os-distribution = "alpine"}
+  ["pkgconfig"] {os-distribution = "nixos"}
+  ["pkg-config"] {os = "macos" & os-distribution = "homebrew"}
+  ["pkgconfig"] {os = "macos" & os-distribution = "macports"}
+  ["pkgconf"] {os = "freebsd"}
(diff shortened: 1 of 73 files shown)


kit-ty-kate:
I’m a bit confused with the content of the PR. The title and description seems to imply you want to add a new version of `conf-pkg-config` but the PR doesn’t add it

kit-ty-kate:
Closing temporarily to avoid putting pressure in the CI, while the state of this PR clears-up. Feel free to reopen at any time.

hannesm:
indeed, sorry - I forgot to add the conf-pkg-config.3 package, which I now addedf.

hannesm:
There are lots of failures, I can see several paths forward (maybe @kit-ty-kate has an opinion):
- (1) Leave conf-pkg-config as is (with it depending on the OpenBSD port and installing a symlink on OpenBSD)
- (2) split this PR into smaller pieces that may pass CI
- (3) mark the solo5-bindings* and solo5-kernel* packages as `available: false` and only deal with other packages (solo5 is now a single package since 2021-12-31 -- supported on both mirage3 and mirage4), the bindings and kernel packages are only for historical purposes around.

kit-ty-kate:
> on OpenBSD the system pkg-config was not powerful enough, and we needed the pkgconfig from ports

what was the missing features on OpenBSD. Is that still the case? I have a personal OpenBSD machine on hand that I can test things with but I need those information to be able to make an informed decision for this PR.

Also don't worry about the failures, as far as I can see, none of the failures are relevant to this PR so they can be ignored for now.

hannesm:
commit 4cc57b1054c468985ec12e6c84e21a97c0a3db8c
Author: Martin Lucina <martin@lucina.net>
Date:   Fri May 18 17:41:28 2018 +0200

    conf-pkg-config: Install pkg-config symlink on OpenBSD
    
    The /usr/bin/pkg-config provided by OpenBSD does not support
    ${pcfiledir}, which breaks .pc files used by MirageOS. The OpenBSD
    devel/pkgconf package does support this feature, but does not install
    itself as "pkg-config" on the PATH. This change adds a symlink to the
    correct "pkg-config" in the in the OPAM "bin" directory at package
    installation time.
    
    We also inform the user, since this will change the "pkg-config" binary
    on their PATH when using OPAM..
    
    Additionally, we mark conf-pkg-config.1.0 as unavailble on OpenBSD to
    prevent installation of the buggy version.

hannesm:
It has been some time, I wonder what is missing to move this PR to being merged. As mentioned in the title, this makes opam & pkg-config less confusing on OpenBSD systems (by using the system's pkg-config instead of the ports one) -- and removes this workaround that was needed for mirage3.

madroach:
I'm an OpenBSD user and would be very pleased to get rid of that symlink!

mseri:
I think this is fine

PR #23401, opened 2023-02-23
ulex-camlp5 1.3: point to upstream tarball

mseri:
The content is identical (modulo the fix to the opam file which is included upstream)

### DIFF
--- a/packages/ulex-camlp5/ulex-camlp5.1.3/opam
+++ b/packages/ulex-camlp5/ulex-camlp5.1.3/opam
@@ -20,9 +20,9 @@ depends: [
 ]
 url {
   src:
-    "https://github.com/sacerdot/ulex/archive/refs/tags/v1.3-camlp5.tar.gz"
+    "https://github.com/whitequark/ulex/archive/refs/tags/v1.3.tar.gz"
   checksum: [
-    "md5=32033b89d244886d227801437f4b68fa"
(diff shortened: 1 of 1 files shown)


sacerdot:
Hmmm, good. I only have one remark: v1.1 and v1.2 have been released upstream for camlp4 only (master branch) and used by opam package ulex; v1.3 is now released upstream only for camlp5 (camlp5 branch) and used only by opam package ulex-camlp5. To clean up the mess @whitequark should release upstream on GitHub both v1.3 of the master branch and v1.3 of the camlp5 branch, to be used respectively by the ulex and ulex-camlp5 opam packages. Currently there is no opam package for ulex v1.3 (the one working with camlp4), but that may just reflect the fact that no users of ulex+camlp4+ocaml5 remain.

whitequark:
To everyone involved: should I maybe transfer ulex to ocaml-community organization? Sedlex already lives there.

avsm:
@whitequark that sounds great.  I'll send you an invite to ocaml-community so that you can do the transfer (I think that's all it takes)

whitequark:
@avsm I think I lost admin access to it though, e.g. I cannot re-enable issues which seems like it'd be OK now that it lives in -community.

avsm:
@whitequark admin access restored to you.

whitequark:
Thanks!

kit-ty-kate:
Thanks!

PR #23402, opened 2023-02-23
[new release] capnp-rpc, capnp-rpc-unix, capnp-rpc-net, capnp-rpc-mirage and capnp-rpc-lwt (1.2.3)

talex5:

##### CHANGES:

- Update to cmdliner 1.1.0 (@MisterDA mirage/capnp-rpc#249).

- Fix tests on Windows, avoid using Unix domain sockets (@talex5 mirage/capnp-rpc#251).

- Fix warning on `Unix.ENOTCONN` under macos (@talex5 mirage/capnp-rpc#252).

- Add `Capnp_rpc_unix.Cap_file.save_uri` (@talex5 mirage/capnp-rpc#253).

- Update README to use MDX (@talex5 mirage/capnp-rpc#254).

- Update tests to Cap'n Proto 0.10.3 (@MisterDA mirage/capnp-rpc#257 mirage/capnp-rpc#260).

- Update to mirage-crypto-rng 0.11 and dns 7.0.0 API changes (@hannesm mirage/capnp-rpc#261).

- Update for opam 2.1+ installations (@frumioj mirage/capnp-rpc#259).

- Install Cap'n Proto binaries in Windows GHA (@MisterDA mirage/capnp-rpc#248).

### DIFF
--- a/packages/capnp-rpc-unix/capnp-rpc-unix.1.2.3/opam
+++ b/packages/capnp-rpc-unix/capnp-rpc-unix.1.2.3/opam
@@ -28,11 +28,11 @@ depends: [
   "asetmap" {with-test}
 ]
 conflicts: [
-  "jbuilder" {with-test}
+  "jbuilder"
 ]
 build: [
   ["dune" "build" "-p" name "-j" jobs]
-  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test & os != "macos"}
 ]
 dev-repo: "git+https://github.com/mirage/capnp-rpc.git"
 url {
@@ -44,4 +44,3 @@ url {
   ]
 }
 x-commit-hash: "86427f68fa9a851fad6317cfda5e8b596add7fe9"
-x-ci-accept-failures: ["macos-homebrew"]
--- a/packages/capnp-rpc-unix/capnp-rpc-unix.1.2.3/opam
+++ b/packages/capnp-rpc-unix/capnp-rpc-unix.1.2.3/opam
@@ -1,0 +1,47 @@
+opam-version: "2.0"
+synopsis:
+  "Cap'n Proto is a capability-based RPC system with bindings for many languages"
+description:
+  "This package contains some helpers for use with traditional (non-Unikernel) operating systems."
+maintainer: "Thomas Leonard <talex5@gmail.com>"
+authors: "Thomas Leonard <talex5@gmail.com>"
+license: "Apache-2.0"
+homepage: "https://github.com/mirage/capnp-rpc"
+doc: "https://mirage.github.io/capnp-rpc/"
+bug-reports: "https://github.com/mirage/capnp-rpc/issues"
+depends: [
+  "ocaml" {>= "4.08.0"}
+  "capnp-rpc-net" {= version}
+  "cmdliner" {>= "1.1.0"}
+  "cstruct-lwt"
+  "astring"
+  "fmt" {>= "0.8.7"}
+  "logs"
+  "extunix"
+  "base64" {>= "3.0.0"}
+  "dune" {>= "3.0"}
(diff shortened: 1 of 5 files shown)


packages/capnp-rpc-unix/capnp-rpc-unix.1.2.3/opam
```diff
@@ -0,0 +1,47 @@
...
+url {
+  src:
+    "https://github.com/mirage/capnp-rpc/releases/download/v1.2.3/capnp-rpc-1.2.3.tbz"
+  checksum: [
+    "sha256=828002d67b9591d1645266c504e3fabc66b750229244a68b0a846c3c93f73715"
+    "sha512=c29f13ada74f3f8c80aa591f0fad60801ea72aea6aaa5299b2edee08e080061c5ac054392678ed8910962b1348f1e61790ce30febfc391ddb8c5ac01d56f3160"
+  ]
+}
+x-commit-hash: "86427f68fa9a851fad6317cfda5e8b596add7fe9"
+x-ci-accept-failures: ["macos-homebrew"]
```
kit-ty-kate, line 47: Only the tests fail so only the tests should be disabled. `x-ci-accept-failures` is not made for this
```suggestion
```

packages/capnp-rpc-unix/capnp-rpc-unix.1.2.3/opam
```diff
@@ -0,0 +1,47 @@
...
+  "mdx" {>= "2.2.1" & with-test}
+  "lwt" {>= "5.6.1"}
+  "asetmap" {with-test}
+]
+conflicts: [
+  "jbuilder" {with-test}
+]
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
```
kit-ty-kate, line 35:
```suggestion
  ["dune" "runtest" "-p" name "-j" jobs] {with-test & os != "macos"}
```

packages/capnp-rpc-unix/capnp-rpc-unix.1.2.3/opam
```diff
@@ -0,0 +1,47 @@
...
+  "dune" {>= "3.0"}
+  "alcotest" {>= "1.6.0" & with-test}
+  "alcotest-lwt" { >= "1.6.0" & with-test}
+  "mirage-crypto-rng-lwt" {>= "0.11.0"}
+  "mdx" {>= "2.2.1" & with-test}
+  "lwt" {>= "5.6.1"}
+  "asetmap" {with-test}
+]
+conflicts: [
+  "jbuilder" {with-test}
```
kit-ty-kate, line 31:
```suggestion
  "jbuilder"
```
`with-test` is not defined in `conflicts` so it reduces into `"jbuilder" {false}`. I'm not sure why it works for you on OCaml-CI. I'm guessing this is a bug in it.

MisterDA: Removing the conflict and setting the `re >= 1.9.0` lower bound should get us the same result.

kit-ty-kate: maybe but in this case I believe the conflict is the correct solution here as it's trying to avoid dune showing warnings that only happen when libraries it's using use jbuilder.

packages/capnp-rpc-unix/capnp-rpc-unix.1.2.3/opam
talex5: I guess. Ideally we only want to disable the tests in opam-repo-ci, not on macos in general.

kit-ty-kate: not really. As long as you run them using opam they will fail on macOS so constraining them in the opam file is the correct way to do it.

kit-ty-kate:
Thanks

PR #23409, opened 2023-02-24
release omake-0.10.6

gerdstolpmann:
new upstream release. omake is now compatible with OCaml-5.

### DIFF
--- a/packages/omake/omake.0.10.6/opam
+++ b/packages/omake/omake.0.10.6/opam
@@ -20,19 +20,13 @@ install: [
   [make "install"]
 ]
 
-remove: [
-  [ "rm" "-f" "%{prefix}%/bin/omake" ]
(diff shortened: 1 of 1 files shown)


packages/omake/omake.0.10.6/opam
```diff
@@ -0,0 +1,43 @@
...
+build: [
+  ["./configure" "-prefix" "%{prefix}%"]
+  [make]
+]
+
+install: [
+  [make "install"]
+]
+
+remove: [
```
mseri, line 23: This can be dropped now

mseri: (done in separate commit)

mseri:
Can you please check if the failres are general incompatibility with ocaml 5 (they involve bigarray of unix.a after all) or if they are related to omake?

kit-ty-kate:
Thanks

PR #23410, opened 2023-02-24
mirage 4.3.4: relax dune bound

hannesm:
We thought (for Mirage CI and chamelon-based unikernels) this to be a great idea, but unfortunately our mirage-www website unikernel fails (due to opam-monorepo deciding to use dune 3.7.0 and then selecting mirage 4.3.3).

The whole story is https://github.com/mirage/mirage-www/pull/795#issuecomment-1443564307 combined with https://github.com/mirage/mirage/pull/1401#issuecomment-1440762994

For the medium term, we already have another solution in mind.

### DIFF
--- a/packages/mirage/mirage.4.3.4/opam
+++ b/packages/mirage/mirage.4.3.4/opam
@@ -19,7 +19,7 @@ build: [
 
 depends: [
   "ocaml" {>= "4.08.0"}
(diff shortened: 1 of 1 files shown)


hannesm:
it would be great to get this merged sooner than later, since without this metadata change basically all mirageos unikernel compilations fail :/

kit-ty-kate:
Thanks

PR #23411, opened 2023-02-24
Release merlin and merlin-lib v4.8-{500,414}

voodoos:


voodoos:
Look like a package (`gopcaml`) breaks with this new release of Merlin. I added a constraint on it, but I am not sure that this is the way to go ? Should I add a "conflict" instead ? Hopefully this will be fixed upstream at some point (cc @Gopiandcode)

voodoos:
The error persists, so I guess the package was already broken ?

voodoos:
@kit-ty-kate makes sense :-)

kit-ty-kate:
Thanks!

PR #23244, opened 2023-02-06
[new release] colibrilib, colibrics and colibri2 (0.4)

bobot:
CHANGES:

  * Decision: Fix delayed decisions handling
  * Array: experimental theory
  * Quantifier: Improve eager instanciation heuristic
  * Quantifier: Avoid creating new term
  * Simplex: Fix redundant run
  * Fix compilation in 32bit
  * Fix sign of sqrt

PR #23414, opened 2023-02-25
[Dolmen] forbid ocaml 5.0.0

bobot:
@Gbury  cherry-pick from #23244 in order to streamline the CI

### DIFF
--- a/packages/dolmen/dolmen.0.4/opam
+++ b/packages/dolmen/dolmen.0.4/opam
@@ -6,7 +6,7 @@ build: [
   ["dune" "build" "-p" name "-j" "1"]
 ]
 depends: [
-  "ocaml" {>= "4.02.3"}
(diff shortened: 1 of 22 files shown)


avsm:
Do you intend to specifically forbid 5.0.0 or anything higher?  Presumably, older releases will also break with OCaml 5.0.1 and higher.

Gbury:
The intent is indeed to forbid `5.0.0` explicitly. This is because `dolmen` uses `Gc.finalise_last` which presents a bug in `5.0.0` (see https://github.com/ocaml/ocaml/issues/11995 ) where the finaliser is called too early (before the value is unreachable), which closes a file descriptor even though the file is still being read. This bug is already fixed in `trunk` (see https://github.com/ocaml/ocaml/pull/12001 ), so it should be fixed in any subsequent release, in particular in the `5.0.1`.

kit-ty-kate:
Thanks!

PR #23416, opened 2023-02-25
mlmpfr: make 4.1.1 version compatible with 4.2.0

thvnx:
The new 4.2.0 version of the C mpfr library is compatible with the previous 4.1.1 one, so, make the latter compatible with it in order to allow users to install the mlmpfr package on up to date systems.

thvnx/mlmpfr#28


mseri:
Succeeds where it can

thvnx:
Thanks!

PR #23433, opened 2023-03-02
Added new version 1.4.1 for 'gappa'

MSoegtropIMC:
This PR adds a new version to the numeric proof generator GAPPA.

Keeping the configure patch at opam level (rather than merging it upstream) has been agreed on with the author - the next version is expected to use a different configure system, which should not have this issue.

### DIFF
--- a/packages/gappa/gappa.1.4.1/opam
+++ b/packages/gappa/gappa.1.4.1/opam
@@ -1,0 +1,45 @@
+opam-version: "2.0"
+maintainer: "7895506+MSoegtropIMC@users.noreply.github.com"
+authors: "Guillaume Melquiond"
+bug-reports: "https://gitlab.inria.fr/gappa/gappa/-/issues"
+homepage: "https://gitlab.inria.fr/gappa/gappa"
+dev-repo: "git+https://gitlab.inria.fr/gappa/gappa.git"
+license: "CeCILL"
+patches: [
+  "0001-Added-configure-for-c-11.patch"
+]
+build: [
+  [ "autoreconf" ]
+  # Note: configure.in seems to reference this file
+  [ "touch" "stamp-config_h.in" ]
+  [ "./configure"
+    # If someone knows how to ask MacPorts for the usual include and lib path, please tell me
+    "CXXFLAGS=-I/opt/local/include" { os-distribution = "macports" & os = "macos" }
+    "LDFLAGS=-L/opt/local/lib" { os-distribution = "macports" & os = "macos" }
+    # Support installing on Apple Silicon with Homebrew
+    "CXXFLAGS=-I/opt/homebrew/include" { os-distribution = "homebrew" & os = "macos" & arch = "arm64"}
+    "LDFLAGS=-L/opt/homebrew/lib" { os-distribution = "homebrew" & os = "macos" & arch = "arm64"}
+    # Support for windows
+    "--build=%{arch}%-pc-cygwin" { os = "win32" & os-distribution = "cygwinports" }
+    "--host=%{arch}%-w64-mingw32" { os = "win32" & os-distribution = "cygwinports" }
+    "--target=%{arch}%-w64-mingw32" { os = "win32" & os-distribution = "cygwinports" }
+    "--prefix=%{prefix}%" 
+  ]
+  [ "./remake" "--jobs=%{jobs}%" ]
+]
+install: [ "./remake" "-d" "install" ]
+depends: [
+  "conf-g++" {build}
+  "conf-autoconf" {build}
+  "conf-automake" {build}
+  "conf-gmp"
+  "conf-mpfr"
+  "conf-boost"
+  "conf-bison" {build}
+  "conf-flex" {build}
+]
+synopsis: "Tool intended for formally proving properties on numerical programs dealing with floating-point or fixed-point arithmetic"
+url {
+  src: "https://gitlab.inria.fr/gappa/gappa/-/archive/gappa-1.4.1.tar.gz"
+  checksum: "sha512=c881eb35ad867320c2232c200dde66149fcd8a27972b8dfac1256a477a53e1885a7f89f82e70dbef553ce152e4afb0778bc347ed28cd4d698a1a69e308114aa2"
+}
--- a/packages/gappa/gappa.1.4.1/opam
+++ b/packages/gappa/gappa.1.4.1/opam
@@ -39,6 +39,7 @@ depends: [
   "conf-flex" {build}
 ]
 synopsis: "Tool intended for formally proving properties on numerical programs dealing with floating-point or fixed-point arithmetic"
+available: os-family != "alpine"
 url {
   src: "https://gitlab.inria.fr/gappa/gappa/-/archive/gappa-1.4.1.tar.gz"
   checksum: "sha512=c881eb35ad867320c2232c200dde66149fcd8a27972b8dfac1256a477a53e1885a7f89f82e70dbef553ce152e4afb0778bc347ed28cd4d698a1a69e308114aa2"
--- a/packages/gappa/gappa.1.4.1/opam
+++ b/packages/gappa/gappa.1.4.1/opam
@@ -4,7 +4,7 @@ authors: "Guillaume Melquiond"
 bug-reports: "https://gitlab.inria.fr/gappa/gappa/-/issues"
 homepage: "https://gitlab.inria.fr/gappa/gappa"
 dev-repo: "git+https://gitlab.inria.fr/gappa/gappa.git"
-license: "CeCILL"
+license: "CECILL-2.1"
 patches: [
   "0001-Added-configure-for-c-11.patch"
 ]
--- a/packages/gappa/gappa.1.4.0/opam
+++ b/packages/gappa/gappa.1.4.0/opam
@@ -4,7 +4,7 @@ authors: "Guillaume Melquiond"
 bug-reports: "https://gitlab.inria.fr/gappa/gappa/-/issues"
 homepage: "https://gitlab.inria.fr/gappa/gappa"
 dev-repo: "git+https://gitlab.inria.fr/gappa/gappa.git"
-license: "CeCILL"
+license: "CECILL-2.1"
 patches: [
   "0001-Added-configure-for-c-11.patch"
 ]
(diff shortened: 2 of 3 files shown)


packages/gappa/gappa.1.4.1/opam
```diff
@@ -0,0 +1,45 @@
...
+    # Support installing on Apple Silicon with Homebrew
+    "CXXFLAGS=-I/opt/homebrew/include" { os-distribution = "homebrew" & os = "macos" & arch = "arm64"}
+    "LDFLAGS=-L/opt/homebrew/lib" { os-distribution = "homebrew" & os = "macos" & arch = "arm64"}
+    # Support for windows
+    "--build=%{arch}%-pc-cygwin" { os = "win32" & os-distribution = "cygwinports" }
+    "--host=%{arch}%-w64-mingw32" { os = "win32" & os-distribution = "cygwinports" }
+    "--target=%{arch}%-w64-mingw32" { os = "win32" & os-distribution = "cygwinports" }
+    "--prefix=%{prefix}%" 
+  ]
+  [ "./remake" "--jobs=%{jobs}%" ]
```
kit-ty-kate, line 28: This custom build system doesn’t work on Alpine (tmpnam doesn’t take the `TMPDIR` env variable into account and thus fail with the sandbox) and won’t work on macOS either given any network (even local) access is forbidden by the macOS sandbox (builtin macOS sandbox restriction).

Is there any chance you could be using something a little less custom to build your package?

MSoegtropIMC: It definitely does work on MacOS with MacPorts and Homebrew - I am using both myself.

With "custom build system" you refer to `remake` - it is not that uncommon. If there are issues with it, I guess I can have them fixed.

MSoegtropIMC: I can look into what issues conf-bison and conf-flex have on Homebrew MacOS, but it seems to be a CI artefact. It works fine in GitHub based MacOS CI and it also works fine on my private machine (I have a x86 Homebrew and a ARM Macports machine). I will check on my private Homebrew machine if the installation (not just checking) of bison and flex works.

MSoegtropIMC: @kit-ty-kate : I had a look at Alpine. As far as I understand it uses this implementation for `tempname`:

https://git.musl-libc.org/cgit/musl/tree/src/stdio/tempnam.c

which indeed does not check any environment variables. Doesn't this lead to many other issues with opam?

I can ask the author of remake if they want to work around this. Otherwise I would state that a light wight OS like Alpine is not the target audience for proof generators.

kit-ty-kate: > It definitely does work on MacOS with MacPorts and Homebrew - I am using both myself.
> 
> With "custom build system" you refer to `remake` - it is not that uncommon. If there are issues with it, I guess I can have them fixed.

via opam? I'm surprised

avsm:
I think part of the reason for the unexpected package building issues in opam is that ... gappa is not an OCaml program.

I realise there are previous versions of gappa in opam-repo, and I'm not suggesting rejecting this package, but... what's it doing here?  If it's a Coq dependency, wouldn't it be better off being distributed using native OS package managers that handle autotools and C/C++ compilation as their bread and butter?

packages/gappa/gappa.1.4.1/opam
MSoegtropIMC: Yes, via opam. Coq Platform is completely opam based and I have gappa in Coq Platform CI since years, which tests on MacOS (homebrew), Ubuntu and Windows. There was never any problem with this package.

Do you have a pointer to the documentation of what should be writeable?

MSoegtropIMC:
@avsm: I originally put this package here and asked back then if you want it here or not and someone decided you want it here. I have absolutely no issue moving it to the Coq opam repo, but it is is not Coq specific. GAPPA can e.g. also produce proofs for HOL and HOL-light is written in OCaml (like Coq).

It is a similar question on why we do have an opam package for Z3. In the end it is a package which is useful for some OCaml developments, so it is convenient to be able to install it via opam. I don't think that there are gappa packages in common operating systems - it is way too exotic. And honestly I would say in terms of compatibility it is not that bad. Alpine is more an embedded OS and not the target audience for proof assistants, so I would simply exclude it in opam.

I think one can also claim that it is a bug in Opam if the hard wired temp directory of the system (in Alpine it appears to be hard wired) is not writeable by OCaml.

MSoegtropIMC:
P.S.: IMHO opam is a terribly convenient meta build system for just anything.

MSoegtropIMC:
@kit-ty-kate : I had a discussion with the author of remake. He says that opam documents that /tmp is writeable here:

https://opam.ocaml.org/doc/FAQ.html#Why-does-opam-require-bwrap

(2nd paragraph). So I don't quite understand what is going on.

How about simply disabling gappa on alpine? As I said, I don't think many people run interactive proof assistants on Alpine.

avsm:
Thanks for the additional background @MSoegtropIMC about Gappa.  I have no problem putting solvers that are useful and actively maintained by the verification community in opam-repository, but we do need to also recognise that build tools off the beaten path do take up a disproportionate amount of maintainer time here.  So please don't take the questioning here as being critical, but us simply trying to figure out how whether we should treat C/C++ tools as a more first class citizen in the opam repo (for example, by support packages that can do environment probing and CFLAGS/LDFLAGS setting more systematically across z3/gappa/gmp and other packages).

Regarding the specific issue of /tmp, the FHS and other standards do mandate that `/tmp` exists, but not how big it might be.  In practical terms, the reason to use `$TMPDIR` is that this is how operating systems would assign per-user temp directories with reasonable space accounting, and this is still a useful property to have.  It's not uncommon to have very small (or memory-backed) `/tmp` directories in build servers or smaller devices like Raspberry Pis.

I would very much like to keep Alpine support if possible, please. It is an extremely popular distribution for those using Docker or embedded Linux, and if our dependency graphs start excluding it without good reason then it makes containerisation hard.  It doesn't sound like there's a fundamental issue with Alpine -- just this /tmp question -- and so we should figure out how to sort that out (either by fixing the opam sandbox to permit /tmp, or by modifying remake to support TMPDIR as a courteous thing to do on UNIX-like systems, or both).

MSoegtropIMC:
@avsm: fine with me.

I don't think one can say remake does anything wrong - it uses the OS supplied function to create a temporary file and this fails. I think it is unreasonable to ask the remake authors to work around this. Since in Alpine the temp folder is hard wired to /tmp I would say the only reasonable fix is to allow write access to /tmp in Alpine in the sandbox.

avsm:
I'm not saying remake is doing anything wrong: I'm saying it's just not being very polite about not using the environment variable we (and the [SUS](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08_03) and many other specifications) recommend.  There are plenty of OS-supplied functions that are not portable, and require workarounds for. At a guess, musl (the libc in Alpine) isn't directly checking for the environment variable since it's not threadsafe to query environment variables, and the C function can't know the context it's called in.  So the onus is on the caller to do so, and in this case, we know that remake can trivially do so as it's sequentially executed (and indeed, is vendored in your tool's build system, so it should be even easier to patch). 

I've created https://github.com/ocaml/opam/issues/5462 to record the need for a writeable `/tmp` within the opam sandbox. It'll take some time for us to get that through a release cycle though, and a workaround from gappa would be appreciated to keep it working on Alpine.  If you really don't want to, then we'll mark it (unhappily) as broken on Alpine.

MSoegtropIMC:
@avsm : still I don' think it makes sense to check the environment variable if the C library doesn't. I think one can reasonably assume that basic system / C library functions work without any workaround. The author of remake is in the business of writing formally verified code and likely has strong opinions on what is the proper way of doing things.

avsm:
> I think one can reasonably assume that basic system / C library functions work without any workaround.
> The author of remake is in the business of writing formally verified code.

See my [SibylFS paper](https://dl.acm.org/doi/10.1145/2815400.2815411) from SOSP 2013 on why those two statements do not support each other (specifically Section 7.3 on the vagaries of the POSIX specification), and the OCaml runtime itself (full of essential workarounds for allegedly basic system functions).  I don't know the author of remake on whose behalf you are speaking for, but they'd be well advised to not trust in the portability of C library functions too much.

I don't have the time to debate this further with respect to the opam-repository. If the other maintainers are ok with it, I'm also happy for this package to be disabled on Alpine; it's the coq-platform's decision as to whether or not they support this OS.  It'll have a reduced testing cone from us as a result.

MSoegtropIMC:
@avsm: since Gappa did never work on Alpine, I would say the most reasonable solution is to disable it now and re-enable it when (https://github.com/ocaml/opam/issues/5462) is implemented. I don't see value in creating some sort of work around in remake if a better solution is on the way. I might reconsider this if there are actual complaints from users.

Thanks for the pointer to your paper - interestingly I did some years back something similar (but much simpler - tailored to test a specific file system to its specs). I am not sure if I follow your conclusion, though. If a system does not adhere to clear Posix specifications it should be fixed and not worked around. That there is system dependent behaviour which is not covered by Posix is an entirely different story. I fully agree that one should not rely on such behaviour. But the spec that /tmp shall be writeable is a clear Posix spec. IMHO things will fall apart if we pile up work arounds instead of fixing root causes.

MSoegtropIMC:
P.S.: I will change my PR to disable Alpine.

packages/gappa/gappa.1.4.1/opam
```diff
@@ -0,0 +1,45 @@
...
+  "conf-g++" {build}
+  "conf-autoconf" {build}
+  "conf-automake" {build}
+  "conf-gmp"
+  "conf-mpfr"
+  "conf-boost"
+  "conf-bison" {build}
+  "conf-flex" {build}
+]
+synopsis: "Tool intended for formally proving properties on numerical programs dealing with floating-point or fixed-point arithmetic"
```
mseri, line 41: available: os-family != "alpine"

mseri:
I marked it unavailable on alpine. Thanks

MSoegtropIMC:
@mseri : thanks for the fix and the merge!

PR #23429, opened 2023-03-01
[new release] mrmime (0.6.0)

dinosaure:
Mr. MIME

- Project page: <a href="https://github.com/mirage/mrmime">https://github.com/mirage/mrmime</a>
- Documentation: <a href="https://mirage.github.io/mrmime/">https://mirage.github.io/mrmime/</a>

##### CHANGES:

- Be able to set field parsers when we parse an entire email (@dinosaure, mirage/mrmime#89)
- Be able to wrap a multipart email into a part (@dinosaure, mirage/mrmime#90)
- Don't use RNG to produce the example to be compatible with OCaml 5 (@dinosaure, mirage/mrmime#93)
- Upgrade to cmdliner.1.1.0 (@hannesm, mirage/mrmime#94)
- Use a smaller email to test isomorphism (@dinosaure, 931c6ad)
- Lint dependencies (@dinosaure, mirage/mrmime#95)
- Delete `rresult` dependency (@dinosaure, @hannesm, mirage/mrmime#96)
- Delete `fmt` dependency (@dinosaure, mirage/mrmime#97)

PR #23438, opened 2023-03-02
mark mrmime.0.6.0 revdeps breakages

raphael-proust:
https://github.com/ocaml/opam-repository/pull/23429 releases 0.6.0 which breaks some revdeps relying on transitive dependencies.


dinosaure:
Thanks, I was going to do it but I hadn't found the time yet!

PR #23441, opened 2023-03-02
Fix revdeps from lablgtk PR

mseri:
This should account for all fixable failures.
All seen on https://github.com/ocaml/opam-repository/pull/23384

### DIFF
--- a/packages/rfsm/rfsm.1.0/opam
+++ b/packages/rfsm/rfsm.1.0/opam
@@ -26,6 +26,7 @@ depends: [
   "menhir" {< "20211215"}
   "lascar" {< "0.6.0"}
 ]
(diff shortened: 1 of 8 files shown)


mseri:
Not going to fix the lower bounds here

packages/rfsm/rfsm.1.0/opam
```diff
@@ -26,6 +26,7 @@ depends: [
   "menhir" {< "20211215"}
   "lascar" {< "0.6.0"}
 ]
+available: opam:version < "2.0.9"
```
kit-ty-kate, line 29:
```suggestion
available: opam-version < "2.0.9"
```

PR #23453, opened 2023-03-04
Package errpy.0.0.7

jasontatton:
### `errpy.0.0.7`
Errpy: An Error Recovering Rust Python Parser

---

---


jasontatton:
Looks like I just need to downgrade the version of rust to 2018 and fix up the lint errors

kit-ty-kate:
You seem to be using cargo without vendoring. Vendoring all your dependencies in the archive is required to be able to build this package in the opam sandbox which forbids network calls. You can have a look at other packages doing this such as `tezos-rust-libs`
```
#=== ERROR while compiling errpy.0.0.7 ========================================#
# context              2.2.0~alpha~dev | linux/x86_64 | ocaml-base-compiler.5.0.0 | pinned(https://github.com/facebook/errpy/archive/refs/tags/0.0.7.tar.gz)
# path                 ~/.opam/5.0/.opam-switch/build/errpy.0.0.7
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p errpy -j 255 @install
# exit-code            1
# env-file             ~/.opam/log/errpy-7-7cfae2.env
# output-file          ~/.opam/log/errpy-7-7cfae2.out
### output ###
# File "src/dune", line 13, characters 0-182:
# 13 | (rule
# 14 |  (targets libffi_ocaml.a)
# 15 |  (mode fallback)
# 16 |  (locks /cargo)
# 17 |  (action
# 18 |   (progn
# 19 |    (run cargo build --release)
# 20 |    (run cp ../../../target/release/libffi_ocaml.a libffi_ocaml.a))))
# (cd _build/default/src && /usr/sbin/cargo build --release)
# error: failed to get `bstr` as a dependency of package `ffi_ocaml v0.0.0 (/home/opam/.opam/5.0/.opam-switch/build/errpy.0.0.7)`
# 
# Caused by:
#   failed to load source for dependency `bstr`
```

jasontatton:
Thanks for this @kit-ty-kate do you know of a way I can test for this offline without having to raise a PR via the `opam publish`?

kit-ty-kate:
`opam install ./` or `opam pin add ./` will do that

jasontatton:
Thanks @kit-ty-kate I will run opam install./ from now on previously I was just using dune build @install. 

One thing which might be a blocker however, is that many of the dependencies of this project require rust 2021. I'm ok downgrading the main errpy code to 2018 but these dependencies are not in our control and I doubt the maintainers would be interested in a downgrade.

Is there any possibility of upgrading opam to make use of rust 2021?

jasontatton:
Hi @kit-ty-kate Thanks for your help previously. We've fixed the two previously identified issues of: 1). using `conf-rust-2021`  and 2). vendoring in all of the rust dependencies to `vendor_external` . But it seems that the `oUnit` dependency is missing, does this need to be explicitly defined (I thought oUnit was included by default)?

jasontatton:
We've upgraded ERRPY to use dependencies which are OCaml 5.0 compatible and included the ounit dependency, these seemed to be the major problems reported here. We will raise this as a new PR

PR #23456, opened 2023-03-04
Package srt.0.3.0

toots:
### `srt.0.3.0`
Binding for the Secure, Reliable, Transport protocol library
Secure Reliable Transport (SRT) is an open source transport technology
that optimizes streaming performance across unpredictable networks, such
as the Internet.
This package provides OCaml bindings to the C implementation library.

---
* Homepage: https://github.com/savonet/ocaml-srt
* Source repo: git+https://github.com/savonet/ocaml-srt.git
* Bug tracker: https://github.com/savonet/ocaml-srt/issues

---


mseri:
The failure to build `conf-srt-openssl` in debian (and thus this package) is expected?

packages/srt/srt.0.3.0/opam
```diff
@@ -0,0 +1,47 @@
...
+synopsis: "Binding for the Secure, Reliable, Transport protocol library"
+description: """
+Secure Reliable Transport (SRT) is an open source transport technology
+that optimizes streaming performance across unpredictable networks, such
+as the Internet.
+This package provides OCaml bindings to the C implementation library.
+"""
+maintainer: ["The Savonet Team <savonet-users@lists.sourceforge.net>"]
+authors: ["The Savonet Team <savonet-users@lists.sourceforge.net>"]
+license: "GPL-2.0"
```
mseri, line 12:
```suggestion
license: "GPL-2.0-only"
```
or
```suggestion
license: "GPL-2.0-or-later"
```
?

toots:
> The failure to build `conf-srt-openssl` in debian (and thus this package) is expected?

Not it's not but it's also most likely a ubuntu bug. I suspect `libsrt-dev` does not install the corresponding SSL dev package. This package could be from gnutls or openssl so it should really be `libsrt-dev`'s responsibility to install the one it's been compiled with.

toots:
License clarification backported upstream thanks!

packages/srt/srt.0.3.0/opam
```diff
@@ -0,0 +1,47 @@
...
+synopsis: "Binding for the Secure, Reliable, Transport protocol library"
+description: """
+Secure Reliable Transport (SRT) is an open source transport technology
+that optimizes streaming performance across unpredictable networks, such
+as the Internet.
+This package provides OCaml bindings to the C implementation library.
+"""
+maintainer: ["The Savonet Team <savonet-users@lists.sourceforge.net>"]
+authors: ["The Savonet Team <savonet-users@lists.sourceforge.net>"]
+license: "GPL-2.0-only"
```
kit-ty-kate, line 12: Note that according to your files (e.g. https://github.com/savonet/ocaml-srt/blob/a10d0785c188236c3c11e70d53f4ce9a021f09e1/src/srt.mli#L9) this should be `GPL-2.0-or-later` instead

kit-ty-kate:
Thanks

toots:
License fixed again. Thanks for the attention to details y'all!

PR #22281, opened 2022-10-15
[new release] geojsone and geojson (0.1.0)

patricoferris:
Streaming GeoJSON library

- Project page: <a href="https://github.com/geocaml/ocaml-geojson">https://github.com/geocaml/ocaml-geojson</a>

##### CHANGES:

Initial release

PR #23458, opened 2023-03-05
[new release] topojsone and topojson (0.1.0)

patricoferris:
Streaming TopoJSON library

- Project page: <a href="https://github.com/geocaml/ocaml-topojson">https://github.com/geocaml/ocaml-topojson</a>

##### CHANGES:

Initial release

### DIFF
--- a/packages/topojson/topojson.0.1.0/opam
+++ b/packages/topojson/topojson.0.1.0/opam
@@ -1,0 +1,43 @@
+opam-version: "2.0"
+synopsis: "A pure OCaml library for working with the TopoJSON format"
+description:
+  "This library provides a pure OCaml library for working with TopoJSON."
+maintainer: ["patrick@sirref.org"]
+authors: ["Patrick Ferris" "Jay Dev Jha"]
+license: "MIT"
+homepage: "https://github.com/geocaml/ocaml-topojson"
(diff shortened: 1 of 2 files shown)


patricoferris:
*sigh* I did the same thing as I did in https://github.com/ocaml/opam-repository/pull/22281 sorry, I wonder if ocaml-ci can catch these kinds of things. I'll have to re-release the package. Sorry for the noise!

avsm:
What's the error you're referring to? A missing `(package)` to a tests stanza?

patricoferris:
Sorry, I should have been more informative! Yes that's right, I suppose it's less about that specific failure case and more related to the fact that OCaml CI does an unconventional installation in order to support the features that it does:

 - Only runs tests on the intersection of the supported platforms for _all_ of the packages in a given repo hence https://github.com/geocaml/ocaml-topojson only gets tested on OCaml 5 and I never catch the fact that it doesn't work on `< 4.08`. I think this is being tracked: https://github.com/ocurrent/ocaml-ci/issues/470
 - And yes, the problem you specify, which I think is related to ocaml-ci not treating a multi-package repository as a multi-single-package repo if that makes sense.

avsm:
Second problem is probably tracked by https://github.com/ocurrent/ocaml-ci/issues/257

PR #23473, opened 2023-03-08
Release of esperanto.0.0.3

dinosaure:
- Delete the `pth` relicat and directly use `pthread` from Cosmopolitan (dinosaure/esperanto#33, @dinosaure)
- Delete the `-mnop-mcount` and fix the issue with `dune.3.7.0` (dinosaure/esperanto#34, @dinosaure) The `-mnop-mcount` is important only if we use `-pg` which is not the case. The initial issue is a conflict with this option and `-fPIC` introduced by `dune.3.7.0`.
- Upgrade Esperanto with Cosmopolitan 2.2-96-gae365928c (dinosaure/esperanto#36, @dinosaure) This specific version of Cosmopolitan fixes an issue about Esperanto and the LWT support. However, this specific version of Cosmopolitan loses the support of `speed_t` (all values of this type disappeared).


kit-ty-kate:
Thanks

PR #23480, opened 2023-03-09
Fix ocamlfind.1.9.6 upgrade

dra27:
Upgrading from ocamlfind 1.9.5 to 1.9.6 is failing on OCaml 5 switches. If you run:

```
opam switch create 5.0 ocaml-base-compiler.5.0.0
opam install ocamlfind.1.9.5
opam install ocamlfind.1.9.6
```

and then run `ocamlfind list` then none of the compiler's packages are included. This is fixed in https://github.com/ocaml/ocamlfind/pull/61, but as the fix only affects the `configure` file, it seems reasonable to patch it immediately in opam-repository.

Note that there is a workaround - `opam reinstall ocamlfind.1.9.6` permanently fixes the problem.


packages/ocamlfind/ocamlfind.1.9.6-1/opam
```diff
@@ -0,0 +1,46 @@
...
+  [make "all"]
+  [make "opt"] {ocaml:native}
+]
+install: [
+  [make "install"]
+  ["install" "-m" "0755" "ocaml-stub" "%{bin}%/ocaml"] {ocaml:preinstalled}
+]
+dev-repo: "git+https://github.com/ocaml/ocamlfind.git"
+url {
+  src: "http://download.camlcity.org/download/findlib-1.9.6.tar.gz"
```
mseri, line 41: The patch is missing

packages/ocamlfind/ocamlfind.1.9.6/opam
```diff
@@ -13,7 +13,7 @@ authors: "Gerd Stolpmann <gerd@gerd-stolpmann.de>"
 homepage: "http://projects.camlcity.org/projects/findlib.html"
 bug-reports: "https://github.com/ocaml/ocamlfind/issues"
 depends: [
-  "ocaml" {>= "4.00.0"}
+  "ocaml" {>= "3.08.0"}
```
mseri, line 16:
```suggestion
  "ocaml" {>= "3.08.0" & < "5.0"}
```
Should we make this unavailable on ocaml 5?

dra27:
All hopefully done. I also added a disabling commit for Windows (I'm separately looking into what's wrong there, but that will almost certainly need a new release)

packages/ocamlfind/ocamlfind.1.9.6-1/opam
```diff
@@ -0,0 +1,46 @@
...
+authors: "Gerd Stolpmann <gerd@gerd-stolpmann.de>"
+homepage: "http://projects.camlcity.org/projects/findlib.html"
+bug-reports: "https://github.com/ocaml/ocamlfind/issues"
+depends: [
+  "ocaml" {>= "3.08.0"}
+]
+depopts: ["graphics"]
+build: [
+  # See https://github.com/ocaml/ocamlfind/pull/61
+  ["sed" "-ib" "s|stdlib/META|stdlib.cmi|" "configure"]
```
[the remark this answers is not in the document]
dra27: Changed to a patch - `sed` is used in a few (older) compiler packages, but I agree that it's bad to have a _latest_ version doing this!

kit-ty-kate:
Thanks a lot

ejgallego:
Hi @dra27 , I'm having some ocamfind windows issues on Opam 2.2, with 1.9.5 ; I see 1.9.6 is still disabled.

Do you have a pointer to the issues you found, to see if there would be a duplicate?

PR #23497, opened 2023-03-12
Release Conjury 2.2.

jhwoodyatt:
Compatibility with OCaml 5.

### DIFF
--- a/packages/conjury/conjury.2.2/opam
+++ b/packages/conjury/conjury.2.2/opam
@@ -7,14 +7,14 @@ tags: "org:conjury.org"
 homepage: "https://bitbucket.org/jhw/conjury/"
 bug-reports: "https://bitbucket.org/jhw/conjury/issues"
 depends: [
-  "ocaml" {with-test & ocaml:native & >= "4.08"}
-  "ocamlfind" {with-test & ocaml:native & >= "1.7.3"}
+  "ocaml" {with-test & >= "4.08"}
+  "ocamlfind" {with-test & >= "1.7.3"}
   "omake" {>= "0.10.3"}
   "ounit2" {with-test & >= "2.2"}
 ]
 build: [
   ["omake" "--dotomake.omake" "--force-dotomake" "--configure" "build"]
-  ["omake" "--dotomake.omake" "--force-dotomake" "test"] {with-test & native}
+  ["omake" "--dotomake.omake" "--force-dotomake" "test"] {with-test & ocaml:native}
 ]
 install: ["omake" "--dotomake.omake" "--force-dotomake" "install"]
 dev-repo: "git+https://bitbucket.org/jhw/conjury"
--- a/packages/conjury/conjury.2.2/opam
+++ b/packages/conjury/conjury.2.2/opam
@@ -7,10 +7,10 @@ tags: "org:conjury.org"
 homepage: "https://bitbucket.org/jhw/conjury/"
 bug-reports: "https://bitbucket.org/jhw/conjury/issues"
 depends: [
-  "ocaml" {with-test & native >= "4.08"}
-  "ocamlfind" {with-test & native & >= "1.7.3"}
+  "ocaml" {with-test & ocaml:native & >= "4.08"}
+  "ocamlfind" {with-test & ocaml:native & >= "1.7.3"}
   "omake" {>= "0.10.3"}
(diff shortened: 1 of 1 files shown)


packages/conjury/conjury.2.2/opam
```diff
@@ -0,0 +1,24 @@
...
+maintainer: "james woodyatt <jhw@conjury.org>"
+authors: "james woodyatt <jhw@conjury.org>"
+license: "BSD-2-Clause"
+tags: "org:conjury.org"
+homepage: "https://bitbucket.org/jhw/conjury/"
+bug-reports: "https://bitbucket.org/jhw/conjury/issues"
+depends: [
+  "ocaml" {with-test & native >= "4.08"}
+  "ocamlfind" {with-test & native & >= "1.7.3"}
+  "omake" {>= "0.10.3"}
```
raphael-proust, line 12: Should `omake` be marked as a build-only dependency? (I'm not sure because most build tools should but some (e.g., `dune`) should not.)

raphael-proust:
I was confused by the `with-test` in the ocaml dependency but I had a look at the project and it makes sense. (Also the non-build dependency to `omake` and all that.)

packages/conjury/conjury.2.2/opam
```diff
@@ -0,0 +1,24 @@
...
+synopsis: "Conjury library for OMake"
+maintainer: "james woodyatt <jhw@conjury.org>"
+authors: "james woodyatt <jhw@conjury.org>"
+license: "BSD-2-Clause"
+tags: "org:conjury.org"
+homepage: "https://bitbucket.org/jhw/conjury/"
+bug-reports: "https://bitbucket.org/jhw/conjury/issues"
+depends: [
+  "ocaml" {with-test & native >= "4.08"}
+  "ocamlfind" {with-test & native & >= "1.7.3"}
```
kit-ty-kate, line 11:
```suggestion
  "ocaml" {with-test & ocaml:native & >= "4.08"}
  "ocamlfind" {with-test & ocaml:native & >= "1.7.3"}
```

packages/conjury/conjury.2.2/opam
```diff
@@ -0,0 +1,24 @@
...
+  "omake" {>= "0.10.3"}
+  "ounit2" {with-test >= "2.2"}
```
kit-ty-kate, line 13:
```suggestion
  "ounit2" {with-test & >= "2.2"}
```

jhwoodyatt:
The `opam lint` failures are happening because I tried to use package variables in the dependency formulas to rule out configurations where the native OCaml compilers are not available. If there's a better way to do that, then I'm open to suggestions. Otherwise, I probably need to figure out how to make the build and install logic work without requiring the native compiler in this release rather than a forthcoming one.

packages/conjury/conjury.2.2/opam
```diff
@@ -0,0 +1,24 @@
...
+synopsis: "Conjury library for OMake"
+maintainer: "james woodyatt <jhw@conjury.org>"
+authors: "james woodyatt <jhw@conjury.org>"
+license: "BSD-2-Clause"
+tags: "org:conjury.org"
+homepage: "https://bitbucket.org/jhw/conjury/"
+bug-reports: "https://bitbucket.org/jhw/conjury/issues"
+depends: [
+  "ocaml" {with-test & ocaml:native & >= "4.08"}
+  "ocamlfind" {with-test & ocaml:native & >= "1.7.3"}
```
kit-ty-kate, line 11:
```suggestion
  "ocaml" {with-test & >= "4.08"}
  "ocamlfind" {with-test & >= "1.7.3"}
```

packages/conjury/conjury.2.2/opam
```diff
@@ -0,0 +1,24 @@
...
+  "omake" {>= "0.10.3"}
+  "ounit2" {with-test & >= "2.2"}
+]
+build: [
+  ["omake" "--dotomake.omake" "--force-dotomake" "--configure" "build"]
+  ["omake" "--dotomake.omake" "--force-dotomake" "test"] {with-test & native}
```
kit-ty-kate, line 17:
```suggestion
  ["omake" "--dotomake.omake" "--force-dotomake" "test"] {with-test & ocaml:native}
```

PR #23502, opened 2023-03-13
[new release] lambdapi (2.3.1)

fblanqui:

### Fixed

- Opaque definitions are not kept in memory and in lpo files anymore
- A few bug fixes.

### Changed

- Why3 dependency updated to 1.6.

### DIFF
--- a/packages/lambdapi/lambdapi.2.3.1/opam
+++ b/packages/lambdapi/lambdapi.2.3.1/opam
@@ -1,0 +1,87 @@
+opam-version: "2.0"
+synopsis: "Proof assistant for the λΠ-calculus modulo rewriting"
+description: """
+This package provides:
+- A lambdapi command for checking .lp or .dk files,
+translating .dk files to .lp files and vice versa,
+or launching an LSP server for editing .lp files.
+- A library of logic definitions and basic definitions and proofs
+on natural numbers and polymorphic lists.
+- A rich Emacs mode based on LSP (available on MELPA too).
+- A basic mode for Vim.
+- OCaml libraries.
+A VSCode extension is also available on the VSCode Marketplace.
+
+Find Lambdapi user manual on https://lambdapi.readthedocs.io/.
+
+Lambdapi provides a rich type system with dependent types.
+In Lambdapi, one can define both type and function symbols
+by using rewriting rules (oriented equations).
+A symbol can be declared associative and commutative.
+Lambdapi supports unicode symbols and infix operators.
+The declaration of symbols and rewriting rules is separated
+so that one can easily define inductive-recursive types.
+
+Lambdapi checks that rules are locally confluent (by checking
+the joinability of critical pairs) and preserve typing.
+Rewrite rules can also be exported to the TRS and XTC formats
+for checking confluence and termination with external tools.
+
+Lambdapi does not come with a pre-defined logic. It is a
+powerful logical framework in which one can easily define
+its own logic and build and check proofs in this logic.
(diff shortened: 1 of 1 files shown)


mseri:
This version is failing on macos and ppc64: https://opam-ci.ci3.ocamllabs.io/github/ocaml/opam-repository/commit/9271226fcba9aa04723ffb2534424501d1eb5405

Do you know what is going on?

fblanqui:
@gabrielhdt @TheoWinterhalter do you have any idea if you use a Mac?

TheoWinterhalter:
Do I understand correctly that it's complaining that a file may not be found? If so it might a problem related to case sensitivity? Like maybe you renamed `foo.dk` to `Foo.dk`? This kind of stuff can be a cause for a lot of problems on macOS.

TheoWinterhalter:
Ah yes, in tests you have both `bool.lp` and `Bool.lp`. Same for `nat.lp` and `Nat.lp`. I suspect this might be the cause.

fblanqui:
Ah, thank you. I didn't know about this. But it didn't show up in previous versions... @mseri Something has been changed on opam CI ?

fblanqui:
Hi. I modified the package by changing the names of some test files. I solved all issues but one. I don't understand the last error. It doesn't do test ` 78   OK/rule_order.lp`. I don't understand why, and there is no error message.
https://opam-ci.ci3.ocamllabs.io/github/ocaml/opam-repository/commit/679a2d941fb40f78f20fc911079e72f35a02fa7c/variant/extras,ppc64-ocaml-4.14,lambdapi.2.3.1,tests

fblanqui:
@mseri I don't know what else to do. Is it mandatory to pass this test to merge this PR?

TheoWinterhalter:
There is no error message, but there is something that looks like a prefix of one:
```
# File "tests/dune", line 2, characters 8-13:
# 2 |  (names ok_ko rewriting purity kernel)
#             ^^^^^
```
I don't know what this means however. Also this time it seems to only affect PowerPC, so I have no idea about it myself.

fblanqui:
Hi @mseri . I have no way to know what is the problem and fix it. Is it mandatory to pass the test on PowerPC to merge this PR? Please let me know.

mseri:
Not a problem. Do you prefer for me to disable the tests or make the architecture unavailable? The kind of failure makes me think that it may actually segfault there, so I am more prone to the latter.

fblanqui:
Ok let's do that (make arch unavailable). Thank you!

packages/lambdapi/lambdapi.2.3.1/opam
```diff
@@ -0,0 +1,87 @@
...
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
+url {
```
mseri, line 79:
```suggestion
available: arch != "ppc64"
url {
```

fblanqui:
@mseri Do you know what I should change in my dune-project file if I want my opam file to be generated by dune automatically?

kit-ty-kate:
> Do you know what I should change in my dune-project file if I want my opam file to be generated by dune automatically?

I don't think this is supported by dune. I would discourage using dune autogeneration but if you really want to keep using it you can create a `lambdapi.opam.template` and put just the `available` field there. See https://dune.readthedocs.io/en/stable/opam.html#opam-template

Thanks

PR #23508, opened 2023-03-14
Octez 16.0 packages

killian-delarue:
Packages for Octez version 16.0


kit-ty-kate:
We'll remove the faulty tests later as per previous versions of the tezos packages.. Good to merge in the meantime as this is going to take too long to retest once again

PR #23518, opened 2023-03-16
[NEW] Add autofonce.0.8, modern runner for autoconf testsuites

lefessan:
Autofonce is a modern runner for Autoconf Testsuites. Though old, the format of Autoconf Testsuites is easy to understand and to use. Autofonce makes them even more interesting, with automatic parallel execution and automatic promotion of results to fix tests.


lefessan:
I think it’s ok now, CI fails with an internal error…

kit-ty-kate:
Thanks

kit-ty-kate:
I love the name btw :D

PR #23521, opened 2023-03-16
Package errpy.0.0.7

jasontatton:
### `errpy.0.0.7`
Errpy: An Error Recovering Rust Python Parser

---

---


jasontatton:
Seems like we are still having a lot of trouble with `[tests (failed: Library "oUnit" not found.)]`. To fix this I will amend the dune config to reference `ounit2` instead of `oUnit`. Hopefully this will resolve the problem. Troublingly when running the local build `dune build @install @runtest` or `opam install .` this does not manifest as a problem

PR #23526, opened 2023-03-17
[new release] ezxenstore (0.4.3)

psafont:
An easy-to-use interface to xenstore

- Project page: <a href="https://github.com/xapi-project/ezxenstore">https://github.com/xapi-project/ezxenstore</a>

##### CHANGES:

* xenstore_transport.unix is deprecated
* Avoid failing tests in CI without access to xenstore

### DIFF
--- a/packages/ezxenstore/ezxenstore.0.4.3/opam
+++ b/packages/ezxenstore/ezxenstore.0.4.3/opam
@@ -1,0 +1,33 @@
+opam-version: "2.0"
+maintainer: "jonathan.ludlam@citrix.com"
+authors: ["xen-api@lists.xensource.com"]
+license: "ISC"
+homepage: "https://github.com/xapi-project/ezxenstore"
+bug-reports: "https://github.com/xapi-project/ezxenstore/issues"
+dev-repo: "git+https://github.com/xapi-project/ezxenstore.git"
+build: [[ "dune" "build" "-p" name "-j" jobs ]]
+depends: [
+  "ocaml"
+  "dune" {>= "1.4"}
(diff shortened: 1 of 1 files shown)


psafont:
Most of the failures are because xenctrl cannot be built in opam's sandbox

psafont:
Xenctrl hasn't had a release for 6 years and it has gotten more unmaintained over time. Unlike the mirage bindings I don't think it's usable outside xapi/xenserver. I'll close this issue. Maybe when Xen has rid of the unstable bindings and xenctrl/xenopsd is free of them we can have a good story for releasing all these libraries :(

packages/ezxenstore/ezxenstore.0.4.3/opam
```diff
@@ -0,0 +1,33 @@
...
+]
+synopsis:
+  "An easy-to-use interface to xenstore"
+description: """
+An easy-to-use xenstore library with a simplified interface geared
+towards use within a daemon that maintains a single connection to
+xenstored."""
+url {
+  src:
+    "https://github.com/xapi-project/ezxenstore/releases/download/0.4.3/ezxenstore-0.4.3.tbz"
```
kit-ty-kate, line 27: This url 404s

psafont: Oops, forgot to run the release step, should be available now

mseri:
Indeed, can we do anything about 
https://github.com/xapi-project/ocaml-xen-lowlevel-libs/blob/master/configure#L3
?
It feels like a small patch would be enough to enable it.

PR #23574, opened 2023-03-24
Add a missing lower-bound to dune >= 2.0.0

kit-ty-kate:
Ports https://github.com/ocaml/dune/pull/7396
Should only be merged after the next dune release to avoid rebuilds for people who use opam < 2.2

cc @emillon


emillon:
Sounds good. There should be a 3.7.1 in the next couple weeks.

PR #23586, opened 2023-03-28
Add HOMEBREW_NO_AUTO_UPDATE=1 to homebrew packages

samoht:
See https://github.com/Homebrew/brew/issues/14663#issuecomment-1437321723 for the rational.

/cc @kit-ty-kate


avsm:
Shouldn't this just be added by the opam depext integration? There doesn't seem to be a case where you wouldn't want this variable set on macOS when using Homebrew with opam.

mseri:
This needs to go there, I think you are right, but it would not solve the issue for most users since the current opam and previous depext would not pick up the fix. 

I think this was also the discussion with had when `HOMEBREW_NO_INSTALL_FROM_API = "1"` was introduced.

samoht:
I'm merging this to unlock the CI on macOS, but I've also opened https://github.com/ocaml-opam/opam-depext/issues/151 to track the proper fix.

PR #23590, opened 2023-03-29
Add libtorch 2.0.0

LaurentMazare:
Add the libtorch binary packages first before moving the torch packages to 2.0.0 too.

### DIFF
--- a/packages/libtorch/libtorch.2.0.0+macos-x86_64/opam
+++ b/packages/libtorch/libtorch.2.0.0+macos-x86_64/opam
@@ -1,0 +1,32 @@
+opam-version: "2.0"
+homepage: "https://github.com/LaurentMazare/ocaml-torch"
+maintainer: "lmazare@gmail.com"
+bug-reports: "https://github.com/LaurentMazare/ocaml-torch/issues"
+dev-repo: "git+https://github.com/LaurentMazare/ocaml-torch.git"
+license: "BSD-3-Clause"
+authors: [
+  "Laurent Mazare"
+]
+install: [
+  [
+    "sh"
+    "-c"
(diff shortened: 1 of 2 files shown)


packages/libtorch/libtorch.2.0.0+macos-x86_64/opam
```diff
@@ -0,0 +1,32 @@
...
+libtorch library."""
+extra-source "libtorch-macos.zip" {
+  src: "https://download.pytorch.org/libtorch/cpu/libtorch-macos-2.0.0.zip"
+  checksum: "md5=b1a87518e95f24452af71759b64f2e1f"
+}
+extra-source "mklml-macos.tgz" {
+  src: "https://github.com/intel/mkl-dnn/releases/download/v0.17.2/mklml_mac_2019.0.1.20181227.tgz"
+  checksum: "md5=a8b4b158dc8e7aad13c0d594a9a8d241"
+}
+available: arch = "x86_64" & os = "macos"
```
avsm, line 32: Is switching to https://github.com/OpenImageDenoise/mkl-dnn something that might work here, so that it compiles on macos-arm64 as well?

LaurentMazare: arm64 support would be great to add, the mkl part might work with the repo you mentioned but I'm afraid the libtorch part doesn't have pre-built binaries at the moment, see https://github.com/pytorch/pytorch/issues/96046 .
A possibility would be to extract the libtorch bit from the python package, see https://github.com/LaurentMazare/tch-rs/issues/629 for more details but that's certainly a bit hacky (and I don't have a mac at hand to test it out, I should probably get one as lots of the issues for tch-rs/ocaml-torch end up being arm64/mac folks trying to use it).

avsm: I think if upstream pytorch doesn't have it yet, it's certainly not a blocker for this PR.  ARM Macs are recommended; they're fast :-)

LaurentMazare: Thanks, marking this as resolved for now and hopefully I'll get a mac soonish.

mseri:
Thanks

PR #23598, opened 2023-03-30
not-ocamlfind 0.10: extend not-ocamlfind preprocess

chetmurthy:
not-ocamlfind preprocess should accept args that `ocamlfind ocamlc` accepts, ignoring those that aren't relevant, so that you can just replace `ocamlc` with `preprocess`.  This fixes a gap in that rule.


mseri:
Thanks

PR #23601, opened 2023-03-31
add coqide.8.17.0

palmskog:
Followup to #23584.

cc: @MSoegtropIMC @Zimmi48

### DIFF
--- a/packages/coqide/coqide.8.17.0/opam
+++ b/packages/coqide/coqide.8.17.0/opam
@@ -1,0 +1,44 @@
+opam-version: "2.0"
+synopsis: "The Coq Proof Assistant --- GTK3 IDE"
+description: """
+Coq is a formal proof management system. It provides
+a formal language to write mathematical definitions, executable
+algorithms and theorems together with an environment for
+semi-interactive development of machine-checked proofs.
+
+This package provides the CoqIDE, a graphical user interface for the
+development of interactive proofs."""
+maintainer: ["The Coq development team <coqdev@inria.fr>"]
+authors: ["The Coq development team, INRIA, CNRS, and contributors"]
+license: "LGPL-2.1-only"
+homepage: "https://coq.inria.fr/"
+doc: "https://coq.github.io/doc/"
+bug-reports: "https://github.com/coq/coq/issues"
+depends: [
+  "dune" {>= "2.9"}
+  "ocamlfind" {build}
+  "conf-findutils" {build}
+  "conf-adwaita-icon-theme"
+  "coqide-server" {= version}
+  "cairo2" {>= "0.6.4"}
+  "lablgtk3-sourceview3" {>= "3.1.2"}
+]
+build: [
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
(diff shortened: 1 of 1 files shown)


palmskog:
The error on OCaml 5.0 is seemingly due to the (transitive) dependency library Cairo:
```
# /home/opam/.opam/5.0/.opam-switch/build/cairo2.0.6/_build/default/src/cairo_stubs.c:1821: undefined reference to `alloc_custom'
# /usr/bin/ld: /home/opam/.opam/5.0/lib/cairo2/libcairo_stubs.a(cairo_stubs.o):/home/opam/.opam/5.0/.opam-switch/build/cairo2.0.6/_build/default/src/cairo_stubs.c:1822: more undefined references to `alloc_custom' follow
# Error: Error during linking (exit code 1)
```
I don't think we can do much about this in the package here.

MSoegtropIMC:
For Coq Platform this is not an issue since I am anyway using OCaml 4.14, so I would suggest to merge this with OCaml restricted to <5~. I expect that the overlap of OCaml 5 users and CoqIDE-only users is quite small.

palmskog:
@MSoegtropIMC but the problem is here not with the CoqIDE package. I'd argue the `coqide` package here should be kept compatible with OCaml 5. I think the Cairo problem is restricted to only some platform (macOS?).

MSoegtropIMC:
Just to make sure I get this correctly: the issue is that the OCaml wrapper for cairo is in general OCaml 5 compatible, but on some platforms not?

MSoegtropIMC:
> I think we just specify "cairo2" { >= "0.6.4" } and be done with it?

Sounds reasonable to me. I would prefer this general restriction over having this restriction only for OCaml 5 so that we don't have unnecessary deviations which might lead to hard to find bugs.

I guess you report at the same time upstream that cairo < 0.6.4 is incompatible with OCaml 5?

palmskog:
> I guess you report at the same time upstream that cairo < 0.6.4 is incompatible with OCaml 5?

I don't know whether this is specific to how Coq indirectly uses cairo2 or it's a general issue. So we could add the constraint into the Coq repo, but no cycles here to investigate the root cause.

palmskog:
Now the only failures are a mysterious "Job failed: Disconnected: Switch turned off" message and a macOS failure for the gmp system package dependency. Not sure I can do much about those...

MSoegtropIMC:
Why does coqide fail on GMP and coq not?

palmskog:
@MSoegtropIMC if you look in the [last CI run for Coq 8.17.0](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/736478400d5ccaa20c573d19f86f06b3a448cafc), it **did** fail on macOS due to gmp.

MSoegtropIMC:
It is more an issue in the Opam main repo - it would make more sense to report it there and put me on CC.

mseri:
It seems to me that the remaining errors are due to infrastructure issue. I am checking them but I don't think I see anything that would prevent merging.

I am rerunning them just to check if they are transient failures

kit-ty-kate:
Thanks

PR #23604, opened 2023-03-31
Add release of Morbig 0.10.4

Niols:
It is an old release but it was never published to OPAM. Other software actually depend on that version so it is worth publishing now in my opinion.

### Morbig 0.10.4

Morbig is a parser for shell scripts written in the POSIX shell script language. It parses the scripts statically, that is without executing them, and constructs a concrete syntax tree for each of them. The concrete syntax trees are built using constructors according to the shell grammar of the POSIX standard.


mseri:
Thanks

PR #23607, opened 2023-04-01
New release of Camlp5 (8.01.00), breaking change in API

chetmurthy:
The original Camlp5 API for installing new grammars doesn't allow to detect that somebody has installed a new grammar twice (which is almost always a bug); this gets fixed.

But several packages relied on the old behaviour, and I need to send out PRs to fix them.  I'm disabling them (with upper-bound constraints on their dependency on camlp5) here so that I can release this, and then can get to work on the PRs.


chetmurthy:
OK, looks like everything passed, except for the MacOS jobs that failed b/c of missing conf-perl-* packages; once this gets released, I can release the pa_ppx packages that build on this, and then everything will no longer rely on conf-perl-* .

chetmurthy:
I should clarify: the failed jobs were all "revdeps" (things that depend on Camlp5) that failed b/c they never built previously on MacOS (or other platforms where these conf-perl-* packages weren't available).  When this PR gets released, I can push new versions of the revdeps (and send PRs to the two packages that I don't develop) that will fix things so that the conf-perl-* packages are no longer needed.

mseri:
Thanks

PR #23616, opened 2023-04-03
new release of pa_ppx 0.12

chetmurthy:
* compatible with camlp5 8.01.00

* remove all perl scripts that depend on conf-perl-* packages

  -- and all but one simple perl script

### DIFF
--- a/packages/pa_ppx/pa_ppx.0.12/opam
+++ b/packages/pa_ppx/pa_ppx.0.12/opam
@@ -1,0 +1,73 @@
+synopsis: "PPX Rewriters for Ocaml, written using Camlp5"
+description:
+"""
+This is a collection of PPX rewriters, re-implementing those based on ppxlib
+and other libraries, but instead based on Camlp5.  Included is also a collection
+of support libraries for writing new PPX rewriters.  Included are:
+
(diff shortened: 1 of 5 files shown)


chetmurthy:
OK, everything builds, except on macos-homebrew, where there are two sorts of errors:

1. revdeps packages (that depend on this package) that didn't build previously due to conf-perl-* packages: this is innocuous

2. there's some sort of build error that only happens on macos-homebrew-ocaml-*-arm64 -- where ocamlfind seems to be constructing the wrong ocamlc invocation.  On other platforms it seems to be working fine.  It's a nonessential build-step, so I could eliminate it.

I don't have access to a Mac, or I would debug this.  If it is allowed, I can submit a smaller package that only allows building on these two macos platforms, and see if I can figure out what' wrong.  Or if there's a way I can submit jobs to the opam CI infrastructure, I could do that.

Or, since these are experimental platforms, we could just ignore ?

raphael-proust:
It seems like the build scripts are not very portable in some ways or another (something related to shell quote plugin for perl). It's fine as is: users can probably still install the package if they go through some additional steps on their setup.

PR #23618, opened 2023-04-04
Release version 1.1.3 of Orsetto.

jhwoodyatt:
Release Conjury 1.1.3.

### DIFF
--- a/packages/orsetto/orsetto.1.1.3/opam
+++ b/packages/orsetto/orsetto.1.1.3/opam
@@ -1,0 +1,73 @@
+opam-version: "2.0"
+synopsis: "A library of assorted structured data interchange languages"
+maintainer: "james woodyatt <jhw@conjury.org>"
+authors: "james woodyatt <jhw@conjury.org>"
+homepage: "https://bitbucket.org/jhw/orsetto/"
+bug-reports: "https://conjury.atlassian.net/browse/ORS"
+dev-repo: "git+https://bitbucket.org/jhw/orsetto"
+description: """
+Orsetto is a standalone library comprising a core toolkit...
+
+    - Core functional data structures and processes.
+    - Unicode transport, normalization, parsing and formatting.
+    - General purpose packet format encoder-decoder processes.
+
+...used to implement streaming parsers and formatters for a useful variety of
+structured data interchange languages...
+
+    - JavaScript Object Notation (JSON).
+    - Concise Binary Object Representation (CBOR).
+    - DjB's 64-bit Temps Atomique International (TAI64).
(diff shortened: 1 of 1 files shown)


jhwoodyatt:
Hi, so I have a problem. This version of Orsetto requires a version of the Conjury OMake library that cannot build on systems without OCaml native mode compilers. That's why it currently fails on some of the supported architectures for OCaml 5.0. If the policy is not to merge releases that cannot build on `bytecode-only` switches, then I need figure out how to make this package unavailable on those switches. Alas, I've studied the OPAM documentation, and it doesn't seem possible to make a package that excludes `bytecode-only` compilers.

Please advise?

jhwoodyatt:
Hi OPAM Repository Curators,

I have a problem attempting to release Orsetto 1.1.3. This release is a patch update to correct a couple of errors that were preventing the 1.1 series from compiling on OCaml 5.0, but it still uses the older Conjury V2 build logic that doesn't have support for `bytecode-only` switches. Accordingly, the OPAM-CI check is reporting failures attempting to build Orsetto 1.1.3 on the various architectures and platforms that OCaml 5.0 currently supports only with `bytecode-only` switches.

It will be a few weeks more before I can release Orsetto 1.2, which uses the newer Conjury V3 build logic that supports `bytecode-only` switches, and which also contains a few additional feature improvements. My recommendation is that you allow Orsetto 1.1.3 to be released despite it still being broken on `bytecode-only` switches. (It has never worked on `bytecode-only` switches.)

mseri:
That's not a problem. Can you add
```
conflicts: [ "ocaml-option-bytecode-only" ]
```
in the affected opam files?

packages/orsetto/orsetto.1.1.3/opam
```diff
@@ -0,0 +1,73 @@
...
+]
+license: "BSD-2-Clause"
+depends: [
+    "ocaml" { >= "4.08.1" }
+    "conjury" { build & >= "2.2" & < "3.0~" }
+    "omake" { build & >= "0.10.3" }
+    "uucd" { build & = "15.0.0" }
+    "ounit2" { build & with-test & >= "2.2" }
+]
+build: [
```
mseri, line 46:
```suggestion
conflicts: [ "ocaml-option-bytecode-only" ]
build: [
```

jhwoodyatt:
Test failures on those `macOS` Homebrew "experimental" targets appear to be related to OPAM-CI issues (the test user does not appear to have permission to access Homebrew API cache). It doesn't seem to me that the problem is with the Orsetto package on those targets. Do you agree?

mseri:
That's fine, thanks a lot

PR #23630, opened 2023-04-04
Release LLVM 16

alan-j-hu:
I have been hired by the OCaml Software Foundation to ensure the LLVM bindings stay up to date with respect to OCaml 5 and the LLVM opaque pointer API changes under supervision of @jberdine. At his request, this release contains patches for removing naked pointers so that the package will work with OCaml 5 as well as new function aliases to the API to ease the transition from typed to opaque pointers. (Note that latter change means that the API has more functions than upstream and the versions in distro package managers!) I've never released LLVM on opam before, so I'll let the CI run and see if the package installs.

CC @Kakadu

### DIFF
--- a/packages/conf-llvm/conf-llvm.16.0.0/files/configure.sh
+++ b/packages/conf-llvm/conf-llvm.16.0.0/files/configure.sh
@@ -1,0 +1,24 @@
+#!/bin/bash -ex
+
+version=${1/%.?.?/}
+
+if hash brew 2>/dev/null; then
+    brew_llvm_config="$(brew --cellar)"/llvm*/${version}*/bin/llvm-config
+fi
+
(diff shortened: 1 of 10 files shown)


Kakadu:
@alan-j-hu It looks like current CI run is not meaningful. Could you try to add package 'conf-llvm' of appropriate version?

kit-ty-kate:
Hi @alan-j-hu, do you have some time to talk in a voice chat (e.g. Jitsi) about this PR (and #22142) at some point? If you do I'll send you a link via email

alan-j-hu:
Sure, I can do a voice chat if you think it's necessary to work on these PRs.

kit-ty-kate:
#24330 was accepted and merged so let's close this one for now.

PR #23641, opened 2023-04-05
update pa_ppx_* packages (after pa_ppx.0.12) [try 2]

chetmurthy:
New packages: pa_ppx_{static,regexp,quotation2extension}

updated packages: pa_ppx_{hashcons,unique,migrate}

Major change is to remove all dependency on Perl

Also Makefile reorganization


chetmurthy:
Everything passed, except macos, where:

1. during tests, system packages libpcre and pkg-config somehow were found to not be installed (even though at least some of the runs I checked, they *had* been installed), and then the reinstallation failed.  Very weird, and I have no idea what might be going wrong.

2. some of the install runs seem to blow on some sort of odd zfs error.  These don't even get started with the install.

But everything else passed.

PR #23645, opened 2023-04-06
Tezt: sync opam metadata from upstream repo

hhugo:
See https://gitlab.com/nomadic-labs/tezt/-/merge_requests/27
cc @rbardou


mseri:
Thanks

PR #23646, opened 2023-04-06
[new release] jsoo-react (0.1)

jchavarri:
Project page: https://github.com/ml-in-barcelona/jsoo-react

Release notes: https://github.com/ml-in-barcelona/jsoo-react/releases/tag/0.1

It is the first version of this package published in opam, and also the first time I publish a package myself. 🙂


packages/jsoo-react/jsoo-react.0.1/opam
```diff
@@ -0,0 +1,42 @@
+# This file is generated by dune, edit dune-project instead
+opam-version: "2.0"
+synopsis: "Bindings to ReactJS for js_of_ocaml, including JSX ppx"
+maintainer: ["Javier Chávarri <javier.chavarri@gmail.com>"]
+authors: ["Javier Chávarri <javier.chavarri@gmail.com>"]
+license: "MIT"
+homepage: "https://github.com/ml-in-barcelona/jsoo-react"
+bug-reports: "https://github.com/ml-in-barcelona/jsoo-react/issues"
+depends: [
+  "dune" {>= "2.7" & >= "2" & < "4"}
```
mseri, line 10:
```suggestion
  "dune" {>= "2.7"}
```
This can also be fixed by removing dune as a dependency in the dune-project file (it is added automatically). I would not add an upper bound here, dune tends to be backward compatible and the CI thoroughly checks this upon release

packages/jsoo-react/jsoo-react.0.1/opam
```diff
@@ -0,0 +1,42 @@
...
+  "ocaml" {>= "4.12.0" & < "5.0.0"}
+  "js_of_ocaml" {>= "4.0.0" & < "5.2.0"}
```
mseri, line 12:
```suggestion
  "js_of_ocaml" {>= "4.0.0"}
```
Are you sure about this upper bound? We check if these are needed when the packages are release and we re-add them manually where needed if that is the case

packages/jsoo-react/jsoo-react.0.1/opam
```diff
@@ -0,0 +1,42 @@
...
+  "gen_js_api" {>= "1.0.8" & < "1.2.0"}
```
mseri, line 13:
```suggestion
  "gen_js_api" {>= "1.0.8"}
```
Are you sure about this upper bound? We check if these are needed when the packages are release and we re-add them manually where needed if that is the case

packages/jsoo-react/jsoo-react.0.1/opam
```diff
@@ -32,7 +32,6 @@ build: [
     "-j"
     jobs
     "@install"
-    "@runtest" {with-test}
```
mseri: I would revert this. The failure to install conf-npm in the CI is not preventing the merge and having the tests run in all other environments will ensure that this package is properly tested each time an update of its dependencies is released

jchavarri: The removal was not due to the conf-npm error, but a different one:

```
Error: Cannot find module 'react-dom/test-utils'
```

Can be seen [here](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/8cf8a29a6bbd5fe9a3629a0852ad091c87849542/variant/compilers,4.13,jsoo-react.0.1,tests).

The reason is that to run the tests we need to install some dependencies from npm first. I guess we should add `conf-npm` as a `with-test` dependency, and then somehow make sure that `npm install` is called before or as part of the `runtest` alias. I am not sure how to do this last part, so for now I thought I would disable the tests as a workaround while I figure it out.

mseri: Oh, I see. Thanks for the clarification. In that case, unless you can bundle the dependencies already in the tarball, let's disable it.

jchavarri:
@mseri Thanks for your review, I think all your suggestions have been tackled.

I am working on the `runtest` alias re-addition already, I will create separately a new minor release to add them back as soon as it's working.

mseri:
Thanks. Everything was already tested and the new commit passes the lint, I am going to merge

mseri:
Feel free to announce the release on discuss.ocaml.org. As you know we have the community category and the announcement tag for this purpose

PR #23653, opened 2023-04-09
3 packages from mefyl/timmy at 1.0.0

mefyl:
This pull-request concerns:
- `timmy.1.0.0`: Time and calendar library
- `timmy-jsoo.1.0.0`: Js_of_ocaml bindings for Timmy
- `timmy-unix.1.0.0`: Unix clock implementation for Timmy

---
* Homepage: https://github.com/mefyl/timmy
* Source repo: git+https://github.com/mefyl/timmy.git
* Bug tracker: https://github.com/mefyl/timmy/issues

---


packages/timmy/timmy.1.0.0/opam
```diff
@@ -0,0 +1,50 @@
...
+  "ppx_here"
+  "ppx_deriving"
+  "ptime"
+  "alcotest" {with-test}
+  "odoc" {with-doc}
+  "cmdliner" {dev}
+  "opam-file-format" {dev}
+  "sexplib" {dev}
+  "stdio" {dev}
+  "shexp" {dev}
```
kit-ty-kate, line 23: What is the purpose of these dependencies?

packages/timmy/timmy.1.0.0/opam
```diff
@@ -0,0 +1,50 @@
...
+synopsis: "Time and calendar library"
+maintainer: ["mefyl <mefyl@gruntech.org>"]
+authors: ["mefyl <mefyl@gruntech.org>"]
+license: "LGPL-3.0-or-later"
+homepage: "https://github.com/mefyl/timmy"
+bug-reports: "https://github.com/mefyl/timmy/issues"
+depends: [
+  "dune" {>= "3.0"}
+  "base"
+  "fmt"
```
kit-ty-kate, line 12:
```suggestion
  "fmt" {>= "0.8.7"}
```

packages/timmy/timmy.1.0.0/opam
mefyl: We have a utility executable (`.logistic/dune/extdeps/extdeps.exe`) that generates iOS cross compilation opam files amongst other things; it requires these. I've tried to get dune to NOT build it by default, but didn't succeed, thus I added them to be able to use OCaml CI. An alternative would be to erase these rules when building on OCaml-CI, but that would require maintaining a separate branch just for releases, which is a bit inconvenient.

Regarding dune, it seems there is no way to override the @check default alias, and `enabled_if` in `executable` stanzas support a very limited set of variables and notably not `lib_available:opam-file-format` :-/

Any tip appreciated !

kit-ty-kate: Ah i see :/ To workaround ocaml-ci's behaviour I can't see any cleaner solution than doing this:
```suggestion
  "cmdliner" {with-test & dev}
  "opam-file-format" {with-test & dev}
  "sexplib" {with-test & dev}
  "stdio" {with-test & dev}
  "shexp" {with-test & dev}
```
It's not the greatest but at least it works and does not bring unnecessary dependencies for regular users in 99% of cases.

mefyl: My understanding was that `dev` dependencies were not installed when installing via an archive (as opposed to a git repository), is it not the case ?

In any case, applied.

packages/timmy/timmy.1.0.0/opam
```diff
@@ -0,0 +1,50 @@
...
+  "cmdliner" {dev}
+  "opam-file-format" {dev}
+  "sexplib" {dev}
+  "stdio" {dev}
+  "shexp" {dev}
+]
+depopts: [
+  "js_of_ocaml"
+  "schematic" {>= "0.17.0" & < "0.18.0"}
+]
```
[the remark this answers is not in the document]
mefyl: In that case this is used to add instrumentation to the code when schematic (unreleased for now) is available, my idea there is that if an uncompatible version is installed, we might as well still allow installation and not add the instrumentation ? Granted it can make debugging why the instrumentation is or isn't present a bit harder, but sounds more flexible wrt the set of installable packages.

We don't really care internally as we always use the latest versions, so I'm leaving the final decision to you.

packages/timmy/timmy.1.0.0/opam
kit-ty-kate: yes, but we do support this use-case too. For example an user might want to do `opam pin --dev timmy` and in this case the dependencies shouldn't be needed.

packages/timmy/timmy.1.0.0/opam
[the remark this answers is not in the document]
kit-ty-kate: The problem is that looking at https://github.com/mefyl/timmy/blob/7120fc0462f8d30e171e761f7ee7110b9d8c394f/timmy/lib/dune#L13 compilation will fail if schematic is detected but the API isn't compatible. Also with the way the depopt is currently used the package might be installed in parallel with `timmy` and cause problem there as well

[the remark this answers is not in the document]
mefyl: I think dune does not `select` the file if the optional dependency doesn't match the constraint ?

In any case I'll trust your judgment regarding what will make the package more usable, modifications applied.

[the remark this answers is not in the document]
kit-ty-kate: > I think dune does not `select` the file if the optional dependency doesn't match the constraint ?

What do you mean by that? Dune doesn't know anything about the constraints from the opam file (even when defined in the dune-project file for autogeneration purpose, if that's what you were thinking of)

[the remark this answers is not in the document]
mefyl: I thought it did, I suppose it *could* thanks to the `dune-project`. In any case if it does not I stand corrected!

mefyl:
I have fixed everything related to the lower-bound builds - awesome feature by the way. What remain are two internal errors and the test suite not passing on Ubuntu 18 because I don't think there is a good way to force the timezone for test purpose - the environment variable does not work like on other / more recent Linuxes. I suppose it's an acceptable state, let me know.

Thanks !

kit-ty-kate:
Thanks

PR #23655, opened 2023-04-10
[new release] zar (0.9.2)

bagnalla:
Formally verified sampling from discrete probability distributions

- Project page: <a href="https://github.com/bagnalla/ocamlzar">https://github.com/bagnalla/ocamlzar</a>

##### CHANGES:

* API change: include Stream module in root Zar module.

### DIFF
--- a/packages/zar/zar.0.9.2/opam
+++ b/packages/zar/zar.0.9.2/opam
@@ -1,0 +1,43 @@
+opam-version: "2.0"
+synopsis:
+  "Formally verified sampling from discrete probability distributions"
+description:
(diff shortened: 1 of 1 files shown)


packages/zar/zar.0.9.2/opam
```diff
@@ -0,0 +1,43 @@
...
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test & ocaml:version >= "4.08"}
```
kit-ty-kate, line 30: Unnecessary given the package already restricts itself to OCaml >= 4.08
```suggestion
    "@runtest" {with-test}
```

kit-ty-kate:
Thanks. Could you return the fix in the opam file upstream so that it can be taken into account for future releases?

PR #23659, opened 2023-04-11
OCaml 5.1: developement packages for the 5.1 branch

Octachron:
This PR takes care of the various dev packages for the newly tagged OCaml 5.1 branch.

More precisely, this PR creates two packages:
- ocaml-variants.5.2.0+trunk to track the latest trunk
- ocaml.5.2.0 to unblock the installation of the package above

and updates two other packages
- ocaml-variants.5.1.0+trunk to track the 5.1 branch rather than the latest trunk.
- ocaml.5.1.0 to allow installing pre-release versions of OCaml 5.1

### DIFF
--- a/packages/ocaml/ocaml.5.2.0/opam
+++ b/packages/ocaml/ocaml.5.2.0/opam
@@ -7,10 +7,10 @@ and polls it to initialise specific variables like `ocaml:native-dynlink`"""
 maintainer: "platform@lists.ocaml.org"
 depends: [
   "ocaml-config" {>= "3"}
-  "ocaml-base-compiler" {= "5.1.0"} |
-  "ocaml-variants" {>= "5.1.0~" & < "5.1.1~"} |
-  "ocaml-system" {>= "5.1.0~" & < "5.1.1~"} |
-  "dkml-base-compiler" {>= "5.1.0~" & < "5.1.1~"}
+  "ocaml-base-compiler" {= "5.2.0"} |
+  "ocaml-variants" {>= "5.2.0~" & < "5.2.1~"} |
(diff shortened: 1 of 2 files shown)


packages/ocaml/ocaml.5.2.0/opam
```diff
@@ -0,0 +1,36 @@
+opam-version: "2.0"
+license: "LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception"
+synopsis: "The OCaml compiler (virtual package)"
+description: """
+This package requires a matching implementation of OCaml,
+and polls it to initialise specific variables like `ocaml:native-dynlink`"""
+maintainer: "platform@lists.ocaml.org"
+depends: [
+  "ocaml-config" {>= "3"}
+  "ocaml-base-compiler" {= "5.2.0"} |
```
yawaramin, line 10: Hi, I'm not understanding how this works. There is no `ocaml-base-compiler.5.2.0` in opam, as far as I can tell: https://ocaml.org/p/ocaml-base-compiler/latest

Also not understanding how the `ocaml` package can be at version 5.2.0. Isn't this PR about creating packages for the 5.1.0 branch?

This relates to https://github.com/ocaml/opam/issues/5554

Octachron: This part of the PR is about unblocking the dev version of the compiler, which is indeed at version `5.2.0+dev0-2023-04-11` currently. The `ocaml` package is a metapackage that requires one of the four (base, variants, dkml or system) possible implementation (notice the `|` at the end of the line). Currently, for `5.2.0`, the only possible way to fulfill this dependency is to install `ocaml-variants.5.2.0+trunk`.

PR #23660, opened 2023-04-12
Add alpine linux target for conf-pandoc

tmcgilchrist:
Not sure if I need a new package version for this or I can add it to an existing one.

Alpine linux has pandoc available https://pkgs.alpinelinux.org/packages?name=pandoc


mseri:
I would also add the freebsd one fwiw

kit-ty-kate:
Thanks

PR #23668, opened 2023-04-14
First alpha release for OCaml 5.1.0

Octachron:
This PR adds the base and variants+option package for the first alpha release of OCaml 5.1.0.
The ocaml metapackage has also been updated to allow installing the vanilla compiler.

### DIFF
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.5.1.0~alpha1/files/ocaml-base-compiler.install
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.5.1.0~alpha1/files/ocaml-base-compiler.install
@@ -1,0 +1,1 @@
+share_root: ["config.cache" {"ocaml/config.cache"}]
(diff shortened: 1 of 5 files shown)


kit-ty-kate:
Thanks

avsm:
Something's not quite right here; I'm getting this after testing this:

image

avsm:
(The actual compiler installed is `ocaml.5.0.0` and `ocaml-base-compiler.5.0.0`)

kit-ty-kate:
@avsm That sounds like an opam issue (the message isn't correct). Could you open a ticket there with a reproduction case?

Otherwise to upgrade to the new alpha compiler you need to install it explicitly (`opam upgrade` will not do it for you) as they have been tagged with `avoid-version` to avoid everyone upgrading to the alpha versions:
```
opam install ocaml-base-compiler.5.1.0~alpha1
```

PR #23672, opened 2023-04-14
Fix ppx_cstubs version constraints

tjammer:
The one in the opam repo has two version constraints which make it impossible to use with OCaml 5. The opam file in the ppx_cstubs repo at the 0.7 tag doesn't have
them (https://github.com/fdopen/ppx_cstubs/blob/0.7.0/ppx_cstubs.opam). This PR removes the version constraint to fix this, see relevant discussion here https://github.com/fdopen/ppx_cstubs/issues/10


kit-ty-kate:
Thanks

PR #23686, opened 2023-04-18
[new release] merlin and merlin-lib (4.9-501-preview)

voodoos:
CHANGES:

Tue Apr 24 16:55:42 CEST 2023

  + merlin binary 
    - Preview support for OCaml 5.1-alpha1. Short path is temporary disabled and inline records might 
      not behave as expected.


voodoos:
> Shouldn't the version be `510-preview` or `51-preview`?

You're right ! 

Do you think it is worth re-releasing ?

kit-ty-kate:
Usually the version number for previews is of the following form: `4.9~5.1preview`

PR #23714, opened 2023-04-25
Release of Solo5.0.8.0

dinosaure:
* Be able to build `spt`, `virtio`, `muen` and `xen` targets on OpenBSD (@adamsteen, solo5/solo5#544). This change does not allow us to "run" these targets on OpenBSD
* Fix linker scripts with TLS (Thread Local Storage) sections (@palainp, @hannesm, @dinosaure, solo5/solo5#542)
* Export TLS symbols (@palainp, @hannesm, @dinosaure, solo5/solo5#546) **breaking change** due to solo5/solo5#542 & solo5/solo5#546, tenders must be **upgraded**. Indeed, solo5.0.7.* tenders will not be able to load correctly unikernels compiled with solo5.0.8.0. The internal ABI version for `solo5-hvt`/`solo5-spt` was upgraded accordingly.

  This version implements Thread Local Storage. The user can initialise a TLS block with `solo5_tls_init` on a pointer to `solo5_tls_size()` free bytes. Then, the user is able to set the `tp` (Thread Pointer) pointer via `solo5_set_tls_base(solo5_tls_tp_offset(tls_block))`. More details are available into `solo5.h`.

  Note: this change does not allow a Solo5 unikernel to use multiple cores! It only provides an API required by OCaml 5 / pthread to launch, at most, one thread.
* Fix tests reported by NixOS (@greydot, @ehmry, solo5/solo5#547)
* Split out the `time.c` implementation between Muen and HVT (@dinosaure, @Kensan, solo5/solo5#552)
* User hypercall instead of TSC-based clock when the user asks for the wall-clock (@dinosaure, @reynir, solo5/solo5#549, solo5/solo5#550)

  Note: only hvt & virtio are updated to avoid a clock drift on the wall-clock. Indeed, when the unikernel is suspended, the wall-clock is not updated. Muen & Xen still use a TSC-based wall-clock. The spt target was already in sync with the host's wall-clock.
* Improve the muen clock (@Kensan, solo5/solo5#553)
* Fix the `.bss` section according to solo5/solo5#542 & solo5/solo5#546. The `.bss` section is tagged with `PT_LOAD`. Tenders are available to load this section properly. (@Kensan, @dinosaure, solo5/solo5#551, solo5/solo5#554)
* Fix the cross-compilation of Solo5 for `aarch64` (@dinosaure, @palainp, @hannes, solo5/solo5#555)

### DIFF
--- a/packages/solo5-cross-aarch64/solo5-cross-aarch64.0.8.0/opam
+++ b/packages/solo5-cross-aarch64/solo5-cross-aarch64.0.8.0/opam
@@ -1,0 +1,45 @@
+opam-version: "2.0"
+maintainer: "martin@lucina.net"
+authors: [
+  "Dan Williams <djwillia@us.ibm.com>"
+  "Martin Lucina <martin@lucina.net>"
+  "Ricardo Koller <kollerr@us.ibm.com>"
+]
+homepage: "https://github.com/solo5/solo5"
+bug-reports: "https://github.com/solo5/solo5/issues"
+license: "ISC"
+dev-repo: "git+https://github.com/solo5/solo5.git"
+build: [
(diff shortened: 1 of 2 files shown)


mseri:
Looks fine

PR #23715, opened 2023-04-26
[ANN] Add header-check.0.1.0, a tool to check all headers and manage them

lefessan:


packages/header-check/header-check.0.1.0/opam
```diff
@@ -0,0 +1,50 @@
...
+license: "LGPL-2.1-only WITH OCaml-LGPL-linking-exception"
+synopsis: "A tool to check and update source headers, using checksums"
+description: """\
+The header-check tool scans a source project, lists all existing headers and
+can be used to update them. It uses checksums to create identifiers for 
+headers and works with OCaml, C/C++ and Shell kinds of comments.
+"""
+authors: ["Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>"]
+maintainer: ["Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>"]
+homepage: "https://ocamlpro.github.io/header-check"
```
raphael-proust, line 11: I have a 404 on this link. Is this expected for now? Is there a different URL to add?

Same remark on the doc link below.

This is not necessarily blocking, but if you want to fix it, you might as well do it before the pull.

lefessan: Thanks for noticing, I decided to push a website there, so it should be ok for now.

packages/header-check/header-check.0.1.0/opam
```diff
@@ -0,0 +1,50 @@
...
+depends: [
+  "ocaml" {>= "4.07.0"}
+  "dune" {>= "2.7.0"}
+  "ocplib_stuff" {>= "0.1"}
+  "ez_file" {>= "0.3"}
+  "base-unix" {>= "base"}
+  "ppx_inline_test" {with-test}
+  "ppx_expect" {with-test}
+  "odoc" {with-doc}
+  "ocamlformat" {with-test}
```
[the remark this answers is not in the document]
lefessan: Actually, I would prefer to wait and see, and add versions only when it becomes necessary.

packages/header-check/header-check.0.1.0/opam
mseri: It's working now

kit-ty-kate:
Thanks

PR #23717, opened 2023-04-26
[new release] yojson (2.1.0)

Leonidas-from-XIV:
Yojson is an optimized parsing and printing library for the JSON format

- Project page: <a href="https://github.com/ocaml-community/yojson">https://github.com/ocaml-community/yojson</a>
- Documentation: <a href="https://ocaml-community.github.io/yojson">https://ocaml-community.github.io/yojson</a>

##### CHANGES:

*2023-04-26*

### Added

- Add `Yojson.Raw.Util` module to provide combinators for extracting fields
  from `Yojson.Raw.t` values. (@tmcgilchrist, ocaml-community/yojson#163)

- Add `Util.path` function to recurse into an object through a list of keys.
  (@cuihtlauac, @Leonidas-from-XIV, ocaml-community/yojson#157)

### DIFF
--- a/packages/yojson/yojson.2.1.0/opam
+++ b/packages/yojson/yojson.2.1.0/opam
@@ -1,0 +1,48 @@
+opam-version: "2.0"
+synopsis:
+  "Yojson is an optimized parsing and printing library for the JSON format"
+description: """
+Yojson is an optimized parsing and printing library for the JSON format.
+
+ydump is a pretty-printing command-line program provided with the
+yojson package."""
+maintainer: [
+  "paul-elliot@tarides.com" "nathan@tarides.com" "marek@tarides.com"
+]
+authors: ["Martin Jambon"]
+license: "BSD-3-Clause"
+homepage: "https://github.com/ocaml-community/yojson"
+doc: "https://ocaml-community.github.io/yojson"
+bug-reports: "https://github.com/ocaml-community/yojson/issues"
+depends: [
+  "dune" {>= "2.7"}
+  "ocaml" {>= "4.02.3"}
+  "cppo" {build}
+  "alcotest" {with-test & >= "0.8.5"}
+  "seq" {>= "0.2.2"}
+  "odoc" {with-doc}
+]
+build: [
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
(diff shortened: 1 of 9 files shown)


packages/yojson/yojson.2.1.0/opam
```diff
@@ -0,0 +1,48 @@
...
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
+dev-repo: "git+https://github.com/ocaml-community/yojson.git"
+url {
+  src:
+    "https://github.com/ocaml-community/yojson/releases/download/2.1.0/yojson-2.1.0.tbz"
```
kit-ty-kate, line 42: This URL 404s

Leonidas-from-XIV: Oops, forgot to run `dune-release publish`. Fixed, the URL should work now.

kit-ty-kate:
frama-c seems to be failing due to the new change. Could you have a look and constrain the dependency if that is expected?
```
# File "src/libraries/utils/json.mll", line 374, characters 8-16:
# Error: This expression has type
#          (string list -> Yojson.Basic.t -> Yojson.Basic.t option,
#           [> `Assoc of (String.t * Yojson.Basic.t) list
#            | `List of Yojson.Basic.t list ]
#           as 'a)
#          Hashtbl.t =
#            (string list -> Yojson.Basic.t -> Yojson.Basic.t option, 'a)
#            Hashtbl.t
#        but an expression was expected of type
#          (Filepath.Normalized.t, Yojson.Basic.t) Hashtbl.t
#        Type string list -> Yojson.Basic.t -> Yojson.Basic.t option
#        is not compatible with type Filepath.Normalized.t 
```

Leonidas-from-XIV:
Thanks for catching it. I added a constraint on all Frama-C versions that had the constraint 2.0.0 when testing. This was unexpected but indeed the code https://git.frama-c.com/pub/frama-c/-/blob/master/src/libraries/utils/json.mll#L337 uses `path` as an argument, then opens the `Util` module (which comes with `path` since 2.1.0) thus overwriting its own `path` and adds unexpected things into the `Hashtbl.t`. Definitely not what should happen.

@bobot maybe you could fix this upstream? There seem to be minor incompatibilities already since the 2.0.0 release.

PR #23722, opened 2023-04-27
[ANN] Add module-graph.0.1.0, computes the graph of dependencies between compiled modules 

lefessan:
`module-graph` computes the graph of dependencies between modules using object files (`.cmt`, `.cmti`, `.cmi`) instead of `ocamldep`.

### DIFF
--- a/packages/module-graph/module-graph.0.2.0/opam
+++ b/packages/module-graph/module-graph.0.2.0/opam
@@ -1,0 +1,48 @@
+opam-version: "2.0"
+license: "LGPL-2.1-only WITH OCaml-LGPL-linking-exception"
+synopsis:
+  "The module-graph tool generates a graph of dependencies between OCaml modules using compiled object files"
+description: """\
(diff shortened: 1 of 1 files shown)


packages/module-graph/module-graph.0.2.0/opam
```diff
@@ -0,0 +1,49 @@
...
+depends: [
+  "ocaml" {>= "4.07.0"}
+  "dune" {>= "2.7.0"}
+  "ocaml-base-compiler" {}
+  "ez_file" {>= "0.3"}
+  "ocplib_stuff" {>= "0.3"}
+  "ppx_inline_test" {with-test}
+  "ppx_expect" {with-test}
+  "odoc" {with-doc}
+  "ocamlformat" {with-test}
```
mseri, line 43: Same comment as in the other PR. This tends to be unstable. It is ok to add the bound if/once it breaks

PR #23736, opened 2023-04-28
Remove constraints on package versions in conflicts, it's unsupported

Leonidas-from-XIV:
After a debugging session with @kit-ty-kate as part of fixing https://github.com/tarides/opam-monorepo/issues/349 and https://github.com/ocaml-opam/opam-0install-solver/pull/46 we realized that OPAM doesn't support resolving package variables in `conflicts` fields (the linter complains about this in `depends` fileds but not in `conflicts` fields, but the solver never resolves these variables). So this constraint doesn't do anything in OPAM and it causes issues in the opam-0install-solver which attempts to resolve them.

### DIFF
--- a/packages/ocaml-options-vanilla/ocaml-options-vanilla.1/opam
+++ b/packages/ocaml-options-vanilla/ocaml-options-vanilla.1/opam
@@ -8,12 +8,7 @@ depends: [
 conflicts: [
   "ocaml-option-32bit"
   "ocaml-option-afl"
-  "ocaml-option-bytecode-only"
-    {ocaml-system:version < "5" |
-     ocaml-base-compiler:version < "5" |
-     ocaml-variants:version < "5" |
-     arch = "arm64" |
-     arch = "x86_64" }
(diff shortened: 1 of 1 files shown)


mseri:
@kit-ty-kate can we merge this or we should wait to avoid triggering rebuilds?

tmcgilchrist:
The package linting is pointing out that some fields are missing.

```
2023-05-12 23:30.49: Error in ocaml-options-vanilla.1:            warning 68: Missing field 'license'
2023-05-12 23:30.49: Error in ocaml-options-vanilla.1:            warning 36: Missing field 'bug-reports'
2023-05-12 23:30.49: Error in ocaml-options-vanilla.1:            warning 35: Missing field 'homepage'
2023-05-12 23:30.49: Error in ocaml-options-vanilla.1:            warning 25: Missing field 'authors'
```

mseri:
No, the ocaml-options packages are special in that respect, they only turn on or off flags in the ocaml compiler packages and mark mutual incompatibilities. You can ignore these warnings and errors in this case

avsm:
Lets go ahead and get this in, as it's a violation of the opam format spec.

PR #23742, opened 2023-05-02
Jane Street packages v0.16.0

dkalinichenko-js:
This PR adds Jane Street packages v0.16.0.

PR #23777, opened 2023-05-11
Github CI actions

toots:
Hi!

I noticed that the official CI is currently being rebuilt so I dusted off my old https://github.com/ocaml/opam-repository/pull/20322 PR. Hopefully, this can help move things forward while the regular CI is rebuilt?

The actions do not include everything that the official CI supports, in particular extra features and architectures but it does check against a bunch of OCaml compiler versions as well as platforms and revdeps, hopefully enough to start merging some PRs!

It should also be possible to cherry-pick the one commit from this PR, add it to another pending PR, check the result of the actions and remove it before the merge if y'all don't want to include the files in the main branch. Here's an example with the pending `ogg.0.7.4`: https://github.com/ocaml/opam-repository/actions/runs/4942730965

Example run with two packages (`faad.0.4.0` and (artificially picked) `liquidsoap.2.0.1`): https://github.com/savonet/opam-repository/actions/runs/4942469715

I tested it against the Janest PR https://github.com/ocaml/opam-repository/pull/23742 but, oh my, no this one is still way too big for it.. See: https://github.com/savonet/opam-repository/actions/runs/4942483639 🙂

### DIFF
--- a/.github/check-installable/action.yml
+++ b/.github/check-installable/action.yml
@@ -1,0 +1,35 @@
+name: Check if package is installable
+inputs:
+  opam:
+    required: true
+  package:
+    required: true
+outputs:
+  installable:
+    description: "Is the package installable?"
+    value: ${{ steps.installable.outputs.installable }}
+runs:
+  using: composite
+  steps:
+    - name: Get opam cli option
+      id: cli
+      shell: sh
+      run: |
+        if [ "${{ inputs.opam }}" = "opam-2.1" ]; then
+          echo 'cli=--cli=2.1' >> $GITHUB_OUTPUT
+        fi
+    - name: Check if package is installable
+      id: installable
+      shell: sh
+      run: |
+        if opam ${{ steps.cli.outputs.cli }} list --readonly --external --resolve=${{ inputs.package }} >/dev/null 2>&1; then
+          echo 'installable=true' >> $GITHUB_OUTPUT
+        else
+          echo "Package is not installable";
+        fi
+    - name: Annotate result
+      if: ${{ !steps.installable.outputs.installable }}
+      uses: actions/github-script@v5
+      with:
+        script: |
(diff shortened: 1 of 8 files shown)


Niols:
The CI on this very PR reports the following “annotations”:

> Error when evaluating 'strategy' for job 'compilers'. .github/workflows/ci.yml (Line: 50, Col: 18): Matrix vector 'package' does not contain any values

Should we be worried? I suppose this might only happen when there are no packages to check, in which case the CI does what is expected. (That is not much.)

toots:
> The [CI on this very PR](https://github.com/ocaml/opam-repository/actions/runs/4942745316) reports the following “annotations”:
> 
> > Error when evaluating 'strategy' for job 'compilers'. .github/workflows/ci.yml (Line: 50, Col: 18): Matrix vector 'package' does not contain any values
> 
> Should we be worried? I suppose this might only happen when there are no packages to check, in which case the CI does what is expected. (That is not much.)

The current setup doesn't account for the empty state, i.e. when there are no changed packages (also I think no revdeps).

I still agree with #20322 decision, having two CIs might be a maintenance burden so I didn't polish that part. However, the execution with actual changed packages are potentially useful at the moment as a substitute.

smorimoto:
I'm concerned that this will put pressure on the CI parallelism of the OCaml organisation and delay the CI of other projects. We've already experienced some decent CI delays for GitHub Actions, and I think the opam-repository has been well-tested with ocurrent, so I'm not sure how much benefit there would be in adding GitHub Actions to this. GitHub has increased the parallelism for us once, but it's still not enough. I think we probably need as much as the Rust team... (Of course, more optimisation, and so on!)

mseri:
Thanks for thinking about this. The CI is back, I am going to close this for now. But I will raise the point at pur next meeting

PR #23799, opened 2023-05-16
Fix bad checksum for ppx_rapper 3.1.0

smorimoto:
@mseri I missed your comment! This should indeed work well!

### DIFF
--- a/packages/ppx_rapper/ppx_rapper.3.1.0/opam
+++ b/packages/ppx_rapper/ppx_rapper.3.1.0/opam
@@ -31,7 +31,7 @@ build: [
 ]
 dev-repo: "git+https://github.com/roddyyaga/ppx_rapper.git"
 url {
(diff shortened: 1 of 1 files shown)


packages/ppx_rapper/ppx_rapper.3.1.0/opam
```diff
@@ -31,7 +31,7 @@ build: [
 ]
 dev-repo: "git+https://github.com/roddyyaga/ppx_rapper.git"
 url {
-  src: "https://github.com/roddyyaga/ppx_rapper/archive/3.1.0.tar.gz"
+  src: "https://raw.githubusercontent.com/ocaml/opam-source-archives/HEAD/ppx_rapper-3.1.0.tar.gz"
```
smorimoto, line 34: The intention is to use links that don't hardcode the name of the default branch, to avoid the same churn that happened with master -> main...

mseri:
Thanks!

PR #23804, opened 2023-05-18
new file:   packages/parany/parany.14.0.0/opam

UnixJunkie:
going back to the fork-plus-marshal based backend; instead of ocaml5 "threads"

added Parany.Parmap.array_pariter and array_pariteri


UnixJunkie:
for any program depending on parany which is broken by this new major version, we could put a < 14.0.0 version constraint

mseri:
The failures seem all related to the fact that alpine tempdir creation does not look at env variables and fails in the sandbox. There does not seem to be any breakage in the revdeps

UnixJunkie:
thanks a lot for the review

PR #23807, opened 2023-05-21
Ocplib-endian uses bytes

samoht:
Without this patch, the compilation (on OCaml5) fails with:

```
  #=== ERROR while compiling ocplib-endian.1.2 ==================================#
  # context     2.1.4 | macos/arm64 |  | https://opam.ocaml.org#f38d24d2
  # path        ~/git/mirage-crypto/_opam/.opam-switch/build/ocplib-endian.1.2
  # command     ~/.opam/opam-init/hooks/sandbox.sh build dune build -p ocplib-endian -j 9 @install
  # exit-code   1
  # env-file    ~/.opam/log/ocplib-endian-989-8ac923.env
  # output-file ~/.opam/log/ocplib-endian-989-8ac923.out
  ### output ###
  # [...]
  # -> required by _build/install/default/lib/ocplib-endian/META
  # -> required by _build/default/ocplib-endian.install
  # File "src/dune", line 75, characters 35-40:
  # 75 |  (libraries ocplib_endian bigarray bytes))
  #                                         ^^^^^
  # -> required by library "ocplib-endian.bigstring" in _build/default/src
  # -> required by _build/default/META.ocplib-endian
  # -> required by _build/install/default/lib/ocplib-endian/META
  # -> required by _build/default/ocplib-endian.install
  ```
  
  /cc @chambart

### DIFF
--- a/packages/ocplib-endian/ocplib-endian.1.2/opam
+++ b/packages/ocplib-endian/ocplib-endian.1.2/opam
@@ -17,6 +17,7 @@ depends: [
   "ocaml" {>= "4.03.0"}
   "cppo" {>= "1.1.0" & build}
   "dune" {>= "1.0"}
+  "base-bytes"
 ]
 build: [
   "dune"

### CODE REVIEW

File: packages/ocplib-endian/ocplib-endian.1.2/opam [hunk 1]
```diff
@@ -17,6 +17,7 @@ depends: [
   "ocaml" {>= "4.03.0"}
   "cppo" {>= "1.1.0" & build}
   "dune" {>= "1.0"}
+  "base-bytes"
```
mseri, line 20: base-bytes is there already. Is that something to do with dune?

mseri: The fix in the past was a cumbersome set of bounds (see https://github.com/ocaml/opam-repository/wiki/PR-checks#some-common-failures) but here it seems that it should have not been a problem

avsm:
See https://github.com/OCamlPro/ocplib-endian/pull/26 -- I pinged @chambart about this a month ago but have not received a response.  We can consider patching this for OCaml 5 directly in opam-repo with a version bump since it is so core to the dependency cone for other packages in the repository.

avsm:
How come this passes the OCaml 5.0 build on check.ocamllabs.io? https://check.ci.ocaml.org/log/1682776300-97d160f899f402d5b0454aa1d3f7880432aed3f7/5.1/good/ocplib-endian.1.2

PR #23813, opened 2023-05-22
Package afl-persistent.1.4

stedolan:
### `afl-persistent.1.4`
Use afl-fuzz in persistent mode
afl-fuzz normally works by repeatedly fork()ing the program being
tested. using this package, you can run afl-fuzz in 'persistent mode',
which avoids repeated forking and is much faster.

---
* Homepage: https://github.com/stedolan/ocaml-afl-persistent
* Source repo: git+https://github.com/stedolan/ocaml-afl-persistent.git
* Bug tracker: https://github.com/stedolan/ocaml-afl-persistent/issues

---

### DIFF
--- a/packages/afl-persistent/afl-persistent.1.4/opam
+++ b/packages/afl-persistent/afl-persistent.1.4/opam
@@ -1,0 +1,38 @@
+opam-version: "2.0"
+maintainer: "stephen.dolan@cl.cam.ac.uk"
+authors: ["Stephen Dolan"]
+homepage: "https://github.com/stedolan/ocaml-afl-persistent"
+bug-reports: "https://github.com/stedolan/ocaml-afl-persistent/issues"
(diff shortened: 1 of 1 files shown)


stedolan:
That's amazing! This is actually the bug in `calendar` that that test originally found, but it took many iterations of `afl-fuzz` to hit it. I assumed it wouldn't actually hit that by doing a few normal quickchecks, but it can find it with a certain small probability and looks like we got (un?)lucky here.

(So yes, feel free to merge, that's not a very robust test in crowbar)

raphael-proust:
Maybe the test run by opam should not fail when the example test does fail. Basically there could be two aliases in the `dune` file of the example with one that just runs the example (which may actually fail) and one that's `@runtest` where it ignores the return code. Something to put in the next release of crowbar?

But that's on the separate package and I think it shouldn't block this PR.

PR #23824, opened 2023-05-25
Prevent install ocamlformat-rpc using a dummy package

Julow:
The RPC feature is now part of the main `ocamlformat` package and this package is no longer useful.

This has been noticed in https://github.com/ocamllabs/vscode-ocaml-platform/issues/1094. Users are instructed to install the package, which has the effect of downgrading ocamlformat.

I initially planned to add a message:
```
  ∗ install   odoc-parser         1.0.1            [required by ocamlformat]
  ↘ downgrade ocaml-version       3.6.1 to 3.5.0   [required by ocamlformat]
  ↘ downgrade ocamlformat-rpc-lib 0.25.1 to 0.21.0
          [required by ocamlformat-rpc]
  ↻ recompile ocamlformat-lib     0.25.1           [uses ocaml-version]
  ↘ downgrade ocamlformat         0.25.1 to 0.21.0
          [required by ocamlformat-rpc]
  ∗ install   ocamlformat-rpc     0.21.0           
          DEPRECATED. This package is not needed and not compatible with
          OCamlFormat >= 0.22.4
===== ∗ 2   ↻ 1   ↘ 3 =====
Do you want to continue? [Y/n] 
```

But this message is not precise (`ocamlformat-rpc.0.21.0` is not deprecated and is useful in combination of `ocamlformat.0.21.0`, if someone wants to use that) and the unwanted outcome is not prevented (the downgrade).

This adds a new `ocamlformat-rpc` version compatible with `ocamlformat >= 0.22.4` that does not install anything and show an information message.

This prevents the default behavior of downgrading `ocamlformat`, which is not the intention:
```
∗ install ocamlformat-rpc removed  DEPRECATED. OCamlFormat ships the RPC mode by default since version
                                   0.22.4. This package is not necessary.

∗ installed ocamlformat-rpc.removed
Done.
```

Though the inverse workflow is not idea as it requires an explicit version constraint:
```
  ↗ upgrade ocamlformat     0.21.0 to 0.25.1
  ∗ install ocamlformat-rpc removed           DEPRECATED. OCamlFormat ships the RPC mode by default since version 0.22.4. This package is not necessary.
===== ∗ 1   ↗ 1 =====
Do you want to continue? [Y/n] 
```

### DIFF
--- a/packages/ocamlformat-rpc/ocamlformat-rpc.removed/opam
+++ b/packages/ocamlformat-rpc/ocamlformat-rpc.removed/opam
@@ -1,0 +1,18 @@
+opam-version: "2.0"
+synopsis: "Auto-formatter for OCaml code (RPC mode)"
+description:
+  "OCamlFormat is a tool to automatically format OCaml code in a uniform style. This package defines a RPC interface to OCamlFormat"
+maintainer: ["OCamlFormat Team <ocamlformat-dev@lists.ocaml.org>"]
+authors: ["Josh Berdine <jjb@fb.com>"]
+homepage: "https://github.com/ocaml-ppx/ocamlformat"
+bug-reports: "https://github.com/ocaml-ppx/ocamlformat/issues"
+depends: []
+build: []
+dev-repo: "git+https://github.com/ocaml-ppx/ocamlformat.git"
+license: ["MIT" "LGPL-2.1-only WITH OCaml-LGPL-linking-exception"]
+messages: [
+  "DEPRECATED. OCamlFormat ships the RPC mode by default since version 0.22.4. This package is not necessary." {installed}
+]
+conflicts: [
+  "ocamlformat" {< "0.22.4"}
+]
--- a/packages/ocamlformat-rpc/ocamlformat-rpc.removed/opam
+++ b/packages/ocamlformat-rpc/ocamlformat-rpc.removed/opam
@@ -11,7 +11,7 @@ build: []
 dev-repo: "git+https://github.com/ocaml-ppx/ocamlformat.git"
 license: ["MIT" "LGPL-2.1-only WITH OCaml-LGPL-linking-exception"]
 messages: [
-  "DEPRECATED. OCamlFormat ships the RPC mode by default since version 0.22.4. This package is not necessary." {installed}
+  "DEPRECATED. OCamlFormat ships the RPC mode by default since version 0.22.4. This package is no longer necessary." {installed}
 ]
 conflicts: [
   "ocamlformat" {< "0.22.4"}
--- a/packages/ocamlformat-rpc/ocamlformat-rpc.removed/opam
+++ b/packages/ocamlformat-rpc/ocamlformat-rpc.removed/opam
@@ -11,7 +11,7 @@ build: []
 dev-repo: "git+https://github.com/ocaml-ppx/ocamlformat.git"
 license: ["MIT" "LGPL-2.1-only WITH OCaml-LGPL-linking-exception"]
 messages: [
-  "DEPRECATED. OCamlFormat ships the RPC mode by default since version 0.22.4. This package is no longer necessary." {installed}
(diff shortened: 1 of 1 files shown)


packages/ocamlformat-rpc/ocamlformat-rpc.removed/opam
```diff
@@ -0,0 +1,18 @@
...
+maintainer: ["OCamlFormat Team <ocamlformat-dev@lists.ocaml.org>"]
+authors: ["Josh Berdine <jjb@fb.com>"]
+homepage: "https://github.com/ocaml-ppx/ocamlformat"
+bug-reports: "https://github.com/ocaml-ppx/ocamlformat/issues"
+depends: []
+build: []
+dev-repo: "git+https://github.com/ocaml-ppx/ocamlformat.git"
+license: ["MIT" "LGPL-2.1-only WITH OCaml-LGPL-linking-exception"]
+messages: [
+  "DEPRECATED. OCamlFormat ships the RPC mode by default since version 0.22.4. This package is not necessary." {installed}
```
mseri, line 14:
```suggestion
  "DEPRECATED. OCamlFormat ships the RPC mode by default since version 0.22.4. This package is no longer necessary." {installed}
```

Julow: Thanks.

packages/ocamlformat-rpc/ocamlformat-rpc.removed/opam
```diff
@@ -0,0 +1,18 @@
...
+maintainer: ["OCamlFormat Team <ocamlformat-dev@lists.ocaml.org>"]
+authors: ["Josh Berdine <jjb@fb.com>"]
+homepage: "https://github.com/ocaml-ppx/ocamlformat"
+bug-reports: "https://github.com/ocaml-ppx/ocamlformat/issues"
+depends: []
+build: []
+dev-repo: "git+https://github.com/ocaml-ppx/ocamlformat.git"
+license: ["MIT" "LGPL-2.1-only WITH OCaml-LGPL-linking-exception"]
+messages: [
+  "DEPRECATED. OCamlFormat ships the RPC mode by default since version 0.22.4. This package is no longer necessary." {installed}
```
avsm, line 14:
```suggestion
  "The `ocamlformat` package ships the RPC mode since version 0.22.4. This package is no longer necessary." {installed}
```

Let's not shout at our users :-)

avsm (commented):
Just a tweak to the comment, but I don't entirely understand why we need this package.

Why not just patch the VSCode plugin to stop prompting users to install the ocamlformat-rpc package?  This should be an invisible detail...

packages/ocamlformat-rpc/ocamlformat-rpc.removed/opam
Julow: Thanks!

Julow:
The VScode plugin contains instructions asking the user to install this package, which is not updated at the moment (https://github.com/ocamllabs/vscode-ocaml-platform/issues/1094). Older documentation might still be around and users might be used to it and might not seek for updated documentation.

This message should appear while installing the deprecated package, which seems like a good place to put it.

avsm:
Why not just remove the offending message from the VSCode plugin?  This PR only seems to add to the packaging confusion (since presumably you will need ocamlformat-rpc if on an older ocamlformat for your project).  /cc @rgrinberg

Julow:
I believe users are install this by manually typing `opam install` and might continue to do so until warned. 

I think the extra annoyance when installing an older version (requires an explicit version or to install `ocamlformat` and `ocamlformat-rpc` at the same time, which also requires an explicit version) is worth it in comparison to automatic downgrades.

mseri:
As a rule of thumb I think we should do both. Remove the message from the plugin and have this transition package to warn the users that go from habit

PR #23825, opened 2023-05-25
Package yuujinchou.5.0.1

favonia:
**This is intended to replace `yuujinchou.5.0.0`; the OPAM file for `yuujinchou.5.0.0` is changed to be unavailable. Please let us know if we failed to follow the protocol to unpublish a package.**

### DIFF
--- a/packages/yuujinchou/yuujinchou.5.0.0/opam
+++ b/packages/yuujinchou/yuujinchou.5.0.0/opam
@@ -1,3 +1,4 @@
+available: false
 opam-version: "2.0"
 synopsis: "A library for hierarchical names and lexical scoping"
 description: """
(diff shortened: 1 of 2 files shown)


mseri:
That's the right way to do it. Vesion 5.0.0 will become uninstallable unless you manually force the installation by an explcit request to ignore the opam constraints

avsm:
is there any particular reason you want 5.0.0 to be actually uninstallable?

In the normal course of events, your users will just get the freshest version (5.0.1), but will have the option to specifically install 5.0.0 if manually requested.  This doesn't require making 5.0.0 unavailable (which is something of a nuclear option)

favonia:
@avsm That's a valid question. The reason is that version 5.0.0 should not exist from the beginning. There's an incompatible, small, but critical API fix introduced by 5.0.1, and version 5.0.0 without the fix doesn't make much sense. It's our oversight. :-( No one should use version 5.0.0. A related point is that a user would have to specify ">= 5.0.1" instead of just ">= 5.0" in their OPAM to get the correct version, because the lower bounds should still make sense. The alternative is to release 6.0.0 instead, but still, no one should use version 5.0.0.

avsm:
Thanks for the clarification! 5.0.1 will be available in the main repo shortly.

PR #23832, opened 2023-05-28
ocamlbuild: add patch to work on windows from fdopen

avsm:
This patch has been lingering externally since 2016 (https://github.com/ocaml/ocamlbuild/pull/70), and Windows builds don't get very far without it any more, and it's blocking development of other upstream Windows ports (https://github.com/ocaml-multicore/eio/pull/530).

So barring strong objections, I'd like to pull this into opam-repository mainline. No version bump required since it only patches Windows, which is broken anyway without this fix.

Source patch from https://github.com/fdopen/opam-repository-mingw/blob/opam2/packages/ocamlbuild/ocamlbuild.0.14.2/opam

cc @dra27 @gasche @patricoferris

### DIFF
--- a/packages/ocamlbuild/ocamlbuild.0.14.2+win/opam
+++ b/packages/ocamlbuild/ocamlbuild.0.14.2+win/opam
@@ -31,7 +31,7 @@ url {
   ]
 }
 extra-source "ocamlbuild-0.14.2.patch" {
-  src: "https://raw.githubusercontent.com/fdopen/opam-repository-mingw/354a87b397856f2a70024c5c83fc5001074935b6/packages/ocamlbuild/ocamlbuild.0.14.2/files/ocamlbuild-0.14.2.patch"
+  src: "https://raw.githubusercontent.com/ocaml-opam/opam-repository-mingw/354a87b397856f2a70024c5c83fc5001074935b6/packages/ocamlbuild/ocamlbuild.0.14.2/files/ocamlbuild-0.14.2.patch"
   checksum: "sha256=a9b7e1829a3304e5a073d8ddea29d3d8272698e93b7e1ee659ae5e31e5cfb6b9"
 }
 patches: "ocamlbuild-0.14.2.patch"
--- a/packages/ocamlbuild/ocamlbuild.0.14.2+win/opam
+++ b/packages/ocamlbuild/ocamlbuild.0.14.2+win/opam
@@ -1,0 +1,38 @@
+opam-version: "2.0"
+synopsis:
+  "OCamlbuild is a build system with builtin rules to easily build most OCaml projects"
+maintainer: "Gabriel Scherer <gabriel.scherer@gmail.com>"
+authors: ["Nicolas Pouillard" "Berke Durak"]
+license: "LGPL-2.0-or-later WITH OCaml-LGPL-linking-exception"
+homepage: "https://github.com/ocaml/ocamlbuild/"
+doc: "https://github.com/ocaml/ocamlbuild/blob/master/manual/manual.adoc"
+bug-reports: "https://github.com/ocaml/ocamlbuild/issues"
+depends: [
+  "ocaml" {>= "4.03"}
+]
+conflicts: [
+  "base-ocamlbuild"
+  "ocamlfind" {< "1.6.2"}
+]
+build: [
+  [make "all"]
+]
+install: [
+  [make "install"]
+  ["mkdir" "-p" "%{lib}%/ocamlbuild"]
+  ["install" "-m" "0644" "META" "%{lib}%/ocamlbuild"]
+]
+dev-repo: "git+https://github.com/ocaml/ocamlbuild.git"
+url {
+  src: "https://github.com/ocaml/ocamlbuild/archive/refs/tags/0.14.2.tar.gz"
+  checksum: [
+    "md5=2f407fadd57b073155a6aead887d9676"
+    "sha512=f568bf10431a1f701e8bd7554dc662400a0d978411038bbad93d44dceab02874490a8a5886a9b44e017347e7949997f13f5c3752f74e1eb5e273d2beb19a75fd"
+  ]
+}
+extra-source "ocamlbuild-0.14.2.patch" {
+  src: "https://raw.githubusercontent.com/fdopen/opam-repository-mingw/354a87b397856f2a70024c5c83fc5001074935b6/packages/ocamlbuild/ocamlbuild.0.14.2/files/ocamlbuild-0.14.2.patch"
+  checksum: "sha256=a9b7e1829a3304e5a073d8ddea29d3d8272698e93b7e1ee659ae5e31e5cfb6b9"
+}
+patches: "ocamlbuild-0.14.2.patch"
+available: os = "win32"


gasche:
I think it would still be nicer to apply fixes upstream instead of patching opam-repository. (This would help people that are not using opam-repository for example; I could see Diskuv building things from source with pre-cooked recipes.)

Some quick remarks:
- the pin of @patricoferris is to a patch https://github.com/patricoferris/ocamlbuild/commit/b72c0dfdd6f1066d50da4b924b8196237696f66b that looks different from https://github.com/ocaml/ocamlbuild/pull/70 to me, it includes more things and some of it is a bit scary. The larger change is the one proposed here, and it includes things that have not been proposed for upstreaming in the past.
- the change duplicates some of the stdlib code into ocamlbuild, we have to think just a minute about license compatibility here (quick browsing indicates that ocamlbuild is LGPL2 while stdlib is LGPL2.1; I think that we would need to upgrade ocamlbuild to 2.1 to do the copy).
- it is not completely obvious to me that non-windows machines are not affected by this patch at all ; given that the patch has probably never been tested on non-windows machine (and not written with upstreaming in mind) I think a review would be useful
- one thing that stalled https://github.com/ocaml/ocamlbuild/pull/70 is the lack of people willing to help with the review; if we decide now that this issue is more important that it seemed (because people want to stop relying on fdopen's repository), maybe we could try to find reviewers again (for example maybe @MisterDA would be willing to lend a hand?).

avsm:
Thanks for the rapid review @gasche.  Note that this PR does not affect non-Windows installations, as the patch is only applied if `{os="win32"}` and otherwise behaviour is entirely the upstream one.

I'll defer your comments on upstreaming the patch for the purposes of merging this PR, as it's important that we unblock the heart of the dependency graph in opam-repo. I do hope @MisterDA or others can shepherd the patch upstream, as it hasn't aged quite as gracefully after lingering for 7 years as a nice glass of scotch might.

@patricoferris might you be able to confirm that this PR removes the need for pin-depends in eio/windows for you?

dra27:
> as it's important that we unblock the heart of the dependency graph in opam-repo

It pains me somewhat (!!), but I think this reason is indeed the argument for making an exception solely for ocamlbuild and incorporating this patch here for the time being. I also agree with @gasche that there are aspects of the patch which are "scary" in terms of blindly upstreaming.

It would of course be good get upstream to a state where the patch isn't needed. That said, it's slightly painful to prioritise it given both ocamlbuild's largely flawed architecture where Windows is concerned (it's quite heavily Cygwin-reliant compared to Dune - I can't remember what the state of its parallel engine is) and its somewhat reduced usage these days.

gasche:
I feel that I am missing context to participate to the discussion. From my uninformed perspective, it is hard to guess why a patch that nobody cared about for 10 years suddenly needs to be urgently applied at the opam-repository level to get the property, if I understand correctly, that all the vendored patches that eio relies on are instead merged upstream (for the clean ones, I suppose) or vendored directly in the opam repository (for the ugly ones).

If at some point someone is interested in upstreaming these changes in ocamlbuild, I think that a nice first step would be to re-split this patch into a series of independent commits, with commit messages explaining what the changes are doing. (Some of the changes here are also included in https://github.com/ocaml/ocamlbuild/pull/70, and there the PR message has some explanations that could be reused.)

avsm:
@gasche wrote:
>  From my uninformed perspective, it is hard to guess why a patch that nobody cared about for 10 years suddenly needs to be urgently applied at the opam-repository level to get the property, 

Because:
- [fdopen's opam-repo is sunsetted now](https://discuss.ocaml.org/t/sunsetting-opam-repository-mingw/11632)
and 
- it takes a miraculous confluence of the build stars to do any OCaml 5 development on Windows right now for a non-expert of that platform. This makes it a little easier.

I've long been a proponent of putting patches like this into opam-repo _with appropriate constraints_ so that they do not affect mainline users, yet help us advance the upstreaming process.  In this case, as @dra27 notes,  this patch allows _other_ Windows development to happen in parallel while its own upstreaming process continues.

patricoferris:
I'll try to answer a few of these remarks as best I can, note I'm very much not a Windows programmer.

Firstly, @gasche you are quite right that the pinned ocamlbuild in my fork and the original PR are not identical. I tried to make this clear in the Eio PR saying "which I think is the essence of this PR https://github.com/ocaml/ocamlbuild/pull/70" -- apologies for the confusion here. 

> might you be able to confirm that this PR removes the need for pin-depends in eio/windows for you?

As far as I could tell (I definitely tried both), the PR does not fix ocamlbuild on Windows and the patch does. That's as far as I went with checking what the patch actually does. I'll take another look.

> it is hard to guess why a patch that nobody cared about for 10 years suddenly needs to be urgently applied at the opam-repository level to get the property

I think this is less about Eio specifically, and more that we care about good Windows support in Eio. With opam-repository-mingw begin discontinued for nearly two years and [opam.2.2.0](https://github.com/ocaml/opam/projects/2) around the corner, I look at this more as helping the "OCaml on Windows" effort more than just Eio. Some popular packages (e.g. `mtime`, `fpath`, `logs` etc.) rely on ocamlbuild and the reverse dependency list is big.

> If at some point someone is interested in upstreaming these changes in ocamlbuild, I think that a nice first step would be to re-split this patch into a series of independent commits, with commit messages explaining what the changes are doing

Yep, that's a reasonable request, I'll try to do that and open a new PR upstream on ocamlbuild.

dra27:
In amongst my pontifications on [Discuss](https://discuss.ocaml.org/t/sunsetting-opam-repository-mingw/11632#what-do-i-do-when-things-are-broken-2) is a little bit of the context (the dates mentioned in it have gone up in flames somewhat with my health...). I think the point is that many people "cared" about the patch, they just didn't know they did 🙂 The patch was automatically available when opam was initialised using the only available distribution of it on Windows. Breaking opam-repository's normal policy on non-upstreamed patches and letting that patch go here would allow a return to that status quo.

I agree that formatting the patches into a commit series is a good idea (see, for example, my doing this with the patches in opam-repository-mingw [for the compiler itself](https://github.com/dra27/ocaml/commits/4.08-patches)) but just to add that that doesn't immediately yield a PR... there's then the decision on whether they're the patches which are wanted. Various of the patches in opam-repository-mingw for the compiler, ocamlfind and ocamlbuild patch those tools in order to work around inadequacies of the build systems of other packages - i.e. rather than carrying patches for _n_ packages in opam-repository-mingw, workaround patches were applied further up the graph. I can see the sense of that when maintaining something which is entirely a fork, but it does mean that the patches necessarily want to go upstream "as is". (that's all opinion, obviously!)

packages/ocamlbuild/ocamlbuild.0.14.2/opam
```diff
@@ -26,8 +26,14 @@ build: [
     "OCAMLBUILD_MANDIR=%{man}%"
     "OCAML_NATIVE=%{ocaml:native}%"
     "OCAML_NATIVE_TOOLS=%{ocaml:native}%"
-  ]
-  [make "check-if-preinstalled" "all" "opam-install"]
+  ] {os != "win32"}
+  [make "check-if-preinstalled" "all" "opam-install"] {os != "win32"}
+  [make "all"] {os = "win32"}
```
dra27, line 31:
```suggestion
  [make "check-if-preinstalled" {os != "win32"} "all" "opam-install" {os != "win32"}]
```
? Although, see below...

packages/ocamlbuild/ocamlbuild.0.14.2/opam
```diff
@@ -26,8 +26,14 @@ build: [
...
+]
+install: [
+  [make "install"] { os = "win32" }
+  ["mkdir" "-p" "%{lib}%/ocamlbuild"] { os = "win32" }
+  ["install" "-m" "0644" "META" "%{lib}%/ocamlbuild"] { os = "win32" }
 ]
```
dra27, line 37: I'm not sure if opam will realise on non-Windows that this package _doesn't_ have an install section (@rjbou?) which technically does impact non-Windows users.

Could we get away with adding an additional package for Windows (say `ocamlbuild.0.14.2+fdopen`) and marking this one as `available : os != "win32"`?

packages/ocamlbuild/ocamlbuild.0.14.2/opam
```diff
@@ -36,4 +42,5 @@ url {
     "md5=2f407fadd57b073155a6aead887d9676"
     "sha512=f568bf10431a1f701e8bd7554dc662400a0d978411038bbad93d44dceab02874490a8a5886a9b44e017347e7949997f13f5c3752f74e1eb5e273d2beb19a75fd"
   ]
-}
\ No newline at end of file
+}
+patches: "ocamlbuild-0.14.2.patch" {os = "win32"}
```
dra27, line 46: The patch is 16K - I think it'd be better to do this using extra-source and reference the commit from the sunset repository instead?
```suggestion
}
extra-source "ocamlbuild-0.14.2.patch" {
  src: "https://raw.githubusercontent.com/ocaml-opam/opam-repository-mingw/354a87b397856f2a70024c5c83fc5001074935b6/packages/ocamlbuild/ocamlbuild.0.14.2/files/ocamlbuild-0.14.2.patch"
  checksum: "sha256=a9b7e1829a3304e5a073d8ddea29d3d8272698e93b7e1ee659ae5e31e5cfb6b9"
}
patches: "ocamlbuild-0.14.2.patch" {os = "win32"}
```

dra27 (commented):
A few technical observations - the main one is referencing the patch file rather than adding it.

dra27 (commented):
A few technical observations - the main one is referencing the patch file rather than adding it.

patricoferris:
Quick update, I managed to get a [working version of the code](https://github.com/patricoferris/ocamlbuild/tree/fdopen-windows%2Brm) which is https://github.com/ocaml/ocamlbuild/pull/70 + https://github.com/ocaml/ocamlbuild/commit/cb12e405ad185ccdae780e9448f1bab39414bfa8 which is part of the patch in opam-repository-mingw.  This seems to fix the file finding problems and the latter commit fixes the hygiene problems (currently described in https://github.com/ocaml/ocamlbuild/issues/316). Hopefully this is a less invasive change that fixes the immediate issues at hand.

avsm:
I've rebased this patch to incorporate all of @dra27's suggestions.  Can easily modify the patch to be @patricoferris', or just use the existing (and long existing) fdopen version.

My preference would be to merge this PR with fdopen's patch, and then for upstream to work through Patrick's simplified version at their own pace.  Am open to other suggestions that move us forward...

packages/ocamlbuild/ocamlbuild.0.14.2/opam
avsm: done in 469a6255cfd0ec96611ac93eedcdb218abd6a699

packages/ocamlbuild/ocamlbuild.0.14.2/opam
dra27: The solver will select the +win variant anyway, but is it worth having `available: os != "win32"` in ocamlbuild.0.14.2/opam just for "belt and braces"?

dra27:
> My preference would be to merge this PR with fdopen's patch, and then for upstream to work through Patrick's simplified version at their own pace. Am open to other suggestions that move us forward...

Agreed - fdopen's patch is already extensively battle-tested... who knows, by the time there's a new release with a simpler patch, we may have Windows package testing up and running 🙂

avsm:
Thanks everyone for the feedback and reviews.  I'll leave this open for a few days to solicit any lingering feedback and testing, and then merge.

packages/ocamlbuild/ocamlbuild.0.14.2+win/opam
```diff
@@ -0,0 +1,38 @@
...
+dev-repo: "git+https://github.com/ocaml/ocamlbuild.git"
+url {
+  src: "https://github.com/ocaml/ocamlbuild/archive/refs/tags/0.14.2.tar.gz"
+  checksum: [
+    "md5=2f407fadd57b073155a6aead887d9676"
+    "sha512=f568bf10431a1f701e8bd7554dc662400a0d978411038bbad93d44dceab02874490a8a5886a9b44e017347e7949997f13f5c3752f74e1eb5e273d2beb19a75fd"
+  ]
+}
+extra-source "ocamlbuild-0.14.2.patch" {
+  src: "https://raw.githubusercontent.com/ocaml-opam/opam-repository-mingw/354a87b397856f2a70024c5c83fc5001074935b6/packages/ocamlbuild/ocamlbuild.0.14.2/files/ocamlbuild-0.14.2.patch"
```
mseri, line 34: should we have this file directly in the tree instead?

avsm: exactly the [opposite review above](https://github.com/ocaml/opam-repository/pull/23832/files#diff-6728a620d9aaae70089e9099543035a7522657d5b5753414e0bc4590b0de41b6) from @dra27 ;-) By holding it externally, we dont hold a fairly large patch in opam-repo, so I think external is better in this case.

dra27: The only downside of external sources is if they disappear (this happened with a very old compiler patches, IIRC), but in general in-tree is not great for patches both because it forces _all_ users to download them and because often the patches end up being applied to more than one package, so they get duplicated on users' machines as well.

dra27: Or if GitHub/git changes the patch files (cf. #23837) but note that this change here isn't affected because we're referencing a patch file in a repo, rather than a commit.

mseri: What if we move it to the opam-source-archive so that at least we control the repo where it comes from?

dra27: Indeed, it's in ocaml-opam/opam-repository-mingw rather than fdopen's. We could move it - I did have in mind many months ago to put a branch for patches in opam-repository itself, but I haven't moved forward with that proposal. ocaml-opam/opam-repository-mingw will hopefully at some point be "archived", but it should never be deleted.

hhugo:
FYI, here are two CI runs failing to install camlbuild.0.14.2+win
https://github.com/ocsigen/js_of_ocaml/actions/runs/5310258265/jobs/9611988249?pr=1471
https://github.com/ocaml-multicore/domain-local-timeout/actions/runs/5298966353/jobs/9591634270

PR #23820, opened 2023-05-24
8 packages from ocaml/opam at 2.1.5

rjbou:
This pull-request concerns:
- `opam-client.2.1.5`: Client library for opam 2.1
- `opam-core.2.1.5`: Core library for opam 2.1
- `opam-devel.2.1.5`: Bootstrapped development binary for opam 2.1
- `opam-format.2.1.5`: Format library for opam 2.1
- `opam-installer.2.1.5`: Installation of files to a prefix, following opam conventions
- `opam-repository.2.1.5`: Repository library for opam 2.1
- `opam-solver.2.1.5`: Solver library for opam 2.1
- `opam-state.2.1.5`: State library for opam 2.1

---
* Bug tracker: https://github.com/ocaml/opam/issues

---
:camel: Pull-request generated by opam-publish v2.2.0

PR #23833, opened 2023-05-28
ocamlgraph.1.8.1: needs autoconf to generate config files

avsm:
spotted in lower bounds for #23820


avsm:
Just lint errors but builds fine, so merging to help advance #23820

PR #23834, opened 2023-05-29
Release v0.16 of Base

dkalinichenko-js:
This PR adds Base v0.16.


dkalinichenko-js:
Feel free to close this PR should it overload the cluster on its own.

dkalinichenko-js:
I don't think so, v0.16 has the same dependencies as v0.15.

avsm:
Looks good -- the upper bounds on other packages stops Base v0.16 being selected too widely.  Onto Core v16!

PR #23836, opened 2023-05-30
Package z3.4.12.2

Hhro:


packages/z3/z3.4.12.2/opam
```diff
@@ -0,0 +1,40 @@
...
+  "ocamlfind" {build}
+  "zarith"
+  "conf-gmp"
+  "conf-python-3" {build}
+  "conf-c++" {build}
+]
+x-ci-accept-failures: [
+  "centos-7" "oraclelinux-7" # C compiler is too old
+]
+synopsis: "Z3 solver"
```
mseri, line 32:
```suggestion
conflicts: [
  "ocaml-option-bytecode-only"
]
synopsis: "Z3 solver"
```

`z3` does not work for bytecode-only compilers:
```
# Testing ocamlc...
# Testing ocamlopt...
#   File "/home/opam/.opam/5.0/.opam-switch/build/z3.4.12.2/scripts/mk_make.py", line 18, in <module>
#     mk_bindings(API_files)
#   File "/home/opam/.opam/5.0/.opam-switch/build/z3.4.12.2/scripts/mk_util.py", line 3066, in mk_bindings
#     check_ml()
#   File "/home/opam/.opam/5.0/.opam-switch/build/z3.4.12.2/scripts/mk_util.py", line 465, in check_ml
#     raise MKException('Failed testing ocamlopt compiler. Set environment variable OCAMLOPT with the path to the Ocaml native compiler. Note that ocamlopt may require flexlink to be in your path.')
# mk_exception.MKException: 'Failed testing ocamlopt compiler. Set environment variable OCAMLOPT with the path to the Ocaml native compiler. Note that ocamlopt may require flexlink to be in your path.'
```

packages/z3/z3.4.12.2/opam
```diff
@@ -0,0 +1,43 @@
...
+depends: [
+  "ocaml" {>= "4.08.0"}
+  "ocamlfind" {build}
+  "zarith"
+  "conf-gmp"
+  "conf-python-3" {build}
+  "conf-c++" {build}
+]
+x-ci-accept-failures: [
+  "centos-7" "oraclelinux-7" # C compiler is too old
```
mseri, line 30:
```suggestion
  "archlinux" # C compiler is too old
```
The other distributions are no-longer tested and the build fails on archlinux due to
```
error: ‘uintptr_t’ does not name a type
```
which is an optional type in C++11 or later and makes me believe that the arch image is shipping an old compiler or one with a peculiar configuration

fangyi-zhou: https://github.com/Z3Prover/z3/issues/6722
Building with GCC 13 causes the problem, I've encountered that with Fedora 38. The patch is https://github.com/Z3Prover/z3/pull/6723

PR #23842, opened 2023-05-31
document the ocaml-option-nnpchecker package

avsm:
from discussion in #23835


kayceesrk:
Thanks @avsm for documenting this. LGTM.

PR #23846, opened 2023-05-31
Add jasmin.2022.09.3

vbgl:


raphael-proust:
There's a CI failure on macos arm64. It seems that the `conf-ppl` package cannot build on that arch probably because the build rule doesn't include the right directories:

```
build: [
  ["sh" "-c"
	"cc test.c -lppl_c -lppl" {os != "freebsd" & os != "openbsd"}
	"cc test.c -I/usr/local/include -L/usr/local/lib -lppl_c -lppl" {os = "freebsd" | os = "openbsd"}]
]
```

We can either mark jasmin as not available on macos (the easy solution) or fix the `conf-ppl` package (more complicated, it happens on arm64 but not amd64, I don't have access to a mac so I can't easily try locally).

Also, the `conf-gmp` package seems to have issues as well.

@vbgl do you want to mark your package as not available on macos?

raphael-proust:
Actually, nevermind, let's not mark this package. We'll just merge this PR if the rest of the CI works. And we'll look at the deps issues separately.

mseri:
The few failures are dependency related and for some of them there are already discussions and PRs ongoing for potential fixes, this can be merged

vbgl:
Thanks!

PR #23852, opened 2023-06-01
Fix for building capnproto on opensuse-tumbleweed.

### LINKED ISSUE #23851: capnproto not building on opensuse-tumbleweed-4.14
As evidenced by https://ocaml.ci.dev/github/ocurrent/ocaml-ci/commit/9408e5f35cbcaad2825777f05772a9ad763e7c88/variant/opensuse-tumbleweed-4.14_opam-2.1#L796-802 capnproto is not building on opensuse-tumbleweed-4.14. 

Suspect that this is because it is being missed by `os-family=suse`

`opam depext --flags` (for opensuse-tumbleweed-4.14) shows:

```
> opam depext --flags
# Depexts vars detected on this system: arch=x86_64, os=linux, os-distribution=opensuse-tumbleweed, os-family=opensuse
```

novemberkilo:
Fixes #23851

### DIFF
--- a/packages/conf-capnproto/conf-capnproto.2/opam
+++ b/packages/conf-capnproto/conf-capnproto.2/opam
@@ -23,6 +23,7 @@ depexts: [
   ["capnp"] {os-distribution = "homebrew" & os = "macos"}
   ["capnproto"] {os-distribution = "macports" & os = "macos"}
   ["capnproto" "libcapnp-devel"] {os-family = "suse"}
+  ["capnproto" "libcapnp-devel"] {os-family = "opensuse"}
   ["capnproto"] {os = "freebsd"}
 ]
 x-ci-accept-failures: [


raphael-proust:
I think we also need to update the os-family/os-distribution wiki page to add info about this… but actually, do you know if there is somewhere a page which keeps this info? Maybe we don't need to replicate it.

raphael-proust:
I fixed the wiki. If anyone knows of a resource which lists those fieldnames, let me know.

raphael-proust:
cancelled all the jobs of the CI and restarted only the opensuse jobs

tmcgilchrist:
I feel that the root problem is a bug in opam with parsing the `/etc/os-release` fields because the order of those has swapped between `opensuse-leap` and `opensuse-tumbleweed`. But this change will help until a fix is done for https://github.com/ocaml/opam/issues/5565.

tmcgilchrist:
Link to the [wiki definitions](https://github.com/ocaml/opam-repository/wiki/Depexts-os-distribution---os-family-values).

PR #23856, opened 2023-06-01
Relax metrics-unix constraint on mtime

patricoferris:
I think this constraint can be removed, I tried it out locally and metrics-unix only makes use of `Mtime_clock.elapsed_ns`, `Mtime_clock.now` and `Mtime.to_uint64_ns` which I believe are all unchanged across versions of Mtime.


mseri:
This has already run in a number of tests. Thanks

PR #23863, opened 2023-06-02
[new release] ssl (0.6.0)

anmonteiro:
Bindings for OpenSSL

- Project page: <a href="https://github.com/savonet/ocaml-ssl">https://github.com/savonet/ocaml-ssl</a>

##### CHANGES:

- Raise an error when `Ssl.flush` isn't successful (savonet/ocaml-ssl#104, savonet/ocaml-ssl#120)
- Add an API-compatible `Ssl.Runtime_lock` module. The functions in this module
  don't release the OCaml runtime lock. While they don't allow other OCaml
  threads to run concurrently, they don't perform any copying in the underlying
  data, leading certain workloads to be faster than their counterparts that
  release the lock. (savonet/ocaml-ssl#106)
- Guarantee `Ssl.output_string` writes the whole string by retrying the
  operation with unwritten bytes (savonet/ocaml-ssl#103, savonet/ocaml-ssl#116)
- Fix calls in C stubs that need to call `ERR_clear_error` before the underlying
  OpenSSL call (savonet/ocaml-ssl#118)
- Add a module `Ssl.Error` to retrieve OpenSSL errors in a structured way (savonet/ocaml-ssl#119)
- Deprecate Ssl.{SSLv23,SSLv3,TLSv1,TLSv1_1}, which were were formally
  deprecated in March 2021 and earlier (savonet/ocaml-ssl#115).

### DIFF
--- a/packages/ssl/ssl.0.6.0/opam
+++ b/packages/ssl/ssl.0.6.0/opam
@@ -13,7 +13,7 @@ run-test: [
 ]
 depends: [
   "ocaml" {>= "4.02.0"}
-  "dune" {>= "2.0.7"}
+  "dune" {>= "2.7"}
   "dune-configurator"
   "base-unix"
   "conf-libssl"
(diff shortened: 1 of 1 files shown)


anmonteiro:
This is the last release of `ocaml-ssl` supporting OpenSSL 1.0.2 and 1.1.0, which is EOL as per https://www.openssl.org/source/ 


> All older versions (including 1.1.0, 1.0.2, 1.0.0 and 0.9.8) are now out of support and should not be used.

anmonteiro:
Will resubmit with 4.03 support.

PR #23872, opened 2023-06-04
Package errpy.0.0.8

jasontatton:
### `errpy.0.0.8`
Errpy: An Error Recovering Python Parser implemented in Rust

---
* Source repo: git+https://github.com/facebook/errpy.git

---


avsm:
This now requires a really recent rustc, which isn't packaged on any of the CI distros (including debian-unstable).  Do you definitely need something this recent?  We do need to do some work in opam-repo on conf-rustc to help bring in newer versions, but it's not there yet.

jasontatton:
I think it's going to be a bit tricky to remove the newer rustc dependency as this is needed by the `tree-sitter` dependency. Possibly we can downgrade to an earlier version of `tree-sitter` though this would add some of it's own regressions. A promising approach could be vendoring in the whole library and removing the newer rustc dependency but this would only work if there were no code changes dependant upon it.

Do you know how much work it would be to resolve the debian distro problem, is this something I could help with?

avsm:
> Do you know how much work it would be to resolve the debian distro problem, is this something I could help with?

To meaningfully test these, we need an opam package that bring in the right version of rustc. Are there some cargo commands or equivalent that can bring up a specific version of rustc?

mseri:
Yes, but they don't work in the sandbox. The error is clear enough that users can install the correct rust manually if they need the library.

In the past, for these libraries we accepted to merge knowing that they will not be tested on the whole matrix (we have the `x-ci-accept-failure` for these cases). In this case it works fine on many oses and fails with a clear error in the others, so I would consider this good to merge.

avsm:
I'm ok with merging this, but we can't continue on forever without testing Rust packages.  One approach would be to create an `opam-rust-repository` which contains source-level packages for the various rustc versions, and then depend on that.

mseri:
We are testing it though. Only the revdeps are missing. As a compromise we could consider making the revedps image configurable (I think was one of the suggestions in the past).

In any case, I agree that we should agree on a reasonable policy and implement whatever solution that requires. We should probably move that discussion from the PR to an issue or some other space.

We have the same issue with libraries containing bindings to R, some bindings to python, and some C bindings. What we discuss/decide should probably address them as well.

avsm:
I don't think we are testing errpy in this PR, are we?  The forward checks are either skipped or failing due to a too-old rustc.  Agreed on the need for figuring out this for Python/R/etc too...

mseri:
I don't understand why not. Fedora, ubuntu, arch, macos all work also with the `with-test` flag: https://toxis.caelum.ci.dev/github/ocaml/opam-repository/commit/80f3afd1cd832d726c6c3a74bad8df96c545b2b5/variant/distributions,fedora-36-ocaml-4.14,errpy.0.0.8,tests

jasontatton:
Awesome, thanks everyone!

PR #23876, opened 2023-06-04
mirage doesn't work with dune 3.8.0

hannesm:
see https://github.com/ocaml/dune/issues/7875 for further information. The compilation of mirage itself works nicely, but cross-compiling a unikernel that somewhere uses ppx does not work at all.

### DIFF
--- a/packages/mirage/mirage.4.0.0/opam
+++ b/packages/mirage/mirage.4.0.0/opam
@@ -19,7 +19,7 @@ build: [
 
 depends: [
   "ocaml" {>= "4.08.0"}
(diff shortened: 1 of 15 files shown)


avsm:
Oh dear... merging this as the constraints look straightforward.  I guess none of the mirage package tests in the opam world do a full end-to-end build, so the revdeps checks in CI didn't catch this.

hannesm:
Though, unfortunately with opam-monorepo in the game, I'm not sure whether this actually works (since opam-monorepo seems to always want to use the latest dune version -- see the Mirage CI at https://github.com/mirage/mirage/pull/1423)

PR #23881, opened 2023-06-05
conf-fts: fix package name for alpine 3.17

emillon:
In 3.17, `fts.h` is available from `musl-fts-dev` instead of `fts-dev`.


avsm:
Thanks!

PR #23893, opened 2023-06-07
Exclude `mtime.1.4.0` for `metrics-unix`

metanivek:
This specific version of `mtime` fails to build, but is now fixed in `mtime.2.0.0`.

See https://github.com/dbuenzli/mtime/issues/42

I just noticed this as `irmin` builds have started failing (after https://github.com/ocaml/opam-repository/pull/23856 merged) with the error in the above issue.

```
Error: Files /home/opam/.opam/5.0/lib/metrics-unix/metrics_gnuplot.cmxa
       and /home/opam/.opam/5.0/lib/mtime/clock/os/mtime_clock.cmxa
       make inconsistent assumptions over interface Mtime_clock
```

### DIFF
--- a/packages/metrics-unix/metrics-unix.0.4.0/opam
+++ b/packages/metrics-unix/metrics-unix.0.4.0/opam
@@ -18,7 +18,7 @@ depends: [
   "dune" {>= "1.4"}
   "uuidm" {>= "0.9.6"}
   "metrics" {= version}
-  "mtime" {>= "1.0.0"}
+  "mtime" {>= "1.0.0" & != "1.4.0"}
   "lwt" {>= "2.4.7"}
   "metrics-lwt" {= version & with-test}
   "conf-gnuplot"


mseri:
A number of jobs has passed, merging this will prevent the metrics-unix failures in JS and other PRs

hannesm:
I'm not entirely sure about this -- to me: mtime 2.0.0 got released, where depending on mtime.clock wasn't a good idea anymore (that's why ea8053b5902b7a46e30e3c13b547ac94597ba561 introduced bounds).

Then, https://github.com/ocaml/opam-repository/pull/23856 "relaxed" the bounds, and suddenly "irmin" fails to build.

Now. this PR adds some conflicts, with the expectation that earlier & newer mtime will work?

In the meantime, https://github.com/mirage/metrics/pull/58 was proposed and merged (which from my point of view looks like the right thing to do with mtime 2.x).

As community service, I opened https://github.com/ocaml/opam-repository/pull/23895 which includes the fix mentioned above, and should be fine with any mtime release.

PR #23899, opened 2023-06-08
Resto-cohttp-server is incompatible with cohttp-lwt.5.1.0

pirbo:
Because of mirage/ocaml-cohttp#982, a file descriptor is leaked for each never ending streaming RPC opened because the connection is never closed.

### DIFF
--- a/packages/resto-cohttp-server/resto-cohttp-server.0.5/opam
+++ b/packages/resto-cohttp-server/resto-cohttp-server.0.5/opam
@@ -13,7 +13,7 @@ build: [
 
 depends: [
   "ocaml" { >= "4.07" }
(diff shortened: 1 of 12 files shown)


packages/resto-cohttp-server/resto-cohttp-server.0.5/opam
```diff
@@ -16,7 +16,7 @@ depends: [
   "dune" { >= "1.7" }
```
mseri, line 16:
```suggestion
  "dune" { >= "1.11" }
```

mseri:
I fixed the linter error. With this it is fine to merge, thanks

PR #23906, opened 2023-06-10
[new release] tracy-client (0.1)

c-cube:
Client bindings to the Tracy profiler (v0.9.1)

- Project page: <a href="https://github.com/imandra-ai/ocaml-tracy">https://github.com/imandra-ai/ocaml-tracy</a>

##### CHANGES:

initial release

### DIFF
--- a/packages/tracy-client/tracy-client.0.1/opam
+++ b/packages/tracy-client/tracy-client.0.1/opam
@@ -1,0 +1,37 @@
+opam-version: "2.0"
+synopsis: "Client bindings to the Tracy profiler (v0.9.1)"
+maintainer: ["Simon Cruanes"]
+authors: ["Bartosz Taudul" "Simon Cruanes"]
+license: "BSD-3-Clause"
+homepage: "https://github.com/imandra-ai/ocaml-tracy"
+bug-reports: "https://github.com/imandra-ai/ocaml-tracy/issues"
+depends: [
+  "dune" {>= "2.7"}
+  "ocaml" {>= "4.08"}
+  "trace"
+  "odoc" {with-doc}
+]
+build: [
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
+dev-repo: "git+https://github.com/imandra-ai/ocaml-tracy.git"
+url {
+  src:
+    "https://github.com/imandra-ai/ocaml-tracy/releases/download/v0.1/tracy-client-0.1.tbz"
+  checksum: [
+    "sha256=287fa7ad450f43951925ac119ad6b3b1e485a3942652e5bf9b483e49800ae772"
+    "sha512=72152274402cd2d2ad0c7e3ba0b3896c0bc1ab92d021460da94a2f9fe634759b2021d70d8175b3269aa6197b7917449007e96b53a8ccd8d534b50debd1b5162b"
+  ]
+}
+x-commit-hash: "74116b30dad14747f8f3ded01d582e3b991e4994"


c-cube:
Only alpine fails, it seems like it doesn't have the right headers installed. Given that this also requires Tracy to be installed to be useful, I think it's fair to ask the user to fix it themselves.

abhsrivastava:
Looks good to me. consider for merge.

packages/tracy-client/tracy-client.0.1/opam
```diff
@@ -0,0 +1,37 @@
...
+authors: ["Bartosz Taudul" "Simon Cruanes"]
+license: "BSD-3-Clause"
+homepage: "https://github.com/imandra-ai/ocaml-tracy"
+bug-reports: "https://github.com/imandra-ai/ocaml-tracy/issues"
+depends: [
+  "dune" {>= "2.7"}
+  "ocaml" {>= "4.08"}
+  "trace"
+  "odoc" {with-doc}
+]
```
avsm, line 13:
```suggestion
]
depexts: [
  ["linux-headers"] {os-distribution = "alpine"}
]
```

packages/tracy-client/tracy-client.0.1/opam
```diff
@@ -0,0 +1,40 @@
...
+synopsis: "Client bindings to the Tracy profiler (v0.9.1)"
+maintainer: ["Simon Cruanes"]
+authors: ["Bartosz Taudul" "Simon Cruanes"]
+license: "BSD-3-Clause"
+homepage: "https://github.com/imandra-ai/ocaml-tracy"
+bug-reports: "https://github.com/imandra-ai/ocaml-tracy/issues"
+depends: [
+  "dune" {>= "2.7"}
+  "ocaml" {>= "4.08"}
+  "trace"
```
haochenx, line 11:
```suggestion
  "trace" {>= "0.1"}
```

please consider adding a lower bound

packages/tracy-client/tracy-client.0.1/opam
```diff
@@ -0,0 +1,40 @@
...
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
+dev-repo: "git+https://github.com/imandra-ai/ocaml-tracy.git"
+url {
+  src:
+    "https://github.com/imandra-ai/ocaml-tracy/releases/download/v0.1/tracy-client-0.1.tbz"
```
haochenx, line 34: I tried to download the tarball, and it seems that it contains two identical physical copies of tracy distribution (one under vendors/tracy, and another under src/bindings/tracy)

This is probably a non-blocker but imho worthy to look into.

c-cube:
Oh, damn, it must expand the symlink 😱. I need to see how to coerce dune-release into only expanding the one in src then.

avsm:
Good catch @haochenx!  This could create a problem if the release is vendoered in a monorepo... do you want to respin the release @c-cube? (I also fixed the Alpine issue with a depext)

packages/tracy-client/tracy-client.0.1/opam
c-cube: it's the lowest version, is there an issue?

c-cube:
I can try to redo a release. Right now dune is driving me nuts while I'm trying to have it copy a directory from vendor/ to src/some/path :scream: .

c-cube:
> This could create a problem if the release is vendoered in a monorepo

I don't think so, I assume the duplicated code is the vendored C++ code which would be compiled and installed only once.

packages/tracy-client/tracy-client.0.1/opam
haochenx: Oh, indeed. I was following opam-repository's [PR review guideline](https://github.com/ocaml/opam-repository/wiki/PR-checks).

I guess it's not going to cause a problem in this case in general, but there is a technical possibility when an incompatible version with a lower version number (e.g. `0.0.1`?) is published on OPAM. And I guess it's good to add it here so we do not have to double check whether it indeed depends on the lowest version.

(Maybe I'm missing something like if the lower bound is not supplied CI assume the lowest version for that specific package as the lowest bound. In this case it is indeed non-necessary in general)

Actually I think I need help from a more experienced maintainer of opam-repository to settle this.
@avsm can you take a look at this?

haochenx:
>I don't think so, I assume the duplicated code is the vendored C++ code which would be compiled and installed only once.

Yeah, that's the reason I didn't think it's a blocker for this PR. (I should have made myself clearer on the reason..)

I think if @c-cube can figure out a quick way to prevent the code duplication it's good to have the release updated. But if not, (and other opam-maintainers are also okay with this) the only downside is inflated tarball size (9.7M vs 5M) and I believe the package will install without problem.

packages/tracy-client/tracy-client.0.1/opam
haochenx: Great. In this case, everything looks good to me.

haochenx: Just out of curiosity: does the CI assume the lowest version available for a package when doing the lower-bounds test when the lower bound is not explicitly specified in the OPAM file? @mseri

mseri: The lowest co-installable

mseri: When it happens that a release breaks some lower bounds, then we go and we fix them. I think so far it only happened once with a widespread package being ported to ocaml 5 or with old packages from before the lower bounds test was added

avsm: Thanks @haochenx, keeping those guidelines uptodate is really important, so appreciate you clarifying them as well.  Note that there is absolutely no harm in putting an explicit lower bound even if the package is the lowest one -- this guards against the (unlikely) situation where a package is pinned at a lower version than the previous lowest revision.

avsm:
Ok, let's go ahead and merge this one as-is, and trust @c-cube to sort out future releases!

c-cube:
Thank you! And thanks @haochenx for the remarks :). I'll keep that in mind for future releases :)

PR #23908, opened 2023-06-11
Petrol 1.2.0 links to wrong src tag.

oemmerson:
Was linking to 1.2.3.


Gopiandcode:
I see. Thanks for bringing this to my attention! Ah, my bad. I agree with what you've proposed @mseri, changing it might break packages expecting the 1.2.3 release. I propose we keep this the same, and I'll be releasing a new version in the coming weeks that appropriately tags the version.

mseri:
Sounds good to me. I will close both PRs then

PR #23916, opened 2023-06-13
hacl-star-raw requires ocamlopt

emillon:
This is declared in the latest releases but old versions are affected too. By conflicting with ocaml-option-bytecode-only we prevent installation on bytecode-only architectures including ocaml 5.


emillon:
(this doesn't need the full revdeps to go through)

mseri:
Seen on #23866

PR #23927, opened 2023-06-14
[new release] rangeSet (0.2.1)

kandu:
# RangeSet #

## Sets over ordered ranges

  This module implements two range set data structure: continuous and discrete range set.

  The elements are represented as ranges internally, and is therefore efficient for large set of adjacent elements.

## Document
  The document is available [here](https://kandu.github.io/caml_doc/rangeSet/RangeSet/index.html).

### DIFF
--- a/packages/rangeSet/rangeSet.0.2.1/opam
+++ b/packages/rangeSet/rangeSet.0.2.1/opam
@@ -1,0 +1,24 @@
+opam-version: "2.0"
+maintainer: "zandoye@gmail.com"
+authors: [ "ZAN DoYe" ]
+homepage: "https://github.com/kandu/rangeSet/"
+bug-reports: "https://github.com/kandu/rangeSet/issues"
+license: "MIT"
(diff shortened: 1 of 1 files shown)


packages/rangeSet/rangeSet.0.2.1/opam
```diff
@@ -0,0 +1,24 @@
...
+license: "MIT"
+dev-repo: "git+https://https://github.com/kandu/rangeSet"
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+]
+depends: [
+  "ocaml" {>= "4.02.3"}
+  "dune" {>= "1.1.0"}
+]
+synopsis: "rangeSet"
```
mseri, line 15:
```suggestion
synopsis: "RangeSet: a library for sets over ordered ranges"
```

mseri: I think the synopsis should be more descriptive

PR #23928, opened 2023-06-14
dns-client: restrict happy eyeballs for preparation of a new release

hannesm:
this change does not do anything in the current opam-repository state, but there's an upcoming API breaking change. I checked the other users of happy-eyeballs, and they should be fine.


mseri:
thanks

PR #23931, opened 2023-06-14
removing package cosovo, as it is inferior to package 'csv'

barko:


mseri:
We have a policy of not removing packages unless there are major bugs (in which case we don't remove them but mark them as unavailable). In this case this does not seem the case, there may be users of the library whose workflow would break for this removal. What is wrong in keeping the library in the repository?

avsm:
Unfortunately, it's not possible to figure out if there no other downstream dependencies, since opam-repositories can be layered and so someone else might be depending on this package without us knowing.  We'd really prefer that you leave this package as-is (unless there is something really urgent like a big security issue), and just change the description to mark that it's deprecated.

barko:
Let's go with your preference! I'll close out this PR.

PR #23941, opened 2023-06-16
magic-trace is incompatible with core.v0.16

dkalinichenko-js:


avsm:
Thanks; looks from the CI failure that it might also be Linux-only -- you might want to do a followup PR with `os = "linux"` in the availability section as well, if that's the case.

PR #23943, opened 2023-06-16
Update conf-trexio

scemama:
pkg-config expects `trexio` and not `libtrexio`

### DIFF
--- a/packages/conf-trexio/conf-trexio.0.1/opam
+++ b/packages/conf-trexio/conf-trexio.0.1/opam
@@ -5,7 +5,7 @@ homepage: "http://trex-coe.github.io/trexio"
 bug-reports: "https://github.com/trex-coe/trexio/issues"
 license: "BSD-3-Clause"
 build: [
-  ["sh" "-c" "cc test.c $(pkg-config --libs --cflags libtrexio)"]
+  ["sh" "-c" "cc test.c $(pkg-config --libs --cflags trexio)"]
 ]
 depends: ["conf-pkg-config" {build}]
 


packages/conf-trexio/conf-trexio.0.1/opam
```diff
@@ -5,7 +5,7 @@ homepage: "http://trex-coe.github.io/trexio"
 bug-reports: "https://github.com/trex-coe/trexio/issues"
 license: "BSD-3-Clause"
 build: [
-  ["sh" "-c" "cc test.c $(pkg-config --libs --cflags libtrexio)"]
+  ["sh" "-c" "cc test.c $(pkg-config --libs --cflags trexio)"]
 ]
 depends: ["conf-pkg-config" {build}]
```
mseri, line 11: Is libtrexio packaged by distrubutions? If so, could you please add the necessary depexts? See https://github.com/ocaml/opam-repository/wiki/Depexts-os-distribution---os-family-values and another conf package as an example

scemama: Thanks to both of you for your comments.

No, trexio is not yet packaged for distributions. There is a package sento to Debian, but it is not yet available.


After seeing that the CI fails, I googled a bit and I saw that ocaml-yaml provides the C library with the OCaml binding. The only dependence Trexio is HDF5, which can be de-activated at configure time, so I believe the most robust fix would be to compile libtrexio with the ocaml binding in the ocaml trexio package, and remove the conf-trexio package.

1. In your opinion, should I keep conf-trexio or should I compile libtrexio in the ocaml package?
2. If you advise to compile the library, can you give me some links to other packages compiling a C library? My knowledge of dune and opam is quite limited so examples will help me a lot.

mseri: If this package is just there to make sure that the library is installed, I think we can wait to add it until there is a distribution packaging it or add a failure message asking the user to install libtrexio manually.

If you have a package that will depend on this you could build the library as part of the package as done by ocaml-yaml (I think it is the bext example to consider for this case) and then you don't need the conf package

scemama: > If this package is just there to make sure that the library is installed, I think we can wait to add it until there is a distribution packaging it or add a failure message asking the user to install libtrexio manually.

The failure message is already present :-)
```
<><> conf-trexio.0.1 troubleshooting ><><><><><><><><><><><><><><><><><><><><><>
=> This package requires installation of the trexio library: https://github.com/TREX-CoE/trexio/releases/
```

> If you have a package that will depend on this you could build the library as part of the package as done by ocaml-yaml (I think it is the bext example to consider for this case) and then you don't need the conf package

OK. I will go in this direction then and remove the conf-trexio package.

Thanks for your help!

PR #23949, opened 2023-06-18
add reason upper bound for melange 0.2.0

anmonteiro:
caught by https://github.com/ocaml/opam-repository/pull/23948


avsm:
Thanks!

PR #23964, opened 2023-06-20
Add afl-2.57b package.

tmcgilchrist:
This is the last official release from https://github.com/google/AFL/releases.

It's not clear how I can test this locally within opam. I have applied the patch manually to the project source and that works ok. Submitting here to get testing on various platforms.

### DIFF
--- a/packages/afl/afl.2.57b/opam
+++ b/packages/afl/afl.2.57b/opam
@@ -1,0 +1,31 @@
+opam-version: "2.0"
+maintainer: "meetup@yomimono.org"
+homepage: "https://github.com/google/AFL"
+dev-repo: "git+https://github.com/google/AFL.git"
+bug-reports: "https://groups.google.com/forum/#!forum/afl-users"
+license: "Apache-2.0"
+build: [
+  [make] {arch = "x86_64"}
+  [make "AFL_NO_X86=1"] {arch != "x86_64"}
+]
+install: [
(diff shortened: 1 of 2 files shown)


avsm:
Just restarting a failed macos/arm build (SIGBUS on git reset?!), but otherwise looks fine.

packages/afl/afl.2.57b/opam
```diff
@@ -0,0 +1,31 @@
...
+remove: [
+  [make "PREFIX=%{prefix}%" "uninstall"]
+]
+patches: [
+  "add-uninstall-target.patch"
+]
+synopsis:
+    "American Fuzzy Lop fuzzer by Michal Zalewski, repackaged for convenient use in opam"
+authors: "Michal Zalewski"
+depends: ["ocaml"]
```
mseri, line 24: I don't understand this, the package only depends on a c compiler, no?

avsm: Woops, I merged before seeing this. I think @mseri is right...

tmcgilchrist: Yes it only depends on a c compiler. 
What's the correct way to depend on that? @mseri

mseri: I think it is `(conf-gcc | conf-clang)` but I am not 100% sure. If there is ocaml, most likely there is also the c compiler, so even though it is not strictly required perhaps is a good idea to keep it as is

PR #23967, opened 2023-06-20
[new release] tar, tar-unix and tar-mirage (2.5.1)

reynir:

##### CHANGES:

- Treat headers with link indicator '0' or '\000' (`Normal`) as directories for backward compatibility (reported in mirage/ocaml-tar#129, fix by @reynir)

### DIFF
--- a/packages/tar-mirage/tar-mirage.2.5.1/opam
+++ b/packages/tar-mirage/tar-mirage.2.5.1/opam
@@ -1,0 +1,61 @@
+opam-version: "2.0"
+synopsis: "Read and write tar format files via MirageOS interfaces"
+description: """
+tar is a simple library to read and write tar files with an emphasis on
+streaming.  This library is functorised over external OS dependencies
+to facilitate embedding within MirageOS.
+"""
+maintainer: ["dave@recoil.org"]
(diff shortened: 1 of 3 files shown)


reynir:
The test failure on Mac is not fatal. I have a fix in https://github.com/mirage/ocaml-tar/commit/95bd151a9237db1f8b7b4339245139ccd6f7886d - not sure if I should re-publish the package with that patch.

mseri:
If you already have a fix, it would be nice to have the fixed version so that we can leave the tests on without affecting the logs

packages/tar-mirage/tar-mirage.2.5.1/opam
```diff
@@ -0,0 +1,61 @@
...
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "--promote-install-files=false"
+    "@install"
+    "@runtest" {with-test}
```
mseri, line 47:
```suggestion
    "@runtest" {with-test & os != "macos"}
```
Otherwise I would suggest to disable the tests on macos

reynir:
Ok, thank you for the feedback. I will resubmit with the patch.

PR #23970, opened 2023-06-20
Add Z3 4.12.2 compilation fix for GCC 13

fangyi-zhou:


mseri:
The patch is included upstream: https://github.com/Z3Prover/z3/pull/6723
This is a strict improvement, so I am going to merge, thanks!

PR #23990, opened 2023-06-23
[new release] duff (0.5)

dinosaure:
Rabin's fingerprint and diff algorithm in OCaml

- Project page: <a href="https://github.com/mirage/duff">https://github.com/mirage/duff</a>
- Documentation: <a href="https://mirage.github.io/duff/">https://mirage.github.io/duff/</a>

##### CHANGES:

- Require OCaml 4.07 and remove compatibility packages (@hannesm, mirage/duff#11)
- Fix the compilation on OCaml 5.1 (@dinosaure, mirage/duff#12)

### DIFF
--- a/packages/duff/duff.0.5/opam
+++ b/packages/duff/duff.0.5/opam
@@ -30,8 +30,8 @@ depends: [
 url {
   src: "https://github.com/mirage/duff/releases/download/v0.5/duff-0.5.tbz"
   checksum: [
-    "sha256=d1eaa97cf58d38762391c8d74676544d4145d3f2fd13e4cda0ea8a0844cde612"
-    "sha512=f66ba0016327236c6bdd29cbd3c3d62b96dbe8aa5cdbc88be956a6152c1fd1dcb70ec1f7598bcf4b56126f0edb127566617f49898e50b412be2fecbe93ce85d4"
+    "sha256=f9453cf4aa3b6850efe8cc6f13f053ebe5f2111faf177ceabfbb03e5d9ad6ede"
+    "sha512=a063a549af860db2e66be5ec0c4bfb827e1fe7284c669562455cdd5ae5fa760cbeb878663ceadf93a8d7350f7eab926740fdef2e7bf7eaf2c2e605ec532eb70e"
   ]
 }
-x-commit-hash: "20259ac3021f9675f12972e04871a45f6651d483"
+x-commit-hash: "c82285ecc147ca5713d65c5c04a994a2245398a4"


moyodiallo:
Why rewriting on the same version instead of new one? I'm getting this error when I'm building the Docker file on [solver-service](https://github.com/ocurrent/solver-service).

```
> git clone git@github.com:ocurrent/solver-service.git
> docker build .
...
#0 35.88 [ERROR] The sources of the following couldn't be obtained, aborting:
#0 35.88           - duff.0.5: Bad checksum
#0 35.88 
....
```

tmcgilchrist:
Are the installation failures of cca a problem? @dinosaure

dinosaure:
Oh sorry I believed that `duff` was not released - I saw some failures about `duff` on `ocaml-ci` the last week and I was not sure that I did the release of this one. However, I replaced the tag on my repository. So this checksum is the good one (the only one I mean). About `cca`, I think the issue is unrelated to this PR.

mseri:
Unfortunately the old source is not in the cache

hannesm:
FWIW https://opam.robur.coop/cache/sha256/d1/d1eaa97cf58d38762391c8d74676544d4145d3f2fd13e4cda0ea8a0844cde612

mseri:
@dinosaure can you please get the old tarball from the cache above and restore the tag, and release this as new version?

hannesm:
So, I downloaded and extracted the old tarball and the new tarball, with the following diff:
```OCaml
diff -ur old/duff-0.5/CHANGES.md duff-0.5/CHANGES.md
--- old/duff-0.5/CHANGES.md     2022-04-08 16:10:55.000000000 +0200
+++ duff-0.5/CHANGES.md 2023-06-23 16:25:40.000000000 +0200
@@ -1,9 +1,8 @@
-v0.5 2022-04-08 Paris (France)
+v0.5 2023-06-23 Paris (France)
 ------------------------------------
 
-* `duff` works only for OCaml >= 4.07 (@hannesm, #11)
-* Update with `ocamlformat.0.21.0` (@hannesm, #11)
-* Remove the `bigarray-compat` dependency (@hannesm, #11)
+- Require OCaml 4.07 and remove compatibility packages (@hannesm, #11)
+- Fix the compilation on OCaml 5.1 (@dinosaure, #12)
 
 v0.4 2021-15-03 Paris (France)
 ------------------------------------
diff -ur old/duff-0.5/lib/duff.ml duff-0.5/lib/duff.ml
--- old/duff-0.5/lib/duff.ml    2022-04-08 16:10:55.000000000 +0200
+++ duff-0.5/lib/duff.ml        2023-06-23 16:25:40.000000000 +0200
@@ -666,27 +666,27 @@
   let entries = ref max in
 
   while !idx >= 0 do
-    let hash = hash source (!idx + 1) (bigstring_length source) in
+    let hashv = hash source (!idx + 1) (bigstring_length source) in
 
-    if Uint32.compare hash !previous = 0
+    if Uint32.compare hashv !previous = 0
     then (
       unpacked.(!rev - 1) <-
         { (unpacked.(!rev - 1)) with offset = !idx + _window } ;
       decr entries (* keep the lowest consecutive indentical blocks *))
     else (
-      previous := hash ;
+      previous := hashv ;
 
       unpacked.(!rev) <-
         ({
            offset = !idx + _window;
-           hash;
-           next = htable.(Uint32.(to_int (hash land hmask)));
+           hash = hashv;
+           next = htable.(Uint32.(to_int (hashv land hmask)));
          }
           : unpacked_entry) ;
 
-      htable.(Uint32.(to_int (hash land hmask))) <- Entry !rev ;
-      hcount.(Uint32.(to_int (hash land hmask))) <-
-        hcount.(Uint32.(to_int (hash land hmask))) + 1 ;
+      htable.(Uint32.(to_int (hashv land hmask))) <- Entry !rev ;
+      hcount.(Uint32.(to_int (hashv land hmask))) <-
+        hcount.(Uint32.(to_int (hashv land hmask))) + 1 ;
       rev := !rev + 1) ;
 
     idx := !idx - _window
```

(Which is exactly https://github.com/mirage/duff/pull/12/commits/dee5d7282e0ff5deead0900887dc30b0207ffee9). Thus, I'm very fine with just merging this PR, and finally have all the CI systems that complain about a bad checksum be happy.

The altarnative is:
- restore the old tarballs somewhere (e.g. ocaml/opam-source-archives -- but I'm really not convinced that storing tarballs in a git repository is the path to go)
- edit the duff 0.5 opam metadata to point to the old tarballs
- adapt the version number in this PR to duff 0.5.1

--> sounds like a bit too much work for no gain, or is there something I'm missing?

mseri:
Mostly that the policy is to have the repo as immutable as possible: usually we allow hash changes in case the sources are effectively unchanged.

Comparing the two manually, indeed, there is no reason not to merge the update in this case

anmonteiro:
This is very unfortunate, and broke the hash in https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/ocaml-modules/duff/default.nix

mseri:
And this teaches us why we should avoid making these changes as much as possible. Thanks for the notification

dinosaure:
Yes, sorry for my mistake and the time spent by everyone on this task, I should look deeper next time when I see `ocaml-ci`. :+1:. Thanks for all of you.

gridbugs:
The hash change broke dune's monorepo benchmark as it uses an opam-monorepo lockfile containing the old hash. Can we have a policy in this repo of never updating hashes in-place? If we lose access to the original archive of a package then I think that should merit a new release instead, even if there is no code change.

gridbugs:
Also a trick I found for digging up old archives that aren't in the opam cache is to ask people to look in their ~/.opam/download-cache directory. I just checked on my machine and I have ` ~/.opam/download-cache/sha256/d1/d1eaa97cf58d38762391c8d74676544d4145d3f2fd13e4cda0ea8a0844cde612` which is a bz2 archive containing the original source for `duff-0.5`.

gridbugs:
To elaborate on why I feel that hashes should always be immutable, it's because users already need to account for the possibility that a package source archive will be unavailable or will have changed in place. Since package source archives are just urls and it's up to package authors to keep them available, it's always possible that the server hosting the file will be down or the package author will have accidentally or unwittingly changed a package archive in place causing its hash to change. I maintain the [dune monorepo benchmark](https://github.com/ocaml-dune/ocaml-monorepo-benchmark/tree/main/benchmark) which depends on > 1000 packages and I see this sort of thing all the time, especially since github occasionally re-archives packages sources (though this seems to happen much less frequently as of recently).

If users are allowed to assume that package hashes in the opam-repo never change then when a tool finds that the stored hash and computed hash differ, there is a single consistent narrative for what has happened - the package is unavailable or broken due to something on the package author's side - a possibility inherent in opam's design. Allowing the hash to update in opam-repo adds another dimension to the possibility space when debugging broken packages.

It also makes it harder to organize packages in a table keyed by hash (such as is done on opam.ocaml.org/cache) as now we have to think about the case when the hash changes which complicates the story.

hannesm:
I don't think this is the right issue to discuss such a policy (NB: I'm in favour of tarballs and hashes thereof to never change) -- but it seems the opam-repository maintenance team does not have a congruent policy (years ago I left that team exactly why people kept on changing hashes) -- in this issue I proposed to accept the change of the hash, and in retrospect this introduced quite some burden, and I'd advise against it in the future.

To me, it looks there is still some policy discussion needed for the opam-repository maintenance, plus enforcement via CI (or git hooks).

PR #24005, opened 2023-06-27
dune.3.8.0: fix duplicate conflicts field

avsm:
error introduced in  #23882


avsm:
fast merge to avoid hosing the CI

PR #24021, opened 2023-06-28
8 packages from ocaml/opam

rjbou:
This pull-request concerns:
- `opam-client.2.2.0~alpha`: Client library for opam 2.2
- `opam-core.2.2.0~alpha`: Core library for opam 2.2
- `opam-devel.2.2.0~alpha`: Bootstrapped development binary for opam 2.2
- `opam-format.2.2.0~alpha`: Format library for opam 2.2
- `opam-installer.2.2.0~alpha`: Installation of files to a prefix, following opam conventions
- `opam-repository.2.2.0~alpha`: Repository library for opam 2.2
- `opam-solver.2.2.0~alpha`: Solver library for opam 2.2
- `opam-state.2.2.0~alpha`: State library for opam 2.2

---
* Bug tracker: https://github.com/ocaml/opam/issues

---

### DIFF
--- a/packages/odep/odep.0.1.0/opam
+++ b/packages/odep/odep.0.1.0/opam
@@ -11,9 +11,9 @@ depends: [
   "sexplib"
   "ppx_sexp_conv" {>= "v0.13"}
   "parsexp"
-  "opam-core" {>= "2.1.0"}
-  "opam-state" {>= "2.1.0"}
-  "opam-format"
+  "opam-core" {>= "2.1.0" & < "2.2"}
+  "opam-state" {>= "2.1.0" & < "2.2"}
+  "opam-format" {< "2.2"}
   "ocamlfind" {>= "1.8.1"}
   "cmdliner" {>= "1.1.0"}
   "bos"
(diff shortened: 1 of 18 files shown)


avsm:
Excellent!  Would it be appropriate to add

```
flags: [ avoid-version ]
```

to each package, given it's an alpha release and not intended for general upgrades?

rjbou:
Agree, i'll add `avoid-version` flag.
The `available` proposal is for what purpose ?

rjbou:
Should I update PR with constraint on revdeps ? Or open a new one ?

rjbou:
Also, should we add the constraint only on failing packages, or on all package that have opam libs as dependency ?

avsm:
Updating the revdeps in this PR works; just the broken packages is fine.  The opam-version check is because the semantics of avoid-version were changed/fixed in 2.1, iirc.

raphael-proust:
Updating the revdeps (or other affected but not directly related) packages in the same PR works. Although my preference is to do it in a separate PR because it makes the CI (and diff but that's less important) more readable.

In this case, the main issue seem to be the `opam-build` and `opam-test`. And maybe the issues are related to the fact that the `opam-version` constraints are unsatisfiable so it just doesn't install the `opam-client` deps.

rjbou:
Opened another PR

avsm:
opam-test is also failing, but can also be fixed in a followup PR. Merging!

PR #24036, opened 2023-07-02
[new release] sel (0.3.0)

gares:
Simple Event Library

- Project page: <a href="https://github.com/gares/sel">https://github.com/gares/sel</a>

##### CHANGES:

- Major code cleanup and documentation
- Removal of recurrent events
- APIs `pop` and `wait` can now be freely mixed
- CI on Linux, Windows and MacOS

### DIFF
--- a/packages/sel/sel.0.3.0/opam
+++ b/packages/sel/sel.0.3.0/opam
@@ -1,8 +1,7 @@
 opam-version: "2.0"
 opam-version: "2.0"
 synopsis: "Simple Event Library"
-description:
-  "This library is the result of our experience in using threads and the Lwt async monad to tame the problem of writing a server which has to listen and react to multiple sources of events. The library itself is just sugar atop Unix.select. You can read more about the library on https://github.com/gares/sel"
+description: "This library is the result of our experience in using threads and the Lwt async monad to tame the problem of writing a server which has to listen and react to multiple sources of events. The library itself is just sugar atop Unix.select. You can read more about the library on https://github.com/gares/sel"
 maintainer: ["Enrico Tassi <enrico.tassi@inria.fr>"]
 authors: ["Enrico Tassi"]
 license: "MIT"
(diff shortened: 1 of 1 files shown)


packages/sel/sel.0.3.0/opam
```diff
@@ -0,0 +1,41 @@
+opam-version: "2.0"
+synopsis: "Simple Event Library"
+description: "Some sugar atop Unix.select"
```
mseri, line 3:
```suggestion
description: """This library is the result of our experience in using threads and the Lwt async monad to tame the problem of writing a server which has to listen and react to multiple sources of events. The library itself is sugar atop Unix.select. 
You can read more about the philosopy of the library on https://github.com/gares/sel
"""
This looks good to merge but I think the description could be a bit more informative
```

packages/sel/sel.0.3.0/opam
```diff
@@ -0,0 +1,41 @@
...
+maintainer: ["Enrico Tassi"]
```
mseri, line 4: Would it be possible to have an email with the maintainer?

gares:
done (also upstream)

mseri:
I think if you don't use a multiline string, it needs to be all in one line. You can check it with `opam lint $filename.opam`

gares:
Thanks for fixing, but it was dune generating that file.
Did I do anything wrong here: https://github.com/gares/sel/blob/main/dune-project ?

mseri:
Actually I think dune generated a fine opam file, I don't understand why our CI is crashing during the lint right now

packages/sel/sel.0.3.0/opam
```diff
@@ -0,0 +1,42 @@
+opam-version: "2.0"
```
mseri, line 1:
```suggestion
```

mseri:
Let's see if this works, there seems to be a repeated `opam-version` entry

mseri:
Ok, that was the issue. The opam file in your repository looks good, there is no need to make further changes there

PR #24048, opened 2023-07-04
bap-ghidra: accept failures in CI

emillon:
The package relies on <ghidra/loadimage.hh> without a corresponding
depext. So we allow it to fail in the common configuration.


mseri:
Thanks but I am going to close this since it fails to build on all distributions anyway

PR #24060, opened 2023-07-05
Package mccs.1.1+16

rjbou:
### `mccs.1.1+16`
MCCS (which stands for Multi Criteria CUDF Solver) is a CUDF problem solver
developed at UNS during the European MANCOOSI project

---
* Homepage: https://www.i3s.unice.fr/~cpjm/misc/
* Source repo: git+https://github.com/ocaml-opam/ocaml-mccs.git
* Bug tracker: https://github.com/ocaml-opam/ocaml-mccs/issues

---
:camel: Pull-request generated by opam-publish v2.2.0

PR #24026, opened 2023-06-29
Constraint revdeps to `opam-* < "2.2"`

rjbou:
Packages

* odep.0.1.0
* opam-build.0.1.0
* opam-custom-install.0.2
* opam-custom-install.0.3
* opam-publish.2.1.0
* opam-publish.2.2.0
* opam-test.0.1.0
* spectrum.0.4.0
* spectrum.0.5.0
* spectrum.0.6.0

PR #24061, opened 2023-07-05
Update opam-test dependencies opam-client < 2.2

rjbou:
It seems not compliant with opam 2.2
cf https://github.com/ocaml/opam-repository/pull/24060#issuecomment-1621658164
& #24026 #24021

### DIFF
--- a/packages/opam-test/opam-test.0.1.0/opam
+++ b/packages/opam-test/opam-test.0.1.0/opam
@@ -11,9 +11,7 @@ build: ["dune" "build" "-p" name "-j" jobs]
 depends: [
   "ocaml" {>= "4.02"}
   "dune" {>= "2.0"}
-  "opam-client"
-  {opam-version >= "2.1" & opam-version < "2.2" & >= "2.1.0" & < "2.2"}
-  "opam-client" {>= "2.2" & < "2.3" & opam-version >= "2.2" & opam-version < "2.3"}
+  "opam-client" {opam-version >= "2.1" & opam-version < "2.2" & >= "2.1.0" & < "2.2"}
   "cmdliner" {>= "1.0"}
 ]
 available: opam-version >= "2.1" & opam-version < "2.3"
--- a/packages/opam-test/opam-test.0.1.0/opam
+++ b/packages/opam-test/opam-test.0.1.0/opam
@@ -14,7 +14,7 @@ depends: [
   "opam-client" {opam-version >= "2.1" & opam-version < "2.2" & >= "2.1.0" & < "2.2"}
   "cmdliner" {>= "1.0"}
 ]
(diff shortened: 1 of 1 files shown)


packages/opam-test/opam-test.0.1.0/opam
```diff
@@ -11,9 +11,7 @@ build: ["dune" "build" "-p" name "-j" jobs]
 depends: [
   "ocaml" {>= "4.02"}
   "dune" {>= "2.0"}
-  "opam-client"
-  {opam-version >= "2.1" & opam-version < "2.2" & >= "2.1.0" & < "2.2"}
-  "opam-client" {>= "2.2" & < "2.3" & opam-version >= "2.2" & opam-version < "2.3"}
+  "opam-client" {opam-version >= "2.1" & opam-version < "2.2" & >= "2.1.0" & < "2.2"}
```
mseri, line 14:
```suggestion
  "opam-client" {opam-version >= "2.1" & opam-version < "2.3" & >= "2.1.0" & < "2.2"}
```

mseri: Right now it fails with opam-client not found, I think because with the opam-version bounds it is ignoring the constraint (the CI has opam 2.2). Should we try this alternative?

mseri: Should we make it conflict with opam-version outside the range otherwise?

rjbou: From what i understand, it needs to have the same opam-lib version than the used binary. And there is an api update, it is not compatible with 2.2.x. So maybe have avilability on opam-version=2.1? and kee lib opam-client >= "2.1.0" & < "2.2" ?

note: it was the same issue for opam-build in #24026, cf [failing ci](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/2526f77cadfdab5dfcb2aec30e276da538b17430)

packages/opam-test/opam-test.0.1.0/opam
```diff
@@ -11,9 +11,7 @@ build: ["dune" "build" "-p" name "-j" jobs]
...
   "cmdliner" {>= "1.0"}
 ]
 available: opam-version >= "2.1" & opam-version < "2.3"
```
rjbou, line 17:
```suggestion
available: opam-version >= "2.1" & opam-version < "2.2"
```
?

packages/opam-test/opam-test.0.1.0/opam
rjbou:
```suggestion
  "opam-client" {>= "2.1.0" & < "2.2"}
```

mseri:
Thanks!

PR #24070, opened 2023-07-06
why3: ensure configure works with Cygwin

maroneze:
Why3 is moving to Dune soon, but until then, its `./configure` script should work on Cygwin.

Currently, configure fails due to extraneous `\r` characters present at the end of some calls to `ocamlfind`. This commit adds a patch file with `| tr -d '\r'` to each of such calls, which fixes the issue.

I didn't report it directly upstream because (1) we need it fairly quickly, and why3 is not going to release anytime soon (especially not for such a "minor" issue that only affects Cygwin users), and (2) why3 is moving to Dune, to the `./configure` script will likely be replaced with `dune-configurator` anyway.

### DIFF
--- a/packages/why3/why3.1.6.0/files/cygwin.patch
+++ b/packages/why3/why3.1.6.0/files/cygwin.patch
@@ -1,0 +1,183 @@
+diff --git a/configure.in b/configure.in
+index 0002665c0..9cb7151eb 100644
+--- a/configure.in
++++ b/configure.in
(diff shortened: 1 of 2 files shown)


mseri:
This looks fine. I was wondering if we should move this to `1.6.0-1` but the build is broken on windows anyway and the patch does not affect other systems, so I think it is fair to do it here in this case

mseri:
Thanks

PR #24074, opened 2023-07-06
Package apron.0.9.14-beta1

antoinemine:
### `apron.0.9.14-beta1`
APRON numerical abstract domain library
Apron is a library to represent properties of numeric variables, such as variable bounds or linear relations between variables, and to manipulate these properties through semantic operations, such as variable assignments, tests, conjunctions, entailment. Apron is intended to be used in static program analyzers, in order to infer invariants of numeric variables, i.e., properties that hold for all executions of a program. It is based on the theory of Abstract Interpretation.

---
* Homepage: https://antoinemine.github.io/Apron/doc/
* Source repo: git+https://github.com/antoinemine/apron.git
* Bug tracker: https://github.com/antoinemine/apron/issues

---

### DIFF
--- a/packages/apron/apron.0.9.14-beta1/opam
+++ b/packages/apron/apron.0.9.14-beta1/opam
@@ -1,0 +1,53 @@
+opam-version: "2.0"
+synopsis: "APRON numerical abstract domain library"
+authors: [
+  "Bertrand Jeannet"
+  "Antoine Miné"
+  "https://github.com/antoinemine/apron/graphs/contributors"
+]
+license: "LGPL-2.1-only WITH OCaml-LGPL-linking-exception"
+homepage: "https://antoinemine.github.io/Apron/doc/"
+maintainer: "Nicolas Berthier <m@nberth.space>"
+dev-repo: "git+https://github.com/antoinemine/apron.git"
+bug-reports: "https://github.com/antoinemine/apron/issues"
+build: [
+  [
+    "sh"
+    "./configure"
+    "--prefix"
+    "%{share}%/apron"
+    "--no-ppl" {!conf-ppl:installed}
+    "--no-pplite" {!conf-pplite:installed}
+    "--no-ocaml-plugins" {os = "freebsd"}
+    "--absolute-dylibs" {os = "macos"}
+    "--ext-dll" {os = "win32"} "dll" {os = "win32"}
+    "--debug"
+    "--no-strip"
+  ]
+  [make "-j%{jobs}%"]
+]
+install: [
+  [make "install"]
+]
+depends: [
+  "ocaml"
+  "ocamlfind" {build}
+  "camlidl"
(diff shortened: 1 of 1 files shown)


antoinemine:
It should be OK to remove the references to `Stdlib`. so that the package compiles for OCaml < 4.07 (no need to add a dependency if it's not useful).

For elina 1.3, the checksum is wrong (or maybe the archive changed since the Opam package was made). I guess it should be fixed on the elina package?

For the macos-homebrew-ocaml-4.14-arm64, GMP is not found. No sure how to fix this.

For OCaml 5, it is supported by this version of Apron, but the mlgmpidl version with OCaml 5 support (on which Apron depends) does not seem to be in Opam yet. See https://github.com/nberth/mlgmpidl/issues/17

packages/apron/apron.0.9.14-beta1/opam
```diff
@@ -0,0 +1,53 @@
...
+    "--debug"
+    "--no-strip"
+  ]
+  [make "-j%{jobs}%"]
+]
+install: [
+  [make "install"]
+]
+depends: [
+  "ocaml"
```
mseri, line 33:
```suggestion
  "ocaml" {>= "4.07"}
```
Otherwise you could just add a lower bound here

mseri:
> For elina 1.3, the checksum is wrong (or maybe the archive changed since the Opam package was made). I guess it should be fixed on the elina package?

Yes, we should look for the old tarball on the cache or software heritage, upload it on opam-source-archives and link to that. This can be done separately and does not depend on this PR

mseri:
@antoinemine can you move the opam file from `packages/apron/apron.0.9.14-beta1/opam` to `packages/apron/apron.0.9.14~beta1/opam`? Otherwise the beta will have higher version than the final version once released

antoinemine:
> @antoinemine can you move the opam file from `packages/apron/apron.0.9.14-beta1/opam` to `packages/apron/apron.0.9.14~beta1/opam`? Otherwise the beta will have higher version than the final version once released

I'm following the instructions at https://opam.ocaml.org/doc/Packaging.html for packages developed on GitHub.
`~`  does not seem to be accepted as git tag: https://git-scm.com/docs/git-check-ref-format

mseri:
You can keep the tag with a dash, but use the tilde for the opam folder

mseri:
I think `opam publish --version=0.9.14~beta1 https://github.com/antoinemine/apron/archive/refs/tags/v0.9.14-beta1.tar.gz` should do the trick, or something like that (I am more familiar with dune release and don't have opam publish installed at the moment)

antoinemine:
OK, I will try a new `opam  publish -v ...`. And also fix the Stdlib issue. I guess this will open a new PR.

antoinemine:
#24090 should take care of this. Closing...

PR #24084, opened 2023-07-07
Mark utop.2.13.0 as not available on windows

emillon:
See ocaml-community/utop#447


emillon:
And #24085 publishes a fixed version

raphael-proust:
Thanks

PR #24090, opened 2023-07-10
Package apron.v0.9.14~beta.2

antoinemine:
### `apron.v0.9.14~beta.2`
APRON numerical abstract domain library
Apron is a library to represent properties of numeric variables, such as variable bounds or linear relations between variables, and to manipulate these properties through semantic operations, such as variable assignments, tests, conjunctions, entailment. Apron is intended to be used in static program analyzers, in order to infer invariants of numeric variables, i.e., properties that hold for all executions of a program. It is based on the theory of Abstract Interpretation.

---
* Homepage: https://antoinemine.github.io/Apron/doc/
* Source repo: git+https://github.com/antoinemine/apron.git
* Bug tracker: https://github.com/antoinemine/apron/issues

---

### DIFF
--- a/packages/apron/apron.v0.9.14~beta.2/opam
+++ b/packages/apron/apron.v0.9.14~beta.2/opam
@@ -1,0 +1,53 @@
+opam-version: "2.0"
+synopsis: "APRON numerical abstract domain library"
+authors: [
+  "Bertrand Jeannet"
+  "Antoine Miné"
+  "https://github.com/antoinemine/apron/graphs/contributors"
+]
+license: "LGPL-2.1-only WITH OCaml-LGPL-linking-exception"
+homepage: "https://antoinemine.github.io/Apron/doc/"
+maintainer: "Nicolas Berthier <m@nberth.space>"
+dev-repo: "git+https://github.com/antoinemine/apron.git"
+bug-reports: "https://github.com/antoinemine/apron/issues"
+build: [
+  [
+    "sh"
+    "./configure"
+    "--prefix"
+    "%{share}%/apron"
+    "--no-ppl" {!conf-ppl:installed}
+    "--no-pplite" {!conf-pplite:installed}
+    "--no-ocaml-plugins" {os = "freebsd"}
+    "--absolute-dylibs" {os = "macos"}
+    "--ext-dll" {os = "win32"} "dll" {os = "win32"}
+    "--debug"
+    "--no-strip"
+  ]
+  [make "-j%{jobs}%"]
+]
+install: [
+  [make "install"]
+]
+depends: [
+  "ocaml"
+  "ocamlfind" {build}
+  "camlidl"
+  "mlgmpidl"
+  "ocamlbuild" {build}
+  "conf-perl"
+]
+depopts: [
+  "conf-ppl"
+  "conf-pplite"
+]
+description:"""
+Apron is a library to represent properties of numeric variables, such as variable bounds or linear relations between variables, and to manipulate these properties through semantic operations, such as variable assignments, tests, conjunctions, entailment. Apron is intended to be used in static program analyzers, in order to infer invariants of numeric variables, i.e., properties that hold for all executions of a program. It is based on the theory of Abstract Interpretation."""
+url {
+  src:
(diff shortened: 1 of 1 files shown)


packages/apron/apron.v0.9.14~beta.2/opam
```diff
@@ -0,0 +1,53 @@
...
+  "ocaml"
+  "ocamlfind" {build}
+  "camlidl"
+  "mlgmpidl"
+  "ocamlbuild" {build}
+  "conf-perl"
+]
+depopts: [
+  "conf-ppl"
+  "conf-pplite"
```
sim642, line 42: There's no conf-pplite package in opam-repository though.

haochenx:
CI reports revdep elina.1.3 fails due to bad checksum, which indeed seems to be the case, but apparently unrelated to this PR.

```
[ERROR] Failed to get sources of elina.1.3: Bad checksum

#=== ERROR while fetching sources for elina.1.3 ===============================#
OpamSolution.Fetch_fail("https://github.com/eth-sri/ELINA/archive/1.3.tar.gz (Bad checksum, expected md5=e0c78d7b0a7afaee871a057b83d06a76)")


<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
| - fetch elina 1.3
+- 
- No changes have been performed
"/bin/bash" "-c" "opam reinstall elina.1.3;
        res=$?;
        test "$res" != 31 && exit "$res";
        export OPAMCLI=2.0;
        build_dir=$(opam var prefix)/.opam-switch/build;
        failed=$(ls "$build_dir");
        partial_fails="";
        for pkg in $failed; do
          if opam show -f x-ci-accept-failures: "$pkg" | grep -qF "\"debian-11\""; then
            echo "A package failed and has been disabled for CI using the 'x-ci-accept-failures' field.";
          fi;
          test "$pkg" != 'elina.1.3' && partial_fails="$partial_fails $pkg";
        done;
        test "${partial_fails}" != "" && echo "opam-repo-ci detected dependencies failing: ${partial_fails}";
        exit 1" failed with exit status 40
2023-07-11 07:36.37: Job failed: Failed: Build failed
2023-07-11 07:36.37: Log analysis:
2023-07-11 07:36.37: >>> 
[ERROR] Failed to get sources of elina.1.3: Bad checksum
 (score = 25)
2023-07-11 07:36.37: Failed to get sources of elina.1.3: Bad checksum
```

packages/apron/apron.v0.9.14~beta.2/opam
antoinemine: > There's no conf-pplite package in opam-repository though.

Yes. This was copied from ppl. Given that (unlike ppl) there's no pplite package for standard distributions (that I know of), adding a `conf-pplite` package may not be very useful. I suggest we remove this, and let the `./configure` script decide if pplite should be compiled in. What do you think?

sim642: A conf-pplite package might not be that unreasonable because the `./configure` check might be close to how packages like conf-ppl do the check as well. Not having system packages and depexts maybe isn't an issue for having a conf-* package.

opam-repository maintainers probably know the best how they'd like things. Seems like opam and the CI don't really care that a depopt doesn't exist at all, so maybe it's fine as-is. If a conf-pplite package is created, then it'll also configure this package.

mseri: I am fine with this, but I'd like to have the `conf-pplite` package as well, even if for the time being the list of depexts is empty. I think you can base it on https://github.com/ocaml/opam-repository/blob/master/packages/conf-ppl/conf-ppl.1/opam (or even simpler using conf-pkg-config if ppxlite supports pkg-config)

antoinemine: As far as I know, pplite has no package installer from standard distribution, so, it's quite different from ppl. I think it is better to remove the dependency on conf-pplite until there is some more reason to include such a package.

antoinemine:
Is there anything preventing the merge of this opam package?

mseri:
It looks fine, sorry for the long wait. I am just re-running the failing tests to make sure that I have fixed the necessary bounds

PR #24091, opened 2023-07-10
[new release] saturn_lockfree and saturn (0.4.0)

Sudha247:
Lock-free data structures for multicore OCaml

- Project page: <a href="https://github.com/ocaml-multicore/saturn">https://github.com/ocaml-multicore/saturn</a>
- Documentation: <a href="https://ocaml-multicore.github.io/saturn">https://ocaml-multicore.github.io/saturn</a>

##### CHANGES:

- Add docs and rename/refactor to add a lockfree package (@lyrm)
- CI clean up and set up Windows CI (@lyrm)
- Adopt OCaml Code of Conduct (@Sudha247)
- Mark alcotest as a test dependency (@Khady)
- Set QCHECK_MSG_INTERVAL to avoid clutter in CI logs (@jmid)
- Fix space leaks in MS Queue (@polytypic, @lyrm)
- Add STM tests for current data structures (@lyrm, @jmid)

### DIFF
--- a/packages/saturn/saturn.0.4.0/opam
+++ b/packages/saturn/saturn.0.4.0/opam
@@ -1,0 +1,31 @@
+opam-version: "2.0"
+maintainer:"KC Sivaramakrishnan <sk826@cl.cam.ac.uk>"
+authors: ["KC Sivaramakrishnan <sk826@cl.cam.ac.uk>"]
+homepage: "https://github.com/ocaml-multicore/saturn"
+doc: "https://ocaml-multicore.github.io/saturn"
+synopsis: "Parallelism-safe data structures for multicore OCaml"
+license: "ISC"
+dev-repo: "git+https://github.com/ocaml-multicore/saturn.git"
+bug-reports: "https://github.com/ocaml-multicore/saturn/issues"
+depends: [
+  "ocaml" {>= "4.12"}
+  "dune" {>= "3.0"}
+  "domain_shims" {>= "0.1.0"}
+  "qcheck" {with-test & >= "0.18.1"}
+  "qcheck-stm" {with-test & >= "0.2"}
+  "qcheck-alcotest" {with-test & >= "0.18.1"}
+  "alcotest" {with-test & >= "1.6.0"}
+  "yojson" {with-test &>= "2.0.2"}
+  "dscheck" {with-test & >= "0.1.0"}
+]
+available: arch != "x86_32" & arch != "arm32" & arch != "ppc64"
+build: ["dune" "build" "-p" name "-j" jobs]
+url {
+  src:
+    "https://github.com/ocaml-multicore/saturn/releases/download/0.4.0/saturn-0.4.0.tbz"
+  checksum: [
+    "sha256=7c7bec95a27055b41aa83540fcc1c6a87c9b7ad61bc511a532b8605ea33788fb"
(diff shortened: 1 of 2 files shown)


Sudha247:
Looks like there's a dependency issue on `saturn`. Will look into fixing it.
(I do wonder how the CI didn't catch it though).

packages/saturn/saturn.0.4.0/opam
```diff
@@ -0,0 +1,32 @@
...
+doc: "https://ocaml-multicore.github.io/saturn"
+synopsis: "Parallelism-safe data structures for multicore OCaml"
+license: "ISC"
+dev-repo: "git+https://github.com/ocaml-multicore/saturn.git"
+bug-reports: "https://github.com/ocaml-multicore/saturn/issues"
+depends: [
+  "ocaml" {>= "4.12"}
+  "dune" {>= "3.0"}
+  "domain_shims" {>= "0.1.0"}
+  "saturn_lockfree" {>= "0.4.0"}
```
mseri, line 14:
```suggestion
  "saturn_lockfree" {= version}
```
are you sure you don't prefer to lock the versions?

Sudha247: Makes sense, thanks!

packages/saturn_lockfree/saturn_lockfree.0.4.0/opam
```diff
@@ -0,0 +1,31 @@
...
+  "dune" {>= "3.0"}
+  "domain_shims" {>= "0.1.0"}
+  "qcheck" {with-test & >= "0.18.1"}
+  "qcheck-stm" {with-test & >= "0.2"}
+  "qcheck-alcotest" {with-test & >= "0.18.1"}
+  "alcotest" {with-test & >= "1.6.0"}
+  "yojson" {with-test &>= "2.0.2"}
+  "dscheck" {with-test & >= "0.1.0"}
+]
+available: arch != "x86_32" & arch != "arm32" & arch != "ppc64"
```
avsm, line 21: What's the intention behind this restriction -- is it that they don't work on those architectures, or that you need native code support? Given ppc64 will be back in 5.2, we should use a different constraint if the latter.

Sudha247: To be fair, that was an unfortunate consequence of getting the CI to work reliably. The library is terribly slow on bytecode-only platforms to a point I doubt they're even useful on those. For context, the CI was routinely timing out on bytecode-only platforms without any real failures.

However, on second thoughts, we should perhaps not impose this restriction and let users decide how they want to use them. A reasonable compromise should be to disable tests on those. Good point about ppc64, I'll enable it on this one. I think we can decide on 32-bit systems later.

mseri:
Thanks!

packages/saturn_lockfree/saturn_lockfree.0.4.0/opam
polytypic: TIL that, on 32-bit native OCaml 4, loops that don't allocate may not have safepoints, which means that systhread switches may not happen when running such loops and that can cause code to never terminate.  I had tests in Kcas that occasionally failed to terminate on 32-bit OCaml 4 backends due that.

Is it known that bytecode is causing the library to just run more slowly or could it be that it is actually some other difference that causes either code in the Saturn library or in the tests to not terminate?

Basically, to be safe to use on 32-bit OCaml 4, it is necessary to make sure that retry loops that depend on another systhread to make progress (e.g. a barrier or lock implementation), must include a safe point such as an explicit call to `Thread.yield ()`.

PR #24094, opened 2023-07-11
Package cryptoverif.2.07

bruno-blanchet:
### `cryptoverif.2.07`
CryptoVerif: Cryptographic protocol verifier in the computational model
CryptoVerif is an automatic protocol prover sound in the computational model. It can prove

  - secrecy;
  - correspondences, which include in particular authentication;
  - indistinguishability between two games.

It provides a generic mechanism for specifying the security assumptions on cryptographic primitives, which can handle in particular symmetric encryption, message authentication codes, public-key encryption, signatures, hash functions.

The generated proofs are proofs by sequences of games, as used by cryptographers. These proofs are valid for a number of sessions polynomial in the security parameter, in the presence of an active adversary. CryptoVerif can also evaluate the probability of success of an attack against the protocol as a function of the probability of breaking each cryptographic primitive and of the number of sessions (exact security).

This software is under development; please use it at your own risk. Comments and bug reports welcome.

---
* Homepage: https://cryptoverif.inria.fr
* Source repo: git+https://gitlab.inria.fr/bblanche/CryptoVerif.git
* Bug tracker: Bruno.Blanchet@inria.fr

---

### DIFF
--- a/packages/cryptoverif/cryptoverif.2.07/opam
+++ b/packages/cryptoverif/cryptoverif.2.07/opam
@@ -1,0 +1,33 @@
+opam-version: "2.0"
+synopsis: "CryptoVerif: Cryptographic protocol verifier in the computational model"
+description: """
+CryptoVerif is an automatic protocol prover sound in the computational model. It can prove
+
+  - secrecy;
+  - correspondences, which include in particular authentication;
+  - indistinguishability between two games.
+
+It provides a generic mechanism for specifying the security assumptions on cryptographic primitives, which can handle in particular symmetric encryption, message authentication codes, public-key encryption, signatures, hash functions.
+
+The generated proofs are proofs by sequences of games, as used by cryptographers. These proofs are valid for a number of sessions polynomial in the security parameter, in the presence of an active adversary. CryptoVerif can also evaluate the probability of success of an attack against the protocol as a function of the probability of breaking each cryptographic primitive and of the number of sessions (exact security).
+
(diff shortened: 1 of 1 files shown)


packages/cryptoverif/cryptoverif.2.07/opam
```diff
@@ -0,0 +1,33 @@
...
+
+This software is under development; please use it at your own risk. Comments and bug reports welcome. 
+"""
+maintainer: "Bruno Blanchet <Bruno.Blanchet@inria.fr>"
+authors: "Bruno Blanchet <Bruno.Blanchet@inria.fr>, David Cadé <David.Cade@normalesup.org>, and Benjamin Lipp <Benjamin.Lipp@mpi-sp.org>"
+license: "CECILL-B"
+homepage: "https://cryptoverif.inria.fr"
+dev-repo: "git+https://gitlab.inria.fr/bblanche/CryptoVerif.git"
+bug-reports: "Bruno.Blanchet@inria.fr"
+depends: [ "ocaml" { >= "4.04" } "ocamlfind" { post } "cryptokit" { post } "conf-m4" { post } ]
```
avsm, line 22: Why are these dependencies marked `{post}` -- to break a dependency cycle somewhere?

avsm:
Sounds good, thanks! We don't have a notion of runtime dependencies in opam, so this is the closest alternative.  

You may want to announce this on <https://discuss.ocaml.org>, where we have a Community category and an `announce` tag for this purpose.

PR #24101, opened 2023-07-12
Add ctypes 0.21.1, ctypes-foreign 0.21.1

yallop:
This release includes a switch to `dune`, which is a fairly major change (https://github.com/yallop/ocaml-ctypes/pull/588), and may need some iteration before it's ready to merge.


emillon:
I just rebased and included pointers to all the upstream fixes in the commit messages. I don't expect important failures in that run so let's see how it goes.

avsm:
Would it be possible to include the patches in this PR in a release, and merge that instead? Where possible, we prefer to not carry upstream patches directly in the opam repository.

yallop:
I've merged the patches into ctypes 0.21.1, and updated the PR to point to that release.

avsm:
Thanks @yallop and @emillon for getting this over the line to be released!

PR #24104, opened 2023-07-13
Package mutaml.0.1.0

jmid:
### `mutaml.0.1.0`
A mutation tester for OCaml
Mutaml is a mutation testing tool for OCaml.
It uses a ppxlib-based preprocessor to make a series of small
breaking changes to a program's source code and then runs
the program's testsuite for each of them to catch uncaught
misbehaviour.

---
* Homepage: https://github.com/jmid/mutaml
* Source repo: git+https://github.com/jmid/mutaml.git
* Bug tracker: https://github.com/jmid/mutaml/issues

---

### DIFF
--- a/packages/mutaml/mutaml.0.1.0/opam
+++ b/packages/mutaml/mutaml.0.1.0/opam
@@ -1,0 +1,48 @@
+opam-version: "2.0"
+
+synopsis:    "A mutation tester for OCaml"
+license:     "BSD-2-Clause"
+homepage:    "https://github.com/jmid/mutaml"
+doc:         "https://github.com/jmid/mutaml"
+bug-reports: "https://github.com/jmid/mutaml/issues"
+
+dev-repo:    "git+https://github.com/jmid/mutaml.git"
+authors:     "Jan Midtgaard <mail@janmidtgaard.dk>"
+maintainer:  "Jan Midtgaard <mail@janmidtgaard.dk>"
+tags: [
+  "test"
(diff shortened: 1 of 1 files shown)


jmid:
Added a `conf-which` dependency (to please archlinux) and made it unavailable on alpine (missing `diff --color`)

jmid:
Added a `conf-diffutils` dependency and fixed `dune subst` to please the opam linter.

jmid:
Overall, `mutaml` contains a collection of cram tests which are challenging to ensure cross-platform.

For the remaining opam-ci red lights:
- `x86_32-ocaml-4.14` is failing cram test `diff`s due to linker warnings being emitted https://github.com/ocaml/ocaml/issues/12166
- `ppc64-ocaml-4.14` is failing because a mutated test fails with a timeout instead of a failure (as the only platform)
- `macos-homebrew-*` is failing due to a slightly different `diff` being output. 

All of these are harmless. 
For the latter, I've just checked locally on a amd64-based MacBook with homebrew:
After a `brew install diffutils` (which installed 3.10) the test suite completes successfully. 
I therefore suspect it is caused by a too old `diffutil` version on the macOS CI runners (my Linux box has 3.7 which is also sufficient).

mseri:
Thanks. I think we can merge as is. I would not disable the tests on the opam-repo for the moment, it can be useful to go and check the failures in case some breaking changes in the dependencies are suspected

PR #24114, opened 2023-07-17
[new release] moonpool (0.3)

c-cube:

##### CHANGES:

- add `Fork_join` for parallelizing computations. This is only
    available on OCaml 5.x because it relies on effects.
- add `Fork_join.{for_,map_array,map_list}`
- add `Fork_join.all_{list,init}`
- add `Pool.with_`
- add a channel module
- add `Runner`, change `Pool` to produce a `Runner.t`
- add a `Lock` module
- add support for domain-local-await when installed
- add `Fut.await` for OCaml >= 5.0

- fix: Fork_join.both_ignore now has a more general type

- expose `Suspend_` and its internal effect with an unstability alert.
    This is intended for implementors of `Runner` only.
- port `cpp.ml` from containers, replace previous codegen with it.
    This will provide better flexibility for supporting multiple versions
    of OCaml in the future.
- add `Pool.run_wait_block`; rename `Pool.run` into `Pool.run_async`
- fix: in blocking queue, `pop` works on a non empty closed queue

### DIFF
--- a/packages/moonpool/moonpool.0.3/opam
+++ b/packages/moonpool/moonpool.0.3/opam
@@ -1,0 +1,44 @@
+opam-version: "2.0"
+synopsis: "Pools of threads supported by a pool of domains"
+maintainer: ["Simon Cruanes"]
+authors: ["Simon Cruanes"]
+license: "MIT"
+tags: ["thread" "pool" "domain"]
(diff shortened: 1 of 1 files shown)


mseri:
Thanks

PR #24118, opened 2023-07-18
DkML 2.0.0

jonahbeckford:
DkML installer and tools for version 2.0.0

### DIFF
--- a/packages/conf-dkml-sys-opam/conf-dkml-sys-opam.1/opam
+++ b/packages/conf-dkml-sys-opam/conf-dkml-sys-opam.1/opam
@@ -1,0 +1,69 @@
+opam-version: "2.0"
+maintainer: ["opensource+diskuv-ocaml@support.diskuv.com"]
+homepage: "https://diskuv.gitlab.io/diskuv-ocaml"
+bug-reports: "https://github.com/diskuv/dkml-runtime-apps/issues"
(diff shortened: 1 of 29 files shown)


jonahbeckford:
I won't be submitting this PR for recreating the DkML distribution. But opam lint is saying an error when it is not an error ... and I'd like to get that fixed or skipped for future versions:

```
2023-07-18 05:45.13: Error in dkml-component-staging-desktop-full.2.0.0: Your dune-project file indicates that this package requires at least dune 3.6 but your opam file only requires dune >= 1.0. Please check which requirement is the right one, and fix the other.
2023-07-18 05:45.13: Error in dkml-component-staging-desktop-ci.2.0.0: Your dune-project file indicates that this package requires at least dune 3.6 but your opam file only requires dune >= 1.0. Please check which requirement is the right one, and fix the other.
```

(Yes, it should make sense to select a specific Dune version because that is exactly the version of Dune that DkML 2.0.0 is installing).

The `dune-project` has:

```scheme
(lang dune 3.6)

(package
 ; ...
 (depends
  ; ...
  (dune (or (= 3.8.3) (= 3.8.3+shim)))
```

and that generates in `.opam`:

```perl
depends: [
  # ...
  "dune" {>= "3.6" & = "3.8.3" | = "3.8.3+shim"}
```

Source code: https://gitlab.com/diskuv-ocaml/components/dkml-component-desktop/-/tree/99ca77ddba31b4521a6a6d002ff3d736085f9cbd

PR #24050, opened 2023-07-04
OCaml 5.1.0, first beta release

Octachron:
This PR adds the two packages for the first beta release of OCaml 5.1.0 .

Compared to the last alpha, we have two subtle and hopefully invisible runtime fixes, and one windows fix. Overall, the beta looks in a very good shape for a first beta.

----------------------------------------

### Runtime

- 12037: Fix some data races by using volatile when necessary
  (Fabrice Buoro and Olivier Nicole, review by Guillaume Munch-Maccagnoni,
  Gabriel Scherer and Luc Maranget)

- 12253, 12342: Fix infinite loop in signal handling.
  (Guillaume Munch-Maccagnoni, report by Thomas Leonard, review by
   KC Sivaramakrishnan and Sadiq Jaffer)

### Windows

- 12184, 12320: Sys.rename Windows fixes on directory corner cases.
  (Jan Midtgaard, review by Anil Madhavapeddy)

PR #24130, opened 2023-07-19
Correct 32-bit architectures for 5.1 and 5.2

dra27:
Alas, in #24075 I correctly fixed the bytecode-only 64-bit architectures in 5.1 and 5.2 and incorrectly broke the 32-bit architectures.

32-bit architectures can be written as `arch = "x86_32" | arch = "arm32" | arch = "ppc32"`, but I think it's better to put what @Octachron had originally put in the 5.1.0~alpha2 package and have a conjunction of the _supported_ native architectures which is what I've done here (cf. also https://github.com/ocaml/opam/issues/4633).

cc @tmcgilchrist

### DIFF
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.5.1.0~alpha1/opam
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.5.1.0~alpha1/opam
@@ -14,7 +14,7 @@ depends: [
   "base-domains" {post}
   "base-nnp" {post}
   "ocaml-options-vanilla" {post}
-  "ocaml-option-bytecode-only" {arch = "ppc64" | arch = "s390x"}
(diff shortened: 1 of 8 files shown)


tmcgilchrist (approved):
~~Looks like the right set of options to me.~~

EDIT The reported `arch` needs to change after checking on a riscv machine.

dra27:
Just double-checking the `arch` value for riscv…

dra27:
Oops - I took the `"riscv"` value from #24050, but that's OCaml's architecture value (OCaml has `Config.arch = "riscv"` and `Config.model = "riscv64"`, differing from `uname -m`'s value)

avsm:
Thanks! We do need to document the 'useful' `arch` values somewhere; it's a perpetual source of ambiguity.

PR #24133, opened 2023-07-19
ZArith version 1.13

xavierleroy:
~~Now that ZArith's configure script uses pkg_config, I removed the macos special cases in the opam file, let's see if this works.~~  It didn't work, so I'm using the same opam file as for 1.12.

### DIFF
--- a/packages/zarith/zarith.1.13/opam
+++ b/packages/zarith/zarith.1.13/opam
@@ -1,0 +1,59 @@
+opam-version: "2.0"
+name: "zarith"
+maintainer: "Xavier Leroy <xavier.leroy@inria.fr>"
+authors: [
+  "Antoine Miné"
+  "Xavier Leroy"
+  "Pascal Cuoq"
+]
+homepage: "https://github.com/ocaml/Zarith"
+bug-reports: "https://github.com/ocaml/Zarith/issues"
+dev-repo: "git+https://github.com/ocaml/Zarith.git"
+license: "LGPL-2.0-only WITH OCaml-LGPL-linking-exception"
+build: [
+  ["./configure"] {os != "openbsd" & os != "freebsd" & os != "macos"}
+  [
+    "sh"
+    "-exc"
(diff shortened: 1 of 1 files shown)


packages/zarith/zarith.1.13/opam
```diff
@@ -0,0 +1,59 @@
...
+  ] {os = "macos" & os-distribution = "homebrew" & arch = "arm64" }
+  [make]
+]
+install: [
+  [make "install"]
+]
+depends: [
+  "ocaml"     {>= "4.04.0"}
+  "ocamlfind"
+  "conf-gmp"
```
avsm, line 43:
```suggestion
  "conf-gmp"
  "conf-pkg-config"
```

If you want to use pkg-config, you can add this and remove the macOS specific bits and see if that works.  Homebrew's gmp package does install a valid-looking gmp.pc file:

```
% /opt/homebrew/lib/pkgconfig/gmp.pc     
prefix=/opt/homebrew/Cellar/gmp/6.2.1_1
exec_prefix=${prefix}
includedir=${prefix}/include
libdir=/opt/homebrew/Cellar/gmp/6.2.1_1/lib

Name: GNU MP
Description: GNU Multiple Precision Arithmetic Library
URL: https://gmplib.org
Version: 6.2.1
Cflags: -I${includedir}
Libs: -L${libdir} -lgmp
```

xavierleroy: That's another option.  Right now it looks like something is wrong on macos ARM, probably due to a recent change in ZArith itself, so I need to debug this before going further with this PR.

xavierleroy: Found the problem in ZArith: LDFLAGS was not fully honored owing to an oddity / a possible bug in ocamlmklib.  Made a new release.  Updated this PR.  Now waiting for the macOS CI runners to kick in and confirm that the macOS ARM issue is gone.

avsm:
ARM builds all passed, merging!

PR #24147, opened 2023-07-22
Release camlp5 8.02.00 with ocaml 5.1.0 compatibility

chetmurthy:
Also release bugfix for camlp5-buildscripts


chetmurthy:
OK, I have no idea why these jobs are failing: they're all revdeps, but I don't seee how they depend on Camlp5.  And there are a TON of such revdeps, too.  It's a little perplexing, and I wonder if the revdeps calculation is broken?  Maybe it's finding too many revdeps ?

kit-ty-kate:
Sadly it is not a bug. The number comes from the fact that:
* `lablgl.1.07` depends on camlp5
* `lablgl` is an optional dependency of `lablgtk`
* `lablgtk` is an optional dependency of `lwt < 3.0.0`

so anything that uses `lwt` unconstrained can technically be a reverse dependency of `camlp5` through this one version of `lablgl`.

Having something that's more precise and actually check that dependency path is possible but fairly hard and costly to do right. If you want you can open a ticket to https://github.com/ocurrent/opam-repo-ci/issues

chetmurthy:
OK, so I take your point about the "too many revdeps".  But I looked at a few, and they're all failing for different reasons, and (it seems) all unrelated to camlp5/camlp5-buildscripts.  It looks like the test-running environment is corrupted, in truth.

chetmurthy:
OK, I've investigated two of the test-failures: "uring.0.1" and "http-cookie.4.1.0".  In both cases, I reran the dockerfile locally, got the failure, then *removed* the package-under-test (camlp5 or camlp5-buildscripts), and reran the dockerfile, getting the same (or a similar) error.

Has something changed in the test-running apparatus of opam?  How can I proceed to demonstrate this problem?  I could give you a Dockerfile you could run ?

kit-ty-kate:
Looks good, thanks. Sorry for the wait and the lack of feedbacks (i'll try to explain on https://github.com/ocaml/opam-repository/issues/24221)

PR #24154, opened 2023-07-24
Add `avoid-version` flag to latest Merlin preview

voodoos:
As discussed with @kit-ty-kate, the latest Merlin preview for OCaml 5.1 has inconsistent versioning.

@kit-ty-kate  I guess I don't need to propagate these changes to the upstream repository ?


avsm:
Thanks!

Issue #24151, opened 2023-07-24
Consider marking base.v0.16.2 as unavailable with MSVC


jonahbeckford:
Not picking on this one package, but it is pretty essential. base.v0.16.1 worked fine with MSVC (and is included in DkML), so this is a regression.

See https://github.com/janestreet/base/issues/157 for the MSVC failure.

avsm:
A PR to constrain this in the opam-repo is welcome... we don't have any automated testing for Windows yet, unfortunately.

jonahbeckford:
What is the preferred filter to use in this scenario?

kit-ty-kate:
there is no MSVC specific filter/packages currently in opam-repository so I would say:
```
available: os != "win32"
```
which makes the package unavailable on Windows as a whole.

PR #24155, opened 2023-07-24
Mark base.v0.16.2 as unavailable on Win32

jonahbeckford:
Root cause https://github.com/janestreet/base/issues/157
Fixes https://github.com/ocaml/opam-repository/issues/24151


kit-ty-kate:
Thanks

PR #24156, opened 2023-07-24
Base v0.16.2 does not build on Windows

dkalinichenko-js:
Add an OS restriction to Base v0.16.2.


kit-ty-kate:
Already done in #24155

PR #24191, opened 2023-08-01
Fix checksums for timmy.1.0.0.

mefyl:
For a reason that escapes my comprehension, the checksums for Timmy 1.0.0 seem to be wrong.

Fixes https://github.com/mefyl/timmy/issues/1.


mseri:
We need to check the content against the previous versions, and if possible instead of changing the checksums we should upload the old tarballs on github.com/ocaml/opam-source-archives and link to those

kit-ty-kate:
I can't find the archive anywhere, even https://archive.softwareheritage.org/browse/search/?q=opam.ocaml.org%2Fpackages%2Ftimmy&with_visit=true&with_content=true&visit_type=opam hasn't archived it, so let's merge this. Sorry for the wait.

PR #24192, opened 2023-08-01
Packages for OCaml 5.1.0 first release candidate

Octachron:
The usual two packages for hopefully the last version published of OCaml 5.1.0 before the last release.
Compared to the beta, we have only a safe bugfix on IO that solves a GC concurrency issue.

There are also two configuration tweaks, for openBSD and ocamlmktop that should be safe.

Overall, I expect to publish the full release next week or the week after.

-----------------------

Changes since the beta
-------

### Bug fix

- 12445: missing GC root registrations in runtime/io.c
  (Gabriel Scherer, review by Xavier Leroy and Jeremy Yallop)

### Configuration fixes

- 12372: Pass option -no-execute-only to the linker for OpenBSD >= 7.3
  so that code sections remain readable, as needed for closure marshaling.
  (Xavier Leroy and Anil Madhavapeddy, review by Anil Madhavapeddy and
  Sébastien Hinderer)

* 11745, +12358: Debugger and toplevels: embed printer types rather than
  reading their representations from topdirs.cmi at runtime.
  This change also removes the ocamlmktop initialization module introduced
  in 11382 which was no longer useful.
  This change breaks toplevel scripts relying on the visibility of `Topdirs`
  in the initial toplevel environment without loading `topfind`.
  Since the opam default `.ocamlinit` file loads `topfind`, it is expected
  that only scripts run with `ocaml -noinit` are affected.
  For those scripts, accessing `Topdirs` now requires the `compiler-libs`
  directory to be added to the toplevel search path with
  ```
    #directory "+compiler-libs";;
  ````
 as was already the case for the other modules in the toplevel interface
 library.
 (Sébastien Hinderer, review by Florian Angeletti, Nicolás Ojeda Bär and
 Gabriel Scherer)

### Documentation changes

- 12201: in the tutorial on modules, replace priority queue example by
  a simpler example based on FIFO queues.
  (Xavier Leroy, review by Anil Madhavapeddy and Nicolás Ojeda Bär).

- 12352: Fix a typo in the documentation of Arg.write_arg
  (Christophe Raffalli, review by Florian Angeletti)

### DIFF
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.5.1.0~rc1/files/ocaml-base-compiler.install
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.5.1.0~rc1/files/ocaml-base-compiler.install
@@ -1,0 +1,1 @@
+share_root: ["config.cache" {"ocaml/config.cache"}]
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.5.1.0~rc1/opam
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.5.1.0~rc1/opam
@@ -1,0 +1,48 @@
+opam-version: "2.0"
+synopsis: "First release candidate of OCaml 5.1.0"
+maintainer: "platform@lists.ocaml.org"
+license: "LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception"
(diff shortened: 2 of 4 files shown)


avsm:
Thanks!

PR #24193, opened 2023-08-01
[new release] cairn (1.0)

VincentPenelle:
A derivation explorer and logger for menhir parser

- Project page: <a href="https://github.com/VincentPenelle/cairn">https://github.com/VincentPenelle/cairn</a>

##### CHANGES:

- Initial public release


packages/cairn/cairn.1.0/opam
```diff
@@ -0,0 +1,45 @@
...
+authors: ["Vincent Penelle"]
+license: "GPL-2.0-or-later"
+homepage: "https://github.com/VincentPenelle/cairn"
+bug-reports: "https://github.com/VincentPenelle/cairn/issues"
+depends: [
+  "menhirSdk"
+  "menhirLib"
+  "ocaml"
+  "dune" {>= "2.9"}
+  "menhir"
```
tmcgilchrist, line 15: This needs a suitable lower bounds on both menhir and the OCaml version.
Eg 
```suggestion
 "ocaml" {>= "4.08"}
  "menhir" {>= "20230608"}
```

4.08 is implied by the use of `String.split_on_char` (since 4.04) and Int module since 4.08. Not sure about the `Zed_string` warning?

VincentPenelle: Indeed, many thanks @tmcgilchrist (and sorry for not having though about that myself, it is my first package contribution, I didn't think to check the version of the functions I use).
For the Zed_string warning, I've searched quickly on zed homepage, I do not see versions for the functions in the API. Maybe let's see if the warning persist after your changes (I accepted to quickly and don't find anymore the log). At worst, would it be a good idea to put the version I used as a lower bound?

VincentPenelle:
Hum, indeed I had not thought about the zed dependency. I've added it. Hope it solves the lower bounds problems I had.

avsm:

PR #24206, opened 2023-08-03
[new release] multicore-magic (2.0.0)

polytypic:
Low-level multicore utilities for OCaml

- Project page: <a href="https://github.com/ocaml-multicore/multicore-magic">https://github.com/ocaml-multicore/multicore-magic</a>

## 2.0.0

- Changed the semantics of `copy_as_padded` to not always copy and to not
  guarantee that `length_of_padded_array*` works with it. These semantic changes
  allow better use of the OCaml allocator to guarantee cache friendly alignment.
  (@polytypic)


avsm:

PR #24209, opened 2023-08-03
Package foxes.1.0.0

AMTitan:
### `foxes.1.0.0`
An API for all your fox image needs
An API for all your fox image needs

---
* Homepage: https://github.com/Foxes-cool/Foxes-Ocaml
* Source repo: git+https://github.com/Foxes-cool/Foxes-Ocaml.git
* Bug tracker: https://github.com/Foxes-cool/Foxes-Ocaml/issues

---

### DIFF
--- a/packages/foxes/foxes.1.0.0/opam
+++ b/packages/foxes/foxes.1.0.0/opam
@@ -1,0 +1,28 @@
+opam-version: "2.0"
+maintainer: "Arthur Melton <me@arthurmelton.com>"
+authors: [
+  "Arthur Melton <me@arthurmelton.com>"
+]
+homepage: "https://github.com/Foxes-cool/Foxes-Ocaml"
+bug-reports: "https://github.com/Foxes-cool/Foxes-Ocaml/issues"
+license: "MIT"
+dev-repo: "git+https://github.com/Foxes-cool/Foxes-Ocaml.git"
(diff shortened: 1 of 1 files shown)


yawaramin:
Sorry to be a downer, but do we really want a library in opam to get URLs of fox images from a specific website? Seems like this could be a GitHub Gist...

AMTitan:
I mean, technically, no, but the main advantage of Opam is the version control. It would also be nice for users to be able to know if the version of the installed API is out of date.

raphael-proust:
An alternative is to host your own opam repository. Essentially a repository like this one which has a `packages/` directory with the structure that opam expect from it.

Then the package becomes available to users after they add your own repository as sources that opam can pick from:
```
opam repository set-url <name> <git-url>
```

It's not as good for discoverability (without the repository-adding command, `opam search` won't show the result), but it's nice for the other aspects of package management (e.g., for users to be able to ensure that a specific version is installed).

AMTitan:
@kit-ty-kate When I do the `opam repository add ...` and `opam install foxes`, is there anything else I have to do? Opam says the library is installed but dune does not believe so.

kit-ty-kate:
Your package does not install anything: https://github.com/Foxes-cool/Foxes-Ocaml/blob/main/lib/dune

You need to use `(public_name foxes)` instead of `(name foxes)` if you want it to be available outside your project directory/installed

PR #24213, opened 2023-08-04
Package lwt.5.7.0

raphael-proust:
### `lwt.5.7.0`
Promises and event-driven I/O
A promise is a value that may become determined in the future.

Lwt provides typed, composable promises. Promises that are resolved by I/O are
resolved by Lwt in parallel.

Meanwhile, OCaml code, including code creating and waiting on promises, runs in
a single thread by default. This reduces the need for locks or other
synchronization primitives. Code can be run in parallel on an opt-in basis.

---
* Homepage: https://github.com/ocsigen/lwt
* Source repo: git+https://github.com/ocsigen/lwt.git
* Bug tracker: https://github.com/ocsigen/lwt/issues

---

### DIFF
--- a/packages/lwt/lwt.5.7.0/opam
+++ b/packages/lwt/lwt.5.7.0/opam
@@ -1,0 +1,57 @@
+opam-version: "2.0"
+
+synopsis: "Promises and event-driven I/O"
+license: "MIT"
+homepage: "https://github.com/ocsigen/lwt"
+doc: "https://ocsigen.org/lwt"
+bug-reports: "https://github.com/ocsigen/lwt/issues"
+
+authors: [
+  "Jérôme Vouillon"
+  "Jérémie Dimino"
+]
+maintainer: [
+  "Raphaël Proust <code@bnwr.net>"
+  "Anton Bachin <antonbachin@yahoo.com>"
+]
+dev-repo: "git+https://github.com/ocsigen/lwt.git"
+
+depends: [
+  "cppo" {build & >= "1.1.0"}
+  "dune" {>= "1.8.0"}
+  "dune-configurator"
+  "ocaml" {>= "4.08"}
+  "ocplib-endian"
+
+  # Until https://github.com/aantron/bisect_ppx/pull/327.
+  # "bisect_ppx" {dev & >= "2.0.0"}
+  "ocamlfind" {dev & >= "1.7.3-1"}
+]
+
+depopts: [
+  "base-threads"
+  "base-unix"
+  "conf-libev"
+]
+
+build: [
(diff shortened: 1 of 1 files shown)


raphael-proust:
There are a few deprecated functions that are removed in this release so possibly a few packages that need additional constraints. I'll monitor the revdeps.

raphael-proust:
Notes for myself to prepare another PR adding constraints:

- `dune-rpc-lwt` uses `Lwt_result.catch`
- `crunch.2.2.0` uses `Result.result`
- `aws-s3-lwt` uses `Lwt_result.catch`
- `ocsigen-start` uses `Lwt_stream.Error`

Otherwise likely unrelated to this PR: `kinetic-client` seem to have some left-over bytes-vs-string build issues?

I think that's it. I'll open a PR for fixing the revdeps constraints.

yawaramin:
Out of curiosity, doesn't Lwt use Semantic Versioning?

raphael-proust:
It doesn't follow semantic versioning strictly. We do use the last digit for bug-fixes only but we're not too strict on the top two numbers.

sim642:
On a somewhat related note, I'm wondering whether previous lwt versions should get an `"ocaml" {< "5.1"}` constraint due to https://github.com/ocsigen/lwt/pull/995.
Since [there are very few users of `Lwt_io.read_value`](https://sherlocode.com/?q=Lwt_io.read_value), it probably doesn't affect most Lwt users and could be overly restrictive to require Lwt >= 5.7 on OCaml 5.1. However, then downstream packages need to know about the incompatibility and add an opam conflict, in order to avoid crashes from marshaling.

raphael-proust:
> could be overly restrictive

We could add the restriction, and also ask potential users if they need a bugfix 5.6.2 release I guess.

I'm reluctant to do such a release if it's not used just because of time/resources, so I'll only do it if there is demand for it. Currently 5.1 is cutting edge so I don't know that it's unreasonable to expect users to be ok with a recent Lwt too.

raphael-proust:
I made https://github.com/ocaml/opam-repository/pull/24243 to mark incompatibilities. Note that pre-5.6 we already have the hard constraint for ocaml<5.0 so the bug affects only 5.6 so we might not need to do a bugfix release.

avsm:
Let's get this merged ahead of the 5.1 release. Is it just the upper bounds on use of Lwt_result.catch remaining, @raphael-proust ?

raphael-proust:
I think it's ready to be merged. The PR https://github.com/ocaml/opam-repository/pull/24228 addressed all the issues with `Lwt_result.catch` that I could find in the CI.

avsm:
In it goes!

PR #24217, opened 2023-08-05
[new release] backoff (0.1.0)

polytypic:
Exponential backoff mechanism for OCaml

- Project page: <a href="https://github.com/ocaml-multicore/backoff">https://github.com/ocaml-multicore/backoff</a>

## 0.1.0

- Initial version based on backoff from kcas (@lyrm, @polytypic)

### DIFF
--- a/packages/backoff/backoff.0.1.0/opam
+++ b/packages/backoff/backoff.0.1.0/opam
@@ -1,0 +1,38 @@
+opam-version: "2.0"
+synopsis: "Exponential backoff mechanism for OCaml"
+maintainer: ["Carine Morel <carine@tarides.com>"]
+authors: ["KC Sivaramakrishnan <sk826@cl.cam.ac.uk>"]
+license: "ISC"
+homepage: "https://github.com/ocaml-multicore/backoff"
+bug-reports: "https://github.com/ocaml-multicore/backoff/issues"
+depends: [
+  "dune" {>= "3.3"}
+  "ocaml" {>= "4.13"}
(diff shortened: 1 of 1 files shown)


packages/backoff/backoff.0.1.0/opam
```diff
@@ -0,0 +1,38 @@
+opam-version: "2.0"
+synopsis: "Exponential backoff mechanism for OCaml"
+maintainer: ["Carine Morel <carine@tarides.com>"]
+authors: ["KC Sivaramakrishnan <sk826@cl.cam.ac.uk>"]
```
avsm, line 4:
```suggestion
authors: ["KC Sivaramakrishnan <kc@tarides.com>"]
```

I suspect this is sadly long out of date :-) /cc @kayceesrk 

Submitter of PR is different from author/maintainers; should you also be in there somewhere @polytypic?

polytypic: I'm in the LICENSE.md.

For background, there are several exponential backoff implementations in various multicore libraries.  I redesigned the backoff implementation in Kcas to reduce overheads and to allow functions to be efficiently parameterized with configured backoffs while working on Kcas and we are now factoring it out as a separate library (work done by @lyrm) so that it can be conveniently used in other places (e.g. Kcas, Saturn (in general), implementation of scheduler friendly locks, ... any place where someone wants to write code using atomics).  Perhaps at some point we could have backoff in the Stdlib.

avsm:

PR #24228, opened 2023-08-08
Lwt revdeps constraints

raphael-proust:
Addressing revdeps errors which are linked to removal of deprecated values in Lwt. See https://github.com/ocaml/opam-repository/pull/24213#issuecomment-1667998101 for details.

Note that none of the packages changed here are at their latest version. Hence I don't think we need to ping the authors/maintainers.


kit-ty-kate:
Thanks

jchavarri:
> Note that none of the packages changed here are at their latest version. Hence I don't think we need to ping the authors/maintainers.

Ftr this was not true for `aws-s3-lwt`, created an issue for it: https://github.com/andersfugmann/aws-s3/issues/35.

PR #24242, opened 2023-08-09
Package ppx_update.0.81

bluddy:
### `ppx_update.0.81`
PPX library to optimize record updates
ppx_update optimizes away record updates when new record contents match old ones

---
* Homepage: https://github.com/bluddy/ppx_update
* Source repo: git+https://github.com/bluddy/ppx_update.git
* Bug tracker: https://github.com/bluddy/ppx_update/issues

---


bluddy:
The only test that fails now is a unit test on s390x. I measure the time taken for 100,000 mutations vs. avoided mutations in a record and expect the mutations to take longer.

Is there a reason this would fail specifically on the s390x? I use `Sys.time ()`. Could it be inaccurate on the s390x?

packages/ppx_update/ppx_update.0.81/opam
```diff
@@ -0,0 +1,40 @@
...
+opam-version: "2.0"
+synopsis: "PPX library to optimize record updates"
+description:
+  "ppx_update optimizes away record updates when new record contents match old ones"
+maintainer: ["yotambarnoy@gmail.com"]
+authors: ["Yotam Barnoy"]
+license: "MIT"
+tags: ["ppx"]
+homepage: "https://github.com/bluddy/ppx_update"
+doc: "https://url/to/documentation"
```
kit-ty-kate, line 11: This field is optional, if there is no documentation page this should probably be removed
```suggestion
```

bluddy: done

kit-ty-kate (approved):

bluddy:
Thanks guys.

PR #24243, opened 2023-08-09
Mark existing Lwt versions as incompatible with ocaml.5.1

raphael-proust:
Because: https://github.com/ocsigen/lwt/pull/995

Also see: https://github.com/ocaml/opam-repository/pull/24213#issuecomment-1669306812


mseri:
Thanks

kit-ty-kate:
We should've waited for the new release to be merged before merging that because that breaks people's setup if they use OCaml 5.1

raphael-proust:
@kit-ty-kate because then there is no installable version available at all under 5.1?

kit-ty-kate:
yeah exactly

PR #24245, opened 2023-08-09
Revert "Mark existing Lwt versions as incompatible with ocaml.5.1"

kit-ty-kate:
It was merged too hastily and we need more time to review https://github.com/ocaml/opam-repository/pull/24213
Will reapply after https://github.com/ocaml/opam-repository/pull/24213 is merged

PR #24246, opened 2023-08-09
Mark existing Lwt versions as incompatible with OCaml 5.1

kit-ty-kate:
See https://github.com/ocaml/opam-repository/pull/24245
- [x] to merge after https://github.com/ocaml/opam-repository/pull/24213


avsm:
Now suitable to merge as #24213 is in.

PR #24250, opened 2023-08-10
Release Caqti 2.0.0

paurkedal:
This major release includes the new packages caqti-eio, caqti-mirage, and caqti-driver-pgx.  Release notes: https://github.com/paurkedal/ocaml-caqti/releases/tag/v2.0.0


paurkedal:
I can see that many of the remaining errors are due to a lower bound check where the `result` library shadows the stdlib module.  I tried to add
```
conflicts: [
  "result" {< "1.5"}
]
```
to caqti, which avoided the issue for caqti itself (I think, I didn't look though all cases), but does not prevent the package from being downgraded and causing sihl* packages to fail.  Is `conflicts` the right clause to use here, and should be add the cause to the sihl* packages, as well?

paurkedal:
**Please don't merge.** I would like to replace this release with a 2.0.1 release to support Eio from the start, and due to some documentation updates.  Maybe I'll reuse this PR to keep the comments.

paurkedal:
This PR is superseded by #24264.

PR #24251, opened 2023-08-10
[new release] opam-check-npm-deps (1.0.0)

jchavarri:
An opam plugin to check for npm depexts inside the node_modules folder

- Project page: <a href="https://github.com/jchavarri/opam-check-npm-deps">https://github.com/jchavarri/opam-check-npm-deps</a>

##### CHANGES:

- Initial version
- Read `depexts` field from all the opam files of the installed packages in the
  switch
- Check the depexts that include filters using the `npm-version` variable
  against the installed npm packages in `node_modules`

### DIFF
--- a/packages/opam-check-npm-deps/opam-check-npm-deps.1.0.0/opam
+++ b/packages/opam-check-npm-deps/opam-check-npm-deps.1.0.0/opam
@@ -13,7 +13,7 @@ depends: [
   "reason" {>= "3.8.1" & < "4.0.0"}
   "dune" {>= "3.8"}
   "opam-client" {>= "2.1.2" & < "2.2.0"}
-  "mccs" {>= "1.11+14"}
+  "mccs" {>= "1.1+14"}
   "angstrom"
   "bos"
   "lwt_ppx"
--- a/packages/opam-check-npm-deps/opam-check-npm-deps.1.0.0/opam
+++ b/packages/opam-check-npm-deps/opam-check-npm-deps.1.0.0/opam
@@ -35,7 +35,6 @@ build: [
     "-j"
     jobs
     "@install"
-    "@runtest" {with-test}
     "@doc" {with-doc}
   ]
 ]
--- a/packages/opam-check-npm-deps/opam-check-npm-deps.1.0.0/opam
+++ b/packages/opam-check-npm-deps/opam-check-npm-deps.1.0.0/opam
@@ -15,6 +15,7 @@ depends: [
   "opam-client" {>= "2.1.2" & < "2.2.0"}
   "mccs" {>= "1.1+14"}
   "angstrom" {>= "0.15.0"}
+  "fmt" {>= "0.9.0"}
   "bos"
   "lwt_ppx"
   "ppx_deriving_yojson"
--- a/packages/opam-check-npm-deps/opam-check-npm-deps.1.0.0/opam
+++ b/packages/opam-check-npm-deps/opam-check-npm-deps.1.0.0/opam
@@ -12,7 +12,7 @@ depends: [
   "ocaml" {>= "4.14.0"}
   "reason" {>= "3.8.1" & < "4.0.0"}
   "dune" {>= "3.8"}
-  "opam-client" {>= "2.1.2" & < "2.2.0"}
+  "opam-client" {>= "2.1.3" & < "2.2"}
   "mccs" {>= "1.1+14"}
   "angstrom" {>= "0.15.0"}
   "fmt" {>= "0.9.0"}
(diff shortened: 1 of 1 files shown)


jchavarri:
Hm there seems to be some issues with the `mccs` builds, I have no clue if they are related to the package, nor how to solve them.

haochenx:
The mccs lower bounds issue seems to be caused by transitive dependency. It should go away if we can somehow raise lower bounds for mccs to 1.11+14 for now. But since mccs is required as a transitive dependency, we should probably no be specifying its lower bounds in the opam file of this package.

Digging deeper, it seems that the problem might [have something to do with the version of c++ compiler installed on the system](https://stackoverflow.com/questions/51235355/comparison-object-being-invocable-as-const), which seems to be quite tricky to resolve perfectly here without touching the upstream (or patching).

Routes that can be taken IMO:
1. directly add lower bounds of mccs (i.e. list mccs >= 1.11+14 as a dependency of this package)
    - dirty but quick hack to merge this otherwise perfect packages
2. add lower bounds of mccs to opam-solver's opam file
3. patch mccs

If (2)-(3) is taken, whether to merge this packages is arguable, as mccs.1+11 could well be installable on many other machines. But even if we do not want to merge it as is, we can temporarily do (1) while waiting for (2)-(3).

haochenx:
ping: @AltGr @rjbou

jchavarri:
@haochenx thanks for the pointers, I applied the approach specified in 1 to the PR, and the errors are gone (not sure why GitHub still shows the check with "Failed" state).

haochenx:
@jchavarri thank you for adopting (1)

I think with the step taken, this package is mergeable. But I think we the opam-repository maintainers should still take steps to fix this problem. I will create an issue shortly.

haochenx:
- Issue filed: https://github.com/ocaml/opam-repository/issues/24256

jchavarri:
I think all ci checks have been fixed. There is some remaining failure in the amd64 builds on macos but they are marked as experimental, so I understand they are not a blocker.

packages/opam-check-npm-deps/opam-check-npm-deps.1.0.0/opam
```diff
@@ -0,0 +1,53 @@
...
+  "fmt" {>= "0.9.0"}
+  "bos"
+  "lwt_ppx"
+  "ppx_deriving_yojson"
+  "ppx_expect"
+  "ppx_inline_test"
+  "ppx_let"
+  "ppx_sexp_conv"
+  "ocaml-lsp-server" {dev}
+  "ocamlformat" {dev}
```
kit-ty-kate, line 27: The `dev` variable is not meant to be used to install packages that are meant to only be useful for your own dev setup (see your own PR ^^ https://github.com/ocaml/opam/issues/4959)
```suggestion
```

packages/opam-check-npm-deps/opam-check-npm-deps.1.0.0/opam
```diff
@@ -0,0 +1,53 @@
...
+  "odoc" {with-doc}
+]
```
kit-ty-kate, line 29: This is necessary to avoid having the client mismatch the version of the library requested as dependency and slowing up both in some cases: https://github.com/ocaml/opam/issues/4554. There might also be similar unpatched issues so i think we should be careful here
```suggestion
]
available: opam-version >= "2.1" & opam-version < "2.2"
```

packages/opam-check-npm-deps/opam-check-npm-deps.1.0.0/opam
```diff
@@ -0,0 +1,53 @@
...
+maintainer: ["Javier Chávarri"]
+authors: ["Javier Chávarri"]
+license: "MIT"
+homepage: "https://github.com/jchavarri/opam-check-npm-deps"
+bug-reports: "https://github.com/jchavarri/opam-check-npm-deps/issues"
+depends: [
+  "ocaml" {>= "4.14.0"}
+  "reason" {>= "3.8.1" & < "4.0.0"}
+  "dune" {>= "3.8"}
+  "opam-client" {>= "2.1.2" & < "2.2.0"}
```
kit-ty-kate, line 15: It’s better to have at least 2.1.3 to have this patch: https://github.com/ocaml/opam/issues/4554
Also `< "2.2.0"` also includes 2.2.0~alpha which is probably not what you want. `< "2.2"` will remove that from the selected set
```suggestion
  "opam-client" {>= "2.1.3" & < "2.2"}
```

jchavarri:
thanks for the review @kit-ty-kate ! I think I applied all the change requests, and CI seems to be happy now :)

packages/opam-check-npm-deps/opam-check-npm-deps.1.0.0/opam
jchavarri: right ... 🤦 i was too used to `dev` already, will keep in mind!

kit-ty-kate:
Thanks!

PR #24254, opened 2023-08-11
fix metadata typos in pyast opam files

haochenx:
There seems to be typos in the hompage, doc, bug-reports, and dev-repo fields in pyast's opam files.


haochenx:
just let you know: @thierry-martinez

mseri:
Thanks

PR #24268, opened 2023-08-14
Package llvm.15.0.7+nnp-1

kit-ty-kate:
### `llvm.15.0.7+nnp-1`
The OCaml bindings distributed with LLVM
Note: LLVM should be installed first.

---
* Homepage: http://llvm.moe
* Source repo: git+https://github.com/kit-ty-kate/llvm-dune.git
* Bug tracker: http://llvm.org/bugs/

---

### DIFF
--- a/packages/conf-llvm/conf-llvm.15/opam
+++ b/packages/conf-llvm/conf-llvm.15/opam
@@ -13,7 +13,7 @@ depends: [
 depexts: [
   ["llvm@15"] {os-distribution = "homebrew" & os = "macos"}
   ["llvm-15"] {os-distribution = "macports" & os = "macos"}
-  ["llvm-15-dev"] {os-family = "debian"}
+  ["llvm-15-dev" "zlib1g-dev"] {os-family = "debian"}
   ["llvm15-dev"] {os-distribution = "alpine"}
   ["llvm"] {os-family = "arch"}
   ["llvm15-devel"] {os-family = "suse"}
--- a/packages/conf-llvm/conf-llvm.15/opam
+++ b/packages/conf-llvm/conf-llvm.15/opam
@@ -13,7 +13,7 @@ depends: [
 depexts: [
   ["llvm@15"] {os-distribution = "homebrew" & os = "macos"}
   ["llvm-15"] {os-distribution = "macports" & os = "macos"}
-  ["llvm-15-dev" "zlib1g-dev"] {os-family = "debian"}
+  ["llvm-15-dev" "zlib1g-dev" "libzstd-dev"] {os-family = "debian"}
   ["llvm15-dev"] {os-distribution = "alpine"}
   ["llvm"] {os-family = "arch"}
   ["llvm15-devel"] {os-family = "suse"}
--- a/packages/conf-llvm/conf-llvm.15/opam
+++ b/packages/conf-llvm/conf-llvm.15/opam
@@ -1,0 +1,26 @@
+opam-version: "2.0"
+maintainer: "Kate <kit.ty.kate@disroot.org>"
+authors: "The LLVM team"
+homepage: "http://llvm.org"
+bug-reports: "https://llvm.org/bugs/"
+license: "MIT"
+build: [
+  ["bash" "configure.sh" version]
+]
+depends: [
+  "conf-bash" {build}
+]
+depexts: [
+  ["llvm@15"] {os-distribution = "homebrew" & os = "macos"}
+  ["llvm-15"] {os-distribution = "macports" & os = "macos"}
+  ["llvm-15-dev"] {os-family = "debian"}
+  ["llvm15-dev"] {os-distribution = "alpine"}
+  ["llvm"] {os-family = "arch"}
+  ["llvm15-devel"] {os-family = "suse"}
+  ["llvm15-devel"] {os-distribution = "fedora"}
+  ["llvm15-devel" "epel-release"] {os-distribution = "centos"}
+  ["devel/llvm15"] {os = "freebsd"}
+]
+synopsis: "Virtual package relying on llvm library installation"
+extra-files: ["configure.sh" "md5=633155a6495a7afd1c87ffd0b94e8cf9"]
+flags: conf
--- a/packages/conf-llvm/conf-llvm.15/files/configure.sh
+++ b/packages/conf-llvm/conf-llvm.15/files/configure.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+set -e
+set -x
+
+version=${1/%.?.?/}
+
+if hash brew 2>/dev/null; then
+    brew_llvm_config="$(brew --cellar)"/llvm*/${version}*/bin/llvm-config
+fi
+
+shopt -s nullglob
+for llvm_config in llvm-config-$version llvm-config-${version}.0 llvm-config${version}0 llvm-config-mp-$version llvm-config-mp-${version}.0 llvm${version}-config llvm-config-${version}-32 llvm-config-${version}-64 llvm-config $brew_llvm_config; do
+    llvm_version="`$llvm_config --version`" || true
+    case $llvm_version in
+    $version*)
+        echo "config: \"$llvm_config\"" >> conf-llvm.config
+        echo "version: \"$llvm_version\"" >> conf-llvm.config
+        exit 0;;
+    *)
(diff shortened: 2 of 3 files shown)


kit-ty-kate:
cc @alan-j-hu @Kakadu

packages/conf-llvm/conf-llvm.15/opam
```diff
@@ -0,0 +1,26 @@
...
+build: [
+  ["bash" "configure.sh" version]
+]
+depends: [
+  "conf-bash" {build}
+]
+depexts: [
+  ["llvm@15"] {os-distribution = "homebrew" & os = "macos"}
+  ["llvm-15"] {os-distribution = "macports" & os = "macos"}
+  ["llvm-15-dev"] {os-family = "debian"}
```
kit-ty-kate, line 16: That looks like a debian bug (it should be a dependency), but oh well
```suggestion
  ["llvm-15-dev" "zlib1g-dev"] {os-family = "debian"}
```

kit-ty-kate: cc @sylvestre as you seem to be the LLVM package maintainer in Debian
```
# /usr/bin/ld: cannot find -lz: No such file or directory
```
and links fine if `zlib1g-dev` is installed.
The `-lz` comes from `llvm-config --system-libs --link-static --libs Core`

packages/conf-llvm/conf-llvm.15/opam
```diff
@@ -0,0 +1,26 @@
...
+build: [
+  ["bash" "configure.sh" version]
+]
+depends: [
+  "conf-bash" {build}
+]
+depexts: [
+  ["llvm@15"] {os-distribution = "homebrew" & os = "macos"}
+  ["llvm-15"] {os-distribution = "macports" & os = "macos"}
+  ["llvm-15-dev" "zlib1g-dev"] {os-family = "debian"}
```
kit-ty-kate, line 16:
```suggestion
  ["llvm-15-dev" "zlib1g-dev" "libzstd-dev"] {os-family = "debian"}
```
Debian bug once again, this time on Debian 13 and newer only (stable works fine for some reason)

cc @sylvestre

alan-j-hu:
I want to make sure I understand the change from OCamlfind to Dune. Is the following correct?

- Prior opam releases of LLVM use OCamlfind, which is supported via the in-tree META file. The package has a script that installs versions of the bindings that link statically or dynamically to LLVM (if each is available), by patching the META file to use the predicate llvm.static to choose between them.
- The llvm.static predicate can be set using OCamlfind but cannot be set from Dune.
- This release uses Dune, which it supports by running a script that generates Dune files in the tree. The script's generated Dune files use the virtual library feature to define versions of the library that link statically or dynamically to LLVM (if each is available).
- OCamlfind-based packages don't understand the Dune virtual library feature and therefore cannot use the "default implementation" and must select the static or dynamic implementation explicitly.
- You believe that the package builds faster and is easier to maintain with Dune.

Context: https://discuss.ocaml.org/t/ask-for-suggestions-for-the-next-z3-package-release/8922/4 and https://discuss.ocaml.org/t/using-output-of-ocamlfind-params-only-show-in-a-dune-build/4187.

Please correct me if I misunderstood any aspect of this.

kit-ty-kate:
yep everything is spot-on

Kakadu:
I found some time to check that PR. 

1) It looks like the function `Llvm.set_opaque_pointers context false;` has disappeared some where, but it is mentioned in documentation comments of Llvm module
2) My code started to generate ill-typed IR. Don't know what is the reason yet....

packages/conf-llvm/conf-llvm.15/opam
sylvestre: you suggest that I should add this package in the build dep? right?

alan-j-hu:
Good catch, and my apologies. I neglected to update the deprecation messages to remove the mention of the set_opaque-pointers function. I made a PR at https://github.com/kit-ty-kate/llvm-dune/pull/13.

packages/conf-llvm/conf-llvm.15/opam
kit-ty-kate: yes, both `zlib1g-dev` and `libzstd-dev`

kit-ty-kate:
> My code started to generate ill-typed IR. Don't know what is the reason yet....

did you manage to track down where it was coming from?

alan-j-hu:
Based on its source code, build_call2 seems to expect a function type: https://github.com/llvm/llvm-project/blob/8dfdcc7b7bf66834a761bd8de445840ef68e4d1a/llvm/lib/IR/Core.cpp#L3918

Kakadu:
Could you look at this?
https://gist.github.com/Kakadu/3e2f80864dc5f1d2408a77e8710e1df2

alan-j-hu:
Okay, I investigated. The following change fixes the issue:

```diff
diff --git a/demo.ml b/demo.ml
index f25d4d6..d88af85 100644
--- a/demo.ml
+++ b/demo.ml
@@ -46,7 +46,7 @@ let () =
     (Llvm.string_of_lltype (Llvm.type_of applyN));
   Llvm.dump_module module_;
   let return_val =
-    Llvm.build_call2 (Llvm.type_of applyN) applyN
+    Llvm.build_call2 (Llvm.var_arg_function_type i64_type [| i64_type; i64_type |]) applyN
       [|
         Llvm.const_int i64_typ 1;
         Llvm.const_int i64_typ 2;
```

The following change also fixes the issue:

```diff
diff --git a/demo.ml b/demo.ml
index f25d4d6..16838ef 100644
--- a/demo.ml
+++ b/demo.ml
@@ -24,10 +24,11 @@ let () =
   in
   let add_virt_binding ~key v = Hashtbl.add virt_of_named_hash key v in
 
+  let fty = Llvm.var_arg_function_type i64_type [| i64_type; i64_type |] in
   let _ =
     (* void* lama_applyN(void* f, int32_t, ...) *)
     Llvm.declare_function "rukaml_applyN"
-      (Llvm.var_arg_function_type i64_type [| i64_type; i64_type |])
+      fty
       module_
   in
 
@@ -46,7 +47,7 @@ let () =
     (Llvm.string_of_lltype (Llvm.type_of applyN));
   Llvm.dump_module module_;
   let return_val =
-    Llvm.build_call2 (Llvm.type_of applyN) applyN
+    Llvm.build_call2 fty applyN
       [|
         Llvm.const_int i64_typ 1;
         Llvm.const_int i64_typ 2;
```

This C program:


```c
#include <llvm-c/Analysis.h>
#include <llvm-c/Core.h>
#include <stddef.h>

int main()
{
  LLVMContextRef ctx = LLVMGetGlobalContext();
  LLVMContextSetOpaquePointers(ctx, 0);
  LLVMTypeRef opaque = LLVMPointerTypeInContext(ctx, 0);
  LLVMModuleRef m = LLVMModuleCreateWithNameInContext("mod", ctx);
  LLVMTypeRef fty = LLVMFunctionType(LLVMInt32Type(), NULL, 0, 0);
  LLVMValueRef main = LLVMAddFunction(m, "main", fty);
  LLVMBasicBlockRef bb = LLVMAppendBasicBlockInContext(ctx, main, "entry");
  LLVMBuilderRef b = LLVMCreateBuilderInContext(ctx);
  LLVMPositionBuilderAtEnd(b, bb);
  LLVMValueRef call = LLVMBuildCall2(b, LLVMTypeOf(main), main, NULL, 0, "rec");
  LLVMBuildRet(b, call);
  LLVMDumpModule(m);
  LLVMVerifyFunction(main, LLVMPrintMessageAction);
  LLVMDisposeModule(m);
  return 0;
}
```

compiled with `` cc test.c `llvm-config-15 --cflags --ldflags --libs core analysis` ``

produces the following output:


```
; ModuleID = 'mod'
source_filename = "mod"

define i32 @main() {
entry:
  %rec = call i32 () @main()
  ret i32 () %rec
}
Called function is not the same type as the call!
  %rec = call i32 () @main()
Function return type does not match operand type of return inst!
  ret i32 () %rec
 i32
 ```

whereas using `fty` instead of `LLVMTypeOf(main)` works.

There must be something funky with `LLVMTypeOf` that is not the fault of the OCaml bindings.

Kakadu:
Many thanks! I approve the changes.

kit-ty-kate:
Superseded by https://github.com/ocaml/opam-repository/pull/24268 to integrate https://github.com/kit-ty-kate/llvm-dune/pull/13

We should now be finally good to go! Thank you all!

PR #24286, opened 2023-08-16
Add a Github Action that removes the "needs reporter action" label when the issue or PR has been commented on

kit-ty-kate:
Labeling PRs and issues allows opam-repository maintainers to more efficiently see which PRs need a review or which ones are waiting for inputs from the person who opened the PR/issue.

This PR adds a bot to opam-repository which will be triggered everytime there is a comment and removes the label `needs reporter action` so that we don’t skip the PR if we missed the notification that the action has been taken.

### DIFF
--- a/.github/workflows/unlabel.yml
+++ b/.github/workflows/unlabel.yml
@@ -4,8 +4,9 @@ on:
     types: [created]
 jobs:
   remove_label:
+    if: ${{ !contains(fromJSON(vars.MAINTAINERS), github.event.comment.user.login) && contains(github.event.issue.labels.*.name, fromJSON(vars.LABEL_TO_REMOVE)) }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions-ecosystem/action-remove-labels@v1
         with:
-          labels: "needs reporter action"
+          labels: ${{ fromJSON(vars.LABEL_TO_REMOVE) }}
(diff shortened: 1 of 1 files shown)


.github/workflows/unlabel.yml
```diff
@@ -0,0 +1,11 @@
+name: Remove Label
+on:
+  issue_comment:
```
mseri, line 3: I would have done it on PR commits, but perhaps makes more sense to remove it on comments. Is it possible to avoid running it when one of us is commenting?

kit-ty-kate: doing it on PR commits is fairly annoying to do as it will add that one action in the checks for the PR and for example if our CI is broken and no output is result has been sent yet or github broke somehow it will show the PR as "checked and good to go" even though it is very not. There is a github feature request asking for a way to hide them but it hasn't seen any activity in 2 years: https://github.com/orgs/community/discussions/26246

> Is it possible to avoid running it when one of us is commenting?

I've done that though. I've used github variables (https://docs.github.com/en/actions/learn-github-actions/variables) to store that information and the name of the label too, to make it easier to modify it in the future.

avsm:
An excellent idea, thanks Kate!

mseri (approved):

raphael-proust (approved):
Looks good to me. I'm not super familiar with github actions but it looks alright.

PR #24289, opened 2023-08-17
Octez 17.3

killian-delarue:
Packages for Octez v17.3

### DIFF
--- a/packages/octez-accuser-PtMumbai/octez-accuser-PtMumbai.17.2/opam
+++ b/packages/octez-accuser-PtMumbai/octez-accuser-PtMumbai.17.2/opam
@@ -23,6 +23,7 @@ build: [
   ["dune" "runtest" "-p" name "-j" jobs] {with-test}
 ]
 synopsis: "Tezos/Protocol: accuser binary"
+available: false
 url {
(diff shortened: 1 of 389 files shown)


raphael-proust:
> And here we go again with a large swathe of packages being marked unavailable, and no sign of a reduction of the number of individual packages that comprise the Octez release.

This is a security fix for the v17 release branch. Essentially it replaces 17.2 packages with 17.3 packages, the delta being a (security) bugfix. See the changelog for this release: https://tezos.gitlab.io/CHANGES.html#version-17-3

I think that it's reasonable to set 17.2 packages as unavailable.

> Do you have a roadmap for when these will be reduced in number, and the churn on available packages in the public opam repository reduced?

The v18 branch already includes a reduction in the number of packages. Currently it's on the order of 25 packages. (I don't remember the exact number, maybe @killian-delarue has an exact figure?) There are further reductions planned too (notably, https://gitlab.com/tezos/tezos/-/merge_requests/9478 which removes even more packages; again, if @killian-delarue might have an exact figure).

killian-delarue:
> 

This release is a security update of Otcez v17 that does not contain the reduction of packages.

A first step toward the full reduction will be introduced in Octez v18.0.

PR #24306, opened 2023-08-24
[new release] moonpool (0.4)

c-cube:

##### CHANGES:

- add `Fut.{reify_error,bind_reify_error}`
- full lifecycle for worker domains, where a domain
    will shutdown if no thread runs on it, after a
    short delay.

- fix: generalize type of `create_arg`
- perf: in `Bb_queue`, only signal condition on push if queue was empty

### DIFF
--- a/packages/moonpool/moonpool.0.4/opam
+++ b/packages/moonpool/moonpool.0.4/opam
@@ -1,0 +1,44 @@
+opam-version: "2.0"
+synopsis: "Pools of threads supported by a pool of domains"
+maintainer: ["Simon Cruanes"]
+authors: ["Simon Cruanes"]
+license: "MIT"
+tags: ["thread" "pool" "domain"]
+homepage: "https://github.com/c-cube/moonpool"
+bug-reports: "https://github.com/c-cube/moonpool/issues"
+depends: [
+  "ocaml" {>= "4.08"}
+  "dune" {>= "3.0"}
+  "either" {>= "1.0"}
+  "trace" {with-test}
+  "qcheck-core" {with-test & >= "0.19"}
+  "odoc" {with-doc}
+  "mdx" {>= "1.9.0" & with-test}
+]
+depopts: [
+  "domain-local-await" {>= "0.2"}
+]
+build: [
+  ["dune" "subst"] {dev}
+  [
+    "dune"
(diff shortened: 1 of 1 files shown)


c-cube:
So that's going to be interesting to debug. What machine does this run on?

avsm:
How many domains are the tests trying to create? It'll be running on a many core x86 machine in that first test, but may be running into cgroups resource limits.

c-cube:
Potentially n-1, where n is recommended domain count. But this time it will create many more domains than that over the life time of the test, it just has at most n-1 alive at a time.

c-cube:
So this is quite puzzling. I tried on a c3.4xlarge instance and a friend tried on his M1, to no avail — the test always passes. Can it be the cgroup settings?

avsm:
It's very likely to do with cgroups restrictions in opam-repo-ci. /cc @mtelvers to perhaps shed light on this.

mtelvers:
This has been a real puzzle for me this afternoon, taking a long look a cgroup limits on the cluster workers, running the spec file on different workers, and getting some expert advice from @dra27.  The test fails when `nproc` >= `Max_domains`.  The machines where it fails include: cree, pima and marpe which have `nproc` of 128, 128 and 256, respectively.  On smaller machines where `nproc` < 127, there is no issue.  This has also been tested as a bare metal installation on pima (no Docker or OBuilder) and the results are the same.

dra27:
I'm not sure where the bug lies, but the domains are never being joined. I was able to reproduce the failures very easily with a custom compiler with `Max_domains` at 16 on the same machine (that triggered even more test failures with the same exception). I hacked together some domain joining logic (the hack was to record the `Domain.t` with the domain state and always join on it before re-using the slot). With that version, I'm not seeing the crash.

c-cube:
Thank you @mtelvers and @dra27 for the thorough investigation and the fix!!

avsm:
Such a success disaster, that we have machines with so many cores that we're blowing past the maximum available domains in OCaml already ;-)  Thank you for the thorough investigation @mtelvers and @dra27, and @c-cube for the revised package!

c-cube:
Thank y'all again for the help! That was a tricky one :)

PR #24309, opened 2023-08-24
reason-react: add depexts

jchavarri:
Now that `opam-check-npm-deps` plugin [has been published](https://github.com/ocaml/opam-repository/pull/24251), this change makes use of it by adding a `depexts` field to the `reason-react` package to reflect the dependency on the underlying `react` and `react-dom` npm packages.

I am not sure how to reflect the dependency on the opam plugin version 1.0.0, so I am using the soon to be added `with-dev-setup` variable for it, as this will be mostly a development time check.

### DIFF
--- a/packages/reason-react/reason-react.0.11.0/opam
+++ b/packages/reason-react/reason-react.0.11.0/opam
@@ -22,7 +22,7 @@ depends: [
   "reactjs-jsx-ppx"
   "reason" {>= "3.6.0"}
   "ocaml-lsp-server" {with-test}
-  "opam-check-npm-deps" {"1.0.0" = with-dev-setup}
+  "opam-check-npm-deps" {= "1.0.0" & with-dev-setup}
   "odoc" {with-doc}
 ]
 build: [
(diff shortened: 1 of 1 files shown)


packages/reason-react/reason-react.0.11.0/opam
```diff
@@ -22,6 +22,7 @@ depends: [
   "reactjs-jsx-ppx"
   "reason" {>= "3.6.0"}
   "ocaml-lsp-server" {with-test}
+  "opam-check-npm-deps" {"1.0.0" = with-dev-setup}
```
mseri, line 25:
```suggestion
  "opam-check-npm-deps" {= "1.0.0" & with-dev-setup}
```
The way it was written cannot work. Do you mean that this should be installed with exactly with version 1.0.0 and only when using a new dev-setup flag?

jchavarri: Yes, that's what I meant, thanks!

mseri:
Thanks

PR #24316, opened 2023-08-25
DkSDKFFIOCaml_Std +Export-linux_x86_64 1.0.0~1

jonahbeckford:
First preview release of DkSDK FFI OCaml.

I believe the following:

```
[(lint) (failed: Error in DkSDKFFIOCaml_Std.1.0.0~1: warning 41: Some packages are mentioned in package scripts or features, but there is no dependency or depopt toward them: "DkSDKFFIOCaml_StdExport-linux_x86_64")](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/da15c1287cf3a23b86a1f55258972680ae118bf7/variant/(lint))
```

is harmless and only present because I'm not publishing each package individually.


packages/DkSDKFFIOCaml_Std/DkSDKFFIOCaml_Std.1.0.0~1/opam
```diff
@@ -0,0 +1,59 @@
...
+# This file is generated by dune, edit dune-project instead
+opam-version: "2.0"
+synopsis: "DkSDK FFI for OCaml"
+description: """
+DkSDK FFI OCaml provides within-process communication from OCaml to another DkSDK FFI language (ex. DkSDK FFI Java).
+This opam package downloads pre-built binaries for a limited set of platforms to avoid the C build tool differences between DkSDK and OCaml.
+Full source code and other platforms are available with a DkSDK license from https://diskuv.com/pricing, and is free for security engineers, educators and related-field researchers on request."""
+maintainer: ["Diskuv, Inc. <opensource+dkml@support.diskuv.com>"]
+authors: ["Diskuv, Inc. <opensource+dkml@support.diskuv.com>"]
+license: "OSL-3.0"
```
avsm, line 11: Just checking on the licensing requirements here. My understanding of OSL 3.0 is that source code must be distributed along with the binaries; is that the case here? (It's fine to have a closed source package, but the license field here has to be accurate).

jonahbeckford: Q1: Regarding the download of a binary: The https://rosenlaw.com/OSL3.0-explained.htm explainer section "Source Code Disclosure" goes over linking code. The first bullet point precisely describes what is happening:
> "If linking (by whatever technical means) can be accomplished by making and using unmodified copies of the Original Work, then 1(a) and 1(c) permit that; only the Source Code of the Original Work must be disclosed."

That can be dissected:
- the `extra-source` of the `opam` file is the "by whatever technical means"; the `opam` file is part of the Original Work.
- the `foreign_archive` in the `dune` file is the "linking"; again, the `dune` files are part of the Original Work
- the fact that it builds with `opam install` is evidence that it "can be accomplished by making and using unmodified copies of the Original Work" (more on this below)
- All of that implies "only the Source Code of the Original Work must be disclosed".

Q2: Why does what builds with `opam install` == the "Original Work"?

- OSL-3.0 starts with `This Open Software License (the “License”) applies to any original work of authorship (the “Original Work”) whose owner (the “Licensor”) has placed the following licensing notice adjacent to the copyright notice for the Original Work`. And it has section 3: `The term “Source Code” means the preferred form of the Original Work for making modifications to it and all available documentation describing how to modify the Original Work.`. That means Diskuv, and nobody else, defines what the Original Work is. And I do! Each source code file required to build the opam packages has an OSL-3.0 licensing notice.

But this reminds me that I can put the licensing notice and the copyright notice in the `opam` file for clarity (even though OSL-3.0 says that is not required). **I'll do that.**

PS. I had the licensing reviewed by an attorney a few weeks ago. But do not construe the above as legal advice. Regardless, I think the above is pretty clear, which is why I really like OSL-3.0.

---

There is an open question I've had for a very long time about why we don't model our transitive licenses accurately. If I look at the [zarith.1.13/opam](https://github.com/ocaml/opam-repository/blob/master/packages/zarith/zarith.1.13/opam) it claims to be:
> `license: "LGPL-2.0-only WITH OCaml-LGPL-linking-exception"`

which is only accurate if the "license" field is not transitive. Why? Because GMP (the library that `zarith` links to) has a flatly incompatible license at https://gmplib.org/manual/Copying: LGPL3+ with no exceptions, or GPL2+ with no exceptions.

And that `zarith/libgmp` situation may seem superficially similar to `DkSDKFFIOCaml/stdexport-linux_x86_64.tar.gz`. But the former has two incompatible licenses and two copyright owners, while the latter has compatible licenses and one copyright owner.

Having said that, I'm not willing to be the test case for transitive dependencies.

packages/DkSDKFFIOCaml_Std/DkSDKFFIOCaml_Std.1.0.0~1/opam
```diff
@@ -0,0 +1,59 @@
...
+homepage: "https://diskuv.com/cmake/help/latest"
+bug-reports:
+  "https://gitlab.com/diskuv/distributions/1.0/dksdk-ffi-ocaml/-/issues"
```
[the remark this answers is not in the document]
jonahbeckford: Will fix or remove it. Thanks for noticing.

packages/DkSDKFFIOCaml_Std/DkSDKFFIOCaml_Std.1.0.0~1/opam
avsm: I've gone over your (much appreciated and thorough) explanation and am satisfied by it; I'm glad you've found a licensing solution that meets both your business needs and for use in the community opam repository.  It might be worth a short FAQ somewhere (perhaps the `description` field in the package) to capture your explanation for users.

jonahbeckford: Updated description.

avsm:
Thanks for the fixes, now merged!

PR #23771, opened 2023-05-09
Add package for Topiary v0.1.0

Niols:
This pull requests adds a package for [Topiary](https://github.com/tweag/topiary) in version 0.1.0 (“Benevolent Beech”). From the point of view of opam-repository and OCaml users, Topiary is an alternative formatter for OCaml. The upstream changelog can be found [here](https://github.com/tweag/topiary/blob/556b2a10df3894a905f2654f774c39933c1f439a/CHANGELOG.md#010-benevolent-beech---2023-03-09).

The main point to note about this package is that is is _not_ an OCaml package: Topiary is written in Rust. However, it can be interesting to publish in opam-repository because it allows formatting OCaml files and could be useful for the community. Some context on this question as well as the idea behind the packaging solution can be found in [this discuss thread](https://discuss.ocaml.org/t/two-questions-about-what-is-appropriate-to-package-with-opam/12030/).

Apart from the problem of whether we accept packaging a Rust program in opam, there is also a technical issue: Topiary requires Cargo/Rustc in version at least 1.65. This version is fairly recent and might not be present in all distributions. In fact, the CI of the [tweag/topiary-opam](https://github.com/tweag/topiary-opam) repository keeps track of this. Out of 11 distributions provided by the `ocaml/opam` DockerHub images:

- 3 can install Topiary without issues (Fedora, Ubuntu and Ubuntu LTS)
- 5 cannot install Topiary due to not having a recent enough version of Cargo/Rustc (Alpine, Debian stable, Debian testing, Debian unstable, Oracle Linux)
- 3 cannot install Topiary due to other problems (Archlinux, CentOS, openSUSE) (The problem with Archlinux is only due to a download issue and I expect it to be resolved rather soon in the `ocaml/opam` DockerHub images.)

The errors are always fairly understandable. In the case of Debian stable that does not even have the 2021 edition of Cargo/Rustc:
```
<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+- The following actions failed
| - build conf-rust-2021 1
+- 
- No changes have been performed
[ERROR] The compilation of conf-rust-2021 failed at "/usr/bin/rustc --edition 2021 test.rs".

#=== ERROR while compiling conf-rust-2021.1 ===================================#
# context              2.0.10 | linux/x86_64 | ocaml-base-compiler.5.0.0 | file:///home/opam/opam-repository
# path                 ~/.opam/5.0/.opam-switch/build/conf-rust-2021.1
[...]

PR #24319, opened 2023-08-25
Package rsdd.0.1.0

mattxwang:
### `rsdd.0.1.0`
Bindings for RSDD
OCaml bindings for the Rust Decision Diagram (RSDD) library, generated with ocaml-rs.

---
* Homepage: https://github.com/neuppl/rsdd-ocaml
* Source repo: git+https://github.com/neuppl/rsdd-ocaml.git
* Bug tracker: https://github.com/neuppl/rsdd-ocaml/issues

---


mattxwang:
Note for others: I believe we're in a similar boat as https://github.com/ocaml/opam-repository/pull/23771 in terms of Rust versions supported on CI. I'm going to do a bit more digging and unmark this as draft + write more in the PR description.

mattxwang:
Hm, I'm not sure why I'm getting an error about `/home/opam/.cargo` being read-only - this wasn't an issue when I first submitted this PR. Doing some more investigating 🔍

mattxwang:
Could I get some guidance on how to deal with the `freebsd (experimental)` CI failure? 

For context, these are OCaml bindings for a Rust library. I'm relying on `conf-rust-2021` to build the library; `cargo` (and thus, `conf-rust-2021`) aren't available on the experimental image which causes an error. I thought I'd have to then use `x-ci-accept-failures`, but I don't think I'm doing this properly (see: [37dbe15](https://github.com/ocaml/opam-repository/pull/24319/commits/37dbe151993a94d7c0f4de0ee8b481d8e44f6cc3)).

packages/rsdd/rsdd.0.1.0/opam
```diff
@@ -0,0 +1,45 @@
...
+opam-version: "2.0"
+synopsis: "Bindings for RSDD"
+description:
+  "OCaml bindings for the Rust Decision Diagram (RSDD) library, generated with ocaml-rs."
+maintainer: ["Matt Wang <mxw@cs.washington.edu>"]
+authors: [
+  "Steven Holtzen <s.holtzen@northeastern.edu>"
+  "Matt Wang <mxw@cs.washington.edu>"
+]
+license: "MIT"
```
haochenx, line 11: There does not seem to be a license on the upstream / in the tarball.
Is it possible for you to add it?

packages/rsdd/rsdd.0.1.0/opam
```diff
@@ -0,0 +1,45 @@
...
+homepage: "https://github.com/neuppl/rsdd-ocaml"
+bug-reports: "https://github.com/neuppl/rsdd-ocaml/issues"
+depends: [
+  "ocaml" {>= "4.08.0" & < "5"}
+  "dune" {>= "2.7" & >= "3.0"}
```
haochenx, line 16: can be simply `{>= "2.7"}`. this is likely due to having a version constraint on [this line](https://github.com/neuppl/rsdd-ocaml/blob/9bb86e392e04e8455f8d6088efa3f1999ee11830/dune-project#L18C1-L18C1)

 If you are depending on any features / fixes that are only available on version 3.0 or later, it's probably a better idea to have `(lang dune 3.0)` on [this line](https://github.com/neuppl/rsdd-ocaml/blob/9bb86e392e04e8455f8d6088efa3f1999ee11830/dune-project#L1)

packages/rsdd/rsdd.0.1.0/opam
```diff
@@ -0,0 +1,45 @@
...
+url {
+  src:
+    "https://github.com/neuppl/rsdd-ocaml/archive/refs/tags/0.1.0-alpha.tar.gz"
+  checksum: [
+    "md5=83caf7c551b6cfb023338f080550a56f"
+    "sha512=83a514a9b0b7d139597b9eb0afe85a8a86b20929390f192ad9c7954d269a773cea69b964ac0eca04989ffaa43b851bb68b90442a0ce1347f08ef805f5c26d0e1"
+  ]
+}
+x-ci-accept-failures: [
+	  "freebsd (experimental)" # cargo (and thus conf-rust-2021) is not available
```
haochenx, line 44:
```suggestion
	  "freebsd" # cargo (and thus conf-rust-2021) is not available
```

packages/rsdd/rsdd.0.1.0/opam
mattxwang: Appreciate it (and I don't think we're depending on dune 3); this was autogenerated earlier and I hadn't realized that there was a conflict.

haochenx: Thanks for taking actions! Just for you info: https://github.com/neuppl/rsdd-ocaml/commit/fc2965dac9a63c061445a3e6e493bcac9e3e053d#r129203968

packages/rsdd/rsdd.0.1.0/opam
haochenx: Also for your information, you can have a [`rsdd.opam.template` file](https://dune.readthedocs.io/en/stable/reference/packages.html#:~:text=If%20a%20pkg.opam.template%20file%20is%20found) in your repository to have Dune generated opam file contain these fields from the beginning (so to avoid the hassle to add them manually for every release)

mattxwang: Thanks! I had tried just using `"freebsd"` earlier and it still reported an `X` in the CI run; for the commit that I just force-pushed with `opam publish`, the [CI is still reporting the same error](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/46550be36bb9867df897692b8913e55aa2349835) (ex: [freebsd on 4.14](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/46550be36bb9867df897692b8913e55aa2349835/variant/freebsd,freebsd-ocaml-4.14-amd64,rsdd.0.1.0)).

packages/rsdd/rsdd.0.1.0/opam
haochenx: Thanks!

haochenx (approved):
It seems that the CI is still failing for freebsd. I will ask other maintainers what's the correct way to ignore for it.

But that should not be a blocker for this PR and it looks mergeable to me!

mseri:
The failure is on conf-rust, in this case I would suggest to accept the fact that the conf package is failing and accept the package as is

packages/rsdd/rsdd.0.1.0/opam
```diff
@@ -0,0 +1,44 @@
...
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
+dev-repo: "git+https://github.com/neuppl/rsdd-ocaml.git"
+x-ci-accept-failures: [
+	  "freebsd" # cargo (and thus conf-rust-2021) is not available
```
mseri, line 36:
```suggestion
	  "freebsd" # cargo (and thus conf-rust-2021) is not available
```
No need to have it here, this needs to go to conf-rust in case

mseri:
Builds fine now. I have reverted the change and will merge as soon as it passes the lint

mattxwang:
Thanks everyone for your help, much appreciated!

PR #24320, opened 2023-08-26
Package ocolor.1.3.1

marc-chevalier:
### `ocolor.1.3.1`

---

---


marc-chevalier:
CI is panicking because of revdeps test:
[simlog.0.0.3 (failed: Failed to get sources of simlog.0.0.3: Bad checksum)](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/6ecbb9946ff9a02568814ed405476a3353d02a4d/variant/compilers,4.14,ocolor.1.3.1,revdeps,simlog.0.0.3)
I checked: CI is right. I do not own/maintain/... simlog.
Not sure what should be done from that.

mseri:
Nothing from your side. Thanks for the submission

PR #24326, opened 2023-08-28
cohttp-mirage: add missing avoid-version

samoht:


kit-ty-kate:
This is missing `available: opam-version >= "2.1.0"`

PR #24332, opened 2023-08-30
Constrain Links.0.9.7

dhil:
The installation of Links.0.9.7 with OCaml 5.0 silently fails (c.f. links-lang/links#1176). This patch puts an upper bound on the required version of the OCaml compiler such that OPAM will no longer attempt to install Links.0.9.7 with OCaml 5.0.

We have fixed the compatibility issues with OCaml 5.0 in our development repository. We will make a separate release to have an OCaml 5 compatiable package of Links available on OPAM.

### DIFF
--- a/packages/links/links.0.9.7/opam
+++ b/packages/links/links.0.9.7/opam
@@ -15,7 +15,7 @@ build: [
 ]
 
 depends: [
-  "ocaml" {>= "4.08.0" & < "5.0"}
+  "ocaml" {>= "4.08.0" & < "5.0~"}
   "dune" {>= "2.7"}
   "ppx_deriving"
   "ppx_deriving_yojson" {>= "3.3"}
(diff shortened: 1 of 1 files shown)


packages/links/links.0.9.7/opam
```diff
@@ -15,7 +15,7 @@ build: [
 ]
 
 depends: [
-  "ocaml" {>= "4.08.0"}
+  "ocaml" {>= "4.08.0" & < "5.0"}
```
avsm, line 18:
```suggestion
  "ocaml" {>= "4.08.0" & < "5.0~"}
```

Just a minor tweak - the tilda will be the lowest version of 5.0, which will catch any beta releases etc.

dhil: Thanks!

PR #24339, opened 2023-08-30
Update camlp5-buildscripts for freebsd 2

chetmurthy:
trying again to update camlp5-buildscripts for freebsd


chetmurthy:
It looks like this hit some snags unrelated to the packages being updates:

1. revdeps craziness
2. some system packages didn't install (under "extras")
3. and one test failed to build dune (a prereq for something)

But otherwise, this looks like it passed fine.

PR #24341, opened 2023-08-31
Synchronise compiler opam file with ocaml/ocaml

dra27:
Reduces the diff with ocaml/ocaml, having spotted some missing things in ocaml/ocaml's copy:
- Cosmetic changes just to reduce the noise of the diff (there are some fields which necessarily differ, so it's useful to have ones which don't in the same place and the same format)
- The parallelism message if very old and the compiler's build is very regularly tested at very high `-j` - I think we can risk dropping that message!
- Irrelevant 32-bit options removed

(cf. ocaml/ocaml#12524; cc @Octachron)


avsm:
Thanks!

PR #24330, opened 2023-08-29
Package llvm.15.0.7+nnp-2

kit-ty-kate:
### `llvm.15.0.7+nnp-2`
The OCaml bindings distributed with LLVM
Note: LLVM should be installed first.

---
* Homepage: http://llvm.moe
* Source repo: git+https://github.com/kit-ty-kate/llvm-dune.git
* Bug tracker: http://llvm.org/bugs/

---
:camel: Pull-request generated by opam-publish v2.2.0

PR #24344, opened 2023-09-01
Add Frama-Clang package, take 2

vprevosto:
This PR supersedes #22222 , especially now that `conf-libclang` includes `libclang-cpp` packages in depexts for distributions that separates C and C++ bindings (i.e. #22442). Beyond frama-clang itself, I've also changed a few things in conf-llvm and conf-libclang:

- updated `conf-libclang.15`: the `arch` depext must now install `llvm15` and `clang15`, as the unsuffixed packages now install v16.
- updated `conf-llvm.14.0.6`: the `arch` depext must now install `llvm14`
- added `conf-llvm.15.0.7` (this is the patchlevel on Arch for package `llvm15`).

Of course, this last point can be removed if this conflicts with #24330

I understand that support for llvm 16 is the subject of PR #23630, hence I haven't tried to touch that.

### DIFF
--- a/packages/conf-llvm/conf-llvm.15/opam
+++ b/packages/conf-llvm/conf-llvm.15/opam
@@ -15,7 +15,7 @@ depexts: [
   ["llvm-15"] {os-distribution = "macports" & os = "macos"}
   ["llvm-15-dev" "zlib1g-dev" "libzstd-dev"] {os-family = "debian"}
   ["llvm15-dev"] {os-distribution = "alpine"}
-  ["llvm"] {os-family = "arch"}
+  ["llvm15"] {os-family = "arch"}
   ["llvm15-devel"] {os-family = "suse"}
   ["llvm15-devel"] {os-distribution = "fedora"}
   ["llvm15-devel" "epel-release"] {os-distribution = "centos"}
--- a/packages/conf-libclang/conf-libclang.15/opam
+++ b/packages/conf-libclang/conf-libclang.15/opam
@@ -15,7 +15,7 @@ depends: [
 ]
 depexts: [
   ["llvm@15"] {os = "macos"}
-  ["llvm" "clang"] {os-distribution = "arch"}
+  ["llvm15" "clang15"] {os-distribution = "arch"}
   ["libclang-15-dev" "libclang-cpp15-dev" "llvm-15-dev"]
     {(os-distribution = "ubuntu" & os-version >= "22.10")}
   ["libclang-14-dev" "libclang-cpp14-dev" "llvm-14-dev"]
--- a/packages/conf-llvm/conf-llvm.14.0.6/opam
+++ b/packages/conf-llvm/conf-llvm.14.0.6/opam
@@ -15,7 +15,7 @@ depexts: [
   ["llvm-14"] {os-distribution = "macports" & os = "macos"}
   ["llvm-14-dev"] {os-family = "debian"}
   ["llvm14-dev"] {os-distribution = "alpine"}
-  ["llvm"] {os-family = "arch"}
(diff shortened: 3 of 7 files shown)


kit-ty-kate (changes requested):
#24330 was merged so the conf-llvm.15.0.7 package should be removed from this PR

kit-ty-kate:
On macOS and Debian < 12 your package fails with:
```
#=== ERROR while compiling frama-clang.0.0.14 =================================#
# context              2.2.0~alpha3~dev | linux/x86_64 | ocaml-base-compiler.5.0.0 | pinned(https://git.frama-c.com/pub/frama-clang/-/archive/0.0.14/frama-clang-0.0.14.tar.bz2)
# path                 ~/.opam/5.0/.opam-switch/build/frama-clang.0.0.14
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p frama-clang -j 255 --promote-install-files=false @install
# exit-code            1
# env-file             ~/.opam/log/frama-clang-7-df4f98.env
# output-file          ~/.opam/log/frama-clang-7-df4f98.out
### output ###
# Error: execve(./build.sh): No such file or directory
# -> required by _build/default/framaCIRGen
# -> required by _build/install/default/bin/framaCIRGen
# -> required by _build/default/frama-clang.install
```

vprevosto:
Thanks for the report. I'm not sure why someone put `#!/usr/bin/bash` as shebang for this two-lines script, but the fix should be quite easy...

vprevosto:
@kit-ty-kate https://github.com/vprevosto/opam-repository/tree/963be4b3b7671b97e61a59bdf4e488d9d615c5b6 modifies `conf-llvm.15` to ask for `llvm15` on arch: as of now arch's `llvm`'s package will give you v16, not v15.

kit-ty-kate:
Thanks. Let's ignore all those errors for now

kit-ty-kate (approved):

packages/conf-llvm/conf-llvm.15/opam
```diff
@@ -15,7 +15,7 @@ depexts: [
   ["llvm-15"] {os-distribution = "macports" & os = "macos"}
   ["llvm-15-dev" "zlib1g-dev" "libzstd-dev"] {os-family = "debian"}
   ["llvm15-dev"] {os-distribution = "alpine"}
-  ["llvm"] {os-family = "arch"}
+  ["llvm15"] {os-family = "arch"}
```
kit-ty-kate, line 18: These changes broke the conf-llvm package on Archlinux. Would you mind reverting those? (there is no such package, the current version of the `llvm` package is 15, there is no alias called `llvm15`)

vprevosto: That's strange. I'm running Arch, and this is what I get:
```
❯ pacman -Q llvm llvm15
llvm 16.0.6-1
llvm15 15.0.7-1
```
As far as I can tell, these packages have been made available in the second half of August. As a matter of fact, I made the change because on my machine conf-llvm would complain that the LLVM version didn't match. I don't mind reverting of course, but I'd be interested to know why we end up with a different view of Arch's packages.

kit-ty-kate: Ah, interesting... I use ArchlinuxARM and it seems their repository is pretty outdated:
```
$ pacman -Q llvm llvm15
llvm 15.0.7-3
error: package 'llvm15' was not found
```
Let's leave things as-is then. Sorry for the ping

vprevosto: NB: I imagine that `conf-libclang` should be modified as well? (according to `pacman`, `clang15` was release at the same time as `llvm15`)

PR #24349, opened 2023-09-04
uunf and uuseg: adjust next Unicode release bounds.

dbuenzli:
The next Unicode release is going to be a point release. 

`uucd` and `uucp` follow the Unicode release version numbers so this needs to be adjusted as mentioned.


mseri:
Thanks

Issue #24308, opened 2023-08-24
gsl fails to install since recently


glondu:
Since recently, gsl, hence orpie, fails to install in a 4.13.1 switch:
```
[ERROR] The compilation of gsl.1.24.3 failed at "dune build -p gsl -j 1".

#=== ERROR while compiling gsl.1.24.3 =========================================#
# context     2.1.2 | linux/x86_64 | ocaml-base-compiler.4.13.1 | https://opam.ocaml.org#1da29ec1
# path        ~/.opam/4.13.1/.opam-switch/build/gsl.1.24.3
# command     ~/.opam/4.13.1/bin/dune build -p gsl -j 1
# exit-code   1
# env-file    ~/.opam/log/gsl-2228388-3a2392.env
# output-file ~/.opam/log/gsl-2228388-3a2392.out
### output ###
# File "src/dune", line 78, characters 0-101:
# 78 | (rule
# 79 |   (targets sf.mli sf.ml)
# 80 |   (deps config/do_sf.exe sf.mli.q)
# 81 |   (action (run config/do_sf.exe))
# 82 | )
# (cd _build/default/src && config/do_sf.exe)
# Fatal error: exception Sys_error("gsl_include.sexp: No such file or directory")
```

As far as I can tell, it wasn't the case on 2023-08-21. I haven't seen any other suspicious upgrade since then... However, biniou did downgrade from 1.2.2 to 1.2.1 in my switch.

Context: https://salsa.debian.org/ocaml-team/opam-debian-switch

kit-ty-kate:
Mmh, I can't reproduce it locally and it was supposed to be fixed by https://github.com/ocaml/opam-repository/pull/19244
Is this deterministically reproducible with your setup?

glondu:
It is strange. It seems deterministically reproducible when I (try to) install all [these packages](https://salsa.debian.org/ocaml-team/opam-debian-switch/-/blob/master/opams-to-install.txt) (except ppx_globalize) at once in an empty 4.13.1 switch. However, installing orpie alone in an empty switch, or installing it after all the others are installed, succeds.

kit-ty-kate:
I've tried to reproduce locally with all those packages at once but it worked for me

glondu:
[...]

PR #24354, opened 2023-09-05
Fix dependency issue in gsl (#24308)

glondu:


kit-ty-kate:
Thanks. Could you also apply same patch to all the previous affected versions? I believe all the dune/jbuilder versions are affected.

packages/gsl/gsl.1.24.3/opam
```diff
@@ -37,3 +37,4 @@ url {
     "sha512=c6e2578618591d1bef428693b69026cdea0f1606cd25d9f02d637a90256a5685eee70ecd0259d2595a1cd7b292c34e3c913c007e2706aa125af045d37e55d9c4"
   ]
 }
+patches: ["fixdeps.patch"]
```
kit-ty-kate, line 40: The extra-files field is missing
```suggestion
patches: ["fixdeps.patch"]
extra-files: ["fixdeps.patch" "md5=???"]
```

glondu:
> Thanks. Could you also apply same patch to all the previous affected versions? I believe all the versions with dune/jbuilder are affected.

...which means testing/adapting the patch to all those versions? It is not part of my QA routine... so I won't bother/care.

kit-ty-kate:
Ok no worries. Thanks!

PR #24355, opened 2023-09-05
[new release] lintcstubs-arity (0.2.0)

edwintorok:
Generate headers for C bindings

- Project page: <a href="https://github.com/edwintorok/lintcstubs-arity">https://github.com/edwintorok/lintcstubs-arity</a>

##### CHANGES:

* Add `lintcstubs_arity_cmt` tool that supports generating prototypes for unboxed arguments in primitives (requires OCaml 4.10+)
* Add a README.md that shows how to use the tool and full examples with both `dune` and `Makefile`
* Add a minimal example that shows how to embed/build the tool as part of another project
* Enable tests in the CI

### DIFF
--- a/packages/lintcstubs-arity/lintcstubs-arity.0.2.0/opam
+++ b/packages/lintcstubs-arity/lintcstubs-arity.0.2.0/opam
@@ -8,7 +8,7 @@ license: "LGPL-2.1-or-later"
 homepage: "https://github.com/edwintorok/lintcstubs-arity"
 bug-reports: "https://github.com/edwintorok/lintcstubs-arity/issues"
 depends: [
-  "dune" {>= "2.7"}
+  "dune" {>= "3.0"}
   "ocaml" {>= "4.08"}
   "odoc" {with-doc}
 ]
--- a/packages/lintcstubs-arity/lintcstubs-arity.0.2.0/opam
+++ b/packages/lintcstubs-arity/lintcstubs-arity.0.2.0/opam
@@ -1,0 +1,41 @@
+opam-version: "2.0"
+synopsis: "Generate headers for C bindings"
+description:
+  "Generates .h files from 'external' declarations in .ml or .cmt files. Can be used to spot mismatches in number of arguments between C primitive declared in OCaml and its implementation in the .c file."
+maintainer: ["Edwin Török <edwin.torok@cloud.com>"]
+authors: ["Edwin Török <edwin.torok@cloud.com>"]
+license: "LGPL-2.1-or-later"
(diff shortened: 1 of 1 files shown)


edwintorok:
This is the 2nd in a series of tools to be presented at the OCaml workshop later this week (more to come in separate PRs).

packages/lintcstubs-arity/lintcstubs-arity.0.2.0/opam
```diff
@@ -0,0 +1,41 @@
...
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
```
kit-ty-kate, line 28: busybox's xargs does not support `--verbose`
```suggestion
    "@runtest" {with-test & os-family != "alpine"}
```

packages/lintcstubs-arity/lintcstubs-arity.0.2.0/opam
```diff
@@ -0,0 +1,41 @@
...
+synopsis: "Generate headers for C bindings"
+description:
+  "Generates .h files from 'external' declarations in .ml or .cmt files. Can be used to spot mismatches in number of arguments between C primitive declared in OCaml and its implementation in the .c file."
+maintainer: ["Edwin Török <edwin.torok@cloud.com>"]
+authors: ["Edwin Török <edwin.torok@cloud.com>"]
+license: "LGPL-2.1-or-later"
+homepage: "https://github.com/edwintorok/lintcstubs-arity"
+bug-reports: "https://github.com/edwintorok/lintcstubs-arity/issues"
+depends: [
+  "dune" {>= "2.7"}
```
kit-ty-kate, line 11: This is probably a little low as it fails with 2.7.0 over an `(enabled_if )` construct

packages/lintcstubs-arity/lintcstubs-arity.0.2.0/opam
edwintorok: Thanks, I'll drop that flag in the next version it is only meant as a debugging aid.

packages/lintcstubs-arity/lintcstubs-arity.0.2.0/opam
edwintorok: Looks like this doesn't work on any 2.x, bumped it to 3.0 for now. I'll come back to this later to see whether I can make it work on 2.x (maybe by separating the package based on the OCaml version it supports).

kit-ty-kate:
Thanks

PR #24356, opened 2023-09-05
10 packages from savonet/liquidsoap-release-assets

toots:
This pull-request concerns:
- `liquidsoap.2.2.1`: Swiss-army knife for multimedia streaming
- `liquidsoap-core.2.2.1`: Liquidsoap core library and binary
- `liquidsoap-js.2.2.1`: Liquidsoap language - javascript wrapper
- `liquidsoap-lang.2.2.1`: Liquidsoap language library
- `liquidsoap-libs.2.2.1~1`: Liquidosap standard library
- `liquidsoap-libs-extra.2.2.1~1`: Liquidosap standard library -- extra functionalities
- `liquidsoap-mode.2.2.1`: Liquidosap emacs mode
- `prometheus-liquidsoap.2`: Virtual package installing liquidsoap dependencies for prometheus optional features
- `sdl-liquidsoap.3`: Virtual package installing liquidsoap dependencies for SDL optional features
- `tls-liquidsoap.1`: Virtual package install liquidosap dependencies for TLS optional features

---
* Homepage: https://github.com/savonet/liquidsoap
* Source repo: git+https://github.com/savonet/liquidsoap.git
* Bug tracker: https://github.com/savonet/liquidsoap/issues

---


kit-ty-kate:
> liquidsoap-libs.2.2.1~1

Is the `~1` expected? I'm curious, why do you need this version to be before 2.2.1? Is this a release candidate or something similar?

toots:
> > liquidsoap-libs.2.2.1~1
> 
> Is the `~1` expected? I'm curious, why do you need this version to be before 2.2.1? Is this a release candidate or something similar?

We want to reserve the option to do point-release with just the `libs` so we have these dependencies in `liquidsoap`:
```opam
  "liquidsoap-libs" {>= "2.2.1~1" & < "2.2.1"}
  "liquidsoap-libs-extra" {>= "2.2.1~1" & < "2.2.1"}
```

and new release would be `2.2.1~2`, `2.2.1~3`, etc.

kit-ty-kate:
> and new release would be 2.2.1~2, 2.2.1~3, etc.

in that case why not have `-1`, `-2`, `-3` instead of `~1`, `~2`, `~3` ?

kit-ty-kate:
or more typical: have `liquidsoap-libs.2.2.1` and if another revision is needed that would be `liquidsoap-libs.2.2.1-1`, `liquidsoap-libs.2.2.1-2`, etc

toots:
> or more typical: have `liquidsoap-libs.2.2.1` and if another revision is needed that would be `liquidsoap-libs.2.2.1-1`, `liquidsoap-libs.2.2.1-2`, etc

I think the original thinking was that the `liquidsoap` package itself might need point releases.. `2.2.1-1` might need libraries that are different than `2.2.1` (or incompatible) ...

toots:
This is all very theoretical though..

toots:
> > I think the original thinking was that the `liquidsoap` package itself might need point releases.. `2.2.1-1` might need libraries that are different than `2.2.1` (or incompatible) ...
> 
> I'm not sure to understand. Why would `~1` solve this issue but not `-1` ?

It might not indeed. Do you think versioning `libs` with an additional prefix like `2.2.1.1` would be better? Any suggestion on how to better achieve this?

mseri:
I don't see the problem with 4 digits, we have multiple packages done that way.
If you release 2.2.1 and 2.2.2, afaik, 2.2.1.1 will be ordered in between the two

toots:
Thanks for the feedback y'all. For some reason, `opam publish` created a new PR https://github.com/ocaml/opam-repository/pull/24376 so I'm closing this one.

PR #24360, opened 2023-09-06
Add oracle linux support for libflac

tmcgilchrist:


packages/conf-libflac/conf-libflac.1/opam
```diff
@@ -12,6 +12,7 @@ depexts: [
   ["libflac-dev"] {os-family = "debian"}
   ["flac-devel"] {os-distribution = "centos"}
   ["flac-devel"] {os-distribution = "fedora"}
+  ["flac-libs"] {os-distribution = "ol" & os-version >= "8"}
```
mseri, line 15: This does not seem to provide the devel components for the library

mseri:
This still fails because–if I am understanding it correctly–the devel packages is provided in an optional package repository in oraclelinux :-/

tmcgilchrist:
That’s unfortunate, maybe we should just constraint away oracle Linux for
the time being.

I’ll have a look at what we put into the oracle linux base images on
opam-repo-ci on Monday. If it’s an optional package repo we can’t assume it
is there. Thanks for persisting with it.

mseri:
For the time being I'll close this PR. We will re-add the dependency once it is installable on oraclelinux

PR #24364, opened 2023-09-06
Package conf-readline.1

vtourneur:
### `conf-readline.1`
Virtual package relying on a readline system installation
This package can only install if the readline lib is installed on the system.

---
* Homepage: https://gitlab.inria.fr/vtourneu/readline-ocaml
* Bug tracker: https://gitlab.inria.fr/vtourneu/readline-ocaml/issues

---


packages/conf-readline/conf-readline.1.0/opam
```diff
@@ -0,0 +1,23 @@
...
+maintainer: ["vincent.tourneur@inria.fr"]
+authors: ["sylvain.pogodalla@inria.fr" "vincent.tourneur@inria.fr"]
+depends: [
+  "conf-pkg-config" {build}]
+depexts: [
+  ["libreadline8" "libreadline-dev"] {os-family = "debian"}
+  ["libreadline8" "libreadline-dev"] {os-family = "ubuntu"}
+  ["readline" "readline-devel"] {os-distribution = "centos"}
+  ["readline" "readline-devel"] {os-distribution = "fedora"}
+  ["readline" "readline-dev"] {os-family = "alpine"}
```
kit-ty-kate, line 11: I believe the dev packages already require the library itself (that would be a bug otherwise)
```suggestion
  ["libreadline-dev"] {os-family = "debian"}
  ["libreadline-dev"] {os-family = "ubuntu"}
  ["readline-devel"] {os-distribution = "centos"}
  ["readline-devel"] {os-distribution = "fedora"}
  ["readline-dev"] {os-family = "alpine"}
```

kit-ty-kate (changes requested):
Could you also make this package `conf-readline.1`? The `.0` doesn't make much sense for conf packages

vtourneur:
Ok, I renamed the pull request and the directory in the repo, is there other places where I should rename?

kit-ty-kate (approved):

mseri:
Works on macports, I am not sure why it does not work on homebrew.

mseri:
It fails on the distributions not providing the pkg-config file. 
We could fix it by adding ad-hoc checks but I am not sure it is worth.
In my opinion it can be merged, @kit-ty-kate what do you think?

packages/conf-readline/conf-readline.1/opam
```diff
@@ -0,0 +1,24 @@
+opam-version: "2.0"
+maintainer: ["vincent.tourneur@inria.fr"]
+authors: ["sylvain.pogodalla@inria.fr" "vincent.tourneur@inria.fr"]
+build: ["pkg-config" "--short-errors" "--print-errors" "readline"]
```
avsm, line 4:
```suggestion
```

packages/conf-readline/conf-readline.1/opam
```diff
@@ -0,0 +1,24 @@
...
+depends: [
+  "conf-pkg-config" {build}]
```
avsm, line 6:
```suggestion
```

avsm (commented):
I think the right thing to do here is to skip the pkg-config probe, since the depext specification will install the right packages. Most of the distros do not wrap readline in pkg-config, so the test is unnecessary.

packages/conf-readline/conf-readline.1/opam
```diff
@@ -0,0 +1,21 @@
+opam-version: "2.0"
+maintainer: ["vincent.tourneur@inria.fr"]
+authors: ["sylvain.pogodalla@inria.fr" "vincent.tourneur@inria.fr"]
+depexts: [
```
mseri, line 4:
```suggestion
license: "GPL-3.0-or-later"
depexts: [
```

mseri:
Thanks

PR #24367, opened 2023-09-06
reason-react: add upper bound to `reason` dep

jchavarri:
The next version of the `reason` package will include a change that might break users of `reason-react`: https://github.com/reasonml/reason/pull/2721. 

This PR adds an upper bound in advance, to prevent any future issues.

cc @anmonteiro @davesnx.

### DIFF
--- a/packages/melange/melange.1.0.0/opam
+++ b/packages/melange/melange.1.0.0/opam
@@ -13,7 +13,7 @@ depends: [
   "dune-build-info"
   "cppo" {build}
   "ounit" {with-test}
-  "reason" {with-test}
+  "reason" {with-test & < "3.10.0"}
   "ppxlib" {>= "0.24.0"}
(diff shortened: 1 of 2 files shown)


anmonteiro:
can you add the same upper bound to Melange 1.0? https://github.com/ocaml/opam-repository/blob/57e2bdd752a666175e519d6c17b96bd3f9158821/packages/melange/melange.1.0.0/opam#L16

jchavarri:
@anmonteiro I can, but I don't think it will change anything. That dep already has `with-test` and the `build` field of that version doesn't include the `@runtest` alias, so the code that depends on `reason` should never run.

https://github.com/ocaml/opam-repository/blob/d46904ea12e3b5bf29aaac4fd7973c0d64315260/packages/melange/melange.1.0.0/opam#L26-L37

mseri:
Thanks

PR #24327, opened 2023-08-29
batteries-3.7.0

UnixJunkie:
I will issue the final tag once opam CI tests are OK

PR #24370, opened 2023-09-07
Batteries 3.7 upper bounds

mseri:
It seems that in all cases the failure is related to the use of `obuild`.
Adding a conflict with obuild or batteries in either package seems overkill, instead I have added upper bounds on the packages that failed to build.

Failures were seen in https://github.com/ocaml/opam-repository/pull/24327


AndrewScheller2006:
❤

PR #24385, opened 2023-09-09
git.robur.io is now git.robur.coop

hannesm:
github.com/roburio is now github.com/robur-coop

I will open another PR with same changes to other packages.


mseri:
Thanks

PR #24409, opened 2023-09-12
[new release] beluga (1.1.1)

MartyO256:

- Project page: <a href="http://complogic.cs.mcgill.ca/beluga/">http://complogic.cs.mcgill.ca/beluga/</a>

### Added

- Added support for fixity pragmas before declarations.

### Changed

- Updated the Beluga mode for Emacs to use the `'cl-lib` library instead of `'cl`.

### Fixed

- Colouring of error messages is fixed for the Beluga mode for Emacs using the `'ansi-color` library (requires Emacs >= v28.1).


mseri:
Thanks

PR #24416, opened 2023-09-13
disable Conf diffutils for freebsd 2

chetmurthy:
can't figure out how to make it work, so disable it (so packages that require it will refuse to installl, instead of failing)

### DIFF
--- a/packages/conf-diffutils/conf-diffutils.1.2/opam
+++ b/packages/conf-diffutils/conf-diffutils.1.2/opam
@@ -1,0 +1,24 @@
+opam-version: "2.0"
+homepage: "https://www.gnu.org/software/diffutils/"
+bug-reports: "chetsky@gmail.com"
+maintainer: "chetsky@gmail.com"
+authors: "GNU Project"
+license: "GPL-3.0-or-later"
+build: [["diff" "--help"]]
+depexts: [
+  ["diffutils"] {os-family = "debian"}
+  ["diffutils"] {os-distribution = "fedora"}
+  ["diffutils"] {os-distribution = "rhel"}
+  ["diffutils"] {os-distribution = "centos"}
+  ["diffutils"] {os-distribution = "alpine"}
+  ["diffutils"] {os-distribution = "nixos"}
+  ["diffutils"] {os-family = "suse" | os-family = "opensuse"}
+  ["diffutils"] {os-distribution = "ol"}
+  ["diffutils"] {os-distribution = "arch"}
+  ["diffutils"] {os = "macos" & os-distribution = "homebrew"}
+  ["diffutils"] {os = "freebsd"}
+]
+synopsis: "Virtual package relying on diffutils"
+description:
+  "This package can only install if the diffutils binary is installed on the system."
+flags: conf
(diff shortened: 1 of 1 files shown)


jmid:
I think the first commit is on the right path!

The CI log shows that the package is being installed alright - until a final check `build: [["diff" "--help"]]` from the opam file fails:
https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/352eb2525d7f096868c025cd5bdb4113f77e3c23/variant/freebsd,freebsd-ocaml-4.14-amd64,conf-diffutils.1.2

I suspect that the reason is that the installed GNU diff binary is named `gdiff`: https://www.freshports.org/textproc/diffutils/

> Note that this port will install these utilities with a 'g' prefix, for example gdiff, gdiff3, gcmp, [...]

and that it is then a matter of putting the installed binaries in the `PATH` under the right name (or aliasing?) to shadow the BSD ones, so that the right one (supporting `--help`) will be picked up.

chetmurthy:
@jmid I think that did it.  I'll have to write something to pick up the right "diff"/"gdiff" command, but that should be straightforward enough.

jmid:
Great! :smiley: 

My initial thought was to alter the `PATH` or add an alias in the opam package so that we could just uniformly use `diff` to invoke the GNU version after installing `conf-diffutils`. This should avoid dependent packages from having to add special FreeBSD cases. However I can also see a point in not shadowing FreeBSD's `diff` to avoid later surprises.

Anyway, for now I can see the `mutaml` test suite works fine on FreeBSD even with the non-GNU `diff` in `PATH`, so I'm good... :wink:

chetmurthy:
I admit that it's a little unsettling to have to conditionally select which diff to use, and I'd rather fix that in conf-diffutils, but I don't see a clean way to do that at the moment.  And since the code to conditionally choose the name is only a few lines in a Makefile, I'm going with that for the moment.

jmid:
Good point about not relying on GNU extensions (if possible)! 

> If there's need for a GNU diff, I suggest to provide a package `conf-GNU-diff` [...]

In fairness, isn't `diffutils` the name associated to GNU `diff` and friends? https://www.gnu.org/software/diffutils/
(serious not-intended-as-flame-bait question from this BSD-illiterate Linux user)

For `mutaml` I recall struggling with an ancient GNU `diff` under macOS and Alpine's reduced busybox `diff` in the past.
For that reason, I'm a bit biased to let this one keep its meaning as GNU diffutils... :wink:

raphael-proust:
If you don't use the `diff` GNU extensions, then maybe your `diffutils` dependency should  be with `{ os != freebsd }`? Would that work in mutaml?

It's a bit strange to depend on a package differently depending on which system you are on… but it's a system's package so it makes some level of sense?

chetmurthy:
I've renumbered.

As to why this package was needed: originally there was a platform (I forge which) where diff wasn't even present, so it needed to be installed.  freebsd is different (there's a diff, but it's a little different from GNU diff), but since we have to have this package anyway (for that diff-less platform) it seems like I might as well conditionalize the selection of the diff executable.  It turned out to be only a few lines of script anyway.

chetmurthy:
Hi, is there anything else I need to do for this?  I applied the change you suggested, and this passes all tests AFAICT.

kit-ty-kate:
Thanks! We'll have to look into the packages that use this to possibly only depend on conf-diffutils when not on FreeBSD because i feel like most of them do not take into account gdiff

PR #24418, opened 2023-09-13
s/team AT robur dot io/team AT robur dot coop/

reynir:
We moved away from robur.io to robur.coop - I think these are the last occurences of `robur{., dot }io`.


kit-ty-kate:
Thanks!

PR #21975, opened 2022-08-12
[new release] goblint-cil (2.0.0)

sim642:
A front-end for the C programming language that facilitates program analysis and transformation

- Project page: <a href="https://github.com/goblint/cil">https://github.com/goblint/cil</a>

##### CHANGES:

* Wrap library into `GoblintCil` module (goblint/cil#107).
* Remove all MSVC support (goblint/cil#52, goblint/cil#88).
* Port entire build process from configure/make to dune (goblint/cil#104).
* Add C11 `_Generic` support (goblint/cil#48).
* Add C11 `_Noreturn` support (goblint/cil#58).
* Add C11 `_Static_assert` support (goblint/cil#62).
* Add C11 `_Alignof` support (goblint/cil#66).
* Add C11 `_Alignas` support (goblint/cil#93, goblint/cil#108).
* Add partial C11 `_Atomic` support (goblint/cil#61).
* Add `_Float32`, `_Float64`, `_Float32x` and `_Float64x` type support (goblint/cil#8, goblint/cil#60).
* Add Universal Character Names, `char16_t` and `char32_t` type support (goblint/cil#80).
* Change locations to location spans and add additional expression locations (goblint/cil#51).
* Add synthetic marking for CIL-inserted statement locations (goblint/cil#98).
* Expose list of files from line control directives (goblint/cil#73).
* Add parsed location transformation hook (goblint/cil#89).
* Use Zarith for integer constants (goblint/cil#47, goblint/cil#53).
* Fix constant folding overflows (goblint/cil#59).
* Add option to disable constant branch removal (goblint/cil#103).
* Add standalone expression parsing and checking (goblint/cil#97, goblint/cil#96).
* Improve inline function merging (goblint/cil#72, goblint/cil#85, goblint/cil#84, goblint/cil#86).
* Fix some attribute parsing cases (goblint/cil#71, goblint/cil#75, goblint/cil#76, goblint/cil#77).
* Fix global NaN initializers (goblint/cil#78, goblint/cil#79).
* Fix `cilly` binary installation (goblint/cil#99, goblint/cil#100, goblint/cil#102).
* Remove batteries dependency to support OCaml 5 (goblint/cil#106).

PR #24423, opened 2023-09-13
[new release] goblint (2.2.1)

sim642:
Static analysis framework for C

- Project page: <a href="https://goblint.in.tum.de">https://goblint.in.tum.de</a>
- Documentation: <a href="https://goblint.readthedocs.io/en/latest/">https://goblint.readthedocs.io/en/latest/</a>

##### CHANGES:

* Bump batteries lower bound to 3.5.0.
* Fix flaky dead code elimination transformation test.

### DIFF
--- a/packages/lintcstubs/lintcstubs.0.4.7/opam
+++ b/packages/lintcstubs/lintcstubs.0.4.7/opam
@@ -11,7 +11,7 @@ depends: [
   "dune" {>= "3.0"}
   "ocaml" {>= "4.13"}
   "lintcstubs-gen" {= version}
-  "goblint" {>= "2.1.0"}
+  "goblint" {>= "2.1.0" & < "2.2.0"}
   "goblint-cil"
   "dune-compiledb"
   "fpath"
--- a/packages/goblint/goblint.2.2.1/opam
+++ b/packages/goblint/goblint.2.2.1/opam
@@ -1,0 +1,93 @@
+opam-version: "2.0"
+synopsis: "Static analysis framework for C"
+maintainer: [
+  "Simmo Saan <simmo.saan@gmail.com>"
+  "Michael Schwarz <michael.schwarz93@gmail.com>"
+]
+authors: [
+  "Simmo Saan"
+  "Michael Schwarz"
+  "Julian Erhard"
+  "Sarah Tilscher"
+  "Ralf Vogler"
+  "Kalmer Apinis"
+  "Vesal Vojdani"
+]
+license: "MIT"
+homepage: "https://goblint.in.tum.de"
+doc: "https://goblint.readthedocs.io/en/latest/"
+bug-reports: "https://github.com/goblint/analyzer/issues"
+depends: [
+  "dune" {>= "3.6"}
+  "ocaml" {>= "4.10"}
+  "goblint-cil" {>= "2.0.2"}
+  "batteries" {>= "3.5.0"}
+  "zarith" {>= "1.8"}
+  "yojson" {>= "2.0.0"}
+  "qcheck-core" {>= "0.19"}
+  "ppx_deriving"
+  "ppx_deriving_hash"
+  "ppx_deriving_yojson" {>= "3.7.0"}
+  "ounit2" {with-test}
+  "qcheck-ounit" {with-test}
+  "odoc" {with-doc}
+  "fpath"
+  "dune-site"
+  "dune-build-info"
+  "json-data-encoding"
+  "jsonrpc" {>= "1.12"}
+  "sha" {>= "1.12"}
+  "fileutils" {>= "0.6.4"}
+  "cpu"
+  "arg-complete"
+  "yaml" {>= "3.0.0"}
+  "uuidm"
+  "catapult"
+  "catapult-file"
+  "conf-gmp" {>= "3"}
+  "conf-ruby" {with-test}
+  "benchmark" {with-test}
+  "conf-gcc"
+]
+depopts: ["apron" "z3"]
+conflicts: [
+  "result" {< "1.5"}
+]
+build: [
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "--promote-install-files=false"
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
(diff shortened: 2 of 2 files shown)


sim642:
All the CI failures aren't really related nor fixable here.

tmcgilchrist:
The macOS failure is curious, batteries not finding a real gcc? Perhaps it is related to the FreeBSD failure which is caused by conf-gcc not having an entry for FreeBSD. 
Add FreeBSD for conf-gcc https://github.com/ocaml/opam-repository/pull/24429

sim642:
The "couldn't find real gcc" error is from goblint-cil, not goblint. I'm surprised to see this again because in https://github.com/ocaml/opam-repository/pull/21975#issuecomment-1213132109 I said:

> With conf-gcc it's depext forces homebrew to install a real gcc.

Depext also shows gcc being installed via homebrew, so I'm not sure what the issue is.

packages/lintcstubs/lintcstubs.0.4.7/opam
```diff
@@ -11,7 +11,7 @@ depends: [
   "dune" {>= "3.0"}
   "ocaml" {>= "4.13"}
   "lintcstubs-gen" {= version}
-  "goblint" {>= "2.1.0"}
+  "goblint" {>= "2.1.0" & < "2.2.0"}
```
sim642, line 14: @edwintorok Goblint's library interface hasn't been designed for external use and regularly breaks between releases, so you'll probably want to adapt lintcstubs. Previous Goblint releases are incompatible with OCaml 5.0, which also means lintcstubs is, which is unfortunate (OCaml 5 C stubs have to be analyzed under OCaml 4.14 instead).

edwintorok: No problem, I'll add some tighter bounds in my repo, thanks for fixing it up. In general I intend to track the latest release of goblint, I think master of goblint already works with OCaml 5.

edwintorok: https://github.com/edwintorok/lintcstubs/tree/goblint2.2.1-compat I've done an experiment to upgrade to latest goblint on this branch, there are some new warnings from goblint, but seems to work so far and the diff is fairly small: https://github.com/edwintorok/lintcstubs/commit/ce8150a60e7a013c33a93090b80596d4a1d0b769

raphael-proust:
On the macos CI failure I see the following in the logs.

First, it installs the `conf-gcc` package and that seems to happen alright.

```
- ==> Installing gcc
- ==> Pouring gcc--13.2.0.ventura.bottle.1.tar.gz
- No entry for terminal type "unknown";
- using dumb terminal settings.
- 🍺  /usr/local/Cellar/gcc/13.2.0: 1,590 files, 421.7MB
```

Then as part of installing `goblint-cil`, it tries to discover the local available the gcc by checking different versions. It starts with just plain `gcc`:

```
# (cd _build/default/bin && ./realGccConfigure.exe)
# run: gcc --version
# -> process exited with code 0
# -> stdout:
#  | Apple clang version 14.0.3 (clang-1403.0.22.14.1)
#  | Target: x86_64-apple-darwin22.5.0
#  | Thread model: posix
#  | InstalledDir: /Library/Developer/CommandLineTools/usr/bin
# -> stderr:
```

And then tries to go down from version to version:

```
# run: gcc-12 --version
# -> process exited with code 127
# -> stdout:
# -> stderr:
#  | sh: gcc-12: command not found
# run: gcc-11 --version
# -> process exited with code 127
# -> stdout:
# -> stderr:
#  | sh: gcc-11: command not found
# run: gcc-10 --version
# -> process exited with code 127
# -> stdout:
# -> stderr:
#  | sh: gcc-10: command not found
# run: gcc-9 --version
# -> process exited with code 127
# -> stdout:
# -> stderr:
#  | sh: gcc-9: command not found
# run: gcc-8 --version
# -> process exited with code 127
# -> stdout:
# -> stderr:
#  | sh: gcc-8: command not found
# run: gcc-7 --version
# -> process exited with code 127
# -> stdout:
# -> stderr:
#  | sh: gcc-7: command not found
# Fatal error: exception Failure("couldn't find real gcc")
```

As you can see, it misses the actual version that `goblint-cil` is actually installed at.

I'm a bit confused because AFAICT it should try versions 16 and lower. I think the issue is that `goblint-cil<2.02` should be marked as incompatible with `macos`? or `conf-gcc`? or something like that?

raphael-proust:
@sim642 sorry to be using your PR to investigate and fix some not-totally-related issues, just trying to take the opportunity to improve the general health of the opam-repository.

edwintorok:
Latest goblint-cil builds just  fine on my Mac OS (with gcc-13). Earlier 'goblint-cil' indeed doesn't build *on the latest version* of 'gcc' that homebrew installs. If you install a different version of 'gcc' it might work, but I don' think you can express that in 'conf-gcc'. To make the CI happier you could indeed mark older version of 'goblint-cil' as not supported, but please only do that on the CI (there are some x-ci-accept-failures tags), it may be possible for someone using an actual Mac to get it working.

sim642:
> I'm a bit confused because AFAICT it should try versions 16 and lower. I think the issue is that `goblint-cil<2.02` should be marked as incompatible with `macos`? or `conf-gcc`? or something like that?

Ah! That lintcstubs macos job is forced to go with goblint-cil 2.0.1, which I didn't notice before. goblint-cil 2.0.2 extends the range of gcc binaries it looks for such that it works.

Indeed, it works on macos if an older gcc from Homebrew is present. So goblint-cil 2.0.1 isn't outright incompatible with macos but depends on some depext versions on which constraints aren't possible.

raphael-proust:
Is it likely to work if we just add a `patch` file for `goblint-cil` to extend gcc version detection?

We could add this patch to the package. I think that's the solution which keeps the most compatibility. Does anyone know if `goblint-cil` might have incompatibilities with higher versions of gcc?

sim642:
It should be fine. This patch should do it: https://github.com/goblint/cil/commit/28d3e7e7b3373e03b4a2e35ef6f446bd66489743.

raphael-proust:
I made https://github.com/ocaml/opam-repository/pull/24493

raphael-proust:
I think this is ready to merge.

dinosaure:
Thanks!

PR #24425, opened 2023-09-13
Added FreeBSD depext for vim

mtelvers:


haochenx:
CI is failing on unrelated platforms. You can probably try to add the following to the end of the opam file:

```
x-ci-accept-failures: [
  "alpine-3.18"
  "archlinux"
  "opensuse-15.5"
  "opensuse-tumbleweed"
  "ubuntu-22.10"
]
```

mseri:
This is a strict improvement over what we had. I would let it fail for now, it will remind us to keep adding the necessary depexts

PR #24426, opened 2023-09-13
Added FreeBSD depext for wget

mtelvers:


haochenx:
same as https://github.com/ocaml/opam-repository/pull/24425#issuecomment-1718104677: you can try add an `x-ci-accept-failures` field.

mseri:
This is a strict improvement, plus we are dropping oraclelinux from the CI matrix, so there only remains to fix opensuse in a future PR

PR #24429, opened 2023-09-14
Add FreeBSD for conf-gcc

tmcgilchrist:


tmcgilchrist:
The failures on bitwuzla-bin should be resolved by constraining that package away from FreeBSD.

mseri:
Thanks

PR #24432, opened 2023-09-14
improve depexts for several conf packages

maroneze:
There are two commits:

- the first replaces an `os-distribution` with `os-family`, after testing on an Arch-based system (I don't remember which one exactly, sorry) which had the same package name;
- the second one adds several Cygwin package names for `conf-` packages used by Frama-C.

I'm using MobaXterm, with the following relevant config environment:

```
arch              x86_32                                                   # Inferred from system
opam-version      2.0.7                                                    # The currently running opam version
os                cygwin                                                   # Inferred from system
os-distribution   cygwin                                                   # Inferred from system
os-family         windows                                                  # Inferred from system
os-version        unknown                                                  # Inferred from system
```

So I'm not sure I should use `os` or `os-distribution` in the depexts I added; for now I used `os-distribution`, but if you prefer `os`, I can change it.

### DIFF
--- a/packages/conf-autoconf/conf-autoconf.0.1/opam
+++ b/packages/conf-autoconf/conf-autoconf.0.1/opam
@@ -23,6 +23,7 @@ depexts: [
   ["autoconf"] {os-distribution = "ol"}
   ["autoconf"] {os-distribution = "rhel"}
   ["system:autoconf"] {os = "win32" & os-distribution = "cygwinports"}
+  ["autoconf"] {os-distribution = "cygwin"}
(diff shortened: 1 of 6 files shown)


mseri:
Thanks

dra27:
Just for the Googling record, Cygwin-specific packages should have `os = "cygwin"` rather than `os-distribution = "cygwin"` (`os = "cygwin" & os-distribution = "cygwin"` ⇒ running opam in Cygwin and building Cygwin programs, `os = "win32" & os-distribution = "cygwin"` ⇒ running native Windows and using Cygwin to provide the cross-compiled native libraries.

PR #24442, opened 2023-09-14
mark bitwuzla-bin.0.0.0 as deprecated

recoules:
This package seems to have compilation errors with newer compilers (https://github.com/ocaml/opam-repository/pull/24429#issuecomment-1718724110).
As of today, the version `1.0.0` seems to still work.

Note: this package has little things to do in the opam repository as it contains only C/C++ sources.
I published it a while ago because I thought it could help people to automatically install the bitwuzla SMT solver (substitute for `apt` et cie.). If it is possible, it can be removed -- I think nobody is using it and no Opam package depends on it.
Ocaml user should use `bitwuzla`, `bitwuzla-c` or the latest version `bitwuzla-cxx` for access to the API.

### DIFF
--- a/packages/bitwuzla-bin/bitwuzla-bin.0.0.0/opam
+++ b/packages/bitwuzla-bin/bitwuzla-bin.0.0.0/opam
@@ -1,5 +1,6 @@
 opam-version: "2.0"
 synopsis: "Bitwuzla SMT solver executable"
+flags: deprecated
 description: """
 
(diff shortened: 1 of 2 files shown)


avsm:
Thanks! Do you also want to push a note into the `description` field pointing users to the right packages?

recoules:
Good point, I updated the description field of the package.

mseri:
Thanks

PR #24461, opened 2023-09-18
[new release] mirage-crypto, mirage-crypto-rng, mirage-crypto-rng-mirage, mirage-crypto-rng-lwt, mirage-crypto-rng-eio, mirage-crypto-rng-async, mirage-crypto-pk and mirage-crypto-ec (0.11.2)

hannesm:
Simple symmetric cryptography for the modern age

- Project page: <a href="https://github.com/mirage/mirage-crypto">https://github.com/mirage/mirage-crypto</a>
- Documentation: <a href="https://mirage.github.io/mirage-crypto/doc">https://mirage.github.io/mirage-crypto/doc</a>

##### CHANGES:

* mirage-crypto-rng-eio: improve portability by using eio 0.7's monotonic clock
  interface instead of mtime.clock.os. (mirage/mirage-crypto#176 @TheLortex)
* mirage-crypto-rng-eio: update to eio 0.12 (mirage/mirage-crypto#182 @talex5)
* mirage-crypto-rng: fix typo in RNG setup (mirage/mirage-crypto#179 @samueldurantes)
* macOS: on arm64 with clang 14.0.3, avoid instcombine (due to miscompilations)
  reported by @samoht https://github.com/mit-plv/fiat-crypto/issues/1606#issuecomment-1560122239
  re-reported in https://github.com/ulrikstrid/ocaml-jose/issues/63 and https://github.com/mirleft/ocaml-tls/issues/478
  (mirage/mirage-crypto#185 @hannesm @kit-ty-kate)
* avoid "stringop-overflow" warning on PPC64 and S390x (spurious warnings) when
  in devel mode (mirage/mirage-crypto#178 mirage/mirage-crypto#184 @avsm @hannesm)
* stricter C prototypes, unsigned/signed integers (mirage/mirage-crypto#175 @MisterDA @haesbaert
  @avsm @hannesm)
* support DragonFlyBSD (mirage/mirage-crypto#181 @movepointsolutions)
* support GNU/Hurd (mirage/mirage-crypto#174 @pinotree)

PR #24462, opened 2023-09-18
mark old ca-certs as unavailable (test cases fail on modern Linux systems)

hannesm:
as observed in https://github.com/ocaml/opam-repository/pull/24461 they fail on alpine and debian 12 systems. better mark them unavailable -- 0.2.3 is still present and works fine on these systems :)


kit-ty-kate:
It looks like only the tests fail. Shouldn’t only the tests be removed in that case? What is the underlying reason for the failures?

hannesm:
The underlying reason is how the package and the tests are achieved:
- the package itself provides access to the OS trust anchors (i.e. /usr/local/etc/ssl/cert.pem)
- the tests have been constructed by the time of writing to establish a TLS connection to a remote host, and the test is "using our current trust anchors, let's see whether they validate"

now, different OS add and remove trust anchors, the world moves forward, google replaces their certificate chain, etc. and it turns out that more and more Linux distributions do not support the certificate chains google presented in 2020 ;)

Looking at the library ca-certs and its commits, I do not see much value in keeping old versions thereof alive/installable. But of course, if you insist to "remove the tests", I'm happy to obey. Should such a removal be guarded by specific distributions, or entirely?

hannesm:
A much better test would be to use network access and test whether "using the current trust anchors, can we establish TLS connections to various endpoints" -- but here the opam sandbox avoids from doing that.

avsm:
The only possible problem with making them unavailable is that packages that formerly compiled with an old OCaml (e.g. 4.07) will no longer compile as the newer version has newer dependencies.

If the problem is just the tests failing, as opposed to the builds failing, another option is to mark them with an `avoid-version` flag so that they are not selected unless that version is explicitly specified. However, I am also not against marking the older versions unavailable if that's easier.  It's just nice to try to keep older software packages buildable (I haven't checked to see if this PR makes anything uninstallable).

hannesm:
> The only possible problem with making them unavailable is that packages that formerly compiled with an old OCaml (e.g. 4.07) will no longer compile as the newer version has newer dependencies.

True. I'm sorry for those stuck on such an old compiler. My lifetime is limited, and I won't deal with such an old compiler anyways.

> If the problem is just the tests failing, as opposed to the builds failing, another option is to mark them with an avoid-version flag so that they are not selected unless that version is explicitly specified.

Does `avoid-version` help with opam CI?

> However, I am also not against marking the older versions unavailable if that's easier. It's just nice to try to keep older software packages buildable (I haven't checked to see if this PR makes anything uninstallable).

If you have an old compiler or want to install old software, why not use an old checkout of opam-repository. I honestly don't see the value of that.

mseri:
So far we decided to stick with the current policy and keep the old package available, to clean the CI I have disabled the opam tests (they can still be run manually of course)

PR #24467, opened 2023-09-19
higher-bound constraint on ocaml version for mdx packages

raphael-proust:
There have been some errors in the revdeps of the CI of some other packages pointing to mdx issues. This PR adds missing upper-bound constraints.

### DIFF
--- a/packages/mdx/mdx.1.11.1/opam
+++ b/packages/mdx/mdx.1.11.1/opam
@@ -21,7 +21,7 @@ homepage: "https://github.com/realworldocaml/mdx"
 bug-reports: "https://github.com/realworldocaml/mdx/issues"
 depends: [
   "dune" {>= "2.7"}
-  "ocaml" {>= "4.02.3"}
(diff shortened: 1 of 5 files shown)


kit-ty-kate:
Example of such error:
```
# File "lib/top/mdx_top.ml", lines 217-220, characters 27-44:
# Warning 8 [partial-match]: this pattern-matching is not exhaustive.
# Here is an example of a case that is not matched:
# Pextra_ty (_, _)
```
Fixed upstream by https://github.com/realworldocaml/mdx/pull/435

mseri:
Thanks

talex5:
Marking MDX as incompatible with OCaml 5.1 means that projects using MDX are not getting much testing in ocaml-ci. For example, Eio is now only tested on one platform (debian-12-5.0_opam-2.1). All the other distributions and architectures try to test with 5.1, which is no longer available.

mseri:
We need to push for a fixed release of mdx

PR #24470, opened 2023-09-19
[new release] melange-numeral (0.0.1)

jchavarri:
Melange bindings for Numeral

- Project page: <a href="https://github.com/ahrefs/melange-numeral/">https://github.com/ahrefs/melange-numeral/</a>

##### CHANGES:

Initial release.

Migrated to Melange v2.

### DIFF
--- a/packages/melange-numeral/melange-numeral.0.0.1/opam
+++ b/packages/melange-numeral/melange-numeral.0.0.1/opam
@@ -18,8 +18,8 @@ depends: [
   "dune" {>= "3.8"}
   "melange" {>= "2.0.0"}
   "reason" {>= "3.10.0"}
-  "opam-check-npm-deps" {with-test} # todo: use with-dev-setup once opam 2.2 is out
-  "ocaml-lsp-server" {with-test}
+  "opam-check-npm-deps" {with-dev-setup}
+  "ocaml-lsp-server" {with-dev-setup}
   "odoc" {with-doc}
 ]
 build: [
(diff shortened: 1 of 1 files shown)


packages/melange-numeral/melange-numeral.0.0.1/opam
```diff
@@ -0,0 +1,51 @@
...
+tags: ["melange" "org:ahrefs"]
+homepage: "https://github.com/ahrefs/melange-numeral/"
+bug-reports: "https://github.com/ahrefs/melange-numeral/issues"
+depends: [
+  "ocaml"
+  "dune" {>= "3.8"}
+  "melange" {>= "2.0.0"}
+  "reason" {>= "3.10.0"}
+  "opam-check-npm-deps" {with-test} # todo: use with-dev-setup once opam 2.2 is out
+  "ocaml-lsp-server" {with-test}
```
kit-ty-kate, line 22:
```suggestion
  "opam-check-npm-deps" {with-dev-setup}
  "ocaml-lsp-server" {with-dev-setup}
```

jchavarri: I prefer `with-dev-setup` :) but then is there a way to install add these packages as deps while still using opam 2.1? I am imagining a way to set variables to true dynamically from the command line?

haochenx (approved):
CI is failing at this moment on macos machines but it seems that they are all the CI's fault. I have restarted the jobs.

Yet since all other CI jobs succeeded and there is no reason to think macos will fail,
It looks to me that this PR is mergeable.

packages/melange-numeral/melange-numeral.0.0.1/opam
kit-ty-kate: ah nevermind i now understand what you were saying. Sorry for my misunderstanding. Mmh, i guess the `opam var` incantation above is fine as a temporary solution although if it's just for yourself i would encourage you to use opam 2.2 directly instead.

kit-ty-kate:
Thanks!

PR #24448, opened 2023-09-15
[new release] ocamlformat-lib, ocamlformat and ocamlformat-rpc-lib (0.26.1)

Julow:
Auto-formatter for OCaml code

- Project page: <a href="https://github.com/ocaml-ppx/ocamlformat">https://github.com/ocaml-ppx/ocamlformat</a>

### Changed

- Compatible with OCaml 5.1.0 (ocaml-ppx/ocamlformat#2412, @Julow)
  The syntax of let-bindings changed sligthly in this version.
- Improved ocp-indent compatibility (ocaml-ppx/ocamlformat#2428, @Julow)
- \* Removed extra break in constructor declaration with comment (ocaml-ppx/ocamlformat#2429, @Julow)
- \* De-indent the `object` keyword in class types (ocaml-ppx/ocamlformat#2425, @Julow)
- \* Consistent formatting of arrows in class types (ocaml-ppx/ocamlformat#2422, @Julow)

### Fixed

- Fix dropped attributes on a begin-end in a match case (ocaml-ppx/ocamlformat#2421, @Julow)
- Fix dropped attributes on begin-end in an if-then-else branch (ocaml-ppx/ocamlformat#2436, @gpetiot)
- Fix non-stabilizing comments before a functor type argument (ocaml-ppx/ocamlformat#2420, @Julow)
- Fix crash caused by module types with nested `with module` (ocaml-ppx/ocamlformat#2419, @Julow)
- Fix ';;' formatting between doc-comments and toplevel directives (ocaml-ppx/ocamlformat#2432, @gpetiot)

PR #24476, opened 2023-09-19
diskuvbox.0.1.1 does not work with ocaml.5.1+

jonahbeckford:
Fix `revdeps` failure in https://github.com/ocaml/opam-repository/pull/24448 caused by `diskuvbox.0.1.1`


mseri:
Thanks

PR #24485, opened 2023-09-20
[new release] reason-react (2 packages) (0.13.0)

davesnx:
Reason bindings for React.js

- Project page: <a href="https://reasonml.github.io/reason-react">https://reasonml.github.io/reason-react</a>
- Documentation: <a href="https://reasonml.github.io/reason-react">https://reasonml.github.io/reason-react</a>

##### CHANGES:

* Support for Melange v2
* Remove legacy `ReactDOMRe` and `ReasonReact` modules (@anmonteiro in
  [reasonml/reason-react#782](https://github.com/reasonml/reason-react/pull/782))
* Fix `React.useCallback*` for callbacks with multiple arguments (@anmonteiro
  in [reasonml/reason-react#786](https://github.com/reasonml/reason-react/pull/786))

### DIFF
--- a/packages/reason-react-ppx/reason-react-ppx.0.13.0/opam
+++ b/packages/reason-react-ppx/reason-react-ppx.0.13.0/opam
@@ -18,7 +18,6 @@ depends: [
   "ocaml" {>= "5.1.0"}
   "reason" {>= "3.10.0"}
   "ppxlib" {>= "0.28.0"}
-  "merlin" {= "4.9-501preview" & with-test}
   "ocamlformat" {= "0.24.0" & with-dev-setup}
   "odoc" {with-doc}
 ]
--- a/packages/reason-react-ppx/reason-react-ppx.0.13.0/opam
+++ b/packages/reason-react-ppx/reason-react-ppx.0.13.0/opam
@@ -14,7 +14,8 @@ doc: "https://reasonml.github.io/reason-react"
 bug-reports: "https://github.com/reasonml/reason-react/issues"
 depends: [
   "dune" {>= "3.9"}
-  "ocaml" {>= "5.1.0" & < "5.2.0"}
+  "melange" {>= "2.0.0"}
+  "ocaml" {>= "5.1.0"}
   "reason" {>= "3.10.0"}
   "ppxlib" {>= "0.28.0"}
   "merlin" {= "4.9-501preview" & with-test}
--- a/packages/reason-react-ppx/reason-react-ppx.0.13.0/opam
+++ b/packages/reason-react-ppx/reason-react-ppx.0.13.0/opam
@@ -20,6 +20,7 @@ depends: [
   "ppxlib" {>= "0.28.0"}
   "merlin" {= "4.9-501preview" & with-test}
   "ocamlformat" {= "0.24.0" & with-dev-setup}
+  "reason-react" {with-test & =version & post}
   "odoc" {with-doc}
 ]
 build: [
--- a/packages/reason-react-ppx/reason-react-ppx.0.13.0/opam
+++ b/packages/reason-react-ppx/reason-react-ppx.0.13.0/opam
@@ -20,7 +20,6 @@ depends: [
   "ppxlib" {>= "0.28.0"}
   "merlin" {= "4.9-501preview" & with-test}
   "ocamlformat" {= "0.24.0" & with-dev-setup}
-  "reason-react" {with-test & =version & post}
   "odoc" {with-doc}
 ]
 build: [
@@ -33,7 +32,6 @@ build: [
     "-j"
     jobs
     "@install"
-    "@runtest" {with-test}
     "@doc" {with-doc}
   ]
 ]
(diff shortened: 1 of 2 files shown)


davesnx:
Tried to use `{post}` in [`d53a5da` (#24485)](https://github.com/ocaml/opam-repository/pull/24485/commits/d53a5da4ac0751000bfde62a8a80996289423de6) and it fails. Is there anything I'm missing @haochenx?

packages/reason-react-ppx/reason-react-ppx.0.13.0/opam
```diff
@@ -0,0 +1,47 @@
...
+homepage: "https://reasonml.github.io/reason-react"
+doc: "https://reasonml.github.io/reason-react"
+bug-reports: "https://github.com/reasonml/reason-react/issues"
+depends: [
+  "dune" {>= "3.9"}
+  "melange" {>= "2.0.0"}
+  "ocaml" {>= "5.1.0"}
+  "reason" {>= "3.10.0"}
+  "ppxlib" {>= "0.28.0"}
+  "merlin" {= "4.9-501preview" & with-test}
```
kit-ty-kate, line 21: do the test really rely on that strict of a constraint?

davesnx: We run ocamlmerlin on the cram tests (https://github.com/reasonml/reason-react/blob/main/ppx/test/keys.t/run.t#L24) and need to match the version of the OCaml compiler with Melange.

So, I'm afraid that yes, but I will cc @jchavarri since was his piece of code.

jchavarri: Afaik we don't need to constrain that much. I only added `with-test` in https://github.com/reasonml/reason-react/commit/e49bac3b435632afe35088fc20eec59e2dcaac17.

@davesnx you added a stricter constrain in https://github.com/reasonml/reason-react/commit/47663d0fbee1f54bbbb41399d86ffa66ff13ab48, then @anmonteiro refined it in https://github.com/reasonml/reason-react/commit/e1f6eeb0df2205369e332ebf1eb1310f54908d2b.

haochenx: @davesnx after some discussion with @kit-ty-kate, it seems that `{post}` will not do anything for this.

We reached the conclusion that the best way to handle it is to attach the tests to the `reason-react` packages (by changing https://github.com/reasonml/reason-react/blob/3971e42df555f3c8fb5b4eab94a26e8e51d79f02/ppx/test/dune#L2).

It seems that OPAM currently cannot handle this kind of circular dependency yet.

There are other tiny details regarding the test: it depends on `jq`, which means you should probably include `conf-jq` as a dependency.

We can (and probably should) merge this PR without enabling the testing.

davesnx: I remember making stricter. Since opam would resolve differently on CI and wanted to ensure the same version between envs. (It failed out snapshot tests due to a different version)

davesnx: So our options are:

- Don't run the tests on opam-repository (as now)
- Change the tests to run inside reason-react
- Remove the circular dependency all together (ship the reason-react-ppx under reason-react)

davesnx: I believe we are happy with the current approach and if it faces more issues we will go with unification of the package

kit-ty-kate: Actually since the tests are not ran in opam-repository there is no point in installing the tests dependencies in the first place
```suggestion
```

jchavarri: > Actually since the tests are not ran in opam-repository there is no point in installing the tests dependencies in the first place

In that case, is it possible to keep this line in the opam repository? It should be harmless, and it allows us to keep the opam file in the project repo in sync with the one submitted here more easily.

Alternatively, we can remove all the test and with-dev-setup deps to a new opam file `reason-react-dev.opam`. 🤔

haochenx: One issue (?) with leaving `"merlin" {= "4.9-501preview" & with-test}` in the opam file is that opam ci will assume there exist tests for the package (reason-react-ppx in this case) if `with-test` occurs in the opam file. This will waste the ci's time running opam install again with `--with-test`. But that's as far as the problem goes as I could see.

haochenx: > Change the tests to run inside reason-react (but still add conf-jq), but I'm not sure if it will work since both tests need each-other

Looks the best to me. It should work because currently reason-react lists reason-react-ppx as its dependency. So if you can get the tests run happily locally after changing https://github.com/reasonml/reason-react/blob/3971e42df555f3c8fb5b4eab94a26e8e51d79f02/ppx/test/dune#L2 to `(package reason-react)`, ci should be happily run the ppx tests as part of reason-react's tests.

IIUC changing which packages the tests are attached should make no difference to local development.

davesnx:
Looks ready to merge :D

mseri:
Thanks

davesnx:
Thanks @kit-ty-kate @haochenx and @mseri

PR #24487, opened 2023-09-20
[new release] trace (2 packages) (0.4)

c-cube:
A stub for tracing/observability, agnostic in how data is collected

- Project page: <a href="https://github.com/c-cube/ocaml-trace">https://github.com/c-cube/ocaml-trace</a>

##### CHANGES:

- add `?data` to `counter_int` and `counter_float`
- add `float` to user data
- add `add_data_to_current_span` and `add_data_to_manual_span`
- make `explicit_span.meta` mutable
- trace-tef: write to `trace.json` if env variable `TRACE` is either 1 or true
- trace-tef: emit function name, if provided, as a metadata key/value pair
- re-export trace.core in trace

- perf: in trace-tef, use broadcast instead of signal in the job queue

### DIFF
--- a/packages/trace-tef/trace-tef.0.4/opam
+++ b/packages/trace-tef/trace-tef.0.4/opam
@@ -1,0 +1,43 @@
+opam-version: "2.0"
+synopsis:
+  "A simple backend for trace, emitting Catapult/TEF JSON into a file"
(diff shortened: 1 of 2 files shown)


mseri:
Thanks

c-cube:
Thank you! :)

PR #24493, opened 2023-09-21
goblint-cil build patch to detect more recent gcc

raphael-proust:
addressing https://github.com/ocaml/opam-repository/pull/24423#issuecomment-1727804077


raphael-proust:
@sim642 The patch seems to work: gcc is correctly found on macos. There are now different errors.

- Is it ok to merge with these errors? (Do you want to mark the package a unavailable on some platforms because the errors indicate that the software is not correct there? Do you think those errors are ok?)
- If you want to copy the error logs you should do so before the PR is merged.

sim642:
I've copied the logs into upstream issues to look into. I think oraclelinux-9 and freebsd are both new for opam CI, so there might be something up with that because all the tests fail.

mseri:
`oraclelinux` should disappear soon while `freebsd` is here to stay :)

mseri:
Thanks

PR #24503, opened 2023-09-22
update not-ocamlfind (0.12): change build to avoid freebsd-specific build error

chetmurthy:
on Freebsd somehow the timestamps can get messed-up, causing a part of the build that should not run, to run anyway.  So I disabled that part of the build.


mseri:
Thanks

PR #24506, opened 2023-09-22
Add bounds for removed dune ctypes 0.1 extension

emillon:
The ctypes extension in Dune is documented as experimental (and has
version 0.x) so is not subject to the same compatibility as the rest of
Dune.

Dune 3.11.0 removed versions 0.1 and 0.2 of this extension, so packages
using it need to build with `dune < 3.11`.

Affected packages need to migrate to version 0.3. This should be just a
matter of changing the version of the extension in `dune-project`, but:
- it requires `(lang dune 3.7)`
- if C code refers to local include paths, these are now relative to the
  where the stanza is located. This avoids the need to specify paths
  twice. See ocaml/dune#6883.


emillon:
cc:

- @bobot (colibrics colibrilib flint)
- @pukkamustard (monocypher)
- @raphael-proust (tezos packages)

emillon:
I'll wait until this merged to submit the 3.11.0 run.

mseri:
Thanks

PR #24508, opened 2023-09-22
Package ocsigenserver.5.1.0

balat:
### `ocsigenserver.5.1.0`
A full-featured and extensible Web server
Ocsigen Server implements most features of the HTTP protocol, and has a very powerful extension mechanism that makes it very easy to plug your own OCaml modules for generating pages. Many extensions are already implemented, like a reverse proxy, content compression, access control, authentication, etc.

---
* Homepage: http://ocsigen.org/ocsigenserver/
* Source repo: git+https://github.com/ocsigen/ocsigenserver.git
* Bug tracker: https://github.com/ocsigen/ocsigenserver/issues/

---

### DIFF
--- a/packages/ocsigenserver/ocsigenserver.5.1.0/opam
+++ b/packages/ocsigenserver/ocsigenserver.5.1.0/opam
@@ -55,7 +55,7 @@ depends: [
   "re" {>= "1.11.0"}
   "cryptokit"
   "ipaddr" {>= "2.1"}
-  "cohttp-lwt-unix" {>= "5.0.0"}
+  "cohttp-lwt-unix" {>= "5.0.0" & < "6.0.0~"}
   "conduit-lwt-unix" {>= "2.0.0"}
   "hmap"
   "xml-light"
--- a/packages/ocsigenserver/ocsigenserver.5.1.0/opam
+++ b/packages/ocsigenserver/ocsigenserver.5.1.0/opam
@@ -1,0 +1,75 @@
+opam-version: "2.0"
+maintainer: "dev@ocsigen.org"
+synopsis: "A full-featured and extensible Web server"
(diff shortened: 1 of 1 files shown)


tmcgilchrist:
Looks like this is missing a dependency on Pcre library or the downstream packages like Eliom are missing that dependency. 
Perhaps things have changed since https://github.com/ocaml/opam-repository/pull/20902 which added Eliom.10.0.0?

glondu:
One feature of the new release is to drop the dependency on pcre!

I've opened a PR on eliom: https://github.com/ocsigen/eliom/pull/764. But CI fails because ocsigenserver.5.1.0 is missing in opam...

Meanwhile, maybe the simplest thing to do is to add depend/conflict relationships so that eliom.10.0.0 cannot be installed with ocsigenserver.5.1.0?

mseri:
Indeed, with the upper bounds it is all green

packages/ocsigenserver/ocsigenserver.5.1.0/opam
```diff
@@ -0,0 +1,75 @@
...
+  "react"
+  "ssl" {>= "0.5.8"}
+  "lwt" {>= "3.0.0"}
+  "lwt_ssl"
+  "lwt_react"
+  "lwt_log"
+  "re" {>= "1.11.0"}
+  "cryptokit"
+  "ipaddr" {>= "2.1"}
+  "cohttp-lwt-unix" {>= "5.0.0"}
```
mseri, line 58:
```suggestion
  "cohttp-lwt-unix" {>= "5.0.0" & < "6.0.0~"}
```
Do you know if this is compatible with the new cohttp alphas? If not, I suggest to add this bound alraedy

mseri:
Thanks

balat:
> Thanks 

Thank you!

PR #24313, opened 2023-08-25
5 packages from c-cube/qcheck

jmid:
This pull-request concerns:
-`ppx_deriving_qcheck.0.4.1`: PPX Deriver for QCheck
-`qcheck.0.21.2`: Compatibility package for qcheck
-`qcheck-alcotest.0.21.2`: Alcotest backend for qcheck
-`qcheck-core.0.21.2`: Core qcheck library
-`qcheck-ounit.0.21.2`: OUnit backend for qcheck

---
* Homepage: https://github.com/c-cube/qcheck/

---
:camel: Pull-request generated by opam-publish v2.1.0

PR #23805, opened 2023-05-19
4 packages from c-cube/qcheck at 0.21.1

jmid:
This pull-request concerns:
-`qcheck.0.21.1`: Compatibility package for qcheck
-`qcheck-alcotest.0.21.1`: Alcotest backend for qcheck
-`qcheck-core.0.21.1`: Core qcheck library
-`qcheck-ounit.0.21.1`: OUnit backend for qcheck

---
* Homepage: https://github.com/c-cube/qcheck/
* Source repo: git+https://github.com/c-cube/qcheck.git
* Bug tracker: https://github.com/c-cube/qcheck/issues

---
:camel: Pull-request generated by opam-publish v2.1.0

PR #23741, opened 2023-05-02
5 packages from c-cube/qcheck

jmid:
This pull-request concerns:
-`ppx_deriving_qcheck.0.4.0`: PPX Deriver for QCheck
-`qcheck.0.21`: Compatibility package for qcheck
-`qcheck-alcotest.0.21`: Alcotest backend for qcheck
-`qcheck-core.0.21`: Core qcheck library
-`qcheck-ounit.0.21`: OUnit backend for qcheck

---
* Homepage: https://github.com/c-cube/qcheck/

---
:camel: Pull-request generated by opam-publish v2.1.0

PR #24540, opened 2023-09-28
Patch prbnmcn-stats 0.0.2 and 0.0.3 tests

jmid:
`prbnmcn-stats.0.0.2` and `0.0.3` contain a 'convergence test' which fails on OCaml 5.
The failure triggers whenever I roll a new release of QCheck, on which `prbnmcn-stats` depends, and thus adds needless noise. See, e.g., these 3 most recent occurrences: https://github.com/ocaml/opam-repository/pull/24313 https://github.com/ocaml/opam-repository/pull/23805 https://github.com/ocaml/opam-repository/pull/23741

The test failure does not trigger on OCaml 4 - with 4.12 being the lower bound in the opam file.
This PR therefore contains a small patch to disable the failing test on OCaml 5+.
A cruder fix would be to disable all tests on OCaml 5+.


jmid:
I suppose an even cruder fix would be to disable these two versions on OCaml 5+... :slightly_smiling_face:

mseri:
Thanks. Yes, we could have just disabled them, but the fix is so small... why not

PR #24375, opened 2023-09-08
conf-unwind.1 with a fix for os-family=ubuntu

jmid:
This PR contains a `conf-unwind.1` with a small fix for `os-family=ubuntu`.
This should help new users wanting to try out `5.1.0~rc3+tsan` after the presentation at tomorrow's OCaml workshop.

Without the fix I got:
```
#=== ERROR while compiling conf-unwind.0 ======================================#
# context     2.1.4 | linux/x86_64 |  | https://opam.ocaml.org#62c9cbe6
# path        ~/.opam/5.1.0~rc3+tsan/.opam-switch/build/conf-unwind.0
# command     ~/.opam/opam-init/hooks/sandbox.sh build pkg-config --short-errors --print-errors libunwind
# exit-code   1
# env-file    ~/.opam/log/conf-unwind-1330963-be162d.env
# output-file ~/.opam/log/conf-unwind-1330963-be162d.out
### output ###
# No package 'libunwind' found
```

I'm on Linux Mint, an Ubuntu-derivative which reports
```
$ opam var os-family 
ubuntu
```
and for which the relevant deb-package is also just called `libunwind-dev`

```diff
$ diff packages/conf-unwind/conf-unwind.0/opam  packages/conf-unwind/conf-unwind.1/opam 
10a11
>   ["libunwind-dev"] {os-family = "ubuntu"}
```

CC: @OlivierNicole @dra27

PR #24545, opened 2023-09-28
Fix conf-openbabel on openbabel 3

jmid:
Here's a third PR to reduce noise - this one in `conf-openbabel`.
The error triggers when I release a new QCheck release, e.g. in #24313 because of the dependency of `lbvs_consent`  on `conf-openbabel` which therefore fails with, e.g., `lbvs_consent.2.1.2 (failed: conf-openbabel.0.1 failed to build)`

The issue here is that the `openbabel` version obtained by `apt-get install libopenbabel-dev` [is installed in `/usr/include/openbabel3`](https://packages.debian.org/sid/amd64/libopenbabel-dev/filelist)
I've also checked locally on my homebrew mac, that `brew install open-babel` installs into `/usr/local/include/openbabel3`

For backwards compatibility, I've left the `openbabel-2.0` entries as is.

Finally, `opam var os-family` yields `ubuntu` on Linux Mint, so I included a fix akin to https://github.com/ocaml/opam-repository/pull/24375

### DIFF
--- a/packages/conf-openbabel/conf-openbabel.0.1/files/test.cpp
+++ b/packages/conf-openbabel/conf-openbabel.0.1/files/test.cpp
@@ -1,8 +1,0 @@
-#include <openbabel/mol.h>
-
-// compile with: c++ -I/usr/include/openbabel-2.0 test.c -lopenbabel
-
-int main () {
-  OpenBabel::OBMol mol;
-  return 0;
-}
--- a/packages/conf-openbabel/conf-openbabel.0.1/opam
+++ b/packages/conf-openbabel/conf-openbabel.0.1/opam
@@ -4,22 +4,8 @@ homepage: "http://openbabel.org/"
 license: "GPL-1.0-or-later"
 authors: "http://openbabel.org/wiki/THANKS"
 bug-reports: "https://github.com/ocaml/opam-repository/issues"
-build: [
-  [ "c++"
-    "-I/usr/include/openbabel-2.0"
-    "-I/usr/include/openbabel3"
-    "test.cpp"
-    "-lopenbabel"
-  ] {os != "macos"}
-  [
-    "c++"
-    "-I/usr/local/include/openbabel-2.0"
-    "-I/usr/local/include/openbabel3"
-    "test.cpp"
-    "-L/usr/local/lib"
-    "-lopenbabel"
-  ] {os = "macos"}
-]
+build: [["pkg-config" "openbabel-3"]]
+depends: ["conf-pkg-config" {build}]
 x-ci-accept-failures: ["debian-11" "debian-unstable" "alpine-3.18"]
 depexts: [
   ["libopenbabel-dev"] {os-family = "debian" | os-family = "ubuntu"}
@@ -34,5 +20,4 @@ synopsis: "Virtual package relying on openbabel library installation"
 description: """
 This package can only install if the openbabel devel library is installed
 on the system."""
-extra-files: ["test.cpp" "md5=0a2f6ad381a1624089cd12718bfbfdd8"]
 flags: conf
--- a/packages/conf-openbabel/conf-openbabel.0.1/opam
+++ b/packages/conf-openbabel/conf-openbabel.0.1/opam
@@ -22,7 +22,7 @@ build: [
 ]
 x-ci-accept-failures: ["debian-11" "debian-unstable"]
 depexts: [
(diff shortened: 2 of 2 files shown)


jmid:
I realize that it would probably make sense to switch to something like `pkg-config`...

To eliminate red lights that happen for a different reason, the latest string of commits add support on Arch, Fedora, opensuse, Oracle Linux, and FreeBSD. 
I couldn't find an `*open*babel*` package for Alpine on https://pkgs.alpinelinux.org/packages

jmid:
7344ad4 suggests to switch the test to using `pkg-config` like many other `conf-*` packages.
I've checked locally that this also works on macOS with homebrew.

jmid:
CI is full of red lights. 3a09c95 fixes the Fedora package to use `openbabel-devel`.
I need to go over the other distro package names again.

jmid:
OK, fixed the package names on opensuse and oracle.
The only remaining error should be on Debian 10 which has a version 2.* of the package.
The `pkg-config` query therefore has to be adjusted accordingly...

jmid:
The Debian 10 and opensuse 15.5 error should be gone with 6c0a5c0 which tests for either version 2.0 or 3.
There's still an Oracle error though :thinking:

jmid:
From https://github.com/ocaml/opam-repository/pull/24360#issuecomment-1709640614 I learned that Oracle Linux does not search the EPEL repo where the system package is located https://yum.oracle.com/repo/OracleLinux/OL9/developer/EPEL/x86_64/index.html

I could revert the Oracle addition and/or add a suitable `x-ci-accept-failures` for it?

jmid:
The `lbvs_consent` package fails because it hard codes the `/usr/include/openbabel-2.0` directory:
https://github.com/UnixJunkie/consent/blob/master/src/Makefile#L2

One can
- rewrite the above to use `pkg-config` instead so that it will also work with `openbabel3`
- split this PR into two opam packages: `conf-openbabel.2` and  `conf-openbabel.3`

@UnixJunkie WDYT?

UnixJunkie:
If that doesn't break anything, then I am OK with the proposed changes.
It does become simpler than what I initially wrote.

jmid:
9fa6afb (and 38a92ee) adds the proposed `x-ci-accept-failures` from above.

> If that doesn't break anything, then I am OK with the proposed changes.
> It does become simpler than what I initially wrote.

lbvs_consent.2.1.1, 2.1.2, 2.1.3, 2.2.0 all fail the CI during  `make obmaccs` in `src/` because of the hard coded `/usr/include/openbabel-2.0` directory, when the tested CI system offers `openbabel3` instead.
I don't know enough about openbabel to tell whether `lbvs_consent` and other software will work equally well with openbabel-2.0 and  openbabel3.

If we go with the current version that accepts either 2.0 or 3, `lbvs_consent` will have to add a `conf-pkg-config` dependency and adjust `src/Makefile` with suitable `pkg-config --libs --cflags openbabel-2.0` and `pkg-config --libs --cflags openbabel-3` calls to work cross platform, AFAICS.

jmid:
Something like the following seems sufficient to get https://github.com/UnixJunkie/consent/blob/master/src/Makefile working with both openbabel 2.0 and 3:
```diff
@@ -1,6 +1,9 @@
+CFLAGS := `pkg-config --silence-errors --cflags openbabel-2.0 || pkg-config --cflags openbabel-3`
+LDFLAGS := `pkg-config --silence-errors --libs openbabel-2.0 || pkg-config --libs openbabel-3`
+
 ob_maccs: ob_maccs.cpp
-       c++ -W -Wall -I/usr/include/openbabel-2.0 ob_maccs.cpp \
-         -o lbvs_consent_ob_maccs -lopenbabel -Wl,-R/usr/lib
+       c++ -W -Wall $(CFLAGS) ob_maccs.cpp \
+         -o lbvs_consent_ob_maccs $(LDFLAGS) -Wl,-R/usr/lib
 
 clean:
        \rm -f lbvs_consent_ob_maccs

```

jmid:
Seeing many unrelated `ppx_rapper` errors triggered. Fix PR in https://github.com/ocaml/opam-repository/pull/24589.
I'll rebase once that one is in.

jmid:
Rebased on `master` after the merge of https://github.com/ocaml/opam-repository/pull/24589

mseri:
Sorry for the long delay and thanks a lot for the fix! Looks like nothing is broken and instead compatibility is better now. I am going to merge

PR #24546, opened 2023-09-28
Support conf-unwind on FreeBSD, Arch, OpenSuse, and Oracle Linux

jmid:
This little PR adds `conf-unwind` support for FreeBSD where [the underlying system package should be installable as `libunwind`](https://www.freshports.org/devel/libunwind/)

I expect errors from the still unsupported Linux distros (arch, opensuse, oraclelinux) and macOS.

### DIFF
--- a/packages/conf-unwind/conf-unwind.0/opam
+++ b/packages/conf-unwind/conf-unwind.0/opam
@@ -18,6 +18,11 @@ depexts: [
   ["libunwind-headers"] {os-distribution = "homebrew" & os = "macos"}
   ["libunwind"] {os = "freebsd"}
 ]
+x-ci-accept-failures: [
+￼ "oraclelinux-7"
+￼ "oraclelinux-8"
+￼ "oraclelinux-9"
+]
 synopsis: "Virtual package relying on libunwind"
 description:"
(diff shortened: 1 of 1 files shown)


jmid:
Seeing errors in CI for the remaining two Linux distros tested, the last two commit add support for them.
This leaves only macOS.

jmid:
The required `*.pc` files for `pkg-config` are missing in homebrew https://github.com/Homebrew/brew/pull/5539 so the last commit disables the `pkg-config` test there.

jmid:
I adjusted package names on Oracle and OpenSuse which were the only two remaining distros failing CI.
(there are dependencies that fail though)

jmid:
From https://github.com/ocaml/opam-repository/pull/24360#issuecomment-1709640614 I learned that the Oracle Linux does not search the EPEL repo where the system package is located https://yum.oracle.com/repo/OracleLinux/OL9/developer/EPEL/x86_64/index.html

I could revert the Oracle addition and/or add a suitable `x-ci-accept-failures` for it?

jmid:
In 950bf78 I propose a suitable `x-ci-accept-failures`

avsm:
Thanks!

PR #24547, opened 2023-09-28
satyrographos.0.0.2.13: Update the lower version constraint on core

na4zagin3:
`satyrographos.0.0.2.13` actually compiles with `core.v0.14`.  As `satysfi.0.0.8`, which is in a custom repo, requires `base.v0.14`, the lower version constraint on core should be updated.


mseri:
thanks

PR #24548, opened 2023-09-28
Add an upper bound for facile.1.1

jmid:
Facile.1.1 uses `Pervasives` and hence doesn't work on OCaml 5+

### DIFF
--- a/packages/facile/facile.1.1/opam
+++ b/packages/facile/facile.1.1/opam
@@ -10,7 +10,7 @@ patches: [
 ]
 install: [make "install"]
 synopsis: "Constraint programming library over integer finite domains"
-depends: ["ocaml"]
+depends: ["ocaml" {< "5.0.0"}]
 extra-files: [
   ["ocaml_4.00.patch" "md5=cbc214d5dd007d1062821b4d96e85ad0"]
   ["facile.install" "md5=a03d20fcd54d7277982caf03b05354fa"]


jmid:
There were 9 internal failures and some opam linting errors.

f6ad25d adds the missing fields from the 1.1.4 opam file to pass linting.

jmid:
The only remaining error on the latest run is an opam lint failure about a missing license.
I therefore downloaded the tar-ball and checked that it lists a LGPL-2.1 (or later) and so added that field.

haochenx (approved):
All looks good to me.

Unrelated to this patch, but maybe it's worth considering adding sha256 checksums for the tarball, esp considering the URL is HTTP (instead of HTTPS)

jmid:
There's already an m5sum. Do you mean you want to replace it?
https://github.com/ocaml/opam-repository/blob/df27dc1d73b552201c02b73a11d761b97ba51394/packages/facile/facile.1.1/opam#L26-L27

haochenx:
I don't think there is a formal policy / recommendation for opam-repository regarding this at this moment, but since md5 is essentially broken, adding a stronger hash is good thing.

You can have multiple checksum, e.g. https://github.com/ocaml/opam-repository/blob/3f23c19538db9c4a721c575dfe79f3463df0ee0a/packages/dune/dune.3.10.0/opam#L52-L55

PR #24492, opened 2023-09-20
add bunjs (https://bun.js) packages

haochenx:
Packages generated with https://github.com/haochenx/opam-bunjs-gen

- Bun homepage: https://bun.sh/
- Bun GitHub: https://github.com/oven-sh/bun

PR #24557, opened 2023-09-30
add conf-bunjs.1

haochenx:
This PR adds the packages `conf-bunjs` that checks whether Bun (https://bun.sh) is installed on the system.

This is a rebake of https://github.com/ocaml/opam-repository/pull/24492 to instead of adding binary packages to opam-repository, add a conf package that only checks whether `bun` and `bunx` is already installed and available on the system.

Since currently there is no good way to install `bun` on the CI machines (as depexts cannot handle npm packages, nor does Bun has good system package manager support), this package will likely fail all the CI tests.

### DIFF
--- a/packages/conf-bunjs/conf-bunjs.1/opam
+++ b/packages/conf-bunjs/conf-bunjs.1/opam
@@ -9,6 +9,9 @@ build: [
   ["bun" "--version"]
   ["bunx" "bun" "--version"]
 ]
+depexts: [
+  [ "bun" ] {npm-version = "^1.0.0"}
+]
 post-messages: [
   """
 Command `bun` and/or `bunx` is not available on the system.
--- a/packages/conf-bunjs/conf-bunjs.1/opam
+++ b/packages/conf-bunjs/conf-bunjs.1/opam
@@ -1,0 +1,32 @@
+opam-version: "2.0"
+maintainer: "haochenx@acm.org"
+authors: ["Jarred Sumner" "Bun Developers and Contributors"]
+homepage: "https://github.com/oven-sh/bun"
+bug-reports: "https://github.com/oven-sh/bun/issues"
+dev-repo: "git+https://github.com/oven-sh/bun.git"
+license: "MIT"
+build: [
+  ["bun" "--version"]
+  ["bunx" "bun" "--version"]
+]
+post-messages: [
+  """
+Command `bun` and/or `bunx` is not available on the system.
+You can install them by running e.g. `npm install -g bun`
+or consult https://bun.sh/docs/installation for the official
+installation instructions for Bun (https://bun.sh).
+""" { failure }
+]
+synopsis: "Virtual package relying on Bun (https://bun.sh)"
+description: """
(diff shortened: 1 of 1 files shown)


haochenx:
>Since currently there is no good way to install bun on the CI machines (as depexts cannot handle npm packages, nor does Bun has good system package manager support), this package will likely fail all the CI tests.

Questions to opam-repository maintainers: should I add `x-ci-accept-failures` for the failing jobs? Or it's better to leave it as it as a reminder for future fix? Or is there a better way to deal with this situation?

packages/conf-bunjs/conf-bunjs.1/opam
```diff
@@ -0,0 +1,32 @@
...
+authors: ["Jarred Sumner" "Bun Developers and Contributors"]
+homepage: "https://github.com/oven-sh/bun"
+bug-reports: "https://github.com/oven-sh/bun/issues"
+dev-repo: "git+https://github.com/oven-sh/bun.git"
+license: "MIT"
+build: [
+  ["bun" "--version"]
+  ["bunx" "bun" "--version"]
+]
+post-messages: [
```
mseri, line 12:
```suggestion
depexts: [
  [ "bun" ] {npm-version = "^1.0.0"}
]
post-messages: [
```

packages/conf-bunjs/conf-bunjs.1/opam
```diff
@@ -0,0 +1,35 @@
...
+homepage: "https://github.com/oven-sh/bun"
+bug-reports: "https://github.com/oven-sh/bun/issues"
+dev-repo: "git+https://github.com/oven-sh/bun.git"
+license: "MIT"
+build: [
+  ["bun" "--version"]
+  ["bunx" "bun" "--version"]
+]
+depexts: [
+  [ "bun" ] {npm-version = "^1.0.0"}
```
mseri, line 13: This allows one to use https://github.com/jchavarri/opam-check-npm-deps to check if the npm deps are properly installed, so it won't really solve this problem

But opam will support npm after 2.2, so perhaps is not bad to keep this here

haochenx:
>This is just a runtime replacement for javascript right? Are there packages that do not work without bun? I am not 100% sure of the need for this package

Since there are packages that deal with JavaScript / TypeScript (e.g. js_of_ocaml & melange, and our unpublished [bindoj](https://github.com/kxcdev/bindoj)), it is sensible for packages that require a JavaScript runtime to do testing, source/config generation etc.

Bun is another runtime replacement of JavaScript, but it has its unique set of features (e.g. being fast and executing TypeScript out-of-the-box) it is sometimes desirable to use it over e.g. npm.

packages/conf-bunjs/conf-bunjs.1/opam
haochenx: >But opam will support npm after 2.2

Do you have a reference to it?

haochenx:
close to reduce noise as there is currently no clear path to go forward.

PR #24564, opened 2023-10-02
Move legacy Coq archives

maximedenes:
The current URLs (https://coq.inria.fr/distrib/xxx) are going to break soon.


maximedenes:
I just want to update the URLs, I'm not maintaining these packages. Is there a way to bypass the linter and check if installation succeeds?

avsm:
Since the archive checksums haven't changed, this doesn't trigger a recompilation of the packages. You can use https://check.ci.ocaml.org to see if these packages are building normally as part of the opam repo bulk build runs.  In the meanwhile, I'll merge these -- thanks for the fixes!

maximedenes:
Thanks a lot!

PR #23845, opened 2023-05-31
Release v0.16 of Core

dkalinichenko-js:
This PR adds Core v0.16 & dependencies.

PR #24565, opened 2023-10-02
Package ppx_inline_test.v0.16.1

dkalinichenko-js:
### `ppx_inline_test.v0.16.1`
Syntax extension for writing in-line tests in ocaml code
Part of the Jane Street's PPX rewriters collection.

---
* Homepage: https://github.com/janestreet/ppx_inline_test
* Source repo: git+https://github.com/janestreet/ppx_inline_test.git
* Bug tracker: https://github.com/janestreet/ppx_inline_test/issues

---

### DIFF
--- a/packages/ppx_inline_test/ppx_inline_test.v0.16.1/opam
+++ b/packages/ppx_inline_test/ppx_inline_test.v0.16.1/opam
@@ -13,7 +13,7 @@ depends: [
   "ocaml"    {>= "4.14.0"}
   "base"     {>= "v0.16" & < "v0.17"}
   "time_now" {>= "v0.16" & < "v0.17"}
-  "dune"     {>= "3.8.0"}
+  "dune"     {>= "3.11.0"}
   "ppxlib"   {>= "0.28.0"}
 ]
 available: arch != "arm32" & arch != "x86_32"
--- a/packages/ppx_inline_test/ppx_inline_test.v0.16.1/opam
+++ b/packages/ppx_inline_test/ppx_inline_test.v0.16.1/opam
@@ -13,7 +13,7 @@ depends: [
   "ocaml"    {>= "4.14.0"}
   "base"     {>= "v0.16" & < "v0.17"}
   "time_now" {>= "v0.16" & < "v0.17"}
-  "dune"     {>= "3.11.0"}
+  "dune"     {>= "3.11.1"}
   "ppxlib"   {>= "0.28.0"}
 ]
 available: arch != "arm32" & arch != "x86_32"
--- a/packages/ppx_inline_test/ppx_inline_test.v0.16.1/opam
+++ b/packages/ppx_inline_test/ppx_inline_test.v0.16.1/opam
@@ -1,0 +1,31 @@
+opam-version: "2.0"
+maintainer: "Jane Street developers"
+authors: ["Jane Street Group, LLC"]
+homepage: "https://github.com/janestreet/ppx_inline_test"
+bug-reports: "https://github.com/janestreet/ppx_inline_test/issues"
+dev-repo: "git+https://github.com/janestreet/ppx_inline_test.git"
+doc: "https://ocaml.janestreet.com/ocaml-core/latest/doc/ppx_inline_test/index.html"
+license: "MIT"
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+]
+depends: [
+  "ocaml"    {>= "4.14.0"}
+  "base"     {>= "v0.16" & < "v0.17"}
+  "time_now" {>= "v0.16" & < "v0.17"}
+  "dune"     {>= "3.8.0"}
+  "ppxlib"   {>= "0.28.0"}
+]
+available: arch != "arm32" & arch != "x86_32"
+synopsis: "Syntax extension for writing in-line tests in ocaml code"
+description: "
+Part of the Jane Street's PPX rewriters collection.
+"
+url {
+  src:
+    "https://github.com/janestreet/ppx_inline_test/archive/refs/tags/v0.16.1.tar.gz"
+  checksum: [
+    "md5=470acc022eb7015862307dad8fe9bcba"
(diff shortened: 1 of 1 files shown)


haochenx:
Can you confirm whether [this pattern](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/eb16cc9e3f080884b88cc170a7be9f399992f9fe/variant/compilers,4.14,ppx_inline_test.v0.16.1,revdeps,caisar.0.2.1) of the revdep failure is a regression?
```
=== recompile 1 package
  - recompile caisar 0.2.1

Processing  1/4:
-> retrieved caisar.0.2.1  (cached)
Processing  2/4: [caisar: dune build]
+ /home/opam/.opam/opam-init/hooks/sandbox.sh "build" "dune" "build" "-p" "caisar" "-j" "255" "--promote-install-files=false" "@install" "@runtest" (CWD=/home/opam/.opam/4.14/.opam-switch/build/caisar.0.2.1)
- File "lib/ir/dune", line 11, characters 1-15:
- 11 |  (inline_tests)
-       ^^^^^^^^^^^^^^
- Error: Rule dependencies are configured to require sandboxing, but the rule
- has no actions that could potentially require sandboxing.
[ERROR] The compilation of caisar.0.2.1 failed at "dune build -p caisar -j 255 --promote-install-files=false @install @runtest".

#=== ERROR while compiling caisar.0.2.1 =======================================#
# context              2.2.0~alpha3~dev | linux/x86_64 | ocaml-base-compiler.4.14.1 | file:///home/opam/opam-repository
# path                 ~/.opam/4.14/.opam-switch/build/caisar.0.2.1
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p caisar -j 255 --promote-install-files=false @install @runtest
# exit-code            1
# env-file             ~/.opam/log/caisar-7-7caebd.env
# output-file          ~/.opam/log/caisar-7-7caebd.out
### output ###
# File "lib/ir/dune", line 11, characters 1-15:
# 11 |  (inline_tests)
#       ^^^^^^^^^^^^^^
# Error: Rule dependencies are configured to require sandboxing, but the rule
# has no actions that could potentially require sandboxing.
```

dkalinichenko-js:
`ppx_inline_tests.v0.16.0` was released in #23845, but I can't load the logs for that PR to check. @hhugo could this error be caused by your patch?

hhugo:
It could be the consequence of my patch but sounds like an issue inside dune cc @rgrinberg

hhugo:
I can confirm there is an issue in the dune implementation of inline_tests. When no inline test is declared (aka empty partition list) in a library using `(inline_tests)`, dune complains. This PR should probably stay on hold until dune is fixed.

haochenx:
@hhugo @dkalinichenko-js thank you for the confirmation. I've marked this PR as do-not-merge until the fix per your suggestion.

hhugo:
x-ref https://github.com/ocaml/dune/issues/8848

hhugo:
I've proposed a fix in https://github.com/ocaml/dune/pull/8849

hhugo:
I don't know if we should mark broken packages as incompatible with this new ppx_inline_tests or if we should have ppx_inline_tests depend on a fixed version of dune.

haochenx:
IMO the latter will be easier to specify and will also cover private projects that might have been affected.

So unless there is a strong preference to publish this package ASAP then I think that'd be the preferred approach.

We can also mark this package with `flags: [ avoid-version ]` and publish for now and add the dependency to a fixed version of Dune later and a middle ground.

hhugo:
There is nothing pressing. When do you expect the next release of dune ?

haochenx:
I see that now https://github.com/ocaml/dune/pull/8849 is merged upstream. It's at Dune's maintainer's discretion on when to make a new release.

@emillon @rgrinberg, what do you think about the perspective of making a (patch?) release sooner rather than later to expedite this PR?

emillon:
I plan to do a patch release early next week. That should not prevent this from being merged though, since we'll need to get the right metadata regarding the dune versions that this release requires. Is there an existing version of dune that's compatible with this release?

haochenx:
>I plan to do a patch release early next week.

Thanks for the response and great to hear that!

>That should not prevent this from being merged though, since we'll need to get the right metadata regarding the dune versions that this release requires.

I was imaging to put a `dune {>= "3.11.1"}` constraint (giving `3.11.1` being the patch release containing https://github.com/ocaml/dune/pull/8849) so the metadata would be correct.

>Is there an existing version of dune that's compatible with this release?

It's not clear to me whether such a version may exist but I'm happy to spend some time tomorrow and try to find one. Do you have a suggestion on which version to start looking at?

hhugo:
> Is there an existing version of dune that's compatible with this release?

@emillon, the only change in this release is the switch to dune lang 3.8 and the use of partitions to run inline tests.
It only exist to start running inline tests in parallel.

Adding a `dune {>= "3.11.1"}` constraint seems to be the way to go.

packages/ppx_inline_test/ppx_inline_test.v0.16.1/opam
```diff
@@ -13,7 +13,7 @@ depends: [
   "ocaml"    {>= "4.14.0"}
   "base"     {>= "v0.16" & < "v0.17"}
   "time_now" {>= "v0.16" & < "v0.17"}
-  "dune"     {>= "3.8.0"}
+  "dune"     {>= "3.11.0"}
```
hhugo, line 16: It should be `3.11.1` once released ..

dkalinichenko-js: Fixed

emillon:
ok! dune release is in https://github.com/ocaml/opam-repository/pull/24607

hhugo:
This can be rebased. Dune 3.11.1 has been released

mseri:
The failures all seem independent from this PR

PR #24567, opened 2023-10-03
2 packages from jtcoolen/ocaml-pari at 0.1

jtcoolen:
This pull-request concerns:
- `pari.0.1`: Type-safe wrapper over the PARI library
- `pari-bindings.0.1`: OCaml bindings to the PARI library

---
* Homepage: https://github.com/jtcoolen/ocaml-pari
* Source repo: git+https://github.com/jtcoolen/ocaml-pari.git
* Bug tracker: https://github.com/jtcoolen/ocaml-pari/issues

---


packages/pari/pari.0.1/opam
```diff
@@ -0,0 +1,42 @@
...
+maintainer: ["Julien Coolen <jtcoolen@pm.me>"]
+authors: ["Julien Coolen"]
+license: "GPL-2.0"
+homepage: "https://github.com/jtcoolen/ocaml-pari"
+bug-reports: "https://github.com/jtcoolen/ocaml-pari/issues"
+depends: [
+  "ocaml" {>= "4.10"}
+  "dune" {>= "3.7" & >= "3.7"}
+  "core" {>= "v0.14.0"}
+  "pari-bindings" {dev}
```
raphael-proust, line 13:
```suggestion
  "pari-bindings" {= version}
```

The `version` version is a special value that means "the same version as this package". So it essentially means "this package (pari) depends on that package (pari-bindings), but that package (pari-bindings) can only be installed with the same (=) version as this package (pari)".

raphael-proust:
There are a few errors.

- In linting, there is an error in the license field. The license field should have one of the entries from the Identifer column of the table on this page https://spdx.org/licenses/. Probably `GPL-2.0-only` or `GPL-2.0-or-later`.

- The build error for pari-bindings seems to be in the build files. I'd recommend to try to build on a different machine if you have access to one.

- The build error for pari I think might be fixed by the suggestion above. Essentially `{dev}` doesn't make real dependencies.

packages/pari/pari.0.1/opam
jtcoolen: Thanks! Applied your piece of advice in the upstream sources.

jtcoolen:
Thanks for your review @raphael-proust ! It should be fixed now.

raphael-proust:
There seems to be a problem on freebsd (`ld: library not found for -lgmp`) but I think that's more likely to be an issue with the `conf-gmp` package or possibly with the CI (the platform is marked as experimental).

There's another error that's just https://github.com/ocaml/ocaml/issues/11948  
We could possibly add an `available` rule to prevent installation on arm32+ocaml4.14

packages/pari-bindings/pari-bindings.0.1/opam
```diff
@@ -0,0 +1,42 @@
...
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
```
raphael-proust, line 34:
```suggestion
]
available: [ arch != "arm32" | "ocaml-version" > "4.14" ]
```

jtcoolen: Applied!

jtcoolen:
The package fails to build on FreeBSD and on ARM Macs.

jtcoolen:
Seems to be good to go! :rocket: @raphael-proust @kit-ty-kate @mseri

PR #24572, opened 2023-10-04
conf-gtksourceview3: Must fail on oraclelinux-9 as well

sacerdot:
No package available


sacerdot:
Both failures were probably there in the previous version as well. One is a reverse dependency. I don't know what to do.

haochenx:
Thank you for the PR! As the same as https://github.com/ocaml/opam-repository/pull/24573#issuecomment-1746917585, we'd usually like a patch to the existing conf package than creating a new version.

You can ignore the reverse dependency failure. The linter error can also be ignored as well.

sacerdot:
@haochenx done, thank you!

haochenx (approved):
Thanks! Although there are some CI failures (due to https://github.com/ocaml/opam-repository/issues/24580), I don't think they are related in anyway.

This PR looks good to me.

avsm:
Thanks!

PR #24573, opened 2023-10-04
conf-expat: Add freebsd support

sacerdot:
freebsd should work as well


sacerdot:
All tests pass, but I don't know if the PR actually fixes the problem with ocaml-expat not compiling on freebsd. How to force recompilation of ocaml-expat in a clean way?

haochenx:
Thank you for the PR!

Actually, for conf packages, we'd rather patch the existing version (`packages/conf-expat/conf-expat.1/opam` in this case) rather than creating a new version.

Could you please modify the PR to do that?

sacerdot:
@haochenx done, thank you!

haochenx (approved):
The CI infrastructure seems to have some trouble.. But I believe this PR is a pure improvement (and the CI ran successfully on the previous commit.) Therefore this PR looks perfectly good to me.

avsm:
Thanks!

PR #24586, opened 2023-10-05
Update multicont-1.0.1.tbz

dhil:
This patch updates the hashes of multicont-1.0.1.tbz. For whatever reasons, the archive was not compressed using bzip2, which would cause OPAM to fail installation in CIs.

I have recompressed the archive myself, thus causing the change in the hashes.


avsm:
We don't normally allow changing archives for released packages, but since this is a case of an originally corrupted archive combined with being submitted by the maintainer, we'll let this one go through.

PR #24598, opened 2023-10-07
conf-readline - add support for oraclelinux

haochenx:


tmcgilchrist:
Oracle Linux won't be tested in CI due to https://github.com/ocurrent/opam-repo-ci/pull/237 @haochenx

mseri:
Do we know when it will disappear from the CI? 

In the meantime I am going to merge this: it is not be the only package with ol support, can still be of use if somebody is actually using it and cleans a bit our current CI logs.

haochenx:
Thanks!

>Oracle Linux won't be tested in CI due to https://github.com/ocurrent/opam-repo-ci/pull/237

noted. I will be discerned about Oracle Linux in the future.

PR #24618, opened 2023-10-10
Fix z3 4.8.5 package wrt Python2

cmovcc:
This removes Python2 dependencies for this version of z3, that is still used by FStar, although this is hopefully going to change soon, see https://github.com/FStarLang/FStar/discussions/2974.
Disclaimer: I only tested x86_64-linux.
cc @R1kM @msprotz @tahina-pro


mseri:
Thanks, would you mind releasing this as [packages/z3/z3.4.8.5-1/opam]?

cmovcc:
Sure! I used z3 4.8.8 and 4.8.9 as examples, as both also have an additional -1 package version. That being said, I did not know whether the available flag should be set to false for z3 4.8.5, it is the case for z3 4.8.9 but not z3 4.8.8.

mseri:
I would not make this unavailable since it should work if python 2 is present

mseri:
Thanks!

PR #24633, opened 2023-10-16
new file:   conf-pip-rdkit/conf-pip-rdkit.1.0.0/opam

UnixJunkie:
this is the modern way to install rdkit, on most distros and operating systems

### DIFF
--- a/packages/conf-pip-rdkit/conf-pip-rdkit.1.0.0/opam
+++ b/packages/conf-pip-rdkit/conf-pip-rdkit.1.0.0/opam
@@ -1,0 +1,21 @@
+opam-version: "2.0"
+authors: "https://github.com/rdkit/rdkit/graphs/contributors"
+homepage: "http://www.rdkit.org/"
+maintainer: "unixjunkie@sdf.org"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: "BSD-3-Clause"
+build: ["pip3" "install" "rdkit"]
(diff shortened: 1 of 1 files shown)


UnixJunkie:
the section x-ci-accept-failures in the opam file is left there in case the CI environment starts to annoy us

mseri:
This cannot work, the sandbox does not allow internet access or the modification of external folders :(

UnixJunkie:
Well, this is a valid opam package !

UnixJunkie:
For a very famous dependency in my field...

mseri:
Yes but it does not to what it is meant to be doing. The package should install and check that rdkit is installed. We can skip the installation if it cannot be done, like in this case, since pip would escape opam sandbox and fail, but should have the check that the package is present. So in this form it is not really fulfilling its purpose.

To support pip depext, you could suggest the feature on the ocaml/opam repository, as it was recently done for npm. Once that is in opam, we can let opam install rdkit as well in this conf package.

UnixJunkie:
Ok, I see.
Currently, the important thing for me is to install the depext automatically.
I can add test of the proper install to this package though.

mseri:
Thanks!

Python is such a common requirement that I think there could be value in integrating pip with opam fwiw. Feel free to open an issue about it in the ocaml/opam repository, perhaps mention that something similar is being done for npm and could also be reasonable to think of it broader (including R and rust for example).

mseri:
I am going to close this for now since it cannot work

PR #24642, opened 2023-10-18
omd archive link update

haochenx:
- another take of https://github.com/ocaml/opam-repository/pull/23600 (which seems to be inactive)
- will proceed to replace all versions with available tarballs and use software heritage for the else

### DIFF
--- a/packages/omd/omd.0.5/opam
+++ b/packages/omd/omd.0.5/opam
@@ -26,6 +26,6 @@ markdown features."""
 flags: light-uninstall
 extra-files: ["omd.install" "md5=0395736894a46f718a77f59ec6fbf1fd"]
 url {
-  src: "http://pw374.github.io/distrib/omd/omd-0.5.tar.gz"
-  checksum: "md5=d343deacc270674828260b0fdc55f74c"
+  src: "https://github.com/pw374/omd/archive/refs/tags/0.5.tar.gz"
+  checksum: "md5=64e40582aa879f37c9867f0e10b6268e"
 }
--- a/packages/omd/omd.0.5/opam
+++ b/packages/omd/omd.0.5/opam
@@ -26,6 +26,6 @@ markdown features."""
 flags: light-uninstall
 extra-files: ["omd.install" "md5=0395736894a46f718a77f59ec6fbf1fd"]
 url {
-  src: "https://github.com/pw374/omd/archive/refs/tags/0.5.tar.gz"
-  checksum: "md5=64e40582aa879f37c9867f0e10b6268e"
+  src: "https://github.com/ocaml/omd/archive/refs/tags/0.5.tar.gz"
+  checksum: "sha256=fc462bf7b8f896f924fef610da436d9152168676adca2d606233101eb67e9432"
 }
--- a/packages/omd/omd.1.2.0/opam
+++ b/packages/omd/omd.1.2.0/opam
@@ -34,6 +34,6 @@ with 0.9.x."""
 flags: light-uninstall
 extra-files: ["omd.install" "md5=0395736894a46f718a77f59ec6fbf1fd"]
 url {
-  src: "http://pw374.github.io/distrib/omd/omd-1.2.0.tar.gz"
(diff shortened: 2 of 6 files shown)


packages/omd/omd.0.5/opam
```diff
@@ -26,6 +26,6 @@ markdown features."""
 flags: light-uninstall
 extra-files: ["omd.install" "md5=0395736894a46f718a77f59ec6fbf1fd"]
 url {
-  src: "http://pw374.github.io/distrib/omd/omd-0.5.tar.gz"
-  checksum: "md5=d343deacc270674828260b0fdc55f74c"
+  src: "https://github.com/ocaml/omd/archive/refs/tags/0.5.tar.gz"
+  checksum: "sha256=fc462bf7b8f896f924fef610da436d9152168676adca2d606233101eb67e9432"
```
mseri, line 30:
```suggestion
  checksum: [
    "sha256=fc462bf7b8f896f924fef610da436d9152168676adca2d606233101eb67e9432"
    "md5=d343deacc270674828260b0fdc55f74c"
  ]
```

packages/omd/omd.1.2.0/opam
```diff
@@ -34,6 +34,6 @@ with 0.9.x."""
 flags: light-uninstall
 extra-files: ["omd.install" "md5=0395736894a46f718a77f59ec6fbf1fd"]
 url {
-  src: "http://pw374.github.io/distrib/omd/omd-1.2.0.tar.gz"
-  checksum: "md5=2a1aa5144466a466e33b34de855bcd1c"
+  src: "https://github.com/ocaml/omd/archive/refs/tags/1.2.0.tar.gz"
+  checksum: "sha256=180465a930f27be1bf82bb4709f900cad0d7c69820c0c4efa98573242287637e"
```
mseri, line 38:
```suggestion
  checksum: [
    "sha256=180465a930f27be1bf82bb4709f900cad0d7c69820c0c4efa98573242287637e"
    "md5=2a1aa5144466a466e33b34de855bcd1c"
  ]
```

And the same for all of them. Are the tarballs identical? If now, we should try to find them in the cache and upload them on the opam-source-archives repo

haochenx: I hoped them to be the same, but they are different (at least for the ones that I have checked.) That's why I removed the old checksum.

>If now, we should try to find them in the cache and upload them on the opam-source-archives repo

Do you mean if they are not different, we should cache and upload?

I was not able to find archives that have the same hash. I tried https://github.com/ocaml/omd, https://github.com/pw374/omd, and Software Heritage, and they are all different from the recorded hash

packages/omd/omd.0.5/opam
haochenx: if the archives are not the same, do we still want the md5 hash?

packages/omd/omd.1.2.0/opam
mseri: The cache that I mean is either of

- https://opam.robur.coop/cache/md5/2a/2a1aa5144466a466e33b34de855bcd1c -- this package is not there unfortunately
- https://opam.ocaml.org/cache/md5/2a/2a1aa5144466a466e33b34de855bcd1c -- this package is not there unfortunately

there is a similar URL for the other hashes.

In this case, in the past we marked the package as unavailable since we could not compare the tarball contents with the original ones. I think we should discuss this specific case at the next maintainers meeting

haochenx: Thanks for clarification. Let's discuss this further next week.

haochenx: @mseri do you mind to add it as an item to https://github.com/ocaml/opam-repository/wiki/Meeting-notes? I don't have the permission to modify the wiki page

mseri: Modified in https://github.com/ocaml/opam-repository/wiki/Meetings-agenda

mseri:
This cannot be built as is, one needs to regenerate the setup.ml.
WE will open a new PR to mark them unavailable

PR #24643, opened 2023-10-19
[new release] melange-atdgen-codec-runtime (1.0.0)

jchavarri:
A Melange runtime for atdgen

- Project page: <a href="https://github.com/ahrefs/melange-atdgen-codec-runtime">https://github.com/ahrefs/melange-atdgen-codec-runtime</a>

##### CHANGES:

Initial release.

Migrated to Melange v2.

### DIFF
--- a/packages/melange-atdgen-codec-runtime/melange-atdgen-codec-runtime.1.0.0/opam
+++ b/packages/melange-atdgen-codec-runtime/melange-atdgen-codec-runtime.1.0.0/opam
@@ -1,0 +1,46 @@
+opam-version: "2.0"
+synopsis: "A Melange runtime for atdgen"
+description: """A Melange runtime for atdgen, based on the Js.Json.t type
+provided by Melange and the combinators from melange-json
+"""
+maintainer: "Ahrefs"
+authors: "Ahrefs"
+license: "MIT"
+homepage: "https://github.com/ahrefs/melange-atdgen-codec-runtime"
+bug-reports: "https://github.com/ahrefs/melange-atdgen-codec-runtime/issues"
+depends: [
+  "dune" {>= "3.8"}
+  "ocaml"
+  "melange" {>= "2.0.0"}
+  "atd"
+  "atdgen"
+  "melange-json"
+  "melange-jest" {with-test}
+  "reason" {with-test}
+  "opam-check-npm-deps" {with-test}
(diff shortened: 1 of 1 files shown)


packages/melange-atdgen-codec-runtime/melange-atdgen-codec-runtime.1.0.0/opam
```diff
@@ -0,0 +1,45 @@
...
+provided by Melange and the combinators from melange-json
+"""
+maintainer: "Ahrefs"
+authors: "Ahrefs"
+license: "MIT"
+homepage: "https://github.com/ahrefs/melange-atdgen-codec-runtime"
+bug-reports: "https://github.com/ahrefs/melange-atdgen-codec-runtime/issues"
+depends: [
+  "dune" {>= "3.8"}
+  "ocaml"
```
haochenx, line 13:
```suggestion
  "ocaml" {>= "5.1"}
```
it's encouraged to add lower bound for OCaml. I'm guessing 5.1 since melange.2.0.0 requires it

haochenx (commented):
one suggestion but otherwise look good to me.

packages/melange-atdgen-codec-runtime/melange-atdgen-codec-runtime.1.0.0/opam
jchavarri: Yeah.. I initially had a configuration like this, but later I thought because `melange` will always be the main dependency of these libraries and `melange` already sets a constraint on `ocaml`, better leave the latter floating on libraries?

That way, if we eventually had to change the `ocaml` constraint in `melange`, we wouldn't need to update all packages depending on `melange`. Wdyt?

mseri: I think the requirement is very important for "root" libraries or libraries that fails with the lower bound of their dependencies. I think it is fine to leave it out as long as the library compiles also in lower-bound checks, and we can add bounds on a case by case basis if it starts failing to build after some new release

haochenx: Good. I think we got a policy clarification so please disregard this suggestion.

PR #24648, opened 2023-10-20
Deprecate ocaml-migrate-parsetree

pitag-ha:
Hello :wave:

I'm deprecating ocaml-migrate-parsetree. I think this time, the whole community is behind it. I wrote a [discuss post](https://discuss.ocaml.org/t/rfc-deprecating-ocaml-migrate-parsetree-in-favor-of-ppxlib-also-as-a-platform-tool/13240) about deprecating it and nobody has commented on it.

I've added `flags: deprecated` and a note in the description in all opam packages. If you'd like me to do something else, such as a `post-message`, please let me know.

### DIFF
--- a/packages/ocaml-migrate-parsetree/ocaml-migrate-parsetree.0.6/opam
+++ b/packages/ocaml-migrate-parsetree/ocaml-migrate-parsetree.0.6/opam
@@ -1,4 +1,5 @@
 opam-version: "2.0"
+flags: deprecated
 maintainer: "frederic.bour@lakaban.net"
 authors: [
   "Frédéric Bour <frederic.bour@lakaban.net>"
@@ -33,6 +34,8 @@ conflicts: [
 ]
 synopsis: "Convert OCaml parsetrees between different versions"
 description: """
+Deprecated. Please, use Ppxlib instead.
+
 This library converts parsetrees, outcometree and ast mappers between different OCaml versions.
 High-level functions help making PPX rewriters independent of a compiler version."""
 url {
--- a/packages/ocaml-migrate-parsetree/ocaml-migrate-parsetree.0.6/opam
+++ b/packages/ocaml-migrate-parsetree/ocaml-migrate-parsetree.0.6/opam
@@ -35,6 +35,7 @@ conflicts: [
 synopsis: "Convert OCaml parsetrees between different versions"
 description: """
 Deprecated. Please, use Ppxlib instead.
+More info on https://ocaml.org/changelog/2023-10-23-omp-deprecation
 
 This library converts parsetrees, outcometree and ast mappers between different OCaml versions.
 High-level functions help making PPX rewriters independent of a compiler version."""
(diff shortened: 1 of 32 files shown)


pitag-ha:
My understanding of the lint failure here is: I've modified all OMP packages to mark them as deprecated, including the very old ones. So the lint is getting triggered on all those packages. The 4 year old packages `ocaml-migrate-parsetree.1.0.11` and `ocaml-migrate-parsetree.1.1.0` don't pass the modern lint: They already depended on dune, while the dune-project file was added after the 1.1.0 release. Is that also your understanding of what's happening?

avsm:
I think your understanding is right.  There's a natural deprecation happening here since omp.2.4.0 isnt compatible with OCaml 5.1.0, and never will be.  If we add the deprecation flags for all the old OMPs, then someone compiling an older opam package that depended on OMP will see deprecation warnings there too (which I think we don't intend -- it's an already existing package dependency that works fine, for that package set).

How about just making the latest version of omp as flags: deprecated (but modify all the descriptions as you've done)?  This would mean someone installing the very latest version would get the warning.  I'm also fine with the existing PR too, if you'd prefer just to mark the whole lot as no longer maintained.

pitag-ha:
> then someone compiling an older opam package that depended on OMP will see deprecation warnings there too (which I think we don't intend -- it's an already existing package dependency that works fine, for that package set).

Right, good point! However, one downside of adding the deprecation flag only to OMP 2.4.0 is that the solver will then solve for 2.3.0 if possible, and `opam` will happily go ahead with the installation of OMP 2.3.0 without a deprecation warning. At least, that's my understanding of how `opam` handles deprecation flags in packages. Is that right? (I can find out on Monday if that's right.)

As you've mentioned, OMP will already be implicitly deprecated by being incompatible with new compilers and by having a deprecation note in the package description. So, we have three options here and all have downsides:
1. Keep the deprecation implicit on all OMP versions. Downside: People who've missed the transition to ppxlib, might also be stuck on OCaml 4, and submit new opam packages depending on OMP (the latter has happened to us).
2. Make the deprecation explicit only on OMP 2.4.0. Downside: What I've pointed out above.
3. Make the deprecation explicit on all OMP versions. Downside: What you've pointed out above.

pitag-ha:
My understanding of how `opam` handles the deprecation flag is right: https://opam.ocaml.org/doc/Manual.html#opamflag-deprecated. So only adding a deprecation flag to 2.4.0 would mean that `opam` would install 2.3.0 without a deprecation warning (whenever possible).

So IMO it's better to either keep the deprecation implicit on all OMP versions (my option 1. above) or to make it explicit on all versions (my option 3. above). I personally prefer the latter, since I like making things explicit and since I'd like to avoid that people create new packages depending on OMP. That would mean leaving the PR as is. If either you, @avsm, or any of the other opam-repo maintainers thinks differently, I'm happy to change the PR, though. You all have a better overview of the impact of each option than I do.

pitag-ha:
I've just added the mentioned link, so the PR is ready for review/merging.

avsm:
Looks good to me! Thanks for the comprehensive documentation link on all the packages; that'll really help other users.

PR #24653, opened 2023-10-21
[new release] melange-moment (0.3.0)

jchavarri:
Melange bindings for Moment.js

- Project page: <a href="https://github.com/ahrefs/melange-moment">https://github.com/ahrefs/melange-moment</a>

##### CHANGES:

Migrated to Melange v2.

### DIFF
--- a/packages/melange-moment/melange-moment.0.3.0/opam
+++ b/packages/melange-moment/melange-moment.0.3.0/opam
@@ -25,7 +25,6 @@ build: [
     "-j"
     jobs
     "@install"
-    "@runtest" {with-test}
     "@doc" {with-doc}
   ]
 ]
--- a/packages/melange-moment/melange-moment.0.3.0/opam
+++ b/packages/melange-moment/melange-moment.0.3.0/opam
@@ -1,0 +1,44 @@
+opam-version: "2.0"
+synopsis: "Melange bindings for Moment.js"
+maintainer: "Ahrefs"
+authors: "Jiayu Liu <jimexist@gmail.com>"
+license: "MIT"
+tags: ["melange" "org:ahrefs"]
+homepage: "https://github.com/ahrefs/melange-moment"
+bug-reports: "https://github.com/ahrefs/melange-moment"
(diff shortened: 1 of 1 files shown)


packages/melange-moment/melange-moment.0.3.0/opam
```diff
@@ -0,0 +1,43 @@
...
+tags: ["melange" "org:ahrefs"]
+homepage: "https://github.com/ahrefs/melange-moment"
+bug-reports: "https://github.com/ahrefs/melange-moment"
+depends: [
+  "dune" {>= "3.9"}
+  "ocaml"
+  "melange" {>= "2.0.0"}
+  "reason"
+  "melange-jest" {with-test}
+  "opam-check-npm-deps" {with-test}
```
haochenx, line 15: `with-dev-setup` is used in melange-numeral and reason-react; but `with-test` is used in this package as well as melange-test. Is this intentional? What's the difference?

haochenx (approved):
Besides https://github.com/ocaml/opam-repository/pull/24653#discussion_r1368152635, everything looks good to me.

Since there are prior-art wrt `with-test` on `opam-check-npm-deps`, I'm also marking this PR as mergeable.

packages/melange-moment/melange-moment.0.3.0/opam
jchavarri: I started using `with-dev-setup` as a way to make the packages more compatible with opam 2.2. But over time i have realized it will still take quite some time for that version to be used across the board, so I am rolling back to `with-test` for now.

The difference is that `with-dev-setup` won't install the deps with opam 2.1.

jchavarri: Once opam 2.2 becomes the main version used, I will go through the packages that depend on `opam-check-npm-deps` and update them.

haochenx: Thanks for the clarification. It's clear to me now!

PR #24654, opened 2023-10-21
Use macports npm compatible with active node lts version.

dboris:


haochenx (approved):
CI does not do any relevant tests but looks mergeable to me since
- https://ports.macports.org/port/npm9/details/ is a valid macport "port"
- this should be a net improvement for macport users (except who's on ancient macs and/or macOS versions)

avsm:
Thanks!

PR #24656, opened 2023-10-22
[new release] melange-recharts (4.0.0)

jchavarri:
Melange bindings for recharts

- Project page: <a href="https://github.com/ahrefs/melange-recharts/">https://github.com/ahrefs/melange-recharts/</a>

##### CHANGES:

- Migrate to Melange v2.

### DIFF
--- a/packages/melange-recharts/melange-recharts.4.0.0/opam
+++ b/packages/melange-recharts/melange-recharts.4.0.0/opam
@@ -1,0 +1,50 @@
+opam-version: "2.0"
+synopsis: "Melange bindings for recharts"
+description: "Melange bindings for recharts JavaScript library."
+maintainer: [
+  "Liubomyr Mykhalchenko <liubomyr.mykhalchenko@ahrefs.com>"
+]
+authors: [
+  "Liubomyr Mykhalchenko <liubomyr.mykhalchenko@ahrefs.com>"
+]
+license: "MIT"
+tags: ["melange" "org:ahrefs"]
+homepage: "https://github.com/ahrefs/melange-recharts/"
+bug-reports: "https://github.com/ahrefs/melange-recharts/issues"
+depends: [
+  "dune" {>= "3.8"}
+  "ocaml"
+  "melange" {>= "2.0.0"}
+  "reason" {>= "3.10.0"}
+  "reason-react"
+  "reason-react-ppx"
+  "ocaml-lsp-server" {with-test}
+  "odoc" {with-doc}
+]
+build: [
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
+dev-repo: "git+https://github.com/ahrefs/melange-recharts.git"
+depexts: [
+  ["recharts"] {npm-version = "^2.1.12"}
+]
+url {
+  src:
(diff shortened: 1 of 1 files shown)


haochenx:
CI failures should be caused by the CI infra. I have hit rebuild on all of the failing jobs.

packages/melange-recharts/melange-recharts.4.0.0/opam
```diff
@@ -0,0 +1,50 @@
...
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
+dev-repo: "git+https://github.com/ahrefs/melange-recharts.git"
+depexts: [
+  ["recharts"] {npm-version = "^2.1.12"}
```
haochenx, line 40: it's perhaps missing a dependency to `opam-check-npm-deps`? (if my understanding of including this depexts here is correct)

jchavarri: A package can define depexts to npm packages, but not need to check integrity of npm packages itself.

This is one of those cases: because `melange-recharts` doesn't include a `package.json` nor perform any kind of action / test using npm packages, we can skip the dep on `opam-check-npm-deps.

haochenx: I still haven't grasped the whole picture, but if it's intentional, this PR looks good to me.

(I'm guessing this is the documented [case for binding libraries](https://github.com/ahrefs/opam-check-npm-deps#for-bindings-authors-defining-constraints), but I cannot see how this package is different to say melange-moment and melange-numeral..)

haochenx (approved):
Looks good to me. The CI failures should be unrelated 
image

packages/melange-recharts/melange-recharts.4.0.0/opam
jchavarri: the tl;dr is: if your package has a `package.json` file (for dev purposes), then it should add `opam-check-npm-deps` as dep with `with-dev-setup`. Otherwise, there is no need.

Re: the examples you mentioned:
- `melange-moment` has [a package.json](https://github.com/ahrefs/melange-moment/blob/main/package.json), so in that case it makes sense to have `opam-check-npm-deps` as a dependency, to make sure at dev time that we have the right set of npm packages installed.
- You are right about `melange-numeral`. In that case, there is no `package.json`, so there was no need to have `opam-check-npm-deps`, I guess I got confused or missed it 😕 I can remove it from the opam file of the released version if needed.

In the case of `melange-recharts`, as there is no `package.json`, I didn't include `check-npm-deps`.

Please let me know if this helps clarifying!

jchavarri: Ended up creating https://github.com/ocaml/opam-repository/pull/24664 to update `melange-numeral`.

haochenx: Thank you very much for explaining! It's much clearer for me now! (and sorry for making you create an extra PR 😅)

avsm:

packages/melange-recharts/melange-recharts.4.0.0/opam
jchavarri: np! thanks for the feedback while we figure this out 😄

jchavarri:
Thanks for the suggestion! I think it's a great idea. There are a few more libraries that we want to publish, I think I might wait until all of them are out there to write a single announcement that mentions all of them, in order to avoid too much noise on the forum.

PR #24629, opened 2023-10-13
[new release] flint (4 packages) (0.3.2)

bobot:
Stub of the C library Flint2

- Project page: <a href="https://github.com/bobot/ocaml-flint">https://github.com/bobot/ocaml-flint</a>

##### CHANGES:

  - Bump to dune 3.7 using ctypes stanza 0.3

PR #24658, opened 2023-10-23
[new release] flint (4 packages) (0.3.3)

bobot:
Stub of the C library Flint2

- Project page: <a href="https://github.com/bobot/ocaml-flint">https://github.com/bobot/ocaml-flint</a>

##### CHANGES:

  - Fix library name for archlinux: arb instead of flint-arb


Gbury:
@bobot I opened a PR in your repository against this branch since I couldn't push to it directly: https://github.com/bobot/opam-repository/pull/1

haochenx:
Looking at the current CI result, it's essentially doing exactly the same as the last version (i.e. [flint.0.3.2](https://github.com/ocaml/opam-repository/pull/24629))

[@Gbury's suggestion](https://github.com/ocaml/opam-repository/pull/24658#issuecomment-1775419990) might be a valuable addition since the archlinux CI job is failing, and the changelog for this version states adding archlinux support

haochenx:
Although mergeable as is IMO, it's still nice to consider https://github.com/ocaml/opam-repository/pull/24658#issuecomment-1775419990, so I'll mark this PR as "needs reporter action" for now.

But I want to be clear that there is no pressure to adopt https://github.com/bobot/opam-repository/pull/1 for this PR to be merged in my opionion.

Gbury:
> [@Gbury's suggestion](https://github.com/ocaml/opam-repository/pull/24658#issuecomment-1775419990) might be a valuable addition since the archlinux CI job is failing, and the changelog for this version states adding archlinux support

Note that even with my PR, only `flint`, `arb` and `antic` (and the associated `conf-*`) packages would likely build on the CI, given that the depext for `calcium` is in the AUR (see https://github.com/ocaml/opam/issues/5709 for the related issue in opam). However, on a personal machine, it should be easy for users to make it work (by manually installing calcium using one's favorite AUR helper).

haochenx:
Thanks for the details! Let's see how @bobot reacts. (I can neither modify this PR directly..)

Meanwhile, I will mark this PR as mergeable, as we can always add those depext lines later.

packages/conf-calcium/conf-calcium.1/opam
```diff
@@ -30,6 +30,12 @@ depexts: [
   ["calcium" "calcium-devel"] {os-distribution = "ol"}
   ["calcium-devel"] {os-family = "suse"}
   ["calcium"] {os-distribution = "nixos"}
+  ["calcium"] {os-distribution = "arch" }
```
haochenx, line 33: not sure this helps if `calcium` cannot be install by pacman, but it shouldn't hurt.

haochenx:
Thanks! The CI seem to haze succeeded for archlinux

packages/conf-calcium/conf-calcium.1/opam
Gbury: If/when opam can configure the package manager, then that will work. In the meantime, I think users can already alias `pacman` to their pacman wrapper of choice to make that work (but I'll have to check once this PR is merged).

Gbury:
Actually, wait a bit, the `arb` and `antic` packages seem to have disappeared from the archlinux package database overnight...

haochenx:
>Actually, wait a bit, the arb and antic packages seem to have disappeared from the archlinux package database overnight...

https://archlinux.org/packages/extra/x86_64/flint/

It looks like the package is updated on 10/20 and states to replace antic, arb.

This is still a net improvement, so I don't think it's a blocker for this PR.

Gbury:
Alright, so, this actually does change things quite a bit: flint [just released](https://github.com/flintlib/flint/releases/tag/v3.0.0) a new major version (version 3) a few days ago, version 3 (and true to its rolling release nature, archlinux packaged it already) and it happens to actually directly integrate `arb`, `antic` and `calcium`, which were previously developed as separate libraries (see https://github.com/flintlib/flint#overview ). In particular this will require another change to the build system for archlinux (since the `arb`, `antic` and `calcium` C library no longer exists, and thus the c flags need to be adjusted), but I hope that it should be reasonable to fix, and I'm working on that.

So in summary, I think we should wait on this PR until I fix that situation because as it stands, the build of the `0.3.3` packages are actually broken on archlinux. The fixes will probably require either a new version `0.3.3` or a `0.3.4` (depending on @bobot 's preference). Luckily, all of this is a result of a rather rare coincidence in terms of releases and packages, and it should help prepare for when other distributions start packaging version `3` of `flint`.

haochenx:
I see. In that case, I’m marking this package (at least temporary) as do-not-merge. This PR can be reused or closed in favor of a new one once the necessary changes has been performed upstream.

samoht:
What should we do with this PR? Seems that @filipeom fixed a few things on archlinux since then.

PR #24664, opened 2023-10-23
[update] melange-numeral, remove unused dep

jchavarri:
Turns out this package is not using this dep, thanks @haochenx for pointing it out in https://github.com/ocaml/opam-repository/pull/24656#discussion_r1368995319.


haochenx (approved):
Thanks! Looks good to me!

PR #24666, opened 2023-10-24
Release three Ortac packages 0.1.0

n-osborne:
The initial opam release of `ortac-core`, `ortac-runtime` and `ortac-qcheck-stm`.

Co-authored-by: Samuel Hym <samuel.hym@rustyne.lautre.net>

### DIFF
--- a/packages/ortac-core/ortac-core.0.1.0/opam
+++ b/packages/ortac-core/ortac-core.0.1.0/opam
@@ -1,0 +1,56 @@
+opam-version: "2.0"
+synopsis:
+  "Ortac (OCaml Runtime Assertion Checking) core tool and library based on Gospel"
+description: """\
+Ortac (OCaml Runtime Assertion Checking) is a tool to turn
+executable Gospel specifications into code to test they hold.
+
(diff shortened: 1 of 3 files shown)


raphael-proust:
The CI errors are unrelated issues with afl-persisten installation failing on some experimental backends.

shym:
For my education, is it possible to work around the problem with `afl-persistent`? Because I see no build failure on the `monolith` package (which is the dependency bringing `afl-persistent` along) released in June, even on macos.

shym:
Thank you for the digging!
What I find weird is that is compiled just fine last Spring. My hypothesis then is:
- that `sh` on macos is creating a temporary file in the current directory for here documents,
- that `/` is now (soundly) unwritable in CI.
If that’s the case, changing the script so that it runs fine on macos too is easy, but this would require a proper diagnostic.

raphael-proust:
I opened an issue upstream: https://github.com/stedolan/ocaml-afl-persistent/issues/13
I think we can merge this PR anyway

mseri:
Thanks

PR #24681, opened 2023-10-26
Mark conflict in zarith_js_stubs

raphael-proust:
Calling the extract function makes a runtime failure: `[failure] ml_z_extract_small not implemented`

See also https://github.com/janestreet/zarith_stubs_js/issues/10

Note that the error is only triggered when calling the missing function, so it's not necessarily a strict conflict, but there is precedent in the other zarith version for similar issues.

### DIFF
--- a/packages/zarith_stubs_js/zarith_stubs_js.v0.15.0/opam
+++ b/packages/zarith_stubs_js/zarith_stubs_js.v0.15.0/opam
@@ -14,7 +14,7 @@ depends: [
   "dune"  {>= "2.0.0"}
 ]
 # The conflict is a run-time failure when versions do not match:
-conflicts: "zarith" {< "1.12"}
+conflicts: [ "zarith" {< "1.12" | >= "1.13" } ]
 synopsis: "Javascripts stubs for the Zarith library"
 description: "
 This library contains no ocaml code, but instead implements
--- a/packages/zarith_stubs_js/zarith_stubs_js.v0.15.0/opam
+++ b/packages/zarith_stubs_js/zarith_stubs_js.v0.15.0/opam
@@ -15,7 +15,7 @@ depends: [
 ]
 # The conflict is a run-time failure when versions do not match:
 # There are two separate failures for <1.12 and >1.12
-conflicts: [ "zarith" {!= "1.12" } ]
+conflicts: [ "zarith" {< "1.12" | >= "1.13" } ]
 synopsis: "Javascripts stubs for the Zarith library"
 description: "
 This library contains no ocaml code, but instead implements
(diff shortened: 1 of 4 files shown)


packages/zarith_stubs_js/zarith_stubs_js.v0.15.0/opam
```diff
@@ -14,7 +14,7 @@ depends: [
   "dune"  {>= "2.0.0"}
 ]
 # The conflict is a run-time failure when versions do not match:
-conflicts: "zarith" {< "1.12"}
+conflicts: [ "zarith" {< "1.12" | >= "1.13" } ]
```
mseri, line 17:
```suggestion
conflicts: [ "zarith" {< "1.12" | >= "1.13" } ]
```
Is there any compatible zarith version in this case?

raphael-proust: I guess zarith.1.12 is the only version compatible with that release of zarith-stubs-js. But I'm a bit tired so I don't really trust myself on having gotten the comparison and operators right.

Btw, the previous constraint is from https://github.com/ocaml/opam-repository/pull/21056

mseri: I was sleeping, I saw equal signs on both sides :)

mseri:
```suggestion
conflicts: [ "zarith" {!= "1.12" } ]
```

raphael-proust: I pushed essentially this but added a comment too

raphael-proust:
Well it seems there are more conflicts because there's a `# Fatal error: exception Failure("ml_z_mul_overflows not implemented")` with zarith.1.12 and zarith_stubs_js.0.12.0. But it might actually be that 0.12.0 is not actually copletely compatible with any of the zarith version? I'm not sure. I'll look into it  bit more later.

packages/zarith_stubs_js/zarith_stubs_js.v0.15.0/opam
avsm: The original < and >= was a better constraint, since it allows for any variants of zarith.1.12 to also match (e.g. `zarith.1.12+dune`, that sort of thing)

raphael-proust:
See also https://github.com/ocaml/opam-repository/pull/24755

packages/zarith_stubs_js/zarith_stubs_js.v0.15.0/opam
raphael-proust: Reverted to `<` and `>=`

mseri:
Thanks a lot

PR #24689, opened 2023-10-27
Added cloudi.2.0.7

okeuday:


okeuday:
Only `macos-homebrew-ocaml-5.1-arm64 (experimental)` and `macos-homebrew-ocaml-4.14-arm64 (experimental)` failed due to not having GMPlib for zarith to use.

mseri:
Strange, gmp should be there. I think homebrew uses a different path on arm, maybe the build script is looking in the wrong place

okeuday:
@mseri The failure occurs in the `zarith-1.13` release that is included in the repository, for compilation in isolation.  So, the GMPlib search fix should occur as part of `zarith` (its build system is custom and isn't yet attempting to use autoconf/automake or something similar to avoid problems).

mseri:
Thanks!

mseri:
Perhaps I was too fast here at merging. What is the point of requiring zarith but also vendoring it?

okeuday:
@mseri The approach is intentional to catch problems early and improve the in-place compilation in the main repository (https://github.com/CloudI/CloudI/tree/develop/src/api/ocaml) as necessary.  Using this approach, there is always the assurance that the build of the OCaml CloudI API is the same as in the main repository.

packages/cloudi/cloudi.2.0.7/opam
```diff
@@ -0,0 +1,21 @@
...
+bug-reports: "https://github.com/CloudI/cloudi_api_ocaml/issues"
+license: "MIT"
+dev-repo: "git+https://github.com/CloudI/cloudi_api_ocaml.git"
+build: [
+  [make]
+]
+depends: [
+  "ocaml" {>= "4.03.0"}
+  "base-unix"
+  "zarith" {>= "1.13"}
```
mseri, line 14: In this case you don't really need this line

okeuday: While it might not be strictly necessary, I wanted to be transparent about the dependency on zarith, so it appears in the cloudi package information.

mseri: Thanks for the clarification

PR #24692, opened 2023-10-27
Frama-C: new release (28.0~beta-Nickel)

AllanBlanchard:
On behalf of @maroneze 

Main changes with respect to Frama-C 27 (Cobalt) include:

#### Kernel
- Frama-C can now generate more default clauses (in particular terminates and exits)
- Removed deprecated options `-no-type` and `-no-obj`

#### Alias
- New plugin Alias, implements a points-to analysis

#### E-ACSL
- More efficient code arithmetic calculations

#### Eva
- Support for simple `\let` bindings
- Removed deprecated Db.Value API
- Fixed unsoundness about initialization with goto and switch

#### WP
- New ACSL extensions for defining automatic proof strategies
- WP generates default exits and terminates, removed old options `-wp-*-terminate`
- Fixed cache for interactive provers

#### Ivette
- Basic component for WP
- Many bug fixes

### DIFF
--- a/packages/frama-c/frama-c.28.0~beta/opam
+++ b/packages/frama-c/frama-c.28.0~beta/opam
@@ -1,0 +1,187 @@
+opam-version: "2.0"
+available: opam-version >= "2.1.0"
+flags: avoid-version
+synopsis: "Platform dedicated to the analysis of source code written in C"
+description:"""
+Frama-C gathers several analysis techniques in a single collaborative
+framework, based on analyzers (called "plug-ins") that can build upon the
+results computed by other analyzers in the framework.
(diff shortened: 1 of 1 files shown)


maroneze:
We are going to take a look into why FreeBSD jobs failed, but we don't believe this should prevent this beta version from being published (making another beta release for that seems overkill...).

mseri:
Thanks for looking into it. I am going to merge this version in the meantime

maroneze:
By the way, is there some way to run the FreeBSD tests without resorting to creating a pull request? I could create a fake one with some modified code to do the testing, but if there already is some infrastructure in place to help with this kind of testing, it would be good to know about it.

PR #24682, opened 2023-10-26
conf-glade support on Linux Mint, Arch, Alpine, OpenSuse, and FreeBSD

jmid:
This little PR adds `conf-glade` support on Linux Mint and FreeBSD

I can see support is also missing for Arch, OpenSuse, Oracle, and macOS and expect the CI to fail on those as a result.

PR #24693, opened 2023-10-27
Add ocaml upper bounds for lablgtk.2.18.11-2.18.13

jmid:
lablgtk.2.18.11-2.18.13 use `String.create` which was first deprecated and then removed in OCaml 5.
This causes errors (as seen in #24682):
```
# ocamlc.opt -I +unix  -w s-3-6+52 -c  lablgladecc.ml
# File "lablgladecc.ml", line 355, characters 16-29:
# 355 |       let buf = String.create 1024 in
#                       ^^^^^^^^^^^^^
# Error: Unbound value String.create
```

The PR therefore adds upper bounds for lablgtk.2.18.11-2.18.13.
For earlier versions (lower) upper bounds are already present.

### DIFF
--- a/packages/lablgtk/lablgtk.2.18.11/opam
+++ b/packages/lablgtk/lablgtk.2.18.11/opam
@@ -13,7 +13,7 @@ install: [
   [make "install"]
 ]
 depends: [
-  "ocaml" {>= "4.05"}
+  "ocaml" {>= "4.05" & < "5.0.0"}
   "ocamlfind" {>= "1.2.1"}
   "conf-gtk2" {build}
 ]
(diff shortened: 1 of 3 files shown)


jmid:
There was a number of internal CI failures for which I've triggered a rerun.

#24700 adds FreeBSD support to `conf-gtk2` which should take of the FreeBSD failures... :crossed_fingers:

mseri:
Thanks

garrigue:
A heads-up would have been welcome: lablgtk-2.18.13 was explicitly intended as a compatibility release for 5.0...
I suppose this was missed because the problem is only triggered if conf-glade is installed (otherwise the incriminated file is not compiled).

mseri:
Fair point, sorry. I'll see if we can modify the bounds adding a conflict for 2.18.13 instead

PR #24694, opened 2023-10-27
`conf-libgccjit`: external dependency for bindings to gccjit

lukstafi:
For upcoming https://github.com/lukstafi/ocaml-gccjit


lukstafi:
I'd appreciate help on the failing CI test, of which there should hopefully be few now.

lukstafi:
I'm confused by this fail: https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/909e58c952d7ce91226c5798cd55368b1dafbcf8/variant/distributions,ubuntu-23.04-ocaml-4.14,conf-libgccjit.13
I have this in conf-libgccjit.13, which I thought excludes v13 from running on "23.04":
```
available: [
  (os-family = "debian" & (os-version > "12")) |
  (os-family = "ubuntu" & (os-version >= "23.10")) |
  (os-family = "fedora" & (os-version >= "38")) |
  (os-family = "ol" & (os-version > "8")) |
  (os-family != "debian" & os-family != "ubuntu" & os-family != "fedora" &
   os-family != "ol" & os != "macos" & os-family != "freebsd")
]
```

packages/conf-libgccjit/conf-libgccjit.12/opam
```diff
@@ -0,0 +1,61 @@
...
+  ["sh" "-exc" "gcc -c $CFLAGS -I/usr/local/include test.c"] {os != "macos"}
+  [
+    "sh"
+    "-exc"
+    "cc -c $CFLAGS -I/opt/homebrew/include -I/opt/local/include -I/usr/local/include test.c"
+  ] {os = "macos"}
+]
+depends: [ "conf-gcc" ]
+# Check that the default version of gcc is not too high.
+available: [
```
mseri, line 24: I am in favour of removing this and make the package fail in ci for those cases. Users can install the library manually, e.g. from sources, and use it also on those systems. Setting availability would prevent them from doing it

packages/conf-libgccjit/conf-libgccjit.13/opam
```diff
@@ -0,0 +1,57 @@
...
+  ["sh" "-exc" "echo \"#include  \\\"libgccjit.h\\\"\" > test.c"]
+  ["sh" "-exc" "gcc -c $CFLAGS -I/usr/local/include test.c"] {os != "macos"}
+  [
+    "sh"
+    "-exc"
+    "cc -c $CFLAGS -I/opt/homebrew/include -I/opt/local/include -I/usr/local/include test.c"
+  ] {os = "macos"}
+]
+depends: [ "conf-gcc" ]
+available: [
```
mseri, line 23: As in the other comment, I would remove this

jmid:
Drive-by-comment: 
On Ubuntu-derivatives such as Linux Mint `opam var`s differ a bit:
```
$ opam var os-distribution
linuxmint
$ opam var os-family
ubuntu
```
(these are not tested on the opam-repo CI though)

If I understand your rules correctly, `gcc-12` (or `gcc-13`) would not be required on such a system.
For `libgccjit-12-dev` in apt-cache I however see
```
Depends: gcc-12-base (= 12.3.0-1ubuntu1~22.04), libgccjit0 (>= 12.3.0-1ubuntu1~22.04)
```
I thus wonder if `gcc-12-base` is sufficient?

packages/conf-libgccjit/conf-libgccjit.12/opam
```diff
@@ -0,0 +1,61 @@
...
+  (os-family = "freebsd") |
+  (os-family = "ol" & (os-version > "8")) |
+  (os-family != "debian" & os-family != "ubuntu" & os-family != "fedora" &
+   os-family != "archlinux" & os-family != "ol" & os != "macos" & os-family != "freebsd")
+]
+# By depending on the specific version of gcc below, we might upgrade
+# the default gcc binary, but we will not downgrade it.
+depexts: [
+  ["gcc-12"] {os-distribution = "centos" | os-distribution = "fedora"
+             | os-distribution = "debian" | os-distribution = "ubuntu"
```
mseri, line 38:
```suggestion
             | os-family = "debian" | os-family = "ubuntu"
```
I would go with the family and make it broader

lukstafi:
> If I understand your rules correctly, `gcc-12` (or `gcc-13`) would not be required on such a system. For `libgccjit-12-dev` in apt-cache I however see
> 
> ```
> Depends: gcc-12-base (= 12.3.0-1ubuntu1~22.04), libgccjit0 (>= 12.3.0-1ubuntu1~22.04)
> ```
> 
> I thus wonder if `gcc-12-base` is sufficient?

If `libgccjit-12-dev` already has the dependency, then `gcc-12-base` depexts would be redundant. But it seemed to me installing `gcc-12/13` updates the default `gcc` binary (symlink), while installing `gcc-12/13-base` would not since installing `libgccjit-12/13-dev` didn't. But I am no longer sure if I properly confirmed it.

packages/conf-libgccjit/conf-libgccjit.12/opam
```diff
@@ -0,0 +1,45 @@
...
+# Check that the default version of gcc is not too high.
+# By depending on the specific version of gcc below, we might upgrade
+# the default gcc binary, but we will not downgrade it.
+depexts: [
+  ["libgccjit-12-dev"] {os-family = "debian"}
+  ["libgccjit-12-dev"] {os-family = "ubuntu"}
+  ["gcc-toolset-12-libgccjit-devel"] {os-distribution = "centos"}
+  ["libgccjit-devel.12"] {os-distribution = "fedora"}
+  ["gcc-toolset-12-libgccjit-devel"] {os-distribution = "ol" | os-distribution = "oraclelinux"}
+             | os-family = "debian" | os-family = "ubuntu"
```
jmid, line 32: This line breaks the `depexts` format as also reported on the CI. I'm guessing it is just a copy-paste mistake.
I recommend locally running `opam lint packages/conf-libgccjit/conf-libgccjit.12/opam` to confirm that things are OK (and shamefully admits that I sometimes forget to do so myself... :sweat_smile:)

jmid:
> If `libgccjit-12-dev` already has the dependency, then `gcc-12-base` depexts would be redundant. But it seemed to me installing `gcc-12/13` updates the default `gcc` binary (symlink), while installing `gcc-12/13-base` would not since installing `libgccjit-12/13-dev` didn't. But I am no longer sure if I properly confirmed it.

You may very well be right, I'm no expert on the gcc deb-packages!
I just spotted the dependency and wanted to mention it.

packages/conf-libgccjit/conf-libgccjit.12/opam
lukstafi: This is an artifact of an accepted suggestion, I don't see it in my repository...

lukstafi:
- The failures of this form: `[conf-libgccjit.12 (failed: The compilation of conf-libgccjit.12 failed at "sh -exc gcc --version | grep 12".)](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/3b1e0a2b014be37f1b41a9160321c6f70547f0b2/variant/distributions,archlinux-ocaml-4.14,conf-libgccjit.12) -- these are acceptable. We installed a downgraded version of gcc which is not the default, (or even an upgraded version if this is Ubuntu LTS?).
  - Maybe with some exceptions related to shell compatibility?
- The failures of this form: [conf-libgccjit.12 (failed: System package install failed with exit code 1 at command:)](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/3b1e0a2b014be37f1b41a9160321c6f70547f0b2/variant/distributions,fedora-37-ocaml-4.14,conf-libgccjit.12) -- most likely I don't know the name of the package on the system.
- The failures of this form: [conf-libgccjit.12 (failed: The compilation of conf-libgccjit.12 failed at "sh -exc gcc -c $CFLAGS -I/usr/local/include test.c".)](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/3b1e0a2b014be37f1b41a9160321c6f70547f0b2/variant/distributions,alpine-3.18-ocaml-4.14,conf-libgccjit.12) -- I would need to debug on the system, probably the include paths are insufficient / different.
-

lukstafi:
Even better, some distros are easy to install on Windows via `wsl install -d`.
- Debian 12: works great, installs `conf-libgccjit.12`
- OpenSUSE Tumbleweed @ gcc version 13: installing OCaml via opam fails. Will need to revisit.
- OpenSUSE 15.5: @gcc version 7, I don't intend to support it.
- OracleLinux_9: @gcc version 11, adding `conf-libgccjit.11` to support it. But there is a complication: while installing the versioned package works with `yum`, it fails the check `rpm -q --whatprovides`.

lukstafi:
I fixed and verified OpenSUSE (within WSL). The only distributions I'm giving up on: `Oracle Linux 8` (too outdated) and `Alpine` (under WSL, I get: `configure: error: C compiler cannot create executables` when installing `ocaml-base-compiler`). The only distribution still not debugged: `MacOS` (I will have access to a Mac in two weeks). In all other cases one version of `conf-libgccjit` passes.

lukstafi:
Summary of CI results -- configurations where no `conf-libgccjit` version passes:

1. alpine-3.18 (tricky to debug)
2. debian-10, debian-11 **too old**
3. opensuse-15.5 **too old**
4. oraclelinux-8  **too old**
5. ubuntu-20.04  **too old**
6. macos-homebrew-ocaml-4.14-arm64 (but macos-homebrew-ocaml-4.14-amd64 passes)
7. macos-homebrew-ocaml-5.1-amd64 (but macos-homebrew-ocaml-5.1-arm64 passes)

Both _6_ and _7_ fail due to the same problem, missing `_stdio.h` included from `stdio.h`:
```
# + gcc-13 -c -I/opt/homebrew/include -I/opt/local/include -I/usr/local/include test.c
# In file included from /opt/homebrew/include/libgccjit.h:23,
#                  from test.c:1:
# /opt/homebrew/Cellar/gcc/13.2.0/lib/gcc/current/gcc/aarch64-apple-darwin23/13/include-fixed/stdio.h:78:10: fatal error: _stdio.h: No such file or directory
#    78 | #include <_stdio.h>
#       |          ^~~~~~~~~~
```

lukstafi:
Since quite a lot of non-Debian distributions don't require an explicitly versioned `libgccjit-N-dev` package, maybe we should instead make `conf-libgccjit` un-versioned, and require the latest or the appropriate version in the specific distros with explicit versioning? Yes, I'll migrate the code in that direction.

lukstafi:
Ready for review now.
I excluded OracleLinux <= 8 because it's unclear how to install libgccjit corresponding to the system version of gcc on them.
On MacOS, I install a specific version of gcc to future-proof the test in the build rule.

Note: the MacOS failures are flaky. E.g. [in this run](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/0ed65bce16c770e18634c2b2b584d1523e257872) there are 3 successes and 1 failure.

lukstafi:
@mseri @jmid friendly ping -- now also supporting Linux Mint.

mseri:
Thanks for the submission, I tried to reproduce the failure locally on mac and have your same experience. I think this can be merged

packages/conf-libgccjit/conf-libgccjit.1/opam
```diff
@@ -0,0 +1,64 @@
...
+    "David Malcolm <dmalcolm@redhat.com>"
+]
+homepage: "https://gcc.gnu.org/wiki/JIT"
+license: "GPL-2.0-or-later"
+# On Debian systems, there is libgccjit0, but not libgccjit0-dev, so we need
+# to explicitly match the libgccjit-dev version to the gcc version.
+# On MacOS, both cc and gcc bind to the clang compiler, so the gccjit bindings
+# might not work out-of-the-box and we use a specific version of gcc below.
+available: [
+  ((os-distribution != "ol" & os-distribution != "oraclelinux") | os-version >= "9")
```
mseri, line 13: We can move this later to an x-ci-accept-failure if anybody requires it on ol

lukstafi: I would not exclude a whole distribution, if anyone still uses OL 8 they have the option to update. This case specifically is about old versions of OL which are misconfigured: the default gcc is older than the oldest available package for libgccjit.

mseri: Yes, sorry I was too sloppy in my comment. I mean if somebody intends to use it in older ol, we can always keep it available and and let them compile their own gcc. In any case I don't think we have ol users at the moment

mseri: Yes, sorry I was too sloppy in my comment. I mean if somebody intends to use it in older ol, we can always keep it available and and let them compile their own gcc. In any case I don't think we have ol users at the moment

PR #24680, opened 2023-10-26
Add minivpt lower bounds

jmid:
In #24677 there are lower bound failures in lower bound dependency... :sweat_smile: 
This causes:
```
#=== ERROR while compiling minivpt.2.0.0 ======================================#
# context              2.2.0~alpha3~dev | linux/x86_64 | ocaml-base-compiler.4.03.0 | file:///home/opam/opam-repository
# path                 ~/.opam/4.03/.opam-switch/build/minivpt.2.0.0
# command              ~/.opam/opam-init/hooks/sandbox.sh build obuild build lib-minivpt
# exit-code            2
# env-file             ~/.opam/log/minivpt-7-cd9550.env
# output-file          ~/.opam/log/minivpt-7-cd9550.out
### output ###
# uncaught exception
# Fatal error: exception Failure("unknown option: lib-minivpt")
```

After having checked locally, this PR therefore adds a lower bound requirement of `obuild.0.0.3`

PR #24677, opened 2023-10-26
Patch lbvs_consent 2.1.1-2.2.0 to support openbabel-3

jmid:
The previous PR #24545 patched `conf-openbabel` to install openbabel-2 or 3.

The `lbvs_consent` however still hardcodes an openbabel-2 include path which causes CI build errors, and thus adds noise:
```
lbvs_consent.2.1.1 (failed: The compilation of lbvs_consent.2.1.1 failed at "make obmaccs".)
lbvs_consent.2.1.2 (failed: The compilation of lbvs_consent.2.1.2 failed at "make obmaccs".)
lbvs_consent.2.1.3 (failed: The compilation of lbvs_consent.2.1.3 failed at "make obmaccs".)
lbvs_consent.2.2.0 (failed: The compilation of lbvs_consent.2.2.0 failed at "make obmaccs".)
```

This follow-up PR thus patches the Makefiles for `lbvs_consent` 2.1.1-2.2.0 to utilize #24545.

I've filed an upstream PR with the corresponding change here: https://github.com/UnixJunkie/consent/pull/15

PR #24696, opened 2023-10-27
Bump minivpt lower bound further

jmid:
`minivpt` is a dependency in #24677 which is now causing `lbvs_consent` compilation to fail with:
```
[ERROR] The compilation of lbvs_consent.2.1.1 failed at "dune build -p lbvs_consent -j 255".

#=== ERROR while compiling lbvs_consent.2.1.1 =================================#
# context              2.2.0~alpha3~dev | linux/x86_64 | ocaml-base-compiler.4.03.0 | pinned(https://github.com/UnixJunkie/consent/archive/v2.1.1.tar.gz)
# path                 ~/.opam/4.03/.opam-switch/build/lbvs_consent.2.1.1
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p lbvs_consent -j 255
# exit-code            1
# env-file             ~/.opam/log/lbvs_consent-7-697f60.env
# output-file          ~/.opam/log/lbvs_consent-7-697f60.out
### output ###
# File "src/dune", line 6, characters 49-56:
# 6 |   (libraries dolog batteries bitv parmap str zip minivpt cpm))
#                                                      ^^^^^^^
# Error: Library "minivpt" not found.
```

I believe this is an issue with the `obuild` version, not making the name `minivpt` visible to `dune`.
Locally I can reproduce the issue - and bumping the `obuild` requirement to 0.1.0 solved it.
The change should benefit other packages that use `dune` and would want to depend on `minivpt`.

For context, #24680 added the first `obuild` lower bound.

While I was at it, I removed a period in the synopsis to make the opam files lint cleanly.

### DIFF
--- a/packages/minivpt/minivpt.1.0.0/opam
+++ b/packages/minivpt/minivpt.1.0.0/opam
@@ -18,7 +18,7 @@ remove: [
 depends: [
   "ocaml"
   "ocamlfind"
-  "obuild" {build & >= "0.0.3"}
+  "obuild" {build & >= "0.1.0"}
 ]
 synopsis: "Minimalist vantage point tree implementation in OCaml"
 description: """
--- a/packages/minivpt/minivpt.1.0.0/opam
+++ b/packages/minivpt/minivpt.1.0.0/opam
@@ -20,7 +20,7 @@ depends: [
   "ocamlfind"
   "obuild" {build & >= "0.0.3"}
 ]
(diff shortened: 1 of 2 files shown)


jmid:
The only two CI failures are from `lbvs_consent` which the ongoing #24677 is trying to address.

mseri:
Thanks

PR #24697, opened 2023-10-27
Update conf-lame opensuse depext

artur-shaik:
According to this page: [https://software.opensuse.org/package/libmp3lame-devel](https://software.opensuse.org/package/libmp3lame-devel) and my experience with installing lame with opam, development package in opensuse should be: `libmp3lame-devel`


mseri:
Thanks

PR #24702, opened 2023-10-27
Add Linux Mint support to conf-pkg-config

jmid:
This little PR adds Linux Mint support to `conf-pkg-config`


jmid:
This one-line change makes a difference for me as a Linux Mint opam user.
I believe that none of the reported 576 failed CI jobs are caused by the change.
Being a QCheck maintainer, I put in a bit of work in #24703 to address the QCheck failures I spotted in the list...

mseri:
Thanks

PR #24705, opened 2023-10-27
Add Mint support to conf-autoconf

jmid:
This little PR adds Linux Mint support to `conf-autoconf`


jmid:
Again, extending an existing `depext` entry to work on Mint too does not cause any of the reported 185 failures, but does make a difference for Linux Mint `opam` users.

mseri:
Thanks

PR #24708, opened 2023-10-27
Add FreeBSD conf-rdkit support

jmid:
This PR adds conf-rdkit support for FreeBSD.

I'm unsure of the right include dir and compile flags on FreeBSD and so mimic the Linux ones for a start.

### DIFF
--- a/packages/conf-rdkit/conf-rdkit.1/opam
+++ b/packages/conf-rdkit/conf-rdkit.1/opam
@@ -27,6 +27,7 @@ x-ci-accept-failures: [ # RDKit does not exist by default on these distributions
   "opensuse-15.2"
   "oraclelinux-7"
   "oraclelinux-8"
+  "oraclelinux-9"
 ]
 post-messages:
     "This package requires rdkit to be installed. You might want to try:
--- a/packages/conf-rdkit/conf-rdkit.1/opam
+++ b/packages/conf-rdkit/conf-rdkit.1/opam
@@ -25,6 +25,7 @@ x-ci-accept-failures: [ # RDKit does not exist by default on these distributions
   "archlinux"
   "centos-7"
   "centos-8"
+  "fedora-37"
   "opensuse-15.2"
   "oraclelinux-7"
   "oraclelinux-8"
--- a/packages/conf-rdkit/conf-rdkit.1/opam
+++ b/packages/conf-rdkit/conf-rdkit.1/opam
@@ -5,8 +5,8 @@ maintainer: "unixjunkie@sdf.org"
 bug-reports: "https://github.com/ocaml/opam-repository/issues"
 license: "BSD-3-Clause"
 build: [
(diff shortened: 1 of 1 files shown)


packages/conf-rdkit/conf-rdkit.1/opam
```diff
@@ -6,7 +6,7 @@ bug-reports: "https://github.com/ocaml/opam-repository/issues"
 license: "BSD-3-Clause"
 build: [
   ["sh" "-c" "c++ test.cpp -o test -I/usr/local/include/rdkit -L/usr/local/lib -lRDKitRDGeneral -DSWIG"] {os = "macos"}
-  ["sh" "-c" "c++ test.cpp -o test -I/usr/include/rdkit -lRDKitRDGeneral -DSWIG || c++ test.cpp -o test -I/usr/include/rdkit -lRDGeneral -DSWIG"] {os = "linux"}
+  ["sh" "-c" "c++ test.cpp -o test -I/usr/include/rdkit -lRDKitRDGeneral -DSWIG || c++ test.cpp -o test -I/usr/include/rdkit -lRDGeneral -DSWIG"] {os = "linux" | os = "freebsd"}
```
hannesm, line 9: it's more likely you'll need to go the macos path... headers in /usr/local/include/rdkit; libraries in /usr/local/lib:

`-I/usr/local/include/rdkit -L/usr/local/lib`

jmid: Thanks Hannes! Second attempt in a17dc33...

jmid:
After a bit of trial and error, a combination of 2 include paths make things work on FreeBSD.
e493083 and c5e704c adds acceptable CI failure versions for Alpine and Oracle.
Finally 630ee40 attempt the extra include dir on macOS which is currently failing.

jmid:
Cleaned up the history a bit after realizing that include dirs won't change anything on macOS unless some depext is listed.
I then marked the tested Fedora and OpenSuse version as acceptable CI failures too.

The remaining failures are:
- lbvs_consent - should be handled by #24677
- OpenSuse Tumbleweed - not sure how to mark it in `x-ci-accept-failures`
- macOS - for which no depext is listed

jmid:
The latest commit also marks `fedora-38` an acceptable CI failure.
Despite these annotations the Fedora runs are still marked as failures.
I suspect that this is because the system package installation fails (pkg not available on these newer Fedoras).

jmid:
The above summary should still be accurate.

For context, #24633 explains that the modern way to install `rdkit` is through `pip` for which the following `opam` issue has been opened: https://github.com/ocaml/opam/issues/5703

Adding FreeBSD support and adding a few `x-ci-accept-failures` is a low-effort approach to get a few more CI green lights until the above is supported.

mseri:
This is a strict improvement over what we had. Thanks!

PR #24716, opened 2023-10-30
Package abella.2.0.8

chaudhuri:
### `abella.2.0.8`
Interactive theorem prover based on lambda-tree syntax

---
* Homepage: https://abella-prover.org
* Source repo: git+https://github.com/abella-prover/abella.git
* Bug tracker: https://github.com/abella-prover/abella/issues

---


chaudhuri:
It appears that `npm install` is not allowed during the buld process by `opam-ci`. Are we expected to store the compiled js in the release and avoid using `npm`?

chaudhuri:
Will fix the build process upstream and submit an updated package later. No need to merge this.

PR #24717, opened 2023-10-30
Add conf-boost support for Linux Mint

jmid:
This little PR add `conf-boost` support for Linux Mint.

Since this is enabling an existing `depexts` on a platform not currently tested by opam-ci, it should not trigger any different behaviour (read: I expect failures to be pre-existing).

### DIFF
--- a/packages/conf-boost/conf-boost.1/opam
+++ b/packages/conf-boost/conf-boost.1/opam
@@ -5,7 +5,7 @@ authors: "Beman Dawes, David Abrahams, et al."
 bug-reports: "https://github.com/ocaml/opam-repository/issues"
 license: "MIT"
 depexts: [
-  ["libboost-dev"] {os-family = "debian"}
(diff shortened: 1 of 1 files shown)


mseri:
It adds support to ubuntu :)

jmid:
> It adds support to ubuntu :)

Apologies if I'm splitting hairs, but I believe on Ubuntu, `opam var os-family` yields `Debian` because Ubuntu is a Debian derivative. The problem then arises on Mint which is an Ubuntu derivative and yields
```
$ opam var os-family
ubuntu
```
More details and suggestions for a proper solution here: https://github.com/ocaml/opam/issues/4440
Until then, I am just playing the `os-family ubuntu` card to get things working on Mint (and other Ubuntu derivatives).

Thanks for the merge!

mseri:
Yes, I meant derivatives sorry

PR #24718, opened 2023-10-30
Add conf-python-3 support for Linux Mint

jmid:
This little PR adds conf-python-3 support for Linux Mint.

Locally, I can see the opam file is not `opam lint`ing cleanly:
```
$ opam lint packages/conf-python-3/conf-python-3.9.0.0/opam
/home/jmi/software/opam-repository/packages/conf-python-3/conf-python-3.9.0.0/opam: Warnings.
  warning 62: License doesn't adhere to the SPDX standard, see https://spdx.org/licenses/: "PSF-2.0"
```
despite that being a valid license ID: https://github.com/kit-ty-kate/spdx_licenses/blob/main/src/licenseIDs.ml#L363

I suspect it is just a consequence of me running an older `opam` though (2.1.2)


mseri:
Yes, that can depend on the version of opam. Thanks a lot

PR #24725, opened 2023-10-31
[new release] sail_sv_backend, sail_smt_backend, sail_output, sail_ocaml_backend, sail_manifest, sail_lem_backend, sail_latex_backend, sail_doc_backend, sail_coq_backend, sail_c_backend, sail and libsail (0.17)

Alasdair:
Project page: <a href="https://github.com/rems-project/sail">https://github.com/rems-project/sail</a>

##### Performance improvements

This release is primarily intended to fix performance issues. Overall
the Sail to C compilation can be almost 10x faster, and consumes
significantly less memory.

##### Order parameters deprecated

The order parameter on the bitvector and vector types no longer does
anything. The `default Order <ord>` statement now sets the bitvector
and vector ordering globally. In practice only POWER uses increasing
bit order, and there is never a valid reason to mix them in a
specification. Overall they added significant complexity to the
language for no real gain. Over subsequent releases a warning will be
added before they are eventually removed from the syntax.

##### String append pattern rework

For a while string append patterns `x ^ y` have been marked with a
special non-executable effect that forbids them from being used. Now
the implementation has been removed due to the deleterious effect
the generated code has on performance. Such clauses are now eagerly
removed from the syntax tree during rewriting pending a revised
implementation.

##### SystemVerilog backend (EXPERIMENTAL)

Sail can now produce SystemVerilog output using the -sv flag. Note
that this is not intended to be human readable or produce a
synthesizable design, but is instead intended to be used with
SystemVerilog verification tools like JasperGold.

### DIFF
--- a/packages/libsail/libsail.0.17/opam
+++ b/packages/libsail/libsail.0.17/opam
@@ -1,0 +1,69 @@
+opam-version: "2.0"
+synopsis:
+  "Sail is a language for describing the instruction semantics of processors"
+description: """
+Sail is a language for describing the instruction-set
+architecture (ISA) semantics of processors. Sail aims to provide a
+engineer-friendly, vendor-pseudocode-like language for describing
+instruction semantics. It is essentially a first-order imperative
+language, but with lightweight dependent typing for numeric types and
+bitvector lengths, which are automatically checked using Z3. It has
+been used for several papers, available from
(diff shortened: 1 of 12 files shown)


Alasdair:
Had some bug reports for commits in this release, so I will close this for now.

PR #24726, opened 2023-10-31
Fix Ortac 0.1.0 packages

shym:
This PR fixes the packages introduced by #24666. Those packages use plugins as set up by `dune-site` but were not calling `dune install` during the `build`; as a consequence, the command-line executable can’t find its plugins directory.


jmid:
The only CI error is that of `afl-persistent` failing to install on macOS which was also present in #24666 and [reported upstream](stedolan/ocaml-afl-persistent#13).

avsm:
thanks!

PR #24730, opened 2023-11-01
change build command of `ocaml` to support compilation in a path with spaces

Khady:
Related to https://github.com/ocaml/opam/issues/5679

This is a naive proposal to be able to compile the `ocaml` package with a version of ocaml that has been compiled in a folder with spaces in its name.


kit-ty-kate:
I don’t think this is a reasonable solution for the reason given in https://github.com/ocaml/opam/issues/5679#issuecomment-1788875621

PR #24687, opened 2023-10-26
[new release] h2 (7 packages) (0.11.0)

anmonteiro:
A high-performance, memory-efficient, and scalable HTTP/2 library for OCaml

- Project page: <a href="https://github.com/anmonteiro/ocaml-h2">https://github.com/anmonteiro/ocaml-h2</a>

##### CHANGES:

- h2-eio: adapt to Eio v0.12
  ([anmonteiro/ocaml-h2#225](https://github.com/anmonteiro/ocaml-h2/pull/225))
- h2-eio: track sw for gluten
  ([anmonteiro/ocaml-h2#215](https://github.com/anmonteiro/ocaml-h2/pull/215))
- h2: Add hpack {= version} constraint
  ([anmonteiro/ocaml-h2#214](https://github.com/anmonteiro/ocaml-h2/pull/214))
- h2, hpack, h2-async: Add opam constraints for angstrom and gluten-async
  ([anmonteiro/ocaml-h2#212](https://github.com/anmonteiro/ocaml-h2/pull/212))

PR #24732, opened 2023-11-01
grpc-examples: add upper bound on eio

avsm:
due to Eio.Flow.two_way object use, and interface change in eio.0.12

fixes revdeps in #24687


avsm:
cc @quernd

PR #24703, opened 2023-10-27
Align QCheck 0.9 and 0.10 versions

jmid:
The `qcheck*` opam files allow mixing versions across releases such as
```
  - install   qcheck        0.10   [required by fadbadml]
  - install   qcheck-core   0.21.2 [required by qcheck]
  - install   qcheck-ounit  0.21.2 [required by qcheck]
```
which ends up triggering a fadbadml lower bound failure in #24702:
```
#=== ERROR while compiling fadbadml.0.1.2 =====================================#
# context              2.2.0~alpha3~dev | linux/x86_64 | ocaml-base-compiler.5.1.0 | file:///home/opam/opam-repository
# path                 ~/.opam/5.1/.opam-switch/build/fadbadml.0.1.2
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p fadbadml -j 71 @install @runtest
# exit-code            1
# env-file             ~/.opam/log/fadbadml-7-6ebe6b.env
# output-file          ~/.opam/log/fadbadml-7-6ebe6b.out
### output ###
# (cd _build/default && /home/opam/.opam/5.1/bin/ocamlc.opt -w -40 -g -bin-annot -I test/unit/.run_test.eobjs/byte -I /home/opam/.opam/5.1/lib/bytes -I /home/opam/.opam/5.1/lib/ocaml/unix -I /home/opam/.opam/5.1/lib/ounit2 -I /home/opam/.opam/5.1/lib/ounit2/advanced -I /home/opam/.opam/5.1/lib/qcheck -I /home/opam/.opam/5.1/lib/qcheck-core -I /home/opam/.opam/5.1/lib/qcheck-core/runner -I /home/opam/.opam/5.1/lib/qcheck-ounit -I /home/opam/.opam/5.1/lib/seq -I /home/opam/.opam/5.1/lib/stdlib-shims -I src/.fadbad.objs/byte -no-alias-deps -open Dune__exe -o test/unit/.run_test.eobjs/byte/dune__exe__Run_test.cmo -c -impl test/unit/run_test.ml)
# File "test/unit/run_test.ml", line 43, characters 16-39:
# 43 |       match res.QCheck.TestResult.state with
#                      ^^^^^^^^^^^^^^^^^^^^^^^
# Error: Unbound record field QCheck.TestResult.state
```

For all `qcheck*` releases from 0.11 and above their versions are aligned.
This PR extends that range to uniformly include 0.9 and 0.10 too.

PR #24737, opened 2023-11-02
QCheck revdeps fixes - take 2

jmid:
This PR fixes a next batch of missing qcheck lower bounds spotted in #24703


jmid:
I've added another batch of fixes and I'm starting to run out of steam...
Because of additional missing lower bounds, this is no longer about qcheck revdeps - but revdeps of (patched) revdeps :grimacing:

mseri:
I suggest we fix whatever remaining failure when the chance comes. This PR is already large enough and each re0run takes lots of resources.


Thanks!

PR #24727, opened 2023-10-31
Add Topiary v0.3.0

Niols:
Website: https://topiary.tweag.io/
Repository: https://github.com/tweag/topiary
Release: https://github.com/tweag/topiary/releases/tag/v0.3.0
Change log: https://github.com/tweag/topiary/blob/v0.3.0/CHANGELOG.md

PR #24744, opened 2023-11-04
2 packages from mt-caret/polars-ocaml at 0.0.1

mt-caret:
This pull-request concerns:
- `polars.0.0.1`: OCaml bindings to the Polars dataframe library
- `polars_async.0.0.1`: Async OCaml bindings to the Polars dataframe library

---
* Homepage: https://github.com/mt-caret/polars-ocaml
* Source repo: git+https://github.com/mt-caret/polars-ocaml.git
* Bug tracker: https://github.com/mt-caret/polars-ocaml/issues

---


mt-caret:
I think the remaining errors here are due to the Rust version on these distributions being too old; i.e. the same reason as pointed out in https://github.com/ocaml/opam-repository/pull/24727#issuecomment-1788691286.

@Niols FYI while I agree there is no way to completely fix this (without some sort of change in opam / additional conf-* packages), I think minimizing the version dependencies (as described in https://github.com/mt-caret/polars-ocaml/blob/main/HACKING.md#adding-or-modifying-rust-dependencies) helps somewhat.

mt-caret:
Sorry, *now* the library should be ready for release.

Niols:
> @Niols FYI while I agree there is no way to completely fix this (without some sort of change in opam / additional conf-* packages), I think minimizing the version dependencies (as described in https://github.com/mt-caret/polars-ocaml/blob/main/HACKING.md#adding-or-modifying-rust-dependencies) helps somewhat.

That's great! I didn't know about this but I'll definitely be keeping it in mind for the future. Sadly, [Topiary's dependencies are exact version numbers](https://github.com/tweag/topiary/blob/79b93527d9bd59533f9a79fe490567963193fafd/Cargo.toml). Do you have a way around that that doesn't involve fully rewriting the set of dependencies myself?

mt-caret:
> > @Niols FYI while I agree there is no way to completely fix this (without some sort of change in opam / additional conf-* packages), I think minimizing the version dependencies (as described in https://github.com/mt-caret/polars-ocaml/blob/main/HACKING.md#adding-or-modifying-rust-dependencies) helps somewhat.
> 
> That's great! I didn't know about this but I'll definitely be keeping it in mind for the future. Sadly, [Topiary's dependencies are exact version numbers](https://github.com/tweag/topiary/blob/79b93527d9bd59533f9a79fe490567963193fafd/Cargo.toml). Do you have a way around that that doesn't involve fully rewriting the set of dependencies myself?

Hmm, sorry; I don't think I can think of anything :(

Niols:
Fair enough; thanks :-)

mseri:
Thanks. It builds where it can. For the other rust packages we also accept the fact that not all releases ship recent enough compilers

PR #24765, opened 2023-11-08
Some versions of omd are no longer available

mseri:
See https://github.com/ocaml/opam-repository/pull/24642


kit-ty-kate:
The archives are in the Software Heritage database but they sadly do not store the original archive :(
However they could serve as a new base for the archives to be moved to opam-source-archive and to make sure they haven’t been tempered beyond what they release process did they could be diffed with the tags for those releases.

mseri:
Thanks for checking, this is a useful place at least to check the tarballs contents indeed. In this case the package is not building so we can merge this PR without breaking anything

PR #24770, opened 2023-11-09
Frama-C: fix minimal version of dune-site

AllanBlanchard:
We discovered that while the compilation of Frama-C succeeds with `dune-site` 2.x, the installed Frama-C crashes when plug-ins are loaded. We are sure that with a `dune-site` with version >= to the minimal version it works, thus we choose to use this constraint.

Note that on 26.x and 27.x, we have a lower bound `3.5.0` on macOS, but the version of `dune-site` does not need this version, we only need `3.5` for binary signature.

The issue remained unnoticed for a long time because to detect the problem, one has to run Frama-C with options that guarantee that plug-ins are loaded. Thus, I don't know whether the CI process could be improved to detect this kind of problems since running the tests is obviously not a solution (since tests might have different version constraints).

### DIFF
--- a/packages/frama-c/frama-c.26.0/opam
+++ b/packages/frama-c/frama-c.26.0/opam
@@ -110,7 +110,7 @@ depends: [
   "dune" { (>= "3.2.0" & os!="macos") | (>= "3.5.0" & os="macos") }
   "dune-configurator"
   "dune-private-libs"
-  "dune-site"
+  "dune-site" { >= "3.2.0" }
 
   ( "alt-ergo-free" | "alt-ergo" )
   "conf-graphviz" { post }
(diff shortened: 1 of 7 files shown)


AllanBlanchard:
All CI failures were expected (and are the same as before):
- 26 beta had a problem with macOS
- 27 beta had a problem with the GCC flags during preprocessing
- 28 beta does not succeed on freebsd (and previous Frama-C versions have the same problem, but that's not a surprise).

avsm:
>  one has to run Frama-C with options that guarantee that plug-ins are loaded. 

The CI will run anything with the `{with-test}` predicate in the build instructions, so you would need to include a test in the package with the above test, and then modify the opam file to run this particular test.

PR #24779, opened 2023-11-13
[new release] loop (0.0.1)

leostera:
Unbounded loops with continues and breaks

- Project page: <a href="https://github.com/leostera/loop">https://github.com/leostera/loop</a>

## 0.0.1

Bootstrap with `loop`, `break`, and `continue`


avsm:

PR #24781, opened 2023-11-13
camlidl 1.10: add missing ocaml 5 upper bound

mseri:
camlidl 1.12 is the first release compatible with ocaml 5


mseri:
Apron's failures can be addressed in a separate PR

PR #24783, opened 2023-11-13
add tezt.4.0.0

rbardou:
Version 4.0.0 of [Tezt](https://gitlab.com/nomadic-labs/tezt) has been released.

### DIFF
--- a/packages/tezt/tezt.4.0.0/opam
+++ b/packages/tezt/tezt.4.0.0/opam
@@ -1,0 +1,41 @@
+opam-version: "2.0"
+maintainer: "contact@tezos.com"
+authors: ["Tezos devteam"]
+homepage: "https://gitlab.com/nomadic-labs/tezt/"
+bug-reports: "https://gitlab.com/nomadic-labs/tezt/issues"
+dev-repo: "git+https://gitlab.com/nomadic-labs/tezt.git"
+license: "MIT"
+depends: [
+  "dune" { >= "3.0" }
+  "ocaml" { >= "4.12" }
+  "re" { >= "1.7.2" }
+  "lwt" { >= "5.6.0" }
+  "base-unix"
+  "ezjsonm" { >= "1.1.0" }
+  "clap" { >= "0.3.0" }
+  "conf-npm" { with-test }
+  "js_of_ocaml" { with-test }
+  "js_of_ocaml-lwt" { with-test }
+  "ocamlformat" { with-test & = "0.21.0" }
+]
+depopts: [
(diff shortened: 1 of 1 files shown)


rbardou:
The CI shows a test that fails. This test doesn't fail on my laptop nor in Tezt's CI, so that's weird. It is a test about signal handling and timeouts. The logs show:
```
[08:42:56.178] Will now terminate process.
[08:42:56.178] Send SIGTERM to process.
[08:42:56.178] Process.wait returned after 1.00163 seconds from start.
```
instead of the expected:
```
[08:48:12.094] Will now terminate process.
[08:48:12.094] Send SIGTERM to process.
[08:48:12.094] [process] received SIGTERM, ignoring
[08:48:13.094] Send SIGKILL to process.
[08:48:13.095] process was killed by signal SIGKILL.
[08:48:13.096] Process.wait returned after 2.00257 seconds from start.
```
So it looks like the following lines of code, which are supposed to make the tested process ignore SIGTERM, does not work in opam-ci:
```
let on_sigterm _ = print_endline "received SIGTERM, ignoring"
let () = Sys.set_signal Sys.sigterm (Signal_handle on_sigterm)
```
Is it possible that opam-ci's environment causes SIGTERM to not be handled somehow?

rbardou:
About failing revdeps: this is expected — Tezt 4.0.0 contains some breaking changes.

Should this PR also add `< 4.0.0` constraints to existing packages that depend on Tezt? (There are a lot of them, I think it will also cause the CI to have to recheck all of them?)

rbardou:
Thanks for your feedback, I'll try to make a PR for revdeps.

> Is this a new test in 4.0.0 ? Has it been updated or modified on your side?

It's a new test. Well, actually, it's an old test that was broken and that I fixed in 4.0.0. It used to accept invalid behavior such as the one that we observed in this PR's CI.

rbardou:
https://github.com/ocaml/opam-repository/pull/24785 adds a version constraint to `octez-shell-libs`.

The revdep check also fail for `tezos-rpc-http-server` but I think this is completely unrelated to Tezt (my bet is on a missing constraint on another package like `cohttp`).

rbardou:
It looks like the 2 remaining issues detected by the CI can be ignored:
- failing Tezt test: seems specific to this CI somehow (would be great to understand it but out of scope for Tezt 4.0.0);
- failing `tezos-rpc-http-server` build: seems unrelated to Tezt.

I think this is thus ready to merge?

raphael-proust:
I agree this is ready to merge

PR #24785, opened 2023-11-13
constrain octez-shell-libs to tezt < 4.0.0

rbardou:
https://github.com/ocaml/opam-repository/pull/24783 upgrades Tezt to version 4.0.0, but some revdeps failed in the CI:
- `octez-shell-libs` depends on features that changed in Tezt 4.0.0;
- `tezos-rpc-http-server` failed too for reasons that seem unrelated.

This MR should fix `octez-shell-libs`. I checked in Octez's codebase and this is the only library where the feature of Tezt that changed was used; and the feature was only used starting from v18.0 of Octez. So I think this is the only package that needs the `< 4.0.0` constraint.

Regarding `tezos-rpc-http-server`, I think the package is broken for reasons unrelated to Tezt. Looks like it is missing a version constraint on some package, maybe `cohttp`?

### DIFF
--- a/packages/octez-shell-libs/octez-shell-libs.18.0/opam
+++ b/packages/octez-shell-libs/octez-shell-libs.18.0/opam
@@ -14,7 +14,7 @@ depends: [
   "ringo" {>= "1.0.0"}
   "aches" {>= "1.0.0"}
   "prometheus" {>= "1.2"}
-  "tezt" {>= "3.1.1"}
+  "tezt" {>= "3.1.1" & < "4.0.0"}
   "tezt-tezos" {with-test & = version}
   "octez-alcotezt" {with-test & = version}
   "astring" {with-test}


rbardou:
The CI fails with a bunch of errors that look like this:
```
# (cd _build/default/src/lib_p2p/test && ./test_p2p_io_scheduler.exe)
# Fatal error: exception Operation not permitted
# Raised by primitive operation at Tezt_tezos__Port.fresh.(fun) in file "tezt/lib_tezos/port.ml", line 42, characters 2-65
# Called from Stdlib__Fun.protect in file "fun.ml", line 33, characters 8-15
# Re-raised at Stdlib__Fun.protect in file "fun.ml", line 38, characters 6-52
# Called from Dune__exe__Test_p2p_io_scheduler in file "src/lib_p2p/test/test_p2p_io_scheduler.ml", line 242, characters 18-44
```
I don't see how adding the constraint `tezt < 4.0.0` could make anything fail since there are no versions of Tezt greater than 3.1.1. So I think this error should be ignored.

Also, even though it may look like this is in fact Tezt-related (since the error is about `tezt/lib_tezos/port.ml`), the library in `tezt/lib_tezos` is a different library than Tezt itself. It looks like the way socket ports are managed in this library doesn't work in opam's CI? In any case, I don't think this particular set of errors should block this PR.

raphael-proust:
Failures are unrelated AFAICT

PR #24793, opened 2023-11-15
**DO NOT MERGE** Test if gappa package passes CI without patch

MSoegtropIMC:
This is just a test if the gappa package passes opam CI without the patch I added a few years back. It is expected that modern OSes support CC+11 without any extra messing (this was not so when I originally added the package).


MSoegtropIMC:
Closing this (the test result is that this patch is still required on some linux flavors).

Issue #24771, opened 2023-11-09
Should ocamlbuild.0.14.2 be blacklisted on Windows?


maroneze:
I had some issues with a package relying on ocamlbuild, while using Cygwin+MinGW. For some reason, I had ocamlbuild 0.14.2 installed, and it seemed to work fine, but when installing `ptime`, which uses ocamlbuild, the compilation failed with a not-so-trivial error message. By chance, I realized that a `ocamlbuild.0.14.2+win` package existed, installed it, and things seem to work fine.

Then, my question is: shouldn't `ocamlbuild.0.14.2` be blacklisted on a Windows system, to avoid this kind of issues? Or is there some reason to keep it available?

raphael-proust:
Yes, the package should be blacklisted on windows, and so should all ocamlbuild packages which do not have the `+win` suffix.

PR #24797, opened 2023-11-15
mark ocamlbuild as not available on windows

raphael-proust:
Fixes https://github.com/ocaml/opam-repository/issues/24771


mseri:
We only need the lint here, no need to test the world, and those are fine

dra27:
Thanks!

PR #24804, opened 2023-11-16
[new release] links, links-postgresql, links-sqlite3 and links-mysql (0.9.8)

frank-emrich:

##### CHANGES:

This release includes minor bug fixes, improvements, and a breaking change regarding the syntax of handlers.

## Queries mixing set and bag semantics
Links now provides experimental support for SQL queries with grouping and
aggregation. These require the _mixing_ normaliser (`mixing_norm=on` in the
configuration file).

The result of grouping over a relation is represented as a finite map, which in
Links is treated as a list of (grouping key, associated subrelation) pairs.
Aggregation can then be applied groupwise to a finite map to obtain again a
relation. Such Links queries are translated to SQL queries using `group by` and
aggregates.

Further information on this feature is provided in the [Links GitHub
wiki](https://github.com/links-lang/links/wiki/Grouping-and-aggregation).

## New syntax for handlers

The syntax for handlers has changed. Instead of
```
handle (...) {
  case Op(params, resumption) -> ...
  case Return(x) -> x
}
```
we write
```
handle (...) {
  case <Op(params) => resumption> -> ...
  case x -> x
}
```

We can also write `case <Op(params) -> resumption> -> ...`.
For now, the semantics for `->` and `=>` are the same.

## Control-flow linearity
Links now tracks control-flow linearity when the flag `--control-flow-linearity`
is enabled. This extension fixes a long-standing soundness bug (see issue
[links-lang/links#544](https://github.com/links-lang/links/issues/544)) with the interaction
between exceptions, multi-shot effect handlers and session-typed channels. More
details about this extension can be found in the [wiki
page](https://github.com/links-lang/links/wiki/Control-flow-linearity).

## Other changes and fixes
* The package `links-mysql` now relies on the `mysql8` package as its underlying
  database driver.
* Links now supports OCaml 5.
* The `SessionFail` effect can now be supressed by disabling the
  `expose_session_fail` setting. When doing so, the `SessionFail` effect
  is included in the `wild` effect instead.
* The (incomplete) support for distributed session delegation has been removed.
* The standard library now provides functions for issuing SPARQL queries.

### DIFF
--- a/packages/links-mysql/links-mysql.0.9.8/opam
+++ b/packages/links-mysql/links-mysql.0.9.8/opam
@@ -14,7 +14,7 @@ build: [
 ]
 
 depends: [
-  "ocaml" {>= "4.08.0"}
+  "ocaml" {>= "4.14.0"}
   "dune" {>= "2.7"}
   "conf-mysql"
   "mysql8"
--- a/packages/links-mysql/links-mysql.0.9.8/opam
+++ b/packages/links-mysql/links-mysql.0.9.8/opam
@@ -1,0 +1,31 @@
+opam-version: "2.0"
+maintainer: "James Cheney <jcheney@inf.ed.ac.uk>"
+authors: "The Links Team <links-dev@inf.ed.ac.uk>"
+synopsis: "MySQL database driver for the Links Programming Language"
+description: "MySQL database driver for the Links Programming Language"
+homepage: "https://github.com/links-lang/links"
+dev-repo: "git+https://github.com/links-lang/links.git"
+bug-reports: "https://github.com/links-lang/links/issues"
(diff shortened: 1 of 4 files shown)


frank-emrich:
It seems like all of the test failures are due to database-related packages that we depend on failing to build, most notably `mysql8`, which our own `links-mysql` depends on.
I'm not really familiar with how the opam CI works, but is it possible that the package  `mysql8` has gone "stale" in the sense that the CI would fail if run on `mysql8` itself, independently from our release of the `links-*` packages?

mseri:
I think it may be also due t version mismatches since on some distributions it works fine. This is not a problem of links though, we can merge

PR #24814, opened 2023-11-20
[new release] riot (0.0.3)

leostera:

## 0.0.3

* Big namespace refactor. `Riot.Runtime` includes the lower-level runtime
  blocks, and everything else that is more user-friendly lives at the `Riot.*`
level.
* Introduce reduction counting, so processes will run up to N iterations unless
  they finish, or they execute an unhandled effect.
* Introduce the `Application` interface for managing the lifecycle of the system
* Fix `Riot.Logger` to fit the `Application` interface
* Add a new `Riot.Telemetry` backend for doing async telemetry

### DIFF
--- a/packages/riot/riot.0.0.3/opam
+++ b/packages/riot/riot.0.0.3/opam
@@ -1,0 +1,44 @@
+opam-version: "2.0"
+synopsis: "An actor-model multi-core scheduler for OCaml 5"
+description:
+  "Riot is an actor-model multi-core scheduler for OCaml 5. It brings Erlang-style concurrency to the language, where lighweight process communicate via message passing"
+maintainer: ["Leandro Ostera <leandro@abstractmachines.dev>"]
+authors: ["Leandro Ostera <leandro@abstractmachines.dev>"]
+license: "MIT"
+tags: ["topics" "multicore" "erlang" "actor" "message-passing" "processes"]
+homepage: "https://github.com/leostera/riot"
+bug-reports: "https://github.com/leostera/riot/issues"
+depends: [
+  "ocaml" {>= "5.1"}
+  "dune" {>= "3.10"}
+  "ptime" {>= "1.1.0"}
+  "iomux" {>= "0.3"}
+  "bigstringaf" {>= "0.9.1"}
+  "uri" {>= "4.4.0"}
+  "telemetry" {>= "0.0.1"}
+  "odoc" {with-doc & >= "2.2.2"}
+]
+build: [
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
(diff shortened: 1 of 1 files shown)


leostera:
There's a failing revdep (`atacama 0.0.1`) that I own but I won't be releasing a 0.0.2 until later this week. I don't care that its broken atm.

Checking the timeouts now, looks like one of the tests is flaky.

mseri:
We do care about broken packages though.
I am going to add the upper bound on riot in a separate PR and merge this once that is merged

leostera:
I appreciate that, thanks @mseri 🙏🏼 – a follow up question: what is the etiquette for me to handle several separate/depending packages? 

For ex. riot I'll likely be publishing more patch releases indefinitely (0.0.1...0.0.N) until it stabilizes, and I expect all of those to be considered unstable (hence the major version 0). 

Right now I have another 6 packages (only 1 released, 5 to be released slowly). I could fix the version like `(riot (= "0.0.3"))` and then every time those are ready bump the version up, 
but I'd not want to flush y'all with PRs that update dependants.

What I'm considering/wondering here is:

* semver here indicates things are unstable, so the expectations should be clear – am I expected to only publish stable releases to opam?

* the review process is still manual, and I don't want to put unnecessary stress on y'all – if the barrier to using the language is lowered (and that's part of the current OCaml Tooling Roadmap), we'll see an uptick in the number of packages published. how should I expect this to scale re: publishing times?

* i happen to be responsible for this revdep! - what would be the procedure if reverse deps finds breaking requirements on other people's packages? Would I be waiting on fixing those packages before publishing mine?

Hope this makes sense, as I'm sure y'all have better things to do than review/fix my alpha software breaking around.

As always, super grateful for your time! 🙌🏼

mseri:
Thanks for being so thoughtful!

In my opinion it is perfectly fine to publish unstable releases and I think it is fine to have lax bounds (as long as they build fine when releasing). It is also perfectly fine to use the CI run as a way to see if revdeps need upper bounds.

It is nice (and convenient for us) if you send us also a PR to fix the upper bounds from the revdeps when you submit your packages, but even there there is no strict rule. Sometimes we notice the issue first or you may not have time and we do it ourselves, which may delay a bit the release but normally not too substantially, and sometimes the PRs owners help us with those.

Concerning publishing times: these are usually very fast for small PRs where nothing is broken or just some metadata needs fixing, and slower for larger PRs or where major changes are required. In many cases, my experience is that after the first couple of slower release reviews, the metadata are fine and the maintainers are acquainted with the process, so it gets relatively fast.

Of course a lot depend also on our external working constraints, so sometimes we are very reactive and in other periods much slower. So if you also send a PR to fix the revdeps of your release this speeds up things a bit :)

PR #24816, opened 2023-11-21
Use clang on macOS for tsan support

tmcgilchrist:
Using the default cc on macOS does not provide ThreadSantizer support, I get this error on macOS Sonoma 14.1:
```
$ opam switch create 5.1.0+tsan --no-install

<><> Installing new switch packages <><><><><><><><><><><><><><><><><><><><>  🐫 
Switch invariant: ["ocaml-variants" {= "5.1.0+tsan"}]
Constructing initial basis...
Number of 0-1 knapsack inequalities = 9
Constructing conflict graph...
Conflict graph has 6 + 1 = 7 vertices

<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><>  🐫 
∗ installed base-bigarray.base
∗ installed base-threads.base
∗ installed base-unix.base
⬇ retrieved ocaml-variants.5.1.0+tsan  (cached)
∗ installed conf-pkg-config.3
∗ installed conf-unwind.0
[ERROR] The compilation of ocaml-variants.5.1.0+tsan failed at "./configure --prefix=/Users/tsmc/.opam/5.1.0+tsan
        --docdir=/Users/tsmc/.opam/5.1.0+tsan/doc/ocaml -C CC=cc --enable-tsan --disable-warn-error".

#=== ERROR while compiling ocaml-variants.5.1.0+tsan ==========================#
# context     2.1.5 | macos/x86_64 |  | https://opam.ocaml.org#2dcd8684
# path        ~/.opam/5.1.0+tsan/.opam-switch/build/ocaml-variants.5.1.0+tsan
# command     ~/.opam/opam-init/hooks/sandbox.sh build ./configure --prefix=/Users/tsmc/.opam/5.1.0+tsan --docdir=/Users/tsmc/.opam/5.1.0+tsan/doc/ocaml -C CC=cc --enable-tsan --disable-warn-error
# exit-code   1
# env-file    ~/.opam/log/ocaml-variants-93286-9def9a.env
# output-file ~/.opam/log/ocaml-variants-93286-9def9a.out
### output ###
# [...]
# checking whether fma works... yes
# checking for cc options needed to detect all undeclared functions... none needed
# checking for unistd.h... (cached) yes
# checking whether getentropy is declared... no
# checking for getrusage... yes
# checking for times... yes
# checking for secure_getenv... no
# checking for __secure_getenv... no
# checking for issetugid... yes
# checking for mach_timebase_info... yes
# checking for mach_absolute_time... yes
# configure: error: thread sanitizer not supported with compiler cc"
```
With this patch I can create a tsan switch. I have tested on both Ventura and Sonoma (x86_64) but nothing previous to that.

### DIFF
--- a/packages/ocaml-variants/ocaml-variants.5.1.0+tsan/opam
+++ b/packages/ocaml-variants/ocaml-variants.5.1.0+tsan/opam
@@ -39,7 +39,8 @@ build: [
     "--disable-flat-float-array" {ocaml-option-no-flat-float-array:installed}
     "--enable-flambda" {ocaml-option-flambda:installed}
     "--enable-frame-pointers" {ocaml-option-fp:installed}
-    "CC=cc" {!ocaml-option-32bit:installed & !ocaml-option-musl:installed & (os="openbsd"|os="macos")}
+    "CC=cc" {!ocaml-option-32bit:installed & !ocaml-option-musl:installed & (os="openbsd")}
+    "CC=clang" {!ocaml-option-32bit:installed & !ocaml-option-musl:installed & (os="macos")}
     "CC=musl-gcc" {ocaml-option-musl:installed & os-distribution!="alpine"}
     "CFLAGS=-Os" {ocaml-option-musl:installed}
     "--enable-tsan"
--- a/packages/ocaml-variants/ocaml-variants.5.1.0~rc3+tsan/opam
+++ b/packages/ocaml-variants/ocaml-variants.5.1.0~rc3+tsan/opam
@@ -32,7 +32,8 @@ build: [
     "--disable-flat-float-array" {ocaml-option-no-flat-float-array:installed}
     "--enable-flambda" {ocaml-option-flambda:installed}
     "--enable-frame-pointers" {ocaml-option-fp:installed}
-    "CC=cc" {!ocaml-option-32bit:installed & !ocaml-option-musl:installed & (os="openbsd"|os="macos")}
+    "CC=cc" {!ocaml-option-32bit:installed & !ocaml-option-musl:installed & (os="openbsd")}
+    "CC=clang" {!ocaml-option-32bit:installed & !ocaml-option-musl:installed & (os="macos")}
     "CC=musl-gcc" {ocaml-option-musl:installed & os-distribution!="alpine"}
     "CFLAGS=-Os" {ocaml-option-musl:installed}
     "--enable-tsan"
--- a/packages/ocaml-variants/ocaml-variants.5.2.0+trunk/opam
+++ b/packages/ocaml-variants/ocaml-variants.5.2.0+trunk/opam
@@ -43,6 +43,7 @@ build: [
     "--enable-frame-pointers" {ocaml-option-fp:installed}
     "--enable-tsan" {ocaml-option-tsan:installed}
     "CC=cc" {!ocaml-option-32bit:installed & !ocaml-option-musl:installed & (os="openbsd"|os="macos")}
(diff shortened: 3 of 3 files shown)


dra27:
Just for consistency, please could the change be applied to `ocaml-variants.5.0.0+tsan` and `ocaml-variants.5.1.0~rc3+tsan` (assuming they do both also work on macOS!).

I don't have access to x86 macOS hardware, but are `cc` and `clang` the same binary? i.e. is the issue that clang behaves differently depending on its basename?

A similar change could do with being made to `ocaml-variants.5.2.0+trunk`, but that probably needs to be more conservative - i.e. leave `CC=cc` for the default case and specifically have `CC=clang` if `os="macos" & ocaml-option-tsan:installed` and, with apologies for yet another shopping list item, that change also needs to go in a PR for ocaml/ocaml (cf. #24341)

mseri:
Yes, if I am not mistaken they have the same inode:
```
❯ stat /usr/bin/cc
16777232 1152921500312505605 -rwxr-xr-x 76 root wheel 0 168160 "Nov  2 23:44:56 2023" "Nov  2 23:44:56 2023" "Nov  2 23:44:56 2023" "Nov  2 23:44:56 2023" 4096 32 0x80020 /usr/bin/cc
❯ stat /usr/bin/clang
16777232 1152921500312505605 -rwxr-xr-x 76 root wheel 0 168160 "Nov  2 23:44:56 2023" "Nov  2 23:44:56 2023" "Nov  2 23:44:56 2023" "Nov  2 23:44:56 2023" 4096 32 0x80020 /usr/bin/clang
```

mseri:
And
```
❯ cc --version
Apple clang version 15.0.0 (clang-1500.0.40.1)
Target: arm64-apple-darwin23.1.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
❯ clang --version
Apple clang version 15.0.0 (clang-1500.0.40.1)
Target: arm64-apple-darwin23.1.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

tmcgilchrist:
Confirming that cc and clang are the same inode and reported version for my mac.
 
I've updated `ocaml-variants.5.1.0~rc3+tsan` and `ocaml-variants.5.2.0+trunk` as suggested @dra27 and validated they build, with some cursory testing of tsan itself.

Note:
 * `opam switch create 5.0.0+tsan --no-install` works as is on MacOS Ventura
 * `opam switch create 5.1.0~rc3+tsan --no-install` fails on MacOS Ventura without the latest change.
 * `opam switch create 5.2.0+tsan --packages=ocaml-variants.5.2.0+trunk,ocaml-option-tsan` also fails
 
 Also the macOS tests aren't finding a solution in opam-repo-ci so aren't actually testing this PR. @mseri

tmcgilchrist:
Created an issue https://github.com/ocurrent/opam-repo-ci/issues/248

MisterDA:
I've submitted a patch for this issue on OCaml: https://github.com/ocaml/ocaml/pull/12768. The problem is that TSan support was activated using a variable called `cc_basename`, which holds the basename of the compiler executable. The configure script would enable TSan if `cc_basename` matched `gcc*` or `clang*`. On macOS however, the default compiler is named `cc` (but it's really Clang).
The problem can be reproduced on a machine where `cc` links to `gcc` (if Autoconf defaults to gcc) with `./configure --enable-tsan CC=cc`.
You might want to carry the patch from my PR into opam-repository.

PR #24841, opened 2023-11-24
why3: really ensure configure works with Cygwin (also in non-dev mode)

maroneze:
The previous patch worked in a dev environment (which regenerates `configure` from `configure.in`), but in the general case, it was insufficient to ensure that why3 compiles under Cygwin.

This patch, also applied to `configure`, should hopefully fix that.

Note that the previous patch to `configure.in` must remain, otherwise the wrong `configure` will be regenerated in a `dev` build.

Also, the patch has been applied upstream to why3, so its future releases won't need it anymore.


mseri:
Thanks!

PR #24842, opened 2023-11-24
Mark monadlib as unavailable

Leonidas-from-XIV:
The Github user is gone and neither of the archives can be found in the OPAM cache.


Leonidas-from-XIV:
Not sure what to do about `jhupllib` 0.1, 0.1.1, 0.2.1 and 0.2.2 which depend on `monadlib` and thus also become uninstallable. But I guess the solver will know that these can't ever be fulfilled so I've left them as-is.

Leonidas-from-XIV:
@mseri Ah, that's much better, thanks!

mseri:
Thanks

Issue #24786, opened 2023-11-13
omd archives


rjbou:
Most of omd archives are present in opam repo caches (at least opam.ocaml.org & ocamlpro ones, and mine).
Here is a full list, done by grepping on hashes. Most of missing archive are the extra-files omd.install (0395736894a46f718a77f59ec6fbf1fd). No archives lines with (*) means that i have that archive.
```
omd.0.2:0395736894a46f718a77f59ec6fbf1fd: no archive
omd.0.2:22ad98d046ef208ff13f42d5aedb6e5b: https://gitlab.ocamlpro.com/OCamlPro/opam-repository/-/raw/cached/cache/md5/22/22ad98d046ef208ff13f42d5aedb6e5b
omd.0.2:22ad98d046ef208ff13f42d5aedb6e5b: https://opam.ocaml.org/cache/md5/22/22ad98d046ef208ff13f42d5aedb6e5b
omd.0.3:0395736894a46f718a77f59ec6fbf1fd: no archive
omd.0.3:522e5c2937e77b48dcdc40d85437d29e: https://gitlab.ocamlpro.com/OCamlPro/opam-repository/-/raw/cached/cache/md5/52/522e5c2937e77b48dcdc40d85437d29e
omd.0.4:0395736894a46f718a77f59ec6fbf1fd: no archive
omd.0.4:8c81e10c2ae6f2f64aeb78bbdcfbd48e: https://gitlab.ocamlpro.com/OCamlPro/opam-repository/-/raw/cached/cache/md5/8c/8c81e10c2ae6f2f64aeb78bbdcfbd48e
omd.0.5.4:0395736894a46f718a77f59ec6fbf1fd: no archive
omd.0.5.4:9e4e2a4a9348cb0260d170edb3e2ce9e: https://gitlab.ocamlpro.com/OCamlPro/opam-repository/-/raw/cached/cache/md5/9e/9e4e2a4a9348cb0260d170edb3e2ce9e
omd.0.5.5:0395736894a46f718a77f59ec6fbf1fd: no archive
omd.0.5.5:8de7bbba97c53ebad68eff9fa259200f: https://gitlab.ocamlpro.com/OCamlPro/opam-repository/-/raw/cached/cache/md5/8d/8de7bbba97c53ebad68eff9fa259200f
omd.0.5:0395736894a46f718a77f59ec6fbf1fd: no archive
[...]

PR #24850, opened 2023-11-25
omd update old links

haochenx:
- fixes https://github.com/ocaml/opam-repository/issues/24786
- pended by: https://github.com/ocaml/opam-source-archives/pull/23 
  - **this PR won't build without ↑ being merged first**


haochenx:
We are getting `UnsignificantlyChanged` from the analysis step which result the CI not running..

Not sure what to do.

mseri:
Thanks a lot. The lint issues in the old versions don't need to be addressed imo

mseri:
The "not significantly changed" is on purpose. The lint has checked that the tarballs are reachable and have the right checksums, which is what matters here. It is not really worth to spend weeks and hours of CI to fix lower bounds for such ancient packages

PR #24577, opened 2023-10-04
add missing dependencies to ocaml in dkml-install-runner packages

raphael-proust:
The lint job of PR #24542 shows the following errors:

```
2023-09-28 10:11.50: Error in dkml-install-runner.0.3.0:            warning 41: Some packages are mentioned in package scripts or features, but there is no dependency or depopt toward them: "ocaml"
2023-09-28 10:11.50: Error in dkml-install-runner.0.2.0:            warning 41: Some packages are mentioned in package scripts or features, but there is no dependency or depopt toward them: "ocaml"
2023-09-28 10:11.50: Error in dkml-install-runner.0.1.0:            warning 41: Some packages are mentioned in package scripts or features, but there is no dependency or depopt toward them: "ocaml"
2023-09-28 10:45.54: Job failed: 3 errors
```

This PR adds the OCaml dependencies. The version constraints are based on [opam-health-check results](https://check.ci.ocaml.org/?comp=4.08&comp=4.09&comp=4.10&comp=4.11&comp=4.12&comp=4.13&comp=4.14&comp=5.0&comp=5.1&available=4.08&available=4.09&available=4.10&available=4.11&available=4.12&available=4.13&available=4.14&available=5.0&available=5.1&show-only=good&show-only=partial&show-only=bad&show-only=not-available&show-only=internal-failure&maintainers=&logsearch=dkml-install-runner&logsearch_comp=4.14)

PR #24852, opened 2023-11-25
4 packages from diskuv/dkml-install-api at 0.5.2

jonahbeckford:
This pull-request concerns:
-`dkml-install.0.5.2`: API and registry for DkML installation components
-`dkml-install-installer.0.5.2`: Build tools for DkML installers
-`dkml-install-runner.0.5.2`: Runner executable for DkML installation
-`dkml-package-console.0.5.2`: Console setup and uninstall executables for DkML installation

---
* Homepage: https://github.com/diskuv/dkml-install-api
* Source repo: git+https://github.com/diskuv/dkml-install-api.git
* Bug tracker: https://github.com/diskuv/dkml-install-api/issues

---

### DIFF
--- a/packages/dkml-install-runner/dkml-install-runner.0.5.2/opam
+++ b/packages/dkml-install-runner/dkml-install-runner.0.5.2/opam
@@ -1,0 +1,48 @@
+opam-version: "2.0"
+synopsis: "Runner executable for DkML installation"
+description:
+  "The runner executable is responsible for loading and running all DkML installation components."
+maintainer: "opensource+dkml@support.diskuv.com"
+authors: "Diskuv, Inc. <opensource+dkml@support.diskuv.com>"
+license: "Apache-2.0"
+homepage: "https://github.com/diskuv/dkml-install-api"
+bug-reports: "https://github.com/diskuv/dkml-install-api/issues"
+depends: [
+  "dune" {>= "2.9"}
+  "alcotest" {>= "1.4.0" & with-test}
+  "dkml-install" {= version}
+  "ppx_expect" {>= "v0.14.1"}
+  "astring" {>= "0.8.5"}
+  "bos" {>= "0.2.0"}
+  "cmdliner" {>= "1.1.1"}
+  "fmt" {>= "0.8.9"}
+  "logs" {>= "0.7.0"}
+  "diskuvbox" {>= "0.1.1"}
+  "odoc" {with-doc}
+]
+available: os = "win32" | os = "linux" | os = "macos"
+build: [
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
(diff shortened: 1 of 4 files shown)


packages/dkml-install-runner/dkml-install-runner.0.5.2/opam
```diff
@@ -0,0 +1,48 @@
...
+synopsis: "Runner executable for DkML installation"
+description:
+  "The runner executable is responsible for loading and running all DkML installation components."
+maintainer: "opensource+dkml@support.diskuv.com"
+authors: "Diskuv, Inc. <opensource+dkml@support.diskuv.com>"
+license: "Apache-2.0"
+homepage: "https://github.com/diskuv/dkml-install-api"
+bug-reports: "https://github.com/diskuv/dkml-install-api/issues"
+depends: [
+  "dune" {>= "2.9"}
```
haochenx, line 11: Likely missing

```suggestion
  "ocaml" {>= "4.10"}
  "dune" {>= "2.9"}
```

as per https://github.com/ocaml/opam-repository/pull/24577 pointed out. Maybe it's better to upstream this?

haochenx (commented):
CI had a few cancelled jobs which I restarted.

I'm not entirely sure the inline comment needs to be addressed, and if it does need address, why its not caught by linter on this PR.

@raphael-proust @mseri please take a look and advice

jonahbeckford:
Clicked the `Rebuild failed` button because the failing jobs had "Cancelled" due to 120 minute timeout. Nothing to do with the PR; sounds like the CI machines were stalled for other reasons.

packages/dkml-install-runner/dkml-install-runner.0.5.2/opam
jonahbeckford: Why is `ocaml.4.10+` necessary?

haochenx: I suggested `ocaml.4.10+` only because it is so in [`dkml-install-runner.0.5.1`](https://github.com/ocaml/opam-repository/blob/e8686e51e05af0100e5232719f2895091f580a99/packages/dkml-install-runner/dkml-install-runner.0.5.1/opam). I don't have a strong opinion here.

It seems that the lower bound wasn't picked by you. As the CI is currently succeeding, we probably do not need this dependency. Yet there is a policy to have a dependency to `ocaml`, so it would be nice to add it like:
```suggestion
  "ocaml"
  "dune" {>= "2.9"}
```

mseri:
We did add the ocaml dependency because it was caught by the linter, but it was not essential since `dkml-install` already comes with an ocaml lower bound. The linter is no longer complaining about that, so I suggest we merge

mseri:
Thanks to both of you

PR #24837, opened 2023-11-24
Why3 1.7.0

silene:

PR #24856, opened 2023-11-27
frama-c: add constraints for upcoming why3 version

maroneze:
The upcoming why3 1.7.0 version has some incompatibilities with Frama-C, so we must add these version constraints.


maroneze:
Superseded by #24837, which I had not seen before.

PR #24857, opened 2023-11-28
hts_shrink_301: correct opam file

UnixJunkie:
does not depend on: parmap, bitv, ocamlgraph, ptmap


UnixJunkie:
those libraries are not in the dune file, they are not needed

mseri:
Thanks

Issue #24263, opened 2023-08-14
Old result packages shadows the `Result` modules for newer builds


paurkedal:
The [CI from my latest Caqti PR](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/2340fa3f28379eb7062a8d17af3fa3389df4faf2) shows that some packages fail during revdep builds since they require features from the `Result` module from stdlib which gets shadowed by older releases of the result package due to implicit dependencies.  That is, library A depends on B which depends on result and can tolerate an old version, but A assumes it gets the `Result` module from a newer compiler distribution.

After adding
```
conflicts: [
  "result" {< "1.5"}
]
```
to my own package avoids the issue.  However, there are other packages having the same issues.  The build discovers sihl-core and sihl, but there could be more.  A simpler but more pervasive solution may be to add an `"ocaml" {< "4.08"}` constraints to the old result packages?

hannesm:
see https://github.com/ocaml/opam-repository/issues/19880 - for reasons still unclear to me, this "breaks lockfiles", and somehow those "lockfiles" are important.

In any case, maybe "result" packages should carry the conflict burden, seems to be accepted in the opam-repository "for exceptional cases" (see https://github.com/ocaml/opam-repository/pull/22512).

TL;DR: so far the "maintainer" of the result package, and some opam repostory maintainers prefer to carry the burden on your shoulder to add "conflicts with result < 1.5", even if your package does not have a dependency on result.

avsm:
We will have a window to fix this over the coming months when we roll out the "mainline opam repository only supports OCaml 4.08 or higher", as discussed in https://discuss.ocaml.org/t/raising-the-minimum-tested-version-of-ocaml-to-4-05-or-4-08-from-4-02/12464

This restriction will in turn lift up the floor of result packages such that the conflict will no longer be required, I believe...

mseri:
Honestly I don't see why not adding the upper bound and relieve the self-inflicted pain of adding a conflict on every other package released. The conflict is now on so many packages that it is almost equivalent to an upper bound, we may as well just to that and forget about it.

We discussed this at the maintainers meeting and there is no objection to send a PR once the CI is a bit freer (now there are tezos and dune, so we will not overload it more)

mseri:
I suggest we close this for the time being since it does prevent installability of a number of packages (and potentially can be breaking if somebody uses lock files). I will look for a solution for the lower bound checker that takes result into account.

PR #24868, opened 2023-11-29
Add upper bound on ocaml 4.08 to result <1.5

mseri:
This will make the constraints no longer necessary.
See https://github.com/ocaml/opam-repository/issues/24263

### DIFF
--- a/packages/result/result.1.0/opam
+++ b/packages/result/result.1.0/opam
@@ -12,7 +12,7 @@ description: """
 Projects that want to use the new result type defined in OCaml >= 4.03
 while staying compatible with older version of OCaml should use the
 Result module defined in this library."""
-depends: ["ocaml" {< "4.14"}]
+depends: ["ocaml" {< "4.08"}]
 url {
   src: "https://github.com/janestreet/result/archive/1.0.tar.gz"
   checksum: "md5=01391a66385ab1a43f90455dfb5c6843"
--- a/packages/result/result.1.1/opam
+++ b/packages/result/result.1.1/opam
@@ -11,7 +11,7 @@ description: """
 Projects that want to use the new result type defined in OCaml >= 4.03
 while staying compatible with older version of OCaml should use the
 Result module defined in this library."""
-depends: ["ocaml" {< "4.14"}]
+depends: ["ocaml" {< "4.08"}]
 url {
   src: "https://github.com/janestreet/result/archive/1.1.tar.gz"
   checksum: "md5=1e0005783d4f70e1a9772723fa0e846b"
--- a/packages/result/result.1.2/opam
+++ b/packages/result/result.1.2/opam
@@ -14,7 +14,7 @@ description: """
 Projects that want to use the new result type defined in OCaml >= 4.03
 while staying compatible with older version of OCaml should use the
 Result module defined in this library."""
-depends: ["ocaml" {< "4.14"}]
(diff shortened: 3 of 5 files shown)


hannesm:
I do appreciate this change, though I think that @dra27 should jump in (from https://github.com/ocaml/opam-repository/issues/19880#issuecomment-954121920): "it's annoying having to add a conflict on a package which isn't (necessarily) related to you, but it is the "correct" thing here (lock files really do matter)." If I understand this correctly, @dra27 has insight into active usages of lockfiles that, if this PR is merged, will break all the users thereof.

mseri:
The point on the uninstallable packages is fair, although I would be curious to see if they are actually installable right now. So I am fine in closing this PR just for that.

Unfortunately using avoid-version requires a bound for opam >= 2.1, so this would also be breaking in some sense. But we can look into some other solution. The convenience of the lower bound check is not just the CI but also ensuring as much as we can that when people change their switch or downgrade something they can do it safely without breaking their switches.

Note that this PR is not discouraging to use result if somebody needs it: result 1.5 is compatible with all ocaml versions. I would be in favor of discouraging the use of older result packages, though: these are shadowing the result module and making all its functions unavailable, leading to nearly everything nowadays conflicting with it. Having an old result package in the lockfile is actually preventing any practical use of the result module, so it should probably be encouraged to update it to require result 1.5 instead, but that is not my decision to make for them.

hannesm:
I'm pretty disappointed by the discussion, I'm not motivated to contribute anything to this repository, being it that there are foggy "policies" that nobody cares to write down, and the argument being that there may be something that may break - instead of seeing the burden you impose on package authors.

So long and thanks for all the fish, opam-repository.

dra27:
> The point on the uninstallable packages is fair, although I would be curious to see if they are actually installable right now. So I am fine in closing this PR just for that.

They are indeed all installable (I verified both their installability and, as it was much quicker to, that the change definitely changed the installability before commenting) - that was how I discovered #24891.

> Unfortunately using avoid-version requires a bound for opam >= 2.1, so this would also be breaking in some sense. But we can look into some other solution. The convenience of the lower bound check is not just the CI but also ensuring as much as we can that when people change their switch or downgrade something they can do it safely without breaking their switches.

Adding `avoid-version` doesn't require a bound on opam - opam 2.0 ignores flags it doesn't understand. That's why we still have the ocaml-beta-repository dance for opam 2.0 with the compiler's beta versions, but opam 2.0 doesn't matter for the lowerbounds check (or in general - I don't think mccs in opam 2.0 is able to do the lowerbounds solves anyway, at least in the lifetime of the universe 🙂). Might be worth an experiment, given that (I expect) it's a lot easier to try than editing the solver directly.

> Note that this PR is not discouraging to use result if somebody needs it: result 1.5 is compatible with all ocaml versions. I would be in favor of discouraging the use of older result packages, though: these are shadowing the result module and making all its functions unavailable, leading to nearly everything nowadays conflicting with it. Having an old result package in the lockfile is actually preventing any practical use of the result module, so it should probably be encouraged to update it to require result 1.5 instead, but that is not my decision to make for them.

Oh, indeed - those (old) packages which are constrained to result < 1.5 have done that for a reason, though (1.5 is a breaking change for anything that assumed only `Result.result` was defined by the library). The main issue is that no package which requires 4.03+ can possibly _need_ to use the result library. We end up with these shims hanging around in codebases seemingly forever!

PR #24871, opened 2023-11-29
merlin is not compatible with OCaml 5.1.1

kit-ty-kate:
(needs to implement the new Compression module)
```
$ ocamlmerlin single locate -position 2:13 -verbosity 0 < main.ml | jq '.value'
"input_value: compressed object, cannot decompress"
```
Being fixed upstream in https://github.com/ocaml/merlin/pull/1714


mseri:
Thanks

PR #24881, opened 2023-12-01
[new release] elpi (1.18.1)

gares:
CHANGES:

Parser:
  - Remove legacy parsing engine based on Camlp5

API:
  - New `RawQuery.compile_ast`, lets one set up the initial state in which the query is run, even if the query is given as an ast.
  - New `solution.relocate_assignment_to_runtime` to pass a query result to another query
  - New `BuiltInPredicate.FullHO` for higher order external predicates
  - New `BuiltInPredicate.HOAdaptors` for `map` and `filter` like HO predicates
  - New `Calc.register` to register operators for `calc` (aka infix `is`)

Library:
  - New `std.fold-right`

Runtime:
  - New clause retrieval through discrimination tree. This new index is enabled whenever the `:index` directive selects only one argument  with a depth `> 1`.

### DIFF
--- a/packages/elpi/elpi.1.18.1/opam
+++ b/packages/elpi/elpi.1.18.1/opam
@@ -1,0 +1,89 @@
+opam-version: "2.0"
+maintainer: "Enrico Tassi <enrico.tassi@inria.fr>"
+authors: [ "Claudio Sacerdoti Coen" "Enrico Tassi" ]
+license: "LGPL-2.1-or-later"
+homepage: "https://github.com/LPCIC/elpi"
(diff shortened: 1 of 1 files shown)


gares:
Hum, CI is not running, maybe because I did open this PR and then close it?
Can someone start ci?

haochenx (approved):
I believe someone restarted the CI and most of the them were building fine. There were still a few transient failures so I restarted them, but I'm not concerned about them.

This PR looks mergeable to me.

mseri:
Thanks

PR #24916, opened 2023-12-07
Package rfsm.2.0

jserot:
### `rfsm.2.0`
A toolset for describing and simulating StateChart-like state diagrams

---
* Homepage: http://github.com/jserot/rfsm
* Source repo: git+https://github.com/jserot/rfsm.git
* Bug tracker: jocelyn.serot@uca.fr

---

### DIFF
--- a/packages/rfsm/rfsm.2.0/opam
+++ b/packages/rfsm/rfsm.2.0/opam
@@ -1,0 +1,38 @@
+opam-version: "2.0"
+synopsis:
+  "A toolset for describing and simulating StateChart-like state diagrams"
+maintainer: ["Jocelyn Sérot <jocelyn.serot@uca.fr>"]
+authors: ["Jocelyn Sérot <jocelyn.serot@uca.fr>"]
+license: "MIT"
+homepage: "http://github.com/jserot/rfsm"
+bug-reports: "jocelyn.serot@uca.fr"
+depends: [
+  "dune" {>= "3.11"}
+  "menhir" {>= "2.0"}
+  "ocamlgraph" {>= "2.0"}
+  "ppx_deriving" {>= "5.0"}
+  "ocaml" {>= "5.0.0"}
+  "odoc" {with-doc}
+]
+build: [
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
+dev-repo: "git+https://github.com/jserot/rfsm.git"
+url {
+  src: "https://github.com/jserot/rfsm/archive/refs/tags/2.0.tar.gz"
+  checksum: [
+    "md5=1b1942e9d27b5ffa6c7dbe3eb81f4f8e"
+    "sha512=fb6db5bb32d965db4584964f7170660163d745ca9f4cec46e61c0e2815bda636205278b019b2e2d7d43ab32fa9bfe8e642bc0d1e94ae22d792a3cd3d45731c37"
+  ]
+}


jserot:
Why is the build failing here ?
When invoking `opam publish`, i get 
```
$ opam publish .
The following will be published:
  - rfsm version 2.0 with opam file at /Users/jserot/Dev/ml/rfsm/rfsm.opam
    archive at https://github.com/jserot/rfsm/archive/refs/tags/2.0.tar.gz
```
If i download the archive file `https://github.com/jserot/rfsm/archive/refs/tags/2.0.tar.gz`, decompress it and invoke `dune build` in the resulting source tree, everything compiles perfectly.

jmid:
> Why is the build failing here ?

From the looks of https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/07b5feaf8f4a342aa4e0438c4338ff3a7c20643f it seems you are failing
- on bytecode-only architectures (32 bit arm+x86, as well as ppc64 which will have the native backend restored in 5.2)
- on FreeBSD

From the former and the looks of https://github.com/jserot/rfsm/blob/2.0/src/guests/std/bin/dune
```
(executable
  (modes byte native)
  (flags (:standard -annot) (:standard -w -26-27-37))
  (preprocess (pps ppx_deriving.show ppx_deriving.make))
  (libraries rfsm full)
  (public_name rfsmc))
```
I'm wondering how `(modes byte native)` is going to fare on a bytecode-only architecture?
The error message in the logs
```
# File "src/guests/std/bin/dune", line 6, characters 15-20:
# 6 |   (public_name rfsmc))
#                    ^^^^^
# Error: No rule found for src/guests/std/bin/rfsmc.exe
```
sounds as if it has a hard time building a native-code executable (.exe) on these.

jmid:
> Ok. Thanks. Does this mean that i have to limit the build to bytecode versions ?
> Or is there a way to tell opam not to install the native version ?

Is it important to offer both a bytecode build and a native when available?
My initial reaction would just be to omit the `(modes byte native)` then `dune` should take of the choice for you (by default building a native executable when available - and otherwise a bytecode executable AFAIR).

> Concerning the second error (on FreeBSD), i'm still wondering. The script complains about a missing file build_options_spec in directory src/host/lib but this file is present. It's a bash script file with permissions set to 0x755. What could prevent FreeBSD (and not other architectures) from executing it ?

I suspect this is the `/bin/bash` invocation failing from within the script.
I would try replacing it with `#!/usr/bin/env bash` as suggested here: https://forums.freebsd.org/threads/script-bash.72779/

Finally since this represents a silent dependency on `bash`, I encourage you to add a requirement of the `conf-bash` opam package.

jserot:
> Is it important to offer both a bytecode build and a native when available?

Well, not absolutely, in this particular case (i do not a native version of the compiler in order to integrate it to pre-built versions of another package, but i can do this on my local machine).

> My initial reaction would just be to omit the (modes byte native) then dune should take of the choice for you (by default building a native executable when available - and otherwise a bytecode executable AFAIR).

Seems a reasonable solution. Will try.

> I suspect this is the /bin/bash invocation failing from within the script.
I would try replacing it with #!/usr/bin/env bash as suggested here: https://forums.freebsd.org/threads/script-bash.72779/

Ok. The classical misleading error: this is not the script which is missing but a command _inside_ the script. Thx for pointing me to the forum message !

> Finally since this represents a silent dependency on bash, I encourage you to add a requirement of the conf-bash opam package.

Ok

Another question : do i have to close this PR and open a new one after making the above mentioned fixes or can i "restart" it (and if yes, how) ?

jmid:
> Another question : do i have to close this PR and open a new one after making the above mentioned fixes or can i "restart" it (and if yes, how) ?

No. If you add new commits and tag the new version as `2.0`, it should be sufficient to rerun your previous `opam publish` command. The result should show up as commits to this existing PR if all goes well :crossed_fingers:

mseri:
You can re-run opam publish. If you want we can also simply mark the package incompatible with bytecode-only compilers or specific architectures. As you prefer

jserot:
Thanks for help @mseri and @jmid. Just made the fixes. I replaced `/bin/bash` by `/bin/sh` in the `build_options_spec`script. Simpler and more portable.

jmid:
The 4.02 is indeed unrelated. I just pushed "rebuild failed" and now that passes too so everything is green :heavy_check_mark: :tada:

jserot:
Great. Thanks !

mseri:
Thanks!

PR #24917, opened 2023-12-07
Release of OCaml 5.1.1

Octachron:
This PR adds the various packages for OCaml 5.1.1:

- ocaml-base-compiler
- ocaml-system
- ocaml-variants.5.1.1+options
- ocaml-src

To unlock the dev version of the compiler it also adds tht new metapackage  for ocaml.5.12 and moves
ocaml.5.1.1+trunk to ocaml.5.1.2+trunk.


packages/ocaml-variants/ocaml-variants.5.1.2+trunk/opam
```diff
@@ -7,7 +7,7 @@ homepage: "https://ocaml.org"
 bug-reports: "https://github.com/ocaml/opam-repository/issues"
```
kit-ty-kate, line 7: removing 5.1.1+trunk will cause problem for the people not using gpatch on macOS. Could you instead use `available: false` and create a new package for 5.1.2+trunk?

Octachron: Done, thanks !

mseri:
Thanks

PR #24919, opened 2023-12-07
Ppxlib: disable tests for 0.30.0 on OCaml 5.1

pitag-ha:
There is a formatting reproducibility problem with those `ppxlib.0.30.0` tests, that test the behavior on OCaml 5.1. The `ppxlib.0.30.0` tests need therefore be disabled on 5.1, which is what this PR does. Funnily, the test conditions were written in a way that the tests don't run on OCaml 5.1.0, but do run on OCaml 5.1.1, so this has been noticed only now. See https://github.com/ocaml-ppx/ppxlib/issues/459.

### DIFF
--- a/packages/ppxlib/ppxlib.0.30.0/opam
+++ b/packages/ppxlib/ppxlib.0.30.0/opam
@@ -21,6 +21,7 @@ bug-reports: "https://github.com/ocaml-ppx/ppxlib/issues"
 depends: [
   "dune" {>= "2.7"}
   "ocaml" {>= "4.04.1" & < "5.2.0" & != "5.1.0~alpha1"}
+  "ocaml" {with-test & ( < "5.1.0" | >= "5.2.0" ) }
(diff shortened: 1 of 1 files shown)


mseri:
So this is broken only on 5.1.x?

pitag-ha:
The tests are broken whenever
```
 (enabled_if
  (and
   (>= %{ocaml_version} "5.1.0~alpha2")
   (< %{ocaml_version} "5.2.0")))
```
is enabled.

I've just changed the opam file to match exactly the upper bound of `5.1.0~alpha2`. Before, I had quickly put a strict `5.1.0` bound, which would unnecessarily also disable the tests on `5.1.0`. Thanks for noticing!

I've also removed the `| >= "5.2.0"` relaxation of that bound given that `>= "5.2.0"` can never be matched due to the general `ocaml` upper bound in ppxlib.0.30.0.

mseri:
Thanks

PR #24921, opened 2023-12-08
Add TSan backport switch: 5.1.1+tsan

OlivierNicole:
5.1.1 has been released and fixes a number of bugs and regressions, so it makes sense have a 5.1.1+tsan backport. The backport is 5.1.1 with the ThreadSanitizer PR backported, along with a few TSan bugfixes.

### DIFF
--- a/packages/ocaml-variants/ocaml-variants.5.1.1+tsan/files/ocaml-variants.install
+++ b/packages/ocaml-variants/ocaml-variants.5.1.1+tsan/files/ocaml-variants.install
@@ -1,0 +1,1 @@
+share_root: ["config.cache" {"ocaml/config.cache"}]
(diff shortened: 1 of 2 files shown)


OlivierNicole:
Hopefully this should be the last TSan backport!

tmcgilchrist:
@OlivierNicole This needs the clang changes from https://github.com/ocaml/opam-repository/pull/24816 to keep MacOS x86 support.

OlivierNicole:
Indeed. Fixed.

tmcgilchrist:
Looks good to me 👍🏻

mseri:
Thanks

mseri:
Do we need to add available only on x86?

tmcgilchrist:
That is more accurate, the support for other architectures is not available in the back port. 
See https://github.com/ocaml-multicore/ocaml-tsan/blob/5.1.1%2Btsan/configure#L16460-L16462

The previous back ports don't appear to have restricted the architecture to x86.

mseri:
The error message is perfectly clear, so I think there is no need for further changes

PR #24924, opened 2023-12-09
Package camlimages.5.0.5

camlspotter:
### `camlimages.5.0.5`
Image processing library
An image processing library, which provides loading and saving various image formats with an interface for the Caml graphics library. It has also an interface with the freetype library to draw texts using truetype fonts.

---
* Homepage: https://gitlab.com/camlspotter/camlimages
* Source repo: git+https://gitlab.com/camlspotter/camlimages.git
* Bug tracker: https://gitlab.com/camlspotter/camlimages/-/issues

---

### DIFF
--- a/packages/camlimages/camlimages.5.0.5/opam
+++ b/packages/camlimages/camlimages.5.0.5/opam
@@ -1,0 +1,51 @@
+opam-version: "2.0"
+synopsis: "Image processing library"
+description:
+  "An image processing library, which provides loading and saving various image formats with an interface for the Caml graphics library. It has also an interface with the freetype library to draw texts using truetype fonts."
+maintainer: "Jun FURUSE <jun.furuse@gmail.com>"
+authors: "Jun FURUSE <jun.furuse@gmail.com>"
+license: "LGPL-2.1-only WITH OCaml-LGPL-linking-exception"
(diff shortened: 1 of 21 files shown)


camlspotter:
Checking the CI failures...

camlspotter:
Analysis of CI failures:

* Old liquidsoap versions: the latest liquidsoap compiles.
* advi.2.0.0 in OCaml 5.1 : It fails to compile due to the use of deprecated Stdlib.Printexc.catch.  Not the issue of camlimages.
* distributions: linker errors due to lack of -fPIC

Not sure what is the best option to handle these "-fPIC" flags...

haochenx (commented):
The only thing concerns me is the revdeps of liquidsoaps. I have created a PR that should fix the failures so please take a look when you have time: https://github.com/camlspotter/opam-repository/pull/2

For the distribution failures I'm afraid that there's not too much thing we can do about. You can look at the `x-ci-accept-failures` option if you want to reduce CI noises but it's not mandatory.

As you mentioned advi is unrelated to this PR so I don't think you have to do anything about it.

haochenx (approved):
Seems like now there are only the distro failures that we cannot do anything about and the advi failure that is not related to this package.

Looks good to me for merging!

mseri:
Thanks a lot!

PR #24942, opened 2023-12-14
ott: fix for building on Windows (without symlink)

MSoegtropIMC:
This PR merges the Windows fix from the Opam MinGW repo:

https://github.com/ocaml-opam/opam-repository-mingw/tree/sunset/packages/ott/ott.0.32

and adds a correspnding fix for ott.0.33.

For older versions of ott the MinGW repo has no patches - not sure if these work or not, I didn't test it.

The change does not influence builds on other systems, so I did not change the version numbers.

### DIFF
--- a/packages/ott/ott.0.32/opam
+++ b/packages/ott/ott.0.32/opam
@@ -14,6 +14,8 @@ depends: [
 build: [
   [make "world"]
   [make "ott.install"]
+  ["rm" "src/ott"] { os = "win32" }
+  ["cp" "src/ott.opt" "src/ott"] { os = "win32" }
 ]
 run-test: [
   [make "-C" "tests/menhir_tests/test_if"]
--- a/packages/ott/ott.0.33/opam
+++ b/packages/ott/ott.0.33/opam
@@ -15,6 +15,9 @@ depends: [
 build: [
   [make "world"] { ocaml:native }
   [make "world.byt"] { !ocaml:native }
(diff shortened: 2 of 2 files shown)


MSoegtropIMC:
@dra27 : after this is merged, should I do a PR to the MinGW sunset repo and remove ott from there? Afair we said that the sunset repo should only be used on top of the main repo and that we want to move everything gradually to the main repo. Such updates look like a good opportunity to do this. Maybe we should leave a readme file in the ott folder of the sunset repo.

MSoegtropIMC:
@dra27 : can I please have a review of this asap - otherwise I have to keep this as local patch in Coq Platform (release is in progress).

mseri:
Since this is only affecting windows and is used in production by coq, I am going to take over the decision and merge it in the meantime. We can always revisit if needed

MSoegtropIMC:
@mseri : thanks, I appreciate it:

@dra27 : the question remains, if I should do a PR to remove this project entirely from the MinGW sunset repo - IMHO the best strategy is to make project by project work on Windows with the main opam repo and delete them from the MinGW repo - until it is empty. This of course requires that everybody uses the MinGW repo as overlay to the main repo - nut sure if this is the communicated policy.

As indicated above, I think we should replace the ott project in the MinGW repo with a Readme file stating that it now lives in the main repo + a link with an explanation of the repo overlaying.

PR #24945, opened 2023-12-15
Num version 1.5

xavierleroy:
Cc @dra27 .

### DIFF
--- a/packages/base58/base58.0.1.2/opam
+++ b/packages/base58/base58.0.1.2/opam
@@ -12,13 +12,12 @@ build: [
 install: [make "install"]
 remove: ["ocamlfind" "remove" "base58"]
 depends: [
-  "ocaml" {>= "4.02.0"}
+  ("ocaml" {>= "4.02.0" & < "5.0"} & "num" | "ocaml" {>= "5.0"} & "num" {< "1.5"})
   "ocamlfind" {build}
   "testsimple" {with-test}
   "bisect" {with-test}
   "bisect_ppx" {with-test}
   "ocamlbuild" {build}
-  "num"
 ]
 synopsis: "Base58 encoding and decoding"
 description:
--- a/packages/cubicle/cubicle.1.2.0/opam
+++ b/packages/cubicle/cubicle.1.2.0/opam
@@ -13,9 +13,8 @@ build: [
   [make]
 ]
 depends: [
-  "ocaml" {>= "4.08.0"}
+  ("ocaml" {>= "4.08.0" & < "5.0"} & "num" | "ocaml" {>= "5.0"} & "num" {< "1.5"})
   "ocamlfind"
-  "num"
 ]
 depopts: ["functory"]
 conflicts: [
--- a/packages/num/num.1.5/opam
+++ b/packages/num/num.1.5/opam
@@ -5,7 +5,7 @@ authors: [
   "Pierre Weis"
   "Xavier Leroy"
 ]
-license: "LGPL-2.1-only with OCaml-LGPL-linking-exception"
+license: "LGPL-2.1-only WITH OCaml-LGPL-linking-exception"
 homepage: "https://github.com/ocaml/num/"
 bug-reports: "https://github.com/ocaml/num/issues"
 dev-repo: "git+https://github.com/ocaml/num.git"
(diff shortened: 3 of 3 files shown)


xavierleroy:
For the record: the linter complains that "The package has a dune-project file but no explicit dependency on dune was found.", but I suspect this dune-project file is not used for building the OPAM package, and adding a dependency from Num to Dune just to silence this warning is dubious.  What's the correct way to deal with this situation?

avsm:
We generally just ignore that warning for these projects that have a good reason why. (although it can be scary to see a big red X -- the UI here can certainly be improved!)

dra27:
I’ve given opam-repo-ci a gentle kick, as a number of the revdeps failures looked transient. When it’s done, I’ll have a look through the other failures to check that there’s nothing strictly to do with num (obviously anyone else is welcome to check, too!). The opam packaging of num got quite an overhaul in this release, so the revdeps testing is more relevant than it might normally be.

dra27 (approved):
The failures on 5.1 for `alt-ergo.2.0.0`, `alt-ergo-free.2.0.0` and `tplib.1.3` are not introduced by this PR (see recent opam-health-check runs).

`base58.0.1.2` and `cubicle.1.2.0` both fail because of a dependence on the num library being installed into the Standard Library. The metadata for base58 isn't even correct; cubicle has had an issue for this for a loong time (https://github.com/cubicle-model-checker/cubicle/issues/1). I've directly pushed an amendment to those packages to constrain them away from num.1.5 when installed on OCaml 5.0+.

Otherwise this looks good to merge, for me.

xavierleroy:
Thanks a lot @dra27 and @avsm for your help!

I hadn't realized until now that a submission for a new package could fix opam files for existing packages.  That's neat!

Re: opam-lint, would there be interest in adding mechanisms to control it better, such as a field in the .opam file to specify opam-lint warnings to turn on or off?

dra27:
> Re: opam-lint, would there be interest in adding mechanisms to control it better, such as a field in the .opam file to specify opam-lint warnings to turn on or off?

This particular warning doesn't come from `opam lint` but from additional linting that opam-repo-ci is doing. The culprit is `Lint.check_dune_constraint` in [ocurrent/opam-repo-ci/lib/lint.ml#L163-L174](https://github.com/ocurrent/opam-repo-ci/blob/afc49aeee32e5807c2c248b0e411d808b204e4b2/lib/lint.ml#L163-L174):

```ocaml
  let check_dune_constraints ~errors ~pkg opam =
    match opam.OpamFile.OPAM.url with
    | Some url ->
        get_dune_project_version ~pkg url >|= fun dune_version ->
        let is_build, dune_constraint = get_dune_constraint opam in
        let errors =
          match dune_constraint, dune_version with
          | _, Error msg -> (pkg, FailedToDownload msg) :: errors
          | None, Ok None -> errors
          | Some _, Ok None -> (pkg, DuneProjectMissing) :: errors
          | None, Ok (Some _) ->
              if is_dune (OpamPackage.name pkg) then
                errors
              else
                (pkg, DuneConstraintMissing) :: errors
```

There is already an exception for Dune itself (that's the `is_dune` part towards the end). This check could be enhanced to look at `opam.OpamFile.OPAM.build` for actual calls to `"dune"`, and so suppress the warning when it finds `dune-project`, but can see that `dune` is never called by the package's build instructions. That would also fix the linting errors for opam-file-format (which has an optional dependency on Dune) and also the compiler, e.g. [5.1.1's linting](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/37d7bd839662f9476660d1b12c0ff300682a8c44/variant/(lint)):

```console
2023-12-07 15:42.05: Warning in ocaml-variants.5.1.2+trunk: The package has a dune-project file but no explicit dependency on dune was found.
2023-12-07 15:42.05: Warning in ocaml-variants.5.1.1+options: The package has a dune-project file but no explicit dependency on dune was found.
2023-12-07 15:42.05: Warning in ocaml-src.5.1.1: The package has a dune-project file but no explicit dependency on dune was found.
2023-12-07 15:42.05: Warning in ocaml-base-compiler.5.1.1: The package has a dune-project file but no explicit dependency on dune was found.
```

mseri:
Thanks, I'll open an issue on the CI repository for the linter. This looks good to merge.

PR #24951, opened 2023-12-17
[new release] mirage-flow, mirage-flow-unix and mirage-flow-combinators (4.0.0)

hannesm:
Flow implementations and combinators for MirageOS

- Project page: <a href="https://github.com/mirage/mirage-flow">https://github.com/mirage/mirage-flow</a>
- Documentation: <a href="https://mirage.github.io/mirage-flow/">https://mirage.github.io/mirage-flow/</a>

##### CHANGES:

- Add ``shutdown : flow -> [ `read | `write | `read_write ] -> unit Lwt.t``
  (@djs55 @hannesm mirage/mirage-flow#16 mirage/mirage-flow#18 mirage/mirage-flow#48)
- Remove SHUTDOWNABLE signature (@djs55 mirage/mirage-flow#17, rebased into mirage/mirage-flow#48)

PR #24952, opened 2023-12-17
restrict tcpip and tls-mirage for mirage-flow < 4.0.0

hannesm:
first PR for #24951 (there may be more to come)


hannesm:
FWIW, the changes are mostly trivial, and I won't fix any revdeps or other compilation issues (the packages were in the same way broken before, this is only to allow a new release of mirage-flow which will be incompatible with the old releases). thanks for considering this PR.

mseri:
Thanks

PR #24961, opened 2023-12-19
Mark let-if and ssh-agent-unix unavailable

reynir:
`let-if` was a fun experiment, but I don't think it's seen any usage and I now think it shouldn't be used.

`ssh-agent-unix` was not meant to be published, but was published sort of by accident due to dune-release picking it up. Previous versions have been of poor quality, and I don't think it's very useful beyond small examples. I would also like to not have to maintain this package (still want to maintain ssh-agent though!)

See also the discussion in https://github.com/ocaml/opam-repository/issues/23789#issuecomment-1862703138


mseri:
I am letting this wait a moment to see if there are users.
There are surely no published packages using them
```
❯ opam list --depends-on=let-if
# Packages matching: depends-on(let-if) & (installed | available)
# No matches found
❯ opam list --depends-on=ssh-agent-unix
# Packages matching: depends-on(ssh-agent-unix) & (installed | available)
# No matches found
```

mseri:
No complain here, I am going on with it

PR #24970, opened 2023-12-21
[new release] tty (0.0.2)

leostera:
A library for interacting with teletype and terminal emulators

- Project page: <a href="https://github.com/leostera/tty">https://github.com/leostera/tty</a>

## 0.0.2

* fix: use select to check for reads, instead of setting the stdin fd to
  non-blocking. This caused issues on Mint Tea's rendering of many emojis or
  colored symbols.

* chore: drop dune as a dependency to fix .opam files


mseri:
No problem, and it was already closed so there is no confusion

PR #24973, opened 2023-12-21
Release of Esperanto (v0.0.5)

dinosaure:
- Support Cosmopolitan 3.1 (dinosaure/esperanto#43, @dinosaure)
- Fix the compilation of esperanto with setup-ocaml and GitHub CI (dinosaure/esperanto#44, @dinosaure)
- Extend the support of Esperanto to 4.13 (dinosaure/esperanto#45, @dinosaure)

### DIFF
--- a/packages/aarch64-esperanto/aarch64-esperanto.0.0.5/opam
+++ b/packages/aarch64-esperanto/aarch64-esperanto.0.0.5/opam
@@ -23,5 +23,5 @@ description:
   "An OCaml compiler (toolchain) with Cosmopolitan as the C library"
 url {
   src: "https://github.com/dinosaure/esperanto/releases/download/v0.0.5/esperanto-v0.0.5.tar.gz"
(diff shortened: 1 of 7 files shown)


dinosaure:
- [ ] move the `dune-project` file to the root of the distribution
- [ ] see what happens between `opam.2.0` and `opam.2.1` about the sandboxing and how we tried to install `esperanto-cosmopolitan`

dinosaure:
The error on MacOS seems unrelated. The linter is a bit restrictive: currently we have a `dune-project` file only for `esperanto-cosmopolitan`. `{x86_64,aarch64}-esperanto` don't use `dune` but due to the presence of the `dune-projet`, it seems that we should require `dune` according to the linter (which is not true).

mseri:
So aarch64-esperanto can build on x86 linux but not on macos x86 or arm64, is that still correct?

dinosaure:
Yes, we probably can build everything on any `x86_64` systems (like mac) but the distribution comes with several shell scripts which are used then to compile OCaml source code. I prefer to be restrictive/conservative on this topic (and trust only on a linux machine) instead of "accidentally" support something new and unexpected :+1:.

mseri:
Thanks for the clarificAtion

PR #24976, opened 2023-12-21
Mark ocamlbuild.0.14.3 as not available on Windows

dra27:
Not sure whether we should immediately update the Windows patched version, given that the fix in 0.14.3 is only for trunk/5.2, but regardless this package doesn't build on Windows.


dra27:
Applying the same logic as in https://github.com/ocaml/opam-repository/pull/24797#issuecomment-1813442287, merging this now that the linter has passed.

PR #24977, opened 2023-12-21
Add mlmpfr package update for version 4.2.0

thvnx:


mseri:
Succeeds where it can. I think some distributions would be fixable providing an inclusion and a linker path.

mseri:
Thanks

thvnx:
> Succeeds where it can. I think some distributions would be fixable providing an inclusion and a linker path.

I'll have a look for a future release, thanks!

PR #24980, opened 2023-12-22
Add new version 3.0 for the E-prover automatic theorem prover

MSoegtropIMC:
This PR adds the new version 3.0 of the E-prover automatic theorem prover.

The main difference to version 2.6 is the support for higher order logic (2.6 was first order logic only).

E-prover is written in C++, but used by software written in OCaml, notably automation plugins for the Coq theorem prover.


mseri:
I don't know why freebsd fails, but it is not strictly necessary that you look into it if you don't intend to support the platform

MSoegtropIMC:
@mseri : it is some weird make error - I would have to instal FreeBSD to check. IMHO it is OK to merge it as is, but I will follow up with the author.

mseri:
Thanks

PR #24984, opened 2023-12-22
[new release] fix atacama 0.0.2 and 0.0.3 dependencies on riot

leostera:

## 0.0.6

* Redo packaging to expose a single public library: `riot`
* Fix issue with schedulers busy-waiting
* Introduce separate IO Schedulers for polling
* Switch to `poll` to support kqueue on macOS
* Reuse read-buffers on Io.read loops
* Broaden IO socket types to file descriptors
* Improved polling with shorter poll timeouts and safety checks
* Add `Dashmap.iter` to iterate over a collection
* Add `net_test` with an echo tcp server/client
* Fix bugs with syscall suspension that was introduced with reduction counting


dmmulroy:
lgtm

mseri:
Can you add upper bounds to atacama 0.0.2 and 0.0.3? They are not compatible with this release and I won’t be at my laptop for a while to do it myself

there are a number of strange thread allocation failures that right now may or may not be CI related, I am rerunning the jobs to check

mseri:
Btw I think it may be the same failure that we saw here: https://github.com/ocaml/opam-repository/pull/24306

leostera:
@mseri thanks for bringing this up! 🙏🏼 I've updated the atacama 0.0.2/0.0.3 opam files to have a fixed version of riot instead. Should be dandy now :) 

Thanks and Happy Holidays! ✨

mseri:
The tests are reliably timing out or failing to allocate domains (so exhausting the domains). The timeout is usually related to some deadlocks, can you double check if you can reproduce them? 

you can use the top of the ci message to rerun the same job locally via docker (except for macos)

leostera:
Looking into it!

leostera:
I'm on a stream now and we managed to replicate IF i ask Riot to start ~125 threads per test on a 10-core MBP M1 (which is crazy but could happen if you have 125 cores).

Is this a limitation of the opam-ci? How many threads are available per package execution?

These tests should be spinning one domain per core by default, and I'll keep 'em that way to exercise parallelism and not just concurrency.

What would you recommend here? I'm running CI on my own as well, so I could modify the tests behavior if I find `OPAM_REPO_CI` (for ex. try running a single thread or skip the ones that need multiple threads).

Here's Riot's CI passing on ubuntu/macos (the only platforms i care about atm) – https://github.com/leostera/riot/actions/runs/7302019400/job/19899831184

mseri:
I think it may have to do with the hardcoded number of max domains in the compiler: https://github.com/ocaml/ocaml/blob/2433ab38e358dab153380a09ecca712ce14df5d8/runtime/caml/domain.h#L34 sounds similar to the issue that @c-cube was facing in the PR linked above

If this is the case, we can probably merge this, but perhaps it may be worth noting the failure down somewhere in case some user stumbles upon it

mseri:
Thanks a lot for checking in any case!

mseri:
In moonpool case, the issue was that there may be for a small time more than `recommended_domain_count()-1` domains active at a time. The issue was resolved by joining to ensure cleanup before re-using domains: https://github.com/c-cube/moonpool/pull/9

Indeed the failures appear on CI machines with > 128 cores and where the hard limit can be easily reached (if I am not mistaken)

leostera:
Found something after reading your comments, and I think we *are* going over the limit, since we have a dedicate IO domain now.

This is fixed here: https://github.com/leostera/riot/commit/1164315f0744b0391856a2d327546708b80d8b24 – but there's still lots of unreleased changes that need some testing, so we may wanna put 0.0.6 in the fold, and I'll open one for 0.0.7 sometime soon.

leostera:
Superseded by https://github.com/ocaml/opam-repository/pull/24991

PR #24991, opened 2023-12-25
[new release] riot (0.0.7)

leostera:

## 0.0.7

* Introduce IO module with low-level IO operations such as performing direct
  vectorized (or regular) reads/writes. New operations include:
  * `read`, `write`
  * `single_read`, `single_write` (vectorized)
  * `await_readable`, `await_writeable`, `await`
  * `write_all`
  * `copy` and `copy_buffered`

* Introduce Buffer module with support for converting from and to CStruct and
  String, including position tracking.

* Introduce Read/Reader interface for creating buffered and unbuffered readers
  of arbitrary sources.

* Introduce Write/Writer interfaces for creating unbuffered writers into
  arbitrary destinations/sinks.

* Introduce File module with Reader and Writer implementations

* Implment Reader and Writer interfaces for Net.Socket

* Dropped dependency on Bigstringaf and moved to Cstruct

* Fix max number of domains to always be under the recommended domain count

* Fix issue with tests where the runtime idled after the main would die. Now
  the main process finishing with an exception is considered reason enough to
  shutdown the system.

* Refactor tests to always output `test_name: OK` when everything is fine and
  all modules to end in `_test`.

* Add several IO tests.

## 0.0.6 (included since we didn't publish 0.0.6)

* Redo packaging to expose a single public library: `riot`
* Fix issue with schedulers busy-waiting
* Introduce separate IO Schedulers for polling
* Switch to `poll` to support kqueue on macOS
* Reuse read-buffers on Io.read loops
* Broaden IO socket types to file descriptors
* Improved polling with shorter poll timeouts and safety checks
* Add `Dashmap.iter` to iterate over a collection
* Add `net_test` with an echo tcp server/client
* Fix bugs with syscall suspension that was introduced with reduction counting

### DIFF
--- a/packages/riot/riot.0.0.7/opam
+++ b/packages/riot/riot.0.0.7/opam
@@ -1,0 +1,45 @@
+opam-version: "2.0"
+synopsis: "An actor-model multi-core scheduler for OCaml 5"
+description:
+  "Riot is an actor-model multi-core scheduler for OCaml 5. It brings Erlang-style concurrency to the language, where lighweight process communicate via message passing"
+maintainer: ["Leandro Ostera <leandro@abstractmachines.dev>"]
+authors: ["Leandro Ostera <leandro@abstractmachines.dev>"]
+license: "MIT"
+tags: ["multicore" "erlang" "actor" "message-passing" "processes"]
+homepage: "https://github.com/leostera/riot"
+bug-reports: "https://github.com/leostera/riot/issues"
+depends: [
+  "cstruct" {>= "6.2.0"}
+  "mdx" {with-test & >= "2.3.1"}
+  "ocaml" {>= "5.1"}
+  "odoc" {with-doc & >= "2.2.2"}
+  "poll" {>= "0.3.1"}
+  "ptime" {>= "1.1.0"}
+  "telemetry" {>= "0.0.1"}
+  "uri" {>= "4.4.0"}
+  "dune" {>= "3.11"}
+]
+build: [
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
+dev-repo: "git+https://github.com/leostera/riot.git"
+url {
+  src:
+    "https://github.com/leostera/riot/releases/download/0.0.7/riot-0.0.7.tbz"
+  checksum: [
+    "sha256=b7e3cc061e2b6578bcd9d5258efde72f3657e68d626a005b43d15f1a7aff769e"
+    "sha512=00dcd34cce21058137cca15229a81e0e2c898f9877f22db6903288c8f757847a65965cfe791f84fe251df342f90e7e4aefe04f2fbb6ba282ad45b1a3b979e6cf"
+  ]
+}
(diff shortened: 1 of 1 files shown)


leostera:
@mseri so this time the test that seem to consistently fail are some of the I/O heavy ones:

* run a TCP client/server
* resolve domain names
* create large buffers (~1gb string) and flush it to disk

What constraints am I working with here? 🤔

mseri:
I donk’t know the bounds on the ci, but given the amount if time the tests run, I would separate the tests into normal tests and heavy tests. I would definitely avoid running on the CI the ones writing on the disk.

Keep also in mind that some sandboxes, e.g. mac, are more restrictive in the use of network in the tests (the tests are run via `opam install -t riot` and so will run in the sandbox).

We can also disable the tests and call it a day but in general I think it is worth running tests in the opam repo ci, I think the extra checks do pay off on the long run

mseri:
Test on macos are failing ebcause the sandbox is quite restrictive on mac and they are not allowed to bind.
```
#=== ERROR while compiling riot.0.0.7 =========================================#
# context              2.2.0~alpha4~dev | linux/x86_64 | ocaml-base-compiler.5.1.1 | pinned(https://github.com/leostera/riot/releases/download/0.0.7/riot-0.0.7.tbz)
# path                 ~/.opam/5.1/.opam-switch/build/riot.0.0.7
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p riot -j 71 @install @runtest
# exit-code            1
# env-file             ~/.opam/log/riot-7-60183c.env
# output-file          ~/.opam/log/riot-7-60183c.out
### output ###
# (cd _build/default/test && ./selective_receive_test.exe)
# 19:54:56.27854Z[info] selective_receive: OK
# (cd _build/default/test && ./add_monitor_test.exe)
# 19:54:56.26206Z[info] add_monitor: OK
# File "test/dune", line 56, characters 7-24:
# 56 |  (name net_addr_uri_test)
#             ^^^^^^^^^^^^^^^^^
# (cd _build/default/test && ./net_addr_uri_test.exe)
# 2023-12-26 19:54:56.39695Z ERROR [thread=0] Process <0.1.0> died with unhandled exception Invalid_argument("option is None"):
# Raised at Stdlib.invalid_arg in file "stdlib.ml", line 30, characters 20-45
# Called from Stdlib__Option.get in file "option.ml" (inlined), line 21, characters 41-69
# Called from Dune__exe__Net_addr_uri_test.main in file "test/net_addr_uri_test.ml", line 6, characters 13-77
# Called from Runtime__Import._spawn.(fun) in file "riot/runtime/import.ml", line 92, characters 10-15
# 
# (cd _build/default/test && ./net_reader_writer_test.exe)
# 19:54:56.69135Z[info] net_reader_writer_test: OK
# File "test/dune", line 66, characters 7-15:
# 66 |  (name net_test)
#             ^^^^^^^^
# (cd _build/default/test && ./net_test.exe)
# 19:54:56.83514Z[error] client unix error Connection reset by peer
# 19:54:56.83530Z[error] net_test: bad payload: "empty paylaod"
```
This other failure seems more common

leostera:
Okay I think it may make sense to split up our tests then, I'll put all the I/O heavy ones in a separate tests package that we won't publish.

mseri:
The issue on freebsd is independent of this, thanks!

PR #25002, opened 2023-12-30
New LLVM 16 release to fix Homebrew installing zstd in nonstandard location

alan-j-hu:
Should fix the issue at https://github.com/alan-j-hu/llvm-dune/issues/18. @jordydickinson do you think you could test this out and make sure it works for you?

### DIFF
--- a/packages/llvm/llvm.16.0.6+nnp+fix-homebrew/opam
+++ b/packages/llvm/llvm.16.0.6+nnp+fix-homebrew/opam
@@ -1,0 +1,36 @@
+opam-version: "2.0"
+synopsis: "The OCaml bindings distributed with LLVM"
+description: "Note: LLVM should be installed first."
+maintainer: "Alan Hu <alanh@ccs.neu.edu>"
+authors: [
+  "Alan Hu <alanh@ccs.neu.edu>"
+  "Kate <kit-ty-kate@outlook.com>"
+  "Gordon Henriksen <gordonhenriksen@mac.com>"
+  "whitequark <whitequark@whitequark.org>"
+  "The LLVM team"
+]
+license: "Apache-2.0 WITH LLVM-exception"
+homepage: "http://llvm.moe"
+doc: "http://llvm.moe/ocaml"
+bug-reports: "http://llvm.org/bugs/"
+depends: [
+  "ocaml" {>= "4.00.0"}
+  "dune" {>= "2.7"}
+  "ctypes" {>= "0.4"}
+  "conf-llvm" {build & = "16"}
+]
+build: [
+  ["./setup.sh" conf-llvm:config]
+  ["dune" "build" "--release" "-j" jobs]
+  ["rm" "%{name}%.install"]
+]
+install: ["./install.sh" prefix]
+dev-repo: "git+https://github.com/alan-j-hu/llvm-dune.git"
+url {
+  src:
+    "https://github.com/alan-j-hu/llvm-dune/releases/download/v16.0.6%2Bnnp%2Bfix-homebrew/llvm-dune-full-minified-16.0.6-fix-homebrew.tar.gz"
+  checksum: [
+    "md5=4c242a4e6a9d76b2ca493b2f14eab3ce"
+    "sha512=cc618295f1a618fdb67b65fc65d1bc40e856ad2cb0200e43573776235dca6cab"
+  ]
+}


jordydickinson:
How do I test this?

alan-j-hu:
Can you try `opam repo add llvm-homebrew https://github.com/alan-j-hu/opam-repository.git#llvm-16-fix-homebrew`? Let me know if this doesn't work.

jordydickinson:
It installs with no errors.

alan-j-hu:
The problem which this release fixes involves the interaction of multiple factors between Apple, Homebrew, and LLVM:

Apple's previous Macs used Intel, but its newest Macs use ARM (the Mac M1 and onward). To ease the transition, Apple has an emulation layer called Rosetta that can execute Intel binaries on ARM. Therefore, Intel and ARM binaries may for now coexist on the new ARM-based Macs. The older versions of Homebrew installed packages in the "standard" location, `/usr/local`. However, in the interest of having Intel binaries and ARM binaries coexist, Homebrew began installing packages compiled for ARM in `/opt/homebrew`. This change is discussed at https://github.com/Homebrew/brew/issues/9177. Another reason for the change cited in the discussion is the desire for Homebrew not to install packages in the "standard" location `/usr/local`, where they may collide with other installations. However, this reason seems to be an ex-post-facto justification, as this change in fact breaks other software that expects to find packages in the standard location.

As a security feature on Apple, child processes ignore the `LD_LIBRARY_PATH` variable. There is no way to change the default path for searching for libraries.

The `llvm-config` tool outputs the necessary flags to compile and link programs using the LLVM libraries. The [flags for system libraries are hardcoded](https://github.com/llvm/llvm-project/blob/7cbf1a2591520c2491aa35339f227775f4d3adf6/llvm/tools/llvm-config/llvm-config.cpp#L736-L740) and `llvm-config` tool does not perform any search for packages. The linker will see the `-lzstd` flag and look for the shared object file in the standard locations, but will not find the Homebrew-installed zstd package on ARM-based Macs.

The Homebrew "recipe" for zstd [alters the path](https://github.com/Homebrew/homebrew-core/blob/53abc41dc5ae17bb7616eac4f74474859590ed51/Formula/z/zstd.rb#L55) in the `.pc` file, allowing `pkg-config` to find the location of `zstd`.

So, the causes of the issue are as follows:
- Apple switching from Intel to ARM for its Macs
- Homebrew installing packages to a nonstandard location on ARM-based Macs so Intel binaries and ARM binaries can coexist
- Apple not providing a way to change the default path for looking up library files for security reasons
- `llvm-config` hardcoding the `-lzstd` flag and not performing any logic to look for system dependencies
  - `pkg-config` can find the zstd package because Homebrew changes the prefix in the `.pc` file

I feel that opam packages generally expect Homebrew to work out of the box, treating it as an ordinary package manager, as `os-distribution = "homebrew"` is a supported filter for depexts. The fact that Homebrew is now installing packages in a non-standard location is very inconveniencing, and I wonder if this change breaks other opam packages as well.

It's not clear to me which party is ultimately "responsible" for fixing this issue, but I'm going ahead and fixing the issue in this package by explicitly looking up the zstd package using `pkg-config`, adding a dependency on `conf-pkg-config`. I'm curious what the opam-repository maintainers think about this issue.

Please let me know if my explanation is unclear or incorrect.

alan-j-hu:
Hey, it's been several months since I opened this PR and there has been no action from the opam-repository maintainers. I want to understand what they think of this PR, and of the changes on Mac and Homebrew that broke stuff,

samoht:
Apologies - this is a tricky package, and I suppose none of the maintainers have time to understand the various issues at play. I think that would help if you could review the CI failures (in https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/4d9fad21a5bc93fd3b1cd19733ce12ee7f259505) and report if these errors are expected or not.

alan-j-hu:
@samoht The compile errors for the binsec package are because of the change from the typed to opaque pointer API. Its LLVM dependency needs to have a version bound of < 16.

The other errors relate to failures to install the system LLVM package in some way or another.

alan-j-hu:
The binsec issue is addressed in https://github.com/ocaml/opam-repository/pull/25667. That PR should be merged first, because it's straightforward while this one may involve some consideration.

shonfeder:
Hello @alan-j-hu! I'm just circling around to see if you think we may be in a position to land this, or at least whether we should put this pack in the queue of PRs needing maintainer attention, now that the PR you noted in https://github.com/ocaml/opam-repository/pull/25002#issuecomment-2052705536 seems to be merged. Let me know what you think!

PR #25018, opened 2024-01-03
conf-libssl: Add support for distributions derived from fedora

kit-ty-kate:


arbipher:
The opam lint complains _Package is flagged "conf" but has source, install or remove instructions_.

kit-ty-kate:
this warning can be ignored. The `conf` flag doesn't have any effect afaik

mseri:
Thanks

PR #25043, opened 2024-01-07
Add missing dependency of menhir on menhirCST.

fpottier:


mseri:
Surprising that it didn't fail in the CI when it was submitted

mseri:
Looks good, thanks

fpottier:
I think that it did not fail because Menhir can be built and installed even if `menhirCST` is absent. The library `menhirCST` is needed only for some applications of Menhir. But I prefer to declare a dependency anyway. This way, end users need to know only about the package `menhir` and do not need to know that this package is itself split into several sub-packages.

PR #25047, opened 2024-01-08
Add a conf-ninja virtual package

AltGr:


AltGr:
Ok I could get the depexts right for every tested distro but for oracle-linux, where the hell can the list of packages for that be found ?

mseri:
Don't worry about oracle linux, we are planning to remove it from the ci matrix

AltGr:
Ok thanks!

PR #25056, opened 2024-01-09
[new release] trace (4 packages) (0.6)

c-cube:
A stub for tracing/observability, agnostic in how data is collected

- Project page: <a href="https://github.com/c-cube/ocaml-trace">https://github.com/c-cube/ocaml-trace</a>

##### CHANGES:

- add `ppx_trace` for easier instrumentation.
  * `let%trace span = "foo" in …` will enter a scope `span` named "foo"
  * `let%trace () = "foo" in …` will enter a scope named "foo" with a hidden name
- add `trace-fuchsia` backend, which produces traces in the binary format
    of [fuchsia](https://fuchsia.dev/fuchsia-src/reference/tracing/trace-format).
    These traces are reasonably efficient to produce (~60ns per span on my machines)
    and reasonably compact on disk, at least compared to the TEF backend.

### DIFF
--- a/packages/ppx_trace/ppx_trace.0.6/opam
+++ b/packages/ppx_trace/ppx_trace.0.6/opam
@@ -1,0 +1,42 @@
+opam-version: "2.0"
+synopsis: "A ppx-based preprocessor for trace"
+maintainer: ["Simon Cruanes"]
+authors: ["Simon Cruanes"]
+license: "MIT"
+tags: ["trace" "ppx"]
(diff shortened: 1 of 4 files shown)


c-cube:
… do we have a `config-littleendian` package in opam? :joy:

mseri:
Should we mark it unavailable on s390x?

c-cube:
Sure, yeah. In general it assumes little endian arch.

c-cube:
This should be ready now!

PR #25065, opened 2024-01-11
Release raylib-ocaml 1.2.0 (raylib 5.0.0)

tjammer:
* Upgrade raylib to 5.0.0
* Make Raylib.Matrix.t usable with Rlgl


tjammer:
The freebsd build shouldn't fail, I need to fix this

avsm:

Issue #25075, opened 2024-01-14
Installing conf-openblas 0.2.1 needs PKG_CONFIG_PATH extended.


tmcgilchrist:
Installing the `conf-openblas` package on macOS Sonoma ARM fails with:

```
#=== ERROR while compiling conf-openblas.0.2.1 ================================#
# context     2.1.5 | macos/arm64 | ocaml-base-compiler.5.0.0 | https://opam.ocaml.org#7600b9b6
# path        ~/code/ocaml/par-matmul-profiling/_opam/.opam-switch/build/conf-openblas.0.2.1
# command     ~/.opam/opam-init/hooks/sandbox.sh build sh -exc cc $CFLAGS $(pkg-config --cflags openblas) test.c $(pkg-config --libs openblas)
# exit-code   1
# env-file    ~/.opam/log/conf-openblas-60410-e8868b.env
# output-file ~/.opam/log/conf-openblas-60410-e8868b.out
### output ###
# [...]
# to the PKG_CONFIG_PATH environment variable
# No package 'openblas' found
# ++ pkg-config --libs openblas
# Package openblas was not found in the pkg-config search path.
# Perhaps you should add the directory containing `openblas.pc'
# to the PKG_CONFIG_PATH environment variable
# No package 'openblas' found
# + cc test.c
# test.c:2:10: fatal error: 'cblas.h' file not found
# #include <cblas.h>
#          ^~~~~~~~~
# 1 error generated.



<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><>  🐫 
┌─ The following actions failed
│ λ build conf-openblas 0.2.1
└─ 
```

Installing openblas via brew suggests setting `export PKG_CONFIG_PATH="/opt/homebrew/opt/openblas/lib/pkgconfig"` to get `pkg-config --cflags openblas` to work correctly. Running this manually works and correctly installs the package.

What is the recommended way to set `PKG_CONFIG_PATH` for this package?

PR #25076, opened 2024-01-15
Fix conf-openblas on macOS arm64 hardware

dra27:
Possible fix for #25075 - working correctly on my M2 and didn't break a run in a Debian 12 opam container.

`pkg-config` requires additional configuration after installing the openblas package - it's detailed after installing the package that `PKG_CONFIG_PATH` must be updated for this to work.

The first commit fixes the test command to do exactly that; the second commit also adds the amendment to the output of `opam env`. Note that opam ignores `FOO += ""` in a `setenv`. The second commit is optional, but having installed openblas, it seems logical for opam to ensure that `pkg-config` then actually works.

### DIFF
--- a/packages/conf-openblas/conf-openblas.0.2.1/opam
+++ b/packages/conf-openblas/conf-openblas.0.2.1/opam
@@ -10,7 +10,7 @@ build: [
   [
     "sh"
     "-exc"
-    "cc $CFLAGS $(pkg-config --cflags openblas) test.c $(pkg-config --libs openblas)"
+    "cc $CFLAGS $(PKG_CONFIG_PATH=\"$(brew --prefix openblas)/lib/pkgconfig:$PKG_CONFIG_PATH\" pkg-config --cflags openblas) test.c $(PKG_CONFIG_PATH=\"$(brew --prefix openblas)/lib/pkgconfig:$PKG_CONFIG_PATH\" pkg-config --libs openblas)"
   ] {os = "macos" & os-distribution = "homebrew"}
   ["sh" "-exc" "cc $CFLAGS test.c -lcblas"]
     {os-family = "arch"}
@@ -20,6 +20,9 @@ build: [
     {os = "win32" & os-distribution = "cygwinports"}
   ["sh" "-exc" "cc $CFLAGS test.c -lopenblas"]
     {os-distribution != "fedora" & os-distribution != "centos" & os-family != "suse" & os-family != "opensuse" & os != "macos" & os-family != "arch" & os != "freebsd" & os != "win32"}
+]
+depends: [
+  "conf-pkg-config" {os = "macos" & os-distribution = "homebrew"}
 ]
 depexts: [
   ["libc-dev" "openblas-dev" "lapack-dev"] {os-distribution = "alpine"}
--- a/packages/conf-openblas/conf-openblas.0.2.1/opam
+++ b/packages/conf-openblas/conf-openblas.0.2.1/opam
@@ -10,6 +10,16 @@ build: [
   [
     "sh"
     "-exc"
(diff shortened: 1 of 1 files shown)


packages/conf-openblas/conf-openblas.0.2.1/opam
```diff
@@ -35,6 +45,9 @@ x-ci-accept-failures: [
   "oraclelinux-7"
   "oraclelinux-8"
 ]
```
tmcgilchrist, line 47: Needs a dependency on `conf-pkg-config` somewhere here to work correctly.
```suggestion
]
depends: [
  "conf-pkg-config"
]
```

tmcgilchrist (commented):
Tested with `conf-pkg-config` fix locally on both x86 and Arm MacOS.

packages/conf-openblas/conf-openblas.0.2.1/opam
mseri:
```suggestion
]
depends: [
  "conf-pkg-config" {os = "macos" & os-distribution = "homebrew"}
]
```

mseri: I think I would restrict it to where it is really needed. With macports I never had any troubles

pklehre:
I suspect this commit could be causing problems on my Debian machine. See this issue: https://github.com/ocaml/opam-repository/issues/25194

mseri:
That is surprising though, I wonder why we don’t see issues on debian in the ci

pklehre:
Could it be that a critical environment variable is set to the empty string? The specific error message seems only to be triggered in line 216 of https://github.com/ocaml/opam/blob/master/src/state/opamEnv.ml, perhaps because the variable `var` is bound to an empty string?

hannesm:
I wonder why such a major semantic change (introducing some `setenv` business on all platforms) didn't impose a version bump of the `conf-` package? Is this in line with the best current practises of opam-repository (opam lock, append-only, ...)?

mseri:
You are right, we should have made it with a `-1` and never assume “it is safe”. If you can send a PR to revert this, I can make a new one where we require opam > 2.1.5 for this  ariant package

mseri:
I am not at my laptop until later in the afternoon, the best I can do is a fast merge and then a new PR later

hannesm:
@mseri revert at https://github.com/ocaml/opam-repository/pull/25195 -- thanks for investigations and quick responses (to both @mseri and @pklehre)

hannesm:
`-1` -- well, or since it is a conf-package, just bump it to 0.2.2 (or 0.3 or 1.0)? (read: I don't know anything about conf- package versioning)

mseri:
That is something that at some point we should figure out. So far is mostly just .x, adding 1 to the version for large changes and -1 for smaller, but different packages have different versioning for different (sometimes historical but other times not) reasons 😳

hannesm:
Fine with me. My intuition is: adding "os" "distributions" to an existing conf- package doesn't deserve a version bump, but everything else does.

Is it annoying (i.e. recompiles on every client) if a depexts changes? I do not know (hopefully someone else does).

I remember #10531 still being open, but eventually there should be some concrete policies being written down (plus CI-checked).

dra27:
Oh dear, apologies for this. 

> I wonder why such a major semantic change (introducing some `setenv` business on all platforms) didn't impose a version bump of the `conf-` package? Is this in line with the best current practises of opam-repository (opam lock, append-only, ...)?

In limited defence, the point here is that it was _not_ believed to be a semantic change affecting all packages. I had not registered that the second commit here relied on opam 2.0.11+ or 2.1.5+ to work correctly, as noted by my unfortunately incomplete comment "Note that opam ignores `FOO += ""` in a `setenv`".

I don't know the full history on conf version bumping either - in this PR, I incorrectly believed that the semantic change only fixed things. I'm not sure what the best thing here in terms of testing would be - if this had been put in a new version, we'd still have had all the breakage, it would have "just" been with a different version number.

I'll prepare a replacement PR in a sec.

PR #25077, opened 2024-01-15
Package archetype.1.5.1

guillaumeduhamel:
### `archetype.1.5.1`
Archetype language compiler
Archetype is a general purpose language to develop smart contracts
on the Tezos blockchain, with a specific focus on contract security

---
* Homepage: https://archetype-lang.org
* Source repo: git+https://github.com/completium/archetype-lang.git
* Bug tracker: https://github.com/completium/archetype-lang/issues

---


raphael-proust:
The CI failures are because ocaml5 is bytecode-only on some platforms. There's a way to deal with it but I can't remember it rn. I'll try to find it but someone might beat me to it. It's something like adding a constraint on `ocaml-option-bytecode-only` or something.

guillaumeduhamel:
Hello @raphael-proust, thank you for the explanations. From what I understand, on certain architectures, only bytecode is allowed. Upon reviewing the logs, I identified that the issue originates from the rule responsible for generating the `api` executable. Consequently, I've appended `byte` to the `api`'s `mode` rule to accommodate this constraint. I hope this fix will rectify the problem. However, if the issue persists, I'm considering the elimination of this executable's generation. It's important to note that the primary component of this project is the `compiler` binary, and the `api` serves mainly as an interface for the endpoints in the `js` distribution.

raphael-proust:
If `byte` is appropriate because it's for targetting `js` then your solution makes sense. Note that `dune` also supports `best` to produce a `.exe` when available and a `.bc` otherwise. I don't know that it's possible to then call the built artifact, but it's possible to produce it.

Otherwise, adding the chunk below to your opam file should make your package unavailable on the bytecode-only plateform.

```
conflicts: [
  "ocaml-option-bytecode-only"
]
```

PR #25078, opened 2024-01-15
Dune opam name checking

Leonidas-from-XIV:
The upcoming Dune 3.13 includes stricter checking for OPAM package names in the `depends` field of `dune-project` files. It will now reject packages with dots (only valid as findlib names, not opam names, thus these never pointed to valid dependencies) as well as OPAM style dependency filters (`{`, which dune has never supported using angle bracket syntax, only s-expression syntax). More details to be found in https://github.com/ocaml/dune/pull/9472.

It turns out this breaks a small amount packages (that for the most part already have been fixed or where fixes have already been submitted and/or merged upstream), so this PR adds version constraints to the affected older releases.

Each commit takes care of one particular project, the commits have links to the failing OPAM-repository CI builds of the Dune 3.13~alpha1 test run in #25067.

cc @ElectreAAS

### DIFF
--- a/packages/archetype/archetype.1.0.0/opam
+++ b/packages/archetype/archetype.1.0.0/opam
@@ -23,7 +23,7 @@ on the Tezos blockchain, with a specific focus on contract security
 """
 depends: [
   "ocaml" {>= "4.07.0"}
-  "dune" {>= "1.10.0"}
+  "dune" {>= "1.10.0" & < "3.13"}
   "menhir"
   "digestif" {>= "0.7.2" }
   "num"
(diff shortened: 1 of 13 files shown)


raphael-proust:
The changeset looks good to me. Waiting for the CI to complete.

Leonidas-from-XIV:
Happy to report that https://github.com/ip2location/ip2location-io-ocaml/pull/2 has been merged very quickly upstream so all the affected packages are already fixed upstream (and most have newer releases with fixes)

avsm:
Thanks!

PR #25084, opened 2024-01-16
Add libtorch 2.1.2+linux

mwlon:
This is the first Jane Street-managed release of the OPAM libtorch package. We are taking maintenance responsibility, and this is backed by [Laurent's comment here](https://github.com/ocaml/opam-repository/pull/23742#discussion_r1183116185) and [the original repository's README](https://github.com/LaurentMazare/ocaml-torch).

### DIFF
--- a/packages/libtorch/libtorch.2.1.2+linux-x86_64/opam
+++ b/packages/libtorch/libtorch.2.1.2+linux-x86_64/opam
@@ -1,0 +1,21 @@
+opam-version: "2.0"
+maintainer: "Jane Street developers"
+authors: ["Jane Street Group, LLC"]
+homepage: "https://github.com/janestreet/torch"
+bug-reports: "https://github.com/janestreet/torch/issues"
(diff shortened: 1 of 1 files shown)


raphael-proust:
Is there a public comment anywhere about the change of license? The previous maintainer's repository is under Apache whereas the new one is under MIT. There's a mention of sublicensing in the `LICENSE_NOTES` but not much.

Everything else seems a-ok :)

mwlon:
We have a private comment to that effect, but not a public one. @LaurentMazare would you be able to chime in?

LaurentMazare:
> We have a private comment to that effect, but not a public one. @LaurentMazare would you be able to chime in?

Sure, the change of license from Apache to MIT is good with me, and looking at the history of the project I wrote > 99% of the original code so I think it's all fine to relicense. Also happy with the change of maintainer, thanks @mwlon for taking ownership of this :)

mwlon:
@raphael-proust are we good to merge this?

PR #25100, opened 2024-01-19
fix broken tarball URLs from lri.fr

backtracking:
The websites at lri.fr no longer exist, or will be discontinued very soon.
This PR updates several URLs of tarballs that have been moved to github. In the process, it also updates contact emails and homepages of various packages.


joaosreis:
Hi @backtracking. Could I kindly ask you to include in this PR the URL update for older package versions also referring to lri.fr (e.g. ocamlgraph)? I work on a project that depends on an older version of ocamlgraph, and I imagine some other projects depend on it as well.

backtracking:
Thanks @mseri !

PR #25103, opened 2024-01-20
Why3 1.7.1

silene:


silene:
The error is the same for the three failing architectures (arm32-ocaml-5.1, ppc64-ocaml-5.1, x86_32-ocaml-5.1):

> ocamlfind: Not supported in your configuration: ocamlopt

Do these architectures not support the native compiler?

jmid:
> Do these architectures not support the native compiler?

As of OCaml 5, 32-bit architectures are bytecode only.

ppc64 is bytecode only in 5.0 and 5.1.
The 64-bit native code POWER backend will be restored with the forthcoming 5.2 release: https://github.com/ocaml/ocaml/blob/5.2/Changes#L8

packages/why3/why3.1.7.1/opam
```diff
@@ -0,0 +1,90 @@
...
+  "ppx_sexp_conv" {build}
+  "mlmpfr"
+]
+
+conflicts: [
+  "why3-base"
+  "ocamlgraph" {< "1.8.2"}
+  "mlmpfr" {< "4.0.0"}
+  "base-effects"
+]
```
mseri, line 78:
```suggestion
  "ocaml-option-bytecode-only"
]
```

PR #25123, opened 2024-01-26
lower ocaml bound on hc

zapashcanon:
Hi,

I have been asked if the bound could be lowered to 4.14. I bumped it to 5.0 because I had to change the code after the removal of some functions in the Ephemeron API. But actually it should still be compatible with 4.14.

Sorry for the two PRs...

Thanks!


mseri:
Built fine on 4.14

mseri:
I am merging since the CI is struggling

zapashcanon:
Thanks!

PR #25136, opened 2024-01-30
Enable test suite for Coq 8.19

SkySkimmer:
Let's see if we can resolve the CI failures.


SkySkimmer:
- vm segfault with ocaml 5.1 (test 11170)
- freebsd make not supported by our test suite makefile
- non-marshallable-state failing on arm32 (not sure why)
- tc_declaration_observer doesn't have the expected output on macos arm64 (not sure why, maybe buffered output got dropped??)
- timing test fails on all macos (not sure why, it works in our CI)

PR #25138, opened 2024-01-30
[new-release] melange (3.0.0)

anmonteiro:
Toolchain to produce JS from Reason/OCaml

Project page: https://github.com/melange-re/melange

This release includes a version for each of the following OCaml compiler versions: 4.14 and 5.1.

Starting with 3.0, we're experimenting with releasing a version of Melange that works across more than one OCaml version, following the same release naming as Merlin.


anmonteiro:
opened https://github.com/ocaml/opam-repository/pull/25141 to fix the revdeps issues.

packages/melange/melange.3.0.0-414/opam
```diff
@@ -0,0 +1,48 @@
+version: "3.0.0-414"
```
mseri, line 1:
```suggestion
```

packages/melange/melange.3.0.0-51/opam
```diff
@@ -0,0 +1,48 @@
+version: "3.0.0-51"
```
mseri, line 1:
```suggestion
```

mseri:
Thanks!

PR #25141, opened 2024-01-31
fix: Melange 3 revdeps

anmonteiro:


arbipher:
I guess you can modify (force-pushed) your previous pr #25138 instead of making a new one. The CI should relaunch.

mseri:
Thanks, I thin it is generally better to keep the revdeps separately to spare CI cycles

packages/tablecloth-melange/tablecloth-melange.0.0.9/opam
```diff
@@ -16,7 +16,7 @@ depends: [
   "ocaml" {>= "5.1"}
   "dune" {>= "3.8"}
   "reason"
-  "melange"
+  "melange" {>= "2.0.0" & < "3.0.0"}
```
mseri, line 19: I assume the lower bound is intentional and was incorrect here, but I prefer always to double check :)

mseri: @anmonteiro please let me know and then I'll merge

anmonteiro: oops, not intentional. if this was tested with melange 1.0 then we can revert.

PR #25143, opened 2024-01-31
Package lru_cache.0.4.0

zoggy:
### `lru_cache.0.4.0`
A simple implementation of a Least-Recently-Used cache

---
* Homepage: https://zoggy.frama.io/ocaml-lru-cache/
* Source repo: git+https://framagit.org/zoggy/ocaml-lru-cache.git
* Bug tracker: https://framagit.org/zoggy/ocaml-lru-cache/issues

---

### DIFF
--- a/packages/lru_cache/lru_cache.0.4.0/opam
+++ b/packages/lru_cache/lru_cache.0.4.0/opam
@@ -1,0 +1,39 @@
+opam-version: "2.0"
+synopsis: "A simple implementation of a Least-Recently-Used cache"
+maintainer: "zoggy@bat8.org"
+authors: "Zoggy <zoggy@bat8.org>"
+license: "GPL-3.0-only"
+homepage: "https://zoggy.frama.io/ocaml-lru-cache/"
+doc: "https://zoggy.frama.io/ocaml-lru-cache/refdoc/lru_cache/"
+bug-reports: "https://framagit.org/zoggy/ocaml-lru-cache/issues"
+depends: [
+  "dune" {>= "2.9"}
+  "ocaml" {>= "4.12.0"}
+  "lwt" {with-test}
+  "odoc" {with-doc}
+]
+build: [
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "--promote-install-files=false"
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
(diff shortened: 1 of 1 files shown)


zoggy:
Beware that the package is renamed `lru_cache` instead of `lru-cache`.

mseri:
The new name is conflicting with `lru_cache` from janestreet

zoggy:
> The new name is conflicting with `lru_cache` from janestreet

Jane Street who did not care creating `lru_cache when lru-cache already existed 😠 . I will rename it to `cache_lru`.

mseri:
Unfurtunately I did not notice the clash of names before, otherwise I would have notified it to them when they had the release.

If you decide to change name, I would suggest to add a `lru_cache.transition` that installs the new package or that informs users of the change.

raphael-proust:
I tested locally installing both `lru_cache` and `lru-cache` in opam, then depending on either in `dune`. It works in both cases (i.e., dune successfully builds using either dependencies).

`merlin` is able to jump to `lru_cache` sources but unable to jump to the sources of `lru-cache`. I think this is because the `mli` file is not installed (in the existing version).

Anyway, we'll try to get the CI to have one additional check to avoid these collisions in the future.

mseri:
@zoggy please wait before doing anything, we are going to ask JaneStreet if it is possible to rename their package since it is very recent and the clash of names can already be a problem with the various cache features under experimentation. Ping @dkalinichenko-js

mseri:
Oh that would be fantastic, thanks a lot. We are really sorry for not noticing it earlier, we will add a CI lint test to avoid this issue in the future

zoggy:
Thanks to all :-)

mseri:
Renaming of JS package: https://github.com/ocaml/opam-repository/pull/25155

@zoggy I suggest you add a `lru-cache.transition` installing `lru_cache {>= "0.4.0"}` as a dependency (you can add `conflicts: "lru_cache" {= "v0.16.0"}` to the opam file to ensure avoiding clashes there) with a `post-message` that the package name now will follow the library name. This way your users will be aware of the change.

zoggy:
Since only erssical (of which I am also the author) uses lru-cache, I think a .transition package may not be that useful :-)

mseri:
For now please add the conflict on your package so that it picks the correct library then. I would still add a message on the old packages to ensure that people pick the correct one

zoggy:
Oups, I destroyed my opam repo to start clean and make the changes you proposed. I will resubmit PR when it is merged.

PR #25144, opened 2024-01-31
afl-persistent 1.4: add patch to fix macos build

mseri:

PR #25150, opened 2024-01-31
afl-persistent: add md5 to patch

mseri:
Fixes https://github.com/ocaml/opam-repository/pull/25144#discussion_r1473275516

### DIFF
--- a/packages/afl-persistent/afl-persistent.1.4/opam
+++ b/packages/afl-persistent/afl-persistent.1.4/opam
@@ -10,7 +10,10 @@ build: [
   [ "./config.sh" ]
 ]
 patches: [
-  ["config.patch" "md5=ff93f594181d6508acdce88597e3c7b6"]
+  "config.patch"
+]  
+extra-files: [
+   ["config.patch" "md5=ff93f594181d6508acdce88597e3c7b6"]
 ]
 depends: [
   "ocaml" {>= "4.05"}
--- a/packages/afl-persistent/afl-persistent.1.4/opam
+++ b/packages/afl-persistent/afl-persistent.1.4/opam
@@ -10,7 +10,7 @@ build: [
   [ "./config.sh" ]
 ]
 patches: [
-  "config.patch"
+  ["config.patch" "md5=ff93f594181d6508acdce88597e3c7b6"]
 ]
 depends: [
   "ocaml" {>= "4.05"}


packages/afl-persistent/afl-persistent.1.4/opam
```diff
@@ -10,7 +10,7 @@ build: [
   [ "./config.sh" ]
 ]
 patches: [
-  "config.patch"
+  ["config.patch" "md5=ff93f594181d6508acdce88597e3c7b6"]
```
hannesm, line 13:
```suggestion
  "config.patch"
]  
extra-files: [
   ["config.patch" "md5=ff93f594181d6508acdce88597e3c7b6"]
```

hannesm: including the checksum directly in the `patches` section is not valid opam syntax

mseri: Aaah, thanks, that is why I had not found it in the docs when I made the PR. So we will need to add all patches both as patches and as extra files? Wouldn't make sense to add the hash also for the patches?

hannesm:
ref https://opam.ocaml.org/doc/Manual.html#opamfield-patches

mseri:
Yes, I had read that when I made my original PR and I thought we could not add hashes for pathches. When I read your comment in the other PR, I had misunderstood that we can using the syntax from the comment.

packages/afl-persistent/afl-persistent.1.4/opam
hannesm: yes, well -- there are various paths forward:
- the whole extra-files stuff is quite cumbersome in various areas of opam, maybe it can be sunset (while retaining the option to add patches as extra-sources)?
- another option is to include these patch files directly in the opam file (they'd grow in size)
- in early 2022, Reynir and myself spent a day to figure out what the semantics / intentions of opam and checksums (and extra-files) is, and we discovered some inconsistent behaviour - it may be worth to put some re-thinking on a future roadmap (2.3? 3.0?)
- the checksums were added for repository signing, but as pointed in the refered PR, the data in `files` is copied independent of whether it is listed in the opam file --> this means for repository signing we can't rely on that fact (or need to wait until the PR mentioned is merged, and opam will be more strict).

but, with the current state of affairs, it's best to put the patches and extra-files in the opam file :)

mseri: Makes sense, thanks for the context :)

PR #25159, opened 2024-02-02
Before renaming lru-cache to lru_cache

zoggy:
Add warning in lru-cache that it is renamed to lru_cache since version 0.4.0.
Add conflits in erssical with lru_cache v0.16.0.


mseri:
Thanks

PR #25174, opened 2024-02-04
[new release] Docfd 2.1.0

darrenldl:

Homepage: https://github.com/darrenldl/docfd

Think interactive grep for both text files and PDFs,

and PDF viewers,

Features:


mseri:
Thanks

PR #24187, opened 2023-07-31
Add support for enabling OCaml's thread sanitizer mode

dra27:
Now that https://github.com/ocaml/ocaml/pull/12114, this adds the glue necessary for turning on tsan support in trunk, cf. the long-existing `ocaml-option-afl` for turning on AFL fuzz in a switch, etc.

Various shed colours to choose - precise name of `ocaml-option-tsan` (we have `ocaml-option-address-sanitizer`, etc. but these are for the C compiler - it may be want `ocaml-option-thread-sanitizer` even though it's a bit of a mouthful) and similarly whether it's `conf-unwind` or `conf-libunwind`).

NB Updating `ocaml-options-vanilla` _doesn't_ `ocaml-base-compiler` recompile switches (they do "recompile" ocaml-options-vanilla).

PR #25180, opened 2024-02-05
packages for OCaml 5.2.0~alpha1

Octachron:
This PR adds the two opam files for the 5.2.0~alpha1 version of the compiler with a small update for the `ocaml.5.2.0` metapackage to allow the installation of `5.2.0~*` packages.

The opam files has also been updated to add support for the native power64 port in 5.2.0 .

### DIFF
--- a/packages/ocaml-variants/ocaml-variants.5.2.0~alpha1+options/opam
+++ b/packages/ocaml-variants/ocaml-variants.5.2.0~alpha1+options/opam
@@ -41,6 +41,7 @@ build: [
     "--disable-flat-float-array" {ocaml-option-no-flat-float-array:installed}
     "--enable-flambda" {ocaml-option-flambda:installed}
     "--enable-frame-pointers" {ocaml-option-fp:installed}
+    "--enable-tsan" {ocaml-option-tsan:installed}
     "CC=cc" {!ocaml-option-32bit:installed & !ocaml-option-musl:installed & (os="openbsd"|os="macos")}
     "CC=musl-gcc" {ocaml-option-musl:installed & os-distribution!="alpine"}
     "CFLAGS=-Os" {ocaml-option-musl:installed}
@@ -75,4 +76,5 @@ depopts: [
   "ocaml-option-leak-sanitizer"
   "ocaml-option-address-sanitizer"
   "ocaml-option-static"
+  "ocaml-option-tsan"
 ]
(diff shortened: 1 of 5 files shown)


packages/ocaml-variants/ocaml-variants.5.2.0~alpha1+options/opam
```diff
@@ -0,0 +1,78 @@
...
+  [
+    "./configure"
+    "--prefix=%{prefix}%"
+    "--docdir=%{doc}%/ocaml"
+    "-C"
+    "--with-afl" {ocaml-option-afl:installed}
+    "--disable-native-compiler" {ocaml-option-bytecode-only:installed}
+    "--disable-flat-float-array" {ocaml-option-no-flat-float-array:installed}
+    "--enable-flambda" {ocaml-option-flambda:installed}
+    "--enable-frame-pointers" {ocaml-option-fp:installed}
```
OlivierNicole, line 43:
```suggestion
    "--enable-frame-pointers" {ocaml-option-fp:installed}
    "--enable-tsan" {ocaml-option-tsan:installed}
```
As this is the first release comprising TSan support, should we consider supporting the already existing option (introduced by @dra27 in #24187) `ocaml-option-tsan`?

packages/ocaml-variants/ocaml-variants.5.2.0~alpha1+options/opam
```diff
@@ -0,0 +1,78 @@
...
+  "ocaml-option-afl"
+  "ocaml-option-bytecode-only"
+  "ocaml-option-no-flat-float-array"
+  "ocaml-option-flambda"
+  "ocaml-option-fp"
+  "ocaml-option-musl"
+  "ocaml-option-leak-sanitizer"
+  "ocaml-option-address-sanitizer"
+  "ocaml-option-static"
+]
```
OlivierNicole, line 78:
```suggestion
  "ocaml-option-tsan"
]
```

packages/ocaml-variants/ocaml-variants.5.2.0~alpha1+options/opam
Octachron: You are right, thanks for spotting my mistake !

kit-ty-kate:
There are more tsan related different between this package and `5.2.0+trunk` (the one on macOS), is that expected? I thought one step of the release process was to make sure there was no difference with the similar packages already on opam-repository. If it's not already a step i think it should be added to avoid such issues in the future.

Octachron:
That's already one step, but I forgot to check the tsan configuration on 5.2.0 trunk.

My solution at this point would be to have template for those opam files, otherwise I end up with multiple frayed sources of truth.

kit-ty-kate (approved):

Octachron:
Is there some issue with the packages? What is the reason behind the merging delay?

mseri:
Time :)

Octachron:
Thanks !

PR #25112, opened 2024-01-23
Octez 19.0 packages

killian-delarue:
Packages for Octez v19.0

PR #25183, opened 2024-02-05
Mark old tezos-sapling packages as unavailable

raphael-proust:
- They are broken
- They keep tripping the CI in revdeps builds
- There is no will from the maintainer's employer to fix them


raphael-proust:
Build errors triggered in https://github.com/ocaml/opam-repository/pull/25179 and https://github.com/ocaml/opam-repository/pull/25112 and others

PR #25193, opened 2024-02-06
Mark broken `ogg` packages as unavailable

raphael-proust:
See https://github.com/ocaml/opam-repository/pull/25179#issuecomment-1927442067

### DIFF
--- a/packages/ogg/ogg.0.6.0/opam
+++ b/packages/ogg/ogg.0.6.0/opam
@@ -23,6 +23,9 @@ build: [
     "@doc" {with-doc}
   ]
 ]
+conflicts: [
+  "ocaml" { >= "4.14" }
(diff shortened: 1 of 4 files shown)


mseri:
Is it really `available: false`? I think we only need an upper bound on the ocaml version

raphael-proust:
True…

I've changed for version constraints. However, some of the packages didn't have a dependency to ocaml so I've made it a conflict. Not sure if I should just add the dependency instead.

mseri:
They build with dune, I would say that they should depend on ocaml

mseri:
In any case, this is a strict improvement and a conflict does the job, I don't think it is worth changing it again

mseri:
Thanks

PR #25195, opened 2024-02-07
Revert "Fix conf-openblas on macOS arm64 hardware"

hannesm:
Reverts ocaml/opam-repository#25076

it breaks with some opam versions, maybe < 2.1.5 (as spotted by @pklehre in https://github.com/ocaml/opam-repository/pull/25076#issuecomment-1931923608 https://github.com/ocaml/opam-repository/issues/25194 https://discuss.ocaml.org/t/opam-error-invalid-argument-opamenv-unzip-to/14012)

Issue #25194, opened 2024-02-07
Invalid_argument("OpamEnv.unzip_to") opam error after installing conf-openblas


pklehre:
The following sequence of commands on a fresh opam init (opam v2.1.3) leads to an error that renders the opam switch in an unusable state (e.g., impossible to install new packages)

```
opam switch create 5.1.1+flambda ocaml-variants.5.1.1+options ocaml-option-flambda
opam install conf-openblas
```

Error:

```
[pkl@zeta ~]$ opam env
Fatal error:
Invalid_argument("OpamEnv.unzip_to")
```

mseri:
I see, I think it is due to a bug in opam < 2.1.5:
```
2.1.5:
* [BUG] Variables are now expanded in build-env (as for setenv) [#5352 @dra27]
* Correctly handle empty environment variable additions [#5350 @dra27]
* Skip empty environment variable additions [#5350 @dra27]
[…]
```

mseri:
Ping @dra27, what do you suggest to keep openblas installable? I think we should probably revert your commit and simply ask people to configure PKG_CONFIG_PATH appropriately in a post message

PR #25196, opened 2024-02-07
Fix conf-openblas on macOS arm64 hardware, hopefully without breakages

dra27:
Re-attempt at #25076, following the needed revert of the second part in #25195.

First commit is the same as #25076 - it changes macOS only, introducing a missing dependency (the package already used pkg-config without depending on its conf- package) and amending the call to `pkg-config` for Homebrew only to match Homebrew's recommendation.

The second commit addresses the unintended breakage in #25194 (for which apologies again!). There are two issues at play: `setenv` doesn't support filters (ocaml/opam#5787) and empty updates to environment variables crash opam, which is fixed in 2.1.5, but not in earlier releases (ocaml/opam#5350).

The fix I propose is to put the `setenv` field in a package of its own, and have `conf-openblas` depend on that package only for macOS with Homebrew. Note that `conf-openblas-macOS-env` will only install if `conf-openblas` installed, which means that we know that the environment variable update will not be empty, hence there is no need to check for the opam-version itself.

### DIFF
--- a/packages/conf-openblas-macOS-env/conf-openblas-macOS-env.1/opam
+++ b/packages/conf-openblas-macOS-env/conf-openblas-macOS-env.1/opam
@@ -1,0 +1,25 @@
+opam-version: "2.0"
+maintainer: "David Allsopp"
+authors: [ "David Allsopp" ]
+homepage: "https://opam.ocaml.org"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: "ISC"
+build: [
+  "sh"
+  "-exc"
+  "printf 'opam-version: \"2.0\"\\nvariables {\\n  pkg-config-homebrew: \"%s/lib/pkgconfig\"\\n}' \"$(brew --prefix openblas)\" > %{_:name}%.config"
+]
+depends: [
+  "conf-openblas" {>= "0.2.2"}
+]
+setenv: [
+  PKG_CONFIG_PATH += "%{_:pkg-config-homebrew}%"
+]
+synopsis: "Virtual package to set the environment correctly for OpenBLAS in Homebrew"
+description: """
+Homebrew requires an update to PKG_CONFIG_PATH in order to detect the openblas
+package. opam 2.0.10 and earlier and 2.1.4 and earlier cannot handle a setenv
+update of "", and setenv updates cannot be filtered. This package exists to
+workaround these two limitations."""
+flags: conf
+available: os = "macos" & os-distribution = "homebrew"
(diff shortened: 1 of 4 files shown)


dra27:
(draft because I don't at present have access to my Mac to test it)

mseri:
Oh this lint was not there the last time:
```
2024-02-07 14:39.15: Error in conf-openblas-macOS-env.1:            warning 56: It is discouraged for non-compiler packages to use 'setenv:'
```

hannesm:
would you mind to increase the version of "conf-openblas" to 0.2.2?

dra27:
The lint issue was there last time - although it's at present removed in opam 2.2, because the issue which caused it be a warning has itself disappeared (cf. https://github.com/ocaml/opam/blob/5fa0ac762f4c8e6e2a45272adc6ef1fa4c6dcf28/src/state/opamFileTools.ml#L655-L661)

dra27:
For both changes, @hannesm, or just for the second env-related change?

If for both, presumably the 0.2.1 should be marked as unavailable on macOS arm64 with Homebrew?

mseri:
I would just move `packages/conf-openblas/conf-openblas.0.2.1/opam -> packages/conf-openblas/conf-openblas.0.2.2/opam` and not touch the other. When you run `opam upgrade` you'll get the latest one anyway

dra27:
OK, I don't mind doing it (and will push in a sec!), but it's that bit which feels strange for using a version number bump. It's a _packaging_ error (we're not trying to install an older version of the thing, we're preserving older incorrect packaging instructions for the same thing), but we end up leaving opam-repository where if for any reason a solver on macOS happens to try 0.2.1 (lowerbounds, etc.), then it'll fail.

mseri:
You can make it unavailable for macos on arm64 with homebrew if that is completely uninstallable, I thought it was working but required manually setting the correct PKG_CONFIG_PATH all the time you wanted to install something

dra27:
With `conf-openblas.0.2.1` with macOS Homebrew, `PKG_CONFIG_PATH` needs to be set _both_ when running `opam install` and also when installing/building/running anything which uses the openblas package. The first part - having to set `PKG_CONFIG_PATH` in order to _install_ `conf-openblas` - I think is a bug in the conf-openblas package (so I just pushed a disabling of conf-openblas.0.2.1 for macOS+Homebrew). The second part is more of a "smell" - the `conf-openblas` package is meant to configure the openblas package to be ready for use, but it hadn't (because the `PKG_CONFIG_PATH` is left unconfigured) forcing _dependent_ packages to need to a special case for Homebrew. The setenv part is then what deals with that.

dra27:
I've now tested and verified this on both Intel and ARM Mac hardware

mseri:
Thanks

PR #25217, opened 2024-02-09
Point broken upstreams to opam-source-archives

mseri:
From https://github.com/ocaml/opam-repository/issues/15298#issuecomment-1931475275 and https://github.com/ocaml/opam-repository/issues/12421#issuecomment-1931480130

I did omit the packages that have `available:false`


mseri:
The new archives are correctly retreived, the other lint failures for those packages have always been there

PR #25225, opened 2024-02-10
[new release] html_of_jsx (0.0.1)

davesnx:
Render HTML writing JSX

- Project page: <a href="https://github.com/davesnx/html_of_jsx">https://github.com/davesnx/html_of_jsx</a>

## 0.0.1

- First working version of the ppx and library
- Supports most of features from [JSX](https://reasonml.github.io/docs/en/jsx) (uppercase components, fragments, optional attributes, punning)
- but with a few improvements (lowercase components, no need to add annotations)
- No React idioms (no `className`, no `htmlFor`, no `onChange`, etc...)
- Type-safe, validates attributes and their types ([it can be better thought](https://github.com/davesnx/html_of_jsx/issues/2))
- Minimal
  - `Html_of_jsx.render` to render an element to HTML
  - `Jsx.*` to construct DOM Elements and DOM nodes (`Jsx.text`, `Jsx.int`, `Jsx.null`, `Jsx.list`)
- Works with [Reason](https://reasonml.github.io) and [mlx](https://github.com/andreypopp/mlx)
- Supports some htmx under the ppx (`html_of_jsx.ppx -htmx`)

### DIFF
--- a/packages/html_of_jsx/html_of_jsx.0.0.1/opam
+++ b/packages/html_of_jsx/html_of_jsx.0.0.1/opam
@@ -1,0 +1,42 @@
+opam-version: "2.0"
+synopsis: "Render HTML writing JSX"
+maintainer: ["David Sancho <dsnxmoreno@gmail.com>"]
+authors: ["David Sancho <dsnxmoreno@gmail.com>"]
+license: "MIT"
+homepage: "https://github.com/davesnx/html_of_jsx"
+bug-reports: "https://github.com/davesnx/html_of_jsx/issues"
+depends: [
+  "dune" {>= "3.8"}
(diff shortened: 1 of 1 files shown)


mseri:
On freebsd it fails with
```
# +  /tmp/dune_cram_62cc2e_.component.t/1.sh: ../ppx.sh: not found
```
I don't know if the problem is with this package or dune. It does not block the merge since freebsd in the CI is still experimental.

mseri:
Thanks

davesnx:
Thanks to you @mseri

davesnx:
I can't tell what's the error on freebsd, but on the cram tests we use a ppx.sh script that is used as dep. Probably freebsd fails at symlink it?

PR #25234, opened 2024-02-11
Disable all old alpha/beta/rc releases on Windows

dra27:
Shedding weight from the Windows branch. It is just possible that when I finally switch the compiler packages over to being generated that it will be trivial to re-enable these, but even I don't really see the value in upgrading the alpha, beta & rc packages from old releases to work on native Windows.

It is, however, problematic if they get accidentally selected (e.g. from a solver which ignores avoid-version, etc.) so I propose marking them as unavailable on Windows.


mseri:
Thanks

PR #25235, opened 2024-02-11
Patch conf-pkg-config.3 for Windows

dra27:
Adapts conf-pkg-config.3 for opam 2.2's Windows depext support. The old versions remain available using "OCaml for Windows"'s depext-cygwinports plugin, but are otherwise unavailable.

depext-cygwinports installs a shim executable for pkg-config, which is why it didn't the extra build command, but it's fine to detect via pkgconf as well (i.e. it's OK that the filter on the command is `os = "win32"` not `os = "win32" & os-distribution != "cygwinports"`).


mseri:
Thanks

PR #25238, opened 2024-02-12
dune-release.2.0.0 requires dune < 3.14 for tests

emillon:
ocaml/dune#9851 adds a loc which is visible in the test suite.


mseri:
Thanks

PR #25239, opened 2024-02-12
Minor corrections to compiler dev-repos

dra27:
Spotted in passing while overhauling these (these things are basically typos) - given that they affect only 4.13's options packages (so an older release) and +trunk packages, I figure we can risk the switch churn of updating those separately?


mseri:
Thanks

PR #25244, opened 2024-02-12
archimedes 0.4.19: relax upper bound on oasis

pqwy:
archimedes.0.4.19 builds with any oasis > 0.4.7 currently in the repo.

Hard upper bound makes archimedes.0.4.19 non-installable on a recent compiler (oasis.0.4.6 needs ocaml < 4.06.0), and jupyter-archimedes non-installable totally (jupyter in addition needs ocaml >= 4.10.0).


mseri:
Thanks!

PR #25246, opened 2024-02-12
[new release] gmp (6.2.1-5)

dinosaure:
The GNU Multiple Precision Arithmetic Library

- Project page: <a href="https://github.com/mirage/ocaml-gmp">https://github.com/mirage/ocaml-gmp</a>

##### CHANGES:

- Fix the cross-compilation for `aarch64` (@dinosaure, @shym, mirage/ocaml-gmp#22)
- Fix the compilation on OpenBSD (@kit-ty-kate, mirage/ocaml-gmp#22)

### DIFF
--- a/packages/gmp/gmp.6.2.1-5/opam
+++ b/packages/gmp/gmp.6.2.1-5/opam
@@ -1,0 +1,29 @@
+opam-version: "2.0"
+maintainer: "Lucas Pluvinage <lucas@tarides.com>"
+license: ["LGPL-3.0-only" "LGPL-2.0-only"]
+authors: "Torbjörn Granlund and contributors"
+homepage: "https://github.com/mirage/ocaml-gmp"
+bug-reports: "https://github.com/mirage/ocaml-gmp/issues"
+dev-repo: "git+https://github.com/mirage/ocaml-gmp.git"
+substs: [ "src/build.sh" ]
+build: [
+ [ "dune" "build" "-p" name "-j" jobs ]
+ [ "dune" "runtest" "-p" name "-j" jobs ] {with-test}
+]
+depends: [
+  "ocaml" {>= "4.02.0"}
+  "dune" {>= "2.6"}
+  "conf-m4"
+]
+synopsis: "The GNU Multiple Precision Arithmetic Library"
+description: """Dune packaging of the GMP library, suitable for 
+cross-compilation."""
+url {
+  src:
+    "https://github.com/mirage/ocaml-gmp/releases/download/6.2.1-5/gmp-6.2.1-5.tbz"
+  checksum: [
+    "sha256=26b24412e18b511f1b8af6f9bf891452e214e32773fdf2f68004310dc9f016f5"
+    "sha512=d3f7cb9043bb387c95d2f95fa82d832dc00287a8ba039ac82b573c849943c181045d61af32b3f47ae48d43c85ab52ab23570a8c4095a7d84c854f26aa38a0844"
+  ]
+}
+x-commit-hash: "e15e2024d31973f39d0552b1b306db682e3ddaab"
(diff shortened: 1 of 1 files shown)


mseri:
Are the failures on macos expected?

```
#=== ERROR while compiling gmp.6.2.1-5 ========================================#
# context              2.2.0~alpha3 | macos/arm64 | ocaml-base-compiler.5.1.1 | pinned(https://github.com/mirage/ocaml-gmp/releases/download/6.2.1-5/gmp-6.2.1-5.tbz)
# path                 ~/.opam/5.1.1/.opam-switch/build/gmp.6.2.1-5
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p gmp -j 7
# exit-code            1
# env-file             ~/.opam/log/gmp-84989-3850ea.env
# output-file          ~/.opam/log/gmp-84989-3850ea.out
### output ###
# File "src/dune", line 14, characters 0-344:
# 14 | (rule
# 15 |  (targets gmp.h libgmp.a dllgmp.so)
# 16 |  (deps gmp-6.2.1.tar.xz build.sh)
# 17 |  (action
# 18 |   (with-stdout-to
# 19 |    build.log
# 20 |    (run sh ./build.sh "%{ocaml-config:c_compiler}" "%{ocaml-config:host}"
# 21 |      "%{ocaml-config:target}" "%{ocaml-config:ocamlc_cflags}"
# 22 |      %{ocaml-config:supports_shared_libraries}
# 23 |      "%{ocaml-config:native_c_libraries}"))))
# (cd _build/default/src && /bin/sh ./build.sh cc aarch64-apple-darwin23.1.0 aarch64-apple-darwin23.1.0 '-O2 -fno-strict-aliasing -fwrapv -pthread' true -lpthread) > _build/default/src/build.log
# ./build.sh: line 5: xzcat: command not found
# ./build.sh: line 6: cd: gmp-6.2.1: No such file or directory
```

hannesm:
Looks like newer macOS distributions do not include xz anymore by default... What likely needs to be done is adding a dependency on `xz` (and so far we don't have a conf-xz package in opam).

An alternative could be to integrate gmp as extra-sources in the opam file (instead of hosting it on github). Maybe even the .tar.gz (to avoid the xz dependency)? I suspect apart from the ocaml-CI in the mirage/ocaml-gmp repository, everything should be smooth. WDYT?

packages/gmp/gmp.6.2.1-5/opam
```diff
@@ -0,0 +1,29 @@
...
+build: [
+ [ "dune" "build" "-p" name "-j" jobs ]
+ [ "dune" "runtest" "-p" name "-j" jobs ] {with-test}
+]
+depends: [
+  "ocaml" {>= "4.02.0"}
+  "dune" {>= "2.6"}
+  "conf-m4"
+]
+synopsis: "The GNU Multiple Precision Arithmetic Library"
```
mseri, line 18:
```suggestion
depexts: [
  "xz" {os = "macos" & os-distribution = "homebrew"}
]
synopsis: "The GNU Multiple Precision Arithmetic Library"
```

mseri: Since this is only a problem on macos with homebrew (macports installs xz by default), I this does not warrant a conf package. We can either use a tar.gz or simply add the depext for maxos with homebrew for this release

dinosaure: I'm ok with this solution :+1:.

packages/gmp/gmp.6.2.1-5/opam
```diff
@@ -0,0 +1,32 @@
...
+ [ "dune" "build" "-p" name "-j" jobs ]
+ [ "dune" "runtest" "-p" name "-j" jobs ] {with-test}
+]
+depends: [
+  "ocaml" {>= "4.02.0"}
+  "dune" {>= "2.6"}
+  "conf-m4"
+]
+depexts: [
+  "xz" {os = "macos" & os-distribution = "homebrew"}
```
mseri, line 19:
```suggestion
  [ "xz" ] {os = "macos" & os-distribution = "homebrew"}
```

PR #25249, opened 2024-02-13
Octez 19.1

killian-delarue:
Packages for Octez v19.1


raphael-proust:
The CI failures are all on experimental or known-to-be-broken platforms. Seems ready for merging to me.

mseri:
There are some that did not look expected, like the initial ones on 4.14 but they all seem to fail due to busted ci (no space left on device)

PR #25252, opened 2024-02-13
[new release] ocaml-protoc (4 packages) (3.0.2)

c-cube:
Pure OCaml compiler for .proto files

- Project page: <a href="https://github.com/mransan/ocaml-protoc">https://github.com/mransan/ocaml-protoc</a>

##### CHANGES:

- fix segfault in pbrt in bytecode

### DIFF
--- a/packages/pbrt_yojson/pbrt_yojson.3.0.2/opam
+++ b/packages/pbrt_yojson/pbrt_yojson.3.0.2/opam
@@ -15,7 +15,7 @@ depends: [
   "base64" {>= "3.0"}
 ]
 build: [
-  ["dune" "subst"] {pinned}
+  ["dune" "subst"] {dev}
   [
     "dune"
     "build"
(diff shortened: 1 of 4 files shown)


arbipher:
> 'dune subst' should always only be called with {dev} (i.e. ["dune" "subst"] {dev}) If your opam file has been autogenerated by dune, you need to upgrade your dune-project to at least (lang dune 2.7).

This is reported by the CI. `["dune" "subst"] {dev}` is always suggested than `["dune" "subst"] {pinned}`.

c-cube:
I know, and I'm still annoyed I have to change this. Can we just move forward and ignore the linter?

packages/pbrt_yojson/pbrt_yojson.3.0.2/opam
```diff
@@ -0,0 +1,40 @@
...
+bug-reports: "https://github.com/mransan/ocaml-protoc/issues"
+depends: [
+  "dune" {>= "2.0"}
+  "ocaml" {>= "4.08"}
+  "odoc" {with-doc}
+  "yojson" {>= "1.6"}
+  "base64" {>= "3.0"}
+]
+build: [
+  ["dune" "subst"] {pinned}
```
mseri, line 18:
```suggestion
  ["dune" "subst"] {dev}
```

mseri:
@arbipher Indeed this happens for all projects using dune < 2.7. There is not much we can do about it before submission, but adding suggestions and committing them solves the problem in a few seconds so we tend to just do that :)

c-cube:
Thank y'all!

PR #25269, opened 2024-02-16
Add a preview release for merlin on OCaml 5.2

kit-ty-kate:
This temporary preview package brings merlin for OCaml 5.2 while it is in its pre-release phase.


raphael-proust:
LGTM: diff with the 4.13.1 version is just ocaml version support, src values, and opam metadata related to preview status.

CI failure (everything skipped) is to be expected considering 5.2 is not tested for yet.

PR #25278, opened 2024-02-17
[new release] random (0.0.1)

leostera:
Easy to use, cryptographically safe random data.

- Project page: <a href="https://github.com/leostera/random">https://github.com/leostera/random</a>

## 0.0.1

Initial release of the library with combinators for:

* cstruct
* int8
* int16
* int32
* int64
* char
* int
* float
* bytes
* bigarray
* string
* alphanum
* seq
* list

### DIFF
--- a/packages/random/random.0.0.1/opam
+++ b/packages/random/random.0.0.1/opam
@@ -1,0 +1,40 @@
+opam-version: "2.0"
+synopsis: "Easy to use, cryptographically safe random data"
+maintainer: ["Leandro Ostera <leandro@abstractmachines.dev>"]
+authors: ["Leandro Ostera <leandro@abstractmachines.dev>"]
+license: "MIT"
+tags: ["random" "random data"]
+homepage: "https://github.com/leostera/random"
+bug-reports: "https://github.com/leostera/random/issues"
+depends: [
+  "mirage-crypto" {>= "0.11.2"}
+  "mirage-crypto-rng" {>= "0.11.2"}
+  "ocaml" {>= "4.14.0"}
+  "randomconv" {>= "0.1.3"}
+  "dune" {>= "3.14"}
+  "odoc" {with-doc}
+]
+build: [
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
+dev-repo: "git+https://github.com/leostera/random.git"
+url {
+  src:
+    "https://github.com/leostera/random/releases/download/0.0.1/random-0.0.1.tbz"
+  checksum: [
+    "sha256=c8373956a728671d22ab4b7674574d6eb6cc67e2f8843a519d65b6fa37dcbdbb"
+    "sha512=1a584e7bee8ebe95960804378acb5487e4596bcf895acd080192168172be0aeca5ddc9120bbb44f8e853917b9a42c202e881d90e3d5b09389cd618fadbd2d178"
+  ]
+}
+x-commit-hash: "e89623218ee95d6b4ad39b62063bafbecd3383ee"


zapashcanon:
Are we really going to start publishing packages made of 15 lines of code that does nothing but combining two functions from an already existing package ?

sabine:
I think not exposing `Cstruct.t` is actually a net gain here. Okay, the package isn't very heavy at this point. :shrug:

A chatter on my stream suggests that, maybe the convenience functions should be provided by `mirage-crypto-rng` instead.

zapashcanon:
> A chatter on my stream suggests that, maybe the convenience functions should be provided by mirage-crypto-rng instead.

That would be much better indeed..

dinosaure:
> A chatter on my stream suggests that, maybe the convenience functions should be provided by mirage-crypto-rng instead.

What is the status of this PR then? Do you want a release of this package or re-integrate what you have done into `mirage-crypto`?

hannesm:
Hmm, so I have some issues with the `random` package proposed here:
- why not submit a pull request to the randomconv package that this package relies on heavily?
- the RNG initialization is done inside of this `random` package, which is not a good idea (RNG are best fed some further entropy while the application is running --> this is why there's mirage-crypto-rng-{eio,async,lwt,mirage} around
- `randomconv` itself has been crafted without dependencies on the underlying rng (at the downside that a `(int -> Cstruct.t)` function is needed at each call site

To me, this looks very similar to `castore` - which is very similar to `ca-certs-nss` / `ca-certs`.

Of course, all contributions as opam packages are welcome. But wouldn't it be more fruitful to enhance existing packages than to develop (and hopefully maintain) your own package with a handful lines of code? In the end I don't mind - go ahead with convenience package as you like.

leostera:
Hi folks 👋🏼 I'm noticing a lot of pushback lately and I'd love to understand where it's coming from.

I get this "old-guard gatekeeping" feeling when I break one of those implicit rules that get y'all commenting on what could otherwise be a very clean merge and publish (and in an ideal world would be completely automated), and to be frank it is frustrating and discouraging 😞

My guiding principle when publishing things is that **iteration is king**. The faster I can put code in front of devs, the faster I can find and fix bugs, and the faster the library evolves in the right direction.

Random may not be up to those implicit standards but it is:

* **easy to find**, it has a clean name that won't make you ask any questions about what it does (like "what is mirage" or "why do i need crypto"), and its positioned to help you get what you need directly.

* **easy to learn**, it has a clean, predictable API that mostly follows the standard Random module, and it asks very little of you. In Elixir you get `Enum.random`, in Erlang you have the `:rand` module, in Rust there's a `rand` package, Go has a `math.Rand` package. All of these have slightly different capabilities but they all give you the most idiomatic possible way of getting random data.

* **easy to extend**, since it is small, it can be grown in other directions as we see need such as combinators for choosing amongst many values, or random complex values, a-la qcheck.

* **hard to misuse**, there's a clear pit-of-success with it where the path of least resistance takes you to the result you want.

Some of you are definitely domain experts and can probably conjure up these definitions on the spot whenever you need them, so you may not see the value in this approach and these priorities. You can still do your thing.

For everyone else who just needs some random data, Random is likely good enough.

Now that I've made my position clear, can you help me understand what parts of it you disagree with? The best outcome here would be that we found we are in agreement over these priorities/values and can narrow down on what we disagree on and deal with that directly.

hannesm:
I don't think your "random" is "hard to misuse". Please read further about PRNG, seeding, entropy feeding if in doubt. Sorry, I've not more energy nor time to discuss your elaborate comment.

avsm:
Just to be clear, I don't view any of the objections to this package are blockers to it being merged.  This package introduces a valid OCaml library, with OCamldoc, and valid dependencies and license. Whether or not using this package is a good idea is not a judgement that the opam-repository has historically made.

mseri:
> I get this "old-guard gatekeeping" feeling when I break one of those implicit rules that get y'all commenting on what could otherwise be a very clean merge and publish (and in an ideal world would be completely automated), and to be frank it is frustrating and discouraging 😞

I can get the frustration but I disagree with dismissing the comments as "old-guard gatekeeping".

I think trying to contribute to the other libraries to improve them in general is a good idea, especially for thin shims where a library could benefit from having a `basic` or `simple` package or module exposing a simpler, easier to use, api. This improves the ecosystem and reduces fragmentation, even if that costs a slower turnaround, especially since dune vendoring makes it trivial to ship your modified version in your sources until the changes are all upstream.

Similarly for comments that could improve the library or reduce risks, like the one about intialization of RNG for example.

As per the discussion on naming convention which is I think the other thing you are referring to. I think it was unlucky timing, we have been revising/writing the policies of the repo, so the PR happened in a moment in which we are actively discussing these subjects and at the same time I was too busy at work to dedicate time to the opam-repo. Apologies if it had slowed down the merge but I think it was useful to start the discussion, and also to take the opportunity to share some ideas and concerns. 

This happens on PRs every now and then, definitely not just yours. I surely did not mean to attack you or personally derange your contributions, apologies if that seemed the case.

PR #25285, opened 2024-02-19
mbr-format.2.0.0 doesn't depend on lwt

reynir:
After making this change I discovered it was unused since 1.0.0 even.


mseri:
Thanks

PR #25283, opened 2024-02-19
Release Caqti 2.1.0.

paurkedal:
I would like to release Caqti 2.1.0 ([release notes](https://github.com/paurkedal/ocaml-caqti/releases/tag/v2.1.0)).  This release includes one new package, `caqti-tls`.  There are actually 3 more TLS-related packages in the tarball, which I'm withholding for now, in the hope I of reducing the number of additional packages and 2 packages which are unchanged:

Package                 | 2.1.0
----------------------- | -----
caqti                   |  yes
caqti-async             |  yes
caqti-eio               |  yes
caqti-lwt               |  yes
caqti-mirage            |  yes
caqti-tls-async         |  no, omitted for now
caqti-tls-eio           |  no, omitted for now
caqti-tls-lwt           |  no, omitted for now
caqti-tls               |  yes
caqti-driver-mariadb    |  yes
caqti-driver-pgx        |  yes
caqti-driver-postgresql |  yes
caqti-driver-sqlite3    |  yes
caqti-dynload           |  no, unchanged
caqti-type-calendar     |  no, unchanged

I'll take a look at the CI results later today.

PR #25289, opened 2024-02-19
Release Caqti 2.1.1.

paurkedal:
(This replaces the just closed PR #25283.)

I would like to release Caqti 2.1.1 ([release notes](https://github.com/paurkedal/ocaml-caqti/releases/tag/v2.1.1)).  This release includes one new package, `caqti-tls`.  There are actually 3 more TLS-related packages in the tarball, which I'm withholding for now, in the hope I of reducing the number of additional packages and 2 packages which are unchanged:

Package                 | 2.1.1
----------------------- | -----
caqti                   |  yes
caqti-async             |  yes
caqti-eio               |  yes
caqti-lwt               |  yes
caqti-mirage            |  yes
caqti-tls-async         |  no, omitted for now
caqti-tls-eio           |  no, omitted for now
caqti-tls-lwt           |  no, omitted for now
caqti-tls               |  yes
caqti-driver-mariadb    |  yes
caqti-driver-pgx        |  yes
caqti-driver-postgresql |  yes
caqti-driver-sqlite3    |  yes
caqti-dynload           |  no, unchanged
caqti-type-calendar     |  no, unchanged

### DIFF
--- a/packages/caqti-driver-pgx/caqti-driver-pgx.2.1.1/opam
+++ b/packages/caqti-driver-pgx/caqti-driver-pgx.2.1.1/opam
@@ -1,0 +1,32 @@
+opam-version: "2.0"
+maintainer: "Petter A. Urkedal <paurkedal@gmail.com>"
+authors: [
+  "Petter A. Urkedal <paurkedal@gmail.com>"
+]
+license: "LGPL-3 with OCaml linking exception"
+homepage: "https://github.com/paurkedal/ocaml-caqti/"
+doc: "https://paurkedal.github.io/ocaml-caqti/index.html"
+bug-reports: "https://github.com/paurkedal/ocaml-caqti/issues"
+depends: [
+  "ocaml" {>= "4.08"}
+  "caqti" {>= "2.1.0" & < "2.2.0~"}
+  "domain-name"
+  "dune" {>= "3.9"}
+  "ipaddr"
+  "pgx" {>= "2.0"}
+]
+build: [
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+]
+dev-repo: "git+https://github.com/paurkedal/ocaml-caqti.git"
+synopsis: "PostgreSQL driver for Caqti based on the pure-OCaml PGX library"
+url {
+  src:
+    "https://github.com/paurkedal/ocaml-caqti/releases/download/v2.1.1/caqti-v2.1.1.tbz"
+  checksum: [
+    "sha256=483a535f41e2641917fc1832ce4ad15ffc3f4e8283b1b3018a2617349583090a"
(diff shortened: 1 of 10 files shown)


packages/caqti-driver-pgx/caqti-driver-pgx.2.1.1/opam
```diff
@@ -0,0 +1,32 @@
+opam-version: "2.0"
+maintainer: "Petter A. Urkedal <paurkedal@gmail.com>"
+authors: [
+  "Petter A. Urkedal <paurkedal@gmail.com>"
+]
+license: "LGPL-3 with OCaml linking exception"
```
mseri, line 6:
```suggestion
license: "LGPL-3 WITH OCaml linking exception"
```

paurkedal: It's actually supposed to be "LGPL-3.0-or-later WITH LGPL-3.0-linking-exception".  Thanks for pointing it out.

paurkedal:
Apart from the license specification, there is at least one issue left: One of the tests for the caqti package must be moved to another package to avoid a dependency cycle.  However, I think I'll omit it by tweaking the OPAM file instead of making another release, since the test coverage is good enough without it.  But I'll let the CI run out first in case it discovers more issues.

mseri:
Sure. To avoid circular dependencies you can add `post` to the dependency filter. It is brittle but works more often than not

paurkedal:
I wasn't familiar with `post`, but I don't think it will help here.  (The main test suite dynamically loads `caqti-driver-sqlite3`.  This is not an issue for the `caqti-{lwt,async,eio,mirage}` instantiations of the test suite, but it is for the blocking instantiation since it's currently attached to the `caqti` packages on which `caqti-driver-sqlite3` depends.)

I'll give it another try tomorrow.

packages/caqti/caqti.2.1.1/opam
```diff
@@ -0,0 +1,66 @@
...
+  "lwt-dllist"
+  "mtime" {>= "2.0.0"}
+  "ocaml" {>= "4.08.0"}
+  "odoc" {with-doc}
+  "ptime"
+  "re" {with-test}
+  "tls"
+  "uri" {>= "2.2.0"}
+  "x509"
+]
```
mseri, line 31:
```suggestion
  "caqti-driver-sqlite3" {post & with-test}
]
```
This should ensure that `caqti-driver-sqlite3` is available when running the tests, if I am not mistaken

paurkedal:
It seems like the `post` dependency [didn't work in this case](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/4b5b4a698f1d7a74e8feadcb7a1a8ed90d3491e0), so I pushed a commit to disable the problematic part of the testsuite.

Most of the errors yesterday was due to a more interesting issue:  When using dune-site, one needs to invoke `dune install`, since it performs substitution needed to find the site files.

I've been pushing new commits instead of amending, so we may want to squash before merging.

paurkedal:
I suggest this is ready to merge now.  Thanks for the help sorting out the issues.

PR #25297, opened 2024-02-20
Add conf-sqlite support for opam var os-family=ubuntu

jmid:
This little PR adds `conf-sqlite` support for Linux Mint and other derivative distributions with
```
$ opam var os-family
ubuntu
```


mseri:
Thanks

PR #25298, opened 2024-02-20
base 0.16 requires bash

AltGr:


mseri:
Thanks! This is due to https://github.com/janestreet/base/blob/ff1270481090f3f34db7d1d35d97144102bce4d0/src/dune#L51

PR #25313, opened 2024-02-22
ez-conf-lib.2

nberth:


mseri:
Please release this change as the `.2` version of the library

packages/ez-conf-lib/ez-conf-lib.1/opam
```diff
@@ -18,6 +18,6 @@ depends: [
   "conf-findutils"
 ]
 url {
-  src: "https://github.com/nberth/ez-conf-lib/archive/main.tar.gz"
-  checksum: "sha512=f223a9cd0ecd3dada1bc06f0ff1492eb776ce63b6c3f79f97f6e1246a85a8f159be45d6b5a6b44f42c21ba00a536f7e763d934a397fcd3428319095bb7cec6aa"
+  src: "https://github.com/nberth/ez-conf-lib/archive/1.tar.gz"
+  checksum: "sha512=4682cad22e7bd4502aaf021e225d8cb222bed9cbcdf34703b8a1824f26d6e459f71a6b97eaa1f05cc3e6af2ceab71a41b300b3ad3d99c429c09d313683677f92"
```
mseri, line 22: I have the original tarball with the correct checksum. Would it be ok if I send it to you and you replace `1.tar.gz` so that we can avoid modifying this checksum?

nberth: Ok if yuou wish. (I did not know I could just upload a tarball like this.)
I'm wondering, 

> I have the original tarball with the correct checksum. Would it be ok if I send it to you and you replace `1.tar.gz` so that we can avoid modifying this checksum?

If you prefer it this way (can I really upload a tarball like this on github?).
But, I'm wondering, at this stage I doubt there are many installations of that library. All that was quite temporary when I did the original PR, and I wanted to do some adjustments before a more "official" release, after some tests by users.

mseri: We had CIs and nix repositories break in the past when we changed checksum of packages we believed were not very used, so I'd rather stick with not changing the checksum. I'll send you a screenshot in the email to show you how to upload the tarball (it is what dune-release usually does for you for example).

I did not notice the package was pointing to `main` otherwise I would have told you already when you submitted the first PR, sorry about that.

PR #25338, opened 2024-02-24
[new release] hol2dk (1.0.0)

fblanqui:
HOL-Light to Dedukti/Lambdapi and Coq translator

- Project page: <a href="https://github.com/Deducteam/hol2dk">https://github.com/Deducteam/hol2dk</a>

### Added

- add-links script
- command rewrite to simplify prf files
- command purge to compute useless theorems
  these two commands greatly reduce the size of generated proofs
- command simp is equivalent to rewrite and purge
- command dump-simp is equivalent to dump, pos, use and simp
- allow simultaneous dumping
- alignments of the types option, Sum and list
- command split to generate a pos/use/sti/nbp file for each named theorem
  (an sti file contains the starting index of the corresponding proof)
- command theorem to generate the lp files corresponding to a named theorem
- Makefile to generate and check lp and coq files generated with split
- command size to get statistics on the size of terms
- option --print-stats to print statistics on hash tables at exit
- use hash tables instead of maps to build abbreviations maps
- use sharing when building canonical types and terms
- add option --use-sharing for using sharing in lp output when defining term abbreviations
- command nbp to get the number of proof steps
- commands patch, unpatch and link to call the corresponding scripts
- command env to print $HOL2DK_DIR and $HOLLIGHT_DIR

### Modified

- identifier renamings
- merged the command dg in the command mk
- fusion.ml: do not generate new theorems for empty instantiations

### Fixed

- valid dedukti and lambdapi identifiers

### DIFF
--- a/packages/hol2dk/hol2dk.1.0.0/opam
+++ b/packages/hol2dk/hol2dk.1.0.0/opam
@@ -1,0 +1,37 @@
+opam-version: "2.0"
+synopsis: "HOL-Light to Dedukti/Lambdapi and Coq translator"
+description: "HOL-Light to Dedukti/Lambdapi and Coq translator"
+maintainer: ["Frédéric Blanqui"]
+authors: ["Frédéric Blanqui" "Anthony Bordg"]
+license: "CeCILL-2.1"
+homepage: "https://github.com/Deducteam/hol2dk"
+bug-reports: "https://github.com/Deducteam/hol2dk/issues"
+depends: [
+  "ocaml" {>= "4.13"}
+  "dune" {>= "3.7"}
+  "odoc" {with-doc}
+]
+build: [
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
+dev-repo: "git+https://github.com/Deducteam/hol2dk.git"
+url {
+  src:
+    "https://github.com/Deducteam/hol2dk/releases/download/1.0.0/hol2dk-1.0.0.tbz"
+  checksum: [
+    "sha256=b9a304e56acda8367944c8590497ede7a386850b423432186309f36cc3aaed98"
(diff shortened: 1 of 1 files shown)


packages/hol2dk/hol2dk.1.0.0/opam
```diff
@@ -0,0 +1,37 @@
+opam-version: "2.0"
+synopsis: "HOL-Light to Dedukti/Lambdapi and Coq translator"
+description: "HOL-Light to Dedukti/Lambdapi and Coq translator"
+maintainer: ["Frédéric Blanqui"]
+authors: ["Frédéric Blanqui" "Anthony Bordg"]
+license: "CeCILL-2.1"
+homepage: "https://github.com/Deducteam/hol2dk"
```
haochenx, line 7: it seems that the `doc:` field is removed. please confirm whether this is intentional
https://github.com/ocaml/opam-repository/blob/0758a2c8add170e79d41976d4a9d2991dc0f7686/packages/hol2dk/hol2dk.0.0.1/opam#L8

haochenx (commented):
everything looks good to me except for the inline comment

packages/hol2dk/hol2dk.1.0.0/opam
fblanqui: Yes, indeed, because the opam file is generated by dune but I get an error with dune-release. I don't know why.

mseri:
```suggestion
homepage: "https://github.com/Deducteam/hol2dk"
doc: "https://github.com/Deducteam/hol2dk/blob/master/README.md"
```

mseri:
Thanks

packages/hol2dk/hol2dk.1.0.0/opam
haochenx: @fblanqui  I see. A `(document ..)` field in your `dune-project` file might help.

ref: https://dune.readthedocs.io/en/stable/dune-files.html#stanza-documentation

fblanqui:
Thank you!

PR #25355, opened 2024-02-26
[new release] mdx (2.4.0)

gpetiot:
CHANGES:

#### Added

- Handle the error-blocks syntax (realworldocaml/mdx#439, @jonludlam, @gpetiot)
- Allow execution of included OCaml code blocks (realworldocaml/mdx#446, @panglesd)
- Make MDX compatible with OCaml 5.2 (realworldocaml/mdx#448, @gpetiot)

#### Fixed

- Reduce false-positives while detecting warnings (realworldocaml/mdx#440, @Julow)

PR #25369, opened 2024-02-28
Add mdx constraints bc breaking changes

raphael-proust:
Because https://github.com/ocaml/opam-repository/pull/25355


mseri:
There are a number of build failures but they are independent of this PR

PR #25377, opened 2024-02-29
saturn_lockfree: disable tests

mseri:
They require saturn, which requires saturn_lockfree

Ping @polytypic, I think the issue with saturn_lockfree tests has not to do just with the other package but is that saturn_lockfree tries to run saturn tests as well.

### DIFF
--- a/packages/saturn_lockfree/saturn_lockfree.0.4.1/opam
+++ b/packages/saturn_lockfree/saturn_lockfree.0.4.1/opam
@@ -27,7 +27,6 @@ build: [
     "-j"
     jobs
     "@install"
-    "@runtest" {with-test}
     "@doc" {with-doc}
   ]
 ]


mseri:
Saturn tests are relayably deadlocking on 4.14 on the CI:
https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/de6821f45af64f6a027fb55fbb2f4263f7e2811f/variant/compilers,4.14,saturn_lockfree.0.4.1,revdeps,saturn.0.4.1

mseri:
This is not a regression in the sense that this used to be the case also before, we just did not know... I just wanted to let you know

polytypic:
> Ping @polytypic, I think the issue with saturn_lockfree tests has not to do just with the other package but is that saturn_lockfree tries to run saturn tests as well.

Yes, that is also true.  It [has been fixed](https://github.com/ocaml-multicore/saturn/pull/113) in the Saturn repo already, but no new package has been published.  We probably should publish a new package soon.

polytypic:
> Saturn tests are relayably deadlocking on 4.14 on the CI

Yes, I have [a PR to Saturn](https://github.com/ocaml-multicore/saturn/pull/128) that fixes the tests to work on OCaml 4 and then some.

PR #25381, opened 2024-03-01
Add conf-findutils sys dep for FreeBSD and Ubuntu derivatives

jmid:


jmid:
The CI failures are unrelated to the PR's changes.

mseri:
Thanks

hannesm:
First of all, thanks for working on getting more opam package work flawlessly on FreeBSD.

I have to admit I do not know the details of how `findutils` is used in dependencies -- but I have to mention that a FreeBSD base system comes with `find`, `xargs`, and `locate`.

The `findutils` package are the GNU findutils, and installed as `gfind`, `gxargs`, and `glocate`. Thus, if the users of `conf-findutils` use these names on FreeBSD, adding the depext sounds fine to me -- but my gut feeling is that the users of `conf-findutils` are fine using FreeBSD's `find` etc.

TL;DR: I propose to remove the depext for FreeBSD, and am happy to read what others think and what the concrete issue is that is solved (i.e. which opam package was broken before and compiles now fine on FreeBSD) by adding this depext.

mseri:
I am fine to revert the freebsd change until something fails

jmid:
Sorry, I missed that the BSD `findutils` package installed the GNU variants with the `g` prefix.
I'm fine with reverting that one.

I've just been taking a (too) quick pass over the top ~130 packages to improve the situation for FreeBSD and Ubuntu-derivatives (and a few other things that I spotted).

hannesm:
Thanks @mseri and @jmid. And indeed, @jmid thanks for your work on that :)

PR #25382, opened 2024-03-01
Expand conf-gssapi platform support

jmid:
This PR expands conf-gssapi platform support to
- Ubuntu derivatives
- Alpine
- Arch
- Fedora
- (Open)Suse
- FreeBSD


jmid:
The macOS and Oracle CI failures are expected as they are unsupported (for now).
The ocamlnet failures seem like an internal CI system failure.

I'm unsure why the OpenSuse installation fails.

mseri:
Thanks

PR #25384, opened 2024-03-01
Broaden conf-sdl-* support

jmid:
This PR broadens the support of the conf-sdl-{gfx,image,mixer,ttf,net} packages to support
- Ubuntu-derivatives
- Arch
- Fedora
- (Open)Suse
- FreeBSD


jmid:
The CI fails on Alpine, Oracle, and macOS - none of which the PR attempts to support.
As such, this should be good to go in.

mseri:
Thanks

PR #25385, opened 2024-03-01
Support conf-perl-* under FreeBSD and Ubuntu-derivatives

jmid:
This PR adds FreeBSD and Ubuntu-derivative support for
- conf-perl-ipc-system-simple
- conf-perl-string-shellquote


jmid:
Only an unrelated revdep failure remains: `frama-clang.0.0.14 (failed: ld returned 1 exit status)`
so this should be good to go.

mseri:
Thanks

PR #25386, opened 2024-03-01
Add Ubuntu derivative support for conf-{c++,g++}

jmid:


jmid:
Only unrelated revdeps on the CI, so this should be good to go

PR #25388, opened 2024-03-01
Use os-family for debian/ubuntu in conf-srt* packages

jmid:
This PR switches
- conf-srt
- conf-srt-gnutls
- conf-srt-openssl

to use `os-family` for debian and ubuntu in order to broaden the package to support debian and ubuntu derivatives.


mseri:
Thanks

PR #25389, opened 2024-03-01
Expand conf-* FreeBSD support

jmid:
This PR add FreeBSD support for
- conf-arb
- conf-flint
- conf-git
- conf-mad
- conf-python-3-dev

The latter includes `python3-config` according to https://freebsd.pkgs.org/13/freebsd-amd64/python3-3_3.pkg.html but I'm less sure of the `Python.h` header... :thinking:

### DIFF
--- a/packages/conf-arb/conf-arb.1/opam
+++ b/packages/conf-arb/conf-arb.1/opam
@@ -1,5 +1,4 @@
 opam-version: "2.0"
-version: "1"
 maintainer: "bobot"
 homepage: "http://arblib.org/"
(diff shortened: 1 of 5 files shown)


jmid:
There was one FreeBSD failure only, from conf-mad, failing on `pkg-config --exists mad`.
According to https://freebsd.pkgs.org/14/freebsd-aarch64/libmad-0.16.3.pkg.html the pkg-config-registered name may be `libmad` under FreeBSD, so cfd9875 adjusts the `build` test accordingly.

PR #25391, opened 2024-03-01
Fix conf-jack FreeBSD support

jmid:
conf-jack fails on FreeBSD on https://freebsd.check.ci.dev/ as it uses the wrong sys package name `jack`
The right name is `audio/jack` or `jackit` according to https://www.freshports.org/audio/jack


jmid:
This fails on Oracle Linux but is otherwise green. It should be good to go.

mseri:
Thanks

PR #25392, opened 2024-03-01
Fix conf-libopus FreeBSD support

jmid:
conf-libopus fails on FreeBSD on https://freebsd.check.ci.dev/ as it uses the wrong sys package name libopus
The right name is audio/opus or opus according to https://www.freshports.org/audio/opus/


jmid:
The FreeBSD runs now succeed.

I couldn't help but notice that OpenSuse was failing with what I think is due to the wrong sys package name.
I've thus changed it in d2fa50f to match the Suse entries I see on https://pkgs.org/search/?q=libopus

mseri:
Thanks

PR #25393, opened 2024-03-01
Fix conf-llvm.16 FreeBSD support

jmid:
conf-llvm.16 fails on FreeBSD on https://freebsd.check.ci.dev/

From the log https://freebsd.check.ci.dev/log/1708435041-55225ede92751fbc1bc34328c453e95a06a6cec6/5.1.0/bad/conf-llvm.16 one can see it repeatedly tries various `llvm-config` variations except `llvm-config16` (no dash before 16)
which should be the right name according to https://freebsd.pkgs.org/14/freebsd-aarch64/llvm16-16.0.6_10.pkg.html

### DIFF
--- a/packages/conf-llvm/conf-llvm.16/opam
+++ b/packages/conf-llvm/conf-llvm.16/opam
@@ -17,7 +17,8 @@ depexts: [
   ["llvm16-dev"] {os-distribution = "alpine"}
   ["llvm16"] {os-family = "arch"}
   ["llvm16-devel"] {os-family = "suse" | os-family = "opensuse"}
-  ["llvm16-devel"] {os-distribution = "fedora"}
+  ["llvm16-devel"] {os-distribution = "fedora" & os-version >= "39"}
+  ["llvm-devel-16"] {os-distribution = "fedora" & os-version < "39"}
   ["llvm16-devel" "epel-release"] {os-distribution = "centos"}
   ["devel/llvm16"] {os = "freebsd"}
 ]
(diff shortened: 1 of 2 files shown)


jmid:
84a3bdc updates the md5 hash to reflect that `configure.sh` was changed

jmid:
CI fails on OpenSuse, Fedora, and Oracle as well as on a revdep (frama-clang.0.0.14, unrelated).

a44691a should fix the situation for OpenSuse.

For Fedora, according to https://pkgs.org/search/?q=llvm
- on Fedora 39 the package name `llvm16-devel` as stated seems right
- on Fedora 38 (which along with 37 is tested on the CI) the name seems to be `llvm-devel-16` :shrug:

packages/conf-llvm/conf-llvm.16/opam
```diff
@@ -16,11 +16,11 @@ depexts: [
   ["llvm-16-dev" "zlib1g-dev" "libzstd-dev"] {os-family = "debian"}
   ["llvm16-dev"] {os-distribution = "alpine"}
   ["llvm16"] {os-family = "arch"}
-  ["llvm16-devel"] {os-family = "suse"}
+  ["llvm16-devel"] {os-family = "suse" | os-family = "opensuse"}
   ["llvm16-devel"] {os-distribution = "fedora"}
```
mseri, line 20:
```suggestion
  ["llvm16-devel"] {os-distribution = "fedora" & os-version >= "39"}
  ["llvm-devel-16"] {os-distribution = "fedora" & os-version < "39"}
```

jmid:
Hm. Fedora 38 still fails CI. Apparently the Fedora 38 package name is just `llvm-devel`, sorry:
https://fedora.pkgs.org/38/fedora-x86_64/llvm-devel-16.0.0-2.fc38.x86_64.rpm.html

On Fedora 37 `llvm-devel` will install version 15 though, according to https://rpmfind.net/linux/RPM/fedora/updates/37/x86_64/Packages/l/LByName.html
hence the package could be marked with 'not available' on that platform

PR #25430, opened 2024-03-08
Package slack.0.1

sewenthy:
### `slack.0.1`
Slack API implementation
OCaml interface for accessing Slack APIs and receiving events.

---
* Homepage: https://github.com/ahrefs/slack
* Source repo: git+https://github.com/ahrefs/slack.git
* Bug tracker: https://github.com/ahrefs/slack/issues

---

### DIFF
--- a/packages/slack/slack.0.1/opam
+++ b/packages/slack/slack.0.1/opam
@@ -1,0 +1,51 @@
+opam-version: "2.0"
+license: "MIT"
+maintainer: "Ahrefs <github@ahrefs.com>"
+authors: "Ahrefs <github@ahrefs.com>"
+homepage: "https://github.com/ahrefs/slack"
+bug-reports: "https://github.com/ahrefs/slack/issues"
+dev-repo: "git+https://github.com/ahrefs/slack.git"
+
+synopsis: "Slack API implementation"
+description: "OCaml interface for accessing Slack APIs and receiving events."
+
+depends: [
+  "dune" {>= "3.4"}
+  "atdgen" {> "2.0.0"}
+  "atdgen-runtime" {> "2.0.0"}
+  "biniou"
+  "cstruct"
+  "cmarkit"
+  "cmdliner"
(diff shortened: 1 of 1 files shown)


packages/slack/slack.0.1/opam
```diff
@@ -0,0 +1,36 @@
...
+authors: "Ahrefs <github@ahrefs.com>"
+homepage: "https://github.com/ahrefs/slack"
+bug-reports: "https://github.com/ahrefs/slack/issues"
+dev-repo: "git+https://github.com/ahrefs/slack.git"
+
+synopsis: "Slack API implementation."
+description: "OCaml interface for accessing Slack APIs and receiving events."
+
+depends: [
+  "dune"
```
mseri, line 13:
```suggestion
  "dune" {> "3.4"}
```
according to the dune-project file

yawaramin: If there is a dune-project file, isn't an explicit dune dependency redundant?

mseri: No, opam needs to know that dune has to be installed. If your opam files are dune generated, then you can omit it since dune adds it itself to the opam file, otherwise you need to manually add it with the correct lower bound

mseri:
@sewenthy can you upstream the fix to the opam file?

sewenthy:
@mseri I'm not sure what was the best way to do that--I added a new commit to the main repo and then run `opam publish` again, hope that's ok 🙏

mseri:
In general, for such changes, you can simply update the opam file in your repository without the need to re-release. But re-releasing as you did is fine as well

Khady:
@sewenthy the build instructions are also missing from this opam file

sewenthy:
@mseri just added to this PR and @Khady added to upstream already, thanks!

mseri:
Thanks. I'll merge this since easy format is tracked by a separate issue

PR #25447, opened 2024-03-10
Patch conf-binutils on FreeBSD

jmid:
https://freebsd.check.ci.dev/ indicates that `conf-binutils` fails on FreeBSD with `unsupported OS freebsd`

This PR attempts to
- install the relevant system package [containing `objdump`, `cxxfilt` and `readelf`](https://github.com/freebsd/freebsd-ports/blob/main/devel/binutils/pkg-plist)
- patch the test script accordingly

### DIFF
--- a/packages/conf-binutils/conf-binutils.0.3/files/find-binutils.ml.in
+++ b/packages/conf-binutils/conf-binutils.0.3/files/find-binutils.ml.in
@@ -207,7 +207,7 @@ let t =
 let () =
   try
     match "%{os}%" with
-    | "linux" -> write @@ collect t @@ all_paths ()
+    | "linux" | "freebsd" -> write @@ collect t @@ all_paths ()
     | "macos" ->
       let t = match cmd "brew --cellar" with
         | None -> t
--- a/packages/conf-binutils/conf-binutils.0.3/opam
+++ b/packages/conf-binutils/conf-binutils.0.3/opam
@@ -29,4 +29,4 @@ depends: [
 ]
 substs: [ "find-binutils.ml" ]
 flags: [ conf ]
-extra-files: ["find-binutils.ml.in" "md5=2bde83d0b7658cac4ffa113f9bcd87c2"]
+extra-files: ["find-binutils.ml.in" "md5=c58b89b0755853f68458f3421270af7e"]


jmid:
This is now green on FreeBSD

mseri:
ping @hannesm for the freebsd change

hannesm:
I've not followed the conf- packages much (apart from those that are in my dependency cone and annoying).

I also have to admit that FreeBSD ships in the base system with objdump and readelf (from llvm) -- but what is conf-binutils expected to deliver / ship with? If conf-binutils is supposed to check for GNU binutils, this change is fine. if only objdump / readelf / c++filt are required, only the patch for `find-binutils.ml.in` should be necessary.

jmid:
Thanks for the feedback Hannes - and sorry for the mess! I've dropped the first commit.

I'm no FreeBSD expert - just trying to improve the situation on https://freebsd.check.ci.dev/ :shrug: 

What's a good resource for searching the contents of the FreeBSD base system so that I can avoid making this mistake going forward? (A quick search suggested searching the source code tree https://svnweb.freebsd.org/base/head/)

hannesm:
An excellent question. The source repository indeed contains what is the base system of FreeBSD -- but there's unfortunately not a pkg-plist -- I put the output of `find /bin /sbin /usr/bin /sbin /usr/sbin` at https://data.robur.coop/freebsd14-base.txt -- maybe that is helpful?

There's as well some work in progress to use the pkg system for the base system -- https://wiki.freebsd.org/PkgBase -- but I couldn't find pkg-plist for these packages.

If in doubt, don't hesitate to ping me in issues / PRs :)

mseri:
Thanks to both of you

jmid:
Thanks! :pray: 

It seems FreeBSD fails on the CI with `objdumps not found`, which indicates that it is not finding the `objdump` executable.

I spotted the following in `find-binutils.ml.in` explaining why `conf-binutils` fails on FreeBSD when finding the `llvm-`-prefixed version: `llvm-objdump`:
```ocaml
let has_prefix str pref =
  let len = String.length pref in
  len <= String.length str &&
  String.sub str 0 len = pref

let is_llvm_tool file = has_prefix file "llvm"

let check ~suffix file =
  Filename.check_suffix file suffix &&
  not (is_llvm_tool file)
```

I haven't looked carefully enough to understand whether this is accidental, e.g., because `/usr/bin/llvm-objdump` is processed before `/usr/bin/objdump`, and whether the intention is a hard requirement of incompatibility with the LLVM-variants.

hannesm:
The issue arising here:
- until FreeBSD 12, they shipped GNU binutils (and llvm-binutils with the llvm- prefix)
- starting in FreeBSD 13 (what eventually the CI systems are using, very unclear which version of FreeBSD the CI machine is actually testing and why), they stopped shipping GNU binutils - so there's only llvm-objdump in the base system
- starting in FreeBSD 14, the llvm- prefix got removed, and /usr/bin/objdump is now the llvm objdump

I've no idea what conf-binutils is expecting (GNU tools or whether the llvm ones are fine as well -- likely should be tested / evaluated for the reverse dependencies of conf-binutils), and which version of FreeBSD the CI is testing (and what the reasoning behind it is).

jmid:
Thanks Hannes! 

Regarding the tested version, I believe it is 13.2: https://infra.ocaml.org/by-use/freebsd-x86_64 found from [this blog post](https://tarides.com/blog/2023-10-04-porting-obuilder-to-freebsd/):
> _Currently we have a single server running FreeBSD 13.2 providing OCaml 4.14 and 5.1 builds on x86_64. This modest Dual Xeon machine (16 cores) is easily handling the load from opam-repo-ci, ocaml-ci, and opam-health-check, [...]_

mseri:
I think we can leave it as is for now, and wait until somebody uses bap from freebsd and tells us what is needed. Of course feel free to work on this if you want

PR #25448, opened 2024-03-10
Patch conf-clang on Ubuntu-derivatives and FreeBSD

jmid:
This conf-clang PR
- adds support for Ubuntu derivatives
- adds an explicit `llvm` dependency on FreeBSD


jmid:
Still green on FreeBSD

mseri:
ping @hannesm for the freebsd change

hannesm:
Again, I've no clue which package uses conf-clang -- but FreeBSD in the base system ships clang :) (with FreeBSD 14, clang 16 is shipped). Any idea which binaries or files are required for users of conf-clang?

mseri:
I think just `bap-std`

mseri:
No worries, I pinged Hannes because I also don't know enough BSDs but I thought clang would have been there

mseri:
Thanks to both of you!

PR #25449, opened 2024-03-10
conf-libsodium: support Ubuntu-derivatives, Arch, Fedora, and Open/Suse

jmid:


jmid:
CI fails on Oracle, which is expected.

Looking more closely, on OpenSuse it is listed as `conf-libsodium.1 (failed: [SKIP] Package not available)`. I'm unsure why though.

mseri:
I don't know but at least is informative for suse users incurring in the failure

PR #25451, opened 2024-03-10
conf-tk patches

jmid:
For conf-tk this PR
- add Ubuntu derivative support
- removes a duplicate `alpine` entry which listed both `tk` and `tk-dev` as dependencies in separate lines. I believe the latter one, containing the headers is the correct one


jmid:
This is still green on alpine, so it should be good to go.

PR #25452, opened 2024-03-10
conf-m4: add support for Ubuntu derivatives

jmid:
For conf-m4 this PR
- adds support for Ubuntu derivatives
- documents that m4 [is included in FreeBSD's base system](https://svnweb.freebsd.org/base/head/usr.bin/m4/)


jmid:
This PR doesn't change anything for systems tested on the CI

PR #25322, opened 2024-02-22
Update the OCaml 5.2 preview for merlin

kit-ty-kate:

PR #25453, opened 2024-03-11
Update ppxlib.0.32.1~5.2preview

NathanReb:
Preview release with OCaml 5.2 compatibility


NathanReb:
CC @kit-ty-kate 

Should I mark the previous preview release as `available: false`?

kit-ty-kate (changes requested):
Adding a new version should not be necessary, the point of a preview release is also to avoid having to add more packages than necessary in opam-repository. Could you instead simply upgrade the url and checksum of 0.32.1~5.2preview with that of the new commit?

kit-ty-kate:
An example of that can be found in https://github.com/ocaml/opam-repository/pull/25322

NathanReb:
Ah yes sure, my bad!

kit-ty-kate (approved):

NathanReb:
Nothing on my side. I haven't gone through all the failures but from a quick look I agree they do seem familiar!

We would really need for opam-ci to be able to detect when a failure is actually triggered by the new release and when it's not. Having so many false positive really makes it hard and time consuming to actually find and fix real errors.

PR #25440, opened 2024-03-08
Add Visual Studio conf packages (conf-msvc32.1, conf-msvc64.1, msvs-detect.0.7.0)

dra27:
These packages are used as part of my proposal for packaging the MSVC port of OCaml in opam-repository, but the packages themselves are entirely independent.

`msvs-detect` packages two scripts from [metastack/msvs-tools](https://github.com/metastack/msvs-tools) which are then used by `conf-msvc32.1` and `conf-msvc64.1`.

The two conf- packages can be simultaneously installed. They provide environment settings which can then be used by other packages, either in `build-env` or `setenv`:

```
C:\Devel>opam config list conf-msvc64
...
conf-msvc64:package   Visual Studio Community 2022 (17.9.34511.98)
conf-msvc64:script    "C:\Program Files\Microsoft Visual Studio\2022\Preview\VC\Auxiliary\Build\vcvarsall.bat" x64
conf-msvc64:ml        ml64
conf-msvc64:msvs-bin  C:\Program Files\Microsoft Visual Studio\2022\Preview\VC\Tools\MSVC\14.39.33428\bin\HostX64\x64;C:\Program
                      Files\Microsoft Visual Studio\2022\Preview\Common7\IDE\VC\VCPackages;C:\Program Files\Microsoft Visual
                      Studio\2022\Preview\Common7\IDE\CommonExtensions\Microsoft\TestWindow;C:\Program Files\Microsoft Visual
                      Studio\2022\Preview\Common7\IDE\CommonExtensions\Microsoft\TeamFoundation\Team Explorer;C:\Program
                      Files\Microsoft Visual Studio\2022\Preview\MSBuild\Current\bin\Roslyn;C:\Program Files (x86)\Microsoft Visual
                      Studio\Shared\Common\VSPerfCollectionTools\vs2019\x64;C:\Program Files (x86)\Microsoft Visual
                      Studio\Shared\Common\VSPerfCollectionTools\vs2019\;C:\Program Files (x86)\Microsoft
                      SDKs\Windows\v10.0A\bin\NETFX 4.8 Tools\x64\;C:\Program Files\Microsoft Visual Studio\2022\Preview\Team
                      Tools\DiagnosticsHub\Collector;C:\Program Files (x86)\Windows Kits\10\bin\10.0.22621.0\x64;C:\Program Files
                      (x86)\Windows Kits\10\bin\x64;C:\Program Files\Microsoft Visual
[...]

PR #25454, opened 2024-03-11
Add mingw-w64-shims.0.1.0

dra27:
This package is the mingw-w64 equivalent of #25440. It ensures that the `sys-root` bin directory for the required compiler gets added to `PATH` and also sets up shims for the internal Cygwin installation in opam 2.2.0.


dra27:
As with #25442, it's separately reviewable (and can be being reviewed), but doesn't need to be merged until just prior to its approval.

PR #25458, opened 2024-03-12
[new release] ppx_deriving (5.3.0)

NathanReb:
Type-driven code generation for OCaml

- Project page: <a href="https://github.com/ocaml-ppx/ppx_deriving">https://github.com/ocaml-ppx/ppx_deriving</a>
- Documentation: <a href="https://ocaml-ppx.github.io/ppx_deriving/">https://ocaml-ppx.github.io/ppx_deriving/</a>

##### CHANGES:

* Port standard plugins to ppxlib registration and attributes
  ocaml-ppx/ppx_deriving#263
  (Simmo Saan)

* Optimize forwarding in eq and ord plugins
  ocaml-ppx/ppx_deriving#252
  (Simmo Saan)

* Delegate quoter to ppxlib
  ocaml-ppx/ppx_deriving#263
  (Simmo Saan)

* Introduce `Ppx_deriving_runtime.Stdlib` with OCaml >= 4.07.
  This module already exists in OCaml < 4.07 but was missing otherwise.
  ocaml-ppx/ppx_deriving#258
  (Kate Deplaix)

### DIFF
--- a/packages/ppx_deriving/ppx_deriving.5.3.0/opam
+++ b/packages/ppx_deriving/ppx_deriving.5.3.0/opam
@@ -1,0 +1,40 @@
+opam-version: "2.0"
+maintainer: "whitequark <whitequark@whitequark.org>"
+authors: [ "whitequark <whitequark@whitequark.org>" ]
+license: "MIT"
+homepage: "https://github.com/ocaml-ppx/ppx_deriving"
+doc: "https://ocaml-ppx.github.io/ppx_deriving/"
+bug-reports: "https://github.com/ocaml-ppx/ppx_deriving/issues"
+dev-repo: "git+https://github.com/ocaml-ppx/ppx_deriving.git"
+tags: [ "syntax" ]
+build: [
+  ["dune" "subst"] {dev}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+  ["dune" "build" "@doc" "-p" name "-j" jobs] {with-doc}
+]
+depends: [
+  "ocaml" {>= "4.05.0"}
+  "dune" {>= "1.6.3"}
+  "cppo" {>= "1.1.0" & build}
+  "ocamlfind"
+  "ppx_derivers"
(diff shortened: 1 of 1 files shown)


sim642:
As I feared, https://github.com/ocaml-ppx/ppx_deriving/pull/263 leads to a lot of revdep breakage, despite the effort to maintain compatibility.

So far the big one appears to be ppx_let which declares the `map` extension and ppx_deriving.5.3.0 which declares the `derive.map` extension. I wonder how these things have co-existed so far and what could be done about it. `[%map: ...]` from ppx_deriving doesn't appear to be used on sherlocode (not sure what the use case for it would be), so maybe it could be removed.

NathanReb:
It appears the main problem is the `with_path` argument to ppx_deriving.show thing, I'll look into fixing that.

NathanReb:
The conflicting extension thing should be fixable with prefixes, I'm a bit surprised it's causing trouble at that level.

Is it because `ppx_let` is recording it without a prefix? If so I'll open a PR on `ppx_let`. That won't solve everything I guess as there could be conflicts from the user's perspective, i.e. some code might have to invoke the extension with a prefix but I guess that's okay.

Either way we'll have to cancel this release for now but we should be able to come up with a couple fixes that should make it usable. Probably worth releasing it as a 6.0 instead if it does break some stuff.

sim642:
> It appears the main problem is the `with_path` argument to ppx_deriving.show thing, I'll look into fixing that.

ppx_deriving has tests for it though, which makes this odd.

sim642:
The issue with `with_path` is when it's used for deriving in signatures. In my PR, `intf_generator` for show doesn't take arguments, only `impl_generator` does.
I guess we should just allow the argument also in signatures, even though it has no effect there.

NathanReb:
Yeah we probably should as it apparently used to define it there before. I'll work on a fix!

In the meantime I opened https://github.com/janestreet/ppx_let/issues/14!

NathanReb:
Let's close this for now, we'll re-release once we have fixed everything!

PR #25474, opened 2024-03-12
[new release] dune (15 packages) (3.14.2)

Leonidas-from-XIV:
Fast, portable, and opinionated build system

### Fixed

- fix compilation on non-glibc systems due to `signal.h` not being pulled in
  spawn stubs. (ocaml/dune#10256, @emillon)

### DIFF
--- a/packages/chrome-trace/chrome-trace.3.14.2/opam
+++ b/packages/chrome-trace/chrome-trace.3.14.2/opam
@@ -1,0 +1,40 @@
+opam-version: "2.0"
+synopsis: "Chrome trace event generation library"
+description:
+  "This library offers no backwards compatibility guarantees. Use at your own risk."
+maintainer: ["Jane Street Group, LLC <opensource@janestreet.com>"]
+authors: ["Jane Street Group, LLC <opensource@janestreet.com>"]
+license: "MIT"
+homepage: "https://github.com/ocaml/dune"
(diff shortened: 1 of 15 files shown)


kit-ty-kate:
Note for opam-repository maintainers: this release fixes a critical issue for dune on Windows (support of unicode characters in PATH, e.g. usernames with accents) and i would suggest all the older versions of dune to be restricted to `available: os != "win32"` in a separate PR, independently of this one (this one should be merged first though).

emillon:
I don't understand the reasoning. This is something important to fix, but this is not a regression or anything. This is something that has never worked. The old versions of dune are working well for a bunch of windows users. Why force them to upgrade? The solver will pick the new version anyway.

kit-ty-kate:
Well there was never any Windows users using opam-repository as opam-repository is not currently available on Windows. People currently use forks that are not currently updated.
So for the opam-repository support on Windows that is coming soon, the reasoning is that it would be nice for every user to have a dune version that works regardless of whether or not they have unicode characters in their environment.

Leonidas-from-XIV:
We evaluated the revdep failures and didn't see any regressions that were due to changes between 3.14.0 and 3.14.2 so this should be ok to merge.

raphael-proust:
Thank you!

PR #25477, opened 2024-03-12
ocaml-migrate-parsetree.2.3.0 is a 404 right now

emberian:
I don't know if it is due to some GitHub fault or some intentional activity somehow, but this link is busted right now.

Use the (equivalent?) github-made tarball.


sim642:
The old link seems to be fine now.

GitHub-made archives aren't reliable: GitHub does not guarantee that they remain unchanged, so their hashes might spuriously change.

mseri:
Thanks for the PR, it looks like it was a temporary issue with github. I am going to close for the same reasons mentioned by @sim642

PR #25482, opened 2024-03-13
Add missing conf-c++ dependencies to mccs

dra27:
#19829 added the `conf-c++` dependency to the mccs packages, but as it wasn't upstreamed back to the source repo, the releases since then have missed it. I've upstreamed it in https://github.com/ocaml-opam/ocaml-mccs/pull/52 and here are the updates to the 3 packages which are missing it.


mseri:
Thanks!

PR #25487, opened 2024-03-13
Add FreeBSD support for conf-antic

jmid:


jmid:
CI is green on FreeBSD!
It fails on Arch, OpenSuse, Oracle, and macOS.

PR #25495, opened 2024-03-13
Add conf-radare2 support for FreeBSD

jmid:


jmid:
CI fails on Suse and Oracle (understandably)

jmid:
d572b97 is a stab at fixing the Open/Suse failure.

jmid:
CI now only fails on Oracle (and on opam.2.0)

Issue #25498, opened 2024-03-14
iri 0.6.0 checksum changed


mseri:
```
#=== ERROR while fetching sources for iri.0.6.0 ===============================#
OpamSolution.Fetch_fail("https://framagit.org/zoggy/ocaml-iri/-/archive/0.6.0/ocaml-iri-0.6.0.tar.bz2 (Bad checksum, expected md5=40574cd88a7fc984adae1839fedb2b10)")
```
See https://github.com/ocaml/opam-repository/pull/25486, link to the old tarball is in the PR comments (cc @zoggy who is already aware)

PR #25499, opened 2024-03-14
iri: change archive urls to be more stable

zoggy:
Urls now points to real archives, no on-the-fly export from gitlab.


mseri:
That was fast, thanks!
Fixrd #25498

PR #25501, opened 2024-03-14
add upper bounds for asn1-combinators, there'll be a new release which breaks API

hannesm:
(a) instead of cstruct.t, string and bytes are used
(b) the representation of integer is now a string (not a Z.t)


mseri:
Lint is fine, thanks!

PR #25510, opened 2024-03-15
[new release] Docfd 3.0.0

darrenldl:

https://github.com/darrenldl/docfd

Think interactive grep for text files, PDFs, DOCXs, etc,

and PDF viewers,

Features:


darrenldl:
Following the dune's guide on cram tests is not bullet proof I guess. I'm removing the runtest line from the opam file.

mseri:
cram tests don't always play nice with the sandbox

mseri:
Thanks

darrenldl:
Yeah I couldn't get it to find `docfd` command without doing `dune build @all` etc first on my end either. It's weird that the cram test dep clause also doesn't seem to trigger a rebuild of the exe? Anyhow.

PR #25506, opened 2024-03-14
Create `conf-*` packages for OpenGL related system libraries

Leonidas-from-XIV:
Given the depexts for OpenGL can be kinda hairy with complicated dependency names, this PR attempts to move the dependencies of `lablgl` into its own `conf-*` packages (and updates raylib to use them where there is agreement).

The libraries in question:

* libGL
* libGLU (part of Mesa but still a separate library)
* freeglut (implementation of GLUT)

PR #25511, opened 2024-03-15
Replace `oraclelinux` with `ol`

Leonidas-from-XIV:
As @jmid kindly pointed out in #25506 the `os-distribution` value for Oracle Linux seems to be `ol`. In opam-repository some packages use `oraclelinux` which is presumably always false and confusing at best so this PR removes the mentions.

It also changes the logic of `conf-libgccjit` which seemed to check that the distribution is Oracle Linux OR the version was 9, which would filter out distributions with lower version numbers accidentally. I assume the idea was to make it not available on Oracle Linux <9.


mseri:
Thanks

PR #25512, opened 2024-03-15
Add flexdll.0.35-flexdll.0.43 and winpthreads.20240209-1

dra27:
This I think is the last prequisite before opening the "main" PR updating the compiler packages for Windows support.

Both of these packages provide sources needed in _addition_ to the compiler's own sources when compiling on Windows. Windows OCaml has required [FlexDLL](https://github.com/ocaml/flexdll) since OCaml 3.11, but it had to be installed separately. Adding to the complexity, the main part of FlexDLL is written in OCaml. Since OCaml 4.03.0, it has been possible to build OCaml and FlexDLL simultaneously, but the compiler does not distribute the required sources, and those what this flexdll package provides (this required support in FlexDLL was first added in 0.35, which is the reason for the older releases going back no further - these are the only releases there can have ever been).

OCaml 5.3.0 will introduce a requirement for the MSVC ports of OCaml in a very similar way with the [winpthreads](https://github.com/ocaml/winpthreads) library.

These packages simply install required sources to the `share` directory of the switch which are then picked up by the `configure` script of OCaml (using it's `--with-flexdll` and `--with-winpthreads-msvc` options). The packages are not constrained to Windows as the _opam packages_ aren't particularly Windows-specific (and, indeed, at some point cross-compilers will certain want the FlexDLL sources).

flexdll 0.35-0.43 are already released, which is why they reference `flexdll.install` from the master branch of the repository. When it's released, FlexDLL 0.44's tarball will include the .install file.

### DIFF
--- a/packages/flexdll/flexdll.0.35/opam
+++ b/packages/flexdll/flexdll.0.35/opam
@@ -1,0 +1,24 @@
+opam-version: "2.0"
+authors: "Alain Frisch"
+maintainer: "David Allsopp <david@tarides.com>"
+bug-reports: "https://github.com/ocaml/flexdll/issues"
+dev-repo: "git+https://github.com/ocaml/flexdll.git"
+homepage: "https://github.com/ocaml/flexdll#readme"
+license: "Zlib"
+synopsis: "FlexDLL Sources"
+description: "Source package for FlexDLL. Installs the required files for
+bootstrapping FlexDLL as part of an OCaml build."
+url {
+  src: "https://github.com/ocaml/flexdll/archive/refs/tags/0.35.tar.gz"
+  checksum: [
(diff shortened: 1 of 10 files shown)


dra27:
Obviously OCaml 5.3.0 is not yet even branched, but the winpthreads package is already needed for `ocaml-variants.5.3.0+trunk`, and I have a proposed back-port of the MSVC port to 5.2.0 I propose to make available in the same way as the `+tsan` variants.

dra27:
> The packages are not constrained to Windows as the _opam packages_ aren't particularly Windows-specific (and, indeed, at some point cross-compilers will certain want the FlexDLL sources).

How embarrassing, I only tested these packages on Windows 😉 It turns out the winpthreads package _is_ Windows-specific, which I've updated it to reflect!

mseri:
Thanks

PR #25521, opened 2024-03-18
Package duppy.0.9.4

toots:
### `duppy.0.9.4`
Library providing monadic threads

---
* Homepage: https://github.com/savonet/ocaml-duppy
* Source repo: git+https://github.com/savonet/ocaml-duppy.git
* Bug tracker: https://github.com/savonet/ocaml-duppy/issues

---


arbipher:
I am lost on this [log](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/421eb5d111de581862364d2d7b3b5a4ee710faa1/variant/compilers,4.07,duppy.0.9.4,lower-bounds) why after using the solver `builtin-0install`,  `opam reinstall duppy.0.9.4` will have a downgrade for `dune` and `re`? Is it caused by the opam internal solver prefers the latest package version (e.g. dune at 3.14.2) while `0install` prefers something package version (e.g. dune at 2.8.5).

mseri:
@arbipher that is the lower-bounds test, it tries to install the oldest possible versions of the packages to ensure that the lower bounds in the metadata are correct

packages/duppy/duppy.0.9.4/opam
```diff
@@ -0,0 +1,37 @@
...
+synopsis: "Library providing monadic threads"
+maintainer: ["The Savonet Team <savonet-users@lists.sourceforge.net>"]
+authors: ["Romain Beauxis <toots@rastageeks.org>"]
+license: "LGPL-2.1-only"
+homepage: "https://github.com/savonet/ocaml-duppy"
+bug-reports: "https://github.com/savonet/ocaml-duppy/issues"
+depends: [
+  "ocaml" {>= "4.07.0"}
+  "dune" {>= "2.7"}
+  "re"
```
mseri, line 12:
```suggestion
  "re" {>= "1.7.2"}
```
This is where `Re_pcre` (and similar) became `Re.Pcre`

mseri:
The failure with liquidsoap may be due an incompatibility between jbuilder and dune

mseri:
Let's see if it happens again

packages/duppy/duppy.0.9.4/opam
arbipher: I see. Thanks!

arbipher:
It's still incorrect. May I ask about _who (which step)_ reports this error?

```
[ERROR] The compilation of liquidsoap-core.2.2.0 failed at "dune build -p liquidsoap-core -j 39 --promote-install-files=false @install".
 (score = 20)
2024-03-19 09:15.57: >>> 
# Error: Unbound module Pcre
 (score = 48)
2024-03-19 09:15.57: Unbound module Pcre
```

mseri:
This is a dependendant (liquidsoap-core) that has issues, it is not related to this package

toots:
Hi all! Let me know if I can help. This release is a one-liner from the previous one..

mseri:
I will address liquidsoap on a separate PR

mseri:
@toots the issue is here: https://github.com/savonet/liquidsoap/blob/main/src/core/dune#L24

Before, `duppy` was linking `pcre` and passing it transitively to `liquidsoap-core`. Now that you use `re`, you need to explicitly link against `pcre` or to use `re` (and it would be better also in this case to explicitly link against it)

PR #25536, opened 2024-03-20
dune 3.x: fix windows bounds

mseri:
Opening since more users are hitting this issue

see https://github.com/ocaml/opam-repository/pull/25484


mseri:
Lint is green, I’ll spare the load to the ci

dra27:
Looks perfect, thanks! I guess this pattern `os != "win32" | os-distribution = "cygwinports"` is a rune we may have to get used to 🙂

dra27:
I have been musing if CI could assist here, but I’m not sure it really would - I’m very loathe for us to spend time/energy actually testing opam-repository-mingw

PR #25539, opened 2024-03-21
Package memprof-limits.0.2.1

gadmm:
### `memprof-limits.0.2.1`
Memory limits, allocation limits, and thread cancellation
Memprof-limits provides per-thread global memory limits, per-thread
allocation limits, and cancellation of threads, with ways to
ensure resource-safety after interruption.

Global memory limits let you bound the memory consumption of a task,
in terms of the major heap size.

Allocation limits let you bound the execution time of a task measured
in number of allocations. Allocation limits do not count
deallocations, and are therefore a measure of the work done, which can
be more suitable (reliable, portable, deterministic) than wall-clock
time.

Token limits lets you cancel a (CPU-bound) task preemptively and at a
distance.

Tasks are interrupted by raising an asynchronous exception.
Memprof-limits provides resource-management features and guidance for
reasoning about the consistency of state in the presence of such
interrupts.

The implementation uses OCaml's Statmemprof engine with a low sampling
rate that does not affect performance. A reimplementation of the
Memprof interface compatible with Memprof-limits running at the same
time is provided for profiling needs.

---
* Homepage: https://gitlab.com/gadmm/memprof-limits/
* Bug tracker: https://gitlab.com/gadmm/memprof-limits/issues/

---

### DIFF
--- a/packages/memprof-limits/memprof-limits.0.2.1/opam
+++ b/packages/memprof-limits/memprof-limits.0.2.1/opam
@@ -1,0 +1,51 @@
+opam-version: "2.0"
+maintainer: "Guillaume Munch-Maccagnoni <Guillaume.Munch-Maccagnoni@inria.fr>"
+authors: [
+  "Guillaume Munch-Maccagnoni <Guillaume.Munch-Maccagnoni@inria.fr>"
+]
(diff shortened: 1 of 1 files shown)


mseri:
Thanks

PR #25540, opened 2024-03-21
conf-alsa: add FreeBSD support

jmid:
Based on https://www.freshports.org/audio/alsa-lib/


jmid:
Wow, all green in the first attempt! :smiley:

mseri:
Thanks

PR #25541, opened 2024-03-21
conf-bluetooth: Add support for Ubuntu-derivatives

jmid:


hannesm:
conf-bluetooth is used for mindstorms, and that seems to be developed with linux in mind. unless you're keen to get mindstorms on a FreeBSD machine, I wouldn't bother... (maybe mark conf-bluetooth as `available: [ os != "freebsd" ]` with a comment about it's usage of `include <bluetooth/bluetooth.h>` etc -- which is very linux-specific.

PR #25544, opened 2024-03-21
conf-clang-format: add FreeBSD and Ubuntu-derivative support

jmid:
For FreeBSD https://www.freshports.org/devel/llvm lists an explicit `bin/clang-format` in pkg-plist
whereas Hannes's base list contains no such thing: https://data.robur.coop/freebsd14-base.txt


jmid:
CI fails only on Alpine (3.19).

1f61d6f adds support for Alpine 3.18-3.19 based on https://pkgs.org/download/clang-extra-tools

jmid:
CI all green!

PR #25545, opened 2024-03-21
conf-csdp: add support for FreeBSD and Ubuntu-derivatives

jmid:
Source: https://www.freshports.org/math/csdp/


jmid:
CI fail on FreeBSD - and also Alpine, Arch, OpenSuse, Oracle, and macOS.

7795765 corrects the FreeBSD port name and
c278534 correct the `os-distribution` for Arch

The result will still fail on Alpine, OpenSuse, Oracle, and macOS :shrug:

mseri:
Thanks

PR #25546, opened 2024-03-21
conf-dssi: add support for FreeBSD and Ubuntu-derivatives

jmid:
Source: https://www.freshports.org/audio/dssi/

For now, I've not added an `conf-alsa` dependency on FreeBSD (depends on #25540) because it is unclear to me whether it is needed, as there is no 'package test' included in `conf-dssi`. I suppose we will find out when rev-deps fail on FreeBSD... :shrug:

### DIFF
--- a/packages/conf-dssi/conf-dssi.1/opam
+++ b/packages/conf-dssi/conf-dssi.1/opam
@@ -8,12 +8,12 @@ depends: [
   "conf-alsa" {os = "linux"}
 ]
 depexts: [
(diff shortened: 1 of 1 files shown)


jmid:
CI fails on macOS and is all-green on FreeBSD.
No revdeps are triggered though. As such, I think this should be good to go.

mseri:
Thanks, I have opened an issue for the macos failure https://github.com/ocaml/infrastructure/issues/106

PR #25548, opened 2024-03-21
Add Z3 4.13.0

wintersteiger:


wintersteiger:
There are a couple of linking failures on RISC V platforms (undefined reference to `__atomic_exchange_1'). I don't know those platforms at all, so I can't tell whether that's a build config problem or a platform issue. Since they are marked as `extra`, I guess the failures are not a blocker, but I'll ask over at the Z3 side to see whether there's an easy fix for this.

packages/z3/z3.4.13.0/opam
```diff
@@ -0,0 +1,43 @@
...
+  "conf-c++" {build}
+]
+x-ci-accept-failures: [
+  "centos-7" "oraclelinux-7" # C compiler is too old
+]
+conflicts: [
+  "ocaml-option-bytecode-only"
+]
+synopsis: "Z3 solver"
+url {
```
mseri, line 36:
```suggestion
available: "arch" != "riscv64"
url {
```
we can mark in unavailable on risc V, however the same issue was probably already there in the previous releases

mseri:
The issues are known (same problem as easy-format). The new images with the new opam 2.2 version will not have this problem, it should be fixed next week

packages/z3/z3.4.13.0/opam
wintersteiger: Yeah, that's very likely! I'll try and see whether I can get a VM up to reproduce the problem.

wintersteiger:
The `riscv64-ocaml-4.14` build is very close to the timeout of 120m, after restarting it a couple of times, it finally did finish in time. You might want to consider raising the timeout for the RISC V builds.

mseri:
Thanks

PR #25551, opened 2024-03-21
conf-flex: add support for Ubuntu derivatives

jmid:


jmid:
There are two revdeps failures on `gappa.1.4.{0,1}`
These are not caused by this PR though, which has no effect on the tested CI systems.
As such, this should be good to merge.

mseri:
gappa failure will be fixed by the new opam 2.2 update, likely next week

PR #25553, opened 2024-03-21
conf-ftgl: add support for Arch, Alpine, Suse, FreeBSD, and Ubuntu-derivatives

jmid:


jmid:
CI fails on Oracle and macOS - but is otherwise green!
This should be OK to merge.

PR #25554, opened 2024-03-21
conf-glfw3: add FreeBSD support

jmid:
Source: https://www.freshports.org/graphics/glfw/


jmid:
CI fails on Oracle, but is otherwise green! :smiley:

PR #25555, opened 2024-03-21
conf-glpk: add FreeBSD support

jmid:
Source: https://www.freshports.org/math/glpk/

### DIFF
--- a/packages/conf-glpk/conf-glpk.1/opam
+++ b/packages/conf-glpk/conf-glpk.1/opam
@@ -18,6 +18,7 @@ depexts: [
   ["glpk-devel"] {os-family = "suse" | os-family = "opensuse"}
   ["glpk"] {os-family = "arch"}
   ["glpk"] {os = "macos" & os-distribution = "homebrew"}
+  ["glpk"] {os = "freebsd"}
 ]
 x-ci-accept-failures: [
   "oraclelinux-7" # package not present by default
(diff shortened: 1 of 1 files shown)


jmid:
CI failed on Oracle, FreeBSD, and macOS.
The latter two because it couldn't find the header file.
cb39314 is an attempt at adding a proper include flag.

mseri:
I think homebrew on arm64 installs the files in `/opt/local` or `/opt/homebrew/local`. If it comes with a pc file, maybe on mac it is better to use pkg-config

jmid:
9503f91 fixes the macOS situation.

I didn't go for a pkg-config solution, because doing that only for some platforms, e.g., macOS, but not on others, e.g., FreeBSD (https://www.freshports.org/math/glpk/ is missing a `.pc` file AFAICS) felt inconsistent :shrug: 

Modulo Oracle, this is now all green! :smiley:

jmid:
1c04c74 extends `x-ci-accept-failures` to include oraclelinux-9 too.
The result should thus be all green.

mseri:
Thanks

packages/conf-glpk/conf-glpk.1/opam
```diff
@@ -6,7 +6,9 @@ authors: ["Andrew Makhorin" "GNU"]
 license: "GPL-3.0-or-later"
 build: [
   ["sh" "-c" "echo '#include <glpk.h>' > test.c"]
-  ["cc" "-E" "test.c"]
+  ["cc" "-E" "test.c"] {os = "linux"}
+  ["cc" "-E" "-I/usr/local/include" "test.c"] {os = "freebsd"}
+  ["cc" "-E" "-I/usr/local/include" "-I/opt/homebrew/include" "test.c"] {os = "macos" & os-distribution = "homebrew"}
```
mseri, line 11: This will still fail on all nonstandard homebrew installations, but it is generally a problem of many packages

PR #25557, opened 2024-03-21
conf-gobject-introspection: add support for FreeBSD and Ubuntu derivatives

jmid:
Source: https://www.freshports.org/devel/gobject-introspection


jmid:
CI is green on FreeBSD! (and is red on Arch and Oracle which we make to attempt to patch).

mseri:
thanks

PR #25558, opened 2024-03-21
conf-goocanvas2: add support for Ubuntu derivatives

jmid:


jmid:
CI fails on Alpine, OpenSuse, Oracle, and macOS
- 27679bc corrects the OpenSuse package name according to https://opensuse.pkgs.org/tumbleweed/opensuse-oss-x86_64/goocanvas2-devel-2.0.4-2.10.x86_64.rpm.html
- Alpine and Oracle simply fail because they are not supported
- macOS fails because `gooconvas` on homebrew installs version 3: https://formulae.brew.sh/formula/goocanvas#default

None of these are related to the PR's patch, BTW

PR #25559, opened 2024-03-21
conf-guile: add support for Ubuntu derivatives

jmid:


jmid:
CI fails on
- OpenSuse 15 - which install guile-2.0 (not 3.0)
- Oracle which can't find the mentioned package
- FreeBSD uses the wrong package name. 

0254b9e fixes the latter based on https://www.freshports.org/lang/guile3

None of the above are failures caused by this PR.

PR #25560, opened 2024-03-21
conf-haveged: add support for Ubuntu derivatives

jmid:


jmid:
All green! :smile: (with an empty build test...)

PR #25561, opened 2024-03-21
conf-libargon2: add FreeBSD support

jmid:
Source: https://www.freshports.org/security/libargon2


jmid:
CI is green on FreeBSD! (it is red on Oracle, as usual)

PR #25562, opened 2024-03-21
conf-libdw: add FreeBSD support

jmid:
Source: https://www.freshports.org/devel/elfutils


jmid:
CI is green on FreeBSD!

- It is red on macOS (expected).
- there is also an unrelated revdep failure from orun.0.0.1

This should be good to merge

PR #25563, opened 2024-03-21
conf-libfuse: add support for FreeBSD and Ubuntu derivatives

jmid:
Source: https://www.freshports.org/sysutils/fusefs-libs/


mseri:
Thanks

PR #25564, opened 2024-03-21
conf-libgsasl: add support for Ubuntu derivatives

jmid:


jmid:
CI fails on Oracle, but is otherwise green! Good to merge.

PR #25565, opened 2024-03-21
conf-liblinear-tools: add support for FreeBSD and Ubuntu derivatives

jmid:
Source: https://www.freshports.org/science/liblinear/


jmid:
CI fails on Arch, Alpine, OpenSuse, and Oracle.

The PR does not attempt supporting either of these. As such, it should be good to go.

PR #25568, opened 2024-03-21
conf-libmaxminddb: add support for Ubuntu derivatives

jmid:


jmid:
CI fails on Arch and Oracle.

Both are unrelated to the PR, so it should be good to go.

PR #25569, opened 2024-03-21
conf-libmosquitto: strengthen build test + add support for Arch, Ubuntu, Suse, Fedora, FreeBSD

jmid:
Sources:
https://archlinux.org/packages/extra/x86_64/mosquitto/
https://opensuse.pkgs.org/tumbleweed/opensuse-oss-x86_64/mosquitto-devel-2.0.15-1.5.x86_64.rpm.html
https://packages.fedoraproject.org/pkgs/mosquitto/mosquitto-devel/fedora-39.html
https://www.freshports.org/net/mosquitto


jmid:
Wow, CI is green on all these in the first attempt!

It is red on Oracle (expected) - so this should be good to go!

PR #25573, opened 2024-03-24
conf-pkg-config: rename nixos package pkgconfig to pkg-config

timbertson:
`pkg-config` has been the correct name since 2018 ([reference](https://github.com/NixOS/nixpkgs/commit/9bb3fccb5b55326cb3c2c507464a8a28d44d1730)), but `pkgconfig` was an alias that worked for many years, so I never noticed it was the non-preferred option.

Now that `pkgconfig` doesn't work, I've retrospectively changed it in historical versions since it's not as if the package name actually changed, it was previously relying on an alias to the correct name.


mseri:
Thanks!

PR #25574, opened 2024-03-24
add patch fixing lablgtk3 #178

garrigue:
Recent versions of clang and gcc trigger a type error in a newly added function in ml_gtk.c (https://github.com/garrigue/lablgtk/issues/178).
This commit adds a patch fixing that.
The fix is already merged in lablgtk3, but making a new release will take a while.

### DIFF
--- a/packages/lablgtk3/lablgtk3.3.1.4/files/ml_gtk.c.diff
+++ b/packages/lablgtk3/lablgtk3.3.1.4/files/ml_gtk.c.diff
@@ -1,0 +1,13 @@
+diff --git a/src/ml_gtk.c b/src/ml_gtk.c
+index deb1b4c7a9..0dcc104e6f 100644
+--- a/src/ml_gtk.c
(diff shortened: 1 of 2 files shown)


garrigue:
If I do that, then I need to also modify lablgtk3-{sourceview3,gtkspell3,rsvg2,goocanvas2} since they all depend on using the same version of lablgtk3.
This fix is only needed if somebody could not install lablgtk3.3.1.4. If the install was successful, it is not needed, and the code is unchanged. So couldn't we keep it just as fix, rather than a new version.
I will make a new version as soon as I have time to do it, but it may take a while, and in the midtime people cannot use lablgtk3 and dependents.

mseri:
Ok, since this seems to be a (backward compatible) strict improvement and the patch is small enough, I think it is a reasonable compromise to merge it in the meantime

mseri:
Thanks

PR #25577, opened 2024-03-25
OCamlMPI version 1.06

xavierleroy:
- Removed unused code that breaks in OCaml 5.2
- Removed error handler that turns some MPI fatal errors into OCaml exceptions, it was raising an OCaml exception from within a blocking section, which is unsupported
- Fix issue with OCaml's memory-cleanup-at-exit mode
- Improved compatibility with OCaml 5.0 and up


mseri:
Thanks

Issue #25583, opened 2024-03-26
merlin packages incompatibilities


raphael-proust:
`merlin.4.14-414`  depends on `"merlin-lib" {= version}` and on `"dot-merlin-reader" {>= "4.9"}`

Currently the only available versions of `dot-merlin-reader>=4.9` are `4.14-502~preview` and `4.9`.
- The former requires `merlin-lib=version` (meaning `=4.14-502`) so it's incompatible with `merlin.4.14-414`.
- The latter requires `"merlin-lib" {>= "4.9" & < "4.14"}` so it's incompatible with `merlin.4.14-414`.

I don't think that `merlin.4.14-414` is installable at all.

Are we missing a release of `dot-merlin-reader.4.14-414`? Do we need to relax constraints on `dot-merlin-reader` requirement for `merlin-lib`?

Related: https://github.com/ocaml/opam-repository/pull/25275

PR #25584, opened 2024-03-26
relax bound on dot-merlin-reader -> merlin-lib constraint

raphael-proust:
attempting to fix https://github.com/ocaml/opam-repository/issues/25583


mseri:
Thanks

PR #25473, opened 2024-03-12
Specifying where to find ppl and gmp in the build in case of macOS with homebrew

rtetley:
On my setup, trying to install conf-ppl results on the following error:
```
[ERROR] The compilation of conf-ppl.1 failed at "sh -c cc test.c -lppl_c -lppl".

#=== ERROR while compiling conf-ppl.1 =========================================#
# context     2.1.5 | macos/arm64 | ocaml-base-compiler.4.13.1 | https://opam.ocaml.org#c55e2f61
# path        ~/.opam/4.13.1/.opam-switch/build/conf-ppl.1
# command     ~/.opam/opam-init/hooks/sandbox.sh build sh -c cc test.c -lppl_c -lppl
# exit-code   1
# env-file    ~/.opam/log/conf-ppl-63991-0708ed.env
# output-file ~/.opam/log/conf-ppl-63991-0708ed.out
### output ###
# test.c:1:10: fatal error: 'ppl_c.h' file not found
# #include <ppl_c.h>
#          ^~~~~~~~~
# 1 error generated.



<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+- The following actions failed
| - build conf-ppl 1
+-
- No changes have been performed
# Run eval $(opam env) to update the current shell environment
```

This is an attempt at addressing this issue.

PR #25592, opened 2024-03-27
Provides path to lib and include dirs for homebrew distribution.

clebreto:
Leaves the option of having a flexible location for homebrew installation.

It is a slightly different alternative to that of #25473.

### DIFF
--- a/packages/conf-ppl/conf-ppl.1/opam
+++ b/packages/conf-ppl/conf-ppl.1/opam
@@ -5,6 +5,7 @@ bug-reports: "http://bugseng.com/products/ppl/bugs"
 license: "GPL-1.0-or-later"
 build: [
   ["sh" "-c"
+	"cc test.c -I${HOMEBREW_PREFIX:-/opt/homebrew}/include -L${HOMEBREW_PREFIX:-/opt/homebrew}/lib -lppl_c -lppl" {os = "macos" & os-distribution = "homebrew"}
 	"cc test.c -lppl_c -lppl" {os != "freebsd" & os != "openbsd"}
 	"cc test.c -I/usr/local/include -L/usr/local/lib -lppl_c -lppl" {os = "freebsd" | os = "openbsd"}]
 ]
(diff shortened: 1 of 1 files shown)


clebreto:
CI fails on oraclelinux and alpine, these OSs have been added since the last successful pipeline ? 
The changes brought to the recipe should have no interference with these OSs.

I am very unfamiliar with these distributions, how to ensure that ppl is at least installed on the CI machines ?

samoht:
Thanks for your patch! The failure seems to indicate that we are missing entries in the `depexts` field for those distributions. Would you mind trying to add some of those? It's also fine to merge just as is if @rtetley or @jmid have an opinion vs. #25473

mseri:
`oraclelinux` does not provide libppl afaik and alpine has it on the edge repository, so I don't think we can improve it for the moment.

I'd wait to hear if there are other opinions, but this is good to merge also for me

jmid:
LGTM too! :+1: 

If we want to get it to zero red CI lights, adding something like the following should work (not completely sure about the `alpine-3.19` part though):
```diff
+x-ci-accept-failures: [
+  "oraclelinux-8" "oraclelinux-9" "alpine-3.19"
+]
```

packages/conf-ppl/conf-ppl.1/opam
```diff
@@ -5,6 +5,7 @@ bug-reports: "http://bugseng.com/products/ppl/bugs"
 license: "GPL-1.0-or-later"
 build: [
   ["sh" "-c"
+	"cc test.c -I${HOMEBREW_PREFIX:-/opt/homebrew}/include -L${HOMEBREW_PREFIX:-/opt/homebrew}/lib -lppl_c -lppl" {os = "macos" & os-distribution = "homebrew"}
 	"cc test.c -lppl_c -lppl" {os != "freebsd" & os != "openbsd"}
 	"cc test.c -I/usr/local/include -L/usr/local/lib -lppl_c -lppl" {os = "freebsd" | os = "openbsd"}]
 ]
```
mseri, line 11:
```suggestion
]
x-ci-accept-failures: [
  "oraclelinux-8" "oraclelinux-9" "alpine-3.19"
]
```

mseri:
Let's do it, I don't see the situation improving on those distros anyway

PR #25595, opened 2024-03-27
Set upper bound for tls compatibility.

toots:
The latest `ocaml-tls` release has an API change that breaks optional compilation with `liquidsoap`.


packages/liquidsoap-core/liquidsoap-core.2.2.0/opam
```diff
@@ -109,6 +109,7 @@ conflicts: [
   "taglib" {< "0.3.10"}
   "sdl-liquidsoap" {< "2"}
   "theora" {< "0.4.0"}
+  "tls" {>= "0.7.4"}
```
hannesm, line 112:
```suggestion
  "tls" {>= "0.17.4"}
```

toots: Fixed, thanks!

mseri:
Thanks

PR #25603, opened 2024-03-29
Mark older versions of `result` only compatible with <4.08

Leonidas-from-XIV:
We ran across this when attempting to release Dune 3.15.0~alpha1, where we removed our reference to the `result` package. Thus in the [lower version check](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/68b4763e3857d91953e14b698593ea06ced07650) it would use `result.1.0`. The code would then fail because of

```
# Error: This expression has type (string, exn) Result.result Lwt.t
#        but an expression was expected of type (string, exn) result Lwt.t
#        Type (string, exn) Result.result is not compatible with type
#          (string, exn) result 
```

Given the types should be equal on 4.08 (and later, the same failure keeps happening for the whole 4.x branch) something is fishy here.

Before https://github.com/janestreet/result/pull/9 on 4.08+ the types `Result.result` and `result` would not unify, thus causing compilation errors when code that was using both ended up interfacing. The fix has been part of `result.1.5`.

Given Dune doesn't depend on `result` anymore we can't constrain it to `result >= 1.5`, we'd need to add a `conflict` and that would solve the problem, however it seems weird to solve this in Dune in the first place. @emillon pointed out that a number of other packages already have a conflict for `result < 1.5` so it seems like this is an issue that packages keep running into the same solution: excluding older versions of result.

### DIFF
--- a/packages/result/result.1.0/opam
+++ b/packages/result/result.1.0/opam
@@ -12,7 +12,7 @@ description: """
 Projects that want to use the new result type defined in OCaml >= 4.03
 while staying compatible with older version of OCaml should use the
 Result module defined in this library."""
-depends: ["ocaml" {< "4.14"}]
+depends: ["ocaml" {< "4.08"}]
 url {
   src: "https://github.com/janestreet/result/archive/1.0.tar.gz"
   checksum: "md5=01391a66385ab1a43f90455dfb5c6843"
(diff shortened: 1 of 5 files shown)


Leonidas-from-XIV:
Ah, thanks; I wasn't aware that this had been discussed time and time again. It's a bit of a disappointing outcome, as this is an issue that keeps happening to multiple projects but I guess for time being we'll have to add the conflict in Dune.

avsm:
I believe this change will be ok to add in once 4.08 becomes the lower bound of testing in the opam repository, following the community discussions in https://github.com/ocaml/opam-repository/issues/23789

PR #25611, opened 2024-04-02
Relax dependency on merlin-lib

voodoos:
See comment https://github.com/ocaml/opam-repository/commit/dcb0fb6e629e707e29a45e65cf72603f3a7ea2d4#r140497176

This version of dot-merlin-reader should be co-installable with merlin-lib `4.14-501`.

In the future we will need to specifiy which variants (>= xxx-502) actually conflict with older dot-merlin-readers.


mseri:
Thanks

Issue #25591, opened 2024-03-27
Error installing conf-lapack in ALPINE


nlaykh:
Hello, I am trying to install csdp (which depends on Lapack) on Alpine and I get the following error:

```
$ opam install -v conf-lapack
<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+ /bin/sh "-exc" "cc $CFLAGS test.c ${LACAML_LIBS:--llapack}" (CWD=/root/.opam/4.12.1/.opam-switch/build/conf-lapack.1)
- + cc test.c -llapack
- /usr/lib/gcc/x86_64-alpine-linux-musl/13.2.1/../../../../x86_64-alpine-linux-musl/bin/ld: /tmp/ccDLIEFM.o: in function `main':
- test.c:(.text+0x2a): undefined reference to `dlamch_'
- collect2: error: ld returned 1 exit status
[ERROR] The compilation of conf-lapack.1 failed at "sh -exc cc $CFLAGS test.c ${LACAML_LIBS:--llapack}".

#=== ERROR while compiling conf-lapack.1 ======================================#
# context     2.1.5 | linux/x86_64 | ocaml-base-compiler.4.12.1 | https://opam.ocaml.org#aef07c8f
# path        ~/.opam/4.12.1/.opam-switch/build/conf-lapack.1
# command     /bin/sh -exc cc $CFLAGS test.c ${LACAML_LIBS:--llapack}
# exit-code   1
# env-file    ~/.opam/log/conf-lapack-225-4a4852.env
# output-file ~/.opam/log/conf-lapack-225-4a4852.out
### output ###
# + cc test.c -llapack
# /usr/lib/gcc/x86_64-alpine-linux-musl/13.2.1/../../../../x86_64-alpine-linux-musl/bin/ld: /tmp/ccDLIEFM.o: in function `main':
# test.c:(.text+0x2a): undefined reference to `dlamch_'
# collect2: error: ld returned 1 exit status



<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+- The following actions failed
| - build conf-lapack 1
+- 
- No changes have been performed
# Run eval $(opam env) to update the current shell environment
'opam install -v conf-lapack' failed.
```
Could you give me any hint to fix this?

erikmd:
Cc @mmottl FYI

mmottl:
[...]

PR #25664, opened 2024-04-12
Fixes for Alpine

nlaykh:
This pull request addresses issue #25591 with the conf-lapack opam package failing to compile on Alpine Linux.

It complained about a missing symbol (dlamch_). The problem arose because the lapack  package alone is not enough in Alpine. We also need the openblas package.

Hence the additional dependence.

Cc: @mmottl


mseri:
Thanks

PR #25667, opened 2024-04-12
Add upper bound to binsec's depopt on llvm

alan-j-hu:


raphael-proust:
This looks ok to me. The only reservation I have is that if this is only an issue on homebrew, then it'd be good to only have the constraint on homebrew. But I don't know that it'd be worth the added complexity anyway.

alan-j-hu:
@raphael-proust This is separate from the Homebrew issue. The reason why binsec doesn't compile with LLVM 16 is that the LLVM API changed. However, somehow the compile errors weren't caught when submitting LLVM 16 to opam-repository.

raphael-proust:
@alan-j-hu thanks for the contribution and for the explanation <3

PR #25670, opened 2024-04-13
[new release] ocaml-protoc-plugin (6.0.0)

andersfugmann:

- Project page: <a href="https://github.com/andersfugmann/ocaml-protoc-plugin">https://github.com/andersfugmann/ocaml-protoc-plugin</a>
- Documentation: <a href="https://andersfugmann.github.io/ocaml-protoc-plugin/">https://andersfugmann.github.io/ocaml-protoc-plugin/</a>

### New features
- Implement json serialization and deserialization (andersfugmann/ocaml-protoc-plugin#5)
- Support special json mapping for google types (andersfugmann/ocaml-protoc-plugin#9)
- Add deprecation annotations for deprecated fields, services etc (andersfugmann/ocaml-protoc-plugin#8)
- Add option to prefix generated files with their package name
- Copy documentation from proto files into generated ocaml bindings

### Bug fixes
- Fix file output name if files contains a '-'
- Resolve bug for Request/Response module aliases leading to generating uncompilable code. (andersfugmann/ocaml-protoc-plugin#21)
- Fix codegen bug for messages with out fields and setting singleton_records = true (andersfugmann/ocaml-protoc-plugin#20)
- In Services, the package field is now correctly set to None if the service if not defined in a package scope (andersfugmann/ocaml-protoc-plugin#24)

### Misc changes
- Unify serialization and deserialization spec and optimize oneof handling
- Simplify types used in code generation to improve readaility
- *Replace `val name': unit -> string` with `val name: unit -> string` which will only return the full protobuf name
- Optimize merge functions by applying eager evaluation
- Change signature of `to_proto'` to return unit and not a writer

(`*` indicates breaking change)

### Notes
  `Message.name': unit -> string` has been renamed to `Message.name:
  unit -> string`, and is now contains the fully qualified protobuf
  message name. Before the name was the ocaml module name of the
  message.

  `Service.Message` signature has been deprecated and replaced with
  `Spec.Message` signature. `Service.Message` is now an alias for
  `Spec.Message` and will be removed in future releases.

### DIFF
--- a/packages/ocaml-protoc-plugin/ocaml-protoc-plugin.6.0.0/opam
+++ b/packages/ocaml-protoc-plugin/ocaml-protoc-plugin.6.0.0/opam
@@ -26,7 +26,7 @@ depends: [
   "conf-pkg-config" {build}
   "dune-configurator" {with-test}
   "yojson" {with-test}
-  "base64" {>= 3.0.0}
+  "base64" {>= "3.0.0"}
   "ptime"
 ]
 
--- a/packages/ocaml-protoc-plugin/ocaml-protoc-plugin.6.0.0/opam
+++ b/packages/ocaml-protoc-plugin/ocaml-protoc-plugin.6.0.0/opam
@@ -26,7 +26,7 @@ depends: [
   "conf-pkg-config" {build}
   "dune-configurator" {with-test}
   "yojson" {with-test}
-  "base64" {>= "3.0.0"}
+  "base64" {>= "3.1.0"}
   "ptime"
 ]
 
--- a/packages/ocaml-protoc-plugin/ocaml-protoc-plugin.6.0.0/opam
+++ b/packages/ocaml-protoc-plugin/ocaml-protoc-plugin.6.0.0/opam
@@ -26,7 +26,7 @@ depends: [
   "conf-pkg-config" {build}
   "dune-configurator" {with-test}
   "yojson" {with-test}
-  "base64"
+  "base64" {>= 3.0.0}
   "ptime"
 ]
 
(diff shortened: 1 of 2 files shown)


raphael-proust:
The CI fails in a strange way and so I opened an issue in opam-repo-ci for it: https://github.com/ocurrent/opam-repo-ci/issues/291

talex5:
I don't know why the error isn't rendering properly, but running `opam lint` manually shows:
```
    error  2: File format error at line 29, column 16: '.' is not a valid token
```
(in `"base64" {>= 3.0.0}`)

packages/ocaml-protoc-plugin/ocaml-protoc-plugin.6.0.0/opam
```diff
@@ -0,0 +1,59 @@
...
+  "ocaml" {>= "4.08.0"}
+  "ppx_expect"
+  "ppx_inline_test"
+  "ppx_deriving" {with-test}
+  "bisect_ppx" {with-test}
+  "odoc" {with-doc}
+  "conf-pkg-config" {build}
+  "dune-configurator" {with-test}
+  "yojson" {with-test}
+  "base64" {>= 3.0.0}
```
avsm, line 29:
```suggestion
  "base64" {>= "3.0.0"}
```

andersfugmann: thanks!

andersfugmann:
There are multiple errors at play:
* The test build uses a native stub, but its not using the correct flags which may explain some of the compilation errors
* Segv and Abrt failures I expect are caused by a bug in the google library (which has been fixed in newer versions) I will rewrite the tests to circumvent this. 
* Some tests does not match on expected output. This is also caused by the bug mentioned above.
* MacOS failures may be solved by lack of default cflags. If not, I will add macos to `x-ci-accept-failures`

I will prepare a 6.1.0 release, as I've fixed other bugs found during testing and close this PR.

PR #25675, opened 2024-04-15
[new release] ppx_deriving (6.0.0)

NathanReb:
Type-driven code generation for OCaml

- Project page: <a href="https://github.com/ocaml-ppx/ppx_deriving">https://github.com/ocaml-ppx/ppx_deriving</a>
- Documentation: <a href="https://ocaml-ppx.github.io/ppx_deriving/">https://ocaml-ppx.github.io/ppx_deriving/</a>

##### CHANGES:

* Fix a bug in `[@@deriving make]` that caused errors when it was used on a set
  of type declarations containing at least one non record type.
  ocaml-ppx/ppx_deriving#281
  (@NathanReb)

* Embed errors instead of raising exceptions when generating code
  with `ppx_deriving.make`
  ocaml-ppx/ppx_deriving#281
  (@NathanReb)

* Remove `[%derive.iter ...]`, `[%derive.map ...]` and `[%derive.fold ...]`
  extensions
  ocaml-ppx/ppx_deriving#278
  (Simmo Saan)

* Port standard plugins to ppxlib registration and attributes
  ocaml-ppx/ppx_deriving#263
  (Simmo Saan)

* Optimize forwarding in eq and ord plugins
  ocaml-ppx/ppx_deriving#252
  (Simmo Saan)

* Delegate quoter to ppxlib
  ocaml-ppx/ppx_deriving#263
  (Simmo Saan)

* Introduce `Ppx_deriving_runtime.Stdlib` with OCaml >= 4.07.
  This module already exists in OCaml < 4.07 but was missing otherwise.
  ocaml-ppx/ppx_deriving#258
  (Kate Deplaix)

### DIFF
--- a/packages/current/current.0.2/opam
+++ b/packages/current/current.0.2/opam
@@ -34,6 +34,7 @@ depends: [
   "prometheus"
   "dune" {>= "1.9"}
   "re" {>= "1.9.0"}
+  "result"
   "lwt-dllist"
   "alcotest-lwt" {>= "1.0.1" & < "1.2.0" & with-test}
 ]
--- a/packages/current/current.0.3/opam
+++ b/packages/current/current.0.3/opam
@@ -34,6 +34,7 @@ depends: [
   "prometheus"
   "dune" {>= "2.0"}
   "re" {>= "1.9.0"}
+  "result"
   "lwt-dllist"
   "alcotest-lwt" {>= "1.0.1" & < "1.2.0" & with-test}
 ]
--- a/packages/ppx_deriving/ppx_deriving.6.0.0/opam
+++ b/packages/ppx_deriving/ppx_deriving.6.0.0/opam
@@ -1,0 +1,39 @@
+opam-version: "2.0"
+maintainer: "whitequark <whitequark@whitequark.org>"
+authors: [ "whitequark <whitequark@whitequark.org>" ]
+license: "MIT"
+homepage: "https://github.com/ocaml-ppx/ppx_deriving"
+doc: "https://ocaml-ppx.github.io/ppx_deriving/"
+bug-reports: "https://github.com/ocaml-ppx/ppx_deriving/issues"
+dev-repo: "git+https://github.com/ocaml-ppx/ppx_deriving.git"
+tags: [ "syntax" ]
+build: [
+  ["dune" "subst"] {dev}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+  ["dune" "build" "@doc" "-p" name "-j" jobs] {with-doc}
+]
+depends: [
+  "ocaml" {>= "4.05.0"}
+  "dune" {>= "1.6.3"}
+  "cppo" {>= "1.1.0" & build}
+  "ocamlfind"
+  "ppx_derivers"
+  "ppxlib" {>= "0.32.0"}
+  "ounit2" {with-test}
+]
+synopsis: "Type-driven code generation for OCaml"
+description: """
+ppx_deriving provides common infrastructure for generating
+code based on type definitions, and a set of useful plugins
+for common tasks.
+"""
+url {
+  src:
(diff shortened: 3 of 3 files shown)


sim642:
Most CI errors seem to be related to https://github.com/ocaml-ppx/ppx_deriving/pull/279 (which isn't in the changelog):
```
#=== ERROR while compiling ppx_deriving_yojson.3.7.0 ==========================#
# context              2.2.0~beta3~dev | linux/x86_64 | ocaml-base-compiler.4.14.2 | file:///home/opam/opam-repository
# path                 ~/.opam/4.14/.opam-switch/build/ppx_deriving_yojson.3.7.0
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p ppx_deriving_yojson -j 39
# exit-code            1
# env-file             ~/.opam/log/ppx_deriving_yojson-7-9f30e6.env
# output-file          ~/.opam/log/ppx_deriving_yojson-7-9f30e6.out
### output ###
# (cd _build/default && /home/opam/.opam/4.14/bin/ocamlc.opt -w -40 -g -bin-annot -I src/.ppx_deriving_yojson_runtime.objs/byte -I /home/opam/.opam/4.14/lib/ppx_deriving/runtime -I /home/opam/.opam/4.14/lib/result -intf-suffix .ml -no-alias-deps -o src/.ppx_deriving_yojson_runtime.objs/byte/ppx_deriving_yojson_runtime.cmo -c -impl src/ppx_deriving_yojson_runtime.ml)
# File "src/ppx_deriving_yojson_runtime.ml", line 20, characters 32-45:
# 20 | type 'a error_or = ('a, string) Result.result
#                                      ^^^^^^^^^^^^^
# Error: Unbound type constructor Result.result
# (cd _build/default && /home/opam/.opam/4.14/bin/ocamlopt.opt -w -40 -g -I src/.ppx_deriving_yojson_runtime.objs/byte -I src/.ppx_deriving_yojson_runtime.objs/native -I /home/opam/.opam/4.14/lib/ppx_deriving/runtime -I /home/opam/.opam/4.14/lib/result -intf-suffix .ml -no-alias-deps -o src/.ppx_deriving_yojson_runtime.objs/native/ppx_deriving_yojson_runtime.cmx -c -impl src/ppx_deriving_yojson_runtime.ml)
# File "src/ppx_deriving_yojson_runtime.ml", line 20, characters 32-45:
# 20 | type 'a error_or = ('a, string) Result.result
#                                      ^^^^^^^^^^^^^
# Error: Unbound type constructor Result.result
```

NathanReb:
It looks to me like this errors are side effects of this change indeed but come from the fact that those package use `Result` without explicitly depending on it.

The fix is likely to add that dependency to `ppx_deriving_yojson`, I'll take care of that!

NathanReb:
Hmm ok my bad, this is worse than I thought, the dependency is not declared in the dune file for `ppx_deriving_yojson` so patching the opam file won't be enough here.

NathanReb:
Okay this is more complicated than I thought. While removing result we seem to have messed up `Ppx_deriving_runtime` which for some reason re-includes the standard library in a weird way.

Since we removed an overriden `Result` module there that was defining the type equality `('a, 'err) t = ('a, 'err) result` (which is already defined by the result compat package btw), including `Ppx_deriving_runtime` as is done in `ppx_deriving_yojson` makes the standard Result module shadow the compat one and therefore causes this error because the standard `Result` module does not define a `result` type, only `t`.

NathanReb:
We could either add an upper bound on ppx_deriving in ppx_deriving_yojson or simply fix this small part of the change in ppx_deriving.

Given both `ppx_deriving.api` and `ppx_deriving.runtime` are to be deprecated, I'd be slightly in favor of fixing this in ppx_deriving so it has one last clean version before we deprecate the deriver creation API.

NathanReb:
I'm closing once again, will release 6.0.1 with the fix to `Ppx_deriving_runtime`.

PR #25686, opened 2024-04-16
conf-emacs and conf-zlib: update depext for Fedora 40

maroneze:
I just upgraded to Fedora 40 (still in beta), and some package names seem to have changed, causing a very annoying behavior: whenever I try to install any package, I get messages such as:

    [WARNING] Opam packages conf-emacs.1 and conf-zlib.1
    depend on the following system packages that are no longer installed: emacs-nox zlib-devel

This leads to lots of reinstallation and recompilation; for instance, simply trying to install Merlin results in:

```
  ∗ install   merlin-lib        4.14-501 [required by merlin]
  ↻ recompile conf-zlib         1        [upstream or system changes]
  ∗ install   dot-merlin-reader 4.9      [required by merlin]
  ↻ recompile camlzip           1.11     [uses conf-zlib]
  ∗ install   merlin            4.14-501
  ↻ recompile why3              1.7.1    [uses camlzip]
  ↻ recompile tuareg            3.0.1    [uses merlin]
```

opam tries to recompile `conf-zlib`, which succeeds, since it finds the actual new package name:

    Package zlib-ng-compat-devel-2.1.6-2.fc40.x86_64 is already installed.

And it recompiles every package depending on conf-zlib.

So, in the end, everything works, but it stills wastes lots of time recompiling things.

I assume the issue is due to the package names in the `conf-*` opam files, so I tried updating them, but to be honest I'm not entirely sure this will solve it.

### DIFF
--- a/packages/conf-emacs/conf-emacs.1/opam
+++ b/packages/conf-emacs/conf-emacs.1/opam
@@ -9,7 +9,8 @@ depexts: [
   ["emacs"] {os-distribution = "ubuntu"}
   ["emacs-nox"] {os-distribution = "centos"}
   ["emacs-nox"] {os-distribution = "rhel"}
-  ["emacs-nox"] {os-distribution = "fedora"}
+  ["emacs-nox"] {os-distribution = "fedora" & os-version < "40" }
+  ["emacs-nw"] {os-distribution = "fedora" & os-version >= "40" }
   ["emacs-nox"] {os-distribution = "alpine"}
   ["emacs-nox"] {os-family = "suse" | os-family = "opensuse"}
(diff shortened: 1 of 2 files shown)


samoht:
Did you try to `pin` those packages on your system and checked that they are working as expected?

samoht:
(we should also probably add support for Fedora 40 in opam-repo-ci /cc @tmattio)

maroneze:
Thanks, I forgot that was the command to try them.

I just did it, and indeed the messages disappeared, so it works for me.

samoht:
Many thanks! So let's merge this now while the team adds Fedora 40 to the CI matrix.

PR #25687, opened 2024-04-17
new file:   packages/bst/bst.7.1.0/opam

UnixJunkie:
New minor version of my bst library for bisector trees.

There is a new query functionality: any_neighbor.
Should be way faster than actually running a range query or looking at the distance of the nearest neighbor.


samoht:
Thanks!

PR #25694, opened 2024-04-18
Why3 1.7.2

silene:


silene:
The only failure is the `lint` job, due to the closeness between the names `why` and `why3`.

mseri:
Thanks

PR #25715, opened 2024-04-23
update ocaml-solo5 synopsis, as suggest by @shym 

hannesm:
https://github.com/mirage/ocaml-solo5/pull/133


hannesm:
CI issues are not relevant -- indeed the ocaml-solo5 opam package has a dune-project file (for testing the unikernel in that repository) -- but the compilation (as seen in the dependencies and in the build instructions) do not use dune.

PR #25720, opened 2024-04-24
[new-release] ppx_deriving_yojson 3.8.0

NathanReb:
## Changes

  * Port deriver to ppxlib (#149)  @sim642

### DIFF
--- a/packages/pa_ppx/pa_ppx.0.10/opam
+++ b/packages/pa_ppx/pa_ppx.0.10/opam
@@ -50,7 +50,7 @@ depends: [
   "sexplib" { with-test & >= "v0.14.0" }
   "ppx_import" { with-test & >= "1.7.1" }
   "ppx_deriving" { with-test }
-  "ppx_deriving_yojson" { with-test & >= "3.5.2" }
+  "ppx_deriving_yojson" { with-test & >= "3.5.2" & < "3.8.0" }
   "ppx_here" { with-test & >= "v0.13.0" }
   "ppx_sexp_conv" { with-test & >= "v0.13.0" }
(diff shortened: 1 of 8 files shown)


NathanReb:
I'm going through the errors here:
- [ ] `goblint` & co: This is a legit error here, similar to what happened with `ppx_deriving`, we broke the `Ppx_deriving_yojson_runtime` API when dropping the `result` compat package dependency. I submitted a fix in ocaml-ppx/ppx_deriving_yojson#157
- [ ] `morbig` & co: Same as above, same fix.
- [ ] satyrographos: Same
- [ ] safemoney: they were trying to use `ppx_deriving_yojson` on exceptions which was silently failing so far. The port to ppxlib turned that silent error into an actual one which is causing the build failure. I'll update the upper bound in the relevant opam files and already reported this issue upstream: gborough/safemoney#5
- [ ] pa_ppx: The test suite fails, I haven't quite put my finger on the exact reason yet but it does seem to be a legitimate error so I'll look into it.

NathanReb:
I've looked into `pa_ppx`'s test failures both here and with `ppx_deriving`. It seems the issue comes from how the rewriters are applied in the test suite. I didn't quite get how they were running ppx-es there but it seems the input name is not properly set for ppxlib.

I'll simply add an upper bound here as if a fix is needed, it's either on `pa_ppx` or `ppxlib`'s side.

NathanReb:
`morbig` still does not build but this time it seems to be unrelated to this release! I'll undraft this!

mseri:
pa_ppx_q_ast requires a lower bound on pa_ppx, from which it takes pcre2 transitively. I will send a fix separately. Thanks

PR #25727, opened 2024-04-24
owi.0.2

zapashcanon:
Hi,

This is a new release of [Owi](https://github.com/ocamlpro/owi). Here's the changelog:

## 0.2 - 2024-04-24

- use a subcommand system for the `owi` binary
- add `owi c`, `owi fmt`, `owi opt`, `owi sym`, `owi validate` and `owi wasm2wat` subcommands
- add a fuzzer
- add a profiling mode
- make `spectec` print stuff for real

### DIFF
--- a/packages/owi/owi.0.2/opam
+++ b/packages/owi/owi.0.2/opam
@@ -1,0 +1,77 @@
+opam-version: "2.0"
+synopsis:
+  "OCaml toolchain to work with WebAssembly, including an interpreter"
+description:
+  "owi is an OCaml toolchain to work with WebAssembly. It provides an interpreter as an executable and a library."
+maintainer: ["Léo Andrès <contact@ndrs.fr>"]
+authors: [
+  "Léo Andrès <contact@ndrs.fr>"
+  "Pierre Chambart <pierre.chambart@ocamlpro.com>"
+  "Filipe Marques <filipe.s.marques@tecnico.ulisboa.pt>"
+  "Eric Patrizio <epatrizio@mpns.fr>"
+  "Arthur Carcano <arthur.carcano@ocamlpro.com"
+]
+license: "AGPL-3.0-or-later"
+tags: ["owi" "ocaml" "webassembly" "wasm" "interpreter" "compiler"]
+homepage: "https://github.com/ocamlpro/owi"
+bug-reports: "https://github.com/ocamlpro/owi/issues"
+depends: [
+  "dune" {>= "3.0"}
+  "ocaml" {>= "5.1"}
+  "integers" {>= "0.5.1"}
+  "cmdliner"
+  "conf-clang"
+  "sedlex"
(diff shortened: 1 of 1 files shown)


zapashcanon:
I fixed most of CI errors by adding some missing packages. The remaining are due to too old distributions not providing some packages or too old versions of such packages.

There is one job failing because of a segfault and this one is due to a bug in Z3/the Z3 bindings. I see it from times to times when asking the solver for a model. It can be ignored.

packages/owi/owi.0.2/opam
```diff
@@ -0,0 +1,82 @@
...
+bug-reports: "https://github.com/ocamlpro/owi/issues"
+depends: [
+  "dune" {>= "3.0"}
+  "ocaml" {>= "5.1"}
+  "integers" {>= "0.5.1"}
+  "cmdliner"
+  "conf-clang"
+  "conf-python-3-dev"
+  "sedlex"
+  "menhir" {build & >= "20220210"}
```
avsm, line 26: Are you sure you want this to be a `build` dependency? The generated code from Menhir can diverge and cause the build to be inconsistent on an upgrade.

packages/owi/owi.0.2/opam
```diff
@@ -0,0 +1,82 @@
...
+available: (arch = "x86_32" | arch = "x86_64" | arch = "arm64") & os != "win32"
+depexts: [
+  [ "llvm" "lld" "wabt"]  {os-family = "debian"}
+  [ "llvm" "wabt"]  {os-family = "homebrew"}
+  [ "llvm" "lld" "wabt" ] {os-family = "arch"}
+  [ "llvm" "lld" "wabt" ] {os-family = "fedora"}
+  [ "llvm" "lld" ] {os-family = "opensuse"}
+  [ "llvm" "lld" "wabt" ] {os-family = "bsd"}
+]
+pin-depends: [
```
avsm, line 73: This still has a pin-depends ... that needs to be removed before it can go into mainline opam-repo.

packages/owi/owi.0.2/opam
zapashcanon: Yes, it already was a build dependency on previous release and we never had an issue so it should be just fine.

packages/owi/owi.0.2/opam
zapashcanon: Actually it is only a dev dependency and is not part of the automatic testsuite, the fuzzer has to be run manually so I'll remove it. :)

packages/owi/owi.0.2/opam
avsm: I took a quick look at why this works currently, and it's only because `menhir` has a strong dependency on `menhirLib` which is itself not a build dependency.

avsm:

zapashcanon:
Thanks!

PR #25729, opened 2024-04-26
conf-librsvg2: add support for Ubuntu family + Oracle

jmid:


jmid:
CI fails only for Oracle.
9fcd95d adds support for it, since it is apparently available on the App Stream repo enabled on the CI: https://pkgs.org/download/librsvg2-devel

jmid:
CI passes on Oracle now.

The opam linting failure `External dependencies should not contain spaces nor empty string: "librsvg +quartz")`
is related to the existing macOS depexts which this PR does not touch:
```
  ["librsvg +quartz" "libxml2"] {os = "macos" & os-distribution = "macports"}
```

jmid:
This should be good to go.

PR #25730, opened 2024-04-26
conf-libsvm: add support for Ubuntu family and FreeBSD

jmid:
FreeBSD package: https://www.freshports.org/science/libsvm/

### DIFF
--- a/packages/conf-libsvm/conf-libsvm.3/files/test.c
+++ b/packages/conf-libsvm/conf-libsvm.3/files/test.c
@@ -1,5 +1,6 @@
 #include <stdio.h>
-#ifdef DARWIN
+#include <sys/param.h>
+#if defined DARWIN || defined __FreeBSD__
 #include <svm.h>
 #else
 #include <libsvm/svm.h>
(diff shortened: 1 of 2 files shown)


jmid:
CI fails on Alpine, OpenSuse, Oracle, FreeBSD, and macOS
The first 3 are natural, as there is no current attempt to support them.
FreeBSD and macOS fails to find the svm.h header.

jmid:
- fea0fff should add OpenSuse support whereas
- d9fd138 marks Alpine and Oracle as acceptable CI failures
- 4f5c33c and 1a804a5 add include paths on FreeBSD and macOS

jmid:
CI was failing the file hash check. b9cf173 updates the md5sum accordingly.

mseri:
Seems to be still failing on macos ARM and freebsd

jmid:
Thanks.
Despite the red lights, there's progress: now both FreeBSD and macOS found the header and are instead complaining about not finding the library. 59ab3c9 therefore sets a library path.

jmid:
This is now all green, modulo opam lint warnings about package name clashes.
It should be good to go.

mseri:
Thanks

PR #25731, opened 2024-04-26
conf-libsvm-tools: add support for Ubuntu family and FreeBSD

jmid:
On FreeBSD this requires the same package as `conf-libsvm` in #25730: https://www.freshports.org/science/libsvm/


jmid:
CI fails on Alpine, Fedora, and Oracle.
- cabda84 should fix the package name on Fedora
- 744af9d accepts CI failures on Alpine and Oracle where I believe the package is not available

mseri:
Thanks

PR #25736, opened 2024-04-26
conf-libxrandr: add support for Ubuntu family and macOS homebrew

jmid:


jmid:
CI fails on macOS.

0bc570f should fix it based on https://formulae.brew.sh/formula/libxrandr#default

mseri:
Thanks

PR #25737, opened 2024-04-26
conf-lz4: add FreeBSD support

jmid:
Package info source: https://www.freshports.org/archivers/liblz4/


jmid:
This passes except for an opam lint package name collision warning.
It should be good to go.

PR #25745, opened 2024-04-26
conf-postgresql: fix package name on FreeBSD

jmid:
I spotted in https://freebsd.check.ci.dev/ that the FreeBSD package name (version, really) was off.

According to https://www.freshports.org/databases/postgresql15-client/
`postgresql15-client`
- contains `bin/pg_config`
- has been the default version since 2023-09-08

### DIFF
--- a/packages/conf-postgresql/conf-postgresql.1/opam
+++ b/packages/conf-postgresql/conf-postgresql.1/opam
@@ -4,7 +4,10 @@ authors: "Markus Mottl <markus.mottl@gmail.com>"
 homepage: "https://www.postgresql.org"
 dev-repo: "git+https://github.com/mmottl/postgresql-ocaml.git"
 license: "PD"
-build: [["pg_config"]]
+build: [
+  ["pg_config"] {os != "macos" | os-distribution != "homebrew"}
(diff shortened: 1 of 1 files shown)


packages/conf-postgresql/conf-postgresql.1/opam
```diff
@@ -4,7 +4,10 @@ authors: "Markus Mottl <markus.mottl@gmail.com>"
 homepage: "https://www.postgresql.org"
 dev-repo: "git+https://github.com/mmottl/postgresql-ocaml.git"
 license: "PD"
-build: [["pg_config"]]
+build: [
+  ["pg_config"] {os != "macos" | os-distribution != "homebrew"}
+  ["$(brew --prefix postgresql@15)/bin/pg_config"] {os = "macos" & os-distribution = "homebrew"}
```
mseri, line 9: I think this does not work, you need to invoke it behind sh:
```suggestion
  ["sh" "-c" "\"$(brew --prefix postgresql@15)\"/bin/pg_config"] {os = "macos" & os-distribution = "homebrew"}
```
similarly on the other failing PRs

mseri:
Thanks

PR #25747, opened 2024-04-26
conf-ghostscript: fix package name on FreeBSD

jmid:
https://freebsd.check.ci.dev/ is showing a failure
```
[ERROR] System package install failed with exit code 1 at command:
            sudo pkg install -y ghostscript
[ERROR] These packages are still missing: ghostscript
```
https://freebsd.check.ci.dev/log/1713371636-f089d8564db935c703bea70a4b3d8335ffa1bef3/4.14.1/internal-failure/conf-ghostscript.1

Use https://www.freshports.org/print/ghostscript10/ instead

### DIFF
--- a/packages/conf-ghostscript/conf-ghostscript.1/opam
+++ b/packages/conf-ghostscript/conf-ghostscript.1/opam
@@ -15,7 +15,8 @@ depexts: [
   ["ghostscript"] {os-family = "alpine"}
   ["ghostscript"] {os-family = "suse" | os-family = "opensuse"}
   ["ghostscript"] {os-family = "arch"}
-  ["ghostscript"] {os = "freebsd"}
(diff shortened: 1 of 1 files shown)


jmid:
CI is now green on FreeBSD too! :tada:

There is a revdep failure - `advi.2.0.0` succeeds on ocaml.4.14 but fails on ocaml.5.1 with:
```
# File "src/main.ml", line 175, characters 0-14:
# 175 | Printexc.catch (Misc.handle_fatal_error main) ();;
#       ^^^^^^^^^^^^^^
# Error (alert deprecated): Stdlib.Printexc.catch
# This function is no longer needed.
```

That failure is however unrelated to the fixed FreeBSD support.
As such, this should be good to go.

PR #25750, opened 2024-04-26
conf-snappy: add include dir on FreeBSD

jmid:
On FreeBSD this fails with https://freebsd.check.ci.dev/log/1713371636-f089d8564db935c703bea70a4b3d8335ffa1bef3/4.14.1/bad/conf-snappy.1
```
# test.cpp:1:10: fatal error: 'snappy.h' file not found
# #include <snappy.h>
#          ^~~~~~~~~~
# 1 error generated.
```

The PR therefore adds a (standard) FreeBSD include directory.

### DIFF
--- a/packages/conf-snappy/conf-snappy.1/opam
+++ b/packages/conf-snappy/conf-snappy.1/opam
@@ -16,6 +16,10 @@ depexts: [
   ["snappy"] {os-family = "suse" | os-family = "opensuse"}
   ["snappy"] {os = "freebsd"}
 ]
+x-ci-accept-failures: [
+  "oraclelinux-8"
+  "oraclelinux-9"
+]
 synopsis: "Virtual package relying on snappy"
 description:
   "This package can only install if the snappy library is installed on the system."
--- a/packages/conf-snappy/conf-snappy.1/opam
+++ b/packages/conf-snappy/conf-snappy.1/opam
@@ -3,7 +3,8 @@ maintainer: "gabriel.scherer@inria.fr"
 homepage: "https://google.github.io/snappy/"
 bug-reports: "https://github.com/ocaml/opam-repository/issues"
 build: [
-  ["c++" "test.cpp" "-lsnappy" "-std=c++11"]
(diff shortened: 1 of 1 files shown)


packages/conf-snappy/conf-snappy.1/opam
```diff
@@ -3,7 +3,8 @@ maintainer: "gabriel.scherer@inria.fr"
 homepage: "https://google.github.io/snappy/"
 bug-reports: "https://github.com/ocaml/opam-repository/issues"
 build: [
-  ["c++" "test.cpp" "-lsnappy" "-std=c++11"]
+  ["c++" "test.cpp" "-lsnappy" "-std=c++11"] {os != "freebsd"}
+  ["c++" "test.cpp" "-I/usr/local/include" "-lsnappy" "-std=c++11"] {os = "freebsd"}
```
mseri, line 7: Does not seem to work. In this case, I think you could also do
```suggestion
  ["c++" "test.cpp"
   "-I/usr/local/include" {os = "freebsd"}
   "-lsnappy" "-std=c++11"] 
```

jmid:
Progress - now the header is found on FreeBSD - but the library isn't:
```
### output ###
# ld: error: unable to find library -lsnappy
# c++: error: linker command failed with exit code 1 (use -v to see invocation)
```

- 6fd6708 sets a FreeBSD library path
- acc4d6b silences Oracle CI errors
- 092e762 adds an include and library path since the header wasn't being found on macOS ARM64 either:
```
### output ###
# test.cpp:1:10: fatal error: 'snappy.h' file not found
# #include <snappy.h>
#          ^~~~~~~~~~
# 1 error generated.
```

mseri:
Now only macos fails. I am inclined to merge it as is, what do you think?

jmid:
I would like to fix the remaining macOS ARM64 failure first. :pray:
I suspect this is just missing the `sh -c` invocation you've suggested in the other PRs? :thinking:

PR #25757, opened 2024-04-28
secondary packages to compile the JS flambda2 compiler

avsm:
This (draft) PR allows the flambda2 compiler to be installed with:

```
opam switch create flambda2 5.1.1+flambda2
```

There are some bleeding-edge limitations with the compiler, so its also helpful to add:

```
opam repo add duni https://github.com/dune-universe/opam-overlays.git
```

to get more dune overlay packages (as ocamlbuild compilation fails).

### DIFF
--- a/packages/ocaml-variants/ocaml-variants.5.1.1+flambda2/opam
+++ b/packages/ocaml-variants/ocaml-variants.5.1.1+flambda2/opam
@@ -1,0 +1,36 @@
+opam-version: "2.0"
+version: "5.1.1+flambda2"
+synopsis: "Jane Street compiler variant with Flambda2 backend"
+depends: [
+  "ocaml" {= "5.1.1" & post}
+  "base-unix" {post}
+  "base-bigarray" {post}
+  "base-threads" {post}
+  "dune-secondary" {="3.8.1"}
+  "menhir-secondary"
+  "conf-autoconf" {build}
+]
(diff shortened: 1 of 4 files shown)


avsm:
I see draft PRs aren't filtered by the CI...

packages/ocaml-variants/ocaml-variants.5.1.1+flambda2/opam
```diff
@@ -0,0 +1,36 @@
...
+  "ocaml" {= "5.1.1" & post}
+  "base-unix" {post}
+  "base-bigarray" {post}
+  "base-threads" {post}
+  "dune-secondary" {="3.8.1"}
+  "menhir-secondary"
+  "conf-autoconf" {build}
+]
+conflict-class: "ocaml-core-compiler"
+flags: compiler
```
dra27, line 14: Just to persist a thought: I initially thought, "should we put `avoid-version` here to stop flambda2 accidentally being activated"? However, the fact of `flambda2` is less than `options` means that it should not be accidentally activated, so I think it's just fine as it is!

dra27 (commented):
I've not reviewed in detail, but a thought at this stage - I think it's probably worth adding `base-flambda2` even while there's only the one version? (as we did for `base-domains` and `base-effects` before multicore was merged)

avsm:
I'm closing this for now to avoid unnecessary CI runs.

PR #25759, opened 2024-04-29
[new release] libevent (0.9.0)

Khady:
This adds support for ocaml 5

### DIFF
--- a/packages/libevent/libevent.0.9.0/opam
+++ b/packages/libevent/libevent.0.9.0/opam
@@ -1,0 +1,43 @@
+opam-version: "2.0"
+maintainer: "ygrek@autistici.org"
+license: "BSD-3-Clause"
+authors: [
+  "Maas-Maarten Zeeman <mmzeeman@xs4all.nl>"
+  "ygrek <ygrek@autistici.org>"
+]
+homepage: "https://github.com/ygrek/ocaml-libevent"
+dev-repo: "git+https://github.com/ygrek/ocaml-libevent.git"
+bug-reports: "https://github.com/ygrek/ocaml-libevent/issues"
+tags: ["org:ygrek"]
+build: [
+  [make "depend"]
+  [make]
+  [make "test"] {with-test & os != "macos"}
+  [make "doc"] {with-doc}
+]
+install: [
+  [make "install"]
+]
+depends: [
+  "ocaml" {>= "4.02.0"}
+  "ocamlfind" {build}
+  "base-unix"
+  "base-bytes" {with-test}
+  "ounit" {with-test}
+  "conf-libevent" {build}
+]
+synopsis: "OCaml wrapper for the libevent API"
+description: """
+libevent provides a mechanism to execute a callback function
+when a specific event occurs on a file descriptor or
+after a timeout has been reached. Furthermore, libevent also
+supports callbacks due to signals or regular timeouts."""
+url {
+  src: "https://ygrek.org/p/release/ocaml-libevent/ocaml-libevent-0.9.0.tar.gz"
+  checksum: [
+    "md5=00851213da0c9ecf37fbd3984a266789"
+    "sha256=e9bcbde7d554f18e52b761601eb4fb8a91df5d160dee03f10882c77a5a83c078"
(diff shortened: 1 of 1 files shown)


packages/libevent/libevent.0.9.0/opam
```diff
@@ -0,0 +1,43 @@
...
+]
+depends: [
+  "ocaml" {>= "4.02.0"}
+  "ocamlfind" {build}
+  "base-unix"
+  "base-bytes" {with-test}
+  "ounit" {with-test}
+  "conf-libevent" {build}
+]
+synopsis: "OCaml wrapper for the libevent API"
```
mseri, line 29:
```suggestion
conflicts: [ "ocaml-option-bytecode-only" ]
synopsis: "OCaml wrapper for the libevent API"
```
I cannot see the error right now, the CI has restarted on its own, but from @avsm comment I believe this is the fix

mseri:
On freebsd it fails with `gcc` not found, perhaps we should mark it unavailable there, see if using `conf-gcc` works or allow the failure in the CI.

On macos/homebrew it fails with
```

#=== ERROR while compiling libevent.0.9.0 =====================================#
# context              2.2.0~beta1 | macos/arm64 | ocaml-base-compiler.4.14.2 | pinned(https://ygrek.org/p/release/ocaml-libevent/ocaml-libevent-0.9.0.tar.gz)
# path                 ~/.opam/4.14.2/.opam-switch/build/libevent.0.9.0
# command              ~/.opam/opam-init/hooks/sandbox.sh build make depend
# exit-code            2
# env-file             ~/.opam/log/libevent-54764-131391.env
# output-file          ~/.opam/log/libevent-54764-131391.out
### output ###
# Makefile:99: depend: No such file or directory
# gcc -I /Users/mac1000/.opam/4.14.2/lib/ocaml -MM *.c > depend
# event_stubs.c:13:10: fatal error: 'event.h' file not found
# #include <event.h>
#          ^~~~~~~~~
# 1 error generated.
# make: *** [depend] Error 1
```
which I think is due to homebrew unusual include paths, we could allow the failure in the CI or make it incompatible with macos

rr0gi:
or mac ci could add homebrew include path to C_INCLUDE_PATH (+ same for LIBRARY_PATH i guess) which is what  homebrew users are supposed to do?

mseri:
I don't know, I don't use homebrew

packages/libevent/libevent.0.9.0/opam
Khady: I’m a bit surprised as we are able to compile this lib with byte code binaries and use it in things like utop. Also the previous version didn’t have this restriction and the changes in the code have been very minimal

mseri: I think that is because the CI did not have a bytecode only compiler at that point. The failure is now accessible again:
```

#=== ERROR while compiling libevent.0.9.0 =====================================#
# context              2.2.0~beta3~dev | linux/x86_32 | ocaml-base-compiler.5.1.1 | pinned(https://ygrek.org/p/release/ocaml-libevent/ocaml-libevent-0.9.0.tar.gz)
# path                 ~/.opam/5.1/.opam-switch/build/libevent.0.9.0
# command              ~/.opam/opam-init/hooks/sandbox.sh build make
# exit-code            2
# env-file             ~/.opam/log/libevent-7-c4eb3c.env
# output-file          ~/.opam/log/libevent-7-c4eb3c.out
### output ###
# ocamlfind ocamlc -package unix -c -ccopt " -Wall " event_stubs.c
# ocamlmklib -oc mloevent event_stubs.o -levent 
# ocamlfind ocamlc -package unix -c  libevent.mli
# ocamlfind ocamlc -package unix -c  -nolabels libevent.ml
# ocamlmklib -o liboevent libevent.cmo -oc mloevent -levent 
# ocamlfind ocamlopt -package unix -c  -nolabels libevent.ml
# ocamlfind: Not supported in your configuration: ocamlopt
# make: *** [Makefile:95: libevent.cmx] Error 2
```
You compile unconditionally both the bytecode and native versions, if no native compiler is available at all then this fails (the conflict is with system that only have bytecode compilers)

avsm:
Agreed not a blocker - merging and can refine Homebrew later.

PR #25794, opened 2024-05-04
release not-ocamlfind 0.13

chetmurthy:
add "-impl", "-intf" to "preprocess" command (pass thru to camlp5).
Also fix Freebsd build process.

### DIFF
--- a/packages/camlp5/camlp5.8.00.05/opam
+++ b/packages/camlp5/camlp5.8.00.05/opam
@@ -34,7 +34,7 @@ depends: [
   "conf-diffutils" { with-test & os-distribution = "alpine" }
   "re"
   "pcre"
-  "ounit" { with-test }
+  "ounit2" { with-test }
   "rresult"
   "bos"
   "fmt"
--- a/packages/camlp5/camlp5.8.00.05/opam
+++ b/packages/camlp5/camlp5.8.00.05/opam
@@ -51,6 +51,7 @@ install: [make "install"]
 conflicts: [
    "ocaml-option-bytecode-only"
 ]
+x-ci-accept-failures: ["freebsd"]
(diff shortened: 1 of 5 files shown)


chetmurthy:
OK, there are three sorts of errors:

1. macos-homebrew: not-ocamlfind depends on "conf-m4" and that doesn't seem to be building on this platform.

2. freebsd: freebsd's diff seems to behave differently from GNU diff, and is throwing errors during a camlp5 test.

3. frama-clang is failing to build on Debian.  I checked doing the following:

  a. take a failing testcase
  b. run the Docker script (uses Debian, OCaml 4.14) on my Ubuntu machine: it fails.
  c. remove my PR from the script, so its just based on master: it still fails in the same way.
  d. then swap out the Debian image for an Ubuntu image: it WORKS.

So I conjecture that there's a problem with frama-clang on Debian.

So it seems none of these test-failures have anything to do with this PR.

raphael-proust:
> 2. freebsd: freebsd's diff seems to behave differently from GNU diff, and is throwing errors during a camlp5 test.

IIRC, you can use `gdiff` when on freebsd. But it's cumbersome to have to check what system you are in and all…

I'm tempted to say that we can merge regardless. But we could also add

```
x-ci-accept-failures: ["freebsd"]
```

chetmurthy:
Good idea, I'll do that.  Also, in more-recent versions I do check for freebsd and use a different diff; since the failures are with older versions, I'll just skip it.

chetmurthy:
alright, looks like

1. still conf-m4 failures on macos-homebrew
2. frama-clang "ld" failures (I didn't look into them, b/c they're appearing in the same places as they did the previous time)

PR #25804, opened 2024-05-06
Add conf-libblake3 virtual package.

dbuenzli:
Add a configuration package for the [extremely spotty support](https://repology.org/project/blake3/versions) of the `libblake3` C library. I guess that's what happens when tools get written in rust :-(


dbuenzli:
I suspect it's normal for these `conf` packages to fail, that's what they are for no ? 

Btw. since I need this on `debian` quickly I have created the [`distrib-crutch.opam`](https://github.com/dbuenzli/distrib-crutch.opam) opam package repository. I don't want to become a package repository maintainer or start fiddling with CI and QA but I may not object simple additions there.

dbuenzli:
Ping. Is there anything more I need to get this merged ?

PR #25811, opened 2024-05-07
Add conf-libmd package.

dbuenzli:


dbuenzli:
Trying to fix `opensuse` and `freebsd` any idea where I can find a package list for `oraclelinux` ?

hannesm:
On FreeBSD, libmd is shipped as part of the base system -- you don't need to have external dependencies specified, just link with `-lmd`. The downside is there's no pkg-config file installed for libmd.

dbuenzli:
Ok so it's rather going to be an unsupported plaform :-) No need to complicate the life of devs. But I guess it works for the other bsds https://pkgsrc.se/devel/libmd

hannesm:
I'm not sure what your mileage is. From https://www.hadrons.org/software/libmd/: "This library provides message digest functions found on BSD systems either on their libc (NetBSD, OpenBSD) or libmd (FreeBSD, DragonflyBSD, macOS, Solaris) libraries and lacking on others like GNU systems."

As I read it, libmd is part of the libc in NetBSD and OpenBSD (so I wouldn't expect a port/package will need to be installed) -- I'm not sure whether pkg-config files are delivered in NetBSD and OpenBSD.

On FreeBSD (and DragonflyBSD) I attempted to explain it earlier. On macOS I suspect you've much more experience and know what is needed there.

Indeed, writing portable code using C libraries on different Linux/BSD platforms is a nightmare. Even with pkg-config - which exists in multiple implementations and versions that have all kinds of special behaviour - be warned.

If supporting 85% of users is sufficient for you, just stick to debian and macOS.

dbuenzli:
So here is the status of this:

1. I removed `freebsd` from the depext @hannesm can you perhaps convince your package manager to install `pkg-config` files for "standard" system libraries. It makes life easier for cross platform development.
2. opensuse still seems to be failing but it seems to be the right [depext name](https://build.opensuse.org/package/show/openSUSE:Factory/libmd)
3. I have no idea where to get information about `oraclelinux`. 

I'd say this is good enough for merging.

dbuenzli:
> If supporting 85% of users is sufficient for you, just stick to debian and macOS.

Everyone's time's limited if you are a minority platform then better provide a good environment for software to be delivered on it. For better or worse `pkg-config` is what we have for now to allow reasonable cross-platform dev in C and I'm not interested in writing conditional build logic for every operating system out there. Writing all these different package names in `depexts` is already ridiculous enough. Is it `$LIB`, `$LIB-dev` or `$LIB-devel` 😂 ?

hannesm:
Sorry to hear that you interpreted my comments that you should do some work. This is not the case - I mainly wanted to help out and avoid you spending time on "why does the FreeBSD CI not work with the proposed package and depext lines".

> This library provides message digest functions found on BSD systems either on their libc (NetBSD, OpenBSD) or libmd (FreeBSD, DragonflyBSD, macOS, Solaris) libraries and lacking on others like GNU systems.

No, as you mention in another comment, "Everyone's time is limited". Why should I bother with this?

> I'm not interested in writing conditional build logic for every operating system out there.

That's fine. Really.

Enjoy your evening, and please don't bother with the "minority platforms". It's not worth it.

dbuenzli:
> Enjoy your evening, and please don't bother with the "minority platforms". It's not worth it.

Please don't make it as if I didn't bother. 

Why would I spend my time going over all these niche linux and bsd distributions to actually find if they named their package `$LIB`, `$LIB-dev` or `$LIB-devel` if I didn't ?

But when I see responses like yours I'm actually asking myself why I actually bother.

jmid:
> 3. I have no idea where to get information about oraclelinux.

I found a couple of Oracle package lists for the two repos available (BaseOS + AppStream) to the CI machines:
- https://yum.oracle.com/repo/OracleLinux/OL9/baseos/latest/x86_64/index.html
- https://yum.oracle.com/repo/OracleLinux/OL9/appstream/x86_64/index.html

(not an Oracle user myself)

hannesm:
Daniel. sorry for my wording. My intention was not to claim that you do not bother. My intention was firstly to help, and secondly to explain some frustration I encountered with "platform-agnostic C libraries".

dbuenzli:
Ping. Is there anything more I need to get this merged ?

packages/conf-libmd/conf-libmd.1.0.0/opam
```diff
@@ -0,0 +1,28 @@
+opam-version: "2.0"
+maintainer: "opam-repository maintainers"
+homepage: "https://www.hadrons.org/software/libmd/"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+authors: ["Guillem Jover" "Aaron D. Gifford" "Markus Friedl" "The NetBSD Foundation, Inc"
+          "Todd C. Miller" "Poul-Henning Kamp" "Colin Plumb"]
+license: "BSD-3-clause AND BSD-2-Clause-NetBSD AND ISC"
+build: [
+  ["pkg-config" "--atleast-version=1.0.0" "libmd"]
```
mseri, line 9:
```suggestion
  ["pkg-config" "--atleast-version=1.0.0" "libmd"] { os-distribution != "freebsd" }
```

dbuenzli:
Thanks !

PR #25812, opened 2024-05-07
[new release] lice (0.0.1-dev)

gabyfle:
The Lice Programming language libraries and interpreter

- Project page: <a href="https://lice.gabyfle.dev/">https://lice.gabyfle.dev/</a>

##### CHANGES:

This is the first version with a working prototype of the Lice interpreter using bytecode and a virtual machine instead of a tree-walk algorithm. Improvements have been made on memory usage and execution times.

### Modified

- Switched the whole interpreter to a bytecode interpreter. This should
  improve performance and reduce memory usage.

- Changed the way the lice library and the interpreter is compiled, making Lice a true OCaml library.

- Reformated the way types works inside the language and made an overall code cleanup.

- Functions are first class citizens.

### Fixed

- Fixed the lexing of float numbers.

Note: This is a pre-release and it's not intended to be used yet.

### DIFF
--- a/packages/lice/lice.0.0.1-dev/opam
+++ b/packages/lice/lice.0.0.1-dev/opam
@@ -1,0 +1,40 @@
+opam-version: "2.0"
+synopsis: "The Lice Programming language libraries and interpreter"
+description:
+  "This package provides the Lice programming language libraries and interpreter. Lice is a simple programming language that is easy to learn and use. It is an imperative programming language that is meant to be interoperable with OCaml."
+maintainer: ["Gabriel Santamaria <gaby.santamaria@outlook.fr>"]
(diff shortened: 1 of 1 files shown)


mseri:
I think the version may not be correct: you are trying to release version `0.0.1-dev` but shouldn't it be `0.0.1`?

mseri:
If you don't want users to install it, then I think it is too early to release it

gabyfle:
@mseri  Yes, actually it was also to test out the way publishing packages on `opam` works, as it was my first time doing it. Also, it looks like the `flambda` backend can't terminate the compilation when `-O2` flag is on.

I'll just close this pull request for the moment and create a new one once this is ready for a real release ! Thanks for your time !

PR #25813, opened 2024-05-08
Retry release camlp5 8.03.00 for ocaml 5.2.0 compat

chetmurthy:
The changes are pretty minimal, *except* that [these changes are documented in the CHANGES file]:

1. the "pr_o.cmo" printer module requires that the "o_keywords.cmo" module be loaded first.  If you instead use the Findlib module "camlp5.pr_o", this is done automatically, but there are some users of Camlp5 who do it the old low-level way.

2. there's a new field "kwds" in the "lexer" record-type.  This is to support pretty-printing raw-identifiers.  Since Camlp5 supports multiple syntaxes, different syntaxes can have different sets of keywords.  There are a couple of projects that create their own lexers, so those need to be updated.

I've sent PRs/emails to the affected projects, and put them in the "conflicts" stanza of the opam file, so users will not be able to install this new Camlp5 with those backlevel projects until they're updated.

### DIFF
--- a/packages/camlp5/camlp5.8.03.00/opam
+++ b/packages/camlp5/camlp5.8.03.00/opam
@@ -34,7 +34,7 @@ depends: [
   "camlp5-buildscripts" { >= "0.02" }
   "conf-diffutils" { with-test & os-distribution = "alpine" }
   "re" { >= "1.11.0" }
-  "ounit" { with-test }
+  "ounit2" { with-test }
   "pcre2" { with-test }
   "rresult"
   "bos"
--- a/packages/camlp5/camlp5.8.03.00/opam
+++ b/packages/camlp5/camlp5.8.03.00/opam
@@ -57,6 +57,7 @@ conflicts: [
    "matita" { <= "0.99.5" }
    "lablgl" { <= "1.07" }
 ]
+x-ci-accept-failures: [ "opensuse" ]
 dev-repo: "git+https://github.com/camlp5/camlp5.git"
 url {
   src: "https://github.com/camlp5/camlp5/archive/refs/tags/8.03.00.tar.gz"
--- a/packages/camlp5/camlp5.8.03.00/opam
+++ b/packages/camlp5/camlp5.8.03.00/opam
@@ -1,0 +1,61 @@
+
+opam-version: "2.0"
+synopsis: "Preprocessor-pretty-printer of OCaml"
+description: """
+Camlp5 is a preprocessor and pretty-printer for OCaml programs. It also provides parsing and printing tools.
+
+As a preprocessor, it allows to:
+
+extend the syntax of OCaml,
+redefine the whole syntax of the language.
+As a pretty printer, it allows to:
+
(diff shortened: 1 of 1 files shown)


chetmurthy:
Oh, I see that 5.2~rc1 is not one of the compiler-versions that is tested.  I guess that means I should wait to release this version of Camlp5 until after 5.2.0 is released ?  B/c otherwise, it won't be tested against 5.2.0 -- it fails against 5.2~beta2 (I found a few pretty-printing bugs).

packages/camlp5/camlp5.8.03.00/opam
```diff
@@ -0,0 +1,66 @@
...
+  [make "-C" "test" "clean" "all"] { with-test & os != "macos" }
+#  [make "-C" "scripts" "clean" "test"] { with-test }
+]
+install: [make "install"]
+conflicts: [
+   "ocaml-option-bytecode-only"
+   "pa_ppx" { < "0.15" }
+   "p5scm" { <= "0.3.1" }
+   "matita" { <= "0.99.5" }
+   "ocaml" { = "5.2.0~beta2" }
```
kit-ty-kate, line 58: `ocaml` is a meta package does not have the beta versions
```suggestion
   "ocaml-base-compiler" { = "5.2.0~beta2" }
   "ocaml-variants" { = "5.2.0~beta2+options" }
```

chetmurthy:
Looking at the test-runs,  I see that OCaml 5.2.0~rc1 wasn't part of the tests.  So I suppose I should wait until the release happens, and rerun the CI.  At that point, I can just remove mention fo 5.2.0~beta2 completely.

chetmurthy:
It looks like OCaml 5.2.0 is not yet part of the opam-ci set of tested compiler releases.  So this code didn't actually get tested with the new OCaml.  I'll resubmit in a couple of days; hopefully by then 5.2.0 will be included.

chetmurthy:
I tried both "opensuse" and "opensuse-tumbleweed", in the following:
```
-x-ci-accept-failures: [ "opensuse" ]
+x-ci-accept-failures: [ "opensuse-tumbleweed" ]
```
and I still see opensuse test-failures called-out in the CI run.  Not sure what I need to do to make them go away, but in any case, they're due to busybox "diff" being different from GNU diff (and I need GNU diff for the tests).

chetmurthy:
OK, it looks like the only test-failures left are expected:

1. macos: conf-m4 failure
2. frama-clang: for some reason doesn't build on Debian
3. opensuse: problem with diff for tests, can't figure out how to ignore test-failure
4. freebsd: test fails to even start

I think this is good-to-go.

mseri:
You cannot avoid the error on suse: it is a failure in the install of a system package, the way things work this propagates through

mseri:
I am re-running one of the freebsd jobs now that the machine seems to work again

chetmurthy:
Looks like all the Freebsd went thru OK.

mseri:
Thanks and sorry for the long delay

PR #25821, opened 2024-05-09
Package cudajit.0.1.1

lukstafi:
### `cudajit.0.1.1`
Bindings to the `cuda` and `nvrtc` libraries with a unified interface
Bindings to manually selected parts of `lcuda` and `lnvrtc`, with a few types and conversion functions to facilitate use.

---
* Homepage: https://github.com/lukstafi/ocaml-cudajit
* Source repo: git+https://github.com/lukstafi/ocaml-cudajit.git
* Bug tracker: https://github.com/lukstafi/ocaml-cudajit/issues

---


raphael-proust:
There are CI failures in most (all?) distributions apart from Debian. They are all related to the `cuda-toolkit` package rather than this here package being released.

I don't think these failures are blockers for this PR. Although it would be good to fix the `cuda-toolkit` package too (in a separate PR, later).

lukstafi:
> There are CI failures in most (all?) distributions apart from Debian. They are all related to the `cuda-toolkit` package rather than this here package being released.
> 
Depending on `cuda-toolkit` unconditionally was a very bad idea, because users may not have this package for good reasons.

I decided to follow the recommended practice and have a separate `depexts` package `conf-cuda`. Due to opam's technical limitations, this first became two packages `conf-cuda-config` and `conf-cuda`, and then I gave up on `depexts` and instead `conf-cuda` prints suggested installation instructions. The best case scenario would be that we had passing CI for Ubuntu and ArchLinux, but the current limitation on package variables means we will not have any passing CI.

I will re-submit `cudajit.0.1.1` once the `conf-cuda` PR is submitted. Apologies for this distracting PR!

PR #25826, opened 2024-05-10
Stop using platform@lists.ocaml.org as a maintainer

dra27:
In the midst of the various changes to the compiler packages I have been working on for Windows, I have been adding missing fields to core support packages in order to satisfy `opam lint`. While doing that, I spotted that many packages were still tagged `maintainer: "platform@lists.ocaml.org"`. Upon grep'ing, I discovered that it's exclusively compiler and infrastructure packages which still use it!

It seems sensible to have named maintainers: the `bug-reports` field tells you were to file the issue (here) and the maintainers list gives a hint as to who to tag. In particular, it seems sensible to me to have the `maintainers` fields of these packages indicate people who would expect to be made aware of changes. `maintainers` is intended to hold maintainers of the opam packages, rather than strictly maintainers of the software itself.

To that end, this PR removes the use of platform@lists.ocaml.org from all remaining packages. Note that updating the `maintainer` field does not trigger switch recompilation in any version of opam. I have split the changes into 5 commits (which can be further split if necessary):
- **Infrastructure**: compiler infrastructure (ocaml-config, ocaml-option-\*, ocaml-options-\* and the ocaml package itself). These relate to the implementation of OCaml expressed by opam within opam-repository. The packages were all written by opam maintainers.
- **Modern**: the "modern" (4.08+) versions of OCaml (ocaml-base-compiler, ocaml-system, ocaml-variants and the ocaml-beta support package).
- **Variants**: 4.08+ packages in ocaml-variants which are actual variants of OCaml (+rescript and +BER).
- **Museum**: packages for OCaml 4.07 and earlier.
- **Secondary**: packages supporting the auxiliary installation of a more recent OCaml in an older switch (ocaml-secondary-compiler and ocamlfind-secondary).

As [your starter for ten](https://en.wikipedia.org/wiki/University_Challenge#Question_format), I propose that I indeed actively maintain these packages (in the case of the secondary compiler, for my many sins and in the case of the "Museum", for my many quirks...) and that @Octachron co-maintains the "Modern" compiler packages.

If you:
- Disagree (!)
- Would like to be added to a list (or, in @Octachron's case, would like to be removed from a list!)
- Think we should do something completely different

Please chime in! The only thing I absolutely believe we should do is stop using an email address which I don't think still works and which I certainly know one of the principal maintainers of these packages isn't checking...!

cc @avsm, @AltGr, @Octachron, @kit-ty-kate (list of people from git log who have clearly maintained these packages, looking at the git log), @ocaml/opam-dev, @yallop (BER variants), @bobzhang (+rescript variants)


avsm:
Thanks, this invalid email address has been bugging me for ages ;-)

avsm:
I'm going to go ahead and merge this in the interests of saving the CI an enormous number of rebuilds, and it can be refined subsequently with the questions David notes above.  It is clearly an improvement over the current invalid maintainer address. /cc @mtelvers as the CI may be overloaded by this.

hannesm:
I highly appreciate this. Thanks for stepping up and fixing the metadata. 👍🏾

dra27:
Thanks! I think this kind of PR only triggers a lint check, because the Analysis phase reports InsignificantlyChanged for each package (it's the complete opposite when compiler packages are significantly edited, where I think there is a huge amount of not-strictly-necessary churn)

PR #25827, opened 2024-05-10
conf-cuda: check for or install CUDA libraries

lukstafi:
In the first commit, dependencies for distributions that have CUDA packages.
In the second commit, I added ArchLinux compatibility and specified installation instructions in case of failure.


lukstafi:
On further thought, this should be done better, I'll also create a `conf-cuda-config` package to skip `depexts` in case CUDA is already (suspected to be) installed, and to check for WSL.

PR #25828, opened 2024-05-10
Backport ocaml/ocaml#13100 to 5.1.x ocaml-variants

dra27:
When building with ocaml-option-musl, zstd detection in OCaml fails if pkg-config and libzstd are installed for the host. This is fixed in 5.2.0. ocaml-base-compiler is unaffected.

This PR is a draft and should not be merged until after opam 2.2.0 beta3 is released as it breaks ocaml-variants.5.1.1+effects, ocaml-variants.5.1.0+options and ocaml-variants.5.1.1+options on Windows (https://github.com/ocaml/opam/pull/5921).


mseri:
@dra27 Is this safe to merge?

dra27:
It should be, yes (🤞)

mseri:
Thanks

PR #25861, opened 2024-05-10
Update ocaml-base-compiler, ocaml-system and ocaml-variants 4.13.0+ with support for native Windows

dra27:
This PR augments the three compiler packages (`ocaml-base-compiler`, `ocaml-system` and `ocaml-variants`) with support for the MSVC and mingw-w64 native Windows ports for 4.13.0+. I intend to extend this support back to 4.08 as part of an ongoing overhaul of the compiler packages, but that is beyond the scope of this PR.

## Principles

The implementation presented here stems from three underlying principles:

1. All four native Windows ports should be able to be co-installed in the same opam root (i.e. `~/.opam`) as separate switches (as with opam-repository-mingw). That is, the user should not be forced to choose permanently between MSVC/mingw-w64 and/or amd64/i686 at `opam init`.
2. The Windows ports should not be made to look like Tier 2 or alternate platforms; i.e. the instructions to create a Windows OCaml switch should not be _fundamentally_ different from other platforms.
3. Depexts requirements should be precise, so the installation of a `conf-` package should not speculatively install more dependencies than necessary for the given switch. In particular, if a switch is configured with i686 OCaml, the installation of conf-libfoo should install the required packages for i686 libfoo, not both i686 and x86_64 libfoo.

<details>
  <summary>Corollaries</summary>
    
There are some immediate observations and problems:
- The first principle matches the existing capability of [opam-repository-mingw](https://github.com/ocaml-opam/opam-repository-mingw).
- The second principle prohibits either _requiring_ `opam exec` or needing a Windows-specific wrapper (such as `with-dkml` or `ocaml-env`).
- The second principle also has implications for the `ocaml-base-compiler` package, since there is no concept of a "default" C compiler on Windows. The mingw-w64 and MSVC ports of OCaml are not interchangeable in the way that GCC or Clang-based OCaml is.
- The depext system in [OCaml for Windows](https://fdopen.github.io/opam-repository-mingw) was based on opam 2.0's opam-depext plugin, and never adopted opam 2.1's integrated support. While the use of the plugin means that it suffers from the same solving problems that caused depext to be integrated directly into the solver in opam 2.1, it did mean that `opam depext` knew which compiler was installed in the switch, because it was necessarily run _after_ the switch had been created.
- The `depexts` section of an opam file can only be filtered on _global_ variables. While we can (just about) set switch-specific global variables, this would be both awkward (users would need to know of an extra step), but would also contradict the second principle, adding a requirement to Windows-specific parts of the workflow. This means that the depexts for Windows need to be recorded in separate `conf-` packages, and the dependency graph of a switch needs to record sufficient information about the architecture and C compiler to select the correct package (in short, we require more `base-`-like packages).
</details>

<details>
    <summary>Other related work</summary>
    
This work overlaps with some considerable ongoing additional work on the compiler's opam packages:
- #17541 added the `ocaml-option-` packages to reduce the number of variants, but this applies to OCaml 4.12+ only. This facility needs extending (on principle) back to 4.11 and earlier. It makes little sense to update all the individual variants for 4.11 and earlier without first addressing this issue (it's side-stepped for now as 4.13 is the minimum version updated for a different technical reason).
- As I note in https://github.com/ocaml/opam-repository/pull/24075#issuecomment-1625271501, #22246 forged the wrong approach to fixing bytecode-only 5.x architectures started in #21510. This PR doesn't attempt to fix those issues, but the new virtual packages are added in both "input" (`ocaml-option-*`) form and "output" (`base-*`), which I believe to be the fundamental error I (/we) failed to see in #21510.

I was originally (back in September) of the opinion that it would be better to solve these two issues _first_ and then add Windows support as an extension of these fixes. However, while it's tempting to engineer things such that Windows becomes a "minor" addition, the compatibility concerns for these two fixes make them much higher risk than the Windows packages, which have no compatibility story to worry about. I've therefore restructured the changes so that the alterations are made Windows-only _for now_, with the fixes to Unix following later, lifting these "limitations".
</details>

## How it works

With opam 2.2.0 beta2, `opam init` (if pointed to this branch) will create an OCaml 5.1.1 switch. Concretely, on a clean Windows 11 system:

```
winget install Git.Git
winget install opam
```

and then in fresh terminal followed by:

```batch
rem Accept all defaults for opam init
opam init git+https://github.com/dra27/opam-repository.git#windows-initial
opam exec -- ocaml
```

will give OCaml 5.1.1! When creating a switch, an `arch-` or `system-` package can simply be added just as for the `ocaml-option-` packages. For example, assuming the user has installed Visual Studio which, amongst methods, may be done with:

```
winget install Microsoft.VisualStudio.2022.BuildTools --override "--add Microsoft.VisualStudio.Workload.VCTools --includeRecommended --passive"
```

then a 32-bit MSVC 4.14.2 switch may be created with:

```
opam switch create 4.14.2-msvc32 ocaml.4.14.2 system-msvc arch-x86_32
echo print_endline "Hello, world" > hello.ml
opam exec -- ocamlopt -o hello.exe hello.ml
```

Note that these steps do _not_ require the user to start a Visual Studio Tools Command Prompt or do anything beyond installing Visual Studio.

## Under the hood

In more detail, at present, the `ocaml.x.y.z` package encodes the version of OCaml being installed. To this, at present for Windows only, I have added two more sets of packages:
- `arch-x86_32` and `arch-x86_64` allow the choice between the i686 and amd64 architectures.
- `system-mingw` and `system-msvc` provide the choice between the mingw-w64 and Microsoft Visual Studio (MSVC) ports.

_Both_ `ocaml-base-compiler` and `ocaml-variants` use these two sets of packages. `ocaml-base-compiler` remains "OCaml in its default configuration", but it becomes possible to control exactly which C compiler configuration it's using.

The default compiler is amd64 mingw-w64 (i.e. `arch-x86_64` and `system-mingw` will be automatically added if no other `arch-` or `system-` package has been selected) for two reasons:
- We can't detect MSVC using opam's depext system at present (but we can automatically install mingw-w64), so mingw-w64 as a default means that `opam init` always builds a working OCaml
- Cygwin (and MSYS2) are not available for 32-bit systems, so users will be on 64-bit Windows.

For now, it is intentionally not possible to install the Cygwin port of OCaml using native Windows opam.

Where the _user_ installs a `system-` and an `arch-` package, there are also new sets of `host-arch-` and `host-system-` which are installed by all opam switches. The idea here is that one `host-system-` and one `host-arch-` package are _always_ installed in a switch (be that `ocaml-base-compiler`, `ocaml-variants` **or** `ocaml-system`).

<details>
  <summary>More information for users</summary>
    
The key rule is that `arch-` and `system-` should never be used in opam files because there are packages (such as `ocaml-system`) which don't use them. These are also packages which may also disappear in the future if opam gains a way to specify configuration options for to packages at installation time.

From the user's perspective, specifying `arch-x86_64` vs `host-arch-x86_64` is similar to the difference between specifying `ocaml-base-compiler.4.14.1` vs `ocaml.4.14.1`. `ocaml-base-compiler.4.14.1` instructs opam to build 4.14.1 from source, where `ocaml.4.14.1` permits the use of a system compiler.
</details>

The motivation for this change is to be able to indicate precisely where packages are not supported:
- `available: os != "win32"` is the sledgehammer: no Windows support at all
- `conflicts: "host-system-msvc"`: this package works with the mingw-w64 ports, but doesn't work with the Visual Studio ports
- `conflicts: "host-arch-x86_32"`: this package doesn't work on 32-bit Intel
- `depends: "host-system-mingw"`: this package _only_ works with the mingw-w64 ports.

## Notes

I've attempted to organise the changes into a meaningful commit series, which is slightly easier to review than the entire diff in one go. I've added missing metadata fields packages in order to pass `opam lint`.

### DIFF
--- a/packages/conf-c++/conf-c++.1.0/opam
+++ b/packages/conf-c++/conf-c++.1.0/opam
@@ -4,7 +4,12 @@ authors: "C++ compiler developers"
 homepage: "https://github.com/ocaml/opam-repository"
 bug-reports: "https://github.com/ocaml/opam-repository/issues"
 license: "GPL-2.0-or-later"
-build: ["c++" "--version"]
+build: [
+  "i686-w64-mingw32-c++" {os = "win32" & host-arch-x86_32:installed}
+  "x86_64-w64-mingw32-c++" {os = "win32" & host-arch-x86_64:installed}
+  "c++" {os != "win32" | (!host-arch-x86_32:installed & !host-arch-x86_64:installed)}
+  "--version"
+]
 depexts: [
   ["gcc-c++"] {os-distribution = "centos"}
   ["gcc-c++"] {os-distribution = "fedora"}
@@ -16,6 +21,9 @@ depexts: [
   ["gcc"] {os-distribution = "arch"}
   ["gcc-g++"] {os = "win32" & os-distribution = "cygwinports"}
 ]
+depends:
+  ("host-arch-x86_32" {os = "win32" & os-distribution != "cygwinports"} & "conf-mingw-w64-g++-i686" {os = "win32" & os-distribution != "cygwinports"}) |
+  ("host-arch-x86_64" {os = "win32" & os-distribution != "cygwinports"} & "conf-mingw-w64-g++-x86_64" {os = "win32" & os-distribution != "cygwinports"})
 synopsis: "Virtual package relying on the c++ compiler"
 description:
   "This package can only install if the c++ compiler is installed on the system."
--- a/packages/conf-c++/conf-c++.1.0/opam
+++ b/packages/conf-c++/conf-c++.1.0/opam
@@ -22,8 +22,8 @@ depexts: [
   ["gcc-g++"] {os = "win32" & os-distribution = "cygwinports"}
 ]
 depends:
-  ("host-arch-x86_32" {os = "win32" & os-distribution != "cygwinports"} & "conf-mingw-w64-g++-i686" {os = "win32" & os-distribution != "cygwinports"}) |
-  ("host-arch-x86_64" {os = "win32" & os-distribution != "cygwinports"} & "conf-mingw-w64-g++-x86_64" {os = "win32" & os-distribution != "cygwinports"})
+  (("host-arch-x86_32" {os = "win32" & os-distribution != "cygwinports"} & "conf-mingw-w64-g++-i686" {os = "win32" & os-distribution != "cygwinports"}) |
+   ("host-arch-x86_64" {os = "win32" & os-distribution != "cygwinports"} & "conf-mingw-w64-g++-x86_64" {os = "win32" & os-distribution != "cygwinports"}))
 synopsis: "Virtual package relying on the c++ compiler"
 description:
   "This package can only install if the c++ compiler is installed on the system."
--- a/packages/host-arch-arm64/host-arch-arm64.1/opam
+++ b/packages/host-arch-arm64/host-arch-arm64.1/opam
@@ -1,0 +1,16 @@
+opam-version: "2.0"
+synopsis: "OCaml on AArch64 (64-bit)"
+description: """
+This package is installed if the underlying OCaml compiler is for 64-bit ARM.
+
+Precisely, this means `ocamlopt -config-var architecture` equals `arm64`.
+
+This package may be used in depends or conflicts fields of dependent packages
+to indicate either a requirement or an incompatibility with this
+architecture."""
+maintainer: "David Allsopp <david@tarides.com>"
+authors: "David Allsopp"
+license: "CC0-1.0+"
+homepage: "https://opam.ocaml.org"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+conflict-class: "ocaml-host-arch"
--- a/packages/host-arch-ppc64/host-arch-ppc64.1/opam
+++ b/packages/host-arch-ppc64/host-arch-ppc64.1/opam
@@ -1,0 +1,17 @@
+opam-version: "2.0"
+synopsis: "OCaml on 64-bit IBM POWER"
+description: """
+This package is installed if the underlying OCaml compiler is for
+64-bit IBM POWER.
+
+Precisely, this means `ocamlopt -config-var architecture` equals `power`.
+
+This package may be used in depends or conflicts fields of dependent packages
+to indicate either a requirement or an incompatibility with this
+architecture."""
+maintainer: "David Allsopp <david@tarides.com>"
+authors: "David Allsopp"
+license: "CC0-1.0+"
+homepage: "https://opam.ocaml.org"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+conflict-class: "ocaml-host-arch"
--- a/packages/host-arch-riscv64/host-arch-riscv64.1/opam
+++ b/packages/host-arch-riscv64/host-arch-riscv64.1/opam
@@ -1,0 +1,16 @@
+opam-version: "2.0"
+synopsis: "OCaml on 64-bit RISC-V"
+description: """
+This package is installed if the underlying OCaml compiler is for 64-bit RISC-V.
+
+Precisely, this means `ocamlopt -config-var architecture` equals `riscv`.
+
+This package may be used in depends or conflicts fields of dependent packages
+to indicate either a requirement or an incompatibility with this
+architecture."""
+maintainer: "David Allsopp <david@tarides.com>"
+authors: "David Allsopp"
+license: "CC0-1.0+"
+homepage: "https://opam.ocaml.org"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+conflict-class: "ocaml-host-arch"
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.4.13.0/opam
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.4.13.0/opam
@@ -8,21 +8,65 @@ license: "LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception"
 authors: "Xavier Leroy and many contributors"
 homepage: "https://ocaml.org"
 bug-reports: "https://github.com/ocaml/opam-repository/issues"
-dev-repo: "git+https://github.com/ocaml/ocaml"
+dev-repo: "git+https://github.com/ocaml/ocaml#4.13"
 depends: [
+  # This is OCaml 4.13.0
   "ocaml" {= "4.13.0" & post}
+
+  # General base- packages
   "base-unix" {post}
   "base-bigarray" {post}
   "base-threads" {post}
+
+  # Architecture (non-Windows)
+  # opam-repository at present requires that ocaml-base-compiler is installed
+  # using an architecture which matches the machine's, since arch is used in
+  # available fields. Cross-compilation at this stage is an unstable accident.
+  "host-arch-arm32" {arch = "arm32" & post}
+  "host-arch-arm64" {arch = "arm64" & post}
+  "host-arch-ppc64" {arch = "ppc64" & post}
+  "host-arch-riscv64" {arch = "riscv64" & post}
+  "host-arch-s390x" {arch = "s390x" & post}
+  # The Windows ports explicitly select the architecture (see below) this
+  # facility is not yet available for other platforms.
+  "host-arch-x86_32" {os != "win32" & arch = "x86_32" & post}
+  "host-arch-x86_64" {os != "win32" & arch = "x86_64" & post}
+  "host-arch-unknown" {os != "win32" & arch != "arm32" & arch != "arm64" & arch != "ppc64" & arch != "riscv64" & arch != "s390x" & arch != "x86_32" & arch != "x86_64" & post}
+
+  # Port selection (Windows)
+  # amd64 mingw-w64 / MSVC
+  ("arch-x86_64" {os = "win32" & arch = "x86_64"} &
+    ("system-mingw" & "mingw-w64-shims" {os-distribution = "cygwin" & post} |
+     "system-msvc") |
+  # i686 mingw-w64 / MSVC
+   "arch-x86_32" {os = "win32"} &
+    ("system-mingw" & "mingw-w64-shims" {os-distribution = "cygwin" & post} |
+     "system-msvc") |
+  # Non-Windows systems
+   "host-system-other" {os != "win32" & post})
+
+  # OCaml with default configuration (no flambda, TSAN, etc.)
   "ocaml-options-vanilla" {post}
+
+  # Support Packages
+  "flexdll" {>= "0.36" & os = "win32"}
 ]
 conflict-class: "ocaml-core-compiler"
 flags: compiler
+build-env: MSYS2_ARG_CONV_EXCL = "*"
 setenv: CAML_LD_LIBRARY_PATH = "%{lib}%/stublibs"
+x-env-path-rewrite: [
+  [CAML_LD_LIBRARY_PATH (";" {os = "win32"} ":" {os != "win32"}) "target"]
+]
 build: [
   [
     "./configure"
+    "--host=x86_64-pc-windows"  {system-msvc:installed & arch-x86_64:installed}
+    "--host=x86_64-w64-mingw32" {os-distribution = "cygwin" & system-mingw:installed & arch-x86_64:installed}
+    "--host=i686-pc-windows"    {system-msvc:installed & arch-x86_32:installed}
+    "--host=i686-w64-mingw32"   {os-distribution = "cygwin" & system-mingw:installed & arch-x86_32:installed}
     "--prefix=%{prefix}%"
+    "--with-flexdll=%{flexdll:share}%" {os = "win32" & flexdll:installed}
     "-C"
     "CC=cc" {os = "openbsd" | os = "macos"}
     "ASPP=cc -c" {os = "openbsd" | os = "macos"}
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.4.13.0/opam
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.4.13.0/opam
@@ -35,13 +35,13 @@ depends: [
 
   # Port selection (Windows)
   # amd64 mingw-w64 / MSVC
-  ("arch-x86_64" {os = "win32" & arch = "x86_64"} &
-    ("system-mingw" & "mingw-w64-shims" {os-distribution = "cygwin" & post} |
-     "system-msvc") |
+  (("arch-x86_64" {os = "win32" & arch = "x86_64"} &
+     (("system-mingw" & "mingw-w64-shims" {os-distribution = "cygwin" & post}) |
+      "system-msvc")) |
   # i686 mingw-w64 / MSVC
-   "arch-x86_32" {os = "win32"} &
-    ("system-mingw" & "mingw-w64-shims" {os-distribution = "cygwin" & post} |
-     "system-msvc") |
+   ("arch-x86_32" {os = "win32"} &
+     (("system-mingw" & "mingw-w64-shims" {os-distribution = "cygwin" & post}) |
+      "system-msvc")) |
   # Non-Windows systems
    "host-system-other" {os != "win32" & post})
 
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.5.0.0/opam
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.5.0.0/opam
@@ -8,25 +8,65 @@ license: "LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception"
 authors: "Xavier Leroy and many contributors"
 homepage: "https://ocaml.org"
 bug-reports: "https://github.com/ocaml/opam-repository/issues"
-dev-repo: "git+https://github.com/ocaml/ocaml"
+dev-repo: "git+https://github.com/ocaml/ocaml#5.0"
 depends: [
+  # This is OCaml 5.0.0
   "ocaml" {= "5.0.0" & post}
+
+  # General base- packages
   "base-unix" {post}
   "base-bigarray" {post}
   "base-threads" {post}
   "base-domains" {post}
   "base-nnp" {post}
+
+  # Architecture (non-Windows)
+  # opam-repository at present requires that ocaml-base-compiler is installed
+  # using an architecture which matches the machine's, since arch is used in
+  # available fields. Cross-compilation at this stage is an unstable accident.
+  "host-arch-arm32" {arch = "arm32" & post}
+  "host-arch-arm64" {arch = "arm64" & post}
+  "host-arch-ppc64" {arch = "ppc64" & post}
+  "host-arch-riscv64" {arch = "riscv64" & post}
+  "host-arch-s390x" {arch = "s390x" & post}
+  # The Windows ports explicitly select the architecture (see below) this
+  # facility is not yet available for other platforms.
+  "host-arch-x86_32" {os != "win32" & arch = "x86_32" & post}
+  "host-arch-x86_64" {os != "win32" & arch = "x86_64" & post}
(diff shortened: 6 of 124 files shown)


dra27:
I have a lot of markdown notes on the full compiler package infrastructure, as well as further notes on the changes here which I'll consolidate into a single piece of markdown which can finally form some documentation (probably to go in the wiki here).

dra27:
(Some of the packages are triggering a lint warning which will be addressed in opam 2.2.0 beta3 by the final version of https://github.com/ocaml/opam/pull/5927)

jonahbeckford:
If you are making changes to `ocaml-config`, I have been stuck not being able to submit DkML into the opam repository because of https://gitlab.com/dkml/distributions/dkml/-/releases/2.1.1#likely-permanent-incompatibilities

The root cause are my changes at https://github.com/diskuv/diskuv-opam-repository/tree/main/packages/ocaml-config so that the delegate executables (ex. `ocamlc-real.exe`) used by the DkML shims (ex. `ocamlc.exe`) are recognized. There was some restriction on OCaml 5 only for `ocaml-config.3` which makes it impossible for me to upstream.

dra27:
ocaml-config.3 is OCaml 5 only when not on Windows - is that insurmountable? The change made here is to be explicit that Windows _always_ requires ocaml-config.3. I think this may be a separate discussion (especially if those delegate executables relate to relocation).

packages/host-arch-ppc64/host-arch-ppc64.1/opam
```diff
@@ -0,0 +1,17 @@
+opam-version: "2.0"
+synopsis: "OCaml on 64-bit IBM POWER"
+description: """
+This package is installed if the underlying OCaml compiler is for
+64-bit IBM POWER.
+
+Precisely, this means `ocamlopt -config-var architecture` equals `power`.
```
avsm, line 7: Just a "nit" question here: why `ppc64` instead of the OCaml `power` term in the package?

avsm:
A very charismatic PR, thank you @dra27 :-) I'm not quite sure what the best way to review this; might you have any suggestions?  There are three considerations:

- new arch/host package scheme: this looks useful for other packages as well (for example for simpler selection of 64-bit architectures ?).
- changes to ocaml-base-compiler packages for existing non-Windows architectures: is the expected standard that "no behavioural changes are expected" here?
- do the new compilers work on Windows: do we have a story aside from manual testing yet? Manual testing is just fine for now, but it would be a useful guidance to opam-repo maintainers to be explicit here.

packages/host-arch-ppc64/host-arch-ppc64.1/opam
dra27: Indeed - the namespacing here is just horrible from a usability perspective (recalling near errors over riscv vs riscv64, etc.!) - so many different names, and often in strings 😱

I opted that opam-based consistency was the best thing (or "least-worst"), so went with the values opam would use for `%{arch}%` on the basis that both are likely to be used in opam files and while it's odd that `ocaml -config-var architecture` gives `power`, it's probably stranger to have to write `"host-arch-power" {arch = "ppc64"}` in the opam files.

opam's `arch` variable is whatever `uname -m` gives with normalisations as given in `OpamSysPoll.normalise_arch` which pretty much reduce them to x86_64/x86_32, ppc64/ppc32, arm64/arm32, falling back to the `uname -m` value.

dra27:
Thanks 🙂 I certainly intend to extend the `arch-` and `system-` packages to allow better control of the compiler build on Unix systems too (for example, selecting between gcc and clang and being able to control the selection of arm64 or x86_64 on macOS). The host-arch- and host-system- packages work completely on Windows because opam is able to encode the defaults, but to do this on Unix will require some changes in opam (which, probably unsurprisingly, I'm already scheming towards!), but the packages as given here should at least encode the current slightly limited behaviour (cf. things like https://github.com/ocaml/opam/issues/5949). The main thing is that we need a way to communicate a tiny bit more information about the C compiler to the solver (as, for example, we already do about any OCaml "system" compiler).

It should indeed be the case that nothing has changed for non-Windows apart from the installation of `host-system-other` and a `host-arch-` package.

There isn't a testing story as yet beyond manual testing, but there isn't a story for the Unix compiler packages either. At the moment, ocaml-variants packages are over-tested because of a "bug" in docker-base-images and ocaml-base-compiler packages aren't tested at all.

dra27:
I wonder whether a synchronous review may be best?

raphael-proust:
I think a synchronous review would definitely help me… I'm pinging you offband about it

dra27:
> At the moment, ocaml-variants packages are over-tested because of a "bug" in docker-base-images and ocaml-base-compiler packages aren't tested at all.

A slight refinement - _new_ compiler releases aren't tested at all, but patches to the latest release of existing branches are (although, again, it is more that they're tested by accident than deisgn; cf. [ocaml-base-compiler.4.14.2](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/87c3caa8e8025fbeb6be3fea49759b482963b0b2/variant/compilers,4.14,ocaml-base-compiler.4.14.2), [ocaml-base-compiler.5.0.0](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/87c3caa8e8025fbeb6be3fea49759b482963b0b2/variant/compilers,5.0,ocaml-base-compiler.5.0.0), [ocaml-base-compiler.5.1.1](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/87c3caa8e8025fbeb6be3fea49759b482963b0b2/variant/compilers,5.1,ocaml-base-compiler.5.1.1))

packages/system-msvc/system-msvc.1/opam
```diff
@@ -0,0 +1,20 @@
+opam-version: "2.0"
+synopsis: "Build OCaml for Microsoft Visual Studio"
+description: """
+This package specifies OCaml built with Microsoft Visual Studio and is presently
+presently available for i386/x86_32 and amd64/x86_64.
```
raphael-proust, line 5:
```suggestion
This package specifies OCaml built with Microsoft Visual Studio and is presently
available for i386/x86_32 and amd64/x86_64.
```

"presently" is duplicated

packages/host-arch-riscv64/host-arch-riscv64.1/opam
```diff
@@ -0,0 +1,16 @@
...
+
+This package may be used in depends or conflicts fields of dependent packages
+to indicate either a requirement or an incompatibility with this
+architecture."""
+maintainer: "David Allsopp <david@tarides.com>"
+authors: "David Allsopp"
+license: "CC0-1.0+"
+homepage: "https://opam.ocaml.org"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+conflict-class: "ocaml-host-architecture"
```
raphael-proust, line 16:
```suggestion
conflict-class: "host-arch"
```

to match the now-reserved prefix for package names

needs to be applied to other packages too

raphael-proust: actually maybe using just the prefix is a bit short, especially for some other packages like `arch-`. So possibly we should have `meta-package-<prefix>` or `opam-setup-selection-<prefix>` or something like that.

We'll discuss this during the meeting (or anyone should feel free to drop comments here) and decide on something.

dra27:
First rebase to clear conflicts with #25875 and #25911; now addressing review

packages/system-msvc/system-msvc.1/opam
dra27: Oops - fixed!

packages/host-arch-riscv64/host-arch-riscv64.1/opam
dra27: What was the conclusion for these ones? The PR at the moment introduces:
- `msys2-environment`
- `ocaml-architecture-setting` & `ocaml-system-setting`
- `ocaml-host-architecture` & `ocaml-host-system`
- `ocaml-mingw-env` & `ocaml-msvc-env`

Prior to this PR, opam-repository's only use of `conflict-class` is `ocaml-core-compiler` (which is why the feature was added)

packages/conf-c++/conf-c++.1.0/opam
```diff
@@ -4,7 +4,12 @@ authors: "C++ compiler developers"
 homepage: "https://github.com/ocaml/opam-repository"
 bug-reports: "https://github.com/ocaml/opam-repository/issues"
 license: "GPL-2.0-or-later"
-build: ["c++" "--version"]
+build: [
+  "i686-w64-mingw32-c++" {os = "win32" & host-arch-x86_32:installed}
```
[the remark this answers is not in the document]
dra27: Not quite - 64-bit hosts can run 32-bit executables, so 32-bit is alive and well, it's just that you can't (easily) run opam on actual 32-bit Windows 10.

The point of the comment is that I've put `(arch = "x86_64" | arch = "arm64")`, but we're extremely unlikely ever to see `arch = "x86_32"`

packages/ocaml-base-compiler/ocaml-base-compiler.5.0.0/opam
```diff
@@ -8,25 +8,65 @@ license: "LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception"
 authors: "Xavier Leroy and many contributors"
 homepage: "https://ocaml.org"
 bug-reports: "https://github.com/ocaml/opam-repository/issues"
-dev-repo: "git+https://github.com/ocaml/ocaml"
+dev-repo: "git+https://github.com/ocaml/ocaml#5.0"
```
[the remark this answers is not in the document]
dra27: Hah - I think this is an artefact of how I applied the patches across them - I would have diff'd 4.14.1 and 5.0.0 and not spotted the diff on the dev-repo but when I applied the changes to 5.1.0 I saw the diff, corrected the 5.0 URL but then didn't make the connection about the 4.13 and 4.14 packages! Done...!

packages/ocaml-base-compiler/ocaml-base-compiler.4.13.0/opam
```diff
@@ -10,19 +10,63 @@ homepage: "https://ocaml.org"
...
+  # amd64 mingw-w64 / MSVC
+  ("arch-x86_64" {os = "win32" & arch = "x86_64"} &
+    ("system-mingw" & "mingw-w64-shims" {os-distribution = "cygwin" & post} |
+     "system-msvc") |
+  # i686 mingw-w64 / MSVC
+   "arch-x86_32" {os = "win32"} &
+    ("system-mingw" & "mingw-w64-shims" {os-distribution = "cygwin" & post} |
+     "system-msvc") |
+  # Non-Windows systems
+   "host-system-other" {os != "win32" & post})
```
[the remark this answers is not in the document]
dra27:
```suggestion
  (("arch-x86_64" {os = "win32" & arch = "x86_64"} &
     (("system-mingw" & "mingw-w64-shims" {os-distribution = "cygwin" & post}) |
      "system-msvc")) |
  # i686 mingw-w64 / MSVC
   ("arch-x86_32" {os = "win32"} &
     (("system-mingw" & "mingw-w64-shims" {os-distribution = "cygwin" & post}) |
      "system-msvc")) |
  # Non-Windows systems
   "host-system-other" {os != "win32" & post})
```

I'm not sure it's clearer, but I don't mind either way - what do you think?

packages/host-arch-riscv64/host-arch-riscv64.1/opam
raphael-proust: I want to strive for regularity so I think the conflict classes should be as follows:

- `msys2-env` (consistent use of `env` with other classes)
- `ocaml-arch` (consistent use of `arch` as per package name, `setting` doesn't appear in package name) & `ocaml-system`
- `ocaml-host-arch` & `ocaml-host-system`
- `ocaml-mingw-env` & `ocaml-msvc-env`

Moreover, would it make sense for the `ocaml-{mingw,msvc}{32,64}-env` to be named `ocaml-{mingw,msvc}-env-{32,64}` instead? The arch packages have the prefix set and then afterwards the variant. Whereas the mingw/msvc env packages have the variant in the middle of the prefix.

packages/conf-c++/conf-c++.1.0/opam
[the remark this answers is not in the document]
raphael-proust: thanks for the explanations

packages/ocaml-base-compiler/ocaml-base-compiler.4.13.0/opam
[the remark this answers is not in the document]
raphael-proust: It's easier to navigate in a text editor (what with pairing) so I'm in favour of this change.

packages/host-arch-riscv64/host-arch-riscv64.1/opam
dra27: `conflict-class`s all updated. I like the switch to `ocaml-env-...` but I've stuck with the existing set of `mingw32`, `mingw64`, `msvc32` and `msvc64` principally because these serve the same purpose as the MSYS2 Environments. i.e. `ocaml-env-mingw64` is performing the Cygwin-related function that `msys2-mingw64` performs.

packages/ocaml-base-compiler/ocaml-base-compiler.4.13.0/opam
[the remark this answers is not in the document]
dra27: Done!

packages/host-arch-riscv64/host-arch-riscv64.1/opam
raphael-proust: the conflict classes are good as they currently stand, thanks

raphael-proust:
hmm

There are actually some weird results like successes for installing all the arch package in some default runners. Anyway, I still think that teh analysis phase needs to be updated.

--------------------------

This MR can be merged before we update the analysis phase bc those arch/system packages shouldn't be modified too often.

packages/host-arch-arm64/host-arch-arm64.1/opam
```diff
@@ -0,0 +1,16 @@
...
+
+This package may be used in depends or conflicts fields of dependent packages
+to indicate either a requirement or an incompatibility with this
+architecture."""
+maintainer: "David Allsopp <david@tarides.com>"
+authors: "David Allsopp"
+license: "CC0-1.0+"
+homepage: "https://opam.ocaml.org"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+conflict-class: "ocaml-host-arch"
```
raphael-proust, line 16: These new packages should have `available` fields to avoid the CI attempting to install them.

raphael-proust: ping @dra27 this can be done later in another PR, I'll let you manage your own TODO list

raphael-proust: FYI this is the errors in the CI

```
extras
    arm32-ocaml-4.14
        host-arch-arm32.1 (failed: Failed: Build failed)
        host-arch-x86_64.1 (failed: Failed: Build failed)
        host-system-mingw.1 (failed: Failed: Build failed)
        host-system-other.1 (failed: Failed: Build failed)
        msys2-clang64.1 (failed: Failed: Build failed)
        ocaml-base-compiler.4.13.0 (failed: Failed: Build failed)
        ocaml-config.0 (failed: Failed: Build failed)
        ocaml-env-msvc64.1 (failed: Failed: Build failed)
        ocaml-system.4.13.0 (failed: Failed: Build failed)
        ocaml-variants.4.13.2+trunk (failed: Failed: Build failed)
        ocaml.4.14.0 (failed: Failed: Build failed)
        system-mingw.1 (failed: Failed: Build failed)
    arm32-ocaml-5.2
        host-arch-s390x.1 (failed: Failed: Build failed)
        host-arch-x86_64.1 (failed: Failed: Build failed)
        host-system-mingw.1 (failed: Failed: Build failed)
        host-system-msvc.1 (failed: Failed: Build failed)
        msys2-mingw64.1 (failed: Failed: Build failed)
        ocaml-base-compiler.4.14.1 (failed: Failed: Build failed)
        ocaml-base-compiler.5.2.0~alpha1 (failed: Failed: Build failed)
        ocaml-option-mingw.1 (failed: Failed: Build failed)
        ocaml-options-only-no-flat-float-array.1 (failed: Failed: Build failed)
        ocaml-system.5.1.1 (failed: Failed: Build failed)
        ocaml-variants.4.14.3+trunk (failed: Failed: Build failed)
        ocaml.4.13.0 (failed: Failed: Build failed)
    arm64-ocaml-4.14
        conf-g++.1.0 (failed: Failed: Build failed)
        msys2-clang64.1 (failed: Failed: Build failed)
        msys2.0.1.0 (failed: Failed: Build failed)
        ocaml-base-compiler.5.2.0 (failed: Failed: Build failed)
        ocaml-base-compiler.5.2.0~beta2 (failed: Failed: Build failed)
        ocaml.4.14.0 (failed: Failed: Build failed)
        ocaml.4.14.3 (failed: Failed: Build failed)
    arm64-ocaml-5.2
        host-arch-x86_64.1 (failed: Failed: Build failed)
        host-system-msvc.1 (failed: Failed: Build failed)
        msys2-clang64.1 (failed: Failed: Build failed)
        msys2-clangarm64.1 (failed: Failed: Build failed)
        msys2-mingw32.1 (failed: Failed: Build failed)
        system-mingw.1 (failed: Failed: Build failed)
        system-msvc.1 (failed: Failed: Build failed)
```

dra27: What `available` field should they have? No other base packages have available fields?

dra27: Although I can see they might temporarily want to conflict older compilers which haven't been updated to use them - I will indeed put that on the TODO list

dra27:
Thank you for the reviews and for the merge, @raphael-proust!

avsm:
Congratulations on the merge, @dra27!  A huge amount of work from you on this over many years.

nojb:
Congratulations @dra27 (and everyone else who helped)!

MSoegtropIMC:
@dra27 : can it be that this breaks OCaml on MacOS. I updated my repo and now get for e.g. 4.14.2:

Switch invariant: ["ocaml-variants" {= "4.14.2+options"} "ocaml-option-flambda"]
[ERROR] Could not determine which packages to install for this switch:
  * Missing dependency:
    - ocaml-option-32bit
    unmet availability conditions: 'arch = "x86_64" & (os = "linux" | os = "macos")'

I am installing opam and switches with scripts I didn't change, so I guess it is a result of this PR.

MSoegtropIMC:
I am on on ARM Mac, btw, so no idea how I get to x86_64 - my opam is clearly arm.

packages/host-arch-arm64/host-arch-arm64.1/opam
raphael-proust: They should have `available` fields matching their architecture. E.g., `host-arch-s390x` should have `available: [ arch = "s390x" ]`. This will avoid the CI trying to build the package on, say, arm32.

MSoegtropIMC:
Update: I rolled back to the previous version of ocaml-variants.4.12.2+options and this works.

dra27:
Sorry about this, it turns out it's a silly mistake in this conjunction:

```
 (("host-arch-x86_32" {os != "win32" & arch = "x86_64" & post} & "ocaml-option-32bit" {os != "win32"}) |
   "host-arch-x86_64" {os != "win32" & arch = "x86_64" & post})
```

For `arch != "x86_64"` and `os != "win32"` this is reducing to the single atom `"ocaml-option-32bit"` which is of course failing, as that package is not available on arm64.

I have already tested a fixing commit (https://github.com/dra27/opam-repository/commit/aef43ad8c6df0060c8619ab7e9d857d46a445651), but as there's a related issue with ocaml-ci and this disjunction, I'm just double-checking what needs to happen (that commit will definitely be in a PR by the end of today, no matter what).

PR #25863, opened 2024-05-11
conf-cuda: CUDA libraries -- configure installation path, offer install tips, validate

lukstafi:
The package `conf-cuda-config` collects information, and then the package `conf-cuda` verifies that the CUDA library headers are installed. `conf-cuda` provides instructions to the user on how to install the CUDA headers and library objects, if the headers are missing. NVidia drivers are out of scope for `conf-cuda`.

### DIFF
--- a/packages/conf-cuda-config/conf-cuda-config.1/opam
+++ b/packages/conf-cuda-config/conf-cuda-config.1/opam
@@ -10,15 +10,24 @@ build: [
     if [ -z "$CUDA_PATH" ]; then
       if [ -d "/usr/local/cuda" ]; then
         CUDA_PATH="/usr/local/cuda"
-        HAS_CUDA="true"
+        CUDA_PREINSTALLED="true"
       elif [ -d "/opt/cuda" ]; then
         CUDA_PATH="/opt/cuda"
-        HAS_CUDA="true"
+        CUDA_PREINSTALLED="true"
+      elif [ -d "/usr/cuda" ]; then
+        CUDA_PATH="/usr/cuda"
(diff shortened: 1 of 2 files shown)


lukstafi:
Request for help: `depexts:` are ignored! I'm stuck. From the CI, both Ubuntu and ArchLinux targets should be passing. Failures of other targets are as expected.

lukstafi:
Since `depexts:` cannot make informed decisions, all CI targets are expected to fail. Particular targets (Debian, Ubuntu, ArchLinux) will provide direct installation instructions, others will point to the NVidia CUDA webpages.

mseri:
Can you open an issue on ocaml/opam to request the availability of flags for depexts?

lukstafi:
> Can you open an issue on ocaml/opam to request the availability of flags for depexts?

As Kate mentioned, there's already a PR about it but it's controversial.
https://github.com/ocaml/opam/pull/5455

lukstafi:
I'll also note that the CUDA system packages (Ubuntu, OpenSUSE, Arch Linux) we can depend on are very heavy so would put a strain on the CI.

mseri:
In my opinion the conf-cuda packages being heavy is not a problem, I would add the ones that can be added. If it turns out to nbe problematic, it will be a good way to tweak the CI

lukstafi:
Friendly ping. What's the "question"?

mseri:
Thanks for the ping. I think we can merge it as is, we can revise later on the choice of not adding depexts for the few systems that provide the packages already

PR #25864, opened 2024-05-12
[new release] bam (3 packages) (0.2)

francoisthire:
A property-based testing library with internal shrinking

- Project page: <a href="https://github.com/francoisthire/bam">https://github.com/francoisthire/bam</a>
- Documentation: <a href="https://francoisthire.github.io/bam/">https://francoisthire.github.io/bam/</a>

##### CHANGES:

- francoisthire/bam#2,francoisthire/bam#5: Add a PPX for Bam derving generators from a type or a module type

- francoisthire/bam#1: Fix the `Bam.Std.oneof` generator

### DIFF
--- a/packages/bam-ppx/bam-ppx.0.2/opam
+++ b/packages/bam-ppx/bam-ppx.0.2/opam
@@ -11,7 +11,7 @@ doc: "https://francoisthire.github.io/bam/"
 bug-reports: "https://github.com/francoisthire/bam/issues"
 depends: [
   "ocaml" {>= "4.14" & < "5.2"}
-  "dune" {>= "3.7" & >= "3.7"}
+  "dune" {>= "3.7"}
   "ppxlib" {>= "0.32.0"}
   "dmap" {>= "0.5"}
   "odoc" {with-doc}
--- a/packages/bam-ppx/bam-ppx.0.2/opam
+++ b/packages/bam-ppx/bam-ppx.0.2/opam
@@ -1,0 +1,42 @@
+opam-version: "2.0"
+synopsis: "A PPX deriving generators for OCaml types"
+description:
+  "Provides a way to automatically get generators for a given type"
+maintainer: ["François Thiré"]
+authors: ["François Thiré"]
+license: "MIT"
+tags: ["test" "pbt" "shrinking" "internal" "bam" "ppx"]
+homepage: "https://github.com/francoisthire/bam"
+doc: "https://francoisthire.github.io/bam/"
+bug-reports: "https://github.com/francoisthire/bam/issues"
+depends: [
+  "ocaml" {>= "4.14" & < "5.2"}
+  "dune" {>= "3.7" & >= "3.7"}
+  "ppxlib" {>= "0.32.0"}
+  "dmap" {>= "0.5"}
+  "odoc" {with-doc}
+]
+build: [
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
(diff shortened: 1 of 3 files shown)


francoisthire:
The current issue seems to concern `32` bits architectures. I am not sure where it comes from. I am using `dune` to generate the opam files. I did not find a way to specify the architecture via the `dune-project` file.

raphael-proust:
Not all fields of the opam file are available in dune. The missing ones can be added via the use of a separate file. See https://dune.readthedocs.io/en/stable/reference/packages.html in which you find

> If a `pkg.opam.template` file is found, the fields found in this opam file are added to the generated file, overriding the generated ones in case of collision. This can be used as an escape hatch to add arbitrary opam fields.

So I think you need this template file to contain
```
conflicts: [
  "ocaml-option-bytecode-only"
]
```

packages/tezt-bam/tezt-bam.0.2/opam
```diff
@@ -0,0 +1,43 @@
...
+bug-reports: "https://github.com/francoisthire/bam/issues"
+depends: [
+  "ocaml"
+  "dune" {>= "3.7" & >= "3.7"}
+  "tezt" {>= "4.0"}
+  "bam"
+  "bam-ppx"
+  "mtime" {>= "2.0"}
+  "odoc" {with-doc}
+]
```
raphael-proust, line 19:
```suggestion
]
conflicts: [
  "ocaml-option-bytecode-only"
]
```

We can publish the package with this (if the CI passes after it's applied). You should still use the template thing upstream to avoid having to add this in future releases.

raphael-proust: hmm I might have gotten the specific option wrong. It's a 32 bit issue.

raphael-proust:
The actual addition you want is probably

```
available: arch != "arm32" & arch != "x86_32"
```

Sorry for the confusion, I copied the wrong fix.

packages/tezt-bam/tezt-bam.0.2/opam
```diff
@@ -0,0 +1,45 @@
...
+maintainer: ["François Thiré"]
+authors: ["François Thiré"]
+license: "MIT"
+tags: ["test" "tezt" "pbt" "shrinking" "internal" "bam"]
+homepage: "https://github.com/francoisthire/bam"
+doc: "https://francoisthire.github.io/bam/"
+bug-reports: "https://github.com/francoisthire/bam/issues"
+depends: [
+  "ocaml"
+  "dune" {>= "3.7" & >= "3.7"}
```
mseri, line 13:
```suggestion
  "dune" {>= "3.7"}
```
This is fixed on the repo itself by removing `dune` from the dependency list in the dune-project file (it is added automatically)

packages/bam-ppx/bam-ppx.0.2/opam
```diff
@@ -0,0 +1,42 @@
...
+maintainer: ["François Thiré"]
+authors: ["François Thiré"]
+license: "MIT"
+tags: ["test" "pbt" "shrinking" "internal" "bam" "ppx"]
+homepage: "https://github.com/francoisthire/bam"
+doc: "https://francoisthire.github.io/bam/"
+bug-reports: "https://github.com/francoisthire/bam/issues"
+depends: [
+  "ocaml" {>= "4.14" & < "5.2"}
+  "dune" {>= "3.7" & >= "3.7"}
```
mseri, line 14:
```suggestion
  "dune" {>= "3.7"}
```
This is fixed on the repo itself by removing `dune` from the dependency list in the dune-project file (it is added automatically)

francoisthire:
The CI was green before, and not it is pending/with failurs. Is it expected?

mseri:
The issues are on the CI now, this looks good. Thanks

PR #25874, opened 2024-05-14
zarith*: use sha256 instead of md5 for extra-files checksums

hannesm:
Dear opam-repository maintainers,

I'm keen to get rid of md5 checksums in the opam-repository. I hope this is a shared goal with you as well.

To keep the ball rolling, I submitted a new lint for opam-repo-ci https://github.com/ocurrent/opam-repo-ci/pull/304 which yells if only MD5 is used as checksum -- so no new md5 checksums should be accepted in this repository.

A second step is to convert the existing md5 checksums to sha256. I started with `opam admin update-extrafiles --hash=sha256` (from opam 2.2). Here is a tiny output, I can as well push a PR with all the changes (though some should be avoided where the existing checksum is sha512). Please let me know what you think about the roadmap and this specific PR. If you have other plans, we can also drop this PR and take a different ship.

Thanks,

Hannes

### DIFF
--- a/packages/zarith-freestanding/zarith-freestanding.1.10/opam
+++ b/packages/zarith-freestanding/zarith-freestanding.1.10/opam
@@ -23,10 +23,22 @@ arbitrary-precision integers. It uses GMP to efficiently implement
 arithmetic over big integers. Small integers are represented as Caml
 unboxed integers, for speed and space economy."""
 extra-files: [
-  ["mirage-uninstall.sh" "md5=33f0c41ca3ef941f73fc566977b32453"]
-  ["mirage-install.sh" "md5=4eb793d6246c80eacebb201694cf3a3e"]
-  ["mirage-build.sh" "md5=ccca36c5e3632bbb788bce15b4487f77"]
-  ["no-dynlink.patch" "md5=2437531740d788bbeba56241d54debc2"]
+  [
+  "mirage-build.sh"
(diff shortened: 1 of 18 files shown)


mseri:
I like the idea. The linter seems to have been merged and is polluting all the CI logs. If you make the large PR and we merge it immediately (we should coordinate on this), would this force useless recompilations to all users? If not let's do it asap, if yes, let's do it as soon as we merge a dune release or a JS release, which will force recompilations in many switches anyway. @raphael-proust what do you think?

hannesm:
Ok, having read a bit more, I have the following suggestion for a change:
- extra-files isn't an optimal thing (see CONTRIBUTING.md), so let's do once a change to push all these extra files in a separate git repository and use extra-sources instead (then using sha256 / sha512) --> here we can as well introduce another lint that extra-files shouldn't be used when submitting new packages
- the checksum issue is also the case for url, so in the same thing we move these checksums to sha256

I can work on scripts and branches for these changes this week, so in the case there's no pushback we can apply this change to opam-repository next week.

For migrating extra-files to extra-sources, some git repository that is maintained by the opam maintainers would be very welcome. maybe some new one in the ocaml or ocaml-opam organisation (if anyone can create such a thing, please do :))?

I'm also happy to write an announcement once this change has been done about the motivation of this change.

hannesm:
let me close this, and open an issue in this repository.

PR #25873, opened 2024-05-14
Package why3find.1.0

btjorge:
### `why3find.1.0`
A Why3 Package Manager
The why3find utility is designed for managing packages for why3 developpers and associated OCaml extracted code.

---
* Homepage: https://git.frama-c.com/pub/why3find
* Source repo: git+https://git.frama-c.com:pub/why3find.git
* Bug tracker: https://git.frama-c.com/pub/why3find/issues

---
:camel: Pull-request generated by opam-publish v2.3.0

PR #25866, opened 2024-05-13
[new release] alt-ergo (4 packages) (2.5.4)

bclement-ocp:
The Alt-Ergo SMT prover

- Project page: <a href="https://alt-ergo.ocamlpro.com/">https://alt-ergo.ocamlpro.com/</a>
- Documentation: <a href="https://ocamlpro.github.io/alt-ergo">https://ocamlpro.github.io/alt-ergo</a>

### Bug fixes

* Fix a long-standing soundness issue in the FPA module (OCamlPro/alt-ergo#1122, originally
  reported in OCamlPro/alt-ergo#1111)
* Dolmen frontend incorrectly allowed semantic triggers outside of theory
  extensions (OCamlPro/alt-ergo#1122)

PR #25878, opened 2024-05-15
alt-ergo build requires native compiler

raphael-proust:
The following error showed up on the CI of https://github.com/ocaml/opam-repository/pull/25873

```
#=== ERROR while compiling alt-ergo.2.4.2 =====================================#
# context              2.2.0~beta3~dev | linux/arm32 | ocaml-base-compiler.5.2.0 | file:///home/opam/opam-repository
# path                 ~/.opam/5.2/.opam-switch/build/alt-ergo.2.4.2
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p alt-ergo -j 79
# exit-code            1
# env-file             ~/.opam/log/alt-ergo-6-c6418a.env
# output-file          ~/.opam/log/alt-ergo-6-c6418a.out
### output ###
# File "src/plugins/fm-simplex/dune", line 17, characters 5-25:
# 17 |     (FmSimplexPlugin.cmxs as plugins/fm-simplex-plugin.cmxs)
#           ^^^^^^^^^^^^^^^^^^^^
# Error: No rule found for src/plugins/fm-simplex/FmSimplexPlugin.cmxs
# File "src/plugins/AB-Why3/dune", line 24, characters 5-22:
# 24 |     (ABWhy3Plugin.cmxs as plugins/AB-Why3-plugin.cmxs)
#           ^^^^^^^^^^^^^^^^^
# Error: No rule found for src/plugins/AB-Why3/ABWhy3Plugin.cmxs
```

IIRC, cmxs files are produced by ocamlopt and so it seems that the build is incompatible with bytecode-only installs.

It's likely that other versions of alt-ergo should get the same change. I'll try to investigate which ones or maybe we can just apply a blanket change? (ping @Stevendeo ?)

### DIFF
--- a/packages/alt-ergo/alt-ergo.2.4.2/opam
+++ b/packages/alt-ergo/alt-ergo.2.4.2/opam
@@ -33,6 +33,9 @@ build: [
   ["dune" "subst"] {dev}
   ["dune" "build" "-p" name "-j" jobs]
 ]
+conflicts: [
+  "ocaml-option-bytecode-only"
+]
 url {
   src: "https://github.com/OCamlPro/alt-ergo/archive/refs/tags/2.4.2.tar.gz"
   checksum: [


raphael-proust:
The new release is already mark incompatible with bytecode-only: https://github.com/ocaml/opam-repository/pull/25866

bclement-ocp:
Applying a blanket change to all 2.X.Y releases should be fine, older releases may work with bytecode-only (but we really don't recommend using a bytecode version of Alt-Ergo).

FWIW the hard dependency on .cmxs will be removed in the upcoming 2.6 release.

mseri:
@raphael-proust do you plan to apply the change to also other packages or we can merge?

PR #25881, opened 2024-05-15
Frama-C: new release (29.0~beta-Copper)

tmarti2:
Main changes with respect to Frama-C 28 include:

# Kernel
- Introduce \plugin:: prefix for ACSL extensions
- Refactor current location handling mechanism
- Removal of Db (Db.Main.extend is deprecated). Features related to asynchronous
  interactions are now handled in module Async
- Various fixes and improvements

# Alias
- Better analysis results in the presence of structures
- Rework the API and improved the documentation

# Eva
- Better reporting for garbled mix
- Improved handling of \valid and \valid_read

# E-ACSL
- Fix TLS segment start address and size
- Remove option -e-acsl-version

# WP
- Generation of counter examples, see option -wp-counter-examples
- Upgrade to Why3 1.7.x and improved prover selection
- Extended support for Ivette

# Ivette
- Revamped workspace (tabs, views, dock, alerts, …)
- Types and Globals navigation
- Better feedback on Eva values evaluation
- Extended support for WP

### DIFF
--- a/packages/frama-c/frama-c.29.0~beta/opam
+++ b/packages/frama-c/frama-c.29.0~beta/opam
@@ -1,0 +1,197 @@
+opam-version: "2.0"
+available: opam-version >= "2.1.0"
+flags: avoid-version
+synopsis: "Platform dedicated to the analysis of source code written in C"
+description:"""
+Frama-C gathers several analysis techniques in a single collaborative
+framework, based on analyzers (called "plug-ins") that can build upon the
+results computed by other analyzers in the framework.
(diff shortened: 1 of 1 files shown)


maroneze:
The CI failures are:
- test oracles in FreeBSD (we'll probably disable the tests in the final release);
- some failures related to alt-ergo 2.5.1, which seem unrelated to Frama-C.

In any case, this is a beta release, so if possible we'd like this to be merged so that our users can start beta-testing it. If you prefer that we disable the tests on FreeBSD nevertheless, we'll do so and publish it again.

PR #25914, opened 2024-05-21
conf-unwind: Add support for fedora-base distributions

kit-ty-kate:


mseri:
Do we need to do this on all conf- packages?

packages/conf-unwind/conf-unwind.0/opam
```diff
@@ -10,7 +10,7 @@ depexts: [
   ["libunwind-dev"] {os-family = "debian" | os-family = "ubuntu"}
   ["libunwind-devel"] {os-distribution = "centos"}
   ["libunwind-devel"] {os-distribution = "rhel"}
-  ["libunwind-devel"] {os-distribution = "fedora"}
+  ["libunwind-devel"] {os-distribution = "fedora" | os-family = "fedora"}
```
mseri, line 13:
```suggestion
  ["libunwind-devel"] {os-family = "fedora"}
```
Why not this? Can it have fedora distribution but not family?

kit-ty-kate: mmh, good question, I thought Fedora itself had `ID_LIKE = rhel` or something but it turns out they do not (maybe they had in the past?). So I guess this should be safe to do.

mseri: I'll keep it as is, can be revisited in the future

PR #25916, opened 2024-05-21
ocamlnet does not work on OCaml 5 so far

Leonidas-from-XIV:
ocamlnet does not support OCaml 5 at the moment. There is a [PR on the repo](https://gitlab.com/gerdstolpmann/lib-ocamlnet3/-/merge_requests/21) but so far it is not merged.

Thus I used opam admin add-constraint to constrain the existing versions to `< 5` where necessary. Some conditions have been thus reshuffled.


mseri:
Thanks

PR #25918, opened 2024-05-22
[new release] ppx_deriving 6.0.2

NathanReb:


NathanReb:
Let's hope this one works! I opened a draft this time to test out the revdeps builds and added upper bounds to `pa_ppx` as mentioned [here](https://github.com/ocaml/opam-repository/pull/25720#issuecomment-2124286072)

NathanReb:
There's a lot of request time out in the CI. Outside of those I couldn't spot any errors related to `ppx_deriving` but I might have missed something. @mseri does anything seem fishy to you here?

mseri:
I fixed the merge conflict. I'll let the CI re-run, hopefully we will see less timeouts. I think it's fine to merge

NathanReb:
Awesome, thanks! I'll go through with the actual release and undraft!

mseri:
I think this release or one of its dependencies has dropped a dependency on result. I'll send a PR to fix all the broken stuff before merging, sorry for the extra delay this will add

mseri:
Bounds added in https://github.com/ocaml/opam-repository/pull/25951

I think this are safe to add, since they were using result in their code but taking it as transitive dependence from ppx_deriving

NathanReb:
Wouldn't the fix be to add the result dependency instead? That way they could still build with ppx_deriving.

No worries for the delay, I forgot to come back and check if they were new failures so that's partly on me.

NathanReb:
Ah yes you're right! My bad!

PR #25939, opened 2024-05-26
ocamlbuild for windows OCaml 5.2

hhugo:
This PR provides a windows variant of ocamlbuild.0.14.3.
It re-uses the patch used to fix ocamlbuild.0.14.2 for windows and make ocamlbuild.0.14.3 available on windows with OCaml 5.02.

### DIFF
--- a/packages/ocamlbuild/ocamlbuild.0.14.3+win/opam
+++ b/packages/ocamlbuild/ocamlbuild.0.14.3+win/opam
@@ -1,0 +1,38 @@
+opam-version: "2.0"
+synopsis:
+  "OCamlbuild is a build system with builtin rules to easily build most OCaml projects"
+maintainer: "Gabriel Scherer <gabriel.scherer@gmail.com>"
+authors: ["Nicolas Pouillard" "Berke Durak"]
+license: "LGPL-2.0-or-later WITH OCaml-LGPL-linking-exception"
+homepage: "https://github.com/ocaml/ocamlbuild/"
+doc: "https://github.com/ocaml/ocamlbuild/blob/master/manual/manual.adoc"
+bug-reports: "https://github.com/ocaml/ocamlbuild/issues"
+depends: [
+  "ocaml" {>= "4.03"}
+]
+conflicts: [
+  "base-ocamlbuild"
+  "ocamlfind" {< "1.6.2"}
+]
+build: [
+  [make "all"]
+]
+install: [
+  [make "install"]
+  ["mkdir" "-p" "%{lib}%/ocamlbuild"]
+  ["install" "-m" "0644" "META" "%{lib}%/ocamlbuild"]
+]
+dev-repo: "git+https://github.com/ocaml/ocamlbuild.git"
+url {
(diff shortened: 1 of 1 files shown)


hhugo:
cc @kit-ty-kate @gasche

hhugo:
I've tested this in https://github.com/ocsigen/js_of_ocaml/pull/1615

kit-ty-kate (changes requested):
I'd rather avoid having a separate package for this.
I think we should either have the patch applied conditionally on Windows in the 0.14.3 package or wait a couple weeks to have an upstream version (this is next on my todo-list after the release of opam 2.2.0~rc1)

hhugo:
> I'd rather avoid having a separate package for this. I think we should either have the patch applied conditionally on Windows in the 0.14.3 package or wait a couple weeks to have an upstream version (this is next on my todo-list after the release of opam 2.2.0~rc1)

I don't think we should wait before making ocamlbuild available on windows 5.2. There are enough issues on the windows side already.

We could make it part of the main opam file. I don't know the rational for splitting the files initially. I just kept the setup in place to unlock the situation as fast as possible.

hhugo:
I think it would be enough to merge this into https://github.com/ocaml-opam/opam-repository-mingw

Maybe we should move ocamlbuild.0.14.2+win as well. What do you think ?

kit-ty-kate:
> Maybe we should move ocamlbuild.0.14.2+win as well. What do you think ?

Packages can't be removed until https://github.com/ocaml/opam-repository/issues/23789 is implemented/formally accepted.

But overall I think it would be fine as part of the main opam file. I really don't understand why it was split in the first place.
It would just be a case of doing:
```
patches: [
  "ocamlbuild-0.14.2.patch" {os = "win32"}
]
```

hhugo:
> > Maybe we should move ocamlbuild.0.14.2+win as well. What do you think ?
> 
> Packages can't be removed until #23789 is implemented/formally accepted.
> 
> But overall I think it would be fine as part of the main opam file. I really don't understand why it was split in the first place. It would just be a case of doing:
> 
> ```
> patches: [
>   "ocamlbuild-0.14.2.patch" {os = "win32"}
> ]
> ```

I guess the `+win` variant makes it clear that it's a modified version. There are a fair amount of changes in this patch. Changes that have never been reviewed or proposed upstream.

hhugo:
That said, if you prefer to merge the patch in the main opam, it's fine.

hhugo:
Can we unlock the situation on windows and merge this ?

mseri:
Done. Is there any chance this patch could be reviewed and merged into upstream ocamlbuild any time soon? Windows support shouldn't be an ad-hoc opam-repo thing imho

gasche:
The problem is that this patch is a hodge-podge of stuff that was only ever tested on Windows, so it feels very unsafe (to me) to merge as-is for everyone. Just splitting this into proper patches and understanding the intent of each change is a substantial amount of work. In principle I would be delighted to see this work being done, but in practice few people have the energy to do it. @hhugo has done some of it in the last few weeks, and we try to make his contributions as pleasant as possible to keep him motivated to do more :-)

PR #25943, opened 2024-05-27
provider 0.0.4: allow base & etc >= v0.16

mbarbin:
I wasn't aware of a workflow in use in the opam-repository by which when new js packages versions are released, bounds are added if/where needed (upon witnessing actual breaks).

In light of this, I think the dependency setup that I initially set for provider may be relaxed.

In particular, I have checked that this package is compatible with the `v0.17` suite that was recently released, and I'd like to allow this combination.

Thank you!


mseri:
Thanks, indeed we can add the upper bounds before a package that breaks it is released. Usually the CI is able to detect these

PR #25952, opened 2024-05-28
[Metadata update] Package niru

EruEri:
- Fix a typo in the description 
- Change source code location but keep the same checksum


mseri:
I think it would be inaccessible for most people. Can you open an issue on the repo, so that we discuss it as a policy?

EruEri:
Yes, it's created

PR #25972, opened 2024-05-30
Expand conf-nauty platform support

jmid:
This PR expands conf-nauty support to include Alpine, Ubuntu, Open/Suse, macOS/homebrew, and FreeBSD.

Again, I couldn't find any mention of it on the relevant Oracle 7/8/9 repos linked from https://yum.oracle.com/index.html and hence mark it an acceptable CI failure.

### DIFF
--- a/packages/conf-nauty/conf-nauty.1.0/opam
+++ b/packages/conf-nauty/conf-nauty.1.0/opam
@@ -15,6 +15,11 @@ depexts: [
   ["nauty-devel"] {os-family = "suse" | os-family = "opensuse"}
   ["nauty"] {os = "freebsd"}
 ]
+x-ci-accept-failures: [
+  "oraclelinux-7"
+  "oraclelinux-8"
+  "oraclelinux-9"
+]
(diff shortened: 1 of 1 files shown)


jmid:
CI fails on
- Arch (missing a case in the opam file) and
- OpenSuse (unsure why)

jmid:
I think OpenSuse is failing due to a missing .pc file, causing the pkg-config lookup to fail
See `Files` here for package contents - without a `nauty.pc`:
https://opensuse.pkgs.org/15.5/opensuse-oss-x86_64/nauty-devel-2.8.6-bp155.1.7.x86_64.rpm.html

jmid:
I can see a package here https://archlinux.org/packages/extra/x86_64/nauty/ from the "extra" repo.
I'm unsure whether that repo is supported in opam-ci though... :thinking:

jmid:
I've now added `x-ci-accept-failures` for OpenSuse and Arch.

mseri:
Thanks

PR #25976, opened 2024-05-30
Expand conf-numa platform support

jmid:
This PR expands conf-numa support to include Ubuntu, Oracle, and FreeBSD.

Because of the relatively loose `build` check (ha!) I expect this to be all green...


mseri:
The failure is independent of this. Thanks

PR #25978, opened 2024-05-30
Expand conf-ode platform support

jmid:
This adds conf-ode support for Ubuntu, Open/Suse, Arch, Fedora, macOS-homebrew, and FreeBSD.

Alpine is under way it seems: https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/60122


jmid:
CI is all green, modulo an lint package name collision warning

Issue #25969, opened 2024-05-29
policy question: existing packages without checksums


hannesm:
Dear everyone,

there's a bunch of packages without checksum. I wanted to ensure everyone is on the same page.

My goal is to discuss what to do with the packages that are not checked (i.e. that are available etc. in the opam-repository). I put them into categories and suggested what I'd do. If you agree, let me know - I'm happy to prepare a PR with the suggestions I noted.

## arbitrary packages (my suggestion: mark as `available: false` - each package has a newer version available in opam-repository)
see https://github.com/ocaml/opam-repository/pull/25981
- [x] sedlex.1.99.1 (src: "git+https://github.com/ocaml-community/sedlex.git")
- [x] ppx_tools.0.1 (src: "git+https://github.com/alainfrisch/ppx_tools.git")
- [x] ppx_test.1.5.1 (available: false)
- [x] base.v0.15.1~5.0preview (flags: avoid-version)
- [x] ocamlformat.0.19.0~4.13preview (available: false)
- [x] ppxlib.0.25.0~5.00preview (available: false)
- [x] camltc.0.9.3 (src: "git+https://github.com/toolslive/camltc.git#camltc-0.9.3")
- [x] camltc.0.9.4 (src: "git+https://github.com/toolslive/camltc.git#317e1e4606ce1e86d09b7b83ec2cb1709a81d437")
- [x] camltc.0.9.5 (src: "git+https://github.com/toolslive/camltc.git#camltc-0.9.5")
- [x] camltc.0.9.6 (src: "git+https://github.com/toolslive/camltc.git#camltc-0.9.6")
- [x] quickcheck.0.0.3 (src: "git+https://github.com/avsm/ocaml-quickcheck.git")

## merlin and lsp (inconsistent, there are further `preview` versions with `available: false`), suggestion: mark the remaining two with `available: false`
- [x] ocaml-lsp-server.1.9.0~4.13preview (available: false)
- [x] ocaml-lsp-server.1.9.2~4.14preview (available: false)
- [x] ocaml-lsp-server.1.13.2~5.0preview (flags: avoid-version)
- [x] dot-merlin-reader.4.4~5.0.preview (src: "git+https://github.com/kit-ty-kate/merlin.git#500")
- [x] merlin-lib.4.4~5.0.preview (src: "git+https://github.com/kit-ty-kate/merlin.git#500")
- [x] merlin.4.3.2~4.13preview (available: false)
[...]

PR #25981, opened 2024-05-30
mark some packages unavailable: use git branches and no checksum

hannesm:
there are newer versions of all the packages available from this repository

discussion in #25969 //cc @mseri


mseri:
Thanks

PR #25983, opened 2024-05-30
Expand conf-pam platform support

jmid:
This adds conf-pam support for Ubuntu-family, Open/Suse, Oracle, and Arch


jmid:
All green, modulo linting warnings.

PR #25984, opened 2024-05-31
Expand conf-pandoc platform support

jmid:
This adds conf-pandoc support for Ubuntu-family and FreeBSD


jmid:
All green, modulo `(lint) (failed: Warning in conf-pandoc.0.1: Possible name collision with package 'conf-pango')` 

Should be good to go.

shonfeder:
Yep looks great. Thank you for this useful updatet1

PR #25985, opened 2024-05-31
Expand conf-pango platform support

jmid:
This adds conf-pango support for Ubuntu-family and Oracle


jmid:
This is all green, modulo a lint warning and should be good to go :tada: 

IMO, the name clash warnings add needless noise forcing readers to have to check something which should have just had a green checkmark...

shonfeder:
> IMO, the name clash warnings add needless noise forcing readers to have to check something which should have just had a green checkmark

I agree.

PR #25990, opened 2024-05-31
conf-python-3-7: add Ubuntu-family support

jmid:


jmid:
This is failing CI spectacularly, despite I've only changed a platform not tested on the CI:
- Arch, Fedora and macOS are failing because they install Python 3.12, but the test covers only 3.7-3.11
- opensuse-15.5 is failing because python3 installs Python 3.6.15 there - so less than 3.7
- Oracle fails with `Error: Unable to find a match: python38`
- FreeBSD fails with `pkg: No packages available to install matching 'lang/python37' have been found in the repositories`

jmid:
The latter 3 issues should be handled by 4fdd349, 34c2625, and d340759.

Fixing the former requires changes to the https://raw.githubusercontent.com/ocaml/opam-source-archives/main/patches/conf-python-3-7/configure.sh - which is now located elsewhere.
What's the policy regarding a change to that (along the lines of the following)?
```diff
- for version in 3.11 3.10 3.9 3.8 3.7; do
+ for version in 3.12 3.11 3.10 3.9 3.8 3.7; do
```

packages/conf-python-3-7/conf-python-3-7.1.0.0/opam
```diff
@@ -11,25 +11,27 @@ depends: [
...
-  ["python3"] {os-family = "debian"
+  ["python3"] {(os-family = "debian" | os-family = "ubuntu")
     & !(os-distribution = "ubuntu" & os-version < "20.04")}
   ["python3"] {os-distribution = "nixos"}
   ["python3"] {os-distribution = "alpine"}
   ["python37" "epel-release"] {os-distribution = "centos"}
   ["python3"] {os-distribution = "fedora"}
-  ["python38"] {os-distribution = "ol"} # No python37 package
+  ["python3"] {os-distribution = "ol" & os-version >= "9"}
+  ["python38"] {os-distribution = "ol" & os-version < "9"} # No python37 package
```
mseri, line 21: This is also not really python 3.7, but I guess this will install other versions in most other systems anyway

mseri:
Feel free to submit a patch to https://raw.githubusercontent.com/ocaml/opam-source-archives/main/patches/conf-python-3-7/configure.sh and then update the file and the checksum here

packages/conf-python-3-7/conf-python-3-7.1.0.0/opam
jmid: Note: the package description says
 `This package can only install if a Python interpreter >=3.7 is available on the system.`
so as long as it is greater that 3.7 we should be good.

I checked that Oracle 9 installs 3.9 https://yum.oracle.com/repo/OracleLinux/OL9/baseos/latest/x86_64/index.html
Finally, Oracle is now gone from opam-repo-ci :shrug:

jmid:
I've now added a patch PR in opam-source-archives here: https://github.com/ocaml/opam-source-archives/pull/32

shonfeder:
Marking this as pending @jmid's action, since, iiuc, a required change is pending on https://github.com/ocaml/opam-source-archives/pull/32  -- Sorry of this is in error!

jmid:
I've updated https://github.com/ocaml/opam-source-archives/pull/32 as requested last week.
Once that's merged, this PR needs updating to point to `conf-python-3-7.1/configure.sh`

mseri:
@jmid thanks! I have merged the other PR

jmid:
Modulo a curious opam lint error and risc-v runners still waiting for a worker, this is all green :tada:

mseri:
Thanks! The lint is expected, "usual" conf packages do not have sources. This does not apply to this package but it is fine

PR #25993, opened 2024-05-31
Expand conf-rocksdb platform support

jmid:
This adds conf-rocksdb platform support for Arch, Alpine, Fedora, OpenSuse, Ubuntu-family, FreeBSD, and macOS.

For OpenSuse https://pkgs.org/download/rocksdb-devel only lists an entry for openSUSE Tumbleweed. The other one will probably fail.

I also expect FreeBSD and macOS to need linker path fiddling.


jmid:
CI is failing on
- Oracle - because it is `os-family = "fedora"` and fails to install the system package `rocksdb-devel`
- FreeBSD and macOS are both failing with linking errors - both missing a link path

jmid:
5fac4e8 takes care of the first of these

PR #25995, opened 2024-05-31
conf-scdoc: add FreeBSD support

jmid:
https://www.freshports.org/textproc/scdoc/


jmid:
CI is failing on Oracle but is otherwise green. 23e362f should address it.

PR #25996, opened 2024-05-31
Expand conf-sdl2-net platform support

jmid:
This add conf-sdl2-net support for Arch, Fedora, OpenSuse, Ubuntu-family, FreeBSD, and macOS.

For OpenSuse https://pkgs.org/download/SDL2_net-devel lists some mismatch between the Tumbleweed and Leap 15.5 package names, which may require dispatching between the two... :thinking:


jmid:
All green - and good to go! :tada:

PR #25997, opened 2024-05-31
Expand conf-sdpa platform support

jmid:
This add FreeBSD and Ubuntu-family support.

I've not been able to find a system package for this on other platforms that the CI tests :thinking:


jmid:
This passes on Debian, Ubuntu, and FreeBSD as expected - and failing the rest.

PR #25998, opened 2024-05-31
conf-secp256k1: add FreeBSD and Ubuntu-family support

jmid:
https://www.freshports.org/math/secp256k1/


jmid:
Note: this is missing a `build` test despite the package description `"This package can only install if the secp256k1 lib is installed on the system."`

mseri:
Good point, let's mention it explicitly

PR #25999, opened 2024-05-31
Expand conf-tidy platform support

jmid:
This adds Alpine and Ubuntu-family support


jmid:
This is failing on
- Oracle - because a system package is missing
- FreeBSD and macOS - because an include path is missing

jmid:
b202ccd should take care of the former

PR #26001, opened 2024-05-31
Expand conf-vim platform support

jmid:
This adds Arch, Alpine, Ubuntu-family, and OpenSuse support.

Unsure how this useful this change is in practice, I suspect most if not all of these systems include it as part of their base systems :shrug:


jmid:
This fails linting but is otherwise green. It should be good to go.

mseri:
Thanks

PR #25994, opened 2024-05-31
[new release] yojson and yojson-five (2.2.0)

Leonidas-from-XIV:
Yojson is an optimized parsing and printing library for the JSON format

- Project page: <a href="https://github.com/ocaml-community/yojson">https://github.com/ocaml-community/yojson</a>
- Documentation: <a href="https://ocaml-community.github.io/yojson">https://ocaml-community.github.io/yojson</a>

##### CHANGES:

*2024-05-31*

### Added

- Added support for JSON5 (@dhilst, @gorm-issuu, @gertsonderby, ocaml-community/yojson#152)

### Removed

- Remove CPPO dependency to make the Yojson installation lighter
  (@Leonidas-from-XIV, ocaml-community/yojson#175)

PR #26003, opened 2024-05-31
Make `yojson-bench` unavailable

Leonidas-from-XIV:
There is no point in ever installing this package, it's just a hack to be able to install development dependencies and set ocaml versions (`core_bench` which includes a lot of dependencies that we don't want to have in `yojson`). In new releases like in #25994 I try not to submit the package in the first place.

Another option is to just delete it from opam-repository altogether which is probably more useful as it then doesn't clog up search results etc. but I'll refer to the maintainers.

### DIFF
--- a/packages/yojson-bench/yojson-bench.2.0.0/opam
+++ b/packages/yojson-bench/yojson-bench.2.0.0/opam
@@ -32,3 +32,4 @@ url {
   ]
 }
 x-commit-hash: "9c5cee131ae0ec79713e62397076d597486844ed"
+available: false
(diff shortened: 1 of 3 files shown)


avsm:
What's the harm in leaving it available? It's fairly common to quickly try a few benchmarks with experimental compiler variants, and it's useful to have some benchmarks lying around for that in the opam repository.

Leonidas-from-XIV:
Not much, but they're not really supported by the Yojson authors for other users and there's no guarantees that they have any use for people who are not Yojson developers. It's a bit like `stdune` which is also not meant to be used by others and has no guarantees.

If you prefer to keep them, that's fine with me.

mseri:
We are going to keep them for now, and then we will move it with the other unmaintained packages to the opam-repository archive once the policy is implemented. We have noted the package down in https://github.com/ocaml/opam-repository/wiki/Known-stale-packages-to-archive

PR #26007, opened 2024-06-02
Package cobs.0.1

Merlin04:
### `cobs.0.1`
Consistent Overhead Byte Stuffing (COBS) encoding and decoding
A lightweight Consistent Overhead Byte Stuffing (COBS) encoder and decoder written in pure OCaml with no dependencies

---
* Homepage: https://github.com/Merlin04/ocaml-cobs
* Source repo: git+https://github.com/Merlin04/ocaml-cobs.git
* Bug tracker: https://github.com/Merlin04/ocaml-cobs/issues

---


shonfeder:
Hello again :) Glad to see cobs making its way into opam in a practical form!

Could you take a look at [the ci errors](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/54c019abd05d282dbdd16735de38a0c38e653232)? Builds are failing on ocaml =< 4.12. You should be able to get compatibility with those easily with https://github.com/thierry-martinez/stdcompat or by writing your own shims or replacements of those functions. Alternatively, you could put a lower bound on supported ocaml versions.

Merlin04:
I'd be happy to take a look at those - is there a way to make the compat functions only be used on the older versions of ocaml where it's necessary (to avoid increasing size/relying on compat functions when the real ones are available)? Otherwise I'll probably just set a lower bound

shonfeder:
> is there a way to make the compat functions only be used on the older versions of ocaml where it's necessary

You can use preprocessing, as discussed in https://discuss.ocaml.org/t/ifdef-conditional-compilation-and-ocamlbuild/3244 but, if it were me, I would just implement the functions (or do without). iiuc, there are only two functions at issue:

- [Option.value](https://github.com/Merlin04/ocaml-cobs/blob/f0128b7b7310933e059aeb99e7ca58f544ee8b8b/lib/cobs.ml#L12) which can be defined on one line or expanded to `match o with Some v -> v | None -> default` and 
- [String.of_bytes](https://github.com/Merlin04/ocaml-cobs/blob/f0128b7b7310933e059aeb99e7ca58f544ee8b8b/test/test_cobs.ml#L3) which, iiuc, can be replaced with [`Bytes.to_string`](https://kcsrk.info/webman/api/Bytes.html) to get compat all the way back to OCaml 4.02.

That said, nothing wrong with lower bounds :)

packages/cobs/cobs.0.1/opam
```diff
@@ -0,0 +1,39 @@
...
+description:
+  "A lightweight Consistent Overhead Byte Stuffing (COBS) encoder and decoder written in pure OCaml with no dependencies"
+maintainer: ["Merlin04"]
+authors: ["Merlin04"]
+license: "MIT"
+homepage: "https://github.com/Merlin04/ocaml-cobs"
+doc: "https://github.com/Merlin04/ocaml-cobs"
+bug-reports: "https://github.com/Merlin04/ocaml-cobs/issues"
+depends: [
+  "ocaml"
```
mseri, line 13:
```suggestion
  "ocaml" {>= "4.08"}
  "ocaml" {with-test & >= "4.13"}
```
This should fix the issues by restricting the compiler. You can go the route you prefer

Merlin04:
Ok, I fixed it in my repo (and also fixed a minor bug I found) - should I close this PR and use `opam publish` to file a new one for version `0.1.1`? Should I try to update this one somehow?

shonfeder:
You can do either way. It's most common for people to update the pr. But if you prefer just close this and create another that is fine.

shonfeder:
Thanks for all the work to get this packaged up nicely! :)

The only CI failure is the specious name collision detection.

LGTM.

packages/cobs/cobs.0.1.1/opam
```diff
@@ -0,0 +1,41 @@
+# This file is generated by dune, edit dune-project instead
+opam-version: "2.0"
+synopsis: "Consistent Overhead Byte Stuffing (COBS) encoding and decoding"
+description:
+  "A lightweight Consistent Overhead Byte Stuffing (COBS) encoder and decoder written in pure OCaml with no dependencies"
+maintainer: ["Merlin04"]
+authors: ["Merlin04"]
```
shonfeder, line 7: @mseri Just to confirm: does a github handle satisfy our policy for "a valid maintainer so that we can ping or send an email to them if needed"?

mseri: It would be preferrable to have an email, if that is possible

shonfeder: Thanks!

Are you ok with supplying an email address for contact here @Merlin04?

Merlin04: Sure, I could set up a custom alias on my domain for this? I'm worried about spam but I think if I use a custom alias then if it gets picked up by any scrapers I could just disable it. Has this been an issue for anyone in the past?

mseri: A custom alias is perfectly fine. I don't get different amount on spam on the address I use to sign my commits compared to the ones that I don't use on github, what I get is picked up from the spam filter and I don't really know which part is coming from being exposed here or from my emails being in some leak or some other website

Merlin04: fixed!

mseri:
Thanks for the patience. Feel free to announce the release on discuss.ocaml.org. We have a community category and an announcement tag for this purpose

Issue #26005, opened 2024-06-01
`bin_prot.0.17.0` fails to build on OpenSUSE and FreeBSD with OCaml 5.0


shonfeder:
Observed in https://github.com/ocaml/opam-repository/pull/25956 and https://github.com/ocaml/opam-repository/pull/25957#issuecomment-2137543695:

On OpenSUSE with OCaml 5.2 we have

```
#=== ERROR while compiling bin_prot.v0.17.0 ===================================#
# context              2.2.0~beta3~dev | linux/x86_64 | ocaml-base-compiler.5.2.0 | file:///home/opam/opam-repository
# path                 ~/.opam/5.2/.opam-switch/build/bin_prot.v0.17.0
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p bin_prot -j 255
# exit-code            1
# env-file             ~/.opam/log/bin_prot-7-1d174a.env
# output-file          ~/.opam/log/bin_prot-7-1d174a.out
### output ###
# (cd _build/default/xen && /usr/bin/bash -e -u -o pipefail -c './cflags.sh > cflags.sexp')
# Package mirage-xen was not found in the pkg-config search path.
# Perhaps you should add the directory containing `mirage-xen.pc'
# to the PKG_CONFIG_PATH environment variable
# No package 'mirage-xen' found
# File "xen/dune", line 8, characters 0-215:
#  8 | (library
#  9 |  (foreign_stubs
# 10 |   (language c)
# 11 |   (flags
# 12 |    (:include cflags.sexp)))
# 13 |  (name bin_prot_xen)
# 14 |  (public_name bin_prot.xen)
# 15 |  (optional)
# 16 |  (virtual_deps mirage-xen-ocaml)
# 17 |  (libraries)
# 18 |  (preprocess no_preprocessing))
# (cd _build/default && /home/opam/.opam/5.2/bin/ocamlmklib -g -o xen/bin_prot_xen_stubs xen/blit_stubs.o)
# /usr/lib64/gcc/x86_64-suse-linux/7/../../../../x86_64-suse-linux/bin/ld: xen/blit_stubs.o: warning: relocation against `caml_leave_blocking_section' in read-only section `.text'
[...]

PR #26008, opened 2024-06-03
bin_prot v0.17.0: fix freebsd error

mseri:
See https://github.com/ocaml/opam-repository/issues/26005


mseri:
That is not enough, the error remains as is. The cflags.sh file seem to not be there?

PR #26017, opened 2024-06-04
[new release] melange-recharts (4.0.3)

liubko:
Melange bindings for recharts

- Project page: <a href="https://github.com/ahrefs/melange-recharts/">https://github.com/ahrefs/melange-recharts/</a>

##### CHANGES:

- add `filterNull` to `Tooltip`
- support `melange.3.0.0` and `reason-react.0.14.0`

### DIFF
--- a/packages/melange-recharts/melange-recharts.4.0.3/opam
+++ b/packages/melange-recharts/melange-recharts.4.0.3/opam
@@ -18,7 +18,6 @@ depends: [
   "reason" {>= "3.10.0"}
   "reason-react" {>= "0.14.0"}
   "reason-react-ppx" {>= "0.14.0"}
-  "ocaml-lsp-server" {with-test}
   "odoc" {with-doc}
 ]
 build: [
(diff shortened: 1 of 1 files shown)


packages/melange-recharts/melange-recharts.4.0.3/opam
```diff
@@ -0,0 +1,50 @@
...
+homepage: "https://github.com/ahrefs/melange-recharts/"
+bug-reports: "https://github.com/ahrefs/melange-recharts/issues"
+depends: [
+  "dune" {>= "3.8"}
+  "ocaml"
+  "melange" {>= "3.0.0"}
+  "reason" {>= "3.10.0"}
+  "reason-react" {>= "0.14.0"}
+  "reason-react-ppx" {>= "0.14.0"}
+  "ocaml-lsp-server" {with-test}
```
mseri, line 21:
```suggestion
  "ocaml-lsp-server" {with-dev-setup}
```
Is it really a test dependency or just a dev setup one?

jchavarri: I think we can remove this package from the deps directly. The reason why I'd rather not use `with-dev-setup` is that the current version of opam (2.1) doesn't support it, so there's no way to call `opam install --with-dev-setup` until 2.2 is out.

liubko: I removed `ocaml-lsp-server` dependency

mseri:
Thanks

PR #26009, opened 2024-06-03
Fix revdeps from Yojson PR

mseri:
See https://github.com/ocaml/opam-repository/pull/25994
Failures in commits

PR #26020, opened 2024-06-05
Add missing dependency on cppo that used to be included by Yojson

Leonidas-from-XIV:
[Build fails](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/dd1f7a904024287a3f7f1288feb4a3694f856d28/variant/compilers,4.14,yojson.2.2.1,revdeps,pa_ppx_q_ast.0.09) with

```
/bin/sh: 1: cppo: not found
```

Which is due to Yojson not depending on CPPO anymore. Similar to #26009.


mseri:
It finished now, looks good thanks

PR #26031, opened 2024-06-06
Add MacPorts support for conf-autoconf and conf-libtool

MSoegtropIMC:
This PR adds MacPorts depext entries to conf-autoconf and conf-libtool.

I did not increase the version numbers because this doesn't change anything on OS variants where it did work before, so there is no need to recompile something on such platforms.


MSoegtropIMC:
The CI failures are mostly timeouts. The changes can only have an effect on macPorts MacOS and I don't see this tested at all in CI. So IMHO this is good to merge.

PR #26032, opened 2024-06-06
opam-graph and ohex are backed up to opam-source-archivs now

hannesm:
requires https://github.com/ocaml/opam-source-archives/pull/29

as suggested by @palainp @dinosaure


avsm:
Upstream PR merged, in this goes!

PR #26037, opened 2024-06-07
Package cvc5.1.1.3-unreleased

joaomhmpereira:
### `cvc5.1.1.3-unreleased`
OCaml bindings for the cvc5 SMT solver
OCaml bindings for the cvc5 SMT solver

---
* Homepage: https://github.com/formalsec/ocaml-cvc5
* Source repo: git+https://github.com/formalsec/ocaml-cvc5.git
* Bug tracker: https://github.com/formalsec/ocaml-cvc5/issues

---


packages/cvc5/cvc5.1.1.3-unreleased/opam
```diff
@@ -0,0 +1,42 @@
...
+    "@doc" {with-doc}
+  ]
+]
+dev-repo: "git+https://github.com/formalsec/ocaml-cvc5.git"
+url {
+  src:
+    "https://github.com/formalsec/ocaml-cvc5/archive/refs/tags/v1.1.3-unreleased.tar.gz"
+  checksum: [
+    "md5=2c34475ff7dfd5e92fca5ce1ba1c0d6d"
+    "sha512=cc1cae0cc1cddb2c1bb9fa7978bc43cc3a862c6fc4a8d7f79ae66c08e0af7fe82cac3c00d351e01806b064fce317705870f0a91bd0dafbb0ec06023786186bcb"
```
filipeom, line 40: Do a release with all the submodules included. Then, change the link and hashes here

packages/cvc5/cvc5.1.1.3-unreleased/opam
```diff
@@ -0,0 +1,42 @@
...
+homepage: "https://github.com/formalsec/ocaml-cvc5"
+bug-reports: "https://github.com/formalsec/ocaml-cvc5/issues"
+depends: [
+  "dune" {>= "3.10"}
+  "ocaml" {>= "4.12"}
+  "conf-gcc" {build}
+  "conf-g++" {build}
+  "conf-gmp" {build}
+  "conf-cmake" {build}
+  "conf-python-3" {build}
```
filipeom, line 17:
```suggestion
  "conf-python-3" {build}
  "conf-python-3-dev" {build}
```

packages/conf-python3-tomli/conf-python3-tomli.1/opam
```diff
@@ -0,0 +1,28 @@
...
+]
+depexts: [
+  ["python3-tomli"] {os-family = "debian"}
+  ["python3-tomli"] {os-family = "ubuntu"}
+  ["python3-tomli"] {os-distribution = "fedora"}
+  ["python-tomli"] {os-family = "suse" | os-family = "opensuse"}
+  ["python-tomli"] {os-family = "arch"}
+  ["py3-tomli"] {os-family = "alpine"}
+  ["py-tomli"] {os = "macos" & os-distribution = "macports"}
+  ["py-tomli"] {os = "freebsd"}
```
filipeom, line 22:
```suggestion
  ["py39-tomli"] {os = "macos" & os-distribution = "macports"}
  ["py39-tomli"] {os = "freebsd"}
```

Maybe try this also?

joaomhmpereira:
I think this is finally ready for review. A note on the remaining errors:

- `smtml` errors happen because the current release was using an older pin of the bindings. This will be fixed on `smtml`'s next release
- `s390x` tests keep failing with a killed process error

mseri:
Is the version correct? Right now it is going to be released as `cvc5` with version `1.1.3-unreleased`

Did I understand correctly that we need to add an upper bound to smtml to xcx5 < 1.1.3 ?

joaomhmpereira:
> Is the version correct? Right now it is going to be released as `cvc5` with version `1.1.3-unreleased`

The version `1.1.3-unreleased` refers to the latest version of [cvc5](https://github.com/cvc5/cvc5/releases/tag/latest) that is available but has not yet been formally released. Maybe it's better to keep the versioning of cvc5 OCaml bindings independent to the version of the actual solver and start at `0.1.0`?

 

> Did I understand correctly that we need to add an upper bound to smtml to xcx5 < 1.1.3 ?

This is only the case for `smtml.0.1.1` as the issue has been addressed in the following releases

shonfeder:
> The version `1.1.3-unreleased` refers to the latest version of [cvc5](https://github.com/cvc5/cvc5/releases/tag/latest) that is available but has not yet been formally released. Maybe it's better to keep the versioning of cvc5 OCaml bindings independent to the version of the actual solver and start at `0.1.0`?

If the intent and nature of these bindings is that are tied to that specific version of cvc5, and entail installation of it and only support compatibility with it, then I think it makes sense to have the version of this package track the version of cvc5 directly. WDYT, ts that the case?

mseri:
I am not against this. In this case, I think you should release it as version `1.1.3~unreleased`. In this way, once 1.1.3 is released, if there are changes, version `1.1.3` will appear as the latest version

joaomhmpereira:
> I am not against this. In this case, I think you should release it as version `1.1.3~unreleased`. In this way, once 1.1.3 is released, if there are changes, version `1.1.3` will appear as the latest version

I have corrected the version, I think now it's ready

mseri:
I will send a pr with the upper bounds of smtml and then publish. Can you look at the failure on debian-unstable? Is that expected?

PR #26038, opened 2024-06-07
Update z3 to 4.13.0-2, add "--gmp" flag to build

smuenzel:
Z3 requires the "--gmp" flag to actually use gmp, but we weren't providing it, even though we have gmp as a dependency.

### DIFF
--- a/packages/z3/z3.4.13.0-2/opam
+++ b/packages/z3/z3.4.13.0-2/opam
@@ -1,0 +1,49 @@
+opam-version: "2.0"
+maintainer: "weng@cs.jhu.edu"
+authors: "MSR"
+homepage: "https://github.com/Z3prover/z3"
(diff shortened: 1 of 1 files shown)


smuenzel:
nuscr 2.0.0 had incorrect dependencies for `base` specified, also fixing this to make the revdep build pass

mseri:
Thanks. I'll look into nuscr separately

wintersteiger:
That's not correct; Z3 does not require the `--gmp` flag. It's optional and may have significant and unexpected performance implications (see `mpz.cpp`). I think we should revert this change in the next package release.

mseri:
Should be never used or should conf-gmp be used as a depopt with optional linking? I think in that case it would be linked an almost all systems, so maybe it is not a good idea either

wintersteiger:
PR: https://github.com/ocaml/opam-repository/pull/26147

PR #26045, opened 2024-06-09
Restrict existing packages depending on mirage-crypto to < 1.0.0

hannesm:
The reasoning is that soon there'll be a mirage-crypto 1.0.0 release which has major breaking changes: Cstruct/bigarray is no longer being used, instead string/bytes are used. Since this is not backward-compatible, let's put bounds on everything and get new releases out rather sooner than later :)


hannesm:
I don't plan to fix any of the CI issues in this PR. This only adds upper bounds to not-yet released packages, so the change won't add any new failures. That the CI system fails here and there is a sign for me that we should get on to #23789 (or rephrased, the current state of opam-repository is not that you can install any old version and it will just work -- and I don't think this should be the goal).

mseri:
Packages that will appear to be working fine can have their upper bounds fixed later. I am going to merge to avoid overloading the CI

PR #26064, opened 2024-06-12
new file:   packages/conf-autoconf/conf-autoconf.0.2/opam

UnixJunkie:
@antoinemine tells me the name has changed on gentoo


mseri:
We could even just update the 0.1, however I think it is a good opportunity to drop dependence on which in this case

UnixJunkie:
Thanks for the improvements!

raphael-proust:
There are only revdeps failures and they seem unrelated.

PR #26065, opened 2024-06-12
Fix repo syntax (reported in https://discuss.ocaml.org/t/ocaml-ci-failure/14777)

hannesm:
The `||` is not valid, it should be `|`.

//cc @mseri @kit-ty-kate


hannesm:
See the report in https://discuss.ocaml.org/t/ocaml-ci-failure -- it is about line 13, character 45-46.

PR #26069, opened 2024-06-12
Add a basic Github Action CI for Windows

kit-ty-kate:
On behalf of the opam team, this PR brings Windows to the platforms tested by opam-repository, as opam 2.2.0 is nearing its final release.

For now it only checks the last compiler version (you'll need to remove the cache to update it) and might timeout for large enough PRs due to the fact that it tests all changed packages iteratively. It is also not clever and will also test even minor changes that do not affect the build (e.g. metadata changes), but i think it is better than nothing to start with

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -27,7 +27,6 @@ jobs:
         if: steps.cache-opam.outputs.cache-hit != 'true'
         run: |
           Invoke-Expression "& { $(Invoke-RestMethod https://raw.githubusercontent.com/kit-ty-kate/opam/windows-installer.debug/shell/install.ps1) } -OpamBinDir 'D:\opam\bin' -NoSetPath -NoAdmin"
-          "D:\opam\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
 
       - name: Init opam
         if: steps.cache-opam.outputs.cache-hit != 'true'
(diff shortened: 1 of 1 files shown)


.github/workflows/windows.yml
```diff
@@ -0,0 +1,74 @@
...
+
+      - name: Install opam
+        if: steps.cache-opam.outputs.cache-hit != 'true'
+        run: |
+          Invoke-Expression "& { $(Invoke-RestMethod https://raw.githubusercontent.com/kit-ty-kate/opam/windows-installer.debug/shell/install.ps1) } -OpamBinDir 'D:\opam\bin' -NoSetPath -NoAdmin"
+          "D:\opam\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+
+      - name: Init opam
+        if: steps.cache-opam.outputs.cache-hit != 'true'
+        run: opam init -yn .
```
raphael-proust, line 34:
```suggestion
        run: opam init --yes --no-setup .
```

.github/workflows/windows.yml
```diff
@@ -0,0 +1,74 @@
...
+        id: changed-files
+        uses: tj-actions/changed-files@v44
+
+      - name: List all changed packages
+        id: changed-packages
+        env:
+          ALL_CHANGED_FILES: ${{ steps.changed-files.outputs.all_changed_files }}
+        run: |
+          $output = @()
+          Foreach ($file in ($env:ALL_CHANGED_FILES).Split(" ")) {
```
raphael-proust, line 55: There is a lint check for spaces inside package names.

raphael-proust:
I'm merging because it looks ok and, even though I'm not very proficient in PowerShell, we can rollback anyway if need be.

PR #26072, opened 2024-06-12
Update conf-libssl for the mingw-w64 ports

dra27:
This commit is extracted from the [opam 2.2.0 Windows announcement](https://opam.ocaml.org/blog/opam-2-2-0-windows/) is intended in part to serve as a demonstration of the changes which need to be made to `conf-` packages to support the opam 2.2.0 depext scheme for native Windows following the merge of #25861.

The key parts:
- Constraining old versions of conf- packages
- Adding conf-mingw-w64-* packages for each architecture with MSYS2 and Cygwin package names
- Updating the latest version of the main conf- package with a dependency formula and adapting pkg-config to use `--personality`, etc.

### DIFF
--- a/packages/conf-libssl/conf-libssl.1/opam
+++ b/packages/conf-libssl/conf-libssl.1/opam
@@ -38,3 +38,4 @@ extra-source "osx-build.sh" {
     "md5=faf81a6bf30e346abe94610f136c6193"
   ]
 }
+available: os != "win32" | os-distribution = "cygwinports"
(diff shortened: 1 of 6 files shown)


mseri:
Wow, that is quite involved, it will make it hard to review and understand them

dra27:
Indeed, although it should be largely one-time pain (it's not like the conf- packages get materially updated that frequently) and, as in #26102, I'm very happy to be pinged to review them (even if that's slightly weird for the ones I've actually opened!)

PR #26073, opened 2024-06-12
GHA Windows: add a special case for packages that are not available

kit-ty-kate:
Should fix the issue seen in https://github.com/ocaml/opam-repository/pull/26072


dra27 (approved):
Thanks - that does indeed fix the CI run for #26072

PR #26080, opened 2024-06-13
GHA windows: do not fail when none of the packages can be installed

kit-ty-kate:
Somehow the last `$LASTEXITCODE` seems to be carried over to the whole process

Should fix the issue seen in https://github.com/ocaml/opam-repository/pull/26079


dra27 (approved):
Tested on a PR I'm about to open which also hit this, but I haven't actually tested with a failing PR

PR #26088, opened 2024-06-14
melange.4.0.0-52: require dune 3.16

emillon:
When dune 3.16 uses ocaml 5.2 (or melange acting like a 5.2 compiler), it will pass `-bin-annot-occurences`, but `melange.4.0.0-52` doesn't support it (4.0.1-52 does support it). See melange-re/melange#1132 and ocaml/dune#10622.


anmonteiro (approved):

raphael-proust:
Thanks!

PR #26102, opened 2024-06-17
Add configuration for gmp under mingw

hhugo:

PR #26103, opened 2024-06-17
[new release] dolmen (6 packages) (0.10)

Gbury:
A parser library for automated deduction

### UI

- Improved the printing of some data-structures (PR#190)

### Std

- Replace the `Plain` statement with the `Other` statement,
  which is a more general version (PR#190)

### Parsing

- Treat quoted symbols from the stdlib as symbols, regardless of
  their contents. Previously a `|assert|` would be understood as the
  reserved work `assert`, allowing e.g. `(|assert| false)`. From now on
  these are understood as symbols, so one can `(declare |assert| () Bool)`
  (PR#198)
- Add parsing extensions for the smtlib2 language (PR#190, PR#194)
- Better split elements of clauses in `cnf` TPTP statements (PR#190)
- Ensure illegal chars raise the correct error during lexing
  (Issue#191, PR#192)

### Printing

- Add printers for smtlib identifiers (PR#198)
- Printing of typed expressions (i.e. the Std.Expr module)
  can now print the tags (PR#210)

### Typing

- Enforce some missing constraints on bitvectors
  indexes and sizes for the smtlib2 BV theory (PR#172, PR#175)
- Slightly improve wording for errors and warnings concerning
  non linearity and other arithmetic restrictions (PR#184)
- More information for reserved Id, resulting in more precise
  errors when smt2 scripts use reserved ids (PR#193)
- Expose implicit declarations/definitions that happen during
  typechecking (PR#199)
- Treat smtlib `:named` annotations as implicit definitions as
  required by the spec (PR#199)
- Add a warning for unknown attributes in smtlib2. This replaces
  the `unbound id` error that some files could raise before when
  using non-standard attribtues (PR#207)
- Only type annotations on quantified formulas and binders once.
  Previously, these were typed twice so that they can be attached to
  both the body of the bound formula and the quantified formula itself.
  (PR#207, PR#210)

### Loop

- Add the String theory to the ALL logic of smtlib2 (PR#182)


Gbury:
- The windows failure seem to be due to missing depexts for zarith/gmp (maybe it willl be fixed by #26102 ? )
- The CI failures on linux were because of missing lower bounds on `qcheck` and `mdx` (which are tests deps). I added some lower bounds (see [the upstream commit](https://github.com/Gbury/dolmen/commit/092fc45b94674cf8c29c49efc6861569563e3de0) )

Gbury:
Upper bounds on the rev-dep  `alt-ergo-lib` have been added (since this new version of `dolmen` changes the API).

Gbury:
After the fixes:
- windows CI is still failing (as expected) because zarith fails to install (but it's not caused by this PR)
- opam-ci seems to be stuck waiting on some ressources for the revdeps test on alt-ergo-libs with the following last lines on the log:
  > 2024-06-17 19:33.31: Waiting for resource in pool OCluster
  > 2024-06-17 19:33.32: Waiting for worker…
  
  otherwise all tests seem to have passed

avsm:

PR #26112, opened 2024-06-18
mlcuddidl.3.0.8

nberth:
Adds support for building native plugins

### DIFF
--- a/packages/mlcuddidl/mlcuddidl.3.0.8/opam
+++ b/packages/mlcuddidl/mlcuddidl.3.0.8/opam
@@ -1,0 +1,28 @@
+opam-version: "2.0"
+maintainer: "Nicolas Berthier <m@nberth.space>"
+authors: ["Bertrand Jeannet" "Nicolas Berthier"]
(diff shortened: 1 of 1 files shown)


avsm:
Fallout for camlidl for the caml.inria.fr server being down, but also some C pointer conversion warnings with newer glibc. Do you want to look over those @nberth?

nberth:
> Fallout for camlidl for the caml.inria.fr server being down, but also some C pointer conversion warnings with newer glibc. Do you want to look over those @nberth?

@avsm Sure!
Updated specifications for `camlidl` are available in PR #26138.
The warning about pointer conversion is actually due to some code generated by `camlidl` itself. I tried working that around, but did not really obtain anything cleaner without a warning appearing somewhere. Fixing this properly may actually require somme changes in `camlidl` (which hardcodes some assignements to `NULL` without any means to cast). The code involved is about propagating exceptions thrown by callbacks into OCaml, that we have to intercept at the C level to deal with CUDD's memory management, and then re-raise. `camlidl` may be too limited when it comes to manipulating exception values.

mseri:
Thanks. If the compiler issue is fixed in camlidl, I guess this will automatically start working in the two remaining systems

PR #26120, opened 2024-06-19
Deleting oma.20240106 because three functions contain serious bugs.

fpottier:
These bugs are fixed in the next release (submitted today).


raphael-proust:
We generally prefer the packages to be marked with

```
flags: [ avoid-version ]
```

rather than deleted. And this applies to buggy packages too. Is there any reason to delete rather than mark to be avoided?

fpottier:
No, no reason. I didn't know about `avoid-version`. I can use that.

fpottier:
Done.

fpottier:
The checks have passed.

PR #26115, opened 2024-06-18
Add configuration for zlib under cygwin/mingw

zilbuz:

PR #26130, opened 2024-06-22
Update some Cygwin depexts

dra27:
Spotted while reviewing #26115 and to expand on https://github.com/ocaml/opam-repository/pull/24432#issuecomment-2183879312.

Cygwin provides two things in the Windows OCaml ecosystem:
1. It's an entire distribution of Unix built on top of Windows. In this context, it has GCC, libraries and so forth and from OCaml's perspective behaves like a BSD flavour with some Linuxisms. opam running in this context has `os = "cygwin"` and `os-distribution = "cygwin"`
2. It's a provider of both cross-compilers and cross-compiled libraries for the mingw-w64 native port of OCaml. opam running in this context has `os = "win32"` and `os-distribution = "cygwin"`.

opam for native Windows also uses tooling from Cygwin (for example, `make`). This means that there are two possible filters we can need for depexts:
- `os = "cygwin"` is for things _specific_ to opam running in Cygwin itself. This will generally by C libraries only.
- `os-distribution = "cygwin"` is generally for tools used by both opam running in Cygwin but also by native Windows opam when building packages.

The latter correspond to depexts prefixed `system:` for `os-distribution = "cygwinports"`. This PR updates depexts presently in opam-repository for Cygwin and was based on a search for `"system:` (to catch existing opam-repository-mingw depexts) and `os-distribution = "cygwin"` (to catch incorrectly tagged Cygwin-only depexts).

The commit messages contain further explanations, as conf-pkg-config and conf-git are both exceptions to the rules above.

### DIFF
--- a/packages/conf-zig/conf-zig.1/opam
+++ b/packages/conf-zig/conf-zig.1/opam
@@ -17,6 +17,7 @@ depexts: [
   ["zig"] {os = "freebsd"}
   ["zig"] {os = "macos" & os-distribution = "homebrew"}
   ["system:zig"] {os = "win32" & os-distribution = "cygwinports"}
+  ["zig"] {os-distribution = "cygwin"}
 ]
 synopsis: "Virtual package relying on zig"
 description:
--- a/packages/conf-zig/conf-zig.1/opam
+++ b/packages/conf-zig/conf-zig.1/opam
@@ -16,8 +16,6 @@ depexts: [
   ["zig"] {os-distribution = "arch"}
   ["zig"] {os = "freebsd"}
   ["zig"] {os = "macos" & os-distribution = "homebrew"}
-  ["system:zig"] {os = "win32" & os-distribution = "cygwinports"}
-  ["zig"] {os-distribution = "cygwin"}
 ]
 synopsis: "Virtual package relying on zig"
 description:
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -80,14 +80,17 @@ jobs:
           $failed = $false
           opam update
           Foreach ($pkg in $pkgs) {
-            opam install --confirm-level=unsafe-yes "$pkg"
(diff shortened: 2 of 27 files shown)


packages/conf-zig/conf-zig.1/opam
```diff
@@ -17,6 +17,7 @@ depexts: [
   ["zig"] {os = "freebsd"}
   ["zig"] {os = "macos" & os-distribution = "homebrew"}
   ["system:zig"] {os = "win32" & os-distribution = "cygwinports"}
+  ["zig"] {os-distribution = "cygwin"}
 ]
```
avsm, line 21:
```suggestion
]
depends: [ "conf-pkg-config" ]
```

due to 
```
#=== ERROR while compiling conf-zlib.1 ========================================#
"pkg-config": command not found.
```

dra27: The error was in conf-zlib.1, not conf-zig.1 (but see note below)

dra27:
It turns out that the depext for "zig" appears to have been a result of a conf- file being produced with sed in #22030 - zig has never been packaged for Cygwin, so I've updated it to remove the depext instead. Similarly, while re-reviewing, I see that the Cygwin depext ought to be in the questionable conf-gtksourceview3.0+2, so added it there.

The other new commit could go in a separate, but it serves a useful demo here - it makes the output from GitHub Actions much clearer by folding the `opam install` logs and colouring the output.

I'd like to ignore the Windows failures, if that's OK - these conf- packages move from a bad state to a less bad state, even if not yet to a perfect state.

conf-asciidoc and conf-libtool won't fail in the Cygwin port of opam (which is what these updates were for) and that isn't tested here. They won't fail in either opam 2.2.1 or 2.3 for native Windows, either, but it's not necessarily important to fix that right now.

The other packages are all libraries, and need considerable updates for mingw-w64 (cf. #26072, #26102, etc.)

mseri:
thanks

PR #26144, opened 2024-06-25
packages not available on windows because they embed an oasis configu…

hhugo:
…re script broken on windows

### DIFF
--- a/packages/assertions/assertions.0.1/opam
+++ b/packages/assertions/assertions.0.1/opam
@@ -19,6 +19,7 @@ synopsis: "Basic assert statements"
 description: """
 Provides basic assert statements, like `assert_equal` and `assert_raises`.
 To be used independently or as a supplement to `ounit` or `pa_ounit."""
+available: [os != "win32"]
 flags: light-uninstall
 url {
   src: "https://github.com/bennn/assertions/archive/0.1.tar.gz"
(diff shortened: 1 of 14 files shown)


hhugo:
This one should be easy to merge.

mseri:
Can you paste here a protypical error message?

mseri:
Ping @jonahbeckford, is that ok dkml-wise?

jonahbeckford:
This doesn't seem like a Windows problem. It is just failing because the lack of shell/configure support on whichever Windows machine was tested.

So, without me having to do parallel testing on every opam package, the exclude from all Windows  is too broad.

hhugo:
This is definitely a windows issue, these packages have their setup file generated from an old oasis version. The issue come from the following code that give the wrong location to ocamlbuild library
```
let fix_args args extra_argv =
    List.flatten
      [
        if (os_type ()) = "Win32" then
          [
            "-classic-display";
            "-no-log";
            "-no-links";
            "-install-lib-dir";
            (Filename.concat (standard_library ()) "ocamlbuild")
          ]
        else
```

Ocamlbuild is no longer installed as a sub directory of the stdlib.

jonahbeckford:
Ah, thanks for correcting me. Definitely applies to all Windows machines. Disable it!

mseri:
Thanks for the clarifications and for checking

PR #26147, opened 2024-06-25
Z3: remove dependency on gmp

wintersteiger:


wintersteiger:
The Windows/Cygwin CI fails because the Visual C Compiler `cl.exe` is not available (but mingw-gcc is). Is there a way to get the "currently opam-installed" C++ compiler?

PR #26149, opened 2024-06-26
[new release] html_of_jsx (0.0.4)

davesnx:
Render HTML with JSX

- Project page: <a href="https://github.com/davesnx/html_of_jsx">https://github.com/davesnx/html_of_jsx</a>

## 0.0.4

- [BREAKING] Handle HTML encoding for `'` (@davesnx)
- Handle HTML encoding for `"` (from `&davesnx/html_of_jsx#34;` to `&quot;`) (@davesnx)
- Improved performance of `JSX.render` (@davesnx)
- [BREAKING] Remove `Fragment` in favor of `JSX.list` (@davesnx)
- Remove unused `Component (unit -> element)` since it isn't needed (@davesnx)
- [BREAKING] Change attributes representation (@andreypopp)
- [BREAKING] Remove melange dependency (@andreypopp)
- [BREAKING] Lower the OCaml bound to 4.14 (@davesnx)
- Make lib wrapped (@andreypopp)

### DIFF
--- a/packages/html_of_jsx/html_of_jsx.0.0.4/opam
+++ b/packages/html_of_jsx/html_of_jsx.0.0.4/opam
@@ -10,7 +10,7 @@ bug-reports: "https://github.com/davesnx/html_of_jsx/issues"
 depends: [
   "dune" {>= "3.12"}
   "ocaml" {>= "4.14"}
-  "ppxlib" {> "0.23.0"}
+  "ppxlib" {>= "0.25.0"}
   "alcotest" {with-test}
   "benchmark" {with-test}
   "reason" {>= "3.10.0" & with-test}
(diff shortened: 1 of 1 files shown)


davesnx:
The error on 4.14, it installs ppxlib 0.24 downgraded from 0.32 and it depends on ocaml compiler libs and fails with `Unbound value Location.error_extensionf`. I will lift the minimum version of 0.25 (where this method is available and see if the CI is happy)

davesnx:
The error comes from `freebsd (experimental)` and I remember ignoring in my previous PRs, do we still forget about those cases?

avsm:
It's ignorable for this merge, but also solveable for the next release: you need to not assume that `/bin/bash` exists, which ppx.sh seems to in your dune rules.

PR #26046, opened 2024-06-09
ppx_base 0.17.0 can't install on windows

hhugo:
See https://github.com/janestreet/ppx_base/issues/2

PR #26186, opened 2024-07-04
Fix available for ppx_base

hhugo:
ppx_ base was marked as unavailable in #26046, this incompatibility was resolved by #26142.


mseri:
Thanks

PR #26066, opened 2024-06-12
opam-state 2.2.0~beta3: add upper bounds to revdeps

mseri:
See https://github.com/ocaml/opam-repository/pull/26053

PR #26187, opened 2024-07-04
Allow opam-state 2.2.0 instead of 2.2.0~beta3 upper bound

sim642:
These were added in PR #26066 due to opam-state API change in 2.2.0~beta3.

However, the final version of opam-state 2.2.0 undid the API change: https://github.com/ocaml/opam/pull/6058. This allows the final version to be used by these packages again.

opam-dune-lint 0.{1,2} are excluded because they were made entirely incompatible in 364f7dd34e8e6bf87f393e507093d0b21cde7a98.

I hope `>= "2.2.0"` means anything newer than alpha/beta/rc. If not, then please suggest the correct bound.

It would have make more sense (at least to me) to just replace the upper bound with a conflict for
```
"opam-state" {>= "2.2.0~beta3" & < "2.2.0"}
```
but according to `opam lint`, conjunctions are not allowed in conflicts apparently.

### DIFF
--- a/packages/odep/odep.0.2.0/opam
+++ b/packages/odep/odep.0.2.0/opam
@@ -12,7 +12,7 @@ depends: [
   "ppx_sexp_conv" {>= "v0.13"}
   "parsexp"
   "opam-core" {>= "2.1.0"}
-  "opam-state" {>= "2.1.0" & < "2.2.0~beta3"}
+  "opam-state" {>= "2.1.0" & (< "2.2.0~beta3" | >= "2.2.0")}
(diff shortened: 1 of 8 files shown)


sim642:
The Windows CI is failing with some linking thing:
```
  # ** Cannot resolve symbols for D:\opamroot\default\lib\opam-core\opam_core.a(opamStubs.o):
  #  win_create_process_native
  # Error: Error during linking (exit code 2)
```

mseri:
Thanks

PR #26188, opened 2024-07-04
[new release] Five Ortac packages 0.3.0

n-osborne:
This release brings a number of new features and improvements:

- Improvement of the user interface:
  + Switch to a module based configuration to simplify the command line
  + Release a `dune` plugin to generate Dune boilerplate
  + Overwrite the default `cleanup` function if there is one defined in the configuration module
- Extend supported Gospel subset for models:
  + Add support for custom ghost type as model: you are no longer limited to the Gospel standard library to define the model of your OCaml types
  + Add support for functional type in model: you can use the arrow type as a model
- Extend coverage of generated tests:
  + Add support for testing functions without a SUT argument
  + Add support for tuples, both as argument and returned value
- Fix:
  + Add an error when none of the functions can be tested rather than generating illegal OCaml code
  + Handle correctly cast from `int` to `integer` included by Gospel when computing the expected returned value in failure message
  + Fix field access translation


n-osborne:
Update version constraint on Monolith.

mseri:
Thanks

PR #26190, opened 2024-07-04
wryd.1.4.6: fix missing archive by using opam-source-archives

kit-ty-kate:
Depends on https://github.com/ocaml/opam-source-archives/pull/31


mseri:
Thanks

PR #26194, opened 2024-07-04
[new release] Docfd 7.0.0

darrenldl:

https://github.com/darrenldl/docfd

Think interactive grep for text files, PDFs, DOCXs, etc,

and PDF viewers,

Features:


packages/docfd/docfd.7.0.0/opam
```diff
@@ -0,0 +1,82 @@
...
+    name
+    "-j"
+    jobs
+    "@install"
+    "@doc" {with-doc}
+  ]
+]
+pin-depends: [
+  [ "nottui.0.3" "git+https://github.com/let-def/lwd.git#a337a778001e6c1dbaed7e758c9e05f300abd388" ]
+]
```
shonfeder, line 74: As flagged by our linting rules, `pin-depends` are not permitted in the repo. Do you think you could work on getting this merged into nottui first?

shonfeder (commented):
Thank you for adding the new release :)

packages/docfd/docfd.7.0.0/opam
darrenldl: I know someone is working on a nottui fork, I'll wait for that first then. Thanks.

PR #26207, opened 2024-07-07
Camlzip 1.12

xavierleroy:
- Add full support for ZIP64 archives
- Fix memory leak when a `Zlib.stream` is finalized and `Zlib.deflate_end` / `Zlib.inflate_end` was not called before

### DIFF
--- a/packages/camlzip/camlzip.1.12/opam
+++ b/packages/camlzip/camlzip.1.12/opam
@@ -14,7 +14,7 @@ build: [
 ]
 install: [make "install"]
 depends: [
-  "ocaml" {>= "4.07.0"}
+  "ocaml" {>= "4.13.0"}
   "ocamlfind" {build}
   "conf-zlib"
 ]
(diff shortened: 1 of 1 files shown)


mseri:
The support for Zip64 removed the `extra` field, I need to add a few upper bounds before I can merge

xavierleroy:
You're right, sorry for the additional work.

Exposing the "extra" ZIP data as a string was a design mistake of Camlzip, as it's really a list of (tag, string) lists.  With ZIP64, some of these tags have special meaning, so I though it would be best to not expose the extra data at all, hoping that nobody uses it.

I'll try to check the Camlzip clients that break with this release, to understand how they use the `extra` field and suggest them not to use it.

mseri:
It seems to be a very limited breakage. The use was introduced in bap-byteweight.2.5. Let me ping @ivg so that they are aware.

mseri:
Thanks!

PR #26209, opened 2024-07-07
Cryptokit 1.20

xavierleroy:
- Name space depollution: make C implementations of ciphers local to the OCaml/C stub code, so that they do not conflict with other C libraries implementing crypto functions with the same names


xavierleroy:
Something is wrong in the OPAM build (that did not show up when testing locally).  Closing this PR until the issue is solved.

PR #26221, opened 2024-07-09
camllib.1.3.3: update availability range

nberth:
For a reason I don't really grasp (maybe an unintended change of the archive that I host on my own website), I am unable to find an archive that corresponds to `camllib.1.3.3` with the previous checksums. I am not sure what is the policy regarding such a case (there's no real reason to make a new version that would be identical to the previous one).

Note that package is quite old and seems very rarely used anyways.

### DIFF
--- a/packages/camllib/camllib.1.3.3/opam
+++ b/packages/camllib/camllib.1.3.3/opam
@@ -19,6 +19,7 @@ depends: [
   "ocamlbuild" {build}
   "stdlib-shims"
 ]
+available: [ os != "win32" & os != "cygwin" ]
 synopsis: "Utility Library (including various datatypes)"
 url {
   src: "https://nberth.space/pool/camllib/camllib-1.3.3.tar.gz"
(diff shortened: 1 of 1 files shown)


packages/camllib/camllib.1.3.3/opam
```diff
@@ -2,28 +2,29 @@ opam-version: "2.0"
...
+  "ocaml" {>= "4.02.0" & < "5"}
   "ocamlfind" {build}
   "ocamlbuild" {build}
   "stdlib-shims"
 ]
+available: [ os != "win32" & os != "cygwin" ]
 synopsis: "Utility Library (including various datatypes)"
 url {
-  src: "http://nberth.space/pool/camllib/camllib-1.3.3.tar.gz"
+  src: "https://nberth.space/pool/camllib/camllib-1.3.3.tar.gz"
```
mseri, line 25: The original tarball can be downloaded from https://gitlab.ocamlpro.com/OCamlPro/opam-repository/-/blob/cached/cache/md5/59/59424473c5a79c4a38da034de36b0293
would it be possible to keep the old hashes and upload this one on `nberth.space` instead?

mseri: Awesome, thanks a lot!

PR #26223, opened 2024-07-09
[new release] flint (4 packages) (0.4.0)

bobot:
Stub of the C library Flint3

- Project page: <a href="https://github.com/bobot/ocaml-flint">https://github.com/bobot/ocaml-flint</a>

##### CHANGES:

  - As in the C library regroup all package into flint
  - Bump to flint version 3.0.0

### DIFF
--- a/packages/antic/antic.0.4.0/opam
+++ b/packages/antic/antic.0.4.0/opam
@@ -1,0 +1,36 @@
+opam-version: "2.0"
+synopsis: "deprecated use flint"
+maintainer: ["François Bobot"]
+authors: ["François Bobot"]
+license: "LGPL-2.1-only"
+homepage: "https://github.com/bobot/ocaml-flint"
+bug-reports: "https://github.com/bobot/ocaml-flint/issues"
+depends: [
+  "dune" {>= "3.7"}
+  "flint" {= version}
+  "odoc" {with-doc}
+]
+build: [
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
+dev-repo: "git+https://github.com/bobot/ocaml-flint.git"
+url {
+  src:
+    "https://github.com/bobot/ocaml-flint/releases/download/0.4.0/flint-0.4.0.tbz"
+  checksum: [
+    "sha256=65625a5c4ab3922a9b39bca0f42a57b3ee18844786c4672c68da98a7bbc863d8"
+    "sha512=c92b1099cacac6bf7a181db0c52444d36180b02a1b2aa5c8aff943318f7c3440457126fb0c2dcf56bf341c05c0f287a67b8df9d05f19fe0321b992a0e04e4a8e"
+  ]
+}
+x-commit-hash: "4cd9eaee849258fcec12549488ccc501fdc9d4a8"
--- a/packages/arb/arb.0.4.0/opam
+++ b/packages/arb/arb.0.4.0/opam
@@ -1,0 +1,36 @@
+opam-version: "2.0"
+synopsis: "deprecated use flint"
+maintainer: ["François Bobot"]
+authors: ["François Bobot"]
+license: "LGPL-2.1-only"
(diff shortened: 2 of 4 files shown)


mseri:
The packages are building where they can.

Note that extra files are no longer allowed on this repository, so packages/conf-flint/conf-flint.3.0/files/test.c needs to be uploaded to https://github.com/ocaml/opam-source-archives/tree/main/patches and referenced from there (with a checksum)

Gbury:
The CI failed to build on archlinux because `conf-flint` failed during install (see https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/930d908b89a0cdb37c68ce56d5a1b00e5646ef3c/variant/distributions,archlinux-ocaml-4.14,conf-flint.3.0 ). Unless I missed something, there seems to be a missing line in the depexts for `conf-flint` for archlinux.

packages/conf-flint/conf-flint.3.0/opam
```diff
@@ -0,0 +1,37 @@
...
+  ["libflint-dev"] {os-family = "ubuntu"}
+  ["flint"] {os = "macos" & os-distribution = "homebrew"}
+  ["flint"] {os-distribution = "macports" & os = "macos"}
+  ["flint" "flint-devel"] {os-distribution = "centos"}
+  ["flint" "flint-devel"] {os-distribution = "fedora"}
+  ["flint" "flint-devel"] {os-distribution = "ol"}
+  ["flint-dev"] {os-distribution = "alpine"}
+  ["flint-devel"] {os-family = "suse"}
+  ["libflint-devel"] {os = "win32" & os-distribution = "cygwinports"}
+  ["flint"] {os-distribution = "nixos"}
```
Gbury, line 30:
```suggestion
  ["flint"] {os-distribution = "nixos"}
  ["flint"] {os-distribution = "arch"}
```
(note: I'm unsure whether it's best to use `os-distribution` or `os-family` as both examples seem to exist)

bobot: I made the change locally, I trying to understand why the test file doesn't work.

Gbury: I don't know if that's relevant or not, but at least on `arch`, the include files are in `/usr/include` and not `/usr/local/include`

mseri:
@bobot sure, generating it on the fly is perfectly fine. I think there are also a couple of other packages doing it (for the same reason)

packages/conf-flint/conf-flint.3.0/opam
bobot: I just though the last item was ubuntu-24 (where it should work) and it is ubuntu-22 (where it should not work). I tried to use os-version to separate the cases.

mseri:
Some distributions derived from ubuntu show `os-family` ubuntu nowadays, that is why some users have started adding it to conf packages

packages/conf-flint/conf-flint.3.0/opam
```diff
@@ -0,0 +1,47 @@
...
+  ["flint" "flint-devel"] {os-distribution = "centos"}
+  ["flint" "flint-devel"] {os-distribution = "fedora"}
+  ["flint" "flint-devel"] {os-distribution = "ol"}
+  ["flint-dev"] {os-distribution = "alpine"}
+  ["flint-devel"] {os-family = "opensuse"}
+  ["libflint-devel"] {os = "win32" & os-distribution = "cygwinports"}
+  ["flint"] {os-distribution = "nixos"}
+  ["flint"] {os-distribution = "arch"}
+]
+available: [
```
mseri, line 34: I am also ok in letting these distributions fail in the CI. Users can still install flint3 manually and then install this

bobot: On previous version users can use transparently conf-flint.1 , which will use previous release of ocaml-flint. The new release of flint have backward compatible API with the old one. So it is easier for them to just ask the installation of flint and dependending on the version of the OS they will get flint v2 or flint v3 under the hood.

mseri: Sounds good!

mseri:
If you are fine with it, I am happy to merge since this is building everywhere it can now.

mseri:
Thanks a lot

PR #26225, opened 2024-07-10
[new release] xedbindings (0.9)

percontation:
Bindings to [Intel XED](https://github.com/intelxed/xed)

https://github.com/percontation/ocaml-xed-bindings/


percontation:
Yeah, I think I can get more of those failures working (or at least add a platform-availability filter if XED really won't build on some of these).

percontation:
Alright, should be good now. The only two failures are the opensuse-15.6 runners noticing that installation on platforms with python3 version <3.8 will fail (with an appropriate error message), but AFAIK this is the best option.

mseri:
Thanks. Looks fine

PR #23330, opened 2023-02-16
conf-gmp: use pkg-config where available

mseri:
See https://github.com/ocaml/opam-repository/pull/18129 and https://github.com/ocaml/opam-repository/pull/22954

I need to test the dependencies in `opam list --depends-on conf-gmp` before un-drafting

PR #26226, opened 2024-07-10
ZArith version 1.14

xavierleroy:
https://github.com/ocaml/Zarith/issues/148, https://github.com/ocaml/Zarith/pull/149: Fail unmarshaling when it would produce non-canonical big ints
    https://github.com/ocaml/Zarith/pull/145, https://github.com/ocaml/Zarith/pull/150: Use standard hash function for Z.hash and add Z.seeded_hash
    https://github.com/ocaml/Zarith/issues/140, https://github.com/ocaml/Zarith/pull/147: Add fast path for Z.divisible on small arguments

### DIFF
--- a/packages/zarith/zarith.1.14/opam
+++ b/packages/zarith/zarith.1.14/opam
@@ -1,5 +1,4 @@
 opam-version: "2.0"
-name: "zarith"
 maintainer: "Xavier Leroy <xavier.leroy@inria.fr>"
 authors: [
   "Antoine Miné"
--- a/packages/zarith/zarith.1.14/opam
+++ b/packages/zarith/zarith.1.14/opam
@@ -38,7 +38,7 @@ install: [
   [make "install"]
 ]
 depends: [
-  "ocaml"     {>= "4.04.0"}
+  "ocaml"     {>= "4.07.0"}
   "ocamlfind"
   "conf-gmp"
 ]
--- a/packages/zarith/zarith.1.14/opam
+++ b/packages/zarith/zarith.1.14/opam
@@ -1,0 +1,59 @@
+opam-version: "2.0"
+name: "zarith"
+maintainer: "Xavier Leroy <xavier.leroy@inria.fr>"
+authors: [
+  "Antoine Miné"
+  "Xavier Leroy"
+  "Pascal Cuoq"
+]
+homepage: "https://github.com/ocaml/Zarith"
+bug-reports: "https://github.com/ocaml/Zarith/issues"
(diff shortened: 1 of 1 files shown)


packages/zarith/zarith.1.14/opam
```diff
@@ -0,0 +1,58 @@
...
+    "sh"
+    "-exc"
+    "LDFLAGS=\"$LDFLAGS -L/opt/local/lib -L/usr/local/lib\" CFLAGS=\"$CFLAGS -I/opt/local/include -I/usr/local/include\" ./configure"
+  ] {os = "macos" & os-distribution = "homebrew" & arch = "x86_64" }
+  [
+    "sh"
+    "-exc"
+    "LDFLAGS=\"$LDFLAGS -L/opt/homebrew/lib\" CFLAGS=\"$CFLAGS -I/opt/homebrew/include\" ./configure"
+  ] {os = "macos" & os-distribution = "homebrew" & arch = "arm64" }
+  [make]
```
mseri, line 34: Now that [`zarith` supports `pkg-config`](https://github.com/ocaml/Zarith/pull/132) I think it may be worth just using that. At least on macos we could simply add `conf-pkg-config` as a dependency and use the same invocation as linux. Maybe this works also on freebsd

xavierleroy: It is tempting to rely on pkg-config, indeed.  It works very well for developing on Linux and macOS.  But I'm afraid of breaking what seems to be working today...  

At any rate, if we go the pkg-config way, I'd like all the platforms to use it, no special cases if at all possible.  But I realize I don't know how things work for Windows.

mseri: I don't know about windows, so I would not touch anything there, but according to the tests I made for https://github.com/ocaml/opam-repository/pull/23330 (for all packages with a checkmark I was also testing all the revdeps I could install on my computer) there seems to be no breakage

I would suggest to add a commit to use `["./configure"]` on all systems and then we see from the CI if this works also on freebsd. Ping @kit-ty-kate, do you know if openbsd would work as well or needs a custom call?

kit-ty-kate: as far as i know openbsd should work out of the box with pkg-config. The base system ships with pkg-config, its only problem is that it doesn't support some of the arguments but i'm not seeing the use of them in zarith so it should be fine.

xavierleroy: YOLO.  See latest commit.

mseri:
The CI logs look good. Thanks a lot!

PR #25896, opened 2024-05-17
Added new version 1.4.2 for gappa and changed URL for older gappa

MSoegtropIMC:
This PR adds new version 1.4.2 for gappa - a tool to automatically generate arithmetic proofs.

I also changed the tar ball URLs for all older versions to a stored and not on the fly generated source. Recently the hash of gappa 1.3.5 changed magically (this happens sometimes on github).

The C++ 11 config patch should not be required any more for version 1.4.2, but we need to see what opam CI says to this.

PR #26228, opened 2024-07-10
Fixing gappa link and checksum

rtetley:


packages/gappa/gappa.1.3.5/opam
```diff
@@ -40,8 +40,8 @@ depends: [
 ]
 synopsis: "Tool intended for formally proving properties on numerical programs dealing with floating-point or fixed-point arithmetic"
 url {
-  src: "https://gitlab.inria.fr/gappa/gappa/-/archive/gappa-1.3.5.tar.gz"
-  checksum: "sha512=29ce59af97e6d60547a193b43538f4812ff74fb01a812cda7109855219457fa7a47f59ea39aff2a5e03fd70181e024a3296b4f48300818a81f62fd2d8629c389"
+  src: "https://gappa.gitlabpages.inria.fr/releases/gappa-1.3.5.tar.gz"
+  checksum: "sha512=60b5719e3a321df43e33045fa8f4511fc02a4218d1ae7e476e7c6ebcf90ae208832881f6eea5b99a3296dfcc3a18c7e1f4ea9dbea446fc502e14306b6975f6e6"
```
mseri, line 44: We are trying to avoid changing the checksums. Can you get the release from the opam cache and replace the tarball at the new address? 
Caches are listed here: https://github.com/ocaml/opam-repository/wiki/FAQ#how-to-find-lost-archives-of-packages

MSoegtropIMC:
@rtetley, @mseri : this is the same as https://github.com/ocaml/opam-repository/pull/25896 (minus the new version). I think we should close this PR in favour of the other.

Unfortunately I did not yet have time to look into the CI issues in in the PR mentioned above. @rtetley : do you want to look into this.

Regarding the hash and path change: it is not so the the file got lost, it is so that due to some reconfiguration of the git server the hash changed (this does happen). So we want to change the path to the actual release tar balls copied at the time of release instead of auto generated tar balls.

Btw.: rtetley and me are working together but obviously not talking as much as we should :-)

mseri:
I am closing this one then, thanks for the heads up :)

PR #26233, opened 2024-07-11
Ctypes does not compile using msvc

pirbo:
See https://github.com/yallop/ocaml-ctypes/pull/685
Work is needed to be able to use ctypes in a swicth that uses msvc compiler on Windows...


dra27 (approved):
[This GitHub Actions run on my fork](https://github.com/dra27/opam-repository/actions/runs/9977798999/job/27573267983?pr=19#step:12:590) shows the older ctypes packages correctly skipping, rather than erroring, as it combines this PR with #26255.

mseri:
Thanks

PR #26237, opened 2024-07-12
Package mopsa.1.0~pre6

rmonat:
### `mopsa.1.0~pre6`
MOPSA: A Modular and Open Platform for Static Analysis using Abstract Interpretation
MOPSA is a generic framework for building sound static analyzers based on Abstract Interpretation.
It features a modular architecture to support different kinds of languages, iterators, and abstract domains.
For the moment, MOPSA can analyze programs written in a subset of C and Python.
It reports run-time errors on C programs and uncaught exceptions on Python programs.

---
* Homepage: https://gitlab.com/mopsa/mopsa-analyzer
* Source repo: git+https://gitlab.com/mopsa/mopsa-analyzer.git
* Bug tracker: https://gitlab.com/mopsa/mopsa-analyzer/issues

---

### DIFF
--- a/packages/mopsa/mopsa.1.0~pre6/opam
+++ b/packages/mopsa/mopsa.1.0~pre6/opam
@@ -30,18 +30,20 @@ depends: [
   "dune" {>= "3.7"}
   "ocamlfind"
   "apron" {>= "0.9.15"}
-  "menhir"
+  "menhir" {>= "20180528"}
   "mlgmpidl"
   "yojson" {>= "1.6.0"}
-  "zarith" {>= "1.4"}
+  "zarith" {>= "1.10"}
   "odoc" {with-doc}
 ]
 depopts: ["elina"]
 available: !(arch = "x86_32")
 build: [
-  ["./configure"]
+  ["./configure"] {os != "macosx"}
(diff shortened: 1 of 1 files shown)


antoinemine:
`macosx`, sorry... `macos` is working ;) 

(well, there are still some MacOS-specific C AST nodes and types we don't parse, but well...)

rmonat:
I'm curious, is there a way to require `opam >= 2.1` to avoid the last issue we have in the CI?

antoinemine:
It seems that we should add `opam-version >= "2.1.0"` in the `available` .

antoinemine:
Also: at some point, the opam file was generated by `dune-project`. This is no longer the case (`(generate_opam_files true)` in `dune-project`) and we modified by hand the once-generated opam file.
It seems that we can revert to an automatically generating the opam file at build if we put everything that dune cannot generate into a `mopsa.opam.template`, which it will append to the generated file. This would include `depexts` and `available`, but also possibly `build` and `install`. I don't know if there is a big benefit in doing it that way.

rmonat:
@mseri I would be glad to have your input here: I've listed the `cygwin` os as not available in ab70276, but the windows CI still triggers.  Am I missing something?

mseri:
Windows is complicated and the CI is still experimental.
Similarly almost no system uses opam < 2.1, the test is there mainly for us to check that there is no regression in the solver.

For the time being I would not add the `available` field, instead I suggest that I accept the failure in the CI.

rmonat:
Thank you for your feedback! Closing in favor of #26259

PR #26253, opened 2024-07-17
ctypes prior to 0.4.0 doesn't support Windows

dra27:
Windows support was added in https://github.com/yallop/ocaml-ctypes/pull/190 in 0.4.0.

Putting this in a separate PR rather than in suggesting in https://github.com/ocaml/opam-repository/pull/26233 in order to limit the revdeps.


mseri:
Thanks

PR #26255, opened 2024-07-17
Windows CI tweaks

dra27:
Four tweaks to the GitHub Actions workflow being used for Windows testing:
- While reviewing #26233, there were failures on old packages because the workflow attempted to downgrade OCaml to a version before 4.13.0. 4.12 and earlier have not (yet) been updated with Windows support. I've added a switch invariant to the workflow so that these don't fail but instead correctly report that they can't be installed.
- I also noticed that the compiler was rebuilding, which made me realise that CI is still running an opam 2.2.0 beta, so I added a display of the version number in the step where opam is added to the cache
- At the moment, the workflow will never update opam - always download install.ps1 and use the hash of that to compute the cache key
- Finally, install.ps1 is switched from a debugging script to download from opam master instead (which updates it from 2.2.0~beta3 to the 2.2.0 release)

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -18,6 +18,11 @@ jobs:
       - name: Checkout tree
         uses: actions/checkout@v4
 
+      - name: Download install.ps1
+        run: |
+          [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+          (New-Object System.Net.WebClient).DownloadFile("https://raw.githubusercontent.com/kit-ty-kate/opam/windows-installer.debug/shell/install.ps1", ".\install.ps1")
+
       - name: Restore opam cache
         id: cache-opam
         uses: actions/cache/restore@v4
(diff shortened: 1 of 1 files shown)


raphael-proust:
I don't see anything wrong with this, but I also don't have any experience of `.ps1` files at all so I don't really feel like I can merge.

dra27:
The .ps1 it's downloading is the "official" opam one (the equivalent of `install.sh`) - I'm not sure its content matters in terms of reviewing here? The semantic change is that the expression before:

```powershell
Invoke-Expression "& { $(Invoke-RestMethod https://raw.githubusercontent.com/kit-ty-kate/opam/windows-installer.debug/shell/install.ps1) } -OpamBinDir 'D:\opam\bin' -NoSetPath -NoAdmin"
```
which is just the equivalent of `bash -c "sh <(curl -fsSL https://raw.githubusercontent.com/ocaml/opam/master/shell/install.sh)`. has been split into a separate download and execute (so `Invoke-RestMethod` becomes a `System.Net.WebClient.DownloadFile` call).

That then means that every run of CI will check for an updated `install.ps1` script from opam and, if it has changed (which it will on release, because the new release URL gets added) then the hash will change and invalidate the cache, causing a new binary to be downloaded.

mseri:
Sounds good, thanks

PR #26263, opened 2024-07-20
ocaml-lsp-server: remove avoid-version now ocaml.5.2.0 is out

avsm:
Without this, attempting to `opam install ocamlformat ocaml-lsp-server` is resulting in the compiler being downgraded to 5.1.1 rather than install the right LSP server in 5.2.0

Noticed by @yminsky and me as part of the Gitpod template-ocaml install process

### DIFF
--- a/packages/ocaml-lsp-server/ocaml-lsp-server.1.18.0~5.2preview/opam
+++ b/packages/ocaml-lsp-server/ocaml-lsp-server.1.18.0~5.2preview/opam
@@ -46,8 +46,6 @@ depends: [
   "merlin-lib" {>= "5.1"}
   "ocaml-index" {>= "1.0" & post}
 ]
-flags: avoid-version
-available: opam-version >= "2.1.0"
 dev-repo: "git+https://github.com/ocaml/ocaml-lsp.git"
 build: [
   ["dune" "subst"] {dev}


mseri:
This is the preview version, I would rather not change the avoid-version. There should be a non-preview release that supports 5.2

avsm:
But it's the only version available for 5.2.0! Surely it's preferable that `opam install lsp-server` installs _a_ version of lsp-server as opposed to downgrading my entire compiler to 5.1.1 automatically. This breaks users very unexpectedly...

mseri:
But why? Wasn't `avoid-version` made to prevent automatic choice of the installation but allow to still be able to install the package? The request to downgrade seems a bug if that is the case

Ping @kit-ty-kate that has more experience with both opam and these preview packages.

avsm:
I suspect `avoid-version` is working as intended -- opam has avoided installing the latest version of ocaml-lsp-server, and found a version which it doesnt need to avoid on ocaml-5.1.1. And given that there's no formula constraint on a default opam init holding it there, it's made a valid selection.

What i don't understand is why we're using avoid-version on lsp-server at all: a simple version constraint on `>=5.2 & < 5.3` should suffice for the preview one, as it should never be selected on an earlier version of the compiler anyway.

kit-ty-kate (approved):
for lsp it indeed doesn't matter as it's already fixed to a given version. The addition of the field was more mechanical on my part, i'll remember for next time.

PR #26271, opened 2024-07-23
[new release] reason (2 packages) (3.12.0)

anmonteiro:
Reason: Syntax & Toolchain for OCaml

- Project page: <a href="https://reasonml.github.io/">https://reasonml.github.io/</a>

##### CHANGES:

- Add `\u{hex-escape}` syntax (@anmonteiro,
  [reasonml/reason#2738](https://github.com/reasonml/reason/pull/2738))
- Support local open and let bindings (@SanderSpies) [reasonml/reason#2716](https://github.com/reasonml/reason/pull/2716)
- outcome printer: change the printing of `@bs.*` to `@mel.*` (@anmonteiro, [reasonml/reason#2755](https://github.com/reasonml/reason/pull/2755))
- Fix outcome printing of optional arguments on OCaml 5.2 (@anmonteiro, [reasonml/reason#2753](https://github.com/reasonml/reason/pull/2753))
- support parsing and printing of `external%extension` (@anmonteiro, [reasonml/reason#2750](https://github.com/reasonml/reason/pull/2750), [reasonml/reason#2766](https://github.com/reasonml/reason/pull/2766), [reasonml/reason#2767](https://github.com/reasonml/reason/pull/2767))
- install `refmt` manpage (@anmonteiro, [reasonml/reason#2760](https://github.com/reasonml/reason/pull/2760))
- add support for parsing / printing of refutation clause in `switch` (@anmonteiro, [reasonml/reason#2765](https://github.com/reasonml/reason/pull/2765))
- support `let%ppx` in signatures (@anmonteiro, [reasonml/reason#2770](https://github.com/reasonml/reason/pull/2770))

PR #26278, opened 2024-07-24
Fix quickjs dep constraint in server-reason-react

raphael-proust:
There is a failure in the CI of https://github.com/ocaml/opam-repository/pull/26271 which seem to be related to server-reason-react.

When checking the revdeps `server-reason-react.0.3.0`, the following type mismatch happens.

```
# File "packages/melange.js/Js.ml", line 496, characters 44-73:
# 496 |   let fromString : string -> t = fun str -> Quickjs.RegExp.compile str ""
#                                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
# Error: This expression has type
#          "(Quickjs.RegExp.t,
#           [ `Invalid_escape_sequence
#           | `Malformed_unicode_char
#           | `Nothing_to_repeat
#           | `Unexpected_end
#           | `Unknown of string ] * string)
#          Stdlib.result"
#        but an expression was expected of type "t" = "Quickjs.RegExp.t"
```

ping @davesnx

### DIFF
--- a/packages/server-reason-react/server-reason-react.0.3.0/opam
+++ b/packages/server-reason-react/server-reason-react.0.3.0/opam
@@ -11,7 +11,7 @@ depends: [
   "reason" {>= "3.10.0"}
   "melange" {>= "3.0.0"}
   "ppxlib" {> "0.23.0"}
-  "quickjs" {>= "0.1.1"}
+  "quickjs" {>= "0.1.2"}
   "promise" {>= "1.1.2"}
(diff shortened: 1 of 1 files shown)


raphael-proust:
Doesn't seem to fix the error… I took a wild guess based on the constraint for the constraints in the next release of `server-reason-react` but that's not it apparently. Closing, but anyone should feel free to provide a fix if they know what's up.

PR #26282, opened 2024-07-25
Fix quickjs dep constraint in server-reason-react 0.3.0

davesnx:
Fixing https://github.com/ocaml/opam-repository/pull/26278


mseri:
Thanks

PR #24159, opened 2023-07-26
Package forester.2.2

jonsterling:
### `forester.2.2`
A tool for tending mathematical forests

---
* Homepage: https://github.com/jonsterling/ocaml-forester
* Source repo: git+https://github.com/jonsterling/ocaml-forester.git
* Bug tracker: https://github.com/jonsterling/ocaml-forester/issues

---
:camel: Pull-request generated by opam-publish v2.2.0

PR #26300, opened 2024-07-28
Package forester.4.2.0

jonsterling:
### `forester.4.2.0`
A tool for tending mathematical forests

---
* Homepage: https://sr.ht/~jonsterling/forester/
* Source repo: git+https://git.sr.ht/~jonsterling/ocaml-forester
* Bug tracker: https://todo.sr.ht/~jonsterling/forester
* Release notes: http://www.jonmsterling.com/jms-00WK.xml
---

### DIFF
--- a/packages/forester/forester.4.2.0/opam
+++ b/packages/forester/forester.4.2.0/opam
@@ -1,0 +1,52 @@
+opam-version: "2.0"
+synopsis: "A tool for tending mathematical forests"
+maintainer: "Jonathan Sterling"
+authors: "Jonathan Sterling"
+license: "GPL-3.0-or-later"
+homepage: "https://sr.ht/~jonsterling/forester/"
+bug-reports: "https://todo.sr.ht/~jonsterling/forester"
+depends: [
+  "menhir" {>= "20240715"}
+  "ocamlgraph" {>= "2.1.0"}
+  "ocaml" {>= "5.1.1"}
+  "dune" {>= "3.13"}
+  "ppx_deriving"
+  "cmdliner" {>= "1.2.0"}
+  "dune-build-info"
+  "uucp" {>= "15.1.0"}
+  "eio_main" {>= "1.1"}
+  "ptime" {>= "1.1.0"}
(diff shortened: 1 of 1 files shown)


jonsterling:
Damn, it looks like these issues are cropping up again: https://github.com/ocaml/opam-repository/pull/24159 + https://github.com/jonsterling/ocaml-forester/commit/f1b8b6da040164bb9d481867d15a3759ba03019c

I really am at a loss...

avsm:
Looks very green now, so merging! You may want to announce this on <https://discuss.ocaml.org>, where we have a Community category and an `announce` tag for this purpose.

jonsterling:
Thanks for the encouragement! I've posted an announcement here: https://discuss.ocaml.org/t/ann-forester-4-2/15043.

Regarding the build issues I experienced in multiple forester releases (linked) when checking lower bounds, each time it was seemingly caused by some subtle behaviour of Menhir; running the reproduction commands via Docker as instructed in the build log takes at least an hour on my machine, and after that I could not figure out what is necessary to test whatever change I might make to resolve the problem without needing another hour — considering that many attempts might be necessary as I pinpoint the issue, it seemed to me that I was being asked to spend at least several hours... The edit-to-debug ratio involved in that process is too extreme for me to undertake it on a regular basis, as seems to be required, so I will in the future be removing my dependency on Menhir. 

I understand the problem may ultimately stem from a complicated interaction between Menhir's versioning practices and my own, so it may not ultimately be any one entity's fault — but life is too short for me to find out, so I plan to just hand-write my parsers in the future.

avsm:
You're also not obliged to use a range of Menhir versions. If you've only tested on a particular version for a given release of Forester, you can constrain the dependency here to that exact version with a `=`, and that's totally fine too.

jonsterling:
That is a fair point, but I am trying to be a good citizen in the package ecosystem, and making Forester uninstallable (without a fresh switch) for users who happen to have an only slightly different release of menhir installed seems to be a bit brutal for my users.

PR #26302, opened 2024-07-28
Package git_split.0.0.1

Artamus:
### `git_split.0.0.1`

---

---


packages/git_split/git_split.0.0.1/opam
```diff
@@ -0,0 +1,41 @@
+# This file is generated by dune, edit dune-project instead
+opam-version: "2.0"
+synopsis: "A tool to split a git commit into multiple"
+description:
+  "When one commit becomes too large and needs to be split up, the general git workflow that involves resetting the commit and then using interactive staging to generate multiple commits from it can be quite cumbersome. In addition to that, the git interactive staging tool's UX can be quite bad, therefore this tools aims to provide this functionality in a nicer way."
+maintainer: ["Artamus"]
```
mseri, line 6: It could be useful to add a way to contact you here

packages/git_split/git_split.0.0.1/opam
```diff
@@ -0,0 +1,41 @@
...
+homepage: "https://github.com/Artamus/git-split"
+bug-reports: "https://github.com/Artamus/git-split/issues"
+depends: [
+  "ocaml"
+  "dune" {>= "3.15"}
+  "re"
+  "notty"
+  "feather"
+  "re"
+  "odoc" {with-doc}
```
mseri, line 18:
```suggestion
  "ppx_deriving"
  "odoc" {with-doc}
```

packages/git_split/git_split.0.0.1/opam
```diff
@@ -0,0 +1,43 @@
...
+  "dune" {>= "3.15"}
+  "re"
+  "notty"
+  "feather" {>= "0.3.0"}
+  "re"
+  "ppx_deriving"
+  "alcotest" {with-test}
+  "odoc" {with-doc}
+]
+build: [
```
mseri, line 22:
```suggestion
conflicts: [ "result" {< "1.5"} ]
build: [
```

Should fix the `Result.bind` issue in the lower bounds

Artamus: Thank you so much for your time @mseri!
Can you explain in a few words why this is happening? Is it a conflict of some "shim" library that exists for OCaml versions without the stdlib Result that is causing problems?

mseri:
Old versions of this library (pulled in by some of the dependencies) were shadowing the additional functions introduced in the stdlib. Adding this conflict prevents those versions to be installed.

packages/git_split/git_split.0.0.1/opam
mseri: Old versions of this library (pulled in by some of the dependencies) were shadowing the additional functions introduced in the stdlib. Adding this conflict prevents those versions to be installed.

Artamus: Ah, I see that I could also have just stuck with 4.14 and it would have been fine. But I'll keep at 4.13 for this version.

mseri:
Thanks!

PR #26313, opened 2024-07-30
Make spotify_ml.push unavailable, the version should have been 0.0.1

mseri:
@raphael-proust I think we should revert the original PR or remove the package altogether.

in the meantime merging this as a fix

### DIFF
--- a/packages/spotify_ml/spotify_ml.push/opam
+++ b/packages/spotify_ml/spotify_ml.push/opam
@@ -32,6 +32,7 @@ build: [
     "@doc" {with-doc}
   ]
 ]
+available: false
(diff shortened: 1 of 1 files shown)


mseri:
Ping @maxRN (I saw the discussion on discort)

maxRN:
Thank you! @mseri

maxRN:
Is there anything else I need to do? I'm fine with deleting the version.

maxRN:
@mseri Sorry to bother you again about this, but the wrong (old) version is still showing as latest on ocaml.org and opam.ocaml.org. Do you know if this is a bug with those sites not respecting the `available` property?

PR #26314, opened 2024-07-31
[new release] uring-trace (2 packages) (0.1.0)

koonwen:
IO-uring tracing tool using eBPF probes

- Project page: <a href="https://github.com/koonwen/uring-trace">https://github.com/koonwen/uring-trace</a>
- Documentation: <a href="https://github.com/koonwen/uring-trace">https://github.com/koonwen/uring-trace</a>

##### CHANGES:

- Initial release.

### DIFF
--- a/packages/conf-liburing/conf-liburing.0.1.0/opam
+++ b/packages/conf-liburing/conf-liburing.0.1.0/opam
@@ -1,12 +1,13 @@
 opam-version: "2.0"
 synopsis: "Virtual package for system installation of liburing"
-maintainer: ["Lee Koon Wen"]
+maintainer: ["koonwen@gmail.com"]
 authors: ["Lee Koon Wen"]
 license: ["ISC" "BSD-3-Clause"]
 homepage: "https://github.com/koonwen/uring-trace"
 doc: "https://koonwen.github.com/koonwen/uring-trace"
 bug-reports: "https://github.com/koonwen/uring-trace/issues"
 available: [ os = "linux" ]
+dev-repo: "git+https://github.com/koonwen/uring-trace.git"
 
 depexts: [
   ["liburing-dev"]   { os-distribution = "ubuntu" }
@@ -14,12 +15,4 @@ depexts: [
   ["liburing-devel"] { os-distribution = "fedora" }
 ]
 flags: conf
-url {
-  src:
-    "https://github.com/koonwen/uring-trace/releases/download/v0.1.0/uring-trace-0.1.0.tbz"
-  checksum: [
-    "sha256=d1469550363df1af0fcb331ac86f6347fb021dfa844532c370317d3ad30d1e91"
-    "sha512=73a519f681d60aa83c4cba8439fc94fe207b0449ef0bff1fef0e976c4f68d6e56f543444cab936910d3e00cb9436abe2828739cc1e27c28d2e8c9d1e9218b1e8"
-  ]
-}
 x-commit-hash: "732869041422ada630d065f937ba210c20b8bf5c"
--- a/packages/conf-liburing/conf-liburing.0.1.0/opam
+++ b/packages/conf-liburing/conf-liburing.0.1.0/opam
@@ -1,0 +1,25 @@
+opam-version: "2.0"
+synopsis: "Virtual package for system installation of liburing"
+maintainer: ["Lee Koon Wen"]
+authors: ["Lee Koon Wen"]
+license: ["ISC" "BSD-3-Clause"]
+homepage: "https://github.com/koonwen/uring-trace"
+doc: "https://koonwen.github.com/koonwen/uring-trace"
+bug-reports: "https://github.com/koonwen/uring-trace/issues"
+available: [ os = "linux" ]
+
+depexts: [
+  ["liburing-dev"]   { os-distribution = "ubuntu" }
+  ["liburing-dev"]   { os-distribution = "debian" }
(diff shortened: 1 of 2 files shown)


koonwen:
This tool has a tight dependency on the linux kernel version to work as intended. It should be built with kernels >= 6.1 but I had no way to express this in the opam file. I guess a workaround is to enforce constraints on the OS versions? Although users could still be running different kernel versions underneath.

packages/conf-liburing/conf-liburing.0.1.0/opam
```diff
@@ -0,0 +1,25 @@
...
+  ["liburing-devel"] { os-distribution = "fedora" }
+]
+flags: conf
+url {
+  src:
+    "https://github.com/koonwen/uring-trace/releases/download/v0.1.0/uring-trace-0.1.0.tbz"
+  checksum: [
+    "sha256=d1469550363df1af0fcb331ac86f6347fb021dfa844532c370317d3ad30d1e91"
+    "sha512=73a519f681d60aa83c4cba8439fc94fe207b0449ef0bff1fef0e976c4f68d6e56f543444cab936910d3e00cb9436abe2828739cc1e27c28d2e8c9d1e9218b1e8"
+  ]
```
shonfeder, line 23: Afaiu, `conf-` are not meant to include sources, but just contain the depexts. That is why this lint check is failing in our CI:

```
error 46: Package is flagged "conf" but has source, install or remove instructions
```

Is this source actually needed here?

packages/conf-liburing/conf-liburing.0.1.0/opam
```diff
@@ -0,0 +1,25 @@
+opam-version: "2.0"
+synopsis: "Virtual package for system installation of liburing"
+maintainer: ["Lee Koon Wen"]
```
shonfeder, line 3: Please add a contact here and in the uring-trace package (e.g., email address).

packages/conf-liburing/conf-liburing.0.1.0/opam
```diff
@@ -0,0 +1,25 @@
...
+authors: ["Lee Koon Wen"]
+license: ["ISC" "BSD-3-Clause"]
+homepage: "https://github.com/koonwen/uring-trace"
```
shonfeder, line 6: To fix the missing lint

```
warning 37: Missing field 'dev-repo'
```

I think you may just add

```suggestion
dev-repo: "https://github.com/koonwen/uring-trace"
```

here and in the other package.

shonfeder (commented):
Hello! :wave:

I'm not sure of the best way to handle the kernel constraint, but you you could at least

- Make this requirement explicit in the package description.
- Add [`post-message`](https://opam.ocaml.org/doc/Manual.html#opamfield-post-messages) warning users of this requirement.

Additional, perhaps you could add a test that checks the kernel version, or checks functionality that depends on this, with a clear error message communicating the requirement if it fails.

Any better ideas @mseri ?

koonwen:
Hey @shonfeder, thanks for the review!

I've made most of the changes you've suggested except for the test for the kernel version since the build failure should now show the post message. I'll probably add this in the next minor version though to make the source of the error more clear. I'm guessing it'll need to be a script that's pre-run before the package attempts to build?

packages/uring-trace/uring-trace.0.1.0/opam
```diff
@@ -0,0 +1,53 @@
...
+homepage: "https://github.com/koonwen/uring-trace"
+doc: "https://github.com/koonwen/uring-trace"
+bug-reports: "https://github.com/koonwen/uring-trace/issues"
+depends: [
+  "ppx_deriving" {>= "6.0.2"}
+  "eio" {>= "1.1"}
+  "ocaml"
+  "dune" {>= "3.13"}
+  "dune-site"
+  "cmdliner"
```
shonfeder, line 16: Noticed in the CI that we are installing

```
-> retrieved cmdliner.1.0.4  (cached)
```

on some platforms.

But you are using the `Cmdliner.Cmd` module which is only available from `1.1.0`, so we need:

```suggestion
  "cmdliner" {>= "1.1.0"}
```

This reflected in the lowerbound errors that turned up after I reran the failing CI jobs (to clear up some transient failures):

```
compilers

    5.1
        uring-trace.0.1.0
            [lower-bounds (failed: Unbound module Cmd)](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/9141711b2bae6019cf673865e320c305bd9e7ab9/variant/compilers,5.1,uring-trace.0.1.0,lower-bounds)
    5.2
        uring-trace.0.1.0
            [lower-bounds (failed: Unbound module "Cmd")](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/9141711b2bae6019cf673865e320c305bd9e7ab9/variant/compilers,5.2,uring-trace.0.1.0,lower-bounds)
```

shonfeder (commented):
Awesome! It's looking good to me :) I think I've found just one other problem with the lower bounds on cmdliner.

shonfeder (approved):
Looks great to me. Thanks for making this available on opam!

shonfeder:
For repo maintainers:

The few outstanding CI failures are expected build failures on unsupported kernels, afaik. In each case the expected `post-message` is printed:

```
=> This package requires your kernel version to be at least 6.1.0
```

mseri:
Thanks!

PR #26316, opened 2024-07-31
conf-automake: add Cygwin support

rjbou:
I've added Cygwin support to `conf-automake` in the same way it is done in `conf-autoconf`. I've tested it with opam 2.2 and a Cygwin install (`os-distribution = cygwin`), but I didn't tested the `cygwinports` part.


mseri:
Thanks, this is a strict improvement, any further fix can be added later from users

PR #26323, opened 2024-08-02
Add -mno-outline-atomics to fix static build on Ubuntu arm64

### LINKED ISSUE #26216: `opam switch create` with `static` option fails on Ubuntu aarch64
Consider this Dockerfile:

```docker
FROM ubuntu:latest

RUN apt update
RUN apt install -y --no-install-recommends ca-certificates git rsync opam musl-tools
RUN opam init -y --bare --disable-sandboxing
RUN opam switch create 5.2.0 --packages ocaml-option-static
```

On x64, it works perfectly fine.

However, on aarch64, `opam switch create` gives me the following error:

```
11.15 [ERROR] The compilation of ocaml-variants.5.2.0+options failed at "make -j9".
11.16
11.16 #=== ERROR while compiling ocaml-variants.5.2.0+options =======================#
11.16 # context              2.1.5 | linux/arm64 | ocaml-option-static.1 | https://opam.ocaml.org#f11abb65
11.16 # path                 ~/.opam/5.2.0/.opam-switch/build/ocaml-variants.5.2.0+options
11.16 # command              /usr/bin/make -j9
11.16 # exit-code            2
11.16 # env-file             ~/.opam/log/ocaml-variants-7-75a8ef.env
11.16 # output-file          ~/.opam/log/ocaml-variants-7-75a8ef.out
11.16 ### output ###
11.16 # make[1]: Entering directory '/root/.opam/5.2.0/.opam-switch/build/ocaml-variants.5.2.0+options'
11.16 # [...]
11.16 #   CC runtime/interp.b.o
11.16 #   CC runtime/startup_byt.b.o
11.16 #   CC runtime/zstd.b.o
11.16 #   MKLIB runtime/libcamlrun.a
11.16 #   MKEXE runtime/ocamlrun
11.16 # /usr/bin/ld: /usr/lib/gcc/aarch64-linux-gnu/13/libgcc.a(lse-init.o): in function `init_have_lse_atomics':
11.16 # (.text.startup+0xc): undefined reference to `__getauxval'
11.16 # collect2: error: ld returned 1 exit status
11.16 # make[1]: *** [Makefile:1309: runtime/ocamlrun] Error 1
11.16 # make[1]: Leaving directory '/root/.opam/5.2.0/.opam-switch/build/ocaml-variants.5.2.0+options'
11.16 # make: *** [Makefile:830: world.opt] Error 2
```

cknitt:
Fixes #26216.

Only made this change for the 5.x release versions for now. If the change is ok, I can extend the PR to cover all the other versions, too.

Note that the problem does not occur on Alpine arm64 (as it doesn't have glibc), so the condition could be extended to check for `os-distribution!="alpine"`, too.

### DIFF
--- a/packages/ocaml-variants/ocaml-variants.5.0.0+options/opam
+++ b/packages/ocaml-variants/ocaml-variants.5.0.0+options/opam
@@ -85,7 +85,8 @@ build: [
     "--enable-frame-pointers" {ocaml-option-fp:installed}
     "CC=cc" {!ocaml-option-32bit:installed & !ocaml-option-musl:installed & (os="openbsd"|os="macos")}
     "CC=musl-gcc" {ocaml-option-musl:installed & os-distribution!="alpine"}
-    "CFLAGS=-Os" {ocaml-option-musl:installed}
+    "CFLAGS=-Os" {ocaml-option-musl:installed & arch!="arm64"}
+    "CFLAGS=-Os -mno-outline-atomics" {ocaml-option-musl:installed & arch="arm64"}
     "LDFLAGS=-Wl,--no-as-needed,-ldl" {ocaml-option-leak-sanitizer:installed | (ocaml-option-address-sanitizer:installed & os!="macos")}
     "CC=gcc -ldl -fsanitize=leak -fno-omit-frame-pointer -O1 -g" {ocaml-option-leak-sanitizer:installed}
     "CC=gcc -ldl -fsanitize=address -fno-omit-frame-pointer -O1 -g" {ocaml-option-address-sanitizer:installed & os!="macos"}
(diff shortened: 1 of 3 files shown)


packages/ocaml-variants/ocaml-variants.5.0.0+options/opam
```diff
@@ -85,7 +85,8 @@ build: [
     "--enable-frame-pointers" {ocaml-option-fp:installed}
     "CC=cc" {!ocaml-option-32bit:installed & !ocaml-option-musl:installed & (os="openbsd"|os="macos")}
     "CC=musl-gcc" {ocaml-option-musl:installed & os-distribution!="alpine"}
-    "CFLAGS=-Os" {ocaml-option-musl:installed}
+    "CFLAGS=-Os" {ocaml-option-musl:installed & arch!="arm64"}
+    "CFLAGS=-Os -mno-outline-atomics" {ocaml-option-musl:installed & arch="arm64"}
```
raphael-proust, line 89: AFAICT this is going to break on Debian 10 because it ships a version of GCC that's too old.

Is this a non-issue for some reason? E.g., if `ocaml-option-musl:installed` is impossible on debian 10.

I don't think it's a hard blocker… and if it's a blocker then the proper fix would be to introduce some constraint on gcc versions which we maybe don't want to do anyway?

raphael-proust: Debian 10 is outside of LTS so let's say it's not a blocker. If people start having issues on debian 10 we'll see what to do.

PR #26325, opened 2024-08-02
[new release] provider (0.0.6)

mbarbin:
### `provider.0.0.6`
Parametrize your OCaml library with values that behave like objects but aren't

---
* Homepage: https://github.com/mbarbin/provider
* Source repo: git+https://github.com/mbarbin/provider.git
* Bug tracker: https://github.com/mbarbin/provider/issues

---
### Changed

- Reduce `provider` package dependencies - reduce from `base` to `sexplib0`.

### Fixed

- Make sure to select the right most implementation in case of overrides, as per specification.

### Removed

- Removed `Trait.Uid.Comparable.S` as this requires `Base`. Make it compatible with `Comparable.Make (Trait.Uid)` and add tests for this use case.

### DIFF
--- a/packages/provider/provider.0.0.6/opam
+++ b/packages/provider/provider.0.0.6/opam
@@ -1,0 +1,41 @@
+opam-version: "2.0"
+synopsis:
+  "Parametrize your OCaml library with values that behave like objects but aren't"
+maintainer: ["Mathieu Barbin"]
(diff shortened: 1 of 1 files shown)


mbarbin:
Please disregard for now - there is some inconsistencies with the dune version constraints as raised by the ci. I'm looking into it.

PR #26330, opened 2024-08-05
[new release] tar (4 packages) (3.0.0)

dinosaure:

##### CHANGES:

- Fix `Header.marshal` and the checksum and the length (@reynir, mirage/ocaml-tar#145)
- Delete a mutable field about the level into the header (@hannesm, mirage/ocaml-tar#141)
- **BREAKING**: de-functorize the package (@hannesm, @reynir, @dinosaure, mirage/ocaml-tar#140, mirage/ocaml-tar#143, mirage/ocaml-tar#146)

  These PRs attempt to de-functorize `Tar` so that users can implement I/O
  themselves, using `Tar`'s own element serialization/deserialization functions
  to take advantage of read/write methods. This avoids imposing on the user the
  implementation of a module that is too rigid in his/her case (which could have
  performance implications).

  `Tar` offers functions for serializing/deserializing tar-specific elements
  from `string`. It is then up to the user to know how to obtain or write these
  `strings`.

  To this, these PRs add "logics" (see `'a Tar.t`) requiring read and/or write
  implementations and describing how to extract all entries from a tar file or
  how to write a tar file according to a "dispenser" (like `Seq.to_dispenser`)
  of entries.

  These logics do not depend on a particular "scheduler", and these PRs propose
  a derivation of these logics with `tar-unix`, `tar-eio` and `tar-mirage`.
  These latter derivations mean that the API for these packages has only been
  extended, and there are no breaking changes as such.

  These logics also make it easy to offer a compression/decompression layer with
  `decompress`, so you can easily manipulate and/or create a .tar.gz file.

### DIFF
--- a/packages/tar-eio/tar-eio.3.0.0/opam
+++ b/packages/tar-eio/tar-eio.3.0.0/opam
@@ -1,0 +1,53 @@
+opam-version: "2.0"
+synopsis: "Decode and encode tar format files using Eio"
+description: """
+tar is a library to read and write tar files with an emphasis on
+streaming.  This library uses Eio to provide a portable tar library.
+"""
+maintainer: ["Reynir Björnsson <reynir@reynir.dk>" "dave@recoil.org"]
+authors: [
+  "Dave Scott"
+  "Thomas Gazagnaire"
+  "David Allsopp"
+  "Antonin Décimo"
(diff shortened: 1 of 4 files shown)


raphael-proust:
Thanks for the contribution!

PR #26339, opened 2024-08-06
Add m4 rule for macos

mtelvers:
Wouldn't this address https://github.com/ocaml/infrastructure/issues/151

### DIFF
--- a/packages/conf-m4/conf-m4.1/opam
+++ b/packages/conf-m4/conf-m4.1/opam
@@ -7,7 +7,10 @@ license: "GPL-3.0-only"
 build: [
   ["sh" "-exc" "echo | m4"] {os != "macos"}
   ["sh" "-exc" "echo | gm4"] {os = "macos"}
-] 
+]
+install: [
+  ["sh" "-c" "ln -s \"$(which gm4)\" \"%{bin}%/m4\""] {os-family = "homebrew"}
+]
 depexts: [
   ["m4"] {os-family = "debian"}
   ["m4"] {os-family = "ubuntu"}
@@ -19,8 +22,6 @@ depexts: [
   ["m4"] {os-family = "suse" | os-family = "opensuse"}
   ["m4"] {os-distribution = "ol"}
   ["m4"] {os-distribution = "arch"}
-  ["m4"] {os = "macos" & os-distribution = "homebrew"}
-  ["m4"] {os = "macos" & os-distribution = "macports"}
   # On FreeBSD m4 is part of the base system
 ]
 synopsis: "Virtual package relying on m4"
--- a/packages/conf-m4/conf-m4.1/opam
+++ b/packages/conf-m4/conf-m4.1/opam
@@ -4,7 +4,10 @@ homepage: "http://www.gnu.org/software/m4/m4.html"
 bug-reports: "https://github.com/ocaml/opam-repository/issues"
 authors: "GNU Project"
 license: "GPL-3.0-only"
-build: [["sh" "-exc" "echo | m4"]]
+build: [
+  ["sh" "-exc" "echo | m4"] {os != "macos"}
+  ["sh" "-exc" "echo | gm4"] {os = "macos"}
+] 
 depexts: [
   ["m4"] {os-family = "debian"}
   ["m4"] {os-family = "ubuntu"}
--- a/packages/conf-m4/conf-m4.1/opam
+++ b/packages/conf-m4/conf-m4.1/opam
@@ -16,6 +16,8 @@ depexts: [
   ["m4"] {os-family = "suse" | os-family = "opensuse"}
   ["m4"] {os-distribution = "ol"}
   ["m4"] {os-distribution = "arch"}
+  ["m4"] {os = "macos" & os-distribution = "homebrew"}
+  ["m4"] {os = "macos" & os-distribution = "macports"}
   # On FreeBSD m4 is part of the base system
 ]
 synopsis: "Virtual package relying on m4"
--- a/packages/conf-m4/conf-m4.1/opam
+++ b/packages/conf-m4/conf-m4.1/opam
@@ -9,7 +9,7 @@ build: [
   ["sh" "-exc" "echo | gm4"] {os = "macos"}
 ]
 install: [
-  ["sh" "-c" "ln -s \"$(which gm4)\" \"%{bin}%/m4\""] {os-family = "homebrew"}
+  ["sh" "-c" "ln -s \"$(which gm4)\" \"%{bin}%/m4\""] {os = "macos" & (os-distribution = "homebrew" | os-distribution = "macports")}
 ]
 depexts: [
   ["m4"] {os-family = "debian"}


mseri:
Good point.

It used to work a long time ago (maybe because it was needed by ocaml) and then started failing. I did not think about adding a dependency because usually m4 came directly with xcode.

mseri:
Seems like it is not woking properly:
```
#=== ERROR while compiling conf-m4.1 ==========================================#
# context              2.2.0 | macos/x86_64 | ocaml-base-compiler.4.14.2 | pinned
# path                 ~/.opam/4.14.2/.opam-switch/build/conf-m4.1
# command              ~/.opam/opam-init/hooks/sandbox.sh build sh -exc echo | m4
# exit-code            72
# env-file             ~/.opam/log/conf-m4-56626-d58bb8.env
# output-file          ~/.opam/log/conf-m4-56626-d58bb8.out
### output ###
# + echo
# + m4
# xcode-select: Failed to locate 'm4', and no install could be requested (perhaps no UI is present). Please install manually from 'developer.apple.com'.
```

mseri:
The error makes me think that something is off, I checked on my mac and `m4` is the one coming with the official cli tools and compiler: `/usr/bin/m4`

mseri:
I think this is wrong. We should not be messing with the users environment paths for the compilers.

On any opam user Mac, that conf package already works without having to do anything since the binary of m4 comes from Apple developer tools. We should not install an extra unnecessary copy just for it to work on the ci.

mtelvers:
It's very curious.  The command line tools are already installed as they provide the compiler and other tools.  We wouldn't get very far without them!

They are installed with `xcode-select --install`.  Running it again reports that they are already installed:

```
% xcode-select --install
xcode-select: note: Command line tools are already installed. Use "Software Update" in System Settings or the softwareupdate command line interface to install updates
```

`/usr/bin/m4` is just a stub which opens the Xcode installer.  I have run through the installation via that route but `m4` still isn't available.  I've tested on a worker and on my MacBook Pro and it hasn't worked in either case.

mtelvers:
Update: `m4` on my MacBook and on all the macOS workers is installed but it's called `gm4`.

All of these systems run macOS Sonoma.

```
% whereis gm4
gm4: /usr/bin/gm4 /Library/Developer/CommandLineTools/usr/share/man/man1/gm4.1
% gm4 --version
GNU M4 1.4.6
Copyright (C) 2006 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

Written by Rene' Seindal.
```

See also https://github.com/apple-oss-distributions/gm4

mseri:
Do you have xcode installed? I start thinking it may matter.

I tried on my desktop, where I only have the command line tools installed, and indeed it goes in a loop in the same way that you describe. I am sure m4 used to work in the past, this may be a bug that has now been going on for a while. And, even more annoying, they ship `gm4` as you point out above: `/usr/bin/gm4` works fine for me (without any extra package...).

Could we create an `m4` for mac which is a symlink to `/usr/bin/gm4` in the `.opam` bin folder? That may resolve the issue

mtelvers:
No, I don't have e Xcode installed, just the command line tools. I wonder if there has been a change in macOS. What version are you running?

We could create a symlink, but would that be representative of the end user environment?

If Apple have moved from `m4` to `gm4`, do we need packages to be updated to take that into consideration?

How does the compiler itself deal with this as that needs `m4` for `autoconf`?

mseri:
No idea, it seems to work seamlessly when I use opam

mseri:
In fact it seems to work fine also in the ci, at least for all, packages not explicitly calling m4

shonfeder:
> I think this is wrong. We should not be messing with the users environment paths for the compilers.
> 
> On any opam user Mac, that conf package already works without having to do anything since the binary of m4 comes from Apple developer tools. We should not install an extra unnecessary copy just for it to work on the ci. 
> 

This makes sense to me.

Fwiw, I didn't mean this to advocate for a particular solution, but just to help illuminate the problem space.

mseri:
Of course, sorry if it sounded negative. I was also wrong, since it does not work if xcode is not installed 😅

mseri:
I have been looking around, we are not the only ones noticing this: 
- https://github.com/DepoXy/depoxy/blob/dc9962422e4db2cb7b02c4013cad96b0ffcaf13d/bin/onboarder/install-ohmyrepos.sh#L172
- https://github.com/cirruslabs/cirrus-ci-docs/issues/1272
- https://github.com/facebookincubator/velox/issues/9190

mseri:
I checked on my mac what is going on. I have macports.

I asked chatGPT which suggested to grep from the autoconf code (it is a perl script). It calls `automa4te` whose code defines m4 as
```
my $m4 = $ENV{"M4"} || '/opt/local/bin/gm4';
```

Unless it is overwritten by the `M4` env variable, it uses `gm4` installed by macports.

mseri:
I am canceling the build on the CI since it passed on mac and it has infinitely many revdeps that we don't need to test

@mtelvers @shonfeder I suggest we merge this version. It would be nice to be able to set the `M4` env variable to `$(which gm4)` but I think we can see if this solves the issue for now or not. It is surely an improvement on the previous version. What do you think?

packages/conf-m4/conf-m4.1/opam
```diff
@@ -4,10 +4,10 @@ homepage: "http://www.gnu.org/software/m4/m4.html"
 bug-reports: "https://github.com/ocaml/opam-repository/issues"
 authors: "GNU Project"
 license: "GPL-3.0-only"
-build: [
-  ["sh" "-exc" "echo | m4"] {os != "macos"}
-  ["sh" "-exc" "echo | gm4"] {os = "macos"}
-] 
+build: [["sh" "-exc" "echo | m4"]]
+install: [
+  ["sh" "-c" "ln -s \"$(brew --prefix m4)/bin/m4\" \"%{bin}%/m4\""] {os-family = "homebrew"}
```
shonfeder: Nice. This looks like a very nice approach afaict.

shonfeder (approved):

mtelvers:
I've read the links.  I've failed to install GT with `opam install GT` (on a Mac).  I tried setting the environment variable `M4` and that doesn't work as m4 is called by name rather than by ${M4} in `src/dune`.

We either need developers to update their projects to expect `gm4` rather than `m4` on macOS, or we need to update the system/machine in such as way that it will still be applicable on the end users machine.

I've added an install section here which creates the symlink for `m4` in the opam switch pointing to `$(which gm4)`.  I think this will maximise the compatibility without requiring developers to update their projects nor does it need `brew install m4`.

packages/conf-m4/conf-m4.1/opam
mtelvers: Yes, I like it.  I've since updated it to use `gm4` as it occurred to me that we don't need the version from `brew`.

shonfeder:
This approach makes sense to me, prima facie. I will note that it conflicts with our policy around `conf-` packages, as per the linting rule:

```
2024-08-07 16:26.43: Error in conf-m4.1:              error 46: Package is flagged "conf" but has source, install or remove instructions
2024-08-07 16:26.58: Job failed: Error in conf-m4.1:              error 46: Package is flagged "conf" but has source, install or remove instructions
```

but I don't have enough context to know whether this change (which looks reasonable to me given the constraints) is actually a violation of the spirit of that rule.

mseri:
Let’s see how it goes 🤞

PR #26367, opened 2024-08-16
[new release] thread-local-storage (0.2)

c-cube:
Efficient thread local storage for OCaml

- Project page: <a href="https://github.com/c-cube/thread-local-storage">https://github.com/c-cube/thread-local-storage</a>
- Documentation: <a href="https://c-cube.github.io/thread-local-storage">https://c-cube.github.io/thread-local-storage</a>

##### CHANGES:

- Incompatible API change, making things simpler.
- It is now safe to read TLS from asynchronous callbacks (e.g. memprof
  callbacks).

### DIFF
--- a/packages/thread-local-storage/thread-local-storage.0.2/opam
+++ b/packages/thread-local-storage/thread-local-storage.0.2/opam
@@ -1,0 +1,37 @@
+opam-version: "2.0"
+synopsis: "Efficient thread local storage for OCaml"
+maintainer: ["Simon Cruanes"]
+authors: ["Vesa Karvonen"]
+license: "MIT"
+tags: ["topics" "threads" "tls"]
+homepage: "https://github.com/c-cube/thread-local-storage"
+doc: "https://c-cube.github.io/thread-local-storage"
+bug-reports: "https://github.com/c-cube/thread-local-storage/issues"
+depends: [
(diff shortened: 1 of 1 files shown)


c-cube:
breaking API, so the failures are expected.

avsm:
Looks good, but need to push upper bounds to the packages broken by the revdeps.

polytypic:
I like the overall idea of the simplified API and this is perhaps not the best place to comment, but I'd recommend some minor changes to the API.

In particular, I'd recommend defining a new non-parameterized exception constructor for the case when `get` has no value to return and then using `raise_notrace`.  This allows one to use such a `get` operation to simultaneously query for existence and to also obtain the value and to do so _without allocations_. I'd recommend renaming such a `get` to `get_exn` to signal to the caller that it uses exceptions for control flow:

```ocaml
match TLS.get_exn key with
| value -> ...
| exception TLS.Not_set -> ...
```

The use of a unique exception constructor also helps in cases where someone mistakenly doesn't use the above pattern.

I don't want to catch `Failure _` exceptions anywhere in my code (except in tests to check exceptions are raised as specified), because a `Failure _` is way too generic and even if code might be raising a`Failure _` today, changes in the future might cause different `Failure _`s to be raised.

PR #26361, opened 2024-08-15
Package opam-depext.1.2.3

kit-ty-kate:
### `opam-depext.1.2.3`
Install OS distribution packages
opam-depext is a simple program intended to facilitate the interaction between
OPAM packages and the host package management system. It can query OPAM for the
right external dependencies on a set of packages, depending on the host OS, and
call the OS's package manager in the appropriate way to install them.

---
* Homepage: https://github.com/ocaml/opam-depext
* Source repo: git+https://github.com/ocaml/opam-depext.git#2.0
* Bug tracker: https://github.com/ocaml/opam-depext/issues

---
:camel: Pull-request generated by opam-publish v2.3.1

PR #26369, opened 2024-08-16
opam-depext is not compatible with Windows

kit-ty-kate:
As described in https://github.com/ocaml/opam-repository/pull/26361#issuecomment-2293782445


kit-ty-kate:
More people are getting the error so emergency merge.

PR #26373, opened 2024-08-17
Package git-split.0.0.2

Artamus:
### `git-split.0.0.2`

---

---


Artamus:
Please pardon the two different users, unfortunately the automation does not deal super well with having multiple different ssh keys for different users on the same machine.

I actually changed the name of the package from `git_split` to `git-split`, because I'm still figuring out where dune allows hyphens and where must they be underscores.
I figured I would later send a PR that deletes the original `git_split@0.0.1` OPAM package, since there's practically no functional difference between the initial release and this one - please let me know if that is undesirable.

dinosaure:
> I figured I would later send a PR that deletes the original git_split@0.0.1 OPAM package, since there's practically no functional difference between the initial release and this one - please let me know if that is undesirable.

The policy suggests that no package should disappear from the repository - see https://github.com/ocaml/opam-repository/wiki/Policies#1-removal-of-packages-should-be-avoided. Cleaning and archiving phases are performed, but answer to another (longer) cycle. You can mark your old package as unavailable (`available: false`), which will help us later to consider it as an archive that absolutely nobody has installed.

shonfeder:
Looks good to me:

- Windows failure is due to notty, not this backage.
- The opam-ci failure is just a collision warning with the other package name.

avsm:

PR #26375, opened 2024-08-19
+ mldonkey 3.2.1

ygrek:


dinosaure:
It seems that `mldonkey` does not compile on FreeBSD and MacOS. You can fix on these system your package or mark them as unavailable on these system if you want:
```
available: [ os-distribution != "freebsd" & os != "macos" ]
```

ygrek:
thanks, updated

PR #26377, opened 2024-08-20
[new release] elpi (1.19.6)

gares:

##### CHANGES:

- Runtime:
  - Fix bug in unification code for "automatic intro" (as in the intro tactic)
    of UVar arguments

### DIFF
--- a/packages/elpi/elpi.1.19.6/opam
+++ b/packages/elpi/elpi.1.19.6/opam
@@ -1,0 +1,90 @@
+opam-version: "2.0"
+maintainer: "Enrico Tassi <enrico.tassi@inria.fr>"
+authors: [ "Claudio Sacerdoti Coen" "Enrico Tassi" ]
+license: "LGPL-2.1-or-later"
+homepage: "https://github.com/LPCIC/elpi"
+doc: "https://LPCIC.github.io/elpi/"
+dev-repo: "git+https://github.com/LPCIC/elpi.git"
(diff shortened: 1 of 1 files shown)


dinosaure:
It seems that your tests fail on 32-bits architecture (eg. `x86_32` & `arm32`). Not sure how to interpret that but you can mark your package as unavailable on 32-bits system or fix your tests on this kind of machine.

gares:
I believe these tests always failed. If the policy for accepting the package changed I can blacklist these architectures, but they don't seem bug specific to elpi at all (eg. elpi uses no threads for example).

dinosaure:
> I believe these tests always failed. If the policy for accepting the package changed I can blacklist these architectures, but they don't seem bug specific to elpi at all (eg. elpi uses no threads for example).

We probably should add something like `with-test & os != "macos" & os-distribution != "alpine" & os-distribution != "freebsd" & arch = "x86_64" | arch = "arm64"` to avoid the test on 32-bits architecture. WDYT?

gares:
I'm fine with that change

mseri:
In this case I am going to merge as is. The issue relates to the use of resources in some machines and we saw it in other packages.

We did not disable the tests at the time because further investigation on the issue was needed and was not strictly a problem of the package. There was a discussion in the past on how to fix them for good but was not very short term, I'll try to find it.

gares:
thanks!

PR #26385, opened 2024-08-21
Package mccs.1.1+18

kit-ty-kate:
### `mccs.1.1+18`
MCCS (which stands for Multi Criteria CUDF Solver) is a CUDF problem solver
developed at UNS during the European MANCOOSI project

---
* Homepage: https://www.i3s.unice.fr/~cpjm/misc/
* Source repo: git+https://github.com/ocaml-opam/ocaml-mccs.git
* Bug tracker: https://github.com/ocaml-opam/ocaml-mccs/issues

---

### DIFF
--- a/packages/mccs/mccs.1.1+18/opam
+++ b/packages/mccs/mccs.1.1+18/opam
@@ -1,0 +1,36 @@
+opam-version: "2.0"
+synopsis: """\
+MCCS (which stands for Multi Criteria CUDF Solver) is a CUDF problem solver
+developed at UNS during the European MANCOOSI project"""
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
The failure in the revdeps (opam-devel) is due to a change in the CI: https://github.com/ocurrent/opam-repo-ci/pull/343 (cc @shonfeder @mtelvers) which sets `OPAM_SOLVER` to something that opam 2.0 doesn't know about and that make it fail.
I was able to reproduce the issue locally but it went away after removing that environment variable.

I would suggest opam-repo-ci use `--solver` instead of setting that environment variable

dinosaure:
Thanks, the revdep is fixed now.

PR #26389, opened 2024-08-22
[new release] miou (0.3.0)

dinosaure:
Composable concurrency primitives for OCaml

- Project page: <a href="https://git.robur.coop/robur/miou">https://git.robur.coop/robur/miou</a>
- Documentation: <a href="https://docs.osau.re/miou/">https://docs.osau.re/miou/</a>

##### CHANGES:

- Set the internal pipe used to interrupt a domain to a non-blocking mode and
  remove the usage of an atomic which protect how we fill the pipe
  (@haesbaert, @dinosaure, robur-coop/miou#28)
- Expose option to reuse addr/port when we `Miou_unix.bind_and_listen`
  (@ada2k, @dinosaure, robur-coop/miou#27)
- Protect an illegal access to the orphan from a possibly parallel task which
  does not own the orphan value
  (@poytypic, @dinosaure, robur-coop/miou#31, robur-coop/miou#32)
- Be able to pin a specific domain when we want to launch a parallel task
  (@dinosaure, robur-coop/miou#34)
- Expose the `Miou.Backoff` module which can be useful for users
  (@dinosaure, robur-coop/miou#35)
- Fix or improve (from the maintainance point-of-view) the `Miou.Queue` module
  and some internal parts of Miou about the usage of atomics
  (@dinosaure, @polytypic, robur-coop/miou#36, robur-coop/miou#33)
- Prefer to require a `finaliser` function for the `events` value and actually
  close the internal `Unix.pipe` used to interrupt domain than to use
  `Gc.finaliser` and possibly leak file-descriptors
  (spotted by @hannesm, @dinosaure, robur-coop/miou#37)

### DIFF
--- a/packages/miou/miou.0.3.0/opam
+++ b/packages/miou/miou.0.3.0/opam
@@ -1,0 +1,37 @@
+opam-version: "2.0"
+maintainer:   "Romain Calascibetta <romain.calascibetta@gmail.com>"
+authors:      "Romain Calascibetta <romain.calascibetta@gmail.com>"
+homepage:     "https://git.robur.coop/robur/miou"
+bug-reports:  "https://git.robur.coop/robur/miou/issues"
+dev-repo:     "git+https://github.com/robur-coop/miou.git"
+doc:          "https://docs.osau.re/miou/"
+license:      "MIT"
+synopsis:     "Composable concurrency primitives for OCaml"
+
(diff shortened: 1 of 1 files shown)


hannesm:
looks good to me, thanks a lot. you may want to announce at https://discuss.ocaml.org :)

PR #26391, opened 2024-08-22
Fix revdeps about tcpip.8.2.0

dinosaure:
Add an upper-bound about `tcpip.8.2.0` which breaks the API (and functors needed).


hannesm:
looks fine to me

PR #26417, opened 2024-08-24
[new release] hardcaml-lua (alpha+31)

jrrk2:
CHANGES:

* Tweak packaging syntax


jrrk2:
That flambda compiler seems to be faulty and should be removed from the regression suite

PR #26423, opened 2024-08-25
[new release] tls-miou-unix (1.0.1)

hannesm:
CHANGES:

* tls-miou-unix: fix file descriptor leak (mirleft/ocaml-tls#508 @dinosaure)
* tls-miou-unix: fix fuzz test (mirleft/ocaml-tls#507 @dinosaure)


dinosaure:
Thanks!

PR #26425, opened 2024-08-26
[new release] provider 0.0.8

mbarbin:
Going directly from `0.0.4` to `0.08` - skipping internal releases published in my custom opam-repository.

I added a commit to allow `>= sexplib.0.17` in the hope things will be compatible going forward - this may be adjusted later.

Detailed changelog below: 

### Fixed

- Follow-up fixes to binding renaming.

### Added

- Added dedicated tests for internal functions.

### Changed

- More renamings. `Implementation => Binding`, `Interface => Handler`. This brings the lib closer to the original namings from Eio.

### Changed

- Rename `Provider.Trait.Implementation` as `Provider.Implementation` to expose the concepts in a more flat way.
- Reduce `provider` package dependencies - reduce from `base` to `sexplib0`.

### Fixed

- Make sure to select the right most implementation in case of overrides, as per specification.

### Removed

- Removed `Trait.Uid.Comparable.S` as this requires `Base`. Make it compatible with `Comparable.Make (Trait.Uid)` and add tests for this use case.

### Added

- Added dependabot config for automatically upgrading action files.

### Changed

- Upgrade `ppxlib` to `0.33` - activate unused items warnings.
- Upgrade `ocaml` to `5.2`.
- Upgrade `dune` to `3.16`.
- Upgrade `eio` to `1.0` (no change required).
- Upgrade base & co to `0.17`.

### DIFF
--- a/packages/provider/provider.0.0.8/opam
+++ b/packages/provider/provider.0.0.8/opam
@@ -1,0 +1,40 @@
+opam-version: "2.0"
+synopsis: "Dynamic Dispatch with Traits"
+maintainer: ["Mathieu Barbin"]
+authors: ["Mathieu Barbin"]
+license: "ISC"
+homepage: "https://github.com/mbarbin/provider"
+doc: "https://mbarbin.github.io/provider/"
+bug-reports: "https://github.com/mbarbin/provider/issues"
+depends: [
+  "dune" {>= "3.16"}
+  "ocaml" {>= "5.2"}
+  "bisect_ppx" {dev & >= "2.8.3"}
+  "ppx_js_style" {dev & >= "v0.17" & < "v0.18"}
+  "sexplib0" {>= "v0.17" & < "v0.18"}
+  "odoc" {with-doc}
+]
+build: [
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
+dev-repo: "git+https://github.com/mbarbin/provider.git"
+url {
+  src:
+    "https://github.com/mbarbin/provider/releases/download/0.0.8/provider-0.0.8.tbz"
+  checksum: [
+    "sha256=433b981763a0734e107890d04368315969422440e6a845c01bfc0bf7c24d680d"
+    "sha512=da903fb935b9ae6eff6b75168eddeb6d67f66be13221d6c705d532723d7bc8fd8ff54324a28bf070ce38aae42c322f07ee7232f3a822bff3ed67186139171b9e"
+  ]
+}
+x-commit-hash: "b3164ad630b2dbf5ee20965e4f6653afeb5392d1"
--- a/packages/provider/provider.0.0.8/opam
+++ b/packages/provider/provider.0.0.8/opam
@@ -10,8 +10,8 @@ depends: [
   "dune" {>= "3.16"}
   "ocaml" {>= "5.2"}
   "bisect_ppx" {dev & >= "2.8.3"}
-  "ppx_js_style" {dev & >= "v0.17" & < "v0.18"}
-  "sexplib0" {>= "v0.17" & < "v0.18"}
+  "ppx_js_style" {dev & >= "v0.17"}
+  "sexplib0" {>= "v0.17"}
   "odoc" {with-doc}
 ]
 build: [


mbarbin:
Looking at the CI, I saw mdx test failing. I did not expect them to be run - I think there was something inconsistent in the provider repo between the dependencies for the main package and the package attached to the mdx stanza in the documentation.

I think it's probably best not to attach the documentation (and mdx page compilation) to the main package.

I created a new version (0.0.9) where I attempted to fix this, and attached the mdx stanza to a test package instead. https://github.com/mbarbin/provider/commit/395c07380bb33a247220e4378a3e25d6e7a3f5ab

I updated this PR with this new version to see if this fixes the CI failures. This new version `0.0.9` cancels and replace `0.0.8`. I updated the title of the PR but don't know how to rename the branch (I think this doesn't matter). Alternatively, I can create a new separate PR if you prefer. Thank you.

packages/provider/provider.0.0.9/opam
```diff
@@ -0,0 +1,40 @@
...
+authors: ["Mathieu Barbin"]
+license: "ISC"
+homepage: "https://github.com/mbarbin/provider"
+doc: "https://mbarbin.github.io/provider/"
+bug-reports: "https://github.com/mbarbin/provider/issues"
+depends: [
+  "dune" {>= "3.16"}
+  "ocaml" {>= "5.2"}
+  "bisect_ppx" {dev & >= "2.8.3"}
+  "ppx_js_style" {dev & >= "v0.17"}
```
mseri, line 13:
```suggestion
  "bisect_ppx" {with-dev-setup & >= "2.8.3"}
  "ppx_js_style" {with-dev-setup & >= "v0.17"}
```
Are you sure this is `dev` and not [`with-dev-setup`](https://opam.ocaml.org/doc/Manual.html#pkgvar-with-dev-setup)? The latter is the flag to install "developer tools"

mseri:
Just a small comment, otherwise it looks good

packages/provider/provider.0.0.9/opam
mseri: Afaik the `with-dev-setup` flag was introduced for linting and instrumentation, for the people that wanted to add a dependency to ocamlformat or ocp-indent for example. The history for this is here: https://github.com/ocaml/opam/issues/4959

The CI will not install these and run the linting steps, in the same way as it does not do it with dev (`dev` is only true if a user wants to install the head of the master branch of a opam package with `opam install --dev ...` and is never true for releases).

> Given that I also have a test package in the same repo, is it important that I add these with-dev-setup section to the main package, or is that enough to simply list them as dependencies to the test package?

I am not sure what you mean. They can stay also in the main package, it depends on how you run the tests and the lint. To use them in your CI is enough to make sure that opam is called with the `--with-dev-setup` flag

mbarbin: Thanks a lot @mseri ! 

I updated the repo to use `--with-dev-setup` and added `ocamlformat` there as well. https://github.com/mbarbin/provider/pull/10

Given that there are no other changes, I also reported the changes to the opam file directly in this PR. Is that OK, even if technically the opam file from the 0.0.9 archive will not be exactly equal to the opam file checked into the opam repo for that version? (or is it cleaner to cancel this release and create a new `0.0.10` release?).

I took the opportunity to apply the same change to other repos, as well as a [template](https://github.com/mbarbin/opam-package-template) I use. Thank you for your help!

mbarbin: > I am not sure what you mean. They can stay also in the main package, it depends on how you run the tests and the lint. To use them in your CI is enough to make sure that opam is called with the --with-dev-setup flag

What I mean is this: In the provider repo, I have two opam files.
- [provider.opam](https://github.com/mbarbin/provider/blob/main/provider.opam)
- [provider-tests.opam](https://github.com/mbarbin/provider/blob/main/provider-tests.opam)
The CI for that project now includes [--with-dev-setup](https://github.com/mbarbin/provider/blob/3a46d62cb848066d98b403915179747e74cbfe06/.github/workflows/ci.yml#L34)

I am guessing that when you are at the root of the repo and run:
```sh
$ opam install . --with-doc --with-test --with-dev-setup
```
it is enough for the dev dependencies to be listed in any of the present opam files to be picked up by the install command. Thus, my question is : what are the pros of including the dev-setup dependencies of `provider.opam`. Is there a context in which not having them there is going to make something fail?

mseri: Yes that will install the dev-setup and test dependencies for all the packages. If you only need it for development and not for the main package, then you can just have them all in the `-tests` package

mseri:
I am fine to merge as is, or to remove them, whatever you prefer. In either case we can do it from here and you don't need to re-tag or re-release

mbarbin:
Got it, thanks a lot @mseri. I am fine to merge as is - and I'll think about it for the next release whether I want to tweak it further as discussed. Thank you for the help!

mseri:
Thanks

PR #26429, opened 2024-08-26
[new release] dns (13 packages) (9.0.0)

dinosaure:
An opinionated Domain Name System (DNS) library

- Project page: <a href="https://github.com/mirage/ocaml-dns">https://github.com/mirage/ocaml-dns</a>
- Documentation: <a href="https://mirage.github.io/ocaml-dns/">https://mirage.github.io/ocaml-dns/</a>

##### CHANGES:

* Remove `Cstruct.t` and use `string`/`bytes` (@hannesm, @dinosaure, @palainp, mirage/ocaml-dns#351)
* Add a Miou implementation of `ocaml-dns` (@dinosaure, @hannesm, mirage/ocaml-dns#352)

PR #26432, opened 2024-08-26
Add some upper-bounds about letsencrypt and dns.9.0.0

dinosaure:
To be able to merge #26429


hannesm:
thanks. there may as well be upper bounds for letsencrypt-dns to dns < 9.0.0

PR #26393, opened 2024-08-22
Update the out-of-date opam warning after the release of opam 2.2.1

kit-ty-kate:
To merge once the Discuss post is published

PR #26435, opened 2024-08-27
Improve INFO and WARNING for the opam version messages

mseri:
Suggested by @dra27 on https://github.com/ocaml/opam-repository/pull/26393

### DIFF
--- a/repo
+++ b/repo
@@ -3,10 +3,10 @@ browse: "https://opam.ocaml.org/pkg/"
 upstream: "https://github.com/ocaml/opam-repository/tree/master/"
 announce: [
 """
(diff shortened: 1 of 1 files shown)


dra27 (approved):
Feels strange to "approve" my own suggestion, but @rjbou had also :+1:'d it!

hannesm:
I now get the following warning with opam 2.1.6:
```
[default] no changes from https://opam.ocaml.org
default (at https://opam.ocaml.org): 
    [WARNING] opam %{opam-version}% is out-of-date. Please consider updating it (https://opam.ocaml.org/doc/Install.html)
```

I find this unexpected and confusing.

mseri:
I agree. I'll send a PR to revert it. @rjbou is string interpolation not supported in the announce opam strings?

PR #26436, opened 2024-08-27
arrayjit, neural_nets_lib: upper bind ppx_minidebug

lukstafi:
in preparation for a non-backward-compatible release: ppx_minidebug 2.0.

### DIFF
--- a/packages/arrayjit/arrayjit.0.3.3/opam
+++ b/packages/arrayjit/arrayjit.0.3.3/opam
@@ -27,7 +27,7 @@ depends: [
   "ppxlib"
   "ppx_jane"
   "ppx_expect"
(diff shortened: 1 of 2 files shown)


lukstafi:
Note: Windows CI failure due to `conf-libgccjit`.
(Going forward, arrayjit 0.4 has only an optional dependency on libgccjit, so this will no longer be a problem.)

The following CI targets also fail due to `conf-libgccjit` (for some peculiar reason). 
- debian-testing-ocaml-5.2
- debian-unstable-ocaml-5.2

All other CI targets pass (3 stragglers looking like they'll pass).

mseri:
Thanks

PR #26448, opened 2024-08-28
mdx: doesnt compile under ocaml 5.3.0-dev

avsm:
fails with:
Error: This constant has type "string" but an expression was expected of type "Unit_info.t option"


mseri:
Thanks

PR #26493, opened 2024-09-05
frama-clang Version 0.0.16

vprevosto:


mseri:
I think there is not much we can do right now to improve the situation. I would suggest to merge as it is and if we figure out a way to harmonize conf-llvm and conf-clang, then we proceed to do it separatelu

shonfeder:
Is this a problem people can address when trying to install on their own machines? If so is this something we should/could improve in the meantime with a post message on failure? That way, if this fails to install for some people, they will at least have tips on the way forward?

raphael-proust:
IIRC we had similar issues in the past and essentially depending on the os, then some or-dependencies had to be one way or another. We could fix this by adding constraints in the dependencies but it's hard to figure out the right set.

Ideally, users would let us know about failures and how they solved them (e.g., "using this distro, had to first install this dep, and then could successfully install that package") and we could add the correct dependencies based on those reports. But we haven't received such reports AFAIK.

I say we merge as is and improve things on-demand.

PR #26496, opened 2024-09-05
Mark u2f unavailable

reynir:
The U2F API has been deprecated and removed in most browsers (IIRC Firefox removed their API in July 2023 and Google Chrome in 2022), and the library is now unmaintained.


reynir:
It's possible to migrate to [webauthn](https://github.com/robur-coop/webauthn), and maybe it could be nice to include a message about this? I don't know.

avsm:
Rather than making it unavailable (the code itself could be useful to others), you could add a `messages` field with the deprecation notice in there, and also add the pointer to webauthn.  This way, programs that used the `u2f` library will still remain installable but the deprecation warning will show up whenever they are installed.  Consider, for example, an application that used u2f as a non-essential part of its operation -- we wouldnt want to make that entirely uninstallable despite this API being deprecated

mseri:
Using `flags: deprecated` and a message field will inform the users of the deprecation. In a few months, once we will start the slimming of the repo, it will be moved to the opam archive and disappear from here

mseri:
Thanks, I'll merge the deprecation + message, this should give any remaining user the time to switch.

PR #26497, opened 2024-09-05
[new release] mirage (2 packages) (4.7.0)

hannesm:
The MirageOS library operating system

- Project page: <a href="https://github.com/mirage/mirage">https://github.com/mirage/mirage</a>
- Documentation: <a href="https://mirage.github.io/mirage/">https://mirage.github.io/mirage/</a>

##### CHANGES:

- Adapt bounds for mirage-crypto-rng-mirage {>= "1.0.0"}, tls {>= "1.0.0"},
  conduit {>= "7.0.0"}, dns {>= "9.0.0"}, happy-eyeballs {>= "1.2.0"},
  tcpip {>= "8.2.0"}, mirage-qubes {>= "1.0.0"}, git {>= "3.17"}.
  The changes are in mirage-crypto to use string instead of cstruct.t
  (mirage/mirage#1559 @hannesm)

PR #26439, opened 2024-08-28
[new release] re (1.12.0)

rgrinberg:
RE is a regular expression library for OCaml

- Project page: <a href="https://github.com/ocaml/ocaml-re">https://github.com/ocaml/ocaml-re</a>

##### CHANGES:

* Add `Re.split_delim` (ocaml/ocaml-re#233)
* Fix handling of empty matches in splitting and substitution functions (ocaml/ocaml-re#233)
* Add support for character classes in `Re.Posix` (ocaml/ocaml-re#263)

PR #26502, opened 2024-09-05
conan: restrict re to < 1.12.0

hannesm:
observed in https://github.com/ocaml/opam-repository/pull/26497 //cc @dinosaure @rgrinberg https://github.com/ocaml/opam-repository/pull/26439

```
== ERROR while compiling conan.0.0.5 ========================================#
context              2.3.0~alpha~dev | linux/x86_64 | ocaml-base-compiler.4.14.2 | file:///home/opam/opam-repository
path                 ~/.opam/4.14/.opam-switch/build/conan.0.0.5
command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p conan -j 39
exit-code            1
env-file             ~/.opam/log/conan-8-2d12a7.env
output-file          ~/.opam/log/conan-8-2d12a7.out
 # output ###
(cd _build/default && /home/opam/.opam/4.14/bin/ocamlc.opt -w -40 -g -bin-annot -I src/.conan.objs/byte -I /home/opam/.opam/4.14/lib/ptime -I /home/opam/.opam/4.14/lib/re -I /home/opam/.opam/4.14/lib/re/pcre -I /home/opam/.opam/4.14/lib/seq -I /home/opam/.opam/4.14/lib/uutf -intf-suffix .ml -no-alias-deps -open Conan__ -o src/.conan.objs/byte/conan__Serialize.cmo -c -impl src/serialize.ml)
File "src/serialize.ml", line 82, characters 6-10:
82 |       !lst;
           ^^^^
Error: This expression has type Re__Cset.c list
       but an expression was expected of type int list
       Type Re__Cset.c is not compatible with type int
```

### DIFF
--- a/packages/conan/conan.0.0.5/opam
+++ b/packages/conan/conan.0.0.5/opam
@@ -11,7 +11,7 @@ doc: "https://mirage.github.io/conan/"
 bug-reports: "https://github.com/mirage/conan/issues"
 depends: [
   "ocaml"               {>= "4.08.0"}
-  "re"                  {>= "1.11.0"}
+  "re"                  {>= "1.11.0" & < "1.12.0"}
   "dune"                {>= "2.9.0"}
   "uutf"
   "ptime"


hannesm:
merging, lint is fine.

rgrinberg:
I suppose this is the culprit: https://github.com/mirage/conan/blob/main/src/serialize.ml#L76

We could introduce something for `cset` in `Re.View` so that this hack isn't necessary

dinosaure:
> We could introduce something for cset in Re.View so that this hack isn't necessary

From my point of view, it seems to me that this part of re should not be exported and can potentially break between releases (without notifying a breaking change). When I started using this part, I accepted the fact that conan could break between re releases and that I shouldn't bother you (but take all the load) to ensure backward compatibility between conan and re.
As such, work on re is not necessarily necessary :) but would be welcome!

hannesm:
FWIW https://github.com/mirage/conan/pull/34

PR #26501, opened 2024-09-05
Luv 0.5.14: async I/O

aantron:
From the [changelog](https://github.com/aantron/luv/releases/tag/0.5.14):

> Bugs fixed
>
> - `const`-ness wrapper was harmful on GCC 14 and ctypes 0.23.0 (aantron/luv#159, reported Olaf Hering).
> - Need quotes in path expansion during build (aantron/luv#160, @tobil4sk).

PR #26508, opened 2024-09-07
[ci] Add cygwin bin directories to PATH

tobil4sk:
Without these PATH entries, cygwin's binaries never get used. This means that `make` and `bash` and other executables are not used from cygwin, but from other preinstalled packages in the github actions runner (e.g. git bash and the preinstalled mingw).

This causes build issues with certain packages, see:
https://github.com/ocaml/opam-repository/pull/26501
https://github.com/aantron/luv/issues/162

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -41,6 +41,8 @@ jobs:
         run: |
           D:\opam\bin\opam --version
           "D:\opam\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+          "D:\opamroot\.cygwin\root\usr\x86_64-w64-mingw32\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+          "D:\opamroot\.cygwin\root\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
 
       - name: Init opam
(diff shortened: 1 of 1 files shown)


dra27:
Something else is going on here - it shouldn’t be necessary to do this in a package build - I’ll endeavour to have a look tomorrow/next week

tobil4sk:
This seems somewhat related to https://github.com/ocaml/opam/issues/6181. To me the idea solution seems to be for `opam env`/`opam exec` to take care of setting the `PATH` correctly so that it is not necessary to set the `PATH` manually.

tobil4sk:
Should mingw-w64-shims also add `/usr/x86_64-w64-mingw32/bin` to PATH maybe? It already adds `/usr/x86_64-w64-mingw32/sys-root/mingw/bin`: https://github.com/dra27/mingw-w64-shims/blob/3c4a82700d9b03443ad6f2927449f7a8f25a77cf/mingw-w64-shims.opam#L8

The problem is that the `luv` library builds a native library using autotools, and the configure script expects an `ar` executable to be available in PATH. Since `/usr/x86_64-w64-mingw32/bin` isn't present in PATH, the build fails (both in CI and on local systems where the PATH hasn't been configured like this).

Does adding this to `mingw-w64-shims` sound like a reasonable idea? If so, I could try to send a patch there instead and close this PR.

PR #26521, opened 2024-09-09
Package git-split.0.0.3

Artamus:
### `git-split.0.0.3`

---

---


Artamus:
Unfortunately the windows CI fail is because of a dependency (notty) that does not work on windows.

mseri:
How to deal with Windows failures is a bit tricky still and honestly I don't yeat understand it, but we consider the CI experimental still for that and we can live with the failure.

Thanks a lot

PR #26526, opened 2024-09-10
Add: qrc.0.2.0

dbuenzli:
* Add: `qrc.0.2.0` [home](https://erratique.ch/software/qrc), [doc](https://erratique.ch/software/qrc/doc), [issues](https://github.com/dbuenzli/qrc/issues)  
  *QR code encoder for OCaml*

---

#### `qrc` v0.2.0 2024-09-10 Zagreb

- `Qrc.Matrix.to_svg`, change matrix rendering strategy. The result doesn't
  use xlink (no risk of identifier clashes when embedding), renders faster in
  browsers and is more compact ([#1](https://github.com/dbuenzli/qrc/issues/1)). Thanks to Alain Frisch for the patch.

- `qrtrip` tool. Fix `stdin` reading bug, newlines were being dropped
  and arbitrary binary input was impossible ([#5](https://github.com/dbuenzli/qrc/issues/5)).

- `qrtrip` tool. Incompatible change to fix the footgun UI. Before,
  the optional positional argument was the message to QR encode, but
  it feels natural to specify a file to QR encode. In which case you
  would silently QR encode a file path… The optional argument is now a
  file to read by default. Use option `-m/--message` to treat the
  argument as the message and make the tool operate as it used to.

- Require OCaml 4.14.0.

---

Use `b0 -- .opam publish qrc.0.2.0` to update the pull request.

### DIFF
--- a/packages/qrc/qrc.0.2.0/opam
+++ b/packages/qrc/qrc.0.2.0/opam
@@ -1,0 +1,42 @@
+opam-version: "2.0"
+synopsis: "QR code encoder for OCaml"
+description: """\
+Qrc encodes your data into QR codes. It has built-in QR matrix
+renderers for SVG, ANSI terminal and text.
+
+Qrc is distributed under the ISC license. It has no dependencies.
+
+Homepage: https://erratique.ch/software/qrc"""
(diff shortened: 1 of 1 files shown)


mseri:
Thanks

PR #26543, opened 2024-09-12
Make the 5.3+ compiler packages depend on base-effects (effect syntax)

kit-ty-kate:
While fixing https://github.com/ocaml-omake/omake/pull/159 in opam-repository, i realised that package already conflicts with `base-effects` (the meta package supposed to show the presence of the `effect` syntax)
But that package is not linked with OCaml >= 5.3 which now has the `effect` keyword and syntax, so this PR makes OCaml >= 5.3 compilers depend on that meta package


Octachron (approved):
That's look good to me, thanks for the fix !

avsm:
Good catch, thanks!

PR #26555, opened 2024-09-12
+angstrom.0.16.1

seliopou:


avsm:
xrefing to https://github.com/mirage/ocaml-uri/issues/178 -- this includes the Lazy-safety fix. Thanks @seliopou !

dinosaure:
The revdeps is huge but the change seems right. Consider for merging but waiting the CI to see of everything is ok :+1:.

avsm:
It’s good I think - I pushed stronger checksums but the previous revdeps run looked ok.

PR #26585, opened 2024-09-18
Remove parallelism warning from 4.14+

dra27:
While clearing the decks for something more important, these diffs were irritating.

The parallelism warning has been mostly removed (note that several of the +options packages from recent releases do _not_ include it). It's highly unlikely to be the root cause, and the suggestion just causes the poor user to waste time and energy on a sequential build.

While here, there are three trivial updates to `+trunk` packages when building on MSYS2.

Changes to `post-messages` should not trigger rebuilds 🤞


raphael-proust:
> Changes to post-messages should not trigger rebuilds 🤞

"should not" as in "I don't think it will" or "should not" as in "wouldn't it be nice if opam didn't"? I.e., should we open an issue on `opam`?

mseri:
Thanks

PR #26177, opened 2024-07-02
OCaml compiler patches for GCC 14

mtelvers:
With the release of GCC 14, which is included in Fedora 40 and the latest Arch Linux releases, we are seeing a number of failures to build older releases of OCaml, specifically < 4.08.  See https://github.com/ocurrent/docker-base-images/issues/279.  The configuration scripts for the compiler prior to 4.08 test for the presence of various library functions by attempting to compile minimal C programs using these functions.  These test builds now fail because GCC 14 makes implicit function declarations an error rather than a warning.  The additional compiler flag `-Wno-implicit-function-declaration` is needed.

This is not the first example of a patch being necessary to support the old compiler.  Previously, we have had the `-fno-common` patch required for GCC 10 compatibility.

Working on this issue with @dra27, and accepting that future patches will be required to maintain the older compilers, this PR set a precedence for holding compiler patches in [ocaml-opam/ocaml](https://github.com/ocaml-opam/ocaml) using the branches `x.yy-patches`.

The first commit updates the existing GCC 10 patch into this format, and the second commit adds the GCC 14 patch.

Issue #25819, opened 2024-05-08
Switches are configured with `OCAML_TOPLEVEL_PATH` but the variable name  is `OCAMLTOP_INCLUDE_PATH` 


dbuenzli:
I was trying to change `omod` and `down` to stop pooping `omod.top` and `down.top` into `ocaml`'s library directory and leave it for `$(opam var lib)/toplevel` to handle it.

This should be possible since OCaml 4.08.0 when https://github.com/ocaml/ocaml/issues/7808 was resolved by @nojb's https://github.com/ocaml/ocaml/pull/1841. 

Unfortunately the compilers packages [set the `OCAML_TOPLEVEL_PATH`](https://github.com/ocaml/opam-repository/blob/1d393046209a3bf9a41284cde888d54879545c5e/packages/ocaml/ocaml.5.2.0/opam#L18) instead of `OCAMLTOP_INCLUDE_PATH` so that doesn't work.

Not sure exactly who's responsible for opam compiler packages so I tentatively cc @Octachron and @dra27.

```sh
# With a down modified to no longer install `down.top` in $(opam var lib:ocaml)

> OCAML_TOPLEVEL_PATH=/Users/dbuenzli/.opam/4.14.2/lib/toplevel ocaml -noinit 
OCaml version 4.14.2
Enter #help;; for help.

# #use "down.top";;
Cannot find file down.top.
# ^D
> OCAMLTOP_INCLUDE_PATH=/Users/dbuenzli/.opam/4.14.2/lib/toplevel ocaml -noinit
OCaml version 4.14.2
Enter #help;; for help.

# #use "down.top";;
Down v0.3.0-5-g8eef77b loaded. Type Down.help () for more info.
# 

```

dra27:
> Not sure exactly who's responsible for opam compiler packages so I tentatively cc @Octachron and @dra27.

Suggests that I really should move on getting the precise value in https://github.com/dra27/opam-repository/commit/7813e70e155ac3a9c11927ab40b620d7741117d4 sorted out and upstreamed 🙂

dbuenzli:
@dra27 do you have an ETA on that one ? I should likely do an `omod` and `down` release and I would have liked to stop installing stuff in OCaml's library directory.

dra27:
[...]

PR #26594, opened 2024-09-20
Packages for OCaml 5.3.0~alpha1

dra27:
This PR adds ocaml-base-compiler.5.3.0\~alpha1 and ocaml-variants.5.3.0\~alpha1+options.

As a trial for this release cycle, in a similar vein to #17541, there is a small tweak to the compiler's packaging, intended to reduce the maintenance burden in future. The change should be almost completely transparent to users - please either install `ocaml.5.3.0`, `ocaml-base-compiler.5.3.0~alpha1`, or `ocaml-variants.5.3.0~alpha1+options` as you would have done before.

The restructuring introduces a new package ocaml-compiler.5.3.0~alpha1. The intention of this package is similar to ocaml-config, in that it carries the source tarball, dependencies and build instructions for the compiler. This package becomes a single source of truth for a given release of the compiler and means that when changes such as #16722, #19855 and #26177 come along in future only _one_ package per release has to be updated. It also reduces the number of packages to which configuration improvements need to be added such as in #26323.

In terms of review, if one compares the _current_ packages/ocaml-variants/ocaml-variants.5.3.0+trunk/opam with packages/ocaml-compiler/ocaml-compiler.5.3.0\~alpha1/opam, the only pertinent difference is that the new ocaml-compiler packages do _not_ have the `conflict-class` field. The new ocaml-base-compiler.5.3.0\~alpha1 and ocaml-variants.5.3.0\~alpha1 packages then just have to depend on ocaml-compiler.5.3.0\~alpha1 and retain the `conflict-class` field. Note that ocaml-base-compiler.5.3.0\~alpha1 still depends on ocaml-options-vanilla.  That package conflicts all the ocaml-option-\* packages, and so all the `:installed` tests being false reduces the myriad options passed to `configure` to be equivalent to those passed in ocaml-base-compiler.5.2.0 with one exception. `ASPP=cc -c` is no longer passed to `configure` for OpenBSD and macOS, however the default in OCaml's `configure` has already catered for this since OCaml 4.08 (so there is no semantic difference).

The ocaml-variants.5.3.0+trunk package is updated to use ocaml-compiler.5.3. This version ("5.3") of ocaml-compiler is "cunningly" picked as it is _older_ than release versions, which avoids any temptation of opam to try to upgrade to it. It's necessary to use the structure for ocaml-variants.5.3.0+trunk now as, uniquely, during a release cycle we have avoid-version set on the alpha releases which otherwise makes ocaml-variants.5.3.0+trunk more tempting to the solver, because it involves one package installation fewer.

I have tested this packaging locally on Linux and Windows (mingw-w64 and msvc64) with `opam switch create ocaml-5.3 --formula '"ocaml" {>= "5.3" & < "5.4"}'` and verified that in each case ocaml-base-compiler.5.3.0~alpha1 is the default selected package.

Finally, there is a minor tweak to ocaml.5.3.0 to allow the use of pre-release ocaml-base-compiler packages and in the environment commands the beginnings of a fix for #25819 (on which more soon).

### DIFF
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.5.3.0~alpha1/opam
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.5.3.0~alpha1/opam
@@ -1,0 +1,29 @@
+opam-version: "2.0"
+license: "LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception"
+synopsis: "First alpha release of OCaml 5.3.0"
+maintainer: [
+  "David Allsopp <david@tarides.com>"
+  "Florian Angeletti <florian.angeletti@inria.fr>"
+]
+authors: [
+  "Xavier Leroy"
+  "Damien Doligez"
+  "Alain Frisch"
+  "Jacques Garrigue"
+  "Didier Rémy"
+  "KC Sivaramakrishnan"
+  "Jérôme Vouillon"
+]
+homepage: "https://ocaml.org"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+dev-repo: "git+https://github.com/ocaml/ocaml.git#5.3"
+depends: [
+  "ocaml-compiler" {= "5.3.0~alpha1"}
+
+  "ocaml-beta" {opam-version < "2.1.0"}
+
+  # OCaml with default configuration (no flambda, TSAN, etc.)
+  "ocaml-options-vanilla" {post}
+]
+conflict-class: "ocaml-core-compiler"
+flags: [ compiler avoid-version ]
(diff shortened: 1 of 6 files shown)


kit-ty-kate (approved):

avsm:
Thanks!

Octachron:
Thanks for the opam packages! I will send the announce for the alpha release once the opam repository has been updated.

kit-ty-kate:
> I will send the announce for the alpha release once the opam repository has been updated.

this is now the case

Octachron:
Thanks for the ping ! I have sent the announces.

PR #26601, opened 2024-09-22
[new release] ppx_deriving_jsonschema (0.0.1)

Khady:
Jsonschema generator for ppx_deriving

- Project page: <a href="https://github.com/ahrefs/ppx_deriving_jsonschema">https://github.com/ahrefs/ppx_deriving_jsonschema</a>
- Documentation: <a href="https://ahrefs.github.io/ppx_deriving_jsonschema/">https://ahrefs.github.io/ppx_deriving_jsonschema/</a>

##### CHANGES:

- Initial alpha with basic features + @name @key @ref


m-laniakea:
dune-release's check of the maintainer field is incomplete. 
Something like `khady@ahrefs.com` would do.

Then the packages likely need more detailed constraints.

Khady:
I've changed the constraint on ppxlib to `"ppxlib" {>= "0.30.0"}`. I think that I could put a limit lower than that, but I don't know of an easy way to check except by actually installing all the older versions one by one.

Khady:
I got it back all the way down to 0.24.0. I think that's enough.

avsm:

PR #26610, opened 2024-09-24
Add dependency on an OCaml package for `semaphore-compat`

Leonidas-from-XIV:
Dune's package management is more strict about dependencies so package that don't depend on `ocaml` don't have an OCaml compiler available.

This PR adds the dependency.


mseri:
Thanks

Issue #26596, opened 2024-09-20
OCaml 5.3.0 release readiness


Octachron:
This issue tracks the build and readiness information of core platform tools and libraries in preparation for the OCaml 5.3 release. It’s intended as a meta-issue and a home for TODO lists and links to other milestones/issues rather than general discussion (which is welcomed on Discuss or on linked issues).

Packages that are fixed upstream, have a PR ready, or can be used as-is for testing purpose, are in [kit-ty-kate/opam-alpha-repository](https://github.com/kit-ty-kate/opam-alpha-repository).  The latest state of the opam ecosystem can be seen [in opam-health-check](http://check.ocamllabs.io) and compared against OCaml 5.3.

## Platform tools:

- [x] opam
- [x]  dune
- [x]  merlin
- [x]  lsp-server / vscode-plugin 
- [x]  odoc
- [x]  ocamlformat
- [x]  ocp-indent

## Ecosystem:

- [x]  ocamlfind
- [x]  ocamlbuild
- [x]  ppxlib
- [x]  mdx
- [x]  utop
- [x]  dune-release
- [x]   opam-publish

## Infrastructure:

- [ ] ocaml.org
- [ ] docker base containers 
- [ ] ocaml-ci
- [ ] opam-repo-ci

dbuenzli:
OCaml 5.3 has been released.

PR #26614, opened 2024-09-25
Allow frame pointers on macOS from 5.3 onwards.

tmcgilchrist:
OCaml 5.3.0 includes the option for compiling with frame pointers on macOS x86_64 [change log entry](https://github.com/ocaml/ocaml/blob/5.3/Changes#L88-L89).

It can be tested locally using these commands to create a switch:
```
opam repo add tmcgilchrist https://github.com/tmcgilchrist/opam-repository.git\#macos_frame_pointers
opam switch create 5.3.0~alpha1+fp ocaml-variants.5.3.0~alpha1+options ocaml-option-fp
```

Relates to https://github.com/ocaml/opam-repository/issues/26596


dra27 (approved):
LGTM, thanks. This support is added in 5.3 via ocaml/ocaml#13163.

PR #26615, opened 2024-09-25
Add `ocaml` dependency to `rusage`

Leonidas-from-XIV:
Compiling rusage requires an OCaml compiler, thus this PR adds that dependency.

The reasoning is the same as for #26610.

Upstream PR here: https://github.com/craigfe/ocaml-rusage/pull/3


Leonidas-from-XIV:
Also includes a `license` field with the value taken from the `LICENSE` file in the upstream repo to make the linter happy.

mseri:
Thanks, looks like it compiles with all ocaml version in the CI

Leonidas-from-XIV:
Yes, I was positively surprised I didn't need any constraints at all. It doesn't seem to work on Windows which is probably unsurprising, given it's some POSIX call and there's only limited support on Windows. Thanks for merging.

Issue #26604, opened 2024-09-23
Discussing the release of a `loc` package


mbarbin:
I am using a type to represent a range of lexing positions from a file, which I called `Loc.t`. It's [here](https://github.com/mbarbin/loc).

I use it as dependency to decorate AST nodes built by parsers, enabling located error messages during file processing (compilers, interpreters, linters, refactor tools, etc.).

I am considering submitting it to the opam-repository, but I have some reservations.

- Firstly I am unsure how to package it for publication. Would there be push-back about grabbing the global "loc" name on opam?
- Secondly, this is very similar to Dune's `Stdune.Loc` module (and actually uses `stdune` under the hood). Perhaps it would make more sense to work on extracting the Dune one as a foundational package (such as `pp`, `dyn`, and `ordering`), rather than maintaining this one separately.

My motivation is that I have other libraries that I'd like to eventually publish to opam that depend on it. Some examples below:

- [file-rewriter](https://github.com/mbarbin/file-rewriter)
- cmdlang's [Err](https://github.com/mbarbin/cmdlang/blob/main/lib/err/src/err.mli) module

Similar libraries or modules:

- Dune's [Loc.t](https://github.com/ocaml/dune/blob/dedfb76837bad6ea6de97de07e61f2ac10442127/otherlibs/stdune/src/loc.mli#L1)

- Parsexp's [ranges](https://github.com/janestreet/parsexp/blob/14af9ab942251783de6abb20e0d0e0eec6080062/src/positions.mli#L59)

- Ppxlib's [Location.t](https://github.com/ocaml-ppx/ppxlib/blob/456d1a99d354f8b3f34f01d1a7b61dc43ef678b6/src/location.ml#L4)

- Compilerlibs's [loc](https://github.com/ocaml/ocaml/blob/81003f23918729fa730c01ccffec25939cd730ff/utils/warnings.mli#L23)

I would appreciate any thoughts or feedback on this matter. Thank you!

mbarbin:
I've asked the question about Dune to the Dune devs here https://github.com/ocaml/dune/discussions/10948

avsm:
Thanks for all your work in establishing consensus here, @mbarbin. From the opam-repo perspective, we have no objection...

mbarbin:
Thank you. With this feedback, I feel comfortable opening #26625 for review.

I will be closing this issue then, the Dune discussion part can stay open as needed.

Thanks a lot!

PR #26625, opened 2024-09-26
[new release] loc (0.2.0)

mbarbin:
This is the initial release for `Loc`, an OCaml library to manipulate code locations, which are ranges of lexing positions from a parsed file. More info at https://github.com/mbarbin/loc

See discussion in #26604 about publishing a package named `loc` to opam.

See discussion https://github.com/ocaml/dune/discussions/10948 about more sharing with stdune [^1]

[^1]: Regarding the potential follow up with dune's stdlib, I suppose it should be possible to replace an opam package from a certain new major semver in the future. I think it is fine to start with this `loc` and reconsider at a later point. Feel free to comment on the Dune discussion.


shonfeder:
Thanks for publishing a useful utility and for being so considerate about the namespace etc.

All LGTM :)

PR #26626, opened 2024-09-26
[conf-gmp] [depext] 32bit dev package for Debian

ejgallego:
I think this should be fine, and makes the `zarith` opam package build with a 32bit OCaml on Ubuntu / Debian.

In the case of a very old distro, this should be OK as `libgmp-dev:i386` is rewritten to `libgmp-dev`.

### DIFF
--- a/packages/conf-gmp/conf-gmp.4/opam
+++ b/packages/conf-gmp/conf-gmp.4/opam
@@ -28,6 +28,8 @@ depends: [
 depexts: [
   ["libgmp-dev"] {os-family = "debian"}
   ["libgmp-dev"] {os-family = "ubuntu"}
+  ["libgmp-dev:i386"] {os-family = "debian" & arch = "x86_32"}
+  ["libgmp-dev:i386"] {os-family = "ubuntu" & arch = "x86_32"}
   ["gmp"] {os = "macos" & os-distribution = "homebrew"}
   ["gmp"] {os-distribution = "macports" & os = "macos"}
   ["gmp" "gmp-devel"] {os-distribution = "centos"}


jmid:
IIUC, there are two ways to support 32-bit:

- there's running on a good old x86 32-bit processor (or in a Docker container simulating it) or
- there's [multi-arch support](https://wiki.debian.org/Multiarch/Implementation) allowing to compile and run 32-bit executables on a 64-bit "amd64/x86_64" processor

It seems to me, that this PR offers `conf-gmp` packages for the latter, while the CI is failing on the former with
- x86_32-ocaml-4.14 - conf-gmp.4 (failed: [SKIP] Package not available)
- x86_32-ocaml-5.2 - conf-gmp.4 (failed: [SKIP] Package not available)

I can't think of a good way to distinguish the two 32bit modes unfortunately :shrug: 
`ocaml-option-32bit` for the latter uses `gcc -m32` as its C-compiler - but that is definitely hackish! :sweat_smile:

ejgallego:
Hi @jmid , thanks for the analysis. I would believe this solution would have worked on case 1 too as in this case `libgmp-dev:i386` is an alias for `libgmp-dev`, but indeed maybe we need to be more careful.

Do you have a link to the CI failure? TIA

ejgallego:
I see, so indeed it is depexts who doesn't understand the `:i386` prefix, `apt` should.

I wonder if it is worth patching depext to understand this?

smorimoto:
Any updates?

shonfeder:
IIUC, the proposed packaging fix here requires an update to opam to support the suffix in depext. As part of clearing out our backlog of stale packages, I will close this for know, but please feel free to re-open if there is another way to move this forward or if the requirement is met! Thanks :)

kit-ty-kate:
cf. https://github.com/ocaml/opam/issues/6413

ejgallego:
Thanks folks, indeed it seems reasonable to follow up at the opam repos, cheers.

PR #26627, opened 2024-09-26
Package mutaml.0.2

jmid:
### `mutaml.0.2`
A mutation tester for OCaml
Mutaml is a mutation testing tool for OCaml.
It uses a ppxlib-based preprocessor to make a series of small
breaking changes to a program's source code and then runs
the program's testsuite for each of them to catch uncaught
misbehaviour.

---
* Homepage: https://github.com/jmid/mutaml
* Source repo: git+https://github.com/jmid/mutaml.git
* Bug tracker: https://github.com/jmid/mutaml/issues

---


jmid:
CI was failing linting with a missing `dev-repo` warning.
I've thus added a `source`-entry to `dune-project` and regenerated the opam file which should address this.

On the first CI run I also saw
- opensuse failing to install `conf-diffutils` - due to sys-package-installation conflict  (hence unrelated)
- an error while running the tests with 4.14 on ppc64. 

Note:
- The latter was also present in #24104 for the 0.1 release, but I've eliminated two other platform CI failures
- The package installs cleanly, it is just the test run that fails
- I believe this is due to missing overflow protection with 4.x on ppc64 (see ocaml/ocaml#11415) as the tests run cleanly with 5.x on ppc64, which has a different stack layout

avsm:
You can push constraints to the with-test stanza to stop these architectures from being tests (especially the known-based 4.14 ppc64 ones). That'll prevent these tests from popping up in other CI runs in the future.

raphael-proust:
In general, tests that rely on timeout don't play very nicely with opam releases. The CI for opam-repository is often different enough from your own which makes different things fail at different times. In general I'd recommend not to have those tests in the opam package (maybe make it a different target which is not called in the opam file).

Anyway, in the meantime, these errors are acceptable so I'm merging.

PR #25527, opened 2024-03-19
8 packages from savonet/ocaml-ffmpeg at 1.1.11

toots:
This is a bugfix release of `ocaml-ffmpeg` addressing segfaults related to the new OCaml concurrent GC. Should be functionally identical and compatible with the previous release.

---
:camel: Pull-request generated by opam-publish v2.3.0

PR #26631, opened 2024-09-27
2 packages from smimram/ocaml-dblp at 0.1.0

smimram:
This pull-request concerns:
- `dblp.0.1.0`: Commandline program do query the DBLP bibliographic database
- `dblp-api.0.1.0`: Library to query the DBLP bibliographic database

---
* Homepage: https://github.com/smimram/ocalm-dblp
* Source repo: git+https://github.com/smimram/ocalm-dblp.git
* Bug tracker: https://github.com/smimram/ocalm-dblp/issues

---


avsm:
Some test failures here; are they serious @smimram?

smimram:
No, they are not serious at all. They are due to
- a missing dependency on `dblp` for the `runtest` alias (thus the failure for `dblp-api` tests)
- the absence of network available on the test machines

I have a CI which is working well on github.

I can fix the first if you insist on it (it is already done on the git and will be in future versions), but for the second I am afraid I cannot do much more for now...

shonfeder:
In https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/4f459584657e9303ae0d4a88c445a3775830c157/variant/compilers,4.09,dblp.0.1.0

I've noticed

```
[ERROR] The compilation of dblp.0.1.0 failed at "dune build -p dblp -j 39 @install".

#=== ERROR while compiling dblp.0.1.0 =========================================#
# context              2.3.0~alpha~dev | linux/x86_64 | ocaml-base-compiler.4.09.1 | pinned(https://github.com/smimram/ocaml-dblp/archive/refs/tags/v0.1.0.tar.gz)
# path                 ~/.opam/4.09/.opam-switch/build/dblp.0.1.0
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p dblp -j 39 @install
# exit-code            1
# env-file             ~/.opam/log/dblp-7-afff3b.env
# output-file          ~/.opam/log/dblp-7-afff3b.out
### output ###
<snip>
# File "src/dblp.ml", line 69, characters 4-26:
# 69 |     Filename.quote_command "x-www-browser" ["http://doi.org/" ^ doi] |> Sys.command |> ignore
#          ^^^^^^^^^^^^^^^^^^^^^^
# Error: Unbound value Filename.quote_command
```

It appear this is because [Filename.quote_command](https://ocaml.org/manual/5.2/api/Filename.html#VALquote_command) was not added until 4.10.

To fix this, we can either write a replacement for for this function, or increase the lower bound on `ocaml`. WDYT?

shonfeder (changes requested):
Hello! This looks like a very useful package. Thank you for taking the time to publish it!

There are several linting fixes to be made to the package data, which I've noted in my review. (The PR has inspired https://github.com/ocurrent/opam-repo-ci/issues/370, which will be a useful addition :) ).

I would be in favor of getting the tests fixed here before we merge it. While we can make exceptions, every exception we make contributes to making the CI results less accurate, undercutting the value proposition in our promising (and IMO, already quite successful) experiment in ecosystem-wide CI/CD.

There are two main benefits:

- Benefit for users of a package: Users can have good confidence that it will work as expected on tested platforms (of which there are [many](https://github.com/ocurrent/opam-repo-ci/blob/master/doc/platforms.md#platforms-matrix)).
- Benefit for authors of a package: When the package tests are reliable, we are able to give reliable results in our reverse dependency tests. This means, if any of your dependencies introduce changes that break your package, we will ensure that unintended breakage is fixed or that intended breakage is guarded by adding an upper bound to the opam file.

With this in mind, I would propose that we address

> the absence of network available on the test machines

following the example of https://github.com/ocaml/opam-repository/pull/25527#issuecomment-2017155985, where tests that require network access are moved into a distinct dune target https://github.com/savonet/ocaml-ffmpeg/blob/f97819265c35375b79232443665fbc413fda45f2/test/dune#L6. Conversely, I think you could change [the build instructions](https://github.com/smimram/ocaml-dblp/blob/7ec799e64d87fe4fe637b1c6567bad82f620cbfd/dblp-api.opam#L21-L32) to use a different target for `with-test`, but I think that would go against the grain of the dune packaging flow.

If you are open to one of these fixes, but don't have the bandwidth to put it in place, I'd be happy to lend a hand.

smimram:
Thanks @shonfeder for your clear and detailed review. I will address the points you mentioned and update the PR soon.

smimram:
I had to make a new release and therefore I made a new PR (#26658) which hopefully fixes the above problems.

shonfeder:
Closing in favor of https://github.com/ocaml/opam-repository/pull/26658

PR #26632, opened 2024-09-27
Update provider

mbarbin:
- Add maintainer email to existing provider packages

- Flagged older `provider.0.0.4` as deprecated.

This version has a bug in the binding lookup I'd be happy to forget. `opam.ocaml.org` says this package is not used ~~so I propose to simply delete this release. Is that technically acceptable?~~

### DIFF
--- a/packages/provider/provider.0.0.4/opam
+++ b/packages/provider/provider.0.0.4/opam
@@ -1,7 +1,7 @@
 opam-version: "2.0"
 synopsis:
   "Parametrize your OCaml library with values that behave like objects but aren't"
-maintainer: ["Mathieu Barbin"]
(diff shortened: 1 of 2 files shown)


avsm:
We can't remove packages, unfortunately -- there may be other opam repositories that depend on this ourside of ours. However, you can add a `messages` to indicate it is deprecated...

mbarbin:
OK thanks for letting me know. I looked at the manual and discovered a wealth of options I wasn't aware of, such as ... a deprecated flag for releases with known bugs! Thanks, updated :smiley:

avsm:
Looks perfect, thanks for the flexibility!

PR #26537, opened 2024-09-11
odoc is not compatible with OCaml 5.3 (compiler-libs change)

kit-ty-kate:
```
#=== ERROR while compiling odoc.2.4.2 =========================================#
# context              2.3.0~alpha~dev | linux/x86_64 | ocaml-variants.5.3.0+trunk | file:///home/opam/opam-repository
# path                 ~/.opam/5.3/.opam-switch/build/odoc.2.4.2
# command              ~/.opam/5.3/bin/dune build -p odoc -j 1 @install
# exit-code            1           
# env-file             ~/.opam/log/odoc-20-7bf2bf.env 
# output-file          ~/.opam/log/odoc-20-7bf2bf.out 
### output ###                    
# (cd _build/default && /home/opam/.opam/5.3/bin/ocamlopt.opt -w -40 -g -w -18-53 -g -I src/syntax_highlighter/.syntax_highlighter.objs/byte -I src/syntax_highlighter/.syntax_highlighter.objs/native -I /home/opam/.o
pam/5.3/lib/ocaml/compiler-libs -intf-suffix .ml -no-alias-deps -o src/syntax_highlighter/.syntax_highlighter.objs/native/syntax_highlighter.cmx -c -impl src/syntax_highlighter/syntax_highlighter.pp.ml)
# File "src/syntax_highlighter/syntax_highlighter.ml", lines 4-146, characters 2-22:
# Error (warning 8 [partial-match]): this pattern-matching is not exhaustive.
# Here is an example of a case that is not matched:
# (METAOCAML_ESCAPE|METAOCAML_BRACKET_OPEN|METAOCAML_BRACKET_CLOSE|EFFECT)
# (cd _build/default && /home/opam/.opam/5.3/bin/ocamlc.opt -w -40 -g -w -18-53 -g -bin-annot -bin-annot-occurrences -I src/loader/.odoc_loader.objs/byte -I /home/opam/.opam/5.3/lib/astring -I /home/opam/.opam/5.3/l
ib/camlp-streams -I /home/opam/.opam/5.3/lib/fpath -I /home/opam/.opam/5.3/lib/ocaml/compiler-libs -I /home/opam/.opam/5.3/lib/odoc-parser -I /home/opam/.opam/5.3/lib/result -I src/document/.odoc_document.objs/byte 
[...]

PR #26640, opened 2024-09-27
odoc is not compatible with OCaml 5.3 (uses compiler-libs)

kit-ty-kate:
Already done in https://github.com/ocaml/opam-repository/pull/26537 but a new version popped-up since


avsm:
Thanks! /cc @jonludlam

PR #26649, opened 2024-09-30
Change src URL to new home

jonahbeckford:


jonahbeckford:
The lint error:

`Error in DkSDKFFIOCaml_Std.1.0.0~1:            warning 41: Some packages are mentioned in package scripts or features, but there is no dependency or depopt toward them: "DkSDKFFIOCaml_StdExport-linux_x86_64"`

looks like a false positive. Regardless, this is just an attribute edit to an existing package.

avsm:
LGTM, no checksum change. Thanks!

PR #26650, opened 2024-09-30
[new release] file-rewriter (2 packages) (0.0.3)

mbarbin:
This is the initial release for 2 packages for applying small rewrites to tweak or refactor your files. Short synopsis below, more info at https://github.com/mbarbin/file-rewriter.

### file-rewriter

`File_rewriter` is an OCaml library for applying small rewrites to tweak or refactor your files. It provides a convenient interface to apply surgical textual substitutions on the fly, while navigating the contents of a file through an abstract representation containing code locations.

### sexps-rewriter

`Sexps_rewriter` is a specialized version of the main library dedicated to rewriting sexp files.

## Motivation

We created these libraries as part of an ongoing work to create linting and refactoring tools for the many `dune` files found in big monorepos (see [dunolint](https://github.com/mbarbin/dunolint)).

### DIFF
--- a/packages/file-rewriter/file-rewriter.0.0.3/opam
+++ b/packages/file-rewriter/file-rewriter.0.0.3/opam
@@ -1,0 +1,40 @@
+opam-version: "2.0"
+synopsis: "Apply small rewrites to tweak or refactor your files"
+maintainer: ["Mathieu Barbin <opensource@mbarbin.org>"]
+authors: ["Mathieu Barbin"]
+license: "Apache-2.0"
+homepage: "https://github.com/mbarbin/file-rewriter"
+doc: "https://mbarbin.github.io/file-rewriter/"
+bug-reports: "https://github.com/mbarbin/file-rewriter/issues"
+depends: [
+  "dune" {>= "3.16"}
+  "ocaml" {>= "5.2"}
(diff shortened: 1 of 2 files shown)


avsm:

mbarbin:
Thanks @avsm . I saw the package landed on `ocaml.org`. It doesn't have odoc pages on its [home page](https://ocaml.org/p/file-rewriter/0.0.3) (says "! No Docs"). I'd like to fix this. How can I enable it? Thanks!

shonfeder:
Sorry for the delay, @mbarbin . It looks to me like we had a solver failure when trying to generate the docs: https://docs.ci.ocaml.org/package/file-rewriter.0.0.3

Given that it passed CI here, this is obviously a problem with our infrastructure. I'll let you know once I figure something out!

mbarbin:
Thank you for letting me know @shonfeder. I looked at the CI log you linked, indeed I see:

```text
- ocaml -> (problem)
    User requested <= 5.1.1
    ...
    file-rewriter 0.0.3 requires >= 5.2
```

I don't know where the `User requested <= 5.1.1` comes from.

`>= 5.2` does indeed come from the opam file. In theory, this bound isn't needed for this package, but I wanted to avoid committing to supporting older versions.

PR #26653, opened 2024-10-02
Camlp5 release to support upcoming Ocaml 5.3.0

chetmurthy:
what it says on the tin.


shonfeder:
Windows failure is 

```
  #=== ERROR while compiling conf-m4.1 ==========================================#
``

The only ocaml-ci failures are revdeps related and expected, as per https://github.com/ocaml/opam-repository/pull/25813#issuecomment-2113530835

Thanks for the pro-active update!

PR #26659, opened 2024-10-03
Release of MetAcsl.0.7

tmarti2:
Release of MetAcsl compatible with Frama-C 29.0 (Copper)


tmarti2:
Windows Ci fail is normal because it is a plug-in of Frama-C, which does not pass windows CI either.

mseri:
Thanks

PR #26668, opened 2024-10-04
2 packages from ocsigen/lwt at 5.8.0

raphael-proust:
This pull-request concerns:
- `lwt.5.8.0`: Promises and event-driven I/O
- `lwt_ppx.5.8.0`: PPX syntax for Lwt, providing something similar to async/await from JavaScript

---
* Homepage: https://github.com/ocsigen/lwt
* Source repo: git+https://github.com/ocsigen/lwt.git
* Bug tracker: https://github.com/ocsigen/lwt/issues

---


raphael-proust:
The only CI errors I see are:

- some misc revdeps (curl timeout, hash mismatch, generic build fail, etc.)
- some jailing issue in macos (networking syscalls failing during the tests)
- also jailing issues(?) or some other system issue(?) on freebsd

As far as I'm concerned, this is ready to merge.

hannesm:
I'm only worried a bit about the following error:

```
#=== ERROR while compiling lambda-term.3.3.2 ==================================#
# context              2.4.0~alpha1~dev | linux/x86_64 | ocaml-base-compiler.4.14.2 | file:///home/opam/opam-repository
# path                 ~/.opam/4.14/.opam-switch/build/lambda-term.3.3.2
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p lambda-term -j 255 @install
# exit-code            1
# env-file             ~/.opam/log/lambda-term-7-dfcc94.env
# output-file          ~/.opam/log/lambda-term-7-dfcc94.out
### output ###
# File "src/dune", line 11, characters 11-27:
# 11 |     (names lTerm_term_stubs lTerm_unix_stubs lTerm_windows_stubs)
#                 ^^^^^^^^^^^^^^^^
# (cd _build/default/src && /usr/bin/gcc -O2 -fno-strict-aliasing -fwrapv -pthread -fPIC -D_FILE_OFFSET_BITS=64 -g -I /home/opam/.opam/4.14/lib/ocaml -I /home/opam/.opam/4.14/lib/bytes -I /home/opam/.opam/4.14/lib/logs -I /home/opam/.opam/4.14/lib/lwt -I /home/opam/.opam/4.14/lib/lwt/unix -I /home/opam/.opam/4.14/lib/lwt_react -I /home/opam/.opam/4.14/lib/mew -I /home/opam/.opam/4.14/lib/mew_vi -I /home/opam/.opam/4.14/lib/ocaml/threads -I /home/opam/.opam/4.14/lib/ocplib-endian -I /home/opam/.opam/4.14/lib/ocplib-endian/bigstring -I /home/opam/.opam/4.14/lib/react -I /home/opam/.opam/4.14/lib/result -I /home/opam/.opam/4.14/lib/trie -I /home/opam/.opam/4.14/lib/uchar -I /home/opam/.opam/4.14/lib/uucp -I /home/opam/.opam/4.14/lib/uuseg -I /home/opam/.opam/4.14/lib/uutf -I /home/opam/.opam/4.14/lib/zed -o lTerm_term_stubs.o -c lTerm_term_stubs.c)
# In file included from lTerm_term_stubs.c:10:
# /home/opam/.opam/4.14/lib/lwt/unix/lwt_unix.h:19:28: error: expected declaration specifiers or '...' before '(' token
#    19 |     caml_convert_flag_list((flags), (int *)(table))
#       |                            ^
# /home/opam/.opam/4.14/lib/ocaml/caml/alloc.h:60:16: note: in expansion of macro 'caml_convert_flag_list'
#    60 | CAMLextern int caml_convert_flag_list (value, int *);
#       |                ^~~~~~~~~~~~~~~~~~~~~~
# /home/opam/.opam/4.14/lib/lwt/unix/lwt_unix.h:19:37: error: expected declaration specifiers or '...' before '(' token
#    19 |     caml_convert_flag_list((flags), (int *)(table))
#       |                                     ^
# /home/opam/.opam/4.14/lib/ocaml/caml/alloc.h:60:16: note: in expansion of macro 'caml_convert_flag_list'
#    60 | CAMLextern int caml_convert_flag_list (value, int *);
#       |                ^~~~~~~~~~~~~~~~~~~~~~
```

I'm not sure what is the root cause, but it looks like lwt_unix.h usage of `caml_convert_flag_list` doesn't play nicely (anymore?)? is this a regression?

For FreeBSD failures, these seem to be test failures. Not sure whether they're meant to succeed (or did succeed in earlier releases).

raphael-proust:
@hannesm thanks for pointing this error out.

I notice that lambda-term `#include <lwt_unix.h>` so it gets the new macro that was introduced in [`#1022](https://github.com/ocsigen/lwt/pull/1022)'s `0da9177c79`.

I don't know to what extent we should make an effort to keep backwards compatibility in `lwt_unix.h`. It's not really meant for "regular" use. So I think we should just mark some incompatibility and ping the devs. We should also `opam grep` for other libraries which include `lwt_unix.h` just in case.

Ping @MisterDA @smorimoto

raphael-proust:
```
$ opam grep --depends-on=lwt ".*include.*lwt_unix\.h.*"
[Info] Getting the list of all known opam packages..
[Info] Fetching and grepping using rg..
baardskeerder matches your regexp.
hidapi-lwt matches your regexp.
kafka matches your regexp.
kafka_lwt matches your regexp.
lambda-term matches your regexp.
lwt matches your regexp.
lwt_camlp4 matches your regexp.
lwt_ppx matches your regexp.
lwt_ppx_let matches your regexp.
lwt_react matches your regexp.
mirage-block-unix matches your regexp.
named-pipe matches your regexp.
plebeia matches your regexp.
reason matches your regexp.
rtop matches your regexp.
usb matches your regexp.
```

MisterDA:
The problem is that `lwt_unix.h` defines this for compatibility:
```c
#if OCAML_VERSION < 50000
#define caml_convert_flag_list(flags, table) \
    caml_convert_flag_list((flags), (int *)(table))
#endif
```

the real function is defined in `caml/alloc.h`, but `lwt_unix.h` _doesn't_ include this header. So, when packages do
```c
#include <lwt_unix.h>
/* ... */
#include <caml/alloc.h>
```
the prototype in `<caml/alloc.h>` gets rewritten with this macro to
```c
extern int caml_convert_flag_list((value), (int *)(int *));
```
triggering the compilation failure.

I suggest we either:
1. define an internal header for Lwt with this macro and possibly others in the future, for compatibility; this is totally invisible for packages including `lwt_unix.h`.
2. include `caml/alloc.h` in `lwt_unix.h` *before* defining this macro; as `caml/alloc.h` has header guards, the prototype won't be substituted after the macro is defined. This won't work if, for some reason (such as compatibility), a package includes `lwt_unix.h` and later on provides it's own definition of `caml_convert_flag_list`.

MisterDA:
There's a third option, which is simply to change the namespace and provide `lwt_convert_flag_list` wrapping `caml_convert_flag_list`.

raphael-proust:
thanks @MisterDA 

I think it makes more sense to include this in 5.8.0 rather than pusblish a 5.8.1. It keeps the constraints/conflicts from other packages to the simplest they can be. I'll ammend 5.8.0

raphael-proust:
The CI now is green for `lambda-term`. Is there a way to search for strings in the whole of the logs? It'd be nice to search for `caml_convert_flag_list`…

smorimoto:
I think we can move forward now.

mseri:
> Is there a way to search for strings in the whole of the logs? It'd be nice to search for `caml_convert_flag_list`…

I don't think we have a way to search in the whole logs, but we can in principle search in the whole opam: https://github.com/kit-ty-kate/opam-grep

mseri:
The logs look fine, modulo the massive amount of `curl failed`

PR #26671, opened 2024-10-04
Edit availability constraints for `checked_oint` v0.2.0

hirrolot:
See [this discussion](https://github.com/ocaml/opam-repository/pull/26662#discussion_r1787311683).

### DIFF
--- a/packages/checked_oint/checked_oint.0.2.0/opam
+++ b/packages/checked_oint/checked_oint.0.2.0/opam
@@ -17,17 +17,7 @@ depends: [
   "alcotest" {>= "1.7.0"}
   "odoc" {with-doc}
 ]
-available: arch != "arm32" & arch != "x86_32" & arch != "ppc64" & os != "win32"
-x-ci-accept-failures: [
-  "fedora-38"
-  "fedora-39"
-  "fedora-40"
-  "opensuse-15.5"
-  "opensuse-tumbleweed"
-  "oraclelinux-8"
-  "oraclelinux-9"
-  "freebsd"
-]
+available: arch != "arm32" & arch != "x86_32" & arch != "ppc64" & os != "win32" & os != "freebsd" & os-distribution != "fedora" & os-distribution != "ol" & os-family != "opensuse" & os-family != "suse"
 build: [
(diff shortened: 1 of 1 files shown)


hirrolot:
I tried `os-distribution != "opensuse"`, `os-family != "opensuse"`, `os != "opensuse"`, but nothing works to disable building on openSUSE. Any ideas?

hirrolot:
It's interesting that `os-family != "opensuse"` disables `opensuse-tumbleweed` but not `opensuse-15.6`...

packages/checked_oint/checked_oint.0.2.0/opam
```diff
@@ -17,17 +17,7 @@ depends: [
...
-  "fedora-38"
-  "fedora-39"
-  "fedora-40"
-  "opensuse-15.5"
-  "opensuse-tumbleweed"
-  "oraclelinux-8"
-  "oraclelinux-9"
-  "freebsd"
-]
+available: arch != "arm32" & arch != "x86_32" & arch != "ppc64" & os != "win32" & os != "freebsd" & os-distribution != "fedora" & os-distribution != "oraclelinux" & os-family != "opensuse" & switch != "5.3~alpha1"
```
mseri, line 20: You need this
```suggestion
available: arch != "arm32" & arch != "x86_32" & arch != "ppc64" & os != "win32" & os != "freebsd" & os-distribution != "fedora" & os-distribution != "oraclelinux" & os-family != "opensuse" & os-family != "suse"
```

if the library is not compatible with ocaml 5.3 I would suggest an upper bound on the ocaml dependency instead. This is the alpha release of the compiler though, so I thunk it is better not to add any extra compiler constraint for the moment

mseri: I forgot to mention that the various os flags are documented here https://github.com/ocaml/opam-repository/wiki/Depexts-os-distribution---os-family-values

hirrolot:
Thanks for helping me @mseri, the CI works right now except `5.3~alpha1`.

mseri:
Thanks a lot and sorry for the slow response!

PR #26675, opened 2024-10-05
edit ascend v0.1.2 package

m-laniakea:
Help menu was missing a single command.


mseri:
Please don't modify the hashes, package contents on the repository are considered immutable. This needs to be released separately. If you no longer have the tarballs you can mark this as unavailable and make a new release, otherwise we can try to recover it from the opam cache (see https://github.com/ocaml/opam-repository/wiki/FAQ#how-to-find-lost-archives-of-packages)

m-laniakea:
Sorry about that!
The link in master points to the correct tarball.

Versions 0.1.0 and 0.1.2 are strictly inferior to their respective next versions.
Do you recommend leaving these as-is?

Moving this to: https://github.com/ocaml/opam-repository/pull/26692

mseri:
Yes, the old versions can stay. Opam will always pick the latest. Soon we will implement a system that will periodically move unmaintained older version to an archive repository, then they will disappear from there (but still remain available for the people that want to reach them)

mseri:
Can I close this PR?

PR #26691, opened 2024-10-08
[new release] Five Ortac packages 0.4.0

n-osborne:
This release brings a number of new features and improvements:

- New features:
  + [#247](https://github.com/ocaml-gospel/ortac/pull/247): Generated tests cover now functions with multiple SUT arguments
  + [#253](https://github.com/ocaml-gospel/ortac/pull/253): Generated tests cover now functions returning SUT values
  + [#259](https://github.com/ocaml-gospel/ortac/pull/259): Generated tests can be run in a separated process with a timeout
- Improvements:
  + [#245](https://github.com/ocaml-gospel/ortac/pull/245): Fix the analysis of function signature to explicitly not support SUTs inside another type
  + [#251](https://github.com/ocaml-gospel/ortac/pull/251): Fix the display of the runnable scenario for value returned by a function that could have raised an exception

PR #26697, opened 2024-10-09
[package update] Strenghten dependencies constraints in Ortac packages

n-osborne:
This PR updates some dependencies constraints for some Ortac packages.

When an Ortac package depends on another Ortac package, make the constraints ask for the exact same version.
The reason why is that version 0.4.0 brings some breaking changes (see #26691) and we can't guarantee that it will be an isolate case.

Constraints on the Gospel package are also added, as the next release is expected to bring breaking changes.

### DIFF
--- a/packages/ortac-core/ortac-core.0.1.0/opam
+++ b/packages/ortac-core/ortac-core.0.1.0/opam
@@ -30,7 +30,7 @@ depends: [
   "ppxlib" {>= "0.26.0"}
   "gospel" {>= "0.2.0"}
   "alcotest" {with-test & >= "0.8.1"}
-  "ortac-runtime" {with-test}
+  "ortac-runtime" {with-test & = version}
 ]
 build: [
   ["dune" "subst"] {dev}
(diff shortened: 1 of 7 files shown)


jmid (commented):
Overall, with the Ortac packages we are trying to make things work together within versions, while avoiding issues arising from cross-version combinations, like ortac-core.0.4 and ortac-qcheck-stm.0.4.

The changes LGTM for doing so!

I think there is also an occurrence in packages/ortac-core/ortac-core.0.3.0/opam that needs `= version` but then everything should be version-aligned AFAICS!
```
  "ortac-runtime" {with-test}
```

n-osborne:
Good catch, I missed this one!

jmid:
The `ppx_deriving.5.2.1` failure on `ocaml.5.3.0~alpha1` is unrelated.

I've just filed #26702 to add missing upper bounds to address it.

jmid:
This is all green modulo the `ppx_deriving` issue fixed in #26702.
It should therefore be good to merge, so that #26691 can be rebased and rerun its CI tests :slightly_smiling_face:

mseri:
Thanks

PR #26720, opened 2024-10-10
dummy packages should not have been uploaded to begin with

raphael-proust:
I know that we don't normally remove packages. I can switch to marking them as unavailable if that's preferred.

These packages are internal test artifacts from lwt. I (most likely it's me) made a mistake including them in a release previously.


mseri:
I did not know they were dummy packages. Why not marking them `available:false`? We may still break stuff deleting files, and we will do a big delete within a few weeks anyway

PR #26725, opened 2024-10-11
Package openbsd.1.0

semarie:
Bindings for some OpenBSD syscalls
Provides bindings around [pledge(2)](https://man.opensbd.org/pledge.2) and [unveil(2)](https://man.opensbd.org/unveil.2) syscalls.

---
- Homepage: https://codeberg.org/semarie/ocaml-openbsd/
- Source repo: git+https://codeberg.org/semarie/ocaml-openbsd.git


packages/openbsd/openbsd.1.0/opam
```diff
@@ -0,0 +1,37 @@
...
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+]
+dev-repo: "git+https://codeberg.org/semarie/ocaml-openbsd.git"
```
mseri, line 30: If I understand correctly, you don't need
```
available: os-distribution = "openbsd"
```
because if you don't have pledge and unveil it basically does nothing (and you have the function that check if the functionality is available which is working everywhere).

semarie: yes. it was designed to be installable on all platforms in order to facilitate portability.

Calling `Pledge.promises "stdio"` on Linux will generate a `Failure` exception. A constant `Pledge.supported` is provided to check at runtime if it is supported or not (and call a different subsystem if it isn't the case for example).

The following code will get `pledge(2)` called only on OpenBSD, and others os-distribution will gracefully ignore it.

```ocaml
let open Openbsd in
if Pledge.supported then
  Pledge.promises "stdio rpath"
```

mseri:
Thanks

PR #26726, opened 2024-10-11
Add Z3 4.13.2

wintersteiger:


wintersteiger:
Waiting for https://github.com/ocaml/opam-source-archives/pull/33

packages/z3/z3.4.13.2/opam
```diff
@@ -0,0 +1,52 @@
...
+authors: "MSR"
+homepage: "https://github.com/Z3prover/z3"
+bug-reports: "https://github.com/Z3prover/z3/issues"
+license: "MIT"
+dev-repo: "git+https://github.com/Z3prover/z3.git"
+depexts: [
+  ["python3-distutils"] {os-family = "debian"}
+]
+patches: [
+  "gccstd-2a.patch" { (os = "opensuse") | (os-distribution = "ubuntu" & os-version < "20.04") }
```
mseri, line 12:
```suggestion
  "gccstd-2a.patch" { (os-family = "opensuse") | (os-distribution = "ubuntu" & os-version < "20.04") }
```

mseri: From https://github.com/ocaml/opam-repository/wiki/Depexts-os-distribution---os-family-values

wintersteiger: Awesome, thanks for the pointer!

wintersteiger:
I think this one's ready now. Apparently opensuse has a very old C/C++ compiler (gcc 7) installed by default and Z3 requires C++20, so I had to disable it for those systems.

mseri:
Is the windows failure expected?
```
   #=== ERROR while compiling z3.4.13.3 ==========================================#
  # context     2.2.1 | win32/x86_64 | ocaml.5.2.0 | file://D:/a/opam-repository/opam-repository
  # path        D:\opamroot\default\.opam-switch\build\z3.4.13.3
  # command     D:\opamroot\.cygwin\root\bin\make.exe -C build -j 3
  # exit-code   2
  # env-file    D:\opamroot\log\z3-8008-cfe39a.env
  # output-file D:\opamroot\log\z3-8008-cfe39a.out
  ### output ###
  # make: Entering directory '/cygdrive/d/opamroot/default/.opam-switch/build/z3.4.13.3/build'
  # make: cl: No such file or directory
  # make: *** [Makefile:2684: smt\smt_statistics.obj] Error 127
  # make: *** Waiting for unfinished jobs....
  # make: Leaving directory '/cygdrive/d/opamroot/default/.opam-switch/build/z3.4.13.3/build'
```

wintersteiger:
It is to me, but only because we ignored it the last time too! 

The problem is that the Visual Studio C/C++ compiler (cl.exe) is not installed on the CI machines, or at least not in the path. What is the preferred compiler for Windows, and is it in the corresponding env vars?

mseri:
@dra27 do you know the answer to the question above?

mseri:
> The problem is that the Visual Studio C/C++ compiler (cl.exe) is not installed on the CI machines, or at least not in the path. What is the preferred compiler for Windows, and is it in the corresponding env vars?

Yes, I wasn't planning on blocking this on windows further. I meant more about the question of the preferred windows compiler (if any)

dra27:
Ah, oops - stopped at the first question 🤦 The preference would be to work for both - but the opam needs updating to change the command line to the appropriate one based on whether `host-system-mingw` or `host-system-msvc` is installed (from what I can see, the invocation of the Python script needs to be very different when building for mingw-w64)

PR #26728, opened 2024-10-11
[new release] rust-staticlib-gen (3 packages) (0.1.0)

Lupus:
Rust staticlib generation utility

- Project page: <a href="https://github.com/Lupus/rust-staticlib-gen">https://github.com/Lupus/rust-staticlib-gen</a>

##### CHANGES:

* Initial release

### DIFF
--- a/packages/dune-cargo-build/dune-cargo-build.0.1.0/opam
+++ b/packages/dune-cargo-build/dune-cargo-build.0.1.0/opam
@@ -1,0 +1,40 @@
+opam-version: "2.0"
+synopsis:
+  "Tool to invoke Cargo for building Rust crates within the dune sandbox"
+description:
+  "dune-cargo-build is a tool that runs cargo build in offline mode for a specified crate, ensuring compatibility with dune's/opam's sandboxing. It parses Cargo's JSON output to determine the produced artifacts and copies them to the current directory, renaming them to match what OCaml expects for foreign stubs. This tool is useful for integrating Rust build processes into OCaml projects managed by dune."
+maintainer: ["Konstantin Olkhovskiy <lupus@oxnull.net>"]
+authors: ["Konstantin Olkhovskiy <lupus@oxnull.net>"]
+license: "Apache-2.0"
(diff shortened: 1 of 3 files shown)


mseri:
Thanks

packages/rust-staticlib/rust-staticlib.0.1.0/opam
```diff
@@ -0,0 +1,35 @@
+opam-version: "2.0"
+synopsis: "A dummy package to define staticlib virtual library"
```
mseri, line 2: Is this expected to be released?

It may be worth to add a bit of description to the three opam files to clarify better the differences between the three (or an example of use). You have a very nicely detailed readme where to extract the bits

Lupus: Yeah, it's expected to be released, it's some sort of a sentinel or a marker (virtual dune) library that tells the end user that there are Rust dependencies somewhere down the transitive dependency closure of their project, and in order to successfully link they need to have Rust staticlib generated in their project (generated staticlib will implement that virtual dune library, and linking it in final executable should resolve dune complaints *and* provide the necessary `external` function definitions for the linker). Not an ideal way, but still much better than just staring at the linker errors complaining about missing Rust stubs.
I'll come up with something more descriptive to reflect the relationships between those 3 packages, and explain why "dummy" lib needs to be on opam repository. As my plan somewhat suggests that once users find missing virtual library implementation error from dune - they go googling what's that, they should find this opam package and it should really give them some solid explanation with further pointers on how to unblock themselves.
How much text is considered a good size for synopsis? dune does not allow multi-line strings in dune files :(

Lupus: I've added description fields with more verbose explanation on what are the packages, could you please take a look? Thanks!

mseri:
Thanks! Feel free to announce the release on discuss.ocaml.org. We have a community category and an announcement tag for this purpose

PR #26734, opened 2024-10-13
ppx_expect 0.17.1 is broken on windows

hhugo:
similar to #26222 

FIx proposed upstream in https://github.com/janestreet/ppx_expect/pull/57


mseri:
Thanks

PR #26737, opened 2024-10-14
nqsb.io was sunset -- update links in all packages that refer there.

hannesm:


avsm:
Many thanks, merging quickly due to the need to rebuild the docs and such to avoid linking to the old site.

PR #26738, opened 2024-10-14
Package dates_calc.0.0.7

rmonat:
### `dates_calc.0.0.7`
A date calculation library
A date calculation library, with exact operators to add a given
 number of days to a date, and approximate operators to add months or
 years.

---
* Homepage: https://github.com/CatalaLang/dates-calc
* Source repo: git+https://github.com/CatalaLang/dates-calc.git
* Bug tracker: https://github.com/CatalaLang/dates-calc/issues

---


packages/dates_calc/dates_calc.0.0.7/opam
```diff
@@ -0,0 +1,42 @@
...
+authors: ["Aymeric Fromherz" "Denis Merigoux" "Raphaël Monat" "Louis Gesbert"]
+license: "Apache-2.0"
+homepage: "https://github.com/CatalaLang/dates-calc"
+bug-reports: "https://github.com/CatalaLang/dates-calc/issues"
+depends: [
+  "dune" {>= "3.11"}
+  "ocaml" {>= "4.11.0"}
+  "alcotest" {with-test & >= "1.5.0"}
+  "qcheck" {with-test & >= "0.15"}
+  "odoc" {with-doc}
```
mseri, line 18:
```suggestion
  "conf-python3" {with-test}
  "odoc" {with-doc}
```

packages/dates_calc/dates_calc.0.0.7/opam
```diff
@@ -0,0 +1,43 @@
...
+authors: ["Aymeric Fromherz" "Denis Merigoux" "Raphaël Monat" "Louis Gesbert"]
+license: "Apache-2.0"
+homepage: "https://github.com/CatalaLang/dates-calc"
+bug-reports: "https://github.com/CatalaLang/dates-calc/issues"
+depends: [
+  "dune" {>= "3.11"}
+  "ocaml" {>= "4.11.0"}
+  "alcotest" {with-test & >= "1.5.0"}
+  "qcheck" {with-test & >= "0.15"}
+  "conf-python3" {with-test}
```
mseri, line 18:
```suggestion
  "conf-python-3" {with-test}
```

raphael-proust:
Are the test failures due to the mismatch of python version?

It works on some distros (e.g., arch with python 3.12) but not on others (e.g., debian with python 3.9).

```
#   File "/home/opam/.opam/5.2/.opam-switch/build/dates_calc.0.0.7/_build/default/lib_python/src/dates_calc/dates.py", line 43
#     match month:
#           ^
# SyntaxError: invalid syntax
```

Don't know if you want to do anything about it. I think it's ok to merge and expect users to manage their python environment (although I'd recommend documenting the requirement somewhere).

rmonat:
> Are the test failures due to the mismatch of python version?
> 
> It works on some distros (e.g., arch with python 3.12) but not on others (e.g., debian with python 3.9).
> 
> ```
> #   File "/home/opam/.opam/5.2/.opam-switch/build/dates_calc.0.0.7/_build/default/lib_python/src/dates_calc/dates.py", line 43
> #     match month:
> #           ^
> # SyntaxError: invalid syntax
> ```
> 
> Don't know if you want to do anything about it. I think it's ok to merge and expect users to manage their python environment (although I'd recommend documenting the requirement somewhere).

Yes, this is pattern matching introduced by a recent Python version > 3.9. I would be in favor of merging as-is please!

yawaramin:
@rmonat is there any reason to require conf-python-3 for this package? The functionality of the OCaml package does not depend on Python, if I understand correctly? If an OCaml/opam user wants to use the OCaml `dates_calc` library, they don't _really_ need Python, right?

If so, I would suggest removing the Python dependency for submission to opam-repository and if the user needs to use the Python version of the package, they can install it using their Python tooling which they would need to manage anyway to ensure they have the required minimum version.

rmonat:
@yawaramin thank you for your suggestion! The tests dune runs are for all languages: C, Python -- so having `conf-python-3` would be nice, although we would indeed require a higher version.

yawaramin:
Understood, I just want to avoid dependency bloat down the line. If some packages and projects in the future use this, their entire dependency cone would be pulling in Python despite it not being needed at all. I understand that you are using dune to run the Python library tests, but it seems like we should be able to find another way instead of forcing potential users of this package to install Python.

mseri:
An option could be to disable the tests here, or to make a separate package for the python tests (in the same repo) that runs only on the package CI and not here.

rmonat:
@mseri I totally agree! I removed the dependency for this release, and will do this in a future release. 

Thank you for spotting this too!

packages/dates_calc/dates_calc.0.0.7/opam
```diff
@@ -0,0 +1,42 @@
...
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
```
mseri, line 30:
```suggestion
#    "@runtest" {with-test} # require python 3.12 to run
```

mseri: To disable them in your repo, what you could do is

- add a dummy `dates_calc_python_test.opam`
_ add `(package dates_calc_python_test)` to [`test/dune](https://github.com/CatalaLang/dates-calc/blob/9125d78833e092b5be009f23b7013a753018466a/test/dune#L7)

Then on release you only release `dates_calc`

rmonat:
@mseri @shonfeder ready to merge on my side.

yawaramin (approved):

shonfeder:
Thanks for publishing the update and making sure the it its in good shape for the CI! :pray:

PR #26744, opened 2024-10-14
Package hol_light_module.1.0

aqjune:
### `hol_light_module.1.0`
A flag for compiling HOL Light core to a bytecode and native module
Installing this package makes the hol-light package to build the
bytecode and native module of HOL Light core as well.

NOTE: This extends the trusted base of HOL Light to include its inliner script,
inline_loads.ml. inline_loads.ml is an OCaml program in HOL Light that receives
an HOL Light proof and replaces the loads/loadt/needs function invocations with
their actual contents. Please install this package only if having this
additional trusted base is considered okay.

---
* Homepage: https://hol-light.github.io
* Source repo: git+https://github.com/jrh13/hol-light.git
* Bug tracker: https://github.com/jrh13/hol-light/issues

---


aqjune:
The main hol_light package (which should use this) will be created soon.

aqjune:
I am ready to get reviewed. About the linter failure - its license does not really fit in one of known categories. What would be a good solution in this case?

raphael-proust:
The SPDX standard says you can use `AND` when your code has different licenses combined. So if your repository has code with distinct licenses, you should have `"MIT AND Apache-2.0 AND …"` or something like that.

Otherwise, maybe you can have a non-standard value in the field.

Is it possible to install LICENSE files with opam? It's common for archlinux packages to install a LICENSE file in `share/licenses/$packagename/` and maybe you can do the same?

aqjune:
> Otherwise, maybe you can have a non-standard value in the field.

Yeah, I am in this case I think. This package does not have multiple licenses. This just has a non-standard license: https://github.com/jrh13/hol-light/blob/master/LICENSE

> Is it possible to install LICENSE files with opam? It's common for archlinux packages to install a LICENSE file in share/licenses/$packagename/ and maybe you can do the same?

Could you provide an example package in opam-repository that is doing this? I could not quickly find one.

mseri:
The license is unusual but we have other 3 packages with nonstandard ones and a similar trick to point at github. Until we figure it out, I think it can go as is

shonfeder:
The only CI failure looks to be an issue on the builder, nothing to do with this package. Sounds like this is good to go from the discussion! Thanks for releasing the package!

PR #26745, opened 2024-10-15
Package hol_light.3.0.0

aqjune:
### `hol_light.3.0.0`
The HOL-Light interactive theorem prover
HOL Light is a computer program written by John Harrison to help users prove
interesting mathematical theorems completely formally in higher order logic.
It sets a very exacting standard of correctness, but provides a number of
automated tools and pre-proved mathematical theorems (e.g. about arithmetic,
basic set theory and real analysis) to save the user work. It is also fully
programmable, so users can extend it with new theorems and inference rules
without compromising its soundness. There are a number of versions of HOL,
going back to Mike Gordon’s work in the early 80s. Compared with other HOL
systems, HOL Light uses a much simpler logical core and has little legacy code,
giving the system a simple and uncluttered feel. Despite its simplicity, it
offers theorem proving power comparable to, and in some areas greater than,
other versions of HOL, and has been used for some significant
industrial-scale verification applications.

This package will install `hol.sh` which will run OCaml REPL with HOL Light
loaded.
To use a compiled module of HOL Light, please install with the
hol_light_module OPAM package.

---
* Homepage: https://hol-light.github.io/
* Source repo: git+https://github.com/jrh13/hol-light.git
* Bug tracker: https://github.com/jrh13/hol-light/issues

---

### DIFF
--- a/packages/hol_light/hol_light.3.0.0/opam
+++ b/packages/hol_light/hol_light.3.0.0/opam
@@ -1,0 +1,89 @@
+opam-version: "2.0"
+synopsis: "The HOL-Light interactive theorem prover"
+description: """
+HOL Light is a computer program written by John Harrison to help users prove
+interesting mathematical theorems completely formally in higher order logic.
+It sets a very exacting standard of correctness, but provides a number of
+automated tools and pre-proved mathematical theorems (e.g. about arithmetic,
+basic set theory and real analysis) to save the user work. It is also fully
+programmable, so users can extend it with new theorems and inference rules
+without compromising its soundness. There are a number of versions of HOL,
+going back to Mike Gordon’s work in the early 80s. Compared with other HOL
+systems, HOL Light uses a much simpler logical core and has little legacy code,
+giving the system a simple and uncluttered feel. Despite its simplicity, it
+offers theorem proving power comparable to, and in some areas greater than,
+other versions of HOL, and has been used for some significant
+industrial-scale verification applications.
+
+This package will install `hol.sh` which will run OCaml REPL with HOL Light
+loaded.
+To use a compiled module of HOL Light, please install with the
+hol_light_module OPAM package.
+"""
+authors: "HOL Light"
+license: "https://github.com/jrh13/hol-light/blob/master/LICENSE"
+homepage: "https://hol-light.github.io/"
+bug-reports: "https://github.com/jrh13/hol-light/issues"
+dev-repo: "git+https://github.com/jrh13/hol-light.git"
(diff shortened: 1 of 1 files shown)


aqjune:
This PR is ready for review.
It seems the Windows failure is due to missing m4:
```
  [ERROR] The compilation of conf-m4.1 failed at "sh -exc echo | m4".
```
... and opam-2.0-ocaml-4.14 is due to missing Perl plugins:
```
| - build conf-perl-ipc-system-simple 3
| - build conf-perl-string-shellquote 3
```
Two FreeBSDs are failing when copying `hol.sh` (which is the output artifact of this package) to the bin dir.

raphael-proust:
> missing m4

You should probably add a dep to `conf-m4`. Although it won't fix the issue for windows because `conf-m4` doesn't have a windows entry… but it'll fix your package when it's eventually added.

aqjune:
> > missing m4
> 
> You should probably add a dep to `conf-m4`. Although it won't fix the issue for windows because `conf-m4` doesn't have a windows entry… but it'll fix your package when it's eventually added.

This package does not compile anything; it is supposed to be used as a package flag, as coq-native does (https://ocaml.org/p/coq-native/1). The existence of this hol_light_module package will be used by the hol_light package to determine its additional build flag. I wonder still adding conf-m4 as its dependency is necessary though.

mseri:
If you explicitly need m4, please add the dependency. Otherwise, if not-ocamlfind stops needing it, this package will unnecessarily break.

The license is unusual but we have other 3 packages with nonstandard ones and a similar trick to point at github. Until we figure it out, I think it can go as is

aqjune:
Hi @mseri, this package (hol_light) does not explicitly use m4. m4 is being used by not-ocamlfind which is used by camlp5-buildscripts which is by camlp5. So it is fine with hol_light package to not explicitly have it in dependencies.

Thanks for supporting about the current license form! :)

aqjune:
I removed Windows and FreeBSDs from available OSes because HOL Light does not generate hol.sh on these operating systems.

aqjune:
The one remaining failure is due to missing apt packages in the Debian machine of CI
```
The packages you requested declare the following system dependencies. Please make sure they are installed before retrying:
    libipc-system-simple-perl libstring-shellquote-perl
```
https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/86fa2296e6c8132b1e29abac6a6f2e5d2cc075ef/variant/extras,opam-2.0-ocaml-4.14,hol_light.3.0.0
I think this patch is ready to land. :)

shonfeder:
Thanks for publishing this update, and for taking the time to work through and explain the CI results so carefully! Please consider announcing the new release on discuss.ocaml.org.

PR #26751, opened 2024-10-16
[new release] datalog (0.7)

c-cube:
An in-memory datalog implementation for OCaml

- Project page: <a href="https://github.com/c-cube/datalog">https://github.com/c-cube/datalog</a>
- Documentation: <a href="https://c-cube.github.io/datalog">https://c-cube.github.io/datalog</a>

##### CHANGES:

- compat with OCaml 5.2
- fix location reporting in parse errors

### DIFF
--- a/packages/datalog/datalog.0.7/opam
+++ b/packages/datalog/datalog.0.7/opam
@@ -13,7 +13,7 @@ depends: [
   "mdx" {>= "1.3" & with-test}
 ]
 build: [
-  ["dune" "subst"] {pinned}
+  ["dune" "subst"] {dev}
   [
     "dune"
     "build"
--- a/packages/datalog/datalog.0.7/opam
+++ b/packages/datalog/datalog.0.7/opam
@@ -1,0 +1,38 @@
+opam-version: "2.0"
+synopsis: "An in-memory datalog implementation for OCaml"
+maintainer: ["simon.cruanes.2007@m4x.org"]
+authors: ["Simon Cruanes"]
+license: "BSD-2-Clause"
+tags: ["datalog" "relational" "query" "prolog"]
+homepage: "https://github.com/c-cube/datalog"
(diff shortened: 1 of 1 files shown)


packages/datalog/datalog.0.7/opam
```diff
@@ -0,0 +1,38 @@
...
+homepage: "https://github.com/c-cube/datalog"
+bug-reports: "https://github.com/c-cube/datalog/issues"
+depends: [
+  "dune" {>= "2.0"}
+  "ocaml" {>= "4.08"}
+  "odoc" {with-doc}
+  "mdx" {>= "1.3" & with-test}
+]
+build: [
+  ["dune" "subst"] {pinned}
```
shonfeder, line 16:
```suggestion
  ["dune" "subst"] {dev}
```

shonfeder (commented):
Thank you for publishing the package update :)

I see several CI failures that look to be failing tests on some platforms. Could you take a look at those when you have a chance? 

We also have this dune subst thing. If you don't mind, I can just commit the fix in directly.

c-cube:
seems good to me, honestly. It seems that on x32 the tests sometimes fail because the hashtbl traversal order is different, which is not a big deal.

shonfeder:
> seems good to me, honestly. It seems that on x32 the tests sometimes fail because the hashtbl traversal order is different, which is not a big deal.

Since hey are not signs of problems, makes sense not to block on this. But fixing the tests so that they are reliable across different platforms will prevent having to face this kind of snag in the future. Would accept a PR to make the tests more robust across platforms?

---

Thanks very much for publishing the update, and of course for the nice library and utility :)

c-cube:
Sure, yeah, in this case it's 10 yo tests that I didn't bother making deterministic (by sorting) :)

PR #26760, opened 2024-10-19
hol_light: fix wrong paths in hol.sh, copy src to lib

aqjune:
This patch fixes hidden bugs in hol_light's opam which did not reveal when installing hol_light locally, but appear when installed from remote git.
Also, this patch copies the source files in the build directory to hol_light:lib because all .ml files can actually matter when hol.sh is used in HOL Light. My previous patch was my first contribution to OPAM, so these points were missed. :/ Since I didn't publicly announce the new HOL Light package, and this is a bug fix in its OPAM configuration, I believe (hope) incrementing its version to 3.0.1 is not necessary.

### DIFF
--- a/packages/hol_light/hol_light.3.0.0/opam
+++ b/packages/hol_light/hol_light.3.0.0/opam
@@ -58,19 +58,18 @@ available: os = "linux" | os = "macos" | os = "cygwin"
 build: [
   [make] {!hol_light_module:installed}
   [make "HOLLIGHT_USE_MODULE=1"] {hol_light_module:installed}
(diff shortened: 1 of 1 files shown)


aqjune:
The one remaining failure in CI is a reproduction of this: https://github.com/ocaml/opam-repository/pull/26745#issuecomment-2421522725 .. which is orthogonal to this change. This patch is ready to get reviewed.

mseri:
Sometimes we release as 3.0.0-1, but since this is broken, only a metadata change and not yet announced, I think it make sense to speedily update the build metadata

aqjune:
Thanks. :)

PR #26769, opened 2024-10-21
Add tophide.1.1.0

kit-ty-kate:
Adds OCaml 5.3 compatibility per https://github.com/mjambon/tophide/pull/5

### DIFF
--- a/packages/tophide/tophide.1.1.0/opam
+++ b/packages/tophide/tophide.1.1.0/opam
@@ -1,0 +1,23 @@
+opam-version: "2.0"
+maintainer: "andy.ray@ujamjar.com"
+authors: "Martin Jambon"
+homepage: "https://github.com/mjambon/tophide"
+bug-reports: "https://github.com/mjambon/tophide/issues"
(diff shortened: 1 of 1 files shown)


packages/tophide/tophide.1.1.0/opam
```diff
@@ -0,0 +1,23 @@
+opam-version: "2.0"
+maintainer: "andy.ray@ujamjar.com"
+authors: "Martin Jambon"
+homepage: "https://github.com/mjambon/tophide"
+bug-reports: "https://github.com/mjambon/tophide/issues"
+dev-repo: "git+https://github.com/mjambon/tophide"
+build: make
```
mseri, line 7:
```suggestion
license: "BSD-3-Clause"
build: make
```
I think this is the one in the repo

kit-ty-kate: this is not the same to the letter (close 3 is missing mention of the contributors) but close enough, i think the spirit is the same.

mseri:
Thanks

PR #26777, opened 2024-10-23
[new release] vcs (5 packages) (0.0.9)

mbarbin:
This is the initial release for 5 packages providing a type-safe and direct-style API to programmatically perform Git operations. More information at https://github.com/mbarbin/vcs. Short synopsis below:

## Synopsis

**vcs** the user facing library for the system, with reduced dependencies.

**vcs-git-provider** a helper library for building providers for the vcs library.

**vcs-git-blocking** a provider for the vcs library built with OCam's Stdlib and offering blocking calls to git.

**vcs-git-eio** a provider for the vcs library built atop Eio for non blocking calls to git.

**vcs-base** an extension of `Vcs` for base users.

### DIFF
--- a/packages/vcs-base/vcs-base.0.0.9/opam
+++ b/packages/vcs-base/vcs-base.0.0.9/opam
@@ -1,0 +1,49 @@
+opam-version: "2.0"
+synopsis: "An Extension of Vcs to use with Base"
+maintainer: ["Mathieu Barbin <opensource@mbarbin.org>"]
+authors: ["Mathieu Barbin"]
+license: "LGPL-3.0-or-later WITH LGPL-3.0-linking-exception"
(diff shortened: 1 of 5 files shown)


mbarbin:
A feedback from the CI:
- Warning in vcs-base.0.0.9: Possible name collision with package 'ojs-base'

Is it safe to ignore? I think the name are sufficiently different to avoid mixing one for the other.

Unrelated, but noting for next-step: in retrospect I want to check whether the non-eio packages of vcs are compatible with ocaml 4.14. I plan on looking into it before enabling for review.

mbarbin:
In light of recent discussions on the discuss forum about memory representation guarantees and Obj.magic, I will withdraw my submission for now, while I am investigating an alternative provider implementation that can be type safe. (provider being a dependency of vcs).

PR #26780, opened 2024-10-23
Fix install of conf-libclang.15 when using macports on macOS.

dboris:
This PR makes it possible to install this package on macOS when using macports.


dboris:
The configure script https://raw.githubusercontent.com/ocaml/opam-source-archives/main/patches/conf-libclang/configure.sh.15 may need a patch. Not sure how to provide it.

Macports installs `llvm-config` to /opt/local/bin/llvm-config

mseri:
Thanks

PR #26807, opened 2024-10-29
Package theora.0.4.1

toots:
### `theora.0.4.1`
Bindings to libtheora

---
* Homepage: https://github.com/savonet/ocaml-theora
* Source repo: git+https://github.com/savonet/ocaml-theora.git
* Bug tracker: https://github.com/savonet/ocaml-theora/issues

---


packages/theora/theora.0.4.1/opam
```diff
@@ -0,0 +1,38 @@
...
+bug-reports: "https://github.com/savonet/ocaml-theora/issues"
+depends: [
+  "conf-libtheora"
+  "conf-pkg-config"
+  "dune" {>= "2.0"}
+  "dune-configurator"
+  "ogg" {>= "0.7.0"}
+]
+build: [
+  ["dune" "subst"] {pinned}
```
shonfeder, line 17: To address the failing CI lint

```
Warning in theora.0.4.1: Dubious use of 'dune subst'. 'dune subst' should always only be called with {dev} (i.e. ["dune" "subst"] {dev}) If your opam file has been autogenerated by dune, you need to upgrade your dune-project to at least (lang dune 2.7).
```

```suggestion
  ["dune" "subst"] {dev}
```

toots: Thanks. Fixed and bumped dune version upstream.

packages/theora/theora.0.4.1/opam
```diff
@@ -0,0 +1,38 @@
+# This file is generated by dune, edit dune-project instead
+opam-version: "2.0"
+synopsis: "Bindings to libtheora"
+maintainer: ["The Savonet Team <savonet-users@lists.sourceforge.net>"]
+authors: ["The Savonet Team <savonet-users@lists.sourceforge.net>"]
+license: "GPL-2.0"
```
shonfeder, line 6: Sorry I didn't catch this before:

I think this is not a big deal for now now, but be we are getting a linting error here because this a deprecated license identifier (see the last section of https://spdx.org/licenses/).

If you can switch to one of the current identifiers in your code base (e.g., `GPL-2.0-or-later` or `GPL-2.0-only`) here and/or in your code base, we won't hit this next time you do an update.

shonfeder:
The only CI failure is unrelated to this package (something up with dune on 4.02, I'll investigate).

shonfeder:
Thanks for publishing the update!

packages/theora/theora.0.4.1/opam
toots: This has been fixed, thanks!

PR #26810, opened 2024-10-30
gptar.1.0.0: remove excess whitespace in description

reynir:
This is not a big problem, but I noticed the opam file I published has a loooot of leading spaces. This looks silly in `opam show gptar`.

This removes the excess white space. I hope this does not cause issues to modify the opam file after it's published. Please feel free to reject this PR in that case.


mseri:
Thanks

PR #26811, opened 2024-10-30
Add Mingw support to conf-sdl2

mtelvers:


shonfeder (commented):
Thank you for these fixes for windows compat!

PR #26813, opened 2024-10-30
Add Mingw support for conf-libpcre

mtelvers:


mseri:
I am not sure what is wrong, seems something with the package selection. There is a new attempt [here](https://github.com/ocaml/opam-repository/pull/26893) that seems to work.

PR #26816, opened 2024-10-30
[new release] dune (15 packages) (3.16.1)

maiste:
Fast, portable, and opinionated build system

- Project page: <a href="https://github.com/ocaml/dune">https://github.com/ocaml/dune</a>
- Documentation: <a href="https://dune.readthedocs.io/">https://dune.readthedocs.io/</a>

### Fixed

- Call the C++ compiler with `-std=c++11` when using OCaml >= 5.0
  (ocaml/dune#10962, @kit-ty-kate)

PR #26821, opened 2024-10-31
Mark `rml` unavailable as sources are unavailable

Leonidas-from-XIV:
`rml.lri.fr` and `reactiveml.org` are both dead/redirecting to Github and the tarballs are 404'ing. The OPAM cache doesn't have them so the best bet seems to point to the Github releases.

I've noticed that while inspecting the revdeps of Dune in #26816.

### DIFF
--- a/packages/rml/rml.1.09.07/opam
+++ b/packages/rml/rml.1.09.07/opam
@@ -13,7 +13,7 @@ install: [
  [make "install"]
 ]
 depends: [
-  "ocaml" {>= "4.07" & < "5.1"}
+  "ocaml" {>= "4.07"}
   "num"
 ]
 depexts: ["patch"] {os-distribution = "alpine"}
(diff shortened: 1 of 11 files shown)


Leonidas-from-XIV:
After running CI it turns out that most (but not all) archives don't have `configure` so can't be built even when the URLs and checksums are updated, so I marked them as `unavailable` since they can't be built unless someone wants to step up and fix the build instructions.

The newest release works with the build instructions, but fails on OCaml 5.1+ so I've updated the constraints.

packages/rml/rml.1.09.07/opam
```diff
@@ -27,9 +27,9 @@ In ReactiveML, the reactive model is integrated at the language level
 (not as a library) which leads to safer and more natural programming."""
 authors: "Louis Mandel louis@reactiveml.org"
 url {
-  src: "http://rml.lri.fr/distrib/rml-1.09.07-2021-07-26.tar.gz"
+  src: "https://github.com/reactiveml/rml/archive/refs/tags/rml-1.09.07-2021-07-26.tar.gz"
   checksum: [
-    "sha256=74acc5319df33b4694d6894b0b8023c1ce3f8f0749c6f4dc587a42c102d809b0"
-    "md5=4fc5ca8f941f6a7f060fb9252389b8b4"
+    "sha256=5beeb70dbd8688597999377b54342c1306ac92a3eefb68978a38311c531de5ca"
```
mseri, line 32: If the hashes have changed, this should also be marked as unavailable, unless we can find one in the cache to compare the contents

Leonidas-from-XIV: Ok, I undid the changes and just marked it as unavailable.

shonfeder:
Thanks for this maintenance! Only linting checks are failing, but I don't think it should be an issue for an unavailable package.

PR #26823, opened 2024-10-31
Older datalog versions FTBFS on newer OCaml

Leonidas-from-XIV:
Something about how it is built is wrong, a mix of `-for-pack` and not `-for-pack` causes it to fail:

```
File "datalog_cli.ml", line 1:
Error: "datalog/lexer.cmx" was built with "-for-pack Datalog", but the
       current unit "Datalog_cli" is not
Command exited with code 2.
```

This seems to work on 4.14 and 5.0, but not on 5.1. Newer versions of Datalog that use Dune don't seem to be affected.

[Example of failing build](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/ed800ba87cfbe293bcb59dfb40db5252bc65f6f5/variant/compilers,5.2,dune-configurator.3.16.1,revdeps,datalog.0.1).

### DIFF
--- a/packages/datalog/datalog.0.1/opam
+++ b/packages/datalog/datalog.0.1/opam
@@ -7,7 +7,7 @@ remove: [
   ["rm" "%{bin}%/datalog_cli"]
 ]
 depends: [
-  "ocaml"
+  "ocaml" {< "5.1"}
   "ocamlfind"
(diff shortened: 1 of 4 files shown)


mseri:
Thanks!

PR #26473, opened 2024-09-03
[new release] picos (8 packages) (0.5.0)

polytypic:
Pico scheduler interface

- Project page: <a href="https://github.com/ocaml-multicore/picos">https://github.com/ocaml-multicore/picos</a>

##### CHANGES:

- Major additions, changes, bug fixes, improvements, and restructuring
  (@polytypic, @c-cube)

  - Additions:

    - Minimalistic Cohttp implementation
    - Implicitly propagated `Flock` of fibers for structured concurrency
    - Option to terminate `Bundle` and `Flock` on return
    - `Event` abstraction
    - Synchronization and communication primitives:
      - Incremental variable or `Ivar`
      - Countdown `Latch`
      - `Semaphore`
      - `Stream` of events
    - Multi-producer, multi-consumer lock-free queue optimized for schedulers
    - Multithreaded (work-stealing) FIFO scheduler
    - Support `quota` for FIFO based schedulers
    - Transactional interface for atomically completing multiple `Computation`s

  - Changes:

    - Redesigned resource management based on `('r -> 'a) -> 'a` functions
    - Redesigned `spawn` interface allowing `FLS` entries to be populated before
      spawn
    - Introduced concept of fatal errors, which must terminate the scheduler or
      the whole program
    - Simplified `FLS` interface

  - Improvements:

    - Signficantly reduced per fiber memory usage of various sample schedulers

  - Picos has now been split into multiple packages and libraries:

    - pkg: `picos`
      - lib: `picos`
      - lib: `picos.domain`
      - lib: `picos.thread`
    - pkg: `picos_aux`
      - lib: `picos_aux.htbl`
      - lib: `picos_aux.mpmcq`
      - lib: `picos_aux.mpscq`
      - lib: `picos_aux.rc`
    - pkg: `picos_lwt`
      - lib: `picos_lwt`
      - lib: `picos_lwt.unix`
    - pkg: `picos_meta` (integration tests)
    - pkg: `picos_mux`
      - lib: `picos_mux.fifo`
      - lib: `picos_mux.multififo`
      - lib: `picos_mux.random`
      - lib: `picos_mux.thread`
    - pkg: `picos_std`
      - lib: `picos_std.event`
      - lib: `picos_std.finally`
      - lib: `picos_std.structured`
      - lib: `picos_std.sync`
    - pkg: `picos_stdio`
      - lib: `picos_stdio`
      - lib: `picos_stdio.fd`
      - lib: `picos_stdio.select`
[...]

PR #26825, opened 2024-11-01
[new release] picos (8 packages) (0.6.0)

polytypic:
Pico scheduler interface

- Project page: <a href="https://github.com/ocaml-multicore/picos">https://github.com/ocaml-multicore/picos</a>

##### CHANGES:

- Added a futex-like `Awaitable` abstraction as the `picos_std.awaitable`
  library (@polytypic)
- Changed the core Picos library to be internally built from a single `.ml` file
  (@polytypic)
- Optimized heap and stack usage of fibers and resource cleanup mechanisms and
  added workarounds for compiler generated space leaks due to closures
  (@polytypic)
- Added `lastly` as a safe alternative to `Fun.protect` (@polytypic)
- Workarounds for the `Uri` library not being threadsafe (@polytypic)
- Fixed to raise proper error when `Picos_io_select` has not been configured
  properly (@polytypic)
- Forbid cancelation propagation during `release` calls in the
  `picos_std.finally` library (@polytypic)
  - This is a change in behaviour and could be seen as a breaking change, but it
    should really be considered a bug fix.
- Renamed `(Ivar|Stream).poison` to `(Ivar|Stream).poison_at` and added
  `(Ivar|Stream).poison` with optional `?callstack:int` (@polytypic)


polytypic:
Oh... There seems to be package level dependency issues (those are not checked by OCaml-CI).  I'll fix that and update.

polytypic:
The request resolution timeout

```
        Try to specify a more precise request, use a different solver, or increase the allowed time by setting OPAMSOLVERTIMEOUT to a bigger value (currently, it is set to 1000.0 seconds).
```

probably due to dependency to cohttp happened also on previous release.  See [here](https://github.com/ocaml/opam-repository/pull/26473#issuecomment-2329456423).

mseri:
Thanks

PR #26837, opened 2024-11-04
[new release] Camlkit v0.3.0

dboris:
New version release.


shonfeder:
Looks great! Thanks for publishing the update.

dboris:
I wonder why docs are shown as not available on https://ocaml.org/p/camlkit-base/latest
`dune build @doc` should work, and I added a `docs` directory to the source. Is it because the package is not available on Linux?

shonfeder:
@dboris it's a problem with the documentation service. See: https://github.com/ocurrent/ocaml-docs-ci/issues/176

PR #26838, opened 2024-11-05
[new release] postgresql (5.1.0)

mmottl:

##### CHANGES:

- Made Postgresql.null a now unique, empty string.

- Fixed license format and a typo.

- Fixed some odoc references.

- Added GitHub workflows.

- Added automatic formatting with ocamlformat and clang-format.

- Ported the config discovery script to pkg-config

  Thanks to Antonio Nuno Monteiro for the patch.

- Improved Dune rules.

- Used new OCaml 4.12 C-macros.

- Switched to Dune lang 2.7.

### DIFF
--- a/packages/conf-mingw-w64-postgresql-i686/conf-mingw-w64-postgresql-i686.1/opam
+++ b/packages/conf-mingw-w64-postgresql-i686/conf-mingw-w64-postgresql-i686.1/opam
@@ -1,0 +1,19 @@
+opam-version: "2.0"
+synopsis: "postgresql for i686 mingw-w64 (32-bit x86)"
+description: "Ensures the i686 version of postgresql for the mingw-w64 project is available"
+maintainer: "David Allsopp <david@tarides.com>"
+authors: "Markus Mottl"
+license: "blessing"
+homepage: "http://www.postgresql.org"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+flags: conf
+available: os = "win32"
+build: ["pkgconf" "--personality=i686-w64-mingw32" "postgresql"]
+depends: [
+  "conf-pkg-config" {build}
+  "conf-mingw-w64-gcc-i686" {build}
+]
+depexts: [
+  ["mingw64-i686-postgresql"] {os = "win32" & os-distribution = "cygwin"}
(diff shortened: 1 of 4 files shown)


mmottl:
I hope this new release will work fine with the transition to `pkg-config`. It is likely that `conf-postgresql` will need to be adjusted for different platforms, but it's a start.

Here are more details about the change: https://github.com/mmottl/postgresql-ocaml/pull/48

packages/conf-mingw-w64-postgresql-i686/conf-mingw-w64-postgresql-i686.1/opam
```diff
@@ -0,0 +1,19 @@
+opam-version: "2.0"
+synopsis: "postgresql for i686 mingw-w64 (32-bit x86)"
```
mseri, line 2:
```suggestion
synopsis: "Postgresql for i686 mingw-w64 (32-bit x86)"
```

packages/conf-mingw-w64-postgresql-x86_64/conf-mingw-w64-postgresql-x86_64.1/opam
```diff
@@ -0,0 +1,19 @@
+opam-version: "2.0"
+synopsis: "postgresql for x86_64 mingw-w64 (64-bit x86_64)"
```
mseri, line 2:
```suggestion
synopsis: "Postgresql for x86_64 mingw-w64 (64-bit x86_64)"
```

mseri:
The failure on macos is to be expected unfortunately
```
- libpq is keg-only, which means it was not symlinked into /opt/homebrew,
- because conflicts with postgres formula.
- 
- If you need to have libpq first in your PATH, run:
-   echo 'export PATH="/opt/homebrew/opt/libpq/bin:$PATH"' >> ~/.profile
- 
- For compilers to find libpq you may need to set:
-   export LDFLAGS="-L/opt/homebrew/opt/libpq/lib"
-   export CPPFLAGS="-I/opt/homebrew/opt/libpq/include"
- 
- For pkg-config to find libpq you may need to set:
-   export PKG_CONFIG_PATH="/opt/homebrew/opt/libpq/lib/pkgconfig"
```
but I don't think we can do better

avsm:

PR #26841, opened 2024-11-05
[new release] http-mirage-client (0.0.8)

hannesm:
HTTP client for MirageOS

- Project page: <a href="https://github.com/robur-coop/http-mirage-client">https://github.com/robur-coop/http-mirage-client</a>

##### CHANGES:

- Do not call [f_body] with a redirect body (robur-coop/http-mirage-client#2 @hannesm)

### DIFF
--- a/packages/http-mirage-client/http-mirage-client.0.0.8/opam
+++ b/packages/http-mirage-client/http-mirage-client.0.0.8/opam
@@ -1,0 +1,51 @@
+opam-version: "2.0"
+synopsis: "HTTP client for MirageOS"
+maintainer: ["team@robur.coop"]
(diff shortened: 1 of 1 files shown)


hannesm:
fine to merge

mseri:
Do you know why the two opam-2.1 tests are failing? Is it something to do with the sandbox?

hannesm:
I don't, but they had before for this package.

now, the test failure is ENOENT on some file where the test results should be in -- but the tests actually open a socket on localhost and communicate through that. so, indeed I don't know what happened between opam 2.1 and opam 2.2 in this respect.

PR #26842, opened 2024-11-05
2 packages from bcpierce00/unison at 2.53.7

jhjourdan:
This pull-request concerns:
- `unison.2.53.7`: File-synchronization tool for Unix and Windows
- `unison-gui.2.53.7`: Pseudo-package for Unison GUI

---
* Homepage: https://www.cis.upenn.edu/~bcpierce/unison/
* Source repo: git://github.com/bcpierce00/unison.git
* Bug tracker: https://github.com/bcpierce00/unison/issues

---

### DIFF
--- a/packages/unison-gui/unison-gui.2.53.7/opam
+++ b/packages/unison-gui/unison-gui.2.53.7/opam
@@ -1,0 +1,31 @@
+opam-version: "2.0"
+maintainer: "unison-hackers@lists.seas.upenn.edu"
+authors: [
+  "Trevor Jim"
+  "Benjamin C. Pierce"
+  "Jérôme Vouillon"
+]
+license: "GPL-3.0-or-later"
+homepage: "https://www.cis.upenn.edu/~bcpierce/unison/"
+bug-reports: "https://github.com/bcpierce00/unison/issues"
+dev-repo: "git://github.com/bcpierce00/unison.git"
+depends: [
+  "unison" {= version}
+  "lablgtk3"
+  "ocamlfind"
+]
+synopsis: "Pseudo-package for Unison GUI"
+description: """
+Unison is a file-synchronization tool for Unix and Windows. It allows
+two replicas of a collection of files and directories to be stored on
+different hosts (or different disks on the same host), modified
+separately, and then brought up to date by propagating the changes in
+each replica to the other."""
+url {
+  src:
+    "https://github.com/bcpierce00/unison/archive/refs/tags/v2.53.7.tar.gz"
+  checksum: [
+    "md5=e49a78c1e159d4d2315fb5264c83ed59"
+    "sha512=11bd1d2792fb84fa2b29426516d7b91a2295febcb84052118d492a29d9ddaa23265b66cc88113019ec6782f3edcf596f7a37c8637e673b2928188248b9d63d60"
+  ]
+}
--- a/packages/unison/unison.2.53.7/opam
+++ b/packages/unison/unison.2.53.7/opam
@@ -11,6 +11,7 @@ bug-reports: "https://github.com/bcpierce00/unison/issues"
 dev-repo: "git://github.com/bcpierce00/unison.git"
 build: [make "NATIVE=%{ocaml:native}%" "-j" jobs]
 install: [make "NATIVE=%{ocaml:native}%" "PREFIX=%{prefix}%" "install"]
+available: os != "win32"
 depends: [
   "ocaml" {>= "4.08"}
 ]
--- a/packages/unison/unison.2.53.7/opam
+++ b/packages/unison/unison.2.53.7/opam
@@ -1,0 +1,35 @@
+opam-version: "2.0"
+maintainer: "unison-hackers@lists.seas.upenn.edu"
+authors: [
+  "Trevor Jim"
+  "Benjamin C. Pierce"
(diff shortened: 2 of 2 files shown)


jhjourdan:
Is there a way to silence the lint about the missing dependency on Dune? There is a dune file, but the opam package file does not depend on it.

In addition, the package does not seem to work on Windows. I don't have a Windows machine to test and fix this, so is there a way to just tell that this package is not wompatible with Windows ?

raphael-proust:
> is there a way to just tell that this package is not wompatible with Windows ?

You can add `available: os != "win32"` to the opam file.

packages/unison-gui/unison-gui.2.53.7/opam
```diff
@@ -0,0 +1,31 @@
...
+separately, and then brought up to date by propagating the changes in
+each replica to the other."""
+url {
+  src:
+    "https://github.com/bcpierce00/unison/archive/refs/tags/v2.53.7.tar.gz"
+  checksum: [
+    "md5=e49a78c1e159d4d2315fb5264c83ed59"
+    "sha512=11bd1d2792fb84fa2b29426516d7b91a2295febcb84052118d492a29d9ddaa23265b66cc88113019ec6782f3edcf596f7a37c8637e673b2928188248b9d63d60"
+  ]
+}
```
raphael-proust, line 31:
```suggestion
}
available: os != "win32"
```

packages/unison/unison.2.53.7/opam
```diff
@@ -0,0 +1,35 @@
...
+separately, and then brought up to date by propagating the changes in
+each replica to the other."""
+url {
+  src:
+    "https://github.com/bcpierce00/unison/archive/refs/tags/v2.53.7.tar.gz"
+  checksum: [
+    "md5=e49a78c1e159d4d2315fb5264c83ed59"
+    "sha512=11bd1d2792fb84fa2b29426516d7b91a2295febcb84052118d492a29d9ddaa23265b66cc88113019ec6782f3edcf596f7a37c8637e673b2928188248b9d63d60"
+  ]
+}
```
raphael-proust, line 35:
```suggestion
}
available: os != "win32"
```

raphael-proust:
> Is there a way to silence the lint about the missing dependency on Dune?

I don't think that's possible but we can just ignore the error.

raphael-proust:
Re: the windows failure, `pkg-config` has a different name on windows. I think you can fix it by changing the call you make on windows:
- call `pkgconf` (instead of `pkg-config`
- add the `--personality=x86_64-w64-mingw32` (or other personality value)

You can do that directly in the opam file:

```
build: [
  ["pkgconf" {os = "win32" & os-distribution = "cygwin"}
   "--personality=i686-w64-mingw32" {os = "win32" & os-distribution = "cygwin" & host-arch-x86_32:installed}
   "--personality=x86_64-w64-mingw32" {os = "win32" & os-distribution = "cygwin" & host-arch-x86_64:installed}
   "pkg-config" {os != "win32" | os-distribution != "cygwin"}
   REST OF THE PKGCONF ARGUMENTS HERE   
  ]
  REST OF THE BUILD INSTRUCTIONS HERE
]
```

You might be able to get away with a `$(PKGCONF)` variable if you are using a Makefile too.

But setting as unavailable on windows (as per the code suggestions above) is also ok.

avsm:
> Is there a way to silence the lint about the missing dependency on Dune?

You can add `dune` to the `depopts` field instead. This indicates that the package will be recompiled if `dune` appears in the future, but is not required to install it without dune present.

jhjourdan:
@raphael-proust Unison in itself does not directly depend on pkg-config. Instead, conf-gtk3 (and, I guess, lablgtk3) does.

Has anyone published a package using lablgtk3 working on Windows?

In the meantime, I pushed a version where Windows is disabled.

jhjourdan:
The build failure is suprious:

```
fatal: unable to access 'https://github.com/ocaml/opam-repository.git/': Could not resolve host: github.com
```

packages/unison-gui/unison-gui.2.53.7/opam
```diff
@@ -0,0 +1,31 @@
...
+maintainer: "unison-hackers@lists.seas.upenn.edu"
+authors: [
+  "Trevor Jim"
+  "Benjamin C. Pierce"
+  "Jérôme Vouillon"
+]
+license: "GPL-3.0-or-later"
+homepage: "https://www.cis.upenn.edu/~bcpierce/unison/"
+bug-reports: "https://github.com/bcpierce00/unison/issues"
+dev-repo: "git://github.com/bcpierce00/unison.git"
```
mseri, line 11: I think something is off, this package does not have `build` and `install` instructions, right?

jhjourdan: Indeed: it simply has an additional dependency on lablgtk. The makefile of the main package decides to build the GUI depending on whether lablgtk is alrady installed.

jhjourdan: That's the reason why lablgtk is a depopt of unison

mseri: Oh I see, so that indeed forces a recompilation. Thanks!

PR #26845, opened 2024-11-07
[new release] sel (0.5.0)

gares:
Simple Event Library

- Project page: <a href="https://github.com/gares/sel">https://github.com/gares/sel</a>

##### CHANGES:

- Resilient to multiple copies of the same event on the same file descriptor
- A system event is given enough rounds to complete if his priority is the lowest


avsm:

gares:
Thanks

PR #26853, opened 2024-11-08
Frama-C: new beta release (30.0~beta-Zinc)

vprevosto:


shonfeder:
Hitting a syntax error on ocaml 5.3:

```
# Error: Syntax error
# (cd _build/default/src/plugins/e-acsl/contrib/libdlmalloc && /usr/bin/make)
# CC           dlmalloc.c
# AR           libeacsl-dlmalloc.a
# RANLIB       libeacsl-dlmalloc.a
# ar: `u' modifier ignored since `D' is the default (see `U')
# (cd _build/.sandbox/a998f1bcf9c0cafa5851c393866115da/default && .ppx/fee8513703f5c2401b04afb573c16d6a/ppx.exe --cookie 'library-name="eva"' -o src/plugins/eva/partitioning/auto_loop_unroll.pp.ml --impl src/plugins/eva/partitioning/auto_loop_unroll.ml -corrected-suffix .ppx-corrected -diff-cmd - -dump-ast)
# File "src/plugins/eva/partitioning/auto_loop_unroll.ml", line 169, characters 23-29:
# 169 | let add_written_var vi effect =
#                              ^^^^^^
# Error: Syntax error
```

Do you want an upper bound on supported ocaml versions for this version of the package or to fix the source to avoid the `effect` keyword?

vprevosto:
> Hitting a syntax error on ocaml 5.3:
> 

> 
> Do you want an upper bound on supported ocaml versions for this version of the package or to fix the source to avoid the `effect` keyword?

Thanks for the report. For this version, it is indeed probably best to say that 5.3 is not supported, and we will make sure that the final 30.0 release is compatible with OCaml 5.3.

shonfeder:
Excellent. Thanks! The other CI errors do not look relevant: solver timeout on old opam versions and what looks like an internal CI errors.

Thank you again for publishing!

PR #26860, opened 2024-11-10
Package malfunction.0.7

stedolan:
(From this version onwards, I've started putting explicit upper bounds on `ocaml` in the opam file from release - Malfunction is extremely unlikely to ever work with an untested OCaml version since it uses so much of compiler-libs)


shonfeder (approved):
Thank you publishing the updates! :) Please consider announcing the update on discuss.ocaml.org :camel:

PR #26864, opened 2024-11-11
Package ocamlfind.1.9.7

kit-ty-kate:
### `ocamlfind.1.9.7`
A library manager for OCaml
Findlib is a library manager for OCaml. It provides a convention how
to store libraries, and a file format ("META") to describe the
properties of libraries. There is also a tool (ocamlfind) for
interpreting the META files, so that it is very easy to use libraries
in programs and scripts.

---
* Homepage: http://projects.camlcity.org/projects/findlib.html
* Source repo: git+https://github.com/ocaml/ocamlfind.git
* Bug tracker: https://github.com/ocaml/ocamlfind/issues

---

### DIFF
--- a/packages/ocamlfind/ocamlfind.1.9.7/opam
+++ b/packages/ocamlfind/ocamlfind.1.9.7/opam
@@ -1,11 +1,12 @@
 opam-version: "2.0"
 synopsis: "A library manager for OCaml"
-description: """\
+description: """
 Findlib is a library manager for OCaml. It provides a convention how
 to store libraries, and a file format ("META") to describe the
 properties of libraries. There is also a tool (ocamlfind) for
 interpreting the META files, so that it is very easy to use libraries
-in programs and scripts."""
+in programs and scripts.
+"""
 maintainer: "Thomas Gazagnaire <thomas@gazagnaire.org>"
 authors: "Gerd Stolpmann <gerd@gerd-stolpmann.de>"
 license: "MIT"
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
cc @dra27

packages/ocamlfind/ocamlfind.1.9.7/opam
```diff
@@ -0,0 +1,49 @@
...
+    "-bindir"
+    bin
+    "-sitelib"
+    lib
+    "-mandir"
+    man
+    "-config"
+    "%{lib}%/findlib.conf"
+    "-with-relative-paths-at"
+    prefix
```
dra27, line 30: IIUC this causes `ocamlfind` to depend on the `OPAM_SWITCH_PREFIX` environment variable being correct which seems potentially more brittle than the 1.9.6 and previous approach which doesn't depend on them. At least for now, it would sensible to remove this setting?

dra27 (commented):
While it definitely works well for Dune developer preview, for current opam-repository, I can only see disadvantage to using `-with-relative-paths-at`.

Worth updating ocamlfind-secondary with it as well, although if there's a 1.9.8 coming soon, that could perhaps wait.

mseri:
Let us know when this is ready

hhugo:
@dra27, @kit-ty-kate, what's the status of this ? There are few fixes for windows in this release, can we do something to have it available sooner than later ?

kit-ty-kate:
Superseded by https://github.com/ocaml/opam-repository/pull/27253

PR #26871, opened 2024-11-12
add selfie 0.1

mpu:
An experimental process snapshotting library.

The build of `selfie.c` fails on alpine because of a missing include file `asm/prctl.h`, so I made the package unavailable for this distribution.


mseri:
Consider announcing the release on discuss.ocaml.org. We have a community category and an announcement tag for this purpose.

PR #26878, opened 2024-11-14
Replace pkg-config with pkgconf when the distribution is brew

emjin:
Homebrew has deprecated pkg-config. `brew install pkg-config` now installs pkgconf. Accordingly, this PR adjusts the expectations of `conf-pkg-config` to depend on `pkgconf`.


mseri:
Thanks a lot! I have merged this with #26891

PR #26886, opened 2024-11-14
Patch to fix async_ssl build

raphael-proust:
DO NOT MERGE: wait until the CI succeeds, then merge https://github.com/ocaml/opam-source-archives/pull/34, then change the URL for the patch, then merge

### DIFF
--- a/packages/async_ssl/async_ssl.v0.15.0/opam
+++ b/packages/async_ssl/async_ssl.v0.15.0/opam
@@ -6,7 +6,6 @@ bug-reports: "https://github.com/janestreet/async_ssl/issues"
 dev-repo: "git+https://github.com/janestreet/async_ssl.git"
 doc: "https://ocaml.janestreet.com/ocaml-core/latest/doc/async_ssl/index.html"
 license: "MIT"
-patches: [ "no-incompatible-pointer-types.patch" ]
 build: [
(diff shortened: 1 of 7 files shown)


mseri:
I have merged the patches before checking this, sorry

mseri:
But the build of async_ssl is fine on the whole matrix, there are only a few ssl issues that are unrelated as far as I can see

mseri:
Should we make this a patch `-1` release instead? Or it will mess up with strict versioning of some packages?

raphael-proust:
> Or it will mess up with strict versioning of some packages?

How so? I guess I can make a script that would break with this update, but I'm not sure there are standard tools around opam that would break.

But yeah, we might as well fix it just in case. Question is: should we mark the unpatched package as not available because it's broken several plateforms?

mseri:
What I mean is if there are janestreet packages with version v0.17 requiring async_ssl with the `=version` flag. In this case I think we should merge as is, if not, I think we should merge a `-1` patch. In the latter case I would not mark anything as unavailable because there could be still people happyly using them which do not need or want to move to the new one

raphael-proust:
I see. Done, it's now a patch version (which makes sense because it's a patch I guess).

mseri:
Tests have already run, the lint and many tests have already passed

mseri:
Thanks

PR #26890, opened 2024-11-14
Octez 21.0 packages

killian-delarue:
Packages for Octez v21.0

### DIFF
--- a/packages/octez-client/octez-client.21.0/opam
+++ b/packages/octez-client/octez-client.21.0/opam
@@ -1,0 +1,79 @@
+opam-version: "2.0"
+maintainer: "contact@tezos.com"
+authors: ["Tezos devteam"]
+homepage: "https://www.tezos.com/"
+bug-reports: "https://gitlab.com/tezos/tezos/issues"
+dev-repo: "git+https://gitlab.com/tezos/tezos.git"
+license: "MIT"
+depends: [
+  "dune" { >= "3.11.1" }
+  "ocaml" { >= "4.14" }
+  "octez-libs" { = version }
+  "octez-shell-libs" { = version }
+  "uri" { >= "3.1.0" }
+  "octez-protocol-020-PsParisC-libs" { = version }
+  "octez-protocol-021-PsQuebec-libs" { = version }
+]
+depopts: [
+  "tezos-client-genesis"
+  "tezos-client-demo-counter"
+  "octez-protocol-000-Ps9mPmXa-libs"
+  "octez-protocol-001-PtCJ7pwo-libs"
+  "octez-protocol-002-PsYLVpVv-libs"
+  "octez-protocol-003-PsddFKi3-libs"
+  "octez-protocol-004-Pt24m4xi-libs"
+  "octez-protocol-005-PsBabyM1-libs"
+  "octez-protocol-006-PsCARTHA-libs"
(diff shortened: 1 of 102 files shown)


raphael-proust:
Are the packages expected to run on windows? The windows CI fails but it seems to be network related or something. Idk. Still, should these packages have an `available:` clause with an `os` condition?

raphael-proust:
`octez-rust-deps` fails to build:

```
#=== ERROR while compiling octez-rust-deps.21.0 ===============================#
# context              2.4.0~alpha1~dev | linux/x86_64 | ocaml-base-compiler.4.14.2 | pinned(https://gitlab.com/tezos/tezos/-/archive/octez-v21.0/octez-source-21.0.tar.gz)
# path                 ~/.opam/4.14/.opam-switch/build/octez-rust-deps.21.0
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p octez-rust-deps -j 39
# exit-code            1
# env-file             ~/.opam/log/octez-rust-deps-7-ec9a10.env
# output-file          ~/.opam/log/octez-rust-deps-7-ec9a10.out
### output ###
# File "src/rust_deps/dune", lines 12-28, characters 0-515:
# 12 | (rule
# 13 |  (targets liboctez_rust_deps.a dlloctez_rust_deps.so wasmer.h wasm.h)
# 14 |  (deps
# ....
# 26 |   (file ../../src/lib_wasm_runtime/build.rs)
# 27 |   (glob_files_rec ../../src/lib_wasm_runtime/src/**.rs))
# 28 |  (action (no-infer (bash ./build.sh))))
# (cd _build/default/src/rust_deps && /usr/bin/bash -e -u -o pipefail -c ./build.sh)
# ./build.sh: line 20: cargo: command not found
```

Do you need to add a dependency to `conf-rust` or something like that?

packages/octez-client/octez-client.21.0/opam
```diff
@@ -0,0 +1,79 @@
...
+  "octez-protocol-016-PtMumbai-libs"
+  "octez-protocol-017-PtNairob-libs"
+  "octez-protocol-018-Proxford-libs"
+  "octez-protocol-019-PtParisB-libs"
+  "octez-protocol-alpha-libs"
+]
+conflicts: [
+  "tezos-client-genesis" { != version }
+  "tezos-client-demo-counter" { != version }
+  "octez-protocol-000-Ps9mPmXa-libs" { != version }
```
mseri, line 45: It is easier also for the solver is these `!= version` would be replaced by `=version` in the dependency list itself

killian-delarue:
> Are the packages expected to run on windows? The windows CI fails but it seems to be network related or something. Idk. Still, should these packages have an `available:` clause with an `os` condition?

Something like in eeb075424fc9c4eab4425d043dea81112b4b9754 ?

raphael-proust:
> Something like

Yes. Also you should probably add these constraints in the upstream project. That way it won't be an issue for the next release.

killian-delarue:
> > Something like
> 
> Yes. Also you should probably add these constraints in the upstream project. That way it won't be an issue for the next release.

It will 👍🏻 Thank you for the suggestion

shonfeder:
We will close to clear out our backlog of stalled PRs. But please know that we remain available to support you if you when you are ready to resume the publication!

PR #26893, opened 2024-11-14
Make sure that conf-libpcre is available for windows

yosefAlsuhaibani:
It seems like the package `conf-libpcre` isn't available for windows when trying to install dependancies via `setup-ocaml@v3+`. I followed the instructions here (https://opam.ocaml.org/blog/opam-2-2-0-windows/#External-Dependencies-quot-depexts-quot:~:text=External%20Dependencies%20(%22depexts%22)) to add support for a package (similar to the conf-openssl package update mentioned in the release notes). 

I tested this in my own project via first forking this repo, adding these changes, and pinning my fork as the default opam-repo when running `setup-ocaml`, and the build on windows seems to pass.

PR #26894, opened 2024-11-14
Update windows build instructions for conf-libcurl 

yosefAlsuhaibani:
related to: #26893

This just updates the build instructions for windows to use `pkgconf` instead of `curl-conf`, which I ran into issues with when using `setup-ocaml`. Similar testing plan to #26893.

### DIFF
--- a/packages/conf-libcurl/conf-libcurl.3/opam
+++ b/packages/conf-libcurl/conf-libcurl.3/opam
@@ -4,7 +4,10 @@ authors: ["Daniel Stenberg"]
 homepage: "http://curl.haxx.se/"
 license: "BSD-like"
 build: [
-  ["sh" "-exc" "curl-config --libs"] {os = "win32" & os-distribution != "cygwinports"}
+  ["pkgconf" {os = "win32" & os-distribution = "cygwin"}
+   "--personality=i686-w64-mingw32" {os = "win32" & os-distribution = "cygwin" & host-arch-x86_32:installed}
+   "--personality=x86_64-w64-mingw32" {os = "win32" & os-distribution = "cygwin" & host-arch-x86_64:installed}
+   "libcurl" {os = "win32" & os-distribution = "cygwin"}]
   ["curl-config" "--libs"] {os != "win32" | os-distribution = "cygwinports"}
 ]
 depexts: [
--- a/packages/conf-libcurl/conf-libcurl.2/opam
+++ b/packages/conf-libcurl/conf-libcurl.2/opam
@@ -4,7 +4,11 @@ authors: ["Daniel Stenberg"]
 homepage: "http://curl.haxx.se/"
 license: "BSD-like"
 build: [
-  ["sh" "-exc" "curl-config --libs"] {os = "win32" & os-distribution != "cygwinports"}
+  ["pkgconf" {os = "win32" & os-distribution = "cygwin"}
+   "--personality=i686-w64-mingw32" {os = "win32" & os-distribution = "cygwin" & host-arch-x86_32:installed}
+   "--personality=x86_64-w64-mingw32" {os = "win32" & os-distribution = "cygwin" & host-arch-x86_64:installed}
+   "pkg-config" {os != "win32" | os-distribution != "cygwin"}
+   "--print-errors" "--exists" "lbcurl"]
   ["curl-config" "--libs"] {os != "win32" | os-distribution = "cygwinports"}
 ]
 depexts: [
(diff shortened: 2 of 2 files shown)


yosefAlsuhaibani:
Seems like I missed the mark with this PR: I'll work on this a little bit more.

yosefAlsuhaibani:
Alright should be ready for review (@mseri)

packages/conf-libcurl/conf-libcurl.3/opam
```diff
@@ -0,0 +1,37 @@
+opam-version: "2.0"
+maintainer: "blue-prawn"
+authors: ["Daniel Stenberg"]
+homepage: "http://curl.haxx.se/"
+license: "BSD-like"
+build: [
+  ["pkgconf" {os = "win32" & os-distribution = "cygwin"}
+   "--personality=i686-w64-mingw32" {os = "win32" & os-distribution = "cygwin" & host-arch-x86_32:installed}
+   "--personality=x86_64-w64-mingw32" {os = "win32" & os-distribution = "cygwin" & host-arch-x86_64:installed}
+   "libcurl" {os = "win32" & os-distribution = "cygwin"}]
```
mtelvers, line 10: This logic is included in the `conf-mingw-w64-curl-i686` and `conf-mingw-w64-curl-x86_64`.  Why does it need to be added here as well?  Are you having an issue running `["sh" "-exc" "curl-config --libs"] {os = "win32" & os-distribution != "cygwinports"}`

yosefAlsuhaibani: Yes: for some reason on a github action running on a windows machine, `setup-ocaml` fails that step because it can't find `curl-config`.

mtelvers: Can you point me to a log a full log output?

The current version passes the tests with the GitHub runner.  See https://github.com/ocaml/opam-repository/actions/runs/11560280665/job/32176785255

yosefAlsuhaibani: Hi @mtelvers: unfortunately the actions run is in a private repo however

I was able to reproduce in a small public repo here: https://github.com/yosefAlsuhaibani/test-ocaml-setup-windows

See the PR: https://github.com/yosefAlsuhaibani/test-ocaml-setup-windows/pull/1

More specifically, the run: https://github.com/yosefAlsuhaibani/test-ocaml-setup-windows/actions/runs/11902381139/job/33167272712?pr=1

Done by the workflow: https://github.com/yosefAlsuhaibani/test-ocaml-setup-windows/actions/runs/11902381139/workflow?pr=1

It fails with two packages

1 - Libcurl: that I was able to solve with the fix above
2 - ocamlfind: which I still am struggling to fix

mtelvers: Do you get a failure when you run it manually, outside of `setup-ocaml`?

I have forked your repo and used OCaml-CI to test.  Both `conf-libcurl` and `ocamlfind` compile.  `ocurl` is the next failure point.  See https://ocaml.ci.dev/github/mtelvers/test-ocaml-setup-windows/commit/4f4912a295e9c08ae5489c651fa004f6e05bb740/variant/windows-server-2022-amd64-4.14_opam-2.3.

It is very interesting to get different results from the two CI approaches.

yosefAlsuhaibani: Wild! 

I don't have a windows machine, so I can't really build this manually; I've only tested this in CI.
 
Thank you for looking into this with me, that updates I have: currently, running the step that installs dependancies via opam without `shell:bash` in windows makes `ocamlfind` compile normally; I assume there is a bad interaction with `ocamlfind`'s installation process + bash + windows. I still keep getting CI failures with conf-libcurl however ☹️

yosefAlsuhaibani: Oh interesting: I now am running into `ocurl` after pinning to my fork 😭

mtelvers: I have tested `conf-libcurl.2` on an Windows machines and I was able to install it with `opam install conf-libcurl`.  opam was setup accepting all the defaults.  I wonder if this issue is related to https://github.com/ocaml/setup-ocaml/pull/859

yosefAlsuhaibani: Hi: yeah I also thought it might be `setup-ocaml` not setting PATH variables right, as I was looking into an unrelated issue where if you use `setup-ocaml@v3` with `musl/static` compiler option, some libraries can't find C header files installed.

I retried the workflow with replacing `setup-ocaml@v3` to `tobil4sk/setup-ocaml@269db9639ccdfb945844b19e3f7e4b65dd79cc8b` (i.e with the proposed fix), and still I'm facing the same issue.

yosefAlsuhaibani: I'll mark this PR closed, as I do think its an issue with setup-ocaml; I'll look into modifying/filing an issue in setup-ocaml and try to get it to work. Thank you for your time!

PR #26909, opened 2024-11-18
Release of OCaml 5.2.1

Octachron:
This PR adds the base package for the release of OCaml 5.2.1. 

As a notable changescompared to previous releases is that I have switched the reference url to an explicit binary archive rather than using the archive generated on demand by Github in order to avoid depending on the `Github` archive format.

Except for this change, we have the usual 3 base packages:

- ocaml-system.5.2.1
- ocaml-base-compiler.5.2.1
- ocaml-variants.5.2.1+options

The PR also switches the active trunk package from

- ocaml-variants.5.2.1+trunk

to
- ocaml-variants.5.2.2+trunk

which requires to add the `ocaml` 5.2.2 metapackage

- ocaml.5.2.2

The `ocaml-src` is not included in this release, since it is dependendent on a PR to the `opam-source-repository`.

### DIFF
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.5.2.1/opam
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.5.2.1/opam
@@ -1,0 +1,88 @@
+opam-version: "2.0"
+synopsis: "Official release 5.2.1"
+maintainer: [
+  "David Allsopp <david@tarides.com>"
+  "Florian Angeletti <florian.angeletti@inria.fr>"
+]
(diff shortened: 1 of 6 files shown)


mseri:
Thanks

shym:
> The `ocaml-src` is not included in this release, since it is dependendent on a PR to the `opam-source-repository`.

Is there such a PR already cooking somewhere?
I was thinking that, with the new organisation, it would make more sense to add a generic `META.in` to the source-archives repository, so that the `ocaml-src` package could be released with the other compiler packages with no extra dependency.

PR #26919, opened 2024-11-20
mark old releases of mpp and glical as unavailable.

hannesm:
the reasoning is that the source tarballs are unavailable.

discovered via https://opam.robur.coop/status -- also manually tried to download these archives without success.


mseri:
Ping @pw374

Thanks!

pw374:
> Ping @pw374
> 
> Thanks!

Thank you! It's now back online...

PR #26921, opened 2024-11-20
some opam packages pointed to tarballs where GitHub replied with "multiple choices"

hannesm:
thanks that we have the checksum, I figured alternative download urls and also added the sha256 checksum to these tarballs

powered by https://opam.robur.coop/status where md5 is reported as d41d8cd98f00b204e9800998ecf8427e


hannesm:
please note that in this PR I don't plan to fix any builds or reverse dependencies, this is only a revival of those packages that are otherwise uninstallable by clients.

mseri:
Thank you!

PR #26928, opened 2024-11-22
[new release] elpi (2.0.0)

gares:

##### CHANGES:

- Compiler:
  - Change the pipeline completely to make unit relocation unnecessary. Current
    phases are (roughly):
    1. `Ast.program` —[`RecoverStructure`]—> `Ast.Structured.program`
    2. `Ast.Structured.program` —[`Scope`,`Quotation`,`Macro`]—> `Scoped.program` (aka `API.Compile.scoped_program`)
    3. `Scoped.program` —[`Flatten`]—> `Flat.program`
    4. `Flat.program` —[`Check`]—> `CheckedFlat.program` (aka `API.Compile.compilation_unit`)
    5. `CheckedFlat.program` —[`Spill`,`ToDbl`]—> `Assembled.program`

    Steps 4 and 5 operate on a base, that is an `Assembled.program` being
    extended. `ToDbl` is in charge of allocating constants (numbers) for global
    names and takes place when the unit is assembled on the base. These
    constants don't need to be relocated as in the previous backend that
    would allocate these constants much earlier.
  - Change compilation units can declare new builtin predicates
  - Fix macros are hygienic
  - New type checker written in OCaml. The new type checker is faster,
    reports error messages with a precise location and performs checking
    incrementally when the API for separate compilation is used.
    The new type checker is a bit less permissive since the old one would
    merged together all types declaration before type checking the entire
    program, while the new one type checks each unit using the types declared
    inside the unit or declared in the base it extends, but not the types
    declared in units that (will) follow it.
  - Remove the need of `typeabbrv string (ctype "string")` and similar
  - New type check types and kinds (used to be ignored).

- API:
  - Change quotations generate `Ast.Term.t` and not `RawData.t`. The data
    type `Ast.Term.t` contains locations (for locating type errors) and
    has named (bound) variables and type annotations in `Ast.Type.t`.
  - New `Compile.extend_signature` and `Compile.signature` to extend a
    program with the signature (the types, not the code) of a unit
  - New `Ast.Loc.t` carries a opaque payload defined by the host application
  - Remove `Query`, only `RawQuery` is available (or `Compile.query`)

- Parser:
  - Remove legacy parser
  - New `% elpi:if version op A.B.C` and `% elpi:endif` lexing directives
  - New warning for `A => B, C` to be disabled by putting parentheses
    around `A => B`.

- Language:
  - New infix `==>` standing for application but with "the right precedence™",
    i.e. `A ==> B, C` means `A => (B, C)`.
  - New `pred` is allowed in anonymous predicates, eg:
    `pred map i:list A, i:(pred i:A, o:B), o:list B` declares that the first
    argument of `map` is a predicate whose first argument is in input and
    the second in output. Currently the mode checker is still in development,
    annotations for higher order arguments are ignored.
  - New attribute `:functional` can be passed to predicates (but not types).
    For example, `:functional pred map i:list A, i:(:functional pred i:A, o:B), o:list B`
    declares `map` to be a functional predicate iff its higher order argument is
    functional. Currently the determinacy checker is still in development, these
    annotations are ignored.
  - New `func` keyword standing for `:functional pred`. The declaration above
    can be shortened to `func map i:list A, i:(func i:A, o:B), o:list B`.
  - New type annotations on variables quantified by `pi` as in `pi x : term \ ...`
  - New type casts on terms, as in `f (x : term)`
  - New attribute `:untyped` to skip the type checking of a rule.

- Stdlib:
  - New `std.list.init N E L` builds a list `L = [E, ..., E]` with length `N`
  - New `std.list.make N F L` builds the list `L = [F 0, F 1, ..., F (N-1)]`
  - New `triple` data type with constructor `triple` and projections `triple_1`...

- Builtins:
  - Remove `string_to_term`, `read`, `readterm`, `quote_syntax`

- REPL:
  - Remove `-no-tc`, `-legacy-parser`, `-legacy-parser-available`
  - New `-document-infix-syntax`

### DIFF
--- a/packages/elpi/elpi.2.0.0/opam
+++ b/packages/elpi/elpi.2.0.0/opam
@@ -14,7 +14,7 @@ build: [
 ]
 
 depends: [
-  "ocaml" {>= "4.08.0" }
+  "ocaml" {>= "4.13.0" }
   "stdlib-shims"
   "ppxlib" {>= "0.12.0" }
   "menhir" {>= "20211230" }
--- a/packages/elpi/elpi.2.0.0/opam
+++ b/packages/elpi/elpi.2.0.0/opam
@@ -1,0 +1,90 @@
+opam-version: "2.0"
+maintainer: "Enrico Tassi <enrico.tassi@inria.fr>"
+authors: [ "Claudio Sacerdoti Coen" "Enrico Tassi" ]
+license: "LGPL-2.1-or-later"
+homepage: "https://github.com/LPCIC/elpi"
+doc: "https://LPCIC.github.io/elpi/"
+dev-repo: "git+https://github.com/LPCIC/elpi.git"
+bug-reports: "https://github.com/LPCIC/elpi/issues"
+
+build: [
+  ["dune" "subst"] {dev}
+  ["dune" "build" "-p" name "-j" jobs]
+  [make "tests" "DUNE_OPTS=-p %{name}%" "SKIP=performance_HO" "SKIP+=performance_FO" "SKIP+=elpi_api_performance"] {with-test & os != "macos" & os-distribution != "alpine" & os-distribution != "freebsd"}
+]
+
+depends: [
+  "ocaml" {>= "4.08.0" }
+  "stdlib-shims"
+  "ppxlib" {>= "0.12.0" }
+  "menhir" {>= "20211230" }
+  "re" {>= "1.7.2"}
+  "ppx_deriving" {>= "4.3"}
+  "ANSITerminal" {with-test}
+  "cmdliner" {with-test}
+  "fileutils" {with-test}
+  "dune" {>= "2.8.0"}
+  "conf-time" {with-test}
+  "atdgen" {>= "2.10.0"}
+  "atdts" {>= "2.10.0"}
+  "odoc" {with-doc}
+]
+synopsis: "ELPI - Embeddable λProlog Interpreter"
+description: """
+ELPI implements a variant of λProlog enriched with Constraint Handling Rules,
+a programming language well suited to manipulate syntax trees with binders.
+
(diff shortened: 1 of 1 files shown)


packages/elpi/elpi.2.0.0/opam
```diff
@@ -0,0 +1,90 @@
...
+bug-reports: "https://github.com/LPCIC/elpi/issues"
+
+build: [
+  ["dune" "subst"] {dev}
+  ["dune" "build" "-p" name "-j" jobs]
+  [make "tests" "DUNE_OPTS=-p %{name}%" "SKIP=performance_HO" "SKIP+=performance_FO" "SKIP+=elpi_api_performance"] {with-test & os != "macos" & os-distribution != "alpine" & os-distribution != "freebsd"}
+]
+
+depends: [
+  "ocaml" {>= "4.08.0" }
```
raphael-proust, line 17:
```suggestion
  "ocaml" {>= "4.13.0" }
```

Because of the following error in the CI:  
`failed: Unbound value String.starts_with`

The Stdlib documentation indicates "Since 4.13" in the documentation of `String.starts_with`.

-----------------

Please update this upstream for future releases.

raphael-proust (commented):
There also seems to be some issues on some 32 bit architectures during the tests. Some call to `Array.make` fails:

```
# File "src/runtime/dune", line 14, characters 12-19:
# 14 | (test (name test_bl) (libraries elpi.runtime) (modules test_bl) (preprocess (pps ppx_deriving.std)))
#                  ^^^^^^^
# (cd _build/default/src/runtime && ./test_bl.exe)
# Fatal error: exception Invalid_argument("Array.make")
# make: *** [Makefile:85: tests] Error 1
```

You should look into it. Possibly make your package not available on the offending architectures?

PR #26925, opened 2024-11-21
[new release] cohttp (14 packages) (6.0.0)

avsm:
An OCaml library for HTTP clients and servers

- Project page: <a href="https://github.com/mirage/ocaml-cohttp">https://github.com/mirage/ocaml-cohttp</a>
- Documentation: <a href="https://mirage.github.io/ocaml-cohttp/">https://mirage.github.io/ocaml-cohttp/</a>

##### CHANGES:

- bump minimum dune version to 3.8 (@avsm)
- cohttp-eio: Use system authenticator in example.
- http, cohttp: remove the scheme field from requests. This means that
  [Request.uri] no longer returns the same URI as was to create the request
  with [Request.make] (@rgrinberg 1086)
- cohttp-eio: Remove unused `Client_intf` module (talex5 mirage/ocaml-cohttp#1081)
- cohttp-eio: Make server response type abstract and allow streaming in cohttp-eio (talex5 mirage/ocaml-cohttp#1024)
- cohttp-{lwt,eio}: server: add connection header to response if not present (ushitora-anqou mirage/ocaml-cohttp#1025)
- cohttp-curl: Curl no longer prepends the first HTTP request header to the output. (jonahbeckford mirage/ocaml-cohttp#1030, mirage/ocaml-cohttp#987)
- cohttp-eio: client: use permissive argument type for make_generic
- cohttp-eio: Improve error handling in example server (talex5 mirage/ocaml-cohttp#1023)
- cohttp-eio: Don't blow up `Server.callback` on client disconnections. (mefyl mirage/ocaml-cohttp#1015)
- http: Fix assertion in `Source.to_string_trim` when `pos <> 0` (mefyl mirage/ocaml-cohttp#1017)
- cohttp: `Cohttp.Request.make_for_client` no longer allows setting both
  `~chunked:true` and `~body_length`.
- cohttp-lwt-unix: Don't blow up when certificates are not available and no-network requests are made. (akuhlens mirage/ocaml-cohttp#1027)
  + Makes `cohttp-lwt.S.default_ctx` lazy.

PR #26930, opened 2024-11-22
graphql-cohttp: upper bound

avsm:
from revdeps for #26925


packages/graphql-cohttp/graphql-cohttp.0.10.0/opam
```diff
@@ -15,7 +15,7 @@ depends: [
   "ocaml" {>= "4.03.0"}
   "dune"
   "graphql" {>= "0.9.0"}
-  "cohttp" {>= "2.0.0"}
+  "cohttp" {>= "2.0.0" & < "6.0.0~~"}
```
raphael-proust, line 18: I'm unclear why you need the `~~` suffix with the version number… What's the reason behind it?

avsm: It's the "lowest possible", i.e. lower than 6.0.0~beta1

PR #26937, opened 2024-11-23
[new release] elpi (2.0.2)

gares:

##### CHANGES:

Requires Menhir 20211230 and OCaml 4.13 or above.

- Really fix tests on 32 bits systems

### DIFF
--- a/packages/elpi/elpi.2.0.2/opam
+++ b/packages/elpi/elpi.2.0.2/opam
@@ -1,0 +1,90 @@
+opam-version: "2.0"
+maintainer: "Enrico Tassi <enrico.tassi@inria.fr>"
+authors: [ "Claudio Sacerdoti Coen" "Enrico Tassi" ]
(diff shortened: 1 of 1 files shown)


gares:
I managed to solve the "thread create" issue that affected previous versions (it was a dune problem with a large stack).

I also managed to fix the test suite on 32 bit systems.

I have no clue what is wrong with the two failing jobs, but IMHO the package can be merged.

gares:
Btw, if you think the failures are important, I'm fine disabling the package on 32bit + 5.2, but I'm not sure I know opam's syntax well enough for doing so. Suggestions are welcome.

shonfeder:
Thanks for the updates, and for helping analyze the CI issues. These are test failures on those platforms, rather than build failures, so it is not as if the package cannot be installed or used there.

I think its fine to merge as is.

gares:
Thanks

PR #26943, opened 2024-11-24
Package checked_oint.0.4.0

hirrolot:
### `checked_oint.0.4.0`
An OCaml library for checked integer arithmetic

---
* Homepage: https://github.com/hirrolot/checked_oint
* Source repo: git+https://github.com/hirrolot/checked_oint.git
* Bug tracker: https://github.com/hirrolot/checked_oint/issues

---
:camel: Pull-request generated by opam-publish v2.3.0

PR #26944, opened 2024-11-25
Update the `checked_oint` & `mazeppa` checksums

hirrolot:
The following happened:
 1. I've rebased the `master` branch with `git-filter-repo` only to change commit metadata.
 2. GitHub re-created release archive files with slightly different compressed byte content but same content after decompression.
 3. Therefore, the MD5 and SHA-512 hashes of the source code archives became different.

Interestingly, GitHub [doesn't even guarantee] stability of source code archives, so this situation could have happened without a forceful push. Nonetheless, the easiest solution would be to just update the checksums according to `md5sum` and `sha512sum` for respective archives.

[doesn't even guarantee]: https://docs.github.com/en/repositories/working-with-files/using-files/downloading-source-code-archives#stability-of-source-code-archives

Needed for https://github.com/ocaml/opam-repository/pull/26943.


hannesm:
FWIW, for the future, topkg-care pioneered an approach to generate a reproducible tarball on release time which dune-release picked up. And GitHub allows you to upload a custom tarball on release time. So if you use e.g. dune-release for your releases you won't end up relying on GitHub.

For the present PR, I'm not a fan of changing the checksums, but instead finding the tarballs in one of the caches and uploading them somewhere. I'm not sure what the opam-repository policy in this respect is at the moment.

hirrolot:
> finding the tarballs in one of the caches and uploading them somewhere

Where should I locate them? I couldn't find any information about where GitHub stores previously generated source code tarballs, and I don't have them locally.

shonfeder:
> Where should I locate them? I couldn't find any information about where GitHub stores previously generated source code tarballs, and I don't have them locally.

Try having a look at the archives at https://github.com/ocaml/opam-repository/wiki/FAQ#how-to-find-lost-archives-of-packages . LMK if you need any support in hunting them down!

PR #26947, opened 2024-11-25
add ocaml-src 5.2.1

hannesm:
//cc @Octachron @shym @reynir

I tried to find the ocaml-src release instructions in the ocaml/ocaml repository, but couldn't.

The gist is, as discussed with @shym, that an empty META file is sufficient for `ocamlfind query` purposes. So, in this opam file we execute `touch META`.

### DIFF
--- a/packages/ocaml-src/ocaml-src.5.2.1/opam
+++ b/packages/ocaml-src/ocaml-src.5.2.1/opam
@@ -1,0 +1,15 @@
+opam-version: "2.0"
+maintainer: "opam-devel@lists.ocaml.org"
+license: "LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception"
+authors: "OCaml contributors"
+homepage: "http://ocaml.org/"
+build: [ "touch" "META" ]
+install: ["cp" "-r" "." "%{lib}%/ocaml-src"]
+synopsis: "Compiler sources"
+depends: [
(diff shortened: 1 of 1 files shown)


Octachron:
Thanks a lot for taking care of this package! If creating an empty META file is good enough for ocamlfind, this looks like a clear improvement to me.

shym:
What I tried to do is to look at all the packages that depend on `ocaml-src` and see whether, and how, they use the ocamlfind `META` file. I may have looked a bit too fast but the only usage I saw was indeed just to locate the directory, in which case an empty file is enough.

For the curious, quite a few packages come with empty `META` files, so this is probably often good enough. For instance, for `menhir` I get:
```bash
$ ocamlfind query -l menhir
package:     menhir
description: [n/a]
version:     [unspecified]
archive(s):
linkopts:
location:    $OPAM_SWITCH_PREFIX/lib/menhir
```

mseri:
Thanks

avsm:
This is a divergence from the other `ocaml-src` packages; should we just adopt the same empty META file convention for all of them to avoid potential heisenbugs in the future?

hannesm:
@avsm I opened https://github.com/ocaml/opam-repository/pull/26972 to synchronize the other ocaml-src packages with this change (empty META file).

hannesm:
anything needed from my site to merge now that there's a consider-for-merge label?

PR #26950, opened 2024-11-25
Cohttp6 upper bounds

avsm:
revdep failures for #26925


mseri:
The linter is fine, there is one lower bound issue that we can investigate separately

avsm:
Thanks for the review! I fixed the graphql lower bound and merged to make progress with cohttp 6.0.0

PR #26927, opened 2024-11-22
New version of package color: 0.3.0

ElectreAAS:
cc @art-w :heart:

PR #26954, opened 2024-11-25
Fix color package

mtelvers:
Followup to https://github.com/ocaml/opam-repository/pull/26927

### DIFF
--- a/packages/color/color.0.3.0/opam
+++ b/packages/color/color.0.3.0/opam
@@ -1,0 +1,43 @@
+opam-version: "2.0"
+synopsis: "Convert between different color formats"
+description:
+  "Library that converts between different color formats. Right now it deals with RGB[A], HSL[A], OkLab, and Oklch formats. All those format convert to and from [Gg.Color.t]"
+maintainer: [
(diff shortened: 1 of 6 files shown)


packages/color/color.0.3.0/opam
```diff
@@ -0,0 +1,43 @@
...
+depends: [
+  "dune" {>= "2.0"}
+  "ocaml" {>= "4.08.0"}
+  "gg" {>= "1.0.0"}
+  "alcotest" {with-test}
+  "ocaml-lsp-server" {with-dev-setup}
+  "ocamlformat" {= "0.26.2" & with-dev-setup}
+]
+build: [
+  ["dune" "subst"] {pinned}
```
shonfeder, line 22:
```suggestion
  ["dune" "subst"] {dev}
```

ElectreAAS:
For the record here: the problem with #26927 was that line `url: {` instead of `url {`, is that correct?

shonfeder:
That's a funky quirk of the opam format.

mseri:
Thanks!

PR #26958, opened 2024-11-26
Update the archive URLs for `checked_oint` & `mazeppa`

hirrolot:
See https://github.com/ocaml/opam-repository/pull/26944.

For the curious, the new and old archives were different byte-wise because GitHub inserts the commit's SHA-1 into a custom `.tar` metadata field called `comment`:

```bash
$ gzip -d mazeppa-0.4.3.tar.gz
$ gzip -d mazeppa-0.4.3-old.tar.gz
$ md5sum mazeppa-0.4.3.tar
da53390343565823e32ce43dfb2d805b  mazeppa-0.4.3.tar
$ md5sum mazeppa-0.4.3-old.tar
8c989a48fc858f7441e3cd4e33332142  mazeppa-0.4.3-old.tar
$ head -c 2048 mazeppa-0.4.3.tar | grep -ao "comment=[^[:space:]]*"
comment=0c4e7b577d3607433340a46344427bac8780c0e1
$ head -c 2048 mazeppa-0.4.3-old.tar | grep -ao "comment=[^[:space:]]*"
comment=b3d044040f82d4611c475930da889e4e804ed956
```

If we compare the decompressed folders, they will be the same content-wise:

```bash
$ mkdir new old
$ tar -xf mazeppa-0.4.3.tar -C new
$ tar -xf mazeppa-0.4.3-old.tar -C old
$ diff -r new old
# Empty output.
```

In the future, I'll attach custom GitHub release archives to make sure they don't change under any circumstances. The current PR implements it for the existing versions of `checked_oint` and `mazeppa`.

PR #26966, opened 2024-11-26
Update `checked_oint` v0.4.0 to a stable archive URL

hirrolot:
In https://github.com/ocaml/opam-repository/pull/26943, we forgot to update the package URL to a stable archive (not dynamically generated), in accordance with the approach implemented in https://github.com/ocaml/opam-repository/pull/26958.


mseri:
Thanks

PR #26986, opened 2024-11-28
Add `ocaml-manual.5.2.1`

shym:
The manual didn’t change between 5.2.0 and 5.2.1 so the same source archive can be used.


shym:
As Windows CI failed (though I don’t understand in the log what failed), I simplified the command to use `cp` also on Windows, as it is also what is used in `ocaml-src`. (IIUC on Windows, Cygwin `cp` will be used in that case)

mseri:
Thanks

Issue #20793, opened 2022-02-24
Non `conf-*` packages with depexts


rjbou:
There is 330 packages (last version) that declares `depexts`, and 118 of them are not conf package. It comes from legacy. If we take all version of packages, there is 1123 depexts fields in the repo, and 842 of them no conf package.

To get the detailled list, it includes some of packages that already externalised their depexts:
```
opam list -a -s --field=depexts | grep -v conf- | xargs opam show -f name,version,depexts:
```
Current ones:
* [ ] 0install
* [ ] aacplus
* [ ] aio
* [ ] albatross
* [ ] arakoon
* [ ] assimp
* [ ] augeas
* [ ] bap-llvm
* [ ] bap-std
* [ ] bwrap
* [ ] ca-certs
* [ ] camldm
* [ ] camlhighlight
* [ ] cmark
* [ ] cmarker
* [ ] core
* [ ] core_unix
* [ ] cwe_checker
* [ ] dlm
* [ ] dose
* [ ] extunix
* [ ] faad
* [ ] farmhash
* [ ] fdkaac
* [ ] flowtype
* [ ] freetds
* [ ] freetennis
* [ ] fstar
* [ ] gammu
* [ ] gavl
* [ ] geoip
* [ ] gettext-stub
* [ ] gles3
* [ ] glMLite
* [ ] glpk
* [ ] goblint-cil
* [ ] gperftools
* [ ] hardcaml-vpi
* [ ] hdf
* [ ] hdf5
* [ ] hvsock
* [ ] imagemagick
* [ ] irrlicht
* [ ] jemalloc
* [ ] kafka
* [ ] krb5
* [ ] kremlin
* [ ] kyotocabinet
* [ ] lablgl
* [ ] lablgtk3-gtkspell3
* [ ] labrys
* [ ] lemonade-sqlite
* [ ] libbinaryen
* [ ] libssh
* [ ] libvirt
* [ ] liquidsoap
* [ ] lmdb
* [ ] lzo
* [ ] magic
* [ ] milter
* [ ] misuja
* [ ] mlmpfr
* [ ] mm
* [ ] not-ocamlfind
* [ ] ocaml-buddy
* [ ] ocaml-expat
* [ ] ocaml-option-32bit
* [ ] ocaml-option-musl
* [ ] ocaml-option-static
* [ ] ocamlsdl
* [ ] ocaml-systemd
* [ ] ocaml-zmq
* [ ] ocsigen-start
* [ ] odbc
[...]

PR #26991, opened 2024-11-28
Add new conf-gcc-multilib package

sim642:
goblint-cil.2.0.5 does some optional compilation based on whether `cpp -m32` and `cpp -m64` work, so it might make sense for it to have an optional dependency on such conf-gcc-multilib package.
ocaml-option-32bit.1 has an explicit depext for gcc-multilib (and also g++-multilib). In the spirit of #20793 it might also make sense for the depext to be delegated to a conf-* package.

So this is my attempt at making such package, although I'm not particularly confident about it. Most non-debian-based distributions don't seem to offer such a package (or I couldn't easily find one).
Also I'm not too sure about the build checks being general/flexible/necessary.

Thus, this is very open to feedback.

**EDIT:** I did some digging to find packages for more distros, although it made me realize that perhaps gcc-multilib itself is the odd one out. It might be 32bit glibc that I've really aimed at here.

### DIFF
--- a/packages/conf-gcc-multilib/conf-gcc-multilib.1.0/opam
+++ b/packages/conf-gcc-multilib/conf-gcc-multilib.1.0/opam
@@ -1,0 +1,18 @@
+opam-version: "2.0"
+maintainer: "Simmo Saan <simmo.saan@gmail.com>"
+authors: ["Simmo Saan"]
+homepage: "https://github.com/ocaml/opam-repository"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: "GPL-2.0-or-later"
+build: [
+  ["sh" "-exc" "echo '#include <limits.h>' | cpp -m32 -"]
+  ["sh" "-exc" "echo '#include <limits.h>' | cpp -m64 -"]
+]
+depexts: [
+  ["gcc-multilib"] {os-distribution = "debian"}
+  ["gcc-multilib"] {os-distribution = "ubuntu"}
+]
+synopsis: "Virtual package relying on the gcc compiler multilib support"
+description:
+  "This package can only install if the gcc compiler multilib support is installed on the system."
+flags: conf
--- a/packages/conf-gcc-multilib/conf-gcc-multilib.1.0/opam
+++ b/packages/conf-gcc-multilib/conf-gcc-multilib.1.0/opam
@@ -9,9 +9,14 @@ build: [
   ["sh" "-exc" "echo '#include <limits.h>' | cpp -m64 -"]
 ]
 depexts: [
+  ["glibc-devel.i686"] {os-distribution = "centos"}
+  ["glibc-devel.i686"] {os-distribution = "fedora"}
+  ["gcc-32bit"] {os-family = "suse" | os-distribution = "opensuse"}
   ["gcc-multilib"] {os-distribution = "debian"}
   ["gcc-multilib"] {os-distribution = "ubuntu"}
+  ["lib32-glibc"] {os-distribution = "archlinux"}
 ]
+available: arch != "s390x"
 synopsis: "Virtual package relying on the gcc compiler multilib support"
 description:
   "This package can only install if the gcc compiler multilib support is installed on the system."
--- a/packages/conf-gcc-multilib/conf-gcc-multilib.1.0/opam
+++ b/packages/conf-gcc-multilib/conf-gcc-multilib.1.0/opam
@@ -11,10 +11,10 @@ build: [
 depexts: [
   ["glibc-devel.i686"] {os-distribution = "centos"}
   ["glibc-devel.i686"] {os-distribution = "fedora"}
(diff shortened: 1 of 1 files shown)


shonfeder:
Sorry for the delay here. I'm afraid this package slipped between the cracks over the holidays. Do you have any thoughts on the CI failures we see here? https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/1d9aa3a77d42360c809fa99e0d6c021fc2db3644

packages/conf-gcc-multilib/conf-gcc-multilib.1.0/opam
```diff
@@ -0,0 +1,23 @@
...
+  ["sh" "-exc" "echo '#include <limits.h>' | cpp -m32 -"]
+  ["sh" "-exc" "echo '#include <limits.h>' | cpp -m64 -"]
+]
+depexts: [
+  ["glibc-devel.i686"] {os-distribution = "centos"}
+  ["glibc-devel.i686"] {os-distribution = "fedora"}
+  ["gcc-32bit"] {os-family = "suse" | os-distribution = "opensuse"}
+  ["gcc-multilib"] {os-distribution = "debian"}
+  ["gcc-multilib"] {os-distribution = "ubuntu"}
+  ["lib32-glibc"] {os-distribution = "archlinux"}
```
raphael-proust, line 17:
```suggestion
  ["lib32-glibc"] {os-family = "arch"}
```

See https://github.com/ocaml/opam-repository/wiki/Depexts-os-distribution---os-family-values

Also, orthogonally, I wonder if this should be `lib32-gcc-libs`

packages/conf-gcc-multilib/conf-gcc-multilib.1.0/opam
```diff
@@ -0,0 +1,23 @@
...
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: "GPL-2.0-or-later"
+build: [
+  ["sh" "-exc" "echo '#include <limits.h>' | cpp -m32 -"]
+  ["sh" "-exc" "echo '#include <limits.h>' | cpp -m64 -"]
+]
+depexts: [
+  ["glibc-devel.i686"] {os-distribution = "centos"}
+  ["glibc-devel.i686"] {os-distribution = "fedora"}
+  ["gcc-32bit"] {os-family = "suse" | os-distribution = "opensuse"}
```
raphael-proust, line 14:
```suggestion
  ["gcc-32bit"] {os-family = "suse" | os-family = "opensuse"}
```

raphael-proust (commented):
pretty sure these two changes will fix arch and suse, idk about fedora because it seems to be an error during the installation of the package

i'd recommend we remove the fedora line for now, someone with a fedora box can contribute the correct  names later possibly

raphael-proust:
@sim642 do you think the arch package should be instead https://archlinux.org/packages/core/x86_64/lib32-gcc-libs/ ?

packages/conf-gcc-multilib/conf-gcc-multilib.1.0/opam
```diff
@@ -0,0 +1,23 @@
...
+]
+depexts: [
+  ["glibc-devel.i686"] {os-distribution = "centos"}
+  ["glibc-devel.i686"] {os-distribution = "fedora"}
+  ["gcc-32bit"] {os-family = "suse" | os-family = "opensuse"}
+  ["gcc-multilib"] {os-distribution = "debian"}
+  ["gcc-multilib"] {os-distribution = "ubuntu"}
+  ["lib32-glibc"] {os-family = "arch"}
+]
+available: arch != "s390x"
```
mtelvers: There is no RISC-V version of this package either.
```suggestion
available: arch != "s390x" & arch != "riscv64"
```

sim642:
> @sim642 do you think the arch package should be instead https://archlinux.org/packages/core/x86_64/lib32-gcc-libs/ ?

Maybe it should be both lib32-glibc and lib32-gcc-libs? Arch seems to have a multilib-devel metapackage which might be the most appropriate thing. It depends on the latter. But lib32-gcc-libs also depends on lib32-glibc.

Ubuntu's gcc-multilib is also a metapackage which transitively depends on a lot of things. 

> i'd recommend we remove the fedora line for now, someone with a fedora box can contribute the correct names later possibly

Something strange seems to be happening there. Even in the latest CI, it successfully installs the package and then for some reason looks it up again and fails:

```
+ /usr/sbin/sudo "yum" "install" "-y" "glibc-devel.i686"
[...]
- Installing:
-  glibc-devel        i686   0:2.43-8.fc44   updates                            2.3 MiB
[...]
- Complete!
+ /usr/sbin/rpm "-q" "--whatprovides" "glibc-devel.i686"
- no package provides glibc-devel.i686
[ERROR] System package install failed with exit code 1 at command:
            rpm -q --whatprovides glibc-devel.i686
[ERROR] These packages are still missing: glibc-devel.i686
```

PR #27000, opened 2024-11-29
avoid dependency to '"ocaml" {= "0"}'

hannesm:
This has been used for marking packages unavailable.


hannesm:
note that the changed mirage-net-xen packages are already marked as "available: false".

mseri:
Thanks

PR #27003, opened 2024-11-30
Add upper-bounds on capnp-rpc-{net,unix}

talex5:
The new 2.0 API will be incompatible.

Some old versions of `current_rpc` used `implicit_transitive_deps`, so they need bounds even with `capnp-rpc-lwt`.

I split this from #27002 to make it easier to see where the real errors are. Since this is just adding upper bounds that don't affect anything at present, any errors must be pre-existing problems.

I'll update #27002 once this is merged.


mseri:
Some lower bounds need fixing but we can take care of those few ones separately

PR #27010, opened 2024-12-02
[new release] cmdlang (5 packages) (0.0.9)

mbarbin:
I took the opportunity to improve the generation of opam files for the repo by using template files, so as to add tags, description and x-maintenance-intent fields.

### Added

- Added an example of migration from `core.command` to `climate`.
- Added migration utils.
- Improve code coverage, added tests.

### Changed

- Document presence in stdlib-runner help (required, default, etc.).
- Minor refactor in stdlib-runner.
- Upgrade to `climate.0.3.0`.

### Fixed

- Fix trailing dot additions in `to-cmdliner` for cases such as `?.` and `..`.

### Removed

- Removed config option `auto_add_short_aliases` from to-base translation (not useful).

### DIFF
--- a/packages/cmdlang-stdlib-runner/cmdlang-stdlib-runner.0.0.9/opam
+++ b/packages/cmdlang-stdlib-runner/cmdlang-stdlib-runner.0.0.9/opam
@@ -1,0 +1,53 @@
+opam-version: "2.0"
+synopsis: "A basic execution runner for cmdlang based on stdlib.arg"
+maintainer: ["Mathieu Barbin <opensource@mbarbin.org>"]
+authors: ["Mathieu Barbin"]
+license: "MIT"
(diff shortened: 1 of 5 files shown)


mbarbin:
I have looked at the CI error (1 job). Somewhere in the log shows:

```sh
 - downgrade ppxlib                      0.33.0 to 0.32.1   [uses dune]
```

The (new) package `cmdlang-to-base` requires `ppxlib.0.33` but I failed to include that constraint in the opam file for this package.

This cause an error later due to earlier versions of ppxlib not supporting `-unused-code-warnings=force`.

Question: I should add the missing constraint (will look into pushing this next). However, given that this job has a downgrade step for ppxlib, is this going to work? Is this job required to pass validation?

Thanks!

avsm:

PR #27014, opened 2024-12-02
Update pkg provider

mbarbin:
Latest version (0.0.11):

- Add tags & description.
- Add [x-maintenance-intent] per packaging archiving call to action.

Older version (0.0.9):

- Emphasis desire to archive (mark as deprecated).


avsm:
Thanks!

PR #27020, opened 2024-12-03
Kakadu's x-maintanenece-intents (ready for review)

Kakadu:


Kakadu:
> Error in conf-qt.5.2.1: error 46: Package is flagged "conf" but has source, install or remove instructions

Can we close our eyes about this?

mseri:
Thanks a lot!

> > Error in conf-qt.5.2.1: error 46: Package is flagged "conf" but has source, install or remove instructions
> Can we close our eyes about this?

Yes, that is not a problem

PR #27024, opened 2024-12-04
add ocaml-solo5 0.8.5

hannesm:

PR #22353, opened 2022-10-24
[new release] http, cohttp, cohttp-top, cohttp-server-lwt-unix, cohttp-mirage, cohttp-lwt, cohttp-lwt-unix, cohttp-lwt-jsoo, cohttp-eio, cohttp-curl, cohttp-curl-lwt, cohttp-curl-async and cohttp-async (6.0.0~alpha0)

mseri:
Cohttp is an OCaml library for creating HTTP daemons. It has a portable HTTP parser, and implementations using various asynchronous programming libraries:

- Http provides essential type definitions used in Cohttp and an extremely fast http parser. It is designed to have no dependencies and make it easy for other packages to easily interoperate with Cohttp.
- Cohttp_lwt_unix uses the [Lwt](https://ocsigen.org/lwt/) library, and specifically the UNIX bindings. It uses [ocaml-tls](https://github.com/mirleft/ocaml-tls) as the TLS implementation to handle HTTPS connections.
- Cohttp_async uses the [Async](https://realworldocaml.org/v1/en/html/concurrent-programming-with-async.html) library and async_ssl to handle HTTPS connections.
- Cohttp_lwt exposes an OS-independent Lwt interface, which is used by the [Mirage](https://mirage.io/) interface to generate standalone microkernels (use the cohttp-mirage subpackage).
- Cohttp_lwt_jsoo compiles to a JavaScript module that maps the Cohttp calls to XMLHTTPRequests. This is used to compile OCaml libraries like the GitHub bindings to JavaScript and still run efficiently.
- Cohttp_curl uses libcurl (via ocurl), and also provides an lwt and an async backend.
- Cohttp_eio uses eio to leverage new features from multicore ocaml.

Links:
- Project page: <a href="https://github.com/mirage/ocaml-cohttp">https://github.com/mirage/ocaml-cohttp</a>
- Documentation for the alpha version needs to be generated locally with `dune build @doc`

##### CHANGES:

- cohttp-eio: ensure "Host" header is the first header in http client requests (bikallem mirage/ocaml-cohttp#939)
- cohttp-eio: add TE header in client. Check TE header is server (bikallem mirage/ocaml-cohttp#941)
- cohttp-eio: add User-Agent header to request from Client (bikallem mirage/ocaml-cohttp#940)
- cohttp-eio: add Content-Length header to request/response (bikallem mirage/ocaml-cohttp#929)
- cohttp-eio: add cohttp-eio client api - Cohttp_eio.Client (bikallem mirage/ocaml-cohttp#879)
[...]

PR #27031, opened 2024-12-05
mark solo5 packages that use -Werror as avoid-version

hannesm:
since solo5 0.8.1, there's no more `-Werror` for releases


hannesm:
my intention is that other CI checks, such as #27024, won't pick the avoid-version versions in their dependencies, esp. the lower-bounds runs.

mseri:
I thought that we need to use `available: opam-version >= "2.1.0" ` when we use `avoid-version`. What happens if a package has `avoid-version` and an older opam is used? If nothing, then I think we can merge, if instead it fails, then I think it may be worth discussing it

hannesm:
My assumption is that using opam < 2.1 with this ignores `avoid-version`.

mseri:
Indeed, that is the case: https://github.com/ocaml/opam-repository/pull/22353#issuecomment-1289184915

mseri:
Thanks

PR #27040, opened 2024-12-06
Add: jsont.0.1.1

dbuenzli:
* Add: `jsont.0.1.1` [home](https://erratique.ch/software/jsont), [doc](https://erratique.ch/software/jsont/doc), [issues](https://github.com/dbuenzli/jsont/issues)  
  *Declarative JSON data manipulation for OCaml*

---

#### `jsont` v0.1.1 2024-12-06 La Forclaz (VS)

- `Jsont.Object.Mems.map` make encoding and decoding optional. Like
   in overy other map.
- `Jsont.Array.map` make encoding and decoding optional. Like
   in every other map.
- `Jsont_bytesrw.encode` change the default buffer size
  to match the one hinted by the writer rather than
  `Bytesrw.Bytes.Slice.io_buffer_size`.
- `jsont.{bytesrw,brr}` export all requires.
- `jsont` tool remove spurious dependency on `b0.std` ([#2](https://github.com/dbuenzli/jsont/issues/2)).

---

Use `b0 -- .opam publish jsont.0.1.1` to update the pull request.

### DIFF
--- a/packages/jsont/jsont.0.1.1/opam
+++ b/packages/jsont/jsont.0.1.1/opam
@@ -1,0 +1,62 @@
+opam-version: "2.0"
+synopsis: "Declarative JSON data manipulation for OCaml"
+description: """\
+Jsont is an OCaml library for declarative JSON data manipulation. It
+provides:
+
+- Combinators for describing JSON data using the OCaml values of your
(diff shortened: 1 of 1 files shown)


avsm:

PR #27041, opened 2024-12-06
Release dot-merlin-reader 4.18-414

voodoos:
merlin `4.18-414` and merlin-lib `4.18-414` were not co-installabe.

This should fix that issue by releasing a new version of dor-merlin-reader matching `4.18-414`.

We will switch to always release identical versions (with 
 {= version} constraints) for all merlin related packages in the future to prevent such issues.


voodoos:
The failing tests are unrelated to this release: one test in Merlin changes behavior with Dune 3.17.

PR #27047, opened 2024-12-08
Support conf-libpcre2-8 on mingw

tobil4sk:
Modelled after #26893

### DIFF
--- a/packages/conf-libpcre2-8/conf-libpcre2-8.2/opam
+++ b/packages/conf-libpcre2-8/conf-libpcre2-8.2/opam
@@ -1,0 +1,42 @@
+opam-version: "2.0"
+maintainer: "https://github.com/ocaml/opam-repository/issues"
+authors: [
+  "Philip Hazel"
+  "Zoltan Herczeg"
+]
+homepage: "https://www.pcre.org/"
+dev-repo: "git+https://github.com/PCRE2Project/pcre2.git"
+license: "BSD-3-Clause"
+build: [
+  ["pkgconf" {os = "win32" & os-distribution = "cygwin"}
+   "--personality=i686-w64-mingw32" {os = "win32" & os-distribution = "cygwin" & host-arch-x86_32:installed}
+   "--personality=x86_64-w64-mingw32" {os = "win32" & os-distribution = "cygwin" & host-arch-x86_64:installed}
+   "pkg-config" {os != "win32" | os-distribution != "cygwin"}
+   "--print-errors" "--exists" "libpcre2-8"
+  ]]
+depends: [
+  "conf-pkg-config" {build}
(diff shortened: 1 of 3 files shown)


packages/conf-libpcre2-8/conf-libpcre2-8.2/opam
```diff
@@ -0,0 +1,42 @@
+opam-version: "2.0"
+maintainer: "https://github.com/ocaml/opam-repository/issues"
```
shonfeder, line 2: AFAIU, maintainer should name particular people (or organizations) who have
taken on the responsibility for maintaining a package. I don't think a link to
the opam repo issue tracker serves this end, and in any case it duplicates the
data we already have in the `bug-reports` field.

Would you be willing to maintain this package?

packages/conf-libpcre2-8/conf-libpcre2-8.2/opam
```diff
@@ -0,0 +1,42 @@
+opam-version: "2.0"
```
shonfeder, line 1: What does the `-8` mean in the package name?

shonfeder (changes requested):
Thank you for providing this system package! With the caveat that I don't have a lot of experience reviewing `conf-` packages yet, I don have one request and one question about the main opam file.

packages/conf-libpcre2-8/conf-libpcre2-8.2/opam
tobil4sk: Hi, for `conf-libpcre2-8`, I am just updating the dependencies and the `pkg-config` check however there is already a previous version in the repository: `conf-libpcre2-8.1`. This previous version also put the issue tracker in place of the maintainer field. I'm not sure if I have the knowledge to maintain this package for every platform.

I feel more confident putting myself as maintainer for the newly added `conf-mingw-w64-pcre2-i686` and `conf-mingw-w64-pcre2-x86_64`.

packages/conf-libpcre2-8/conf-libpcre2-8.2/opam
tobil4sk: There is an explanation for the library name in: #21349

packages/conf-libpcre2-8/conf-libpcre2-8.2/opam
shonfeder: Oh, sorry I missed that this was just update! My comment is not relevant then.

packages/conf-libpcre2-8/conf-libpcre2-8.2/opam
shonfeder: Got it, thanks! Sorry again that I missed the context of the previous release.

shonfeder:
Only a single revdep failure from `pa_ppx_regexp` 0.02. I've confirmed (locally) that this is already failing to install with tests prior to this release.

Thanks again for your work packaging here!

shonfeder (approved):

PR #27053, opened 2024-12-08
5 packages from aryx/ocaml-elm-playground at 0.1.6

aryx:
This pull-request concerns:
- `elm_core.0.1.6`: Imitate core Elm functions for ease of porting
- `elm_playground.0.1.6`: Interface of the Playground
- `elm_playground_native.0.1.6`: Native TDSL-based playground platform
- `elm_playground_web.0.1.6`: Web vdom-based playground platform
- `elm_system.0.1.6`: Imitate system Elm functions for ease of porting

---
* Homepage: https://github.com/aryx/ocaml-elm-playground
* Source repo: git+https://github.com/aryx/ocaml-elm-playground.git
* Bug tracker: https://github.com/aryx/ocaml-elm-playground/issues

---


aryx:
I've put in the dune file:
```
(library
  (public_name elm_playground_native)
  (implements elm_playground)
  ; problems with tsdl on windows I think
  (enabled_if
  (<> %{os_type} Win32))
  (libraries
    str
    logs
    cairo2 tsdl
    imagelib.unix curl
    
    elm_playground
  )
)

```
but the enabled_if does not seem to work; Opam-ci still tries to build the library on windows.

aryx:
hmm it's because the error is at the opam installation step.
How can I disable windows in the dune-project file (I'm using `(generate_opam_files true)` so I need
the directive in the dune-project).

shonfeder:
Hi, @aryx! Thanks you for the care with getting this package working on CI :pray: 

> How can I disable windows in the dune-project file

You can use opam's [`available`](https://opam.ocaml.org/doc/Manual.html#opamfield-available) field to indicate which platforms a package is available on.

The [dune `package` stanza](https://dune.readthedocs.io/en/stable/reference/dune-project/package.html) doesn't expose this (I'm not sure why), but you can add it in via a `<pgk>.opam.template` files. Here is an example of what it should end up looking like in an opam file:

https://github.com/shonfeder/opam-repository/blob/master/packages/ascend/ascend.0.1.2/opam#L32

```
available: [ os-family != "windows" ]
```

And here's an example of a `opam.template` file used for just this purpose.

For the publication of this package, we just add this field to your opam file directly, and you can update your package data upstream.

But you may want to remove the unneeded complication of the `enabled_if` in your dune file for your release, unless you plan to have conditional builds for different platforms

LMK if you need any clarification on these points of have more questions!

aryx:
Ok let me try what you proposed in 0.1.7.

Issue #23697, opened 2023-04-21
Policy for non-functional packages


emillon:
This is the summary of a discussion between @kit-ty-kate and I regarding how non-functional packages should be handled in the repository, more precisely in the context of #23649.

The current policy for opam-repository is to only add new packages, and not remove any. However, the maintainers fix package metadata so that installing a package should not result in a build failure.

However, there are some circumstances where there are benefits to making some releases unavailable. For example:
- when a package release has a known security issue
- when a package relies on an external service that does not exist anymore, or through an API that does not exist anymore

In the latter case, the package can be installed, but is not useful at runtime.

In addition, these non-functional packages cost a non-negligible amount of resources in CI: because they are included in the revdeps run of some "core" packages (`dune`, `lwt`,`uri`, the compilers...), they consume time and energy.

The goal of this issue is to list the possible solutions and discuss what should be implemented.

(In addition, having a growing number of packages has some negative consequences in itself: it is more work for the opam maintainers, it takes up disk space (and inodes) and slows down `opam update`. But this is a separate issue.)

One thing we want to avoid is to make solutions suddenly unavailable: if `opam update` triggers a reinstall of the project but it has now no solutions, this is a bad experience with no escape hatch. This also breaks the concept of lock file since an explicit version changes in a major way.

#### 1: `git rm`

This is the most invasive solution - just remove the package.

#### 2: `available: false`

This ensures that the package is not installable, but keeps the metadata history. It makes it possible to copy the opam file in a repository with just this line removed and have it to work.

#### 3: `avoid-version`

This flag puts a version at the end of the list of candidates, without explicitly forbidding it. (a variant of this opam feature is to only select a version only if explicitly requested but it has not been documented). We could use that as a signal not to trigger revdeps run in CI.

#### 4: `x-skip-ci`

An x-field, ignored by all users, but implemented in CI to filter packages in the revdeps stage.

#### 5: a configuration option in `opam-repo-ci`

This would be a way to configure `opam-repo-ci` to ignore some packages, especially in revdeps runs.

#### 6: using a flag package installed in CI

The idea would be to add `depends: "allow-broken-package"` or `conflicts: "forbid-broken-packages"` to these packages. Normal users would allow installation, while CI would reject the installation (by putting `forbid-broken-package` in the switch invariant, for example).

#### 7: encoding CI-ness in `available:`

Similar to the above, except it's encoded in `available` instead of package dependencies.
[...]

PR #27063, opened 2024-12-10
avoid mirage-* packages that require solo5-bindings-* which are deprecated

hannesm:
I'm still on the fence, we could as well mark these as unavailable. They are bitrot, and requires solo5-bindings-* packages which are deprecated (and modern C compilers error compiling them due to usage of `-Wall -Werror`.

The list in here is taken from https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/327d6ea9ffe66fe2ac64d4ee6373cc6db7845a01


hannesm:
related to #23697

mseri:
I know that they are equivalent, but why not "deprecated" instead? This way it will be easier to spot it for cleanup. (Sorry for not mentioning it in the other PR)

The inconvenience with `avoid-version` is that it is being used for beta packages or preview packages and sometimes the stable version is missing, or a beta version is still being used in the wild and should not be removed. With `deprecated` we give a clear message that it is meant to no longer be used

hannesm:
as you wish :)

mseri:
Thanks

PR #27071, opened 2024-12-11
Frama c.30.0

vprevosto:
Packages for Frama-C 30.0 Zinc release and its associated plug-ins frama-clang and frama-c-metacsl

This PR also adds `x-maintenance-intent` fields for the three packages.

### DIFF
--- a/packages/frama-c-metacsl/frama-c-metacsl.0.8/opam
+++ b/packages/frama-c-metacsl/frama-c-metacsl.0.8/opam
@@ -1,0 +1,58 @@
+opam-version: "2.0"
+synopsis: "MetAcsl plugin of Frama-C for writing pervasives properties"
+description: """\
+MetAcsl let users write properties that need to be checked at particular
+contexts (e.g. each time a location is written to inside a given set
+of functions). It will then generate all the corresponding ACSL
+annotations, leaving it to analysis plug-ins (e.g. WP) to prove the
+resulting clauses."""
+maintainer: "Virgile.Prevosto@cea.fr"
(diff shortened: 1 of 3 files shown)


mseri:
Thanks. All failures seem to be related to dependencies not building or too old LLVM. I'd say that it is good to merge. 

Can you double check if I missed something?

vprevosto:
Thanks for your comment. I agree with you that most failures are indeed usual issues with getting a coherent set of packages for llvm and libclang installed through depext. Do you think it could be I worth it if I try to propose something in this area (in a separate PR, and this is not a strong commitment :sweat_smile: )? In particular, I'd say that the version detection algorithm should be unified (and probably `conf-libclang` should simply rely on `conf-llvm` to get its version number). I don't see how it would be possible to have a working configuration of `conf-llvm.x` and `conf-libclang.y` with `x!=y`.

Back to `frama-c*` packages, there's one failure that worries me a bit, namely [opam-2.0,ocaml-5.2](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/71ded5f52bc36f0bd95eab261effae1e7c4de5ee/variant/extras,opam-2.0-ocaml-5.2,frama-c.30.0) which apparently is caused by the absence of `autoconf`. `conf-autoconf` is only an indirect dependency of frama-c, but shouldn't have it been installed by  the `opam depext` step?

mseri:
The conf-autoconf failure seem to happen only with the old opam 2.0, maybe it is due to some bug in a older opam-depext version. I would not worry about that since every distribution ships more recent opam versions

mseri:
Let's merge this. It would indeed be nice if we can harmonize clang and llvm conf packages, but this can be done separately

PR #27084, opened 2024-12-12
alt-ergo*2.5.0: mark as deprecated instead of avoid-version

hannesm:
//cc @Halbaroth hope that is the intention


hannesm:
@Halbaroth so, avoid-version may have many reasons. And in the upcoming package archival (read more at https://discuss.ocaml.org/t/proposed-package-archiving-policy-for-the-opam-repository), we decided that we keep `avoid-version` packages as is, but remove in the future those marked `deprecated`. If you like to retain alt-ergo 2.5.0 in opam-repository, please disregard this PR. If you're fine with it getting pushed into the opam-repository-archive, this can be done by approving this PR. :)

Halbaroth:
It makes sense to deprecate it. Go for it!

mseri:
Thanks both!

PR #27090, opened 2024-12-12
fix url for opam2web and pla

hannesm:
same as #26921 -- we get "multiple choices" since a branch and a tag exist with the same name.

now, we can specify a different URL and get the tarballs of the tag. that's what is done in this PR.


mseri:
Thanks

PR #27094, opened 2024-12-13
dune-rpc-lwt 3.0.3 has an incorrect bound on result

mseri:
Which makes it uninstallable


mseri:
Seen on https://github.com/ocaml/opam-repository-archive/pull/2

avsm:
Good catch!

PR #27116, opened 2024-12-15
twirp_core, smaws-lib: there's no with-dev, only with-dev-setup

hannesm:
/cc @c-cube for twirp_core, and @chris-armstrong for smaws-lib (who may want to push such a change to their upstream repository)

See https://opam.ocaml.org/doc/Manual.html#pkgvar-with-dev-setup for the available opam variables.


mseri:
Thanks

PR #27118, opened 2024-12-16
release pa_ppx_migrate.0.13

chetmurthy:
This release splits off the OCaml AST migrations (which were taking massive time, b/c lots of version<->version code) into a separate package.  So this package should build fast.


packages/pa_ppx_migrate/pa_ppx_migrate.0.13/opam
```diff
@@ -0,0 +1,49 @@
...
+  "cppo" { >= "1.6.9" }
+  "camlp5-buildscripts" { >= "0.02" }
+  "camlp5"      { >= "8.03.00" }
+  "pa_ppx"      { >= "0.15" }
+  "not-ocamlfind" { >= "0.10" }
+  "ounit" { >= "2.2.7" & with-test}
+  "fmt"
+  "bos" { >= "0.2.0" }
+]
+conflicts: [
```
mseri, line 35: I have the same comment about the conflicts as in the other PR. We should move them to the respective packages if this one is an explicit dependency

chetmurthy:
OK, everything worked except for 2 opensuse builds that failed b/c the opensuse base image already contains a "diff", and when I try to install GNU diffutils, it won't allow it -- so the build fails. I've marked both packages with
```
x-ci-accept-failures: [ "opensuse-tumbleweed" ]
```
but I'm still getting the "opam-ci" task marked as a faiure.

PR #27119, opened 2024-12-16
fixup pa_ppx_regexp.0.02 dependency (to avoid changed behaviour of re 0.11 -> 0.12)

chetmurthy:
re.1.11 had a bug, so pa_ppx_regexp had test that relied on that bug's behaviour.  when re.1.12 fixed the bug, pa_ppx_regexp.0.03 updated the test. But pa_ppx_regexp.0.02 still has the old test, we need to cap the dependency on re.


mseri:
Thanks

PR #27120, opened 2024-12-16
pa_ppx_regexp: pin re version on tests only

mseri:
See https://github.com/ocaml/opam-repository/pull/27119

This should pin re only on the tests, if I understood correctly the issue only concerns a test. Ping @chetmurthy


chetmurthy:
Oh, very nice.  I didn't think of adding two deps to the same package with different constraints.  Very, very nice!

mseri:
Let's double check if it works first :D

mseri:
It worked!

PR #27121, opened 2024-12-16
first release of pa_ppx_migrate_ocaml_parsetree, 0.01

chetmurthy:
This is broken out of pa_ppx_migrate (to make that build much faster), provides migrations between ocaml 4.02.0 <-> 5.3.0 and all intermediate versions.

### DIFF
--- a/packages/pa_ppx_migrate_ocaml_parsetree/pa_ppx_migrate_ocaml_parsetree.0.01/opam
+++ b/packages/pa_ppx_migrate_ocaml_parsetree/pa_ppx_migrate_ocaml_parsetree.0.01/opam
@@ -35,6 +35,6 @@ install: [make "install"]
 url {
   src: "https://github.com/camlp5/pa_ppx_migrate_ocaml_parsetree/archive/refs/tags/0.01.tar.gz"
   checksum: [
-    "sha512=a80f9647a9484bd09ea7d10497ac0619ac7e1a596d789a469ed6d8e8048087296feec0462b58c8539fc9824d98e196a9c25df9e3ee47f227dbe31e1898f8173b"
+    "sha512=0f9313842c6eb07ae6719f49c27858e3a2a94f8cc1dd6d00a4aec27d85b1c44c0a664a9eae511d0edd37c638754ca8a47b1ee5d2ee1cd9b2b38d0cd9e5f03381"
   ]
 }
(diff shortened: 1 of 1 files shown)


chetmurthy:
There are two failures, both due to mangled object-files produced by compile-steps.  I took the docker script for the "4.12" build, and ran it myself locally.   I found that it would sometimes go thru and sometimes fail.  I'm going to "rebuild failed" and see if after a few tries, things go thru, but at this point, I don't think the problem is with the actual CI run.

mseri:
Do you want to investigate this further?

chetmurthy:
It's a little difficult to investigate, b/c it takes so long to run the test (even on my local Radeon server).  I -did- do some investigation.  I found that splitting the install into "deps-only" and then "the package" made the failure go away.  But as I noted above, even the original Docker script sometimes succeeds on my local machine (though never, so far, on the CI cluster).

```
--- DOIT	2024-12-16 14:19:28.133508927 -0800
+++ DOIT.2	2024-12-16 17:25:25.822002683 -0800
@@ -23,21 +23,8 @@
 RUN opam update --depexts || true
 RUN opam pin add -k version -yn pa_ppx_migrate_ocaml_parsetree.0.01 0.01
-RUN opam reinstall pa_ppx_migrate_ocaml_parsetree.0.01; \
-    res=$?; \
-    test "$res" != 31 && exit "$res"; \
-    export OPAMCLI=2.0; \
-    build_dir=$(opam var prefix)/.opam-switch/build; \
-    failed=$(ls "$build_dir"); \
-    partial_fails=""; \
-    for pkg in $failed; do \
-    if opam show -f x-ci-accept-failures: "$pkg" | grep -qF "\"debian-12\""; then \
-    echo "A package failed and has been disabled for CI using the 'x-ci-accept-failures' field."; \
-    fi; \
-    test "$pkg" != 'pa_ppx_migrate_ocaml_parsetree.0.01' && partial_fails="$partial_fails $pkg"; \
-    done; \
-    test "${partial_fails}" != "" && echo "opam-repo-ci detected dependencies failing: ${partial_fails}"; \
-    exit 1
+RUN opam install --deps-only -t pa_ppx_migrate_ocaml_parsetree.0.01
+RUN opam install pa_ppx_migrate_ocaml_parsetree.0.01
 
 END-OF-DOCKERFILE
 docker build --progress plain -f ../Dockerfile .
```

chetmurthy:
OK, all the tests passed.  I'm glad you pushed me about what I thought was a build-system glitch.  It was in fact a bug in my Makefile that only showed up under a poorly-timed race condition.  But fixed now.

mseri:
Thanks for the patience!

PR #27123, opened 2024-12-16
recent lwt_ppx generates code that uses reraise, introduced in recent lwt

raphael-proust:
This PR adds the corresponding constraint


mseri:
Thanks

PR #27143, opened 2024-12-18
rdf, ojs-base, lwt-parallel: use archives from opam-source-archives

hannesm:


shonfeder:
Thanks for fixing these sources!

PR #27145, opened 2024-12-18
OCaml 5.3.0: release candidate

Octachron:
This PR adds the three packages (`ocaml-compiler`, `ocaml-base-compiler` and `ocaml-variants...+options`) for the release candidate of OCaml 5.3.0.

Compared to the second beta release, the release candidate contains one fix for the new linking order check (in presence of pack), and fix for type expression generalization.

Simultaneously, this PR removes the `compiler` flag from the `ocaml-compiler` package  and the non-windows `host-system-*` dependencies as discussed with @dra27 and @kit-ty-kate .

---

### Type system regression fix

- 13690: some type expressions were incorrectly not generalized (because they
   were assigned to the wrong level pool)

### Documentation update

- 13666: Rewrite parts of the example code around nested lists in Chapter 6
  (Polymorphism and its limitations -> Polymorphic recursion) giving the
  "depth" function [in the non-polymorphically-recursive part of the example]
  a much more sensible behavior; also fix a typo and some formatting.
  (Frank Steffahn, review by Florian Angeletti)

### Compiler user-interface and warnings:

* 12084, +13669, +13673: Check link order when creating archive and when using
   ocamlopt.

### DIFF
--- a/packages/ocaml-base-compiler/ocaml-base-compiler.5.3.0~rc1/opam
+++ b/packages/ocaml-base-compiler/ocaml-base-compiler.5.3.0~rc1/opam
@@ -1,0 +1,29 @@
+opam-version: "2.0"
+license: "LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception"
+synopsis: "First release candidate release of OCaml 5.3.0"
+maintainer: [
+  "David Allsopp <david@tarides.com>"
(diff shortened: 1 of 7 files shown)


kit-ty-kate (approved):
lgtm. Tested locally on Linux/arm64

PR #27147, opened 2024-12-19
Add support for conf-mbedtls on windows

tobil4sk:
Updates conf-mbedtls with the following updates:
- Uses gcc instead of cc on windows, since there is no cc symlink
- Updated metadata
- Removed duplicate macos homebrew dependency.
- Depends on new `conf-mingw-w64-mbedtls-x86_64` package.

There is no official `mingw64-x86_64-mbedtls` package for cygwin, however, it's possible that it has been installed from source.

### DIFF
--- a/packages/conf-mbedtls/conf-mbedtls.2/opam
+++ b/packages/conf-mbedtls/conf-mbedtls.2/opam
@@ -4,6 +4,7 @@ maintainer: "Andy Li <andy@onthewings.net>"
 homepage: "https://www.trustedfirmware.org/projects/mbed-tls/"
 bug-reports: "https://github.com/Mbed-TLS/mbedtls/issues"
 license: "Apache-2.0"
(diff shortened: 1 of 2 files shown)


tobil4sk:
Currently failing due to: https://github.com/ocurrent/opam-repo-ci/issues/420.

Also failing because it uses `extra-source` to build a test program to ensure that mbedtls is installed. This is not a new addition, the previous version also worked like this. Ideally, pkg-config would be used to check if it is installed, however, mbedtls only recently [started supporting pkg-config](https://github.com/Mbed-TLS/mbedtls/pull/8691) and it means that on many systems with older versions it won't work.

I'm not sure what would be best to do here.

mseri:
I think this is all right, thanks

mseri:
And sorry for the massive delay

tobil4sk:
Thanks!

Issue #27131, opened 2024-12-17
sources not available - gforge.inria.fr


hannesm:
Dear everybody,

I figure that we're still using gforge.inria.fr in a few packages, and these source tarballs are not available (instead, they return 403). Does anyone have these tarballs at hand? I tried to get them from the opam.ocaml.org/cache without success.

Here are the urls and their checksums.

| package | url | md5 checksum |
|---------|-----|--------------|
| dose.3.2.2 | https://gforge.inria.fr/frs/download.php/file/33677/dose3-3.2.2.tar.gz | 3ffed1987a040024076c08f4a7af9b21 |
| javalib.2.3.1 | https://gforge.inria.fr/frs/download.php/file/34920/javalib-2.3.1.tar.bz2 | 31e33a2c462f8bbda79218f9041f3639 |
| javalib.2.3.2 | https://gforge.inria.fr/frs/download.php/file/36092/javalib-2.3.2.tar.bz2 | 17ec3fabb1472dd3324ec68fb63b22a6 |
| javalib.2.3.3 | https://gforge.inria.fr/frs/download.php/file/36307/javalib-2.3.3.tar.bz2 | a4d4b06e8f4860db34c128e760fa8397 |
| javalib.2.3.4 | https://gforge.inria.fr/frs/download.php/file/37154/javalib-2.3.4.tar.bz2 | 4707eda130e41d1d7f0506cc0c77eff1 |
| javalib.2.3.5 | https://gforge.inria.fr/frs/download.php/file/37655/javalib-2.3.5.tar.bz2 | 7111616e58b366ca269baf4867677082 |
| sawja.1.5.1 | https://gforge.inria.fr/frs/download.php/file/34921/sawja-1.5.1.tar.bz2 | 3a9b750faa7a368ec6aaaa0206fac40b |
| sawja.1.5.2 | https://gforge.inria.fr/frs/download.php/file/36093/sawja-1.5.2.tar.bz2 | 12afa3651ea39413d2cb0e1775f587c4 |
| sawja.1.5.3 | https://gforge.inria.fr/frs/download.php/file/37403/sawja-1.5.3.tar.bz2 | 25ff421a3f932881234ed5b05b94ac8d |
| sawja.1.5.4 | https://gforge.inria.fr/frs/download.php/file/37656/sawja-1.5.4.tar.bz2 | ccb408c9cced1e5e3a0d04e4fb12a40f |

hannesm:
[...]

PR #27149, opened 2024-12-19
mark some more packages unavailable

hannesm:
see https://github.com/ocaml/opam-repository/issues/27131

this PR is based on observations from https://opam.robur.coop/status

EDIT: I tried to find the sources, on my laptop, on opam.ocaml.org/cache - without success.


shonfeder:
Thanks for this valuable maintenance work!

PR #27150, opened 2024-12-19
adapt the checksum of the ocaml 5.1 reference manual

hannesm:
I'm not sure why the checksum changed (in december 2023) -- but since this is no code, only the reference manual, let's just use the artifact that caml.inria.fr provides.


mseri:
Are the webpages the same in the cache and in that tarball?

hannesm:
I'm unable to find the cache tarball, otherwise I'd have uploaded that instead.

PR #27153, opened 2024-12-19
mark compiler prereleases as deprecated

hannesm:
alpha/beta/rc of compiler releases are deprecated now

as discussed earlier today with @Octachron


hannesm:
This marks them with the flag deprecated, eventually removes the avoid-version flag, and sets availability to opam-version >= 2.2 (that's when the deprecated flag was introduced).

Octachron:
The changes seem fine to me. In particular, I think we could deprecate prerelease version of the compiler as soon as the version is released, since I don't see a use case where it would be preferable to stay on the alpha1 or rc1 once the full version is available.

mseri:
Thanks!

PR #27169, opened 2024-12-22
[new release] allegro5 (0.1)

Frigory33:
OCaml Allegro 5

- Project page: <a href="https://github.com/Frigory33/ocaml-allegro-5">https://github.com/Frigory33/ocaml-allegro-5</a>

##### CHANGES:

First release. Includes elements of the following Allegro 5 modules:

- Displays
- Events
- Graphics
- Keyboard
- Mouse
- System
- Time
- Timer
- Font addons
- Image addon
- Primitives addon

### DIFF
--- a/packages/allegro5/allegro5.0.1/opam
+++ b/packages/allegro5/allegro5.0.1/opam
@@ -1,0 +1,46 @@
+opam-version: "2.0"
+synopsis: "OCaml Allegro 5"
+description: "An OCaml binding of the Allegro 5 game & multimedia C library"
+maintainer: ["Sylvain Chiron <chironsylvain@orange.fr>"]
+authors: ["Sylvain Chiron <chironsylvain@orange.fr>"]
+license: "LGPL-3.0-or-later"
+homepage: "https://github.com/Frigory33/ocaml-allegro-5"
+bug-reports: "https://github.com/Frigory33/ocaml-allegro-5/issues"
+depends: [
+  "ocaml" {>= "4.08"}
+  "dune" {>= "2.9"}
+  "conf-allegro5"
+  "odoc" {with-doc}
+]
+build: [
+  ["dune" "subst"] {dev}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "--promote-install-files=false"
+    "@install"
+    "@runtest" {with-test}
+    "@doc" {with-doc}
+  ]
+  ["dune" "install" "-p" name "--create-install-files" name]
+]
+dev-repo: "git+https://github.com/Frigory33/ocaml-allegro-5.git"
+post-messages: [
+  """\
+We're sorry that your install failed. Please check in the README.md file
+what minimal version of Allegro this package requires."""
+    {failure}
+]
+url {
+  src:
+    "https://github.com/Frigory33/ocaml-allegro-5/releases/download/0.1/allegro5-0.1.tbz"
+  checksum: [
+    "sha256=330291289db836b9f555cfd73a673654e27bf47e05115bf0990db76c23cc770b"
+    "sha512=e29664381a381bb846d244144fefc427c6287fb1064bae0f7bf755a6d72683b385d21bc2d14750a10bcc5c362ef5d9a4adfb5b610cf99e7b45273cd132000ffc"
+  ]
+}
+x-commit-hash: "a9c2b91f6a298cfc4c7aa0fd7e22db4bc4150636"
--- a/packages/conf-allegro5/conf-allegro5.1/opam
+++ b/packages/conf-allegro5/conf-allegro5.1/opam
@@ -1,0 +1,44 @@
+opam-version: "2.0"
+maintainer: "allegro-developers@lists.liballeg.org"
+homepage: "https://liballeg.org"
+authors: "The Allegro authors"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+license: "Giftware"
+build: ["pkg-config" "allegro-5"]
+depexts: [
+  [
+    "allegro5-devel"
+    "allegro5-addon-acodec-devel"
+    "allegro5-addon-audio-devel"
+    "allegro5-addon-dialog-devel"
+    "allegro5-addon-image-devel"
+    "allegro5-addon-physfs-devel"
+    "allegro5-addon-ttf-devel"
+    "allegro5-addon-video-devel"
+  ] {os-distribution = "centos" | os-distribution = "fedora"}
+  ["allegro"] {os-family = "suse" | os-family = "opensuse"}
+  [
+    "liballegro5-dev"
+    "liballegro-acodec5-dev"
+    "liballegro-audio5-dev"
+    "liballegro-dialog5-dev"
+    "liballegro-image5-dev"
+    "liballegro-physfs5-dev"
+    "liballegro-ttf5-dev"
+    "liballegro-video5-dev"
+  ] {os-family = "debian" | os-family = "ubuntu"}
+  ["allegro5"] {os-distribution = "nixos"}
+  ["allegro"] {os-family = "arch"}
+  ["allegro-dev"] {os-family = "alpine"}
+  ["allegro"] {os-distribution = "homebrew"}
+  ["allegro5"] {os = "freebsd"}
+  ["allegro5"] {os = "netbsd"}
+  ["mingw-w64-allegro"] {os-distribution = "msys2"}
+]
+synopsis: "Virtual package relying on Allegro 5 development libraries"
+description:
+  "This package can only install if Allegro 5 development libraries are accessible through pkg-config."
+depends: [
+  "conf-pkg-config" {build}
+]
+flags: conf


Frigory33:
I don’t know how to add an external dependency to my library. It needs the Allegro 5 development library and its addons. Thus the build failed. What should I do?

shonfeder:
First, thank you for taking the time to publish your work on these bindings!

> I don’t know how to add an external dependency to my library. It needs the Allegro 5 development library and its addons. Thus the build failed. What should I do?

opam is able to records system dependencies for packages available in a number of popular distros via the [`depext` field](https://opam.ocaml.org/doc/Manual.html#opamfield-depexts). As noted in those docs, the `depext` field should be used in a `conf-` package, that declares a (reusable) system dependency. Here's an example of the `conf-` package for `git`: https://github.com/ocaml/opam-repository/blob/master/packages/conf-git/conf-git.1.1/opam

For systems where the dependency is not available from a supported package manager, you can add a [post-install message](https://opam.ocaml.org/doc/Manual.html#opamfield-post-messages) that is printed only on failure, pointing to users towards how the dependency can be met. Here's an example:

https://github.com/toots/opam-repository/blob/d31b2b983a4cae4ba367ae61c522f8570ff6783a/packages/ffmpeg-avdevice/ffmpeg-avdevice.1.2.1/opam#L36-L41

For these cases, we an also add the [`x-ci-accept-failure`](https://github.com/ocaml/opam-repository/wiki/How-to-deal-with-CI#annex) field to help the CI know that failures on the given platforms is expected.

Let us know if you have any additional questions, and/or don't hesitate to ask on https://discuss.ocaml.org/

Frigory33:
Thank you very much for your support!

I don’t know how to deal with MSYS2. There is no Allegro package in Cygwin but [there is one in MSYS2](https://packages.msys2.org/base/mingw-w64-allegro).

And I don’t know how to deal with windows with `x-ci-accept-failure`.

packages/conf-allegro5/conf-allegro5.1/opam
```diff
@@ -0,0 +1,41 @@
...
+    "liballegro-image5-dev"
+    "liballegro-primitives5-dev"
+    "liballegro-ttf5-dev"
+  ] {os-family = "debian" | os-family = "ubuntu"}
+  ["allegro5"] {os-distribution = "nixos"}
+  ["allegro"] {os-family = "arch"}
+  ["allegro-dev"] {os-family = "alpine"}
+  ["allegro"] {os-distribution = "homebrew"}
+  ["allegro5"] {os = "freebsd"}
+  ["allegro5"] {os = "netbsd"}
```
shonfeder, line 32: It looks like we might get some windows packaging support via msys2, via https://packages.msys2.org/base/mingw-w64-allegro

If you're interested in giving that a shot, you could add

```suggestion
  ["allegro5"] {os = "netbsd"}
  ["mingw-w64-allegro"] {os-distribution = "msys2"}
```

Of course, you don't need to do everything at once, so if you'd just like us to move forward with publising this as is, I think that is fine!

packages/allegro5/allegro5.0.1/opam
```diff
@@ -0,0 +1,47 @@
...
+maintainer: ["Sylvain Chiron <chironsylvain@orange.fr>"]
+authors: ["Sylvain Chiron"]
+license: "LGPL-3.0-or-later"
+homepage: "https://github.com/Frigory33/ocaml-allegro-5"
+bug-reports: "https://github.com/Frigory33/ocaml-allegro-5/issues"
+depends: [
+  "ocaml"
+  "dune" {>= "3.11"}
+  "odoc" {with-doc}
+  "conf-allegro5"
```
shonfeder, line 13: To address the test failure in https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/ec7ebf3d7e33040e1e6f69592341b57233ab0173/variant/distributions,alpine-3.20-ocaml-4.14,allegro5.0.1,tests which looks like

```
https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/ec7ebf3d7e33040e1e6f69592341b57233ab0173/variant/distributions,alpine-3.20-ocaml-4.14,allegro5.0.1,tests
```

I think we'll need to declare the `dune-site` package as a test dependency. This should do the trick:

```suggestion
  "conf-allegro5"
  "dune-site" {with-test}
```

shonfeder: For future releases, you can reflect this change back into your https://github.com/Frigory33/ocaml-allegro-5/blob/main/dune-project

shonfeder (changes requested):
Thank you the updates! Things are coming along well here :)

I have suggested one fix for the tests failures here, and one suggestion to extend some windows packaging support.

Could you also take a look some of the build failures to see if they can be fixed or if they are expected?

Frigory33:
I’ve done what I could, but now there are problems with `pkg-config`. My packages just need the `pkg-config` command. I saw in the `conf-sdl2` opam file that there is a weird case for Cygwin, using `pkgconf` instead of `pkg-config`. Should I simply use `pkgconf` instead of `pkg-config` everywhere?

I also changed my test and my release of the `ocaml-allegro-5` repository, but I don’t know how to make the tests here to use the new files, while it’s still the release “0.1.”

shonfeder:
Hello! Thank you again for your careful work here! I apologize for the delay: we are still getting caught up again after the holiday.

> there are problems with pkg-config

Ah yes, I see. It looks like this is a problem with `mingw-w64-shims`:

```
  === recompile 1 package
    - recompile mingw-w64-shims 0.2.0  [uses conf-pkg-config]
```

This needs to be fixed in that conf package, iiuc, so out of scope here.

> I also changed my test and my release of the ocaml-allegro-5 repository, but I don’t know how to make the tests here to use the new files, while it’s still the release “0.1.”

The cleanest way would be to update this PR to point to the corrected, updated version of your package. With this PR, you can just update the version number in the directory name, and update the sources and sha.  If you are using `opam-publish` or `dune-release` you can run them in a mode to produce the new opam file, and replace the content here with that. Or you may open a new PR with your updated release and close this one.

This is advisable, because it will ensure that when we check your package during reversed dependency checks, we do not get false positives. Otherwise, you will not protected by our CI against breaking changes of your dependencies. Please let me know if you have questions about how to do this! I am happy to support.

Frigory33:
I think I’ve done what you said, but now there’s a new failing check (maintenance-intent present)… I don’t understand.

shonfeder:
That's a newly added check and it's not stable yet (perhaps should be removed from CI for the moment). Sorry for the noise there!

Frigory33:
The build fails with OCaml 4 variants. I get the same problem locally with an opam switch. I don’t know what to do… Do you have an idea of where this bug comes from? (The errors are not specific to the `display` and `timer` modules of my library; I get the same errors if I keep only two modules.) Should I require OCaml 5 for my package? It might be the only solution…

Also, can you help me on publishing the documentation of my package on GitHub with dune-release? I don’t understand what I should do.

shonfeder:
I will take a look at both the builds on OCaml 4 and your documentation situation this weekend! I am optimistic that I'll be able to help on both accounts. Thanks again for your persistence and patience here!

shonfeder:
> The build fails with OCaml 4 variants.

I wanted to note quickly that the build failures here are not limited to OCaml 4.x. See, e.g., https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/da45543bcf6ada03ab09d168b588f94a1c22a7cb/variant/distributions,fedora-41-ocaml-5.3,allegro5.0.2

```
[ERROR] The compilation of allegro5.0.2 failed at "dune build -p allegro5 -j 71 --promote-install-files=false @install".

#=== ERROR while compiling allegro5.0.2 =======================================#
# context              2.4.0~alpha1~dev | linux/x86_64 | ocaml-base-compiler.5.3.0 | pinned(https://github.com/Frigory33/ocaml-allegro-5/releases/download/0.2/allegro5-0.2.tbz)
# path                 ~/.opam/5.3/.opam-switch/build/allegro5.0.2
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p allegro5 -j 71 --promote-install-files=false @install
# exit-code            1
# env-file             ~/.opam/log/allegro5-7-8d9a0a.env
# output-file          ~/.opam/log/allegro5-7-8d9a0a.out
### output ###
# File "lib/dune", lines 19-26, characters 0-277:
# 19 | (library
# 20 |  (name al5)
# 21 |  (public_name allegro5)
# 22 |  (foreign_stubs
# 23 |   (language c)
# 24 |   (names al5 display events graphics keyboard mouse system time timer font image primitives)
# 25 |   (flags -std=c11 -Wall -O2 -I -fPIC (:include c_flags.sexp)))
# 26 |  (c_library_flags (:include lib_flags.sexp)))
# (cd _build/default && /home/opam/.opam/5.3/bin/ocamlmklib -g -o lib/al5_stubs lib/timer.o lib/time.o lib/system.o lib/primitives.o lib/mouse.o lib/keyboard.o lib/image.o lib/graphics.o lib/font.o lib/events.o lib/display.o lib/al5.o -ldopt -lallegro_image -ldopt -lallegro_primitives -ldopt -lallegro_ttf -ldopt -lallegro_font -ldopt -lallegro)
# /usr/bin/ld: lib/timer.o: relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a shared object; recompile with -fPIC
# /usr/bin/ld: failed to set dynamic section sizes: bad value
```

shonfeder:
Please note that I am not at all expert at working with C-stubs!

I have been trying to dig into this error. In case it is useful, these are some of the resources and discussions I've consulted:

- https://dune.readthedocs.io/en/stable/reference/foreign-stubs.html
- https://github.com/ocaml/dune/issues/1948
- https://github.com/ocaml/ocaml/pull/8631
- https://stackoverflow.com/questions/19364969/compilation-fails-with-relocation-r-x86-64-32-against-rodata-str1-8-can-not

However, I wonder if the solution may lie in https://github.com/ocaml/dune/issues/5809 . The latter issue includes a proposed workaround in the issue description, and tho it was supposedly fixed in an earlier version of dune, a possible regression is reported there.

In the switch where you are able to reproduce this, could show what you see in the build logs, as in https://github.com/ocaml/dune/issues/5809#issuecomment-1831231370 ? So, could you run something like

```
$ dune --version
$ dune build
$ grep '/usr/bin/gcc' _build/log
```

and share the results? Perhaps also just investigating the `_build/log` on your own to see if you the compilation of the c stubs looks like it may answer to the error 

```
# /usr/bin/ld: lib/timer.o: relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a shared object; recompile with -fPIC
```

If this does not help point the way, you may want to consider seeking help on https://discuss.ocaml.org/ ! Many helpful folks with deep expertise interfacing with c there :D

Frigory33:
In the build log, there is nothing more than the error report we get in Opam-CI. So I’m now asking [on discuss.ocaml.org](https://discuss.ocaml.org/t/linking-problem-in-compiling-my-binding-of-allegro-5/15972).

Frigory33:
OK, so now my tests are failing with stack overflow with OCaml 4.

I cannot reproduce it locally because I can’t install `dune-site` or, more precisely, its dependency `dyn`. It seems it looks for some files in my `default` opam switch instead of the current `ocaml-base-compiler.4.14.2` switch… Are you able to install `dune-site` with the following commands?
```
opam switch create ocaml-base-compiler.4.14.2
eval $(opam env --switch=ocaml-base-compiler.4.14.2)
opam install dune-site
```
It does not work for me.

shonfeder:
That is very curious. The error is

```
Processing  2/3: [allegro5: dune build]
+ /home/opam/.opam/opam-init/hooks/sandbox.sh "build" "dune" "build" "-p" "allegro5" "-j" "71" "--promote-install-files=false" "@install" "@runtest" (CWD=/home/opam/.opam/4.12/.opam-switch/build/allegro5.0.2)
- File "test/dune", line 10, characters 7-25:
- 10 |  (name test_ocaml_allegro)
-             ^^^^^^^^^^^^^^^^^^
- (cd _build/default/test && ./test_ocaml_allegro.exe)
- Fatal error: exception Stack overflow
[ERROR] The compilation of allegro5.0.2 failed at "dune build -p allegro5 -j 71 --promote-install-files=false @install @runtest".

#=== ERROR while compiling allegro5.0.2 =======================================#
# context              2.4.0~alpha1~dev | linux/x86_64 | ocaml-base-compiler.4.12.1 | pinned(https://github.com/Frigory33/ocaml-allegro-5/releases/download/0.2/allegro5-0.2.tbz)
# path                 ~/.opam/4.12/.opam-switch/build/allegro5.0.2
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p allegro5 -j 71 --promote-install-files=false @install @runtest
# exit-code            1
# env-file             ~/.opam/log/allegro5-7-84eda8.env
# output-file          ~/.opam/log/allegro5-7-84eda8.out
### output ###
# File "test/dune", line 10, characters 7-25:
# 10 |  (name test_ocaml_allegro)
#             ^^^^^^^^^^^^^^^^^^
# (cd _build/default/test && ./test_ocaml_allegro.exe)
# Fatal error: exception Stack overflow
```

Have you tried running the docker builds to get reproduction? A docker recipe is included in each job log. E.g., for one [compilers,4.14,allegro5.0.2,tests](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/e573944dde8fbb1db612e0257e42584b0daf41c3/variant/compilers,4.14,allegro5.0.2,tests), you should be able to drop this into a script, run it, and get a reproduction that way:

```
cd $(mktemp -d)
git clone --recursive "https://github.com/ocaml/opam-repository.git" && cd "opam-repository" && git fetch origin "refs/pull/27169/head" && git reset --hard e573944d
git fetch origin master
git merge --no-edit 4d67ef48e6a36b5ba83d930da14d544d46025751
cat > ../Dockerfile <<'END-OF-DOCKERFILE'
FROM ocaml/opam:debian-12-ocaml-4.14@sha256:a4b2191841204ae998edfb1969d3f7cc2689fdcccc16ecebde7b39f99ffd5324
USER 1000:1000
WORKDIR /home/opam
RUN sudo ln -f /usr/bin/opam-dev /usr/bin/opam
RUN opam init --reinit -ni
RUN opam option solver=builtin-0install && opam config report
ENV OPAMDOWNLOADJOBS="1"
ENV OPAMERRLOGLEN="0"
ENV OPAMPRECISETRACKING="1"
ENV CI="true"
ENV OPAM_REPO_CI="true"
RUN rm -rf opam-repository/
COPY --chown=1000:1000 . opam-repository/
RUN opam update --depexts || true
RUN opam pin add -k version -yn allegro5.0.2 0.2
RUN opam reinstall allegro5.0.2; \
    res=$?; \
    test "$res" != 31 && exit "$res"; \
    export OPAMCLI=2.0; \
    build_dir=$(opam var prefix)/.opam-switch/build; \
    failed=$(ls "$build_dir"); \
    partial_fails=""; \
    for pkg in $failed; do \
    if opam show -f x-ci-accept-failures: "$pkg" | grep -qF "\"debian-12\""; then \
    echo "A package failed and has been disabled for CI using the 'x-ci-accept-failures' field."; \
    fi; \
    test "$pkg" != 'allegro5.0.2' && partial_fails="$partial_fails $pkg"; \
    done; \
    test "${partial_fails}" != "" && echo "opam-repo-ci detected dependencies failing: ${partial_fails}"; \
    exit 1
RUN (opam reinstall --with-test allegro5.0.2) || true
RUN opam reinstall --with-test --verbose allegro5.0.2; \
    res=$?; \
    test "$res" != 31 && exit "$res"; \
    export OPAMCLI=2.0; \
    build_dir=$(opam var prefix)/.opam-switch/build; \
    failed=$(ls "$build_dir"); \
    partial_fails=""; \
    for pkg in $failed; do \
    if opam show -f x-ci-accept-failures: "$pkg" | grep -qF "\"debian-12\""; then \
    echo "A package failed and has been disabled for CI using the 'x-ci-accept-failures' field."; \
    fi; \
    test "$pkg" != 'allegro5.0.2' && partial_fails="$partial_fails $pkg"; \
    done; \
    test "${partial_fails}" != "" && echo "opam-repo-ci detected dependencies failing: ${partial_fails}"; \
    exit 1

END-OF-DOCKERFILE
docker build -f ../Dockerfile .
```

Unless your tests are building a *very* large file (which can indeed cause SO, this may be a dune bug. If you are building a very large file, you may want to define those tests in a target which is not run with `dune test` but only with something like `dune build @large-file-test` or something.

Frigory33:
Hello, sorry for the delay. My test is not expensive, does not build a large file. It is a simple test program, like an example program of Allegro.

I don’t know how to manage other build targets, so I’ve simply moved my test to another repository, I think this will make things simpler.

I was able to install ocaml 4.14.2 with dune-site on Windows and the bug wasn’t present here. But I was able to reproduce the problem locally with Docker (I had never used Docker before). But I didn’t know what to do with that.

I’ve just updated everything, this should work know… But what do we do for Windows?

Also, you said you could help me for the documentation. How can I get my documentation published? I suppose this will be on a github.io site but I don’t know how to create it.

Frigory33:
Finally, I got it to the documentation. I just had to add the `documentation` stanza in my `dune-project` file, following the scheme given by dune-release for the URI, then `dune-release publish doc` created a branch that I had to set as the Pages branch in my repository settings.

So now, you might merge my pull request, I think.

shonfeder:
Sorry for the delayed reply here!

shonfeder:
I

shonfeder (approved):

PR #27175, opened 2024-12-24
morbig 0.9.* is not compatible with ocaml >= 5.1

mseri:
Fails with
```
ar: /home/opam/.opam/5.2/lib/ocaml/libcamlstr.a: No such file or directory
```


mseri:
Merged by mistake, but fixes the issue anyway :)

PR #27178, opened 2024-12-24
Prepare for LLVM 18, link to local repo for scripts and patches

alan-j-hu:
This references the scripts and patches for my fork of opam-source-archive, just so I can see if things work on CI... If this is accepted, they will need to be changed to point to ocaml/opam-source-archive.

Context: https://github.com/ocaml/opam-source-archives/pull/40

### DIFF
--- a/packages/llvm/llvm.18-shared/opam
+++ b/packages/llvm/llvm.18-shared/opam
@@ -46,7 +46,7 @@ extra-source "AddOCaml.cmake.patch" {
   src:
     "https://raw.githubusercontent.com/ocaml/opam-source-archives/main/patches/llvm/AddOCaml.cmake.patch.18"
   checksum: [
-    "sha256=cf9023f52075bd7ec7fb808a12fdecb46ca0690cc24dd8d9defd051bcd506493"
-    "md5=667a95b875d607b5d7567d7c6cf7a494"
+    "sha256=a4fac2d6ebfaa707906dc576b99f80efe42ee8857647eb881a07f8fb5ebcddd2"
+    "md5=5a6c39055106cfc700eef2ad0255cba6"
   ]
 }
--- a/packages/llvm/llvm.18-shared/opam
+++ b/packages/llvm/llvm.18-shared/opam
@@ -36,7 +36,7 @@ url {
 }
 extra-source "install.sh" {
   src:
-    "https://raw.githubusercontent.com/ocaml/opam-source-archives/refs/heads/main/patches/llvm/install.sh.18"
+    "https://raw.githubusercontent.com/ocaml/opam-source-archives/main/patches/llvm/install.sh.18"
   checksum: [
     "sha256=91311c9a90258e6819adda4132f9f64df18f1f0b9df905e7f88be325ed18dce7"
     "md5=70201a5ad01e104f7572e1fa0d498eaa"
@@ -44,9 +44,9 @@ extra-source "install.sh" {
 }
 extra-source "AddOCaml.cmake.patch" {
   src:
-    "https://raw.githubusercontent.com/ocaml/opam-source-archives/refs/heads/main/patches/llvm/AddOCaml.cmake.patch.18"
+    "https://raw.githubusercontent.com/ocaml/opam-source-archives/main/patches/llvm/AddOCaml.cmake.patch.18"
   checksum: [
-    "sha256=3c9abe430b0868e5937e9944133d521f62677244e0a345e511f66cb9502e406a"
-    "md5=a9ef328c5a793812c25a5f8777316ad9"
+    "sha256=cf9023f52075bd7ec7fb808a12fdecb46ca0690cc24dd8d9defd051bcd506493"
+    "md5=667a95b875d607b5d7567d7c6cf7a494"
   ]
 }
--- a/packages/llvm/llvm.18-shared/opam
+++ b/packages/llvm/llvm.18-shared/opam
@@ -44,7 +44,7 @@ extra-source "install.sh" {
 }
 extra-source "AddOCaml.cmake.patch" {
   src:
-    "https://raw.githubusercontent.com/ocaml/opam-source-archives/main/patches/llvm/AddOCaml.cmake.patch.18"
+    "https://github.com/alan-j-hu/opam-source-archives/raw/e5d49c1c15149d3a239d237634f5fc85d903718d/patches/llvm/AddOCaml.cmake.patch.18"
   checksum: [
     "sha256=a4fac2d6ebfaa707906dc576b99f80efe42ee8857647eb881a07f8fb5ebcddd2"
     "md5=5a6c39055106cfc700eef2ad0255cba6"
--- a/packages/llvm/llvm.18-shared/opam
+++ b/packages/llvm/llvm.18-shared/opam
@@ -36,7 +36,7 @@ url {
 }
 extra-source "install.sh" {
   src:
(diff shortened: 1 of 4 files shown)


alan-j-hu:
I'm getting cascading errors because of this:

```
Error in conf-llvm-static.18: error 46: Package is flagged "conf" but has source, install or remove instructions
Error in conf-llvm-static.18: No package source directory provided.
Error in conf-llvm-shared.18: error 46: Package is flagged "conf" but has source, install or remove instructions
Error in conf-llvm-shared.18: No package source directory provided.
```

This is despite the fact that this all worked on my computer testing locally, and `conf-llvm.17` and prior all have `build` fields and were accepted. Today, running `opam lint` on `conf-llvm.17` gives this error. Is this something new?

alan-j-hu:
Okay, I misunderstood the CI error message.

alan-j-hu:
I'm changing this PR from "draft" to "ready to review" because the opam maintainers have not seen my PR over at the opam-source-archives repository. They may have only glanced at this one, and misunderstood its status because of the "draft" state.

I need the opam maintainers to look at this PR. However, as it stands, it points at scripts on my fork of opam-source-archives. Therefore, my fork of opam-source-archives therefore needs to be merged first, so I can change the URLs on this PR.

mseri:
You can link the opam-source-archive patches and files now, thanks!

alan-j-hu:
I have changed the links now.

alan-j-hu:
Hold on, don't merge this yet. I think there might be a problem with bytecode executables, which I need to check out.

alan-j-hu:
Bytecode executables were not working. (I'm not sure if they were even working before, as bytecode executables do not compile under the LLVM 14 package for a different reason.)

To fix, please merge https://github.com/ocaml/opam-source-archives/pull/42 first, and then merge this.

alan-j-hu:
Argh, I spoke too soon. The bytecode executable seems to work, but I think I still made a mistake in the patch, which I'm checking out right now.

After the creation of the opam-source-archives repository, it's really, really difficult to support packages that depend on files in the opam-repository, because I need to move between two repositories and make two merge requests, leaving a lot of room for mistakes.

alan-j-hu:
Okay, https://github.com/ocaml/opam-source-archives/pull/42 should be ready to merge, then this can be merged after.

Right now, I have four projects open:
- opam-repository
- opam-source-archives
- llvm-project
- My OCaml LLVM test project to make sure the LLVM package works

I need to make sure that the llvm-project test suite passes, and that the LLVM bindings also work properly when installed out-of-tree. I realized that bytecode executables didn't work when the `-custom` flag was passed (as it was), and set out to fix the issue, which is what the commotion was about.

alan-j-hu:
AFAIK, in LLVM 14 (the last version opam to use CMake to build), bytecode compilation wasn't working, because of a different issue that has since been fixed: https://discuss.ocaml.org/t/llvm-symbol-not-found-for-bytecode-compilation/8728. Now, I would like to pay closer attention to this package to make sure all the moving parts work.

mseri:
Thanks for the effort!

By the way, you don’t need the opam-source-archive until the patch is ready. It could live on your repo or in a gist for example, giving you fast iteration, and once everything is ready and tested, we can move it to the archive

packages/llvm/llvm.18-shared/opam
```diff
@@ -0,0 +1,52 @@
...
+  src:
+    "https://raw.githubusercontent.com/ocaml/opam-source-archives/main/patches/llvm/install.sh.18"
+  checksum: [
+    "sha256=91311c9a90258e6819adda4132f9f64df18f1f0b9df905e7f88be325ed18dce7"
+    "md5=70201a5ad01e104f7572e1fa0d498eaa"
+  ]
+}
+extra-source "AddOCaml.cmake.patch" {
+  src:
+    "https://raw.githubusercontent.com/ocaml/opam-source-archives/main/patches/llvm/AddOCaml.cmake.patch.18"
```
mseri, line 47:
```suggestion
    "https://github.com/alan-j-hu/opam-source-archives/raw/e5d49c1c15149d3a239d237634f5fc85d903718d/patches/llvm/AddOCaml.cmake.patch.18"
```

to iterate on the patch and test it while you are working on it, this or any other remote location would be fine

packages/llvm/llvm.18-static/opam
```diff
@@ -0,0 +1,53 @@
...
+  src:
+    "https://raw.githubusercontent.com/ocaml/opam-source-archives/refs/heads/main/patches/llvm/install.sh.18"
+  checksum: [
+    "sha256=91311c9a90258e6819adda4132f9f64df18f1f0b9df905e7f88be325ed18dce7"
+    "md5=70201a5ad01e104f7572e1fa0d498eaa"
+  ]
+}
+extra-source "AddOCaml.cmake.patch" {
+  src:
+    "https://raw.githubusercontent.com/ocaml/opam-source-archives/refs/heads/main/patches/llvm/AddOCaml.cmake.patch.18"
```
mseri, line 47:
```suggestion
    "https://github.com/alan-j-hu/opam-source-archives/raw/e5d49c1c15149d3a239d237634f5fc85d903718d/patches/llvm/AddOCaml.cmake.patch.18"
```

Drup:
Just a gentle ping, because I also would like this to work. :)

alan-j-hu:
https://github.com/alan-j-hu/llvm-ocaml-ci I updated my CI repo to match the hashes, and can confirm that the build still works.

mseri:
Every now and then it fails with
```
# [  0%] Building OCaml library llvm
# ld: error: unable to find library -lzstd
# cc: error: linker command failed with exit code 1 (use -v to see invocation)
```
but somehow inconsistently. I see the failure on the conf package, and later on the package itself (using this conf) builds fine. Do you know why?

I am rerunning some of the builds to double check

alan-j-hu:
Ugh, first this is a problem with llvm-config. It's just hardcoded to output `-lzstd`, but it doesn't add any flags to look for zstd in the right paths. I searched online and see similar issues at https://github.com/ziglang/zig/issues/19617 and https://github.com/rust-lang/rust/issues/135503.

Furthermore, Homebrew started installing libraries under `/'opt/homebrew` instead of `/usr/local` under Silicon Macs, breaking stuff for a lot of people. It looks like the Mac build works on Amd64, but not on Arm64 (i.e. Silicon Macs). I had assumed my patch would handle both cases, because I use the `HOMEBREW_PREFIX` environment variable. However, I wonder if `HOMEBREW_PREFIX` is not being set to `/opt/homebrew` in the Arm64 build.

I'm not sure of a good way to fix these issues.

I'm not familiar with FreeBSD, or its package manager/repositories. I wonder if FreeBSD requires some `zstd` depext?

alan-j-hu:
I wonder if the LLVM.18-shared package will build on Silicon Macs so long as the user properly sets the `HOMEBREW_PREFIX` environment variable prior to installing.

alan-j-hu:
I tried adding "zstd" as a depext to conf-llvm, but the FreeBSD CI still failed, so looks like that wasn't the issue.

mseri:
Let's leave it as is for now. This way homebrew and freebsd users will have a chance to try it out and see if they can help us with a fix. What do you think?

alan-j-hu:
Ok, sounds good to me.

mseri:
Thanks for such an amount of wark and your patience with all my requests 🙏

PR #27180, opened 2024-12-26
Fix an oversight in opam file for pa_ppx_regexp.0.03

chetmurthy:
the version constraint for pa_ppx was "= 0.15", but needed to be ">=".


chetmurthy:
ugh.  there's a mistake in the pa_ppx_regexp.0.03 release.  I'll do a 0.04 release and withdraw the 0.03 one.

PR #27100, opened 2024-12-13
Archiving opam-repository Phase 1 (unavailable and uninstallable packages)

hannesm:
~~Do not merge before Jan 1st 2025.~~ See https://github.com/ocaml/opam-repository-archive/pull/3

Opening here for completeness, CI shouldn't care.

PR #27208, opened 2025-01-05
result < 1.5: mark as available only with ocaml < 4.08

### LINKED ISSUE #24263: Old result packages shadows the `Result` modules for newer builds

hannesm:
Result is a compatibility package for OCaml versions that do not provide the Result module in the standard library (< 4.08). There's an issue with result < 1.5, namely that with OCaml compilers >= 4.08 it does lack the type equality which leads to compilation errors due to the Result.result (provided by the result package) not being compatible with the Result.t from the standard library.

Now, at [several](https://github.com/ocaml/opam-repository/pull/25603) [occasions](https://github.com/ocaml/opam-repository/pull/24868) there was an attempt to lower the lower bound - to avoid random packages failing in CI, and requiring the magic `conflicts: [ "result" {< "1.5"} ]` line in arbitrary opam files (see also [discussions](https://github.com/ocaml/opam-repository/issues/19880) and [more](https://github.com/ocaml/opam-repository/issues/24263)).

Time has moved on, we're now in the happy situation (thanks to https://github.com/ocaml/opam-repository/pull/25960 and https://github.com/ocaml/opam-repository/pull/27100) that we're able to remove packages. This is great.

Within our archival process, soon (Feb 1st) we'll move all our `< 4.08` packages to the archive. To avoid further burden on ocaml package developers, I propose (as discussed in opam-repository maintainers meetings) to lower the lower bound of earlier result packages - so we finally can close #24263.

Obviously since there has been some discussions and disapproval by @dra27 (and eventually others, asking for @avsm @kit-ty-kate opinions), my invitation is to discuss this PR - whether there are still concerns or the time is ready. From my discussions with the opam-repository maintainers, the timing is good.

### DIFF
--- a/packages/result/result.1.0/opam
+++ b/packages/result/result.1.0/opam
@@ -12,7 +12,7 @@ description: """
 Projects that want to use the new result type defined in OCaml >= 4.03
 while staying compatible with older version of OCaml should use the
 Result module defined in this library."""
-depends: ["ocaml" {< "4.14"}]
+depends: ["ocaml" {< "4.08"}]
 url {
   src: "https://github.com/janestreet/result/archive/1.0.tar.gz"
   checksum: [
--- a/packages/result/result.1.1/opam
+++ b/packages/result/result.1.1/opam
@@ -11,7 +11,7 @@ description: """
 Projects that want to use the new result type defined in OCaml >= 4.03
 while staying compatible with older version of OCaml should use the
 Result module defined in this library."""
-depends: ["ocaml" {< "4.14"}]
+depends: ["ocaml" {< "4.08"}]
 url {
   src: "https://github.com/janestreet/result/archive/1.1.tar.gz"
   checksum: [
--- a/packages/result/result.1.2/opam
+++ b/packages/result/result.1.2/opam
@@ -14,7 +14,7 @@ description: """
 Projects that want to use the new result type defined in OCaml >= 4.03
 while staying compatible with older version of OCaml should use the
 Result module defined in this library."""
-depends: ["ocaml" {< "4.14"}]
+depends: ["ocaml" {< "4.08"}]
 url {
   src: "https://github.com/janestreet/result/archive/1.2.tar.gz"
   checksum: [
--- a/packages/result/result.1.3/opam
+++ b/packages/result/result.1.3/opam
@@ -7,7 +7,7 @@ bug-reports: "https://github.com/janestreet/result/issues"
 license: "BSD-3-Clause"
 build: [["jbuilder" "build" "-p" name "-j" jobs]]
 depends: [
-  "ocaml" {< "4.14"}
+  "ocaml" {< "4.08"}
   "jbuilder" {>= "1.0+beta11"}
 ]
 synopsis: "Compatibility Result module"
--- a/packages/result/result.1.4/opam
+++ b/packages/result/result.1.4/opam
@@ -7,7 +7,7 @@ bug-reports: "https://github.com/janestreet/result/issues"
 license: "BSD-3-Clause"
 build: [["dune" "build" "-p" name "-j" jobs]]
 depends: [
-  "ocaml" {< "4.14"}
+  "ocaml" {< "4.08"}
   "dune" {>= "1.0"}
 ]
 synopsis: "Compatibility Result module"


kit-ty-kate:
With the archival process in mind i personally think this PR is totally fine and welcome. If this is merged i would also invite to also open a PR to remove the now unnecessary `conflicts: ["result" {< "1.5"}]` from the numerous packages that have it.

hannesm:
@kit-ty-kate that's a great idea, and once this PR is merged I'll follow up (it may only be safe when the result packages have been archived - otherwise on a e.g. 4.06 switch someone may install result 1.3 and get into trouble with some packages).

avsm:
Agreed with @kit-ty-kate! My only objection to it previously was the supported status of pre-4.08 versions. It'll also be really nice to remove all the conflict clauses on result.

dra27:
It's a data point rather than a veto, but as in https://github.com/ocaml/opam-repository/pull/24868#issuecomment-1838940115, there will still I think be packages which aren't (yet) archived but which become uninstallable on OCaml 4.08-4.13 as a result of this.

hannesm:
Right @dra27. To repeat your list -- here with the transitive closure of revdeps:
-  `alcotest.0.8.0`, `alcotest.0.8.1`, `alcotest.0.8.3`, `alcotest.0.8.4`, `alcotest.0.8.5` -- makes `irmin-test.2.0.0` unavailable, newer alcotest versions available
-  `cmdliner.1.0.0`, `cmdliner.1.0.1`, `cmdliner.1.0.2` - no further revdeps, newer releases available
-  `containers.2.6`, `containers.2.6.1`, `containers.2.7` - makes `smbc.0.5` and `smbc.0.6` uninstallable (there's `smbc.0.6.1`) -- also newer containers releases available
-  `datakit-client.0.12.0`, `datakit-client.0.12.2`, `datakit-client.0.12.3`, `datakit-client.1.0.0` - makes `datakit-bridge-github`, `datakit-bridge-local-git`, `datakit-ci`, `datakit-client-9p`, `datakit-github` uninstallable
-  `depyt.0.2.0` - no further revdeps, newer versions available
-  `irmin.1.3.2`, `irmin.1.4.0` - plays a role in `datakit`, also in `ezirmin` -- newer irmin versions available
-  `logs.0.6.2` - makes `logs-async.1.0` uninstallable (there's newer versions), also newer logs versions available
-  `mirage-block-lwt.1.0.0`, `mirage-block-lwt.1.1.0` - no further revdeps, newer versions available
-  `ocaml-version.2.3.0` - no further revdeps, newer versions available
-  `promise.1.0.2` - no further revdeps, newer versions available
-  `tls.0.9.2`, `tls.0.9.3`, `tls.0.10.1` - no further revdeps, newer versions available
-  `tsdl.0.9.1`, `tsdl.0.9.2`, `tsdl.0.9.3`, `tsdl.0.9.4`, `tsdl.0.9.5`, `tsdl.0.9.6` - makes `bogue.20190717`, `bogue.20190920`, and `bogue.20200630` uninstallable (there are newer bogue versions), also `tsdl-mixer.0.2` (newer version available) -- there are newer tsdl versions available
-  `tsdl-image.0.2.0` - no further revdeps, newer versions available
-  `x509.0.6.2` - no further revdeps, newer versions available
-  `zed.2.0`, `zed.2.0.1`, `zed.2.0.2`, `zed.2.0.3`, `zed.2.0.4`, `zed.2.0.5` - no further revdeps, newer versions available

So, below the line I only see datakit as an issue -- and am wondering whether this is an actively used project? There's not much activity on https://github.com/moby/datakit -- is this still used?

hannesm:
Oops, there's maybe some more impact by `containers`, with:
- `calculon.0.1` and `calculon.0.2` (newer version exists)
- `lilis.0.2.1` no newer version available
- `opam-query.1.3` no newer version available

I'm unaware of the usage and impact of lilis and opam-query. They both don't have any revers dependencies, I am not sure whether they are used.

hannesm:
paging @avsm @djs55 @samoht @MagnusS @talex5 as datakit authors
also @whitequark for opam-query
and @Drup for lilis

the question is: are these three packages actively used and maintained, so would it be really sad to archive them (as a collateral damage of result < 1.5) or not. thanks.

djs55:
Regarding datakit: I've not used it for several years. We've not added a commit for 6 years. I think it's ok to archive.

whitequark:
`opam-query` was intended as a CLI tool. It is sad to see that a tool is broken in that way but I do not object to it.

raphael-proust:
The change looks fine to me and I think it'll help have simpler packages with fewer arcane you-just-need-to-know-about-it recipes for how to make a good package.

Thanks for the work @hannesm!

PR #27211, opened 2025-01-05
functoria/functoria-runtime: mark deprecated

hannesm:
this is part of mirage since 4.5.0.


shonfeder (approved):
Given how focused these changes are, I think all the failing CI test must be unrelated. Thanks for the clean up!

PR #27212, opened 2025-01-05
tar-format: mark as deprecated

hannesm:
This package has been renamed to "tar".


shonfeder (approved):
Thanks!

CI errors can't be related.

PR #27217, opened 2025-01-06
Add x-maintenance-intent for Irmin packages

art-w:
The deprecated `irmin-mem` and `irmin-unix` packages were moved to the main `irmin` package some time ago (under `irmin.mem` and `irmin.unix`).


hannesm:
Thanks! :)

for `imin-mem` and `irmin-unix` a `flags: deprecated` in all versions would be fine as well, but I the `x-maintenance-intent: [ "(none)" ]` is good as well.

There's no need for specifying the x-reason-for-archival, this will be added when the package(s) are moved to opam-repository-archive.

packages/irmin-mem/irmin-mem.2.3.0/opam
```diff
@@ -35,3 +35,5 @@ url {
     "sha512=ff527303850c2e36853ae0857f2ecbc22b5096f87197521dd049f80a234228c74f840aadef6d7c826acceda7b0a49e1dc2b5c105bb65d7c9551f0f8f377b2fc6"
   ]
 }
+x-reason-for-archiving: [ "maintenance-intent" ]
```
hannesm, line 38:
```suggestion
```

packages/irmin-tezos-utils/irmin-tezos-utils.3.6.1/opam
```diff
@@ -43,3 +43,5 @@ url {
   ]
 }
 x-commit-hash: "0d241e62f69036e4b26c6bddbab49826051ab3af"
+x-reason-for-archiving: [ "maintenance-intent" ]
```
hannesm, line 46:
```suggestion
```

art-w:
Ok sure! Thanks for the clarifications :)

mseri:
Thanks

hannesm:
I don't think the last commit (adding sources to irmin-unix) is necessary -- since back then irmin used optional dependencies, and the irmin-unix was just a package depending on irmin and the optional dependencies to get irmin.unix built -- i.e. irmin-unix didn't install anything (and there's no ocamlfind package for irmin-unix) -- so no sources are necessary.

art-w:
Yeah hm you're right I removed that last commit.

shonfeder (approved):
Thanks!

PR #27218, opened 2025-01-06
[new release] starred_ml (0.0.6)

paulosuzart:
CHANGES:

- Adds a `language_slug` attribute to the `item` (repository being rendered) with the slug of the language.
- Languages sent to `jingoo` template now are the slug version of the language.


shonfeder:
Thanks for looking at the CI results :) That is an upstream failure on windows, and needs to addressed on the re package, so nothing to block your publication here. Thanks!

PR #27227, opened 2025-01-07
Flag every versions of the ocaml package that haven't been released yet with avoid-version

kit-ty-kate:
https://github.com/ocaml/opam/pull/6358 would display more clearly which packages are tagged with avoid-version/deprecated. However in the case of `ocaml` where we would expect `5.4.0` to be shown in the same way to avoid confusion as to which version is the latest, it is more complicated due to the fact that detecting if such a package needs to be displayed differently would need a request from the solver for each version which seems too costly.

Thus i'm proposing to tag the `ocaml` packages that are still unreleased or in their pre-release phase with `avoid-version` and to untag them at the same time as the final release (in this case 5.4.0).

cc @Octachron does that seem ok to you? If so this step would be to be added to the release documentation.

Not to be merged before https://github.com/ocaml/opam/pull/6358 has been discussed and merged.

### DIFF
--- a/packages/ocaml/ocaml.4.02.4/opam
+++ b/packages/ocaml/ocaml.4.02.4/opam
@@ -28,4 +28,4 @@ authors: [
   "Didier Rémy"
   "Jérôme Vouillon"
 ]
-flags: conf
+flags: [conf avoid-version]
--- a/packages/ocaml/ocaml.4.03.1/opam
+++ b/packages/ocaml/ocaml.4.03.1/opam
@@ -28,4 +28,4 @@ authors: [
   "Didier Rémy"
   "Jérôme Vouillon"
 ]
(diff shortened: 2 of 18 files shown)


Octachron:
That seems very-reasonable to me, I imagine that we should also flags 5.3.1?

kit-ty-kate:
> I imagine that we should also flags 5.3.1?

I forgot about it but yes indeed

kit-ty-kate:
https://github.com/ocaml/opam/pull/6358 has just been merged so this is ready to review. I also took the time to flag previous unreleased versions of ocaml (e.g. 5.0.1, 4.02.4, …)

mseri:
Thanks a lot! We need to keep this in mind for future compiler releases

PR #27228, opened 2025-01-07
Mark all non-latest opam-* pre-releases as deprecated

kit-ty-kate:
As asked in https://github.com/ocaml/opam/issues/6348
Although i'm not sure adding `flags: deprecated` is the best way to do this but we can talk about it later


packages/opam-client/opam-client.2.2.0~alpha/opam
```diff
@@ -46,5 +46,4 @@ url {
     "sha512=b7899244ab620213f2997409c9d8ac63e1a7ba99ff36abdd0acf41ae545ca7e0534e5e2f6c01a9700c10abf2df4f07944cc64fb5dd315a566f6c5b6b19464729"
   ]
 }
-available: opam-version >= "2.1.0"
-flags: avoid-version
+flags: deprecated
```
avsm, line 49: Won't this break opam 2.1 users if we remove the opam-version constraint as well?

kit-ty-kate: unknown flags are ignored so on opam 2.0 and 2.1 these packages will just not have any special behaviour.
In practice it shouldn't matter anymore as more recent versions are available. It was critical to have it at the time of release because it would've been the latest version and be picked up automatically by the solver, but this is not the case anymore

mseri:
Now we can use `x-maintained: false` to achieve the same and would avoid the full package deprecation warning

shonfeder:
IIUC, this will need to be recreated if we want to move forward?

hannesm:
I took this over, and instead of deprecated, marked these with `x-maintained: false`. I also rebased on top of master.

mseri:
Thanks!

PR #27230, opened 2025-01-08
Add x-maintenance-intent to ctypes.

yallop:
Closes yallop/ocaml-ctypes#791.

Add `x-maintenance-intent` to the `ctypes` package.

I'm starting with a conservative policy that should make it possible to archive around 55% of the current `ctypes` versions.  I may switch to a more aggressive policy in the future.


mseri:
Thanks!

PR #27231, opened 2025-01-08
release of OCaml 5.3.0

Octachron:
This PR adds the opam package for the newly release OCaml 5.3.0 compiler:
More precisely, it adds  the core compiler package

-  ocaml-compiler.5.3.0

and its two variants

- ocaml-base-compiler.5.3.0
- ocaml-variants.5.3.0+trunk

completed by the system version

- ocaml-system.5.3.0 (where I have removed the non-windows arch package)

Simultaneously, this PR moves the dev package 

- `ocaml-variants.5.3.0+trunk` to `ocaml-variants.5.3.1+trunk`

 and adds the necessary metapackage

- ocaml.5.3.1 

This time, I have included the two related packages

- ocaml-src.5.3.0
- ocaml-manual.5.3


Octachron:
cc @mseri @shonfeder @raphael-proust

(I am not sure what it the best way to signal new release of the compiler?)

mseri:
Seems to include the various changes for host- and arch- packages for windows and to unbreak lock files on other systems that we discussed recently. Ping @dra27 to double check

packages/ocaml-system/ocaml-system.5.3.0/opam
```diff
@@ -0,0 +1,54 @@
...
+  "base-bigarray" {post}
+  "base-threads" {post}
+  "base-domains" {post}
+  "base-nnp" {post}
+  "base-effects" {post}
+
+  # System (Windows-only at present)
+  "host-system-mingw" {?sys-ocaml-arch & sys-ocaml-libc = "msvc" & sys-ocaml-cc = "cc" & post}
+  "host-system-msvc" {?sys-ocaml-arch & sys-ocaml-cc = "msvc" & post}
+  "host-system-other" {?sys-ocaml-arch & sys-ocaml-libc != "msvc" & post}
```
dra27, line 27: From https://github.com/dra27/opam-repository/pull/20:
```suggestion
# Architecture (Windows-only at present)
  "host-arch-x86_32" {os = "win32" & ?sys-ocaml-arch & sys-ocaml-arch = "i686" & post}
  "host-arch-x86_64" {os = "win32" & ?sys-ocaml-arch & sys-ocaml-arch = "x86_64" & post}
  "host-arch-unknown" {os = "win32" & (!(?sys-ocaml-arch) |
                       sys-ocaml-arch != "i686" &
                       sys-ocaml-arch != "x86_64") & post}

  # System (Windows-only at present)
  "host-system-mingw" {os = "win32" & ?sys-ocaml-arch & sys-ocaml-libc = "msvc" & sys-ocaml-cc = "cc" & post}
  "host-system-msvc" {os = "win32" & ?sys-ocaml-arch & sys-ocaml-cc = "msvc" & post}
  "host-system-other" {os = "win32" & ?sys-ocaml-arch & sys-ocaml-libc != "msvc" & post}
```

dra27 (commented):
I think this is correct, because opam has been fixed on macOS et al, but we definitely rename ocaml-variants.5.3.0+trunk to ocaml-variants.5.3.1+trunk (rather than leaving an `available: false` ocaml-variants.5.3.0+trunk), now?

The changes in ocaml-compiler / ocaml-base-compiler / ocaml-variants are as already tested in #27145 - minor tweak to ocaml-system.5.3.0, just to match the proposal in https://github.com/dra27/opam-repository/pull/20/ (keeps system compilers consistent with from-source compilers on Windows).

Octachron:
Concerning the `+trunk` package, I mostly followed the change in #27100 which removed all not-available opam files. It might makes sense to mark the file as unavailable temporarily?

raphael-proust:
> It might makes sense to mark the file as unavailable temporarily?

the only difference it'd make is that the packages would end up in the archive at the next clean-up

the purpose of the archive is to allow people to restore their setup to a previous state (e.g., they had a binary compiling at some point and they want to have access to the packages that existed at that point)

I don't think we need that for packages which have never eneterd the repo at all, so I suggest we don't add those packages with the false availibility

Octachron:
Is there anything blocking the merge?

raphael-proust:
:camel: :bangbang:

Octachron:
Thanks !

PR #27238, opened 2025-01-09
[new release] lsp (3 packages) (1.21.0)

voodoos:
LSP protocol implementation in OCaml

## Features

- Add a new server option `standardHover`, that can be used by clients to
  disable the default hover provider.  When `standardHover = false`
  `textDocument/hover` requests always returns with empty result. (ocaml/ocaml-lsp#1416)


avsm:

PR #27239, opened 2025-01-10
utop 2.15.0: switch to the archive with the substituted version number

Octachron:
This small PR points the opam package for utop 2.15.0 to the updated release archive for utop 2.15.0 in which the utop version number has been correctly substituted.

This avoids leaving %%VERSION%% in the utop banner

> Welcome to utop version %%VERSION%% (using OCaml version 5.3.0)!"

### DIFF
--- a/packages/utop/utop.2.15.0-1/opam
+++ b/packages/utop/utop.2.15.0-1/opam
@@ -1,0 +1,50 @@
+opam-version: "2.0"
+synopsis: "Universal toplevel for OCaml"
+description:
+  "utop is an improved toplevel (i.e., Read-Eval-Print Loop or REPL) for OCaml. It can run in a terminal or in Emacs. It supports line edition, history, real-time and context sensitive completion, colors, and more. It integrates with the Tuareg mode in Emacs."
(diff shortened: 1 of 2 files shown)


shonfeder:
I'm not sure if it fits opam policy to change the sources for this kind of fix. Isn't this essentially a small patch fix, and so deserving of an patch version increment?

I'll leave that to more experienced maintainers, cc @mseri and @raphael-proust.

Octachron:
This feels slightly more like a packaging change than a version change from my point of view. In particular, this changes of archives doesn't correspond to a change in the utop repository.

mseri:
Lint is fine and the build had already run, I am merging

Octachron:
Thanks a lot ! For record's sake, do I understand correctly that the opam policy is to use `%{internal_version}-%{package_version}` scheme for updating package versions?

mseri:
I am not sure I understand the name of the variables. Usually what we do for a patch release is to publish the same version dash an incremental number starting from 1 (integer increments for each eventual new update)

PR #27256, opened 2025-01-13
Mark all jbuilder packages as depracted

maiste:
Following the discussion on [this Dune issue](https://github.com/ocaml/dune/issues/11272), this PR is the first step to make clearer the intention about the maintenance policy of Dune.

This PR marks all the `jbuilder` packages as `deprecated` as Dune should be used instead.

### DIFF
--- a/packages/jbuilder/jbuilder.1.0+beta1/opam
+++ b/packages/jbuilder/jbuilder.1.0+beta1/opam
@@ -9,6 +9,8 @@ build: [
   ["ocaml" "bootstrap.ml"]
   ["./boot.exe" "-j" "1"]
 ]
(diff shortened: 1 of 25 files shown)


hannesm:
Thanks a lot, after further discussion, would you instead of marking everything as deprecated put a `x-maintenance-intent: [ "(none)" ]` into the latest release? Thanks a lot!

maiste:
I can mark the latest as `x-maintenance-intent: [ "(none)" ]`. However, do you want me to keep the deprecated flags everywhere or to remove it?

mseri:
I think deprecated for jbuilder is fine, sice it has been renamed to dune and is no-longer supported

PR #27261, opened 2025-01-13
parany-14.0.1 opam file: added x-maintenance-intent

UnixJunkie:
added x-maintenance-intent

I think @hannesm sent me an e-mail about this


avsm:
Thanks!

PR #27265, opened 2025-01-14
[new release] starred_ml (0.0.7)

paulosuzart:
CHANGES:

- IMPORTANT: now use `starred_ml render` for rendering the template.
- Removed `slug` dependency that was actually not enough for the intention.
- Encourages raw jingoo filters for encoding urls. As in `item.language | urlencode`.
- The `dafault.jingoo` template now takes a the github repositories (the `starred list` type) and uses `groupby` Jingoo's feature to print each repository under the corresponding language.


mseri:
Thanks

PR #27272, opened 2025-01-15
Add the indexer as a post-dependency to ocaml-lsp-server

voodoos:
Latest Merlin and OCaml-LSP enable project-wide renaming and it requires an additional binary to be installed.


mseri:
Thanks

PR #27275, opened 2025-01-16
Add x-maintenance-intent field for agrep, camlidl, camlp-streams, camlzip, cryptokit, dbm, graphics, num, mpi, pringo, zarith

xavierleroy:


packages/agrep/agrep.1.0/opam
```diff
@@ -5,6 +5,7 @@ homepage: "https://github.com/xavierleroy/ocamlagrep/"
 bug-reports: "https://github.com/xavierleroy/ocamlagrep/issues"
 dev-repo: "git+https://github.com/xavierleroy/ocamlagrep.git"
 license: "LGPL-2.0-or-later"
+x-maintenance-intent: [("latest")]
```
hannesm, line 8:
```suggestion
x-maintenance-intent: ["(latest)"]
```

Sorry if this is documented differently somewhere. The quotes should be around the `(latest)`.

xavierleroy: That was cut and pasted from one of your posts on discuss.ocaml.org.  I suggest you edit your posts ASAP.

hannesm: Yes, thanks @xavierleroy. I already edited the post on discuss.

xavierleroy:
Concerning the CI failure on the num package, see https://github.com/ocaml/num/issues/44 .

mseri:
The dune lint failure is a false positive, don't worry about that. I'll bring it up at the next meeting to see if we can do better. It is there since in the past we had many instances of packages forgetting to add dune to their dependency even though it was used for the build

PR #27279, opened 2025-01-16
minor updates for the archiving policy as discussed today in the meeting

hannesm:
introduction of x-maintained field

Once merged, I plan to publish a discuss post (that will hopefully clarify some bits) -- suggestions / comments welcome:

## x-maintenance-intent

We've had some further discussions on Phase 3 and the semantics of the
`x-maintenance-intent` field.

### Goal

Our aim is to be not disruptive for the common OCaml programmer or user. The
opam-repository supports (from February 1st on) OCaml 4.08 and greater. This
means that if you install OCaml 4.08 you should be able to install all the
packages that have ever been released with 4.08 support.

This means the revised semantics of "(latest)" is "the latest version of this
package, so that every supported OCaml version will have an installation
candidate".

#### Example

Let me give you an example, consider the package "basic" which exists in three
versions:
basic.1.0.0 with the dependency "ocaml" {>= "4.05" & < "5"}
basic.1.0.1 with the dependency "ocaml" {>= "4.08" & < "5"}
basic.2.0.0 with the dependency "ocaml" {>= "4.14" & < "5"}

Here, if the `x-maintenance-intent: [ "(latest)" ]` is present, we will only
(try to) archive basic.1.0.0 -- since 1.0.1 is needed for OCaml 4.08 .. 4.13.
I hope this is clear.

### Default value

The default value of `x-maintenance-intent` will for now be `"(any)"` - so all
versions are kept. In the future, we may change this default to `"(latest)"`,
but will announce this ahead of the change with plenty of time.

This default value is agreed on by the non-disruptive agreement to cause the
least trouble.

## x-maintained

In addition to the `x-maintenance-intent` - which covers the semantics of all
versions of an opam package, we support another field, `x-maintained: BOOL`.
This is an overwrite for a specific opam package version, and allows to declare
whether it is maintained or not.

It is useful in the setting where you've lots of pre-releases that are no
longer maintained and you like to state this without writing a global intent
for the opam package (e.g. for the OCaml compiler packages, the alpha, beta, and
rc versions). Here, `x-maintained: false` is a nice setting. NB: earlier we
proposed `flags: deprecated` - but we stay away from the flags, since there may
be packages that are deprecated but still maintained (opam prints a warning if
you install a package with the deprecated flag).

If you have a private project and depend on a specific version of an opam
package, you can as well PR the `x-maintained: true` field for that opam file
(please specify when, who, and why). This will ensure that this opam file stays
in the opam repository.

## Phase 3

In Phase 3, we will consider all packages marked with `x-maintenance-intent`
(the versions not matching the intent) and `x-maintained: false` to be archived.

We plan to ensure that (a) all supported OCaml versions will retain an
installation candidate (b) all reverse dependencies will still be installable.
As a note, if you have an availability condition (some version will only work
on some OS), we won't take that into consideration -- you will need to specify
the `x-maintenance-intent` to cover your versions.

Our plan is to publish the list of packages to be archived by February 15th
on this discourse. It is likely we'll have candidate lists PRed to the
[opam-repository-archive](https://github.com/ocaml/opam-repository-archive)
earlier. We have lots of ideas and plans for CI systems to give feedback which
opam versions are falling into the maintenance intent when you open a PR to
the opam-repository (but we're not there yet).

## Future

As noted above, the default value of `x-maintenance-intent` may change in time.
If this is decided, we will announce this with plenty of time before.

Also, at some point in the future we will bump the OCaml lower bound (from
February 1st it is 4.08).

## Action

For the smooth shrinking of the opam-repository, please don't hesitate to fill
in your x-maintenance-intent (especially "(none)" and "(latest)" are fine
and safe choices).

If you want to contribute more, the opam-repository needs help for triaging and
merging PRs - why not become a maintainer? See the old but still valid
['call for new opam-repository maintainers'](https://discuss.ocaml.org/t/call-for-new-opam-repository-maintainers/12041)
if you're interested.

### DIFF
--- a/governance/policies/archiving.md
+++ b/governance/policies/archiving.md
@@ -4,7 +4,7 @@
 
 - The primary opam repository, (referred to here as the "primary repo") is located at [ocaml/opam-repository](https://github.com/ocaml/opam-repository). The primary repo is curated to ensure that compatible packages are co-installable on as many supported platforms as possible, and it is the default package repository.
 - "Primary repo criteria" refers to the criteria used to decide whether a version of a package is suitable for continued inclusion in the primary repo.
-- The archive opam repository, (referred to here as the "archive repo")" is located at [ocaml/opam-repository-archive](https://github.com/ocaml/opam-repository-archive) and records packages that were once in the primary repo, but no longer meet the primary repo criteria.
+- The archive opam repository, (referred to here as the "archive repo") is located at [ocaml/opam-repository-archive](https://github.com/ocaml/opam-repository-archive) and records packages that were once in the primary repo, but no longer meet the primary repo criteria.
 - "Supported platforms" are those listed under [Tier 1 and Tier 2 by the OCaml compiler](https://github.com/ocaml/ocaml?tab=readme-ov-file#overview)
 - A package's "maximum compiler version" is the upper bound of the OCaml compiler versions supported by a package. Support is derived either based on explicit version bounds set on the `ocaml` dependency in a package's dependency cone, or implicitly based on failed installs detected through our CI and health check processes.
 - The "compiler cutoff threshold" is an OCaml compiler versions stipulated by the opam repository maintainers. It sets a lower bound on the compiler versions supported by the primary repo. 
@@ -15,10 +15,10 @@
 
 ### Compiler cutoff threshold
 
-The current compiler cutoff threshold is `4.08`.
+The current (2025-02-01) compiler cutoff threshold is `4.08`.
 
 This threshold is subject to change by the opam repo maintainers.
-The threshold is based on the oldest ocaml compiler version available 
+The threshold is based on the oldest ocaml compiler version available
 in the maintained[^1] distributions. It determines the minimum
 compiler version used in tests for the opam-repository CI[^2].
 
@@ -34,7 +34,7 @@ A version of a package may be published on the primary repo, and will continue t
 3. The package version falls within a package's maintenance intent, or is a dependency of a package satisfying the primary repo criteria.
 4. The package version is installable on at least one of the supported platforms. (Note that it is not required that CI tests are passing, since working installation may require manual system configuration.)
 
-Note that this property is transitive along a package's dependency tree: if a package satisfies the primary repo criteria, then its dependencies do as well. 
+Note that this property is transitive along a package's dependency tree: if a package satisfies the primary repo criteria, then its dependencies do as well.
 
 ### Periodic pruning process
 
@@ -52,7 +52,7 @@ At regular intervals, no less than every six months, the opam repo maintainers w
 
 When it has been decided that a set of package versions (aka "versions") should be archived, archiving will proceed according to the following process:
 
-- A PR will be made to add the archived versions to the archive repo. 
+- A PR will be made to add the archived versions to the archive repo.
     - The opam file of each version will be extended as follows:
         - Any dependencies without upper bounds will have upper bounds added, recording the latest available version of the dependency in the primary repo at the time of archiving.
         -  The field `x-reason-for-archiving` will be added.
@@ -88,6 +88,7 @@ When it has been decided that a set of package versions (aka "versions") should 
     - Meaning:
         - Expresses the declared intent of the maintainers to maintain only certain versions ranges of a package.
         - The value of the `x-maintenance-intent` on the latest published package will precedence.
+    - Default value: ["(any)"] on 2025-01-17, please note that this default may change
     - Examples:
         - `["(latest)"]` the maintainer will only maintain the latest version
         - `["(latest)" "(latest-1)"]` the maintainer will only maintain the latest `X.Y.Z` version and `(X-1).Y.Z`
@@ -98,6 +99,16 @@ When it has been decided that a set of package versions (aka "versions") should 
         - `["(none)"]` the maintainer will not maintain any version
         - `["1.3"]` the maintainer will maintain the latest  version of "1.3.Z"
         - `["2.(latest)"]` the maintainer will maintain the latest minor version specifically of version "2" of the package
+- `x-maintained`:
+    - Allowed values: `true` and `false`
+    - Meaning:
+        - Expresses that this opam package version is maintained (if `true`) or not (if `false`).
+        - Overwrites the `x-maintenance-intent` field
+        - Useful to abandon pre-releases
+        - Also for packages that are used in private opam-repositories or developments
+    - Examples:
+        - `true # used by @bactrian, added on 2025-01-24` where @bactrian wants to keep this package in the opam-repository
+        - `false` if this meets the `x-maintenance-intent`, but is actually not maintained
 
 ## References
 


governance/policies/archiving.md
```diff
@@ -88,6 +88,7 @@ When it has been decided that a set of package versions (aka "versions") should
     - Meaning:
         - Expresses the declared intent of the maintainers to maintain only certain versions ranges of a package.
         - The value of the `x-maintenance-intent` on the latest published package will precedence.
+    - Default value: ["(any)"] on 2025-01-17, please note that this default may change
```
mseri, line 91:
```suggestion
    - Default value: ["(any)"] on 2025-01-17, please note that this default is expected to change in the future
```
I would make it even more explicit

governance/policies/archiving.md
```diff
@@ -98,6 +99,16 @@ When it has been decided that a set of package versions (aka "versions") should
         - `["(none)"]` the maintainer will not maintain any version
         - `["1.3"]` the maintainer will maintain the latest  version of "1.3.Z"
         - `["2.(latest)"]` the maintainer will maintain the latest minor version specifically of version "2" of the package
+- `x-maintained`:
+    - Allowed values: `true` and `false`
+    - Meaning:
+        - Expresses that this opam package version is maintained (if `true`) or not (if `false`).
+        - Overwrites the `x-maintenance-intent` field
+        - Useful to abandon pre-releases
+        - Also for packages that are used in private opam-repositories or developments
```
mseri, line 108:
```suggestion
        - Also for packages that are used in private opam-repositories or developments and that a maintainer would like to keep in the repository independently of the global package intent
```
I think this is not yet clear enough, maybe this edit? What do you think?

governance/policies/archiving.md
```diff
@@ -88,6 +88,7 @@ When it has been decided that a set of package versions (aka "versions") should
     - Meaning:
         - Expresses the declared intent of the maintainers to maintain only certain versions ranges of a package.
         - The value of the `x-maintenance-intent` on the latest published package will precedence.
+    - Default value: ["(any)"] on 2025-01-17, please note that this default may change
     - Examples:
         - `["(latest)"]` the maintainer will only maintain the latest version
```
mseri, line 93:
```suggestion
        - `["(latest)"]` the maintainer will only maintain the latest version[^caveat]
        [^caveat]: note that this will retain the latest versions of this package so that every supported OCaml version will have an installation candidate.
```
I think we should make this semantic explicit

mseri:
Thanks a lot for the update. I only have some very minor comments

mseri (approved):

governance/policies/archiving.md
```diff
@@ -15,10 +15,10 @@
 
 ### Compiler cutoff threshold
 
-The current compiler cutoff threshold is `4.08`.
+The current (2025-02-01) compiler cutoff threshold is `4.08`.
```
shonfeder, line 18: I'm not sure how this date helps. AFAIU, whatever the current threshold is should be stated here, and when it is updates, this should be updated. If someone wants to see when this was written, they can look at git blame.

governance/policies/archiving.md
```diff
@@ -98,6 +100,16 @@ When it has been decided that a set of package versions (aka "versions") should
         - `["(none)"]` the maintainer will not maintain any version
         - `["1.3"]` the maintainer will maintain the latest  version of "1.3.Z"
         - `["2.(latest)"]` the maintainer will maintain the latest minor version specifically of version "2" of the package
+- `x-maintained`:
+    - Allowed values: `true` and `false`
+    - Meaning:
+        - Expresses that this opam package version is maintained (if `true`) or not (if `false`).
+        - Overwrites the `x-maintenance-intent` field
```
shonfeder, line 107:
```suggestion
        - Overrides the `x-maintenance-intent` field
```

governance/policies/archiving.md
```diff
@@ -98,6 +100,16 @@ When it has been decided that a set of package versions (aka "versions") should
...
+        - Useful to abandon pre-releases
```
shonfeder, line 108: It is not clear to me how this works or what this means. Since the policy document should stand alone, I think expanding this to explain is a good idea.

governance/policies/archiving.md
```diff
@@ -98,6 +100,16 @@ When it has been decided that a set of package versions (aka "versions") should
...
+        - Also for packages that are used in private opam-repositories or developments and that a maintainer would like to keep in the repository independently of the global package intent
```
shonfeder, line 109: I am unclear here too. I would not think the opam repository policies would have any bearing on private repositories.

governance/policies/archiving.md
```diff
@@ -98,6 +100,16 @@ When it has been decided that a set of package versions (aka "versions") should
...
+    - Examples:
+        - `true # used by @bactrian, added on 2025-01-24` where @bactrian wants to keep this package in the opam-repository
+        - `false` if this meets the `x-maintenance-intent`, but is actually not maintained
```
shonfeder, line 112: This seems confusing to me. Why no just update the intent to be accurate, rather than introducing a new field that can conflict with the existing one and then having it override?

governance/policies/archiving.md
```diff
@@ -88,8 +88,10 @@ When it has been decided that a set of package versions (aka "versions") should
     - Meaning:
         - Expresses the declared intent of the maintainers to maintain only certain versions ranges of a package.
         - The value of the `x-maintenance-intent` on the latest published package will precedence.
+    - Default value: ["(any)"] on 2025-01-17, please note that this default is expected to change in the future
     - Examples:
-        - `["(latest)"]` the maintainer will only maintain the latest version
+        - `["(latest)"]` the maintainer will only maintain the latest version[^caveat]
+        [^caveat]: note that this will retain the latest versions of this package so that every supported OCaml version will have an installation candidate.
```
shonfeder, line 94: This caveat is not very clear to me. I am also worried that introducing footnotes make the document harder to read. IMO, we should put all the needed details in the policy directly, without introducing marginal qualifications.

Regarding the content of the caveat: AFAIU the policy, the maintenance intent does not dictate what is "retained". This is determined based on satisfying the inclusion criteria. The intent is way to tell us whether we can just remove a package that fails to satisfy the criteria, or whether we try to contact a maintainer. My understanding of the policy as written is: If a package is no repaired to satisfy the inclusion criteria and the maintainer does not respond within the stated window, the package should be archived no matter what the stated intent was. Conversely, if a package version continues to satisfy the inclusion criteria, it is retained even if it is out-side of the maintenance intent. Given that understanding, I do not understand what this note qualifying retainment relative to ocaml versions adds.

shonfeder: I see that this is a suggestion from @mseri, so tagging.

governance/policies/archiving.md
hannesm: I'm sure everyone has opinions, that's why we are meeting to discuss the consensus. Sorry that you couldn't make it last week.

The short story is: there's no "conflict". As example, take the OCaml compiler releases. I don't think anyone will write down an x-maintenance-intent for them, but nevertheless the alpha/beta/rc of old versions can safely be archived.

governance/policies/archiving.md
hannesm: I struggle what you mean. The whole purpose of archiving things is to reduce the load on opam clients.

Now, if there's package a in version 1.2.3 and 1.2.4 and 1.2.5 with the dependencies all the same, of course we would like to archive 1.2.3 and 1.2.4.

So, the whole addition here is to clarify that a maintenance-intent of latest means that for each supported OCaml compiler, the latest version of that package will be retained. And not only the latest and greatest (in case it requires a newer than the cut-off OCaml compiler). Hope that makes it clearer.

governance/policies/archiving.md
hannesm: done in df67251905, thanks

governance/policies/archiving.md
hannesm: I clarified in df67251905, hope that is understandable.

governance/policies/archiving.md
hannesm: I amended the text in df67251905, is this understandable for you now?

governance/policies/archiving.md
shonfeder: My worry is that it will lead people to think that this has become "out of date" since this date will probably only change very infrequently. E.g., in 2027, this may still be the current cutoff threshold, but someone is likely to read it and think, "huh, current as of 2025, huh? Well that is quite old, this must be out of date." In other words, I don't think this date answers a questions anyone will have ("current just means current") but I think it is likely to raise irrelevant questions.

That said, it's not a big deal, so if you are sure this is an improvement, I am not really  worried about. If we find it to cause confusion later, we can deal with it at that time.

governance/policies/archiving.md
shonfeder: Yes this helps me a lot. Thanks!

governance/policies/archiving.md
shonfeder: I sorry I missed the meeting! I'm glad we also have a review process in addition to sync discussions tho :)

More generally, I am worried about the apparent growing complexity of this maintenance data. The more we add new fields and elaborate the bespoke syntax, the more worried I am that we are introducing something that is too esoteric to be that helpful. E.g., if we think we need this `x-maintained` field, maybe we don't need `x-maintenance-intent` at all? Wouldn't it simplify things a lot to just have a boolean flag on each version stating whether or not it is maintained? Then we don't have to worry about inventing a new specification format for all kinds of versions formats :)

governance/policies/archiving.md
shonfeder: Yes, I am reconsidering too much of the policy here. I retract this comment. We can talk it out at a meeting maybe if it's still worth discussing.

governance/policies/archiving.md
mseri: X-maintenance-intent is global, affects all releases of a package. This allows to override o specify the policy on a specific version of a release. It resolves (manually) all the corner cases we had do far and gives us an alternative to the deprecated flag, which is also package-wide

shonfeder: I believe I understand the intended semantics. But since we've decided a boolean flag is necessary, and now considering how that interacts with, and compares against, the prior `x-maintenance-intent`, I have become very uncertain whether the latter is actually a good solution for the problem we are trying to solve.

I know that this is a bit late in the process to raise this concern, but I think it is better to fix things a little late than it is to stay committed to something that adds complexity and may cause more problems down the line. Here is my current thinking:

Consider: with `X-maintenance-intent`, to determine if any given package version is actually ("intended") to be maintained, you have to:

- find the latest version of that package
- parse out it's `X-maintenance-intent` field into a (from a not fully determined syntax)
- then run the (not fully defined algorithm) to determine if the package version of interest happens to meet the criteria

This requires various forms of knew knowledge and functionality to work correctly, including:

- package publishers to learn 2 new opam fields
- package publishers to learn a novel (albeit simple) DSL for package versions
- us to fully specify a syntax and semantics for this DSL
- the CI and archiving tooling to do back and forth comparing different versions to reason out what is supported.

By contrast, suppose we just have `x-maintained`. To determine if a given package version is maintained, you just:

- check the boolean value of `x-maintained` on the package version

All that is needed to support this is:

- package authors learn one new opam field

So the latter seems like its just a much simpler and more elegant solution to our need to identify whether a package version is maintained. What are the down sides?

The only downside I can think of is that for some packages with a lot of versions, authors may need to make quite a few small updates. But the semantics and logistics of such updates are trivial, and we can very easily offer tooling (e.v., via `opam admin`) to make this easier.

To sum up, thru this discussion, and the surrounding discussions abound the `x-maintenance-intent`, I've come to think that the latter causes more problems that it solves, and that this boolean flag `x-maintained` is all we really need for our purposes. 

I think a principle we can appeal to is that meta-data formats should be as simple as possible, and the corollary that it is better to have simple, unambiguous data formats (with the business logic in tooling to generate and manipulate the data) than it is to enrich the data format with various ad hoc domain-specific logic. I suspect it is much easier for us to write and maintain a tool that can insert a boolean flag into a bunch of opam files, than it will be for our the whole opam-repository socio-technical system to figure out how to manage `x-maintenance-intent`.

WDYT?

governance/policies/archiving.md
mseri: I agree with @hannesm mostly because I am lazy and would never do the git blame 😅 

We have to update the threshold every now and then anyway, at that point we can update also the date. And if we see that it gets stale, we can always drop it later

governance/policies/archiving.md
mseri: I am still not convinced. We aim at having a x-maintained-intent with a default that works for most packages and allows us to do a cleanup. I see this is a path of least resistance and least effort.
A tag that needs to be edited continuously would be a massive pain to maintain for us and the maintainers imo. Instead the x-maintained is there for the few corner cases only, and should almost never be needed.

But I think it is worth exploring this space better. Let's have another discussion at the meeting tomorrow

shonfeder: We talked about this at the maintenance meeting today. I am convinced that the current solution is expedient, and the best on the table given the current social practices and the stat of the database and packing system.

If we had a more advanced package management UI, different packaging practices, or support for a different data layout in our existing tooling, then it would make sense for us to pursue a cleaner solution. But given the current state of all this, I am convinced that there will be more value to the ecosystem in adopting a sub-optimal solution that enables us to move forward with the effort than to block trying to achieve a more prefect implementation that would require non-trivial adaptations to our publication process and package tooling.

We also agreed to write a followup PR to clarify some of these points in the policy.

Thank you for the discussion and for your work, @hannesm and thanks for the discussion @mseri.

PR #27280, opened 2025-01-17
mark packages I marked as deprecated with x-maintenance-intent: ["(none)"]

hannesm:
since I didn't read the documentation back then, and the flags: deprecated won't help me with archiving.


mseri:
Thanks

PR #27262, opened 2025-01-14
[new release] carton (4 packages) (1.0.0)

dinosaure:
Implementation of PACKv2 file in OCaml

- Project page: <a href="https://git.robur.coop/robur/carton">https://git.robur.coop/robur/carton</a>
- Documentation: <a href="https://robur-coop.github.io/carton/">https://robur-coop.github.io/carton/</a>

##### CHANGES:

- New version of Carton

PR #27287, opened 2025-01-17
Prepare release carton.1.0.0 (add some bouds with incompatible packages

dinosaure:
This PR is not a release as it but it puts some contraints to smoothly merge then #27262 (which needs some fixes).


mseri:
Thanks

PR #27290, opened 2025-01-18
Fix conf-mad check for freebsd

toots:
This is according to: https://github.com/savonet/ocaml-mem_usage/issues/8#issuecomment-2599463702


mseri:
Is this a recent change in freebsd? I am mainly asking to understand if we should release this as conf-mad.2 or not

toots:
Not sure. Let's do a `.2` indeed.

shonfeder:
The only CI error is an incorrect lint on conf packages.

PR #27291, opened 2025-01-19
Package mem_usage.0.1.2

toots:
### `mem_usage.0.1.2`
Cross-platform stats about memory usage

---
* Homepage: https://github.com/savonet/ocaml-mem_usage
* Source repo: git+https://github.com/savonet/ocaml-mem_usage.git
* Bug tracker: https://github.com/savonet/ocaml-mem_usage/issues

---


toots:
This should be good to merge now.

I'm not exactly sure how to name the external dependency. I think that `conf-sysinfo` makes more sense than `conf-mem_usage` as it refers to a specific system library.

The library is available by default on linux systems and optional on freebsd systems (at least). At the moment the `conf-sysinfo` package is only available on `freebsd` but it could also be made available on linux systems as well.

Another option could be to explicitely rename it `conf-sysinfo-freebsd` or something like this.

mseri:
This seems fine to me, but I'll bring it up at the next maintainers meeting on Thursday

toots:
Sounds good no rush here.

mseri:
We think it is good as is, thanks a lot!

PR #27298, opened 2025-01-20
new file:   packages/mmo/mmo.1.0.1/opam

UnixJunkie:


UnixJunkie:
first realease of mmo: a few math modules which can be useful for molecular mechanics

shonfeder (approved):
Thank you for publishing your work for the community! The only CI failures are install issues with nlopt on some platforms. This seems to be an upstream problem, not an issue with this package, which installs fine on most targets.

PR #27322, opened 2025-01-27
relase lablgtk 2.18.14

garrigue:
This the long delayed release of lablgtk 2.18.14.
(Just forgot to push it)

### DIFF
--- a/packages/lablgtk/lablgtk.2.18.14/opam
+++ b/packages/lablgtk/lablgtk.2.18.14/opam
@@ -1,0 +1,62 @@
+opam-version: "2.0"
+maintainer: "garrigue@math.nagoya-u.ac.jp"
+authors: ["Jacques Garrigue et al., Nagoya University"]
+homepage: "https://garrigue.github.io/lablgtk/"
+bug-reports: "https://github.com/garrigue/lablgtk/issues"
+dev-repo: "git+https://github.com/garrigue/lablgtk.git"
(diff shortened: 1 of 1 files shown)


shonfeder:
Thank you for the update!

I notice that some of the revdeps errors seem to be relaed to lablgtk. E.g. (https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/0fa6b40613207ef373f2223148aa0b170f22dc36/variant/compilers,4.14,lablgtk.2.18.14,revdeps,frama-c.22.0):

```
# Error: The files /home/opam/.opam/4.14/lib/lablgtk3/gWindow.cmi
#        and /home/opam/.opam/4.14/lib/ocamlgraph_gtk/graph_gtk__DGraphContainer.cmi
#        make inconsistent assumptions over interface Pango
```

Do you know if these may indicate the need for an upper bound on those reverse dependencies?

garrigue:
lablgtk anf lablgtk3 are distinct libraries, but they use identical module names.
For this reason, one cannot use both in the same program.
Somehow it seems that this is what is done. Or maybe just a wrong package specification.
Anyway, this is an old version of Frama-C, newer versions seem to compile fine.

shonfeder:
Thank you for looking and explaining the situation! Sound like we are OK here then :)

shonfeder (approved):

PR #27331, opened 2025-01-29
[new release] ocamlmig (5.2-20250129)

v-gb:
Ocaml source code rewriting tool

- Project page: <a href="https://github.com/v-gb/ocamlmig">https://github.com/v-gb/ocamlmig</a>

##### CHANGES:

Initial release.

### DIFF
--- a/packages/ocamlmig/ocamlmig.5.2-20250129/opam
+++ b/packages/ocamlmig/ocamlmig.5.2-20250129/opam
@@ -1,0 +1,67 @@
+opam-version: "2.0"
+synopsis: "Ocaml source code rewriting tool"
+description:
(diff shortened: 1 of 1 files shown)


v-gb:
I'm confused about two things here:

- the code fails to build on windows due to the core_unix dependency, fine. But why is the build even attempted ? Ideally, opam would say "this is not compatible with windows", and the windows CI would say "ok, nothing to see here". Aren't there missing constraints on the core_unix package?
- I have changes to a library, here ocamlformat. I'm using a pin to build it, but it's rejected here. So I looked for a standard way to do this, but found nothing.  Is there a standard way to use a modified dependency? Or just manually copying the dependency into my repository?

PR #27338, opened 2025-01-30
Release ppxlib and ppxlib-tools 0.35.0

NathanReb:
As usual, starting with a draft to see the revdeps builds.

**CHANGES:**

- Allow use of effect syntax in preprocessed files without causing migration failures. (ocaml-ppx/ppxlib#552, @hhugo, @NathanReb)

- Remove support for compilers older than 4.08 (ocaml-ppx/ppxlib#556, @NathanReb)

- Add a --use-compiler-pp flag to the standalone driver. This flag can be set when the driver outputs source code to make it use the compiler's Pprintast instead of ppxlib's. (ocaml-ppx/ppxlib#555, @NathanReb)


NathanReb:
This should be good to go now!

PR #27349, opened 2025-01-31
Compiler unmaintained

hannesm:
instead of using "deprecated", specify x-maintained: false -- so in our phase 3 we can clean these up.


mseri:
Thanks

PR #27353, opened 2025-02-03
[new release] ocamlmig (5.2-20250202)

v-gb:
OCaml source code rewriting tool

- Project page: <a href="https://github.com/v-gb/ocamlmig">https://github.com/v-gb/ocamlmig</a>
- Documentation: <a href="https://github.com/v-gb/ocamlmig/blob/main/README.md">https://github.com/v-gb/ocamlmig/blob/main/README.md</a>

##### CHANGES:

Correct a failure to use some opam-installed files.

### DIFF
--- a/packages/ocamlmig/ocamlmig.5.2-20250202/opam
+++ b/packages/ocamlmig/ocamlmig.5.2-20250202/opam
@@ -1,0 +1,69 @@
+opam-version: "2.0"
+synopsis: "OCaml source code rewriting tool"
+description:
+  "Ocamlmig is a command line tool to rewrite ocaml source code, especially to make updating to newer interfaces easier."
+maintainer: ["Valentin Gatien-Baron <valentin.gatienbaron@gmail.com>"]
+authors: ["Valentin Gatien-Baron <valentin.gatienbaron@gmail.com>"]
+license: "LGPL-2.1-only WITH OCaml-LGPL-linking-exception"
+homepage: "https://github.com/v-gb/ocamlmig"
+doc: "https://github.com/v-gb/ocamlmig/blob/main/README.md"
+bug-reports: "https://github.com/v-gb/ocamlmig/issues"
+depends: [
+  "ocaml" {>= "5.2" & < "5.3"}
+  "dune" {>= "3.15"}
+  "base"
+  "core"
+  "core_unix"
+  "csexp"
+  "ppx_partial"
+  "ocaml" {>= "4.08"}
+  "alcotest" {"1" = "0" & >= "1.3.0"}
+  "base" {>= "v0.12.0"}
+  "cmdliner" {>= "1.1.0"}
+  "dune"
+  "dune-build-info"
+  "either"
+  "fix"
+  "fpath" {>= "0.7.3"}
+  "menhir" {>= "20201216"}
+  "menhirLib" {>= "20201216"}
+  "menhirSdk" {>= "20201216"}
+  "ocaml-version" {>= "3.5.0"}
+  "ocamlformat-rpc-lib" {"1" = "0" & = version}
+  "ocp-indent" {"false" = "false" & >= "1.8.0" | "1" = "0" & >= "1.8.1"}
+  "stdio"
+  "uuseg" {>= "10.0.0"}
+  "uutf" {>= "1.0.1"}
+  "csexp" {>= "1.4.0"}
+  "astring"
+  "camlp-streams"
+  "re" {>= "1.10.3"}
+  "odoc" {with-doc}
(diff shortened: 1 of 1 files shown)


packages/ocamlmig/ocamlmig.5.2-20250202/opam
```diff
@@ -0,0 +1,69 @@
...
+depends: [
+  "ocaml" {>= "5.2" & < "5.3"}
+  "dune" {>= "3.15"}
+  "base"
+  "core"
+  "core_unix"
+  "csexp"
+  "ppx_partial"
+  "ocaml" {>= "4.08"}
+  "alcotest" {"1" = "0" & >= "1.3.0"}
```
mseri, line 20:
```suggestion
  "alcotest" {"1" = "0" & >= "1.3.0"}
```
Why you have these funny bounds? If you want you could use `{with-dev-setup}` and make them available only when invoking opam with that flag

v-gb: The package vendors (a modified copy of) ocamlformat, which means the dependencies section of my opam file copies the dependencies section from ocamlformat's opam file, which is where the alcootest dependency and others come from.
In the ocamlformat package, alcootest is listed as `{with-dev-setup & >= "1.3.0"}`, but that doesn't make sense for my package, since my tests don't use alcootest. So I mechanically replace all the `with-dev-setup` that come from ocamlformat with `false`. But since syntactically that's not allowed (not sure if that's a shortcoming of dune-project or opam itself), instead of `false`, I use `"1" = "0"` which means the same thing :/.

v-gb: The import of ocamlformat's dependencies is done by a script. Dropping the dependencies entirely would involve parsing opam dependencies, which is much more involved than a call to sed.

v-gb: Actually it would involve parsing ocamlformat's dune-project, but the general point is the same.

mseri: I meant more if it was ok for me to remove them here :)

v-gb:
I suppose this is fine, but the previous release already has them, and
future releases will keep adding them back, so that seems like unnecessary
manual work.

Le mer. 5 févr. 2025, 10:37, Marcello Seri ***@***.***> a
écrit :

mseri:
It does not take much, and avoid unnecessary checks in the dependency solver. I'll keep an eye out and remove it also in future releases

mseri:
By the way, why not leaving it with `with-dev-setup` instead? Opam will practically ignore them, and they will not be installed in any switch anyway unless you specifically use that flag

v-gb:
The ocp-indent change is not correct though.

packages/ocamlmig/ocamlmig.5.2-20250202/opam
```diff
@@ -0,0 +1,66 @@
...
+  "dune"
+  "dune-build-info"
+  "either"
+  "fix"
+  "fpath" {>= "0.7.3"}
+  "menhir" {>= "20201216"}
+  "menhirLib" {>= "20201216"}
+  "menhirSdk" {>= "20201216"}
+  "ocaml-version" {>= "3.5.0"}
+  "stdio"
```
mseri, line 31:
```suggestion
  "ocp-indent" {>= "1.8.0"}
  "stdio"
```

v-gb:
> By the way, why not leaving it with `with-dev-setup` instead? Opam will practically ignore them, and they will not be installed in any switch unless you specifically use that flag

I'll likely want to use with-dev-setup for my own purpose.
If you'd be fine if the opam file said with-dev-setup, why would a static condition be more expensive?

mseri:
I think the flags are filtered a priori. But that is not really my point, it was more to cleanup the dependencies and to have filters that are easier to read. I am going to open an issue in dune to support `:false`.

For `ocp-indent` you could just have the bound without `"false" = "false"` though.

v-gb:
Thanks, I'll try that!

v-gb:
So I [tried this](https://github.com/v-gb/ocamlmig/compare/main...push-nmwnrkzvpyyt]), and even though the opam files look fine, dune itself doesn't understand what's going on, which causes `dune pkg lock` to drop the dependency on ocp-indent (I assume because `dune` treats the variable "true" as unset, hence false), which breaks the build.

mseri:
For the true case, the true should be completely unnecessary though, you only need “version bound or (false and version bound)”. Am I wrong?

PR #27355, opened 2025-02-03
Package synopsis.0.2

ptrichr:
### `synopsis.0.2`
Gather information about module usage, top-level bindings, and function calls

---
* Homepage: https://github.com/umd-cmsc330/restriction-checker
* Source repo: git+https://github.com/umd-cmsc330/restriction-checker.git
* Bug tracker: https://github.com/umd-cmsc330/restriction-checker/issues

---
:camel: Pull-request generated by opam-publish v2.5.0


mseri:
As it stands this packages does not install anything

ptrichr:
Ah. I think I've been approaching my issue the wrong way. I've been trying to figure out a way to make this piece of code globally available so that it can be used across all my projects without the need of duplicating the source file each time, and I figured a package might be a solution but I think I've been misunderstanding how packages work. Do you think a library/archive would be a better solution?

ptrichr:
Thanks for the input! I'll publish a new package release.

ptrichr:
Thanks for the input! I will publish a new package with the changes.

mseri:
Yes, publishing this package as a library is fine. But indeed, as shown in @zapashcanon PR (thanks a lot!), you need to export the library that you would like to see installed

mseri:
Feel free to submit a new release with the change in place

PR #27360, opened 2025-02-03
release ocaml-rocksdb-0.02

chetmurthy:
This ocaml-rocksdb interface is built using cppffigen, a tool that generates OCaml FFI bindings to C++ libraries that follow the Google C++ Style Guide for memory-ownership.

This might not be the best name for this package.  There are already "orocksdb" and "ahrocksdb" packages (both based on ctypes); if you think this should be named something else, I'd be happy to change it.


packages/conf-rocksdb-ge-5-17-2/conf-rocksdb-ge-5-17-2.5.17.2/opam
```diff
@@ -0,0 +1,37 @@
...
+  (("host-arch-x86_32" {os = "win32" & os-distribution != "cygwinports"} & "conf-mingw-w64-zstd-i686" {os = "win32" & os-distribution != "cygwinports"}) |
+   ("host-arch-x86_64" {os = "win32" & os-distribution != "cygwinports"} & "conf-mingw-w64-zstd-x86_64" {os = "win32" & os-distribution != "cygwinports"}))
+]
+build: [
+
+  ["pkgconf" {os = "win32" & os-distribution = "cygwin"}
+   "--personality=i686-w64-mingw32" {os = "win32" & os-distribution = "cygwin" & host-arch-x86_32:installed}
+   "--personality=x86_64-w64-mingw32" {os = "win32" & os-distribution = "cygwin" & host-arch-x86_64:installed}
+   "pkg-config" {os != "win32" | os-distribution != "cygwin"}
+   "--atleast-version=7.8.3" "rocksdb"] {
```
raphael-proust, line 21: is this `7.8.3` constant correct? what's its relation to the "ge-5-17-2" aim?

packages/conf-rocksdb-ge-5-17-2/conf-rocksdb-ge-5-17-2.5.17.2/opam
```diff
@@ -0,0 +1,37 @@
...
+    !(os-distribution = "debian" & (os-version = 11 | os-version = 12)) &
+    !(os-distribution = "ubuntu" & (os-version = "20.04" | os-version = "22.04"))
+   }
+]
+synopsis: "Virtual package relying on a system installation of RocksDB (version >= 5.17.2)"
```
raphael-proust, line 26: I understand the need for this kind of package, and I wish we had a good solution (or at least a standard solution) to offer but right nnow it's an unsolved problem for the opam repo.

I'm not very happy with adding version _constraints_ as part of the name of a package. Although I can see how it's needed and how we can't reasonably generate conf-rocksdb packages for each available version of rocksdb. Or maybe that's reasonable but would require having a separate dedicate opam repository.

Anyway… Anyone has ideas?

Otherwise I guess we can possibly merge for now and possibly revisit.

packages/ocaml-rocksdb/ocaml-rocksdb.0.02/opam
```diff
@@ -0,0 +1,59 @@
+
```
raphael-proust, line 1: Normally we avoid having ocaml as part of the name of packages. See https://github.com/ocaml/opam-repository/tree/master/governance/policies#7-new-package-names-should-avoid-the-ocaml-prefixsuffix

I don't think this porject necessarily needs an exception. (Unless you can make a case for it?)

chetmurthy: I'm happy to make it "rocksdb" ?  Or any other name you would propose *grin*.

packages/conf-rocksdb-ge-5-17-2/conf-rocksdb-ge-5-17-2.5.17.2/opam
chetmurthy: It's actually worse and worse.  When I started with trying to get this to work, I foolishly assumed that `pkg-config` would be available to check rocksdb's package-version.  But (haha, silly me) that wasn't the case with these older versions of rocksdb like (sigh) current Debian (11, 12) and backlevel Ubuntu (20.24, 22.04).  And they were all backlevel versions of Rocksdb (5.17.2, 6.11.4).   Other platforms had 7.8.3, though a few even failed the check.

I could put these changes into conf-rocksdb ?

The bigger problem of not being able to state version-constraints is .... painful.  Some packages (like Rocksdb) appear to be changing their API pretty rapidly.  But OTOH this is a big change to opam, and I sure think there are better ways to spend developer resource than on this.

shonfeder:
Quite a few failing CI checks here on various platforms. Sounds like we'll need investigation there, in addition to the naming question. `rocksdb` seems fine, IMO.

Marking this accordingly.

packages/conf-rocksdb-ge-5-17-2/conf-rocksdb-ge-5-17-2.5.17.2/opam
raphael-proust: I think that we can accept the `ge-…` name for now

And maybe someone with time available can write a generator for many conf-packages where the version number is just substituted.

PR #27376, opened 2025-02-05
Add rust 2024 edition

mseri:
It is now being stabilized (see https://github.com/rust-lang/rust/pull/133349)

May need an update to the test.rs and distributions once it is out, but will help with packages that need recent rust


mseri:
This fails everywhere for now, it will start succeeding after the upcoming rust release

